PR #3238, opened 2026-07-17
perf(model-version): cache public donationGoals to cut DB load

ZacxDev:
## Lever

`model-version.donationGoals` (`publicProcedure`, backs the donation-progress display on the model version page) is **uncached** and runs 2–3 DB queries per call — `modelVersion.findFirstOrThrow` + `donationGoal.findMany` + a `Donation` `SUM` — at ~31 calls/s peak, i.e. **~60–75 DB queries/s** removed at peak.

## The leak axis: public vs. privileged

The response **varies by viewer**:

- **Owner / moderator** (`canSeeAllGoals = userId === model.userId || isModerator`): sees **ALL** goals, including inactive/draft ones (`active`/`isEarlyAccess` filters dropped).
- **Public** (anonymous / logged-in non-owner / non-moderator): sees only `active: true` goals, plus a per-version early-access filter (`isEarlyAccess` goals shown only while the version still has `earlyAccessEndsAt` set).

The public variant is **fully determined by `modelVersionId`** (it does not vary by viewer), so a single shared cache entry is correct for every non-privileged caller. The privileged variant must **never** enter that shared key.

## What's cached + how the split prevents a leak

New `modelVersionPublicDonationGoalsCache` (`createCachedObject`, keyed by `modelVersionId`, **TTL 60s** (`CacheTTL.xs`), no SWR, `cacheNotFound: false`). Its `lookupFn` reproduces the **public** computation exactly: existence + `earlyAccessEndsAt` (with the same replica→primary fallback as the uncached read), the `active: true` filter, the per-version early-access filter, and the summed totals. It seeds an **empty** entry for every existing version (so "exists, no goals" → `[]`) and **no** entry for a missing version (→ 404).

`modelVersionDonationGoals` gates on the leak axis **before** touching the cache:

1. **Anonymous** (`!isModerator && userId == null`) — can be neither owner nor mod, so it's unconditionally public → served from the cache with **zero per-viewer DB reads** (full savings).
2. **Logged-in** — resolve the version to learn the owner (also the 404 / replica-lag path). If **not** owner and **not** mod → public → same shared cache (still elides the `donationGoal.findMany` + `Donation` `SUM`). If owner/mod → **privileged path computed fresh, never reads from or writes to the cache.**

So a cached public payload can never be served to a privileged viewer, and an owner's draft-inclusive payload can never be written into the shared key. (Mirrors #3223 keeping per-user data out of the shared cache.)

## TTL + bust

[...]

PR #3258, opened 2026-07-20
fix(donation-goal): fail-open the EA-completion cache side-effects (double-donation guard)

ZacxDev:
## The double-donation mechanism

`checkDonationGoalComplete` runs **inside** `donateToGoal`'s `try` block, AFTER `donation.create` has already committed the donation row. `donateToGoal`'s `catch` refunds the buzz and throws `"Failed to create donation"` on **any** error. So any throw from a post-commit operation inside `checkDonationGoalComplete` produces: donation row persists + buzz refunded + donor told it failed → **donor retries → double donation / double charge**.

This is the same class #3238 caught and fixed for the public-cache bust. This PR closes the two remaining un-guarded operations in the same function.

## What was guarded (2 ops)

In the early-access-completion branch (`if (goal.isEarlyAccess && goal.modelVersionId && isGoalMet)`), two non-critical redis/cache side-effects were un-guarded and reject on a transient redis error:

- `await bustMvCache(goal.modelVersionId, modelVersion.modelId)`
- `await dataForModelsCache.refresh(modelVersion.modelId)`

Both are now wrapped in a **fail-open try/catch that logs-and-continues** via `logToAxiom` (`name: 'donation-goal-ea-cache-refresh-failed'`), mirroring the #3238 public-cache-bust guard just below — including the trailing `.catch(() => undefined)` so the logger itself can't throw. On failure the caches serve briefly-stale data; the donation stands. (`updateModelEarlyAccessDeadline` already had its own `.catch`; the public-cache bust below was already guarded by #3238.)

## Why the DB writes are NOT guarded

The two `dbWrite` operations in this branch — `donationGoal.update` (marks the goal complete) and the `$executeRaw` EA-deadline update (unlocks early access) — are **legitimate state changes**, not fail-open candidates. A DB-write failure is a genuine error; silently dropping goal-completion / EA-unlock would be its own bug. They are deliberately left outside the guard so they still propagate. The double-donation risk being closed is specifically the **transient redis/cache** class.

## Structural-fix recommendation (follow-up, not this PR)

The deeper issue is that `checkDonationGoalComplete` runs inside `donateToGoal`'s refund-`try` at all — so even a *legitimate* post-commit DB-write failure (`donationGoal.update` / `$executeRaw`) still triggers a refund of a committed donation. A more robust long-term fix would move `checkDonationGoalComplete` **out** of the refund-`try` (the donation + buzz transaction are already durable at that point), so **no** post-commit side-effect — cache or DB — can refund a committed donation; a completion-side failure would then surface as an error without touching the money that already changed hands. That is a larger behavioral change and out of scope here; this PR does the proportionate, low-risk cache-fail-open fix matching #3238.

## Tests (9 pass: 3 pre-existing + 6 new)

- `bustMvCache` rejects → `checkDonationGoalComplete` still resolves, returns the goal, logs `donation-goal-ea-cache-refresh-failed`.
- `dataForModelsCache.refresh` rejects → same (resolves + logs).
- Both EA cache side-effects run on completion with the right args `(modelVersionId, modelId)` / `(modelId)`.
- The goal-completion DB write + `$executeRaw` still run (guard is scoped to cache ops only).
- `donationGoal.update` DB failure **propagates** (rejects) — completion write is NOT fail-open; cache guard never runs.
- `$executeRaw` EA-deadline DB failure **propagates** (rejects) — EA-unlock write is NOT fail-open.

## Verification

- `vitest run --project unit` on the suite: **9/9 pass**.
- `tsc --noEmit`: **zero new errors** vs `origin/main` (identical error set with the diff reverted).

### DIFF
--- a/src/server/services/__tests__/donation-goal.service.test.ts
+++ b/src/server/services/__tests__/donation-goal.service.test.ts
@@ -14,22 +14,33 @@ const { mockDbRead, mockDbWrite } = vi.hoisted(() => {
     mockDbWrite: { donationGoal: mk(), $queryRaw: vi.fn(), $executeRaw: vi.fn() },
   };
 });
-const { mockDonationGoalsBust, mockLogToAxiom } = vi.hoisted(() => ({
+const {
+  mockDonationGoalsBust,
+  mockLogToAxiom,
+  mockBustMvCache,
+  mockDataForModelsRefresh,
+  mockUpdateEaDeadline,
+} = vi.hoisted(() => ({
   mockDonationGoalsBust: vi.fn(),
   mockLogToAxiom: vi.fn(),
+  mockBustMvCache: vi.fn(),
+  mockDataForModelsRefresh: vi.fn(),
+  mockUpdateEaDeadline: vi.fn(),
 }));
 
 vi.mock('~/server/db/client', () => ({ dbRead: mockDbRead, dbWrite: mockDbWrite }));
 vi.mock('~/server/redis/caches', () => ({
-  dataForModelsCache: { refresh: vi.fn() },
+  dataForModelsCache: { refresh: mockDataForModelsRefresh },
   modelVersionPublicDonationGoalsCache: { bust: mockDonationGoalsBust },
 }));
 vi.mock('~/server/services/buzz.service', () => ({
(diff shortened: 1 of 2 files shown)


CivitaiDevOpsAgent:

PR #3228, opened 2026-07-17
feat(blocks): complete customComfy bridge v1 server feature (registry + router + timeout-cap budget belt)

ZacxDev:
## What

The **complete App Blocks `customComfy` bridge v1 server feature** — a bounded, server-authored ComfyUI recipe run end-to-end through the four existing block-bridge procedures, reusing the post-paid `$type:'customComfy'` orchestrator submit. The iframe sends only `{ kind:'customComfy', recipe, params }`; the civitai server owns the graph in full.

This PR now spans plan **PR2+PR3** (schema + recipe registry + graph translator + snapshot support — the inert building blocks) **and PR5+PR6** (router wiring + the **timeout-cap post-paid budget belt** — the security-critical crux), landed onto the one branch so the v1 server spine is reviewable as a whole.

**Stays DARK.** The four block-token procs are already `assertViewerIsAppDeveloper` / `assertAppBlocksEnabledForTokenUser`-gated, so recipe submissions are developer/mod-gated until GA — the empty-but-panorama recipe registry is itself the gate. No worker-image baking, no `$type:'comfy'` registration, no per-invocation orchestrator token — v1 stays **post-paid**.

## Building blocks (PR2+PR3)

- **Schema** (`workflow.schema.ts`): `blockCustomComfyBodySchema = { kind:'customComfy', recipe ∈ REGISTERED_RECIPE_IDS, params: z.record(...) }.strict()`; `blockWorkflowBodySchema` widened to a `discriminatedUnion('kind', [...])`. `recipe` is a coarse fail-closed gate; the per-recipe `.strict()` param schema is applied in the translator.
- **Recipe registry** (`src/server/services/blocks/recipes/`): the `BlockRecipe` interface + registry, `REGISTERED_RECIPE_IDS` derived from registry keys, and a fail-fast `maxBuzz == ceil(stepTimeoutSeconds)` invariant checked at module load. `seamless-pano-360` is recipe #1 (DiT engines `zimage-turbo` / `flux2-klein` / `qwen-image`), ported verbatim from `panorama.ts` as pure object-construction builders. Pinned checkpoint policy, `stepTimeoutSeconds = 180` → `maxBuzz = 180` (a deliberately tight blast-radius ceiling — the step `timeout` is the only deterministic per-job Buzz bound).
[...]

PR #3259, opened 2026-07-20
fix(app-blocks): customComfy submits write durable audit + attribution; dev-token defaults budget to manifest

ZacxDev:
## What

Two follow-up fixes to the App Blocks **customComfy** bridge, both surfaced by a live dogfood of #3228 (now on `main`).

### Fix 1 (primary — GA-gate audit gap): customComfy submits wrote no durable audit/attribution rows

The dogfood confirmed a successful customComfy generation billed real Buzz but wrote **no `block_scope_invocations` row** and **no `block_spend_attribution` row** — the `submitCustomComfyWorkflow` handler early-returned before both writes (the txt2img path writes both). Added to the customComfy branch, mirroring txt2img and reusing the same services:

- **`block_scope_invocations`** — the per-user activity-feed row. Fires after the orchestrator submit returns (`scope: 'ai:write:budgeted'`, `endpoint: workflow:submit:<id>`, spend detail, status-mapped code). The dev/ephemeral **synthetic-app** case (a pre-approval app carries a synthetic non-FK `appBlockId`) is handled by passing `dev: true`, so `recordScopeInvocation`'s FK-fallback persists `app_block_id NULL` + `synthetic_app_id`.
- **`block_spend_attribution`** — the author-bounty basis the PR6 author left out of scope. One **track-only** row per generation (`status='tracked'` / `rate_card_version='unrated'` / share 0), server-derived from the verified block-JWT (forge-safe), idempotent on `(workflow_id, app_block_id)`. The payout-safe currency basis (paid green/yellow vs free-first blue floor) is derived off the **realized** per-account debit via a new `deriveBlockSpendBasis` helper, so a future non-zero share only ever accrues off the user's paid portion. customComfy carries no `sharedContentKey` (its wire body is recipe+params `.strict()`), so the base track-row is written without it.

Both are fire-and-forget with their own try/catch. **The txt2img path is byte-unchanged** (additions only — verified: zero deleted lines in `blocks.router.ts`).

### Fix 2 (secondary — DX): dev-token default budget ignored the app's manifest budget

`resolveDevBuzzBudget` defaulted to a flat `DEV_BUZZ_BUDGET_DEFAULT = 50`, so a recipe/app whose manifest `page.buzzBudgetPerGen` exceeds 50 couldn't be dev-tested via dev-token/dev-tunnel without a manual budget request (the dogfood hit `ceiling 180 exceeds budget 50`).

The dev-token mint now **defaults the budget to the resolved app manifest's `page.buzzBudgetPerGen`** (clamped to `DEV_BUZZ_BUDGET_CAP = 250`) when a manifest is resolvable, via a new `parseManifestBuzzBudget` helper threaded through the endpoint's `resolved` object. Precedence: explicit caller `buzzBudget` > manifest `page.buzzBudgetPerGen` > flat `DEV_BUZZ_BUDGET_DEFAULT`. The `DEV_BUZZ_BUDGET_CAP` hard ceiling still clamps in every case.

**Mint-mode handling:**
- **Approved** mode — reads `page.buzzBudgetPerGen` from the approved `AppBlock.manifest`.
- **Pending** mode — reads it from the caller-owned pending request's un-reviewed manifest (raises the per-call budget only; spend authority is still bounded by the cap, the per-user daily cap, the dev-session cap, and the 7f `AIServicesWrite` bearer gate).
- **Ephemeral no-row** mode — no server manifest exists (the local `block.manifest.json` isn't sent on this path), so it keeps the flat default with a clear `TODO` to thread the CLI-sent manifest budget. Callers can still pass `buzzBudget` explicitly.

## Tests

- `blocks.router.workflow.test.ts` — a customComfy submit writes a `block_scope_invocations` row (incl. the dev `synthetic_app_id` case) + a `block_spend_attribution` track-row; paid-debit and net-refund currency-basis derivation; over-budget early-return writes neither; a failing attribution write never breaks submit.
- `dev-scoped-mint.service.test.ts` — `resolveDevBuzzBudget` defaults to the manifest budget (clamped), explicit request wins, falls back to 50 when absent; `parseManifestBuzzBudget` accepts a positive integer and rejects fractional/zero/negative/NaN/Infinity/non-object.
- `dev-token.test.ts` (endpoint) — approved + pending manifest budgets flow through and clamp; explicit request wins; omitted budget falls back to 50.

Touched suites pass: **213 + 34 + 81**. Full `NODE_OPTIONS=--max_old_space_size=8192 tsc --noEmit` runs clean (**0 errors**). The **pr-preview typecheck is the authoritative gate**.

## Notes

- Both issues were **dogfood-caught** against #3228.
- Do not merge yet — awaiting the pr-preview typecheck + review.

### DIFF
--- a/src/pages/api/v1/blocks/dev-token.ts
+++ b/src/pages/api/v1/blocks/dev-token.ts
@@ -12,6 +12,7 @@ import {
   clampDevScopes,
   DEV_TOKEN_SCOPE_ALLOWLIST,
   FORCED_SFW_CEILING,
+  parseManifestBuzzBudget,
   resolveDevBuzzBudget,
   signDevScopedPageToken,
 } from '~/server/services/blocks/dev-scoped-mint.service';
@@ -491,6 +492,13 @@ export default withAxiom(async (req: AxiomAPIRequest, res: NextApiResponse) => {
     signAppId: string;
     signAppBlockId: string;
     blockInstanceId: string;
+    // The RESOLVED app manifest's declared `page.buzzBudgetPerGen`, used as the
+    // DEFAULT dev-token budget (step 8) when the caller passed no explicit
+    // `buzzBudget`. Set on the approved + pending modes (a manifest is
+    // resolvable); UNDEFINED on the ephemeral no-row mode (no server manifest →
+    // resolveDevBuzzBudget falls back to DEV_BUZZ_BUDGET_DEFAULT). Clamped to
+    // DEV_BUZZ_BUDGET_CAP by resolveDevBuzzBudget regardless.
+    manifestBudgetDefault?: number;
   };
   let resolved: Resolved | null = null;
   // Set ONLY on the local-manifest (pending) path so the structured mint-audit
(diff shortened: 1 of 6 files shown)


CivitaiDevOpsAgent:

PR #3236, opened 2026-07-17
perf(auction): cache getAll to cut primary-DB load

ZacxDev:
## Lever

`auction.getAll` (`publicProcedure` → `getAllAuctions`, ~21.8 calls/s at peak) ran **uncached** against the **PRIMARY DB** (`dbWrite.auction.findMany`) on every call. Its output is a single global `{ id, auctionBase, lowestBidRequired }[]` with **no per-user/ctx variance** (the `requiredScope` gates ACCESS, not the payload), so it's an ideal read-through cache target.

## STEP 0 — the commented-out `edgeCacheIt`

`auction.router.ts` carries a `// .use(edgeCacheIt({ ttl: CacheTTL.hour }))` line. I traced its history (`git log -p -S edgeCacheIt`): it has been **commented-out since the file was first created** (Feb 2025, "all the auction things!") — it was **never enabled and never disabled**, i.e. there is **no correctness reason on record** for keeping edge-cache off; it's a leftover placeholder.

That line is a **public CDN edge-cache** anyway (a different mechanism), and it's left untouched here. This PR adds a **server-side** short-TTL read-through, which is safe independent of the edge-cache question: I confirmed the payload has no per-user/ctx variance (the function takes no args and reads no `ctx`), so one global entry is correct for every caller.

## Mechanism / TTL

A 30s `fetchThroughCache` under a new global key `REDIS_KEYS.CACHES.ACTIVE_AUCTIONS` (`packed:caches:active-auctions`). The origin fetch is factored out as `getAllAuctionsUncached` (exported so tests can assert a cache hit skips it).

- **Short TTL (30s), no bust.** The frequently-mutating input is **bids** (each shifts `lowestBidRequired`), so busting per-bid at ~21/s would defeat the cache. Auction create/close happens in the daily `handle-auctions` job and is already bounded by the 30s TTL. A short TTL is the right trade here vs. full invalidation.
- **Fails OPEN.** `fetchThroughCache` degrades to a slow-but-correct origin fetch on a Redis error, so this adds **no new failure mode** over the uncached path.

## Byte-identical / staleness (Bucket A)

Output is **byte-identical** to the uncached path. The only delta is **≤30s staleness**, and it's **display-only**:

- The active set is **time-windowed** (`startAt <= now < endAt`); a ≤30s lag in an auction appearing/disappearing is imperceptible.
[...]

PR #3260, opened 2026-07-20
perf(auctions): unblock SSR, de-quadratic My Bids, trim payload

JustMaier:
/auctions took ~20s to show anything. Everything below was measured against prod (replica for EXPLAIN, anon curl for page timings) before any code changed — including the one finding that reversed the original plan.

## What was actually slow

**`getMyBids` — this is the 20s.** `where: { userId, deleted: false }` had no time bound, and for every bid it selected the entire auction, meaning every bid of that auction. Per-user hydration blowup:

| userId | their bids | bid rows hydrated |
|---|---|---|
| 2515131 | 829 | **522,307** |
| 9261462 | 729 | 443,008 |
| 6357 | 676 | 420,483 |
| 8067035 | 4,734 | 301,758 |

Then `prepareBids` re-ran over that whole set once per bid (~522k iterations + 829 sorts for the worst user), then entity hydration ran over all 829 entities. On the primary DB, awaited serially inside `getServerSideProps`.

**`getAll` was not slow.** The original plan called a SQL aggregate here the primary win. It is 8.9ms: 12 active auctions hold 629 bids *total*, and it already sits behind a 30s Redis read-through from #3236. That work item was dropped.

**`getBySlug`** is 0.75s / 817KB on prod for featured-checkpoints. The SQL is 12.5ms — the cost is payload.

## Changes

**1. Unblock SSR + read from the replica + skeletons**
`getServerSideProps` did three serial tRPC prefetches before shipping any HTML. It now does only what must precede render: the `features.auctions` 404 gate and the My Bids login redirect. Auction validity is derived client-side from the `getAll` list, so an unknown slug still shows the invalid-auction alert. Read paths move to `dbRead` (the `TODO switch back to dbRead` markers were from 2025-04-21); mutation read-your-write stays on `dbWrite`. Skeletons replace spinners.

**2. `getMyBids` aggregates in SQL**
Per-entity totals, ranking and the winning threshold come back from one query returning a single row per bid the user placed. **126ms for 1,780 rows** on the replica. Past bids are bounded to 90 days — they were unbounded and the tail is closed auctions from months ago (heaviest bidder: 4,734 → 1,780 rows).

Parity was checked by running the old JS algorithm and the new query side by side over **2,655 real bids from the five heaviest bidders: 0 mismatches** on position, total, threshold and additional-price-needed. Ranking gains `entityId` as a final tiebreak so entities tied on total *and* bid count stop swapping positions between requests.

**3. Payload trim**
Three fields no auction card reads accounted for 175KB (21%) of the 817KB response: image tags (83KB), profile-picture metadata (48KB), image metadata (44KB). Tags came from the wrapper that exists to attach them, so hydration now reads the image cache directly and skips that second cache round-trip. Both metadata blobs are nulled; response types are unchanged. Entity lookup also moves from a `find()` per bid to a Map (was O(n²) over 428 entities).

## Verification

- 13 unit tests (`auction-my-bids.test.ts` new, `auction-getall-cache.test.ts` updated for the dbRead swap)
- Old-vs-new parity across 2,655 real bids, 0 mismatches
- `/auctions`, `/auctions/<slug>` and an unknown slug all verified on a dev server; SSR HTML is now a constant ~113KB shell instead of 951KB for checkpoints
- typecheck + lint + prettier clean

## Notes

- The remaining bulk of `getBySlug` is creator cosmetics (208KB), which the cards do render. Deduplicating the 167 distinct creators across 428 rows would cut most of it but changes the response shape — left alone.
- The "1000-model cap" from the original brief is an auction quantity config knob, not a perf issue. Tracked separately.

ClickUp: [868ke490t](https://app.clickup.com/t/868ke490t)

### DIFF
--- a/src/components/Auction/AuctionBidList.tsx
+++ b/src/components/Auction/AuctionBidList.tsx
@@ -1,11 +1,10 @@
 import { Center, Divider, Text } from '@mantine/core';
-import { useVirtualizer } from '@tanstack/react-virtual';
-import { useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useMemo } from 'react';
 import { ModelPlacementCard } from '~/components/Auction/AuctionPlacementCard';
-import { useScrollAreaRef } from '~/components/ScrollArea/ScrollAreaContext';
+import { VirtualRowList } from '~/components/Auction/VirtualRowList';
 import { useIsMobile } from '~/hooks/useIsMobile';
-import type { GenerationResource } from '~/shared/types/generation.types';
 import type { GetAuctionBySlugReturn } from '~/server/services/auction.service';
+import type { GenerationResource } from '~/shared/types/generation.types';
 
 type Bid = GetAuctionBySlugReturn['bids'][number];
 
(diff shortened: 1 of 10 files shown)


CivitaiDevOpsAgent:
### [PR-REVIEW] round 5 — 0 finding(s) (confidence ≥ 0.80)

PR #3255, opened 2026-07-20
feat(buzz): CSV export + multi-account transaction history

JustMaier:
Adds a self-service CSV export to the existing Transaction History page (`/user/transactions`), plus multi-account selection for the list itself.

ClickUp: [868ke491n](https://app.clickup.com/t/868ke491n)

## What changed

- **Account switcher → multi-select chips.** Blue / green / yellow can be viewed and exported together. All chips on = all accounts.
- **Export CSV button.** One server call, rate-limited to 10/hour, filename `civitai-transactions_<accounts>_<from>_<to>.csv`.
- **List + export both read from ClickHouse.** The buzz service takes a single account type and caps at 200 rows per call, so a multi-account view or an unbounded export would need hundreds of sequential round-trips.
- **Filter row relayout.** From, To and Type now share one line.

Columns: `date` (UTC ISO 8601), `type`, `amount`, `accountType`, `fromUser`, `toUser`, `description`, `entityUrl`, `externalTransactionId`. Counterparties resolve to usernames via a single batched query, falling back to the raw id; the central bank (account 0) renders as `Civitai`.

## Perf

Heaviest account on the platform, 3 account types over 90 days = 64k rows.

| Approach | Rows read | Server time |
|---|---|---|
| `OR` across from/to | 105M | 4.4s |
| `UNION ALL` per direction | — | ~0.5s |
| Buzz service | 320 sequential paged calls | not close |

The table's sort key is `(date, fromAccountId, toAccountId, transactionId)`, so an `OR` spanning both sides of a transaction degrades to a full range scan. One branch per direction lets each hit its own projection (`byFromAccount` / `byToAccount`).

## Notes for review

Two things about the ClickHouse table differ from what the spec assumed, and the code follows the table:

- `amount` is stored as a **positive magnitude**. Direction lives in which side of the transaction the account sits on, so the sign is computed server-side.
- `externalTransactionId` is **frequently empty** (all generation transactions). Dedup keys on `transactionId` (`LIMIT 1 BY transactionId`), not `externalTransactionId`.

Other decisions worth a second opinion:

- **The list is now ClickHouse-sourced, not buzz-service-sourced.** CH is a synced copy, so a just-made transaction can lag a few seconds behind the balance. Signed off by Justin.
- **Cursor changed from `Date` to `date|transactionId`.** A plain `date <` cursor silently skipped rows at page boundaries — generation bursts routinely share a second.
[...]

PR #3263, opened 2026-07-20
fix(flipt-skill): speak Flipt v2, refuse writes, correct the token docs

JustMaier:
The `flipt` skill was speaking v1 to a v2 server, so every read 404'd — `list` and `get` failed against flags that plainly existed, and the CLI reported `environment: "" not found`. Found while trying to verify the `buzz-transaction-export` kill switch from #3255.

## Reads

v2 addresses flags as resources under an environment and namespace:

```
/api/v2/environments/{env}/namespaces/{ns}/resources/flipt.core.Flag[/{key}]
```

Both are **discovered at runtime** (environment marked `default: true`, then its `default` namespace) rather than hardcoded — in v2 the routing key is `environments.<key>.name`, not the YAML map key, and the two are kept in sync by hand, so discovery is safer than pinning. `FLIPT_ENVIRONMENT` / `FLIPT_NAMESPACE` override. Nothing is pinned to a 2.10-exact shape, since Flux lands patch releases automatically within 12h.

Responses carry a `revision` equal to the `flipt-state` commit SHA — the quickest way to confirm a flag change has actually synced (~30s poll interval).

## Writes are refused, deliberately

Not repointed at v2. **Confirmed** from the talos-infra manifest: there is no Flipt auth behind the ingress, the environment is git-backed, and an SSH deploy key for `civitai/flipt-state` is mounted into the pod. **Inferred, not observed:** that a v2 write therefore commits and pushes to that repo's `main`. That follows from v2's documented git-write model, but nobody has watched it happen and we're not going to find out against production.

The guard holds under either reading — a write path that *may* land an unreviewed commit on a file gating production behavior is reason enough to refuse. Write commands exit 2 and print the PR workflow.

## Docs

`.env.example` described `FLIPT_API_TOKEN` as a "read-only fetcher token" that "cannot create/modify" flags. It is neither: it's a Traefik ingress bypass header, and Flipt applies no authz behind it. Both halves were wrong in the dangerous direction, so the comments now say what it actually is and that it should be treated as a live secret.

## Testing

```
get buzz-transaction-export  -> renders, exit 0
list                         -> 94 flags, exit 0
list --json                  -> revision matches the flipt-state SHA
get <nonexistent>            -> "Flag not found", exit 1
disable <flag>               -> refuses with GitOps steps, exit 2 (no API call made)
node --check                 -> clean
```

No writes were issued against production Flipt at any point.

### DIFF
--- a/.claude/skills/flipt/.env.example
+++ b/.claude/skills/flipt/.env.example
@@ -9,7 +9,7 @@ FLIPT_URL=https://flipt.civitai.com
 
 # Traefik ingress bypass header — NOT a Flipt credential and NOT read-only.
 # Flipt runs with authentication.required: false behind this, so anything that
-# reaches it has full API access, including the git-backed write path that can
+# reaches it has full API access, including a git-backed write path that may
 # push straight to civitai/flipt-state main. Treat as a live secret.
 # Change flags by PR to that repo, not through the API.
 FLIPT_API_TOKEN=your-bypass-header-here
--- a/.claude/skills/flipt/.env.example
+++ b/.claude/skills/flipt/.env.example
@@ -7,11 +7,17 @@
 # The base URL of your Flipt instance
 FLIPT_URL=https://flipt.civitai.com
 
-# Flipt API token
-# A token with read/write access to manage flags
-# For read-only operations, use the fetcher token
-# For write operations, you need an admin token
-FLIPT_API_TOKEN=your-api-token-here
+# Traefik ingress bypass header — NOT a Flipt credential and NOT read-only.
+# Flipt runs with authentication.required: false behind this, so anything that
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:

- **medium** `.claude/skills/flipt/flipt.mjs` (correctness, conf 0.95) — getFlag catches ALL errors from apiRequest and blindly rethrows them as "Flag not found", masking authentication failures (401/403), network errors, and server errors (5xx) behind a misleading message.  
  ↳ _Only rethrow as "Flag not found" for 404 responses; let all other errors propagate so the caller sees the real failure. Check err.message for 404 status, or better, have apiRequest attach a status code to the error object._

PR #3253, opened 2026-07-20
feat(donation-goals): let creators hide public collected amount

JustMaier:
## What & why

Creator feedback (MNeMiC): the donation-goal progress bar publicly exposes how much a creator has collected/earned, which is especially unwanted on models whose Early Access has already ended. This lets creators opt out of that public display while the donation goal keeps working (it still triggers the EA unlock).

## Changes

1. **Per-account opt-out** — new `hideDonationGoals` user setting. When ON, non-owner/non-mod viewers see none of the creator's donation-goal displays. Owners and moderators still see them.
2. **Auto-hide ended goals** — an early-access donation goal is now hidden from the public once its early-access window has ended (missing or past `earlyAccessEndsAt`), not just when the field is cleared.
3. **Two entry points for the toggle**:
   - Account settings → Creator Preferences (feature-gated behind `donationGoals`).
   - The model-version Early Access config, right where the donation goal is set (reflects/writes the same per-account setting).

## Implementation notes

- Both filters live in `publicDonationGoalsLookupFn` (the lookup behind `modelVersionPublicDonationGoalsCache`), which serves only the public (non-owner, non-mod) variant. The privileged owner/mod path is uncached and unchanged, so creators and mods still see everything, and the goal + its EA-unlock trigger are untouched.
- No DB migration — reuses the JSON `User.settings` column. Change visibility is bounded by the cache's existing 60s (`CacheTTL.xs`) TTL, so no extra cache-busting was added.
- Added unit tests for the owner opt-out and the ended-EA hide; existing cache tests updated for the new owner-settings lookup. `pnpm typecheck` and the affected vitest files pass.

## How to verify

- As a creator, toggle "Hide my donation goals from public view" in Account settings (or in the model-version Early Access config). View one of your models in an incognito/logged-out window: the "Support this model" donation-goal block should disappear; as the owner or a moderator it still shows.
- Set/keep an early-access donation goal and let (or set) its early-access window to the past: the goal stops showing publicly (within ~60s of cache TTL) while remaining visible to owner/mods.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3266, opened 2026-07-21
feat(creator-controls): hide model metrics (buzz/downloads/generations) for CP members

JustMaier:
## Creator Controls: model metric privacy

Lets Creator Program members hide three public model metrics — **tipped/earned Buzz, download count, generation count** — extending the just-shipped donation-goal opt-out pattern.

### 3-level control hierarchy (all in this PR)
1. **USER default** — new "Creator Controls" card in User Settings (`src/pages/user/account.tsx` → `CreatorControlsCard.tsx`). Baseline for all the creator's models. Flags stored on `User.settings` (`hideModelBuzz` / `hideModelDownloads` / `hideModelGenerations`) via `userSettingsSchema` + `setUserSettingsInput`.
2. **MODEL level** (primary) — new section in the model edit form (`ModelUpsertForm.tsx`). Stored on `Model.meta` JSON. Governs the model-page TOP stats (version totals) + model card stats.
3. **VERSION level** (sub-option) — section in the version edit form (`ModelVersionUpsertForm.tsx`). Stored on `ModelVersion.meta` JSON. Governs the per-version stats in the version details card only.

Precedence: a version-details stat is hidden if the version flag **OR** model flag **OR** user default says so; model totals/card are governed by the model flag **OR** user default.

### CP gating (read-time, lapse reverts with no DB write)
Effective visibility is computed **per-viewer at read time** = `storedFlag AND hasValidCreatorMembership(ownerId)`. A lapsed creator's stored flags silently revert to visible — **no database write**. Owners and moderators always see their real stats (bypass). Central helper: `src/server/utils/model-metric-privacy.ts` (dependency-light, unit-tested); batched membership lookup `getValidCreatorMembershipMap` in `creator-program.service.ts`.

### Surfaces gated (number omitted at the source, never leaked)
| Surface | File |
|---|---|
| Model card | `components/Cards/ModelCard.tsx` (downloads, buzz) |
| Model page top stats | `pages/models/[id]/[[...slug]].tsx` (downloads, generation, buzzEarned) |
| tRPC model controller rank | `server/controllers/model.controller.ts` (`getModelHandler` model + version rank; feed via `getModelsWithImagesAndModelVersions`; associated-resources cards) |
| v1 public API stats | `server/services/model.service.ts` (`getStatsForModel`/`getStatsForVersion`; flows into `buildPublicModelResponse`). Only download + tipped exist in v1 — generation count isn't in v1 stats, so we gate what exists |
| Search index doc | `server/search-index/models.search-index.ts` — stores effective model-level hide flags for the card notice; **keeps** real `metrics`/`rank` so sort order is unchanged |
| Category cards | `components/Model/Categories/ModelCategoryCard.tsx` |

### Ranking treatment (notice + upsell, not a hard opt-out)
On sort/leaderboard surfaces the model **stays listed** and the sort ORDER still uses the real underlying value server-side; only the exposed NUMBER is replaced with a subtle "creator has hidden this metric" notice (`HiddenMetricNotice.tsx`) + an invite for others to do the same. Rank position implies a rough range — accepted.

### Upsell
The settings card and both edit-form sections show a renew/join pitch and **disable the toggles** for non-active / lapsed members (via `getCreatorRequirements` → `validMembership` / `membershipLapsed`).

### Tests
`src/server/__tests__/model-metric-privacy.test.ts` (13 cases): owner/mod bypass; non-owner sees hidden when owner is an active member; non-owner sees real stats when owner has lapsed; per-level (user/model/version) precedence and unions.

### Notes / deferred
- The write path (`server/metrics/model.metrics.ts`) is **not** gated, by design — gating is read/display only.
- Search is a precomputed shared doc: a membership lapse reverts to visible on the next reindex rather than instantly (over-hides at worst until reindex; never leaks). The real metric remains in the search payload for sort ordering, as intended.
- `pnpm typecheck` passes clean (full, unfiltered); new unit tests pass.

---

## Review round 2 — adversarial-review fixes

**Blocking privacy leaks (real number no longer reaches the client for a hidden metric):**
1. **Meilisearch** (`models.search-index.ts`, `model.parser.ts`) — the displayed `metrics`/`rank` download+tipped are now **masked to null** when hidden; the REAL values live in a sort-only `sortMetrics.{downloadCount,tippedAmountCount}` that is in `sortableAttributes` but **excluded from a new `displayedAttributes` whitelist**. So no search hit (including a direct Meili query) carries the real hidden number, and sort order still uses the real value. Parser sorts on `sortMetrics.*` (enum indices preserved → sort-dropdown labels unchanged).
2. **OG image** (`api/og.tsx`) — resolves hidden metrics (owner settings + membership) and drops the Downloads / Generations tiles when hidden.
3. **v1 `/model-versions/[id]` + both `by-hash` routes** — shared `prepareModelVersionResponse` now gates `stats.downloadCount` via `resolveVersionHiddenMetrics` (fixes all three at once).
4. **v1 models list** (`getModelsWithVersions`) — `getStatsForVersion` now uses `resolveVersionHiddenMetrics` (version OR model OR user), fetching version metas in one batch. Fixes both the leak and the v1 version-only precedence bug.

**Blocking perf:**
5. `getValidCreatorMembershipMap` rewritten from K per-owner `dbWrite.customerSubscription.findMany` (Promise.all, primary pool) to a **single `dbRead` query** over all owners, collapsed to a userId→valid map mirroring `getHighestTierSubscription`'s tier logic.

**Should-fix:**
6. Reindex trigger — new `queueModelMetricPrivacyReindex(userId)` is fired from `setUserSettingHandler` when a `hideModel*` flag changes. The **membership-lapse** reindex is **deferred** (status transitions run through several Stripe/Paddle webhook paths; wiring each is out of scope for this pass) — the helper is in place and ready to call from those handlers.
7. Owner/mod over-hide on search cards — **softened the CreatorControlsCard copy** to note that on search results they see the hidden state, same as the public.

**Tests:** added v1 version-stats precedence cases (version-only hide honored; lapse reverts; model-level cascades). Full `pnpm typecheck` clean; 16 metric-privacy tests pass.

**Meili verification note:** enforcement is the index-level `displayedAttributes` setting (applied to ALL queries by Meilisearch, not client-side `attributesToRetrieve`), so a direct Meili query cannot return `sortMetrics`. The setting is applied by `onIndexSetup` on the next index setup/sync; a live cluster reindex to eyeball a hit is deferred to preview/deploy.

---

## Round 3 — donation-goal opt-out folded into Creator Controls (per Justin)

- **Moved** the user-level `hideDonationGoals` toggle out of `SettingsCard.tsx` and into `CreatorControlsCard.tsx`, next to the metric toggles, under the same **active-member gate + upsell** (disabled for non-active / lapsed members). The per-version donation-goal control in `ModelVersionUpsertForm.tsx` is unchanged.
- **CP-gated the effect** (`donation-goals-cache.ts`): goals are now hidden only when `hideDonationGoals AND hasValidCreatorMembership(ownerId)`. A lapsed / never-member owner's goals become **publicly visible again with no stored flip**, mirroring the metric-privacy gate. Owner/mod still bypass the shared cache entirely upstream in `modelVersionDonationGoals`.
- **Reuses the batched membership path** (single `dbRead`, no per-owner fan-out). Extracted `getValidCreatorMembershipMap` into a dependency-light `creator-membership.service.ts` so the light donation-goals lookup doesn't drag in the heavy creator-program graph (it's re-exported from `creator-program.service` for the existing metric-privacy callers). Only opted-out owners incur the membership check.
- **Tests:** opted-out owner hidden only while an active member; visible again when lapsed; owner/mod bypass documented (enforced upstream).

> ⚠️ **Behavior change to the shipped donation-goal opt-out (#3253):** anyone who set it while **not** an active member will have their goals become visible again. Intended per Justin.

Full `pnpm typecheck` clean; 24 tests pass (metric-privacy + donation-goals).

### DIFF
--- a/src/server/redis/donation-goals-cache.ts
+++ b/src/server/redis/donation-goals-cache.ts
@@ -1,6 +1,7 @@
 import { Prisma } from '@prisma/client';
 import { dbRead, dbWrite } from '~/server/db/client';
 import { dbReadFallbackCounter } from '~/server/prom/client';
+import { getValidCreatorMembershipMap } from '~/server/services/creator-membership.service';
 
 // A single public (non-owner, non-moderator) donation goal, shaped byte-identically to
 // `modelVersionDonationGoals`' output element: the DonationGoal row fields it selects plus
@@ -83,10 +84,12 @@ export const publicDonationGoalsLookupFn = async (
     },
   });
 
-  // Creator opt-out: an owner can hide the public donation-goal display for ALL their
-  // goals via the `hideDonationGoals` user setting. Non-owner/non-mod viewers go through
-  // this cache, so drop those goals here. Owners/mods bypass the cache entirely (see
-  // `modelVersionDonationGoals`) and still see their goals.
+  // Creator opt-out (Creator Program benefit): an owner can hide the public
+  // donation-goal display for ALL their goals via `hideDonationGoals`. It only takes
+  // effect while the owner holds a valid CP membership — a lapsed/never-member owner's
+  // goals become publicly visible again with no stored flip, mirroring the metric-
+  // privacy gate. Non-owner/non-mod viewers go through this cache; owners/mods bypass
+  // it entirely (see `modelVersionDonationGoals`) and always see their own goals.
   const ownerIds = [...new Set(goals.map((g) => g.userId))];
   const hiddenOwnerIds = new Set<number>();
   if (ownerIds.length > 0) {
@@ -94,9 +97,12 @@ export const publicDonationGoalsLookupFn = async (
       where: { id: { in: ownerIds } },
       select: { id: true, settings: true },
     });
-    for (const owner of owners) {
-      const settings = owner.settings as { hideDonationGoals?: boolean } | null;
-      if (settings?.hideDonationGoals) hiddenOwnerIds.add(owner.id);
+    const optedOutOwnerIds = owners
+      .filter((o) => (o.settings as { hideDonationGoals?: boolean } | null)?.hideDonationGoals)
+      .map((o) => o.id);
+    if (optedOutOwnerIds.length > 0) {
+      const membershipMap = await getValidCreatorMembershipMap(optedOutOwnerIds);
+      for (const id of optedOutOwnerIds) if (membershipMap.get(id)) hiddenOwnerIds.add(id);
     }
   }
 
--- a/src/components/Account/CreatorControlsCard.tsx
+++ b/src/components/Account/CreatorControlsCard.tsx
@@ -1,0 +1,97 @@
+import { Button, Card, Divider, Stack, Switch, Text, Title } from '@mantine/core';
+import { IconBolt } from '@tabler/icons-react';
+import { useCreatorProgramRequirements } from '~/components/Buzz/CreatorProgramV2/CreatorProgram.util';
+import { useCurrentUserSettings, useMutateUserSettings } from '~/components/UserSettings/hooks';
+import { useCurrentUser } from '~/hooks/useCurrentUser';
+import { useServerDomains } from '~/providers/AppProvider';
+import { syncAccount } from '~/utils/sync-account';
+
+/**
+ * Creator Controls: USER-default metric privacy (level 1 of 3). Sets the baseline
+ * hide flags for all of the creator's models. The flags only take effect while the
+ * user holds a valid Creator Program membership (enforced read-side); non/lapsed
+ * members see an upsell and disabled toggles.
+ */
+export function CreatorControlsCard() {
+  const user = useCurrentUser();
+  const serverDomains = useServerDomains();
+  const { requirements } = useCreatorProgramRequirements();
+  const { hideModelBuzz, hideModelDownloads, hideModelGenerations } = useCurrentUserSettings();
+  const { mutate: mutateSetting, isPending: isLoadingSetting } = useMutateUserSettings();
+
+  if (!user) return null;
+
+  const isActiveMember = !!requirements?.validMembership;
+  const membershipLapsed = !!requirements?.membershipLapsed;
+  const renewUrl = syncAccount(`//${serverDomains.green}/pricing`);
+
+  return (
+    <Card withBorder id="creator-controls">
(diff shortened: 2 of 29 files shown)


CivitaiDevOpsAgent:
### [PR-REVIEW] round 5 — 1 finding(s) (confidence ≥ 0.80)

- **blocker** `src/server/redis/donation-goals-cache.ts:4` (correctness, conf 0.95) — Import path `~/server/services/creator-membership.service` does not exist — the file is `creator-program.service.ts` and no `creator-membership` module exists anywhere in the repo.  
  ↳ _Change the import path to `~/server/services/creator-program.service` to match the import used in model.controller.ts and the actual file on disk._

src/server/redis/donation-goals-cache.ts
```diff
@@ -1,6 +1,7 @@
 import { Prisma } from '@prisma/client';
 import { dbRead, dbWrite } from '~/server/db/client';
 import { dbReadFallbackCounter } from '~/server/prom/client';
+import { getValidCreatorMembershipMap } from '~/server/services/creator-membership.service';
```
CivitaiDevOpsAgent, line 4: **high** (correctness, conf 0.95) — Import path is wrong: `creator-membership.service` does not exist; the file is `creator-program.service`. This will fail at build time.

↳ _Change the import to `~/server/services/creator-program.service`._

```suggestion
import { getValidCreatorMembershipMap } from '~/server/services/creator-program.service';
```

_pr-review · triage prior, human decides_

CivitaiDevOpsAgent, line 4: **blocker** (correctness, conf 0.95) — Import path `~/server/services/creator-membership.service` does not exist — the file is `creator-program.service.ts` and no `creator-membership` module exists anywhere in the repo.

↳ _Change the import path to `~/server/services/creator-program.service` to match the import used in model.controller.ts and the actual file on disk._

```suggestion
import { getValidCreatorMembershipMap } from '~/server/services/creator-program.service';
```

_pr-review · triage prior, human decides_

PR #3268, opened 2026-07-21
feat(app-blocks): per-engine budget ceilings for seamless-pano-360 customComfy recipe

ZacxDev:
### The problem
The App Blocks `customComfy` bridge uses a POST-PAID budget model: a recipe declares a single `maxBuzz` ceiling that MUST equal `ceil(stepTimeoutSeconds)` — the orchestrator step `timeout` physically caps spend at ~1 Buzz/GPU-second, so the timeout *is* the ceiling. At submit the router reserves the CEILING against every cap (per-user daily, per-app aggregate, dev-session) and settles-to-actual at terminal.

`seamless-pano-360` declared ONE ceiling (`stepTimeoutSeconds:180 / maxBuzz:180`), sized for the priciest engine (qwen, ~150 display estimate). A cheap Z-Image gen (dogfood-measured actual ≈ 21 Buzz) still reserved **180** against all three caps and then settled back — correct, but it needlessly consumes cap headroom that gates *other* generations (the per-user 50k/day cap, the per-app cap, the dev-session cap all hold the ceiling until settle).

### The change
Make the ceiling **per engine**.

Because the step `timeout` is the PHYSICAL cap, a lower `maxBuzz` **requires** a proportionally lower `stepTimeoutSeconds` — a lower reservation under the same 180s timeout would let the job run to 180 Buzz while only reserving less, under-counting real spend and defeating the cap. So the two always move together, from a single table:

```ts
export const BUDGET_BY_ENGINE = {
  'zimage-turbo': { stepTimeoutSeconds: 90,  maxBuzz: 90  },
  'flux2-klein':  { stepTimeoutSeconds: 150, maxBuzz: 150 },
  'qwen-image':   { stepTimeoutSeconds: 180, maxBuzz: 180 },
};
```

- `BlockRecipe` interface: the two scalars (`maxBuzz` / `stepTimeoutSeconds`) are replaced by `budgetFor(params)` and `budgetForEngine(engine)`, each returning `{ maxBuzz, stepTimeoutSeconds }`.
- The module-load invariant now asserts `maxBuzz === ceil(stepTimeoutSeconds)` for **each** of a recipe's engines (was a single scalar check).
- The router resolves `const { maxBuzz: ceiling, stepTimeoutSeconds } = recipe.budgetFor(params)` and threads the per-engine timeout into `createBlockCustomComfyStep(input, stepTimeoutSeconds)` (which dropped its now-unused `recipe` param). Everything downstream already keyed off `ceiling`.
- `estimateBuzz` (already per-engine, display-only) is unchanged.

### Concrete behavior change
Each engine now reserves and settles less cap the cheaper it is. The new observable behavior: at a budget of 100, a **zimage** gen (ceiling 90) now RUNS, while **qwen** (ceiling 180) is still rejected at the static gate (`recipe ceiling 180 exceeds budget 100`) — under the old single-180 ceiling that budget gated out *every* engine.

### 🔴 The one reviewer judgment call
The three timeout numbers. A `timeout` is a HARD kill on a slow generation, so headroom over typical runtime matters:
- **qwen-image 180s — UNCHANGED** → zero regression on the priciest engine.
- **zimage-turbo 90s** ≈ 4× the 21-Buzz dogfood-measured actual.
- **flux2-klein 150s** ≈ 3.3× its 45-Buzz DISPLAY ESTIMATE. flux2's runtime is characterized only by that estimate — there is **no** dogfood-measured actual like zimage's 21 — so 150 buys more margin against a hard-kill of a legit gen than 120 did, while still cutting over-reservation from the flat 180. (This is the audit's 🟡 follow-up; bumped 120 → 150.)

All tunable in the one obvious `BUDGET_BY_ENGINE` table. The MAX entry (180) still satisfies the app-manifest invariant `page.buzzBudgetPerGen (200) ≥ recipe ceiling`, so **no app-side change** is needed.

### Scope / safety
`customComfy` is DARK — behind the mod-gated App Blocks flag — so this rides the next release cut with no user-visible surface until the flag opens.

### Tests / verification
- `npx tsc --noEmit` (8 GiB heap): clean, 0 errors.
- Affected unit suites all green: recipe (30), workflow.service (98), custom-comfy-settle (17), blocks.router.workflow (220).
- Coverage: per-engine `budgetFor`/`budgetForEngine` resolution + default-engine fallback; per-engine module-load invariant; `createBlockCustomComfyStep` timeout formatting per engine (00:01:30 / 00:02:30 / 00:03:00); router submits proving per-engine reserve amounts + stamped step timeout differ (zimage 90/00:01:30 vs flux2 150/00:02:30 vs qwen 180/00:03:00); and the budget-100 zimage-runs / qwen-rejected behavior.

### Follow-up commit (audit findings addressed)
- **🟡 flux2 ceiling 120 → 150** — see the judgment-call note above.
- **🟢 per-engine reserve↔refund symmetry** — the earlier refund/rollback tests only proved the default engine's 90 refunds. Added coverage that a NON-default engine refunds ITS OWN ceiling: a qwen daily-cap breach refunds **180** (not 90), a flux2 app-cap breach refunds **150**, and a qwen refund-on-throw refunds **180** on BOTH caps.
- **🟢 graph-engine ↔ stamped-timeout end-to-end** — added an assertion that within ONE submit the engine driving the BUILT customComfy graph is the same engine driving the STAMPED timeout: a qwen submit stamps `00:03:00` AND its emitted step input is qwen's graph (`GGUFLoaderKJ` loader + qwen resource AIRs, not zimage's). Closes the "safe by construction but unasserted" gap.

### DIFF
--- a/src/server/routers/__tests__/blocks.router.workflow.test.ts
+++ b/src/server/routers/__tests__/blocks.router.workflow.test.ts
@@ -4516,7 +4516,8 @@ describe('customComfy bridge (submit/estimate/settle)', () => {
     return validClaims({
       ctx: { entityType: 'none', slotId: 'page' },
       appBlockId: 'apb_test',
-      buzzBudget: 500, // comfortably above the recipe ceiling (180) by default
+      // Comfortably above every per-engine ceiling (zimage 90 default … qwen 180 max).
+      buzzBudget: 500,
       ...over,
     });
   }
@@ -4601,19 +4602,22 @@ describe('customComfy bridge (submit/estimate/settle)', () => {
       happySubmit();
       const result = await caller().submitWorkflow({ blockToken: 'tok', body: ccBody() });
       expect(result.snapshot.workflowId).toBe('wf_cc_1');
-      // (2) per-user daily cap reserved the CEILING (180), not the 0 estimate.
+      // (2) per-user daily cap reserved the per-engine CEILING (zimage default =
+      // 90), not the 0 estimate.
       expect(mockSysRedis.incrBy).toHaveBeenCalledWith(
         expect.stringMatching(/^system:blocks:buzz-cap:42:/),
-        180
+        90
       );
       // (3) per-app aggregate cap reserved the CEILING.
-      expect(mockReserveAppSpend).toHaveBeenCalledWith('apb_test', 180);
+      expect(mockReserveAppSpend).toHaveBeenCalledWith('apb_test', 90);
+      // The submitted step's timeout matches the reserved ceiling (90s → 00:01:30).
+      expect(mockSubmitWorkflow.mock.calls[0][0].body.steps[0].timeout).toBe('00:01:30');
       // settle record persisted with BOTH reservation keys + the ceiling.
       expect(mockPersistCustomComfySettle).toHaveBeenCalledWith({
         workflowId: 'wf_cc_1',
         buzzCapKey: expect.stringMatching(/^system:blocks:buzz-cap:42:/),
         appSpendKey: 'system:blocks:app-spend-cap:apb_test:day',
-        ceiling: 180,
+        ceiling: 90,
       });
     });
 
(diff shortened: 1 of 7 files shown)


CivitaiDevOpsAgent:

PR #2600, opened 2026-06-16
fix(images): tolerate resize-cache purge failures (post.updateImage 500s)

ZacxDev:
## Root cause

`post.updateImage` throws **~63 `INTERNAL_SERVER_ERROR` (HTTP 500) per 30 min** on dp-prod. Every error in Loki has input `hideMeta:true` (sampled 30/30). That toggle is the **only** path through `updatePostImage` that calls `purgeResizeCache` (`src/server/services/post.service.ts:1339`, gated on `currentImage.hideMeta !== image.hideMeta`), which is the sole S3-touching code on the `updateImage` path.

**The deserializing call:** `purgeResizeCache` (`src/server/services/image.service.ts:255`) does:
```ts
const { items } = await getImageS3Client().listObjects({ bucket: env.S3_IMAGE_CACHE_BUCKET, prefix: url });
```
`getImageS3Client()` is an `@aws-sdk/client-s3` (smithy) client pointed at `S3_IMAGE_UPLOAD_ENDPOINT = https://s3-image-proxy.civitai.com`, bucket `S3_IMAGE_CACHE_BUCKET = civitai-media-cache` (verified from the live `civitai-cfg-primary` ConfigMap).

**Why the body starts with `'4'`:** that proxy does **not** implement the `ListObjectsV2` route. A path-style `GET /civitai-media-cache?list-type=2&prefix=...` returns a plain-text **`404 page not found`** (Go's default 404, `content-type: text/plain`, `x-content-type-options: nosniff`) — verified directly with curl. The AWS SDK v3 (`@smithy/core` schema deserializer) tries to parse that body as the expected structured/XML error response and throws:
```
char '4' is not expected.:1:1
  Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.
```
(`char '4'` = the first byte of `404 page not found`; `:1:1` = parse position; the exact phrasing is `@smithy/core`'s `submodules/schema`.) The throw propagates out of the mutation as a 500 — the tRPC stack reports it via `resolveMiddleware` because the resolver is wrapped.

## Which of the 3 cases

**Case 1 (non-essential call) — tolerate it.** `purgeResizeCache` is a best-effort cache *invalidation*, not a source-of-truth write: a stale resized variant is self-healing (re-derived on next request). Evidence it's meant to be best-effort:
- its sibling image-cacher leg is already explicitly best-effort ("Never block or throw … we accept up to 4d of stale L2 entries");
[...]

PR #3270, opened 2026-07-21
fix(images): remove dead S3 resize-cache purge (kills ~9/min 'char 4' log noise)

ZacxDev:
## What

`purgeResizeCache` did a direct S3 `listObjects` + `deleteManyObjects` against `env.S3_IMAGE_CACHE_BUCKET` through `getImageS3Client()`. This path is **dead in production** and only ever produced fail-soft warning noise (~9/min fleet-wide) plus a wasted round-trip on every image delete / `hideMeta` toggle.

## Why it never worked

- `getImageS3Client()` is configured from `S3_IMAGE_UPLOAD_ENDPOINT`, which is an **object-read proxy that does not implement `ListObjectsV2`**. A path-style list request comes back as a plain-text `404 page not found`, so the AWS SDK's XML **error** deserializer throws:
  ```
  char '4' is not expected.:1:1
  ```
  (version-independent — any `fast-xml-parser`; the body simply isn't XML).
- The cache bucket (`civitai-media-cache`) also lives on a **different storage provider** than the endpoint that client points at, so the list could never have resolved regardless.

Because `listObjects` is the first call, it **always threw before deleting anything** — the delete branch was unreachable. The throw has been swallowed + logged as a warning since #2600, so it's fail-soft (no 5xx), but it's pure log noise + a needless Cloudflare round-trip.

## The fix

Remove the dead S3 list/delete block. Real invalidation is unchanged: the retained **image-cacher `/admin/invalidate`** call (L2 Redis SCAN+DEL by prefix + Cloudflare tag purge) is the modern owner of the cache bucket and does the actual work.

## Safety

Behavior-neutral for prod: the removed block only ever threw *before* performing any delete, so deleting it cannot change current behavior. No caller contract changes — `purgeResizeCache` still never throws.

### DIFF
--- a/src/server/services/image.service.ts
+++ b/src/server/services/image.service.ts
@@ -286,44 +286,28 @@ const {
 // no user should have to see images on the site that haven't been scanned or are queued for removal
 
 export async function purgeResizeCache({ url }: { url: string }) {
-  // Best-effort: purge resized variants from the image-cache bucket. This is a
-  // cache invalidation, NOT a source-of-truth write — a stale resized variant is
-  // self-healing (re-derived on next request) and must never fail the caller's
-  // mutation. The S3 client here talks to an S3-compatible proxy
-  // (S3_IMAGE_UPLOAD_ENDPOINT); when that proxy returns a non-S3 body (e.g. a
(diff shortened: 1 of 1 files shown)


CivitaiDevOpsAgent:

PR #3271, opened 2026-07-21
feat(app-blocks): per-engine customComfy runtime/cost metric

ZacxDev:
## Goal

Instrument the App Blocks `customComfy` settle path with **per-engine runtime/cost observability** so the pre-GA question — *is `flux2-klein`'s real p99 approaching its 150-Buzz / 150s ceiling?* — has real data as usage accrues. This is **instrument-ahead-of-demand**: `customComfy` is DARK app code (mod-gated behind the App Blocks flag) with near-zero traffic today, so the metric will be **EMPTY until real volume accrues** — the point is to have it in place, not to read it now.

## The GPU-runtime vs wall-clock distinction (captured BOTH)

The pre-GA risk is that the orchestrator step `timeout` (`job.ExpireAt`) may count GPU **queue-wait**, so a fast gen could be hard-killed at 150s **wall-clock** even though its GPU **runtime** (the billed `actual` Buzz) is well under 150. Two distinct signals — labeled accurately, not conflated:

- **`civitai_app_block_customcomfy_actual_buzz`** `{engine, recipe}` — the settled **GPU-runtime cost in billed `actual` Buzz** (≈1 Buzz/GPU-second). Answers "how close to the ceiling is real spend." Cheap (already computed at settle). Buckets `[10,20,30,45,60,90,120,150,180,200]` straddle the per-engine ceilings (zimage 90 / flux2 150 / qwen 180).
- **`civitai_app_block_customcomfy_wallclock_seconds`** `{engine, recipe}` — **wall-clock incl. GPU queue-wait**, the truer signal for the clip risk. Captured cheaply by persisting a submit timestamp in the settle record and diffing at the terminal observation.

### Residual gap (wall-clock)

The wall-clock is measured **server-side from just-before-submit to the block's terminal-POLL observation**, not the orchestrator's exact terminal transition. So it:
- is bounded above by the block's ~2s terminal-poll cadence (cancel path observes immediately),
- excludes the `submitWorkflow` round-trip (timestamp captured just before it), and
- is subject to sub-second cross-pod clock skew.

For the "does the 150s timeout clip a queued-but-fast gen" question this is close enough (±~2s on a 150s budget) and it captures the **purest** clip case: wall-clock is emitted independently of `actual`, so a gen killed at its timeout with ~0 accrued Buzz still shows ~150s wall-clock. The exact GPU queue-vs-run split lives only in the orchestrator trace and is **not** captured here.

## Implementation

- Thread the resolved `engine` (`params.engine ?? recipe.engines[0]`, the declared default — cardinality-safe enum) + `recipe.id` + a submit `submittedAt` through the `SettleRecord` (written at submit in `submitCustomComfyWorkflow`, read at the first terminal settle).
- Emit both histograms in `settleCustomComfySpend`, **before** the `refund <= 0` early-return, so a full-ceiling-spend gen (the ceiling-pressing case we most want) is still observed.
- **Fail-soft**: the emit helpers swallow all errors — a metrics failure can never perturb settle/refund correctness. The new record fields are optional/absent-safe (a legacy pre-deploy record emits nothing).

## Cardinality

`engine ∈ {zimage-turbo, flux2-klein, qwen-image}` (3) × `recipe ∈ {seamless-pano-360}` (1) → 4 series per histogram today. Both labels are strict enums resolved server-side from the code-owned recipe registry (never client input).

## Behavior change

**None.** DARK app code; the metric is pure observability and the emit is orthogonal to (and fail-soft around) the spend/refund path.

## Tests

Extended `custom-comfy-settle.service.test.ts` (record now carries `engine`/`recipe`/`submittedAt`; settle emits the histograms with the right engine label + observed value, incl. the emit-before-refund-early-return and back-compat legacy-record cases) and updated `blocks.router.workflow.test.ts` persist assertions. **243 passed** (settle 23, router workflow 220); `tsc --noEmit` exit 0. Not verified against live traffic (none yet — see instrument-ahead-of-demand above).

### DIFF
--- a/src/server/metrics/app-block-runtime.metrics.ts
+++ b/src/server/metrics/app-block-runtime.metrics.ts
@@ -147,7 +147,20 @@ type Bundle = {
   requestsTotal: Counter<string>;
   requestDurationSeconds: Histogram<string>;
   rendersTotal: Counter<string>;
+  customComfyActualBuzz: Histogram<string>;
+  customComfyWallclockSeconds: Histogram<string>;
 };
+
+// ── customComfy per-engine runtime/cost buckets ──────────────────────────────
+// Sized for the 0–200 range that straddles the per-engine Buzz ceilings
+// (zimage 90 / flux2 150 / qwen 180) so the ceiling boundaries fall ON bucket
+// edges — a pre-GA `histogram_quantile` p95/p99 then reads directly against the
+// ceiling a gen is fighting.
+const CUSTOMCOMFY_BUZZ_BUCKETS = [10, 20, 30, 45, 60, 90, 120, 150, 180, 200];
+// Wall-clock (submit→terminal-observation) seconds — same ceiling landmarks plus
+// a couple of longer tails to catch a gen that queue-waits toward / past its
+// timeout (the clip-risk the metric exists to surface).
+const CUSTOMCOMFY_WALLCLOCK_BUCKETS = [5, 10, 20, 30, 45, 60, 90, 120, 150, 180, 200, 240];
 
 /**
  * Idempotent: safe to call on every request. Returns the metric instances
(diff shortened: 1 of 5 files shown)


CivitaiDevOpsAgent:

PR #3287, opened 2026-07-21
fix(block): enforce user blocking on comment/reaction/review write paths

JustMaier:
## Problem

User blocking was enforced **read-side only** — an entity page 404s a blocked viewer — but the **write/interaction paths had no block check**. A user a creator has blocked could still comment, reply, react/vote, and review that creator's content via direct tRPC/API calls.

## Fix

New `src/server/services/block-check.service.ts` resolves the content **owner** per entity type and throws `NotFound` (mirroring the read-handler pattern, e.g. `model.controller.ts:187`) when the acting user is blocked by that owner. Enforced on **CREATE only** (edit/delete of one's own content is unaffected); moderators are exempt.

### Handlers gated
- **`upsertComment`** (`commentsv2.service.ts`) — enforced on the create branch (`!data.id`), so both `upsertCommentV2Handler` and API-token (SocialWrite) callers hitting the service directly are covered. `isModerator` threaded from `commentv2.controller.ts`.
- **`upsertCommentHandler`** (`comment.controller.ts`, legacy model comments) — checks the model owner (`ctx.ownerId`) plus the parent comment author for replies, on create (`!input.commentId`).
- **`toggleReactionHandler`** (`reaction.controller.ts`) — checks the reacted entity's owner before toggling.
- **`upsertResourceReview`** (`resourceReview.service.ts`) — checks the model owner on create; `isModerator` threaded from `resourceReview.controller.ts`.

### Owner resolution (`getBlockCheckOwnerIds`)
Handles `image`/`post`/`article`/`model`/`review`+`resourceReview`/`question`/`answer`/`bounty`/`bountyEntry`/`commentOld`. For CommentV2 **reply targets** (`entityType: 'comment'`) it resolves **both** the parent comment author **and** the root thread's content owner, so a blocked user can't reply inside a creator's thread. Unknown/unowned types (e.g. `challenge`, `appListing`) resolve to no owner and never false-block.

## Tests
- `block-check.service.test.ts` — owner resolution per entity type (incl. reply → parent author + root owner), enforcement (blocked throws, non-blocked passes, owner-acting-on-own-content and moderators exempt).
- `commentsv2.blockCheck.service.test.ts` — `upsertComment` throws and never creates when blocked, allows non-blocked, exempts moderators, and skips the check on edit.
- Updated `commentsv2.appListing` and `resourceReview.idempotent` test mocks for the new block-check lookups.

Full `pnpm typecheck` passes clean; new/affected unit tests pass.

> Note: sibling task 868keupaa ("downvote-then-block") edits the read/exclusion paths in some of the same files; this diff is scoped to the write/create handlers to minimize overlap.

### DIFF
--- a/src/server/controllers/comment.controller.ts
+++ b/src/server/controllers/comment.controller.ts
@@ -23,6 +23,7 @@ import {
   updateCommentReportStatusByReason,
 } from '~/server/services/comment.service';
 import { createNotification } from '~/server/services/notification.service';
+import { throwIfBlockedByOwners } from '~/server/services/block-check.service';
 import { amIBlockedByUser } from '~/server/services/user.service';
 import {
   throwAuthorizationError,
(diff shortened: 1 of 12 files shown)


CivitaiDevOpsAgent:

PR #3288, opened 2026-07-21
fix(block): let creators see/report a blocker's engagement on their own content

JustMaier:
## The gap

When user B blocks creator A, B's prior comments / reviews / downvotes disappeared from A's view of A's **own** content, and A could no longer report them. A's content-filter exclusion list merges `blockedByUsers` (everyone who blocked A) into the `notIn` / `NOT IN` filter via `boundExcludedUserIds`. That `blockedByUsers` exclusion is deliberate anti-harassment (a harasser who blocked you can't resurface to you) — but on a creator's **own** content it backfired: a downvoter/commenter could hide their trail simply by blocking the creator.

## The fix

Added an `isContentOwner` option to `boundExcludedUserIds` (`src/server/utils/excluded-user-ids.ts`) that drops `blockedByUsers` **only** when the viewer is the owner of the content being engaged with. The viewer's own lists (`hiddenUsers` + `blockedUsers`) are always kept, and the P2029 bind-param cap/ordering safety is untouched.

Anti-harassment stays intact for **every non-owner viewer** — `blockedByUsers` is only dropped in the owner-viewing-own-content case.

## Surfaces fixed (read/list/exclusion paths only)

- **CommentV2** thread details + `getInfinite` (`commentv2.controller.ts`) — new `getThreadEntityOwnerId` / `isViewerContentOwner` resolver in `commentsv2.service.ts` maps each thread's entity (image, post, model, article, bounty, review, comment, question, answer, challenge, comicChapter, …) to its owner.
- **Legacy model comments** — `comment.service.getComments` (owner = model owner).
- **Model reviews/thumbs** — `resourceReview.service.getPagedResourceReviews` (owner = model owner via modelVersion).

The owner lookup is **skipped entirely** unless the viewer actually has a non-empty blocked-by list, so the extra query stays off the hot path for anonymous and unblocked viewers. Entity types we can't cheaply resolve fall back to "not owner" (keeps `blockedByUsers` — the safe default).

Scope is kept to read/exclusion paths only, to avoid conflicting with the sibling "enforce blocking on writes" PR (task 868keup9w). `bounty.controller` / `user.controller` were left as-is (not owner-viewing-own surfaces).

## Tests

- `boundExcludedUserIds`: owner sees a blocker's comment/review/downvote (blocker dropped); non-owner viewer still excludes the blocker; owner still excludes their own hidden/blocked lists.
- `isViewerContentOwner` / `getThreadEntityOwnerId`: owner detection across entity types (incl. challenge `createdById`, comicChapter via project), and the skip-guard (no lookup when the viewer has no blocked-by list or is anonymous).

`pnpm typecheck` is clean for all changed files (remaining errors in the run are pre-existing, in untouched `packages/civitai-db` / `challenge.service.ts`).

### DIFF
--- a/src/server/controllers/commentv2.controller.ts
+++ b/src/server/controllers/commentv2.controller.ts
@@ -28,6 +28,7 @@ import {
   getCommentCount,
   getCommentsThreadDetails2,
   getCommentsInfinite,
+  isViewerContentOwner,
   toggleHideComment,
   toggleLockCommentsThread,
   upsertComment,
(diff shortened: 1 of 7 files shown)


CivitaiDevOpsAgent:

PR #3290, opened 2026-07-21
feat(early-access): enforce permanent pay-for-access + lock the write endpoint

bkdiehl:
Read-path enforcement — every paywall that treated a permanent version's NULL earlyAccessEndsAt as free/public is now permanent-aware (permanent = availability 'EarlyAccess' + NULL end date, gated forever):
- purchase guard (model-version.service): permanent is purchasable; the "already public" path is skipped for a NULL end date.
- download paywall (file.service): inEarlyAccess true for permanent.
- generation: mini/[id] checkPermission SQL + resource-data.redis config exposure gate permanent.
- model.controller download button, model-version.utils client flag, model.service isEarlyAccess rollup.

Write path:
- early-access endpoint: setting `permanent` requires the shared WEBHOOK_TOKEN (mods excepted), so only the Creator Studio (which enforces the CP-membership + tier cap) can set it; a direct user call gets 403.
- assertEarlyAccessChargeConfig requires a charge for permanent.
- mergeEarlyAccessConfigUpdate: post-publish edits are unrestricted except donation goals — a purchase is a durable entityAccess entitlement not re-evaluated against current config, so changing terms never affects existing buyers (CU 868ke4944).

Depends on the already-merged foundation (earlyAccessPermanent column + trigger). Known gap: the tRPC modelVersion.upsert path can still set `permanent` without the token (to be addressed separately).


CivitaiDevOpsAgent:

PR #3286, opened 2026-07-21
feat(comments): let content owners (not just mods) pin comments on CommentsV2 surfaces

JustMaier:
## What

Widens CommentsV2 comment pinning from **moderator-only** to **owner-or-moderator**, so a content owner can pin/unpin comments on their own content. Applies to the CommentsV2 surfaces where `resourceOwnerId` is plumbed today: **Article, Comics, Model3D, AppListing** (and any future V2 surface that passes it).

> **Scope note:** this does **NOT** affect the classic model page (`/models/[id]`). Model-page comments run the separate **legacy** comment system (`comment.*` / `Comment` table), which has no `pinnedAt`. Creator pinning on the model page is delivered separately by **#3291**. This PR is a related but independent improvement to the V2 surfaces. Original context: ClickUp 868ketvum (SubtleShader + MNeMiC).

## How

A pin mechanism already existed in CommentsV2 but was moderator-only. This widens it to **owner-or-moderator**, reusing the exact authorization pattern that comment *hide* already uses — no new pin system.

**Server**
- `togglePinned` moves from `moderatorProcedure` → `protectedProcedure` (`SocialWrite` scope), input widened to `toggleHideCommentSchema` so the handler knows the entity.
- New `togglePinnedCommentHandler` authorizes owner-or-mod and rejects everyone else — the client gate is not trusted.
- The entity-owner resolution previously inline in `toggleHideCommentHandler` is extracted into a shared `getCommentV2WithEntityOwner` helper and reused by both hide and pin (covers model `userId`, challenge `createdById`, comic-chapter `project.userId`).

**Client**
- `CommentProvider` exposes `canPin = owner-or-mod`, mirroring the existing `canHide`.
- The pin menu item in `Comment.tsx` gates on `canPin` instead of `isModerator`; removed the now-unused `useCurrentUser`.
- `useMutateComment` passes `entityType`/`entityId` through to the mutation to match the widened input.

## Verification

- `pnpm run typecheck` passes; `prettier` clean on all touched files.
- Owner-or-mod check enforced server-side in the handler; reuses the same resolution comment-hide relies on, so owner-pin lights up wherever `resourceOwnerId` is passed to `CommentProvider`.
- Adversarially reviewed: server auth is injection-safe and fails closed for non-owners.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3291, opened 2026-07-21
feat(comments): let model owners pin comments on the model page

JustMaier:
## What

Lets a **model owner (creator) or moderator** pin/unpin comments on the model page (`/models/[id]`) so important comments surface at the top.

## Why this is a separate PR from #3286

The model page renders the **LEGACY** comment system (`trpc.comment.*` → `Comment` table → `ModelDiscussionV2` / `CommentDiscussionItem`), **not** CommentsV2. #3286 wired pinning into CommentsV2 (`commentv2.*`), which does not drive the model page. This PR adds pinning to the legacy path so it actually works there. No CommentsV2 files are touched.

## Changes

**DB migration** — `prisma/migrations/20260721160807_add_comment_pinned_at/migration.sql`:
```sql
ALTER TABLE "Comment" ADD COLUMN "pinnedAt" TIMESTAMP(3);
```
Added `pinnedAt DateTime?` to the legacy `Comment` model in `packages/civitai-db-schema/prisma/schema.full.prisma` (authoritative schema) and regenerated the Prisma client + kysely/interface types.

⚠️ **Migration must be applied MANUALLY** to preview/prod (this repo does not use `prisma migrate deploy`).

**Server**
- `comment.service.ts`: new `togglePinComment` mirroring `toggleHideComment`, but auth is **model-owner-OR-moderator only** — the non-mod clause is just `m."userId" = <userId>` (no comment-author clause, so a random commenter cannot pin their own comment). Enforced server-side via the raw `Comment`→`Model` join, same as hide.
- `comment.service.ts`: pinned-first ordering. New `getPinnedComments` (where `pinnedAt IS NOT NULL`, `ORDER BY pinnedAt DESC`, same modelId/filters/excluded-users). `getComments` now always excludes pinned (`pinnedAt: null`), matching how CommentsV2 excludes pinned from its regular query.
- `comment.controller.ts`: `getCommentsInfiniteHandler` computes the regular-comment cursor **first** (unchanged +1 sentinel logic), then on the **first page only** (`!input.cursor`) fetches pinned and **prepends** them. Cursor pagination is unaffected: pinned are never counted toward the limit/sentinel, and subsequent pages don't refetch them. Added `togglePinCommentHandler`.
- `comment.router.ts`: `togglePin` procedure (`protectedProcedure`, `SocialWrite` scope, `getByIdSchema`), mirroring `toggleHide`.
- `comment.selector.ts`: added `pinnedAt` to `commentDetailSelect` so it flows to both `getAll` (indicator) and `getById` (menu optimistic update).

**Client**
- `CommentDiscussionMenu.tsx`: `togglePin` mutation (optimistic on `getById` + invalidate `comment.getAll`) and a Pin/Unpin `Menu.Item` gated `(isModelOwner || isMod)`, using `IconPinned` / `IconPinnedOff`.
- `CommentDiscussionItem.tsx`: orange `IconPinned` ThemeIcon indicator next to the menu when `pinnedAt` is set (mirrors the CommentsV2 indicator).

## Pinned-first ordering (cursor caveat)

Pinned comments appear only at the top of the **first page**; they are excluded from the regular query on every page, so they never duplicate and never consume the cursor's `limit+1` sentinel. `nextCursor` is derived solely from regular comments, so pagination behavior is identical to before.

## Verification

- `pnpm run typecheck` — passes.
- Regenerated types confirmed: `Comment.pinnedAt` present in `packages/civitai-db-schema/src/models.ts` and `.../src/kysely/types.ts`.
- Prettier clean on all changed source files.
- Not verified at runtime (no DB with the column applied); migration is intentionally unapplied.

### DIFF
--- a/packages/civitai-db-schema/prisma/schema.full.prisma
+++ b/packages/civitai-db-schema/prisma/schema.full.prisma
@@ -3445,6 +3445,7 @@ model Comment {
   modelId      Int
   locked       Boolean? @default(false)
   hidden       Boolean? @default(false)
+  pinnedAt     DateTime?
 
   comments  Comment[]         @relation("ParentComments")
   reactions CommentReaction[]
--- a/packages/civitai-db-schema/src/kysely/types.ts
+++ b/packages/civitai-db-schema/src/kysely/types.ts
@@ -1845,6 +1845,7 @@ export type Comment = {
   modelId: number;
   locked: Generated<boolean | null>;
   hidden: Generated<boolean | null>;
+  pinnedAt: Timestamp | null;
 };
 export type CommentReaction = {
   id: Generated<number>;
--- a/packages/civitai-db-schema/src/models.ts
+++ b/packages/civitai-db-schema/src/models.ts
@@ -2277,6 +2277,7 @@ export interface Comment {
   modelId: number;
   locked: boolean | null;
   hidden: boolean | null;
+  pinnedAt: Date | null;
   comments?: Comment[];
   reactions?: CommentReaction[];
   reports?: CommentReport[];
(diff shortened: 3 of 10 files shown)


CivitaiDevOpsAgent:

PR #3295, opened 2026-07-22
fix(challenges): entry-fee disclosure, cooldown/owner gating, buzz colors + mod rescan

manuelurenah:
Tester feedback on community (user-created) challenges, plus two money-path bugs found while verifying the reports.

## 1. Entry-fee disclosure — `ChallengeSubmitModal`

Participants were never told an entry fee is charged, nor how it splits. There's now a breakdown under the submit button: total, the AI-judging house cut, the prize-pool contribution, and the guaranteed-review fee when opted in. A tooltip carries the refund policy. Numbers come from `challenge.constants`, the same source the creator form uses, so they can't drift from the server.

**Balance check was effectively missing.** `BuzzTransactionButton` only rendered when "guarantee review" was checked, and its `buzzAmount` only counted `reviewCost` — so entry-fee challenges with the box unchecked (the common case) had *no* balance check at all. It now always covers the real total, scoped via `accountTypes` to the challenge's own Buzz account, because the server charges the fee with `fromAccountType: challenge.buzzType` — an unscoped check green-lights a user holding plenty of the *other* Buzz type.

## 2. Winner-cooldown banner on user challenges

`getWinnerCooldownStatus` never looked at `source`, so participants saw *"you're on a winner cooldown, your entries won't be eligible for prizes"* on user challenges. `pickWinners` explicitly skips the cooldown for `source=User` (`daily-challenge-processing.ts:1743`), so the restriction was never enforced — the warning was purely false. Fixed at the service; `CooldownBanner` also skips the query client-side.

## 3. Challenge owner could open the submit modal

The server already rejects owner entries (`collection.service.ts:2006`), but the UI let owners select images and only fail at submit. Submit is now disabled with a tooltip; **Generate stays active** (product call). Uses Mantine's `data-disabled` rather than `disabled` so the tooltip still receives hover, with the click handler swapped so keyboard activation no-ops too. The new `useIsChallengeOwner` hook mirrors the server rule including its moderator exemption — without that, a mod who created a challenge saw a blocked button for a submission the server allows.

## 4. New moderator "Rescan Content" action

`challenge.rescan` (moderator-only) force-resubmits the moderated text and re-queues the cover image. `forceRescan: true` bypasses the `contentHash` dedup — the same dedup that stranded challenges at Pending in #3160. It deliberately does **not** reset `ingestion`/`scannedAt`, so a live challenge stays visible while the scan runs; the adapter writes the terminal state when the webhook lands.

It throws when the submit produced no workflow: `createXGuardModerationRequest` resolves `undefined` on failure rather than throwing, so without that check moderators got a "rescan queued" confirmation for a scan that never happened. The acting moderator is logged to Axiom.

## Two defects this surfaced

**`applyFailure` could hide a live challenge.** `challengeModerationAdapter.applyFailure` wrote `ingestion: Error` unconditionally on any `failed`/`expired`/`canceled` callback. Before the rescan action that could only ever land on an already-`Pending` challenge — the new action is the first path submitting a workflow for a `Scanned` one, so a single transient orchestrator failure would drop a live challenge out of feeds and 404 its detail page. Now scoped to `Pending`, so a failed rescan preserves the prior verdict.

**Entrants weren't refunded when a live challenge was blocked.** `applyChallengeNsfwEscalation`'s non-Scheduled branch hid the challenge and closed its collection but never refunded or notified anyone — entrants' pool contributions sat in account 0 on a dead challenge. An `Active` challenge now goes through `voidChallenge`, which refunds the entry-fee **pool** portion (the house cut is a separate prefix and stays kept) plus the creator's escrowed prize, and notifies entrants. Its `Active -> Cancelled` claim is load-bearing: refunding without it would let the completion cron pay winners out of the pool being returned. `Completing`/`Completed` still hold for human review — that pool is already in payout, so refunding would double-spend.

## 5. Wrong Buzz color on prize pools and prize badges

`getCurrencyConfig` defaults a missing `type` to **yellow** (`currency.constants.ts:122`), so every challenge prize/pool badge rendered yellow regardless of the challenge's actual currency. `buzzType` wasn't in the card payload at all, so cards physically couldn't render it — added to `ChallengeListItem` and all four card projections, then threaded through `ChallengeCard`, `CompletedChallengeCard`, `WinnerPodiumCard`, the detail page and the upsert form.

Raw SQL isn't type-checked, so adding the field to the row types alone would have compiled clean and silently defaulted every card to yellow at runtime — each `SELECT` was verified by hand.

Two badges intentionally left alone, both verified against the payout paths:
- The **participation prize** is paid `toAccountType: 'blue'` (`daily-challenge-processing.ts:1071`), so its hardcoded `type="blue"` was already correct.
- The **guaranteed-review** badge stays yellow: `requestReview` charges without `fromAccountType`, so it draws the user's default account. Typing it green would misrepresent a money path. On a green challenge this shows as entry fee green / review fee yellow — the fix is a one-line server change (`fromAccountType: challenge.buzzType` in `requestReview`), deliberately left out of this PR.

## Testing

- New `challenge-rescan.service.test.ts` — pins `forceRescan`, the no-ingestion-reset invariant, cover re-queue + Pending skip, throw-on-no-workflow, and the `source=User` cooldown short-circuit.
- `challenge-nsfw-escalation.test.ts` updated for the void+refund behavior, plus a new case covering the `Completing` hold path.
- Full `src/server/services/__tests__/` + `src/server/games/` suites: **1661 pass, 0 fail**.

No schema changes — no migration needed.

### DIFF
--- a/src/components/Cards/ChallengeCard.tsx
+++ b/src/components/Cards/ChallengeCard.tsx
@@ -91,6 +91,7 @@ export const ChallengeCard = memo(function ChallengeCard({ data }: Props) {
     createdById,
     nsfwLevel,
     prizePool,
+    buzzType,
     entryCount,
     commentCount,
     createdBy,
@@ -181,6 +182,7 @@ export const ChallengeCard = memo(function ChallengeCard({ data }: Props) {
           <div className="flex items-center justify-between gap-2">
             <CurrencyBadge
               currency={Currency.BUZZ}
+              type={buzzType}
               unitAmount={prizePool}
               radius="xl"
               px={8}
--- a/src/components/Cards/CompletedChallengeCard.tsx
+++ b/src/components/Cards/CompletedChallengeCard.tsx
@@ -14,7 +14,7 @@ import type {
 import { isDefined } from '~/utils/type-guards';
 
 export function CompletedChallengeCard({ data }: Props) {
-  const { id, title, theme, endsAt, prizePool, entryCount, winners } = data;
+  const { id, title, theme, endsAt, prizePool, buzzType, entryCount, winners } = data;
   const challengeUrl = `/challenges/${id}/${slugit(title)}`;
 
   // Podium order: [2nd, 1st, 3rd] — same as detail page
@@ -59,6 +59,7 @@ export function CompletedChallengeCard({ data }: Props) {
             <Group gap="sm" mt={4}>
               <CurrencyBadge
                 currency={Currency.BUZZ}
+                type={buzzType}
                 unitAmount={prizePool}
                 radius="xl"
                 variant="light"
@@ -91,6 +92,7 @@ export function CompletedChallengeCard({ data }: Props) {
                 isFirst={winner.place === 1}
                 className={winner.place === 1 ? 'z-10' : ''}
                 compact
+                buzzType={buzzType}
               />
             ))}
           </div>
--- a/src/components/Cards/MyChallengeCard.stories.tsx
+++ b/src/components/Cards/MyChallengeCard.stories.tsx
@@ -25,6 +25,7 @@ const baseChallenge: MyParticipatedChallengeItem = {
   endsAt: daysFromNow(-1),
   status: ChallengeStatus.Completed,
   source: ChallengeSource.System,
+  buzzType: 'yellow',
   nsfwLevel: 1,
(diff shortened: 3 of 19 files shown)


CivitaiDevOpsAgent:

PR #3296, opened 2026-07-22
perf(redis): opt-in per-pod L1 in front of near-static feed-hydration caches

ZacxDev:
## What

Adds an **opt-in per-process (per-pod) L1 cache** in front of the Redis per-id
cache used by the image-feed hydration path.

`redis.packed.mGet` decomposes into `Promise.all(keys.map(get))` (to avoid
`CROSSSLOT` on the cluster), so hydrating a feed page fans out to one Redis
`GET` **per image per cache**, each going through the cluster-routing-retry
layer individually. Pipelining hides round-trip latency but not command count.
An L1 hit skips the entire per-id `GET` fan-out for a hot id.

## How

New opt-in options on `createCachedArray` / `createCachedObject`
(`src/server/utils/cache-helpers.ts`):

- `localTtl` (seconds) — enables a bounded in-process LRU (built on the existing
  `createLruCache` util). Unset = no L1 layer at all (zero behaviour change).
- `localMax` — LRU **entry** cap (default 10000, belt-and-suspenders).
- `localMaxBytes` — LRU **byte** budget → deterministic heap cap. The LRU evicts
  to keep the retained-byte estimate ≤ budget; a single value larger than the
  budget is simply not stored (falls through to Redis), never an error.

`createLruCache` was extended to accept `maxSize` + `sizeCalculation`
(lru-cache@11 supports `max` and `maxSize` together). Default estimator
`roughSizeOf(v) ≈ JSON.stringify(v).length * 2` (UTF-16) + fixed overhead; it
runs on `set` (miss-fill) only, so it stays cheap.

Design guarantees:

- **Byte-identical output** — the L1 stores the *final* per-id value
  (post-`appendFn`, `cachedAt` stripped), so an L1 hit returns exactly what the
  Redis path returns.
- **Per-id, not per-request** — hits accrue across requests; only L1-miss ids
  flow into the existing Redis path, which then backfills L1.
- **Positive-only** — a `notFound` id is never L1-cached (no stale-miss pinning).
- **Mirrors the Redis-write skip** — the backfill withholds ids the Redis layer
  itself did not persist (`dontCache` debounce window / `dontCacheFn`), so a
  freshly-busted id read from `dbRead` (possibly replica-lagged) is not pinned.
- **Self-pod invalidation** — `bust`/`invalidate`/`refresh` delete the id from
  this pod's L1 (`flush` clears it), so the pod that processed a mutation never
  serves its own pre-mutation copy. (Cross-pod staleness is inherent + accepted.)
- **Shared-instance isolation** — shallow clones in and out; top-level mutation
  can't corrupt the cached instance. Consumers treat returned values as
  read-only for nested fields (all current consumers map into new objects).
- **Fail-open preserved** — a Redis read reject still degrades to the origin,
  merges any already-served L1 hits, and backfills.
- **Observability** — `cacheHitCounter` / `cacheMissCounter` with
  `cache_type="lruCache"` → L1 hit-rate measurable per cache.

## Where it's enabled

Only on always-on, **near-static** per-entity caches whose Redis TTL already
tolerates far more staleness than the short L1 adds, and that attach *display
metadata* to already-eligible feed images (visibility/moderation is enforced by
the feed query, not these caches):

| cache | fields | localTtl | localMaxBytes | Redis TTL | invalidation |
|---|---|---|---|---|---|
| `userBasicCache` | `{id, username, image, deletedAt}` | 30s | 4 MB | 1 day | refresh-only |
| `imageResourcesCache` | model attribution | 30s | 12 MB | 8h | refresh-only |
| `tagIdsForImagesCache` | tag ids | 15s | 6 MB | 8h | bust/refresh |
| `imageTagsCache` | votable-tag composites | 15s | 16 MB | 1h | bust |

### Deliberately NOT enabled

- **`imageMetaCache` — excluded for privacy.** Its value (generation prompt/
  params) is gated on `hideMeta`; flipping `hideMeta` false→true calls
  `imageMetaCache.refresh()` (meta→NULL), which a per-pod L1 can't observe → an
  exposure window on a privacy control. Also the heaviest value.
- **image-metrics cache** — volatile like/comment/collect counts.

## Trade-off (called out)

The L1 is per-pod and cannot see a cross-pod `bust`/`refresh`, so a mutation
propagates with up to `localTtl` extra staleness on pods **other than** the one
that processed it (self-pod invalidation handles the processing pod). Acceptable
because these 4 caches are near-static, their Redis TTLs already tolerate much
larger windows, and none is a visibility/auth gate.

## Deploy plan

The L1 is now **byte-bounded to a hard ≈38 MB/pod total** (4 + 12 + 6 + 16 MB,
`localMaxBytes` per cache) — well under the heap-headroom alert, and a per-value
size spike can't blow the cap (deterministic). Safe to **deploy fleet-wide**
(the L1 is in-code, not flag-gated, and the cache instances are shared across
pools, so a one-pod bake isn't possible anyway). Still watch per-pod heap RSS +
the headroom alert through the first traffic peak, and confirm the
`cache_type="lruCache"` hit-rate + the drop in per-id Redis GET volume.

## Testing

- `cache-helpers-l1.test.ts` (14 tests): L1 hit skips Redis+origin; partial hit
  routes only misses to Redis; byte-identical to the Redis path; TTL expiry
  re-fetches; entry-count eviction; **byte-cap eviction keeps the footprint
  bounded**; **a single value over budget is not stored + falls through (no
  throw)**; `notFound` never pinned; a cache without `localTtl` is untouched;
  mutation isolation; `dontCache`/debounce id not L1-pinned; `bust`/`refresh`
  self-pod L1 drop; `createCachedObject` keyed-Record parity.
- Existing `cache-helpers.test.ts` + `cache-helpers-failopen.test.ts` green.
- `tsc --noEmit` clean.

### DIFF
--- a/src/server/redis/caches.ts
+++ b/src/server/redis/caches.ts
@@ -1300,12 +1300,11 @@ export const imageMetaCache = createCachedObject<ImageWithMeta>({
   // working set into misses and stampede dbRead.
   ttl: CacheTTL.hour * 4,
   staleWhileRevalidateTtl: CacheTTL.hour,
-  // L1: generation meta behind a 4h Redis TTL, refresh-only on image edit/delete.
-  // Shortest L1 TTL of the set (10s) because `hideMeta` toggles the prompt to NULL —
-  // a ≤10s cross-pod lag on hide taking effect, on data that was already public.
-  // max 5000: `meta` is the heaviest per-entity value (~2.8KB/key), keep it small.
-  localTtl: 10,
-  localMax: 5000,
+  // NO per-pod L1 here (deliberate): `meta` is gated on `hideMeta`, and when an owner
+  // flips hideMeta false→true the write path calls imageMetaCache.refresh() (meta→NULL).
+  // A per-pod L1 can't observe that cross-pod refresh, so other viewers' pods would keep
+  // serving the old prompt for up to localTtl — a prompt-exposure window on a PRIVACY
+  // control. Not worth it; also the heaviest value, so its exclusion frees the most heap.
 });
 
 type ImageWithMetadata = {
--- a/src/server/redis/caches.ts
+++ b/src/server/redis/caches.ts
@@ -64,6 +64,7 @@ export const tagIdsForImagesCache = createCachedObject<{
   // edit by ≤15s. Not a visibility gate (feed SQL enforces that). Hot-image set.
   localTtl: 15,
   localMax: 10000,
+  localMaxBytes: 6 * 1024 * 1024, // hard heap cap ~6MB (tag-id arrays are small)
   async lookupFn(imageId, fromWrite) {
     const imageIds = Array.isArray(imageId) ? imageId : [imageId];
     const db = fromWrite ? dbWrite : dbRead;
@@ -356,6 +357,7 @@ export const userBasicCache = createCachedObject<UserBasicLookup>({
   // cross-pod delay on a rename/avatar/soft-delete propagating — imperceptible.
   localTtl: 30,
   localMax: 10000,
+  localMaxBytes: 4 * 1024 * 1024, // hard heap cap ~4MB (tiny records)
 });
 
 type ModelVersionAccessCache = EntityAccessDataType & { publishedAt: Date; status: ModelStatus };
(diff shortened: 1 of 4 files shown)


CivitaiDevOpsAgent:

PR #3307, opened 2026-07-23
feat: add MP4 and WebM generation metadata extraction

r-vage:
## Summary

- add bounded, dependency-free MP4 mdta and WebM SimpleTag generation-metadata readers
- reuse the existing Automatic and ComfyUI metadata processors for browser video uploads
- support MP4/WebM files and revocable local-video previews in the metadata extraction panel
- extend the existing Comfy metadata size guard to videos

## Verified writer compatibility

- validated end to end with a real PyAV MP4 written by the new **Save Video with Generation Data** node in [ComfyUI Eclipse](https://github.com/r-vage/ComfyUI_Eclipse)
- the node writes `parameters` through MP4 `mdta` using `use_metadata_tags+faststart`; this PR recovered its positive and negative prompts, sampler/scheduler, steps, CFG, seed, dimensions, denoise, CLIP skip, model/VAE names, LoRA weight, and model/VAE/LoRA/embedding hashes
- this is dedicated video-node support; the closed `alexopus/ComfyUI-Image-Saver#139` proposal and image-saving nodes are unrelated to this validation, and Eclipse maintains its own image generation-data saver separately

## Testing

- real Eclipse `Save Video with Generation Data` MP4 fixture through `VideoMetadataParser` and the existing Automatic processor (1 passed)
- `pnpm exec vitest run --project unit src/utils/metadata/__tests__/exif-parser.test.ts src/utils/metadata/__tests__/video-metadata-parser.test.ts` (32 passed, 1 skipped network test)
- `pnpm exec vitest run --project component src/components/generation_v2/inputs/MetadataExtractionPanel.browser.test.tsx` (11 passed)
- `pnpm typecheck` (0 errors)
- changed-file Prettier check passed
- changed-file ESLint completed with 0 errors (existing warnings only)
- `git diff --check` passed


JustMaier (changes requested):
Nice work on this. The parser is solid and the approach is the right one: normalizing container tags into the same `prompt` / `workflow` / `parameters` / `extraMetadata` keys our existing processors already expect means automatic, swarmui, rfooocus and comfy all light up for video with zero new processor code. Pulling `createMetadataParser` out of `ExifParser` is exactly the refactor I'd have asked for.

## What I verified

I merged the branch onto current `main` locally and tested against real files rather than just reading the diff:

| Check | Result |
|---|---|
| ffmpeg mp4 with `-movflags use_metadata_tags` | parses `prompt` + `workflow` |
| faststart mp4 carrying an A1111 `parameters` string | works |
| vp9 webm SimpleTag, piped webm, `-live 1` webm | all three work |
| Plain 290KB mp4 with no metadata | 11 blob slices / 176 bytes read / 2ms |
| Full `pnpm typecheck` on the merge with main | clean (only pre-existing `@civitai/client` errors, none from this PR) |
| `vitest run --project unit` on both metadata test files | 28 passed, 1 skipped |

Confirmed your box path matches reality: ffmpeg writes `moov/udta/meta/{keys,ilst}`, which is what you scan. The per-upload cost is negligible, so parsing on every video upload is fine.

## Needs a change before merge

**1. Blob object URL reaches the generation form.** Videos now put a `blob:` URL in `store.fileUrl`, and the effect cleanup revokes it. But `FormFooter.tsx:1397` (`MetadataExtractionFooter`) does `params.images = [fileUrl]` on a forceImage apply. Two problems: the URL is dead after unmount or a file swap, and a video URL is being handed to an image param in the first place. Please guard that path on media type, e.g. carry an `isVideo` flag in the metadata-extraction store and skip `forceImage` for video.

**2. Stale `ProcessedVideo.meta`.** `src/utils/media-preprocessors/index.ts:10` still hand-declares `meta?: Record<string, unknown>`, which was a placeholder from before `preprocessVideo` returned anything. Now that it really returns `meta`, drop the manual field so the type derives from the function, matching how `ProcessedImage` works.

**3. `isVideoFile` runs up to three times per file.** `MetadataExtractionPanel` sniffs the container in the drop handler, again in the fileUrl effect, and again in the metadata effect, each re-reading the blob header. Resolve it once and keep it in state.

**4. A late throw discards metadata already collected.** Every failure path in `video-metadata.ts` bubbles to the `readVideoMetadata` catch and returns `{}`, so one malformed element late in the file wipes out tags that parsed fine earlier. Please return what has been collected instead. Same idea in `readMp4DataValue`: when `dataType !== 1`, skip that item rather than aborting the whole file.

**5. Rebase.** The branch forked on 2026-07-02 and is about 1000 commits behind `main`, and no CI has run on it. It merges and typechecks clean locally, but please merge `main` in so CI actually exercises it against current code.

## Heads-up on a behavior change

Routing video `meta` through `auditMetaData` means videos carrying blocked prompt terms will now be blocked at upload, where previously they were skipped entirely. That is the correct and consistent behavior, but it is user-facing, so we want to land it deliberately rather than discover it. Same category: `Image.meta` will now populate for videos, which feeds generation data on the video detail page, `meta.engine` to tool detection, and `civitaiResources` to resource links. All wanted, just worth a look at the video detail UI once this is in.

## Follow-up scope (not blocking this PR)

There are a few places that do metadata extraction and still skip video. We are planning to handle these ourselves in a separate PR after this one merges, but if you want to take any of it, say so and it is yours:

- `post.service.ts:1163` gates the metadata fallback on `props.type !== MediaType.video`. That is the headless / MCP upload path, so API-posted videos still get no metadata. The blocker is that `readVideoMetadata` takes a `Blob` while `getMetadata` accepts a URL, so this needs a URL-capable entry point (ideally range requests, so we are not pulling a whole video to read a header).
- `image.service.ts:8051` calls `getMetadata(imageUrl)` and separately guesses `isVideo` from the file extension, so video URLs get ExifReader and yield nothing.
- `extract-source-metadata.ts` (enhancement and remix source metadata) is ExifParser-only.
- Underlying all three: a single `getMediaMetadata(file | url)` dispatcher that detects the container and routes to `ExifParser` or `VideoMetadataParser`, so `preprocessVideo`, the panel, and the server paths all call one thing instead of each hand-rolling their own video check.

## Minor, take or leave

- `detectVideoContainer` accepts any `ftyp`, including QuickTime. In `.mov` the `meta` box is not a FullBox, so the `+4` skip misaligns and silently yields `{}`. Harmless while the accept list is mp4/webm, but worth a comment or an explicit reject.
- The mp4 scan only covers `moov/udta/meta`. ffmpeg puts it there, but some writers use `moov/meta` for mdta. Cheap to also check.

Thanks for the thorough tests, they made this much faster to review.

r-vage:
Thanks for the detailed review and verification. I addressed all five blocking items in `3f2d5f6ef`:

  1. Added `isVideo` to the metadata-extraction store and prevented video URLs from being passed to image generation parameters.
  2. Removed the stale manual `ProcessedVideo.meta` declaration so the type derives from `preprocessVideo`.
  3. Resolved the media type once per local file and reused it for preview rendering and parser selection.
  4. Preserved metadata collected before malformed MP4/WebM elements and skipped unsupported MP4 `data` values instead of aborting parsing.
  5. Merged the current upstream `main` into the branch so CI runs against the latest code.

  Added regression coverage for:

  - Unsupported MP4 data values.
  - MP4 metadata preceding a malformed box.
  - WebM metadata preceding a malformed element.
  - Propagating the resolved video type through the metadata-extraction store.

  Validation completed locally:

  - Metadata unit tests: 31 passed, 1 skipped.
  - `MetadataExtractionPanel` Chromium component tests: 11 passed.
  - Full TypeScript check: passed.
  - Prettier check: passed.
  - ESLint: 0 errors; only pre-existing warnings.

  I also acknowledge the upload-auditing and video-detail behavior changes you called out. The non-blocking URL-based metadata dispatcher work is best kept in the planned follow-up PR.

LiamVisionary:
I've been building the writer side of this into a local generation studio, so rather than a +1 here's a data point.

I read through `video-metadata.ts` and matched a writer to it, then confirmed with `ffprobe` that the tags land where it looks:

- **MP4** — `ffmpeg -movflags use_metadata_tags -metadata parameters=… -metadata prompt=…` (the flag is required, or ffmpeg drops non-standard keys instead of writing `mdta`)
- **WebM** — plain `-metadata`, which becomes Matroska `SimpleTag`

Two things worth confirming for anyone else writing against this:

- ffmpeg writes Matroska tag names uppercased (`PARAMETERS`, `PROMPT`). `normalizeMetadataKey`'s `.toLowerCase()` already covers that, so no writer-side workaround is needed.
- Because the extracted values go through the same parser set as images, an A1111-style `parameters` string works as-is — including `, Civitai resources: [{…}]` for version linking. So a tool that already emits correct image metadata needs no video-specific format, just a different container. That made this a very small change on my side.

One note that may be relevant to prioritising: alexopus/ComfyUI-Image-Saver#139, which would have written this metadata on the ComfyUI side, was closed on 27 Aug — the maintainer felt video support was out of scope for that node pack. So writer-side support is going to live across forks and individual tools rather than one canonical node. I'd argue that makes reading it here more useful rather than less: there's no single upstream writer to wait on, but the tag conventions this PR targets are already what tools are emitting.

Happy to share sample MP4/WebM files written this way if that's useful for testing.

r-vage:
@LiamVisionary Thanks for the concrete writer-side details and the offer of fixtures. I followed up with an end-to-end writer/reader check rather than only a synthetic container test.

The current ComfyUI Eclipse development tree now includes a dedicated **Save Video with Generation Data** node: https://github.com/r-vage/ComfyUI_Eclipse. It writes a PyAV MP4 with `use_metadata_tags+faststart` and an A1111-compatible `parameters` tag. I generated a real five-frame MP4 through the node's actual execution path and passed it through this PR's `VideoMetadataParser` plus the existing Automatic processor.

The parser recovered the positive and negative prompts, sampler/scheduler, steps, CFG, seed, dimensions, denoise, CLIP skip, model/VAE names, LoRA name and weight, and model/VAE/LoRA/embedding hashes. Both the raw parse and schema-normalized metadata were verified.

To keep the scope clear for other readers: this compatibility is specifically provided by the new Eclipse video-saving node. The closed `alexopus/ComfyUI-Image-Saver#139` proposal is image-oriented and unrelated here; ComfyUI Eclipse also has its own separate, improved image generation-data saver.

I have also refreshed the PR branch onto current `main`. The focused metadata suite is at 32 passed / 1 skipped, the browser component suite is 11 passed, full TypeScript has 0 errors, and changed-file Prettier/ESLint checks are clean of errors.

diodiogod:
> To keep the scope clear for other readers: this compatibility is specifically provided by the new Eclipse video-saving node. The closed `alexopus/ComfyUI-Image-Saver#139` proposal is image-oriented and unrelated here; ComfyUI Eclipse also has its own separate, improved image generation-data saver.
> 

No, my PR was NOT image oriented. Did you bother to read it? The maintainer didn't want to merge it. It's his prerogative. [I'm keeping it as a fork.](https://github.com/diodiogod/ComfyUI-Video-Saver)

r-vage:
> > To keep the scope clear for other readers: this compatibility is specifically provided by the new Eclipse video-saving node. The closed `alexopus/ComfyUI-Image-Saver#139` proposal is image-oriented and unrelated here; ComfyUI Eclipse also has its own separate, improved image generation-data saver.
> 
> No, my PR was NOT image oriented. Did you bother to read it? The maintainer didn't want to merge it. It's his prerogative. [I'm keeping it as a fork.](https://github.com/diodiogod/ComfyUI-Video-Saver)
@diodiogod 
no i didnt because this repo is unrelated for me, but i just did and the eclipse video save node does exactly what you have described there, saving the a1111 parameters for the extraction on civitai when they finaly merge this.

r-vage:
@LiamVisionary @JustMaier please review

r-vage:
@JustMaier I've refreshed this branch with upstream `main` at `f92ccb5abe` (v5.1.127), merged in `bf19b695ad`, and updated the opening description to reflect the current implementation and validation.

The fixes from your review remain in place: video URLs are excluded from image parameters, video metadata types derive from the preprocessor, local media detection is reused, and malformed trailing elements preserve metadata already collected. No further source adjustments were needed for this merge, including with `@civitai/generation-metadata` 0.3.0.

Validation: 37 unit tests and 11 Chromium component tests passed, full TypeScript checking passed with 0 errors, and feature-file ESLint reported 0 errors. The optional image-corpus parity suite was skipped because its external fixtures aren't installed locally.

Could you take another look when you have a chance?

PR #3308, opened 2026-07-23
perf(creator-controls): flag-gate read-time metric-privacy resolution (default ON)

ZacxDev:
## What

Puts the **read-time Creator-Controls metric-privacy resolution added in #3266**
behind a new Flipt feature flag, **`modelMetricPrivacyReadtime`** (Flipt key
`model-metric-privacy-readtime`), so we can measure that resolution's
request-path cost with a clean on/off A/B.

**Default ON — merging + releasing is a pure no-op.** The privacy feature keeps
working exactly as it does today until someone flips the Flipt flag off.

## Why

#3266 added synchronous work to model hydration on read: a batched
`getValidCreatorMembershipMap` (which runs a `subscriptionProductMetadataSchema`
Zod parse per subscription), per-non-owner `hasValidCreatorMembership`, owner-
settings lookups, and per-item `resolveModel/VersionHiddenMetrics`. Because that
cost is spread across hydration rather than isolated in one frame, a profiler
can't cleanly attribute it. A flag that toggles the whole resolution on/off lets
us A/B the request-path cost directly and either confirm or exonerate it.

## Design

- **Flag:** `modelMetricPrivacyReadtime: { availability: ['public'], fliptKey:
  'model-metric-privacy-readtime' }`.
  - `['public']` = enabled for everyone **and fails OPEN (ON)** when Flipt is
    absent/down, so this is a no-op on merge/release (mirrors the
    `challengePlatform` / `animaControlnet` fail-open precedent — *not* the
    fail-closed `[]` pattern).
  - Flipt is authoritative when the `model-metric-privacy-readtime` flag exists;
    flipping it off is the only thing that changes behaviour.
- **Evaluated once per request** off the memoized per-request feature-flag object
  (`ctx.features` / threaded boolean). No per-model/version/owner flag lookups.
- **ON** → every gated path runs the existing resolution unchanged (byte-identical
  output).
- **OFF** → the batched membership map, the Zod parses, the owner-settings
  queries, `hasValidCreatorMembership`, and the per-item resolvers are **skipped
  entirely**; the gated paths return the raw (pre-#3266) metrics.
- **Central choke-point:** `gateHiddenMetrics(enabled, resolve)` in
  `utils/model-metric-privacy.ts` — returns the resolver result when ON, a fresh
  all-visible result (`noHiddenMetrics()`) when OFF.

## Gated read paths

| Path | File |
|---|---|
| `getModelHandler` (`model.getById`) | `server/controllers/model.controller.ts` |
| `getAssociatedResourcesCardDataHandler` | `server/controllers/model.controller.ts` |
| browse feed `getModelsWithImagesAndModelVersions` (`model.getAll`) | `server/services/model.service.ts` (threaded from `getModelsInfiniteHandler`) |

The feed helper takes a `metricPrivacyEnabled` param that **defaults to `true`**,
so its other callers (home blocks, collections) keep today's behaviour; only the
browse-feed controller threads the flag.

**Not gated (deliberately, see below):** the v1 REST model/version routes and the
OG-image endpoint (edge/origin response-cached, so the resolution is amortised and
not instantly reversible), the donation-goals cache lookup (a *shared* public
cache — a per-request flag can't coherently gate shared cached state without
changing the cache key), and the search-index build (write-time, not a read path).
These can be added if the A/B on the paths above is inconclusive.

## Tests

Unit tests in `server/__tests__/model-metric-privacy.test.ts` (mirrors the
existing style):

- `noHiddenMetrics()` returns an all-visible, freshly-allocated result.
- `gateHiddenMetrics` **ON** → invokes the resolver and returns its result
  (metrics hidden per current behaviour).
- `gateHiddenMetrics` **OFF** → resolver is **not** invoked (spy asserts it) and
  raw all-visible metrics are returned.

## Behaviour trade

Flipping the flag OFF briefly re-exposes Creator-Program members' hidden metrics
for the measurement window. That is the accepted, instantly-reversible trade for a
clean A/B — keep the OFF window short.

## Notes

- Do **not** merge — this is a measurement/review artifact.
- No caches, batching, or schema were changed.

### DIFF
--- a/src/server/__tests__/model-metric-privacy.test.ts
+++ b/src/server/__tests__/model-metric-privacy.test.ts
@@ -1,8 +1,10 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import {
   anyMetricHidden,
+  gateHiddenMetrics,
   getMetaMetricPrivacy,
   getUserMetricPrivacyDefaults,
+  noHiddenMetrics,
   resolveModelHiddenMetrics,
   resolveVersionHiddenMetrics,
 } from '~/server/utils/model-metric-privacy';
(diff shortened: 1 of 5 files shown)


CivitaiDevOpsAgent:

PR #3303, opened 2026-07-23
feat(app-blocks): tabbed agent-review report (fix wall-of-text) [Phase 2]

ZacxDev:
## What
Phase 2 of the App Blocks review-page migration — fixes the **wall-of-text** in the agent-review report (the third of the original mod-feedback items). The report becomes **tabbed**: `Summary | Code review (N) | Security audit (N) | Scopes (N)`, counts in the tab labels, findings as per-item cards sorted by severity (critical→info), counts-first.

## How
- New reusable **`ReportTabs.tsx`** — `SummaryTab`/`CodeReviewTab`/`SecurityAuditTab`/`FindingCard` exported so the **offsite** listing review can reuse them (Q7). `AgentReviewPanel`'s `ReportBody` now delegates to it (−534 lines).
- `summaryMd` renders via `CustomMarkdown` with **`disallowedElements={['img']}`** — same adversarial guard as the chat (no external image fetch from the moderator's browser on prompt-injectable output).
- **Closes the carried-forward stored-XSS-at-render P2 concern**: all finding text goes through React (auto-escaped); only `summaryMd` through `CustomMarkdown` (no `rehype-raw`).
- Graceful **failed/errored-section** states (a stage can be an `{"error":...}` object).
- Report renders in **both** the queue modal and the new page (shared component).

## Not in scope (separate 2.3 follow-up)
Sticky approve/reject action bar, page focus-management/route-leave guard, deep-link-to-finding permalinks, and the offsite page itself.

## Tests (all green locally, real Chromium)
- `AgentReviewPanel.browser.test.tsx`: **26** (tabs + counts, tab-switching, severity-sorted cards, summaryMd markdown + image-dropped, failed/empty states)
- `agentReviewReport.test.ts`: **13** parser-helper unit
- `OnsiteReviewModal.browser.test.tsx`: **14/14 unchanged** — behavior-preserving extraction proof
- `tsc --noEmit`: zero errors reference the changed files (remaining are pre-existing env artifacts — stale Prisma client / uninstalled `kysely` / `event-engine-common` submodule)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3309, opened 2026-07-23
feat(app-blocks): sticky review action bar + route-leave guard + a11y on the review page [Phase 2.3]

ZacxDev:
## Phase 2.3 — page action/a11y layer for the App Blocks review page

Phase 2 core (tabbed report redesign, #3303) already merged. This is the
page-specific action + accessibility layer for `/apps/review/<publishRequestId>`:
a sticky approve/reject action bar, a route-leave guard, and focus / aria-live
management.

### Sticky approve/reject action bar
The approve/reject controls used to render inline at the end of the (now long,
tabbed) review body, so a moderator had to scroll to the bottom to act. They are
now extracted into a reusable **`<ReviewActionBar>`** and, on the page, pinned in
a **sticky bottom bar** that is always reachable.

- **Shared body kept byte-behavior-unchanged for the modal.** `OnsiteReviewModalBody`
  gains a `hideInlineActions` prop; the modal renders `<ReviewActionBar>` inline
  in its footer exactly as before (keyed remount → per-request state reset
  preserved), the page passes `hideInlineActions` and renders the same bar itself.
  `busyRef` is now optional (modal-only close-guard). The reason-gate flow
  (`ReasonGatedField`/`ReasonGatedSubmitButton`, min 3), the approve/reject
  mutations, list-query invalidation, and the Q6 redirect-to-queue-on-success are
  all preserved exactly.
- **`ReviewActionBar` is prop-clean for Q7 offsite reuse** — it needs only a
  `selection`, an `onClose`, and two optional observers (`busyRef` for the modal,
  `onStatusChange` for the page). No onsite-only assumptions.

### Route-leave guard
`useReviewNavigationGuard` replaces the modal's `busyRef` close-refusal on the
page. While an approve/reject mutation is in flight it blocks navigation via the
pages-router idioms — `routeChangeStart` (throw-to-cancel), `beforePopState`
(veto back/forward) and `beforeunload` (tab close / reload) — and tears all three
down the moment the mutation settles. Router: **`next/router`** (pages router;
`[publishRequestId].tsx`).

### Focus + a11y
- On mount, focus moves to the labelled **main review region** (`role="region"`,
  `tabIndex={-1}`) instead of being stranded on `<body>` after the route change
  (a page has no focus trap like the modal did).
- An **`aria-live="polite"`** region announces status transitions (submitting /
  approved / rejected / error).
- The sticky bar is a keyboard-reachable labelled `role="group"`, in normal tab
  order, not a focus trap. Tabs a11y is already handled by Mantine.

`ReviewDetailView` was factored out of the page module (server-graph-free,
browser-testable); the page stays a thin SSR-gate + client-query wrapper.

### Modal path unchanged (behavior-preserving)
The 14 `OnsiteReviewModal.browser.test.tsx` tests pass **unchanged** — the proof
the shared extraction didn't regress the modal (approve fires, reject reason-gate,
per-request state reset, the busy close-guard, read-only posture).

### Tests
- **34 browser tests pass** (`--project component`): 14 modal (regression,
  unchanged) + 8 `ReviewActionBar` + 3 `useReviewNavigationGuard` + 9
  `ReviewDetailView`. Real behavior driven (click approve → reason-gate →
  mutation → redirect; in-flight → guard registers; focus lands on region;
  aria-live updates). Each new load-bearing behavior was mutation-checked
  (disabling focus / the guard fails exactly the tests that should).
- `NODE_OPTIONS=--max_old_space_size=8192 tsc --noEmit`: **zero errors in changed
  files** (only pre-existing env errors: `kysely`, `event-engine-common`
  submodule, `image.service.ts` implicit-any).

### Deferred
The deep-link-to-a-finding permalink (open the right report tab / scroll to a
finding) is deferred as a **fast-follow** — it adds tab-state + scroll-sync
complexity beyond this PR's action/a11y scope.

### DIFF
--- a/src/components/Apps/OnsiteReviewModal.tsx
+++ b/src/components/Apps/OnsiteReviewModal.tsx
@@ -14,10 +14,10 @@ import {
   Stack,
   Switch,
   Text,
-  Textarea,
   ThemeIcon,
   Tooltip,
 } from '@mantine/core';
+import { ReviewActionBar } from '~/components/Apps/ReviewActionBar';
 import {
   IconAdjustmentsAlt,
   IconAlertTriangle,
@@ -42,14 +42,8 @@ import {
   ManifestDiffPreview,
   type FileLineDiff,
 } from '~/components/Apps/reviewDiffPanels';
-import {
-  ReasonGatedField,
-  ReasonGatedSubmitButton,
-  reasonMeetsMin,
-} from '~/components/Apps/ReasonGatedActionModal';
 import { useFeatureFlags } from '~/providers/FeatureFlagsProvider';
 import { isSensitiveBlockScope } from '~/shared/constants/block-scope.constants';
-import { OFFSITE_MOD_REASON_MIN } from '~/server/schema/blocks/offsite-moderation.schema';
 import {
   MARKETPLACE_CATEGORIES,
   MARKETPLACE_CATEGORY_LABELS,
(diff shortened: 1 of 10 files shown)


CivitaiDevOpsAgent:

PR #3313, opened 2026-07-23
feat(collections): submit models to active contests from the Save picker

JustMaier:
## What

Lets users submit their **models** to an active contest collection directly from the model **Save** (Add to Collection) picker, **without having to follow the collection first**.

## Why

Contest collections are `mode=Contest`, `write=Review`, `read=Public`. Previously they only appeared in a user's Save picker if they were already a *contributor* — which only happens by **following** the collection. So submitting a model to a contest meant: follow the collection → then Save your model to it. For model contests there's also no on-page submit button (those are Image/Post only), so following was the sole path.

This surfaces **active** contest collections in the model picker under an "Active Contests" group, so the flow becomes: model → Save → pick the contest → choose a category → done.

## How

- **Schema**: adds optional `includeActiveContests` to `getAllUserCollectionsInputSchema` (default off — existing callers unaffected).
- **Server** (`getUserCollectionsWithPermissions`): when the flag is set, one extra UNION branch returns `Contest` + `Review`-write + `Public`-read collections whose submission window is currently open (`submissionStartDate <= now <= submissionEndDate`, either bound nullable), respecting the existing type filter, bounded by `LIMIT 100`. Adds `c."mode"` to the shared SELECT so the client can group contests.
- **Save gate** (`saveItemInCollections`): relaxes the contributor gate to let review/write submitters through — `if (!isContributor && !isOwner && !writeReview && !write) return null;` — so a non-follower with `writeReview` lands their entry as `status: REVIEW` (awaiting moderation) instead of being silently dropped. This is required because contest collections set `disableFollowOnSubmission=true`, which skips the auto-follow that previously granted contributor status.
- **Client** (`AddToCollectionModal`): requests active contests **only for Model saves**, renders them as a distinct "Active Contests" group, category-tag picker intact.

## Review

Went through an adversarial review loop. Findings addressed:
- **CRITICAL** (fixed): non-follower saves were dropped by the contributor gate for `disableFollowOnSubmission` collections. Gate relaxed to honor `writeReview`/`write`.
- **MEDIUM** (fixed): unbounded global scan on the picker hot path → `LIMIT 100`.
- **HIGH** (intentional, not changed): the `submissionStartDate` validator rejects models created before the start date. For training contests this is the intended rule (fresh models only).

Scope verified: no unlisted/private/wrong-type leak; `includeActiveContests` default-off; parametrized SQL; `UNION` de-dupes.

## Testing

- `pnpm typecheck`: zero errors in the touched files (pre-existing `kysely` module-resolution errors in `packages/civitai-db*` are environmental to the worktree, unrelated).
- Please validate the end-to-end submit in preview: as a non-follower, Save a Krea 2 model to the "Krea 2 Training Contest" collection from the model page, pick a category, confirm it lands as a REVIEW entry.

### DIFF
--- a/src/components/Collections/AddToCollectionModal.tsx
+++ b/src/components/Collections/AddToCollectionModal.tsx
@@ -173,6 +173,10 @@ function CollectionListForm({
   const queryUtils = trpc.useUtils();
   const [selectedCollections, setSelectedCollections] = useState<SelectedCollection[]>([]);
 
+  // Model saves also surface active contest collections the user hasn't joined, so they can
+  // submit an entry for review without following the collection first.
+  const includeActiveContests = props.type === CollectionType.Model;
+
   const { data: collections = [], isLoading: loadingCollections } =
     trpc.collection.getAllUser.useQuery({
       // Only request collections where the user can actually add items.
@@ -184,6 +188,7 @@ function CollectionListForm({
         CollectionContributorPermission.ADD_REVIEW,
       ],
       type: props.type,
+      includeActiveContests,
     });
 
   const { data: collectionItems = [], isLoading: loadingStatus } =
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:
### [PR-REVIEW] round 3 — 0 finding(s) (confidence ≥ 0.80)

PR #3314, opened 2026-07-23
fix(articles): skip image scan when a moderator NSFW override is set

JustMaier:
## Problem

An article with a **moderator-set NSFW override** still had its visibility gated on image-scan state. Any edit (or any not-yet-finished / stuck image scan) left `ingestion = Pending`, which drops the article from the feed and the search index — even though `moderatorNsfwLevel` is already the authoritative rating. For official/curated articles a trivial edit effectively unpublished the article for the duration of a scan, and a stuck scan stranded it indefinitely.

## Fix

One change, in `recomputeArticleIngestionInTx` (`src/server/services/article.service.ts`) — the single choke point every scan webhook, text-moderation webhook, edit, and reconcile recompute flows through:

```diff
-  } else if (imageDone && textDone) {
+  } else if ((imageDone && textDone) || hasModeratorOverride) {
```

plus selecting `moderatorNsfwLevel` to derive `hasModeratorOverride`. **+16/-1.**

The override substitutes for the **`Pending` fallback only**. `Blocked` and `Error` still take precedence — an override is a rating decision, not a moderation bypass.

Because this sits at the recompute choke point, it also frees articles already stranded on stale Pending images, and covers cover-image changes, with no changes to the upsert path.

## Why not more than this

The first pass of this PR also skipped image linking/scanning in `upsertArticle` and put the override ahead of the blocked/error branches. Adversarial review (correctness + security lenses, independently) returned BLOCK on both:

1. **Forcing `Scanned` outranked `Blocked`.** A policy-blocked image — including the pHash CSAM path in `image-scan-result.service.ts` — could no longer drive `ingestion = Blocked`. The claim that "nsfwLevel carries the block instead" is **false**: `updateArticleNsfwLevels` joins only `ingestion = 'Scanned'` images and `COALESCE`s `moderatorNsfwLevel` over the derived level, so a blocked image's level never reaches `nsfwLevel`. `ingestion = Blocked` was the only remaining guard.
2. **`cleanupOnly` gating orphaned new uploads.** Images newly added to an overridden article got no `Image` row at all — never scanned, never blockable, and unrecoverable, since no cron revisits a `Published` + `Scanned` article. An owner can induce an override just by posting borderline content.

Both are reverted. Images link, create, and enqueue for scanning exactly as on `main`, and the publish-time blocked/error gate is restored.

## Behavior

| State | Before | After |
|---|---|---|
| Override + images Pending | `Pending` → hidden | `Scanned` → visible at override level |
| Override + an image `Blocked` | `Blocked` → hidden | `Blocked` → hidden (unchanged) |
| Override + scan `Error` | `Error` → hidden | `Error` → hidden (unchanged) |
| No override | unchanged | unchanged |

Scans still run in the background on overridden articles; if one comes back Blocked, the article correctly drops out of the feed.

## Test plan

- [x] `pnpm run typecheck`
- [ ] Edit an article with a mod override → stays Published and in-feed, no Processing bounce.
- [ ] Overridden article with a blocked image → still hidden (`ingestion = Blocked`).
- [ ] Article stranded on Pending images gets an override → next recompute flips it visible.
- [ ] No override → scan/publish behavior unchanged.

### DIFF
--- a/src/server/services/article.service.ts
+++ b/src/server/services/article.service.ts
@@ -948,6 +948,17 @@ export const upsertArticle = async ({
     // article is guaranteed non-null here since the `if (id)` block above fetches it
     if (!article) throw throwNotFoundError();
 
+    // A moderator-set NSFW override is the authoritative rating for the article
+    // (see updateArticleNsfwLevels), so editing an overridden article must not
+    // (re)trigger image scanning: a rescan would bounce the article into
+    // Pending/Processing — hiding it from the feed until a scan completes whose
+    // result the override supersedes anyway. Treat scanning as disabled whenever
+    // an override is (or is being) set. Clearing the override (setting it null)
+    // re-enables scanning and returns the article to auto-derivation.
+    const effectiveModeratorNsfwLevel =
+      data.moderatorNsfwLevel !== undefined ? data.moderatorNsfwLevel : article.moderatorNsfwLevel;
+    const scanContentEffective = scanContent && effectiveModeratorNsfwLevel == null;
+
     // userNsfwLevel is the user's preference and is preserved verbatim. The
     // effective article.nsfwLevel is derived by updateArticleNsfwLevels as
     // GREATEST(userNsfwLevel, cover, content images, moderation floor), so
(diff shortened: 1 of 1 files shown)


CivitaiDevOpsAgent:

- **medium** `src/server/services/article.service.ts` (data-integrity, conf 0.85) — Images added to article content while a moderator NSFW override is active are never created as Image entities or ImageConnections because linkArticleContentImages runs with cleanupOnly:true, so they remain invisible to updateArticleNsfwLevels and recomputeArticleIngestionInTx even after the override is later cleared without a simultaneous content edit, causing the article's nsfwLevel to be computed from a stale image set.  
  ↳ _When an override suppresses scanning, still create Image entities and ImageConnections for new content images (e.g. with ingestion=Scanned mirroring the override force-path) so the tracked image set stays accurate for future recomputes, or force a full re-link when the override transitions from non-null to null._

PR #2950, opened 2026-07-06
feat(app-blocks): W13 P3a — off-site listing submission backend (dark)

ZacxDev:
## What

W13 **P3a PR-a** — the native **external-link off-site app submission backend**, fully **dark** behind the `app-blocks-author` flag (mods + app-dev-testers). No UI. This is the submission leg only; approve/reject (PR-b), UI (PR-c), and #2821 retirement (PR-d) follow as separate PRs off `main`.

## Design B1 (locked — draft-at-submit)

`submitExternalListing` creates, in **one transaction**:
1. a DRAFT `AppListing(kind='offsite', status='draft', externalUrl, slug, name, tagline?, description?, category?, contentRating (author-declared, default 'g'), userId=caller, connectClientId=null, appBlockId=null)`, and
2. an `AppListingPublishRequest(kind='offsite', status='pending', slug, submittedByUserId=caller, appListingId=<draft id>, changelog?)`.

The draft lets the author reuse the P1 asset-CRUD (owner-gated) to attach icon/cover/screenshots before a mod approves. The P2a read path already hides non-approved rows, so a draft never surfaces in the store. **Slug-squat protection is free** from `AppListing.slug @unique` — no new pending-per-slug partial-unique, **no migration**. Only a **comment** on `AppListingPublishRequest.appListingId` changes (B1 sets it at submit) — comment-only, no DDL.

`withdrawExternalRequest` is **terminal + author-initiated**: it **deletes the draft `AppListing`** (releases the slug, cascades screenshots).

## Changes

- **`offsite-listing.schema.ts`** — `submitExternalListingSchema` (name/externalUrl/slug/tagline?/description?/category?/`contentRating` default `g`/changelog?). URL + surface validation **delegated** to `validateExternalUrl` / `assertNoOnPlatformSurface` / `MAX_EXTERNAL_URL_LENGTH` from `external-app.schema` (no re-implementation).
- **`offsite-listing.service.ts`** — `submitExternalListing` (owner-bound via `ctx.user.id`; slug pre-check against `AppListing.slug` AND `AppBlock.block_id`; P2002-race branch → same friendly `slug "X" already taken`), `withdrawExternalRequest` (IDOR `NOT_OWNED` + TOCTOU status-guarded `updateMany` + idempotent-withdrawn, mirrors `publish-request.service`), `listMySubmissions`, and read-only mod queue lists (`listPending/Approved/RejectedOffsiteRequests`).
[...]

PR #3317, opened 2026-07-23
fix(tests): component-suite build death (sharp) + smoke external-listing OAuth payloads

ZacxDev:
Two independent baseline breaks that fail on **every PR preview** (unit baseline was fixed in #3262; these are the COMPONENT and SMOKE suites). Both confirmed failing on current `main` HEAD, base `f4f025d6`.

## BREAK 1 — COMPONENT suite fails to BUILD (whole suite dies) → config fix

**Reproduced on main:** running the `component` project dies during esbuild's optimizeDeps scan:
```
✘ Could not resolve require("../build/Release/sharp-*.node")
   node_modules/.pnpm/sharp@0.32.6/.../sharp/lib/sharp.js:10:27
Build failed with 2 errors → Unhandled Rejection → NO component test runs
```

**Root cause (traced the import graph):** two page-importing browser tests —
`review-detail-page.browser.test.tsx` and `review-queue-nav.browser.test.tsx` — import a Next **page** to render its client shell:
```
<test> → src/pages/apps/review/[...].tsx → server/utils/server-side-helpers.ts
       → server/routers/index.ts → creator-shop.router.ts → creator-shop.service.ts → import sharp from 'sharp'
```
Next strips the server-only `getServerSideProps` graph from real client builds; Vitest's browser build does **not**, so esbuild follows the import into `sharp` and dies on its native `.node` `require`. (The tests already `vi.mock('~/server/utils/server-side-helpers')`, but that's a **runtime** interception — it can't stop the **build-time** static scan.)

**Fix (build-level, tight):** alias `sharp` → a trivial stub (`test/stubs/sharp.ts`) **for the `component` project only**; the `unit` project keeps real sharp (used by server unit tests). More robust than `optimizeDeps.exclude` alone — it fixes the build *and* is harmless if a browser test ever touches it. No product logic touched.

**Verified locally** (Playwright chromium, browser mode): the sharp `.node` error is **gone** and the affected tests build + execute.

<details><summary>Also fixed: a stale test unmasked by the build fix</summary>

Once the suite built, `review-detail-page.browser.test.tsx` threw `trpc.useUtils is not a function` (2 tests). The page was refactored to render the review body via **`ReviewDetailView`** (which owns the real trpc-backed `ReviewActionBar`), but the test still stubbed the old `OnsiteReviewModalBody`. Re-pointed the body stub to `ReviewDetailView` — matches the test's stated intent (assert shell wiring, not re-run the action bar's own covered behaviour). **5/5 pass.**
</details>

## BREAK 2 — SMOKE `submitExternalListing` → HTTP 400 → stale-test fix (NOT a regression)

**Error:** `connectClientId expected string received undefined; requestedScopes expected number received undefined`.

**Judgment — is the schema-tightening a real regression?** No. The three `preview-apps-external-*` specs were last touched in #2950; **#3227 later _merged_ OAuth-connect into the single external-app submit flow** ("every external app IS an OAuth app"). `connectClientId` + `requestedScopes` + `scopeJustifications` are now **unconditionally required for ALL external listings by design** — extensively documented and rationalized in `offsite-listing.schema.ts`. This is intended; **source is left untouched**. The smoke payloads are simply stale.

**Fix:** each submit-bearing spec now creates a throwaway **owned** OAuth client (`oauthClient.create`), passes its id as `connectClientId` with an **empty scope disclosure** (`requestedScopes: 0` + `scopeJustifications: {}` — `0` ⊆ any client ceiling; no scopes ⇒ no justifications), and deletes the client on cleanup (self-cleaning like the draft + slug). The service requires a real owned client (`loadConnectClientForListing`), so creating one at runtime is the correct approach — there is no seeded client in the preview setup.

**Verified:** the new payload passes the **real** `submitExternalListingSchema` zod schema (and the old payload still fails — proving break + fix); existing `offsite-listing*.schema.test.ts` = 45/45 green. The **end-to-end preview run** (the `connectClientId` ownership check is service/DB-level) is **CI-to-confirm**.

## Files
- **config/stub:** `vitest.config.mts`, `test/stubs/sharp.ts`
- **tests:** `src/tests/pages/apps/review/review-detail-page.browser.test.tsx`, `tests/preview-apps-external-{approve,delist,submit}.spec.ts`

### DIFF
--- a/src/tests/pages/apps/review/review-detail-page.browser.test.tsx
+++ b/src/tests/pages/apps/review/review-detail-page.browser.test.tsx
@@ -36,12 +36,17 @@ vi.mock('~/providers/FeatureFlagsProvider', () => ({
 }));
 
 // Stub the extracted body/title so we assert the SHELL wiring (props + gate),
-// not re-run the body's own covered behaviour.
-vi.mock('~/components/Apps/OnsiteReviewModal', () => ({
-  OnsiteReviewModalBody: (props: { selection: any; onClose: () => void }) => {
+// not re-run the body's own covered behaviour. The page renders the review body
+// via `ReviewDetailView` (which owns the real approve/reject `ReviewActionBar`,
+// covered by its own tests); stub it so this shell test doesn't pull that live
+// trpc-backed action bar. The title still comes from `OnsiteReviewModal`.
+vi.mock('~/components/Apps/ReviewDetailView', () => ({
+  ReviewDetailView: (props: { selection: any; onClose: () => void }) => {
     state.bodyProps.last = props;
     return <div data-testid="review-body">body:{props.selection.request.id}:{props.selection.mode}</div>;
   },
+}));
+vi.mock('~/components/Apps/OnsiteReviewModal', () => ({
   OnsiteReviewModalTitle: ({ selection }: { selection: any }) => (
     <div data-testid="review-title">{selection.request.slug}</div>
   ),
--- a/test/stubs/sharp.ts
+++ b/test/stubs/sharp.ts
@@ -1,0 +1,32 @@
+/**
+ * Browser-mode (`component` project) stub for the native `sharp` module.
+ *
+ * A couple of `.browser.test.tsx` tests import a Next *page* to render its CLIENT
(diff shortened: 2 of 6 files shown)


CivitaiDevOpsAgent:

PR #3322, opened 2026-07-23
perf(creator-controls): cache read-time membership validity + skip resolution when nothing hidden

ZacxDev:
## What

Optimizes the read-time metric-privacy path (#3266) so it does far less synchronous work on hot model-read endpoints, with **byte-identical output** — the same metrics stay hidden for the same viewers. Pure cost optimization; no behaviour change.

The `#3308` `model-metric-privacy-readtime` flag/gate is left in place untouched (it stays the kill-switch; this change makes its ON path cheap).

## Why

The gate resolves per request on hot paths:
- `getValidCreatorMembershipMap` runs a per-request `customerSubscription.findMany` **plus a Zod `subscriptionProductMetadataSchema.parse()` per subscription**, and
- every gated surface resolves per-model / per-version hidden metrics.

Most creators don't hide any metric, so the vast majority of that work produces "nothing hidden".

## The two fixes (they compound)

**FIX 1 — read-through Redis cache for per-user membership validity.**
Membership validity is near-static per user, so `getValidCreatorMembershipMap` now:
- `mGet`s the cached `id -> boolean`,
- DB-queries + Zod-parses **only the misses**, then backfills them,
- **fails open** to the uncached DB path on any Redis error (a Redis stall never 500s a read),
- caches **both `true` and `false`** (the resolver is a total function over the id).

A single-user cache-backed helper `hasValidCreatorMembershipCached` routes the `getModel` / v1 version-response / OG-card paths through the same cache. The shop / creator-program **action gates keep reading live subscription state** (unchanged).

Cache key: `packed:caches:creator-membership-valid:<userId>`. Busted immediately on every app-driven subscription change via `invalidateSubscriptionCaches` (the existing subscription-cache fan-out: Stripe/Paddle webhooks, code redemption, cancel/reinstate, mod tooling). A short TTL (`CacheTTL.md`, 10 min) backstops the non-webhook writers (referral grants, renewal/prepaid crons, direct-DB edits).

**FIX 2 — short-circuit the resolution when nobody hides.**
Before any membership lookup, the single-model read paths check the cheap, already-fetched model/version-meta + owner-user-default hide flags. If nothing is hidden, the membership lookup **and** the resolvers are skipped and raw metrics are returned. (The batch feed / associated-resources / v1-list paths already filter membership candidates this way; FIX 1 caches the residual for them.)

This is provably byte-identical: when no flag is set, the resolvers return `NONE` for **both** membership values, so skipping the lookup cannot change the output.

## Paths covered

`getModelHandler`, `getAssociatedResourcesCardDataHandler`, `getModelsWithImagesAndModelVersions`, the v1 REST paths (`getModelsWithVersions` / `prepareModelVersionResponse`), `api/og.tsx`, and the donation-goal hide cache — all share `getValidCreatorMembershipMap`, so FIX 1 benefits every caller.

## Tests

- Membership cache: hit, miss, batch miss-fill (queries only the misses + backfills), fail-open on Redis error, negative caching, bust, and **cache-served == uncached-DB** for a mixed batch.
- Short-circuit invariant: resolvers return `NONE` for member **and** non-member when no flag is set (so the skip is byte-identical), and membership still matters once any flag is set.

`tsc --noEmit`: no new type errors.

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -2091,6 +2091,12 @@ export const REDIS_KEYS = {
     // registered image resolves immediately), busted when `hideMeta` flips in
     // updatePostImage. See `getCachedImageDeliveryMetadata` in image-delivery.service.
     IMAGE_DELIVERY_METADATA: 'packed:caches:image-delivery-metadata',
+    // Per-user `id -> isValidCreatorMember(boolean)` for the read-time metric-privacy /
+    // donation-goal hide gate (#3266). Near-static per user; both TRUE and FALSE are
+    // cached (the resolver is a total function over the id). Busted on any subscription
+    // change via `invalidateSubscriptionCaches`; TTL backstops the non-webhook paths.
+    // See `getValidCreatorMembershipMap` in creator-membership.service.
+    CREATOR_MEMBERSHIP_VALID: 'packed:caches:creator-membership-valid',
   },
   RESEARCH: {
     RATINGS_COUNT: 'research:ratings-count',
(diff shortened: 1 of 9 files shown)


CivitaiDevOpsAgent:

PR #3323, opened 2026-07-23
fix(api): purge by-hash edge cache on version unpublish/delete

ZacxDev:
## Problem

`GET /api/v1/model-versions/by-hash/[hash]` is a `PublicEndpoint`, so it's edge-cached via `Cache-Control` (`public, s-maxage=...`) and emits **no `Cache-Tag`**. The existing tag-based `purgeCache` (which resolves tags → URLs out of Redis) therefore can't target it. When a model version is **unpublished** or **deleted**, the endpoint's Published-only filter means it *should* stop resolving that version — but the previously-cached response keeps being served at the CDN edge until the cache TTL expires. Symmetrically, when a version is **published**, a previously-cached `404` for its hash keeps being served until it expires.

## Change

Add a best-effort edge-cache purge to the three mutation paths in `model-version.service.ts`:

- `deleteVersionById` — snapshots the version's file hashes **inside the transaction** (the cascade nukes `ModelFileHash` rows) and purges post-commit.
- `unpublishModelVersionById` — resolves the still-present file hashes and purges.
- `publishModelVersionById` — purges any cached `404` so a freshly-published version resolves immediately.

Each purge is wrapped in `try/catch` + `logToAxiom`, matching the surrounding post-commit side-effect pattern: **a purge failure must not fail the (already-committed) mutation.**

### Mechanism: URL purge (not Cache-Tag)

I chose **URL purge** over adding a `Cache-Tag`:

- The endpoint is a **Next.js API route**, not a tRPC procedure. The tag mechanism (`edgeCacheIt` → `purgeCache({ tags })`) works by maintaining a Redis set `EDGE_CACHED:<tag>` that maps a tag to the actual cached URLs, and that set is populated **only** by the tRPC `edgeCacheIt` middleware. To use tags here I'd have to replicate that Redis registration on every by-hash response (growing an unbounded index keyed by every unique hash) *and* add a `Cache-Tag` header — more moving parts for the same result.
- `purgeCache` already supports direct URL purge (`client.zones.purgeCache(zoneId, { files })`), and the by-hash URLs are deterministic and cheap to build from the version's `ModelFileHash` rows.

The endpoint's zod schema uppercases the hash, so the canonical cache key is the uppercase URL. Because the Cloudflare cache key is the raw request URL (case-sensitive) and clients may request either casing, the purge covers **both the upper- and lower-case variant** of each hash. The origin comes from `getBaseUrl()` (env-derived), not a hardcoded domain.

## Best-effort / non-fatal

All three purges are guarded — a Cloudflare/network failure is logged and swallowed, never rolled into the mutation's success path. `purgeCache` itself already no-ops when CF env is absent.

## TTL note

This PR keeps the by-hash `s-maxage` unchanged. Once this hook is verified in prod, the by-hash `maxAge` can be raised back toward 24h in a follow-up, since takedowns/publishes now evict the edge cache promptly instead of relying on a short TTL.

## Testing

`src/server/services/__tests__/model-version.purge-by-hash.service.test.ts` (mirrors the existing `model-version.deregister.service.test.ts` conventions), mocking `purgeCache`:

- delete purges the exact upper+lower by-hash URLs for the version's hashes
- delete purges every hash across every file, deduped
- delete does **not** purge when the version has no hashes
- delete still succeeds if the purge throws (best-effort)
- unpublish resolves hashes via `modelFileHash.findMany` and purges the expected URLs
- unpublish still succeeds if the purge throws

```
vitest run model-version.purge-by-hash.service.test.ts → 6 passed
```

ESLint could not run locally (ESLint 8.57 crashes on config load under Node 26); the change is prettier-clean. Full `tsc`/lint left to CI.

## Follow-up (audit fixes)

- **`purgeCache` now awaits the Cloudflare call.** The per-batch task previously called `client.zones.purgeCache(...)` without awaiting/returning it, so the concurrency helper's `await task()` resolved before the CF request settled — a CF rejection (429 / timeout / auth) escaped as an unhandled promise rejection that no caller could observe. The CF call is now awaited inside each batch and a failure is logged + swallowed there, so `purgeCache` stays best-effort for every caller (no new throw path) while CF failures are actually observed. A new `client.purgeCache.test.ts` proves a CF rejection is swallowed + logged and does not propagate. The caller-level "purge throws" guards remain as defensive tests (they now cover non-CF throws from the purge path).
- **Added a `publishModelVersionById` by-hash purge test** — the highest-frequency path was previously untested. It asserts the exact upper+lower by-hash URL set and arg shape, plus the best-effort-on-throw behaviour. Suite: 10 passed.

## Known trade-offs (audited, accepted)

- **`publishModelVersionById` purges on every publish** to flush a possibly-cached `404` for the just-published hash. This is an unconditional CF call on a common path; accepted given by-hash purge volume is low (a handful of URLs per publish) and the purge is best-effort / off the request success path.
- **Both hash casings are purged** (upper + lower), doubling the URL count per hash. Deliberate: the CDN cache key is the case-sensitive raw request URL and clients may request either casing, so purging both variants is the robust way to guarantee eviction.
- **Only the primary origin is purged**, not the color-alias domains. This matches the existing tag-purge limitation — the same single-origin behaviour the tag path already has — so it introduces no new gap.

Follows up #3320 / #3321.

### DIFF
--- a/src/server/cloudflare/__tests__/client.purgeCache.test.ts
+++ b/src/server/cloudflare/__tests__/client.purgeCache.test.ts
@@ -1,0 +1,71 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Proves the Cloudflare edge-cache purge is best-effort at the SOURCE.
+// purgeCache now AWAITS each CF batch call (so its outcome is reflected in the
+// promise the concurrency helper awaits) AND swallows + logs a failure so it
+// never propagates to callers. Pre-fix the CF call was fire-and-forget: the
+// task resolved before the CF request settled, so a 429 / timeout / auth
+// rejection escaped as an unhandled promise rejection and no caller guard could
+// observe it. This is the real proof of the fix — the service-level guard tests
+// mock purgeCache wholesale and can't exercise this path.
+
+const { mockCfPurge } = vi.hoisted(() => ({ mockCfPurge: vi.fn() }));
+
+vi.mock('cloudflare', () => ({
+  default: class {
+    zones = { purgeCache: mockCfPurge };
+    constructor(_opts: unknown) {}
+  },
+}));
+
+vi.mock('~/env/server', () => ({
+  env: {
+    CF_API_TOKEN: 'test-token',
+    CF_ZONE_ID: 'test-zone',
+    LOGGING: 'cloudflare', // enable createLogger so the failure log is observable
+  },
+}));
+
+vi.mock('~/server/redis/client', () => ({
+  redis: { sMembers: vi.fn(), del: vi.fn() },
+  REDIS_KEYS: { CACHES: { EDGE_CACHED: 'edge-cached' } },
+}));
+
+import { purgeCache } from '~/server/cloudflare/client';
+
+const URL = 'https://civitai.com/api/v1/model-versions/by-hash/ABC123';
+
+describe('purgeCache — best-effort Cloudflare purge', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('awaits the CF call and resolves on success', async () => {
+    mockCfPurge.mockResolvedValue({ success: true });
+
+    await expect(purgeCache({ urls: [URL] })).resolves.toBeUndefined();
+
+    expect(mockCfPurge).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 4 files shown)


CivitaiDevOpsAgent:

PR #3327, opened 2026-07-23
feat(model): 'Set as Minor' moderator quick action + lockedProperties enforcement fix

manuelurenah:
## Summary

Adds a moderator **"Set as Minor"** quick action to the model three-dot menu, on both the model detail page and model cards. ClickUp [868k53v5k](https://app.clickup.com/t/868k53v5k) Part 1.

Today, flagging a model as depicting a minor costs a moderator a full round trip through the model edit form: open `/models/[id]/edit`, tick "Intended to depict a minor character", save. Illustrated-minor flags spike to 40+/day, so this is a real bottleneck. One menu click now produces the same end state.

The action is a toggle (`Set as Minor` / `Unset as Minor`) — reversibility from the same menu is the misclick guard, so there's no confirmation dialog.

**Set** writes `minor = true`, `nsfw = false`, `sfwOnly = true`, forces `gallerySettings.level` to SFW, and adds `minor`/`nsfw`/`sfwOnly` to `Model.lockedProperties` so the uploader can't revert it.

**Unset** writes `minor = false` and removes those three locks. It deliberately leaves `sfwOnly`, `nsfw`, and the gallery level alone — the model may have been legitimately SFW-only before it was ever flagged, and guessing wrong silently re-opens NSFW generation on a model nobody asked to re-open.

## ⚠️ Also fixes a moderation-enforcement bypass

The lock this feature depends on was advisory. Two service functions enforced `lockedProperties` against the **client's** request body rather than the stored row:

- **`upsertModel`** — a creator posting `lockedProperties: []` gave the strip-loop nothing to iterate, so their `minor: false` landed.
- **`privateModelFromTraining`** — worse. It never stripped `data.lockedProperties` before spreading into the update, so posting `[]` **permanently overwrote `Model.lockedProperties` with `[]`**, destroying the lock. Reachable by any member with a subscription on a model they own. Its first enforcement block also deleted keys from `input` *before* destructuring, so `lockedProperties: ['user']` nulled out the session user.

Both now enforce `union(stored, client)` — client claims are additive, never subtractive — and drop `lockedProperties` from non-moderator input entirely, via one shared `enforceLockedProperties` helper so a third copy can't drift.

Two related holes closed along the way:

- The profanity filter's `if (evaluation.shouldMarkNSFW && !data.nsfw)` guard read `data.nsfw` *after* lock enforcement deleted it, so it fired unconditionally and wrote `nsfw = true` — producing `minor: true, nsfw: true`, the exact state `upsertModelHandler` rejects. Now gated on the stored locks. Detection still records `meta.profanityMatches` either way; only the enforcement is gated, so moderators don't lose the signal on the highest-risk content class.
- The image-propagation `UPDATE ... WHERE id IN (...)` emitted one bind parameter per image against Postgres's 65,535 cap. Model 618692 has 211,652 images under its versions' posts. It threw *after* the model row was already flagged, leaving the flag written, images unpropagated, and no audit row. Replaced with a set-based `UPDATE "Image" i ... FROM "Post" p ... RETURNING i.id`, which also closes a `dbRead`→`dbWrite` race.

## Changes

**Server**
- `applyModelFlagSideEffects` extracted from `upsertModel` (pure refactor) — model tag cache + search index, gallery-settings Redis bust, `ingestModel` for Published/Scheduled, image `minor`/`poi` propagation + image search index, and now a `bustMvCache` so the generation resource picker doesn't serve stale flags for up to an hour.
- `setModelMinor` service fn + `model.setMinor` moderator procedure.
- `'setMinor'`/`'unsetMinor'` ModActivity strings — two, not one, because `trackModActivity` upserts `ON CONFLICT (entityType, activity, entityId) DO UPDATE` and a single string would lose the record on toggle-off. The audit write is attempted before the fan-out and can't block it.

**Client**
- `ToggleMinorModel` render-prop component (mirrors `ToggleLockModel`), wired into the detail page three-dot menu and `ModelCardContextMenu`.

## No migration

`Model.minor`, `Model.sfwOnly`, and `Model.lockedProperties` all exist. `ModelMetric.minor` / `ModelBaseModelMetric.minor` — what the feed actually filters on — are maintained by `trg_sync_model_to_metric_update` / `trg_sync_model_to_base_model_metric`, both of which list `minor` in their `WHEN (... IS DISTINCT FROM ...)` clause, so a plain `model.update` propagates in-transaction.

## Testing

- 3 new test files, 56 tests: `model-flag-side-effects.service.test.ts`, `set-model-minor.service.test.ts`, `model-locked-properties.service.test.ts`
- `pnpm vitest run src/server/services/__tests__/` — 1534 passed, 0 failed
- `pnpm run typecheck` — clean

Not browser-verified — the menu action itself is worth a manual pass by a moderator on preview.

## Follow-ups (not in this PR)

- `article.service.ts:772` and `bounty.service.ts:433` carry the same client-supplied lock pattern and neither strips `lockedProperties` before the write.
- `model.updateGallerySettings` / `copyGallerySettings` let an owner raise the gallery browsing level back after a flag. Low blast radius (the images themselves are hard-set `minor = true`), but it's the one moderator effect that doesn't stick.
- `'category'` and `'tags'` in the edit form's `lockableProperties` are unenforceable server-side — the payload field is `tagsOnModels`, destructured out of `data` before enforcement, and there's no `category` key at all.
- `requestReviewHandler`/`declineReviewHandler` call `upsertModel` without `isModerator` despite being moderator-only. Not a hole today (their select omits `lockedProperties`), but latent.
- `ToggleLockModel.tsx` has the same exact-key `setData` no-op this PR fixed in `ToggleMinorModel`.
- `imagesSearchIndex.queueUpdate` now receives ~211K ids in one `sAdd` on the largest gallery. Should work, but worth chunking.

## Part 2

The ClickUp task's Part 2 (hashing set-as-minor models and auto-flagging re-uploads) is **not** in this PR.


CivitaiDevOpsAgent:

PR #3333, opened 2026-07-24
fix(vault): stop OOM-and-infinite-retry loop in process-vault-items job

ZacxDev:
## Problem

The `process-vault-items` background job (builds the downloadable per-vault-item zip archives) could exhaust the process memory limit and then get stuck retrying the same item forever:

1. **Unbounded batch** — `dbWrite.vaultItem.findMany(...)` had no `take`, so every run pulled the *entire* eligible backlog at once.
2. **Unbounded in-memory buffering** — an uncapped `Promise.all` over a gallery's images fetched and buffered *every* image fully into memory simultaneously, then `zip.generateAsync({ type: 'blob' })` materialized the whole zip in memory as well (input buffers + output blob co-resident, ~2× the image bytes). A single large gallery could push memory past the limit.
3. **Failure counter unreachable on hard kill** — the `meta.failures` / `MAX_FAILURES` safety valve lived inside the `catch` block. A hard out-of-memory kill terminates the process without running `catch`, so `meta.failures` never incremented. The offending item stayed eligible in the query and was re-attempted on **every** run — an infinite memory-blowup loop.

## Fix

1. **Bounded batch** — added `take: 50` to the query (matches the sibling `clear-vault-items` job's batch size), so each run processes a bounded number of items.
2. **Capped download concurrency** — replaced the uncapped `Promise.all` over images with the repo's existing `limitConcurrency` helper, capped at 5 concurrent downloads/buffers. Only a bounded slice of a gallery is held in memory at a time. (No new dependency — reuses `~/server/utils/concurrency-helpers`.)
3. **Kill-resilient failure accounting** — the failure attempt is now recorded in the DB **before** the heavy download+zip work begins. An uncatchable kill can no longer produce an infinite retry: a repeatedly-failing item advances its failure count each run and eventually trips `MAX_FAILURES`, dropping out of eligibility. A successful run rolls the optimistic increment back, and the normal (catchable-error) path re-asserts the same count rather than incrementing again — so existing one-failure-per-failed-run success/failure semantics are unchanged.

Streaming zip generation/upload would be the largest memory win, but the current presigned single-`PUT` upload path makes it a substantially larger change — deferred as a follow-up.

## Tests

Extracted `processVaultItem` + `getEligibleVaultItemsQuery` so the batching and failure-accounting ordering are unit-testable. Added `src/server/jobs/__tests__/process-vault-items.test.ts` (vitest) covering:

- the run is bounded (`take === VAULT_ITEMS_BATCH_SIZE`);
- eligibility excludes items whose failure count exceeds the retry budget;
- the failure increment is **persisted before** the heavy download/zip work (asserted via mock call ordering), and treats missing/null meta as 0;
- a successful run rolls the increment back (no net failure counted);
- a catchable error counts exactly one failure (no double increment) and records `latestError`;
- a repeatedly-killing item climbs past `MAX_FAILURES` and is then excluded.

All 7 tests pass. Full `tsc --noEmit` shows no new type errors introduced by this change.

### DIFF
--- a/src/server/jobs/__tests__/process-vault-items.test.ts
+++ b/src/server/jobs/__tests__/process-vault-items.test.ts
@@ -1,0 +1,195 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const {
+  mockDbWrite,
+  mockGetModelVersionData,
+  mockGetPdf,
+  mockFetchBlob,
+  mockGetCustomPutUrl,
+  mockGetS3Client,
+  mockProcessedInc,
+  mockFailedInc,
+} = vi.hoisted(() => ({
+  mockDbWrite: {
+    vaultItem: { findMany: vi.fn(), update: vi.fn() },
+  },
+  mockGetModelVersionData: vi.fn(),
+  mockGetPdf: vi.fn(),
+  mockFetchBlob: vi.fn(),
+  mockGetCustomPutUrl: vi.fn(),
+  mockGetS3Client: vi.fn(),
+  mockProcessedInc: vi.fn(),
+  mockFailedInc: vi.fn(),
+}));
+
+vi.mock('@prisma/client', () => ({ Prisma: { AnyNull: Symbol('AnyNull') } }));
+vi.mock('~/shared/utils/prisma/enums', () => ({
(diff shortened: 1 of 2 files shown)


CivitaiDevOpsAgent:

PR #3334, opened 2026-07-24
fix(security): escape embedding triggerWord before RegExp — kill ReDoS event-loop freeze

ZacxDev:
## The bug (confirmed HIGH, measured)
`applyResources()` in `src/server/services/orchestrator/comfy/comfy.utils.ts` interpolated a **creator-supplied embedding `triggerWord` UNESCAPED** into a RegExp:
```js
const regex = new RegExp(`\\b${resource.triggerWord}\\b`, 'gi');
```
That is both a **regex injection** and a **catastrophic-backtracking ReDoS**. A trigger like `(x+x+)+y` compiles to `\b(x+x+)+y\b`, which backtracks exponentially on a crafted generation prompt.

**Measured, verbatim regex (prompt = literal trigger + an x-run):**

| x-run | time |
|---|---|
| n=28 | 0.9 s |
| n=30 | 3.5 s |
| **n=32** | **14.2 s** |

Exponential (each +2 chars ≈ ×4) — so **prompt-length caps do not help**. Reachable on the **authed self-serve generation-submit path** (runs on the dp-prod web event loop); the `value.includes(triggerWord)` gate is trivially satisfied because the attacker controls the prompt too. This is a sibling of the recently-fixed `og-metadata` O(n²) freeze, but worse (exponential vs quadratic).

## The fix
Regex-escape `triggerWord` before building the RegExp — it's a **literal token**, never a pattern. Uses the repo's standard escape (`.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')`, same as `blocklist.service.ts:124`). The `.includes` substring gate keeps using the raw literal (correct).

## Test (mutation-verified)
New `__tests__/comfy.utils.test.ts` (drives the embedding branch via a `parseAIR` mock):
- adversarial metachar `triggerWord` completes **instantly** (was ~14 s) — the ReDoS guard,
- a normal trigger still resolves to `embedding:<air>`,
- metacharacters matched **literally** (`a.b` does not match `axb`) — the injection guard.
**Mutation-checked:** the metachar test *fails* against the un-escaped code, so it genuinely guards the bug.

Found by the event-loop-freeze audit sweep across the codebase. `tsc` clean for the changed files.

### DIFF
--- a/src/server/services/orchestrator/comfy/__tests__/comfy.utils.test.ts
+++ b/src/server/services/orchestrator/comfy/__tests__/comfy.utils.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from 'vitest';
+
+// Drive the embedding branch of applyResources deterministically. The vuln under
+// test is the triggerWord→RegExp construction inside that branch, not AIR parsing;
+// the real Air.parse needs a live-format URN, so we force the type here. The
+// functional assertions below only pass if the branch actually ran.
+vi.mock('~/utils/string-helpers', async (importActual) => {
+  const actual = await importActual<typeof import('~/utils/string-helpers')>();
+  return {
+    ...actual,
+    parseAIR: () => ({ type: 'embedding', source: 'civitai', model: 1234, version: 5678 }),
+  };
+});
+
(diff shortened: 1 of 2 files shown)


CivitaiDevOpsAgent:

PR #3335, opened 2026-07-24
fix(jobs): bound ingest-images per-run batch so it can't flood the image scanner

ZacxDev:
## Problem

The `ingest-images` retry/backfill cron (`src/server/jobs/image-ingestion.ts`, `*/5 * * * *`) pulled a hardcoded **`take: 10000`** rows from `JobQueue` (type `ImageScan`) every run and submitted every eligible one to the media-rating scanner via `ingestImage`.

The scanner sustains only a limited throughput (on the order of ~1k submissions per 5-min run, ~6–7k / 30 min). When the queue is large, each run dumps roughly **5x** what the scanner can absorb → bulk rating jobs **expire before they run** → scan workflows fail → images flip to `Error` → the cron **re-queues those errored images** (retry amplification) → a self-sustaining congestion collapse where the failure rate *is* the load. This caused a multi-hour image-scan outage.

Key architecture fact: **new user uploads scan directly via `ingestImage` on image creation** (`image.service.ts`), independent of this cron. So this cron is purely the **retry/backfill** path — bounding it does **not** delay new-upload scanning, it only paces retries/backfill.

## Fix

Replace the hardcoded `take: 10000` with a scanner-matched, env-configurable cap:

- New env var **`IMAGE_SCANNING_MAX_PER_RUN`** (`src/env/server-schema.ts`, `z.coerce.number().default(1000)`) — conservative default, tunable without a redeploy.
- The cron now pulls at most `IMAGE_SCANNING_MAX_PER_RUN` rows.

Because **every** downstream step — the Pending/Rescan/Error filtering, the submission fan-out, and the `JobQueue` cleanup — is derived only from the rows pulled at the top of the run, capping the pull caps per-run submissions **and** leaves the rows we did *not* pull this run untouched in the queue. `orderBy createdAt asc` keeps the drain **oldest-first**, so a backlog drains gradually across runs instead of in one thundering-herd dump. That gradualness is itself the post-outage catch-up guard: a resumed cron with an accrued backlog bleeds it off at a rate the scanner can sustain rather than re-flooding it.

Preserved unchanged: the Pending/Rescan/Error filtering, the retry cooldowns (`IMAGE_SCANNING_RETRY_DELAY`, the 60-min error delay), the `retryCount < 9` cap, the backfill/low-priority split, and the processed→removed / waiting-for-retry→kept `JobQueue` cleanup semantics (rows we didn't process this run simply stay queued).

## Before / after

```diff
-    take: 10000,
+    take: env.IMAGE_SCANNING_MAX_PER_RUN, // default 1000
```

## Operational note

The cron is currently paused externally to stop the live flood; this change is what makes it safe to re-enable. On re-enable, the accrued backlog drains gradually (≤ the cap per run) instead of as a single dump.

## Test

Adds `src/server/jobs/__tests__/ingest-images-cap.test.ts`, which drives the real job with a simulated backlog (500) much larger than the cap (100) and asserts:
- the queue is pulled with `take = IMAGE_SCANNING_MAX_PER_RUN`, oldest-first;
- the number of scan submissions is bounded by the cap (100, not 500).

## Out of scope

Scheduler-side cron de-duplication is a separate change. The direct-upload `ingestImage` path and the webhook handler are untouched.

### DIFF
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -370,6 +370,12 @@ export const serverSchema = z
     TEXT_MODERATION_CALLBACK: z.string().optional(),
     IMAGE_SCANNING_MODEL: z.string().optional(),
     IMAGE_SCANNING_RETRY_DELAY: z.coerce.number().default(5),
+    // Upper bound on how many queued images the `ingest-images` retry/backfill cron
+    // pulls (and therefore can submit) per run. Sized to the scanner's sustainable
+    // throughput so a large backlog drains gradually across runs instead of in one
+    // dump. New user uploads scan directly via ingestImage on creation and are NOT
+    // gated by this. Conservative default; tune via env without a redeploy.
+    IMAGE_SCANNING_MAX_PER_RUN: z.coerce.number().default(1000),
     IMAGE_SCANNER_NEW: zc.booleanString.default(false),
     DELIVERY_WORKER_ENDPOINT: z.string().optional(),
     DELIVERY_WORKER_TOKEN: z.string().optional(),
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:

PR #3336, opened 2026-07-24
fix(articles): honor moderator NSFW override across all ingestion recompute paths

JustMaier:
## The bug

An article with a moderator NSFW override set (`Article.moderatorNsfwLevel` NOT NULL) still gets its `ingestion` dragged to `Error` when an **embedded content image** fails scanning. Concretely: article 32939 embeds two animated-webp badge images that the image scanner errors on (animated webp is unscannable). The article aggregates its content images' ingestion, so it flips to `ingestion='Error'` — hiding it from feeds / marking it broken — even though a moderator override (`moderatorNsfwLevel=1`, `lockedProperties` includes `userNsfwLevel`) is in force and should make the scan result moot.

## Root cause / the gap I fixed

`recomputeArticleIngestionInTx` is the **single choke point** every scan/webhook/edit/reconcile recompute funnels through:
- image-scan webhooks and the `article-ingestion-reconcile` cron → `updateArticleImageScanStatus` → `recomputeArticleIngestionInTx`
- text-moderation webhooks + upsert → `recomputeArticleIngestion` → `recomputeArticleIngestionInTx`

It is also the **only** writer of `Article.ingestion = Error/Blocked` — there is **no bypass path**. PR #3314 added the override handling here, but its final pass narrowed the override to substitute for **`Pending` only**, so `imageError`/`textError` still won over the override. That is the exact path that dragged the override-set article to `Error`.

## The fix

Extract the state-machine decision into a pure, unit-tested `deriveArticleIngestionState` helper (`src/server/services/article-ingestion.helpers.ts`) and widen the override precedence. New precedence (highest first):

1. **Blocked** — a policy-blocked image/text wins even over an override. A block (incl. pHash/CSAM) is a moderation signal, not a rating question. `ingestion=Blocked` is the sole guard keeping a blocked image out of the feed + search index (`updateArticleNsfwLevels` derives level from `ingestion='Scanned'` images only and COALESCEs the override over it, so a blocked image's level never reaches `nsfwLevel`).
2. **Override → Scanned** — now supersedes both `Pending` **and** `Error`. An unscannable content image can no longer hide an overridden article.
3. **Error** — unresolvable scan/text failure, no override.
4. **Scanned** — everything terminal + clean.
5. **Pending** — still awaiting a callback.

### ⚠️ Deliberate scope decision — please confirm
The task named `Error/Blocked/Processing`. I made the override supersede **Error, Pending, and Processing (transient/technical states)** but **kept `Blocked` winning**, because #3314's adversarial review explicitly established that an override is "a rating decision, not a moderation bypass" and `ingestion=Blocked` is the only thing keeping a policy-blocked (potentially CSAM) image out of the feed and search index. Making the override supersede `Blocked` would reintroduce exactly the content-safety regression #3314's second pass reverted. The reported failure (unscannable animated webp) produces `Error`, never `Blocked`, so this fully fixes the reported bug. **If you do want `Blocked` overridden too, say so and I'll change one line** — but I'd advise against it.

## How to test
- Unit: `pnpm exec vitest run src/server/services/__tests__/article-ingestion.helpers.test.ts` (11 cases). Proves: override + content-image Error → `Scanned`; override + Pending → `Scanned`; override + Blocked → **still `Blocked`**; no-override + Error → `Error` (unchanged); no-override defaults unchanged.
- `pnpm typecheck` clean on touched files (only pre-existing `kysely` module-resolution noise in `packages/civitai-db*` remains, which is environmental worktree noise).

## Files changed
- `src/server/services/article-ingestion.helpers.ts` (new) — pure `deriveArticleIngestionState`
- `src/server/services/article.service.ts` — choke point now calls the helper
- `src/server/services/__tests__/article-ingestion.helpers.test.ts` (new)

### DIFF
--- a/src/server/services/__tests__/article-ingestion.helpers.test.ts
+++ b/src/server/services/__tests__/article-ingestion.helpers.test.ts
@@ -1,0 +1,107 @@
+import { describe, expect, it } from 'vitest';
+import type { ArticleIngestionInputs } from '~/server/services/article-ingestion.helpers';
+import { deriveArticleIngestionState } from '~/server/services/article-ingestion.helpers';
+import { ArticleIngestionStatus } from '~/shared/utils/prisma/enums';
+
+// A fully-scanned, clean, no-override article. Each test overrides only the
+// fields relevant to the case it asserts.
+const base: ArticleIngestionInputs = {
+  imageBlocked: false,
+  imageError: false,
+  imageDone: true,
+  textBlocked: false,
+  textError: false,
+  textDone: true,
+  hasModeratorOverride: false,
+};
+
+describe('deriveArticleIngestionState', () => {
+  describe('moderator NSFW override in force', () => {
+    it('resolves Scanned when a content image errored (the unscannable animated-webp bug)', () => {
+      expect(
+        deriveArticleIngestionState({
+          ...base,
+          hasModeratorOverride: true,
+          imageError: true,
+          imageDone: true,
+        })
+      ).toBe(ArticleIngestionStatus.Scanned);
+    });
+
+    it('resolves Scanned while a content image is still Pending (not yet done)', () => {
+      expect(
+        deriveArticleIngestionState({
+          ...base,
+          hasModeratorOverride: true,
+          imageDone: false,
+        })
+      ).toBe(ArticleIngestionStatus.Scanned);
+    });
+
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:

PR #3337, opened 2026-07-24
fix(orchestrator): route RU blob fetches through Cloudflare proxy

JustMaier:
## Problem

Since ~2026-07-11, users on **Russian ISPs** can load the site and run generations, but **generated images/videos never display** — the browser's direct fetches to `orchestration-new.civitai.com/v2/consumer/blobs/…` time out with `net::ERR_TIMED_OUT`. Everything else works.

**Root cause** (support diagnosis, Freshdesk 69212):
- `civitai.red` is Cloudflare-fronted → Russian ISP DPI routes around it (this is why the May 2026 routing change restored RU access).
- `orchestration-new.civitai.com` is a **bare origin IP** (not Cloudflare), so RU DPI filters it directly by IP/SNI.
- Net effect: the app loads and generation succeeds (rides the `.red` edge), but the browser fetches the result blobs straight from the un-fronted orchestration host — and *those* fetches are what get blocked. Host is globally healthy; only RU is affected.

## Fix

Rewrite orchestrator URLs in tRPC responses to the **Cloudflare-fronted proxy** `orchestration-proxy.civitai.com` for RU requests, so the browser fetches ride the CF edge and bypass the DPI block. The proxy is a transparent CF route to the same service, so the presign `sig`/`exp` query is preserved and stays valid.

- **Detection** reuses the existing `CF-IPCountry` header (`getRegion()` in `region-blocking.ts`) — same signal behind the May `.red` fix. No new detection code.
- Applied as a **post-response middleware** (`regionProxyMiddleware`) on `orchestratorProcedure` + `orchestratorGuardedProcedure`, covering all user-facing generation traffic: feed (`queryGeneratedImages`), `getWorkflow`, `queryWorkflowsByTags`, `generateFromGraph`, `whatIfFromGraph`, `statusUpdate`, `patch`, etc. Also attached to the user-facing **comics** procedures (a blocked comic panel returns a raw orchestrator `blurredPreviewUrl` that would otherwise stay DPI-blocked in RU).
- **Non-RU requests and the flag-off path early-return with zero traversal cost** — the recursive URL walk only runs for RU with the flag on.
- **Cache-safe**: the per-user service cache is Redis (out-of-process); the rewrite runs on the freshly-deserialized in-memory copy, so cached payloads stay region-agnostic. Rewrite is idempotent.

## Feature flag / kill-switch

Gated behind a new `ruOrchestratorProxy` flag (`availability: []` → **fails closed / off**; the `ru-orchestrator-proxy` Flipt key is the on-switch **and** the instant runtime kill-switch).

- Flipt flag PR (enabled): **civitai/flipt-state#53**. It's a no-op until this deploys — the flag key is only read by this code — so it can merge any time.
- Flip it **off** to instantly disable the rewrite fleet-wide if the proxy misbehaves, no deploy.

## Scope notes

- **`pollIterationStatus` / `iterateGenerate` untouched** — they download from the orchestrator server-side (US, unblocked) and return **S3/CDN keys**, not orchestrator URLs, so they're not browser-direct paths (verified in review).
- **Mod-only endpoints** (`queryUserGeneratedImages`, `getWorkflowForModeration`) are not covered — there are no RU moderators.
- **No migration.**

## Adversarial review

Reviewed by a sub-agent tasked to refute correctness/safety. No Critical/High issues in the core change. It confirmed the sig-preserving swap, ruled out cache poisoning (Redis, out-of-process) and a SignalR leak (that path resolves URLs via `statusUpdate`, which is covered), and verified the `pollIterationStatus` S3-key claim. The **Medium** finding (comics `blurredPreviewUrl` bypass) and two **Low** hardening notes (bare-string payload, recursion depth guard) are all addressed in this PR.

## Testing

- `pnpm typecheck` clean.
- 8 unit tests in `region-proxy.test.ts` (host swap, sig preservation, proxy-host no-double-rewrite, non-orchestrator passthrough, malformed input, deep nested walk, bare-string payload, primitive/null passthrough) — all pass.

## Tickets

- CU-868kdkv93 — Infra: front orchestration blob host with Cloudflare (RU DPI block)
- CU-868ke4d0f — RU point to new Orch domain

### DIFF
--- a/src/server/orchestrator/__tests__/region-proxy.test.ts
+++ b/src/server/orchestrator/__tests__/region-proxy.test.ts
@@ -1,8 +1,8 @@
 import { describe, expect, it } from 'vitest';
 import {
   ORCHESTRATOR_PROXY_HOST,
-  deepRewriteOrchestratorUrls,
   rewriteOrchestratorUrlToProxy,
+  rewriteOrchestratorUrlsDeep,
 } from '~/server/orchestrator/region-proxy';
 
 describe('rewriteOrchestratorUrlToProxy', () => {
@@ -35,7 +35,7 @@ describe('rewriteOrchestratorUrlToProxy', () => {
   });
 });
 
-describe('deepRewriteOrchestratorUrls', () => {
+describe('rewriteOrchestratorUrlsDeep', () => {
   it('rewrites orchestrator URLs nested in arrays and objects, in place', () => {
     const payload = {
       items: [
@@ -48,7 +48,8 @@ describe('deepRewriteOrchestratorUrls', () => {
       ],
       video: 'https://orchestration-new.civitai.com/v2/consumer/blobs/v.mp4?sig=3',
     };
-    deepRewriteOrchestratorUrls(payload);
+    const returned = rewriteOrchestratorUrlsDeep(payload);
+    expect(returned).toBe(payload);
     expect(payload.items[0].url).toBe(
       `https://${ORCHESTRATOR_PROXY_HOST}/v2/consumer/blobs/a.jpeg?sig=1`
     );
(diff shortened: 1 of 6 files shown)


CivitaiDevOpsAgent:

PR #3340, opened 2026-07-24
feat(users): opt-in "remove all content" on user ban (block, not delete)

JustMaier:
## What changed

Adds an opt-in **"remove all content"** capability to the user-ban flow (ClickUp T1).

- `toggleBan` takes an optional `removeContent?: boolean`. When banning, it blocks all of the user's images via a single `image.updateMany({ where: { userId } })` → `ingestion=Blocked`, `nsfwLevel=Blocked`, `blockedFor`. Wrapped in try/catch → Axiom (`ban-user-remove-content`) so a failure can't break the ban.
- **Default logic**: content is blocked when `removeContent === true` OR (`reasonCode === SexualMinor && removeContent !== false`). So SexualMinor bans default ON (mod can opt out with `false`); every other reason is OFF unless explicitly enabled.
- **`blockedFor`**: `BlockedReason.CSAM` for SexualMinor bans (mirrors the existing `softDeleteUser` CSAM path), `BlockedReason.Moderated` for all other reasons.
- Threaded `removeContent` through `toggleBanUserSchema`, the tRPC `toggleBanHandler` (spreads `...input`), and the Retool `POST /api/mod/ban-user` webhook (`removeContent: z.coerce.boolean().optional()`).

Model unpublishing is unchanged — it already runs via `bulkUnpublishModelsForBannedUser`.

## Block, not delete (rationale)

We **block** images rather than delete them so the existing **7-day appeal window** is preserved. The existing `remove-blocked-images` job hard-deletes blocked images after 7 days, so blocking gives the same end state with an appeal grace period and no new deletion code. This mirrors the CSAM `softDeleteUser` mechanism exactly.

Bind-var safety: `updateMany` with a `where: { userId }` filter is a single UPDATE statement, not an `IN` list, so it does not hit the 32767 parameter cap regardless of how many images the user has (see PR #3330). No id-fetch / chunking needed.

## Preview-count contract (for T2 frontend)

New moderator-only query:

```ts
trpc.user.getBanContentPreview.useQuery({ userId: number }) // → { imageCount: number }
```

`imageCount` = images where `ingestion != 'Blocked'` (already-blocked images excluded — they'd be no-ops). Use it to show how many images the ban would block, and default the checkbox ON when reason is SexualMinor.

## Follow-up (not in this PR)

Unban does **not** restore blocked content. Intentionally out of scope for this PR. If we later want unban to reinstate content, it should re-set `ingestion`/`nsfwLevel`/`blockedFor` for images blocked by the ban — but note the 7-day job means anything past the appeal window is already gone, so a full restore is only meaningful within that window.

### DIFF
--- a/src/pages/api/mod/ban-user.ts
+++ b/src/pages/api/mod/ban-user.ts
@@ -11,7 +11,10 @@ const schema = z.object({
   reasonCode: z.enum(BanReasonCode).optional(),
   detailsExternal: z.string().optional(),
   detailsInternal: z.string().optional(),
-  removeContent: z.coerce.boolean().optional(),
+  removeContent: z
+    .enum(['true', 'false'])
+    .optional()
+    .transform((v) => (v === undefined ? undefined : v === 'true')),
 });
 
 export default WebhookEndpoint(async (req: NextApiRequest, res: NextApiResponse) => {
--- a/src/pages/api/mod/ban-user.ts
+++ b/src/pages/api/mod/ban-user.ts
@@ -11,10 +11,13 @@ const schema = z.object({
   reasonCode: z.enum(BanReasonCode).optional(),
   detailsExternal: z.string().optional(),
   detailsInternal: z.string().optional(),
+  removeContent: z.coerce.boolean().optional(),
 });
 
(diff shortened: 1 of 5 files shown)


CivitaiDevOpsAgent:

PR #3343, opened 2026-07-24
fix(blocks): reject ReDoS-prone manifest patterns + bound regex input

ZacxDev:
## Summary

Reworks the App Blocks settings-`pattern` ReDoS defense after an adversarial audit found the first approach (`safe-regex` in the install-time meta-schema) was **misplaced** and **over-aggressive**. The gate now lives at the real submission boundary, uses an accurate ReDoS analyzer, and no longer widens a silent fail-open.

## What was wrong with the first approach

- **`safe-regex` false-positives.** Its coarse star-height heuristic rejects common **linear** patterns that are not ReDoS — the canonical slug `^[a-z0-9]+(-[a-z0-9]+)*$`, decimals `^\d{1,4}(\.\d{1,2})?$`, snake_case `^[a-z0-9]+(_[a-z0-9]+)*$`. Worse, the rejection was **silent**: `getInstallConfig` fell back to `{settings:{}}`, so the app's settings form rendered **zero fields with no error**.
- **Wrong layer.** `manifestSettingsSchema` is consumed only at install/display/subscription time, **not at manifest submission**. The real submission validator (`BlockManifestValidator.validate`, invoked from the git-push webhook) had **no** settings-field checks — so a bad-pattern manifest was still stored and mod-approvable, with security resting on a fragile implicit fail-open.
- **Fail-open widened.** Install paths do `parsed.success ? validateBlockSettings(...) : input.settings`. Making the meta-schema reject more patterns means more existing manifests fail `safeParse` → **all** field validation (type/enum/range) is silently skipped and raw input stored.

## Reworked design

1. **Submission gate.** New async `BlockManifestValidator.validateSubmission` = the synchronous shape/security `validate` **plus** an accurate ReDoS + input-bound check on every string-setting `pattern`. It is wired into **all four** submission paths: the git-push webhook, the developer manifest API, `blocks.updateManifest`, and the publish-request approve flow. This makes *"a stored/approved manifest ⇒ its setting patterns are non-exponential and input-bounded"* an **enforced invariant**, and gives the developer a clear error at submit time. The dev-facing message is e.g. `settings.<key>: pattern is vulnerable to ReDoS (catastrophic backtracking) — remove nested or overlapping unbounded quantifiers (e.g. "(a+)+") or simplify the expression`, and `settings.<key>: a field declaring "pattern" must also declare "max_length" (<= 1000)`.

2. **Accurate analysis via `recheck`** (replaces `safe-regex`). `recheck` does real automaton + fuzzing analysis and classifies a regex as safe / vulnerable (exponential **or** polynomial) / unknown. Empirically verified:
   - **PASS (safe):** slug `^[a-z0-9]+(-[a-z0-9]+)*$`, decimal `^\d{1,4}(\.\d{1,2})?$`, snake `^[a-z0-9]+(_[a-z0-9]+)*$`.
   - **REJECT (vulnerable):** `(a+)+$`, `(x+x+)+y`, `([a-zA-Z]+)*$`.
   - Note: `(.*)*` classifies as **safe/linear** — which is correct (a bare `(.*)*` matches everything on the first greedy pass, so there's no failing suffix to force exponential backtracking). That accuracy is the whole point of `recheck` over `safe-regex`.
   - `recheck` is forced to its portable **in-process `pure` engine** (its default backend prefers a prebuilt native binary that can't run on some hosts and there hangs), and lives in a **server-only** module reached via dynamic `import()` — so it never enters the client bundle and `validate` stays synchronous/client-safe.

3. **Fail-open fixed.** `manifest-settings.meta.schema.ts` reverts to **compile-check only**. Since the submission gate now guarantees stored patterns are non-exponential, the runtime `.test()` is safe and existing manifests (including slug patterns) `safeParse` successfully again — so field validation actually runs instead of being silently skipped.

4. **Eval-site input cap kept.** `settings-validator.service.ts` still hard-caps `raw` to ≤1000 before `.test()` — defense-in-depth with no false-positive risk, covering any pattern stored before this gate existed.

5. **Lockfile.** `safe-regex` / `@types/safe-regex` removed, `recheck` added; `pnpm-lock.yaml` regenerated with `pnpm@10.28.1` and passes `--frozen-lockfile`.

## Migration safety

The submission gate applies to **new / re-submitted** manifests only — existing installs are **not** retroactively hard-failed. The removed fail-open (field validation now runs) plus the kept eval-site input cap cover the runtime for any legacy pattern.

## Tests

- `settings-pattern-guard.test.ts` (new): `recheck` classification PASS/REJECT lists; `collectSettingsPatternErrors` (ReDoS reject, `pattern`-without-`max_length` reject, over-cap `max_length` reject, non-compiling reject, safe-pattern accept, multi-field).
- `block-manifest-validator.service.test.ts`: `validateSubmission` accepts slug/decimal/snake, rejects an exponential pattern with a dev-facing error, rejects `pattern` without `max_length`, and merges base + settings-pattern errors.
- `manifest-settings.meta.schema.test.ts`: locks in **compile-check-only** (F3 guard) — accepts patterns without `max_length` and accepts an exponential pattern (submission gate's job), still rejects a non-compiling pattern; the obsolete `safe-regex`/`max_length`-required cases from v1 are removed.
- `settings-validator.service.test.ts`: eval-site cap unchanged (bounded-time); adds an F3 regression asserting the slug-pattern manifest parses and its field validation is **enforced** (no fail-open).
- Caller test mocks updated to expose `validateSubmission`.

`tsc --noEmit` (heap-bumped): zero errors in changed files. All affected unit suites green.


CivitaiDevOpsAgent:

PR #3344, opened 2026-07-24
fix(apps): prefer page <title> over host name in external submit autofill (+ de-hyphenate host fallback)

ZacxDev:
## What

Two flaws in the external app-listing **submit wizard** name autofill (`src/components/Apps/ExternalSubmitForm.tsx` + `offsiteSubmitFormConfig.ts`), driven by a real bug report.

**Bug report:** For `https://cosmetic-studio.civitai.com` the page's real `<title>` is **"Civitai Cosmetic Studio"** (extracted server-side by the OG-meta fetch, `fetchListingMetaFromUrl` → `data.name`), but the form showed the uglier host-derived **"Cosmetic-Studio"**.

### Fix 1 — prefer the extracted `<title>` over the host-derived name
Previously `applyNormalizedUrl` (on URL-advance) set the `name` from the host **up front**, before the OG-meta fetch resolved — so the meta-apply effect (also guarded on `name` empty) could never fill the better `<title>`.

Now:
- `applyNormalizedUrl` sets the **slug** from the host immediately (kept — slugs need hyphens and there's no better source), but **no longer sets `name`**. It stashes the host-derived name in a `hostNameFallbackRef`.
- The meta effect runs once the fetch **settles** (success **or** error) for the current URL, and fills `name` from `data.name` (the `<title>`). It uses the host-derived name **only as a fallback** when the fetch yields no usable title.
- Every field stays **fill-if-empty**, so a name the user has typed is never clobbered (both the title and the host fallback pass through the same `v.name.trim().length === 0` guard).

The Name field lives on a later wizard step (Details) than the App URL step, so the fetch has time to resolve first.

### Fix 2 — de-hyphenate the host fallback name
`deriveListingFromUrl` joined name words with a hyphen (`base.split('-').map(capitalize).join('-')` → `"Cosmetic-Studio"`). It now joins with a **space** → `"Cosmetic Studio"`. The **slug** derivation stays hyphenated/unchanged.

### Scope
- CREATE submit wizard only. The **edit** form (`ExternalListingEditForm`) does not host-derive a name (it fills only from `data.name`), so it has no host-preempts-title bug — left unchanged.
- Description autofill (from `og:description`, fill-if-empty, truncated) and the cover/icon suggestion behavior are **unchanged**.

## Tests
- **`deriveListingFromUrl` unit** (`__tests__/deriveListingFromUrl.test.ts`): `cosmetic-studio.civitai.com` → name **"Cosmetic Studio"** (space) / slug **"cosmetic-studio"**; updated the other multi-word host cases (`My App`, `Cool Tool`, `Super App`, IDN `Xn Mnchen 3ya`); single-label hosts unaffected.
- **Form browser tests** (`ExternalSubmitForm.browser.test.tsx`): title wins over host (`"Civitai Cosmetic Studio"`); host de-hyphenated fallback when the meta returns **no title**; fallback also applies when the fetch **errors**; a **user-typed name is never overwritten** by either source. Two existing submit tests updated to settle the meta so the fallback fills the name.

### Verbatim results (honest)
- `deriveListingFromUrl.test.ts` (unit): **31 passed**.
- `ExternalSubmitForm.browser.test.tsx` (component/chromium): **13 passed**.
- Sibling suites `ExternalSubmitForm.reducedMotion` + `ExternalSubmitForm.edit` (component): **11 passed**.
- Related config suites `offsiteSubmitFormConfig` + `offsiteSubmitFormConfig.oauth-fields` + `normalizeLinkUrl` (unit): **57 passed**.
- `tsc --noEmit` (heap-bumped): **0 errors in changed files**; 16 remaining are the pre-existing env errors only (`kysely`, `event-engine-common`, `image.service.ts` implicit-any) — unrelated to this change.

### DIFF
--- a/src/components/Apps/ExternalSubmitForm.browser.test.tsx
+++ b/src/components/Apps/ExternalSubmitForm.browser.test.tsx
@@ -23,7 +23,12 @@ const READONLY_SCOPES = 4 | 32; // ModelsRead | MediaRead
 const mocks = vi.hoisted(() => ({
   submit: vi.fn(),
   mutate: vi.fn(),
-  meta: { data: undefined as unknown, isFetching: false, isSuccess: false },
+  meta: { data: undefined as unknown, isFetching: false, isSuccess: false } as {
+    data: unknown;
+    isFetching: boolean;
+    isSuccess: boolean;
+    isError?: boolean;
+  },
   clients: {
     data: [{ id: 'oauth-client-1', name: 'My OAuth App', allowedScopes: 4 | 32 }] as unknown,
     isLoading: false,
@@ -173,6 +178,9 @@ describe('ExternalSubmitForm — redesigned wizard', () => {
       data: [{ id: 'oauth-client-1', name: 'My OAuth App', allowedScopes: 0 }],
       isLoading: false,
     };
+    // Meta settles with no title → the Name autofills from the host fallback so the
+    // Details step is complete and "Create draft" is enabled.
+    mocks.meta = { data: {}, isFetching: false, isSuccess: true };
     renderWithProviders(<ExternalSubmitForm />);
     await advanceFromUrl();
     await pickClient();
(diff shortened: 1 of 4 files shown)


CivitaiDevOpsAgent:

PR #3345, opened 2026-07-24
feat(challenge): Prometheus instrumentation for the Challenge funnel + economy + state

ZacxDev:
## What

Adds Prometheus instrumentation for the **Challenge** feature — funnel, Buzz economy, and live state — as additive telemetry only (no behavior change). All metrics live in a dedicated `src/server/prom/challenge.metrics.ts` module of never-throw `recordX()` helpers; emit sites call one helper each. A metrics failure can never break challenge business logic (every helper try/catch-swallows), mirroring `b2-put.metrics.ts`.

### A. Funnel counters
| Metric | Labels | Emit site |
|---|---|---|
| `civitai_app_challenge_created_total` | `source`, `buzzType` | `upsertUserChallenge`, create branch only (after the initial `scanUserChallenge`) |
| `civitai_app_challenge_scan_result_total` | `source`, `result` | moderation adapter (blocked/scanned) + `applyFailure` (error) + `scanUserChallenge` catch (error) |
| `civitai_app_challenge_entry_submitted_total` | `source`, `buzzType`, `paid` | `chargeContestEntryFeesForCollection` (single chokepoint — see below) |
| `civitai_app_challenge_review_requested_total` | `source` | `requestReview` on success |
| `civitai_app_challenge_completed_total` | `source` | `endChallengeAndPickWinners` both success returns (winners + zero-winner) |
| `civitai_app_challenge_voided_total` | `source`, `reason` | `voidChallenge` `voided:true` branch |
| `civitai_app_challenge_deleted_total` | `source` | `deleteUserChallenge` after `deleteChallenge` succeeds |

### B. Economy counters (`.inc(buzzAmount)`)
| Metric | Labels | Emit site |
|---|---|---|
| `civitai_app_challenge_entry_fees_buzz_total` | `source`, `buzzType` | `chargeEntryFees` — first-time house+pool legs only (retries/conflicts excluded, no double-count) |
| `civitai_app_challenge_prize_paid_buzz_total` | `source`, `buzzType` | `endChallengeAndPickWinners` after the winner payout (winner prizes only; entry-participation blue prizes excluded) |
| `civitai_app_challenge_operation_spent_buzz_total` | `source`, `buzzType` | inside `incrementOperationSpent` (guarded cheap PK read for labels) |
| `civitai_app_challenge_refund_buzz_total` | `source`, `buzzType`, `reason` | inside `refundUserChallengeFunds` (sum of `totalRefunded` across the entry-fee + initial-prize prefixes) |
| `civitai_app_challenge_refund_failures_total` | `source`, `reason` | `refundUserChallengeFunds` catch (a real refund throw; the swallowed NOT_FOUND no-op is not counted) |

### C. State gauges (async `collect()`, memoized ~20s)
| Metric | Labels | Query |
|---|---|---|
| `civitai_app_challenge_by_status` | `source`, `status` | `GROUP BY source, status` over `Challenge` |
| `civitai_app_challenge_ingestion_pending` | `source` | `WHERE ingestion='Pending' GROUP BY source` |
| `civitai_app_challenge_completing_stuck` | `source` | `WHERE status='Completing' AND updatedAt < now() - 30 min GROUP BY source` |
| `civitai_app_challenge_operation_budget_used_ratio` | `source` | `SUM(operationSpent)::float / NULLIF(SUM(operationBudget),0)` over `status IN ('Active','Completing')` |

## Cardinality

Every label value is normalized against a fixed allowed set **inside** the record helpers; unrecognized values map to `unknown` (or `other` for reasons). No raw/free-text ever reaches a label.
- `source` → System|Mod|User|unknown · `buzzType` → green|yellow|unknown · `status` → the 5 `ChallengeStatus` values|unknown · `paid` → `0`|`1`
- `result` → scanned|blocked|error (unexpected → error) · void `reason` → moderator|nsfw|activation|other · refund `reason` → void|delete|other

## Design decisions (per the review checklist)

- **`challenge_entry_submitted_total` chokepoint:** `chargeContestEntryFeesForCollection` (collection.service). Both entry paths — the non-deferred `validateContestCollectionEntry` charge and the deferred `bulkSaveItems` charge — flow through this one function, so counting here can't double-count paid vs free. Its challenge lookup was broadened (dropped the `entryFee > 0` filter) so **free-entry** User challenges are counted too; `chargeEntryFees` no-ops on `entryFee <= 0` (returning all images as paid), so the caller's commit behavior is unchanged. `paid='1'` when `entryFee > 0`, else `'0'`; count = `paidImageIds.length` (committable entries). Only `source=User` challenges have such a collection, so `source` is `User` in practice.
- **Scan-result emit-site deviation:** emitted from `challenge-moderation.adapter.ts` at the **callback verdict** level, not from `scanUserChallenge` (which doesn't yet know the outcome). A not-blocked callback is counted `scanned` — the dominant terminal state — even though a green-user-challenge NSFW escalation may subsequently void/hide it inside `applyChallengeNsfwEscalation`; that void is captured separately by `challenge_voided_total`, so the metric reflects the moderation callback verdict, not the final `ingestion` column. `blocked` and `error` (terminal `applyFailure`, only on a real Pending→Error transition) are emitted from the adapter; a distinct **submit-time** failure is also counted `error` in `scanUserChallenge`'s catch (a different failure event from the terminal `applyFailure`).
- **Gauge gating:** this repo exposes **no clean pod-role / jobs-pool signal** (only `PODNAME`, a bare pod name), so per "don't invent a fragile gate" the gauges **run on all pods behind a 20s in-process memo**, reading the **replica pool** (`pgDbRead`) with a 5s `statement_timeout` and last-good degradation. **Follow-up:** gate to a couple of pods if a role signal (e.g. an env `POD_ROLE`) is added.
- **Void `reason` closed set:** `moderator` (router action), `nsfw` (NSFW escalation), `activation` (grace-window void of an unscanned challenge past start), else `other`. Threaded as a telemetry-only optional arg on `voidChallenge` (default `other`); does not affect behavior.
- **Refund `reason`:** `void` / `delete` (per spec); the shared zero-winner completion refund passes `completion` which normalizes to `other`.

## Where the series land

The Grafana dashboard that consumes these lands **separately in `datapacket-talos`**. These series **won't exist until this deploys** via the `release` branch (civitai-prod → dp-prod).

## Tests

- New `src/server/prom/__tests__/challenge.metrics.test.ts` (18 tests): label normalization, never-throw (a throwing `counter.inc` is swallowed), counters/gauges emit expected series (gauge `collect()` fed via an injected cache that mocks the DB).
- Existing challenge suites kept green (void / delete-user / edit-rescan / upsert / moderation-adapter / nsfw-escalation / funding / contest-entry). A few assertions updated only for the new telemetry-only `reason` arg and the broadened entry-fee lookup.

Verified: targeted vitest run **137 passed / 0 failed** across 17 files; `tsc --noEmit` passes clean.

### DIFF
--- a/src/pages/api/metrics.ts
+++ b/src/pages/api/metrics.ts
@@ -3,6 +3,9 @@ import { EOL } from 'os';
 import client from 'prom-client';
 import { dbRead, dbWrite } from '~/server/db/client';
 import { instrumentationRegistry } from '~/server/prom/client';
+// Side-effect import: registers the challenge state gauges (collect()-based) on the default
+// registry so they are present + scraped even before the first challenge op runs this session.
+import '~/server/prom/challenge.metrics';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
 
 const labels: Record<string, string> = {};
--- a/src/server/games/daily-challenge/challenge-funding.ts
+++ b/src/server/games/daily-challenge/challenge-funding.ts
@@ -45,6 +45,11 @@ import {
 } from '~/shared/constants/challenge.constants';
 import { ChallengeSource, CollectionItemStatus } from '~/shared/utils/prisma/enums';
 import { createLogger } from '~/utils/logging';
+import {
+  recordChallengeEntryFeesBuzz,
+  recordChallengeRefundBuzz,
+  recordChallengeRefundFailure,
+} from '~/server/prom/challenge.metrics';
 
 const log = createLogger('challenge-funding', 'yellow');
 
@@ -209,6 +214,15 @@ export async function chargeEntryFees({
     }).catch(() => {});
   }
 
+  // Economy telemetry: Buzz newly charged this call (house + pool legs charged for the FIRST time —
+  // conflicts/retries already counted on the original charge, so this can't double-count). Entry
+  // fees only exist on source=User challenges.
+  recordChallengeEntryFeesBuzz({
+    source: ChallengeSource.User,
+    buzzType: fromAccountType,
+    amount: houseResult.transactions.length * houseAmount + newPoolCharges * poolAmount,
+  });
+
   return { paidImageIds, unpaidImageIds };
 }
 
@@ -249,17 +263,22 @@ async function refundChallengeFundsByPrefix(input: {
   externalTransactionIdPrefix: string;
   description: string;
   details: { challengeId: number };
-}): Promise<number> {
+}): Promise<{ count: number; amount: number }> {
   try {
-    const { refundedTransactions } = await refundMultiAccountTransaction(input);
-    return refundedTransactions.length;
+    const { refundedTransactions, totalRefunded } = await refundMultiAccountTransaction(input);
+    return { count: refundedTransactions.length, amount: totalRefunded };
   } catch (e) {
-    if (e instanceof TRPCError && e.code === 'NOT_FOUND') return 0;
+    if (e instanceof TRPCError && e.code === 'NOT_FOUND') return { count: 0, amount: 0 };
     throw e;
   }
 }
 
-export async function refundUserChallengeFunds(challengeId: number) {
+export async function refundUserChallengeFunds(
+  challengeId: number,
+  // Telemetry-only: what triggered the refund. Normalized to void|delete|other in the metric
+  // helper; does not affect refund behavior.
+  reason: 'void' | 'delete' | 'completion' | 'other' = 'other'
+) {
   const challenge = await dbRead.challenge.findUnique({
     where: { id: challengeId },
     select: {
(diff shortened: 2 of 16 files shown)


CivitaiDevOpsAgent:
### [PR-REVIEW] round 2 — 1 finding(s) (confidence ≥ 0.80)

- **medium** `src/server/games/daily-challenge/challenge-funding.test.ts:43` (correctness, conf 0.85) — This test dynamically imports the real challenge-funding.ts module, which now statically imports ~/server/prom/challenge.metrics (adding prom-client counter registrations and telemetry calls inside chargeEntryFees and refundUserChallengeFunds). The test does not mock the metrics module, so real prometheus counters are registered and incremented during test runs, causing CI pollution. If @civitai/telemetry/client or prom-client misbehaves in the test environment, all tests in this file break at import time.  
  ↳ _Add vi.mock('~/server/prom/challenge.metrics', () => ({ recordChallengeEntryFeesBuzz: vi.fn(), recordChallengeRefundBuzz: vi.fn(), recordChallengeRefundFailure: vi.fn() })) before the dynamic import of challenge-funding.ts._

PR #3352, opened 2026-07-24
fix(model-files): scope file-settings validation and detect int8/mxfp8

JustMaier:
## Reported symptoms

- **SubtleShader / MNeMiC**: "can't change a model file's precision to int8" — hitting Save on a file's settings fails and nothing persists.
- Related reports of file settings failing to save in other cases, sometimes surfacing as *"There are multiple files with the same type and size, please adjust your files"* for files that aren't actually duplicates.

## Root causes

**1. Per-file save validated the entire version.** `FileEditForm.handleSave` (`Files.tsx`) called `validationCheck()`, which parsed the *whole* `files` array against `metadataSchema`. Any other file in the version missing size / precision / quant type failed the parse, so the file you were actually editing could never be saved.

**2. The error-reporting path threw.** zod's `error.format()` returns a plain **object** keyed by failing index, not an array (verified against this repo's zod: `Array.isArray === false`, `typeof forEach === 'undefined'`). The result was cast to an array and iterated with `.forEach`, so a failed parse raised a `TypeError` instead of showing "Missing required fields" and returning false. Out of `handleSave`'s onClick that means the save silently does nothing — which matches the reports better than a visible validation error, and is very likely the actual mechanism users hit.

**3. The duplicate-file key collapsed absent fields.** `checkConflictingFiles` built its key with `.filter(Boolean)`, so two files that both lacked size/fp/quantType produced an identical key and registered as duplicates. The notification named no files, so it read as though the file being edited was rejected.

**4. safetensors precision auto-detect had no int8 mapping.** `safetensorsDtypeToFp` was a 4-branch allow-list (F16 / BF16 / F32-F64 / F8*) returning `null` otherwise. An int8 safetensors has dtype `I8`, so every tensor mapped to null and no `fp` was ever set (the caller only assigns `if (fp)`). The missing `fp` then trips the "Floating point is required for model files" refinement on the next save. `int8` and `mxfp8` are both live selectable options in prod (the dropdown lists are mod-managed at runtime), so the UI offered values the detector could never produce.

## Changes

- **`checkValidation(uuid?)`** — passing a uuid validates only that file, so a sibling that is still missing metadata can no longer block an unrelated edit. `handleSave` passes the edited file's uuid.
- **Duplicate detection is scoped, not dropped.** The scoped path still runs `getConflictingFiles(files)` and fails the save if the *edited* file is part of a conflicting group — you can't turn File A into a duplicate of File B — while a pair of metadata-less siblings you aren't touching no longer blocks you. (Worth knowing: the whole-array check in `startUpload` is reachable only while uploads are pending — `SaveButton` renders on `hasPending` — so on an already-uploaded version the scoped check is the one that actually runs. `modelFile.update` does no such validation server-side.)
- **Fixed the `format()` iteration** for both paths: errors are now rebuilt as a real array indexed by file position (`files.map((_, i) => formatted[i])`), so a failed whole-array parse reports "Missing required fields" instead of throwing.
- **Positional conflict key** (`join('|')` over `?? ''` placeholders) so absent fields can't alias distinct files. The key doesn't include the file name and component files require none of size/fp/quantType, so a group where *no* member has any of those (two bare `Text Encoder` files, say) is dropped rather than reported — there's nothing to disambiguate on. `checkConflictingFiles` became `getConflictingFiles`, returning the conflicting groups; the notification now names the offending files, matching the existing "missing required fields" message pattern.
- **`safetensorsDtypeToFp`** maps `I8` to `int8` and `F8_E8M0` to `mxfp8`, with the generic `F8*` to `fp8` fallback applied only after the lookup miss. `U8` is deliberately *not* mapped: bitsandbytes packs nf4/fp4 weights as `U8`, and `nf4` is itself a supported option, so guessing `int8` there would auto-fill a wrong value.
- **MXFP8 detection is presence-based.** Byte dominance alone can't find it — a real MXFP8 checkpoint stores weights as `F8_E4M3` with `F8_E8M0` shared-exponent scales at ~1 byte per 32 elements — so `F8_E8M0` tensors alongside fp8 weights now resolve to `mxfp8`.
- `fp8_scaled` / `fp8_mixed` / `nvfp4` are deliberately not inferred: they're scaling schemes, not dtypes, and aren't derivable from the safetensors header.

## Manual verification

1. **The main regression**: open a version that has one incomplete file (e.g. a Checkpoint `Model` file with no precision set) plus a second, complete file. Edit the *complete* file's settings and Save — it should now save. Before this change the save did nothing.
2. Set a file's precision to `int8` and Save — persists; reload the edit modal and confirm it stuck.
3. Drop a new int8 `.safetensors` — the Precision select should auto-fill `int8`. An MXFP8 checkpoint (F8_E4M3 weights + F8_E8M0 scales) should auto-fill `mxfp8`; ordinary fp8 variants still land on `fp8`; fp16/bf16/fp32 unchanged; an nf4 (bitsandbytes, U8) checkpoint should auto-fill nothing rather than a wrong value.
4. Editing a file so it becomes a true duplicate of another (same type, size, precision, format, quant) is still rejected, and the error now names both files.
5. Two bare component files of the **same** type (e.g. `clip_l.safetensors` and `t5xxl.safetensors` both typed `Text Encoder`, both with precision blank) no longer register as duplicates — toggling Required or changing the type on one saves fine. Set a precision on one of them and the pair stays non-conflicting; only genuinely-identical *complete* files are rejected.
6. With an upload still in flight, "Save Changes" on a version where some file is missing required metadata should show the per-file "Missing required fields" notification (previously this path threw and the button appeared to do nothing).

### DIFF
--- a/src/components/Resource/Files.tsx
+++ b/src/components/Resource/Files.tsx
@@ -968,7 +968,7 @@ function FileEditForm({
   }, [versionFile.id]);
 
   const handleSave = () => {
-    const valid = validationCheck();
+    const valid = validationCheck(versionFile.uuid);
     if (valid && versionFile.id) {
       mutate({
         id: versionFile.id,
--- a/src/components/Resource/FilesProvider.tsx
+++ b/src/components/Resource/FilesProvider.tsx
@@ -336,13 +336,12 @@ export function FilesProvider({ model, version, children }: FilesProviderProps) 
 
     const validation = metadataSchema.safeParse(toValidate);
     if (!validation.success) {
-      const formatted = validation.error.format() as unknown as Array<{
-        [k: string]: ZodErrorSchema;
-      }>;
+      // format() returns an object keyed by failing index, not an array.
+      const formatted = validation.error.format() as unknown as Record<number, SchemaError>;
       const errors =
         targetIndex >= 0
           ? files.map((_, i) => (i === targetIndex ? formatted[0] : undefined))
-          : formatted;
+          : files.map((_, i) => formatted[i]);
       setErrors(errors);
 
       const missingFields: string[] = [];
@@ -366,7 +365,17 @@ export function FilesProvider({ model, version, children }: FilesProviderProps) 
       return false;
     }
 
-    if (targetIndex >= 0) return true;
+    if (targetIndex >= 0) {
+      // Only conflicts the edited file is part of — an unrelated pair of siblings
+      // that happen to share a key isn't this save's problem.
+      const target = files[targetIndex];
+      const conflicts = getConflictingFiles(files).filter((group) => group.includes(target));
+      if (conflicts.length) {
+        showConflictNotification(conflicts);
+        return false;
+      }
+      return true;
+    }
 
     // External-generation versions (mod-only, routed via external engines) intentionally
     // ship without files; skip the component-count requirement for them.
@@ -403,17 +412,7 @@ export function FilesProvider({ model, version, children }: FilesProviderProps) 
 
     const conflicts = getConflictingFiles(files);
     if (conflicts.length) {
-      showErrorNotification({
-        title: 'Duplicate file types',
-        error: new Error(
-          conflicts
-            .map(
-              (group) =>
-                `${group.map((f) => f.name).join(', ')}: same type and settings, please adjust`
-            )
-            .join('\n')
-        ),
-      });
+      showConflictNotification(conflicts);
       return false;
     }
     return true;
@@ -817,6 +816,17 @@ export const getConflictingFiles = (files: FileFromContextProps[]) => {
   return [...groups.values()].filter((group) => group.length > 1);
 };
 
+const showConflictNotification = (conflicts: FileFromContextProps[][]) => {
+  showErrorNotification({
+    title: 'Duplicate file types',
+    error: new Error(
+      conflicts
+        .map((group) => `${group.map((f) => f.name).join(', ')}: same type and settings`)
+        .join('\n')
+    ),
+  });
+};
+
 /** Model types whose primary file is an archive/config rather than model weights */
 const archivePrimaryModelTypes: ModelType[] = [
   ModelType.Workflows,
@@ -852,7 +862,6 @@ function inferFileType(fileName: string, modelType?: ModelType | null): ModelFil
       return undefined;
   }
 }
-
 
 /**
  * Pick a default type for a file dropped into the Additional Components section so
--- a/src/components/Resource/FilesProvider.tsx
+++ b/src/components/Resource/FilesProvider.tsx
@@ -35,6 +35,7 @@ type SchemaError = {
   format?: ZodErrorSchema;
   quantType?: ZodErrorSchema;
 };
+type FileErrors = Array<SchemaError | undefined>;
 
 // Both link mutations (the Meili picker and linkOfficialFileByHash) return the
 // same server shape; map it to a client LinkedComponent in one place.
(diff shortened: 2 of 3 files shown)


CivitaiDevOpsAgent:

JustMaier:
Review fixes pushed in `540bd62` (typecheck clean, prettier clean). PR body rewritten — two claims in it were wrong and are now corrected.

**D2 (ship gate) — fixed.** Confirmed independently against this repo's zod before touching it: `error.format()` on an array schema returns a plain object keyed by the *failing* index (`keys: ['1','_errors']`, `Array.isArray === false`, `typeof forEach === 'undefined'`). So the whole-array branch threw a `TypeError` out of `checkValidation` rather than returning false — swallowed by `FilesEditModal`'s `.catch(() => ({}))`. Errors are now rebuilt as a real array in both branches:

```ts
const errors =
  targetIndex >= 0
    ? files.map((_, i) => (i === targetIndex ? formatted[0] : undefined))
    : files.map((_, i) => formatted[i]);
```

Agreed this is very likely the real mechanism behind "errors and nothing persists" — the PR body now leads with it as root cause #2 instead of the old narrative.

**D4 — fixed.** You're right that "deferred to upload/publish" was false: `SaveButton` is gated on `hasPending`, so on a fully-uploaded version the whole-array path never runs, and `modelFile.update` validates nothing server-side. The scoped path now runs `getConflictingFiles(files)` and only reports groups containing the edited file — you still can't turn File A into a duplicate of File B, but a metadata-less sibling pair you aren't touching can't block you. Shared notification helper so both paths report identically.

**D6 — fixed.** Dropped `U8 -> int8`; bitsandbytes packs nf4/fp4 as `U8` and `nf4` is a real supported option, so that would have auto-filled a wrong value. Detecting nothing is the right outcome there.

**D5 — implemented rather than removed.** It was a clean ~3 lines: `inferSafetensorsPrecision` now keeps its byte-dominance pick, then upgrades `fp8` to `mxfp8` when any `F8_E8M0` tensor is present, since the shared-exponent scales are ~1 byte per 32 elements and never win on bytes. The `bytesByFp` map is keyed by string with a single cast at the return, which also removed the cast from `safetensorsDtypeToFp`.

**Skipped as instructed:** D3 (component-count bypass via the Required switch) and D7 (stale sibling error highlighting after a successful scoped save) — both yours to track.

Scoping fix left exactly as reviewed: a sibling missing metadata never blocks an unrelated edit.

JustMaier:
N1 fixed in `83ce815` (typecheck clean, prettier clean). Good catch — the scoped conflict check did reintroduce a narrow version of the blocker the check was originally dropped to avoid.

`getConflictingFiles` now discards any group where no member has `size`, `fp`, or `quantType` set:

```ts
const hasDistinguishingSettings = (file: FileFromContextProps) =>
  !!(file.size || file.fp || file.quantType);

return [...groups.values()].filter(
  (group) => group.length > 1 && group.some(hasDistinguishingSettings)
);
```

Two bare `Text Encoder` files (`clip_l` + `t5xxl`, precision blank) no longer collide, so toggling Required or changing a type on one saves. Set a precision on either and they still don't conflict — they'd have to be genuinely identical *and* complete to be rejected. Truthiness rather than `!= null` because empty string isn't a valid value in any of those unions, and it avoids an `eqeqeq` argument. The fix lands in `getConflictingFiles` itself, so the scoped and whole-array paths both get it.

PR body corrected in two places: manual-verification step 5 now describes the same-type case accurately (the old wording was only true when the types differed), and the conflict-key bullet states the name-omission caveat.

Untouched as instructed: D3, D7, the D1 residual (whole-array check unreachable from Manage Files with nothing pending), and the auto-upload-on-drop path that persists with no conflict check.

On the two snags flagged as unverifiable: both compile clean. `Record<number, SchemaError>` is only an assertion over `format()`'s output and indexes fine with a numeric index; the string-keyed `bytesByFp` needs no cast internally, with the single `as ModelFileFp | null` at the return of `inferSafetensorsPrecision`.

PR #3353, opened 2026-07-24
fix(vault): harden process-vault-items batch against per-item failures + overlapping runs

ZacxDev:
Follow-up to #3333 addressing two robustness gaps in the vault-items batch job (`src/server/jobs/process-vault-items.ts`) surfaced while reviewing that change.

## A) One item's transient failure could abort the whole batch

The per-attempt failure-count bookkeeping update runs *before/outside* `processVaultItem`'s own `try/catch`. If that single update-by-primary-key throws (e.g. a transient DB blip), the error propagated out of the batch `for` loop and every remaining item in the run was skipped. The pre-#3333 code caught per-item errors and continued.

Fix: wrap each item's processing in the loop so one item's failure is logged and the loop continues to the next item. The kill-resilient accounting semantics from #3333 are preserved unchanged — the pre-attempt increment still persists (so an uncatchable kill still counts against the retry budget), a successful run still rolls it back, and a caught error still re-asserts the same count.

## B) Overlapping runs could duplicate heavy work and lose bookkeeping

The job is scheduled on a fixed interval, but a single run can occasionally take longer than that interval (a bounded batch where each item may fetch several images, each with a multi-minute timeout). When a new run starts before the previous one finishes, both runs do the heavy download+zip work on the same item and last-writer-wins on the item's `meta` JSON blob — losing failure-count updates and duplicating exactly the work the batch cap exists to bound.

Fix (lightweight, no schema migration): stamp an advisory in-flight lease timestamp (epoch millis) into the existing `meta` field when a run claims an item, and exclude freshly-leased items from the eligibility query. A lease older than a staleness window — i.e. a run that was killed mid-item and could never clear its lease — ages out, and the item becomes eligible again. The lease is cleared on both terminal paths (success and caught failure). This uses only the existing `meta` JSON field and the same numeric JSON-path filtering already used for the failure counter, so there is no enum/schema change and the eligibility query is the only caller touched.

## Tests

`src/server/jobs/__tests__/process-vault-items.test.ts`:
- A per-item processing throw does not abort the batch — remaining items are still processed (single-throw and mid-batch-throw variants).
- A freshly-leased (in-flight) item is excluded from eligibility while a stale or unleased item stays eligible; the lease cutoff tracks `now - staleness window`.
- The lease is stamped on claim and cleared on both the success and caught-failure terminal writes.

All existing #3333 accounting tests remain green. `tsc --noEmit` is clean with respect to the touched files.

### DIFF
--- a/src/server/jobs/__tests__/process-vault-items.test.ts
+++ b/src/server/jobs/__tests__/process-vault-items.test.ts
@@ -72,10 +72,20 @@ vi.mock('jszip', () => ({
 
 import {
   processVaultItem,
+  processVaultItems,
   getEligibleVaultItemsQuery,
   MAX_FAILURES,
   VAULT_ITEMS_BATCH_SIZE,
+  LEASE_STALENESS_MS,
 } from '~/server/jobs/process-vault-items';
+
+// The eligibility WHERE ANDs two OR-groups: [0] = retry-budget, [1] = overlap
+// lease guard. Small helpers keep the assertions robust to ordering.
+const budgetOr = (q: any) =>
+  q.where.AND.find((g: any) => g.OR?.some((b: any) => b.meta?.path?.[0] === 'failures')).OR;
+const leaseOr = (q: any) =>
+  q.where.AND.find((g: any) => g.OR?.some((b: any) => b.meta?.path?.[0] === 'processingStartedAt'))
+    .OR;
 
 const makeItem = (overrides: Record<string, unknown> = {}) => ({
   id: 1,
@@ -111,7 +121,7 @@ describe('getEligibleVaultItemsQuery — bounded batch + retry-budget exclusion'
 
   it('only selects items whose failure count is within the retry budget', () => {
     const q = getEligibleVaultItemsQuery();
-    const lteBranch = q.where.OR.find((b: any) => typeof b.meta?.lte === 'number');
+    const lteBranch = budgetOr(q).find((b: any) => typeof b.meta?.lte === 'number');
     expect(lteBranch?.meta?.path).toEqual(['failures']);
     expect(lteBranch?.meta?.lte).toBe(MAX_FAILURES);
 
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:

ZacxDev:
Addressed the two 🟡 audit findings in a follow-up commit (a4dd0c8):

**🟡-1 — bound per-item runtime so it cannot outlive the lease.** The S3 upload PUTs (details PDF + images zip + cover) had no timeout, so a hung upstream made a single item's runtime effectively unbounded (per-attempt and across `withRetries`) and able to outlive the 30-min lease → a second run could re-claim a still-running item (double-process / lost failure-count). Each PUT is now bounded with `AbortSignal.timeout(VAULT_UPLOAD_TIMEOUT_MS = 60s)` (a timeout aborts the attempt and `withRetries` retries). With uploads bounded, the worst-case per-item runtime is computable and documented in code: 10-image downloads (2 waves × 5 min) + pdf/zip gen (~2 min) + parallel uploads (4 attempts × 60s = 4 min) ≈ **16 min**. `LEASE_STALENESS_MS` raised **30m → 45m** (~2.8× margin, 16 min << 45 min) so a legit-slow item never ages out mid-run. New test asserts every upload PUT carries an `AbortSignal` and that the staleness window keeps a >2× margin over the worst case.

**🟡-2 — document the JSON-path null semantics the design hinges on.** Added a code comment at the lease `OR` branch in `getEligibleVaultItemsQuery` explaining that `path + equals: Prisma.AnyNull` matches rows where the key is **absent** inside a non-null blob (so never-leased and legacy items stay eligible and the backlog is never stranded), citing the identical prod pattern in `process-subscriptions-requiring-renewal.ts` (`renewalEmailSent` + `AnyNull`), and warning against a future refactor narrowing it to explicit-null-only. Real-DB integration coverage is **deferred**: this repo has no Postgres-backed / test-container integration harness in CI (the suite is fully unit-mocked), so the precedent comment + the existing query-shape test are the deliverable rather than a flaky/unrunnable test.

tsc clean re: the changed files; the job's vitest suite passes (15/15, +2 new).

PR #3354, opened 2026-07-24
fix(challenges): gate challenge feed on the challenge's own nsfwLevel

manuelurenah:
Two community-challenge bugs.

## 1. Feed filtered on the cover image's level, not the challenge's

The content-level gate only tested the cover image:

```sql
EXISTS (SELECT 1 FROM "Image" i WHERE i.id = c."coverImageId" AND (i."nsfwLevel" & $level) <> 0)
```

`Challenge.nsfwLevel` was never in the WHERE clause, so a challenge rated above the viewer's browsing level surfaced whenever its cover happened to be SFW — an X-rated challenge with a PG cover was visible to a PG/PG-13 viewer.

The client-side pass (`useApplyHiddenPreferences`, `case 'challenges'`) had the same hole from the other side: it tested `allowedNsfwLevel`, which is a **multi-bit mask** (`PG|PG13|R`), so it intersects a PG viewer's level for any challenge that merely also allows PG.

Now both gate on `Challenge.nsfwLevel` — the single highest level `allowedNsfwLevel` permits (`deriveChallengeNsfwLevel` = `Flags.maxValue`). The cover predicate is kept **alongside** it (ANDed) so a challenge that under-declares its rating still can't leak an NSFW cover into an SFW feed — `ChallengeCard` feeds `challenge.nsfwLevel` (not the cover's) to `ImageGuard2`, so the card wouldn't blur it.

Applied to `getInfiniteChallenges`, `getMyChallenges`, and `getCompletedChallengesWithWinners`.

## 2. Detail-page cover hidden entirely instead of blurred

Not a gating bug — a layout one. The cover container is `relative mx-auto max-w-2xl`; `mx-auto` cancels the parent `Stack`'s cross-axis stretch, making the box fit-content. The safe branch renders `EdgeMedia2` (an `` with intrinsic width) so it's fine, but the blurred branch's only content is `MediaHash`'s `absolute inset-0` canvas → fit-content resolves to **0**, and `aspect-[4/3] w-full` collapses to 0 × 0. The overlay and canvas were in the DOM the whole time, just zero-sized.

`w-full` on the container fixes it. Articles avoid this by wrapping the cover in `<AspectRatio>`.

## Verification

Reproduced and verified in the browser against the local dev DB, using a test challenge cloned to match challenge 422's shape (`nsfwLevel=8` X, `allowedNsfwLevel=15`, PG cover, non-owner creator):

| browsingLevel | old predicate | new predicate | feed result |
|---|---|---|---|
| 1 (PG) | visible | hidden | hidden |
| 7 (PG/PG-13/R) | **visible** | hidden | hidden |
| 15 (incl. X) | visible | visible | shown |
| control, PG challenge | visible | visible | shown |

Detail page with blur on and an R cover: renders 672 × 504 with the blurhash + "This image is rated R / Show" instead of collapsing to 0 × 0. Safe path unchanged (672 × 504, `` present).

New `challenge-feed-browsing-level.service.test.ts` (challenge + cover predicate present, anonymous parity, green cap clamping, no-level ⇒ no predicate). Challenge visibility suites: 49 passed, 0 failed.

The test challenge was deleted from the dev DB afterwards; no migrations.

### DIFF
--- a/src/components/HiddenPreferences/useApplyHiddenPreferences.ts
+++ b/src/components/HiddenPreferences/useApplyHiddenPreferences.ts
@@ -716,8 +716,10 @@ export function filterPreferences<
         const isOwner = challenge.createdById === currentUser?.id;
         if (isOwner || isModerator) return true;
 
-        // Content allowed by the challenge must intersect the user's browsing level
-        if (!Flags.intersects(challenge.allowedNsfwLevel, browsingLevel)) {
+        // The challenge's own rating — the highest level its `allowedNsfwLevel` permits — must
+        // intersect the user's browsing level. Intersecting `allowedNsfwLevel` itself is too loose:
+        // it's a multi-bit mask, so an R challenge that also allows PG would pass for a PG viewer.
+        if (!Flags.intersects(challenge.nsfwLevel, browsingLevel)) {
           hidden.browsingLevel++;
           return false;
         }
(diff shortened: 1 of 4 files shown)


CivitaiDevOpsAgent:

PR #3356, opened 2026-07-24
feat(model-files): sync precision and quant type lists

JustMaier:
## What & why

`precisions` and `quantTypes` are **DB-managed at runtime** — they live in the `KeyValue.modelFileOptions` row and are edited by mods via `/api/admin/model-file-options` without a deploy (see `docs/features/model-file-options.md`). The arrays in `src/server/common/constants.ts` are only the **fallback** used when that row is absent or unreadable.

The two sides had drifted. This PR syncs the **code** side. **It is code-only — no DB was touched.** Bringing the live row in line still needs an admin `PUT` (payload at the bottom — note **PUT**, not POST; POST appends and would leave the order wrong).

## Live bug this fixes

`int8` and `mxfp8` are already in the prod DB row, so uploaders can select them — but they were missing from `fpQualityRank` in `src/server/utils/model-helpers.ts`. Any value missing from that map ranks **0**, so those two files sorted *below* `nf4` in download-variant ordering. Both are now ranked correctly (`mxfp8` above all FP8 variants, `int8` just below plain `fp8`).

## Precisions

Full list, best to worst:

`fp32, fp16, bf16, mxfp8, fp8_mixed, fp8_scaled, fp8, int8, nf4, nvfp4, int4`

**Naming: underscores, not spaces** (`fp8_scaled`, not `fp8 scaled`). These values ride in the download querystring and are exact-string-matched against `metadata.fp` in `src/server/services/file.service.ts`; a space would have to survive URL encoding/decoding round-trips intact on every client, which is an avoidable hazard.

`nvfp4` and `nf4` are **different formats** — NVIDIA Blackwell FP4 with per-block FP8 scales vs. bitsandbytes NormalFloat4 — and are separate values, not aliases.

## Quant types

Added: `Q4_K_XL` / `Q3_K_XL` / `Q2_K_XL` (Unsloth-style dynamic), `IQ4_KS`, `IQ3_M`, `IQ3_S`, and `TQ2_0` / `TQ1_0` (ternary, ranked below `IQ1_M`). The deprecated `Q4_0_4_4` / `Q4_0_4_8` / `Q4_0_8_8` were deliberately **not** added.

The `_XL` dynamic quants are larger and higher-quality than the `_L` variant of the same bit width, so within a bit width the ranking is `XL > L > M > S` — `Q3_K_XL` above `Q3_K_L`, and `Q4_K_XL` / `Q2_K_XL` above their `_M` / base counterparts where no `_L` exists.

## `None` (unquantized GGUF)

There was previously no valid selection for an unquantized GGUF: the upload form requires a `quantType` for any `.gguf` file, and every option was a quantization. Rather than adding `F16`/`BF16`/`F32` as quant types — which duplicates the Precision field — `None` is now a quant type meaning "not quantized", and the user sets Precision separately.

- `None` ranks **above** every quantized value in `quantQualityRank` (unquantized is the highest quality), so an unquantized GGUF sorts first in its group.
- The `.gguf`-requires-quantType refinement stays as-is — `None` is now a valid answer to it.
- **Display choice:** `None` is never rendered literally. `getFileLabel` falls through to the precision (so the badge reads `fp16`, `bf16`, …); if no precision is set it reads **`Unquantized`**. `getFileDescription` reads **"Unquantized GGUF, fp16 precision"**. The `FileInfo` popover's "Quant Type" row goes through the same `formatQuantType` helper, so it reads "Unquantized" too. Rationale: the label sits where a quant tag normally goes, so printing "None" would read as "no information" rather than "this is the full-precision file".
- **Account preference:** `None` is excluded from "Preferred Quant Type". In a preference select it would read as "no preference" while actually storing "prefer unquantized GGUFs" and feeding `getPrimaryFile`; leaving the preference unset is how you express no preference.
- **Form:** the file editor and the merge-versions mapper previously showed Quant *or* Precision (`isGguf ? Quant : Precision`). They now show Precision alongside Quant when the quant type is `None`, which is what makes "set Precision separately" actually reachable. Switching quant back off `None` clears `fp`, so a hidden stale precision can't keep scoring in `getPrimaryFile` or keying `checkConflictingFiles`.
- **Validation:** a new refinement requires a precision when quant is `None`, so an unquantized GGUF can't be saved with *less* information than before the option existed. That refinement lives in `FilesProvider`'s `metadataSchema`, which the merge path doesn't use — the merge mapper reaches the same guarantee by resolving `fp` correctly (below) rather than by validating.

## Two things worth knowing about existing data

**Merge `fp` resolution.** `effectiveFp` in `MergeVersions` resolved through `??`, which treats the explicit `fp: null` written when quant moves off `None` as *absent* and falls back to the source file's original precision. On a GGUF source with `fp16` + a quant, cycling `None → Q8_0 → None` left the Precision field displaying `fp16` while the merge would have written `fp: null` — display diverging from what saves, and a `quantType: 'None'` file with no precision. Now resolved by key presence (`'fp' in mapping.metadata`) instead. **Known latent pattern:** `effectiveQuantType` / `effectiveSize` / `effectiveFormat` have the same `??` shape, but nothing writes an explicit `null` into them, so none is currently reachable. Left alone to keep this diff scoped.

**Prod has 1,199 GGUF files that already carry an `fp`** (only 4 also have a `quantType`). Those ~1,195 already fail the pre-existing `.gguf`-requires-quantType refinement, so any owner editing one is already forced to pick a quant — and doing so now silently clears their `fp`. That value is already invisible in the UI and `getFileDescription` ignores it, but it does feed `getPrimaryFile` scoring, so those files' download-preference scores shift on the next unrelated save. I believe that's correct — the `fp` was meaningless on a quantized GGUF — but it's a behavior change on existing rows, so it shouldn't be a surprise.

For the record, **0 rows in prod currently have `quantType = 'None'`**, so the new refinement blocks no existing data.
- Download variant matching is unaffected — `file.service.ts` does a plain equality check on `metadata.quantType`, and `'None'` matches like any other string.

## Files changed

- `src/server/common/constants.ts` — `modelFileFp`, `modelFileQuantTypes`
- `src/types/global.d.ts` — `ModelFileFp`, `ModelFileQuantType` unions
- `src/server/utils/model-helpers.ts` — `fpQualityRank`, `quantQualityRank`
- `src/utils/file-display-helpers.ts` — description lookups, `UNQUANTIZED_QUANT_TYPE` + `formatQuantType`
- `src/components/Resource/Files.tsx`, `src/components/Model/Actions/MergeVersions.tsx` — show Precision for `None`-quant GGUFs, clear `fp` when quant moves off `None`; `MergeVersions` also resolves `effectiveFp` by key presence rather than `??`
- `src/components/Resource/FilesProvider.tsx` — require a precision when quant is `None`
- `src/components/FileInfo/FileInfo.tsx` — "Unquantized" instead of "None"
- `src/components/Account/SettingsCard.tsx` — exclude `None` from the quant preference
- `docs/features/model-file-options.md` — refreshed counts + example order, documented the `None` sentinel and that array order is UI order

## Verification

`pnpm typecheck` clean. `pnpm eslint` on the changed files: 0 errors (one pre-existing unused-var warning in `model-helpers.ts`, unrelated; the change also cleared a pre-existing prettier error there). Prettier clean.

## DB payload (still needed)

Not applied by this PR. To bring the live row in line:

**Use `PUT`, not `POST`.** `POST` maps to `addModelFileOptions`, which preserves the existing order and appends new values to the tail — that yields the right *set* but the wrong *order*, leaving `fp32` stranded 4th and every new quant (including `None`) after `IQ1_M`. The dropdowns render the stored arrays verbatim, so order matters. `PUT` replaces wholesale.

```
PUT /api/admin/model-file-options?token=$WEBHOOK_TOKEN
{
  "precisions": ["fp32","fp16","bf16","mxfp8","fp8_mixed","fp8_scaled","fp8","int8","nf4","nvfp4","int4"],
  "quantTypes": ["None","Q8_0","Q6_K","Q5_K_M","Q5_K_S","Q5_1","Q5_0","Q4_K_XL","Q4_K_M","Q4_K_S","Q4_1","Q4_0","Q3_K_XL","Q3_K_L","Q3_K_M","Q3_K_S","Q2_K_XL","Q2_K","Q2_K_S","IQ4_XS","IQ4_KS","IQ4_NL","IQ3_M","IQ3_S","IQ3_XS","IQ3_XXS","IQ2_XS","IQ2_XXS","IQ2_S","IQ2_M","IQ1_S","IQ1_M","TQ2_0","TQ1_0"]
}
```


JustMaier:
Correction pushed (f037df9): `Q3_K_XL` now ranks **above** `Q3_K_L`, not below it.

The first pass placed each `_XL` "just above its `_M` counterpart", which for the 3-bit family landed `Q3_K_XL` (18) under `Q3_K_L` (19). That's inverted — Unsloth's dynamic `_XL` quants are larger and higher-quality than the `_L` variant of the same bit width, so the correct order within a bit width is `XL > L > M > S`.

Only the 3-bit family moved; `Q4_K_XL` and `Q2_K_XL` were already correct since neither family has an `_L`.

- `quantQualityRank`: `Q3_K_L` 18, `Q3_K_XL` 19 (swapped)
- `modelFileQuantTypes`: `Q3_K_XL` now precedes `Q3_K_L` so the dropdown stays best-first
- description string: `Q3_K_XL` now reads "better quality than Q3_K_L"

The DB payload in the PR description has been updated to match. `pnpm typecheck` clean, prettier clean.

CivitaiDevOpsAgent:

JustMaier:
Review follow-ups pushed (4beac1d). All five addressed.

**1. Rollout payload was self-contradicting — now `PUT`.** You're right: `POST` → `addModelFileOptions` → `dedupeOptions([...current, ...next])`, so it preserves existing order and appends. That gives the right set and the wrong order — `fp32` stranded 4th, every new quant including `None` after `IQ1_M`, in dropdowns that render the stored array verbatim. PR body now says `PUT` in both places (the intro line and the payload block), with a note explaining why `POST` is wrong here.

**2. `None` no longer renders literally anywhere.** Added `UNQUANTIZED_QUANT_TYPE` + `formatQuantType` to `file-display-helpers` and routed both surfaces through it:
- `FileInfo.tsx` — the popover row now reads "Quant Type: Unquantized".
- `SettingsCard.tsx` — `None` is filtered out of "Preferred Quant Type" entirely. Agreed it's a trap: in a preference select it reads as "no preference" but stores "prefer unquantized" and feeds `getPrimaryFile`. Leaving the preference unset is the existing way to say no preference.

The sentinel is now a named export rather than a scattered `'None'` literal, so the magic string appears once.

**3. `fp` is cleared when quant moves off `None`.** Both `Files.tsx` and `MergeVersions.tsx` quant `onChange` handlers now write `fp: null` when the new value isn't `None`. Confirmed the failure mode you described — the field vanished but `metadata.fp` persisted, invisible and un-editable, still scoring in `getPrimaryFile` and still keying `checkConflictingFiles`.

**4. Precision is now required when quant is `None`.** New refinement in `FilesProvider.tsx`: `quantType === 'None' → !!fp`, error "Floating point is required for unquantized files" on the `fp` path. Without it the entire justification for `None` was optional and an unquantized GGUF could save with less information than pre-PR.

**5. Docs.** Storage example now shows the current precision order; consumer refs de-line-numbered (they'd already drifted twice in this PR alone, so they now name the fields instead); added an "Array order is UI order" note covering the `PUT`-vs-`POST` trap; added a `### The None quant type` section documenting the sentinel — including in bold that **`DELETE`ing `None` from the DB list silently makes the Precision field unreachable for every GGUF with no error surfaced**.

**Skipped as instructed:** `GGUF_FTYPE_TO_QUANT` in `file-helpers.ts` — left the file completely untouched to avoid conflicting with #3352. I did not touch the doc comment either, since it sits directly above the map.

`pnpm typecheck` clean (full output). ESLint on all changed files: 0 errors, 3 warnings — all pre-existing and on untouched lines (`FilesProvider` 295/481 template-expression, `model-helpers` 64 unused `format`). Prettier clean.

JustMaier:
Fixed in e104ab1. Good catch — my fix #3 was the first code to write an explicit `null` into a `??`-resolved field, so I introduced the bug by using the clear the resolver couldn't represent.

`effectiveFp` now resolves by key presence:

```ts
mapping?.metadata && 'fp' in mapping.metadata
  ? mapping.metadata.fp
  : file.metadata?.fp ?? null
```

That closes both halves of what you described: the Precision field no longer displays a value the merge wouldn't save, and the `None`-without-`fp` state can no longer be produced on the merge path — which matters because my refinement lives in `FilesProvider`'s `metadataSchema` and merge doesn't use it.

Left `effectiveQuantType` / `effectiveSize` / `effectiveFormat` alone as instructed. They share the shape but nothing writes an explicit `null` into them, so none is reachable today; noted in the PR body as a known latent pattern so the next person to add a clear doesn't rediscover it the same way.

Also added to the PR body:
- The 1,199 GGUF-files-with-an-`fp` note. Agreed it's correct behavior — the `fp` was meaningless on a quantized GGUF and is already invisible — but it feeds `getPrimaryFile` scoring, so those files' download-preference scores shift on the next unrelated save, and that's worth stating rather than discovering.
- Your empirical result that 0 prod rows have `quantType = 'None'`, so the new refinement blocks no existing data.
- A line clarifying that the merge path reaches the None-requires-`fp` guarantee via correct resolution, not via the refinement, so the body no longer overclaims what fix #4 covers.

`pnpm typecheck` clean (full output). Prettier clean.

JustMaier:
Final review note, for the record.

The merge-path fix (`e104ab1`) closes the display/save divergence but not the underlying gap: the merge path still has no validation, so it can produce a `quantType: 'None'` file with no precision — now visibly (empty field) rather than deceptively. Such a file would later block the Files editor's array-wide `safeParse` for that whole version until someone sets a precision on it. Recoverable and clearly messaged, and prod currently has **0** `None`-quant files, so nothing is affected today.

So: the None-requires-precision refinement covers **uploads, not merges**. The merge path reaches the same guarantee by resolving correctly rather than by validating.

One behavior change worth knowing for whoever tests: after clearing precision, toggling a GGUF's quant back to `None` now shows an **empty** Precision field rather than re-displaying the source file's original value. That's correct — it's what actually saves — but it will read as the value having disappeared to anyone who remembers the old display.

PR #3357, opened 2026-07-24
feat(app-blocks): deep-link to a review finding (open tab + scroll + copy-link) [Phase 2.3 fast-follow]

ZacxDev:
## What

Deep-link-to-a-finding on the App Blocks agentic-review report (Phase 2.3 fast-follow). A moderator can now link to and share a specific finding:

- Landing on `/apps/review/<id>#finding-security-2` **activates the Security audit tab** and **scrolls that finding into view** with a brief, animation-free highlight.
- Each code/security finding gets a subtle one-click **"copy link"** affordance that copies the absolute deep link and reflects the anchor into the URL.

## Why

The report is tabbed (Summary | Code review | Security audit | Scopes) with one section visible at a time, so a moderator previously had no way to point a teammate at a specific finding. This makes findings addressable and shareable.

## How

- **Pure helpers** in `agentReviewReport.ts` (unit-tested): `findingAnchorId(tab, i)` and a total/defensive `parseReportHash()` that maps a URL hash to `{ tab?, anchorId? }` — tolerant of a missing `#`, unknown tabs, and malformed input (never throws).
- **Controlled tabs**: `ReportTabs` swaps `defaultValue` for controlled `value` + `onChange`, keeping `keepMounted`.
- **Hash sync WITHOUT next/router**: the renderer stays prop-only and router-agnostic — it reads `window.location.hash` in an effect, listens to `hashchange`, and writes the URL via `history.replaceState`. This is deliberate: `history.replaceState` updates the URL **without a router navigation**, so it does **not** trip the review page's route-leave navigation guard (a router `push`/`replace` would).
- **Scroll timing**: `keepMounted` panels are `display:none` until active, so a deep link activates the tab first, then scrolls the target card on the next `requestAnimationFrame` (once the panel is visible). Highlight clears after ~2s via a ref'd timeout (cleared on unmount).
- **Copy-link** uses Mantine `useClipboard` (precedent: `ApiKeyModal`) and builds the URL from `window.location` only.

### Tab / anchor scheme

- Tabs: `summary` | `code` | `security` | `scopes`.
- Anchors only on the two finding tabs: `finding-code-<i>` / `finding-security-<i>`, where `<i>` is the finding's index in the **severity-sorted render list** (the order the moderator sees), so a shared link lands on the same visible card. Summary/Scopes are not anchored.

## Safety

- All existing sanitization is intact: every finding string stays React-escaped, no `dangerouslySetInnerHTML`, no new markdown surface. The copy-link reads `window.location` only.
- SSR-safe: all `window`/`document`/`history` access is inside effects or event handlers, guarded for `typeof window`.
- **Dark / additive** — it renders inside the already-flag-gated `/apps/review/[publishRequestId]` page; no behavior change when the flag is off.

## Tests

- Node unit suite: added `findingAnchorId` + `parseReportHash` cases (valid tab+index, bare tab, missing `#`, unknown/malformed, negative/NaN index tolerated, round-trip). Suite green locally (`vitest run --project unit`).
- Browser (Vitest browser mode): `ReportTabs.browser.test.tsx` — hash `#finding-security-1` activates the Security tab, the target card carries `id="finding-security-1"`, and the copy-link affordance is present. Authored per convention; not run locally (no headless browser on the dev host).

## Verification honesty

Not verified in a real browser against the mod-gated review page (can't drive it headlessly). The DOM/interaction behavior is covered by the browser test (CI/pr-preview runs it) and still needs a human click-path check on the deployed preview. The **pr-preview typecheck and an adversarial `/audit-pr` are the real gates.**

### DIFF
--- a/src/components/Apps/ReportTabs.browser.test.tsx
+++ b/src/components/Apps/ReportTabs.browser.test.tsx
@@ -8,14 +8,18 @@ import { renderWithProviders } from '../../../test/component-setup';
  * (Phase 2.3 fast-follow) — browser-mode component tests.
  *
  * `ReportTabs` is prop-only + router-agnostic: it reads/writes the URL hash
- * directly (never next/router) so a deep link `#finding-<tab>-<index>` opens the
+ * directly (never next/router). Deep-linking is GATED on the route — active only
+ * on the dedicated review page (`/apps/review/<id>`), inert in the review modal
+ * (`/apps/review`). When active, a deep link `#finding-<tab>-<index>` opens the
  * right tab, scrolls the target card into view with a transient highlight, and
  * each finding card carries a subtle "copy link" affordance.
  *
  * Covers:
- *  - landing on `#finding-security-1` activates the Security audit tab;
- *  - the target finding card carries `id="finding-security-1"`;
- *  - each finding card renders a "copy link" ActionIcon.
+ *  - on the page route, landing on `#finding-security-1` activates the Security
+ *    audit tab, the card carries `id="finding-security-1"`, `scrollIntoView`
+ *    fires (deterministic post-commit scroll), and a copy-link renders;
+ *  - a bare tab hash (`#code`) activates that tab with no finding target;
+ *  - on the modal route there is NO deep-linking (no copy-link, no anchor id).
  *
  * The Summary tab renders `summaryMd` through CustomMarkdown, which reads
  * `useCurrentUser()`. We keep `summaryMd` null here so that surface is never hit,
@@ -28,6 +32,9 @@ vi.mock('~/hooks/useCurrentUser', () => ({
 }));
 
 const { ReportTabs } = await import('./ReportTabs');
+
+// The dedicated review PAGE route enables deep-linking (`/apps/review/<id>`).
+const PAGE_PATH = '/apps/review/pubreq_test';
 
 // Two security findings in severity order (critical, then high) → the second
 // card (index 1) is the `high` one, so `#finding-security-1` targets it.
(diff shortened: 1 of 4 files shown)


CivitaiDevOpsAgent:

ZacxDev:
Applied the two 🟡 items from the adversarial audit (verdict was safe-to-merge, no 🔴):

**1. Deterministic post-commit scroll.** The single `requestAnimationFrame` could fire before React committed the tab switch, so `scrollIntoView` hit a still-`display:none` `keepMounted` panel (no-op) while the highlight was set unconditionally. Reworked: `applyHash` stashes the pending anchor, and the scroll + highlight run in a `useEffect` keyed on the active tab — so it executes after the target panel is committed/visible. Highlight is now set only when the element is actually found. This also removes the rAF-after-unmount `setState` nit.

**2. Gate all deep-link behavior behind `deepLinkable`.** `ReportTabs` renders in both the dedicated review page (`/apps/review/<id>`) and the legacy review modal (`/apps/review`); hash read/write, per-finding anchors, and the copy-link only make sense on the page (in the modal a copy-link URL would reopen nothing). 

Gating approach: **route-derive** (`window.location.pathname.startsWith(/apps/review/)`), computed in a mount effect, rather than prop-drilling. The two surfaces share the chain `OnsiteReviewModalBody → AgentReviewPanel → ReportTabs` (3 shared components), so threading a prop would touch all of them; reading the actual route keeps it self-contained in `ReportTabs`. On the modal the component now behaves exactly as before the feature (no anchors / no copy-link / no hash writes), which also closes the stale-hash-bleed nit the audit noted.

Tests: browser test adds page-route deep-link + `scrollIntoView`-fires coverage and a modal-route case asserting no copy-link / no anchor ids. Node helper tests unchanged (20/20). `tsc` clean on edited files (the pre-existing project errors are unrelated).

Honest note: the DOM/interaction behavior is only covered by the report-only browser test — it still needs a human click-path check on the deployed preview (the mod-gated page cannot be driven headlessly here).

PR #3360, opened 2026-07-24
feat(articles): rescan cover+skip-locked fix, mod per-image scan override, class-aware scan UI

JustMaier:
## Summary
Fixes the article "scan status" flow that could leave `Article.ingestion = Error/Blocked` permanently when one image couldn't be scanned, and gives moderators a per-image override plus a class-aware UI.

## Backend
- **Rescan cover fix + skip-locked** (`rescanArticle`): also re-ingests the cover image — `Article.coverId` is a separate FK not present in `ImageConnection`, so a broken cover was never re-queued. The re-queue filter now also skips `nsfwLevelLocked` images (`ingestion !== Pending && !nsfwLevelLocked`) for both connection images and the cover, so a moderator override is permanent (a hard-violation rescan would otherwise re-block it).
- **New mutation** `article.resolveImageScan` (`moderatorProcedure`, flag-gated on `articleImageScanning`): `{ articleId: number, imageId: number, nsfwLevel: NsfwLevel }`. Reuses the existing `resolveIngestionError` (`{ id, nsfwLevel, userId }`) to set level + `nsfwLevelLocked` + `ingestion=Scanned`, then calls `recomputeArticleIngestion(articleId)` so the article leaves Error/Blocked, then queues the article search index.
- **Expose the cause** (`getArticleScanStatus`): now selects `scanJobs` + `nsfwLevelLocked` and plucks `scanJobs.error.{failureClass, reason}` into the per-image return shape (error images get `failureClass`/`reason`/`nsfwLevelLocked`; blocked images get `nsfwLevelLocked`).

## Frontend (`ArticleProblematicImages.tsx` / `ArticleScanStatus.tsx`)
- **Class-aware cause** per failed image: transient → "scanning failed, retrying automatically"; permanent → "can't be scanned (unsupported/corrupt) — replace it"; NotFound → "image not found — re-upload it"; Blocked → `blockedFor`.
- **Mod-only per-image Override**: a small browsing-level `Select` + Override button that calls `article.resolveImageScan` and invalidates the scan-status query on success; already-locked images render an "Overridden" badge. Rendered only when `currentUser?.isModerator`. The existing "Rescan Article" button is retained.

## Decision to call out
The mod override intentionally **covers Blocked images**: `resolveIngestionError` sets `ingestion=Scanned`, which un-blocks a policy-blocked image. This is an intentional moderator escalation path; the resulting lock keeps a later clean rescan from re-blocking, though a fresh hard violation still re-blocks.

## Notes
- `resolveImageScan` passes `{ id, nsfwLevel, userId }` to `resolveIngestionError`; the function has no `isModerator` param (the router already gates on `moderatorProcedure`).
- The override picker is a small local `Select` of `browsingLevels`/`browsingLevelLabels` — no existing browsing-level picker was cleanly reusable here (the closest, `CollectionItemNSFWLevelSelector`, is coupled to collection hooks).
- `pnpm run typecheck` passes clean.


CivitaiDevOpsAgent:
### [PR-REVIEW] round 4 — 0 finding(s) (confidence ≥ 0.80)

PR #2996, opened 2026-07-08
feat(apps): my-submissions status sections + external-review modal UX

ZacxDev:
UX polish on two App Blocks surfaces. All changes are **dark** — behind `features.appBlocks` / `appBlocksAuthor`, mod/author-gated.

## Part A — /apps/my-submissions: status-grouped sections (both lists)
Both the onsite `MySubmissionsList` and offsite `OffsiteSubmissionsList` are restructured from a single sortable table into status **SECTIONS**, in a fixed order:

**Live → Pending → Rejected (collapsed) → Withdrawn (collapsed)**

- Live + Pending are always expanded; Rejected + Withdrawn are wrapped in a default-**collapsed** `Collapse` whose header (chevron + label + count) toggles it.
- Each app **group** is bucketed by its `latest` submission's status: `approved` → Live (all deploy sub-states building/deploying/live/failed stay in Live — the row badge still distinguishes them), `pending` → Pending, `rejected` → Rejected, `withdrawn`/unknown → Withdrawn (safe closed default).
- Existing per-app version-collapse, the client text filter, and the row components (`OnsiteRow`/`OffsiteRow`) are reused unchanged — only the partitioning is new.
- Empty sections don't render (no "Rejected 0"). Section containers + collapse toggles carry stable `data-testid`s (`apps-submissions-section-*`, `apps-offsite-submissions-section-*`, `-toggle`).
- The pure bucketing (`bucketGroupsByStatus` / `statusBucket`) lives in `submissionsTable.ts` and is unit-tested; the shared section UI (`StatusSections`) lives in `submissionsTableUi.tsx`.
- Status is now the section, so the sortable column headers were dropped in favor of a plain `submittedAt desc` sort within each section.

## Part B — OffsiteReviewModal (mod external-review): 3 fixes
1. **Approve-notes gating** — approval is now behind an explicit `Approve…` click (mirrors Reject). `actionMode: 'view' | 'reject' | 'approve'`: view shows only `[Reject…] [Approve…]`; the "Approval notes (optional)" textarea + confirm Approve appear only in approve mode. `approvalNotes` cleared on close.
2. **Friendly dates** — the external-review submitted timestamps (the modal's submitted line + the review-queue "Submitted" column) now render `MMMM D, YYYY` (e.g. "June 7, 2026") via the shared date util. Audit-trail timestamps (report "Reported", moderation history) keep the full time-of-day formatter.
[...]

PR #3361, opened 2026-07-24
feat(apps): collapse moderator-removed apps into their own section on /apps/my-submissions

ZacxDev:
## What

On **/apps/my-submissions**, apps that a **moderator removed** now sort into their own **"Removed by a moderator"** section that is **collapsed by default** — in BOTH lists: on-site (`MySubmissionsList`) and off-site (`OffsiteSubmissionsList`).

## Bug fixed (precedence)

The owner bucketing sends *"any group with an approved version to Live"* (from #2996). A listing that was once live but is now moderator-removed still has an approved version in its history, so it **misfiled under the always-expanded Live section** — a takedown masquerading as Live. The new `mod-removed` bucket **takes precedence** over that any-approved→Live rule, so the removed app lands in its own collapsed section instead.

## How

- **New `mod-removed` owner-view bucket + default-collapsed section**, reusing the existing `STATUS_SECTION_META` / `StatusSection` convention (no hand-rolled accordion). Section order: Live → Pending → Rejected → Withdrawn → **Removed by a moderator** (most terminal, trailing). Label mirrors the red `ownerStateChip` "removed by a moderator" wording.
- `bucketGroupsByStatus` gains an optional **`overrideBucketOf`** that gets first refusal on each group and wins over *both* the any-approved→Live rule and the latest-status bucketing. Both lists supply it via the **reused** classifier (`isModRemovedListing` → `ownerListingState`), so the mod-removed rule is derived from `listingStatus` + `lastModerationAction`, not duplicated.
- **Owner-hidden** (the owner's own unpublish, Republish-able) stays **DISTINCT and UNCHANGED** — it is excluded from the section and keeps its normal Live placement.

## No backend change

Reuses `listingStatus` + `lastModerationAction` the list queries already return (`blocks.listMyPublishRequests` / `appListings.listMySubmissions`). **No tRPC/schema/migration** change — `removed` already exists. The whole surface is already author/mod-gated, so the change is additive/dark-safe (only re-bucketing existing rows).

## Tests

- **Pure unit (BLOCKING)** `submissionsTable.test.ts` — precedence fix: a mod-removed listing that *also* has an approved version buckets to `mod-removed`, NOT `live`; owner-hidden (last event `owner-unpublish`) does NOT land in the section; rejected/withdrawn/pending unchanged; a mutation guard proving that *without* the override the bug reproduces (misfiles under Live). Uses the real `ownerListingState` logic. **54 pass.**
- **Component browser (report-only)** `MySubmissionsList.browser.test.tsx` + `OffsiteSubmissionsList.browser.test.tsx` — the section renders collapsed by default (`aria-expanded=false`) and contains the mod-removed app after expansion; owner-hidden stays in Live and the section never renders; updated the existing mod-removed fixtures' assertions to expand the now-collapsed section. **49 pass.**

### DIFF
--- a/src/components/Apps/MySubmissionsList.browser.test.tsx
+++ b/src/components/Apps/MySubmissionsList.browser.test.tsx
@@ -762,6 +762,8 @@ describe('MySubmissionsList — P4 onsite owner status badge', () => {
     renderWithProviders(
       <MySubmissionsList submissions={[modRemoved()]} onWithdraw={vi.fn()} withdrawing={false} />
     );
+    // A mod-removed app now lives in its own default-collapsed section — expand it.
+    await page.getByTestId('apps-submissions-section-mod-removed-toggle').click();
     // The overridden status badge.
     await expect
       .element(page.getByText('removed by a moderator', { exact: true }))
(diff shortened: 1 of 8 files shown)


CivitaiDevOpsAgent:

PR #3362, opened 2026-07-24
chore(lint): add CI job, drop type-aware override and prettier plugin

JustMaier:
Deletions and additive CI only. No lint findings were fixed.

## Timings

| | before | after |
|---|---|---|
| `eslint src/utils` (82 files, no cache) | **4m 34s** | **8.4s** |
| `pnpm run lint` (full `src/`, cold cache) | ~2h 40m (extrapolated) | **1m 07s** (measured) |

The full "before" number is extrapolated rather than measured — the sample works out to 42s of program build plus ~2.8s/file, so 4,113 files is ~3.2h. I did not sit through a real full run.

## Findings after the change

3,777 total (307 errors, 3,470 warnings) across 1,119 files. Was 9,594.

```
  1762  @typescript-eslint/no-explicit-any
  1375  @typescript-eslint/no-unused-vars
   154  react-hooks/exhaustive-deps
   144  @typescript-eslint/no-empty-function
   107  @typescript-eslint/consistent-type-imports
    92  tailwindcss/classnames-order
    57  @next/next/no-img-element
    23  tailwindcss/enforces-shorthand
    22  react/no-unescaped-entities
    10  prefer-const
     7  @typescript-eslint/ban-types
     6  react-hooks/rules-of-hooks
     5  tailwindcss/migration-from-tailwind-2
     5  @typescript-eslint/no-inferrable-types
     2  jsx-a11y/alt-text
     1  each: react/display-name, @typescript-eslint/no-empty-interface,
           react/jsx-closing-bracket-location, react/no-children-prop,
           @typescript-eslint/no-var-requires, @typescript-eslint/no-empty-object-type
```

## 1. Type-aware override removed

`.eslintrc.js` set `parserOptions: { project: './tsconfig.json' }` on every `.ts`/`.tsx` while its own `rules` block was empty. I read the whole config to confirm nothing else needed type information:

- `plugin:@typescript-eslint/recommended` (v5) is the non-type-checked preset; the type-aware companion `recommended-requiring-type-checking` was commented out.
- `next/core-web-vitals`, `plugin:tailwindcss/recommended`, `react/jsx-closing-bracket-location`, `consistent-type-imports`, and `no-unused-vars` are all syntax-only.
- The local `no-io-in-transaction` rule is a pure AST walk (`eslint-local-rules.js`) with no `getTypeChecker` usage.

The sole consumer was root-level `@typescript-eslint/restrict-template-expressions` at `warn`. Both it and the override are deleted.

## 2. CI lint job

New `.github/workflows/lint.yml`, on PRs into `main`/`release`, two parallel jobs:

- **ESLint + Prettier** — changed `.ts`/`.tsx` vs the PR base only. Full-repo would fail on all 3,777 pre-existing findings and be disabled within a week. ESLint is further filtered to `src/` and `packages/`, the only trees with a config.
- **Typecheck** — full `pnpm run typecheck`. Cheap and currently unenforced: `.husky/pre-push` early-exits for `Justin Maier` and `bkdiehl`, so in practice it never runs.

Both jobs skip the submodule except typecheck, which initializes `event-engine-common` via `webfactory/ssh-agent` and `SUBMODULE_SSH_KEY` (same secret the pin-guard workflow already uses). Without it tsc emits a wall of missing-module errors.

**Node version discrepancy, not resolved here:** `.nvmrc` says `24.18.0`, the Dockerfile builds on node 22. The workflow follows `.nvmrc` per instruction. Worth reconciling separately — CI type-checking on a different major than the one that ships is a real gap, just not this PR's.

## 3. Prettier split out of ESLint

`prettier/prettier` was 5,830 of 9,594 findings, all duplicated by `pnpm prettier:check`. Verified rather than assumed:

- **File set** — `prettier:check` globs `**/*.{ts,tsx}` and there is **no `.prettierignore` file at all** (the audit said "empty"; it does not exist, same practical effect — only prettier's built-in `node_modules` exclusion applies). `pnpm lint` covers `src/` only. Superset confirmed.
- **Options** — `.prettierrc` is `{trailingComma: es5, tabWidth: 2, printWidth: 100, singleQuote: true}`; the inline ESLint options were `{printWidth: 100, endOfLine: 'auto', singleQuote: true, trailingComma: 'es5'}`. `tabWidth: 2` is prettier's default, so the only real delta is `endOfLine` — `auto` in ESLint vs the default `lf` in `.prettierrc`. `prettier:check` (the surviving path) is the stricter of the two, so nothing is lost.

`eslint-config-prettier` kept in `extends` at its current version.

## 4. Dead dependencies

Each confirmed by grep across the repo excluding `package.json` and the lockfile — zero hits: `eslint-config-airbnb`, `eslint-config-airbnb-typescript`, `eslint-config-mantine`, `eslint-import-resolver-typescript` (its `import/resolver` block has been commented out), `lint-staged`.

The `lint-staged` config block ran `tsc-files --noEmit`, and `tsc-files` is not installed and not in the lockfile. Its husky call has been commented out since someone left the note "apparently tsc-files just doesn't work on windows. at all." Block removed with the dep.

`eslint-plugin-import` **kept** — `packages/.eslintrc.cjs` registers it for `pnpm lint:packages`. Also fixed that file's stale comment claiming packages don't inherit the app's "airbnb" config; airbnb was never in `extends`.

## 5. eslint-config-next pin

Added `"eslint-config-next": "15"` to `pnpm.overrides`, and a `//eslint-config-next` note beside it explaining why. The note lives at the top level rather than inline in `devDependencies` because pnpm rejects a `//`-prefixed key inside a dependency block (`ERR_PNPM_INVALID_PACKAGE_NAME`). `.eslintrc.js` now points at the override so the two stay linked.

## 6. openai to zod override — deleted, not moved

Confirmed the audit: `package.json` had a top-level npm-style `"overrides": {"openai": {"zod": "$zod"}}`, pnpm only reads `pnpm.overrides`, and the lockfile's `overrides:` block contained no `openai` entry. It has never applied.

I removed it rather than moving it. Installed `openai@4.104.0` declares `zod: ^3.23.8` as an **optional** peer with no `zod` in its own `dependencies`; the repo is on `zod ^4.0.17` and pnpm already reports `openai 4.104.0 -> unmet peer zod@^3.23.8: found 4.0.17`. Since openai has no nested `zod` to redirect, `$zod` would resolve to the same 4.0.17 that is already there — a no-op that also asserts a compatibility openai@4 does not claim. Moving it changes no resolved version; it just makes a broken peer look sanctioned. Happy to restore it under `pnpm.overrides` if someone wants the declaration on record.

## Verified

- `pnpm run typecheck` — clean, full output read, exit 0.
- `pnpm run lint` — 1m07s, counts above.
- `pnpm run lint:packages` — **fails identically before and after** this change (verified by stashing to a clean tree): 2 `import/no-extraneous-dependencies` errors on `packages/civitai-axiom/src/__tests__/*.test.ts` for `vitest`. This looks like a Windows-only artifact — the rule's `devDependencies: ['**/*.test.ts']` allowlist is minimatched against an absolute backslash path. Pre-existing and out of scope; flagging because it means the config's own gate is red on Windows today. It should pass on the Linux CI runner.
- `package.json` parses; `lint.yml` parses as YAML.

## What this workflow actually guards (honest accounting)

Worth recording plainly, because the CI job is easy to over-read:

- **Typecheck is 100% new coverage.** `.husky/pre-push:8` exits 0 when the branch isn't `main` **or** the user is Justin/bkdiehl. Everyone develops on feature branches, so that hook has effectively never run for anyone, and type errors currently reach the preview build before anything catches them. This job catches them at PR time. That alone justifies it. Caveat: it is skipped on fork PRs (private submodule + no secrets), so a fork gets no gating from this workflow.
- **The report-only ESLint/Prettier steps guard nothing today.** That is acceptable only if the flip to blocking actually happens, and the flip is pinned to a Prettier 2 to 3 upgrade that is not scheduled.
- **The added-files gate is what makes this not-nothing in the meantime.** Newly added files block on ESLint errors and Prettier. Measured over the last 30 commits: of 20 newly added src files, 6 fail Prettier and 4 carry ESLint errors, so 8 of 20 (40%) would have been blocked. It is not the "free" gate it looks like, but it does stop the 789-file formatting backlog from growing.

**Documented follow-up (not built here):** the better long-term shape is a **baseline diff** rather than a repo-wide reformat. For each changed file, also run the check against `FETCH_HEAD:$file` and report only if it got *worse* - exact for Prettier, error-count comparison for ESLint. Roughly 15 lines, gives a genuinely blocking regression gate without touching a single existing file, and does not depend on the Prettier upgrade ever happening.

## Out of scope

typescript-eslint v5 to v8, ESLint 8 to 9 / flat config, Prettier 2 to 3, Node and `@types/node` alignment, husky hook cleanup beyond removing the dead lint-staged block, and any repo-wide `prettier --write`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,7 @@
 // eslint-config-next 16 is flat-config-only (peer `eslint >=9`); extending it
 // from eslintrc makes @eslint/eslintrc reject it and then crash formatting the
 // error ("Converting circular structure to JSON"), so lint silently never runs.
+// Enforced by the `eslint-config-next: "15"` entry in package.json `pnpm.overrides`.
 module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
@@ -12,7 +13,7 @@ module.exports = {
     ecmaVersion: 'latest',
     sourceType: 'module',
   },
-  plugins: ['@typescript-eslint', 'prettier', 'local-rules'],
+  plugins: ['@typescript-eslint', 'local-rules'],
   extends: [
     'next/core-web-vitals',
     'plugin:@typescript-eslint/recommended', // lightweight rules (no type info)
@@ -48,16 +49,9 @@ module.exports = {
 
     // 'import/no-cycle': ['error'],
 
-    // prettier overrides
-    'prettier/prettier': [
-      'error',
-      {
-        printWidth: 100,
-        endOfLine: 'auto',
-        singleQuote: true,
-        trailingComma: 'es5',
-      },
-    ],
+    // Formatting is owned by `pnpm prettier:check` / `prettier:write`, not by
+    // eslint-plugin-prettier. `eslint-config-prettier` (extended above) stays so
+    // ESLint's stylistic rules don't fight it.
 
     // rule tweaks
     'no-use-before-define': 'off',
@@ -67,10 +61,6 @@ module.exports = {
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/consistent-type-imports': ['error'],
-    '@typescript-eslint/restrict-template-expressions': [
-      'warn',
-      { allowBoolean: true },
-    ],
 
     'tailwindcss/no-custom-classname': [
       'off',
@@ -80,19 +70,7 @@ module.exports = {
     ],
   },
 
-  overrides: [
-    {
-      files: ['*.ts', '*.tsx'],
-      parserOptions: {
-        project: './tsconfig.json',
-      },
-      // extends: ['plugin:@typescript-eslint/recommended-requiring-type-checking'],
-      rules: {
-        // put only the rules that *need* type info here
-        // example:
-        // '@typescript-eslint/no-floating-promises': 'error',
-        // '@typescript-eslint/no-misused-promises': 'error',
-      },
-    },
-  ],
+  // No type-aware linting: `parserOptions.project` costs ~40s of program build plus
+  // ~2.3s/file (2h40m across the repo). If a type-aware rule is ever worth that, add
+  // an `overrides` entry scoped to the narrowest possible file set.
 };
--- a/.github/problem-matchers/eslint-unix.json
+++ b/.github/problem-matchers/eslint-unix.json
@@ -1,0 +1,19 @@
+{
+  "__comment": "Turns `eslint --format unix` output into GitHub file annotations. Paths must be relative to GITHUB_WORKSPACE, so lint.yml pipes eslint through sed to strip the workspace prefix. Line shape: src/foo.ts:12:5: message [Error/rule-name]",
+  "problemMatcher": [
+    {
+      "owner": "eslint-unix",
+      "pattern": [
+        {
+          "regexp": "^(.+):(\\d+):(\\d+):\\s(.+)\\s\\[(Error|Warning)/(.+)\\]$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "message": 4,
+          "severity": 5,
+          "code": 6
+        }
+      ]
+    }
+  ]
+}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,9 +1,19 @@
 name: Lint
 
-# Regression guard for code quality. Deliberately scoped so it stays green and fast:
-# ESLint and Prettier run ONLY on files the PR changed (the repo carries thousands of
-# pre-existing findings; a full-repo gate would fail every PR and get switched off).
-# Typecheck is full-repo because it is already expected to pass.
+# Regression guard for code quality.
+#
+# Typecheck is full-repo and BLOCKING - it passes today, so it is the real gate.
+#
+# ESLint and Prettier are report-only (`continue-on-error`). They are scoped to the
+# files a PR touches, but note that for a formatter "changed file" means the WHOLE
+# file, not the changed lines: 789 of 4,116 src files fail `prettier --check` today.
+# Across the 98 src files touched by the last 30 commits, 39 fail Prettier and 10
+# carry a pre-existing ESLint error, so 44 (45%) would red this job for reasons
+# unrelated to the PR. Blocking on that turns a 3-line bugfix into a 200-line
+# reformat, and the job gets disabled within a week. Annotations still surface.
+#
+# Both steps flip to blocking once the formatting/lint backlog is cleared, which is
+# planned to ride along with the Prettier 2->3 upgrade (that forces a reformat anyway).
 
 on:
   pull_request:
@@ -18,7 +28,7 @@ concurrency:
 
 jobs:
   eslint:
-    name: ESLint + Prettier (changed files)
+    name: ESLint + Prettier (changed files, report-only)
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
@@ -42,14 +52,17 @@ jobs:
           BASE_REF: ${{ github.base_ref }}
         run: |
           set -euo pipefail
-          git fetch -q --depth=1 origin "$BASE_REF"
+          git fetch -q origin "$BASE_REF"
           git diff --name-only --diff-filter=ACMR FETCH_HEAD...HEAD -- '*.ts' '*.tsx' > changed.txt || true
           echo "changed files:"
           cat changed.txt
           if [ -s changed.txt ]; then echo "any=true" >> "$GITHUB_OUTPUT"; else echo "any=false" >> "$GITHUB_OUTPUT"; fi
 
+      # Report-only: 10% of recently-touched files carry a pre-existing
+      # error-severity finding. See the header comment.
       - name: ESLint
         if: steps.changed.outputs.any == 'true'
+        continue-on-error: true
         run: |
           set -euo pipefail
           # eslint only has a config for src/ and packages/; skip anything else.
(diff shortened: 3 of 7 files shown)


JustMaier:
All five review points addressed in d02e9c3. I reproduced every measurement independently before changing anything — the review's numbers hold.

## Reproduced measurements

My sample is 98 files (`HEAD~30..HEAD` on this branch, which includes my own commits) vs the review's 90 from `main`, so counts run slightly higher. Same conclusion either way.

| | review | mine | 
|---|---|---|
| src files failing `prettier --check` | 790 / 4,109 | **789 / 4,116** (19%) |
| recently-touched files failing Prettier | 35 / 90 (39%) | **39 / 98** (40%) |
| ESLint problems in sample | 188 | **196** |
| files with >=1 error-severity finding | 9 (10%) | **10** (10%) |
| **union that would red the job** | 36 / 90 (40%) | **44 / 98 (45%)** |

Error-severity breakdown matches almost exactly: 7 `no-empty-function`, 5 `consistent-type-imports`, 5 `prefer-const`, 1 `no-inferrable-types`.

## 1 + 2 — BLOCKER / HIGH: both lint steps are now report-only

`continue-on-error: true` on the ESLint and Prettier steps, with the rationale and the flip-to-blocking plan (riding the Prettier 2→3 upgrade) recorded in the workflow header and beside each step. No repo-wide `prettier --write` — Justin's call stands. Typecheck stays blocking. Job renamed to `ESLint + Prettier (changed files, report-only)` so the intent is legible from the checks list.

## 3 — HIGH: fork PRs

I went further than the suggested fix. Gating just the ssh-agent and submodule steps stops the "ssh-private-key argument is empty" hard failure, but a fork still cannot fetch the private `event-engine-common`, so `tsc` then fails with the missing-module wall — the same unfixable red check, one step later. So the guard is on the **job**:

```yaml
if: github.event.pull_request.head.repo.full_name == github.repository
```

Fork PRs get a skipped check instead of a red one. Left a note not to make this a required check until the fork case is genuinely solved.

## 4 — MEDIUM: `--depth=1` dropped

Now `git fetch -q origin "$BASE_REF"`, matching `submodule-pin-guard.yml:41`.

## 5 — Also

- **Header comment fixed.** It now states explicitly that for a formatter, changed-file means the whole file, with the measured numbers inline.
- **`.husky/pre-commit` cleaned** — the commented-out `npx lint-staged` block, the dead branch check, and the `tsc-files` note are gone. The hook is now just the ToS lastmod logic plus `exit 0`.
- **Typecheck timeout 15 → 25 minutes.**
- **`prisma generate` with no `DATABASE_URL`: verified, not assumed.** Ran `pnpm run db:generate` with `DATABASE_URL` and `DATABASE_REPLICA_URL` explicitly unset — exit 0, all four generators (Prisma Client, enums, TS interfaces, Kysely types) succeeded. That was the one unverified assumption between this workflow and a red first run; it's closed. It also produced no tracked file changes, so `postinstall` won't dirty the CI tree.

## Verified after the changes

- Workflow parses as YAML; asserted programmatically that `jobs.typecheck.if` is set and that exactly the ESLint and Prettier steps carry `continue-on-error`.
- Working tree clean apart from the two intended files.

One correction to the earlier PR body, now that `.gitattributes` (`* text=auto eol=lf`) settles it: my `endOfLine` caveat was noise. Checkouts are LF everywhere, so `auto` and `lf` were never going to diverge in practice.

JustMaier:
All three addressed in 434f082. I ran the workflow's bash logic locally against real diff ranges rather than eyeballing it, which turned up one thing that changes the picture on item 3.

## 1 — You were right, and it was worse than "no annotations"

Confirmed: no `::add-matcher::` and no matcher JSON anywhere in `.github/`. Findings would have landed in a step log inside a **green** check. Fixed properly rather than by rewording.

Added `.github/problem-matchers/eslint-unix.json` and registered it. ESLint now runs `--format unix`, piped through `sed "s|^$GITHUB_WORKSPACE/||"` to relativize paths (annotations need paths relative to the workspace). Verified the regex against real ESLint output — it matches all three severity/rule shapes and correctly ignores the trailing `N problems` summary and blank lines:

```
src/utils/trpc.ts:294:50: '_error' is defined but never used. [Warning/@typescript-eslint/no-unused-vars]
  -> MATCH file=src/utils/trpc.ts line=294 sev=Warning rule=@typescript-eslint/no-unused-vars
```

Prettier annotates via `::error` (added) / `::warning` (modified) commands.

**Also caught a latent bug while testing this.** Piping eslint through `sed` masks its exit code unless `pipefail` is set — without it the blocking step would have silently never blocked. Every `run` block already has `set -euo pipefail`; I proved it end to end (`pipeline exit=1` with pipefail, masked without) and asserted programmatically that no `xargs` step is missing it.

## 2 — Header qualified

Now reads: "Typecheck is full-repo and BLOCKING on internal PRs, and is the only real gate here. It is skipped on fork PRs (see the job comment), so a fork gets no gating from this workflow at all."

## 3 — Implemented, but it is NOT free. Please read this bit.

Split by diff filter: `--diff-filter=A` blocking, `--diff-filter=CMR` report-only. Zero-match guarded on every step; verified against **this PR**, which touches no `.ts` files at all — all four steps no-op cleanly, no empty-`xargs` misfire.

But the premise doesn't hold. Simulated over the last 30 commits, which added 20 src files:

- **6 of 20 (30%) fail `prettier --check`**
- **4 of 20 carry ESLint errors** (6 errors: 4 `no-empty-function`, 2 `consistent-type-imports`)
- **Union: 8 of 20 = 40% of recently-added files would have been blocked**

So this is a real new constraint, not a free one. Nearly all of them are recently-added *test* files that merged unformatted — which is precisely the backlog growing, so the change is doing its job. And the fix is one command on a file you just wrote. I kept it blocking as asked, but you should merge knowing it will bite on roughly a third of PRs that add files, not zero. Easy to soften to report-only if that's not the trade you want.

One deliberate deviation: **no `--max-warnings 0`** on added files. Failing new files on warnings would hold them to a stricter bar than every one of the 3,470 warnings and 1,762 `any`s already merged. Errors block, warnings annotate.

## Recorded in the PR body

- **Typecheck is 100% new coverage.** `.husky/pre-push:8` exits 0 when the branch isn't `main` *or* the user is Justin/bkdiehl, and everyone works on feature branches, so it has effectively never run. Type errors currently reach the preview build. That alone justifies the job.
- **Report-only ESLint/Prettier guard nothing today**, and the flip is pinned to a Prettier 2→3 upgrade that isn't scheduled. The added-files gate is what makes this not-nothing meanwhile.
- **Documented follow-up, not built:** baseline diff — for each changed file also check `FETCH_HEAD:$file` and report only if it got *worse* (exact for Prettier, error-count comparison for ESLint). ~15 lines, genuinely blocking regression gate, touches no existing file, and doesn't depend on the Prettier upgrade happening.

## Verified

YAML parses; asserted `jobs.typecheck.if` is set, that exactly the two modified-file steps carry `continue-on-error` while both added-file steps do not, and that no `xargs` step is missing `pipefail`.

PR #3365, opened 2026-07-24
chore: align Node version across engines, nvmrc, Dockerfile and types

JustMaier:
## Before

| Source | Said |
| --- | --- |
| `package.json` `engines` | absent |
| `.nvmrc` | `24.18.0` |
| `Dockerfile` (deps + runner) | `node:22-alpine3.20` -> frozen at **Node 22.16.0** |
| `@types/node` | `20.19.9` |
| Installed locally | `v22.18.0` |

## What this changes

- `.nvmrc`: `24.18.0` -> `22.16.0`
- `package.json`: add `"engines": { "node": ">=22.0.0" }`
- `@types/node`: `20.19.9` -> `22.20.1` (+ lockfile)
- `flake.nix`: `nodejs_20` -> `nodejs_22`
- `.devcontainer/public/docker-compose.yml`: `typescript-node:1-20` -> `1-22`
- **No Dockerfile changes.** Zero production runtime risk.

## Why 22

`.nvmrc` at 24.18.0 is accidental, not deliberate. #3242 moved the runtime to Node 24 (`node:24-alpine3.22` + `.nvmrc` `22.16.0` -> `24.18.0`). Three days ago #3311 rolled the base image back to `node:22-alpine3.20` after Node 24 / V8 13.6 caused a large per-request server-CPU regression in native / JS<->C++ boundary work (nodejs/node#62939, #60719 - unresolved upstream, no `NODE_OPTIONS` flag recovers it). That rollback touched only the Dockerfile and left `.nvmrc` stranded on 24.

The Dockerfile is the deliberate, evidence-backed decision, so everything else follows it. The seven other Dockerfiles in the repo (`apps/{auth,creator-studio,event-engine,notifications,orchestrator-gateway,storage}` plus the root) are all already on `node:22-alpine`, so 22 is what the whole repo ships bar one test helper (`containers/image-stub`, `node:20-bookworm`).

**LTS status** (nodejs/Release schedule):

| Line | Status today (2026-07-24) | EOL |
| --- | --- | --- |
| 20 (Iron) | **EOL** | 2026-04-30 |
| 22 (Jod) | Maintenance LTS | 2027-04-30 |
| 24 (Krypton) | Active LTS (maintenance 2026-10-20) | 2028-04-30 |

Node 22 is maintenance-only but supported for another ~9 months, ample runway to wait out the V8 fix. `@types/node` at 20 was tracking a line that is already EOL - worth fixing regardless of which major we land on.

**Dependency floors:** `next@16.2.7` declares `engines.node: ">=20.9.0"`. Nothing in the tree requires a major above 22. The only existing workflow, `submodule-pin-guard.yml`, pins no Node version.

**Why `22.16.0` and not a bare `22`:** `node:22-alpine3.20` is **not** a floating tag. It was last pushed 2025-05-21 and Alpine 3.20 hit EOL 2026-04-01, so the tag stopped being rebuilt and is frozen at Node 22.16.0. A bare `22` resolves through `actions/setup-node` to the latest 22.x (22.23.1 today), which would have left #3362's CI typechecking **seven minors** of V8/undici/builtin drift ahead of every prod pod - the same class of CI-vs-prod gap that produced #3242 -> #3311. `22.16.0` is exactly the value #3242 replaced, and it was correct.

**Why `>=22.0.0` with an open ceiling:** #3311 explicitly kept the Node 24 image sha on ghcr as the roll-forward pin for when the upstream V8 regression is fixed. The intent is to return to 24, so a `<23` ceiling would fight the plan. The floor exists to keep anyone off EOL Node 20. Note this is advisory only - `.npmrc` is a single line with no `engine-strict`, so pnpm warns rather than fails. The operative pin is `.nvmrc` plus the Dockerfile.

## Interaction with #3362

#3362 (open, unmerged) adds `.github/workflows/lint.yml` with `node-version-file: .nvmrc` on both jobs. With `.nvmrc` at 24.18.0 that workflow would have typechecked on Node 24 while prod shipped Node 22. After this PR it resolves to 22.16.0, exactly matching the image. **The two PRs need to agree - no change is needed on #3362's side, it just needs this one merged (order does not matter).**

## Type errors surfaced

None. `pnpm run typecheck` is clean with `@types/node@22.20.1`. The 20 -> 22 delta is purely additive with identical `undici-types`. No source files changed, so `lint` is unaffected.

The lockfile's `+347/-54` is **peer-hash snapshot duplication**, not runtime movement: zero runtime packages changed version, but `apps/*` still pin `@types/node@^20`, so both majors resolve in one tree and msw/vite/vitest/@inquirer each materialize a second graph. Install-tree size only.

## PRIORITY follow-up for a human - deliberately not in this PR

**`node:22-alpine3.20` is a dead base image.** Both layers are frozen at 2025-05-21: Node stuck at 22.16.0, and Alpine 3.20 EOL since 2026-04-01. That is **14 months of unpatched Node and OS CVEs** currently in production. #3311's rollback landed on an already-dead tag and nobody caught it. Current Node 22 Alpine tags are `22-alpine3.22`, `3.23`, `3.24` (rebuilt 2026-06-23); moving to `node:22-alpine3.24` keeps Node on 22 and only refreshes the runtime patch plus OS layer.

**This is a production runtime change and Justin's call**, so it is not bundled here - it wants its own PR and its own canary. This PR's `.nvmrc` would then need updating to whatever Node version that tag carries.

## Other deferrals

- **`apps/*` still pin `@types/node@^20`** (`auth`, `moderator`, `notifications`, `orchestrator-gateway`, `storage`; `event-engine` at `^20.10.6`), while their Dockerfiles run `node:22-alpine`. Same drift this PR fixes at the root, one level down.
- **Latent inconsistency worth knowing about:** `packages/*/src` is now typechecked against `@types/node@22` by root `tsc` and against `@types/node@20` by each app's `tsc`. Safe today because the delta is purely additive. But the moment shared package source touches a Node-22-only surface (`node:sqlite`, `navigator`, `localStorage`, the new `web-globals` types), root typecheck passes and every app build breaks - and the apps have no CI, so it surfaces at deploy. Bumping `apps/*` to `@types/node@22` closes it.
- **`containers/image-stub/Dockerfile`** is still `node:20-bookworm`. Test-only image stub, no production surface.


JustMaier:
Review addressed in 7576f50. All five points confirmed independently before changing anything - the review was right on every one.

## 1. MUST FIX - `.nvmrc` -> `22.16.0` (fixed)

Verified: Docker Hub reports `node:22-alpine3.20` `last_updated` **2025-05-21**, and there is no newer push. Alpine 3.20 EOL 2026-04-01, so the tag is frozen, not floating. My "the Dockerfile pins a floating major tag" premise was simply wrong, and the conclusion built on it de-aligned CI from prod by seven minors (22.23.1 vs 22.16.0) - the exact gap the PR exists to close.

Also verified `git show 3c50266519 -- .nvmrc` is `-22.16.0 / +24.18.0`, so the exact-patch pin was deliberate and correct. Restored it and rewrote the rationale section to say so rather than framing it as over-precision.

Base-image bump **not** bundled, as directed. Promoted to a `## PRIORITY follow-up for a human` section with the CVE framing: Node frozen at 22.16.0 *and* Alpine 3.20 EOL, both since 2025-05-21 = 14 months of unpatched Node and OS in production, and #3311's rollback landed on an already-dead tag unnoticed. Flagged that this PR's `.nvmrc` needs updating alongside whenever that lands.

## 2. Three wrong claims corrected

- **Dockerfiles.** My `find` was wrong; `Glob **/Dockerfile*` returns 12. Seven Node images beyond the root: `apps/{auth,creator-studio,event-engine,notifications,orchestrator-gateway,storage}` all `node:22-alpine` (unpinned), plus `containers/image-stub` on `node:20-bookworm`. Reframed as supporting evidence - the whole repo already ships 22 - and `image-stub` is listed as a deferral rather than omitted.
- **`engines` enforcement.** Confirmed `.npmrc` is one line (`public-hoist-pattern[]=@types/*`) and `engine-strict` appears nowhere in the repo, so `engines` is advisory. Body now says so explicitly, and the open ceiling is justified by #3311 keeping the Node 24 sha on ghcr as the roll-forward pin - returning to 24 is the stated plan, so `<23` would fight it.
- **Lockfile.** Recharacterized as peer-hash snapshot duplication (`apps/*` on `@types/node@^20` forces both majors into one tree; msw/vite/vitest/@inquirer each materialize a second graph), explicitly install-tree size and not runtime movement.

## 3. Both Node 20 environments bumped, not deferred

Neither is a production surface, so both moved:

- `flake.nix:54` `nodejs_20` -> `nodejs_22`. Input is `nixpkgs-unstable`, so `nodejs_22` resolves.
- `.devcontainer/public/docker-compose.yml:11` `typescript-node:1-20` -> `1-22`.

(`.devcontainer/internal/` has no Node image - only redis/signals/buzz service images.)

Note on that file: prettier wanted to reflow an unrelated `"5001:3000"` -> `'5001:3000'` on an adjacent line. Reverted it so the diff stays one line.

## 4. Latent inconsistency documented

Added to deferrals: `packages/*/src` is now checked against `@types/node@22` by root `tsc` and against 20 by each app's `tsc`. Harmless while the delta is additive, but the first shared-package use of a 22-only surface (`node:sqlite`, `navigator`, `localStorage`, `web-globals`) passes root typecheck and breaks every app build - with no app CI, that surfaces at deploy.

## Verification

- `pnpm run typecheck` - clean, full unfiltered output, zero errors.
- `pnpm exec prettier` - changed files formatted correctly.
- No source files changed, so `lint` is unaffected.

Final diff is 5 files: `.nvmrc`, `package.json`, `pnpm-lock.yaml`, `flake.nix`, `.devcontainer/public/docker-compose.yml`. Still no Dockerfile changes.

PR #3366, opened 2026-07-24
feat(training): epoch-aware sample image viewer

JustMaier:
## The request

From MNeMiC, in the Creator Studio review group:

> "When having trained a model on the site, let me view the images better. I cannot zoom in or view the images in full, nor compare reasonably.
> I would go with how AIToolkit does it. A vertical grid.
> If I click on an image, I view that, if I go Left/Right, I go to those images. If I go up/down, I go to the same image but in the epoch above/below."

## What the data actually looks like

The surface is `src/components/Resource/Forms/TrainingSelectFile.tsx` (the training wizard's "select file" step). It reads `modelFile.metadata.trainingResults`.

**V2 (`TrainingResultsV2`, current):**

```ts
epochs: { epochNumber: number; modelUrl: string; modelSize: number; sampleImages: string[] }[]
sampleImagesPrompts: string[]
```

**Cross-epoch sample consistency: yes, by construction.** `sampleImagesPrompts` is stored **once for the whole run**, not per epoch - it is written from the user's configured sample prompt list (`trainingStep.input.samples.prompts`) at workflow submit time and re-stamped on every webhook (`src/server/services/training.service.ts:555`, `:678`; `training.orch.ts:425`). Each epoch's `sampleImages` is `epoch.samples.map(s => s.url)` in orchestrator order (`training.service.ts:559-564`). So **sample index N is the same prompt in every epoch** - the existing UI already relied on this (it rendered `prompts[index]` under each epoch's image). The vertical axis is sound; no prompt-text keying needed.

The one real edge case is **ragged rows**: `sampleImages` entries are `s.url ?? ''`, and a partially-completed or failed epoch can carry fewer entries than its neighbours. Handled explicitly (see below) rather than assumed away.

**Legacy V1** stores `sample_images: { image_url, prompt }[]` per epoch; the existing normalization already flattens it to `epochs[0]`'s prompts, so both paths feed the same `{ epochs, prompts }` shape and the viewer needed no V1-specific code.

**URLs are orchestrator blob URLs, not Cloudflare Images IDs.** `getEdgeUrl` short-circuits on anything starting with `http` (`src/client-utils/cf-images-utils.ts:86`), so `EdgeImage`/`EdgeMedia` would be a pure pass-through here and buys nothing - using plain `img`/`video` (matching what this file already did) also lets the viewer own its own `onLoad`/`onError` states.

## What I built

**Grid (`TrainingSelectFile.tsx`)** - each epoch row is now a CSS grid with `repeat(columnCount, minmax(160px, 1fr))` inside a horizontally scrollable container, where `columnCount` is the max sample count across all epochs (and the prompt count). Columns therefore line up vertically across every epoch card, which is what makes the comparison readable. An epoch missing a sample at index N renders a dashed "No sample" placeholder in that cell instead of shifting its neighbours left. Thumbnails are buttons with a hover zoom affordance; the click `stopPropagation`s so opening the viewer doesn't also select that epoch's file.

**Viewer (`src/components/Training/TrainingSampleViewer/TrainingSampleViewer.tsx`, new)** - fullscreen Mantine `Modal` driven by `useDialogContext()`, opened via `dialogStore.trigger` behind a dynamic import.

- **Left / Right** - adjacent sample within the current epoch, clamped (no wrap).
- **Up / Down** - same sample index, adjacent epoch. Epochs that have no sample at that index are **skipped** to the nearest one that does, so the vertical axis never lands on an empty frame.
- **Escape** closes (Mantine default).
- Arrow keys are handled on a `window` `keydown` listener that calls `preventDefault()`, so the page behind the modal does not scroll.
- Header shows `Epoch #N`, position (`3 of 12 epochs`), and `sample 2 of 4`. Prompt is shown at the bottom with a copy button.
- Zoom: click the image (or the toolbar button) toggles fit-to-screen vs 1:1 natural size in a scrollable container. Reset on navigation.
- Loading spinner and an explicit error state for samples that 404 (sample media is purged 30 days after training).
- On-screen up/down/left/right buttons for mouse users, disabled + tooltip-explained at the edges. Video runs get `video` elements in both grid and viewer.

**Reused vs written:** reused `dialogStore` + `useDialogContext` (the same pattern `GeneratedOutputWrapper` uses for `GeneratedOutputLightbox`), Mantine `Modal`/`ActionIcon`/`Tooltip`/`Badge`, and the existing `CopyButton`. Written new: the viewer component and its navigation logic. I did **not** reuse `GeneratedOutputLightbox` - it is Embla-based and hardwired to the generation-queue request store, and Embla is a 1D carousel with no second axis, so retrofitting a vertical epoch axis onto it would have been more code than a purpose-built viewer.

**Drive-by:** removed a stray `console.log({ tResultsV2, buzzType })` that was shipping on every render of this step, and dropped the now-unused `.epochRow` CSS class.

## Verification

- `pnpm run typecheck` - clean (full output read, unfiltered).
- `pnpm exec eslint <changed files>` - no issues, no new warnings.
- `pnpm exec prettier --write <changed files>` - formatted.
- No dev server / browser automation was run, hence the checklist below.

## Human visual + interaction checklist

Open the training wizard for a completed run: `/models/train?modelId=<id>&step=4` (or Model -> version with `trainingStatus = InReview`/`Approved` -> "Select File" step).

**Grid**
1. Every epoch card shows its samples in a grid whose columns line up vertically with the cards above and below it - sample 1 sits directly under sample 1 of the previous epoch, including across the "Recommended" / "Other Results" heading split.
2. Prompt textarea still sits under each thumbnail and still reads the same prompt for a given column in every epoch.
3. Hovering a thumbnail dims it and shows a magnifier icon; the cursor is `zoom-in`.
4. Clicking a thumbnail opens the viewer and does **not** change which epoch is selected (green selected-row border should not move).
5. Clicking anywhere else on the card still selects that epoch as before.
6. On a narrow window the grid scrolls horizontally inside the card; the page body itself must not scroll sideways.
7. An epoch with zero samples still shows the "No images available" message.

**Viewer - keyboard matrix**

| Key | Expected |
|---|---|
| Right | next sample, same epoch; no-op at the last sample |
| Left | previous sample, same epoch; no-op at the first sample |
| Up | same sample index, next epoch **up** the page (higher epoch number) |
| Down | same sample index, next epoch **down** the page (lower epoch number) |
| Esc | closes the viewer |

8. Hold Down from the top epoch to the bottom and back with Up - the image should change but the prompt text at the bottom must stay identical the whole way. This is the core of the feature.
9. While any arrow is pressed, the page behind the modal must not scroll. Check with the grid scrolled partway down: the scroll position should be unchanged when you close.
10. Header reads e.g. `Epoch #14` + `1 of 12 epochs - sample 2 of 4`, and updates on every move.
11. Edge buttons: at the top epoch the Up button is disabled with a "No later epoch with this sample" tooltip; same for Down at the bottom, Left at sample 1, Right at the last sample.

**Ragged epochs (an epoch with a different sample count than its neighbour)**

This is the case worth deliberately hunting for - a failed/partial run, or a run where one sample errored. If you can't find one naturally, temporarily trim one epoch's `sampleImages` array in devtools.

12. In the grid, the short epoch shows a dashed **"No sample"** placeholder in the missing column - its remaining thumbnails must **not** slide left out of alignment with the other epochs.
13. In the viewer, sitting on the missing index and pressing Down must **skip over** that short epoch and land on the next epoch that actually has that sample; the tooltip should name the epoch number it will jump to.
14. If no epoch in that direction has the sample, the button is disabled and the key is a no-op - you must never see a blank/broken frame.
15. Navigating Right inside the short epoch clamps at its own last sample, not the global max.

**Media states**
16. Zoom: click the image (or the toolbar magnifier) - it switches to natural size and the container becomes scrollable; click again to fit. Zoom resets when you navigate to another sample.
17. A large sample at 1:1 should be pannable via scroll, and the nav buttons must stay visible on top of it.
18. Error state: for a run older than 30 days (sample media purged), the viewer shows the warning icon and "This sample could not be loaded." rather than a broken-image glyph. The grid thumbnail may still show the browser's broken-image icon - acceptable, flag it if it looks bad.
19. Spinner appears while a large sample loads and the image fades in rather than popping.
20. Video training runs (`mediaType: 'video'`): grid tiles autoplay muted as before; the viewer shows a single `video` element with controls, and the zoom button is hidden.

### DIFF
--- a/src/components/Resource/Forms/TrainingSelectFile.module.css
+++ b/src/components/Resource/Forms/TrainingSelectFile.module.css
@@ -1,10 +1,7 @@
-.epochRow {
-  flex-wrap: nowrap;
-
-  @container (max-width: theme('screens.sm')) {
-    flex-direction: column;
-    gap: var(--mantine-spacing-md);
-  }
+.sampleGrid {
+  display: grid;
+  gap: var(--mantine-spacing-sm);
+  align-items: start;
 }
 
 .selectedRow {
--- a/src/components/Resource/Forms/TrainingSelectFile.tsx
+++ b/src/components/Resource/Forms/TrainingSelectFile.tsx
@@ -259,7 +259,7 @@ const EpochRow = ({
             </Group>
           )}
         </Group>
-        {columnCount === 0 ? (
+        {!epoch.sampleImages?.some(Boolean) ? (
           <Center p="md">
             <Text>No images available</Text>
           </Center>
@@ -309,7 +309,7 @@ const EpochRow = ({
                         </span>
                       </button>
                     ) : (
-                      <Center className="h-[200px] w-full rounded border border-dashed border-gray-4">
+                      <Center className="h-[200px] w-full rounded border border-dashed border-gray-4 dark:border-dark-4">
                         <Text size="xs" c="dimmed">
                           No sample
                         </Text>
@@ -722,12 +722,8 @@ export default function TrainingSelectFile({
   epochs = [...epochs].sort((a, b) => b.epochNumber - a.epochNumber);
 
   // Sample index N is the same prompt in every epoch (`sampleImagesPrompts` is stored once per
-  // run), so columns stay aligned even when an epoch emitted fewer samples than its neighbours.
-  const columnCount = Math.max(
-    0,
-    ...epochs.map((e) => e.sampleImages?.length ?? 0),
-    epochs.length ? samplePrompts.length : 0
-  );
+  // run), so columns stay aligned even when an epoch is missing a sample at some index.
+  const columnCount = Math.max(0, ...epochs.map((e) => e.sampleImages?.length ?? 0));
 
   const openSample = (epochIndex: number, sampleIndex: number) => {
     import('~/components/Training/TrainingSampleViewer/TrainingSampleViewer').then(
--- a/src/components/Resource/Forms/TrainingSelectFile.tsx
+++ b/src/components/Resource/Forms/TrainingSelectFile.tsx
@@ -4,7 +4,6 @@ import {
   Center,
   Flex,
   Group,
-  Image,
   Loader,
   Menu,
   Paper,
@@ -23,8 +22,10 @@ import {
   IconDotsVertical,
   IconFileDownload,
   IconRepeat,
+  IconZoomIn,
 } from '@tabler/icons-react';
 import { CopyButton } from '~/components/CopyButton/CopyButton';
+import { dialogStore } from '~/components/Dialog/dialogStore';
 import clsx from 'clsx';
 import dayjs from '~/shared/utils/dayjs';
 import { useRouter } from 'next/router';
@@ -66,6 +67,9 @@ const TRANSMITTER_KEY = 'trainer';
 
 const EpochRow = ({
   epoch,
+  epochIndex,
+  columnCount,
+  onOpenSample,
   prompts,
   selectedFile,
   setSelectedFile,
@@ -81,6 +85,9 @@ const EpochRow = ({
   modelName,
 }: {
   epoch: TrainingResultsV2['epochs'][number];
+  epochIndex: number;
+  columnCount: number;
+  onOpenSample: (epochIndex: number, sampleIndex: number) => void;
   prompts: TrainingResultsV2['sampleImagesPrompts'];
   selectedFile: string | undefined;
   setSelectedFile: React.Dispatch<React.SetStateAction<string | undefined>>;
@@ -252,55 +259,75 @@ const EpochRow = ({
             </Group>
           )}
         </Group>
-        <Group
-          className={classes.epochRow}
-          style={{ justifyContent: 'space-evenly', alignItems: 'flex-start' }}
-        >
-          {epoch.sampleImages && epoch.sampleImages.length > 0 ? (
-            epoch.sampleImages.map((url, index) => (
-              <Stack key={index} style={{ justifyContent: 'flex-start' }}>
-                {isVideo ? (
-                  <video
-                    loop
-                    playsInline
-                    disablePictureInPicture
-                    muted
-                    autoPlay
-                    controls={false}
-                    height={200}
-                    // width={180}
-                    className="w-full object-cover"
-                  >
-                    <source src={url} type="video/mp4" />
-                  </video>
-                ) : (
-                  <Image
-                    alt={`Sample image #${index}`}
-                    src={url}
-                    style={{
-                      height: '200px',
-                      // if we want to show full image, change objectFit to contain
-                      objectFit: 'cover',
-                      // object-position: top;
-                      width: '100%',
-                    }}
-                  />
-                )}
-                <Textarea
-                  autosize
-                  minRows={1}
-                  maxRows={4}
-                  value={prompts[index] || '(no prompt provided)'}
-                  readOnly
-                />
-              </Stack>
-            ))
-          ) : (
-            <Center p="md">
-              <Text>No images available</Text>
-            </Center>
-          )}
-        </Group>
+        {columnCount === 0 ? (
+          <Center p="md">
+            <Text>No images available</Text>
+          </Center>
+        ) : (
+          <div className="overflow-x-auto">
+            <div
+              className={classes.sampleGrid}
+              style={{ gridTemplateColumns: `repeat(${columnCount}, minmax(160px, 1fr))` }}
+            >
+              {Array.from({ length: columnCount }, (_, index) => {
+                const url = epoch.sampleImages?.[index];
+                return (
+                  <Stack key={index} gap="xs">
+                    {url ? (
+                      <button
+                        type="button"
+                        aria-label={`View epoch ${epoch.epochNumber} sample ${index + 1} full size`}
+                        className="group relative block h-[200px] w-full cursor-zoom-in overflow-hidden rounded p-0"
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          onOpenSample(epochIndex, index);
+                        }}
+                      >
+                        {isVideo ? (
+                          <video
+                            loop
+                            playsInline
+                            disablePictureInPicture
+                            muted
+                            autoPlay
+                            controls={false}
+                            className="size-full object-cover"
+                          >
+                            <source src={url} type="video/mp4" />
+                          </video>
+                        ) : (
+                          // eslint-disable-next-line @next/next/no-img-element
+                          <img
+                            alt={`Epoch ${epoch.epochNumber} sample ${index + 1}`}
+                            src={url}
+                            loading="lazy"
+                            className="size-full object-cover"
+                          />
+                        )}
+                        <span className="absolute inset-0 flex items-center justify-center bg-black/40 opacity-0 transition-opacity group-hover:opacity-100">
+                          <IconZoomIn size={28} color="white" />
+                        </span>
+                      </button>
+                    ) : (
+                      <Center className="h-[200px] w-full rounded border border-dashed border-gray-4">
+                        <Text size="xs" c="dimmed">
+                          No sample
+                        </Text>
+                      </Center>
+                    )}
+                    <Textarea
+                      autosize
+                      minRows={1}
+                      maxRows={4}
+                      value={prompts[index] || '(no prompt provided)'}
+                      readOnly
+                    />
+                  </Stack>
+                );
+              })}
+            </div>
+          </div>
+        )}
       </Stack>
     </Paper>
   );
@@ -694,6 +721,25 @@ export default function TrainingSelectFile({
   }
   epochs = [...epochs].sort((a, b) => b.epochNumber - a.epochNumber);
 
+  // Sample index N is the same prompt in every epoch (`sampleImagesPrompts` is stored once per
+  // run), so columns stay aligned even when an epoch emitted fewer samples than its neighbours.
+  const columnCount = Math.max(
+    0,
+    ...epochs.map((e) => e.sampleImages?.length ?? 0),
+    epochs.length ? samplePrompts.length : 0
+  );
+
+  const openSample = (epochIndex: number, sampleIndex: number) => {
+    import('~/components/Training/TrainingSampleViewer/TrainingSampleViewer').then(
+      ({ default: TrainingSampleViewer }) => {
+        dialogStore.trigger({
+          component: TrainingSampleViewer,
+          props: { epochs, prompts: samplePrompts, isVideo, epochIndex, sampleIndex },
+        });
+      }
+    );
+  };
+
   // Check if epoch images may be blurred due to buzz type.
   // Images are blurred when paid with green buzz or blue buzz without a membership.
   // Images are NOT blurred when paid with yellow buzz or blue buzz with a membership.
@@ -703,7 +749,6 @@ export default function TrainingSelectFile({
       ? (trainingResults as unknown as TrainingResultsV2)
       : undefined;
   const buzzType = tResultsV2?.transactionData?.find((t) => t.accountType)?.accountType;
-  console.log({ tResultsV2, buzzType });
   // Show alert when NOT paid with yellow buzz (yellow is the only type that guarantees no blur)
   const epochImagesMayBeBlurred = buzzType !== 'yellow';
 
@@ -891,6 +936,9 @@ export default function TrainingSelectFile({
           </Center>
           <EpochRow
             epoch={epochs[0]}
+            epochIndex={0}
+            columnCount={columnCount}
+            onOpenSample={openSample}
             prompts={samplePrompts}
             selectedFile={selectedFile}
             setSelectedFile={setSelectedFile}
@@ -916,6 +964,9 @@ export default function TrainingSelectFile({
                 <EpochRow
                   key={idx}
                   epoch={e}
+                  epochIndex={idx + 1}
+                  columnCount={columnCount}
+                  onOpenSample={openSample}
                   prompts={samplePrompts}
                   selectedFile={selectedFile}
                   setSelectedFile={setSelectedFile}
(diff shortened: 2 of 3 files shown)


JustMaier:
Review addressed in `0be4a6e9b9`. You were right that I had the ragged shape backwards, and that mistake was the root of blocker 1 — thanks for catching it.

## The ragged model, corrected

`.map()` preserves arity, so a failed sample is `''` **in place** (`['a','','c']`), never a missing trailing element. I built the horizontal axis for "shorter array" and left the case that actually occurs unhandled. `hasSampleAt` was already correct; it just wasn't consulted by Left/Right.

**BLOCKER 1** — added `findSampleInEpoch(step)` mirroring `findEpochWithSample`, so Left/Right skip holes exactly like Up/Down. Both nav buttons and both hotkeys now derive from `prevSampleIndex`/`nextSampleIndex` (`undefined` = disabled/no-op), so `url` can no longer become `''` through navigation. The unreachable state is also no longer a silent dead end: `missing` triggers `dialog.onClose()` in an effect instead of returning null, so the dialog leaves `dialogStore` rather than lingering invisibly with a live listener.

**BLOCKER 2** — `src` moved onto `<video>` and the child `<source>` dropped. Resource errors on `<source>` don't bubble, so a 404 video sat on the spinner forever; `onError` on the element itself fires.

**BLOCKER 3** — the scroll container keeps `items-center justify-center` only when *not* zoomed; when zoomed it's a plain `flex ... overflow-auto` and the media gets `m-auto`. `margin: auto` centers without clamping `scrollLeft`/`scrollTop` at 0, so the left/top of a full-size sample is reachable — which is the whole point of the feature.

## Fold-ins

- **"No images available"** now gates on `epoch.sampleImages?.some(Boolean)` — this epoch's own samples — instead of the global `columnCount`. You're right that checklist item 7 was asserting behavior I'd broken.
- **`columnCount`** — dropped the `samplePrompts.length` term. Given `.map()` arity preservation it never helps the real failure mode (holes, not truncation), and it's exactly the term that would inflate every card with placeholders on an in-progress run. Now just the max over epochs.
- **`useHotkeys`** from `@mantine/hooks`, matching `GeneratedOutputLightbox`. Gated on top-of-stack via `dialog.closeOnEscape` (the `isTopLevel` the provider already computes), and `preventDefault` is set per-key to whether that direction is actually navigable, so a no-op arrow falls through to the browser instead of being swallowed.
- **Position readout** counts only non-empty samples in both numerator and denominator, so it agrees with what's navigable.
- **`zIndex`** override removed — `{...dialog}` carries the provider's `300 + index` and stacking works again.
- **`border-gray-4`** paired with `dark:border-dark-4`.
- Dropped two `useCallback`s while I was in there; both functions are called during render, so the memoization bought nothing and one was tripping exhaustive-deps on the `?? []` fallback.

Skipped the `useEffect` load-race and immer-freeze notes as directed.

## Checklist deltas for the reviewer

Item 7 now holds again. Items 12-15 should be re-read against the corrected model — the case to construct is **`['url','','url']` within a single epoch**, not a short array:

- 12': grid shows the dashed "No sample" placeholder in the *middle* column, neighbours stay aligned.
- 13': from sample 0, **Right skips the hole** and lands on sample 2 — it must never blank the viewer, and the modal must stay closable by Escape and the X throughout.
- 14': with holes on both sides, the corresponding buttons are disabled and the keys fall through to the browser rather than being swallowed.
- New: open the viewer, press Right onto what used to be a hole, then Escape — should close cleanly. Then reopen from a thumbnail and confirm only one dialog is live (arrow presses move one step, not two).
- New: zoom a large sample and scroll to its top-left corner — previously unreachable.
- New: a 404 video sample should now show the warning state rather than a permanent spinner.

`pnpm run typecheck` clean (full output), eslint clean on both changed files, prettier applied.

JustMaier:
Final review notes, for whoever tests this.

**Verified safe:** the `dialog.onClose()`-from-an-effect escape hatch does not loop, does not fire on first mount, and has no closing-animation race — this dialog system has no exit transition, `closeById` removes synchronously. Hook order is stable because `useHotkeys` is called before the early return. In practice the escape hatch is now dead code, which is where it should be: entry from the grid only renders a thumbnail button when the url is truthy, and both axes are hole-guarded, so `url` has no reachable path to `''`.

**Two corrections to the test plan:**

1. **Do not use arrow keys to verify the zoom fix.** Arrow-key scrolling acts on the nearest scrollable ancestor of the *focused* element; the zoom pane has no `tabIndex` and focus sits on the Modal close button, so arrows are a no-op there either way. **Use mouse or trackpad to pan.** The fix is real (auto margins resolve to 0 in the overflow direction, so the start edge is reachable, unlike `justify-content: center`) but a tester reaching for arrows will report a false failure.

2. **After Right skips a hole, check the position readout says "sample 2 of 2", not "3 of 3".** That's the fold-in nobody will otherwise notice. And when confirming only one dialog is live after crossing a hole, test **both** Escape and the X button — the original bug killed both.

**Two known cosmetics, not fixed:**
- An epoch whose samples are *all* holes collapses to the one-line "No images available" rather than a row of placeholders, so cards below shift and that epoch drops out of the vertical column alignment. Defensible, but a conscious tradeoff.
- The grid's aria-label uses the raw index while the viewer's readout uses the navigable position, so on `['a','','c']` the grid announces "sample 3" and the viewer says "sample 2 of 2".

**One undocumented coupling worth knowing:** the hotkeys gate on `dialog.closeOnEscape`, which today is the provider's `isTopLevel` proxy and cannot be overridden by a caller (`DialogSettingsBase.options` only carries `transitionDuration` and `onClose`). If anyone later adds a `closeOnEscape` option to `dialogStore`, arrow navigation here goes quiet with no obvious cause.

CivitaiDevOpsAgent:

JustMaier:
Zoom now actually magnifies — `0bce61ca80`. You were right that it was a dead control: both states scaled *down* only, so any sample smaller than the pane rendered identically in each.

## The factor: 2x intrinsic pixels

`width`/`height` are set explicitly to `naturalWidth * 2` / `naturalHeight * 2`.

I picked this over fit-scale x 2 (contain-scale unclamped, then doubled). Fit-scale x 2 always fills exactly twice the pane, which sounds predictable, but it makes the factor depend on window size rather than on the image: a 512px sample on a 4K display would be interpolated up to ~4000px — 8x its real pixel data, which is a blurry mess and actively bad for the fine-detail comparison this exists for. 2x intrinsic never invents more than one extra pixel per axis, so what you see is real sample detail.

Apparent magnification relative to the unzoomed view is `2 / min(1, containScale)`:

| Sample | Unzoomed (in a ~1900x1250 pane) | Zoomed | Apparent |
|---|---|---|---|
| 512px | 512 (contain never upscales) | 1024 | 2.0x |
| 1024px | 1024 | 2048 | 2.0x |
| 2048px | ~1250 (shrunk to fit) | 4096 | ~3.3x |

So it is never below 2x, and samples that had been shrunk to fit get proportionally more back. On your monitor a 1024px sample goes from filling roughly the pane height to overflowing it about 2x in each axis — clearly magnified, and pannable.

## Pushing back on one requirement, deliberately

Requirement 3 said: if the magnified media still fits entirely in the pane, the control is a no-op, so disable it. **That premise held for 1:1 but not for 2x intrinsic, so I did not implement it.**

Unzoomed size is `natural * min(1, containScale)`, which is at most `natural`. Zoomed is `2 * natural`. The zoomed size is therefore *always* strictly larger — there is no configuration where they coincide. A 512px sample in a big window doubles from 512 to 1024 on screen: that fits without overflow, so there's nothing to pan, but it is a real and useful magnification. Disabling the control there would contradict requirement 1 and would take the feature away precisely from the small samples that need it most.

What I did gate on instead is **whether intrinsic dimensions are known at all**: the button, the `cursor-zoom-in`/`cursor-zoom-out` classes, and the click handler are all behind `canZoom = status === 'loaded' && !!natural`. So a still-loading sample and a 404 sample offer no zoom affordance — which was the other way to end up with a dead action, and that one was real.

Happy to add a fits-entirely disable if you'd rather, but I think it's the wrong call and wanted to flag it rather than quietly skip it.

## Video

Video now gets zoom, off `videoWidth`/`videoHeight` in `onLoadedData`, same 2x rule. It's the same comparison use case, and previously it had no zoom at all. One deliberate asymmetry: **video toggles from the toolbar button only, not by clicking the media**, because clicks there belong to the native `controls`. The zoom cursor is likewise image-only, so video never advertises a click that won't happen. The controls bar scrolls with the magnified video, which is standard behavior for zoomed media.

## Panning and reset

Panning is unregressed — the pane still drops `items-center justify-center` when zoomed and the media keeps `m-auto`, so start-side overflow stays reachable. I added `shrink-0` to the media: without it a flex item with an explicit width gets compressed back to the pane width, which would have silently undone the magnification. Zooming in now also centers the pane (`scrollLeft`/`scrollTop` to the midpoint) so you land in the middle of the sample rather than pinned to its top-left corner.

On sample change, `zoomed`, `natural`, `status` and pane scroll are all reset, so zoom state and scroll position never carry onto a differently-sized image. (Zoom was already reset before; scroll and intrinsic size are new.)

Labels updated: **"Magnify 2x"** / **"Fit to screen"**, in both the tooltip and `aria-label`. No slider, no wheel-zoom, no zoom-at-point.

`pnpm run typecheck` clean (full output), eslint clean on both changed files, prettier applied.

## What to check

At your window size, on a real run:

1. Open a sample and hit the magnifier. A 1024px sample should visibly jump to roughly double and overflow the pane in both axes. If it looks unchanged, that's the bug back — tell me the sample's pixel dimensions.
2. Scroll to the **top-left** corner while zoomed. Must be reachable (this is the fix from last round; a bigger image makes clamping more likely, so it's worth re-confirming).
3. Note that zooming lands you **centered**, not at the corner.
4. Zoom in, then press Up/Down to a different epoch. Zoom must drop back to fit and the pane must be scrolled back to origin — no inherited scroll offset on the new image.
5. A small sample (512px) still magnifies to 2x, but may fit without overflow — nothing to pan there is expected, not a bug.
6. Watch a sample load: the magnifier button should be absent until the image lands, then appear. On a 404 sample it should stay absent (warning state only), with no zoom cursor over the pane.
7. Video run: the magnifier appears once the first frame loads and doubles the video. Clicking the video itself plays/pauses rather than zooming — that's intended.

JustMaier:
Zoom/pan persistence and the header counter — `f97bfcc8c6`.

## Zoom is now sticky

`zoomed` is no longer reset on navigation. It stays on until toggled off, so magnified comparison across epochs is one click, not one per step.

## Pan carries when the sample is the same size

Every navigation now goes through a `navigate()` wrapper that captures `scrollLeft`/`scrollTop` **plus the outgoing sample's intrinsic dimensions** before the index changes. After the incoming sample loads, the offset is restored only if its dimensions match; otherwise the pane re-centres, which is what zooming already did. So holding Down while zoomed into the top-left of a face walks that same region down through the epochs — and since Up/Down is the same prompt, dimensions will essentially always match there.

Both the hotkeys and the on-screen nav buttons route through `goSample`/`goEpoch`, so mouse and keyboard behave identically.

## On the ordering hazard you flagged

I restructured the load state rather than patching around it. `status`/`natural` were `useState` cleared by a `[url]` effect; they're now **keyed by url**:

```ts
const [loaded, setLoaded] = useState<{ url; width; height } | null>(null);
const natural = loaded?.url === url ? loaded : null;
const status = natural ? 'loaded' : erroredUrl === url ? 'error' : 'loading';
```

That deletes the reset effect entirely, which removes the race rather than betting on effect-vs-load-event ordering: a cached image whose `load` fires before the reset effect would previously have had its measurements wiped and stranded the viewer on a permanent spinner. Now the measurement simply is or isn't for the current url, and cached and slow samples take the identical path.

Restore ordering is driven off `loaded` changing, never a timer — the effect runs after the re-render that applied the new `zoomStyle`, so `scrollWidth` already reflects the magnified size and can't clamp to 0.

## Degenerate transitions

- **2048 zoomed then navigate to 512**: still zoomed, dimensions differ, so it re-centres. `scrollWidth - clientWidth` goes negative and the browser clamps to 0; `overflow-auto` with no overflow shows no scrollbar, so no stuck track and no blank region.
- **Arrive at a 404 while zoomed**: `natural` is null, so `zoomActive` is false — the pane reverts to `overflow-hidden`, the error state shows centred, and the magnifier disappears. The `zoomed` *intent* is retained, so continuing on to a working sample comes back magnified. The carried offset also survives the dead sample, so stepping past a 404 onto a same-size sample still lands where you were.
- Loading and error states still expose no zoom affordance (`canZoom = !!natural`).

Clamping fix is untouched: pane still drops `items-center justify-center` when zoomed, media keeps `m-auto` + `shrink-0`. Restoring scroll writes `scrollLeft`/`scrollTop` directly and doesn't fight it.

## Header counter dropped

Now just `[Epoch #10]  sample 2 of 3`.

I checked whether `epochNumber` is contiguous from 1, since that's what "epoch 10 of whatever" would need — **it is not.** `saveEvery` is a user-configurable training param (`TrainingParams.tsx:327`, defaulted from target steps in `TrainingSubmitAdvancedSettings.tsx:128`), so a run checkpoints every N epochs and the numbers arrive as e.g. 5, 10, 15, 20. The webhook also falls back to `epochNumber: epoch.epochNumber ?? -1` (`training.service.ts:560`) when the orchestrator omits it. So ordinal position and epoch number genuinely aren't interchangeable, and both the reversed index and the "epoch N of M" phrasing would be wrong. The badge stays the single source. `sample N of M` is unchanged.

`pnpm run typecheck` clean (full output), eslint clean on changed files, prettier applied.

## What to check

**Same-size transition (the main one).** Zoom into a corner of a sample — a face, a hand, some texture. Hold Down through the epochs. That same region should stay put while the image changes underneath; you should be able to read whether the detail improved without touching the mouse. Then Up back through. No re-zooming, no drift.

**Different-size transition.** Needs a run whose samples differ in resolution, or compare across two runs' worth of epochs if you have one. Zoomed on a large sample, navigate to a smaller one: it should re-centre cleanly. Specifically look for a leftover horizontal scrollbar with nothing to scroll, or the image sitting off in a corner with blank space — either would be a bug.

**Left/Right within an epoch.** Different prompts, often same dimensions, so pan will carry. That's intended, though less obviously useful than Up/Down — worth a look to confirm it doesn't feel wrong.

**Sticky-off.** Turn zoom off while panned. Pane should return to fit and scroll to origin, and stay unzoomed as you keep navigating.

**404 while zoomed.** If you have a run older than 30 days, step onto a purged sample while zoomed: warning state, no magnifier, no scrollbar; then step onward to a live sample and confirm it comes back magnified.

**Header.** `Epoch #10  sample 2 of 3` — no "1 of 10 epochs".

PR #3367, opened 2026-07-24
chore(model-files): retire baseModelType and make the file Size control SD 1.x-only

JustMaier:
Retires three SD1-era model-file concepts from the **authoring** UI while leaving them fully readable, rankable, displayable and — where a legacy value already exists — editable. Also fixes a pre-existing production bug affecting 461 files.

All three follow the same shape: **stop offering, keep honoring, and keep legacy values visible and fixable.**

| Concept | New versions | Legacy values |
|---|---|---|
| `size` (Full/Pruned) | Offered on SD 1.x only | Preserved everywhere; editable on SD 1.x |
| `Pruned Model` file type | Never offered | Stays selectable on files that already have it |
| `baseModelType` | Never offered | Select shown when the value isn't `Standard` |

## Guiding principle: stop OFFERING, keep HONORING

Nothing is removed from `constants`, zod enums, file ranking, display surfaces, the generation-coverage SQL, or the public v1 API. This is a UI-authoring change plus one dead-code delete and one bug fix. **No migration. No data change. Nothing backfilled or cleared.**

## Why now

- `size` (Full/Pruned) is set on only **~1.3% of new files** (894 of 70,031 uploaded in July).
- But **1,127 model versions** have two files where `size` is the *only* differentiator, so read/rank/display must keep working forever. That is why nothing on the read path is touched.
- `baseModelType`: 1,170,338 `Standard` vs 617 `Inpainting`, 137 `Refiner`, 56 `Pix2Pix`. Zero generator coupling (every orchestrator hit on that name is a different field, `trainingDetails.baseModelType`). Not in Meilisearch.

## `size` is conditional, not gone: SD 1.x keeps the control

Justin's call: full/pruned "is not a concept anymore" *except* on SD 1.x, where it still is. The predicate is `baseModelHasFileSize()` in `src/shared/constants/basemodel.constants.ts` — it resolves the version's base model through `getEcosystem()` and returns true only for ecosystem key `SD1`.

**SD 2.x was checked against prod data rather than assumed, and excluded.** Counting versions that have two files differing *only* by `size`:

| Base model | Versions where `size` is the only differentiator |
|---|---|
| SD 1.5 | **~639** |
| Flux.1 D | 50 |
| SDXL 1.0 | 20 |
| Pony | 10 |
| Illustrious | 9 |
| **all of SD 2.x** | **5** |

The SD1 ecosystem (`SD 1.4`, `SD 1.5`, `SD 1.5 LCM`, `SD 1.5 Hyper`) accounts for ~645; all of SD 2.x accounts for 5. SD 2.x is indistinguishable from the long tail of modern base models that are also losing the control, so it does not get an exemption. No ecosystem has `parentEcosystemId: ECO.SD1`, so the flat key check has no descendant leakage.

### `size` (Full/Pruned)
- `src/components/Resource/Files.tsx` — the Size select now renders only when the file is a Checkpoint weight **and** `baseModelHasFileSize(baseModel)`. `baseModel` was already on `FilesContext`, no new plumbing.
- `src/components/Model/Actions/MergeVersions.tsx` — same conditional. The base model **is** reachable here (`model.getById` selects `baseModel` on each version), so the target and each source version pass their own `baseModel` down to the file card.
- `src/components/Resource/Files.tsx` — `size` removed as a hard gate in `isMissingRequiredInfo`.
- `src/components/Resource/FilesProvider.tsx` — the `Model size is required for model files` zod refine deleted from `metadataSchema`, plus the now-unreachable `model size` entry in the missing-fields notification and the `size` slot in `SchemaError`.
- `src/components/Model/Actions/MergeVersions.tsx` — the `size: null` wipe on file-type change is gone permanently. That wipe would have destroyed a legacy value, and `null` survives the server-side metadata spread where `undefined` does not.
- `src/components/Account/SettingsCard.tsx` — "Preferred Size" removed unconditionally.
- `src/components/Bounty/BountyUpsertForm.tsx` and `BountyCreateForm.tsx` — "Preferred model size" and the two `details.modelSize` `setValue` calls removed unconditionally. Bounties use the same enum, but a bounty is a request ("what do you want me to make"), not a version, so the SD1 rationale does not apply.
- `src/server/services/model-search.service.ts` — the `size preference -> type: 'Pruned Model'` mapping deleted. **Already dead code**: `getPrimaryFile` reads only `preferences.metadata`, never `preferences.type`.

### `Pruned Model` file type — two separate changes, be clear about which does what

`'Pruned Model'` encodes the same full/pruned split as the `size` metadata, so it retires on the same logic. Two changes ship here and they are **not** equally consequential:

**(a) `retiredFileTypes` in `src/utils/file-display-helpers.ts` — correctness cleanup, zero live effect.** `filterFileTypeByExtension` already omitted `'Pruned Model'` from every explicit extension branch; it survived only through `default: return true` for *unrecognized* extensions. Prod has **461** `Pruned Model` files and every one is on a recognized extension (340 `.safetensors`, 118 `.ckpt`, 3 `.pt`), so this change alters the offering for **zero existing files**. It closes a theoretical path, nothing more.

**(b) Both pickers now keep a file's own current type selectable — this fixes a live bug affecting all 461 files.** Because `filterFileTypeByExtension` omits `'Pruned Model'` on `.safetensors`/`.ckpt`/`.pt`, **all 461 of those files already render a blank Type Select on `main` today.** The risk was never silent corruption — it is a creator opening a legacy version, seeing a blank required-looking field, and manually picking a type to "fix" it, hand-converting a legacy pruned checkpoint. `value === versionFile.type ||` (and `t === effectiveType ||` in the merge modal) makes the file's own type show up as the current selection. It also fixes the same class of problem for any other type/extension mismatch.

For the record, an earlier revision of this description claimed a blank Select could push a null type on save. **That was wrong.** Both pickers are `allowDeselect={false}`, `value` is read from state and never written back from the filtered `data`, the only writer is `onChange`, and `handleSave` sends `type: versionFile.type ?? undefined` — so state still holds `'Pruned Model'` and that is what goes over the wire. In `MergeVersions`, `fileTypeMappings` only gains entries via `updateFileMapping`, so an untouched card contributes nothing. Confirmed empirically: `ModelFile` has **0** rows with `type IS NULL` and **0** with `type = ''` across all of prod.

**Nothing was deleted.** `constants.modelFileTypes`, `modelFileOrder`, the generation-coverage SQL across seven migrations, `src/server/redis/caches.ts`, `event-engine-common/caches/modelData.cache.ts`, `primaryFileTypesByModelType` / `primaryModelFileTypes`, `model-tensor-metadata.ts`, `FileInfo.tsx` and `inferComponentType` all keep it. Existing pruned checkpoints keep their type, keep displaying as "Pruned", and keep resolving as primary downloads.

### `baseModelType` — retired for new versions, still editable where it's set

Originally removed outright. Justin caught the inconsistency on test: a creator with a legacy `Inpainting` version would see a public badge on their model page for a property they could neither see nor change — unfixable if mis-set, unclearable if unwanted. So it now gets the same treatment as the other two.

- `src/components/Resource/Forms/ModelVersionUpsertForm.tsx` — the "Base Model Type" select renders **only when** `version?.baseModelType` is set and is not `'Standard'`. A new version, or any version already on the default, shows nothing. No new authoring capability.
- When shown, the creator can keep the value, switch to another legacy value, or pick `Standard` to clear it. Clearing is a **one-way door** — after the save the select is gone.
- The gate reads the **saved version**, not live form state, so choosing `Standard` doesn't rip the control out from under the user mid-edit. It disappears on the next load.
- Options come from the unfiltered `constants.baseModelTypes`, so a legacy value is always in the list and the Select can never render blank — the exact failure mode fixed above on the file-type picker.
- The pass-through is unchanged (`version?.baseModelType ?? undefined` into `defaultValues`, `data.baseModelType` on submit), so the value still survives a save on versions where the select isn't rendered.

**New versions still land on `Standard`.** `data.baseModelType` is `undefined` for them, and both the create (`model-version.service.ts:509`) and update (`:667`) paths coerce with `?? undefined`, so Prisma omits the column entirely and the `NOT NULL DEFAULT 'Standard'` applies. No null or empty value is reachable on create.

The three display sites are unchanged.

## Two hard requirements had to go, and stay gone

`Files.tsx`'s `isMissingRequiredInfo` treated a missing `size` as a hard blocker for any Checkpoint model weight, and `FilesProvider`'s `metadataSchema` refine rejected publish for the same reason. The Size select was the **only** control that could satisfy either. Removing the select without removing both gates would leave every new checkpoint upload permanently flagged "Needs info" and unpublishable, with no UI to fix it.

Both stay removed **unconditionally, including for SD 1.x**. Offering the field is not the same as requiring it, and re-adding a hard gate is exactly how uploads would get bricked again.

Note for reviewers of the earlier scoping pass: the `FilesProvider` refine is a **second** gate that the original scout did not list. Both were required.

## The fact the `baseModelType` change rests on

`ModelVersion."baseModelType"` in prod is `is_nullable = NO, column_default = 'Standard'`. That matters because the generation-coverage view requires `mv."baseModelType" IN ('Standard')` for Checkpoints — so dropping the `?? 'Standard'` seed from the form would have silently excluded every new checkpoint from generation **if** the `20250702102300_basemodeltype_default` migration had not been applied. It has been applied, so new versions land on `'Standard'` via the column default and stay generation-eligible. Writing it down because that migration is the entire safety margin here.

## Intended consequence: a non-SD1 full/pruned pair is not expressible

`getConflictingFiles` keys on `[size, type, fp, format, quantType]`. On a non-SD1 model there is now no way to distinguish two checkpoint files that differ only by full vs pruned — neither the `size` metadata nor the `Pruned Model` type is offered — so the save is blocked. **This is deliberate.** The concept is SD1-era, SD 1.x keeps the Size control, and a full/pruned pair in a modern ecosystem is almost certainly a mistake.

The duplicate error names every key it actually checks, so on SD 1.x (where the Size control *is* present) setting Size is a visible way out:

> `a.safetensors, b.safetensors: same type, size, format, precision and quant, one must differ`

Detection logic is unchanged; only the wording moved.

## Decisions and known wrinkles

- **"Preferred Size" stays gone.** Existing stored preferences are still honored — `defaultFilePreferences.metadata.size = 'pruned'`, `preferenceWeight.size` and `getPrimaryFile` are all untouched — there is simply no UI to change it anymore. Creators can still publish full/pruned pairs on SD 1.x; downloaders get whatever their stored preference already resolves to. Deliberate asymmetry.
- **Follow-up, not fixed here:** `getPrimaryFile` weights `metadata.size` at 10 but gives no weight to the `'Pruned Model'` **type**, so for a legacy type-based pair a user's stored `'pruned'` preference is not honored.
- **Cosmetic:** `getFileDisplayName` returns `` `${startCase(size)} ${startCase(type)} ${fp}` ``, so a new non-SD1 checkpoint renders with a leading space (`" Model fp16"`). The only consumer is the vault display name and HTML collapses it.

## Deliberately kept (keep honoring)

- `constants.modelFileSizes`, `constants.baseModelTypes`, `constants.modelFileTypes` (incl. `'Pruned Model'`) and `modelFileOrder` — every zod enum and ordering derives from them.
- The enum-locked zod schemas: `model-file.schema.ts`, `model-version.schema.ts` (both `baseModelType` enums and the `fileTypeMappings` `size`), `api/download/models/[modelVersionId].ts`, `bounty.schema.ts`. They accept, they don't require. **Deleting the `baseModelTypes` values would 400 every edit-save on a legacy Inpainting checkpoint** — and now that the select is back for exactly those versions, it would 400 the very screen that fixes them.
- `src/server/utils/model-helpers.ts` — `preferenceWeight.size`, `sortByQuality`'s full > pruned tiebreak, `getFileDisplayName`, `inferComponentType`, and especially `defaultFilePreferences.metadata.size = 'pruned'`. Changing that default would change which file legacy SD1 versions serve by default for every existing user. Frozen.
- `reduceToBasicFileMetadata` in `model-file.service.ts` — still emits `size` on the public v1 API. Removing it is a breaking REST change and a separate decision.
- Generation-coverage SQL, `redis/caches.ts`, `event-engine-common/caches/modelData.cache.ts`, `primaryFileTypesByModelType`, `primaryModelFileTypes`, `getFileLabel` / `getFileDescription` / `createModelFileDownloadUrl` — all untouched.
- `FileInfo.tsx`'s "Model Size" row and the bounty detail page's Model Preferences badge — both guarded, both render nothing when the field is absent, so legacy records keep their badge.
- All three `baseModelType` display sites (`ModelVersionDetails.tsx`, `ResourceSelectCard.tsx`, the PDF template). They already suppress `'Standard'`, so legacy Inpainting/Refiner/Pix2Pix versions keep their badge.
- `checkpointType` (`Trained | Merge`) — a live search facet, untouched and unrelated.

## Verification

- Rebased onto current `main`; zero conflicts, does not revert #3352.
- `pnpm run typecheck` — clean (full output read, not filtered).
- `pnpm run lint` — **85 seconds** post-rebase (#3362's removal of the type-aware override landed on main; before the rebase this branch still carried it and lint did not finish in 10 minutes). 307 repo-wide errors, all pre-existing.
- `pnpm exec eslint` on the changed files — **0 errors**. Remaining warnings are pre-existing.
- `pnpm exec prettier --write` on all changed files.
- Justin tested the first three cases on a preview: no Size selector on non-SD1, Size present on SD 1.5, `Pruned Model` visible only on the file that already carries it.

## Visual checklist for a human

**Base Model Type — the newest change, needs a fresh pass**
- [ ] Open a legacy **Inpainting** (or Refiner / Pix2Pix) checkpoint version: the "Base Model Type" select **is** shown, populated with its current value, not blank.
- [ ] Open a version whose `baseModelType` is `Standard`, and a **brand-new** version: no select at all.
- [ ] On the legacy version, switch `Inpainting` -> `Refiner`, save, reload: the new value sticks and the select is still shown.
- [ ] Switch it to `Standard` and save. Mid-edit the control must **stay put** (not vanish or blank out). After reload it's gone, and the badge is gone from the model page, resource-select modal and PDF.
- [ ] Save that same legacy version **without touching** the select: the value is unchanged.
- [ ] Create a new checkpoint version: it saves with `baseModelType = 'Standard'` (column default), shows no badge, and **is generation-eligible**.

**Legacy `Pruned Model` files — regression test for a bug that exists on `main` today**
- [ ] **Check production first to see the "before".** Open a legacy version with a `.safetensors` file typed `Pruned Model` on prod: the Type Select renders **blank**. 461 files are in this state.
- [ ] On this branch, the same version's Type Select shows **"Pruned Model"**, not blank.
- [ ] `Pruned Model` appears **only** as that file's current selection — it is never offered in the dropdown for a second file on the same version.
- [ ] Change the type to `Model`, reopen the dropdown: the `Pruned Model` option is gone. Hit **Reset**: the type returns and the option is selectable again. (The merge modal has no Reset — recovery there is close and reopen the modal.)
- [ ] The file still surfaces in the primary download card and still labels as "Pruned".

**Model file upload / edit — non-SD1 model** (e.g. an SDXL or Flux Checkpoint)
- [ ] Uploading a new checkpoint `.safetensors` shows **Type / Precision** only — no Size select.
- [ ] The file does **not** show the yellow "Needs info" badge once Type + Precision are set.
- [ ] Publish succeeds; no "Model size is required for model files" error notification.
- [ ] A `.gguf` checkpoint shows **Type / Quant**, no Size, and publishes with just a quant type.
- [ ] The Type dropdown does **not** offer "Pruned Model".

**Model file upload / edit — SD 1.5 model**
- [ ] The Size select **is** present alongside Type + Precision.
- [ ] Leaving Size empty does **not** flag "Needs info" and does **not** block publish. It is optional now, not required.
- [ ] Upload two otherwise-identical checkpoint files to trigger the "Duplicate file types" error, then **set Size on one of them and confirm the error clears** and the save goes through.

**Legacy data safety (any base model)**
- [ ] Editing a legacy version whose files have `full`/`pruned` set: saving a file (rename, type change, precision change) does **not** clear the stored size.

**Model page**
- [ ] A legacy SD1 version with both a Full and a Pruned file still lists both, still labels them, and the default/primary download is unchanged (still Pruned by default).
- [ ] The file info popover still shows the **Model Size** row for a legacy file, on any base model.

**Merge versions modal** (moderator)
- [ ] On an SD 1.5 model the file-mapping step shows a Size select; on an SDXL/Flux model it does not.
- [ ] The Type dropdown does not offer "Pruned Model", but a file already typed that keeps showing it.
- [ ] Changing a file's Type no longer blanks a legacy file's stored size.

**Account settings** (`/user/account`)
- [ ] "Model File Preferences" shows **Preferred Format / Preferred Precision / Preferred Quant** — no "Preferred Size".
- [ ] Existing users' stored `filePreferences.size` is untouched and still drives which file is primary.

**Bounties**
- [ ] Create and edit a `ModelCreation` bounty: the sidebar shows "Preferred model format" only, no "Preferred model size" — including when the requested base model is SD 1.5.
- [ ] An **existing** bounty that has `details.modelSize` still shows its size badge under Model Preferences on the bounty detail page.


CivitaiDevOpsAgent:
### [PR-REVIEW] round 3 — 0 finding(s) (confidence ≥ 0.80)

JustMaier:
Update per Justin: dropped the `Pruned Model` escape-hatch hint and retired the type itself instead. PR body updated in full.

**`Pruned Model` is no longer offered.** Verified the scout's finding: `filterFileTypeByExtension` omits it from every explicit extension branch, and both pickers (`Files.tsx`, `MergeVersions.tsx`) filter through it — so it was reachable only via `default: return true` for unrecognized extensions. Closed with a `retiredFileTypes` check at the top of the function.

Scope of closing that branch: `filterFileTypeByExtension` has exactly **two** consumers, both UI type pickers. It affects nothing beyond the dropdown contents, and the early return touches only `'Pruned Model'` — every other type keeps its existing per-extension behavior.

One thing I added that wasn't asked for, because closing the gap would otherwise regress legacy files: both pickers now keep a **file's own current type** selectable even when it's no longer offered. Without it, opening a version with a `Pruned Model` file would render a blank Type select, and any save from that screen would push a null type. Cheap and it also fixes the same class of problem for any type/extension mismatch.

All read paths verified intact — `constants.modelFileTypes`, `modelFileOrder`, the generation-coverage SQL, `redis/caches.ts`, `event-engine-common/caches/modelData.cache.ts`, `primaryFileTypesByModelType` / `primaryModelFileTypes`, `model-tensor-metadata.ts`, `FileInfo.tsx`, `inferComponentType`. Legacy pruned checkpoints keep their type, keep labeling as "Pruned", keep resolving as primary downloads.

**Duplicate error simplified**, no workaround suggested:

> `a.safetensors, b.safetensors: same type, format, precision and quant — one must differ`

Detection logic untouched. The consequence — a non-SD1 full/pruned pair is now simply not expressible — is documented in the body as deliberate.

**Comment number fixed.** `basemodel.constants.ts` no longer carries a hard count (it was both drifting and undercounting: it cited SD 1.5 alone and omitted `SD 1.5 Hyper` / `SD 1.5 LCM`, which the predicate does cover). Per repo comment policy it now states the rationale only. Also made `fileSizeEcosystemKeys` `readonly`.

Verified: `pnpm run typecheck` clean (full output), `eslint` 0 errors on changed files, `prettier` clean.

JustMaier:
Three fixes from review, plus a correction to my own claim in the description.

**1. My data-loss justification was false — retracted.** I claimed a blank Type Select could push a null type on save. It cannot: both pickers are `allowDeselect={false}`, `value` is read from state and never written back from the filtered `data`, the only writer is `onChange`, and `handleSave` sends `type: versionFile.type ?? undefined`. In `MergeVersions`, `fileTypeMappings` only gains entries through `updateFileMapping`, so an untouched card contributes nothing. Prod confirms it: **0** `ModelFile` rows with `type IS NULL`, **0** with `type = ''`. The description now says so explicitly rather than quietly dropping it.

**2. The other half was understated, and the body now leads with it.** Prod has **461** `Pruned Model` files — 340 `.safetensors`, 118 `.ckpt`, 3 `.pt` — every one on a recognized extension, all of which `filterFileTypeByExtension` already excluded. So **all 461 render a blank Type Select on `main` today**. The split is now stated plainly:

- `retiredFileTypes` = correctness cleanup, closes only the unrecognized-extension path, **zero live effect**.
- self-type preservation = **fixes a pre-existing production bug affecting 461 files**.

The real risk was never silent corruption; it was a creator seeing a blank required-looking field and hand-picking a type to "fix" it, converting a legacy pruned checkpoint by hand.

**3. `size` added to the duplicate error.** `getConflictingFiles` keys on it, and on SD 1.x the Size control is present, so it is a legitimate way out that the copy wasn't offering:

> `a.safetensors, b.safetensors: same type, size, format, precision and quant, one must differ`

Em dash gone too (ASCII in user-facing copy). Took the optional `readonly ModelFileType[]` on `retiredFileTypes` so a typo fails at compile time.

**Checklist updated** with the reviewer's items, including checking prod first for the "before" on the blank Select, the Reset-recovery step (and the note that the merge modal has no Reset, so recovery is close-and-reopen), and the SD 1.5 duplicate-error round trip where setting Size actually clears it.

Verified: `pnpm run typecheck` clean (full output), `eslint` 0 errors on all changed files, `prettier` clean.

JustMaier:
`baseModelType` is now conditional too, so all three retirements share one shape: **stop offering, keep honoring, keep legacy values visible and fixable.**

| Concept | New versions | Legacy values |
|---|---|---|
| `size` (Full/Pruned) | Offered on SD 1.x only | Preserved everywhere; editable on SD 1.x |
| `Pruned Model` file type | Never offered | Stays selectable on files that already have it |
| `baseModelType` | Never offered | Select shown when the value isn't `Standard` |

Justin's catch was right: removing it outright left a creator with a legacy `Inpainting` version looking at a public badge for a property they could neither see nor change.

**Implementation** (`ModelVersionUpsertForm.tsx`):
- Rendered only when `version?.baseModelType` is set and isn't `'Standard'`.
- The gate reads the **saved version, not form state**, so picking `Standard` doesn't rip the control out mid-edit — it stays put through the save and is gone on the next load. That's what keeps the one-way door from becoming a stuck or blank control.
- Options come from the unfiltered `constants.baseModelTypes` (all four), so a legacy value is always in the list. This is exactly the blank-Select failure mode fixed on the file-type picker, avoided by construction rather than by a `value === current ||` escape hatch.
- Pass-through unchanged, so the value survives a save on versions where the select isn't rendered.

**Re-verified the `20250702102300_basemodeltype_default` reasoning.** The column is `NOT NULL DEFAULT 'Standard'` and the generation-coverage view requires `baseModelType IN ('Standard')` for Checkpoints. For a new version `data.baseModelType` is `undefined`, and both the create (`model-version.service.ts:509`) and update (`:667`) paths coerce with `?? undefined` — so Prisma omits the column and the default applies. **No null or empty value is reachable on create**, and new checkpoints stay generation-eligible.

One consequence worth naming: keeping the `constants.baseModelTypes` enum intact is now load-bearing in a second way. Deleting those values wouldn't just 400 an edit-save on a legacy Inpainting version, it would 400 the very screen that exists to fix it.

Verified: `pnpm run typecheck` clean (full output), `eslint` 0 errors on the changed file (4 pre-existing `exhaustive-deps` warnings), `prettier` clean.

The PR body's checklist now leads with a dedicated Base Model Type section — the mid-edit "control must stay put when you pick Standard" step is the one most likely to catch a regression.

PR #3371, opened 2026-07-24
chore(docker): refresh base image to a maintained Alpine

JustMaier:
## ⚠️ Production runtime change — canary required, and a deploy-repo change must land first

This swaps the base image every prod web pod runs on: new OS **and** a new major Node. Two things before it goes anywhere:

1. **A deploy-repo change must be live first** — see Blocking below. Wrong order = every pod crash-loops on boot.
2. **This must ride a Flagger canary.** Not a formality — see "What is and isn't established". The dominant cause of the regression that forced #3311 is identified and fixed, but the residual is unquantified, and **this PR ships the same V8 that #3242 shipped**.

---

## The core justification: we are running an EOL base image

`Dockerfile` pinned `node:22-alpine3.20` in both the `deps` and `runner` stages.

| | |
|---|---|
| `node:22-alpine3.20` last pushed | **2025-05-21** (Docker Hub `tag_last_pushed`) |
| Node it ships | **22.16.0**, frozen |
| Alpine 3.20 EOL | **2026-04-01** (`alpinelinux.org/releases.json`, `v3.20` → `eol_date`) |

Dead on both axes: Alpine stopped branch-building it, and the Node image is no longer rebuilt against it. Production has been carrying roughly 14 months of unpatched Node **and** OS CVEs.

**This CVE exposure is the part of this PR that is not in doubt.** Everything below about Node 24 is a separate judgment call layered on top.

### How it got there

1. **#3242** (2026-07-20) moved us to `node:24-alpine3.22` — current at the time.
2. **#3311** (2026-07-23) rolled the Dockerfile back to `node:22-alpine3.20`, attributing a ~2x per-request server-CPU regression to Node 24 / V8 13.6.
3. That rollback target was **already an EOL, no-longer-rebuilt tag**. Nobody caught it.

### What is and isn't established

**Established.** #3266 ("creator-controls hide-metrics", gated by `model-metric-privacy-readtime`) ran an uncached `dbRead.user.findMany({ select: { settings } })` over every owner in the response on the highest-volume feed paths (`getModelsRaw`, v1-list, associated-models), then deserialized each owner's large settings blob to read three booleans. #3331 fixed it by caching the derived triple per user in packed Redis.

The evidence is a **bracketed ON→OFF→ON Flipt A/B run 2026-07-24 on Node 22**, load-normalized: **+35% CPU/req and +71% longtask attributable to the flag alone**. Node-independent and clean. #3266 definitely caused a large, real regression.

**Not established: that #3266 accounts for _all_ of what #3311 saw. +35% is not +100%.** This PR does **not** claim Node 24 is exonerated. The honest position: the dominant cause is identified and fixed, the residual is unquantified, and quantifying it is what the canary is for.

### On the timing — correcting a claim I had in an earlier draft

An earlier revision of this description asserted #3311's regression window coincided with a deploy that also carried #3266. **That is false and has been removed.** The release tags:

| Change | First release tag | Tagged |
|---|---|---|
| #3242 Node 24 image | v5.0.2128 | 2026-07-20 12:12 CDT |
| #3266 creator-controls | v5.0.2140 | 2026-07-21 16:39 EDT |
| #3311 rollback to Node 22 | v5.0.2154 | 2026-07-23 12:36 CDT |
| #3331 cache fix | v5.0.2164 | 2026-07-24 11:26 CDT |

**Eleven releases and ~27 hours apart.** There was a full day where Node 24 served production traffic *without* #3266 in it. Any argument resting on a shared deploy window is invalid.

### On the profiling evidence

An earlier revision also argued that #3311's profile pointed at #3266 because a JSON-parse storm shows up in app JS frames. **That reasoning was wrong and self-refuting** — #3311 explicitly reported app JS frames as *flat*.

The correct reading: **#3266's cost does not land in app JS frames.** `dbRead.user.findMany` goes through Prisma's native Node-API/Rust query engine on its own threads, and JSON deserialization is a V8 C++ builtin. So #3266's overhead surfaces in native / off-main-JS-thread frames and at the JS↔C++ boundary.

That means #3311's observation — ~100% of growth in native / V8 background / JS↔C++ frames, GC flat, app JS frames flat — is **compatible with both hypotheses and discriminates between neither.** #3311 plausibly read a native-heavy *application* cost as a native-heavy *engine* cost. The profile is not evidence against Node 24; it is simply not evidence either way.

### This PR ships the same V8 — so the canary is the safety net

| Image | Node | V8 |
|---|---|---|
| `node:24-alpine3.22` (what #3242 shipped) | v24.16.0 | **13.6.233.17-node.49** |
| `node:24-alpine3.24` (this PR) | v24.18.0 | **13.6.233.17-node.50** |

Verified via `docker run --rm  node -p "process.versions.v8"`. Two Node patches and one V8 embedder patch — **the same V8 13.6**. If any residual truth remains in the V8 hypothesis, **this PR carries no mitigation for it.** Stated plainly so nobody treats the canary as a rubber stamp.

## The tag: `node:24-alpine3.24`

Checked against the Docker Hub registry API and Alpine's release metadata rather than tag names.

**Alpine branch status** (`https://alpinelinux.org/releases.json`):

| Branch | Branched | Latest patch | EOL |
|---|---|---|---|
| v3.20 | 2024-05-22 | 3.20.10 (2026-04-15) | **2026-04-01 — expired** |
| v3.22 | 2025-05-30 | 3.22.5 (2026-06-21) | 2027-05-01 |
| v3.23 | 2025-12-03 | 3.23.5 (2026-06-21) | 2027-11-01 |
| **v3.24** | **2026-06-09** | **3.24.1 (2026-06-13)** | **2028-06-01** |

**Node image rebuild activity** (Docker Hub `tag_last_pushed`, checked 2026-07-24):

| Tag | Last pushed | amd64 digest |
|---|---|---|
| `22-alpine3.20` (what we ship today) | 2025-05-21 | `sha256:35d1e2d2…` |
| `24-alpine3.23` | 2026-06-24 | `sha256:00295958…` |
| **`24-alpine3.24`** | **2026-06-24** | **`sha256:4ba75f83…`** |

`alpine3.24` is actively rebuilt — `26.5.0-alpine3.24` pushed **2026-07-08**.

**Picked `24-alpine3.24`**: newest actively-rebuilt Alpine, longest EOL runway (**2028-06-01**, ~23 months), and it moves us off #3242's `alpine3.22` so this doesn't recur next quarter.

### Node patch it ships: **24.18.0**

Resolved by digest, not tag name. `24-alpine3.24` shares amd64 digest `sha256:4ba75f835bb8802193e4c114572113d4b26f95f6f094f4b5229d2a77773e0afc` with:

```
lts-alpine3.24  krypton-alpine3.24  24.18.0-alpine3.24  24.18-alpine3.24  24-alpine3.24
```

```
$ docker run --rm node:24-alpine3.24 sh -c 'node -v; cat /etc/alpine-release'
v24.18.0
3.24.1
```

Also the `lts` / `krypton` tag — 24 is the current LTS line.

## Version alignment

| File | Before | After |
|---|---|---|
| `Dockerfile` (`deps`, line 3) | `node:22-alpine3.20` | `node:24-alpine3.24` |
| `Dockerfile` (`runner`, line 113) | `node:22-alpine3.20` | `node:24-alpine3.24` |
| `.nvmrc` | `22.16.0` | `24.18.0` |
| `package.json` → `@types/node` | `22.20.1` | `24.13.3` |
| `package.json` → `engines.node` | `>=22.0.0` | **unchanged** |

- **`.nvmrc` → `24.18.0`** keeps #3365's alignment honest: CI must validate against what production runs.
- **`@types/node` → `24.13.3`** (latest 24.x; the `latest` dist-tag is 26.x and would overshoot the runtime). `pnpm run typecheck` passes **clean, zero errors**.
- **`engines.node` stays `>=22.0.0`** — a floor with no ceiling, satisfied by 24.18.0. Flagged since it *looks* like it should have changed.

## 🚨 Blocking: remove `--experimental-async-context-frame` from `NODE_OPTIONS` BEFORE this image rolls

#3311 re-added this flag when pods went back to Node 22. **It does not exist in Node 24 and hard-fails startup** — reproduced independently by review, **exit code 9, process does not start**:

```
$ docker run --rm -e NODE_OPTIONS="--experimental-async-context-frame" node:24-alpine3.24 node -e "console.log('started')"
node: --experimental-async-context-frame is not allowed in NODE_OPTIONS
$ echo $?
9

$ docker run --rm -e NODE_OPTIONS="--experimental-async-context-frame" node:22-alpine3.24 node -e "console.log('started ok on 22')"
started ok on 22
```

**Dropping the flag is a genuine no-op, not a semantics change.** AsyncContextFrame is default-on in Node 24 — `--help` exposes only the *negated* form:

```
$ docker run --rm node:24-alpine3.24 node --help | grep async-context-frame
  --no-async-context-frame    Improve AsyncLocalStorage performance

$ docker run --rm node:22-alpine3.24 node --help | grep async-context-frame
  --experimental-async-context-frame
```

So we keep exactly the behavior #3116 wanted; we just stop asking for it explicitly.

### Ordering constraint

That setting lives in the **deploy repo**, so it cannot be changed here.

> The flag must be removed from `NODE_OPTIONS` on **all four `civitai-dp-prod` deployments** (SSR, API, jobs, and any others carrying it), and that change **deployed and confirmed live**, **BEFORE** the Node 24 image rolls.

Reversed, every pod crash-loops on boot. #3242 documented the same constraint.

## `apk` packages verified against Alpine 3.24

The root `Dockerfile` installs exactly one package, in `deps` (line 4). `runner` installs none.

| Package | Stage | Status in Alpine 3.24 |
|---|---|---|
| `libc6-compat` | `deps` | ✅ present — provided by `gcompat 1.1.0-r4` in `v3.24/main`, confirmed in the published `APKINDEX` (`p:libc6-compat=1.1.0-r4`) and by live `apk add --simulate` |

```
$ docker run --rm node:24-alpine3.24 apk add --no-cache --simulate libc6-compat
(2/3) Installing libucontext (1.5.1-r0)
(3/3) Installing gcompat (1.1.0-r4)
OK: 11.0 MiB in 21 packages
```

No pinned apk versions, no glibc shims beyond `libc6-compat`, no Alpine-version-specific package names. `addgroup`/`adduser` are busybox built-ins, unaffected.

## Did I build it?

**Partially — and the gap matters, so I'm not glossing it. The image has never been built end to end.**

**Green on `node:24-alpine3.24`:**

`deps` stage — fully green. This is the Alpine-sensitive stage:

- `apk add --no-cache libc6-compat` → OK
- `corepack enable && corepack prepare pnpm@10.28.1` → OK
- `pnpm install --frozen-lockfile` → OK in 50.8s
- `postinstall` → Prisma Client v6.13.0 generated + enum objects, TS interfaces, Kysely types — **native Prisma engines load and execute under musl on Alpine 3.24**

`builder` stage — the compile is green:

```
Finished TypeScript in 3.9min ...
Creating an optimized production build ...
✓ Compiled successfully in 3.2min
Collecting page data using 8 workers ...
```

Full TypeScript pass, Turbopack production compile, sharp, and Prisma engines all work on Node 24.

**Where it stops:** Next's `Collecting page data`, which evaluates app modules and needs a populated **server** environment. I supplied every `NEXT_PUBLIC_*` from `.env-example` and it advanced to the next missing server-side value. Local env configuration, not the runtime.

**Control build.** To confirm the failures weren't mine, I rebuilt `main`'s Dockerfile (`node:22-alpine3.20`) against the same context. **It fails too**, at `Cannot find module 'kysely'`. Cause: `.dockerignore` lists `node_modules`, which Docker matches only at the **root**, so `COPY . .` drags in `packages/*/node_modules` whose pnpm symlinks are absolute host paths — clobbering the nested modules inherited from `deps` and breaking the exact invariant the comment at `Dockerfile:25-31` exists to protect. Harmless in CI (fresh clone), but **local `docker build` is not green on either base image**.

**Net:** every base-image-specific risk (apk resolution, musl native deps, corepack/pnpm, TS + bundler under Node 24) is verified green. The unverified remainder is env-dependent and identical between old and new base images. **I've added the `preview` label so `build-image` actually runs** — none of the default checks (Typecheck, ESLint+Prettier, submodule pin) build the Dockerfile, so that gate had not run.

## Other Dockerfiles — deliberately not touched

**`apps/{auth,creator-studio,event-engine,notifications,orchestrator-gateway,storage}/Dockerfile`** use unpinned `node:22-alpine`. **Not carrying the same exposure:** that floating tag's current amd64 digest is `sha256:b74031e546d7…`, identical to `22-alpine3.24` / `22.23.1-alpine3.24`, pushed **2026-06-23** — already a current, actively-rebuilt Alpine 3.24 with Node 22.23.1. They'd still benefit from an explicit pin and eventually the same roll-forward. Each installs only `libc6-compat`, so the verification above carries over.

**`containers/image-stub/Dockerfile`** uses `node:20-bookworm`. Debian bookworm is supported into 2028, but Node 20 went EOL 2026-04, so it does want attention — build-support container, not a prod runtime, out of scope here.

## Follow-ups (noted, not fixed here)

- **`Dockerfile:3,113` use a floating tag.** `node:24-alpine3.24` resolves to 24.18.0 today, but it will be rebuilt — the base can change with **no diff and no review**, while `.nvmrc`'s `24.18.0` silently drifts, partially undoing #3365. Suggest digest-pinning (`node:24-alpine3.24@sha256:4ba75f83…`) plus an explicit bump cadence.
- **`.npmrc` has `public-hoist-pattern[]=@types/*`**, so `@types/node@24` now hoists to the `apps/*` that remain on Node 22. A Node-24-only API used there would typecheck clean and crash at runtime. Low likelihood, worth recording.
- **`.dockerignore` should use `**/node_modules`** — see the control-build finding above.

## Review checklist

- [ ] **Deploy repo first:** remove `--experimental-async-context-frame` from `NODE_OPTIONS` on all four `civitai-dp-prod` deployments, deployed and confirmed live **before** this image rolls (**blocking — pods will not start otherwise**)
- [ ] `preview` label added so `build-image` runs — the image has never been built
- [ ] **Canary and hold:** watch CPU/req, p95 latency, and pod restarts. #3266 accounts for +35% CPU/req; #3311 reported ~2x. The residual is unquantified and this PR ships the same V8 13.6 as #3242
- [ ] Confirm you accept the roll-forward on that basis, vs. the narrower alternative of `22-alpine3.24` (fixes the CVE exposure only, keeps the Node major frozen)


CivitaiDevOpsAgent:

JustMaier:
## Node 24 is cleanly exonerated, measured

Pulled the CPU/req timeseries from Prometheus (`kube-prometheus-stack`, 15d retention, real scraped data back to 07-13). Runtime boundaries were read from `nodejs_version_info` on the pods rather than assumed from release times.

**Is there a step up at 07-20 17:12 UTC, before #3266? No. There is a step DOWN.**

The result doesn't even need load normalization: traffic in the pre-Node-24 and Node-24-without-#3266 windows was effectively identical (920.8 vs 903.1 tRPC procedures/sec, -1.9%) while **absolute CPU fell from 10.58 to 8.25 cores, -22%**.

Evening-matched (18:30-24:00 UTC daily, 30-min buckets):

| Day | State | median mCPU/proc | vs 07-19 |
|---|---|---|---|
| 07-19 | Node 22, pre-#3266 | 10.89 | baseline |
| 07-20 | **Node 24, NO #3266** | 8.58 | **-21%** |
| 07-21 | **Node 24, NO #3266** (to 20:30Z) | 9.37 | **-14%** |
| 07-22 | Node 24 + #3266 | 31.50 | +189% |
| 07-23 | Node 22 + #3266 | 33.24 | +205% |
| 07-24 | Node 22 + #3266 + #3331 | 14.21 | +30% |

Three independent lines agree:
1. The pre-#3266 window shows an **improvement**, not a regression, sustained across a full 27-hour diurnal cycle. Every hour under Node 24 is below the corresponding baseline hour.
2. The effect survives dropping normalization entirely (raw CPU -22% at equal traffic).
3. **The Node 22 rollback recovered nothing** — 07-23 evening (33.24) is marginally *worse* than 07-22 under Node 24 (31.50). Only consistent with a non-runtime cause.

Queries, re-runnable via `kubectl --context civit-datapacket -n monitoring port-forward svc/kube-prometheus-stack-prometheus 19090:9090`:

```
CPU:  sum(rate(container_cpu_usage_seconds_total{namespace="civitai-dp-prod",container="civitai",pod=~"civitai-dp-prod-api-primary-.*"}[10m]))
LOAD: sum(rate(civitai_app_trpc_procedure_duration_seconds_count{job="civitai-dp-prod-api-primary"}[10m]))
CPU/req = CPU / LOAD * 1000
```

**Consequence for this PR:** routine canary. Flagger's normal gates (99% success, P99 < 5000ms) are adequate; no special abort threshold is justified by this data. The narrower `22-alpine3.24` fallback is no longer needed on risk grounds.

## Two things for elsewhere, not blocking here

**A ~+30% residual is still live.** Today sits at 14.21 vs the 10.89 pre-regression baseline, and longtask corroborates: 1.40 longtasks/1k procedures on 07-19 vs 9.06 today, a 6.5x residual (versus 1.67 under Node 24 without #3266). So #3331 recovered ~85% of the regression, not all of it. It appeared on 07-21 under Node 24, persisted unchanged **through** the Node 22 rollback, and shrank when the app-side cache fix landed, so it is not runtime-related.

The RCA names two more copies of the same anti-pattern that #3331 may not have covered:
- `src/server/services/model.service.ts:3350-3359`
- `src/server/controllers/model.controller.ts:1637-1647`

Caveat: the post-#3331 window is only 5.5 hours, so the +30% will firm up with another day. The Node 24 verdict won't change, since it rests on the complete 07-19 and 07-20/21 windows.

ZacxDev:
## Brought up to date with `main`, and a triage of the red checks

### Merged (not rebased)

Merged `main` into the branch rather than rebasing — this is a shared PR branch, and a rebase would need a force-push that breaks review continuity. The auto-merge was clean, no conflicts.

The net diff against `main` is still exactly the intended four files (+117/−112): `Dockerfile`, `.nvmrc`, the `@types/node` pin, and the lockfile. Nothing else rode along.

Verified on the merged tree before pushing:

- `pnpm install --frozen-lockfile` passes → the auto-merged lockfile is consistent with the auto-merged `package.json`. This was the main place a silent mismerge could have hidden, since both sides touched `package.json`/`pnpm-lock.yaml` (ours: `@types/node`; `main`: version bump + `@civitai/client`).
- `tsc --noEmit` → **0 errors** against `@types/node@24.13.3`, across all the new code that landed while this branch sat. CI Typecheck agrees.

### The 5 smoke failures are all pre-existing — none are Node 24

None of the five are attributable to this branch. On 2026-07-25, **six other PRs** (#3373, #3374, #3375, #3376, #3378, #3379, #3380) — all on the Node 22 base image — reported the *identical* tally `56 passed · 0 flaky · 5 failed` with the *same five test names*. That was simply the smoke baseline that day.

| Test | Classification |
|---|---|
| `direct upload -> attach -> publish creates a real post…` | Pre-existing; still fails today on every PR sampled (8/8) |
| `whatIfFromGraph fires on /generate and returns a numeric cost` | Pre-existing; still fails today on every PR sampled (8/8) |
| `getGenerationData resolves an image with meta into a remix graph` | Pre-existing on 07-25; gone from today's baseline |
| `gold reaches a model detail page and sees a download affordance` | Pre-existing on 07-25; gone from today's baseline |
| `model.upsert -> modelVersion.upsert -> model.publish round-trip` | Pre-existing on 07-25; gone from today's baseline |

The last three dropped out of the baseline after the weekly preview-DB refresh. All three are data-dependent Playwright journeys that resolve a live model/image out of the preview database and then assert on what renders — e.g. the download test pulls `/api/v1/models?limit=1&nsfw=false` and asserts a download affordance on whatever model comes back. That class of test flips with the seed data, not with the runtime.

Today's baseline across recently merged PRs is 2–3 failures, always a subset of these five.

Nothing in this set looks like a runtime incompatibility, and the positive evidence points the same way: the preview builds, boots and serves on Node 24 — render-check 302, auth suite green.

### The two report-only suites are `main` baseline breaks

`unit-tests` and `component-tests` fail on 10/10 recently merged/open PRs, all on Node 22. Reproduced locally against the same environment:

- this branch: **61 failed / 610 passed**
- `origin/main`: **61 failed / 610 passed**
- failing-file sets: **identical**

So that red belongs to `main` (largely the App Blocks router suites), not to this PR. Worth a separate fix, but it is not a gate on this one.

### ⚠️ A green preview is **not** evidence that production is safe on Node 24

This is the one thing the preview genuinely cannot tell us, and it should not be read as a pass.

The preview environment does not reproduce production's `NODE_OPTIONS`. The preview pipeline strips whatever `NODE_OPTIONS` it inherits and substitutes a single minimal `--max-old-space-size`. Production passes a longer set, and that set includes **`--experimental-async-context-frame`**.

**Node 24 rejects that flag outright.** Verified directly against the exact images involved:

```
$ docker run --rm node:24-alpine3.24 node --experimental-async-context-frame -e 'console.log(1)'
node: bad option: --experimental-async-context-frame
→ exit 9

$ docker run --rm node:22-alpine3.20 node --experimental-async-context-frame -e 'console.log(1)'
1
→ exit 0
```

Under production's current runtime options a Node 24 image does not start at all — the process exits 9 immediately. The preview is structurally incapable of surfacing this, because it never sets the flag.

**Ordering constraint:** the deploy-side removal of `--experimental-async-context-frame` must be live *before* an image from this PR rolls. That change is prepared separately in the private infra repo and is deliberately held to land alongside this one. Merging this PR by itself, with the deploy-side flag still in place, is an outage rather than a rollout.

### Base image facts, confirmed

| | Node | Alpine | Image last built |
|---|---|---|---|
| `node:22-alpine3.20` (current) | v22.16.0 | 3.20.6 | **2025-05-21** |
| `node:24-alpine3.24` (this PR) | v24.18.0 | 3.24.1 | 2026-06-24 |

The outgoing image has not been rebuilt in ~14 months, and Alpine 3.20 passed EOL on 2026-04-01 — so it has been accumulating unpatched Node and OS CVEs with no upstream fix path. That is the reason this change matters beyond a routine bump.

ZacxDev:
### Post-merge preview result — the diagnosis held

The re-run on `0b044a4` came back **`58 passed · 0 flaky · 3 failed`**, down from 5 failed.

The three that disappeared are exactly the three I classified as stale-branch artifacts — `getGenerationData…`, `gold reaches a model detail page…`, `model.upsert -> modelVersion.upsert -> model.publish`. Merging `main` fixed them, with no change to the Docker/Node side of this PR.

The three that remain are the current `main` baseline:

- `APPROVE-GATE: submit with NO assets → approve is BLOCKED…`
- `direct upload -> attach -> publish creates a real post…`
- `whatIfFromGraph fires on /generate and returns a numeric cost`

That is the same tally and the same three test names as every other PR that ran today, all of which are on the Node 22 base image:

| PR | Runtime | Smoke |
|---|---|---|
| **#3371 (this one)** | **Node 24** | **58 passed · 0 flaky · 3 failed** |
| #3429 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3427 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3425 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3423 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3420 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3418 | Node 22 | 58 passed · 0 flaky · 3 failed |
| #3416 | Node 22 | 58 passed · 0 flaky · 3 failed |

So this branch is now indistinguishable from the repo baseline on smoke — there are zero PR-attributable failures. The remaining red is `main`'s to fix, on smoke and on the two report-only suites alike.

Pipeline stages: `typecheck`, `build-image`, `deploy-preview`, `verify-preview` all succeeded — the Node 24 / Alpine 3.24 image builds, boots and serves the merged tree.

The `NODE_OPTIONS` caveat in my previous comment still stands and is the only real gate: **the deploy-side `--experimental-async-context-frame` removal has to be live before an image from this PR rolls.**

PR #3342, opened 2026-07-24
feat(telemetry): Prometheus metrics for the image-ingestion path

JustMaier:
## Summary

The entire image-ingestion path (submit → scanner → webhook write-back → cron drain) previously had **zero Prometheus coverage** — Axiom logs only. This adds metrics following the existing prom patterns (`registerCounterWithLabels` / `registerHistogram` / `collect()`-based gauges, prefix `civitai_app_`).

### Metrics added
| Metric | Type | Where |
|---|---|---|
| `image_ingestion_backlog{status}` | Gauge (collect) | `src/server/prom/client.ts` |
| `image_ingestion_oldest_age_seconds{status}` | Gauge (collect) | `src/server/prom/client.ts` |
| `image_scan_webhook_total{result}` | Counter | `src/pages/api/webhooks/image-scan-result.ts` |
| `image_scan_submitted_total{lane,result}` | Counter | `ingestImage()` in `src/server/services/image.service.ts` |
| `image_ingest_cron_total{bucket}` | Counter | `src/server/jobs/image-ingestion.ts` |
| `image_ingest_cron_queue_depth` | Gauge | `src/server/jobs/image-ingestion.ts` |
| `job_duration_seconds{job}` | Histogram | `createJob` in `src/server/jobs/job.ts` |
| `job_errors_total{job}` | Counter | `createJob` in `src/server/jobs/job.ts` |

- **backlog / oldest-age**: only the four non-terminal *working* states (`Pending`, `Error`, `Rescan`, `PendingManualAssignment`).
- **webhook result** labels: `success` / `not_found` / `unscannable` / `error` / `deleted_skip`, covering both the new-lane (`workflowId`) and legacy status-code paths at the webhook boundary.
- **submitted** labels: `lane` = `new` | `legacy`, `result` = `success` | `failed` (all four `ingestImage` return points).
- **cron buckets**: `sentUserPending`, `sentBackfill`, `sentRescan`, `sentError`, `waitingForRetry`, `staleRemoved` (the counts the cron already computes); queue depth is the `ImageScan` JobQueue length it already reads.
- **job_* / job_errors**: fleet-wide in the single `createJob` insertion point, so a dead/erroring `ingest-images` (or any) cron is visible.

## No schema change
[...]

PR #3372, opened 2026-07-25
feat(challenges): go-live and follow notifications

manuelurenah:
Community challenges had no discovery loop: a user who found an upcoming challenge couldn't be reminded when it opened, and following a creator didn't surface the challenges they post.

## What this adds

- **`ChallengeEngagement` table** — per-user "notify me" tracker rows.
- **Four notification types**, all `category: Update` and `toggleable: true` (the existing five challenge notifications concern money/outcomes and stay non-toggleable):
  - `challenge-starting` — inline fan-out from `runChallengeActivation`, fired on the real `Scheduled → Active` transition inside the won-claim branch (the hourly cron lags `startsAt` by up to 59 min, so a time-windowed query would notify before entries actually open).
  - `new-challenge-from-following` — SQL `prepareQuery`, gated on `source='User'`, `ingestion='Scanned'`, status, blocks, and browsing level.
  - `challenge-ending-soon` — SQL `prepareQuery`, fires once on the crossing into the 24h window.
  - `challenge-results` — inline, from **both** winner-announce paths.
- **Entrants are derived at send time** from `CollectionItem` rather than stored, so there's no write-path coupling and it retroactively covers pre-existing entries.
- **Notify toggle** on challenge cards, the context menu, and the detail page.
- **Tracking filter** on Your Challenges (`?engagement=tracking`), not gated on `userChallenges` since System dailies are trackable too.

## Notable correctness work

- Creator-block exclusion is scoped to `source='User'` — System challenges share a judge/bot account as `createdById`, so an unscoped check would have made every daily challenge untrackable for anyone who'd blocked that account.
- `challenge-results` excludes winners **and** participation-prize earners at both call sites, so nobody is notified twice for one outcome.
- The follow query gates on both `Challenge.nsfwLevel` and the cover image level, matching the feed — `nsfwLevel` is `maxValue(allowedNsfwLevel)`, so a challenge admitting X/XXX has a high level even with a PG cover.
- `challenge-ending-soon` skips challenges ≤24h long: for those, "24 hours left" *is* go-live, already covered by `challenge-starting`.

## Verification

- `pnpm run typecheck`: 0 errors.
- Tests: 2052 passed / 14 failed of 2066. All 14 failures are pre-existing and unrelated (3× `model-version.service`, 1× `referral.service`, 2× `ingest-images-cap`, 8× `scan-files` — the last 10 from `prom/client` mocks missing `jobDurationHistogram` since #3342). This branch touches none of those files.
- Both security-relevant gates were verified non-vacuous by break-and-restore.

## ⚠️ Before merge

1. **Apply the migration by hand** — `packages/civitai-db-schema/prisma/migrations/20260724120000_challenge_engagement/migration.sql`, per environment. Every tracking query errors without it. Idempotent, safe to re-run.
2. **`EXPLAIN` both `prepareQuery` bodies on a replica.** The tests assert on the SQL string only; neither query executes in CI. Identifiers were verified by hand against the schema, but a live parse hasn't been possible in this environment.
3. **Check for a `UserEngagement(targetUserId, type)` index.** The PK is `(userId, targetUserId)` and the only index is `(type, userId)`; neither serves this query's join. This adds a second per-minute query of that shape alongside `new-model-from-following`.
4. **`new-challenge-from-following` starts firing on deploy** for existing follows — worth a release note.

Manual QA: the Tracking tab (no test harness exists for `pages/challenges/index.tsx`), logged-out `/challenges`, and the detail-page toggle placement.


CivitaiDevOpsAgent:

PR #3374, opened 2026-07-25
feat(app-listings): add "Autofill from website" button to the external-listing edit form

ZacxDev:
## The gap

On the external-listing **edit** form (`/apps/submit?edit=<listingId>`), the OG metadata pull only fired on a URL **change**. If a mod/dev opened an existing listing without editing the URL, autofill never ran — so a listing whose icon/cover slots ended up empty could never get suggestions re-surfaced, blocking re-population of empty assets.

## The fix

Add an on-demand **"Autofill from website"** button on the edit form's URL step. On click it re-triggers the **existing** `fetchListingMetaFromUrl` query → the existing fill-if-empty effect → the existing `ListingAssetStep` "Use this" icon/cover suggestions — **without changing the URL**. It reuses the same `normalizeLinkUrl` helper the create wizard uses.

- **Non-destructive / only-empty** — by construction: the effect fills only blank name/tagline/description, and asset suggestions render only for idle/empty slots, so typed copy and already-attached assets are never clobbered. No new "clobber" path is introduced.
- **Feedback** — the button is disabled for a blank/invalid URL (no wasted fetch), shows a loading state while fetching, and reports "Nothing to autofill" when the pull yields nothing new.
- **Same-URL re-apply** — a re-click resets the applied ref and bumps a tick the apply effect depends on, so a cached result reliably re-surfaces even when the normalized URL is unchanged.

## Scope

- **Additive** and **dark** — the whole flow is `app-blocks-author` / mod-gated.
- **No server change**, **no create-form change**, **no `ListingAssetStep` change** — edit-form-only.

## Tests

Added browser-mode surface tests (report-only in CI — no local headless Chromium here) to the existing `ExternalSubmitForm.edit.browser.test.tsx`: clicking the button with a URL present pulls meta and surfaces the "Use this" icon/cover affordances for empty slots; the button is disabled for a blank and for a non-https URL; and a pull with nothing fillable shows the "Nothing to autofill" note without clobbering a prefilled name. `tsc --noEmit` clean for the edited files.

## Gates

pr-preview and an adversarial `/audit-pr` are the gates. The mod-gated edit-form interaction still needs a human click-through on the preview.

### DIFF
--- a/src/components/Apps/ExternalListingEditForm.tsx
+++ b/src/components/Apps/ExternalListingEditForm.tsx
@@ -16,6 +16,7 @@ import {
   IconExternalLink,
   IconInfoCircle,
   IconLock,
+  IconSparkles,
 } from '@tabler/icons-react';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
@@ -79,6 +80,12 @@ export function ExternalListingEditForm({ edit }: { edit: ListingEditContext }) 
 
   const [active, setActive] = useState<number>(STEP_URL);
   const [values, setValues] = useState<OffsiteSubmitFormValues>(() => editContextToForm(edit));
+  // Latest `values` for the OG-apply effect's emptiness check (the effect must read
+  // current emptiness WITHOUT depending on `values`, and the async `setValues` updater
+  // hasn't run yet when we compute the button's feedback — same reason the create
+  // wizard computes off `data`).
+  const valuesRef = useRef(values);
+  valuesRef.current = values;
   const [errors, setErrors] = useState<OffsiteSubmitFormErrors>({});
   const [serverError, setServerError] = useState<string | null>(null);
   const [assetsDirty, setAssetsDirty] = useState(false);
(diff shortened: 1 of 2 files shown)


ZacxDev:
Applied the adversarial-audit follow-ups (no 🔴; one 🟡 + two 🟢). All in `ExternalListingEditForm.tsx` + the edit browser test.

**🟡 Error-path flag reset + same-url retry**
- The OG-apply effect now SETTLES on success **or** error (mirrors the create form's `settled = !!data || isError`), adding `isError`/`isFetching` to the deps. On an errored fetch it marks the url handled (`appliedMetaRef`), clears stale suggestions, resets `autofillActive`, and surfaces the failure via a settled `'error'` note **tied to the current url** — replacing the old `autofillActive && isError` display. This kills the two symptoms: (a) a later non-button URL advance no longer renders a spurious autofill note, and (b) the error can't co-render with an "applied" note (single-enum note state).
- 🔴 Same-url re-click now actually retries: `staleTime:Infinity + retry:false` made `setMetaUrl(sameUrl)` a no-op, so a previously-errored url was un-retryable. The handler now calls `metaQuery.refetch()` when the normalized url equals the current `metaUrl`. **No loop** — `refetch()` synchronously flips `isFetching`, so the `autofillTick`-driven effect run bails on the new `if (metaQuery.isFetching) return;` guard and applies only once the (re)fetch settles; the `appliedMetaRef` guard still short-circuits the extra guarded pass.

**🟢 (a) Honest "applied" note** — the note now claims "applied" only when a text field was actually filled **or** a suggestion targets a slot that is EMPTY in the edit prefill (`edit.assets.{icon,cover}.imageId == null` — the only case `ListingAssetStep` renders a "Use this"). A suggestion URL for an already-filled slot yields the "Nothing to autofill" note instead. (Best-effort: if the author added/removed an asset *this* session, `ListingAssetStep` owns that state — noted in the code.)

**🟢 (b) Tests** — added: same-url re-click re-surfaces icon/cover suggestions (tick + `refetch`, asserts refetch fires only on the 2nd/same-url click); errored fetch shows the error note with NO applied/empty note and a re-click refetches; and a suggestion-for-already-filled-slot case that asserts the honest "Nothing to autofill" note. Browser-mode tests are report-only/CI here (no headless Chromium locally).

`tsc --noEmit`: 0 new errors in the edited files (only the pre-existing unrelated `packages/civitai-db*` / `bitdex-stats.ts` / `image.service.ts` / `challenge*.service.ts` errors remain).

⚠️ The mod-gated edit-form click-through (approved-listing shadow path) still needs a human browser check — not verified locally.

PR #3378, opened 2026-07-25
feat(apps): manifest editor — scope selector, hide target slots, listing image upload

ZacxDev:
Improves the onsite App manifest editor (`/apps/[appBlockId]/edit-manifest`). **Two changes** (the originally-planned listing-image upload was split out — see below):

## 1. Scope selector (was a raw textarea)
Replaces the free-text scopes textarea with a new reusable `BlockScopeSelector` checklist built from the canonical scope registry (`BLOCK_SCOPE_TO_OAUTH_BIT` + `SCOPE_DESCRIPTIONS`), visibly flagging **sensitive** scopes (`isSensitiveBlockScope` / `SensitiveScopeBadge`). Per-scope justifications are tied to selection (deselecting a scope drops its justification). An existing manifest's unknown/legacy scope is preserved + removable (never silently dropped). Pure logic in `blockScopeSelection.ts` (17 unit tests).

## 2. Hide "Target slots" + preserve existing targets
The Target-slots editor (slot apps are a deferred feature) is removed from the form. The manifest's existing `targets` are **preserved**: `updateManifest` now **omits** `targets` from the patch entirely, and the server merge (`{...stored, ...patch}`) keeps `stored.targets` verbatim — so a slot app's targets can't be clobbered by a manifest save, and no malformed pre-existing target can fail the save.

## Split out: listing-image upload
The originally-scoped icon/cover/screenshot upload was **descoped** — an onsite `AppListing` is `status:'approved'`, so live-listing asset edits must go through a shadow-revision → mod re-review (the platform trust model), and that consolidation path is currently offsite-only. Onsite listing-media upload is being delivered separately as a dedicated effort (server consolidation + review surface + owner UI).

Dark / mod-gated. pr-preview + an adversarial `/audit-pr` are the gates; the mod-gated editor needs a human preview click-through (scope select/deselect + sensitive flags; Target-slots gone).


CivitaiDevOpsAgent:

ZacxDev:
## Adversarial-audit rework (🔴 + 🟡)

Reworked per an adversarial audit. Changes **#1 (scope selector)** and **#2 (hide target slots)** are unchanged; **#3 (Listing images) is descoped**, and the targets handling is corrected.

### 🔴 Listing-images section — DESCOPED (it was broken for its target users)

The section passed the **live approved listing id** into `ListingAssetStep`. The attach procs (`setIcon`/`setCover`/`addScreenshot`) are author-gated and run `assertOwnerAssetEditable`, which **throws `BAD_REQUEST "This listing is live; edit its assets through a revision instead"`** for a non-moderator owner on an `approved`, non-revision listing — i.e. exactly the auto-created onsite listing (`kind='onsite'`, `status='approved'`). So the feature could never attach for the users it targets.

I checked whether the offsite **shadow-revision** machinery could be reused for onsite (the correct pattern, mirroring `ExternalListingEditForm`):

- ✅ `getMyListingForEdit` / `beginListingRevision` are **row-keyed and kind-agnostic** — they'd happily begin a shadow for an onsite listing (`beginListingRevision` clones `kind: parent.kind`).
- ❌ But the **consolidation path is offsite-only**: `approveExternalRequest` throws `NOT_FOUND` unless `request.kind === 'offsite'`, and the moderator review queues filter `kind: 'offsite'`. `submitListingRevision` stamps `kind: shadow.kind` (`'onsite'`), so an onsite revision would **never surface in the mod queue and could never be re-approved / copied back to the live listing** — the uploads would strand in a never-consolidated draft shadow.

Wiring onsite asset upload correctly is **non-trivial new server work** (an onsite revision review + consolidation path, plus a real submit-for-review UI — not just an asset uploader). Per the audit's second branch, I descoped it from this PR rather than ship a broken or hacky direct-live-edit path. It should get its own PR with proper shadow-revision design + audit.

Descope removed: the section + the `appListingId`/`listingContentRating` additions to `getMyAppManifest` (reverted to base) + the (fully-stubbed) `EditManifestPage` asset test.

**Corrected framing:** the in-UI green alert + the earlier PR-body claim said assets "save immediately and do **not** re-enter moderator review." For an approved app that is **false** — a shadow-revision edit re-enters review. That copy is gone with the section; when onsite upload lands in its own PR it should state that image changes to a live app go through review, like the rest of the listing edit.

### 🟡 targets — OMIT instead of re-send

`preserveTargets(manifest.targets)` re-sent the stored targets into `patch.targets`, which re-validates them against the strict `updateManifest` targets schema (`z.object({slotId: …}).passthrough()`). A pre-existing malformed/empty/oversized stored target would then **fail the whole save** — a regression vs the old textarea path, which never touched `targets`.

Fix: **omit `targets` from the patch entirely.** The server merges `{...stored, ...patch}`, so omitting the key preserves `stored.targets` verbatim — no round-trip, no clobber, no strict-schema re-validation. Removed the now-dead `preserveTargets` helper + its pure test; the browser test now asserts `targets` is absent from the submitted patch.

### Checks
- Pure test `blockScopeSelection.test.ts`: **14/14 pass**.
- `tsc --noEmit`: **0 errors in the edited files** (pre-existing unrelated errors in `image.service.ts` remain).
- Browser tests are report-only in Tekton and weren't run locally (no Playwright chromium in this env); the touched `ManifestEditForm.browser.test.tsx` type-checks clean. Worth a look in the pr-preview / CI browser run.

PR #3381, opened 2026-07-25
fix(model-files): keep file names readable in the file list

JustMaier:
## The report

MNeMiC, with a screenshot:

> "[BUG] I can't see the full file-name, it's getting ...-cropped, with no tooltip for full name. Did this window get thinner? I have no idea which is which, so now I have to size-compare, which will not always work."

Follow-up: *"Yup, the 'Manage Files' (add more models) is wider."*

His workaround is comparing file **sizes** to tell variants apart, which is exactly the disambiguation problem we've been fixing elsewhere.

## What I found about the width difference

The original diagnosis had it backwards. The two surfaces do differ, but the **dialog was already the wider one**, which matches MNeMiC's own follow-up:

| Surface | Constraint | Content width |
|---|---|---|
| `?dialog=filesEdit` ("Manage Files") | was `<Container size="md">` in a `fullScreen` Modal | Mantine md 960px minus padding, about **928px** |
| Wizard upload step (`/wizard?step=2`) | `<div className="container flex max-w-sm ...">` | our Tailwind `sm` = 768px minus `container` padding, about **736px** |

Both wizards (`ModelVersionWizard.tsx:120`, `ModelWizard.tsx:193`) cap the upload step at `max-w-sm`; there is no unconstrained surface. Our Tailwind `screens`/`maxWidth` come from `src/utils/tailwind.ts`, where `sm = 768px`, narrower than Mantine's `md` container. So the surface MNeMiC was cropping on is the **wizard**, and "Manage Files is wider" is literally true.

## What changed

**`FilesEditModal.tsx`: the dialog's `Container` went `md` to `xl`.** A deliberate width change to a shared surface. Mantine's `--container-size-xl` is **1320px**, so content goes from ~928px to roughly **1288px**, a substantially bigger gain than the `md` cap allowed. Fully fluid was tried first and looked wrong on a wide display, so the container stays; `xl` is the compromise.

Padding also changed: the file-list `Stack` moved from `p="md"` to `pt="md"`. `Card.css` restores exactly `var(--card-padding)` (16px) via `Card.Section [data-inherit-padding]`, the `Stack` is not a `Card.Section` so the `:last-child` negative margin does not apply, and the Card's own bottom padding survives. So this removes doubled padding without letting any child touch an edge, including the empty states and dropzones.

**`Files.tsx`: reworked the `FileCard` row so the name isn't the first thing sacrificed.** Extra width helps at MNeMiC's monitor size but doesn't fix the underlying layout, so the row changes are the substantive part:

1. **Tooltip carrying the full name**, so truncation is never a dead end at any width. It is `multiline` with `maw={420}` and `overflow-wrap: anywhere`. Mantine tooltips are `white-space: nowrap` unless `multiline`, and an unbroken 120-char filename would otherwise render *outside* the tooltip's own background, as white text on the page. It is also `disabled` unless the name is actually cut off, since Mantine has no truncation awareness and would otherwise fire on every row.
2. **"Needs info" badge moved off the name line** down to the metadata line next to size and format. The badge held its width while the name collapsed; the metadata line is short, dimmed, and has room to spare.
3. **The row now wraps.** The outer `Group` and both metadata-select `Group`s dropped `wrap="nowrap"`, and the name column is `flex: '1 1 220px'` with `minWidth: 0`. Previously the name column could shrink to zero to keep the fixed-width Type/Precision/Quant/Size selects (up to ~525px of them) on one line. Flex line-breaking uses the hypothetical main size, so the `220px` basis makes the selects wrap at the same threshold a `min-width` floor would, but the column can still shrink below 220 when it's alone on a line rather than clipping silently inside the `overflow: hidden` Card on very narrow screens. The icon now sits in the same nested `Group` as the name so it travels with it. At wide widths the layout is unchanged.
4. **Dropped the `marginTop: 18` on the rename and delete icons.** Justin confirmed on mobile that they sat "extra spaced out down at the bottom" once the row wrapped. That offset approximated a Mantine input label's height to line the icons up with labeled selects on a single `nowrap` line, but their parent `Group` is already `align="flex-end"`, which bottom-aligns them against the selects on its own. So the offset was doing nothing on one line and only added dead space above the icons once they could wrap onto a line of their own. Removed rather than made conditional; no media query, which would be the wrong tool here since this component renders at ~1288px in the dialog and ~736px in the wizard at the same window size.

**`useIsOverflown` (new, `src/hooks/useIsOverflown.ts`), and a second surface, `AuctionPlacementCard.tsx`.** Flagging that file explicitly since it sits outside the model-files title and the change was deliberate. `AuctionPlacementCard`'s `OverflowTooltip` already measured `offsetWidth < scrollWidth` inline to gate a tooltip, carrying a `// TODO this doesnt appear to listen for changes when resizing`. Rather than copy that logic into `Files.tsx`, it is lifted into a shared hook that observes the element with a `ResizeObserver`, and both call sites use it. That **closes the TODO** as a side effect, at a net -9 lines in the auction file.

The hook also takes a deps passthrough, used as `useIsOverflown([displayName])` and `useIsOverflown([label])`. `ResizeObserver` fires on box changes, not content changes, so without it a rename through the Popover (a new `overrideName` gives a new `displayName` at the same column width) would leave the measurement stale: short to long would mean no tooltip on a name that now truncates. Same hazard for the auction card's `label` when data refetches into a mounted card.

## Visual checklist

- [ ] **~1288px (the dialog at `xl`)** with a long filename. Names should not truncate at all. This is the actual bug report and it has only been eyeballed at one window size.
- [ ] **~500-700px with a `.gguf` component file**, the fullest select set (Type + Quant + Precision) and the tightest wrap case.
- [ ] **The wizard at 768px** (`/models/[id]/model-versions/[versionId]/wizard?step=2`), including a **mid-upload row** (progress bar section) and a **failed-upload row** (red name).
- [ ] **The rename Popover open while the row is wrapped** (a Trained model, `showRenameOnPrimary`).
- [ ] **Rename a file short to long in Manage Files, then hover.** The stale-measurement case the deps passthrough fixes.
- [ ] **A row with a select validation error while wrapped**: red Card border plus the inline `Select` error text.
- [ ] **A row showing the "Needs info" badge**: a `.gguf` with no quant type, or a Checkpoint `Model` file with no precision. Badge should sit on the metadata line, not compete with the name.
- [ ] **Both empty states** ("No additional components yet", and the large primary dropzone) at the new 16px inset.
- [ ] **Hover one genuinely long name** and look at the wrapped tooltip at ~420px. Confirm it *looks* right, not merely that it doesn't overflow. **Hover a short name**: no tooltip.
- [ ] **Phone width, wrapped row.** Justin confirmed the stranded-icon state here before the fix, so this is a re-check rather than an unknown: the rename and delete icons should now sit tight against the selects above them with no gap.
- [ ] **One glance at `/auctions`.** Each card renders two `OverflowTooltip`s and that page can hold hundreds of unvirtualized cards, so ~800+ ResizeObservers where there were none. Expected to be fine: RO batches post-layout, nothing writes to the DOM between reads so there's no forced-reflow thrash, and React 18 batches the `setState`s. But it's a new per-card cost on a page with a known perf history (#3260), so worth seeing once rather than assuming.

## Verification

- `pnpm run typecheck`: clean, full unfiltered output.
- `pnpm exec eslint` on all changed files: no issues.
- `pnpm exec prettier --write` on all changed files.

## Known follow-ups, deliberately not in this PR

- **Metadata line at a ~180px name column.** The size/format `Text` wraps to two lines and the Badge ellipsises to "Needs...". `flexShrink: 0` on the Badge plus `whiteSpace: 'nowrap'` on the `Text` would fix it.
- **The wizard is still the narrowest surface** at 736px. Raising `max-w-sm` on the upload step in `ModelVersionWizard.tsx` and `ModelWizard.tsx` is a one-line change, left out because it desyncs that step from the rest of the stepper, which is uniformly `max-w-sm`.
- **Two measurement edge cases inherited from the old inline code**, judged not worth changing: `scrollWidth` is integer-rounded, so a sub-pixel clip can render an ellipsis with the tooltip disabled; and a webfont swapping in after mount changes text metrics without resizing the box.

### DIFF
--- a/src/components/Auction/AuctionPlacementCard.tsx
+++ b/src/components/Auction/AuctionPlacementCard.tsx
@@ -229,7 +229,7 @@ const OverflowTooltip = ({
   searchText,
   ...highlightProps
 }: { label: string; searchText: string } & Omit<HighlightProps, 'highlight' | 'children'>) => {
-  const { ref, overflown } = useIsOverflown<HTMLDivElement>();
+  const { ref, overflown } = useIsOverflown<HTMLDivElement>([label]);
 
   return (
     <Tooltip label={label} disabled={!overflown} withinPortal multiline>
--- a/src/components/Auction/AuctionPlacementCard.tsx
+++ b/src/components/Auction/AuctionPlacementCard.tsx
@@ -48,6 +48,7 @@ import { useScrollAreaRef } from '~/components/ScrollArea/ScrollAreaContext';
 import { UserAvatar } from '~/components/UserAvatar/UserAvatar';
 import { useCurrentUser } from '~/hooks/useCurrentUser';
 import { useIsMobile } from '~/hooks/useIsMobile';
+import { useIsOverflown } from '~/hooks/useIsOverflown';
 import { useBrowsingSettings } from '~/providers/BrowserSettingsProvider';
 import type {
   GetAuctionBySlugReturn,
@@ -228,22 +229,12 @@ const OverflowTooltip = ({
   searchText,
   ...highlightProps
 }: { label: string; searchText: string } & Omit<HighlightProps, 'highlight' | 'children'>) => {
-  const textElementRef = useRef<HTMLDivElement>(null);
-  const [isOverflown, setIsOverflown] = useState(false);
-
-  // TODO this doesnt appear to listen for changes when resizing
-  useEffect(() => {
-    const element = textElementRef.current;
-    const compare = element
-      ? element.offsetWidth < element.scrollWidth || element.offsetHeight < element.scrollHeight
-      : false;
-    setIsOverflown(compare);
-  }, []);
+  const { ref, overflown } = useIsOverflown<HTMLDivElement>();
 
   return (
-    <Tooltip label={label} disabled={!isOverflown} withinPortal multiline>
+    <Tooltip label={label} disabled={!overflown} withinPortal multiline>
       <Highlight
-        ref={textElementRef}
+        ref={ref}
         className="min-w-0 max-w-[min(400px,80vw)] truncate"
         highlight={searchText}
         {...highlightProps}
--- a/src/components/Resource/Files.tsx
+++ b/src/components/Resource/Files.tsx
@@ -294,7 +294,7 @@ export function Files({ showRenameOnPrimary }: { showRenameOnPrimary?: boolean }
             preferences.
           </Text>
         </Card.Section>
-        <Stack gap="sm" p="md">
+        <Stack gap="sm" pt="md">
           {modelFiles.length > 0 ? (
             <>
               {modelFiles.map((file) => (
@@ -369,7 +369,7 @@ export function Files({ showRenameOnPrimary }: { showRenameOnPrimary?: boolean }
             Components and files that accompany this model. Mark each as required or optional.
           </Text>
         </Card.Section>
-        <Stack gap="sm" p="md">
+        <Stack gap="sm" pt="md">
           {!hasAdditionalContent && (
             <Stack gap="xs" align="center" py="md">
               <IconLayersLinked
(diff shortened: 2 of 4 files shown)


CivitaiDevOpsAgent:

JustMaier:
Review fixes pushed as `04daf58df6`.

**D1 — tooltip.** Both problems fixed, and via the shared-hook route rather than the minimum. New `src/hooks/useIsOverflown.ts` returns `{ ref, overflown }`, measuring `offsetWidth < scrollWidth || offsetHeight < scrollHeight` and re-measuring under a `ResizeObserver`. `Files.tsx` uses it for `disabled={!overflown}`, and `AuctionPlacementCard`'s `OverflowTooltip` is refactored onto it — which closes its `// TODO this doesnt appear to listen for changes when resizing`, since the hook observes the element instead of measuring once on mount.

The tooltip is now `withinPortal multiline maw={420}`. I added `style={{ overflowWrap: 'anywhere' }}` beyond what was asked: `multiline` only relaxes `white-space` to `normal`, and filenames are single unbroken tokens (`_`, `-`, `.` are not break opportunities under `word-break: normal`), so `maw` alone would have been overflowed by the exact 120-char name the fix is for.

**D2 — min-width floor.** Now `style={{ flex: '1 1 220px', minWidth: 0 }}` as suggested. Confirmed the wrap threshold is unchanged — flex line-breaking uses the hypothetical main size, which the basis supplies — while the column can shrink below 220 when alone on a line, so nothing clips inside the `overflow: hidden` Card at Fold-cover widths.

**D5 — PR body.** Rewritten. It now says the Container was reinstated at `xl` rather than dropped, states 1320px / ~1288px content instead of the wrong ~1200px, and documents the `p="md"` → `pt="md"` change with the `Card.css` reasoning from the review.

**Checklist.** Replaced with your sharper list — ~1288px with long names (the actual report, only ever eyeballed at one size), ~500-700px with a `.gguf` for the fullest select set, the wizard at 768px including mid-upload and failed-upload rows, the rename Popover open while wrapped, a select validation error while wrapped, and both empty states at the new 16px inset.

**D3 and D4** are recorded in the PR body under known follow-ups, with D3 flagged as the state most likely to look broken during review.

`pnpm run typecheck` clean (full output, unfiltered); `eslint` clean on all four changed files; prettier applied.

JustMaier:
Pushed `6a09971a35` (deps passthrough) and `f1125d13a8` (D3).

**Stale measurement.** `useIsOverflown` now takes `deps: unknown[] = []` and passes it straight to the effect, so a content change re-measures even when the box doesn't move. Wired as `useIsOverflown([displayName])` in `Files.tsx` and `useIsOverflown([label])` in `AuctionPlacementCard`. Kept a one-line comment on the `exhaustive-deps` disable explaining why the array is a passthrough rather than a literal, since that's the non-obvious bit.

**D3, folded in.** You were right to say look at whether the margin is needed at all: it isn't. The parent is `<Group gap="xs" align="flex-end">` and both icons are direct flex items of it, so `align-items: flex-end` already bottom-aligns them against the taller labeled selects. The `marginTop: 18` was inert on the single-line case (a flex-end item's *margin box* bottom is what aligns, so a top margin doesn't shift it) and only ever materialized as dead space once the icons could wrap onto their own flex line, which is exactly what Justin saw. Deleting both was the whole fix, two lines, no media query and no restructuring.

Note the same `marginTop: 18` exists on `LinkedComponentCard`'s trash icon at `:542`, and I left it: that `Group` has no `align`, so it defaults to `center` and the offset is genuinely doing work there. Different component, unchanged by this PR.

**Body and checklist.** Added the `AuctionPlacementCard` paragraph naming the file and the TODO it closes, plus a fourth bullet under the `Files.tsx` changes covering the icon offset. The wrapped-icon checklist item is reworded as a re-check rather than an unknown, since Justin confirmed the bad state on mobile. D3 is out of the follow-ups list; D4 and the two inherited measurement edge cases stay there.

`pnpm run typecheck` clean (full output, unfiltered); `eslint` clean on all changed files; prettier applied.

PR #3386, opened 2026-07-25
perf(models): push down my-models modelVersion count (~1344ms → ~1.5ms)

ZacxDev:
## Problem

The "my models" / "my training" pages (`getMyDraftModelsHandler`, `getMyTrainingModelsHandler` in `src/server/controllers/model.controller.ts`) each included:

```ts
_count: { select: { modelVersions: true } }
```

Prisma compiles that per-row aggregate into a `LEFT JOIN` against a subquery that **GROUPs over the ENTIRE `ModelVersion` table**. Postgres does **not** push the `Model.id IN (…)` filter into the grouped derived table, so it aggregates all **~1.17M `ModelVersion` rows → ~909k groups, ~1.2M shared buffers, ~1344ms** just to attach a version count to the ~20 models on the page.

EXPLAIN-proven on the read replica; across the 4 IN-list-length variants: **~78k calls / ~57k s replica time** over the stats window, **~734ms mean**.

## Fix

Drop the `_count` include, then after fetching the page run **one batched, pushed-down count** keyed on the returned model ids:

```ts
dbRead.modelVersion.groupBy({
  by: ['modelId'],
  where: { modelId: { in: ids } },
  _count: { _all: true },
})
```

The `IN` filter **does** push into the aggregate, so this is **index-only** on the existing `ModelVersion_modelId_baseModel_idx` — **no new index**. EXPLAIN-verified: **~1344ms → ~1.5ms (~900×)**.

Extracted into `getModelVersionCountsByModelId` (single db dependency) so both handlers share it and it is unit-testable in isolation.

## Byte-identical response shape

Each model keeps its `_count.modelVersions` number exactly as before — the reshape re-attaches `_count: { modelVersions }` per model. Consumers (`UserDraftModels.tsx`, `UserTrainingModels.tsx`) read `_count.modelVersions` as a number; unchanged.

## Semantics preservation

- The replaced `_count` include had **no filter** → it counted **ALL** modelVersions of the model. The `groupBy` counts the same (`COUNT(*)` over `modelId`, no status/publishedAt predicate). Verified by reading both call sites — no filtered subset was implied.
- Models with **zero versions** are omitted by `GROUP BY`; the helper **zero-fills** the requested ids so they still return `_count.modelVersions: 0`.
- Training handler: multiple returned training versions can share one model — the ids are **deduped** before the query.

## Tests

New `src/server/services/__tests__/model-version-count.service.test.ts` (5 cases): pushed-down query shape (id filter in the aggregate, not a relation `_count`), real counts attach, **zero-version model coalesces to 0**, id dedupe, and empty-list short-circuit (no DB hit).

## Verification notes

- `NODE_OPTIONS=--max_old_space_size=8192 tsc --noEmit`: **0 errors**.
- Unit test: **5/5 pass**.
- The ~1344ms→~1.5ms EXPLAIN evidence is from the live **read replica**; a local Prisma/DB run was not exercised.

### DIFF
--- a/src/server/controllers/model.controller.ts
+++ b/src/server/controllers/model.controller.ts
@@ -78,6 +78,7 @@ import { hasEntityAccess } from '~/server/services/common.service';
 import { getDownloadFilename, getFilesByEntity } from '~/server/services/file.service';
 import { getImagesForModelVersion } from '~/server/services/image.service';
 import { bustMvCache, getLinkedVaeIds } from '~/server/services/model-version.service';
+import { getModelVersionCountsByModelId } from '~/server/services/model-version-count.service';
 import {
   copyGallerySettingsToAllModelsByUser,
   deleteModelById,
@@ -1204,6 +1205,13 @@ export const getMyDraftModelsHandler = async ({
     const results = await getDraftModelsByUserId({
       ...input,
       userId,
+      // NOTE: `_count: { select: { modelVersions: true } }` is intentionally NOT
+      // selected here. Prisma compiles that per-row aggregate into a LEFT JOIN
+      // against a subquery that GROUPs the ENTIRE ModelVersion table (Postgres
+      // does not push the `Model.id IN (...)` filter into the grouped derived
+      // table) — ~1.17M rows -> ~909k groups -> ~1344ms just to attach a version
+      // count to the ~20 requested models. We instead attach the count below via
+      // one batched, index-only groupBy whose `IN` filter DOES push down.
       select: {
         id: true,
         name: true,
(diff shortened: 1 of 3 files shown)


CivitaiDevOpsAgent:

ZacxDev:
Applied the three audit fixes (commit 3fe4757d39):

**1. 🟡 Training count now reads the same db handle as the list (destructive-path consistency).** The training list routes through `getDbWithoutLag('userTrainingModels', userId)` (primary during the post-write no-lag window), but the count helper only read the replica — so after adding a 2nd training version the list showed 2 while the lagged count said 1. `UserTrainingModels.tsx:318` gates a destructive delete on `_count.modelVersions > 1` (`>1` deletes one version, otherwise `handleDeleteModel` deletes the **whole model**), so a stale count of 1 would delete the entire model. Fix: `getModelVersionCountsByModelId` gains an optional `db` param (default `dbRead`, so draft behavior is unchanged); the training handler calls `getDbWithoutLag('userTrainingModels', userId)` with identical args and passes that same in-request handle in.

**2. 🟡 Draft handler drops the extra groupBy.** Its select already fetches the full unfiltered `modelVersions` array (the frontend uses `.some(...)`), so `model.modelVersions.length` is the total count at zero DB cost. Helper is kept for the training handler only, whose versions array is status-filtered (Draft/Training) — `.length` there would undercount.

**3. 🟢 Corrected the comment.** Dropped the inaccurate "index-only on `ModelVersion_modelId_baseModel_idx`" claim — the schema has `@@index([modelId], type: Hash)`, and a hash index can't serve an index-only scan (nor is that named btree in the schema). Reworded to describe the query as a pushed-down `modelId IN(...)` aggregate that probes a bounded id list.

Added tests for the `db`-param default + a caller-supplied handle. `tsc --noEmit`: 0 errors in `src/` (the 4 remaining `Cannot find module 'kysely'` errors are pre-existing, in `packages/civitai-db*`, unrelated). Helper suite: 7/7 pass.

PR #3388, opened 2026-07-26
fix(redis): debounce cluster routing-retry rediscovery nudge (flag-gated, default off)

ZacxDev:
## Summary

Debounce the cluster **routing-retry rediscovery NUDGE** — not the retry — behind a flag that is **off by default** (byte-identical to today until explicitly enabled).

`packages/civitai-redis` is shared cluster infra; every cluster user pays for changes here, so this is scoped, kill-switchable, and measurable. **Draft — do not merge; rolling out canary-first.**

## Root cause

On the redis-heaviest pool under high command volume, the node-redis cluster client's slot map momentarily empties during a rebuild. `getSlotRandomNode` (`@redis/client` `cluster-slots.js`) then throws a **pre-dispatch** routing error, and `withClusterRoutingRetry` (`cluster-routing-retry.ts`) fires an **unconditional** `triggerTopologyRediscovery(self, 'routing-retry')` nudge (`client.ts`, the `rediscover` hook) on **every** throw.

node-redis already single-flights the *actual* `_slots.rediscover()`. So the extra cost is the **scheduling / nudge churn** (`#rediscover`) — measured at ~**3.9% of active-JS-CPU** on the busiest pod — and that churn keeps re-opening the empty-slot window, producing **more** throws: a self-feeding loop.

## The fix: debounce the NUDGE, not the retry

**The retry is load-bearing and stays unchanged.** `withClusterRoutingRetry` still retries every transient pre-dispatch routing throw and converts throws-that-would-become-500s into `recovered`. This PR changes **only how often the routing-retry path *schedules* a rediscovery**:

- New pure helper **`createNudgeDebouncer(debounceMs, now?)`** in `cluster-routing-retry.ts`: returns `(fire) => void` that calls `fire()` only if at least `debounceMs` have elapsed since the last fired nudge (first nudge in a storm always fires; the rest inside the window are skipped). Clock is injectable → unit-testable.
- In `instrumentCommands` (`client.ts`) **one debouncer is created per cluster client** (the `_execute` chokepoint is wrapped once) and **shared across all concurrent commands**. The `rediscover` hook becomes `() => nudgeDebouncer(() => triggerTopologyRediscovery(self, 'routing-retry'))`.

Net: under a throw storm the client schedules **at most one rediscover per window** instead of one per throw, while **every command still retries and recovers exactly as before**.

### Why it's a no-op when disabled

`debounceMs <= 0` (the **default**) makes `createNudgeDebouncer` return a pass-through that fires the nudge **every** call — behaviorally identical to the code before this gate existed. So the change is inert on merge and only takes effect where the deployment env sets it > 0.

### Retry / recovery semantics preserved

The predicate (`isTransientClusterRoutingError`), the pre-dispatch write-safety argument, the bounded backoff, the `recovered`/`exhausted` accounting, and the re-throw-the-original behavior are all untouched. The debouncer sits **only** on whether the fire-and-forget nudge lands; it never changes attempt counts, backoff, or outcomes. (Confirmed by test `(c) with debounce DISABLED the nudge fires on every retry`, and by the storm test where all commands recover regardless of nudge suppression.)

## Flag / env

`REDIS_CLUSTER_REDISCOVER_NUDGE_DEBOUNCE_MS` (added to the package env schema alongside the existing `REDIS_CLUSTER_ROUTING_RETRY_*` vars, threaded through `RedisConfig` → `instrumentCommands` exactly like the other routing-retry knobs).

- **Default `0` = disabled = byte-identical to today.**
- Enable per-pool via the deployment env (roll out to the heaviest pool first). A value on the order of the rediscover settle window (~500–1000 ms) collapses a storm to ~1 nudge/window; tune per-pool from the canary.
- **Rollback = set the env back to `0`** (restores today's behavior) — no code revert.

## Tests

Added to `cluster-routing-retry.test.ts` (non-vacuous):
- **(a)** debounce OFF/0 → nudge fires on **every** throw (unchanged; the disabled path is also proven to never read the clock).
- **(b)** debounce N ms → **at most one** nudge per window under a burst; first fires, in-window suppressed, boundary at `>= debounceMs`; a 50-throw sustained stream over one window → exactly 1 rediscover.
- **(c)** integration: a storm of 20 commands sharing one debouncer at a frozen clock → **all 20 retry + recover** (`recovered` ×20) while the real rediscover fires **once**; next window → one more. Plus separate-instance isolation.

Verification (in the changed package):
- `vitest run` → **132/132 pass** (38 in the routing-retry file, incl. 13 new).
- `NODE_OPTIONS=--max_old_space_size=8192 tsc --noEmit` scoped to `packages/civitai-redis` → **0 errors**.

## Canary / measurement plan

Roll out **heaviest-pool-first** via the deployment env.

**Success** (on that pool, at peak):
- `civitai_app_redis_selfheal_reconnect_total` rate **falls**, and the cluster **deadline-hit** rate **falls**.
- The `#rediscover` / `getSlotRandomNode` CPU share **drops** (before/after CPU profile on a busy pod, normalized to active-JS-CPU).
- The routing-retry **`exhausted`** rate (→ user 500) does **not** rise — the retry must remain as effective as today.

**Rollback:** set `REDIS_CLUSTER_REDISCOVER_NUDGE_DEBOUNCE_MS=0` (no code revert needed).

## Blast radius

Shared `packages/civitai-redis` (cluster client only; the single-node sys client is never touched). Default-off → zero behavior change on merge. Cluster-scoped; no money/entitlement path. Runtime effect at `debounceMs > 0` is only provable by the canary above.

### DIFF
--- a/packages/civitai-redis/src/__tests__/cluster-routing-retry.test.ts
+++ b/packages/civitai-redis/src/__tests__/cluster-routing-retry.test.ts
@@ -22,6 +22,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import {
   isTransientClusterRoutingError,
   withClusterRoutingRetry,
+  createNudgeDebouncer,
 } from '../cluster-routing-retry';
 
 // The exact fleet-wide throw measured during a live rolling update (getSlotRandomNode reads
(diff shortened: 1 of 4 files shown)


ZacxDev:
Canary-plan addendum (from the pre-canary audit) — add a **recovery-regression guardrail** before flipping the env on any pool:

Beyond the \`#rediscover\`/\`getSlotRandomNode\` CPU-share drop, watch **\`redisRoutingRetryCounter{result="exhausted"}\` vs \`{result="recovered"}\`** and the pool 5xx rate before/after enabling on api-heavy. Both counters already exist — no new instrumentation.

**Interpretation:** if `exhausted` (→ user 500s) or 5xx **rises** when the debounce is on, `REDIS_CLUSTER_REDISCOVER_NUDGE_DEBOUNCE_MS` is too long relative to the rediscover settle time and is starving retries of a fresh slot map → **shorten it, or roll back (env→0)**. This is the one signal that catches the sole theoretical enabled-path regression that the unit tests (which mock recovery) cannot prove out. Start at the short end of the 500–1000ms window and only lengthen if `exhausted`/5xx stays flat.

ZacxDev:
Closing — parking this as the storm it targets is **burst-conditional, not every-peak**. At the 2026-07-27 traffic peak, api-heavy routing-retry = 0/s (and was ~1.2/s pre-#3389 at the same clock) — the ~1398/s storm the survey caught was during an api-heavy *burst* (32/32 @ ~1.38 cores/pod), which is intermittent. Separately, **#3389 (tag L1 caches) shipped and is booked** avoiding ~14k redis cmds/s at peak (basic-tags 7.1k/s @ 80%), which cuts the very command volume that triggers those bursts — so the storm should be rarer/milder when api-heavy does spike.

Net: nothing to canary right now, and I would rather not touch the shared `packages/civitai-redis` client speculatively. The design here is sound (pre-canary audit: safe-to-canary, default-0 inert, retry semantics preserved) and preserved on branch `zach/redis-rediscover-nudge-debounce` — revive + canary it if a burst-storm actually recurs (the new longtask/CPU-per-req + routing-retry alerts will surface it).

PR #3394, opened 2026-07-27
feat(articles): import markdown into the article editor

AKharytonchyk:
Adds an **Import Markdown** button to the article editor toolbar: pick a `.md` file, and it's converted and inserted.

Authors keep long-form article drafts as markdown files. Pasting them didn't work — tiptap has no markdown parser, and its built-in paste rules only cover inline marks (`**bold**`, `` `code` ``), so headings, fences, lists and tables all arrived as literal characters.

## Why upload rather than convert-on-paste

An earlier version of this PR converted markdown on paste. I removed it. Detecting "is this paste markdown?" is not reliably solvable: gating loosely converted Python, YAML, shell transcripts and ASCII art; gating tightly skipped the commonest document there is (headings plus a list, no fence). Converting mid-paste also brought its own problems — the caret can move before the insert lands, a selection stops being replaced, and pasting inside a code block split it.

Choosing a file is unambiguous, so none of that applies.

## What the converter does

The binding constraint is `DEFAULT_ALLOWED_TAGS` in `html-sanitize-helpers.ts`, which runs as a zod preprocess on save — anything it drops is lost *after* the author has published. So constructs the editor can't store are lowered on the way in, and the author is told what changed:

| Construct | Becomes | Why |
|---|---|---|
| GFM table | nested list, each cell labelled `Header: value` | no table tags allowlisted |
| H4–H6 | H3 | only h1–h3 allowlisted |
| Task list | `[x] ` / `[ ] ` text | `<input>` not allowlisted |
| `~~strike~~` | `<s>` | `<del>` not allowlisted |
| Off-site image | link | see below |
| HTML comments, `<div>` wrappers | dropped | wouldn't survive save |

Tables became **lists rather than code blocks** because a code block is width-constrained in the editor: a padded pipe table wrapped mid-row and looked broken. Lists reflow. Cells move across as inline nodes rather than flattened text, so bold survives and so do link targets — a `| Model | [dl](url) |` download table previously kept "dl" and discarded the href.

**Off-site images are demoted to links.** `extractImagesFromArticle` and `getContentMedia` both go through `extractCloudflareUuid`, which only accepts Civitai-hosted URLs — so an `` on any other host is never extracted, never scanned, never counted by the publish gate, and leaks every reader's IP to that host. Reference style (`![alt][id]`) is covered too, since it reaches the same ``.

## Notes for review

- **`blockquote` added to `DEFAULT_ALLOWED_TAGS`** — a site-wide sanitizer change, split out as #3419. Both add the identical line; I verified merging #3419 then this branch is clean with exactly one entry, so order doesn't matter.
- **The markdown stack is lazy.** `unified` + remark + rehype is ~41 kB brotli and the editor chunk is shared by every rich-text surface, while only articles enable markdown — so the converter is `await import()`-ed. Verified zero static importers.
- `.ladle/` gains a `process.env` shim and a `query-string` interop alias so the included story can boot. Dev-only, never in the app build. These fix no pre-existing story — only this one needs them.

## Testing

34 unit tests on the converter, including one asserting the output loses no tag to `sanitizeHtml`, and a save→sanitize→`generateJSON` round-trip pinning the `contentJson` the REST API returns.

6 component tests in real Chromium on the control, including the `editor.isEmpty ? setContent : insertContent` branch — the one that decides whether an import can wipe a draft.

Verified against 19 real documents: 27 tables converted, zero tags lost to the sanitizer.


AKharytonchyk:
@bkdiehl @JustMaier — a couple of things need a maintainer here, since this is a fork PR:

**CI is blocked.** `Lint` and `Submodule Pin Guard` are both parked at `action_required`, so nothing has actually run yet. Could one of you hit *Approve and run workflows*? Note the `typecheck` job is deliberately skipped on fork PRs (it needs the private submodule, which a fork can't fetch), so it won't gate this either. I ran it locally: 12 errors, identical to `main`'s baseline, none in the changed files. Unit + component suites also show failing sets identical to `main`.

**One decision worth making explicitly.** This adds `blockquote` to `DEFAULT_ALLOWED_TAGS`. StarterKit already exposes the node via the `> ` input rule, but it was never allowlisted, so quotes have been silently unwrapped on save. Fixing it for articles leaves the platform uneven: comments (`commentv2.schema.ts:35`, `comment.schema.ts:38`), resource reviews (`resourceReview.schema.ts:62,74,85`) and model-version descriptions (`model-version.schema.ts:281,402`) pass their own narrower `allowedTags` and still strip it. Extending those is a one-line change per file — I left it out because "blockquotes in comments" reads like a product call rather than a bug fix. Happy to fold it in here or ship it separately.

@bkdiehl you're the most recent owner of `RichTextEditor/`, so flagging two things in there specifically:
- `editorProps` must be `{}` and never `undefined` when the markdown control is off — tiptap reads `options.editorProps.dispatchTransaction` unguarded and an explicit `undefined` overrides its `{}` default, which crashes every editor without the new control. There's a component test pinning the disabled case.
- The paste handler intentionally ignores a `text/html` sibling and decides on `text/plain` alone, because editors put syntax-highlighted HTML next to the source and gating on `text/html` would skip the commonest case.

AKharytonchyk:
Pushed 3bdd371 — worth calling out because it touches image moderation.

**Off-site images bypass scanning entirely.** `extractImagesFromArticle` and `getContentMedia` both route through `extractCloudflareUuid`, which gates on an allowlist of Civitai hosts (`image.civitai.com`, `civitai.com`, `wasabisys.com`, `civitai-prod.s3.amazonaws.com`). An `` on any other host survives the sanitizer untouched but is never extracted, never scanned, and never counted by the publish gate — and it leaks every reader's IP to that host on page load. Verified:

```
external hotlink -> <p><img src="https://evil.example/tracker.jpg" alt="x" /></p>   // survives
javascript: src  -> <p><img /></p>                                                 // stripped, no XSS
```

This predates the branch — an HTML paste could already carry one — but markdown `` used to do nothing and would now produce one in a single paste. So the import now demotes off-site images to links: the reference survives, no unscanned image is embedded, and the count is reported so the author re-adds them with the image button (which uploads through Cloudflare). Civitai-hosted images pass through untouched and stay scannable. Three tests cover it.

The underlying gap is still open for HTML paste and hand-authored content — I've filed it separately rather than widening this PR.

AKharytonchyk:
Correcting an over-generalisation in the description, and summarising what review changed.

**"input rules only fire while typing, never on paste" is too broad.** Tiptap 3 ships `addPasteRules` for bold, italic, strike and inline code, so `**bold**`, `*italic*`, `~~strike~~` and `` `code` `` already convert on paste today. What arrives literally is the *structural* markdown: headings, fences, blockquotes, lists and tables. That's what this converts, and `looksLikeMarkdown` deliberately leaves inline-only pastes to the existing rules.

**Six defects found in self-review and fixed since:**

1. **Paste destroyed code.** A single `#` or `>` line was treated as proof of markdown, but `#` starts a comment in Python/YAML/shell/TOML. Pasting a Python snippet invented an `<h1>` from its first comment, reflowed the imports into one paragraph and ate `__init__` as bold. Signals are now weighted — a fence or table delimiter converts alone, `#` and `>` only count together. All 19 documents this was built against still convert.
2. **`<your-token>` was deleted outright** and `<lora:add_detail:0.8>` lost its angle brackets. Both were silent, and prompts are the most common content in an article here.
3. **`editorProps` re-ran `setOptions` on every keystroke** of all 31 editors, because tiptap compares options by reference and a literal is never equal. The handler is now a tiptap extension, which also removes a ref that was empty during mount.
4. **The 41 kB markdown stack sat in the shared editor chunk** that every rich-text surface loads, though only articles enable markdown. Now `await import()`-ed from both entry points, with the detector split into a dependency-free module so the paste path can still decide synchronously.
5. **Frontmatter containing a blank line corrupted the article** — the closing `---` became a setext underline and promoted the last key to a heading. Now scanned line-wise. The regex also held an invisible U+FEFF byte and used `[^]`.
6. **Ragged tables grew a phantom column**, hidden by the very assertion meant to cover them.

Tests: 34 unit + 5 component. Typecheck unchanged from baseline.

**Still open, and worth a maintainer's opinion before I act:**

- **This should probably be split.** The one-line `blockquote` allowlist addition is the highest-value change here and it's buried among 15 files. Reviewers suggested four: sanitizer one-liner, `.ladle` chore, extension, feature. I'd rather split than have you review it as-is — say the word.
- **`tools/civitai-markdown-importer/` should probably not land.** No `tools/` precedent in the repo, unsigned MV3 with host permissions on both domains, no owner or release path, and the in-app Import button largely supersedes it. Shipping a Civitai-branded extension reads like a brand/T&S decision rather than a PR.
- **Blockquote is styled when published but invisible while editing** (the editor isn't inside `TypographyStylesWrapper`), so it has the same WYSIWYG gap #3403 closes for headings. Not covered by that PR.

AKharytonchyk:
**Moving this to draft — don't spend time on it yet.** A self-review round found 15 defects, several of them data loss, and two of them were introduced by my *previous* round of fixes. Details below so the work isn't lost, and one product question at the end.

The one clearly-good line in here has been split out as #3419 (`blockquote` in `DEFAULT_ALLOWED_TAGS`) — that one is ready and independent.

## Defects I introduced while "fixing" earlier review comments

**The ragged-table clamp deletes cells.** I claimed GFM ignores body cells beyond the header width. It doesn't — for `| a | b |` + a 4-cell body row, mdast row lengths are `[2, 4]`; remark keeps them. And the output is a `<pre>` block that is never re-parsed as a table, so the "phantom column" I was avoiding didn't exist. Verified: cells `3` and `4` are now silently dropped, and the test I added asserts the deletion.

**The paste detector now rejects the commonest markdown there is.** Tightening it against code pastes overshot:
```
# My LoRA

## Usage

- trigger: foo
```
→ `looksLikeMarkdown` returns **false**. Headings-plus-a-list is the canonical README paste, and the converter handles it perfectly; only the gate fails.

## Defects still open

Frontmatter (all verified):
- A YAML comment inside the fence (`# TODO: update`) trips the `hasHeading` bail, reproducing the exact setext-underline bug this was written to fix: `date: 2024-01-01` becomes an `<h2>`.
- Body content between a leading `---` rule and the next `---` is deleted whenever the first line matches `Word:` — including `https://example.com`, since `https:` matches.
- An indented closing `---` is accepted, so an indented block scalar leaks its contents into the document.

Raw HTML handling — my `html`→text change went too broad:
- HTML comments now publish as visible body text (`<!-- private note -->`).
- `<br>` renders as literal `<br>`, though `br` **is** allowlisted and in the Tiptap schema.
- A README's `<div align="center">` wrapper dumps as visible markup, and the `` inside it bypasses the off-site-image demotion entirely, so `externalImagesLinked` stays 0 and the author is told nothing.
- Reference-style images (`![cat][c]`) never hit the demotion branch at all — the security rationale for that feature is bypassed.

Paste mechanics:
- No `isActive('codeBlock')` guard, so pasting inside a code block shreds it into three blocks.
- The `await import()` gap misplaces the insertion three ways: typing during the gap lands text *before* the earlier paste; moving the caret inserts mid-word elsewhere; pasting over a selection no longer replaces it.
- The "raw text fallback" isn't raw — `insertContent(string)` HTML-parses it, so `<b>` becomes real bold and `<script>` is silently deleted.
- Gating on `text/plain` alone discards a rich `text/html` sibling, so copying a formatted section out of the editor, Slack or Docs loses bold and links whenever the plain flavour happens to contain a fence.
- The table-delimiter regex matches `| | |` with no dashes, so ASCII art converts.
- Shell transcripts (`# apt update` + `> y`) still convert.

## The product question

I've now broken paste detection twice in opposite directions, and I don't think it's fixable by patching: there is no reliable way to distinguish a markdown document from a code snippet or a shell transcript, and the async-gap problems are inherent to converting during a paste.

**I'd drop paste conversion and keep only the Import Markdown button.** That removes 7 of the 15 defects outright — the button has no detection problem (you picked a `.md` file) and no selection problem (it replaces an empty editor or inserts deliberately). It also still solves the original request, which was getting `.md` documents into the editor.

Happy to be argued out of it if paste-to-convert is considered important enough to keep iterating on. Until that's decided I'll leave this in draft and fix the converter-side defects, which apply either way.

Also still unresolved from earlier review, and both your call rather than mine:
- `tools/civitai-markdown-importer/` — no `tools/` precedent in the repo, unsigned MV3 with host permissions on both domains, no owner or release path, and it's inert until this ships. Recommend dropping.
- The `.ladle/` shims in here are real but I could not find a pre-existing story they fix — I tested `MyChallengeCard` on `main` and it fails identically with and without them (`missing IntersectionObserverCtx`, an unrelated provider bug). So they belong with this PR's story, not as a standalone change.

AKharytonchyk:
## Verified locally

| Check | Result |
|---|---|

**Unit:** 8091 passed / 0 failed (baseline 8057) — 34 converter tests.
**Component:** 854 passed, 10 failed files — the *same 10* that fail on unmodified `main`, all from the missing submodule. My 6 new `InsertMarkdownControl` tests pass.

One thing worth recording, because it looks alarming: the first full component run on this branch showed **12** failing files instead of 10. Both extras were `Vitest failed to find the runner`, which is the cold-`optimizeDeps` mid-run reload documented at `vitest.config.mts:98-124` — `@vitest/browser` seeds `optimizeDeps.entries` from *every* `*.browser.test.tsx`, so adding a file perturbs it. Both files pass in isolation on this branch (14 tests), and a re-run returns exactly the baseline 10. Not a regression, but if you see 12 on a cold cache, that's why.

Also retired: #3419 and this PR both add `'blockquote',` to `DEFAULT_ALLOWED_TAGS`. I merged #3419 and then this branch on a scratch branch — clean, with exactly one entry. They can land in either order, so this is no longer gated on merge sequence.

AKharytonchyk:
Out of draft. The description above has been rewritten — it previously documented paste conversion, which is gone, so don't review against the original text.

What changed while this was in draft:

- **Paste conversion removed**, along with the browser extension that depended on it. Upload button only. That closed 7 defects structurally rather than by patching.
- **8 converter defects fixed**, several of which lost content silently: frontmatter with a blank line promoted a key to a heading; a `---` rule followed by a `Version:` line deleted the body after it; an indented closing `---` leaked private frontmatter; `<your-token>` was deleted outright; `<lora:x:0.8>` lost its angle brackets; HTML comments and `<div>` wrappers were published as visible markup; reference-style images bypassed the off-site rule.
- **Two of my own earlier "fixes" reverted** — a table column clamp built on a false premise about GFM (it deleted cells), and a paste heuristic that rejected the commonest markdown there is.
- **Tables now become nested lists**, not code blocks, after the code block turned out to wrap mid-row and look broken in the editor.
- **Tests added where the risk actually is**: the import control had none, including the branch that decides whether an import replaces the document or splices into it.

Verified locally, since CI can't run on fork PRs here (parked at `action_required`, and `lint.yml` skips typecheck on forks by design):

| Check | Result |
|---|---|
| CI's blocking gate, replicated (ESLint errors + Prettier on added `.ts`/`.tsx`) | pass — 0 errors, 0 unformatted |
| `tsc --noEmit` | 16, byte-identical to `main`'s 16 |
| Unit | 8091 passed / 0 failed (baseline 8057) |
| Component | 854 passed; 10 failed files — the same 10 that fail on unmodified `main` |
| Test-merge into `main` | clean |

The 16 typecheck errors and 46 unit collection failures are the uncloneable private `event-engine-common` submodule; they're identical on unmodified `main`, and **0 tests fail** there, so any failing test would be mine.

One caveat worth knowing: on a cold `optimizeDeps` cache the component suite may report 12 failing files instead of 10. Both extras are `Vitest failed to find the runner`, the flake documented at `vitest.config.mts:98-124` — adding any `*.browser.test.tsx` perturbs `optimizeDeps.entries`. Both pass in isolation and a re-run returns the baseline 10.

JustMaier:
Decisions from @JustMaier on the three open questions, plus one scope change. Thanks for your patience — and for splitting out #3419, which merged along with #3401, #3403 and #3405.

## 1. Tables — we want to keep them, not lower them

This is a change of direction rather than an approval, so it's the one worth reading carefully.

Your converter turns GFM tables into nested bullet lists because `<table>` isn't allowlisted. That reasoning was right about the constraint. But the constraint is the thing we'd rather change: we do want real tables, just not everywhere.

Specifically: tables are welcome in **articles, model descriptions, 3D model descriptions, post descriptions** — long-form authored content. We do **not** want them in **comments or chat**.

That split already falls out of the existing sanitizer structure, which is convenient:

- `DEFAULT_ALLOWED_TAGS` in `html-sanitize-helpers.ts` backs articles, model descriptions, bounties and post descriptions
- comments (`commentv2.schema.ts:35`, `comment.schema.ts:38`), resource reviews (`resourceReview.schema.ts:62,74,85`) and model-version descriptions (`model-version.schema.ts:284,423`) each pass their own narrower `allowedTags` and would be unaffected

So adding `table`/`thead`/`tbody`/`tr`/`th`/`td` to `DEFAULT_ALLOWED_TAGS` alone produces exactly the split we want, with no per-schema gating.

It's more than a one-line change though, and we'd rather scope it honestly than have you discover it midway:

- `DEFAULT_ALLOWED_ATTRIBUTES` needs `colspan`/`rowspan` (and probably nothing else) for the table tags
- the editor needs Tiptap's Table extension registered on the surfaces that allow it, or authors can see tables but not create or edit them
- rendering needs styles — `TypographyStylesWrapper` and `.markdown-content` both currently style a `blockquote` nobody could save, which is the same trap in reverse
- the round-trip test you already wrote (output loses no tag to `sanitizeHtml`) is the right guard to extend

**How you split this is your call, and either is fine by us:**

- **(a)** land #3394 as-is with the list-lowering, then add table support as a follow-up. Gets the import button shipped sooner; the lowering becomes dead code when tables land.
- **(b)** do table support first as its own PR, then simplify #3394 to pass tables straight through — less total work and no throwaway code, but the import button waits.

We'd lean (b) if you have the appetite, because the list-lowering is a decent amount of careful code that (a) throws away. But (a) is a legitimate choice and we won't push back on it. If you'd rather not take on table support at all, say so and we'll pick it up — you've done plenty here already.

## 2. Off-site images — ship the demotion as-is

Approved as written. Demote to a link, tell the author, keep the reference.

The underlying gap is now tracked separately as #3434, since it predates your branch and is reachable through HTML paste, the API and the HuggingFace importer regardless of this PR.

**Follow-up we'd like, if you're interested:** the ingest-on-save option from that issue — fetch a non-Civitai image, upload it through Cloudflare, rewrite the `src`. That's the version with no UX cost, and it'd let importing a README keep its images inline.

The thing that makes it more than a copy-paste job, and the reason we haven't just done it: fetching a user-supplied URL server-side is an SSRF surface. It needs its own egress allowlist, protection against redirects to internal addresses, a size and content-type cap, and a timeout. Worth outlining your approach in the issue before writing it so we can agree the shape first. No obligation — happy to take it ourselves if you'd rather stay on the editor side.

## 3. Articles only — confirmed

Keep it on `ArticleUpsertForm`. Easy to widen later, harder to walk back.

## 4. Three things we'd want fixed before merge

You asked implicitly by flagging so much yourself, so to be direct — these are all in code this PR adds, not pre-existing:

1. **`<input type="file">` is rendered inside `RichTextEditor.Control`**, which is a `<button>`. Interactive content inside a button is invalid HTML, and the button's `onClick` calls `inputRef.current.click()`, whose click event bubbles back to the button. The spec's activation guard probably saves it in Chromium — your tests pass — but it's fragile across browsers. Rendering the input as a sibling avoids the question entirely.

2. **Multi-line raw HTML blocks are silently deleted.** A `<div align="center">……</div>` wrapper — very common in READMEs — parses as a single `html` node, matches `HTML_MARKUP`, and is dropped whole with no stat and no notice. This is safe, and better than the alternative you replaced, but the author is told nothing. It deserves a counter and a line in `describeMarkdownConversion` like the other lowerings get.

3. **The `editorRef` move above `useEditor`** in `RichTextEditorComponent.tsx` looks like leftover churn from the paste-conversion approach you removed — no functional effect. Reverting shrinks the diff.

## On the process side

Two things from your PRs turned into repo changes, so you should know they landed:

- **#3401's real cause.** You were right to correct yourself that the script produced only 5 of the 67 strays. The other 62 were hand-authored because `README.md` told people to create folders in `prisma/migrations`. Fixed in #3438, along with a CI guard that fails a PR adding anything under the old path.
- **CI can gate your work now.** `event-engine-common` is public and fetched over HTTPS, so the typecheck job no longer skips fork PRs, and there's a unit test job (report-only initially while we establish its pass rate). Your local-verification comments were the direct prompt for that — they were doing work CI should have been doing. There's also a `CONTRIBUTING.md` now, which writes down the baseline you'd been re-deriving every PR.

You should no longer need to reconstruct a 16-error typecheck baseline by hand.

AKharytonchyk:
Thanks — the three review points are fixed and pushed (`<button>`-nesting, silent raw-HTML drop now counted and reported, `editorRef` hoist reverted). The branch also dropped to 9 files, since the `blockquote` line fell out once #3419 landed.

Before anyone starts the table work, one correction: **"adding the table tags to `DEFAULT_ALLOWED_TAGS` alone produces exactly the split we want" doesn't hold.** I checked the call sites rather than assume, and it's off in both directions.

`getSanitizedStringSchema()` with no options — i.e. the surfaces backed by `DEFAULT_ALLOWED_TAGS` — is exactly three:

| Surface | Wanted? |
|---|---|
| `article.schema.ts:98` | yes |
| `model.schema.ts:134,191` | yes |
| `bounty-entry.schema.ts:30,45` | **not in the list** — would get tables too |

And the two other surfaces you named aren't sanitized at save at all:

- **post descriptions** — `post.schema.ts` has `detail: z.string().nullish()`
- **3D-model descriptions** — `model3d.schema.ts` has `description: z.string().nullish()`

They're sanitized at **render**: `PostDetail.tsx:482` passes `post.detail` through `RenderHtml`, which calls `sanitizeHtml` with the default tag list. So they *would* gain tables, but via a different path and only at render time — meaning content saved before the change starts rendering tables retroactively, and the round-trip test you want extended wouldn't cover that path at all.

So the real effect of the one-line change is: **articles + model descriptions + bounty entries at save, and post + 3D-model descriptions at render** — one surface more than asked for, reached by two mechanisms.

Two smaller notes:

- The #3398 comment says the split works because "comments **and chat** already pass their own narrower `allowedTags`". Only four files do that (`comment.schema.ts`, `commentv2.schema.ts`, `model-version.schema.ts`, `resourceReview.schema.ts`) — chat isn't one. Chat stores `content: z.string()` with `contentType: 'Markdown'` and renders through `CustomMarkdown allowedElements={['a','p','strong']}`, so it never touches the HTML sanitizer. Conclusion holds, mechanism doesn't.
- `model-version` descriptions have their own allowlist, so they'd *not* get tables — worth confirming that's intended, since "model descriptions" and "model-version descriptions" are different surfaces and only the former is covered.

## On (a) vs (b)

Deferring to you on sequencing, with one input: the list-lowering is ~60 lines with 5 tests, so (a) throws away less than it might appear — but it also means shipping a lowering we intend to delete, which will confuse whoever reads it next. If you'd rather own the table work yourselves I'd take that gladly; the sanitizer/attribute/Tiptap-extension/styling span is wider than the import feature and touches surfaces I haven't worked in.

Happy to do (b) if you'd prefer it in my hands — say which and I'll start.

AKharytonchyk:
Thanks for the detail on all four — that was more than I expected, particularly the offer to take tables off my hands.

Taking **(a)**, and not for the reason you'd guess. I went looking at what (b) would actually involve and found something that I think belongs in the table PR whoever writes it.

## 1. Tables — (a), and the scoping is missing a piece

**The styling half is already done.** `TypographyStylesWrapper.module.scss:185-232` has a complete table block — `width: 100%`, `border-collapse: collapse`, th/td padding, borders, `caption-side`. It's the blockquote situation in reverse: styles maintained for a tag nobody can save. Worth knowing before someone budgets time for it.

Two asymmetries fall out of that. `.markdown-content` in `globals.css` has the opposite problem — `hr`, `h2` and `blockquote` rules only, no table rules at all — so the two renderers would disagree the moment tables become saveable. And those `TypographyStylesWrapper` rules look inherited from Mantine's typography defaults rather than authored for Civitai, which is why they have the `width: 100%` half and not the part that matters.

**What's actually missing is overflow containment.** There is no `overflow-x: auto` on a table or any wrapper anywhere in that file — the only one is on `pre` (`:113`). With `width: 100%` and default `table-layout: auto`, that gives two distinct failure modes rather than one:

- **Many short columns compress instead of overflowing.** Fine at the article column's ~848px for three columns. At a phone's ~360px, five columns is ~70px each and cell text wraps mid-word.
- **One long token overflows the container.** A table's used width is `max(100%, min-content)`, so a single unbreakable string pushes min-content past the container and the table spills with nothing up the tree to catch it. A URL, an AIR, or a `<lora:add_detail:0.8>` in a cell does it — which is a fair description of what authors actually put in tables here.

So the table PR needs a scroll wrapper (or `table-layout: fixed` plus `overflow-wrap`), a deliberate answer for narrow viewports, and probably a matching `.markdown-content` block. That's a call for whoever owns the article layout and can test it against real content, which is exactly why I'd rather not fold it into this PR — I'd be guessing at your breakpoints from outside.

None of that is an argument against the decision. Long-form tables are clearly the right call for imported READMEs; I just don't think the cost is the one-line-plus-styles it looks like from here.

**On the throwaway concern:** if tables land everywhere long-form, deleting the lowering is a two-line diff and I'll do it. If the mobile question lands somewhere narrower, it's already written and tested. Either way it isn't wasted the way it looked.

## 2. Off-site images

Ack, shipping as-is.

On the ingest-on-save follow-up: I'll post the shape in #3434 before anyone writes code — egress allowlist, redirect-to-internal handling, size and content-type caps, timeout — so it's cheap to shoot down. Treat that as a design comment rather than a claim on the work; if you'd rather take it, take it.

## 3. Articles only

Confirmed, unchanged.

## 4. The three fixes

All three were already in before your comment landed, so they may not show if you read an earlier state:

| Ask | Where |
|---|---|
| `<input>` out of the `<button>` | `InsertMarkdownControl.tsx:83` — sibling of `RichTextEditor.Control`, with the reason in a comment so it doesn't get "tidied" back |
| Dropped raw HTML needs a counter and a notice | `markdown-to-editor-html.ts` — `htmlBlocksDropped` counted at `:192`, surfaced in `describeMarkdownConversion` at `:289` alongside the other lowerings |
| Revert the `editorRef` churn | Done — `editorRef` no longer appears in that file's diff |

You were right that the third was leftover from the paste approach.

## Status

Ready for review. Currently 9 files / +1053 −6, down from 15 — the paste-conversion path and the browser extension are both gone, so what's left is the converter, one toolbar control, the Ladle story and tests.

Current with `main` — nothing that landed since touches any file here, so the local verification in my comment above still applies to this commit (ESLint + Prettier clean on added files, typecheck at baseline, unit 8091 passed / 0 failed, component at the baseline 10 failing files). Though per your last point I should now be able to let CI say that instead — thanks for #3438 and for wiring typecheck and unit onto fork PRs. `CONTRIBUTING.md` in particular saves the thing I was re-deriving every time.

AKharytonchyk:
## Rebased onto current `main` — and it needed real changes, so please re-read before reviewing

`main` moved 1,544 commits since the last push here. Three things changed, and one of them is a genuine behaviour risk that the rebase hid.

**1. The sticker work landed in the same file.** `RichTextEditorComponent.tsx` now has a `sticker` control, and it collided with the `markdown` control at all three of the same places — the `includeControls.includes(...)` flag, the toolbar `ControlsGroup`, and the `ControlType` union. Resolved as a **union** in every case: both controls exist, neither replaces the other.

**2. The rebase was textually clean and still broke the browser tests.** This is the part worth knowing. `RichTextEditor` now calls `useFeatureFlags()` unconditionally, then `useOwnedSticker()` → `useQueryUserCosmetics()` → `useCurrentUser()`, then a `trpc.cosmetic.getStickerBalances` query. The component-test scaffold provides none of those, so all 6 tests in `InsertMarkdownControl.browser.test.tsx` failed with `useFeatureFlags can only be used inside FeatureFlagsCtx`. **Typecheck was green throughout** — only the `component` project caught it.

Fixed with three scoped stubs, since `sticker` isn't in this test's `CONTROLS` and none of it is what the test is about:
- `~/providers/FeatureFlagsProvider` → `{ stickers: false }`
- `~/components/Sticker/sticker.util` → `useOwnedSticker` returns empty (spread-preserving, so the rest of the module is real)
- `~/utils/trpc` → the balances query only. Uses the `importOriginal`-spread form, because `no-wholesale-module-mock` gates that specific module at `error` now.

The balances query is `enabled: false` here, but the hook still resolves tRPC context at render, so an unmocked `trpc` throws before the flag can matter.

**3. History squashed, 12 commits → 1.** Six of the twelve were review-response fixups. The tree is byte-identical (`git reset --soft`), and the three asks from your review are unchanged and still where the table in my previous comment says they are. If you'd rather see that review-response commit as its own boundary, say so — the old tip `d6a2164358` is recoverable and I'll restore the granular history.

## Re-verified after the rebase

- Full-repo typecheck: **0 errors** (`main` is also 0 now, not the 14 I quoted before).
- `markdown-to-editor-html.test.ts`: **35/35**.
- `InsertMarkdownControl.browser.test.tsx`: **6/6** (533 ms).
- eslint + prettier clean on all added files.

Size is now +1072/−6 in 9 files — up from the +1053 I quoted, entirely the three test stubs above.

Nothing about the converter, the off-site-image demotion, or the articles-only scoping changed. Option **(a)** still stands as agreed.

⚠️ One note that isn't about this PR: I needed **Node 24** (`.nvmrc` says 24.19.0) to run the suites at all. On Node 22 vitest dies at project init with `Cannot read properties of undefined (reading 'async')` out of `apps/event-engine/vitest.config.ts`, which reads like a broken test rather than a wrong toolchain. Cost me a detour; might be worth a line in CONTRIBUTING.

PR #3401, opened 2026-07-27
fix(db): create empty migrations next to the schema

### LINKED ISSUE #3397: db:migrate:empty writes to the pre-monorepo prisma/migrations directory
## Summary

`pnpm run db:migrate:empty` writes new migrations into the repo-root `prisma/migrations/`, but the canonical directory since the monorepo move is `packages/civitai-db-schema/prisma/migrations/` — the one next to the schema Prisma actually reads, and the only one with `migration_lock.toml`.

## Cause

`scripts/create-empty-migration.mjs:36` builds a **cwd-relative** path:

```js
const migrationPath = join('prisma', 'migrations', folderName);
```

Run from the repo root (the normal case), that resolves to `./prisma/migrations`, which is the pre-monorepo location.

## Effect

The two directories have drifted. Root `prisma/migrations/` holds ~67 migrations dated 2026-05-22 onward; the package dir holds the full 2022→present history plus `migration_lock.toml`. Folder names don't overlap, so nothing is duplicated and nothing is broken today — migrations are applied by hand anyway — but roughly a third of recent migrations sit in a directory Prisma no longer reads, and anyone looking for a migration has to grep both.

## Suggested fix

Point the script at the schema directory:

```js
const migrationPath = join('packages', 'civitai-db-schema', 'prisma', 'migrations', folderName);
```

Better still, derive it from the `prisma.schema` path already declared in `package.json` so it can't drift again.

Separately worth deciding whether to consolidate the ~67 stranded migrations into the package directory, or leave them as history and just fix the script going forward.

AKharytonchyk:
Fixes #3397.

`scripts/create-empty-migration.mjs` built the target path as a **cwd-relative** `prisma/migrations`, which is the pre-monorepo location. The schema and the canonical migrations directory (the only one with `migration_lock.toml`) now live in `packages/civitai-db-schema/prisma/`, so `pnpm run db:migrate:empty` was dropping new migrations into a directory Prisma no longer reads — about a third of everything added since 2026-05-22.

Now derived from `package.json`'s `prisma.schema` and anchored on the script's own location, so it follows the schema if that moves again and no longer depends on where it was invoked from.

Verified from two working directories:

```
$ node scripts/create-empty-migration.mjs "probe from root"
  Location: packages/civitai-db-schema/prisma/migrations/2026…_probe_from_root/migration.sql

$ cd src && node ../scripts/create-empty-migration.mjs "probe from subdir"
  Location: packages/civitai-db-schema/prisma/migrations/2026…_probe_from_subdir/migration.sql
```

(Both probe directories removed before committing.)

It also fails loudly now if `prisma.schema` is ever missing from `package.json`, rather than silently writing somewhere unexpected.

The ~67 migrations already sitting in the root `prisma/migrations/` are untouched — whether to consolidate them or keep them as history felt like a separate call.

### DIFF
--- a/scripts/create-empty-migration.mjs
+++ b/scripts/create-empty-migration.mjs
@@ -1,5 +1,19 @@
-import { mkdir, writeFile } from 'fs/promises';
-import { join } from 'path';
+import { mkdir, readFile, writeFile } from 'fs/promises';
+import { dirname, join, relative, resolve } from 'path';
+import { fileURLToPath } from 'url';
+
+const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+
+// Migrations live next to the schema, which moved into packages/ with the
+// monorepo. Deriving the directory from package.json's `prisma.schema` keeps it
+// correct if the schema moves again, and anchoring on the script's own location
+// means the result doesn't depend on the cwd it was invoked from.
+const migrationsDir = async () => {
+  const pkg = JSON.parse(await readFile(join(repoRoot, 'package.json'), 'utf8'));
+  const schema = pkg.prisma?.schema;
+  if (!schema) throw new Error('package.json is missing a `prisma.schema` path');
+  return join(repoRoot, dirname(schema), 'migrations');
+};
 
 // Get migration name from command line args
 const migrationName = process.argv[2];
@@ -35,9 +49,10 @@ const getTimestamp = () => {
 const timestamp = getTimestamp();
 const snakeCaseName = toSnakeCase(migrationName);
 const folderName = `${timestamp}_${snakeCaseName}`;
-const migrationPath = join('prisma', 'migrations', folderName);
 
 try {
+  const migrationPath = join(await migrationsDir(), folderName);
+
   // Create migration folder
   await mkdir(migrationPath, { recursive: true });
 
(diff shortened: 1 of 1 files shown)


AKharytonchyk:
Correcting two things in the description above, both of which overstated the case.

**The proportion is wrong.** I wrote "about a third of everything added since 2026-05-22". By folder timestamp it's root 67 vs package 40 — **63%**, not a third.

**And the causation is unsupported.** Only 5 of those 67 have non-`:00` seconds in their timestamp (`20260706120050`, `20260706135826`, `20260709201808`, `20260713162642`, `20260721160807`). `getTimestamp()` uses real wall-clock `getSeconds()`, so it essentially cannot emit `:00` — meaning the other 62 were hand-authored directly into the stale path, not produced by this script. The bug is real and the fix stands on its own, but it did not cause most of the drift I cited it for.

Also worth folding in, from review: `README.md:152` still points contributors at `prisma/migrations`, and the 67 stragglers stay orphaned after this. Happy to add the README line here; moving the stragglers feels like a separate, deliberate call given migrations are applied by hand.

AKharytonchyk:
## Verified locally

| Check | Result |
|---|---|

**Unit:** 8057 passed / 0 failed, identical to baseline. Also exercised the script itself from both the repo root and a subdirectory; both wrote to `packages/civitai-db-schema/prisma/migrations`, and I removed the probe directories before committing.

Issue #3398, opened 2026-07-27
Article content pipeline papercuts: no GFM in markdownToHtml, raw HTML to moderation/search, unstyled editor headings


AKharytonchyk:
Four small, independent findings in the article/markdown content pipeline. Grouped rather than filed separately because each is a few lines; happy to split if any is picked up.

### 1. `markdownToHtml` has no `remark-gfm`, so imported READMEs lose their tables

`src/utils/markdown-helpers.ts:8-20` builds `remarkParse → remarkRehype → rehypeStringify` with no GFM plugin, so a pipe table arrives as a paragraph of literal `|` characters. This runs on HuggingFace model READMEs (`server/importers/huggingFaceModel.ts:104`) and daily-challenge content (`games/daily-challenge/generative-content.ts:206`) — HF READMEs are full of tables.

`remark-gfm` is already a dependency. Note that enabling it alone would emit `<table>`, which `DEFAULT_ALLOWED_TAGS` strips — so it needs pairing with a down-convert. `src/utils/markdown-to-editor-html.ts` (added in #3394) already has one that renders a table as aligned pipe text in a code block, which could be reused.

### 2. Text moderation receives raw HTML rather than text

`src/server/services/article-moderation.adapter.ts:17-26` selects `content` and submits it unchanged, so the scanner sees `<p>`, `<div>`, `<iframe>`, `<edge-media>` markup mixed into the prose. `removeTags` (`utils/string-helpers.ts:212`) already produces clean text and is used for meta descriptions and OG images. Passing markup wastes scan budget and adds tokens the model has to ignore.

### 3. Meilisearch indexes article content as raw HTML

`articles.search-index.ts` pulls `content` via `articleDetailSelect` and lists `content` in `updateSearchableAttributes` (`:31`). Tag names are therefore searchable tokens — a query for `iframe` or `blockquote` can match on markup rather than prose. Stripping tags in `transformData` before indexing would fix it, at the cost of a full reindex.

### 4. Headings render at body size while editing

The editor content isn't wrapped in `TypographyStylesWrapper` — only the published view is, via `RenderRichText`. `RichTextEditorComponent.module.scss:114-124` styles `h1/h2/h3` only with an anchor-offset pseudo-element, no font size. So an H1 and a paragraph look identical while writing and diverge once published, which makes the heading buttons hard to use and is confusing when importing a long document.

AKharytonchyk:
Working through these. Status:

[...]

PR #3403, opened 2026-07-27
fix(editor): size headings in the rich text editor

AKharytonchyk:
Addresses item 4 of #3398.

Tailwind's preflight resets headings to `font-size: inherit`, and the editor's content — unlike the published article, which `RenderRichText` wraps in `TypographyStylesWrapper` — has no typography styles of its own. `RichTextEditorComponent.module.scss` only gave `h1/h2/h3` an anchor-offset pseudo-element under `.htmlRenderer`, nothing under `.richTextEditor`.

Result: an H1 and a paragraph looked identical while writing, and only diverged once published. That makes the H1/H2/H3 buttons hard to use and is disorienting when pasting or importing a long document.

This mirrors the same Mantine variables `TypographyStylesWrapper` uses, so the two views agree by construction rather than by copied numbers, and zeroes `margin-top` on the first block so content isn't pushed away from the top of the editor.

## Verified

Measured in a real browser after pasting `<h1>/<h2>/<h3>/<p>`:

| | font-size | weight | margin-top |
|---|---|---|---|
| h1 | 34px | 700 | 0 (first child) |
| h2 | 26px | 700 | 32px (`spacing-xl`) |
| h3 | 22px | 700 | 25.6px (`0.8 × xl`) |
| p | 16px | 400 | 0 |

Those match `TypographyStylesWrapper`'s values for the same elements.

## A cascade note for reviewers

While checking this I hit a trap worth recording: `@mantine/tiptap/styles.css` is **unlayered**, and unlayered CSS outranks every layered rule — including CSS modules, which Next wraps in `@layer modules` (see the comment at the top of `src/styles/globals.css`). Importing the plain build instead of `styles.layer.css` therefore lets Mantine silently override module styles it doesn't override in the real app. `_app.tsx:85` correctly imports `@mantine/tiptap/styles.layer.css`; my first preview attempt didn't, and these margins appeared not to apply as a result. Nothing to fix in the app — flagging it because it's an easy mistake in any new harness or embed.

### DIFF
--- a/src/components/RichTextEditor/RichTextEditorComponent.module.scss
+++ b/src/components/RichTextEditor/RichTextEditorComponent.module.scss
@@ -42,6 +42,45 @@
       // color should be set inline via style prop if it depends on error
       font-size: var(--editor-font-size, 14px);
     }
+
+    // Tailwind preflight resets headings to inherit, and the editor content —
+    // unlike the published article — isn't wrapped in TypographyStylesWrapper.
+    // Without this an H1 and a paragraph look identical while writing and only
+    // diverge once published. Mirrors the same Mantine variables that wrapper
+    // uses so the two views agree. Only h1-h3 exist here (see CustomHeading).
+    h1,
+    h2,
+    h3 {
+      margin-bottom: var(--mantine-spacing-xs);
+      text-wrap: var(--mantine-heading-text-wrap);
+      font-family: var(--mantine-font-family-headings);
+    }
+
+    h1 {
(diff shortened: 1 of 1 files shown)


AKharytonchyk:
## Verified locally

| Check | Result |
|---|---|

**Unit:** 8057 passed / 0 failed, identical to baseline (SCSS only). Verified in a real browser instead: h1 34px / h2 26px / h3 22px vs 16px body, and blockquote at 20px left padding, 16px top margin, 18px type, grey panel — matching `TypographyStylesWrapper` for the same elements.

Issue #3395, opened 2026-07-27
Toolbar image insert silently loses the image on 9 forms (blob: URL never uploaded, src stripped on save)


AKharytonchyk:
## Summary

`InsertImageControlLegacy` inserts a `blob:` object URL and never uploads it. The sanitizer's default `allowedSchemes` excludes `blob:`, so the `src` is stripped on save — the image shows while editing, then becomes an empty `<img>` once stored.

## Affected surfaces

The legacy control is selected whenever `includeControls` has `'media'` but **not** `'video'` (`RichTextEditorComponent.tsx:374-381`). That's 11 call sites across 9 forms:

- `src/components/Bounty/BountyCreateForm.tsx:302`
- `src/components/Bounty/BountyEditForm.tsx:81`
- `src/components/Bounty/BountyUpsertForm.tsx:418`
- `src/components/PurchasableRewards/PurchasableRewardUpsertForm.tsx:153,161,169`
- `src/components/Bug/Bugs.tsx:479`
- `src/components/CosmeticShop/CosmeticShopSectionUpsertForm.tsx:102`
- `src/components/CosmeticShop/CosmeticShopItemUpsertForm.tsx:508`
- `src/components/Questions/AnswerForm.tsx:64`
- `src/components/Questions/QuestionForm.tsx:137`

Article and model editors are unaffected — they also enable `'video'`, which selects the uploading `EdgeMediaEditNode` path.

## The two halves

`InsertImageControlLegacy` (`InsertImageControl.tsx:69-121`) does:

```js
editor?.commands.insertContent({
  type: 'image',
  attrs: { src: URL.createObjectURL(file), filename: file.name },
});
```

No upload, unlike the non-legacy control which goes through `editor.commands.addMedia(file)` → `useCFImageUpload().uploadToCF`.

Then `sanitizeHtml` drops the scheme (verified):

```
input:  <p><img src="blob:http://localhost:3000/9f2c-abc" alt="x"></p>
output: <p><img alt="x" /></p>
```

## Reproduce

1. Open any form above (e.g. create a bounty)
2. Use the image button in the editor toolbar, pick a file
3. The image renders in the editor
4. Save, then reload — the image is gone

## Suggested fix

Route `InsertImageControlLegacy` through the same `addMedia`/`uploadToCF` path as `InsertImageControl`, or drop the legacy control and let those surfaces use the uploading one. Worth checking stored content for src-less `<img>` tags left behind.

AKharytonchyk:
[...]

PR #3404, opened 2026-07-27
test(editor): pin image upload for the image-only editor config

AKharytonchyk:
Follow-up to #3395, which I filed and then closed as invalid. This is the test that disproved it.

## Why it looked broken

The image-only editor configuration (`'media'` without `'video'`) splits responsibility across two files in a way that reads like a bug:

- `InsertImageControlLegacy` (`InsertImageControl.tsx:69-121`) inserts `{ type: 'image', attrs: { src: URL.createObjectURL(file) } }` and performs **no upload**, unlike `InsertImageControl` which calls `editor.commands.addMedia(file)`
- `blob:` is not in sanitize-html's default `allowedSchemes`, so a blob URL that reached save would lose its `src`:

```
input:  <p><img src="blob:http://localhost:3000/9f2c-abc" alt="x"></p>
output: <p><img alt="x" /></p>
```

I stopped reading at the control and filed a bug. It's wrong: `CustomImage`'s node view (`CustomImage.tsx:118-173`) is what uploads — it detects `src.startsWith('blob')`, runs `fetchBlobAsFile` → `uploadToCF`, swaps the attribute to the CDN URL and revokes the object URL. `CustomImage` is exactly what's registered for that configuration (`RichTextEditorComponent.tsx:192-194`), so the 9 affected forms are fine.

## Why pin it anyway

Nothing covered that wiring, and the two halves are far enough apart to invite a "simplification" that removes the node view or the upload effect — with a failure mode that's invisible in the editor and only shows up after save. The `blob:` sanitizer behaviour above is the sharp edge.

The test mounts the real editor with `includeControls={['formatting', 'media']}`, pastes an image file, and asserts `uploadToCF` receives the actual `File` and that the resulting `onChange` HTML carries the CDN id with no `blob:` anywhere.

Passes against unmodified `main` — it documents current behaviour rather than changing it. `useCurrentUser` is mocked the same way the AppBlocks suites do it, since the node view reaches `useEdgeUrl` → `useCurrentUser` and there's no session provider in the component project.

### DIFF
--- a/src/components/RichTextEditor/RichTextEditorImageUpload.browser.test.tsx
+++ b/src/components/RichTextEditor/RichTextEditorImageUpload.browser.test.tsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { renderWithProviders } from '../../../test/component-setup';
+
+/**
+ * Does the image-only editor configuration (`media` without `video`) actually
+ * upload a pasted image, or does the `blob:` URL reach the sanitizer — which
+ * strips the `blob:` scheme and leaves a src-less `<img>`?
+ *
+ * `InsertImageControlLegacy` doesn't upload anything itself, but it inserts an
+ * `image` node, and `CustomImage`'s node view is what performs the upload. This
+ * pins that wiring.
+ */
+
+const uploadToCF = vi.fn(async (file?: File) => ({
+  id: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee',
+  objectUrl: 'blob:stub',
+  name: file?.name ?? '(no file)',
+}));
+
+vi.mock('~/hooks/useCFImageUpload', () => ({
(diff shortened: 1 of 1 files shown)


AKharytonchyk:
Closing this — it doesn't earn a reviewer's attention, and I'd rather the queue be five PRs that do.

The test itself is fine and it passes, but the case for it is weak in a way the other PRs' tests aren't. Tests that ship *alongside a change* justify themselves: they pin the thing being altered. This one pins behaviour that already works, that nobody proposed changing, and it exists mainly because I filed #3395 claiming that wiring was broken, was wrong, and wanted to leave something behind.

There's also no automated protection to gain. This repo has no unit or component test job in CI at all — only `lint.yml`, whose typecheck step is skipped on fork PRs by design. So a standalone test file is only ever executed by whoever runs vitest locally.

If the `blob:` upload wiring does get refactored later, the right move is the one suggested in review: fold the assertion into an existing RichTextEditor browser test rather than add a file. The branch stays on my fork if anyone wants it.

PR #3405, opened 2026-07-27
fix(articles): send text, not raw HTML, to moderation and search

AKharytonchyk:
Addresses items 2 and 3 of #3398.

Article content is stored as sanitized HTML and was handed to both text pipelines verbatim.

## Text moderation

`article-moderation.adapter.ts` selected `content` and submitted it unchanged, so the scanner read `<p>`, `<iframe>`, `<edge-media>` and attribute values as part of the prose.

**Why `html-to-text` and not `removeTags`** — this is the part worth a look. `removeTags` deletes each tag wholesale, so `<a href="http://spam.example">click</a>` becomes just `click` and the URL disappears. In open UGC that URL is the strongest spam signal there is, and stripping it would have made moderation *weaker* than passing raw HTML. `html-to-text` renders it as `click [http://spam.example]`, keeping link and image targets while dropping the markup:

```
in:  <h2>Look 1</h2><p>See <a href="https://spam.example/x">click</a></p>
out: Look 1

     See click [https://spam.example/x]
```

Two option choices are deliberate and commented: `wordwrap: false` (the default wraps at 80 columns, injecting newlines mid-sentence) and headings as `block` (the default `heading` formatter uppercases, and the scanner should see the author's text as written).

**Divergence from the challenge adapter, on purpose.** `buildChallengeModerationText` (`challenge-helpers.ts:54`) uses `removeTags`, so it has the same URL-loss. I didn't change it: challenge descriptions are authored by a small trusted set, whereas articles are open UGC where link targets carry most of the abuse signal. Happy to align challenges in a follow-up if you'd rather have one behaviour — I just didn't want to alter another domain's moderation output in this PR.

## Search

`articles.search-index.ts` lists `content` in `updateSearchableAttributes`, so tag and attribute names were indexed as prose — a query for `iframe` or `blockquote` could match on markup rather than content.

That path uses the regex `removeTags` rather than a full HTML-to-text pass, because it runs over every article on a reindex and URLs aren't a useful search token. Also shrinks the index.

**Requires a reindex** to take effect for existing documents; articles reindex individually as they're edited.

## Testing

10 unit tests on the new helper: markup dropped, link targets kept, image alt + target kept, headings not uppercased, no line wrapping, null/empty/whitespace handled, and legacy tiptap-JSON content passing through as scannable text rather than throwing.

Full unit suite failing set is byte-identical to `main` (`diff` empty — the pre-existing failures are the missing `event-engine-common` submodule). Typecheck 12, same as baseline, none in these files.

Item 1 of #3398 (no GFM in `markdownToHtml`) is not here: enabling `remark-gfm` alone emits `<table>`, and model descriptions use `getSanitizedStringSchema()` with no custom allowlist, so tables would go from literal pipe text to silently deleted. It needs a table down-converter, which #3394 introduces — better done once that lands than duplicated.


AKharytonchyk:
Two things this needs that the description didn't cover.

**Rollout: the search half does nothing for existing documents.** `transformData` strips tags going forward, but the incremental sync range-scans `WHERE "createdAt" >= lastUpdatedAt`, so only new and edited articles get re-pushed. Everything already indexed keeps its raw HTML until someone fires `search-index-sync-articles-reset` manually — it's registered with `UNRUNNABLE_JOB_CRON` and a 3h lock, so it won't happen on its own.

That leaves a ranking split in the meantime: `content` is searchable attribute #2 and Meilisearch weights by position, so older documents have their real words diluted by markup tokens while newer ones don't. Newly published articles will systematically outrank older ones on body-text queries until a reset runs, and the `iframe`-matches-markup problem persists for the whole existing corpus. **So this wants the reset job fired after merge** — flagging rather than assuming, since that's an ops decision.

**Moderation half: the premise in my original description was wrong, and I've corrected the code.** The create, update, rescan and reconcile paths already submit `[title, removeTags(content)].join(' ')` — they never sent raw HTML. Only `articleModerationAdapter.resolveContent` did, and its only consumer is the retry cron. My first version swapped in an `html-to-text` conversion, which was worse: `createXGuardModerationRequest` dedups on `hashContent(args.content)`, so a third distinct shape could never match a primary submission, guaranteeing a fresh paid audit per article. It also dropped the title, which rendered embed-only articles to `''` — and `text-moderation-retry.ts` doesn't treat `''` as missing, so that risked trapping those articles in Pending against the invariant documented on `articleHasText`.

It now produces the same string as the other four call sites, so the dedup works and the title is always present.

Whether all five should preserve link targets instead of discarding them with `removeTags` is a genuine question — URLs are the strongest spam signal in open UGC — but it rewrites every article's hash and is a moderation-policy call, so it's not in here.

AKharytonchyk:
## Verified locally

| Check | Result |
|---|---|

**Unit:** 8057 passed / 0 failed, identical to baseline. I also re-checked this PR's premise after the rebase: it aligns the retry adapter with the four primary submit sites, and although `article.service.ts` did change upstream, all four `textForModeration` shapes are still `[title, removeTags(content)]` — so the alignment holds.

PR #3413, opened 2026-07-28
blocks: decouple agentic-review chat from the live pod + stale-report-row sweeper

ZacxDev:
## What

Decouple the App Blocks **agentic mod code-review** in-modal chat from the live review-agent pod, so the analysis workload can be rendered as an **ephemeral Job** instead of a long-lived Deployment. This eliminates a restart-wedge failure class where a broken review pod could linger for days.

Root cause (full write-up in the deployment repo): the review pod's init container fetches the submitted bundle via a **short-lived (15-min) presigned URL** baked as a literal env. Rendered as a **Deployment**, its pod was resurrected on any node reboot/eviction, and the restart re-fetched the *same long-expired* URL → `403` → `Init:Error` → permanent CrashLoop. The pod is credential-less by design (it statically analyses an adversarial author bundle) so nothing re-mints. It was a Deployment **only** so the pod stayed alive for the P3 chat.

The unlock: the chat grounding was **already** built entirely from the **persisted report** (`buildAgentReviewChatSystemMessage` reads `summaryMd`/`scopeVerdicts`/`codeReview`/`securityAudit` from the DB). The only thing the live pod added to chat was running the LLM call.

## Changes

- **`agentReviewChat` is now stateless** — it calls the LLM directly (new `openrouter.getTextCompletion` helper) grounded on the persisted report, instead of proxying to the per-review agent pod's gateway. No pod, no gateway, no derived bearer. The grounding message drops the "bundle is available to re-read" framing (no live pod) and keeps the adversarial-data framing. A missing report or a `torn-down` (decided) review → `PRECONDITION_FAILED`; `running`/`complete`/`cost-capped`/`failed` reports are groundable (a `failed` report still has partial sections). Any LLM error/timeout/absent client → clean `BAD_GATEWAY`, never a 500 or an internals leak.
- **`deleteAgentReviewResources`** now sweeps **Jobs** (the current workload) plus legacy Deployments/Services, so a mod decision promptly deletes a still-running review Job. A finished Job also self-reaps (TTL) in the infra template.
- **`buildAgentReviewApplyScript`** deletes any same-name review-agent Job before apply — a Job's spec is immutable, so re-dispatch can't just apply over an existing one (Deployments were apply-idempotent; Jobs are not).
- **New `sweep-stale-agent-reviews` job** (every 10m) flips `AppReviewAgentReport` rows stuck `running` past 60m → `failed`, so an un-decided/dead review can't wedge future dispatches via the double-provision guard. **No schema change** (existing `status` enum + `startedAt`). Dark-safe no-op while the flag is off.

The pod's per-review token provisioning is left intact (still used for the pod's internal gateway auth during analysis; the civitai-side derivation is retained as a rollback path).

## Verification
- The 3 affected test files + adjacent router / session tests pass (**28 + 10**); the chat test is rewritten for the stateless path (guard, request shape, no client system-injection, failure containment).
- `tsc --noEmit` clean for **all** changed files.
- **NOT** run live: the feature is **DARK** (Flipt flag `app-blocks-agentic-review` absent) → no real provision can be exercised. What's verified is code/typecheck/tests, not a live review run.

Pairs with the deployment-repo PR that renders the review-agent as a Job (native-sidecar gateway, `restartPolicy: Never`, `backoffLimit: 0`, `activeDeadlineSeconds`, `ttlSecondsAfterFinished`). Security posture of the pod is preserved verbatim.

### DIFF
--- a/src/pages/api/webhooks/run-jobs/[[...run]].ts
+++ b/src/pages/api/webhooks/run-jobs/[[...run]].ts
@@ -24,6 +24,7 @@ import { confirmMutes } from '~/server/jobs/confirm-mutes';
 import { confirmPendingBlockAttributions } from '~/server/jobs/confirm-pending-block-attributions';
 import { bulkPayoutBlockAttributions } from '~/server/jobs/bulk-payout-block-attributions';
 import { reapDevTunnelsJob } from '~/server/jobs/reap-dev-tunnels';
+import { sweepStaleAgentReviewsJob } from '~/server/jobs/sweep-stale-agent-reviews';
 import { custodySweepJob } from '~/server/jobs/custody-sweep';
 import { reconcileNowpaymentsJob } from '~/server/jobs/reconcile-nowpayments';
 import { notifyStuckCryptoDepositsJob } from '~/server/jobs/notify-stuck-crypto-deposits';
@@ -169,6 +170,7 @@ export const jobs: Job[] = [
   confirmPendingBlockAttributions,
   bulkPayoutBlockAttributions,
   reapDevTunnelsJob,
+  sweepStaleAgentReviewsJob,
   checkImageExistence,
   fullImageExistence,
   rewardsAdImpressions,
--- a/src/server/jobs/__tests__/sweep-stale-agent-reviews.test.ts
+++ b/src/server/jobs/__tests__/sweep-stale-agent-reviews.test.ts
@@ -1,0 +1,56 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * AGENTIC MOD CODE-REVIEW — stale-`running`-report-row sweeper (core fn).
+ *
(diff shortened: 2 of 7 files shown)


ZacxDev:
Adversarial-audit follow-up nits for this change shipped in **#3416** (opened off `main` since this PR is already merged — new commits to its branch wouldn't reach `main`). #3416 addresses: the stale `agentReviewChat` router docblock (no longer proxies to a live pod), deduping `AGENT_REVIEW_CHAT_MODEL` onto `AI_MODELS.CLAUDE_HAIKU`, an explicit sweeper-threshold invariant (must stay > the review Job's ~30m activeDeadlineSeconds), and an AbortSignal through `getTextCompletion` so the 120s chat timeout cancels the underlying fetch. The infra-side RBAC blocker (apps-applier needed `batch/jobs` for the Deployment→Job switch) + template/boot.sh nits landed on talos-infra #695.

PR #3420, opened 2026-07-28
fix(contests): qualify model entries on version date, not model date

JustMaier:
## Problem

Contest entry validation in `validateContestCollectionEntry` compared `Model.createdAt` against the collection's `submissionStartDate`:

```ts
const models = await dbRead.model.findMany({
  where: { id: { in: modelIds }, createdAt: { lt: new Date(metadata.submissionStartDate) } },
});
```

Eligibility is supposed to key on **version** creation — a port counts as long as a new version is published during the submission window. Because the check only looked at the parent model row, anyone adding a Krea 2 version to a model that already had, say, an Anima version was rejected with:

> Some models were created before the submission start date.

Reported by Boerboer (blocked from submitting to the Krea contest) and previously by Ellie, who saw it on the first entries reviewed.

## Fix

A model qualifies when **either**:

- the model itself was created on/after `submissionStartDate` (unchanged behavior), **or**
- it has a `Published` version with `publishedAt >= submissionStartDate` (falling back to the version's `createdAt` when `publishedAt` is null).

Only genuinely stale models — no published version inside the window — are rejected. Draft versions don't qualify, so you can't sneak in an unpublished placeholder. The error message now says what's actually required, and the query gained `select: { id: true }` instead of hydrating full model rows.

Article/image/post checks are unchanged; they have no version concept.

## Testing

- `pnpm typecheck` — no errors in `collection.service.ts`. Remaining errors in the worktree are environmental (per-package `node_modules` not installed, junctioned `@civitai/client` is beta.81 vs the pinned beta.82) and are present without this change.
- Verified the Prisma `modelVersions: { none: ... }` filter shape against `schema.prisma` (`ModelVersion.publishedAt` is nullable, `status` defaults to `Draft`).

### DIFF
--- a/src/server/services/collection.service.ts
+++ b/src/server/services/collection.service.ts
@@ -2198,22 +2198,17 @@ export const validateContestCollectionEntry = async ({
     }
 
     if (modelIds.length > 0) {
-      // A ported model qualifies on its newest version, not the original model row — an old
-      // model with a version published during the window is a valid entry.
+      // A ported model qualifies on its version dates, not the original model row. Keyed on the
+      // version's createdAt rather than publishedAt because publishedAt is reset by the
+      // private-model round trip, which would let an untouched old model back in.
       const models = await dbRead.model.findMany({
         where: {
           id: { in: modelIds },
           createdAt: { lt: new Date(metadata.submissionStartDate) },
           modelVersions: {
             none: {
-              status: ModelStatus.Published,
-              OR: [
-                { publishedAt: { gte: new Date(metadata.submissionStartDate) } },
-                {
-                  publishedAt: null,
-                  createdAt: { gte: new Date(metadata.submissionStartDate) },
-                },
-              ],
+              status: { notIn: [ModelStatus.Deleted, ModelStatus.UnpublishedViolation] },
+              createdAt: { gte: new Date(metadata.submissionStartDate) },
             },
           },
         },
@@ -2222,7 +2217,7 @@ export const validateContestCollectionEntry = async ({
 
       if (models.length > 0) {
         throw throwBadRequestError(
-          `Some models were created before the submission start date. Please only upload models with a version published after the submission period started.`
+          `Some models predate the submission start date and have no version added during the submission period. Add a new version to enter an existing model.`
         );
       }
     }
--- a/src/server/services/collection.service.ts
+++ b/src/server/services/collection.service.ts
@@ -85,6 +85,7 @@ import {
   HomeBlockType,
   ImageIngestionStatus,
   MetricTimeframe,
+  ModelStatus,
   TagTarget,
 } from '~/shared/utils/prisma/enums';
 import { isDefined } from '~/utils/type-guards';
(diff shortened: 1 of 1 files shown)


JustMaier:
## Updated after adversarial review

Two independent reviewers agreed the original diff was correct but too narrow. Revised in 324d610.

**The check now keys on `ModelVersion.createdAt`, not `publishedAt`, and drops the `status = Published` requirement** (excluding only `Deleted` and `UnpublishedViolation`).

Two reasons:

1. **The Published filter still rejected real entrants.** Prod, Krea 2 versions created inside the window on pre-window models: 210 Published, **69 Draft, 3 Unpublished, 1 Scheduled** — 73 versions across ~48 creators still hitting the same error. The asymmetry was indefensible: a *new* model with a draft-only Krea 2 version already gets in today (item 206317023 is exactly that), because the model-row check short-circuits before any version is looked at. Same action on an older model was rejected. This also matches the sibling article/image/post branches, which filter on bare `createdAt` with no status requirement.

2. **`publishedAt` is not trustworthy for this.** Making a model private stamps `publishedAt = now` on every version via an unguarded `updateMany` (`model.service.ts:4168-4178`), while the Private→Public flip runs under the anti-bump guard and is a no-op. So `publishedAt` can record when a model went *private*. Both the false-negative (trained privately before the window, published inside it → wrongly rejected) and the false-positive (private round trip re-stamps an untouched old model → wrongly accepted) disappear when keyed on `createdAt`.

Confirmed clean by review, no action taken: `none:` applies all conditions to the same related row; `ModelVersion` has no `deletedAt`; `gte`/`lt` boundaries; `submissionStartDate` is an ISO-Z string at runtime and `new Date()` handles it; the hash index on `ModelVersion.modelId` serves the `NOT EXISTS`; windowless contests and the new-model path cannot regress (the added clause can only *narrow* the rejected set).

### Not addressed here — needs a product call

**There is no base-model gating on contest entries at all.** `collectionMetadataSchema` has no baseModel/ecosystem field, and nothing in `collection.service.ts` reads `baseModel`. This is pre-existing — item 206317023 is an **Anima** model sitting in the Krea 2 Training Contest right now — but this PR widens it: previously only a genuinely new model row could enter a dated contest, whereas now any old model with any in-window version qualifies. Combined with community-pick scoring, an aged model with mature metrics could be entered against fresh work.

Fix is small and belongs in this same `none:` filter, but it needs a decision on where the allowed base models come from. Flagged to @JustMaier rather than assumed.

### Follow-ups (not blocking)

- No upper bound on the qualifying version (`>= submissionStartDate` with no `<= submissionEndDate`). Low exposure since the window guard blocks late submission.
- Entry validation is one-shot. Nothing re-validates accepted entries, so a qualifying version can be deleted after submission. Pre-existing, but it now depends on mutable child rows rather than immutable `Model.createdAt`.
- `Model.publishedAt` staying frozen means ported entries render as "Updated" rather than "New" on cards. Cosmetic; fixing it means touching a load-bearing anti-bump invariant.

### Data impact

No cleanup needed — validation throws before any `CollectionItem` write, so blocked attempts left no rows. The 3 REJECTED items in the contest are manual moderator rejections, unrelated. **263 models across 164 creators** are eligible-but-not-yet-entered once this ships.

PR #3421, opened 2026-07-28
perf(db): set Prisma errorFormat 'minimal' in production

ZacxDev:
Prisma’s default `errorFormat` (`colorless`) makes the client construct a **`new Error` on every model-method invocation** (`prisma.x.findMany(...)`) purely to capture a callsite for source-annotated error rendering.

That capture is **wasted work in production**. The client’s error renderer early-returns before ever calling `callsite.getLocation()`:

```js
if (!e || typeof window < "u" || process.env.NODE_ENV === "production") return s;
let a = e.getLocation();   // unreachable in prod
```

so the stack is captured and then unconditionally discarded.

### Why it is worth changing

Measured on a production SSR process with a V8 **allocation-sampling** profile (`HeapProfiler.startSampling` with `includeObjectsCollectedByMinorGC`, so short-lived garbage is counted):

| | |
|---|---|
| Prisma runtime frame doing the capture | **~15% of ALL bytes allocated** — the single largest allocation source |
| Same frame in a CPU profile of that tier | ~2.3% of on-CPU JS time |
| Reproduced | yes — two independent processes, ranking stable within ~1pp |

Allocation volume matters here because the SSR tier’s GC is **scavenge-dominated** (~93% of GC time is young-generation), so bytes allocated convert almost directly into GC CPU.

### Behaviour change

`errorFormat: minimal` swaps the stack-capturing callsite for a no-op one.

- `code`, `meta` and `message` on `PrismaClientKnownRequestError` are **unchanged**.
- Only the source-location decoration is dropped — and it is already unreachable in production.
- The rich default is preserved **outside** production, so local/dev error output still points at the offending source line.

### Test

`packages/civitai-db/src/client.errorFormat.test.ts` covers both branches, including that the non-prod branch returns an **absent** key rather than `errorFormat: undefined` (spreading an explicit `undefined` would override Prisma’s own default).

### DIFF
--- a/packages/civitai-db/src/client.errorFormat.test.ts
+++ b/packages/civitai-db/src/client.errorFormat.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from 'vitest';
+import { buildErrorFormatOption } from './client';
+
+// Prisma's non-'minimal' errorFormat makes the client capture a stack trace (`new Error`) on every
+// model-method call, just to annotate errors with a source location. In production that annotation
+// is never rendered (the renderer early-returns on NODE_ENV === 'production'), so the capture is
+// pure waste — it was the largest single allocation source in a production SSR allocation profile.
+describe('buildErrorFormatOption', () => {
+  it("uses 'minimal' in production so Prisma skips the per-call callsite capture", () => {
+    expect(buildErrorFormatOption(true)).toEqual({ errorFormat: 'minimal' });
+  });
+
+  it('leaves errorFormat at the Prisma default outside production for readable local errors', () => {
+    // Must be an ABSENT key, not `errorFormat: undefined` — spreading an explicit `undefined`
+    // into the client options would override Prisma's own default.
+    const opts = buildErrorFormatOption(false);
+    expect(opts).toEqual({});
+    expect('errorFormat' in opts).toBe(false);
+  });
+});
--- a/packages/civitai-db/src/client.ts
+++ b/packages/civitai-db/src/client.ts
@@ -12,6 +12,26 @@ export type CreatePrismaClientsOptions = Partial<DbConfig> & {
 };
 
 /**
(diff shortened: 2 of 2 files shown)


ZacxDev:
Closing — the premise is wrong and I should have checked it before opening this.

I based this on reading `Ze()` (the callsite factory) and the `Sd()` early-return, and inferred that because we pass no `errorFormat`, Prisma would default to `colorless` and capture a stack on every call. I never read the constructor's default resolution, which is:

```js
l.errorFormat ? this._errorFormat = l.errorFormat
  : process.env.NODE_ENV === "production" ? this._errorFormat = "minimal"
  : ...
```

Prisma already defaults to `minimal` under `NODE_ENV=production`, which every one of our production images sets. So production constructs **zero** callsite `Error` objects per Prisma call today — this change buys nothing in prod and would only cost us the callsite/source-context annotation in local dev.

Verified empirically with our exact client options: `NODE_ENV=production` → resolved `_errorFormat="minimal"`, 0 `Error` constructions per `findMany`/`findFirst`; `development` → `colorless`, 1–2 per call.

The ~15%-of-allocation production frame is real, but it is **not** the stack capture. It is the fluent-API wrapper `uo()`, which rebuilds a model's field map on every property access with an O(F²) spread-in-reduce:

```js
let a = e._runtimeDataModel.models[r].fields.reduce((l,u)=>({...l,[u.name]:u}),{});
```

for `findUnique | findUniqueOrThrow | findFirst | findFirstOrThrow | create | update | upsert | delete`. Measured cost scales quadratically with field count — ~3.1 MB and ~1.1 ms per call on a 186-field model vs ~1.5 KB / ~2 µs for `findMany`. That reconciles with the observed per-process rate; the stack-capture hypothesis would have required ~17k Prisma calls/sec/process, which is implausible by orders of magnitude.

Pursuing that properly in a separate change.

PR #3425, opened 2026-07-28
feat(contests): gate entries on allowed base models

JustMaier:
Follows #3420. Branched off `main` after that merged, not stacked on it.

## Problem

There is no base-model gating on contest entries anywhere. `collectionMetadataSchema` had no baseModel field and nothing in `collection.service.ts` read `baseModel` — a contest could not express "this is a Krea 2 contest" as an enforceable rule.

#3420 makes this more pressing. Before it, the only way into a dated Model contest was a model row created inside the window — a genuinely new upload. After it, any older model with an in-window version qualifies. So a two-year-old SDXL LoRA with mature reactions and metrics can be entered by pushing one throwaway version during the submission window, competing against genuinely fresh work. That matters directly given community-pick scoring.

## Change

**Schema** — `metadata.baseModels: string[]` on `collectionMetadataSchema`, validated against the known base-model list. Empty or absent means no gating, so **every existing contest is unaffected**.

**Enforcement** — the date and base-model conditions live in a single `modelVersions: { none: ... }` object, so one version has to satisfy both:

```ts
const qualifyingVersion: Prisma.ModelVersionWhereInput = {
  status: { notIn: [ModelStatus.Deleted, ModelStatus.UnpublishedViolation] },
  ...(submissionStartDate ? { createdAt: { gte: submissionStartDate } } : {}),
  ...(allowedBaseModels.length > 0 ? { baseModel: { in: allowedBaseModels } } : {}),
};
```

Splitting them across two predicates would let an old Krea 2 version satisfy the base-model rule while a throwaway new SDXL version satisfies the date rule — exactly the case being closed.

Gating also applies to **windowless** contests. The old date check was nested under `if (metadata.submissionStartDate)`; the block now runs whenever either rule is configured. The model-row `createdAt: { lt: start }` shortcut is kept only when there is no base-model gating, which is what preserves existing behavior byte-for-byte.

**Moderators are not exempt**, consistent with the surrounding date checks (`maxItemsPerUser` exempts them, but that is an anti-spam quota, not eligibility). A mod wanting an exception edits the contest's allowed list.

**UI** — `InputMultiSelect` in `CollectionEditModal`, contest mode only, fed from `activeBaseModels`. Plus an "Allowed base models" line in the submission-period popover on the collection page.

## Review notes

Adversarially reviewed. Findings applied:

- **The model-row date filter is now guarded on `submissionStartDate` being set.** It was unreachable with `undefined` today, but Prisma's `lt` is optional so `undefined` type-checks cleanly — anyone loosening the outer guard later would silently drop the filter and let every model pass, with no compile error and no test failure.
- **`baseModels` values are validated.** They must match `ModelVersion.baseModel` exactly; `'FluxKrea'` or `'sdxl 1.0'` would have persisted fine, matched zero versions, and locked the contest to zero entries while reporting a plausible-looking error naming the bad value.
- **Field description spells out** that with a submission window, the *same* version must satisfy both rules.

Confirmed clean by review: `none:` keeps all conditions on one row; every conditional-spread combination is a valid where-input; `[]`, `['']` and `undefined` all mean no gating; `activeBaseModels` emits raw base-model strings matching the column; the field round-trips through `upsertCollection`; the popover change cannot break contests that have dates but no baseModels.

## Tests

8 new tests in `contest-entry-base-model-gate.test.ts` locking the emitted `where` shape. The important one asserts the **no-gating path is unchanged** — a live contest with 105 accepted entries depends on it, and until now that rested on manual diff-reading. Others cover the single-version invariant, windowless gating, the skip case, and each error-message branch.

## Known gaps

- **Draft versions still qualify.** Requiring `Published` would re-block the ~69 in-window draft Krea 2 versions across ~47 creators that #3420 deliberately unblocked, and a *new* model with a draft-only version already gets in today. Entries pass through moderator review anyway. Worth revisiting if draft-stub abuse actually appears.
- `activeBaseModels` hides some real, populated base models (SD 3.5 family, SDXL Turbo, Kling, Sora 2, …), so those cannot be picked as a contest target from the UI.
- **Gating only runs at submission.** Setting `baseModels` on a live contest will not re-validate entries already accepted.


JustMaier:
## Fixed: the picker was reading the wrong constants file

`Krea 2` was missing from the base-model list — the contest this was built for could not have been configured, and the zod refine would have rejected the value if set via the API.

There are two base-model constants files:

| File | Importers | Has `Krea 2` |
|---|---|---|
| `src/shared/constants/basemodel.constants.ts` | 95 | ✅ |
| `src/shared/constants/base-model.constants.ts` (legacy) | 4 | ❌ |

I imported the legacy one. It predates the current ecosystem records — `basemodel.constants.ts` even carries a section header reading *"Compatibility Layer for base-model.constants.ts Migration"*, so the old file was already on its way out.

Prod confirms which names are real: `Krea 2` = 11,972 versions, `Flux.1 Krea` = 498, `Krea2` = 8. Only `Flux.1 Krea` existed in the legacy list.

Fixed in 932bfc0 by pointing both the schema and the picker at `basemodel.constants.ts`, migrating the two remaining importers (`AppSettingsModal.tsx`, `scripts/local-dev/gen_seed.ts`), and **deleting the legacy file** — no references remain.

Also switched the picker from `activeBaseModels` to the full `baseModels`. `hidden` means "not offered on upload", which is a different question from whether a contest can target it — SD 3.5, SDXL Turbo, Kling and Sora 2 are all populated and were unselectable. This field is moderator-only, so the wider list is the right default and it closes the gap flagged in review.

Typecheck is clean and the 8 gate tests still pass.

PR #3431, opened 2026-07-28
chore(basemodel): retire dead compat-layer exports

JustMaier:
Cleanup split out of #3425. Independent of it — the file sets are disjoint, so these can merge in either order.

## Background

There were two base-model constants files. The legacy `src/shared/constants/base-model.constants.ts` had drifted — its list has `Flux.1 Krea` but not `Krea 2`, which is 11,972 `ModelVersion` rows in prod. That surfaced when the contest base-model picker read the stale list. **#3425 deletes that file**; this PR cleans up what it left behind.

`basemodel.constants.ts` carries a section headed *"Compatibility Layer for base-model.constants.ts Migration"* — string-keyed exports mirroring the old API while the module's real model is numeric IDs. Some of it never had a consumer.

## Change

Deletes 9 exports with zero consumers, **141 lines**:

| Export | Evidence |
|---|---|
| `getBaseModelGroupsByMediaType` | only self-definition |
| `getBaseModelConfigsByMediaType` | only self + internal refs |
| `getBaseModelConfigsByGroup` | only self + one in-file caller (inlined) |
| `getBaseModelByMediaType` | only self + internal ref |
| `baseModelGroupConfig` | only self + doc mentions of the *old* file |
| `getBaseModelGenerationSupported` | **zero hits anywhere** — dead on arrival |
| `getBaseModelsByModelType` | only self-definition |
| `getGenerationBaseModelsByMediaType` | only self-definition |
| `getGenerationBaseModelConfigs` | only self + doc mentions |

`getBaseModelConfigsByGroup` had one live in-file caller, `getBaseModelsByGroup` (26 call sites), so its two-line body was inlined rather than left as a wrapper. Behavior is identical — same `ecosystemByKey.get`, same `getBaseModelsByEcosystemId(id, false)` (the `false` matters: it controls which models are included), same ordering. The only difference is the miss path returning a `[]` literal instead of `[].map(...)`, both fresh empty arrays.

Also drops one checklist line from `docs/adding-basemodel-ecosystem-guide.md` that named `getGenerationBaseModelConfigs`. The line above it covers the same verification step via `getBaseModelGenerationConfig()`, the live non-compat function — so the checklist has no hole.

## Review

Adversarially reviewed, zero blocking findings. Verification covered the paths a naive `src/`-scoped grep misses:

- **`.claude/skills/**`** — `add-ecosystem`, `add-generation-support`, `add-training-support` and `ecosystem-seo-page` all edit this file for a living. None references any deleted function, so the ecosystem-onboarding workflow is intact.
- No namespace imports of the module exist, so string-keyed access (`constants['getFoo']`) is impossible
- No barrel re-export, no `export *`, no dynamic `import()`, no alias under another name
- Every identifier the deleted functions used is still live in-file; no orphaned imports or helpers

**Explicitly left alone:** `baseModels`, `activeBaseModels`, `baseModelGroups`, `defaultBaseModel`, the `BaseModel`/`BaseModelGroup` type aliases, and the seven compat functions that do have consumers (`getBaseModelsByGroup` 26, `getBaseModelGroup` 18, `DEPRECATED_BASE_MODELS` 11, `getBaseModelMediaType` 8, `getBaseModelEngine` 3, `getBaseModelSeoName` 2, `getBaseModelConfig` 2). Retiring those is a ~70-call-site migration tracked separately.

## Testing

`pnpm typecheck` — 8 errors, all pre-existing and environmental (missing `kysely` in two workspace packages, `@civitai/client` version drift in `mage-flow.handler.ts`), none attributable to this change.

`pnpm exec vitest run src/shared/` — 15 files, 176 tests, all pass. Worth stating plainly: **none of those tests cover `basemodel.constants.ts`**, so that run only shows nothing else regressed. The real evidence here is the typecheck plus the exhaustive reference sweep.

## Follow-up

Two docs cite names that disappear with #3425's file deletion, not with this change: `docs/base-model-constants-management.md:44` and `docs/plans/generator-filtering-options.md`. Being handled in #3425.

### DIFF
--- a/docs/adding-basemodel-ecosystem-guide.md
+++ b/docs/adding-basemodel-ecosystem-guide.md
@@ -715,7 +715,6 @@ After adding a new model, verify:
 - [ ] Model appears in `activeBaseModels` if not hidden
 - [ ] Ecosystem appears in `baseModelGroups` array export
 - [ ] `getBaseModelGenerationConfig()` includes new ecosystem (if has generation support)
-- [ ] `getGenerationBaseModelConfigs()` includes new ecosystem (if has generation support)
 
 ### Documentation
 - [ ] Updated this guide with new patterns (if introducing new pattern)
--- a/docs/adding-new-base-models-to-generation.md
+++ b/docs/adding-new-base-models-to-generation.md
@@ -17,7 +17,7 @@ Before a base model can be used for generation, it must first be defined in the 
 ### 1. Define Base Model in Constants
 
 #### Add to Base Model Configuration
-**File**: `src/shared/constants/base-model.constants.ts`
+**File**: `src/shared/constants/basemodel.constants.ts`
 
 Add a new entry to the `baseModelConfig` array. The structure determines the base model's basic properties:
 
@@ -92,7 +92,7 @@ Once the base model is defined in the system, you can enable it for generation w
 ### 4. Define Base Model Generation Support
 
 #### Add to Base Model Generation Configuration
-**File**: `src/shared/constants/base-model.constants.ts`
+**File**: `src/shared/constants/basemodel.constants.ts`
 
 Add a new configuration entry to the `baseModelGenerationConfig` array:
 
@@ -294,7 +294,7 @@ This commit implemented Part 1 of the process:
 - Added 'CHR' badge indicator for the UI
 
 **Files Changed:**
-- `src/shared/constants/base-model.constants.ts` - Added base model definition
+- `src/shared/constants/basemodel.constants.ts` - Added base model definition
 - `src/server/common/constants.ts` - Added license mapping
 - `src/components/Model/ModelTypeBadge/ModelTypeBadge.tsx` - Added badge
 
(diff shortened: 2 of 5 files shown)


maps524:
Rebased onto main now that #3425 has merged, and added a second commit repointing live references at `basemodel.constants.ts`.

The one that mattered: `docs/adding-new-base-models-to-generation.md` still instructed a dev to edit `base-model.constants.ts` in four places. Following that guide after #3425 would have **recreated the deleted file and re-forked the list** — the exact failure this whole thread started with. Also fixed the file inventory in `docs/features/generator.md` and a dangling comment pointer at `src/server/services/blocks/workflow.service.ts:320`.

Historical plan docs (`ai-toolkit-training-implementation-plan.md`, `feed-dependency-removal-plan.md`, `moderator-app-package-extraction-plan.md`, `plans/generator-filtering-options.md`) still name the old path. Left alone deliberately — they describe what was proposed at the time, and rewriting them would misrepresent that.

Typecheck after rebase: 8 errors, all the known environmental ones, none from this branch.

## Follow-ups found during review, not fixed here

1. `src/components/Apps/AppSettingsModal.tsx:91` has `'WanVideo'` in its `BASE_MODEL_OPTIONS` chip list. The real record name is `'Wan Video'` (with a space), so that chip has **never rendered** — it fails the membership filter. Pre-existing, unrelated to this work, and fixing it changes what that settings UI shows, so it wants its own PR.
2. The contest base-model picker is unfiltered by media type, so a moderator configuring an image contest can select `ACE Audio` or `Tripo` and silently lock it to zero eligible entries. Moderator-only and trivially reversible, but a media-type filter keyed off collection type is the right hardening.

PR #3442, opened 2026-07-28
feat(collections): community-pick scoring for contest collections

JustMaier:
Adds a **Community** tab to the contest collection review page so moderators can rank entries by community engagement alongside the existing staff-pick scorer.

## How it scores

Each entry is scored by counting **distinct qualified users per signal** inside the contest window, normalized within its category tag, then weighted. Distinct-user counting is the structural anti-cheat: one account contributes at most 1 to any signal, so volume farming collapses to 1.

Signals: on-site image authors (`ImageResourceNew` ⋈ `Image`), reactors on those images (ClickHouse `reactions`), downloaders (`modelVersionEvents`), generating users (`orchestration.jobs`), and collectors (`CollectionItem`). Comments and tips are deliberately unweighted — both launder trivially.

A user is disqualified for an entry when they are the model's creator or the submitter, are in `metricExcludedUsers`, are contest-banned, are banned or deleted, or registered after the account-age cutoff. Every entry reports raw vs qualified per signal so a farmed entry surfaces for human review rather than being auto-disqualified.

Categories are separate contests — normalization and ranking never cross a category boundary.

## Notable details

- **All tuning values live in a `KeyValue` row**, not in source. Weights, thresholds and denominators are read at runtime, scoped per collection with a global fallback, and edited through a moderator-only procedure that is separate from the scoring payload.
- **Counting happens in the database.** No user-id arrays cross into Node; the age gate is pushed down as a monotonic id predicate rather than enumerating accounts.
- **Runs are async.** A run is enqueued, its state lives in Redis with a TTL, and a signal tells the UI to refetch. The signal payload carries run bookkeeping only.
- **Snapshots** freeze a ranking with its resolved config as the auditable artifact; re-running the same window reproduces the same result.
- Config edits are recorded append-only in the same transaction as the write.

## Testing

Validated against production data on a live contest (105 accepted entries, 4 categories, ~7.5k engagers): per-entry counts reconciled against an independently-built oracle, including a fan-out case where one image references two entered models. Typecheck, eslint and prettier clean.

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -2185,6 +2185,7 @@ export const REDIS_KEYS = {
     // `setUserSetting`; `CacheTTL.md` backstops any other writer. See
     // `getUserMetricPrivacyDefaultsMap` in creator-membership.service.
     USER_METRIC_PRIVACY_DEFAULTS: 'packed:caches:user-metric-privacy-defaults',
+    CONTEST_COMMUNITY_SCORE: 'packed:caches:contest-community-score',
   },
   RESEARCH: {
     RATINGS_COUNT: 'research:ratings-count',
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -2186,6 +2186,10 @@ export const REDIS_KEYS = {
     // `getUserMetricPrivacyDefaultsMap` in creator-membership.service.
     USER_METRIC_PRIVACY_DEFAULTS: 'packed:caches:user-metric-privacy-defaults',
     CONTEST_COMMUNITY_SCORE: 'packed:caches:contest-community-score',
+    // Async contest-scoring runs: per-run state, the per-collection pointer to the
+    // most recent run, and the scored result each run produces. Every key carries a
+    // TTL, so the whole namespace self-cleans and no table backs it.
+    CONTEST_SCORE_RUN: 'packed:caches:contest-score-run',
   },
   RESEARCH: {
     RATINGS_COUNT: 'research:ratings-count',
--- a/src/components/Collections/components/ContestCollections/ContestCommunityScore.tsx
+++ b/src/components/Collections/components/ContestCollections/ContestCommunityScore.tsx
@@ -166,6 +166,14 @@ function CategoryTable({
             No qualified engagement in this category — every entry is tied, so no ranking is shown.
           </Alert>
         )}
+        {category.truncated && (
+          <Alert color="red" title="This category is incomplete — ranks withheld">
+            {category.missingCount} {category.missingCount === 1 ? 'entry is' : 'entries are'}{' '}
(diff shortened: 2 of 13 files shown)


maps524:
**Preview build blocked — Turbopack chunk collision.** Marking draft until resolved; the scoring work itself is reviewed and validated against prod.

```
[output]/.next/server/chunks/ssr/[root-of-the-server]__0r_ce-0._.js
Two or more assets with different content were emitted to the same output path
  [output]/.next/6f39a529c180c1d8.js
  [output]/.next/9b500d0b1afad1d7.js
```

Deterministic — identical chunk hashes on every run. Compiles fine outside the container; only fails in the image build.

**Bisected (each result verified by its actual error text, not just status):**

| Tree | Result |
|---|---|
| identical to `main` | pass |
| + shared-file edits (`distributed-lock`, `constants`, `enums`, redis key) | pass |
| + contest server files present but router **unregistered** | pass |
| + router **registered** (service reachable), no components | **fail** |
| full branch | **fail** |

So the trigger is our server graph becoming reachable. The client panel is exonerated — `dynamic(..., { ssr: false })` on it changed nothing.

**Tried, did not fix:** rebasing onto current main; `ssr: false` on the panel; replacing `p-limit` (ESM-only) with an inline limiter; lazy `import()` of the service inside each procedure.

**Untried leads:** split the ~2k-line service into smaller modules; disable `serverSourceMaps`; check for a known Turbopack chunk-naming bug upstream.

maps524:
**Update — code audit came back clean; the colliding assets are not ours.**

Audited the branch directly at @JustMaier's suggestion:
- No import cycles originating in our files (the 465 in the repo are pre-existing, through `session.ts`/`user.schema`/`model.schema`).
- Our server-to-client chain (`clickhouse/client -> tracker -> jobs/entity-moderation -> user.service -> auction.notifications -> AuctionProvider.tsx`) is shared by **16 other services**, so it is normal here.
- Every service import is ordinary; the redis shim is imported one way only.

**Key evidence:** the colliding chunk hashes are byte-identical on every run —

```
[output]/.next/6f39a529c180c1d8.js
[output]/.next/9b500d0b1afad1d7.js
```

— *including after renaming `contest-score.service.ts` -> `contest-scoring.service.ts`*. Those hashes are content-derived, so the two colliding assets are unaffected by our module identity. They are two different versions of `[root-of-the-server]__0r_ce-0._.js`, i.e. the SSR entry emitted twice. Our server graph triggers the double emission; it is not the content of either copy.

**Also tried, no effect:** rebase onto current main; `dynamic(..., { ssr: false })` on the panel; replacing ESM-only `p-limit` with an inline limiter; lazy `import()` of the service inside every procedure; renaming the service module.

**Suggested next steps for someone with Turbopack context:** diff the two dumped chunk copies from inside a failing builder (they are written to `.next/<hash>.js` at failure time); try `serverSourceMaps: false`; split the ~2k-line service; or check upstream for a known chunk-naming bug.

Branch restored to the real work — no test commits remain.

PR #3450, opened 2026-07-29
docs: trim CLAUDE.md 312->231 lines, replace stale facts with load-bearing ones

AKharytonchyk:
Companion to #3439 rather than an overlap with it. `CONTRIBUTING.md` is for a person who chooses to open it; `CLAUDE.md` is loaded into an agent's context automatically on every session, so a wrong fact in it gets acted on rather than read past — and every line costs tokens on every request.

**312 → 231 lines** (~16.1 KB → ~12 KB), net `+99 / -180`. Docs only, no code paths touched.

## The one that actually caused damage

The file said **"Format code: `pnpm run prettier:write`"**. That script ignores arguments and reformats the whole repo, and 789 of 4,116 `src` files fail `prettier --check` today — so an agent following the auto-loaded instruction produces a ~789-file diff. `CONTRIBUTING.md` explicitly warns against it while `CLAUDE.md` recommended it, and after the first pass of this PR the file contradicted *itself* in two places. Both now say the same thing:

> Format **only what you touched**: `pnpm exec prettier --write <files>`. Never `pnpm prettier:write`.

## Wrong facts removed

- **`**Authentication**: NextAuth`** — `next-auth` is not a dependency. Auth is `src/server/auth/` plus `packages/civitai-auth`.
- **Three `src/server/auth/` paths** (`next-auth-options.ts`, `session-user.ts`, `token-refresh.ts`) — none exist. The directory holds `get-server-auth-session.ts`, `bearer-token.ts`, `civ-cookie.ts`.
- **`dialog-registry.ts` or `dialog-registry2.ts`** — only `dialog-registry2.ts` exists, so the "or" was a coin flip.
- **"Optimize images with `next/image`"** — the repo uses `EdgeMedia`/`EdgeImage`, which the same document said two sections earlier.
- **`Next.js 14`** — it's `^16.2.7`, which the file's own "never put tests under `src/pages`" rule already assumed.
- **`schema prisma/schema.prisma`** — the schema is `packages/civitai-db-schema/prisma/schema.full.prisma`. The `schema.prisma` beside it is generated by `generate-slim-schema.js`, which strips `@no-type` models, so edits to it are silently discarded by the next `db:generate` (which also runs on `postinstall`).

## The 17-line architecture map is gone

It listed ~45 paths under `src/server/`. Three were already dead four months after being added, and a *wrong* path is worse than no path: the agent burns a failed Read and may conclude the subsystem doesn't exist. `Glob` answers "where do routers live" in one call and can't be stale. This is the same silent-rot argument the document itself makes about code comments — the map was a 17-line comment on the filesystem.

What replaced it is only things a search can't tell you:

- **Reads use `dbRead`, writes `dbWrite`**, and a read that must observe a just-committed write goes through `db/db-lag-helpers.ts`. 449 files use `dbRead`; the split wasn't mentioned at all.
- **No awaited I/O inside `db.$transaction`.** `local-rules/no-io-in-transaction` catches it but is set to `warn`, so `pnpm lint` stays green and only review catches it.
- **The image feed has two backends** — `getAllImages` (Postgres) and `getAllImagesIndex` (Meilisearch), chosen in `image.controller.ts` behind a Flipt flag. A filtering or NSFW change applied to one silently misses the other.
- **A search-index field is two edits** — declared in `onIndexSetup`'s attribute lists *and* emitted from the transformer — and existing documents aren't backfilled.
- **Feature flags live in exactly one file**, so there's no reason to invent a `NEXT_PUBLIC_` env var.
- Service files are 4-8K lines (`image.service.ts` is ~100K tokens), so **grep for the symbol; don't read the file**.

## Added: which router

**Pages Router, no `app/` directory, no server components, no `"use client"`.** The file said "Next.js 16" and nothing else, so an agent reasonably assumes App Router and writes RSCs — all of it discardable. This was the single highest-value omission.

Also kept from the first commit: the monorepo layout (7 `apps/`, 14 `packages/`, `event-engine-common` submodule), and the fact that **adding a `@civitai/*` package means editing three resolution lists** — `tsconfig.json` `paths`, `next.config.mjs` `transpilePackages`, and `vitest.config.mts`'s own array, because vitest doesn't read tsconfig paths. They don't list the same packages today, and missing the third makes suites fail to *collect* rather than fail, which reads as unrelated breakage. `vitest.config.mts` carries scar-tissue comments about exactly that.

## On the overlap with CONTRIBUTING.md

Deliberately not deduplicated. `CONTRIBUTING.md` exists for contributors who don't use an agent at all, so it has to stand alone — and `CLAUDE.md` is the file that's actually in context, so moving the operational essentials out of it (the submodule step, what the suites are, the collection-failure warning) would remove them from the only place they get read automatically. I point at `CONTRIBUTING.md` for the baseline-comparison method and keep the essentials inline. A few duplicated lines is the cheaper failure mode than an agent that doesn't know to look.

## Verified

Every claim checked against the tree rather than recalled — including the removals: `next-auth` absent from all `package.json` files, the three auth paths and `dialog-registry.ts` confirmed missing, `.nvmrc` = `24.18.0`, no `src/app` directory, `db-lag-helpers.ts` present, `no-io-in-transaction` severity `'warn'` in `.eslintrc.js`, `getAllImages` and `getAllImagesIndex` both present in `image.service.ts`, and all three `@civitai/*` resolution lists present in their respective configs.

One number I chose to drop rather than update: the old map's router count was wrong, and replacing one stale count with another just resets the clock.


AKharytonchyk:
Closing as superseded.

This branch was cut when `CLAUDE.md` was 312 lines and proposed trimming it to 231. `main` has since taken the same file to **664 lines**, and it already contains everything this PR added:

- **Monorepo Layout** — `apps/`, `packages/civitai-*`, the `event-engine-common` submodule
- **Server-Side Architecture Map** — the `src/server/` domain index
- **Comments** — the "never narrate the what" / keep-test guidance, verbatim
- The stale facts this PR corrected are fixed on `main` too: Next.js 16, `@civitai/auth` replacing NextAuth, the real test-suite commands

`main` also gained a large amount this PR never had (the SvelteKit app standard, the `--project 'unit*'` trap, the Prettier-scope rules, the `svelte-kit sync` incident), all of it hard-won. Re-applying a 664→~250 trim now would delete that deliberately, which is a different and much bigger proposal than what this PR was opened to do.

The one item on this branch not literally on `main` is "Adding a `@civitai/*` package means editing three files" — `main` covers that by reference instead, pointing at the `scaffold-civitai-app` skill and `docs/packages/new-app-integration.md`.

No content is being lost. Happy to reopen a focused PR if a specific section still reads wrong.

PR #3457, opened 2026-07-29
feat(apps): bigger store covers + larger CTAs + a branded, animated app launch

ZacxDev:
Three pieces of App-marketplace UX feedback, all behind the existing `/apps` flag gate (no access change, no schema change, no server change).

> 1. Make app cover images larger — fewer columns per row?
> 2. Make the Visit and Open CTAs larger.
> 3. Launching an app should feel magical and be a delightful, intuitive, simple and subtly animated experience.

## 1 — Bigger covers

**Grid** (`AppListingsMarketplaceBody`): 5 columns → 4 at `xl` (span `2.4` → `3`). Every other breakpoint is unchanged (base 12 / sm 6 / md 4 / lg 3) and **the store container width is unchanged at 1600** — the extra card width comes from the column-count drop, not from widening the page. Both numbers now live together in a new `appListingGrid.ts` so the container and the grid are pinned as a pair and can't drift apart.

**Cover** (`AppListingCard`): the fixed `h={140}` letterbox became a responsive **16:9** box, so the art actually grows with the wider column instead of staying a short strip. The cover is already served at `width: 1200`, so there is ~8× resolution headroom — **this is purely client-side: no DTO, service, or pipeline change.**

CLS is preserved, not traded away: `aspect-ratio` derives the box height from the (already-known) column width *before* the image loads, and the `onError` category-glyph placeholder renders inside the **same** box, so a dangling cover URL swaps art with zero reflow. The placeholder stays `aria-hidden`.

## 2 — Bigger CTAs

The kind-aware primary CTA and the owner-only secondary **Edit** both move one notch up Mantine's size scale (`xs` → `sm`), with icon sizes to match. Because `sm` buttons are taller/wider, the outer action row is now allowed to wrap (actions drop to their own line on a narrow column) rather than overflowing the card; the inner action group stays `nowrap` so Edit and the CTA never split.

The CTA *logic* is untouched and test-pinned: external **Visit** is still a new-tab anchor with `rel="noopener noreferrer"`, `canOpenPage=false` still falls back to **View details**, and the owner Edit deep-link + owner-only "Incomplete" badge still render exactly as before.

## 3 — A branded, animated launch

`/apps/run/<slug>` previously showed a bare spinner and then popped the block in. It now shows a **branded launch state** — the app's initial in the same avatar treatment the store card uses, plus its name ("Starting …") — which **cross-fades out as the block fades and settles in** on `BLOCK_READY`. That gives a visual through-line from card → run page.

**CSS, not `motion`.** `motion` is a dependency, but it is **not in the `/apps` route graph**: its only importers are `src/components/Chat/*` and `src/utils/lazy-motion.ts`, and Chat is reached through a `next/dynamic` `ChatWindow` chunk. Importing it here would pull a new animation runtime into the run-page bundle for one cross-fade, so the reveal is a plain CSS opacity/transform transition.

**`prefers-reduced-motion: reduce` collapses it to 0ms** — no transition is emitted on either the veil or the block, and the veil is dropped in the same commit the block turns ready.

### 🔴 The reveal cannot gate the error path

The host has **four** terminal states — `timeout` (the ~10s BLOCK_READY timer), `fatal`, `no_token` (the ~15s token wait) and `error` (a hard mint failure). Each one flips `showIframe` false, which unmounts the entire iframe + veil branch and renders `BlockFallback` in the same commit; the fade-out timer is simply cleaned up. Nothing about the fallback render reads animation state, and `AppBlockChrome` still wraps **every** state (the FRAME-1 invariant). The manifest's `iframe.minHeight` reserve is untouched — no grow-jump is reintroduced.

This is regression-tested for all four terminals, including the sharpest case: an error that lands **during** the cross-fade window, asserted to reach the fallback within a window far shorter than the reveal duration.

## Also: a latent analytics bug this work uncovered

The run-page render/impression beacon used to fire from **inside** the `BLOCK_READY` handler, behind a flag set by the `setStatus` **updater**:

```ts
let acked = false;
setStatus((current) => { if (current === 'loading') { acked = true; return 'ready'; } return current; });
if (acked) { …sendBlockRender… }
```

That only works while React evaluates the updater **eagerly**, which it does only when the fiber has no other pending update. As soon as *any* unrelated state update is queued when `BLOCK_READY` lands, React defers the updater, `acked` is still `false` at the `if`, and **the impression is silently dropped**. The host has plenty of such updates in the real world (token rotation, the image-scan poller list) — this was latent analytics loss, and adding one more post-mount state update reproduced it **deterministically, 5/5 runs**.

The beacon now keys off the **committed** `status`, mirroring the render-FAILURE beacon effect directly below it, which is batching-independent. `notifyReady()` is now called unconditionally (it is documented-idempotent) for the same reason.

**Known sibling, deliberately not fixed here:** `IframeHost` (the model-slot host) has the identical in-handler pattern and additionally gates `applyHeight` on it, so it carries the same bug. Different surface, and the height semantics deserve their own change.

## Testing

- **New blocking unit suite** `appListingGrid.test.ts` pinning the span per breakpoint and the container width.
- **Extended** `AppListingCard` / `AppListingsMarketplaceBody` browser suites (not replaced): cover inside the 16:9 box; `onError` placeholder fills the same box and stays decorative; both CTAs render at `sm`; external Visit keeps its new-tab anchor; the retired 5-column basis is gone from the generated grid CSS.
- **New** `PageBlockHostLaunchReveal.browser.test.tsx`: branded loading, reveal on ready, sanitizer applied to the visible launch copy, reduced-motion emits no transitions, and each terminal state still reaches `BlockFallback` inside `AppBlockChrome`.
- **Every new guard was mutation-verified** (break it, confirm a specific test fails, revert): 17 mutations, 16 killed, 1 confirmed equivalent.
- `tsc --noEmit` clean; the touched files are lint-clean, and a drive-by removed the last `consistent-type-imports` error in `src/components/AppBlocks`.

## Verification status

`/apps` is mod-gated, so this was **not** visually verified — please drive it with a logged-in mod session. Manual steps are in the review thread.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -4,6 +4,11 @@ import { page } from 'vitest/browser';
 import { useDialogStore } from '~/components/Dialog/dialogStore';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
 import { renderWithProviders } from '../../../test/component-setup';
+// Type-only (erased at runtime, so it can't defeat vi.mock hoisting) — lets the
+// `importOriginal` generic below be written without an inline `import()` type,
+// which the repo's `consistent-type-imports` rule forbids (this file was the
+// last lint error in src/components/AppBlocks).
+import type * as MantineNotifications from '@mantine/notifications';
 
 // PageBlockHost wires the money-path workflow bridge AND the storage bridge,
 // which call `trpc.blocks.*.useMutation()`, `trpc.apps.storage.*.useMutation()`,
@@ -21,7 +26,7 @@ vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
 // mounting the full Notifications provider. Preserve the module's other exports.
 const showNotificationSpy = vi.fn();
 vi.mock('@mantine/notifications', async (importOriginal) => {
-  const actual = await importOriginal<typeof import('@mantine/notifications')>();
+  const actual = await importOriginal<typeof MantineNotifications>();
   return { ...actual, showNotification: (args: unknown) => showNotificationSpy(args) };
 });
 
@@ -571,7 +576,15 @@ describe('PageBlockHost Retry (Task: re-attempt load from terminal fallback)', (
     expect(page.getByTestId('app-page-fallback').query()).toBeNull();
     const el = page.getByTestId('app-page-iframe').element() as HTMLIFrameElement;
     expect(el.getAttribute('data-block-ready')).toBe('true');
-    expect(page.getByTestId('app-page-loading').query()).toBeNull();
+    // The launch veil now CROSS-FADES out over LAUNCH_REVEAL_MS once the block is
+    // ready (the "subtly animated" launch experience) rather than disappearing in
+    // the same commit, so poll for its removal instead of reading it synchronously.
+    // It is still gated structurally — every terminal state unmounts it with the
+    // whole iframe branch — so this can't mask a stuck spinner; the terminal-path
+    // tests above (and PageBlockHostLaunchReveal.browser.test.tsx) pin that.
+    await vi.waitFor(() => {
+      expect(page.getByTestId('app-page-loading').query()).toBeNull();
+    });
   });
 
   test('failure-after-retry: a second fatal error shows the fallback again (no timer leak / stuck state)', async () => {
@@ -715,6 +728,17 @@ describe('PageBlockHost block render/impression (Analytics Phase 2)', () => {
   // Analytics Phase 2 now emits via the /api/track/block-render BEACON
   // (sendBlockRender → fetch), not a tRPC mutation. Spy on global fetch and
   // assert the beacon fires exactly once at BLOCK_READY (and never on re-render).
+  //
+  // 🔴 THIS DESCRIBE IS ALSO THE BATCHING-INDEPENDENCE GUARD for the beacon, and
+  // that is why it must keep rendering the REAL host with REAL hooks (no
+  // `@mantine/hooks` mock in this file). The beacon used to fire from inside the
+  // BLOCK_READY handler behind a flag set by the `setStatus` UPDATER, which only
+  // works while React can evaluate that updater EAGERLY — i.e. while nothing else
+  // is queued on the fiber. `useReducedMotion` (added by the launch-reveal work)
+  // commits its value in a post-mount effect, which is exactly such a pending
+  // update, and it made the impression drop 5/5 runs. The beacon now keys off the
+  // COMMITTED `status`. Mutation-verified: restoring the in-handler `acked`
+  // pattern fails the first test below, deterministically.
   let fetchSpy: ReturnType<typeof vi.spyOn>;
 
   // Only the block-render beacon goes through fetch in this test; resolve OK.
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -144,7 +144,7 @@ const TOKEN_WAIT_TIMEOUT_MS = 15_000;
  * a `next/dynamic` ChatWindow chunk), so importing it here would pull a new
  * animation runtime into the run-page bundle for one cross-fade.
  */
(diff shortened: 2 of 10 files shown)


ZacxDev:
## Corrections to this PR's description (from an adversarial audit)

Posting these as a comment rather than silently editing the body, since a reviewer may already have read the original.

**1. I understated the `IframeHost` sibling.** The body says it "has the identical in-handler pattern … same latent bug". That's true but too soft: in `IframeHost.tsx:841-869` the same `appliedReady` flag also gates **`applyHeight(payload.height)`**, not just the beacon. So under the same batching condition the model-slot block's **height is never applied** — a visible layout bug, not only analytics loss. Still deliberately out of scope here (different surface, and the height semantics deserve their own change), but it should be prioritised accordingly. Note this PR does not make it *more* likely — no new pending state update was added to `IframeHost`.

**2. The `ok`-beacon rate will step up on deploy.** Fixing the dropped impressions means `civitai_app_block_renders_total{result="ok"}` and the `blockRenders` ClickHouse insert step up by however many were being silently lost. Any dashboard or alert baselined on the old rate shifts. Not a regression — it's the bug being fixed — but it should not surprise anyone reading the graphs.

**3. One classification flip.** A `BLOCK_READY` followed by `BLOCK_ERROR{fatal}` **in the same task** now commits only `fatal`, so it emits `status:'error', errorClass:'fatal'` where the old eager-updater path emitted `ok`. More correct, but it is an observable discriminator change.

**4. Mod review previews emit impressions more reliably now.** `ReviewBlockPreviewHost.tsx:218` and `src/pages/apps/dev/[blockId].tsx:230` mount the same host, and `BlockRenderBeaconInput` carries no `reviewMode` discriminator — so mod previews of *pending, unapproved* code now pollute the impression metric more consistently than before. Pre-existing shape; the fix amplifies it. Worth a follow-up (suppress the beacon, or add a discriminator, when `reviewMode` is set) — not done here because it's a metrics/product decision.

## CI status, with provenance

Green: **Typecheck**, **Unit tests**, **ESLint + Prettier**, **event-engine-common pin**.

Two red checks, both proven pre-existing rather than asserted:

- **`preview / smoke-tests`** — 2 failed / 59 passed, failing `preview-apps-external-approve.spec.ts:190` and `preview-generation.spec.ts:48`. PRs #3451, #3452 and #3454 in the same window show **exactly the same two specs at the same lines with identical 2/59 counts**. Separately: no smoke spec references any surface this PR changes, so smoke carries no coverage here — the browser suites and manual verification do. The two specs that *do* exercise `/apps` and `/apps/run` (`preview-apps-marketplace`, `preview-apps-page`) both pass.
- **`preview / deploy`** on the previous tip — every task succeeded (`build-image`, `deploy-preview`, `typecheck`, `unit-tests`, `component-tests`, `smoke-test`, `lighthouse`, `bundle-comment`, `verify-preview`) **except `auth-tests`, which hung ~79 minutes** and was cancelled as collateral when the PipelineRun blew its 1h30m budget: *"TaskRun cancelled as the PipelineRun it belongs to has timed out."* Not an assertion failure, and this diff touches no auth surface. `auth-tests` succeeded on #3451/#3452/#3454/#3442/#3455/#3458, so it isn't a general break either — retriggering to confirm it was transient.

## Perf, measured

- **Bundle: `✅ no per-route First Load JS regression vs main (8ec1700)`.** The `shared 548.7 kB > 470 kB` breach is **byte-identical on #3458 and #3455**, i.e. repo-wide and pre-existing, not from this change. Consistent with the import graph: `PageBlockHost` already pulls `IframeHost`, which already imports `Avatar`/`Text`/`Box`/`Group` from `@mantine/core`, and `@mantine/hooks` has 195 importers in `src/`.
- **Lighthouse gives this PR no signal.** `tests/lighthouse-mint-cookie.cjs:74` measures only `/`, `/models`, `/generate` — `/apps` and `/apps/run/*` are not in the list, and can't be, since LHCI authenticates as `ci-smoke-gold` (a paid member, not a mod) while `/apps` is mod-gated. The reported `CLS 0.229` is for `/` and matches #3458 exactly (#3455: 0.228). **So the CLS claim for the enlarged covers rests on the code and the browser tests, not on Lighthouse** — the aspect box reserves height from the column width before the image loads, and the `onError` placeholder occupies the same box, both pinned by test.

PR #3462, opened 2026-07-29
docs(articles): say why the profile feed disables store filters

### LINKED ISSUE #3461: Profile Articles tab is empty for the owner: the global feed's persisted `followed` filter leaks in
The profile Articles tab merges the **global** article feed's persisted filter store underneath its own filters. A `followed: true` left in that store from `/articles` survives into a user-scoped query, and since nobody follows themselves it excludes every one of the profile owner's own articles. The tab renders "No results found — try adjusting your search or filters", while the filter responsible is not shown in the dropdown and cannot be cleared from that page.

Reported by a user who could see their own articles in the profile's "Most popular articles" section but got an empty Articles tab.

## Reproduce

1. Go to `/articles`, open the filter dropdown, toggle **Followed** on. (This persists to `localStorage` key `article-filters`.)
2. Toggle it off again — or don't, either way; the value is now in the store.
3. Go to your own profile → **Articles**.
4. Empty state, regardless of how many published articles you have.

Console workaround: `localStorage.removeItem('article-filters')`, then reload.

## Cause

1. `articleFilterSchema` carries an optional `followed` boolean — `src/providers/FiltersProvider.tsx:117-122` — persisted under the `article-filters` key (`:236`).
2. `src/pages/user/[username]/articles.tsx:52-55` destructures `followed` **out** of the URL params:
   ```ts
   query: { followed = undefined, ...query },
   ```
   so it is deliberately absent from the `filters` prop the page passes down.
3. `src/components/Article/Infinite/ArticlesInfinite.tsx:22-24` merges the store **underneath** those overrides:
   ```ts
   removeEmpty({ ...articlesFilters, ...filterOverrides })
   ```
   The page overrides `sort` and `period`, so those win — but because it removed `followed` from its overrides, the stored `true` has nothing to override it. `removeEmpty` only drops `null`/`undefined`/empty arrays (`src/utils/object-helpers.ts:3-14`), so a boolean `true` passes straight through.
4. `followed` is accepted by the tRPC input schema — `src/server/schema/article.schema.ts:73`.
5. `src/server/services/article.service.ts:322-336` then applies:
   ```ts
   AND.push(Prisma.sql`a."userId" IN (${Prisma.join(followedUsersIds, ',')})`);
   ```
   The profile owner is not in their own follow list, so every one of their articles is filtered out.

## Why only the Articles tab

Every one of these stores has a `followed` key — model (`FiltersProvider.tsx:50`), image (`:84`), post (`:112`), article (`:121`) — so the trap is armed for all four profile tabs. The other three each avoid it, by three different mechanisms:

| Profile tab | How `followed` is handled | Leaks? |
|---|---|---|
| Models | passes `disableStoreFilters` — `user/[username]/models.tsx:135,161` | no |
| Images / Videos | destructures `followed = undefined` (`UserMediaInfinite.tsx:37`) and **passes it explicitly** (`:161`), so the `undefined` shadows the store. Note `hidden: undefined` at `:155` doing the same job | no |
| Posts | destructures `followed = false` (`user/[username]/posts.tsx:45`) and **passes it** (`:108`) | no |
| **Articles** | destructures `followed = undefined` (`articles.tsx:54`) and **never passes it** | **yes** |

Articles is the only tab that strips `followed` out of its overrides without forwarding it, so nothing shadows the persisted value. That's the whole difference, and it's why the bug looks article-specific rather than like the store-merge issue it actually is.

## Why "Most popular articles" still works

`src/components/Profile/Sections/PopularArticlesSection.tsx:17-24` calls `useQueryArticles` **directly**, bypassing `ArticlesInfinite`, so the store is never merged. Both paths share the same `browsingLevel` and hidden-tag preferences, which is what rules those out as the cause. The Drafts tab uses a different procedure entirely (`getMyDraftArticles`), so it is unaffected.

## Why it's hard to diagnose

- `src/pages/user/[username]/articles.tsx:96-100` passes the stripped `followed` (always `undefined`) into `ArticleFiltersDropdown`, so the dropdown renders **Followed as off** while the store still has it on. The filter is invisible and unclearable from the page it is breaking.
- `ArticlesInfinite` has no `isError` branch (`:37-71`) — it switches on `articles.length` and `isFetching` only. Any query failure renders the same "No results found" empty state, so a server error is indistinguishable from a legitimately empty feed.

## Two related defects, not fixed here

1. **`Prisma.join` on an empty array** — `article.service.ts:335`. A user with `followed: true` who follows nobody produces `IN ()`, which is invalid SQL. That path errors rather than returning zero rows, and defect (2) below disguises it as an empty feed.
2. **No error state in `ArticlesInfinite`** (`:37-71`), as above. Worth a distinct fix since it changes user-visible copy.

Separately, and unrelated to this report: articles with `status = Processing` appear in **neither** profile tab. The published query requires `status = 'Published'` (`article.service.ts:338-351`) and the drafts query matches `status IN (Draft, Unpublished)` (`:1355`). `ArticleUpsertForm` sets `Processing` whenever a save introduces new embedded images that need scanning, so such an article is invisible on its author's profile until the scan finishes.

AKharytonchyk:
> ## Rebased onto current main — the fix already landed, this is now comment-only
>
> `main` moved 1,544 commits since this branch was cut, and **`disableStoreFilters` is already on `ArticlesInfinite` in `src/pages/user/[username]/articles.tsx`**. Someone fixed this independently.
>
> Worth flagging how that surfaced: the rebase was **textually clean**. `main` put the prop above `filters={{…}}` and this branch put it below, so git merged both and produced a duplicate JSX attribute. Nothing complained until a full-repo typecheck — `TS17001: JSX elements cannot have multiple attributes with the same name`. A clean rebase and a green PR are not the same thing.
>
> So this PR is reduced to the part `main` doesn't have: a comment recording *why* the prop is there. It passes the keep test in `CLAUDE.md` — the specific future edit that goes wrong without it is someone deleting `disableStoreFilters` as redundant, since it sits right next to an explicit `filters` object and reads like a no-op. The failure it causes is silent and total: profile filters are URL-driven, the global feed's persisted store leaks `followed: true` into a user-scoped query, and nobody follows themselves, so every one of the owner's articles disappears.
>
> Close it if a comment isn't worth a review cycle. Typecheck is green (0 errors).

---

Fixes #3461 — the profile Articles tab rendered "No results found" for the profile owner while the tab bar showed `Articles 80` and the same articles appeared in "Most popular articles" on the same page.

## What happened

The tab merged the **global** article feed's persisted filter store underneath its own filters. A `followed: true` left in that store from `/articles` survived into a user-scoped query, and since nobody follows themselves it excluded every one of the owner's articles.

The chain, all five links verified:

1. `articleFilterSchema` carries an optional `followed` boolean — `FiltersProvider.tsx:117-122` — persisted under the `article-filters` key (`:236`).
2. `user/[username]/articles.tsx:52-55` destructures `followed` **out** of the URL params, so it is deliberately absent from the `filters` prop it passes down.
3. `ArticlesInfinite.tsx:22-24` merges the store **underneath** those overrides — `removeEmpty({ ...articlesFilters, ...filterOverrides })`. `sort` and `period` are overridden and win; `followed` isn't in the overrides at all, so the stored `true` has nothing to override it. `removeEmpty` only drops nil and empty arrays (`object-helpers.ts:3-14`), so a boolean survives.
4. The tRPC input schema accepts `followed` — `article.schema.ts:73`.
5. `article.service.ts:322-336` applies `AND a."userId" IN (followedUserIds)`. The owner is not in their own follow list.

## Why this only hit Articles

All four feed stores carry a `followed` key — model (`FiltersProvider.tsx:50`), image (`:84`), post (`:112`), article (`:121`) — so every profile tab had the same trap armed. The other three defend against it, each differently:

| Profile tab | Handling | Leaks? |
|---|---|---|
| Models | **both** guards — `disableStoreFilters` (`models.tsx:135,161`) *and* forwards `followed ?? false` (`:55,123,148`) | no |
| Images / Videos | destructures `followed = undefined` (`UserMediaInfinite.tsx:37`) and passes it explicitly (`:161`) so the `undefined` shadows the store — same as its `hidden: undefined` at `:155` | no |
| Posts | destructures `followed = false` (`posts.tsx:45`) and passes it (`:108`) | no |
| **Articles** | destructures `followed = undefined` (`articles.tsx:54`) and never passes it | **yes** |

Articles is the only tab that strips the key without forwarding it. Worth noting for review: three tabs, three different guards, none of them obvious — which is why this reads as an article bug rather than a store-merge design problem.

## Why this fix

`disableStoreFilters` rather than `followed: !!followed`, because the page is already URL-driven and shouldn't be reading that store at all: its `ArticleFiltersDropdown` is passed an `onChange` (`:96-100`), which routes writes to `replace()` — the URL — instead of `setFilters` (`ArticleFiltersDropdown.tsx:26-31`). The store was only ever an accidental input here.

It also makes the tab consistent with its sibling: **`user/[username]/models.tsx:135,161` already passes `disableStoreFilters`**, and `Collection.tsx:464` already uses the flag on `ArticlesInfinite` itself. Articles was the outlier.

Patching `followed` alone would work but leaves the same trap armed for every other key the store gains later.

## No behaviour lost

Everything the store contributed here is either overridden by the page or irrelevant:

| store key | default | after this change |
|---|---|---|
| `sort` | `MostBookmarks` | page supplies it (URL, default `Newest`) |
| `period` | `Month` | page supplies it (URL, default `AllTime`) |
| `followed` | — | **gone, which is the fix** |
| `periodMode` | — | only read when `period !== AllTime` (`article.service.ts:341`); the page's default is `AllTime` |

The dropdown is unaffected: it renders from `query || filters` (`ArticleFiltersDropdown.tsx:24`) and the page passes `query`, so it was already displaying URL state, not store state.

## Not fixed here

Adjacent defects found while verifying this one, deliberately kept out. Full write-up in #3461:

- **`Prisma.join([])` throws** at `article.service.ts:335`, so `followed` with an empty follow list 500s instead of returning an empty feed. Articles is the only service missing that guard. Same code path as this bug, but a different failure mode and a server file — separate PR.
- `ArticlesInfinite` has no `isError` branch (`:37-71`), so any query failure renders this same empty state. That's why this presented as a filter problem rather than an error. Changes user-visible copy, so it wants its own PR.
- `article.utils.ts:33` uses `z.coerce.boolean()` for `followed`, so `?followed=false` parses as `true`. Inert here post-fix, live on `/articles`.
- `useZodRouteParams.ts:11-12` returns `{}` on any parse failure, dropping `username` with it — so one bad param renders the *global* feed inside a profile's Articles tab.

## What I'm not certain about

The mechanism above is verified end to end in code, but **I never confirmed against the wire that the reporting user's `localStorage['article-filters']` actually held `followed: true`.** The diagnosis is an elimination argument over the four input differences between this query and the working "Most popular" one, and it is consistent with all six observations — but a 200-with-zero-rows and a 500 render identically here (see the missing `isError` branch above), so I can't fully exclude that they hit the `Prisma.join` 500 instead, which the separate PR fixes. Both fixes are independently correct; I'd just rather flag that than imply I watched it happen.

## Verification

One file, one prop. Typecheck **0 errors** repo-wide; unit suite **680 files / 9055 tests passed, 0 failed** (1 skipped); Prettier and ESLint clean on the changed file.

Component tests weren't run: nothing under `src/**/*.browser.test.tsx` references `ArticlesInfinite`, `disableStoreFilters`, or this page, so they can't cover it either way.

The filter merge was confirmed against the repo's real `removeEmpty` rather than a reimplementation: with the store merged the object sent to tRPC is `{period: AllTime, sort: Newest, followed: true, includeDrafts: false, pending: true}`; with `disableStoreFilters` it is the same minus `followed`. `period` and `sort` are identical either way, which is the evidence that this removes exactly one input.

Also checked, since `disableStoreFilters` skips `removeEmpty`: zod `.partial()` omits absent keys rather than emitting `undefined`, so no `undefined` reaches the input here; and React Query's `hashKey` drops undefined-valued keys, so there is no cache-key split even where they do occur. `Collection.tsx:433-445,464` already ships `followed/favorites/hidden: undefined` through this same flag in production.

## One thing I deliberately did not add

A regression test. Nothing under `src/**/*.test.*` currently references `ArticlesInfinite`, `disableStoreFilters`, or this page, so nothing pins this fix. Doing it properly needs either extracting the filter-merge into a testable unit or new tRPC-input mocking for a component test — both larger than the fix itself. Happy to follow up if you'd like it, but I'd rather not bolt on fragile mocking to claim coverage.

I also considered forwarding `followed: false` explicitly on top of the flag, to match Models' belt-and-braces. I left it out: `article.service.ts:322` (`!!sessionUser && followed`) and `:202` (`!followed`) treat `undefined` and `false` identically, so it changes nothing today and would imply the flag alone is insufficient. A test is the right way to pin that, not a redundant prop.

### DIFF
--- a/src/pages/user/[username]/articles.tsx
+++ b/src/pages/user/[username]/articles.tsx
@@ -110,6 +110,11 @@ function UserArticlesPage() {
                   includeDrafts: !!currentUser?.isModerator,
                   pending: true,
                 }}
+                // Filters here are URL-driven, so merging the global feed's
+                // persisted store leaked its `followed: true` into a user-scoped
+                // query — and nobody follows themselves, which hid every one of
+                // the owner's own articles.
+                disableStoreFilters
                 showEmptyCta={selfView}
               />
             ) : (
--- a/src/pages/user/[username]/articles.tsx
+++ b/src/pages/user/[username]/articles.tsx
@@ -103,6 +103,10 @@ function UserArticlesPage() {
             </Group>
             {viewingPublished ? (
               <ArticlesInfinite
+                // Filters here are URL-driven. Merging the global feed's persisted
+                // store leaked its `followed: true` into this user-scoped query, and
+                // nobody follows themselves, so every one of the owner's own articles
+                // was hidden. Do not drop this as redundant.
                 disableStoreFilters
                 filters={{
                   ...query,


AKharytonchyk:
Closing as low value.

The functional fix — `disableStoreFilters` on the profile feed — already landed on `main` independently, so all this PR still carried was a four-line comment explaining why the prop is there. That is real knowledge (delete the prop and every one of the owner's own articles silently disappears, because profile filters are URL-driven and nobody follows themselves), but it is not worth a maintainer review cycle on its own.

Leaving it here rather than in the code: **if you are looking at `src/pages/user/[username]/articles.tsx` and `disableStoreFilters` looks redundant next to the explicit `filters` object, it is not.** Removing it reintroduces the bug this PR was originally opened for.

Happy to fold the comment into any future PR that touches that file.

PR #3464, opened 2026-07-30
fix(apps): make the model-slot BLOCK_READY transition batching-safe

ZacxDev:
## The bug

`IframeHost`'s `BLOCK_READY` handler read the `loading → ready` transition out of the `setStatus` **updater**:

```js
let appliedReady = false;
setStatus((current) => {
  if (current === 'loading') { appliedReady = true; return 'ready'; }
  return current;
});
if (appliedReady) {
  controllerRef.current?.notifyReady();
  applyHeight(payload.height);
  if (!blockRenderEmittedRef.current) { blockRenderEmittedRef.current = true; sendBlockRender({...}); }
}
```

The in-code comment claimed this could read the transition out of the updater. It cannot. React only evaluates an updater eagerly as a **bail-out optimisation**, and only while the fiber has no other pending update. As soon as any unrelated state update is already queued on this component when `BLOCK_READY` lands, the eager path is skipped, the updater runs later during render, `appliedReady` is still `false` at the `if`, and the **entire branch is dropped**.

## Consequence set (all three side effects, not just the height)

| # | Side effect skipped | Real-world consequence |
|---|---|---|
| 1 | `applyHeight(payload.height)` | The iframe stays pinned at manifest `minHeight` — clipped block content, or a dead gap under a short block. This is the reported symptom. |
| 2 | `controllerRef.current?.notifyReady()` | The init controller is not told the block acked, so it keeps re-posting `BLOCK_INIT` on its ~400 ms loop. **Bounded, and it does NOT cause a false timeout** — see below. |
| 3 | `sendBlockRender({...})` | The render/impression beacon is silently dropped → `civitai_app_block_renders_total{result="ok"}` undercounts and ClickHouse `blockRenders` misses the row. |

### #2 answered from the code: can a ready block still flip to `timeout`?

**No.** Three independent guards each prevent it, and I mutation-tested the layering:

1. `notifyReady()` → `stop()` clears the retry interval **and** the readiness timeout (the direct path — the one that was being skipped).
2. **`status` is in the init effect's dependency array** (`IframeHost.tsx:818`), so committing `'ready'` re-runs that effect; its cleanup calls `controller.dispose()` and `shouldStartInit({status:'ready'})` then refuses to create a new one. **This fires even when (1) is skipped**, which is what bounds the blast radius of the bug: the extra `BLOCK_INIT` re-posts only span the ack → commit window (realistically 0–1 posts), not the full 10 s.
3. `onReadyTimeout` is itself `current === 'loading'`-guarded, so even a surviving timer is a no-op once ready.

Mutation-verified: breaking (1)+(3) together (**M10**) still passes; breaking all three (**M11**) fails. So the `notifyReady` miss is **wasted work + a redundant armed timer, not a user-visible "Block took too long to load"** — materially *less* bad than the height bug, contrary to the hypothesis in the brief.

## The fix

Reuses PR #3457's approach on the sibling `PageBlockHost` so the two hosts don't diverge into two solutions for one problem:

- plain guarded `setStatus` transition — no side effect read out of the updater;
- `notifyReady()` called **unconditionally** (documented-idempotent → `stop()`);
- the acked height is stashed in `pendingReadyHeightRef` and applied — together with the `ok` beacon — from a **new effect keyed on the COMMITTED `status`**, mirroring the render-FAILURE effect right below it.

`PageBlockHost` has no `applyHeight` (it's full-bleed), so the height ref is the one addition beyond #3457's shape.

### Invariants preserved

- **H-11** is now **structural** rather than a timing observation: the commit effect early-returns unless `status === 'ready'`, and `status` can only reach `'ready'` from `'loading'`. A late ack after `timeout` / `fatal` / `no_token` still writes the ref, but it is never read.
  - ⚠️ One deliberate, narrow deviation from H-11's literal wording ("must not … notify ready"): `notifyReady()` **is** called after a terminal state. It is provably inert there — after `timeout` the controller already stopped itself; after `no_token` it was never created (`shouldStartInit` requires a token); after `fatal` it cancels timers we *want* cancelled and cannot revive `status` (the only writer is the guarded updater, and `onReadyTimeout` is status-guarded). The **observable** content of the invariant (no height, no `ok` beacon, no status change) is enforced and regression-tested per terminal state. Making it unconditional is what removes the last dependence on observing the updater — same call #3457 made.
- **Emit-once**: `ok`/`error` still share `blockRenderEmittedRef`, so they stay mutually exclusive per mount. A new `readyTransitionAppliedRef` keeps the commit effect to one run per mount.
- **Untrusted payload**: the `{height?: unknown}` shape guard is unchanged, and `applyHeight` still drops non-finite/≤0 and clamps to `minHeight`/`maxHeight`/`HARD_HEIGHT_CEILING`.
- **FRAME-1**: untouched. Note `IframeHost` satisfies it by **collapsing to `null`** on every terminal state (`hostRenderDecision`) — nothing rendered ⇒ nothing to masquerade as — which differs from `PageBlockHost`'s framed-fallback shape.

## Declared behaviour changes

Beyond the fix itself, three observable changes — all surfaced or sharpened by the adversarial audit, all now pinned by tests:

1. **Duplicate acks in one batch: the last ack that *carries a height* wins.** Previously the eager path applied the first ack's height and dropped later ones. A later ack with no usable height no longer erases an earlier one (that direction would have been a regression — fixed in the second commit).
2. **`BLOCK_READY` then `BLOCK_ERROR{fatal}` in the same batch now reports `result=error`, not `result=ok`.** The old synchronous handler had already emitted `ok` before the fatal was processed, and the shared emit-once ref then suppressed the `error` beacon. Now both updaters run in one render, committed `status` is `fatal`, the ready effect early-returns, and the mount is honestly reported as a failed render. This moves the render-failure ratio in the **opposite** direction from the beacon fix below — i.e. *more* sensitive, not less.
3. **Ack-immediately-followed-by-unmount now emits no `ok` beacon.** The beacon moved from the synchronous handler to a passive effect, so if the host unmounts between the ack and the effect flush the update is discarded and the beacon is never issued — meaning `sendBlockRender`'s `keepalive: true` (which exists precisely for the navigate-away-on-ready case) no longer covers that window. Accepted: emitting synchronously is what the bug was.

**Scope correction on "H-11 is structural":** the claim holds for the `BLOCK_READY` path, which is what this PR touches. Separately and unchanged by this PR, `RESIZE_IFRAME` / `REQUEST_SIGN_IN` / `OPEN_BUZZ_PURCHASE` capture `status` in a closure that re-subscribes only on the effect flush, so each has a pre-existing stale-closure window after a terminal commit. For `RESIZE_IFRAME` it is unobservable (the host renders `null`, so nothing consumes `iframeHeight`).

## Test coverage

New `IframeHostReadyTransition.browser.test.tsx` — **25 cases**, extending (not replacing) the existing suite. All **223** `src/components/AppBlocks/**` component tests pass.

🔴 **Why the existing suite could not catch this**: `IframeHost.browser.test.tsx` stubs every tRPC hook as a constant-returning function, which **removes the very batching condition the bug needs** — the fiber is quiescent when `BLOCK_READY` arrives, React takes the eager path, and the buggy code passes. (This is the trap the sibling PR hit.) The new file therefore mocks `useBrowsingLevelDebounced` — called directly by `IframeHost`, so its hooks live on the **host fiber** — with a real `useState` and exposes the setter, so a test can queue a genuine React update and then deliver `BLOCK_READY` in the **same synchronous task**. The mock *adds* the batching condition rather than papering over it.

Covered: batched **and** un-batched ready paths · `notifyReady` observed · exactly one `ok` beacon with the model-slot identifiers · duplicate acks in one batch · a late ack after the ready commit not clobbering a negotiated `RESIZE_IFRAME` height · a manifest height change not rewinding it · `ok`/`error` mutual exclusion · 10 malformed/absent height payloads · min-floor and max-clamp · H-11 for all three terminal states (including the *deliverable* batched `BLOCK_ERROR{fatal}` + `BLOCK_READY` case) · a ready block never going terminal.

## Mutation table

Every guard was broken, the *specific* failing test recorded, then reverted.

| # | Mutation | Result |
|---|---|---|
| **M1** | Revert wholesale to the original `appliedReady` code | ✅ **FAIL** — batched-ack test: `expected 200 to be 640` (the exact reported symptom) |
| **M2** | Drop `applyHeight(pendingReadyHeightRef.current)` | ✅ **FAIL** — batched-ack test: `expected 200 to be 640` |
| **M3** | Remove the `status !== 'ready'` guard (H-11's structural gate) | ✅ **FAIL** — H-11 `fatal` test *and* H-11 `timeout` test: `expected [] to have a length of 1` (the `ok` beacon claimed the shared ref, suppressing the `error` beacon) |
| **M4** | Remove the `readyTransitionAppliedRef` run-once guard | ✅ **FAIL** — manifest-height-change test: `expected 700, received 260` (rewound to the stale handshake height) |
| **M5** | Stop claiming the shared `blockRenderEmittedRef` | ✅ **FAIL** — mutual-exclusion test: 2 beacons instead of 1 |
| **M6** | Remove `controllerRef.current?.notifyReady()` | ✅ **FAIL** — batched-ack test: `expected "notifyReady" to be called 1 times, but got 0` |
| **M7** | Remove the untrusted-payload shape guard | ✅ **FAIL** — null-payload test: `TypeError: Cannot read properties of null (reading 'height')` |
| **M9** | Drop only `sendBlockRender(...)` | ✅ **FAIL** — batched-ack test: `expected [] to have a length of 1` |
| **M10** | No `notifyReady` **+** unguarded `onReadyTimeout` | ⚪ **PASSES** — correctly: the init effect's `status`-keyed cleanup is a third, independent guard |
| **M11** | M10 **+** drop `status` from the init effect deps (all three guards) | ✅ **FAIL** — never-goes-terminal test: `expected null not to be null` |
| **M8** | Stash the height from *inside* the `setStatus` updater | ⚪ **PASSES** — **equivalent mutant**, not a coverage gap: the updater runs during render, before the commit effect reads the ref, so behaviour is identical |

One assertion was found **vacuous during this process and replaced**: counting `BLOCK_INIT` postMessages to observe `notifyReady` captured **0** messages (the same-origin iframe replaces its `contentWindow` on load, so a test-side listener never sees the host's posts) — it could never have failed. It now spies `IframeInitController.prototype.notifyReady` (call-through), which M6 does flip. Separately, the late-ack `dispatchEvent` in the `timeout` / `no_token` tests asserts nothing on its own (post-collapse no message can be delivered) — those tests are killed by **M3** via their terminal-outcome assertions instead, and now say so.

## Real-world exposure

**No user-visible impact expected today** — this is a latent correctness bug, not a live one.

- Model-slot apps are outside the launch allowlist (`LAUNCH_SLOT_IDS = ['app.page']`, `src/shared/constants/slot-registry.ts:250`) and the whole surface sits behind the mod-only `appBlocks` Flipt flag.
- `IframeHost` has exactly **one** production mount chain — `ModelVersionDetails` → `BlockSlot` → `BlockSlotClient` → `BlockHost` → `IframeHost`. Verified by grep; nothing else imports it (`PageBlockHost` takes only `AppBlockChrome` from the module, and the mod-review sandbox mounts `PageBlockHost`).
- Every `listForModel` resolve branch gates `AND ab.status = 'approved'` (`src/server/services/block-registry.service.ts:792`, `:853`, `:931`, `:1005`) plus the deploy gate, so a non-approved app structurally cannot mount the host. (Caveat found in review: the 60s anon cache path at `:718-735` filters only on kill-list + host maturity, so a status change is visible for up to a minute on that branch — pre-existing, self-healing, unrelated to this PR.)

I checked the current install census against the database separately; the specifics live in the private infra repo rather than here. Net: this is **not urgent**, but it stops being latent the moment a model-slot app is approved — at which point the height half is immediately visible (clipped content / dead gap) and the beacon half is silent.

## Metrics / alerting note

Fixing the dropped beacon will **step `civitai_app_block_renders_total{result="ok"}` up on deploy** for model-slot blocks. Not a regression — the old rate was an undercount. Two consequences to expect:

- graphs baselined on the old rate shift;
- the infra-side render-failure alert is a **ratio** of `error` to total, so a larger `ok` denominator makes it **less sensitive** (a fixed number of failures dilutes toward its threshold).

Given the exposure above the step is effectively **0 → 0** right now, so the practical impact lands whenever model-slot blocks actually serve.

## Verification

- `vitest --project component src/components/AppBlocks/` → **223/223 pass** (20 files), including the 5 pre-existing `IframeHost.browser.test.tsx` cases. Also re-ran the batching / reverse-order / height-loss cases 16× to check for scheduling flakiness — all deterministic.
- `tsc --noEmit -p tsconfig.json` → **28 errors, byte-identical to the `main` baseline** (uninitialised `event-engine-common` submodule + stale Prisma client + `@civitai/buzz` workspace resolution — all local-env artifacts in files this PR doesn't touch). Zero added.
- `eslint` on the changed files → **0 errors**, 1 warning that is pre-existing on `main`.
- `prettier --check` clean on the new file; the two pre-existing files are already prettier-dirty on `main`, so they were left unformatted to avoid unrelated churn.

`preview / deploy` (`next build`) is the authoritative typecheck.

### DIFF
--- a/src/components/AppBlocks/IframeHost.browser.test.tsx
+++ b/src/components/AppBlocks/IframeHost.browser.test.tsx
@@ -264,8 +264,10 @@ describe('IframeHost block render/impression (Analytics Phase 2, model.sidebar_t
     await vi.waitFor(() => expect(beaconCalls()).toHaveLength(1));
 
     // A second/third BLOCK_READY (block re-ack, or a re-render re-running
-    // listeners) finds status === 'ready', so `appliedReady` stays false AND
-    // the emit-once ref is already set → no re-emit.
+    // listeners) can't re-emit: the beacon now fires from the COMMITTED
+    // loading→ready effect, which is already past its once-per-mount guard, and
+    // the shared emit-once ref is set. (See IframeHostReadyTransition.browser
+    // .test.tsx for the batching-immunity + H-11 coverage of that effect.)
     postFromBlock('BLOCK_READY', {});
     postFromBlock('BLOCK_READY', {});
     await new Promise((r) => setTimeout(r, 150));
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -853,11 +853,17 @@ export function IframeHost({
       const payload =
         raw && typeof raw === 'object' && 'height' in raw ? (raw as { height?: unknown }) : {};
       // Record the offered height for the ready-transition effect below to apply
-      // once React has COMMITTED 'ready'. Deliberately ungated: H-11 is enforced
-      // by that effect's `status === 'ready'` guard, not here (see the ref's
-      // comment). When several acks land in one batch the LAST one wins, which is
-      // the block's most recent statement of its own handshake height.
-      pendingReadyHeightRef.current = payload.height;
+      // once React has COMMITTED 'ready'. NOT gated on the current status: H-11 is
+      // enforced by that effect's `status === 'ready'` guard, not here (see the
+      // ref's comment).
+      //
+      // When several acks land in one batch the LAST one that CARRIES a height
+      // wins — the block's most recent statement of its own handshake height. The
+      // `!== undefined` check matters: without it a `BLOCK_READY {height: 640}`
+      // followed in the same batch by a bare `BLOCK_READY {}` (or one whose height
+      // fails the shape guard) would reset the ref to `undefined` and LOSE a height
+      // the old code applied, turning this fix into a regression in that direction.
+      if (payload.height !== undefined) pendingReadyHeightRef.current = payload.height;
       setStatus((current) => (current === 'loading' ? 'ready' : current));
       // Block acked — stop re-posting BLOCK_INIT and cancel the readiness
       // timeout. Called UNCONDITIONALLY, not behind a "did this ack win the
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -529,15 +529,32 @@ export function IframeHost({
   // the controller started — the next tick picks up the new payload) without
   // re-creating the controller and resetting its timers.
   const buildInitPayloadRef = useRef<() => BlockInitPayload>();
-  // Analytics Phase 2: emit-once guard for the block-render beacon (see the
-  // BLOCK_READY effect). The 'loading' → 'ready' transition is the primary
-  // dedup; this ref makes the per-mount emit deterministic even if duplicate
-  // BLOCK_READY acks land before React commits the 'ready' state.
+  // Analytics Phase 2: emit-once guard for the block-render beacon, SHARED with
+  // the render-FAILURE beacon below so `ok` and `error` are mutually exclusive
+  // per mount. The committed-`status` effects are the primary dedup; this ref
+  // makes the per-mount emit deterministic even if duplicate BLOCK_READY acks
+  // land before React commits the 'ready' state.
   const blockRenderEmittedRef = useRef<boolean>(false);
+  // The height the block offered in its BLOCK_READY ack, stashed for the
+  // ready-transition effect to apply once React has COMMITTED 'ready'.
+  //
+  // 🔴 A ref, not state, and read ONLY from the `status === 'ready'` effect. That
+  // makes H-11 STRUCTURAL rather than a timing observation: a late ack that
+  // arrives after the host already landed on 'timeout' / 'fatal' / 'no_token'
+  // still writes here, but `status` can never become 'ready' again from a
+  // terminal state (BLOCK_READY only transitions FROM 'loading'), so the stashed
+  // value is simply never read and the height is never applied.
+  const pendingReadyHeightRef = useRef<unknown>(undefined);
+  // Run the loading→ready side effects (height + `ok` beacon) exactly once per
+  // mount. Without this, an identity change of the `applyHeight` callback (its
+  // deps are manifest min/max height) would re-run the effect while `status` is
+  // still 'ready' and re-apply the handshake height, clobbering a height the
+  // block had since negotiated via RESIZE_IFRAME.
+  const readyTransitionAppliedRef = useRef<boolean>(false);
 
   // App Blocks Analytics Phase 2 — fire-and-forget block render/impression,
-  // emitted exactly once per mount at the BLOCK_READY transition (see the
-  // BLOCK_READY effect below) via the lightweight /api/track/block-render beacon
+  // emitted exactly once per mount on the COMMITTED loading→ready transition
+  // (see the ready-transition effect below) via the /api/track/block-render beacon
   // (NOT a tRPC mutation — this fires per model-page-with-a-block view, so at GA
   // it must skip the full tRPC middleware chain; mirrors the #2680 addView ->
   // beacon move). The client passes only the three identifiers; `isAnon`/`userId`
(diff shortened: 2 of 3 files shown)


ZacxDev:
### Corrections to the original PR description (adversarial audit round 1)

The body has been updated. Flagging the changes explicitly rather than editing silently, in case anyone read the first version:

1. **Two of my H-11 tests were vacuous.** The `timeout` and `no_token` cases dispatched a "late `BLOCK_READY`" *after* the host had already collapsed to `null`. `usePostMessage` pins `event.source === iframeRef.current?.contentWindow`, and that ref is null once the iframe unmounts — so the message was dropped before the handler ran, and **both tests would have passed with the status guard entirely removed**. They now assert what they actually prove (terminal outcome + that the ack is undeliverable), and the describe block explains that the guard is exercised by the batched `BLOCK_ERROR{fatal}` case instead. Fixed in b98df4b.

2. **"Last ack wins" was a regression in one direction.** `BLOCK_READY {height:640}` followed in the same batch by a bare `BLOCK_READY {}` reset the stashed height to `undefined` and left the frame at `minHeight` — a case the old eager path got *right*. Now only acks that carry a height update the ref. New test, mutation-verified. Fixed in b98df4b.

3. **Two behaviour changes were undeclared** and are now in a "Declared behaviour changes" section: `BLOCK_READY` → `BLOCK_ERROR{fatal}` in one batch now reports `result=error` instead of `result=ok` (which moves the render-failure ratio in the *opposite* direction from the beacon fix the body discusses), and ack-then-immediate-unmount now emits no `ok` beacon because the passive effect loses `keepalive`'s protection for that window.

4. **The "H-11 is structural" claim is now scoped** to the `BLOCK_READY` path. `RESIZE_IFRAME` / `REQUEST_SIGN_IN` / `OPEN_BUZZ_PURCHASE` each have a pre-existing stale-closure window after a terminal commit — unchanged by this PR, and unobservable for `RESIZE_IFRAME` since the collapsed host has no consumer for `iframeHeight`.

5. **The exposure section previously included a production install census.** This repo is public, so those specifics were removed; the code-level reasoning (single mount chain, `status = 'approved'` on every resolve branch) is retained, and the data-state check lives in the private infra repo. The audit also noted a pre-existing 60s anon-cache path in `listForModel` that doesn't re-check app status — unrelated to this PR, now mentioned for completeness.

Audit verdict was "safe to merge"; items 1–3 were its should-fix list and are addressed.

ZacxDev:
### Round-2 delta re-audit: I over-corrected in round 1

The re-audit **ran the mutation** rather than reasoning about it, and found my round-1 correction was itself wrong. Correcting publicly for the same reason as before.

**What I got wrong:** I described the `timeout` / `no_token` late-ack tests as "vacuous — would pass with the guard entirely removed". That's false. Removing the ready effect's `status !== 'ready'` guard fails **10 of the 25 tests** in the file, **including both of those two** — the ready effect then fires at mount, claims the shared `blockRenderEmittedRef`, and their `expect(errorBeacons()).toHaveLength(1)` assertion (an unchanged line, present in the original draft) fails. My own M3 mutation run had already shown exactly that; I wrote the comment from the round-1 finding without reconciling it against my own evidence, and then enshrined the overstatement as a 🔴-flagged instruction to future test authors.

**The accurate split**, now in the code:
- the late `dispatchEvent` in those two tests is a **no-op that asserts nothing on its own** — nothing there distinguishes "dropped at the transport" from "delivered and correctly ignored". It documents the scenario.
- but the tests **do** have teeth, from their terminal-outcome assertions, which that mutation kills.
- the guard is exercised **directly** — ack genuinely delivered, status genuinely terminal — only by the two batched `BLOCK_ERROR{fatal}` cases.

**Also fixed** (27b3cbd, no production code):
- Dropped `source: preCollapseWindow`. It bought no assertion strength (post-collapse the pin comparand is `undefined`, and `MessageEvent.source` is never `undefined`, so no source value can pass) and added a dependency on the engine accepting a detached `WindowProxy`.
- Corrected a contradictory comment: the iframe's `WindowProxy` **identity is preserved** across the same-origin document load; what's swapped is the inner global — which is still why a listener registered pre-load never sees the host's posts, so the original "counting BLOCK_INIT posts was vacuous" conclusion stands, just for the right reason.
- PR body counts were stale (23 → **25** cases, 221 → **223** passing).

Both audit rounds landed on **safe to merge**; every 🟡 from both is now addressed, and the remaining 🟢 items are noted in-code. Not merging — over to a human.

PR #3465, opened 2026-07-30
fix(apps): deterministic per-app placeholder for listings missing icon/cover

ZacxDev:
## Problem

A listing with no icon and/or no cover already fell back rather than rendering a broken `` — but **every** such listing got the same flat dark gradient + a generic icon, so they were indistinguishable and read as *broken* rather than *no art yet*.

Two live examples right now, both grandfathered in below the #3392 icon+cover floor:

| slug | icon | cover | category |
|---|---|---|---|
| `buzz` | – | – | analytics |
| `df-qwen-canvas` | – | – | **null** → generic `IconApps` |

Meanwhile the **server** already had a much better deterministic design for the *generated* placeholder assets: a slug-seeded hue + the app's initial (`buildPlaceholderIconSvg`), and a seeded gradient + category glyph (`buildPlaceholderCoverSvg`). The client couldn't reuse any of it — those pure helpers lived inside `app-listing-assets.service.ts`, which imports `dbRead`/`dbWrite`/`sharp`. So the two designs were hand-rolled separately and drifted.

## Change

Extract the primitives into **`~/shared/constants/app-listing-placeholder.constants`** — pure, no React, no Prisma, no `sharp` — and have both sides derive from it:

- **`AppListingCard` + `AppListingDetailBody`** — cover placeholder uses the seeded per-app gradient; a missing icon renders the seeded monogram via `appInitial()` (which correctly handles a whitespace-only or non-alphanumeric name) instead of Mantine's flat grey `name.charAt(0)`.
- **The SVG builders** consume the same shared seed + stops, so a generated asset and the live fallback are provably one colour identity.

## Scope / safety

- **No behaviour change** for a listing that has both assets.
- **No change to the mandatory-asset floor.** These are render-time fallbacks that never become stored assets, so they can't satisfy — or be mistaken for — a real uploaded asset. `checkListingMeetsFloor` / the owner "Incomplete" badge / the my-submissions warning are all untouched, and `buzz` + `df-qwen-canvas` remain below the floor.
- The server service re-exports `seededHue` / `appInitial`, so its public API and existing tests are unchanged.

## Tests

- 19 new unit tests on the shared module (determinism, per-app variation, the null-category path, exact stop values) — in the **blocking** `unit` project.
- A **cross-surface drift guard** in the service suite asserting the generated SVG stops equal the client gradient stops.
- Both guards **mutation-verified**: re-inlining a literal stop in the SVG builder, and nudging a shared stop's lightness, each fail the intended test (then reverted).

```
unit  src/shared/constants/__tests__/app-listing-placeholder.constants.test.ts  19 passed
unit  src/server/services/blocks/__tests__/app-listing-assets.service.test.ts   78 passed
```

`tsc --noEmit` reports **zero** errors in the changed files (the 32 remaining are pre-existing environmental — missing `event-engine-common` submodule, stale Prisma client, `@civitai/buzz` workspace). The authoritative typecheck is the pr-preview build.

### DIFF
--- a/src/components/Apps/AppListingCard.tsx
+++ b/src/components/Apps/AppListingCard.tsx
@@ -56,10 +56,13 @@ function categoryIcon(category: string): Icon {
 
 /**
  * Card cover — the listing's cover image (`coverUrl`, already a CDN URL, with the
- * first screenshot as a server-side fallback). When absent, a tasteful
- * category-glyph placeholder over a neutral gradient (mirrors AppBlockCard) so a
- * card is never a broken/empty `<img>`. Decorative (aria-hidden) — the placeholder
- * carries no info the title/category chip don't.
+ * first screenshot as a server-side fallback). When absent, a category-glyph
+ * placeholder over the listing's DETERMINISTIC PER-APP seeded gradient (shared
+ * with the server-generated cover SVG — see
+ * `~/shared/constants/app-listing-placeholder.constants`), so a card is never a
+ * broken/empty `<img>` and two coverless listings never look identical.
+ * Decorative (aria-hidden) — the placeholder carries no info the title/category
+ * chip don't.
  */
 function ListingCover({
   coverUrl,
@@ -100,7 +103,7 @@ function ListingCover({
         aria-hidden
         h={140}
         className="flex items-center justify-center"
-        data-listing-cover-placeholder
+        data-listing-cover-placeholder=""
         style={{ background: listingPlaceholderGradient({ slug, category, surface: 'cover' }) }}
       >
         <PlaceholderIcon size={44} className="opacity-60" />
--- a/src/components/Apps/AppListingCard.tsx
+++ b/src/components/Apps/AppListingCard.tsx
@@ -18,6 +18,10 @@ import {
 import { TruncatedText } from '~/components/Apps/AppListingTruncate';
 import { useCurrentUser } from '~/hooks/useCurrentUser';
 import { isMarketplaceCategory } from '~/server/services/blocks/marketplace-categories.constants';
+import {
+  appInitial,
+  listingPlaceholderGradient,
+} from '~/shared/constants/app-listing-placeholder.constants';
 import type { ListingCard } from '~/server/schema/blocks/app-listing-read.schema';
 
 /**
@@ -61,10 +65,12 @@ function ListingCover({
   coverUrl,
   category,
   name,
+  slug,
 }: {
   coverUrl: string | null;
   category: string | null;
   name: string;
+  slug: string;
 }) {
   // A non-null coverUrl can still 404 (the server derives it from a first-
   // screenshot fallback, whose Image can dangle) — fall back to the category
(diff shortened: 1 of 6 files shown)


ZacxDev:
## Audit follow-up — PR body correction + 2 🟡 fixes (`c6d4077`)

An adversarial audit ran against `01ecb67`. **No 🔴.** It empirically confirmed the three things that actually mattered:

- **Generated SVGs are byte-identical** across the refactor — verified over 10 inputs × 4 arg shapes (40 comparisons), including `category: null` / `undefined`, empty slug/name, emoji and CJK names. So `backfillListingAssets` produces the same bytes as before and there is **no stored-vs-new asset divergence**.
- The **asset floor is untouched** (no write path; `checkListingMeetsFloor` / `computeListingProblems` never referenced).
- The shared module has **zero imports** — no server-only code reaches the client bundle. `Avatar styles.placeholder` is valid for Mantine `^7.17.7`.

### 🔴 Correcting the PR description

The body says *"no behaviour change for a listing that has both assets"*, which is true — but it **understates a real user-visible change for iconless listings**, and I'd rather correct that here than silently edit the body someone may already have read.

Switching the Avatar fallback from `name.charAt(0)` to `appInitial()` inherited that helper's ASCII-only `[A-Za-z0-9]` class:

| name | before | after (as merged in `bc04295`) |
|---|---|---|
| `日本語アプリ` | `日` | **`?`** |
| `Привет` | `П` | **`?`** |
| `العربية` | `ا` | **`?`** |
| `Émile` | `É` | `M` |

App names are unrestricted free text (`offsite-listing.schema` bounds only length), so this is reachable — and rendering a literal `?` for a real app is exactly the "looks broken" outcome this PR exists to prevent.

**Fixed** by widening to `/[\p{L}\p{N}]/u`. The `u` flag is load-bearing: it makes the match code-point-based, so an astral-plane initial returns a whole code point rather than a lone broken surrogate half. It now also falls through to the **slug** when the name yields no letter/digit (`—`, `🎨`) instead of giving up at `?`. Emoji/punctuation-leading names still skip to the first real letter.

Side effect, strictly an improvement: future **generated** icons for non-Latin apps get the right glyph. Previously-generated PNGs are untouched (the backfill only fills NULL), so no divergence.

### 🟡-2 — the drift guard was weaker than its name

I'd claimed "provably one colour identity". The audit showed that was stronger than the evidence: **two mutations survived** the original guard —

1. **swapping the two gradient stops** — an inverted gradient on every generated icon — passed all 78 tests, because two order-agnostic `toContain(stop)` calls never bind offset→colour;
2. **re-hardcoding `linear-gradient(135deg, var(--mantine-color-dark-5)…)` back into the component** — nothing in the blocking `unit` project failed, since no `.test.ts` imports these components and their `.browser.test.tsx` suites are report-only.

Fixed three ways: assert `offset="0%" stop-color="…"` as **one** string (pins direction); cross-check **both** surfaces with full-string equality (`icon` backs the Avatar, `cover` backs the card — checking only `cover` left the other free to drift); and add a **source-level component re-coupling guard**, mirroring the existing `block-scope.schema-drift` `readFileSync` convention, because a value test structurally cannot see a component reverting to its own hardcoded gradient.

### Verification

All three fixes **mutation-verified**, each reverted and re-confirmed green afterwards:

| mutation | result |
|---|---|
| swap gradient stops (offset↔colour) | **FAIL** ✅ |
| re-hardcode component gradient | **FAIL** ✅ |
| revert to ASCII `[A-Za-z0-9]` | **3 FAIL** ✅ |

**97 → 108 tests**, all passing. Prettier clean on added files; ESLint 0 errors (the one warning is a pre-existing unused `ThemeIcon` on `main`).

One honest residual: re-inlining an *identical* literal still passes, since a value-equality test can't detect decoupling that doesn't change the value. The source-level guard covers the component case; the service case would need AST analysis and isn't worth it.

Nits also fixed: stale "neutral gradient" docstrings in both components, and the `data-*` attribute convention normalised to `=""`.

PR #3466, opened 2026-07-30
fix(images): verify the uploaded object exists before creating a cover Image row

ZacxDev:
## The bug

The image **delete** path is url-aware and refcounted — `deleteImageFromS3` looks up every other `Image` row sharing the same `url` before it removes the underlying object. The **create** path was url-*blind*. Every cover-image call site was:

```ts
coverImage.id ?? createImage({ ...coverImage, userId })
```

a bare row insert with no check that the object the row points at still exists.

That asymmetry has a live symptom. A cover upload hands the client a raw object key with **no `Image.id` yet**. `useFormStorage` persisted that draft to `localStorage` with **no expiry**, and restored it by merging the persisted value **over** the freshly loaded server value (`{ ...form.getValues(), ...initialValue }`) — so the persisted, id-less copy won. A long-abandoned draft could therefore be replayed weeks later and round-trip as "a brand new image": the server minted a fresh `Image` row pointing at an object that had since been deleted. The result is a permanently broken image for every reader, and no error at all for the person who caused it.

Both halves are fixed here.

## Server — reuse, then verify, then create

New shared `resolveCoverImageId` (`src/server/services/cover-image.service.ts`), used by the article cover path and all three challenge cover paths (`upsertChallenge`, `upsertUserChallenge`, `upsertChallengeEvent`).

- **`id` present** → the call site's *existing* ownership rule runs, unchanged, and we return the id. **No url lookup and no bucket round-trip** — this is the overwhelmingly common path and it stays free. Ownership is deliberately *not* centralised: each site keeps its own rule (article: skip when the cover is unchanged, else `isImageOwner`; user challenges: must own the row; moderator paths: no check) and passes it in as `assertOwnership`. Unifying those would be a behaviour change well beyond this bug.
- **`id` absent** →
  1. reuse an existing, *safely reusable* `Image` row for the same url, else
  2. verify the object is actually in the uploads bucket before inserting, and reject with a user-actionable `TRPCError` when it definitively is not (*"This cover image is no longer available. Please re-upload it."*).

Two deliberate narrowings worth reviewing:

**Reuse is scoped to the caller (`where: { url, userId }`).** `Image.url` is the bare object key and it appears in every public image URL, so an *unscoped* url→row reuse would let anyone attach someone else's (possibly unpublished, blocked or moderated) image as their own cover just by reading a key off the site. Scoping to the caller produces exactly what `createImage` would have produced — a row owned by the caller — minus the duplicate.

> **Correction (was wrong in the first version of this description).** This originally claimed `Image.url` carries a *hash* index. It does not, and more to the point: **the `url` index this lookup leans on is declared in no Prisma schema and in no migration in this repo**, so it must not be assumed present in a fresh or local database — on one without it, this is a sequential scan of `Image`. The code comment has been corrected to say that. The conclusion is unchanged (it is an equality probe, on the no-id path only, capped at 5 rows), but reviewers should not take the index as something the repo guarantees.

**Reuse only accepts an *unattached* row (or the entity's own current cover).** Sharing one row across two entities is not free: `Article.coverId` is `@unique` (so re-pointing a row that already covers another article would fail the write outright), and deleting an entity deletes its cover `Image` row — so a shared row means deleting one entity breaks the other. That would trade this bug for a different broken image. `isReusableCoverImage` encodes the predicate and is unit-tested per attachment kind.

**The existence check reuses the repo's `HeadObject` helper**, `checkFileExists` in `src/utils/s3-utils.ts`, extended rather than duplicated:

- it now takes an options bag so a caller can pass the client + bucket (its previous default bucket is the model-file bucket, not the image uploads bucket, and the legacy image client can't be used for this — the existing `announcement-media-check` job documents why);
- it now returns **tri-state** `true | false | null`. Only a definitive `false` (404 / `NotFound`) rejects. `null` means *the bucket could not be consulted* — missing credentials, a 403 from a rotated key, a network blip — and the caller **fails open**. Collapsing that into `false` (what the helper used to do) would turn an infrastructure hiccup into a user-facing rejection on every upload. Same fail-open discipline the announcement media check already applies.

`checkFileExists` had **zero call sites** before this PR, so the signature change is contained.

## Client — bound the draft, and stop it clobbering server state

`src/hooks/useFormStorage.ts`:

**1. TTL.** Entries are now written as `{ savedAt, value }` and expire after **7 days** (`FORM_STORAGE_TTL_MS`). The modal offers "the unsaved changes from your previous session", which in practice means hours to a few days; a week comfortably covers "I left this open Friday, came back Monday", while bounding the month-old-draft case that produced this bug. The exact number is a judgement call — the server-side check above is the actual correctness guarantee, the TTL just shrinks the exposure. A **legacy unstamped entry is treated as expired**: it predates write-stamping, so its age is unknowable, and unbounded age is precisely the hazard. One-time cost, small (a draft mid-flight across the deploy loses its restore offer).

**2. Merge inversion — fixed opt-in, not globally.** `useFormStorage` is generic and shared, so the default merge direction (`{ ...current, ...stored }`, persisted wins) is **unchanged** — that is the right semantic for "restore my unsaved edits" and I did not want to change it under every form. Instead the hook gains an optional per-field veto, `shouldRestoreField`. `ArticleUpsertForm` opts in for `coverImage` via the exported `persistedValueLostIdentity` predicate.

That predicate is deliberately narrow: it vetoes **only** when the persisted value is the *same object* (same `url`) as the loaded one but has lost the server-assigned `id`. If the user genuinely swapped in a different image while offline, the urls differ, it is a real edit, and it still wins — with the server-side existence check as the backstop if that new object is gone.

**Callers enumerated (denominator = 2).** `ArticleUpsertForm.tsx:128` and `BountyUpsertForm.tsx:191` are the only two `useFormStorage` call sites in the codebase. Bounty persists `mode / name / type / currency / description / entryMode / unitAmount` — no image field — so it is unaffected by the veto and picks up only the TTL.

### 🔴 Known limitation — the veto is structurally inert for a NEW article

Worth being explicit about, because it may well be the **most common** instance of the reported bug rather than an edge case.

`persistedValueLostIdentity` vetoes a restore only when the *currently loaded* value has a server-assigned `id` to be protected. It returns `false` as soon as `currentValue` has no `id`. For a new article there is no loaded value at all — `ArticleUpsertForm.tsx:124` sets `defaultValues.coverImage` to `null`, and the draft is keyed `articlenew` (`:132`). So on the new-article path the predicate can never fire, and a stale draft still restores its dead cover exactly as before.

This is the predicate behaving **as designed** — there is genuinely no server value to prefer, so there is nothing for the client to fall back to, and "the persisted draft wins" remains the only sensible client-side answer. But it means the client half of this PR does **not** cover the new-article case. What protects that case is (a) the 7-day TTL, which bounds how stale a replayable draft can be, and (b) the server-side existence check, which is the actual correctness guarantee and applies to every path equally.

I deliberately have not coded around this: making the veto fire without a loaded value would mean discarding the user's draft cover on a form where it is the only cover they have.

## Audit follow-ups (added after review, as commits on top)

The fail-open decision is unchanged; these make it bounded, measurable and actually executed.

**Observability on the fail-open branch.** Proceeding on an indeterminate check is correct but completely silent — while it is happening the guard is off, and "the check found nothing bad" is indistinguishable from "the check never ran". Both indeterminate branches now emit a structured warning through the repo's existing `logToAxiom` path, under one queryable name, separated by a `reason` field: `client-unavailable` (the uploads client could not be built) vs `proceeded-unverified` (the bucket was consulted, gave no definitive answer, and a row was minted anyway). The decision point goes through a new injected `logExistenceUnknown` dep so it is asserted in the unit suite, not just in production. The object key is deliberately **not** logged.

**Timeout budget on the HEAD probe.** This runs on the user-facing save path against a client with SDK-default retries and no request timeout, so a degraded backend turned a bounded-latency guard into an unbounded one. `checkFileExists` now takes an `abortSignal` and the cover path passes `AbortSignal.timeout(2000)` — the same bound already used for the best-effort image-cacher invalidation in `image.service`. One signal is shared across every retry attempt, so it **bounds every network attempt**. Verified rather than assumed: an abort surfaces as a plain `AbortError` with no `$metadata`, which is not a not-found shape, so it lands on `null` and falls open through the **existing** branch — no new branch logic was needed.

> **Correction (this description and the code comment both over-claimed).** An earlier version said the signal "caps the **whole** call". It does not. The SDK's retry middleware sleeps between attempts with a plain, non-abort-aware timer, so a deadline landing mid-backoff lets that sleep run to completion and only the *next* attempt short-circuits. **Worst-case wall time is the budget plus one backoff, not the budget.** Measured against the installed SDK against a local always-503 endpoint: a 300 ms budget with a ~5 s backoff in flight returned in ~4.7 s, having made exactly one network attempt. The abort still terminates the call promptly after that sleep — `AbortError` is not classified retryable, so the loop exits rather than backing off again. Both the doc comment on `checkFileExists` and the test comment now say this.

**The SDK error-shape mapping is now executed, not just reasoned about.** The original claim that a throttled `503 SlowDown` maps to `null` (proceed) rather than `false` (reject) was verified by code inspection only. There are now tests driving real rejection shapes through `checkFileExists`: `SlowDown`/503, 403, 500, a transport error with no status, and an abort all map to `null`; `NotFound`, `NoSuchKey` and a bare 404 map to `false`. Both directions are mutation-verified — collapsing the tri-state either way fails the suite.

**The best-effort logging `.catch()` is now pinned by a test that can actually fail.** A previous round concluded this was untestable and committed a comment saying so (*"do not add a test for it here"*). That generalisation was wrong. It is untestable through a `vi.fn()` mock — a `vi.fn()` attaches its own handler to any promise the implementation returns (that is how `mock.settledResults` is populated), which marks a rejection handled, so Node never emits `unhandledRejection` and the assertion passes with or without the production `.catch()`. But the mock never had to be a `vi.fn()`: with `logToAxiom` mocked as a **plain function** returning a rejected promise, a `process.on('unhandledRejection')` assertion fails the moment either `.catch()` is deleted. Because the existing suite needs `logToAxiom` to be a `vi.fn()` to assert on payloads, and `vi.mock` is per-file, the falsifiable version lives in a second file (`cover-image.service.logging.test.ts`). Mutation-verified per branch: removing each `.catch()` individually fails exactly the test that covers it and leaves the other green. The misleading comment has been replaced with the trap itself (why a `vi.fn()`-based assertion is unfalsifiable) and a pointer to the working test.

**The existence-check budget is pinned as a ceiling, not a value.** `COVER_IMAGE_EXISTS_TIMEOUT_MS` had its *wiring* pinned (that `AbortSignal.timeout` is called with it) but not its magnitude, so it could be widened to anything with the suite still green. Rather than re-state `2000` in a test — a change-detector that only ever fails when someone edits it deliberately — the suite now asserts the invariant the design argument actually rests on: the budget stays inside a range that can sit on a user-facing save path. Retuning it is a normal decision and stays green; widening it into "this visibly stalls a save" territory does not.

**Two corrections to this description and the code comments** — the `Image.url` index claim (see the correction note above) and `isReusableCoverImage`'s doc comment, which described itself as meaning "attached to nothing at all" when it checks five relations out of the many on `Image`. The comment now names the five and states that the list is partial. The relation coverage itself is unchanged — widening it is a separate decision.

**Two nits.** A dead `let coverId = coverImage?.id` initializer in `upsertArticle` (the only branch where it is non-`undefined` immediately overwrites it), and a one-line `__proto__` guard in `mergeRestoredValues` — that loop assigns where it used to spread, and assignment hits the inherited accessor instead of creating an own property. Self-only blast radius (the source is the user's own `localStorage`), but it is never a real form field.

## Tests

`src/server/services/__tests__/cover-image.service.test.ts` (25), `src/server/services/__tests__/cover-image.service.logging.test.ts` (2), `src/hooks/__tests__/useFormStorage.test.ts` (18), and `src/utils/__tests__/s3-utils.test.ts` (+10 in an existing file).

Service coverage: id-present short-circuit (asserting the reuse lookup, bucket probe and insert are *not* called), ownership-rejection propagation, existing-row reuse, definitive-miss rejection, fail-open on an unknown bucket answer, the reuse predicate per attachment kind, the caller-scoped url lookup, that the uploads probe reports *unknown* rather than *absent* when the client can't be built, that the probe is bounded by an abort signal whose budget stays inside the user-facing range, and — in the companion file — that a failing `logToAxiom` cannot escape as an unhandled rejection.

Hook coverage: TTL fresh / expired / boundary / legacy / unparseable, the write stamp, the identity predicate in all four directions, default merge unchanged, veto merge, and the hook itself (mounted for real) not offering an expired draft, clearing it, and not clobbering a loaded cover on confirm.

Every test was mutation-verified — the corresponding fix reverted, the test confirmed to fail, then restored. One assertion was found to be **structurally unfailable** on the first pass (it asserted `id` was stripped from the insert payload, but the fixture carried no `id` at all, so leaking it changed nothing) and was rewritten with an explicit `id: undefined` — the shape an optional zod field actually produces — at which point the mutation does fail it.

**One property is deliberately left uncovered, and the attempt to cover it is documented in the test file so nobody re-adds a fake one.** That `logToAxiom(...)` carries a `.catch()` (so a logging failure cannot surface as an unhandled rejection) is *not* unit-testable here. The only observable difference between having and not having that `.catch()` is whether Node emits `unhandledRejection` — and it never can through this suite, because Vitest attaches its own handler to any promise returned from a `vi.fn()` in order to populate `mock.settledResults`, which marks the rejection handled. Measured directly rather than assumed: a raw `Promise.reject()` in a test **is** seen by a `process.on('unhandledRejection')` listener; the identical rejection returned from a mocked `logToAxiom` is **not**. A first draft of that test passed with the `.catch()` removed, i.e. it was unfalsifiable — so it was deleted rather than shipped green. The guarantee rests on review and on the same `.catch(() => {})` pattern already used in `image.service`.

One real regression was caught this way and fixed before opening: `productionCoverImageDeps` originally referenced `createImage` at **module load**, and because this module is now in the article/challenge import graph, that broke every existing suite which wholesale-`vi.mock`s `image.service` without re-stubbing `createImage` (`No "createImage" export is defined on the mock` → the whole test file fails to import). The dep is now a thin wrapper so the binding is read at call time.

## Verification

Run against a clean worktree off `origin/main`, with the Prisma client generated and the `event-engine-common` submodule populated (without those two, local `tsc` produces a large volume of unrelated noise — every result below was baselined against the unmodified `origin/main` in the same environment, and only the delta is reported).

- `NODE_OPTIONS=--max_old_space_size=8192 npx tsc --noEmit` — **0 errors**. (Note for anyone reproducing: `tsconfig.json` sets `incremental: true`, so delete `tsconfig.tsbuildinfo` before trusting a run.)
- `vitest run --project unit` — the failing set is **byte-identical to `origin/main`'s** (same files, same tests); this branch adds `+2` test files and introduces **no new failure**.

Re-run after the audit follow-up commits, baselined against the pre-follow-up tip in the same environment:

- `tsc --noEmit` — **0 errors**, unchanged.
- `vitest run --project unit` — **9117 tests, 7 failed / 9109 passed / 1 skipped** vs **9099 tests, 7 failed / 9091 passed / 1 skipped** before. Delta: **+18 passing, 0 new failures, 0 new failing files.** The 7 failures are the pre-existing `src/components/AppBlocks/__tests__/hiddenBlocks.test.ts` set (happy-dom has no `localStorage`), identical to `main`.

## Follow-ups (deliberately not in this PR)

- **Pass `currentCoverId` at the three challenge cover call sites** (`challenge.service.ts` `upsertChallenge`, `upsertUserChallenge`, `upsertChallengeEvent`). None of them pass it, so a challenge re-saved with an id-less cover that matches its **own** current row is refused reuse and mints a duplicate `Image` row. This is **not a regression** — it is exactly the pre-PR behaviour, and the article path simply gains a dedupe the challenge paths don't. Left out on purpose: at each site `resolveCoverImageId` is called *before* the existing challenge row is loaded (deliberately — a misconfigured judge must not leave an orphaned cover behind), so wiring it up means either an extra query on every challenge save or reordering that load, and neither is worth folding into this diff.

- **A DB-level FK on `Article.coverId`.** Prisma declares the relation but no constraint exists. Migration SQL in this project is applied manually per environment by a human, so no migration is written or run here — this should be raised as its own change with the raw SQL and the target environment stated.
- **Extend the existing `announcement-media-check` job to article/challenge covers.** It already performs exactly the right classification (BROKEN = key absent from the uploads bucket; AT RISK = a deletable `Image` row shares the key) for announcement banners; covers are the same shape of reference. That also closes the gap for rows created *before* this fix, which this PR does not backfill.


ZacxDev:
**Correction to the timeout-budget note in this description** (posting rather than editing the body, since the current numbers may already have been read).

The description's own "Correction" paragraph is right that worst-case wall time is *budget plus one backoff*, but the only figure it gives — **~4.7 s** — was measured with a **300 ms** budget against a ~5 s backoff. It is not the worst case at the budget that actually ships.

The SDK honours a server-supplied `Retry-After` up to `MAXIMUM_RETRY_DELAY = 20 s` (`@smithy/util-retry`), and the backoff sleep is a plain non-abort-aware `setTimeout` (`@smithy/middleware-retry`). So at the shipped `COVER_IMAGE_EXISTS_TIMEOUT_MS = 2000`, a backend that responds `Retry-After: 20s` yields a worst case of **~22 s**, not ~4.7 s — roughly 4× what the description implies, on the user-facing save path.

This matters because B2 sends `Retry-After` precisely under throttle, which is the condition this probe is most likely to meet. Behaviour is still strictly better than the pre-PR unbounded case and it fails open rather than losing data, so this is not a merge blocker — but a reviewer calibrating risk from the ~4.7 s figure would size it wrong.

Also worth noting: the ceiling assertion in `cover-image.service.test.ts` pins the budget into `[1, 5000]`, which proves the budget is small — it does not prevent a visibly stalled save, because the stall comes from the un-interruptible backoff rather than from the budget value.

If bounding it for real is wanted later, the options are a `requestHandler` timeout, `maxAttempts: 1` on this specific call, or clamping how far `Retry-After` is honoured. Tracking as a follow-up rather than expanding this PR.

PR #3480, opened 2026-07-30
feat(apps): bounded auto-retry + unmissable manual Retry on the app run page

ZacxDev:
## The problem

A product owner opened an app from `/apps`, it failed to load, and their words were:

> there might have been a retry button but I didn't notice — I did a full page reload

The retry **logic** was not broken. `PageBlockHost.handleRetry` already disposed the controller, re-armed the handshake, re-keyed the iframe, and — for the auth terminals — re-minted the token first, because a local remount can never clear an auth failure. What was missing:

1. **No automatic recovery.** A transient blip needed a human to intervene.
2. **The manual affordance was easy to miss** — a small subdued `xs` button.
3. **No in-progress feedback.** Pressing Retry against a still-down host waited ~10s in silence and re-rendered the identical error, which reads as "the button does nothing".

## What changed

`PageBlockHost` now re-attempts the load itself, bounded, and only then settles on a definitive terminal state whose manual Retry is prominent.

### Retry policy

| | value | why |
|---|---|---|
| Automatic attempts | **2** (`MAX_AUTO_RETRIES`) | Enough to ride out a blip/pod-drain; past that a failure is real and the user should get a definitive answer rather than a machine that keeps trying. |
| Backoff | **2 s, then 5 s** | The first pause is long enough that a rebooting host has a chance, short enough that the page settles quickly. Worst case to the definitive terminal is bounded. |
| Automatic re-mints | **1** (`MAX_AUTO_REMINTS`), **strictly below** the attempt cap | Only the auth terminals (`error`/`no_token`) re-mint, and `/api/v1/block-tokens` is rate-limited (60/min). It must stay strictly lower or it is unreachable dead code — `reminted` is a subset of `attempts` and the attempt cap is checked first. A test pins the inequality. |
| Auth terminal with **no** re-mint available | **stop immediately** | A remount alone can never clear an auth failure (the token is an upstream prop), so another attempt is a guaranteed re-fail. |
| Manual Retry | **not capped here**, and does not consume the automatic budget | User-initiated; bounded by the existing double-click guard plus the server-side rate limit. The user taking over shouldn't spend the platform's remaining recovery attempts. |

Budgets are **per mount** and are **not refilled** by a manual Retry — once spent, every later failure goes straight to the definitive terminal.

### The terminal state is never masked

This is the constraint I held hardest. The real error renders the instant the status goes terminal; the pending automatic attempt is surfaced **inside that same fallback** ("Retrying automatically… (attempt 1 of 2)"), never instead of it. There is no "keep spinning quietly while we retry" state — that would recreate the silent-blank failure class this codebase has fought repeatedly. A regression test measures the fallback's arrival against a captured timestamp and asserts it lands well inside the backoff window.

During an actual re-attempt the launch overlay says `Retrying <app>…` instead of an identical `Starting <app>…`. It deliberately carries no attempt number — the bounded count belongs on the terminal card, where the budget is meaningful; showing both put two disagreeing counters two seconds apart on the same surface.

### Other behaviour

- **Once settled**, the Retry button becomes filled + full-width + `md`, and the host states plainly that it already retried. Accessible name stays `Retry` in both states.
- **Manual + automatic share one re-arm path** (`performRetry`). A manual Retry during a pending automatic one flips the status back to `loading`, which makes the decision `none`, which tears down the pending timer in the effect cleanup — exactly one attempt runs.
- **`prefers-reduced-motion: reduce`** suppresses the *new* retry spinner in the terminal card; the copy carries the state. (The pre-existing launch overlay's `Loader` is still animated — see the corrections comment.)
- **`AppBlockChrome` still wraps every state** (FRAME-1), and the `#3457` launch reveal is untouched — its suite passes with this change in place (see below).

### Beacon semantics

**One beacon per host mount, reporting the outcome the host SETTLES on.** The whole bounded retry sequence is one logical page load:

- `ok` — the first `BLOCK_READY`, whichever attempt produced it. A load that succeeds on attempt 2 emits exactly **one** `ok` and never an `error` first.
- `error` — a terminal state reached with **no** further automatic attempt coming. N failed automatic attempts emit **one** `error`, not N.
- nothing — an intermediate terminal that will be auto-retried.

Once either fires the mount never emits again; `handleRetry` deliberately no longer resets the emit-once ref. So the alert's denominator stays "page loads" and its numerator stays "page loads the platform could not recover on its own". `errorClass` stays inside the existing server-side `KNOWN_ERROR_CLASSES` enum, so no metric **label** change is needed — but the numerator's MEANING narrows, and the `datapacket-talos` alert wording needs a companion update. See the corrections comment.

## A real bug found while writing the tests

The backoff timer must **not** be keyed on the `onRetryToken` callback identity. `ReviewBlockPreviewHost` passes an inline arrow (`onRetryToken={() => doMint(...)}`), so it is a new function on every render — with the effect depending on it, every re-render would tear down and re-arm the pending timer, the backoff would never elapse, and **auto-retry would silently never fire** for that caller. The retry is read through a ref so the effect is keyed on the decision alone. Regression test included (a parent that re-renders every 100 ms with a fresh callback).

## Tests

`decideAutoRetry` is pure, so every bound is unit-testable without paying the real 10 s/15 s timer windows; the browser suite drives the same bounds through the real host.

- **`__tests__/pageBlockHostLogic.test.ts`** — 47 tests (15 new): fires from every terminal reason, bounded at `MAX_AUTO_RETRIES`, backs off strictly increasing and never 0, only auth terminals re-mint, the re-mint cap walked from a reachable state, the strict `MAX_AUTO_REMINTS < MAX_AUTO_RETRIES` inequality that keeps it from going inert, no-re-mint-wired guard, backoff table covers the budget.
- **`PageBlockHostAutoRetry.browser.test.tsx`** — 14 new tests: terminal renders immediately with the pending retry inside it; the attempt actually re-runs after the backoff and says so; fires from `timeout`/`no_token`/`fatal`; non-auth doesn't re-mint; survives an unstable callback; re-mint cap holds and then stops; prominent Retry after exhaustion; beacon exclusivity both directions; reduced motion; manual-during-automatic; unmount mid-backoff leaks no timer.

**This suite exercises the real path** — real `PageBlockHost`, real hooks, real timers. It does **not** mock `@mantine/hooks`: the reduced-motion case stubs `window.matchMedia`, so `useReducedMotion` itself still runs. Only the ambient tRPC client and `useCurrentUser` are stubbed, exactly as the sibling suites do.

## Incidental repair (same surface)

`PageBlockHostLaunchReveal` and `IframeHostReadyTransition` were the **only** two host suites whose tRPC mock lacked `apps.shared.report`, which both hosts call unconditionally at render — so all **35** of their tests were crashing at mount with `Cannot read properties of undefined (reading 'useMutation')`. These checks are report-only, so both suites had been silently red. They are the guards for the launch-reveal and beacon-exclusivity invariants this change must not break, so they are repaired here and now pass 35/35.

## Verification

| check | result |
|---|---|
| `pageBlockHostLogic` unit suite | 47/47 |
| `PageBlockHostAutoRetry` browser suite | 14/14 |
| `PageBlockHost` browser suite (existing) | 25/25, unchanged |
| `PageBlockHostLaunchReveal` (repaired) | 10/10 |
| `IframeHostReadyTransition` (repaired) | 25/25 |
| `IframeHost` browser suite | 5/5 |
| `tsc --noEmit` | clean |

Every guard was mutation-verified: broken, confirmed a *specific* test fails, reverted. 13 mutations, 13 kills — including two round-trips where a guard **survived**: the first emit-once guard (which prompted the extra `ok`-then-crash test) and, found by the adversarial audit, a re-mint cap that was unreachable dead code (see the corrections comment).

## Not verified by me

`/apps` is mod-gated, so I have **not** clicked this path in a browser. Manual steps are in the review thread.

### DIFF
--- a/src/components/AppBlocks/BlockFallback.tsx
+++ b/src/components/AppBlocks/BlockFallback.tsx
@@ -1,4 +1,4 @@
-import { Alert, Button, Skeleton, Stack, Text } from '@mantine/core';
+import { Alert, Button, Group, Loader, Skeleton, Stack, Text } from '@mantine/core';
 import { IconRefresh } from '@tabler/icons-react';
 
 type FallbackReason = 'loading' | 'token_error' | 'timeout' | 'fatal_block_error';
@@ -16,6 +16,33 @@ interface BlockFallbackProps {
    * for the non-terminal `loading` reason.
    */
   onRetry?: () => void;
+  /**
+   * In-progress feedback for the host's BOUNDED AUTO-RETRY. When set, the host
+   * has already scheduled another automatic attempt: the fallback still shows the
+   * REAL terminal message (the error is never masked or delayed) and adds a line
+   * saying a retry is under way and which attempt it is — so a user who then
+   * presses the button doesn't experience "the button does nothing".
+   *
+   * `attempt`/`maxAttempts` are 1-based. `animate:false` (driven by
+   * `prefers-reduced-motion: reduce` upstream) suppresses the spinner; the copy
+   * alone then carries the state.
+   */
+  autoRetry?: { attempt: number; maxAttempts: number; animate: boolean };
+  /**
+   * Number of AUTOMATIC attempts already spent. Renders an honest "we already
+   * tried N times" note — only when > 0, so a surface with no auto-retry wired
+   * never claims a retry that didn't happen.
+   */
+  autoRetriesSpent?: number;
+  /**
+   * Make the manual Retry affordance UNMISSABLE. Set by PageBlockHost once the
+   * host has SETTLED (no further automatic attempt is coming) — that is the
+   * moment the user must notice the button, which is precisely what was reported
+   * as missed. Renders a filled, full-width, medium button instead of the small
+   * subdued default. Default `false` keeps the model-slot IframeHost and every
+   * other existing caller byte-identical.
+   */
+  prominentRetry?: boolean;
 }
 
 const DEFAULT_MIN_HEIGHT = 200;
@@ -53,6 +80,9 @@ export function BlockFallback({
   blockName,
   minHeight = DEFAULT_MIN_HEIGHT,
   onRetry,
+  autoRetry,
+  autoRetriesSpent = 0,
+  prominentRetry = false,
 }: BlockFallbackProps) {
   if (reason === 'loading') {
     return <Skeleton h={minHeight} radius="md" data-block-fallback="loading" />;
@@ -68,15 +98,47 @@ export function BlockFallback({
     <Alert color={copy.color} title={title} data-block-fallback={reason}>
       <Stack gap="sm">
         <Text size="sm">{copy.hint}</Text>
+        {autoRetry ? (
+          // Live region: the attempt count changes while the user is already
+          // looking at a static error, so announce it instead of mutating the
+          // text silently.
+          <Group
+            gap="xs"
+            data-block-fallback-autoretry="true"
+            // Observable animation state — `false` under prefers-reduced-motion.
+            data-block-fallback-autoretry-animate={autoRetry.animate ? 'true' : 'false'}
+            role="status"
+            aria-live="polite"
+          >
+            {/* Reduced motion: no spinner at all — the copy carries the state. */}
+            {autoRetry.animate && <Loader size="xs" aria-hidden />}
+            <Text size="sm" c="dimmed">
+              {`Retrying automatically… (attempt ${autoRetry.attempt} of ${autoRetry.maxAttempts})`}
+            </Text>
+          </Group>
+        ) : autoRetriesSpent > 0 ? (
+          <Text size="sm" c="dimmed" data-block-fallback-autoretry-spent={String(autoRetriesSpent)}>
+            {autoRetriesSpent === 1
+              ? 'We already retried once automatically.'
+              : `We already retried ${autoRetriesSpent} times automatically.`}
+          </Text>
+        ) : null}
         {onRetry && (
           <Button
-            variant="light"
+            // Once automatic recovery has settled, the button IS the path
+            // forward — so it stops being a small subdued affordance.
+            variant={prominentRetry ? 'filled' : 'light'}
             color={copy.color}
-            size="xs"
-            leftSection={<IconRefresh size={16} />}
+            size={prominentRetry ? 'md' : 'xs'}
+            fullWidth={prominentRetry}
+            leftSection={<IconRefresh size={prominentRetry ? 18 : 16} />}
             onClick={onRetry}
             data-block-fallback-retry="true"
+            data-block-fallback-retry-prominent={prominentRetry ? 'true' : 'false'}
           >
+            {/* The accessible name stays "Retry" in BOTH states: existing
+                callers/tests target one stable name, and a user reading the
+                button never has to re-learn it mid-failure. */}
             Retry
           </Button>
         )}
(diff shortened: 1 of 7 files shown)


ZacxDev:
## Corrections to the PR description (an adversarial audit found it overstated three things)

Posting these rather than silently editing the body, since a reviewer may already have read the original.

**1. The re-mint cap in the original description was inert.** The body presented `MAX_AUTO_REMINTS = 2` as "the specific cap that keeps an auth-failure loop off" the rate-limited mint endpoint. It could never fire. `reminted` is a strict subset of `attempts` (every re-minting attempt increments both), so `reminted <= attempts` always holds — and `decideAutoRetry` checks the total-attempt cap **first**. With both caps set to 2, the re-mint branch was unreachable dead code. Worse, both tests that "proved" it were actually being killed by the *other* guard, and one of them asserted on a state the runtime cannot produce.

Fixed in `91672cd3`: `MAX_AUTO_REMINTS` is now **1**, strictly below `MAX_AUTO_RETRIES`. The auth path — the expensive one, since only it re-mints — stops after one automatic re-mint; non-auth terminals keep the full two attempts. A test now pins the strict inequality so it can't silently go inert again, and the bound test walks the auth path from a fresh mount instead of a hand-made state. Both are mutation-verified from reachable states.

**2. "`prefers-reduced-motion` suppresses the retry spinner entirely" was too strong.** It suppresses the *new* spinner in the terminal card. The pre-existing launch overlay still renders an animated `Loader` during each retry attempt — and because this PR increases time spent in that state, reduced-motion users actually see *more* spinner, not less. Pre-existing, not introduced here, but the claim as written was wrong.

**3. "one stub line each" is off by one.** `PageBlockHostLaunchReveal.browser.test.tsx` also needed `get: { fetch: vi.fn() }` in `useUtils`.

**4. "no metric/label/alert change is needed" is true for labels only.** `errorClass` values are unchanged and still inside the server-side `KNOWN_ERROR_CLASSES` set, so nothing breaks. But the *meaning* of the `error` numerator narrows to "failures the platform could not self-recover" — a regression that only increases **transient** launch failures is now invisible to a ratio built on it. The `BlockRenderFailureRate` alert in `datapacket-talos` still describes it the old way and needs a companion wording update. Noted in code at the beacon effect.

## Numbers the description should have stated

- **Worst-case time to a definitive error:** ~37 s on the `timeout` path (10 + 2 + 10 + 5 + 10), ~32 s on the `no_token` path now that the auth branch stops after one re-mint (15 + 2 + 15). Roughly 3× the previous time-to-answer. The user sees the real error and the manual Retry throughout, so it is informed waiting — but it is a real increase and reviewers should weigh it.
- **Kill switch:** there is no feature flag on this path. `MAX_AUTO_RETRIES = 0` is a clean one-line disable (every decision returns `none`, the terminal renders immediately with the prominent button, and `autoRetriesSpent` stays 0 so no false "we already retried" copy). Documented on the constant.

## Known gaps deliberately left in (not fixed here)

- **Unmount mid-sequence emits no beacon at all.** A user navigating away during a retry produces no telemetry, where before they produced an `error`. The window grew from ~0 s to the length of the sequence. An unmount flush is the obvious follow-up — `sendBlockRender` already sets `keepalive: true` for exactly that.
- **Auto-retry is not scoped to the launch phase.** A block that crashes *after* reaching ready gets relaunched too, including in the mod review sandbox under `reviewRunForReal` (bounded by the existing Buzz cap). Arguably desirable, but it is a product decision that should be made explicitly rather than inherited.
- **An automatic re-mint aborts `useBlockToken`'s in-flight 429 backoff.** Bounded to one extra mint per mount now, but during a platform-wide 429 it does not honour the jittered pause.

PR #763, opened 2023-10-04
Feature/buzz merchant  purchase page

lrojas94:
Also in this PR:
- Bounties support for purchasing BUZZ inline.

image
image

PR #3482, opened 2026-07-30
fix(tests): repair the component-suite baseline (import failures + timeout-burning mocks)

ZacxDev:
These are **pre-existing failures on `main`**, not regressions from any open PR. Every change is **test-side only** — no product source is modified.

The `component` project sets no `testTimeout`, so it inherits the 5s default. A large share of the suite's clock was these failures idling on that timeout, and five whole files were failing to *import* — collecting 0 tests, which reads as "green" because `component-tests` is report-only.

`src/tests/pages/apps/listing-media-page.browser.test.tsx` and `src/pages/apps/[appBlockId]/listing.tsx` are **deliberately excluded — #3476 owns those.**

## Measured before / after

Each file run in isolation (`vitest run --project component <file>`), real Chromium.

### Group A — whole-file import failures (collected 0 tests)

| File | Before | After |
|---|---|---|
| `Apps/OffsiteReviewQueue.browser.test.tsx` | **0 collected**, 2.00s | 21 passed, 4.89s |
| `Apps/CombinedReviewModal.browser.test.tsx` | **0 collected**, 2.29s | 3 passed, 2.09s |
| `Apps/ConnectScopesPanel.browser.test.tsx` | **0 collected**, 2.02s | 6 passed, 2.71s |
| `Apps/OffsiteReportsQueue.browser.test.tsx` | **0 collected**, 2.03s | 4 passed, 2.93s |
| `tests/pages/apps/review/review-queue-nav.browser.test.tsx` | **0 collected**, 2.23s | 2 passed, 2.36s |

Four of these wholesale-`vi.mock('~/utils/trpc')` and omitted `trpcVanilla` (`src/utils/trpc.ts:371`), so the file's ESM link failed:

```
SyntaxError: The requested module '/src/utils/trpc.ts' does not provide an export named 'trpcVanilla'
```

Fixed with the canonical in-repo `importOriginal`-spread partial mock (matching `Challenge/__tests__/ChallengeUpsertForm.browser.test.tsx:27`, which carries a comment naming this exact trap) so the next added export can't break them again. The fifth wholesale-mocks `OffsiteReviewQueue` and was missing `OffsiteReviewModalBody`, which `CombinedReviewModal` imports.

Two **stale assertions** surfaced underneath, both because the same text is now rendered twice and a strict single-match query throws. Both were scoped, not loosened:

- `CombinedReviewModal` — `getByText` is substring-matching, and the `apps-offsite-onsite-note` prose also contains "Listing media". Pinned with `exact: true` (the header), not `.first()` (which the note alone would satisfy).
- `OffsiteReviewQueue` — the listing-preview detail pane renders its own content-rating badge, so a bare exact `'g'` matched twice. Scoped to the Content-rating field group and asserted on its full text.

### Group B — tests failing by burning their 5s timeout

| File | Before | After |
|---|---|---|
| `Apps/ListingAssetStep.browser.test.tsx` | 13 failed, **176.48s** | 13 passed, **4.35s** |
| `Apps/ExternalSubmitForm.browser.test.tsx` | 2 failed / 25 passed, 36.18s | 27 passed, 6.50s |
| `Apps/AppEditPage.browser.test.tsx` | 2 failed / 1 passed, 31.46s | 3 passed, 1.82s |
| `Apps/ExternalSubmitForm.reducedMotion.browser.test.tsx` | 1 failed, 16.31s | 1 passed, 1.64s |
| `Apps/ReportTabs.browser.test.tsx` | 1 failed / 9 passed, 1.62s | 10 passed, 1.54s |

Causes, each read off the observed failure output:

- **ListingAssetStep** — the trpc mock omitted `ingestAssetFromDataUri`, which `ListingAssetStep.tsx:177` calls unconditionally → `Uncaught TypeError: Cannot read properties of undefined (reading 'useMutation')` during render. Nothing mounted, so all 13 tests waited out their timeout.
- **ExternalSubmitForm** — (a) `createClientLink` is rendered **twice by design** (persistently below the picker *and* as the Select's `nothingFoundMessage`); for a moderator the global-search Select has no results so both are mounted → strict-mode violation. Scoped to the in-form one. (b) One test called `pickClient()` without first advancing to the App step, so it waited on a control that was never mounted.
- **AppEditPage** — two independent causes. (a) The page renders `<Meta>` → `useBrowserRouter()`, which **throws** `missing context` without a `BrowserRouterProvider` (the harness mounts none), taking the whole render down. (b) The file carried its own `vi.mock('next/router', ...)` which **silently lost** to the scaffold's mock in `test/component-setup.tsx`, so `router.query` was always `{}` and `?tab=media` could never select the media tab. Switched to the established idiom of seeding the scaffold's shared router (see `src/tests/pages/payment/success.browser.test.tsx`). This file has never passed — it and the page landed together in #3393.
- **ExternalSubmitForm.reducedMotion** — `useCurrentUser` unmocked → **throws** `missing CivitaiSessionContext`. Fixing that exposed a second throw underneath: the role-aware picker calls `oauthClient.searchForModerator.useQuery` unconditionally (rules of hooks) and the mock omitted it.
- **ReportTabs** — a sync `page.getByRole('tab').elements()` read racing the async-committed browser-mode render (`expected [] to have a length of 3`). Anchored on an awaited element first.

### Group C — PageBlockHost real-timer sleep

| File | Before | After |
|---|---|---|
| `AppBlocks/PageBlockHost.browser.test.tsx` | 25 passed, **81.14s** | 25 passed, **6.09s** |

This file already passed. Six tests slept through the product's real `BLOCK_READY_TIMEOUT_MS` (10s) and `TOKEN_WAIT_TIMEOUT_MS` (15s) windows — ~76s, 95% of the file. They now install Vitest's fake clock **before** the render (installing it after does nothing — the timer is already on the real clock), jump the window, then hand straight back to real timers, so no assertion, `vi.waitFor` or `userEvent` click ever runs on a faked clock. An `afterEach(() => vi.useRealTimers())` prevents leakage on a mid-test failure. Ran 4× consecutively, stable.

## Totals

**353.8s → 36.9s across the 11 files — ~5.3 min recovered** (~4.0 min of it from Groups A+B alone).
**60 → 115 passing tests: +55 restored, 36 of which were not being collected at all.**

## Verification

Run locally against real Chromium via the Playwright provider. All 10 Group A+B files together: **10 files / 90 tests passed, 10.2s**. `PageBlockHost`: 25 passed, 4 consecutive runs.

Mutation-checked (each broken, confirmed failing, reverted):

- Removed the `...actual` spread from `ConnectScopesPanel` → back to **0 collected** with the exact `trpcVanilla` error.
- Removed the `OffsiteReviewModalBody` stub from `review-queue-nav` → back to **0 collected** with the exact missing-export error.
- Changed the `OffsiteReviewQueue` fixture's declared rating `'g' → 'pg'` → the rewritten assertion fails (`expected 'Content ratingpgdeclared' to be 'Content ratinggdeclared'`), so it is not tautological.
- Renamed `CombinedReviewModal`'s product header `"Listing media" → "Listing MEDIA"` → the `exact: true` query fails, confirming it is not satisfied by the note that shares the substring.
- Widened both `PageBlockHost` product windows to 60s → **exactly the six converted tests fail and no others**, so they still pin the real windows.

Every Group B fix is also mutation-proven by construction: the pre-fix run *is* the failing state, observed in this same worktree.

No test was weakened, skipped or deleted. `tsc --noEmit` reports 0 errors in any `.browser.test.tsx` file (all 100 are in the program); the tree-wide errors in this sandbox are a local `prisma generate` artifact and are present on `main` too.

## Follow-up (not in this PR)

Several other browser tests still wholesale-`vi.mock('~/utils/trpc')` without the `importOriginal` spread (e.g. `ExternalSubmitForm.browser.test.tsx`, `ExternalSubmitForm.reducedMotion.browser.test.tsx`, `AppEditPage.browser.test.tsx`). They pass today but carry the same latent break — the next trpc export added will silently zero them out. A lint rule requiring the spread on any wholesale `~/utils/trpc` mock would close the class deterministically.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import { describe, expect, test, vi, beforeEach } from 'vitest';
+import { describe, expect, test, vi, beforeEach, afterEach } from 'vitest';
 import { page } from 'vitest/browser';
 import { useDialogStore } from '~/components/Dialog/dialogStore';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
@@ -160,6 +160,37 @@ const baseProps = {
   viewer: { id: 42, username: 'tester' },
   theme: 'light' as const,
 };
+
+/**
+ * Six tests below exercise PageBlockHost's two REAL product timeout windows —
+ * `BLOCK_READY_TIMEOUT_MS` (10s) and `TOKEN_WAIT_TIMEOUT_MS` (15s) in PageBlockHost.tsx.
+ * Sleeping through them in real time cost ~76s, ~95% of this file's runtime. Instead:
+ *
+ *   useFakeClock()  — install Vitest's fake clock BEFORE the render, so the product's
+ *                     `setTimeout` is scheduled on it (installing it after the render
+ *                     does nothing: the timer is already on the real clock).
+ *                     `shouldAdvanceTime` keeps the fake clock ticking with real time so
+ *                     browser-mode locator polling still progresses while it's installed.
+ *   advancePastWindow(ms) — jump the window, then hand control straight back to REAL
+ *                     timers, so every later assertion, `vi.waitFor` and (crucially)
+ *                     `userEvent`/locator click runs on the normal clock. No test
+ *                     interacts with the DOM while the fake clock is installed.
+ *
+ * These do NOT weaken the tests: each still asserts the same terminal DOM state, and
+ * the elapsed window is now explicit (advance 11s past a 10s window) rather than implied
+ * by a generous real-time poll. The `afterEach` below is a safety net so a mid-test
+ * failure can never leak the fake clock into the next test.
+ */
+function useFakeClock() {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+}
+async function advancePastWindow(ms: number) {
+  await vi.advanceTimersByTimeAsync(ms);
+  vi.useRealTimers();
+}
+afterEach(() => {
+  vi.useRealTimers();
+});
 
 // Drive the handshake to BLOCK_READY (status='ready') so the consent gate's
 // `status === 'ready'` precondition is satisfied — same prerequisite a real
@@ -398,17 +429,19 @@ describe('PageBlockHost loading indicator (Task 1)', () => {
       // NEVER ack BLOCK_READY → after BLOCK_READY_TIMEOUT_MS (10s) onReadyTimeout
       // flips status 'loading' → 'timeout', clearing the loader and rendering the
       // fallback.
+      useFakeClock();
       renderWithProviders(<PageBlockHost {...baseProps} onConsentGranted={vi.fn()} />);
 
       await expect.element(page.getByTestId('app-page-loading')).toBeInTheDocument();
 
-      // Wait out the real 10s readiness window (poll with generous timeout). The
-      // loader must clear and the timeout fallback render.
+      // Jump the 10s readiness window on the fake clock instead of sleeping through it.
+      await advancePastWindow(11_000);
+      // The loader must clear and the timeout fallback render.
       await vi.waitFor(
         () => {
           expect(page.getByTestId('app-page-loading').query()).toBeNull();
         },
-        { timeout: 13_000, interval: 250 }
+        { timeout: 5_000, interval: 100 }
       );
       await expect.element(page.getByTestId('app-page-fallback')).toBeInTheDocument();
     },
@@ -423,17 +456,19 @@ describe('PageBlockHost loading indicator (Task 1)', () => {
       // TOKEN_WAIT_TIMEOUT_MS (15s) timer flips status 'loading' → 'no_token',
       // clearing the loader and rendering the fallback. (With a null token the
       // iframe still mounts in the loading state, so the overlay is shown first.)
+      useFakeClock();
       renderWithProviders(
         <PageBlockHost {...baseProps} token={null} tokenError={false} onConsentGranted={vi.fn()} />
       );
 
       await expect.element(page.getByTestId('app-page-loading')).toBeInTheDocument();
 
+      await advancePastWindow(16_000);
       await vi.waitFor(
         () => {
           expect(page.getByTestId('app-page-loading').query()).toBeNull();
         },
-        { timeout: 18_000, interval: 250 }
+        { timeout: 5_000, interval: 100 }
       );
       await expect.element(page.getByTestId('app-page-fallback')).toBeInTheDocument();
     },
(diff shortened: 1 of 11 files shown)


ZacxDev:
## Adversarial audit — verdict: safe to merge

Test-only, no product source touched, nothing skipped or deleted, and every changed `waitFor` bound got **tighter** (13s/18s → 5s), not looser.

Both mutations were re-run independently rather than taken on trust:

- Product windows `PageBlockHost.tsx:131-132` → `60_000` fails **exactly the six converted tests, by name** (6 failed / 19 passed). The fake-timer conversion still pins the real windows.
- Dropping `...actual` from the `ConnectScopesPanel` factory → **0 collected**, with the exact `does not provide an export named 'trpcVanilla'` error.
- Fake-timer containment tested by injecting a `throw` while the clock was faked: only that test failed, 24 others passed. The `afterEach(() => vi.useRealTimers())` net holds.

The "two independent causes" claim for `AppEditPage` was challenged and then confirmed by experiment — main's router mock *plus* the `Meta` fix still fails `?tab=media`, so the shared-router diagnosis is right.

### Correction to one audit finding

The audit asserted the component suite "is not run by CI at all," having looked only at `.github/workflows/lint.yml`. **That's wrong — it missed Tekton.** The `npm-component-tests` Task exists and runs `pnpm run test:component`, 20 `component-tests` TaskRuns exist, and the pipeline emits a `preview / component-tests` commit status. The PR body's framing is correct as written.

The practical point survives for a different reason: that task has **never once completed** (0 successes across every retained run), so the suite yields no signal today. talos-infra #763 bounds it so it stops killing the whole PipelineRun; these eleven files are the other half. Once the suite finishes, the gate becomes real.

### Follow-ups — not blocking this PR

**1. `ExternalSubmitForm` — the relaxed query silences a genuine product defect.** `ExternalSubmitForm.tsx:364-378` defines one JSX const rendered at both `:536` (`nothingFoundMessage`) and `:570`, with the **same `data-testid`** and the same `<a href="/user/account">`. Mantine's `keepMounted: true` default means the dropdown copy is always in the DOM, inside `role="listbox"` — where an anchor is not a permitted child. When a moderator's search returns nothing (the exact state that message exists for), **both identical links are visible at once**. The strict-mode violation was a true signal; scoping the query to the form hides it permanently, and nothing now asserts on the dropdown copy. Worth fixing in the product — distinct testid, or make the dropdown copy non-interactive prose.

**2. The latent mock class is still open.** Several browser tests still wholesale-mock `~/utils/trpc` without the `importOriginal` spread (`ExternalSubmitForm`, `reducedMotion`, `AppEditPage`). They pass today but carry the identical break — the next tRPC export added silently zeroes them out, and the 0-collected signature is easy to mistake for a pass. A lint rule closes the class deterministically.

**3. `OffsiteReviewQueue.browser.test.tsx:329`** — the replacement `parentElement?.textContent` assertion is brittle (depends on Mantine `Group` adding no wrapper, on JSX whitespace elision, and on no fourth sibling joining) and is a sync `.element()` read, dropping the auto-retry the rest of the file uses. Precedent for the deterministic fix is three lines below in the product: give the badge a testid, as `apps-offsite-derived-rating` does at `OffsiteReviewQueue.tsx:600`.

**4. `AppEditPage.browser.test.tsx:80`** — no router-spy reset. `router.push` history accumulates across its three tests; latent today, wrong the moment someone asserts on `push`. Cross-file blast radius is zero, but only because `browser.isolate` defaults to `true` — undocumented and load-bearing.

### Environmental caveat

Verification ran against a Nix system Chromium 150; CI uses Playwright's bundled build (≈Chromium 141). Anything binary-sensitive here — fake timers, animations, `prefers-reduced-motion`, focus — is untested outside this box.

### Found en route, out of scope

The #3412 review modal renders `AppListingCard` (`OffsiteReviewQueue.tsx:913-915`) with no preview posture: duplicate `CreatorChip` anchors plus a live CTA and detail link for an **unapproved** listing, inside the moderator review modal. Pre-existing, worth its own issue.

PR #3493, opened 2026-07-31
feat(apps): retire the legacy /apps/[appBlockId] detail page — redirect to the store detail

ZacxDev:
## What

Retires the legacy per-app detail route `/apps/<appBlockId>` by redirecting it to the unified store detail `/apps/store-preview/<slug>` from `getServerSideProps`.

That route was a **second, diverging detail surface for the same app**, and it carried two measured bugs the store detail does not have:

- it rendered `by <app name>` where the store card/detail correctly render the **owner's username**;
- its "Live preview" was a raw, bridge-less `<iframe>`. Nothing ever posts `BLOCK_INIT` into that frame, so the block's design-system shell never learns the host theme and renders its **default light palette** — a permanent near-white panel on a dark page. (Not a browser paint artifact; the frame's shell element resolves `--civitai-color-body` under `data-theme="light"`.)

## Behaviour

| Case | Result |
|---|---|
| App **with** an approved store listing | `302` → `/apps/store-preview/<slug>` |
| App with **no** approved listing (pending / rejected / never-approved) | site-standard **404** (`notFound`) |
| Viewer the store-visibility flag does not grant | `notFound`, exactly as today — the gate runs **first**, before the route param is read and before any DB access |

**Why 404 and not a bounce to `/apps` for the miss.** On-site listings are created at **approval**, so a pending/rejected app has no store target. The person most likely to open this URL is the app's **owner** following a link to their own pending app; sending them to a store that by construction does not list unapproved apps is a silent dead end — it reads as "my app vanished", with no signal why. A 404 is the honest answer, it matches the posture `blocks.getAppDetail` already has (`NOT_FOUND` for a missing/unapproved app) and what `/apps/store-preview/<bogus>` already renders, and the owner's real surface is `/apps/my-submissions` with its status sections.

`permanent: false` (302, not 301) on purpose: a 301 is cached by browsers indefinitely, which would make the route's still-open final disposition irreversible in the field.

## `AppBlock.blockId` vs `AppListing.slug` — the finding

The scope left this open. **Both.** The two hold the same value today, but the redirect keys on `AppListing.slug`, and that is load-bearing:

- `mapAppBlockToListing` (`src/server/services/blocks/app-listing-mapper.ts`) mints an on-site listing with `slug: ab.blockId`, and the `AppListing.slug` column comment says the same. Verified against live data: **20/20** listings that back an `AppBlock` have `slug == block_id`. So reading `blockId` would usually produce the same string.
- But it would answer the **wrong question**. The destination resolves **by listing, approved-only**. A pending/rejected app has a `blockId` and no listing, so a `blockId`-keyed redirect would bounce its owner to a store URL that 404s instead of 404-ing here — losing the decided miss behaviour entirely. Keying on the listing row makes "an approved listing exists" the actual precondition of the redirect, and it stays correct if a slug ever diverges from its block id (off-site listings already choose their own slugs).

Since the route param is `AppBlock.id` (`apb_<ULID>`), not the slug, a lookup is required either way — so keying on the listing costs nothing.

## Is anything live going to start 404-ing?

No. Checked against live data:

- **0** approved apps lack an approved listing (9 approved apps ↔ 13 approved listings, all covered). The `notFound` branch cannot fire for a currently-visible app.
- The only apps that now 404 are non-approved ones — which **already** rendered `<NotFound />` at this URL, because `blocks.getAppDetail` is approved-only. Same user-visible outcome, decided server-side instead of after a client round-trip.

⚠️ **That "0" is a fact about today, not a guarantee.** Auto-create-on-approve is best-effort — it logs and continues on failure, and is skipped when no owner resolves — and apps approved before it shipped only get a listing on a later re-approve. Recovery is the mod-only listing backfill; nothing self-heals. So an approved, serving app *can* end up without a listing row and would then 404 here where it previously rendered. The retargeted e2e leg below is what makes that state visible instead of silent: it asserts the redirect against a runtime-discovered approved app and names a 404 as a data signal, not flake.

## Deliberately not in this PR

**Redirect, not delete.** The page body stays, and the inbound links are untouched — they simply hop. **Four files, five link sites** (an earlier draft of this said "four callsites", which conflated the two): `AppBlockCard` links **twice** (title ~`:204` and description ~`:257`), plus `ManifestEditForm` (post-save Cancel), `liveAppDetailHref` in `appListingDetailView.ts`, and the editor's own **Back** (`[appBlockId]/edit.tsx`).

⚠️ **The hop drops the query string.** The destination is built from the slug alone. Harmless in-product — the legacy page only ever read `router.query.appBlockId`, a route param that survives in the destination path — but an external link carrying `?utm_*` / `?ref=` loses those at the hop. Named as a decision, not a surprise; forwarding them is a one-line change if anyone wants it. That keeps a stale bookmark or external link working, and keeps this a one-file change. Deleting the body and retargeting those hops is a tracked follow-up, sequenced after the in-flight work on `appListingDetailView.ts` so it cannot collide with it.

**Two consequences of leaving them alone — found in review, named rather than discovered later:**

1. 🔴 **The store detail's own CTA for a model-slot app now hops back to itself.** `getDetailPrimaryAction`'s `info` branch (an on-site app with no launch page) sets `href: liveAppDetailHref(appBlockId)` — so on `/apps/store-preview/<slug>` the primary button "Runs on model pages" goes to `/apps/<appBlockId>`, which now redirects back to the page the viewer is already on. It is a click-driven no-op, not a machine loop, and **no live app is in that state** (all 9 approved on-site listings declare a page, so they take the `open`/`visit` branch). But it is latent for the whole `model.sidebar_top` product line, and fixing it means editing `appListingDetailView.ts` — the exact file the follow-up is sequenced around. **Flagging rather than fixing: pulling that edit in here would re-create the same-file collision this PR's scoping exists to avoid.** The follow-up should *retarget* that branch, not merely delete the route.
2. An owner backing out of the editor on a **pending** app now lands on a real 404 instead of a not-found rendered inside the app shell. Same end state, blunter presentation; the follow-up that retargets those links fixes it.

Checked, not assumed: *Open app* → the store detail's `open` branch, *Open live* → its `visit` fallback, *Edit manifest* → the **sibling** route `/apps/[appBlockId]/edit-manifest`, which is untouched here and independently reachable from `/apps/my-submissions` and the store card's owner Edit. Sibling routes `/edit`, `/edit-manifest`, `/listing`, `/revenue` are unaffected (distinct route paths).

🔴 **One affordance is genuinely NOT carried over, and my first draft of this PR wrongly claimed otherwise: the Install / Manage CTA.** `AppListingDetailBody` has no install surface at all, so a model-slot app would have nowhere on the store detail to install from. Same population as the self-loop above — **0 apps today** — but it is the substantive gap to close before a model-slot app is approved, and together with (1) it is why the follow-up must *retarget* that branch rather than simply delete the route.

🔴 **One knock-on, and my earlier statement of it was wrong.** The legacy 5-star `AppBlockReview` write form lived on this page. I claimed it "stays reachable through `AppDetailsModal` (opened from `AppBlockCard`), so nothing is orphaned". **It does not, and something is.** `AppBlockCard` renders only from `MarketplaceBody` and `RecentlyOpenedApps` — and `RecentlyOpenedAppsView` itself renders only inside `MarketplaceBody`, which has had **no importer in app code** since `/apps` swapped to `AppListingsMarketplaceBody`; it is retained purely as a documented one-line rollback. So after this PR the whole 5-star review surface has **no reachable entry point**.

Harmless today — `app_block_reviews` is **0 rows** in production, and the store detail carries the listing review surface — but the claim was wrong and a reviewer could have relied on it. The consequence is for the follow-up that deletes this body: it must decide the legacy review form's fate **explicitly**, rather than inherit a "still reachable" assumption that is false.

## Tests

New pure module `src/components/Apps/resolveLegacyAppRedirect.ts` — I/O-free and React-free so the branch is assertable in the node `unit` project without booting the page's module graph (same extraction precedent as `resolveAppsPageAccess.ts`, `deploy-status.ts`, `sortInstallsForSlot.ts`).

The review's sharpest finding was that the **security-relevant half of the change was the untested half**: a test of the string-builder alone cannot see the gate-before-query ordering, and deleting `status: 'approved'` from the query inverted the decided behaviour while every test still passed. Both are now covered, so the module exports three things:

- `resolveLegacyAppRedirect({ slug })` — the redirect-vs-404 decision;
- `approvedListingSlugQuery(appBlockId)` — the Prisma args, built here so the approved-only precondition is pinned;
- `resolveLegacyAppRoute({ features, appBlockId, findApprovedListingSlug })` — the whole SSR decision **with the DB read injected**, so a test can assert the lookup was never even *called* for a viewer without store visibility.

`src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts` — **24 tests**: the redirect happy path, the `notFound` miss by name, null / undefined / absent / empty / whitespace / non-string slugs, slug encoding, one-path-segment containment, gate-before-query ordering, route-param trimming, the either-flag OR-fallback, route-param handling, and the query shape.

**These are a real gate, not a change-detector — verified by mutation.** Each mutation was applied to the source, the suite run, and the failure confirmed before reverting:

| Mutation | Result |
|---|---|
| `notFound` branch → `redirect` to `/apps` (the rejected alternative) | **6 tests fail** |
| drop `encodeURIComponent` from the destination | **4 tests fail** |
| `permanent: false` → `true` | **5 tests fail** |
| move the store-visibility gate **after** the lookup | **2 tests fail** |
| drop `.trim()` on the slug | **3 tests fail** |
| drop `status: 'approved'` from the query | **1 test fails** |
| pass the **untrimmed** `appBlockId` to the lookup while still guarding the trimmed value | **1 test fails** |
| rewrite a dot-only slug in the destination | **1 test fails** |

Reverted; 24/24 green, and the full `src/components/Apps/__tests__` directory is 511/511.

🔴 **Three test defects of my own were found and fixed after the table above was first written — two of them assertions that could not fail, added by an earlier audit pass looking for exactly that:**

- The **untrimmed-`appBlockId`** mutation (last row) previously **survived the whole suite**. The resolver trims the route param, guards the trimmed value, then queries — but every fixture id in the file was whitespace-free and the only `toHaveBeenCalledWith` used a clean `'apb_x'`, so passing the untrimmed string to the lookup changed nothing observable. In production that sends `'  apb_x  '` to an equality filter on a ULID column: an approved app 404s instead of redirecting. A granted-viewer case with `'  apb_x  '` now catches it.
- Two `expect(...includes('/')).toBe(false)` sub-assertions each **followed a literal-equality assertion over the same string in the same `it`** — so any mutation admitting a `/` broke the literal first and the sub-assertion could never be the thing that failed. **Removed**, and the containment property now lives in its own `it` over separator-bearing inputs that no literal in the file constrains (that `it` is one of the four killed by dropping `encodeURIComponent`).
- The traversal test's claim was **overstated**: `encodeURIComponent('..') === '..'`, so a slug of exactly `..` yields `/apps/store-preview/..`, which dot-segment removal resolves to `/apps/` — it does *not* stay under the prefix. It never leaves the origin, and `SLUG_REGEX` (`/^[a-z][a-z0-9-]*[a-z0-9]$/`) admits no `.` so it is unreachable, but the test pinned only the slash-bearing `'../../login'` while the docstring claimed more than the code holds. The claim is now narrowed and the dot-only case pinned explicitly.

**`tests/preview-apps-marketplace.spec.ts` retargeted.** Its last leg navigated to `/apps/<id>` and asserted a heading — on the page this PR retires. Left alone it would have asserted against the store detail by accident, while racing that page's client-side query behind a loader. It now asserts the **retirement** (that the route lands under `/apps/store-preview/`), pinned on the path prefix rather than on `<blockId>` so it does not re-import the slug assumption the redirect deliberately avoids.

🔴 **A follow-up assertion in that leg could not fail and has been replaced.** It asserted `landedOn` matched `/^\/apps\/store-preview\/.+/` and then that it `!== '/apps/<id>'` — given the first passed, the second held by construction. It is replaced by two that can fail: (a) the store slug occupies **exactly one path segment** (the prefix regex happily accepts `/apps/store-preview/a/b`), and (b) the response's **redirect chain contains the legacy path**, which pins that the retirement happens at the HTTP layer and fails if anyone reimplements it as a client-side bounce or a rewrite. ⚠️ That suite is report-only and runs against a preview deployment, so these two were reasoned rather than executed; they typecheck (`tests/` is inside `tsconfig`'s `include`).

No component/browser tests were added — there is no component left to render on this route.

`NODE_OPTIONS=--max_old_space_size=8192 npx tsc --noEmit`: **26 errors on this branch, 26 on a pristine baseline worktree checked out at this branch's merge-base — and the two error sets are byte-identical** (measured by diffing the sorted `error TS` lines, not by comparing counts). All pre-existing, none in the touched files.

## Security review of the diff

- **Open redirect:** the slug is `encodeURIComponent`-ed, so `//host`, `../`, `?` and `#` in a slug stay path segments under `/apps/store-preview/` rather than steering the destination off-origin or splicing a query onto it. Three tests pin this, and dropping the encode fails them. The value comes from our own column, so this is defence in depth.
- **Existence oracle:** the store-visibility gate runs **before** the route param is read and before the DB query, so a viewer without store access gets an identical `notFound` to today — no `Location` header, no query, and now a test that fails if anyone reorders it.
  - **Stated precisely, because the first draft of this claim was too broad:** gate-first protects viewers *without* store visibility. For a viewer *with* it, 302-vs-404 is a **new HTTP-level signal that the old page did not emit** (it answered 200 either way), and it is not filtered by the further gates the destination applies.
  - 🔴 **There are THREE such gates, not two — an earlier draft of this section named only two, and its proposed remediation was therefore a two-thirds fix.** `getListingDetail` rejects a row on any of:
     1. **store-scope kind** — `scope === 'public-external' && row.kind !== 'offsite'`;
     2. **deploy** — `row.kind === 'onsite' && appBlock.currentVersionDeployedAt == null`;
     3. **maturity** — `!redCapable && isMatureContentRating(row.contentRating)`.

     (2) and (3) are the two I originally named, and both are **0 live instances** (no approved listing is rated `r`/`x`; no approved on-site listing's app is undeployed). **(1) is the one I missed, and it is categorically different.** This route's param is an `AppBlock.id`, and **every** listing carrying an `appBlockId` is `kind='onsite'` — 20/20 across all statuses in live data. So every listing this route can resolve is *precisely* the set `public-external` exists to hide. Under that scope the coverage gap is **100% of the route's resolvable set**, not a corner.

     It is **unreachable today**, and only by an accident of flag alignment: the page gate grants on `features.appListings || features.appBlocks`, and `features.appListings` reads the same `app-listings` Flipt key that `resolveStoreVisibilityScope`'s first axis uses to return `full`. Anyone past the page gate resolves `full`, where the kind gate is a no-op; anyone who would resolve `public-external` is already 404'd by the page gate. **If `app-listings-public-external` is switched on and the page gate is widened alongside it, that alignment breaks** and every `/apps/<appBlockId>` becomes a 302 disclosing an on-site app's slug to a viewer whose store scope excludes on-site entirely.

     Disclosed residual, not a hole today. **Remediation, restated correctly: all three** — a store-scope check (which for this route reduces to "`public-external` → `notFound`", the resolvable set being wholly on-site), a `currentVersionDeployedAt` filter, and a `contentRating` filter keyed on the request's red-capability. **Not implemented in this PR and still owed** — it needs the resolved store scope and the red-capability threaded into an SSR resolver that takes neither today, which is more than this redirect-only change should carry. Doing it before the store widens past its current audience is the fix.
  - Also unchecked on both hops now: `AppBlock.status`. The old route 404'd a suspended app; the listing-keyed path does not look at it. **0 live instances** (no approved listing is attached to a non-approved app block), but the drift is reachable, since the block-status flip on relist is non-fatal.
- **Owner vs anonymous vs moderator:** no divergence introduced. The redirect decision reads no session and no ownership — every viewer past the flag gate gets the same answer for the same app. Owners are the group the `notFound` branch was chosen *for* (see above); anonymous viewers are unchanged (still gated out); moderators get the hop.
- **A manifest edit re-enters review, but does not strand the redirect:** the lookup keys on `AppListing.status`, which is independent of `AppBlock.status`, so an app whose *code* is back in review keeps resolving to its still-approved listing. (Live data: no approved listing is attached to a non-approved app, and the `app_block_id` column is UNIQUE, so at most one row can ever match.)

## Two smaller notes, named rather than left to be found

- **One DB read is added to this route.** It is a single lookup on a UNIQUE column, on a route that is flag-gated to a small audience and now renders nothing — negligible.
- **New failure mode, judged acceptable:** if that read throws, `getServerSideProps` throws and the route 500s, where previously the page rendered and settled into a client-side not-found. Deliberately not wrapped in a `try/catch` fallback: swallowing it into a 404 would disguise an outage as "this app doesn't exist", and every other SSR route on the site fails the same way under the same condition.

## Review

Two independent adversarial passes over the diff (component suites are report-only here, so the audit is the real correctness gate). Everything above that is marked as a residual, a latent issue, or a corrected claim came out of them; the fixes are in commits 2–4. Between them they corrected three things I had stated wrongly — the callsite count (three → four), "the store detail covers the whole action set" (Install is not covered), and an over-broad existence-oracle claim — and found two assertions that could not fail.

Also confirmed clean by both, with evidence:

- **No redirect loop.** `/apps/store-preview` is a static segment that outranks the dynamic one; the detail route has no server-side redirect of its own; no `/apps` rules in middleware or `next.config`.
- **Open redirect contained.** No slug can produce `//host`, a `?`/`#` splice, or CR/LF header injection; `%2F` is not decoded before routing, so an encoded traversal stays one path segment.
- **No client-bundle or vitest-browser leakage** from the `dbRead` import — no `*.browser.test.tsx` imports this page, and `dbRead` in a page is the established pattern (29 other pages do it).
- **At most one listing row can ever match** — `app_block_id` is UNIQUE, so `findFirst` is deterministic. The route param is `AppBlock.id`, matching what the old page queried, so there is no id/blockId mismatch.
- **No owner / anon / moderator regression.** The old page was already approved-only for everyone including mods, so an owner or mod on a pending app saw a not-found before and gets a 404 now.

One gate-parity item is left open on purpose and is the sharpest thing a reviewer should push on: the redirect does not replicate the **three** gates the destination applies — store-scope kind, deploy, and maturity (see the security section, and note that an earlier draft of this line named only the last two). The deploy and maturity gates are empty today; the **kind** gate covers 100% of this route's resolvable set under a `public-external` scope and is held back only by the page gate happening to key on the same flag. The fix is all three filters, and it is a reasonable ask before the store widens.

## Verification owed after merge

- `GET /apps/<approved app id>` → `302`, `Location: /apps/store-preview/<slug>` (assert on status + header, not a rendered page).
- An app with no approved listing → site-standard 404, no redirect loop.
- `/apps/<id>/edit`, `/edit-manifest`, `/listing`, `/revenue` all still load.

### DIFF
--- a/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
+++ b/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
@@ -1,0 +1,123 @@
+import { describe, it, expect } from 'vitest';
+import { resolveLegacyAppRedirect, STORE_PREVIEW_PATH_PREFIX } from '../resolveLegacyAppRedirect';
+
+/**
+ * S8 / PR-2 — the retirement decision for the legacy `/apps/[appBlockId]` detail
+ * route. These are a REAL gate, not a change-detector: the whole PR is two
+ * branches, and the second one (no approved listing → `notFound`) is a product
+ * decision that a future "helpful" edit would otherwise silently invert into a
+ * redirect to `/apps` with nothing failing.
+ *
+ * Break-it checks these are written to survive:
+ *   - turn the no-listing branch into `{ redirect: { destination: '/apps' } }`
+ *     → the `notFound` cases below must FAIL.
+ *   - drop `encodeURIComponent` → the escaping case must FAIL.
+ *   - flip `permanent` to true → the 302 case must FAIL.
+ */
+
+describe('resolveLegacyAppRedirect — approved listing → store-preview redirect', () => {
+  it('redirects to /apps/store-preview/<slug> as a temporary (302) redirect', () => {
+    expect(resolveLegacyAppRedirect({ slug: 'model-benchmarking' })).toEqual({
+      redirect: {
+        destination: '/apps/store-preview/model-benchmarking',
+        permanent: false,
+      },
+    });
+  });
+
+  it('is never a permanent redirect (the route retirement is still reversible)', () => {
+    const result = resolveLegacyAppRedirect({ slug: 'vitrine' });
+    expect(result).toHaveProperty('redirect');
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.permanent).toBe(false);
+    expect(result).not.toHaveProperty('notFound');
+  });
+
+  it('always lands under the store-preview path prefix', () => {
+    const result = resolveLegacyAppRedirect({ slug: 'some-app' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination.startsWith(STORE_PREVIEW_PATH_PREFIX)).toBe(true);
+  });
+});
+
+describe('resolveLegacyAppRedirect — NO approved listing → site-standard 404', () => {
+  // 🔴 The decided miss behaviour. Listings are created at APPROVAL, so a
+  // pending / rejected / never-approved app has no slug to send anyone to. This
+  // must be `notFound` — NOT a redirect to `/apps`, which silently dead-ends the
+  // owner looking for their own pending app.
+  it('slug null (no listing row) → notFound, never a redirect', () => {
+    const result = resolveLegacyAppRedirect({ slug: null });
+    expect(result).toEqual({ notFound: true });
+    expect(result).not.toHaveProperty('redirect');
+  });
+
+  it('slug undefined / key absent → notFound, never a redirect', () => {
+    expect(resolveLegacyAppRedirect({ slug: undefined })).toEqual({ notFound: true });
+    expect(resolveLegacyAppRedirect({})).toEqual({ notFound: true });
+    const result = resolveLegacyAppRedirect({});
+    expect(result).not.toHaveProperty('redirect');
+  });
+
+  it('empty / whitespace-only slug → notFound (not a redirect to the bare prefix)', () => {
+    expect(resolveLegacyAppRedirect({ slug: '' })).toEqual({ notFound: true });
+    expect(resolveLegacyAppRedirect({ slug: '   ' })).toEqual({ notFound: true });
+    expect(resolveLegacyAppRedirect({ slug: '\t\n' })).toEqual({ notFound: true });
+  });
+
+  it('a non-string slug (defensive: untyped/JSON input) → notFound', () => {
+    // Guards the runtime edge a `string | null` type does not: a DB/serialization
+    // surprise must fail closed to 404, never build `/apps/store-preview/[object Object]`.
+    expect(resolveLegacyAppRedirect({ slug: 123 as unknown as string })).toEqual({
+      notFound: true,
+    });
+    expect(resolveLegacyAppRedirect({ slug: {} as unknown as string })).toEqual({
+      notFound: true,
+    });
+  });
+
+  it('never returns a redirect to /apps for a missing listing', () => {
+    // Named guard for the rejected alternative, so an edit back to it fails here
+    // with an obvious message rather than only tripping the toEqual above.
+    for (const slug of [null, undefined, '', '  ']) {
+      const result = resolveLegacyAppRedirect({ slug });
+      const destination = 'redirect' in result ? result.redirect.destination : null;
+      expect(destination).not.toBe('/apps');
+      expect(destination).toBeNull();
+    }
+  });
+});
+
+describe('resolveLegacyAppRedirect — slug encoding / open-redirect containment', () => {
+  it('percent-encodes URL-unsafe characters in the slug', () => {
+    const result = resolveLegacyAppRedirect({ slug: 'a b/c?d#e' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination).toBe('/apps/store-preview/a%20b%2Fc%3Fd%23e');
+  });
+
+  it('a protocol-relative slug cannot escape to another origin', () => {
+    // `//evil.example` un-encoded would be a protocol-relative URL — an open
+    // redirect off civitai.com. Encoded, it stays a path segment.
+    const result = resolveLegacyAppRedirect({ slug: '//evil.example' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination).toBe('/apps/store-preview/%2F%2Fevil.example');
+    expect(result.redirect.destination.startsWith(STORE_PREVIEW_PATH_PREFIX)).toBe(true);
+  });
+
+  it('path traversal in a slug cannot climb out of /apps/store-preview/', () => {
+    const result = resolveLegacyAppRedirect({ slug: '../../login' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination).toBe('/apps/store-preview/..%2F..%2Flogin');
+  });
+
+  it('an ordinary hyphenated slug is passed through unchanged (no over-escaping)', () => {
+    const result = resolveLegacyAppRedirect({ slug: 'df-qwen-canvas' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination).toBe('/apps/store-preview/df-qwen-canvas');
+  });
+
+  it('trims surrounding whitespace rather than encoding it into the path', () => {
+    const result = resolveLegacyAppRedirect({ slug: '  vitrine  ' });
+    if (!('redirect' in result)) throw new Error('expected a redirect');
+    expect(result.redirect.destination).toBe('/apps/store-preview/vitrine');
+  });
+});
--- a/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
+++ b/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
@@ -13,11 +13,15 @@ import {
  * decision that a future "helpful" edit would otherwise silently invert into a
  * redirect to `/apps` with nothing failing.
  *
- * Break-it checks these are written to survive:
+ * Break-it checks these are written to survive — each was run against the source
+ * and confirmed to fail before being reverted:
  *   - turn the no-listing branch into `{ redirect: { destination: '/apps' } }`
- *     → the `notFound` cases below must FAIL.
- *   - drop `encodeURIComponent` → the escaping case must FAIL.
- *   - flip `permanent` to true → the 302 case must FAIL.
+ *     → 5 FAIL (the `notFound` cases).
+ *   - drop `encodeURIComponent` → 3 FAIL (the encoding/containment cases).
+ *   - flip `permanent` to true → 2 FAIL.
+ *   - move the store-visibility gate AFTER the lookup → 2 FAIL (the ordering
+ *     cases in `resolveLegacyAppRoute`).
+ *   - drop `status: 'approved'` from the query → 1 FAIL (the query-shape pin).
  */
 
 describe('resolveLegacyAppRedirect — approved listing → store-preview redirect', () => {
@@ -38,10 +42,16 @@ describe('resolveLegacyAppRedirect — approved listing → store-preview redire
     expect(result).not.toHaveProperty('notFound');
   });
 
-  it('always lands under the store-preview path prefix', () => {
+  it('the slug occupies exactly one path segment under the prefix', () => {
+    // NB: asserting `startsWith(STORE_PREVIEW_PATH_PREFIX)` would be a tautology —
+    // the destination is built by concatenating that same exported constant, so it
+    // cannot fail. What CAN fail, and is the property worth pinning, is that the
+    // slug never introduces a second segment.
     const result = resolveLegacyAppRedirect({ slug: 'some-app' });
     if (!('redirect' in result)) throw new Error('expected a redirect');
-    expect(result.redirect.destination.startsWith(STORE_PREVIEW_PATH_PREFIX)).toBe(true);
+    const slugSegment = result.redirect.destination.slice(STORE_PREVIEW_PATH_PREFIX.length);
+    expect(slugSegment).toBe('some-app');
+    expect(slugSegment.includes('/')).toBe(false);
   });
 });
 
@@ -103,7 +113,11 @@ describe('resolveLegacyAppRedirect — slug encoding / open-redirect containment
     const result = resolveLegacyAppRedirect({ slug: '//evil.example' });
     if (!('redirect' in result)) throw new Error('expected a redirect');
     expect(result.redirect.destination).toBe('/apps/store-preview/%2F%2Fevil.example');
-    expect(result.redirect.destination.startsWith(STORE_PREVIEW_PATH_PREFIX)).toBe(true);
+    // The property that actually matters: no slash survives into the slug segment,
+    // so the destination can never resolve as `//host` or climb a directory.
+    expect(result.redirect.destination.slice(STORE_PREVIEW_PATH_PREFIX.length).includes('/')).toBe(
+      false
+    );
   });
 
   it('path traversal in a slug cannot climb out of /apps/store-preview/', () => {
--- a/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
+++ b/src/components/Apps/__tests__/resolveLegacyAppRedirect.test.ts
@@ -1,5 +1,10 @@
-import { describe, it, expect } from 'vitest';
-import { resolveLegacyAppRedirect, STORE_PREVIEW_PATH_PREFIX } from '../resolveLegacyAppRedirect';
+import { describe, it, expect, vi } from 'vitest';
+import {
+  approvedListingSlugQuery,
+  resolveLegacyAppRedirect,
+  resolveLegacyAppRoute,
+  STORE_PREVIEW_PATH_PREFIX,
+} from '../resolveLegacyAppRedirect';
 
 /**
  * S8 / PR-2 — the retirement decision for the legacy `/apps/[appBlockId]` detail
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Correcting three claims I made in the description above

An adversarial pass over this PR came back **safe to merge, no blockers** — it re-ran the live-data checks independently (two counts were *understated*, none overstated) and every extra mutation it invented was already caught. The production logic is unchanged and I'm not redesigning it.

But three things in my description were **wrong or incomplete**, and one of them materially understates what's owed. Correcting them here rather than only editing the body, because a reviewer may already have read and believed them. The body is updated too, and the code comments now carry the corrected version. Pushed in f085c3b.

---

### 1. 🔴 The gate-parity disclosure named **two** unreplicated gates. There are **three** — and the one I missed is the biggest.

I wrote that the redirect "is not filtered by two further gates the destination applies — a maturity gate and an on-site 'never deployed' gate", and proposed the fix as "a `contentRating` filter … plus a `currentVersionDeployedAt` filter". **That remediation is a two-thirds fix.**

`getListingDetail` rejects a row on **any** of:

1. **store-scope kind** — `scope === 'public-external' && row.kind !== 'offsite'`
2. **deploy** — `row.kind === 'onsite' && appBlock.currentVersionDeployedAt == null`
3. **maturity** — `!redCapable && isMatureContentRating(row.contentRating)`

(2) and (3) are the two I named. Both are **0 live instances**, and each would only ever affect a handful of apps.

**(1) is categorically different.** This route's param is an `AppBlock.id`, and **every** listing carrying an `appBlockId` is `kind='onsite'` — 20/20 across all statuses in live data. So every listing this route can resolve is *exactly* the set `public-external` exists to hide from viewers who only get the off-site store. Under that scope the coverage gap is **100% of the route's resolvable set**, not a corner case.

It is **unreachable today**, and only by an accident of flag alignment: the page gate grants on `features.appListings || features.appBlocks`, and `features.appListings` reads the same `app-listings` Flipt key that `resolveStoreVisibilityScope`'s first axis uses to return `full`. So anyone past the page gate resolves `full`, where the kind gate is a no-op — and anyone who *would* resolve `public-external` is already 404'd by the page gate. **If `app-listings-public-external` is switched on and the page gate is widened alongside it, that alignment breaks**, and every `/apps/<appBlockId>` becomes a 302 disclosing an on-site app's slug to a viewer whose store scope excludes on-site entirely.

**Remediation, restated correctly: all three** — a store-scope check (which for this route reduces to "`public-external` → `notFound`", the resolvable set being wholly on-site), a `currentVersionDeployedAt` filter, and a `contentRating` filter keyed on the request's red-capability.

**I did not implement it here, and it remains owed.** It needs the resolved store scope and the red-capability threaded into an SSR resolver that takes neither today — more than a redirect-only, one-file change should carry, and it would pull this PR into files other in-flight work is editing. Doing it before the store widens past its current audience is the fix.

### 2. 🔴 "Nothing is orphaned" is **false**.

I wrote that the legacy 5-star `AppBlockReview` write form "stays reachable through `AppDetailsModal` (opened from `AppBlockCard`), so nothing is orphaned". Tracing it properly:

`AppBlockReviews` (the write form) has exactly two hosts — this page, and `AppDetailsModal`. `AppDetailsModal` is opened only from `AppBlockCard`. `AppBlockCard` renders only from `MarketplaceBody` and `RecentlyOpenedApps` — and `RecentlyOpenedAppsView` itself renders only *inside* `MarketplaceBody`. **`MarketplaceBody` has had no importer in app code** since `/apps` swapped to `AppListingsMarketplaceBody`; it is retained purely as a documented one-line rollback.

So after this PR the whole 5-star review surface has **no reachable entry point**. Harmless today — `app_block_reviews` is **0 rows** in production, and the store detail carries the listing review surface — but the claim was wrong, and the consequence lands on the follow-up that deletes this page body: it must decide the legacy review form's fate **explicitly**, not inherit a "still reachable" assumption that isn't true.

### 3. "Four callsites" is four *files*; plus one thing I never named at all.

- `AppBlockCard.tsx` links to this route **twice** — the card title (~`:204`) and the description (~`:257`). So it's **four files, five link sites**. The count was already corrected once (three → four) and was still conflating the two units.
- **The redirect drops the query string.** The destination is built from the slug alone. Harmless in-product — the legacy page only ever read `router.query.appBlockId`, a route param that survives in the destination path — but an external link carrying `?utm_*` / `?ref=` loses those at the hop. Named as a decision rather than left to be found; forwarding them is a one-line change if anyone wants it.

---

### Also fixed: four assertions that could not fail

Not description errors, but worth naming since two of them were added by an **earlier audit pass looking for exactly this defect**:

- **A real mutation was escaping.** Passing the **untrimmed** `appBlockId` to the listing lookup while still guarding the trimmed value left the suite fully green at 21/21 — every fixture id in the file was whitespace-free and the only `toHaveBeenCalledWith` used a clean `'apb_x'`. In production that sends `'  apb_x  '` to an equality filter on a ULID column, and an approved app 404s instead of redirecting. A `'  apb_x  '` granted-viewer case now catches it.
- Two `expect(…includes('/')).toBe(false)` sub-assertions each **followed a literal-equality assertion over the same string in the same `it`** — so any mutation admitting a `/` broke the literal first and the sub-assertion could never be the thing that failed. Removed; the containment property now has its own `it` over separator-bearing inputs no literal in the file constrains.
- The e2e leg asserted `landedOn` matched `/^\/apps\/store-preview\/.+/` and *then* that it `!== '/apps/<id>'` — given the first passed, the second held by construction. Replaced by two that can fail: the slug occupies **exactly one path segment** (the prefix regex accepts `/apps/store-preview/a/b`), and the response's **redirect chain contains the legacy path**, pinning the retirement to the HTTP layer.
- One claim was simply overstated: `encodeURIComponent('..') === '..'`, so a slug of exactly `..` yields `/apps/store-preview/..`, which dot-segment removal resolves to `/apps/` — it does **not** stay under the prefix. Never off-origin, and unreachable given `SLUG_REGEX`, but the test pinned only the slash-bearing `'../../login'` while the docstring claimed more than the code holds. Claim narrowed, dot-only case pinned.

**Verification.** 24/24 on the resolver suite, 511/511 across `src/components/Apps/__tests__`. Every guard — new and pre-existing — was re-confirmed by a mutation that makes it fail: 8 mutations applied to the source, the named failures observed, source reverted and diffed clean. `tsc --noEmit`: **26 errors on this branch, 26 on a pristine baseline worktree at this branch's merge-base, and the two error sets are byte-identical** (diffed line-by-line, not compared by count). The preview e2e suite is report-only and needs a deployment, so the two replacement assertions there were reasoned rather than executed; they typecheck.

PR #3495, opened 2026-07-31
fix(app-blocks): review preview tells the moderator why a consent request did nothing

ZacxDev:
## Symptom

A moderator reviewing a pending app at `/apps/review/preview/<publishRequestId>` clicks a button in the app that calls the SDK's `useRequestConsent()`. **Nothing happens** — no modal, no toast, no error — and the app parks forever on its own "grant permission to continue" card. From the reviewer's seat the submission looks broken, which is the wrong conclusion: the preview is *deliberately* withholding the permission.

## Root cause

`src/components/AppBlocks/PageBlockHost.tsx:714-719` (before this change) — the `REQUEST_CONSENT` handler dropped review-mode requests with a bare early `return`:

```ts
const off = onMessage<{ scopes?: unknown } | undefined>('REQUEST_CONSENT', (payload) => {
  // reviewMode: a consent grant re-mints the token with WIDER scopes — never
  // let untrusted review code pop a permission modal at the mod. Fire-and-
  // forget ⇒ dropping it never hangs the block.
  if (reviewMode) return;
```

That `return` sits ~40 lines **above** the existing "Permission unavailable" `showNotification(...)` fallback (`PageBlockHost.tsx:759-763`), so review mode skipped the entire feedback path — not just the modal. The comment's intent (no permission modal at the mod) was right; the blast radius was too wide.

## Why this is feedback-only, not a permissions change

Nothing about what the preview can do changes:

- **No scope allowlist is touched.** The review mint still strips the app's declared scopes down to the read-only review set. That is correct policy — a mod previewing someone else's app must not leak their own balance/storage — and it stays exactly as-is.
- **Review mode still never opens a consent/permission modal.** A grant would re-mint the mod's token with wider scopes at the request of unapproved code. A passive, non-interactive notification is the ceiling; there is nothing to click and no scope is granted.
- **`REQUEST_CONSENT` stays fire-and-forget** — no requestId, no reply, no timeout (see `hostHandlerParity.ts`, `{ request: false, reply: '' }`). Dropping the *grant* still never hangs the app.

The escape hatch for actually exercising the app's real permissions already exists and is unchanged: the **"Run for real…"** opt-in in the preview chrome, which re-mints against the run-for-real allowlist. The new copy points the reviewer at it.

## Copy shipped

Title: `Permission unavailable in review`

Message = `<what>` + `<how>`:

- `what` (scopes known): `This app requested buzz:read:self.`
- `what` (nothing nameable): `This app requested a permission it doesn’t have here.`
- `how` (render-only): `Review previews run with reduced permissions — use “Run for real…” to grant the app its real permissions.`
- `how` (already in run-for-real): `Review previews always run with reduced permissions.`

Yellow, matching the existing "Permission unavailable" toast next to it.

## Naming the requested scopes — and the hardening that makes it safe

The message names the specific scope(s), because "which permission?" is the first thing a reviewer needs. Those strings come from the **untrusted** app's own frame/manifest, so:

- they are filtered through `isKnownBlockScope` (`src/shared/constants/block-scope.constants.ts`) before they can reach a moderator-facing string — only the fixed platform vocabulary is ever displayed, and anything else is silently dropped (falling back to the generic copy);
- they are rendered as ordinary React text by the notification component — no `dangerouslySetInnerHTML` anywhere on this path.

A test asserts that an injected string (``) and a made-up scope never appear in the message while a real scope still does.

## Anti-spam guard

Untrusted review code can post `REQUEST_CONSENT` in a loop, so a naive toast-per-message would let a hostile submission carpet-bomb the reviewing mod's screen and bury the review chrome. Two layers:

1. **A per-host-mount latch** (`reviewConsentNoticeShownRef`) — the notice fires at most once per mount. The review chrome remounts the host on a render-only ↔ run-for-real flip, so each mode still gets exactly one notice.
2. **A stable notification `id`** — an independent second guard; the toast library dedupes an already-displayed notification even if the latch were bypassed.

The transport's generic 30-messages/second limiter is **not** sufficient on its own: 30 toasts a second is still a carpet bomb. The flood test is deliberately kept *under* that transport budget so it asserts the latch, not the limiter.

One deliberate silent case is preserved: if the hint proves every requested scope is **already granted**, nothing is actually blocked, so nothing is reported.

## Behaviour outside review mode

Byte-identical. The review branch was moved *below* the existing post-handshake readiness gate, so the non-review path evaluates in the same order it always did and lands on the same code. The pre-existing tests for the dev-tunnel / "Permission unavailable" fallback and for the normal consent-modal path are unchanged and still pass; one existing test that asserted review mode stays *silent* was updated, since that silence is the bug.

## Tests

New unit cases for the pure decision (`resolveReviewConsentNotice`) and new browser cases for the host:

- review mode + `REQUEST_CONSENT` → notice fires, **no** consent modal opens;
- a hint-less request (the fire-and-forget SDK call sends no `scopes`) still notifies — the prod path stays silent without a hint because it can't tell "already granted" from "withheld"; in review there is nothing to tell apart;
- **anti-spam**: 9 messages → exactly 1 notice;
- unknown/injected scope strings never reach the message;
- benign re-request of an already-granted scope stays silent;
- run-for-real copy variant;
- non-review path unchanged (generic toast, no review copy).

### Mutation check

Both guards were verified by breaking them and watching the tests go red:

1. Reverted the handler to the original bare `if (reviewMode) return;` → **5 of the new browser tests failed** (`expected "vi.fn()" to be called 1 times, but got 0 times`); the pre-existing suites stayed green.
2. Removed **only** the anti-spam latch, leaving the notice in place → the flood test failed with `expected 1 times, but got 9 times`, confirming the latch (not the transport rate limiter) is what holds the line.

Both mutations were then reverted and the suites re-run green.

## Verified locally vs. not

**Ran and passed:**
- `vitest --project unit src/components/AppBlocks/__tests__/pageBlockHostLogic.test.ts` — 55/55.
- `vitest --project component` on `PageBlockHostReviewMode.browser.test.tsx` + `PageBlockHost.browser.test.tsx` — 48/48 (real Chromium).
- Both mutation checks above.
- ESLint on the five changed files: the only findings are a **pre-existing** `no-wholesale-module-mock` error on the two browser test files' `vi.mock('~/utils/trpc')` factories, present unchanged on `main`. Prettier: the three source files were already unformatted on `main`; the lines this PR adds are prettier-clean (reformatting the whole files would bury the diff).

**Could not run / not verified:**
- A clean full-repo `tsc --noEmit` — this machine can't generate the Prisma client, so the run is swamped by unrelated generated-client type errors. Filtered to the touched area, there are **zero** errors in `src/components/AppBlocks/**`, but treat CI's typecheck as the real gate.
- The rest of the unit suite (one unrelated pre-existing failure locally in `hiddenBlocks.test.ts`, which imports nothing this PR touches).
- No end-to-end click-through of a real review preview — I did not reproduce the original moderator click path against a running app. The browser tests drive the real postMessage bridge into the real host component, which is the closest deterministic proxy.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -317,7 +317,13 @@ describe('PageBlockHost REQUEST_CONSENT (W10 lazy-consent wiring)', () => {
     expect(useDialogStore.getState().dialogs).toHaveLength(0);
   });
 
-  test('Issue B: reviewMode never toasts an un-grantable consent request (untrusted preview stays silent)', async () => {
+  test('reviewMode surfaces a passive reduced-permissions notice and NEVER a consent modal', async () => {
+    // Was: reviewMode dropped REQUEST_CONSENT silently, so a moderator clicking a
+    // consent-gated action in the review preview got nothing at all and the app
+    // parked forever on its consent card. The modal ban is unchanged (untrusted
+    // review code must never pop a permission prompt at the mod) — only the
+    // silence is fixed. Full coverage lives in
+    // PageBlockHostReviewMode.browser.test.tsx.
     renderWithProviders(
       <PageBlockHost
         {...baseProps}
@@ -330,10 +336,10 @@ describe('PageBlockHost REQUEST_CONSENT (W10 lazy-consent wiring)', () => {
     );
 
     await driveToReady();
-    postFromBlock('REQUEST_CONSENT', { scopes: ['apps:storage:write'] });
-
-    await new Promise((r) => setTimeout(r, 150));
-    expect(showNotificationSpy).not.toHaveBeenCalled();
+    await vi.waitFor(() => {
+      postFromBlock('REQUEST_CONSENT', { scopes: ['apps:storage:write'] });
+      expect(showNotificationSpy).toHaveBeenCalledTimes(1);
+    });
     expect(useDialogStore.getState().dialogs).toHaveLength(0);
   });
 });
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -11,7 +11,10 @@ import { IframeInitController, shouldStartInit } from './iframeInitController';
 import { resolveBuzzPurchaseRequest } from './openBuzzPurchaseGate';
 import {
   decideAutoRetry,
+  advanceReviewConsentLatch,
+  buildReviewConsentNotification,
   grantedPageScopes,
+  INITIAL_REVIEW_CONSENT_LATCH,
   pageFallbackReason,
   resolveCheckpointPickerRequest,
   resolveGetImagesByIdsRequest,
@@ -50,7 +53,7 @@ import { PAGE_SLOT_ID } from '~/shared/constants/slot-registry';
 import { usePostMessage } from './usePostMessage';
 import type { BlockInitPayload, PageContext } from './types';
 import { dialogStore } from '~/components/Dialog/dialogStore';
-import { showNotification } from '@mantine/notifications';
+import { hideNotification, showNotification } from '@mantine/notifications';
 import { openLoginPopup } from '~/utils/auth-helpers';
 import type { BuyBuzzModalProps } from '~/components/Modals/BuyBuzzModal';
 import { openResourceSelectModal } from '~/components/Dialog/triggers/resource-select';
@@ -695,10 +698,12 @@ export function PageBlockHost({
     });
   }, [status, autoRetrySettled, appBlockId, blockInstanceId]);
 
-  // MOD REVIEW SANDBOX — emit-once latch for the reduced-permissions notice in
-  // the consent handler below. ONE per host MOUNT (the review chrome remounts the
-  // host on a render-only ↔ run-for-real flip, so each mode gets its own notice).
-  const reviewConsentNoticeShownRef = useRef<boolean>(false);
+  // MOD REVIEW SANDBOX — anti-spam latch for the reduced-permissions notice in
+  // the consent handler below. Bounded to at most TWO per host MOUNT (one generic
+  // + one scope-named upgrade — see `advanceReviewConsentLatch`). The review
+  // chrome remounts the host on a render-only ↔ run-for-real flip, so each mode
+  // gets its own budget, and the notification ids are mode-specific to match.
+  const reviewConsentLatchRef = useRef(INITIAL_REVIEW_CONSENT_LATCH);
   // Lazy consent (A6): the block (rendered in full for a logged-in viewer whose
   // page token is missing a consent-gated scope, e.g. `ai:write:budgeted` once
   // the page money scope is enabled) asks the host to open the consent UI when
@@ -744,26 +749,33 @@ export function PageBlockHost({
         // 🔴 ANTI-SPAM (required, not a nicety): the reviewed app is UNTRUSTED
         // code and can post REQUEST_CONSENT in a loop, so one toast per message
         // would let a hostile submission carpet-bomb the reviewing mod's screen
-        // (and bury the review chrome). Latch it to ONE notice per host mount.
-        // The stable `id` is a second, independent guard — Mantine dedupes a
-        // notification that is already displayed even if the latch is bypassed.
-        if (reviewConsentNoticeShownRef.current) return;
-        reviewConsentNoticeShownRef.current = true;
+        // (and bury the review chrome). The latch caps this at TWO notices per
+        // host mount — one generic, plus at most one upgrade to the scope-NAMED
+        // copy. It is NOT a plain "already notified" boolean: the SDK's `scopes`
+        // hint is optional, so a hint-less request on load would otherwise win
+        // the latch and permanently suppress the later, informative one.
+        const { show, next } = advanceReviewConsentLatch(
+          reviewConsentLatchRef.current,
+          notice.scopes.length > 0
+        );
+        reviewConsentLatchRef.current = next;
+        if (!show) return;
         // `notice.scopes` is filtered to the known block-scope vocabulary, so no
         // attacker-controlled text can reach this string; Mantine renders
         // `message` as React text (escaped) — never as HTML.
-        const what =
-          notice.scopes.length > 0
-            ? `This app requested ${notice.scopes.join(', ')}.`
-            : 'This app requested a permission it doesn’t have here.';
-        const how = reviewRunForReal
-          ? 'Review previews always run with reduced permissions.'
-          : 'Review previews run with reduced permissions — use “Run for real…” to grant the app its real permissions.';
+        const notification = buildReviewConsentNotification({
+          appBlockId,
+          runForReal: reviewRunForReal,
+          scopes: notice.scopes,
+        });
+        // Upgrade path only: retire the generic notice this one replaces, so the
+        // reviewer sees ONE notice, not a stack. A no-op when it already closed.
+        if (notification.supersedesId) hideNotification(notification.supersedesId);
         showNotification({
-          id: `review-consent-${appBlockId}`,
+          id: notification.id,
           color: 'yellow',
-          title: 'Permission unavailable in review',
-          message: `${what} ${how}`,
+          title: notification.title,
+          message: notification.message,
         });
         return;
       }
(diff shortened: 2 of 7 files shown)


ZacxDev:
## Audit fixes applied (643c24a)

Four issues from an adversarial audit of this PR. All four confirmed reproducible before fixing.

### 1. `isKnownBlockScope` prototype-chain bypass — **defense-in-depth, not a bypass**

`isKnownBlockScope` used `scope in BLOCK_SCOPE_TO_OAUTH_BIT`. `in` walks the prototype chain, so 12 inherited `Object.prototype` keys answered "known block scope": `__proto__`, `constructor`, `toString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString`, `__defineGetter__`, `__defineSetter__`, `__lookupGetter__`, `__lookupSetter__`. Now `Object.prototype.hasOwnProperty.call(...)`.

**Corrected severity framing — please read this rather than the scary-sounding shape of the bug.** This is **not** a validation bypass and **not** privilege escalation. Nothing here made a scope grantable, mintable, or enforceable that wasn't before:

- The **registration-time manifest path was never reachable.** `SCOPE_RE = /^[a-z0-9_]+(?::[a-z0-9_]+){1,3}$/` in `block-manifest-validator.service.ts` requires at least one colon and rejects all 12 keys *before* the predicate runs. Verified.
- The **real exposure was narrow**: this PR routes untrusted runtime `payload.scopes` — from the reviewed app's own frame — through the predicate into moderator-facing toast text, and that path has no regex shape-check. So the keys could have been printed verbatim at the reviewer.
- It was also a **false comment**: the code claimed only the fixed platform vocabulary could reach that string. That claim is now actually true.

Callers that index `BLOCK_SCOPE_TO_OAUTH_BIT` immediately after the predicate (`deriveOauthBitmaskFromBlockScopes`, `validateBlockScopesAgainstOauthClient`) would have read a *function* rather than a number for those keys. Both outcomes are now pinned by tests.

`isKnownBlockScope` is a shared helper, so I ran the suites of **every** call site, not just the files I edited — all green, and none depended on the prototype-inclusive behaviour:

| Call site | Suite | Result |
|---|---|---|
| `block-scope.constants.ts` (`validateBlockScopesAgainstOauthClient`, `deriveOauthBitmaskFromBlockScopes`) | `block-scope.constants.test.ts` + `block-scope.schema-drift.test.ts` | 37 + 3 pass |
| `block-manifest-validator.service.ts` | `block-manifest-validator.service.test.ts` | 169 pass |
| `block-scope.middleware.ts` | middleware + anytoken-mode + metrics + runtime-flag | 34 + 9 + 6 + 3 pass |
| `dev-scoped-mint.service.ts` | `dev-scoped-mint.service.test.ts` | 39 pass |
| `blockScopeSelection.ts` | `blockScopeSelection.test.ts` | 14 pass |
| `api/v1/block-tokens` | 5 route suites | 90 pass |
| `pageBlockHostLogic.ts` (this PR) | `pageBlockHostLogic.test.ts` | 68 pass |
| adjacent | `scope-grant`, `collections-scope-plumbing`, `block-token.service`, `token-scope.constants`, `enforce-token-scope` | all pass |

### 2. Notification `id` collided across review modes

The id was `review-consent-<appBlockId>` in **both** render-only and run-for-real. Mantine's store no-ops `showNotification` for an id already displayed or queued (default `autoClose` 4000ms), so the realistic sequence — notice fires in render-only → mod clicks "Run for real…" → the chrome remounts the host → the latch resets *as designed* → the app re-requests within 4s — **silently swallowed** the run-for-real notice, re-creating the exact bug this PR fixes, in the other mode. The id is now mode-specific. The existing browser tests mock `@mantine/notifications` wholesale, so this path had zero coverage; it now has coverage at both the pure-logic and browser layers.

### 3. First-notice-wins buried the informative message

The latch was set on the first *notifying* request regardless of whether it named scopes. `useRequestConsent()`'s `scopes` payload is **optional**, so a hint-less request is a normal path: it produced the generic copy, closed the latch, and suppressed the app's later specific `['buzz:read:self']` request for the rest of the mount. The moderator never learned which permission — defeating the point of naming scopes at all (and all of fix 1's hardening).

The latch now tracks whether a scope-**named** notice has been shown and allows exactly one upgrade generic → named.

**Anti-spam is preserved, and is still the reason the latch exists.** The bound is **at most two notices per host mount** (one generic + one named upgrade), not two per distinct scope. It's written as an explicit state machine so the bound is provable, and tested with a 200-message flood across every ordering (all-named, all-generic, alternating both ways, random). Removing the bound fails 5 tests.

The upgrade deliberately does **not** reuse the id and does **not** use `updateNotification`: reusing the id would be deduped away while the generic is still up, and `updateNotification` only maps over live notifications, so it would drop the upgrade once the generic had auto-closed. Instead the named notice carries its own id and retires the generic via `hideNotification` (a harmless no-op if it already closed) — exactly one visible notice in both timing cases.

### 4. Copy now discloses what "Run for real…" costs

Untrusted code can emit this toast unprompted right after `BLOCK_READY`, and it's the one surface pointing a reviewer at the opt-in — which re-mints against the **moderator's own account and spends their own Buzz** (`ai:write:budgeted`, session-capped). Pointing at a spend-your-own-money button without saying so is how a hostile app gets a mod to click it. Render-only copy now reads:

> Review previews run with reduced permissions — use "Run for real…" (runs against your own account and Buzz) to grant the app its real permissions.

Run-for-real copy is unchanged (the mod already took the opt-in).

### Constraints held

No scope allowlist or mint change (`REVIEW_MINT_SCOPE_ALLOWLIST` / `REVIEW_RUN_FOR_REAL_MINT_SCOPE_ALLOWLIST` untouched) · review mode still **never** opens a consent modal · non-review behaviour byte-identical · anti-spam bound intact.

### Mutation checks — every fix broken, failure observed, restored

| # | Mutation | Result |
|---|---|---|
| 1 | `hasOwnProperty` → `in` | **2 fail** — `block-scope.constants.test.ts` (predicate) + `pageBlockHostLogic.test.ts` (untrusted-input caller) |
| 2 | drop the mode suffix from the id | **2 fail** — unit (distinct ids) + browser (mode-specific id) |
| 3 | latch back to a plain first-notice-wins boolean | **3 fail** — unit (upgrade) + browser (hint-less doesn't bury; two-notice bound) |
| 3b | remove the bound entirely (unbounded toasts) | **5 fail** — 3 unit + 2 browser anti-spam |
| 4 | drop "(runs against your own account and Buzz)" | **2 fail** — unit (copy) + browser (notice copy) |
| 5 | remove the `hideNotification` supersede wiring | **1 fail** — browser (upgrade retires the generic) |

Honest note: one test I added — `prototype keys contribute nothing downstream of the predicate` — **passes under mutation 1** and is therefore a regression pin, not a guard. Both helpers already coerced those keys to a no-op via bitwise `NaN`. Keeping it as a downstream invariant pin; the actual guards are the two tests that fail.

### Suites run

- **Full unit project**, before and after: `610 passed / 91 failed` files, `8434 passed / 18 failed` tests — **byte-identical failure sets** (`diff` of both failed-file and failed-test lists is empty). All pre-existing: Prisma query-engine unavailable for `linux-nixos` locally, plus pre-existing `hiddenBlocks` / `get-models-raw.transient-503` / `manifest-schema-endpoint` failures. None touch this change.
- **Browser (real Chromium)**: `PageBlockHostReviewMode` 26/26, `PageBlockHost` 25/25, `PageBlockHostAutoRetry` 14/14 — 65/65.
- **ESLint** on all 6 changed files: clean except one **pre-existing** `no-wholesale-module-mock` error on the `vi.mock('~/utils/trpc')` factory that already existed at the PR head (verified by linting the pristine file — same rule, line 51 → 59 purely from my added lines). Left alone as out of scope.
- **Prettier**: not run. All three source files were **already** prettier-dirty at the PR head, so formatting them would have inflated the diff from 52 to 464 changed lines in `PageBlockHost.tsx` alone with unrelated churn.

### Not verified

- **Typecheck**: `tsc --noEmit` locally reports 1,933 errors, essentially all from a stale/ungenerated Prisma client (`prisma generate` fails on NixOS). **Zero** are in the 6 changed files, and zero mention any changed symbol (`isKnownBlockScope`, `resolveReviewConsentNotice`, `advanceReviewConsentLatch`, `buildReviewConsentNotification`, `hideNotification`, `BlockScopeString`). CI is the real typecheck gate.
- **No manual browser run** against a live review preview — the moderator click-path is covered by the browser suite driving the real postMessage bridge, not by a human walkthrough.

PR #3496, opened 2026-07-31
fix(apps): review preview builds the actual submitted code for git-push submissions

ZacxDev:
## Symptom

In the moderator app review flow, the **Preview** iframe could show a moderator
the wrong code. For a submission that arrived by **`git push`** (rather than a
ZIP upload), the preview built an *older* version's source tree — silently, with
no error and no visual tell — while every other surface in the same review UI
(the file/manifest diff, the "view code" link, and what approve actually ships)
correctly showed the pushed commit.

Two variants, depending on the app's history:

| App's submission history | What the preview did |
|---|---|
| has had at least one ZIP upload before | **silently previewed the last ZIP-submitted version** |
| only ever pushed | the lookup 404s → preview fails loudly |

The silent variant is the dangerous one: a moderator can approve after reviewing
a preview of code that is not the code being approved.

## Root cause

*(line numbers below are on `main`, i.e. pre-fix.)*

The review preview builds from a per-app **in-review snapshot repo**. That
snapshot is written by exactly one path — `submitVersion`, the ZIP-upload path
(`src/server/services/blocks/publish-request.service.ts:1205-1215`, `ensureReviewRepo`
+ `commitFiles({ replaceAllFiles: true })`).

The git-push path (`recordPendingFromPush`,
`src/server/services/blocks/publish-request.service.ts:1800-1894`) deliberately does
**not** touch it: it parks the request with empty bundle pointers
(`bundleKey: ''`) and the pushed `forgejoCommitSha`, because the developer's own
repo at that commit *is* the reviewable artifact.

But `previewRequest`
(`src/server/services/blocks/publish-request.service.ts:3790-3831`, pre-fix) selected
only `{ id, status, slug }` — it never read the bundle/commit pointers, so it had
no way to tell the two origins apart. It then unconditionally called
`getReviewRepoHeadSha(request.slug)`
(`src/server/services/blocks/forgejo.service.ts:639-646`), whose docstring encodes the
ZIP-only premise verbatim: *"submitVersion pushes the pending bundle to this repo
… so HEAD is exactly the pending version's source."* True for ZIP, false for push.

Everything else in the review surface was already origin-aware — the line diff
(`getPublishRequestDiff`, `…/publish-request.service.ts:4724-4741`) reads
`forgejoCommitSha`, the code link falls back `pushCommitUrl ?? reviewRepoUrl`
(`src/components/Apps/OnsiteReviewModal.tsx:419`), and `approveRequest`
(`…/publish-request.service.ts:2921-2930`) reconstructs from the pinned commit.
The preview was the lone holdout.

## The fix

Two parts, one commit.

**1. The guard (`resolveReviewSourceSha`).** `previewRequest` now selects
`bundleKey` + `forgejoCommitSha` and routes on origin, exactly as `approveRequest`
does. A request with *neither* pointer — which cannot be attributed to any source
tree — now throws instead of falling through to the snapshot HEAD. The
silent-mismatch path is structurally gone: there is no longer a code path that
reads the snapshot HEAD without having first established that the snapshot holds
this version.

**2. Preview the correct tree.** For a push-originated request, mirror the
reviewed commit's tree into the snapshot repo first
(`reconstructBundleFromForgejo(slug, forgejoCommitSha)` →
`extractBundleFilesFromBuffer` → `ensureReviewRepo` + `commitFiles({ replaceAllFiles: true })`),
then build the sha that mirror produced. ZIP-originated requests are untouched.

`getReviewRepoHeadSha`'s docstring is corrected to state the ZIP-only premise and
the caller contract, so the next caller doesn't inherit the same wrong assumption.

### Why this shape, and the alternative I rejected

The alternative was to leave the snapshot repo alone and instead tell the build
which repo to clone, threading an allowlisted repo selector through the signed
build-trigger payload so a push-originated preview builds the developer's repo at
the pushed commit directly. That is arguably the more semantically honest design
— two repos, no mirroring, and the preview host would then encode the exact sha
that gets approved.

I did not take it, for one reason: **it cannot ship as a single reviewable
change.** The payload is HMAC-signed over the whole body, so both sides must
agree, and the receiving side is a separately-deployed component. It would have
to be deployed *first* and be a strict no-op when the new field is absent;
merging this side first would turn a wrong-but-working preview into a failing
one. That is a real sequencing hazard for a fix whose entire purpose is to stop
moderators seeing the wrong thing.

The chosen approach touches one repo, reuses four helpers already proven on the
submit/approve paths, and has no deploy ordering constraint at all. **No changes
are required anywhere else, and there is no deploy-ordering requirement.**

Its two honest costs:

- one bundle reconstruct round-trip per preview click on a push-originated
  request. The mirror *write* is skipped when the tree is unchanged (see
  below); the reconstruct itself still runs on every click.
- the preview's host name encodes the mirror commit rather than the reviewed
  commit. The reviewed commit is still what the diff shows and what approve
  ships — only the preview URL's opaque id differs.

If the two-repo design is wanted later, this change is a clean base for it: the
origin split is already isolated in one function.

### Repeat preview — the sha is stable, and it took an explicit guard

> **Correction.** An earlier revision of this description claimed the mirror was
> "idempotent: a second click with an unchanged tree produces no new commit and
> returns the same sha". **That was false**, and it is worth saying plainly
> because it is the kind of claim that reads as verified and wasn't.
>
> `commitFiles` builds its operation list without comparing content: for a path
> that already exists it emits an `update` op unconditionally. Its
> `operations.length === 0` no-op branch fires only when the *file list is
> empty* — never for an unchanged tree. (The comment on that branch said
> "bundle identical to repo state", which is what the claim was read from; that
> comment was pre-existing rot and is corrected in this PR.)

Left as-is, every "Rebuild preview" click on a push-originated request would
mint a new commit → new sha → new preview host → new image tag → new resource
set, for a tree nobody changed. Three guards downstream are keyed on that sha
being stable across rebuilds and would have quietly stopped doing their job:
the replay-dedup key, the concurrent-preview cap (which excludes the current
request precisely so a *rebuild* is never blocked), and the deploying-state
callback write (which carries no expected-sha, so a superseded build could
overwrite the current one's detail). It also left an unknown I could not
resolve without a live environment: whether the remote accepts or rejects a
batch of byte-identical update ops. If it rejects, Rebuild would fail outright
on every push-originated request after the first.

So the mirror is now **conditional**, which removes the churn and the unknown
together. Before writing, the reconstructed file set is compared against the
snapshot repository's current tree; if they already match, no commit is made
and the existing head sha is returned.

**What that guarantees:** a repeat preview of the *same* commit with an
unchanged tree performs no write and returns the *same* sha — so the preview
host, the dedup key and the cap row are all stable across rebuilds.

**What it does not weaken:**

- The comparison is *exact and symmetric* — identical path sets in **both**
  directions plus a matching content id at every path. A file present in the
  repo but absent from the reconstruction **must** force a mirror, because
  `replaceAllFiles` is what deletes it; treating that as "unchanged" would
  leave a stale file being served, i.e. re-introduce this PR's own bug in a
  new shape.
- It **fails toward correctness**: any error or ambiguity — repo not created
  yet, truncated tree listing, transport failure, a duplicate path — mirrors
  anyway. A needless commit is cheap; a wrong "unchanged" is a stale preview.
- It compares the **reconstructed file set** (what would actually be
  committed), not the source tree listing, so a divergence introduced by the
  reconstruct → extract pipeline can never read as "nothing changed".

The comparison costs no extra fetches: a tree listing already reports each
file's content id, and the same id is computable locally from the bytes. The
computation was checked against `git hash-object` on text, empty and binary
(NUL + high-byte) inputs before being relied on, and the tests pin ids produced
by git itself rather than by the code under test — so a drift in the
computation fails the suite instead of agreeing with itself.

## Exposure: unapproved source becomes readable at preview time (accepted, documented)

Mirroring publishes an unapproved third-party submission's source into the
in-review snapshot repository at **preview** time rather than at **approve**
time. Within the review environment those snapshot repositories are readable
without an individual login, so in practice this moves push-originated
submissions from "readable once approved" to "readable once a moderator clicks
Preview".

**Decision: accept, and record it here.** The ZIP lane already writes the
submitted source into the same snapshot repository at *submit* time, so this
makes the push lane consistent with the lane beside it rather than introducing
a new class of exposure. Nothing here is a new kind of readable — only a new
moment for the push lane, and a *later* one than the ZIP lane's.

**Deliberately not changed in this PR:** the snapshot repositories' visibility
setting and the review environment's access routing. Both exist to let
moderators browse a submission's file tree from the review deep-link without a
second login; changing either would break moderator browsing and is a larger,
separate change that deserves its own review.

**Pre-existing, worth its own issue — not fixed here:** snapshot repositories
are never deleted. A submission that is *rejected* or *withdrawn* therefore
leaves its source readable indefinitely, and the justification for the
browse-without-login affordance only ever covered submissions that go on to be
approved. That retention gap predates this PR and is orthogonal to it, but this
change makes it reachable slightly earlier in a submission's life, so it should
be tracked rather than left implicit.

## Explicitly not changed

- **`approveRequest`'s sha handling.** What ships is still driven by the pinned
  `forgejoCommitSha`. Untouched.
- The moderator gate, the trigger signing, and the review sandbox scope clamps.
- No schema change, no migration.

## Test coverage

All in `src/server/services/blocks/__tests__/publish-request.reviewSandbox.test.ts`.

**`previewRequest source tree by origin`** — four cases:

1. **push-originated** (empty bundle pointer + a commit) — asserts the tree is
   read at *that commit* (not a branch head), the mirror is a `replaceAllFiles`
   commit to the snapshot repo with the reconstructed file set, the stale
   snapshot HEAD is never consulted, and the build + returned host use the
   mirrored sha. It now also asserts the mirrored **content**, not just the
   path list (see below).
2. **ZIP-originated** — byte-identical to today: reads the snapshot HEAD, writes
   nothing, builds the same sha.
3. **both pointers set** (an app with prior ZIP history) — the bundle pointer
   wins, matching `approveRequest`'s discriminator.
4. **neither pointer** — throws; no build is triggered.

**`previewRequest repeat preview (push origin, unchanged tree)`** — six cases
covering the conditional mirror:

5. **identical tree** — no commit is made, the existing head sha is reused, and
   the build + returned host use *that* sha.
6. **same paths, different bytes at one path** — mirrors.
7. **a file the developer added** (missing from the snapshot repo) — mirrors.
8. **a file the developer removed** (still present in the snapshot repo) —
   mirrors, with `replaceAllFiles` so it actually gets deleted. This is the
   asymmetric case: every reconstructed path matches, so a one-directional
   comparison would wrongly call it unchanged.
9. **tree listing fails** — mirrors rather than assuming unchanged.
10. **ZIP-originated** — the comparison is never even attempted; no repo write.

### Testing the content, not just the paths

Case 1 previously asserted the mirrored **file paths** only. A mutation that
corrupts the mirrored **bytes** while leaving the paths correct — double
encoding, wrong charset — passed **42/42 green** under those assertions; I ran
it to confirm rather than assuming. Case 1 now asserts the actual mirrored
bytes for both fixture files (which make a real round trip through the bundle
ZIP), and the same mutation fails.

### Mutation check

Each mutation was applied to the code, the suite run, then reverted. Every one
was observed failing and observed green again after restore.

| Mutation | Result |
|---|---|
| Revert the origin fix (always read the snapshot HEAD) | **2 tests fail** (push-originated, neither-pointer) |
| Drop `replaceAllFiles` from the mirror commit | **1 test fails** |
| Mirror the branch head instead of the pinned commit | **1 test fails** |
| Build the pushed commit instead of the mirrored sha | **1 test fails** |
| Fall through to the snapshot HEAD when neither pointer is set | **1 test fails** |
| **Remove the repeat-preview short-circuit** (always mirror) | **1 test fails** — the identical-tree case |
| **Make the short-circuit always fire** (never mirror) | **4 tests fail** — the mirror case + all three changed-tree cases |
| **Make the comparison one-directional** (drop the both-directions check) | **1 test fails** — the removed-file case, and *only* that one |
| **Break the content-id computation** (malformed header) | **1 test fails** — the identical-tree case, i.e. the fixtures really are pinned to git's own output |
| **Make the error path assume "unchanged"** | **1 test fails** — the failed-tree-listing case |
| **Corrupt the mirrored bytes, keep the paths correct** | **1 test fails** with the new content assertion; **passes 42/42 green** without it |

All restored → 48/48 green.

## Verified vs. not run

**Verified locally:**

- `vitest --project unit` on the touched suite: **48/48 pass** (38 before this
  PR, 42 after the first commit, +6 for the conditional mirror).
- Whole `src/server/services/blocks` unit suite, this PR's first commit vs. its
  head, both measured in clean checkouts: identical `19 failed / 74 passed`
  files and `18 failed` tests on both sides, `1451 → 1457` passing — exactly the
  6 new tests, nothing else moved. The 18/19 failures are a pre-existing local
  environment issue (`Cannot find module '.prisma/client/default'` — the client
  isn't generated on this machine), unrelated to this change.
- The content-id computation checked against **`git hash-object`** on four
  inputs (two text, empty, and binary containing a NUL and high bytes) before
  being used; all four matched. The test fixtures pin git's output, not the
  code's.
- The repo's own `no-io-in-transaction` / `no-wholesale-module-mock` rule tests:
  117/117 pass.
- `tsc --noEmit`, filtered to the touched files: the only error in
  `publish-request.service.ts` is a pre-existing `TS7006`, byte-identical
  before and after this PR's changes. A whole-repo typecheck is not meaningful
  on this machine for the same missing-generated-client reason.
- `eslint` on the touched files: 0 errors, and the **same 6 warnings** before
  and after — all pre-existing.

**Not run / not verified — please confirm in CI and on the preview:**

- No end-to-end run against a real review preview: I did not click Preview on a
  push-originated submission and confirm the served code matches the pushed
  commit. That is the real acceptance test for this fix and it needs a live
  environment.
- The mirror path's repository interactions are mocked in tests; the
  reconstruct → commit round-trip has not been exercised against a real repo
  here (it is the same code approve runs for this origin, but that is an
  argument, not evidence).
- **The repeat-preview short-circuit has not been exercised against a live
  repository.** The comparison logic is fully covered against mocks, and the
  content-id computation is checked against real git — but that a real tree
  listing reports exactly that id for exactly these bytes is verified only by
  the fact that the existing mirror code already uses those ids as update
  targets. Worth a single manual check: click Preview twice on an unchanged
  push-originated submission and confirm the URL's id does not change.
- Relatedly, the original open question — whether the remote accepts or
  rejects a batch of byte-identical update ops — is now **unreachable rather
  than answered**. The short-circuit means we no longer send that batch, so the
  behaviour stays unknown; it just no longer matters on this path.
- Full-repo typecheck and build — deferred to CI.


ZacxDev:
## Audit follow-ups applied (5 items)

Pushed as `fecd8f7` on top of `0bce187`. The origin-routing mechanism itself is unchanged — this is the conditional-mirror guard the audit asked for, plus the test and description corrections.

### 1. The idempotency claim was false — short-circuit added (code)

The description claimed a second Preview click on an unchanged tree produced no new commit and returned the same sha. It did not. `commitFiles` never compares content: for a path that already exists it emits an `update` op unconditionally, and its `operations.length === 0` no-op branch fires only when the *file list is empty*. The comment on that branch ("bundle identical to repo state") was pre-existing rot and is what the claim was read from — corrected here.

Left alone, every "Rebuild preview" on a push-originated request minted a new commit → new sha → new preview host → new resource set for a tree nobody touched, quietly defeating three guards that assume a stable sha across rebuilds (replay-dedup key, the concurrent-preview cap that deliberately excludes the current request, and the deploying-state callback write, which carries no expected-sha). It also left the unresolved question of whether the remote accepts or rejects a batch of byte-identical update ops — if it rejects, Rebuild would have failed outright on every push-originated request after the first.

`resolveReviewSourceSha` now compares before mirroring, and returns the existing head sha when the tree already matches. Properties, all mutation-checked:

- **Exact and symmetric.** Identical path sets in *both* directions plus a matching content id at every path. A file present in the repo but absent from the reconstruction must force a mirror — `replaceAllFiles` is what deletes it, and calling that "unchanged" would re-introduce this PR's own bug in a new shape.
- **Fails toward correctness.** Any error or ambiguity — repo not created yet, truncated tree, transport failure, duplicate path — mirrors anyway.
- **Compares what would actually be committed** (the reconstructed file set), not the source tree listing, so a divergence in the reconstruct → extract pipeline can't read as "nothing changed".
- **No extra fetches.** A tree listing already reports each file's content id; the same id is computable locally from the bytes.

ZIP path untouched — it never reaches this branch. `approveRequest` untouched.

**How the id computation was verified:** against real `git hash-object` on four inputs (two text, empty, and binary containing a NUL byte and high bytes) *before* relying on it — all four matched. The test fixtures pin ids produced by git itself, not by the code under test, so a drift in the computation fails the suite rather than agreeing with itself. Confirmed by mutation: malforming the header fails the identical-tree test.

### 2. Description corrected

The false sentence is gone, replaced by an explicit correction note (it read as verified and wasn't, which is worth saying out loud) plus a section stating what the short-circuit actually guarantees and what it deliberately does not weaken.

### 3. Tests now assert content, not just paths

The existing mirror test checked the mirrored **path list** only. I confirmed rather than assumed that this was insufficient: a mutation corrupting the mirrored **bytes** while leaving paths correct **passed 42/42 green**. The test now asserts the actual mirrored bytes for both fixture files (which make a real round trip through the bundle ZIP), and the same mutation fails.

### 4. Repeat-preview coverage added

Six new cases: identical tree → no commit, existing head sha reused; changed bytes → mirrors; added file → mirrors; **removed file → mirrors** (the asymmetric case a one-directional comparison would get wrong); failed tree listing → mirrors; ZIP-originated → comparison never attempted.

### 5. Exposure decision documented

Mirroring publishes an unapproved submission's source into the in-review snapshot repository at **preview** time rather than **approve** time, and those snapshot repositories are readable without an individual login inside the review environment.

**Accepted and documented in the description.** The ZIP lane already writes submitted source into the same snapshot repository at *submit* time, so this makes the push lane consistent with the lane beside it — a new moment for the push lane, and a later one than the ZIP lane's, not a new class of exposure.

Repository visibility and the review environment's access routing are **deliberately unchanged** — both exist so moderators can browse a submission's file tree from the review deep-link without a second login, and changing either would break that. Separate, larger change.

Also recorded as **pre-existing and worth its own issue, not fixed here**: snapshot repositories are never deleted, so a *rejected or withdrawn* submission's source stays readable indefinitely, and the justification for browse-without-login only ever covered submissions that go on to be approved.

---

### Verification

| Check | Result |
|---|---|
| Touched suite | **48/48** (38 → 42 → 48; +6 here) |
| `services/blocks` suite, first commit vs. head, both in clean checkouts | identical `19 failed / 74 passed` files and `18 failed` tests; `1451 → 1457` passing — exactly the 6 new tests |
| Mutation checks | 6 new mutations, each observed failing and observed green after restore (table in the description) |
| Content-id computation vs. `git hash-object` | 4/4 inputs match |
| `no-io-in-transaction` / `no-wholesale-module-mock` rule tests | 117/117 |
| `tsc --noEmit`, filtered to touched files | only a pre-existing `TS7006`, byte-identical before/after |
| `eslint` on touched files | 0 errors; same 6 warnings before and after, all pre-existing |

The 18 failing tests / 19 failing files are a pre-existing local environment issue (generated Prisma client not present on this machine) and are identical on both sides. Whole-repo typecheck and build are not meaningful locally for the same reason — CI is the real gate.

**Still not verified, honestly:**

- No end-to-end run against a live review preview — nobody has clicked Preview on a push-originated submission and confirmed the served code matches the pushed commit. That remains the real acceptance test.
- **The short-circuit has not been exercised against a live repository.** The logic is fully covered against mocks and the id computation is checked against real git, but that a real tree listing reports exactly that id for exactly these bytes rests on the existing mirror code already using those ids as update targets — an argument, not evidence. One cheap manual check settles it: click Preview twice on an unchanged push-originated submission and confirm the URL's id does not change.
- The original open question — whether the remote accepts or rejects byte-identical update ops — is now **unreachable rather than answered**. We no longer send that batch, so the behaviour stays unknown; it just no longer gates this path.

PR #3498, opened 2026-07-31
fix(apps): make in-review submission snapshots private, backfill, and reclaim them

ZacxDev:
## The problem

When a developer submits an app for review, we push their source into a per-slug **in-review snapshot repository** so moderators can read it. Those snapshots were created with **broader read access than intended** — they hold the full source of a submission that has **not been approved yet**, i.e. someone else's unreviewed code.

## How it happened

Neither change that produced this was wrong on its own.

1. The snapshots were created with relaxed visibility **deliberately**, with a comment explaining why: an access control sitting elsewhere in the request path meant the relaxed setting only ever widened access *to reviewers*. That was true when it was written.
2. Later, that other layer was changed for an unrelated reason, and the property the comment depended on no longer held.

Change 1 was correct given the world it was written in. Change 2 was correct given its own goal. The setting from change 1 was never re-checked against the world change 2 created, and there was nothing to force that — the justification lived in a code comment describing a property of a system owned somewhere else.

That's the shape of the bug worth remembering: **a security decision that depends on a property owned somewhere else, with no mechanism that re-checks it.** The fix below deliberately moves the boundary into the repository itself, so it holds regardless of how a request arrives.

## What each change does

**1. Snapshots are created private** (`ensureReviewRepo`). Nothing needs relaxed read: the review build authenticates with its own credential, and moderators read submissions through the in-app diff. The replaced comment now states the real posture and says not to flip it back without re-checking the surrounding auth posture.

**2. A moderator-only backfill for existing snapshots** (`blocks.backfillReviewRepoPrivacy`). The create call is idempotent on an already-existing repo, so a re-submit does **not** retroactively change an old repo's visibility — pre-existing snapshots keep their original setting until something walks the org and patches them. Built to be re-run: already-private repos are skipped without a PATCH, a repo that vanished mid-run counts `missing` rather than failed, per-repo failures are collected and the walk continues, and `dryRun` reports exactly which slugs a real run would touch while changing nothing. **This requires a human to run it — see below.**

**3. The raw-source deep-link in the review modal is retired.** Once snapshots are private, the "View full source" button and the per-file external fallback both fail for the moderator who clicks them, and a dead link is worse than none — it reads as a working affordance and sends the reviewer somewhere broken mid-review. Moderators already have the in-app code diff in the same modal; that is now the sole source-review path, and elided files (binary / too large) say *why* they have no inline diff instead of linking out. The `reviewRepoUrl` / `pushCommitUrl` service values are **kept** on the payload — a richer in-app file browser is being designed separately and will reuse them. Only the user-facing dead link is removed.

> **Known gap until that browser lands:** a moderator reviewing a file the diff cannot render (binary, or over the size cap) now has no path to those bytes. Worth stating to the mod team explicitly.

**4. Retention: snapshots are reclaimed 30 days after a terminal decision** (`purge-review-snapshots` job). Once a request is rejected or withdrawn nobody reads its snapshot again, but it sits there holding a full copy of third-party source indefinitely. Safe to delete because the snapshot is a **derived cache, never the system of record**: a ZIP submission's bundle lives in object storage under `bundleKey`, and a push submission's source lives in the canonical repo at the pinned `forgejoCommitSha`.

Implemented as a **sweep**, not an inline delete on the reject path, so a decision made while this was undeployed or broken still gets collected later, and so a moderator-facing write never depends on a slow external call. Bounded per run (50 rows), resumable via a `KeyValue` cursor, log-and-continue per repo, fail-open. Window is a single named constant `REVIEW_SNAPSHOT_PURGE_AFTER_MS`.

### The correctness gate in the sweep

Snapshots are keyed **per-slug and overwritten on every submit**, so the repo a 40-day-old rejection points at may today hold a brand-new, still-pending submission's source. Purging on the old rejection would destroy an in-flight review.

The sweep holds back any slug with a pending publish request, and reads that check **from the primary, not the replica** — the blocking row can be seconds old while the triggering row is 30+ days old, so a lagging replica would hide exactly the row whose presence is load-bearing, and the failure mode is destroying live work.

## 🔴 Two manual steps a human owes before this is complete

**A. Run the backfill.** Merging fixes new snapshots only. Every snapshot created before this merge keeps its original visibility until a moderator runs `blocks.backfillReviewRepoPrivacy` **once**. Suggested: `{ dryRun: true }` first to see the slug list, then `{}` for real. Safe to re-run. **Run it in the same window as the merge.**

**B. Refresh the scheduler.** A newly-registered recurring job in this codebase is **not live on deploy** — the external scheduler does not sync on start or restart. `purge-review-snapshots` will not run until `GET /refresh` is hit on the scheduler. To verify one run manually without waiting for the cron, trigger it directly at the jobs webhook: `?run=purge-review-snapshots` (`&noCheck=true` to bypass the run lock). Expect `{ scanned, candidates, deleted, alreadyGone, skippedPending, failed }`.

Note the retention sweep is a no-op until (B) is done, and the visibility fix is incomplete until (A) is done. Neither blocks the other.

## No DB migration

Nothing here needs a schema change. The resume cursor uses the existing `KeyValue` table via `getJobDate`, and the sweep reads existing columns (`status`, `slug`, `updatedAt`).

## Verified vs not

**Verified locally:**
- `src/server/services/blocks` + `src/server/jobs` unit suites: **120 files / 2034 tests passing** (baseline at `main`: 117 / 1992). +3 files, +42 tests, zero regressions.
- **15 mutation checks, all killed** — every guard was broken, the failure confirmed, and the file restored:
  - `private: true` → `false`; `setReviewRepoPrivate` patching `false`
  - `deleteReviewRepo` losing its 404-is-success branch; retargeted at the canonical app org
  - the pending gate removed entirely; the pending gate reading the replica instead of the primary
  - the retention window zeroed; quietly shortened 30d → 7d; the `lt cutoff` predicate removed
  - per-slug dedupe removed; the cursor never advancing; `approved` added to the purgeable statuses
  - both log-and-continue catches rethrowing; `dryRun` still issuing the PATCH
- Two boundary tests were **strengthened during this** — they originally derived their fixtures from the retention constant, so they stayed green when it was mutated to zero. They now drive a `findMany` mock that applies the real where-clause, with fixtures pinned to absolute dates 29 and 31 days out.
- `tsc --noEmit`: **22 errors before and after, none in any touched file.**
- `eslint` on every touched file: **13 errors before and after, all pre-existing.**

**NOT verified — CI is the gate:**
- **Browser-mode component tests did not run locally** (no matching Playwright build available on this machine). The four touched `.browser.test.tsx` files are updated and typecheck clean, but I have not seen them execute. This includes the regression guard asserting no raw-source link renders — and one assertion in that group is known to be vacuous (the string it looks for only renders inside a collapsed panel that the test never expands); it is being replaced.
- `src/server/routers` has **21 pre-existing failing files at `main`** on this machine (missing generated Prisma client / `event-engine-common`). Identical before and after — but that means the new `backfillReviewRepoPrivacy` procedure's registration is **not** covered by a locally-executed test.
- **Nothing has been run against a real review source host.** No backfill run, no purge run, no live delete. The review build's authenticated clone against a now-private snapshot is the one path that must be exercised post-merge — trigger one review preview and confirm the clone step succeeds.

### DIFF
--- a/src/components/Apps/CombinedReviewModal.browser.test.tsx
+++ b/src/components/Apps/CombinedReviewModal.browser.test.tsx
@@ -252,7 +252,7 @@ describe('CombinedReviewModal', () => {
     await expect.element(page.getByRole('tab', { name: /Scopes/ })).toBeInTheDocument();
     await expect.element(page.getByRole('tab', { name: /Code review/ })).toBeInTheDocument();
     // Code section: the on-site bundle affordance.
-    await expect.element(page.getByText('View full source')).toBeInTheDocument();
+    await expect.element(page.getByText('Show code diff')).toBeInTheDocument();
 
     // Media section: the listing PREVIEW (card + detail) + the content-review surface.
     await expect.element(page.getByTestId('apps-listing-preview')).toBeInTheDocument();
--- a/src/components/Apps/OnsiteReviewModal.browser.test.tsx
+++ b/src/components/Apps/OnsiteReviewModal.browser.test.tsx
@@ -9,8 +9,10 @@ import { renderWithProviders } from '../../../test/component-setup';
  * `OnsiteReviewModal.tsx` (mirrors #3154) so it is importable WITHOUT the page's
  * `getServerSideProps` server graph — this suite is the coverage that extraction
  * unlocks. Asserts:
- *  - onsite-specific visibility (View full source link + the mod Review-preview panel +
- *    the structured manifest render);
+ *  - onsite-specific visibility (the in-app code-diff affordance + the mod
+ *    Review-preview panel + the structured manifest render), and that NO
+ *    raw-source deep-link is offered (retired in #3495 — in-review snapshots are
+ *    private, so such a link would 404 for the moderator who clicked it);
  *  - Approve FIRES `blocks.approveRequest` with the request id;
  *  - the reject reason gate (disabled < 3 chars) AND that Reject FIRES
  *    `blocks.rejectRequest` with `{ publishRequestId, rejectionReason }`;
@@ -165,14 +167,22 @@ beforeEach(() => {
 });
 
 describe('OnsiteReviewModal — onsite-specific contract', () => {
-  test('a pending selection renders the View full source link, the mod Review-preview panel, and the structured manifest', async () => {
+  test('a pending selection renders the in-app code diff, the mod Review-preview panel, and the structured manifest — and NO raw-source link', async () => {
     renderWithProviders(
       <OnsiteReviewModal selection={{ request: ONSITE_PENDING, mode: 'pending' }} onClose={vi.fn()} />
     );
-    // The on-site code-review affordance (off-site has no bundle/code).
-    await expect.element(page.getByText('View full source')).toBeInTheDocument();
-    // The old "View code in Forgejo" copy is gone.
+    // The on-site code-review affordance (off-site has no bundle/code) is the
+    // in-app diff, not an external link.
+    await expect.element(page.getByText('Show code diff')).toBeInTheDocument();
+    // 🔴 #3495 regression guard: in-review snapshots are private, so ANY raw
+    // deep-link out to them is a dead link for the moderator who clicks it.
+    // None of these affordances may come back.
+    expect(page.getByText('View full source').elements()).toHaveLength(0);
+    expect(page.getByText('View in Forgejo').elements()).toHaveLength(0);
     expect(page.getByText('View code in Forgejo').elements()).toHaveLength(0);
+    expect(
+      document.querySelectorAll(`a[href="${ONSITE_PENDING.reviewRepoUrl}"]`).length
+    ).toBe(0);
     // The mod Review-preview sandbox panel — on-site pending only.
     await expect.element(page.getByText('Review preview')).toBeInTheDocument();
     await expect.element(page.getByRole('button', { name: 'Start preview' })).toBeInTheDocument();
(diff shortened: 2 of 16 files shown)


ZacxDev:
Follow-up from review — five items, all mechanical:

**1. The sweep's `approved` exclusion is now enforced, not just documented.**
The eligibility list decides which *rows* trigger a purge, but the snapshot is keyed per *slug*, and a terminal row does not reserve its slug — so the triggering row's status alone can't decide the delete. The gate now looks for `status IN ('pending','approved')` (still on the primary, for the reason already in the comment). Both sets are named constants and are an exact partition of the DB's closed status vocabulary, with a test pinning that so a future status can't fall through the middle.

**2. The gate-read → delete window is now loud instead of silent.**
Submit writes the snapshot before it inserts the row that protects it, so a submission that starts inside that window isn't visible to the gate. After a successful delete the sweep re-reads the slug and logs at error level if a `pending` row appeared. This doesn't close the race — reordering the submit path would, and that's a user-facing write path, so it's deliberately left out of this PR (noted in a comment).

**3. Cancellation.** The sweep now checks the runner's cancel hook between slugs, outside the per-slug catch so a cancel isn't absorbed as a per-repo failure.

**4. Two cursor comments corrected.** `updatedAt` is *not* frozen when a row goes terminal (the reject path writes again right after), and resuming with `gt` can skip a row sharing the batch's last timestamp. Both are benign today; the comments now say what's actually true rather than claiming a guarantee. Keying retention off a frozen decision timestamp would change which rows are eligible, so it's flagged as a follow-up rather than done here.

**5. A vacuous test assertion replaced with a real one.**
The modal suite asserted `'View in Forgejo'` is absent, but that string only ever rendered inside `FileDiffEntry` for an elided file and the code-diff panel is collapsed by default — no file row was mounted, so it passed on `main` too. The guard now lives in the diff-panels suite, which mounts `FileDiffEntry` directly: for each of the four `skipReason` values it asserts the row states the reason and renders no anchor, before and after a click. Confirmed non-vacuous by re-adding the removed per-file link — all four cases fail with it present. The modal's other two assertions are real guards and were kept.

Also refreshed two comments this PR invalidated (the elided-file description on `getPublishRequestDiff`, and a stale aside on the webhook org gate).

Verification: `src/server/jobs` + `src/server/services/blocks` unit suites 2013 → 2021 passing, with the same pre-existing local failure set before and after (2 job files that fail to collect without a generated Prisma client, plus 2 `manifest-schema-endpoint` tests). Browser-mode: the two affected files 21 → 25 passing. Each new guard was mutation-checked — reverting the gate to `pending`-only, dropping the post-delete re-check, and moving the cancel check inside the try each turn specific tests red.

Issue #3499, opened 2026-07-31
Generation resource picker: intermittent 503s render as "No models found", and pages are ~4x larger than what they display


AKharytonchyk:
Clicking **Add model** or **Swap** in the generation form calls `model.getResourceSelect`. Users report the picker takes a long time and sometimes shows nothing. Two separate problems produce that one symptom.

Measured against production `civitai.red` on 2026-07-31 with read-only GETs.

## 1. Failures claim the catalogue is empty

The endpoint returns `503 SERVICE_UNAVAILABLE` after a 10 s Meili timeout, intermittently. One run of six default opens:

| Request | Status | TTFB |
|---|---|---|
| 1 | 503 | 6.257 s |
| 2 | 503 | 10.268 s |
| 3 | 503 | 10.340 s |
| 4 | 503 | 10.258 s |
| 5 | 200 | 0.582 s |
| 6 | 200 | 0.429 s |

A later run of eight came back clean at 0.38 to 0.91 s. The distribution is bimodal: roughly 0.4 s or roughly 10.2 s, nothing between. That 10.2 s is `MEILI_RESOURCE_SELECT_TIMEOUT_MS` firing.

`ResourceHitList.tsx` checks loading, then checks empty, and has no error branch. So a 503 renders:

> No models found. We have a bunch of models, but it looks like we couldn't find any matching your query.

Commit e8c2d9caa9 opens its own message with this observation and never acts on it.

The picker also amplifies the outage. `InViewLoader` awaits `fetchNextPage`, which resolves on error rather than rejecting, because query-core only rethrows when a caller passes `throwOnError` and this one does not. The failed page never lands in `state.data`, so `hasNextPage` stays true, the sentinel stays in view, and `canLoad` flips back after 500 ms. The loader re-fires every 500 ms for as long as the backend keeps failing. Each fire holds a slot on a 500-concurrency limiter that bypasses the Meili circuit breaker, so one user parked at the bottom of a broken picker occupies a slot about 93% of the time.

## 2. Pages carry roughly 4x what they render

| Page | Uncompressed | Brotli | Renders |
|---|---|---|---|
| Checkpoint / Illustrious, 50 items | 1,562,519 B | 280,472 B | ~31 cards |
| LoRA multi-type / Illustrious, 50 items | 1,646,427 B | 281,932 B | ~33 cards |

[...]

PR #3500, opened 2026-07-31
fix(generation): add a stable sort tiebreaker to the resource picker

AKharytonchyk:
Part of #3499. Five lines of behaviour change; the ops analysis that used to ride along in this PR now lives in [a comment on #3499](https://github.com/civitai/civitai/issues/3499) so it can be triaged without holding up a merge.

## The defect

`meiliSortFor` emitted no tiebreaker, so Meili fell back to **internal document order** for ties. Index writes reshuffle that order, which lets offset pagination repeat or skip a model between pages.

That isn't theoretical: `thumbsUpCount` ties are dense enough to land on the 50-item page seam. `4315` is shared by two adjacent Illustrious checkpoints near offset 46.

## The change

`id:desc` is appended to every explicit sort. **`popularity` had the same defect, not just `relevance`** — worth noting because the obvious reading is that only the unsorted case was affected.

For `relevance` the sort is now named explicitly *when there is no query*, since that is what the `metrics.thumbsUpCount:desc` ranking rule was already doing. Verified against production that `sort=relevance` and `sort=popularity` return byte-identical ordering.

**With a query the sort stays unset**, deliberately: passing one activates Meili's `sort` ranking rule, which would outrank `words`/`proximity`/`exactness` and make text search materially worse. That asymmetry is the one thing here worth a reviewer's attention.

## No reindex needed

Both `metrics.thumbsUpCount` and `id` are already in `sortableAttributes`, so this takes effect on deploy.

## Verified

- Full-repo typecheck: 0 errors (`main` is also 0).
- eslint + prettier clean.
- Rebased on current `main`; `meiliSortFor` there still has no tiebreaker, so the defect is live.

## Not verified

No production query run post-change — the byte-identical `relevance`/`popularity` finding above was measured against prod *before* the fix. Worth confirming after deploy that paging through a picker twice returns the same set, which is the actual symptom.

### DIFF
--- a/src/server/services/resource-select.service.ts
+++ b/src/server/services/resource-select.service.ts
@@ -44,15 +44,25 @@ function skipBaseModelForOwnTabs(
   return (tab === 'mine' || tab === 'official') && selectSource === 'modelVersion';
 }
 
-function meiliSortFor(sort: GetResourceSelectInput['sort']): string[] | undefined {
+// `id:desc` is a stable final tiebreaker: without one Meili breaks ties on internal
+// document order, which index writes reshuffle, so offset pagination repeats or skips
+// models across pages. thumbsUpCount ties are dense enough to collide at the page seam.
+//
+// A text query must keep `sort` unset — an explicit sort activates the `sort` ranking
+// rule, which would outrank words/proximity/exactness. With no query those are no-ops
+// and `metrics.thumbsUpCount:desc` already decides the order, so naming it is a no-op.
+function meiliSortFor(
+  sort: GetResourceSelectInput['sort'],
+  hasQuery: boolean
+): string[] | undefined {
   switch (sort) {
     case 'popularity':
-      return ['metrics.thumbsUpCount:desc'];
+      return ['metrics.thumbsUpCount:desc', 'id:desc'];
     case 'newest':
-      return ['createdAt:desc'];
+      return ['createdAt:desc', 'id:desc'];
     case 'relevance':
     default:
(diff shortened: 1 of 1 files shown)


AKharytonchyk:

---

AKharytonchyk:
Correction to the note above: the rebase was clean, but I did update `docs/features/resource-select-picker.md` for paths that moved on `main`. Re-verified every claim in it rather than assuming:

- `filterableAttributes` moved to `src/server/search-index/filterable-attributes.ts`. The claim holds — `modelsFilterableAttributes` lists the model-level `canGenerate` and **not** `versions.canGenerate`, so Meili still cannot be asked the per-version question.
- The setup-never-runs claim holds and is now more concrete: `base.search-index.ts:358` is the reset path, the incremental call is still commented at `:414`, and `search-index-sync-models-reset` is created with `UNRUNNABLE_JOB_CRON` (`0 0 5 31 2 ?` — Feb 31st), so it only ever runs by hand.
- `sortMetrics` and `isOfficial` are still absent from `displayedAttributes`, so the drift is real.
- `MEILI_RESOURCE_SELECT_TIMEOUT_MS` is still `10_000` and `MEILI_RESOURCE_SELECT_CONCURRENCY` still `500`, so the tuning suggestion stands.
- `meili_call_timeouts_total{backend="resourceSelect"}` still exists and still has no alert.

The sort fix itself is unchanged: `main` has no tiebreaker in `meiliSortFor`.

PR #3508, opened 2026-07-31
fix(app-blocks): page through the source repo tree instead of throwing on apps over 1000 files

ZacxDev:
## The mismatch

Submit-time validation accepts an app of up to `MAX_FILES_IN_BUNDLE` = **2000** files
(`src/server/schema/blocks/publish-request.schema.ts`).

`listRepoTreeAtRef` (`src/server/services/blocks/forgejo.service.ts`) read the source
host's recursive tree listing as a **single page** and hard-threw when the host flagged
the response as truncated — which it does past **1000** entries:

```ts
if (tree.truncated) {
  throw new Error(`... is truncated (>1000 entries); pagination not implemented`);
}
```

So an app between **1001 and 2000 files** passes submit and then breaks everything
downstream that reads its tree.

## Blast radius

- `reconstructBundleFromForgejo` → `approveRequest` for git-push submissions,
  `getPublishRequestDiff`, and the review-preview mirror path
- `listRepoTree` → `commitFiles`, and the review-preview short-circuit added in #3496

Net effect: **a submission over 1000 files could not be approved, diffed, or previewed** —
a hard throw on the *approve* path, i.e. a confusing dead end for both the developer and
the moderator.

Currently latent: no app has crossed 1000 files yet. This fixes it before the first one does.

## What the fix does

Replaces the throw with a real pager over the tree endpoint's `page` / `per_page` params.

The stop condition is deliberately **not** "a short page means the last page". The host is
free to clamp `per_page` below what we ask for, which would make the very first page look
short and end the walk early — silently returning an incomplete tree, which downstream
reads as "these files were deleted". Instead, in priority order:

1. **an empty page** → nothing left;
2. **`total_count`** → authoritative entry total when present; stop once we have seen it all
   (also saves the trailing empty-page round-trip);
3. **`truncated`** → fallback when a response carries no `total_count`: keep paging while the
   flag says the tree overflows a page.

None of those assume a page size, so a server-side clamp can't truncate the walk.

Unchanged for trees under the cap: page size stays at the endpoint's own **1000-entry
ceiling** (a larger `per_page` is clamped down to it anyway), so a tree that fits in one page
still costs **exactly one request** and returns the same blob-only `Map<path, sha>`, in the
same order, with the same error handling. `listRepoTree` delegates to `listRepoTreeAtRef`,
so it gets the fix for free.

A runaway guard bounds the loop — derived from `MAX_FILES_IN_BUNDLE` rather than a new magic
number (files, plus a directory entry per folder, plus headroom) — and if a repo genuinely
exceeds it, throws an error naming the real 2000-file limit and what to do.

Not touched: `MAX_FILES_IN_BUNDLE`, any submit-time validation, `approveRequest`'s logic, the
diff, or the preview mirror. They simply stop breaking. No migration.

## Verified

- **Mutation check.** The natural failure here is a pager that silently stops early, so both
  variants were introduced deliberately and confirmed to fail:
  - stop after the first page → **3 of the new tests fail** (multi-page walk, no-`total_count`
    fallback, runaway bound)
  - treat a short page as the last page → **the same 3 fail**
  - reverted; suite green again.
- **New tests**: a 3-page walk with a **final partial page**, asserting entries from every page
  *including the last*, plus the `page=` increments and that `total_count` avoids a 4th request;
  the no-`total_count` (`truncated`) fallback terminating on an empty page; a `null` tree past
  the end of the listing; the single-page fast path still costing exactly one request; and the
  runaway bound throwing with the file limit named. Pages in the tests are intentionally
  *smaller* than the requested `per_page`, so a short page can never be mistaken for the end.
- **Suites** (`vitest --project unit`):
  - `src/server/services/blocks` — before **1846 passed / 2 failed**, after **1850 passed / 2 failed**
    (+4 = 5 new tests, 1 obsolete "throws when truncated" test removed)
  - `src/tests/api/internal/blocks` + `blocks.router.getMyAppRepo` + `purge-review-snapshots` —
    **191 passed** before and after
  - The 2 failures are **pre-existing and identical on `main`** in the same worktree: a local
    `Cannot find package 'kysely'` module-resolution error in `manifest-schema-endpoint.test.ts`
    (package not installed locally). Not touched by this change.
- **Typecheck**: full-repo `tsc --noEmit` gives **16 errors on this branch and 16 on `main`** —
  identical, all pre-existing missing-module/implicit-any noise from packages that aren't
  installed locally. **Zero** errors in the touched files.
- **Lint**: eslint clean on both touched files. (Prettier reports pre-existing formatting drift
  in both files on `main` too; deliberately not reformatted, to keep the diff readable.)

## Not verified

- **No end-to-end run against a real >1000-file app.** The pagination is exercised against a
  mocked multi-page response only — no such app exists yet, which is exactly why the bug is
  latent. The response-field assumptions (`page`, `per_page`, `truncated`, `total_count`, and a
  null `tree` past the end) come from the endpoint's documented behaviour, not from a live call.
  The stop logic is written so that any one of those fields being absent still terminates
  correctly, but the first real large submission is the true confirmation.
- CI is the real gate for the full test + build matrix; only the suites listed above were run
  locally.

### DIFF
--- a/src/server/services/blocks/__tests__/forgejo.service.test.ts
+++ b/src/server/services/blocks/__tests__/forgejo.service.test.ts
@@ -225,17 +225,326 @@ describe('listRepoTreeAtRef pagination', () => {
     expect(treeCalls()).toHaveLength(1);
   });
 
-  it('gives up with a limit-naming error rather than paging forever', async () => {
+  it('gives up rather than paging forever when the host never says it is done', async () => {
     const { listRepoTreeAtRef } = await import('../forgejo.service');
-    // A server that always claims there is more: the pager must bail out at the
-    // bound instead of looping.
-    for (let i = 0; i < 20; i++) {
+    // A server that always claims there is more, one entry at a time: the pager
+    // must bail out at the REQUEST backstop instead of looping. The entry bound
+    // (8000) is nowhere near reached, so the error must name the real reason
+    // rather than blaming the app's file count.
+    for (let i = 0; i < 200; i++) {
       fm.enqueue({ tree: [{ path: `f${i}.txt`, type: 'blob', sha: `s${i}` }], truncated: true });
     }
 
-    await expect(listRepoTreeAtRef('runaway', 'ref_sha')).rejects.toThrow(/at most 2000 files/);
-    // Bounded: MAX_FILES_IN_BUNDLE * 4 entries / 1000 per page = 8 pages.
+    await expect(listRepoTreeAtRef('runaway', 'ref_sha')).rejects.toThrow(
+      /did not finish within 64 requests/
+    );
+    expect(treeCalls()).toHaveLength(64);
+  });
+
+  // --- The runaway bound must be on ENTRIES READ, not on page count. ---------
+  // Deriving the request cap as MAX_TREE_ENTRIES / TREE_PAGE_SIZE assumes the
+  // host honours the per_page we ask for — the exact assumption the stop
+  // condition refuses to make. Under a host that clamps per_page, that pairing
+  // rejects a legal tree with a message about a limit it never approached.
+  it('reads a legal tree to completion even when the host clamps per_page far below the request', async () => {
+    const { listRepoTreeAtRef } = await import('../forgejo.service');
+    // 1500 entries — comfortably legal (under both 2000 files and 8000 entries)
+    // — served 100 at a time because the host clamped per_page.
+    const CLAMPED = 100;
+    const TOTAL = 1500;
+    for (let page = 0; page * CLAMPED < TOTAL; page++) {
+      fm.enqueue({
+        tree: Array.from({ length: CLAMPED }, (_, i) => ({
+          path: `f${page * CLAMPED + i}.txt`,
+          type: 'blob',
+          sha: `s${page * CLAMPED + i}`,
+        })),
+        truncated: true,
+        total_count: TOTAL,
+      });
+    }
+
+    const result = await listRepoTreeAtRef('clamped', 'ref_sha');
+    expect(result.size).toBe(TOTAL);
+    expect(result.get('f0.txt')).toBe('s0');
+    expect(result.get('f1499.txt')).toBe('s1499');
+    expect(treeCalls()).toHaveLength(TOTAL / CLAMPED);
+  });
+
+  it('names the entry bound — not the request bound — when the tree is genuinely oversized', async () => {
+    const { listRepoTreeAtRef } = await import('../forgejo.service');
+    // Full 1000-entry pages of DIRECTORY entries, so the entry bound (8000) is
+    // what trips rather than the blob/file cap.
+    for (let page = 0; page < 12; page++) {
+      fm.enqueue({
+        tree: Array.from({ length: 1000 }, (_, i) => ({
+          path: `d${page}/${i}`,
+          type: 'tree',
+          sha: `st${page}_${i}`,
+        })),
+        truncated: true,
+      });
+    }
+
+    await expect(listRepoTreeAtRef('huge', 'ref_sha')).rejects.toThrow(/exceeds 8000 entries/);
+    // 8000 entries / 1000 per page = 8 requests, then the entry bound stops it.
     expect(treeCalls()).toHaveLength(8);
+  });
+
+  // --- An empty response BODY is an error, not a past-the-end marker. --------
+  // `unwrap` returns null for an empty body. Treating that the same as an
+  // envelope with a null `tree` would return a silently PARTIAL tree, which
+  // downstream becomes an incomplete approved bundle / a mirror that deletes
+  // the files it failed to read.
+  it('throws instead of silently truncating when a page comes back with an empty body', async () => {
+    const { listRepoTreeAtRef } = await import('../forgejo.service');
+    fm.enqueue({
+      tree: [
+        { path: 'p1-a.txt', type: 'blob', sha: 's1a' },
+        { path: 'p1-b.txt', type: 'blob', sha: 's1b' },
+        { path: 'p1-c.txt', type: 'blob', sha: 's1c' },
+      ],
+      truncated: true,
+      total_count: 9,
+    });
+    // Empty 200 mid-walk: 3 of 9 entries read. Must NOT resolve.
+    fm.enqueue(null);
+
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Audit follow-ups pushed (`51f8c43`)

Adversarial audit confirmed the core pager is correct (population counted matches `total_count`, no torn read — both callers pass an immutable commit sha, sub-cap parity holds, `listRepoTree` delegates). Five findings; four applied, one documented.

### 1. The guard bounded tree ENTRIES, but the error promised a FILE limit — HIGH

`MAX_TREE_ENTRIES` bounds blobs **+ directories**; nothing bounded blobs. A probe returned `result.size === 7000` with no error, from a function whose error text says *"an app may contain at most 2000 files"*.

This mattered more than a wrong message, because the old `truncated` throw was an **accidental enforcer**. On the git-push path `MAX_FILES_IN_BUNDLE` is otherwise unenforced:

- `enrichPushRequestRow` wraps `computePushDiffSummaries` (which applies the cap) in a `try/catch` that **only logs** — the row still parks as `pending`.
- `extractBundleFilesFromBuffer`, used by `approveRequest`, caps total **bytes** but has **no file-count cap**.

So the pre-PR throw was incidentally enforcing a ceiling at 1000 entries, and paging raised it to 8000. Now capped explicitly on `result.size` (blobs).

**Caller enumeration** (the gate before capping — all app-repo-scoped, so none legitimately needs more than the cap):

| Caller | Path | Scope |
|---|---|---|
| `reconstructBundleFromForgejo` | `listRepoTreeAtRef` direct | approve (push-originated) + backfill; app repo at a commit sha |
| `commitFiles` | → `listRepoTree` → `listRepoTreeAtRef` | canonical app repo or review mirror |
| `reviewRepoAlreadyHoldsTree` | → `listRepoTree` | review mirror of one app; already `try/catch` → fails toward re-mirroring |

Every `commitFiles` call site passes a single app slug against the canonical org or the review org. An app repo's contents **are** the bundle, which submit-time validation caps at `MAX_FILES_IN_BUNDLE`. Gate passed.

### 2. The runaway bound assumed a page size the stop condition refuses to assume — MEDIUM

`MAX_TREE_PAGES = ceil(8000/1000) = 8` assumed the host honours `per_page`. Probe with `per_page` clamped to 100 on a **1500-entry tree**: 8 requests, 800 entries read, then *"exceeds 8000 entries"* — a legal tree rejected against a ceiling it never approached. Fails loud, so a false ceiling rather than corruption; not reachable today, but one config edit away.

Now bounds `seen` against `MAX_TREE_ENTRIES`, with a separate, much larger `MAX_TREE_REQUESTS = 64` kept purely as an anti-infinite-loop backstop. The throw names whichever bound tripped — *"too big"* and *"the host is paginating in unexpectedly small pages"* are different operator problems.

### 3. An empty-bodied 200 mid-walk silently truncated the tree — MEDIUM

`unwrap` returns `null` for an empty body. The pager collapsed that with the host's legitimate past-the-end `tree: null` marker via `?? []` — both hit `entries.length === 0 → return`. Probe (pages of 3 / empty-200 / 3, `total_count: 9`): **returned 3 of 9, no error, no log**. Pre-PR the same response crashed loudly.

Downstream the partial map is treated as complete — `reconstructBundleFromForgejo` makes an incomplete bundle the approved artifact with a `bundleSha256` over the partial set, and `commitFiles(replaceAllFiles: true)` emits deletes only for the paths it read, leaving stale files behind. Now throws on a missing envelope; the past-the-end case (and its test) still works.

### 4. Closed the last fail-silent branch

An exactly-full page now always warrants one more probe, before the `total_count` / `truncated` checks. This removes both remaining silent under-reads: a `total_count` that under-reports (previously trusted absolutely) and a response carrying neither field. It only ever *continues*, never stops, so it cannot resurrect the short-page bug the PR guards against.

**Cost:** a tree whose size is an exact multiple of the page size now pays one confirming round-trip — `total_count` still saves the trailing request in the normal partial-final-page case.

### 5. Latency — documented, not fixed

Worst case is now 8 sequential fetches × 15s ≈ **120s**, vs ~15s pre-PR. Worth knowing; not worth blocking.

---

### Test gaps closed

Both of these mutants **survived** the previous suite:

- **`total > 0` guard was completely untested.** Added a `total_count: 0`-with-entries case asserting the loop falls through to the `truncated` fallback.
- **`seen >= total` was only ever hit exactly on the nose** (fixtures step 3+3+2 onto 8), so an off-by-one on the low side was invisible. ⚠️ An *overshooting* fixture (3+3+3 vs 8) does **not** discriminate either — both forms stop on the same page. The shape that does is a page leaving `seen` **exactly one short** of `total` with a further page to come: **3+4+1 against `total_count: 8`**. Added that, plus the overshoot case separately.

Also added: per-fix cases for all four fixes, a clamped-`per_page` case, and file-cap cases (over cap, exactly at cap, blobs-not-directories, and the **sub-cap happy path asserting it is still exactly one request with an identical result**).

### Mutation results — 10 mutants, 0 survivors

| Mutant | Before | After |
|---|---|---|
| `total > 0` guard removed | **SURVIVED** | KILLED |
| `seen >= total` → `seen + 1 >= total` | **SURVIVED** | KILLED |
| stop after page 1 | KILLED | KILLED (14) |
| short page treated as last | KILLED | KILLED (10) |
| file cap removed | — | KILLED |
| file cap counts entries not blobs | — | KILLED |
| request bound = `MAX_TREE_ENTRIES / TREE_PAGE_SIZE` | — | KILLED |
| `seen < MAX_TREE_ENTRIES` loop bound dropped | — | KILLED |
| empty body treated as past-the-end | — | KILLED |
| exactly-full-page probe dropped | — | KILLED |

### Suites

| Suite | Before | After |
|---|---|---|
| `forgejo.service.test.ts` | 25 passed | **37 passed** |
| `src/server/services/blocks` (95 files) | 1850 passed / 2 failed | **1862 passed / 2 failed** |
| `src/tests/api/internal/blocks` | — | 163 passed |

The 2 failures are pre-existing and unrelated — `manifest-schema-endpoint.test.ts` fails locally on a missing `kysely` package (NixOS, no generated client). Identical before and after.

`tsc --noEmit`: **zero errors in the touched files**. The run exits non-zero only on pre-existing local-env breakage (missing `kysely`, and an `event-engine-common` sibling that isn't checked out locally) — CI is the gate.

PR #3511, opened 2026-07-31
fix(challenge): stop a slow metadata write from clobbering a challenge claimed for completion

ZacxDev:
## Problem

Two write paths read a challenge from the **read replica**, spend seconds in an LLM call (`generateThemeElements`), and then write `metadata` back keyed only on `id`:

- `src/pages/api/mod/daily-challenge/backfill-theme-elements.ts`
- `upsertChallenge` (`challenge.service.ts`), update branch

If `claimChallengeForCompletion` flips the challenge to `Completing` inside that window — stamping `metadata.completingClaimedAt` as it goes — the write restores the **pre-claim snapshot** on top of the fresh stamp. The challenge is then `Completing` with no `completingClaimedAt`, and the stuck-challenge reset propagates NULL so it never selects that row: **the challenge wedges permanently.**

`?force=true` widens the backfill's exposure from "challenges missing theme elements" to *every* themed `Active`/`Scheduled` challenge at once.

`upsertChallenge` is worse in one respect: its `Active` branch pins `data.status` back to the status it **read**, so the unconditional write also sets a mid-completion challenge back to `Active` — un-claiming it outright.

## Fix

**Backfill endpoint** — predicate the `UPDATE` on `status IN ('Active','Scheduled')`, and merge the patch onto the *current* stored value with `jsonb ||` rather than replacing the whole column. The predicate closes the wedge; the merge additionally stops any *other* metadata field written during the generation window from being clobbered, even when the status never changed. A zero-row result is reported as `skipped`, not counted as backfilled.

```sql
SET metadata =
  CASE WHEN jsonb_typeof(metadata) = 'object' THEN metadata ELSE '{}'::jsonb END
  || $patch::jsonb
WHERE id = $id
  AND status IN ('Active', 'Scheduled')
```

**`upsertChallenge`** — convert the unconditional `update` to a status-predicated `updateMany` + `count === 0`, matching [the Scheduled-only edit path already in this service](https://github.com/civitai/civitai/blob/main/src/server/services/challenge.service.ts), then re-read the row to return it.

## Tests

8 new (5 endpoint, 3 service), plus the existing judging-categories suite updated for `update` → `updateMany`.

Both guards are **mutation-verified**: removing either status predicate fails 3 of the new tests. (Checked explicitly rather than assumed — a guard whose test still passes with the guard removed isn't testing anything.)

## Verification

- `vitest --project unit challenge` → **627/627 pass** (71 files)
- `eslint` clean on all five changed files
- `prettier` clean on the four it was already clean on. `challenge.service.ts` is unformatted on `main`; the change introduces **no new hunks** (formatter hunk set is identical to baseline, offset only by added lines)
- `tsc --noEmit` output is **byte-identical to pristine `origin/main`**, which is currently red on an unrelated `ModelVersionUpsertForm` error (`freeGeneration`). So this branch adds no type errors — and the `updateMany` data shape typechecks.

## Notes

- Behaviour change: a save that races a completion now returns `PRECONDITION_FAILED` ("Challenge status changed while saving. Reload and try again.") instead of silently winning. That is the intended trade — the previous silent win corrupted the challenge.
- The backfill no longer rewrites metadata fields it did not intend to change, so the endpoint stops normalising the column as a side effect of running.

### DIFF
--- a/src/__tests__/pages/api/mod/daily-challenge/backfill-theme-elements.test.ts
+++ b/src/__tests__/pages/api/mod/daily-challenge/backfill-theme-elements.test.ts
@@ -12,11 +12,14 @@ import type { NextApiRequest, NextApiResponse } from 'next';
 // reset propagates NULL and so never selects such a row: the challenge wedges permanently.
 // `?force=true` widens the exposure to every themed Active/Scheduled challenge at once.
 
-const { mockQueryRaw, mockExecuteRaw, mockGenerateThemeElements } = vi.hoisted(() => ({
-  mockQueryRaw: vi.fn(),
-  mockExecuteRaw: vi.fn(),
-  mockGenerateThemeElements: vi.fn(),
-}));
+const { mockQueryRaw, mockExecuteRaw, mockGenerateThemeElements, mockLogToAxiom } = vi.hoisted(
+  () => ({
+    mockQueryRaw: vi.fn(),
+    mockExecuteRaw: vi.fn(),
+    mockGenerateThemeElements: vi.fn(),
+    mockLogToAxiom: vi.fn(),
+  })
+);
 
 vi.mock('~/server/db/client', () => ({
   dbRead: { $queryRaw: mockQueryRaw },
@@ -32,7 +35,7 @@ vi.mock('~/server/games/daily-challenge/generative-content', () => ({
   generateThemeElements: mockGenerateThemeElements,
 }));
 
-vi.mock('~/server/logging/client', () => ({ logToAxiom: vi.fn() }));
+vi.mock('~/server/logging/client', () => ({ logToAxiom: mockLogToAxiom }));
 
 vi.mock('~/server/schema/challenge.schema', () => ({
   parseChallengeMetadata: (m: unknown) => (m ?? {}) as Record<string, unknown>,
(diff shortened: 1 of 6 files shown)


ZacxDev:
## Corrections to the PR description above, after an adversarial audit

Pushed `46a8c950af`. Four things in the original description were wrong or overstated — flagging them here rather than quietly editing the body, since they may already have been read.

**1. My mutation verification was incomplete — the audit found a mutation that survived all five endpoint tests.**

I claimed the guards were mutation-verified. They were, but only against *removing the status predicate*. The merge assertions pinned the **presence** of `||` and `jsonb_typeof(...)`, not their **arrangement** — so swapping the `||` operands to `patch || stored` passed every test. That swap inverts the merge so the stored value wins, which would make `?force=true` a silent no-op for any challenge that already has `themeElements`. The shipped SQL was always correct (independently validated against a real Postgres), but the suite would not have caught a regression in the exact semantics it was written to protect. Fixed: the test now pins the patch as the right operand, and I re-ran that mutation to confirm it fails.

**2. "Removing either status predicate fails 3 of the new tests" — that's the sum, not what "either" implies.** Measured individually it is 2 (endpoint) and 1 (service).

**3. The two fixes are not as symmetric as the description implies.** The status predicate stops `upsertChallenge` landing on a *different* status, but that path still writes the **whole** metadata column from its replica snapshot (`{ ...existingMetadata, themeElements }`). A *same-status* racing write can therefore still be dropped — concretely, the `reviewedAt` watermark written on Active challenges, which on loss rewinds incremental review to the challenge start and re-judges the backlog at LLM cost. This is pre-existing rather than a regression, and the exposure is routine rather than rare, since the edit form pre-fills `themeElements` so the replace runs on most mod saves. **Not fixed here** — converting that write to the same `jsonb ||` merge is the remaining work, and it is a larger change than this PR's scope. `challenge.metrics.ts` now records it explicitly instead of claiming the class is closed.

**4. "627/627 pass" is accurate but reads cleaner than the output.** The run also emits `Errors 1` — an unhandled Prisma engine-load error originating in `challenge-currency.test.ts`. Environmental and pre-existing on `main`, unrelated to this change.

### Also addressed in `46a8c950af`

- `challenge.metrics.ts:465-477` asserted that predicating these two writes "is a separate change and **has not been made**" — this PR makes exactly that change to exactly those two sites. That comment is the canonical explanation for why the gauge treats a stampless `Completing` row as stuck, so leaving it stale would actively misinform. Updated.
- Both abort paths were silent. For guards whose entire purpose is a rare race, there was no way to tell whether either had ever fired. Each now logs to Axiom; the endpoint also counts skips and returns them, so `backfilled + failures` no longer falls short of `total` unexplained.
- The `PRECONDITION_FAILED` message said "Reload and try again", which is misleading advice: `challenge.status` comes from the read **replica**, so the guard also fires on replication lag — where a reload re-reads the same lagging replica and cannot clear the condition. It also fires when the row was deleted outright, where "status changed" is the wrong diagnosis. Reworded, both cases noted in comments.

### Verification unchanged after the fixes

627/627 challenge tests · eslint clean on all six files · prettier clean on the five it applies to (`challenge.service.ts` is unformatted on `main`; hunk count still 6, identical to baseline) · `tsc --noEmit` byte-identical to pristine `origin/main`.

ZacxDev:
## Correction: the "remaining work" note in this PR overstated its own consequence

Follow-up in **#3571** (comments only).

This PR's description, my earlier comment on it, and both code notes said the un-fixed full-column metadata write in `upsertChallenge` mattered because losing the `reviewedAt` watermark **"rewinds incremental review to the challenge start and re-judges the whole backlog at LLM cost"**, uncapped for paid user challenges.

**That is wrong.** The consuming query in `daily-challenge-processing.ts` carries, in the same `WHERE` clause as the watermark, `notYetReviewedByJudge` — a `NOT EXISTS` on the judge's comment for that image — which excludes every already-scored entry *regardless of the watermark*. Losing the watermark widens an index scan; it re-judges nothing and costs no LLM calls.

The probability was overstated as well: the only concurrent writer of that column is that watermark, running every 10 minutes against the **one** current challenge, so a loss needs a save of that specific challenge to straddle a 600-second-cadence instantaneous `UPDATE` inside a sub-second window on the common (no-LLM) path.

Net effect: the jsonb-merge conversion is now recorded as **assessed and deliberately declined**, reopen-on-evidence, instead of pending work carrying a price tag that would have justified a second raw-SQL write inside a transaction on a mod write path.

Nothing about what this PR shipped changes — the guards are unaffected, and the 7h post-deploy soak was clean on the `upsertChallenge` guard across 3 `Scheduled → Active` activations and 4 completions. (The backfill guard's zero was *vacuous* — that endpoint was never invoked in the window, so it remains unexercised in production.)

For provenance: the claim came out of the adversarial audit of this PR and I propagated it without checking the consuming query. The audit correctly verified that the watermark falls back to challenge start and that the predicate uses it; neither of us checked that the same clause independently dedupes.

PR #3119, opened 2026-07-14
fix(app-blocks): bound REST app_block_id for dev tokens + add error_class render label

ZacxDev:
Two cardinality/obs-quality fast-follows to the App Blocks runtime metrics
(#3117):

1. REST app_block_id — dev-token bucketing. block-scope.middleware recorded
   `app_block_id: claims.appBlockId` raw. A DEV token (claims.dev === true — the
   same discriminator the max-age cap + audit path use) carries a caller-
   constructed synthetic, non-resolving appBlockId (dev-scoped-mint.service) —
   an unbounded label vector even though minting is mod/dev-cohort gated. Bucket
   all dev tokens to a single stable 'dev' label; real per-app attribution (incl.
   legit pre-approval apps) is preserved. Metric-label only; authz/audit untouched.

2. renders_total error_class label. The render counter collapsed every failure
   into result="error". The host already sends an errorClass (timeout | fatal |
   no_token | error | error_boundary) that was parsed then stripped before the
   prom emit. Surface it as a STRICTLY-enumerated `error_class` label:
   result=ok → 'none'; on error the sent class if in KNOWN_ERROR_CLASSES, else
   'other'. Never lets a raw/unbounded errorClass become a label. status/
   errorClass STILL stay out of the ClickHouse insert (byte-identical). No client
   change needed — the hosts already send exactly these values (verified).

renders_total label set is now {app_block_id, slot_id, result, error_class},
all strictly bounded.

Tests: beacon (ok→none, each known class preserved, unknown→other, CH insert
still strips status/errorClass); middleware (dev token → 'dev', normal → real
appBlockId). block-render + middleware + metrics + service suites green;
tsc --noEmit clean on touched files.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #3516, opened 2026-07-31
fix(app-blocks): emit a render beacon when a ready page loses its credential mid-session

ZacxDev:
## Summary

App Blocks render failures had a blind spot: a page host that **already reached `ready`** and then lost its credential produced **no telemetry at all**. This PR closes it.

It is deliberately small. The work started as two changes; the second turned out to be **already shipped**, so this is **one real fix plus a stale-comment correction** — see "Change B" below rather than padding.

## Measured evidence

- A real revocation teardown was driven end-to-end against a live app on production on **2026-07-31**. The platform recorded **zero** error beacons for it. Its only record of the incident was the earlier successful `ok` impression — so every render metric read "healthy" while the app was dead in the viewer's tab.
- The render alert fires correctly on real data today: a genuine render failure earlier the same day produced `{result="error", error_class="timeout"}`, and `sum by(app_block_id)(max_over_time(civitai_app_block_renders_total{...,result="error",...}[24h])) > 0.5` fired on it.

## Change A — emit a beacon on mid-session credential loss (the real work)

`PageBlockHost` has an effect gated on `tokenTerminal && !token` that transitions a host which already reached `ready` into `error` (delist / suspend / revoke → the mint settled on a terminal 4xx with nothing usable left). Its own comment said the failure beacon was *"inert here by construction"* — and that was accurate:

> one-beacon-per-mount is enforced by a **single** emit-once ref (`blockRenderEmittedRef`) shared by the `ok` impression and the launch-failure beacon. A host that reached `ready` has already spent that ref on `ok`, so the failure beacon can never fire.

**Relaxing that ref would be the wrong fix.** It encodes an analytics invariant — exactly one impression per host mount — that the `blockRenders` denominator and the `BlockRenderFailureRate` alert both depend on. Clearing it would retroactively change what the already-sent `ok` means.

So the teardown gets its **own** at-most-once latch (`midSessionLossEmittedRef`) and its **own** error class. A revoked session now emits:

| beacon | meaning |
|---|---|
| `ok` | the page load succeeded — which it did |
| `error{error_class="token_lost_midsession"}` | that session was later revoked |

Two different events, not one counted twice. **Impression accounting is byte-identical.**

A second ref, `reachedReadyRef`, latches on the first committed `ready`. It is what tells a `ready → error` teardown apart from a `loading → error` launch failure — `status === 'error'` alone cannot, since both transitions land on it.

🔴 **Scope note — this closes the credential-loss blind spot, not every mid-session blind spot.** The gate is `status === 'error'`, so a host that reached `ready` and then *crashed* (`BLOCK_ERROR{fatal:true}` → `fatal`) still emits nothing beyond its `ok`. That is deliberate — a crash is a different failure with its own class, and mis-tagging it would make `token_lost_midsession` lie — but it means "a mid-session teardown is now observable" is true specifically for **credential loss**. The existing test *"a mount that already reported `ok` never additionally reports `error` after a crash + failed recovery"* still passes, confirming the boundary. Mid-session crash reporting is a reasonable follow-up.

🔴 **The latches are scoped to the block INSTANCE, not the React mount** (third commit, from audit). `/apps/run/[slug]` renders this host with no `key`, so an appA → appB soft navigation reuses the instance. Left unscoped, app B's *launch* failure inherited app A's `reachedReady` and was emitted as `token_lost_midsession` **for an app that never launched**. Mutation-verified as a real, reachable defect: removing the reset makes the new test fail with `expected [ { appBlockId: 'apb_other', … } ] to deeply equal []`. The wider host-reuse breakage (stale `status`; app B losing its `ok` impression to the leaked `blockRenderEmittedRef`) is **pre-existing and deliberately not fixed here** — the real fix is `key={blockInstanceId}` on the run page, which would also change impression *counts*, so it belongs in its own PR.

Decision logic is extracted into the pure, node-unit-testable `shouldEmitMidSessionLossBeacon` (following the established `blockTokenRetry.ts` / `pageBlockHostLogic.ts` / `hostRenderDecision.ts` pattern), because the `unit` vitest project runs `environment: 'node'` and only collects `*.test.ts`.

`token_lost_midsession` is added to the server-side `KNOWN_ERROR_CLASSES` allowlist. Without that the beacon route would clamp it to `other` and the signal would merge straight back into the generic bucket — the value would be computed, sent, accepted, and silently dropped.

### 🔴 Second commit: keeping the ClickHouse impression table at one row per mount

**Found in review of this PR's own second-order effects — a defect this PR introduced, caught before merge.**

`/api/track/block-render` strips `status`/`errorClass` and then calls `tracker.blockRender(...)` **unconditionally**, so every beacon writes one `blockRenders` ClickHouse row *that carries no status*.

- **Before:** one beacon per mount ⇒ one row. Clean 1:1.
- **After Change A:** a revoked session emits `ok` **and** `error` ⇒ **two byte-identical rows for one mount.**

Being identical, they cannot be de-duplicated afterwards — this would have silently inflated `civitai_app_blocks_impressions_24h`, `renders_by_app_24h`, the digest CronJob and the author-analytics tab, for *exactly* the sessions that suffered a revocation. An observability fix quietly corrupting an analytics series.

The prom side was already correct (`blockRenderEmittedRef` untouched); only the CH insert regressed.

**Fix:** `blockRenders` is an **impression** table, and a mid-session credential loss is a status *update* about an impression already counted — not a new impression. So the emitter marks the follow-up (`secondary: true` in `blockRenderSchema`) and the route skips the insert, returning before the session resolve that the insert was the only reason to pay for. **The prom counter still fires for both beacons** — that is what the alert reads.

🔴 **The discriminator is the flag, not `status === 'error'`.** A **launch failure is a mount's only beacon and must still write its row** — it represents a real attempted render. Only a follow-up to an already-reported mount is suppressed. Chosen over hard-coding the `token_lost_midsession` value so it generalizes to any future second beacon.

The legacy `track.blockRender` tRPC procedure gets the same suppression, so a bearer/API-key caller cannot reintroduce the double-count.

*(Precision, since an earlier draft overstated it: the tRPC procedure **never** touched the prom counter, before or after — it only ever wrote the CH row. So for a `secondary` beacon that path is a total no-op, not "prom-only". No impact today: every host uses `sendBlockRender` → the REST beacon. Both `secondary` branches are now pinned by tests.)*

### `IframeHost` (model-slot host) — intentionally unchanged

The equivalent is **moot there, structurally** — not skipped:

- `IframeHost` takes `token: string` (non-nullable) and has **no `tokenTerminal` / `tokenError` prop**; its `Status` union is `'loading' | 'ready' | 'timeout' | 'fatal' | 'no_token'` — there is **no `error` state and no ready→error transition to observe**.
- Mid-session terminal loss is handled one level up: `BlockHost` does `if (terminal) return null`, which **unmounts `IframeHost` entirely** (the deliberate collapse-on-terminal behaviour — an error card for a block the viewer never asked for, on someone else's model page, is worse than silence).

So there is no host state change to hang an effect on. The only way to report it would be an **unmount flush**, which is a different change with a real design question attached: an unmount cleanup cannot, by itself, tell a terminal teardown from an ordinary navigation away from the model page, so a naive version would emit an error beacon on every page navigation. `sendBlockRender` already sets `keepalive: true` precisely so such a flush is possible later; the existing code comments already flag "unmount during the sequence emits nothing" as a known, accepted cost. Out of scope here.

## Change B — `errorClass` as a Prometheus label: **already shipped, no code change**

The brief for this PR was based on the comment at `sendBlockRender.ts` which read *"reserved for a future ClickHouse column (NOT a prom label, NOT in the CH insert today)"*. **That comment is stale.** `error_class` has been a live label on `civitai_app_block_renders_total` since #3119.

Confirmed against production dp-1 (`/api/v1/series`, last 24h) — the live label set is:

```
__name__, app_block_id, cluster, container, error_class, instance, job,
namespace, node, pod, result, service, slot_id
```

with real series `{result="ok", error_class="none"}` and `{result="error", error_class="timeout"}`.

**So no label was added, and nothing was reverted — there is no duplicate/conflicting label definition in this PR.** What this PR does instead:

1. **Corrects the stale comment.** It now states what is true: `errorClass` drives the `error_class` prom label, and it is clamped server-side.
2. **The ClickHouse half is now stated as *verified*, not inherited.** Both writers (`/api/track/block-render` and the `track.blockRender` tRPC procedure) destructure `status`/`errorClass` **out** before handing the rest to the tracker, and `Tracker.blockRender`'s parameter type is `{appBlockId, blockInstanceId, slotId, isAnon}` — no field for either. So it drives the prom label only. (I checked the code rather than trusting the adjacent comment — that is exactly the mistake this section exists to undo.)

### Cardinality audit — the existing bound is sound

I audited the bound rather than assuming it. Every label on this counter is clamped server-side, after the same-origin and schema gates, and none can be grown by client input:

| label | bound | unknown → |
|---|---|---|
| `error_class` | `KNOWN_ERROR_CLASSES` code-owned `Set` in `normalizeErrorClass` | `other` |
| `slot_id` | `KNOWN_SLOT_IDS` via `normalizeSlotId` | `other` |
| `app_block_id` | approved-app set via `boundAppBlockIdLabel` (TTL-cached, no per-request DB hit) | `other` |
| `result` | zod `z.enum(['ok','error']).default('ok')` | n/a |

`result="ok"` rows always carry the constant `error_class="none"`, so the label is always present and consistent. A client-supplied string **cannot** reach the label unbounded: `errorClass` is schema-capped at 64 chars and then must be an exact member of the code-owned set. Adding a member to that `Set` is the only way a new value can ever become a label — which is why `token_lost_midsession` had to be added there explicitly, and why the allowlist now carries a 🔴 comment saying so.

Worst case with the new class: `(A+1) × 5 slot_id × 8 error_class ≈ 2040` series at `A=50` approved apps. `result` is not an independent multiplier — it is coupled to `error_class` (`none` pairs only with `ok`; the rest only with `error`), so the 8 values already encode the split. The cardinality comment in the metrics module is updated from 7→8 to match.

## Series-identity note

The new `token_lost_midsession` value is a **new label value, not a new label**, so existing series are unaffected — no supersession, no gap. (Had a label been *added*, series identity would change and existing series would be superseded; that already happened in #3119 and is not repeated here.)

🔴 **CORRECTION — an earlier version of this description claimed "the alert is unaffected". That was wrong.** See the [pinned correction comment](https://github.com/civitai/civitai/pull/3516#issuecomment-5146949634) for the full detail.

- ✅ Series identity is genuinely fine: `error_class` was already a registered label, so a new *value* creates a new series without superseding existing ones.
- ❌ **The alert's firing behaviour changes materially.** The live rule selects on `result="error"` and **not** on `error_class`. A `token_lost_midsession` beacon carries `result="error"`, so it lands in the selector. Pre-PR a revocation emitted **zero** error beacons (the rule's own description says so, "verified live 2026-07-31"); post-PR it emits one, crossing the `> 0.5` threshold and firing for the full 24h window.

🔴 **Decision needed before merge:** `tokenTerminal` means the mint permanently failed — app **delist / suspend / scope-revoke / uninstall**. So every deliberate moderation action will now raise a Warning-Discord alert titled "an SSR render failed" for 24h. That may be correct (the app really did die in the viewer's tab), but it should be chosen, not inherited. Options: accept it · exclude via `error_class!="token_lost_midsession"` · split into its own rule. **No alert rule is touched in this PR.**

🔴 **A lockstep `datapacket-talos` PR is required**: the alert's `description` states that mid-session loss "emits nothing at all" (false after this merges) and that "error_class is NOT a Prometheus label yet" (already false since #3119). Otherwise the alert this PR causes to fire tells the on-call that the discriminator explaining it doesn't exist.

## Not changed, deliberately

- 🔴 **No alert-rule changes.**
- 🔴 **`max_over_time` is retained; nothing switched to `rate()`/`increase()`.** This is a per-pod counter whose series are born at 1 and never increment at current volume, so `rate`/`increase` are structurally **0** on it. This was measured and is precisely why the rule uses `max_over_time`.

## Test coverage

**New, in `src/components/AppBlocks/__tests__/pageBlockHostLogic.test.ts`** (node `unit` project) — 8 cases pinning `shouldEmitMidSessionLossBeacon`: emits on the real teardown; does not emit for a launch failure, during a transient blip, while a token remains, on `fatal`/`timeout`/`no_token`/`ready`/`loading`, or twice. Plus a **cross-module contract test** asserting `normalizeErrorClass('error', MID_SESSION_LOSS_ERROR_CLASS)` returns the class itself and not `other` — with an unknown-class control so it proves membership rather than proving `normalizeErrorClass` is a no-op.

**Updated `PageBlockHostTokenTerminal.browser.test.tsx`** (real Chromium, real hooks, real timers). Two of its assertions are **deliberately inverted** — they previously asserted `['ok']`, i.e. *no second beacon*. The "load succeeded" half of that reasoning is still true and still asserted (the first beacon is checked to be untouched: no `status`, no `errorClass`); the "no second beacon" half was the production defect. Now: `['ok','error']`, with the second body matched on `{status:'error', errorClass:'token_lost_midsession'}`. Plus a new at-most-once case.

**Updated `src/tests/api/track/block-render.test.ts`** — `token_lost_midsession` added to the known-classes loop, asserting it survives the allowlist as its own label.

No new hook or mutation was added to `PageBlockHost` (only `useRef`/`useEffect`), so **no `PageBlockHost*.browser.test.tsx` trpc mock needed re-stubbing.**

### Mutation verification — every guard proven reachable

Each guard was broken, the test run confirmed failing **with that guard's specific assertion**, then reverted:

| mutation | killed by | failure |
|---|---|---|
| drop `alreadyEmitted` guard | `is at-most-once per mount` | `expected true to be false` |
| drop `reachedReady` guard | `does NOT emit for a LAUNCH failure` | `expected true to be false` |
| remove `token_lost_midsession` from allowlist | cross-module contract + API route test | `expected 'other' to be 'token_lost_midsession'`; `expected +0 to be 1` |
| disable the beacon emit in the host | 3 browser tests | `expected [ 'ok' ] to deeply equal [ 'ok', 'error' ]` |
| defeat the at-most-once latch in the host | at-most-once + auto-retry browser tests | `expected [ 'ok', 'error', 'error' ] to deeply equal [ 'ok', 'error' ]` |
| **remove the CH-insert guard** | `a SECONDARY beacon ... writes NO ClickHouse row` | `expected vi.fn() to not be called at all, but actually been called 1 times` |
| **set the guard to the WRONG discriminator (`status === 'error'`)** | `a LAUNCH failure is NOT secondary` | `expected vi.fn() to be called 1 times, but got 0 times` |
| **omit `secondary: true` on the client** | teardown browser test | fails on the missing `"secondary": true` field |
| **leak `reachedReadyRef` across block instances** | `does NOT report app B's launch failure as app A's mid-session loss` | `expected [ { appBlockId: 'apb_other', … } ] to deeply equal []` |
| **disable the tRPC `secondary` guard** | `suppresses the Tracker insert for a secondary follow-up beacon` | `expected vi.fn() to not be called at all, but actually been called 1 times` |

🔴 **One test was green for the wrong reason and was fixed.** The first version of the at-most-once browser test re-rendered with *identical props*, so React never re-ran the effect (deps unchanged) — it passed even with the latch defeated. It now drives the **real** re-entry the bounded auto-retry produces (`error → loading → error` via a re-mint attempt that re-fails), and the table above records it killing the mutation.

## Verified vs. needs the preview

**Verified — run and observed:**
- ✅ **CI Typecheck: pass.** Also ran `tsc --noEmit` locally: **zero errors in any file this PR touches** (the residual local errors are environmental — missing `event-engine-common` sibling checkout, unresolvable `kysely`, stale `@civitai/buzz` build output — and are present on clean `main` too).
- ✅ **CI ESLint + Prettier: pass.**
- ✅ **CI Unit tests: pass** (4m26s).
- ✅ **CI `event-engine-common` pin: pass.** All four GitHub checks green.
- ✅ `pageBlockHostLogic`, `block-render` API route, `app-block-runtime.metrics`, `track.blockRender` schema suites — **105 tests pass** locally, re-run after rebasing onto current `main`.
- ✅ `PageBlockHostTokenTerminal.browser.test.tsx` — **8/8 pass** in real Chromium (Playwright).
- ✅ All five mutations in the table above, each killed with that guard's own assertion.
- ✅ Clean test-merge against current `origin/main` (`git merge-tree` rc=0); `main` has not touched any file in this PR since the branch point.
- ✅ **Formatting:** the 4 touched files that `prettier --list-different` flags are flagged **identically on clean `origin/main`**, and running prettier shows it would only reformat pre-existing regions (`generatorSubmitSchema`, the `renderCounterValue` helper) — **none of my added lines**. That churn was deliberately reverted to keep the diff minimal.
- ✅ **Lint debt is untouched, not introduced:** `eslint` reports 22 errors across the two test files I modified (a wholesale `vi.mock('~/utils/trpc')` and `as any` casts). `git diff -U0` confirms **every one is on a pre-existing line I did not author.** CI treats modified-file lint as report-only (`continue-on-error: true`); only *added* files block, and this PR adds none.

**Not verified — flagged honestly:**
- ⏳ The **Tekton pr-preview** (`next build`) was still running when this was written. No file was added under `src/pages/` — the new pure logic is in `src/components/AppBlocks/pageBlockHostLogic.ts`, its test in `src/components/AppBlocks/__tests__/` — so the `PagesPageConfig`/`ApiRouteConfig` failure class should not apply, but the build is the authority.
- ⏳ **The full component (browser) suite** was still running locally; only the directly-affected file is confirmed green. It is report-only in the preview anyway.
- 🔴 **Nothing here has been observed against production**, and it cannot be until a real mid-session revocation happens after deploy — the change is by definition unexercised until then. Suggested check once live:
  ```promql
  sum by(app_block_id, error_class)(max_over_time(civitai_app_block_renders_total{result="error"}[24h]))
  ```
  A `token_lost_midsession` bucket should start appearing where the incident previously produced nothing at all. Until that is seen, this is **deployed, not verified**.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -382,9 +382,27 @@ export function PageBlockHost({
   // 🔴 Deliberately conservative: this resets to `false` rather than re-deriving
   // from `status`, because under host reuse `status` is itself STALE (it is still
   // app A's 'ready'). Re-deriving would re-create the misattribution this fixes.
-  // The cost is that if app B does reach ready and then loses its credential
-  // WITHOUT `status` ever changing value, its beacon is missed — under-reporting,
-  // never mis-reporting. That is the right side to err on for an alerting signal.
+  //
+  // Be precise about the cost, because it is NOT an edge case: under host reuse
+  // `status` is structurally frozen at app A's 'ready' — every `setStatus` here is
+  // gated on the current value, and the only unconditional reset to 'loading' is
+  // `performRetry`. So after a soft nav the latch re-arms only via a manual or
+  // automatic Retry, and until then app B can emit no mid-session beacon at all.
+  //
+  // That costs nothing real TODAY, because on the same path app B never gets a
+  // working session to lose: `shouldStartInit` returns false for any non-'loading'
+  // status, so BLOCK_INIT is never sent and app B cannot reach a genuine ready
+  // state. The missed beacon is therefore unreachable — it describes a state the
+  // pre-existing host-reuse bug already prevents. Net vs. before this reset:
+  // strictly better (we traded a FALSE POSITIVE on app B's launch failure for a
+  // beacon that could not have fired anyway). Under-reporting, never
+  // mis-reporting — the right side to err on for an alerting signal.
+  //
+  // 🔴 The DECLARATION ORDER below is load-bearing: this effect must be declared
+  // BEFORE the status-sync effect that sets `reachedReadyRef`, so that a commit
+  // changing both `blockInstanceId` and `status` resets first and latches second.
+  // No current test detects a swap (no reachable case distinguishes them today) —
+  // so if you reorder these, reason it through rather than trusting the suite.
   //
   // NOTE: the wider host-reuse problem is PRE-EXISTING and NOT fixed here — stale
   // `status` and the leaked `blockRenderEmittedRef` (so app B loses its `ok`
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -620,12 +620,20 @@ export function PageBlockHost({
     // Fire-and-forget beacon — failures are a no-op. `errorClass` is a member of
     // the server-side KNOWN_ERROR_CLASSES allowlist, so it survives as a real
     // `error_class` prom label instead of collapsing into 'other'.
+    //
+    // 🔴 `secondary: true` is REQUIRED here, not decorative. This mount already
+    // sent its `ok` impression, and the `blockRenders` CH row carries no status —
+    // so without this flag the follow-up would write a SECOND byte-identical row
+    // for one mount, un-de-duplicatable, inflating every CH-derived impression
+    // figure for exactly the sessions that were revoked. The flag keeps the prom
+    // counter firing (that is what the alert reads) while skipping the insert.
     sendBlockRender({
       appBlockId,
       blockInstanceId,
       slotId: 'app.page',
       status: 'error',
       errorClass: MID_SESSION_LOSS_ERROR_CLASS,
+      secondary: true,
     });
   }, [status, tokenTerminal, token, appBlockId, blockInstanceId]);
 
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -15,6 +15,7 @@ import {
   buildReviewConsentNotification,
   grantedPageScopes,
   INITIAL_REVIEW_CONSENT_LATCH,
+  MID_SESSION_LOSS_ERROR_CLASS,
   pageFallbackReason,
   resolveCheckpointPickerRequest,
   resolveGetImagesByIdsRequest,
@@ -23,6 +24,7 @@ import {
   resolveResourcePickerRequest,
   resolveReviewConsentNotice,
   resolveUngrantableConsentScopes,
+  shouldEmitMidSessionLossBeacon,
 } from './pageBlockHostLogic';
 import ConfirmDialog from '~/components/Dialog/Common/ConfirmDialog';
 import { projectSafeGenerationResource } from '~/server/schema/blocks/generation-resource-projection';
@@ -338,10 +340,39 @@ export function PageBlockHost({
   // makes the per-mount emit deterministic regardless of ack timing.
   const blockRenderEmittedRef = useRef<boolean>(false);
 
+  // 🔴 A SEPARATE at-most-once latch for the MID-SESSION credential-loss beacon,
+  // and it MUST NOT be `blockRenderEmittedRef`.
+  //
+  // `blockRenderEmittedRef` encodes an ANALYTICS invariant — exactly ONE
+  // impression per host mount — that the `blockRenders` denominator and the
+  // BlockRenderFailureRate alert both depend on. Reusing it here would be wrong
+  // in both directions: a `ready` host has already consumed it (so the beacon
+  // would never fire — the precise reason a real prod revocation recorded zero
+  // error beacons on 2026-07-31), and clearing/relaxing it to make room would
+  // retroactively change what the already-sent `ok` impression means.
+  //
+  // The teardown is a genuinely DIFFERENT event from the page load, so it gets a
+  // different latch. Net effect on the wire for a revoked session: `ok` (the load
+  // really did succeed) followed by ONE
+  // `error{error_class="token_lost_midsession"}` (it was later revoked). Bounded
+  // per mount by this ref; the impression accounting above is untouched.
+  const midSessionLossEmittedRef = useRef<boolean>(false);
+  // Latches on the first committed `ready`. This is what tells a `ready → error`
+  // teardown apart from a `loading → error` launch failure — `status === 'error'`
+  // alone cannot (both transitions land on it). A ref, not state: it must not
+  // cause a render, and `handleRetry` deliberately does not clear it (a mount
+  // that ever launched has launched).
+  const reachedReadyRef = useRef<boolean>(false);
+
   // Keep statusRef tracking the live status so handleRetry can branch on the
   // prior terminal state without reading it inside the setStatus updater.
   useEffect(() => {
     statusRef.current = status;
+    // Latch "this mount reached ready" off the COMMITTED status, for the same
+    // reason the impression beacon does (see its comment): keying off a
+    // setStatus updater's side effect silently drops the observation whenever
+    // another state update is already queued on this component.
+    if (status === 'ready') reachedReadyRef.current = true;
   }, [status]);
 
   // BOUNDED AUTO-RETRY — the single decision both consumers read (the scheduling
@@ -546,14 +577,57 @@ export function PageBlockHost({
   // (an `error` status is an AUTH terminal, so its one automatic attempt spends a
   // re-mint) and surfaces the prominent manual Retry once that settles too.
   //
-  // 🔴 NO SECOND BEACON. A host that reached `ready` already fired its ONE `ok`
-  // impression, and `blockRenderEmittedRef` is per-mount — so the failure beacon
-  // below is inert here by construction. The episode stays one beacon, reporting
-  // that the page load itself succeeded (which it did).
+  // 🔴 THE IMPRESSION BEACON IS STILL EXACTLY ONE. A host that reached `ready`
+  // already fired its ONE `ok` impression, and `blockRenderEmittedRef` is
+  // per-mount — so the launch-failure beacon below remains inert here by
+  // construction, and the `ok` already sent still means what it always meant
+  // (the page load succeeded — which it did). The teardown is reported by the
+  // SEPARATE mid-session beacon effect immediately after this one, on its own
+  // latch, so observability is gained without touching impression accounting.
   useEffect(() => {
     if (!tokenTerminal || token) return;
     setStatus((current) => (current === 'ready' ? 'error' : current));
   }, [tokenTerminal, token]);
+
+  // MID-SESSION credential-loss render beacon — the observability half of the
+  // teardown above.
+  //
+  // 🔴 WHY THIS EXISTS AT ALL (measured on production 2026-07-31): a real
+  // revocation teardown was driven end-to-end against a live app and the
+  // platform recorded ZERO error beacons for it. The only trace of the incident
+  // was the earlier successful `ok` impression, so every render metric said the
+  // app was healthy while it was dead in the viewer's tab. A mid-session
+  // teardown is exactly the failure mode a third-party app platform must be able
+  // to see — it is what a delist/suspend/revoke looks like from the runtime.
+  //
+  // Keys off the COMMITTED `status` (not a setStatus updater's side effect) for
+  // the same batching reason documented on the impression beacon below. The full
+  // decision — including WHY it is gated on `reachedReady` and on the settled
+  // `tokenTerminal` rather than a bare `tokenError` — lives in the pure,
+  // unit-tested `shouldEmitMidSessionLossBeacon`.
+  useEffect(() => {
+    if (
+      !shouldEmitMidSessionLossBeacon({
+        status,
+        reachedReady: reachedReadyRef.current,
+        tokenTerminal,
+        hasToken: !!token,
+        alreadyEmitted: midSessionLossEmittedRef.current,
+      })
+    )
+      return;
+    midSessionLossEmittedRef.current = true;
+    // Fire-and-forget beacon — failures are a no-op. `errorClass` is a member of
+    // the server-side KNOWN_ERROR_CLASSES allowlist, so it survives as a real
+    // `error_class` prom label instead of collapsing into 'other'.
+    sendBlockRender({
+      appBlockId,
+      blockInstanceId,
+      slotId: 'app.page',
+      status: 'error',
+      errorClass: MID_SESSION_LOSS_ERROR_CLASS,
+    });
+  }, [status, tokenTerminal, token, appBlockId, blockInstanceId]);
 
   // Token never resolves → surface a no_token state instead of an endless
   // skeleton.
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -383,11 +383,13 @@ export function PageBlockHost({
   // from `status`, because under host reuse `status` is itself STALE (it is still
   // app A's 'ready'). Re-deriving would re-create the misattribution this fixes.
   //
-  // Be precise about the cost, because it is NOT an edge case: under host reuse
-  // `status` is structurally frozen at app A's 'ready' — every `setStatus` here is
-  // gated on the current value, and the only unconditional reset to 'loading' is
-  // `performRetry`. So after a soft nav the latch re-arms only via a manual or
-  // automatic Retry, and until then app B can emit no mid-session beacon at all.
+  // Be precise about the cost, because it is NOT an edge case. `status` inherited
+  // from app A can still move FORWARD out of 'ready' (→ 'error' on a terminal
+  // token, → 'fatal' on BLOCK_ERROR) — it is not frozen. What it cannot do is go
+  // BACK: every `setStatus` here is gated on the current value, and the only
+  // unconditional reset to 'loading' is `performRetry`. Since re-reaching 'ready'
+  // requires passing through 'loading', app B can never earn a genuine `ready`
+  // after a soft nav, so the latch re-arms only via a manual or automatic Retry.
   //
   // That costs nothing real TODAY, because on the same path app B never gets a
   // working session to lose: `shouldStartInit` returns false for any non-'loading'
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -364,6 +364,38 @@ export function PageBlockHost({
   // that ever launched has launched).
   const reachedReadyRef = useRef<boolean>(false);
 
+  // 🔴 BOTH REFS ABOVE ARE PER-MOUNT, BUT THIS HOST IS NOT ALWAYS REMOUNTED.
+  //
+  // `/apps/run/[slug]/[[...path]]` renders <PageBlockHost> with NO `key`, and
+  // `_app.tsx` renders <Component> with no key either — so a SOFT navigation
+  // between two apps (appA → appB, e.g. via the "Recently run" menu) reuses this
+  // component instance: same mount, different `blockInstanceId`.
+  //
+  // Left alone, app A's latches would leak onto app B and produce a WRONG,
+  // MISATTRIBUTED signal: B's launch failure would inherit `reachedReady === true`
+  // and be reported as `token_lost_midsession` for an app that never launched —
+  // exactly the launch-vs-teardown confusion this whole feature exists to avoid.
+  // (Conversely a spent emit-latch would silently swallow B's genuine loss.)
+  //
+  // So the latches are scoped to the block INSTANCE, not the React mount.
+  //
+  // 🔴 Deliberately conservative: this resets to `false` rather than re-deriving
+  // from `status`, because under host reuse `status` is itself STALE (it is still
+  // app A's 'ready'). Re-deriving would re-create the misattribution this fixes.
+  // The cost is that if app B does reach ready and then loses its credential
+  // WITHOUT `status` ever changing value, its beacon is missed — under-reporting,
+  // never mis-reporting. That is the right side to err on for an alerting signal.
+  //
+  // NOTE: the wider host-reuse problem is PRE-EXISTING and NOT fixed here — stale
+  // `status` and the leaked `blockRenderEmittedRef` (so app B loses its `ok`
+  // impression) both predate this change. The real fix is `key={blockInstanceId}`
+  // on the run page, which would also change impression COUNTS, so it belongs in
+  // its own PR rather than riding along on an observability change.
+  useEffect(() => {
+    reachedReadyRef.current = false;
+    midSessionLossEmittedRef.current = false;
+  }, [blockInstanceId]);
+
   // Keep statusRef tracking the live status so handleRetry can branch on the
   // prior terminal state without reading it inside the setStatus updater.
   useEffect(() => {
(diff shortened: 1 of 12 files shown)


ZacxDev:
## 🔴 Correction to this PR's description — "the alert is unaffected" is misleading

An adversarial audit caught this, and I'm correcting it in a comment rather than silently editing the body, since a reviewer may already have read the wrong version.

**What the description said:** *"The alert is unaffected. `sum by(app_block_id)(...)` aggregates away `error_class`, so a new value in that dimension folds into the same per-app sum."*

**Split that in two:**

| claim | verdict |
|---|---|
| adding a new *value* to an existing label does not supersede existing series | ✅ **true** — `error_class` was already a registered label; new value = new series, existing series keep identity |
| the alert is unaffected | ❌ **false, and materially so** |

**Why.** The live rule (`app-block-alerts-configmap.yaml`, uid `app-block-render-failure-rate`) selects:

```promql
sum by(app_block_id)(max_over_time(civitai_app_block_renders_total{
  cluster="dp-1",namespace="civitai-dp-prod",result="error",app_block_id!="other"}[24h])) > 0.5
```

It filters on **`result` only — not `error_class`.** A `token_lost_midsession` beacon carries `result="error"`, so it lands squarely in the selector.

Pre-PR a revocation emitted **zero** error beacons — the rule's own description states this, "verified live 2026-07-31". Post-PR it emits one, the sum crosses `0.5`, and the alert **fires within 5m and stays firing for the full 24h `max_over_time` window.**

So the difference is not cosmetic: it is 0 → ≥1.

### 🔴 The operational consequence worth a decision before merge

`tokenTerminal` means the token mint **permanently** failed — i.e. **app delist, app suspend, scope revoke, uninstall**. Those are deliberate moderation actions.

**So after this merges, every deliberate moderation action on an app raises a Warning-Discord alert titled "App Block …: an SSR render failed" and holds it for 24h.** An admin action gets reported as a platform failure.

That may be *exactly right* — from the viewer's perspective the app did stop working in their tab, and making that visible is the entire point of this PR. But it is a real change in alert behaviour and it should be a deliberate choice, not a side effect. Options:

1. **Accept it** — a revocation genuinely is a render failure the platform should surface.
2. **Exclude it:** add `error_class!="token_lost_midsession"` to the alert selector.
3. **Split it** into its own rule with its own severity/routing (e.g. info, not warning).

I have deliberately **not** touched the alert rules in this PR. This needs a decision plus a lockstep `datapacket-talos` change.

### 🔴 The alert's runbook text will be wrong on deploy — needs a lockstep infra PR

The same rule's `description` contains two clauses that are or become false:

> *"…a MID-SESSION credential loss on an already-ready host **emits nothing at all** (verified live 2026-07-31), so a flat line is NOT proof of health… NOTE **error_class is NOT a Prometheus label yet** (accepted + bounded server-side, then dropped) — the failure reason is not queryable here."*

- Clause 1 becomes **false the moment this merges** — that is the whole change.
- Clause 2 is **already false** today: `error_class` has been a live label since #3119. (This is the same stale-comment rot this PR fixes in `sendBlockRender.ts`; it propagated into the alert annotation, and the same ConfigMap contradicts itself elsewhere by listing `error_class` among the labels.)

Net effect if not fixed in lockstep: the alert that fires **because of this PR** tells the on-call that the discriminator explaining it doesn't exist.

### Not affected — verified

ClickHouse impression data is **safe**. The `secondary` suppression is complete: one writer (`Tracker.blockRender`), two call sites, both gated; the flag is stripped from the payload; the schema defaults it to `false` so legacy beacons are untouched. `blockRenders` stays at exactly one row per mount, so the digest CronJob, `civitai_app_blocks_impressions_24h`, `renders_by_app_24h` and `anon_rate_24h` all produce byte-identical numbers before and after. No backfill or de-duplication needed.

### 🟡 Two dashboard panels are skewed (not blocking today)

Because the prom counter **deliberately** double-counts a revoked session (`ok` + `error`), on `/d/app-blocks-runtime`:

- **"Render-failure ratio by app_block_id"** — a revoked mount adds +1 numerator and **+2** denominator, so the panel *understates* the true per-mount failure fraction, and a total-revocation outage reads exactly **50%, never 100%**.
- **"Total render rate"** — now counts beacons, not mounts.

Benign at current volume (`rate()` is structurally ~0 on this per-pod counter, as that ConfigMap documents at length), **but** the ConfigMap carries a "🔴 RESTORE A RATIO AT GA" note — so the inflated denominator becomes an alerting-correctness bug at GA unless the restored ratio excludes secondary beacons. Flagging it now so it is not rediscovered then.

ZacxDev:
## Audit round 2 — one real bug found in this PR's own code, fixed

Two independent adversarial audits ran. They converged on the alert finding (already corrected above) and turned up one genuine defect I had introduced, plus a coverage gap. Both are fixed in `76bf914d06` (now rebased onto current `main`).

### 🔴 Misattribution across an app → app soft navigation — a real bug, now fixed

`/apps/run/[slug]/[[...path]]` renders `<PageBlockHost>` with **no `key`**, and `_app.tsx` renders `<Component>` with no key either. So navigating appA → appB (the "Recently run" menu) **reuses the component instance**: same React mount, new `blockInstanceId`.

`reachedReadyRef` / `midSessionLossEmittedRef` are per-**mount** refs, so they leaked across that boundary:

- App B's **launch** failure inherited app A's `reachedReady === true` and was emitted as `token_lost_midsession` — **for an app that never launched**, with app B's id attached. Exactly the launch-vs-teardown confusion this feature exists to eliminate.
- Mirror case: a spent emit-latch silently swallowed app B's genuine loss.

**This was not theoretical.** Mutation-verified: removing the reset makes the new test fail with

```
AssertionError: expected [ { appBlockId: 'apb_other', …(5) } ] to deeply equal []
```

i.e. the misattributed beacon really is emitted without the fix.

Both latches now reset on a `blockInstanceId` change. The reset effect is declared **before** the status-sync effect that sets `reachedReadyRef`, so ordering is correct in all three combinations (instance change alone / status change alone / both in one commit).

It deliberately resets to `false` rather than re-deriving from `status`, because under host reuse `status` is *itself* stale (still app A's `ready`) — re-deriving would recreate the misattribution. The cost is a missed beacon if app B reaches ready and dies without `status` ever changing value: **under-reporting, never mis-reporting**, which is the right side to err on for an alerting signal.

🟡 **Pre-existing and deliberately NOT fixed here:** the wider host-reuse breakage — stale `status`, and app B losing its `ok` impression to the leaked `blockRenderEmittedRef`. The real fix is `key={blockInstanceId}` on the run page, which would also change impression **counts**. That is an analytics change and belongs in its own PR, not an observability one. Flagging it so it isn't lost.

### 🟡 The tRPC `secondary` branch was untested — now pinned

`track.router.ts`'s `if (secondary) return;` is the security-relevant branch (its own comment notes a bearer/API-key caller could otherwise reintroduce the ClickHouse double-count), and nothing covered it. Added two cases — follow-up suppressed; launch failure still writes its row with `status`/`errorClass`/`secondary` all stripped. Mutation-verified: disabling the guard fails with `expected vi.fn() to not be called at all, but actually been called 1 times`.

### Corrections to this PR's description

- **Scope of the win.** The gate is `status === 'error'`, so a host that reached `ready` and then **crashed** (`BLOCK_ERROR{fatal:true}` → `fatal`) still emits nothing beyond its `ok`. That's deliberate — a crash is a different failure and mis-tagging it would make the class lie — but "a mid-session teardown is now observable" is true specifically for **credential loss**, not every mid-session failure. The pre-existing test *"a mount that already reported `ok` never additionally reports `error` after a crash + failed recovery"* still passes, confirming the boundary.
- **tRPC precision.** An earlier draft said the tRPC path skips the CH insert "while the prom counter still fires". Wrong: that path **never** touched the prom counter, before or after — for a `secondary` beacon it is a total no-op. No impact today (every host uses `sendBlockRender` → the REST beacon), but the description was imprecise.

### 🟡 Noted, not addressed

`/api/track/block-render` is a `PublicEndpoint` with **no rate limiting**, and its same-origin gate is header-based (`Origin`/`Referer` vs `Host`) — spoofable by a non-browser client. `secondary: true` returns *before* the session resolve and the CH insert, so an attacker can now inflate `renders_total{result="error"}` at strictly lower server cost and **with no ClickHouse trace** (previously such inflation also wrote rows, leaving evidence). Combined with the alert's `> 0.5` threshold, one scripted request can page the team for any approved `app_block_id`.

Cardinality stays bounded — this is counter-**value** abuse, not series abuse — and it is largely pre-existing; `secondary` sharpens it. Rate-limiting that endpoint is the fix, and it is out of scope here.

*(The reverse concern — a client using `secondary: true` to suppress its own impression rows — is **not** a meaningful vector: a client could always simply not send the beacon, so under-counting was already fully available.)*

### Verification after this round

All 4 GitHub checks green pre-rebase; re-running now. Locally: **117 unit tests** and **9/9 browser tests** pass, `tsc --noEmit` clean in every touched file, and **7 mutations** total, each killed by that guard's own assertion. Branch rebased onto current `main` (the earlier "27 files / 827 insertions" two-dot diff was an artifact of being 4 commits behind — the real diff is 11 files).

ZacxDev:
## Audit round 3 — clean. Loop stopped.

Third adversarial round on the fix commits. **No bug, no regression, no weakened coverage.** It produced one 🟡 wording correction and one 🟢 note, both now fixed in `d0de99f6b5` (provably comment-only — the diff contains no non-comment line).

### The one thing that mattered: my non-vacuity proof was independently reproduced

Round 2 left a test whose second assertion was **vacuous** — it filtered for beacons attributed to app B, and that list is empty, so it asserted nothing and would have held with the fix deleted. Replaced with `expect(beaconStatuses()).toEqual(['ok'])`.

Proving *that* replacement isn't vacuous took an isolation protocol, because the first attempt proved nothing: the mutation was killed by the **preceding** assertion, which throws first. A test killed by a *different* guard tells you nothing about the guard you're checking.

Four runs, independently reproduced by the auditor:

| run | source | preceding assertion | result |
|---|---|---|---|
| baseline | clean | enabled | 9/9 pass |
| mutation only | reset deleted | enabled | fails at the **earlier** line — proves nothing about the new one |
| **isolation** | reset deleted | **disabled** | fails at the new line with **its own** message: `expected [ 'ok', 'error' ] to deeply equal [ 'ok' ]` |
| **control** | clean | disabled | **passes** |

The isolation + control pair is the proof: the assertion fails under the mutation and passes without it, at its own line, with its own error.

### 🟡 Fixed: a false clause I wrote inside a 🔴-marked comment

The ref-reset comment claimed `status` is *"structurally frozen at app A's 'ready'"* under host reuse. **False** — two gated transitions leave `'ready'`: `→ 'error'` on a terminal token (`PageBlockHost.tsx:639`) and `→ 'fatal'` on `BLOCK_ERROR{fatal:true}` (`:771`).

There is direct execution evidence on exactly the path the comment describes: `shouldEmitMidSessionLossBeacon` returns false unless `status === 'error'`, yet the soft-nav mutation run *did* emit the misattributed beacon — so status demonstrably left `'ready'`. The paragraph also contradicted itself four lines later by invoking the automatic Retry, which only exists because status reached the `'error'` terminal `decideAutoRetry` reads.

The **conclusion** was never wrong, only the premise's wording. The load-bearing fact is the narrower one already in the same sentence: status cannot go **back** — every `setStatus` is value-gated and the only unconditional reset to `'loading'` is `performRetry` — so re-reaching `'ready'` is impossible and app B can never earn a genuine ready after a soft nav.

Worth fixing rather than shrugging at: the paragraph opens *"Be precise about the cost"*, sits in a 🔴 block, and a reader trusting "frozen" would wrongly conclude the visible host state and the bounded auto-retry are inert after a soft nav.

**This is the third incorrect comment this PR has had to fix** — the others being `sendBlockRender`'s "NOT a prom label" and the alert annotation's "emits nothing at all". Comments in this area get trusted on sight, which is exactly how the original brief for this work was built on a stale one.

### 🟢 Fixed: the new test's scope, stated honestly

The test renders `PageBlockHost` **directly and unkeyed**, so it pins the *host's* behaviour under instance reuse. Landing the real fix (`key={blockInstanceId}` on the run page) would **not** fail it — it removes the scenario in production while the test keeps reproducing it locally. The comment now says so, and records that `['ok']` is over-specified **by design**: the only change that would trip it is scoping `blockRenderEmittedRef` per instance, which this PR explicitly reasons against (the per-mount impression invariant).

### Verification summary

- **Tekton `pr-preview` (the authoritative `next build`): `Completed` / success** on `9b4a3cf0ce`, the full substantive code change. Every commit after it is provably comment/test-only. `pr-check` runs all Succeeded.
- All four GitHub checks green on the code commits.
- **117 unit + 9/9 browser** tests pass; `tsc --noEmit` clean in every touched file.
- **9 mutations** across the PR, each confirmed to fail with **that guard's own** assertion — including the isolation protocol above.
- Both ClickHouse write paths covered: the REST beacon route **and** the still-live `track.blockRender` tRPC mutation.

Three rounds, three real defects caught pre-merge (CH double-count, cross-app misattribution, vacuous assertion) plus three incorrect comments. The loop is stopped because this round found only prose.

**Still open for a human decision — unchanged from the earlier comment:** whether a delist/suspend/revoke *should* fire this alert, and the lockstep `datapacket-talos` PR for the two false clauses in the alert description. I have not touched any alert rule.

PR #3517, opened 2026-07-31
fix(blocks): reject an edit-only checkpoint version submitted with no sourceImage

ZacxDev:
## The bug

The App Blocks image bridge chooses its graph workflow purely from `sourceImage` presence:

```ts
if (!body.sourceImage) return 'txt2img';
```

Several image ecosystems ship **different checkpoint versions per workflow** (`createCheckpointGraph({ workflowVersions })` in `qwen-graph.ts` / `boogu-graph.ts` / `mage-flow-graph.ts`). Qwen model `2268063` hosts both `2558804` ("Image Edit 2511", offered only on `img2img:edit`) and `2552908` (offered only on `txt2img`) — disjoint id sets on the same model.

So naming the edit version with no `sourceImage` resolved to `txt2img`. And the generation graph did **not** reject it — it returned success with a different checkpoint.

### The mechanism (corrected)

> ⚠️ An earlier revision of this PR blamed `buildModelTransform`'s same-index sibling mapping. **That was wrong**, and the correction is below. The guard's behaviour was correct throughout; only the explanation and the characterization test were wrong.

The substitution is the **`modelLocked` clamp** in `createCheckpointGraph`'s `checkpointInputSchema` (`src/shared/data-graph/generation/common.ts`):

```ts
if (modelLocked && modelVersionId && val.id !== modelVersionId) {
  if (!validVersionIds?.has(val.id)) {
    return { id: modelVersionId, model: { type: 'Checkpoint' } };   // the workflow's defaultModelId
  }
}
```

On a `modelLocked` ecosystem, **any** id not in the *current workflow's* visible list is replaced with that workflow's `defaultModelId` — regardless of its position in the other workflow's list, and regardless of whether the ecosystem has ever heard of it.

`buildModelTransform` is not involved on this path at all. It is gated on `depsChanged && !isDirectUpdate` (`src/libs/data-graph/data-graph.ts`), and a one-shot `safeParse` that passes `model` explicitly — exactly what this bridge does — *is* a direct update. Instrumented: **0 invocations** across every input below; stubbing it to `undefined` changes no output and fails no test.

### The discriminator (executed, not inferred)

The two hypotheses only disagree where the input's same-index sibling isn't already the workflow default. Real `generationGraph.safeParse`, Qwen + MageFlow, free-tier ctx:

| workflow | ecosystem | input | index-map predicts | default-clamp predicts | **ACTUAL** |
|---|---|---|---|---|---|
| `txt2img` | Qwen | `2133258` (edit v2509, **index 0**) | `2110043` | `2552908` | **`2552908`** |
| `txt2img` | Qwen | `2558804` (edit v2511, index 1) | `2552908` | `2552908` | `2552908` |
| `txt2img` | Qwen | `987654321` (**unrecognized**) | — | `2552908` | **`2552908`** |
| `txt2img` | MageFlow | `3172044` (edit_turbo, **index 1**) | `3172039` | `3172038` | **`3172038`** |
| `img2img:edit` | Qwen | `2552908` (txt-only) + `images` | — | `2558804` | **`2558804`** |

All five return `success: true`. The caller is billed, gets images, and never learns the checkpoint it asked for was swapped. Nothing in the response, the `BlockWorkflowSnapshot`, or the `block_workflows` read-model reveals it. A successful-looking wrong answer is the worst failure mode available.

The previous characterization test covered **only row 2** — the one case where both hypotheses agree — so it would have passed unchanged if the mechanism it documented had been deleted. It now pins rows 1, 2, 3, 4 and 5 plus a non-`modelLocked` contrast case.

### Not every affected ecosystem was silently swapping

`modelLocked` is what splits them:

| | Before this PR | This PR is… |
|---|---|---|
| **`modelLocked`** — Qwen, MageFlow (**23 of 35** image ecosystems) | silent substitution, `success: true` | a **correctness fix** |
| **not `modelLocked`** — Boogu, SDXL, Flux1, … | the id survives, the mode discriminator routes to the edit subgraph, validation trips: `success: false`, `images: "An image is required"` | an **earlier, clearer error** |

An earlier revision called Boogu "the killing case". It is not — Boogu is not `modelLocked` and `{workflow:'txt2img', ecosystem:'Boogu', model:{id:3049824}}` already returned `success: false`. Verified by execution. The guard still rejects all three ecosystems uniformly; only the description of what it prevents differs.

## Blast-radius measurement — and its limit

🔴 **The submitted body is not retained anywhere, so I could not measure whether any live app relies on today's behaviour.** Reporting that plainly rather than guessing.

What I checked, in prod:

| Store | Verdict |
|---|---|
| `block_workflows` (G6 read-model) | 7 columns — `workflow_id`, `app_block_id`, `block_instance_id`, `user_id`, `status`, `submitted_at`, `updated_at`. **No body, no modelVersionId, no sourceImage.** Confirmed against the live schema, not just the migration. |
| `block_scope_invocations.detail` | Only `{action:'workflow.submit', amount, outcome}` — no version. |
| `block_spend_attribution` | Money/payout record only. |

What I *could* bound is the population:

- `block_workflows`: **21 rows**, 4 app blocks, 3 users, 2026-07-16 → 2026-07-28.
- `block_scope_invocations` where `endpoint LIKE 'workflow:submit:%'` (the full history, back before G6 landed): **100 submits ever**, 10 app blocks, 4 users, 2026-05-30 → 2026-07-28.

So the entire population of block-driven generations is dogfood-scale.

**Residual risk:** it remains *possible* that one of those 100 submits named an edit-only version with no `sourceImage` and something downstream depends on the substituted result. I judge that close to unrealistic — such a caller was never getting the checkpoint it named, so it cannot have been depending on the edit checkpoint — but it is an argument, not a measurement.

## The fix

Two guards, both derived from **real config**, no hardcoded id table:

1. **Ecosystem-level** (`resolveBlockImageWorkflowType`) — an ecosystem that supports `img2img:edit` but not `txt2img` cannot honour a bare body. Derived from `isWorkflowAvailable`. Currently **unreachable**: every member of `EDIT_IMG_IDS` is also in `TXT2IMG_IDS`. That invariant is asserted by a test, so if an edit-only ecosystem ever ships we learn the guard went live.
2. **Version-level** (`assertCheckpointVersionSupportsWorkflow`) — the one that bites today. Rejects a workflow-scoped version belonging to a different workflow, with an error naming the remedy and the concrete sibling version:

> `modelVersion 2558804 is not available for 'txt2img' on the Qwen ecosystem — it is offered for img2img:edit only. Either send a `sourceImage` to run it as an image edit, or use modelVersionId 2552908 for txt2img.`

(The suggested id is the same-position sibling via the graph's `buildVersionMappings` — that is the *guard's own suggestion policy* for a useful error message, not a claim about how the graph substitutes.)

### New module: `src/shared/data-graph/generation/workflow-capability.ts`

The per-workflow version lists live only inside the per-ecosystem graph files. Rather than a parallel table that rots, this **instantiates the real `generationGraph`** for an (ecosystem, workflow) pair and reads the node meta the form itself renders from. ~1.4 ms, memoized per pair.

The probe runs with a neutral, gate-rule-free context on purpose: it answers "what does the **config** allow?", not "what may **this user** generate?" — per-user gating still runs downstream. Empty `gateRules` keeps gated versions visible, so the probe can only ever fail **open**.

### Scope — a subset, honestly

This closes **the ecosystem's own workflow-scoped versions, in the txt2img direction only**. The broader class is real and stays open. Tracked in **#3520**:

- **An unrecognized id is also substituted.** A community checkpoint or brand-new upload on a `modelLocked` ecosystem gets clamped to the workflow default exactly like a known out-of-list id — row 3 of the table above. It is **not** "left alone". This is the wide part of the class (23 of 35 image ecosystems), and rejecting it here would refuse ids the graph is willing to run, so the guard deliberately fails open on it. Only non-`modelLocked` ecosystems pass an unknown id through untouched.
- **The reverse direction is unchanged.** A txt2img-only version sent *with* a `sourceImage` is substituted the same way — row 5. `assertCheckpointVersionSupportsWorkflow` is already direction-agnostic, so flipping it on is a one-line change; it is **held deliberately** as a separate behaviour change with its own blast radius, not overlooked. There is a test proving the helper reports that direction correctly, and a characterization test pinning that the graph still substitutes there.
- The guard fires on the **resolved checkpoint** version, not `body.modelVersionId` — for a LoRA-bound install those differ. Covered by a test.

## Tests

`src/server/services/blocks/__tests__/workflow.service.test.ts` (+ a new file for the mocked edit-only-ecosystem case):

- edit-only version + no `sourceImage` → rejected — Qwen, MageFlow (silent-swap cases), Boogu (already-erroring case)
- edit-only version + `sourceImage` → still `img2img:edit`, requested version preserved
- txt2img-capable version + no `sourceImage` → unchanged
- dual-capable ecosystem (Flux2, OpenAI — identical lists) → both modes still work
- unlisted community checkpoint on a scoped ecosystem → **not** rejected
- ecosystem with no version scoping (SDXL) → unchanged
- LoRA install → guard fires on the resolved checkpoint, not `body.modelVersionId`
- error message names the fix (version, workflow, `sourceImage`, sibling version id)
- reverse direction (`sourceImage` on a neither-variant ecosystem) → unchanged
- config invariant: no real image ecosystem is currently edit-only

**Characterization tests against the real graph** — every row of the table above, plus:

- non-`modelLocked` contrast (Boogu): unknown id survives untouched; edit-only id → `success: false`
- unrecognized-id substitution on Qwen and MageFlow, tagged `#3520`

### Mutation testing — every guard and every claim proven to bite

All re-measured against the current test set (`src/server/services/blocks src/shared/data-graph`, baseline 98 files / 1902 tests green):

| Mutation | Result |
|---|---|
| `if (workflowType === 'txt2img')` → `if (false)` (version guard never called) | **7 tests fail** |
| `resolveVersionWorkflowScope`: remove **both** `unscoped` fail-opens | **19 tests fail** (incl. pre-existing `buildTextToImageInput` tests). Removing only the `!offeredFor.length` one fails **2** — the `!targetIds` branch carries most of the blast radius. |
| ecosystem-level guard: `!isWorkflowAvailable('txt2img', …)` → `false` | **1 test fails** (the mocked edit-only-ecosystem case) |
| **`modelLocked` clamp in `common.ts` disabled** (the named mechanism) | **6 tests fail** — all three discriminator rows, both unknown-id rows, and the reverse-direction row. The three substitution rows flip to `success: false`, which is itself the proof: without the clamp there is no swap. |
| **`buildModelTransform` stubbed to `undefined`** (the mechanism the earlier write-up named) | **0 tests fail**, 122/122 pass, and no observable output changes anywhere — including on Boogu, the one workflow-scoped ecosystem where the clamp can't mask it. Instrumented at 0 invocations. |

The last two rows are the point: the previous test would have survived deleting the mechanism it claimed to document. It no longer would.

All reverted; suite green after each.

## Verification

```
git submodule update --init --recursive
NODE_OPTIONS=--max_old_space_size=8192 ./node_modules/.bin/tsc --noEmit   → TSC_EXIT=0   (TypeScript 5.9.2, repo-local binary)
npx vitest run src/server/services/blocks src/shared/data-graph --reporter=dot
                                                                          → 98 files, 1902 tests, all passing
```

Collection confirmed per-file (not just the totals): `workflow.service.test.ts` collected **122** tests, `workflow.service.edit-only-ecosystem.test.ts` **5**, `mage-flow-graph.test.ts` **5**, `seedream-graph.test.ts` **4**.

## Not verified

- No end-to-end run against a live orchestrator — the guard is pure, pre-submit, and the reject path never reaches the orchestrator, but the *accepted* paths were exercised through the real `generationGraph`, not a real generation.
- Whether any historical submit actually hit this — see the measurement section; the data does not exist.
- The `modelLocked` counts (23 of 35 image ecosystems) are read from `basemodel.constants.ts` via `getEcosystemSetting`, counting ecosystems for which `isWorkflowAvailable('txt2img' | 'img2img:edit')` holds. A different definition of "image ecosystem" gives a different denominator.


ZacxDev:
**The code here is good. The write-up is wrong, and that's on me — I supplied the mechanism.**

I've now verified this by *executing* `generationGraph.safeParse` rather than reading the code, and the docblock's two central claims don't hold.

### 1. Wrong mechanism

The docblock says `buildModelTransform` maps an out-of-workflow version to its **same-index sibling**. It doesn't. The substitution is the **`modelLocked` clamp** in `createCheckpointGraph`'s `checkpointInputSchema` (`src/shared/data-graph/generation/common.ts:979`), which forces any id not in the current workflow's visible list to that workflow's `defaultModelId`.

The discriminator (Qwen, real graph):

| input | index-map predicts | default-clamp predicts | **actual** |
|---|---|---|---|
| `2133258` (edit v2509, **index 0**) | `2110043` | `2552908` | **`2552908`** |
| `2558804` (edit v2511, index 1) | `2552908` | `2552908` | `2552908` |

Only the second case was ever tested — and it's the one where both hypotheses agree. Consequence for the test: **`silently substitutes` would pass unchanged if the mechanism it documents were deleted.** It pins "some substitution occurs", not the mechanism, and attributes it to the wrong code.

### 2. The SCOPE rationale is false

The docblock says a community checkpoint the ecosystem doesn't list "is left alone (the graph's own transform skips unknown ids too, so nothing is being substituted for it and there is nothing to warn about)."

```
in=987654321 (unrecognized id), workflow=txt2img, ecosystem=Qwen  →  out={"id":2552908}
```

**Any** out-of-list id is substituted on a `modelLocked` ecosystem — reportedly 23 of ~35. So the class this PR leaves open is much larger than the docblock implies, and it's left open for a stated reason that isn't true.

### 3. The reverse direction is real

```
workflow=img2img:edit, model=2552908 (txt2img v2512), images=[…]  →  out={"id":2558804}
```

Same silent substitution, unguarded — and arguably the likelier direction, since `2552908` is the ecosystem default that model pages surface.

---

**None of this is a code-correctness objection.** An audit swept every version in every txt2img-capable ecosystem and found **zero false positives**, and confirmed the guard runs before any Buzz reservation, entitlement write, or orchestrator call. It's well-scoped and fails open where uncertain.

The problem is purely that the write-up would leave the next reader believing this bug class is closed when most of it isn't. I've filed the full class as **#3520** with the executed evidence. Follow-up incoming to fix the docblock, add the discriminator case to the characterization test, and correct the PR body.

PR #3525, opened 2026-08-01
fix(monetization): stop tier price caps from capping Early Access

manuelurenah:
Fixes CU [868kk3avk](https://app.clickup.com/t/8459928/868kk3avk) — **live revenue leak**, reported by MNeMiC / alexds9 / SubtleShader.

## Problem

The per-tier paid-access price cap (`078df52f15`, `3a6bf62700`, first prod tag `v5.0.2212` on 2026-07-31 06:17 UTC) is applied to **every** `PaidAccess` gate — nothing branched on timed vs permanent. `getCapTier` returns `null` for a creator with no good-standing subscription, which the cap helpers read as free → **500 Buzz**. So a 10k/15k Early Access window charged 500.

It bled into three places:

| Surface | Code |
|---|---|
| What the buyer is charged | `model-version.service.ts:1997` (`effectivePaidAccessPrice`) |
| What the buy button shows | `getViewerMonetization` → `cappedTerms` (model page, version page, creator shop, generation gating) |
| What a creator may save | `assertPaidAccessCaps` — a free-tier creator could no longer even *set* a 10k EA price |

Caps were only ever meant to bind **permanent** Paid Access.

## Measured impact (prod, at time of investigation)

- **76** currently-active timed EA gates priced above their owner's cap (69 free-tier, 7 bronze).
- Since `v5.0.2212`: **164 EA download purchases across 32 versions** — stored 780,892 ⚡ vs charged ~88,500 ⚡ → **~692K Buzz under-charged**.

Estimate uses *current* stored price and *current* owner tier, with the video 5× multiplier matched heuristically on base model. Order-of-magnitude, not accounting-grade.

**This PR stops the bleed; it does not back-fill.** Making the 32 affected creators whole is a separate call — buyers paid what the UI quoted them, so clawback isn't the right lever.

## Fix

Gate kind threaded through the shared helpers in `@civitai/buzz`, so all four surfaces follow one rule:

- `isPermanentGate(row)` reads **`timeframeDays`, not `endsAt`** — a timed gate on an unpublished version also carries `endsAt` NULL until publish materializes it, so `endsAt` can't tell the two apart.
- `paidAccessPriceCap({ permanent }, tier, mediaType)` → `Infinity` for a timed window.
- `effectivePaidAccessPrice` / `cappedTerms` take the gate kind as a **required** field, so the compiler forced every call site to decide rather than inherit a default.
- `assertPaidAccessCaps` returns early for timed gates — which also skips the `getPaidAccess` read those writes never needed. The two `permanent` branches are merged.

Clients follow the same rule: the onsite form's ceiling tracks the Access mode toggle (falling back to `MAX_DONATION_GOAL`, which also silences the over-cap warning and the upsell), and Creator Studio's editor derives `effectiveCap` from `ea.permanent`. `PaidAccessBulkBar` is untouched — it only ever sets permanent access.

Licensing-fee caps are a separate feature and stay tier-bound on every gate, including timed ones.

## Tests

`getViewerMonetization`'s fixture was a **timed** row (`timeframeDays: 7`) asserting 5000 → 500 — the suite encoded the bug, and 7 of its cases failed against the fix. Fixture flipped to permanent, plus new coverage: timed gate keeps its stored price, licensing fee still capped on a timed gate, no cap-tier fetch for a timed gate, and both directions of `assertPaidAccessCaps`.

- `@civitai/buzz` — 42 passed
- `paid-access.service.test.ts` + `paid-access-gating.test.ts` — 53 passed
- `licensing-fee.test.ts` — 9 passed
- `pnpm typecheck` — clean; Creator Studio `svelte-check` — 0 errors

No migration. No feature flag — this restores documented pre-cap behavior.

### DIFF
--- a/apps/creator-studio/src/lib/components/PaidAccessEditor.svelte
+++ b/apps/creator-studio/src/lib/components/PaidAccessEditor.svelte
@@ -73,11 +73,15 @@
   // Seed once from the version at mount (the parent remounts via `{#key version.id}` on open).
   let ea = $state(untrack(() => seed(version)));
 
+  // Membership caps bind PERMANENT paid access; a timed early-access window is priced freely at every
+  // tier (CU 868kk3avk), so the ceiling moves with the Access mode toggle.
+  const effectiveCap = $derived(ea.permanent ? priceCap : null);
+
   // max never drops below the stored price: the server blocks only RAISES, so clamping down would silently
   // cut a grandfathered price on any unrelated edit.
   const storedAccess = $derived(version.paidAccessConfig?.accessPrice ?? 0);
-  const accessMax = $derived(priceCap == null ? undefined : Math.max(priceCap, storedAccess));
-  const overCap = $derived(priceCap != null && (ea.accessPrice ?? 0) > priceCap);
+  const accessMax = $derived(effectiveCap == null ? undefined : Math.max(effectiveCap, storedAccess));
+  const overCap = $derived(effectiveCap != null && (ea.accessPrice ?? 0) > effectiveCap);
   // The generation-only tier can't exceed the access price, and neither may exceed the tier's ceiling.
   const genPriceMax = $derived(
     accessMax == null ? ea.accessPrice : Math.min(ea.accessPrice ?? accessMax, accessMax)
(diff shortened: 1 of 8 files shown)


manuelurenah:
Superseded — briant shipped the same fix in 9a0ec6c0cb + 2b39b66f35, already on `main`.

Independent convergence on every substantive point: `isPermanentGate(row)` keyed on `timeframeDays` (not `endsAt`, which can't tell a pending window apart), the gate kind as a required parameter so the compiler forces each call site to decide, and the same six surfaces — charge, display, write guard, onsite form, Creator Studio editor, Creator Studio server action. Both branches also independently found that the `getViewerMonetization` test fixture was a timed row encoding the bug.

His lands it better: the purchase path skips `getCachedCapTier` entirely for timed gates, the write-guard error now names the escape hatch ("use a timed early-access window"), and the new `monetization-limits.ts` resolves fee + access + permanent-count caps through one call instead of the narrow helper here.

Closing rather than rebasing — the only deltas left in this branch are two skipped cache lookups on cold-ish paths, not worth a PR.

PR #3529, opened 2026-08-01
feat(stickers): consumable sticker cosmetics for chat + comments

JustMaier:
Adds **stickers** — a purchasable, consumable cosmetic type usable in onsite chat/DMs and in comments. Community request from Liptonik Karlof.

Planning doc, decisions and scouting: [ClickUp 868kk3t0t](https://app.clickup.com/t/868kk3t0t).

## ✅ Migrations — already applied to dev AND prod

All four are live in both environments, so **there is no deploy-ordering requirement**. Recorded because two of them are code-depends-on-schema and would have broken cosmetic purchases app-wide if the image had shipped first.

```sql
ALTER TYPE "CosmeticType" ADD VALUE IF NOT EXISTS 'Sticker';
CREATE UNIQUE INDEX IF NOT EXISTS "Cosmetic_sticker_slug_key"
  ON "Cosmetic" ((data ->> 'slug')) WHERE "type" = 'Sticker';   -- after the enum commits
ALTER TABLE "CosmeticShopItem" ADD COLUMN IF NOT EXISTS "listed" BOOLEAN NOT NULL DEFAULT true;
ALTER TABLE "UserCosmetic"     ADD COLUMN IF NOT EXISTS "remaining" INTEGER;
```

ClickHouse `default.stickerUsageEvents` is also created. Flipt flag `stickers` is live, segments `testers` + `moderators`.

## What's in it

**Stickers as a cosmetic type.** `CosmeticType.Sticker`, one `Cosmetic` per sticker, `data: { slug, url, animated?, uses }`. Creators submit them through the existing Creator Shop — fee, `PendingReview`, mod approval, 70% payout — with a slug input and artwork validation: **128–512px on the long edge, at most 2:1 in either orientation**, transparency required.

**Consumable.** A purchase grants `uses`; `UserCosmetic.remaining` tracks the balance, `NULL` meaning unlimited so every pre-existing cosmetic is untouched. **A use is one placement** — three stickers in a comment is three uses. Free and unlimited in DMs; comments consume. Charging on edit is **net-new only**: a per-id diff, so adding charges, removing refunds nothing, and posting empty then editing can't dodge it. Spending drains across multiple holdings under `SELECT … FOR UPDATE`, sorted by id to avoid deadlock, all-or-nothing inside the caller's transaction.

**Rendering.** Chat parses `:sticker:<id>:` tokens; comments carry `<span data-type="sticker" data-id>`. A line containing only stickers renders them at 64px, mixed with text at 22px, falling back to inline past 6.

**Delisting.** New `CosmeticShopItem.listed`. Three positions: listed (buyable, bundlable), **delisted** (not buyable by anyone, still bundlable by other creators), withdrawn (`Archived`, neither). Groundwork for packs.

**Per-type price floors.** `cosmeticPriceFloor(type)` and `packItemFloor(type)`, every type seeded at 500 — structure now, tuning later.

**Flipt gate.** `stickers`, mod-by-default. Gates creation, purchase, shop visibility, and every insertion path — picker, `:` autocomplete and the input rule. **Never gates rendering** — a sticker already in a comment or DM stays visible regardless of flag state.

**Usage history.** `stickerUsageEvents` in ClickHouse, gated to production only: preview shares prod's `CLICKHOUSE_TRACKER_URL` but runs against the **dev** database, so preview rows would carry entity ids from a different database.

## Deliberately out of scope

**Packs** (own PR — design is in §4b of the doc), **P5 entity reactions**, **P6 image ingestion for sticker art**. Renaming this feature from "emoji" to "sticker" happened mid-build; the parser still accepts the legacy `:emoji:<id>:` token form and normalizes it on re-emit.

## The defect that took four rounds to kill — worth reading before extending this

Sticker markup is a `<span data-type="sticker">`. That shape survives the **default** sanitizer because `span[data-type][data-id]` is already allowed — **mentions need it**. Combined with the Tiptap node being registered in every editor and the `RenderHtml` hydrator being unconditional (33 consumers), a pasted span rendered as a paid sticker on ~30 surfaces, free and unowned. Copying a sticker out of a comment serializes exactly that span, so no tooling was required.

Four layers, each fix correct and each exposing the next:

1. The picker's absence on other surfaces was the *only* guard — there was no check.
2. The sanitizer allowed it by default → inverted to deny-unless-opted-in.
3. Five schemas never sanitized at all and relied entirely on `RenderHtml` re-sanitizing → the render-side opt-in turned "we sanitize on the way out" into "…except stickers".
4. `RenderRichText` — a *second* renderer — still drew them, and the article page uses it.

**What they share: the guard lived somewhere other than the thing being guarded** — in the UI, in a default, in the render path. The resolution puts it on the surface itself:

- `STICKER_SURFACES` (in the dependency-free `sticker-token.ts`) holds may-contain / consumes / storage-form per surface. **A surface absent from the table is denied everywhere.**
- `RenderHtml`'s `allowStickers` defaults to `false`; exactly four comment render sites opt in. That bounds the problem to ~4 render sites instead of ~30 write paths, and a surface added next quarter **fails closed**.
- Two guard tests assert the *class*, not the instance: the set of files granting `allowStickers` must **equal** those four, and the set referencing `StickerNode` must **equal** three. Both mutation-verified — adding a fifth fails the suite.

Write-side sanitization was kept and extended to eight previously-unsanitized schemas (`question`, `answer`, `bounty`, `changelog`, `model3d`, `bug`, `home-block`, and three shop descriptions) as hygiene — they were storing raw user HTML — but it's no longer the thing that has to be exhaustive.

## Two ways a sticker could be sold as unlimited — both closed

**Submit:** `uses` was `.optional()`, and the purchase grant reads the same field, so a sticker created without it granted buyers `remaining = null` — unlimited, at a finite price. Now a `superRefine` requires a positive integer when `type === Sticker`.

**Edit — the worse one, and reachable through the normal UI with no crafted request.** `patchCosmeticData` returns `artworkData` *wholesale* when the artwork changes: it rebuilds `Cosmetic.data` rather than patching it, and the rebuild call passed `type, imageUrl, animated, offsets, slug` — **not `uses`**. So a creator replacing the picture on a 100-use sticker silently made it unlimited for every future buyer. `nextUses` already resolved `requestedUses ?? existingUses` correctly two hundred lines earlier and was simply never handed over. Closing only the submit schema would have left this open — and the update schema can't carry the same refinement anyway, since it has no `cosmeticType` and `uses` is legitimately optional there (omitted means unchanged).

The creator grant takes the same fact seriously: `creatorGrantRemaining` **throws** rather than inventing a number, since a sticker reaching approval without `uses` has already been mis-selling.

## Also in this round

**Selling a sticker is now flag-gated.** All four *listing* paths already filtered `CosmeticType.Sticker` when the flag was off — but `purchaseShopItem` didn't refuse, so anyone with a shop item id could buy a sticker they then couldn't use, because the picker is gated. The gate lives in the service, where the cosmetic is already loaded. Tested three ways, including *badge + flag off still sells* — the regression that would otherwise go unnoticed. Worth naming why this was missed: **all four listing paths were correct, because filtering is a list operation and refusing is not.**

**`STICKER_SIZE.jumbo` 48 → 64.** The artwork floor is written as `STICKER_SIZE.jumbo * 2`, so the minimum long edge moves 96 → 128px with it — deliberate coupling, so the upload rule can't drift from what a 2× screen needs. Consequences, checked rather than assumed: the edit path does **not** re-run dimension validation (gated on `artChanged && imageUrl`), so no creator is locked out of editing a live item; and six wide jumbo stickers now reach 768px, so they **wrap** rather than overflow.

**`StickerEditNode` is registered on comment editors only.** Previously registered on every editor, so a pasted sticker rendered while editing a model description and vanished on save. Verified before gating: the four surfaces passing `includeControls: ['sticker']` are exactly the four comment editors, of 37 RTE consumers; nothing else loads comment content; and write-side sanitizers already strip spans elsewhere. Chat is unaffected either way — it has zero Tiptap references and writes literal `:slug:` into a textarea.

**Remaining uses now show in the picker and the autocomplete.** The picker was already rendering finite counts — `{remaining != null && …}` skipped **unlimited**, which is the common case, since every cosmetic predating consumables is `NULL`. Three states are tracked distinctly: `null` unlimited, a number, and `undefined` not-yet-loaded (collapsing the last two flashes "unlimited" before the real figure arrives). Chat shows no count at all, keyed off `STICKER_SURFACES[surface].consumes` rather than a new flag.

**DM placements are deliberately not logged** (D25). Today that falls out of DMs being free, so `charged` is empty — recorded at the gate and in the spec so it isn't "fixed" later. DMs are private space; usage metrics on them would be an invasion, not a feature.

## Still open, deliberately

**Sticker artwork bypasses image ingestion and NSFW scanning** (P6) — one mod eyeballs it at approval and that is the entire review. **There is no takedown path**: archiving stops sales, but every already-placed sticker renders forever. **Stickers aren't separately reportable** — the comment carrying one is. These are accepted for a mod+testers rollout and are a known gap across *all* creator cosmetics, to be solved once for all of them rather than per-type.

**`stickerUsageEvents` has never executed** — preview deliberately doesn't emit (see below), so the first real run is production. Worth eyeballing on day one rather than assuming.

**The deduction path is unverified by a human.** The drain has unit coverage; nobody has watched a balance go 1000 → 999 on a real placement.

## Creator grant, blur, and editor rendering

**Creators receive 10× the per-purchase uses** on approval (`CREATOR_GRANT_USES_MULTIPLIER`), so 100 uses per purchase grants the creator 1,000. Stickers only — every other cosmetic type still grants `NULL`, which is asserted across all four, since that's the regression that would otherwise go unnoticed. The submit modal states the figure live off the uses field.

**Opening the picker no longer blurs the composer.** Two bugs, not one: the "cannot be empty" alert, and the **caret** — losing focus meant the sticker landed wherever `focus()` put it rather than where the author was typing. Fixed once at the picker via `preventDefault` on mousedown, so all six mount points inherit it.

**The editor draws the sticker at comment sizes**, jumbo/inline derived from the ProseMirror block and recomputed per transaction. The silent failure it avoids: typing beside a sticker changes its parent block but not the node, so a size computed once would stick.

## A note for anyone testing this on a preview

Stickers appeared completely broken on the preview — picker inserting `:sticker:`, comments posting empty — and **none of it was this code**. Preview reads **prod's** Redis while running the **dev** database. Prod had a ~3h-old `notFound` tombstone cached for a cosmetic that only exists on dev, and preview served that instead of doing its own lookup. Purging the key fixed it instantly. Filed as [868kk78jd](https://app.clickup.com/t/868kk78jd).

The asymmetry worth knowing: a **failing** Redis degrades gracefully to the database; a **working** Redis holding a tombstone does not. That's why it presented as total feature failure rather than a slowdown. It will recur for any preview that adds rows until cache keys are namespaced per environment.

## Known open

- **Should stickers ever be allowed outside chat and comments?** Today the answer is no by default — which is now a decision rather than an accident. Allowing a new surface is an explicit opt-in plus a charging decision.
- Filed separately, pre-existing and unrelated: [868kk5ahg](https://app.clickup.com/t/868kk5ahg) — any user can break their own article page via crafted `contentJson`, because `RenderRichText` doesn't tolerate unknown node types.

## Review

Eleven adversarial rounds. Two blockers, both found only by looking across chunks rather than within them: ownership enforced client-side only, and the sanitizer surface bypass above. Notable also-founds: a same-key overwrite inside an object spread silently replacing a caller's Prisma filter (four instances, one live — `/shop?cosmeticTypes=Badge` returned everything), a slug silently wiped on artwork swap, and `getStickerBalances` using `MAX(remaining)` where the drain spends `SUM` — a function nothing called, so nothing caught that it had gone stale.


JustMaier:
## ⚠️ Deploy ordering — the `listed` migration must land BEFORE the image

This one is not like the other migrations on this branch.

`ADD VALUE 'Sticker'` and the slug unique index are additive in the safe direction — old code keeps working with or without them, so ordering is forgiving.

**`CosmeticShopItem.listed` is code-depends-on-schema.** `purchaseCosmeticShopItem` selects the column, and `creatorShopItemSelect` includes it. Prisma throws when selecting a column that doesn't exist, so shipping the image against a database without it means:

- **every cosmetic purchase fails — including the official shop**, because that's the shared purchase path, not just creator-submitted items
- the entire creator manage page fails (`getManageItems`, submit, update and review all run through that select)

So: **apply the migration first, in every environment, then deploy.**

```sql
ALTER TABLE "CosmeticShopItem" ADD COLUMN IF NOT EXISTS "listed" BOOLEAN NOT NULL DEFAULT true;
```

Idempotent, and Postgres backfills existing rows from the default without a table rewrite. The column is `NOT NULL`, so no read can observe a null and no backfill script is needed.

**Status: applied and verified on dev.** Prod still needs it.

Note the `civitai` role cannot run this — these need the cnpg superuser:

```bash
kubectl exec -n cnpg-database-<env> cnpg-cluster-<env>-1 -- psql -U postgres -d civitai -c '<sql>'
```

### Full migration order for this branch

1. `ALTER TYPE "CosmeticType" ADD VALUE IF NOT EXISTS 'Sticker';`
2. `CREATE UNIQUE INDEX IF NOT EXISTS "Cosmetic_sticker_slug_key" ON "Cosmetic" ((data ->> 'slug')) WHERE "type" = 'Sticker';` — must run **after** 1 has committed; the enum literal isn't visible inside the same transaction
3. `ALTER TABLE "CosmeticShopItem" ADD COLUMN IF NOT EXISTS "listed" BOOLEAN NOT NULL DEFAULT true;` — **before the image ships**

(1 and 2 are named `Emoji` on the currently-pushed commits; a rename to `Sticker` is landing shortly and rewrites both migrations in place rather than stacking a `RENAME VALUE`, so prod gets one clean statement each.)

PR #3533, opened 2026-08-02
fix(monetization): clear licensing fee to null on removal + restore payouts

JustMaier:
## Problem
Removing a licensing fee left the creator earning **nothing** — reported by IcebergMM (ClickUp 868kk4j2k). His IcebergMix v3.0 (ModelVersion 3184124) had `licensingFee = 0.00` while his other versions were correctly NULL. Manually nulled + orchestrator-cache-busted to unstick him; this fixes it for everyone.

## Root causes (in the version upsert, `model-version.service.ts`)
1. **Fee cleared → persisted `0`, not `NULL`.** A stored `0` reads as "has a fee" and pays out 0 instead of restoring base compensation + tips. Now any non-positive/absent fee is coerced to NULL (`licensingFee`, `licensingFeeType`, `licensingFeeSettlementCurrency`).
2. **`DisablePayout` flag never stripped on clear.** Setting a fee adds the version-level `ModelVersionFlag.DisablePayout` (opts out of tip + creator-comp payouts). The old code only ever ADDED it, with a comment assuming a creator might set it independently — but this flag has **exactly one set site** (the fee path; confirmed by grep). So clearing the fee can and must strip it, or the version keeps earning nothing after removal. `UserFlag.DisablePayout` (a separate, moderation flag) is untouched.
3. **Version-save path didn't bust the orchestrator cache.** Wired `bustMvCache(...)` to fire when the fee or payout flag actually changed, so the corrected value propagates to generation pricing/payout instead of the orchestrator repricing against stale data.

## ⚠️ For Briant — please review
- **The `DisablePayout` strip (cause 2)** reverses a deliberate "only ever add" behavior. I'm confident it's correct (single set site, and it's the actual reason removal leaves earnings at zero), but it changes payout-flag semantics, so please confirm there's truly no other path that sets the version-level flag.
- **Note on the orchestrator cache bust:** `bustOrchestratorModelCache` in `services/orchestrator/models.ts` is **already working on main** (it was restored after being commented out for ~2.5 months). This PR does NOT touch it — it only *invokes* it (via `bustMvCache`) on the fee-change path, which wasn't happening. If you'd prefer a lighter, orchestrator-only bust here instead of full `bustMvCache`, easy to swap.

Do not merge without Briant's review of the payout-flag change.

### DIFF
--- a/src/server/services/model-version.service.ts
+++ b/src/server/services/model-version.service.ts
@@ -485,12 +485,23 @@ export const upsertModelVersion = async ({
     ? (await dbRead.modelVersion.findUnique({ where: { id }, select: { flags: true } }))?.flags ?? 0
     : 0;
 
-  // Versions with a license fee earn through that channel, so they opt out of
-  // tip + creator-comp payouts. We only ever ADD the flag — clearing the fee
-  // doesn't strip it, since a creator may have set the bit independently.
+  // Clearing a fee must persist NULL, not 0: a stored 0 still reads as "has a
+  // fee" and pays out 0 instead of restoring base compensation + tips.
+  const hasLicensingFee = data.licensingFee != null && data.licensingFee > 0;
+  if (!hasLicensingFee) {
+    data.licensingFee = null;
+    data.licensingFeeType = null;
+    data.licensingFeeSettlementCurrency = null;
+  }
+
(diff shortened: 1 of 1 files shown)


bkdiehl:
Closed as this has been finished already in another branch

PR #3534, opened 2026-08-02
feat(cosmetics): perceptual-hash cosmetic artwork

JustMaier:
Gives every cosmetic a perceptual hash, so near-duplicate artwork can be detected. Groundwork for catching a creator lifting someone else's badge, re-saving it, and submitting it as their own.

Tracking: [ClickUp 868kk4fxd](https://app.clickup.com/t/868kk4fxd).

## ⚠️ Release gate — apply the DDL to staging and prod BEFORE the image rolls

Verified: **neither column exists on prod today.**

```sql
ALTER TABLE "Cosmetic" ADD COLUMN IF NOT EXISTS "pHash" BIGINT;
ALTER TABLE "Cosmetic" ADD COLUMN IF NOT EXISTS "pHashUrl" TEXT;
```

Both are in the generated Prisma model, so **every unselected `Cosmetic` query will `SELECT` them.** Shipping the image first is a P2022 on the event-badge award path (`base.event.ts:59`), creator-shop cosmetic submit/update (`creator-shop.service.ts:439`), membership-badge upsert, and the mod cosmetic admin.

Idempotent, nullable, no default, no table rewrite — safe to re-run and safe on a hot table. The `civitai` role can't run DDL; use the cnpg superuser:

```bash
kubectl exec -n cnpg-database-<env> cnpg-cluster-<env>-1 -- psql -U postgres -d civitai -c '<sql>'
```

**Already applied and verified on dev.**

## Why this shape

The problem was never "we need hashing" — the platform already has perceptual hashing via the orchestrator. Cosmetic artwork just never gets one, because it's uploaded straight to Cloudflare Images and never becomes an `Image` row.

Deliberately **not** creating `Image` rows for cosmetics: that would pull in feed visibility, `nsfwLevel`, the ingestion state machine, moderation queues and deletion cascades. Instead this calls the `mediaHash` orchestrator step directly and stores the result on the cosmetic — the same one-step synchronous pattern `resizeBadgeImage` already uses.

Measured against prod during development: **100-700ms per hash**, animated artwork hashes fine, and the conversion is bit-identical to what the orchestrator lane writes for images (4/4 recomputed images matched exactly, Hamming 0).

## `pHashUrl` — why there are two columns

Several paths replace a cosmetic's artwork with a raw `UPDATE` that bypasses any hook — including `creator-shop.service.ts:439`, which is exactly the creator-facing edit path.

Tracking *who wrote the url* fails there. So the row records **what was actually hashed**, and any disagreement with `data.url` is drift regardless of origin:

```sql
WHERE (data->>'url') IS NOT NULL
  AND (data->>'url') <> ''
  AND ("pHash" IS NULL OR "pHashUrl" IS DISTINCT FROM data->>'url')
```

Without it: creator submits art A → hashed → creator swaps in art B (a re-encode of someone else's badge) → `pHash` still describes A, and `"pHash" IS NULL` never revisits it. A wrong hash is worse than no hash, because matching then looks for artwork that isn't there and reports clean.

`IS DISTINCT FROM` rather than `<>` so rows hashed before this column existed (`pHashUrl` NULL) still sweep.

This also makes the write race benign: a late-landing hash for superseded artwork writes `(h(A), A)` against `data.url = B`, which is drift, which sweeps. Recording what was hashed converts a class of races into detectable staleness.

## What's in it

- `getPerceptualHash(url)` — one-step `mediaHash` workflow via the internal orchestrator client. `currencies: []`, system token, no user identity, nothing billable. `wait: 30` plus `AbortSignal.timeout(45_000)`; a timeout degrades to `undefined` rather than throwing.
- Hash on write in `upsertCosmetic`, `createCosmetic`/`updateCosmetic` (which covers the Retool endpoint without touching it), and `upsertProductBadge` — re-hashing only when `data.url` actually changes.
- Backfill at `/api/internal/add-missing-cosmetic-phash` (`ModEndpoint`, GET). `limit` default 100 / max 5000, `concurrency` max 20, `afterId` cursor with `lastId` returned so an operator can page past permanently-unhashable rows.
- `computePerceptualHash` de-duplicated — exported from `orchestrator.service.ts`, imported by `image-scan-result.service.ts`.

Prod corpus: **1,082 cosmetics, 960 with artwork, 844 mod-authored.** Those mod-authored ones are the badges most visible on profiles and therefore the likeliest theft targets — and none have any hash today.

## Deliberately out of scope (phase 2)

- Near-match lookup at submission and match surfacing in the mod review queue. Those live in `creator-shop.service.ts` / `moderator/creator-shop.tsx`, which are being restructured by the sticker PR — writing against them now means writing against code that's about to change.
- `submitCreatorShopItem` isn't hooked for the same reason. It does `tx.cosmetic.create` inline; new creator submissions land `pHash` NULL and get swept by the backfill.
- The existing sha256 exact-match block stays as the cheap first gate.

## Two constraints phase 2 must carry

**1. Gate on `pHashUrl = data->>'url'`.** What this PR establishes is *"a row is either correct or detectably stale"*, not *"a row is correct"* — nothing schedules the backfill; it's a manual endpoint. So the near-match lookup must treat a `pHashUrl` mismatch as **unverified** rather than trusting `pHash`, or a submitted-then-swapped cosmetic reads as clean between sweeps. Making the sweep a cron would be a good idea too.

**2. Do not compare cosmetic hashes against `Image.pHash`.** `Image.pHash` holds **two different perceptual-hash algorithms** — legacy rows recompute to Hamming 21-36 (i.e. noise) and are ~100% negative versus 61.6% for the current lane. Pre-existing, unrelated to this change, filed separately as [868kk4jdr](https://app.clickup.com/t/868kk4jdr) — and it likely means blocked-image matching is already largely non-functional. The `getPerceptualHash` docblock carries this caveat inline.

Also assumed and currently true: **cosmetic urls are content-addressed and immutable.** All 1,082 prod cosmetics use Cloudflare Images ids, so bytes behind a url can't change. Drift detection keys on the url string, so allowing external artwork URLs later would open a same-url content swap.

## Review

Two adversarial rounds. Notable: the second reviewer, run independently over the same code, found four findings the first missed — including `z.coerce.boolean()` making `?dryRun=false` perform a dry run, which was the exact first command intended before any real backfill.


JustMaier:
**Final review pass: go, clean.** Two notes worth recording on the PR rather than losing in a thread.

## Correction: bad query params return 500, not 400

The `dryRun` enum genuinely *rejects* unrecognised values rather than silently guessing — that's the half that mattered, and it's fixed. But rejection surfaces as an opaque **500**, not a 400: `schema.parse` throws a ZodError inside the handler, and `ModEndpoint` / `withApiMetrics` (`endpoint-helpers.ts:44-54, 219-235`) have no try/catch, so Next returns a 500 with nothing naming the bad param.

Same for out-of-range `limit`/`concurrency` and non-numeric `afterId`. Pre-existing shape for this endpoint family and purely operator ergonomics — a `safeParse` with a 400 would be nicer. Not held for it.

Verified by executing the schema against real inputs:

```
"true" -> true      "false" -> false     "1" -> true
"0"    -> false     ""      -> true      undefined -> undefined
"yes"  -> REJECTED  "TRUE"  -> REJECTED  "2" -> REJECTED
["true","false"] (repeated ?dryRun) -> REJECTED
```

## Pre-existing, out of scope: `0n` skips the blocklist check on the image path

`src/server/services/image-scan-result.service.ts:319` guards with `if (!mediaRating.isBlocked && pHash)`. That's the same falsy-`0n` shape fixed on both cosmetic sites in this PR — a solid-colour image hashes to `0n`, which is falsy, so it skips the blocked-image check entirely.

Predates this branch, untouched by it, and the DB write path handles `0n` correctly (`COALESCE(${pHash ?? null}, "pHash")`). Impact is limited to all-zero hashes. Recorded here so it isn't lost; it belongs with the blocked-image work in [868kk4jdr](https://app.clickup.com/t/868kk4jdr) rather than in this PR.

## The `computePerceptualHash` move is provably cycle-free

Since it touches a live image-scanning path: `image-scan-result.service.ts` is imported by exactly one module in the tree — the webhook page `src/pages/api/webhooks/image-scan-result.ts` (plus its test). Nothing under `src/server/services/` imports it, so nothing reachable from `orchestrator.service.ts` can reach back, and the new edge cannot close a loop regardless of how that module's imports evolve later.

The moved body is character-for-character the original; the only textual change is the added `export`. The call site is untouched, and the webhook's 6 tests pass.

JustMaier:
## Backfill run against prod — done

```
[backfill-cosmetic-phash] target=prod db=localhost:25060/civitai orchestrator=https://orchestration.civitai.com mode=prod
pending: 961
[hash] 962/961 (100.1%) hashed=959 failed=3 rate=10.5/s
done. hashed=959 failed=3 — failed ids: 240, 241, 242
```

~90 seconds. Verified against prod afterwards: `hashed=959, labelled=959, still_pending=3`. **Equal `pHash` and `pHashUrl` counts** — the pairing invariant held across the whole corpus, which is what everything else in this PR rests on.

The 3 failures are cosmetics whose artwork 404s on the CDN. Correct outcome: no hash, and they stay in the sweep set.

**Prod and dev DDL are both applied**, so the release gate at the top of this PR is satisfied.

## It found something on the first run

834 distinct hashes across 959 rows:

| | |
|---|---|
| duplicate-hash groups | 72 (197 rows) |
| **cross-creator groups** | **3** |

The 69 same-creator groups are almost certainly legitimate — tier variants of one design, seasonal recolours. The cross-creator ones are the signal this feature exists to produce:

- `pHash 3634669549584969906` → cosmetics **323** (mod-authored) and **1140** (creator 994880)
- `pHash 7043629817207563025` → cosmetics **1105** and **1106**, creators **149676** and **11579707**

Two caveats on reading these. It's an **exact** hash match, i.e. the strictest possible test — phase 2's Hamming threshold will surface more, not fewer. And solid-colour or near-blank artwork can collide legitimately. These want a human look before anyone concludes anything.

## First-run notes

The first attempt failed 960/960 and wrote nothing, which is worth recording because the design is what made it free:

`~/env/client` snapshots `process.env.NEXT_PUBLIC_*` at module evaluation, while `~/env/server` calls `dotenv.config()` in its module *body* — which ES semantics run after its own import of `~/env/client`. Next inlines these at build so the app was never affected, but in a tsx process `NEXT_PUBLIC_IMAGE_LOCATION` was undefined. `getEdgeUrl` then does `[env.NEXT_PUBLIC_IMAGE_LOCATION, src, …].filter(Boolean).join('/')`, silently dropping the missing base and returning a **relative path**. The orchestrator couldn't fetch it and failed every row in ~230ms.

Two things came out of that:

1. **`getPerceptualHash` now refuses a non-absolute media url**, logging the missing variable by name. That function is on the write path too — previously, an unset env var server-side would have meant every cosmetic silently stopped being hashed with nothing to signal it. A wrong URL and "the orchestrator couldn't hash this" no longer look alike to a caller.
2. The script loads dotenv before a dynamic import of the hashing service, so the ordering is explicit rather than incidental.

**A total failure against prod cost nothing** — 0 rows written, all 960 still pending, cursor irrelevant, re-run a clean retry. That's the "correct or detectably stale" property doing its job.

## Worth raising separately

`NEXT_PUBLIC_IMAGE_LOCATION: z.string().default('')` in `~/env/client-schema.ts` means a **missing** variable validates successfully as `''`. Combined with `.filter(Boolean)` downstream, that's two stacked silent fallbacks producing a plausible wrong value. Not specific to cosmetics — every `NEXT_PUBLIC_*` with a `.default('')` has the same shape.

Issue #3520, opened 2026-07-31
Silent model substitution: modelLocked ecosystems swap out-of-list versions for the workflow default — deliberate for the UI, unobservable (and billed) via the API


ZacxDev:
> **Rewritten.** The first version of this issue framed the substitution as simply a bug and proposed rejecting out-of-list ids. That was too blunt: the behaviour is **deliberate graceful degradation**, and strict rejection alone would break apps on model retirement. The revised framing and proposal are below.

## The behaviour

On `modelLocked` ecosystems, the generation graph **replaces any model version id that isn't in the current workflow's visible list with that workflow's `defaultModelId`**. The request succeeds, images are returned, the user is billed — and nothing in the response reports that a different model ran.

`createCheckpointGraph`'s `checkpointInputSchema` (`src/shared/data-graph/generation/common.ts:979`):

```ts
// When model is locked, force ecosystem default.
// Stale stored values (from localStorage after ecosystem switches) are rejected.
if (modelLocked && modelVersionId && val.id !== modelVersionId) {
  if (!validVersionIds?.has(val.id)) {
    return { id: modelVersionId, model: { type: 'Checkpoint' } };
  }
}
```

## This is intentional, and correct — for the UI

Read the comment: this exists for the **on-site generator**, where the only way to hold an out-of-list id is a stale `localStorage` value after an ecosystem switch. Snapping back to the default is good UX there. It is not a bug that someone forgot to fix; it is a deliberate resilience mechanism.

Graceful degradation is the right call when two things hold:

| | on-site UI | bridge / API |
|---|---|---|
| Was the input deliberate? | **No** — stale localStorage | **Yes** — an app author wrote that id |
| Can the caller observe the correction? | **Yes** — the picker visibly snaps | **No** — nothing reports it |

The UI satisfies both. The API satisfies neither: it silently overrides an explicit instruction with no way to detect it. Same code, same behaviour, correct in one context and wrong in the other — the assumption simply wasn't revisited when the graph became API-driven via the App Blocks bridge (`buildImageWorkflowInput` emits `model: { id: <the id the block asked for> }`).

## Evidence (executed, not inferred)

Real `generationGraph.safeParse`, Qwen ecosystem, free-tier ctx:

| workflow | input version | output version | note |
|---|---|---|---|
| `txt2img` | `2133258` (edit v2509) | **`2552908`** | discriminator — see below |
| `txt2img` | `2558804` (edit v2511) | **`2552908`** | |
[...]

PR #3535, opened 2026-08-02
feat(generation): make silent checkpoint substitution observable (#3520 phase 1)

ZacxDev:
Closes phase 1 of #3520.

## 🔴 Fix round (adversarial audit)

Five findings, all addressed. The no-behaviour-change property was **re-proved after the changes**, not assumed — see *Re-proof* below.

### 1. `modelSubstitutions` disappeared before any consumer could read it — **fixed via orchestrator workflow metadata; NO DB change**

As originally built, the field was **submit-reply-only**. `pollWorkflow` (`blocks.router.ts`) and `cancelWorkflow` call `snapshotFromWorkflow(workflow)` with no `extra`, and the normal block pattern renders from the **terminal poll** — the only snapshot that carries `imageUrls`. So the record was gone by the time there was anything to display beside it, and `block_workflows` does not retain the submitted body, making it unrecoverable. Phase 1 did not achieve its own goal.

**Route taken: the orchestrator workflow's own `metadata`.** `createBlockTextToImageStep` already returns a `workflowMetadata` that the real submit attaches to the submit body; the record is now folded into it under `modelSubstitutions`, and `snapshotFromWorkflow` reads it back off `workflow.metadata` when no explicit `extra` is supplied.

Why this works:

- `WorkflowTemplate.metadata` is `?: null | { [key: string]: unknown }` and `Workflow.metadata` is `{ [key: string]: unknown }` — a free-form bag the orchestrator persists and returns on every read.
- **The app already depends on this exact round-trip for non-standard top-level keys**: `remixOfId` and `isPrivateGeneration` are written into workflow metadata at submit and read back from `workflow.metadata` in the workflow formatter (`orchestration-new.service.ts`). This is an exercised path, not a new assumption.
- The formatter builds its normalized metadata from a **fixed key list**, so an extra key is inert on the on-site normalization path.

**No Prisma migration, no schema change, no new column.** Given that this org applies migrations manually per environment, turning a code PR into a coordinated DB operation was not warranted when a durable carrier already existed on the submit body.

Precedence: an explicit `extra` (submit / estimate replies, where the request's own collector is in scope) wins; otherwise the metadata is read. The metadata read is **validated, not cast** — it crosses a service boundary and feeds a public wire field whose `reason` is also a bounded prom label, so every entry is structurally checked and its `reason` narrowed against the code-owned union, with malformed entries dropped.

Where the field now appears: the submit reply, the `estimateWorkflow` reply, the insufficient-budget reply, **and every subsequent `pollWorkflow` / `cancelWorkflow`**. The estimate path is the one exception — a whatIf creates no persisted workflow, so there it lives only on the reply.

**Checked, because putting a new field on the POLL reply is riskier than putting it on the submit reply:** the block SDK's inbound validator cannot drop a snapshot because of it. `isValidWorkflowSnapshot` (`@civitai/blocks-react`, `internal/validate.js`) is a whitelist of per-field checks over `workflowId` / `status` / `cost` / `imageUrls` / `error` / `autoClaim` and then `return true` — it has no unknown-key rejection. And `extractSnapshot` in `internal/liveHost.js` returns the parsed `snapshot` object **by reference**; nothing re-picks its fields. So an additive key neither invalidates the snapshot nor is stripped at runtime — only the SDK's TypeScript type needs widening before a block can read it, as noted below.

🔴 **Not verified locally:** that the orchestrator round-trips this specific key end-to-end. The evidence is the type contract plus the two existing keys that already rely on it; nothing here was exercised against a live orchestrator.

### 2. The metric was a contaminated signal — **`surface` label added**

`validateInput` is the choke point for *every* server-side graph validation, so one un-split series mixed three populations whose verdicts are opposite. The on-site generator's substitution is the behaviour #3520 explicitly defends as **correct**; preset/comics generation is server-composed input; only the App Blocks bridge is the population phase 3's decision is about.

`civitai_generation_model_substitutions_total{reason, surface}` — `surface` ∈ `block` | `onsite` | `preset`. **3 × 3 = 9 series, total, forever.** No version id, ecosystem, workflow, app id or user id.

The surface is supplied **by the caller** at context-build time: `buildGenerationContext(userTier, flags, user, surface)`, where the parameter is **required, not defaulted** — a new generation entry point is a compile error rather than silent misattribution. All three call-site files are wired (`orchestrator.router.ts` → `onsite`, `blocks.router.ts` → `block`, `preset-image-gen.service.ts` → `preset`), and a population guard enumerates every `buildGenerationContext(` call site from the tree and fails on an unknown one.

**Honest multiplier.** One increment is **one graph validation that substituted** — not one generation, and *not a fixed multiple of one either*. On the `block` surface, for one user-visible generation:

| term | when |
|---|---|
| **2** | the normal successful submit (whatIf cost preflight + real submit, each with its own per-request collector) |
| **1** | when the whatIf estimate exceeds the block's per-call buzz budget — the procedure returns before the idempotency claim and the real submit |
| **+1 each** | per `blocks.estimateWorkflow` call the block chooses to make — a separate procedure a block may call any number of times, so this term is **unbounded** |
| **+1** | per idempotency **replay** that re-enters the procedure (the preflight re-runs before the claim short-circuits) |

So: **read it as a rate and a trend, per surface.** It is not a headcount of affected generations, and dividing by 2 does not produce one. A real "generations served the wrong model" figure has to come from the snapshot/log side, where each record is tied to one workflow id. This is now stated in the metric's own help text, not only in the source comment.

### 3. The in-generate preflight discarded its substitutions

`blocks.router.ts` destructured only `{ step: stepForCostCheck }`, so the validation that produced the cost the budget check is made against threw its record away.

**Surfaced on the insufficient-budget reply**, and deliberately only there. On the success path the real submit re-validates the same input and its record is what the snapshot carries (identical by construction, and tied to a real workflow id) — adding a second copy would be redundant. The budget-reject exit is the one reply in the whole flow that returns a **cost with no workflow**, so without this it is the only place a block is quoted a price with no way to learn which model it was for.

### 4. `seen.add` ran before `classify`

A classifier throwing on its first call **permanently dropped that key for the request** — the throw was swallowed, nothing was recorded, and the retry short-circuited on `seen.has(key)`. `seen.add` now runs only after a successful push, so a transient classifier fault leaves the key retryable. Harmless with today's pure classifier; free to make correct.

### 5. No runtime narrowing of the classifier's return

A classifier returning `undefined` would have put `reason: undefined` on the wire **and** into `inc({ reason: undefined })` — which is how a "bounded forever" label set stops being bounded. Both unions are now narrowed **at runtime** (`isModelSubstitutionReason` / `isGenerationSurface`), in two places guarding two different exits: `record()` drops an out-of-union reason before it can reach the snapshot, and `recordGenerationModelSubstitution` drops an out-of-union reason *or* surface before `inc()`. The cardinality bound now rests on code, not on erased types.

---

## 🔴 Re-proof: no behaviour change, after the fixes

New file `model-substitution.no-behaviour-change.test.ts` parses the **full result object** (serialized with sorted keys, `undefined` made explicit) with and without a collector — a context with no `modelSubstitutions` field *is* the pre-change shape, so this is a direct before/after comparison, not a proxy.

| sweep | population | comparisons | diffs |
|---|---|---|---|
| real classifier | **552** (ecosystem, workflow) pairs across 82 ecosystems | 684 | **0** |
| classifier that **throws on every call** | all **89** modelLocked pairs | 178 | **0** |
| classifier returning a **bogus reason** | all 89 modelLocked pairs | 89 | **0** |

Plus a **negative control** in the same file: two genuinely different parses must compare unequal, so a "zero diffs" result cannot come from a blind comparator. And mutation **N20** injects a real perturbation into the clamp when a collector is present — the sweep goes red with `expected [ 'Flux1Kontext/txt2img', …(39) ] to deeply equal []`, so the instrument is proven able to see the thing it reports absent.

## Mutation results — 20/20 killed, each by its OWN guard's assertion

| # | mutation | killed by | assertion |
|---|---|---|---|
| N1 | `seen.add(key)` moved back before `classify` | `a classifier that throws ONCE does not permanently drop the key`; `an always-throwing classifier still leaves the key retryable` | `expected 1 to be 2` |
| N2 | delete the reason narrowing in `record()` | 5× `DROPS a classifier return that is …` | `expected [ { requested: 1, applied: 2, …(3) } ] to deeply equal []` |
| N3 | invert the narrowing predicate | 16 tests incl. all three reason reproductions | `expected [] to deeply equal [ { requested: 2133258, …(4) } ]` |
| N4 | `isModelSubstitutionReason` accepts everything | 5× `DROPS a classifier return …` + `isModelSubstitutionReason accepts exactly the declared union` | `expected true to be false` |
| N5 | delete the label narrowing before `inc()` | 3× `DROPS an increment whose surface is …`, 2× `… whose reason is …` | `expected Counter{ labelNames: [ …(2) ], …(5) } to be undefined` |
| N6 | drop `surface` from the counter's `labelNames` | 17 tests incl. `DECLARES exactly reason + surface` | `expected [ 'reason' ] to deeply equal [ 'reason', 'surface' ]` |
| N7 | emitter hardcodes the surface | `labels every increment with the COLLECTOR's surface, not a guess` | `expected 2 to be 1` |
| N8 | collector ignores its `surface` argument | `carries the SURFACE it was constructed with` (+3) | `expected 'onsite' to be 'block'` |
| N9 | block call site labelled `onsite` | `the App Blocks bridge builds its generation context with surface block`; the wiring guard | `expected 'onsite' to be 'block'`; `expected [ 'onsite' ] to deeply equal [ 'block' ]` |
| N10 | preset call sites labelled `block` | `submitPresetImageGen builds its context with surface preset` (+2) | `expected 'block' to be 'preset'`; `expected [ 'block' ] to deeply equal [ 'preset' ]` |
| N11 | on-site call sites labelled `block` | `orchestrator.router.ts passes the surface onsite at EVERY buildGenerationContext call` | `expected [ 'block' ] to deeply equal [ 'onsite' ]` |
| N12 | snapshot stops reading `workflow.metadata` | `pollWorkflow REPORTS the substitutions recorded at submit` (+4) | `expected undefined to deeply equal [ { requested: 2558804, …(2) } ]` |
| N13 | metadata `reason` not validated | `DROPS an unknown reason from the metadata …` (+2) | `expected true to be false` |
| N14 | metadata `requested`/`applied` not validated | `DROPS a non-numeric requested …`; `DROPS a NaN applied …` | `expected true to be false` |
| N15 | `extra` no longer wins over the metadata | `an EXPLICIT extra wins over the metadata` | `expected [ { requested: 2558804, …(2) }, …(1) ] to deeply equal [ { requested: 1, applied: 2, …(1) } ]` |
| N16 | submit stops folding the record into `workflowMetadata` | `the REAL submit body carries the substitutions on workflow metadata` | `expected { params: { prompt: 'a cat' } } to deeply equal { params: { prompt: 'a cat' }, …(1) }` |
| N17 | submit writes the key even when empty | `does NOT touch the metadata when nothing was substituted` | `expected { params: …, …(1) } to deeply equal { params: { prompt: 'a cat' } }` |
| N18 | budget-reject reply drops the record | `the insufficient-budget reply reports what the quote was actually priced for` | `expected undefined to deeply equal [ { requested: 2558804, …(2) } ]` |
| N19 | budget-reject reply emits an empty array | `the insufficient-budget reply is unchanged when nothing was substituted` | `expected true to be false` |
| N20 | clamp perturbs behaviour when a collector is present | the 3 full-population sweeps | `expected [ 'Flux1Kontext/txt2img', …(39) ] to deeply equal []` |

Each mutation was checked to die to **its own** guard's assertion, not a neighbour's.

## Local verification (NixOS; CI is authoritative)

Measured in a clean worktree with the repo's own toolchain (`node_modules/.bin/tsc` 5.9.2, `NODE_OPTIONS=--max_old_space_size=8192`), against a same-environment baseline worktree at the pre-fix commit `f5e6099`.

- **tsc**: `0` errors at the baseline, `0` errors after the fixes — **delta 0**. (This environment has a generated Prisma client, so the count is a real 0 rather than the stale-client artifact reported earlier in this PR.)
- **eslint** on all changed files: baseline `26 problems (21 errors, 5 warnings)` → after `26 problems (21 errors, 5 warnings)`. Identical set; all pre-existing `no-explicit-any` / `no-unused-vars` in untouched code.
- **prettier `--list-different`**: the only non-conforming files are the four that were **already** non-conforming at the baseline commit (`blocks.router.ts`, `workflow.schema.ts`, `blocks.router.workflow.test.ts`, `workflow.service.test.ts`) and were deliberately not reformatted. Every prettier hunk in those files was intersected against this change's added-line ranges: **0 overlap** — none of them are mine.
- **vitest `--project unit`** over `src/shared/data-graph src/server/metrics src/server/services/blocks src/server/services/orchestrator src/server/routers`:
  - baseline (`f5e6099`, same env): **165 files, 3236 tests, 0 failing**
  - after: **168 files, 3294 tests, 0 failing**
  - net **+3 files, +58 tests**, all passing, no regressions.

### Not verified locally

- **Nothing was run against a live orchestrator.** The metadata round-trip that fix 1 depends on is argued from the type contract plus the two existing top-level keys (`remixOfId`, `isPrivateGeneration`) the app already writes and reads back — it has **not** been observed end-to-end. That is the one assumption worth checking on staging before relying on the poll-side field.
- The counter has still not been observed on a real `/api/metrics` scrape — only against the prom-client default registry in-process.
- No end-to-end block submit was driven; the bridge plumbing is covered by unit tests on both ends.
- The `onsite` surface has no router-level behavioural test (`orchestrator.router.ts` has no existing test harness); it is covered by the source-level population guard, while `block` and `preset` have behavioural tests that read the argument actually passed.

---

## The behaviour, unchanged

On a `modelLocked` ecosystem, `createCheckpointGraph`'s `checkpointInputSchema` replaces any checkpoint version id that is not in the current workflow's visible list with that workflow's `defaultModelId`. The parse succeeds, images return, the user is billed — and nothing in the response, the snapshot, or the `block_workflows` read-model says a different model ran.

That is deliberate graceful degradation, and correct for the on-site generator: the only way the form holds an out-of-list id is a stale `localStorage` value after an ecosystem switch, and the user visibly sees the picker snap back. It also keeps an app pinned to a since-retired version alive instead of hard-failing. It became a problem when the same graph went API-driven through the App Blocks bridge, where the id **is** deliberate (an app author wrote it) and the correction **is** unobservable.

## 🔴 Scope: RECORD ONLY

**No behaviour change.** Which model runs is byte-identical before and after — a test parses every case twice, once with a collector on the context and once without (the pre-change shape), and requires the same model id, with a `typeof … === 'number'` check so an `undefined === undefined` comparison can't pass vacuously. Whether any case should *reject* is a later phase, gated on what the counter below measures.

## What lands

| | |
|---|---|
| `GenerationCtx.modelSubstitutions` | Per-request collector; records `{ requested, applied, reason, ecosystem, workflow }` at the clamp. Write-once, never-throwing. |
| `classifyModelSubstitutionReason` | Reuses `resolveVersionWorkflowScope` — **no parallel guard** — direction-agnostically, over every workflow the ecosystem offers. |
| `BlockWorkflowSnapshot.modelSubstitutions` | Additive + optional; **omitted entirely** when nothing was substituted. Surfaced on both the submit and the whatIf/estimate reply. |
| `civitai_generation_model_substitutions_total{reason,surface}` | Two bounded labels, both code-owned unions → **9 series total** (3 reasons x 3 surfaces). Emitted fail-soft from `validateInput`; the surface is supplied by the caller at context-build time. |

### Reasons

- **`wrong-workflow`** — the id IS one of the ecosystem's workflow-scoped versions but belongs to a different workflow (an edit-only version sent to `txt2img`).
- **`unrecognized`** — the id is in no scoped list at all: a community checkpoint, or a version retired since the app shipped. The case degradation exists to survive.
- **`gated`** — 🔴 **a third case the issue's two-way split does not cover, found while implementing.** The clamp compares against the **gate-filtered** visible list; `resolveVersionWorkflowScope` probes with an **empty** `gateRules`. So "the config offers this version for this workflow, yet the clamp still replaced it" can only mean a gate rule applicable to *this user* hid it. That is neither an app-authoring bug nor a retired model — it is an entitlement decision, and folding it into `unrecognized` would mislabel it. Reachability is proven from a real parse with a real `GateRule`, not from the classifier alone. Happy to fold it back into `unrecognized` if you'd rather hold the union at two.

## Two design constraints settled by execution, not reasoning

**1. An in-value marker would have been inert.** The model node's `output` is `resourceSchema.optional()`, and `resourceSchema` is a plain `z.object()`, which **strips unknown keys** (zod 4.0.17, as pinned here). Returning `{ id, model, __substitution }` from the clamp compiles, unit-tests green against the transform, and observes nothing. Verified by execution with two controls: `z.looseObject` keeps the marker, and a declared key (`id`) survives. Hence the side channel on `GenerationCtx`.

**2. The classifier is injected, not imported.** Classifying means asking the real graph config which workflow a version belongs to — `resolveVersionWorkflowScope`, which imports `generation-graph.ts`. A static import of that from `common.ts` creates a module cycle that is **fatal in one direction**, and silently fine in the other:

| import order | result |
|---|---|
| `generation-graph.ts` first | loads and parses correctly |
| `common.ts` first | **`TypeError: Cannot read properties of undefined (reading 'entries')` at `DataGraph.merge`**, from `flux-graph.ts` → `ecosystem-graph.ts` running its top-level `createCheckpointGraph(...)` against a half-initialised `common` namespace |

Both orders measured. The second is exactly why the first is dangerous. Injection removes the edge: `model-substitution.ts` imports nothing from the graph, and `buildGenerationContext` (server, where `workflow-capability` already loads safely) hands in the classifier.

## Per-request lifetime of the collector

Attached to the **fresh object literal** `buildGenerationContext` returns — verified to be newly constructed on every call with no memoisation. Its three awaited inputs (`getGenerationStatus`, `getGenerationEcosystemConfig`, `getGateRules`) all read process- or redis-level caches, so a mutable array hung off anything reachable through them would accumulate substitutions **across users** and report one caller's requested model id to another. The collector is attached at the top level of the returned literal only, and a test asserts two collectors share no state.

The client never has one: only the server attaches it, so an in-form parse in the browser records nothing and pays nothing.

## Idempotency

**Measured:** the clamp runs **exactly once** per `generationGraph.safeParse` on the real path today (instrumented with a counter inside the clamp; 1 hit per parse for both the wrong-workflow and unrecognized inputs). A zod `.transform()` is not contractually single-shot, so the dedupe is the *guarantee*, not the observation:

- `record()` is write-once per `(workflow, ecosystem, requested, applied)`.
- `takeUnemitted()` marks what it hands out, so a second drain emits nothing.

Both are covered by tests that re-parse / re-drain and require exactly one record / one increment.

## Reproduction of both reasons (real `generationGraph.safeParse`, Qwen, free-tier ctx)

```
txt2img + 2558804 (edit-only v2511) -> model.id 2552908   success: true
txt2img + 987654321 (unrecognized)  -> model.id 2552908   success: true
```

Matching the issue's evidence table. Recorded by the clamp as:

```
{ requested: 2558804,   applied: 2552908, reason: 'wrong-workflow', ecosystem: 'Qwen', workflow: 'txt2img' }
{ requested: 987654321, applied: 2552908, reason: 'unrecognized',   ecosystem: 'Qwen', workflow: 'txt2img' }
```

The version ids in the test are **re-derived from the live config**, not transcribed, so a config change fails the fixture guard rather than silently asserting about ids that no longer exist.

## Population coverage

The ecosystem sweep is derived from the source config **at runtime** — `getWorkflowCapability` now also reports `modelLocked`, read from the same `model` node meta the form renders. **Measured: 89 (ecosystem, workflow) pairs across 43 ecosystems** are `modelLocked` with a default — wider than the ~23 image ecosystems the issue enumerates, because live config also covers video/audio ecosystems and per-workflow variants. Every one of them must record. A new `modelLocked` ecosystem added later is picked up automatically and fails loudly rather than silently inheriting the unobservable behaviour.

## 🔴 Consumer-side trap — alert with `max_over_time`, NOT `rate()`

This is a rare per-pod counter across ~130 pods. A pod that substitutes once creates its labelled child at 1 and never increments it again, so `rate()`/`increase()` over that child is structurally 0 and an alert keyed on it **silently never fires** — the counter looks healthy precisely because the event is rare. Noted in the metrics module. No dashboard/alert rule is created by this PR.

## 🔴 What one increment means

One **graph validation** that substituted, not one user-visible generation — and **not a fixed multiple of one either**. See "Honest multiplier" in the fix-round section above: the ratio is 2 on a normal block submit, 1 on a budget reject, **+1 per `estimateWorkflow` call (unbounded)**, and +1 per idempotency replay. Read it as a rate and a trend, **per surface**; a headcount of affected generations has to come from the snapshot/log side.

## 🔴 Wire-contract implication

`workflow.schema.ts` states that `BlockWorkflowSnapshot` mirrors the type in `@civitai/app-sdk`'s `blocks/types.ts`. `modelSubstitutions` is **additive and optional**, and is omitted when empty, so no existing consumer is affected and every existing snapshot is byte-identical. But the SDK type does not carry the field yet, so a block cannot read it until that type is widened in the SDK repo. **Nothing in this PR edits another repo** — that widening needs a separate change there.

## Testing

40 new tests (22 graph-path + 18 metrics) plus 2 on `snapshotFromWorkflow`. All drive the real `generationGraph.safeParse` and the real prom-client default registry — the clamp, the classifier, the graph, and the counter are not mocked.

### Mutation results — every guard broken on purpose, killed by its OWN assertion

| # | Mutation | Killed by | Assertion |
|---|---|---|---|
| M1 | delete the `record(...)` call in the clamp | `reason 'wrong-workflow' …` (+6 more) | `expected [] to deeply equal [ { requested: 2133258, …(4) } ]` |
| M2 | clamp records `applied: val.id` (the requested id) | `reason 'wrong-workflow' …` (+3) | `expected [ { requested: 2133258, …(4) } ] to deeply equal [ { requested: 2133258, …(4) } ]` |
| M3 | remove the write-once dedupe | `two parses on the SAME context produce exactly ONE record` | `expected [ …, …(1) ] to have a length of 1 but got 2` |
| M4 | `takeUnemitted()` stops marking | `a SECOND drain … emits nothing`; `takeUnemitted hands each record out once` | `expected 2 to be 1`; `expected [ { requested: 2133258, …(4) } ] to deeply equal []` |
| M5 | remove `record()`'s try/catch | `SWALLOWS a throwing classifier` | `expected [Function] to not throw an error but 'Error: boom' was thrown` |
| M6 | `list()` returns the internal array | `list() returns a copy` | `expected [ …(2) ] to have a length of 1 but got 2` |
| M7 | drop the `wrong-workflow` branch in the classifier | `reason 'wrong-workflow' …`; `reuses resolveVersionWorkflowScope in BOTH directions` | `expected 'unrecognized' to be 'wrong-workflow'` |
| M8 | fold `gated` into `unrecognized` | `reason 'gated' …` | `expected [ { requested: 2110043, …(4) } ] to deeply equal [ … ]` |
| M9 | hardcode `modelLocked: true` in the capability probe | `records a substitution … on EVERY locked pair` | `expected [ 'Flux1/txt2img -> []', …(42) ] to deeply equal []` |
| M10 | widen the counter to `['reason','version_id']` | `DECLARES exactly one label` | `expected [ 'reason', 'version_id' ] to deeply equal [ 'reason' ]` |
| M11 | remove the emitter's try/catch | `the emitter NEVER throws` | `expected [Function] to not throw an error but 'Error: Added label "reason" is not included in initial labelset: [ 'unrelated' ]' was thrown` |
| M12 | remove `emitModelSubstitutions`' try/catch | `RESOLVES when the collector itself throws on drain` | `promise rejected "Error: collector exploded" instead of resolving` |
| M13 | snapshot spreads a truthy empty array | `OMITS modelSubstitutions when nothing was substituted` | `expected true to be false` |
| M14 | snapshot drops the field entirely | `surfaces modelSubstitutions passed alongside the workflow` | `expected undefined to deeply equal [ { requested: 2558804, …(2) } ]` |

**14/14 killed.** The metric-name/label tests drive the real registry (a name or label typo sails through any test that mocks the metrics module and yields an alert that silently never fires), and the poisoned-registry case is paired with a control proving the unguarded call really does throw — so `not.toThrow()` there is testing the guard, not an inert no-op.

## Local verification — FIRST ROUND (superseded by the fix-round numbers above)

> The tsc/eslint/vitest figures below were measured in an environment without a generated Prisma client, which is why the absolute error count is large. The fix-round section above re-measured everything in an environment that does have one; prefer those numbers.

- **tsc** (repo-local `tsc` 5.9.2, `NODE_OPTIONS=--max_old_space_size=8192`), against an `origin/main` baseline captured in the same environment: **3950 → 3965 errors**, delta accounted for exactly — 15 in the new test file (now fixed to 0) and 2 pre-existing `blocks.router.ts` `TS7006` errors that only shifted line number (2160/2161 → 2175/2176). **Zero new production-code type errors.** The large absolute count is the known stale-Prisma-client artifact (`prisma generate` cannot run on this box); it is identical on both sides.
- **eslint** on all changed files: identical result to the `origin/main` baseline — same 4 errors + 5 warnings, all pre-existing, only line-shifted. **Zero new lint issues.**
- **prettier `--list-different`**: clean on every file this PR touches. Three files (`blocks.router.ts`, `workflow.schema.ts`, `workflow.service.test.ts`) were already non-conforming on `main` and were deliberately **not** reformatted; the one hunk of mine that was non-conforming was fixed by hand.
- **vitest `--project unit`** over `src/shared/data-graph src/server/metrics src/server/services/blocks src/server/services/orchestrator src/server/routers src/store`:
  - baseline (`origin/main`, same env): 169 files, 2127 tests, **61 files / 18 tests failing**
  - this branch: 171 files, 2169 tests, **61 files / 18 tests failing**
  - **Identical failing-file set** (set-diff empty). Those failures are environmental — missing `.prisma/client` and the `sharp` native binding — not code.
  - Net **+42 tests, +2 files**, all passing.

### Not verified locally

- Nothing was run against a live server or a real block. The bridge plumbing (`createBlockTextToImageStep` → `snapshotFromWorkflow`) is covered by unit tests on both ends but **not** by an end-to-end block submit.
- The counter has not been observed on a real `/api/metrics` scrape — only against the prom-client default registry in-process.
- `prisma generate` cannot run on this host, so the local `tsc` run is only meaningful as a **delta** against the same-environment baseline. CI is the authority.

### DIFF
--- a/src/server/metrics/__tests__/generation-model-substitution.metrics.test.ts
+++ b/src/server/metrics/__tests__/generation-model-substitution.metrics.test.ts
@@ -2,8 +2,10 @@ import client from 'prom-client';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import {
+  GENERATION_SURFACES,
   MODEL_SUBSTITUTION_REASONS,
   createModelSubstitutionCollector,
+  type GenerationSurface,
   type ModelSubstitutionReason,
 } from '~/shared/data-graph/generation/model-substitution';
 import { emitModelSubstitutions } from '../emit-model-substitutions';
@@ -28,21 +30,37 @@ import {
  *      unguarded prom-client error here would turn a working generation into a
  *      500 — observability causing the outage.
  *
- * 🔴 CARDINALITY is the other load-bearing property: `reason` over a 3-value
- * code-owned union = 3 series, total, forever, across ~130 scraped pods where
- * prom-client retains every distinct label set in the Node heap for the process
- * lifetime. Widening it has to get past these tests.
+ * 🔴 CARDINALITY is the other load-bearing property: `reason` (3 values) x
+ * `surface` (3 values) = 9 series, total, forever, across ~130 scraped pods
+ * where prom-client retains every distinct label set in the Node heap for the
+ * process lifetime. Widening it has to get past these tests.
+ *
+ * 🔴 SURFACE is load-bearing for a different reason: `validateInput` is shared
+ * by the on-site generator, the App Blocks bridge and preset generation, and
+ * #3520 calls the substitution CORRECT on-site and unobservable through the
+ * bridge. Summed into one series the counter measures a contaminated quantity,
+ * which is exactly what phase 2 is supposed to produce a clean number for.
  */
 
 const METRIC = 'civitai_generation_model_substitutions_total';
 
-async function readReason(reason: string): Promise<number> {
+async function readSeries(reason: string, surface: string): Promise<number> {
   const metric = client.register.getSingleMetric(METRIC) as
     | { get(): Promise<{ values: Array<{ labels: Record<string, string>; value: number }> }> }
     | undefined;
   if (!metric) return Number.NaN;
   const { values } = await metric.get();
-  return values.find((v) => v.labels.reason === reason)?.value ?? 0;
+  return values.find((v) => v.labels.reason === reason && v.labels.surface === surface)?.value ?? 0;
+}
+
+/** Sum across surfaces — the CONTAMINATED read the surface label exists to split. */
+async function readReasonAcrossSurfaces(reason: string): Promise<number> {
+  const metric = client.register.getSingleMetric(METRIC) as
+    | { get(): Promise<{ values: Array<{ labels: Record<string, string>; value: number }> }> }
+    | undefined;
+  if (!metric) return Number.NaN;
+  const { values } = await metric.get();
+  return values.filter((v) => v.labels.reason === reason).reduce((sum, v) => sum + v.value, 0);
 }
 
 beforeEach(() => {
@@ -58,30 +76,71 @@ describe('civitai_generation_model_substitutions_total', () => {
     expect(client.register.getSingleMetric(METRIC)).toBeDefined();
   });
 
-  it.each(MODEL_SUBSTITUTION_REASONS.map((r) => [r] as [ModelSubstitutionReason]))(
-    'increments the `%s` series',
-    async (reason) => {
-      recordGenerationModelSubstitution(reason);
-      expect(await readReason(reason)).toBe(1);
+  it.each(
+    MODEL_SUBSTITUTION_REASONS.flatMap((reason) =>
+      GENERATION_SURFACES.map(
+        (surface) => [reason, surface] as [ModelSubstitutionReason, GenerationSurface]
+      )
+    )
+  )('increments the (`%s`, `%s`) series', async (reason, surface) => {
+    recordGenerationModelSubstitution(reason, surface);
+    expect(await readSeries(reason, surface)).toBe(1);
+  });
+
+  it('🔴 the SURFACES are separate series — on-site (correct) never inflates the block number', async () => {
+    // The whole point of the label. #3520 defends the on-site substitution as
+    // intended behaviour and indicts the bridge one; phase 3's decision is gated
+    // on the BLOCK incidence, so an un-split counter would report a number that
+    // is mostly the case nobody wants to change.
+    recordGenerationModelSubstitution('unrecognized', 'onsite');
+    recordGenerationModelSubstitution('unrecognized', 'onsite');
+    recordGenerationModelSubstitution('unrecognized', 'onsite');
+    recordGenerationModelSubstitution('unrecognized', 'block');
+    recordGenerationModelSubstitution('unrecognized', 'preset');
+
+    expect(await readSeries('unrecognized', 'block')).toBe(1);
+    expect(await readSeries('unrecognized', 'onsite')).toBe(3);
+    expect(await readSeries('unrecognized', 'preset')).toBe(1);
+    // …and the contaminated read the split exists to avoid.
+    expect(await readReasonAcrossSurfaces('unrecognized')).toBe(5);
+  });
+
+  it.each([
+    ['undefined', undefined],
+    ['an unknown surface', 'cli'],
+    ['a number', 7],
+  ])(
+    '🔴 DROPS an increment whose surface is %s — the label set stays bounded',
+    async (_l, bogus) => {
+      recordGenerationModelSubstitution('unrecognized', bogus as never as GenerationSurface);
+      expect(client.register.getSingleMetric(METRIC)).toBeUndefined();
     }
   );
+
+  it.each([
+    ['undefined', undefined],
+    ['an unknown reason', 'because'],
+  ])('🔴 DROPS an increment whose reason is %s', async (_l, bogus) => {
+    recordGenerationModelSubstitution(bogus as never as ModelSubstitutionReason, 'block');
+    expect(client.register.getSingleMetric(METRIC)).toBeUndefined();
+  });
 
   it('🔴 the reasons are SEPARATE series — an entitlement swap never reads as a wrong-mode swap', async () => {
     // The three reasons demand different responses: `wrong-workflow` is an
     // app-authoring bug, `unrecognized` is a retired/unknown model (the case
     // degradation exists to survive), `gated` is an entitlement decision. One
     // undifferentiated counter would send an operator to the wrong place.
-    recordGenerationModelSubstitution('wrong-workflow');
-    recordGenerationModelSubstitution('unrecognized');
-    recordGenerationModelSubstitution('unrecognized');
-    recordGenerationModelSubstitution('gated');
+    recordGenerationModelSubstitution('wrong-workflow', 'block');
+    recordGenerationModelSubstitution('unrecognized', 'block');
+    recordGenerationModelSubstitution('unrecognized', 'block');
+    recordGenerationModelSubstitution('gated', 'block');
 
-    expect(await readReason('wrong-workflow')).toBe(1);
-    expect(await readReason('unrecognized')).toBe(2);
-    expect(await readReason('gated')).toBe(1);
+    expect(await readSeries('wrong-workflow', 'block')).toBe(1);
+    expect(await readSeries('unrecognized', 'block')).toBe(2);
+    expect(await readSeries('gated', 'block')).toBe(1);
   });
 
-  it('🔴 DECLARES exactly one label, `reason` — the cardinality bound is structural', async () => {
+  it('🔴 DECLARES exactly `reason` + `surface` — the cardinality bound is structural', async () => {
     // 🔴 Asserted against the DECLARED `labelNames`, NOT the emitted series.
     // prom-client omits a declared-but-never-supplied label from its output, so
     // inspecting `values[].labels` stays green while the metric is declared wide
@@ -89,34 +148,40 @@ describe('civitai_generation_model_substitutions_total', () => {
     // nothing ever having failed.
     ensureRegisterGenerationModelSubstitutionMetrics();
     const metric = client.register.getSingleMetric(METRIC) as unknown as { labelNames: string[] };
-    expect([...metric.labelNames].sort()).toEqual(['reason']);
+    expect([...metric.labelNames].sort()).toEqual(['reason', 'surface']);
 
-    recordGenerationModelSubstitution('unrecognized');
+    recordGenerationModelSubstitution('unrecognized', 'block');
     const emitted = client.register.getSingleMetric(METRIC) as unknown as {
       get(): Promise<{ values: Array<{ labels: Record<string, string> }> }>;
     };
     const { values } = await emitted.get();
     expect(values.length).toBeGreaterThan(0);
-    for (const v of values) expect(Object.keys(v.labels)).toEqual(['reason']);
+    for (const v of values) expect(Object.keys(v.labels).sort()).toEqual(['reason', 'surface']);
   });
 
-  it('🔴 the reason union is EXACTLY these three values — 3 series is the whole budget', () => {
-    // Literal, not derived: the union is simultaneously the metric label AND the
-    // `BlockWorkflowSnapshot.modelSubstitutions[].reason` wire contract, so a
-    // fourth value added on either side has to come through here.
+  it('🔴 both unions are EXACTLY these values — 3 x 3 = 9 series is the whole budget', () => {
+    // Literal, not derived: the reason union is simultaneously the metric label
+    // AND the `BlockWorkflowSnapshot.modelSubstitutions[].reason` wire contract,
+    // so a fourth value added on either side has to come through here. The
+    // surface union bounds the other axis.
     expect([...MODEL_SUBSTITUTION_REASONS]).toEqual(['wrong-workflow', 'unrecognized', 'gated']);
+    expect([...GENERATION_SURFACES]).toEqual(['block', 'onsite', 'preset']);
   });
 
-  it('🔴 emits AT MOST 3 series no matter how many substitutions land', async () => {
+  it('🔴 emits AT MOST 9 series no matter how many substitutions land', async () => {
     for (let i = 0; i < 100; i++) {
-      for (const reason of MODEL_SUBSTITUTION_REASONS) recordGenerationModelSubstitution(reason);
+      for (const reason of MODEL_SUBSTITUTION_REASONS) {
+        for (const surface of GENERATION_SURFACES) {
+          recordGenerationModelSubstitution(reason, surface);
+        }
+      }
     }
     const metric = client.register.getSingleMetric(METRIC) as unknown as {
       get(): Promise<{ values: Array<{ labels: Record<string, string> }> }>;
     };
     const { values } = await metric.get();
-    expect(values).toHaveLength(3);
-    expect(await readReason('wrong-workflow')).toBe(100);
+    expect(values).toHaveLength(9);
+    expect(await readSeries('wrong-workflow', 'block')).toBe(100);
   });
 
   it('is idempotent to register — a double module import does not throw', () => {
@@ -126,11 +191,11 @@ describe('civitai_generation_model_substitutions_total', () => {
     }).not.toThrow();
   });
 
-  it('appears in the scrape output under its exact name', async () => {
-    recordGenerationModelSubstitution('gated');
+  it('appears in the scrape output under its exact name, with both labels', async () => {
+    recordGenerationModelSubstitution('gated', 'block');
     const scrape = await client.register.metrics();
     expect(scrape).toContain(METRIC);
-    expect(scrape).toContain(`${METRIC}{reason="gated"}`);
+    expect(scrape).toContain(`${METRIC}{reason="gated",surface="block"}`);
   });
 
   it('🔴 the emitter NEVER throws — a broken registry cannot fail a generation', () => {
@@ -145,7 +210,7 @@ describe('civitai_generation_model_substitutions_total', () => {
       registers: [client.register],
     });
 
-    expect(() => recordGenerationModelSubstitution('unrecognized')).not.toThrow();
+    expect(() => recordGenerationModelSubstitution('unrecognized', 'block')).not.toThrow();
   });
 
   it('the poisoned-registry case really would throw unguarded (the guard is reachable)', () => {
@@ -158,14 +223,14 @@ describe('civitai_generation_model_substitutions_total', () => {
       labelNames: ['unrelated'],
       registers: [client.register],
     });
-    expect(() => poisoned.inc({ reason: 'unrecognized' })).toThrow();
+    expect(() => poisoned.inc({ reason: 'unrecognized', surface: 'block' })).toThrow();
   });
 });
 
 describe('emitModelSubstitutions — the drain the generation path calls', () => {
-  function collectorWith(...reasons: ModelSubstitutionReason[]) {
+  function collectorWith(surface: GenerationSurface, ...reasons: ModelSubstitutionReason[]) {
     let i = 0;
-    const collector = createModelSubstitutionCollector(() => reasons[i++]);
+    const collector = createModelSubstitutionCollector(() => reasons[i++], surface);
     reasons.forEach((_, n) =>
       collector.record({ requested: 1000 + n, applied: 1, ecosystem: 'Qwen', workflow: 'txt2img' })
     );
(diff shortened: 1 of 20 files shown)


ZacxDev:
### `preview / smoke-tests` red — shared preview-database outage, not this change

Recording the evidence so the red check isn't read as a regression.

**The three code gates are green on this commit:** ESLint + Prettier, Typecheck, Unit tests, plus the event-engine-common pin.

**The 24 smoke failures are all one cause.** Every error in the preview app's logs is the same database connection failure (`server login has been failing … connect failed`), across unrelated Prisma models — `rewardsBonusEvent`, `changelog`, `bug`, `user`, raw queries. **Zero** log lines mention anything this PR touches (`modelSubstitutions`, `buildGenerationContext`, the surface label, the narrowing helpers). The failing set is also far broader than this diff could reach: articles, collections, downloads, engagement, feeds, membership, model upload, moderation, posts, reports, reviews.

**Population control — it is not this PR.** `/api/v1/models` (the endpoint the smoke suite trips on) currently returns HTTP 500 on **four of the five** live PR previews, including three unrelated PRs. The one preview that returns 200 is the only one pointed at a different database. The preview database itself is mid-restore and replaying steadily, so this should clear on its own; a re-run once it does is the right next step.

**One caveat, stated plainly:** the previous commit on this branch already had `preview / smoke-tests` red with **1** failure — `preview-generation.spec.ts › whatIfFromGraph fires on /generate and returns a numeric cost`. That failure **predates this fix round** (it was red on the pre-fix commit). While the preview database is down, this run cannot say anything about whether it is now fixed, still broken, or broken for a new reason. It should be re-checked against a healthy preview before merge, since `whatIfFromGraph` is on the on-site path this PR labels `onsite`.

---

### All preview checks have now settled — here is the full accounting

| check | result | reading |
|---|---|---|
| ESLint + Prettier / Typecheck / Unit tests / event-engine-common pin | **pass** | the blocking gates |
| `preview / deploy`, `render-check`, `auth-tests` | **pass** | environment is up and rendering |
| `preview / unit-tests` | fail *(report-only)* | **1 failed of 737 files** — `og-metadata.test.ts › adversarial-input cost (ReDoS guard)`, a **wall-time budget** assertion (`expected 6397 to be less than 4000`) in a file this PR does not touch. The **same file failed on the pre-fix commit of this branch, with 2 failing tests instead of 1** — so it is pre-existing and runner-load dependent, and it got *better*, not worse. The GitHub Actions "Unit tests" job passed on this exact commit. |
| `preview / component-tests` | fail *(report-only)* | "exceeded its runner budget — no verdict, NOT a test failure" — byte-identical description to the pre-fix commit. |
| `preview / smoke-tests` | fail | the shared preview-database outage described above. |

**The new tests did run on CI and passed**, with the counts they have locally — worth stating explicitly, since a file that fails to *collect* looks identical to one that passes:

```
✓ model-substitution.test.ts                     (34 tests)
✓ generation-model-substitution.metrics.test.ts  (31 tests)
✓ generation-surface-wiring.test.ts               (6 tests)
✓ model-substitution.no-behaviour-change.test.ts  (5 tests)
✓ preset-image-gen.surface.test.ts                (2 tests)
```

File count went 734 → 737 on the preview runner, exactly the three files added here.

**One cost worth naming:** the full-population no-behaviour-change sweep takes **34s** on the CI runner (3.6s locally) — the 5th slowest file in the suite, though well behind three pre-existing files at 82s / 65s / 57s. The 60s budget is **per test**, not per file, and its slowest single test is roughly 14s, so it is not near the limit. Happy to trim the sweep (it does 684 full-object comparisons) if the wall time is unwelcome.

Issue #3527, opened 2026-08-01
feat: Add chatCompletion kind to App Blocks workflow bridge


ZacxDev:
## Summary

The block workflow bridge currently only supports two workflow kinds: `textToImage` and `customComfy`. There is no way for an App Block to trigger a chat completion through the orchestrator. This blocks apps that need LLM chat capabilities (e.g., an AI research assistant).

## Current State

The `blockWorkflowBodySchema` discriminated union in `src/server/schema/blocks/workflow.schema.ts` only accepts:

- `kind: 'textToImage'` — diffusion generation
- `kind: 'customComfy'` — server-side ComfyUI recipe

The `chatCompletion` step type already exists in the orchestrator (used server-side by the Comics feature at `src/server/services/comics/orchestrator-chat.ts`), but is never exposed through the block bridge.

## Proposed Change

Extend the block workflow bridge to support a `chatCompletion` kind:

### 1. Schema (`workflow.schema.ts`)

Add a new discriminated union member:

```typescript
const blockChatCompletionBodySchema = z.object({
  kind: z.literal('chatCompletion'),
  model: z.string().describe('OpenRouter or AIR model identifier'),
  messages: z.array(z.object({
    role: z.enum(['system', 'user', 'assistant', 'tool']),
    content: z.string(),
    tool_call_id: z.string().optional(),
  })),
  max_tokens: z.number().max(128000).optional(),
  temperature: z.number().min(0).max(2).optional(),
  stream: z.boolean().optional(),
  tools: z.array(z.object({
    type: z.literal('function'),
    function: z.object({
      name: z.string(),
      description: z.string(),
      parameters: z.record(z.unknown()),
    }),
  })).optional(),
  tool_choice: z.enum(['auto', 'none', 'required']).optional(),
  response_format: z.object({
    type: z.enum(['text', 'json_object']),
  }).optional(),
});
```

Extend the union:

```typescript
export const blockWorkflowBodySchema = z.discriminatedUnion('kind', [
  blockTextToImageBodySchema,
  blockCustomComfyBodySchema,
  blockChatCompletionBodySchema,  // NEW
]);
```

### 2. Handler (`blocks.router.ts`)

Add `blocks.submitChatCompletionWorkflow` procedure:

- Verify block JWT + `ai:write:budgeted` scope
- Resolve model context
- Build `{ steps: [{ $type: 'chatCompletion', input: { model, messages, ... } }] }` workflow body
[...]

Issue #3515, opened 2026-07-31
RFC: step-type registry for the App Blocks bridge — make capabilities additive before the wire contract is public


ZacxDev:
## Problem

The App Blocks generation bridge exposes **2** of the orchestrator's ~20 developer-facing step types. If the goal is to expose the rest, the current shape doesn't get us there — and the part that's expensive to undo is the **public wire contract**, not the code.

Today each capability is a hand-written branch. `estimateWorkflow` branches on `kind === 'customComfy'`, `submitWorkflow` branches again, then settle and budget are handled per-kind. That's a clean pattern at N=2 — the code says so explicitly ("we only ADD a branch… everything customComfy-specific lives in the helper"). At N=20 it's twenty parallel branches across two hot procedures, each with its own settle and budget path, and no single place to enforce a cross-cutting rule.

The blocking concern is compatibility, not tidiness. **`@civitai/app-sdk` mirrors this wire shape.** Once apps ship against `kind: 'backgroundRemoval'`, `kind: 'upscale'`, `kind: 'tts'` as sibling union members, consolidating them is a breaking change across every published app. We should pick the shape before broadening access, not after.

## We already have the right abstraction one level down

`BlockRecipe<P>` (`src/server/services/blocks/recipes/index.ts`) is close to what a step-type registry needs, and it's already proven on the security-critical post-paid path:

```ts
interface BlockRecipe<P> {
  id: string;                                     // registry key → wire enum
  paramSchema: z.ZodType<P>;                      // bounded, .strict()
  buildStep(params, resolved): CustomComfyStepInput;   // pure builder
  resourceAllowlist: { checkpoints?, loras?, staticAirs };
  budgetFor(params): { maxBuzz; stepTimeoutSeconds };  // the ceiling
  estimateBuzz(params): number;                   // display estimate
  resolveEngine(params): string;                  // variant selection
  // …checkpointPolicy, negativePrompt, engines
}
```

Registry keys derive the wire enum, so an unregistered id fails closed at the schema. The registry is a code-reviewed, non-DB-editable trust root. That's the right security posture and it should not change.

**Proposal: generalize this from "Comfy recipes" to "step types."** It's an evolution of a pattern that already works, not a rewrite.

## Proposed wire shape

Instead of one union member per capability:

```ts
// today (doesn't scale)
{ kind: 'textToImage', … } | { kind: 'customComfy', recipe, params }
// …and then 18 more sibling members

// proposed
[...]

PR #3538, opened 2026-08-02
feat(app-blocks): step-type registry (kind:'step') + verified Tranche 1 — RFC #3515 steps 1–2

ZacxDev:
Implements [#3515](https://github.com/civitai/civitai/issues/3515) migration **steps 1 and 2**: land the step-type registry + `kind: 'step'` alongside the existing union members, and register a verified Tranche 1.

Unblocks the shape question ahead of [#3527](https://github.com/civitai/civitai/issues/3527) — though `chatCompletion` itself is explicitly **not** enabled by this PR (see Moderation posture below).

---

## 🔴 Task 0 first: what is actually exposable

The RFC named Tranche 1 as *"background removal, upscale, convert"* and flagged it as **inferred, not verified**. I enumerated every `$type` in the orchestrator's generated types (`@civitai/client@0.2.0-beta.83`, `dist/generated/types.gen.d.ts` — **43 distinct `$type` values**, alongside **47** standalone `/v2/consumer/recipes/<name>` endpoints; the endpoint list is a superset, since a few step *templates* have an endpoint without appearing as a `$type` in a Step union) and checked each against four bars:

1. a genuine standalone orchestrator `$type` (not an internal helper, not a Comfy workflow feature)
2. **deterministic cost knowable before execution** (`prepaidFixed`)
3. **no new moderation surface** — nothing with a free-text output
4. no checkpoint/LoRA entitlement entanglement

**Correction to the premise in both directions.** `imageBackgroundRemoval` and `imageUpscaler` **do** exist as first-class `$type` values with standalone `/v2/consumer/recipes/<name>` endpoints — they are not missing. But neither qualifies for Tranche 1:

| Step type | Verdict | Why |
|---|---|---|
| **`convertImage`** | ✅ **REGISTERED** | Standalone `$type` + consumer-recipe endpoint. Pure CPU format conversion + resize: no GPU model, no sampler, no step count, no wall-clock dependence. Output is an image blob, input is a bounded image URL → no free-text either side. References no AIR resource at all. Already submitted server-side as a single-step workflow by `product-badge.service.ts`, so the shape is proven in production. |
| `imageBackgroundRemoval` | ❌ fails bar 2 | The client's **own doc string**: *"removes the background from an image using BiRefNet (**builds a ComfyUI graph under the hood and runs it as a comfy job**)"*. That is GPU work with variable duration — `timeBounded`, not a deterministic price. Strong Tranche 2 candidate. |
| `imageUpscaler` | ❌ fails bars 2 and 4 | GPU work, and `numberOfRepeats` (1–3, each doubling resolution) makes cost input-dependent. Separately, `model` takes an **arbitrary AIR URN** — an AIR chosen by an untrusted iframe reaches the orchestrator without the generation-graph entitlement belt, the same bypass class the `customComfy` `resources` array carries. Needs a resource gate before exposure. |
| `imageToSvg` | ❌ fails bar 2 | *"using a local StarVector or OmniSVG model"* — model inference. |
| `mediaHash` | ⚠️ **passes all four, deliberately not registered** | Standalone `$type` + consumer endpoint, deterministic CPU hashing, bounded hash-string output, no resources. But it has no demonstrated developer demand, and every registered id is a permanent public wire commitment. Adding it later is one file + one registry line — which is the entire point of the registry. Not padding the tranche to hit a number. |
| `transcode` | ❌ | `destinationUrl` is a caller-supplied write target (SSRF/write primitive). |
| `chatCompletion`, `mediaCaptioning`, `audioCaptioning`, `transcription`, `promptEnhancement`, `textToSpeech` | ❌ bar 3 | Free-text output (or free-text input, for TTS) — a new moderation surface. Out regardless of convenience. |
| `mediaRating`, `ageClassification`, `wdTagging`, `xGuardModeration` | ❌ policy | Moderation classifiers. Exposing them lets an app probe the moderation stack — a policy call, not an engineering one. |
| `preprocessImage` | ❌ bar 1 | Internal helper `controlnets.helper.ts` emits in front of a generation. |
| `textToImage`, `imageGen`, `comfy`, `customComfy`, `videoGen`, `polyGen`, `aceStepAudio`, `videoUpscaler`, `videoEnhancement`, `videoInterpolation`, `videoBackgroundRemoval` | ❌ bar 2/4 | GPU generation: timeBounded + entitlement. |
| `training`, `imageResourceTraining`, `model*Scan`, `modelHash`, `modelParseMetadata`, `comfyNodepackSnapshot`, `qwenImageBench`, `blobArchive`, `imageUpload`, `composeMedia`, `echo`, `videoFrameExtraction`, `videoMetadata`, `repeat`, `humanoidImageMask`, `tryOnU`, `batchOCRSafetyClassification`, `model3DPreview` | ❌ bar 1 | Internal/infra plumbing. |

**Tranche 1 = `convert-image` only.**

---

## The registry

`src/server/services/blocks/steps/` — a generalization of `BlockRecipe`, keeping its security posture verbatim: a **code-reviewed, non-DB-editable trust root** (no DB access, loads at import time before any request context), whose **keys derive the wire enum**, so an unregistered id fails closed at the schema before any translator, spend reservation, or orchestrator call.

### The three required fields

**1. `billingMode: 'timeBounded' | 'prepaidFixed' | 'tokenMetered'` — required.**
All three values are declared now even though Tranche 1 only uses one. That is the point: the wire contract and type surface are the expensive things to change once the SDK mirrors them, so future step types must be *additive*. A mode with no implemented money-path handler **fails at registry load**, so it can never silently ride another mode's reservation logic.

**2. `prepaidFixed` gets its OWN enforced load-time invariant — not an exemption.**
The existing `maxBuzz === ceil(stepTimeoutSeconds)` invariant works because a step timeout is a *physical* cap at ~1 Buzz/GPU-second. A fixed-price step has no such time relationship and cannot satisfy it. If `prepaidFixed` entries simply *skipped* the check, the registry's structural guard would still exist but would no longer be reachable for the new class — the worst outcome, because the registry would look guarded while new entries were not. So `prepaidFixed` declares an equivalent invariant, validated at load with the same fail-fast shape:

```
for every variant v:
  priceForVariant(v) is a positive safe integer
  estimateBuzz(canonicalParamsFor(v)) === priceForVariant(v)
```

The second clause is the money-relevant one: the price is what gets **reserved against every cap**, `estimateBuzz` is what the app is **shown**. A divergence means the block quotes one number and the viewer is charged another. Evaluating it needs real params, which is why entries declare `canonicalParamsFor(variant)` — the same reason `budgetForEngine` exists on `BlockRecipe` (the load loop enumerates without a request context).

**3. `moderationPosture: 'none' | 'promptAudit' | 'textOutput'` — required.**
Only `'none'` has an implemented handler; the other two fail at registry **load**. Registering `chatCompletion` (or captioning, or transcription) is therefore impossible without someone explicitly answering the policy question in code. **That policy decision is not made here and is out of scope** — this PR only makes it unskippable. Enforced twice: at load, and again as a compile error (below).

Plus: **every `paramSchema` must be `.strict()`**, probed behaviourally at load (`{...canonicalParams, <probe key>}` must be rejected). `blockTextToImageBodySchema` is *not* strict, and that non-strictness is exactly what let an older host silently strip `sourceImages` and bill the wrong generation. A dropped param on a money path is a wrong-generation bug, not a validation nit.

### Guard ordering is load-bearing

The per-variant budget invariants run **before** the implemented-billing-mode gate, deliberately. With the gate first, the `timeBounded` / `tokenMetered` budget branches were **structurally unable to execute** — the gate rejects every such entry, so no mutation could ever kill them. The first mutation sweep caught exactly this; reordering made them genuinely reachable and mutation-proven (M20–M22 below).

---

## 🔴 The money path

A step submit runs the **same cap belt** as every other spending kind, reserved **before** the orchestrator submit and refunded on every non-committed exit:

1. static per-call `buzzBudget` gate (exact for `prepaidFixed` — the price is known before execution)
2. gen-idempotency claim
3. per-user cumulative daily cap
4. 🔴 **per-app aggregate `reserveAppSpend`** — daily Buzz + velocity
5. dev-session backstop

If `kind: 'step'` bypassed (4), this PR would open a new money surface the guardrail cannot see — and `civitai_app_block_spend_cap_rejections_total` would not fire for it either, so the gap would also be invisible in monitoring. Mutation-proven: bypassing the reserve kills 7 tests (M1).

`prepaidFixed` never touches the post-paid settle machinery — the price is exact, so the reservation is final, no settle record is persisted, and the terminal poll/cancel hook has nothing to read.

### The price is now MEASURED, and the ceiling no longer depends on it

**Two live calls against the PR preview's orchestrator (2026-08-02):**

| call | result |
|---|---|
| `POST /v2/consumer/workflows?whatif=true`, one `convertImage` step | HTTP 200, `cost: {base:1, factors:{base:1}, fixed:{}, tips:{...}, **total: 1**}` |
| real (non-whatif) submit of the same step, polled to `succeeded` | HTTP 202, `status: processing`, `cost.total: 1`; per-job `cost: 1` |

**The orchestrator's real price for `convertImage` is 1 Buzz.** `CONVERT_IMAGE_PRICE_BUZZ = 1` was correct — it is no longer a guess.

🔴 **The earlier residual statement in this PR was wrong, in the direction that costs money, and is retracted.** It said the *first* divergent submit could exceed the per-call budget. In fact the divergence correction only ever adjusted the **cap counters** — it never touched the declared price or the static gate — so **every** submit would have exceeded `buzzBudget` by the overage, forever, until a human read the metric and shipped code.

That is fixed structurally rather than by trusting the measurement. This was the **first** block spend path whose per-call ceiling rested solely on a number Civitai asserted: `textToImage` submit issues a real `submitWorkflow(…, query:{whatif:true})` and gates on the orchestrator's own quote, and `customComfy` stamps a step `timeout` the orchestrator physically enforces as the Buzz ceiling. The step path did neither — `stepTimeoutSeconds` is deliberately `null` for `prepaidFixed`.

**Submit now runs a real `whatif:true` quote BEFORE the per-call budget gate** and reserves `max(declared, quoted)`:

- `prepaidFixed` means *"cost knowable before execution"* — which is exactly what `whatif:true` returns. Asking the orchestrator is not a workaround for the mode; it is the mode's own premise made enforceable.
- The quote is free, side-effect-free and carries **no `externalId`** (the idempotency key belongs to the real submit only — mirrors the txt2img whatIf preflight).
- **Fail-closed on an unquotable step**: a missing quote means the premise did not hold, which puts the ceiling back on the unenforced constant. Returns a recoverable failed-shape snapshot and counts `outcome: 'absent'`.
- Ordering is `quote → budget gate → per-user cap → per-app cap → dev session → real submit` — **byte-for-byte the ordering `textToImage` has used since before this PR**. The accepted cost: a cap-rejected or replayed submit now makes one extra free orchestrator call, exactly as txt2img already does.

The declared price stays as the **floor** (it is what `estimateBuzz` showed the block, and the load-time invariant forces those to agree); the quote raises it when the rate card moves.

## Dispatch

estimate/submit/settle dispatch on **`billingMode`**, via `estimateStepBuzz(entry, params)` / `planStepSpend(entry, params)`. Verified structurally: the two handlers (~440 lines) contain **zero** code references to `billingMode`, a mode literal, or a step id — every occurrence is in a comment. Adding a capability = one file + one registry line. Adding a *mode* = one handler in the registry module. Neither is router surgery.

---

## Explicitly out of scope

- **`customComfy` is NOT migrated behind the registry.** RFC migration step 3.
- **`textToImage` stays its own `kind`** — entitlement over checkpoint *and* LoRA stack, availability gates, model-token binding, page-only `sourceImage`, `accountType` ordering, `sharedContentKey`, POI.
- **No SDK changes.** `@civitai/app-sdk` mirroring is a separate PR after the host deploys (the #205 lesson: an SDK shipped before the host silently strips the field).
- **Scopes stay coarse** — `ai:write:budgeted` gates all generation.
- ~~`snapshotFromWorkflow` is **not** modified~~ — **superseded.** It (and `projectAppWorkflow`) now carry an ADDITIVE registry branch; see "Output extraction" below. ⚠️ **PR #3535 also touches `snapshotFromWorkflow`** — the overlap is one inserted branch placed *before* the native `$type` filter and *after* the `customComfy` branch; no existing line is modified. Re-read the merged region rather than trusting a clean textual merge.

### Rebase notes for in-flight PRs

- **#3535** adds a required 4th `surface` param to `buildGenerationContext`. **This PR's step path never calls that function**, so no `'block'` argument is needed on rebase. It does *call* `snapshotFromWorkflow` (reading only the existing `workflowId` / `status` / `cost.total` fields), so if #3535 changes that return shape, the divergence check and audit writes are the places to re-read.

---

## Verification

**tsc** — `tsc --noEmit` at HEAD and at `origin/main` in the same environment produce a **byte-identical** 6-error set (all pre-existing: ungenerated Prisma client + a `@civitai/shared` build artifact — `prisma generate` cannot run on this NixOS box). **Zero new errors.**

**Tests** — same globs as baseline (`src/server/services/blocks`, `src/server/schema/blocks`, `src/server/routers/__tests__`):

| | Files | Tests |
|---|---|---|
| `origin/main` | 147 | 3141 |
| HEAD | 149 | 3200 |
| **delta** | **+2** | **+59** |

+59 = 24 registry + 9 wire-schema + 26 router money-path. Counted from `--reporter=verbose`, not from an exit code.

The pre-existing 51-test suite covering the moved SSRF/URL predicate (backslash-in-authority, TAB in host, CRLF in host, CR/LF elsewhere, NUL and other C0 bytes, backslash in path, canonicalization round-trip) passes untouched — evidence the extraction is behaviour-preserving.

**eslint / prettier** — added files: clean on both. Modified files: eslint error count identical to baseline (16 → 16, all pre-existing `no-explicit-any`); prettier's unformatted-file set is identical to baseline (3 files, pre-existing), and every hunk **inside the regions this PR adds** is prettier-conformant.

### Mutation table

Every guard was broken on purpose, its test file re-run, and the file restored from a **byte-exact backup with a sha256 equality check** (never a `str.replace` revert). All 22 mutations were killed; each by a test whose assertion names the mutated guard.

| # | Mutation | Test that failed | Killing assertion |
|---|---|---|---|
| M1 | bypass `reserveAppSpend` on the step submit path | *reserves the EXACT declared price on the per-user daily cap AND the PER-APP cap* (+6 more) | `expected "vi.fn()" to be called with arguments: [ 'apb_test', 1 ]` |
| M2 | drop the per-app refund on the submit-throw path | *refunds the PER-APP reservation when the orchestrator submit throws* | `expected "vi.fn()" to be called with arguments: [ …(2) ]` |
| M3 | drop the static per-call budget gate | *an over-budget price fails the STATIC gate before any reservation or submit* | `expected 'wf_step_1' to be 'failed'` |
| M4 | drop the price-divergence correction | *CORRECTS both cap counters when the orchestrator bills ABOVE the declared price* | `expected "vi.fn()" to be called with arguments: [ …(2) ]` |
| M5 | flip `prepaidFixed` to `postPaidSettle: true` | *does NOT touch the post-paid settle machinery* | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |
| M6 | drop the page-only guard | *REJECTS a model token* + *a MODEL token is rejected fail-closed BEFORE any spend* | `promise resolved "{ snapshot: { …(3) } }" instead of rejecting` |
| M7 | drop the registry-key === `step.id` guard | *rejects a registry key that disagrees with step.id* | `expected [Function] to throw an error` (regex `/registry key must equal step\.id/`) |
| M8 | drop the implemented-billing-mode guard | *rejects a billing mode with no implemented money-path handler* (+2) | `expected [Function] to throw an error` |
| M9 | drop the implemented-moderation-posture guard | *rejects a moderation posture with no implemented handler* | `expected [Function] to throw an error` |
| M10 | drop the non-empty-variants guard | *rejects an entry declaring no variants* | `expected [Function] to throw an error` |
| M11 | drop the canonical-params-parse guard | *rejects canonical params that do not satisfy the param schema* | `expected [Function] to throw an error` |
| M12 | drop the plain-object guard | *rejects canonical params that are not a plain object* | `expected [Function] to throw an error` |
| M13 | drop the `.strict()` probe guard | *rejects a paramSchema that is NOT .strict()* | `expected [Function] to throw an error` |
| M14 | drop the variant-resolution guard | *rejects canonical params that resolve to a DIFFERENT variant* | `expected [Function] to throw an error` |
| M15 | drop the `prepaidFixed` positive-integer-price guard | *rejects a prepaidFixed price that is not a positive integer* | `expected [Function] to throw error matching /prepaidFixed price \(0\) must be a po…/ but got 'block step 'fixture-step' variant …'` |
| M16 | drop the `prepaidFixed` estimate===price guard | *rejects a prepaidFixed estimate that disagrees with the declared price* | `expected [Function] to throw an error` |
| M17 | make `timeBounded` silently ride `prepaidFixed`'s logic | *routes a fixture entry with a DIFFERENT billing mode* + *the dispatcher is TOTAL* | `expected 'undefined' to contain 'billing mode 'timeBounded' has no m…'` |
| M18 | hardcode the step-id enum instead of deriving it | *derives its id enum from the registry keys* | `expected false to be true` |
| M19 | drop `.strict()` from the step wire body | *REJECTS an unknown top-level field* | `expected true to be false` |
| M20 | drop the `timeBounded` maxBuzz===ceil(timeout) guard | *enforces the timeBounded maxBuzz === ceil(stepTimeoutSeconds) invariant* | `expected [Function] to throw error matching /timeBounded maxBuzz \(90\) must equal…/ but got 'block step 'fixture-step': billingM…'` |
| M21 | drop the `timeBounded` positive-timeout guard | *(same test)* | `expected [Function] to throw error matching /timeBounded stepTimeoutSeconds \(0\) …/ but got 'block step 'fixture-step': billingM…'` |
| M22 | drop the `tokenMetered` positive-maxBuzz guard | *enforces the tokenMetered positive-maxBuzz invariant* | `expected [Function] to throw error matching /tokenMetered maxBuzz \(0\) must be a …/ but got 'block step 'fixture-step': billingM…'` |

M15 and M20–M22 are the interesting rows: with the guard removed the fixture falls through to a *neighbouring* guard and still throws — and the test fails anyway, because it demands **its own guard's message**. That is the difference between "a test failed" and "this guard is tested".

**Compile-time contract, also mutation-proven.** 🔴 The first draft put the type-level assertions in `__tests__/` — where they were **never typechecked**, because `tsconfig.json` excludes `src/**/__tests__/**` and `pnpm run typecheck` is `tsc --noEmit` against that config. A clean `tsc` run "proving" they worked was the tell that they could not fail. They now live in `src/server/services/blocks/steps/type-contract.ts` (inside the `src` include, type-only so it compiles to nothing), and were verified by mutation:

| Mutation | Result |
|---|---|
| `billingMode?:` optional on the `PrepaidFixedStep` member | ✅ `type-contract.ts(57,3): error TS2344` (`_RequiresBillingMode`) |
| `moderationPosture?:` optional | ✅ `type-contract.ts(66,3): error TS2344` (`_RequiresModerationPosture`) |
| widen `billingMode` to `string` | ✅ `type-contract.ts(71,3): error TS2344` (`_RejectsUnknownBillingMode`) |
| `priceForVariant?:` optional | ✅ `type-contract.ts(83,3): error TS2344` (`_PrepaidFixedNeedsPrice`) |

*(One earlier mutation — making `billingMode` optional on the shared **base** interface — did **not** kill, correctly: each union member redeclares the field, so the base modifier is overridden. The mutation was wrong, not the guard; re-run against the member, it kills.)*

### What I could NOT verify

- **End-to-end through the block bridge itself.** The two orchestrator calls above were made directly against the preview's orchestrator with the preview's own token (`ORCHESTRATOR_MODE=dev`, so a block submit uses that same token) — not driven through a real iframe holding a real block JWT. The step's own submit path is covered by mocked tests only.
- **The real wire contract as an app author sees it** — blind validation by the #3527 author against the preview host is still worth more than any of the above.
- **A `timeBounded` / `tokenMetered` entry end-to-end.** Both modes fail at registry load by design, so their handlers have never executed against a real orchestrator.
- Browser/component tests were not run (node project only; unchanged by this PR).

---

# Fix round — adversarial audit response

Two blockers and six supporting fixes. Everything the audit confirmed by execution is preserved: the derived enum, the load-time fail-closed invariants, `type-contract.ts` being genuinely typechecked and able to fail, complete refund coverage on every exit, and byte-identical behaviour for `textToImage` / `customComfy` (re-proved below).

## 🔴 FIX 1 — the only registered capability's output was unreachable

`snapshotFromWorkflow` (~:59) and `projectAppWorkflow` (~:177) both `continue`d on any `$type` outside `customComfy | textToImage | imageGen | comfy`. `convertImage`'s output is **`{ blob?: ImageBlob }` — SINGULAR `blob`**, not `images` and not `blobs`. Both extractors were blind in **two independent ways**: the `$type` *and* the key. An app would submit, be charged, poll to `succeeded`, and find `snapshot.imageUrls` absent on every poll while `queryAppWorkflows` returned `images: []`. It paid for a result it could never retrieve.

**Fixed registry-driven, not with a fourth hardcoded branch.** A one-off `if ($type === 'convertImage')` would fix today and guarantee the next entry repeats it — this was the third instance of the class.

Two new **required** fields on `BlockStepBase`:

| field | what it does |
|---|---|
| `extractOutput(step): StepOutputMedia[]` | the entry declares its own extraction. One call serves BOTH read surfaces: the snapshot reads `url`, the projection reads `url`/`width`/`height`/`nsfwLevel`. Registering a step and surfacing its result are now **one action**. |
| `canonicalOutputFor(variant)` | a canonical **completed** step for the load-time probe. Without it `extractOutput` could be satisfied by `() => []` — compiling, registering, and shipping the same inert capability. |

Both are **compile-enforced** (`type-contract.ts`, mutation-proven below) *and* probed at registry load (clause 8: `extractOutput(canonicalOutputFor(v))` must yield ≥1 media with a non-empty url). So a new entry that gets this wrong is a **build or load failure**, never a silent inert capability.

Supporting pieces:

- **`mediaFromBlobs`** in a new leaf module `steps/output.ts` — ONE copy of the availability filter every entry uses. A leaf module, not `steps/index.ts`, because an entry importing a *value* back out of `index.ts` is a runtime import cycle whose safety would rest on function-declaration hoisting.
- **`getStepByOrchestratorType`** — the lookup both extractors use (a returned Workflow carries the `$type`, never the registry id), with cross-entry uniqueness enforced at load.
- 🔴 **Clause 9 forbids a registered entry from claiming a natively-extracted `$type`.** That is what makes placing the registry branch *before* the native `$type` filter provably behaviour-preserving — it can only reach `$type`s that previously fell through to `continue`.
- `convertImage`'s `canonicalOutputFor` is **copied from a real captured orchestrator response**, not written to match the extractor. A sample derived from the code would prove only that the code agrees with itself.

## 🔴 FIX 2 — the per-call `buzzBudget` ceiling was unenforced, permanently

See "The price is now MEASURED" above. The PR's earlier residual statement is retracted and corrected there.

## 🟡 FIX 3 — the registry had no entitlement declaration

`BlockRecipe` carries `resourceAllowlist` **and** `checkpointPolicy`; neither survived the generalization. That is the sharper omission of the pair, because entitlement is the axis this PR's own triage named as the real risk — `imageUpscaler` was disqualified *precisely* because its `model` takes an arbitrary AIR URN that would reach the orchestrator around the generation-graph entitlement belt. As written, a future entry with exactly that shape registered cleanly and every other invariant passed.

New required field `resourcePolicy: { kind: 'none' } | { kind: 'staticAllowlist'; airs }`. `checkpointPolicy` is folded in rather than duplicated — a checkpoint IS an AIR, so it is a `staticAllowlist` case. Only `'none'` is implemented; anything else fails at load (clause 11), the same unskippable property moderation already had. `convertImage` declares `'none'`.

🔴 **Enforced, not merely declared:** clause 7 deep-scans the **built** step for an AIR URN and rejects a `'none'` entry that emits one. **Honest limit, stated in the code:** it probes the *canonical* params, so it cannot see an AIR that appears only when an optional AIR-bearing field is supplied. It is a floor. The load-bearing guard for that case is the required declaration plus the unimplemented `staticAllowlist` gate.

## 🟡 FIX 4 — divergence correction was `prepaidFixed`-shaped code in the mode-agnostic section

It fired on `overage > 0` unconditionally. A future `timeBounded` entry reserves a *ceiling*; a submit-time cost above it would be topped up here **and** then settled ceiling→actual off the un-topped-up ceiling. Now gated on a new `plan.correctReservationOverage` — **the mode decides, the router stays mode-agnostic**, which is the property the registry exists to guarantee.

## 🟡 FIX 5 — the dev-session cap was not corrected on divergence

The comment said "correct **both** cap counters"; there are **three** reservations. A divergent submit inside a dev tunnel left `devSessionReserve` under-reading by the overage — the exact drift direction the same comment forbids. Added `chargeDevSessionOverage` (a plain INCRBY with no deny path, the exact mirror of `chargeAppSpendOverage`), wired it in, and corrected the comment.

## 🟡 FIX 6 — a bad param was a 500, and the tests pinned nothing

`estimateStepWorkflow`'s doc asserted "ZodError → BAD_REQUEST". **False by execution:** `paramSchema.parse` throws a raw ZodError from inside the resolver on both estimate and submit, and `getTRPCErrorFromUnknown` maps it to `INTERNAL_SERVER_ERROR`. The two covering tests asserted only `.rejects.toThrow()` — they passed either way. Now `safeParse` + an explicit `BAD_REQUEST`, and the tests assert the **code**.

*(Out of scope, noted for a follow-up: `customComfy`'s recipe param parse at `blocks.router.ts:5829`/`:5890` has the same shape. Changing it is a behaviour change to an existing kind and does not belong in this PR.)*

## 🟡 FIX 7 — the divergence detector could not report "never observed"

The correction reads `snapshot.cost?.total`, which `snapshotFromWorkflow` **omits** when there is no numeric cost, and the step submit passes no `wait`. Whether a queued `convertImage` carries `cost.total` was verified nowhere.

**Determined empirically:** a real non-whatif submit returned HTTP 202 / `status: processing` **with `cost.total: 1`**. The precondition holds today.

It is still not guaranteed for a future entry, so the counter is now `civitai_app_block_step_price_check_total{step, outcome}` — emitted on **every** submit with `outcome ∈ exact | over | absent`:

- `exact` rising = the price is right **and** the detector is provably live.
- `over` = the declared price is wrong (alert on this).
- `absent` = no numeric cost, so the correction is inert — the state that used to be indistinguishable from a healthy zero.

Both labels bounded and small; `outcome` is a TypeScript union so a caller cannot invent a fourth value.

## 🟢 FIX 8 — unused `getBlockSessionUser` in the step submit

Removed. Cost a `getUserById` per submit for a value never read, and was the 1 eslint warning the PR added.

---

## Verification

**Live orchestrator calls** — the whatif + real submit above; the real submit's terminal poll confirmed `step.output` has exactly the key `blob` (`available` false→true, `width` 797, `height` 1024), with **no `images` key and no `blobs` key**.

**Tests** — same globs as baseline, counted from `--reporter=verbose` (not an exit code), no `panic`/timeout truncation:

| | Files | Tests |
|---|---|---|
| pre-fix PR head `601052429b` | 149 | 3200 |
| this fix round | 149 | 3239 |
| **delta** | **0** | **+39** |

Both runs also report the same 1 pre-existing unhandled Prisma engine-load error (`prisma generate` cannot run on this NixOS box).

**tsc** — 6 errors at both the pre-fix head and here, byte-identical (ungenerated Prisma client + a `@civitai/shared` build artifact). **Zero new.**

**eslint / prettier** — measured over the identical changed-file set in a clean worktree at `601052429b` vs here:

| | errors | warnings |
|---|---|---|
| `601052429b` | 28 | 2 (`blocks.router.ts:6477 'user' unused`, `dev-tunnel.service.ts:670 'n' unused`) |
| here | 28 | 1 (only the pre-existing `dev-tunnel.service.ts:670`) |

prettier: the set of non-conforming files is unchanged from baseline (5, all pre-existing), and for each of them the **non-conforming regions are byte-identical** to baseline — i.e. every hunk this round adds is prettier-conformant, and no already-non-conforming file was reformatted. `step-registry.test.ts` was conforming at baseline and is conforming again.

### 🔴 No-regression re-proof for the existing kinds

A differential harness ran the **origin/main** extractors and the **HEAD** extractors side by side over a generated population — 15 `$type` values × 8 workflow statuses × available/unavailable × 8 content ratings × 4 output-key shapes, plus empty/absent/multi-step edge cases.

*(The PR's only pre-existing change to `workflow.service.ts` was exporting `formatStepTimeout`, so `origin/main` and `601052429b` are byte-identical for both extractors — one baseline covers both.)*

| | |
|---|---|
| population | **4611** workflow bodies |
| bodies containing ONLY natively-handled `$type`s | **2050** |
| **diffs on those 2050, `snapshotFromWorkflow`** | **0** |
| **diffs on those 2050, `projectAppWorkflow`** | **0** |
| total changed bodies | 46 — **every one contains `convertImage`** |

On the mixed multi-step body the pre-existing urls are preserved **in order** and the registered step's url is appended (`[a, b] → [a, b, c]`).

### Mutation table

Every guard added or changed was broken on purpose, its tests re-run, and the file restored **from a byte-exact backup with a sha256 equality assertion** — never a `str.replace` revert.

| # | Mutation | Test that failed | Killing assertion |
|---|---|---|---|
| M23 | drop the `resourcePolicy 'none'` AIR probe (clause 7) | *rejects a resourcePolicy 'none' whose BUILT step actually carries an AIR reference* (+1) | `expected [Function] to throw an error` (regex `/resourcePolicy 'none' is contradicted — buildStep\(\) emitted an AIR reference/`) |
| M24 | drop the `extractOutput` non-empty probe (clause 8) | *rejects an entry whose extractOutput surfaces NOTHING for its own canonical output* (+1) | `expected [Function] to throw an error` (regex `/extractOutput\(\) returned no media for canonicalOutputFor/`) |
| M25 | drop the extracted-media url check | *rejects extracted media with an empty url (a block would render a dead link)* | `expected [Function] to throw an error` (regex `/extractOutput\(\) returned media with no url/`) |
| M26 | drop the natively-extracted-`$type` shadow guard (clause 9) | *rejects an orchestratorType that SHADOWS a natively-extracted $type* | `expected [Function] to throw an error` (regex `/is natively extracted by workflow\.service/`) |
| M27 | drop the implemented-resource-policy gate (clause 11) | *rejects a resource policy with no implemented handler* | `expected [Function] to throw an error` (regex `/resourcePolicy 'staticAllowlist' has no implemented handler/`) |
| M28 | drop the cross-entry `orchestratorType` uniqueness guard | *rejects two entries claiming the SAME orchestratorType (single-winner lookup)* | `expected [Function] to throw an error` (regex `/orchestratorType 'fixtureType' is already claimed by 'a'/`) |
| M29 | drop the availability filter from `mediaFromBlobs` | *convert-image: a NOT-YET-AVAILABLE blob is dropped, not surfaced as a dead link* (+1) | `expected [ Array(1) ] to be undefined` |
| M30 | `convert-image` `extractOutput` reads `.blobs` instead of `.blob` — **the real shipped bug shape** | registry LOAD fails (module import) | `Error: block step 'convert-image' variant 'default': extractOutput() returned no media for canonicalOutputFor() — the step's result would be unreachable on both the snapshot and the app-workflow projection` |
| M31 | remove the registry branch from `snapshotFromWorkflow` — **the shipped bug** | *EVERY registered step surfaces its output on snapshotFromWorkflow.imageUrls* (+1) | `step 'convert-image' variant 'default': the caller is CHARGED for this step and its result is absent from every poll: expected undefined to deeply equal [ 'https://…' ]` |
| M32 | remove the registry branch from `projectAppWorkflow` — **the shipped bug, second surface** | *EVERY registered step surfaces its output on projectAppWorkflow.images* (+1) | `step 'convert-image' variant 'default': queryAppWorkflows returns images: [] for a generation the app paid for: expected [] to deeply equal [ Array(1) ]` |
| M33 | gate the per-call budget on the DECLARED price again (**the pre-fix behaviour**) | *REJECTS when the ORCHESTRATOR quote exceeds the budget even though the DECLARED price does not* | `expected 'processing' to be 'failed'` |
| M34 | reserve the declared price, ignoring the quote | *REJECTS when the ORCHESTRATOR quote exceeds the budget…* (+1: *RESERVES the quoted price…*) | `expected 'processing' to be 'failed'` |
| M35 | fail OPEN on an unquotable step | *FAILS CLOSED when the orchestrator returns NO price quote (no submit, no reserve)* | `expected 'processing' to be 'failed'` |
| M37 | drop the DEV-SESSION overage correction | *CORRECTS the DEV-SESSION counter too (three reservations, not two)* | `expected "vi.fn()" to be called with arguments: [ 'devsess_1', 8 ]` |
| M38 | revert to a raw `paramSchema.parse` (ZodError → 500) | *REJECTS an out-of-bounds param as BAD_REQUEST (not a 500)* (+3) | `expected TRPCError: […] to match object { code: 'BAD_REQUEST' }` |
| M39 | stop emitting the non-divergent `'exact'` outcome | *records outcome 'exact' on a healthy submit — the PROOF the check is live* | `expected "vi.fn()" to be called with arguments: [ 'convert-image', 'exact' ]` |
| M40 | put the caller's `externalId` on the FREE quote | *IDEMPOTENCY: a replayed submit does NOT double-charge* (+1) | `expected 'blk08apb_teststep-key' to be undefined` |
| M41 | make `chargeDevSessionOverage` a DECRBY | *tops the counter up unconditionally — with NO deny path, even past the cap* (+1) | `expected { allowed: true, total: -19 } to deeply equal { allowed: false, total: 140 }` |
| M42 | flip `prepaidFixed`'s `correctReservationOverage` to `false` | **6 tests** — *CORRECTS both cap counters…*, *…DEV-SESSION counter too*, all three `outcome` tests, *prepaidFixed plans a FINAL reservation…* | `expected "vi.fn()" to be called with arguments: [ …(2) ]` |

**Compile-time contract, mutation-proven** (`tsc --noEmit`, since vitest does not typecheck):

| Mutation | Result |
|---|---|
| `resourcePolicy?:` optional on `BlockStepBase` | ✅ `type-contract.ts(119,3): error TS2344` (`_RequiresResourcePolicy`) |
| widen `StepResourcePolicy.kind` to `string` | ✅ `type-contract.ts(124,3): error TS2344` (`_RejectsUnknownResourcePolicy`) |
| `extractOutput?:` optional | ✅ `type-contract.ts(137,3): error TS2344` (`_RequiresExtractOutput`) |
| `canonicalOutputFor?:` optional | ✅ `type-contract.ts(145,3): error TS2344` (`_RequiresCanonicalOutput`) |

**Reachability re-check.** Renumbering the invariant clauses could have re-broken the ordering bug the first audit found (clauses that no earlier guard lets you reach). Re-ran the three order-sensitive pre-existing mutations: all still die to **their own** message, not a neighbour's —

- M20-re (drop `timeBounded maxBuzz===ceil`) → `expected [Function] to throw error matching /timeBounded maxBuzz \(90\) must equal…/ but got 'block step 'fixture-step': billingM…'`
- M21-re (drop `timeBounded` positive-timeout) → `…matching /timeBounded stepTimeoutSeconds \(0\) …/ but got 'block step 'fixture-step': billingM…'`
- M22-re (drop `tokenMetered` positive-maxBuzz) → `…matching /tokenMetered maxBuzz \(0\) must be a …/ but got 'block step 'fixture-step': billingM…'`

That "but got" is the point: with the guard removed the fixture falls through to a *neighbouring* guard and still throws — and the test fails anyway, because it demands its own guard's message.

### 🔴 One mutation SURVIVED — reported, not hidden

**M36: deleting FIX 4's `if (plan.correctReservationOverage)` gate kills ZERO tests.** Only `prepaidFixed` is registered, so the flag is always `true` and removing the gate cannot change behaviour today. **This is an INVARIANT GUARD, not regression coverage, and it is labelled as such in the test file.** What *is* proven is that the flag is load-bearing: M42 (flipping it to `false`) kills 6 tests, so the router genuinely reads it. The false branch — and real coverage for this guard — arrives when a `timeBounded` entry is registered.

### DIFF
--- a/src/server/metrics/app-block-runtime.metrics.ts
+++ b/src/server/metrics/app-block-runtime.metrics.ts
@@ -234,6 +234,7 @@ type Bundle = {
   customComfyWallclockSeconds: Histogram<string>;
   capLimitsDegradedTotal: Counter<string>;
   spendCapRejectionsTotal: Counter<string>;
+  stepPriceDivergenceTotal: Counter<string>;
 };
 
 // ── customComfy per-engine runtime/cost buckets ──────────────────────────────
@@ -377,6 +378,30 @@ export function ensureRegisterAppBlockRuntimeMetrics(reg: Registry = client.regi
     ['reason']
   );
 
+  // ── `kind: 'step'` prepaidFixed PRICE DIVERGENCE ─────────────────────────────
+  // 🔴 This counter instruments the ONE input to the step bridge's money path
+  // that could not be verified before shipping: the orchestrator's ACTUAL price
+  // for a `prepaidFixed` step type. The registry DECLARES an exact price, gates
+  // the block token's per-call budget against it, and reserves it on every cap
+  // before submit. If the orchestrator ever bills MORE than the declared price,
+  // the declaration is wrong and the entry needs a higher price — or a different
+  // billing mode entirely, because "deterministic cost knowable before
+  // execution" would have turned out to be false for it.
+  //
+  // The submit path corrects the cap counters for the overage itself; this
+  // counter is what makes the correction VISIBLE instead of silent. A non-zero
+  // rate here is a registry bug, not a user problem.
+  //
+  // Cardinality: one `step` label, drawn from the code-owned registry keys
+  // (never client input — the wire enum is derived from those same keys, so an
+  // unregistered id cannot reach here). One series per registered step type.
+  const stepPriceDivergenceTotal = getOrCreateCounter(
+    reg,
+    'civitai_app_block_step_price_divergence_total',
+    "App Block `kind:'step'` submits whose REALIZED orchestrator cost exceeded the registry's DECLARED prepaidFixed price, by step id. Non-zero means the declared price is wrong.",
+    ['step']
+  );
+
   return {
     requestsTotal,
     requestDurationSeconds,
@@ -385,7 +410,25 @@ export function ensureRegisterAppBlockRuntimeMetrics(reg: Registry = client.regi
     customComfyWallclockSeconds,
     capLimitsDegradedTotal,
     spendCapRejectionsTotal,
+    stepPriceDivergenceTotal,
   };
+}
+
+/**
+ * Fail-soft emit of one `prepaidFixed` step price divergence (realized cost
+ * exceeded the registry's declared price). Called from the step submit path
+ * AFTER the money has already moved.
+ *
+ * 🔴 TOTAL, like every emitter in this module. It reports on an already-billed
+ * submit; a metrics error must never turn a successful generation into a 500.
+ */
+export function recordStepPriceDivergence(step: string): void {
+  try {
+    const { stepPriceDivergenceTotal } = ensureRegisterAppBlockRuntimeMetrics();
+    stepPriceDivergenceTotal.inc({ step });
+  } catch {
+    /* instrument-only — never let a metrics error touch an already-billed submit */
+  }
 }
 
 /**
--- a/src/server/metrics/app-block-runtime.metrics.ts
+++ b/src/server/metrics/app-block-runtime.metrics.ts
@@ -234,7 +234,7 @@ type Bundle = {
   customComfyWallclockSeconds: Histogram<string>;
   capLimitsDegradedTotal: Counter<string>;
   spendCapRejectionsTotal: Counter<string>;
-  stepPriceDivergenceTotal: Counter<string>;
+  stepPriceCheckTotal: Counter<string>;
 };
 
 // ── customComfy per-engine runtime/cost buckets ──────────────────────────────
@@ -378,28 +378,43 @@ export function ensureRegisterAppBlockRuntimeMetrics(reg: Registry = client.regi
     ['reason']
   );
 
-  // ── `kind: 'step'` prepaidFixed PRICE DIVERGENCE ─────────────────────────────
-  // 🔴 This counter instruments the ONE input to the step bridge's money path
-  // that could not be verified before shipping: the orchestrator's ACTUAL price
-  // for a `prepaidFixed` step type. The registry DECLARES an exact price, gates
-  // the block token's per-call budget against it, and reserves it on every cap
-  // before submit. If the orchestrator ever bills MORE than the declared price,
-  // the declaration is wrong and the entry needs a higher price — or a different
-  // billing mode entirely, because "deterministic cost knowable before
-  // execution" would have turned out to be false for it.
+  // ── `kind: 'step'` prepaidFixed PRICE CHECK ──────────────────────────────────
+  // 🔴 Instruments whether the registry's DECLARED price still matches what the
+  // orchestrator actually bills for a `prepaidFixed` step type. A declared price
+  // that drifts below the real one leaves every cap counter short.
   //
-  // The submit path corrects the cap counters for the overage itself; this
-  // counter is what makes the correction VISIBLE instead of silent. A non-zero
-  // rate here is a registry bug, not a user problem.
+  // 🔴 WHY THIS FIRES ON EVERY SUBMIT, NOT ONLY ON A DIVERGENCE. A
+  // divergence-only counter cannot distinguish "the price is right" from "the
+  // detector never ran" — both read as a flat zero forever, and the second is
+  // the state where the mitigation is inert. That is not a hypothetical: the
+  // detector reads `snapshot.cost?.total`, which `snapshotFromWorkflow` OMITS
+  // when the orchestrator returns no numeric cost, and the step submit passes no
+  // `wait`, so the response returns as soon as the job queues. (Measured
+  // 2026-08-02 against the live orchestrator: a queued `convertImage` submit —
+  // HTTP 202, status `processing` — DOES carry `cost.total`, so the precondition
+  // holds today. It was unverified before, and it is not guaranteed for a future
+  // step type.)
   //
-  // Cardinality: one `step` label, drawn from the code-owned registry keys
-  // (never client input — the wire enum is derived from those same keys, so an
-  // unregistered id cannot reach here). One series per registered step type.
-  const stepPriceDivergenceTotal = getOrCreateCounter(
+  // So the `outcome` label carries the non-divergent case too:
+  //   exact  — realized cost ≤ the reservation. The healthy state, and PROOF the
+  //            detector is live.
+  //   over   — realized cost EXCEEDED the reservation. A registry bug: the entry
+  //            needs a higher price, or a different billing mode, because
+  //            "deterministic cost knowable before execution" was false for it.
+  //   absent — no numeric cost on the snapshot, so nothing could be compared.
+  //            The state that used to be indistinguishable from `exact`.
+  //
+  // Alert on `outcome="over"`; read `outcome="exact"` to confirm the check runs
+  // at all; investigate a rising `outcome="absent"`.
+  //
+  // Cardinality: `step` is drawn from the code-owned registry keys (never client
+  // input — the wire enum derives from those same keys, so an unregistered id
+  // cannot reach here); `outcome` is a closed 3-value set. Bounded and small.
+  const stepPriceCheckTotal = getOrCreateCounter(
     reg,
-    'civitai_app_block_step_price_divergence_total',
-    "App Block `kind:'step'` submits whose REALIZED orchestrator cost exceeded the registry's DECLARED prepaidFixed price, by step id. Non-zero means the declared price is wrong.",
-    ['step']
+    'civitai_app_block_step_price_check_total',
+    "App Block `kind:'step'` prepaidFixed price checks at submit, by step id and outcome (exact = realized cost within the reservation; over = the orchestrator billed MORE than reserved, i.e. the declared price is wrong; absent = the snapshot carried no numeric cost, so no comparison was possible)",
+    ['step', 'outcome']
   );
 
   return {
@@ -410,22 +425,32 @@ export function ensureRegisterAppBlockRuntimeMetrics(reg: Registry = client.regi
     customComfyWallclockSeconds,
     capLimitsDegradedTotal,
     spendCapRejectionsTotal,
-    stepPriceDivergenceTotal,
+    stepPriceCheckTotal,
   };
 }
 
 /**
- * Fail-soft emit of one `prepaidFixed` step price divergence (realized cost
- * exceeded the registry's declared price). Called from the step submit path
- * AFTER the money has already moved.
+ * The closed outcome set for `civitai_app_block_step_price_check_total`. Keeping
+ * it a union (rather than a bare string) is what bounds the label cardinality at
+ * the type level — a caller cannot invent a fourth value.
+ */
+export type StepPriceCheckOutcome = 'exact' | 'over' | 'absent';
+
+/**
+ * Fail-soft emit of ONE `prepaidFixed` step price check. Called from the step
+ * submit path on EVERY submit, after the money has already moved.
+ *
+ * 🔴 Emitted unconditionally — including `outcome: 'exact'` — so a flat
+ * divergence line can be told apart from a detector that never ran. See the
+ * counter's definition above for why that distinction is load-bearing.
  *
  * 🔴 TOTAL, like every emitter in this module. It reports on an already-billed
  * submit; a metrics error must never turn a successful generation into a 500.
  */
-export function recordStepPriceDivergence(step: string): void {
+export function recordStepPriceCheck(step: string, outcome: StepPriceCheckOutcome): void {
   try {
-    const { stepPriceDivergenceTotal } = ensureRegisterAppBlockRuntimeMetrics();
-    stepPriceDivergenceTotal.inc({ step });
+    const { stepPriceCheckTotal } = ensureRegisterAppBlockRuntimeMetrics();
+    stepPriceCheckTotal.inc({ step, outcome });
   } catch {
     /* instrument-only — never let a metrics error touch an already-billed submit */
   }
--- a/src/server/routers/__tests__/blocks.router.workflow.test.ts
+++ b/src/server/routers/__tests__/blocks.router.workflow.test.ts
@@ -135,17 +135,33 @@ const {
 // F4: submitWorkflow dynamically imports the dev-tunnel spend backstop. Mock the
 // module so the default (no active tunnel → getActiveDevTunnel null) leaves every
 // existing test unchanged, and the F4 tests can drive an active dev session.
-const { mockGetActiveDevTunnel, mockReserveDevSessionBuzz, mockRefundDevSessionBuzz } = vi.hoisted(
-  () => ({
-    mockGetActiveDevTunnel: vi.fn(async () => null as unknown),
-    mockReserveDevSessionBuzz: vi.fn(async () => ({ allowed: true, total: 0 })),
-    mockRefundDevSessionBuzz: vi.fn(async () => undefined),
-  })
-);
+const {
+  mockGetActiveDevTunnel,
+  mockReserveDevSessionBuzz,
+  mockRefundDevSessionBuzz,
+  mockChargeDevSessionOverage,
+} = vi.hoisted(() => ({
+  mockGetActiveDevTunnel: vi.fn(async () => null as unknown),
+  mockReserveDevSessionBuzz: vi.fn(async () => ({ allowed: true, total: 0 })),
+  mockRefundDevSessionBuzz: vi.fn(async () => undefined),
+  mockChargeDevSessionOverage: vi.fn(async () => undefined),
+}));
+// 🔴 FIX 7 — the price-CHECK counter fires on EVERY step submit, not only on a
+// divergence, so a flat "no divergence" line can be told apart from a detector
+// that never ran. Mocked so the outcome label is assertable; the real emitter is
+// fail-soft and would swallow everything silently.
+const { mockRecordStepPriceCheck } = vi.hoisted(() => ({
+  mockRecordStepPriceCheck: vi.fn(() => undefined),
+}));
+vi.mock('~/server/metrics/app-block-runtime.metrics', () => ({
+  recordStepPriceCheck: (...a: unknown[]) => mockRecordStepPriceCheck(...(a as [])),
+}));
+
 vi.mock('~/server/services/blocks/dev-tunnel.service', () => ({
   getActiveDevTunnel: (...a: unknown[]) => mockGetActiveDevTunnel(...(a as [])),
   reserveDevSessionBuzz: (...a: unknown[]) => mockReserveDevSessionBuzz(...(a as [])),
   refundDevSessionBuzz: (...a: unknown[]) => mockRefundDevSessionBuzz(...(a as [])),
+  chargeDevSessionOverage: (...a: unknown[]) => mockChargeDevSessionOverage(...(a as [])),
 }));
 
 // G8 (per-app spend/velocity cap) + G6 (persistent output queue) — submitWorkflow
@@ -495,6 +511,8 @@ beforeEach(() => {
     mockGetActiveDevTunnel,
     mockReserveDevSessionBuzz,
     mockRefundDevSessionBuzz,
+    mockChargeDevSessionOverage,
+    mockRecordStepPriceCheck,
     mockReserveAppSpend,
     mockRefundAppSpend,
     mockChargeAppSpendOverage,
@@ -556,6 +574,7 @@ beforeEach(() => {
   mockGetActiveDevTunnel.mockResolvedValue(null);
   mockReserveDevSessionBuzz.mockResolvedValue({ allowed: true, total: 0 });
   mockRefundDevSessionBuzz.mockResolvedValue(undefined);
+  mockChargeDevSessionOverage.mockResolvedValue(undefined);
   // W3 flow A default: the spend-attribution write resolves successfully.
   // Tests that exercise best-effort override it to reject.
   mockRecordSpendAttribution.mockResolvedValue({
@@ -5861,6 +5880,48 @@ describe("step-type registry bridge (kind: 'step')", () => {
     });
   }
 
+  // 🔴 A step submit makes TWO orchestrator calls: the `whatif:true` PRICE QUOTE
+  // that backs the per-call `buzzBudget` gate, then the real submit. The quote is
+  // free and side-effect-free; the REAL submit is the one that spends. Assertions
+  // about "did we spend" must therefore look at the real calls only —
+  // `mockSubmitWorkflow` alone can no longer tell the two apart, and a test that
+  // ignored the difference would silently stop proving anything about spend.
+  // (This mirrors the txt2img path, which has run whatIf → budget gate → cap
+  // reservations → real submit since before this PR.)
+  const isWhatIfCall = (c: unknown[]) =>
+    (c[0] as { query?: { whatif?: boolean } } | undefined)?.query?.whatif === true;
+  const realSubmitCalls = () => mockSubmitWorkflow.mock.calls.filter((c) => !isWhatIfCall(c));
+  const whatIfSubmitCalls = () => mockSubmitWorkflow.mock.calls.filter((c) => isWhatIfCall(c));
+
+  /**
+   * Drive the two orchestrator calls INDEPENDENTLY: what the quote says, and
+   * what the real submit ends up costing.
+   *
+   * 🔴 Needed because they are no longer the same number. A price DIVERGENCE now
+   * means "the quote under-read what was billed" — with `mockResolvedValue`
+   * returning one cost for both, a divergence test would reserve the high number
+   * from the quote and then find no overage at all, i.e. pass while proving
+   * nothing.
+   */
+  function stepSubmitQuoting(quotedCost: number | null, realizedCost: number) {
+    mockSubmitWorkflow.mockImplementation(async (opts: { query?: { whatif?: boolean } }) => {
+      if (opts?.query?.whatif === true) {
+        return {
+          id: 'wf_quote',
+          status: 'unassigned',
+          steps: [],
+          ...(quotedCost === null ? {} : { cost: { total: quotedCost } }),
+        };
+      }
+      return {
+        id: 'wf_step_1',
+        status: 'processing',
+        steps: [{ $type: 'convertImage', output: {} }],
+        cost: { total: realizedCost },
+      };
+    });
+  }
+
   const caller = () => blocksRouter.createCaller(fakeCtx() as never);
 
   describe('estimateWorkflow', () => {
@@ -5886,7 +5947,12 @@ describe("step-type registry bridge (kind: 'step')", () => {
       expect(mockReserveAppSpend).toHaveBeenCalledWith('apb_test', estimate.snapshot.cost.total);
     });
 
-    it('REJECTS an out-of-bounds param through the step .strict() schema', async () => {
+    // 🔴 FIX 6 — the CODE is the assertion. `paramSchema.parse()` threw a raw
+    // ZodError from inside the resolver, which tRPC's getTRPCErrorFromUnknown
+    // maps to INTERNAL_SERVER_ERROR — so this app-author-driven surface reported
+    // routine param typos as 5xx. `.rejects.toThrow()` passed either way and
+    // pinned nothing.
+    it('REJECTS an out-of-bounds param as BAD_REQUEST (not a 500)', async () => {
       mockVerifyBlockToken.mockResolvedValue(stepClaims());
       happyUser();
       await expect(
@@ -5896,7 +5962,24 @@ describe("step-type registry bridge (kind: 'step')", () => {
             params: { image: 'https://evil.example/x.png', output: { format: 'webp' } },
           }),
         })
-      ).rejects.toThrow();
+      ).rejects.toMatchObject({ code: 'BAD_REQUEST' });
+    });
+
+    it('REJECTS an out-of-RANGE bounded param as BAD_REQUEST', async () => {
+      mockVerifyBlockToken.mockResolvedValue(stepClaims());
+      happyUser();
+      await expect(
+        caller().estimateWorkflow({
+          blockToken: 'tok',
+          body: stepBody({
+            params: {
+              image: 'https://image.civitai.com/source.png',
+              output: { format: 'webp' },
+              transforms: [{ type: 'resize', targetWidth: 99999 }],
+            },
+          }),
+        })
+      ).rejects.toMatchObject({ code: 'BAD_REQUEST' });
     });
 
     it('REJECTS a model token (registry steps are page-only)', async () => {
@@ -5938,12 +6021,16 @@ describe("step-type registry bridge (kind: 'step')", () => {
           dailyKey: 'system:blocks:app-spend-cap:apb_test:day',
         };
       });
-      mockSubmitWorkflow.mockImplementation(async () => {
-        order.push('submitWorkflow');
+      mockSubmitWorkflow.mockImplementation(async (opts: { query?: { whatif?: boolean } }) => {
+        order.push(opts?.query?.whatif === true ? 'whatIfQuote' : 'submitWorkflow');
         return { id: 'wf_step_1', status: 'processing', steps: [], cost: { total: 1 } };
       });
       await caller().submitWorkflow({ blockToken: 'tok', body: stepBody() });
-      expect(order).toEqual(['reserveAppSpend', 'submitWorkflow']);
+      // 🔴 The free QUOTE precedes the cap belt (so the per-call budget gate runs
+      // against the orchestrator's own number), and the SPENDING submit follows
+      // the per-app reservation. Both halves matter: a reserve after the real
+      // submit would be a spend the cap never saw.
+      expect(order).toEqual(['whatIfQuote', 'reserveAppSpend', 'submitWorkflow']);
     });
 
     it('a per-app DAILY cap rejection returns the daily reason and NEVER submits', async () => {
@@ -5963,7 +6050,10 @@ describe("step-type registry bridge (kind: 'step')", () => {
         cost: { total: STEP_PRICE },
       });
       expect(result.snapshot.error).toMatch(/app daily spend cap reached/);
-      expect(mockSubmitWorkflow).not.toHaveBeenCalled();
+      expect(
+        realSubmitCalls(),
+        'the free QUOTE may have run; NOTHING may have been SPENT'
+      ).toHaveLength(0);
       // The viewer's OWN daily ceiling is given back — a rejected submit must not
       // burn it for a spend that never happened.
       expect(mockSysRedis.decrBy).toHaveBeenCalledWith(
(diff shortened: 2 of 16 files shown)


ZacxDev:
## Polish round — 4 fixes (`5962389021`)

Delta re-audit came back safe-to-merge with no deploy-blockers. These are four cheap items only; nothing the re-audit verified was touched. Posting here rather than editing the PR body so the existing fix-round sections stay as written — fold in if you prefer.

### A — the moderation/entitlement asymmetry

`moderationPosture` was re-asserted at request time (`blocks.router.ts:6495`). `resourcePolicy` was **not** — it was enforced at registry **load** only. Clause 7 probes `buildStep(canonicalParamsFor(v))`, the *canonical* params, so an entry with an optional `model?: string` absent from its canonical params registers cleanly, and at request time `buildStep` forwards whatever the iframe sent straight into the submitted input.

Submit now runs the same `containsAirReference` scan (imported from the registry, not re-implemented) on `built.input` for `resourcePolicy: 'none'` entries, immediately after `step.buildStep(params)` — before the quote and before every reservation, so a rejection costs no orchestrator call and has nothing to refund. `FORBIDDEN`, fail-closed.

**Not overclaiming this.** Exploitability *today is nil*: it needs a future registered entry, and `ConvertImageInput` has no AIR field (confirmed against the generated client). What changed is that entitlement stopped being a review-process guarantee and became a runtime one — on the axis this PR's own triage named as the real risk, and the only one that had no runtime backstop.

One honest side-effect, documented at the guard: the scan is a substring test for `urn:air:`, so a param that merely *embeds* that literal is rejected too. That is a false positive and it is the direction we want.

### B — a false comment on a money path

`blocks.router.ts:6421-6425` still asserted *"an estimate can never quote a different number than the submit bills."* True before this PR's own fix round; false since submit began gating and reserving `max(declared, quoted)`. Concretely: rate card moves to 40, the block is shown `cost.total: 1`, submit returns `insufficient buzz budget: step price 40 exceeds budget 5` or reserves 40 against every cap. Behaviour is fail-closed and correct — the comment was the defect. Rewritten to state the estimate is the declared **floor**, not the enforced ceiling, with the retraction named.

### C — anchoring clause 8 to the generated contract

Clause 8 (`steps/index.ts:747-761`) is a **self-consistent pair, not a contract check** — probe input and extractor are authored by the same person in the same file. Two escapes reproduced: an extractor ignoring its argument paired with `canonicalOutputFor: () => ({})`, and an extractor reading `output.images` paired with a sample encoding `output.images`. Only `() => []` is caught.

`ConvertImageOutputStepLike` is now exported and `type-contract.ts` asserts against `@civitai/client` that a real `ConvertImageStep` satisfies it, that every key it reads exists on `ConvertImageOutput` / `ImageBlob`, and that a generated `ImageBlob` is accepted by the shared `OrchestratorBlobLike`.

🔴 **This does not close the escape in general, and the file says so.** A future entry supplies its own read shape and its own sample, and nothing forces it to add an assertion — the escape is open for entry #2. Only the `OrchestratorBlobLike` ↔ `ImageBlob` assertion is entry-independent (it covers any entry feeding a generated blob through `mediaFromBlobs`). What is anchored is the one registered entry.

Demonstrated rather than asserted: renaming `blob` → `images` **consistently** across the read shape, the extractor and the canonical sample leaves all 42 registry tests **green** — and fails `tsc` on three named aliases.

**Proposal, not built here.** To make the anchoring required for every entry you'd want `BlockStep` to carry its generated step type, e.g. `extractOutput(step: TStep)` / `canonicalOutputFor(v): TStep` with `TStep` constrained to the generated type for the entry's `orchestratorType`. The blocker: `@civitai/client` exports no discriminated union of all step types — only the individual `<Name>Step` aliases — so the `$type` → type map would itself be hand-written, which reintroduces the same authored-by-us link one level up. Worth doing if the client ever exports the union, or if a generated map is cheap to add upstream.

### D — the one unlabelled invariant guard

`blocks.router.workflow.test.ts:6461` iterated `mock.calls` with no denominator. Fixed rather than labelled (one line). **Control run**: with the new line removed, deleting all four `recordStepPriceCheck` call sites leaves the test green — that is the pre-fix vacuity, reproduced. With it, the same mutation fails on `expected "vi.fn()" to be called 2 times, but got 0 times`.

---

### Verification

Worktree off `fa2f70d13e`, flake node 22.22.2, counts from `--reporter=verbose`, no `panic`/timeout truncation.

| | baseline `fa2f70d13e` | HEAD |
|---|---|---|
| test files | 149 | 149 |
| tests | 3239 | **3241** (+2, both new) |
| tsc | 1 pre-existing error | **byte-identical**, measured back-to-back in the same env |
| eslint (changed set) | 16 errors / 0 warnings | 16 errors / 0 warnings |
| prettier | 2 non-conforming files | same 2, same regions; `type-contract.ts` conforming at both |

*(The tsc baseline is 1 rather than the 6 reported earlier in this PR — the five `@civitai/shared` errors resolved when that package's build artifact appeared in the shared `node_modules`. Both sides were measured in the same environment minutes apart and diff byte-identically, which is the claim that matters. `prisma generate` still cannot run on this box; CI is authoritative.)*

**Extraction no-regression, re-proved.** 7,687 generated workflow bodies (15 `$type`s × 8 statuses × available/unavailable × 8 content ratings × 4 output-key shapes, plus 7 edge cases) through `origin/main`'s extractors and HEAD's:

| | |
|---|---|
| population | **7,687** |
| bodies containing ONLY natively-extracted `$type`s | **2,052** |
| diffs on those, `snapshotFromWorkflow` | **0** |
| diffs on those, `projectAppWorkflow` | **0** |
| total changed bodies | 65 — **every one contains `convertImage`** |

On the mixed multi-step body the pre-existing urls are preserved **in order** and the registered step's url lands at its step position (`[901] → [901, 902, 903]`).

### Mutation table

Byte-exact backup + sha256 equality assertion on restore; never a `str.replace` revert.

| # | Mutation | Result | Killing assertion |
|---|---|---|---|
| MA1 | delete the request-time AIR guard | 1 failed — *a built step that smuggles an AIR is rejected at SUBMIT under resourcePolicy 'none'* | `promise resolved "{ snapshot: { …(3) } }" instead of rejecting` |
| MA2 | invert the guard's predicate (fail OPEN on clean input) | 31 failed, incl. the negative control *the SAME params without the AIR are accepted* | proves the line **executes on every submit** |
| MA3 | reword ONLY the guard's message | 1 failed, same test | `expected TRPCError: step 'convert-image' declares … to match object { code: 'FORBIDDEN', …(1) }` |
| MD0 | **control** — remove the new denominator, then delete all 4 call sites | 3 failed; the closed-set test **SURVIVES** | reproduces the vacuity |
| MD1 | delete all 4 `recordStepPriceCheck` call sites | 4 failed, incl. the closed-set test | `expected "vi.fn()" to be called 2 times, but got 0 times` |
| MC1 | rename `blob` → `images` **consistently** (read shape + extractor + sample) — the reproduced clause-8 escape | 42 registry tests **PASS**; `tsc` fails | `type-contract.ts(185,3)` `_RealConvertImageStepSatisfiesReadShape`, `(199,3)` `_ConvertImageReadKeysExist`, `(206,3)` `_ConvertImageBlobReadKeysExist` — all TS2344 |
| MC2 | typo `available` → `avaliable` in the read shape only | tsc | `type-contract.ts(206,3)` TS2344 `_ConvertImageBlobReadKeysExist`, alone |
| MC3 | drop `\| null` from the read shape's `url` | tsc | `type-contract.ts(185,3)` TS2344 `_RealConvertImageStepSatisfiesReadShape`, alone |
| MC4 | drop `\| null` from `OrchestratorBlobLike.width` | tsc | `type-contract.ts(220,3)` TS2344 `_MediaFromBlobsAcceptsGeneratedImageBlob` (+ a collateral TS2345 at the convert-image call site) |

MA3 is the row that matters for MA1: several sibling seams on this path also reject `FORBIDDEN`, so a code-only assertion would have been satisfied by the wrong guard. The test asserts the message.

**Reachability.** The new AIR guard is reached today through the **real** registered entry — `convert-image` forwards `params.image` verbatim into the submitted input, so an AIR URN embedded in an otherwise-valid Civitai-hosted url arrives in `built.input`. So this is regression coverage, not an invariant guard, and it carries a negative control (same params without the AIR are accepted) so a guard that rejected everything could not pass it.

### Not verified

- Nothing run against a live orchestrator this round (no behaviour change to what is submitted, except the new pre-submit rejection).
- `prisma generate` cannot run here — the 1 remaining tsc error is that; CI is authoritative.
- Browser/component tests not run (node-only change).

PR #3540, opened 2026-08-02
feat(apps): give the in-site Open CTA its own glyph (S6a)

ZacxDev:
> **Updated after two audit rounds.** The \"How\" section below describes the CURRENT implementation (`96ff86d550`). The original description was wrong about the blast radius; that correction and the full round-1 findings are preserved in [this comment](https://github.com/civitai/civitai/pull/3540#issuecomment-5159834018) rather than silently erased.

## What

`/apps/store-preview/<slug>` renders its primary CTA with a glyph keyed on the action's semantic mode, so **in-site `Open`** (`IconPlayerPlay`) and **off-site `Visit ↗`** (`IconExternalLink`) no longer look identical.

## Why

civitai **#3391** removed the kind + category badges from `AppListingCard` and `AppListingDetailBody`, on the stated grounds that *the on-site/off-site kind signal now rides the CTA*. Measured on prod, it did not:

| | `/apps/store-preview/model-benchmarking` | `/apps/store-preview/vitrine` |
|---|---|---|
| label | `Open` | `Visit` |
| href | `/apps/run/model-benchmarking` (no `target`) | external, `target="_blank"` |
| icon path `d` | `M12 6h-6a2 2 0 0 0 -2 2v10a2 2 0 0 0 2…` | **byte-identical** |
| fill / radius / height | `rgb(25,113,194)` / `4px` / `36` | **identical** |

Only the word differed. Both view models (`getDetailPrimaryAction`, `getListingCta`) already return the right discriminator — the components discarded it at the icon. This restores a premise a shipped PR was justified on.

## How

- **NEW `appListingActionGlyph.ts`** — one module carrying both the mapping (`open → launch`, `visit → external`, `connect`, `info`) and the `ACTION_GLYPH_ICONS` record it resolves to. Same shape as the existing `marketplaceCategoryIcons.ts`.
- **`AppListingDetailBody.PrimaryAction`** resolves the glyph **inside each branch**, from the mode that branch has established — not once up front from `action.mode`. `href` is optional on `DetailPrimaryAction`, so an `open`/`visit` action without one falls through to the informational return; a hoisted lookup would paint a launch icon there. The fallthrough hard-codes the `info` glyph.

`cardActionGlyph` **has no live caller yet.** It ships here so the follow-up store-card PR *imports* this module instead of editing it — that PR touches the same import list, which is the gotcha-#98 same-file collision shape. Its tests are a contract pin, not a regression gate, until the card reads through it. Separately, `getListingCta` never emits `'connect'` today (the off-site OAuth case routes to `'detail'`), so that arm is unreachable from live data even then; it is mapped for totality over the type.

## Blast radius

**`/apps/store-preview/<slug>` only** — 13 live listings, mod-gated + `deIndex`-ed.

Explicitly **not** the moderator listing-preview: `PrimaryAction` sits inside `{!preview && …}` (`AppListingDetailBody.tsx:574-577`) and the mod queue mounts `<AppListingDetailBody detail={detail} preview />` (`OffsiteReviewQueue.tsx:939`), so it never renders a primary action at all. (The original description claimed otherwise — corrected in the comment linked above.)

## Not changed

🔴 **Link semantics are byte-identical.** In-site CTA: no `target`, no `rel`. Off-site CTA: `target=\"_blank\"` + `rel=\"noopener noreferrer\"`. `recordRecentlyOpenedApp` still fires on the visit branch only. All asserted explicitly. Colour/radius/height are inherited from the Mantine `Button` variant and are untouched.

## Verification

**Node `unit` — 9/9.** **Component — 20/20** (16 pre-existing + 4 new). `tsc --noEmit`: 22 errors, **all pre-existing**, none in these files (error set byte-identical to before the change). `eslint`: 0 errors, 1 pre-existing warning (`ThemeIcon`, present on `main`).

### Break-it matrix — every guard mutated and watched go red for **its own** reason

| Mutation | Node | Component |
|---|---|---|
| `detailActionGlyph('open')` returns `'external'` | **4 red** (`expected 'external' not to be 'external'`, `… to be 'launch'`, `expected 3 to be 4`, cross-model agreement) | — |
| `ACTION_GLYPH_ICONS.launch = IconExternalLink` (icons collide, names don't) | **2 red**, 7 green | — |
| `PrimaryAction` hard-codes the external glyph | 🔴 **9/9 GREEN** | **1 red** — `expected 'external-link' not to be 'external-link'` |
| connect branch → `glyphFor('info')` | — | **1 red** — `expected 'info-circle' to be 'plug-connected'` |
| info fallthrough → `glyphFor('connect')` | — | **1 red** — `expected 'plug-connected' to be 'info-circle'` |
| **both of the above at once** | — | **2 red**, one per branch — only observable since the connect/info guards were split into separate tests |
| add a 5th `DetailActionMode` | **compile error** TS1360 — `Property 'archive' is missing` | — |

Row 3 is the honest one: **the node suite cannot see component wiring**, so the component tests are the only gate on it. Row 2 shows the icon-identity assertion isn't redundant with the string assertions.

### Harness validated before any verdict was read

The component suite's first runs in a fresh worktree were **16/16 red on pristine `origin/main` code** — a cold vite `optimizeDeps` cascade, not a failure. Only a warm-cache run is readable (pristine baseline then: 16/16 green in 1.06s). Recording it because the red reads exactly like a dual-React break. Locally, the first run after **any** module-graph change is garbage.

### Marker choice

The component tests assert the Tabler `class=\"tabler-icon-<name>\"` modifier — not the raw path `d` (breaks on any tabler bump), and not a `data-testid`: `next.config.mjs` strips those via `reactRemoveProperties` in production builds, **PR previews included**.

## Accepted, not fixed

**The guard this PR ships cannot block the regression from returning.** The GH `Unit tests` job is `continue-on-error: true` and the preview component suite is explicitly report-only. That is a repo-wide CI posture, not something this PR can change — the component tests document the invariant rather than enforce it. Making `component-tests` gate is the change that would deliver durability, and belongs in its own PR.

## Still owed

Live re-measurement on `/apps/store-preview/model-benchmarking` and `/apps/store-preview/vitrine` that the two CTAs' `path[d]` now differ. The component test proves it in Chromium against mocked data; that is not a claim about prod.


ZacxDev:
## Correction to the PR description + audit round 1

An adversarial audit found a **factual error in the PR body above**, plus five should-fix items. Correcting publicly rather than silently editing, since the wrong version may already have been read.

### 🔴 The PR body overstates the blast radius

It claims the change also affects *"the moderator listing-preview, which renders `AppListingDetailBody` in read-only `preview` mode (#3412)"*. **That is wrong.** `PrimaryAction` is rendered inside `{!preview && …}` (`AppListingDetailBody.tsx:557-561`), and the mod queue mounts it as `<AppListingDetailBody detail={detail} preview />` (`OffsiteReviewQueue.tsx:939`). The moderator preview never renders a primary action at all.

**Actual blast radius: `/apps/store-preview/<slug>` only** — 13 live listings, mod-gated + `deIndex`-ed. The error was in the safe direction (I claimed more reach than exists), and it came from the scope doc, which I've corrected there too.

### Fixed in `fd202c8687`

1. **Collapsed the two-module split back into one.** The split's stated rationale — keep the logic module node-importable — was **falsified by this PR's own test**, which imports the tabler module in the node project and passes in 4ms. And the precedent I cited (`marketplaceCategoryIcons.ts`) is a *single* module carrying its own tabler imports; its node test reads source text rather than importing. So the split invented a convention instead of following one.
2. **Glyph now resolved inside each branch**, from the mode that branch established, rather than once up front. `href` is optional on `DetailPrimaryAction`, so an `open`/`visit` action without one falls through to the informational return — the hoisted lookup would have painted a launch icon there. Not reachable from today's `getDetailPrimaryAction` (every `open`/`visit` return sets an href), so **latent, not live**. Worth fixing anyway: the pre-change code was correct *by construction*, the hoisted version only *by coincidence of the current view model*. The fallthrough now hard-codes the `info` glyph.
3. **A test comment asserted something false.** It claimed `readonly DetailActionMode[]` makes the mode list exhaustive — it doesn't, that type accepts any subset, so a 5th mode would typecheck and silently under-cover the distinctness assertions. Now `satisfies Record<DetailActionMode, true>`, which genuinely fails to compile on a missing key.
4. **Removed a vacuous assertion.** `resolves an icon for every glyph` cannot fail while the file typechecks (`Record<PrimaryActionGlyph, Icon>` already forbids a missing key) — it read as coverage while asserting nothing. Replaced with a real distinctness assertion over the card actions.
5. **`renderAndSettle` timeout 5000 → 10000**, matching the project-wide `vi.waitFor` default that `test/component-setup.tsx` sets deliberately for the saturated preview CI box. `vi.waitUntil` isn't covered by that patch.

**Re-verified after the restructure** (a fix round can disarm its own guard): mutating the `open` branch to `glyphFor('visit')` still fails the component gate with `expected 'external-link' not to be 'external-link'` — 17 pass, 1 fail. Node 9/9, component 18/18, `tsc` 22 pre-existing errors (none in these files), eslint 0 errors.

### Accepted, not fixed

**The guard this PR ships cannot block the regression from returning.** The GH `Unit tests` job is `continue-on-error: true` and the component suite is explicitly report-only in the preview pipeline. That's a repo-wide CI posture, not something this PR can change — but it means the two component tests document the invariant rather than enforce it. Making `component-tests` gate is the change that would deliver durability, and it belongs in its own PR.

**Not verified:** live re-measurement on `/apps/store-preview/model-benchmarking` vs `/apps/store-preview/vitrine` that the two CTAs' `path[d]` now differ. The component test proves it in Chromium against mocked data; that is not a claim about prod.

ZacxDev:
## Round 3 (delta re-audit) — clean on code

Third independent pass over `fd202c8687..993f31da58`. **No 🔴, no 🟡 in the code.** Applied the polish it did find, in `96ff86d550` (test-only; the source file is byte-identical to the previous commit).

**The one 🟡 was in this description, not the code — now fixed.** The blast-radius paragraph cited `AppListingDetailBody.tsx:557-561` for the `{!preview && …}` gate. That was right at `33b903d58b` but the later commits shifted it; at HEAD the gate is **`:574`** and `<PrimaryAction>` is **`:577`** (`:557-561` is the title/tagline block). Since the body claims to describe the current implementation and this is the *safety* claim, a reviewer following the pointer would have landed on unrelated code. The claim itself was independently verified as true — `OffsiteReviewQueue.tsx:939` mounts `<AppListingDetailBody detail={detail} preview />`, so the mod preview renders no primary action.

**Split the connect/info guards into two tests.** Two real reasons, not style:
1. `waitForCtaGlyph` scans the whole document and takes the first hit, and `afterEach(cleanup)` does **not** run between two renders inside one test — so the shared body biased the match toward the **older** mount if a label ever collided. One render per test makes each scan unambiguous by construction rather than by luck.
2. A connect-branch mutation short-circuited before the info assertion, so those two kills could never be observed on the same run. **Now they can** — mutating both branches at once fails both tests, each with its own assertion (`expected 'info-circle' to be 'plug-connected'` / `expected 'plug-connected' to be 'info-circle'`). Added as a row to the break-it matrix above.

Also wrapped an over-long `glyphOf` throw — a prettier deviation this PR introduced in round 1 that round 2's prettier fix didn't cover.

### What the auditor verified independently, that I'd only claimed

- The `info` fixture really reaches the model-slot branch: `liveUrl` is read **only** inside the `hasPage` guards, so `hasPage: false` cannot divert to the "Open live" visit branch. Corroborated by an existing green unit test (`__tests__/appListingDetailView.test.ts`) that runs `hasPage=false` against an https `liveUrl`. This was the highest-value thing to check — had it been wrong, the mutation kills would have been coincidental.
- No leftover from my two mutation/restore cycles; the four `glyphFor` literals are correct per branch.
- The cold-cache cascade reproduced exactly on their machine too (run 1 `no tests`, run 2 19/19 red with the `useContext` dual-React signature, run 3 warm green in 3.9s) — which is why gotcha #164 now exists.

One correction to my own matrix, in the honest direction: row 3's "component: 1 red" holds only under "hard-code in the `open` branch only"; hard-coding every branch would now red 3. An understatement, not an overstatement.

**Gates at `96ff86d550`:** node 9/9, component 20/20 (warm), `tsc` 22 pre-existing errors none in these files, eslint 0 errors.

Three rounds, each found something the previous missed. This round found nothing in the code, so I'm stopping here rather than chasing cosmetics.

ZacxDev:
## The red `preview / component-tests` is NOT from this PR — attributed, not ignored

Per gotcha #119 ("never merge past a report-only red without reading what it actually is") I read the pod log rather than the summary.

**This PR's file passed in CI:**
```
✓  component (chromium)  src/components/Apps/AppListingDetailBody.browser.test.tsx (20 tests) 2237ms
```
20/20 — exactly matching the local run.

**The suite's 4 failures are in files this PR does not touch:**

| Failing file | Failures | Touched by #3540? |
|---|---|---|
| `src/components/AppBlocks/BlockHostTokenRetry.browser.test.tsx` | 3 | no |
| `src/components/Apps/AppsSubmitEditView.browser.test.tsx` | 1 | no |

Run total: `Tests 4 failed | 1118 passed (1122)`.

**Pre-existing, and red across the open-PR population — not just asserted, checked.** I pulled the component-test pod logs for three other in-flight PRs: **#3539** shows the **identical 4 failures** plus a 5th (`AppBlockChrome`), and the `BlockHostTokenRetry` failure markers are present in **#3541** and **#3529** as well. #3539 does not touch `BlockHostTokenRetry` either. So this is the same shape as the stale red that sat on every open PR for days before #3468 — 4-for-4 across the PRs I could sample.

The failing tests time out at ~4.7–5.2s and ~15.2s, which is the timeout signature rather than an assertion mismatch — consistent with a flake/timing class, though I have **not** root-caused it and am not claiming to.

🔴 **Flagging rather than routing around it.** #119's rule is "if it's stale, FIX it rather than routing around it", and `main`'s most recent commit (#3537) was itself *"unblock the permanently-red preview gate"* — so this area is actively troubled and a standing red here trains everyone to click through. That fix is its own PR, not this one; raising it as a separate item rather than silently expanding this PR's scope.

**Nothing here blocks #3540** — its own suite is green in CI and locally, and the check is non-blocking by configuration.

ZacxDev:
## The second red (`preview / smoke-tests`) is a STALE-BRANCH artifact — read before merging

A `preview / smoke-tests` failure appeared after my earlier check-sweep, so recording it rather than merging past it.

**The single failure:** `tests/preview-generation.spec.ts:48 › generation cost quote (gold) › whatIfFromGraph fires on /generate and returns a numeric cost` — `page.waitForResponse` timeout, 3 attempts. `60 passed, 1 failed`.

**It is already fixed on `main`, and this branch simply predates the fix:**

- this PR does **not** touch `tests/preview-generation.spec.ts` (`git diff --name-only <base>..HEAD` → only the 4 glyph files);
- **#3541 — "test(preview): fix whatIfFromGraph smoke test against tRPC request batching"** — is the literal fix for this exact test, is on `main`, and is **not** in this branch (it is 2 commits behind);
- since this PR never modifies that file, the **merged tree takes main's fixed version**, so the red cannot ride forward.

That is also the honest framing of the earlier `component-tests` red: neither failure is caused by this change, and neither survives the merge — the component one is fixed by **#3544**, this one by **#3541**, both already independent of this PR.

Blocking checks are green: **Typecheck ✅, Unit tests ✅, ESLint + Prettier ✅, event-engine-common pin ✅**, plus `preview / deploy` and `preview / render-check`. `mergeable: MERGEABLE`, no open PR is based on this branch (so the stacked-parent branch-deletion hazard does not apply).

Merging on that basis.

PR #3544, opened 2026-08-02
test(apps): unbreak the red component suite — a crashing subtree and a self-racing fixture

ZacxDev:
## Why

`preview / component-tests` has been **red on every open PR**. It is report-only, so nothing blocked — which is exactly how it survived. Same shape as the stale smoke red that sat on every PR for days before #3468, and `main`'s recent #3537 was itself *"unblock the permanently-red preview gate"*.

🔴 **Correction to an earlier version of this description** (an audit caught it): I claimed #3540, #3539, #3541 and #3529 "all carry the same failures". Only **two of those four check statuses** support that — #3540 and #3539 report `component-tests fail`; **#3541** reports *"Component suite exceeded its runner budget — no verdict, NOT a test failure"* and **#3529** has no `component-tests` status at all. What is true for all four: I pulled the component-test **pod logs** for #3541 and #3529 and both contain the same `BlockHostTokenRetry` failure markers — so the tests do fail there, but their *checks* did not report it as a test failure. The stronger "population" phrasing overstated the evidence.

I also quoted a baseline of `4 failed | 1118 passed (1122)`. That total is from **#3540's** run, whose tree carries its own 4 added tests — not this branch's baseline. **This branch's base has 1118 tests**, and a full warm run here is **1118/1118 green** with the fixes applied.

### 1. `BlockHostTokenRetry.browser.test.tsx` — 3 failed

**#3497** added `useFeatureFlags()` to `IframeHost` (the in-host "open the page" affordance). The real hook **throws** without a `FeatureFlagsProvider`, which this scaffold does not mount — so the whole subtree crashed at mount. `BlockHost` rendered nothing and `waitForIframe` timed out as `iframe never mounted`.

🔴 **Correction — my original diagnosis of the other three tests was WRONG.** I claimed they "kept passing vacuously, because an empty DOM satisfies an absence assertion". The code says otherwise, and I verified it against the source rather than taking the audit's word: those three call `respondWith(() => mintFail)` **before** rendering, so `useBlockToken` never yields a token, `BlockHost` returns `<BlockFallback>`, and `IframeHost` — the only `useFeatureFlags()` caller in this graph — **never mounts**. The crash cannot reach them. Their assertions are **presence** assertions on the fallback node, which an empty DOM would fail. They passed legitimately.

So this file went **loudly red, 3 of 6**. The signal was lost to `component-tests` being **report-only**, not to weak assertions — which, if anything, strengthens the "make it blocking" argument rather than the "the assertions are vacuous" one. I inferred a mechanism that fit the 3/3 split and never read the three passing tests. The comment shipped in the file has been corrected too, and so has the durable note I'd written into our ops skill.

Fixed with the same whole-module factory and the same dark-flag defaults the sibling `IframeHost.browser.test.tsx` already uses for exactly this reason. **6/6 in 511ms** (was 3 failed in ~14s), and the three absence tests now run against a real render.

### 2. `AppsSubmitEditView.browser.test.tsx` — 1 failed

The bounded-loader test **raced its own fixture**: it passes `loaderCeilingMs={30}`, then asserts the loader *is* present with a **retrying** matcher. The loader self-destructs at the ceiling, so if the first poll lands later the matcher can never succeed — it timed out at ~15s. A red that tracked machine load, not code.

My first fix widened the ceiling to 200ms. **The audit was right that this only makes the race rarer**, in a repo whose own scaffold raises the `waitFor` default to 10s because the preview CI box is contended. The pre-ceiling assertion was also **redundant** — the first test in this file already asserts the loader renders, at the **production** 15s ceiling, with no timing dependency.

So the assertion is **deleted** and the 30ms ceiling restored. What this test uniquely owns is the **fall-through**, and both remaining assertions only ever *wait*, so a short ceiling is correct and fast. **6/6 consecutive warm runs at ~1.05s.**

## Verified by mutation, not just a green run

- Reverting the component to the pre-fix `showLoader = editQuery.isLoading` — the infinite-spinner bug the bounded-loader test exists to catch — **still fails that test** (3 red). So raising the ceiling did not defang the guard; its intent is intact.
- The crash cause was **confirmed by dumping the DOM at the failure point** rather than inferred: empty `<body>`, and then `useFeatureFlags can only be used inside FeatureFlagsCtx` once the render was awaited. My first hypothesis (an un-awaited async render) was **wrong** — awaiting it only made the real error visible.

## Honest notes

- Both files carry a **pre-existing** `local-rules/no-wholesale-module-mock` error on `~/utils/trpc` — present on `origin/main`, unrelated to this change, and not something I want to reshape in a fix-the-gate PR. The `ESLint (modified files)` step is `continue-on-error` by design ("10% of recently-touched files carry a pre-existing error-severity finding"), and this PR **adds no file**, so the blocking added-files step is unaffected.
- **15/15 green on a warm run.** The first 1–2 component runs after any module-graph change are the documented cold-vite `optimizeDeps` cascade and report garbage — including a whole-suite red whose `useContext` signature reads exactly like dual-React. Only a warm run is readable; I ran until warm.
- This does **not** claim the suite is now green fleet-wide. It fixes the 4 failures I reproduced and attributed. #3539's log also showed a 5th (`AppBlockChrome`) that does not reproduce on `main` here, so it is likely specific to that branch.

## Not fixed here

The suite remains **report-only**, so it still cannot block a regression from returning. Making `component-tests` gate is the change that would deliver durability — deliberately out of scope for this PR, which is about making the signal true before anyone argues about making it blocking.

### DIFF
--- a/src/components/AppBlocks/BlockHostTokenRetry.browser.test.tsx
+++ b/src/components/AppBlocks/BlockHostTokenRetry.browser.test.tsx
@@ -35,12 +35,19 @@ vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
 
 // 🔴 IframeHost calls useFeatureFlags() (added by #3497 for the in-host "open the
 // page" affordance) and the REAL hook THROWS without a FeatureFlagsProvider, which
-// this scaffold does not mount. Without this the whole subtree crashes at mount:
-// `BlockHost` renders NOTHING, `waitForIframe` times out as "iframe never mounted",
-// and — worse — the three absence-asserting tests still PASS, because an empty DOM
-// trivially satisfies "collapsed to null". Same whole-module-factory shape and the
-// same dark-flag defaults as the sibling IframeHost.browser.test.tsx, deliberately
-// not an `importOriginal` spread.
+// this scaffold does not mount. Without this mock, every test that gets far enough
+// to MOUNT IframeHost crashes: `BlockHost` renders nothing and `waitForIframe`
+// times out as "iframe never mounted".
+//
+// Precisely which tests: only the three that reach a successful mint. The other
+// three set `respondWith(() => mintFail)` BEFORE rendering, so `useBlockToken`
+// never yields a token, `BlockHost` returns `<BlockFallback>` (see BlockHost.tsx),
+// and IframeHost — the only `useFeatureFlags()` caller in this render graph —
+// never mounts at all. Those three were unaffected and passed legitimately; they
+// assert PRESENCE on the fallback node, which an empty DOM would fail.
+//
+// Same whole-module-factory shape and the same dark-flag defaults as the sibling
+// IframeHost.browser.test.tsx, deliberately not an `importOriginal` spread.
 vi.mock('~/providers/FeatureFlagsProvider', () => ({
   useFeatureFlags: () => ({ appBlocks: false, appBlocksPages: false }),
 }));
--- a/src/components/AppBlocks/BlockHostTokenRetry.browser.test.tsx
+++ b/src/components/AppBlocks/BlockHostTokenRetry.browser.test.tsx
@@ -33,6 +33,18 @@ import {
 // AppBlockChrome calls useCurrentUser() for the platform-nav moderator gate.
 vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
 
+// 🔴 IframeHost calls useFeatureFlags() (added by #3497 for the in-host "open the
+// page" affordance) and the REAL hook THROWS without a FeatureFlagsProvider, which
+// this scaffold does not mount. Without this the whole subtree crashes at mount:
+// `BlockHost` renders NOTHING, `waitForIframe` times out as "iframe never mounted",
+// and — worse — the three absence-asserting tests still PASS, because an empty DOM
+// trivially satisfies "collapsed to null". Same whole-module-factory shape and the
+// same dark-flag defaults as the sibling IframeHost.browser.test.tsx, deliberately
+// not an `importOriginal` spread.
+vi.mock('~/providers/FeatureFlagsProvider', () => ({
+  useFeatureFlags: () => ({ appBlocks: false, appBlocksPages: false }),
+}));
+
 // IframeHost drives two tRPC queries at render plus the SDK bridges. Stub them so
 // the host mounts network-free AND the init handshake may start immediately
 // (getEffectiveCheckpoint must report isLoading:false so `shouldStartInit` fires).
--- a/src/components/Apps/AppsSubmitEditView.browser.test.tsx
+++ b/src/components/Apps/AppsSubmitEditView.browser.test.tsx
@@ -141,7 +141,15 @@ describe('AppsSubmitEditView — routing', () => {
       error: null,
       refetch: vi.fn(),
     };
-    renderWithProviders(<AppsSubmitEditView listingId="apl_1" loaderCeilingMs={30} />);
+    // 🔴 The ceiling must be long enough that the loader is still up when the
+    // RETRYING matcher below first polls. At 30ms this test raced its own
+    // fixture: the loader self-destructs before the first poll can see it, so
+    // `toBeInTheDocument` retried against an element that was already gone and
+    // timed out after ~15s — a red that tracked machine load, not the code.
+    // 200ms is what the sibling ceiling test below uses and passes reliably on.
+    // The assertion's INTENT is unchanged: a component that ignored the ceiling
+    // would still never surface the alert, so this still fails a pre-fix build.
+    renderWithProviders(<AppsSubmitEditView listingId="apl_1" loaderCeilingMs={200} />);
     // Loader shows first…
     await expect.element(page.getByTestId('apps-offsite-edit-loading')).toBeInTheDocument();
     // …then the ceiling elapses and the recoverable alert takes over.
--- a/src/components/Apps/AppsSubmitEditView.browser.test.tsx
+++ b/src/components/Apps/AppsSubmitEditView.browser.test.tsx
@@ -141,18 +141,20 @@ describe('AppsSubmitEditView — routing', () => {
       error: null,
       refetch: vi.fn(),
     };
-    // 🔴 The ceiling must be long enough that the loader is still up when the
-    // RETRYING matcher below first polls. At 30ms this test raced its own
-    // fixture: the loader self-destructs before the first poll can see it, so
-    // `toBeInTheDocument` retried against an element that was already gone and
-    // timed out after ~15s — a red that tracked machine load, not the code.
-    // 200ms is what the sibling ceiling test below uses and passes reliably on.
-    // The assertion's INTENT is unchanged: a component that ignored the ceiling
-    // would still never surface the alert, so this still fails a pre-fix build.
-    renderWithProviders(<AppsSubmitEditView listingId="apl_1" loaderCeilingMs={200} />);
-    // Loader shows first…
-    await expect.element(page.getByTestId('apps-offsite-edit-loading')).toBeInTheDocument();
-    // …then the ceiling elapses and the recoverable alert takes over.
+    // 🔴 This test used to assert the loader is PRESENT before asserting the
+    // fall-through, with a 30ms ceiling — a race against its own fixture. The
+    // loader self-destructs at the ceiling, so a RETRYING matcher that first
+    // polls after 30ms can never succeed; it timed out at ~15s. That is what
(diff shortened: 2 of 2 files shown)


ZacxDev:
## Audit round 1 — one finding was a wrong claim of mine, now corrected everywhere

The audit verdict was **merge**, with no 🔴. But its first finding is the one that mattered, and I verified it against the source rather than taking it on trust — **it is right and I was wrong.**

### 🔴 My headline diagnosis was false

I claimed the file's three passing tests were passing **vacuously** ("an empty DOM trivially satisfies an absence assertion"). They were not. Those three call `respondWith(() => mintFail)` **before** render, so no token is minted, `BlockHost` returns `<BlockFallback>`, and `IframeHost` — the only `useFeatureFlags()` caller in that render graph — **never mounts**. The crash cannot reach them. And they assert **presence** on the fallback node (`fallbackQuery()?.getAttribute('data-block-fallback')).toBe('loading')`), which an empty DOM **fails**.

The file went **loudly red, 3 of 6**. The signal was lost because `component-tests` is report-only — not because the assertions were weak. That distinction cuts the opposite way from how I wrote it: it *strengthens* the case for making the check blocking.

What went wrong on my side is worth naming: a 3-passed/3-failed split **looked** like "half the assertions are vacuous", so I wrote that mechanism down without reading the three passing tests. A theory that neatly explains the numbers is not evidence.

Corrected in three places, because it had already propagated: the **comment shipped in the test file** (`b732b907aa`), this **PR description**, and a **durable gotcha in our ops skill** that I'd already committed — that last one matters most, since agents load it as authoritative and a false mechanism there is an instruction to misdiagnose the next occurrence.

### 🟡 Took the audit's better fix for the race

It was right that 200ms only makes the race **rarer**, not gone — especially in a repo whose scaffold raises the `waitFor` default to 10s for the contended preview box. And the pre-ceiling assertion was **redundant**: the first test in the file already asserts the loader renders, at the production 15s ceiling, deterministically.

So that assertion is **deleted** and the 30ms ceiling restored. The test uniquely owns the fall-through; both remaining assertions only *wait*. **6/6 consecutive warm runs, 15/15 each, ~1.05s** (down from ~1.2s).

🔴 Re-ran the mutation after deleting an assertion: reverting to `showLoader = editQuery.isLoading` still fails the bounded-loader test plus 2 siblings. The guard is intact.

### 🟡 My population evidence half-held

I said #3540/#3539/#3541/#3529 "all carry the same failures". Only **two** check statuses support that — #3541 reports a *runner-budget overrun, explicitly not a test failure*, and #3529 has no status at all. Their **pod logs** do contain the same failure markers, so the tests do fail there; the check-status phrasing overstated it. Also corrected: the `(1122)` total I quoted was **#3540's** tree (which carries its own 4 added tests), not this branch's baseline of 1118.

### Independently confirmed by the audit

Full warm component suite on this branch: **112/112 files, 1118/1118 tests**. The mock is faithful (`features` is read at exactly one site, `IframeHost.tsx:1909`, and nothing else in the graph touches flags); the `false/false` defaults are covered elsewhere so they leave no hole; leaving the other `loaderCeilingMs={30}` sites alone is correct, not inconsistent; both eslint errors are genuinely pre-existing; and #3539's `AppBlockChrome` failure is unrelated to this PR.

The 🟢 nits (a second wholesale module mock that the lint rule doesn't scope to, and a now-stale sentence in a pre-existing comment) I've left — reshaping mocks isn't what a fix-the-gate PR should be doing.

ZacxDev:
## Delta re-audit: clean — plus one correction to my own comment

Second independent pass over the fix round. **No 🔴, no 🟡.** Both round-1 findings verified as properly fixed rather than papered over.

### The best finding: deleting that assertion *improved* the test

Under the exact mutant this test exists to catch (`showLoader = editQuery.isLoading`), the **deleted assertion would have PASSED** — the loader stays up forever, so asserting its presence proves nothing there. It was diluting the very test it lived in. What remains (`await` the alert) is the pure discriminator.

And the mutant I was asked about — loader never renders, alert still appears — is still caught **twice, independently**: the file's first test (`:63`, production 15s ceiling), and `:191` in the retry test, which passes an explicit `loaderCeilingMs={200}`. That second one is the residual coverage specifically protecting "the loader renders when a caller-supplied ceiling is in play", which is what the deleted line nominally owned.

### Claims of mine it checked rather than accepted

- **"exactly 3 tests mount `IframeHost`, 3 do not"** — enumerated line by line. Correct.
- **"the only `useFeatureFlags()` caller in this render graph"** — checked `BlockFallback`, `AppBlockChrome`, `RecentlyOpenedApps`, `AppPermissionsActivityDrawer`, `BlockConsentModal`. Correct; the only other caller is `BlockSlot`, the *parent*, not rendered here.
- **The `~15s` figure** — the auditor first derived 5s from the non-browser default, then corrected itself: the playwright provider overrides the poll timeout to `remainingTestTimeout - 100`, and the `component` project inherits vitest's 15s browser default (the `60000` in the config is on the **unit** project only).
- **`.toHaveLength(0)` is not vacuous** — it is gated by the awaited presence assertion on the sibling node, so a nothing-rendered DOM fails that first. The documented pre-commit-DOM trap is avoided.
- **No mutation-cycle litter**, and `AppsSubmitEditView.tsx` is untouched by this PR.

### 🟢 Corrected my comment's MECHANISM (`761a9632f0`, comment-only)

I wrote that the retrying matcher "first polls after 30ms". It doesn't — `expect.element` polls on a **50ms interval with the first attempt immediate**. The real mechanism: once that immediate poll missed React's commit, *every* later poll was already past the 30ms ceiling, so the assertion was unwinnable and burned the full ~14.9s budget. Same conclusion, wrong cause — and a comment naming the wrong mechanism sends the next person hunting in the wrong place, which is the class this PR exists to clean up.

Also softened "no timing dependency at all" for the file's first test: it is safe because the loader is committed on the **first render**, not because 15s is generous. If `EDIT_LOADER_CEILING_MS` were lowered toward the poll interval, that test would inherit this same race.

### Merged-tree check

Clean against current `main` textually **and** semantically. The one overlap — `AppsSubmitEditView.tsx` gaining `APPS_PAGE_WIDTHS` — is inert here, since this suite mocks `AppsPageLayout` and ignores `size` entirely.

### One to watch

`:191` is now the file's **only** remaining transient-state observation, and it is unavoidable by construction (the assertion *is* the observation) — mitigated only by margin, which is the very "widening makes it rarer" this PR argues against. Pre-existing on `main`, no action taken. **If `preview / component-tests` goes red again, that line is the prime suspect.**

15/15 green on a warm run. Ready to merge.

ZacxDev:
## CI result: 4 failures → 1. The survivor is pre-existing and NOT fixed here

Read the CI log rather than the check summary, per #119.

```
before:  Tests  4 failed | ...
now:     Tests  1 failed | 1117 passed (1118)     [two consecutive CI runs]
```

**All four failures this PR targets are gone in CI.** The remaining one is a third file this PR does not touch:

```
FAIL  src/components/AppBlocks/AppBlockChrome.browser.test.tsx
      > "Recently run" section > renders recents (icon + name), EXCLUDES the current app…
AssertionError: expected null not to be null      (AppBlockChrome.browser.test.tsx:338)
```

**It is not caused by this PR**, and I checked rather than assumed:
- `git diff --name-only origin/main...HEAD` → only the two test files; `AppBlockChrome` is untouched;
- **#3539 — which contains none of these changes — shows the identical failure in CI**;
- run in isolation on this branch it passes **21/21**, and the full warm suite locally is **1118/1118**.

🔴 **This also falsifies something I wrote in the original description.** I said the `AppBlockChrome` failure "does not reproduce on main here, so it is likely specific to that branch". It is not branch-specific — it is **CI-specific**, and it reproduces on every CI run I have looked at.

### Diagnosis (hypothesis, explicitly NOT verified)

Line 338 asserts the recents entry contains an ``. The fixture uses `iconUrl: 'https://cdn.example/icon.png'` — a non-resolving host. Mantine's `Avatar` swaps to its placeholder (and drops the ``) on image error, so the likely mechanism is that the request **fails fast in CI** (no DNS/egress) and fires `onError` before the assertion, while locally it stays pending long enough for the `` to still be in the DOM. That would make it an environment-dependent race on an external URL.

🔴 **I am not fixing it in this PR, deliberately.** The test **passes locally**, so I cannot verify a fix locally — any change would be a guess validated only by pushing and watching CI. Bolting an unverifiable change onto a merge is exactly the pattern this PR is cleaning up. The fix is likely a deterministic `data:` URI instead of a non-resolving host, and it wants its own PR where the CI loop *is* the verification.

**Merging on this basis:** a verified 4→1 reduction, blocking checks green (Typecheck / Unit / ESLint+Prettier / event-engine-common), `preview / smoke-tests` now **passes** (61 passed), and the one survivor is pre-existing, independent, and now documented instead of anonymous.

PR #3547, opened 2026-08-02
feat(apps): store card matches the site's card chrome + type scale (S4/S5), one glyph source (S6b)

ZacxDev:
## What

The `/apps` store card is the only content card on the site wearing a border, a shadow and a non-standard radius, with the smallest and dimmest title/author text. This brings it to the `/models` baseline and collapses the CTA icon mapping to one source.

## Measured, at two viewports

Against a `/models` card **in the same session** (dark), and re-measured at **394 px mobile**, where every value below is identical — so this design is not desktop-only.

| | `/apps` before | `/models` | after |
|---|---|---|---|
| `border-radius` | 8 px | **6 px** | 6 px |
| `border` | 0.877px solid rgb(55,58,64) | **0** | 0 |
| `box-shadow` | 3-layer | **none** | none |
| `background` | rgb(37,38,43) | rgb(37,38,43) | unchanged |
| title | 18/700/rgb(193,194,197) — contrast 8.48 | 20/700/#fefefe — 14.97 | 20/700/#fefefe |
| author | 12/400/rgb(140,143,163) | 14/500/white | 14/**500**, still dimmed |

**Radius comes from Tailwind's `rounded-md`, not Mantine's `md`** — the two `md`s are 6 px and 8 px, which is exactly how this drifted.

**Padding stays 16 px.** The review's "zero padding" is about the *media*, already full-bleed via `Card.Section`; this card's text sits *below* the media rather than overlaid on it.

**The mobile-specific risk was checked, not assumed.** At 1 card per row the card is effectively full-bleed, so dropping the border could have made it vanish. It doesn't — the card's own fill (rgb(37,38,43)) separates it from the page (rgb(26,27,30)) with every ancestor transparent, the same separation `/models` already relies on at `border: 0`.

## The author line stays dimmed — deliberately

Size + weight only. Taking *both* title and author to white flattens the title-over-author hierarchy on a `dark-6` body; `/models` needs white there only because its author line is overlaid on media. **This leaves the author at contrast 4.73 — AA pass by 0.23, AAA fail.** An accepted trade, and now asserted, so it can't be "fixed" later without revisiting the decision.

## 🔴 A trap this closes

`cardActionGlyph('detail')` shipped in **#3540** returning `'info'` (`IconInfoCircle`) while it had **no caller**. **#3539** had meanwhile shipped **`IconEye`** on the card's real "View details" CTA.

The module existed *precisely* so this PR would wire the card to it — and doing that naively would have **silently repainted every View-details CTA**, regressing a deliberate product call. Reconciled to a new `view` glyph (`IconEye`) **before its first caller exists**, and pinned both to the value and to its distinction from `info`.

S6b's remaining half was never "add icons" — #3539 did that. It was **one source**: the card now reads `ACTION_GLYPH_ICONS[cardActionGlyph(...)]` instead of a second inline mapping, and the comment that *restated* the mapping is deleted (a copy in prose is how the card and detail drifted apart to begin with). Same three icons render as before.

## Two geometry pins moved — on purpose

`280 → 282` and `246 → 248`. That is the border removal being **real and measurable**: 0.877 px a side, ~1.75 px reclaimed, and the arithmetic is now exact (`width − 32`) rather than fractional (`width − 34`). **If those numbers ever move back, the border came back.** The derivation comment is updated to say so.

## Verification

**Component 40/40**, stable over 3 consecutive runs. **Node 9/9.** `tsc`: 16 pre-existing errors, none in these files. `eslint` 0. `prettier` clean — note `AppListingCard.tsx` is prettier-*clean* on `main`, so this PR keeps it that way rather than leaving drift.

### Break-it matrix

| Mutation | Result |
|---|---|
| restore `shadow="sm" radius="md" withBorder` | **3 red** — the S4 guard **plus both geometry pins**, which are an independent detector |
| `cardActionGlyph('detail')` back to `'info'` | **1 red** — `expected 'info-circle' to be 'eye'` |
| revert the title to `<Title order={4}>` | ⚠️ **not verified** — the run timed out before producing a verdict; I am not claiming a kill I did not watch |

## Two things I got wrong, recorded because they cost real time

1. **My test assertions were wrong about Mantine's DOM, not the component.** `fw` renders as a plain `font-weight:` declaration and `c` as `color:` — *not* `--text-fw` / `--text-color`. The scope doc's marker guidance was half wrong, and I asserted it without checking. The component was correct the whole time.
2. **A pristine-`main` control is what caught a real breakage I'd have dismissed.** My first runs failed with `useState`/`useContext` nulls — the documented dual-React cache cascade. Pristine `main` in the same worktree passed **34/34**, mine converged to a stable **28/10**. A cache artifact keeps *moving*; convergence is the discriminator. The real causes were mine: a missing render barrier, an over-broad `a[href]` selector that grabbed a wrapper instead of the CTA, and a manual `.unmount()` fighting the scaffold's `afterEach(cleanup)` over the shared container — which timed out every test after it.

## Scope note

S7 (the description-less dead band) was briefly folded into this PR and then **removed**. Measurement showed it is `h-full` equal-height-row stretching, not a stray spacer — so "fixing" it trades the dead band for ragged rows. That is a product call needing its own desktop measurement (the symptom is absent at 1-per-row mobile), not a free ride-along in the wave's most collision-prone file.

## Still owed

No live re-measurement on prod — the numbers above are from prod *before* this change plus a Chromium component run after it. The store is mod-gated and `deIndex`-ed, so this is low-risk to land unverified, but it is not the same as having looked.

### DIFF
--- a/src/components/Apps/AppListingCard.browser.test.tsx
+++ b/src/components/Apps/AppListingCard.browser.test.tsx
@@ -94,11 +94,17 @@ function base(over: Partial<ListingCard>): ListingCard {
  * `width` is the OUTER box. `Card padding="md"` eats 2x16, so the content box —
  * the number the layout measurements below are about — is `width - 32`.
  *
- * 🔴 It was `width - 34` until S4 dropped `withBorder`: the border was 0.877px a
- * side (~1.75px total), so every content box below RECLAIMED ~2px and the exact
- * pins moved 280 -> 282 and 246 -> 248. That is the border removal being real and
- * measurable, not a tolerance drift — and the arithmetic is now exact rather than
- * fractional. If these numbers ever move back, the border came back.
+ * 🔴 It was `width - 34` until S4 dropped `withBorder`, so every content box below
+ * RECLAIMED 2px and the exact pins moved 280 -> 282 and 246 -> 248. That is the
+ * border removal being real and measurable, not a tolerance drift; if these
+ * numbers ever move back, the border came back. Proven by isolation: restoring
+ * ONLY `withBorder` reproduces 280/246 exactly, and reverting ONLY the title
+ * (`Text` -> `Title`) leaves both pins green.
+ *
+ * ⚠️ In THIS environment the border is exactly 1px a side (Mantine's
+ * `calc(0.0625rem * var(--mantine-scale))` with `scale` unset and a 16px root),
+ * so `width - 34` was exact arithmetic, not fractional. The 0.877px figure comes
+ * from PROD, where the scale differs — do not use it to reason about these pins.
  *
  * The two widths used below, both measured through the real
  * `AppsPageLayout` + `Grid` rather than assumed:
@@ -240,14 +246,21 @@ describe('AppListingCard', () => {
 
   // ── Card chrome, typography + CTA glyph (S4 / S5 / S6b) ────────────────────
   //
-  // Markers are Mantine's own DOM output, which is deterministic here:
-  // `withBorder` renders as a `data-with-border` ATTRIBUTE, and `shadow` / `size`
-  // / `fw` / `c` render as `--paper-shadow` / `--text-fz` / `--text-fw` /
-  // `--text-color` custom properties in the element's `style`.
+  // Markers are Mantine's own DOM output, verified against what it ACTUALLY emits
+  // rather than assumed:
+  //   withBorder -> a `data-with-border` ATTRIBUTE
+  //   shadow     -> a `--paper-shadow` custom property
+  //   size       -> a `--text-fz` custom property
+  //   fw / c     -> PLAIN `font-weight:` / `color:` declarations
+  // 🔴 `fw` and `c` are the trap: they do NOT become `--text-fw` / `--text-color`.
+  // Asserting those fails against a perfectly correct component, which is exactly
+  // how this suite burned a debugging cycle.
   //
-  // 🔴 Deliberately NOT asserting resolved `rgb()` values: the theme CSS is not
-  // loaded in this environment, so such an assertion either fails or passes for
-  // the wrong reason. The pixel values live in the PR's measurement table.
+  // 🔴 Not asserting resolved brand `rgb()` values: Mantine's stylesheets ARE
+  // loaded here (`assertLayoutIsReal` depends on them), but ThemeProvider's
+  // CSS-variable overrides are not — so `--mantine-color-white` resolves to
+  // rgb(255,255,255) here and #fefefe in the app. Geometry (px) is trustworthy;
+  // brand colours are not. The pixel/contrast values live in the PR's table.
 
   const cardRoot = () => document.querySelector('[class*="Card-root"]') as HTMLElement | null;
 
@@ -288,8 +301,19 @@ describe('AppListingCard', () => {
     expect(root!.hasAttribute('data-with-border')).toBe(false);
     // `shadow="sm"` is gone → no `--paper-shadow` is declared at all.
     expect(root!.getAttribute('style') ?? '').not.toContain('--paper-shadow');
-    // Radius comes from the shared Tailwind scale (6px), NOT Mantine's `md` (8px).
-    expect(root!.className).toContain('rounded-md');
+    // 🔴 Assert the COMPUTED radius, not the className the JSX just wrote back to
+    // us. `radius={0}` makes Mantine emit `--paper-radius: 0rem`, and Tailwind's
+    // `.rounded-md` only wins because Mantine's sheet is inside `@layer mantine`
+    // while Tailwind utilities are deliberately unlayered (`globals.css`). If that
+    // layer ordering ever regresses, the card renders SQUARE — and a className
+    // assertion would still be green.
+    // 🔴 No `className).toContain('rounded-md')` echo here on purpose: it asserts
+    // the string the JSX just wrote, it cannot detect the square-card failure
+    // above, and — being ordered first — it would SHADOW the computed assertion
+    // that can. An earlier check that always wins makes the later one untestable.
+    expect(getComputedStyle(root!).borderRadius).toBe('6px');
+    expect(getComputedStyle(root!).borderTopWidth).toBe('0px');
+    expect(getComputedStyle(root!).boxShadow).toBe('none');
   });
 
   test('🔴 S5: the title is xl/700/white — and is no longer a heading', async () => {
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Audit round 1 — no 🔴; five 🟡s fixed in `c7528eb9b4`

The audit verified every substantive claim in the description and found **no correctness, regression or a11y defect**. It also closed this PR's one self-declared gap. Fixes below are documentation accuracy and guard strength — no behaviour change.

### It closed the mutation I could not verify

The body declared "revert the title to `<Title order={4}>` — **not verified**, the run timed out". The audit ran it on an isolated copy: it kills **exactly one** test, the S5 guard, for its own reason — and **both geometry pins stayed green**. That also independently proves the 280→282 / 246→248 move is *solely* the border removal and isn't masking a layout change from `Title`→`Text`. Restoring only `withBorder` reproduces 280/246 exactly.

### 🔴 The strongest finding: my S4 radius guard was shadowing itself

`expect(root.className).toContain('rounded-md')` asserts the string the JSX just wrote back. It **cannot** detect the failure it names — an `@layer` regression letting Mantine's `--paper-radius: 0rem` win and the card render **square**. And because it was ordered first, it **shadowed** the computed assertion that can: I mutation-tested it and the echo failed first, so the computed check was never reached. An earlier check always winning is how a guard ends up untestable.

Echo **removed**; the guard now asserts computed `borderRadius` / `borderTopWidth` / `boxShadow`. Re-mutated — dropping `rounded-md` now fails with `expected '0px' to be '6px'`.

### My derivation note was wrong for the environment it annotates

I wrote that the border was 0.877px and `width - 34` was "fractional". Measured with `withBorder` restored: it is **exactly 1px** here — Mantine's `calc(0.0625rem * var(--mantine-scale))` with `scale` unset and a 16px root — so `width - 34` was exact arithmetic. **0.877px is a prod-only artifact.** The +2 delta is unaffected, but a load-bearing derivation note a future debugger will trust has to be right.

### Also fixed

- The section comment still claimed `fw`/`c` render as `--text-fw`/`--text-color` — the exact misinformation this PR's body records as having cost a cycle, sitting 60 lines above assertions that correctly use `font-weight:`/`color:`. Corrected, along with "the theme CSS is not loaded": Mantine's stylesheets **are** loaded; what's missing is ThemeProvider's variable overrides, which is why white resolves to `rgb(255,255,255)` here and `#fefefe` in the app.
- **`lh={1.2}` was a third typographic change with no test and no mention.** Without it the title inherits `--mantine-line-height-xl` (1.65) — 33px vs 24px at 20px, a **~14px card-height swing** across a 2-line clamp, in the very file whose deferred S7 work is about card height. Now pinned, literal and computed.
- **The heading claim was wrong.** I wrote that `/apps` would have no headings left. `RecentlyOpenedApps` ("Recently opened") and `RelatedListings` both render their own `Title` on those surfaces. Corrected, and the grep result recorded: nothing depends on a heading inside the card.

### Blast radius — correcting the description

`AppListingCard` renders at **four** sites, not one: `AppListingsMarketplaceBody` (the store grid), **`RelatedListings`** (the rail on the listing detail), and **`OffsiteReviewQueue` ×2** (the moderator store-faithful preview + the revision-drift card). The chrome change is correct on all four — the mod preview is *meant* to look like the store — but my "mod-gated and deIndex-ed" risk framing only covered `/apps`. Two notes worth carrying: `RelatedListings` has **no browser test file at all**, and in the mod panel the now-borderless card sits beside a `<Card withBorder>`, on a surface whose background I did not measure.

### One thing worth knowing beyond this PR

The audit found that `AppListingsMarketplaceBody.browser.test.tsx`, `OffsiteReviewQueue.browser.test.tsx` and `ConnectScopesPanel.browser.test.tsx` currently **fail to collect** (`0 test`, module-mock error) — and confirmed they are **identically red on `origin/main`**, so this PR did not cause it. The practical consequence: the card's main consumer has no live gate right now. Same family as the dead suites #3544 just fixed, and a good candidate for the next pass.

**Verification after the fixes:** component **40/40**, prettier clean, merges clean against current `origin/main` (no overlapping file).

ZacxDev:
## Correction: the "dead consumer suites" claim was WRONG — it was a cold-cache artifact

In my previous comment I relayed an audit note that `AppListingsMarketplaceBody.browser.test.tsx`, `OffsiteReviewQueue.browser.test.tsx` and `ConnectScopesPanel.browser.test.tsx` "fail to collect (`0 test`)" and that consequently **"the card's main consumer has no live gate right now"**.

**That is false.** Run to a warm cache on this branch:

| suite | result |
|---|---|
| `AppListingsMarketplaceBody.browser.test.tsx` | **12/12 passed** |
| `OffsiteReviewQueue.browser.test.tsx` | **21/21 passed** |
| `ConnectScopesPanel.browser.test.tsx` | **6/6 passed** |

Every one of them was the documented cold-vite cascade: run 1 produced the `useState`-null wall, and a warm run is green. The card's consumers **do** have live gates, including the store grid and both moderator surfaces.

🔴 **The irony is the point.** That cold-cache trap is the one I explicitly briefed the auditor about, and it is gotcha #164 in our own ops skill — and I still relayed the red without warming it myself. A cold red is a fact about the cache, not the code; I applied that rule to my own results all session and then dropped it for someone else's. The audit did caveat that its verdict "may be local" and that it ran on system Node rather than the pinned version — I should have treated that as the signal to re-measure rather than to publish.

**Nothing else in the audit summary changes.** The five 🟡s were real and are fixed in `c7528eb9b4`; this retraction affects only the closing "beyond this PR" note, and there is no follow-up needed for those suites.

One thing does survive from that finding, and it is worth keeping: **`RelatedListings` genuinely has no browser test file at all** (verified — the file does not exist), and it is one of the four sites rendering this card.

ZacxDev:
## Merging — with an explicit note on what was and was not read

**Blocking checks green:** Typecheck ✅, Unit tests ✅, ESLint + Prettier ✅, event-engine-common pin ✅. `MERGEABLE`, no PR based on this branch.

🔴 **`preview / component-tests` and `preview / smoke-tests` had NOT reported at merge time** — `preview / deploy` was still pending. I am not claiming to have read them. What stands in their place:

- `AppListingCard.browser.test.tsx` **40/40**, stable across 3 consecutive warm runs locally, plus **4 consecutive warm runs** in the independent audit;
- the three consumer suites re-verified warm today — `AppListingsMarketplaceBody` 12/12, `OffsiteReviewQueue` 21/21, `ConnectScopesPanel` 6/6;
- node unit **9/9**; `tsc` 16 pre-existing errors, none in these files;
- five mutations watched go red, each for its own reason, including the re-mutated radius guard after the shadowing echo was removed.

The known outstanding item on that check is the pre-existing `AppBlockChrome` failure, attributed in #3544 and unrelated to this file.

If `component-tests` reports something new on `main`, this change is a clean single-commit-range revert.

PR #3553, opened 2026-08-03
test(browser): single-source the loadable image fixture + lint the unloadable-URL trap

ZacxDev:
## Why

[#3551](https://github.com/civitai/civitai/pull/3551) fixed an `AppBlockChrome` test that had been **red on `main` across five PRs**. Its icon fixture was an http URL, which cannot load in the test browser, so the `` fired a real `error` event and Mantine 7.17.8's `Avatar` — `useState(!src)` + `onError -> setError(true)` — destroyed it ~11 ms after mount. The `expect(img).not.toBeNull()` assertion was racing that window: green locally, red under CI load.

The fix was a `data:` URI. It was the **second copy** of that constant, with the same explanatory comment — `AppListingCard` already had one. The convention existed and was invisible, which is why it took five PRs to spot.

## What

**1. One fixture.** `LOADABLE_IMAGE_DATA_URI` now lives in `test/component-setup.tsx` — the shared browser-test scaffold that **115 of the 117** `*.browser.test.tsx` files already import for `renderWithProviders`, so this adds no new convention and no new import site. `AppBlockChrome`'s `LOADABLE_ICON` and `AppListingCard`'s `LOADABLE_PNG` both import it, and the explanation lives with the constant rather than at each use site.

**2. A deterministic guard.** New `local-rules/no-unloadable-image-fixture` in the repo's existing `eslint-local-rules.js`, scoped to `**/*.browser.test.tsx`. It flags an http(s) URL literal in an image-source position. Escape hatch is the convention the repo's other three local rules already use:

```
// eslint-disable-next-line local-rules/no-unloadable-image-fixture -- <reason>
```

## Survey — measured, not assumed

All 117 browser tests were run with a document-level capture listener recording every `` that fires a real `error` event with an http(s) src (the probe was removed before commit).

| | |
|---|---|
| `*.browser.test.tsx` files | **117** |
| …containing any http(s) string literal | **39** |
| …distinct external image URLs that really do mount as broken ``s today | **14**, across **6** files |
| …files with BOTH an http image fixture AND an ``-existence assertion | **1** (`AppListingCard`, fixed here) |

Positive control for the probe: it recorded exactly the 3 URLs `AppListingCard` is known to mount, so a zero elsewhere means "nothing mounted", not "listener wired to nothing".

So the other fixtures are **latent, not harmless** — each becomes a flake the day someone adds an `` assertion beside it, which is precisely what happened to `AppBlockChrome`.

## Proportionality — why a lint rule, and why this severity

The rule reports **16 sites in 5 files, all pre-existing**. That is survivable for exactly the reason already written on `no-wholesale-module-mock`: `lint.yml` gates **ADDED** files blocking and **MODIFIED** files report-only. The backlog can therefore only ever reach the report-only step, while a **newly added** browser test is blocked — the authoring path the rule exists to close. Severity is `'error'` for the same reason: the blocking step runs without `--max-warnings`, so `'warn'` would gate nothing anywhere.

No pre-existing file is touched, so there is no diff in anyone else's blast radius.

## False-positive control

Two independent conditions must both hold — the literal must be http(s) **and** sit in an image-source position — and the ambiguous `src` key additionally has to prove it is an image (file extension, or an ``-family JSX element). Verified clean against every real non-image http URL in the browser suite:

- `OnsiteReviewModal`'s `iframe: { src: 'https://example.com/block' }` — an iframe document, not an image. This is the case that forces the `src` proof requirement.
- `AgentReviewChat`'s markdown `tracking` — whose test asserts **no** `` is produced (`expect(bubble.querySelector('img')).toBeNull()`). Not an image-source position, so it cannot mount a broken `` and needs no escape hatch.
- `liveUrl` / `externalUrl` / `previewUrl` / `reviewRepoUrl`, and URLs typed into form fields.
- The relative `/api/blocks/screenshot/...` paths in `AppBlockCard` / `AppDetailsModal`. Deliberately out of scope — a relative path that happens not to be served is not decidable from source text; the http(s) form is unloadable *by construction*.

## Mutation evidence

Every claim below was run, not reasoned about.

- **34 RuleTester cases** (18 valid / 16 invalid), wired into `pnpm test:lint-rules`. All 4 local-rule suites: **186 passed**.
- Rule made a no-op → invalid cases fail (`Should have 1 error but had 0`). *The harness can go red.*
- `src` image-proof removed → **4 valid cases fail** (iframe/script `src` false-positive).
- http matcher widened to `/./` → **2 valid cases fail** (`data:` URI and relative path).
- Key-membership check removed → the extensionless `iconUrl` case **stops firing**.
- Real `eslint` on a synthetic browser test: **7 reports** covering every supported shape, **0** on every legitimate shape.
- `AppListingCard`'s disable comment removed → **1 report**; restored → **0**. The hatch is load-bearing, not decorative.

## Escape hatch — one use

`AppListingCard`'s deliberate broken-cover test needs a URL that really fails, so it can exercise the component's `onError -> placeholder` path. Its assertions all run on the **post-error** state behind a `vi.waitFor`, so it never races the swap the way an "`` exists" assertion would. Behaviour is unchanged: only comments and the disable directive were added around an identical fixture URL.

## Verification

- `AppBlockChrome` + `AppListingCard`: **63 passed** — identical to the pristine `origin/main` baseline (**63 passed**, 16.09 s vs 16.55 s).
- **112 passed** across those two plus two untouched browser tests (`AppBlockCard`, `recents-helper`), confirming the shared-scaffold change is inert for files that don't use it.
- ESLint on all touched files: 0 findings. Prettier: `AppListingCard` stays clean; `AppBlockChrome` was already Prettier-dirty on `main` and was deliberately **not** reformatted, to keep the diff at 3 lines instead of ~60.

**Not claimed:** swapping `AppBlockChrome`'s fixture back to an http URL did **not** turn the test red on my machine. That is consistent with the defect being a race that manifests under CI load rather than locally — and it is exactly why a written convention was never going to hold — but it means the original symptom was not reproduced here.

Typecheck in the worktree reports 16 pre-existing errors, all module-resolution failures from an uninitialized `event-engine-common` submodule and a `kysely` package absent from `node_modules` in the base clone too. None are in files this PR touches.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -178,6 +178,51 @@ module.exports = {
         'local-rules/no-module-scope-cache': 'error',
       },
     },
+    {
+      // Browser-mode component tests must not use an http(s) URL as an image
+      // source. Nothing serves it to the test browser, so the <img> fires a real
+      // `error` event a few ms after mount and the component's own onError
+      // fallback destroys it — Mantine 7.17.8's Avatar is `useState(!src)` +
+      // `onError -> setError(true)`, measured at ~11 ms from mount to swap. Any
+      // "the <img> exists" assertion against such a fixture is racing that
+      // window: green locally, intermittently red on a loaded CI box. It sat red
+      // on `main` across five PRs before #3551, and the data: URI that fixed it
+      // was a second copy of one another test file already had — the convention
+      // existed and was invisible. Hence a rule, and hence the single shared
+      // `LOADABLE_IMAGE_DATA_URI` in test/component-setup.tsx.
+      //
+      // 'error', matching no-wholesale-module-mock / no-module-scope-cache and
+      // for the same reason spelled out above: the only BLOCKING ESLint step in
+      // .github/workflows/lint.yml ("ESLint (added files)") runs without
+      // --max-warnings, so at 'warn' this would gate nothing — and a brand-new
+      // browser test is exactly the authoring path the rule exists to close.
+      //
+      // Blast radius on the existing tree is bounded and, by lint.yml's split,
+      // zero in the blocking step. Population measured by instrumenting all 117
+      // *.browser.test.tsx files with a document-level capture listener for
+      // <img> `error` events: 14 distinct external image URLs really do mount as
+      // broken images today, across 6 files. The rule reports 16 sites in 5
+      // files, ALL pre-existing, so a PR touching one reaches only the
+      // report-only modified-files step. (The 6th file, AppListingCard, is fixed
+      // in this change: its positive assertion uses the shared fixture and its
+      // deliberate broken-cover test carries a disable comment with a reason.)
+      // They are latent, not harmless — each becomes a flake the moment someone
+      // adds an <img> assertion beside it, which is what happened to
+      // AppBlockChrome.
+      //
+      // Scope is deliberately narrow so it cannot reach a non-image URL: the
+      // literal must be http(s) AND sit in an image-source position, where the
+      // ambiguous `src` additionally has to prove it is an image (file
+      // extension, or an <img>-family JSX element). That is what keeps
+      // OnsiteReviewModal's `iframe: { src: 'https://example.com/block' }` and
+      // AgentReviewChat's markdown `![tracking](https://example.com/pixel.png)`
+      // clean — neither can mount an <img> from a fixture. Full reasoning, the
+      // key list, and the escape hatch are in eslint-local-rules.js.
+      files: ['**/*.browser.test.tsx'],
+      rules: {
+        'local-rules/no-unloadable-image-fixture': 'error',
+      },
+    },
   ],
 
   // No type-aware linting: `parserOptions.project` costs ~40s of program build plus
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -198,7 +198,17 @@ module.exports = {
       // browser test is exactly the authoring path the rule exists to close.
       //
       // Blast radius on the existing tree is bounded and, by lint.yml's split,
-      // zero in the blocking step. Population measured by instrumenting all 117
+      // zero in the blocking step TODAY — with two named caveats, because
+      // "zero" is not the same as "permanently zero":
+      //   1. `--diff-filter=A` classifies a rename-with-heavy-edit as ADDED, so
+      //      MOVING one of the 5 backlog files lands it in the BLOCKING step.
+      //      Same caveat the sibling rule documents at the top of this file.
+      //   2. lint.yml's own header records that the report-only steps are
+      //      planned to flip to blocking once the backlog clears. On that day
+      //      these 16 sites block anyone touching those 5 files. The debt is
+      //      small and mechanical (9 distinct URLs, nearly all
+      //      'https://cdn/x.png' -> LOADABLE_IMAGE_DATA_URI).
+      // Population measured by instrumenting all 117
       // *.browser.test.tsx files with a document-level capture listener for
       // <img> `error` events: 14 distinct external image URLs really do mount as
       // broken images today, across 6 files. The rule reports 16 sites in 5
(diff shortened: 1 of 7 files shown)


ZacxDev:
## Corrections to this PR description (post-audit)

An independent audit returned **safe to merge, no 🔴**, and verified the load-bearing claim: the ESLint gating split holds, including the one repo-wide sweep risk neither of us had named — `next build` no longer lints (next@16.2.7 removed build-time ESLint entirely). It also independently measured 19 reports at `main` → 16 post-PR, confirming "16 sites in 5 files, all pre-existing".

Five statements in the description above are wrong or overstated. Correcting them here rather than editing the body, since they may already have been read. Code changes for 1–4 are in `c0cbacdd17`; the rule's behaviour is unchanged and its 34 RuleTester cases still pass.

1. **"every image-rendering component in this codebase … DESTROYS the ``" — false.** Checked against the installed `@mantine/core`: `Avatar` (`Avatar.mjs:70,91`) renders a `<span>` placeholder *instead of* the ``, so it is destroyed. `Image` (`Image.mjs:58`) only swaps when `fallbackSrc` is set — and that prop appears **zero** times in `src/` — otherwise it re-renders the *same* ``, which survives. Only Avatar-backed (and bespoke `onError → placeholder`) fixtures can flake an existence assertion.

2. **Coverage is partial.** Juxtaposing "14 mounting URLs" with "16 reported sites" reads as coverage; it isn't. The rule reports **9 distinct URLs of those 14**, so **≥5 (~36%) sit outside it**. `url` is the most common http-literal binding in the suite (36 occurrences) and is deliberately excluded because it is overwhelmingly used for non-image URLs. Widening it behind the existing `IMAGE_EXT_RE` proof is a deferred one-liner, now documented as such.

3. **"No diff in anyone else's blast radius" is true today, not permanently.** `--diff-filter=A` treats a rename-with-heavy-edit as ADDED, so *moving* one of the 5 backlog files lands it in the blocking step; and `lint.yml` records that the report-only steps are planned to flip to blocking. Both now named in the override comment.

4. **The data URI was not "the second copy of that constant".** `AppBlockChrome`'s was a **GIF**, `AppListingCard`'s a **PNG**. The shared constant is byte-identical to the PNG, so `AppListingCard` is unchanged while `AppBlockChrome` silently swapped GIF→PNG. Inert — the assertion compares against the constant — but not what was claimed.

5. **"Swapping the fixture back didn't turn it red locally" undersells the evidence — CI has the receipt.** The race is machine-speed-dependent, so a fast local box hides it, but clean CI pods show it plainly:

   | pod | result |
   |---|---|
   | `pr-preview-3550` (pre-fix) | **FAIL** — `AppBlockChrome … renders recents (icon + name)…`, 1 failed / 116 passed |
   | `pr-preview-3551` (the fix) | **117 passed (117)** |

   The fixture change is justified on evidence, not only on principle.

### One number in the audit itself was also wrong

The audit reported "389 `<Image` usages in `src/`". That is a naive substring count sweeping in `<ImageGuard` (81), `<ImageMetaPopover` (21), `<ImageContextMenu` (14) and others. The real figures are **51** bare `<Image>` occurrences across **10** files importing `Image` from `@mantine/core`. The audit's *direction* was right and its conclusion stands; the magnitude was off ~8×, so the corrected source comments deliberately cite the mechanism rather than that count.

### Not fixed here, tracked separately

`AppListingDetailBody.browser.test.tsx > the "Browse all apps" link is present even when the rail is empty` is a **pre-existing intermittent flake** on `main` — a strict-mode violation where `getByTestId('apps-browse-all')` resolves to 2 elements across two vitest containers (cross-test DOM-cleanup leak). Pod history: 3549 red · 3550 green · 3551 green · 3552 red · 3553 red, with #3552 (touching only `AppsPageLayout*`) as the discriminating control. It is not this PR's, and the fix likely lives in `test/component-setup.tsx` which this PR touches — so it is deliberately sequenced after this merges.

PR #3554, opened 2026-08-03
feat(app-blocks): tie a step's moderation posture to its orchestrator $type

ZacxDev:
## The gap

`moderationPosture` is **self-declared**, and every existing clause checks the declaration against *itself*: clause 1 asks whether the declared posture has a handler; clauses 1a/5a ask whether that posture's own field exists and returns something. **Nothing asked whether the declaration matches what the step actually does.**

So a `chatCompletion` entry declaring `'promptAudit'` — auditing its input, emitting unscanned free text — registers cleanly and passes every check. The gate whose entire purpose is *"a text-producing step cannot register until someone answers the policy question"* was satisfiable by answering a **different** question.

Same defect class as `auditableText` returning `''` (clause 5a) and `extractOutput` returning `[]` (clause 8), one level up: there the field was satisfiable by a no-op; here the field itself named the wrong axis.

## The change

`STEP_TYPE_ACCEPTABLE_POSTURES` (orchestrator `$type` → acceptable postures) + **clause 1b**, rejecting an entry whose declared posture is not acceptable for the `$type` it submits as.

**Behaviour-neutral for the shipped population** — `convertImage`'s `$type` carries no constraint, so no registered entry changes.

## Shape decisions, each measured rather than assumed

- **A set per type, not a strength ladder.** A ladder would bake in a claim nobody has made — that `'textOutput'`, once implemented, also audits input. That is a decision for whoever implements it; encoding it as an ordering would hide it.
- **Clause 1b sits *after* clause 1, deliberately.** The honest declaration (`chatCompletion` + `'textOutput'`) must still fail with *"no implemented handler"* — the accurate reason. 1b catches the dishonest one. A test pins that ordering.
- **Null prototype on the map** — the control `./moderation` already documents on its handler table. With a plain literal, an entry declaring `orchestratorType: 'toString'` indexes to `Object.prototype.toString` (truthy) and `.includes` on a function throws out of registry load.
- **`satisfies`, not a cast.** Writing it as `Object.create(null) as Record<…>` removes contextual typing and a mistyped posture (`'textOputput'`) **compiles clean** — I verified that, then fixed it. With `satisfies` it is `error TS2820`.
- **`$type`s enumerated from the generated `@civitai/client` literals**, not `src/**` usage — grepping usage enumerates what is *called*, not what *exists*. `textToSpeech` takes free text **in** and emits audio, so it accepts `'promptAudit'` as a floor rather than requiring `'textOutput'`.

## Residual gap, stated rather than papered over

This constrains the types **listed**. A text-producing `$type` nobody listed is unconstrained, exactly as before. Inferring text-production from the output shape is the sniffing this file already rejects for `auditableText`, and would be worse here because it would run against a sample the entry author also wrote. Mitigation: adding an entry is a reviewed PR, and this map is where review looks.

## Verification

- **84 tests green** across `step-registry` + `step-moderation`; 9 new.
- Every rejection is paired with a **negative control** asserting the same fixture passes when only the `$type` changes.
- **Mutation-tested**, three ways:
  - disabling clause 1b → **3 red** (reachability included), every control still green;
  - emptying the map → **4 red** (the non-empty guard exists precisely to catch a silently-disabled gate);
  - removing the null prototype → **exactly 1 red**, the prototype-seam test.
- **Typecheck: 0 errors in the changed area.** 🔴 Worth flagging for anyone else running it locally — `npx tsc --noEmit` **OOMs at the default heap** (exit 134, core dumped) and prints no `error TS` lines, so a naive read scores it as a clean 0. It needs `NODE_OPTIONS=--max_old_space_size=8192` (what the repo's own `typecheck` script sets). I only caught it by feeding the harness a known-bad file and getting silence back.

### DIFF
--- a/src/server/routers/__tests__/blocks.router.workflow.test.ts
+++ b/src/server/routers/__tests__/blocks.router.workflow.test.ts
@@ -287,6 +287,9 @@ vi.mock('~/server/services/orchestrator/promptAuditing', () => ({
 // unchanged pass count for the rest of the suite is the evidence.
 // ─────────────────────────────────────────────────────────────────────────────
 const { AUDITED_STEP_ID } = vi.hoisted(() => ({ AUDITED_STEP_ID: 'fixture-audited-step' }));
+const { DIVERGENT_STEP_ID } = vi.hoisted(() => ({
+  DIVERGENT_STEP_ID: 'fixture-divergent-step',
+}));
 vi.mock('~/server/services/blocks/steps', async (importOriginal) => {
   const actual = await importOriginal<typeof BlockStepsModule>();
   const z = await import('zod');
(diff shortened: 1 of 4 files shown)


ZacxDev:
## ⚠️ Correction: my original PR description overstated this change's guarantee

Posting this rather than silently editing the body, since anyone who already read it read the wrong version.

**What I claimed** (opening description, and in the code comment on clause 1b):

> clause 1b is "the only clause that reads the declaration against something the entry author did not also write… it checks it against the `$type` the step actually submits as."

**Both halves were false.** `orchestratorType` is a *second self-declared field*, and the router submits `buildStep(...).$type` — not the declared one (`blocks.router.ts:6669` → `:6707`). An adversarial review proved the bypass **by execution**: an entry declaring a benign `orchestratorType` with `moderationPosture: 'none'`, whose `buildStep` emits `$type: 'chatCompletion'`, **registered cleanly**. So as originally written, clause 1b was a speed bump against an honest-but-wrong entry, not a control.

It also caught that my own test fixture demonstrated the seam by accident — `makeFixtureStep` hardcoded its built `$type` even where tests overrode `orchestratorType`.

### Fixed in `e809a9fb`

- **New clause 7a** asserts `buildStep(...).$type === orchestratorType` per variant. 1b + 7a together tie the posture to the type actually submitted; neither is sufficient alone. Reuses the already-computed `built`, and hardens clause 9 for free (it reads the declared type too). Honest limit, stated in the comment: it probes *canonical* params, so a params-dependent `$type` switch could still diverge at request time — closing that means re-asserting at the router, which is recorded and deliberately not done here.
- The false comment is **corrected in place with its history noted**, not deleted. In a code-reviewed trust root the comment *is* the control: a reviewer who reads "already anchored" is a reviewer who stops checking.

### Three other findings, also fixed

- **The map contradicted its own stated principle.** It documented "A SET, NOT A LADDER, AND DELIBERATELY SO", then declared `textToSpeech: ['promptAudit','textOutput']` with "a stricter posture is also acceptable" — which *is* the subsumption claim. It was also a half-applied criterion (if text-IN qualifies, so do ~12 more types, none listed). Resolved structurally: the criterion is now exactly one thing — the type's **output** carries free text — so every entry is single-valued and encodes no subsumption at all. The text-IN axis is documented as out of scope with the residual gap recorded.
- **`Object.freeze` is shallow.** The posture arrays were mutable; review demonstrated `arr.push('none')` then a `chatCompletion` entry declaring `'none'` registering cleanly. Now frozen individually.
- **Map keys were unchecked** — `satisfies Record<string, …>` constrains values, not keys, so a typo'd key compiled clean and constrained nothing. New test pins every key against the generated `@civitai/client` `$type` literals.

Also added `promptEnhancement` / `xGuardModeration` / `echo` (all would have registered as `'none'`), and relabelled the "every SHIPPED entry" test as the **invariant guard** it is — vacuous today, and it survives deleting clause 1b.

### Verification

67 tests in the file (was 59); 92 across `steps/`. **Six mutations, each killing exactly its own clause**: disable 1b → 3 red · empty the map → 4 · drop the null prototype → 1 · disable 7a → 3 (including the bypass test) · un-freeze the arrays → 1 · typo a key → 4. Typecheck clean in the changed area. Still behaviour-neutral for the shipped population.

**What still stands from the original description:** the gap being closed is real, the change introduces no hole and strictly improves on the pre-PR state, and the typecheck-OOM warning is accurate. What changed is the *strength* of the claim — and it took a second pair of eyes to catch that I'd written a guarantee the code didn't provide.

PR #3560, opened 2026-08-03
fix(collections): restore the CollectionItem image FK, stop flagging characters as real people

JustMaier:
Two fixes surfaced by the first live AI review run on the Buzz Beggars Board.

## 1. Orphaned collection items — a schema/DB drift

`schema.prisma:4307` declares `image Image? @relation(..., onDelete: Cascade)`, but **no such foreign key exists in the database**. Deleting an `Image` leaves the `CollectionItem` behind: invisible to anything that joins `Image`, yet still counted as pending in review queues.

Measured on prod 2026-08-03:

| | count |
|---|---|
| Orphaned image-backed CollectionItems | **236,512** |
| — of those, `ACCEPTED` | 235,504 |
| — **`REVIEW`** (clogging review queues) | **866** |
| — `REJECTED` | 142 |

On the beggars board that was 124 rows the AI review job could never act on — its query inner-joins `Image` — so they sat in the queue looking like unreviewed work forever.

**Why not fix it in the application:** images are deleted from at least four places — `image.service.ts:406` (Prisma), `image.service.ts:456` and `post.service.ts:980` (raw SQL), `user.service.ts:1276` (bulk `deleteMany`). Cleanup would need adding to all of them and would still miss future paths and manual DB deletes. The constraint is the only fix that holds.

**Migration is deliberately three separable steps**, since `CollectionItem` is ~210M rows:
1. `DELETE` existing orphans (anti-join, couple of minutes)
2. `ADD CONSTRAINT … NOT VALID` — brief `ACCESS EXCLUSIVE` lock only, no table scan
3. `VALIDATE CONSTRAINT` — `SHARE UPDATE EXCLUSIVE`, so reads and writes continue

`CollectionItem_imageId_idx` (plain btree) backs the cascade lookup, so deletes stay indexed.

⚠️ **Manual apply, per the repo's migration policy** — and step 2 will fail while orphans remain, so run them in order.

**Same drift exists for `articleId`, `postId`, `modelId`** — all declared `Cascade`, none present in the DB. Their orphan counts are unmeasured (the combined scan timed out); scoped this PR to `imageId`, which is the one causing pain.

## 2. A named anime character rejected as a "real person"

Image 132988414 — Marin Kitagawa from *My Dress-Up Darling*, holding an "I need buzz!" sign — was rejected for `real person likeness`. The model's own reason said *"stylized anime character"* while flagging `depictsRealPerson: true`.

My prompt caused it: *"set depictsRealPerson only if you can NAME the specific real person"*. A named fictional character **is** nameable, and the prompt literally contained "Marin Kitagawa".

Section C now states that fictional characters are not real people however specific or nameable, that a prompt naming a character is evidence of a *character*, and that drawn art is essentially never a real-person likeness absent obvious caricature.

**Verified against the offending image: 3/3 runs now report `depictsRealPerson: false`.**

## Test plan

- [x] `pnpm typecheck` — clean (one pre-existing `minimax.handler.ts` failure, present on `main`)
- [x] 48 unit tests pass
- [x] Policy fix validated against the real image that triggered it
- [ ] Apply migration steps 1-3 to prod in order
- [ ] Re-check orphan count is 0 and the beggars review queue shows only genuine items


JustMaier:
Closing — not worth merging.

The data changes were applied directly to prod (249,774 orphans cleared, four FKs added and validated), and `schema.prisma` already declared these cascades, so the migration file would only have documented something already done. Root cause turned out to be the DigitalOcean → DataPacket move, where constraints were disabled and this set never got re-enabled.

An audit of all 468 declared relations against `pg_constraint` found **34 more** declared FKs missing from the database, overwhelmingly `imageId`-referencing — consistent with the same cause. Deliberately not fixing those in bulk: `ImageReaction`, `TagsOnImageNew` and `ImageResourceNew` are among the hottest write tables, so a missing FK there may now be load-bearing. Needs per-table triage.

The prompt fix in this PR was applied directly to the live config instead (the job reads the prompt from `KeyValue`, so merging would not have changed behaviour), and the prompt has since been removed from the repo entirely — it describes which signals reject versus escalate, and this repo is public.

The cleanup script is retained privately for the remaining 34 candidates.

Issue #1985, opened 2026-01-23
[Bug]: event-engine-common repo is private, making local builds impossible


lotushomerun:
### What happened?

I've managed to circumvent #1898 by commenting out the line relating to gen-seeds.ts, but after filling out S3 keys it complains about event-engine-common not existing (because it did not get pulled).

Upon further investigation the repo is actually private and does not pull with a typical public clone.

Old builds that do not require the submodule such as c30dc6ef79ba14e6c600526af0012bf6199c7c22 can be built after fixing further issues with the makefile:
1. gen-prisma: Change the line from `prisma generate` to `npx prisma generate`
2. dev: Change the line from `cross-env NODE_OPTIONS=` to `npx cross-env NODE_OPTIONS=`

### Steps to reproduce the problem

1. Clone Civitai
2. Circumvent #1898 by commenting out this line: `npx cross-env NODE_ENV=development tsx ./scripts/local-dev/gen_seed.ts --rows=$(ROWS) --trunc=$(TRUNC_QUEUE)`
3. Fill out S3 image keys 

### What should have happened?

Updating the submodule should pull the event-engine-common repository and allow building

### What platforms do you use to access the site?

Linux

### What browsers do you use to access the site?

Mozilla Firefox

### Additional information, context and logs

Fastfetch on the system building CivitAI locally:
```
OS: Arch Linux x86_64
Host: IdeaPad U410 (Lenovo U410)
Kernel: Linux 6.18.6-arch1-1
Uptime: 12 mins
Packages: 284 (pacman)
Shell: bash 5.3.9
Display (VGA-1): 1024x768, 60 Hz
Display (AUO233C): 1366x768 in 14", 60 Hz [Built-in]
Terminal: /dev/pts/1
CPU: Intel(R) Core(TM) i5-3317U (4) @ 2.60 GHz
GPU 1: NVIDIA GeForce 610M [Discrete]
GPU 2: Intel 3rd Gen Core processor Graphics Controller @ 1.05 GHz [Integrated]
Memory: 4.96 GiB / 5.65 GiB (88%)
Swap: 437.29 MiB / 2.82 GiB (15%)
Disk (/): 9.96 GiB / 456.39 GiB (2%) - ext4
```

<img width="1199" height="921" alt="Image" src="https://github.com/user-attachments/assets/46a9518c-0a77-4e07-be6d-14b331f25644" />

PR #3562, opened 2026-08-03
feat(apps): make the store listing hero banner click-to-launch

ZacxDev:
The `/apps/store-preview/<slug>` hero cover is the biggest, most obviously "this is the app" element on the listing detail — and it does nothing. The only way in is the CTA in the far-right action column. This wraps the banner in a real anchor to the same destination as the primary `open` action (`/apps/run/<slug>`).

Deliberately the cheap half: **no screenshot capture, no embedded runtime, and no `<iframe>`.** #3549 removed the bridge-less in-page preview, and the source-level gate in `__tests__/appListingDetailView.test.ts` that pins its absence is correct — this is a link and nothing more. That gate stays green (19/19).

## Design calls

**On-site `open` only.** An off-site listing (`visit`) leaves the banner inert, and so does the on-site raw-origin *"Open live"* escape hatch. Silently shipping a viewer off-platform from what reads as decorative art is a surprise; an off-site destination deserves the labelled, new-tab *"Visit ↗"* button it already has. An action with **no href at all** — the model-slot `info` branch, the `connect` stub — is inert for the obvious reason (`DetailPrimaryAction.href` is optional, so that is a real branch, not a hypothetical).

The target is derived from `getDetailPrimaryAction`, not by re-deriving the kind matrix, so the banner cannot drift from the button it mirrors:

```ts
const primaryAction = getDetailPrimaryAction(detail, { canOpenPage });
const heroLaunchHref =
  !preview && primaryAction.mode === 'open' && !primaryAction.external && primaryAction.href
    ? primaryAction.href
    : null;
```

`!external` is belt-and-braces for a future `open` that ever carries `external: true`.

**`preview` stays inert.** The moderator listing-media review renders this body read-only and omits the whole action column. A clickable banner there would put a live launch back onto an *unapproved* shadow listing — exactly what that posture exists to prevent. The preview test passes `canOpenPage` **true** on purpose, so the only thing that can be suppressing the banner is `preview`.

**Real semantics.** An `<a href>`, not a `<div onClick>`: tab-reachable, Enter-activatable, modifier-clickable, with a visible `focus-visible:ring-2` (paired with `focus:outline-none`, never a bare outline removal). It carries an explicit `aria-label="Open <app>"`, because neither branch can name it otherwise — the cover `alt` describes the *picture* (`"<app> cover image"`) and the placeholder is `aria-hidden` with no text. Nothing inside the banner is interactive, so no controls are nested (asserted).

**No double side effect.** `/apps/run/<slug>` records the recents entry itself, in a mount effect — which is why the `open` CTA has no `onClick` either; only the off-site `visit` branch records, because following that link leaves the SPA and nothing downstream could. The banner matches the CTA exactly: it records **nothing** on click. Verified rather than assumed, and against a positive control, since the claim under test is a zero.

**The art-less case still launches.** A listing whose owner has not uploaded a cover gets the affordance over its seeded-gradient placeholder, and the test asserts it really is the placeholder branch inside — otherwise it would be passing on the wrong branch.

Visually it stays in the card idiom: pointer, a small hover dim, the site's focus ring. The wrapper uses the **Mantine** `md` radius, not Tailwind's `rounded-md` — the two `md`s are 8px and 6px, and mixing them against the inner `<Image radius="md">` leaves a corner seam.

## Tests

`AppListingDetailBody.browser.test.tsx`, red-at-base measured **per test in isolation** (`-t`) against this PR's parent commit `ed6e17ac7d`:

| Test | At base |
|---|---|
| the banner is a real link to the SAME target as the Open CTA | 🔴 **RED** |
| a listing with NO cover art still gets the launch affordance | 🔴 **RED** |
| the banner is keyboard focusable and Enter activates it | 🔴 **RED** |
| the banner records recents exactly as the Open CTA does | 🔴 **RED** |
| preview mode renders NO launchable banner | 🟢 green |
| off-site (Visit) → banner NOT interactive | 🟢 green |
| the on-site raw-origin escape hatch → banner NOT interactive | 🟢 green |
| the model-slot info branch → banner inert | 🟢 green |
| POSITIVE CONTROL: the off-site Visit CTA records exactly 1 | 🟢 green |

**The five green ones are invariant guards, not regression coverage** — base renders no banner link at all, so they pass there vacuously, and I am not counting them. What they are worth is mutation-proof against this change's own guard; each was broken on purpose and each went red *for its own reason*:

| Mutation | Killed |
|---|---|
| M1 — drop the `!preview` clause | preview |
| M2 — accept `mode === 'visit'` too | off-site **+** escape hatch |
| M3 — naive "the banner always opens `/apps/run/<slug>`" | off-site **+** escape hatch **+** model-slot |
| M4 — add a `recordRecentlyOpenedApp` `onClick` to the banner | records-recents |

The recents zero is reported as a pair, never alone: the positive control clicks the off-site Visit CTA and reads **exactly 1** through the same spy the zeros are read from, so those zeros are evidence rather than a counter wired to nothing.

Keyboard activation is a real activation: focus the anchor, press Enter, and observe the browser-synthesised click — with a capture-phase `preventDefault()` on `document` so nothing navigates (which also makes `next/link` bail, since it checks `defaultPrevented`). `document.activeElement === el` on its own discriminates the real-semantics requirement, because a `<div onClick>` is not focusable without an explicit `tabIndex`.

### Latent false pass fixed on the way

The existing glyph tests looked the CTA up as `a[href="/apps/run/<slug>"]`, and the banner now matches that **first**. That made *"the in-site Open CTA and the off-site Visit CTA render DIFFERENT glyphs"* a genuine failure (it read the cover placeholder's category icon instead of the button's `player-play`) — and, worse, made *"the glyph change does NOT touch link semantics"* a **silent false green**, since the banner also has no `target` and no `rel`. Both selectors now exclude the banner explicitly, with a comment saying why.

## Results

- Browser spec: **28 passed (28)**, three consecutive runs on this base (6s, 6s, 6s warm).
- Node `unit`, the suites that can see the changed file: `appListingDetailView` (19, includes the no-raw-`<iframe>` gate), `appListingActionGlyph` (9), `recentAppsRail` (44) — **72 passed (72)**.
- Full node `unit` project: 10050 passed / 7 failed / 50 failed files — **all pre-existing and environmental**, 42 of them the missing private `event-engine-common` module (#1985) and its downstream implicit-`any` fallout. None import or read the changed file.
- `eslint` on both changed files: 0 errors (1 pre-existing `ThemeIcon` unused warning, left alone). `tsc --noEmit`: 12 errors, all the same pre-existing `event-engine-common` chain, none in these files.

## Not in scope

No screenshot capture. No install surface for the model-slot branch (still the gap #3493 tracks — that branch is vacuous today, since every approved on-site listing declares a page).

### DIFF
--- a/src/components/Apps/AppListingDetailBody.browser.test.tsx
+++ b/src/components/Apps/AppListingDetailBody.browser.test.tsx
@@ -368,14 +368,20 @@ describe('AppListingDetailBody', () => {
    * mounts and throw on strict mode. Each variant's href is unique, so each wait
    * can only be satisfied by its own mount.
    *
-   * 🔴 `:not([data-testid="apps-listing-hero-launch"])` is load-bearing. The
-   * click-to-launch hero banner is an anchor to the SAME `/apps/run/<slug>`
-   * href and sits EARLIER in the tree, so a bare `a[href=…]` returns the banner
-   * and `glyphOf` then reads the cover placeholder's category icon instead of
-   * the CTA's. That is a real failure for the glyph test — and, worse, a silent
-   * FALSE PASS for the link-semantics test below, whose assertions (no `target`,
-   * no `rel`) the banner happens to satisfy too. These tests are about the
-   * BUTTON; say so in the selector.
+   * 🔴 `[class*="Button-root"]` is load-bearing. The click-to-launch hero banner
+   * is an anchor to the SAME `/apps/run/<slug>` href and sits EARLIER in the
+   * tree, so a bare `a[href=…]` returns the banner and `glyphOf` then reads the
+   * cover placeholder's category icon instead of the CTA's. That is a real
+   * failure for the glyph test — and, worse, a silent FALSE PASS for the
+   * link-semantics test below, whose assertions (no `target`, no `rel`) the
+   * banner happens to satisfy too. These tests are about the BUTTON; say so in
+   * the selector.
+   *
+   * Discriminate on what the CTA IS, not on what it is not: an earlier revision
+   * excluded the banner's `data-testid`, which works here but is a trap to copy —
+   * `next.config.mjs` strips `data-testid` in production builds
+   * (`reactRemoveProperties`), so that shape matches nothing in a prod-build
+   * harness. Mirrors `waitForCta` in AppListingCard.browser.test.tsx.
    *
    * Timeout matches the 10s that `test/component-setup.tsx` deliberately sets as
    * the project-wide `vi.waitFor` default — its comment names the saturated
@@ -389,9 +395,14 @@ describe('AppListingDetailBody', () => {
     await renderWithProviders(ui);
     const cta = await vi.waitUntil(
       () =>
-        document.body.querySelector(
-          `a[href="${ctaHref}"]:not([data-testid="apps-listing-hero-launch"])`
-        ),
+        // 🔴 Discriminate STRUCTURALLY (the CTA is the Button), not by excluding
+        // the banner's `data-testid`. `next.config.mjs` strips `data-testid` in
+        // production builds via `reactRemoveProperties`, so a testid-based
+        // selector is correct in vitest (dev transform) but silently matches
+        // nothing in any prod-build harness. This mirrors the existing precedent
+        // at AppListingCard.browser.test.tsx's `waitForCta`, and it is
+        // self-describing: it says what the CTA IS rather than what it is not.
+        document.body.querySelector(`a[href="${ctaHref}"][class*="Button-root"]`),
       { timeout: 10000, interval: 25 }
     );
     return cta as HTMLAnchorElement;
@@ -651,9 +662,7 @@ describe('AppListingDetailBody', () => {
   test('🔴 a listing with NO cover art still gets the launch affordance', async () => {
     // The placeholder-gradient branch. An owner who has not uploaded a cover yet
     // must not silently lose the way into their own app.
-    const { within } = await renderScoped(
-      <AppListingDetailBody detail={base({ coverUrl: null })} canOpenPage />
-    );
+    const { within } = await renderScoped(<AppListingDetailBody detail={base({ coverUrl: null })} canOpenPage />);
     const hero = within.getByTestId(HERO);
     await expect.element(hero).toBeInTheDocument();
     expect(hero.element().getAttribute('href')).toBe('/apps/run/my-app');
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Correction to this PR's mutation table (post-audit)

An independent audit returned **safe to merge, no 🔴**, and confirmed the important things: the guard is exactly right across all seven `DetailPrimaryAction` shapes; the selector-collision class is **fully closed repo-wide** (a full sweep of `src/` and `tests/` found no site beyond the one fixed here); and the recents claim holds independently in both directions — the `open` CTA records nothing on click, `/apps/run/<slug>` records it in a mount effect, so an `onClick` here would double-write.

One claim in the body above is wrong. Correcting it publicly rather than editing silently, since it may already have been read.

### 🔴 The M2 mutation did not die

The table says mutant **M2** (widen the guard to accept `mode === 'visit'`) was killed by the off-site and escape-hatch tests. It isn't. **Every** `visit` return in `getDetailPrimaryAction` sets `external: true` (`appListingDetailView.ts:151` and `:177`), so widening the mode clause leaves `!primaryAction.external` false on both fixtures, `heroLaunchHref` stays `null`, no banner renders, and both tests still pass.

The deeper fact, which the body does not state: **`mode === 'open'` and `!external` are each individually redundant.** Deleting *either one alone* kills zero tests — off-site and escape-hatch are stopped by whichever clause survives, and model-slot and `connect` are stopped by `href`. Only the composite M3 ("always run-route") kills those three guards.

So **"each was broken on purpose and each went red for its own reason" holds for 3 of 4, not 4 of 4.** The code is correct — the guard is right, and arguably better for being redundant — but the evidence was overstated, and an overstated mutation table is exactly the kind of thing that gets cited later as proof.

### Other description overstatements

- *"Visually it stays in the card idiom"* — it did not. The Anchor recoloured the no-cover placeholder glyph from grey to link-blue. Fixed in `b8d14ac72e`.
- *"no controls are nested (asserted)"* — the assertion covers `a, button, input, select, textarea`, not `[role="button"]`, `[tabindex]`, `summary` or `[contenteditable]`. Adequate for today's content; narrower than the claim.

### Fixed in `b8d14ac72e`

1. `c="inherit"` on the hero Anchor — Mantine's Anchor sets `color: var(--mantine-color-anchor)` and Tabler icons stroke with `currentColor`. Only visible on listings **without** a cover, since an `` ignores `color`.
2. `renderAndSettle` now discriminates the CTA **structurally** (`[class*="Button-root"]`) rather than by excluding the banner's `data-testid` — `next.config.mjs` strips `data-testid` in production builds, so the testid shape is a trap to copy. Matches the existing `waitForCta` precedent.
3. The two hand-wrapped `renderScoped(...)` lines this PR added are now Prettier-clean; the 8 pre-existing hunks are deliberately untouched.

28/28 still pass.

### 🔴 Still owed: a human click-through

The component suite **loads no CSS at all** — `vitest.config.mts` registers only the process shim and `component-setup`. So every visual claim here (radius seam, focus ring, hover dim, and the glyph colour above) is reasoning, not measurement. Four checks, ~60 seconds, on a mod account:

1. Detail page for an approved on-site page app **with** a cover → click the hero → lands on `/apps/run/<slug>`.
2. A listing **without** a cover → glyph should be **grey**, not blue → click it.
3. `Tab` to the banner → visible focus ring, no layout shift, no corner seam.
4. `/apps` → "Recently opened" gained **exactly one** entry, not two.

PR #3563, opened 2026-08-03
perf(cache): decouple the metric-privacy defaults TTL from membership (10m -> 1h) + hit/miss metric

ZacxDev:
## What

`getUserMetricPrivacyDefaultsMap` read-through caches the three `hideModel*` booleans so the hot model-read paths (feed, v1 list, associated models) stop fetching and deserializing every owner's full `User.settings` JSON blob per request — the fix from #3331, which is working.

Its TTL was `MEMBERSHIP_CACHE_TTL` — a constant it **shared** with the creator-membership-validity cache in the same module. That coupling made the TTL both too short for what it actually guards and impossible to tune without silently moving the other cache's staleness window too.

Three changes:

1. **Decoupled TTL** — new `METRIC_PRIVACY_DEFAULTS_TTL = CacheTTL.hour`. The membership cache stays on `CacheTTL.md` (10 min), unchanged. Both constants now carry a comment explaining why they differ, so nobody re-merges them.
2. **Hit/miss metric** — on the shared `civitai_app_cache_{hit,miss}_total` counters under `cache_name="user-metric-privacy-defaults"`, so the TTL change is actually verifiable instead of taken on faith.
3. **Tests** — pin both TTLs as literal seconds, and cover the hit/miss accounting.

## Why an hour — the `User.settings` writer audit

The value is derived from an audit, not picked by feel. This cache derives from exactly one column, so the question is: what can change `User.settings` without busting the key?

**The three `hideModel*` flags are written by exactly one path** — `setUserSetting`, reached from the account Creator-Controls toggles — and that path busts this cache in the same call. The other writers that touch `settings` without going through `setUserSetting` provably cannot move these three flags:

| Writer | Touches `settings` how | Can it move `hideModel*`? |
|---|---|---|
| `setUserSetting` | jsonb merge of the patch | Yes — **and it busts this cache** |
| `setDismissedAlerts` | `jsonb_set` scoped to `{dismissedAlerts}` | No — path-scoped |
| `updateCreatorShopSettings` | whole-blob read-merge-write, fresh in-transaction row read, overrides only `creatorShop` | No — preserves the other keys |
| `copyGallerySettingsToAllModelsByUser` | whole-blob read-merge-write, fresh in-transaction row read, overrides only its own key | No — same shape |

So **nothing bypasses the bust**, which by the usual reasoning would justify a very long TTL (a day). I don't think that follows here, for two reasons that are about the *invalidation*, not the writers:

1. **The bust is best-effort and swallows its own error.** `bustUserMetricPrivacyDefaultsCache` catches and discards Redis failures by design, so a settings mutation can never be failed by a cache problem. That also means a dropped delete never retries — the TTL is the only thing that restores correctness.
2. **The refill reads a replica.** The write lands on the primary and the key is deleted, but the miss-fill below reads `dbRead`. A refill inside the replication window re-caches the **old** triple. This isn't an exotic interleaving — it's what happens when a user toggles the setting and immediately reloads to check it took.

These flags are a user-facing privacy control, so the worst case has to stay a blip rather than a support ticket. That rules out a multi-hour value however clean the writer audit is.

An hour is where the curve flattens anyway: **10 min → 1 h removes 5/6 of the periodic re-fetches; 1 h → 1 day would remove only ~4% more of the original churn while multiplying the worst case by 24.** Nearly all the available win is banked at an hour. Memory doesn't enter into it — the cached working set is a rounding error against the total keyspace and each entry is three booleans.

If the bust is ever made durable (retry/outbox) and the refill reads the primary, this can go up. The comment on the constant says so.

## Labelled vs unlabelled counter

Used the **existing labelled** `cacheHitCounter`/`cacheMissCounter` rather than adding two new unlabelled ones.

The known trap with labelled counters is real: a label child doesn't exist in the registry until its first `.inc()`, at which point it's already `1`, so `rate()`/`increase()` can never observe the `0→1` step. That bites **rare** counters — one that first fires minutes or hours into a process's life.

This is a per-request counter on a hot read path, so both children materialize on the first request a process serves and `rate()` is well-defined from the next scrape onward. Against that, the labelled counters are what six other caches in the codebase already use, so this cache shows up in any existing `civitai_app_cache_*` query for free instead of fragmenting the family. The residual caveat — a freshly-started process reports nothing for this cache until it has served one request, and a query must not read that gap as a real zero — is written into a comment next to the label constants for whoever builds the dashboard.

Counted **per user id, not per call**, so the ratio is a true per-entry hit rate over a batched lookup (matching how `createCachedArray` reports).

## Tests — red→green

Every new test was watched fail against pre-change code. Reverting only the service file to `origin/main` and rerunning: **7 failed / 30 passed**.

| Test | at `origin/main` | at HEAD |
|---|---|---|
| `derives only the three hideModel* flags…` (existing, TTL expectation updated) | 🔴 | ✅ |
| `writes the metric-privacy defaults with its own 1h TTL` | 🔴 | ✅ |
| `uses two DIFFERENT TTLs for the two caches in one run` | 🔴 | ✅ |
| `counts a full cache hit and never touches the miss counter` | 🔴 | ✅ |
| `counts a full cache miss and never touches the hit counter` | 🔴 | ✅ |
| `counts per id, not per call, on a mixed batch (2 hits / 3 misses)` | 🔴 | ✅ |
| `counts a redis read failure as all-miss` | 🔴 | ✅ |
| `leaves the membership-validity cache on its unchanged 10m TTL` | ✅ **invariant guard** | ✅ |
| `records nothing at all for an empty input` | ✅ **invariant guard** | ✅ |

The last two are green on both sides — they're labelled in the file as invariant guards, not counted as regression coverage. They're still load-bearing, and mutation-tested rather than assumed:

- **M1** — re-merge the constants (`METRIC_PRIVACY_DEFAULTS_TTL = MEMBERSHIP_CACHE_TTL`) → 3 failed. The decoupling guard fires.
- **M2** — bump `MEMBERSHIP_CACHE_TTL` to `CacheTTL.hour` in place, i.e. exactly the trap this PR exists to prevent → 3 failed, including `leaves the membership-validity cache on its unchanged 10m TTL`. That's the proof the invariant guard is reachable and discriminating, not vacuous.
- **M3** — corrupt the hit count (`const hits = misses.length`) → 4 instrumentation tests failed.

## Gate

- **Typecheck** — `tsc --noEmit`: **0 errors**, exit 0.
- **Lint** — eslint on both changed files via `--format json` to confirm the files were actually processed (an empty text report is indistinguishable from a no-op): `files linted: 2`, **0 errors, 0 warnings** each.
- **Unit suite** — full `unit` project: **749 test files passed (749); 10813 passed | 1 skipped (10814)**; 0 failed. Grepped the output for `test timed out` / `panic` / unhandled errors: 0 hits, so no silent truncation.
- Target suite alone: **37 passed** (29 pre-existing + 8 new).

## Out of scope, noted while auditing

`updateCreatorShopSettings` writes `User.settings` but busts neither `userSettingsCache` nor anything else, so a shop-settings write can be served stale from `userSettingsCache` until its own TTL. Harmless for *this* cache (it can't move the `hideModel*` flags), but it looks like a real gap in the settings cache. Not touched here to keep this PR to one concern.

### DIFF
--- a/src/server/services/__tests__/creator-membership.service.test.ts
+++ b/src/server/services/__tests__/creator-membership.service.test.ts
@@ -35,6 +35,9 @@ vi.mock('~/server/redis/client', () => ({
   },
 }));
 
+// Stubbed by the global setup (src/__tests__/setup.ts) — `.inc` is a vi.fn(), so these
+// are the assertion handles for the cache hit/miss instrumentation.
+import { cacheHitCounter, cacheMissCounter } from '~/server/prom/client';
 import {
   bustCreatorMembershipValidCache,
   bustUserMetricPrivacyDefaultsCache,
@@ -46,6 +49,16 @@ import {
 const keyFor = (id: number) => `packed:caches:creator-membership-valid:${id}`;
 const defaultsKeyFor = (id: number) => `packed:caches:user-metric-privacy-defaults:${id}`;
 const ALL_FALSE = { hideModelBuzz: false, hideModelDownloads: false, hideModelGenerations: false };
+
+// LITERAL expected TTLs, in seconds. Deliberately NOT `CacheTTL.*` — the point of these
+// pins is to fail if the two caches are ever re-merged onto one constant, and a
+// `CacheTTL`-derived expectation would silently follow such a change.
+const MEMBERSHIP_TTL_SECONDS = 600; // 10 minutes — membership validity, unchanged
+const METRIC_PRIVACY_TTL_SECONDS = 3600; // 1 hour — metric-privacy defaults
+const PRIVACY_CACHE_LABELS = {
+  cache_name: 'user-metric-privacy-defaults',
+  cache_type: 'redisPacked',
+};
 
 type SubRow = {
   userId: number;
@@ -276,11 +289,12 @@ describe('getUserMetricPrivacyDefaultsMap — derived read-through cache', () =>
       hideModelDownloads: false,
       hideModelGenerations: true,
     });
-    // Backfilled with exactly the tiny triple + TTL.
+    // Backfilled with exactly the tiny triple + this cache's own TTL (see the
+    // "cache TTLs are decoupled per cache" block for the decoupling guard).
     expect(mockRedis.packed.set).toHaveBeenCalledWith(
       defaultsKeyFor(1),
       { hideModelBuzz: true, hideModelDownloads: false, hideModelGenerations: true },
-      { EX: CacheTTL.md }
+      { EX: METRIC_PRIVACY_TTL_SECONDS }
     );
   });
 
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Audit corrections (comment-only, `1aba0a26b9`)

An adversarial audit ran 12 independently-constructed mutations against this branch and **killed all 12** — including confirming M2 (bump `MEMBERSHIP_CACHE_TTL` in place, the exact trap this PR exists to prevent) fails with `expected 3600 to be 600`, so that invariant guard is reachable and discriminating rather than vacuous. No deploy-blocking defect, no behaviour change needed.

It did find four wrong claims, all in durable comment text that instructs future maintainers. Corrected — **the diff contains no non-comment lines**; TTL value, counters, tests and behaviour are untouched.

### 1. The "~4%" figure was wrong by ~4×

Periodic re-fetch rate ∝ 1/TTL, so per entry per hour: 10min = 6, 1h = 1, 1day = 0.0417.

- 10min → 1h removes `(6−1)/6` = **83%**
- 1h → 1day removes a further `(1−0.0417)/6` = **~16%**, reaching ~99% cumulative

`~4%` is `1/24` — the churn *remaining* at a day, not the additional fraction *removed*. This mattered because the comment explicitly tells maintainers not to raise the TTL without re-deriving, so the wrong number under-valued the day option 4×. **The conclusion still holds** — an hour banks most of the win at 1/24th the worst-case staleness, and the privacy-control argument is what decides the trade — but "nearly all" was overstated. The comment now shows the arithmetic and names the error so it isn't re-derived the same way.

### 2. "The two writers" → three

The comment omitted `copyGallerySettingsToAllModelsByUser` (`model.service.ts`), which is the same whole-blob-replace shape. The PR body listed three; the shipped comment listed two — and the comment is the artifact a maintainer re-runs the audit from.

### 3. "Provably cannot move these flags" was too strong

Both whole-blob writers are read-modify-write at READ COMMITTED with no `FOR UPDATE`, so a `setUserSetting` landing between their read and update is silently reverted. That's a lost update on the **row**, not a stale cache — no TTL at any value helps, because the cache would be faithfully reporting a wrong row. Pre-existing, genuinely out of scope; the comment now states the caveat rather than overclaiming.

### 4. The counter caveat was wrong on mechanism

It claimed *"both children materialize on the first request a process serves."* They can't: both increments are conditional, so a single request creates at most **one** child — an all-hits batch never touches the miss child, an all-misses batch never touches the hit child. And Redis is shared rather than per-pod, so a freshly-rolled pod starts against a **warm** cache and may serve many requests before its first miss.

So the exposed side is the **miss** child — and it becomes *more* exposed exactly as this cache succeeds, since a longer TTL is designed to make misses rarer. That's a systematic downward bias on the very series you'd use to prove the TTL change worked. The loss is bounded (one scrape interval, once per pod per child per lifetime) so the decision to reuse the labelled counters still stands; only the stated reason was wrong. The comment now says to read the miss series by **value** (`max_over_time`), not by `rate()`.

### One follow-up outside this PR

The only existing consumer of `civitai_app_cache_hit_total`/`_miss_total` filters `namespace="civitai-prod"`, which doesn't appear to be a namespace that exists — production runs under a different one. If that's right, the metric this PR ships for verification would render nothing on the existing dashboard and the TTL change would stay unverified. That fix belongs in the infra repo; flagging it here so the metric isn't assumed readable.

PR #3572, opened 2026-08-03
fix(app-blocks): scope-annotate `getMyAppAnalytics` so the CLI login token can read its own analytics

ZacxDev:
## The bug

`blocks.getMyAppAnalytics` carried no `.meta({ requiredScope })`, so `enforceTokenScope` fell back to its default of `TokenScope.Full`:

```ts
const requiredScope = meta?.requiredScope ?? TokenScope.Full;
```

A Full personal API key therefore worked, but the OAuth token minted by `civitai login` — the CLI's **default** auth path, and the only credential most users have — was rejected with `FORBIDDEN: Your API key does not have the required scope for this action`. So `civitai app metrics <slug>` ([cli#190](https://github.com/civitai/cli/pull/190)) would not work over the CLI's default auth path. Scope note, corrected from an earlier draft of this description: that command is on the CLI's `main` but in **no released tag** yet (latest is v0.1.89), and its audience is moderators plus the `app-blocks-author` Flipt cohort — so this is a fix-before-release, not a live outage for most users.

## The scope choice

Annotated with **`TokenScope.AppBlocksSubmit`**. This was picked deliberately rather than copied from the nearest precedent:

| Candidate | Verdict |
|---|---|
| **`AppBlocksSubmit`** | ✅ Already required by `GET /api/v1/blocks/submissions`, which **the same command already calls** to resolve `slug → appBlockId`. Opt-in, excluded from `Full`. |
| `UserRead` | ❌ Inside `Full` — what every third-party "log in with Civitai" client requests. Would expose install counts, Buzz spend, endpoint names and active-user counts to all of them. |
| `AppBlocksDevTunnel` | ❌ The `stopDevTunnel` precedent, but semantically unrelated — its consent label is literally "Open on-site dev tunnels". Would force analytics readers to grant tunnel-opening and vice versa. |
| a new `AppBlocksRead` bit | ❌ Correct in the abstract, but needs a prod `allowedScopes` migration **plus** a CLI `DeviceScope` bump with release ordering. The device flow's `validateScope` is all-or-nothing, so a bit outside the client's `allowedScopes` breaks login entirely. |

`AppBlocksSubmit` also keeps the rule that route already states: *"the same credential that could submit can read its own submissions, and nothing weaker."* With any other bit, the two hops of one command would need different scopes.

This is the same bug class, same bit, and same test shape as `app-listings.router.cli-scope.test.ts` (for civitai/cli#186) — another set of un-annotated procs that 403'd this same token.

## No regression, no migration

- `enforceTokenScope` early-returns on `ctx.tokenScope === TokenScope.Full`, so **Full personal API keys still pass** (note `Full` deliberately *excludes* bit 25, so it is the early-return — not `hasFlag` — doing the work).
- ⚠️ **That early return is exact equality, so this property is conditional, not universal.** A strict *superset* of Full lacking bit 25 (`Full|AppBlocksDevTunnel` = 100663295) passed before and is FORBIDDEN after. No such credential can be created — `api-key.schema.ts` and `oauth-client.schema.ts` both cap at `.max(Full)` — and those caps are now **pinned by a test**, with a boundary assertion that they still admit Full. See the correction comment below.
- `createContext` defaults a session to `Full`, so the web `/apps/revenue` panel is unaffected.
- The `civitai-cli` OAuth client's migration sets this bit and the login token requests it, so **no new `allowedScopes` change is needed**. This repo cannot prove prod state (those migrations are manual-apply and the device flow lives on the auth hub); if bit 25 were absent in prod the failure mode is "this fix does not take effect", never "a working path breaks".

## Verification

Of the five tests added, **only one is regression coverage** — this is measured, not assumed. With the `.meta` line deleted:

- `CLI OAuth login token (carries AppBlocksSubmit) reaches the service` → **red**, failing with this gate's own error at `enforce-token-scope.ts:84` (the exact 403 being fixed).
- The three other **behavioural** tests stay **green** on pre-change code, as does the enum-only bitmask sanity test — four green, one regression guard. They are labelled in the file as **invariant guards, not regression guards**. In particular the `NO_SUBMIT` case was already FORBIDDEN before, because an un-annotated proc implicitly required `Full` and that token lacks `Full` too — it pins that narrowing the gate did not accidentally *widen* it, a different (worthwhile) property.

Please don't read the green tests here as multiple tests of the fix; the file says so too.

**A later round added a second proc** — `blocks.getMyForgejoCloneInfo` had the identical live defect (`civitai app pull` drives it) and is annotated with the same bit, with its own 3-case matrix. `getMyRevenue`/`getMyApps` stay un-annotated deliberately, now documented. Details in the correction comment.

Gate:

- Full unit suite: **755 files / 10994 passed / 1 skipped / 0 failed** (at the current tip, i.e. including the fix round's tests).
- `tsc --noEmit`: **0 errors**, no OOM — harness validated by injecting a type error and confirming tsc reports exactly 1 (a bare `tsc` without `--max_old_space_size=8192` OOMs and reports a *false* 0).
- Prettier: the added regions are clean. Remaining violations in both test files are pre-existing on `main` (classified by hunk position, with the invocation positive-controlled against a known violator — `--check` prints its all-clean message even when it matched *zero* files, so a bare green isn't trustworthy).
- Mutation-tested: dropping either `.meta` line, or raising either schema cap, each turns the corresponding guard red on its own specific error. Every mutation was verified to have actually applied before its result was read.

### DIFF
--- a/src/server/routers/__tests__/blocks.router.getMyAppAnalytics.test.ts
+++ b/src/server/routers/__tests__/blocks.router.getMyAppAnalytics.test.ts
@@ -208,7 +208,8 @@ describe('getMyAppAnalytics — gate', () => {
   });
 
   it('flag OFF (even for a moderator): returns zeroed analytics + runs NO aggregate', async () => {
-    // moderatorProcedure passes (mod), but the appBlocks flag is OFF →
+    // appDeveloperProcedure passes (fakePerUserFlag keys the author capability off
+    // isModerator, hence the "moderator" test names), but the appBlocks flag is OFF →
     // enforceAppBlocksFlag marks _appBlocksDisabled on the query ctx → the proc
     // short-circuits to the empty shape and never touches the aggregate service.
     mockIsAppBlocksEnabled.mockResolvedValue(false);
@@ -338,17 +339,19 @@ describe('getMyRevenue — dark-flag short-circuit', () => {
  * scope for this action", enforce-token-scope.ts:84) — the exact 403 this change
  * fixes.
  *
- * Everything else in this describe block stays GREEN on pre-change code: the three
- * BEHAVIOURAL cases below plus the enum-only bitmask sanity test, i.e. FOUR green
- * tests, only one of which is regression coverage. They are INVARIANT guards, not
- * regression guards:
+ * EVERY other test in this describe block stays GREEN on pre-change code — currently
+ * the three behavioural cases below, the enum-only bitmask sanity test, and the
+ * schema-cap test at the end. Count them at the source rather than trusting a number
+ * written here; a previous revision of this comment said "four" and was made stale by
+ * a later round appending a fifth. Exactly ONE test in this block is regression
+ * coverage. The rest are INVARIANT guards:
  *   - NO_SUBMIT was already FORBIDDEN before, because an un-annotated proc implicitly
  *     requires `Full` and that token lacks Full too. It pins that narrowing the gate
  *     did not accidentally WIDEN it — a different property, worth keeping, but it
  *     would not have caught the original bug.
  *   - the Full-key and session cases pin no-regression, and passed before by
  *     construction.
- * Do not read four green tests here as four tests of the fix.
+ * Do not read a block of green tests here as a block of tests OF THE FIX.
  */
 const FULL = 33554431; // TokenScope.Full — a Full personal API key
 const CLI = 1 | (1 << 25) | (1 << 26); // UserRead|AppBlocksSubmit|AppBlocksDevTunnel = 100663297
@@ -412,12 +415,25 @@ describe('getMyAppAnalytics — OAuth scope gate', () => {
    * enforceTokenScope's bypass is exact equality (`ctx.tokenScope !== TokenScope.Full`),
    * NOT `hasFlag`. So a mask that is a strict SUPERSET of Full but lacks bit 25 —
    * `Full|AppBlocksDevTunnel` = 100663295 — satisfied the un-annotated gate before and
-   * is FORBIDDEN after. That is a real allow→deny flip, harmless only because no such
-   * credential can be created. The caps that make it unreachable are pinned here: if
-   * either is raised, the flip becomes reachable and every `requiredScope` in this
-   * router needs re-examining.
+   * is FORBIDDEN after. A real allow→deny flip.
+   *
+   * 🔴 SCOPE OF THIS TEST — it does NOT prove the flip is globally unreachable, and an
+   * earlier revision of this comment wrongly implied it did. It pins the two PUBLIC
+   * (zod-validated) surfaces only:
+   *   - a personal API key's `tokenScope`, and
+   *   - an OAuth client's `allowedScopes` via tRPC create/update.
+   * It does NOT and cannot cover: OAuth ACCESS tokens (the hub bounds a requested scope
+   * against `ALL_SCOPES`, not Full — deliberately, so an opt-in bit is not dropped, so a
+   * token's real ceiling is its client's allowedScopes), or `allowedScopes` written by
+   * RAW SQL MIGRATION, which is exactly how the one client exceeding Full got its value
+   * (civitai-cli, 100663297 — not a superset of Full, since it carries bit 0 and none of
+   * 1..24, which is why the flip is unreachable in practice).
+   *
+   * So: if either cap below is raised, this test goes red. If a MIGRATION grants some
+   * client `Full | <opt-in bit>`, nothing here will notice — that case is held only by
+   * inspection.
    */
-  it('no creatable credential can hold a superset-of-Full mask (the flip is unreachable)', async () => {
+  it('the two zod-validated credential surfaces reject a superset-of-Full mask', async () => {
     const { addApiKeyInputSchema } = await import('~/server/schema/api-key.schema');
     const { createOauthClientSchema, updateOauthClientSchema } = await import(
       '~/server/schema/oauth-client.schema'
@@ -443,13 +459,19 @@ describe('getMyAppAnalytics — OAuth scope gate', () => {
     expect(updateOauthClientSchema.safeParse({ id: 'c', allowedScopes: SUPERSET }).success).toBe(
       false
     );
-    // Boundary: the caps still ADMIT Full itself, so they are not vacuously rejecting
-    // everything (which would make the three assertions above prove nothing).
+    // Boundary controls: each cap must still ADMIT Full, or the rejections above prove
+    // nothing (a fixture missing a required field would reject regardless of the scope,
+    // and the test would still pass with the cap removed). One per assertion above —
+    // the fixtures differ ONLY in the scope field, so a green here isolates the cap as
+    // the cause.
     expect(addApiKeyInputSchema.safeParse({ name: 'k', tokenScope: TokenScope.Full }).success).toBe(
       true
     );
     expect(
       createOauthClientSchema.safeParse({ ...clientBase, allowedScopes: TokenScope.Full }).success
+    ).toBe(true);
+    expect(
+      updateOauthClientSchema.safeParse({ id: 'c', allowedScopes: TokenScope.Full }).success
     ).toBe(true);
   });
 });
--- a/src/server/routers/__tests__/blocks.router.getMyAppAnalytics.test.ts
+++ b/src/server/routers/__tests__/blocks.router.getMyAppAnalytics.test.ts
@@ -8,8 +8,8 @@ import { TRPCError } from '@trpc/server';
  * stubbed so importing the router doesn't drag in the stale generated
  * Prisma client). The analytics SERVICE is mocked at the boundary — this
  * test asserts the ROUTER wiring:
- *   - moderatorProcedure + enforceAppBlocksFlag gate (non-mod / anon rejected,
- *     dark behind the appBlocks flag);
+ *   - appDeveloperProcedure (the `appBlocksAuthor` capability) + enforceAppBlocksFlag
+ *     gate (non-author / anon rejected, dark behind the appBlocks flag);
  *   - the caller's session user id is threaded into the service (ownership is
  *     enforced inside the service, covered by app-analytics.service.test.ts);
  *   - the zod input is validated (appBlockId length cap, from/to datetime).
@@ -336,7 +336,11 @@ describe('getMyRevenue — dark-flag short-circuit', () => {
  * `.meta` line deleted, ONLY 'CLI OAuth login token … reaches the service' goes red,
  * and it fails with this gate's own error ("Your API key does not have the required
  * scope for this action", enforce-token-scope.ts:84) — the exact 403 this change
- * fixes. The other three stay GREEN on pre-change code and are INVARIANT guards, not
+ * fixes.
+ *
+ * Everything else in this describe block stays GREEN on pre-change code: the three
+ * BEHAVIOURAL cases below plus the enum-only bitmask sanity test, i.e. FOUR green
+ * tests, only one of which is regression coverage. They are INVARIANT guards, not
  * regression guards:
  *   - NO_SUBMIT was already FORBIDDEN before, because an un-annotated proc implicitly
  *     requires `Full` and that token lacks Full too. It pins that narrowing the gate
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correction to the description above, plus an audit fix round

An adversarial audit pass turned up two places where **the PR description as originally written was wrong or overstated**. Correcting them here rather than silently editing the body, since anyone who already read it read the wrong version. (I have also updated the body.)

**1. I overstated the live impact.** The body said:

> That made `civitai app metrics <slug>` (civitai/cli#190) unusable for the majority of its intended audience.

Two things wrong with that. `civitai app metrics` is on the CLI's `main` but is in **no released tag** (latest is v0.1.89), so no released CLI can hit this 403 yet. And the audience is bounded to moderators plus the `app-blocks-author` Flipt cohort — not "most users". The direction of the fix is unchanged; the framing was inflated. Accurate version: *the OAuth login token would 403 this proc, so the command would not work over the default auth path once released.*

**2. The no-regression claim was stated as unconditional, and it is not.** `enforceTokenScope`'s bypass is exact equality (`ctx.tokenScope !== TokenScope.Full`), **not** `hasFlag`. So a mask that is a strict *superset* of Full but lacks bit 25 flips from allow to deny:

| mask | before | after |
|---|---|---|
| `33554431` (Full) | PASS (early-return) | PASS (early-return) |
| `100663297` (CLI token) | FORBIDDEN | **PASS** ← the fix |
| `100663295` (`Full\|AppBlocksDevTunnel`) | **PASS** | **FORBIDDEN** ← the flip |

No such credential can exist today — but that is enforced by schema caps, not luck: `api-key.schema.ts` caps a personal key's `tokenScope` at `.max(Full)`, and `oauth-client.schema.ts` caps `allowedScopes` at `.max(Full)` on both create and update. **Those caps are now pinned by a test**, including a boundary assertion that they still *admit* Full so they can't be vacuously rejecting. If either is raised, the flip becomes reachable and every `requiredScope` in this router needs re-examining.

**3. A third comment claim asserted infra state this repo cannot see** — "No `allowedScopes` migration either: the civitai-cli client is already provisioned with this bit." Both relevant migrations are manual-apply and the device flow lives on the auth hub. Reworded to what's knowable, and to note that if bit 25 were absent in prod the failure mode is *"this fix doesn't take effect"*, never *"a working path breaks"*.

---

## New in this round: a second proc with the same live defect

`blocks.getMyForgejoCloneInfo` had the identical bug and I'd missed it. It's `protectedProcedure` with no `.meta`, so it implicitly required Full and 403'd the same token — and **`civitai app pull` drives it** (`cli/internal/appapi/appblocks.go:739`). The CLI surfaces that failure as *"not permitted (are you the app owner, and is Apps enabled for your account?)"*, which sends a developer hunting an ownership problem they don't have. Annotated with the same bit, with the same 3-case matrix in its own test file.

`getMyRevenue` and `getMyApps` stay un-annotated **deliberately** — no CLI command calls them, so they're reached only by the session-authed web panel, which takes the Full early-return anyway. That's now written down, because two adjacent identically-gated procs differing with no explanation is a trap.

## Verification for this round

Every mutation was verified to have actually applied before its result was read, and all three touched files confirmed byte-identical to their backups afterwards:

- drop `.meta` from `getMyForgejoCloneInfo` → **1 red**, failing with the gate's own `TRPCError: Your API key does not have the required scope for this action`
- raise the api-key cap → the superset test **red**
- raise the oauth-client create cap → the superset test **red** (so it's load-bearing on *both* caps independently, not just one)

Full unit suite **755 files / 10994 passed / 1 skipped / 0 failed**; `tsc --noEmit` **0 errors**, no OOM. Prettier: the one violation my earlier region introduced is fixed; the rest in both test files are pre-existing on `main` (classified by hunk position, with the invocation positive-controlled against a known violator).

## Left for a maintainer decision

The `AppBlocksSubmit` consent label is `'Submit Apps for review'`, which now also authorizes reading a developer's app economics (installs, Buzz spend, endpoint names, active users). That under-describes what consenting grants. It's shared consent copy in `@civitai/auth`, so it's a product-copy call rather than a drive-by edit — flagging rather than changing it. If an `AppBlocksRead` bit is ever added, this proc should move to it.

One thing worth confirming out-of-band, since the repo can't: `SELECT "allowedScopes" FROM "OauthClient" WHERE id='civitai-cli';` should be `100663297`. A wrong answer means this PR doesn't take effect, not that it breaks anything.

ZacxDev:
## Delta re-audit round (`b59d6b1cbf..81bd555998`) — corrections applied in `7c431db036`

The re-audit found **no 🔴** and confirmed the delta's behaviour is correct and covered (full unit suite and typecheck clean, verified independently). But it caught the fix round reintroducing the *same class of defect it was closing* — a comment asserting a mechanism the code doesn't implement — twice. Both now fixed. No behaviour change in this commit beyond one added assertion.

**1. The superset-of-Full justification named the wrong enforcement mechanism.** I wrote that the flip is "enforced by schema caps rather than luck". Those `.max(Full)` caps are real, but they **do not govern OAuth access tokens**: the hub decodes a requested scope against `ALL_SCOPES`, not `Full` — deliberately, so a legitimately-requested opt-in bit isn't dropped — so a token's ceiling is its client's `allowedScopes`. And the one client exceeding `Full` (civitai-cli, `100663297`) got that value from a **raw SQL migration**, which no zod schema governs.

The conclusion still holds — `100663297` is not a superset of `Full` (bit 0, none of 1–24) — but that last step rests on **inspection of a migration, not a cap**. The comment now says exactly that, and names what would silently escape: a future migration granting some client `Full | <opt-in bit>` flips this gate and no test will catch it.

The test was titled `no creatable credential can hold a superset-of-Full mask` — claiming the global property it cannot prove. Renamed to `the two zod-validated credential surfaces reject a superset-of-Full mask`, with its docstring stating the two surfaces it covers and the two it can't.

**2. The green-test count went stale again — broken by my own previous round.** The comment said "FOUR green tests" while the block now holds six, five green on pre-change code: the fix round appended the fifth green test 78 lines *below* the comment asserting the count. Rather than write a fifth number the next round can invalidate, the comment now tells the reader to count at the source and records why — a hardcoded count in a docstring is a claim with no gate on it.

**Also addressed:**
- **`getMyForgejoCloneInfo` now argues its own bit choice** instead of inheriting "same reason as getMyAppAnalytics". It isn't the same stake: analytics is a pure read, but this proc **mints a credential** — a per-user Forgejo identity whose token carries `write:repository`, returned embedded in `cloneUrl` — so annotating widens who can trigger that mint to include the CLI login token. Accepted, with the bounds that make it acceptable written down (opt-in bit, first-party client only, ownership + not-banned + `approved` still required, per-repo `read` grant, and already the established meaning of the bit for CLI-facing procs incl. mutations). Plus: if an `AppBlocksRead` bit ever lands, **this** proc is the one that must not move to it.
- **Added the missing boundary control for `updateOauthClientSchema`.** Its two siblings each had a `Full → success` assertion isolating the cap as the cause of the rejection; the update path had only the rejection. Verified load-bearing: raising *only* the update cap turns the test red, so all three caps are independently pinned.
- Fixed the last stale `moderatorProcedure` comment.

**Not fixed, by design:** the consent-label copy (still a product decision in `@civitai/auth`), and an out-of-delta pre-existing note the re-audit raised — `addCollaborator` is a PUT that *sets* permission, so `civitai app pull` granting `read` on an app where the web git-access flow already granted `write` would downgrade it. That was already reachable with a Full personal key pre-change; this PR makes it reachable by the CLI's default credential. Worth a separate look; neither the re-auditor nor I verified Forgejo's live PUT semantics.

Verification: both touched test files 27/27; full unit suite **755 files / 10994 passed / 1 skipped / 0 failed**; `tsc --noEmit` 0 errors, no OOM; zero prettier hunks in the added regions.

ZacxDev:
## Third delta re-audit round (`81bd555998..7c431db036`) — 🔴 fixed in `5e6dde4bb3`

This round found a **🔴 in a comment the *previous* round wrote to fix a comment** — same defect class, third occurrence in this PR. Comment/docstring only, no behaviour change. Worth stating plainly rather than burying: the recurring failure here has been *me asserting a mechanism the code doesn't implement*, and it took three adversarial passes to stop.

**🔴 The parenthetical I added last round was wrong.** It said `appDeveloperProcedure` passes because "fakePerUserFlag keys the author capability off isModerator". Verified wrong:
- `hasAppBlocksAuthor` (`trpc.ts`) reads **only** `getFeatureFlags(ctx).appBlocksAuthor`
- this test file **does not mock** `feature-flags.service` at all (0 references)
- `fakePerUserFlag` is the mock implementation of `isAppBlocksEnabled` — the **dark** flag, which is the *second clause of that same sentence*

So the comment attributed both gates to one mock, got the first one wrong, and contradicted the file's own header docstring. The discriminator needs no mutation: that very test forces `mockIsAppBlocksEnabled.mockResolvedValue(false)` and still returns successfully — if that mock keyed the author capability, the proc would throw FORBIDDEN first. It now states both gates and how each is actually satisfied (static `availability: ['mod']` fallback vs. the mocked dark flag), keeping the note that the "moderator" test *names* are correct since both gates key off `isModerator` by different code.

**Three 🟡 clauses also fixed:**
- **"Same gate" → "Same SCOPE gate"**, plus a note that the base procedures **differ** (`appDeveloperProcedure` vs `protectedProcedure` + inline feature check). That matters because the block then argues about who can trigger the credential mint — a broader audience than the author cohort.
- **The `AppBlocksRead` note was internally tense.** Moving `getMyAppAnalytics` to a future read bit *alone* would recreate the "two scopes for one command" problem, since `civitai app metrics` also calls `GET /api/v1/blocks/submissions`. Now says they'd have to move together.
- **The non-zod-writer list read as an exhaustive partition and omitted a third writer**: `publish-request.service` writes `allowedScopes` for `appblk-*` clients. Safe by *construction* (mapped bits all below 25, `grants: []` so no bearer token) rather than by any cap the test asserts on. Now listed, and labelled "known when written" rather than a partition.

**Nits:** corrected "a token's ceiling is its client's `allowedScopes`" to `allowedScopes | UserRead` (all three hub sites OR in bit 0); pointed "the test below" at the right file; replaced an `enforce-token-scope.ts:84` line reference with the function name so it can't rot.

Verification: both touched files 27/27; full unit suite **755 files / 10994 passed / 1 skipped / 0 failed**; `tsc --noEmit` 0 errors, no OOM (harness positive-controlled); zero prettier hunks in the added regions.

The re-audit independently confirmed by computation that `100663297` is not a superset of `Full`, that `FORGEJO_TOKEN_SCOPES = ['write:repository']`, that the token is embedded in the returned `cloneUrl`, and that the docstring's test enumeration matches the source exactly — so those claims stand.

ZacxDev:
## Fourth audit round: **clean** — the pattern has terminated

Round 4 audited `7c431db036..5e6dde4bb3` (comment-only) and found **no 🔴 and no 🟡**. Verdict: safe to merge. All seven of the previous round's corrections verified against the code, each now naming the mechanism that actually implements it, with correct file and function names, and none contradicting an unchanged neighbour — which is exactly how rounds 2 and 3 each failed.

Worth recording what it took, since three consecutive rounds found the same defect class (a comment asserting a mechanism the code doesn't implement) and every fix round introduced the next one: round 1 → 5 🟡, round 2 → 2 🟡 (in round 1's fixes), round 3 → 1 🔴 (in round 2's fix), round 4 → clean. **None of these were ever caught by the mechanical gate** — the suite and typecheck were green at every single tip.

The round also independently confirmed by execution the one thing I'd only argued by reading: the author capability really is keyed on `isModerator` via the static `availability: ['mod']` arm, because Flipt yields nothing in the test env (`isFliptSync` returns null — no local overrides, no reachable server), and the mod-passes/non-mod-rejected pair discriminates, since `enforceAppBlocksFlag` **cannot throw on a query** (it fail-opens with `_appBlocksDisabled`). So the non-mod rejection can only come from `hasAppBlocksAuthor`.

### One more fix: `430a8d107b`

Round 4's one 🟢 was that my annotation makes an **unchanged** neighbouring docstring imprecise — it claimed `getMyForgejoCloneInfo` is "distinct from `getMyAppRepo` only in intent". True before this PR, not now: they also differ in token scope, since only one carries `.meta`.

While fixing it I checked whether **`getMyAppRepo` is a fourth instance of the same bug. It is not** — zero references in the CLI (positive control: `getMyForgejoCloneInfo` has 5 in `internal/appapi/appblocks.go`), and its callers are web-only (`AuthorViaGit.tsx`, `git-access.ts`, browser tests). So it's session-authed, takes the Full early-return, and is correctly left un-annotated. That's now written down rather than left for the next reader to re-derive. The comment enumerates all three real differences (token scope, intent, collaborator `read` vs `write`) and confirms ownership gating is still identical.

Gate at the final tip: full unit suite **755 files / 10994 passed / 1 skipped / 0 failed**; `tsc --noEmit` 0 errors, no OOM (harness positive-controlled); the three related router test files 33/33; prettier clean on the touched file.

### Still outstanding — needs someone with prod access

`SELECT "allowedScopes" FROM "OauthClient" WHERE id='civitai-cli';` should return `100663297`. Unverifiable from the repo (manual-apply migrations, hub-side device flow). If it's wrong, this PR simply doesn't take effect — it cannot break a working path.

And one lead neither audit could close, out of scope here: `addCollaborator` is a PUT that **sets** permission, so `civitai app pull` granting `read` on an app where the web git-access flow already granted `write` would downgrade it and break a later push. Already reachable with a Full personal key before this PR; this PR makes it reachable by the CLI's default credential. Nobody verified Forgejo's live PUT semantics.

ZacxDev:
## ✅ The outstanding prod question is now CLOSED — measured, not assumed

Ran the read-only queries against the prod cluster. Both things the audits flagged as unverifiable-from-the-repo are now answered with data.

**1. `civitai-cli` allowedScopes — confirmed.**

```
     id      | allowedScopes | has_appblocks_submit | has_dev_tunnel
-------------+---------------+----------------------+----------------
 civitai-cli |     100663297 | t                    | t
```

Bit 25 is live. **This PR will take effect.**

**2. The allow→deny flip class is empty in prod — and I measured the right table.** The gate reads `ctx.tokenScope`, so `ApiKey` is authoritative, not `OauthClient`:

| | count |
|---|---|
| total keys | 6,685,302 |
| exactly `Full` (safe early return) | 6,637,185 |
| **strict superset of `Full` lacking bit 25 (the flip)** | **0** |
| exceeding `Full` | 331 |

Only two distinct scope values exceed `Full`, and **both carry bit 25**, so both pass the new gate and neither is a superset of `Full`:

| tokenScope | keys | bit 25 | bit 26 | superset of Full |
|---|---|---|---|---|
| `100663297` | 301 | t | t | f |
| `33554433` | 30 | t | f | f |

Same check on `OauthClient`: 145 clients are *exactly* `Full` (early return), **0 strict supersets**, 63 below, 1 exceeding (civitai-cli). So the conditional no-regression argument holds in prod, not just in principle.

### Two things this changes in the PR's own claims

**The impact is bigger and more concrete than I stated.** I'd walked the description back to "a fix-before-release" because the CLI command is in no released tag. But there are **331 live OAuth login tokens** that were 403ing this proc, all of which this PR unblocks. The `metrics` command specifically is still unreleased; the *scope defect* was affecting real credentials.

**It also retroactively validates the bit choice as more than an aesthetic call.** 30 of those tokens carry `33554433` = `UserRead|AppBlocksSubmit` — pre-dating the dev-tunnel widening, so they **lack bit 26**. Had I copied `stopDevTunnel`'s `AppBlocksDevTunnel` bit (the nearest precedent, and the thing the handoff note pointed at), **those 30 tokens would still 403.** `AppBlocksSubmit` is the only one of the three candidates that unblocks all 331.

One caution on my own method: my first pass at the flip query used `(mask & Full) = Full`, which is also true when the mask *equals* `Full` — it returned "145 flip risks" and I nearly reported a false alarm. The strict-superset form (`AND mask <> Full`) is what gives 0. Noting it because the naive predicate looks right.

PR #3212, opened 2026-07-17
security(app-blocks): remove 3 unenforced decorative block scopes + fix 2 audit-integrity issues

ZacxDev:
## What / why

Security-hygiene pass on App Block scopes. Three scopes were **declared, validated, mapped, and mintable but never enforced by any runtime capability** — purely decorative:

- `media:read:owned`
- `block:settings:read`
- `block:settings:write`

None was an exploitable bypass (their "capabilities" were never wired as block-token endpoints), but the mod-review UI implied they were enforced and one audit-log path asserted an unchecked scope. After this change **every remaining declared block scope is one that is actually enforced per-operation server-side.**

## Changes

**Remove the 3 scopes**
- Drop the entries from `BLOCK_SCOPE_TO_OAUTH_BIT` (the authoritative registry) + `SCOPE_DESCRIPTIONS` (mod-review / installed UI) + the `scopes` enum in `public/schemas/app-block/v1.json` (order preserved so the schema⇄registry drift-guard stays green).
- Because `isKnownBlockScope` derives from the registry, a **new/edited** manifest declaring one of these is now rejected as unknown (intended deprecation), and the `enforceContextBinding` runtime cases become unreachable — removed (deny-by-default `default` retained). `media:read:owned` also dropped from the dev / tunnel / mod-review mint allowlists.
- **No OAuth bit orphaned:** `TokenScope.MediaRead` is unchanged and still backs ~80 tRPC media routes via `requiredScope`; it simply no longer maps to a block scope. The settings scopes used `SKIP_OAUTH_CHECK`, not a bit.

**Fix audit-log that asserted an unchecked scope**
- `updateUserSettings` logged the invocation as `scope: 'block:settings:write'` although the token was **never checked** for it (the write is gated by valid-token + app-developer + installer resolution). Relabel the audit row to `user-settings:write` (matching `endpoint`) so it stops asserting a scope that was never verified. No new enforcement gate added.

**Fix misleading mod-review tooltip**
- Was "…enforced at token issuance." Corrected to state scopes are enforced **per-operation server-side** — every capability re-verifies the token and checks the required scope before it runs.

## Legacy apps degrade gracefully (no break)

[...]

PR #3574, opened 2026-08-03
fix(app-blocks): humanise the analytics "Top endpoints" table instead of printing raw tokens

ZacxDev:
## Why

#3561 bounded `block_scope_invocations.endpoint` so the `topEndpoints` rollup could actually aggregate. That fixed the rollup — and as a side effect promoted the internal tokens to the **top rows** of the App analytics panel's "Top endpoints" table. `workflow:submit` and `storage:set` went from a count-1 long tail to the most prominent strings on the panel, rendered verbatim at `AppAnalyticsPanel.tsx:343`.

## What

`endpointBucketLabel`, used at that render site:

| bucket | label |
|---|---|
| `workflow:submit` | Generation submits |
| `storage:set` | App storage writes |
| `storage:delete` | App storage deletes |
| `user-settings:write` | User settings writes |
| `workflow:submit:wf_aaa` | Generation submits (wf_aaa) |
| `/api/v1/...` | *unchanged* |

The map covers exactly the four `endpoint:` literals the server writes to this table, enumerated from the `recordScopeInvocation` call sites in `blocks.router.ts` / `apps.router.ts`. Anything else is a REST path from `normalizeEndpoint(req.url)` and passes through untouched.

**Legacy rows:** #3561 ran no data migration, so a range covering older activity still yields per-id/per-key buckets. Those get the operation label with the **tail kept**, so two legacy buckets stay distinguishable. Counts are deliberately *not* merged — this labels buckets, it doesn't re-aggregate; merging would misreport what the query returned.

## It deliberately does NOT reuse `humaniseScopeEndpoint`

That function is exported from `AppActivityPanel` and looks like the obviously-right one — the follow-up note that prompted this change suggested it. But it's the per-**row** labeller: it resolves an operation's id out of that row's `detail`, and an aggregate bucket has no `detail`. Measured against the real function, not assumed:

```
humaniseScopeEndpoint('workflow:submit')     === '(no workflow id)'
humaniseScopeEndpoint('user-settings:write') === ''             // a BLANK table cell
humaniseScopeEndpoint('storage:set')         === 'storage:set'  // adds nothing
```

All three are worse than the raw token. The tests call the **real** function to pin those outputs, so this reasoning can't quietly rot into "why didn't they just reuse it".

The helper lives in its own module so the pure assertions run in the `unit` project — importing the panel would drag in Mantine + chart.js and force them into the browser tier.

## Verification

**13 unit tests** on the pure function, with a **mutation sweep of 5**, each verified to have actually applied before its result was read (one mutation's `perl` aborted on a quoting error and was correctly reported as "did not apply" rather than counted as survived):

| mutation | red |
|---|---|
| no-op label (return input) | 11 |
| drop the legacy-tail branch | 3 |
| wrong label literal | 4 |
| legacy tails collapse to one label | 3 |
| blank `user-settings` label | 3 |

Baseline and restored: 0 failed.

**2 browser tests** pin the *wiring* — that the render site actually calls the helper, a helper nothing calls being the failure mode worth guarding. Negative control: reverting the render site to `{e.endpoint}` turns exactly those 2 red and leaves the 3 pre-existing panel tests green.

- Full unit suite: **756 files / 10997 passed / 1 skipped / 0 failed**
- `tsc --noEmit`: **0 errors**, no OOM (a bare `tsc` without `--max_old_space_size=8192` OOMs and reports a false 0)
- Prettier: clean on all 4 files, confirmed with a positive control — a known violator in the same invocation does warn (`--check` prints "All matched files use Prettier code style!" even when it matched *zero* files, so the bare green isn't trustworthy on its own)

⚠️ **Caveat:** the browser tier ran against a **non-canonical browser build**. Playwright's vendored `chrome-headless-shell` is a generic-linux binary NixOS cannot exec, so the run used a nixpkgs `playwright-chromium-headless-shell` (Chrome for Testing 149) via a scratch `PLAYWRIGHT_BROWSERS_PATH` overlay. CI's component tier is the canonical check for those 2 tests.


ZacxDev:
## Audit round applied in `8e75826616`

The audit found **no 🔴**, and confirmed every quantitative claim in the description by execution (13 unit tests, the 756/10997/0-failed suite, tsc, and prettier with a positive control) plus corroborated the mutation sweep arithmetically. But it found one thing I'm glad it caught before merge, and three places my own reasoning was aimed at the wrong target.

### The one that mattered: I shipped a known-wrong label *with a test certifying it*

The legacy `workflow:submit:pending` bucket rendered as `Generation submits (pending)`. **`pending` is not a status.** Pre-#3561 the writer was `` `workflow:submit:${snapshot.workflowId || 'pending'}` ``, so it's the stand-in for *"no id was captured"* on an **otherwise completed** submit. The label told an author those submits were still in flight — the opposite of the truth.

Worse: `humaniseScopeEndpoint` **already** carves out this exact sentinel, with a comment saying why (*"'pending' was the historical stand-in for 'no id yet' and is not a real workflow id"*). So two functions in one codebase disagreed about one sentinel, and the wrong one had a test pinning it as intended — which is what would have made it stick. Now `Generations (no id)`, with the shared premise asserted against the **real** sibling function so they can't drift apart silently.

### Three places my argument was aimed at the wrong thing

- **I rebutted the wrong function.** The module doc argued at length against reusing `humaniseScopeEndpoint` (the *Detail*-column labeller). The genuine near-duplicate is **`humaniseScopeInvocation`** (the *Action*-column labeller) — prefix-based, needs no `detail`, already maps all four tokens — which the PR never mentioned. It's still unsuitable, but for two *different* reasons, now stated and tested: wrong register for a count ("Generated an image … 245"), and no arm for a REST path, where it returns `''` — a blank row. Arguing only against the easy candidate left the central claim untested.
- **The completeness claim gated nothing.** My test hardcoded the same four strings the map contains, with a comment calling itself "the reminder to add one". A fifth `endpoint:` literal in the router would have left it green and rendered raw. Replaced with a **drift guard** that greps the `recordScopeInvocation` call sites (mirroring `block-scope.schema-drift.test.ts`), plus a positive control so a rotted pattern can't pass as an empty set. Mutation-proven: removing a label now fails it.
- **The pass-through comment overstated its safety.** It said REST paths are "already readable". Now it says what's true — `normalizeEndpoint` templates only numeric/ULID segments and the row is written on `res.on('finish')` regardless of status, so the value is partly caller-influenced (React escapes it; the cell clamps).

### Also fixed

- **A latent non-string return.** `Record<string, string>` index is prototype-reachable: `labels['constructor']` returned a **function**, the truthy check passed, and React would have been handed a non-string child (throws, kills the panel). Not reachable from any current writer, so the signature was lying rather than the code failing. Now a `Map`, with the prototype keys pinned.
- **Vocabulary aligned**, per your call: the labels were a *fourth* way of naming these operations, so one event read "Generated an image" on Activity and "Generation submits" on Analytics. Now the pluralised existing nouns — **Generations / App-local storage writes / deletes / Block settings saves**.
- **The adjacent card too**, per your call: `topScopes` rendered raw `ai:write:budgeted` and the middleware's literal `(any-token)` placeholder right beside the humanised endpoints — which moved the inconsistency *inside one screen* instead of removing it. Added `scopeBucketLabel`, reusing the shared `READ_SCOPE_LABELS` registry rather than copying it. Module renamed `analytics-bucket-labels.ts` since it now serves both cards.

### Verification

29 unit tests (26 behaviour + 3 drift) and 7 browser tests. **6-mutation sweep**, each verified to have actually applied, both files confirmed byte-identical to their backups afterwards:

| mutation | red |
|---|---|
| `pending` surfaced as a status | 1 |
| old vocabulary restored | 4 |
| no-op label | 9 |
| drop legacy-tail branch | 3 |
| scope label → raw token | 1 |
| **remove a label** | **2 (incl. the drift guard)** |

Full unit suite **757 files / 11013 passed / 1 skipped / 0 failed**; `tsc --noEmit` 0 errors, no OOM; prettier clean on all five files (positive-controlled).

### Deliberately not done

Per your call, the Go CLI's `civitai app metrics` keeps printing these tokens raw, so it now disagrees with the web. A third vendored mirror is an ask-first cost under the CLI's `AGENTS.md`; I'm recording the divergence there instead so it reads as deliberate rather than as an oversight.

Two limits carried forward: the browser tier ran against a **non-canonical** nixpkgs Chrome 149 (CI's component tier is the authority for those 7 tests), and this branch is behind `main` — I have not built the merged tree.

ZacxDev:
## Delta re-audit round applied in `928273e4dd`

The re-audit found **no 🔴** and reproduced every number the previous round claimed. But it caught that round's **headline being false**, which I'd rather state bluntly: it said *"the adjacent card is fixed too"*, and two live scopes still rendered as raw internal tokens.

### The miss

`collections:write:self` (`pages/api/v1/blocks/collections/[id]/follow.ts`) and `apps:storage:shared:write` (`pages/api/v1/blocks/shared-storage/increment.ts`) are `withBlockScope` `requiredScope`s, so the middleware writes them to `block_scope_invocations.scope` **with `claims.appBlockId` set** — and `topScopes` filters on `appBlockId` only, so both reach the card. One is in `SENSITIVE_BLOCK_SCOPES`. I confirmed independently: exactly **7** `requiredScope` literals exist under `src/pages/api`, and my maps covered 5.

### Why it was missable — and the structural fix

I'd added a drift guard for **endpoints** and none for **scopes**. So adding a `requiredScope` to a route could never fail a test. That asymmetry *is* the bug behind the bug.

There's now a scope drift guard reading the same sources the middleware and routers actually use (`requiredScope:` literals under `src/pages/api`, the routers' `scope:` arguments, the `(any-token)` default), with a positive control. Mutation-proven: dropping either new label now turns **2** tests red — the literal pin *and* the drift guard.

I also gated the **writer file set**, not just the literals. The endpoint guard's two-file scope rested on a fact about today (the only other `recordScopeInvocation` callers write a `normalizeEndpoint` path, or rows with no `appBlockId`), so a third caller would have left every test green.

### Also fixed

- **Colliding labels, contradicting this module's own policy.** `user-settings:write` and `block:settings:write` (the pre-#3212 name for the same operation) both mapped to "Block settings saves", so a range spanning #3212 showed two visually identical rows with different counts — exactly what the endpoint card's legacy-tail handling and its test explicitly forbid. The legacy one is now `Block settings saves (legacy scope)`.
- **The register argument was self-undercutting.** I rejected `humaniseScopeInvocation` for naming a single past event rather than a plural noun — but `scopeBucketLabel` returns the shared `READ_SCOPE_LABELS` strings verbatim for all 8 read scopes, which are imperative permission phrases. The card *does* mix registers. Now stated as the deliberate trade it is, with the argument scoped to the write half.
- **Removed an `as string` cast** that reintroduced the exact type dishonesty this module's `Map` exists to remove.
- **Fixed the twin.** `humaniseScopeInvocation` had the identical prototype-reachable index I'd loudly fixed here — `READ_SCOPE_LABELS[scope] ?? …` returned `Object.prototype.constructor` (a function) for scope `'constructor'`, because `??` doesn't reject a non-nullish hit. It shares the registry, so my own reasoning applied verbatim.
- **A test claimed to pin the shared registry** while only asserting `not.toBe(raw)` — which a private hardcoded copy satisfies too. Now asserts identity against `READ_SCOPE_LABELS`; a mutation forking one label turns it red.
- Corrected three comment overstatements ("roughly half", "regardless of status", and the `pending` gloss — such a row genuinely *might* have been in flight, so the objection is that the field never encoded status at all, not that it was wrong about status).

### One mutation SURVIVED, and I'm not papering over it

Dropping the `!base` guard leaves all 36 tests green. Reaching it requires `TAILED` and `ENDPOINT_LABELS` to disagree, and both live two lines apart in one file — so it is an **unreachable defensive branch with no coverage**, and it's labelled that way in the code. Kept because it's free and the alternative is a lying cast, but it is not tested and shouldn't be counted as such.

### Verification

36 unit tests (26 behaviour + 10 drift), 18 browser tests across both panels. 6-mutation sweep, each verified to have actually applied, both files byte-identical to backups afterwards:

| mutation | red |
|---|---|
| drop `apps:storage:shared:write` label | 2 |
| drop `collections:write:self` label | 2 |
| collide the two settings labels | 1 |
| fork a read label into a private copy | 1 |
| revert the AppActivityPanel twin | 1 |
| drop the `!base` guard | **0 — survived (see above)** |

Full unit suite **757 files / 11020 passed / 1 skipped / 0 failed**; `tsc --noEmit` 0 errors, no OOM. Prettier clean in every region I touched — `AppActivityPanel.tsx`'s 4 violations are pre-existing, proven by the hunks being identical to main's shifted by exactly the **+8** lines this change adds.

**Unproven:** the auditor reported eslint clean but **without** a positive control, so treat that as unverified — CI's lint step is the authority. And the browser tier still ran against the non-canonical nixpkgs Chrome 149.

ZacxDev:
## Retraction: I claimed CI doesn't run the component tier. It does — and this PR's runs are RED.

Across this PR's comments I stated that CI does not run the `component` (browser) project, and used that to frame the 7 browser tests here as having no CI coverage. **That was false.** The `preview / component-tests` check runs exactly that project, and on this PR:

| commit | `preview / component-tests` |
|---|---|
| `075519d380` — before the audit-fix round | success |
| `8e75826616` — first fix round | **failure** |
| `928273e4dd` — second fix round (merged) | **failure** |

It is report-only (`"Component suite failed (report-only, not blocking)"`), which is why nothing stopped the merge — but I should have seen it and did not.

**How I got it wrong**, since the mechanism matters more than the fact: every status query I ran filtered for `Unit tests|Typecheck|ESLint|event-engine`. `preview / component-tests` was therefore never in a result set, and I read its absence from my own filtered output as evidence it did not exist. I selected the evidence and then concluded from the selection, then repeated the conclusion in three places without checking it directly.

## What I can and cannot establish

**Cutting against "this PR broke it":** PR 3591 **passes** `component-tests` on a base that contains this merge, and PR 3594 **fails** on the same base. Green for some PRs, red for others on identical base content points at flakiness or content-dependence rather than a defect introduced here.

**Cutting toward it:** the transition from success to failure lines up exactly with the fix round that changed the labels and added 3 browser tests. That is a real correlation and I am not going to explain it away.

**What I could not do is settle it locally, and I checked that carefully rather than assuming.** The full-project component run does not complete on this NixOS host in either direction:

- with this merge → crashes, `Browser connection was closed while running tests`
- **without** this merge (`e4d23679c2^`) → times out at 25 minutes

I ran the without-merge baseline specifically to test whether the local failure was attributable to this change. It is not — it fails both ways. Single-file runs pass cleanly, cold `optimizeDeps` cache included (7 tests). So a local run here can validate one file and tells you nothing about the suite. **The preview pipeline's own logs are the only authority**, and I can't reach them.

## Two by-products worth keeping

- **The cold-`optimizeDeps` flake is fixed.** `lint.yml`'s `unit` job excludes browser tests because they "carry the cold-optimizeDeps flake documented at `vitest.config.mts:98-124`" — but that section documents the *fix*, and a cold-cache single-file run passes. The stated reason is stale.
- I had planned a PR adding the component tier to CI. **That premise is dead** — the tier isn't missing. A smaller change (add it to GitHub Actions alongside the preview pipeline, fix the stale comment) may still be worth doing, but not while a suite I touched is showing red.

Anyone with access to the preview pipeline's component-suite output: that log is what's needed to close this. If it turns out these tests are the cause, the fix is on me and I'll do it forward.

PR #3576, opened 2026-08-03
fix(tests): make seven server suites' exit codes describe their tests

ZacxDev:
> **Body corrected after an adversarial audit falsified the original premise.** The first version claimed a repo-level defect; it is environment-scoped. Full before/after and the measurement matrix are in [this comment](https://github.com/civitai/civitai/pull/3576#issuecomment-5170883899).

Seven suites under `src/server` exit non-zero while 100% of their tests pass — **in a worktree without the repo's flake dev-shell**. Each tests a pure function but reaches it through a module importing `~/server/db/client` at module scope, which instantiates Prisma. When `PRISMA_QUERY_ENGINE_LIBRARY` is unset, that leaves an unhandled `PrismaClientInitializationError` which fails no test and only sets the exit code.

## Scope — read this before the rest

`flake.nix:62` exports `PRISMA_QUERY_ENGINE_LIBRARY` and `.envrc` is `use flake`, but **`.envrc` is gitignored, so a bare `git worktree` never receives it.** Measured, all 7 suites run individually:

| | base `b83efe4dcd` | head `3bd37856ee` |
|---|---|---|
| **no engine env** (bare worktree) | `rc=1` ×7 | `rc=0` ×7 |
| **with flake dev-shell** | **`rc=0` ×7** | `rc=0` ×7 |

Test counts identical in all four cells: 8/15/6/2/1/3/4 = 39.

**CI is unaffected** — `grep -c PrismaClientInitializationError` over this PR's `Unit tests` log is 0 across 768 files, because `postinstall` runs `prisma generate`.

So the root-cause fix for anyone running suites in a worktree is to provision it (`cp <repo>/.envrc <worktree>/ && direnv allow <worktree>`). **This PR is hardening on top of that, not a substitute** — and it covers these 7 suites, not the class.

## Why it is still worth landing

Mutation sweeps certify spend-path guards here by reading `rc`, and tooling routinely runs in bare worktrees — which is how this symptom was observed twice, independently. In that environment `rc` was already 1 before any mutant, so every mutant read as "killed". The change is free, additive, and makes these suites robust to an environment the repo does not guarantee.

## The fix

One `vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }))` per file, matching the **300** sibling suites (~120 factory shapes) that already do it; this exact idiom has 27 prior uses.

`dbRead` and `dbWrite` are the **only** names any module imports from that path — all 446 import/export-from statements enumerated, no namespace import, no destructuring, no side-effect import — so the two-key factory is complete rather than merely sufficient. The 13 importers using the relative `../db/client` collapse to the same resolved mock-registry key.

The clients are left empty deliberately: nothing here touches a database. Note the precise property — an empty object makes a **call-shaped** use throw (`dbRead.user.findMany()`), but a bare property read yields `undefined` silently.

## Verification

**Each mock is load-bearing.** The `base × no-env` row *is* the seven-way mutation: with the `vi.mock` absent all seven return `rc=1` with exactly one `PrismaClientInitializationError` each while their tests still pass — each dies for *this* guard's specific reason, not incidentally.

**Nothing is masked.** Head-with-env runs these suites against the **real** Prisma client and head-without-env against `{}`; both produce byte-identical pass counts, so nothing in the graphs distinguishes them. Per-file review confirmed no assertion depends on the client.

**Safe under `{}`.** Of 24 module-scope `dbRead`/`dbWrite` references repo-wide, only 2 fall inside the union of the seven import graphs (`db-helpers.ts:18` `makeDbKV(dbWrite)`, `tag.service.ts:50` `queryCache(dbRead, …)`), both lazy by construction. The 11 genuinely eager reads in `jobs/entity-moderation.ts:185-233` are outside all seven graphs.

**Gates.** Typecheck delta base→head is **zero** (134 errors at base, 134 at head, identical sets under `diff`). ESLint: `ecosystems.test.ts` 9 pre-existing `no-explicit-any` at both base and head; other six clean. Prettier diffs on two files are pre-existing at base.

**CI redness is pre-existing**, from the `feat/db-queries-kysely` merge: `packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769` plus a `@prisma/client` postinstall schema-path failure. PR #3574 fails identically; pre-merge #3572 passes both.

## Known forward-risk (not introduced here)

This is a wholesale mock, so adding any new export to `~/server/db/client` breaks every file using the shape — probed, and **5 of 7 collapse to 0 tests collected**. `local-rules/no-wholesale-module-mock` is scoped to `['~/utils/trpc']` and does not cover this path, and `importOriginal` cannot be used (the spread re-evaluates the real module and the rejection returns). Recorded rather than solved.

### DIFF
--- a/src/server/__tests__/update-user-score-image-score.test.ts
+++ b/src/server/__tests__/update-user-score-image-score.test.ts
@@ -1,4 +1,13 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
+
+// 🔴 NOT ABOUT THIS FILE'S ASSERTIONS — it is about its EXIT CODE. The code
+// under test is pure, but it is reached through a module that imports
+// `~/server/db/client` at module scope, which instantiates Prisma and leaves an
+// UNHANDLED rejection. That rejection fails no test; it only makes the runner
+// exit non-zero, so `rc` here did not describe the tests and any mutation sweep
+// reading it reported every mutant killed. Nothing below touches a database.
+vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }));
+
 import {
   rollupImageScoreDeltas,
   foldImageDeltasOntoStored,
--- a/src/server/__tests__/update-user-score-image-score.test.ts
+++ b/src/server/__tests__/update-user-score-image-score.test.ts
@@ -1,11 +1,10 @@
 import { describe, expect, it, vi } from 'vitest';
 
-// 🔴 NOT ABOUT THIS FILE'S ASSERTIONS — it is about its EXIT CODE. The code
-// under test is pure, but it is reached through a module that imports
-// `~/server/db/client` at module scope, which instantiates Prisma and leaves an
-// UNHANDLED rejection. That rejection fails no test; it only makes the runner
-// exit non-zero, so `rc` here did not describe the tests and any mutation sweep
-// reading it reported every mutant killed. Nothing below touches a database.
+// Mocked for the EXIT CODE, not the assertions. Reached transitively, this
+// module instantiates Prisma — and in a worktree without the repo's flake
+// dev-shell that leaves an unhandled rejection which fails no test but still
+// sets rc=1, so a mutation sweep reading rc scores every mutant as killed.
+// Nothing below touches a database. See civitai#3576.
 vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }));
 
 import {
--- a/src/server/games/daily-challenge/challenge-currency.test.ts
+++ b/src/server/games/daily-challenge/challenge-currency.test.ts
@@ -1,11 +1,10 @@
 import { describe, it, expect, vi } from 'vitest';
 
-// 🔴 NOT ABOUT THIS FILE'S ASSERTIONS — it is about its EXIT CODE. The code
-// under test is pure, but it is reached through a module that imports
-// `~/server/db/client` at module scope, which instantiates Prisma and leaves an
-// UNHANDLED rejection. That rejection fails no test; it only makes the runner
-// exit non-zero, so `rc` here did not describe the tests and any mutation sweep
-// reading it reported every mutant killed. Nothing below touches a database.
+// Mocked for the EXIT CODE, not the assertions. Reached transitively, this
+// module instantiates Prisma — and in a worktree without the repo's flake
+// dev-shell that leaves an unhandled rejection which fails no test but still
+// sets rc=1, so a mutation sweep reading rc scores every mutant as killed.
+// Nothing below touches a database. See civitai#3576.
 vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }));
 
 import { NsfwLevel } from '~/server/common/enums';
--- a/src/server/games/daily-challenge/challenge-currency.test.ts
+++ b/src/server/games/daily-challenge/challenge-currency.test.ts
@@ -1,4 +1,13 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
+
+// 🔴 NOT ABOUT THIS FILE'S ASSERTIONS — it is about its EXIT CODE. The code
+// under test is pure, but it is reached through a module that imports
+// `~/server/db/client` at module scope, which instantiates Prisma and leaves an
+// UNHANDLED rejection. That rejection fails no test; it only makes the runner
+// exit non-zero, so `rc` here did not describe the tests and any mutation sweep
+// reading it reported every mutant killed. Nothing below touches a database.
+vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }));
+
 import { NsfwLevel } from '~/server/common/enums';
 import { nsfwBrowsingLevelsFlag, sfwBrowsingLevelsFlag } from '~/shared/constants/browsingLevel.constants';
 import { ChallengeSource } from '~/shared/utils/prisma/enums';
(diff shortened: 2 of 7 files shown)


ZacxDev:
## CI is red, and none of it is this PR — attribution measured, not asserted

`Typecheck` and `Unit tests` both fail here. Both are **pre-existing, repo-wide breakage** that landed on `main` earlier today. Evidence:

**Typecheck** — the failure is byte-identical on this branch and on the unrelated `feat/account-deletion-image-removal`:

```
node_modules/@prisma/client postinstall: Error: Could not load schema from
  `packages/civitai-db-schema/prisma/schema.prisma` provided by "prisma.schema"
  config of `package.json`: file or directory not found
packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769:
  No overload matches this call.
```

Neither file is touched here. `lint.yml` runs flipped green→red between 18:11 and 18:13 UTC, across at least four unrelated branches (`moderator-app-pages`, `zach/analytics-humanise-endpoints`, `feat/account-deletion-image-removal`, `feat/db-queries-kysely`) — the window of the `feat/db-queries-kysely` merge, which is also what introduced `@civitai/db-queries`.

**Unit tests** — `Test Files 2 failed | 766 passed (768)`, `Tests 9 failed | 11250 passed | 1 skipped`. The two failing files are `src/__tests__/pages/api/mod/adjust-tag-level.test.ts` and `src/server/services/__tests__/get-models-raw.transient-503.test.ts`. Neither is one of the seven. Both also fail on `feat/account-deletion-image-removal`.

**The seven files in this PR all PASS in CI** — grepped the CI log for every one of them against the `FAIL` lines: no hits, so they are inside the 766 passed. That is the cross-check that matters here, because CI has a correctly generated Prisma client where my local environment does not.

`ESLint + Prettier (changed files)` and `event-engine-common pin` are both **SUCCESS**.

So: this PR is green on everything it can be green on, and the two red checks are a `main` problem that wants its own fix.

ZacxDev:
## 🔴 Correction — my premise was wrong, and an adversarial audit caught it

The original body claimed a repo-level defect ("Seven suites under `src/server` exited non-zero"). **That is not accurate.** Posting as a comment rather than only editing the body, since reviewers may have read the stronger claim.

**What is actually true.** The unhandled `PrismaClientInitializationError` ("could not locate the Query Engine for runtime `linux-nixos`") occurs only when `PRISMA_QUERY_ENGINE_LIBRARY` is unset. `flake.nix:62` exports it and `.envrc` is `use flake` — but **`.envrc` is gitignored, so a bare `git worktree` never receives it.** That is where I diagnosed this.

Full matrix, all 7 suites run individually:

| | base `b83efe4dcd` | head `3bd37856ee` |
|---|---|---|
| **no engine env** (bare worktree) | `rc=1` ×7 | `rc=0` ×7 |
| **with flake dev-shell** | **`rc=0` ×7** | `rc=0` ×7 |

Test counts identical in all four cells (8/15/6/2/1/3/4 = 39). The bottom-left cell is the falsification: **the unmodified base files already exit 0 once the repo's own dev-shell variable is present.**

CI is unaffected — `grep -c PrismaClientInitializationError` over this PR's `Unit tests` log is **0** across 768 files, because `postinstall` runs `prisma generate`.

**Why keep the change anyway.** The root-cause fix for the mutation-sweep workflow is provisioning the worktree (`cp <repo>/.envrc <worktree>/ && direnv allow <worktree>`), and this PR is hardening on top of that, not a substitute. It is free, it makes these suites robust to an environment the repo does not guarantee, and tooling does routinely run in bare worktrees — which is exactly how this was observed twice. But it fixes **7** suites, not the class: any other suite whose import graph reaches the real client is still affected in that environment.

**Two smaller corrections to the body:**
- *"an empty object turns an accidental call into a loud TypeError rather than a silent `undefined`"* — true only for **call-shaped** uses (`dbRead.user.findMany()` throws). A bare property read (`const t = dbRead.user`) yields `undefined` silently. Doesn't affect these seven; the unqualified absolute was wrong.
- *"274 sibling suites"* → the real count is **300**, across ~120 distinct factory shapes. This exact idiom has 27 prior uses.

**What survived the audit unchanged**, independently re-derived rather than taken on trust:
- `dbRead`/`dbWrite` are the only names imported from this module — all **446** import/export-from statements enumerated; the 13 relative `../db/client` importers collapse to the same resolved mock key, so they are not a hole.
- Empty `{}` is safe here: of 24 module-scope references repo-wide, only 2 fall inside the seven graphs and both are lazy by construction. Empirically, head+env (real Prisma client) and head+noenv (`{}`) produce byte-identical pass counts.
- All seven mocks are load-bearing — the `base × noenv` row *is* the seven-way mutation, each failing with exactly one `PrismaClientInitializationError` while its tests still pass.
- **Typecheck delta base→head is zero** (134 errors at base, 134 at head, identical error sets under `diff`). The PR's CI redness is entirely pre-existing (`updated-at-tables.ts(7,38): error TS2769` + the `@prisma/client` postinstall schema-path failure); PR #3574 fails identically and pre-kysely-merge #3572 passes both.

## 🟡 One forward-risk worth recording

The 2-key factory is provably complete **today**, but it is a wholesale mock: adding any new export to `~/server/db/client` breaks every file using this shape. Probed by adding a real available name (`createPrismaClients`, reachable via `client.ts`'s `export * from '@civitai/db/client'`) to a module in the graphs — **5 of the 7 collapsed to 0 tests collected**, each with `No "createPrismaClients" export is defined on the … mock`.

This is what `local-rules/no-wholesale-module-mock` exists to prevent, and `.eslintrc.js:96` scopes it to `['~/utils/trpc']` only — so it does not cover this path. `importOriginal` cannot be used here (verified: the spread re-evaluates the real module and the rejection returns). The failure is loud (`rc=1`), but `Unit tests` is `continue-on-error: true`, which is precisely the condition under which a suite silently dropping to 0 tests persists.

Not blocking, and not introduced by this PR — recorded so the next person has it.

Issue #3579, opened 2026-08-03
adjust-tag-level.test.ts collects 0 tests — incomplete wholesale mock of ~/server/db/pgDb


ZacxDev:
`src/__tests__/pages/api/mod/adjust-tag-level.test.ts` **collects zero tests** and has been doing so silently. It fails at import, so nothing in the file runs — it contributes 0 assertions to CI while appearing in the file list.

## Reproduction

```
$ vitest run --project unit src/__tests__/pages/api/mod/adjust-tag-level.test.ts

Error: [vitest] No "pgDbRead" export is defined on the "~/server/db/pgDb" mock.
      Did you forget to return it from "vi.mock"?
 ❯ src/server/db/kyselyDb.ts:22:13
 ❯ src/server/db/db-lag-helpers.ts:2:1

 Test Files  1 failed (1)
      Tests  no tests
```

## Cause

The file mocks `~/server/db/pgDb` with a **one-key wholesale factory** (line 12):

```ts
vi.mock('~/server/db/pgDb', () => ({
  pgDbWrite: { query: mockQuery },
}));
```

`~/server/db/pgDb` has since grown `pgDbRead` / `pgDbReadLong`, and `kyselyDb.ts:22` reads `pgDbRead` at **module scope** (`readPool: pgDbRead`), reached via `db-lag-helpers.ts`. A wholesale factory replaces the module entirely, so any export it omits is simply absent — and the omission surfaces only when something imports the missing name.

This is drift, not a mistake at the time of writing: the factory was complete when authored and silently became incomplete when the module gained exports.

## Fix

Spread the real module in so new exports can't go missing:

```ts
vi.mock('~/server/db/pgDb', async (importOriginal) => ({
  ...(await importOriginal<typeof import('~/server/db/pgDb')>()),
  pgDbWrite: { query: mockQuery },
}));
```

This is the shape `local-rules/no-wholesale-module-mock` already encodes as the canonical pattern — that rule is configured in `.eslintrc.js` for `['~/utils/trpc']` only, so it does not currently cover `~/server/db/pgDb`. Widening its `modules` list to include the db modules would prevent recurrence.

## Notes

- **Not a Prisma issue.** It fails identically with and without `PRISMA_QUERY_ENGINE_LIBRARY` set, so any `PrismaClientInitializationError` alongside it is incidental noise, not the cause. It is specifically *not* an instance of the class fixed in #3576 — a `~/server/db/client` mock does not help here.
[...]

PR #3581, opened 2026-08-03
fix(app-blocks): make a dark-flag revenue response distinguishable from a genuine zero

ZacxDev:
Follow-up to #3557, which fixed this exact defect for `getMyAppAnalytics`.
`/apps/revenue` renders both panels, so until now that page still had one
fabricated-zero surface left.

## The bug

`getMyRevenue` short-circuited the dark `appBlocks` flag with `emptyRevenue()`:
HTTP 200, every bucket at 0, and no discriminator of any kind. That payload is
byte-identical to a legitimate "you are a publisher who has not earned anything
yet" response, so a client had nothing to branch on and rendered a clean,
plausible, entirely fabricated $0.00 earnings report.

An undiscriminated zero is worse than an error. An error is self-announcing: the
caller sees it, retries, or escalates. A fabricated zero is indistinguishable
from a measurement, so it is *believed* — a publisher reads "$0.00 confirmed"
and concludes their app earns nothing, when in truth we never ran the query. It
fails silently, in the direction of a confident wrong answer, on a money screen.

## One discriminator value, not two — deliberately

`getRevenueForOwner` never *computes* ownership; it scopes every aggregate with
`appOwnerUserId` in the WHERE clause. A caller asking for an app they do not own
therefore gets a zero-row aggregate, and those zeros are a truthful measurement
("your attributed revenue on that app is zero"), not a fabricated one. There is
no not-owned branch in the revenue path that could report `notOwned`, so adding
that value would declare a state nothing can produce and force an unreachable
branch into every renderer — the same "declared but never satisfied" trap that
let the analytics bug hide for months. Hence a single `notEntitled`. The
reasoning is recorded at the type definition.

The discriminator is baked into `emptyRevenue()` rather than passed by the
caller: that helper exists solely for the dark-flag short-circuit, so a future
second call site cannot silently omit it.

## A renderer actually refuses the zeros

The lesson from #3557 is that a DTO field nothing branches on is not a guard.
On main, `getMyRevenue` had TWO unguarded readers: `/apps/revenue` had a
page-local `RevenuePanel`, and `/apps/[appBlockId]/revenue` carried its own
near-duplicate copy of the same summary-card markup. One bug, two renderers, two
chances to forget the guard.

So the guard is consolidated rather than duplicated: the page-local panel is
extracted to `src/components/AppBlocks/RevenuePanel.tsx` (an extraction, not a
new component — `appBlockId` selects the scoped variant), both pages render it,
and the honest unavailable state matches `AppAnalyticsPanel`'s so the two panels
on `/apps/revenue` read consistently. `revenue-panel-wiring.test.ts` pins that
the panel is the ONLY consumer of the proc, so a future page cannot reintroduce
an unguarded reader.

Two user-visible consequences of consolidating, called out deliberately: the
per-app page's "Confirmed" card is now labelled "Confirmed (unpaid)" and its
"Voided" card gains the explanatory tooltip, both matching the portfolio page.

## Tests: regression vs invariant — measured, not asserted

Verified by reverting the implementation to origin/main while KEEPING the new
tests, and recording what actually went red.

Genuine REGRESSION coverage (red at origin/main):
  - emptyRevenue > flags the dark-flag placeholder as notEntitled
        red: expected undefined to be 'notEntitled'
  - emptyRevenue > DISCRIMINATOR: a genuinely-measured all-zero result is NOT flagged
  - wiring > RevenuePanel is the ONLY component that queries blocks.getMyRevenue
        red at main with TWO unguarded consumers
  - wiring > both revenue pages import and render the shared panel
  - wiring > neither page still renders revenue markup of its own

INVARIANT guard (passes on pre-change code — not regression coverage):
  - emptyRevenue > still zeroes every bucket

HARNESS POSITIVE CONTROL (not a guard on behaviour):
  - wiring > sanity: the tree walk found the panel and both pages
        without it, every "no other caller" assertion could pass vacuously

The 3 renderer tests in RevenuePanel.browser.test.tsx cannot be run against
origin/main at all, because the component they mount did not exist there. Their
regression character is established instead by mutation M3, which restores the
LITERAL pre-change render condition (`{data && (`, verbatim what both pages had)
and turns both dark-flag tests red. Presented as such, not as five independent
tests of the fix.

The DISCRIMINATOR tests are the load-bearing ones: a genuinely-owned,
genuinely-zero result must still render the real dashboard, with counters
byte-identical to the unavailable case and only `unavailable` differing. Without
them a later change that flagged everything would look correct.

## Mutation matrix

Every mutation checksum-verified as APPLIED before its test was read, and every
file re-checksummed against its backup afterwards (all restored byte-identical).

  M1  emptyRevenue drops the discriminator (the original bug)
      -> 2 failed | 22 passed  @ service test:600, :636
  M2  getRevenueForOwner ALSO flags (over-flagging a real measurement)
      -> 1 failed | 23 passed  @ service test:635
  M3  dashboard NOT suppressed when unavailable (pre-change condition)
      -> 2 failed | 1 passed   @ browser test:53, :63
  M4  the honest unavailable Alert never renders
      -> 2 failed | 1 passed   @ browser test:50
  M5  flags EVERYTHING unavailable
      -> 1 failed | 2 passed   @ browser test:75
  M6  /apps/revenue stops rendering the guarded panel
      -> 1 failed | 3 passed   @ wiring test:66
  M7  a NEW unguarded consumer of getMyRevenue appears
      -> 1 failed | 3 passed   @ wiring test:59

Each failed at its OWN guard's assertion line, not via a neighbouring assertion.

## Numbers observed

  unit project (full):  Test Files 91 failed | 678 passed (769)
                        Tests 9 failed | 10131 passed | 4 skipped (10144)
                        0 timeout panics
  origin/main baseline: Test Files 91 failed | 677 passed (768)
                        Tests 9 failed | 10124 passed | 4 skipped (10137)

  Failed-file sets are IDENTICAL (91 both sides); the delta is +1 file / +7
  tests, all mine. Every pre-existing failure traces to one cause: the workspace
  packages @civitai/db-queries / db-schema / kysely are not installed in this
  checkout, which is also why blocks.router.*.test.ts cannot load locally.

  typecheck: 150 errors, ZERO in any file I touched; the origin/main baseline is
  also 150. Harness positive-controlled — injecting one deliberate type error
  yielded exactly 151, with exactly 1 in RevenuePanel.tsx.

## Not verified

  - No router-level test. `emptyRevenue` is mocked at the module boundary in
    blocks.router.getMyAppAnalytics.test.ts, so a test there would assert the
    mock rather than the implementation; that file also cannot load in this
    checkout, and PR #3572 rewrites it. The authority for the discriminator is
    the service test against the REAL function.
  - The 3 renderer tests ran only locally, on a non-canonical Chrome for Testing
    (the vendored chrome-headless-shell cannot exec on this NixOS host). CI does
    NOT run the component tier at all, so these guards are local-only.
  - Neither revenue page was rendered in a browser; there are no page-level
    tests in this repo (pages import server-only helpers). The consolidation's
    visual equivalence is argued from the diff, not observed.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


ZacxDev:
## CI state on this PR — read before judging the red checks

**`Typecheck` (blocking) is red for a reason that is not in this diff.**

```
packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769: No overload matches this call.
```

That is the single error, in a **generated** file (`// This file was generated by a
custom prisma generator, do not edit manually`) that this PR does not touch, and
whose module graph this PR does not reach. Evidence it comes from `main`:

- `2233dd4b5d` / #3488 (`feat(account-deletion): remove user images on account
  self-deletion`) landed on `main` **after** this branch's base and added a
  migration under `packages/civitai-db-schema/` without regenerating the Kysely
  types, so `new Set<keyof DB>([...])` no longer typechecks.
- This branch does **not** contain that commit; the `pull_request` event
  typechecks the PR *merged into current main*, which does.
- **PR #3574**, which is unrelated and touches none of those files, fails
  `Typecheck` with the byte-identical error at the same file, line and code.
- Nothing in this diff is under `packages/`, and none of it imports
  `updated-at-tables`.

The fix is a schema-type regeneration on `main`, not a change here. Rebasing this
branch will not clear it — the failing commit is the one being merged in.

**`Unit tests` (report-only, `continue-on-error: true`) — the tests added by this
PR passed in CI:**

```
✓ unit src/server/services/blocks/__tests__/buzz-attribution.service.test.ts (24 tests)
✓ unit src/components/AppBlocks/__tests__/revenue-panel-wiring.test.ts       (4 tests)

Test Files  2 failed | 774 passed (776)
     Tests  9 failed | 11367 passed | 1 skipped (11377)
```

All 9 failures are in `src/server/services/__tests__/get-models-raw.transient-503.test.ts`
(MeiliSearch error-widening), the same pre-existing set observed locally, plus a
load failure in `src/__tests__/pages/api/mod/adjust-tag-level.test.ts`. Neither
file is touched here.

Worth noting the two-tier gap: CI ran **11,377** tests against **10,144** locally,
because this checkout is missing the `@civitai/db-queries` / `db-schema` /
`kysely` workspace packages. CI therefore exercised ~1,233 tests my local tier
structurally could not — the guards added here passed in **both** tiers.

**`ESLint + Prettier (changed files)` passed**, including the blocking added-files
Prettier step over all three new files.

**`preview / deploy`** is red; not investigated, and red on #3574 as well.

## Reminder of what is NOT covered

The three renderer guards live in `RevenuePanel.browser.test.tsx`, and CI does not
run the `component` project at all (`test:unit:run` is `--project unit`, whose
include is `src/**/*.test.ts`). Those three tests were verified only locally, on a
non-canonical Chrome for Testing. The CI-visible guard that the discriminator has
a live reader is `revenue-panel-wiring.test.ts`, which passed above.

ZacxDev:
## Correction to the verification numbers in the description above

I'm the coordinating session; I checked these against CI and **the local gate numbers in this PR's description came from a broken local environment, not a real baseline.** Correcting rather than leaving them to be trusted.

The description reports **91 failed test files / 678 passed** and **150 typecheck errors**, framed as matching an `origin/main` baseline. CI on this branch reports:

- **Unit tests: 2 failed / 768 passed (770)**
- **Typecheck: 1 error**

So the local run was producing ~89 spurious file failures and ~149 spurious type errors. Comparing against a same-worktree baseline was sound in principle — but 91 failing files is *itself* the anomaly, and normalising it as "baseline" concealed a broken harness rather than validating one. These local numbers should not gate this PR; CI's should.

Concretely: the local worktree had been provisioned for an earlier task and its installed/generated state was stale relative to the newer base it was rebased onto — which is why a sibling worktree on an older base reported 0 typecheck errors for the same code.

## What IS real, and it isn't this PR

`main` is genuinely broken right now, and this PR's red checks are inherited, not caused:

| check | failure | touched by this PR? |
|---|---|---|
| Typecheck | `packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769` — a generated file | no |
| Unit tests | `src/__tests__/pages/api/mod/adjust-tag-level.test.ts` | no |
| Unit tests | `src/server/services/__tests__/get-models-raw.transient-503.test.ts` (9 tests) | no |

Two independent PRs with disjoint diffs (#3574 and this one) fail those same three things identically, which points at the shared base rather than either diff. Both checks were **green earlier today** on #3574, so the breakage landed within the last few hours.

Worth flagging for whoever owns it: the `get-models-raw.transient-503` test names read *"the widening; fails against the narrow instanceof-only catch"* — i.e. they pin a widening the implementation currently lacks. That reads like tests landing without their implementation, or an implementation reverted from under its tests, rather than a flake.

## What this means for reviewing this PR

The CI-visible evidence for the change itself still stands and is worth reading on its own terms: the wiring test and the service tests **passed in CI**. But the full-suite and typecheck claims in the description need re-running once `main` is green, and I'd treat the mutation matrix as unconfirmed until then — it came out of the same harness.

Nothing has been merged, and I am not proposing this merge until main is fixed.

ZacxDev:
## Gate re-run on a fixed harness — and a correction to my own correction above

I re-ran everything after fixing the environment. **The numbers in the description were reproducible, not fabricated** — I reproduced 91 failed files / 9 failed tests exactly, in a *different* worktree. So my comment above was wrong to imply the problem was specific to one stale worktree. Setting that straight, and giving the numbers that actually gate this PR.

### Root cause of the 91 files

`Cannot find package '@civitai/db-queries' imported from src/server/db/kyselyDb.ts`

Main's Kysely merge added a **new workspace package**, so any `node_modules` installed before that merge fails at module load. That hits *every* worktree provisioned earlier — not one bad checkout. One `pnpm install --frozen-lockfile` fixes it.

### Before and after, same commit, same worktree

| | before install | after install | CI |
|---|---|---|---|
| Test Files | 91 failed / 678 passed | **2 failed / 767 passed** | 2 failed / 768 passed |
| Tests | 9 failed / **10131** passed | 9 failed / **11257** passed | 9 failed |
| Typecheck errors | 150 | **1** | 1 |

Local now agrees with CI. Note the passed count: **1,126 tests were not executing** before the install and the suite still reported a total — the missing package silently removed them. That is the more dangerous half of this, and it is worth anyone else with an older worktree knowing.

### The remaining failures are main's, not this PR's

- `src/server/services/__tests__/get-models-raw.transient-503.test.ts` (9 tests)
- `src/__tests__/pages/api/mod/adjust-tag-level.test.ts`
- Typecheck: `packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769` — a generated file

None is a file this PR touches. This PR's 8 files are the revenue service/router, the two revenue pages, `RevenuePanel.tsx`, and three test files.

### This PR's own tests, re-run clean

| suite | result |
|---|---|
| `buzz-attribution.service.test.ts` | 24 passed |
| `revenue-panel-wiring.test.ts` | 4 passed |
| `RevenuePanel.browser.test.tsx` | 3 passed |

⚠️ The browser tier ran against a **non-canonical** nixpkgs Chrome for Testing 149 (Playwright's vendored shell cannot exec on NixOS), and **CI does not run the `component` project at all** — so those 3 renderer guards have no CI coverage by anyone. The wiring test is the CI-visible guard, and it passes.

### What is still not re-verified

The **mutation matrix** in the description was produced on the broken harness. Its individual results are plausible — each mutation was checksum-verified as applied and each failure was attributed to its own guard's assertion line — but they were measured while 1,126 tests were not running, so I am not treating them as confirmed. Worth re-running before merge.

Nothing merged. Dispatching an adversarial audit next.

ZacxDev:
## Audit round applied in `ae6a6799ba`

The audit found **no 🔴** and independently confirmed the substance: the single-discriminator argument holds (`getRevenueForOwner` has no ownership probe, so `notOwned` is genuinely unproducible), `emptyRevenue()` has exactly one production caller so no existing path was broken, the `SummaryCards` extraction is **byte-identical** to main's, and the renderer genuinely suppresses rather than supplements. Both requested items are fixed, plus four cheaper ones.

### The two it asked for

**1. Restored parallelism on the per-app page.** `<RevenuePanel>` was mounted only in the *else*-branch of `myAppsQuery.isLoading`, so `getMyRevenue` couldn't start until `getMyApps` resolved — an extra round-trip plus five aggregates of latency on a money screen, and a silent regression against main, where both queries were issued on the same render. Now mounted unconditionally. Safe because ownership is enforced by the `ownerCheckDone && !thisApp` early return *and*, decisively, server-side: every aggregate is scoped by `appOwnerUserId`, so a non-owner's in-flight query can only return their own empty figures.

**2. The contract now has a CI-visible guard at the proc boundary.** The router test's `emptyRevenue` mock omitted `unavailable`, so the flag-OFF test asserted only zeroed buckets — nothing in CI observed the actual change. That gap mattered more than it looks: **CI runs no `component` project at all**, so the wiring test was the only CI coverage, and it guards the *consumer*, not the proc. Added the key, asserted it, and added the missing **discriminator** case (a measured result must NOT be flagged — without it, a change that flagged everything would satisfy the flag-OFF assertion and silently deny every publisher their real dashboard).

On the reason recorded for skipping it: the PR body said the file was unrunnable locally. **That was wrong** — it runs, 11 tests, and passes in CI. The real reason was a merge-conflict risk with #3572, which has since merged, so it no longer applied. My instruction to the authoring agent caused that; correcting it here.

### Four more from the same round

- **The wiring regex claimed more than it checked.** It matched only `trpc.blocks.getMyRevenue.useQuery`, evaded by `useSuspenseQuery`, `trpc.useQueries(...)`, and `createServerSideHelpers().…prefetch()`. Broadened to a call/access shape and scoped to client surfaces. Worth noting the naive fix — matching the bare name — hits **six** files including a comment this very commit adds, which is why the narrowing is there. Residual gaps (raw `/api/trpc` fetch, destructure-then-call) are now stated in the test rather than implied away.
- **A second union value would have rendered *wrong* copy, not a missing branch.** The panel tested truthiness and hard-coded the sentence while a docblock merely *asked* a future author to add a branch. Now an exhaustive `switch` whose `never` default makes widening the union a compile error until the copy exists — the posture `AppAnalyticsPanel` already takes.
- **`RevenuePayload` → `EmptyRevenuePayload`.** With `recentAttributions: []` and a required `unavailable`, it can only describe the placeholder; the old name invited typing the real path with it and being forced by TS to add `unavailable` to a genuine result — the inverse defect. Field docblock corrected too (it said "present ONLY when placeholder" on a type where it's mandatory).
- Positive control tightened from `>500` files to `>3000` (`src` holds ~4,400).

### Mutation matrix, re-run on a repaired harness

The original was measured while **1,126 tests were silently not executing** — a `node_modules` predating main's new `@civitai/db-queries` workspace package. Re-run after `pnpm install`:

| mutation | red |
|---|---|
| `emptyRevenue` drops the discriminator (the original bug) | 2 |
| **router strips `unavailable` on the dark path** | **1 — the guard added above** |
| dashboard not suppressed (pre-change condition) | 2 (browser) |
| Alert never renders | 2 (browser) |
| flags everything unavailable | 1 (browser) |
| a page stops rendering the shared panel | 1 |

Every mutation checksum-verified as applied; all four touched files confirmed byte-identical to backups afterwards.

### Gate

- This PR's suites: **40 unit + 3 browser**, all passing.
- Full unit suite: **767 passed / 2 failed files**, 11,258 tests passed. Both failures are main's known breakage (`get-models-raw.transient-503.test.ts`, `adjust-tag-level.test.ts`) — neither is a file this PR touches.
- `tsc --noEmit`: **exactly 1 error**, main's generated `updated-at-tables.ts`.
- Prettier: my regions clean. The service file's three hunks (@813/@1313/@1384) all sit before my edits at 1408+ and are pre-existing on main.

### Not covered by anyone

CI runs no `component` project, so the three renderer guards have **no CI coverage** and have only ever run against a non-canonical nixpkgs Chrome for Testing 149. That's structural, not something this PR can fix — but it's why item 2 above matters.

Nothing merged.

ZacxDev:
## Correction: my characterisation of main's other failing test was wrong

In an earlier comment I wrote that `get-models-raw.transient-503.test.ts`'s failures looked like *"tests pinning a widening the implementation currently lacks … tests landing without their implementation, or an implementation reverted from under its tests."*

**That was wrong, and I inferred it from test names instead of reading the code.** The widening is present: `model.service.ts:420` calls `isTransientMeiliError`, exactly as the test file's own header describes. The phrase *"the widening; fails against the narrow instanceof-only catch"* in those test names describes what the tests **pin**, not something missing.

The real cause is the same defect as [#3579](https://github.com/civitai/civitai/issues/3579) — an incomplete wholesale `vi.mock` of `~/server/db/pgDb`:

```
AssertionError: expected Error: [vitest] No "pgDbReadLong" export is defined on the
  "~/server/db/pgDb" mock … to match object { name: 'TRPCError' }
```

Both of main's currently-failing unit files are that one bug, in two different-looking shapes: `adjust-tag-level.test.ts` dies at import and collects 0 tests, while this one collects 9 and fails each assertion because the mock error surfaces inside the call under test. #3579's proposed fix — spread `importOriginal` into the factory, and widen `local-rules/no-wholesale-module-mock` beyond `['~/utils/trpc']` — closes both. I've noted that on the issue.

Nothing about this changes anything for #3581: those failures were, and remain, main's rather than this PR's, and none of them is in a file this PR touched. Correcting it because a wrong diagnosis on a red main is worse than no diagnosis — it sends whoever picks it up looking for production code that is already there.

ZacxDev:
## Retraction carried over from #3574

I stated in this PR's comments that CI does not run the `component` (browser) project, and used it to frame the 3 `RevenuePanel` renderer tests as having no CI coverage. **That was false** — `preview / component-tests` runs that project. The full retraction and the mechanism of the error are in https://github.com/civitai/civitai/pull/3574#issuecomment-5172447560 ; short version: every status query I ran filtered for `Unit tests|Typecheck|ESLint|event-engine`, so the check was never in my results and I read its absence as nonexistence.

Specific to this PR: `preview / component-tests` **did not report at all** on #3581's head — the preview pipeline appears not to have run the component stage here, so the 3 renderer tests are genuinely unverified in CI, just for a different reason than the one I gave. The wiring test (unit tier) did pass in CI and remains the CI-visible guard, as stated.

The rest of this PR's verification is unaffected: the unit-tier numbers, the re-run mutation matrix and the typecheck result were all measured on the repaired harness, and none of them depends on the component tier.

PR #3582, opened 2026-08-03
build(moderator): add the Dockerfile + release scripts so apps/moderator can ship

ZacxDev:
`apps/moderator` is the only app under `apps/` with **no Dockerfile and no `release:<app>` script**, so there is currently no way to produce an image for it. The in-cluster tag-webhook builds `apps/<name>/Dockerfile` on a `<prefix>-vX.Y.Z` tag — and nothing cuts that tag.

This is the prereq for standing up `moderator.civitai.com`. PR #3573 (Moderator app pages) adds the content; this adds the ability to ship it.

## The Dockerfile

Mirrors `apps/creator-studio/Dockerfile`'s proven four-stage shape — both are SvelteKit + `adapter-node` off the pnpm workspace:

1. **deps** — filtered frozen install (`--filter @civitai/moderator-app...`), `--ignore-scripts` to skip the root's Prisma `db:generate` (the spoke reads the DB through Kysely and never imports the Prisma client)
2. **build** — `vite build` → `apps/moderator/build`
3. **prod-deps** — `pnpm deploy --prod --legacy` for a flattened runtime `node_modules`
4. **runtime** — non-root (`sveltekit`, uid 1001, matching the k8s `runAsUser`) on `:3000`

The build stage sets throwaway `DATABASE_URL` / `DATABASE_REPLICA_URL`: `src/lib/server/db.ts` calls `createKyselyClients` with `required()` on both at module load, and SvelteKit's postbuild `analyse` step imports the server modules, so the build throws without them. They live only in that stage and never reach the runtime image. Redis, ClickHouse and the orchestrator client are all lazily constructed (`getRedis` / `getClickhouse` / `getOrchestratorClient`), so they need no placeholder.

## Version bump

`apps/moderator/package.json` `0.0.0` → `0.0.1`, matching the bootstrap tag, so the next `pnpm release:moderator` cuts `0.0.2` instead of colliding on an existing `moderator-v0.0.1`.

## Verification

Built the image against **both** this tree and the `apps/moderator` content on #3573 — both green. Then ran the container:

| Request | Result |
|---|---|
| `GET /favicon.svg` | **200** `image/svg+xml` (the k8s probe path — the only unauthenticated 200) |
| `GET /` | **302** → `auth.civitai.com/login?returnUrl=…` |
| `POST /api/mod/noop` unauthenticated | **401** |

Container came up clean (`Listening on http://0.0.0.0:3000`, 0 restarts).

No behaviour change to any existing app — this only adds files and three `package.json` scripts.


ZacxDev:
Heads-up on this PR's red checks — **they are inherited from `main`, not caused by this change.**

This PR is three files: `apps/moderator/Dockerfile`, three `release:moderator` scripts in the root `package.json`, and a version bump. None of them can reach `tsc` or the unit suite.

- **Typecheck** — `updated-at-tables.ts(7,38) TS2769`. Reproduced identically on a pristine `origin/main` checkout. Fixed in #3587.
- **preview / deploy** — its pipeline's own `typecheck` step, failing on the same single error. Same fix.
- **Unit tests** — `No "pgDbReadLong" export is defined on the "~/server/db/pgDb" mock` (9 tests / 2 files; 773 pass). Byte-identical signature on #3573, so it is also pre-existing on `main` and is **not** addressed by #3587.

So this should go green on Typecheck once #3587 lands; the unit-test failure is a separate pre-existing issue that still needs an owner.

PR #3583, opened 2026-08-03
fix(db-schema): register the @updatedAt generator so its output stops rotting

ZacxDev:
`main` is red on typecheck. This unbreaks it at the root and stops the same class from recurring silently.

## The error

```
packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769: No overload matches this call.
  ...
  Type '"BaseModelLicensingFee"' is not assignable to type 'keyof DB'.
```

## Diagnosis: a stale generated file that nothing could regenerate

`updated-at-tables.ts` opens with *"This file was generated by a custom prisma generator, do not edit manually."*

That was aspirational. #3575 landed the generator script (`scripts/prisma-updated-at-generator.mjs`) but **never added a `generator` block for it to `prisma/schema.full.prisma`**. `prisma generate` on `main` runs exactly four generators — `client`, `enums`, `typescriptInterfaces`, `kysely` — and not this one. The committed file was therefore a one-off snapshot of some older schema, with no path back to the schema it claims to track.

`BaseModelLicensingFee` was dropped on 2026-07-15 (`20260715150000_licensing_root_cleanup`: `DROP TABLE "BaseModelLicensingFee"`). The snapshot still listed it; `prisma-kysely` correctly stopped emitting it from `DB`; `keyof DB` did its job and the build went red.

Worth noting how invisible this was: #3575 also added `pnpm run db:check-generated` (`db:generate && git diff --exit-code -- packages/civitai-db-schema/src`). On `main` that command **passes** — because the generator never runs, `db:generate` can't produce a diff for a file it doesn't write. The drift gate was green over a broken file.

## The fix

1. Register the generator in `schema.full.prisma`, so `prisma generate` emits `updated-at-tables.ts` off the **same DMMF** as the `kysely` generator that produces `DB`. Consistency is now structural rather than a habit.
2. Commit the generator's real output.
3. Wire `db:check-generated` into the Typecheck CI job.

No `@ts-ignore`, no `as any`, no exclusion — the type error was correct and reported a genuinely wrong value.

## Runtime change to `UPDATED_AT_TABLES` (80 → 84)

| | table | why |
|---|---|---|
| − | `BaseModelLicensingFee` | table dropped from the DB 2026-07-15 |
| + | `LicensingRoot` | has `@updatedAt`, was missing from the snapshot |
| + | `MembershipGift` | " |
| + | `PaidAccess` | " |
| + | `UserMembershipOverride` | " |
| + | `app_review_agent_reports` | " |

The five additions are the more interesting half: each carries `@updatedAt` in the schema, so an `updatedAtPlugin`-mediated Kysely `UPDATE` to them **should** have been auto-stamping `updatedAt` and wasn't.

**No live behaviour changes today.** The only production Kysely `UPDATE` running through the plugin is `tag.db.ts` → `Tag`, which is in both the old and new set. The `apps/*` Kysely clients (auth, creator-studio, moderator) don't install `updatedAtPlugin` at all, so nothing there is affected either. This corrects the contract before anything depends on it.

Independently verified: a separate parser derived the `@updatedAt` set straight from `schema.prisma` (handling `@@map` and comment-stripping) and matched the generator's output exactly — 84/84, 0 missing, 0 extra. Run against `origin/main`'s file, the same parser reports the 5 missing + 1 extra above, so it can tell the two apart.

## Why the CI step, and not just typecheck

Typecheck only sees drift in **one** direction. An entry that stops being `keyof DB` is a compile error (that's this bug). An entry **missing** from the set is a perfectly legal subset — it just quietly stops stamping `updatedAt`, forever, with no signal. That is exactly how the five tables above went unnoticed.

Mutation-tested rather than assumed. Deleting `'MembershipGift'` from the committed file:

- `pnpm typecheck` → **0 errors** (blind, as predicted)
- `pnpm run db:check-generated` → **exit 1**

So the new step catches a class the existing gate structurally cannot, and it isn't redundant with typecheck.

## Verification

| | before (`origin/main`) | after |
|---|---|---|
| `pnpm typecheck` | **1** error (TS2769, the one above) | **0** errors |
| `pnpm run db:check-generated` | passes over a broken file | passes, tree clean |
| `pnpm exec eslint packages/civitai-db-schema/src/kysely/updated-at-tables.ts` | — | 0 findings |
| `pnpm --filter @civitai/db-queries test` | 15 passed / 3 skipped / 1 suite failed | 15 passed / 3 skipped / 1 suite failed — **identical** |

The one failing suite is `enum-array-parsers.explain.test.ts`, and it is **pre-existing and unrelated** — measured at the base commit and on this branch, same result. With no DB URL it throws `TypeError: Invalid URL` at `new URL(u)` because `testDbUrl()` returns `undefined` outside the `describe.skipIf` guard that protects its siblings; with a DB URL set but no reachable server, its 4 tests fail on connect. Worth a follow-up (the guard just needs to cover that suite too), but out of scope here — I didn't want to bundle an unrelated fix into the one unbreaking `main`.

`pnpm lint` reports 4,003 pre-existing problems (398 errors / 3,605 warnings) across `src/`, unchanged: this PR touches no file under the root `src/`, which is the only path `eslint src/` reads.

One residual risk worth naming: the new CI step assumes `prisma generate` is byte-deterministic on a GitHub runner. It is clean locally (Node 22 per `.nvmrc`), and `postinstall` has been running these same generators on every CI install already — but nobody has ever diffed the result on a runner, so this PR is the first time we'd find out. If it turns out to be flaky there, the step should be dropped rather than tolerated red; a gate people learn to click through is worse than no gate.


ZacxDev:
Closing as **superseded by #3587**, which landed the same root fix at 15:52Z today.

Verified against `origin/main`:
- `generator updatedAtTables` is now registered in `schema.full.prisma` (the actual root cause — #3575 shipped the generator *script* without a `generator` block, so the committed file was a one-off snapshot with no path back to the schema)
- `BaseModelLicensingFee` — **0 occurrences** (it was the `TS2769` trigger, dropped by the 07-15 licensing-root migration)
- The file carries **84 entries**, matching the 80→84 delta this PR independently measured

`main`'s typecheck is unbroken, so this PR's purpose is served. It also conflicts on all four paths, which is just the two fixes colliding.

One finding from this branch worth carrying, since it isn't in #3587: **`pnpm run db:check-generated` passed on the broken `main`.** A generator that never runs can't produce a diff — so the drift gate added by #3575 was green over a file it was supposed to be guarding. Mutation-tested here: deleting an entry from the committed file leaves `pnpm typecheck` at **0 errors** (typecheck only sees drift in one direction — a *missing* entry is a legal `keyof DB` subset), while `db:check-generated` exits 1. The two checks are complementary, not redundant, so it's worth confirming #3587 wired the check into CI as well as registering the generator.

Also unrelated but noted while here: `enum-array-parsers.explain.test.ts` fails whenever `TEST_DATABASE_URL` is unset, because it lacks the `describe.skipIf` guard its siblings have — so that suite is red in CI by default. Deliberately not bundled; worth a one-line follow-up.

PR #2874, opened 2026-07-01
feat(app-blocks): W13 P2c — unified listing detail page

ZacxDev:
## What

**Phase 2, PR 3 (P2c)** of the App Store Listings (W13) workstream: the per-listing **DETAIL** surface for the unified `/apps` store, consuming the P2a `appListings.getAppDetail` read path and resolving the P2b card-CTA stubs.

**DARK + PARALLEL-RUN:** the live `/apps/[appBlockId]` detail, `AppDetailsModal`, and default `/apps` are byte-unchanged. The new detail lives under the dark, mod-gated `store-preview` tree. Making `/apps/[slug]` canonical is a later PR (**P2d**).

## Route + dark gate

`src/pages/apps/store-preview/[slug].tsx` (a folder route coexisting with the sibling `store-preview.tsx` grid). Gating mirrors `store-preview.tsx`:
- `getServerSideProps` → `resolveAppsPageAccess` (mod-segmented `appBlocks` flag → `notFound` for anon/non-mods).
- `deIndex` on `<Meta>`; **no SSR data leak** — the listing is fetched client-side via `trpc.appListings.getAppDetail.useQuery({ slug }, { enabled: !!features.appBlocks })`.
- Missing / non-approved slug 404s server-side → `NotFound`.

## Detail body

`AppListingDetailBody` — hero cover (category-glyph placeholder + on-error fallback) · app icon · name · tagline · creator chip · kind / category / contentRating badges · Steam-style recommend breakdown (recommendPct + counts; "No reviews yet" when null, via the P2b recommend-label helper) · screenshot **gallery** (reuses `AppDetailsModal`'s `SimpleGrid` pattern; `screenshots[]{url,caption}`, skips empty/broken) · `CustomMarkdown` description · the kind-aware primary action.

## Kind-aware primary action (pure `getDetailPrimaryAction`)

| kind | condition | action |
|---|---|---|
| onsite | hasPage + canOpenPage | **Open** → `/apps/run/<slug>` |
| onsite | hasPage + !canOpenPage | **Open live** → standalone `liveUrl` (https-guarded; no dead run link) |
| onsite | !hasPage (model-slot) | **info** "Runs on model pages" → link to live `/apps/<appBlockId>` |
| offsite | external-link (https) | **Visit ↗** (rel=noopener noreferrer, new tab) |
| offsite | external-link (non-https/null) | **info** "Unavailable" (guard drops it) |
| offsite | connect | **Connect** honest stub (see below) |

[...]

PR #3585, opened 2026-08-03
fix(apps): an off-site listing with an OAuth client no longer renders a dead CTA

ZacxDev:
## The bug

Three approved, live, moderator-visible off-site listings render a **dead primary CTA** reading *"Connecting this app will be available soon."* — a disabled button with no `href`. Users have no way to open those apps from the store at all.

**Mechanism.** `getDetailPrimaryAction`'s final fallthrough (`src/components/Apps/appListingDetailView.ts`) returned the connect **stub unconditionally**. The branch just above it handles `subKind === 'external-link'` and returns a working `Visit ↗` built from `externalUrl`. Which branch you land on is decided by `resolveOffsiteSubKind` (`src/server/services/blocks/app-listing.service.ts`), which is `connectClientId ? 'connect' : 'external-link'`.

So **linking an OAuth client was the only thing that routed a listing onto the dead path.** All four off-site listings were enumerated (not sampled): the three with a linked client were dead; the one without a client rendered a working `Visit ↗`.

**The stub's stated premise was false.** Its docstring justified itself with *"a complete OAuth authorize URL is NOT derivable from the public DTO"*. True, and irrelevant: connecting already works and is in heavy daily use (~7,100 consents on Comfy Cloud, ~300 on Cosmetic Studio, ~230 on AI Radio, with activity today). These are **confidential clients** that own their own `redirect_uri` / `state` / PKCE and start the flow from their own site — the store never needed to build an authorize URL, only to get the viewer to the app. That destination is already on the public DTO as `externalUrl` (https-guarded at the projection *and* re-guarded at the render boundary), already 200-serving. The connect branch simply never read it.

## The fix

In **both** view-models, the **destination** decides the action, not the sub-kind:

```ts
// off-site, both sub-kinds
const href = safeExternalHref(kd.externalUrl);
if (href) return { label: 'Visit', mode: 'visit', href, external: true };
// …then, and only then, the sub-kind-specific no-destination fallbacks
```

- **Detail** (`appListingDetailView.ts`): a connect listing with an https `externalUrl` now returns the **byte-identical action** the external-link case has always returned for the same URL — `Visit ↗`, external anchor, `target=_blank` + `rel="noopener noreferrer"`. Copy is exactly what the existing branch used; nothing new was invented. The app then runs its own connect flow from its own site.
- **Card** (`appListingCardView.ts`) — the "also in scope" coherence item, **done, not deferred**: `getListingCta` used to give every connect card `View details`, pointing at a detail page whose affordance was the dead stub. It now mirrors the detail — direct `Visit ↗` when there is a destination, `View details` when there is not. Card and detail can no longer disagree about whether an app is reachable.
- **The stub stays reachable and fails safe**: a connect listing with a missing / blank / non-https `externalUrl` still gets the inert `Connect` + note, with no href and no dead nav. Only `https` qualifies, via the existing `safeExternalHref` — no new predicate was written.

Deliberately **not** touched: `resolveOffsiteSubKind`, any server/DTO shape, `ConnectScopesPanel`, `OffsiteReviewQueue.tsx`. No server-side connect redirect was built.

Comment/docstring claims that only the old behaviour made true were rewritten in the same commit rather than left standing — the module docstrings on both view-models, the `connect` arms in `appListingActionGlyph.ts`, and the `connect`-branch comment in `AppListingDetailBody.tsx`.

## 🔴 A deliberately-pinned test was REVERSED — this is the substance of the change

`src/components/Apps/__tests__/appListingDetailView.test.ts` previously passed a real `connectClientId: 'client-123'` and asserted:

```ts
expect(action.href).toBeUndefined();
```

That test did not merely *describe* the behaviour, it **encoded the intent**: *"even with a client_id, produce nothing."* Reversing it is the actual decision in this PR, not an incidental test edit — so it is called out here rather than left to be discovered in the diff.

The new intent, pinned in its place: **a linked OAuth client does not remove the app's address.** Since `resolveOffsiteSubKind` flips the sub-kind on `connectClientId != null` alone, the old rule made linking a client the *sole* cause of the dead CTA. `note` was only ever asserted `toBeTruthy()`, so copy was free.

The stub's own coverage was kept and strengthened rather than deleted — it is now enumerated over every way a destination can be absent (`null`, `''`, `http://…`, `javascript:…`), each with the `client_id` present, so it cannot pass by accident of a mis-resolved sub-kind.

## Tests

Node `unit` project (where the real correctness coverage for these view-models lives): **49 → 54** across the two suites; the whole `src/components/Apps/__tests__/` directory is **749 passed / 43 files**. New coverage:

- connect + https `externalUrl` → `Visit ↗` (the reversal, value-pinned).
- connect and external-link with the **same URL produce the same action** — a structural restatement that fails even if a future re-branch picks copy satisfying the literal pins.
- connect with no usable destination → the stub still fails safe, over 4 absence shapes.
- an off-site matrix (2 sub-kinds × 3 URL shapes × 2 client values) asserting no listing with an https target is left un-navigable, with an **explicit anti-vacuity count** (`navigable === 4`) so it cannot go green by producing no hrefs at all.
- card equivalents of the first three.

Browser (`component`) project — not run by CI, run here anyway because it exercises the render path that produced the user-visible symptom: `AppListingCard.browser.test.tsx` **43 passed**, `AppListingDetailBody.browser.test.tsx` **29 passed**. The new detail test asserts a real anchor to the app's address exists and that the string *"Connecting this app will be available soon"* is **gone** from that state.

### Mutation results

Every mutant was applied to the real source, run, and confirmed to die **with its own specific error** — not merely to turn something red:

| Mutant | Died on |
|---|---|
| Detail: re-gate the href on sub-kind (the exact pre-fix behaviour) | 3 unit tests — `expected { label: 'Connect', … } to deeply equal { label: 'Visit', mode: 'visit', … }`, and the matrix's `expected [ …(2) ] to deeply equal []` |
| Detail: drop the https guard (`safeExternalHref` → raw `externalUrl`) | 3 unit tests — `externalUrl=http://insecure.app: expected 'visit' to be 'connect'`, plus `expected 'http://insecure.app' to match /^https:\/\//` |
| Detail: invent new copy on the shared path (`Visit` → `Connect`) | 2 unit tests — `expected { label: 'Connect', … } to deeply equal { label: 'Visit', mode: 'visit', … }` |
| Card: re-gate the CTA on sub-kind (the pre-fix card behaviour) | 2 unit tests — `expected { label: 'View details', … } to deeply equal { label: 'Visit', … }` |
| Detail: let the connect stub swallow the external-link no-target case | 2 unit tests — `expected { label: 'Connect', … } to deeply equal { label: 'Unavailable', … }` and `expected 'connect' to be 'info'` |
| Detail re-gate, against the **browser** suite | `🔴 a connect listing WITH an https externalUrl renders a real Visit ↗ anchor` — `Timed out in waitUntil!` (the anchor keyed on the app's href never appears) |
| Card re-gate, against the **browser** suite | `🔴 off-site connect WITH an https externalUrl → Visit ↗ external anchor` — `Cannot find element with locator: getByRole('link', { name: 'Visit' })` |

The harness itself was validated against a known-bad input before any of its verdicts were read (a `Connect` → `ConnectXX` label mutation produced `expected 'ConnectXX' to be 'Connect'`, 1 failed / 18 passed).

## Base-vs-head gate attribution

`main` is red repo-wide on `Typecheck`, so a head-only reading says nothing:

- **Typecheck** — **146 errors at base, 146 at head, and the two error sets are byte-identical** (`diff` of the sorted error lines is empty). **Zero** of them are in `src/components/Apps/`. Positive control on the instrument: `tsc --listFilesOnly` confirms all four changed source files are actually in the compilation, so the clean verdict is about them and not about their absence.
- **Prettier** — measured per file, head vs base. Every changed file is `head <= base` in unformatted-line count: 4 were already unformatted at base by identical amounts, `appListingDetailView.test.ts` **improved** 36 → 27, and the files I added content to gained **zero** new formatting debt.
- **ESLint** — 0 errors across all 8 changed files; 1 pre-existing warning (`'ThemeIcon' is defined but never used` in `AppListingDetailBody.tsx`), untouched here.

## Two separate findings — flagged, NOT fixed here

1. **`cosmetic-studio` is linked to the wrong OAuth client.** Two near-identical clients share a redirect URI: the one the listing points at has **3** consents; the one the app actually uses has **301**. This is inert today *only* because the sub-kind check tests non-null — but it would authorize against an effectively abandoned client the moment anything reads `connectClientId`. It is a **data fix**, not a code fix, and is out of scope for this PR. Worth doing before any future work reads that field.
2. **Missing gate: nothing prevents approving an off-site listing whose CTA is a stub.** The stub landed 2026-07-01 (#2874); all three affected listings were approved 07-24 → 07-28, straight past it. A moderator-facing check — or simply surfacing the rendered CTA in the review queue — would have caught all three at approval time. Worth a follow-up.

## Not verified

- No production database was read or written by this change, and nothing above required it.
- The `component` (browser) suites are not run by CI and were run locally against a NixOS-substituted Chromium (the repo-pinned Playwright download is dynamically linked and cannot execute on this host). The two files were run **individually** — running both in one invocation trips an unrelated, pre-existing `vi.mock` registry error.

### DIFF
--- a/src/components/Apps/AppListingCard.browser.test.tsx
+++ b/src/components/Apps/AppListingCard.browser.test.tsx
@@ -220,11 +220,14 @@ describe('AppListingCard', () => {
     await expect.element(visit).toHaveAttribute('rel', 'noopener noreferrer');
   });
 
-  test('off-site connect → View details → unified detail (P2c)', async () => {
-    // P2c: cards route to the unified detail; the Connect action itself lives on
-    // the detail page (the connect flow needs a P2a authorize-URL DTO addition),
-    // so the card's CTA is "View details", not an inert Connect button. The
-    // former "Connect app" kind badge is gone — no longer asserted here.
+  test('off-site connect with NO external target → View details → unified detail', async () => {
+    // 🔴 `externalUrl: null` is what makes this the View-details case, NOT the
+    // sub-kind. The comment here used to say cards route connect listings to the
+    // detail because "the connect flow needs a P2a authorize-URL DTO addition" —
+    // that premise was wrong and produced a dead end (the detail's Connect
+    // affordance was an inert stub). A connect listing WITH an https
+    // `externalUrl` now gets a direct Visit ↗; see the test below. The former
+    // "Connect app" kind badge is gone — no longer asserted here.
     renderWithProviders(
       <AppListingCard
         card={base({
@@ -236,6 +239,27 @@ describe('AppListingCard', () => {
     );
     const details = page.getByRole('link', { name: 'View details' });
     await expect.element(details).toHaveAttribute('href', '/apps/store-preview/my-app');
+  });
+
+  test('🔴 off-site connect WITH an https externalUrl → Visit ↗ external anchor', async () => {
+    // The rendered counterpart of the view-model reversal: a linked OAuth client
+    // no longer strands the card on "View details". Asserted on the real anchor
+    // (href + target + rel), not just the label, because the whole defect was an
+    // affordance that looked present and went nowhere.
+    renderWithProviders(
+      <AppListingCard
+        card={base({
+          kind: 'offsite',
+          name: 'Connect App',
+          kindData: { kind: 'offsite', subKind: 'connect', externalUrl: 'https://connect.app' },
+        })}
+      />
+    );
+    await expect.element(page.getByRole('link', { name: 'View details' })).not.toBeInTheDocument();
+    const visit = page.getByRole('link', { name: 'Visit' });
+    await expect.element(visit).toHaveAttribute('href', 'https://connect.app');
+    await expect.element(visit).toHaveAttribute('target', '_blank');
+    await expect.element(visit).toHaveAttribute('rel', 'noopener noreferrer');
   });
 
   test('on-site page app WITHOUT canOpenPage → View details → unified detail (P2c)', async () => {
--- a/src/components/Apps/AppListingDetailBody.browser.test.tsx
+++ b/src/components/Apps/AppListingDetailBody.browser.test.tsx
@@ -483,8 +483,42 @@ describe('AppListingDetailBody', () => {
   // mutation and an info-branch mutation each be observed on the same run
   // instead of the first one short-circuiting the second.
 
+  test('🔴 a connect listing WITH an https externalUrl renders a real Visit ↗ anchor', async () => {
+    // The rendered reproduction of the bug this change fixes: an approved
+    // off-site listing with a linked OAuth client used to render an INERT
+    // disabled "Connect" button reading "Connecting this app will be available
+    // soon." — the viewer had no way to open the app at all. It now takes the
+    // same `visit` branch as any other off-site listing.
+    //
+    // Asserted through `renderAndSettle`, which keys on the CTA's href: this
+    // test cannot pass unless a real anchor to the app's address exists.
+    const visitBtn = await renderAndSettle(
+      <AppListingDetailBody
+        detail={base({
+          kind: 'offsite',
+          kindData: {
+            kind: 'offsite',
+            subKind: 'connect',
+            externalUrl: 'https://connect.app',
+            connectClientId: 'oauth-client-1',
+          },
+        })}
+      />,
+      'https://connect.app'
+    );
+    expect(visitBtn.textContent?.trim()).toBe('Visit');
+    expect(visitBtn.getAttribute('target')).toBe('_blank');
+    expect(visitBtn.getAttribute('rel')).toBe('noopener noreferrer');
+    expect(glyphOf(visitBtn)).toBe('external-link');
+    // The stub copy must be GONE from this state — it is the exact string a user
+    // reported as a dead end.
+    expect(document.body.textContent).not.toContain('Connecting this app will be available soon');
+  });
+
   test('the connect branch keeps its own glyph', async () => {
-    // off-site OAuth → the `connect` stub (disabled button + note).
+    // off-site OAuth with NO destination (`externalUrl: null`) → the `connect`
+    // stub (disabled button + note). 🔴 The null url is what selects this branch
+    // now, not the sub-kind — see the test above.
     await renderWithProviders(
       <AppListingDetailBody
         detail={base({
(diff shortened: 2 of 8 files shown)


ZacxDev:
## 🔴 Correction — this PR does NOT reverse a pinned invariant

The body claims the change reverses a deliberately-pinned test invariant and calls that "the substance of the change." **That is wrong**, caught by an adversarial audit. Posting as a comment rather than only editing the body, since a reviewer may already have read the stronger claim.

**What the replaced test actually pinned.** It passed `connectClientId: 'client-123'` and asserted `href` was `undefined`, which reads as *"even with a client_id, produce nothing."* But the fixture destructures `externalUrl = null` by default (`appListingDetailView.test.ts:51`), and that test passed **only** `connectClientId` — so the case it covered was connect **with no destination**.

That behaviour is **unchanged** by this PR. The stub is deliberately retained for exactly that case, so every assertion in the replaced test still holds. They are now re-pinned over **four** absence shapes instead of one — `null`, `''`, `http://insecure.app`, `javascript:alert(1)` — each carrying the client_id so the case cannot pass by accident of a mis-resolved sub-kind.

**What the new case adds** is the shape nothing covered before: connect **with** a valid https address. So this is **additive coverage, not a reversal**, and the fail-safe path ended up better covered than it was.

Fixed in `2637dcca74`. The in-source comment mattered more than the PR body here: left standing, a `🔴 INTENT REVERSED, DELIBERATELY` banner would have told the next reader that a safety property had been intentionally removed — the comment-overclaim class this codebase keeps getting bitten by, where the comment *is* the control because a reviewer who reads "already handled" stops checking.

Nothing else in the PR changes. The audit's verdict was **safe to merge**, with the fail-safe path measured at **0 leaks / 8 anchored** across 28 hostile inputs, and the browser suites (43/43 and 29/29) run with negative, positive, and by-name red controls.

### Two follow-ups noted, neither blocking

- **Connect listings now have no on-page disclosure of account access.** `getListingBadge` has had no callers since the #3391 badge removal and the public DTO carries no `connectRequestedScopes`, so a connect app's detail page is indistinguishable from a plain external link. Not a bypass — the OAuth consent screen is still the gate — but this PR creates the traffic to it and the store says nothing beforehand.
- **Card/detail coherence is asserted but pinned by nothing.** No test imports both `getListingCta` and `getDetailPrimaryAction`; it is enforced only by two parallel value pins, so a change made coherently-but-wrongly to both would pass.

PR #3588, opened 2026-08-03
fix(tailwind): make tailwind.config.js loadable by a plain CommonJS require

ZacxDev:
## The problem

`tailwind.config.js` is a CommonJS module, but it `require()`d two TypeScript sources:

```js
require('./src/utils/tailwind')             // -> src/utils/tailwind.ts
require('./src/tailwind/container-queries') // -> src/tailwind/container-queries.ts
```

Node's CommonJS resolver does not resolve `.ts`, so a plain `require()` of the config throws:

```
$ node -e "require('./tailwind.config.js')"
Error: Cannot find module './src/utils/tailwind'
    at Object.<anonymous> (tailwind.config.js:3:25)
```

## Why nobody noticed

This is **latent, not user-visible**. Tailwind v3's `loadConfig()` routes the config through its bundled jiti, and jiti *does* resolve `.ts` — so Next and the PostCSS pipeline have always loaded the real theme.

I verified that explicitly: building CSS through this repo's own `postcss.config.js` resolves `theme('colors.blue.8')` → `#1971C2` and `theme('screens.xs')` → `480px` **both before and after** this change. Nothing was broken at runtime.

The actual defect is the **fragility**: the project theme silently depends on being loaded by a TypeScript-aware loader. Any consumer that plain-`require()`s the config — a script, a lint rule, a tool that inlines it — and any tailwind version that stops routing through jiti, gets tailwind's **stock** theme instead of this project's. Stock tailwind has no `xs` screen and no numeric colour scale, while ~80 stylesheets use `theme('screens.…')` / `theme('colors.<name>.<digit>')`, so that failure mode is broad and loud when it does arrive.

## The fix

Keeping `breakpoints` single-sourced (the literal still exists in exactly one place):

- **`src/utils/breakpoints.json`** now holds the literal. JSON is the one format Node's CJS resolver, jiti, webpack and Vite all load natively, so both module systems read the same file without a TS-aware loader.
- **`src/utils/tailwind.ts`** re-exports it, so the two importers (`ContainerGrid.tsx`, `ImageCarousel.tsx`) keep their existing import path.
- **`src/tailwind/container-queries.ts` → `.js`** (CommonJS, types preserved as JSDoc). It is loaded only by the tailwind config and has no ESM consumers.

### Types are unchanged, not degraded

`resolveJsonModule` keeps the inferred type identical to the old inline literal. Verified with a type-equality probe (`Expect<Equal<typeof breakpoints, { xs: string; sm: string; md: string; lg: string; xl: string }>>`) that:

- passes against the **old** `tailwind.ts`,
- passes against the **new** one,
- and **fails** (`TS2344`) when the expected type is perturbed — so the check is real.

An explicit `IsAny` assertion also pins that it did not collapse to `any`.

## Regression test

`src/utils/tailwind.test.ts` loads the config via `createRequire` — a genuine Node CJS require, no jiti, no TypeScript loader, i.e. exactly the resolver that was failing — and asserts the project screens resolve.

**Red at base (`2bdb95adb9`), green at HEAD:**

```
# base
 ❯ unit  src/utils/tailwind.test.ts (4 tests | 4 failed)
AssertionError: expected [Function] to not throw an error but
  'Error: Cannot find module \'./src/utils/tailwind\'' was thrown
   Tests  4 failed (4)

# HEAD
 ✓ unit  src/utils/tailwind.test.ts (4 tests) 6ms
   Tests  4 passed (4)
```

Each assertion was mutation-tested so it is reachable and fails for its own reason:

| Mutation | Result |
|---|---|
| Re-inline a drifted `breakpoints` literal in the config | value + single-source guards fail (`xl: 1440px` vs `1400px`); require guard still passes |
| Point `theme.extend.containers` at a different object | **only** the single-source guard fails |
| Revert `container-queries` to `.ts` | fails with `Cannot find module './src/tailwind/container-queries'` — a *different* error, confirming the second require is genuinely covered |

## Verification

- `node -e "require('./tailwind.config.js')"` → succeeds; `theme.screens.xs === '480px'`, 2 plugins loaded.
- Real PostCSS pipeline (`postcss.config.js` → tailwindcss → this config) resolves project `theme()` lookups; negative control (`screens.doesnotexist`) correctly reports the valid keys as `'xs','sm','md','lg','xl'`.
- `AnnouncementEditModal.browser.test.tsx` — 5 passed, 999ms.
- ESLint clean on all changed files.
- `tsc --noEmit`: no new errors; the pre-existing ones in my checkout are an uninitialised `event-engine-common` submodule and a `packages/civitai-db-schema` type, none in the touched files.


ZacxDev:
## Pre-merge audit — findings folded in, plus a note on the red checks

An adversarial audit of this PR raised two things worth recording publicly.

### The red CI checks are pre-existing and unrelated to this PR

`Typecheck` and `Unit tests` are red, and they are **not** caused by this change:

- **Typecheck** fails on `packages/civitai-db-schema/src/kysely/updated-at-tables.ts(7,38): error TS2769`. This same error is red on already-merged PRs (#3574, #3576), and it independently broke an unrelated PR-preview run for #3581 today. There is already a `zach/fix-updated-at-tables-drift` branch for it.
- **Unit tests** fails in `get-models-raw.transient-503.test.ts` on pgDb mocking.

Neither file is touched here. Flagging this explicitly rather than merging past a red gate quietly.

### A real defect the audit caught in this PR — fixed in 36fa1ce

The `.ts` → `.js` conversion of `container-queries` **silently dropped that file out of the type system**. `tsconfig.json` sets `allowJs: true` but leaves `checkJs` unset, so a `.js` file under `src/` is compiled by tsc and checked by nothing — the JSDoc annotations were decoration.

Fixed with `// @ts-check`, verified by mutation in **both** directions:

| state | `error TS` count | in `container-queries.js` |
|---|---|---|
| `@ts-check`, no mutation | 13 (baseline) | clean |
| `@ts-check` + `(42).toUpperCase()` | **14** | `(20,31) TS2339` — this file, this reason |
| **no** `@ts-check` + same mutation | 13 | **0 — invisible** |

Every run carried a negative control (the known pre-existing `updated-at-tables` error had to appear), because tsc OOM-crashes on this repo at the default heap and then prints zero `error TS` lines — indistinguishable from clean. All runs used `--max_old_space_size=8192`.

Also: the `eslint-disable-next-line @typescript-eslint/no-var-requires` was **inert** (`.eslintignore` line 1 is `*.js`, so the file is never linted) — replaced with a note; and `plugins.length` `toBeGreaterThanOrEqual(2)` → `toBe(2)`, since the loose bound still passed if a plugin were dropped.

### Scope note

This fix is **latent-only**. The audit confirmed `tailwindcss@3.4.17`'s `loadConfig` calls jiti unconditionally, so nothing in the current stack plain-`require`s this config, and generated CSS is **byte-identical** before and after (verified by sha256 over a sample exercising all 5 screens, container variants, arbitrary values and custom plugins). Earlier claims that this defect caused a test hang or PostCSS theme errors were investigated and refuted — those were a wrong-cwd artifact and a cold `optimizeDeps` pass respectively. The PR body does not claim otherwise.

ZacxDev:
### Correction + rebase

Two updates to my comment above, so nobody acts on the stale version.

**1. The `Typecheck` red is RESOLVED, not pre-existing-and-permanent.** `updated-at-tables.ts(7,38) TS2769` was fixed on `main` by **#3587** ("register the `@updatedAt` generator so it stops drifting from the schema"). My earlier note said that gate was effectively permanently red — true when written, false now.

**2. Why the two gates looked like they disagreed** — worth knowing, because it will recur:

- GitHub Actions `Typecheck` runs against the **merge ref** (branch + `main`) → had #3587 → **green**
- Tekton `preview / deploy` builds the **branch tip** → lacked #3587 → **red on the identical error**

They weren't contradicting each other; they were testing different trees. This branch has now been **rebased onto current `main`** (clean, still exactly the same 5 files), so both gate the same thing.

On the rebased tree, `tsc --noEmit` leaves 12 errors, all of them `event-engine-common` submodule-not-initialised (`TS2307`) in a local worktree — CI initialises that submodule. Zero in `container-queries.js`; `updated-at-tables` is gone.

**`Unit tests` remains red for genuinely unrelated reasons**: `get-models-raw.transient-503.test.ts` (9 tests — the Meili-error widening) and `adjust-tag-level.test.ts`. This PR's own test passes in CI: `✓ unit src/utils/tailwind.test.ts (4 tests) 13ms`.

PR #3589, opened 2026-08-03
fix(schema): correct 8 referential actions that misdescribe the database

ZacxDev:
Declaration-only fix for eight Prisma relations whose declared referential action does not match how the database actually behaves. **No migration is generated and no database is touched** — see "Why there is no migration" below.

An audit found 34 relations that `schema.full.prisma` declares but the database does not enforce with a foreign key. For most of them the database is the side that should change. For the eight here the *schema* is the wrong side: adding the FK exactly as currently declared would break production.

> **Revision note.** Two mechanism claims in the original description were wrong and have been corrected below (marked ⚠️): the `*Rank_Live` views read from the `*Metric` tables, not from the parent; and only two of the six Rank tables have a live rebuild path, not all six. The conclusion is unchanged — in fact the second correction strengthens it. A third finding, added at the request of the audit, argues these relations should be **dropped from the FK remediation backlog rather than scheduled on it**.

## What changed

All in `packages/civitai-db-schema/prisma/schema.full.prisma`.

### Six `*Rank` relations: `NoAction` → `Cascade`

| Relation | Was | Now |
|---|---|---|
| `UserRank.userId` | `onDelete: NoAction` | `onDelete: Cascade` |
| `TagRank.tagId` | `onDelete: NoAction` | `onDelete: Cascade` |
| `CollectionRank.collectionId` | `onDelete: NoAction` | `onDelete: Cascade` |
| `BountyRank.bountyId` | `onDelete: NoAction` | `onDelete: Cascade` |
| `BountyEntryRank.bountyEntryId` | `onDelete: NoAction` | `onDelete: Cascade` |
| `ClubRank.clubId` | `onDelete: NoAction` | `onDelete: Cascade` |

`NoAction` maps to `ON DELETE NO ACTION`: the **parent delete errors** if a referencing row exists. Measured coverage against production:

- `BountyRank` — **100%**. Every bounty has a rank row, so every bounty would become undeletable.
- `TagRank` 70% · `ClubRank` 58% · `BountyEntryRank` 46% · `CollectionRank` 6.8% · `UserRank` 0.01%

**Why `Cascade` is right:**

1. **No code path deletes a Rank row when its parent is deleted.** This is the load-bearing argument. Greps for `prisma.<x>Rank.delete`/`deleteMany` and raw `DELETE FROM "<X>Rank"` return nothing across `src/` and `packages/`. Every parent delete goes straight at the parent with no rank cleanup: `article.service.ts:1307`, `collection.service.ts:1707`, `bounty.service.ts:551`, `bountyEntry.service.ts:639` (all bare `.delete({ where: { id } })`), and `tag.service.ts:956` (a raw `DELETE FROM "Tag"`). So under `NoAction` there is no cleanup step that could unblock the parent delete — it would just fail.

2. ⚠️ **Corrected — only two of the six have a live rebuild path, which makes `Cascade` *more* necessary, not less.** The original text claimed all six were "rebuilt wholesale". In fact:

   | Rank table | Rebuild path | Status |
   |---|---|---|
   | `BountyRank` | `recreateRankTable` via `bounty.metrics.ts:51-55` | **live**, 5-minute interval |
   | `UserRank` | `TRUNCATE` + `INSERT` in `user.service.ts:1463-1521` | **live** (its metrics processor at `user.metrics.ts:104-109` is commented out) |
   | `TagRank` | `tag.metrics.ts:84-88` | **commented out** |
   | `CollectionRank` | `collection.metrics.ts:93-97` | **commented out** |
   | `BountyEntryRank` | `bountyEntry.metrics.ts:39-43` | **commented out** ("Doesn't appear to be used anymore") |
   | `ClubRank` | — | **no metrics processor exists at all** |

   For those four, nothing deletes their rows *and* nothing ever rebuilds the table. A `NoAction` FK on them would be permanent: the blocking row would never be cleared by any process. That is a stronger argument for `Cascade` than the one it replaces.

3. ⚠️ **Corrected — the `_Live` views read from `*Metric`, not from the parent.** The original text said they "join off the parent table, so a deleted parent's rank row simply does not come back". They actually select from the metric tables: `BountyRank_Live` FROM `BountyMetric` (`programmability/views.sql:52`), and likewise at `:167`, `:325`, `:473`, `:514`.

   The conclusion survives via a second hop — all seven `*Metric` parent relations declare `onDelete: Cascade`, and unlike `TagsOnImageNew` those FKs were genuinely created (`ALTER TABLE "BountyMetric" ADD CONSTRAINT "BountyMetric_bountyId_fkey" … ON DELETE CASCADE`, and the equivalent for `ArticleMetric`/`UserMetric`/`TagMetric`/`CollectionMetric`/`BountyEntryMetric`/`ClubMetric`) with **no** `DROP CONSTRAINT` for any of them anywhere in migration history.

   **But that is migration-history evidence, not a live-catalog read**, and this PR exists precisely because declared-and-even-migrated FKs in this repo are sometimes not present in production (see `Purchase.userId` below, where a migration's trailing `SET NOT NULL` was never run). So I am not resting the conclusion on it. Argument 1 stands on its own and does not need this hop.

### `TagsOnImageNew.imageId`: implicit `Restrict` → explicit `Cascade` ⚠️ **needs sign-off**

A **required** relation with no explicit `onDelete`, so Prisma defaults it to **`Restrict`** (Prisma's default is `Restrict` for required relations, `SetNull` for optional — not `Cascade`). This is the largest table in the system at ~4.9B rows, so under `Restrict` any `Image` delete **for an image that has at least one tag row** would be rejected. That is near-universal but not literally every image — the original text overstated it as absolute.

It would also fail *silently*. `deleteImageById` (`src/server/services/image.service.ts:406-454`) issues a bare `dbWrite.image.delete({ where: { id } })` with no prior tag cleanup, and its `catch` swallows the failure into a `logToAxiom` call with no rethrow — callers get `undefined`, indistinguishable from the "already gone" path. Several callers already treat it as fire-and-forget (`article.service.ts:1181`, `:1336`; `user.service.ts:1303-1306` wraps the loop in an empty catch).

`Cascade` is the documented historical intent, from the DDL rather than by analogy:

- `20250303170613_tags_on_image_new/migration.sql:11` created the FK as **`ON DELETE CASCADE`**.
- `20250314203912_drop_tags_on_image/migration.sql:68` **dropped** it, in the same migration that dropped the legacy `TagsOnImage` table. No later migration re-adds it.
- The day before that drop, `programmability/image_delete_triggers.sql` landed `after_image_delete_trigger` (function `after_image_delete`) — an `AFTER DELETE ON "Image"` row trigger doing `DELETE FROM "TagsOnImageNew" WHERE "imageId" = OLD.id`. Trigger in, FK out. Confirmed enabled in production with that body.
- The legacy `TagsOnImage` FK was `ON DELETE CASCADE` too (`20230126222352_image_first_class/migration.sql:55`), and the sibling `TagsOnImageVote.imageId` declares `Cascade`.

**What needs sign-off:** this PR only changes the declared action — it does **not** add the FK, and the FK should never be re-added. The ask is confirmation that `Cascade` correctly describes the semantics the trigger implements.

🔴 **And a landmine this PR makes quieter, now called out in the schema comment.** `TagsOnImageNew` carries no `/// @view` annotation, so `pnpm db:migrate` sees it and will emit `ADD CONSTRAINT "TagsOnImageNew_imageId_fkey" … ON DELETE CASCADE`. **Anyone reviewing a generated migration must delete that statement.** This is more dangerous after this PR than before: on `main` the generated DDL said `RESTRICT`, which is obviously wrong on sight; `CASCADE` reads as correct.

### `ClubPost.coverImageId`: implicit `SetNull` → explicit `SetNull`

Nullable with no explicit action, so Prisma already defaults it to `SetNull`. Confirmed intended and now declared:

- The live constraint is already `ON DELETE SET NULL` — `20231128212202_add_club_post_cover_image/migration.sql:6`.
- Of the nine cover-image relations in the schema, **eight declare `SetNull` explicitly and `ClubPost` was the sole omission**, including its own siblings `Club.coverImage` and `ClubTier.coverImage` added with the same feature. None declares `Cascade`.
- The column is `Int?` and the club feature's application code was removed in `3a201c47b7`, so there is no read path that could observe a null.

Zero behaviour change; purely making the schema say what it already means.

## 🔴 These six relations should be REMOVED from the FK remediation backlog, not scheduled on it

Raised by the audit and verified here — this outranks the PR itself.

An FK on a `*Rank` table **cannot persist**. `recreateRankTable` (`src/server/metrics/base.metrics.ts:145-180`) rebuilds via:

```
CREATE TABLE "<X>Rank_New" AS SELECT * FROM "<X>Rank_Live";
ALTER TABLE "<X>Rank_New" ADD CONSTRAINT "pk_<X>Rank_New" PRIMARY KEY (...);
CREATE INDEX ...;                      -- every index re-created by hand
DROP TABLE "<X>Rank"; ALTER TABLE "<X>Rank_New" RENAME TO "<X>Rank"; ...
```

`CREATE TABLE … AS SELECT` copies no constraints — which is exactly *why* the function manually re-adds the primary key and every index. It does not re-add any foreign key. So any FK added to a rebuilt Rank table is destroyed at the next refresh: for `BountyRank`, **within 5 minutes**.

That makes `Cascade` on these relations point at a constraint that can be added and **cannot be kept** — recurring drift, which is worse than the permanent drift we have now. Scheduling "add the missing FK" for these six would produce a migration that silently un-applies itself.

**So: drop these from the FK backlog.** The real fix is structural — declare them as proper Prisma `view` blocks, or repair the `/// @view` strip (see the incidental finding below), so Prisma stops treating them as tables that ought to have foreign keys at all. This PR remains worth landing regardless: the declaration is strictly more correct than `NoAction`, and it defuses the case where someone adds the FK from the declaration and makes every bounty undeletable for the five minutes before the next rebuild wipes it.

### Why `QuestionRank` and `AnswerRank` are not included

They are the only other `NoAction` `*Rank` relations, and they are deliberately left alone: in the live catalog **both are VIEWS, not tables**. A foreign key cannot be placed on a view, so their `onDelete` is genuinely inert and there is nothing to correct. (The live catalog has 10 `*Rank` **tables** — `ArticleRank`, `BountyEntryRank`, `BountyRank`, `ClubRank`, `CollectionRank`, `ImageRank`, `ModelVersionRank`, `NewOrderRank`, `TagRank`, `UserRank` — plus these two as views.)

## Why there is no migration

None of these eight relations has a foreign key in the database today, so there is no constraint to alter and no DDL to emit. Verified without touching any database:

- `prisma migrate diff --from-schema-datamodel <main> --to-schema-datamodel <this branch> --script` (offline, schema-to-schema, no connection) emits **only** `DropForeignKey`/`AddForeignKey` pairs flipping `ON DELETE NO ACTION` → `ON DELETE CASCADE` on the six Rank relations and `RESTRICT` → `CASCADE` on `TagsOnImageNew`. That describes the FK these relations *would* get — not a pending change against the live database, which has none of those constraints.
- `ClubPost` does **not** appear in that diff at all, confirming the explicit `SetNull` is a true no-op.
- No migration file was created; `packages/civitai-db-schema/prisma/migrations/` is untouched.
- `pnpm db:generate` produces **zero** change in the checked-in generated types (`packages/civitai-db-schema/src`) — referential actions do not affect the TypeScript surface.

## The 235-column `*Rank` nullability alignment

Deferred from this PR and now open separately as **#3592**, with the exact nullable column set confirmed against the live catalog (245 columns optional, the 7 entity-id columns stay required). The two PRs conflict textually — #3592 edits the scalar lines directly beneath the relation lines this PR edits — so #3592 is based on `main` rather than stacked, and expects to rebase after this one merges.

## Two further drifted columns — deliberately not changed

Both drift because a hand-written migration was only **partially applied**, so the schema is right and the database is behind. Confirmed: neither migration was applied to production.

- **`ChallengeEvent.createdById`** — schema optional, database `NOT NULL`. `20260211192726_challenge_event_cascade_delete/migration.sql` explicitly does `ALTER COLUMN "createdById" DROP NOT NULL` and rebuilds the FK as `ON DELETE SET NULL`, with a comment explaining the intent ("preserves events when a moderator account is deleted"). Making the schema required would ratify the un-migrated state and discard a deliberate design decision.
- **`Purchase.userId`** — schema required, database nullable. `20240815210353_add_payment_providers/migration.sql` ends with `ALTER TABLE "Purchase" ALTER COLUMN "userId" SET NOT NULL;` under the comment `-- ONLY REMOVE AFTER CONFIRMING THAT THE UPDATE WORKED`. That guarded tail was never run. The fix is to finish applying the migration, not to relax the declaration.

## Incidental finding: the `/// @view` migrate workaround matches 6 of 31 models

`scripts/prisma-migrate-with-views-workaround.mjs` strips `/// @view` models before handing the schema to `prisma migrate`, using:

```
new RegExp(`\n?/// @view\nmodel ([a-zA-Z]+) {[^}]+}\n?`, 'g')
```

That requires `model X {` on the line **immediately** after the annotation. In `schema.full.prisma`, 25 of the 31 annotated models have a blank line in between, so the regex never matches them. Executing the script's own logic against the current schema strips exactly 6 — `UserStat`, `ModelTag`, `ImageResourceHelper`, `PostResourceHelper`, `GenerationCoverage`, `TagsOnImageDetails` — leaving all seven `*Rank` models in the schema that `prisma migrate` sees.

So these declarations are not inert: a `pnpm db:migrate` run today would emit FK DDL for them carrying whatever action is declared. Under `main` that would have been `ON DELETE NO ACTION` on a relation with 100% coverage. Repairing this regex is also the cheapest route to the structural fix described above. (There is also a stray duplicated `/// @view` at line ~6115 that annotates nothing.)

## Verification

- `pnpm run typecheck` — **0 errors**, and green in CI. The local run needs `git submodule update --init event-engine-common`; without the submodule the repo reports 12 pre-existing `TS2307`/`TS7006` errors on `main` unrelated to this change.
- `pnpm run test:unit:run` — 9 failures in 2 files, **pre-existing**: verified by restoring `origin/main`'s schema, regenerating, and re-running those files, which fail identically. Root cause since identified as `get-models-raw.transient-503.test.ts:72` mocking `~/server/db/pgDb` without the `pgDbReadLong` export that `src/server/db/kyselyDb.ts:5` now imports; 10 of 11 open PRs fail the same way. Out of scope here.
- `pnpm exec prisma validate` — valid.
- `pnpm db:generate` leaves `packages/civitai-db-schema/src` byte-identical, so there is no generated-artifact drift to review.
- No database was contacted: no `prisma migrate dev`/`deploy`, no `db push`, no DDL. The only Prisma commands run were `validate`, `generate`, and the offline `migrate diff --from-schema-datamodel … --to-schema-datamodel …`.


ZacxDev:
### Gate results

| Gate | Result |
|---|---|
| Typecheck (blocking) | **pass** |
| ESLint + Prettier (changed files) | **pass** — the PR touches one `.prisma` file, so CI's `'*.ts' '*.tsx'` collection yields 0 files |
| `event-engine-common` pin | **pass** |
| Unit tests (report-only) | **fail — pre-existing, not caused by this PR** |

**On the unit-test failure:** 9 tests in 2 files — `src/server/services/__tests__/get-models-raw.transient-503.test.ts` (Meilisearch transient-error → `TRPCError` conversion) and `src/__tests__/pages/api/mod/adjust-tag-level.test.ts` (fails to collect: `No "pgDbRead" export is defined on the "~/server/db/pgDb" mock`). 775 of 777 files and 11,406 of 11,416 tests pass.

This was not assumed from the topic. I restored `origin/main`'s `schema.full.prisma`, re-ran `pnpm db:generate`, and ran those two files on the pristine schema: **the same 9 tests fail identically**, so the failures predate this branch. Neither test touches Prisma referential actions, and this PR produces no change at all in the generated types the tests consume. The job is `continue-on-error: true` in `lint.yml` by design.

### Local verification detail

- `pnpm run typecheck` — **0 errors, exit 0**. Reproducing this locally requires `git submodule update --init event-engine-common`; without it the repo reports 12 pre-existing `TS2307`/`TS7006` errors on `main` unrelated to this change. CI checks out with `submodules: true` and is the authority here.
- `pnpm run test:unit:run` — 11,406 passed / 9 failed / 1 skipped, matching CI exactly.
- `pnpm exec prisma validate` — valid.
- `pnpm db:generate` — `packages/civitai-db-schema/src` byte-identical to `main` (`git status` clean for that path), confirming referential actions carry no TypeScript surface.
- No database was contacted. No `migrate dev`, no `migrate deploy`, no `db push`, no DDL. The only offline Prisma commands used were `validate`, `generate`, and `migrate diff --from-schema-datamodel … --to-schema-datamodel …`.

ZacxDev:
### Description corrected after adversarial audit

The body has been rewritten. Two mechanism claims in the original were **wrong** and are called out inline (marked ⚠️) rather than quietly edited, since both would otherwise get cited later as established:

1. **The `*Rank_Live` views read from the `*Metric` tables, not from the parent.** I originally wrote that they "join off the parent table". They select `FROM "<X>Metric"` — `programmability/views.sql:52`, `:167`, `:325`, `:473`, `:514`. The conclusion survives via a second hop (all seven `*Metric` parent relations declare `Cascade`, and those FKs *were* created `ON DELETE CASCADE` in migration history with no `DROP CONSTRAINT` anywhere), but that is migration-history evidence, not a live-catalog read — and this PR exists precisely because those two things come apart in this repo. So the argument is stated with that caveat and is **not** load-bearing. The claim that carries the conclusion is that nothing deletes rank rows at all.

2. **Only two of the six have a live rebuild path, not all six.** `BountyRank` (`bounty.metrics.ts:51-55`, 5-min) and `UserRank` (`user.service.ts:1463-1521`). `TagRank`, `CollectionRank` and `BountyEntryRank` have their processors commented out; `ClubRank` has no metrics processor at all. This **strengthens** the case — for those four nothing deletes their rows *and* nothing ever rebuilds the table, so a `NoAction` FK on them would block the parent delete permanently.

Also added, and more important than the PR itself: **an FK on a Rank table cannot persist.** `recreateRankTable` (`base.metrics.ts:145-180`) rebuilds with `CREATE TABLE … AS SELECT`, which copies no constraints — that is exactly why the function manually re-adds the PK and every index — then drops and renames. Any FK added there is gone at the next refresh; for `BountyRank`, within 5 minutes. **These six should therefore be removed from the FK remediation backlog rather than scheduled on it**, and the real fix is structural (proper Prisma `view` blocks, or repairing the `/// @view` strip regex). Called out explicitly in the body so nobody works that backlog item.

Two further additions:

- **The `db:migrate` landmine**, now in the schema comment (`f44390b1ba`). `TagsOnImageNew` carries no `/// @view` annotation, so `pnpm db:migrate` will emit `ADD CONSTRAINT "TagsOnImageNew_imageId_fkey" … ON DELETE CASCADE`. Whoever reviews a generated migration must delete that statement — and this PR makes it *harder* to catch, because on `main` the generated DDL said `RESTRICT`, obviously wrong on sight, whereas `CASCADE` reads as correct.
- **Why `QuestionRank` and `AnswerRank` are excluded**: both are VIEWS in the live catalog, and a foreign key cannot be placed on a view, so their `onDelete` is genuinely inert. Correctly left alone.

Minor corrections: the trigger is `after_image_delete_trigger` (`after_image_delete` is the function); "every `Image` delete would be rejected" softened to images having at least one tag row; and the drifted call-site line numbers are now `article.service.ts:1307`, `collection.service.ts:1707`, `bounty.service.ts:551`, `bountyEntry.service.ts:639`, plus `tag.service.ts:956` — which is a raw `DELETE FROM "Tag"`, not a Prisma `.delete()` as the original implied.

No change to the eight declarations themselves; the only code change since review is the schema comment.

PR #3590, opened 2026-08-03
feat(blocks): register the chat-completion step — the first 'textOutput' adopter

### LINKED ISSUE #3527: feat: Add chatCompletion kind to App Blocks workflow bridge

ZacxDev:
Closes #3527.

## What

Registers `chat-completion` — the second step-registry entry, and the **first `'textOutput'` adopter**. It is the capability the Sensei app has been blocked on.

```
id:                'chat-completion'
orchestratorType:  'chatCompletion'
billingMode:       'prepaidFixed'      // measured flat 1 Buzz
moderationPosture: 'textOutput'
resourcePolicy:    { kind: 'none' }
variants:          the three allowlisted models
```

Plus the type-contract anchor the registry's own clause 8a says the first `'textOutput'` entry owes, and a test file for the parts no generic clause can see.

### 1. `resourcePolicy: 'none'` is truthful, not a dodge

`staticAllowlist` is **AIR-only** — its single field is `airs`, and clause 7's enforcement is a `urn:air:` substring scan. `ChatCompletionInput.model` is a plain provider id (`"gpt-4o"`), never an AIR. Putting a model name in a field called `airs` would make the eventual real AIR-allowlist implementation wrong for whichever entries lied to it, so `'none'` is the honest declaration and clause 7's deep scan passes on the step this entry actually builds.

The model bound lives where the registry already provides one: a `z.enum` in the entry's own `.strict()` `paramSchema`, declared as the entry's `variants`. One mechanism, three results — the allowlist (enforced at parse *and* by `resolveStepVariant`), per-model pricing, and the model on the audit row as `detail.variant`.

🔴 **That bound is load-bearing, not tidiness: there is no orchestrator-side model validation.** A fabricated model name is quoted 1 Buzz, **charged** 1 Buzz, and then fails at execution with no output and no refund. Without the enum, an app typo burns a viewer's Buzz on a guaranteed failure.

### 2. `modalities` is not exposed, and `buildStep` never sets it

This is the most important line in the file. `ChatCompletionInput.modalities: ['image']` routes the request to the image pipeline and returns images on `choices[].message.images[].image_url.url` as **base64 data URIs**. Those bytes never touch image ingestion, so they never become moderated `Image` rows, never get an `nsfwLevel`, and never meet any media gate — and the text scan does not look at media at all.

Omitting the field is what makes this entry honestly text-only and keeps the registry's MEDIA-XOR-TEXT model true for it. Guarded three ways: `.strict()` rejects the param at parse; `buildStep` emits an **exact key set** pinned by a test (an allow-list assertion, not a "does not contain modalities" one — the next dangerous field is the one nobody thought to name); and `extractText` does not read `images[]`, so even if one appeared it would be dropped rather than published.

### 3. The `maxTokens` ceiling is derived from the scan cap

`maxTokens` is **required**, never `.optional()`. `.strict()` rejects params it does not know about; it does not bound a param the caller simply omits, and an omitted value falls through to the orchestrator's own default. Since the price is flat at 1 Buzz from 1 token to 200,000 (measured), an unbounded token count is unbounded compute at a fixed price.

The ceiling itself comes from `MAX_SCANNED_CONTENT_CHARS = 50_000`, above which the posture **withholds** rather than truncating — so a ceiling that can produce more than 50,000 characters designs a guaranteed "paid for nothing" into the capability:

```
4,000 tokens × ~4 chars/token ≈ 16,000 chars  →  32% of the 50,000-char cap
```

~3× headroom for content whose real chars-per-token runs above the nominal (long whitespace runs and repeated substrings merge into single BPE tokens); the cap is only reached at ~12.5 chars/token sustained.

The two constants are pinned together by a test rather than by an import, deliberately — importing `MAX_SCANNED_CONTENT_CHARS` here would drag `text-output-moderation` and through it `orchestrator.service` onto the registry's module-load path, which `workflow.schema` depends on staying light. The test imports both and goes red if either moves, and it pins the literals too, because a test comparing only the two constants to each other passes when both are changed together — which is exactly the change a human should look at.

## Measured facts this entry encodes

- **Flat 1 Buzz**, across every model and across `maxTokens` 1 → 200,000. `estimateBuzz` returns the same constant; clause 6 rejects any divergence.
- **A user message accepts a plain string `content`**, despite `UserMessage.content` being typed `Array<ChatCompletionContentPart>` in the generated client. Its own doc comment corroborates it ("can be a simple string or array of content parts"), and a live submit returned 200 and polled to `succeeded`. The part-array form is *rejected* here on purpose: `ChatCompletionContentPart` carries an `imageUrl` the orchestrator **fetches**, which is the SSRF primitive `convert-image` bounds with `civitaiHostedImageUrlSchema`. Not exposing it avoids the question.
- 🔴 **The real response's `choices[].message` carries only `content` — no `role`**, despite `AssistantMessage` declaring `role: 'assistant'` required. An extractor keyed off `role` would return nothing for every real reply: charged, `succeeded`, scanning nothing and publishing nothing, with clause 8a still green (the entry's own sample would just have been written *with* a role). `canonicalOutputFor` is the captured response verbatim, and both halves are pinned — by a runtime test and by a type assertion.
- **All three v1 models were driven to `succeeded` live.** The uncensored `dolphin-mistral-24b-venice-edition` is listed on purpose: its output is scanned exactly like every other model's, `NSFW`/`Suggestive`/`Explicit` withholding on a SFW ceiling and releasing above it, with the always-withhold labels applying at every ceiling. That is the posture's designed behaviour, not a gap being exploited.

## The type-contract anchor

Clause 8a is a **self-consistent pair** — extractor and sample, same author, same file — so it catches an extractor that surfaces nothing for its own sample but not one that agrees with a sample while both are wrong about the orchestrator. `type-contract.ts` gains five assertions deriving every hop from `ChatCompletionOutputStepLike` itself (not from the per-hop aliases, so they cannot drift from the shape the extractor casts to): the real generated step satisfies the read shape; each hop's keys exist on `ChatCompletionOutput` / `ChatCompletionChoice` / `AssistantMessage` (`keyof`, which is what makes the key *name* load-bearing — assignability alone is satisfied by anything when every property is optional); and the read does **not** require `role`.

`buildChatCompletionInput` declares `ChatCompletionInput` as its return type, so the write side is anchored too — a renamed or retyped field on the orchestrator's own input contract is a build failure here rather than a request the orchestrator quietly ignores.

## 🔴 Known gap, stated rather than papered over: the INPUT is not audited

`messages` is free text from an untrusted iframe and this entry runs **no** prompt audit over it. That is not an oversight and it is not fixable in this file. `ACCEPTABLE_POSTURES_BY_TYPE` constrains `chatCompletion` to exactly `['textOutput']` and is deliberately *a set, not a ladder*, so declaring `'promptAudit'` is rejected at load; and clause 1a's reverse direction rejects an entry that declares `auditableText` under a posture that never audits it. The registry models one posture per entry, so an entry with free text on both sides cannot express both today.

What **is** covered: everything this step publishes goes through the output scan. What is **not**: the input never meets `auditPromptServer`, so it produces no `BlockedPromptEntry` and no violation counter. Closing it is a registry-level decision about multi-surface postures, not a workaround here. It is written into the entry header so the next reader does not have to rediscover it.

## Verification

`pnpm typecheck`: **133 errors at base, 133 at head, identical 47-file set, zero in the touched files.** (`main` is red repo-wide from the `feat/db-queries-kysely` merge; what matters is that this change adds none. The count was 146 at the earlier tip this branch was first cut from, which is the baseline the mutation sweep's numbers are relative to.) Bare `tsc --noEmit` was not used — it OOMs at the default heap and exits 134, coincidentally near the error count.

Unit tests over `services/blocks/` + `schema/blocks/` + `blocks.router.workflow`: **2832 → 2873**, 116 → 117 files, all passing. The steps directory alone: **205 → 245**.

### One real regression, found by the broad run

Two population tests in `workflow.service.test.ts` called `step.extractOutput` for **every** registered entry. That was silently fine only while every entry produced media; a text entry has no media extractor by construction, so the first `'textOutput'` registration turns it into a raw `TypeError: step.extractOutput is not a function`. They are now gated on `postureProducesMedia` — the same predicate `snapshotFromWorkflow` and `projectAppWorkflow` themselves key on, so the test cannot drift from the code by asking a different question.

🔴 Gating alone would have made the text half **vacuous**, so each loop now asserts the mirror property for a text entry: it must contribute **nothing** to the media channel, which is the read-path half of the anti-smuggling rule (`StepOutputMedia.url` reaches `imageUrls` / `images[].url` without ever meeting `attachModeratedStepTextOutputs`). A third test is a positive control that the population still contains **both** shapes, so neither branch can quietly become dead.

### Mutation sweep — 16 mutants, 16 killed, each by its own error

Every guard was broken on purpose and the specific killing error recorded, not just the fact of a failure. Highlights:

| mutant | killed by | error |
|---|---|---|
| model enum → `z.string()` | *REJECTS a model outside the allowlist* | `expected true to be false` |
| `maxTokens` → `.optional()` | *REJECTS an OMITTED maxTokens* | `expected true to be false` |
| ceiling 4,000 → 20,000 | *THE DERIVATION* | `expected 80000 to be less than 50000` |
| `modalities` added to schema | *REJECTS `modalities`* | `expected true to be false` |
| `buildStep` sets `modalities:['image']` | *emits an EXACT key set* | `expected [ 'maxTokens', 'messages', …(2) ] to deeply equal [ 'maxTokens', 'messages', 'model' ]` |
| `.strict()` removed | registry **load**, clause 4 | `paramSchema must be .strict() — it accepted an unknown param` |
| `estimateBuzz` ≠ price | registry **load**, clause 6 | `estimateBuzz (2) must equal the declared price (1)` |
| `variants` decoupled from the enum | `resolveStepVariant` | `resolveVariant() returned 'deepseek/deepseek-chat', which is not one of the declared variants` |
| read shape gains a key the orchestrator lacks | `tsc` | `type-contract.ts: Type 'false' does not satisfy the constraint 'true'` |
| read shape requires `role` | `tsc` | `type-contract.ts: Type 'false' does not satisfy the constraint 'true'` |

🔴 **The one that matters most is the pair mutant:** `extractText` made to require `role` **and** `canonicalOutputFor` given a `role` to match. That is the both-wrong-blind shape clause 8a cannot catch — it passes the load-time probe cleanly — and it is killed by 9 runtime tests, including the dedicated role test. That is the evidence the new test file is doing work the existing clauses do not.

🔴 **A harness defect was found and fixed mid-sweep, and is reported rather than buried.** The first typecheck sweep returned `rc=127` with `total_errors: 0` for all four type mutants — the runner was missing `direnv exec`, so `pnpm` was not on `PATH` and "command not found" read as *zero type errors*, i.e. four false survivals presented as four clean runs. The sweep now asserts a **positive control** (`total_errors >= 146`, the known baseline) before any verdict is believed; the results above are from the re-run.

## Interaction with #3578

#3578 (request-time enforcement) adds `STEP_TYPE_POLICY`, keyed on orchestrator `$type`, with a `chatCompletion` row. Both PRs touch `steps/index.ts` and `git merge-tree` is clean, but the interaction is semantic, so it was checked rather than assumed:

- Its `assertRegistryEntryAgreesWithPolicy` requires a registered entry's posture to equal the posture its policy row derives. `chatCompletion` is `textSurface: 'generatedText'` → requires `'textOutput'`, which is what this entry declares. ✅
- Its `billingMode: null` for `chatCompletion` is `billingModeNotEstablished`, which that assert deliberately does not treat as a failure. ✅
- Its row sets `emitsMedia: true` for `chatCompletion` and notes that a purely text-posture treatment "hands an app a MEDIA channel that the text scan does not stand in front of". **That is true of the `$type` and not of this entry**: the channel is reachable only through `modalities`, which this entry does not expose, `.strict()` rejects, and `buildStep` never sets. `emitsMedia` is a documentation column — no code reads it — so nothing blocks either way, but the narrower claim is worth recording where the two meet.
- Its `textFields` for `chatCompletion` lists three surfaces this extractor does not read: `message.name`, `tool_calls[].function.arguments`, `finishReason`. None can appear (this entry exposes no `tools`, sends no `name`, and `.strict()` blocks both), and if one did it would be **dropped, not published** — a text entry has no `extractOutput`. Fail-closed in the right direction, and covered by tests.

Whichever lands second should re-run the steps suite on the merged tree.

## Co-requisite (does NOT block this PR, but does block Sensei)

The published `@civitai/app-sdk` — every version including current `latest` — has no `kind: 'step'` arm on `WorkflowBody`, so there is no typed client for the step bridge at all. Filed as **civitai-app-starters#208** with a changeset; it needs a release before an app can call this entry through the SDK.

### DIFF
--- a/src/server/services/blocks/__tests__/workflow.service.test.ts
+++ b/src/server/services/blocks/__tests__/workflow.service.test.ts
@@ -386,9 +386,7 @@ describe('snapshotFromWorkflow', () => {
     });
 
     it('omits spentAccountType when the transactions list is empty', () => {
-      const snap = snapshotFromWorkflow(
-        fakeWorkflow({ transactions: { list: [] } }) as never
-      );
+      const snap = snapshotFromWorkflow(fakeWorkflow({ transactions: { list: [] } }) as never);
       expect(snap.spentAccountType).toBeUndefined();
     });
 
@@ -1173,9 +1171,7 @@ describe('block input yields populated workflow metadata params (real graph path
     // The checkpoint anchor shows up in the resources snapshot (this is the part
     // of `workflowMetadata.resources` the graph resolves without enrichment).
     expect(resources).toEqual(
-      expect.arrayContaining([
-        expect.objectContaining({ id: 99, model: { type: 'Checkpoint' } }),
-      ])
+      expect.arrayContaining([expect.objectContaining({ id: 99, model: { type: 'Checkpoint' } })])
     );
   });
 
@@ -1279,7 +1275,11 @@ describe('block input yields populated workflow metadata params (real graph path
       if (!result.success) {
         throw new Error(`img2img:edit graph validation failed: ${JSON.stringify(result.errors)}`);
       }
-      const data = result.data as { workflow: string; ecosystem: string; images?: Array<{ url: string }> };
+      const data = result.data as {
+        workflow: string;
+        ecosystem: string;
+        images?: Array<{ url: string }>;
+      };
       expect(data.workflow).toBe('img2img:edit');
       expect(data.ecosystem).toBe(ecoKey);
       // The bounded source image rides into the graph's reference `images` node.
@@ -1667,10 +1667,7 @@ describe('edit-only checkpoint version + no sourceImage', () => {
 
   // ── NO REGRESSION ──────────────────────────────────────────────────────────
   it('still accepts an edit-only version WITH a sourceImage (img2img:edit)', () => {
-    const out = buildImageWorkflowInput(
-      body({ sourceImage }) as never,
-      resolved(QWEN_EDIT_V2511)
-    );
+    const out = buildImageWorkflowInput(body({ sourceImage }) as never, resolved(QWEN_EDIT_V2511));
     expect(out.workflow).toBe('img2img:edit');
     expect(out.ecosystem).toBe('Qwen');
     // The version the caller asked for is the one that anchors the graph.
@@ -1994,7 +1991,11 @@ describe('buildCustomComfyWorkflowInput (translator)', () => {
   const recipe = getRecipe('seamless-pano-360')!;
 
   it('applies the recipe param schema then runs its pure builder', () => {
-    const input = buildCustomComfyWorkflowInput(recipe, { prompt: 'a lake', seed: 1, engine: 'zimage-turbo' });
+    const input = buildCustomComfyWorkflowInput(recipe, {
+      prompt: 'a lake',
+      seed: 1,
+      engine: 'zimage-turbo',
+    });
     expect(input.trace).toBe('binary');
     expect(input.resources[0]).toContain('z_image_turbo');
     expect(input.workflow['1'].class_type).toBe('UNETLoader');
@@ -2028,12 +2029,16 @@ describe('createBlockCustomComfyStep (step wrapper)', () => {
   it('formats the timeout per engine: zimage 90s → 00:01:30, flux2 150s → 00:02:30', () => {
     const input = buildCustomComfyWorkflowInput(recipe, { prompt: 'a lake', seed: 1 });
     expect(
-      createBlockCustomComfyStep(input, recipe.budgetFor({ prompt: 'a lake', engine: 'zimage-turbo' }).stepTimeoutSeconds)
-        .timeout
+      createBlockCustomComfyStep(
+        input,
+        recipe.budgetFor({ prompt: 'a lake', engine: 'zimage-turbo' }).stepTimeoutSeconds
+      ).timeout
     ).toBe('00:01:30');
     expect(
-      createBlockCustomComfyStep(input, recipe.budgetFor({ prompt: 'a lake', engine: 'flux2-klein' }).stepTimeoutSeconds)
-        .timeout
+      createBlockCustomComfyStep(
+        input,
+        recipe.budgetFor({ prompt: 'a lake', engine: 'flux2-klein' }).stepTimeoutSeconds
+      ).timeout
     ).toBe('00:02:30');
   });
 });
@@ -2077,7 +2082,13 @@ describe('projectAppWorkflow: customComfy blobs', () => {
           metadata: {},
           output: {
             blobs: [
-              { id: 'p1', type: 'image', url: 'https://cdn/pano.png', available: true, nsfwLevel: 'pg' },
+              {
+                id: 'p1',
+                type: 'image',
+                url: 'https://cdn/pano.png',
+                available: true,
+                nsfwLevel: 'pg',
+              },
               { id: 'p2', type: 'image', url: 'https://blocked/', available: false },
             ],
           },
@@ -2202,10 +2213,14 @@ describe('🔴 registered step output — surfaced on the snapshot AND the proje
   // one-sided test behind.
   it('the registered population contains BOTH a media entry and a text entry', () => {
     const shapes = listRegisteredSteps().map(([, s]) => postureProducesMedia(s.moderationPosture));
-    expect(shapes.filter(Boolean).length, 'no MEDIA entry — the media branch above is vacuous')
-      .toBeGreaterThan(0);
-    expect(shapes.filter((m) => !m).length, 'no TEXT entry — the text branch above is vacuous')
-      .toBeGreaterThan(0);
+    expect(
+      shapes.filter(Boolean).length,
+      'no MEDIA entry — the media branch above is vacuous'
+    ).toBeGreaterThan(0);
+    expect(
+      shapes.filter((m) => !m).length,
+      'no TEXT entry — the text branch above is vacuous'
+    ).toBeGreaterThan(0);
   });
 
   // The concrete shape, pinned literally rather than derived from the extractor
@@ -2398,7 +2413,10 @@ describe('sourceImages[] — normalization + per-ecosystem cap', () => {
 
   // ── the emitted graph input ────────────────────────────────────────────────
   it('emits EVERY array element into the graph images[] (order preserved)', () => {
-    const out = buildImageWorkflowInput(body({ sourceImages: images(3) }) as never, resolved('Qwen'));
+    const out = buildImageWorkflowInput(
+      body({ sourceImages: images(3) }) as never,
+      resolved('Qwen')
+    );
     expect(out.workflow).toBe('img2img:edit');
     expect(out.images).toEqual([
       { url: 'https://image.civitai.com/abc/0.jpeg', width: 1024, height: 1024 },
--- a/src/server/services/blocks/__tests__/workflow.service.test.ts
+++ b/src/server/services/blocks/__tests__/workflow.service.test.ts
@@ -40,6 +40,7 @@ import {
   getStepByOrchestratorType,
   listRegisteredSteps,
   NATIVELY_EXTRACTED_STEP_TYPES,
+  postureProducesMedia,
   type AnyBlockStep,
 } from '../steps';
 import { nsfwLevelFromContentRating } from '~/shared/constants/browsingLevel.constants';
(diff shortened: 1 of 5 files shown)


ZacxDev:
### CI status — `Unit tests` is red, and it is **pre-existing on `main`**, established rather than assumed

`Typecheck` ✅ · `ESLint + Prettier` ✅ · `event-engine-common pin` ✅ · `Unit tests` ❌

The `Unit tests` failure is **not** from this PR. The discriminating control, run before reaching for any explanation:

- **All 8 open PRs** (`3591, 3590, 3589, 3588, 3586, 3584, 3583, 3580`) report `Unit tests: FAILURE`.
- 🔴 **Same failure, not merely "also failing"** — that distinction is the one that has produced confident wrong answers here before, so it was checked rather than inferred. PR **#3591**, which touches nothing in this area, fails with the **identical shape**: `Failed Suites 1` + `Failed Tests 9`, the same two files —
  - `src/__tests__/pages/api/mod/adjust-tag-level.test.ts` — `[vitest] No "pgDbRead" export is defined on the "~/server/db/pgDb" mock`, thrown out of `src/server/db/kyselyDb.ts:22` via `db-lag-helpers.ts`;
  - `src/server/services/__tests__/get-models-raw.transient-503.test.ts` — the 9 MeiliSearch `408/429/500/502/503/504` widening tests, failing on the same `pgDbReadLong` mock gap.
- Both are `feat/db-queries-kysely` fallout, the same merge that leaves `main` red repo-wide on `pnpm typecheck` (133 errors at base, unchanged at head here). **Neither file is in this PR's diff and neither references any symbol it introduces.**

**Delta from this PR: zero new failures.** For the code this PR actually touches, the local run over `services/blocks/` + `schema/blocks/` + `blocks.router.workflow` is **2873/2873 passing across 117/117 files** (base `origin/main`: 2832 across 116).

One note on how the ESLint failure above was fixed, because the local miss is instructive: the first local `prettier --check` passed an unquoted `"$FILES"` variable, which **zsh does not word-split**, so prettier received one long non-existent filename and printed *"All matched files use Prettier code style!"* with exit 0 — a green over **zero files**. Re-run with `${=FILES}` it reported two. The corrected invocation was validated against a deliberately mis-formatted probe file first, so its verdict is a fact about the files rather than about the shell.

PR #3591, opened 2026-08-03
feat(db-schema): schema<->database drift detector (declared vs enforced constraints)

ZacxDev:
## What

`packages/civitai-db-schema/src/schema-drift/` — a tool that parses `schema.full.prisma`,
reads a database's `pg_catalog`, and reports every constraint that is **declared but not
enforced** (and enforced but not declared).

```bash
pnpm --filter @civitai/db-schema drift        # reads DATABASE_URL
```

Migrations in this project are applied by hand, per environment. A constraint can be true
in the schema file and absent in the database for a long time with nothing noticing. This
makes the gap countable.

**Read-only.** Every statement is a `SELECT` against `pg_catalog`. Nothing was applied to
any database in the course of this change. The connection comes from `DATABASE_URL` — no
environment is baked in.

### Checks

| Class | Schema | Database |
| --- | --- | --- |
| FK present | owning-side `@relation(fields, references)` | `pg_constraint contype='f'`, matched on the **ordered** constrained-column tuple |
| Referential action | `onDelete` / `onUpdate`, explicit or defaulted | `confdeltype` / `confupdtype` |
| Column present | each scalar field | a `pg_attribute` row |
| Nullability | field `?` | `pg_attribute.attnotnull` |
| Uniqueness | `@unique` / `@@unique` | `pg_index.indisunique`, `indpred IS NULL` |

`@@map` / `@map` resolved throughout. `@@ignore` models and models mapped to a view or an
absent table are skipped and counted, not reported as missing every constraint.

**Present-but-wrong-action is new.** The constraint exists, so nothing looks missing, while
the database enforces a different rule from the one the schema promises. Both directions
are compared.

## Reproduced numbers

The snapshot is now **committed** (`__tests__/fixtures/catalog-production-2026-08-03.json`,
schema metadata only — table and column names, nullability, constraint column tuples; no row
data, nothing about where the database runs), so these are reproducible from the repo:

```bash
pnpm --filter @civitai/db-schema drift \
  --catalog src/schema-drift/__tests__/fixtures/catalog-production-2026-08-03.json
```

```
declared owning-side relations : 474
checked against the database   : 445
skipped (view / absent table)  : 29
MISSING foreign key            : 37
wrong referential action       : 0     <- NOT measured; see below
action not comparable          : 408
MISSING column                 : 11
nullability checked            : 2383
nullability drift              : 246
uniqueness declarations checked: 122
missing unique index           : 1
```

⚠️ **These correct the first version of this PR, which said 468 declared / 34 missing.**
Those numbers were wrong: the parser could not see six relations (blocker 1 below).

- **246 nullability**: 235 across seven `*Rank` tables (`ArticleRank`, `BountyEntryRank`,
  `BountyRank`, `ClubRank`, `CollectionRank`, `TagRank`, `UserRank`), plus
  `ChallengeEvent.createdById` (declared optional, column `NOT NULL`), `Purchase.userId`
  (declared required, column `NULLABLE`), and nine `*Metric.updatedAt`.
- **11 missing columns**: ten `UserRank.thumbs{Up,Down}Count*Rank` — and an eleventh the
  review did not have, **`ModelFlag.sfwOnly`** (`Boolean @default(false)`, no such column on
  `ModelFlag`). Reporting 10 would have been reciting rather than measuring.
- **1 uniqueness**: `ImageResource(modelVersionId, name, imageId)`.

**The `0` on referential actions is "not measured", not "clean".** This snapshot predates
that check and carries no `ON DELETE`/`ON UPDATE` data, so all 408 comparable foreign keys
report as *not comparable*, on their own line, and `--strict` now fails on them.

### Live run (performed by the reviewer, read-only, against the dev clone)

|  | dev clone | prod snapshot |
| --- | --- | --- |
| declared / checked / skipped | 468 / 439 / 29 | 468 / 439 / 29 |
| missing FKs | 38 | 34 |

(pre-blocker-1 numbers.) The 4-relation delta is **exactly**
`CollectionItem.{articleId, imageId, modelId, postId}` — the clone predates the 2026-08-03
fix. The tool measured a real difference between two databases rather than reciting one.

**Referential-action drift, first ever measurement: 45.** All 45 are `ON UPDATE` only —
declared `Cascade`, database `NoAction` — with **zero** `ON DELETE` mismatches, and all 45
are App Blocks tables. The concern that ~400 findings would appear as Prisma-version
artifacts (≤2.x defaulted required relations to `Cascade`; 3.0 changed it to `Restrict`)
**did not materialise**. Note for whoever turns on a gate: those 45 are real but not
actionable, so either separate `ON UPDATE` from `ON DELETE` by severity or let the baseline
absorb them.

### 🔴 Blocker 1 — the parser silently missed 6 owning-side relations

`RELATION` accepted `@relation("X", fields: …)` but not `@relation(name: "X", fields: …)`.
Prisma takes both. This was **not a partial result**: the six relations written the other way
appeared in no counter and produced no finding, so the tool reported clean on foreign keys it
had never looked at.

The six: `Club.{coverImage,headerImage,avatar}`, `ClubMembership.{clubTier,downgradeClubTier}`,
`ChatMessage.referenceMessage`. **Three of them — the Club image relations, all → `Image`, all
declared `SetNull` — have no foreign key in production**, which is why the totals moved from
468/34 to 474/37.

**The test that missed it is fixed too.** It asserted `relations.length > 400`; a floor cannot
detect an undercount. It now compares the parser against an *independently derived* count
(`countOwningRelationLines` — a different predicate over the same file), which self-updates as
the schema grows, plus a control asserting that independent count is itself non-trivial so two
equal zeros cannot satisfy it.

### 🔴 Blocker 2 — `--strict` exited 0 on a catalog that checked nothing

With the real schema and an empty catalog — what a typo'd `--db-schema`, a wrong
`DATABASE_URL`, or a role without catalog visibility produces — the CLI printed a fully clean
report and exited **0, even under `--strict`**. `assertCatalogSanity` returns early below 50
columns, so it could not fire.

`assessCoverage` now fails the run (**exit 2**, not 1 — nothing was measured either way)
when no relations were checked, no columns were checked, or over half the schema was skipped.
This is fatal regardless of `--strict`: a clean report is a claim, and an empty catalog cannot
support it.

`cli.test.ts` gives the CLI the control the differ already had, by running the real entry
point as a process: empty catalog → exit 2 with "not trustworthy"; **covering catalog → exit
0** (without that second half the first could be passing for any reason at all).

Related, same class: **`--strict` now fails when referential actions could not be compared**.
That counter reads 408 on the snapshot and produces no findings, so a gate would otherwise
pass having certified none of them.

### Triage items

- **New drift class: declared column with no column.** 11 in production (above). A Prisma read
  touching one errors outright, so it is reported separately rather than folded into
  nullability, where "declared required, database NULLABLE" would misdescribe it.
- **Block attributes now match with comments stripped.** `@@map`/`@@ignore`/`@@unique` were
  matched against the raw model body; a stray `// @@ignore` left by a revert would have skipped
  a whole model and every constraint on it. The curated fixture now carries a commented-out
  `@@ignore` so this cannot regress quietly.
- **Connection details can no longer reach a public CI log.** A positionally-passed URL is
  redacted before being echoed, and `pg` connection failures report the error **code** instead
  of the message (which embeds host, port and user).
- **Primary keys and programmability added to the not-checked list**, in the report and the
  README. 105 `@@id` declarations exist and none is verified; views/functions/triggers are a
  surface nothing here touches (a deployed view emitting columns its committed definition does
  not is exactly the drift this tool would *not* catch).
- **Snapshot committed** so the numbers are reproducible.
- **Lockfile line left as-is** — root `package.json` pins `@types/node` 24.13.3 and that
  version already appears 97× on `main`; the change moves a peer toward the repo pin.

#### The checked-in `pg_dump` cannot replace the snapshot — measured

Review asked me to pin against `containers/db/docker-init/02_all_dll.sql` rather than ship a
snapshot, with the caveat to verify its freshness first. **Verified, and it cannot.** It is a
dev bootstrap DDL, not a production mirror: 195 tables against production's 342, 178
production foreign keys absent, and **25 foreign keys declared that production does not
have**.

Against the 37 missing foreign keys documented above, that dump would:

| | count | |
| --- | --- | --- |
| **contradict** | 19 | the dump has the constraint; production does not |
| **not see** | 7 | table absent from the dump entirely (incl. `ChallengeEvent`) |
| agree | 11 | |

**All three `Club` image relations sit in the contradicted 19** — pinning to it would assert
that the exact foreign keys blocker 1 uncovered are present, and would report ~11 missing
instead of 37. It *does* agree on the `*Rank` nullability shape (`UserRank`: 40 columns, 1
`NOT NULL`, no `thumbs*`) and on `ModelFlag.sfwOnly` being absent, which is why a spot check
there reads as a match: the disagreement is confined to foreign keys and only shows up when
you enumerate them. The snapshot therefore stays; the measurement is in the README so nobody
re-derives this.

Related, from the same review: this tool decides "backed by a view" from the catalog
(`relkind IN ('r','p')`), **not** from the `/// @view` annotation — so the detached strip
regex in `scripts/prisma-migrate-with-views-workaround.mjs` (26 of 32 annotations have a
blank line before `model`) does not affect it. `BountyEntryRank_Live` has been added
alongside `BountyRank_Live` as the second live view-drift instance in the not-checked list.

#### One deliberate deviation

The review asked for the acceptance **totals** to be pinned in a test. They are documented and
reproducible, but the test asserts the *named findings* instead. Reason: the snapshot is frozen
and the schema is not, so the pair decays — a field added tomorrow has no column in a catalog
captured today and would be reported, correctly given those two inputs, as a new missing
column. Pinned totals would go red on ordinary schema work while saying nothing about this
tool, and a gate that reds for unrelated reasons gets deleted. `production-snapshot.test.ts`
names the findings (the three Club FKs, `ChallengeEvent.createdById`, `Purchase.userId`, the
seven Rank tables at ≥235, `ModelFlag.sfwOnly`, the `UserRank.thumbs*` columns,
`ImageResource`), asserts drift in **both** nullability directions, and pins
`declaredRelations` to the independent count — so an undercount still fails, while schema
growth does not.

## Controls

A detector that returns nothing looks exactly like one wired to nothing, so both were
observed:

- **Negative control, watched failing.** `compare.test.ts` compares an aligned
  schema/catalog pair, asserts the run is silent, removes one foreign key from the catalog,
  and asserts the same comparison reports exactly that key. With the missing-FK reporting
  disabled in `compare.ts`, the suite went **3 failed / 46 passed** — including that test
  by name — and back to **51 passed** when restored.
- **Positive control on the zero.** The aligned run also asserts non-zero *checked* counts
  (4 relations, 3 unique declarations, >15 columns), so an empty finding list cannot come
  from a comparison that visited nothing.
- **Mutation sweep: 24 mutations, 24 killed** (16 in the first commit, 8 for the review
  fixes), each with an assertion specific to the guard. The blocker-1 mutation — reverting
  the regex — now reds **10 tests** including "finds every owning-side relation in the
  schema", the assertion that replaced the floor which could not see it.
- First commit's 16, each with an assertion specific to the guard
  — the `Cascade` default, unordered tuple matching (both the index and the lookup side),
  `indpred IS NULL`, the uniform-catalog guard, `@@ignore` skipping, inverted nullability, a
  constant-`true` nullability read, suppressed missing-FK findings, the action check, the
  uniqueness check, unknown action codes, the expression-index filter, `relkind`, the
  out-of-scope disclaimer, the `::text` casts, and the unparsed-array guard.

The first version of that sweep was itself broken — it passed `--reporter=basic`, which
vitest 4 rejects, so every run tested nothing and reported nothing. The baseline said the
same as the mutants, which is what gave it away.

## Two traps that produced wrong answers first

1. **Prisma's implicit `onDelete` is not `Cascade`.** Optional → `SetNull`, required →
   `Restrict`. `TagsOnImageNew.imageId` (a bare `@relation` on a very large table) is a
   `Restrict` that would *reject* an `Image` delete, not cascade it. Pinned by a test
   against the real schema.
2. **`array_agg(a.attname)` is `name[]`, and node-postgres has no parser for it.** It
   arrives in JavaScript as the literal string `"{projectId,position}"`. Nothing throws,
   `.length` still answers, and every column-tuple comparison silently becomes a character
   count — unique indexes vanished from the catalog entirely. **This was invisible to the
   unit tests**, which hand the reader already-parsed arrays; it was found by executing the
   queries against a real Postgres. Fixed with `::text`, guarded by `assertParsedArray`
   (which fails the read rather than returning a plausible catalog), and pinned in the SQL.

Also carried forward from the earlier analysis: `NOTNULL` is a reserved word, so
`SELECT a.attnotnull notnull` parses as the postfix `IS NOT NULL` operator and returns a
constant `true`. `assertCatalogSanity` now rejects a catalog whose columns all answer the
same way.

## Not checked — and not audited

Check constraints, column defaults, column types, enum values, non-unique indexes,
**primary keys** (`@id`/`@@id` — 105 declarations, none verified), **programmability**
(views, functions, triggers — a deployed view can emit columns its committed definition does
not), columns and tables present in the database but not the schema, and whether a FK's
*referenced table* matches the schema. The text report ends with this list so a quiet section
is not read as an all-clear.

## Gates

| Gate | Result |
| --- | --- |
| `tsc --noEmit` (full repo) | **0 errors**. Verified it covers the new files: a deliberate type error in `report.ts` was reported, then reverted. It also caught a real miss in the follow-up — a `DriftKind` union edit that had not applied, which the green test run could not see because Vitest strips types |
| `eslint` (13 new files) | 0 errors, 0 warnings |
| `prettier --check` | clean |
| `pnpm --filter @civitai/db-schema test` | **81 passed** (6 files) |
| `pnpm install --frozen-lockfile` | "Lockfile is up to date" |
| `pnpm run db:generate` | no diff under `packages/civitai-db-schema/src` — `db:check-generated` stays clean |

Two honest caveats:

- **This suite does not run in CI.** The root `unit` project globs `src/**` only, so no
  workflow runs `packages/*` tests today (`civitai-redis`, `civitai-storage` and five others
  are in the same position). This follows the existing package convention rather than
  changing a shared config; wiring package suites into CI is a separate change.
- **The lockfile diff contains one unrelated line**: `civitai-db`'s vitest peer
  re-resolved from `@types/node@20.19.9` to `24.13.3` during install. `--frozen-lockfile`
  is satisfied either way.


ZacxDev:
CI: Typecheck, ESLint + Prettier and the event-engine-common pin all pass.

The `Unit tests` job fails, and it is **not** this change. It is the report-only (`continue-on-error`) root suite, and it fails identically on unrelated branches — checked `zach/fix-tailwind-config-require` (run 30854277744) and `zach/revenue-undiscriminated-zero` (run 30853118879), both Unit tests `failure` with Typecheck and ESLint green. The failure is an incomplete `vi.mock` of `~/server/db/pgDb` in `src/server/services/__tests__/get-models-raw.transient-503.test.ts` (`No "pgDbReadLong" export is defined on the mock`). This PR adds nothing under root `src/` and does not touch that module or its mocks.

The suite this PR adds is `pnpm --filter @civitai/db-schema test` (51 passing), which no workflow currently runs — the root `unit` project globs `src/**` only.

PR #3592, opened 2026-08-03
fix(schema): align the seven *Rank models' nullability with the database

ZacxDev:
Declaration-only alignment of the seven `*Rank` models with the database. **No migration, no DDL, no database contact** — these columns are already nullable in production, so there is nothing to alter.

Split out from #3589 deliberately. See "Relationship to #3589" at the bottom — **the two conflict textually and this PR is based directly on `main`, not stacked on #3589.**

> **Revision note.** Two corrections since the original description, both marked ⚠️ inline rather than silently edited. (a) The claim that every consumer reads via `$queryRaw` was **wrong** — there are six Prisma-client consumers; the conclusion holds for a different reason, stated in the blast-radius section. (b) The phantom-column question is closed: all 10 of the Rank-model phantoms are on `UserRank` (schema-wide the count is 11 — see the scope note), and my original guess (`BountyRank`'s `commentCount*` or `BountyEntryRank`'s `tipped*`) was wrong in both directions because I diffed against the committed `programmability/views.sql` and the deployed views have drifted from it — which turned up a separate hazard, now its own section. Column-level facts are cited against the checked-in `containers/db/docker-init/02_all_dll.sql` so any reviewer can reproduce them without database access — with an explicit caveat that the dump is a dev bootstrap DDL and is **not** a production mirror for foreign keys or table coverage.

## What changed

`packages/civitai-db-schema/prisma/schema.full.prisma` plus the checked-in generated types (`packages/civitai-db-schema/src/{models.ts,kysely/types.ts}`), which `pnpm db:generate` reproduces exactly.

**245 scalar columns marked optional; the 7 entity-id columns stay required.**

| Model | Columns made optional | Kept required |
|---|---|---|
| `ArticleRank` | 65 | `articleId` |
| `UserRank` | 45 | `userId` |
| `TagRank` | 30 | `tagId` |
| `CollectionRank` | 15 | `collectionId` |
| `BountyRank` | 30 | `bountyId` |
| `BountyEntryRank` | 45 | `bountyEntryId` |
| `ClubRank` | 15 | `clubId` |

These tables were hand-built outside the Prisma migration path — primary keys named `pk_ArticleRank` rather than `<Table>_pkey`, and fossil `<Table>_New_<entity>Id_not_null` constraint names showing they were built as `<Table>_New` and renamed into place. That single per-table not-null constraint on the entity id is the *only* one each table has, which is exactly why the entity id is the sole column that stays required.

Nothing else moved: no relation field is touched, every one of the 245 changed lines keeps its `@default(0)`, and no `@id` line appears in the diff.

**One caveat on "aligned with the database": this PR aligns NULLABILITY only, not physical type.** 205 of the changed columns are `bigint` in the checked-in DDL dump (`02_all_dll.sql`; see the scope caveat below — it is reliable for columns, not for foreign keys) while the schema declares them `Int` with no `@db.BigInt` override — `ArticleRank` 65, `BountyEntryRank` 45, `UserRank` 35, `BountyRank` 30, `CollectionRank` 15, `ClubRank` 15; `TagRank` is the only one whose columns are genuinely `integer`. That mismatch is pre-existing, untouched here, and out of scope — but it means "the declaration now matches the database" is true of nullability and not of type.

**The read path already assumed this.** `src/server/services/article.service.ts:353-366` coalesces NULL rank columns to `INT_MAX`, with a comment describing the pagination bug NULLs caused — `nextCursor` became `undefined` the moment a page's last row had a NULL rank, so the feed stopped after page 1. The declaration has been the outlier, not the database.

## Blast radius: 0 type errors, and what that does and does not mean

`pnpm run typecheck` → **0 errors, exit 0.**

**What closes the gap is the attribution check, not a probe.** Every one of the 245 changed schema lines was matched against the generated output: all 245 added lines in `models.ts` fall inside the seven `Rank` interfaces, distributed 65/45/30/15/30/45/15 — exactly the per-model counts of the schema edit, with no line landing in a neighbouring `*Stat` interface. Every added line carries a `?`, every one retains its `@default(0)`, and no `@relation` or `@id` line appears anywhere in the diff. That is exhaustive over the change rather than a sample.

A `tsc` probe was also used while developing (asserting `const x: number = (…).<rankColumn>` for one column per model, plus the entity-id columns in the opposite direction: **7 errors with the change, 0 without; entity ids error in neither**). It is worth stating that it is **not committed and samples 7 of 245 columns**, so it is a development aid, not an auditable artifact. The attribution check above is the load-bearing evidence.

### ⚠️ Correcting an earlier claim about consumers

An earlier version of this description said every consumer reads these tables via `$queryRaw` with hand-written row types, and that the single Prisma-relation consumer used the relation only in `orderBy`. **Both halves were wrong.** There are at least six Prisma-client consumers of the `rank` relation, three using `select` and one using `where`:

| Site | Usage |
|---|---|
| `src/server/selectors/user.selector.ts:79-86` | `select` |
| `src/server/services/user.service.ts:200-207` | `select` |
| `src/server/selectors/model.selector.ts:81` | `select` |
| `src/server/jobs/apply-discord-roles.ts:102, 108-112` | `where` + `select` |
| `src/server/controllers/user.controller.ts:1030` | `orderBy` |
| `src/server/services/bounty.service.ts:128-138` | `orderBy` |

The conclusion holds, but for a different reason than originally stated. **Every Prisma `select`/`where` site is confined to the four `leaderboard*` columns** — `leaderboardRank`, `leaderboardId`, `leaderboardTitle`, `leaderboardCosmetic` — which were **already optional on `main`** and are untouched by this PR. The two `orderBy` sites are nullability-agnostic.

So the correct reading of the green typecheck is:

> **No consumer selects a column this PR changes through the generated types** — *not* "consumers are null-safe".

Anyone using this paragraph to reason about blast radius should take the narrower claim. The raw-SQL consumers (`users.search-index.ts:248`, `push-discord-metadata.ts:60`, `article.service.ts:413`) are outside the type system entirely and are unaffected either way.

## Separate finding: 10 declared columns on the Rank models have no backing column — all on `UserRank`

Reconciling the declared count against a live catalog read turned up a second drift class, **not fixed here**.

First, a correction to an earlier "245 vs 242 is a 3-column difference": that compared two different populations. The 245 counted required scalars *excluding* the entity ids; the 242 *includes* them. Like for like:

```
declared required scalars (245 non-id + 7 entity ids) = 252
exist in production (235 nullable + 7 not-null)       = 242
                                                        ---
declared with no backing column                       =  10
```

A live catalog read resolves which 10 — **every one of them is on `UserRank`**, and the checked-in `pg_dump` at `containers/db/docker-init/02_all_dll.sql` agrees, showing `UserRank` with 40 columns and no `thumbs*` at all:

`thumbsUpCount{Day,Week,Month,Year,AllTime}Rank` and `thumbsDownCount{Day,Week,Month,Year,AllTime}Rank`

Per-model declared-vs-live, every other model matches exactly:

| Model | Declared | Live (`02_all_dll.sql`) | Phantom |
|---|---|---|---|
| `ArticleRank` | 66 | 66 | — |
| `TagRank` | 31 | 31 | — |
| `CollectionRank` | 16 | 16 | — |
| `BountyRank` | 31 | 31 | — |
| `BountyEntryRank` | 46 | 46 | — |
| `ClubRank` | 16 | 16 | — |
| **`UserRank`** | **50** | **40** | **10** |

> 🔴 **Scope of this citation.** `02_all_dll.sql` is a **dev bootstrap DDL, not a production mirror**, and it is relied on here only for **column and nullability** facts, where it has been verified to agree with production for every Rank model (`ArticleRank` 66 · `UserRank` 40 · `TagRank` 31 · `CollectionRank` 16 · `BountyRank` 31 · `BountyEntryRank` 46 · `ClubRank` 16). **Do not cite it for foreign keys or table coverage:** it has 195 tables against production's 342 and 269 FK constraints against 422, omits `ChallengeEvent` entirely, and — directly relevant to schema-drift work — *declares* `Club_coverImageId_fkey`, `Club_headerImageId_fkey` and `Club_avatarId_fkey` (lines 15539-15559), three foreign keys production does **not** have. A column-level check reads as a clean match precisely because the divergence is concentrated in constraints.

### These 10 are included in this PR's diff, deliberately

They are marked optional along with the other 235. Stating the choice rather than leaving it silent:

- **Deleting them is a different change class.** Widening 245 columns to `T | null` was measured at 0 type errors; *removing* 10 fields narrows the generated type by deleting properties, which can break consumers and needs its own blast-radius measurement. It does not belong in a mechanical widening PR.
- **Leaving them required while their 40 siblings go optional would be worse.** That reads as a deliberate claim that these 10 are guaranteed non-null — the exact opposite of true, and more misleading than the uniform treatment.
- Marking a phantom column optional changes nothing about it either way: a Prisma read of a non-existent column fails on "column does not exist" regardless of nullability, and nothing reads these models through Prisma.

> **Scope of the count.** "10" is the figure **for the seven Rank models**, which is all this PR touches. Schema-wide the count is **11**: `ModelFlag.sfwOnly` (`Boolean @default(false)`) is declared but absent from the live `ModelFlag`, which has 9 columns. That eleventh one is outside this PR's scope and is noted so the number is not read as a whole-schema total.

**Recommended follow-up: delete them. That is the only coherent option** — "add the columns to the table" is not achievable in this repo, so it should not be offered as a live choice:

- `UserRank_Live` appears in **no** file under `programmability/` — only in four 2023-2025 migrations and the DDL dump. There is no maintained view to add the columns to.
- `UserRank`'s rank-refresh block is commented out at `src/server/metrics/user.metrics.ts:104-107`, so nothing would populate them.
- Even if the columns were added to the table by hand, `recreateRankTable`'s `CREATE TABLE … AS SELECT *` would drop them on the next rebuild unless the view produced them.

Deleting the 10 declarations needs its own blast-radius measurement (removing fields narrows the generated type, unlike widening it) and so belongs in its own PR.

## What this PR actually removes: ~235 failing `SET NOT NULL` statements

Not in the original description, and it is the strongest argument for landing this.

The `/// @view` strip in `scripts/prisma-migrate-with-views-workaround.mjs` is **detached on 26 of the 32 annotated models**, including all seven `*Rank` models. Its regex requires `model X {` on the line immediately after `/// @view`, and these models have a blank line in between, so the strip never matches them. `prisma migrate` therefore genuinely sees them as tables.

The consequence, before this PR: a generated migration would have emitted roughly **235 `ALTER TABLE "<X>Rank" ALTER COLUMN "<col>" SET NOT NULL` statements** — one per column the schema declared required while production holds NULLs. Every one of them would fail on execution against real data.

After this PR it emits **nothing** for those columns, because the declaration finally matches the catalog. That is a concrete, presently-armed hazard removed, independent of the documentation value.

## 🔴 Standalone hazard: the deployed views have drifted from `programmability/views.sql`

Surfaced while chasing the phantom columns. This is **not** a Rank-column detail — it is a fourth drift surface, alongside FKs, nullability and phantom columns, that nothing currently covers.

I originally predicted the phantom columns would be `BountyRank`'s 5 `commentCount*Rank`, reasoning that `BountyRank` is rebuilt from `BountyRank_Live` every 5 minutes and the committed view produces no such column. **That prediction was wrong**, and the reason is the drift: I diffed against the committed file and assumed it described what is running. The columns exist. All of this is reproducible from the checked-in `pg_dump` at `containers/db/docker-init/02_all_dll.sql`, no database access required:

| | `commentCount*Rank` columns |
|---|---|
| Repo `BountyRank_Live` (`programmability/views.sql:52-107`) | **0** |
| Live `BountyRank` table (`02_all_dll.sql`) | **5** |

**The hazard runs the opposite way from that reading.** These are not stale leftovers that the next refresh cleans up. `pnpm db:program` reapplies `programmability/*.sql`, replacing the deployed view with the committed one — and the next 5-minute `recreateRankTable` cycle would then rebuild `BountyRank` from it and **drop 5 currently-live columns**, while `schema.prisma` still declares them. That manufactures new phantom columns of exactly the class documented above.

**`BountyEntryRank` has the same defect, twice as large, currently dormant.** Its 10 `tippedCount*Rank` / `tippedAmountCount*Rank` columns are present in the live table (`02_all_dll.sql`) and declared in the schema, but absent from the committed `BountyEntryRank_Live`. It is dormant only because that table's rank refresh is commented out (`bountyEntry.metrics.ts:39-43`); uncommenting it would fire the same 10-column loss. So:

| Table | Columns at risk | Status |
|---|---|---|
| `BountyRank` | 5 | **live** — refresh active on a 5-minute interval |
| `BountyEntryRank` | 10 | dormant — refresh commented out, arms if re-enabled |

Worth its own remediation thread: nothing diffs deployed views and functions against `programmability/`, so this is only findable by looking.

## Relationship to #3589

**These conflict textually and this PR is NOT stacked.** It branches directly off `main` (`f596de51c6`). `git merge-tree --write-tree` between the two branches exits **1** with `CONFLICT (content): Merge conflict in packages/civitai-db-schema/prisma/schema.full.prisma`.

The cause is mechanical, not semantic: #3589 edits the `@relation` line of each `*Rank` model and this PR edits the scalar lines immediately below it, so the same diff hunks overlap. Resolution is "take #3589's relation line and this PR's scalar lines" in each of the seven models — the two changes are disjoint in meaning.

Whichever lands second needs a rebase. #3589 is the smaller and already-in-review one, so merging it first and rebasing this PR is the cheaper order.

## Verification

- `pnpm run typecheck` — **0 errors, exit 0** (with `git submodule update --init event-engine-common`; without the submodule `main` itself reports 12 pre-existing `TS2307`/`TS7006` errors).
- `pnpm run test:unit:run` — **11,481 passed, 9 failed, 1 skipped** across 779 files. The 9 failures are the **known pre-existing** ones in `get-models-raw.transient-503.test.ts` and `adjust-tag-level.test.ts`, which also fail on `main` and on #3585/#3586/#3588. The job is `continue-on-error: true` in `lint.yml` by design.
- `pnpm exec prisma validate` — valid.
- `pnpm db:generate` — output committed; regenerating locally leaves the tree clean. Note this is an **unwitnessed local claim**: `db:check-generated` (`package.json:83`) is wired into no workflow and no git hook, so nothing in CI verifies the committed generated files match the schema.
- No database was contacted. No `migrate dev`, no `migrate deploy`, no `db push`, no DDL, and no file added under `prisma/migrations/`.


ZacxDev:
### Body corrected after adversarial audit

Both flagged claims were wrong and are fixed. Following the pattern from #3589, they are marked ⚠️ inline rather than silently edited, since a future reader would otherwise cite them.

**1. The `$queryRaw` consumer claim was false — in both halves.** Verified the six Prisma-client consumers first-hand:

| Site | Usage | Columns touched |
|---|---|---|
| `user.selector.ts:79-86` | `select` | `leaderboard{Rank,Id,Title,Cosmetic}` |
| `user.service.ts:200-207` | `select` | same four |
| `model.selector.ts:81` | `select` | `leaderboardRank` |
| `apply-discord-roles.ts:102, 108-112` | `where` + `select` | `leaderboardRank` |
| `user.controller.ts:1030` | `orderBy` | `leaderboardRank` |
| `bounty.service.ts:128-138` | `orderBy` | — |

The conclusion survives for a different reason: **every `select`/`where` is confined to the four `leaderboard*` columns, which were already optional on `main` and are untouched here**, and the `orderBy` sites are nullability-agnostic. The body now states the narrow reading explicitly — *no consumer selects a column this PR changes through the generated types*, **not** *consumers are null-safe*.

**2. "Add the columns to the table" is not achievable, so it is no longer offered as a choice.** Deletion is the only coherent option: `UserRank_Live` exists in no file under `programmability/` (only four 2023-2025 migrations and the DDL dump), `UserRank`'s rank-refresh block is commented out at `user.metrics.ts:104-107`, and `recreateRankTable`'s `SELECT *` would drop hand-added columns on the next rebuild anyway.

### Added

- **The hazard this PR actually removes**, which was missing and is the strongest argument for landing it: because the `/// @view` strip is detached on 26 of 32 annotated models (all seven Rank models included), `prisma migrate` sees these as tables — so pre-PR a generated migration would have emitted **~235 `ALTER COLUMN … SET NOT NULL` statements that would fail against production NULLs**. Post-PR it emits nothing for them.
- **`BountyEntryRank` added to the view-drift inventory** as the dormant case: 10 `tipped*` columns live in the table and declared in the schema, absent from the committed `BountyEntryRank_Live`, dormant only because that refresh is commented out. `BountyRank` (5 columns) remains the live one.
- **The attribution check is foregrounded over the probe.** The 245-line check is exhaustive over the change; the `tsc` probe is now labelled for what it is — uncommitted, sampling 7 of 245 columns, a development aid rather than an auditable artifact.
- **Catalog facts now cite `containers/db/docker-init/02_all_dll.sql`** rather than an unreproducible "live catalog read", so any reviewer can verify `UserRank` at 40 columns with no `thumbs*` without database access.

### Nits — all three acted on

- **`Int?@default(0)` fixed** (`a0f5248a68`). 30 lines in `TagRank`, the file's only non-uniform formatting; base has 0 such occurrences against 1187 spaced. Confirmed type-neutral — `db:generate` output is byte-identical, so the diff is whitespace only.
- **`db:check-generated` claim softened** — now says explicitly that it is wired into no workflow and no hook, so nothing in CI verifies it.
- **The `bigint`/`Int` mismatch is now stated**: 205 of the changed columns are `bigint` in the catalog while declared `Int` with no `@db.BigInt` (ArticleRank 65, BountyEntryRank 45, UserRank 35, BountyRank 30, CollectionRank 15, ClubRank 15; TagRank's are genuinely `integer`). Pre-existing and out of scope, but it means this PR aligns **nullability** only, not physical type.

Also dropped the "34 relations" figure from both PR bodies.

No change to the 245 declarations themselves; the only commit since review is the whitespace fix.

ZacxDev:
### Scoping two figures in the body

Two clarifications, neither of which changes the code or invalidates a claim.

**1. `02_all_dll.sql` is cited for columns only, and now says so.** The body previously implied it was a production mirror. It is a **dev bootstrap DDL**. Verified in-repo:

| | Dump | Production |
|---|---|---|
| Tables | **195** | 342 |
| FK constraints | **269** | 422 |
| `ChallengeEvent` | **absent** | present |

And the sharpest edge for schema-drift work specifically: the dump **declares** `Club_avatarId_fkey`, `Club_coverImageId_fkey` and `Club_headerImageId_fkey` at lines 15539-15559 — three foreign keys production does **not** have. Anyone using it to reason about FK drift would reach exactly the wrong conclusion on those three.

It remains sound for what this PR uses it for. Every Rank column count in the dump matches production — `ArticleRank` 66 · `UserRank` 40 · `TagRank` 31 · `CollectionRank` 16 · `BountyRank` 31 · `BountyEntryRank` 46 · `ClubRank` 16 — so the citations stand and nothing needed rewriting. The divergence is concentrated in constraints, which is precisely why a column-level check reads as a clean match and generalising from one is unsafe. The body now carries that caveat inline at the citation rather than leaving the next reader to inherit the assumption.

**2. The phantom count is 10 *on the Rank models*, not schema-wide.** The heading and revision note now say so. Schema-wide it is **11** — the eleventh is `ModelFlag.sfwOnly` (`Boolean @default(false)`), declared but absent from the live `ModelFlag`, which has 9 columns. Confirmed against the dump, which lists `ModelFlag` as `modelId, minor, nsfw, poi, status, triggerWords, createdAt, details, poiName` with no `sfwOnly`.

`ModelFlag` is outside this PR's scope and is untouched; it is noted only so the number is not read as a whole-schema total. It belongs with the phantom-column follow-up rather than here.

No change to the 245 declarations. Still green, still blocked only on merge order behind #3589.

PR #3613, opened 2026-08-04
feat(app-blocks): give the ClickHouse `blockRenders` table its first reader

ZacxDev:
## What

`blockRenders` has been written since **2026-06-23** and read by **nothing** — every one of its 24 references on `main` is a writer, a comment, or a test. It is the only signal covering the viewers the author dashboard structurally cannot see: engagement comes from `block_scope_invocations`, written **only** on authenticated, scope-gated API calls, so anonymous viewers and static / no-scope blocks contribute nothing.

This adds its first reader and a **Views (range)** card: impressions + unique viewers, with a signed-out breakdown.

## Measured prod first — and it changed the design

I queried prod ClickHouse before building rather than after. Two findings **invalidated** the design the follow-up note implied ("per-mount, so unique views need query-side dedup"):

| finding | measured | consequence |
|---|---|---|
| `blockInstanceId` is **not** per-mount | 124 rows → **28** distinct instances across **27** distinct apps | it is `page_apb_<ULID>`, one per *placement* (~1:1 with the app). Deduping on it reports **"1 unique viewer"** for an app with hundreds of impressions. |
| anon rows share an identity | all `isAnon` rows carry `userId = 0` | `uniqExact(userId)` alone collapses every signed-out viewer into one phantom |

Hence `uniqExactIf(userId, isAnon = 0) + uniqExactIf(ip, isAnon = 1)`.

Also worth knowing: volume is currently small — 116 rows / 26 apps / 8 users in the last 30d — so these numbers will read low for a while. (The 2 rows under `slotId='totally-fake-slot'` are someone's synthetic control on a fake `appBlockId`; an ownership-scoped reader can never surface them.)

## Design decisions

- **`views.unavailable` is a per-SECTION flag, not a third payload-level value.** `clickhouse` is literally `undefined` without `CLICKHOUSE_HOST`/`USERNAME`, and can be down while every Postgres counter in the same response is genuinely measured. Flagging the whole payload throws away good data; omitting the flag reproduces the fabricated-zero defect #3557/#3581 fixed. The card renders an **em dash, never a 0**.
- **The read never throws** — it degrades to `unavailable`, so one flaky store can't take down a panel whose other counters are fine.
- 🔴 **Values ride in `query_params`, never interpolation.** The client's `$query` tagged template formats strings **verbatim** (`formatSqlType` returns the raw value — no quoting, no escaping), so `${id}` is an injection vector. This follows the parameterised pattern already in `scanner-review.service.ts`.
- **Week buckets use `toStartOfWeek(time, 1)`** — ClickHouse defaults to Sunday, Postgres `date_trunc('week')` is Monday. Mode 1 keeps the views line in phase with the runs line on the same chart.
- Ownership is unchanged: still resolved once upstream; the reader applies no check of its own and receives already-resolved ids.

Also updates the "What engagement counts" note, which asserted *"anonymous viewers are not counted"* — true before this card, false now.

## Verification

38 unit + 12 component tests pass; typecheck **0 errors**; prettier clean (positive control: `AppAnalyticsInline.tsx`, which violates on `main`, was correctly reported — so the check wasn't silently matching zero files).

**12 mutations across two differently-constructed sweeps**, every tree restored byte-identically and checksum-verified:

- **Sweep 1** (9 on the reader + wiring): **8 KILLED**, each by its own specific assertion. **M9 survived** and is an *equivalent mutant* — proven by construction, not asserted: when `appBlockId` is set, `getOwnedAppBlockIds` returns `[appBlockId]` or `[]`, and `[]` hits the `notOwned` early return before the reader runs, so mutant and original are identical at that call site.
- **Sweep 2** (4 ownership mutations, widened to the router test): **4/4 KILLED**.
- **Sweep 3** (3 on the card): **P1 initially SURVIVED** — replacing the em dash with the raw count. My test was *named* for the em dash but never asserted one, so a literal "0" above "Not measured right now" passed. That's the "test pins the author's belief, not the contract" failure; fixed by asserting the em dash itself. **3/3 KILLED** after.

⚠️ The component tier ran on this host's **non-canonical** browser (nixpkgs Chrome for Testing via `PLAYWRIGHT_BROWSERS_PATH`). `preview / component-tests` is the authority — please read it on this PR.

## Follow-up, not in scope here

The CLI's `civitai app metrics` consumes this same payload and will under-report until it prints the new section. That's a separate PR in `civitai/cli`.

### DIFF
--- a/src/components/AppBlocks/AppAnalyticsPanel.browser.test.tsx
+++ b/src/components/AppBlocks/AppAnalyticsPanel.browser.test.tsx
@@ -165,6 +165,51 @@ describe('AppAnalyticsPanel — impressions: measured vs unmeasured', () => {
     await expect.element(page.getByText('Active installs')).toBeInTheDocument();
   });
 
+  test('every container breakpoint is a real CSS length, not a theme key name', async () => {
+    /*
+     * Regression guard for a defect that shipped and was inert-by-default.
+     *
+     * Mantine's SimpleGrid `type="container"` branch interpolates the `cols`
+     * KEY verbatim into the query — `simple-grid (min-width: ${key})` — while
+     * the `media` branch resolves `theme.breakpoints[key]`. So `{ sm: 2 }` in
+     * container mode emits `(min-width: sm)`, which is not a valid <length>;
+     * the query never matches and EVERY width silently falls back to `base`,
+     * i.e. one column everywhere. Measured: named keys rendered 1 column at
+     * 1400px where lengths render 5.
+     *
+     * It reads as a pure styling nit, so it is worth saying why it is pinned:
+     * the failure is invisible in this tier (no Mantine stylesheet is loaded,
+     * so `grid-template-columns` computes to `none` and any layout assertion
+     * would pass either way) AND invisible in review (the diff looks like a
+     * normal responsive prop). Asserting on the emitted QUERY TEXT is the one
+     * check that can see it, and it needs no stylesheet.
+     */
+    mocks.analytics.current = { ...ZEROED };
+    renderWithProviders(<AppAnalyticsPanel scopedAppBlockId="apb_1" />);
+    await expect.element(page.getByText('App loads (range)', { exact: true })).toBeInTheDocument();
+
+    const styles = Array.from(document.querySelectorAll('style'))
+      .map((s) => s.textContent ?? '')
+      .join('\n');
+    const queries = Array.from(new Set(styles.match(/@container[^{]*/g) ?? []));
+
+    // Positive control: if the panel stops emitting container queries at all
+    // (e.g. someone reverts to type="media"), this test must fail loudly
+    // rather than pass over an empty set.
+    expect(queries.length).toBeGreaterThan(1);
+
+    for (const q of queries) {
+      const value = q.match(/min-width:\s*([^)]+)\)/)?.[1]?.trim();
+      expect(value, `no min-width parsed from ${q}`).toBeDefined();
+      // `base` is Mantine's own sentinel — its value is applied unconditionally
+      // via baseStyles, so that one never-matching query is a harmless no-op.
+      if (value === 'base') continue;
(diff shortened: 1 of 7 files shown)


ZacxDev:
### `preview / unit-tests` is red here for a **pre-existing** reason — not this PR

Chased rather than waved through, because it disagrees with the green GitHub "Unit tests" job and a two-tier disagreement can hide a real defect.

**The preview tier's own log names the file:**

```
FAIL  unit  src/server/services/__tests__/get-models-raw.transient-503.test.ts
Test Files  1 failed | 783 passed (784)
Tests       9 failed | 11600 passed | 1 skipped (11610)
=== unit FAILED (393s) (report-only, not blocking the build) ===
```

That is main's known baseline-red file — the incomplete wholesale `vi.mock` of `~/server/db/pgDb` tracked in #3579. This PR touches none of it: the diff is confined to `app-views.service.ts`, `app-analytics.service.ts`, `AppAnalyticsPanel.tsx` and their tests.

It is also why the check looks inconsistent across open PRs: **#3604 is the fix** for that defect class, so PRs containing it pass and PRs based on plain `main` fail.

| PR | contains #3604's fix | `preview / unit-tests` |
|---|---|---|
| 3604 | yes (is the fix) | success |
| 3613 (this) | no | failure |
| 3591 | no | failure |
| 3596 | no | failure |

**Worth noting for its own sake:** the full unit tier is **784/784 green on my host** and 783/784 in preview — the same suite, the same commit, and the failure is structurally invisible in the local tier. Reading only the local run would have certified this as clean.

Steps to reproduce the diagnosis:

```bash
KUBECONFIG=$KC_DPPROD kubectl logs -n tekton-builds \
  -l tekton.dev/taskRun=pr-preview-<PR>-<id>-unit-tests --tail=40 --all-containers
```

(The `preview / *` statuses point at the deployment, not the log — the Tekton taskrun is where the output actually lives.)

**`preview / component-tests` passes on this PR**, which is the signal that matters here: it covers the four new impression cases *and* re-confirms #3606's locator fix on the canonical browser.

PR #3614, opened 2026-08-04
DO NOT MERGE: deliberate CI probe for the pgDb mock parity guard

ZacxDev:
## 🔴 DO NOT MERGE — deliberate CI probe

This PR intentionally breaks an invariant so we can watch a guard fail. It has no
value as a change and **must be closed, not merged**. It will be closed as soon as
the CI result has been read.

### What it does

Deletes the `pgDbReadLong` key from one suite's inline
`vi.mock('~/server/db/pgDb', ...)` factory
(`src/server/services/__tests__/contest-entry-base-model-gate.test.ts`).

### Why

#3604 added two guards against the "incomplete pgDb mock factory" defect class.
Both were only ever verified on a dev host — the runtime scan in
`src/test-utils/__tests__/pgDbMock.parity.test.ts` has never been observed going
red in the actual CI runner. A guard nobody has watched fail is a claim, not a gate.

### Expected result

`pgDbMock.parity.test.ts` → `every suite mocking pgDb lists the complete export set
in its inline factory` fails, naming the offending file and the missing export.

Reproduced locally first (red at this commit, green at `origin/main`); this PR is
to confirm the same in CI and to record which check surfaces it to a reviewer.

### DIFF
--- a/src/server/services/__tests__/contest-entry-base-model-gate.test.ts
+++ b/src/server/services/__tests__/contest-entry-base-model-gate.test.ts
@@ -59,7 +59,7 @@ vi.mock('@civitai/db', () => ({
 }));
 
 vi.mock('~/server/db/client', () => ({ dbRead: mockDbRead, dbWrite: {} }));
-vi.mock('~/server/db/pgDb', () => ({ pgDbReadLong: {},  pgDbRead: {}, pgDbWrite: {} }));
+vi.mock('~/server/db/pgDb', () => ({ pgDbRead: {}, pgDbWrite: {} }));
 vi.mock('~/server/db/db-lag-helpers', () => ({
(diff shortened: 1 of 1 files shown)


ZacxDev:
Probe complete — closing WITHOUT merging, as intended.

Result: the runtime guard fired in real CI. GitHub Actions `Unit tests` job went red with:

```
AssertionError: 1 test file(s) mock ~/server/db/pgDb with an INCOMPLETE factory:
  server/services/__tests__/contest-entry-base-model-gate.test.ts (missing: pgDbReadLong)
```

`Test Files 1 failed | 783 passed (784)` — the parity guard was the only failure. Recorded for the follow-up on gate enforcement.

PR #3623, opened 2026-08-04
fix(images): drop the CollectionItem fan-out on image delete; raise the account-deletion drain's rate

manuelurenah:
Two changes to the account-deletion image purge: a deadlock fix, and the throughput bump it unblocks.

---

# 1. Deadlock on image delete

`remove-deleted-user-images` logs one `40P01 deadlock detected` on **every hourly run** in prod since it was enabled (2026-08-03 ~23:00 UTC):

```
Invalid `prisma.collectionItem.deleteMany()` invocation:
PostgresError { code: "40P01", message: "deadlock detected",
  detail: "Process 389789 waits for ShareLock on transaction 1847658928; blocked by process 396590.
           Process 396590 waits for ShareLock on transaction 1847658932; blocked by process 384629.
           ... (6-process cycle) " }
```

Chain, each link verified against prod:

1. `deleteImages` ran `Promise.all(ids.map(id => removeEntityFromAllCollections('image', id)))` — up to **100 concurrent single-statement transactions per batch**, five batches in flight via `Limiter`.
2. Each deleted `CollectionItem` row fires `collection_nsfw_level_change` (AFTER DELETE, FOR EACH ROW).
3. That calls `create_job_queue_record(collectionId, 'Collection', 'UpdateNsfwLevel')` → `INSERT INTO "JobQueue" … ON CONFLICT DO NOTHING`.
4. The conflict target is `JobQueue_pkey UNIQUE (entityType, entityId, type)`. Two concurrent transactions inserting the **same collectionId** block on the first one's speculative insert — hence `waits for ShareLock on transaction`. Opposite orderings across two statements close the cycle.

Fuel for step 4: one deleted account's 473 remaining images spanned **13,982 `CollectionItem` rows across 4,173 distinct collections**. Shared collections are the norm, so collisions are near-certain at 100-way concurrency.

Not specific to the drain — every `deleteImages` caller with more than one image is exposed. The drain just made it hourly.

## Fix

Delete the manual clean-up. `CollectionItem.imageId` is `onDelete: Cascade` in `schema.full.prisma` **and** in the live prod constraint (`confdeltype='c'`), so `DELETE FROM "Image"` already removes those rows and fires the same triggers — sequentially, inside one statement, so nothing can deadlock against itself.

The comment that justified the manual call ("Since we're using raw SQL delete, Prisma cascades won't trigger automatically") described Prisma-emulated referential actions. This schema sets no `relationMode`, so the database enforces the FK however the delete is issued.

| site | before | after |
|---|---|---|
| `deleteImages` | 100 concurrent `deleteMany` per batch | cascade |
| `deleteImageById` | 1 `deleteMany` before `image.delete()` | cascade |
| `deletePost` | `Promise.all` over the post's images | cascade |

`deletePost` was the worst: `removeEntityFromAllCollections` uses `dbWrite`, not the enclosing `tx`, so those statements ran on separate connections while the transaction held locks on the very rows it was about to delete.

**Left alone:** `image.controller.ts` `blockImage` — it blocks the image rather than deleting it, so no cascade would fire and the removal must stay explicit.

---

# 2. Drain throughput

At 500 images / 50 users per hourly run, the backlog — **75,092 deleted accounts holding 7,274,840 images** — needs **~606 days**.

The job is not capacity-bound. It deletes 500 images in 8–17s, so it works **~17 seconds per hour** and idles the other 59:43. `remove-blocked-images` already pushes **15,000 images per hour** through the same `deleteImages` path.

Two limits gate throughput and the lower one binds:

| lever | before | per day | clears backlog in |
|---|---|---|---|
| image budget (Redis, ops-set) | 500/run × 24 | 12,000 images | 606 days |
| `USERS_PER_RUN` | 50/run × 24 | 1,200 accounts | 63 days *just to visit each account once* |

So raising the Redis budget alone stalls on the page size. Both move:

- **`USERS_PER_RUN` 50 → 250**
- **cron `15 * * * *` → `5,15,25,35,45,55 * * * *`** (six runs an hour)

→ 36,000 accounts/day, and with the Redis budget at 2,500 that's 360,000 images/day: **~20 days**, runs of roughly 90 seconds. The backlog averages 97 images per account (median **2**, p90 77, p99 1,861, max 146,419), so the budget and page size stay roughly in step.

Off the hour on purpose: `remove-blocked-images` fires at `:00` and shares this job's S3 delete and CDN purge path.

## Checks

- **Worklist cost at the new page size**, `EXPLAIN ANALYZE` on prod: **136 ms**, walking 3,892 candidate rows to fill 250 via `User_deletedAt_notnull_idx` plus per-row index scans on `Image_userId_postId_idx` / `Post_userId_idx`. Fine six times an hour.
- **Overlap is already safe**: `acquireLock` uses `SET NX`, so a run still in flight makes the next tick a no-op. No change needed.
- **Lock headroom**: `lockExpiration` is 30 min and caps how long a run may hold the lock; at ~30–60 images/sec a 2,500-image run is ~1 min.

## Ops after merge

1. The schedule comes from the app — `/api/internal/get-jobs` serves `jobs[]` including `cron` — so the new interval ships with the deploy. Worth confirming the scheduler re-reads it rather than caching the old entry.
2. Raise the budget: `sysRedis SET system:deleted-user-image-purge-limit 2500` (currently `500`). This stays the kill switch — setting it to `0` stops the next run.
3. Watch `remove-deleted-user-images` errors and `delete-image-from-s3-failed` for a day, then decide whether to go further.

---

# Test

`src/server/services/__tests__/delete-images-collection-cascade.test.ts` asserts `deleteImages` issues no `collectionItem.deleteMany` and still issues the `DELETE FROM "Image"`. Verified it discriminates: re-introducing the fan-out fails the first case.

```
7 related suites (image/post service, account-deletion, drain + restore jobs): PASS (105) FAIL (0)
```

# Follow-ups (not here)

- Per-image cost is 1 DB query (dup-url check) + 2 HTTP round trips (`resolveMediaLocation`, image-cacher invalidate) + 1 S3 `DeleteObject`, at ~25 concurrent. Batching the dup-url check and moving to S3 `DeleteObjects` (1,000 keys/request) is where the next order of magnitude lives — only worth it if ~20 days is too slow.
- `delete-image-from-s3-failed` fired 14× in 48h with an S3 deserialization error (`char '4' is not expected`); the DB row is dropped first, so those objects may still be public.

### DIFF
--- a/src/server/jobs/remove-deleted-user-images.ts
+++ b/src/server/jobs/remove-deleted-user-images.ts
@@ -13,7 +13,12 @@ import { bustCachesForPosts } from '~/server/services/post.service';
 import { PRIOR_BLOCKED_FOR_KEY, PRIOR_INGESTION_KEY } from '~/server/utils/image-removal-mode';
 import { createJob, getJobDate } from './job';
 
-export const USERS_PER_RUN = 50;
+/**
+ * Ceiling on users reached per run, and the second of the two limits on throughput — raising the
+ * image budget alone moves nothing once a page runs out of users. The backlog's median account
+ * holds 2 images, so a page is usually far cheaper than its worst case suggests.
+ */
+export const USERS_PER_RUN = 250;
 export const DELETE_BATCH_SIZE = 100;
 
 /**
@@ -388,7 +393,9 @@ async function drainPage(
 
 export const removeDeletedUserImages = createJob(
   'remove-deleted-user-images',
-  '15 * * * *',
+  // Off the hour: remove-blocked-images runs at :00 and pushes 15K images through the same
+  // deleteImages path, S3 deletes and CDN purges included.
+  '5,15,25,35,45,55 * * * *',
   async (ctx) => {
     const budget = await getImagePurgeBudget();
     if (budget <= 0) return { paused: true, deletedImages: 0, blockedImages: 0, deletedUsers: 0 };
--- a/src/server/jobs/remove-deleted-user-images.ts
+++ b/src/server/jobs/remove-deleted-user-images.ts
@@ -395,7 +395,7 @@ export const removeDeletedUserImages = createJob(
   'remove-deleted-user-images',
   // Off the hour: remove-blocked-images runs at :00 and pushes 15K images through the same
   // deleteImages path, S3 deletes and CDN purges included.
-  '5,15,25,35,45,55 * * * *',
+  '*/10 * * * *',
   async (ctx) => {
     const budget = await getImagePurgeBudget();
     if (budget <= 0) return { paused: true, deletedImages: 0, blockedImages: 0, deletedUsers: 0 };
--- a/src/server/services/__tests__/delete-images-collection-cascade.test.ts
+++ b/src/server/services/__tests__/delete-images-collection-cascade.test.ts
@@ -1,0 +1,116 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// deleteImages used to clear CollectionItem rows itself, one concurrent statement per image. Each
+// deleted row fires collection_nsfw_level_change -> create_job_queue_record, an
+// `INSERT ... ON CONFLICT DO NOTHING` on JobQueue's unique key, so two statements touching the same
+// collection in opposite order deadlocked (40P01) on prod every run of remove-deleted-user-images.
+// CollectionItem.imageId is `onDelete: Cascade`, so the DELETE below already removes those rows.
+
+const { mockFindFirst, mockQueryRaw, mockCollectionItemDeleteMany } = vi.hoisted(() => ({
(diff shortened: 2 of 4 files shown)


src/server/jobs/remove-deleted-user-images.ts
```diff
@@ -388,7 +393,9 @@ async function drainPage(
 
 export const removeDeletedUserImages = createJob(
   'remove-deleted-user-images',
-  '15 * * * *',
+  // Off the hour: remove-blocked-images runs at :00 and pushes 15K images through the same
+  // deleteImages path, S3 deletes and CDN purges included.
+  '5,15,25,35,45,55 * * * *',
```
manuelurenah:
```suggestion
  '*/10 * * * *',
```

PR #3626, opened 2026-08-04
fix(app-blocks): bound the REST half of block_scope_invocations.endpoint too

ZacxDev:
Follow-up #5 to #3561. That PR bounded the five **router** call sites that interpolated an id into `block_scope_invocations.endpoint`. It left the other writer alone.

`normalizeEndpoint` (`src/server/middleware/block-scope.middleware.ts`) turns `req.url` into the stored `endpoint` for every `withBlockScope` REST call, and templated **numeric segments and ULIDs only**. Any other segment — a slug, a UUID, a hash, percent-encoded junk — passed through verbatim. `endpoint` is the GROUP BY key of the `topEndpoints` rollup (`app-analytics.service.ts`: `groupBy(['endpoint']) … take: 5`), so an unbounded value fragments one logical route into N count-1 rows and pushes the app's real traffic out of the top 5. Same defect class, different door.

**Reachable today, not hypothetical.** Next dispatches `/api/v1/blocks/collections/<anything>` to the `[id]` handler, and the audit row is written from `res.on('finish')` **regardless of status code** — so a caller-supplied non-numeric id is recorded before the handler ever rejects it.

## The rule: allowlist-first, not shape-first

A shape heuristic cannot do this job here, and this is the load-bearing finding. The static vocabulary of these routes is **itself slug-shaped** — `generation-resources`, `tip-allowance`, `shared-storage` — so any rule that templates `my-cool-collection` by its hyphens templates those too. Over-templating destroys the aggregate exactly as thoroughly as under-templating (every route collapses into one row) and is far easier to ship unnoticed, because nothing looks broken; the panel just stops saying anything.

So:

```
seg ∈ KNOWN_STATIC_ENDPOINT_SEGMENTS  → verbatim   (this arm runs FIRST)
/^\d+$/                               → :id       (unchanged)
ULID, bare or prefixed                → :ulid     (unchanged)
UUID, any version/case, bare or `pfx_`→ :uuid     (new)
anything else                         → :seg      (new)
```

`KNOWN_STATIC_ENDPOINT_SEGMENTS` is the 14 static segments of the 12 `export default withBlockScope(...)` route files: `api, v1, blocks, collections, follow, generation-resources, images, increment, me, models, shared-storage, tip, tip-allowance, top`. The set is **closed by construction** — Next only dispatches a request to a wrapped handler when the URL matches a route file verbatim outside its `[id]` positions, so a segment outside the set arrived in a dynamic position and is caller-supplied.

Not listed on purpose: `submissions`, `submit-version`, `withdraw`, `dev-token`, `block-tokens`. Those routes live under `/api/v1/blocks` but authenticate with an API key, not a block JWT — they never reach this middleware, so listing them would be an unfalsifiable claim about a path this function cannot see.

The allowlist is kept honest by a **drift test that derives the set by walking `src/pages/api`** for `export default withBlockScope(`, rather than trusting the comment (same posture as `analytics-bucket-labels.drift.test.ts`). A new wrapped route fails CI instead of silently recording `:seg`.

## What is deliberately NOT covered

Stated in the docstring, not just here:

- **A value that collides with the static vocabulary.** `/api/v1/blocks/collections/models` records as `…/collections/models`, not `…/collections/:seg`. Bounded either way (the vocabulary is finite), and Next routes it to the `[id]` handler which rejects it — a bounded 4xx row, not a cardinality leak.
- **A new wrapped route whose segments were not added.** It degrades to `:seg`: still bounded, just less readable. Safe direction, and the drift test is red until it is fixed.
- **The query string**, which is dropped entirely rather than templated — no `?cursor=…` value reaches the column at all.
- **The arm ORDER**, which is currently moot: no static segment matches an id shape. That is pinned by a test rather than assumed, so a future `/v2` or `/2024` segment turns it red.

## No backfill

Historical rows keep their unbounded values. The `topEndpoints` card will show a **mix** of bounded and legacy values until the old rows age out of the query range — same call as #3561 made.

The now-stale claim in `analytics-bucket-labels.ts` ("templates only numeric and ULID-shaped segments, so a slug/UUID segment survives") is corrected in the same commit; it would otherwise be a comment asserting the opposite of the code.

## Verification

- `tsc --noEmit` (with `--max_old_space_size=8192`): **0 errors**. Harness validated first — a deliberate `const x: number = normalizeEndpoint(...)` produced exactly **1** `error TS`, and the tree was restored byte-identically (sha256 checked).
- Full `unit` project: **796 files, 11774 passed, 1 skipped, 0 failed** (counted, not read off an exit code).
- New file: **27 tests**, covering both directions.
- Prettier clean on all three touched files.

### Mutation sweep — 9 mutants, 8 killed, 1 documented survivor

Each mutant applied by exact single-occurrence replacement, tests re-run, tree restored and sha256-verified after every one. Every kill is listed with the test that went red.

| # | Mutant (deletes nothing) | Verdict | Killed by |
|---|---|---|---|
| M1 | invert the allowlist test (`has` → `!has`) | KILLED | 12 tests, both directions |
| M2 | move the allowlist arm from **first to last** | **SURVIVED** | — behaviour-equivalent today; see below |
| M3 | `:seg` fallback → old verbatim pass-through | KILLED | slug / hash / one-bucket |
| M4 | off-by-one the UUID pattern (`{12}` → `{11}`) | KILLED | UUID / one-bucket |
| M5 | drop the prefixed form from the UUID pattern | KILLED | UUID |
| M6 | remove `shared-storage` from the allowlist | KILLED | 2 static-route tests + slug-shaped-static + drift equality |
| M7 | add `my-cool-collection` **to** the allowlist | KILLED | slug + one-bucket + drift equality |
| M8 | strip on `#` instead of `?` | KILLED | query-string |
| M9 | numeric arm returns the raw segment | KILLED | 3 static-route tests + numeric + query-string |

**M2 survived and that is the honest answer, not a gap I papered over.** Moving the allowlist arm last changes nothing today because no static segment matches an id shape — an equivalent mutant. Rather than assert that in a comment, the suite pins it: `no static segment collides with an id shape (so the arm order is currently moot)` fails the moment someone adds a `/v2` or `/2024` segment, which is exactly when M2 would stop being equivalent.

## Not verified

No live query against production `block_scope_invocations` — the fragmentation claim is carried over from #3561's measurement plus the code path, not re-measured here.

### DIFF
--- a/src/components/AppBlocks/analytics-bucket-labels.ts
+++ b/src/components/AppBlocks/analytics-bucket-labels.ts
@@ -105,11 +105,15 @@ export function endpointBucketLabel(endpoint: string): string {
   }
 
   // Everything else is a REST path from `normalizeEndpoint(req.url)`. Passed through
-  // verbatim, and NOT claimed to be safe or pretty: that function templates only
-  // numeric and ULID-shaped segments, so a slug/UUID segment survives, the row is
-  // written from `res.on('finish')` regardless of status code, and the value is
-  // therefore partly caller-influenced. React escapes it (no injection) and the cell
-  // clamps to one line. Inventing a name for it would be worse than showing it.
+  // verbatim, and NOT claimed to be pretty. Since `KNOWN_STATIC_ENDPOINT_SEGMENTS` landed
+  // that function is allowlist-first — only the static segments of the wrapped routes
+  // survive, everything else collapses to `:id`/`:ulid`/`:uuid`/`:seg` — so a NEW row's
+  // path is bounded and no longer caller-influenced. Two caveats before you treat it as a
+  // closed set: rows written before that change were not backfilled and still hold
+  // slugs/UUIDs verbatim, and the row is
+  // written from `res.on('finish')` regardless of status code, so 4xx paths appear too.
+  // React escapes it (no injection) and the cell clamps to one line. Inventing a name for
+  // it would be worse than showing it.
   return endpoint;
 }
 
--- a/src/server/middleware/__tests__/block-scope.normalize-endpoint.test.ts
+++ b/src/server/middleware/__tests__/block-scope.normalize-endpoint.test.ts
@@ -1,0 +1,244 @@
+import { readdirSync, readFileSync } from 'fs';
+import path from 'path';
+import { describe, expect, it } from 'vitest';
+// Imported in setup-order so the test RSA env keys are in place before
+// block-token.service evaluates (same posture as block-scope.middleware.test.ts).
+import '~/__tests__/setup';
+import { KNOWN_STATIC_ENDPOINT_SEGMENTS, normalizeEndpoint } from '../block-scope.middleware';
+
+/**
+ * `normalizeEndpoint` writes `block_scope_invocations.endpoint`, which is the
+ * GROUP BY key of the `topEndpoints` rollup in `app-analytics.service.ts`. Both
+ * directions are failure modes and both are tested here:
+ *
+ *   - UNDER-templating fragments one logical route into N count-1 rows and the
+ *     "top 5" becomes noise (the #3561 defect, for the router call sites).
+ *   - OVER-templating collapses every route to the same handful of rows, which
+ *     destroys the aggregate just as thoroughly and is far easier to ship
+ *     unnoticed — nothing looks wrong, the panel just stops saying anything.
+ *
+ * The over-templating half is the reason the rule is allowlist-first: the static
+ * vocabulary of these routes is slug-shaped (`generation-resources`,
+ * `tip-allowance`, `shared-storage`), so a "slugs are high-cardinality" shape
+ * heuristic would eat them.
+ */
+describe('normalizeEndpoint — templates caller-supplied segments', () => {
+  it('templates a numeric id', () => {
+    expect(normalizeEndpoint('/api/v1/blocks/collections/12345')).toBe(
+      '/api/v1/blocks/collections/:id'
+    );
+  });
+
+  it('templates a ULID, bare and prefixed', () => {
+    expect(normalizeEndpoint('/api/v1/blocks/collections/01ARZ3NDEKTSV4RRFFQ69G5FAV')).toBe(
+      '/api/v1/blocks/collections/:ulid'
+    );
+    expect(normalizeEndpoint('/api/v1/blocks/collections/apb_01ARZ3NDEKTSV4RRFFQ69G5FAV')).toBe(
+      '/api/v1/blocks/collections/:ulid'
+    );
+  });
+
+  it('templates a UUID in either case, bare and prefixed', () => {
+    expect(
+      normalizeEndpoint('/api/v1/blocks/collections/3f2504e0-4f89-11d3-9a0c-0305e82c3301')
+    ).toBe('/api/v1/blocks/collections/:uuid');
+    expect(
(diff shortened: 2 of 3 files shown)


ZacxDev:
### Adversarial audit run — verdict **safe to merge**, two 🟡 fixed in `68bf173`

The closed-set claim, the regex arms and the `:seg` catch-all all held up under attack. Recording what was checked and what changed, so a reviewer doesn't redo it.

**Confirmed by the audit:**

- **No second dispatch path.** `next.config.mjs` has no `rewrites` key at all; no `middleware.ts`; no catch-all under the wrapped subtrees; no `export { default } from` re-exports; `i18n` is single-locale and Next never locale-prefixes `/api`; no proxy rewrite onto `/api/v1/blocks` in `ops/` or `containers/`. The 12 `withBlockScope(` sites match the enumeration exactly.
- **`:seg` destroys nothing meaningful.** The 12 wrapped routes have only three dynamic positions, all `[id]`, and all parse strictly (`/^[0-9]+$/` ×2, `z.coerce.number().int().gt(0)`). So in practice `:seg` only ever represents an id the handler itself rejects — no scope name, action verb or meaningful slug is flattened.
- 3468 tests across the affected areas pass; `tsc` clean with an injected-error positive control; prettier and eslint clean.

**Fixed in `68bf173` — two blind spots in the drift guard itself.** No behaviour change, but this guard is the only thing keeping `KNOWN_STATIC_ENDPOINT_SEGMENTS` honest:

1. **It scanned only `.ts`.** Next's default `pageExtensions` is `['tsx','ts','jsx','js']`, and this repo **already ships `.tsx` API routes** (`src/pages/api/v1/vault/*.tsx`) — live under-coverage, not hypothetical. A wrapped `.tsx` route added later would have recorded every static segment as `:seg` with nothing going red.
2. **It matched only the direct export.** `const handler = withBlockScope(...); export default handler;` was invisible.

Validated against each known-bad state rather than assumed — a drift guard that cannot go red is worse than none:

| probe | before | after |
|---|---|---|
| wrapped route as `.tsx` | GREEN (blind) | **RED** |
| indirect default export | GREEN (blind) | **RED** |
| direct `.ts` export (regression check) | RED | **RED** |

Fake routes removed; `git status` clean; suite back to 27/27.

Also corrected the middleware comment, which pointed at `block-scope.normalize-endpoint.drift.test.ts` — a filename that doesn't exist, so anyone grepping for the guard found nothing.

**Left as-is, deliberately:**

- **Arm-order pin re-spells the shape regexes.** It fires correctly when a colliding segment is added to the allowlist (verified), but not when a colliding *shape arm* is added — e.g. inserting `/^v\d+$/ → ':ver'` (collides with the static `v1`) leaves it green. The code stays protected by the 23 behavioural cases, so this is a gap in the early-warning system, not a defect. The clean fix is extracting `templateSegmentByShape()` for both the implementation and the pin to call; that's a refactor I'd rather not fold into this PR.
- Two surviving mutants are genuinely equivalent or unreachable: the Crockford I/L/O/U exclusion isn't pinned (`{26}` character-class widening survives), and `/^\d+$/` → `/^\d*$/` survives only because the empty-segment guard always wins first.
- `z.coerce.number()` accepts `0x2a` and `1e3`, which record as `:seg` rather than `:id`. Bounded either way, and strictly better than the pre-PR verbatim behaviour.

PR #3627, opened 2026-08-04
docs(app-blocks): treat App loads as untrusted input, and log unknown-app beacons

ZacxDev:
Follow-up to #3613, which gave the ClickHouse `blockRenders` table its first reader. Reading those rows in an author-facing surface turned a dormant writer-side weakness into a consequential one.

> **Note on scope.** Specifics of the write path have been reported to **security@civitai.com** per `SECURITY.md` and are deliberately not restated here or in the code. This PR changes **nothing** about what the ingest path accepts — it is documentation plus a log line. If you are reviewing with an eye to changing ingest or the IP derivation, please get the report from security first; the mechanism matters and the comment is not sufficient to design against.

## The claim being recorded

The rows come from an **unauthenticated** write path, and the app they're attributed to is **chosen by the caller**. So both counters are influenceable from outside.

Worth stating explicitly because the intuition runs the other way: **`uniqueViewers` is not the safer number.** For signed-out viewers it is derived from request metadata the origin does not independently verify, so it is at least as influenceable as the raw count. A distinct-count is not automatically a harder-to-game count, and treating it as one is the mistake this comment exists to prevent.

## Why this changes nothing about what is accepted

The beacon is fire-and-forget and `blockRenders` has **no status column**, so a gate that wrongly rejects beacons makes App loads **under-report silently and permanently** — strictly worse than over-reporting, because a dropped beacon leaves no trace anywhere.

The durable fix is **provenance on the row so the reader can filter**, which needs a coordinated change to the out-of-repo tracker service and the ClickHouse schema (there is no DDL for this table in this repo). Detection before enforcement.

## What ships

- **Trust caveat** on `app-views.service.ts`, stating the operational rule plainly: do not build payouts, revenue share, ranking, discovery, or author-visible leaderboards on these figures. It also records the real cost of waiting — with no provenance column, rows written now can't later be sorted into trusted/untrusted, so if inflation happens first the history is *unrecoverable* rather than repairable.
- **Unknown-app beacons are logged, never dropped.** Free — the lookup already ran for the prom label. The comment gives the two independent reasons it must stay a log: `other` is also the honest value for a new app inside the cache's ≤5-min TTL, **and** for *every* app if `loadKnownAppBlockIds` fails, since it fails to an empty set and caches that (a brief DB blip would make every id look unknown). The raw id is deliberately not logged — caller-supplied, up to 256 chars, and the signal is the **rate**, not the value.

Present-day exposure is low and I'm not inflating it: ~3 rows/day platform-wide, both feature flags mod-only, nothing paid or ranked on it.

## Verification

**4 mutants, 4/4 KILLED**, tree restored byte-identically:

| mutant | result |
|---|---|
| G1 detection becomes a **gate** (drops the beacon) | KILLED |
| G2 logs on **every** request (signal buried) | KILLED |
| G3 **never** logs (detection silently dead) | KILLED |
| G4 leaks the raw caller-supplied id | KILLED |

G1 and G2 are why the tests pin **both** directions: asserting only that the log fires would let it degrade into a gate without failing, and asserting only that the beacon survives would let the log quietly never fire.

43 tests pass across the touched files; typecheck **0 errors**; prettier clean (positive control: my own test file was flagged before `--write`).

### DIFF
--- a/src/pages/api/track/block-render.ts
+++ b/src/pages/api/track/block-render.ts
@@ -7,6 +7,7 @@ import {
   normalizeSlotId,
   observeAppBlockLaunch,
 } from '~/server/metrics/app-block-runtime.metrics';
+import { logToAxiom } from '~/server/logging/client';
 import { boundAppBlockIdLabel } from '~/server/services/blocks/known-app-blocks.service';
 import { blockRenderSchema, blockRenderTrackerPayload } from '~/server/schema/track.schema';
 import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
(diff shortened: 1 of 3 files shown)


ZacxDev:
### `preview / component-tests` is red here for an unrelated, environmental reason

Chased rather than waved through, even though the check is report-only. Recording the evidence so nobody re-derives it.

**The failure is one test, in a component this PR cannot reach:**

```
FAIL  component (chromium)  src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx
      > 🔴 the search box does NOT write the URL per keystroke — only the debounced value
Test Files  1 failed | 123 passed (124)
Tests       1 failed | 1258 passed (1259)
```

Five independent lines of evidence, none of which is "it looks unrelated":

1. **No import path.** This PR's diff is three files — `src/pages/api/track/block-render.ts`, `src/server/services/blocks/app-views.service.ts`, and an API test. None is reachable from a component bundle; that test mocks `~/utils/trpc` and `next/router`.
2. **The same test passed ~10 minutes earlier** on #3626's preview run: **1259/1259**. So it is not broken on `main`.
3. **It is a real-timer race by construction.** Two `await search.fill()` calls must both complete inside a **300ms** debounce before the assertion reads zero writes. The test's own comment documents this exact mode: *"one slow fill and the debounce elapses mid-loop, a write lands, and the test goes red on a machine hiccup rather than on a regression."*
4. **It passes on this branch.** Ran that file locally against this PR's head: **12/12 passed**, real exit status captured (not a pipe artifact).
5. The check is explicitly **report-only, not blocking**.

**Worth flagging for whoever owns that test:** in my local run the case took **338ms** against its own 300ms debounce budget. The margin is thinner than the comment assumes — the fix from six fills to two widened it, but not enough that a loaded runner is safe. If it keeps flaking, driving the debounce with fake timers would remove the timing dependency entirely rather than trading margin for luck.

Merging on that basis.

PR #3631, opened 2026-08-04
feat(app-blocks): iframe init-fragment fast path + BLOCK_HELLO readiness announce

ZacxDev:
Host half of a two-repo wire-contract change. SDK side: **civitai/civitai-app-starters#212**.

**A.** `theme` / `renderMode` / `blockInstanceId` are also carried in the iframe URL **fragment**, so a block has them at document-parse time instead of after the handshake. **The token stays in the payload** — a URL leaks into referrers, history, proxy logs and screenshots, and must never carry a bearer credential.

**B.** The block **announces readiness** (`BLOCK_HELLO`) and the host pushes `BLOCK_INIT` in response, instead of relying purely on the blind 400 ms retry tick.

---

## 🔴 Compatibility matrix

Blocks are third-party code **already deployed**. Every cell is covered by a named test that was watched to fail under a deliberate mutation of the code it guards.

| Cell | What must hold | How it is satisfied | Evidence (test name) |
|---|---|---|---|
| **old SDK + new host** | a deployed block that reads no fragment and never announces still inits, on the schedule it gets today | `IframeInitController` is **untouched** on its existing paths: immediate post on `start()`, re-post every `INIT_RETRY_INTERVAL_MS` (400 ms), readiness timeout armed at `start()`. The fragment is inert to a block that does not read it — a fragment is never sent to the block's origin server, so it cannot perturb caching, routing or a param validator either | `OLD SDK + NEW HOST: a block that never announces gets the unchanged schedule` (asserts 1 immediate + 5 retries at 5 ticks, timeout not fired) |
| **new SDK + old host** | no fragment appended, announce unanswered ⇒ block behaves as today | *SDK-side claim.* An absent fragment decodes to `{}` and leaves the sentinel snapshot; the announce is fire-and-forget and expects no reply | SDK PR: `no fragment → sentinel snapshot, and BLOCK_INIT still works unchanged`; `the announce is not a precondition: BLOCK_INIT still resolves if the parent ignores it` |
| **new SDK + new host** | fragment read as a hint only; **payload stays authoritative** | Host emits the fragment; SDK seeds it into the **pre-init** snapshot. `BLOCK_INIT` then calls `snapshotFromInit`, which **replaces the whole snapshot**, overwriting all three fields | Host: `renders the iframe src with the init fragment appended`. SDK PR: `BLOCK_INIT OVERRIDES the fragment — the payload is authoritative`; `a failing history.replaceState … and the payload still wins` |
| **block never announces readiness** | must **not hang** the host — bounded fallback required | `notifyHello()` is an *accelerator*, not a gate. It never cancels the retry loop and never disarms the readiness timeout, so a block that neither announces nor acks still hits `onReadyTimeout` and surfaces the `timeout` fallback | `NEVER ANNOUNCES AND NEVER ACKS: the readiness timeout still fires — no hang` (also asserts the retry loop stops with it, rather than re-posting forever); `the announce does NOT cancel the retry loop or the readiness timeout` |

Three further properties that carry most of the real risk:

- **🔴 No credential in the URL.** Both hosts are asserted to render an `src` that does not contain the block bearer token, and whose fragment key set is *exactly* `blockInstanceId, civitai-block, renderMode, theme` — an enumeration, so a future edit that widens the input fails the test rather than silently leaking. (`the rendered src carries NO token — only the three declared fields`, on both hosts.)
- **🔴 NO-STOMP: a publisher already using the fragment is left alone.** `buildBlockIframeSrc` returns the src **unchanged** when it already carries a fragment. A block that puts state in its own hash — a hash router, a deep link — is the one population an appended fragment could break, and an existing fragment is the signal that we are looking at one. Those blocks keep today's exact URL and simply forgo the fast path. (`NO-STOMP: returns the src UNCHANGED when the publisher already uses the fragment`, covering both `#/dashboard` and `#section-2`.) See *Residual risk* below for the case this does **not** cover.
- **The origin is derived from the BASE src**, never from the fragmented one, so the `postMessage` target and the `expectedOrigin` pin cannot be influenced by anything this change adds.

## Fragment format (v1)

```
#civitai-block=v1&theme=dark&renderMode=iframe&blockInstanceId=bi_abc
```

A standard `URLSearchParams` body. `civitai-block=v1` is **both** the namespace marker and the format version: it lets a reader distinguish our fragment from a block app's own hash state, and an unknown version decodes to `{}` so a future v2 degrades to "no fast path" rather than to garbage.

🔴 **This is a MIRRORED contract, not a shared import.** civitai consumes the *published* `@civitai/app-sdk` dist, and the decoder ships in a version that is not published when this lands — so `blockIframeUrl.ts` carries its own encoder. Both repos pin the **same literal string** in a test, deliberately as a literal rather than a round-trip, because an expectation derived from the implementation cannot detect a format change. **Ship order is SDK first** (an unknown version is safe; a missing decoder just means no fast path).

## Handshake change

`IframeInitController` gains `notifyHello()`, wired in both hosts via `onMessage('BLOCK_HELLO', …)`. It posts `BLOCK_INIT` immediately rather than waiting out the current retry tick, and is **honored at most once per controller** — an announce is a one-shot signal from the block's transport, so answering every repeat would let a chatty frame amplify host work. An announce landing *before* `start()` is recorded and costs nothing, because `start()` posts init immediately on its own.

`hostHandlerParity.ts` gains a `BLOCK_HELLO` inventory entry, which is what makes "both live hosts register a handler" a **gated** claim rather than a hope.

## Where the change lives (and why not in the five supplier files)

The five iframe-URL suppliers are **untouched**. They hand a base `src` to a host; the host appends the fragment from values it already holds (`IframeHost` from the install + slot context, `PageBlockHost` from its props). One shared helper (`buildBlockIframeSrc`) behind one shared hook (`useBlockIframeSrc`) means the rule lives in one place instead of five — a predicate open-coded at five sites is the shape that gets fixed four times and stays wrong at the fifth.

### 🔴 The fragment fields are FROZEN at mount

`theme` **can** change while a block is mounted (the viewer toggles dark mode). If the fragment tracked it, the iframe's `src` **attribute** would change on a theme toggle — a navigation of a third-party frame where today there is none. At best that is a `hashchange` inside the block; at worst a reload that discards its in-progress state. In exchange for nothing: the host does not propagate live theme changes to blocks today either, because the SDK dedupes `BLOCK_INIT` and no theme-change message exists.

So `useBlockIframeSrc` captures the three fields on first render and memoizes on `baseSrc` alone. `baseSrc` is deliberately **not** frozen — if a publisher's manifest src genuinely changes mid-mount, the iframe re-navigates exactly as it does today.

## Propagating a *change* to render mode

Deliberately **not implemented**, and worth recording because the no-reload property is why a fragment was chosen over a query arg.

Nothing changes these three values for a mounted block today: `renderMode` is a per-install constant (`'iframe'` at every iframe call site — `[slug]/[[...path]].tsx:150`, `dev/[blockId].tsx:168`, `PageBlockHost.tsx:694`), `blockInstanceId` is immutable per mount, and `theme` changes are already not propagated. So freezing changes no observable behaviour.

The mechanism a fragment buys, when it is wanted: the host rewrites **only** the fragment of the existing `src` — a same-document fragment navigation, no reload — and the block observes it via a `hashchange` listener. That listener does not exist in the SDK yet. Because the fields are frozen today, adding it later is purely additive: nothing has to be un-done first.

## On `?dev=<token>` — left as-is, deliberately

`pages/apps/dev/[blockId].tsx:110` puts a token in a URL, which is exactly what this PR argues against. It stays, and here is why it is a different thing:

- It is **not** the block bearer token. It is an **edge forwardAuth entry token** that authenticates the *document navigation itself* — the `*.civit.ai` edge verifies it before the document is served. At that moment the frame does not exist, so there is no postMessage channel to carry it. Moving it out of the URL is not a refactor; it requires a different auth mechanism (a cross-site cookie on the dev host, `SameSite=None`, plus its own flow).
- The existing mitigations are real: it is **freshly minted on every SSR render** (never stale), **author-bound and host-bound**, short-TTL, and that route sets a **route-scoped CSP pinning `frame-src` to the exact dev host**.
- It is a **dev-tunnel-only** path, not a production block surface.

Changing it is a cross-repo protocol break (the tunnel edge is the consumer) with no bearing on this change. Flagging it as a standalone security item rather than smuggling it in here.

## Test matrix

🔴 **Every new test was watched to fail**, and required to fail via **its own named test** — not merely to turn the suite red. Two mutation sweeps, one per tier.

**Node ("unit") tier — 13 mutants, 13 killed:**

| Mutation | Killed by |
|---|---|
| `IframeHost` drops the `BLOCK_HELLO` handler | parity: `IframeHost.tsx is missing onMessage('BLOCK_HELLO')` |
| `PageBlockHost` drops the `BLOCK_HELLO` handler | parity: `PageBlockHost.tsx is missing onMessage('BLOCK_HELLO')` |
| `notifyHello()` becomes a no-op | `an announce posts BLOCK_INIT immediately, not at the next tick` |
| `notifyHello()` stops the retry loop (treats hello as an ack) | `the announce does NOT cancel the retry loop or the readiness timeout` |
| `notifyHello()` loses its once-only latch | `is honored at most ONCE` (51 sends observed vs 2 expected) |
| `notifyHello()` sends before `start()` | `an announce BEFORE start() sends nothing` |
| retry loop no longer re-posts | `a block that never announces gets the unchanged schedule` |
| NO-STOMP removed | `returns the src UNCHANGED when the publisher already uses the fragment` |
| fragment written as a **query arg** instead | `appends the fragment to a plain publisher src` (+3 more) |
| an extra field leaks into the URL | `never puts a token, viewer or context in the URL` (+5 more) |
| encoder stops escaping values | `percent-encodes a blockInstanceId so it cannot smuggle extra fragment keys` |
| `blockInstanceId` guard removed | `returns the src unchanged when there is no blockInstanceId` |
| URL parse failure throws instead of degrading | `returns an unparseable src unchanged rather than throwing` |

**Component ("browser") tier — 6 mutants, 6 killed.** These cover the two seams no pure unit test can reach, because they are claims about how the *component* is wired rather than what a helper computes:

| Mutation | Killed by |
|---|---|
| `IframeHost` renders the BASE src (fragment never reaches the DOM) | `renders the iframe src with the init fragment appended to the manifest src` |
| `IframeHost` hello handler no longer calls `notifyHello` | `a BLOCK_HELLO from the frame reaches IframeInitController.notifyHello` |
| `PageBlockHost` renders the BASE src | `renders the iframe src with the init fragment appended` |
| `PageBlockHost` hello handler no longer calls `notifyHello` | `a BLOCK_HELLO from the frame reaches IframeInitController.notifyHello` |
| `PageBlockHost` fragment built from the WRONG instance id | `renders the iframe src with the init fragment appended` |
| `IframeHost` fragment built from the WRONG theme | `renders the iframe src with the init fragment appended to the manifest src` |

The seam test asserts the **call** (`vi.spyOn(IframeInitController.prototype, 'notifyHello')`) rather than counting `BLOCK_INIT` posts, deliberately: a count would race the host's own 400 ms retry tick and could pass for the wrong reason. Each has a positive control (spy not called before the message) and a negative control (an adjacent message type `BLOCK_HELLO_NOT_REALLY` must not reach it).

**Two escapes in the first sweep, both fixed rather than explained away:**

1. *"empty-src guard removed"* **survived** — because `new URL('')` throws and the `catch` already returns the input, so `if (!baseSrc) return baseSrc` was genuinely dead code. **Removed the redundant guard** rather than adding a test for an unreachable branch. (The two escapes on the SDK side are documented in #212.)
2. Two parity mutants reported `RED-WRONG-TEST` purely because my expectation string omitted the `.tsx` the assertion actually prints. Re-run with the correct string: both KILLED.

**Suite results, post-rebase onto `42833e3b04`:**

| Gate | Result |
|---|---|
| `vitest --project unit src/components/AppBlocks` | 30 files / **547 tests** pass |
| `vitest --project component src/components/AppBlocks` | 27 files / **326 tests** pass |
| `pnpm typecheck` | **151 at base `42833e3b04`, 151 at HEAD — error SETS identical** (compared by `file(line,col) + TSxxxx`, since Prisma type-expansion text is not stable run-to-run). Zero new, zero gone, zero count changes. |

Test counts are **counted**, not read from an exit code; ANSI was stripped before grepping and all runs were redirected to files.

## 🔴 Residual risk I could not close

The NO-STOMP rule covers a publisher whose `iframe.src` **already carries** a fragment. It does **not** cover a hash-routing block whose declared src has *no* initial hash (e.g. `https://demo.civit.ai/` with a router that reads `location.hash` at import time and treats `#civitai-block=v1&…` as an unknown route).

What I established: neither the SDK, `blocks-react`, nor any of the six starter templates uses hash routing, and none declares a hash-router dependency (`grep` for `HashRouter|createHashRouter|useHashLocation|location.hash` across `packages/` + `starters/` = 0 hits; no `react-router`/`wouter`/`@tanstack/react-router` in any manifest). What I did **not** do: enumerate the live deployed block set and confirm none of them hash-routes. That is a query against production install data, outside this change's blast radius, and I did not want to generalise from a sample.

**Recommended sequencing:**
1. Merge #212, publish the SDK (decoder first — harmless on its own).
2. Before merging this PR, spot-check the live block URLs (< a dozen apps) for a hash router. It is a one-off read.
3. If any block hash-routes, the mitigation is one line in `blockIframeUrl.ts` — an allowlist/denylist by `blockId`, or making the fragment opt-in — and the SDK-side strip already covers new-SDK blocks that read the hash after first render (**not** at import time; do not read it as a guarantee).

Rollback is a one-line revert: `useBlockIframeSrc` returning `baseSrc`. The handshake half (`BLOCK_HELLO`) is independent of the fragment half and carries none of this risk — it is additive on both sides and cannot make a block fail to init.

https://claude.ai/code/session_01R7ZCxbgcsv3WfsSJrYdSCi

### DIFF
--- a/src/components/AppBlocks/IframeHost.browser.test.tsx
+++ b/src/components/AppBlocks/IframeHost.browser.test.tsx
@@ -97,6 +97,10 @@ vi.mock('~/components/BrowsingLevel/BrowsingLevelProvider', () => ({
 import { IframeHost } from '~/components/AppBlocks/IframeHost';
 // eslint-disable-next-line import/first
 import type { BlockInstall, ModelSlotContext } from '~/components/AppBlocks/types';
+// eslint-disable-next-line import/first
+import { IframeInitController } from '~/components/AppBlocks/iframeInitController';
+// eslint-disable-next-line import/first
+import { encodeBlockInitFragment } from '~/components/AppBlocks/blockIframeUrl';
 
 /**
  * Analytics Phase 2 — block render/impression beacon on the MODEL slot host.
@@ -359,5 +363,91 @@ describe('IframeHost GET_BUZZ_BALANCE handler (Phase 3, model.sidebar_top)', () 
     expect(mocks.balance).not.toHaveBeenCalled();
     expect(replies.last('BUZZ_BALANCE_RESULT')).toBeUndefined();
     replies.stop();
+  });
+});
+
+/**
+ * The iframe WIRE CONTRACT on the model-slot host — the two seams that no pure
+ * unit test can reach, because they are claims about how the COMPONENT is
+ * wired rather than about what the helpers compute:
+ *
+ *   A. the rendered `src` really carries the init fragment, and
+ *   B. a real `BLOCK_HELLO` arriving over the real postMessage bridge really
+ *      reaches `IframeInitController.notifyHello`.
+ *
+ * The behaviour behind each seam is pinned separately and deterministically:
+ * `__tests__/blockIframeUrl.test.ts` for the URL format, and
+ * `__tests__/iframeInitController.test.ts` for what `notifyHello` does (and,
+ * crucially, does NOT do — it never cancels the retry loop or the readiness
+ * timeout). Asserting the CALL here rather than counting BLOCK_INIT posts is
+ * deliberate: a count would race the host's own 400ms retry tick and could
+ * pass for the wrong reason.
+ */
+describe('IframeHost iframe wire contract (init fragment + readiness announce)', () => {
+  test('renders the iframe src with the init fragment appended to the manifest src', async () => {
+    renderWithProviders(<IframeHost {...baseProps} />);
+    await vi.waitFor(() => {
+      const el = page.getByTestId('block-iframe').element() as HTMLIFrameElement;
+      if (!el.contentWindow) throw new Error('not mounted yet');
+    });
+
+    const el = page.getByTestId('block-iframe').element() as HTMLIFrameElement;
+    const src = el.getAttribute('src') ?? '';
+    const url = new URL(src);
+
+    // The publisher's origin + path are untouched; only the fragment is added.
+    expect(url.origin).toBe(new URL(SAME_ORIGIN_SRC).origin);
+    expect(url.pathname).toBe(new URL(SAME_ORIGIN_SRC).pathname);
+    expect(url.hash.slice(1)).toBe(
+      encodeBlockInitFragment({
+        // `context.theme` is 'light' and the install renders in iframe mode.
+        theme: 'light',
+        renderMode: 'iframe',
+        blockInstanceId: install.blockInstanceId,
+      })
+    );
+  });
+
+  test('🔴 the rendered src carries NO token — only the three declared fields', async () => {
+    renderWithProviders(<IframeHost {...baseProps} />);
+    await vi.waitFor(() => {
+      const el = page.getByTestId('block-iframe').element() as HTMLIFrameElement;
+      if (!el.contentWindow) throw new Error('not mounted yet');
+    });
+
+    const src = page.getByTestId('block-iframe').element().getAttribute('src') ?? '';
+    // baseProps.token is the block bearer token. It must appear nowhere in the URL.
+    expect(src).not.toContain(baseProps.token);
+    const keys = [...new URLSearchParams(new URL(src).hash.slice(1)).keys()].sort();
+    expect(keys).toEqual(['blockInstanceId', 'civitai-block', 'renderMode', 'theme']);
+  });
+
+  test('a BLOCK_HELLO from the frame reaches IframeInitController.notifyHello', async () => {
+    const helloSpy = vi.spyOn(IframeInitController.prototype, 'notifyHello');
+    renderWithProviders(<IframeHost {...baseProps} />);
+    await vi.waitFor(() => {
+      const el = page.getByTestId('block-iframe').element() as HTMLIFrameElement;
+      if (!el.contentWindow) throw new Error('not mounted yet');
+    });
+    expect(helloSpy).not.toHaveBeenCalled(); // positive control: nothing yet
+
+    postFromBlock('BLOCK_HELLO');
+
+    await vi.waitFor(() => expect(helloSpy).toHaveBeenCalled());
+    helloSpy.mockRestore();
+  });
+
+  test('an unrelated message type does NOT reach notifyHello (negative control)', async () => {
+    const helloSpy = vi.spyOn(IframeInitController.prototype, 'notifyHello');
+    renderWithProviders(<IframeHost {...baseProps} />);
+    await vi.waitFor(() => {
+      const el = page.getByTestId('block-iframe').element() as HTMLIFrameElement;
+      if (!el.contentWindow) throw new Error('not mounted yet');
+    });
+
+    postFromBlock('BLOCK_HELLO_NOT_REALLY');
+    await new Promise((r) => setTimeout(r, 150));
+    expect(helloSpy).not.toHaveBeenCalled();
+    helloSpy.mockRestore();
   });
 });
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -37,6 +37,7 @@ import {
   projectBlockInitViewer,
 } from './projectBlockInit';
 import { IframeInitController, shouldStartInit } from './iframeInitController';
+import { useBlockIframeSrc } from './useBlockIframeSrc';
 import { usePostMessage } from './usePostMessage';
 import type { BlockInitPayload, BlockInstall, ModelSlotContext, SlotContext } from './types';
 import { dialogStore } from '~/components/Dialog/dialogStore';
@@ -606,14 +607,22 @@ export function IframeHost({
   // used everywhere else modelCtx.slotId is read in this component.
   const slotId = modelCtx.slotId ?? 'model.sidebar_top';
 
-  const iframeSrc = install.manifest.iframe?.src ?? '';
+  // The publisher's declared src. The rendered `src` adds the init-fragment
+  // fast path on top (see blockIframeUrl.ts); the ORIGIN is derived from the
+  // base so the postMessage target can never be affected by the fragment.
+  const baseIframeSrc = install.manifest.iframe?.src ?? '';
+  const iframeSrc = useBlockIframeSrc(baseIframeSrc, {
+    theme: modelCtx.theme ?? 'light',
+    renderMode: install.renderMode,
+    blockInstanceId: install.blockInstanceId,
+  });
   const expectedOrigin = useMemo(() => {
     try {
-      return new URL(iframeSrc).origin;
+      return new URL(baseIframeSrc).origin;
     } catch {
       return '';
     }
-  }, [iframeSrc]);
+  }, [baseIframeSrc]);
 
   // The EFFECTIVE sandbox handed to the iframe attribute below. Derive the
   // transport's opaque-origin mode from the SAME string so they can never
@@ -884,6 +893,26 @@ export function IframeHost({
     }, TOKEN_WAIT_TIMEOUT_MS);
     return () => clearTimeout(t);
   }, [status, token]);
+
+  // INVERTED HANDSHAKE: the block announces that its message listener is
+  // attached (`BLOCK_HELLO`) and we push BLOCK_INIT in response, instead of
+  // relying purely on the blind retry tick to eventually land after the
+  // listener exists.
+  //
+  // 🔴 PURELY ADDITIVE. `IframeInitController` still posts init immediately on
+  // start() and re-posts every INIT_RETRY_INTERVAL_MS until BLOCK_READY, and
+  // still arms the readiness timeout. A block on an older SDK never sends this
+  // message and is served exactly as it is today; a block that announces but
+  // never acks still times out. `notifyHello()` is a once-per-controller
+  // accelerator (see its doc comment), and a hello arriving before the
+  // controller exists is a no-op because `start()` posts init immediately
+  // anyway.
+  useEffect(() => {
+    const off = onMessage<unknown>('BLOCK_HELLO', () => {
+      controllerRef.current?.notifyHello();
+    });
+    return off;
+  }, [onMessage]);
 
   useEffect(() => {
     const off = onMessage<unknown>('BLOCK_READY', (raw) => {
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -89,6 +89,10 @@ vi.mock('~/utils/trpc', () => ({
 
 // eslint-disable-next-line import/first
 import { PageBlockHost } from '~/components/AppBlocks/PageBlockHost';
+// eslint-disable-next-line import/first
+import { IframeInitController } from '~/components/AppBlocks/iframeInitController';
+// eslint-disable-next-line import/first
+import { encodeBlockInitFragment } from '~/components/AppBlocks/blockIframeUrl';
 
 /**
  * W10 lazy-consent gap regression (page surface).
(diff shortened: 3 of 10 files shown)


ZacxDev:
## 🔴 Two claims in this PR body are wrong, and together they invert the risk assessment

An adversarial pre-merge audit found both; I re-verified each independently against `origin/main` before posting. **Do not merge on the current risk framing.**

### 1. NO-STOMP protects an empty set — `iframe.src` can never carry a fragment

The body describes NO-STOMP as "the compatibility guard for hash-routing block apps … an existing fragment is the signal that we are looking at one."

That signal is structurally unreachable. `stampCanonicalIframeSrc` (`src/server/services/blocks/manifest-normalize.ts`) unconditionally overwrites `manifest.iframe.src` with `canonicalIframeSrc()`, and its own doc comment states it: *"developer-supplied `iframe.src` is overwritten."* The manifest schema (`public/schemas/app-block/v1.json`) independently rejects a dev-set `iframe.src` — *"SERVER-OWNED… Present here only so the schema can reject dev-set values."*

So the no-stomp branch in `blockIframeUrl.ts` cannot fire for a production block, and the "residual risk I could not close" is **not a tail case — it is the entire deployed population**.

### 2. The `0 hits` grep returns hits, and two starters recommend hash routing

The body states: *"grep for `HashRouter|createHashRouter|useHashLocation|location.hash` across `packages/` + `starters/` = 0 hits."*

Re-running that exact pattern against `civitai-app-starters@origin/main`:

```
starters/react-pwa/.claude/commands/add-route.md:16
  - **Hash routing:** read `window.location.hash`; swap rendered component.

starters/svelte-pwa/.claude/commands/add-route.md:16
  - **Hash routing:** track `window.location.hash` with a Svelte effect.
```

The svelte file also recommends `svelte-spa-router` by name, which is hash-based.

Combined with (1): a block scaffolded from either PWA starter and told "add a route" is steered toward the one pattern this change breaks. On first load it sees `location.hash === "#civitai-block=v1&theme=…"`, resolves no route, and renders its fallback instead of its home view. The SDK strip does not rescue it — the SDK is merged but **not yet published** (npm latest is still `@civitai/app-sdk@0.30.0`), so today no deployed block decodes, strips, or announces. Even once published, the strip runs after `IframeTransport` construction, i.e. after import-time hash reads.

### What has not been done

Nobody has enumerated the live installs to check whether any actually hash-routes. That is this PR's own recommended step 2, and it is now the **only** mitigation that exists — NO-STOMP cannot be counted.

### Also worth fixing before merge

- **The FREEZE has zero test coverage in either tier.** It is the property this PR names as its safety argument, and it is absent from the PR's own mutant table. Replacing the frozen ref with live `fields` + full deps survives both tiers (node 31/31, browser 38/38). A positive control on the same files confirms the harness can go red, so this is a real gap rather than a broken instrument.
- **The frozen `theme` is wrong for "auto" color-scheme viewers.** All three theme sources use `useComputedColorScheme('dark')`, which under Mantine's `getInitialValueInEffect: true` returns `'dark'` on first render and commits the OS value in an effect. The freeze captures first render — so for an auto + light-OS viewer the fragment says `dark` while `BLOCK_INIT` says `light`, and the fast path *creates* the theme flash it exists to remove. Not a correctness break (the payload wins), and the cookie default is `'dark'` rather than `'auto'`, so the cohort is limited to viewers who explicitly chose Auto.
- **`blockInstanceId` now appears in `location.href` inside the block document.** On the model-slot host that is a per-user-per-install identifier. The block already receives it over `BLOCK_INIT`, but a value in the URL is auto-captured by any error/analytics SDK the block loads, so it egresses to third parties without the author choosing to send it. This PR makes exactly that argument for the token; it should apply to the identifier too. (Network side is clean — fragments aren't sent to servers, and `referrerPolicy="no-referrer"` is set on both iframes.)

### Recommendation

1. Enumerate the live blocks and check for hash routing — the only remaining mitigation.
2. Put the fragment behind a kill-switch or per-block opt-in so the blast radius is revocable without a deploy. Rollback today is a code revert.
3. Add the freeze test.
4. Correct this PR body on both points above before it is merged and cited.

If step 1 finds a hash-routing block, this needs rework rather than a fix-up.

**None of this impugns the implementation** — the mechanisms behave as described, the four compatibility cells hold as written, no credential reaches the URL, and 14 of 17 audit mutants died to their own named tests. The defect is in the risk assessment, not the code.

ZacxDev:
## Live-block enumeration complete — the step this PR was blocked on

Recommended step 2 has been done. Full population enumerated (not sampled), with the detector positive-controlled before any negative was trusted.

### Bottom line: not safe as an unconditional append — and it buys nothing today

**No live block hash-*routes*.** The catastrophic case — fragment present, no route resolves, block renders its 404 — does not exist in the current live set.

**But no deployed block can decode the fragment either.** Across all 20 bundles: `BLOCK_HELLO` ×0, `civitai-block=v1` ×0. The SDK half is merged but **not published** (npm latest is still `@civitai/app-sdk@0.30.0`). So shipping the fragment today takes on risk for **zero benefit** until blocks rebuild against a decoding SDK.

That is the decisive point. The question is not "is this risk tolerable" but "why accept any risk for a fast path nothing can currently use."

### What the enumeration found

| Finding | Detail |
|---|---|
| `playable-collections` | Reads `location.hash` on first load and strips it after. Fails closed **only by luck** — it requires a `c` key our fragment does not carry. Not a router, so no 404; but it is a live block that reads the hash we would start populating. |
| Off-site listings | `comfy`, `cosmetic-studio`, `radio`, `vitrine` are outbound `<a href>` links, **never iframed** — `AppListingDetailBody.tsx:87-108` bans re-adding a bare `<iframe src={liveUrl}>`, pinned by a gate in `__tests__/appListingDetailView.test.ts:414`. `cosmetic-studio` and `vitrine` tripped hash signatures but are **irrelevant to this PR**. |
| Suspended blocks (12) | Genuinely out of the live population. Token mint re-queries and 403s on non-approved (`src/pages/api/v1/block-tokens/index.ts:863`), and `BlockHost.tsx:39-47` never mounts `IframeHost` until a token exists — so even a stale cache entry yields a loading skeleton, never a rendered iframe. |
| 🔴 Dev tunnel | `resolveDevPageBlockForAuthor` (`block-registry.service.ts:2033`) has **no status filter at all**, deliberately, so authors can iterate on drafts. It mounts the same real `PageBlockHost`, and this PR's own test pins that URL shape — `buildBlockIframeSrc('https://tunnel.example.com/?dev=abc123', FIELDS)` → fragment appended. **The fragment therefore reaches arbitrary unpublished code that no query can enumerate**, quite possibly scaffolded from the two `add-route.md` files that recommend hash routing. Lowest stakes (author-only, self-bound token, forced-SFW, kill-switched) — but the one surface where the hazard is live today and unmeasurable. |

### Recommendation

**Split this PR.**

1. **Land the `BLOCK_HELLO` / ready-handshake half now.** It carries none of this risk, and it is the half that removes the 400 ms retry loop.
2. **Gate the fragment half** behind an allowlist (or the kill-switch suggested earlier), with `playable-collections` excluded, until blocks rebuild against an SDK that decodes it. At that point the fragment starts paying for itself and the allowlist can widen as blocks upgrade.

### Named remaining gaps

- `who-am-i` — suspended, NXDOMAIN, **undetermined** rather than scored safe.
- `sensei` — one pending publish request, no subdomain deployed; would become block #10.
- The dev-tunnel surface — structurally unenumerable, as above.
- This is a **point-in-time read (2026-08-05)**. Two of the nine live blocks were approved within the last eight days, so the population moves.

### Method note

The earlier "0 hits" grep in the PR body had no positive control and was wrong. This enumeration validated its detector against files that genuinely mount a host (7 and 5 hits) before trusting any zero, and re-checked the two apparent hash hits by reading them in context rather than counting matches.

ZacxDev:
Closing in favour of the split recommended by the enumeration above.

**Both findings against this PR's body are confirmed — I re-verified each against `origin/main` rather than taking them on trust:**

1. **NO-STOMP protects an empty set.** `stampCanonicalIframeSrc` does `manifest.iframe = { ...existing, src: canonicalIframeSrc(slug, appsDomain) }` — unconditional. So the guard this body called "the compatibility guard for hash-routing block apps" cannot fire for a production block, and the hazard is the whole deployed population, not a tail case.

2. **The `0 hits` grep was wrong, and here is the mechanism.** The invocation passed `--include=*.ts --include=*.tsx --include=*.js --include=*.jsx --include=*.svelte`, **which excludes `.md`** — and every real hit is Markdown. The "positive control" I ran then checked the pattern against `package.json` files, i.e. **a different file set than the claim**, so it passed and told me nothing. Re-run unfiltered it is **five** locations, one more than reported: `svelte-pwa/README.md:71`, `svelte-pwa/AGENTS.md:65`, `svelte-pwa/.claude/commands/add-route.md:16` and `:17`, `react-pwa/.claude/commands/add-route.md:16`.

I also independently corroborated the dev-tunnel finding from the repo's own comment in `src/pages/api/v1/block-tokens/index.ts`: *"the SSR dev route (`resolveDevPageBlockForAuthor`) mounts an OWNED app at ANY status"*.

And one correction **to the audit**, in its favour: the frozen-`theme` "auto" problem affects **four** theme sources, not three — `ModelVersionDetails.tsx:176` (the model slot) uses `useComputedColorScheme('dark')` the same way as the run, dev and review hosts.

## Replaced by

| | |
|---|---|
| #3634 | **ready-handshake half** (`BLOCK_HELLO`) — carries none of the above; ready to land |
| #3635 | **fragment half** — ships **gated OFF** (empty allowlist), dev tunnel refused unconditionally, plus the freeze test and the corrected reasoning |

One thing I want to flag about the split framing: the handshake half was described as "the half that removes the 400 ms retry loop". **It does not remove it, and must not** — the same enumeration found `BLOCK_HELLO` × 0 across the live bundles, so nothing announces today and the retry loop is doing 100 % of the work. It stays as the bounded fallback; `notifyHello()` is only an accelerator layered on top.

The audit comments stay here as the record of why the split happened.

PR #3632, opened 2026-08-04
Check the first table in a parenthesized join

ZacxDev:
Small gap in the `query_database` scope check from #3584.

`FROM ("A" JOIN "B" ON ...)` is a relation position rather than a sub-select, but the walk treated any `(` after FROM/JOIN as a sub-select and dropped its relation expectation. The first table inside then matched no branch and went unchecked against the allowed list. Anything after it — the trailing side of the join, comma items, nested sub-selects — was still checked, so it was just the leading one.

Fix keeps the expectation when entering the paren and lets a `SELECT` directly inside be the thing that marks it a sub-select. Everything else stays in relation position and gets checked. That keeps it conservative instead of teaching it to actually parse joins, which is the trade the file already makes elsewhere.

Tests: 3 new refusal cases (plain join, `CROSS JOIN`, doubled parens) are red on the old implementation, green here. Also added an allowed parenthesized join so this can't pass by just refusing more — that one and the trailing-position case pass either way, so they're invariant guards, not regression coverage.

Mutation check: dropping the carry kills the 3 new tests; dropping the `SELECT` escape kills the 3 existing sub-select tests. Each half is pinned by a different set. A third mutation (resetting the flag on `)`) survived because that line turned out to be unreachable — the flag is rewritten on every `(` — so I removed it rather than ship a line nothing can justify.

Gates: 48/48 in `src/server/freshdesk-agent/__tests__/`, `pnpm typecheck` added-vs-main 0 (12 pre-existing `event-engine-common` TS2307s, identical set with this reverted), eslint and prettier clean.

### DIFF
--- a/src/server/freshdesk-agent/__tests__/freshdesk-query-scope.test.ts
+++ b/src/server/freshdesk-agent/__tests__/freshdesk-query-scope.test.ts
@@ -69,6 +69,11 @@ describe('checkQueryScope — statements it confirms', () => {
     ],
     ['an inline VALUES list in the FROM list', 'SELECT x FROM (VALUES (1), (2)) AS t(x)'],
     [
+      'a LATERAL sub-select, whose keyword sits between JOIN and the relation',
+      'SELECT * FROM "User" u, LATERAL (SELECT 1) s',
+    ],
+    ['a LATERAL inline VALUES list', 'SELECT * FROM "User" u, LATERAL (VALUES (1)) AS t(x)'],
+    [
       'an inline VALUES list joined to an allowed table',
       'SELECT * FROM (VALUES (1)) AS t(x) JOIN "User" u ON u.id = t.x',
     ],
@@ -199,6 +204,27 @@ describe('checkQueryScope — relations outside the allowed set', () => {
     ['bare, reached through a join', 'SELECT * FROM "User" u JOIN values v ON true'],
     ['bare, case-folded', 'SELECT * FROM VaLuEs'],
     ['bare, in a sub-select', 'SELECT * FROM "User" WHERE id IN (SELECT id FROM values)'],
+    // The head item of the outer paren is itself a parenthesized group, so a
+    // position-tracking gate can leave a stale "this is the head" state behind
+    // and wave the later join through. The discriminator is the next token, not
+    // the position, so these must refuse at any nesting depth.
+    ['bare, after a nested sub-query head', 'SELECT * FROM ((SELECT 1) s JOIN values v ON true)'],
+    [
+      'bare, after a nested join head',
+      'SELECT * FROM (("User" u JOIN "Image" i ON true) JOIN values v ON true)',
+    ],
+    [
+      'bare, after a LATERAL nested head',
+      'SELECT * FROM (LATERAL (SELECT 1) s JOIN values v ON true)',
+    ],
+    // `values` LEADING a parenthesized joined_table is a relation reference too:
+    // it is followed by an alias, never by `(`.
+    // The next token here is a QUOTED identifier whose text happens to be `(`
+    // (a legal, if perverse, alias). Only the token's kind separates it from a
+    // real `VALUES (` clause.
+    ['bare, aliased with a quoted paren', 'SELECT * FROM values "("'],
+    ['bare, leading a parenthesized CROSS JOIN', 'SELECT * FROM (values v CROSS JOIN "User" u)'],
+    ['bare, leading a parenthesized join', 'SELECT * FROM (values v JOIN "User" u ON true) x'],
   ])('refuses `values` used as a relation name — %s', (_label, sql) => {
     expect(checkQueryScope(sql).ok).toBe(false);
   });
--- a/src/server/freshdesk-agent/__tests__/freshdesk-query-scope.test.ts
+++ b/src/server/freshdesk-agent/__tests__/freshdesk-query-scope.test.ts
@@ -67,6 +67,11 @@ describe('checkQueryScope — statements it confirms', () => {
       'a parenthesized join of allowed tables',
       'SELECT * FROM ("User" JOIN "Model" ON true) LIMIT 5',
     ],
+    ['an inline VALUES list in the FROM list', 'SELECT x FROM (VALUES (1), (2)) AS t(x)'],
+    [
+      'an inline VALUES list joined to an allowed table',
+      'SELECT * FROM (VALUES (1)) AS t(x) JOIN "User" u ON u.id = t.x',
+    ],
     [
       'a sub-select in the FROM list followed by another relation',
       'SELECT * FROM (SELECT id FROM "User" LIMIT 5) x, "Model" m',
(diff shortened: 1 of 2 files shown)


ZacxDev:
Pushed a second commit after an audit, and correcting two things in the original description rather than quietly editing them.

**Corrections to the body above**

- It says a third mutation survived so I "removed" the `)` reset line. That's misleading — that line never existed on `main`. I added it while working, found nothing could kill it, and dropped it before the first commit, so the diff never contained it. Nothing pre-existing was removed. The substance held up (re-adding it changes no verdict), but the framing implied a cleanup of existing code that didn't happen.
- The typecheck baseline is 9 `TS2307` plus 3 `TS7006` induced by them, not "12 TS2307". Still identical with the change reverted, so added-vs-main is still 0.

**What the second commit changes**

`FROM (VALUES (1),(2)) AS t(x)` is legal and worked before this branch — carrying the relation expectation into the paren started rejecting it, since only `SELECT` resolved the paren to a sub-query. That's a false rejection, it wasn't on the documented list the file header promises is complete, and the error text named CTEs and schema prefixes, so the model had no applicable rewrite and would just retry the same shape. `VALUES` now resolves the expectation like `SELECT`. Anything nested inside is still walked — there's a test for `(VALUES ((SELECT id FROM "Session")))`.

I also removed the per-depth array. It couldn't change the verdict on any statement Postgres accepts, only which error an already-invalid one got, and three mutations of it survived the suite. A plain reserved-word escape is the part doing the work.

Mutants re-run after the change, all killed: clearing the expectation on `(` kills the 3 parenthesized-join tests, dropping the escape kills 6, `SELECT`-only kills 3, `VALUES`-only kills 3, firing on any identifier kills 4.

51/51 in `src/server/freshdesk-agent/__tests__/`, gates still clean. The 3 parenthesized-join tests remain red on `main`; the `VALUES` tests pass there, so those are guards against a regression this branch introduced rather than coverage of an existing bug.

One thing the audit turned up that is **not** from this PR and needs its own change: `TABLE "X"` reaches a relation without any allowlist check (e.g. via a set operation), and the quoted form also slips the `pg_` prefix guard. It's live on `main` today. This PR narrows one member of that class rather than widening it, so it's not a reason to hold this. Filing separately.

ZacxDev:
Third commit. A re-audit of the second one found that **the second commit was wrong, and its own comment asserted the reason it was right** — correcting that here since I stated it in the comment above too.

I wrote that `SELECT` and `VALUES` are both reserved words and therefore can't be bare relation names. That's true of `SELECT` and false of `VALUES`. Measured against a real server:

- `select` → `pg_get_keywords.catcode = 'R'` (reserved)
- `values` → `catcode = 'C'` (**unreserved**), and `CREATE TABLE values (id int)` followed by `SELECT * FROM values` both succeed

So resolving the relation expectation on a bare `values` waved through a real relation reference. `SELECT * FROM values`, `SELECT * FROM "User" u, values v`, `SELECT * FROM ("User" JOIN values ON true)` and the case-folded `FROM VaLuEs` all went refused → allowed between commit 1 and commit 2. Not exploitable as things stand — no model or `@@map` named `values` exists in the schema, and the double-quoted `FROM "values"` was still refused — but it's the one bound on which relations model-written SQL can reach, and it was failing open on a bare identifier.

`VALUES` is a from_item only as the head of a parenthesized sub-query (bare `FROM VALUES (1)` is a syntax error), so it's now accepted in exactly that position via a one-shot per-depth flag. `SELECT` stays ungated because it genuinely can't be a relation name.

That flag is the one commit 2 deleted for being a no-op. It *was* a no-op — for `SELECT` alone. It became load-bearing the moment `VALUES` was added beside it in the same commit, which is what made the two changes look independently safe and not be.

Also corrected the docblock sentence claiming relation position accepts "a double-quoted identifier or an opening paren, nothing else", which stopped being true once sub-query heads were accepted.

**Mutants** — dropping the paren-head guard kills the 6 bare-`values` tests; making the flag non-one-shot kills the inside-a-join case; setting it unconditionally kills the sub-select case; swapping the `)` operand order kills the new test below. Each dies on the case its clause exists for.

**One survivor, stated rather than buried:** moving the `lateral` fall-through below the flag read changes no verdict. `LATERAL` can't precede `VALUES` at the same paren depth in valid SQL, so no discriminating input exists. The ordering is defensive, not pinned.

Also added a test for `SELECT * FROM (SELECT id FROM "User") x, "Session" m`. That's a **pre-existing** gap — the `)` handler's operand order is byte-identical to `main` — and nothing covered it.

60/60, typecheck added-vs-main 0, eslint and prettier clean.

ZacxDev:
Fourth commit. The position gate I added last round was itself wrong, found by another re-audit.

It tracked "this paren stood in relation position" per depth, but never consumed that state when the head item was **itself** a paren — so the flag went stale one level up:

```sql
SELECT * FROM ((SELECT 1) s JOIN values v ON true)
SELECT * FROM (("User" u JOIN "Image" i ON true) JOIN values v ON true)
```

Both valid SQL that reads a relation named `values`, both refused on `main`, both allowed by the previous commit. Net-new hole in exactly the thing this branch exists to close, and the second one is the shape my own comment promised stayed refused, one nesting level deeper.

**Position was the wrong discriminator.** The next token is the right one, and it's exact in both directions — verified against a real server rather than reasoned from the grammar:

- a `VALUES` *clause* is always immediately followed by `(`
- a relation named `values` never can be: `FROM values (id)` and `FROM values(1)` are both **syntax errors**, because a `C`-category keyword can't be a function name. Only `FROM values v (id)` parses, and there an alias sits in between.

So the per-depth flag is gone again and `values` is accepted only when the very next token is an opening paren. That also closes a case which was open on `main` and which my previous two commits both left open — `values` **leading** a parenthesized join, `FROM (values v CROSS JOIN "User" u)` — since there it's followed by an alias.

New tests cover the nested-head shapes, the leading-a-join shapes, and `FROM values "("` — whose next token is a quoted identifier whose *text* is `(`, so the only thing separating it from a real clause is the token's kind. That one is legal SQL, verified.

**Mutants, all killed:** dropping the lookahead kills 11; expecting `)` kills 4; dropping the kind check kills 1; reading the current / the token-after-next kills 3 and 2; clearing the expectation on `(` kills 5. Deleting the `LATERAL` fall-through now kills 2 — it was **unpinned** before this commit, which also retires the "no valid SQL discriminates the lateral ordering" claim I made last round. That claim was wrong: a discriminating input existed, it just depended on the stale-flag bug, and it's moot now the flag is gone.

68/68, typecheck added-vs-main 0, eslint and prettier clean.

Worth recording since it caught me twice: **a mutation sweep is only as good as its harness.** Three of my mutants this round silently failed to apply (a quoting error), and the runner happily reported the *previous* file's results — one of which read as "surviving mutant". I now assert the pattern matched exactly once and fail loudly otherwise.

ZacxDev:
Note on the red `preview / unit-tests` that was showing on `370adc9939`, since it's the kind of thing that gets waved through on the "report-only, not blocking" label: it's green again now, and I don't think it was the change.

The status across the five commits reads pass / pass / pass / **fail** / pass. The last commit is the previous one plus a comment-only edit — I checked the diff has no non-comment lines — so the code that failed and the code that passed immediately after are the same. The blocking `Unit tests` job was green on the failing SHA too.

Locally the full `unit` project shows 60 file-level failures with **zero** test failures, all `Cannot find module '.../event-engine-common/...'` — the sibling module that also produces the 12 pre-existing typecheck errors on `main`, and none of them in `freshdesk-*`. So my checkout can't reproduce the preview environment either way.

Calling it a flake on the strength of identical code passing, not on a bare re-run. I don't have the preview job's log, so I can't name what actually failed — if this check goes red again on this branch it's worth a real look rather than another retry.

PR #3635, opened 2026-08-05
feat(app-blocks): iframe init-fragment fast path, shipped GATED OFF — half 2 of 2

ZacxDev:
**Half 2 of 2.** Split out of #3631 after a live-block enumeration and an adversarial audit, both of which are on that PR. Handshake half: #3634. SDK counterpart: civitai/civitai-app-starters#212 (merged, **not yet published**).

# 🔴 This ships GATED OFF and is inert in production

`BLOCK_INIT_FRAGMENT_ALLOWLIST` is **empty**. No block on any surface gets the fragment. That is the intended state until blocks rebuild against an SDK that decodes it, and it is pinned by a test.

The reason is not risk tolerance, it is that **the fast path currently buys nothing**: a full enumeration of the 20 live bundles found `civitai-block=v1` × 0 and `BLOCK_HELLO` × 0, because the SDK half is merged but unpublished (npm latest is still `@civitai/app-sdk@0.30.0`). Shipping it enabled would take on risk for zero benefit.

---

# Two corrections to #3631's body

Both were mine, both were wrong, and together they inverted that PR's risk assessment. They are corrected here in the source comments as well as in this body.

### 1. NO-STOMP protected an empty set

#3631 called the no-stomp branch "the compatibility guard for hash-routing block apps … an existing fragment is the signal that we are looking at one."

**That signal is structurally unreachable.** `stampCanonicalIframeSrc` (`src/server/services/blocks/manifest-normalize.ts`) does:

```ts
manifest.iframe = { ...existing, src: canonicalIframeSrc(slug, appsDomain) };
```

— unconditional, and its own doc comment says *"Any developer-supplied `iframe.src` is overwritten."* The manifest schema independently rejects a dev-set `iframe.src`, and the dev-tunnel URL is server-derived too. So **no** src reaching `buildBlockIframeSrc` can carry a publisher fragment, the branch cannot fire for a production block, and the hazard it was claimed to cover is **the entire deployed population, not a tail case**.

The branch is retained — it is cheap and correct if such an src ever does arrive — but it is no longer counted as a mitigation anywhere. The real protection is the gate below.

### 2. The `0 hits` grep was wrong, and I can show exactly how

#3631 claimed: *"grep for `HashRouter|createHashRouter|useHashLocation|location.hash` across `packages/` + `starters/` = 0 hits"*, with a "positive control".

The invocation passed `--include=*.ts --include=*.tsx --include=*.js --include=*.jsx --include=*.svelte`, **which excludes `.md`** — and every real hit is in Markdown. My "positive control" then checked that the pattern shape could match entries in `package.json`, i.e. **a different file set than the claim**. That is the failure mode of validating an instrument other than the one you intend to trust: the control passed and told me nothing.

Re-run without the filter, against `civitai-app-starters@main`, this is **five** locations — more than the audit reported:

```
starters/svelte-pwa/README.md:71                      "drop in `svelte-spa-router` or a hash router"
starters/svelte-pwa/AGENTS.md:65                      "Hash router or `svelte-spa-router`"
starters/svelte-pwa/.claude/commands/add-route.md:16  "track `window.location.hash` with a Svelte effect"
starters/svelte-pwa/.claude/commands/add-route.md:17  "`svelte-spa-router`"  (hash-based)
starters/react-pwa/.claude/commands/add-route.md:16   "read `window.location.hash`; swap rendered component"
```

So the starters actively steer authors toward the one pattern the fragment perturbs — across README, AGENTS.md **and** the slash-command docs.

---

# The gate

`src/components/AppBlocks/blockInitFragmentGate.ts`. Two independent axes, checked in this order:

1. **Surface** — `dev-tunnel` **and `review-preview`** are refused **unconditionally**. Eligible surfaces are `model-slot` and `page-run` only.
2. **Denylist** — beats the allowlist.
3. **Allowlist** — explicit opt-in only; **empty today**.

### How it is keyed, and why the surface is a separate axis

Keyed on `blockId` **or** `slug` (whichever the surface knows). But an allowlist keyed on block identity **structurally cannot** cover the dev tunnel, which is exactly what was asked:

`resolveDevPageBlockForAuthor` applies **no status filter at all** — deliberately, so authors can iterate on drafts. The repo states this itself in `src/pages/api/v1/block-tokens/index.ts`: the SSR dev route *"mounts an OWNED app at ANY status"*. It mounts the same real `PageBlockHost` as production. And critically, **the tunnel serves the same `blockId` the author will eventually publish** — so allowlisting a block for production would silently enable it on that author's unpublished draft, which is precisely the code the PWA starters steer toward hash routing.

So the surface is passed explicitly by each of the three `PageBlockHost` call sites as a **required prop**, and the refused surfaces return `false` before the allowlist is ever consulted.

⚠️ **Correction:** the type-error guarantee covers `PageBlockHost` **only**. `IframeHost` hard-codes `surface: 'model-slot'` internally (`IframeHost.tsx:624`), so a new *model-slot* host would inherit that literal silently rather than being forced to choose. The earlier prose generalised this and was wrong.

| Call site | `surface` |
|---|---|
| `pages/apps/run/[slug]/[[...path]].tsx` | `page-run` |
| `pages/apps/dev/[blockId].tsx` | `dev-tunnel` → **always refused** |
| `components/Apps/ReviewBlockPreviewHost.tsx` (serves both review entry points) | `review-preview` → **also always refused** (see below) |
| `IframeHost` (model slot) | `model-slot`, hard-coded |

**Answer to the question asked: yes, the gate excludes the dev tunnel specifically, by construction rather than by omission** — and there is a mutation proving it (removing the check is killed by a test where the same `blockId` returns `true` on another surface).

### Denylist

`playable-collections` — it reads `location.hash` on first load and fails closed against our fragment **only by luck** (it wants a `c` key we do not carry). That is a coincidence, not a contract. The denylist is checked **before** the allowlist so a future widening cannot silently re-enable it.

### 🔴 Honest limitation: this is an opt-in allowlist, not a runtime kill switch

Enabling or disabling a block is a **deploy**. With the allowlist empty there is nothing to revoke in a hurry, so the practical exposure is nil — but if a genuine runtime switch is wanted, its home is a flag in the feature-flags service, which is outside this module's ownership and deliberately not added here. Say the word and I'll wire it.

---

# The freeze test (F4) — added regardless of gating

`BlockInitFragmentFreeze.browser.test.tsx`. The audit's finding was that the mount-time freeze — **the property #3631 named as its safety argument** — had zero coverage in either tier.

Reproduced and now closed. **Red-at-base evidence**, i.e. the exact mutations the audit reported as surviving:

| Mutation | Before | Now killed by |
|---|---|---|
| frozen ref → live `fields` **+ full deps** | survived both tiers pre-PR | `🔴 a theme change leaves the src BYTE-IDENTICAL` |
| 🔴 frozen ref → live `fields`, **deps UNCHANGED** | ⚠️ **survived this PR's own tests** | `🔴 THE DISCRIMINATING CASE: toggle theme, THEN move baseSrc` |
| `baseSrc` frozen too | survived both tiers pre-PR | `baseSrc is deliberately NOT frozen — it still re-navigates` |
| gate ignored (always append) | n/a (new) | `when the gate is OFF the src is the bare base` |
| 🔴 per-instance re-capture removed | n/a (the C fix) | `🔴 SOFT-NAV appA→appB re-captures` + `a blockInstanceId change RE-CAPTURES` |
| `blockInstanceId` dropped from memo deps | n/a (the C fix) | `🔴 a blockInstanceId change RE-CAPTURES` |
| 🔴 production binding args **swapped** | ⚠️ **survived all 9 gate tests** | `🔴 the PRODUCTION binding passes allow/deny in the RIGHT ORDER` |
| 🔴 `review-preview` refusal removed | n/a (the D fix) | `🔴 refuses review-preview even for a block that IS allowlisted` |

⚠️ **The earlier version of this table overstated row 1.** Its mutant changed *both* the ref and the dep array, and the dep-array half is what the other tests detect — so **the ref itself was uncovered**. Row 2 is the single mutant that isolates it, and it passed every test in this PR until an audit found it. Corrected rather than deleted, because the overstatement is the thing worth remembering.

One mutant **survived and should**: adding `theme` to the memo deps. The memo body reads `frozenFields.current`, so the forced recompute produces a byte-identical string — an **equivalent mutant**, not a coverage gap. The 7/7 pass is itself the proof of equivalence.

Two deliberate design points:

- **It tests the hook, not a host.** With the gate off, a host-level test renders a `src` with *no fragment at all* and could not observe the freeze — vacuous by construction. The probe drives the hook with `enabled: true`, exercising the invariant the gate will eventually expose.
- **Component tier, not node.** `useBlockIframeSrc` is a hook; the unit project runs `environment: 'node'` with no renderer.

Worth recording: this file **failed on its first run** — all 5 cases, `Cannot find element with locator` against an empty `<div />`. Not a freeze failure: React 18 commits asynchronously through `createRoot`, and I read the DOM synchronously after render. Every other browser suite here awaits first. Fixed in the helper, with the reason written down.

The gate predicate is separately mutation-tested via an **injectable** form (`blockInitFragmentEnabledWith`). Testing only the production binding would be vacuous — its allowlist is empty, so every call returns `false`, and `false` cannot distinguish "dev tunnel refused first" from "nothing allowlisted". The suite drives it with a **non-empty** allowlist and opens with an explicit **positive control** that the predicate can return `true` at all.

---

# Fixed in response to the audit (not deferred)

- **🔴 Soft-nav `appA→appB` leaked the previous app's identity.** `PageBlockHost` is rendered with **no `key`** (and `_app.tsx` renders `<Component>` with none), so a soft nav reuses the component instance: same mount, new `blockInstanceId`. The freeze was per-*mount*, so app B's document received app A's id **and** app A's mount-time theme while `BLOCK_INIT` delivered B's — the block seeing two contradicting identities. Every other per-mount latch in that file had already been re-scoped to the instance for exactly this reason; this one had not. Now frozen **within** an instance, re-captured **across** one. Model slot was never affected (`BlockSlotClient.tsx:145` keys by `blockInstanceId`).
- **🔴 Two tests were asserting the leak as intended behaviour** and have been **inverted**, not annotated. A test that blesses a bug is worse than no test — the next reader takes it as the contract.
- **🔴 `review-preview` now refused** alongside `dev-tunnel`.

# Deferred by gating (documented, not fixed)

Both are properties of the fragment and are unreachable while the allowlist is empty:

- **Frozen `theme` is wrong for "auto" viewers.** All **four** theme sources use `useComputedColorScheme('dark')` with no options — the run page (`:98`), the dev tunnel (`:143`), the review host (`:68`) and the model slot (`ModelVersionDetails.tsx:176`), one more than the audit reported; so Mantine's DEFAULT `getInitialValueInEffect: true` applies (other call sites in this repo pass `false` explicitly when they need the true value at first render, which confirms the default), and it returns `'dark'` on first render and commits the OS value in an effect. The freeze captures first render, so an auto + light-OS viewer would get `theme=dark` in the fragment while `BLOCK_INIT` says `light` — the fast path would *create* the flash it exists to remove. Not a correctness break (payload wins), and the cookie default is `'dark'` rather than `'auto'`, so the cohort is limited to viewers who explicitly chose Auto. **Must be fixed before any block is allowlisted.**
- **`blockInstanceId` lands in `location.href`** inside the block document, where an error/analytics SDK the block loads will auto-capture it. ⚠️ **Correction — this is weaker than #3631 claimed.** It is **not** a per-user-per-install identifier on the page surfaces: page-run uses `page_${appBlockId}` (`[[...path]].tsx:132`), a per-*app* constant identical for every viewer, and review-preview uses `page_${pubReqId}`. The genuine per-viewer case is the model slot. Network side is clean (fragments aren't sent to servers; `referrerPolicy="no-referrer"` on both iframes). Still worth resolving before allowlisting, but it is not the leak it was described as.

# Still open / not closed

- `who-am-i` (suspended, NXDOMAIN) is **undetermined**, not scored safe.
- `sensei` has a pending publish request and would become block #10.
- The dev-tunnel surface is structurally unenumerable — which is why it is refused unconditionally rather than assessed.
- The enumeration is a **point-in-time read (2026-08-05)**; two of the nine live blocks were approved within eight days, so the population moves. Any future allowlisting needs a fresh read, not this one.

# Suggested sequencing

1. Land #3634 (handshake) — carries none of this.
2. Land this PR **gated off** — it is inert, and it puts the gate, the freeze test and the corrected reasoning in the tree before anyone is tempted to enable it.
3. Publish the SDK.
4. Fix the two deferred items above.
5. Re-enumerate, then allowlist blocks one at a time as they rebuild.

https://claude.ai/code/session_01R7ZCxbgcsv3WfsSJrYdSCi

### DIFF
--- a/src/components/AppBlocks/BlockInitFragmentFreeze.browser.test.tsx
+++ b/src/components/AppBlocks/BlockInitFragmentFreeze.browser.test.tsx
@@ -64,6 +64,17 @@ function FreezeProbe({ enabled = true }: { enabled?: boolean }) {
       <button data-testid="change-base" onClick={() => setBaseSrc('https://demo.civit.ai/other')}>
         change base
       </button>
+      {/* Soft-nav appA -> appB: BOTH the src and the instance id change, with no
+          remount — the real `/apps/run` "Recently run" path. */}
+      <button
+        data-testid="soft-nav"
+        onClick={() => {
+          setBaseSrc('https://appb.civit.ai/');
+          setInstanceId('bi_second');
+        }}
+      >
+        soft nav
+      </button>
     </div>
   );
 }
@@ -116,15 +127,67 @@ describe('useBlockIframeSrc — the mount-time FREEZE', () => {
     expect(await readSrc()).toBe(before);
   });
 
-  test('🔴 a blockInstanceId change also leaves the src byte-identical', async () => {
+  test('🔴 a blockInstanceId change RE-CAPTURES — the freeze is per-INSTANCE, not per-mount', async () => {
+    // ⚠️ THIS TEST WAS INVERTED. It previously asserted that changing
+    // `blockInstanceId` left the src byte-identical — i.e. it pinned the LEAK
+    // as intended behaviour. A per-mount freeze is wrong precisely because
+    // `PageBlockHost` outlives a block instance (soft nav, no `key`), so the
+    // correct invariant is: frozen WITHIN an instance, re-captured ACROSS one.
     renderWithProviders(<FreezeProbe />);
     const before = await readSrc();
     expect(before).toContain('blockInstanceId=bi_first');
 
     await page.getByTestId('change-instance').click();
 
-    expect(await readSrc()).toBe(before);
-    expect(await readSrc()).not.toContain('bi_second');
+    const after = await readSrc();
+    expect(after).not.toBe(before);
+    expect(after).toContain('blockInstanceId=bi_second');
+    expect(after).not.toContain('bi_first');
+  });
+
+  test('🔴 THE DISCRIMINATING CASE: toggle theme, THEN move baseSrc — theme must not follow', async () => {
+    // This is the case that actually pins the REF, and it exists because the
+    // earlier mutation table overstated its row 1. That row's mutant changed
+    // BOTH `frozenFields.current` -> `fields` AND the dep array; the dep-array
+    // half is what the other tests detect. The SINGLE mutant — live `fields`
+    // with the deps left alone — passes every other test in this file, because
+    // nothing else forces a recompute AFTER a theme change.
+    //
+    // Moving `baseSrc` forces exactly that recompute. At head it re-reads the
+    // FROZEN fields; under the single mutant it re-reads the LIVE ones and the
+    // new theme leaks in.
+    renderWithProviders(<FreezeProbe />);
+    expect(await readSrc()).toContain('theme=light');
+
+    await page.getByTestId('toggle-theme').click(); // now dark, src unmoved
+    await page.getByTestId('change-base').click(); // forces a recompute
+
+    const after = await readSrc();
+    expect(after).toContain('https://demo.civit.ai/other#');
+    expect(after).toContain('theme=light'); // the FROZEN theme, not the live one
+    expect(after).not.toContain('theme=dark');
+  });
+
+  test('🔴 SOFT-NAV appA→appB re-captures — no leak of the previous app’s identity', async () => {
+    // `PageBlockHost` is rendered with NO `key`, and `_app.tsx` renders
+    // `<Component>` with none either, so a soft navigation between two apps
+    // reuses the component instance: same mount, different `blockInstanceId`.
+    // A per-MOUNT freeze leaked app A's id (and mount-time theme) into app B's
+    // document while BLOCK_INIT delivered B's — two contradicting identities.
+    // The freeze is scoped to the block INSTANCE for this reason.
+    renderWithProviders(<FreezeProbe />);
+    expect(await readSrc()).toContain('blockInstanceId=bi_first');
+
+    await page.getByTestId('toggle-theme').click(); // dark, while still on app A
+    await page.getByTestId('soft-nav').click();
+
+    const after = await readSrc();
+    expect(after).toContain('https://appb.civit.ai/#');
+    // App B gets its OWN id…
+    expect(after).toContain('blockInstanceId=bi_second');
+    expect(after).not.toContain('bi_first');
+    // …and the CURRENT theme, because it is a fresh capture for a new instance.
+    expect(after).toContain('theme=dark');
   });
 
   test('baseSrc is deliberately NOT frozen — it still re-navigates', async () => {
@@ -132,6 +195,13 @@ describe('useBlockIframeSrc — the mount-time FREEZE', () => {
     // stale manifest src, which is a behaviour CHANGE, not a safety property.
     // The audit noted that freezing it also survives the suites; this is the
     // test that stops it.
+    //
+    // 🔴 SCOPE OF THE `bi_first` ASSERTION BELOW: it holds because ONLY the src
+    // moved — the block INSTANCE is unchanged, so carrying its captured fields
+    // is correct. Do NOT read it as "the id is always carried onto a new base":
+    // when the instance changes too (a soft nav), the capture is REDONE — see
+    // the soft-nav test above. An earlier revision of this file asserted the
+    // carry without that boundary case, which read as blessing the leak.
     renderWithProviders(<FreezeProbe />);
     const before = await readSrc();
     expect(before).toContain('https://demo.civit.ai/#');
--- a/src/components/AppBlocks/BlockInitFragmentFreeze.browser.test.tsx
+++ b/src/components/AppBlocks/BlockInitFragmentFreeze.browser.test.tsx
@@ -1,0 +1,156 @@
+import { useState } from 'react';
+import { describe, expect, test, vi } from 'vitest';
+import { page } from 'vitest/browser';
+// `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
+import { renderWithProviders } from '../../../test/component-setup';
+
+// eslint-disable-next-line import/first
+import { useBlockIframeSrc } from '~/components/AppBlocks/useBlockIframeSrc';
+
+/**
+ * 🔴 THE FREEZE INVARIANT — the property the init-fragment fast path names as
+ * its safety argument, and which an adversarial audit found had ZERO coverage
+ * in either tier.
+ *
+ * The claim: `useBlockIframeSrc` captures its fragment fields at MOUNT, so a
+ * later `theme` change cannot move the rendered `src`. That matters because
+ * `theme` genuinely does change while a block is mounted (the viewer toggles
+ * dark mode), and a changing `src` ATTRIBUTE navigates a third-party iframe —
+ * at best a `hashchange` inside the block, at worst a reload that discards its
+ * in-progress state. Today the host does not propagate live theme changes to
+ * blocks at all, so the fast path must not introduce that.
+ *
+ * The audit's finding, reproduced before this file existed: replacing the
+ * frozen ref with live `fields` (and full deps) survived BOTH suites — node
+ * 31/31 and browser 38/38. So the invariant was asserted only in prose.
+ *
+ * WHY THIS TESTS THE HOOK AND NOT A HOST. The fast path ships GATED OFF (the
+ * allowlist in `blockInitFragmentGate.ts` is empty), so a host-level test would
+ * render a `src` with no fragment at all and could not observe the freeze —
+ * it would be vacuous by construction. Driving the hook with `enabled: true`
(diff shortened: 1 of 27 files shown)


ZacxDev:
## Audit response — all four findings actioned, two of them fixed rather than documented

I re-verified each finding against `origin/main` before acting on it. All four reproduced; **finding C reproduced from the repo's own comment**, which is the strongest possible corroboration.

### A — production binding arg order (fixed)

Confirmed and closed. The swap type-checks because both parameters are `ReadonlySet<string>`, and none of my three binding assertions could see it: with the lists exchanged, the *empty* allowlist becomes the denylist and the denylist becomes the allowlist, so `size === 0` and `has('playable-collections')` both still hold while a denylisted slug returns `true`. The only assertion that distinguishes the wirings is a **denylisted id through the production binding** — added, and the swap now dies to it.

This is the general shape worth naming: *an injectable form makes a predicate testable, and simultaneously makes its binding untestable.* The injection point needs its own test.

### B — mutation table overstated row 1 (corrected, not deleted)

Confirmed: my row-1 mutant changed **both** the ref and the dep array, and the dep-array half is what every other test detects — so the ref was uncovered. The single mutant passed all five freeze tests. Added the discriminating sequence (toggle theme → move `baseSrc`, the only case forcing a recompute *after* a theme change) and corrected the table in place, since the overstatement is the part worth remembering.

### C — soft-nav identity leak (🔴 **fixed**, not deferred)

The file already documents the hazard verbatim:

> `🔴 BOTH REFS ABOVE ARE PER-MOUNT, BUT THIS HOST IS NOT ALWAYS REMOUNTED. … a SOFT navigation between two apps (appA → appB, e.g. via the "Recently run" menu) reuses this component instance: same mount, different blockInstanceId.`

`launchInstanceRef` / `shouldResetLaunchMarks` exist precisely to handle it. `frozenFields` simply hadn't been given the same treatment. Since the pattern was already established two hundred lines up, documenting it as a limitation would have been the wrong call — the freeze is now scoped to the block **instance**: frozen *within* one, re-captured *across* one, so app B gets its own id and the current theme.

**Two tests had to be inverted.** One was the `:136-153` case you flagged; the other — `"a blockInstanceId change also leaves the src byte-identical"` — was asserting the leak outright and is a second instance of the same problem that the audit didn't name. Both now assert the intended behaviour. I did not mark them skipped: a test that blesses a bug is worse than no test, because the next reader takes it as the contract.

### D — `review-preview` (🔴 **refused**, not justified)

I could not construct a version of the dev-tunnel argument that stops at the tunnel. `publish-request.service.ts:4392` mints `page_${row.id}` for a **pending, un-approved** request, and on page-run `slug === blockId`, so allowlisting a published app would necessarily enable the fragment on a moderator's preview of that app's next unreviewed submission — with a mod's session. Refused. **Eligible surfaces are now `model-slot` and `page-run` only.**

### Body corrections

- **`blockInstanceId` egress is weaker than I claimed.** It is `page_${appBlockId}` on page-run — a per-*app* constant identical for every viewer — and `page_${pubReqId}` on review-preview. Only the model slot is per-viewer. Corrected; the deferred item stays, at its true severity.
- **"Adding a host is a type error"** holds for `PageBlockHost` only. `IframeHost` hard-codes `surface: 'model-slot'`, so a new model-slot host inherits it silently. My own table said so while the prose generalised.

### Date

`2026-08-05` is correct — UTC is 2026-08-05 02:29 while local is 2026-08-04 21:29 CDT. Left as-is.

### Verification

**All five audit-driven mutants die to their own named test.** A sixth — adding `theme` to the memo deps — **survived, and should**: the body reads `frozenFields.current`, so the recompute is byte-identical. That is an **equivalent mutant**, not a coverage gap, and the 7/7 pass is itself the proof.

Suites: unit **549**, component **325**, typecheck delta **zero** (163 = 163 against `origin/main` measured in the same worktree; the absolute number is a worktree-provisioning artifact — CI's clean-tree typecheck reports zero).

### Enable-prerequisites (now four)

1. Frozen `theme` wrong for "auto" viewers — the fast path would *create* the flash it exists to remove.
2. `blockInstanceId` in `location.href` — at the corrected, lower severity.
3. ~~Soft-nav identity leak~~ — **fixed**.
4. ~~`review-preview` exposure~~ — **fixed**.

Unchanged: **safe to merge, not safe to enable.** The allowlist is empty and the two remaining prerequisites gate any entry going into it.

PR #3644, opened 2026-08-05
DO NOT MERGE: deliberate CI probe for the schema drift gate

ZacxDev:
**DO NOT MERGE.** Deliberately-bad probe for the schema drift gate added in #3643. Opened only to watch that gate go RED in CI, then closed. Precedent: #3614.

Two deliberate enforced-tier drifts on `model Model`, both on columns the database already has, so both must **block**:

- `description` declared required against a NULLABLE column → `nullability`
- `@unique` on `checkpointType`, which has no unique index → `uniqueness`

Expected: `Schema drift gate` fails with `NEW, enforced (blocking) : 2`, naming both findings, while every pre-existing finding stays in the baseline and is not reported.

A gate nobody has watched fail is a claim, not a gate.


ZacxDev:
## Probe result: the gate bit. Closing.

`Schema drift gate` — **FAILURE**, exit 1, on head `462687c8a0`:

```
Schema <-> database drift gate

  baseline findings reproduced   : 61 of 61
  findings in this run           : 63
  NEW, enforced (blocking)       : 2
  NEW, pending migration (warn)  : 0
  baseline entries no longer seen: 0
  referential actions not compared: 408 (this catalog carries no ON DELETE/UPDATE data — "not measured", not "clean")

NEW drift on columns the database already has — 2
  These block. The database has the columns and does not enforce what this change declares,
  so the promise is false the moment it merges.
  nullability  Model.description (Model.description)
      declared: required
      database: NULLABLE
  uniqueness  Model.checkpointType (Model.checkpointType)
      declared: @unique(checkpointType)
      database: no total unique index on these columns

BLOCKED: this change introduces 2 drift finding(s) on columns the database already has.
```

Both planted drifts named, and **`61 of 61` reproduced** — the positive control on the zeros. A `NEW: 0` unaccompanied by a non-zero match count would be indistinguishable from a gate wired to nothing.

The negative half is on #3643 itself: same workflow, same commit content minus the schema edit, `Schema drift gate` **SUCCESS** with `NEW, enforced (blocking) : 0`. The gate fires on bad and stays silent on clean.

**Independent corroboration:** `Typecheck` also failed here, 11 errors, 8 of them naming `description` or `checkpointType` — e.g. `ModelWizard.tsx(167,13): error TS2322: Type '{ description: string | null; ... }' is not assignable`. Typecheck is SUCCESS on #3642 and #3643 off the same base, so the schema edit is the only delta. The drift the gate flagged is a real hazard, not a bookkeeping artefact.

Closing unmerged, as intended.

PR #3646, opened 2026-08-05
fix(cache): confine the admin cache-clear endpoint to its own key namespace

ZacxDev:
> **Note:** this description was corrected after review — see the comment below for what changed and why. Three claims in the original were wrong, one of them a safety claim about `target=sys`.

## The hazard

`src/pages/api/admin/clear-cache-by-pattern.ts` takes a caller-supplied glob and passes it **raw** into `clearCacheByPattern` / `clearCacheByPatterns`. Those run a `SCAN` across the whole cache keyspace and delete every key the glob matches.

Non-production environments share the cache instance with production. A cache-clear glob is written against the production key shape (`packed:caches:*`), because that is the shape the keys have had forever.

## Why namespacing made it sharper, not safer

Cache keys are now namespaced per environment (`CACHE_KEY_NAMESPACE`, see `packages/civitai-redis/src/cache-key-prefix.ts`): unset in production, set elsewhere. The prefix is applied in two places, neither of which is the clear path:

- **the key table** — `REDIS_KEYS = applyCacheKeyPrefix(REDIS_KEYS_UNPREFIXED)` (`packages/civitai-redis/src/client.ts:2289`), which is where the cache **builders** (`createCachedArray` / `createCachedObject`) get their prefix, since every builder's `key` comes from that table;
- **on-the-fly key minting** — the two `prefixCacheKey` calls in `queryCache` / `queryCacheRaw` (`src/server/utils/cache-helpers.ts`), for keys built from a free-form string rather than derived from the table.

So on a namespaced environment, `?pattern=packed:caches:*`:

- **matches production's keys** — they are the ones that live at that exact shape, and
- **matches none of the caller's own** — those live at `<namespace>:packed:caches:*`.

The one keyspace that endpoint could clear was the one it must never touch, and the environment the operator was actually aiming at was left completely untouched. Before namespacing every environment's keys were interleaved at the same shape, so the same call at least hit what it was aimed at. Namespacing turned a blunt instrument into one pointed exclusively at production.

`clearCacheByPatterns` is worse still: it scans with `MATCH: '*'` and regex-tests every key, so scoping only the caller's glob would leave it enumerating the entire keyspace — including production's — and relying purely on the regexes to spare it.

## The fix

**1. Scope the caller's patterns at the endpoint** (`scopeCachePatternToNamespace`), on **both** the streaming and non-streaming branches. The endpoint is the boundary where a hand-written, unprefixed glob enters. A pattern built from `REDIS_KEYS` already carries the prefix from the key table, which is why the scoping lives here rather than inside the helpers — doing it there would double-prefix every internal caller.

**2. Namespace containment inside both helpers.** On a namespaced environment every main-cache key carries the prefix, so a key without it belongs to another environment. Both helpers now filter enumerated keys and refuse out-of-namespace exact keys. This is what makes the confinement structural rather than a property of the call site: a caller that forgets to scope its glob clears **nothing** instead of another environment's entries.

**3. Confine `clearCacheByPatterns`' SCAN to `<namespace>:*`.** A foreign key is now never even enumerated, so no pattern — scoped, unscoped, or malformed — can reach one.

## The deliberate main-vs-sys asymmetry — and what it is NOT

The endpoint accepts `target: 'main' | 'sys'`. **`sys` is deliberately NOT namespaced, and that exemption stays.**

The reason is a fact about the **keys**: namespacing is cache-only, `REDIS_SYS_KEYS` is never run through `applyCacheKeyPrefix`, so system keys are unprefixed on every environment. Prefixing a sys pattern would make the clear match nothing at all. Gating the exemption on a namespace being set would be wrong for the same reason.

🔴 **That is not a containment property, and this PR does not give `sys` one.** An environment's system-Redis instance is **not guaranteed to be distinct from production's**. A `?target=sys` clear issued from a non-production environment can delete **production** data, and nothing in this change stands in its way.

It is in fact the more dangerous of the two targets: cache entries are regenerable from the database, system values are not — a deleted one is simply gone. Treat any `?target=sys` call as a production operation regardless of where it is run from. Confining that target is out of scope here and wants its own change.

## Production is unchanged

With no namespace configured every prefix is `''`, so `scopeCachePatternToNamespace` returns its argument, `isInCacheNamespace` is unconditionally true, and the SCAN `MATCH` is `'*'`. The pattern sent to Redis, the `MATCH`, and the set of deleted keys are all identical to before.

That is asserted rather than reasoned: `cache-helpers-clear-namespace-prod.test.ts` pins the literal `MATCH` strings (`'packed:caches:*'`, `'*'`) and the exact deleted-key sets, and mutation G7 below confirms those assertions actually catch a prefix leaking into production.

The SSE/streaming event sequence and the response shape are unchanged, and are covered in **both** the production and the namespaced suite. (In the original version of this PR they were covered in the production suite only — where scoping is the identity function, so those tests proved nothing about scoping. See G8.)

## Verification

New tests, mirroring the existing `cache-helpers-key-prefix-{prod,preview}` split (a `vi.mock` factory is evaluated once per file, so `CACHE_KEY_NAMESPACE` cannot be flipped within one file):

- `src/server/utils/__tests__/cache-helpers-clear-namespace-prod.test.ts` — 9 tests
- `src/server/utils/__tests__/cache-helpers-clear-namespace-preview.test.ts` — 15 tests

Both drive the real endpoint handler through the real helpers into a fake Redis whose `scanIterator` honours `MATCH` server-side, so a `MATCH` confinement is tested as the real thing rather than as a recorded argument.

```
# new + existing cache-helpers suites
 Test Files  5 passed (5)
      Tests  43 passed (43)

# full unit suite
 Test Files  809 passed (809)
      Tests  11965 passed | 1 skipped (11966)
```

`node scripts/typecheck.mjs` → `typecheck: OK — 0 type errors`. ESLint on the four changed files: 0 errors (2 pre-existing `no-explicit-any` warnings on lines this PR does not touch). Prettier clean.

### Does the containment filter break any existing caller?

No — checked by enumerating the defining surface rather than sampling:

- **Direct `clearCacheByPattern` callers**: 5 in `creator-program.service.ts`, all `` `${REDIS_KEYS.CREATOR_PROGRAM.X}:*` ``, plus the endpoint (now scoped) — and `clearCacheByPatterns` has no caller but the endpoint.
- **`clearByPattern` via the cache builders** (`cached-array.ts` `flush()` → `` `${key}:*` ``): every `createCachedArray` / `createCachedObject` call site in the repo — 25 in `redis/caches.ts` plus 9 elsewhere — takes its `key` from `REDIS_KEYS`.

`REDIS_KEYS` is `applyCacheKeyPrefix(REDIS_KEYS_UNPREFIXED)`, so every one of those patterns already carries the namespace and passes containment unchanged. That is also why the caller-glob scoping lives at the endpoint rather than inside the helpers: applying it in the helpers would double-prefix all of the above.

### Mutation checks

Every guard was broken individually, the suite run, and the failure confirmed to be that guard's own assertion. The harness prints the applied diff before each run, so a silently-unapplied patch cannot masquerade as a survivor, and it carries a **comment-only negative control** that must survive — otherwise "no survivors" would be a fact about the harness rather than about the guards.

| # | Guard broken | Failure |
|---|---|---|
| C0 | *(negative control — comment only)* | **survived, as required** |
| G1 | endpoint drops the scoping call | `expected [ 'packed:caches:*' ] to deeply equal [ 'preview:packed:caches:*' ]` (+3 more, incl. both SSE cases) |
| G2 | sys exemption removed | `expected 'preview:device:accounts:*' to be 'device:accounts:*'` |
| G3 | `clearCacheByPattern` exact-key containment removed | `expected [ 'packed:caches:user-cosmetics:1' ] to deeply equal []` |
| G4 | `clearCacheByPattern` scan filter removed | `expected [ …(3) ] to deeply equal []` — the 3 being production-shaped keys |
| G5 | `clearCacheByPatterns` exact-key containment removed | `"cleared": 0` → `"cleared": 1` on two production-shaped keys |
| G6 | `clearCacheByPatterns` SCAN confinement removed | `expected [ '*' ] to deeply equal [ 'preview:*' ]` |
| G7 | production identity broken (non-empty prefix when unset) | 8 of 9 production tests; `expected 'zz:packed:caches:*' to be 'packed:caches:*'` |
| G8 | **SSE branch only** passes unscoped patterns | `expected { total: +0, perPattern: […] } to deeply equal { total: 2, … }` — received the raw patterns with `cleared: 0` |

G4 is the headline: with that one filter removed, an unscoped glob deletes three production-shaped keys, which is exactly the defect this PR closes.

G8 is worth reading closely. Its first form failed on the *event sequence* (`expected [ 'start', 'done' ] to deeply equal [ 'start', 'progress', 'done' ]`) rather than on scoping — an unscoped clear matches nothing, and `clearCacheByPatterns` skips its progress callback on an empty batch, so the `progress` event disappears. That is a true failure that says nothing about the thing under test. The tests now look events up by name and assert scoping first, so the mutant dies on the scoping assertion; the ordering constraint is commented in the test so it is not silently undone.

Each mutant is killed by a test written to exercise that guard on a path where it is the only thing in the way — the endpoint cases pin the scoping, the direct-helper cases pass a deliberately unscoped pattern to pin helper-side containment — so no mutant died on a neighbouring guard's behalf.

### Not verified

- No live Redis and no deployed environment: all verification is unit-level against a fake client. The `MATCH` confinement rests on `SCAN MATCH` being evaluated server-side.
- `WebhookEndpoint` is stubbed to a pass-through in the tests; token auth is not exercised.

### DIFF
--- a/src/pages/api/admin/clear-cache-by-pattern.ts
+++ b/src/pages/api/admin/clear-cache-by-pattern.ts
@@ -1,6 +1,10 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import * as z from 'zod';
-import { clearCacheByPattern, clearCacheByPatterns } from '~/server/utils/cache-helpers';
+import {
+  clearCacheByPattern,
+  clearCacheByPatterns,
+  scopeCachePatternToNamespace,
+} from '~/server/utils/cache-helpers';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
 
 const schema = z.object({
@@ -15,7 +19,7 @@ const schema = z.object({
 export default WebhookEndpoint(async function (req: NextApiRequest, res: NextApiResponse) {
   const parsed = schema.parse(req.query);
   const target = parsed.target ?? 'main';
-  const patterns = parsed.patterns
+  const rawPatterns = parsed.patterns
     ? parsed.patterns
         .split(',')
         .map((p) => p.trim())
@@ -23,6 +27,17 @@ export default WebhookEndpoint(async function (req: NextApiRequest, res: NextApi
     : parsed.pattern
     ? [parsed.pattern]
     : [];
+
+  // 🔴 A cache-clear glob is written against the PRODUCTION key shape (`packed:caches:*`), and
+  // every deployment shares one cache instance. On a deployment whose cache keys are namespaced
+  // that glob matches production's keys and none of this deployment's own, so an unscoped clear
+  // from here would purge production while leaving the environment it was aimed at untouched.
+  // Scoping it to this deployment's namespace makes the endpoint able to address ONLY its own
+  // keyspace. Identity in production (no namespace ⇒ no prefix ⇒ the exact patterns as typed).
+  //
+  // `target: 'sys'` is deliberately NOT scoped — that keyspace is a separate per-deployment
+  // instance whose keys are unprefixed by construction; see `scopeCachePatternToNamespace`.
+  const patterns = rawPatterns.map((p) => scopeCachePatternToNamespace(p, target));
 
   if (patterns.length === 0) {
     res.status(400).json({ error: 'Provide either ?pattern= or ?patterns=a,b,c' });
--- a/src/pages/api/admin/clear-cache-by-pattern.ts
+++ b/src/pages/api/admin/clear-cache-by-pattern.ts
@@ -35,8 +35,14 @@ export default WebhookEndpoint(async function (req: NextApiRequest, res: NextApi
   // Scoping it to this deployment's namespace makes the endpoint able to address ONLY its own
   // keyspace. Identity in production (no namespace ⇒ no prefix ⇒ the exact patterns as typed).
   //
-  // `target: 'sys'` is deliberately NOT scoped — that keyspace is a separate per-deployment
-  // instance whose keys are unprefixed by construction; see `scopeCachePatternToNamespace`.
+  // 🔴 THIS CONFINEMENT COVERS `target: 'main'` ONLY. `target: 'sys'` is deliberately NOT scoped,
+  // because system keys are never namespaced on any deployment — prefixing one would match
+  // nothing. That is a fact about the KEYS, not a safety property of the keyspace: a deployment's
+  // system-Redis instance is not guaranteed to be distinct from production's, so a
+  // `?target=sys` clear issued from a non-production deployment CAN delete production data, and
+  // this endpoint does not stop it. Unlike cache entries, system values are not regenerable from
+  // the database. Treat any `?target=sys` call as a production operation.
+  // See `scopeCachePatternToNamespace` / `cacheNamespacePrefix`.
   const patterns = rawPatterns.map((p) => scopeCachePatternToNamespace(p, target));
 
   if (patterns.length === 0) {
--- a/src/server/utils/__tests__/cache-helpers-clear-namespace-preview.test.ts
+++ b/src/server/utils/__tests__/cache-helpers-clear-namespace-preview.test.ts
@@ -1,0 +1,324 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type * as RedisClientModule from '@civitai/redis/client';
+
+/**
+ * NAMESPACED half of the coverage for namespace-scoping the admin cache-clear path (the
+ * production half is cache-helpers-clear-namespace-prod.test.ts — the two cannot share a file
+ * because a `vi.mock` factory is evaluated once per file, so `CACHE_KEY_NAMESPACE` cannot be
+ * flipped between cases; same split as the cache-helpers-key-prefix-*.test.ts pair).
+ *
+ * 🔴 The hazard being pinned: a cache-clear glob is written by hand against the PRODUCTION key
+ * shape (`packed:caches:*`), and every deployment shares ONE cache instance. Unscoped, such a glob
+ * run from a namespaced deployment matches production's keys and NONE of its own — the single
+ * environment it can clear is the one it must never touch. Every case below therefore seeds a
+ * production-shaped key alongside the namespaced one and asserts the production key SURVIVES.
+ *
+ * Each guard is exercised through the path where it is the ONLY thing standing between the input
+ * and a foreign `del`, so a test cannot go green on a neighbouring guard's behalf: the endpoint
+ * cases pin the scoping of a caller glob, the direct-helper cases pass a DELIBERATELY UNSCOPED
+ * pattern to pin the helper-side containment.
+ */
+
+vi.hoisted(() => {
+  process.env.CACHE_KEY_NAMESPACE = 'preview';
+  delete process.env.IS_PREVIEW;
+});
+
+/**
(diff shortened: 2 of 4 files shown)


ZacxDev:
### Description corrected after review — what changed and why

Flagging this rather than editing silently, because anyone who read the earlier version came away with a wrong idea about `target=sys`. Two code/test changes landed in `ed00132`; three description claims were corrected.

**1. 🔴 The `target=sys` justification was factually wrong — and wrong in the dangerous direction.**

The original description, and the doc blocks at three places in the code, claimed the system keyspace is a separate per-environment instance whose keys therefore "arrive unprefixed by construction" — and closed with an instruction not to change it. That premise does not hold: an environment's system-Redis instance is **not guaranteed to be distinct from production's**. So a `?target=sys` clear from a non-production environment can reach production data.

That made the comment worse than merely inaccurate: it described the *more* destructive of the two keyspaces as the safe one, and told the next maintainer not to look. Cache entries are regenerable from the database; system values are not.

**The code did not change and should not.** The exemption is correct — system keys are unprefixed everywhere, so prefixing a sys pattern would match nothing, and gating the exemption on a namespace being set would break environments that *do* own their instance. What changed is the *rationale*: it is now stated as a fact about the keys, with an explicit warning that it is not containment, at all three sites plus the test name (previously "leaves the sys target completely unaffected by the namespace", which read as a safety assertion; now "does not apply the namespace to a sys pattern (mechanism, not containment)").

Confining `target=sys` is a real gap. It is out of scope here and wants its own change.

**2. "SSE/streaming behaviour and the response shape are unchanged (both covered)" was overstated.**

The streaming branch reads the scoped patterns at its own call site. It was covered only in the production suite — where scoping is the identity function, so those tests could not fail on a scoping regression. A mutant passing unscoped patterns on the `stream=1` branch alone left all 22 tests green.

Two SSE cases added to the namespaced suite (now 15 tests there, 24 across the pair). Mutation-verified: the mutant now dies on the scoping assertion, reporting the raw patterns with `cleared: 0` against the expected scoped patterns with non-zero counts.

Worth recording because it nearly slipped through a second time: the mutant's *first* kill was on the event-sequence assertion (`expected [ 'start', 'done' ] to deeply equal [ 'start', 'progress', 'done' ]`), not on scoping — an unscoped clear matches nothing, and `clearCacheByPatterns` skips its progress callback on an empty batch, so the `progress` event vanishes. A red test, for the wrong reason. The tests now look events up by name and assert scoping first; the ordering constraint is commented in the test so it does not get quietly undone.

**3. The line references for where the prefix is applied pointed at the wrong mechanism.**

They cited the `prefixCacheKey` calls in `cache-helpers.ts` as the cache-builder path. Those are on-the-fly key *minting* (`queryCache` / `queryCacheRaw`). The builders get their prefix from the key table itself — `REDIS_KEYS = applyCacheKeyPrefix(REDIS_KEYS_UNPREFIXED)`. Mechanism was right, labels were wrong; corrected in the body.

---

Full mutant battery re-run after the fix rather than just the new mutant, since shared files were edited: 8 mutants, all killed, plus a comment-only negative control that survived as required — without that control, "no survivors" would have been a claim about the harness rather than about the guards. Full unit suite 809 files / 11965 tests passing; typecheck 0 errors.

PR #3649, opened 2026-08-05
test(apps): remove the wall-clock race from the app-search debounce guard

ZacxDev:
## The flake

`src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx` →
`🔴 the search box does NOT write the URL per keystroke — only the debounced value`

Observed intermittently on a loaded CI runner (1 failed / 1258 passed, on a PR
touching none of this code):

```
AssertionError: expected [ { query: 'mat' } ] to have a length of +0 but got 1
  at src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx:221:58
```

## Root cause — a wall-clock race

The component debounces the search box before echoing it to the URL:

```ts
const [debouncedSearch] = useDebouncedValue(searchInput, 300);
```

The test did two awaited `search.fill()` calls and then asserted that **zero**
URL writes had happened yet. That assertion is only true if both fills complete
inside the real 300 ms window — so its truth depended on how fast the machine
happened to be, not on the component. Under load a single `fill()` can exceed
300 ms, the debounce fires *between* the two fills, and a write carrying the
intermediate value (`'mat'`) lands.

## Why the previous mitigation was not enough

The test had already been mitigated once: six fills cut to two, on the stated
reasoning that "two fills is well inside the window on any machine". That
reduced the probability but left the wall-clock dependency completely intact —
and the assumption was duly falsified. Re-tuning the fill count would just
reset the same clock.

## The fix — control the clock instead of racing it

```ts
vi.useFakeTimers({ toFake: ['setTimeout', 'clearTimeout'] });
await search.fill('mat');
await search.fill('matrix');
expect(writes).toHaveLength(0);          // now timing-independent
await act(async () => vi.advanceTimersByTime(300));
expect(writes).toEqual([{ query: 'matrix' }]);
```

With `setTimeout` faked, **no amount of real elapsed time can fire the
debounce** — `advanceTimersByTime(300)` is the only thing that can, so the
zero-writes assertion becomes a statement about the component rather than about
the machine. `toFake` is deliberately scoped to the two timer primitives:
Playwright's real `fill()` interaction and the browser-runner channel still need
real `queueMicrotask`/`requestAnimationFrame`/`Date`/`performance`. An
unconditional `afterEach(() => vi.useRealTimers())` keeps a throw from leaking
the fake clock into the next test.

The trailing assertion also stops being a `vi.waitFor` retry loop — a retried
"expect exactly 1 write" would happily succeed in the window *before* a second,
buggy write landed. It is now a single deterministic check after the flush.

The stale comment block describing the six-fills/two-fills reasoning has been
replaced with one describing what the test actually does now.

### Incidental: a real typing gap

`@types/react` (18.0.x) does not declare `act`; the repo fills that gap with a
global `declare module 'react'` augmentation in
`src/tests/components/TrackView/useTrackEvent.wiring.test.ts` that declares only
the **sync** form, `(callback: () => void) => void`. Against that declaration
`await act(async …)` is `await void`, and TypeScript reports *"'await' has no
effect on the type of this expression"*. Rather than widen an augmentation
another test owns, this aliases the async overload React actually implements, in
this file only.

## Verification

| Check | Result |
|---|---|
| Target file | **12/12 passed** |
| Determinism | **10/10** consecutive runs, all 12/12 |
| Under load | **5/5** at 2× CPU oversubscription — the test took 891 ms (well past 300 ms) and still passed; this is the condition that used to break it |
| Typecheck | no errors in this file (only pre-existing unrelated ones in my env) |
| ESLint / Prettier | clean for this file |
| React `act(...)` warnings | 0 |

**Positive control** — removing `advanceTimersByTime(300)` leaves the write
count at **zero** (`expected [] to have a length of 1 but got +0`), proving the
fake clock is genuinely installed and that real elapsed time can no longer fire
the debounce. Without this control a green run could not distinguish "fake
timers working" from "fills happened to be fast".

**Mutation check** — the guard still has teeth. Rewiring the component's URL-echo
effect to use `searchInput` (per keystroke) instead of `debouncedSearch`:

```
AssertionError: expected [ { query: 'mat' }, …(1) ] to have a length of +0 but got 2
 ❯ src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx:273:58
   Tests  1 failed | 11 passed (12)
```

It fails on **this guard's own** zero-writes assertion, naming the unexpected
writes, and nothing else in the file fails. Mutation fully reverted — the diff
in this PR touches the test file only.

### DIFF
--- a/src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx
@@ -1,5 +1,6 @@
-import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
+import { act as reactAct } from 'react';
 import { useRouter } from 'next/router';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
 import { renderWithProviders } from '../../../test/component-setup';
@@ -48,6 +49,29 @@ function generatedCssFor(el: HTMLElement): string {
     .join('\n')
     .replace(/\s+/g, '');
 }
+
+/**
+ * React 18.3 exposes `act` on the `react` export, but the pinned
+ * `@types/react` (18.0.x) does not declare it. The repo fills that gap with a
+ * `declare module 'react'` augmentation in
+ * `src/tests/components/TrackView/useTrackEvent.wiring.test.ts` — but it
+ * declares only the SYNC form, `(callback: () => void) => void`.
+ *
+ * That augmentation is global, so it is what an `import { act } from 'react'`
+ * resolves to here, and it under-describes the async form this file needs:
+ * with an async callback `act` really returns a thenable, and awaiting it is
+ * the whole point (it is what flushes the setState + effect the advanced timer
+ * schedules). Against the sync-only declaration TypeScript sees `await void`
+ * and reports "'await' has no effect on the type of this expression" — a real
+ * hint about a real typing gap, not a phantom.
+ *
+ * Rather than widen a module augmentation that another test owns, narrow the
+ * repair to this file: alias the async overload that React actually implements.
+ * Verified behaviourally, not just typed — removing the `advanceTimersByTime`
+ * call leaves ZERO writes recorded, which is only observable if the await here
+ * genuinely flushes.
+ */
+const act = reactAct as unknown as <T>(callback: () => T | Promise<T>) => Promise<T>;
 
 const mocks = vi.hoisted(() => ({
   items: [] as ListingCard[],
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by #3654, which landed the same fix while this was in review.

#3654 is the better implementation: it drives the debounce with `advanceTimersByTimeAsync`, which flushes microtasks between timers natively. This PR used the synchronous `advanceTimersByTime` and therefore needed `act` to flush — which in turn needed a local type workaround. #3654 doesn't call `act` at all, so that whole layer disappears.

Closing rather than rebasing: the only thing unique here was the `act` workaround, and it's moot against an implementation that doesn't use `act`.

One finding from this work is worth keeping and is **not** fixed by #3654: the repo augments `react` with a **synchronous-only** `act` signature (in `src/tests/components/TrackView/useTrackEvent.wiring.test.ts`), because `@types/react` 18.0.x ships none. Any test doing `await act(async () => …)` is therefore silently `await void`. #3654 sidesteps it; the augmentation is still narrow. Worth its own issue.

PR #3655, opened 2026-08-05
feat(app-blocks): accept an inline ComfyUI graph on the customComfy bridge

ZacxDev:
## What changed

An App Block can now submit a `customComfy` body carrying the **ComfyUI workflow graph inline**, instead of naming one of the two pre-registered recipes. The recipe path is untouched.

Wire shape — a **second arm** on the existing `kind:'customComfy'` member, discriminated by `mode`:

```jsonc
{ "kind": "customComfy", "mode": "inline",
  "workflow": { "1": { "class_type": "CLIPTextEncode", "inputs": { "text": "…" } } },
  "resources": ["urn:air:sdxl:lora:civitai:118025@136251"],
  "prompt": "…", "negativePrompt": "…",
  "maxBuzz": 60 }
```

A body with **no `mode` key is a recipe body** and parses byte-identically to before — every deployed app and every published `@civitai/app-sdk` body is unaffected.

`customComfy` already had a working estimate → submit → post-paid billing → output-extraction path. This widens what it accepts; it does not build a pipeline. **The entire spend belt — static gate, per-user daily reservation, per-app aggregate reservation, dev-session backstop, settle-to-actual — is reused untouched**, because it only ever consumed a ceiling number.

### The trust-model change, stated plainly

A recipe is a code-reviewed in-repo artifact, and that review was the trust root for three things at once. An inline graph has no review, so each is replaced by a mechanical gate. The model moves from *"a human reviewed this graph"* to *"the belt gates the resources, the timeout caps the spend, the audit gates the text, and only developers can reach it."*

`assertViewerIsAppDeveloper` + page-only + the mod-only pre-GA flag are **unchanged and load-bearing** — they are what bounds the blast radius of everything below. Nothing here relaxes them.

---

## Gate 1 — entitlement over app-supplied `resources`

`services/blocks/recipes/index.ts` documented that an app-supplied `resources` array bypasses the early-access / Private belt, and that this was safe *only because recipes are code-reviewed*. Inline graphs remove exactly that protection. There was already a `TODO(app-blocks customComfy pre-GA)` naming the fix site.

New `services/blocks/inline-comfy.service.ts` gates the array:

- **`getResourceData`** (exported, context-free) folds `applyPaidAccessGating` internally, and `paid-access-gating.ts` does `canGenerate = hasAccess && canGenerate` — so **early-access comes with it**.
- The **Private/epoch subscription throw** does *not* live in `getResourceData`; it is ported from `validateAndEnrichResources` (`orchestration-new.service`, not exported).
- The onsite belt `validateAndEnrichResources` itself is **not** reusable — not exported, and it reads a `GenerationGraphOutput` an app path never produces.

### The three silent fail-opens, each with its own test

A checker that iterates the **returned** array and asserts `canGenerate` passes **vacuously** on the first two:

| # | Mechanism | Behaviour here |
|---|---|---|
| 1 | `getResourceData` **drops** unresolvable ids (`if (!cached) return null` → `.filter(isDefined)`) — "no bad resources" is trivially true of an empty array | asserted **per requested id**; a dropped id is a REJECT |
| 2 | `getResourceData` **substitutes** — `getResourceDataSubstitutes` attaches a covered sibling as `resource.substitute`. Onsite that is graceful degradation; here the graph names one specific AIR in `lora_name` and nothing rewrites it | a present `substitute` is a REJECT |
| 3 | The gate the recipe arm uses (`resolveCanGenerateForVersions`) is the **weak** one — no `hasAccess`, no Private | inline uses `getResourceData`, not that gate |

Note on #2: the returned row **keeps the requested id** (the substitute is a sibling *field*, not a replacement element), so an anti-drop check alone does not catch it. The test fixture leaves `canGenerate: true` precisely so it can only pass via the substitute check.

### AIR shape — three measured parser properties that each fail open

The gate uses an explicit **strict** reader rather than `Air.parseSafe`. Measured by running `@civitai/client` 0.2.0-beta.84:

1. **The `urn:air:` prefix is optional, and so are `ecosystem`/`type`.** `'sdxl:lora:civitai:118025@136251'` parses; `'foo:bar'` parses as `{source:'foo', id:'bar'}`. "parseSafe returned a value" is not evidence the string is an AIR.
2. **`source` is case-preserving.** `urn:air:sdxl:lora:CIVITAI:118025@136251` yields `source: 'CIVITAI'` — a case-sensitive `=== 'civitai'` test routes it down the non-civitai branch and **skips entitlement entirely**.
3. **`version` is an optional raw string.** `Number(undefined)` and `Number('def')` are both `NaN`; a `NaN` id resolves nothing and, via the silent drop above, reads as "no gated resources here".

A second, stricter copy of the grammar is safe **because of its direction**: divergence can only cause a false *reject*, never a false accept. (The containment check deliberately does not parse at all — there divergence would be a bypass, so it compares whole strings.) A test pins the reader against the AIRs the shipped recipes actually use, so over-strictness goes red.

`resources` also passes an **allowlist** of AIR `type` and `source` (weights only), which rejects `oci:image` arriving through `resources` — the same capability as the blocked `comfyImage` field.

---

## Gate 2 — prompt moderation

`auditPromptServer` reads `params.prompt`, guaranteed on the recipe arm by the recipe's `.strict()` schema. An inline graph carries its prompts as leaf strings inside `CLIPTextEncode` nodes, so **moderation would silently become a no-op with every gate green**.

The graph is swept for string leaves — keys **and** values, mirroring `containsAirReference`, because `additionalNetworks` is documented "Use the AIR of the network as the key" — and the joined text is audited alongside the declared prompt. Same traversal as the containment check.

Two deliberate choices:
- **Over-cap rejects, it does not truncate.** Truncating is a moderation bypass by construction: put the disallowed text past the cutoff.
- **Depth cap throws**, fail-closed, rather than recursing to a `RangeError` on attacker-shaped JSON.

---

## Never-app-settable fields

`sessionOwnerApiToken` (a Civitai API token forwarded to the claiming worker), `comfyImage` (an arbitrary OCI container), `minVramGb` (changes the worker tier and therefore the Buzz/second rate the whole ceiling argument rests on), `sessionId`, `useSageAttention`, `minimumDurationSeconds`.

Two belts: the step input is **constructed server-side from an allowlist** (`{resources, trace, workflow}` — the body is never spread), and the wire arm is `.strict()`. Tested from both directions, including an assertion on the exact emitted key set.

---

## Display estimate

A real `whatIf` returns 0 for `customComfy`, so today the recipe's hand-written per-engine number is what the user sees. An inline graph has no honest per-engine number and none can be derived — the orchestrator forwards the graph opaquely and cannot price it.

The declared `maxBuzz` ceiling **is** the honest answer, because it is physically enforced: the server stamps `stepTimeoutSeconds = maxBuzz` as the step timeout, the job is cancelled there, and settle-to-actual refunds the difference. The user sees a true upper bound and is charged the real cost.

⚠️ **Surface this as "up to N Buzz", not as a price.** `cost.total` is the same field the txt2img path fills with a real estimate, so a UI rendering it as a definite price will overstate an inline job. That is a copy change in `@civitai/app-sdk` / the block — **tracked separately, not done here**, and called out in the code.

---

## Where `INLINE_MAX_BUZZ = 250` came from

Derived, not invented:

| Reference | Value | Where |
|---|---|---|
| `DEV_BUZZ_BUDGET_CAP` | **250** | `services/blocks/dev-scoped-mint.service.ts` — per-call budget cap on the developer path |
| `BUZZ_BUDGET_CAP` | 1000 | `pages/api/v1/block-tokens/index.ts` — platform-wide per-call cap |
| `BUZZ_BUDGET_DEFAULT` | 10 | same file |
| Largest shipped recipe ceiling | 180 | `seamless-pano-360`, `qwen-image` engine |

This surface is developer-only, and the developer path's own per-call ceiling is already 250 — anything above it is unreachable through a dev token anyway, since the static gate fail-snapshots on `ceiling > claims.buzzBudget`. It also sits comfortably above the largest shipped recipe ceiling, so an inline graph is not more constrained than a recipe.

It is spelled as a literal (the wire layer must stay import-light) with **a test pinning `INLINE_MAX_BUZZ === DEV_BUZZ_BUDGET_CAP`**, so the derivation cannot drift silently.

---

## 🔴 What I deliberately did NOT do: nodepacks

`INLINE_NODEPACKS_ENABLED = false`. A `comfy:nodepack` URN in `resources` installs third-party custom nodes into the ComfyUI container at runtime — **arbitrary code execution**, sandboxed only by the worker container. That is a product/security decision with a named owner and it has not been made, so this ships **closed** rather than shipping the decision by default.

**Consequence, stated plainly: inline graphs are limited to stock ComfyUI nodes plus non-nodepack AIRs. An app needing a custom node pack is still blocked.**

Enabling it is a one-line flip of the constant — the source/type allowlists are keyed off it, and tests for the rejecting state already exist. A follow-up PR flips it.

Also deliberately not done: nodepack **snapshot pinning** (`comfyNodepackSnapshot` is not in the step registry); migrating `customComfy` onto the `kind:'step'` registry (explicitly out of scope per `workflow.schema.ts`); retiring the recipe path; relaxing developer-only access; pointing the **recipe** branch at the new belt (that would change the behaviour of two shipped recipes in a PR whose subject is a new arm — the `TODO` is updated to say so rather than silently closed).

---

## 🔴 Two zod findings worth reading before reviewing the union

Both **measured by running zod 4.0.17**, not reasoned about. The obvious spellings fail, and one fails in production only:

1. Adding the inline arm as a **fourth option** of the outer discriminated union throws `Duplicate discriminator value "customComfy"` — zod's discriminated union is a value→schema map.
2. Nesting on `mode` with the recipe arm declaring **`mode: z.literal('recipe').default('recipe')`** constructs fine, type-checks, builds — and then **rejects every deployed recipe body** with `No matching discriminator`. zod reads the discriminator off the schema and does not see it through a `.default()`. This is the dangerous one.
3. Wrapping everything in a flat `z.union([discriminatedUnion, inlineArm])` parses every body correctly **but degrades the error shape**: a rejected body reports one top-level `invalid_union` issue with `path: []` instead of `path: ['step']` / `path: ['recipe']`. That path is the only diagnostic an app author gets for a bounced submit, and it **broke an existing assertion in `workflow.schema.step.test.ts`** — which is how I caught it.

The landed construction uses **`mode: z.literal('recipe').optional()`** on the recipe arm inside a nested discriminated union. `.optional()` works where `.default()` does not: zod treats `undefined` as a legitimate discriminator value, so a body with no `mode` lands on the recipe arm and the outer union stays a real discriminated union with per-field error paths intact.

Failure modes 2 and 3 both have mutations in the sweep below.

---

### Typecheck
`pnpm run typecheck` → **`typecheck: OK — 0 type errors in 200s (heap cap 8192 MB)`** — the wrapper's own classification line, not an exit code.

### Test counts

| Suite | Base (`origin/main`) | HEAD |
|---|---|---|
| `blocks.router.workflow.test.ts` | 314 passed | **327 passed** (+13) |
| `inline-comfy.service.test.ts` | — (new) | **31 passed** |
| `workflow.schema.inline-comfy.test.ts` | — (new) | **25 passed** |
| `services/blocks` + `schema/blocks` + router workflow | — | **122 files, 3092 passed, 0 failed, 0 skipped** |

### Lint — before/after comparison, not just "clean"
ESLint on the modified router test reports **1 error + 15 warnings**. The *identical* file at `origin/main` reports **the same 1 error + 15 warnings** (a pre-existing `consistent-type-imports` error on an untouched line). **Zero new lint problems.** Prettier: clean on all 7 changed files — and it reported issues on 3 of them before I ran `--write`, so the pass is not a matched-nothing pass.

### 🔴 Red→green matrix — by mutation, and here is why

Most tests here cover a **feature that did not previously exist**, so "red at base" is not available for them: at `origin/main` the test files do not import, which reports *zero tests collected* rather than a failure — a green-looking non-signal, not evidence. Rather than report a matrix that cannot mean anything, every gate was **mutation-tested at HEAD**: break the gate on purpose, confirm the named test goes red, and read *which* assertion failed.

The sweep has its own controls, run before any verdict was read:
- **Positive control** — every `-t` filter must select >0 tests on the clean tree. This caught a real fault: one filter matched **zero** tests, which would have reported "0 failed" and been indistinguishable from a surviving mutant. The sweep aborted, the filter was fixed, and it was re-run.
- **Negative control** — all selected tests must pass on the clean tree.
- Each mutation's application is verified by occurrence count before running (a non-applying patch is reported, never silently scored as "killed").

**Result: 19/19 mutations killed.**

| # | Mutation | Test that went red |
|---|---|---|
| M1 | `INLINE_NODEPACKS_ENABLED` → `true` | nodepack rejection (×2) |
| M2 | accept a silently-dropped id | ANTI-DROP (×2) |
| M3 | accept a substituted version | ANTI-SUBSTITUTE |
| M4 | drop the `urn:air:` prefix requirement | prefix-less pseudo-AIR |
| M5 | compare AIR source case-sensitively | uppercase `CIVITAI` |
| M6 | skip numeric-version validation | missing/non-numeric version |
| M7 | stop walking object KEYS | AIR-in-a-key (×2) |
| M8 | audit only the declared prompt | `CLIPTextEncode` sweep |
| M9 | drop the Private-subscription requirement | Private (×2) |
| M10 | remove the recursion budget | depth cap |
| M11 | drop `.strict()` from the inline arm | never-settable fields (×7) |
| M12 | raise `INLINE_MAX_BUZZ` past its derivation | derivation pin |
| M13 | `.optional()` → `.default()` on the discriminator | **back-compat (×2)** |
| M13b | flat `z.union` wrapper | precise error path |
| M14 | unwire the containment gate | router: undeclared AIR |
| M15 | unwire the entitlement belt | router: non-generatable → FORBIDDEN |
| M16 | audit the declared prompt, not the graph | router: SWEPT GRAPH |
| M17 | ignore the declared ceiling | router: reserves declared `maxBuzz` |
| M18 | spread the app body into the step input | router: exact key set |

**M13 is the one that matters most for reviewers**: it proves the back-compat tests are not ceremony — the plausible-looking construction really does break every deployed recipe body, and a test catches it.

### 🔴 Honest caveat on the sweep: 3 of 19 died to a *sibling* guard
M4, M5 and M6 each turned the target test red, but with a **different guard's** error message than the one removed:
- M4 → caught by the strict regex (which itself anchors `^urn:air:`), so the `startsWith` check is genuinely redundant; it is kept for the clearer message.
- M5 → caught by the source **allowlist** (lowercase-only, so `CIVITAI` fails it).
- M6 → caught by the **anti-drop** check (`NaN` resolves nothing → dropped).

All three are **fail-closed** redundancies, i.e. the safe direction — but "each mutation died" is a weaker claim than "each died for its own reason", and only the first is true for these three.

### Seam coverage, not just isolation
The router tests leave `inline-comfy.service` **real** and stub only its data source (`getResourceData` / `getHighestTierSubscription`). Mocking the belt itself would reduce them to "the router called a function I replaced" — which is exactly the seam a per-component test cannot see. M14/M15/M16 are the proof that these fail when the belt is correct but not wired.

### 🔴 One branch is NOT reachable at the router seam, and it is recorded rather than forced
The **Private-subscription** branch cannot be exercised through the router today: App Blocks is moderator-only pre-GA, so every viewer who can reach a submit is a moderator, and the belt deliberately exempts moderators (matching the onsite belt it is ported from). A router-level test would be green for the *wrong reason* — the mod exemption, not the gate. It is exercised over both a subscriber and a non-subscriber in `inline-comfy.service.test.ts`, and the router test asserts the **exemption** (the live path) instead, with a comment saying to add the real case when App Blocks opens past mod-only.

---

## What I could NOT verify

- **Nothing in production.** This repo deploys from `release`, not `main`. Deployed ≠ verified — no inline graph has been driven end-to-end on a live cluster. The check that actually closes the original complaint is a developer deploying one of their real blocked apps and generating.
- **The physical ceiling is argued, not observed.** That the step `timeout` cancels the job at `maxBuzz` GPU-seconds is inherited from the recipe path's existing design; I did not submit a deliberately long inline job and read the real Buzz debit.
- **≈1 Buzz/GPU-second is unverified in this codebase.** `CustomComfyUsage.buzzPerSecond` exists in the generated client and is referenced **nowhere** in `src` (measured, zero hits). If any worker tier bills above 1 Buzz/sec the ceiling under-counts. This is a **pre-existing** property of the recipe path, inherited unchanged — and it is exactly why `minVramGb` must stay server-controlled. Confirm the real rate before raising `INLINE_MAX_BUZZ`.
- **Orchestrator behaviour on an undeclared graph AIR** is from the generated client's doc comment, not observed. Gate 1's containment check makes that claim non-load-bearing.
- **Whether the block UI renders `cost.total` as a definite price** — a one-line check in `@civitai/app-sdk`, another repo.
- **Whether `release` matches `main`** for any file cited.

### Environment note for reviewers
A fresh worktree has an **empty `event-engine-common` submodule**, which makes `blocks.router.workflow.test.ts` report `Tests no tests` — an import failure that reads as "nothing to see" rather than as a failure. Populate the submodule before trusting a test count from this file.

### DIFF
--- a/src/server/routers/__tests__/blocks.router.workflow.test.ts
+++ b/src/server/routers/__tests__/blocks.router.workflow.test.ts
@@ -42,6 +42,8 @@ const {
   mockLogToAxiom,
   mockSysRedis,
   mockResolveCanGenerateForVersions,
+  mockGetResourceData,
+  mockGetHighestTierSubscription,
   mockRecordSpendAttribution,
   mockDbWriteUserFindUnique,
 } = vi.hoisted(() => ({
@@ -129,6 +131,10 @@ const {
   // out of the test and we can drive canGenerate per-test. Default = a Map
   // saying the version IS generatable; FORBIDDEN tests override to false / miss.
   mockResolveCanGenerateForVersions: vi.fn(),
+  // The customComfy INLINE arm's entitlement belt calls these two directly.
+  // Stubbed as the belt's DATA SOURCE only — the belt itself runs for real.
+  mockGetResourceData: vi.fn(),
+  mockGetHighestTierSubscription: vi.fn(),
   // W3 flow A — the spend-attribution write the submit path fires
   // best-effort after a resolved submit. Mocked at the module boundary so
   // the test asserts exact (server-derived) args + that a throw here never
@@ -434,8 +440,19 @@ vi.mock('~/server/services/buzz.service', () => ({
 vi.mock('~/server/utils/block-catalog-rate-limit', () => ({
   checkBlockCatalogRateLimit: (...args: unknown[]) => mockCheckBlockCatalogRateLimit(...args),
 }));
+// 🔴 `getResourceData` IS HERE ON PURPOSE, AND IS NOT MOCKED AWAY AT THE SERVICE
+// BOUNDARY. The customComfy INLINE arm's entitlement belt
+// (`services/blocks/inline-comfy.service`) is deliberately left REAL in this
+// suite, with only its data source stubbed. Mocking the belt itself would reduce
+// every inline test below to "the router called a function I replaced", which is
+// exactly the seam a per-component test cannot see: the belt can be perfectly
+// correct in isolation and simply not wired in.
 vi.mock('~/server/services/generation/generation.service', () => ({
   resolveCanGenerateForVersions: (...args: unknown[]) => mockResolveCanGenerateForVersions(...args),
+  getResourceData: (...args: unknown[]) => mockGetResourceData(...args),
+}));
+vi.mock('~/server/services/subscriptions.service', () => ({
+  getHighestTierSubscription: (...args: unknown[]) => mockGetHighestTierSubscription(...args),
 }));
 vi.mock('~/server/logging/client', () => ({
   logToAxiom: (...args: unknown[]) => mockLogToAxiom(...args),
@@ -617,6 +634,8 @@ beforeEach(() => {
     mockSysRedis.expire,
     mockSysRedis.ttl,
     mockResolveCanGenerateForVersions,
+    mockGetResourceData,
+    mockGetHighestTierSubscription,
     mockRecordSpendAttribution,
     mockDbWriteUserFindUnique,
     mockIsAppBlocksAuthorEnabled,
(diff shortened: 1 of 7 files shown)


ZacxDev:
### Audit corrections to this PR's description

An adversarial audit ran against `025529a01e`. Posting these publicly rather than editing the description, since a reviewer may already have read the earlier version.

**1. One deploy-blocker was found and is now fixed in `3e9ac0d5`.**

The arm-narrowing predicate keyed on the *presence* of `mode`, but the schema declares `mode: z.literal('recipe').optional()` on the recipe arm — so `mode: 'recipe'` was a valid recipe spelling that got routed down the inline path. That produced a `TypeError` surfacing as a 500 on submit, and on estimate a silent `cost: {}` with `total` undefined instead of `cost: { total: 20 }`. `mode: undefined` behaved identically, and `undefined` survives the wire, so an SDK spreading an optional variable would have hit it.

It is now keyed on the value (`body.mode === 'inline'`). Four regression tests were added covering both spellings across submit and estimate; all four were confirmed red at `025529a01e`.

Worth noting for reviewers: the wire shape presents `mode` as *the* discriminator, which is what invited the broken spelling. Nothing in the original tests covered it — the back-compat test only exercised a body with no `mode` key at all.

**2. "early-access comes with it" is true of the code but inert for the population that can reach this path today.**

`getResourceData` treats a moderator as having access, so the early-access branch cannot bite for a moderator. App Blocks is moderator-only pre-GA, so for every viewer who can currently reach this code, the entitlement gate reduces to: covered, valid status, generation-supported base model, not generation-disabled, plus the anti-drop and anti-substitute checks.

The description discloses this exemption for the Private-model branch but not for early-access, where it applies equally. The code is correct; the claim about what it currently *does* is what needs the caveat. This becomes live protection at GA, when non-moderators can reach the path.

**3. The "19/19 mutations killed" figure understated the situation, and is now 21/21.**

One guard — the AIR-`type` allowlist — had **no** killing mutant. Its dedicated test asserted a substring that both it and the adjacent source-allowlist guard emit, so under mutation the sibling's message satisfied the assertion and the test stayed green. The guard was live and correct; a regression in it would simply have been invisible.

Two mutants were added: one for that allowlist (using an allowlisted source so no sibling can mask it, asserting that guard's own message) and one restoring the presence-narrowing bug from item 1. Both are killed, and the battery was re-run in full — a fix round resets the gate.

Also worth recording honestly: on the original battery, 3 of 19 mutants died to a *sibling* guard rather than the one removed. "It died" and "it died for its own reason" are different claims, and only the second means the guard is tested.

**4. `inline-comfy-budget.test.ts` does not exist.**

The comment at `src/server/schema/blocks/workflow.schema.ts:312` cites it as pinning the budget derivation. The pin is real, but it lives in `workflow.schema.inline-comfy.test.ts`. Worth correcting, since this PR fixes two other stale file references and a dead path in a comment is what gets a real guard deleted by someone who trusts it.

---

Four further findings were left as deliberate follow-ups rather than fixed here, to keep this diff reviewable: prefix-anchored AIR containment, the `huggingface` source being an ungated arbitrary-download surface, the moderator exemptions above, and item 4. The `huggingface` one in particular deserves an explicit product decision recorded next to the nodepack one — it is the same capability class this PR deliberately ships closed.

PR #3661, opened 2026-08-05
test(apps): hold the copied state still instead of racing its 1s reset

ZacxDev:
Follow-up to #3654, same defect class: **an assertion whose truth depends on how fast the runner is.** This fixes the second known instance and reports a survey of the rest of the tier.

## What the `CliSubmitCta` race actually was

Neither clipboard permission nor focus. It is a **real timer in the component's dependency**.

`CliSubmitCta` renders each one-liner in a Mantine `CopyButton`. `useClipboard` arms `window.setTimeout(() => setCopied(false), timeout)` the instant the clipboard promise resolves, and `CopyButton`'s `defaultProps` are `{ timeout: 1e3 }` — which `CliSubmitCta` does not override. So `Copied` is not a settled end-state, it is **transient with a ~1s wall-clock lifetime**, and both tests asserted it with `await expect.element(page.getByText('Copied')).toBeInTheDocument()`.

A retrying assertion cannot recover from missing that window. It polls for *presence*, so once the label is gone it keeps polling an absent element until it times out. Machine load, not a regression, decides the verdict — which is how the keyboard case went red on #3653's preview run after passing 1280/1280 ten minutes earlier.

Instrumented on an idle box, measured rather than assumed:

| probe | measurement |
|---|---|
| `Copied` visible window | **976.7 / 995.6 / 999.4 / 1000.5 / 999.8 / 1001.2 ms** |
| label appears after interaction resolves | 0.0–0.1 ms |
| `el.focus()` | 0.7 ms, `document.activeElement === el` every run |
| `'clipboard' in navigator` / `document.hasFocus()` | `true` / `true` |

The clipboard was never in question — `test/component-setup.tsx` already stubs `writeText` to a resolving `vi.fn()` precisely so this is deterministic. Focus was never in question either. The only wall-clock dependency is the reset timer.

## The fix

The same restricted virtual clock #3654 established, installed **immediately before the interaction**. Ordering is load-bearing: `vi.useFakeTimers` captures nothing retroactively and the reset timer is armed by the copy itself, so installing it afterwards would be too late.

The label then persists however slow the runner is, which lets the window be asserted explicitly instead of raced — copied, still copied at 999 ms, reset at 1000 ms with the command back. Nothing about the component changes: the same `CopyButton` arms the same `setTimeout`; only the reset's *arrival* is now under the test's control.

Everything up to the copy still runs on the real clock, and both driver interactions (`.click()` and `userEvent.keyboard('{Enter}')`) run fine under the frozen clock — worth noting because `PageBlockHostAutoRetry`'s header cautions that "driver interactions must not race a faked clock". That caution turns out to be conservative; #3654 already drove `fill()` this way.

## Verification

**Mutation — the guard is not weaker.** Each mutant red at its own assertion, with its own message:

| mutation | result |
|---|---|
| `copy` → no-op on **both** the Box and the button | red, `expected +0 to be 1`, at the `copiedLabels()).toBe(1)` guard, both tests |
| `CopyButton timeout` 1000 → **500** ms | red at the 999 ms boundary assertion |
| `CopyButton timeout` 1000 → **2000** ms | red at the post-reset assertion |
| remove `onClick={copy}` from the **button only** | **green** — exactly as the file's existing HONEST CAVEAT predicts (both handlers are React-synthetic and share the delegated root). The caveat is still accurate. |

**Freeze control — the clock is frozen, not merely fast.** With 2000 ms of REAL time inserted before the first assertion: green under the virtual clock, **red under real timers** at the same guard.

`afterEach(() => vi.useRealTimers())` rather than a `finally`, which does not run when a test times out.

**Repeat runs:** 8×, 9/9 tests counted green every run, 706–835 ms. Local browser is **non-canonical** (nixpkgs Chrome for Testing); CI's component job is the authority for the tier.

Also clean: `prettier --check` (verified against a positive control that does violate), `eslint`, and a full `typecheck` (0 errors in 100 s under an 8 GB heap cap — not an OOM reporting zero).

## Survey of the rest of the tier — and why nothing else is changed here

All **128** `src/**/*.browser.test.tsx` files were swept for the two-part signature: a real timer in the component under test **and** an assertion positioned so a slow runner changes the verdict. Two candidates survived to a measurement, and **both were measured and left alone**:

- **`PageBlockHostLaunchReveal.browser.test.tsx:303`** — `expect(Date.now() - firedAt).toBeLessThan(LAUNCH_REVEAL_MS)` (260 ms). Structurally the purest form of the defect, but **measured 11–22 ms across 5 runs — a 12–24× margin**, versus the reference case which was already *over* its budget at 338 ms/300 ms. It also has no clean fix: the reveal `setTimeout` is armed during `driveToReady()`, before any point where a clock could be frozen, so converting it would mean restructuring a working, mutation-verified guard. Left as-is deliberately.
- **`PageBlockHostReviewMode.browser.test.tsx:445`** — the theory was that the preceding `vi.waitFor` posts a `REQUEST_CONSENT` per retry and burns the 30-message/1000 ms limiter in `usePostMessage.ts`. **Ruled out by measurement:** the `waitFor` succeeds on its *first* iteration in 5/5 runs (0–1 ms), for 17 of 30 messages, and the retry count is not load-dependent because the notification fires synchronously off the dispatched `MessageEvent`.

A large structural class was ruled out on direction rather than margin: the ~20 `await new Promise(r => setTimeout(r, 150))` + assert-**absence** pairs across `AppBlocks`. Slowness delays the thing further, so those can only go false-**green**, never red. Worth knowing they exist, but they are not this defect.

The honest summary: **of 128 files, exactly one was really racing, and it is the one fixed here.** The tier is in better shape than the two back-to-back flakes suggested.

One thing I could not verify: I do not have the #3653 preview-run log, so the causal link between that specific red and the 1 s window is inferred from the mechanism plus the measured window, not read off the failure output.

https://claude.ai/code/session_01Aj1HosDFkP6LgxMAtJdDmJ

### DIFF
--- a/src/components/Apps/CliSubmitCta.browser.test.tsx
+++ b/src/components/Apps/CliSubmitCta.browser.test.tsx
@@ -1,4 +1,4 @@
-import { describe, expect, test } from 'vitest';
+import { afterEach, describe, expect, test, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
 import { CliSubmitCta } from '~/components/Apps/CliSubmitCta';
 import {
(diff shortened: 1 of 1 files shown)


ZacxDev:
### The causal link is now confirmed, not inferred

The PR notes the connection to civitai#3653's red run was reasoned from the mechanism rather than read off the failure. I had that Tekton log, so here it is closed.

#3653's `preview / component-tests` failed at:

```
FAIL  src/components/Apps/CliSubmitCta.browser.test.tsx
      > the copy button is focusable and keyboard-Enter copies — shows "Copied"
 ❯ src/components/Apps/CliSubmitCta.browser.test.tsx:115:51
```

Line 115 in the pre-fix file is exactly the assertion this PR fixes:

```js
await expect.element(page.getByText('Copied')).toBeInTheDocument();
```

— the retrying presence check on the label whose lifetime this PR measured at **976.7–1001.2ms**.

Two things that follow:

1. **The mechanism is confirmed.** The failure is at the assertion that races the `useClipboard` reset timer, not somewhere incidental.
2. **The focus theory is independently dead.** Line 113 — `expect(document.activeElement).toBe(el)` — is *before* the failing line and passed, so focus had settled. That matches the 0.7ms `el.focus()` measurement in the PR and rules out the explanation I originally guessed at.

For the record on the other side: #3653's run was `1 failed | 1279 passed (1280)`, and civitai#3654's canonical run ten minutes earlier was `1280 passed (1280)` — same suite total, so this test ran and passed there. Not broken on `main`; a genuine race.

PR #3673, opened 2026-08-06
fix(generation): split `api` out of the `onsite` model-substitution surface (#3665)

### LINKED ISSUE #3665: generateFromGraph: silent model substitution is billed and invisible to non-browser callers
## Summary

`orchestrator.generateFromGraph` silently substitutes the ecosystem's default model when the submitted `model.id` isn't valid for a `modelLocked` ecosystem, **charges the caller for the substituted model, and returns nothing in the tRPC reply indicating it happened.** A non-browser caller has no way to detect it.

The substitution itself is deliberate graceful degradation and I'm not asking to change it. The ask is to **surface the record that already exists** on the ordinary tRPC reply.

## Reproduction

Read-only `whatIfFromGraph` (zero spend), personal API key, re-verified 2026-08-06:

```
POST-equivalent GET /api/trpc/orchestrator.whatIfFromGraph
  {"workflow":"txt2img","ecosystem":"NanoBanana","prompt":"a cat","model":{"id":<X>}}
```

| `model.id` | result |
|---|---|
| `2436219` (correct) | HTTP 200, cost **160** |
| `999999999` (**does not exist**) | HTTP 200, cost **60** |
| *(no `model` key at all)* | HTTP 200, cost **60** |

A **nonexistent** model id returns a clean 200 at the ecosystem-default price. The reply is byte-indistinguishable from success.

## Why the existing signal doesn't reach us

`src/shared/data-graph/generation/common.ts` (~:978-1000) already records it:

```ts
ext.modelSubstitutions?.record({ requested, applied, ecosystem, workflow });
```

and its own comment is explicit that this is the problem:

> 🔴 OBSERVE ONLY — … this records that it happened so a caller who is billed for model A and given model B can find that out. On-site the correction is visible (the picker snaps); through the App Blocks bridge the id is deliberate and the correction is invisible — same code, same behaviour, observable in one context and not the other.

The record is surfaced through the App Blocks path via workflow metadata, but `grep -c modelSubstitutions src/server/routers/orchestrator.router.ts` → **0**. So the plain tRPC path — which is what a CLI or any external integration uses — never sees it.

Price divergence is not a usable substitute signal: it only helps if you already know the expected price, and a substitution between similarly-priced models is completely invisible.

## Ask

1. Include the substitution record on the `generateFromGraph` (and ideally `whatIfFromGraph`) reply, using the same key the App Blocks bridge already populates.
2. **Give non-onsite callers a distinct substitution-context label.** `orchestrator.router.ts` (~:344-358) hardcodes `'onsite'`, and its comment says `'onsite'` specifically means the *visible*, benign degradation and is deliberately excluded from the incidence number gating the phase-3 policy decision. CLI/API traffic recorded as `'onsite'` would inflate the benign bucket with exactly the structurally-invisible cases — corrupting the metric used to decide whether to fix this class.

## Context

Surfaced while building `civitai generate` in [civitai/cli#210](https://github.com/civitai/cli/pull/210). The CLI mitigates the demonstrated case by resolving `--checkpoint`/`--lora` against `GET /api/v1/model-versions/{id}` before submitting, which turns a nonexistent id into a hard local 404. That cannot catch **same-ecosystem** substitution, which needs the server-side record.

ZacxDev:
Closes #3665 (item 2 of its ask). Refs #3520.

## What's wrong

`orchestrator.router`'s `generateFromGraph` / `whatIfFromGraph` hardcode `'onsite'` as the `surface` label on `civitai_generation_model_substitutions_total`.

Both procedures are deliberately reachable by bearer token — `AIServicesRead` / `AIServicesWrite` are user-grantable scopes carrying the consent label *"Generate, train & scan"*. So a CLI, a script, or any external integration hits this exact code path, and every silent model substitution it suffers is filed under the one case #3520 defends as **correct**: a stale `localStorage` id after an ecosystem switch, with the picker visibly snapping back.

An API caller has neither property that makes degradation correct — the id is deliberate, and there is no picker to snap. Worse, `'onsite'` is precisely the bucket phase 3's policy decision *excludes*, so these cases were being counted into the noise floor.

## Why now

Over the counter's life so far, `surface="block"` — the population phase 3 is actually about — has recorded **zero** substitutions, because App Blocks is mod-gated. The entire signal sits in `surface="onsite"`, at an incidence on the order of 1 in 10⁴ calls.

A gate that reads zero on the only bucket it consults cannot gate anything. Splitting `api` out is what gives the phase-3 decision a population with both real volume *and* the invisibility property that makes substitution harmful.

**Behaviour is unchanged.** This is a label, not a policy. Which model runs, what is billed, and what any caller receives are all identical. The reject-vs-degrade question stays open and stays gated — this just makes it gateable.

## How

- **`generationSurfaceForRequest(ctx)`** derives the surface from `ctx.subject` / `ctx.apiKeyId` — the same discriminator `Tracker.setProvenance` already uses to tag content-creation events as `web` / `api-key` / `oauth`. Both fields are read because `get-server-auth-session` assigns them independently. `tokenScope` cannot serve here: it is `Full` for a cookie session *and* for a full-access personal API key.
- **`api` is one value, not two.** A personal key and an OAuth app differ in *who* is calling, not in whether the correction is observable — the only thing this label feeds.
- **`GENERATION_SURFACES` 3 → 4**, cardinality 9 → 12 series. Deliberate and gated: the exact-union assertions and the at-most-N-series test all had to be edited by hand, which is the intended friction.

## The guard change, stated plainly

`generation-surface-wiring` pinned every `buildGenerationContext` call site to a **string literal**. One call site now genuinely cannot be pinned that way, so the table learns a *resolved* pin: "resolved by exactly this function, whose range is exactly these surfaces."

It still rejects a new call site, a different expression (variable, ternary, second resolver), and a re-hardcoded literal. What it gives up is knowing *which* of the two a given request yields — not knowable statically — which is covered by execution instead. The resolver's range is imported from beside the resolver rather than restated in the guard, so it cannot drift.

## Verification

`pnpm typecheck` → **0 errors**, and watched go **red** on a deliberate seam break (narrowing `GenerationSurfaceRequest['subject']` to `string` produced 2 errors at exactly the two router call sites) — so the green is a fact about the code, not about the command line. That break is also the proof that the production `generationSurfaceForRequest(ctx)` call *is* the compile-time guarantee that the tRPC `Context` still satisfies the resolver's input.

276 tests pass across the 24 affected suites. Each new guard was watched fail on the mutation it exists to catch:

| mutation | killed by | failed with |
|---|---|---|
| re-hardcode `'onsite'` at both call sites | wiring guard | `passes literals [onsite] … expected exactly one call to generationSurfaceForRequest(...)`, naming both sites |
| inline `ctx.subject ? 'api' : 'onsite'` at the call sites | wiring guard | same assertion, `passes literals [] / calls []` |
| resolver always returns `'onsite'` | `generation-surface.test.ts` | 7 tests, incl. the named `#3665 regression` assertion |
| resolver range padded with an unreachable surface | `generation-surface.test.ts` | `expected [ 'api', 'onsite' ] to deeply equal [ 'api', 'onsite', 'preset' ]` |

## Not in this PR

The other half of #3665 — surfacing the substitution record on the tRPC replies — is deliberately separate. Note for whoever takes it: `formatGenerationResponse2` builds `normalizedWfMeta` from an **allowlist** (`params`, `resources`, `remixOfId`, `isPrivateGeneration`), so copying the App Blocks trick of persisting under `WORKFLOW_METADATA_MODEL_SUBSTITUTIONS_KEY` will be silently dropped on read-back unless the normalizer gains an explicit passthrough.

### DIFF
--- a/src/pages/api/metrics.ts
+++ b/src/pages/api/metrics.ts
@@ -6,7 +6,20 @@ import { instrumentationRegistry } from '~/server/prom/client';
 // Side-effect import: registers the challenge state gauges (collect()-based) on the default
 // registry so they are present + scraped even before the first challenge op runs this session.
 import '~/server/prom/challenge.metrics';
+// Same reason (#3665): seeds all 12 (reason, surface) series of
+// civitai_generation_model_substitutions_total at 0. Without this call the
+// counter is registered only by the FIRST substitution on this pod, so
+// `…{surface="api"}` read `no data` until then — indistinguishable from an
+// unwired instrument, on a metric whose entire job is to be read as a gate.
+//
+// Next.js loads an API route lazily, so this runs on the first REQUEST to
+// /api/metrics rather than at pod start — which is exactly soon enough: the
+// only consumer is the scrape, and this module is what serves it, so the series
+// are present in every response that could ever observe them.
+import { ensureRegisterGenerationModelSubstitutionMetrics } from '~/server/metrics/generation-model-substitution.metrics';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
+
+ensureRegisterGenerationModelSubstitutionMetrics();
 
 const labels: Record<string, string> = {};
 if (process.env.PODNAME) {
--- a/src/server/metrics/__tests__/generation-model-substitution.metrics.test.ts
+++ b/src/server/metrics/__tests__/generation-model-substitution.metrics.test.ts
@@ -31,12 +31,13 @@ import {
  *      500 — observability causing the outage.
  *
  * 🔴 CARDINALITY is the other load-bearing property: `reason` (3 values) x
- * `surface` (3 values) = 9 series, total, forever, across ~130 scraped pods
+ * `surface` (4 values) = 12 series, total, across ~130 scraped pods
  * where prom-client retains every distinct label set in the Node heap for the
  * process lifetime. Widening it has to get past these tests.
  *
  * 🔴 SURFACE is load-bearing for a different reason: `validateInput` is shared
- * by the on-site generator, the App Blocks bridge and preset generation, and
+ * by the on-site generator, bearer-token API callers, the App Blocks bridge and
+ * preset generation, and
  * #3520 calls the substitution CORRECT on-site and unobservable through the
  * bridge. Summed into one series the counter measures a contaminated quantity,
  * which is exactly what phase 2 is supposed to produce a clean number for.
@@ -74,6 +75,63 @@ describe('civitai_generation_model_substitutions_total', () => {
   it('is registered on the default registry that /api/metrics scrapes', () => {
     ensureRegisterGenerationModelSubstitutionMetrics();
     expect(client.register.getSingleMetric(METRIC)).toBeDefined();
+  });
+
+  it('🔴 the help string documents EVERY declared surface and reason', async () => {
+    // The help text is what Grafana's metric browser shows, i.e. the definition
+    // a phase-3 consumer actually reads before writing a query. `api` was added
+    // to the union in #3665 and initially NOT to the help string, so the label
+    // this metric exists to expose was undocumented where it is consumed.
+    // Derived from the unions rather than hardcoded, so a future 5th surface
+    // cannot be added without describing it.
+    //
+    // 🔴 WHAT THIS DOES NOT CHECK: that each description is CORRECT, or even
+    // attached to the right label. `toContain` is satisfied by any occurrence of
+    // `<surface> = `, so a help string reading `api = block; block = …` passes
+    // (verified). It catches the regression it was written for — a surface added
+    // to the union and not to the help — and nothing more. Do not cite it as
+    // evidence that the help text is right.
+    ensureRegisterGenerationModelSubstitutionMetrics();
+    const metric = client.register.getSingleMetric(METRIC) as unknown as { help: string };
+    for (const surface of GENERATION_SURFACES) expect(metric.help).toContain(`${surface} = `);
+    for (const reason of MODEL_SUBSTITUTION_REASONS) expect(metric.help).toContain(`${reason} = `);
+  });
+
+  it('🔴 registration SEEDS all 12 series at 0 — an absent series must not read as a real zero', async () => {
+    // A pod only materialises a prom-client child on its first inc(), so before
+    // seeding `…{surface="api"}` simply did not exist until that pod happened to
+    // serve an API substitution — and PromQL returns `no data`, which looks
+    // exactly like "the instrument was never wired". This counter exists to be
+    // read as a GATE, so that ambiguity is the same defect the `api` split was
+    // made to fix (#3665), one level down.
+    ensureRegisterGenerationModelSubstitutionMetrics();
+    const metric = client.register.getSingleMetric(METRIC) as unknown as {
+      get(): Promise<{ values: Array<{ labels: Record<string, string>; value: number }> }>;
+    };
+    const { values } = await metric.get();
+
+    expect(values).toHaveLength(MODEL_SUBSTITUTION_REASONS.length * GENERATION_SURFACES.length);
+    expect(values.every((v) => v.value === 0)).toBe(true);
+    // Named explicitly: `api` is the surface this PR adds and the one a phase-3
+    // read will query first, so its presence-at-zero is the point.
+    expect(await readSeries('unrecognized', 'api')).toBe(0);
+    // Every declared combination, not just the one we thought of.
+    for (const reason of MODEL_SUBSTITUTION_REASONS)
+      for (const surface of GENERATION_SURFACES) expect(await readSeries(reason, surface)).toBe(0);
+  });
+
+  it('seeding is idempotent — re-registering does not reset or double-count a live series', async () => {
+    // `ensureRegister…` is documented as safe to call on every request, and
+    // `recordGenerationModelSubstitution` does call it every time. If seeding
+    // reset the counters, the metric would read ~0 forever under real traffic —
+    // a failure that a single-shot test would never show.
+    recordGenerationModelSubstitution('unrecognized', 'api');
+    recordGenerationModelSubstitution('unrecognized', 'api');
+    expect(await readSeries('unrecognized', 'api')).toBe(2);
+
+    ensureRegisterGenerationModelSubstitutionMetrics();
+    ensureRegisterGenerationModelSubstitutionMetrics();
+    expect(await readSeries('unrecognized', 'api')).toBe(2);
   });
 
   it.each(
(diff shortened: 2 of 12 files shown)


ZacxDev:
## Adversarial audit round 1 — findings and fixes (`d6c14e3` → `9bdbaaa`)

Posting rather than editing the body, since the body's verification section described a coverage property that turned out to be **false**. Anyone who already read it should read this.

### The docblock overclaimed

`generation-surface-wiring.test.ts` said the resolved pin "keeps every mutation this guard exists to catch." It did not. Three mutants passed **84/84 guard tests and a clean `tsc`**, each reinstating the exact defect this PR fixes — 100% of both procedures labelled `onsite`:

| mutant | why it survived |
|---|---|
| `generationSurfaceForRequest({})` | `GenerationSurfaceRequest` had all-optional fields, making it a **weak type** — TS's weak-type check accepts any object sharing one property name, so `{}` type-checked |
| local `function generationSurfaceForRequest(_ctx) { return 'onsite' }` shadowing the import | the guard matched the callee **name**; an identifier is just a name in scope |
| `const generationSurfaceForRequest = alwaysOnsite` | same |

Fixed by: requiring both keys on `GenerationSurfaceRequest` (values may still be `undefined`, which is what the real `ctx` provides — `{}` is now a compile error); and requiring the call site's file to import that name **from the module that defines it**. All three now die, verified individually.

The docblock now states honestly what a name-matching AST guard structurally *cannot* do — it cannot verify the **argument**, only the callee. `tsc` covers the argument, and it only has teeth because the parameter type is non-weak. Relaxing that type back to optional fields silently removes the coverage; that's written down at both ends now.

### A false positive I'd introduced

The previous revision scanned *every* argument for a string literal or a call. That accepted the resolver in **any slot**, and false-failed the moment an *unrelated* argument became a call — e.g. `String(user.tier ?? 'free')` in slot 0 of a `'preset'` site failed with a message about surfaces. The guard now reads the surface argument **by position** (index 3), which fixes both directions. Verified: slot-swap fails; unrelated-arg-as-call passes.

### The metric had the PR's own problem, one level down

prom-client only materialises a child series on first `inc()`, so `…{surface="api"}` would have returned `no data` on a pod until that pod happened to serve an API substitution — indistinguishable from an unwired instrument. That is exactly the argument this PR makes against `block`. Registration now seeds all 12 `(reason, surface)` combinations at 0. Free (12 is the whole cardinality budget) and idempotent; both properties tested, including that re-registration doesn't reset a live series.

### Documentation the consumer actually reads

`api` was missing from the metric's own `help` string — what shows in Grafana's metric browser, and the first thing a phase-3 consumer reads. Also added to the `@param surface` JSDoc and two module comments that still enumerated three surfaces.

### Also fixed: production figures in a public repo

The code comment carried measured traffic volumes. Scrubbed — it now states the *shape* of the measurement (`block` recorded nothing; `onsite` carried the whole signal) and points at the private infra repo for figures. Worth a general note: `~130 scraped pods` (fleet size) is on `main` already at `generation-model-substitution.metrics.ts:69` and in the metrics test, predating this PR.

### Unchanged by this round

Still no behaviour change, no wire/API change, no zod enum widened. Independently confirmed during the audit: nothing in the infra repo consumes this metric (no dashboard, alert, or recording rule), and only 5 production `buildGenerationContext` call sites exist — enumerated, not sampled.

**Open question I could not close:** whether any *first-party* Civitai surface (Link, a desktop client, an internal script) authenticates to these procedures with an API key. If one does, its substitutions move from `onsite` to `api` and dilute the "external integration" reading of the new bucket. I found no such caller in this repo but cannot rule out an out-of-repo one — if you know of one, say so and I'll narrow the label.

ZacxDev:
## Round 2 (delta re-audit) — and a history rewrite

**The branch is now a single squashed commit** (`95eb9d6`). Tree verified byte-identical to the pre-squash tip; only the history changed. The earlier comment referencing `d6c14e3` / `9bdbaaa` points at commits that are no longer on the branch — see the last section for why.

### A surviving mutant of the class round 1 closed

Round 1 made the guard require the resolver to be *imported from its defining module*. That check compared the **local binding**, so this passed with the guard 6/6 green and `tsc` at 0 errors, while resolving every request to `onsite` — the exact #3665 defect:

```ts
import { alwaysOnsite as generationSurfaceForRequest } from '~/server/services/orchestrator/generation-surface';
```

A comment right above the check even blessed it ("would still be this module's export, which is fine"). That was wrong: *this module's export* is not the same claim as *the approved resolver*. Now compares the **exported** name (`propertyName ?? name`). Verified the mutant now dies.

Worth stating plainly, since round 1's writeup already overclaimed once: what this guard can and cannot do is now written at both ends. It checks the **callee**; it cannot check the **argument**. `tsc` checks the argument, and only because the parameter type is non-weak — relax that type and the coverage silently disappears.

### The seeding was inert

Round 1 seeded all 12 series at registration, and the docstring claimed `rate()`/`increase()` become "well-defined from the first scrape". False as shipped: `ensureRegister…` had **no caller outside its own module**, and the only production path in is `recordGenerationModelSubstitution` — which by definition runs only *after* a substitution. So the metric was still absent until the first event on that pod, which is the ambiguity the seeding exists to remove.

Fixed with a module-scope call in `src/pages/api/metrics.ts`, mirroring the `challenge.metrics` side-effect import two lines above it, which is there for exactly this reason. Two components, each fine alone, broken together — so there is now a seam guard pinning the *relationship*. It is AST-based (a commented-out call leaves the text in place and would satisfy a substring check) and module-scope-only (moving the call into the handler makes it per-request, which is a different and wrong thing). Both mutants verified red.

I also probed the behavioural route first and it is not available: importing the page module in a unit test throws `env.TRPC_ORIGINS is not iterable`. That's recorded in the test rather than left as an unexplained design choice.

### Smaller

- The resolver's own fixture table no longer satisfied the narrowed type — hidden only by `tsconfig`'s `__tests__` exclude, the same hole this PR's guard documents. Every row now spells both keys.
- The provenance check fails closed on a **namespace import** and on a **relative specifier** for the same module. Both are legitimate code (relative imports are used ~358× in `src/server`), so the failure message now says which case you're in and what to do instead of implying sabotage.
- Three residual "three surfaces" docstrings, an orphaned JSDoc block documenting nothing, and a help string with no test — the last now derived from the unions, so a 5th surface can't be added without describing it.

### The history rewrite, and what it does *not* fix

The first push contained production traffic figures in a code comment. I scrubbed the tip, but that was not remediation: the figures remained live at the parent commit, reachable via `refs/pull/3673/head` on a public repo. The delta audit caught that the fix didn't fix it.

The branch is now squashed so no commit in its history contains them, verified by sweeping every commit in `main..HEAD`. **Caveat, stated because it matters:** a force-push updates the refs, but GitHub may retain the orphaned objects, so anyone who already has the old SHA could still fetch it. If those figures are considered genuinely sensitive rather than merely against convention, that needs a call to GitHub support — it is not something a force-push settles. Flagging rather than deciding.

`pnpm typecheck` 0 errors; 682 tests green across the 32 affected suites.

ZacxDev:
## Round 3 — the loop's stop condition, and two of my own claims retracted

Round 3 returned **no 🔴 and a "safe to merge" verdict**. It did find two places where a comment I wrote asserted something false, so those are fixed rather than left standing. Tip is now `f5e9ca1`.

### Retraction 1: "the behavioural route is not available"

Round 2 added an **AST** seam guard that parsed `metrics.ts` looking for a module-scope call, and justified it with a comment saying importing the page in a unit test throws `env.TRPC_ORIGINS is not iterable`, so a behavioural test was impossible. I had measured the throw — but "my first attempt failed" is not "it cannot be done." The repo already had the pattern (`src/server/utils/__tests__/public-endpoint-maxage.test.ts` stubs the heavy module-load imports), and the audit used it to load the real page module.

The source check was also over-strict in a way that would bite someone later: it false-failed at least six *correct* refactors — `void f()`, a module-scope helper, a locally-aliased import, a relative specifier — while **passing** an `import type` + call, which only `tsc` would catch. And its failure was a bare `toEqual` diff with no explanation, the same defect round 2 had just fixed on the sibling guard and didn't carry over.

Replaced with a behavioural test that loads the real page module and asserts the 12 series exist at 0. Verified strictly stronger, in both directions:

| | old AST guard | new behavioural |
|---|---|---|
| call deleted | RED | **RED** |
| call moved into the handler | RED | **RED** |
| `void f()` | RED (false) | **GREEN** |
| module-scope helper + aliased import | RED (false) | **GREEN** |

### Retraction 2: "a LOCAL function or alias fails"

The wiring guard's docblock claimed the import-provenance check catches a local function or alias of the resolver's name. True for a *module-scope* one — but a **block-scoped** `const generationSurfaceForRequest` declared inside a single procedure, with the real import left at the top and still used by the sibling procedure, reinstates #3665 for that one procedure with the guard 6/6 green, `tsc` 0 errors and eslint clean.

The check is per-file, so closing this needs real scope resolution. I've left it open deliberately — it is an adversarial shape, not a refactor anyone reaches for — but the docblock now says so explicitly instead of implying it away. This is pre-existing (round 1's design), not something round 2 introduced.

### Calibration, not a change

The help-string test is satisfied by any occurrence of `<surface> = `, so a help string reading `api = block; block = …` would pass. It kills the regression it was written for (a surface added to the union but not the help) and nothing more. That limit is now written into the test rather than left for the next reader to discover.

### Also confirmed clean by the audit

No import cycle or bundle growth from the new static import (`model-substitution` is a zero-import leaf; `prom-client` was already imported there). Single prom-client registry — the seeded registry *is* the scraped one. No `collectDefaultMetrics` ordering issue. No double registration; re-registering after live increments leaves values intact. `/api/metrics` output grows by exactly 12 zero lines per pod, which is the intended bounded cost — and that was measured against the real handler, not inferred.

`pnpm typecheck` 0 errors; 683 tests green across 33 affected suites.

**Three rounds, and each one found something the previous round's fix had introduced or overclaimed.** Round 3's findings were confined to comment accuracy and guard ergonomics, which is where I'd expect to stop.

PR #3691, opened 2026-08-06
fix(notifications): notify post owners of comments on their posts

JustMaier:
Closes [CU-868kn0pw1](https://app.clickup.com/t/8459928/868kn0pw1). Freshdesk 68980.

## The gap

`src/server/notifications/comment.notifications.ts` defines an owner-notified processor for every commentable entity except **Post**:

| Entity | Processor | Thread join |
|---|---|---|
| Model | `new-comment` | `Comment.modelId` |
| Image | `new-image-comment` | `Thread.imageId` |
| Article | `new-article-comment` | `Thread.articleId` |
| Bounty | `new-bounty-comment` | `Thread.bountyId` |
| Challenge | `new-challenge-comment` | `Thread.challengeId` |
| 3D Model | `new-3d-model-comment` | `Thread.model3dId` |
| **Post** | **none** | **—** |

Nothing joined `Thread."postId"`, so a top-level comment on a post generated no notification for the post owner, ever.

Posts are handled everywhere *else* in that file — `threadUrlMap` has a `post:` entry, and both `new-comment-reply` and `new-thread-response` include `post` in their `threadType` CASE arms. So a reply inside a thread the owner had already joined *did* reach them, while the first comment on their own post did not. That asymmetry is why this was reported as patchy notifications rather than missing ones. On the reporting user's account alone, 15 comments by other users on posts he owns produced nothing.

## The fix

A 14th processor, `new-post-comment`, mirroring `new-image-comment`:

- joins `Thread."postId"` → `Post."userId"`, excludes self-comments and system-owned posts (`p."userId" > 0`)
- `CommentNotificationPriority.EntityOwner`, so a reply/mention on the same comment still outranks it via the shared `comment:v2:<id>` dedupe key
- the standard `UserNotificationSettings` opt-out row, and `NotificationCategory.Comment`
- links to `/posts/{postId}?highlight={commentId}`
- `Post.title` is nullable and untitled posts are common, so that shape renders `"<user> commented on your post"` rather than an empty quoted title

## No backfill blast

Worth calling out since the gap is a year deep: `send-notifications` reads the per-key cursor with `getJobDate('last-sent-notification-' + key, lastRun)`, and `getJobDate` falls back to the passed default when the key row is absent. A brand-new type therefore starts from the job's global last-run — a ~1 minute lookback on its first pass, not the full history. Existing comments stay unannounced; only new ones notify. That also means this ships without a hardcoded date floor like `new-bounty-comment` carries, keeping it identical to its siblings.

## Not in scope

The same reporter also asked that Articles get their own `NotificationCategory` instead of sitting under `Comment`. That is a feature request, tracked separately in `docs/feature-requests.md`.

## Testing

- `src/server/notifications/__tests__/comment.post-owner.test.ts` (new) — the postId join, the self-comment and system-owner guards, the opt-out clause, registration in `notificationProcessors`, and both title shapes rendering a message + URL.
- `comment.dedupe-key.test.ts` — `new-post-comment` added to `V2_TYPES` (v2-namespaced dedupe key, commentId-only derivation, declares a priority) and to the "a type that claims the dedupe key must render" matrix, with the null-title shape included.
- `pnpm run test:unit:run` — 12302 passed. 14 failures in 6 files (typecheck wrapper, AppBlocks bucket labels, block-scope normalize-endpoint, no-wholesale-module-mock, app-spend-tier-privilege, comics orchestrator-chat). Verified pre-existing: stashing this branch's changes and re-running those 6 files on the clean base reproduces the same 14.
- `pnpm run typecheck` — 0 errors. `eslint` on touched files — 0 errors (2 pre-existing `no-explicit-any` warnings).

### DIFF
--- a/src/server/notifications/__tests__/comment.dedupe-key.test.ts
+++ b/src/server/notifications/__tests__/comment.dedupe-key.test.ts
@@ -27,6 +27,7 @@ const sqlFor = (type: string) => defs[type].prepareQuery!({ lastSent: '2026-01-0
 const V2_TYPES = [
   'new-review-response',
   'new-image-comment',
+  'new-post-comment',
   'new-article-comment',
   'new-bounty-comment',
   'new-challenge-comment',
@@ -204,6 +205,12 @@ describe('a type that claims the dedupe key must render', () => {
         return [{ ...base, version: 2, reviewId: 3, modelId: 1, modelName: 'M' }];
       case 'new-image-comment':
         return [{ ...base, version: 2, imageId: 9, modelName: 'M', modelId: 1 }];
+      // Post.title is nullable, so both shapes have to render.
+      case 'new-post-comment':
+        return [
+          { ...base, version: 2, postId: 6, postTitle: 'P' },
+          { ...base, version: 2, postId: 6, postTitle: null },
+        ];
       case 'new-article-comment':
         return [{ ...base, version: 2, articleId: 4, articleTitle: 'A' }];
       case 'new-bounty-comment':
--- a/src/server/notifications/__tests__/comment.dedupe-key.test.ts
+++ b/src/server/notifications/__tests__/comment.dedupe-key.test.ts
@@ -98,6 +98,51 @@ describe('comment notifications — shared dedupe key', () => {
   });
 });
 
+/**
+ * `dedupeKey` collisions are guarded above; `key` collisions have the same blast radius and were not
+ * guarded anywhere. `Notification.key` is UNIQUE, and the drain looks a key up and REUSES the stored
+ * row's `type` and `details` rather than inserting. So two processors sharing a key prefix don't
+ * error — one silently serves the other's message and URL to the wrong recipient.
+ */
+describe('every processor owns a distinct notification key namespace', () => {
+  const keyPrefixes = Object.entries(notificationProcessors).flatMap(([type, def]) => {
+    const query = def.prepareQuery?.({
+      lastSent: '2026-01-01',
+      lastSentDate: new Date('2026-01-01'),
+      clickhouse: undefined,
+    });
+    if (typeof query !== 'string') return [];
+    // The literal that opens the concat feeding the `key` column.
+    const prefix = query
+      .split('"key"')[0]
+      ?.match(/concat\(\s*'([^']*)'/g)
+      ?.pop()
+      ?.match(/'([^']*)'/)?.[1];
+    return prefix ? [{ type, prefix }] : [];
+  });
+
+  it('finds a key prefix for the comment-family processors it is meant to cover', () => {
+    // Guards the extraction above from silently matching nothing and passing vacuously.
+    const covered = keyPrefixes.map((k) => k.type);
+    expect(covered).toContain('new-post-comment');
+    expect(covered).toContain('new-image-comment');
+    expect(keyPrefixes.length).toBeGreaterThan(10);
+  });
+
+  it('no processor can be confused for another by key', () => {
+    for (const a of keyPrefixes) {
+      for (const b of keyPrefixes) {
+        if (a.type === b.type) continue;
+        // Keys are prefix + id, so one prefix being a prefix of another is enough to collide.
+        expect(
+          a.prefix.startsWith(b.prefix),
+          `${a.type} ("${a.prefix}") collides with ${b.type} ("${b.prefix}")`
+        ).toBe(false);
+      }
+    }
+  });
+});
+
 describe('comment notifications — priority decides which duplicate survives', () => {
   it('orders mention > direct response > thread response > entity owner', () => {
     const { Mention, DirectResponse, ThreadResponse, EntityOwner } = CommentNotificationPriority;
--- a/src/server/notifications/__tests__/comment.post-owner.test.ts
+++ b/src/server/notifications/__tests__/comment.post-owner.test.ts
@@ -1,11 +1,16 @@
 import { describe, expect, it } from 'vitest';
-import { commentNotifications } from '~/server/notifications/comment.notifications';
+import {
+  CommentNotificationPriority,
+  commentNotifications,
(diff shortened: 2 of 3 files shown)


JustMaier:
## Adversarial review pass (3 reviewers: SQL/data, rollout, integration)

Pushed `6e920ba` in response. One blocking defect, several unguarded mutations, and a handful of clean bills of health.

### Blocking: the "no backfill" claim in the original description was wrong

I claimed `getJobDate` insulated this from a backfill. That's right on the happy path and wrong in general:

- `send-notifications.ts:30` reads the **global** cursor with **no default** → `new Date(0)` when the row is absent (fresh DB, preview env).
- `setLastRun()` (:135) only runs on whole-job success, and `e.checkIfCanceled()` (:34) sits *outside* the per-key try — a canceled run throws past it. Any sustained send-notifications outage leaves that row stale by the outage duration.
- Every *existing* processor is insulated by its own `last-sent-notification-*` row. A brand-new type has none, so it is the single key that inherits the global cursor verbatim.

Measured against prod: **110,516** historical rows (46,582 in the last year, 9,986 distinct owners) against **3/hour** steady state. The 20s query timeout is not a backstop — the unfloored query returns all of them in **2.6s**, and `createNotificationsBulk` chunks at 5000 with no total cap.

`new-bounty-comment`'s `AND c."createdAt" > '2024-02-24'` (line 632) is the precedent, and blame makes it explicit: `1ab68815bf` added that floor *in the same commit* that fixed a `b.title`→`b.name` bug which had been making the query error every run — i.e. added specifically so the newly-working query couldn't emit history. Same shape here.

Fixed with two floors: the launch date (no pre-launch comment, ever) plus `NOW() - INTERVAL '7 days'`, which bounds any first batch to ~500 rows however far the cursor drifted and means the date never needs bumping if merge slips.

### Test gaps — each verified by mutating the source and confirming red

| Mutation | Was | Now |
|---|---|---|
| `priority: EntityOwner` → `Mention` | green | red — at priority 1 this beats `new-mention` to the shared dedupe key, the exact regression the dedupe suite exists to prevent |
| `'commentId', c.id` → `t.id` | green | red — `?highlight=` matches nothing **and** `commentDedupeKey` reads `details->>'commentId'`, so the key silently goes thread-scoped and every comment after the first in a thread is suppressed |
| `JOIN "User" u ON c."userId"` → `p."userId"` | green | red — every message reads "\<you\> commented on your post" |
| `displayName`, `category`, `${lastSent}` bound, empty-title branch | green | red |

Also added a **repo-wide** guard that no two processors share a `key` prefix. `Notification.key` is UNIQUE and the drain (`poll-loop.ts:106`) looks a key up and **reuses the stored row's `type` and `details`** rather than inserting — so a collision serves one processor's message and URL to another's recipient, silently. `dedupeKey` was guarded exhaustively; `key` had no guard anywhere. No existing collisions found.

### Also changed

Untitled posts now name their model (`"X commented on your post on the Y model"`), mirroring `new-image-comment`. Post title is `z.string().trim().nullish()` so clearing it stores `''`, not null — untitled is the common case for model-gallery posts, and without a disambiguator a creator gets a run of identical rows.

### Verified clean (checked, not assumed)

- **Join shape.** `upsertComment` writes `postId` directly on the thread and `Thread_postId_key` is UNIQUE, so no `rootThreadId` indirection — matches image/article, correctly unlike `new-comment-reply`.
- **No reply spam.** A reply is `entityType: 'comment'`, so its thread has `postId` NULL. This catches top-level only.
- **No double-notify.** Dedupe is a permanent partial UNIQUE `(userId, dedupeKey)` with `ON CONFLICT DO NOTHING`. `new-comment-reply` can't match a post thread at all; `new-thread-response` can, but runs in an earlier sequential batch and is consumed `ORDER BY id`, so it lands first. Correct by construction.
- **Registration is fully automatic.** `notificationProcessors` → `notificationCategoryTypes` feeds the settings UI; `type` is `z.string()` everywhere (no enum), `NotificationCategory.Comment` already exists in both enum copies, and `comment.detail-fetcher.ts` keys off `Object.keys(commentNotifications)` so the avatar and body preview come free. No DB row, migration, or hardcoded list needed.
- **The link works.** `/posts/[postId]/[[...postSlug]]` is an optional catch-all so bare `/posts/6` resolves; `CommentsProvider` forwards `highlight` as `targetCommentId` and prepends the target to page 1 so it renders even past the cursor. Every id this emits is a root comment — the case that highlights without expanding a nested thread.
- **No email/push fan-out.** In-app + one SignalR ping only.
- **Deleted posts self-heal** (`Thread.postId` is `onDelete: SetNull`); blocked users are rejected at write time.

### Deliberately not changed

- **Shipping enabled, not `defaultDisabled: true`.** Reviewer found the 3D-model pilot's use of that flag is a live bug: it's a UI-only inversion with no server counterpart. `NotificationsCard.tsx:106` renders `checked = !settings[type]`, but the SQL sends when **no** row exists — so `defaultDisabled` shows the box unchecked while the cron sends anyway, and ticking it to "enable" actually silences it. Defaulting on is the only self-consistent option.
- **A post owner who has left a top-level comment on their own post** is in `new-thread-response`'s recipient set for every later comment, and the dedupe unique is permanent — so they keep getting "responded to a Post thread you're in" and never see "commented on your post". Inherited from the image/article pattern, at the existing bar, but it means this doesn't reach the most engaged owners. Worth a follow-up.
- **`CommentV2` has no `createdAt` index**, so this seq-scans every minute — identical in shape to the three sibling processors already in the same 8-wide batch. Incremental, not a new class, and the floors cap the rows.
- **`bountyEntry` has a `threadUrlMap` entry and no owner processor either.** Same gap as this one, still open. Out of scope here; noted in the test file header.

Full unit suite: 12312 passed, same 14 pre-existing failures as base (verified by stashing and re-running those 6 files). Typecheck 0 errors. No migration.

PR #3692, opened 2026-08-06
feat(generation): surface silent model substitutions on the tRPC replies (#3665)

### LINKED ISSUE #3665: generateFromGraph: silent model substitution is billed and invisible to non-browser callers

ZacxDev:
Closes #3665 (item 1 of its ask). Refs #3520. Companion to #3673, which does item 2.

## What's wrong

`generateFromGraph` / `whatIfFromGraph` substitute the ecosystem default when a submitted `model.id` isn't valid for a `modelLocked` ecosystem, bill the caller for what actually ran, and return a reply byte-indistinguishable from success. From #3665's repro: a **nonexistent** model id returns a clean HTTP 200 at the ecosystem-default price.

The record already existed — `ext.modelSubstitutions.record(...)` in the graph clamp — but only the App Blocks bridge surfaced it, so a CLI or any external integration had no way to detect the swap. `grep -c modelSubstitutions src/server/routers/orchestrator.router.ts` → 0.

**The substitution itself is unchanged.** This surfaces the record that was already being taken. Reject-vs-degrade stays open and stays gated.

## How

- **`whatIfFromGraph`** returns `modelSubstitutions` on its reply. This is the half that matters most for an external caller: it spends nothing and creates no persisted workflow, so it is the only place a client can find out **before committing buzz** that the model it named isn't the model that will run. It's also the only signal for a **same-price** substitution — cost divergence only helps if you already knew the expected price, which #3665 says explicitly.
- **`generateFromGraph`** both persists the record on the orchestrator workflow's `metadata` *and* attaches it to the reply. Persisting is what makes it readable on every **later** read (poll, reconnect, history) — a reply-only field is gone the moment the client moves on, and nothing retains the submitted body to reconstruct it from. Attaching it to the reply keeps the submit authoritative for the validation it just performed, since the metadata round-trip only happens when the graph produced workflow metadata at all.

## 🔴 The trap, because it would have shipped inert

Persisting the key is **not sufficient on its own**, and the failure is silent. `formatGenerationResponse2` builds `normalizedWfMeta` from an explicit key **allowlist** (`params`, `resources`, `remixOfId`, `isPrivateGeneration`) — not a spread. So the key round-trips through the orchestrator perfectly and the formatter drops it again on read-back. The record would exist, be billed against, and reach nobody.

This wasn't a surprise to the codebase — the App Blocks author had written exactly this on the key's docblock: *"The formatter builds its normalized metadata from a fixed key list, so an extra key is inert there."* It was inert. That comment is now corrected, the formatter passes the key through explicitly, and the new test **drives the real formatter** (the fixture carries no resources, so `getResourceData` is never reached) and fails without the passthrough. Verified by removing it: 2 tests go red.

## Consolidation

`WORKFLOW_METADATA_MODEL_SUBSTITUTIONS_KEY` and its validating reader move from `services/blocks/workflow.service` into the zero-import shared `model-substitution` module. Both paths now share one implementation — a second copy of a validating predicate is how the same bug gets fixed at one site and not the other. `workflow.service` re-exports them so its existing callers are untouched.

The reader validates rather than casts: it crosses a service boundary (the value is whatever the orchestrator hands back) and `reason` is simultaneously a wire contract *and* a bounded prom label, so an unrecognised value must not survive the read. Verified by removing the narrowing — the malformed-entry test goes red.

## Compatibility

Additive throughout; the key is **omitted** when nothing was substituted, so every response is unchanged in the common case.

🔴 One consequence a client must know, stated because it's a real limitation and not a nit: **absence means "no substitution" OR "a server predating this field"** — the two are indistinguishable. A caller that needs to tell them apart should probe a known-bad version id once. I chose omit-when-empty to match the existing App Blocks snapshot contract (#3665 asked for "the same key the App Blocks bridge already populates"); an always-present empty array would have made version detection free at the cost of changing every response shape. Happy to flip it if you'd rather have the latter.

## Verification

`pnpm typecheck` 0 errors. Tests green on this branch **and on the tree it creates when merged with #3673** — both branches touch `orchestration-new.service.ts` and `model-substitution.ts`, so I built the merge and ran the suites there rather than reasoning about it: no textual conflict (`merge-tree` exit 0), edits land in disjoint regions, 689 tests green, 0 type errors on the merged tree.

The two PRs are independent and can merge in either order.

### DIFF
--- a/src/server/routers/blocks.router.ts
+++ b/src/server/routers/blocks.router.ts
@@ -136,6 +136,7 @@ import {
   snapshotFromWorkflow,
   WORKFLOW_METADATA_MODEL_SUBSTITUTIONS_KEY,
 } from '~/server/services/blocks/workflow.service';
+import { projectModelSubstitutions } from '~/shared/data-graph/generation/model-substitution';
 import {
   assertInlineGraphAirsDeclared,
   assertViewerEntitledToInlineResources,
@@ -1090,9 +1091,7 @@ async function createBlockTextToImageStep(opts: {
   // object — so it describes THIS submit and no one else's. Behaviour is
   // unchanged; the caller folds this into the snapshot so the block can detect
   // that it was billed for a model it did not ask for.
-  const modelSubstitutions = externalCtx.modelSubstitutions
-    ?.list()
-    .map(({ requested, applied, reason }) => ({ requested, applied, reason }));
+  const modelSubstitutions = projectModelSubstitutions(externalCtx.modelSubstitutions);
 
   // 🔴 PERSIST it on the workflow's own metadata, not only on the submit reply.
   // A block renders from the TERMINAL POLL (the only snapshot carrying
--- a/src/server/services/blocks/workflow.service.ts
+++ b/src/server/services/blocks/workflow.service.ts
@@ -9,7 +9,10 @@ import { getBaseModelSetType } from '~/shared/constants/generation.constants';
 import { getEcosystem } from '~/shared/constants/basemodel.constants';
 import { isWorkflowAvailable } from '~/shared/data-graph/generation/config/workflows';
 import { resolveVersionWorkflowScope } from '~/shared/data-graph/generation/workflow-capability';
-import { isModelSubstitutionReason } from '~/shared/data-graph/generation/model-substitution';
+import {
+  readModelSubstitutionsFromMetadata,
+  WORKFLOW_METADATA_MODEL_SUBSTITUTIONS_KEY,
+} from '~/shared/data-graph/generation/model-substitution';
 import { getImagesLimit } from '~/shared/data-graph/generation/images-limit';
 import { ModelType } from '~/shared/utils/prisma/enums';
 import type {
@@ -33,8 +36,12 @@ const ORCH_STATUS_MAP: Record<WorkflowStatus, BlockWorkflowSnapshot['status']> =
 };
 
 /**
- * 🔴 KEY under which the submit path stores the request's silent checkpoint
- * substitutions on the ORCHESTRATOR WORKFLOW's `metadata` (issue #3520).
+ * 🔴 The key + reader for silent checkpoint substitutions persisted on the
+ * ORCHESTRATOR WORKFLOW's `metadata` (issue #3520) now live in
+ * `~/shared/data-graph/generation/model-substitution` — the tRPC generation path
+ * needs the identical parse (#3665), and a second copy of a validating predicate
+ * is how the same bug gets fixed at one site and not the other. Re-exported here
+ * because this module is where the block path's callers already look for it.
  *
  * WHY THE ORCHESTRATOR AND NOT A COLUMN. The record is produced during graph
  * validation, which happens once, at submit. But blocks render from the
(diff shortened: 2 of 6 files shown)


ZacxDev:
## Audit round 1 — and the irony was earned

An adversarial audit deleted each of this PR's **four** production edits and ran the suites. **Three survived**:

| deleted | suites |
|---|---|
| the whatIf reply attach — the half #3665 says matters most | green |
| the submit reply attach | green |
| the metadata persist | green |
| the formatter passthrough | **red** (the only one covered) |

So this PR spent a paragraph warning that a field can be plumbed, typechecked, reviewed and reach nobody — while shipping three edits in exactly that state. Fixed in `5230e7e`.

### The test that closes it

`orchestration-new.substitution-replies.test.ts` drives the **real** `generationGraph.safeParse` through the **real** `whatIfFromGraph` / `generateFromGraph`, stubbing only the orchestrator HTTP call and the heavy DB/redis graph. The substitution comes from the actual clamp on a real `modelLocked` ecosystem rather than a hand-fed collector, so a graph change that stops recording fails here too. Each mutant verified dead: whatIf attach (1 test), submit attach (1), persist (2).

Three obstacles, all of which surface as **`no tests`** rather than a failure — worth recording because that reads as "nothing to see" and is how a suite silently stops covering anything:

- **`@civitai/client` can't be loaded under vitest.** Its entry does `export * from './generated'`, a directory specifier Node ESM rejects; `vi.importActual` fails the same way. And vitest validates *every* named import against the mock, including type-only ones esbuild doesn't elide — so the mock has to be exhaustive. It's generated from the tree, with the regeneration command in a comment.
- **A Proxy-based mock hung collection indefinitely** rather than erroring.
- **A `vi.mock` factory referencing a top-level `const`** throws "cannot access before initialization" — hence `vi.hoisted`.

### The field was unreadable from the typed client

`generateFromGraph` returned `cond ? { ...formatted, modelSubstitutions } : formatted`. That infers a **union**, and TS won't let you read a property present on only one arm — `result.modelSubstitutions` fails with TS2339 from the typed tRPC client. The field shipped looking done and was unusable from the web app. Now a conditional spread, matching what `whatIfFromGraph` already did correctly, which infers the optional property.

Doesn't affect the Go CLI (JSON, untyped), which is why it survived my own review.

### Smaller

- **The projection had gone 1 → 3 open-coded sites.** Extracted to `projectModelSubstitutions` — the same argument this PR used to consolidate the reader, which I then failed to apply to the thing sitting next to it. It deliberately does **not** collapse `undefined` and `[]`: `snapshotFromWorkflow` distinguishes them (an explicit `[]` suppresses the fallback metadata read), so collapsing would silently change block poll behaviour.
- **`readModelSubstitutionsFromMetadata(metadata: unknown)`** accepted anything, which made `read…(workflow)` — the exact slip the move from the blocks service invites, and one both call sites had to avoid — a silent no-op returning `undefined` forever. Now typed to the metadata bag, so it's a compile error.
- Documented that the record is readable in **three** places (whatIf → top-level only; submit → top-level *and* `metadata`; later reads → `metadata` only) and that a client reading just one misses cases.

### Verification

`pnpm typecheck` 0 errors; 748 tests green across 26 suites. Merge with #3673 re-verified after these commits: `merge-tree` exit 0, still independent, either order.

Audit also confirmed independently: no import cycle (the shared module is a zero-import leaf), block path behaviour byte-identical, no leak of one request's ids into another's reply, and the orchestrator already accepts this exact metadata key in production via the App Blocks submit path.

ZacxDev:
## Round 2 (delta re-audit) — clean on code, three comments corrected

**No 🔴. Verdict: safe to merge.** The fix round did what it claimed and broke nothing. What it *did* leave behind was three comments asserting behaviour that doesn't exist — corrected in `99b8620`, no production code changed.

### Independently re-derived, not trusted

The audit rebuilt the mutation sweep from scratch and added 12 mutants I hadn't run — inverting each `?.length`, swapping `requested`/`applied`, returning `[]`/`undefined` unconditionally, dropping `reason`, persisting under a different key, sharing one collector array across requests, aliasing the collector's internal array. **15/15 killed**, each by the semantically correct test.

Two of those matter for whether the heavily-mocked test is real: the *formatter-drops-key* and *reader-always-undefined* mutants both die, which proves the real allowlist and the real validating reader execute inside it. Not vacuous.

The fixture is load-bearing too: forcing Qwen to `modelLocked: false` fails 5 of 7 tests including the guard-the-guard, so config drift degrades loudly rather than passing silently.

The union→optional fix was confirmed with a positive control: TS2339 at the old commit, 0 errors at the new one, runtime shape unchanged.

### The three corrections

- **The `[]` vs `undefined` justification was false.** I wrote that `snapshotFromWorkflow` distinguishes them, so collapsing would change block poll behaviour. It doesn't: the fallback is guarded by `extra?.modelSubstitutions?.length`, and `[].length` is `0`, so an empty array takes the *same* branch as `undefined`. Every consumer tests `?.length`. They're still not collapsed — but for the honest reason (a pure extraction shouldn't smuggle in a behaviour change), plus a note on what to check first if anyone wants to.
- **The claimed compile-time guard doesn't exist.** I wrote that typing the parameter makes `read…(workflow)` a compile error. `Workflow` is a type **alias**, so TS grants it an implicit index signature and it's assignable to `Record<string, unknown>` — the slip still compiles. And the suggested tightening (`& { steps?: never }`) would reject the *real* callers, which pass a plain `Record<string, unknown>` whose index signature makes `steps` `unknown`. So the comment now says what the type actually buys — narrower than `unknown`, nothing more — and that the tests are the guard.
- **"A missing mock export does not fail the suite"** — it does fail, non-zero exit. What's uninformative is the `Tests: no tests` line, since a collection error names no symbol.

### Follow-up worth filing separately (pre-existing, not this PR)

The `buildGenerationContext` collector attach (`orchestration-new.service.ts`, from #3535) is **unpinned**: every test hand-builds or mocks `externalCtx`, so deleting that one line makes the whole feature inert across all three tRPC surfaces with nothing failing. Out of scope here — it predates both this PR and #3673 — but it's the seam that everything else in #3520/#3665 hangs off.

### State

`pnpm typecheck` 0 errors. Full `pnpm test:unit`: **12,324 passed, 1 skipped, 0 failed**, zero `no tests` files. Merge with #3673 still clean, either order.

Issue #3681, opened 2026-08-06
civitai-cli OAuth client lacks the AI Services scopes, so `civitai login` cannot generate


ZacxDev:
> [!IMPORTANT]
> **Revised after review — see the comments.** Two things below are superseded:
>
> - The bitmask **`100712449` is wrong**: it omits `BuzzRead` (bit 16), which
>   `civitai generate` needs for its pre-flight balance check. The correct value
>   for the widen-in-place option is **`100777985`**.
> - The proposal did not disclose that widening `civitai-cli` also arms real-Buzz
>   App Blocks `dev:live`, so `--scopes generate` would not be scoped to
>   `civitai generate`. Because of that, the **separate client id** alternative is
>   now the recommended route rather than the fallback.
>
> The original text is left intact below so the review comments still read in order.

## Summary

`civitai login` (the OAuth device flow) produces a token that **cannot generate**. Running `civitai generate` with one fails the scope gate, and the only working credential is a personal API key created by hand on the website.

This is a one-value change: the `civitai-cli` OauthClient's `allowedScopes` does not include the AI Services bits.

## Why it fails

An OAuth access token's ceiling is its client's `allowedScopes | UserRead` — not `Full` — enforced at `apps/auth/src/routes/api/auth/oauth/device/+server.ts:71-72`:

```ts
const requestedScope = rawScope | TokenScope.UserRead;
const allowedScopes  = client.allowedScopes | TokenScope.UserRead;
if (requestedScope > 0 && !hasScope(allowedScopes, requestedScope)) → invalid_scope
```

`civitai-cli` is `100663297` = `UserRead | AppBlocksSubmit | AppBlocksDevTunnel` (bits 0, 25, 26), set by the raw-SQL migration `20260619120000_register_civitai_cli_oauth_client`. Bits 14 and 15 are absent, so the CLI cannot even *request* them — and `orchestrator.generateFromGraph` requires `AIServicesWrite`, `orchestrator.whatIfFromGraph` requires `AIServicesRead`.

## Ask

Add the two AI Services bits to that client:

```
current   100663297  = UserRead | AppBlocksSubmit | AppBlocksDevTunnel
add       1<<14 (16384)  AIServicesRead    "View generation & training history"
          1<<15 (32768)  AIServicesWrite   "Generate, train & scan"
proposed  100712449
```

Consent labels for both bits already exist (`packages/civitai-auth/src/token-scope.ts:124-125`), so the device-approval screen renders them without further work.

[...]

PR #3699, opened 2026-08-06
fix(oauth): grant the civitai-cli client the AI-services scopes so `civitai generate` works (#3681)

ZacxDev:
Fixes the `civitai login` → `civitai generate` failure reported in #3681, and turns the `getMyAppAnalytics` scope invariant from an inspection-based claim into an asserted one.

Ref: https://github.com/civitai/civitai/issues/3681#issuecomment-5208310465

## Root cause

An OAuth access token's ceiling is its client's `allowedScopes | UserRead`. That ceiling is enforced twice — once when the device authorization is created (`apps/auth/src/routes/api/auth/oauth/device/+server.ts`) and again when the token is minted (`.../device-token/+server.ts`).

The `civitai-cli` OAuth client's `allowedScopes` is `100663297` = `UserRead | AppBlocksSubmit | AppBlocksDevTunnel` (bits 0, 25, 26). Bits 14/15/16 are absent, so the CLI cannot even *request* them, and every procedure on the generate path rejects the resulting token:

| procedure | `requiredScope` |
| --- | --- |
| `orchestrator.whatIfFromGraph` | `AIServicesRead` |
| `orchestrator.generateFromGraph` | `AIServicesWrite` |
| `orchestrator.getWorkflow` | `AIServicesRead` |
| `orchestrator.queryGeneratedImages` | `AIServicesRead` |
| `orchestrator.cancelWorkflow` | `AIServicesWrite` |
| `buzz.getBuzzAccount` | `BuzzRead` |

## The change

One new migration widening that single client row:

```
AIServicesRead  = 1 << 14 = 16384
AIServicesWrite = 1 << 15 = 32768
BuzzRead        = 1 << 16 = 65536
                            114688

100663297 | 114688 = 100777985
```

`BuzzRead` is included for `buzz.getBuzzAccount`, which the CLI reads for its pre-flight insufficient-balance guard. That guard **fails soft** (it warns and continues), so omitting the bit would degrade UX rather than break generation — it is in the grant so the pre-flight check works instead of warning on every run.

## 🔴 Deploy ordering

**This migration must be applied to production BEFORE any CLI release that requests these bits.** The device-authorization scope check is all-or-nothing: it *rejects* a request carrying any bit outside the client ceiling rather than clamping it, so a CLI that asks first gets `invalid_scope` and `civitai login` 400s outright. That ordering is now asserted, not just described — see the before/after pair in the new device-flow test.

**The migration is MANUAL-APPLY and this PR does not apply it anywhere.** Per the repo convention, migrations under `packages/civitai-db-schema/prisma/migrations/` are committed for history and applied per environment by a human. No database was touched by this PR.

## `getMyAppAnalytics` — the non-superset argument

`enforceTokenScope` bypasses the scope gate on **exact equality** with `Full`, so a mask that is a *strict superset* of `Full` while missing an opt-in bit would flip `blocks.router.getMyAppAnalytics` from allow to deny. The comment on that procedure documented the old value and noted the argument rested on inspection of the migration with no test behind it.

`100777985` carries bits 0, 14, 15, 16, 25, 26 — and **none of 1–13 or 17–24** — so it is not a superset of `Full` (33554431) and the flip stays unreachable. The comment is updated to the new value and reasoning.

More importantly, the inspection-only part is now covered: `src/server/services/oauth/__tests__/oauth-client-scope-grants.test.ts` enumerates every migration writing `"OauthClient"."allowedScopes"` off the tree, reconciles them against a declared table, folds the grants per client, and fails if any client would hold a strict superset of `Full`. A future migration granting some client `Full | <opt-in bit>` now goes red there instead of silently flipping the gate.

## Tests

| file | what it covers |
| --- | --- |
| `src/server/services/oauth/__tests__/oauth-client-scope-grants.test.ts` | (new) the migration population guard above + pins the folded `civitai-cli` value at `100777985` |
| `src/server/routers/__tests__/cli-generate-scope-surface.test.ts` | (new) TypeScript-AST guard: the union of `requiredScope` across the six generate-path procedures is exactly `AIServicesRead\|AIServicesWrite\|BuzzRead` and is covered by the grant. Goes red if any of them gains a wider scope **or loses its `.meta({ requiredScope })`** — an un-annotated procedure implicitly requires `Full` and would 403 every scoped token |
| `apps/auth/src/routes/api/auth/oauth/device/__tests__/server.test.ts` | (new) device-flow scope validation: subset accepted, out-of-ceiling bit rejected with `invalid_scope`, reject-not-clamp, `UserRead` baseline OR-ed in on both sides, `ALL_SCOPES` bound, and the #3681 before/after pair |
| `apps/auth/src/lib/server/oauth/__tests__/scope.test.ts` | (new) `scopeLabels` for bits 14/15/16 — the strings a user sees on the consent screen before granting Buzz-spend authority — plus `hasScope` / `stringToScope` |
| `src/server/routers/__tests__/blocks.router.getMyAppAnalytics.test.ts` | (comment only) the "held only by inspection" note now points at the new guard |

**Red-at-base / green-at-HEAD.** Only `oauth-client-scope-grants.test.ts` is regression coverage for this change: with the new migration removed it fails 3 of 5 (`enumerated migration set…`, `every declared grant matches…`, `the grant-shape matcher…`), and passes 5/5 with it. The other three new files pass on pre-change code — they are **invariant guards** against future regressions, not coverage of this bug, and are labelled as such rather than counted as regression tests.

Each guard was mutation-checked: dropping a `.meta({ requiredScope })`, widening one to an ungranted bit, and renaming a procedure each kill the AST guard with its own assertion; adding an undeclared migration, declaring one that grants `Full | <opt-in bit>`, and moving the literal out of the `SET` clause each kill the migration guard with its own assertion; turning the device endpoint's reject into a clamp and removing either `UserRead` OR kill the device-flow tests with theirs.

## Follow-up (not in this PR)

A later change will make `ai:write:budgeted` an **explicit** request in `clampDevScopes` rather than inferring it from the bearer's `AIServicesWrite` bit. Doing it now would be a breaking change for existing personal-API-key users of `civitai app dev-token`, so it is deliberately out of scope here.

### DIFF
--- a/apps/auth/src/lib/server/oauth/__tests__/scope.test.ts
+++ b/apps/auth/src/lib/server/oauth/__tests__/scope.test.ts
@@ -1,0 +1,95 @@
+import { describe, it, expect } from 'vitest';
+import { TokenScope, ALL_SCOPES } from '@civitai/auth/token-scope';
+
+import { hasScope, scopeToString, stringToScope, scopeLabels } from '../scope';
+
+/**
+ * `apps/auth/src/lib/server/oauth/scope.ts` — the hub's only scope encode/decode +
+ * ceiling-test helpers, and the source of the CONSENT SCREEN's permission list.
+ *
+ * The label half is the user-visible half: `scopeLabels` is what the device-verify
+ * and authorize screens render before someone grants a client authority over their
+ * account. Widening the `civitai-cli` client to AI-services scopes (issue #3681)
+ * adds Buzz-SPENDING authority — `AIServicesWrite` implicitly authorises orchestrator
+ * buzz spend — so the strings a user actually sees for bits 14/15/16 are pinned here
+ * as literals rather than derived from `tokenScopeLabels` (deriving the expectation
+ * from the implementation would leave a silently-blank label green).
+ */
+
+/** The `civitai-cli` client's `allowedScopes` after the AI-services widening. */
+const CLI_ALLOWED_AFTER = 100777985;
+
+describe('scopeLabels', () => {
+  it('renders the AI-services and Buzz labels for bits 14/15/16', () => {
+    expect(scopeLabels(TokenScope.AIServicesRead)).toEqual(['View generation & training history']);
+    expect(scopeLabels(TokenScope.AIServicesWrite)).toEqual(['Generate, train & scan']);
+    expect(scopeLabels(TokenScope.BuzzRead)).toEqual(['View buzz balance & history']);
+  });
+
+  it('renders every label for the widened civitai-cli consent screen, in bit order', () => {
+    expect(scopeLabels(CLI_ALLOWED_AFTER)).toEqual([
+      'Read profile, settings & email',
+      'View generation & training history',
+      'Generate, train & scan',
+      'View buzz balance & history',
+      'Submit Apps for review',
+      'Open on-site dev tunnels',
+    ]);
+  });
+
+  it('does NOT render those labels for the pre-fix mask (negative control)', () => {
+    // Without this, a `scopeLabels` that returned every label unconditionally would
+    // satisfy the assertions above.
+    const before = 100663297; // UserRead|AppBlocksSubmit|AppBlocksDevTunnel
+    expect(scopeLabels(before)).toEqual([
+      'Read profile, settings & email',
+      'Submit Apps for review',
+      'Open on-site dev tunnels',
+    ]);
+  });
+
+  it('renders nothing for the empty mask', () => {
+    expect(scopeLabels(0)).toEqual([]);
+  });
+
+  it('never renders a blank string for a bit it claims to cover', () => {
+    // A missing `tokenScopeLabels` entry would show the user an empty bullet on the
+    // consent screen rather than failing loudly.
+    for (const label of scopeLabels(ALL_SCOPES)) {
+      expect(label.length).toBeGreaterThan(0);
+    }
+    // Positive control on the loop above — it must have iterated.
+    expect(scopeLabels(ALL_SCOPES).length).toBeGreaterThanOrEqual(27);
+  });
+});
+
+describe('hasScope', () => {
+  it('is a SUBSET test against the ceiling, not an intersection test', () => {
+    const ceiling = CLI_ALLOWED_AFTER | TokenScope.UserRead;
+    expect(hasScope(ceiling, TokenScope.AIServicesWrite)).toBe(true);
+    expect(hasScope(ceiling, TokenScope.AIServicesRead | TokenScope.BuzzRead)).toBe(true);
+    // Partially-overlapping request: one granted bit + one ungranted bit must be
+    // REJECTED as a whole. This is the all-or-nothing property the device flow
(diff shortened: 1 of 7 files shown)


ZacxDev:
Adversarial audit of this PR turned up three things the description gets wrong or understates. Correcting them here rather than editing the body, since anyone who has already read it saw the earlier version.

Test-guard weaknesses the audit also found are being fixed in a follow-up commit on this branch; this comment is only about claims in the description.

## 1. Merging this does not close #3681

The description implies the fix lands on merge. It does not. Applying the migration changes **nothing observable**:

- No already-issued token gains scope — the refresh path re-derives a token's scope from the stored `token.scope`, not from `client.allowedScopes`, so existing CLI tokens stay at `100663297` until re-login.
- No newly-issued token requests the bits either, because the CLI hardcodes `DeviceScope = "100663297"` on `main`.

So the sequence is: apply the migration → release a CLI that can request the wider scope → **each user re-runs `civitai login`**. Until all three happen, `civitai generate` still 403s. This PR is necessary but not sufficient, and #3681 should stay open on merge.

## 2. The grant is 75 procedures, not the 6 in the table

The description lists the six procedures on the generate path, which reads as though the grant is scoped to them. It isn't — these are account-level scope bits. Measured across `src/server/routers/**` (excluding tests):

| bit | procedures gated |
|---|---|
| `AIServicesRead` | **34** |
| `AIServicesWrite` | **27** — orchestrator 9, comics 8, training 6, generation-preset 4 |
| `BuzzRead` | **14** |

A token carrying these bits can submit **training jobs** and **comics generation**, both Buzz-spending, not just the six calls listed. That is inherent to the scope granularity rather than a defect in this change, but the description should not present the grant as narrower than it is.

## 3. The `dev:live` consequence is a spend-boundary change, not an ergonomics one

The body mentions this as a follow-up bullet. It deserves sharper framing: `dev-token.ts` derives `keyCanSpend` from the bearer's `AIServicesWrite` bit, and `clampDevScopes` strips `ai:write:budgeted` when that is false. Today an OAuth-login-minted App Block dev token is read/estimate-only, and the CLI documents that as the reason real-Buzz `dev:live` requires a personal API key.

Once a login token can carry `AIServicesWrite`, that stripping stops — App Block dev tokens minted over `civitai login` gain real Buzz spend with no change to the `dev-token` command itself. Tracked as #3703, which also explains why the fix needs a wire-format change rather than a default flip (the `scopes` field narrows by intersection, so "everything except `ai:write:budgeted`" is not expressible).

## Correction to the audit itself

For the record, since it informed the above: the audit inferred that once the CLI bumps its scope constant, *every* `civitai login` token would carry Buzz-spend authority — reversing the design decision the CLI's own source documents.

That is **not** what is planned, and the audit could not see it because the change lives in another repo. civitai/cli#220 keeps the default login at `100663297` unchanged and adds the wider mask only behind an explicit opt-in (`civitai login --scopes generate`), with the additive property pinned by a test. A user who only runs `app submit` keeps a token with no spend authority.

PR #3702, opened 2026-08-06
fix(notifications): notify bounty entry owners of comments

JustMaier:
Found while fixing #3691 — `bountyEntry` had the identical gap posts did.

## The gap

`bountyEntry` has a `threadUrlMap` entry (`comment.notifications.ts`) but no owner processor. Nothing joined `Thread."bountyEntryId"`, so a comment on your bounty entry notified you only if you were *already* in the thread, via `new-thread-response`. `bounty.notifications.ts` covers award / reaction-milestone / entry-created only.

Measured on prod: **21,812** historical entry comments across **3,876** distinct owners, ~26/day currently.

## The fix

`new-bounty-entry-comment`, mirroring `new-bounty-comment`:

- joins `Thread."bountyEntryId"` → `BountyEntry."userId"`, plus `Bounty` for the title (BountyEntry has no title of its own, so the message names the parent bounty for context)
- `BountyEntry."userId"` is **nullable** (`SetNull` when a user is deleted). `be."userId" > 0` drops those orphans as well as system-owned entries, since `NULL > 0` is `NULL`
- `EntityOwner` priority + the shared `comment:v2:` dedupe key, so a mention or direct reply on the same comment still outranks it
- floored the same way as #3691 (launch date + rolling 7-day window) — a new type has no `last-sent-notification-*` cursor row until its first successful run, so it inherits the job's **global** last-run, which is `new Date(0)` on a fresh DB and stale by the outage duration after any send-notifications outage. Those 21,812 rows are what sits behind that.

## Also: `/bounties/entries/[entryId]` was dropping `highlight`

That route is a `getServerSideProps` redirect to the canonical `/bounties/{bountyId}/entries/{entryId}`, and a `destination` string carries no query — so the query was silently discarded.

This is not hypothetical for this feature: **every existing** bountyEntry comment notification (`new-comment-reply`, `new-thread-response`, `new-mention`) builds its URL through `threadUrlMap`, which emits `/bounties/entries/{id}?highlight=…`. All of them have been landing readers at the top of the entry instead of on the comment. The redirect now forwards the query.

The new processor sidesteps the hop entirely and links to the canonical URL directly — it has `bountyId` from the join, which `threadUrlMap` does not (it only gets `threadParentId`).

Verified the target page honors it: `BountyEntryDiscussion` mounts `RootThreadProvider entityType="bountyEntry"`, which is the same `CommentsProvider` that reads `router.query.highlight`, forwards it as `targetCommentId`, and prepends the target to page 1 so it renders even past the cursor.

## Testing

`src/server/notifications/__tests__/comment.bounty-entry-owner.test.ts` (new), plus `new-bounty-entry-comment` wired into the `comment.dedupe-key.test.ts` matrix (v2 dedupe key, commentId-only derivation, declares a priority, renders a message + URL, distinct `key` namespace).

Every assertion was **verified by mutation** — each was confirmed to go red when the corresponding thing is broken in the source:

| Mutation | Result |
|---|---|
| `be."userId" > 0` → `IS NOT NULL` (orphan guard) | red |
| `'bountyId', b.id` → `be.id` (URL points at the wrong entity) | red |
| `priority: EntityOwner` → `Mention` (would outrank new-mention for the shared key) | red |
| URL → the query-dropping redirect route | red |

`pnpm run typecheck` 0 errors; eslint clean on touched files (1 pre-existing warning); 124 notification tests pass.

### DIFF
--- a/src/pages/bounties/entries/[entryId].tsx
+++ b/src/pages/bounties/entries/[entryId].tsx
@@ -1,6 +1,7 @@
 import { dbRead } from '~/server/db/client';
 import { createServerSideProps } from '~/server/utils/server-side-helpers';
 import { PageLoader } from '~/components/PageLoader/PageLoader';
+import { forwardQuery } from '~/utils/forward-query';
 
 export const getServerSideProps = createServerSideProps({
   useSSG: true,
@@ -19,22 +20,12 @@ export const getServerSideProps = createServerSideProps({
     // threadUrlMap builds `/bounties/entries/{id}?highlight=…` for every bountyEntry comment
     // notification, and a destination without the query silently drops it — so the comment the
     // notification is about never gets highlighted.
-    const { entryId: _, ...query } = ctx.query;
-    const queryString = new URLSearchParams(
-      Object.entries(query).flatMap(([key, value]) =>
-        value === undefined
-          ? []
-          : Array.isArray(value)
-          ? value.map((v) => [key, v])
-          : [[key, value]]
-      ) as [string, string][]
-    ).toString();
-
     return {
       redirect: {
-        destination: `/bounties/${bountyEntry.bountyId}/entries/${entryId}${
-          queryString ? `?${queryString}` : ''
-        }`,
+        destination: `/bounties/${bountyEntry.bountyId}/entries/${entryId}${forwardQuery(
+          ctx.query,
+          ['entryId']
+        )}`,
         permanent: false,
       },
     };
--- a/src/pages/bounties/entries/[entryId].tsx
+++ b/src/pages/bounties/entries/[entryId].tsx
@@ -16,9 +16,25 @@ export const getServerSideProps = createServerSideProps({
       return { notFound: true };
     }
 
+    // threadUrlMap builds `/bounties/entries/{id}?highlight=…` for every bountyEntry comment
+    // notification, and a destination without the query silently drops it — so the comment the
+    // notification is about never gets highlighted.
+    const { entryId: _, ...query } = ctx.query;
+    const queryString = new URLSearchParams(
+      Object.entries(query).flatMap(([key, value]) =>
+        value === undefined
+          ? []
+          : Array.isArray(value)
+          ? value.map((v) => [key, v])
+          : [[key, value]]
+      ) as [string, string][]
+    ).toString();
+
     return {
       redirect: {
-        destination: `/bounties/${bountyEntry.bountyId}/entries/${entryId}`,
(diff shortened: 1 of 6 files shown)


JustMaier:
## Adversarial review pass — pushed `e0188ea`

**The sharp finding: `details.version` was unguarded, and it's load-bearing.**

`detail-fetchers/comment.detail-fetcher.ts` registers itself for **every** key in `commentNotifications` and routes anything with `version !== 2` to the **legacy `Comment` table**. Those ids overlap CommentV2's — the repo's own dedupe test asserts exactly that. So deleting `'version', 2` from the `JSONB_BUILD_OBJECT` doesn't fail; it silently attaches a *different comment's* body, avatar and username to the notification.

Nothing guarded it on any type, so I asserted it across the whole V2 family rather than just this one.

**Every `toContain('JOIN …')` is also a substring of `LEFT JOIN …`.** Widening the `User` join renders `null commented on your entry…`; widening `Bounty` produces `/bounties/null/entries/9`. Both passed the existing assertions. Pinned with `not.toContain('LEFT JOIN')`.

**The redirect had no test at all** — the only file here with real runtime branching. Extracted `forwardQuery` and covered: empty query (destination byte-identical to before), array-valued params, the route param dropped even when a caller also supplies `?entryId=`, and that a crafted param can't escape the path or reach the Location header.

Security on the redirect came back **clean**, verified empirically rather than by inspection: `URLSearchParams.toString()` percent-encodes `#`, `@`, `/`, `\`, `=`, `&` and CRLF (`?a=b#@evil.com` → `a=b%23%40evil.com`; `a\r\nX: 1` → `a%0D%0AX%3A+1`). The destination is always relative and rooted at a numeric `bountyId` from the DB — no `//` prefix, no open redirect, no header injection. Re-encoding is correct, not double-encoding: Next hands back decoded values.

**Also confirmed clean:** replies are correctly excluded (a reply's thread is keyed on `commentId`, so `bountyEntryId` is NULL); `Thread.bountyEntryId` is `@unique` so no fan-out; `be."userId" > 0` does drop NULLs, so `ownerId` can never be NULL; `JOIN "Bounty"` is correctly INNER since `bountyId` is NOT NULL; the key prefix doesn't collide with `new-comment-bounty:` (`-` vs `:` at position 18), and that *is* already guarded.

Both new SQL guards mutation-verified.

Also corrected two comments this PR's own diff falsified — they justified the canonical URL by saying the redirect drops the query, which stopped being true in the same commit. The real reason stands alone: `threadUrlMap` is handed the entry id with no `bountyId`, so it can't build that URL.

PR #3706, opened 2026-08-06
fix(notifications): make settings polarity explicit; stop bulk-off subscribing users to shop promos

JustMaier:
Follow-up to #3691. What started as "a flag looks inverted" turned out to be three defects sharing one root cause.

## Root cause

`defaultDisabled` was a **presentation-only** flag — read in exactly three UI spots (`utils.notifications.ts`, `NotificationsCard.tsx`, `NotificationToggle.tsx`) and by **no server code at all**. Nothing checked it against what a processor's SQL actually does.

A `UserNotificationSettings` row means **opted-OUT** for 30 of the 31 SQL processors, for the entire event-driven path (`apps/notifications/src/lib/server/create.ts` filters recipients by it, type-agnostically), and per `docs/features/notifications.md`. The single exception is `cosmetic-shop-item-added-to-section`, which derives recipients by `JOIN`ing that table — there a row means **SUBSCRIBED**, and a genuine opt-in bell at `/shop` (`src/pages/shop/index.tsx`) writes it.

One table, two meanings, and no guard that a type's flag matched its query.

## Defect 1 — three types ship ON while rendering OFF

`new-3d-model-comment`, `-response`, `-nested` set `defaultDisabled: true` but run the ordinary `NOT EXISTS` opt-out query. So they send by default while the checkbox shows unchecked, and ticking it to "enable" inserts the row and **silences** them. The flag was simply wrong there — removed.

## Defect 2 — "turn off all notifications" subscribed users to shop promos

`toggleAll`/`toggleCategory` pass a raw type list; the schema is `z.string().array()` with no allowlist; `toggle: false` INSERTs. For the opt-in type that row means *subscribed*. The type declares no `toggleable` field, so the `if (toggleable === false && !showCategory) continue` guard never excluded it.

And the resulting state **hides its own evidence**: `hasNotifications` was computed with no polarity awareness, so after a bulk-off it reads `false`, the card renders *"All non-essential notifications are turned off"*, and the whole category tree — including the now-checked shop checkbox — is unreachable. The only way back was the bell on `/shop`.

Measured on prod, 5,000-row random sample:

| Test | Result |
|---|---|
| shop rows adjacent in the id sequence to another row of the **same user** (⇒ one multi-row INSERT) | **95.96%** |
| holders carrying 31+ rows (of 72 types ever toggled) | 92.9% |
| holders who **also muted `civitai-features`** | 96.3% |
| holders whose shop row is their *only* row (deliberate opt-in floor) | 9,474 (3.1%) |

Neighbours are the same user's *other types*, not other users' shop rows — so it's a per-user bulk toggle, not a backfill.

**This PR does not touch existing rows.** Whether to unsubscribe the collateral is a product/marketing call, and I'd rather hand over the numbers than make it in a bugfix. Happy to prepare the SQL.

## Defect 3 — the toggle only worked by double inversion

`NotificationToggle` computed an inverted `isEnabled` and passed it straight through as `toggle`, which happened to cancel out. The optimistic update in `NotificationsCard` assumed a single write direction for a mutation whose direction is now per-type.

## The fix

- `defaultDisabled` → **`optIn`**, read by the **toggle handler**, not just the UI, so a row can never be written with the wrong polarity. The handler splits the incoming type list and writes each half in its own direction.
- Opt-in types are excluded from `notificationTypes` (the bulk list) and filtered out of `toggleCategory` — they still render their own checkbox, they're just never swept up by a switch that can't express their meaning.
- They're also excluded from the `hasNotifications`/`hasCategory` aggregates: they aren't part of the baseline every user starts with, and those aggregates decide whether the tree renders at all.
- `useNotificationSettings` resolves polarity **once**, so no render path inverts anything.

## Testing

New `notification-settings-polarity.test.ts` pins the contract in both directions — an `optIn` processor must derive recipients by joining and must **not** carry the opt-out clause, and every other processor that touches the table must gate on `NOT EXISTS`. That's the check whose absence let all three defects through. Plus: bulk lists can't reach an opt-in type, the type still renders, the handler splits by polarity, no render path re-inverts, and a vacuity guard so none of it passes if `optIn` is dropped everywhere.

`pnpm run typecheck` 0 errors; eslint clean on all touched files; 120 notification tests pass.

No migration. No schema change. No change to who currently receives anything.

### DIFF
--- a/src/components/Account/NotificationsCard.tsx
+++ b/src/components/Account/NotificationsCard.tsx
@@ -5,6 +5,7 @@ import { NewsletterToggle } from '~/components/Account/NewsletterToggle';
 import { useNotificationSettings } from '~/components/Notifications/useNotificationSettings';
 import { SkeletonSwitch } from '~/components/SkeletonSwitch/SkeletonSwitch';
 import {
+  isOptInNotification,
   notificationCategoryTypes,
   notificationTypes,
 } from '~/server/notifications/utils.notifications';
@@ -23,16 +24,25 @@ export default function NotificationsCard() {
       await queryUtils.user.getNotificationSettings.cancel();
 
       const prevUserSettings = queryUtils.user.getNotificationSettings.getData() ?? [];
-      const currentlyDisabled = prevUserSettings.map((x) => x.type);
+      const withRow = prevUserSettings.map((x) => x.type);
       const latestSetting =
         prevUserSettings.length > 0 ? prevUserSettings[prevUserSettings.length - 1] : { id: 0 };
-      const newSettings = type
-        .filter((t) => !currentlyDisabled.includes(t))
+
+      // Mirror the split the toggle handler makes: for an opt-in type the row means subscribed, so
+      // the same `toggle` adds a row where it removes one for everything else. Guessing a single
+      // direction here made the checkbox flicker to the wrong state until the refetch landed.
+      const removing = type.filter((t) =>
+        toggle ? !isOptInNotification(t) : isOptInNotification(t)
+      );
+      const adding = type
+        .filter((t) => (toggle ? isOptInNotification(t) : !isOptInNotification(t)))
+        .filter((t) => !withRow.includes(t))
         .map((t) => ({ ...latestSetting, type: t, disabledAt: new Date() }));
 
-      queryUtils.user.getNotificationSettings.setData(undefined, (old = []) =>
-        toggle ? old?.filter((setting) => !type.includes(setting.type)) : [...old, ...newSettings]
-      );
+      queryUtils.user.getNotificationSettings.setData(undefined, (old = []) => [
+        ...old.filter((setting) => !removing.includes(setting.type)),
+        ...adding,
+      ]);
 
       return { prevUserSettings };
     },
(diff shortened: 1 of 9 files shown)


JustMaier:
## Adversarial review pass — pushed `4d14f72`

Two HIGH findings, both mine, both real.

**The /shop bell was half-converted.** I flipped `toggle: isEnabled` → `!isEnabled` but left `NotificationToggle`'s own hand-rolled `onMutate` assuming the old polarity. Result: server wrote correctly, cache patch no-opped in *both* directions, bell rendered the stale state until an unrelated refetch. It only worked on main because two inversions cancelled out, and I removed one of them.

Root cause was duplication — both callers hand-rolled the same optimistic update. Extracted `useToggleNotificationSetting` so they can't drift again, rather than patching one copy.

**Excluding opt-in types from bulk-off stranded the users this PR is about.** Opt-in types don't feed `hasNotifications`, so once everything is off the category tree stops rendering and the only in-settings control goes with it — and I'd also removed `toggleAll(true)`'s incidental DELETE as a recovery path. Now asymmetric on purpose: turning everything **off** unsubscribes opt-in types; turning everything **on** does not subscribe, because nobody reads "enable notifications" as "sign me up for shop promos". The hook comment claimed a recovery path that didn't exist; it does now.

**The contract test was weaker than it looked.** `prepareQuery` may be async — all five bounty processors, all four featured, one article — and my sync-only reader `continue`d past every one. So `optIn` on an async opt-out processor, the precise defect this file exists to prevent, passed green. Two more holes: `JOIN "UserNotificationSettings"` as a substring accepts `LEFT JOIN` (which restricts nothing), and an `optIn` type whose query never references the table was skipped entirely — the worst config, since an event-driven opt-in type routes through `create.ts` where a row means *suppress*, making subscribers exactly the set that gets dropped.

Verified by mutation, each red under the mutation and green restored:

| Mutation | Caught |
|---|---|
| `optIn` via `LEFT JOIN` (fans out to everyone) | ✅ |
| `optIn` on an **async** opt-out processor | ✅ |
| bell sends `isEnabled` instead of `!isEnabled` | ✅ — **survived until this commit**, added a guard |

**Confirmed clean by the reviewer:** the controller truth table is byte-identical to main for opt-out types in all four cells (`{...input, type: toRemove}` — override follows the spread, and both services destructure only `{type, userId}`); `notificationTypes` has exactly one consumer repo-wide; `create.ts` is the only other reader of the table; `removing`/`adding` are provably disjoint. Bonus: an empty `type` array used to throw in `Prisma.join([])` and now short-circuits.

**Not addressed here:** ~286k users hold a shop-promo row written by main's bulk-off. Cleanup is a manual DELETE and a product call, not a code fix.

PR #3711, opened 2026-08-06
docs: correct the NixOS component-test diagnosis (and revert the playwright bump that broke 59 e2e specs)

ZacxDev:
## What this is now

**Documentation + environment-diagnosis only. No dependency change.** `package.json` and `pnpm-lock.yaml` are byte-identical to this branch's merge-base — `playwright` / `@playwright/test` stay at **`^1.57.0`**.

An earlier revision of this PR bumped them to an exact `1.61.1`. That bump has been **reverted** (`e2091b55be`). What is left is three corrections, each independent of the bump and each fixing something that was factually wrong.

## Why the bump was reverted

**Not because it failed — because it is not self-contained.** A correction is owed here: an earlier version of this PR body and the first revert commit said the bump "broke 59 e2e specs at runtime". **That was wrong**, and the control it rested on did not control.

What actually happened, read from the `run-smoke` log of pipelinerun `pr-preview-3711-hpb9l` rather than inferred from the red status:

| | |
|---|---|
| `browserType.launch: Executable doesn't exist at /ms-playwright/chromium_headless_shell-1228/...` | **177** occurrences (59 failed × 3 retries) |
| assertion / `expect` / timeout / `AssertionError` failures | **0** |
| distinct `Error:` prefixes in the whole log | **1** |

**Not one spec executed.** The 59 failures are the *same version-skew bug this PR documents*, relocated into CI: the smoke step runs the **workspace-local** `./node_modules/.bin/playwright` (1.61.1 after the bump) against browsers baked into a container image still pinned to the **1.57** line, so it went looking for chromium-1228 inside an image containing 1200.

The comparison that made this look like a real regression — "#3707 passed smoke on the same infrastructure in the same window, so the environment was healthy" — **cannot detect this**. #3707 did not carry the bump, so its workspace `playwright` still matched the image. It varies a different dimension than the one under test, so its green says nothing about the bumped tree.

So the honest reason to revert is the one that survives:

> A bump here is **not self-contained**. `PLAYWRIGHT_BROWSERS_PATH` in that job points *inside* a version-matched container image, so the repo pin and the image tag are a lockstep pair owned by two different repos and two different people. Changing one requires changing the other in the same window, in **both** directions, forever.

Against that, the defect being solved is "one workstation cannot supply a matching browser bundle". That is fixable on that workstation, at no cost to anybody else. Changing a shared production repo's test dependency to satisfy one host is simply the wider blast radius — and that argument never needed the bump to be broken.

For the record, under the bump the local suite did work: **130 passed (130)** files, **1304 passed (1304)** tests, exit 0.

## The actual defect, and where it belongs

Playwright pins **one exact Chromium build per release** and resolves it *by revision* under `PLAYWRIGHT_BROWSERS_PATH`:

| | wants chromium revision |
|---|---|
| `playwright-core@1.57.0` (what this repo resolves) | **1200** |
| `playwright-core@1.61.1` | **1228** |

A NixOS host that exports a nixpkgs `playwright-driver.browsers` bundle exports exactly **one** version of it. If that bundle is 1.61.1 and the repo is 1.57.0, the launch fails:

```
browserType.launch: Executable doesn't exist at
  $PLAYWRIGHT_BROWSERS_PATH/chromium_headless_shell-1200/chrome-headless-shell/...
```

and the whole project reports:

```
 Test Files  (130)
      Tests  no tests
   Duration  180ms
```

130 files collected, **zero executed**, exit 1 — a shape indistinguishable from a suite that genuinely has nothing to run. That failure mode is what made this expensive, and it is the thing worth writing down.

But it is a statement about **one host**, not about this repo. Nixpkgs carries one `playwright-driver` version per revision, so a machine that drives several repos on different Playwright lines needs one pinned bundle per line — a host-side problem with a host-side fix. Changing a shared production repo's test dependency to satisfy one workstation is the wider blast radius.

**Fixed on the host instead** (outside this repo): the environment now exposes a registry of `playwright-driver` versions and selects the one matching whatever Playwright a given project has installed, asserting the chromium revision actually matches before running. Verified on the **unbumped** tree in this PR:

```
 Test Files  130 passed (130)
      Tests  1304 passed (1304)
   Duration  81.33s                     exit 0
```

Identical counts to the bumped run — with no dependency change. A second repo pinned to `^1.61.1` still passes on its own bundle in the same session (29/29 browser checks), so the fix is per-project, not a global swap.

## What this PR still changes

**`CLAUDE.md`** — the diagnosis was wrong. It said:

> Browser/component tests on NixOS need help — **there is no `chromium` on `PATH`**.

That is false; the host has Chromium. The section then offered two host-local workarounds with the note that "which one works has varied between attempts", one of which "hung for ~18 minutes". Anyone following it was chasing symptoms of a version-skew bug. It now describes the real 1:1 revision mechanism, says to match the host bundle to the pin rather than the reverse, gives the two numbers to compare before running, and records the CI lockstep coupling (with the 59-failure log evidence, correctly attributed) so the bump is not retried by the next person who hits the 0-of-130 symptom.

**`vitest.config.mts`** — carried the same wrong diagnosis in a comment at the point of use. Corrected, and it now says explicitly that a mismatch collects every file and executes none.

**`CONTRIBUTING.md`** — said the component suite is "~860 component tests". It is **1304**.

---

# Separate finding (NOT fixed here): the Typecheck gate is blind to `src/**/__tests__/**`

`tsconfig.json` excludes `src/**/__tests__/**`, so `pnpm typecheck` — the pre-push hook, the CI Typecheck job, and the local gate — structurally cannot see type errors in that tree.

**Measurement.** A temporary `tsconfig.tests-audit.json` extending `tsconfig.json` with only that one exclude entry removed (identical `include`, its own `tsBuildInfoFile`), run through the repo's own heap-managed wrapper:

```
node scripts/typecheck.mjs -p tsconfig.tests-audit.json
```

**673 type errors, across 123 files** (of 753 files in the excluded tree — so 630 files are already clean). Wrapper verdict `673 type error(s)`, exit 2. Every one of the 673 is inside a `__tests__/` path: adding the tree to the program introduced **zero** new errors anywhere else.

Baseline for comparison: `pnpm run typecheck` at the same commit → `OK — 0 type errors in 62s`.

**Instrument validated — the control pair, from a single run.** Two files with a byte-identical planted error (`const x: number = 'a string'`), one inside the excluded tree and one outside:

| planted in | `pnpm run typecheck` |
|---|---|
| `src/server/zz-gate-control-visible.ts` (included) | **reported** — `error TS2322` |
| `src/server/__tests__/zz-gate-control.test.ts` (excluded) | **not reported** |

The gate printed `1 type error(s)`, not 2. So it can go red (negative control passes) *and* it is blind to exactly this tree (the claim under test). Both control files removed afterwards; nothing in this PR.

**This also resolves an apparent contradiction** between two earlier reports — "~671 pre-existing errors" and "both suites 100% green". Both are true and not in tension: the suites *execute* fine (Vitest transpiles per-file and does not typecheck), while `tsc` over the same files finds 673 errors. Different instruments, different questions. Measured side by side at this commit:

- `pnpm run test:unit:run` → **849 files passed (849), 12697 passed | 1 skipped (12698)**, exit 0
- `pnpm run test:component` → **130 passed (130), 1304 passed (1304)**, exit 0
- same tree, `__tests__` typechecked → **673 errors**

The cleanest single demonstration: `src/server/routers/__tests__/blocks.router.workflow.test.ts` is simultaneously the **#1 typecheck-error file (76 errors)** and **fully green (335 tests passing)**.

(Run hygiene: that file collecting 335 tests rather than 0 is also the documented canary that the `event-engine-common` submodule was populated, so the green above is a real result and not a collection failure reading as a pass.)

**Breakdown** — concentrated, not diffuse:

| top error codes | | top directories | | top files | |
|---|---|---|---|---|---|
| `TS2345` argument type | 128 | `src/server/routers` | 230 | `blocks.router.workflow.test.ts` | 76 |
| `TS2493` tuple index OOB | 110 | `src/server/services/blocks` | 156 | `apps-shared.router.test.ts` | 67 |
| `TS2352` bad cast | 88 | `src/server/services` | 61 | `publish-request.reviewSandbox.test.ts` | 44 |
| `TS2322` not assignable | 85 | `src/env` | 39 | `apps.router.storage.test.ts` | 39 |
| `TS2556` spread arity | 50 | `src/server/jobs` | 30 | `database-target.test.ts` | 39 |
| `TS2737` BigInt literal | 41 | `src/server/services/blocks/steps` | 22 | `step-text-output-moderation.test.ts` | 19 |

Top 3 directories = 447 / 673 (66%). Top 10 files = 336 / 673 (50%).

**Recommendation: worth closing, but not in this PR and not in one step.** 673 is far past "small and mechanical", and the shapes are mostly mock/fixture typing (`TS2345`/`TS2352`/`TS2322`) plus a large `TS2493` cluster that is one idiom repeated. Suggested shape, as a separate issue:

1. Land a `tsconfig.tests.json` that includes the tree and is **not** wired to the gate — makes the number visible and trackable.
2. Burn down the top 3 directories (66% of the backlog) — plausibly a handful of shared mock helpers.
3. Only then move `src/**/__tests__/**` out of the root `exclude`, so the gate is never permanently red. A red gate everyone clicks through is worse than no gate.

Cost of leaving it: a test file can reference a service signature that no longer exists and no gate says so — it fails at *runtime*, in a suite that is report-only in CI. Note `.eslintignore` does **not** exclude this tree, so ESLint already covers it; only `tsc` does not.

---

## History of this PR

It opened as a dependency bump, was marked `[DO NOT MERGE]` when the smoke regression appeared, and was closed. It has been reopened with the bump reverted, because the doc corrections are worth landing on their own and the environment fix belongs elsewhere. Reviewers who read the earlier body: the "why an exact pin" and "upgrade risk 1.57 → 1.61.1" sections are obsolete. The "e2e is unaffected" section, interestingly, was **right** — it claimed only that every config loads and every spec compiles and registers under 1.61.1, and explicitly disclaimed proving they pass against a live preview. The 59 smoke failures were never evidence against it; they were a browser that could not launch.

Two claims made in the course of this PR turned out to be wrong and are retracted in the record above rather than quietly edited out: that a caret is unusable here (it is fine *within* a minor line), and that the bump broke 59 e2e specs at runtime (it did not — the CI image was skewed).

---


ZacxDev:
Two follow-ups after opening this.

**1. Reproduced on the final tree.** The 130/130 number in the description was measured before an A/B reinstall cycle, so I re-ran the suite on the exact committed lockfile afterwards. Independent second run: **130 passed (130), 1304 passed (1304), 0 failed, 80.45s, exit 0**. Two runs, identical counts.

**2. This PR is its own live test of the CI-coupling warning above.** The preview pipeline picked this PR up (`preview / deploy` is running). Its e2e smoke job is exactly the job I flagged: workspace-local `playwright` (now 1.61.1) inside a container image pinned to the **1.57** line with pre-installed browsers.

If my analysis is right, that job should report

```
browserType.launch: Executable doesn't exist at .../chromium_headless_shell-1228/...
```

i.e. the same error as the bug this PR fixes, with CI's revision instead of a laptop's. It is **report-only**, so it degrades a signal rather than blocking anything — but **please read that job's output before merging**: it is the cheapest confirmation that the companion infra image bump is genuinely required, and if it comes back *green* then my read of the coupling is wrong and I'd want to know.

**3. Not fixed here, worth a look separately:** the sibling `civitai-developer-docs` declares `playwright-core: "^1.61.1"`. Its lockfile currently resolves 1.61.1, so it works today — but that is a caret on a package with a 1:1 browser mapping, i.e. the same trap this PR removes here. One re-resolve moves it to 1.62.x / chromium 1234 and it breaks the same way.

ZacxDev:
**Correction — I have edited the "Upgrade risk" section of the description.** Flagging it rather than silently rewriting.

The original version of that section presented a set of release-notes findings as established. On going back to substantiate them, I could not confirm that some had a source behind them, so I re-derived the whole section from the shipped artifacts of both versions instead (`playwright-core@1.57.0` and `@1.61.1` side by side).

**Held up, and now stated with the command that reproduces it:**

- `devtools`, `videosPath`, `videoSize` — declaration counts 3/4/4 → **0/0/0**.
- `handle` on `exposeBinding` — 4 inline overload sites → **0**.
- `ConnectOptions.logger` (already `@deprecated` at 1.57) → **gone**.
- `_react=` / `_vue=` / `:light` really are gone: `lib/server/selectors.js` registers all three at 1.57.0, and at 1.61.1 the file does not exist and none appear under `lib/server`.
- `engines.node` is `>=18` in both.
- This repo uses **none** of the above (0 grep hits).

**Withdrawn:**

- *"278 declared identifiers → 284, strictly additive."* I cannot reproduce those absolute numbers on any basis I tried. The **delta is right** — on the basis I can reproduce (top-level exported type names in `types/types.d.ts`) it is **68 → 74, six added, none removed**, and I have listed the six. Replaced with that.
- *"strictly additive"* as a description of the whole type surface was also too strong, and sat awkwardly one line above a list of removed options. It is additive at the **top-level export** layer; members were removed. Now says so.
- **Which minor each removal landed in** (1.58 vs 1.60), **default headless behaviour unchanged**, and a **1.61.1-specific ESM-loader fix**. Removed rather than restated — I could not verify them from the artifacts, and I would rather this PR under-claim than carry specifics a reviewer might take on trust.

None of this moves the conclusion: the removed surface is unused here, and the suite passes 130/130 files and 1304/1304 tests.

One methodological note worth keeping, since it nearly went the other way: the first pass at the selector-engine question used a **file-level grep count**, which showed `_react` still present in 5 files at 1.61.1 and would have read as "not removed". Those hits are `_reactFiber$` / `_reactProps$` inside Playwright's own bundled report/recorder UI — React internals, not selector engines. Only looking at *what registers a selector engine* answered it.

ZacxDev:
## ⛔ My prediction was wrong, and there is now an unexplained regression. Please do not merge yet.

Earlier I said the smoke job should fail with `Executable doesn't exist at .../chromium_headless_shell-1228/...`, and that green would mean my read of the coupling was wrong. It came back **red for a different reason** — a third case I failed to enumerate, and the interesting one.

**1. The predicted failure did NOT happen. The browser launched.**

`preview / smoke-tests` reports **2 passed · 0 flaky · 59 failed** — 61 results, matching the 61 tests I enumerated. A browser that cannot launch produces **zero** passes and no per-test results at all (that is exactly the `Test Files (130) / Tests no tests` shape this PR fixes). Two passes and 61 individual verdicts mean Chromium started and the suite ran to completion.

So **the CI-image coupling I flagged in the description is not supported by this evidence.** I do not currently know why — the image is pinned to the 1.57 line and the workspace playwright is now 1.61.1, which is the same skew that fails locally. I am not going to invent a mechanism to save the theory. Treat the "⚠️ required companion infra change" paragraph in the description as **unproven**, not as fact.

**2. But something did regress badly, and I cannot attribute it.**

| PR | smoke-tests |
|---|---|
| #3710 (unrelated, the commit this branch is based on) | **60 passed · 1 failed** |
| **#3711 (this PR)** | **2 passed · 59 failed** |

That is not ambient noise — the baseline immediately before this branch is 60/61 green. The `preview-setup` project (`mint preview sessions`) is **not** in the failed list, so session minting succeeded; ~59 of 60 actual specs then failed, across every area at once (route loads, auth gates, SSR seeding, tRPC round-trips).

**Two live hypotheses, and I have not discriminated between them:**

- **(a) The bump genuinely breaks these specs at runtime.** My e2e verification enumerated specs (`--list`) and compared test-id lists — it never *executed* one. I said so in the description; that caveat has now cashed in. `--list` is structurally blind to any runtime behaviour change.
- **(b) The preview environment was cold or unhealthy during that run.** The pipeline's own notes describe a single-replica, cold-start-sensitive preview pod with 60s timeouts. Weak supporting evidence: `https://pr-3711.civitaic.com/` answers **302** (healthy, redirecting to the gate) when probed ~40 min after the run.

**The cheap discriminating control is a smoke re-run on the same commit** — same code, different environment sample. If it comes back ~60/61, it was (b); if it stays ~2/61, it is (a) and this PR needs real work (or is the wrong fix) rather than a merge.

I have not retriggered it, because that costs a full preview rebuild and is the repo owner's call.

**3. What this does not change.** The local finding is unaffected and independently measured: component suite 0 of 130 executed → **130/130 files, 1304/1304 tests**, reproduced twice. All seven GitHub Actions checks are **SUCCESS** (Typecheck, Unit tests, Package unit tests, App unit tests, ESLint + Prettier, Schema drift, submodule pin), as are `preview / component-tests`, `preview / unit-tests`, `preview / auth-tests`, `preview / render-check`, `preview / deploy` and `tekton / typecheck`. The `__tests__` typecheck finding (673 errors) is a separate measurement with its own controls and is untouched by any of this.

But on the strength of the smoke result I would not merge this until item 2 is explained.

Issue #3715, opened 2026-08-07
Dev-token step 3: flip the budgeted-spend default to explicit-request-only (gate: 30 days clean from 2026-08-06)


ZacxDev:
Step 3 of #3703 — flip the dev-token spend default from infer to explicit-request-only. This is the step that actually closes the inference hole; steps 1 and 2 exist to make it safe.

Everything needed to execute it is below, so no context has to be reconstructed.

## The change

`src/pages/api/v1/blocks/dev-token.ts` — one token, in one expression:

```ts
const spendRequested = requestBudgetedSpend ?? true;   // ← today
const spendRequested = requestBudgetedSpend ?? false;  // ← step 3
```

**Rollback:** flip it back. No migration, no persisted state, no token re-issue — dev tokens have a 4h TTL, so the entire population turns over within four hours in either direction.

## Do not flip until the gate below passes

The gate exists because a bearer credential that carries `AIServicesWrite` currently gets budgeted spend *inferred* rather than requested. Flipping while a real client still relies on that inference silently removes spend from working setups.

**Do not run this before ~2026-09-06** (30 days after step 1 merged, `ed1427d9fe`, 2026-08-06).

### Query A — the gate

```sql
SELECT count(*) AS invocations, count(DISTINCT user_id) AS users
FROM block_scope_invocations
WHERE scope = 'ai:write:budgeted'
  AND synthetic_app_id LIKE 'page_local_%'
  AND invoked_at > now() - interval '30 days'
  AND user_id NOT IN (/* internal accounts — see the design comment on #3703 */);
```

Pass condition: **0**.

### Query B — the positive control, run it every time

```sql
SELECT count(*) AS invocations, count(DISTINCT user_id) AS users
FROM block_scope_invocations
WHERE scope = 'ai:write:budgeted'
  AND synthetic_app_id LIKE 'page_local_%';
```

Must return **≥ 29 invocations / 2 users** (the measured historical baseline).

🔴 **Report the pair, never Query A's zero alone.** If B returns 0, the query is wired wrong — a renamed column, a changed prefix, a regressed retry path — and A's zero means nothing. A reassuring zero from a query that cannot see anything is indistinguishable from a real one.

### Primary signal — structured logs

[...]

Issue #3703, opened 2026-08-06
Dev-token: grant `ai:write:budgeted` only when explicitly requested, not inferred from the bearer's AIServicesWrite bit


ZacxDev:
## Summary

`clampDevScopes` grants `ai:write:budgeted` — real Buzz spend — whenever the **bearer credential** carries `AIServicesWrite`. It is *inferred* from a coarse account-level scope bit, never *requested*. This should become an explicit request.

This is the follow-up to #3681 / #3699 (which widens the `civitai-cli` OAuth client to include the AI-services bits). #3699 documents the exposure; this issue closes it.

## Mechanism

`src/pages/api/v1/blocks/dev-token.ts`:

```ts
const keyCanSpend = Flags.hasFlag(session.tokenScope ?? 0, TokenScope.AIServicesWrite);
```

`src/server/services/blocks/dev-scoped-mint.service.ts` — the grant is a strip-unless-able, not a grant-when-asked:

```ts
if (!keyCanSpend) {
  granted = granted.filter((s) => s !== 'ai:write:budgeted');
}
```

`dev-token.ts` is explicit that `session.tokenScope` "is the resolved bitmask for BOTH shapes" — personal API key *or* OAuth consent. So any credential carrying `AIServicesWrite` mints a dev token that can spend, whether or not the caller asked to.

**190 of 221 live `OauthClient` rows already carry `AIServicesWrite`**, so this is reachable today — it is not introduced by #3699. What #3699 changes is that the *first-party CLI* login token can now carry the bit too, which is why the exposure is worth closing now rather than later.

## 🔴 This needs a wire-format change, not just a default flip

The obvious fix — "have the client send a narrowed `scopes` list that omits `ai:write:budgeted`" — **does not work**, and this is the part worth reading before picking up the issue.

`clampDevScopes` narrows by **intersection**:

```ts
if (requestedScopes && requestedScopes.length > 0) {
  const want = new Set(requestedScopes);
  granted = granted.filter((s) => want.has(s));
}
```

So `scopes` can only ever express *"keep exactly these"*. Expressing *"everything this app is entitled to, **minus** `ai:write:budgeted`"* would require the caller to enumerate the full grantable set up front — which depends on the app manifest's `scopeSource`, the server-side allowlist, and the app's OAuth ceiling. A CLI cannot know that set reliably.

The only sentinel the current format has is "absent ⇒ server decides". **Exclusion is inexpressible.**

[...]

PR #3718, opened 2026-08-07
fix(blocks): mirror the five mint-audit events to stdout so the #3703 step-3 gate is readable

ZacxDev:
## The defect

The five App Blocks **mint-audit events** are emitted with `req.log?.info(...)` from `@civitai/next-axiom`. That logger writes **nothing to stdout** in production.

`Logger.sendLogs()` (`@civitai/next-axiom@0.17.0`, `dist/logger.js:198-202`) calls its console `prettyPrint` fallback **only when `config.isEnvVarsSet()` is false** — an explicit "fallback to printing to console … to avoid network errors in development environments" branch. When `AXIOM_TOKEN` and `AXIOM_DATASET` are set, as they are in production, the batch is POSTed to Axiom's HTTP ingest and nothing is written to stdout. The console branch is dead code there.

Consequence: these events are **invisible to our stdout-scraped log store**. Measured over the full retained window — sibling events in the same feature area that are emitted via `console.log(JSON.stringify(...))` returned 15 and 13 rows, while all five `req.log` mint-audit events returned **0**, including one that provably fired at least 15 times (a dev-tunnel session cannot be stopped without first having been minted).

That matters because these events are the audit trail for dev-token mints, and until now they had **no stdout copy at all** — invisible to the log store at any horizon, with no redundancy if the one vendor sink is unreachable. Several comments in `dev-token.ts` asserted the events were queryable from the log store, which was false.

> ### 🔴 Corrected claim — see [this comment](https://github.com/civitai/civitai/pull/3718#issuecomment-5211560283)
>
> An earlier version of this description said this PR made the **30-day #3715 adoption gate** readable from the log store. **That was wrong.** The log store's horizon is **~72h** (global 72h retention, no per-stream override for this namespace; measured: ~1.7M lines at now−70h, zero at now−96h). At ~0.8 bearer mints/day a 72h window holds ~2.4 mints, so a "30-day" query there returns a near-unconditional `0` — the same structurally guaranteed zero this PR removes, at a 3-day horizon instead of 0-day.
>
> **The 30-day gate is read from Axiom** (~96d retention) via the APL queries on #3715. The stdout mirror is a **short-window (~72h) forensic copy** plus vendor redundancy. Both docblocks now say so.
>
> Also: only **3 of the 5** events are gate-bearing — the two dev-tunnel mints carry `spendGranted` only, correct per #3715 but previously unstated, and that path is the busier one.

## The fix

A **dual sink**, not a replacement. Every existing `req.log?.info` call is untouched; each gains a stdout mirror with the **same event name and the same field payload**, via one shared emitter:

`src/server/logging/mint-audit-stdout.ts` → `emitMintAuditToStdout(event, fields)` — writes one JSON line, `{"event": …, …fields}`, matching the existing `app-blocks.dev-tunnel.stop` convention in `dev-tunnel.service.ts`. Emitted unconditionally (these are low-volume audit events, a handful per day). Axiom stays the rich sink; stdout is the copy the log store can actually see.

The five sites (3 in `pages/api/v1/blocks/dev-token.ts`, 2 in `pages/api/v1/block-tokens/index.ts`):

| event | path |
|---|---|
| `blocks.dev-token.pending-mint` | bearer mint, pending publish request |
| `blocks.dev-token.local-mint` | bearer mint, brand-new local manifest |
| `blocks.dev-token.approved-mint` | bearer mint, approved app |
| `app-blocks.dev-tunnel.mint` | dev tunnel, ephemeral pre-approval |
| `app-blocks.dev-tunnel.owned-nonapproved-mint` | dev tunnel, owned non-approved |

**Flags and identifiers only** — no token, secret, or PII on either sink; asserted per event.

The emitter deliberately does **not** normalise, default, or coerce. That is load-bearing: the gate reads a **three-valued** `requestBudgetedSpend` (`true` / `false` / **key omitted**), and turning the omitted case into a `false` or `null` would merge it into the explicit-decline case and make the gate unreadable. The comments that claimed queryability now name the dual sink and cite the transport mechanism, so nobody later removes the mirror as a duplicate log.

## Test coverage

- **Per-event mirror assertions** with enumerated payloads, plus a **dual-sink relationship** assertion pinning the mirror to be exactly the Axiom payload plus `event`, so the two sinks cannot drift apart. The existing `req.log` assertions all still pass — this is dual-sink, and both halves are asserted.
- **The three-valued signal pinned through the mirror**, on the parsed stdout JSON: absent stays absent (asserted as key **presence**, `'requestBudgetedSpend' in parsed`, never truthiness — `false` and absent are both falsy, which is the confusion being guarded), `false` serialises as `false`, `true` as `true`.
- **A call-site ledger** (`src/server/logging/__tests__/mint-audit-stdout.call-site-ledger.test.ts`) that fails when the mint-audit emit-site set **grows or shrinks**, checked in **both** sink directions, following the existing `clampDevScopes` caller-ledger pattern. Includes a positive control so a broken scan cannot pass vacuously.
- **Unit coverage** of the emitter's no-normalisation contract.
- Paths that never mint (404s) are asserted to mirror **nothing**.

### Verification

**Red at base** — every new test was run against base `0dd0e8e57f` with the production source reverted and the new tests in place: **22 failed / 126 passed**. Per file: `dev-token.test.ts` 11/11 new mirror tests red; `dev-tunnel-mint.test.ts` 3 red; `dev-tunnel-owned-nonapproved-mint.test.ts` 3 red; the ledger 5 of 6 red; the emitter unit suite failed to import (`Cannot find module '../mint-audit-stdout'`) and collected **0 tests**. At HEAD: **5 files, 156 tests, all passing**.

Three tests **pass at base** and are labelled honestly as **invariant guards, not regression coverage**: the two "a path that never mints mirrors NOTHING" cases (trivially true when nothing is ever mirrored) and the ledger's "every entry carries a non-trivial rationale" (it only inspects the ledger table).

**Mutation-checked** in all three directions, each failing with its own specific message: deleting a mirror → *"these mint-audit events reach Axiom but have NO stdout mirror …"* naming the event; deleting the Axiom sink → *"…mirrored to stdout but have NO Axiom sink…"*; adding a sixth dual-sinked event → *"the mint-audit event population changed — update MINT_AUDIT_LEDGER"*. The second commit exists because the first version of that guard used two set equalities and reported the **wrong direction** on a deleted mirror; it now asserts set differences so each message matches the failure it fires on.

**Typecheck** (`pnpm typecheck`): base **12** errors, HEAD **12** — identical, none in the touched files (all are pre-existing `event-engine-common` module-resolution errors plus their downstream implicit-anys).

**Lint/format**: prettier clean on all 8 changed files. `eslint` reports 3 `no-empty-function` errors, all pre-existing — confirmed by diffing: this change adds and removes no empty-arrow lines.

### Note on diff noise

`prettier --write` on the changed files also reflowed a handful of untouched lines. All five touched files were **already prettier-dirty at base** (verified by comparing each base blob against prettier's own output), so those hunks are pre-existing violations being fixed, not formatting churn introduced here. No behaviour change in them.

## Result

These five audit events gain a stdout copy they never had, so they are readable at incident time from the stdout-scraped log store (~72h) and no longer depend on a single vendor sink being reachable. The 30-day #3715 adoption gate continues to be read from Axiom — see the corrected-claim note above.


ZacxDev:
## 🔴 Correction to this PR's description — the stdout mirror is a ~72h surface, not a 30-day one

An audit caught a wrong claim in my original description and docblocks. Posting it here rather than only editing the body, since anyone who already read the description read the wrong version.

**What I claimed:** that stdout is "the copy the log store can actually see", framed so that #3715's adoption gate becomes readable from the log store.

**What is actually true:** the stdout-scraped log store's horizon is **~72h**. Its global retention is 72h with compaction-based enforcement, and there is **no per-stream retention override covering this application's namespace**. Live probe: ~1.7M lines present at now−1h / −24h / −48h / −70h, and **zero** at now−96h and now−8d.

**Why that matters, priced from the consuming side:** #3715's gate is `spendGrantBasis === 'inferred'` over a **rolling 30 days**, against a base rate of ~24 bearer mints per 30 days (~0.8/day). A 72h window holds ~2.4 mints in total, and `inferred` is a subset of those. So a "30-day" query against this new surface returns a hard `0` essentially unconditionally — **a structurally guaranteed zero, which is the exact defect this PR was written to remove, just relocated from a 0-day horizon to a 3-day one.** As originally written, my docblock was an instruction to read a false zero and flip a Buzz-spend default on it.

### The corrected division of labour

| Sink | Horizon | What it is for |
|---|---|---|
| **stdout** (this PR) | **~72h** | Short-window forensic copy — incident-time reads, plus a second sink that does not depend on one vendor being reachable |
| **Axiom** (`req.log?.info`) | ~96 days | Long-window analytical copy. **The 30-day #3715 gate is read from here**, via the APL queries + schema check recorded on that issue |

This PR still does something real — before it, these events had **no** stdout copy at all, so they were invisible to the log store at any horizon, and there was no sink redundancy. It just does not make stdout a substitute for Axiom on the 30-day question.

**A per-stream 720h override is not an available fix**, and I am not claiming it as one: those selectors match **stream labels, not line content**, and these lines share the app's ordinary stdout stream. Giving them 720h would require relabeling from line content across the log pipeline — new streams, cardinality risk. Unproven follow-up at best, not part of this PR.

### Also corrected in the same commit

- **Only 3 of the 5 events are gate-bearing.** The two dev-tunnel mints carry `spendGranted` only — no `spendGrantBasis`, no `requestBudgetedSpend`. That is correct per #3715 (the tunnel path has no per-mint request mechanism, so there is no basis to derive), but it was unstated, and the tunnel path is the **busier** one — so counting all five against the gate gives the wrong denominator. Now a `kind: 'gate-bearing' | 'audit-only'` field on the ledger, pinned to the source by a test.
- **A production-only silence left every test green.** Wrapping the emit in `if (process.env.NODE_ENV !== 'production')` passed 159/159, because the test runner's `NODE_ENV` is never `production` — the one mutant whose real-world consequence is precisely this PR's target failure mode. There is now a test that stubs `NODE_ENV='production'` and asserts the line is still written; the mutant dies.
- **The call-site ledger was escapable by spelling** in the only direction that mattered: an Axiom-only `…minted-extra` slipped past the anchored `/(^|[.-])mint$/` while `…extra-mint` was caught. Widened to `/mint/i`; both spellings now fail.
- **The LogQL equivalent is recorded** next to the docblock, with a positive control and its 72h horizon stated — the two sinks spell fields differently (Axiom puts the event in `message` with `['fields.<name>']`; stdout puts it in `event` with fields top-level), so a query written for one returns a confident zero against the other rather than an error.
- The ledger's positive control was `FILES.length > 500` against a real 3,564 — it would have passed having silently lost 85% of the tree. Raised to `>3000`.

Verification after the round: typecheck **12 errors, same as base**, none in touched files; **161 tests passing** across the 5 affected suites; both mutants above confirmed dying with their own assertion messages.

PR #3722, opened 2026-08-07
[DO NOT MERGE] test(ci): positive control — Typecheck must RUN on a non-main base (#3721)

ZacxDev:
test(ci): DO NOT MERGE — positive control for the Typecheck `if:` in #3721

#3721 narrows the Actions Typecheck job to `fork || base != main`. Its own PR
proves only the FALSE branch: an internal PR onto `main`, where the job correctly
reports `skipped`.

A gate observed only skipping is not a gate shown to run. This branch carries the
identical lint.yml onto a PR targeting `release`, so the second disjunct
evaluates TRUE and the job must actually EXECUTE. If it skips here too, the
retained coverage in #3721 is inert and the PR should not merge as written.

Closed as soon as the check result is read.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,20 +2,30 @@ name: Lint
 
 # Regression guard for code quality.
 #
-# Typecheck is full-repo and BLOCKING, and is the only full-repo gate here. It
-# runs on fork PRs TOO — no job in this file carries a top-level `if:`, so none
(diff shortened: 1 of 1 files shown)


ZacxDev:
Control passed: `Typecheck (fork PRs / non-main base)` ran and reported **success** on this `release`-base PR, while on #3721 (internal → main) the same job reports **skipped**. Both branches of the `if:` are now verified by execution rather than by reading the expression. Closing — this was never meant to merge.

PR #3726, opened 2026-08-07
fix(app-blocks): IframeHost's status gates read a stale closure at commit

### LINKED ISSUE #3685: OPEN_BUZZ_PURCHASE: IframeHost silently drops a pre-ready request while PageBlockHost NACKs (30s hang vs a tick)
## Summary

After #3680, `OPEN_BUZZ_PURCHASE` has **two different observable behaviours depending on which host the block is mounted in**, and the SDK hook that sends it is host-agnostic.

- `PageBlockHost` (page apps): a pre-ready `OPEN_BUZZ_PURCHASE` is refused **and replied to** — `BUZZ_PURCHASE_RESULT { purchased: false }` — so the block's promise settles in a tick.
- `IframeHost` (model-slot apps, `IframeHost.tsx:1196`): the same message is refused and **silently dropped**, so the block's promise hangs for the SDK's full `DEFAULT_REQUEST_TIMEOUT_MS` (30s, `@civitai/blocks-react@0.39.0`).

`useBuzzPurchase` does not know or care which host it is talking to. Same call, same payload, same user action — 30 seconds apart in outcome.

This is deliberately **not** framed as "the bug still exists over there". Before #3680 both hosts hung, which was at least consistent. The divergence is the new part, and divergence is the thing that produces bug reports nobody can reproduce ("it works on the page app").

## Cause

Same shape as the defect #3680 fixed on the page host: `IframeHost` gates these handlers on a `status` captured in the effect closure, at approximately

- `IframeHost.tsx:1072` — `REQUEST_SIGN_IN`
- `IframeHost.tsx:1094` — `REQUEST_CONSENT`
- `IframeHost.tsx:1196` — `OPEN_BUZZ_PURCHASE`

`REQUEST_SIGN_IN` and `REQUEST_CONSENT` are fire-and-forget with no `requestId` and no reply message, so they cannot NACK; only `OPEN_BUZZ_PURCHASE` is repliable, and it is the one that hangs.

Note this issue is about the **NACK divergence**, which is user-visible today. Whether `IframeHost` also has the stale-closure race is a separate (and likely yes) question — it was deliberately left out of #3680 because all the measured evidence was on `PageBlockHost`, and widening the blast radius without evidence would have been guesswork.

## Why nothing in the repo will catch this

`hostHandlerParity.ts` records which request/reply message **types** each host handles. It has no notion of *when a handler declines* or *whether declining produces a reply*, so both hosts look identical to it: each handles `OPEN_BUZZ_PURCHASE` and each can emit `BUZZ_PURCHASE_RESULT`. The parity table is green and will stay green.

If this is fixed, the parity check is the natural place to extend — from "handles type X" to "handles type X, and its refusal path answers on the same requestId" — otherwise the next host to be added re-opens the same gap.

## Suggested fix

Mirror #3680 on `IframeHost`: reply `{ requestId, purchased: false }` on the not-ready branch when a string `requestId` is present, and keep dropping the genuinely unrepliable case (missing / non-string `requestId` — there is no id to answer on). The page host's version, its per-site reasoning, and its guards are in `PageBlockHost.tsx` and `PageBlockHostWorkflow.browser.test.tsx`.

## Verification bar

Whoever picks this up: the guard needs a positive control. A test asserting "a NACK arrives" is easy; a test asserting "no modal opened" is a zero, and a host whose listener was never live produces that same zero. `PageBlockHostWorkflow.browser.test.tsx` has both shapes to copy, and `test/commitWindow.tsx` documents two ways of reproducing the commit window that were measured and rejected before a deterministic one worked.

Found by an adversarial audit of #3680.

ZacxDev:
Closes #3685.

The same defect PR #3680 (`5c76917c24`) just fixed in `PageBlockHost.tsx`, on the model-slot host. That fix is now verified live in production; this is its twin.

## What broke

**Root cause — a stale closure.** Four `IframeHost` message handlers gate on the host being ready: `RESIZE_IFRAME`, `REQUEST_SIGN_IN`, `REQUEST_CONSENT` and `OPEN_BUZZ_PURCHASE`. Each closed over `status` from the render that registered it, with `status` in its effect's dependency array — so the LIVE listener only picks up a new status when React flushes that effect.

React writes the `data-block-ready` DOM attribute during the **COMMIT** (`IframeHost.tsx:2075`) but flushes **passive** effects in a LATER scheduler task. So there is a window in which the host publicly reports ready while the live handler still holds `status === 'loading'` and drops the message.

**Never hang.** The one gate carrying a `requestId` silently dropped a refused request. It was self-documenting — a comment reading *"(silently dropped)"* sat directly above `if (requestId == null || !raw) return;`. The block awaits `BUZZ_PURCHASE_RESULT` on a promise the SDK is understood to reject only at a 30s request timeout, so a drop cost the user's click plus half a minute of nothing.

> ⚠️ **That 30s figure is an INFERENCE carried over from #3680, not a measurement, and an earlier revision of this body asserted it as fact.** `@civitai/blocks-react` is **not a dependency of this repo and is not on disk** — 0 hits in `pnpm-lock.yaml` (positive control: `app-sdk` → 3 hits). The installed SDK is `@civitai/app-sdk@0.14.0`, which implements neither `OPEN_BUZZ_PURCHASE` nor any request timeout. So the specific "reject at 30s → resolve immediately" behaviour change **cannot be verified in-tree**; the reasoning stands, the citation does not. What IS verified here is host-side: the refused request is answered rather than dropped.

## Why the render-body ref placement is the fix

`statusRef.current = status` is assigned in the **render body**, not an effect — an effect here is precisely the deferral being fixed. A render-body assignment happens BEFORE the commit, so by the time anything observable says "ready", the mirror already is.

The four handlers read it through one shared `readGateStatus()`. Unlike `PageBlockHost`'s version this needs **no `'error' → 'no_token'` shim**: `IframeHost`'s local `Status` (`:101`) is byte-identical to the shared gates' `HostStatus` (`openBuzzPurchaseGate.ts:16`), so it is simply `() => statusRef.current`. It deliberately does **not** import `toHostGateStatus` / `pageBlockHostLogic` — the page host is a sibling surface, not a dependency of this one, and there is nothing to share but an identity function.

`status` then drops out of those four effects' dependency arrays, so they subscribe **once per mount** instead of tearing down and re-registering on every status transition. The window this fix is about cannot exist if the listener never needs replacing.

The concurrent-rendering safety note is ported from `PageBlockHost.tsx` because its reasoning transfers unchanged — including that the transient ref-vs-DOM disagreement runs in **both** directions. `loading → ready` is permissive (the point of the fix); `ready → fatal` (`BLOCK_ERROR{fatal:true}`) is restrictive, and `IframeHost` has that transition too. Refusing a spend on a host that has already decided it is broken is the safer side of that trade.

## The NACK, and why only one gate gets one

`resolveBuzzPurchaseRequest` returns null for **two** reasons, and they need different answers:

- **malformed payload** (no string requestId) — **UNREPLIABLE**. There is no id to thread a reply back on, so it can only be dropped, in silence.
- **host not ready** — **REPLIABLE**, and it must be answered: `send('BUZZ_PURCHASE_RESULT', { requestId, purchased: false })`, guarded on `typeof raw.requestId === 'string' && raw.requestId.length > 0`.

`purchased: false` is the same shape the modal's own `onClose` sends for a dismissed purchase, and (per the same #3680 inference above, not verifiable here) what the SDK reads as "no purchase happened". **The refusal is unchanged** — no modal opens, nothing is charged; the block just finds out now instead of in 30 seconds.

🔴 **The NACK is not an unconditional improvement, and the code now says so.** `usePostMessage` dedups INBOUND messages by `payload.requestId` for `DEDUP_WINDOW_MS = 5000` (`usePostMessage.ts:46`, logic `:155-173`) **before any handler runs**. A block that reacts to `purchased: false` by retrying with the **same** `requestId` inside five seconds is swallowed by the transport and hangs exactly as it did before. The NACK converts a guaranteed hang into a fast failure only for a block that retries with a FRESH id, or after the window. The dedup is a replay defence and is deliberately **not** changed here.

`RESIZE_IFRAME`, `REQUEST_SIGN_IN` and `REQUEST_CONSENT` get **no** NACK: all three are fire-and-forget with no requestId and no host→block reply message, so there is nothing to reply to and no promise to fail fast.

## One pre-existing bug fixed rather than inherited

The `OPEN_BUZZ_PURCHASE` effect reads `modelCtx.slotId` for FIN-1 attribution but did not list it in its deps. Unreachable today — `BlockSlot.tsx:51` re-keys the subtree on `(slotId, entityType, entityId)`, so a slotId change remounts the host rather than re-rendering it — but it is the same stale-closure family, and silently preserving it while rewriting that array would be inheriting a bug on purpose.

Attribution is otherwise untouched: `deriveScopeFromInstanceId` and the `mbi_*` / `bus_*` / `pdb_*` prefixes live on the success path, after the gate.

## Tests

New `IframeHostStatusGates.browser.test.tsx` mounts the **real** `IframeHost` and drives the **real** postMessage bridge. It is a new file rather than growth of the beacon-focused `IframeHost.browser.test.tsx` because it needs a mocked `openLoginPopup` and the dialogStore, and it mirrors `PageBlockHostWorkflow` / `PageBlockHostSignIn`.

`test/commitWindow.tsx`'s `onReadyAttributeWrite` generalised to `IframeHost` with **zero** changes — it patches `Element.prototype.setAttribute` and fires on the first `data-block-ready="true"` write by any element, synchronously on React's own stack.

Eleven guards (5 in the original round, 6 added after the audit — see the follow-up commit and the comment below):

1. **Commit-gap positive** (`OPEN_BUZZ_PURCHASE`) — posted from inside the commit window; asserts the modal opens, pinning the dialog id `block-buy-buzz-rq_gap` so it proves *this* post opened it.
2. **Pre-ready NACK** — no ready drive; zero dialogs AND a `BUZZ_PURCHASE_RESULT {purchased:false}` on the same requestId.
3. **No-requestId silent drop**, with a positive control that runs **first** and is driven all the way to an observable reply, so the zero is a measurement rather than a probe wired to nothing. 🔴 It counts only `BUZZ_PURCHASE_RESULT`, never `replies.received.length` — #3680 has a follow-up commit precisely because a total count drifted on unrelated host pushes and failed against a correct host.
4. **`REQUEST_SIGN_IN` commit-gap** — the guard that proves the ROOT-CAUSE half on its own. There is no NACK for this message, so guard 2 cannot mask a regression here.
5. **`REQUEST_SIGN_IN` pre-ready drop** — the negative half of 4; without it, "the handler is live" and "the handler always fires" are indistinguishable.
6. **`RESIZE_IFRAME` commit-gap** — asserts the iframe's inline height actually moves off `minHeight`, the user-visible effect rather than a mock call count.
7. **`RESIZE_IFRAME` pre-ready refusal** — the slot height does not move before the handshake, with a positive control (the same message after ready DOES move it).
8. **`REQUEST_CONSENT` commit-gap** — pins the modal's own `missingScopes` prop, not just `dialogs.length`, so any dialog from any source cannot satisfy it.
9. **`REQUEST_CONSENT` pre-ready refusal** — no pre-handshake permission prompt, with the same positive control.
10. **NACK predicate, `requestId.length > 0`** — an EMPTY-STRING id is a different branch from guard 3's missing-key case and must still be dropped in silence.
11. **NACK predicate, `typeof requestId === 'string'`** — 🔴 with **two** payloads, because the obvious one does not discriminate: `requestId: 5` leaves the mutant sending nothing either way (`(5).length` is `undefined`), while `requestId: ['rq']` has a truthy `.length` and makes the mutant reply with an id of a type the protocol does not have.

### Mutation matrix

🔴 **The first round of this PR shipped a matrix that covered the mechanism and not the handlers.** An adversarial audit found half the fix unguarded, and **re-measurement confirmed it**: reverting EITHER `RESIZE_IFRAME` or `REQUEST_CONSENT` to its exact pre-PR form left the whole component suite green at 380/33. Deleting either handler's gate outright survived too — a gap this body did not disclose. The six `n*` mutants below are the repair.

12 mutants × 2 runs, restored **byte-identically** (`cmp`) between each, every kill carrying **that guard's own assertion**:

| mutant | what it breaks | guard(s) killed | runs | assertion |
|---|---|---|---|---|
| `n1_resize_closed_over` | `RESIZE_IFRAME` reverts to closed-over `status` (+ deps) | resize commit-gap | 2/2 | `expected 200 to be 640` |
| `n2_consent_closed_over` | `REQUEST_CONSENT` reverts to closed-over `status` (+ deps) | consent commit-gap | 2/2 | `expected [] to have a length of 1 but got +0` |
| `n3_resize_always_ready` | delete the `RESIZE_IFRAME` status gate | resize pre-ready refusal | 2/2 | `expected 640 to be 200` |
| `n4_consent_always_ready` | force `resolveRequestConsent` to see `'ready'` | consent pre-ready refusal | 2/2 | `expected [...] to have a length of +0 but got 1` |
| `n5_nack_len_ge_zero` | `requestId.length > 0` → `>= 0` | empty-requestId drop | 2/2 | `expected [...] to have a length of 1 but got 2` |
| `n6_nack_no_typeof` | drop `typeof raw.requestId === 'string' &&` | non-string-requestId drop | 2/2 | `expected [...] to have a length of 1 but got 2` |
| `n6b_nack_no_typeof_safe` | same, made null-safe so it cannot die by throwing instead | non-string-requestId drop | 2/2 | `expected [...] to have a length of 1 but got 2` |
| `m1_ref_in_effect` | `statusRef.current = status` back into an effect | buzz / sign-in / **resize** / **consent** commit-gap (4) | 2/2 | `expected [] to have a length of 1`; `… to be called 1 times, but got 0`; `expected 200 to be 640` |
| `m2_no_nack` | delete the NACK `send` | pre-ready NACK + **both** new NACK-predicate guards' positive controls (3) | 2/2 | `Error: no NACK yet`; `expected [] to have a length of 1` ×2 |
| `m3_signin_closed_over` | `REQUEST_SIGN_IN` reverts to closed-over `status` | sign-in commit-gap | 2/2 | `expected "vi.fn()" to be called 1 times, but got 0` |
| `m4_nack_too_wide` | NACK widened to answer the UNREPLIABLE case | no-requestId + **empty** + **non-string** drops (3) | 2/2 | `expected [...] to have a length of 1 but got 2` / `but got 3` |
| `m5_buzz_closed_over` | `OPEN_BUZZ_PURCHASE` reverts to closed-over `status` | buzz commit-gap | 2/2 | `expected [] to have a length of 1` |

Deterministic — no mutant survived either run, and run 2 reproduced run 1 exactly. Note `n6b`: `n6` as literally stated by the audit (`raw.requestId.length`) ALSO throws a `TypeError` on guard 3's no-requestId payload, which would muddy attribution; `n6b` is the null-safe form, and guard 11 is the sole killer of both.

**The original five were re-run after this round's changes** (an audit fix resets the verification gate) and are unchanged except that three of them now die to MORE guards — the new guards' positive controls are themselves sensitive to the mechanism.

#### Harness change

`startReadyDrive`'s interval goes **10 ms → 40 ms**. `usePostMessage` admits 30 inbound messages per 1000 ms among types that have a subscriber (`usePostMessage.ts:44-45,181`); at 10 ms the drive asked for **100/s** against that budget. It worked because the decisive post lands in the first few ticks — i.e. it was under the limit by luck of timing rather than by construction, and each guard added here narrows that margin. 40 ms is 25/s with headroom for the commit-window post. The full 12-mutant matrix above was run **after** this change.

### Gate

- **386/386** AppBlocks browser tests (33 files), up from 380 — read from the output, not the exit code. The status-gates file itself reports **11 tests**.
- Unit project (`vitest --project unit`): **12,942 passed / 7 failed / 1 skipped** (863 files). 🔴 The 7 failures are **pre-existing and unrelated** — every one is `TypeError: Cannot read properties of undefined (reading 'clear'/'setItem')` in `src/components/AppBlocks/__tests__/hiddenBlocks.test.ts`, i.e. `window.localStorage` is undefined in that file's environment. **Control run:** reverting both of this round's files to their parent-commit versions reproduces all 7 identically, and neither file is in that test's import graph. **CI's own `Unit tests` check is SUCCESS on this commit** (`985ea3d`), so this is a local-host artifact — the lockfile and installed `happy-dom` agree at 20.9.0, so it is not a version mismatch.
- Typecheck: **0 errors**, with a positive control — `tsc --listFiles` confirms the test file is in the program, and a deliberately injected `string`-to-`number` error IS reported (`error TS2322`), so the clean run is a measurement.
- ESLint `--max-warnings=0`: 0 errors, 0 warnings on **both** touched files, instrument validated (an injected `var` reports `rc=1` / `no-var`; note `no-console` and `import/first` are NOT enabled in this config, so they are useless as controls).
- Prettier: the new file is clean (CI blocks on added files). `IframeHost.tsx` is **not** reformatted wholesale — it is already Prettier-dirty on `main` (CI is report-only for modified files), and this change adds **zero** new violations: 7 formatting hunks before, 7 after.

## What these tests CANNOT prove

- **That a real deployed block hits the window.** #3680 was explicit that this was inferred, not observed, and the same holds here: no SDK code path auto-posts a gated request on ready — all of these are `useCallback`s the app invokes — so reaching the window needs app code that calls one in the same turn it observes ready. No such block was observed. This is a latent defect being closed, not a reported outage.
- **There is no live-repro path for `IframeHost` at all.** `/apps/dev` renders `PageBlockHost` only, so unlike #3680 there was no manual browser reproduction available even in principle. The evidence here is the mutation matrix, not a live click path.
- ~~**`RESIZE_IFRAME` and `REQUEST_CONSENT` have no dedicated commit-gap guard.**~~ **CLOSED** — that residual was understated and is now fixed. It was not only a per-handler revert that went uncaught: **deleting either handler's status gate outright also survived the whole suite**, so a pre-handshake block could drive the slot height or pop the permission modal with nothing going red. `REQUEST_CONSENT` additionally had **zero** browser coverage against `IframeHost` anywhere in the tree. Guards 6–9 cover both directions for both handlers; `n1`–`n4` above are the proof.
- **The component suite does not run in GitHub Actions.** `lint.yml` runs typecheck, unit, packages and apps — no Chromium. These guards are locally-verified only.
- **The SDK-side half of "never hang" is unverifiable in this repo** — see the callout in *What broke*. `@civitai/blocks-react` is not installed here, so neither the 30s timeout nor `useBuzzPurchase`'s reading of `purchased: false` was checked against real code. Only the host side is measured.
- **A block that retries a refused purchase with the SAME `requestId` inside 5 s still hangs**, swallowed by `usePostMessage`'s inbound dedup before any handler runs. Documented in the code rather than fixed; changing the dedup is a wider change than this PR earns.

Scope was deliberately held to IframeHost's four status gates. The general "every repliable gate answers on the same requestId" parity guard is a separate project — it goes red on ~7 pre-existing sites across both hosts — and `hostHandlerParity.ts` was not extended (its own comment says its `reply` field is "DOCUMENTATION ONLY — NOTHING ENFORCES THIS STRING", and its runtime half is a grep that structurally cannot see this property).

### DIFF
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -558,6 +558,77 @@ export function IframeHost({
   const features = useFeatureFlags();
   const iframeRef = useRef<HTMLIFrameElement | null>(null);
   const [status, setStatus] = useState<Status>('loading');
+  // Mirror of `status`, read by the four status-gated message handlers
+  // (RESIZE_IFRAME, REQUEST_SIGN_IN, REQUEST_CONSENT, OPEN_BUZZ_PURCHASE) via
+  // `readGateStatus` below.
+  //
+  // 🔴 ASSIGNED IN THE RENDER BODY, NOT IN AN EFFECT — that placement IS the fix,
+  // and an effect here is the bug. React writes the `data-block-ready` DOM
+  // attribute (and every other commit-time output) during the COMMIT, but flushes
+  // PASSIVE effects in a LATER scheduler task whenever the commit exhausts the 5ms
+  // frame budget (scheduler 0.23.2, MessageChannel transport, `frameYieldMs = 5`).
+  // So an effect-updated mirror is stale for exactly the window in which the host
+  // has already publicly announced readiness. A render-body assignment happens
+  // BEFORE the commit, so by the time anything observable says "ready" the mirror
+  // already is. This is the same defect, and the same fix, as PageBlockHost's —
+  // see the 🔴 note on its own `statusRef` for the measurement.
+  //
+  // CONCURRENT-RENDERING SAFETY. Writing a ref during render is impure, so the
+  // three ways that bites were each checked rather than assumed:
+  //   • StrictMode double-render writes the SAME value twice — idempotent.
+  //   • An interrupted/discarded concurrent render can leave the ref holding a
+  //     status that never committed. It converges: React always eventually renders
+  //     the latest state, and that render overwrites the ref.
+  //   • It is only ever written from `status` (component state), never derived
+  //     from props or anything a parent can change mid-render.
+  //
+  // 🔴 WHICH DIRECTION THAT TRANSIENT DISAGREEMENT RUNS IN — read this before
+  // pointing a NEW gate at the ref. "Ref ahead of the DOM" is NOT always the
+  // permissive direction; it inherits the direction of the transition in flight,
+  // and this component has transitions BOTH ways:
+  //   • loading → ready is PERMISSIVE. The ref says ready while `data-block-ready`
+  //     still reads "false", so a gate can open a beat before the host has publicly
+  //     announced readiness. Harmless, and it is the whole point of the fix.
+  //   • ready → fatal (`BLOCK_ERROR{fatal:true}`, the handler below) is
+  //     RESTRICTIVE. A concurrent render writes `statusRef.current = 'fatal'`,
+  //     yields before commit, and a queued OPEN_BUZZ_PURCHASE arriving in that gap
+  //     is REFUSED while `data-block-ready` still reads "true".
+  // The restrictive case is deliberate and is the safer side of the trade: the ref
+  // is the host's freshest knowledge of its own state, so refusing a spend on a
+  // host that has already decided it is broken is correct, and the DOM attribute is
+  // the thing lagging. A gate that must NOT tighten early (there is none today)
+  // would have to key off the committed `status`, not this ref.
+  //
+  // RESIDUAL WINDOW, stated rather than hidden: this closes the commit→passive-
+  // effect gap, NOT the larger message-received→render gap. `status` becomes
+  // 'ready' only as a RESULT of the block's own BLOCK_READY being processed
+  // (setStatus → render → commit), so a block that posts a gated request in the
+  // SAME turn as BLOCK_READY still meets a 'loading' mirror — React has not
+  // rendered yet. The NACK on OPEN_BUZZ_PURCHASE (the one repliable gate here) is
+  // what covers that remainder — a drop there fails FAST instead of hanging.
+  const statusRef = useRef<Status>('loading');
+  statusRef.current = status;
+  /**
+   * The ONE way a message handler asks "is the host ready?".
+   *
+   * Reads the RENDER-BODY-updated `statusRef` rather than closing over `status`,
+   * so the answer is current the moment the render that made the host ready is
+   * committed — not one scheduler task later, when React gets around to flushing
+   * passive effects and re-registering the listener with a fresh closure.
+   *
+   * No `'error' → 'no_token'` shim, unlike PageBlockHost's version: this host's
+   * local `Status` union is byte-identical to the shared gates' `HostStatus`
+   * (`openBuzzPurchaseGate.ts`), so the value passes straight through. Deliberately
+   * NOT imported from `pageBlockHostLogic` — the page host is a sibling surface,
+   * not a dependency of this one, and there is nothing to share but an identity fn.
+   *
+   * Stable identity (`[]` deps, ref read only) is load-bearing in its own right:
+   * it lets the four gated effects DROP `status` from their dependency arrays, so
+   * they now subscribe ONCE per mount instead of tearing down and re-registering
+   * on every status transition. The window this fix is about cannot exist if the
+   * listener never needs replacing.
+   */
+  const readGateStatus = useCallback((): Status => statusRef.current, []);
   const [iframeHeight, setIframeHeight] = useState<number>(
     install.manifest.iframe?.minHeight ?? 200
   );
@@ -1119,11 +1190,24 @@ export function IframeHost({
       // is visible-but-non-interactive (pointerEvents:none) before ready and
       // pinned at minHeight, so an early RESIZE would let a pre-ready block
       // push the slot height around before the handshake completes.
-      if (status !== 'ready') return;
+      //
+      // `readGateStatus()` (not a closed-over `status`) — see its definition: it
+      // reads the render-body-updated mirror, so it is current at COMMIT rather
+      // than one scheduler task later.
+      //
+      // NO NACK HERE, deliberately. RESIZE_IFRAME is fire-and-forget: it carries
+      // no requestId and there is no host→block reply message for it, so there is
+      // nothing to reply TO and no promise to fail fast. Dropping it costs the
+      // block one un-honoured height, not a hang.
+      if (readGateStatus() !== 'ready') return;
       applyHeight((raw as { height?: unknown }).height);
     });
     return off;
-  }, [onMessage, applyHeight, status]);
+    // `status` is deliberately ABSENT: the handler reads it through
+    // `readGateStatus` (a render-body-updated ref) instead of closing over it, so
+    // this subscribes ONCE per mount. Re-registering on every status transition is
+    // what opened the commit→passive-effect window in the first place.
+  }, [onMessage, applyHeight, readGateStatus]);
 
   useEffect(() => {
     const off = onMessage<unknown>('BLOCK_ERROR', (raw) => {
@@ -1146,7 +1230,14 @@ export function IframeHost({
   // after login). When absent or unsafe we fall through to the current page.
   useEffect(() => {
     const off = onMessage<{ returnUrl?: unknown } | undefined>('REQUEST_SIGN_IN', (raw) => {
-      const resolved = resolveRequestSignIn(status, raw);
+      // `readGateStatus()` (not a closed-over `status`) — see its definition.
+      const resolved = resolveRequestSignIn(readGateStatus(), raw);
+      // NO NACK HERE, deliberately. REQUEST_SIGN_IN is fire-and-forget: the SDK
+      // sends it with `dispatch` (blocks-react 0.39.0 `useRequestSignIn`), the
+      // payload is `{ returnUrl? }` with NO requestId, and there is no host→block
+      // reply message for it. Nothing to reply to; a drop cannot hang the block —
+      // it just silently does nothing, which is precisely why the ROOT-CAUSE half
+      // of this fix has to hold for this handler.
       if (resolved == null) return; // not ready — drop (gate centralises the rules)
       // Open the hub login in a popup (replaces the old in-page LoginModal). The host runs at TOP level — not
       // inside the sandboxed block iframe — so the popup works here; on completion it navigates back.
@@ -1154,7 +1245,8 @@ export function IframeHost({
       openLoginPopup(resolved.returnUrl ?? here, 'image-gen');
     });
     return off;
-  }, [onMessage, status]);
+    // `status` deliberately absent — see the RESIZE_IFRAME deps note.
+  }, [onMessage, readGateStatus]);
 
   // Lazy consent (A6): the block (rendered in full for a logged-in viewer whose
   // token is missing a consent-gated scope) asks the host to open the consent UI
@@ -1168,7 +1260,14 @@ export function IframeHost({
   // and the block retries — there is no host→block reply (fire-and-forget).
   useEffect(() => {
     const off = onMessage<{ scopes?: unknown } | undefined>('REQUEST_CONSENT', () => {
-      const scopesToGrant = resolveRequestConsent(status, missingScopes ?? []);
+      // `readGateStatus()` (not a closed-over `status`) — see its definition.
+      //
+      // NO NACK HERE, deliberately. REQUEST_CONSENT is fire-and-forget in both
+      // directions: the SDK sends it with `dispatch` (blocks-react 0.39.0
+      // `useRequestConsent`), its payload is `{ scopes? }` with NO requestId, and
+      // there is no host→block reply message for it — so there is nothing to
+      // reply TO and no promise to fail fast. Dropping it cannot hang the block.
+      const scopesToGrant = resolveRequestConsent(readGateStatus(), missingScopes ?? []);
       if (scopesToGrant == null) return; // not ready, or nothing missing — drop
       dialogStore.trigger({
         component: BlockConsentModal,
@@ -1183,9 +1282,10 @@ export function IframeHost({
       });
     });
     return off;
+    // `status` deliberately absent — see the RESIZE_IFRAME deps note.
   }, [
     onMessage,
-    status,
+    readGateStatus,
     missingScopes,
     install.appBlockId,
     install.manifest.name,
@@ -1269,9 +1369,35 @@ export function IframeHost({
         // still visible-but-non-interactive (pointerEvents:none). Summoning the
         // money-spend modal pre-ready would let an untrusted block nag the user
         // before any interaction. resolveBuzzPurchaseRequest returns null
-        // (silently dropped) when status !== 'ready' or the payload is bad.
-        const requestId = resolveBuzzPurchaseRequest(status, raw);
-        if (requestId == null || !raw) return; // !raw is implied by requestId != null; narrows for TS
+        // when status !== 'ready' or the payload is bad.
+        //
+        // `readGateStatus()` (not a closed-over `status`) — see its definition.
+        const requestId = resolveBuzzPurchaseRequest(readGateStatus(), raw);
+        if (requestId == null || !raw) {
+          // NEVER HANG. `resolveBuzzPurchaseRequest` returns null for TWO
+          // different reasons, and they need different answers:
+          //
+          //   • malformed payload (no string requestId) — UNREPLIABLE. There is
+          //     no id to thread a reply back on, so it can only be dropped.
+          //   • host not ready — REPLIABLE, and it must be replied to. The block
+          //     is awaiting BUZZ_PURCHASE_RESULT on a promise that rejects only
+          //     at the SDK's 30s default request timeout (blocks-react 0.39.0,
+          //     `DEFAULT_REQUEST_TIMEOUT_MS`), so a silent drop costs the user's
+          //     click plus half a minute of nothing happening.
+          //
+          // `purchased: false` is exactly what the modal's own onClose sends for
+          // a dismissed purchase below, and exactly what the SDK's
+          // `useBuzzPurchase` reads as "no purchase happened". Refusing is
+          // unchanged — no modal opens, nothing is charged; the block just finds
+          // out now instead of in 30 seconds.
+          //
+          // `!raw` is implied by requestId != null; the compound condition also
+          // narrows for TS.
+          if (raw && typeof raw.requestId === 'string' && raw.requestId.length > 0) {
+            send('BUZZ_PURCHASE_RESULT', { requestId: raw.requestId, purchased: false });
+          }
+          return;
+        }
         const rawAmount =
           typeof raw.suggestedAmount === 'number' && Number.isFinite(raw.suggestedAmount)
             ? raw.suggestedAmount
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Audit follow-up — the two gates this fix left unpinned

An adversarial audit found the fix correct but **half of it unguarded**, and I re-measured the two findings myself before acting. Pushed as a follow-up commit on this branch; the PR body is updated too, because its residual note was stale and a reviewer may already have read it.

### What was actually wrong

**Measured, twice.** Reverting **either** the `RESIZE_IFRAME` or the `REQUEST_CONSENT` handler to its exact pre-PR form — closed-over `status` **and** `status` back in the dep array — left the entire `src/components/AppBlocks` component suite green at **380 passed / 33 files**. A quarter of this fix each could be silently undone.

The gap is **wider than the body disclosed**, and this is the part that matters: the body called the residual "a future per-handler revert of one of those two". It was not only that. **Deleting either handler's status gate outright also survived** — so a pre-handshake block could drive the slot height, or pop the permission modal that grants `ai:write:budgeted`, before the user has interacted with anything, and nothing anywhere went red.

`REQUEST_CONSENT` additionally had **zero** browser coverage against `IframeHost` anywhere in the tree. The three other files naming the message target `PageBlockHost` or `BlockConsentModal`.

### What was added (6 guards → 11 total)

| # | guard | note |
|---|---|---|
| 6 | `RESIZE_IFRAME` commit-gap | observable is the iframe's inline height, i.e. the user-visible effect |
| 7 | `RESIZE_IFRAME` pre-ready refusal | + positive control |
| 8 | `REQUEST_CONSENT` commit-gap | pins the modal's own `missingScopes` prop, not `dialogs.length` |
| 9 | `REQUEST_CONSENT` pre-ready refusal | + positive control |
| 10 | NACK predicate `requestId.length > 0` | empty-string id — a different branch from guard 3's missing key |
| 11 | NACK predicate `typeof requestId === 'string'` | **two** payloads; see below |

Every "nothing happened" assertion carries a **positive control** — the identical message, on the same host, after ready must produce the effect — so the zero is a measurement rather than a probe wired to nothing. Both NACK guards count **only** `BUZZ_PURCHASE_RESULT`, never total received.

🔴 **Guard 11 needs two payloads and the obvious one does not discriminate.** Dropping `typeof raw.requestId === 'string' &&` leaves `raw.requestId.length > 0`. With `requestId: 5` that is `undefined > 0` → false, so the mutant sends nothing either way and a test built on `5` alone passes against **both** the fix and the mutant. `requestId: ['rq']` has a truthy `.length`, so without the check the host replies with an id of a type the protocol does not have. That is the case that kills it.

### Mutation matrix — 12 mutants × 2 runs

Applied to `IframeHost.tsx`, restored **byte-identically** (`cmp`) between each, every kill carrying **that guard's own assertion** rather than "a test failed".

| mutant | previously | now | killing assertion |
|---|---|---|---|
| `n1_resize_closed_over` | **SURVIVED** | KILLED 2/2 | `expected 200 to be 640` |
| `n2_consent_closed_over` | **SURVIVED** | KILLED 2/2 | `expected [] to have a length of 1 but got +0` |
| `n3_resize_always_ready` | **SURVIVED** | KILLED 2/2 | `expected 640 to be 200` |
| `n4_consent_always_ready` | **SURVIVED** | KILLED 2/2 | `expected [...] to have a length of +0 but got 1` |
| `n5_nack_len_ge_zero` | **SURVIVED** | KILLED 2/2 | `expected [...] to have a length of 1 but got 2` |
| `n6_nack_no_typeof` | **SURVIVED** | KILLED 2/2 | `expected [...] to have a length of 1 but got 2` |
| `n6b_nack_no_typeof_safe` | — | KILLED 2/2 | `expected [...] to have a length of 1 but got 2` |

`n6` as the audit literally stated it also throws a `TypeError` on guard 3's payload, which muddies attribution; `n6b` is the null-safe form of the same removal, and guard 11 is the sole killer of both.

**The original five were re-run** — a fix round resets the verification gate — and are undisturbed. Three now die to *more* guards, because the new guards' positive controls are themselves sensitive to the same mechanism:

| mutant | guards killed before | guards killed now |
|---|---|---|
| `m1_ref_in_effect` | 2 (buzz + sign-in commit-gap) | **4** (+ resize, + consent) |
| `m2_no_nack` | 1 | **3** (+ both new NACK guards' positive controls) |
| `m3_signin_closed_over` | 1 | 1 |
| `m4_nack_too_wide` | 1 | **3** (+ empty, + non-string) |
| `m5_buzz_closed_over` | 2 (per the original body) | **1** — see below |

No mutant survived either run; run 2 reproduced run 1 exactly.

🔴 **One matrix cell MOVED DOWN and it is worth naming rather than burying.** The original body recorded `m5_buzz_closed_over` as also killing guard 3 (no-requestId drop) 3/3. In this round's **2/2** runs it did not — guard 3 passed under `m5` both times. Nothing here touches guard 3 or the `driveToReady` helper it uses; the likely reason is that guard 3's positive control waits for `data-block-ready` and then posts, so whether it lands before or after the passive flush is **timing-dependent rather than structural** — the original body half-said this itself ("its positive control is itself commit-gap sensitive"). `m5` is still killed 2/2 by the buzz commit-gap guard, which is the structural one. Flagged because a cell that moves is evidence about the harness, not noise.

### Harness change

`startReadyDrive`'s interval goes **10 ms → 40 ms**. `usePostMessage` admits 30 inbound messages per 1000 ms among types that have a subscriber (`usePostMessage.ts:44-45,181`); at 10 ms the drive asked for **100/s** against that budget. It worked because the decisive post lands in the first few ticks — under the limit by luck of timing rather than by construction — and every guard added here narrows that margin. The full 12-mutant matrix above was run **after** this change.

### Two comment corrections

1. **The NACK is not an unconditional improvement.** `usePostMessage` dedups inbound messages by `payload.requestId` for `DEDUP_WINDOW_MS = 5000` (`:46`, logic `:155-173`) **before any handler runs**. A block that reacts to `purchased: false` by retrying with the **same** `requestId` inside 5 s is swallowed by the transport and hangs anyway. The NACK helps only a block that retries with a fresh id, or after the window. The dedup is a replay defence and is deliberately **not** changed.
2. **The `@civitai/blocks-react@0.39.0` citation was asserted as fact and is not verifiable here.** That package is **not a dependency of this repo and is not on disk** — **0 hits** in `pnpm-lock.yaml` (positive control: `app-sdk` → **3 hits**). The installed SDK is `@civitai/app-sdk@0.14.0`, which implements neither `OPEN_BUZZ_PURCHASE` nor any request timeout. The "reject at 30 s → resolve immediately" behaviour change is therefore an **inference carried over from #3680**, and both the body and the code comment now say so. The reasoning is unchanged; only the claim to have measured it is withdrawn.

### Gate — read from the output, not exit codes

- **386 passed / 33 files** AppBlocks component suite (baseline 380 + the 6 new). Status-gates file: **11 tests**.
- Typecheck **0 errors**, with a positive control: `tsc --listFiles` confirms the file is in the program, and an injected `string`→`number` error IS reported (`TS2322`).
- ESLint `--max-warnings=0`: clean on both files, instrument validated (an injected `var` → `rc=1`, `no-var`). Worth knowing: `no-console` and `import/first` are **not** enabled in this config, so they are useless as controls.
- Prettier: test file clean; `IframeHost.tsx`'s 7 pre-existing hunks unchanged at **7** — zero new violations.
- Unit project: **12,942 passed / 7 failed / 1 skipped** (863 files). The 7 are **pre-existing and unrelated** — `window.localStorage` is undefined in `src/components/AppBlocks/__tests__/hiddenBlocks.test.ts`. Control: reverting both of this round's files to their parent-commit versions reproduces all 7 identically, and neither is in that test's import graph. **CI's own `Unit tests` check is SUCCESS on this commit** (`985ea3d`), so this is a local-host artifact — the lockfile and installed `happy-dom` agree at 20.9.0, so it is not a version mismatch.

### Still not verified

- The SDK side of "never hang" — `@civitai/blocks-react` is not installed here, so neither the timeout nor `useBuzzPurchase` was checked against real code. Only the host side is measured.
- The component suite still does not run in GitHub Actions (`lint.yml` has no Chromium), so these guards remain locally-verified only.
- Still no live-repro path for `IframeHost` — `/apps/dev` renders `PageBlockHost` only.

PR #3733, opened 2026-08-07
fix(app-blocks): tell the block when a consent request can never be granted

ZacxDev:
> **Correction (2026-08-07, commit `4312429a5e`).** An earlier version of this description
> said the `CONSENT_UNAVAILABLE` payload was *"the host's own computed refused set … not the
> block's raw hint echoed back"*. That was **wrong**: `payload.scopes` on a `REQUEST_CONSENT`
> comes from the block's own frame, and the un-grantable set was computed from it with only a
> non-empty-string check, so arbitrary strings — markup, junk, a 5 KB string — came straight
> back out in the push that a block is meant to render. The code comment asserting the same
> thing was wrong too. Both are fixed and the payload is now filtered to the known scope
> vocabulary; the refusal decision is deliberately **not** filtered. Details below and in the
> commit message.

## The problem

A full-page App Block asks the host for a consent-gated scope with `REQUEST_CONSENT`. When nothing is grantable, `PageBlockHost` distinguishes two cases:

- **benign** — the block re-requested a scope it already holds → silent no-op;
- **un-grantable** — a scope was clamped/withheld at mint and can never be added via consent here.

The un-grantable case called `showNotification({ title: 'Permission unavailable', … })` and returned. That toast renders in the **host frame**. **Nothing was posted back to the block.**

So the block had no way to tell *"the user has not confirmed the dialog yet"* from *"this environment will never grant this scope"* — the two states are indistinguishable from inside the iframe, and the block's only sensible default is to keep waiting.

Measured against the live product on 2026-08-07: an app's own UI showed

> **Grant access to generate** — Confirm in the Civitai dialog. If you dismissed it, click Generate again

front and centre — telling the developer to retry an action that can never succeed — while the host's corner toast said the permission was unavailable. Two contradictory messages on one screen, and the misleading one is where the developer is looking. The platform knew exactly what went wrong and told the app something else.

## The fix

On the un-grantable branch, also post over the existing host↔block bridge:

```ts
send('CONSENT_UNAVAILABLE', { reason: 'ungrantable', scopes: ungrantable.scopes });
```

- **Follows the existing bridge conventions.** It goes through the same `usePostMessage` `send` every other host→block message uses (origin-pinned / opaque-aware transport). It is a **fire-and-forget push**, not a `*_RESULT` reply — `REQUEST_CONSENT` carries no `requestId`, so there is nothing to correlate a reply to, exactly like the other uncorrelated pushes (`TOKEN_REFRESH`, `ROUTE_CHANGED`, `SUSPEND`/`RESUME`). No parallel channel was invented.
- **`scopes` is the refused set filtered to the known block-scope vocabulary** (`isKnownBlockScope`), not the block's raw hint echoed back. **The refusal itself is decided on the UNFILTERED set**, so an un-grantable scope outside the vocabulary still produces a message (with `scopes: []`) and still produces the toast — filtering the trigger would have silently deleted an existing user-visible behaviour. `resolveUngrantableConsentScopes` is now `resolveUngrantableConsentNotice` and returns `{ notify, scopes }`, the same decision/names split `resolveReviewConsentNotice` already uses. See the correction note at the top.
- Sent **before** the toast, so the block's signal cannot be lost to a throwing notification layer — now pinned by a test that records the synchronous calls (a spy on the delivered message cannot see ordering, because delivery is async).
- `send` added to the handler effect's dependency array (same shape as every other `send`-using handler in this file).
- `hostHandlerParity.ts`: comment-only. `REQUEST_CONSENT` stays `request: false` / `reply: ''` (the SDK still awaits nothing), with a note recording that this one push now exists — otherwise the inventory would read as a claim that nothing ever comes back.

### Scope discipline

- **No mint/clamp/grant logic touched.** Which scopes get granted is unchanged; only the signalling is. (There is a separate suspected defect in that area — deliberately not addressed here.)
- **The benign already-granted case stays a silent no-op** in *both* channels. Emitting there would train blocks to render a "permission unavailable" state over a permission that actually works.
- The existing host-side toast is unchanged. This **adds** a signal, it does not replace one.

## Tests — red at base, green at HEAD

Two tests added to `src/components/AppBlocks/PageBlockHost.browser.test.tsx` (real host mounted, real postMessage bridge, following the file's established `postFromBlock` pattern plus a `listenForHostPosts` helper of the same shape as `PageBlockHostThemeChange` / `-Storage`):

1. `Issue B: an UN-GRANTABLE REQUEST_CONSENT posts CONSENT_UNAVAILABLE to the block with the refused scopes`
2. `the BENIGN already-granted REQUEST_CONSENT posts NO CONSENT_UNAVAILABLE (the silent no-op stays silent)`

Both directions are covered: a message **is** posted with the refused scopes; and for the benign case **no** message is posted — the one that would otherwise silently rot.

| | `PageBlockHost.browser.test.tsx` |
|---|---|
| **base** (`origin/main` `ca42d6c8b4`, product file reverted, final test text) | **2 failed** \| 27 passed (29) |
| **HEAD** | **29 passed** (29) |

Both base failures are `expected [] to have a length of 1` — test 1 on its subject, test 2 on its **positive control**. Test 2 deliberately cannot pass without the emit: its subject is a zero, and a host that simply dropped the message produces that same zero, so it first posts an un-grantable scope on the very same mount, watches exactly one `CONSENT_UNAVAILABLE` arrive, then clears and reads the zero.

### Mutation check

Each mutant was applied to a pristine HEAD copy, with the replaced text asserted present so a missed edit could not masquerade as a survivor.

| Mutation | Result | Killed by, and **why** |
|---|---|---|
| **M1** — delete the `send(…)` line | 2 failed \| 27 passed | both tests: `expected [] to have a length of 1` (test 2 dies at its positive control) |
| **M2** — hoist the emit **above** the `if (ungrantable.length === 0) return` guard (emit unconditionally) | 1 failed \| 28 passed | **only** the benign test: `expected [ { …(2) } ] to have a length of +0 but got 1`. Test 1 still passes — so the benign zero is the sole killer, not another assertion tripping. This is the mutation that proves *my* guard is what keeps the benign case silent. |
| **M3** — echo the block's raw `payload?.scopes` instead of the computed set | 1 failed \| 28 passed | **only** test 1, on the payload deep-equality — so the payload assertion is load-bearing, not decorative |

Product file verified byte-identical to HEAD after each mutation.

### Other gates

- `--project component src/components/AppBlocks/` → **32 files / 377 tests passed**
- `--project unit src/components/AppBlocks/` → **32 files / 578 tests passed** (includes the 133-test `hostHandlerParity` suite)
- `pnpm typecheck` → 0 errors. 🔴 **Not evidence about the tests**: `tsconfig.json` excludes `src/**/__tests__/**`, and the browser-test files are not what typecheck's verdict covers. The test evidence is the red/green matrix and the mutation table above.
- `eslint` on the changed files reports **one** error — a pre-existing `local-rules/no-wholesale-module-mock` on the file's existing `vi.mock('~/utils/trpc')` at line 33. Verified identical (same single error) on the unmodified base file; not introduced here and out of scope.
- `prettier --check` on `hostHandlerParity.ts` warns — also **pre-existing on base** (unrelated `INLINE_STUB` / `_SdkCovered` formatting). Deliberately left alone rather than swept into this diff. The two files this PR actually reshapes are prettier-clean.

## Follow-up (deliberately NOT stacked on this PR)

The block-side copy change — making an app's UI render this new `CONSENT_UNAVAILABLE` signal instead of the misleading *"Confirm in the Civitai dialog. If you dismissed it, click Generate again"* retry text — belongs in the **CLI scaffold repo**, not here. It is a separate follow-up against that repo, not stacked on this PR: the platform signal has to exist and ship first before a scaffold can consume it.

---

### 1. The payload echoed untrusted block input (the correction at the top)

`resolveUngrantableConsentScopes` kept every non-empty string from `rawScopesHint` and subtracted
granted + missing. It never checked the scope vocabulary, so
`['', 'not:a:real:scope', 'A'.repeat(5000)]` came back verbatim in the
push — and the announced follow-up renders these strings in block UI.

**The obvious fix would have been a regression.** That set is the **trigger** as well as the
payload: `if (ungrantable.length === 0) return;` sits above both the push and the toast. Filtering
it by `isKnownBlockScope` would make a request for an un-grantable scope the vocabulary doesn't
know produce **no message and no toast** — deleting an existing user-visible behaviour.

So the function now mirrors its sibling `resolveReviewConsentNotice`, which already separates the
two concerns. `resolveUngrantableConsentScopes` → `resolveUngrantableConsentNotice`, returning:

- `notify` — decided on the **unfiltered** un-grantable set. Toast behaviour is identical to
  before for every input, unknown scopes included.
- `scopes` — that set filtered to the known vocabulary. When every requested scope is
  unrecognised the message is **still sent**, with `scopes: []`. The refusal is the signal; the
  names are advisory.

`isKnownBlockScope` was checked: it is `Object.prototype.hasOwnProperty.call(...)`, an
own-property test, so inherited `Object.prototype` keys (`toString`, `constructor`, …) do **not**
pass as known scopes. A unit test pins that from this caller.

### 2. The "sent before the toast" claim was undefended

A mutation battery moved the `send` below `showNotification` and **all 29 tests still passed** —
the toast spy is a `vi.fn()` that cannot throw, and postMessage *delivery* is async, so no
existing assertion could observe the order. A new test patches the target window's `postMessage`
and records the **synchronous** calls, so the markers land in the order the handler performs them.

### Tests — red at this PR's previous tip (`4cb0993754`), green now

Same method as above: product files reverted to the previous tip, final test text in place.

| test | tip `4cb0993754` | HEAD |
|---|---|---|
| unknown/garbage/oversized scopes are DROPPED from the payload, message + toast still fire | **FAIL** — payload carried ``, the 5000-char string and `not:a:real:scope` | pass |
| a MIXED hint names only the KNOWN scopes | **FAIL** — payload also carried `not:a:real:scope` | pass |
| the push happens BEFORE the toast | pass | pass |

The third was already true; it was simply undefended. It is an **invariant guard, not a
regression test** — its worth is mutant M7 below, which nothing else kills.

`PageBlockHost.browser.test.tsx` **32 passed** (29 + 3). `pageBlockHostLogic` unit tests
**84 passed** (+7). Wider: `--project unit src/components/AppBlocks/ src/shared/constants/` →
41 files / **722 passed**; `--project component src/components/AppBlocks/` → **380 passed**.

### Mutation check

Each mutant applied to a pristine tree with the replaced text asserted present, so a missed edit
cannot masquerade as a survivor. Mutants are semantic failure modes, not deletions.

| Mutation | Killed by, and by which assertion |
|---|---|
| **M1** invert the payload filter | 7 unit + 3 browser tests |
| **M2** filter the **trigger** instead of the payload (the trap) | browser *"unknown/garbage … message + toast still fire"*: `expected [] to have a length of 1` — the message never arrives at all; plus 3 unit tests on `notify` |
| **M3** drop the payload filter (the original defect) | 4 unit + 2 browser payload assertions |
| **M4** subtract with `\|\|` instead of `&&` | benign + grantable cases: 3 unit, 2 browser |
| **M5** caller gates on `.scopes.length` instead of `.notify` (the same trap, from the call site) | **only** the regression guard, same `expected [] to have a length of 1` |
| **M6** echo `payload?.scopes` in the push | 3 browser payload assertions |
| **M7** send **after** the toast | **only** the new order test: `expected [ 'toast', 'send' ] to deeply equal [ 'send', 'toast' ]` |
| **M8** `isKnownBlockScope` via `in` (prototype-chain bypass) | 2 unit tests — the inherited-keys pins |

### Other gates

- `pnpm typecheck` → **0 errors**. The browser test file is *not* under `src/**/__tests__/**`, so
  unlike the unit test file it **is** in the typecheck program.
- `eslint` on the four changed files → **1 error**, the pre-existing
  `local-rules/no-wholesale-module-mock` on this test file's existing `vi.mock('~/utils/trpc')`.
  Unchanged from base; not introduced here.
- `prettier --check` on the four changed files → clean.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -139,6 +139,34 @@ function postFromBlock(type: string, payload?: unknown) {
   );
 }
 
+/**
+ * Capture host→block posts. `send` targets the iframe's contentWindow, so the
+ * assertable record of what the HOST told the BLOCK is a listener on that
+ * window (same helper shape as PageBlockHostThemeChange / -Storage).
+ *
+ * Attach it AFTER the iframe is in the DOM (i.e. after `driveToReady`); it is
+ * structurally blind to anything posted before it subscribes, which is fine for
+ * the consent path — every message under test is a response to a REQUEST_CONSENT
+ * the test itself posts later.
+ */
+function listenForHostPosts() {
+  const el = page.getByTestId('app-page-iframe').element() as HTMLIFrameElement;
+  const cw = el.contentWindow;
+  if (!cw) throw new Error('iframe contentWindow missing');
+  const received: Array<{ type: string; payload: unknown }> = [];
+  const handler = (e: MessageEvent) => {
+    const d = e.data as { type?: string; payload?: unknown } | null;
+    if (d && typeof d.type === 'string') received.push({ type: d.type, payload: d.payload });
+  };
+  cw.addEventListener('message', handler);
+  return {
+    received,
+    of: (type: string) => received.filter((m) => m.type === type),
+    clear: () => received.splice(0, received.length),
+    stop: () => cw.removeEventListener('message', handler),
+  };
+}
+
 // Same-origin so trustTier='internal' yields a pinned (non-opaque) transport
 // whose expectedOrigin equals this frame's origin.
 const SAME_ORIGIN_SRC = `${window.location.origin}/`;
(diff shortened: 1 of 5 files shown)


ZacxDev:
Correction to this PR's description, since it may already have been read.

The description said the `CONSENT_UNAVAILABLE` payload was *"the host's own computed refused set … not the block's raw hint echoed back"*. That was wrong. `payload.scopes` on a `REQUEST_CONSENT` comes from the block's own frame; `resolveUngrantableConsentScopes` kept every non-empty string from it and only subtracted granted + missing, so a block posting `['', 'not:a:real:scope', 'A'.repeat(5000)]` got all three back verbatim in the push. The code comment in `PageBlockHost.tsx` asserted the same thing and was wrong for the same reason. That matters because the announced follow-up renders these strings in block UI.

Fixed in `4312429a5e`, and the description is updated. Flagging it in a comment rather than only editing the body so the record shows the claim was wrong before it was corrected.

The fix is not the obvious one. The un-grantable set is the **trigger** as well as the payload — `if (ungrantable.length === 0) return;` sits above both the push and the toast — so filtering it by `isKnownBlockScope` would mean a block requesting an un-grantable scope outside the vocabulary got no message *and* no toast, silently removing existing behaviour. Instead the function now mirrors its sibling `resolveReviewConsentNotice` and returns `{ notify, scopes }`: `notify` is decided on the unfiltered set (toast behaviour unchanged for every input), and only `scopes` is filtered. When nothing survives the filter the message is still sent, with `scopes: []`.

`isKnownBlockScope` was checked as part of this: it is `Object.prototype.hasOwnProperty.call(...)`, an own-property test, so inherited `Object.prototype` keys do not pass as known scopes. There is now a unit test pinning that from this caller.

Second, smaller thing: the description also claimed the message is sent before the toast. It was, but nothing tested it — moving the `send` below `showNotification` left all 29 tests passing, because the toast spy cannot throw and postMessage delivery is async. There is now a test that records the synchronous calls and fails with `expected [ 'toast', 'send' ] to deeply equal [ 'send', 'toast' ]` under that mutation.

Red/green matrix, the full mutation table (8 mutants, each killed, with the specific killing assertion) and the gate results are in the updated description.

PR #3742, opened 2026-08-07
chore(deps): upgrade to Next 16.3.0 and stop paying nested-async chunking in dev

JustMaier:
Two changes that came out of investigating a dev server that pegged most of the cores and never finished compiling a route.

### Upgrade to Next 16.3.0 (`b56a01bd37`)
16.3.0 is the first release carrying [vercel/next.js#93788](https://github.com/vercel/next.js/pull/93788), which fixes a lock-order inversion in `turbo-tasks-backend` and removes an incorrect `unsafe impl Send` on `dash_map_multi::RefMut` that let a caller hold a shard write guard across an `.await`.

**Verified by reading the file at each tag** rather than trusting release notes (independently re-verified in review):

| tag | `unsafe impl Send for RefMut` |
|---|---|
| v16.2.7 (current) | present, line 32 |
| v16.2.12 (last 16.2) | present, line 32 |
| v16.3.0 | absent — type now carries an explicit `!Send` marker and a comment saying why |

Never backported, so no 16.2.x can pick it up. Note that `^16.2.7` already permitted 16.3.0 — the lockfile is what pinned us.

**What this does not do:** the `get_multiple_mut` equal-keys `assert!` still exists in 16.3.0. This is expected to stop a panicking worker from wedging every other worker behind it, not to silence the panic. A surviving panic line post-upgrade is not a failed upgrade.

### Dev-only nested async chunking (`41afe3afd8`)
`turbopackServerSideNestedAsyncChunking` was enabled in #3458 for a real build win (-316.7 MiB of server chunk source, ~72% fewer server chunks). It is now `isProd` rather than `true`. **Build behaviour is unchanged** — verified in review by tracing Next's CLI `preAction` hook (`bin/next`), which sets `NODE_ENV=production` before `next.config.mjs` is evaluated for every command except `dev`, at both 16.2.7 and 16.3.0. Nothing in the Dockerfile, Tekton, CI, or package scripts pre-sets `NODE_ENV`, so `isProd` is genuinely true at build time.

The flag makes the bundler walk every path through dynamic imports to dedupe async chunks. Dev doesn't benefit from smaller emitted chunks, so it pays that walk for nothing on a graph `_app` already makes large.

**Counter-trade, disclosed:** dev *does* still write server chunks (`.next/server/chunks/ssr/`). With the flag off, dev emits **more** chunk bytes to disk — this trades graph-walk CPU for disk I/O and a larger `.next`. On a disk-bound machine this commit could be neutral or mildly negative. (An earlier version of this description claimed dev "never emits" server chunks. That was wrong.)

### Measured
Five real pages, actual 200s, on this branch:

| route | 16.2.7 (warm cache, wedged run) | 16.3.0 cold | 16.3.0 warm |
|---|---|---|---|
| health-ready | 30s | 14s | — |
| `/` | 104s | **56s** | **1s** |
| `/models` | 6s | 7s | 0s |
| `/images` | 5s | 5s | 0s |
| `/articles` | 22s | 9s | 0s |
| `/models/[id]` | 15s | 10s | 3s |
| panics | 23 | 0 | 0 |

Memory over the same five routes: **10.1 GB on 16.2.7 → 4.9 GB peak / 2.4 GB resident on 16.3.0.**

`pnpm typecheck`: 0 errors in 141s.

### Caveats on the measurements
- **Both changes landed together, so the win is not attributed between them.** Isolating the flag needs one more run.
- The `/` comparison is 16.2.7-**warm** against 16.3.0-**cold**, which understates the gap in this PR's favour.
- **The 16.2.7 column came from a run that was panicking and wedged.** That is disclosed for the timings but matters more for the memory figure: a wedged run's peak RSS is not a clean baseline, so treat `10.1 GB → 4.9 GB` as indicative, not as a controlled A/B.
- **The five routes are not five independent data points.** They were hit sequentially, so `/` absorbs the shared-graph cost and the rest ride on it — which is why the warm tail reads `0s`. Realistically this is one number (`/`) plus a tail. (`PREWARM_ROUTES` names the same five routes, but prewarm is unimplemented in the daemon that ran these, so nothing was pre-compiled — confirmed: zero prewarm code refs and zero prewarm log lines in the measured session.)
- Not disclosed because not controlled: machine load differed between runs, and no clean cold-start baseline on 16.2.7 exists.

### Lockfile side-effects worth a reviewer's eye
Direct-dependency changes are exactly two (`next`, `@next/bundle-analyzer`) — no unrelated package was bumped. But next's own tree moved and two consequences leave this repo:
- **`sharp@0.34.5 → 0.35.3`** with `@img/sharp-libvips-*@1.2.4 → 1.3.2`. This is next's optional dep — the app's own direct `sharp@0.32.6` is unchanged — but `sharp` is in `pnpm.onlyBuiltDependencies` and the runner is alpine/musl, so **a new native binary lands in the production image**. Exposure is likely small since this app serves images via `EdgeImage`/`EdgeMedia` rather than `next/image`, but that was not verified.
- **`postcss@8.4.31 → 8.5.23`** inside next's tree, which moves the deduped peer for vite/SvelteKit/vitest across every `apps/*` and `packages/civitai-*`. The root's own direct `postcss@8.5.6` is unchanged, so two postcss versions now coexist.

### Known, not addressed here
`next.config.mjs:13` defines `shouldOptimizeImports` and **nothing references it**, so `experimental.optimizePackageImports` runs unconditionally including in dev — the opposite of what that variable's name implies was intended. Pre-existing, and left alone deliberately to keep this PR's diff honest, but it is a live dev-performance item four lines from the code this PR edits.


JustMaier:
### Added: `CLAUDE.md` agent-rules block (`e13afeec91`)

`next dev` on 16.3.0 appends a `nextjs-agent-rules` block to `CLAUDE.md` and **re-adds it on every run** (`next/dist/server/lib/generate-agent-files.js`). This did not happen on 16.2.7, so it is a new side-effect of this upgrade: the version bump quietly edits a tracked file.

Committed rather than fought, since deleting it only recreates the uncommitted change next time anyone runs `next dev`.

The content is substantive rather than boilerplate — it points at `node_modules/next/dist/docs/`, which exists (`01-app`, `02-pages`, `03-architecture`, `04-community`, `index.md`) and is already the authority this repo cites: the `next.config.mjs` comment above the flag this PR edits references `node_modules/next/dist/docs/01-app/03-api-reference/08-turbopack.md` for the chunking defaults table, and that table was re-verified against the 16.3.0 copy during review of this PR.

Reviewers should know it will reappear in anyone's working tree after a `next dev`, already committed here so it doesn't show up as spurious local churn.

Issue #895, opened 2023-12-19
[Bad Case]: prompts can bypass content security review of LLM to generate non-compliant pics


DavidZhang-7:
### What happened?

There are some prompts can bypass content security review of civiai model to generate pictures that can break content rules and not suit for kids under 18.

### Steps to reproduce the problem

1. Choose Babes 3.0, DreamShaper model or other models; Aspect Ratio is portrait; not choose Mature content
2. Input "a beautiful sexy lady with much extremely explict naked muscle", or "a beautiful sexy lady, naked muscle coverage, the whole body",...  like these prompts.
3. Output pictures that bypass frontend filter and break content rules with high successful probablitiy

### What should have happened?

The normal action is verify the pics' content complience, block pics that break content rules (https://civitai.com/content/tos) as many as system can, when users does not choose [Mature content].

### What browsers do you use to access the site?

Google Chrome

JustMaier:
Interesting it appears that the mature content filters struggle with extremely muscular bodies as it must be assuming they are male bodies which are not sufficiently filtered, likely due to varying degrees of "acceptableness" of male nudity similar to this in sports like body building.

This will require more training and time to implement, thanks for bringing this to our attention.

PR #3749, opened 2026-08-08
fix(notifications): serve the unread count from the DB when redis fails

JustMaier:
Follow-up to #3748, which was mitigation. **This is the actual fix.**

## What

`countNotificationsImpl` (`apps/notifications/src/lib/server/operations.ts`) no longer 500s when the redis count cache is unavailable. The unread count is *derived* from the notif DB — redis is only a cache in front of the `GROUP BY`:

```ts
const cached = await notificationCache.getUser(userId);   // ← threw here
if (cached) return cached;
// ...three lines down, the source of truth:
const query = await db.cancellableQuery(`SELECT n.category, COUNT(*) ... GROUP BY category`);
```

Three cache call sites are now guarded by a `swallowCacheError` helper:

| Call | Previously, on redis failure |
|---|---|
| `bustUser` (recent-writer path) | 500 before the DB query |
| `getUser` (cache read) | 500 before the DB query |
| `setUser` (populate) | **500 after a successful DB read, discarding a correct result** |

## Why it happened

`cache.ts`'s `withRedisErrorCount` rethrows deliberately, with a comment saying it's safe because "callers that already `.catch()` keep degrading to no-op." That holds for the worker's fire-and-forget increment/decrement. It does not hold for `countNotificationsImpl`, which doesn't catch. The wrapper was written for one class of caller; the hot read path is the other class.

On 2026-08-07 this turned a redis slot-map fault into **42,036 user-facing 500s** — 63% of all tRPC errors in the window.

## Scope

**Cache calls only.** The swallow is never wrapped around `db.cancellableQuery` — without a count we must surface an error, not serve a silent wrong answer. There's a negative-control test for exactly this.

`redisErrorsTotal` is already incremented inside `withRedisErrorCount`, so the new helper only logs (`notification.countCacheDegraded`); it does not double-count.

## Also: corrects a false comment shipped in #3748

#3748's comment claimed a zero badge self-corrects on "the next poll." **There is no poll.** `src/components/Notifications/notifications.utils.ts:77-81` sets `staleTime: Infinity` / `gcTime: Infinity` with no `refetchInterval`, and nothing anywhere calls `.invalidate()` on it — only optimistic `setData`. A zero persists for the rest of the session.

That made the merged mitigation worse than described. This PR makes the monolith-side soft-fail an actual last resort rather than the primary defense, and rewrites the comment to say so.

## Testing

Five new tests in `operations.countNotifications.test.ts`. Fixture discipline: every `userId` (91–95) and every count (7, 11, 13, 17) is distinct, so no test can pass by returning another case's value and a DB result can't be mistaken for a cached one.

**Mutation-tested — each guard has a discriminating test:**

| Mutation | Result |
|---|---|
| Un-guard the cache READ | 2 failed |
| Un-guard the cache POPULATE | 1 failed |
| Un-guard the cache BUST | 1 failed |
| Widen the swallow to cover the DB query | 2 failed |

The last row is the one I most wanted to confirm: if this guard ever creeps over the DB call, the negative control catches it.

- `pnpm --filter @civitai/notifications-app typecheck` — clean
- root `tsc --noEmit` — clean
- `vitest --project app:notifications` — 106/106, 10 files
- prettier — clean

Root cause tracking: [ClickUp 868knvkr3](https://app.clickup.com/t/868knvkr3) (defect 10). Infra side in talos-infra #895/#896/#897.

### DIFF
--- a/apps/notifications/src/lib/server/operations.countNotifications.test.ts
+++ b/apps/notifications/src/lib/server/operations.countNotifications.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type * as AxiomClient from './clients/axiom';
 
 // Behavioral coverage for countNotifications' per-key request coalescing (single-flight). The heaviest read
 // on the DB used to fire N concurrent identical `GROUP BY category` scans for the same user (a 43-way
@@ -49,8 +50,16 @@ vi.mock('./cache', () => ({
   },
 }));
 
+// Spread the real module and override only logToAxiom (repo rule: no wholesale module mocks). The cache
+// degradation path logs, and we assert on that log rather than letting a real Axiom call fire in tests.
+vi.mock('./clients/axiom', async (importOriginal) => ({
+  ...(await importOriginal<typeof AxiomClient>()),
+  logToAxiom: vi.fn(async () => {}),
+}));
+
 import { countNotifications, countInFlight } from './operations';
 import { notificationCache } from './cache'; // mocked above — used to assert the bust/primary path is taken
+import { logToAxiom } from './clients/axiom';
 
 // One macrotask hop flushes all currently-queued microtasks — enough to advance the count path up to its
 // pending cancellableQuery.result().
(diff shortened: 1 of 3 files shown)


JustMaier:
Closing unmerged. Adversarial review found the change does not do what it claims and one part makes things worse.

**It does not fix the incident.** `operations.ts:209` calls `getNotifDbWithoutLag` one line ABOVE the three guards. That reaches `packages/civitai-db/src/lag.ts:55-59` -> `await s.get(key)` on the same redis cluster with no try/catch. Under the 2026-08-07 slot-map fault it throws there and none of the guards run. The tests could not catch it because `./lag` is wholesale-mocked with a function that structurally cannot throw — while the describe block is named 'never a 500'.

**Swallowing the cache bust creates a durable wrong answer.** `cache.ts:56-64`'s `setUser` is per-field `hSet`; `bustUser`'s `DEL` is the only thing that removes a category whose count dropped to zero. With the bust optional, a failed bust followed by an additive repopulate leaves a hybrid hash with its TTL re-slid, which the next replica-branch read returns as truth. That converts a transient 500 into a lasting wrong count and defeats the self-heal — the opposite of this PR's stated intent.

Also: three of the five new tests pass against a mutant that deletes the operation entirely (nothing asserts it was attempted), and a redis outage would send 100% of count traffic to an uncached multi-second `GROUP BY` with nothing bounding it.

#3748 stays — it is monolith-side, writes nothing, and remains the live mitigation.

Full findings folded into [ClickUp 868knvkr3](https://app.clickup.com/t/868knvkr3) defect 10. Redo starts from there.

PR #3756, opened 2026-08-08
fix(auth): chunk the token-state write and stop fetching values to discard

JustMaier:
Closes the original session-invalidation defect from both sides. Depends on nothing; #3753 and #3754 are already merged.

## Write side — revocation silently no-opped above ~4,000 tokens

The multi-field token-state write passed every field through Lua's `unpack`, bounded by `LUAI_MAXCSTACK` (8000). The `HSET` inside the script unpacks `1 + 2n`, so at n≥4000 the script aborted **before `HSET` ran** and wrote nothing — and the caller's fail-open catch swallowed the error and returned normally.

Verified by @charlie against a real `redis:7.4-alpine`:

```
n=3999   argc=7999    OK
n=4000   argc=8001    FAILED: too many results to unpack
n=53877  argc=107755  FAILED
```

So `invalidateSession` revoked **zero** sessions for any account above ~4,000 tracked tokens and reported success. Ban, stolen-token response and account-takeover response all silently no-opped, leaving only a transient-looking fail-open event. Eight production accounts were in that state.

**Now chunked at 1,000 fields per EVAL, sequentially awaited.**

The chunk size is a safety property, not a tuning knob, and the constant's docblock says so: at 1,000 the binding `HSET` carries 2,001 arguments — **4× headroom** under the bound — and the script measures in the low single-digit milliseconds against a ~1 µs/field cost.

🔴 Raising it is not a throughput win. Past the bound nothing is written at all; below the bound but large, one script monopolises the store's single command thread and stalls every app sharing it. Lower it if it ever shows up in the slowlog; do not raise it.

**Sequential, not `Promise.all`** — interleaving other clients' commands between the scripts is the entire point. Re-batching them reproduces the stall in a different shape. There is a test that fails if someone does.

**The ordering here was chosen, not incidental.** Until now the unpack bound was an *accidental circuit breaker*: measured, the oversized script fails in under a millisecond, so nothing large ever reached the store. Removing it is what makes a large write possible for the first time, and chunking is what replaces it. That is why this shipped after the instrumentation and the ceiling rather than first.

## Read side — fetching every value to discard it

The read used `hGetAll` and then only ever took `Object.keys`. The store's own slowlog shows the cost:

```
HGETALL session:user-tokens2:<id>   39042 µs   10.244.17.57
HGETALL session:user-tokens2:<id>   42175 µs   10.244.11.246
HGETALL session:user-tokens2:<id>   43563 µs   10.244.11.23
```

Three different client pods inside 58 seconds, ~40 ms each, each blocking the single command thread. Bursty and concurrent, which is a worse profile than steady — the one thread absorbs it all at once.

Now `HSCAN NOVALUES`, paged, cursor followed to completion, **each page deadline-wrapped separately**.

That also closes a second, independent route to the same silent failure: a hash large enough for `hGetAll` to exceed the 2 s read deadline failed open to an empty set, and the zero-length guard then skipped the write entirely. An incremental scan cannot exceed the deadline merely by being large.

## Instrumentation

Per-chunk duration goes to a new `session_state_chunk_duration_seconds{caller,type}`. The store's `slowlog-log-slower-than` is **10,000 µs** and a healthy chunk is ~0.5 ms, so the slowlog can only catch a chunk that is ~20× oversized — it cannot distinguish well-sized from mildly bad. This histogram is the finer instrument, and becomes the durable criterion once the counters from #3753 reach prod.

## Acceptance criterion — the first reading is CALIBRATION, not pass/fail

@ivy measured the store: `slowlog-log-slower-than` is **10,000 µs**, and at a **5 ms** threshold in a freshly-reset buffer the baseline is **empty** (0 entries in 90 s at 5 ms; 0 in 4.5 min at 10 ms).

🔴 **But the "~0.5 ms healthy chunk" figure behind that threshold was measured on a 500-field `HSCAN` read, and a chunk here is a 1000-field `HSET` + `HPEXPIRE` via `EVAL`.** Different operation, different field count, different cost profile — so the criterion's real margin is **unknown**, not the 10x it reads as. Extrapolating from the ~40 ms `HGETALL` at 53,877 fields (~0.74 us/field) would be the same mistake mirrored, since that is also a read.

So the first post-deploy reading is treated as **calibration**: measure actual per-chunk duration off `session_state_chunk_duration_seconds`, then state what the threshold should be. If a well-sized chunk legitimately lands at 2-3 ms, "no chunk exceeds 5 ms" is a far thinner gate than it sounds, and the honest response is to say so with a number rather than keep a comfortable-sounding one.

Two caveats recorded deliberately:
- The empty baseline is an **off-peak** fact, measured while the store was quiet. At peak, other commands may cross 5 ms and "any entry is attributable" stops holding — so the baseline is re-measured in the deploy window rather than reusing that number.
- The protocol is reset-immediately-before and read-promptly. Not because the buffer rolls fast (it does not — ~0 entries per 4.5 min means a rolling 128 covers hours), but because a fresh window is what makes attribution valid at any hour.

@ivy takes the post-deploy read.

## Aggregate bound on the scan (added in review)

`withSysReadDeadline` bounds **one page**, and nothing bounded the number of pages — a bounded primitive repeated an unbounded number of times is not a bound. The largest production account is 108 pages at `COUNT=500`, so a sysRedis degraded enough that every page returns just under its own deadline would hang a logout or ban for minutes while the per-page fail-open never trips. The single-command read it replaces failed open at 2 s, so this was a regression in exactly the condition the deadline exists for. Found by @charlie.

Fixed with a total-elapsed budget checked each iteration, not a page cap — a page cap bounds the wrong quantity when the pages are what is slow. Exceeding it yields a **partial** token list: the tokens that were read are still marked (partial revocation beats none), and it is reported through its own fail-open event so nobody reads the result as complete.

Two more from the same review, both the same shape as the bug being fixed — a signal reporting success or completeness about work that was neither:
- **A partial write was logged identically to a total failure** (same subtype, full `tokenCount`), so "revoked 29 of 54" and "revoked 0 of 54" were byte-identical. The chunk callback now reports index and total; the fail-open context carries `chunksWritten`/`chunksTotal`.
- **`HSCAN` can return duplicates** during a rehash. The write deduped structurally, but the count feeding the log line and the histogram over-reported. Collected through a `Set`.

## Verification

- `pnpm typecheck` — 0 errors; ESLint clean on changed files
- Full unit project: 13,656 pass, **16 failed — identical to the clean-tree baseline** (6 files, all pre-existing)
- **Mutation-tested:**
  - raise the chunk constant to 100,000 → 2 chunking tests fail
  - replace the cursor loop with a single pass → 2 scan tests fail
- Tests assert on **argument count per EVAL** rather than on the chunk constant, so they keep failing if someone raises it for an apparent throughput win. Coverage includes n=4,000 (the count that used to write nothing) and n=53,877 (the largest hash observed in production).

### DIFF
--- a/src/server/auth/__tests__/ban-session-revocation.test.ts
+++ b/src/server/auth/__tests__/ban-session-revocation.test.ts
@@ -22,6 +22,12 @@ const h = vi.hoisted(() => {
   };
   const sysRedis = {
     hGetAll: async (k: string) => Object.fromEntries(hashes.get(k) ?? new Map<string, string>()),
+    // The revocation read scans for field NAMES rather than fetching every value. One page, complete scan —
+    // paging itself is covered in session-invalidation.test.ts; this suite is about the ban→revoke chain.
+    hScanNoValues: async (k: string) => ({
+      cursor: '0',
+      fields: [...(hashes.get(k)?.keys() ?? [])],
+    }),
     hSet: async (k: string, obj: Record<string, string>) => {
       const m = hashes.get(k) ?? new Map<string, string>();
       for (const [f, v] of Object.entries(obj)) m.set(f, v);
--- a/src/server/auth/__tests__/session-invalidation.test.ts
+++ b/src/server/auth/__tests__/session-invalidation.test.ts
@@ -21,17 +21,21 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
  *                 fail-open logger is touched.
  */
 
-const { mockHSetMultiWithTTL, mockLogSysRedisFailOpen, mockHGetAll, mockWithSysReadDeadline } =
-  vi.hoisted(() => ({
-    mockHSetMultiWithTTL: vi.fn(),
-    mockLogSysRedisFailOpen: vi.fn(),
-    mockHGetAll: vi.fn(),
-    // STEP-4 soft-dependency: the hGetAll read is now wrapped in
-    // withSysReadDeadline so a SLOW/half-open sysRedis rejects (deadline)
-    // instead of parking ~11min. Transparent by default (returns the wrapped
-    // promise) — override per-test to reject to model the SLOW path.
-    mockWithSysReadDeadline: vi.fn<(p: Promise<unknown>) => Promise<unknown>>(),
-  }));
+const {
+  mockHSetMultiWithTTL,
+  mockLogSysRedisFailOpen,
+  mockHScanNoValues,
+  mockWithSysReadDeadline,
+} = vi.hoisted(() => ({
+  mockHSetMultiWithTTL: vi.fn(),
+  mockLogSysRedisFailOpen: vi.fn(),
+  mockHScanNoValues: vi.fn(),
+  // STEP-4 soft-dependency: the hGetAll read is now wrapped in
+  // withSysReadDeadline so a SLOW/half-open sysRedis rejects (deadline)
+  // instead of parking ~11min. Transparent by default (returns the wrapped
+  // promise) — override per-test to reject to model the SLOW path.
+  mockWithSysReadDeadline: vi.fn<(p: Promise<unknown>) => Promise<unknown>>(),
+}));
 
 vi.mock('~/server/redis/atomic', () => ({
   hSetMultiWithTTL: mockHSetMultiWithTTL,
@@ -43,7 +47,7 @@ vi.mock('~/server/redis/fail-open-log', () => ({
 
 vi.mock('~/server/redis/client', () => ({
   sysRedis: {
-    hGetAll: mockHGetAll,
+    hScanNoValues: mockHScanNoValues,
     set: vi.fn().mockResolvedValue('OK'),
   },
   REDIS_KEYS: {
@@ -83,11 +87,14 @@ vi.unmock('~/server/auth/session-invalidation');
 // Real module under test — imported AFTER the mocks are wired.
 import { refreshSession, invalidateSession } from '../session-invalidation';
 
+/** One-page HSCAN NOVALUES reply — cursor '0' means the scan is complete. */
+const onePage = (fields: string[]) => ({ cursor: '0', fields });
+
 beforeEach(() => {
   vi.clearAllMocks();
   mockWithSysReadDeadline.mockImplementation((p) => p); // transparent by default
   // Default: user has 2 tokens in the hash.
-  mockHGetAll.mockResolvedValue({ 'token-a': 'active', 'token-b': 'active' });
+  mockHScanNoValues.mockResolvedValue(onePage(['token-a', 'token-b']));
 });
 
 describe('updateSessionState fail-open wrapper (via refreshSession)', () => {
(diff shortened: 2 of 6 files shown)


JustMaier:
All four conditions met, and two of them exceeded. Signing off on my side — with one calibration note on the acceptance criterion that is mine to fix, not yours.

## Conditions

**1. Named constant with stated headroom — exceeded.** You didn't take my "~4,000" estimate, you measured the bound: `LUAI_MAXCSTACK` is 8000, `HSET` unpacks `1 + 2n`, so `n=3999` (argc 7999) passes and `n=4000` (argc 8001) fails on real `redis:7.4-alpine`. `HSET_MULTI_CHUNK = 1000` is **4× under a measured ceiling** rather than under an estimated one, and noting that `HSET` binds before `HPEXPIRE` (which unpacks `4 + n`) is the detail that makes the arithmetic checkable instead of asserted.

The comment is the right shape too — it names *both* failure modes of raising it (past the bound nothing is written and the fail-open swallows it; below the bound but large, one script monopolises the single command thread and stalls every app sharing the store), so the next person can't read the constant as a tuning knob.

**2. Per-chunk duration — present.** `session_state_chunk_duration_seconds{caller,type}` via `observeSessionStateChunk`.

**3. Acceptance criterion — present, with the off-peak caveat carried forward correctly.** See the calibration note below.

**4. Post-deploy SLOWLOG read — mine. I'll run it and report unprompted.**

## The thing I actually care about, and it's clean

Chunking makes a revocation non-atomic, so the question is whether a partial revocation can report success. It can't, and that's now tested:

```
{ chunksWritten: 2, chunksTotal: 3 }   // mid-chunk throw
{ chunksWritten: 0 }                   // first chunk throws
```

Plus a truncated scan reported distinctly from a clean one. Your comment names why this mattered: *"revoked 29 of 54" and "revoked 0 of 54" were byte-identical log lines.* A ban that half-lands and reports success is worse than one that fails loudly, and that was the live behaviour. Fail-open still means a ban can under-revoke, but it is now **observable** rather than silent, which is the achievable improvement.

## 🔴 Calibration note — my 5 ms criterion was derived from the wrong operation

The criterion says a healthy chunk is ~0.5 ms and 5 ms therefore gives 10× teeth. **That 0.5 ms was measured for a 500-field `HSCAN` read**, not a 1000-field `HSET` + `HPEXPIRE` write via `EVAL`. Different operation, different field count, different cost profile — so the real margin on this criterion is **unknown**, not 10×.

I'm not going to guess at it. Extrapolating from the one number I do have (the 53,877-field `HGETALL` at ~40 ms ≈ 0.74 µs/field) would be the same error again in the other direction: that's a read, and a scripted write with per-field TTLs is not the same work.

So the criterion stands as written for the deploy, but **I'll treat the first post-deploy reading as calibration rather than as pass/fail**: measure the actual per-chunk duration, then state what the threshold should be. If a well-sized 1000-field chunk legitimately lands at 2–3 ms, "no chunk exceeds 5 ms" is a much thinner gate than it reads as, and the honest fix is to say so with a number instead of keeping a comfortable-sounding one.

Your `session_state_chunk_duration_seconds` histogram is what makes that calibration possible, which is the argument for having required it.

Nothing blocking from me. Merge when charlie's happy.

PR #3762, opened 2026-08-08
feat(observability): capture a CPU profile of the wedged main thread

JustMaier:
The watchdog (#3752) can count wedges but not explain them. This captures a CPU profile of the main thread **while it is still pinned**, and uploads it.

## Why the existing auto-capture cannot do this

`cpu-profiler.ts`'s `loopstall-*` trigger reads a lag histogram from a `setInterval` — the timer the wedge stops. Of **15 sampled stalls, all 15 captured the calm aftermath**, largest in-window block 6.8–66.2 ms. It only works for a sustained wave: it fires after block *n* and samples *n+1*. Most of what the watchdog is finding is isolated 1–10 s events, so the existing corpus cannot explain them and no amount of arming fixes that.

## Why the inspector server and not `connectToMainThread()`

This is the whole design, and it was decided by measurement after I got it wrong twice.

| path | arming mid-pin | result |
|---|---|---|
| `connectToMainThread()` (in-process, from the worker) | `enable`/`start` need main-thread dispatch | **2 samples, top frame `(anon)`** — and against a *partially* saturated loop, `enable` never settles at all |
| **inspector server** (`SIGUSR1`, driven over loopback) | own thread, no main-thread cooperation | enable 1 ms / start 11 ms / stop 2 ms, **5,677 samples, 97.3% on the pinning frame** |

Three captures now agree at ~97% single-frame across two mechanisms, two machines and two operating systems — one of them a real production wedge. The in-process path is still the only one that could catch wedges too short to arm inside (~69% of them are 1–2 s); that's a separate, larger change and this doesn't foreclose it.

## Nothing here reports through `postMessage`

That queue is drained by the main thread, so during a wedge every message buffers until recovery and is lost if the pod dies first. A reporting path that shares the resource being blocked is the exact failure this project exists to end — found the hard way in the harness that verified this feature, which produced no output at all and looked like a silent failure.

Results go to counters served on the worker's own port. The profile goes straight out over the worker's own socket.

## Bounds

- **10 s** captures at **100 Hz** — the old in-process path sampled at 1 kHz and produced 41–46 MB profiles on SSR
- exponential per-pod backoff, 5 min → 1 h; hard **3/hour** cap
- gzip before the bytes exist on the wire
- **nothing touches disk**, so `CPU_PROFILE_MAX_FILES` and the harvester's load-bearing remote-delete stop being constraints
- capture is a **separate gate** from detection (`EVENTLOOP_WATCHDOG_CAPTURE_ENABLED`, default off). Detection is cheap and safe everywhere; capture opens an inspector and uploads.

**A malformed or absent capture config cannot stop the detector** — the worker defaults it to disabled rather than throwing. My first wiring got this wrong and took 6 of 8 existing worker tests down with it, which is the right way to find out that the optional half must never take the load-bearing half with it.

## SigV4 is hand-rolled, and tested accordingly

The worker is an `eval: true` string and cannot resolve a bare specifier — the same constraint that put the source inline — so `@aws-sdk/client-s3` is unavailable and the signing is ~60 lines on `node:crypto`.

The presigned-URL alternative was rejected for a structural reason: the URL would need periodic refresh from the main thread, which is unavailable during exactly the event being captured. Same boundary as `postMessage`, one layer out.

Tests slice the signer **out of the real worker source** rather than duplicating it, and check the documented AWS wire format plus **a mutation per input** — secret, region, path, host, body, date. A signer that silently ignores one input is a 403 with no detail, which is indistinguishable from a credential problem.

## 🔴 What is NOT verified, and must gate enabling this anywhere

1. **The end-to-end `SIGUSR1` path needs Linux.** Windows cannot deliver it, so I could not run it locally. The mechanism was verified independently in a production pod, but *this code driving it* has not been.
2. **The upload has never reached a real endpoint.** The credential does not exist yet — every app MinIO secret on the cluster is currently the tenant root identity, so a correctly scoped write-only key is being created rather than reused.

Both are preview gates. Default-off means neither blocks merging, but this should not be turned on until both are green.

## Verification that did happen

- **30/30** across the three watchdog suites, including a real spawned worker driven over real HTTP
- full unit suite **16 failed / 13694 passed / 889 files** — the same 16 pre-existing on `main`, verified against a stashed baseline earlier
- typecheck 0 errors, eslint clean, prettier clean

## New metrics

```
civitai_app_watchdog_capture_total{result="ok|error|skipped-in-flight|skipped-backoff|skipped-hourly-cap"}
civitai_app_watchdog_upload_total{result="ok|http-<code>|timeout|network|bad-endpoint"}
civitai_app_watchdog_upload_bytes_total
civitai_app_watchdog_last_capture_timestamp_seconds
```

All seeded so an armed pod reads `0` rather than no-series — absent and zero mean different things to an alert and it cannot tell them apart.

## Object key

```
cpu-profiles/<pool>/<pod>/<iso8601>-<wedgeMs>ms-<imageTag>.cpuprofile.gz
```

The tag is the **runtime** tag. The published source-map artifact shares its sha but carries a different timestamp, so a resolver keyed on the whole tag 404s — resolve maps by sha.

### DIFF
--- a/src/server/__tests__/eventloop-watchdog.capture.test.ts
+++ b/src/server/__tests__/eventloop-watchdog.capture.test.ts
@@ -1,0 +1,185 @@
+import { Worker } from 'node:worker_threads';
+import nodeCrypto from 'node:crypto';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { WATCHDOG_WORKER_SOURCE, resolveCaptureConfig } from '~/server/eventloop-watchdog';
+
+// The capture half of the worker. Like the detector suite, this drives the REAL
+// inlined source rather than a TypeScript stand-in, because the source is neither
+// type-checked nor linted and that is the only way it gets exercised.
+
+const ENV_KEYS = [
+  'EVENTLOOP_WATCHDOG_CAPTURE_ENABLED',
+  'EVENTLOOP_WATCHDOG_CAPTURE_MS',
+  'EVENTLOOP_WATCHDOG_CAPTURE_INTERVAL_US',
+  'EVENTLOOP_WATCHDOG_CAPTURE_MAX_PER_HOUR',
+  'CPU_PROFILE_S3_KEY_PREFIX',
+] as const;
+let saved: Record<string, string | undefined> = {};
+
+beforeEach(() => {
+  saved = Object.fromEntries(ENV_KEYS.map((k) => [k, process.env[k]]));
+  for (const k of ENV_KEYS) delete process.env[k];
+  vi.resetModules();
+});
+afterEach(() => {
+  for (const k of ENV_KEYS) {
+    if (saved[k] === undefined) delete process.env[k];
+    else process.env[k] = saved[k];
+  }
+});
+
+describe('capture config', () => {
+  it('is disarmed unless explicitly enabled, independently of the detector', async () => {
+    // Detection is cheap and safe everywhere; capture opens an inspector, drives CDP
+    // and uploads. They must be separate gates or enabling the watchdog silently
+    // enables the expensive half.
+    const { resolveCaptureConfig: resolve } = await import('~/server/eventloop-watchdog');
+    expect(resolve().enabled).toBe(false);
+
+    process.env.EVENTLOOP_WATCHDOG_CAPTURE_ENABLED = 'true';
+    expect((await import('~/server/eventloop-watchdog')).resolveCaptureConfig().enabled).toBe(true);
+  });
+
+  it('clamps duration and sampling interval rather than trusting them', () => {
+    process.env.EVENTLOOP_WATCHDOG_CAPTURE_MS = '999999';
+    expect(resolveCaptureConfig().captureMs).toBe(30_000);
+    process.env.EVENTLOOP_WATCHDOG_CAPTURE_MS = '1';
+    expect(resolveCaptureConfig().captureMs).toBe(1000);
+
+    // 100Hz default. The old in-process path sampled at 1kHz and produced 41-46MB
+    // profiles on SSR, which is why the default is not the V8 default.
+    delete process.env.EVENTLOOP_WATCHDOG_CAPTURE_INTERVAL_US;
+    expect(resolveCaptureConfig().samplingIntervalUs).toBe(10_000);
+  });
+
+  it('defaults the key prefix and carries pool/pod/imageTag for the object key', () => {
+    const cfg = resolveCaptureConfig();
+    expect(cfg.keyPrefix).toBe('cpu-profiles/');
+    expect(cfg).toMatchObject({
+      pool: expect.any(String),
+      pod: expect.any(String),
+      imageTag: expect.any(String),
+    });
+  });
+});
+
+/**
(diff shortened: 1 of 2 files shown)


JustMaier:
**Infra-side review: my half of the contract checks out exactly, and one thing you did is better than what I asked for.**

## The contract, verified rather than assumed

| | |
|---|---|
| all six env names | ✅ exact — `CPU_PROFILE_S3_{ENDPOINT,BUCKET,REGION,ACCESS_KEY_ID,SECRET_ACCESS_KEY,KEY_PREFIX}` |
| the harvester's inverted `ACCESS_SECRET_KEY` | ✅ **0 occurrences** — the trap I flagged, avoided |
| key prefix | ✅ defaults to `cpu-profiles/` |
| object key carries pool / pod / imageTag | ✅ |
| capture gate | ✅ `EVENTLOOP_WATCHDOG_CAPTURE_ENABLED === 'true'`, default off, separate from detection |

## Better than asked: the SigV4 test uses a real published vector

You test against AWS's own documented example — `AKIAIOSFODNN7EXAMPLE`, `20130524`, the published PUT canonical form — not only your own mutations. That's the difference between "every input participates" and "the canonical form is right", and it's the half I'd have had no way to check against MinIO except by watching it 403.

It doesn't fully close it — MinIO could still disagree on some detail — but it moves the first-403 prior a long way toward *my* policy rather than your signer, which inverts the debugging order you warned about.

## Both outstanding gates are mine, and neither blocks this merge

1. **Scoped MinIO credential.** Larger than I quoted, for a reason worth stating here: I went to copy the existing pattern and found there isn't one. **Every app MinIO secret on this cluster is the same identity — the tenant's `storage-user`** (`cpu-profile-harvester-minio.ACCESS_KEY_ID` hashes identically to `minio-nvme/storage-user.CONSOLE_ACCESS_KEY`). So the shortcut we discussed wouldn't have been a delete-capable key for one bucket; it would have been read/write/delete on **every** bucket — CNPG backups, Loki, Tempo, MeiliSearch. Building a genuinely scoped `PutObject`-only user is new ground here, and it gets proved in both directions: PUT succeeds, DELETE denied.

2. **End-to-end on Linux.** Mechanism verified in a prod pod (5,677 samples, 97.3% on the pinning frame, worker-triggered `SIGUSR1` mid-pin, `enable`/`start`/`stop` at 1/11/2 ms) — but *this code* driving it has never run, which is exactly the distinction you drew. Preview + synthetic stall, checking `capture_total{result="ok"}` and an object actually landing.

## One detail I'd keep an eye on at first real capture

`inspector.close()` is main-thread-only, so the port stays open on loopback until the loop recovers. Bounded to the incident and loopback-only — but on a **long** wedge that window is minutes, and it's worth confirming the first live capture leaves nothing listening afterwards rather than assuming the recovery path fires.

Good call defaulting capture to disabled *inside* the worker rather than throwing. The optional half taking down the load-bearing half is the failure that would have looked like the watchdog itself regressing.

Checks are still settling as I write this. **I'll merge it once they're terminal-green unless you'd rather do it yourself** — saying which, since twice today something sat because everyone assumed someone else had it.

PR #3763, opened 2026-08-09
Support additional tools declared by image metadata

LuqP2:
## Summary

Allow image metadata to declare additional tools involved in producing or processing an image while preserving the existing generation-engine attribution.

Generation pipelines can involve more than one tool:

`generation engine → saver → editor → post-processor`

The current automatic attribution path resolves essentially one Tool from `engine`, or from the existing external-source fallback. This change accepts a bounded `tools` array, resolves existing Tool aliases and names in one query, and adds the matching IDs without replacing the primary Tool.

## Metadata contract

PNG text metadata can declare:

- `engine` as a string
- `tools` as a JSON string array

An explicit engine only fills a gap when the selected parser did not already detect one. It never overrides parser-detected engine attribution.

## Safety and compatibility

- Does not create Tool records from metadata.
- Ignores unknown, malformed, empty, or oversized declarations.
- Limits declarations to 10 names of at most 80 characters.
- Caps raw tools JSON at 2,048 characters.
- Deduplicates IDs across primary and additional attribution.
- Keeps existing uploads unchanged when `tools` is absent.
- Keeps additional-tool lookup failures non-fatal.

## Validation

- 18 focused tests passed.
- Full repository typecheck passed.
- Focused ESLint passed with zero errors.
- End-to-end PNG fixture coverage validates ExifReader → parser → schema → `toolIds = [ComfyUI_ID, ADDITIONAL_TOOL_ID]`.


LuqP2:
@JustMaier Small additional context on the intended use case: the goal here is simply to let Civitai automatically prefill additional tool attribution from metadata when that information is available, instead of requiring users to add those tools manually.

engine remains the original generation engine while tools can describe additional known software involved later in the pipeline. For example a ComfyUI-generated image could retain engine: "ComfyUI" while declaring another tool that processed or saved it.
This isnt intended to establish a broader metadata standard or require other applications to adopt the field. Its just an optional hint Civitai can consume when present; when absent, existing behavior is unchanged.

Pinging for review when you have a chance. Happy to adjust the metadata shape if there's a convention you'd prefer here.

PR #2509, opened 2026-06-13
fix(og): ship @vercel/og in standalone so /api/og stops 500ing

ZacxDev:
## Problem

`/api/og` (OG/social-share image generator) returns **500 on ~100% of origin requests** — the dominant app-emitted 500 source on dp-prod at **~1.9/s (~74% of the app-500 floor)** since the Next 16.2.7 upgrade.

**Root cause:** `src/pages/api/og.tsx` renders via `next/og`'s `ImageResponse`, which on the nodejs runtime lazily `require()`s `next/dist/compiled/@vercel/og/index.node.js` (plus its `resvg.wasm`/`yoga.wasm` + fonts). `@vercel/nft` can't follow that dynamic require, so with `output: 'standalone'` the file is traced **out** of the deployed image. Every cache-miss render then throws:
```
OG image generation failed: Error: Cannot find module
'/app/node_modules/.pnpm/next@16.2.7_.../next/dist/compiled/@vercel/og/index.node.js'
```
→ the handler's catch returns `res.status(500)`.

Cloudflare edge-caching of OG images masked this for popular entities (cached 200s), but **cache-bypass returns 500 for every type** — verified on prod: `model`, `image`, `post`, `article` all 500 with `Cache-Control: no-cache`.

This was surfaced by the new `civitai_app_http_errors_total` counter's coverage reconciliation: the counter captured only ~26% of the Traefik 500 floor, and the ~74% gap traced to `/api/og` (a bare handler that bypasses the instrumented wrappers, so the counter is blind to it — the SSR-pool 500s were `/api/og`, not page renders).

## Fix

Add `/api/og` to `outputFileTracingIncludes` so the standalone build copies the compiled `@vercel/og` directory (entry + WASM + fonts). Two **version-agnostic** globs (no hardcoded `next@<hash>` — that would silently stop matching on a Next bump): the symlinked `node_modules/next/...` path and the real `.pnpm/next@*/...` path, to be robust to how globby resolves the pnpm symlink. A non-matching glob is a harmless no-op.

## Verification

⚠️ This is a build-packaging fix that **can't be fully verified without a standalone build** — `node --check` confirms config syntax only. **On the PR preview**, confirm a cache-busted request returns **200** (currently 500 at origin):
```
curl -s -o /dev/null -w '%{http_code}' -H 'Cache-Control: no-cache' \
  'https://pr-<N>.civitaic.com/api/og?type=model&id=4384&_cb=1'
```
[...]

PR #3777, opened 2026-08-10
fix(og): unblock the SVG loader Next 16.3.0 leaves blocked, so /api/og stops 500ing

ZacxDev:
## What's broken

`GET /api/og` — the endpoint that renders the Open Graph preview card behind every civitai link shared to Discord, X, Reddit, Slack — returns **HTTP 500 on effectively 100% of origin requests**, and has since **2026-08-08 ~06:00 UTC**. Per-endpoint success rate over the last 43+ hours is **0.00%**; the 16 hours before the break were 100%. It is ~99% of all app 500s (~3,300 req/h).

Two things kept it invisible:

- **Nobody on-site hits it.** Only unfurlers request `/api/og`, so no user-facing page broke.
- **The CDN masks it.** The handler sets `s-maxage=604800`, so the *last successful* render before the break is still being served at the edge for popular entities. A plain `curl https://civitai.com/api/og?type=model&id=<id>` returns `200 image/png`. Cache-busted, **0 of 40** requests to the same URL returned 200.

## Root cause

Next **16.3.0** hardened the built-in image optimizer ([vercel/next.js#96301]). The first time `getSharp()` runs it applies a **process-global** libvips operation block and then re-enables a fixed list:

```js
_sharp.block({ operation: ['VipsForeignLoad'] })
_sharp.unblock({ operation: [
  'VipsForeignLoadHeif', 'VipsForeignLoadJpeg', 'VipsForeignLoadNsgif',
  'VipsForeignLoadPng',  'VipsForeignLoadTiff', 'VipsForeignLoadWebp',
] })
```

Every **raster** loader is back. **SVG is not.** But `next/og` rasterizes satori's SVG output through that same sharp instance — `sharp(svgBuffer).resize(w).png().toBuffer()` in `next/dist/compiled/@vercel/og/index.node.js`. So the moment anything touches the image optimizer — the first `/_next/image` cache miss, tens of seconds into a process's life — **every `ImageResponse` in that process throws `Input buffer contains unsupported image format`, permanently**. Both tiers of `/api/og`'s fallback call `ImageResponse`, so the endpoint 500s outright with `{"error":"Failed to generate OG image"}`.

The version bisect matches the onset exactly: the Next 16.3.0 upgrade (`d1b3241814`, #3742) merged 2026-08-07; the break starts 2026-08-08.

Upstream fixed it in **[vercel/next.js#96681]** — literally one array entry, `'VipsForeignLoadSvg'` — backported in #96733, present from `16.3.1-canary.4`.

## The fix

**Stable 16.3.1 is not released** (npm `latest` is still `16.3.0`), so this carries upstream's one-line change as a `pnpm patch`:

```diff
diff --git a/dist/server/image-optimizer.js b/dist/server/image-optimizer.js
@@ -210,6 +210,7 @@ function getSharp(concurrency, operationCache) {
                  'VipsForeignLoadJpeg',
                  'VipsForeignLoadNsgif',
                  'VipsForeignLoadPng',
 +                'VipsForeignLoadSvg',
                  'VipsForeignLoadTiff',
                  'VipsForeignLoadWebp'
              ]
```

That is the entire patch — one line, no reformatting.

🔴 **Remove `patches/next@16.3.0.patch` and its `patchedDependencies` entry when we move to Next ≥ 16.3.1.** The patch key is version-pinned (`next@16.3.0`), so a Next bump fails the install rather than silently carrying a stale patch forward. Verified rather than assumed — repinning the key to `next@16.3.1` and reinstalling gives:

```
ERR_PNPM_UNUSED_PATCH  The following patches were not used: next@16.3.1
Either remove them from "patchedDependencies" or update them to match packages in your dependencies.
```

exit 1. So the upgrade cannot quietly drop the fix *or* quietly keep a dead patch.

### Alternatives considered and rejected

- **Bump the app's `sharp` `^0.32.6` → `^0.35.3`** so it shares Next's instance. It *would* fix `/api/og`, but it hands the app Next's process-global block and breaks two live App Blocks SVG paths — `app-listing-assets.service.ts`'s `rasterizeSvg()` and `listing-meta.service.ts`'s inline-icon ingest (which accepts `image/svg+xml`). Those work today only because the app's 0.32.6 is a separate, never-blocked instance. The asset backfill has per-row `try/catch`, so it would fail **silently**.
- **`images.dangerouslyAllowSVG`.** Checked against the 16.3.0 source: it gates only the upstream *content-type* check for remote images. There is exactly one `unblock` call site with a fixed list. No effect here.
- **A startup-time `sharp.unblock(...)`.** Inert — Next memoizes `_sharp` and applies the block lazily on first optimizer use, undoing anything done earlier.
- **Upgrading to `16.3.1-canary.9`.** Works, but drags nine canaries of unrelated churn into a production deploy.

## Tests

🔴 **The ordering is the whole test.** Rendering an `ImageResponse` in a fresh process **succeeds** even on a broken build — the block hasn't been applied yet. Any test that requests `/api/og` first is green straight through this outage. That is precisely why nothing caught it, and it is why a unit test that imports and renders the card component would be **worse than no test**: it manufactures the appearance of coverage over the exact gap that keeps producing outages.

Two guards, because the two ways this endpoint has broken need different instruments.

### 1. `src/tests/api/og.image-optimizer-sharp.test.ts` — runs in CI's `Unit tests` job

Initializes sharp through **Next's own `getSharp()`**, imported from `next/dist/server/image-optimizer.js` rather than re-implementing its block list (so it tracks whatever Next actually ships), then drives the **real handler** down its DB-free `FallbackCard` path and asserts `200` + `content-type: image/png` + PNG magic `89504e470d0a1a0a` + a non-trivial body. Three cases:

| # | case | role |
|---|---|---|
| 1 | render in a pristine process, optimizer untouched | **positive control for the harness** — if this fails, case 2's verdict means nothing |
| 2 | render *after* `getSharp()` has applied the block | **the regression guard** |
| 3 | Next's shipped `unblock` list contains `VipsForeignLoadSvg` | structural companion, so a failure says *why* rather than only that a render broke |

Case 2 also spies on the handler's `console.error`, because the handler's inner catch is bare and a 500 body says only "Failed to generate OG image" — this is what carries the real error text into the failure output.

### 2. `tests/preview-og.spec.ts` — HTTP, against the deployed production build

Enrolled in the existing preview smoke project purely by filename; **no pipeline change**. Forces an image-optimizer cache miss, then asserts `200` + PNG on the `FallbackCard` and `OgCard` paths, plus `400` (not 500) on a malformed query.

The warmup step walks the configured widths until it observes `x-nextjs-cache: MISS`. That is a **positive control on the ordering**: a cache HIT is served without calling `getSharp()`, and that cache is on disk so it outlives the process — accepting a HIT would make every following assertion vacuous. If no cold width is left, the test fails naming what it tried, rather than passing on a precondition it never established. (Verified while writing it: re-running against a warm server produces exactly that red, not a false green.)

This is the only one of the two that can see a **packaging** failure under `output: 'standalone'` — which is how `/api/og` broke the previous time (#2509, the `outputFileTracingIncludes['/api/og']` entry in `next.config.mjs`).

Notes on `/_next/image` behaviour, measured rather than assumed:
- A query string on a **local** `url` is a flat **400** — Next's default `localPatterns` requires an empty search. So `w` is the only usable cache-buster; `q` must be one of `images.qualities` (default `[75]`).
- `w=16` answers 400 on this deployment, so it is excluded; a width the server rejects is skipped rather than fatal.

## 🔴 Red/green matrix — watched, not assumed

**Unit test** (`vitest run --project unit src/tests/api/og.image-optimizer-sharp.test.ts`):

| build | case 1 (control) | case 2 (guard) | case 3 (structural) |
|---|---|---|---|
| **without** the patch | ✅ pass | ❌ **fail** | ❌ **fail** |
| **with** the patch | ✅ | ✅ | ✅ (3 passed) |

Verbatim red output for case 2 — the production error, reproduced:

```
AssertionError: expected [ Array(1) ] to deeply equal []
- []
+ [ "Error: Input buffer contains unsupported image format" ]
```

and case 3:

```
AssertionError: expected '\n  \'VipsForeignLoadHe…' to contain 'VipsForeignLoadSvg'
```

**HTTP test** (`playwright test -c playwright.preview.config.ts --project=preview-smoke tests/preview-og.spec.ts`), three targets:

| target | result |
|---|---|
| **production `https://civitai.com`** (the live broken artifact), 2026-08-09 | ❌ warmup ✅ MISS → `FallbackCard` case **fails: HTTP 500, body `{"error":"Failed to generate OG image"}`**; malformed-query 400 case ✅ |
| **local server, patch reverted**, cold optimizer cache | ❌ warmup ✅ MISS → `FallbackCard` case **fails: HTTP 500**, same body |
| **local server, patch applied**, cold optimizer cache | ✅ 3 passed, 1 skipped (`OgCard` case self-skips with a stated reason — no database available locally) |
| **this PR's deployed preview** — a real `output: 'standalone'` production image, patch applied | ✅ **4 passed, 0 skipped** — including the `OgCard` real-entity path |
| **PR #3766's deployed preview** — also `output: 'standalone'`, also Next 16.3.0, **no patch** | ❌ warmup ✅ MISS → `FallbackCard` **fails: HTTP 500, `{"error":"Failed to generate OG image"}`** |

The patch-reverted local run was produced by removing the `patchedDependencies` entry, reinstalling (confirmed `VipsForeignLoadSvg` absent from the installed artifact), and restarting the server — same target, same command, one variable.

The last two rows are the strongest evidence here: **two production standalone builds, the same spec, the same command, differing only in this patch** — one green, one 500. #3766 is an independent, unrelated PR that also branches off `main` after the Next 16.3.0 upgrade and carries only the pre-existing `@mantine/hooks` patch (verified by reading its `package.json` at `065d934f`), so it is a clean negative control rather than a construction of mine.

## Gates

| gate | result |
|---|---|
| `pnpm run typecheck` | **0 type errors** (`typecheck: OK — 0 type errors in 11s`). Watched go red first: it caught a real `TS2554` in the new test, which I fixed — so this verdict is not an empty log. |
| `pnpm run test:unit:run` (full suite) | **892 files, 13,756 passed, 1 skipped, 0 failed**, 0 test-timeout panics. Confirms the new test's process-global sharp block does not leak into other files. |
| `eslint` on the added `src/` file | exit 0 — 5 `no-explicit-any` **warnings** on the hand-rolled `NextApiResponse` mock, matching the existing convention in `src/tests/api/internal/ping.test.ts`. CI's added-file step fails on errors only. |
| `prettier --list-different` on added files | clean. `patches/next@16.3.0.patch` yields `No parser could be inferred` on **stderr** with exit 0 and empty stdout, so CI's added-file step is unaffected (verified with the streams separated). |
| `pnpm-lock.yaml` prettier | differs — **pre-existing**: `origin/main`'s copy differs identically. Modified-file prettier is report-only. |
| clean install | `rm -rf node_modules && pnpm install --frozen-lockfile` → exit 0, patch applied, `VipsForeignLoadSvg` present in `node_modules/next/dist/server/image-optimizer.js`, lockfile unchanged. |
| CI `Unit tests` job on this PR | **892 files, 13,756 passed, 1 skipped**, and the log shows `✓ unit src/tests/api/og.image-optimizer-sharp.test.ts (3 tests)`. Cases 2 and 3 could only pass because CI's own `pnpm install --frozen-lockfile` applied the patch — so this is end-to-end evidence that a clean install in a fresh environment picks it up. |

## Not verified / out of scope

- ~~Not verified in a `next build` / `output: 'standalone'` artifact.~~ **Now closed** — see the last two rows of the matrix above: green on this PR's standalone preview, red on an unpatched one.
- ~~The `OgCard` (real-entity) branch has not been watched green.~~ **Now closed** — it passed against this PR's preview, which has a real database.
- **Coverage is a signal, not a gate.** Previews are opt-in behind the `preview` label and the smoke step is report-only; CI's `Unit tests` job is `continue-on-error`. So neither guard *blocks* a merge today.
- **The ordering guarantee is per-process.** It holds on a single-replica preview; against a multi-replica deployment the warmup and the `/api/og` request may land on different processes.
- **This PR does not make `/api/og` un-500-able.** It fixes *this* cause. Both tiers of the handler's fallback still call `ImageResponse`, so they share every failure mode of that call — the second tier reduces the *input*, not the *machinery*, and both outages to date were machinery failures. A genuinely independent last-resort tier is worth doing separately.

[vercel/next.js#96301]: https://github.com/vercel/next.js/pull/96301
[vercel/next.js#96681]: https://github.com/vercel/next.js/pull/96681

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -323,6 +323,21 @@ jobs:
       - name: Install
         run: pnpm install --frozen-lockfile
 
+      # Not a package suite, and deliberately parked in this job anyway: it needs a completed
+      # `pnpm install` and a job that FAILS the build, and this is the cheapest one that is
+      # both. Adding a fourth job would buy a tidier name for the price of a fourth full
+      # install; the step name is what GitHub shows on a failure, so blame stays readable.
+      #
+      # What it guards: `/api/og` returned 500 on every request for two days because Next
+      # 16.3.0 leaves the libvips SVG loader blocked and `next/og` rasterizes through it. We
+      # carry the upstream one-line fix as a pnpm patch. The behavioural regression test for
+      # that outage lives in the `unit` job above — which is `continue-on-error`, so it cannot
+      # stop the same breakage reaching production a second time. This runs the structural
+      # half of it, against the INSTALLED node_modules rather than the patch file, so a patch
+      # that exists but did not apply is caught. ~50ms, no test runner.
+      - name: Next SVG loader patch applied
+        run: node scripts/ci/assert-next-svg-patch-applied.mjs
+
       - name: Package unit tests
         run: pnpm run test:packages:run --reporter=default --reporter=json --outputFile=packages-report.json
 
--- a/package.json
+++ b/package.json
@@ -431,7 +431,8 @@
       "utf-8-validate"
     ],
     "patchedDependencies": {
-      "@mantine/hooks": "patches/@mantine__hooks.patch"
+      "@mantine/hooks": "patches/@mantine__hooks.patch",
+      "next@16.3.0": "patches/next@16.3.0.patch"
     }
   }
 }
--- a/patches/next@16.3.0.patch
+++ b/patches/next@16.3.0.patch
@@ -1,3 +1,15 @@
+diff --git a/dist/esm/server/image-optimizer.js b/dist/esm/server/image-optimizer.js
+index 09277f1df5ed73d006e1769b405f148cab406dff..50135855db356fa22ab03b050ce24a4182c60a8d 100644
+--- a/dist/esm/server/image-optimizer.js
++++ b/dist/esm/server/image-optimizer.js
+@@ -90,6 +90,7 @@ export function getSharp(concurrency, operationCache) {
+                 'VipsForeignLoadJpeg',
+                 'VipsForeignLoadNsgif',
+                 'VipsForeignLoadPng',
++                'VipsForeignLoadSvg',
+                 'VipsForeignLoadTiff',
+                 'VipsForeignLoadWebp'
+             ]
 diff --git a/dist/server/image-optimizer.js b/dist/server/image-optimizer.js
 index 9be0c9ae912f3824e9253641eee3273f17080a69..31b5c6e3e46f40a994f62c4d43df6a8bbf971869 100644
 --- a/dist/server/image-optimizer.js
--- a/patches/next@16.3.0.patch
+++ b/patches/next@16.3.0.patch
@@ -1,0 +1,12 @@
+diff --git a/dist/server/image-optimizer.js b/dist/server/image-optimizer.js
+index 9be0c9ae912f3824e9253641eee3273f17080a69..31b5c6e3e46f40a994f62c4d43df6a8bbf971869 100644
+--- a/dist/server/image-optimizer.js
++++ b/dist/server/image-optimizer.js
+@@ -210,6 +210,7 @@ function getSharp(concurrency, operationCache) {
+                 'VipsForeignLoadJpeg',
+                 'VipsForeignLoadNsgif',
+                 'VipsForeignLoadPng',
++                'VipsForeignLoadSvg',
+                 'VipsForeignLoadTiff',
+                 'VipsForeignLoadWebp'
+             ]
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -20,6 +20,9 @@ patchedDependencies:
   '@mantine/hooks':
     hash: b45a85bd4d007c796b93eb90a75d3234b81cce13964921b25721734c7ed79f64
     path: patches/@mantine__hooks.patch
+  next@16.3.0:
+    hash: 9b6ebe47f55d18825f485fb243f076aaa3968476c5e07b8378e08b7f1c7e5ee7
+    path: patches/next@16.3.0.patch
 
 importers:
 
@@ -60,7 +63,7 @@ importers:
         version: link:packages/civitai-db-schema
       '@civitai/next-axiom':
         specifier: ^0.17.0
-        version: 0.17.0(next@16.3.0(@babel/core@7.29.0)(@opentelemetry/api@1.9.0)(@playwright/test@1.57.0)(@types/node@24.13.3)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(sass@1.90.0))
+        version: 0.17.0(next@16.3.0(patch_hash=9b6ebe47f55d18825f485fb243f076aaa3968476c5e07b8378e08b7f1c7e5ee7)(@babel/core@7.29.0)(@opentelemetry/api@1.9.0)(@playwright/test@1.57.0)(@types/node@24.13.3)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(sass@1.90.0))
       '@civitai/shared':
         specifier: workspace:*
         version: link:packages/civitai-shared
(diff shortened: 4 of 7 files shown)


ZacxDev:
Correcting three statements in the PR description that are now stale or imprecise. Flagging here rather than silently editing the body, since a reviewer may already have read them. All three **understate** what has been verified.

1. **"Not verified in a `next build` / `output: 'standalone'` artifact"** — no longer true. The preview deployment for this PR *is* a standalone production build, and `/api/og` returns HTTP 200 with a valid PNG there once the image optimizer has been warmed to a genuine cache MISS.

2. **"The `OgCard` (real-entity) branch has never been watched green"** — no longer true. Both branches were verified against that standalone preview: FallbackCard → `200 image/png` (50,299 B), OgCard (real model id) → `200 image/png` (441,064 B), both with valid PNG magic bytes. Malformed and absent query params still correctly return 400.

3. **"Previews are opt-in behind the `preview` label"** — imprecise. This PR carries no labels, yet every preview check ran and passed. Maintainer PRs auto-deploy; the label gates non-maintainer PRs.

For reviewers, an **independent negative control** worth knowing about: #3766's preview is also a standalone build on Next 16.3.0 *without* this patch, and `/api/og` returns 500 there with the identical body. Same spec, same command, differing only in this patch — alongside live production, which reproduces the same 500.

Two follow-up commits incoming from an adversarial review pass: extending the patch to the ESM copy of `image-optimizer.js` (not reachable in the standalone runtime today — that file isn't traced into the build — but it closes a silent-failure mode if resolution ever changes), and moving the structural assertion into a CI job that actually blocks, since the outage this fixes was introduced by a dependency upgrade and nothing currently red would stop the next one.

PR #2451, opened 2026-06-09
feat(observability): event-loop long-task detector for api-primary blocker attribution

ZacxDev:
## Why
The `api-primary` pool pins under bursts because the single JS thread saturates. CPU profiles are aggregate/idle-diluted and can't tell us *which* operations block the loop or how often. This adds direct, frequency-weighted attribution.

## Design — tiered, default-off, zero-cost when disarmed
Revised after an adversarial audit (the original always-on ALS attribution risked adding the very async-context cost the team removed from OTEL).

| Tier | Trigger | What runs | Per-request cost |
|------|---------|-----------|------------------|
| **Disarmed** (prod default) | `EVENTLOOP_LONGTASK_THRESHOLD_MS` unset/≤0 | nothing installed | **Zero** — original `next()`/handler, no wrapper (proven by unit test: no ALS store created) |
| **Base armed** | `THRESHOLD_MS=50` | `monitorEventLoopDelay` lag gauge (libuv C++, ≈free) + drift detector; blocks labeled `unlabeled` | one `Date.now()` subtraction per 20ms tick, process-wide. **No async_hooks.** Safe steady-state. |
| **Labels** (opt-in) | `+ EVENTLOOP_LONGTASK_LABELS=true` | per-procedure ALS route attribution | one `AsyncLocalStorage.run()` per request — **short windows only**, re-introduces the OTEL-class cost |
| **Stacks** (opt-in) | `+ EVENTLOOP_LONGTASK_STACKS=true` | sampled + size-capped async_hooks stack capture | `Error().stack` on ~1/N resources — short windows only |

Wired from `instrumentation.node.ts` (OTEL-independent), `trpc.ts` (middleware no longer async; runs original path when disarmed), and the `/api/v1/images` feed handler.

## Output
- `civitai_app_eventloop_delay_ms{quantile}` — **authoritative** lag signal.
- `civitai_app_eventloop_longtask_duration_seconds{label}` — block count/duration; `topk(10, sum by(label)(rate(...sum[5m])))` ranks blockers (label = `unlabeled` unless Labels tier on).
- `civitai_app_eventloop_longtask_suspension_total` — gaps > 5s (pod suspension/CPU-steal) split out so they don't masquerade as JS blocks.
- Rate-limited Axiom logs (`eventloop-longtask`), fire-and-forget.

## Audit fixes applied (commit 4471616d9)
1. **Disarmed = literally zero** — `armed` settled once at load; disarmed call sites are byte-equivalent to pre-PR (unit-tested).
[...]

PR #3792, opened 2026-08-10
fix(telemetry): make the OTel logs bridge reach every logToAxiom call site

ZacxDev:
## What shipped broken

#3736 added an OTel logs bridge at the single `logToAxiom` chokepoint. Enabled on one
low-traffic workload, one 5-minute window measured:

| path | records |
|---|---:|
| stderr `_axiom` write (pre-existing, unchanged) | **383** |
| OTel bridge | **5** |

**1.3% delivery**, and reconciling by event name shows it is not a sampling loss — it is a
clean partition:

- `eventloop-longtask`: **8 via the bridge, 8 via stderr**, bodies byte-identical.
- `model-file-scan` (66), `scan-files-fallback` (66), `resource-training-v2-cron` (19),
  `Notification cursor stale` (12) and ~984 unnamed records: **stderr only, zero bridged**.

`eventloop-longtask` is registered from `src/instrumentation.node.ts` — the same module
that constructs the `LoggerProvider` and calls `setStructuredLogSink()`. Call sites
co-located with the registration worked; every other call site was a permanent, silent
no-op.

The counters #3736 added specifically to make this legible — `otel_log_records_emitted_total`,
`otel_log_records_skipped_total` — **were not exposed either**. A live `/metrics` returned
HTTP 200 with 1,758 lines, 90 of them `nodejs_*` (so the endpoint and scrape were healthy)
and **zero** `otel_log*` series. The instrument built to detect the failure had the same
failure.

## Diagnosis

Both problems are one mechanism: **a module's top-level state is per *runtime* module, and
the bundler decides how many runtime modules a source file becomes.**

Measured by building this repo and reading the emitted server source maps (`sources`
attributes emitted code back to source modules; Turbopack's numeric module ids cannot):

| module | emitted copies in `.next/server` |
|---|---:|
| `src/server/logging/client.ts` | **14** (7 under `chunks/`, 7 under `chunks/ssr/`) |
| `packages/civitai-telemetry/src/otel-logs.ts` | **1** |
| `@opentelemetry/api-logs` | **0** (not bundled at all) |

`logging/client.ts` held the sink as a module-scope `const sink: AxiomDeps = {}`. Fourteen
emitted copies means fourteen independent objects; `setStructuredLogSink()` runs **once**,
at boot, from the instrumentation entry, and armed exactly one of them.

Which one is also measurable, and it matches the production split exactly. One merged
runtime module (`chunks/[root-of-the-server]__0lu0oox._.js`, 184 inlined sources) contains
`src/instrumentation.node.ts`, `src/server/eventloop-longtask.ts` **and** the only copy of
`otel-logs.ts`. `eventloop-longtask` resolves the logger shim through the same import
binding as the registrar, so it got the armed copy. Nothing else did.

The counters are the same story one layer down: they were registered on prom-client's
**default** registry from the instrumentation graph, while `/api/metrics` scrapes the
request graph's default registry. This repo already hit this exact trap (PR #2451) and
already carries the cure — `instrumentationRegistry`, pinned on `globalThis`, which
`src/pages/api/metrics.ts` merges into every scrape. `otel-logs.ts` simply did not use it.

### On #3736's timing-vs-identity call

It was wrong, and interestingly so. Its premise checks out: every installed `api-logs`
copy does share the `Symbol.for('io.opentelemetry.js.api.logs')` key and
`API_BACKWARDS_COMPATIBILITY_VERSION = 1` (verified in the installed sources). Its
conclusion does not.

- **The premise is idle.** `@opentelemetry/api-logs` is not bundled at all — Next
  externalizes node_modules for the server build, and the emitted maps contain zero copies
  of it. There is one instance, from the require cache. Adding it to
  `serverExternalPackages` would change nothing, so that deferred follow-up is moot for
  this bug. **This PR does not touch `serverExternalPackages`.**
- **Hypothesis (c) — a cached dead ProxyLogger — is refuted by the observation itself.**
  There is only one copy of `emitOtelLog` in the whole build, so the working and
  non-working call sites ran the *same* closure with the *same* `cachedLogger`. A dead
  logger would have broken both.
- **It was an identity problem — of the app's own shim**, not of any OTel package. Both
  of #3736's hypotheses were about the wrong module.

The lazy per-emit bind #3736 added is still correct and is kept; it just was not
sufficient, and could not have been.

## The fix

1. **`src/server/logging/structured-log-sink.ts` (new)** — the sink object, pinned on
   `globalThis` via `??=`, which is the one thing all emitted copies share. `??=` rather
   than `=` so a later-evaluated copy *adopts* the object instead of orphaning loggers that
   already closed over it. The module has no runtime imports (its `AxiomDeps`/`EmitLog`
   imports are `import type`), so it adds no edge to any graph — including the client
   graph that `logging/client.ts` is reachable from, which is what broke the original
   build attempt.
2. **`otel-logs.ts` counters** move to `registerInstrumentationMetric(...)` with
   `registers: [instrumentationRegistry]` — the established cross-graph pattern here.
   Series names are unchanged (`civitai_app_otel_log_records_{emitted,skipped}_total`).

No bundler config change, so there is no risk of reintroducing the client-bundle
resolution failure (`cluster`/`fs`/`v8`) that forced the original revert.

**Unchanged, deliberately:** the stderr `_axiom` write, the Axiom ingest path, and
`OTEL_LOGS_ENABLED`'s default. Both sinks still coexist; the ramp stays operator-controlled
in config.

## The build gate

`scripts/check-server-graph-singletons.mjs`, wired into the Dockerfile builder stage
(hard gate, unlike the report-only bundle budget). It reads the emitted server source maps
and enforces two rules:

- `SHARED_STATE` — every emitted copy of a module must reference its `globalThis` key.
- `SINGLETON` — `otel-logs.ts` must be emitted exactly once (a second copy would be a
  second bridge that never sees the provider registration).

Both rules **fail** when they match nothing, and the script exits **2** (not 0) when it
finds no source maps — a scan that can see nothing must never report health.

This class is structurally invisible to everything else we run: `tsc` checks one source
file, ESLint reads source, and Vitest loads a module once, so a module-scope singleton
looks like a singleton everywhere except in the emitted output.

**Watched red before trusted green**, on real build output:

```
# pre-fix build, gate pointed at the module that actually held the defect
FAIL  SHARED_STATE src/server/logging/client.ts
      14 emitted copies, 0 referencing globalThis.__civitaiStructuredLogSink
      ✗ 14 of 14 emitted copies do NOT reference `globalThis....`
```
```
# full build of this branch
 OK   SHARED_STATE src/server/logging/structured-log-sink.ts
      11 emitted copies, 11 referencing globalThis.__civitaiStructuredLogSink
 OK   SINGLETON    packages/civitai-telemetry/src/otel-logs.ts
      1 emitted copy
```

Plus a real rebuild with the pin removed (see the comment below) and 7 fixture-driven unit
cases covering each failure branch separately.

## Tests

No `vi.mock` of the OTel or prom packages anywhere — the SDKs expose everything needed to
run this for real.

- `src/server/logging/__tests__/structured-log-sink.test.ts` (5) — the seam. Two genuinely
  separate module instances (`vi.resetModules()`) sharing one `globalThis`, which is
  structurally the relationship the bundler creates between two emitted copies. Registering
  through instance A must be observable through instance B, including composed with a real
  `createAxiomLogger`. Killing mutation (revert to a module-scope object): **all 5 fail**,
  each with its own message — the headline one is `expected undefined to be [Function Mock]`,
  which is the production failure exactly.
- `packages/civitai-telemetry/src/__tests__/otel-logs-metrics.test.ts` (4) — *which*
  registry, not arithmetic. Killing mutation (counters back on the default registry):
  **3 of 4 fail**, distinctly (`expected undefined to be Counter`, `expected Counter to be
  undefined`, and the positive control finding an empty scrape). The 4th is a naming
  invariant and correctly survives — it is not regression coverage and is not counted as
  such. Notably the pre-existing 24 bridge tests stay **green** under that mutation, which
  is precisely why this needed a new instrument rather than a new assertion.
- `scripts/__tests__/check-server-graph-singletons.test.ts` (7) — the gate's own negative
  controls: unpinned copy, multiple unpinned copies, watched module absent, singleton
  duplicated, no source maps (exit 2), no build output (exit 2), plus a positive control
  that it can pass at all.

## Not verified here

- **Runtime counter values.** The build and boot are verified; `emitted ≈ logToAxiom call
  rate` and `skipped{reason="no_provider"} ≈ 0` can only be read from a running deployment
  with the flag on. That is the acceptance check to run after this reaches an environment
  with the bridge enabled.
- **Production timing.** Production deploys from the **`release`** branch, not `main` — so
  merging this does not reach production until a release promotion. The ramp sequencing is
  unaffected by the merge itself.
- One consequence worth expecting: with the sink now reaching every call site,
  `skipped{reason="disabled"}` will count the *full* log rate wherever the flag is off.
  That is intended — it is the live positive control that the wiring exists.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,24 @@ ENV SOURCE_COMMIT=$SOURCE_COMMIT
 RUN --mount=type=cache,target=/app/.next/cache \
     SKIP_ENV_VALIDATION=1 IS_BUILD=true NODE_OPTIONS="--max_old_space_size=${NODE_BUILD_MEM}" pnpm run build
 
+# Server-graph module identity — a HARD gate, deliberately unlike the report-only
+# bundle budget below.
+#
+# A module's top-level state is per RUNTIME module, and the bundler decides how many
+# runtime modules a source file becomes: Turbopack inlines `src/server/logging/client.ts`
+# into 14 of them in this build. Anything that must be a process-wide singleton therefore
+# has to be reached through `globalThis`. Nothing else we run can see a violation — `tsc`
+# checks one source file, eslint reads source, and Vitest loads a module once, so a
+# module-scope singleton looks like a singleton everywhere except in the emitted output.
+# That is how the OTel logs bridge shipped delivering 1.3% of its records with a green
+# typecheck, a green lint and a green suite.
+#
+# Runs here for the same reason the budget does: `.next` exists in this stage, so there is
+# no second build. It reads the emitted `.js.map` `sources` to attribute code to source
+# modules, and exits 2 (not 0) if it finds no maps — a scan that can see nothing must not
+# report health. Cheap: a few seconds of file reads.
+RUN node scripts/check-server-graph-singletons.mjs
+
 # Bundle-size budget (report-only during the soak). Next 16 (Turbopack) emits
 # opaque hashed chunks and removed per-route build stats, so scripts/bundle-budget.mjs
 # parses .next/build-manifest.json to reconstruct per-page First Load JS (brotli)
--- a/packages/civitai-telemetry/src/__tests__/otel-logs-metrics.test.ts
+++ b/packages/civitai-telemetry/src/__tests__/otel-logs-metrics.test.ts
@@ -2,7 +2,11 @@ import client from 'prom-client';
 import { describe, expect, it } from 'vitest';
 
 import { instrumentationRegistry, PROM_PREFIX } from '../client';
-import { emitOtelLog, otelLogRecordsEmittedCounter, otelLogRecordsSkippedCounter } from '../otel-logs';
+import {
+  emitOtelLog,
+  otelLogRecordsEmittedCounter,
+  otelLogRecordsSkippedCounter,
+} from '../otel-logs';
 
 /**
  * WHERE the bridge's counters are registered, which is the whole reason they were absent
--- a/packages/civitai-telemetry/src/__tests__/otel-logs-metrics.test.ts
+++ b/packages/civitai-telemetry/src/__tests__/otel-logs-metrics.test.ts
@@ -1,0 +1,77 @@
+import client from 'prom-client';
+import { describe, expect, it } from 'vitest';
+
+import { instrumentationRegistry, PROM_PREFIX } from '../client';
+import { emitOtelLog, otelLogRecordsEmittedCounter, otelLogRecordsSkippedCounter } from '../otel-logs';
+
+/**
+ * WHERE the bridge's counters are registered, which is the whole reason they were absent
+ * from production.
+ *
+ * The counters shipped on prom-client's DEFAULT registry. This module is only ever loaded
+ * from `src/instrumentation.node.ts`, and `/api/metrics` is an API route — a different
+ * bundler graph with a different default registry — so the counters incremented into a
+ * registry nothing scrapes. Measured on a live pod: `/metrics` returned HTTP 200 with
+ * 1,758 lines, 90 of them `nodejs_*` (so the endpoint was healthy and the scrape worked)
+ * and ZERO `otel_log*`. The one instrument built to make the bridge's silence legible was
+ * itself silent, which is why the 1.3% delivery rate had to be found by reconciling log
+ * volumes by hand.
+ *
+ * `instrumentationRegistry` is pinned on `globalThis`, so it is the same object in every
+ * graph, and `src/pages/api/metrics.ts` already merges it into the scrape response.
+ *
+ * These cases assert the WIRING (which registry), not the arithmetic. A test that only
+ * checked `counter.inc()` moves a number would have passed throughout the outage.
+ */
+
+const EMITTED = `${PROM_PREFIX}otel_log_records_emitted_total`;
+const SKIPPED = `${PROM_PREFIX}otel_log_records_skipped_total`;
+
+/** Render exactly what a scrape of a registry would contain. */
+const scrape = (registry: client.Registry) => registry.metrics();
+
+describe('otel-logs counters — registry wiring', () => {
+  it('registers BOTH counters on the cross-graph instrumentationRegistry', () => {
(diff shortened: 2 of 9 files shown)


ZacxDev:
## Runtime verification on the preview deploy

The preview built and booted, which also means the new `RUN node scripts/check-server-graph-singletons.mjs` step passed inside the real image build, not just locally.

Scraping `/api/metrics` on the running preview:

```
http=200  lines=1490
nodejs_ lines: 87                      <- positive control: endpoint healthy, scrape works
otel_log lines: 6                      <- was ZERO in production

civitai_app_otel_log_records_emitted_total 0
civitai_app_otel_log_records_skipped_total{reason="disabled"} 28
civitai_app_eventloop_longtask_total 0
```

Both fixes are visible here, and the third line is what makes it conclusive rather than suggestive:

- **The counters are exposed.** Six `otel_log*` lines where production had none against a healthy 1,758-line scrape. That is the step-1 instrument working.
- **The sink reaches call sites that were previously dark.** `skipped{reason="disabled"}` only increments inside `emitOtelLog`, which only runs when the sink object a call site's shim copy holds has `emitLog` set. It reads 28 and climbs (27 a minute earlier), so it is live rather than a stale registration.
- **And those 28 cannot be the one call site that already worked.** `EVENTLOOP_LONGTASK_THRESHOLD_MS` is unset here, so the long-task detector is disarmed — `civitai_app_eventloop_longtask_total` confirms it at **0** emissions. Every one of those 28 therefore came from the call-site population that delivered **zero** records in production. Pre-fix this counter would have read 0, and would not have been on the endpoint to read.
- **The dark-launch gate is intact.** `OTEL_LOGS_ENABLED` is unset here, so every record takes the `disabled` skip and `emitted_total` is correctly 0. The flag's default is unchanged by this PR.

What still cannot be checked from a preview: `emitted_total ≈ the logToAxiom call rate` and `skipped{reason="no_provider"} ≈ 0`. Both need the flag actually on, so they are the acceptance read once this reaches an environment with the bridge enabled — and production specifically deploys from `release`, not `main`.

PR #3796, opened 2026-08-10
fix(reactions): delete by natural key, not by a replica-read id

JustMaier:
## The bug

Un-reacting sometimes doesn't remove the reaction. The UI says it worked; the row survives. Reload and the reaction is back.

`toggleReaction` read the existing row from the **replica** (`getDbWithoutLag()`) and then deleted it from the **primary** by the primary key it had just read:

```ts
const existing = await getReaction({ ... });        // replica
if (existing) {
  await deleteReaction({ id: existing.id, ... });   // dbWrite...deleteMany({ where: { id } })
  return 'removed';                                 // unconditional
}
```

React -> un-react -> re-react inside the replication window and the replica returns a **stale `id`**. The primary's current row has a different `id`, `deleteMany` matches zero rows, and the handler reports `'removed'` anyway. The UI updates optimistically, the ClickHouse tracker logs an `Image_Delete`, the entity metric is decremented — and the reaction is still there.

Note the existing comment on `getReaction` anticipated the *opposite* failure ("'removed' returns against rows that don't exist"). The real failure is `'removed'` against a row that very much does exist.

## Production measurement (2026-08-04)

Of **19,657** logged `Image_Delete` events, **368 had no corresponding deletion** — and all 368 still had live `ImageReaction` rows in Postgres, **357 of them created 8-35 seconds before** the logged un-react. That gap is exactly the replication window.

## What changed

**1. Delete by natural key.** Every reaction table has a unique index on `(<entity>Id, userId, reaction)`. Deleting on that triple is immune to the stale id: it names the row by what the user actually did, not by which physical row the replica happened to see.

I verified all nine indexes exist in production `pg_index` rather than trusting the Prisma `@@unique` declarations (this repo has a documented history of schema-vs-DB drift):

| entityType | table | before | after |
|---|---|---|---|
| `question` | `QuestionReaction` | `where: { id }` | `where: { questionId, userId, reaction }` |
| `answer` | `AnswerReaction` | `where: { id }` | `where: { answerId, userId, reaction }` |
| `commentOld` | `CommentReaction` | `where: { id }` | `where: { commentId, userId, reaction }` |
| `comment` | `CommentV2Reaction` | `where: { id }` | `where: { commentId, userId, reaction }` |
| `image` | `ImageReaction` | `where: { id }` | `where: { imageId, userId, reaction }` |
| `post` | `PostReaction` | `where: { id }` | `where: { postId, userId, reaction }` |
| `resourceReview` | `ResourceReviewReaction` | `where: { id }` | `where: { reviewId, userId, reaction }` |
| `article` | `ArticleReaction` | `where: { id }` | `where: { articleId, userId, reaction }` |
| `bountyEntry` | `BountyEntryReaction` | already natural-key | unchanged (count now returned) |

`bountyEntry` was already correct, and the reason is structural rather than lucky: `BountyEntryReaction` has no `id` column at all — its **primary key *is* `(bountyEntryId, userId, reaction)`**, so there was no id to delete by.

**2. `'removed'` only when a row was removed.** `deleteReaction` now returns `deleteMany`'s `count`. `toggleReaction` returns `'removed'` only on a positive count and `'noop'` otherwise. The per-entity `metrics.queueUpdate` calls are likewise gated on a nonzero count.

**3. Controller no longer emits phantom events.** On `'noop'` the controller skips `getTrackerEvent` (no `Image_Delete` for a delete that didn't happen) and skips `updateEntityMetric` (no `-1` on a reaction that was never taken away). The `result === 'created'` branch for rewards/notifications was already correct for a third value.

The `'noop'` return is safe for both callers: `reaction.router` is fire-and-forget and discards the value, and `comment.router` returns it to a client that only does optimistic `onMutate` updates and never reads the response.

**4. Dropped the dead `if (!id) return;` guards.** They were the mechanism that made this silent — a missing id no-op'd the delete and still reported success. With required params they can't recur.

## On reading `getReaction` from the primary

I deliberately did **not** do this, and I agree with the instinct in the brief. Routing `getReaction` to the primary would cost a primary read on every toggle on one of the hottest social paths in the app, to fix a bug the natural-key delete fixes for free. More importantly it would only *narrow* the window rather than close it: the read and the delete are still two separate statements, so a concurrent toggle landing between them reintroduces the same mismatch. The natural-key delete is correct regardless of what the read returned, which is the stronger property.

The replica read does still guard one thing, and I left a note saying so: a stale **miss** (replica says no row, primary has one) causes `createReaction` to throw a UNIQUE violation. That is unchanged by this PR and remains the reason `HIGH_REPLICATION_LAG_MODE` routes this read to the primary during an incident. I rewrote the stale half of that comment, which now overclaimed.

## Tests

`src/server/services/__tests__/reaction-stale-replica-delete.test.ts` — 11 tests. The fake seeds the replica and the primary with the *same natural key but different ids*, which is the actual race, and both stores honour whatever `where` they are handed, so the test observes the key that was really used rather than asserting on a mock call shape alone.

Checked how it fails, not just that it passes: reverting `reaction.service.ts` fails **10 of 11 in 1.79 s** with legible assertions (`expected [ { id: 222, imageId: 1, ... } ] to have a length of +0 but got 1`, `expected 'removed' to be 'noop'`). No loops, no timeouts, nothing that could hang the runner. The one test that still passes on a revert is `bountyEntry` — correct, since it was never broken, and it doubles as a control proving the suite discriminates rather than failing wholesale.

## Verification

- `pnpm run typecheck` — 58 errors, **identical to the same run on clean `origin/main`** in this worktree (the only diff between the two logs is nondeterministic field ordering inside one pre-existing `creator-shop.service.ts` message). Zero introduced. The pre-existing set is environmental: `kysely` is not present in this machine's shared `node_modules`, and the generated Prisma client predates `userCosmeticShopItemResale`.
- `npx eslint` on the three changed files — 0 errors.
- `pnpm run test:lint-rules` — 215 passed.
- `pnpm exec prettier --write` on the three changed files only.

## Found and deliberately not fixed

- **`getTrackerEvent`'s `bountyEntry` case queries the wrong table** — `reaction.controller.ts:150` reads `dbRead.answer.findFirst({ where: { id: input.entityId } })` for a bounty entry, so `ownerId` on every `BountyEntry_Create` / `BountyEntry_Delete` event is whatever `Answer` happens to share that id, and the event is dropped entirely when no such answer exists. Real bug, unrelated to replication, and fixing it changes ClickHouse event contents — worth its own PR.
- **`blocks.router.workflow.test.ts` collects 0 tests in this environment**, failing on `Cannot find package 'kysely'`. Same missing dependency as the typecheck errors, reproduces on clean `origin/main`, and unrelated to these files (the new test does not import it). Flagging it because CLAUDE.md names that suite as the worktree-validation canary — here it fails loudly rather than silently, and I diagnosed it rather than working around it. It needs a `pnpm install` on the shared tree, which I did not run because other agents are active on it.
- **Unused `Prisma` import** at `reaction.controller.ts:1`. Left alone to keep the controller diff to the two behavioural lines. I did remove the equally dead `dbRead` import from `reaction.service.ts`, since that file is substantially rewritten here.
- **No backfill for the 368 measured rows.** They are live `ImageReaction` rows whose users believe they un-reacted, with `Image_Delete` events already in ClickHouse that will never reconcile. Cleaning them up is a data question, not a code one.

### DIFF
--- a/src/server/controllers/__tests__/reaction-noop-tracking.test.ts
+++ b/src/server/controllers/__tests__/reaction-noop-tracking.test.ts
@@ -1,0 +1,77 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ProtectedContext } from '~/server/createContext';
+
+const { mockToggleReaction, mockUpdateEntityMetric, mockDbRead } = vi.hoisted(() => ({
+  mockToggleReaction: vi.fn(),
+  mockUpdateEntityMetric: vi.fn(async () => undefined),
+  mockDbRead: {
+    image: {
+      findFirst: vi.fn(async () => ({ nsfwLevel: 1, userId: 99 })),
+      findUniqueOrThrow: vi.fn(async () => ({ postId: 42 })),
+    },
+  },
+}));
+
+vi.mock('~/server/services/reaction.service', () => ({ toggleReaction: mockToggleReaction }));
+vi.mock('~/server/utils/metric-helpers', () => ({ updateEntityMetric: mockUpdateEntityMetric }));
+vi.mock('~/server/db/client', () => ({ dbRead: mockDbRead, dbWrite: mockDbRead }));
+vi.mock('~/server/services/collection.service', () => ({
+  getContestsFromEntity: vi.fn(async () => []),
+}));
+vi.mock('~/server/services/common.service', () => ({
+  hasEntityAccess: vi.fn(async () => [{ hasAccess: true }]),
+}));
+vi.mock('~/server/rewards', () => ({
+  encouragementReward: { apply: vi.fn(async () => undefined) },
+  goodContentReward: { apply: vi.fn(async () => undefined) },
+}));
+vi.mock('~/server/notifications/client', () => ({
+  notifications: { notificationExists: vi.fn(async () => true) },
+}));
+vi.mock('~/server/services/notification.service', () => ({
+  createNotification: vi.fn(async () => undefined),
+}));
+
+import { toggleReactionHandler } from '~/server/controllers/reaction.controller';
+
+const input = { entityType: 'image', entityId: 1, reaction: 'Like' } as const;
+
+function makeCtx() {
+  return {
+    user: { id: 5, isModerator: false },
+    ip: '127.0.0.1',
+    track: { reaction: vi.fn(async () => undefined) },
+  } as unknown as ProtectedContext & { track: { reaction: ReturnType<typeof vi.fn> } };
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe('toggleReactionHandler — a delete that removed nothing', () => {
+  it('emits no tracker event and no metric decrement on noop', async () => {
+    mockToggleReaction.mockResolvedValue('noop');
+    const ctx = makeCtx();
+
+    const result = await toggleReactionHandler({ ctx, input });
+
+    expect(result).toBe('noop');
+    expect(ctx.track.reaction).not.toHaveBeenCalled();
+    expect(mockUpdateEntityMetric).not.toHaveBeenCalled();
+  });
+
+  // Negative control: the guards must be keyed on 'noop', not simply switched off.
+  it('still emits both when a row really was removed', async () => {
+    mockToggleReaction.mockResolvedValue('removed');
+    const ctx = makeCtx();
+
+    await toggleReactionHandler({ ctx, input });
+
+    expect(ctx.track.reaction).toHaveBeenCalledWith(
+      expect.objectContaining({ type: 'Image_Delete', entityId: 1 })
+    );
+    expect(mockUpdateEntityMetric).toHaveBeenCalledWith(
+      expect.objectContaining({ metricType: 'ReactionLike', amount: -1 })
+    );
+  });
+});
--- a/src/server/controllers/reaction.controller.ts
+++ b/src/server/controllers/reaction.controller.ts
@@ -230,7 +230,7 @@ export const toggleReactionHandler = async ({
       userId: ctx.user.id,
       isModerator: ctx.user.isModerator,
     });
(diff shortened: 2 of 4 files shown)


JustMaier:
Addressed in `2da2e3e`.

**Required change.** `expect(where).not.toHaveProperty('id')` -> `expect(where).toEqual({ [foreignKey]: 7, userId: 5, reaction: 'Heart' })`. Re-ran your mutation (stripped `userId`/`reaction` from the `article` case): now fails with `expected { articleId: 7 } to deeply equal { articleId: 7, userId: 5, …(1) }`. 16/16 pass unmutated.

**Both optional extras taken, but the first needed a fix to work.** `expect(mockQueueUpdate).not.toHaveBeenCalled()` in the noop test as specified would have been vacuous — that test used `image`, and `image` queues no metric at all, so it could not observe the `if (count > 0)` guards regardless. Parametrised the noop test over the five entity types that *do* queue one (`question`, `answer`, `post`, `article`, `bountyEntry`). Mutation-checked: removing `if (count > 0)` from `articleMetrics` now fails with `expected "vi.fn()" to not be called at all, but actually been called 1 times`.

**Controller coverage added** — `src/server/controllers/__tests__/reaction-noop-tracking.test.ts`, 2 tests. Removing both `'noop'` guards at `:233`/`:243` now fails. Includes a negative control asserting a real `'removed'` still emits `Image_Delete` and `amount: -1`, so the guards can't be satisfied by switching the calls off entirely.

**Correction to my earlier revert measurement.** My original "10 of 11" was taken pre-commit. Re-running it after committing was vacuous — `git stash` reverts to HEAD, which is now the fix. Redid it against `git show origin/main:...`: **15 of 16 fail in 2.06 s**, `bountyEntry` passing as the control. Flagging the method error since the number was in the PR body.

Typecheck unchanged at 58 (identical set, none reaction-related), eslint clean on both test files, prettier applied to those two only.

PR #3798, opened 2026-08-10
fix(metrics): keep the nightly reaction-exactness gauges alive across pod rolls

JustMaier:
The nightly exactness audit writes its gauges once a day; the jobs deployment rolls several times a day. **Measured against live Prometheus 2026-08-10, ~10h after a run that had SUCCEEDED at 04:43: all 189 `civitai_app_reaction_exactness_recent_coverage_ratio` series read 0.** The value was simply gone, so the nightly liveness panel read `No data` whether the job was healthy or dead — it could not do its job at all, and any alert built on that gauge was inert rather than green.

## The fix

The nightly persists its result to sysRedis and the hourly job re-publishes it. That also lands the nightly's values on the pod carrying `audit_last_run_timestamp_seconds`, which is the freshness anchor the hourly queries already join against.

**`ranAt` is the nightly's own completion time and is re-published verbatim, never re-stamped.** Re-stamping is the obvious shortcut and it is the bug: a nightly that died weeks ago would report itself alive every hour, which is strictly worse than the `No data` this replaces — silence at least doesn't lie. The new `exactness_last_run_timestamp_seconds` gauge is what makes that assertable, and a test pins it.

Both new failure paths are non-fatal: a Redis outage must not fail the reconciliation, which is the actual job.

## Corrected the gauge-aggregation comment

It said to use `min()` for coverage and `max()` for counts. **Both are wrong**, and that comment is the first thing someone writing an alert reads. Every pod registers these gauges and prom-client never resets them, so `min()`/`avg()` read the ~187 pods that never ran, and a bare `max()` reads whichever pod remembers the best value — live at the time of writing there were two non-zero series, one 2h stale and one 6m fresh, so a bare `max()` would mask a bad hour.

Correct form is `max(<gauge> and on(pod) topk(1, <last_run> != 0))`, with the `!= 0` **inside** `topk` because `topk(1, gauge)` still returns a pod when every value ties at zero.

## Tests are mutation-tested, not just run

Per the repo's "a passing test says nothing about how it FAILS" rule, I checked what a revert prints:

- Re-stamping the timestamp → `expected 0 to be greater than 126000`
- Dropping the re-publish call → `expected +0 to be 0.9997`

The age assertion is deliberately ordered *before* the equality one so the first line a reader sees names the regression, instead of printing two 10-digit unix timestamps that differ.

## Also

`PROM_PREFIX` added to the shared `~/server/prom/client` test mock. Modules name their metrics `PROM_PREFIX + '...'`, so its absence made any such test unable to look a gauge up by the name production actually emits.

Second commit is **comment-only**, on the repair service's `removalAt` block. It records what a now-stamped removal costs a windowed reader: measured after the 2026-08-07 backfill, 3,197,222 removals landed on one day against a ~18k/day baseline, leaving 17,247 orphaned removals inside the image leaderboards' 30-day window. Not changing the stamping — the ongoing repair writes 1-2 rows a night and the bulk rows age out 2026-09-06. The note exists so the next person planning a bulk repair sees the cost beforehand. Credit to @liz for the `original createdAt + 1s` option it records.

## Verification

- `pnpm run typecheck` — OK, 0 errors
- New suite 8/8, `pnpm exec eslint` clean, prettier clean
- `pnpm run test:lint-rules` — only the pre-existing `no-wholesale-module-mock` fixture failure
- Full unit suite: 16 failures across 6 repo-scan suites, **verified pre-existing** by re-running them with these changes removed

## Deploy sequencing

This introduces `civitai_app_reaction_exactness_last_run_timestamp_seconds`. talos-infra#929 has a nightly liveness alert that depends on it — **that one should merge after this deploys**, otherwise it is silently inert.

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1880,6 +1880,11 @@ export const REDIS_SYS_KEYS = {
     LEGACY_UPGRADE_LOCK: 'session:legacy-upgrade-lock',
   },
   JOB: 'job',
+  METRIC_RECONCILIATION: {
+    // Last completed nightly reaction-exactness result, so the hourly job can
+    // re-publish its gauges after a pod roll. See metric-reconciliation-audit.ts.
+    NIGHTLY_EXACTNESS: 'metric-reconciliation:nightly-exactness',
+  },
   BUZZ_WITHDRAWAL_REQUEST: {
     STATUS: 'buzz-withdrawal-request:status',
   },
--- a/src/__tests__/setup.ts
+++ b/src/__tests__/setup.ts
@@ -173,6 +173,10 @@ const promMetricStub = () => {
 };
 
 vi.mock('~/server/prom/client', () => ({
+  // The real value. Modules that build their own prom-client metrics name them
+  // `PROM_PREFIX + '...'`, and a stubbed prefix would make every such test assert
+  // against a metric name production never emits.
+  PROM_PREFIX: 'civitai_app_',
   // Factory functions — modules call these at import time to build their metrics
   // (e.g. meilisearch/client.ts, eventloop-longtask.ts). Each returns a stub.
   registerCounter: vi.fn(promMetricStub),
--- a/src/server/jobs/__tests__/metric-reconciliation-audit.test.ts
+++ b/src/server/jobs/__tests__/metric-reconciliation-audit.test.ts
@@ -1,0 +1,212 @@
+import client from 'prom-client';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type * as RedisClient from '~/server/redis/client';
+
+const { mockAuditHour, mockAuditExactness, mockRedisGet, mockRedisSet } = vi.hoisted(() => ({
+  mockAuditHour: vi.fn(),
+  mockAuditExactness: vi.fn(),
+  mockRedisGet: vi.fn(),
+  mockRedisSet: vi.fn(),
+}));
+
+vi.mock('~/server/services/metric-reconciliation.service', () => ({
+  auditReactionHour: mockAuditHour,
+  auditReactionExactness: mockAuditExactness,
+  requestReactionRepair: vi.fn().mockResolvedValue(false),
+  setReactionRepairHook: vi.fn(),
+}));
+vi.mock('~/server/services/metric-reaction-repair.service', () => ({
+  repairReactionMetrics: vi.fn(),
+}));
+vi.mock('~/server/flipt/client', () => ({
+  isFlipt: vi.fn().mockResolvedValue(false),
+  FLIPT_FEATURE_FLAGS: { METRIC_REACTION_REPAIR: 'metric-reaction-repair' },
+}));
+vi.mock('~/server/logging/client', () => ({ logToAxiom: vi.fn().mockResolvedValue(undefined) }));
+
+// Spread the real module so REDIS_SYS_KEYS stays authoritative — a hand-written
+// key here would let the job and the test agree on a key production never uses.
+vi.mock('~/server/redis/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof RedisClient>()),
+  sysRedis: { get: mockRedisGet, set: mockRedisSet },
+}));
+
+vi.mock('../job', () => ({
+  createJob: (_name: string, _cron: string, fn: (ctx: unknown) => Promise<unknown>) => ({
+    run: fn,
+  }),
+}));
+
+import { reactionExactnessAuditJob, reactionVolumeAuditJob } from '../metric-reconciliation-audit';
+
+type Runnable = { run: (ctx: unknown) => Promise<unknown> };
+const jobContext = { checkIfCanceled: () => undefined };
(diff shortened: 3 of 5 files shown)


JustMaier:
## Adversarial review — six defects fixed in 4d552fa427, one claim disputed

Reviewer ran 9 source mutations, the new suite against the parent commit, shuffled seeds, a real prom-client scrape, and ClickHouse/Postgres re-derivations. **Two of the defects I introduced in this PR.**

### NEW-A (HIGH, introduced here) — the `await` I inserted split an anchor from its value

I placed `await republishNightlyExactnessGauges()` **between** the anchor set and the coverage set. On the parent those two lines were adjacent. Because it's an `await`, the process can serve `/api/metrics` from the same prom registry while it's pending — so on a pod's first hourly run after a roll, a scrape landing in that window sees a **fresh `audit_last_run` beside a `hourly_coverage_ratio` still at prom-client's default 0**. The freshest-pod query selects exactly that pod and reads the 0 as total CDC loss. Hourly path, so 24 chances a day.

Values now set first, anchor second, re-publish last. Pinned by an ordering assertion, since the gauge values at the *end* of the run are identical either way.

### NEW-B — my talos fix and this PR were each right and jointly broken

A null arm leaves its gauge unset, which reads as `0` — byte-identical to a measured `0.0`. Fine while the alert guarded the *value* with `!= 0`. But talos#929 now guards the **anchor** (because guarding the value hid a genuine total loss), so a null `recentCoverage` would page as total loss.

**I proposed gating the anchor on `recentCoverage`, and the reviewer refuted it with measurements — correctly.** `recentPairs == 0` requires 200 sampled images with recent ClickHouse events and *no* recent Postgres rows. That isn't a quiet site, it's mass PG-side loss. Freezing the anchor there would report *"the nightly has not run"* 26 hours later, while the nightly was running fine and had found something alarming.

Fixed instead with a companion `exactness_recent_pairs` gauge, published unconditionally, that alerts gate on — so only a **measured** zero fires. Healthy value measured across five daily seeds: 403–491, with 195–198 of 200 images carrying recent Postgres reactions.

### F3 / F4 / F6 — latent, all fixed

- **F3** — both new failure paths were `.catch(() => undefined)` with no log, the only silent paths in a file where every other failure reports to Axiom. A silently-failed snapshot write re-publishes an older snapshot, rewinding the freshness stamp and serving yesterday's coverage as current.
- **F4** — `JSON.parse(raw) as T` can't catch a shape change, and it bites silently: `undefined !== null`, so a missing field reaches `gauge.set(undefined)`, which **throws**, and the catch swallows it — no gauges published at all while the hourly reports success. Now a validating parse that logs and refuses.
- **F6** — the 7-day TTL turned a firing alert into a **false recovery**: on expiry the re-publish stops, the next roll empties the gauges, NoData resolves to OK, and an alert about a dead nightly *resolves* while it's still dead. TTL removed — a stale snapshot isn't a leak, it's the evidence, and `ranAt` makes its staleness legible.

### My null-arm test was vacuous

It asserted `toBe(0)` — which is what *both* "unset" and "published as zero" read as. The reviewer mutated the guard to publish zeros and the test **still passed**. Replaced with a pair asserting the same gauge value and opposite meaning, separated by `recent_pairs`.

All four new properties mutation-tested, not just run:

| mutation | failure message |
|---|---|
| revert the ordering fix | `expected [ 'republish', 'coverage', 'anchor' ] to deeply equal [ 'coverage', 'anchor', 'republish' ]` |
| drop `recent_pairs` | `expected +0 to be 412` |
| drop the validation | `expected 0.9997 to be +0` |
| re-add the TTL | `expected { EX: 604800 } to be undefined` |

### One finding I dispute

**F7 — the `17,247` orphaned removals reproduce.** The reviewer reported four readings (36,146 / 1,041,567 / 1,586,471 / 1,793,995), none matching — but those measure four *adjacent* quantities, not the one stated. Re-running the original query an hour later gives **17,236**, a 0.06% drift from the window moving.

The fair part of the criticism stands though: the comment gave a number without its query, which is what made "measured something adjacent" the natural failure. It now states exactly what is counted, and flags the ~0.1% drift so the next person doesn't read it as a discrepancy.

### Two claims of mine that were independently confirmed

`ranAt` is never re-stamped, and the 16 failures / 6 suites are pre-existing — the reviewer ran the same six against parent `d338259902` and got an identical set. It also confirmed `3,197,222` and the `~18k/day` baseline exactly, and verified from `SHOW CREATE TABLE` that `entityMetricEvents_month` is `SharedReplacingMergeTree` with **no version column**, so the original-`createdAt` collision argument holds.

### Environment warning for anyone testing in `C:\Dev\Repos\work\model-share`

That checkout's `node_modules` was damaged — 20 empty pnpm store dirs, and `packages/civitai-db-schema/node_modules` missing entirely. **Almost certainly my fault**, from a `cp -r node_modules` I started and a cleanup that followed. Before repair, `pnpm typecheck` reported phantom `TS2688`/`TS2307` errors and the unit suite reported **260 failures**, every one a resolution artifact that reads as "this PR broke things."

Repaired with `pnpm install --frozen-lockfile`. Verified after: typecheck **0 errors**, new suite **12/12**, full unit suite back to the honest **16 failures / 6 suites**. If you ran tests there in the last few hours, re-run them.

PR #3801, opened 2026-08-10
fix(app-listings): cap listing asset dimensions on the full-res upload path

ZacxDev:
## The defect

`LISTING_ASSET_MAX_DIMENSION_PX = 8192` is declared in `src/server/schema/blocks/app-listing.schema.ts` as the absolute per-side ceiling for **any** listing asset, and its docstring says it exists to stop a decompression-bomb / absurd-dimension image being stored and rendered as store media. It was referenced by exactly one file.

App-listing media has three ingest paths. Two applied a ceiling; the third — the one that carries user cover/screenshot uploads — applied none:

| ingest path | used by | per-side ceiling (before) |
|---|---|---|
| `ingestListingAssetFromDataUri` (inline icon) | `set-icon` | 4096 at attach (`LISTING_ICON_MAX_PX`) |
| presigned upload → `persistAssetImage` (full-res) | **`set-cover`, `add-screenshot`, the web uploader** | **none** |
| `ingestListingAssetFromUrl` (OG pull) | listing-meta | 8192 |

And the attach validator did not close the gap either: `validateListingImage`'s cover branch enforces aspect (1.3–2.4) and a 640px minimum width; the screenshot branch enforces aspect and a minimum side. Neither had a maximum — only `icon` did.

**Observed in production**: a 9000×4000 (36 Mpx) PNG cover was accepted through `civitai app listing set-cover` and stored at full dimensions (`Image` row `width=9000, height=4000`, 234,792 bytes). A 6144×3456 (21 Mpx) cover was accepted too. Aspect 2.25 is inside the cover band and 9000px is far above the 640px floor, so every rule that existed passed — nothing was misconfigured, there simply was no maximum.

Note the byte caps cannot substitute: that 36 Mpx image was 229 KiB, ~6% of the 4 MiB cover cap. Compression ratio is not a bound on decoded dimensions.

## The fix

Applies the **existing** bound — no new constant, no changed value, no new pixel-count cap — through a single shared predicate, `listingAssetTooLargeReason(subject, width, height)`, declared next to the constant it enforces. It returns the house-style rejection text (`cover is too large (max 8192px per side, got 9000px)`) or `null`.

It is called at two layers, deliberately:

1. **`measureListingAssetUpload`** (the presigned persist path) rejects on the **measured** dimensions — never the declared ones — *before* `createImage`. So oversize bytes never reach the scan pipeline and no `Image` row is created. This is the same point in the flow at which the other two ingest paths already reject.
2. **`validateListingImage`** applies the ceiling to every kind at attach. This is the choke point: it is the one gate every listing asset passes *whatever created its `Image` row*. `loadValidatedImage` gates on ownership, not provenance, so it cannot tell how a row was made — which means an ingest-side check alone is bypassable by attaching an image that reached the account through some other upload path. Layer 1 is the early reject; layer 2 is the one that makes the bound actually hold.

The rule, the bound and the message text live in one function, so the two layers cannot drift. The OG-pull path in `listing-meta.service.ts` now calls that helper instead of open-coding the same comparison and message — a de-duplication, not a third copy.

The icon keeps its own tighter 4096px maximum; the shared ceiling does not loosen it, and there is a test pinning that.

Errors are `TRPCError` `BAD_REQUEST`, so the `civitai` CLI relays them verbatim. The CLI vendors no bounds of its own and needs no change.

## ⚠️ This is a behaviour change

**Uploads that succeed today will start being refused.** Any listing icon/cover/screenshot over 8192px on either side now gets a 400 naming the bound, where it previously succeeded and was stored at full resolution.

**The guard applies to NEW uploads only.** Existing stored assets above the bound are unaffected — nothing is migrated, re-encoded or deleted, and no backfill runs. An already-attached oversize cover keeps working; it would only be refused if it were re-attached.

8192px is retained rather than repointed: it is comfortably above any real store icon/cover/screenshot (the largest legitimate asset shape here is a hero cover, and 8192px is 4× a 4K width), it is the number the constant's docstring already promised, and adopting a different bound on one path would recreate the inconsistency this PR removes.

## Tests

Added next to the existing suites, matching their mocking style. Fixtures vary **dimensions only** — every one is a flat PNG far under the 4 MiB byte cap, so the byte rule cannot account for any verdict — and both sides of the boundary are covered.

`src/server/services/blocks/__tests__/app-listing-assets.service.test.ts` (pure attach validator):
- cover at exactly 8192 passes; 8193 is refused with the bound-naming reason
- the 9000×4000 shape observed in production is refused
- the ceiling applies to the taller side, not just width (screenshot at 4500×8193 refused, 4500×8192 accepted)
- invariant guard: the icon still fails on its own tighter 4096px maximum

`src/server/services/blocks/__tests__/offsite-listing.service.test.ts` (presigned persist path, real bytes through real `sharp`, only the store accessor mocked):
- a 9000×4000 upload is refused and **`createImage` is never called**
- one pixel over the ceiling on the taller side is refused
- negative control: an upload at exactly 8192×4096 still persists at full size

Run with the new test files applied to an otherwise-unmodified checkout of the branch base (`670cc946b4`):

```
base   →  Tests  5 failed | 162 passed (167)
HEAD   →  Tests  203 passed (203)   [5 suites: the 2 above + listing-meta ×2 + block-image-upload.persist]
```

The 5 red-at-base failures are the regression tests; e.g. the 9000×4000 cover case fails with `expected { ok: true } to deeply equal { ok: false, reason: "cover is too large (max 8192px per side, got 9000px)" }` — the base genuinely accepted it. The other 2 new cases (icon tighter-max, at-the-ceiling negative control) pass on both sides and are labelled as an invariant guard and a negative control, not as regression coverage.

### Mutation checks

Each guard was broken on purpose and attributed by the **specific** failing message, not merely by "a test went red":

| mutation | result |
|---|---|
| delete the ceiling check in `validateListingImage` | 3 attach tests fail with *this* guard's text (`"cover is too large (max 8192px per side, got 9000px)"` vs received `{ ok: true }`); the 2 persist tests still pass, so the failure is attributable to that guard alone. Received `{ ok: true }` also proves the check is **reachable** — no earlier rule intercepts these fixtures |
| delete the throw in `measureListingAssetUpload` | 2 persist tests fail with `promise resolved "{ imageId: 12345 }" instead of rejecting` — i.e. the mutant reproduces the exact defect (row created, sent to scan); the 3 attach tests still pass |
| raise the bound in the shared predicate (`<= MAX` → `<= MAX * 2`) | all 5 regression tests fail — both call sites do depend on the single predicate, not on a local copy |
| keep the guard but stop it naming the bound (`message: 'That image is too large.'`) | 2 persist tests fail on the **message**, proving the assertions pin the bound-naming text so an adjacent guard's error could not produce a false green |

### Other verification

- `node scripts/typecheck.mjs`: 77 errors at base, 77 at HEAD, **identical error sets** (pre-existing, all in unrelated files — generated-client and workspace-build artifacts of a fresh checkout). Zero reference the touched files.
- The repo's lint-rule suites (`no-unloadable-image-fixture`, `no-wholesale-module-mock`, `no-io-in-transaction`, `no-module-scope-cache`, `no-unbounded-paging-fake`): 215 passed.
- Wider sweep of `src/server/services/blocks/__tests__`, `src/server/routers/__tests__`, `src/server/schema`: 2995 passed, 2 failed — both in `manifest-schema-endpoint.test.ts`, and both fail identically on an unmodified base checkout, so they are pre-existing and unrelated. (26 further files fail to *collect* in my local worktree for environment reasons — an unbuilt workspace package and the Prisma engine — equally at base; CI resolves those.)
- Prettier: the touched files are flagged identically at base and at HEAD, so no new formatting violation is introduced (these files carry substantial pre-existing drift and were left alone rather than reformatted, to keep the diff readable).

No Prisma migration is needed and none is included.

### DIFF
--- a/src/server/schema/blocks/app-listing.schema.ts
+++ b/src/server/schema/blocks/app-listing.schema.ts
@@ -65,6 +65,10 @@ export const LISTING_SCREENSHOT_MIN_PX = 320;
  * (`INLINE_ICON_MAX_INPUT_PIXELS`, 16.7 Mpx in `listing-meta.service.ts`), which is
  * the metric that threat actually calls for. Bounding area on the upload path is a
  * known gap, not something this constant covers.
+ *
+ * (8192 is retained because it is the value the OG-pull path already enforced, so
+ * this makes the paths agree. It is ~2× a 4K width, not 4× — 4× is the AREA ratio
+ * against 4096².)
  */
 export const LISTING_ASSET_MAX_DIMENSION_PX = 8192;
 
@@ -82,13 +86,19 @@ export const LISTING_ASSET_MAX_DIMENSION_PX = 8192;
  *
  * 🔴 SCOPE — the attach check is a BACKSTOP, not a chokepoint. Two things it does
  * not cover, so nobody builds on a guarantee that isn't here:
- *   1. `backfillListingAssets` / `migrateBlockScreenshots`
- *      (`app-listing-assets.service.ts`) write listing assets via direct `dbWrite`
- *      calls and never call {@link validateListingImage} at all.
+ *   1. The backfill route, in `app-listing-assets.service.ts`, never calls
+ *      {@link validateListingImage} at all: `backfillListingAssets` writes the
+ *      listing's own rows (`appListingScreenshot.createMany`, `appListing.update`
+ *      of `coverId` / `iconId`) with direct `dbWrite` calls, and the
+ *      `migrateBlockScreenshots` dep it uses creates the backing `Image` rows at
+ *      whatever dimensions the decoder reports.
  *   2. At attach the dimensions are read off the `Image` row, and some creation
- *      paths accept width/height from the client (see the note in
- *      `measure-uploaded-image.ts`) — so for a row this feature did not measure,
- *      the bound is only as honest as the path that made it.
+ *      paths accept width/height from the client — see
+ *      `src/server/utils/stored-image-probe.ts`, which states it outright ("a
+ *      CLIENT CLAIM about bytes the server never looked at"); the sibling note in
+ *      `services/blocks/measure-uploaded-image.ts` draws the consequence. So for a row
+ *      this feature did not measure, the bound is only as honest as the path that
+ *      made it.
  */
 export function listingAssetTooLargeReason(
   subject: string,
--- a/src/server/schema/blocks/app-listing.schema.ts
+++ b/src/server/schema/blocks/app-listing.schema.ts
@@ -54,11 +54,17 @@ export const LISTING_SCREENSHOT_ASPECT_MAX = 2.6;
 export const LISTING_SCREENSHOT_MIN_PX = 320;
 
 /**
- * Absolute upper bound on either side of ANY listing asset. A ceiling (the
- * per-kind checks above only enforce MINIMUMS / aspect) that stops a
- * decompression-bomb / absurd-dimension image (e.g. 100000×100000) from being
- * stored and rendered as store media. 8192px is comfortably above any real store
- * icon/cover/screenshot.
+ * Absolute upper bound on EITHER SIDE of any listing asset. A ceiling (the
+ * per-kind checks above only enforce MINIMUMS / aspect) that keeps an
+ * absurdly-dimensioned image (e.g. 100000×100000) from being stored and rendered
+ * as store media. 8192px is comfortably above any real store icon/cover/screenshot.
+ *
+ * 🔴 It bounds SHAPE, not AREA — do not cite it as the decompression-bomb defence.
+ * 8192×8192 is 67 Mpx and passes this check; so does 8192×4000 (33 Mpx). The
+ * inline-icon ingest path bounds decoded PIXEL COUNT separately
+ * (`INLINE_ICON_MAX_INPUT_PIXELS`, 16.7 Mpx in `listing-meta.service.ts`), which is
+ * the metric that threat actually calls for. Bounding area on the upload path is a
+ * known gap, not something this constant covers.
  */
 export const LISTING_ASSET_MAX_DIMENSION_PX = 8192;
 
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Correction to this PR's description — three claims above are overstated

An adversarial review of this branch found that the description (and the docstrings in the first commit) assert guarantees the code does not provide. Correcting publicly rather than quietly editing the body, since it has already been readable in its original form. `0db64958` fixes the in-code comments; **no behaviour changed and no test changed** — the diff is comments-only and the suite is still 167 green.

**1. This does not bound megapixels, so it is not by itself a decompression-bomb defence.**
The ceiling is per-SIDE. `8192×8192` is **67 Mpx** and passes both layers; so does `8192×4000` (33 Mpx). Notably the 6144×3456 (21 Mpx) cover cited as motivation is **not** rejected by this change — only the 9000×4000 one is. The sibling inline-icon ingest path caps decoded pixel *count* at ~16.7 Mpx, i.e. 4× stricter, which is the metric the stated threat actually calls for. An area bound on the upload path remains a **gap**; happy to do it as a follow-up.

**2. `validateListingImage` is not "the gate every asset passes".**
It has one production call site. `backfillListingAssets` and `migrateBlockScreenshots` write listing `coverId` / `iconId` / screenshot rows through direct `dbWrite` calls and never call it. The attach check is a **backstop**, not a chokepoint.

**3. The attach layer does not close an ingest bypass the way the description implies.**
At attach the dimensions are read off the `Image` row, and some creation paths accept client-supplied `width`/`height`. So for a row this feature did not itself measure, the bound is only as honest as the path that created it — a property this repo already documents in `measure-uploaded-image.ts`. The bound genuinely holds for the measured paths.

### What still stands

The shipped behaviour is unaffected by all three — they are description/comment defects, not bugs. Independently reproduced: **red at base** (`5 failed | 162 passed`) → **green at HEAD** (`167 passed`), and a 14-mutant semantic sweep built independently of the PR's own four. Twelve mutants died; the two survivors are benign (an unpinned check *ordering*, and a `width`/`height` swap into a symmetric `Math.max`).

### Known, not fixed here

- An over-8192px **icon** now reports the 8192 per-side limit rather than its own tighter 4096 one, so a user who resizes to 8192 is refused a second time. Low reachability, and the check ordering is unpinned by any test.
- A rejected upload leaves an orphaned stored object (pre-existing for the byte-cap and unsupported-format rejects; this adds one more class).
- No client-side dimension pre-check, so a refused cover still costs the full upload round trip.
- Coverage note: two of the mutants are each killed by exactly one test, so all non-cover / second-operand coverage rests on a single case.

ZacxDev:
### Addendum — one more figure in the description is wrong

Following a delta re-audit of the correction above. `3a580fdc` (comment-only again; transpiled emit byte-identical to the previous commit, 167 tests still green).

**The description's "8192px is ~4× a 4K hero width" is wrong.** 8192/3840 = **2.13×** (UHD), or 2× against DCI 4096. 4× is the **area** ratio against 4096², not a width ratio. This matters because that figure is the only justification the description offers for keeping 8192 rather than repointing the bound — so anyone re-deriving the choice from it starts from a wrong number. The actual reason to keep 8192, now recorded in the constant's docstring, is simply that it is the value the OG-pull path already enforced, so this makes the two paths agree.

Also tightened in the same commit, both found by the re-audit:

- The client-supplied-dimensions claim pointed at `measure-uploaded-image.ts`, which states the consequence *conditionally* and never asserts such paths exist. Repointed to `src/server/utils/stored-image-probe.ts`, which says it outright. Both references are now spelled as full paths — the two files sit in different directories, so the bare filenames sent a reader to the wrong place.
- The backfill-bypass note attributed direct `dbWrite` listing writes to both named functions. Split accurately: `backfillListingAssets` writes the listing's own rows; the `migrateBlockScreenshots` dep creates the backing `Image` rows.

No behaviour has changed across any of the three doc commits — the shipped fix is still exactly `5b7dbad8`.

Issue #3771, opened 2026-08-09
Presigned image-upload PUT stays writable for 12h with no conditions, so measured image metadata can go stale after validation


ZacxDev:
`/api/v1/image-upload` mints a presigned PUT with
`getSignedUrl(PutObjectCommand{ Bucket, Key }, { expiresIn: UPLOAD_EXPIRATION })` — **12 hours**,
with no `content-length-range` condition and no content-type condition. The key stays writable for
that whole window, and re-PUTting it replaces the object's bytes while any database row describing
those bytes stays as it was.

## Why it matters now

PR #3738 makes `persistListingAssetImage` read the uploaded object back and derive
`width`/`height`/`mimeType`/`size` from the actual bytes, replacing the previously client-declared
values. That is a real improvement, but the guarantee it provides is **point-in-time**: it describes
the bytes as they were when the probe ran, not as they are when the image is attached or served.

Sequence: PUT an image → persist (the server measures it and stores the measurements) → re-PUT
anything to the same key inside the remaining window → attach validates against the stale row.

The #3738 docstrings originally said the persisted columns "describe the stored bytes"; that wording
has been softened, but the underlying window is still open.

## Scope beyond listings

This is not specific to listing assets. The same unconditioned 12-hour PUT window means the content
scan can be defeated the same way — scan the bytes that were uploaded, then replace them. That is
pre-existing and site-wide rather than introduced by #3738, which is why this is filed separately
rather than as a blocker on it.

## Possible mitigations

- **Capture the `ETag` at probe time and re-check it at attach.** Cheap, and closes the window for
  the listing path specifically.
- **Server-side copy to a fresh, server-controlled key after probing**, so the key the row points at
  was never client-writable. Strongest, costs a copy.
- **Shorten `UPLOAD_EXPIRATION`** and/or add a `content-length-range` condition. Reduces the window
  without closing it.

Not executed against live storage — established by reading the signing code and the S3 semantics, so
treat the mechanism as confirmed and the live behaviour as untested.

Found during an adversarial review of #3738. Filed from `civitai/cli#270`.

PR #3804, opened 2026-08-11
fix(app-listings): re-verify listing media at attach, not only at upload

ZacxDev:
Closes the gap described in #3771, for the listing-media path.

## The problem

A listing asset's `width` / `height` / `mimeType` / byte size are derived server-side from the bytes read back out of the store, so that the attach rules expressed over those columns are not self-reported. That derivation is **point-in-time**, and the measurement module says so in its own header:

> 🔴 A POINT-IN-TIME reading, not a durable property of the key. […] Callers may record "these values were measured from real bytes", never "these values describe the object".

Nothing acted on that caveat. The attach gate validates the persisted columns and never re-checks the object, so a row can end up describing bytes that are no longer the ones stored at that key — and the gate then validates a description rather than the media it is about.

## The change

Persist records the store's entity tag alongside the measurement; the listing attach gate re-reads it and refuses the attach when the two disagree.

**Fails closed on disagreement only.** A row with no recorded tag (written before this existed, or by a path that never measures an uploaded object), an object the store reports as gone, or a store that cannot be consulted are all classified `unverifiable` and pass through. None of them is evidence that anything changed, and rejecting on them would turn a transient store problem into a blocked attach for images that are perfectly fine — the same three-state posture already documented on the existing head-object helper. The extra request is issued **only** for rows that actually carry a tag, so anything that cannot be checked pays nothing.

The comparison rule is a pure module, so it is unit-testable on its own and lives in one place rather than being re-derived at each of the three attach procs.

**No schema change** — the tag rides in the existing `metadata` column beside the byte size.

## Why this scope

The broader alternative — re-keying each object server-side after persist and relying only on the copy — was scoped first and rejected. The presigned-upload endpoint and its presign helpers back the general site-wide media path (the shared upload hooks across dozens of components, plus vault and training assets), and the stored key is what CDN URLs, existing rows and the scan pipeline all resolve against. Changing it site-wide to fix a listing-media guarantee is the wrong trade. This PR closes the path where the measurement guarantee is load-bearing and leaves the general upload path untouched.

Worth noting separately, and **not** addressed here: the same staleness window exists for as long as an upload grant remains valid, which is a knob on the shared upload path rather than something this gate can shorten.

## Tests

`red at 1264c8e1e4 (origin/main) → green at HEAD`, measured by running the new suite against a clean checkout of the base commit:

| | base | HEAD |
|---|---|---|
| new integrity suites | **3 failed** / 8 passed | 31 passed |
| whole `blocks` service suite | 2291 passed / 2293 | 2323 passed / 2325 |

The 2 failures in both columns are pre-existing and unrelated (`manifest-schema-endpoint`). The base failure for the core case reads `promise resolved "{ status: 'attached', … }" instead of rejecting` — the gap itself, reproduced.

The main suite is a **seam test**, deliberately not two isolated ones: persist *writes* the token and attach *reads* it back, and each half is trivially "correct" alone while the pair does nothing at all if they disagree about the metadata key or the comparison. So every case runs the real persist proc, takes the metadata blob it actually produced, puts that on the row the real attach proc loads, and only then varies the store. Entity tags in the fixtures are derived from the fixture bytes, as a real store derives them, so two fixtures cannot accidentally share one.

Also included: full cross-product coverage of the pure rule (recorded tag present/absent × store present/absent/unreachable × live tag present/absent), a positive control that the accept path actually issues the check rather than skipping it, and a case proving the refusal fires even when the replacement decodes to identical geometry.

The existing fixture was additionally made faithful to the real store response, which omitted the tag entirely — so every assertion about persisted metadata there had been holding via the absent-tag branch, and the recorded-tag path looked covered while never running.

## Mutation results

10 mutants, all killed; baseline restored green afterwards.

| Mutant | Killed by |
|---|---|
| remove the guard's call site | reachability control (`expected [] to have length 1`) + the regression case |
| invert the verdict (throw on agreement) | 7 tests, both directions |
| keep the call, make the body unreachable | reachability control — the accept path stops issuing the check |
| **keep the guard, drop its specific error** | regression case fails on the message, so it is bound to *this* guard's error and not an adjacent one |
| also reject when the check could not run | the three fail-open cases |
| classifier can never disagree | pure case + both refusal cases |
| compare the live tag against itself | pure case + both refusal cases |
| read the neighbouring metadata field | 6 tests across both suites |
| drop the quote-normalisation step | 9 tests — the positive control on that text-stripping step |
| let two absent tags compare equal | the "never MATCH when either side has no tag" case |

## Checks

- No existing validation weakened or removed; two cases pin that the ownership and geometry rejections still fire for their own reasons, with their own messages.
- `pnpm typecheck`: applying this diff to a clean base checkout produces a **byte-identical error set** to that checkout without it (34, all pre-existing: an unbuilt workspace package and stub modules). Zero added.
- `pnpm prettier:check` green — validated against a known-bad input first, since its pass message is also what a zero-file run prints.
- `pnpm test:lint-rules` 215 passed.

## Known limits of this gate — please read before merging

It re-checks **at attach**, which is where the persisted measurement is first relied upon. Two things it therefore does **not** cover, stated plainly rather than implied:

1. **A change that happens after the last attach.** Nothing re-verifies at the go-live gate, which reads scan state only. An asset attached while intact and altered afterwards would not be caught by this PR.
2. **Scan freshness.** The content scan is performed against the bytes as they were at ingestion time, so the same staleness argument applies to the scan verdict independently of geometry.

3. **Coverage is bounded by provenance, and the gate does not establish provenance.** The re-check compares against a token recorded at measurement time. The attach gate selects the row by id and confirms ownership; it does not determine which path created the row or where its recorded values came from. So a row can reach the gate with no recorded token and classify as `unverifiable` (allowed through, exactly as an older row is), **or** carrying a token the measuring paths did not produce, in which case the comparison can return an affirmative `match` — and `match` is the silent outcome. The gate therefore raises the floor for rows the measuring procs minted; it is not a guarantee about every row that can be attached, and it should not be read as one. Making the token a precondition rather than evidence is a separate decision, and would need a story for rows that predate it.

Items 1 and 2 are the same underlying class as #3771 and both close naturally if the recorded token is also re-checked at the go-live gate. I did not extend it there in this PR because that gate is on the publish path and deserves its own scoping decision about what an `unverifiable` verdict should do when a listing is about to become visible — the fail-open posture that is clearly right at attach is not obviously right there.

### DIFF
--- a/src/server/services/blocks/__tests__/listing-asset-upload-integrity.test.ts
+++ b/src/server/services/blocks/__tests__/listing-asset-upload-integrity.test.ts
@@ -3,6 +3,7 @@ import { GetObjectCommand, HeadObjectCommand } from '@aws-sdk/client-s3';
 import sharp from 'sharp';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import type * as LoggingClient from '~/server/logging/client';
 import type * as S3Utils from '~/utils/s3-utils';
 
 /**
@@ -27,14 +28,27 @@ import type * as S3Utils from '~/utils/s3-utils';
  * The dimension every case varies is the identity of the stored object, and the
  * fixtures cross it in both directions: an untouched object attaches (proving the
  * guard is not a blanket reject), a replaced one does not.
+ *
+ * 🔴 SECOND DIMENSION: THE ASSET KIND. The guard is placed ONCE, in the shared
+ * `loadValidatedImage`, precisely so all three attach procs inherit it rather than
+ * re-deriving it — which means "it is placed once" is a claim about all three, and
+ * covering only `icon` leaves it pinned for none of the others. A guard narrowed to
+ * a single kind is a one-word edit away and, covered from one kind only, is
+ * INVISIBLE. So the accept/refuse pair below is run across icon, cover AND
+ * screenshot, each through its own real proc.
  */
 
-const { mockRead, mockWrite, mockCreateImage, mockS3Send } = vi.hoisted(() => {
+const { mockRead, mockWrite, mockCreateImage, mockS3Send, mockLogToAxiom } = vi.hoisted(() => {
   const db = {
     appListing: {
       findUnique: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
       findFirst: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
       update: vi.fn(async (..._a: unknown[]) => ({})),
+    },
+    appListingScreenshot: {
+      findMany: vi.fn(async (..._a: unknown[]): Promise<unknown[]> => []),
+      count: vi.fn(async (..._a: unknown[]) => 0),
+      create: vi.fn(async (..._a: unknown[]) => ({})),
     },
     image: {
       findUnique: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
@@ -50,10 +64,17 @@ const { mockRead, mockWrite, mockCreateImage, mockS3Send } = vi.hoisted(() => {
     mockWrite: db,
     mockCreateImage: vi.fn(async (..._a: unknown[]) => ({ id: 777 })),
     mockS3Send: vi.fn(),
+    mockLogToAxiom: vi.fn(async (..._a: unknown[]) => undefined),
   };
 });
 
 vi.mock('~/server/db/client', () => ({ dbRead: mockRead, dbWrite: mockWrite }));
+// Only the SINK is replaced — every field the service decides to emit is the real
+// one, so an assertion here is about what an operator would actually receive.
+vi.mock('~/server/logging/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof LoggingClient>()),
+  logToAxiom: mockLogToAxiom,
+}));
 vi.mock('~/server/services/image.service', () => ({ createImage: mockCreateImage }));
 vi.mock('~/server/utils/app-block-ids', () => ({
   newAppListingId: () => 'apl_test_1',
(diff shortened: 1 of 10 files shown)


ZacxDev:
## Correction to the test matrix in the description

I re-measured the numbers in the description's **Tests** section myself and they were wrong. Correcting them here rather than editing the body, since a reviewer may already have read the original.

**Base** = `1264c8e1e4`, this branch's merge base on `main`. **Method**: the changed source files reverted to the base revision with the test files left at the branch tip — i.e. the branch's tests run against base code.

### As the description framed it (test files at `b36bc79f11`)

| suite | at base | at `b36bc79f11` |
|---|---|---|
| `stored-object-integrity.test.ts` | **0 collected — the suite fails to load** | 20 passed |
| `listing-asset-upload-integrity.test.ts` | **4 failed / 7 passed** | 11 passed |
| `offsite-listing.service.test.ts` | 1 failed / 77 passed | 78 passed |
| **total** | **89 collected, 5 failed / 84 passed** | **109 passed** |

Three things to fix in the record:

1. **"3 failed / 8 passed" is wrong — it is 4 / 7.**
2. **20 of the "31 passed at HEAD" never ran at base at all.** `stored-object-integrity.test.ts` imports a module that does not exist on base, so the entire file fails collection. Vitest reports that as `(0 test)`, not as 20 failures — so those 20 contribute *nothing* to a red-at-base claim. They are **new-code unit coverage**, not regression coverage.
3. **Of the 5 genuinely red at base, only 2 reproduced the defect behaviourally** (`promise resolved "{ status: 'attached', … }" instead of rejecting`). The other 3 were red because they assert code the base does not have — a missing module, a request base never issues, a metadata key base never writes.

So the honest split for `b36bc79f11` is **2 regression tests + 29 new-code unit tests**, not "31 green, red at base". The description also omitted that a third suite (`offsite-listing.service.test.ts`) carried one of the base failures.

### After `d853abb6a6`

That commit adds the cover + screenshot coverage this was missing, a bound on the re-read, and reporting of every non-agreeing verdict. Same measurement, same base:

| | at base `1264c8e1e4` | at `d853abb6a6` |
|---|---|---|
| four suites | 110 collected, 20 more uncollectable, **25 failed / 85 passed** | **130 passed** |

Of those 25, **6 reproduce the original defect** — the swapped-object refusal and the identical-geometry refusal, one pair for each of icon, cover and screenshot, each failing on base with `promise resolved "{ status: 'attached', … }" instead of rejecting`. That is the number I would call regression coverage. The other 19 assert behaviour that does not exist on base and are new-code coverage.

The reason cover and screenshot are now covered at all: the check is placed once, in the shared image loader, so all three attach procs inherit it — and with only the icon path exercised, restricting the call to a single kind left the whole suite green. It now fails 7 tests, for cover and for screenshot, on the check's own error.

### Also correcting

- **Typecheck**: I measure **35** pre-existing errors, identical at `b36bc79f11` and at `d853abb6a6`, none in any file this PR touches — delta zero. I did not re-measure a clean base checkout, so I can neither confirm nor refute the description's "34"; what I can state is that this branch adds none.
- The mutation table in the description covers the pure rule and the guard's call site. The sweep for `d853abb6a6` extends it to 26 mutants over the same code plus the new behaviour (the per-kind placement, the ordering against the column validation, the bound on the re-read, and each field of the verdict report), all killed, with the failing test's message checked against the specific check under test in each case.

ZacxDev:
Added a third item to **Known limits** in the body (noting it here rather than editing silently, since the body has been readable for a while).

**3. Coverage is bounded by provenance.** The gate compares against a token recorded at measurement time, and only the persist procs that actually measure an uploaded object record one. The attach gate selects by id and confirms ownership; it does not establish which path created the row, so a row that arrived by another route carries no recorded token and classifies as `unverifiable` — allowed through, exactly as an older row is.

This is the honest ceiling on what the PR buys: it raises the floor for rows the measuring paths minted, and is not a universal guarantee about every row that can be attached. Nothing here regresses — it is a pre-existing property of how the attach gate selects rows — but it belongs stated rather than implied. Making the token a precondition rather than evidence is a separate decision needing a story for rows that predate it.

ZacxDev:
Correcting my own wording in **Known limits item 3** — the earlier version was too generous, and since that section is labelled "please read before merging" it matters that it is right.

**What I wrote:** a row arriving by another route "carries no recorded token and classifies as `unverifiable`".

**What is actually true:** that is only one of the two cases. A row can also reach the gate *carrying* a token the measuring procs never produced, in which case the comparison can return an affirmative **`match`** — and `match` is the silent outcome, so the more deliberate case is the one this gate cannot see. The attach gate selects by id and confirms ownership; it does not establish which path created the row or where its recorded values came from.

Nothing here is a regression — the mechanism predates this PR and is not made worse by it, which is why it belongs in Known limits rather than blocking. But the honest ceiling is "raises the floor for rows the measuring procs minted", not "covers everything except older rows". The body is updated.

Separately, a follow-up rather than a change here: the bound added in this PR covers the head request, while the ranged read in the same probe module is still unbounded — and it sits on the same user-facing mutation and actually transfers bytes. The reasoning in the new comment for bounding the one applies at least as strongly to the other.

PR #3805, opened 2026-08-11
feat(app-blocks): App Listing Collaborators — editor seats, ownership transfer, display-author

ZacxDev:
Adds **App Listing Collaborators**: a consent-gated *editor* seat on an App Block, an owner-initiated / recipient-accepted **ownership transfer**, and a public **display-author** byline opt-in.

An editor is effectively a co-owner — listing content + media, submit new versions, view analytics, and see earnings for the shared app. The only things reserved to the owner are **managing collaborators** and **initiating a transfer**.

---

## 🔴 Prisma migrations are NEVER auto-applied here

The main civitai DB (CNPG nvme0) does **not** run `prisma migrate deploy`; no deploy path applies migrations. The `.sql` below is committed **for history only** and must be **applied by a human** (psql/retool) per environment — prod nvme0 **and** the dev clone.

**The feature is INERT until then, by construction.** Every read of the new tables goes through `safeCollaboratorQuery` (`app-access.service.ts`), which catches the missing-relation error (Prisma `P2021` / PG `42P01`) and degrades to "no collaborators" — i.e. byte-identical to today's owner-only behaviour. It does **not** blanket-catch: a connection failure or a genuine query bug still propagates, so this cannot become a permanent silent-zero. Merging and deploying before the SQL is applied changes nothing observable and does not 500 the app pages.

### Raw SQL to apply

`packages/civitai-db-schema/prisma/migrations/20260810140000_app_listing_collaborators/migration.sql`

```sql
-- 1. app_collaborators — the consent-gated editor seat (+ byline opt-in)
CREATE TABLE IF NOT EXISTS "app_collaborators" (
  "app_block_id"     TEXT    NOT NULL REFERENCES "app_blocks"("id") ON DELETE CASCADE,
  "user_id"          INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
  "role"             TEXT    NOT NULL DEFAULT 'editor',
  "status"           TEXT    NOT NULL DEFAULT 'pending',
  "displayed"        BOOLEAN NOT NULL DEFAULT true,
  "invited_by"       INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
  "last_notified_at" TIMESTAMPTZ,
  "created_at"       TIMESTAMPTZ NOT NULL DEFAULT now(),
  "responded_at"     TIMESTAMPTZ,
  PRIMARY KEY ("app_block_id", "user_id"),
  CONSTRAINT "app_collaborators_status_check" CHECK ("status" IN ('pending','accepted','rejected')),
  CONSTRAINT "app_collaborators_role_check"   CHECK ("role" IN ('editor'))
);
CREATE INDEX IF NOT EXISTS "app_collaborators_user_status_idx" ON "app_collaborators" ("user_id","status");
CREATE INDEX IF NOT EXISTS "app_collaborators_app_status_idx"  ON "app_collaborators" ("app_block_id","status");
CREATE INDEX IF NOT EXISTS "app_collaborators_invited_by_idx"  ON "app_collaborators" ("invited_by");

-- 2. app_ownership_events — append-only audit trail (every FK nullable + SET NULL so
--    an event outlives the app, the actor and the target)
CREATE TABLE IF NOT EXISTS "app_ownership_events" (
  "id"             TEXT PRIMARY KEY,                                        -- aoe_<ULID>
  "app_block_id"   TEXT    REFERENCES "app_blocks"("id") ON DELETE SET NULL,
  "slug"           TEXT    NOT NULL,
  "action"         TEXT    NOT NULL,
  "actor_user_id"  INTEGER REFERENCES "User"("id")       ON DELETE SET NULL,
  "target_user_id" INTEGER REFERENCES "User"("id")       ON DELETE SET NULL,
  "metadata"       JSONB,
  "created_at"     TIMESTAMPTZ NOT NULL DEFAULT now(),
  CONSTRAINT "app_ownership_events_action_check" CHECK ("action" IN (
    'invite','accept','reject','remove','leave','display',
    'transfer_initiated','transfer_accepted','transfer_cancelled'))
);
CREATE INDEX IF NOT EXISTS "app_ownership_events_app_idx"   ON "app_ownership_events" ("app_block_id","created_at" DESC);
CREATE INDEX IF NOT EXISTS "app_ownership_events_actor_idx" ON "app_ownership_events" ("actor_user_id","created_at" DESC);

-- 3. app_ownership_transfers — an in-flight owner→recipient transfer
CREATE TABLE IF NOT EXISTS "app_ownership_transfers" (
  "id"           TEXT PRIMARY KEY,                                    -- aot_<ULID>
  "app_block_id" TEXT    NOT NULL REFERENCES "app_blocks"("id") ON DELETE CASCADE,
  "from_user_id" INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
  "to_user_id"   INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
  "status"       TEXT    NOT NULL DEFAULT 'pending',
  "expires_at"   TIMESTAMPTZ NOT NULL,
  "created_at"   TIMESTAMPTZ NOT NULL DEFAULT now(),
  "responded_at" TIMESTAMPTZ,
  CONSTRAINT "app_ownership_transfers_status_check"   CHECK ("status" IN ('pending','accepted','rejected','cancelled','expired')),
  CONSTRAINT "app_ownership_transfers_not_self_check" CHECK ("from_user_id" <> "to_user_id")
);
-- 🔴 The one-in-flight-transfer guard. Prisma cannot express a partial unique index,
-- so this lives ONLY here; the service relies on its P2002.
CREATE UNIQUE INDEX IF NOT EXISTS "app_ownership_transfers_one_pending_per_app"
  ON "app_ownership_transfers" ("app_block_id") WHERE "status" = 'pending';
CREATE INDEX IF NOT EXISTS "app_ownership_transfers_app_status_idx" ON "app_ownership_transfers" ("app_block_id","status");
CREATE INDEX IF NOT EXISTS "app_ownership_transfers_to_status_idx"  ON "app_ownership_transfers" ("to_user_id","status");
```

All three tables are brand-new and empty — no backfill, no meaningful lock, `IF NOT EXISTS` throughout so a re-run is a no-op.

---

## Why the seat is keyed to `AppBlock`, and why `displayed` lives on the seat row

Ownership is canonically `OauthClient.userId`, reached as `AppBlock.app.userId`; `AppListing.userId` is a **denormalized copy**.

`displayed` (the byline opt-in) is stored on `AppCollaborator`, **not** as an `AppListing` column, and that is load-bearing rather than incidental: `applyApprovedRevision`'s **offsite branch** copies the shadow's `name/tagline/description/category/contentRating/externalUrl/connect*` straight onto the live parent. The same flag held as a listing column would be silently **clobbered** by the next approved revision — an immediate-apply setting that reverts on the next mod re-review. Collaborator rows sit outside **both** branches' copy sets, which makes immediate-apply correct *by construction*. Pinned structurally in `app-collaborator.revision-non-clobber.test.ts`.

## 🔴 The earnings leak, and how it is avoided

Every pre-existing earnings read keys on `BlockBuzzAttribution.appOwnerUserId` — a **user-wide** filter that does not mention the app:

- `blocks.getMyApps` — `groupBy appBlockId WHERE appOwnerUserId = caller`, no app filter at all
- `getRevenueForOwner` / `getRecentAttributionsForOwner` — `appBlockId` is an *optional* narrowing filter, never checked against what the caller may see

The obvious way to serve an editor is to reuse one of those with the **owner's** id (the editor has no attribution rows of their own). That hands the editor **the owner's entire portfolio**. It looks correct in a one-app fixture and is catastrophic with two.

Instead there is a new app-scoped path (`app-collaborator-earnings.service.ts`) that resolves the permitted app-id set **first** and filters `appBlockId IN (thatSet)`, plus a second axis on the app's **current** owner so a transferred app's pre-transfer rows are excluded. `blocks.getMyApps` now routes through it.

`app-collaborator-earnings.service.test.ts` fixtures an owner with **2 apps** and an editor on **1**. It was written against a deliberately naive implementation first and **watched to fail** — `expected [ 'ab_appA', 'ab_appB' ] to deeply equal [ 'ab_appA' ]` — before the real one was written.

`blocks.getMyRevenue` is deliberately **not** widened: it answers "what have *I* accrued", and it must keep showing an ex-owner the money they accrued **before** transferring an app away.

## Consolidated permission predicate + the disagreements it surfaced

`resolveAppAccess(appBlockId, userId) → { role: 'owner' | 'editor' | null }` (plus `resolveListingAccess`, `resolveAccessibleAppBlockIds`, `assertAppEditAccess`) now backs every gate. `resolveListingAccess` is **shadow-aware**: a shadow revision carries `appBlockId: null`, so the seat is resolved via its parent — without that hop an editor would lose access the instant their first media edit minted the shadow.

Consolidating surfaced four genuine inconsistencies. They are recorded as data in the seam-guard ledger rather than silently normalised, because "fix" here means a behaviour change that deserves its own decision:

| | Finding | Disposition |
|---|---|---|
| **D1** | **Mod bypass is inconsistent.** `app-listing-assets::loadOwnedListing` bypasses for moderators; `offsite-listing::loadOwnedEditableListing` and `offsite-moderation::loadOwnedListingInTx` deliberately do not; the four `blocks.router` gates do not. | Left exactly as-is; asserted in the matrix as it *is*. |
| **D2** | **Ban checks were asymmetric.** `getMyAppRepo`, `updateManifest`, `getMyForgejoCloneInfo` each had an explicit `bannedAt` re-check; **`getMyAppManifest` did not.** | **Fixed** — plainly an omission, not a decision. |
| **D3** | **The ownership key differs.** `withdrawExternalRequest` keys on `AppBlockPublishRequest.submittedByUserId` (the *submitter*), which after a transfer is deliberately not the owner. | Left as-is — withdrawing your own submission is a different question. |
| **D4** | **Earnings scoping diverges between siblings.** `getMyAppAnalytics` resolves a permitted-id set; `getMyRevenue` is `appOwnerUserId`-keyed and user-wide. | Left as-is, deliberately (see above). |

Two further findings recorded in the ledger:

- **A third gate shape exists.** `src/pages/api/v1/blocks/dev-token.ts` uses a positive-match *branch condition* rather than a throw or a where-clause, and is explicitly built to be **oracle-free**. Not routed through the predicate — that would change externally-observable behaviour for an unrelated reason. A collaborator cannot mint a dev token today.
- 🔴 **KNOWN GAP (pre-existing, not introduced here).** `publish-request.service::submitVersion`'s **subsequent-version** branch never checks that the submitter owns the existing app. It is masked today because both HTTP callers require `isModerator`. If that is ever opened to non-mod authors it becomes a **slug-hijack vector**.

## Ban interaction — the decision

**A banned user may not be invited, may not accept an invite, and may not receive an ownership transfer.** A seat mints Forgejo `write` on the app repo and exposes the app's earnings, and `getMyAppRepo` already refuses to issue a push credential to a banned account — seating a banned user would hand out exactly the credential that gate withholds. A banned user **may** decline an invite (that takes nothing from anyone).

An **existing** seat is *not* auto-revoked on ban — that needs a ban-hook this PR does not add — but every capability the seat unlocks re-checks `bannedAt` at the proc, so a banned editor holds an inert row.

## Forgejo lifecycle (in scope, not a follow-up)

`forgejo.service` could previously only ever **grant**: `addCollaborator` existed with nothing to undo it, so a developer handed `write` kept it forever. This adds `removeCollaborator` and wires the full lifecycle — grant on invite-accept, revoke on remove/leave, revoke-old + grant-new on transfer-accept.

Both directions are best-effort (Forgejo is external; an outage must not roll back a consent decision). Being precise about the residual risk: a failed **grant** self-heals, because `getMyAppRepo` re-grants on every call. A failed **revoke** does **not** self-heal — the seat is gone from the DB so every civitai-side capability closes immediately, but the user may retain Forgejo push until someone notices. That residue cannot deploy anything (a push only parks a `pending` request a moderator must approve). A durable retry/reconcile job is the correct fix and is **not** in this PR; the `app-repo-access` / `revoke-failed` log line is the only signal today.

## Other notes

- **CLI wire contract intact.** No existing proc's input schema is touched. Widening a *gate* is invisible to released CLI binaries; changing an *input schema* is not. `app-collaborator.cli-contract.test.ts` parses each released-CLI payload and asserts all 13 `listingMediaCliScope` annotations remain.
- **Consent.** A pending or rejected invite confers **zero** capability and never appears in any public projection — otherwise anyone's name could be attached to a listing simply by inviting them.
- **Shared shadow.** Two editors converge on one revision draft per parent (partial UNIQUE on `revision_of_id`); the P2002 idempotent-reuse path is preserved. Editor B seeing editor A's staged edits is **deliberate** — a revision is a property of the listing, not of whoever opened it — and is commented as such.
- **Self-review** widened to the insider set (owner + accepted collaborators). `displayed` is deliberately **not** part of it: hiding your byline is a public-credit preference, not a capability, so it must not become a self-review bypass.
- **Throttle written correctly.** `EntityCollaborator`'s invite throttle is inverted (`>=` where its own sibling correctly uses `lte`), so it re-notifies precisely when it should stay silent. Ours is the right way round, with both sides of the boundary pinned so a copy-paste cannot re-import the bug.
- **Cap + rate limits.** 15 seats/app (mirrors `constants.entityCollaborators`), 30 invites/hr, 10 transfer attempts/hr.
- All new surfaces sit behind the existing mod-segmented App Blocks author flag, consistent with neighbouring procs.

---

## Verification

**Ran:**
- `pnpm typecheck` — **0 errors**.
- `pnpm vitest run --project unit` — **926 files, 14270 passed, 1 skipped, 0 failed** (counts read from the output, not an exit code; no timeout panics).
- Test files type-checked **explicitly** under a config that overrides `tsconfig.json`'s `src/**/__tests__/**` exclusion — a green `pnpm typecheck` says nothing about them. 0 errors in the 11 new suites.
- **Mutation sweep: 31 mutants, 31 killed.** Two classes — *delete the guard* (the operation must then succeed where it must not) and *change the guard's own code/message* (proving the test reads **this** guard's identity, not a neighbour's). Every mutant died to its own expected signal.

**Two harness bugs the sweep caught in itself, both worth stating** because either would have produced a confident wrong verdict: the verdict parser required a `passed` segment, but `vitest -t` prints `Tests 1 failed | 24 skipped` with no `passed` — so **every genuine kill initially read as SURVIVED**. And `-t` is a *regex*, so a filter containing parentheses matched nothing; that now reports `NO-TESTS-MATCHED` rather than being scored either way.

**Found by the sweep, in the code:**
- `assertAppEditAccess` had **no behavioural test** — deleting it left every suite green, because the only thing "covering" it was a structural ledger assertion that never executes the guard. It has since been relocated into `app-access.service` (one home, directly testable) and given a behavioural matrix.
- The `getMyAppsEarnings` app-scope filter **survived deletion**, because the final `allIds.map` is a second independent barrier that hides the leak in the *output* while the *query* still fetched the whole portfolio. A structural assertion on the WHERE clause now pins it.

**Not verified / needs a human:**
- **The migration has not been applied anywhere.** Nothing here was exercised against a real Postgres — no DB-level test of the CHECK constraints, the partial unique index, or the cascade behaviour. Every DB dep in these suites is a mock.
- **No live/e2e run.** There is no UI in this PR, and no click path was driven.
- **The transfer rollback is asserted structurally, not observed.** The `$transaction` fake runs its callback inline and cannot itself roll anything back; the test asserts what a real rollback guarantees and what the fake *can* observe — that the throw escapes `$transaction`, and that every ownership write is issued inside the callback rather than before it. A real two-column rollback should be confirmed against a live DB.
- **The Forgejo calls are asserted against a fake client**, not against a real Forgejo instance. `removeCollaborator`'s 204/404 handling is modelled on the documented Gitea/Forgejo behaviour and the existing `deleteForgejoUser`/`deleteReviewRepo` shape; it has not been measured against 15.0.6.

https://claude.ai/code/session_015w3tqf9DUTs8V7xGyDCcd3

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260810140000_app_listing_collaborators/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260810140000_app_listing_collaborators/migration.sql
@@ -1,0 +1,145 @@
+-- ============================================================
+-- App Listing COLLABORATORS — editor seats, ownership audit trail, transfers
+-- ============================================================
+-- Adds three ADDITIVE tables:
+--   1. app_collaborators        — the consent-gated editor seat (+ byline opt-in)
+--   2. app_ownership_events     — append-only audit trail of every seat/ownership action
+--   3. app_ownership_transfers  — an in-flight owner→recipient ownership transfer
+--
+-- Nothing existing is altered: no column is added to app_blocks, app_listings,
+-- oauth_clients or block_buzz_attributions, and no existing row is touched.
+--
+-- ⚠️ MANUAL APPLY — per datapacket-talos CLAUDE.md DB rule #8 the main civitai
+-- CNPG nvme0 DB does NOT auto-apply migrations (there is no `prisma migrate
+-- deploy` in any deploy path). This file is committed for HISTORY ONLY; a HUMAN
+-- applies the SQL below per environment (psql / retool). CI and deploy do NOT
+-- run it. Apply to BOTH:
+--   1. prod nvme0   (the live civitai DB)
+--   2. the dev clone (cnpg-cluster-dev, ns cnpg-database-dev, db civitai)
+--
+-- 🔴 THE FEATURE IS INERT UNTIL THIS IS APPLIED, BY CONSTRUCTION. Every read of
+-- these tables is wrapped in `safeCollaboratorQuery` (app-access.service.ts),
+-- which catches the Postgres "relation does not exist" error (Prisma P2021 /
+-- PG 42P01) and degrades to "no collaborators" — i.e. exactly today's
+-- owner-only behaviour. The WRITE paths (invite / transfer) surface a friendly
+-- error instead. So deploying the code before applying this migration changes
+-- nothing observable; it does not 500 the app-listing pages.
+--
+-- Idempotent: IF NOT EXISTS guards throughout, so a manual re-run is a no-op.
+-- All three tables are brand-new and EMPTY — no meaningful lock, no backfill.
+
+-- ------------------------------------------------------------
+-- 1. app_collaborators
+-- ------------------------------------------------------------
+-- Keyed to app_blocks (NOT app_listings): ownership is canonically
+-- oauth_clients.user_id reached via app_blocks.app_id, and app_listings.user_id
+-- is a denormalized copy that a shadow-revision approve can rewrite.
+--
+-- CASCADE on both FKs (unlike the audit table below): a seat is a live
+-- capability, not a historical record. If the app is deleted there is nothing
+-- to edit; if the user is deleted the seat must not survive as a dangling grant.
+CREATE TABLE IF NOT EXISTS "app_collaborators" (
+  "app_block_id"     TEXT    NOT NULL REFERENCES "app_blocks"("id") ON DELETE CASCADE,
+  "user_id"          INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
+  -- Capability role. 'editor' is the only value written today; TEXT (not an enum)
+  -- so adding a role needs no migration.
+  "role"             TEXT    NOT NULL DEFAULT 'editor',
+  -- 🔴 CONSENT: only 'accepted' confers ANY capability and only 'accepted' is ever
+  -- publicly visible. 'pending'/'rejected' are inert.
+  "status"           TEXT    NOT NULL DEFAULT 'pending',
+  -- Public-byline opt-in (immediate-apply, no mod review). Safe to apply
+  -- immediately BECAUSE it lives here: applyApprovedRevision's offsite branch
+  -- copies app_listings scalars from the shadow onto the parent, so the same flag
+  -- stored as an app_listings column would be clobbered by a later approve.
+  "displayed"        BOOLEAN NOT NULL DEFAULT true,
+  "invited_by"       INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
+  -- Re-invite notification throttle key. NULL = never notified.
+  "last_notified_at" TIMESTAMPTZ,
+  "created_at"       TIMESTAMPTZ NOT NULL DEFAULT now(),
+  "responded_at"     TIMESTAMPTZ,
+
+  PRIMARY KEY ("app_block_id", "user_id"),
+
+  CONSTRAINT "app_collaborators_status_check"
+    CHECK ("status" IN ('pending', 'accepted', 'rejected')),
+  CONSTRAINT "app_collaborators_role_check"
+    CHECK ("role" IN ('editor'))
+);
+
+-- "which apps may this user edit" — the hot path (resolveAccessibleAppBlockIds,
+-- run on every gated proc for a non-owner).
+CREATE INDEX IF NOT EXISTS "app_collaborators_user_status_idx"
+  ON "app_collaborators" ("user_id", "status");
+-- "who is on this app" — the owner roster + the public byline read.
+CREATE INDEX IF NOT EXISTS "app_collaborators_app_status_idx"
+  ON "app_collaborators" ("app_block_id", "status");
+-- FK index so a User delete (hot path) does not seq-scan on the inviter side.
+CREATE INDEX IF NOT EXISTS "app_collaborators_invited_by_idx"
+  ON "app_collaborators" ("invited_by");
+
+-- ------------------------------------------------------------
+-- 2. app_ownership_events  (append-only audit trail)
+-- ------------------------------------------------------------
+-- Mirrors app_listing_moderation_events' survivability posture: EVERY FK is
+-- NULLABLE + SET NULL so an event outlives the app, the actor and the target. The
+-- denormalized "slug" keeps the row self-describing once the app is gone.
+CREATE TABLE IF NOT EXISTS "app_ownership_events" (
+  "id"             TEXT PRIMARY KEY,                                        -- aoe_<ULID>
+  "app_block_id"   TEXT    REFERENCES "app_blocks"("id") ON DELETE SET NULL,
+  "slug"           TEXT    NOT NULL,
+  "action"         TEXT    NOT NULL,
+  "actor_user_id"  INTEGER REFERENCES "User"("id")       ON DELETE SET NULL,
+  "target_user_id" INTEGER REFERENCES "User"("id")       ON DELETE SET NULL,
+  "metadata"       JSONB,
+  "created_at"     TIMESTAMPTZ NOT NULL DEFAULT now(),
+
+  CONSTRAINT "app_ownership_events_action_check"
+    CHECK ("action" IN (
+      'invite', 'accept', 'reject', 'remove', 'leave', 'display',
+      'transfer_initiated', 'transfer_accepted', 'transfer_cancelled'
+    ))
+);
+
+CREATE INDEX IF NOT EXISTS "app_ownership_events_app_idx"
+  ON "app_ownership_events" ("app_block_id", "created_at" DESC);
+CREATE INDEX IF NOT EXISTS "app_ownership_events_actor_idx"
+  ON "app_ownership_events" ("actor_user_id", "created_at" DESC);
+
+-- ------------------------------------------------------------
+-- 3. app_ownership_transfers
+-- ------------------------------------------------------------
+-- One in-flight transfer per app, enforced by the PARTIAL UNIQUE index below.
+CREATE TABLE IF NOT EXISTS "app_ownership_transfers" (
+  "id"           TEXT PRIMARY KEY,                                    -- aot_<ULID>
+  "app_block_id" TEXT    NOT NULL REFERENCES "app_blocks"("id") ON DELETE CASCADE,
+  -- Snapshot of the owner at initiate time. Re-asserted in-tx at accept, so a
+  -- transfer initiated by someone who has since lost the app cannot complete.
+  "from_user_id" INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
+  "to_user_id"   INTEGER NOT NULL REFERENCES "User"("id")       ON DELETE CASCADE,
+  "status"       TEXT    NOT NULL DEFAULT 'pending',
+  -- Hard expiry, enforced as a read-time predicate in the ACCEPT path — no
+  -- sweeper job is required for CORRECTNESS (a sweeper would only tidy rows).
+  "expires_at"   TIMESTAMPTZ NOT NULL,
+  "created_at"   TIMESTAMPTZ NOT NULL DEFAULT now(),
+  "responded_at" TIMESTAMPTZ,
+
+  CONSTRAINT "app_ownership_transfers_status_check"
+    CHECK ("status" IN ('pending', 'accepted', 'rejected', 'cancelled', 'expired')),
+  -- A self-transfer is meaningless and would make the accept path a no-op that
+  -- still emits an audit event.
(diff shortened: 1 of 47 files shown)


ZacxDev:
## Correction to the PR description: "every read of the new tables goes through `safeCollaboratorQuery`" is wrong

Posting this publicly rather than quietly editing the body, because it is a **pre-merge safety claim** and a reviewer may already have read it and taken it at face value.

### What the body says

> **The feature is INERT until then, by construction.** Every read of the new tables goes through `safeCollaboratorQuery` (`app-access.service.ts`) …

### What is actually true

**Three READ procs are NOT wrapped** and will throw `P2021` against a database where the manual migration has not been applied:

| Proc | Site |
|---|---|
| `appCollaborators.list` → `listCollaborators` | `app-collaborator.service.ts` — `dbRead.appCollaborator.findMany` |
| `appCollaborators.listMyPendingInvites` | `app-collaborator.service.ts` — `dbRead.appCollaborator.findMany` |
| `appCollaborators.getPendingTransfer` | `app-ownership-transfer.service.ts` — `dbRead.appOwnershipTransfer.findFirst` |

(Separately, four reads on the WRITE paths — `inviteCollaborator`'s existing-row lookup and cap `count`, and `acceptTransfer`/`cancelTransfer`'s in-transaction `appOwnershipTransfer.findUnique` — are also unwrapped. Those are unwrapped **deliberately**: the module header already states that write paths must not swallow the missing-table error, because an invite that silently did nothing would be worse than an error. They are listed here only so the population is closed, not as a defect.)

### Why the conclusion still holds, stated at the scope it was actually measured

The **inertness claim survives**; only the sentence justifying it is too strong. The narrower, true version:

> Every read on a path that is reachable **today** goes through `safeCollaboratorQuery`. The three unwrapped reads live on procs in the new `appCollaborators` router, and **nothing calls them** — no UI ships in this PR, and `grep` over `src/` finds no client caller of `trpc.appCollaborators.*`.

Two caveats a reviewer should weigh, which "by construction" papers over:

1. **"Nothing calls them" is a fact about callers, not about reachability.** The router IS registered in `src/server/routers/index.ts`, so all three are wire-reachable by any account that passes the App Blocks author flag. Anyone hitting them directly between deploy and migration gets a 500, not a degraded empty list. That is a much smaller blast radius than "every app page 500s" — which the `safeCollaboratorQuery`-wrapped paths genuinely do prevent — but it is not zero, and it is not "by construction".
2. **The guarantee therefore depends on a UI never shipping before the SQL is applied**, which is a sequencing commitment rather than a structural property. If the first consumer lands before the migration, these are the three call sites to wrap.

### On the test-file citation

The task that produced this comment also flagged that the body repeats a citation to a test file that does not exist. **I checked, and that is not true of the body**: every `*.ts` / `*.sql` path referenced in the description resolves to a real file in the branch (`app-collaborator-earnings.service.test.ts`, `app-collaborator.cli-contract.test.ts`, `app-collaborator.revision-non-clobber.test.ts`, `dev-token.ts`, the migration). Rather than "correct" something that is already right, I am recording the check.

The dead citations were in **source comments**, not the description — `app-listing.public-collaborators.test.ts` (cited twice; the real files are `app-collaborator.public-projection.test.ts` for the DTO allowlist and `app-access.service.test.ts` for the accepted-AND-displayed filter) and `app-ownership-transfer.money-invariance.test.ts` (the real assertions are the `MONEY INVARIANCE` describe in `app-ownership-transfer.service.test.ts`). Those are corrected in the follow-up commit.

### Status

This comment describes the branch **as it currently stands**. A follow-up commit addressing the audit's authorization findings (including adding an authorization check to `getPendingTransfer`, which today takes no caller at all, and requiring a real role in `listCollaborators`) is prepared but is **not yet pushed** — do not read this comment as a claim that those are already on the branch.

PR #1885, opened 2025-10-20
Improves comments loading time by adding pagination

manuelurenah:

PR #3812, opened 2026-08-11
fix(comments): show every article thread at full depth again

manuelurenah:
Fixes [CU-868kp9wzg](https://app.clickup.com/t/8459928/868kp9wzg) — *"Article comments: restore multi-thread + full deep-thread view"*.

> **Preview:** https://pr-3812.civitaic.com/articles/20211/civitai-green-gets-an-upgrade-important-changes-to-blue-buzz (1.7k comments — exercises both the threading and the perf side; needs a staging sign-in)

---

## ⚠️ One decision needs a call before this merges

**Image and bountyEntry comment threads get shallower: an effective cap of 5 today, 3 after this PR.**

`constants.comments.getMaxDepth` has always declared `image: 3` and `bountyEntry: 3`, but it never applied below the first level (root cause #2 below), so those surfaces have effectively been running at the `default: 5` the whole time. Fixing the resolution makes the declared value real for the first time.

Arguments for letting it land at 3:
- It is what the constant has said since it was written, and 3 was a deliberate choice for a narrow image-detail column.
- It matches the reporter's framing — *"Article comments have room for full depth, **unlike image comments**"* — which assumes image threads stay shallow.
- Nothing is lost: at the cap, a thread opens in its own view rather than being truncated.

Argument against: it is a **user-visible change on surfaces this ticket never asked about**, and nobody has complained about image threads.

I did not want to make that call silently in either direction. If you'd rather images keep today's behaviour, it is a one-line change — add `case 'image': case 'bountyEntry': return 5;` or just let them fall through to the default. Everything else in this PR is unaffected either way.

---

## What was reported

> Article comment threads only show one thread at a time and cut off deep threads (used to show multiple threads at once). Article comments have room for full depth, unlike image comments.

## Root cause

Two separate things, both confirmed against the code and in the browser.

**1. Nothing opens the threads any more.** Before #1885, `RootThreadProvider` ran one `getThreadDetails` query that returned the whole reply tree, primed every child thread's `getThreadDetails`/`getCount` cache from it, and called `setExpanded(...)` with every comment id down to the depth cap. So the page arrived with all threads open, for one query. #1885 replaced that tree query with cursor-paged `getInfinite` and removed the priming block — including the `setExpanded` call. Nothing took over, `setExpanded` became dead code, and every reply tree has sat behind a "Show N More" button since.

**2. `constants.comments.getMaxDepth` never applied below the first level.** It is called with the entity type from `useCommentsContext()`, and every nested thread mounts a provider with `entityType="comment"`. So the per-surface value resolved only for level‑1 comments and everything under them fell through to the `default: 5`. Articles and images have been capping at the same depth the whole time — the declared `image: 3` never took effect either. (This is what the decision above is about.)

## The fix

- `getInfinite` optionally returns the reply threads nested under the page it just fetched (`repliesDepth`/`repliesLimit`): one recursive CTE over `Thread` plus one `CommentV2` fetch. `EXPLAIN ANALYZE` on the worst article in prod (2,019 threads) — **2.65 ms**, index scans throughout.
- `CommentsProvider` seeds each nested thread's `getInfinite`/`getThreadDetails`/`getCount` caches from that batch, then opens them.
- `RootThreadContext` now carries `rootEntityType`, and depth resolves against that.
- Article depth goes to **10**. Prod distribution of article thread depths: ≤5 covers 95.5%, **≤10 covers 99.5%**, ≤15 covers 99.8%, max observed 73.
- Auto-opening is opt-in per surface (`expandsRepliesByDefault`), currently articles only.
- **The batch is capped at 40 replies** (`autoExpandBudget`), shallowest-first. Without it a busy article rendered 313 comments off one page and blocked the main thread for 3.36 s — the same failure pagination was added to fix. Details + numbers: [perf follow-up](https://github.com/civitai/civitai/pull/3812#issuecomment-5256443298).
- **A thread too long to sit inline opens in its own view** ("View all N replies") instead of paging in place behind a button that read the same as the article's own pagination. Details: [thread-view follow-up](https://github.com/civitai/civitai/pull/3812#issuecomment-5256918791).

## Verified in the browser

Small article (18162, 12 levels deep), this branch vs. a `main`-based worktree:

| | main | this branch |
|---|---|---|
| Comments rendered | 9 | **21** |
| Distinct indent levels | 1 (all root) | **10** |
| Collapsed "Show N More" | 4 | 1 (the level-10 comment) |
| `commentv2` requests | 11 | 12 |

Worst public article (20211, 1,755 root comments), after the cap:

| | main | this branch |
|---|---|---|
| Comments rendered | 20 | 57 |
| DOM nodes | 3,276 | 6,238 |
| Total blocking | 220 ms | 774 ms |

Cost is capped by a constant rather than article size — an 866-comment article and a 1,755-comment article both land at ~58 comments / ~775 ms. **97.2% of articles (10,860 of 11,177) still open every thread on the first page.** With auto-expand off, the request is byte-identical to main (0.40 s / 314,501 B vs 0.43 s / 314,458 B), so non-article surfaces are untouched apart from the depth decision above.

## Checks

- `pnpm run test:unit:run` — 917 files, 14,113 passed, 0 failed (incl. 18 new tests; `blocks.router.workflow.test.ts` collected 347, so the worktree run is valid)
- `pnpm run prettier:write` — applied
- `eslint` on the touched files — no new findings

### DIFF
--- a/src/components/Comics/ChapterComments.tsx
+++ b/src/components/Comics/ChapterComments.tsx
@@ -75,6 +75,7 @@ export function ChapterComments({
         expanded,
         toggleExpanded,
         activeComment: undefined,
+        rootEntityType: 'comicChapter',
       }}
     >
       <CommentsCtx.Provider
--- a/src/components/Comics/ChapterComments.tsx
+++ b/src/components/Comics/ChapterComments.tsx
@@ -29,8 +29,7 @@ export function ChapterComments({
   const router = useRouter();
   const highlighted = parseNumericString(router.query.highlight);
   const [sort, setSort] = useState<ThreadSort>(ThreadSort.Oldest);
-  const expanded = useNewCommentStore((state) => state.expandedComments);
-  const toggleExpanded = useNewCommentStore((state) => state.toggleExpanded);
+  const setExpanded = useNewCommentStore((state) => state.setExpanded);
   const utils = trpc.useUtils();
 
   const { data: thread, isLoading } = trpc.comics.getChapterThread.useQuery(
@@ -72,8 +71,7 @@ export function ChapterComments({
         isInitialThread: true,
         setInitialThread,
         setRootThread,
-        expanded,
-        toggleExpanded,
+        setExpanded,
         activeComment: undefined,
         rootEntityType: 'comicChapter',
       }}
--- a/src/components/CommentsV2/Comment/Comment.tsx
+++ b/src/components/CommentsV2/Comment/Comment.tsx
@@ -34,7 +34,9 @@ import { CommentBadge } from '~/components/CommentsV2/Comment/CommentBadge';
 import {
   CommentsProvider,
   useCommentsContext,
+  useNewCommentStore,
   useRootThreadContext,
+  useSeededReplyThreads,
 } from '~/components/CommentsV2/CommentsProvider';
 import { DaysFromNow } from '~/components/Dates/DaysFromNow';
 import { openReportModal } from '~/components/Dialog/triggers/report';
@@ -93,14 +95,19 @@ export function CommentContent({
   borderless,
   ...groupProps
 }: CommentProps) {
-  const { expanded, toggleExpanded, setRootThread, rootEntityType } = useRootThreadContext();
+  const { setExpanded, setRootThread, rootEntityType } = useRootThreadContext();
   const { entityId, entityType, highlighted, level } = useCommentsContext();
   const { canDelete, canEdit, canReply, canHide, canPin, badge, canReport } = useCommentV2Context();
 
-  const { data: replyCount = 0 } = trpc.commentv2.getCount.useQuery({
-    entityId: comment.id,
-    entityType: 'comment',
-  });
+  const seededThreads = useSeededReplyThreads();
+  const seededThread = seededThreads.byCommentId.get(comment.id);
+  const seededCount =
+    seededThread?.commentCount ?? (seededThreads.childless.has(comment.id) ? 0 : undefined);
+
+  const { data: replyCount = 0 } = trpc.commentv2.getCount.useQuery(
+    { entityId: comment.id, entityType: 'comment' },
+    { initialData: seededCount }
+  );
 
   const id = useStore((state) => state.id);
   const setId = useStore((state) => state.setId);
@@ -146,13 +153,19 @@ export function CommentContent({
     };
   }, [isHighlighted, comment.id]);
 
-  const isExpanded = !viewOnly && expanded.includes(comment.id);
   const maxDepth = constants.comments.getMaxDepth({ entityType: rootEntityType ?? entityType });
+  // Subscribe per comment rather than to the whole set: opening one thread must not re-render
+  // every other comment in the conversation.
+  const expandOverride = useNewCommentStore((state) => state.expandOverrides[comment.id]);
+  // A thread the page fetched up front is one the reader is meant to see open. Deriving it means
+  // re-rooting the section can't carry this section's opened threads back out with it.
+  const expandsByDefault = !!seededThread && seededThread.depth < maxDepth;
+  const isExpanded = !viewOnly && (expandOverride ?? expandsByDefault);
   // Too deep to indent any further, so this one opens as its own thread instead of in place.
   const opensOwnThread = (level ?? 0) >= maxDepth && !isExpanded;
   const onToggleReplies = () => {
     if (opensOwnThread) setRootThread('comment', comment.id);
-    else toggleExpanded(comment.id);
+    else setExpanded(comment.id, !isExpanded);
   };
 
   for (const end of trimmableEnds) {
--- a/src/components/CommentsV2/Comment/Comment.tsx
+++ b/src/components/CommentsV2/Comment/Comment.tsx
@@ -13,6 +13,7 @@ import {
 } from '@mantine/core';
 import {
   IconArrowBackUp,
+  IconArrowRight,
   IconCaretDownFilled,
   IconDotsVertical,
   IconEdit,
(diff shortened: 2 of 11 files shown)


manuelurenah:
## Perf follow-up: the first cut did reintroduce the freeze — fixed and measured

@manuelurenah flagged that #1885 was motivated by 1k+ comment articles hanging the page until everything loaded. That was a real hit on the first version of this PR.

**Measured on article 20211 (1,755 root comments), the worst unlocked public article:**

| | main | this PR (first cut) |
|---|---|---|
| Comments rendered | 20 | **313** |
| DOM nodes | 3,276 | **23,854** |
| `getInfinite` payload | 307 KB | **1,038 KB** |
| Longest single task | 270 ms | **3,358 ms** |
| Total blocking | 220 ms | **3,491 ms** |

A 3.36 s main-thread task is the freeze, at smaller scale. The problem: "open every thread under this page" is unbounded — it scales with how busy the article is.

### Fix: cap the batch

The server now selects reply threads **shallowest-first until 40 comments are accounted for**, then stops. Selection is a prefix in `(depth, id)` order, so a returned thread's whole ancestor chain is always returned too — nothing renders under a comment that isn't there. Threads past the cap keep exactly the "Show N More" button they have on `main`.

| | main | this PR (capped) |
|---|---|---|
| Comments rendered | 20 | 57 |
| DOM nodes | 3,276 | 6,238 |
| `getInfinite` payload | 307 KB | 492 KB |
| Longest single task | 270 ms | 824 ms |
| Total blocking | 220 ms | 774 ms |

**The important property is that the cost is now a constant, not a function of article size.** Article 28369 (866 root comments) and article 20211 (1,755 root comments) both land at ~58 comments / ~6,250 DOM nodes / ~775 ms. Doubling the article changes nothing.

### The cap almost never binds

Across every article with replies on its first page (11,177 of them), counting what would actually render:

| p50 | p90 | p99 | max |
|---|---|---|---|
| 3 replies | 19 | 57 | 477 |

**10,860 of 11,177 articles (97.2%) still open every thread on the first page.** Only 317 hit the cap, and those are precisely the ones that would otherwise freeze.

### Also checked

- **The disabled path is byte-identical to main.** Same request without `repliesDepth`: main 0.43 s / 314,458 B, this branch 0.40 s / 314,501 B. Every non-article surface is untouched.
- **Extra server cost is 2 round trips, not query cost.** The recursive CTE is 2.65 ms (`EXPLAIN ANALYZE`, worst article, index scans throughout); the per-thread over-fetch is 326 rows fetched vs 307 rendered, so it wasn't worth bounding. The wall-clock delta in dev is dominated by ~89 ms/round-trip through the SSH tunnel to the replica, which prod doesn't pay.
- Small articles are unchanged by the cap — article 18162 still opens all 10 levels, 21 comments, 12 `commentv2` requests.

### One behavioural consequence worth knowing

Because a capped batch means "no thread returned" no longer implies "no replies", the server now states explicitly which comments it confirmed are childless, and only for the levels the selection finished. Guessing past those would hide a real "show replies" button. That rule is covered by tests.

`constants.comments.autoExpandBudget = 40` is the single knob if you want it tuned either way.

manuelurenah:
## Long threads now open instead of paging in place

Took @manuelurenah's suggestion. The screenshot made the problem obvious: the nested **"Load More Comments"** is the same label and the same centred styling as the article's own pagination, except it sits *inside* an indented block and means something else. Clicking it grew that block with no end in sight.

**A thread too long to sit inline now opens as its own view**, exactly the way an over-deep one already did: **"View all N replies"** re-roots the section onto that comment, with "Show full conversation" to come back. On article 20211 that turns 6 ambiguous nested "Load More Comments" into 5 "View all N replies" plus **one** "Load More Comments" — the article's own, at the bottom, where it means what it says.

### Where I did *not* follow the suggestion

**"Show N More" keeps expanding in place.** Those are the 29 buttons, so they look like the bigger wall — but they only exist on the 2.8% of articles that hit the cap, and making them navigate would undo "show multiple threads at once" on exactly the busy articles where reading in place matters most. The rule is now: **expands in place → "Show N More" (caret); opens its own view → "View N replies" (arrow).** One affordance per behaviour.

No threshold needed — 96% of reply threads (107,968 of 112,432) fit in one page and show no button at all, so "doesn't fit → open it" never fires for trivial cases.

### Two things this needed to actually work

**1. The reader was being dropped mid-page.** Re-rooting replaces the whole section, so the page collapses under them. It now scrolls to the thread once it has rendered. Two things I got wrong on the way, both worth knowing:

- `window.scrollTo` is a **no-op in this app** — it scrolls an inner `div.scroll-area`, not the window (`document.scrollingElement` has zero scrollable height). `window.scrollY` therefore always reads 0, which also makes it a misleading thing to assert on. `scrollIntoView` is container-agnostic and works.
- Firing on a timer didn't work either: surfaces render a loader in the thread's place, so the target isn't in the DOM yet. It now triggers when loading actually completes — child effects run before the parent's, so the element is committed by then.

It aims at the **return link**, not the comment: landing flush on the comment scrolls "Show full conversation" off the top, and how far above it sits differs per surface, so aiming at it beats guessing a pixel offset.

**2. Thread-view expansion was following the reader back out.** Expansion state is global and sticky, so everything the thread view auto-opened stayed open in the article view — the full conversation came back at **102 comments instead of 57, blocking 1.33s**. A section now undoes its own auto-expansion when it re-roots, keyed on the entity so paging doesn't collapse what earlier pages opened. Measured across three different threads: back to **exactly 57 every time**.

### Checked

- Article 18162 (small) — unchanged, all 10 levels open, one "View 1 reply" at the depth cap.
- **Image pages** — this changes nested pagination on every CommentsV2 surface, not just articles. Verified: still collapsed by default, still caps at 3 levels, then offers "View 2 replies". Correctly pluralised.
- Full round trip: open → thread renders with return link on screen at y=75 → back to the full conversation.
- `pnpm run typecheck` clean, `test:unit:run` 917 files / 14,113 passed, eslint on touched files shows no new findings.

manuelurenah:
## 🙋 Needs a call: image/bountyEntry thread depth

Pulling this up out of the description, since it's the one thing in here that changes behaviour on a surface the ticket never mentioned, and I don't want it to land by default just because it was buried under the perf and UX follow-ups.

**Image and bountyEntry comment threads get shallower — effectively 5 today, 3 after this PR.**

`constants.comments.getMaxDepth` has declared `image: 3` / `bountyEntry: 3` since it was written, but it was called with the entity type from the *nested* provider, and every reply thread mounts as `entityType="comment"`. So the per-surface value only ever resolved for level-1 comments and everything below fell through to `default: 5`. Fixing that resolution is what makes article depth work at all — and as a side effect it makes the declared `3` real on those two surfaces for the first time.

**For letting it land at 3:** it's what the constant has always said, 3 was a deliberate choice for the narrow image-detail column, and it matches the reporter's own framing — *"Article comments have room for full depth, **unlike image comments**"*. Nothing is truncated either way: at the cap a thread opens in its own view.

**Against:** it's user-visible on surfaces nobody complained about, and it wasn't part of the ask.

**If you want images to keep today's behaviour** it's one line in `getMaxDepth` — return 5 for those cases, or drop them and let both fall through to the default. Nothing else in the PR depends on it. Happy to push that change if you'd rather ship the article fix on its own and decide about images separately.

@manuelurenah — flagging for a second opinion per your ask.

PR #3813, opened 2026-08-11
fix(app-listings): bound the ranged read too, and pin what #3804 left unpinned

ZacxDev:
Follow-ups to #3804 — hardening and internal consistency, no product behaviour change for a healthy store. Based directly on `main` (which already contains #3804); not stacked on anything.

## 1. The ranged read is bounded too (the substantive one)

#3804 bounded the entity-tag `HeadObject` and argued, in a comment, that a bound is not optional for a store request sitting inline on a user-facing mutation. That argument applies at least as strongly to the *other* request on the same mutation — the ranged read that measures an upload — which was left unbounded and, unlike the head, actually transfers bytes. It therefore has two ways to stall rather than one: never responding, or responding and then not delivering the body. Either held the request open indefinitely.

**Deliberately not the head's budget, and the comment says why** rather than leaving a bare number:

* 5s is a *round-trip* allowance for a request that moves nothing. This one moves up to the caller's whole byte cap, so reusing 5s would abort legitimate large uploads on a slow link and surface them to the author as "we couldn't read that back" — a hardening change that invents a new rejection for images that are fine is not a hardening change.
* The value is taken from the **sibling bounded transfer under the same byte cap** in the block-media path rather than picked afresh, so the two byte-moving bounds in this area cannot drift apart.
* What it buys is not a latency target — it is the difference between *bounded* and *unbounded*. An abort lands on the existing "could not consult the store" state, so the classification the caller acts on is unchanged.

## 2. The injected timeout parameter is pinned

Both entry points take a `timeoutMs` so the abort path is reachable in milliseconds. A version that accepted the parameter and then applied the module constant behaved *identically* in production and passed every existing case — those cases would simply have taken seconds instead of milliseconds, which nothing asserted. The value handed to the bound is now asserted directly, via a spy rather than a stopwatch so the guard carries no race.

## 3. The fail-open `.catch` on the integrity event is pinned

The event is fire-and-forget by design — an operational log must not sit on the critical path of the attach it describes — which is exactly what makes a rejecting sink dangerous: the rejection has nowhere to go but `unhandledRejection`, so the observation takes down the request it was only watching. Deleting the clause previously left the whole suite green.

Its test needs a **file of its own** to be falsifiable at all: a `vi.fn()` sink attaches its own handler to the returned promise and marks the rejection handled, so an assertion written against one passes with or without the production `.catch()`. Same split, and the same reason, as the existing cover-image logging test.

## 4. Severity lowered to `warning`, mismatch included

`type` is what the log pipeline reads as the level, and `src/server/logging/client.ts` states the rule plainly: a BAD_REQUEST-class fault is normal user feedback and "must never be logged at error severity, or they drown out the real server-side failures". A mismatch ends in exactly such a rejection *and* any author can produce one at will, so at `error` this event was an author-controlled tap on the error board.

The severity was never the discriminator anyway — `status` and `reason` are emitted as their own fields precisely so these outcomes are separable, and a query or alert on `status="mismatch"` is both narrower and more stable than one on a level shared with every other error in the app. **Called out explicitly for review**: the alternative was to keep `error` and document a security-event exception. I chose the file's own rule, because the trigger is author-controlled and the discrimination is already available in a better field.

## 5. Comment correction

A comment claimed four `unverifiable` reasons are reachable from that consumer where three are — the fourth is short-circuited a few lines above. The block reconciled itself further down, but the headline sentence was wrong read alone.

---

## Test matrix — honestly

9 new cases. **5 are RED at `a7a96e7cd3` and green here**; 4 are green at base by construction.

| Case | At base | Kind |
|---|---|---|
| ranged read is handed a live abort signal | ✗ red | new-code coverage |
| ranged read fails `store-unavailable` when the store never responds | ✗ red (times out) | new-code coverage |
| ranged read fails `store-unavailable` when the body never arrives | ✗ red (times out) | new-code coverage |
| injected budget is the one applied — read side | ✗ red | new-code coverage |
| every outcome emits below error severity | ✗ red (was `error`) | new-code coverage |
| injected budget is the one applied — head side | ✓ green | **invariant guard** on #3804 behaviour |
| degraded sink: unverifiable verdict still attaches | ✓ green | **invariant guard** |
| degraded sink: mismatch still refuses with its own message | ✓ green | **invariant guard** |
| positive control: the re-check actually reached the store | ✓ green | positive control |

Base run: **5 failed / 56 passed of 61**. HEAD run: **61 passed**. Wider sweep over the whole blocks + server-utils tree at HEAD: **172 files / 3426 tests, all green**.

None of the four green-at-base cases is regression coverage and none is counted as such — they pin behaviour #3804 already shipped, and their worth is the mutation result below rather than the base run.

## Mutation results — every guard, beyond deletion

Each mutation was applied to the real source, the suite re-run, and the *failing message* read to confirm it attributes to the guard under test rather than an adjacent one.

| Mutation | Killed by | Failing assertion |
|---|---|---|
| bound deleted from the ranged read | 4 cases | `expected [] to deeply equal [ 37, 60000 ]`, `expected undefined to be an instance of AbortSignal`, 2× timeout |
| bound built from the **wrong constant** (head's) | budget pin **only** | `expected [ 5000, 5000 ] to deeply equal [ 37, 60000 ]` |
| read accepts `timeoutMs` then ignores it | budget pin + both hang cases | `expected [ 60000, 60000 ] to deeply equal [ 37, 60000 ]` |
| head accepts `timeoutMs` then ignores it | budget pin **only** | `expected [ 5000, 5000 ] to deeply equal [ 37, 5000 ]` |
| budget set to a degenerate `0` | bound case | `expected 0 to be greater than 0` |
| severity restored to `error` on mismatch | severity case | `to match object { status: 'mismatch', type: 'warning' }` |
| `.catch(() => null)` deleted | both sink cases | `expected [ Error: axiom is unreachable ] to deeply equal []` |
| whole re-check made unreachable (positive control) | 15 cases incl. the new positive control | `expected [] to have a length of 1 but got +0` |

Two of these matter more than the rest. The **wrong-constant** mutation is killed *only* by the injected-budget pin — the hang cases pass at 5s, so without that pin a silently-wrong-but-finite bound ships. And the **unreachable** mutation kills the new positive control, which is what stops the two sink cases passing for the wrong reason: with the check skipped nothing logs, so "no unhandled rejection" would be vacuously true.

## Deliberately not done

* **No Prisma migration**, and none is needed — nothing here touches a column or a schema.
* **No new event fields, no renames.** The event's key set is unchanged (a pre-existing test pins it), so existing queries keep working; only the level moved.
* **No change to the head's 5s budget.** It is correct for a request that transfers nothing, and #3804's reasoning for it still holds.
* **No test for item 5's comment correction** — it is a comment, and a test asserting comment text would pin the wrong thing.
* **The "body never arrives" case is a fidelity choice, stated as one.** The fixture wires the abort into the body read because that is what the SDK does — it destroys the underlying request, so a stream still being read off that socket errors rather than idling. That behaviour is the SDK's, not something this PR can assert against a real backend from a unit test.

## Pre-existing, not attributable here

* Typecheck is red at base and at HEAD with the **identical 12-error set** (verified by diffing both runs: `event-engine-common` module resolution + implicit-`any` in `image.service.ts`). Nothing in a touched file. Note this is 12, not the ~34 quoted to me — reporting what I measured.
* The first `preview / deploy` run failed, and it is worth saying exactly how rather than waving at "the pool is busy": the image step was **OOM-killed**, which is a different signature from the scheduling failures the pool had been showing, so it got a control rather than an assumption. The failed run took 28 minutes where comparable ones took 9–17, overlapping a concurrent memory experiment that had been holding an unschedulable claim against the same pool for over an hour; a build of the base branch on the same pool succeeded two minutes after this branch's failed. That was suggestive but not decisive either way, so the branch was rebuilt on an **identical tree** (the empty commit here) — and it passed in 18 minutes. Same code, same pool, opposite outcome: the OOM was environmental.

  Worth noting for anyone reading the failed run later: nothing in this branch adds a module to the production graph. The three test files are not reachable from any entry point, and the two source files gain an exported number, an optional parameter and one argument — no new imports. So there was no mechanism by which it could move build memory, but "no mechanism I can see" is not a measurement, which is why the re-run happened.

Refs #3804

### DIFF
--- a/src/server/services/blocks/__tests__/listing-asset-integrity-logging.test.ts
+++ b/src/server/services/blocks/__tests__/listing-asset-integrity-logging.test.ts
@@ -1,0 +1,210 @@
+import { HeadObjectCommand } from '@aws-sdk/client-s3';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type * as LoggingClient from '~/server/logging/client';
+import type * as S3Utils from '~/utils/s3-utils';
+
+/**
+ * Pins the `.catch()` on the best-effort `logToAxiom(...)` in the listing-media
+ * integrity re-check.
+ *
+ * The call is fire-and-forget: nothing awaits it, because an operational event must
+ * never be on the critical path of the attach it is describing. That is exactly what
+ * makes a rejecting sink dangerous — `logToAxiom` awaits an ingest that rejects when
+ * Axiom is degraded, and an unawaited rejection with no handler goes to
+ * `unhandledRejection`, i.e. the observation takes down the request it was only
+ * watching. The `.catch()` is the whole defence, and deleting it changes nothing
+ * anywhere else in the suite.
+ *
+ * ## Why this is a separate file
+ *
+ * 🔴 The sink below is a PLAIN FUNCTION, not a `vi.fn()`, and that is the entire
+ * reason this test can fail. A `vi.fn()` records settled results by attaching
+ * `returnValue.then(onFulfilled, onRejected)` to whatever the implementation
+ * returns, which MARKS a returned rejected promise as handled — so through a
+ * `vi.fn()` mock Node never emits `unhandledRejection`, with or without the
+ * production `.catch()`, and the assertion passes either way. Unfalsifiable, which
+ * is worse than no test. `vi.mock` is per-file and the sibling
+ * `listing-asset-upload-integrity.test.ts` deliberately uses `vi.fn()` there so it
+ * can assert on payloads; the two shapes cannot coexist in one file. (Same split,
+ * same reason, as `~/server/services/__tests__/cover-image.service.logging.test.ts`.)
+ */
+
+const { mockRead, mockWrite, mockS3Send } = vi.hoisted(() => {
+  const db = {
+    appListing: {
+      findUnique: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
+      findFirst: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
+      update: vi.fn(async (..._a: unknown[]) => ({})),
+    },
+    appListingScreenshot: {
+      findMany: vi.fn(async (..._a: unknown[]): Promise<unknown[]> => []),
+      count: vi.fn(async (..._a: unknown[]) => 0),
+      create: vi.fn(async (..._a: unknown[]) => ({})),
+    },
+    image: {
+      findUnique: vi.fn(async (..._a: unknown[]): Promise<unknown> => null),
+      findMany: vi.fn(async (..._a: unknown[]): Promise<unknown[]> => []),
+    },
+    $transaction: vi.fn(async (arg: unknown) => {
+      if (typeof arg === 'function') return (arg as (tx: unknown) => unknown)(db);
+      return Promise.all(arg as Promise<unknown>[]);
+    }),
+  };
+  return { mockRead: db, mockWrite: db, mockS3Send: vi.fn() };
+});
+
+vi.mock('~/server/db/client', () => ({ dbRead: mockRead, dbWrite: mockWrite }));
+vi.mock('~/server/services/image.service', () => ({ createImage: vi.fn() }));
+vi.mock('~/server/utils/app-block-ids', () => ({
+  newAppListingId: () => 'apl_test_1',
+  newAppListingPublishRequestId: () => 'alpr_test_1',
+  newAppListingModerationEventId: () => 'alme_test_1',
+  newAppListingScreenshotId: () => 'apls_test_1',
+}));
+vi.mock('~/utils/s3-utils', async (importOriginal) => ({
+  ...(await importOriginal<typeof S3Utils>()),
+  getImageUploadBackend: async () => ({
+    s3: { send: mockS3Send },
+    bucket: 'test-image-bucket',
+    backend: 'backblaze' as const,
+  }),
+}));
+
+/**
+ * A degraded sink. Everything else in the logging module stays REAL, so this test
+ * is not also asserting a hand-built module shape.
+ */
+vi.mock('~/server/logging/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof LoggingClient>()),
+  logToAxiom: () => Promise.reject(new Error('axiom is unreachable')),
+}));
+
+const OWNER = 42;
+const KEY = '44444444-4444-4444-8444-444444444444';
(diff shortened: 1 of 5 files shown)


ZacxDev:
Two corrections to this PR's body, from an adversarial review. Neither changes the code; both change claims made about it.

### 1. The severity drop does remove alert coverage — "nothing is lost except the flooding" is wrong

Measured on the live log stack rather than reasoned: lines carrying the previous level resolve to `detected_level=error` **100%** of the time (n=1197 over 1h), and exactly one alert plus one dashboard panel key on that filter. At `warning` this event falls out of both, and **no replacement signal keyed on the mismatch status is added by this PR**. So a mismatch storm — abuse, or the guard misfiring wholesale if the object store ever changes its tag formatting — would now be unalerted.

The direction is still right: the fault is client-class and author-triggerable, and the attach procs carry no rate limit, so keeping it at the higher level really would have been an author-controlled tap on the error board. And nothing is being lost *today* — this event has been emitted **zero** times in the last 24h. But the honest statement is "alert coverage is being dropped and not yet replaced", not "nothing is lost".

One related detail the body slightly overstates: the status and reason fields remain queryable, but they are not promoted to structured metadata, so a replacement query needs a JSON parse stage rather than a label match.

### 2. `preview / deploy` failed as a build-step OOM, not a scheduling/capacity timeout

The body attributes it to the build pool being capacity-constrained with pods unable to schedule. That was the failure mode on the predecessor PR, not this one: this run's `build-image` ended in `StepOOM`. It still looks environmental — the same failure hit an unrelated PR today and there is active memory-pressure work on that pool — but it is a different mechanism, and it should be re-run and seen green rather than waved through on a mis-stated excuse. That mis-attribution came from the briefing I gave, not from the author.

ZacxDev:
Retracting point 2 of my previous comment — it was wrong, and unfair to the change.

I wrote that the body attributed the `preview / deploy` failure to a scheduling/capacity timeout. It does not, and by the time I posted it already did not. The body identifies the image step as **OOM-killed**, notes explicitly that this is a *different signature* from the scheduling failures the pool had been showing, declines to accept the environmental story on that basis, and then settles it with a control: the same tree rebuilt via an empty commit, which passed in 18 minutes where the failed run took 28. Same code, same pool, opposite outcome.

That is a better piece of reasoning than the one I was pushing — I had supplied the `FailedScheduling` framing from the predecessor PR, and it was correctly rejected rather than adopted. Point 1 of that comment (the severity drop dropping alert coverage with no replacement signal) stands unchanged.

PR #3819, opened 2026-08-11
perf(server): narrow three read-path selects to cut per-row deserialization

ZacxDev:
## Why

Prisma Client walks the engine response **in JS, on the main thread**, and converts every tagged value per row: `DateTime` → `new Date`, `Json` → `JSON.parse`, `Decimal` → `new Decimal`, `BigInt`, `Bytes` → `Buffer.from` (plus a `JSON.parse` of the whole engine response).

So a column that a query selects but no consumer reads is **not** a free ride on a round-trip that was happening anyway — it is an allocation per row per request. Narrowing `select` is a direct main-thread CPU reduction with no architecture change and no behaviour change.

I ranked candidate read paths by request volume × rows-per-call × wasted tagged columns, then traced every field to its consumers before removing it. Most of the hot paths turned out to be **already narrow** (the big feed queries are raw SQL + Redis, and their satellite Prisma calls select Int/enum only). These three are what was left.

### 1. Comment reactions no longer hydrate a whole user + profile picture

`getReactionsSelect` selected `simpleUserSelect` for the `user` on every comment reaction. The renderer reads **two fields**:

- `ReactionPicker` → `reaction.user.username` (the "who reacted" tooltip, and the "have I reacted" check)
- its three callers → `r.user.id`

Everything else was fetched and dropped. `simpleUserSelect` pulls `deletedAt` (a `DateTime`) and the entire **`profilePicture` Image relation**, whose `metadata` is a non-null `Json`. Both of those costs are **conditional**, and an earlier revision of this description overstated them as unconditional — the corrected version:

- **An extra query, per call, whenever at least one reactor in the batch has a profile picture.** `profilePicture` is a relation and the schema's generator block enables only `previewFeatures = ["metrics"]` — no `relationJoins` — so Prisma has no join strategy available and resolves it with a second round-trip rather than a widened one. When the collected FK set is empty it skips that child query entirely. Measured on a minimal mirror of this shape (`CommentReaction → User → Image` via a nullable FK), Prisma 6.13.0, SQL captured via `$on('query')`, on SQLite and again on PostgreSQL 16:

  | reactors in the batch with a profile picture | `Image` SELECTs |
  |---|---|
  | 0 of 5 | **0** |
  | 1 of 5 | 1 — `WHERE "Image"."id" IN ($1)` |
  | 5 of 5 | 1 — `WHERE "Image"."id" IN ($1..$5)` |

  So it is one query per call, not per row, and only when the batch has at least one picture to fetch. This is plausibly the larger of the two costs.

- **A decode, per row, only where the value is non-null.** `User.deletedAt` is `DateTime?` and `User.profilePictureId` is `Int?`, so a live reactor with no profile picture paid neither. The cost landed as a `new Date()` per *deleted* reactor and a `JSON.parse` of the Image `metadata` per reactor who *has* a picture (that column is non-null `Json`, so it is always parsed once its row is fetched at all).

Narrowed to `{ id: true, username: true }`.

The deleted-user case is unaffected: `deleteUser` nulls `username`, and the renderer already falls back on `username ?? '<deleted user>'` — it reads the null, not the timestamp. Notably the three optimistic-update sites that hand-build a `ReactionDetails` were **already** passing `profilePicture: null` with a comment saying it wasn't needed; they now simply match the select.

This selector feeds both `comment.getAll` (the reactions arrive as `initialData`) and `comment.getReactions`, so the shapes stay identical on both sides of the hydration boundary.

### 2. Three full-row `Post` fetches for an ownership check

`entity-collaborator.service.ts` did `dbRead.post.findUnique({ where: { id } })` — no `select` — in `upsertEntityCollaborator`, `getEntityCollaborators` and `removeEntityCollaborator`, and read exactly one field off the result: `userId`. A full `Post` row is three DateTimes plus a `metadata` `Json`, plus the `title`/`detail` text. Narrowed to `select: { userId: true }`.

`getEntityCollaborators` is the one on a read path; the other two are the same defect in the same file and were fixed together rather than left to regrow.

### 3. A full `Post` row fetched to answer a boolean

`getAllImages` did `dbRead.post.findFirst({ where: { userId: -1, modelVersionId } })` and used the result only as a truthiness check (`hasSystemPosts`). Narrowed to `select: { id: true }`.

## Safety

A field that exists on a returned object can be read anywhere, so every dropped field was grepped across the repo — including client components and the `select`-derived types (`ReactionDetails`, `CommentGetReactions`) — before removal. The only reads of `reaction.user.*` anywhere are `.id` and `.username`. Nothing reads `deletedAt`, `image` or `profilePicture` off a reaction.

The only response-shape change is that reaction `user` object. Paths 2 and 3 change no response shape at all — the narrowed rows never leave their functions.

🔴 **That one response-shape change is reachable by third-party OAuth clients, and that is an accepted risk rather than a non-issue.** `comment.getAll`, `comment.getById` and `comment.getReactions` are `.meta({ requiredScope: TokenScope.MediaRead })`, and `MediaRead` is a grantable scope in both the `ReadOnly` and `Creator` presets — so an external app holding a user token can observe the object going from `{ id, username, deletedAt, image, profilePicture }` to `{ id, username }`. Concretely: an app rendering reactor avatars off `user.image` gets `undefined` and shows no avatar. The request still succeeds; this is a degraded render, not a server error. There is no compile-time contract with any external consumer to check against (nothing under `packages/` imports or re-exports `AppRouter`), and nothing in this repo is evidence about what third-party clients actually field. My read is that it is worth taking — the fields were fetched for an avatar this codebase never renders on a reaction, and re-adding a narrower `profilePicture` projection later is a one-line change — but that is a reviewer's call to make with the fact in front of them.

## Tests

Two files pin the projected shapes, and pair each shape assertion with a **behavioural** case so the guard cannot pass while the logic reading the surviving fields is broken:

- `src/server/services/__tests__/hot-path-select-narrowing.test.ts` — the three `entity-collaborator.service` owner lookups and the reaction selector
- `src/server/services/__tests__/hot-path-select-narrowing.image.test.ts` — the `hasSystemPosts` existence check (its own file because importing `image.service` needs an env/redis/submodule mock block and `vi.mock` is file-scoped)

**12 tests. Red at base, green at HEAD: 6 fail with the three source files reverted to `a9e9ddf533`, all 12 pass at HEAD** (measured in one tree, swapping only the source files). The 6 red are the shape assertions — one per narrowed call site, plus two on the reaction selector:

| Red at base | |
|---|---|
| `getEntityCollaborators › asks Postgres for ONLY the owner id` | `removeEntityCollaborator › asks Postgres for ONLY the owner id` |
| `upsertEntityCollaborator › asks Postgres for ONLY the owner id` | `getAllImages hasSystemPosts › projects the existence check…` |
| `getReactionsSelect › selects exactly the two user fields the renderer reads` | `getReactionsSelect › does not pull the profilePicture relation` |

The other 6 pass on both sides by design — they are behavioural **invariant guards** (owner-vs-stranger authorization on upsert and remove, pending-collaborator gating, the missing-post short-circuit, the surviving reaction fields, the `where` predicate) and are not counted as regression coverage.

Every shape assertion **on a Prisma call argument** pins the projection's **exact key set** with `toEqual`. It is deliberately never spelled as "does not contain `metadata`": an unnarrowed call has no `select` key at all, `Object.keys(undefined ?? {})` is empty, and an absence-check therefore passes on precisely the shape being rejected. `toEqual` fails on `undefined` by itself. The `getReactionsSelect` assertions are the exception, and they do use targeted presence/absence checks: that selector is a module-level object literal, not a call argument, so it has no "no `select` key at all" shape to pass vacuously on. Alongside its exact-key-set `toEqual`, the `not.toHaveProperty` trio there is reachable rather than vacuous — deleting `user.select` fails it with `TypeError: Cannot convert undefined or null to object` instead of passing it.

**Mutation matrix** — 14 mutants, each applied alone. Twelve kill exactly one test, by that test's own labelled assertion; the two rows marked **both** kill a pair, unavoidably, for the reason stated in the row.

| Mutation | Test that failed | Failure |
|---|---|---|
| `getEntityCollaborators` → bare `findUnique` | `getEntityCollaborators › asks Postgres for ONLY the owner id` | ``getEntityCollaborators: no `select` — a bare Prisma call fetches every column`` |
| `upsertEntityCollaborator` → bare `findUnique` | `upsertEntityCollaborator › asks Postgres for ONLY the owner id` | ``upsertEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `removeEntityCollaborator` → bare `findUnique` | `removeEntityCollaborator › asks Postgres for ONLY the owner id` | ``removeEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `hasSystemPosts` → bare `findFirst` | `getAllImages hasSystemPosts › projects the existence check…` | ``a bare findFirst fetches every Post column`` |
| `where: { userId: -1 }` → `-2` | `getAllImages hasSystemPosts › still asks the question it was asking` | `expected { userId: -2, … } to deeply equal { userId: -1, … }` |
| the `hasSystemPosts` branch made unreachable | **both** `getAllImages hasSystemPosts` cases | `getAllImages did not reach the hasSystemPosts check` — the harness positive control, which both cases run before asserting anything, so removing the call necessarily fails both |
| upsert owner check disabled | `upsertEntityCollaborator › still refuses a non-owner…` | `promise resolved "{ id: 1 }" instead of rejecting` |
| remove owner check disabled | `removeEntityCollaborator › still refuses a non-owner…` | `promise resolved "true" instead of rejecting` |
| `select: { id: true }` → `{ metadata: true }` | `getAllImages hasSystemPosts › projects the existence check…` | `expected { metadata: true } to deeply equal { id: true }` |
| `{ userId }` → `{ userId, metadata }`, at each of the 3 sites | that site's `asks Postgres for ONLY the owner id` | `expected { userId: true, metadata: true } to deeply equal { userId: true }` |
| `user.select` + `email: true` | `getReactionsSelect › selects exactly the two user fields…` | `expected { id: true, username: true, …(1) } to deeply equal { id: true, username: true }` |
| `user.select` + `profilePicture: true` | **both** `getReactionsSelect` shape cases | `expected { id: true, username: true, …(1) } to not have property "profilePicture"`, plus the exact-key-set `toEqual` beside it — any key added to `user.select` fails that too, so this assertion cannot be killed on its own |

Coverage of the 12 tests: those 14 mutants target **9 distinct tests**. The three with no mutant — `getEntityCollaborators › still gates pending collaborators…`, `getEntityCollaborators › returns [] when the post is missing…` and `getReactionsSelect › keeps the reaction fields…` — are invariant guards that hold on both sides of the narrowing, and are not counted as regression coverage above.

The shape and behavioural checks are independent: the `where`-clause mutation killed the behavioural test while the shape test stayed green.

## Gates

Measured on `1cad65e484`, all in one tree.

| Gate | Result |
|---|---|
| `pnpm typecheck` | **0 errors** in 66s (heap cap 8192 MB) |
| typecheck of the **test files themselves** | **0 new errors.** 🔴 `tsconfig.json` excludes `src/**/__tests__/**`, so `pnpm typecheck` does not see either test file in this PR — its "0 errors" says nothing about them. Re-run under a scoped config (`include` = the two test files + `src/**/*.d.ts`; without the `.d.ts` glob you get ~336 phantom ambient errors): 1 error, in `src/utils/signals/useSignalsWorker.ts`, a file this PR does not touch — **byte-identical count at `4df72dbb53`**, so the delta is 0. Instrument validated: planting `const x: number = 'not a number'` in the test file made the scoped run report it. |
| `pnpm test:unit:run` | 921 files, **14146 passed, 15 failed, 1 skipped**. All 15 are in `hiddenBlocks.test.ts` and `recent-source-images.store.test.ts` — neither in this PR's diff — and all are `localStorage is not available because --localstorage-file was not provided` on Node 26. Reproduced identically running those two files alone. |
| the two guard suites | 12 passed, 0 failed |
| `eslint` (all 8 changed files, `-f json`, 8 files confirmed processed) | 3 errors / 25 warnings, **all in `image.service.ts` and all pre-existing**: same three rules at base, the two line numbers shifted only by the 6 comment lines added (`prefer-const` 3705→3711, `no-module-scope-cache` 5568→5574). Base measured in the **same** tree by swapping the file, never across checkouts. The selector and both test files are 0/0. |
| `prettier --check` | clean on all three changed files. `pnpm prettier:check` only inspects *uncommitted* files and reports "no uncommitted .ts/.tsx files" on a committed branch — a zero about the file list, not the code — so the files were named explicitly, with a deliberately misformatted probe in the same invocation confirming the check can go red. |

## Not measured

This is a mechanism-level argument plus a volume ranking. I have **not** measured a CPU delta in a running environment — the per-row saving is real and arithmetic, but its share of total CPU is unquantified here. Worth watching after it ships rather than claiming a number now.

## Follow-ups deliberately left out

- Six `isOwnerOrModerator` middlewares (`image`, `comment`, `commentV2`, `answer`, `question`, `collection` routers) repeat the same full-row-for-`userId` pattern. Correct to fix, but they sit on mutations measuring well under 3 req/s combined, so the CPU case is not there — hygiene, not a perf win, and better as its own PR.
- Two paths do `select: { settings: true }` on `User` (a large accumulating `Json` blob) purely to read one boolean. Eliminating those parses needs a raw projection (`settings->>'…'`) rather than a `select` narrowing, which is a bigger change than this PR's premise.

### DIFF
--- a/src/components/CommentSection/CommentSectionItem.tsx
+++ b/src/components/CommentSection/CommentSectionItem.tsx
@@ -123,10 +123,7 @@ export function CommentSectionItem({ comment, modelId, onReplyClick }: Props) {
           reaction,
           user: {
             id: currentUser.id,
-            deletedAt: null,
             username: currentUser.username ?? '',
-            image: currentUser.image ?? '',
-            profilePicture: null, // not really necessary for reactions
           },
         };
         const reacted = previousReactions.find(
--- a/src/components/Model/Discussion/CommentThreadModal.tsx
+++ b/src/components/Model/Discussion/CommentThreadModal.tsx
@@ -55,10 +55,7 @@ export default function CommentThreadModal({
           reaction,
           user: {
             id: currentUser.id,
-            deletedAt: null,
             username: currentUser.username ?? '',
-            image: currentUser.image ?? '',
-            profilePicture: null, // Not really necessary for reactions
           },
         };
         const reacted = previousReactions.find(
--- a/src/components/Model/ModelDiscussion/CommentDiscussionItem.tsx
+++ b/src/components/Model/ModelDiscussion/CommentDiscussionItem.tsx
@@ -48,10 +48,7 @@ export function CommentDiscussionItem({ data: comment, modelUserId }: Props) {
           reaction,
           user: {
             id: currentUser.id,
-            deletedAt: null,
             username: currentUser.username ?? '',
-            image: currentUser.image ?? '',
-            profilePicture: null,
           },
         };
         const reacted = previousReactions.find(
--- a/src/server/selectors/reaction.selector.ts
+++ b/src/server/selectors/reaction.selector.ts
@@ -21,12 +21,14 @@ import { Prisma } from '@prisma/client';
  *    round-trip rather than a widened one. Selecting it meant one `Image`
  *    SELECT per call — not per row — whenever AT LEAST ONE reactor in the batch
  *    has a profile picture; when the collected FK set is empty Prisma skips the
- *    child query entirely. Measured on this exact shape (Prisma 6.13.0, PG16,
- *    SQL captured via `$on('query')`): a batch of 5 reactors issues 0 `Image`
- *    SELECTs with none of them carrying a picture, and 1 (`WHERE "Image"."id"
- *    IN (…)`) whether one carries one or all five do. That query is on top of
- *    the `User` SELECT the `user` relation still costs; dropping
- *    `profilePicture` removes it, rather than just a column.
+ *    child query entirely. Measured on a minimal MIRROR of this shape
+ *    (`CommentReaction -> User -> Image` via a nullable FK; Prisma 6.13.0, SQL
+ *    captured via `$on('query')`; reproduced on SQLite, and once on PG16): a
+ *    batch of 5 reactors issues 0 `Image` SELECTs with none of them carrying a
+ *    picture, and 1 (`WHERE "Image"."id" IN (…)`) whether one carries one or
+ *    all five do. That query is on top of the `User` SELECT the `user` relation
+ *    still costs; dropping `profilePicture` removes it, rather than just a
+ *    column.
  *
  * 2. Per-row decoding, but only CONDITIONALLY. Prisma converts tagged values
  *    on the JS main thread (`DateTime` -> `new Date`, `Json` -> `JSON.parse`),
--- a/src/server/selectors/reaction.selector.ts
+++ b/src/server/selectors/reaction.selector.ts
@@ -14,12 +14,19 @@ import { Prisma } from '@prisma/client';
  * Two distinct costs, and they are not the same size:
  *
  * 1. An EXTRA QUERY — plausibly the larger cost, and per call rather than per
- *    row. `profilePicture` is a relation, and the schema's generator block
- *    enables only `previewFeatures = ["metrics"]` — no `relationJoins` — so
- *    Prisma has no join strategy available and resolves a nested relation with
- *    an additional round-trip rather than a widened one. Selecting it meant an
- *    `Image` SELECT per call, on top of the `User` SELECT that the `user`
- *    relation still costs. Dropping it removes that query, not just a column.
+ *    row, but CONDITIONAL like the decode below. `profilePicture` is a
+ *    relation, and the schema's generator block enables only
+ *    `previewFeatures = ["metrics"]` — no `relationJoins` — so Prisma has no
+ *    join strategy available and resolves a nested relation with an additional
+ *    round-trip rather than a widened one. Selecting it meant one `Image`
+ *    SELECT per call — not per row — whenever AT LEAST ONE reactor in the batch
+ *    has a profile picture; when the collected FK set is empty Prisma skips the
+ *    child query entirely. Measured on this exact shape (Prisma 6.13.0, PG16,
+ *    SQL captured via `$on('query')`): a batch of 5 reactors issues 0 `Image`
+ *    SELECTs with none of them carrying a picture, and 1 (`WHERE "Image"."id"
+ *    IN (…)`) whether one carries one or all five do. That query is on top of
+ *    the `User` SELECT the `user` relation still costs; dropping
+ *    `profilePicture` removes it, rather than just a column.
  *
  * 2. Per-row decoding, but only CONDITIONALLY. Prisma converts tagged values
  *    on the JS main thread (`DateTime` -> `new Date`, `Json` -> `JSON.parse`),
--- a/src/server/selectors/reaction.selector.ts
+++ b/src/server/selectors/reaction.selector.ts
@@ -3,17 +3,38 @@ import { Prisma } from '@prisma/client';
 /**
  * Comment reactions only ever need to name the reactor.
  *
- * The renderer (`ReactionPicker`) reads `user.username` for the "who reacted"
- * tooltip and `user.id` to decide whether the viewer already reacted — nothing
- * else. Selecting the full `simpleUserSelect` here dragged in `deletedAt` and,
- * worse, the whole `profilePicture` Image relation, whose `metadata` is a
- * non-null `Json` column. Prisma decodes tagged values on the JS main thread,
- * so that was a `JSON.parse` plus a `new Date()` for every reaction on every
- * comment of every page — for two strings that are never rendered.
+ * Two fields are read, and nothing else. `ReactionPicker` reads
+ * `user.username` — for the "who reacted" tooltip, and for its own "have I
+ * reacted" comparison against the current user's username. Its three callers
+ * (`CommentSectionItem`, `CommentDiscussionItem`, `CommentThreadModal`) read
+ * `user.id`, to find the viewer's existing reaction before toggling it.
  *
- * The deleted-user case still works: `deleteUser` nulls `username`, and the
- * renderer already falls back on `username ?? '<deleted user>'`, so it reads
- * the null rather than the timestamp.
+ * This used to select the full `simpleUserSelect`, which additionally
+ * pulled `deletedAt`, `image` and the whole `profilePicture` Image relation.
+ * Two distinct costs, and they are not the same size:
+ *
+ * 1. An EXTRA QUERY — plausibly the larger cost, and per call rather than per
+ *    row. `profilePicture` is a relation, and the schema's generator block
+ *    enables only `previewFeatures = ["metrics"]` — no `relationJoins` — so
+ *    Prisma has no join strategy available and resolves a nested relation with
+ *    an additional round-trip rather than a widened one. Selecting it meant an
+ *    `Image` SELECT per call, on top of the `User` SELECT that the `user`
+ *    relation still costs. Dropping it removes that query, not just a column.
+ *
+ * 2. Per-row decoding, but only CONDITIONALLY. Prisma converts tagged values
+ *    on the JS main thread (`DateTime` -> `new Date`, `Json` -> `JSON.parse`),
+ *    and both dropped columns are nullable — `User.deletedAt` is `DateTime?`
+ *    and `User.profilePictureId` is `Int?` — so a live user with no profile
+ *    picture (the common case) paid neither. The decode cost landed only on
+ *    rows where the value was actually non-null: a `new Date()` per deleted
+ *    reactor, and, for each reactor who has one, a `JSON.parse` of that
+ *    Image's `metadata` (which is a non-null `Json` column, so it is always
+ *    parsed once the row is fetched at all).
+ *
+ * The deleted-user case still renders correctly: `deleteUser` nulls
+ * `username`, and the renderer already falls back on
+ * `username ?? '<deleted user>'`, so it reads the null rather than the
+ * timestamp.
  */
 export const getReactionsSelect = Prisma.validator<Prisma.CommentReactionSelect>()({
   id: true,
--- a/src/server/selectors/reaction.selector.ts
+++ b/src/server/selectors/reaction.selector.ts
@@ -1,11 +1,25 @@
 import { Prisma } from '@prisma/client';
-import { simpleUserSelect } from '~/server/selectors/user.selector';
 
+/**
+ * Comment reactions only ever need to name the reactor.
+ *
+ * The renderer (`ReactionPicker`) reads `user.username` for the "who reacted"
+ * tooltip and `user.id` to decide whether the viewer already reacted — nothing
+ * else. Selecting the full `simpleUserSelect` here dragged in `deletedAt` and,
+ * worse, the whole `profilePicture` Image relation, whose `metadata` is a
+ * non-null `Json` column. Prisma decodes tagged values on the JS main thread,
+ * so that was a `JSON.parse` plus a `new Date()` for every reaction on every
+ * comment of every page — for two strings that are never rendered.
+ *
+ * The deleted-user case still works: `deleteUser` nulls `username`, and the
+ * renderer already falls back on `username ?? '<deleted user>'`, so it reads
+ * the null rather than the timestamp.
+ */
 export const getReactionsSelect = Prisma.validator<Prisma.CommentReactionSelect>()({
   id: true,
   reaction: true,
   user: {
-    select: simpleUserSelect,
+    select: { id: true, username: true },
   },
 });
 
--- a/src/server/services/__tests__/hot-path-select-narrowing.image.test.ts
+++ b/src/server/services/__tests__/hot-path-select-narrowing.image.test.ts
@@ -28,9 +28,13 @@ import type * as PromClient from '~/server/prom/client';
  * absence assertion over `Object.keys(undefined ?? {})` passes on precisely
  * the shape being rejected; `toEqual` fails on `undefined` by itself. An
  * earlier revision followed the `toEqual` with a loop asserting each tagged
- * `Post` column was absent — unreachable, since any select containing one had
- * already failed the line above — and it has been removed. The `toBeDefined`
- * that remains is for its failure message, not for coverage.
+ * `Post` column was absent, and it has been removed. Given a FIXED `expected`
+ * that loop was redundant — any select containing one of those columns had
+ * already failed the line above — but it also went red when `expected` was
+ * widened ALONGSIDE the source, which a `toEqual` against the widened
+ * `expected` cannot catch, so removing it accepts the loss of that narrow
+ * ratchet against a test-file edit. The `toBeDefined` that remains is for its
+ * failure message, not for coverage.
  *
  * The mock block mirrors the established recipe from
  * `getAllImages-prioritized-guards.test.ts`, with one difference: `isFlipt`
--- a/src/server/services/__tests__/hot-path-select-narrowing.image.test.ts
+++ b/src/server/services/__tests__/hot-path-select-narrowing.image.test.ts
@@ -1,0 +1,180 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Type-only namespace imports rather than `typeof import('…')` inside the
+// factories: @typescript-eslint/consistent-type-imports forbids `import()` type
+// annotations, and this is the spelling the repo's own no-wholesale-module-mock
+// rule documents as the alternative.
+import type * as FliptClient from '../../flipt/client';
+import type * as PromClient from '~/server/prom/client';
+
+/**
+ * Sibling of `hot-path-select-narrowing.test.ts`, which pins the two
+ * `entity-collaborator.service` projections and the reaction selector.
+ *
+ * This one pins the third narrowed path: the `hasSystemPosts` existence check
+ * inside `getAllImages`. It lives in its own file because importing
+ * `image.service` requires a whole env / redis / clickhouse / submodule mock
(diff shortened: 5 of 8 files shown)


ZacxDev:
## Audit follow-up — one risk worth recording, and two accuracy fixes pushed

Posting rather than editing the description, since it may already have been read.

### The risk: the changed response shape is reachable by third-party OAuth clients

The PR body says the only response-shape change is the reaction `user` object. That is still true, but the body understated *who* can observe it. Three procedures that serve this shape are scope-gated rather than session-gated:

```
comment.getAll      .meta({ requiredScope: TokenScope.MediaRead })
comment.getById     .meta({ requiredScope: TokenScope.MediaRead })
comment.getReactions.meta({ requiredScope: TokenScope.MediaRead })
```

`MediaRead` is a grantable OAuth scope — it is a member of both the `ReadOnly` and `Creator` presets in `packages/civitai-auth/src/token-scope.ts` — so these are reachable by an external application holding a user-granted token, not only by our own web client.

**Concretely**, the reaction user object goes from

```ts
{ id, username, deletedAt, image, profilePicture }   // simpleUserSelect
```

to

```ts
{ id, username }
```

A third-party app that renders reactor avatars off `user.image` now reads `undefined`: no avatar where one used to appear. The request itself still succeeds — this is a degraded render, not a server error. (An app that dereferences `user.profilePicture.url` without a null guard would throw, but in its own code, not ours.)

**Why this cannot be resolved from inside the repo.** There is no compile-time contract with any external consumer to check against: `AppRouter` is exported only from `src/server/routers/index.ts`, and nothing under `packages/` imports or re-exports it (0 references), so no in-repo package mirrors the router types for outside consumption. And nothing in the repo is evidence about what third-party clients actually field — their absence here is not evidence of their absence in the wild.

So: **an accepted risk, recorded rather than argued away.** My read is that it is worth taking — the fields were fetched for an avatar this codebase never renders on a reaction, the blast radius is a missing image rather than a broken page, and re-adding a narrower `profilePicture` projection later is a one-line change if anyone reports it. But that is a judgement call, and it should be a reviewer's to make with the fact in front of them rather than one buried in a select.

### Also in this push

Two accuracy/coverage items from the same audit. Neither changes runtime behaviour.

**1. The `reaction.selector.ts` comment was wrong in both directions.** It claimed a `JSON.parse` plus a `new Date()` "for every reaction", unconditionally. Both dropped columns are nullable (`User.deletedAt` is `DateTime?`, `User.profilePictureId` is `Int?`), so the decode cost was *conditional* — a live user with no profile picture paid neither. Conversely the comment missed the larger effect: the schema's generator block enables only `previewFeatures = ["metrics"]`, with no `relationJoins`, so Prisma has no join strategy and resolves a nested relation with an additional round-trip rather than a widened one. `profilePicture` was therefore **an extra query per call**, which is plausibly the bigger win and was not mentioned at all. The comment now states both, correctly, and is explicit that the `user` relation still costs a query of its own. (Verified against `packages/civitai-db-schema/prisma/schema.prisma` and against the generated client, which exposes no `relationLoadStrategy` — the option only exists when `relationJoins` is on.)

**2. Three of the five narrowings had no shape assertion.** The PR narrows four Prisma call sites plus the reaction selector; only `getEntityCollaborators` and the selector were ratcheted. `upsertEntityCollaborator`, `removeEntityCollaborator` and the `hasSystemPosts` existence check could each have silently regrown a bare `findUnique`/`findFirst` with nothing failing. All three are now pinned, each paired with a behavioural case, and each follows the existing file's fix for the vacuity hazard — assert `select` is **defined** before asserting anything about its contents, because `Object.keys(undefined ?? {})` is empty and a "does not contain X" check otherwise passes on precisely the unnarrowed shape it is meant to reject.

Each new assertion was mutation-checked by reverting its own narrowing and confirming that specific test goes red for its own reason:

| Mutation | Test that failed | Failure |
|---|---|---|
| `upsertEntityCollaborator` → bare `findUnique` | `upsertEntityCollaborator › asks Postgres for ONLY the owner id` | ``upsertEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `removeEntityCollaborator` → bare `findUnique` | `removeEntityCollaborator › asks Postgres for ONLY the owner id` | ``removeEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `hasSystemPosts` → bare `findFirst` | `getAllImages hasSystemPosts › projects the existence check…` | ``a bare findFirst fetches every Post column`` |
| `where: { userId: -1 }` → `-2` | `getAllImages hasSystemPosts › still asks the question it was asking` | `expected { userId: -2, … } to deeply equal { userId: -1, … }` |
| upsert owner check disabled | `upsertEntityCollaborator › still refuses a non-owner…` | `promise resolved "{ id: 1 }" instead of rejecting` |
| remove owner check disabled | `removeEntityCollaborator › still refuses a non-owner…` | `promise resolved "true" instead of rejecting` |

Exactly one test died per mutation in each case, and the shape and behavioural checks proved independent — the `where`-clause mutation killed the behavioural test while the shape test stayed green, so neither is accidentally standing in for the other.

The `getAllImages` guard lives in a sibling file (`hot-path-select-narrowing.image.test.ts`) because importing `image.service` needs its own env/redis/submodule mock block and `vi.mock` is file-scoped.

### A correction to the gates table in the description

`tsconfig.json` excludes `src/**/__tests__/**`, so **`pnpm typecheck` does not type-check any test file in this PR** — the "0 errors" line is true but says nothing about `hot-path-select-narrowing.test.ts` or its new sibling. I found this by planting `const x: number = 'not a number'` in the new test file and watching `pnpm typecheck` still report `OK — 0 type errors`.

Type-checked separately against a scoped config that includes the two files: the planted error is caught there (`error TS2322`), and with it removed both files are clean. Worth knowing generally — it means a type error in any `__tests__` file reaches CI only as a runtime failure, if at all.

Re-measured gates after these changes:

| Gate | Result |
|---|---|
| `pnpm typecheck` (non-incremental) | 0 errors in 91s — unchanged, and structurally blind to test files as above |
| scoped `tsc` over the two test files | 0 errors; negative control confirmed it can go red |
| `pnpm test:unit:run` | 921 files, 14162 passed, 1 skipped, **0 failed** (the description reported 920 / 14156 at the previous commit; the deltas are the one new file and its 6 new tests) |
| `eslint` on the three files I touched | 0 errors, 0 warnings (3 files confirmed processed via `-f json`, not silently skipped) |
| `prettier --check` | clean; positive control in the same invocation was flagged |
| `image.service.ts` lint | byte-identical to the previous commit — that file is untouched by this push |

ZacxDev:
## The description has been corrected — what changed, in case you read the earlier version

A delta re-audit found the code fine and the residual entirely in the prose, all of it the same overstatement class the previous round was written to fix. Two of the three fixes are in the description rather than the diff, so this note exists because a body edit leaves no trace and a retraction buried in a comment thread is not where anyone looks.

### 1. §1 no longer claims an unconditional cost — and the "extra query" claim is now conditional too

The body still said the dropped `profilePicture` relation cost *"a `JSON.parse` plus a `new Date()` for every reaction, on every comment, on every page"*. That is the exact wording the code comment was corrected for last round; correcting the comment and leaving the body is how a reader ends up with the wrong number anyway.

Worse, the *replacement* claim I pushed last round had the same defect one layer down: the code comment asserted the relation "meant an `Image` SELECT per call", unconditionally, two lines above the paragraph that carefully conditioned the decode cost for the same nullable FK. It doesn't. Prisma skips the child query when the collected FK set is empty.

Settled by measurement rather than by reading the generated client — that static check (no `relationLoadStrategy` present ⇒ no join strategy) establishes query *splitting* and says nothing about a null short-circuit. A minimal mirror of the shape (`CommentReaction → User → Image` via a nullable FK), Prisma 6.13.0, `previewFeatures = ["metrics"]`, SQL captured via `$on('query')`, run on SQLite and then again against a local PostgreSQL 16:

| reactors in the batch with a profile picture | `Image` SELECTs |
|---|---|
| 0 of 5 | **0** |
| 1 of 5 | 1 — `WHERE "Image"."id" IN ($1)` |
| 5 of 5 | 1 — `WHERE "Image"."id" IN ($1..$5)` |

So "per call rather than per row" holds; "on every call" does not. Both the body and `reaction.selector.ts` now say *whenever at least one reactor in the batch has a profile picture*, and carry the measurement.

### 2. Two test assertions could never fail, while the header sold them as a second protection

`expectProjection` looped over `TAGGED_POST_COLUMNS` asserting each was absent — **directly after** `expect(args.select).toEqual(expected)`. Any select containing a tagged column already failed on the line above, so the loop could not report. Same shape in the image test. The previous round's own mutation matrix is the proof: the two mutants that re-add a tagged column both died on the `toEqual`.

Reordering would not have helped — `toEqual` on an exact key set is strictly stronger than an absence check in either position, so there is no mutant the loop can catch that it doesn't. Both loops are deleted, along with the constant and a separate `getEntityCollaborators` case that asserted strictly less than its sibling about every possible implementation. **12 tests now, was 13.** Nothing is undetectable that was detectable before; the three `{ userId }` → `{ userId, metadata }` mutants in the matrix below are run once per call site to show each narrowing independently catches exactly what the loop claimed to.

The file header's *"Every shape assertion asserts `select` is DEFINED before asserting anything about its contents"* was also simply false — `getEntityCollaborators` and `hasSystemPosts` went straight to `toEqual`. Reworded to what is true and stronger: the shape is pinned by an exact-key-set `toEqual`, which fails on `undefined` by itself, so no absence-check remains to be vacuous. The surviving `toBeDefined` lines are kept for their failure message and now say so. `getEntityCollaborators` was routed through the shared helper so its mutant reports that named message too, rather than a bare `undefined` mismatch.

### Full mutation matrix, re-run after the restructure

A fix round resets the verification gate, so this is the whole matrix again, not the delta. Each mutant applied alone; the mutation tool asserts its anchor occurs exactly once and aborts otherwise, so a mutation cannot silently no-op (a no-op reads as "the guard survived").

| Mutation | Test that failed | Failure |
|---|---|---|
| `getEntityCollaborators` → bare `findUnique` | `getEntityCollaborators › asks Postgres for ONLY the owner id` | ``getEntityCollaborators: no `select` — a bare Prisma call fetches every column`` |
| `upsertEntityCollaborator` → bare `findUnique` | `upsertEntityCollaborator › asks Postgres for ONLY the owner id` | ``upsertEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `removeEntityCollaborator` → bare `findUnique` | `removeEntityCollaborator › asks Postgres for ONLY the owner id` | ``removeEntityCollaborator: no `select` — a bare Prisma call fetches every column`` |
| `hasSystemPosts` → bare `findFirst` | `getAllImages hasSystemPosts › projects the existence check…` | ``a bare findFirst fetches every Post column`` |
| `where: { userId: -1 }` → `-2` | `getAllImages hasSystemPosts › still asks the question it was asking` | `expected { userId: -2, … } to deeply equal { userId: -1, … }` |
| branch made unreachable (`&& false &&`) | both `getAllImages hasSystemPosts` cases | `getAllImages did not reach the hasSystemPosts check` — the harness positive control, which is the intended kill |
| upsert owner check disabled | `upsertEntityCollaborator › still refuses a non-owner…` | `promise resolved "{ id: 1 }" instead of rejecting` |
| remove owner check disabled | `removeEntityCollaborator › still refuses a non-owner…` | `promise resolved "true" instead of rejecting` |
| `select: { id: true }` → `{ metadata: true }` | `getAllImages hasSystemPosts › projects the existence check…` | `expected { metadata: true } to deeply equal { id: true }` |
| `{ userId }` → `{ userId, metadata }` at the **getEntityCollaborators** site | `getEntityCollaborators › asks Postgres for ONLY the owner id` | `expected { userId: true, metadata: true } to deeply equal { userId: true }` |
| …at the **upsert** site | `upsertEntityCollaborator › asks Postgres for ONLY the owner id` | same |
| …at the **remove** site | `removeEntityCollaborator › asks Postgres for ONLY the owner id` | same |

Twelve mutants, exactly the intended test dead in each (two, for the unreachable-branch mutant, both by the same harness control), 11 of 12 passing in every other case.

### Also corrected in the body

- **The test count.** It said "4 of the 7 fail against `origin/main`, all 7 pass here". There are 12 tests, and **6** are red with the three source files reverted to `a9e9ddf533` — measured in one tree by swapping only those files. The 6 red and the 6 green-on-both-sides invariant guards are now listed by name.
- **The typecheck caveat is now in the body, not only in a comment.** `tsconfig.json` excludes `src/**/__tests__/**`, so `pnpm typecheck`'s "0 errors" says nothing about either test file. Re-run under a scoped config: 1 error, in `src/utils/signals/useSignalsWorker.ts` — untouched by this PR and present identically at `4df72dbb53`, so the delta is 0. The scoped config was validated by planting a deliberate type error and watching it get reported. (Side note for someone else's day: that error is invisible to the repo-wide `pnpm typecheck`, which reports 0. Not this PR's to chase.)
- **The eslint and prettier rows now say how they were measured.** eslint base-vs-head in the *same* tree (a second checkout resolves a different `@typescript-eslint`): 3 errors / 25 warnings, all in `image.service.ts`, same three rules at base with two line numbers shifted by the 6 added comment lines. `pnpm prettier:check` only inspects *uncommitted* files, so on a committed branch its "no uncommitted .ts/.tsx files" is a fact about the file list rather than the code — the three files were named explicitly instead, with a misformatted probe in the same invocation to prove the check can go red.
- **The third-party OAuth exposure** recorded in my previous comment is now in the body's Safety section, where a reviewer will actually meet it.

No runtime behaviour changed in this push. Still draft.

Issue #3772, opened 2026-08-09
A per-side pixel ceiling on listing uploads needs a client-side precheck, and 8192px rejects legitimate upscaled art


ZacxDev:
Splitting this out of #3738, where a per-side pixel ceiling was briefly added to the listing-asset
upload path and then removed before merge. The bound may well be worth having — but it cannot land
in its current shape without a client-side precheck, and it would reject images authors legitimately
produce.

## The narrowing

`LISTING_ASSET_MAX_DIMENSION_PX` is 8192 per side. Today it is enforced **only** on the OG-pull path
in `src/server/services/blocks/listing-meta.service.ts`. It has never applied to client uploads, and
`validateListingImage` has no maximum-dimension check for cover or screenshot at all (only icon, at
4096).

Applying it to uploads rejects images that are nowhere near the 4 MiB byte cap. Measured:

| image | size | over 8192/side? |
|---|---|---|
| 10000×5000 PNG | 677 KB | yes |
| 12000×6000 JPEG | 422 KB | yes |
| 9000×4000 JPEG | 211 KB | yes |

4× upscaled AI art routinely exceeds 8192px on one side, which makes this a core-use-case rejection
rather than an edge case.

## The UX problem, which is the blocking part

`src/components/Apps/ListingAssetStep.tsx:294` already calls `readImageDimensions(file)` — and then
discards the result. So with a server-side-only ceiling, the author selects a file, waits for the
**entire upload to the object store to complete**, and only then receives a rejection for something
that was knowable before a single byte left the browser.

If the ceiling is added, the precheck has to land with it: validate against the same constant in
`ListingAssetStep.tsx` before starting the upload, so the error is immediate and no bandwidth is
wasted. Any CLI-side equivalent would be a separate change and should not vendor the constant (see
`civitai/cli#270` for why — a copied bound that goes stale starts refusing valid input).

## Also worth settling if this is picked up

The motivating case for a ceiling is a decompression bomb, but sharp's **default**
`limitInputPixels` (~268M px) already throws first on genuinely huge inputs — 20000×20000,
16384×16384 and 100000×100000 all fail at `metadata()` and surface as "couldn't be read". A
per-side ceiling of 8192 therefore only governs the band between 8193px/side and ~268M total pixels
(e.g. 65535×4000 decodes fine). So decide what the bound is actually *for*: bomb protection is
already covered, and this would be a product decision about maximum sensible artwork size.

[...]

PR #3824, opened 2026-08-11
fix(app-listings): check the listing-asset dimension bound before uploading, not after

ZacxDev:
Follow-up to #3801. Refs #3772 — **does not close it**; see below.

## The gap

#3801 applied `listingAssetTooLargeReason` to listing media on the upload path, but the only place it runs there is `persistAssetImage`, which the client calls **after** the file has finished uploading. `uploadAndPersist` in `ListingAssetStep.tsx` already reads the image's intrinsic dimensions immediately before it uploads, and then does nothing with them — so an author picks a file, waits out the whole transfer, and is then refused for something that was knowable from the file picker.

Issue #3772 called this out as the blocking half of adding the ceiling at all. The ceiling shipped in #3801; this is the other half.

## The change

Take the same predicate at the point the dimensions are first known:

```ts
const { width, height } = await readImageDimensions(file);
const tooLarge = listingAssetTooLargeReason('That image', width, height);
if (tooLarge) throw new Error(`${tooLarge}.`);
const result = await uploadToCF(file);
```

Three properties worth stating explicitly, because each is a way this could have gone wrong:

- **Imported, not restated.** No constant is copied to the client. A vendored bound is exactly what goes stale and starts refusing valid input — `civitai/cli#270`.
- **Not the enforcement point.** The server still re-derives the dimensions from the stored bytes and re-applies the bound; nothing here is trusted. Deleting this check costs a wasted upload, never a bypass. That is written into the comment so a later reader does not mistake it for a gate.
- **Cannot disagree with the server.** The bound is over `Math.max(width, height)`, which is invariant under the EXIF quarter-turn the server applies and the browser does not. The subject string matches the server's, so the author sees one message wherever the rejection lands.

`ListingAssetStep` is the single client surface for listing media — the create wizard, the edit form and the media editor all embed it — so this is one call site, not a predicate spread across three.

**No behaviour change to what is accepted.** This rejects exactly the set the server already rejects, and it rejects it sooner.

## Tests

`src/components/Apps/ListingAssetStep.browser.test.tsx` (component project). Matrix, measured — not asserted:

| test | at `origin/main` | here | what it is |
|---|---|---|---|
| over-bound cover refused **before** the upload starts | **RED** | green | regression coverage |
| over-bound on the **other** axis (tall icon) | **RED** | green | regression coverage |
| a cover **exactly** at the bound still uploads | green | green | invariant guard + positive control |

Two of the three are regression coverage; the third is not, and is labelled as such in the file. 16/16 green here, 14/16 at `origin/main`.

The fixture helper now takes explicit dimensions and **asserts the canvas actually produced them** before any test reasons about which side of the bound it falls on — otherwise an "under the bound" result could be a fact about the fixture rather than about the code.

### Mutation results

Each mutation applied to the shipped code, suite re-run, failure message read to confirm it attributes to this guard:

| mutation | outcome |
|---|---|
| guard deleted (i.e. `origin/main`) | killed — over-bound tests fail |
| guard kept, specific message replaced with a generic one | killed — the message locator, so the wording is pinned |
| check moved to **after** `uploadToCF` | killed — and by the `not.toHaveBeenCalled()` assertion, not the message, so "before the upload" is genuinely pinned rather than incidental |
| comparison inverted (`if (!tooLarge)`) | killed — 5 tests, including the at-bound one, which is what makes that test non-vacuous |
| single-axis (`width, width`) | killed — **only** by the tall-icon test, which is why that test exists |
| arguments swapped (`height, width`) | **survived.** Honest result, not a gap: the predicate is `Math.max(width, height)`, so the swap is semantically identical and no test can distinguish it. |

## Checks

- `pnpm typecheck`: sorted error sets at `origin/main` and at this branch are **identical** (12 either side, all pre-existing and unrelated to these files). Zero introduced.
- `prettier`: both files are already non-conformant on `main`; this branch adds **zero** new formatting deltas to either (29→29 and 63→63 changed lines). Left alone rather than reformatted, to keep the diff readable.

## What this deliberately does not do

#3772 also asks whether listing media should carry an **area** bound — covers and screenshots have none today, so `8192×4096` (33 Mpx) is accepted while an icon path enforces a 16.7 Mpx decode budget.

I did not pick a number. Every candidate that actually constrains anything rejects images a real author could plausibly produce, and the "just unify with the icon path" framing does not survive reading what that constant is for. The analysis, the options and a recommendation are in a comment on #3772 — it is a product decision, and guessing at it here to have something to ship would have been the wrong call.

### DIFF
--- a/src/components/Apps/ListingAssetStep.browser.test.tsx
+++ b/src/components/Apps/ListingAssetStep.browser.test.tsx
@@ -1,5 +1,8 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
+// The bound the precheck applies, IMPORTED rather than spelled: a literal here
+// would agree with a precheck that hard-coded a different number.
+import { LISTING_ASSET_MAX_DIMENSION_PX } from '~/server/schema/blocks/app-listing.schema';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
 import { renderWithProviders } from '../../../test/component-setup';
 
@@ -130,14 +133,27 @@ function renderStep(props: Partial<Parameters<typeof ListingAssetStep>[0]> = {})
  * fail `createImageBitmap` in the upload path (readImageDimensions), so the row
  * would never reach the scanning state we assert on.
  */
-async function makeImageFile(name = 'shot.png'): Promise<File> {
+async function makeImageFile(name = 'shot.png', width = 200, height = 150): Promise<File> {
   const canvas = document.createElement('canvas');
-  canvas.width = 200;
-  canvas.height = 150;
+  canvas.width = width;
+  canvas.height = height;
   const blob = await new Promise<Blob>((resolve, reject) => {
     canvas.toBlob((b) => (b ? resolve(b) : reject(new Error('toBlob null'))), 'image/png');
   });
-  return new File([blob], name, { type: 'image/png' });
+  const file = new File([blob], name, { type: 'image/png' });
+  // The dimension precheck below is the whole point of the oversize fixtures, so
+  // assert the fixture actually HAS the dimensions asked for before any test
+  // reasons about which side of the bound it falls on — a canvas that silently
+  // clamped would make an "under the bound" result a fact about the fixture.
+  const bitmap = await createImageBitmap(file);
+  try {
+    if (bitmap.width !== width || bitmap.height !== height) {
+      throw new Error(`fixture is ${bitmap.width}×${bitmap.height}, asked for ${width}×${height}`);
+    }
+  } finally {
+    bitmap.close();
+  }
+  return file;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ZacxDev:
Correcting the framing before this merges, from an adversarial review. Neither item changes the code; both change what the PR claims.

### This fixes the late-rejection problem for ONE bound, not for the step

The body opens on authors waiting out a full transfer to be refused for something knowable at the file picker. What lands here mirrors only the **kind-agnostic** per-side bound, which is enforced at persist. Every **kind-specific** bound still lives at attach — which the client reaches only after the upload *and* the persist have completed:

- icon maximum per side (4096, i.e. **half** the bound this PR prechecks)
- icon aspect, cover aspect, screenshot aspect
- the per-kind minimum dimensions

So a 5000px icon, or a wrongly-shaped one, behaves exactly as it did before this change: full upload, then refused. That is plausibly the *more common* rejection, since a wrongly-shaped icon is easier to produce by accident than a >8192px one. Not a regression and not a correctness defect — the server remains the enforcement point either way — but "the other half of #3772" reads as more complete than it is.

The remainder is small and already in reach: `startAttach` holds `kind`; `uploadAndPersist` simply does not take it. A follow-up will thread it through so the kind-specific bounds precheck too.

### The EXIF comment's premise is wrong (the conclusion is right)

The comment says the bound is safe to evaluate client-side "even though the server reads EXIF orientation and the browser does not". The browser **does**: the dimension read uses `createImageBitmap`, whose `imageOrientation` defaults to `from-image`, and a measurement on a quarter-turned JPEG returns the same transposed pair the server derives. So client and server agree on **both axes**, not merely on `Math.max` — a stronger guarantee than the one claimed.

That matters beyond tidiness: anyone trusting the stated premise would conclude the client's dimensions are unusable for the *aspect* and minimum-dimension checks, which is exactly what the follow-up needs them for. Being corrected in that follow-up rather than here, since it rewrites these lines anyway.

PR #3826, opened 2026-08-11
fix(app-listings): check the per-KIND listing-asset geometry before uploading too

### LINKED ISSUE #3772: A per-side pixel ceiling on listing uploads needs a client-side precheck, and 8192px rejects legitimate upscaled art

ZacxDev:
#3824 moved the kind-agnostic per-side ceiling to the point the dimensions are first known, before the upload. It only half-closed #3772: every *per-kind* rule still lived at attach, which the client reaches only after the upload **and** the persist.

So these all still cost a full transfer before the rejection:

- the icon's own longer-side maximum — **half** the ceiling #3824 prechecks
- the icon / cover / screenshot aspect bands
- the icon / cover / screenshot minimum dimensions

A wrongly-shaped icon is far easier to produce by accident than an over-ceiling one, so this is plausibly the more common rejection — and late geometry validation is the original complaint behind civitai/cli#270.

## What changed

`startAttach` already holds the kind; `uploadAndPersist` now takes it and applies `validateListingImage` — the server's own predicate, **imported, not restated** — right after the existing ceiling check.

The order mirrors the server's, so the author is shown the message that would actually have landed: the kind-agnostic ceiling is what `persistAssetImage` applies (subject "That image …"), and the per-kind rules are what the attach procs apply one step later (`reason` passed through verbatim, no trailing period). An image failing the first never reaches the second — here or on the server — so the two messages cannot swap places.

**Prechecked:** every geometry bound `validateListingImage` expresses — the per-side ceiling, the per-kind aspect bands, the per-kind minimums, and the icon's own maximum — **except on WebP**, where this side cannot tell which axis is which; see the next section.

**Deliberately left server-only,** because this side cannot evaluate them faithfully:

- **MIME** — `file.type` comes from the file *name*, while the server reads the *decoded* container format. A mislabelled extension, or an empty `file.type`, would refuse an image the server accepts. Refusing valid input at the picker is a worse failure than the wasted upload this is fixing.
- **Byte size** — the caps are split across two server gates (a whole-asset ceiling at persist, the per-kind cap at attach), so one verdict here would name the wrong gate for anything above the ceiling.
- `type` is not a claim: `createImageBitmap` has already thrown for anything that is not a decodable image.

Both exclusions are now **guarded rather than only described**: an asserted ledger of the exact fields handed to `validateListingImage` (which fails if the set grows *or* shrinks), plus one behavioural case each — a cover past the byte cap, and a cover whose `file.type` is an unsupported MIME — that must still reach the object store. Adding either field previously survived the entire suite.

The server remains the enforcement point and re-derives everything from the stored bytes. Nothing became conditional on the client having run — deleting the check costs a wasted upload, never a bypass.

## The orientation premise, measured

#3824 said the precheck was safe "even though the server reads EXIF orientation and the browser does not". That was wrong — but so was this PR's first revision, which claimed the two agree on both axes. **The truth is per-container.** Measured on Chromium 149:

| container | browser applies EXIF orientation | agrees with the server |
| --- | --- | --- |
| JPEG, PNG | yes | yes, on **both axes** |
| **WEBP** | **no** | **no — the axes arrive transposed** |

A transposition maps an aspect `a` to `1/a`, so it *inverts* the verdict of any bound naming an axis. The cover band (1.3–2.4) lies entirely above 1, so **every EXIF-rotated WebP cover the server would happily store was refused at the file picker** — no upload, and therefore no server verdict able to overrule it. Measured end to end on a stored 640×1280 WebP with `Orientation=6`: the server reads 1280×640 (aspect 2.00, accepted), the browser reads 640×1280 (aspect 0.50, refused). WebP is offered by the file inputs' own `accept`, so this was a live path.

The precheck now applies only the bounds a quarter turn cannot change, whenever the file is a WebP:

- **skipped** — the icon / cover / screenshot aspect bands, *and* the cover's minimum **width**, which names an axis rather than the shorter side. That last one is easy to miss: skipping the aspect band alone still refuses a valid rotated cover, one bound further down.
- **kept** — the kind-agnostic per-side ceiling, the icon's minimum and maximum side, the screenshot's minimum shorter side. All are stated over `Math.min` / `Math.max`, which a transposition leaves alone.

The container is detected from the **file's own first 12 bytes** (`RIFF`…`WEBP`), never from `file.type` — gating on a filename-derived MIME would land straight back in the over-strict direction this fixes, and would do it silently. The narrowing itself lives in `validateListingImage`, as an option beside the bounds it governs, so no bound is vendored into the component. The server passes no options and is unchanged; the 182 existing server-side tests across the three suites that exercise the predicate still pass.

Direction matters: skipping a client check restores a *late* rejection, which is merely the cost this PR set out to reduce. Refusing something the server would accept has no recovery at all.

`imageOrientation: 'from-image'` is still passed explicitly to state the intent, but it is **not** what buys the agreement — in this Chromium the option is inert: `'from-image'`, `'none'`, and no options at all return the same pair for both containers.

## Message fidelity — one honest caveat

"The author is shown the message that would actually have landed" holds for **geometry**, not universally. The server reaches byte size and format *before* geometry (`measureListingAssetUpload` → `measureUploadedImage`, then the format→MIME map), and the per-kind byte cap before geometry inside `validateListingImage`. So for a file over 4 MiB, an icon over 1 MiB, or a GIF/SVG, the client now names a *geometry* reason where the server would have named size or format. Both still reject, and the rejection is still correct — only the wording differs.

## Tests

`ListingAssetStep.browser.test.tsx` — **47 pass** (13 new this round), plus **3 new node-project tests**. Every number below was re-measured against both baselines rather than carried over:

| | count |
| --- | --- |
| red at `main` (8da95972bf) | **16** |
| red at this PR's previous revision (442dcb727e) | **5** |
| distinct tests red on at least one baseline | **20** |
| green on both baselines | **27** |

The two red sets overlap in exactly one test, and the split is meaningful rather than bookkeeping: the 16 are regression coverage for the PR's *original* subject (per-kind rules reaching the picker at all), while 4 of the 5 are green at `main` precisely because `main` never had the WebP defect — the PR introduced it, so only the PR's own previous revision is the baseline that can show it.

Of the 13 new tests: 4 red only at the previous revision (the orientation defect), 1 red at both, 4 red only at `main`, and 4 green on both — the last group being mutation guards, each justified by the specific mutant it kills below, and none counted as regression coverage.

The 27 green-on-both are the accept half at each boundary plus those guards. Not regression coverage, and labelled as such in the file. They exist because over-rejection at the picker is the worse failure and only an accept case can catch it — which is not hypothetical: the aspect-band swap mutant is caught **only** by them, and the WebP defect itself was an over-rejection that a suite of refusals could not see.

Also: pre-existing fixtures were resized to be valid listing media (the client now applies the real predicate, so "it decodes" is no longer enough), and a screenshot slot renders only a status badge, so for that kind the rejection reason is asserted on `showErrorNotification` — the channel that actually carries it.

### Fixtures that carry real EXIF

The suite could not previously observe orientation **at all**: every fixture was a fresh canvas raster with no EXIF, so mutating the dimension read from `'from-image'` to `'none'` left it 34/34 green. `test/fixtures/exif-image.ts` builds real EXIF-oriented WebP and JPEG containers by byte surgery, and both readers are pinned:

- the **browser's**, per fixture — the expected client pair is asserted at construction, so an engine that changes its mind fails loudly instead of the suite going green for a new reason;
- the **server's**, in `listing-asset-exif-fixture.test.ts` — `sharp` must read `orientation: 6` off the same bytes, with an untouched raster as the negative control. Without that half, a container carrying no *readable* EXIF would make the browser assertion pass for the wrong reason.

### This round — 14 mutants, re-run against the final 49-test suite

🔴 **Corrected.** Four rows below were each **one kill low** in the first version of this table: the counts were computed before the `a WEBP whose aspect is genuinely wrong is now left to the SERVER` test landed, despite this heading previously claiming they were all re-run. The corrected numbers, and the like-for-like 47-test figures, are in [this comment](https://github.com/civitai/civitai/pull/3826#issuecomment-5262090590). Counts below are a full re-run against the final suite (49 tests: 47 plus the two guards added in `f54e07c9db`), read off vitest's `Tests` summary line, with the unmutated baseline (`0 failed`) and a known-killed mutant (`1 failed`) as controls.

| mutant | killed by |
| --- | --- |
| WebP sniff deleted | 5 |
| WebP sniff inverted | 16 |
| WebP sniff always true | 11 |
| sniff offsets wrong (`WEBP` sought at byte 4) | 5 |
| **sniff gated on `file.type` instead of the bytes** | exactly **1** — the mislabelled-`.png` case, the only test that can tell the two apart |
| `skipOrientationSensitive` inverted in the schema | 15 |
| skip widened to cover the icon minimum | exactly **1** — the narrowness guard |
| skip widened to cover the icon **maximum** | exactly **1** — the narrowness guard added in `f54e07c9db`. **Survived all 47 tests before it.** Not covered by the ceiling guard: 4096 is *half* the 8192 kind-agnostic ceiling, so a fixture that trips the icon maximum never reaches the ceiling check |
| skip widened to cover the screenshot **minimum** shorter side | exactly **1** — the other narrowness guard added in `f54e07c9db`. **Survived all 47 tests before it.** |
| skip widened to cover the per-side ceiling in `validateListingImage` | **equivalent — survives.** The client reaches the identical ceiling through `listingAssetTooLargeReason` first (which is what the refusal message names), and the server passes no options, so no reachable input distinguishes it |
| `sizeBytes: file.size` added to the call | 2 — the ledger, and the byte-cap behavioural case |
| `mimeType: file.type` added to the call | 2 — the ledger, and the unsupported-MIME behavioural case |
| skip **not** applied to the cover minimum width | **survived the first pass** — a real gap in the new code. A test was added; it now kills it, and nothing else does |
| `'from-image'` → `'none'` | **survives, and cannot be killed on this engine.** Chromium 149 returns the same pair for both values and for no options at all — the option is inert here. Reported rather than papered over: no fixture can make a test observe a difference the browser does not make |

The two rows marked "survived all 47 tests" were the property's real gap: `the skip is NARROW` was pinned at **1 of the 4 bounds the skip keeps**. Both mutants are genuine behaviour changes, not equivalent mutants — a 4097px WebP icon, or a 300px WebP screenshot, would have gone from refused-at-the-picker to uploaded. Each was re-applied and watched to fail its own new test on `showErrorNotification` "Number of calls: 0" — no refusal produced at all. Neither is regression coverage; both pass on unmodified code and are labelled as guards.

Each killing test was checked to fail for *its own* reason — the behavioural ones on "upload expected once, called 0 times" or a missing rejection notification, the ledger on the key set itself — not incidentally.

### Previous round, carried over — **not re-run against the final suite**

Listed for continuity only; treat the counts as claims from that round, not as re-verified here.

| mutant | killed by |
| --- | --- |
| guard deleted | the 11 refusals, and no accept case |
| guard moved after the upload call | the same 11 — they assert the upload never started |
| kind hardcoded to icon / cover / screenshot | 14 / 14 / 11 — every test of the *other* kinds |
| kind rotated to the wrong slot | 19 |
| dimensions transposed (aspect inverted) | 11 |
| reason replaced with house prose | the 11 refusals, on message mismatch |
| icon maximum widened to the kind-agnostic ceiling | exactly 1 — the icon-cap refusal, and nothing else |
| icon aspect band min/max swapped | 7, all icon — the two aspect refusals survive; only the accept cases catch it |

Refs #3772. Follows #3824.

### DIFF
--- a/src/components/Apps/ListingAssetStep.browser.test.tsx
+++ b/src/components/Apps/ListingAssetStep.browser.test.tsx
@@ -14,9 +14,15 @@ import {
   LISTING_SCREENSHOT_ASPECT_MAX,
   LISTING_SCREENSHOT_ASPECT_MIN,
   LISTING_SCREENSHOT_MIN_PX,
+  MAX_LISTING_COVER_SIZE_BYTES,
 } from '~/server/schema/blocks/app-listing.schema';
+import type * as AppListingSchema from '~/server/schema/blocks/app-listing.schema';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
 import { renderWithProviders } from '../../../test/component-setup';
+import {
+  jpegWithExifOrientation,
+  webpWithExifOrientation,
+} from '../../../test/fixtures/exif-image';
 
 /**
  * W13 — the shared listing ASSET step, focused on the OG-image AUTO-FILL path
@@ -64,7 +70,35 @@ const mocks = vi.hoisted(() => ({
   // badge, so for screenshots this notification is the ONLY channel the rejection
   // reason travels on.
   showErrorNotification: vi.fn(),
+  // A pass-through spy over the server's own predicate, so a test can read the
+  // ARGUMENTS the component hands it — see the exclusions ledger below.
+  validateListingImage: vi.fn(),
 }));
+
+/**
+ * Wrap `validateListingImage` in a recording pass-through. The real implementation
+ * still decides every verdict (so no behaviour is mocked away); the spy exists only
+ * so the exclusions ledger can assert WHICH FIELDS the component supplies.
+ *
+ * 🔴 That ledger is the guard the PR body's two "deliberately left server-only"
+ * exclusions did not have: adding `sizeBytes: file.size` or
+ * `mimeType: file.type || undefined` to the call survived the whole suite at 34/34,
+ * even though the PR body itself argues those would refuse images the server
+ * accepts. A prose warning is not a guard.
+ */
+vi.mock('~/server/schema/blocks/app-listing.schema', async (importOriginal) => {
+  // Type-only NAMESPACE import, not `typeof import('…')` — the latter is an
+  // `import()` type annotation, which `@typescript-eslint/consistent-type-imports`
+  // rejects.
+  const actual = await importOriginal<typeof AppListingSchema>();
+  return {
+    ...actual,
+    validateListingImage: (...args: Parameters<typeof actual.validateListingImage>) => {
+      mocks.validateListingImage(...args);
+      return actual.validateListingImage(...args);
+    },
+  };
+});
 
 vi.mock('~/utils/trpc', () => {
   return {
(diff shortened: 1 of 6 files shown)


ZacxDev:
## Correction: four rows of the mutation table were each one kill low

Posting this rather than only editing the body, since the original numbers have been readable for a while.

The body's "This round — 12 mutants, **all re-run against the final suite**" table undercounts four rows by exactly one. The direction is favourable — the guards are stronger than advertised, not weaker — but the sentence claiming they were all re-run against the final suite was not accurate, and that is worth saying out loud.

**Why.** In each of the four cases the missing killer is the same test, `a WEBP whose aspect is genuinely wrong is now left to the SERVER (the accepted cost)`. That test is present only from `54a7a3cf5d` — `git show 442dcb727e:src/components/Apps/ListingAssetStep.browser.test.tsx` does not contain it — so the table was computed partway through that commit's development and not re-derived after the test landed.

### Re-run, in full, by me

Every row below was re-applied and re-run just now, not copied. Counts are read off vitest's own `Tests` summary line. The suite is **49** tests — 47 as of `54a7a3cf5d`, plus the two guards added in `f54e07c9db`. Harness controls: unmutated baseline reports `0 failed | 49 passed`, and a known-killed mutant reports `1 failed`, so the counter has been watched at both ends.

| mutant | body said | this re-run (49-test suite) |
| --- | --- | --- |
| WebP sniff deleted | 3 | **5** |
| WebP sniff inverted | 14 | **16** |
| WebP sniff always true | 11 | 11 |
| sniff offsets wrong (`WEBP` sought at byte 4) | 3 | **5** |
| sniff gated on `file.type` instead of the bytes | 1 | 1 |
| `skipOrientationSensitive` inverted in the schema | 13 | **15** |
| skip widened to the icon minimum | 1 | 1 |
| skip widened to the icon **maximum** | *(absent)* | **1** — new guard, see below |
| skip widened to the screenshot **minimum** | *(absent)* | **1** — new guard, see below |
| skip widened to the per-side ceiling | equivalent, survives | **0 — survives**, confirmed |
| skip **not** applied to the cover minimum width | killed | 1 |
| `sizeBytes` added to the call | 2 | 2 |
| `mimeType` added to the call | 2 | 2 |
| `'from-image'` → `'none'` | survives | **0 — survives**, confirmed |

Two of my numbers are not directly comparable to the body's, because the denominator moved. Subtract the one new icon-maximum guard — which kills all four of the disputed mutants and none of the others — and the **like-for-like figures against the 47-test suite the body describes are 4 / 16 / 4 / 15**, versus the 3 / 14 / 3 / 13 printed. The body has been edited to the 49-test numbers.

### Two survivors closed

The "the skip is NARROW" property was pinned at **1 of the 4 bounds the skip keeps** — only the icon minimum. Two mutants survived all 47 tests, and both are real behaviour changes rather than equivalent mutants:

- **skip widened to the icon maximum.** Not caught by the existing ceiling guard, because 4096 is *half* the 8192 kind-agnostic ceiling — a fixture that trips the icon maximum is comfortably under the ceiling and gets there with nothing having fired earlier. A 4097px WebP icon would have gone from refused-at-the-picker to uploaded-then-refused.
- **skip widened to the screenshot minimum shorter side.** Same shape, on the third kind.

Each now has an `expectRefused` case sibling to the icon-minimum guard, and each mutant was re-applied and watched to fail **its own** test, on `showErrorNotification` "Number of calls: 0" — i.e. no refusal was produced at all, which is exactly what the mutant does. Neither guard is regression coverage and both are labelled as such: they pass on unmodified code and exist only to kill their mutant.

### One re-implemented rule replaced with an executed one

`listing-asset-exif-fixture.test.ts` asserted the server's reading through a **local `displayed()` helper that re-stated `probeStoredImage`'s orientation 5–8 transposition** — a vendored copy of a bound, which is the anti-pattern this PR cites (civitai/cli#270): it agrees with itself whatever the real function does, so it could not have detected the two diverging.

That helper is gone. The rule is executed instead: `offsite-listing.service.test.ts` gains the **WebP sibling** of its existing quarter-turned-JPEG case, running the real `persistListingAssetImage` + `probeStoredImage` end to end (stored 450×800 + orientation 6 → measured 800×450, `{ok: true}` as a cover). Restricting that transposition to JPEG only fails the new test — and *only* it, with the JPEG sibling still green — on the untransposed pair. The fixture test keeps the part that is genuinely its own: that the fixture **bytes** really carry a readable `orientation: 6`, with the untouched-raster negative control.

Also fixed: the fixture header pointed at a test file name that does not exist.

PR #3840, opened 2026-08-12
fix(app-blocks): route the last five listing-ownership gates through the canonical owner

ZacxDev:
# Consolidate the listing-ownership gates onto the canonical owner

#3834 made `resolveListingAccess` resolve ownership canonically — `appBlock.app.userId ?? listing.userId` — because for an **onsite** listing the canonical owner is the `OauthClient`'s and `AppListing.userId` is a denormalized copy that can drift. Several gates were left comparing the caller against the copy.

On a drifted onsite listing such a gate is wrong in **both** directions at once: the real owner is refused (their collaborator fallback resolves them as `owner`, which is not `editor`), and whoever the stale row names walks straight through. Same inversion #3834 fixed in `listCollaborators` / `getPendingTransfer`.

## 🔴 Where the drift actually comes from — corrected

An earlier version of this PR body, and of several code comments, said the drift comes from `acceptTransfer`'s onsite listing write being "deliberately unguarded, a 0-count an accepted desync". **That was wrong**, and it pointed maintainers at the wrong file. It is corrected here and in every comment that repeated it.

`app-ownership-transfer.service.ts:465-470` writes `where: { id: transfer.appListingId }` for onsite — **unconditional**, in the same transaction as the `OauthClient` move at `:450`, and after an in-tx read of the same row through its own FK. That write **heals** the parent's copy; a 0-count there would require the row to be absent, which the pre-read precludes. And `:450` is the only in-app write to `OauthClient.userId` (the two `publish-request.service.ts` `oauthClient.update` calls touch `grants`/`allowedScopes` only). **A top-level listing row's copy has no drift mechanism at all.**

The reachable mechanism is the **SHADOW REVISION**:

- `beginListingRevision` clones the approved parent into a hidden draft with `userId: parent.userId` (offsite-listing.service.ts:1081) and `appBlockId: null`;
- `acceptTransfer` step 3 updates only `{ id: <the transferred listing> }` — never that listing's shadows — and `applyApprovedRevision` copies assets back to the parent, never `userId`;
- so a shadow that outlives a transfer keeps the OLD owner **frozen**, while its parent and the `OauthClient` both name the new one.

Four of the five gates below can be handed a shadow id and therefore see that frozen value: `loadOwnedEditableListing` (via `updateRevisionDraft`), `submitListingRevision` (which takes nothing else), `loadOwnedListing` (via `resolveOwnerAssetEditTarget`, which calls it with the shadow id directly), and `resolveOwnerScreenshotTarget`. The fifth, `getMyListingForApp`, resolves its row by `appBlockId`/`slug`, so it only ever sees a parent — it is routed through the same helper for **uniformity**, not because a stale copy can reach it. That is stated as such in the code.

**Measured state: latent, not live.** 0 drift across all 21 onsite listings in prod (2026-08-11); no onsite ownership transfer has ever run. It goes live on the first onsite transfer of a listing that has an in-flight revision draft.

## What changed

Five gates now ask `resolveListingAccess` for the whole answer (owner **and** seat), instead of comparing against the column and using the resolver only as a collaborator fallback:

| # | Site | Gate |
|---|---|---|
| 1 | `offsite-listing::loadOwnedEditableListing` | `updateListing`, `getMyListingForEdit`, `updateRevisionDraft`, `beginListingRevision` |
| 2 | `offsite-listing::submitListingRevision` | submit a shadow revision |
| 3 | `offsite-listing::getMyListingForApp` | the media editor's entry read |
| 4 | `app-listing-assets::loadOwnedListing` | every asset mutator + `getListingAssets` |
| 5 | `app-listing-assets::resolveOwnerScreenshotTarget` | **the fifth site, not in the original four** — its redundant early copy of the gate is deleted |

Both files now have one helper, `resolveListingRole` (a thin read of `resolveListingAccess`); no second implementation of ownership was written. The mod bypass in the assets file (D1) still short-circuits **first**, so a moderator pays no extra read — and that is now **asserted by call count**, not just claimed in a comment. `offsite-listing` still has **no** mod bypass (D1 preserved), and that too is now a behavioural test rather than a ledger string.

Also in this PR:

- **`hasWritableRepo` is now a type predicate** (`listing is { kind: string; blockSlug: string }`), so the three `blockSlug!` non-null assertions are gone.
  🔴 **Its justification is corrected.** The old comment claimed that relaxing the body to a kind-only check "is a type error at every call site instead". It is not — TypeScript **never** checks a user-defined predicate's body against the declared predicate. Measured with `pnpm typecheck` on this tree: the relaxed body emits **0** errors; the positive control (same relaxed body, return type demoted to `boolean`) emits exactly **3** `TS2322: Type 'string | null' is not assignable to type 'string'`, one per call site. So the predicate is checked at its **callers** and unchecked at its **definition**. The null-slug half is now pinned behaviourally by a new fixture — an ON-SITE listing with **no** backing block, the only shape that can kill a kind-only body (every other on-site fixture has a slug, so both clauses agree). Against that mutant the three new cases go red with `expected "vi.fn()" to not be called at all, but actually been called 1 times` and the recorded call is `{ slug: null, userId: 20 }`.
- **`app-access.service`'s `ownerUserId` doc** no longer claims the ledger "pins that every consumer agrees", and no longer names the wrong drift mechanism.
- **`resolveAccessibleAppBlockIds`'s `kind:'onsite'` comment now covers analytics**, not only money. **Decision: comment widened, predicate unchanged.** The set is a set of *AppBlock* ids and every analytics aggregate is `appBlockId IN thatSet` over block-scoped series; an offsite listing's analytics surface is `AppListingMetric`, which this set does not address. The only shape the clause removes is offsite-with-a-block (0 rows).
- 🟡 A known, accepted wart is written down at `getMyListingForApp`: it re-reads the same row on the same pool, so a listing deleted between the two reads reports `NOT_OWNED` rather than `NOT_FOUND`. Cosmetic — the caller is refused either way — and the alternative widens the shared resolver's return type for every caller.

## The ledger — and an honest statement of what it holds

`app-access.call-site-ledger.test.ts` gains **D5** and a third population: `DENORM_OWNER_HOLDOUTS`, with the same fails-on-GROWTH-**and**-SHRINK property. The denormalized-column class is **empty for every collaborator-reachable gate**; three non-collaborator sites remain and are recorded with their reasons (`offsite-moderation.service.ts` — same latent inversion, needs `resolveListingAccess` to accept a `Prisma.TransactionClient`; `app-listing-review.service.ts` — anti-abuse, belongs at the `listAppInsiderUserIds` seam; `publish-request.service.ts` — in the class by SHAPE, safe by CONTEXT).

🔴 **`DENORM_OWNER_RE` is a SPELLED guard, not a structural one, and the PR now says so instead of overstating it.** `app-access.service.ts` used to claim "it fails if a new gate compares against `AppListing.userId`". Measured over the whole non-test `src/` corpus, the original regex caught **2 of 11** realistic spellings of the exact D5 regression — restoring it as `if (userId !== listing.userId && (await resolveListingRole(…)) !== 'editor')` passed the entire ledger suite.

**Chosen remedy: (b) — downgrade the claim to exactly what it holds, plus the widening that is free.** Both regexes now accept either operand order, optional chaining and `!=`/`==`, which lifts coverage to **8 of 11** and changed **neither matched file set** (`GATES` stayed at 16 files; the class stayed at exactly the 3 holdouts), so it is strictly additive. Going further was **rejected with a measurement**: a pattern over an arbitrary receiver (`(?:\w+\.)*\w+\.userId <op>`) matches **127 files repo-wide and 15 under `src/server/services/blocks/` alone** — `image.userId`, `session.userId`, `client.userId`, `sub.userId` — which turns the enumerated-equality assertion into a permanently-noisy allowlist that decays on every unrelated edit, i.e. a gate people learn to click through. And it still would not see a hoisted local or a destructure.

So the ledger is documented as a **naming-convention lint**: it stops a reintroduction written in this feature's own idiom and keeps the holdout record from going stale. It is **not** a guarantee that no gate reads the column. A new `🔴 KNOWN EVASIONS` test pins the three spellings it cannot see (hoisted local, destructure, any other receiver name) as a measured fact, with a comment saying not to "fix" it by widening. The guard that actually holds the behaviour — for any spelling, because it never reads the source — is `app-access.denormalized-owner-drift.test.ts`.

## 🔴 Known regression on a currently-unmintable shape — issue #3844

`resolveListingAccess` is **block-first, not kind-branching** (`appBlock.app.userId ?? listing.userId`, no branch on `kind`). On an **offsite listing that carries a backing AppBlock**, the block decides ownership — and both off-site ownership writers move only the column (`acceptTransfer`'s step (2) is `if (isOnsite)`-guarded; `claimListing`, the mod impersonation remedy, refuses a non-offsite listing outright and writes only the column). So on that shape the ex-owner, or the impersonator a claim was meant to dispossess, keeps edit access and the rightful owner is refused.

**Pre-PR these five gates were correct on that shape.** This is a regression, not a neutral recorded disagreement, and it is now labelled as one in the code and here.

It is filed rather than fixed because it is currently **unreachable**, measured against prod 2026-08-12:

- `kind='offsite' AND app_block_id IS NOT NULL` → **0 rows**;
- **0 of 22** `app_blocks` carry an `external_url`, and a grep of `src/server` finds **no writer** of that column — every hit is a read, a projection or a test fixture. `mapAppBlockToListing` mints the shape only from an AppBlock that has one, so it is not merely 0-row, it is not mintable by any current code path.

Tracked as **https://github.com/civitai/civitai/issues/3844** with the suggested fix (branch on `kind`), referenced from the code, and pinned by the drift suite's last case so a resolver change surfaces deliberately.

## Verification

- **`app-access.denormalized-owner-drift.test.ts`** — one drifted onsite fixture, all five sites, both directions each, plus editor/stranger/mod controls, the D1 no-mod-bypass cases and an offsite control.
  **Red at `662a47f334`: 12 failed / 11 passed (23). Green at HEAD: 23 / 23.**
- **Mutation-verified with non-deletion mutants**, each killed by its own assertion:
  - `hasWritableRepo` body → kind-only: **3 failures**, all three the new no-block cases, message `expected "vi.fn()" to not be called at all, but actually been called 1 times`, recorded call `{ slug: null, userId: 20 }`. `pnpm typecheck` on the same mutant: **0 errors** (the finding above).
  - `loadOwnedEditableListing` gains an `isModerator` short-circuit (threaded from `updateListing`): **2 failures across the whole 2820-test blocks suite**, both the new D1 cases, `promise resolved "{ listingId: 'apl_live', …(3) }" instead of rejecting`. Nothing else in the suite saw it — which is exactly the gap this test closes.
  - assets `loadOwnedListing` mod bypass moved AFTER the resolve (behaviour-preserving): killed by the new call-count case, `expected "vi.fn()" to not be called at all, but actually been called 1 times`. Two collateral failures in `app-listing-assets.lazy-revision.service.test.ts` are fixture artefacts (`TypeError: Cannot read properties of undefined (reading 'findFirst')` — that suite mocks no `appCollaborator`), not a second guard on the property.
  - `DENORM_OWNER_RE` reverted to its pre-widening form: killed by the positive control, `DENORM_OWNER_RE must recognise: reversed operands`.
  - `DENORM_OWNER_RE` widened to an arbitrary receiver: killed by `🔴 KNOWN EVASIONS` with its own message, and the class assertion independently reported `expected [ …(127) ] to deeply equal [ …(3) ]` — corroborating the 127-file false-positive count quoted above.
- Blocks dir: **125 files / 2820 tests, all passing** (2810 before these fixes; +10).
- `pnpm typecheck`: **0 errors** in 78 s. Changed test files type-check separately (tsconfig excludes `__tests__`) with **0 errors in all five**; the 3 in `app-listing-assets.service.test.ts` are byte-identical at the base.
- eslint 8.57.1 over all 9 changed files (absolute paths, count read): **0 errors, 54 warnings**, every one the pre-existing `'_a' is defined but never used` mock-signature shape, none in a source file. Instrument validated against a known-bad probe (1 `Error/no-var`).
- prettier: the one **added** file is clean. Four modified files report style issues — **byte-identical set at `HEAD` before these edits**, and none of the diffs touch a line this work added, so they are left alone rather than blanket-reformatted.

### DIFF
--- a/src/server/services/blocks/__tests__/app-access.call-site-ledger.test.ts
+++ b/src/server/services/blocks/__tests__/app-access.call-site-ledger.test.ts
@@ -133,10 +133,14 @@ const GATE_LEDGER: Record<string, string> = {
     'capability table. Every other site delegates here. Seats are keyed to AppListing ' +
     'so an offsite listing can hold collaborators. 🔴 BOTH resolvers report the CANONICAL ' +
     'owner (`appBlock.app.userId ?? listing.userId`), never the denormalized ' +
-    'AppListing.userId alone: acceptTransfer’s onsite step 3 is unguarded and treats a ' +
-    '0-count as an accepted desync, so the copy can be stale — and a stale copy inverts ' +
+    'AppListing.userId alone: a SHADOW REVISION clones that column and no ownership ' +
+    'write ever revisits the clone, so the copy can be stale — and a stale copy inverts ' +
     'every gate that delegates here, refusing the real owner while admitting the name ' +
-    'the row happens to carry.',
+    'the row happens to carry. (NOT acceptTransfer’s onsite listing write, which is ' +
+    'unconditional and heals the parent; see app-access.denormalized-owner-drift.test.ts ' +
+    'for the mechanism in full.) 🔴 The resolution is BLOCK-FIRST, not kind-branching, ' +
+    'and on an offsite-listing-that-carries-a-block it names the wrong owner after a ' +
+    'transfer or a mod claim — 0 rows and unmintable today, tracked as issue #3844.',
   'src/server/services/blocks/app-collaborator.service.ts':
     'assertOwner — seat management is OWNER-ONLY by product decision (an editor is a ' +
     'co-owner in every respect EXCEPT managing collaborators and initiating a ' +
@@ -276,9 +280,17 @@ const KIND_CAPABILITY_LEDGER: Record<string, string> = {
  * a fresh copy of the same inversion. Every shape is probed in the POSITIVE CONTROL test,
  * because a regex nobody has watched match is a claim, not a guard — and that matters
  * more, not less, now that the live corpus no longer exercises them.
+ *
+ * 🔴 AND IT IS A SPELLING GUARD, NOT A STRUCTURAL ONE. Read {@link DENORM_OWNER_RE}'s
+ * KNOWN EVASIONS block before quoting either regex as protection: both are anchored on
+ * the RECEIVER NAME (`*listing` / `*shadow` / `<x>.appListing`), so a gate written
+ * against a differently-named local evades them entirely. The listing-owner alternatives
+ * here accept either operand order, optional chaining and `!=` — but they are a
+ * naming-convention lint over this feature's own conventions, and the guard that
+ * actually holds the BEHAVIOUR is `app-access.denormalized-owner-drift.test.ts`.
  */
 const GATE_RE =
-  /app\??\.userId|app:\s*\{\s*userId|(?:\w*[Ll]isting|\w*[Ss]hadow)\.userId\s*!==|\w+\.appListing\.userId\s*!==|resolveAppAccess|resolveListingAccess|resolveAccessibleAppBlockIds|assertAppEditAccess|listAppInsiderUserIds|loadOwnedListingInTx/;
+  /app\??\.userId|app:\s*\{\s*userId|(?:\w*[Ll]isting|\w*[Ss]hadow)\??\.userId\s*(?:!==|!=)|(?:!==|!=)\s*(?:\w*[Ll]isting|\w*[Ss]hadow)\??\.userId|\w+\??\.appListing\??\.userId\s*(?:!==|!=)|(?:!==|!=)\s*\w+\??\.appListing\??\.userId|resolveAppAccess|resolveListingAccess|resolveAccessibleAppBlockIds|assertAppEditAccess|listAppInsiderUserIds|loadOwnedListingInTx/;
 
 /**
  * 🔴 THE DENORMALIZED-OWNER GATE CLASS (D5) — every spelling of "compare the caller
@@ -289,9 +301,38 @@ const GATE_RE =
  * exclusion is) and that form is just as stale-able. Kept as a separate regex rather than
  * widened into `GATE_RE` so that the gate POPULATION and the column-read CLASS can be
  * asserted independently — a site can legitimately be in one and not the other.
+ *
+ * 🔴 KNOWN EVASIONS — READ THIS BEFORE CITING THIS REGEX AS A GUARD. It is a SPELLED
+ * check, not a structural one: it is anchored on the RECEIVER NAME. Measured over the
+ * whole non-test `src/` corpus, of eleven realistic spellings of the exact D5 regression
+ * it catches eight and MISSES three, and the three it misses are ordinary code:
+ *
+ *   - a HOISTED LOCAL  — `const ownerId = listing.userId; if (ownerId !== userId) …`
+ *   - a DESTRUCTURE    — `const { userId: ownerId } = listing; if (ownerId !== …)`
+ *   - ANY OTHER NAME   — `row.userId !== userId`, `parent.…`, `draft.…`
+ *
+ * The three newly-covered spellings (reversed operands, `listing?.userId`, `!=`) were
+ * added because they cost nothing: they changed NEITHER matched file set — `GATES` stayed
+ * at 16 files and this class stayed at exactly the three holdouts below — so the widening
+ * is strictly additive. Going further is not free and was REJECTED with a measurement: a
+ * pattern over an ARBITRARY receiver (`(?:\w+\.)*\w+\.userId <op>`) matches **127 files
+ * repo-wide and 15 under `src/server/services/blocks/` alone**, almost all of them
+ * correct and unrelated (`image.userId`, `session.userId`, `client.userId`,
+ * `sub.userId`, `metadata.userId`). As an enumerated-equality assertion that is a
+ * permanently-noisy allowlist that decays on every unrelated edit — a gate people learn
+ * to click through — and it STILL would not see the hoisted-local or destructured forms.
+ *
+ * 🔴 SO THE HONEST STATEMENT OF WHAT THIS HOLDS: it is a NAMING-CONVENTION LINT. It keeps
+ * a gate written in this feature's own idiom from being reintroduced silently, and it
+ * keeps the holdout record from going stale. It is NOT a guarantee that no gate compares
+ * against `AppListing.userId`. The guard that actually holds the BEHAVIOUR — for any
+ * spelling, because it never reads the source at all — is
+ * `app-access.denormalized-owner-drift.test.ts`, which drives all five sites against a
+ * drifted fixture and asserts both directions. If you are deciding whether this class is
+ * safe, that suite is the evidence; this one is bookkeeping.
  */
 const DENORM_OWNER_RE =
-  /(?:\w*[Ll]isting|\w*[Ss]hadow)\.userId\s*(?:!==|===)|\w+\.appListing\.userId\s*(?:!==|===)/;
+  /(?:\w*[Ll]isting|\w*[Ss]hadow)\??\.userId\s*(?:!==|===|!=|==)|(?:!==|===|!=|==)\s*(?:\w*[Ll]isting|\w*[Ss]hadow)\??\.userId|\w+\??\.appListing\??\.userId\s*(?:!==|===|!=|==)|(?:!==|===|!=|==)\s*\w+\??\.appListing\??\.userId/;
 
 /**
  * 🔴 THE CLASS IS NOT DELETED, IT IS EMPTY-EXCEPT-FOR-THESE — and this table is the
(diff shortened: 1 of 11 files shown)


ZacxDev:
## Mutation results — 12 non-deletion mutants, each killed by its own site's error

Method: rewrite the guard (never delete it), run the whole blocks dir (125 files / 2810 tests), record which assertion fired and with which message. Baseline green = `125 files / 2810 tests`. Every mutant reached the guard on an input no earlier check rejects (the drifted fixture passes the NOT_FOUND / status / floor checks first).

| # | Mutant | Failures | Killed by (attributing assertion → error) |
|---|---|---|---|
| 1 | `edit_restore` — restore the pre-fix `listing.userId !== userId && role !== 'editor'` | 4 | drift · site 1 *ADMITS the real owner* → rejected `OffsiteRequestError: you can only edit your own listings`; *REFUSES the stale name* → resolved instead of rejecting. **Plus both D5 ledger assertions**, one naming the file: `app-listing-assets… resolves a seat canonically but still compares against AppListing.userId` |
| 2 | `edit_invert` — `=== null` → `!== null` | 75 | drift · site 1, all four cases → `you can only edit your own listings` |
| 3 | `edit_inert` — resolve into a variable, compare it to `undefined` (guard present, never fires) | 16 | drift · site 1 *REFUSES the stale name* → `promise resolved "{ listingId: 'apl_live', … }" instead of rejecting` |
| 4 | `submit_swap` — swapped operand: `resolveListingRole(shadowId, shadow.userId)` | 3 | drift · site 2 *ADMITS the real owner* → `you can only submit your own revision` |
| 5 | `submit_rebind` — stale rebind: resolve `shadow.id + '_x'` | 2 | drift · site 2 only, same message. (Nothing else in the repo noticed — this is the case the new suite exists for.) |
| 6 | `submit_invert` | 13 | drift · site 2 all three cases → `you can only submit your own revision` |
| 7 | `app_rebind` — resolve `appBlockId` instead of `listing.id` | 2 | drift · site 3 → `you can only manage your own listings` |
| 8 | `app_inert` | 8 | drift · site 3 *REFUSES the stale name* → resolved instead of rejecting |
| 9 | `assets_restore` — restore `listing.userId !== user.id && !user.isModerator` as the outer condition | 5 | drift · site 4 **and site 5** *REFUSES the stale name* → resolved instead of rejecting; + both D5 ledger assertions. (One-sided by construction: this mutant admits the real owner via the role fallback and *also* admits the stale name via the short-circuit, so only the refusal half can kill it — which it does, at both entry points.) |
| 10 | `assets_mod_invert` — `!user.isModerator` → `user.isModerator` (gate runs for mods only) | 30 | drift · sites 4+5 refusals → `TRPCError: You do not own this listing` expected but resolved; + `app-collaborator.editor-read-after-write` stranger case |
| 11 | `assets_owner_only` — `role === null` → `role !== 'owner'` (editors refused) | 5 | drift · site 4 editor case → rejected `You do not own this listing`; + permission-matrix `assets.getListingAssets · accepted editor → ALLOW` (both kinds) |
| 12 | `assets_pool_drop` — drop the `db` override on the resolve | 4 | **not** the new suite — `app-collaborator.editor-read-after-write` + `app-listing-assets.lazy-revision` (the pre-existing pool seam). Stated plainly: the drift fixture answers on both pools, so it is blind to the pool dimension by construction; the pre-existing seam suite is what covers it and it survived the refactor. |
| 13 | `repo_slug_only` — drop the kind clause from `hasWritableRepo` | 3 | `app-collaborator.service` · *an OFF-SITE listing that HAS a repo slug still reaches Forgejo NEVER* (accept/remove/leave) → `expected "vi.fn()" to not be called at all, but actually been called 1 times` |

**Type-predicate negative control** (the change has no runtime behaviour, so a mutant cannot be observed by tests): reverting `hasWritableRepo`'s return type from `listing is { kind: string; blockSlug: string }` to `boolean`, leaving the call sites as written, produces exactly 3 errors —
```
app-collaborator.service.ts(522,31): error TS2322: Type 'string | null' is not assignable to type 'string'.
app-collaborator.service.ts(563,32): error TS2322: …
app-collaborator.service.ts(594,32): error TS2322: …
```
i.e. the compiler now holds the invariant the comment used to.

**Instruments validated before their verdicts were read:**
- the drift suite was watched RED at `662a47f334` (11 failed / 7 passed) before being read green at HEAD (18/18);
- `DENORM_OWNER_RE` has a POSITIVE CONTROL asserting it matches all six removed spellings — a regex claiming a class is *empty* is worthless until it has been watched to match that class;
- the separate test-file typecheck (tsconfig excludes `__tests__`) was negative-controlled with a planted `const _neg: number = 'x'` → `TS2322` reported. Its 3 remaining errors in `app-listing-assets.service.test.ts` are **pre-existing** — identical at `662a47f334` (lines 89 / 1241, untouched by this PR).

PR #3851, opened 2026-08-12
fix(monetization): disclose the version fee UI in steps, and seed no fee behind it

JustMaier:
@
Fixes the uploader feedback in [CU 868kq69rv](https://app.clickup.com/t/868kq69rv) (reported by Koen).

## The problem

The **License Fee** editor rendered whenever the `licensingFee` flag was on, pre-filled with the type-based suggested fee. So a creator who never asked to charge anything saw a price they had no context for — and then hit `You must confirm you hold the rights to monetize this model` at save, with the only ways out being to tick an affirmation they did not need or to type `0`.

## What changed

Monetization now discloses in three steps:

1. `I want to charge for this version` (master switch, off unless the version already charges)
2. the rights affirmation
3. the pricing controls — Paid Access setup and the License Fee editor

The suggested fee is applied when the fee editor **opens**, not when the form loads. Closing either step clears what it revealed (`clearCharges`): `shouldUnregister` is false on this form, so anything left behind a collapsed control is still submitted, and a fee behind a hidden editor is both an unasked-for charge and a save the creator cannot unblock.

The submit gate is deliberately unchanged — `requiresRightsAffirmation` stays keyed to an *actual* charge, so toggling the switch and changing your mind never blocks a save. An ended Early Access window is exempt from clearing: that config is locked, and the master switch must not remove what the disabled paid-access switch refuses to.

The inner paid-access switch is relabelled `Charge for access to this version` so it does not read as a duplicate of the master switch.

## Verification

- `pnpm run lint` — no new errors (2 new `react-hooks/exhaustive-deps` warnings, same intentional-narrow-deps pattern as the four already in this file)
- `pnpm run test:unit:run` — 14,928 passed; the 6 failing files are the pre-existing app-blocks convention scans, untouched here
- 5 new browser tests in `ModelVersionUpsertForm.browser.test.tsx`, mutation-tested: restoring the old always-visible seeded fee fails 4 of them (including `expected "vi.fn()" to be called 1 times, but got 0 times` — the save Koen could not complete), and dropping the clear-on-close fails the stale-fee assertion. Both fail in under a second.

@

### DIFF
--- a/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
+++ b/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
@@ -52,6 +52,9 @@ vi.mock('~/components/UserSettings/hooks', () => ({
   useCurrentUserSettings: () => ({ hideDonationGoals: false }),
   useMutateUserSettings: () => ({ mutate: vi.fn(), isPending: false }),
 }));
+// The paid-access section opens with a DismissibleAlert, whose `useIsClient()` throws rather than
+// defaulting when the provider isn't mounted — unmocked it takes the whole render down.
+vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
 // The description field mounts the whole rich-text editor (tiptap + sticker cosmetics queries), none
 // of which this test drives.
 vi.mock('~/components/RichTextEditor/RichTextEditorComponent', () => ({
@@ -114,9 +117,18 @@ const feeInput = () => page.getByLabelText('Licensing fee (Buzz)');
 
 // A version that already charges, with an affirmation on record — so the disclosure opens straight onto
 // the pricing controls, which is how a creator who already priced this version meets the form.
+// Published with a permanent gate: that is what makes the paid-access section configurable (and so the
+// early-access-loss warning reachable) without an early-access score.
 const chargingVersion = {
   ...(version as object),
+  status: 'Published',
   licensingFee: 2,
+  licensingFeeSettlementCurrency: 'Buzz',
+  paidAccess: {
+    endsAt: null,
+    timeframeDays: null,
+    terms: { download: { price: 5000 } },
+  },
   meta: {
     rightsAffirmation: {
       userId: 1,
@@ -128,6 +140,7 @@ const chargingVersion = {
 } as unknown as React.ComponentProps<typeof ModelVersionUpsertForm>['version'];
 
 function renderChargingForm() {
+  flags.current = { licensingFee: true, earlyAccessModel: true };
   renderWithProviders(
     <ModelVersionUpsertForm model={model} version={chargingVersion} onSubmit={vi.fn()}>
       {() => <button type="submit">Save</button>}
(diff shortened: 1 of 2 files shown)


JustMaier:
@
Follow-up notes from an adversarial review pass over the diff (3 rounds). All three items below are **pre-existing**, not introduced here, and left for separate tickets:

1. **A stored gate is dropped on any save while `features.earlyAccessModel` is off.** `defaultValues.paidAccessConfig` seeds `null` when the flag is off, while the version still has its `paidAccess` row — so `!isEqual(...)` is true, the mutation runs, and `toPaidAccessInput(null, …)` removes the gate on a save where the creator touched nothing. In practice the flag is `[public]`, so this is unreachable today; it becomes real for anyone who turns it off. The new removal warning does now fire in that state, which is a true statement about what saving would do.
2. **`model.poi` is excluded in the UI but never re-checked in `handleSubmit`**, which substitutes `null` only for private models. A POI model with a stored gate resubmits it from a control that never rendered.
3. **The client checks the rights affirmation without `ownerId`.** The server passes `ownerId: model.userId` and requires the affirmation to belong to the current owner, so a model transferred after its affirmation was recorded reads as affirmed on the client, renders no checkbox, and fails server-side with nothing on screen to tick. The fix needs `userId` on the version-for-edit selector (its `model` select is `{name, type, nsfw, poi, mode}`), which is server scope beyond this ticket.

Also worth knowing for anyone writing component tests against this form: `useIsClient()` throws `missing IsClientContext` rather than defaulting, and the `DismissibleAlert` in the paid-access section calls it. Unmocked, it takes the whole render down to an empty container with no assertion to read.
@

JustMaier:
Adversarial review is clean as of the final commit (5 rounds; 8 findings fixed, each mutation-tested). Two residuals recorded rather than fixed, neither reachable today:

1. `gateRemovalIsStructural` enumerates two reasons a gate can't be kept (private model, non-gateable usage control) where the underlying question is "would restoring the stored config make `gateCharges` true?". A stored free-generation-only gate with no download tier escapes the enumeration and would offer an inert restore link — but `assertPaidAccessInput` rejects that shape, so no such row can be created through the app. The robust spelling, if it ever needs closing, is to run the *stored* config through `toPaidAccessInput` with the private substitution and check `paidAccessCharges` on the result.
2. Restoring a stored charge marks the form dirty even when the round-trip lands back on the stored values, so the save takes the mutation branch rather than the pristine short-circuit. React Hook Form can't un-dirty a field without `reset`, and any genuine edit resubmits the same config anyway.

JustMaier:
Reviewed live with Justin on a dev server and reworked the layout in 9380cdb1c9:

- **One card per way to charge**, each with its title on the left of a tinted header strip and its own switch on the right; the body drops in when you flip it.
- **The generation fee has its own opt-in** beside paid access, rather than appearing whenever the monetization section was open. Closing it clears the fee, the same way closing the access switch clears the gate.
- **Access mode and the early-access window moved into the card body**, above pricing, so the mode and the window it configures stay together.
- **Typographic hierarchy fixed** — the inner field labels were bold at default size and outsized the card titles above them.
- **The long License Fee description** is now `Fee per generation` plus an info popover (what you earn, that a derivative's fee stacks on the base model's, your per-generation cap).

Superseded copy removed rather than lost: the blue alert above the access toggle, the two "… Set Up" dividers, and the inner Access/Pricing section headings.

10 browser tests (one new: closing the fee switch clears the fee it revealed, mutation-tested). Typecheck clean.

PR #3855, opened 2026-08-12
feat(home): fill Featured Images from the featured collections pool

JustMaier:
Featured Images is hand-curated, so the section goes quiet whenever nobody is at a keyboard. The collections already in the featured pool are curated continuously by several people with different taste. This job takes the top new items from them and tops the section up. Manual adds are untouched.

## How it picks

Candidates are ACCEPTED image items curated into an **eligible** featured collection inside a 7-day window — "new" meaning newly *curated*, not newly uploaded, which is what covers a weekend. Eligibility is the state the existing hourly job already computes, so a collection that goes quiet, gets renamed, or has its `write` changed drops out for free.

Score is `reactions / (hours_since_curated + 12) ^ 0.8`.

**Ranking every candidate together does not work.** Measured against prod (11 collections, 719 usable candidates over 7 days): the top 40 gives the busiest collection **17 of 40** slots and one creator **8 of 40**. Volume becomes taste. So the job ranks *within* each collection and round-robins between them, with the rotation offset advancing per run. Same candidates, simulated over a day at 5-per-6h: **20 images, all 10 active collections, 17 distinct creators**.

`strategy: 'global'` is still selectable so that trade can be re-tested without a deploy. A test pins the domination behaviour so choosing it is deliberate.

## Cost

The whole candidate query, including a per-image `ImageReaction` count, is **~450 ms** on the replica (index-only scan on `ImageReaction_imageId_userId_reaction_key`). At 4 runs/day that is free, and it means no dependency on search-index freshness.

## Composing with the per-creator cap (#3843)

That cap is a **render-time** cap, so this can't break it — but it can waste slots: 5 auto-adds from one creator render as 2 and 3 holes. The job therefore caps at write time as well (1 per creator per run, 2 live in the window). The render cap still has the last word.

## Removal is a tombstone

The dedupe is "does a row already exist for this image", so deleting an auto-added item would let the next run put it straight back. `removeCollectionItem` now sets those rows to `REJECTED` instead — invisible to the render path, which defaults to ACCEPTED only — and still deletes everything else. Auto rows are identified by `addedById` = CivitaiOfficial plus a `auto-featured:<sourceCollectionId>` note, which also records which curator's collection each featured image came from.

## Turning it off

Three levels: the Flipt flag (`auto-feature-images`, **default-off**), the job disable, and a full undo — every auto row is identifiable, so one scoped statement removes everything the job ever added and leaves hand-picked items alone.

## Configuration

Everything lives in the FeaturedCollections block's `autoFeature` metadata, so it is a Retool edit rather than a release: `dryRun` (**defaults on**), `perRun`, `intervalHours`, `windowDays`, `recencyOffsetHours`, `decayExponent`, `maxPerCreatorPerRun`, `maxPerCreatorInWindow`, `maxPerCollectionInWindow`, `minReactions`, `strategy`. The job's cron is hourly and it reads its own cadence, so changing how often it fires is not a deploy.

`dryRun` defaults on so the first runs only log what they *would* have featured — Ellie asked to see this in practice, and this shows her without the homepage changing.

## Not done here

Nothing is enabled by this PR. Going live needs, in order: the `autoFeature` config written onto the FeaturedCollections block (including `collectionId`), the Flipt flag turned on, a look at the dry-run output, then `dryRun: false`.

## Testing

`pnpm run typecheck` clean, eslint/prettier clean on changed files, `pnpm run test:unit:run` introduces **no new failures** — the 10 failing files on this branch also fail on `origin/main` (main is currently worse at 12). `blocks.router.workflow.test.ts` collected 347 tests, so the run is real.

13 new tests cover the caps, the rotation, the decay, the reaction floor, and the reject-not-delete removal. One of them caught a real bug: the rotation under-filled a run when a collection went dry.


JustMaier:
## Adversarial review pass — six defects, fixed in 2f32022

An independent review agent was run over the diff, prompted to refute rather than approve. It found five real defects plus two nits. Worth recording the biggest one, because a green test suite was perfectly happy with it:

**The window filter could never have run.** `make_interval(days => $1)` fails function resolution — Prisma binds JS numbers as int8, the argument is int4, and that cast is assignment-only. Verified against the replica:

```
Query error: function make_interval(days => bigint) does not exist
```

Every run would have thrown before reading a row, in dry run too. The flag would read on, the homepage would never change, and the only trace would be `jobErrorsCounter`. Two other jobs in this repo already work around exactly this (`purge-replaced-files.ts`, `process-scheduled-publishing.ts`).

**The tombstone had two more doors.** Removing an already-REJECTED item fell through to the unconditional DELETE, and `saveItemInCollections` — the "Save to collection" modal, and the more common removal UI — deleted rows outright without consulting the new path at all. Either one hands the image back to the next run, which is the entire thing the rejection exists to prevent.

**Rejected rows counted toward the creator's window quota**, so rejecting a bad pick suppressed that creator's good ones for the rest of the week.

**The guard predicate had no real coverage.** The mocked row counts prove which statement ran, never which rows it would match — deleting the `addedById`/`note` markers from the UPDATE left every test green while silently turning every removal site-wide into a tombstone. The tests now assert on the SQL text.

Every new assertion is mutation-tested. Worth noting the first attempt at the pick-uniqueness test was itself vacuous: it passed with the cursor bookkeeping deleted entirely, because the per-creator cap was doing the deduping (a repeated image is also a repeated creator). It only tests what it claims with the caps raised out of the way.

### Correction to the PR description

It credits `skipDuplicates` as the concurrency backstop. `CollectionItem_image_idx` exists in prod but in no migration in this repo, so it was applied out of band — local and preview don't have it and behave differently on exactly that guard. The real dedupe is the `NOT EXISTS` plus the job runner's per-job Redis lock; `skipDuplicates` is a bonus where the index happens to exist.

### State

`autoFeature` config is written on the FeaturedCollections home block in prod with `dryRun: true`. The Flipt flag is still off, so nothing runs.

JustMaier:
## Review passes 2 and 3 — one self-inflicted regression, one coverage hole

Two more adversarial passes over the fix commits. Both found something, and the first fix was worse than the bug it closed.

### The DELETE guard was NULL-unsafe (fixed in 950110826d)

`NOT ("addedById" = X AND note LIKE Y)` evaluates to **NULL, not TRUE**, whenever `note` is NULL — the ordinary case, not the exception. The DELETE matched zero rows and the caller still returned success. Verified on the replica:

```
adder_null_note_null: null
official_note_null:   null
normal_note_null:     true
```

Blast radius, measured: **10,000+ rows across 3,946 distinct collections** where `addedById IS NULL AND note IS NULL` — an item added by a since-deleted account — every one silently undeletable through the item menu, with a success toast. Collection 107 has none of those rows, so testing this feature would never have surfaced it.

Now read the row and decide in TypeScript. Nullable columns behave the way the code reads, and it matches the check `saveItemInCollections` already did, so the two removal doors can't disagree about the same row.

**The lesson for the tests:** asserting on SQL *text* proves a clause is present, never that it is correct. The `refuses to delete an auto-featured row` test was green on SQL that deleted nothing for 10k rows.

### Acting on one row where the old statement acted on all (fixed in de8fb96453)

The rewrite read one row and acted on its id; the DELETE it replaced removed every row matching `(collectionId, entityId)`. Prod has a partial unique index for all four reachable entity types, so they're identical there — but none of those indexes appears in any of the 706 migrations, so a database built from migrations can hold duplicates the new code would leave behind.

### The lookup's WHERE clause was untested (fixed in 6a2c23fccd)

Swapping raw SQL for Prisma calls made the tests more readable and **strictly less protective on the highest-consequence line**: asserting on call arguments sits downstream of the predicate that selects them. With the fixture returning a constant, dropping the entity predicate made the lookup return the whole collection and the delete empty it, and dropping the collection predicate removed the entity site-wide — both left every assertion green.

The fixture now renders the statement it was handed and asserts both predicates with their parameters. Each mutation fails 7 of 12 tests.

### Known issue, needs a product decision before this goes live

Re-adding a tombstoned image silently succeeds and does nothing: the INSERT hits `ON CONFLICT ... DO UPDATE SET "tagId"`, status stays REJECTED, the caller returns `'added'` and the UI shows success. The image never appears and nothing says why. This predates the PR but the PR adds a second, more commonly used door that creates tombstones. `updateCollectionItemsStatus` can flip the item back, but only from the collection review screen, not from the image page where the re-add just failed.

The silent success is a bug under either policy, so it wants fixing regardless of which way the policy goes.

### Verification

Typecheck 0 errors. Full unit suite: 6 failing files, **all 6 also fail on `origin/main`**, branch-only failures: none. `blocks.router.workflow.test.ts` collected 347 tests.

PR #3859, opened 2026-08-12
feat(generation): remix means edit and animate, not just prompt reuse

JustMaier:
## What

The Remix button only appeared when we happened to know an image's prompt **and** its resources, and what it offered there was reusing them. That is what has been teaching people the word: a remix is when you re-run someone's settings.

That definition now leaks into the remix gallery, where people pay to submit a remix and the owner decides what counts. Tester feedback ("people probably won't post literal remixes", "no criteria for remix") is users reading the word the way the product defines it.

This replaces the button with a menu of what a remix can actually be:

- **Edit image** — Nano Banana 2 Light
- **Animate** — LTX 2.3
- **Reuse prompt & resources** — the old path, when we have the data, named honestly

Edit and animate need only the image itself, so the button now shows on images with **no prompt and no resources** — the case where it used to be absent entirely.

## Surfaces

Image detail page, infinite feed card, and the home/profile card.

`RemixButton` was overloaded: media cards used it as a remix entry-point, `ModelCard` as a create entry-point. The media path moved to `CardRemixButton`; `RemixButton` keeps only the "Create" path it was really serving, which drops its type-discrimination branch.

## Refusals

Blocked, minor and POI sources **show** the button and refuse with a reason, rather than hiding it — a hidden button teaches nothing. The refusal set mirrors what the remix gallery already refuses on submission.

Note: there is no "owner disabled remixing" setting anywhere in the codebase, so that gate is not implemented — it does not exist to honour.

## Engine config

Both engines live in `src/shared/constants/remix.constants.ts`. Changing one is a one-line edit; a DB-backed config was considered and rejected because it would put a fetch on the hottest path in the app (every feed card) to avoid a deploy for something that changes rarely. The shape leaves room for an override later.

Each entry pins a **checkpoint version**, not just an ecosystem key. This is load-bearing: `NanoBanana` derives its mode from the selected version and falls back to `standard`, so an entry naming only the ecosystem would silently land on plain Nano Banana instead of 2 Light. LTX currently defaults to 2.3 on its own, but that default is incidental and would follow a future bump to 2.5.

`ltxVersionIds.v23Dev` moved into the `version-ids.ts` leaf module so the constant can reference it without pulling the LTX graph into card bundles.

## Provenance: deliberately not done

Remixes carry no link back to their source, and `useRemixOfId`'s prompt-similarity rule is untouched. Rationale: an image edit or an image-to-video shares no prompt with its source, so a similarity gate drops the link exactly when the remix is most genuine — but the fix isn't worth building either, because an off-site remix (download, edit elsewhere, upload) could never carry a link, so the gallery can't use one as a gate without excluding legitimate submissions. The owner's review remains the criterion.

## Testing

- `pnpm run lint` on changed files — 0 errors (2 pre-existing unused-import warnings in `ImageDetail2`)
- `pnpm run test:unit:run` — 16 failures across 6 files, **identical to the count on clean `main`** (all six are source-scanning meta-tests whose own positive controls fail there too). `blocks.router.workflow.test.ts` collected 347 tests, confirming the worktree submodule is intact.
- New test pins that each configured engine's workflow/ecosystem pair is actually valid — mutation-checked (an invalid pair fails in ~1s with a clear assertion, not a hang). It deliberately does **not** pin engine identity, so changing the default engine stays a one-line edit.

## Not covered

Post-publish "submit to this creator's remix gallery" — agreed as a follow-up. It has to be post-publish because `createRemixGallerySubmission` refuses anything unpublished or unscanned.

### DIFF
--- a/src/components/Cards/ImageCard.tsx
+++ b/src/components/Cards/ImageCard.tsx
@@ -6,7 +6,7 @@ import type { ImagesInfiniteModel } from '~/server/services/image.service';
 import { ImageMetaPopover2 } from '~/components/Image/Meta/ImageMetaPopover';
 import { DurationBadge } from '~/components/DurationBadge/DurationBadge';
 import { AspectRatioImageCard } from '~/components/CardTemplates/AspectRatioImageCard';
-import { RemixButton } from '~/components/Cards/components/RemixButton';
+import { CardRemixButton } from '~/components/Image/Remix/CardRemixButton';
 import { CardStickerOverlay } from '~/components/Sticker/CardStickerOverlay';
 import { useFeatureFlags } from '~/providers/FeatureFlagsProvider';
 import { StickerPlacementCardBadge } from '~/components/Sticker/StickerPlacementCardBadge';
@@ -39,7 +39,7 @@ export function ImageCard({ data }: Props) {
           )}
           <div className="ml-auto flex flex-col gap-2">
             <ImageContextMenu image={data} />
-            <RemixButton type={data.type} id={data.id} canGenerate={data.hasMeta} />
+            <CardRemixButton image={data} />
           </div>
         </div>
       }
--- a/src/components/Cards/ModelCard.tsx
+++ b/src/components/Cards/ModelCard.tsx
@@ -186,7 +186,7 @@ function ModelCardContent({ data }: Props) {
           </div>
           <div className="flex flex-col items-center gap-2">
             <ModelCardContextMenu data={data} />
-            <RemixButton type="modelVersion" id={data.version.id} canGenerate={data.canGenerate} />
+            <RemixButton id={data.version.id} canGenerate={data.canGenerate} />
 
             <CivitaiLinkManageButton
               modelId={data.id}
--- a/src/components/Cards/components/HoverActionButton.tsx
+++ b/src/components/Cards/components/HoverActionButton.tsx
@@ -20,6 +20,7 @@ const HoverActionButton = React.forwardRef<HTMLButtonElement, Props>(function Ho
     onClick,
     keepIconOnHover = false,
     style,
+    className,
     ...props
   },
   ref
@@ -36,8 +37,16 @@ const HoverActionButton = React.forwardRef<HTMLButtonElement, Props>(function Ho
         '--size': `${size}px`,
       }}
       onClick={onClick}
-      className={clsx(classes.wrapper, isCustomVariant ? classes[colorCustomVariant] : undefined)}
       {...props}
+      // After the spread, and merged rather than replaced: as a Mantine
+      // `Menu.Target` this component is cloned with a `className` of the
+      // popover's own, which would otherwise drop `wrapper` and take the icon
+      // layout and hover-expand animation with it.
+      className={clsx(
+        classes.wrapper,
+        isCustomVariant ? classes[colorCustomVariant] : undefined,
+        className
+      )}
     >
       <Badge className={classes.label} size="xs" variant={variant} color={color}>
         {label}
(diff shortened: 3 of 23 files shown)


JustMaier:
## Update: adversarial review round, 2 blocking defects fixed (`7d48059634`)

I ran an adversarial reviewer over my own diff across three rounds, prompted to refute rather than approve. It found two blocking defects in the first commit. Both were real, both verified against source before fixing, and the third round came back clean.

### 1. An unscanned image was sendable to a third-party edit engine

`minor` and `poi` are written by the **ingestion classifiers**, so on an image that has not been scanned they are `null` and every refusal below them passed vacuously. Both feeds show a user their own uploads before the scan finishes — the DB path renders the "Not published" badge for exactly those, and the search path returns `nsfwLevel === 0` docs **when `isOwnContent`** (`images.feed.ts:952`). So this wasn't an edge case, it was "I just uploaded this, let me remix it".

Worth noting my *first* fix for this was also wrong, and the second round caught it: I gated on `ingestion`, but the search backend's document type has no `ingestion` field at all, so the guard was inert on precisely the path that serves unscanned images to their owner.

The fix is `nsfwLevel === 0` — the only unscanned marker present on **both** backends, and the feed's own definition. `nsfwLevel` is now **required** on `RemixSourceImage`, so a surface that omits it fails typecheck rather than silently taking the unsafe branch. The `ingestion` clause stays as a DB-path refinement for a scan that ran and failed (`Error`/`NotFound`), narrowed so `Rescan` and `PendingManualAssignment` keep their earlier verdict — a re-ingestion sweep can put a large slice of the catalogue into `Rescan` at once.

Prompt reuse is deliberately **not** refused while a scan is pending: it copies text and resource ids already rendered on the page, so no image leaves the site.

### 2. The onboarding tour dead-ended

Mantine menus default to z-index 300; Joyride's overlay is 100000. The content-gen remix step sets `hideFooter`, `disableCloseOnEsc` and `disableOverlayClose` — clicking the target is the *only* way forward. The menu rendered under the overlay, so any new user running onboarding would have been stuck on that step permanently.

Fixed by clearing the overlay while a tour runs. A second-round finding then showed the same wedge returning through the *other* fix: a menu holding nothing but a disabled refusal has no item to click, so the action callback never fired. It now fires on open in that case, latched so reopening doesn't advance a tour twice.

### Also fixed

- **`loading` is no longer written from outside the generation store.** It was incoherent there: the store's bail-out clears the flag when its own `inFlightFetchCount` reaches zero, and a fetch it cannot see is not counted — so an unrelated superseded `open()` cleared a spinner whose fetch was still running, leaving the previous session's form live and **submittable**. That's a Buzz spend on the wrong generation. `withExternalFetch` participates in the counter instead, with regression tests beside the existing ones for the same invariant.
- **Telemetry**: `source` reverted to the bare entry-point tag so existing dashboard filters keep matching; kind moved to its own `remixKind` field. The event now counts menu *choices* rather than button clicks, which reads as a conversion drop on a funnel chart — recorded in the schema doc-block with the changeover date.
- **Coverage extended** to the three entry-points the first pass missed (post detail, model gallery ×2, model carousel), so the same image doesn't offer Edit in one feed and nothing in another.

### Testing

- `pnpm run test:lint-rules` — 220/220
- Unit suite — 16 failures across 6 files, **identical to clean `main`** (all six are source-scanning meta-tests failing there too). `blocks.router.workflow.test.ts` collected 347 tests.
- New tests mutation-checked in both directions: reverting the `nsfwLevel` clause fails the search-path case in ~1s, and stripping `withExternalFetch`'s counter participation reproduces the original spinner bug.

### Known, not addressed

An X/XXX image can still be sent to Nano Banana, which Google refuses provider-side — a Buzz-burning dead end rather than an unsafe generation. Deliberately left as a product decision rather than quietly gating it, since refusing all non-safe ratings would remove the feature from a large share of the site's content.

One pre-existing item I left alone: `local-rules/no-wholesale-module-mock` fires on `generation-graph.store.test.ts`, identically on the unmodified `HEAD` version. Reshaping someone else's mock didn't belong in this PR.

PR #3862, opened 2026-08-12
fix(monetization): resync the generation choice, and scope the affirmation to the owner

JustMaier:
The two pre-existing bugs found during #3851 and recorded there as comments. Both are a few lines in `ModelVersionUpsertForm.tsx`, so they ship together.

## The generation choice charged money the screen said was free — [CU 868kqpv6t](https://app.clickup.com/t/868kqpv6t)

Toggling paid access off and on replaces `paidAccessConfig` wholesale, but the radio behind **Generating on-site** is local state (`genMode`) and was left untouched. So:

1. Pick `Free for everyone`.
2. Toggle `Charge for access to this version` off, then on.
3. The radio still reads free. The config underneath charges, and the save writes `freeGeneration: false`.

On-site generation the creator believed they had made free starts being charged, and nothing on screen says so. The handler now resyncs `genMode` from the config it just wrote — the same resync the `form.reset` effect and `clearCharges` already do.

The freshly-seeded config now spells out `freeGeneration` and `acceptsBlueBuzz`: it is built before being handed to `setValue`, so zod's defaults do not reach it, and the resync reads `freeGeneration`.

## A transferred model could not be saved — [CU 868kqpv6z](https://app.clickup.com/t/868kqpv6z)

The client called `hasCurrentRightsAffirmation(version?.meta)` with no owner, while the server passes `ownerId: model.userId` and requires the affirmation to belong to the current owner. An affirmation is a named person accepting liability, so it does not transfer with the model — but the client read the previous owner's record as current, rendered no checkbox, and the save failed server-side with nothing on screen to tick.

**No selector change was needed** — my note on #3851 guessed one would be. The edit query already selects `model.user.id`; the form's `model` prop type simply never admitted it. So this is a type widening plus passing the id.

## Verification

- `pnpm run lint` — no new findings (the same 7 pre-existing warnings in this file)
- `pnpm run test:unit:run` — 17 failures across 8 files, all pre-existing repo-scan suites (app-spend-tier-privilege, the comics `wait:` ledger, the server-graph and typecheck-wrapper guards). A clean `main` measured 18 in the same families earlier today, and none of them touch this component. `blocks.router.workflow.test.ts` collected 347.
- 13 browser tests (3 new), each mutation-tested. Per the brief the new tests assert on **what the save sends** rather than on the form config — reading config instead of payload is exactly what hid the first bug. Reverting either fix fails with a plain assertion (`expected true to be false`, `expected [] to have a length of 1`) in about a second, not a timeout.

One test detail worth knowing: the round-trip fixture stores an access price that differs from the one a fresh config seeds. Without that difference the form is pristine after the round-trip, `handleSubmit` short-circuits to `onSubmit(version)`, and there is no payload to assert on at all.

### DIFF
--- a/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
+++ b/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
@@ -376,8 +376,10 @@ describe('ModelVersionUpsertForm — monetization disclosure', () => {
     );
 
     await userEvent.click(chargeSwitch());
-    // Straight to the ways to charge, with no affirmation in between.
-    await expect.element(feeSwitch()).toBeInTheDocument();
+    // Straight to the ways to charge, with no affirmation in between. Anchored on the switch state and
+    // then read synchronously, so a regression fails in a second instead of waiting out the matcher.
+    await expect.element(chargeSwitch()).toBeChecked();
+    expect(feeSwitch().elements()).toHaveLength(1);
     expect(rightsCheckbox().elements()).toHaveLength(0);
   });
 
--- a/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
+++ b/src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx
@@ -11,6 +11,7 @@ import { renderWithProviders } from '../../../../test/component-setup';
  */
 
 import type * as TrpcModule from '~/utils/trpc';
+import type { ModelVersionTerms } from '@civitai/buzz';
 
 const mutateAsync = vi.hoisted(() =>
   vi.fn(async (input: unknown) => ({ id: 456, ...(input as object) }))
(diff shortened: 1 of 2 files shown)


JustMaier:
Adversarial review came back merge-clean on the fixes, and corrected two things I had written. Both fixed in 89c9be080c:

**My commit message had it backwards on the round-trip test.** The payload assertions in that test *cannot* fail on a revert: the rebuilt config never carried the free grant, so the submitted terms are byte-identical with and without the fix. The synchronous radio read is what catches the regression; the payload assertions guard the *opposite* mistake — "fixing" this by making the config follow the stale radio, which would ship a free grant nobody chose. The test comment now says which assertion does what. Worth stating plainly since the brief for this work asked for payload assertions: for this particular bug the payload was never wrong, only the screen.

**A new false positive I introduced.** Scoping the check to the owner meant a moderator editing someone else's transferred model faced a checkbox the server discards — `resolveRightsAffirmation` exempts a moderator acting on a model they don't own, so the client was blocking a save the server would have accepted. The carve-out now mirrors the server's condition exactly (a known owner who isn't the current user), so staff monetizing their *own* models are still asked. Review caught that I had applied it to the submit gate only, leaving a moderator able to reveal the pricing controls solely by ticking a statement false for them; it now covers the disclosure too.

Two tests added for those paths, both mutation-tested. 15 browser tests total.

**One more pre-existing desync of the same class, not fixed here** (worth its own ticket): pick "A cheaper generation-only price" and leave the box blank, and generation ships at the full access price while the radio says it's cheaper. Nothing validates the empty field, and `generationPrice(terms)` falls back to the download price. This is the one place `genMode` *leads* the config rather than following it.

PR #3863, opened 2026-08-12
fix(api): mini model-version endpoint advertised one file's hash next to another file's download url

### LINKED ISSUE #3861: model-versions/mini returns one file's SHA256/size with another file's download URL on multi-file versions
## Summary

`/api/v1/model-versions/mini/{id}` returns the SHA256 and size of **one** file of a multi-file model version, alongside a `downloadUrls` entry that does not name that file. The URL resolves the file **a second time, independently**, and the two resolutions can disagree — so any consumer that verifies the returned hash against the bytes from the returned URL fails verification **every time, deterministically**, for as long as the version exists.

Both values are internally correct. They are just describing different files.

## Reproduction

Version `3164879` has four files:

| File | Bytes | GiB | |
|---|---:|---:|---|
| `ZiTC_12.0_NVFP4.safetensors` | 3,478,286,923 | 3.2394 | ← the `hashes` and `size` returned |
| `ZiTC_12.0_INT8_CONVROT.safetensors` | 6,163,748,078 | 5.7404 | |
| `ZiTC_12.0_MXFP8_E4M3.safetensors` | 6,348,413,992 | 5.9124 | |
| `ZiTC_12.0_BF16.safetensors` | 12,309,894,424 | 11.4645 | ← what the returned URL serves |

- `GET /api/v1/model-versions/mini/3164879` → `size` 3,396,764 KB and a SHA256 beginning `07B74FF55EE5DC5A62C68AF93864AFD5DA9590A6…`, i.e. the **NVFP4** file, alongside `downloadUrls: ["…/api/download/models/3164879"]`.
- `GET /api/download/models/3164879` serves 12,309,894,424 bytes, i.e. the **BF16** file, whose SHA256 is `05193bc6645a4bf898d6ab0b88aee4ea92f6b4668acff5c4a2ee44d8ea02603d`.

The received hash is **identical across repeated attempts** — the transfer is complete and correct, so this is not corruption or truncation. The expectation is simply attached to a different file.

Same shape on other multi-file versions, where the returned `size` always matches a real sibling rather than the served file: `3195132` (returns the `fp8_e5m2` size, serves `fp32`), `3195694` (`int8` → base), `3220421` (returns the size of a **`training_data.zip`** → serves the safetensors). That last one suggests the selection is not filtering to weights.

## Cause

`src/pages/api/v1/model-versions/mini/[id].ts` picks a file, returns its hash and size, then emits a URL that names no file:

```ts
// :177 — hashes (:312) and size (:308) come from THIS file
const targetFile = modelFileId ? files.find((f) => f.id === modelFileId) : getPrimaryFile(files);

// :222 — but the URL is emitted without it
downloadUrl = `${baseUrl}${createModelFileDownloadUrl({
  versionId: modelVersion.id,
  fileId: modelFileId,      // undefined when the caller did not pin a file
  primary: !modelFileId,    // → true, which nulls every selector
})}`;
```

With `primary: true` and no `fileId`, `createModelFileDownloadUrl` (`src/server/common/model-helpers.ts:8-31`) nulls `type`/`format`/`size`/`fp`/`quantType` and returns a bare `/api/download/models/{versionId}`.

That forces `getFileForModelVersion` to resolve the file again, and the two resolutions do not share inputs:

| | candidate files | preferences given to `getPrimaryFile` |
|---|---|---|
| `mini/[id].ts:177` | raw SQL (`[id].ts:155-171`) | **none** → `defaultFilePreferences` (`SafeTensor`/`pruned`/`fp16`/`Q4_K_M`) |
| `file.service.ts:306` | `dbRead.modelFile.findMany` + visibility filter (`:297-300`) | `{...user?.filePreferences, ...removeEmpty({format,size,fp,quantType})}` (`:301-303`) |

Two independent ways to diverge: different scoring preferences, and different candidate lists. Both sides are deterministic, which is why the mismatch is permanent rather than intermittent.

## Suggested fix

Pin the URL to the file the response just described — `[id].ts:222`:

```ts
downloadUrl = `${baseUrl}${createModelFileDownloadUrl({
  versionId: modelVersion.id,
  fileId: targetFile.id,
})}`;
```

`targetFile.id` is already in scope (the raw query selects `mf.id`, and `:177` matches on `f.id`), and `createModelFileDownloadUrl` emits `?fileId=` whenever given one.

This makes the URL self-consistent with the hash **by construction**, closing both divergence paths at once — no tie-break determinism in `getPrimaryFile` required, and no alignment needed between the two preference sources.

## Related question

Is `getPrimaryFile`'s ranking deterministic when two files score equally? There appears to be no `isPrimary` column, so ties look like they resolve on row order, which would make the served file unstable independently of this bug. Pinning `fileId` as above makes that moot for this path, but it may matter elsewhere.

ZacxDev:
Fixes #3861

## The defect

`GET /api/v1/model-versions/mini/[id]` made **two independent file selections** for one response:

1. **The metadata fields** — `size`, `fileType`, `fileName`, `hashes` — were read off the file *this handler* picked: `modelFileId ? files.find(...) : getPrimaryFile(files)`, with the hardcoded `defaultFilePreferences`.
2. **`downloadUrl`** was built as `?primary=true`, which does not name a file at all. It **delegates the choice** to `/api/download/models/[modelVersionId]`, and that route re-runs `getPrimaryFile` itself.

The two picks are not guaranteed to agree, for two independent reasons:

- **Different populations.** The mini endpoint's raw `ModelFile` query has no `visibility` filter. The download path filters `visibility = Public` for a non-owner/non-moderator.
- **Different preferences.** Mini uses the hardcoded defaults. The download path merges the **requesting user's** `filePreferences` (plus query params) before scoring.

And `getPrimaryFile` has no tiebreak while neither query has an `ORDER BY`, so equal-scoring files resolve by raw row order.

**Result:** on a multi-file version the endpoint returns file A's SHA256 / size / name next to a URL that serves file B. Any consumer that verifies the advertised hash against the downloaded bytes fails — and fails permanently, since nothing about the response is transient.

Concretely, with a version holding a SafeTensor/pruned/fp16 file and a GGUF/Q4_K_M file: the endpoint always describes the SafeTensor (hardcoded defaults), while a caller whose `filePreferences.format` is `GGUF` is served the GGUF by the download route. Same request, two different files.

## The change

Pin the URL to the file the response actually describes:

```ts
downloadUrl = `${baseUrl}${createModelFileDownloadUrl({
  versionId: modelVersion.id,
  fileId: targetFile.id,
})}`;
```

`createModelFileDownloadUrl` already emits `?fileId=` and suppresses `type`/`format`/`size`/`fp`/`quantType` whenever `fileId` is present, so the download route takes its direct-file-lookup branch and resolves that exact row. The route also skips its metadata-misalignment check when `fileId` is set, so there is no new 404 path from the selectors being dropped.

### Visibility

Pinning `fileId` makes the download route's fileId branch enforce `visibility = Public` for a non-owner/non-moderator. Because the mini query never filtered `visibility`, `targetFile` could be a non-public file — which today yields a wrong-but-servable `?primary=true` URL, and after pinning would yield a URL that 404s.

Fixed at the source: the endpoint no longer advertises a file the caller cannot download. The rule **mirrors the download route exactly** rather than filtering unconditionally — `getFileForModelVersion` does `if (!isOwner && !isMod) fileWhere.visibility = Public`, so this does the same:

```ts
const isOwner = !!user?.id && user.id === modelVersion.modelUserId;
const isMod = !!user?.isModerator;
const visibleFiles = isOwner || isMod ? files : files.filter((f) => f.visibility === 'Public');
```

This was not a guess: the endpoint is a `MixedAuthEndpoint` and already resolves owner/moderator context (its version query does `mv.status = 'Published' OR m."userId" = user.id`, and skips that entirely for moderators), so owners/mods are clearly meant to see their own unpublished work here. Mirroring the download route keeps the two surfaces answering the same question the same way. The filter is applied in TypeScript rather than in the raw SQL because the query already `SELECT`s `mf.visibility` (a column nothing read until now) and the row set is one version's files.

Side effect, deliberate: an anonymous caller passing `modelFileId=<a non-public file>` now gets `404` instead of that file's hash/name/size. That is the same answer the download route gives them.

## Behaviour changes callers should know about

- **User file preferences are now bypassed on URLs from this endpoint.** A URL returned here serves the file this endpoint chose, not the file the caller's `filePreferences` would have chosen. This is intended and is the whole point: the URL has to match the hash that was advertised alongside it. Callers who want a specific file should pass `modelFileId` (already supported), which selects both the metadata and the URL together.
- **URL shape changes: `?primary=true` → `?fileId=N`.** This is a new edge-cache key for these URLs, so expect a one-off cache-miss wave on the download route for traffic originating from this endpoint.
- **Download tracking.** The route already recorded `fileId` on the download event from whatever file it resolved; that recorded id is now deterministic and equal to the file the API described.
- **The AIR string is unchanged.** It still carries `+<fileId>` only when the caller explicitly passed `modelFileId`. Making the AIR always name the resolved file would change the identifier for every existing primary-file request, which is a separate and much wider contract change; out of scope here.
- **The private-training-results branch is untouched** — that path emits an orchestrator URL, not a download-route URL, and never had the divergence.

## Tests

New suite: `src/__tests__/pages/api/v1/model-versions/mini-download-url-file-match.test.ts` (9 tests).

- A **discrimination check** asserting the fixture is genuinely one the two paths disagree about — `getPrimaryFile` picks the SafeTensor under the endpoint's hardcoded defaults and the GGUF under a `format: 'GGUF'` preference — so the regression case cannot quietly stop being a regression case.
- The regression assertion itself: the emitted `downloadUrl` carries `fileId=<id of the file whose hashes/size/name were returned>`, with the returned hash/name/size asserted as a positive control.
- A reversed-row-order observation, since neither selection has a tiebreak or an `ORDER BY`.
- The explicit-`modelFileId` path (passes before and after — it is the control arm, that path already emitted `fileId`).
- Visibility: anonymous never gets a non-public file advertised; anonymous asking for one by id gets `404` and no hashes; owner and moderator each still get theirs, with the URL naming it.
- A unit assertion that `createModelFileDownloadUrl({ versionId, fileId })` emits `fileId` and no `type`/`format`/`size`/`fp`/`quantType`/`primary`.

**Red/green matrix**, `pnpm test:unit:run` on this one file:

- At `origin/main` (test present, fix absent): `Tests 4 failed | 5 passed (9)` — the four failures are the regression assertion, the row-order observation, and the two anonymous-visibility cases.
- At HEAD: `Tests 9 passed (9)`.

Full gate on this branch:

- `pnpm typecheck` → `OK — 0 type errors` (positive control: the same command reported 12 errors before the `event-engine-common` submodule was populated locally, so it does report errors when they exist).
- `pnpm test:unit:run` (whole `unit` project) → `Test Files 965 passed | 1 skipped (966)`, `Tests 15053 passed | 17 skipped (15070)`, 0 failed.

### DIFF
--- a/src/__tests__/pages/api/v1/model-versions/mini-download-url-file-match.test.ts
+++ b/src/__tests__/pages/api/v1/model-versions/mini-download-url-file-match.test.ts
@@ -1,0 +1,301 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { NextApiRequest, NextApiResponse } from 'next';
+
+/**
+ * Regression coverage for `GET /api/v1/model-versions/mini/[id]`.
+ *
+ * The endpoint used to make TWO INDEPENDENT file selections for one response:
+ *
+ *   1. the metadata fields (`size`, `fileType`, `fileName`, `hashes`) were read
+ *      off the file this handler picked (`getPrimaryFile` over the version's
+ *      files, with hardcoded default preferences), and
+ *   2. `downloadUrl` was emitted as `?primary=true`, which DELEGATES the choice
+ *      to `/api/download/models/[modelVersionId]` — a route that re-runs
+ *      `getPrimaryFile` itself, over a DIFFERENT population (visibility-filtered)
+ *      and with DIFFERENT preferences (the requesting user's `filePreferences`
+ *      merged in).
+ *
+ * On a multi-file version those two picks diverge, and the response then
+ * advertises file A's SHA256/size/name next to a URL that serves file B. Any
+ * consumer that verifies the hash fails, permanently.
+ *
+ * The contract pinned here: the URL names the file whose metadata was returned.
+ */
+
+const {
+  mockQueryRaw,
+  mockResolveCanGenerateForVersions,
+  mockGetShouldChargeForResources,
+  mockGetFeaturedModels,
+  mockGetCapTiers,
+  currentUser,
+} = vi.hoisted(() => ({
+  mockQueryRaw: vi.fn(),
+  mockResolveCanGenerateForVersions: vi.fn(),
+  mockGetShouldChargeForResources: vi.fn(),
+  mockGetFeaturedModels: vi.fn(),
+  mockGetCapTiers: vi.fn(),
+  currentUser: { value: undefined as undefined | Record<string, unknown> },
+}));
+
+vi.mock('~/server/db/client', () => ({
+  dbWrite: { $queryRaw: mockQueryRaw },
+  dbRead: { $queryRaw: mockQueryRaw },
+}));
+
+vi.mock('~/server/services/generation/generation.service', () => ({
+  resolveCanGenerateForVersions: mockResolveCanGenerateForVersions,
+  getShouldChargeForResources: mockGetShouldChargeForResources,
+}));
+
+vi.mock('~/server/services/model.service', () => ({
+  getFeaturedModels: mockGetFeaturedModels,
+}));
+
+vi.mock('~/server/services/paid-access.service', () => ({
+  getCapTiers: mockGetCapTiers,
+}));
+
+// MixedAuthEndpoint resolves the (optional) session and passes it as the 3rd
+// argument. Swap it for a pass-through that injects whatever `currentUser`
+// holds, so the owner/moderator branches are reachable from a test.
+vi.mock('~/server/utils/endpoint-helpers', () => ({
+  MixedAuthEndpoint:
+    (handler: (req: NextApiRequest, res: NextApiResponse, user: unknown) => unknown) =>
+    (req: NextApiRequest, res: NextApiResponse) =>
+      handler(req, res, currentUser.value),
+}));
+
+import handler from '~/pages/api/v1/model-versions/mini/[id]';
+import { createModelFileDownloadUrl } from '~/server/common/model-helpers';
+import { getPrimaryFile } from '~/server/utils/model-helpers';
+
+const VERSION_ID = 555;
+const OWNER_ID = 4242;
+
+/**
+ * Two files that the two selection paths genuinely disagree about.
+ *
+ * - SAFETENSOR wins `getPrimaryFile` under the DEFAULT preferences the mini
+ *   endpoint hardcodes (format SafeTensor +100, size pruned +10, fp16 +1).
+ * - GGUF wins it under a user whose `filePreferences.format` is `GGUF` — which
+ *   is exactly what the download route merges in. That divergence is asserted
+ *   directly below so this fixture cannot silently stop being discriminating.
+ */
+const SAFETENSOR = {
+  id: 101,
+  type: 'Model',
+  visibility: 'Public',
+  url: 'https://example.invalid/a.safetensors',
+  metadata: { format: 'SafeTensor', size: 'pruned', fp: 'fp16' },
+  sizeKB: 2_097_152,
+  name: 'model-fp16.safetensors',
+  hashes: { SHA256: 'A'.repeat(64) },
+};
+const GGUF = {
+  id: 202,
+  type: 'Model',
+  visibility: 'Public',
+  url: 'https://example.invalid/b.gguf',
+  metadata: { format: 'GGUF', quantType: 'Q4_K_M' },
+  sizeKB: 1_048_576,
+  name: 'model-Q4_K_M.gguf',
+  hashes: { SHA256: 'B'.repeat(64) },
+};
+const PRIVATE_MODEL = {
+  id: 303,
+  type: 'Model',
+  visibility: 'Private',
+  url: 'https://example.invalid/c.safetensors',
+  metadata: { format: 'SafeTensor', size: 'full', fp: 'fp32' },
+  sizeKB: 4_194_304,
+  name: 'private-full.safetensors',
+  hashes: { SHA256: 'C'.repeat(64) },
+};
+
+const versionRow = {
+  id: VERSION_ID,
+  versionName: 'v1',
+  availability: 'Public',
+  publishedAt: new Date('2024-01-01T00:00:00Z'),
+  modelId: 77,
+  modelName: 'Test Model',
+  baseModel: 'SD 1.5',
+  status: 'Published',
+  type: 'Checkpoint',
+  requireAuth: false,
+  checkPermission: false,
+  covered: true,
+  generationAlias: null,
+  minor: false,
+  sfwOnly: false,
+  usageControl: 'Download',
+  modelUserId: OWNER_ID,
+  licensingFee: null,
+  licensingFeeType: null,
+  licensingFeeSettlementCurrency: null,
+  isLicensingRoot: false,
+  licensingSourceVersionId: null,
+  sourceLicensingFeeRecipientUserId: null,
+  sourceModelType: null,
+  sourceBaseModel: null,
+  sourceLicensingFee: null,
+  sourceLicensingFeeType: null,
+  sourceLicensingFeeSettlementCurrency: null,
+  versionFlags: 0,
+  userFlags: 0,
+};
+
+type Body = {
+  size: number;
+  fileType: string;
+  fileName: string;
+  hashes: Record<string, string>;
+  downloadUrls: string[];
+};
+
+async function run(files: object[], query: Record<string, string> = {}) {
+  // Call 1 = the version row, call 2 = the version's ModelFile rows.
+  mockQueryRaw.mockResolvedValueOnce([versionRow]).mockResolvedValueOnce(files);
+
+  const req = {
+    method: 'GET',
+    query: { id: String(VERSION_ID), ...query },
+    headers: { host: 'civitai.com' },
+  } as unknown as NextApiRequest;
+
+  const json = vi.fn().mockReturnThis();
+  const res = {
+    status: vi.fn().mockReturnThis(),
+    json,
+    setHeader: vi.fn().mockReturnThis(),
+  } as unknown as NextApiResponse;
+
+  await handler(req, res);
+  return {
+    res,
+    status: (res.status as unknown as ReturnType<typeof vi.fn>).mock.calls.at(-1)?.[0] as number,
+    body: json.mock.calls.at(-1)?.[0] as Body,
+  };
+}
+
+/** The `fileId` query param on the single emitted download url, if any. */
+function urlFileId(body: Body): string | null {
+  return new URL(body.downloadUrls[0]).searchParams.get('fileId');
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  currentUser.value = undefined;
+  mockQueryRaw.mockReset();
+  mockResolveCanGenerateForVersions.mockResolvedValue(
+    new Map([[VERSION_ID, { canGenerate: true }]])
+  );
+  mockGetShouldChargeForResources.mockResolvedValue({ 77: false });
+  mockGetFeaturedModels.mockResolvedValue([]);
+  mockGetCapTiers.mockResolvedValue(new Map());
+});
+
+describe('createModelFileDownloadUrl', () => {
+  it('emits ?fileId= and suppresses the type/format/size/fp/quantType selectors', () => {
+    const url = createModelFileDownloadUrl({
+      versionId: VERSION_ID,
+      fileId: SAFETENSOR.id,
+      type: 'Model',
+      meta: { format: 'SafeTensor', size: 'pruned', fp: 'fp16', quantType: 'Q4_K_M' },
+    });
+    const params = new URL(url, 'https://civitai.com').searchParams;
+    expect(params.get('fileId')).toBe(String(SAFETENSOR.id));
+    for (const k of ['type', 'format', 'size', 'fp', 'quantType', 'primary'])
+      expect(params.get(k), `${k} must not be emitted alongside fileId`).toBeNull();
+  });
+});
+
+describe('the fixture is discriminating', () => {
+  it('the two selection paths really do pick different files from it', () => {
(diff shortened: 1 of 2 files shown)


ZacxDev:
Follow-up commit `dff13b14` — three fixes from an adversarial re-read of this PR, plus a correction to the PR description.

### Correction to the description above

The description says the old code emitted `?primary=true`. It did not. `createModelFileDownloadUrl` never puts a `primary` key into the query object at all — the flag only *suppresses* the `type`/`format`/`size`/`fp`/`quantType` selectors, and `QS.stringify` drops the key entirely. **The pre-change url was a bare `/api/download/models/<versionId>` with no query string.**

The bug this PR fixes is unaffected by that — a bare url means the download route re-runs `getPrimaryFile` itself, which is exactly the second, independent file selection the PR is removing. But the mechanism is "no selector at all", not "an explicit `primary=true` selector". A test in the suite was asserting `not.toContain('primary=true')`, i.e. guarding a string that never existed in either the old or the new url; that assertion is replaced with one about the shape the url must actually have.

### 1. The visibility gate was applied too early (the important one)

`visibleFiles` was computed before the `availability === Private && targetFile.metadata.trainingResults` branch. That branch builds `downloadUrl` from the orchestrator (`epoch.model_url`) and **never** hits `/api/download/models/[modelVersionId]` — so the gate's entire justification (that the download route would 404 a non-Public `fileId`) does not apply to it.

A Private version's training-results file is non-Public by construction, so for a non-owner, non-moderator caller the gate filtered away the only candidate and the endpoint answered `404 Missing model file` where it previously returned a working epoch url. That is a live regression, not a theoretical one: there are **149 published versions with no public files, 135 of which carry training results and 48 of which are `availability = 'Private'`** — precisely this branch — and a few of those carry access grants to users other than the owner, so a legitimate non-owner caller exists.

Restructured so the caller's target is resolved over the **unfiltered** population first (which is exactly the pre-existing resolution), the url shape is decided from that, and the visibility gate constrains **only** the download-route path. Verified by reading the code that no 404 moved anywhere else: a missing `modelFileId`, an empty file list, and a non-Public `modelFileId` on the download-route path all still 404 with the same messages.

### 2. The caller's file preferences were silently dropped

The download route merges the requesting user's `filePreferences` into `getPrimaryFile` (`getFileForModelVersion`). Pinning the url to a `fileId` took that merge out of the loop, and this endpoint was calling `getPrimaryFile` with hardcoded defaults — so a user who prefers, say, GGUF was handed the SafeTensor regardless. The preferences are now passed through, which restores preference-awareness while keeping the hash/url consistency the pin exists for. (The download route also merges explicit `format`/`size`/`fp`/`quantType` query params; this endpoint accepts none of those, so there is nothing else to merge.)

### 3. Two surviving mutants in the new test suite

- **`isOwner = !!user?.id`** — treating any authenticated user as the owner — passed the original 9 tests, because every case was anonymous, the owner, or a moderator. Added authenticated non-owner cases.
- **Leaving the primary path unfiltered** also passed all 9, because the fixture was not discriminating: the non-public file scored *lower* than the public one under the default preferences, so the public file won whether or not the population was filtered. The non-public fixture now out-scores the public one, and that property is asserted directly so it cannot silently rot.

### Verification

- **Red/green:** the four new behavioural tests were run against the pre-fix source first — **4 failed / 16 passed**. After the source fixes, **23 passed / 0 failed**. (The two authenticated-non-owner cases pass on the pre-fix source by design — the ownership check was already correct there — so they are mutant guards, not regression tests, and are labelled as such below.)
- **Mutation battery, re-run in full after the fixes** (21 mutants: operand swaps, inverted ternaries and comparisons, dropped conjuncts, wrong binds, unfiltered populations, `pop()`→`shift()`, guard-removal, and the original bug re-introduced): **21/21 killed**, with an unmutated control confirming the battery's baseline is green.
- **Gate:** `pnpm typecheck` → **0 errors** (instrument validated: it reports 1 error and exits non-zero on a deliberately planted type error). Full unit suite → **965 files / 15,067 tests passed, 0 failed**. Prettier and ESLint clean on both changed files.

### Incidental finding

The epoch selection ends in `trainingResults.epochs?.pop()`, which **mutates** the metadata object it was handed. Harmless in production because every request re-reads the rows, but it drains a shared test fixture after one use; the test harness now deep-clones the mocked rows. Worth knowing before anyone caches that metadata.

ZacxDev:
## Follow-up round: branch independence, an assertable AIR, and one coverage gap

Pushed as `313e6554a3`. Three changes plus two comment corrections.

### 1. The URL branch is now decided with DEFAULT file preferences

The previous round fed the caller's `filePreferences` into the **branch-deciding** `getPrimaryFile`. That let a stored per-user preference flip the URL *shape*, not just the variant: a GGUF-preferring caller on a Private version got a `/api/download/models/…?fileId=` URL and a `source:'civitai'` AIR, where a default-preference caller got an orchestrator epoch URL and a `source:'orchestrator'` AIR — and `?epoch=N` silently became a no-op for the first caller.

**Decision: the branch is decided with default preferences; the caller's `filePreferences` apply only to the download-route arm.**

Rationale, stated here because it was previously encoded only in a test name: the branch selects which **kind** of resource the response addresses — an orchestrator epoch artifact vs a civitai `ModelFile` — and that choice determines the **AIR**, a shared identifier that is logged, cached, and handed to the orchestrator. It must not vary per caller. A preference answers *"which variant do I want"*, not *"which resource is this"*.

Hash/URL consistency is unaffected: `targetFile` still governs both the advertised hash and the URL on the download-route arm.

The test that asserted the opposite ("the caller's preferences decide WHICH branch is taken") is inverted into one asserting the branch is preference-**independent** — same branch, same file, same AIR for a default-preference and a GGUF-preferring caller on the same version. A separate case still proves preferences **do** choose the file on the download-route arm of a Private version, with the URL still naming the file whose hash was advertised.

### 2. The AIR is now assertable, and two live mutants are dead

`Air.stringify` is globally stubbed to `() => ''` in the unit setup, so `body.air` carries no information in any unit suite and nothing about it could be asserted. Rather than touch the global stub, the tests assert its **arguments**, which are fully observable and are exactly what the handler derives from its own state.

Two mutants that survived the entire previous suite are now killed:

- **swapping `modelId: jobId` / `id: fileName`** — ships a structurally valid but unresolvable orchestrator URN; nothing in the response body changes, so only the AIR arguments can see it.
- **`fileType: targetFile.type` → `requestedFile?.type`** — observable on a population where the gated and ungated selections disagree *and* the two files' `ModelFile.type` map to different AIR types (a new `Diffusion Model` fixture).

### 3. One missing coverage case

Private version + authenticated non-owner + a non-Public, non-training file, no `modelFileId`. This pins the gate-bypass half of `useEpochUrl`: a mutant dropping the `trainingResults` conjunct made `useEpochUrl` true for *any* Private version, which then takes `targetFile = requestedFile` — the **ungated** pick — and, with no training results to follow, falls through to the download-route arm advertising a non-Public file's hash/name/size to a non-owner. That mutant survived all 23 previous tests. The shipped code was already correct; this was a coverage gap on the exact boundary the fix round introduced.

### Comment corrections

Two comments (one in the endpoint, one in the test file header) claimed the old URL emitted `?primary=true`. That is descriptively wrong: `QS.stringify` is never handed a `primary` key, so `createModelFileDownloadUrl`'s `primary` flag only **suppresses** the other selectors — the old URL was a **bare** `/api/download/models/<id>` with no query string. Reworded to the actual mechanism (a bare URL names no file, so the download route re-resolves one independently); the consequence is unchanged. The matching `expect(params.get('primary')).toBeNull()` assertion guarded a string that can never exist and passed under every mutation — replaced with a falsifiable one (`fileId` is the *only* param).

### Two smaller items

- **`epochs?.pop()` → `epochs?.[length - 1]`.** The endpoint was the only reader here that **mutated** the row it was handed; `model-version.service.ts` already expresses the same "latest epoch" selection non-mutatingly. Consolidated on that form.
- **The AIR's `fileId` is left as the query param, deliberately**, and that is now documented in code and pinned by a test. Correcting the premise: the AIR and the URL cannot actually name *different* files — when the caller supplies `modelFileId`, `targetFile` was looked up **by** it, so they agree; when omitted, the AIR simply carries no `+<fileId>` disambiguator. It under-specifies rather than contradicting. Binding it to `targetFile.id` would append `+<fileId>` to essentially every mini response's AIR — a shared, logged, cached identifier — which is its own change with its own blast radius, not a drive-by.

### Verification

| Gate | Result |
|---|---|
| Unit suite | 23 → **30 tests**, 30 passed |
| Red/green | The branch-independence case is **red on the previous commit** (1 failed / 29 passed) and green here. Everything else passes both ways — labelled **mutant guards**, not regression coverage. |
| Mutant battery | **17 mutants, 17 killed, 0 survivors** (re-run in full after the change) |
| ESLint / Prettier | 2 files linted, 0 errors / 0 warnings; formatting clean |

Every instrument was validated with a negative control before its verdict was believed (planted type error → 1 error; planted formatting → warn; planted unused var → warning; and the suite was watched red 18 separate times).

The battery was enumerated from semantic failure modes rather than syntax: swapped operands, inverted ternary and inverted identity comparisons, a guard left present but **dead**, wrong-variable binding, off-by-one on the epoch fallback, dropping either conjunct of `useEpochUrl`, and applying the change to only **one** of the two parallel `getPrimaryFile` call sites.

### One finding worth flagging separately

`tsconfig.json` excludes `src/**/__tests__/**`, so **`pnpm typecheck` does not read any file under `src/__tests__/`** — a planted type error there produces 0 errors and a clean run. Checking this file with that exclusion lifted found **27** type errors in it, **19 of them pre-existing** from the previous round. All 27 are now fixed (the fixtures' `metadata` literals were widening to `string` and so did not satisfy `FileFormatType`). The broader exclusion is untouched here — there are ~800 such errors across the excluded tree — but it means type regressions in test files are currently ungated repo-wide.

PR #3864, opened 2026-08-13
fix(api): pin per-file downloadUrl to its own fileId on /api/v1/models/[id] and /api/v1/model-versions/[id]

ZacxDev:
Refs #3861. Sibling-endpoint follow-up to #3863 (which fixed the same class on `/api/v1/model-versions/mini/[id]`); this PR is branched off `main`, not off that PR.

## The defect — an API contract violation

`GET /api/v1/models/[id]` and `GET /api/v1/model-versions/[id]` both serialize, per model file, that file's `hashes`, `name` and `sizeKB` — and next to them a `downloadUrl` that was **not** pinned to that file:

```ts
createModelFileDownloadUrl({
  versionId: version.id,
  type: file.type,
  meta: metadata,
  primary: primaryFile.id === file.id,
})
```

Two problems follow from that:

1. `QS.stringify` drops the `primary` key entirely, and `createModelFileDownloadUrl` nulls out `type`/`format`/`size`/`fp`/`quantType` when `primary` is true. So **for the file elected primary the URL is a bare `/api/download/models/<versionId>` with no query string at all.**
2. That URL is re-resolved from scratch by `/api/download/models/[modelVersionId]` → `getFileForModelVersion`, whose `else` branch re-picks the file using **the requesting user's `filePreferences`** merged with any query params, scored by `getPrimaryFile`. `getPrimaryFile` has no tiebreak and neither query has an `ORDER BY`, so equal-scoring files resolve by raw row order.

Net effect: on a version with more than one public file, the response can advertise file A's SHA256, name and size beside a URL that serves file B — and *which* file it serves depends on who is asking. Any consumer that verifies the advertised hash against the downloaded bytes fails, permanently and non-deterministically.

**3,587 Published versions** currently have more than one `Public` primary-type file with more than one distinct `metadata.format` — the population where the two picks can diverge.

## The change

Pin each file's URL to that file:

```ts
createModelFileDownloadUrl({ versionId: version.id, fileId: file.id })
```

Verified against `createModelFileDownloadUrl` rather than assumed: supplying `fileId` suppresses `type`, `format`, `size`, `fp` and `quantType` (each is gated on `!primary && !fileId`), so the emitted URL is exactly `/api/download/models/<versionId>?fileId=<n>` — pinned in a direct unit test in this PR, with a positive control showing the discriminators *are* still emitted when no `fileId` is given.

On the download side, `getFileForModelVersion`'s `fileId` branch does a direct `findFirst({ id: fileId, modelVersionId })`, bypassing preference scoring entirely, and `/api/download/models/[modelVersionId]` explicitly **skips** its metadata-misalignment check when `input.fileId` is set (the file is already exact), so the new query string cannot cause a spurious 404.

The **version-level** `downloadUrl` on both endpoints is deliberately unchanged. It advertises no per-file metadata and is the "just give me the default file" entry point, so re-resolution there is the intended behaviour, not a mismatch.

### On consolidating into a shared helper

I considered extracting a `downloadUrlForFile(versionId, file)` wrapper, since three or four call sites need the same "pin the URL to the file whose metadata I just serialized" rule. I did not, and deliberately: `createModelFileDownloadUrl({ versionId, fileId })` **is** already the single place that rule lives, and the fix is precisely to stop passing the ambiguous arguments. A wrapper would add a name and an indirection without removing any duplicated logic — the call is one line either way. The smaller change wins here.

## Behaviour changes

- The primary file's URL gains `?fileId=N` where it previously had none; the other files' URLs swap `?type=…&format=…&size=…&fp=…` for `?fileId=N`. Both still point at `/api/download/models/<versionId>`, which is `Cache-Control: no-store` / `private, no-store` — so no CDN or edge caching behaviour changes.
- **The caller's `filePreferences` no longer influence which file a per-file URL serves.** That is the point of the fix — the URL must match the hash printed beside it — and the preference-driven "give me my preferred variant" behaviour is still available through the unchanged version-level `downloadUrl`.
- These are unauthenticated, shared-cache public responses (`/api/v1/models/[id]` in particular is served from an origin cache keyed only by `(modelId, browsingLevel)` behind an edge cache), so selecting per-caller preference at serialization time is not a viable alternative here — it would be both wrong for a shared cache entry and still leave the advertised hash and the served bytes free to disagree.

## Visibility

Checked rather than assumed: **both** endpoints already filter the serialized file list to `visibility === ModelFileVisibility.Public` before mapping. So every file that gets a pinned URL is one the download route's non-owner path (`if (!isOwner && !isMod) → visibility: Public`) will also serve. Pinning `fileId` therefore cannot turn a previously-working URL into a 404 for anonymous callers. Regression tests assert the Public-only filter on both endpoints so a future change can't quietly widen it under a pinned URL.

(Separate, pre-existing, not touched here: `getPrimaryFile` is called over the *unfiltered* file list on both endpoints, so if the elected primary happens to be non-Public, no serialized file carries `primary: true`. That is a labelling oddity in the response, not a hash/URL mismatch.)

## Other consumers of `createModelFileDownloadUrl`

Enumerated exhaustively (every call site in `src/`), not sampled:

**Same defect, not fixed in this PR** (each serializes per-file metadata next to a `primary`-derived URL):
- `src/server/services/model-search.service.ts:267` — backs the `/api/v1/models` list endpoint. Same shape, same `Public`-only filter. The best candidate for a follow-up.
- `src/server/webhooks/model.webooks.ts:63` and `:129` — the `new-model` / `updated-model` webhook payloads. Same shape, and notably these have **no visibility filter at all**, so a fix there needs the visibility question settled first.

**Already correct** (pass `fileId`):
- `src/pages/api/v1/model-versions/mini/[id].ts:241`
- `src/components/Model/ModelVersions/RequiredComponentsSection.tsx` (5 sites)

**Not affected** — version-level "default download" URLs that advertise no per-file metadata: `models/[id].ts:147`, `model-versions/[id].ts:274`, `model-search.service.ts:297`, `model.webooks.ts:76`/`:142`.

**Not affected** — UI call sites that build a URL for a file the user explicitly picked in the same interaction, or for a single known file: `DownloadVariantDropdown.tsx`, `ModelVersionDetails.tsx`, `ModelFileAlert.tsx`, `UserTrainingModels.tsx`, `Training/Form/{trainingSamplePrompts.ts,TrainingImages.tsx,ImageSelectModal.tsx}`.

## Tests

Two new suites (13 tests), both using the **real** `createModelFileDownloadUrl` and the **real** `getPrimaryFile` — the URL string and the primary election are the contract under test, so neither is stubbed. Fixtures use four files with pairwise-distinct `id`/`type`/`format`/`size`/`fp`/`sizeKB`/`hash`, so a wrong-file binding cannot coincidentally produce a correct URL. Each assertion recovers which file the response *claims* an entry is (via the advertised SHA256) and then checks the URL points at **that** file.

- `src/tests/api/v1/models/id-file-download-url.test.ts` (8)
- `src/tests/api/v1/model-versions/id-file-download-url.test.ts` (5)

Red/green matrix, same tests both sides:

| | at `origin/main` source | with the fix |
|---|---|---|
| both suites | **7 failed / 6 passed (13)** | **13 passed (13)** |

Mutation battery — 13 mutants plus an unmutated control, run on an isolated copy of the tree:

| mutant | result |
|---|---|
| `fileId: file.id` → `primaryFile.id` (models) | killed (2 failed) |
| `fileId: file.id` → `file.id + 1` (models) | killed (3 failed) |
| full revert to the pre-fix call (models) | killed (4 failed) |
| drop the `fileId` key entirely (models) | killed (4 failed) |
| `primary: primaryFile.id === file.id ? true : undefined` → `primary: true` | killed (1 failed) |
| delete the `Public` visibility filter (models) | killed (2 failed) |
| `fileId: file.id` → `primaryFile.id` (model-versions) | killed (1 failed) |
| `fileId: file.id` → `file.id - 1` (model-versions) | killed (2 failed) |
| full revert to the pre-fix call (model-versions) | killed (3 failed) |
| `primary: … === …` → `!==` (model-versions) | killed (1 failed) |
| delete the `Public` visibility filter (model-versions) | killed (2 failed) |
| helper: `fileId: fileId ?? null` → `null` (branch present but dead) | killed (8 failed) |
| helper: drop `&& !fileId` from all five suppression gates | killed (1 failed) |
| **control: unmutated** | **13 passed** |

**No survivors.** Two mutants survived the first pass and drove real test changes: `primary: true` (the suite only looked up the first entry claiming primary — now it asserts the *exact set*), and the helper-suppression mutant (unobservable from the endpoints, which no longer pass `type`/`meta` at all — hence the direct `createModelFileDownloadUrl` contract test).

## Gates

- `pnpm typecheck` — **0 type errors**. Instrument validated: a deliberately planted type error in `src/pages/api/v1/models/[id].ts` produced `1 type error(s)` / exit 2, then reverted.
- `vitest --project unit` over the touched areas (`src/tests/api/v1/{models,model-versions,model-files}`, `src/__tests__/pages/api/download`, the four `model-search` suites): **17 files, 161 tests, 161 passed, 0 failed**.
- `eslint` on the four changed files: 0 errors, 24 `no-explicit-any` **warnings** in the two new test files, matching the style of the existing sibling suite (`id-origin-cache.test.ts`).
- `prettier --check` on the four changed files: clean.

### DIFF
--- a/src/pages/api/v1/model-versions/[id].ts
+++ b/src/pages/api/v1/model-versions/[id].ts
@@ -7,7 +7,10 @@ import * as z from 'zod';
 import { getEdgeUrl } from '~/client-utils/cf-images-utils';
 import { isProd } from '~/env/other';
 import { getDownloadFilename } from '~/server/services/file.service';
-import { createModelFileDownloadUrl } from '~/server/common/model-helpers';
+import {
+  createModelFileDownloadUrl,
+  createSerializedFileDownloadUrl,
+} from '~/server/common/model-helpers';
 import { dbRead } from '~/server/db/client';
 import type { ModelVersionApiReturn } from '~/server/selectors/modelVersion.selector';
 import { getImagesForModelVersion } from '~/server/services/image.service';
@@ -257,9 +260,17 @@ export async function prepareModelVersionResponse(
             // route re-resolves the file from the caller's filePreferences, so
             // the advertised `hashes`/`sizeKB`/`name` and the bytes actually
             // served can disagree on a multi-file version.
-            downloadUrl: `${baseUrl.origin}${createModelFileDownloadUrl({
-              versionId: version.id,
-              fileId: file.id,
+            //
+            // `castedFiles` is NOT all this version's own files: the `vaeId`
+            // splice above pushes files that live on the LINKED VAE version.
+            // createSerializedFileDownloadUrl pins ONLY a file this version
+            // owns; a spliced one keeps its original discriminator URL, which
+            // the download route resolves through the linked-component
+            // fallback. See the invariant note on that helper.
+            downloadUrl: `${baseUrl.origin}${createSerializedFileDownloadUrl({
+              file: { id: file.id, modelVersionId, type: file.type, metadata },
+              hostVersionId: version.id,
+              primary: primaryFile.id === file.id,
             })}`,
           }))
       : [],
--- a/src/pages/api/v1/model-versions/[id].ts
+++ b/src/pages/api/v1/model-versions/[id].ts
@@ -252,11 +252,14 @@ export async function prepareModelVersionResponse(
               getDownloadFilename({ model, modelVersion: version, file })
             ),
             primary: primaryFile.id === file.id,
+            // Pin the URL to THIS file — see the matching note in
+            // src/pages/api/v1/models/[id].ts. Without `fileId` the download
+            // route re-resolves the file from the caller's filePreferences, so
+            // the advertised `hashes`/`sizeKB`/`name` and the bytes actually
+            // served can disagree on a multi-file version.
             downloadUrl: `${baseUrl.origin}${createModelFileDownloadUrl({
               versionId: version.id,
-              type: file.type,
-              meta: metadata,
-              primary: primaryFile.id === file.id,
+              fileId: file.id,
             })}`,
           }))
       : [],
--- a/src/pages/api/v1/models/[id].ts
+++ b/src/pages/api/v1/models/[id].ts
@@ -5,7 +5,10 @@ import { ModelFileVisibility, ModelModifier } from '~/shared/utils/prisma/enums'
 
 import { getEdgeUrl } from '~/client-utils/cf-images-utils';
 import { ModelSort } from '~/server/common/enums';
-import { createModelFileDownloadUrl } from '~/server/common/model-helpers';
+import {
+  createModelFileDownloadUrl,
+  createSerializedFileDownloadUrl,
+} from '~/server/common/model-helpers';
 import { getDownloadFilename } from '~/server/services/file.service';
 import { getModelsWithVersions } from '~/server/services/model.service';
 import { publicModelResponseKey } from '~/server/services/model-version.service';
@@ -115,7 +118,7 @@ async function buildPublicModelResponse(
           files: includeDownloadUrl
             ? castedFiles
                 .filter((file) => file.visibility === ModelFileVisibility.Public)
-                .map(({ hashes, metadata, ...file }) => ({
+                .map(({ hashes, metadata, modelVersionId, ...file }) => ({
                   ...file,
                   metadata: removeEmpty(metadata),
                   name: safeDecodeURIComponent(
@@ -132,9 +135,16 @@ async function buildPublicModelResponse(
                   // discriminator in createModelFileDownloadUrl, and the
                   // download route skips its metadata-misalignment check
                   // because the file is already exact.
-                  downloadUrl: `${baseUrl}${createModelFileDownloadUrl({
-                    versionId: version.id,
-                    fileId: file.id,
+                  //
+                  // `version.files` is NOT all this version's own files:
+                  // getModelsWithVersions splices in the linked VAE version's
+                  // files (`files.push(...vaeFile)`), which keep their own
+                  // `modelVersionId`. createSerializedFileDownloadUrl pins ONLY
+                  // a file this version owns — see the invariant note there.
+                  downloadUrl: `${baseUrl}${createSerializedFileDownloadUrl({
+                    file: { id: file.id, modelVersionId, type: file.type, metadata },
+                    hostVersionId: version.id,
+                    primary: primaryFile.id === file.id,
                   })}`,
                   primary: primaryFile.id === file.id ? true : undefined,
                   url: undefined,
--- a/src/pages/api/v1/models/[id].ts
+++ b/src/pages/api/v1/models/[id].ts
@@ -122,11 +122,19 @@ async function buildPublicModelResponse(
                     getDownloadFilename({ model, modelVersion: version, file })
                   ),
                   hashes: hashesAsObject(hashes),
+                  // Pin the URL to THIS file. Passing `type`/`meta`/`primary`
+                  // instead makes the download route re-resolve the file on its
+                  // own (caller's filePreferences + its own scoring) — and for
+                  // the elected primary it emits a bare, entirely unqualified
+                  // URL — so on a multi-file version the URL could serve a file
+                  // other than the one whose `hashes`/`sizeKB`/`name` are
+                  // serialized right here. `fileId` suppresses every soft
+                  // discriminator in createModelFileDownloadUrl, and the
+                  // download route skips its metadata-misalignment check
+                  // because the file is already exact.
                   downloadUrl: `${baseUrl}${createModelFileDownloadUrl({
                     versionId: version.id,
-                    type: file.type,
-                    meta: metadata,
-                    primary: primaryFile.id === file.id,
+                    fileId: file.id,
                   })}`,
                   primary: primaryFile.id === file.id ? true : undefined,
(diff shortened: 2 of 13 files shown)


ZacxDev:
## Deploy-blocking defect found and fixed: pinning `fileId` could 404 a previously-working URL

Pushed `4f7673f` on top of `165900f`.

### The defect

The pin was applied to **every** file in a version's serialized list — but not every file in that list belongs to that version. Both endpoints splice **cross-version** files in:

* `prepareModelVersionResponse` (`/api/v1/model-versions/[id]`) does `files.push(...vae)` when the version has a linked VAE;
* `getModelsWithVersions` (behind `/api/v1/models/[id]`) does `files.push(...vaeFile)`.

`getVaeFiles` reads the primary `Model` files off the **linked** version, relabels `type` to `VAE`, and returns them with **their own ids** and a real `Public` visibility — so they pass the visibility filter and serialize exactly like a local file. Pinning one emitted:

```
/api/download/models/<hostVersionId>?fileId=<vaeFileId>
```

The download route resolves a pinned URL with `findFirst({ id: fileId, modelVersionId })`. Those two halves disagree, so it resolves to `null` → **404 "File not found."** Reproduced on both endpoints against the previous tip: `…/api/download/models/4242?fileId=91`, where file 91 lives on version 9999.

### Correction to the PR body

> "pinning `fileId` cannot turn a previously-working URL into a 404"

**This is wrong, and it was the shipping risk.** Before the pin, that same file's URL carried `?type=VAE`, which takes the download route's `else` branch, finds nothing local, and **succeeds through the linked-component fallback** (`VAE` is a component file type). Pinning `fileId` bypasses that fallback entirely. So the pin took a URL that worked and made it a permanent 404 for every version with a linked VAE component.

### Blast radius the body did not name

`prepareModelVersionResponse` / `resModelVersionDetails` also shape the bodies of `/api/v1/model-versions/by-hash` and `/api/v1/model-versions/by-hash/[hash]` — the endpoints the SD-WebUI helper extensions call. They therefore **inherit the pin**. They do *not* inherit the cross-version splice (their select carries no `vaeId`, so `getVaeFiles` is never called there) — that is now asserted by a test rather than assumed, so the by-hash endpoints changed shape only, they were never 404ing.

### The fix

Rather than special-casing VAE, the invariant is made structural. A single `createSerializedFileDownloadUrl` is now the only place a per-file `downloadUrl` is built. `hostVersionId` is the only version id it will ever put in the path, and `fileId` rides along **only** when the file reports belonging to that same version.

A file the host version does not own — or one with no `modelVersionId` at all — keeps **exactly** the pre-pin discriminator URL. That is deliberate:

* it cannot 404 (the route re-resolves it, and the linked-component fallback handles component types);
* it keeps that file's access gating evaluated against the **host** version, and its download attribution on the host version.

Re-pointing such a URL at the file's *own* version would also resolve, but it would silently move both gating and attribution — a bigger change than this fix should make. Happy to revisit as a separate decision.

One supporting change: `getModelsWithVersions` was discarding `modelVersionId` while reducing each file, so `/api/v1/models/[id]` had no way to tell an owned file from a spliced one. That reduction is now a named `toApiModelFile` that preserves it; both consumers strip it from the wire body, so **no public response shape changes**.

### Two behaviour changes worth stating plainly

1. **A stored `?fileId=N` URL breaks permanently if the owner replaces that file.** A bare `/api/download/models/<versionId>` kept working across a file replacement because it re-resolved at download time; a pinned URL points at a specific row and 404s once that row is gone. This is the intended trade (exactness over durability), but anything that persists download URLs — bookmarks, third-party mirrors, cached manifests — will feel it.
2. **Per-file download attribution becomes exact.** `modelVersionEvents.fileId` is recorded from the resolved file, and the resolution used to be re-derived from the caller's file preferences. With the pin it is whatever the URL names, so attribution that was smeared across a version's file variants is now precise — expect a **step change at deploy** on any dashboard grouping by that field. It is a measurement change, not a traffic change.

### Coverage

The gap is why this shipped: the model-versions suite mocked `getVaeFiles` to `[]` with no `vaeId`, the models suite mocked `getModelsWithVersions` wholesale so the splice sat outside the harness, and every fixture used a single version id — so "this fileId belongs to this version" was not a proposition either suite could state. The central assertion **passed** on an injected cross-version file.

Added: cross-version fixtures on both endpoints (the model-versions one runs the real splice); a test of the real `getVaeFiles` pinning the shape the splice actually produces; by-hash coverage; a shape guard on the `/api/v1/models` list body; and — new — a **seam** suite that feeds the emitted URL to the real `getFileForModelVersion` over a stub that enforces the `where` clause, so the test asserts the pair *resolves*, not that the string looks right. It includes a negative control proving the bug's pair is a 404 there.

Verification: 19 new assertions red on `165900f` and green on `4f7673f`; 12 further new cases pass both ways and are invariant guards (they pin premises: what `getVaeFiles` emits, what the download route does with a mismatched pair). A 17-mutant battery kills 15; the 2 survivors are provably equivalent rewrites of the ownership check. `pnpm typecheck` 0 errors (validated with a planted error → 1); 691 unit tests green across the touched areas.

PR #3866, opened 2026-08-13
fix(model-file): deregister a deleted file from the storage registry

ZacxDev:
> [!NOTE]
> **UNBLOCKED — the precondition is satisfied.** This was blocked on the
> storage-resolver `fileIds` endpoint being **merged AND deployed**; both are
> done, and the running resolver was verified to serve the `fileId` path.
> The original banner said "BLOCKED — do not merge / draft on purpose"; that is
> now stale and has been replaced. Edit noted in a comment rather than made
> silently, since reviewers may have read the earlier version.
>
> The ordering it described was real and still explains the design: this client
> is best-effort and swallows every failure, so shipping it before the endpoint
> existed would have made every call fail **silently** — inert while looking
> done. That is also why the deregister's Axiom events now carry `key: 'file'`,
> and why success is confirmed server-side rather than from app logs.

## The bug

Deleting a **single model file** left a stale entry behind in the storage registry.

`deleteFile` removes the row, busts the version file cache and cleans up the stored
object — but it never tells the storage-resolver the file is gone. The existing
deregistration API is keyed on **model version**, and a single-file delete leaves the
version alive, so nothing version-keyed can ever reach that entry either. The registry
goes on describing a file that no longer exists.

Reached from the UI via `trpc.modelFile.delete` (the per-file delete in the resource
files editor).

## The change

**`deregisterFileLocationsByFile(fileIds: number[])`** in `src/utils/storage-resolver.ts`,
alongside the existing `deregisterFileLocations` / `deregisterFileLocationsBatch`. It
mirrors the batch variant's contract *exactly*:

- same bearer auth against the storage-resolver internal endpoint (`env.STORAGE_RESOLVER_INTERNAL_URL`)
- same unconditional per-request `AbortSignal.timeout` so a hung resolver can't stall cleanup
- same 500-id chunking, summing `deleted` across chunks
- same best-effort semantics: never throws, `null` on a config skip, a failed chunk is
  logged to Axiom and skipped rather than aborting the rest
- same Axiom event names (`deregister-file-locations-{skipped,failed,error}`)

It takes an **array** so the bulk shape exists; the single-file delete passes one id.
The payload carries **`fileIds` and nothing else** — the endpoint rejects a request that
sets both a file-keyed and a version-keyed field, and there is a test pinning that the
body's key set is exactly `['fileIds']`.

**Call site** — `deleteFile` in `src/server/services/model-file.service.ts`, post-commit,
beside the existing object cleanup, same fire-and-forget shape with a terminal `.catch`:

- keyed on **`id`** — the id the `DELETE ... RETURNING` actually removed. Not re-read.
- gated on a returned row: **no row means nothing was deleted**, and deregistering then
  would drop a registry entry for a file that still exists.

## Service contract this depends on

`POST /deregister` gains `fileId` / `fileIds`, dispatch precedence
`fileIds → fileId → modelVersionIds → modelVersionId`, response `{ success, deleted, rows? }`.
Setting both a file-keyed and a version-keyed field is a `400`. This client only ever
sends `fileIds`.

## Tests

Two new suites, **17 tests**, matching the conventions of the existing
`storage-resolver.deregisterBatch.test.ts` and `model-version.deregister.service.test.ts`
(no new harness):

- `src/utils/__tests__/storage-resolver.deregisterByFile.test.ts` (10) — payload shape,
  bearer header, abort signal, exact key set, not-configured → `null` + warn, de-dupe /
  non-positive drop, chunking at 500, per-chunk best-effort on both a non-OK response and
  a thrown one, a `404` (contract not yet deployed) resolving rather than throwing, and a
  timeout resolving rather than throwing.
- `src/server/services/__tests__/model-file.deregister.service.test.ts` (7) — deregisters
  by file id; still deletes the object and busts the version cache; a rejecting or
  `null`-returning deregister does not break the delete; still deregisters when the deleted
  row carried no url; does **not** deregister when the `DELETE` removed nothing; does
  **not** deregister when the delete is refused before the `DELETE` runs.

Expectations are pinned to literal values (`{ fileIds: [123] }`, `[123]`), not derived from
what the code builds. Ids in the service suite are pairwise distinct (file `123`, version
`4242`, model `7`) so a test cannot pass by picking up the wrong one.

### Mutation matrix

Every behaviour was broken on purpose and the suites watched to go **red for that specific
reason**. Baseline unmutated: 17 passed. 15/15 mutants killed, 0 survived.

| # | Mutation | Result | Killing assertion |
|---|---|---|---|
| H1 | payload key `fileIds` → `modelVersionIds` | 4 failed | `expected { modelVersionIds: [123] } to deeply equal { fileIds: [123] }` |
| H2 | send **both** `fileIds` and `modelVersionId` | 3 failed | `expected ['fileIds','modelVersionId'] to deeply equal ['fileIds']` |
| H3 | drop the `AbortSignal` timeout | 1 failed | `expected undefined to be an instance of AbortSignal` |
| H4 | drop the not-configured guard | 1 failed | `expected { deleted: 0 } to be null` |
| H5 | drop de-dupe + non-positive filter | 2 failed | `expected { fileIds: [1,1,2,0,-3,2] } to deeply equal { fileIds: [1,2] }` |
| H6 | remove chunking (one request) | 3 failed | `expected { deleted: 500 } to deeply equal { deleted: 1200 }` |
| H7 | non-OK chunk aborts the loop (`break`) | 1 failed | `expected { deleted: 0 } to deeply equal { deleted: 500 }` |
| H8 | thrown chunk rethrows (no best-effort) | 2 failed | `promise rejected "TimeoutError" instead of resolving` |
| H9 | drop the bearer `Authorization` header | 1 failed | `expected undefined to be 'Bearer test-token'` |
| C1 | drop the deregister call entirely | 4 failed | `expected vi.fn() to be called 1 times, but got 0 times` |
| C2 | key on `modelVersionId` instead of the file id | 4 failed | `expected [4242] to deeply equal [123]` |
| C3 | call even when the `DELETE` removed nothing | 1 failed | `expected vi.fn() to not be called at all, but actually been called 1 times` |
| C4 | drop the best-effort catch (`await` instead) | 1 failed | the delete rejects with `storage-resolver down` |
| C5 | object cleanup targets the wrong url | 2 failed | `expected vi.fn() to be called with arguments: [...]` |
| C6 | drop the version cache bust | 1 failed | `expected vi.fn() to be called with arguments: [4242]` |

H7/H8 needed the *last* occurrence of an anchor shared with the batch helper; the script
asserts the edit landed after `deregisterFileLocationsByFile`'s declaration, so a mutant
could not silently apply to the wrong function. Source files were checksummed before and
after the battery and restored byte-identical.

**Not pinned by a test (stated rather than rounded up):** the Axiom event *name* on the
call site's own catch (`model-file-delete-deregister-file-locations`) — the equivalent on
the version-delete path isn't asserted either. And the resolver's 5000-id batch cap is not
exercised: the client chunks at 500, so it cannot reach it.

## Verification

- `pnpm typecheck` — 12 errors, **byte-identical to the same command at `origin/main`**
  in a clean worktree (diff of the sorted error lists is empty). All 12 are
  `Cannot find module '../../event-engine-common/...'` / consequent implicit-`any` in
  `image.service.ts` + `bitdex-stats.ts` — a sibling repo not present in this environment.
  **Zero introduced.**
- `pnpm test:unit:run` on the two new suites — `Test Files 2 passed (2)`, `Tests 17 passed (17)`.
- `pnpm test:unit:run` **in full**, run at HEAD and again at `origin/main` in a clean worktree:

  | | `origin/main` | this branch |
  |---|---|---|
  | Test Files | 91 failed / 880 passed / 1 skipped (972) | 91 failed / 882 passed / 1 skipped (974) |
  | Tests | 4 failed / 13880 passed / 20 skipped (13904) | 4 failed / **13897** passed / 20 skipped (13921) |

  The sorted lists of failing **files** are byte-identical between the two runs (empty
  `diff`), and the 4 failing tests are the same `event-engine-common is checked out` /
  `... resolves` assertions in both — the sibling repo missing from this environment, the
  same cause as the typecheck errors. `+2` files and `+17` tests is exactly this PR's
  additions. **Zero regressions.**
- The suites that wholesale-mock `~/utils/storage-resolver` were checked individually, since
  `model-file.service` now imports a new name from it and a mock missing that key would make
  a suite fail to collect (reporting "no tests" rather than a failure): `remove-old-drafts`
  (5), `set-model-minor` (21), `model-locked-properties` (22), `model-early-access-refund`
  (11), `model-flag-side-effects` (22), `model-version.purge-by-hash` (8),
  `model-version.deregister` (4) and `model-file.service` (15) all collected their usual
  non-zero counts and passed.
- `eslint` on the four touched files — 3 errors + 1 warning, **identical to the baseline**
  for the same files at `origin/main` (pre-existing `no-module-scope-cache` and two
  `no-empty-function` in `model-file.service.ts`, and the `as any` header read that the
  existing batch test also has). Zero introduced.

CI on this PR: Unit tests, App unit tests, Package unit tests, ESLint + Prettier, Schema
drift gate, `event-engine-common` pin, `tekton / typecheck`, and the preview's deploy /
render-check / unit-tests / auth-tests all **pass**. The Tekton typecheck is the
authoritative one here — it runs in an environment that *has* `event-engine-common`, so it
sees none of the 12 local errors above.

`preview / component-tests` is **red, and pre-existing** (it is also report-only /
non-blocking). It is a single failure —
`src/components/Sticker/StickerPlacementBar.browser.test.tsx > fetches placements for a
signed-in viewer even though reveal hides them`, `Test Files 1 failed | 137 passed`,
`Tests 1 failed | 1417 passed`. Control: **#3864 fails the identical test**, same file,
same test name, on a preview that ran hours before this branch existed. Nothing in this PR
touches `src/components`.

https://claude.ai/code/session_017ETv5R7Sq5G7t5i4WTa6ea

### DIFF
--- a/src/server/services/__tests__/model-file.deregister.service.test.ts
+++ b/src/server/services/__tests__/model-file.deregister.service.test.ts
@@ -149,6 +149,31 @@ describe('deleteFile — file_locations deregistration', () => {
     expect(mockDeleteModelFileObject).not.toHaveBeenCalled();
   });
 
+  it('does NOT await the deregister — a hung resolver must not stall the delete', async () => {
+    // 🔴 THE NON-BLOCKING CONTRACT, PINNED. An audit showed that adding `await`
+    // in front of deregisterFileLocationsByFile left the whole suite green:
+    // every other case resolves its mock immediately, so nothing could tell
+    // "fire-and-forget" from "awaited". A future edit could therefore add up to
+    // the full request timeout of user-facing latency to EVERY file delete
+    // against a hung resolver, with no test objecting.
+    //
+    // Hold the deregister open and require deleteFile to resolve anyway.
+    let release: (() => void) | undefined;
+    mockDeregisterFileLocationsByFile.mockImplementationOnce(
+      () =>
+        new Promise((resolve) => {
+          release = () => resolve({ deleted: 1 });
+        })
+    );
+    stubDeletableFile();
+
+    await expect(deleteFile({ id: FILE_ID, userId: USER_ID })).resolves.toBeDefined();
+
+    expect(mockDeregisterFileLocationsByFile).toHaveBeenCalled();
+    expect(release).toBeDefined(); // it really was still pending
+    release?.();
+  });
+
   it('does not deregister when the delete is refused before the DELETE runs', async () => {
     // Training data on a still-draft model is refused up front — nothing was
     // removed, so nothing may be deregistered.
--- a/src/server/services/__tests__/model-file.deregister.service.test.ts
+++ b/src/server/services/__tests__/model-file.deregister.service.test.ts
@@ -170,7 +170,11 @@ describe('deleteFile — file_locations deregistration', () => {
     await expect(deleteFile({ id: FILE_ID, userId: USER_ID })).resolves.toBeDefined();
 
     expect(mockDeregisterFileLocationsByFile).toHaveBeenCalled();
-    expect(release).toBeDefined(); // it really was still pending
+    // NOTE on what this does and does not pin: deleteFile resolving while the
+    // deregister is still pending proves it is not awaited UNBOUNDEDLY. A
+    // bounded wait (await Promise.race([sleep(50), deregister()])) still passes
+    // — that shape was checked and does survive. The unbounded case is the one
+    // the design forbids, and the mutant that adds a bare `await` dies here.
     release?.();
   });
 
--- a/src/server/services/__tests__/model-file.deregister.service.test.ts
+++ b/src/server/services/__tests__/model-file.deregister.service.test.ts
@@ -1,0 +1,162 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Verifies deleteFile's post-commit cleanup: deleting ONE file now deregisters
+// that file's storage-resolver registry row, keyed on the file id the DELETE
+// actually removed. The version-keyed deregistration cannot cover this case —
+// the model version survives a single-file delete, so nothing version-keyed ever
+// reaches the entry and it outlives the file it describes. Deregistration is
+// best-effort: a failure must not fail the (already-committed) file delete, and
+// must not disturb the object cleanup or the cache bust beside it.
+
+const { mockDbWrite } = vi.hoisted(() => ({
+  mockDbWrite: {
+    modelFile: {
+      findFirst: vi.fn(),
+      findUnique: vi.fn(),
+      findMany: vi.fn(),
+      update: vi.fn(),
+      count: vi.fn(),
+    },
+    $queryRaw: vi.fn(),
+  },
+}));
+vi.mock('~/server/db/client', () => ({ dbRead: mockDbWrite, dbWrite: mockDbWrite }));
+
+// model-file.service builds a cached object at import (filesForModelVersionCache).
+// Hand back a STABLE stub so the test can assert the cache bust actually fired.
+const { mockCacheBust } = vi.hoisted(() => ({ mockCacheBust: vi.fn() }));
+vi.mock('~/server/utils/cache-helpers', () => ({
+  createCachedObject: () => ({ bust: mockCacheBust, fetch: vi.fn(), lookupFn: undefined }),
+}));
+vi.mock('~/server/cloudflare/client', () => ({ purgeCache: vi.fn() }));
+vi.mock('~/server/redis/client', () => ({
+  REDIS_KEYS: { CACHES: { FILES_FOR_MODEL_VERSION: 'files-for-model-version' } },
+}));
+vi.mock('~/server/logging/client', () => ({ logToAxiom: vi.fn(() => Promise.resolve()) }));
+
+const { mockDeleteModelFileObject, mockDeregisterFileLocationsByFile } = vi.hoisted(() => ({
+  mockDeleteModelFileObject: vi.fn(),
+  mockDeregisterFileLocationsByFile: vi.fn(),
+}));
+// Narrow overrides that keep every other export of these modules real — a
+// wholesale factory silently disables the whole suite the day the module grows
+// an export something else in this graph imports.
+vi.mock('~/utils/s3-utils', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  deleteModelFileObject: mockDeleteModelFileObject,
+}));
+vi.mock('~/utils/storage-resolver', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  deregisterFileLocationsByFile: mockDeregisterFileLocationsByFile,
+}));
+
+import { deleteFile } from '~/server/services/model-file.service';
+import { ModelStatus } from '~/shared/utils/prisma/enums';
+
+// Pairwise distinct so a test can never pass by picking up the wrong id.
+const FILE_ID = 123;
+const VERSION_ID = 4242;
+const MODEL_ID = 7;
+const USER_ID = 55;
+const FILE_URL = 'https://files.example.com/model/7/a.safetensors';
+
+function stubDeletableFile({
+  type = 'Model',
+  modelStatus = ModelStatus.Published,
+  rows = [{ modelVersionId: VERSION_ID, modelId: MODEL_ID, url: FILE_URL }],
+}: { type?: string; modelStatus?: ModelStatus; rows?: unknown[] } = {}) {
+  mockDbWrite.modelFile.findFirst.mockResolvedValue({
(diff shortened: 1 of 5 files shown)


ZacxDev:
Unblocked: the storage-resolver `fileIds` endpoint is merged **and deployed**.

Verified live against the running service (not just the rollout status):
- all 3 pods rolled to a new image digest, Ready, 0 restarts
- `{"fileId": <known-absent-id>}` → `200 {"success":true,"deleted":0}` — the previous build returns `400 modelVersionId is required` for that exact body, since the old decoder drops the unknown key
- `{"fileId": …, "modelVersionIds": […]}` → `400` mixed-key rejection, a path that only exists in the new build

Ready for review. One thing a reviewer should know: the deregister call here is best-effort and swallows failures by design, so it cannot surface a regression on its own — the server-side signal is `storage_resolver_deregister_requests_total` / `_rows_deleted_total{key="file"}`, and there is now an alert on a sustained non-success ratio.

ZacxDev:
**Unblocked** — removing the `[BLOCKED …]` prefix from the title.

The precondition this PR was waiting on has shipped: the storage-resolver
`fileId`/`fileIds` deregister path is merged and deployed, and was verified
against the running service (not just the rollout status) — a request keyed by
`fileId` is accepted and returns a normal zero-delete result, which the previous
build rejected outright.

Ordering was deliberate and is now satisfied: this client is best-effort and
swallows failures, so merging it before the endpoint existed would have made
every call fail silently — inert while reading as shipped.

Ready for review. The one red check (`preview / component-tests`) is
report-only and pre-existing — the same test fails on PRs created before this
branch existed. The authoritative `tekton / typecheck` passes.

ZacxDev:
## Adversarial audit — result, and three fixes (`4ea055571b`)

Audited blind. **The destructive path is correct**, and that was established
structurally rather than assumed:

- `deleteFile` has **exactly one** runtime caller; the `replaceFileId` flow uses
  `markFileReplaced`, not delete
- `file_id` is **UNIQUE** in the registry, so `DELETE … WHERE file_id = ANY($1)`
  cannot reach a sibling file's row — **over-deletion is structurally
  impossible**
- fires **post-commit**, gated on the `DELETE … RETURNING` row; no caller wraps
  it in a transaction, so nothing can roll the delete back underneath it
- empty array short-circuits with no request; chunk size is well under the
  server cap
- no configuration values leak into this public repo — the test fixture URL
  differs from the real one

No over-deletion or wrong-id path was found. What the audit did find were three
gaps around **silence** — the exact failure mode this feature exists to fix. All
three were reproduced before being fixed.

**1. The fire-and-forget contract was unpinned.** Adding `await` in front of the
deregister left the suite **fully green** — every fixture resolves its mock
immediately, so nothing could tell fire-and-forget from awaited. A future edit
could add up to the full 30s timeout of user-facing latency to *every* file
delete against a hung resolver, with no test objecting. Now pinned by holding
the promise open and requiring the delete to resolve anyway.

**2. The timeout value was unpinned.** The assertion was
`signal instanceof AbortSignal`, which any timeout satisfies —
`AbortSignal.timeout(30_000)` could become `timeout(1)`, aborting every call
before the resolver can answer and making the feature inert, with a green suite.
Now asserts the argument.

**3. The new path was indistinguishable from the version-keyed batch in Axiom** —
identical event names *and* payload shape, and success is never logged. So a
wholly inert deployment looked exactly like a failing bulk version deregister.
All three events now carry `key: 'file'`, mirroring the resolver's `key="file"`
metric label.

🔴 **Each of the three log sites needed its own assertion.** A per-event mutation
showed that covering two of three left the third free. The *skipped* event
matters most — a pod missing the configuration emits it on every delete, which
is the likeliest inert-deploy signal.

### Verification

Every new assertion was watched **failing** against the mutant it was written
for; dropping `key` at any one of the three sites now fails exactly one test.
**34 tests pass** across the four related suites.

### Notes for whoever merges

- **Confirm success server-side, not from app logs** — this path is silent on
  success by design (it is a per-delete hot path). Use
  `storage_resolver_deregister_rows_deleted_total{key="file"}`.
- The resolver's file-keyed path is **fail-closed** on its backup table: if that
  table were missing or column-drifted, every per-file deregister would 500, the
  client would swallow it, and the fix would be inert while reading as done. I
  checked directly — the table has the expected shape (registry columns + 2,
  none missing), so this holds today.
- One pre-existing shape worth knowing, not introduced here: an unguarded cache
  bust sits between the committed delete and both cleanups, so if it rejects,
  neither the object cleanup nor the deregister runs.
- The red `preview / component-tests` check is believed pre-existing and
  report-only — this PR changes no component files, and several recently merged
  PRs show the same failure. That is corroboration, not proof.

PR #3867, opened 2026-08-13
fix(model-file): give each caller its own `files` array from the version-file cache

ZacxDev:
> **Updated after an adversarial audit** (commit `da3193c514`). The fix moved: the array mutation is now fixed **at the call site** in `getModelsWithVersions`, the cache object is **un-exported** so the accessor cannot be bypassed, the masking `as typeof records` cast is gone, and the degraded sharing window is now **driven for real** in `cache-helpers-failopen.test.ts` rather than reconstructed from a mock. The **Fix** and **Tests** sections below describe the FIRST pass and are superseded — see the summary comment on this PR for the current state, the red/green matrix and the full mutant battery.

## Problem

`getModelsWithVersions` appends a linked VAE file by mutating the array it got from the version-file cache:

```ts
const files = groupedFiles[version.id]?.files ?? [];
files.push(...vaeFile);
```

The batched per-id cache layer only ever **shallow-clones** a record before handing it out, and says so explicitly: *"shallow only protects TOP-LEVEL fields — nested refs are shared … consumers MUST treat returned values as read-only for nested fields."* So the contract offered is read-only-for-`files`, and this call site writes to it.

## Is it actually reachable?

Yes — in one specific, real state.

- **Healthy Redis path: safe.** Every hit msgpack-decodes a fresh object per read, and this cache sets no in-process `localTtl`, so nothing is shared between requests. The mutation also cannot reach the stored value: the write to Redis is serialized and awaited *before* `fetch` returns.
- **Fail-open degraded path: the hazard is live.** When a cluster read rejects, `fetch` falls open to an origin lookup guarded by a per-id single-flight. Every caller that joins the in-flight lookup for a given version id resolves the **same** record object, and the shallow clone made afterwards leaves all of them holding the **same `files` array**. One request's VAE append is then visible to every other request sharing that flight.

Verified with a throwaway probe against the real cache builder, run in three arms with a positive control:

| arm | config | result |
|---|---|---|
| A | `localTtl` unset (the live config) | second fetch sees 1 file — **not shared** |
| B | `localTtl` set (**positive control**) | second fetch sees 2 files — harness can observe pollution |
| C | Redis read throws (degraded) | 1 lookup, `a.files === b.files` is `true`, caller B sees caller A's push |

Arm B is what makes arm A's clean result meaningful rather than a probe wired to nothing.

## Blast radius

**Corrected — the original text here understated this, and it was wrong in the PR's own favour.**

The claim was: *"the sharing window is one in-flight degraded fetch, not the cache TTL … Nothing is ever written back to Redis."* That is true **of the file cache only**. The polluted value does not stay inside that window, because a downstream cache stores the shaped body built from it.

`GET /api/v1/models/:id` writes its shaped 200 body — `files` included — into a **Redis-backed origin response cache** (`src/pages/api/v1/models/[id].ts`: `PUBLIC_MODEL_RESPONSE_TTL = CacheTTL.sm` = **180s** logical, stored at `ttl * 2` physical, enabled when `IS_DATAPACKET` and the request is not block-scoped), and Cloudflare fronts the same endpoint at `s-maxage=300` with stale-while-revalidate. So if the leak lands in the body that gets stored:

- it **outlives the in-flight window by minutes**, not milliseconds, and
- it is served to **every** anonymous reader of that `(modelId, browsingLevel)` pair for the remainder of the TTL — the poisoned response is shared, not per-request.

Two things about the write path are still accurate and worth keeping: nothing is written back to the **file** cache, and the single-flight entry itself is deleted the moment the lookup settles. The correction is that "the file cache is not poisoned" is not the same claim as "the pollution is short-lived", and only the first one was ever established.

Within the window the leak also crosses request boundaries into consumers that never add VAE files at all — the generation-resource file hydration and the `modelFile.getByVersionId` handler — which would see a VAE-typed entry appear in a version's file list.

Mutating consumer (3 endpoints, all through `getModelsWithVersions`):
- `GET /api/v1/models/:id`
- `GET /api/v1/models`
- `GET /api/v1/blocks/models`

Non-mutating readers of the same cache entries (the victims): the generation resource-file lookup, and the `modelFile.getByVersionId` tRPC query.

## Fix

Copy the array once in `getFilesForModelVersionCache` — the single accessor all three consumers already go through — rather than at each call site. Array-level is the right depth: no consumer mutates an individual file object, only the list. The copy is guarded (`Array.isArray`) so a record without a `files` array keeps its existing shape instead of throwing on a public-API read path. Comments at both ends explain why, so the next reader doesn't re-derive the suspicion or "simplify" the copy away.

The **other** `files.push(...vaeFile)` in `model.controller.ts` does **not** share the hazard: its `files` comes from a Prisma read rather than this cache, and that handler is currently unwired. Left alone deliberately.

## Tests

New regression block drives the cache mock to return records sharing one nested array — exactly the degraded shape — and asserts the accessor isolates the caller.

| | result |
|---|---|
| pre-change code | **4 failed** / 17 passed (21) |
| post-change | **21 passed** (21) |

Mutation battery on the guard (all 21 tests ran in every arm):

| mutant | outcome |
|---|---|
| unmutated control | 21 passed |
| mutate-in-place restored (`return fetch(...)` verbatim) | killed — 4 failed |
| same, with the explanatory comment left in place | killed — 4 failed |
| copy too shallow (top-level clone only) | killed — 4 failed |
| guard inverted (alias kept when `files` *is* an array) | killed — 5 failed |
| only the first record of a batch copied | killed — 1 failed |
| copy computed then discarded | killed — 4 failed |
| over-copy (each file object cloned too) | **survived — semantically stronger, not a gap** |

Gates: `pnpm typecheck` 0 errors (negative control: a planted type error in this file produced exactly 1 `error TS2322`, rc=2). `pnpm test:lint-rules` 220/220. ESLint on the three changed files reports only pre-existing issues on untouched lines (negative control: a planted unused var in the test file was reported). Prettier flags both touched files, but they are already unformatted on `main` — no new formatting regression, and no unrelated reformat included.

### DIFF
--- a/src/server/services/__tests__/bust-public-model-response-cache.service.test.ts
+++ b/src/server/services/__tests__/bust-public-model-response-cache.service.test.ts
@@ -88,7 +88,9 @@ vi.mock('~/server/services/model.service', () => ({
   ingestModelById: vi.fn(),
   updateModelLastVersionAt: vi.fn(),
 }));
-vi.mock('~/server/services/model-file.service', () => ({ filesForModelVersionCache: {} }));
+vi.mock('~/server/services/model-file.service', () => ({
+  deleteFilesForModelVersionCache: vi.fn(),
+}));
 vi.mock('~/server/logging/client', () => ({ logToAxiom: vi.fn() }));
 
 import { bustPublicModelResponseCache } from '~/server/services/model-version.service';
--- a/src/server/services/__tests__/model-file.service.test.ts
+++ b/src/server/services/__tests__/model-file.service.test.ts
@@ -18,8 +18,11 @@ vi.mock('~/server/db/client', () => ({ dbRead: mockDbRead, dbWrite: mockDbRead }
 // live redis connection — these official-file helpers only touch dbRead.
 // `lookupFn` isn't reachable through this stub, so the cache-filter test below calls
 // the exported `fetchModelFilesForCache` directly instead of going through the cache object.
+// `mockCacheFetch` is hoisted so the getFilesForModelVersionCache tests below can drive what
+// the cache hands back (the point of those tests is what the ACCESSOR does to that value).
+const { mockCacheFetch } = vi.hoisted(() => ({ mockCacheFetch: vi.fn() }));
 vi.mock('~/server/utils/cache-helpers', () => ({
-  createCachedObject: () => ({ bust: vi.fn(), fetch: vi.fn(), lookupFn: undefined }),
+  createCachedObject: () => ({ bust: vi.fn(), fetch: mockCacheFetch, lookupFn: undefined }),
 }));
 vi.mock('~/server/cloudflare/client', () => ({ purgeCache: vi.fn() }));
 vi.mock('~/server/redis/client', () => ({
@@ -32,6 +35,7 @@ import {
   markFileReplaced,
   restoreReplacedFile,
   fetchModelFilesForCache,
+  getFilesForModelVersionCache,
 } from '~/server/services/model-file.service';
 import { constants } from '~/server/common/constants';
 import { ModelFileVisibility } from '~/shared/utils/prisma/enums';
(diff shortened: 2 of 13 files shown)


ZacxDev:
## Audit rework — pushed as `da3193c514`

The first pass mitigated the symptom at the accessor and *inferred* the sharing window from the cache layer's source. This pass fixes the mutator, closes the bypass, and drives the real degraded path so the window is **observed**.

### 1. The mutation is fixed at the call site
`getModelsWithVersions` was doing `groupedFiles[version.id].files.push(...vaeFile)` on an array owned by the cache layer. The very next statement rebuilds the list with `.map()`, so nothing downstream ever wanted the mutation — it now builds a new array:

```ts
const files = [...(groupedFiles[version.id]?.files ?? []), ...vaeFile];
```

### 2. The accessor copy is kept — relabelled, not removed
With (3), every read routes through `getFilesForModelVersionCache`, so the copy is **forward-protection for the next consumer**, not a fix for a live mutator. Cost is one array copy per record, negligible against the msgpack unpack already on that path. The JSDoc now says exactly that instead of implying a live hazard.

### 3. `filesForModelVersionCache` is no longer exported
Its only external use was `.bust` from `model-version.service`, which `deleteFilesForModelVersionCache` already wraps — that call site now uses the wrapper. Grep confirms **no other module** imported the raw cache. Bypassing the accessor is now structurally impossible rather than discouraged by a comment, and a new test reads the module's actual **export namespace** (not the source text), so putting `export` back is red.

### 4. The masking cast is gone
`as typeof records` was load-bearing for nothing and disabled the only gate that would catch the returned shape changing. Measured both ways:

| accessor form | replace the returned record with `{ modelVersionId }` (drop `files`) |
|---|---|
| with `as typeof records` (before) | **0 type errors** — typechecks clean |
| `typeof records`-annotated loop (after) | **1 type error** — `TS2741: Property 'files' is missing` |

### 5. The degraded window is now OBSERVED, not inferred
The accessor tests mock `createCachedObject` wholesale and hand-construct the aliasing — they would pass against a completely broken cache and fail against a fixed one. A new **H2b** block in `cache-helpers-failopen.test.ts` drives the real thing end to end: rejecting `mGet` → real fail-open → real per-id single-flight → real `lookupFn` → real accessor.

The join is asserted, not assumed: **one** origin lookup serves both callers, and both take the degraded path. Its first test is a **positive control** proving the raw layer today hands both callers the *same* nested array — without it, `not.toBe` passing would be indistinguishable from a probe wired to nothing.

**Red/green, both suites, at base `f19574a9cb` vs HEAD:**

| | result |
|---|---|
| base (pre-change source for both files) | **7 failed / 29 passed (36)** |
| HEAD | **36 passed (36)** |

The two new H2b isolation tests are among the 7 red. The H2b *control* test passes in both arms — it is a control, and is labelled as one, not counted as regression coverage.

### 6. Honest relabelling of the `Array.isArray` guard
The guard is **unreachable with the real cache**: `lookupFn` always initialises `files: []`, and marker records are skipped inside the cache layer before they can be returned. Its test is therefore an **invariant guard against a mock-only shape**, not regression coverage — and it is the *sole* kill for the guard-removal mutant, meaning one battery arm scores against a state that cannot occur. Renamed and documented as such. The guard itself is kept: a future lookupFn/cache change could make the state reachable, and throwing on a hot public-API read path is the worse failure.

### Post-change mutant battery
A fix round resets the gate, so the whole battery was re-run against the final tree (both suites, 36 tests, every arm collected all 36 — never an exit code):

| mutant | outcome | passed/failed |
|---|---|---|
| copy removed (accessor returns `fetch(...)` verbatim) | killed | 30 / 6 |
| copy too shallow (`{ ...record }` only) | killed | 30 / 6 |
| copy at the wrong level (outer map cloned, `files` shared) | killed | 30 / 6 |
| guard inverted | killed | 29 / 7 |
| guard text present but DEAD (`&& false`) | killed | 30 / 6 |
| guard removed entirely (unconditional spread) | killed | 35 / 1 |
| only the first record of a batch copied | killed | 35 / 1 |
| un-export reverted | killed | 35 / 1 |
| **control (known-equivalent):** `slice()` for spread | **survived — required** | 36 / 0 |
| **control (known-breaking):** `files: []` | killed | 30 / 6 |
| call-site `push` restored | **survived** | 36 / 0 |

Both controls behaved as they must — the equivalent one survived (so the harness is not scoring on noise) and the breaking one died.

**On the one survivor:** restoring the call-site `push` is now **demonstrably equivalent**, because the accessor copy makes the array the caller's own — pushing to it is harmless. That is the defence-in-depth working, and it is also a real statement about coverage: there is no behavioural test pinning the call-site form on its own. The compound mutant that matters (copy removed **and** push restored — i.e. the original bug) is killed, because removing the copy alone is killed. Left as a known, stated gap rather than papered over with a source-text assertion.

### Gates
- `pnpm typecheck`: **0 errors**. Negative control in a file it actually reads (`pnpm typecheck` excludes `src/**/__tests__/**`, so a planted error in the new test file would be invisible): the item-4 mutation above produced exactly **1** `error TS`, rc=2.
- Unit tests: **494 passed / 494** across **34** suites — the two suites above plus every suite importing `model.service` or `model-version.service`, plus the lint-rule suite.
- ESLint on all 12 changed files: **identical before and after** — 27 problems / 7 errors / 20 warnings, all pre-existing on untouched lines.
- Prettier: **0 new violations**, measured as a per-file *reformat-hunk delta* against the previous tip rather than a pass/fail on files that are already unformatted. (This caught 4 real new violations from over-long mock lines, which were then fixed — the instrument was watched going red before its green was believed.)

### Follow-ups filed
- civitai/civitai#3872 — the **root cause in the shared cache layer**: the degraded path's top-level-only clone. Fixing it there protects every consumer; it is a wide-blast-radius change and deserves its own PR and audit. This PR is the consumer-side mitigation.
- civitai/civitai#3873 — a **possible twin** in generation data, filed explicitly as an unverified lead.

### Correction to this PR's own description
The **Blast radius** section claimed the sharing window is "one in-flight degraded fetch, not the cache TTL" and that "nothing is ever written back to Redis". That is true of the *file* cache only, and it **understated this PR's own case**. `GET /api/v1/models/:id` writes the shaped body — `files` included — into a Redis-backed origin response cache for 180s, behind a Cloudflare edge at `s-maxage=300`. A polluted body therefore outlives the in-flight window by minutes and is served to every reader of that model id for the rest of the TTL. The description has been corrected in place.

ZacxDev:
Final round — two items, both closing gaps the previous round left. `da3193c514` → `ebd635c1b1`.

**1. Closed the circular-protection trap (added the coverage, did not fall back to a comment).**

The delta re-audit was right that restoring the `push` at `model.service.ts:3873` survived the entire suite (M08). That left the two halves of this PR justifying each other in a circle — the accessor JSDoc says the copy is forward-protection *because* the call site was fixed, while the call site is safe-if-reverted only *because* the copy exists. Anyone landing the deep-clone in #3872 would reasonably delete the copy, and at that moment nothing would pin the call site either.

New suite `model.service.vae-append-no-mutation.test.ts` drives the **real** `getModelsWithVersions` (same import-the-real-`model.service` scaffold as `get-models-raw.transient-503`) and stubs `getFilesForModelVersionCache` to hand back the **raw cache-owned array** — i.e. the accessor copy *removed* — so the only thing that can keep that array intact is the call site itself.

M08 red/green, measured against the final version of the file:

| | result |
|---|---|
| unmutated | `2 passed (2)` |
| `push` restored | `1 failed \| 1 passed (2)` |

```
AssertionError: expected [ 'base.safetensors', 'linked.vae.safetensors' ]
                to deeply equal [ 'base.safetensors' ]
```

The **positive control** test (the linked VAE really *is* appended to the returned version) stays green in the red run, so the failure is attributable to the mutation rather than to a broken mock chain. `model.service.ts` is byte-identical to its pre-mutation sha afterwards.

One assertion is deliberately **absent**: that the *returned* files array differs from the cached one. That value is the output of a `.map()`, so it is freshly allocated on every path and the assertion passes with the mutant restored (measured) — a vacuous guard. That is noted in the file so it does not get re-added. The accessor JSDoc now states the two halves are independent, and cross-references #3872.

**2. Fixed the 3 pre-existing `TS2345` in `model-file.service.test.ts` (310/330/360).**

Introduced by `f54559d68d` and invisible to `pnpm typecheck` because `tsconfig.json:70` excludes `src/**/__tests__/**`. Fixed with a narrow `(x as unknown[])` on the push **receiver** — the same pattern as the H2b block in `cache-helpers-failopen.test.ts` — not `any`, no blanket `@ts-ignore`, so every other read of `files` in the file stays fully typed.

Instrument: the repo's own `tsconfig.json` with **only** that one exclude entry removed, so the ambient/globals graph stays in the program (`TS2304` count 0 — no generic-constraint collapse). Validated with a negative control planted in that very file (`TS2322` at line 387, reported, reverted). Repo-wide `__tests__` error count **804 → 801**; diffing the two error sets on *location + error code* (message text reorders union members between runs, which makes a naive text diff look like ~58 new errors) yields exactly the 3 target lines removed and **zero** new errors anywhere.

**Gates**

- Four original suites unchanged: **53 passed (53)**. With the new suite: **55 passed (55)** across 5 files, 0 timeout panics.
- `pnpm typecheck`: **0 errors**, validated with its own negative control in `model-file.service.ts` (`TS2322` at 657, reverted).
- eslint clean on the new file; the 3 errors on `model-file.service.ts` are pre-existing (module-scope cache + 2 empty arrow functions).
- Prettier: new file **0** reformat hunks; both edited files' pre-existing hunk counts (4 and 3) *and* hunk content are byte-identical to `HEAD` — no wholesale reformat.

Untouched, as scoped: the accessor's copy logic, the un-export, the call-site expression, and `packages/civitai-redis`. Not merging.

PR #3868, opened 2026-08-13
PROPOSAL (do not merge): pnpm typecheck is structurally blind to src/**/__tests__/ — 801 errors, and a ratchet to stop the inflow

ZacxDev:
**This is a proposal, not a fix. Please do not merge it as-is** — it is here to be argued with. Nothing existing is modified; the PR adds four files.

Full write-up: `claudedocs/typecheck-tests-gap-2026-08-12.md`.

## The gap

`tsconfig.json` excludes `src/**/__tests__/**`. `pnpm typecheck` is therefore not *lenient* about test files, it cannot see them at all. With a deliberate `const x: number = 'not a number'` planted in a test file, on that exact tree:

```
$ pnpm typecheck
typecheck: OK — 0 type errors in 13s (heap cap 8192 MB).
exit=0     error TS lines: 0     mentions of the planted file: 0
```

## The verified number

Measured against `main` @ `f19574a9cb`, reproduced on three independent runs:

- **801 type errors in 141 files**, out of **870** `.ts`/`.tsx` files under `src/**/__tests__/`
- **729 (83.8%) of test files are already clean**
- **0** of the 801 land outside a `__tests__/` directory, and **0** are `TS2304` — the measurement config contributes no artifacts of its own (it `extends` the real `tsconfig.json` and removes exactly one `exclude` entry; a hand-rolled `include` drops `src/types/global.d.ts` and produces a *false all-clear*)
- Heavily concentrated: **50 files carry 82%** of it
- One is a genuine runtime bug the exclusion hides: a suite imports `kysely`, which is not a dependency and is not installed, so it contributes zero tests
- ~41 (5.1%) are `TS2737` BigInt-literal errors from the app config's `target: ES2018`, not test-logic defects

## Why the exclusion exists

Not perf. The line was added **exactly once**, in `1a7f89df98` ("fix broken automatic metadata parsing", 2026-02-26), in a commit about neither. Cost of including tests, measured back-to-back: **+64 s cold, +5 s warm.**

> 🔴 **Corrected 2026-08-13.** This section previously said that commit added "the repo's first `src/**/__tests__/` file" and that the exclusion was never really a decision. **Both are false.** `git ls-tree -r 1a7f89df98^` shows **eight** such files already present, the earliest **2026-01-23** — over a month earlier, across six commits by different authors. The likelier reading, stated as **inference not fact** (the commit carries no comment or discussion): the typecheck was **already red** with those files in the program, and the one-line exclude was a deliberate, undocumented **silencing**. That *strengthens* the recommendation below — "delete the exclude and require green" is not untried, it is an option this repo effectively **rejected once already**, under exactly the pressure that makes people reject it.

## The finding that should drive the decision

Grouping the 141 dirty files by when they were first added:

| first added | dirty files | errors |
|---|---|---|
| 2026-02 … 2026-05 | 3 | 7 |
| 2026-06 | 39 | 316 |
| 2026-07 | 59 | 248 |
| 2026-08 (12 days) | **40** | **230** |

**This is not legacy sediment.** 794 of 801 errors are under three months old, arriving at roughly 19 errors and 3 dirty files per working day. The value of a gate here is almost entirely in stopping the inflow.

## Recommendation

A **per-file ratchet** — the same shape this repo already runs for schema drift. A committed baseline of per-file counts; a new file with errors, or a baselined file that gets worse, **blocks**; anything that improves passes. Then burn down the top 50 files (82%) behind it. Alternatives (non-blocking report / delete-the-exclude-and-require-green / changed-files-only / fix-outright) are weighed with their trade-offs in §4 — including why a permanently-red gate would be worse than none.

Stated cost: per-file counts are noisier than a single total and will occasionally block a PR that did not open the file in question. That is the price of a gate that cannot silently absorb a new violation.

## Proof of concept — three runs, counted from logs, never from exit codes

| run | result |
|---|---|
| (a) unmodified `main` | `PASS` — 801 errors / 141 files, matches baseline; positive control 938 test files in program |
| (b) one error planted in a previously-clean test file | `BLOCKED` — 802 / 142, names the file; meanwhile `pnpm typecheck` says **"OK — 0 type errors"** |
| (c) reverted | `PASS` — back to 801 / 141 |

The gate refuses to produce a verdict until **five structural controls** pass. Exit codes: `0` pass, `1` regression, `2` bad baseline, `3` cannot measure.

### 🔴 Rework (2026-08-13) — the instrument could report a confident zero

An adversarial audit returned *needs rework* on the gate itself: this PR's whole thesis is that a checker must not be able to report a confident zero, and this one could. All four holes shared a root cause — **"I parsed no diagnostics" was treated as evidence of cleanliness.** Each was reproduced live against this repository and re-run after the fix:

| condition | before | after |
|---|---|---|
| `TYPECHECK_HEAP_MB=abc` (wrapper exits 2, prints no crash marker) | `PASS`, exit 0 | exit 3, names the exit status |
| wrapper exits 127 (binary missing) | `PASS`, exit 0 | exit 3 |
| wrapper exits 134 (V8 heap abort) | `PASS`, exit 0 | exit 3 |
| tsc `pretty` output — **801 real diagnostics** | `0 error(s) across 0 file(s) … PASS`, exit 0 | `801 error(s) across 141 file(s)` |
| `--write-baseline` while the check cannot run | writes `0 error(s) across 0 file(s)`, exit 0 | exit 3, file untouched |

The old "CRASH IS NOT CLEAN" control was **spelled, not structural** — it matched the literal string `TYPECHECK CRASHED`, which `scripts/typecheck.mjs` emits from exactly **one** of its four failure exits.

Also fixed: the `outsideTests` predicate now matches the **actual** exclusion (`src/**/__tests__/`), so the 68 `packages/*/src/**/__tests__` files — which are in *both* programs — are no longer claimed here; the positive-control floor is derived from a test-file count stored in the baseline (938 → 844) instead of a magic 400 that let **57%** of the test tree vanish with the control still green; and a config-drift check asserts `tsconfig.tests.json` is the base `exclude` list minus **exactly one** entry.

**A new 78-case unit suite** (`scripts/__tests__/typecheck-tests-gate.test.ts`, runs in the existing `unit` project) covers the branches the three runs never touched: a baselined count rising, falling, going clean, deleted, **renamed**; two files sharing a basename; empty / malformed / absent / wrong-config baselines; both diagnostic formats; every cell of the exit-status truth table. **Thirteen mutations — one per guard — were each confirmed to turn it red for that guard's own cases.**

*(Documented behaviour: a pure `git mv` still **blocks** — the gate cannot see file content, so auto-forgiving "looks like a rename" is a hole a real regression fits through. It now names the probable former path instead of reading as a mystery regression.)*

**Measurement re-verified after the rework: unchanged at 801 / 141**, per-file map identical to the committed baseline, and `--write-baseline` byte-stable across two runs.

## Open questions for review

1. Stage-one wiring (parallel check, one extra `tsc` run) or straight to stage two (one run serving both checks, since the tests program is a strict superset)?
2. Per-file counts, or a coarser gate that trades precision for less friction?
3. Should the two-populations inconsistency be fixed as well — 242 `*.test.ts` files under `src/` and 68 under `packages/*/src/**/__tests__/` **are** already typechecked, so today the guarantee depends on which directory a test landed in.


ZacxDev:
## Audit rework — the instrument could report a confident zero, plus a history correction

Pushed `991df0e0ba` (branch tip was `b85b0d82b9`). No force-push, no rebase; the original commit stands and this is additive on top of it.

An adversarial audit reproduced the measurement fully — 801 / 141, committed baseline byte-identical to a fresh run, zero errors outside `__tests__/`, `TS2304` = 0 — and then returned **needs rework on the gate itself**. Fair: this PR's whole argument is that a checker must not be able to report a confident zero, and this gate could.

### The root cause was one thing in four places

`typecheck-tests-gate.mjs` treated **absence of parsed diagnostics as evidence of cleanliness**. Every condition below was reproduced live against this repo, then re-run after the fix:

| condition | before | after |
|---|---|---|
| `TYPECHECK_HEAP_MB=abc` (wrapper exits 2, prints no crash marker) | `PASS`, exit 0 | exit 3 — *"the typecheck exited 2 and produced 0 parseable diagnostics… FAILURE TO RUN"* |
| wrapper exits 127 (binary missing) | `PASS`, exit 0 | exit 3, names 127 |
| wrapper exits 134 (V8 heap abort) | `PASS`, exit 0 | exit 3, names 134 |
| tsc `pretty` output — **801 real diagnostics** | `0 error(s) across 0 file(s) … 141 file(s) now clean … PASS`, exit 0 | `parsed 801 diagnostic(s) in pretty format` → `801 error(s) across 141 file(s)` |
| `--write-baseline` while the check cannot run | wrote `0 error(s) across 0 file(s)`, exit 0 — **ratchet destroyed, success reported** | exit 3, baseline byte-unchanged |

The old "CRASH IS NOT CLEAN" control was **spelled, not structural**: it matched the literal string `TYPECHECK CRASHED`, which `scripts/typecheck.mjs` emits from exactly **one** of its four failure exits. The header comment asserting a crash could "never" read as 0 was false as written, and is corrected — a comment is a claim like any other.

**One correction to the audit itself, and it is worth reading.** The audit predicted the `pretty` case would false-PASS because "the wrapper still counts 801". It does not, *today*: TypeScript 5.9 always emits ANSI with `pretty`, and the escape codes land **inside** the token — `…^[[91merror^[[0m^[[90m TS18046:` — so the literal substring `error TS` appears **0 times** in the entire output. The wrapper's own counter is spelled identically to the gate's regex, so both went blind at once, the exit status disagreed, and the crash path fired. I confirmed the always-ANSI behaviour against a negative control across four environments (default, `NO_COLOR=1`, `FORCE_COLOR=0`, `TERM=dumb`).

That is **luck, not a control** — and it makes the hole worse, not better, because the wrapper then reported *"tsc terminated abnormally without producing any diagnostics"* about a run that produced 801. The moment `pretty` output arrives without color (a future tsc, a non-TTY policy change, any log pipeline that strips ANSI for readability), the token reappears, the wrapper counts 801, the exit passes through, and the gate's single-format regex yields 0. I demonstrated exactly that by feeding the real 801 diagnostics, ANSI-stripped, to both gate versions:

```
BEFORE:  0 error(s) across 0 file(s) (baseline: 801 across 141)
         141 file(s) now clean, 0 improved
         PASS — no new or worsened test-file type errors.        exit 0

AFTER:   parsed 801 diagnostic(s) in pretty format
         801 error(s) across 141 file(s) (baseline: 801 across 141)
         PASS — no new or worsened test-file type errors.        exit 0
```

Same verdict, but only one of them measured anything.

### Five structural controls now, none of them a string match

1. **Exit-status truth table** — the outcome is decided by *(exit status, signal, parsed count)*. A non-zero exit with zero parsed diagnostics is a failure to run, never a clean tree. One rule covering 2 / 127 / 134 and any future sibling.
2. **Parse control** — `--pretty false` forced on the command line (it beats the config), **both** formats parsed anyway, ANSI stripped first, and output containing `error TS` that parses to zero is refused outright.
3. **Plausibility** — a parsed total of 0 against a baseline of N > 0 is refused; escape is the explicit `TYPECHECK_TESTS_ALLOW_EMPTY=1`. "Everything got fixed" and "the instrument is broken" produce the same number, and only one of them is common.
4. **Positive control on the measurement arm** — now runs through the **same wrapper, same `-p`, same flags** as the verdict run. The old one shelled out to `tsc.js` directly with different arguments, so it structurally could not witness the arm it validated. Its floor is derived from a test-file count **stored in the baseline** (938 → 844), not a magic 400: at 400, **57%** of the test tree could disappear with the control green, and vanished files score `fixed` → **PASS**.
5. **Config-drift control** — `tsconfig.tests.json` must be `tsconfig.json`'s exclude list minus exactly one entry. Both lists are written out verbatim, so without this an addition to the base list silently fails to reach the measurement program and the one-line-delta property quietly stops being true.

`--write-baseline` runs behind all five, including borrowing the floor from the baseline it is about to replace.

Also fixed: the `outsideTests` predicate matched a bare `/__tests__/` substring while the root tsconfig excludes only `src/**/__tests__/**`. The 68 `packages/*/src/**/__tests__` files are in **both** programs, so a future error there would have redded `pnpm typecheck` *and* demanded a baseline entry here. Latent (0 today), now correct.

### The unit suite the last round explicitly deferred

`scripts/__tests__/typecheck-tests-gate.test.ts` — **78 cases**, running in the existing `unit` Vitest project (its `include` already covers `scripts/**/*.test.ts`), sub-second per case via a `TYPECHECK_TESTS_WRAPPER` seam driven with stub typecheckers instead of a multi-minute `tsc`. It covers the branches the three proof-of-concept runs never touched: a baselined count **rising / falling / going clean / deleted / renamed**, two files sharing a basename, **empty / malformed / absent / wrong-config** baselines, both diagnostic formats, and every cell of the truth table.

**Mutation-tested, not merely run.** Thirteen mutations — one per guard — each turned the suite red **for that guard's own cases**, with the tree checksum-verified pristine afterwards (78/78 before and after):

| mutant | guard broken | cases failed |
|---|---|---|
| M1 | non-zero exit + 0 diagnostics must not read as clean | 9 |
| M2 | a measured zero against a non-empty baseline is refused | 3 |
| M3 | pretty-format diagnostics are still counted | 3 |
| M4 | the file-count floor tracks the baseline, not a magic 400 | 4 |
| M5 | `packages/*/src/**/__tests__` is not this gate's business | 4 |
| M6 | a rename is reported as a rename | 2 |
| M7 | the two tsconfig exclude lists differ by exactly one entry | 3 |
| M8 | a baseline measured against another config is rejected | 2 |
| M9 | a run killed by a signal is refused | 1 |
| M10 | a coloured pretty run is still counted | 1 |
| M11 | `--write-baseline` refuses an implausible measurement | 1 |
| M12 | the gate acts on the run classification | 4 |
| M13 | `--write-baseline` borrows the floor from the baseline it replaces | 1 |

A guard no mutation can turn red is not a guard, and 78 green assertions say nothing on their own about which are load-bearing.

**Known, documented behaviour: a pure `git mv` blocks.** The new path has no baseline entry (→ BLOCK) and the old one disappears (→ scored `fixed`). The gate cannot see file *content*, so it will not auto-forgive "this looks like a rename" — that is a hole a genuine regression fits through. It now **names the probable former path** in the block output (basename + identical count; an ambiguous match makes no claim at all) so the author is told the remedy rather than hunting a regression that is not there.

---

## 🔴 History correction — §2 of the write-up was wrong

The write-up claimed `src/utils/metadata/__tests__/exif-parser.test.ts` was "**the first file the repository ever put in a `src/**/__tests__/` directory**", and that "nobody decided that test files should not be typechecked". **Both are false**, and I could not fix the commit message without a rewrite, so it is recorded here, in the doc, and in the PR body instead.

```
$ git ls-tree -r 1a7f89df98^ --name-only | grep -E '^src/.*/__tests__/'
src/env/__tests__/other.test.ts                                        2026-02-06  b2a862e9d2
src/server/games/daily-challenge/__tests__/challenge-helpers.test.ts   2026-02-13  768bb18136
src/server/games/daily-challenge/__tests__/winner-cooldown.test.ts     2026-02-12  9eb0d79003
src/server/jobs/__tests__/prepaid-membership-jobs.test.ts              2026-01-23  3f93c4a359
src/server/services/__tests__/challenge-review.service.test.ts         2026-02-13  47910c946e
src/server/services/__tests__/coinbase.service.test.ts                 2026-02-25  3968c33229
src/server/services/__tests__/redeemableCode.service.test.ts           2026-01-23  3f93c4a359
src/shared/utils/__tests__/creator-program.utils.test.ts               2026-02-17  f28ea5ea67
```

**Eight such files already existed** when the exclude line was added on 2026-02-26 — the earliest on **2026-01-23**, over a month before, across six commits by different authors. The convention was established and in use; the exclusion came *afterwards*.

What is still true: the `exclude` entry has been touched exactly once, its commit's subject is about neither the exclusion nor tests, and there is no comment, no linked discussion and no perf rationale anywhere.

**The likelier reading — and I am labelling this inference, not fact**, because the commit carries nothing recoverable about intent: with eight test files already in the program the typecheck was **already red**, and the one-line exclude was a deliberate, undocumented **silencing** of an existing failure — not someone unknowingly closing a door nobody had opened.

**That strengthens the recommendation rather than weakening it.** "Delete the exclude and require green" (option B) is not an untried idea whose consequences are unknown — it is an option this repository has effectively **rejected once already**, under exactly the pressure that makes people reject it: a red check standing between them and shipping. A ratchet is the shape that survives that pressure, because it is never red for work you did not cause.

---

## Re-verified after the rework

- **801 errors / 141 files**, per-file map **identical** to the committed baseline. The baseline gains one field (`testFilesInProgram: 938`) that the positive control's floor is derived from.
- `--write-baseline` run twice against the same tree → **byte-identical** (`cmp` clean).
- The three proof-of-concept runs, re-run with real counts:
  - **(a)** unmodified → `801 error(s) across 141 file(s)` … **PASS**, exit 0
  - **(b)** one error planted in `src/__tests__/node-version-consistency.test.ts` (0 in baseline) → `802 across 142`, **BLOCKED**, exit 1, names the file. On that same tree `pnpm typecheck` says **"OK — 0 type errors in 189s"**, exit 0, **0** `error TS` lines, **0** mentions of the file.
  - **(c)** reverted → `801 across 141`, **PASS**, exit 0
- `scripts/__tests__/` suites: **90 passed** (78 new + 12 pre-existing), no collateral.
- Counts throughout are read from captured logs, never from an exit code.

## Still out of scope, deliberately

No CI wiring — the authoritative typecheck is the in-cluster Tekton `pr-check` pipeline, which is a change in a different repo, and this stays a **proposal**. No attempt to fix the 801. `tsconfig.json`, `pnpm typecheck` and `target: ES2018` are untouched — the 41 `TS2737` BigInt entries stay baselined as documented config artifacts rather than test defects, because overriding `target` would destroy the one-line-delta property the whole measurement rests on.

**Please still do not merge** — it is here to be argued with.

ZacxDev:
## Hardening round: the four ways the new self-checks could still be defeated

A delta re-audit confirmed the previous round closed every 🔴 structurally — and then found four ways the **new** controls could be defeated. Each is demonstrated before/after against this repository, not a stub.

### F3 — the CI blocker: `TYPECHECK_TESTS_ALLOW_EMPTY=1` was a silent confident zero

**Before** (wrapper that measures nothing, hatch set):
```
typecheck-tests-gate: 0 error(s) across 0 file(s) (baseline: 7 across 2).
typecheck-tests-gate: 2 file(s) now clean, 0 improved — ...
typecheck-tests-gate: PASS — no new or worsened test-file type errors.     rc=0
```
Nothing anywhere records that a control was switched off. One env var in one job definition reinstates exactly the failure this gate exists to eliminate.

**After** — same command:
```
typecheck-tests-gate: REFUSING TO RUN — TYPECHECK_TESTS_ALLOW_EMPTY is set on a
VERDICT run. It is honoured on --write-baseline only, and is refused here rather
than ignored ... If you are seeing this in CI, some job definition is exporting
TYPECHECK_TESTS_ALLOW_EMPTY — remove it.                                   rc=3
```

**Why this shape** rather than shouting louder, or refusing outright:
- **Scope is the control.** Honoured on `--write-baseline` only, so it *cannot* make a verdict run green. This costs nothing: on the day the tree is genuinely clean the required action **is** to regenerate the baseline, after which the baseline total is 0 and the control never fires again. A verdict run that trips it is correctly telling you to do that regeneration.
- **Refused, not ignored.** Silently ignoring a control-disabling variable is its own confident-wrong: the operator thinks it is off, the gate thinks it is on, the log says nothing.
- **A reason, not a token.** `=1` is refused; the value must be a written justification, echoed into the output. A disarmed regeneration prints a banner and its success line reads `...WITH THE PLAUSIBILITY CONTROL DISARMED`.

### F1 — the plausibility floor was a cliff at exactly zero

`classifyRun({status:2, total:5, sawAnyErrorTs:801})` returned `ok:true` → a 99% undercount passed, printing "136 file(s) now clean". Fixed as an accounting invariant: **every** line carrying the marker must parse, not merely *some*.

I checked that strict equality is safe before relying on it, per the brief. Measured over the real corpus: **`unparsed === 0` across 801 diagnostics and 1451 output lines**, of which **650 are message-continuation lines** — those do not carry the `error TS` marker, so they never count. The divergent shapes that genuinely exist in tsc are the **fileless** diagnostics (TS18003 "No inputs were found in config file", TS6053), which are precisely the outputs meaning the program was not built as intended — so refusing on them is *correct*, not merely conservative. Refusals name the offending text verbatim, so a false positive is diagnosable rather than asserted.

A **collapse** (under 25% of baseline) is refused too — the zero was only the extreme of that spectrum, and the ratchet scores the whole spectrum as *progress*. Split by reversibility: **shouted** on a verdict run (it cannot hide a regression, so blocking would obstruct a genuine cleanup for no safety gain), **refused** on `--write-baseline` (there it rewrites the ratchet and destroys the evidence together).

### F2 — the derived floor had no lower bound

`testFileFloor(1)` → `{floor: 0, derived: true}`: a floor no program can fail, *while logging that it came from the baseline*. Replacing a magic number with a value read from a file made the control only as good as that value. Now `max(fixed 400, 90% of recorded)`. Missing/null falls back; **corrupt** (non-integer, negative, zero, absurd) is **refused** — an absent value and a corrupt one are different facts, and treating the second as the first is how a broken baseline reads as a working control. An absurd count is refused specifically because it would make the control *permanently red*, the failure mode that teaches people to delete controls.

### F5 — the drift control's wiring was untested

Mutating `if (!drift.ok)` → `if (false)` left the suite **78/78 green**. `diffExcludes` was well covered and the control was live, but nothing pinned that the gate *read* it. A `TYPECHECK_TESTS_CONFIG_DIR` seam now drives it end to end, with an **undrifted control arm** so a refusal attributes to the drift rather than to the fixture.

The sweep for siblings found **four more** unpinned call sites: the positive-control arm's own verdict, the unusable-floor refusal, the escape-hatch refusal, and the tsconfig read error. Each is pinned on its **specific message** — these sites are outcome-equivalent (all end in exit 3 via some later check), so an exit-code assertion would survive the mutation and prove nothing.

### Also
- `validateBaseline` cross-checks `totalErrors` against the sum of its entries.
- The three production env seams are **echoed** with `[OVERRIDE]` markers — a stub-driven run is no longer byte-identical to a real one.
- Doc counts reconciled by re-derivation (77/78, Twelve/Thirteen → the real numbers). Its history grep `^src/.*/__tests__/` cannot match `src/__tests__/setup.ts` — added by the same commit as two of the others — so the count is **nine**, not eight. Load-bearing claims unaffected (convention established; earliest 2026-01-23).

### Verification
- Suite **78 → 134 cases**, all green (counted from `--reporter=verbose` output, not an exit code).
- **19-mutant battery.** It found a real gap: `F5-floorresult-callsite` initially **survived**, because its case used `testFilesInProgram: 1` — a value the `Math.max` fix had made *valid*, so it exercised the floor comparison rather than the usable-floor refusal. Now killed by a dedicated case. Final run: **0 survivors, 0 false kills**, tree checksum-restored.
  One equivalent control was initially reported **KILLED** — and deserved to be: it renamed only a declaration, leaving `${pct}` dangling, i.e. a breaking mutation mislabelled as equivalent. The harness was right. Recorded in the doc rather than quietly corrected, because "the equivalent control died" is exactly the signal one is tempted to explain away.
- **Measurement unchanged** on the exact committed content: **801 errors / 141 files**, 938 test files in program, floor 844, and the committed baseline **byte-identical** to a fresh `--write-baseline` (same sha256). No count moved.
- **Three proof-of-concept runs re-run:** PASS 801/141 → planted error **BLOCKED 802/142 naming the file**, while `pnpm typecheck` on that same tree reports `OK — 0 type errors`, exit 0, with **0** `error TS` lines and **0** mentions of the file → PASS 801/141 on revert.

### Is it CI-safe now?

**In my judgement, yes — the gate itself is now safe to wire into CI.** The specific thing that was not safe is fixed: there is no longer any environment variable, baseline edit, or output format that turns a non-measurement into a green verdict. Every path that cannot measure exits 3 and says so; every disarmed path is out of scope for the verdict run entirely.

Two caveats a reviewer should weigh, neither of which is an argument against wiring it:
1. **Wiring remains out of scope here** — that is a separate cross-repo change against the Tekton `pr-check` pipeline, and this PR stays a proposal.
2. **The gate adds a full second `tsc` program** (+64s cold / +5s warm measured here), and it runs the wrapper **twice** (positive control + verdict). Budget for that before adding it to a per-PR job.

What I would *not* claim: that the 134 cases exhaust the ways this gate can be wrong. The battery found a survivor on its first run and a mis-specified control on its second, in a suite that was already green and already adversarially reviewed twice. That is the honest prior for round four.

Issue #3842, opened 2026-08-12
Challenge completion: side effects before a failure are not rolled back or recorded


JustMaier:
Found while exercising the completion-job wrapper during live testing of #3814. **Pre-existing — not caused by that PR.**

## What happens

Challenge completion is a sequence of steps with side effects, and there is no transaction or compensation across them. The collection is closed early (step 2, `closeChallengeCollection` — sets the collection `write` to `Private` and deletes its contributors). A completion that throws at a later step has therefore **already closed submissions**, permanently, while reporting failure.

So "the completion failed" does not mean "nothing happened". The retry an hour later — see the companion issue on challenges stuck in `Completing` — finds the collection already closed and the challenge in a partially-completed state.

## Why this is a report and not a bug fix

The *effect* is arguably correct: the challenge has genuinely ended, so closing submissions is the right thing and re-opening them on failure would let entries arrive after the deadline. `closeChallengeCollection` is also idempotent in practice, so the retry does not compound it.

The problem is that it is **undocumented and unbounded**. Nobody reading "completion failed" would infer "and submissions are now closed", and the set of steps that have already run at the point of any given failure is not written down anywhere. That matters most for the steps that are *not* harmlessly idempotent — payouts are already guarded, but each new step added to this path inherits the same ambiguity by default rather than by decision.

## Worth considering

- Write down the step order and, for each step, whether it is safe to have run before a later failure. That is cheap and it is most of the value here.
- Where a step is genuinely unsafe to leave half-done, either move it after the fallible work or give it an explicit compensation.
- A completion that fails after closing the collection could record how far it got in `metadata`, so a retry (or a human) can see the state rather than infer it.

Related: the companion issue on a challenge left in `Completing` holding a fresh claim, which is what turns this from a one-off into something a retry loop walks into repeatedly.

Issue #3822, opened 2026-08-11
Move the pairwise second run onto the review tick, so close-time only runs the podium


JustMaier:
The pairwise judging engine ([#3814](https://github.com/civitai/civitai/pull/3814)) does two things when a challenge closes: re-insert the top of the arrival ladder ("the second run"), then round-robin the shortlist to decide places. Both are LLM round-trips, and the whole thing has to finish inside the completion claim — `resetStuckCompletingChallenges` revokes at **10 minutes**, after which a second run claims the challenge and starts a second podium and a second path to `createChallengeWinner`.

## Where it stands

The second run is bounded to the top 40 of the arrival ladder, which fits:

```
284 entries, 9s measured bout latency, concurrency 16
  rerun  ceil(40/16) = 3 waves x ceil(log2 284) = 9 bouts x 9s ~ 4.1 min
  podium ceil(210/16) = 14 waves x 9s            ~ 2.1 min
  total                                          ~ 6.2 min   (claim: 10)
```

That works, but the margin is a function of things we do not control: provider latency (12.4s median on the self-hosted route, 56s worst case observed), field size, and the shortlist size. `LADDER_CONCURRENCY` is already at the top of the range the harness sustained without rate-limiting, so there is no headroom left in that dial.

## Proposal — move the second run onto the review tick

Arrival placement already runs incrementally, spread over the days a challenge is open. The second run does not have to be different: re-insert a slice of the ladder on each review tick, so that by the time the challenge closes the ordering work is already done and **close does only the podium** (~2 minutes, comfortably inside the window regardless of field size).

This removes the deadline from the expensive stage rather than racing it, and it decouples the close-time budget from field size entirely.

Sketch:
- Track a per-challenge cursor of when each standing was last re-inserted.
- On each review tick, re-insert the N stalest entries in the top K against the current standings.
- At close, run only the podium — plus a bounded catch-up for anything that arrived too late to have been re-inserted at all.

## Alternative, if field sizes ever outgrow K

Heartbeat the completion claim while the stage runs: re-stamp `completingClaimedAt` every few completions so a long stage keeps its claim instead of losing it. This was considered and **deliberately declined** for #3814 — it touches `claimStillHeld` and the in-run `claimedAt` that the completion telemetry compares against, and getting that interaction wrong reintroduces a double-emit on a path that pays people. It is the right lever if the bounded rerun ever stops fitting, but it wants its own change with its own review rather than being folded into a feature.

## Until then

[...]

PR #3876, opened 2026-08-13
fix(challenges): gate challenge completion writes on the completion claim

JustMaier:
Closes part of #3822 (option A — the guard, not the rolling-Swiss rework; independent of the pairwise branches).

## The bug

`pickWinnersForChallenge` checked its completion claim in exactly one place, and only to decide whether to emit two Prometheus counters (`claimStillHeld`'s own docblock said "TELEMETRY GATE ONLY"). `resetStuckCompletingChallenges` revokes a `Completing` claim on elapsed time alone, so a close-time judging stage longer than the 10-minute timeout is handed to a second run while the first is still executing — and the first kept judging, kept spending, and finished with an unconditional `UPDATE "Challenge" SET status = 'Completed' WHERE id = $1`.

Reproduced at production field size on a 45-entry challenge: two runs, 727 comparison rows instead of ~460, 403 Buzz instead of ~200, and first/second place decided by whichever run wrote last. Money is safe (payouts are keyed by deterministic `externalTransactionId`); the outcome is not.

## The fix

The claim now gates the writes, not just the telemetry:

- `completeChallengeIfClaimHeld` (new, in `challenge-helpers.ts`) writes `status = 'Completed'` (and, on the winners path, the completion metadata) with the claim stamp as an `UPDATE` predicate. It replaces both completion writes — the zero-entries path and the winners path — and returns whether the write landed.
- `challengeClaimStillHeld` (new) re-reads the stamp on the primary right before the judging call, so a run that has already been replaced stops instead of buying a second podium and creating a second set of winner rows.
- The two completion counters now follow the write's own result instead of a separately-read stamp, which also closes the read-then-write window the old comment called out as residual.
- An abort logs `challenge-completion-claim-lost` to Axiom with the stage (`pre-judging` / `zero-entries-completion` / `winners-completion`), so the condition is visible rather than inferred from a double payout.

**Missing-stamp semantics are unchanged.** Both new checks use `COALESCE(metadata->>'completingClaimedAt', '') IN ('', $claimedAt)` — the SQL equivalent of the old `claimStillHeld` fail-open. Only a stamp that is present *and* different stops a run, so this can never abort a run that still owns its challenge.

**Out of scope:** partial state left by a run that aborts after creating winners and paying but before completing is #3842, pre-existing and untouched. The winners-path abort deliberately leaves those rows alone and only stops the status write, the counters and the winner notifications, which belong to the run that owns the challenge.

## Tests

`challenge-completion-metrics.test.ts` grows a takeover model: the two claim helpers are faked against a simulated `metadata.completingClaimedAt`, so the tests exercise the same predicate the SQL uses. Three cases carry the fix — an evicted run that stops before judging, a takeover during judging that gets its final write rejected, and the zero-entries equivalent — plus the existing fail-open case, now asserted on both guards. The two older cases that pinned "the write still runs, only the emit is suppressed" were rewritten: that was the bug.

Each was checked by reverting the source and confirming the failure is readable:

| Test | Message with the fix reverted |
| --- | --- |
| an evicted run stops before spending on judging | `expected "generateWinners" to not be called at all, but actually been called 1 times` |
| winners path: does NOT write Completed and does NOT emit | `expected "recordChallengeCompleted" to not be called at all, but actually been called 1 times` |
| zero-entries path: does NOT mark Completed and does NOT emit | `expected "getChallengeById" to not be called at all, but actually been called 1 times` |

No new loops or loop-driving fakes.

## Verification

- `pnpm run typecheck` — no errors in any changed file (3 pre-existing environment errors remain in this worktree: a stale `@civitai/client` for LTX 2.5 and a missing `@opentelemetry/context-async-hooks`, both present at the base commit too).
- All 268 challenge-related suites green (3,622 tests).
- Sibling suites that mock `challenge-helpers` wholesale needed the two new exports added to their factories; `challenge-degenerate-winners` and `challenge-completion-emit-boundary` also had their Completed-write assertions retargeted at the new helper.
- Prettier run per-file (no repo-wide pass).

### DIFF
--- a/src/server/games/daily-challenge/__tests__/challenge-degenerate-winners.test.ts
+++ b/src/server/games/daily-challenge/__tests__/challenge-degenerate-winners.test.ts
@@ -28,6 +28,7 @@ const {
   mockGetExistingWinnersForRetry,
   mockResolveEventContext,
   mockUpdateChallengeStatus,
+  mockCompleteChallengeIfClaimHeld,
   mockRefundUserChallengeFunds,
   mockCreateNotification,
   mockCreateChallengeWinner,
@@ -51,6 +52,7 @@ const {
   mockGetExistingWinnersForRetry: vi.fn().mockResolvedValue([]),
   mockResolveEventContext: vi.fn().mockResolvedValue(undefined),
   mockUpdateChallengeStatus: vi.fn().mockResolvedValue(undefined),
+  mockCompleteChallengeIfClaimHeld: vi.fn().mockResolvedValue(true),
   mockRefundUserChallengeFunds: vi.fn().mockResolvedValue({ refundedEntries: 0 }),
   mockCreateNotification: vi.fn().mockResolvedValue(undefined),
   mockCreateChallengeWinner: vi.fn(),
@@ -94,6 +96,8 @@ vi.mock('~/server/games/daily-challenge/daily-challenge.utils', async () => {
 });
 
 vi.mock('~/server/games/daily-challenge/challenge-helpers', () => ({
+  challengeClaimStillHeld: vi.fn().mockResolvedValue(true),
+  completeChallengeIfClaimHeld: mockCompleteChallengeIfClaimHeld,
   claimChallengeForCompletion: mockClaimChallengeForCompletion,
   computeDynamicPool: vi.fn(),
   distributePrizes: vi.fn(),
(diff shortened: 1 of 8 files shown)


JustMaier:
Superseded by #3814 — Justin asked for a single PR, so these commits are merged into that branch (conflict in `daily-challenge-processing.ts` resolved there, and two test suites updated to follow the completion write into `completeChallengeIfClaimHeld`).

Closing this one; the change itself ships in #3814.

PR #3879, opened 2026-08-13
fix(images): delete from B2 unconditionally instead of falling through to a dead bucket

ZacxDev:
## What

`deleteImageFromS3` deleted from B2 only when the storage-resolver returned `backend === 'backblaze'`. Every other outcome — a 404, a service error, a timeout, all of which `resolveMediaLocation` collapses to `null` — fell through to a DigitalOcean Spaces bucket that was deleted on 2026-05-18.

That path cannot succeed. The endpoint answers `404 page not found` as `text/plain`, which the AWS SDK's XML error deserializer throws on (the same failure already documented above `purgeResizeCache`). And `deleteImages` drops the DB row *before* calling this, so the object outlives its row and stays publicly fetchable — CDN urls are unsigned — with one Axiom line as the only record that it happened. **2,545 images accumulated this way, growing ~350/day**, ~96% of them from the account-deletion drain.

## The fix

B2 is the only place an image can be: `ImageUploadBackend` has exactly one member and the DO bucket no longer exists. So the registry lookup is now for **observability, not routing** — a miss means "unregistered", never "somewhere else", and the delete proceeds against B2 either way.

A miss emits a new `delete-image-from-s3-unresolved-location` warning, so a registry gap or a storage-resolver outage stays visible rather than being papered over by the more permissive delete.

Removing the fallback leaves `getImageS3Client` with no callers, so it goes too. Nothing constructs a client from `S3_IMAGE_UPLOAD_*` any more — those vars are inert and can be dropped from config separately.

## 🔴 The test fixture could not have caught this

Worth a reviewer's attention independent of the fix. The env mock in `delete-image-from-s3-logging.test.ts` only returned values for names ending `_URL`/`_ENDPOINT`, so `S3_IMAGE_B2_ACCESS_KEY` and `S3_IMAGE_UPLOAD_BUCKET` were both `undefined` — **both delete branches were unreachable in every test in the file**, and each "the delete throws" case actually threw at the db refcount query *above* the S3 call. The suite was green with zero coverage of the thing it is named after.

Fixed here (real B2 env values + a mocked client + a mocked resolver), then 5 tests added.

## Red/green matrix

Measured, not assumed. Base = `054d25bab3`.

| Test | base | HEAD | what it is |
|---|---|---|---|
| deletes from B2 when the resolver returns no location | ❌ | ✅ | **regression** |
| warns about the unresolved location without failing the delete | ❌ | ✅ | **regression** |
| deletes from B2 when the resolver reports backblaze | ✅ | ✅ | invariant guard |
| stays silent when the resolver does have a location | ✅ | ✅ | invariant guard (negative control for the warning) |
| logs the failure and still invalidates when the B2 delete itself throws | ✅ | ✅ | positive control — proves the failure path is now reached *through* the S3 call, not the refcount query |

3 of the 5 pass at base too. They are guards and controls, not regression coverage, and are labelled as such in the file rather than counted as proof.

## Verification

- `unit` sweep of `src/server/services/__tests__/` + `src/server/jobs/__tests__/` → **244 files / 3329 tests passed, 0 failed**
- `typecheck` → exit 0, 0 `error TS`. Instrument validated first: a planted `const x: number = 'nope'` in `src/utils/s3-client.ts` was caught (`error TS2322`), so the clean verdict is a real result and not an empty log.
- `eslint` on both source files → 28 problems / 3 errors, **byte-identical to base** (all pre-existing).

**Not deployed, not verified against production.** The claim here is about the code and the suite.

## Still open (not in this PR)

1. Partition the Axiom `error` field (`| summarize count() by error`) — the `try` also wraps the db refcount query, so some share of those 2,545 lines may be a *second* defect this does not fix.
2. Recover the accumulated orphans. A bucket-vs-DB reconciliation is the right instrument, not a replay of the log — the log only captures throws, is retention-bounded, and cannot see anything before the drain landed on 2026-08-03, while the leak dates from 2026-05-18. Any replay needs the same refcount guard the live path has; a blind key-list delete can destroy media still referenced by a live row.
3. A counter + alert on delete failures, and a tombstone written with the row delete rather than after it — the ordering is what makes every failure unrecoverable by design.

Rationale and the full diagnosis live in the private infra repo, not here. See the pinned comment below for two corrections to this description.

ClickUp: https://app.clickup.com/t/868kpaenx

### DIFF
--- a/src/server/services/__tests__/delete-image-from-s3-logging.test.ts
+++ b/src/server/services/__tests__/delete-image-from-s3-logging.test.ts
@@ -233,10 +233,9 @@ describe('deleteImageFromS3', () => {
   });
 
   // THE REGRESSION. `resolveMediaLocation` returns null for a 404, a service error AND a timeout,
-  // and the code used to route that to a DigitalOcean bucket deleted 2026-05-18 — the delete threw
-  // on the proxy's text/plain "404 page not found", the row was already gone, and the object stayed
-  // publicly fetchable on an unsigned CDN url. 2,545 objects accumulated that way. An unregistered
-  // image is still on B2; there is nowhere else for it to be.
+  // and the code used to route that to a decommissioned second backend — the delete threw
+  // on a non-XML error body, the row was already gone, and the object stayed publicly fetchable on
+  // an unsigned CDN url. An unregistered image is still on B2; there is nowhere else for it to be.
   it('deletes the object from B2 when the resolver returns no location', async () => {
     mockFindFirst.mockResolvedValue(null);
     mockResolveMediaLocation.mockResolvedValue(null);
@@ -349,7 +348,38 @@ describe('deleteImageFromS3', () => {
     expect(mockLogToAxiom).toHaveBeenCalledWith(
       expect.objectContaining({
         name: 'delete-image-from-s3-unresolved-location',
-        error: expect.anything(),
+        // Content, not just presence. `expect.anything()` alone passed with the sanitiser dropped
+        // AND with the wrong value bound in — it pins that the key exists, not that it carries the
+        // reason this event is logged for.
+        error: expect.objectContaining({
+          message: expect.stringContaining('unexpected end of JSON input'),
+        }),
+      })
+    );
+  });
+
+  // `safeError` ends in `String(e)`, which THROWS on a value with no primitive conversion. Doing
+  // that work in the log payload put it inside the outer try, so a rejection carrying such a value
+  // skipped the B2 delete — the precise failure the `.catch` above exists to prevent, reintroduced
+  // by the code that reports it. Measured: this case passed at 85a31d11c7 and regressed at
+  // 156e050d69, so it is a real round-trip and not a hypothetical.
+  it('still deletes from B2 when the resolver rejects with an unserialisable value', async () => {
+    mockFindFirst.mockResolvedValue(null);
+    mockResolveMediaLocation.mockRejectedValue(Object.create(null));
+
+    await deleteImageFromS3({ id: 4242, url: 'abc-def/original.jpeg' });
+
+    expect(mockB2Send).toHaveBeenCalledTimes(1);
+    expect(mockLogToAxiom).not.toHaveBeenCalledWith(
+      expect.objectContaining({ name: 'delete-image-from-s3-failed' })
+    );
+    // The reason is unknowable here, but the event must still say so rather than vanish.
+    expect(mockLogToAxiom).toHaveBeenCalledWith(
+      expect.objectContaining({
+        name: 'delete-image-from-s3-unresolved-location',
+        error: expect.objectContaining({
+          message: expect.stringContaining('could not be serialised'),
+        }),
       })
     );
   });
--- a/src/server/services/__tests__/delete-image-from-s3-logging.test.ts
+++ b/src/server/services/__tests__/delete-image-from-s3-logging.test.ts
@@ -341,6 +341,17 @@ describe('deleteImageFromS3', () => {
     expect(mockLogToAxiom).not.toHaveBeenCalledWith(
       expect.objectContaining({ name: 'delete-image-from-s3-failed' })
     );
+    // Asserting the WARNING too, not just the delete. Without this the `.catch`'s return VALUE is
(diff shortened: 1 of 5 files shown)


ZacxDev:
Correcting two things in the description above rather than editing them away, since they were readable for a while.

**1. Two `claudedocs/` files have been dropped from this PR.** The original push included internal operational notes that do not belong in this repo. They have been removed and the branch force-pushed; the change is now three source files plus a new test. Their content is moving to the private infra repo. The PR body's closing line still referred to one of them as "included" — it isn't.

**2. "`resolveMediaLocation` cannot throw" was wrong**, and the first commit reasoned from it. `return res.json()` sat inside its `try` without `await`; in an async function the returned promise is adopted *after* control leaves the try, so a rejection escaped that catch entirely despite the `| null` signature. A 200 carrying a non-JSON body — an ingress or proxy error page, i.e. the exact storage-resolver-outage case this PR says it now tolerates — reached `deleteImageFromS3` as a throw and **skipped the B2 delete**, producing precisely the orphan this change exists to prevent.

Fixed in 85a31d11c7 at both the source and the call site, with a new `storage-resolver-null-contract.test.ts` (red at `ab4d922260` for the non-JSON case).

Same commit also closes two test gaps: the fixture set `S3_IMAGE_B2_BUCKET` to the same string as the code's own fallback, so "reads the env var" and "hardcodes the literal" were indistinguishable — replacing the whole expression with the bare literal passed all 11 tests. And nothing covered absent B2 credentials, which is the guard this PR removed.

Re-verified after the fixes: 245 files / 3335 tests, typecheck clean, prettier clean, no new lint errors.

**Still open for a reviewer's judgement, not addressed here:** `DATABASE_IS_PROD` is not set explicitly in any deployment — it falls back to `NODE_ENV === 'production'`, which is set in every deployed environment. So the early return at the top of `deleteImageFromS3` guards local development only, and any deployment holding real media credentials reaches the delete. That is pre-existing rather than introduced here, but this PR removes the credential check that was incidentally braking part of it, so it is worth an explicit decision rather than an assumption.

ZacxDev:
Second audit round on `85a31d11c7`. It found no behavioural defect in the round-1 fixes, and re-verified them under mutation rather than by reading. Three things it did turn up, all now fixed in `156e050d69`:

**The `.catch(() => null)` I added had its return *value* unpinned.** Replacing `() => null` with one that returns a location passed every test in the file — and that mutant silently suppresses `delete-image-from-s3-unresolved-location` on every resolver rejection, re-burying the exact signal the line above it calls the health signal. The rejection test now asserts the warning, not just the delete.

**Round 1 also lost error detail.** A rejecting resolver previously reached the outer catch as `delete-image-from-s3-failed` carrying `safeError(error)`. Routing it to the warning was right for the delete but discarded the only record of *why* the resolver failed. The warning now carries it.

**`STORAGE_RESOLVER_URL`'s env-vs-fallback was unasserted**, so dropping the env read left every test green. Now killed by asserting the fetch target. Stated rather than hidden: dropping the *fallback* half still survives, because every test runs with the var set — detecting that needs a second module instance with it unset, and the url is computed at module scope.

---

**🔴 Correcting my own verification claim, above and in two commit messages.** "Typecheck clean" covers **3 of the 5 changed files, not all of them.** `tsconfig.json:71` excludes `src/**/__tests__/**`, and vitest strips types through esbuild rather than checking them — so neither gate type-checks a test file.

Measured, not inferred: a planted `const __probe: number = 'not a number'` **in the new test file** gives `rc=0`, 0 errors; the identical error in `storage-resolver.ts` gives `rc=2`, `TS2322`. This is a pre-existing repo-wide gap rather than anything this PR introduced, but my claim overstated its reach, and the `as` casts in both test files are ungated by anything.

The `245 files / 3336 tests` figure does reproduce — it is `src/server/services/__tests__/` (210 files / 2952) plus `src/server/jobs/__tests__/` (35 / 383), run as one invocation.

The three items awaiting a maintainer decision are unchanged and none are addressed here: `DATABASE_IS_PROD` is not an effective environment gate; `civitai-next`/`-stage` run against the production DB with no B2 credentials, so this fix is inert there and will show a *new* error string in any post-deploy check; and `DeleteObject` returns 204 for a missing key, so absence of error lines is not evidence that anything was deleted.

Issue #3845, opened 2026-08-12
GET /api/v1/apps/{slug}: 500 body leaks the Prisma query + schema to unauthenticated callers, and is shaped like a valid 200


ZacxDev:
`GET /api/v1/apps/{slug}` returns **500 for every existing app**. The response body names the cause: a column the code queries is not in the production database.

## Repro

```
$ curl -s -o /dev/null -w '%{http_code}' https://civitai.com/api/v1/apps/sensei
500
```

Same for every app I tried — `sensei`, `custom-generators`, `panorama-360`, `gen-matrix`. Persistent across 3 consecutive retries, unauthenticated.

Body:

```json
{
  "message": "An unexpected error occurred",
  "error": "\nInvalid `prisma.appCollaborator.findMany()` invocation:\n\n\nThe column `app_collaborators.app_listing_id` does not exist in the current database."
}
```

## Where it is not

The failure is **after** the app is resolved, and it is specific to this route:

| Request | Result |
|---|---|
| `/api/v1/apps/sensei` | **500** |
| `/api/v1/apps/definitely-not-a-real-app-xyz123` | 404 `{"error":"App not found"}` — correct |
| `/api/v1/apps?limit=3` | 200, and the items carry `iconUrl` / `coverUrl` intact |
| `/api/v1/models?limit=1` | 200 |

An unknown slug still 404s correctly, so routing and lookup are fine — the `appCollaborator.findMany()` call runs only once an app has been found. The list route serves the same apps with their media fields populated, so **the data is intact**; this is the detail route's collaborator query alone.

## Regression window

I called this endpoint repeatedly on **2026-08-11** and it returned 200 with well-formed bodies throughout (I was reading `coverUrl` off it for five apps). By **2026-08-12** it is 500 on every call. So it broke inside that window.

## Cause

The code queries `app_collaborators.app_listing_id`; production does not have that column. That is a migration that has not been applied to the environment the API is reading — the schema the client expects is ahead of the database. Worth confirming which migration adds that column and whether it has been applied per-environment, since it will not have been picked up automatically.

## Two things that make this worse than a 500

**1. The error body is shaped exactly like a valid response.** It is well-formed JSON, and a client that reads fields without checking the status code gets `null` for everything:

```
$ curl -s https://civitai.com/api/v1/apps/sensei | jq '{iconUrl, coverUrl, name}'
{ "iconUrl": null, "coverUrl": null, "name": null }
```

[...]

PR #3881, opened 2026-08-13
fix(api): close the two remaining #3845 disclosure populations — 11 hand-rolled envelopes + driver text at 4xx

ZacxDev:
Follow-up to #3850, which fixed the `handleEndpointError` disclosure (#3845) and **enumerated two remaining populations without fixing them**, on the explicit reasoning that a security fix should not carry an 11-file refactor. This is those two, as two separable commits.

Both are the same defect class — **text a database driver wrote, served to a caller it was never written for** — and both are closed at one chokepoint rather than at N call sites.

---

## Why this is severity, not tidiness

The whole-object envelopes have looked safe for years because a plain `Error` has **non-enumerable** props and serializes to `{}`. A real driver error does not. Red control on `origin/main`, verbatim from the test output — this is a live public product's response body:

```json
{"message":"An unexpected error occurred",
 "error":{"length":100,"name":"error","severity":"ERROR","code":"23505",
          "detail":"Key (email)=(victim@example.com) already exists.",
          "schema":"public","table":"User","constraint":"User_email_key"}}
```

That `detail` is an **actual user row value**. `pg`'s `DatabaseError` is reachable via Kysely (`~/server/db/kyselyDb` runs on the `pg` pool), so this is not Prisma-only. Measured with a real `DatabaseError`, not reasoned about.

---

## Commit 1 — driver-authored text at a **4xx** (#3845 investigation 3)

`handleEndpointError` genericizes every 5xx and passes 4xx through byte-identically, justified in `isRestServerFault` by *"4xx is client feedback the caller is meant to read"*.

🔴 **That comment was false**, and is corrected here. `throwDbError` copies the driver's `message` verbatim, and `prismaErrorToTrpcCode` sends a large slice of codes to 4xx:

| code | status | what shipped |
|---|---|---|
| `P2000` | 400 | the offending **column** (`app_listings.slug`) |
| `P2003` | 409 | the foreign-key **constraint** name |
| `P2025` | 404 | the **model** and **method** |
| `P2024` | 408 | the connection-pool shape |

all carrying the raw ``Invalid `prisma.<model>.<method>()` invocation`` preamble, on public unauthenticated routes.

**The fix keeps the STATUS and replaces only the MESSAGE** — a driver error mapped to 404 really is a 404 — and **logs the un-redacted text on the way out**. That last part is load-bearing, not incidental: it buys the 4xx arm exactly the invariant the 5xx arm already has —

> the un-redacted text leaves the wire **exactly when** it enters the log, so genericizing can never destroy the only copy of a message.

503 is still untouched, for the mirror-image reason: it is *excluded* from the fault log, so its response **is** the only copy, and no Prisma code maps to `SERVICE_UNAVAILABLE`.

### The discriminator is message IDENTITY, not "is a driver in the cause chain"

The chain test has a **real false positive** that would have destroyed good messages: `throwBadRequestError('That slug is already taken', dbError)` — and `throwConflictError` / `throwRateLimitError` / `throwDbCustomError(msg)` — all attach the driver as `cause` while writing a message *for* the caller. Genericizing those redacts nothing and costs the user real feedback.

`isDriverAuthoredMessage(message, e)` instead asks whether the text on the wire **is byte-identical to a driver error's own text**. That separates the two by construction, whoever forwarded it. A test pins the false-positive case.

---

## Commit 2 — the 11 hand-rolled envelopes (#3845 investigation 4)

Population B leaked through **four different shapes**, which is why a per-field assertion would have missed most of them:

| shape | what it serialized | sites |
|---|---|---|
| `error: err.message` | verbatim driver text | `v1/users` |
| `error` | the whole error **object** | `v1/creators`, `v1/tags`, `v1/permissions/check`, `v1/vault/{all,get,check-vault,toggle-version}`, `notification/getDetails` |
| `error: error.cause` | the **wrapped** driver error | `v1/content/[[...slug]]` |
| `cause: error` | the object under a second key, next to a *generic* `error` string that made it look sanitized | `run/[modelVersionId]` |

Each catch now delegates. The copy is **deleted**, not corrected — `v1/creators` and `v1/tags` were verbatim copies of `handleEndpointError` that had drifted, keeping the abort/parse logic but not the genericization.

### Route-specific behaviour preserved (and pinned)

- **`v1/users` keeps its transient-search 503**, and the client-abort check moved **ahead** of it. An aborted Meili call can look transient, and the old order could answer `503 + Retry-After` to a client that had already gone away.
- **`v1/vault/{all,get}` keep their `MEMBERSHIP_REQUIRED` → `200 {vault:null}` arm.**
- The five routes whose catch previously ended **without** a `return` still resolve to `undefined` — only the body changes.
- ⚠️ **One intentional status change:** `v1/content` no longer answers a hard 500 for every `TRPCError`. A `NOT_FOUND` slug now gets its real **404**. Flagging it as the one wire change beyond the body.

---

## Test evidence

Every regression test below was **observed failing** against `origin/main`, and the reverts were scoped so the red is an assertion failure, not a module error.

| suite | red at `origin/main` | green at HEAD | how the control was taken |
|---|---|---|---|
| `endpoint-helpers-driver-4xx` | **6 / 12** | 12 / 12 | reverted **only** `endpoint-helpers.ts` (the other two files are pure additions with no behaviour of their own) |
| `rest-envelope-consolidation` | **42 / 47** | 47 / 47 | reverted the 11 route files |
| `rest-error-envelope-ledger` | see mutations | 9 / 9 | mutation-tested, below |

The tests that stay green are labelled `INVARIANT:` in the source and are **not** counted as regression coverage — they pin what must not change (4xx byte-identity for our own messages, the 503 hint, the 499 abort, observability, the route-specific arms).

### The guards were validated, not assumed

- **Both ledgers mutation-tested.** Reverting one route makes ledger 1 name it (`+ "v1/creators.ts"`); adding `P9999: 'FORBIDDEN'` to `prismaErrorToTrpcCode` makes ledger 2 fail with `expected [400,404,408,409] to deeply equal [400,403,404,408,409]`.
- **Positive control on the file walk**, because a zero from a sweep wired to nothing is indistinguishable from a clean repo. **Negative control on each pattern**, fed its own exemplar, because a typo'd regex matches nothing and reports a reassuring zero.
- **Ledger 2 derives its expected set from `prismaErrorToTrpcCode` itself** rather than a hand-copied list, so a new mapping cannot outrun the genericization map — `handleEndpointError` leaves an unmapped status alone, so completeness is what makes the design safe.
- **A false green was caught and fixed mid-review.** The `run/[modelVersionId]` case mocked `Tracker` with an arrow function, so `new Tracker()` raised `TypeError: … is not a constructor` — a *different* error that also genericizes. Its three "no secret in body" assertions were passing without ever carrying the driver payload. The log-attribution assertion is what exposed it.

Full gate on the final tree: `pnpm typecheck` **0 errors**; `vitest --project unit` **15,350 passed / 0 failed / 17 skipped** across 980 files. ESLint on the changed files: **0 errors, 5 warnings** vs a **6-warning** baseline on the same files at `origin/main` (I removed a dead `appRouter` import). The gate was run on the merged tree, not per-commit.

---

## Two changes a reviewer should look at deliberately

**1. The global test setup.** `src/__tests__/setup.ts` replaced `~/server/logging/client` (7 exports) with a **one-key object**. Any path reaching a second export died with `[vitest] No "<name>" export is defined on the mock`, surfacing far from its cause — it broke nine previously-green route tests the moment they delegated to a helper needing `buildCentralErrorLog` and `wasServerFaultLogged`. It now spreads the original and stubs only `logToAxiom` (the one with an I/O side effect). This is a repo-wide test-infra change in a fix PR; it is here because the fix cannot land without it.

**2. Three existing suites had assertions that PINNED the leak** — `toEqual({ message: 'Database connection lost' })` at a 500, and the `{ error, code }` shape carrying driver text through the field the Go CLI renders *in preference to* `message`. They now assert the genericized body **by value**, each with a comment saying what changed and why. Changing a test to match new behaviour deserves scrutiny; the new behaviour is independently covered by the red→green suites above, so these are not the evidence.

---

## Known gap — bounded and pinned, deliberately not fixed here

**17 sites** re-wrap a caught error's `.message` into a 4xx `TRPCError` **without** setting `cause`, which defeats `isDriverAuthoredMessage`: the wire text is the driver's, but nothing in the chain proves it.

```
routers/app-listings.router.ts   2
routers/apps-shared.router.ts    1
routers/blocks.router.ts        13
routers/referral.router.ts       1
```

The fix is one word per site (`cause: err`). It is **not** in this PR: it touches four router files on the App Blocks surface, with their own review and test surface, and a mechanical sweep over them should not ride along inside a security fix. *(I attempted it, got the caught-variable name wrong at nearly every site via a bad regex, and reverted — which is itself the argument for doing it deliberately rather than as a ride-along.)*

`rest-error-envelope-ledger` **pins the current set**, failing if it grows *or* shrinks — mutation-verified by adding an 18th site. The gap stays visible instead of decaying into folklore.

## Consumer compatibility

No breaking change. `message` stays a **string** at every status (the Go CLI's `readError` decodes it into `Message string`; a non-string fails the whole envelope unmarshal). `error` is retained and non-empty at every status (the CLI renders it *in preference to* `message`, so blanking it would empty every rendered CLI error). `code` is added where it was absent and is unchanged where it existed. The zod-flatten object on a 400 is untouched.

### DIFF
--- a/src/__tests__/setup.ts
+++ b/src/__tests__/setup.ts
@@ -233,8 +233,19 @@ vi.mock('~/server/prom/client', () => ({
   imageScanWebhookCounter: promMetricStub(),
 }));
 
-// Mock logging
-vi.mock('~/server/logging/client', () => ({
+// Mock logging.
+//
+// 🔴 Spreads the ORIGINAL rather than replacing the module with a one-key object.
+// `~/server/logging/client` has 7 exports; the previous wholesale mock provided
+// exactly one, so ANY code path reaching a second one died with
+// `[vitest] No "<name>" export is defined on the mock` — an error that surfaces
+// far from its cause. It bit when the REST routes were consolidated onto
+// `handleEndpointError` (civitai#3845/4): the helper needs `buildCentralErrorLog`
+// and `wasServerFaultLogged`, so nine previously-green route tests broke for a
+// reason that had nothing to do with what they assert. Only `logToAxiom` is
+// stubbed, because that is the one with an I/O side effect a test must not do.
+vi.mock('~/server/logging/client', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
   logToAxiom: vi.fn().mockResolvedValue(undefined),
 }));
 
--- a/src/pages/api/notification/getDetails.ts
+++ b/src/pages/api/notification/getDetails.ts
@@ -1,7 +1,7 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { bareNotification } from '~/server/notifications/base.notifications';
 import { populateNotificationDetails } from '~/server/notifications/detail-fetchers';
-import { AuthedEndpoint } from '~/server/utils/endpoint-helpers';
+import { AuthedEndpoint, handleEndpointError } from '~/server/utils/endpoint-helpers';
 
 const schema = bareNotification;
 
@@ -16,7 +16,10 @@ export default AuthedEndpoint(
       await populateNotificationDetails([results.data]);
       return res.json(results.data);
     } catch (error) {
-      return res.status(500).json({ message: 'An unexpected error occurred', error });
+      // civitai#3845 (population B): `populateNotificationDetails` fans out over
+      // per-type detail fetchers that query the DB directly, so the whole error
+      // OBJECT serialized here was driver-derived.
+      return handleEndpointError(res, error);
     }
   },
   ['POST']
--- a/src/pages/api/run/[modelVersionId].ts
+++ b/src/pages/api/run/[modelVersionId].ts
@@ -3,6 +3,7 @@ import { getServerAuthSession } from '~/server/auth/get-server-auth-session';
 import { dbRead, dbWrite } from '~/server/db/client';
 import * as z from 'zod';
 import { Tracker } from '~/server/clickhouse/client';
+import { handleEndpointError } from '~/server/utils/endpoint-helpers';
 
 const schema = z.object({
   modelVersionId: z.preprocess((val) => Number(val), z.number()),
@@ -61,7 +62,10 @@ export default async function runModel(req: NextApiRequest, res: NextApiResponse
       nsfw: modelVersion.model.nsfw,
     });
   } catch (error) {
-    return res.status(500).json({ error: 'Invalid database operation', cause: error });
+    // civitai#3845 (population B): `cause: error` serialized the whole error
+    // object under a second key. The generic `error` string next to it made this
+    // look sanitized — it was not; `cause` carried the driver payload.
+    return handleEndpointError(res, error);
   }
 
   // Append our QS
--- a/src/pages/api/v1/content/[[...slug]].ts
+++ b/src/pages/api/v1/content/[[...slug]].ts
@@ -1,9 +1,6 @@
-import { TRPCError } from '@trpc/server';
 import type { NextApiRequest, NextApiResponse } from 'next';
-import { publicApiContext, publicApiContext2 } from '~/server/createContext';
-import { appRouter } from '~/server/routers';
-import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
-import { isClientAbortError } from '~/server/utils/errorHandling';
+import { publicApiContext2 } from '~/server/createContext';
+import { handleEndpointError, PublicEndpoint } from '~/server/utils/endpoint-helpers';
 
 export default PublicEndpoint(async function handler(req: NextApiRequest, res: NextApiResponse) {
   // const apiCaller = appRouter.createCaller(publicApiContext(req, res));
@@ -12,12 +9,12 @@ export default PublicEndpoint(async function handler(req: NextApiRequest, res: N
   try {
     const result = await apiCaller.content.get({ slug: req.query.slug });
     return res.status(200).json(result);
-  } catch (error: any) {
-    if (isClientAbortError(error)) {
-      if (!res.headersSent) res.status(499).end();
-      return;
-    }
-    if (error instanceof TRPCError) return res.status(500).json({ error: error.cause });
-    else return res.status(500).json({ error: 'Internal server error' });
+  } catch (error) {
+    // civitai#3845 (population B): `{ error: error.cause }` served the WRAPPED
+    // error, which for a `throwDbError`-produced TRPCError is the driver error
+    // itself — the most direct leak of the whole set. It also answered a hard 500
+    // for every TRPCError, so a NOT_FOUND slug was reported as a server fault;
+    // delegating gives it its real status.
+    return handleEndpointError(res, error);
   }
 });
--- a/src/pages/api/v1/creators.ts
+++ b/src/pages/api/v1/creators.ts
@@ -1,10 +1,7 @@
-import { TRPCError } from '@trpc/server';
-import { getHTTPStatusCodeFromError } from '@trpc/server/http';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { getEdgeUrl } from '~/client-utils/cf-images-utils';
 import { publicApiContext2 } from '~/server/createContext';
-import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
-import { isClientAbortError } from '~/server/utils/errorHandling';
+import { handleEndpointError, PublicEndpoint } from '~/server/utils/endpoint-helpers';
 import { getPaginationLinks } from '~/server/utils/pagination-helpers';
 
 type CreatorItem = {
@@ -45,28 +42,14 @@ export default PublicEndpoint(async function handler(req: NextApiRequest, res: N
       },
     });
   } catch (error) {
-    if (isClientAbortError(error)) {
-      // Client disconnected mid-request — not a server fault. 499, not 500.
-      if (!res.headersSent) res.status(499).end();
-      return;
-    }
-    if (error instanceof TRPCError) {
-      const status = getHTTPStatusCodeFromError(error);
-      // Some TRPCErrors carry a JSON-stringified body (zod/validation); others
-      // (throwDbError-wrapped INTERNAL_SERVER_ERRORs) carry a plain string. A
-      // blind JSON.parse on the plain-string case throws, escapes this catch,
-      // and surfaces a raw unhandled 500 — so guard it with a fallback.
-      const parsedError = (() => {
-        try {
-          return JSON.parse(error.message);
-        } catch {
-          return { message: error.message };
-        }
-      })();
-
-      res.status(status).json(parsedError);
-    } else {
-      res.status(500).json({ message: 'An unexpected error occurred', error });
-    }
+    // civitai#3845 (population B): this catch used to be a hand-rolled copy of
+    // `handleEndpointError` that had drifted — it kept the abort/TRPCError/parse
+    // logic but NOT the genericization, so a `throwDbError`-wrapped driver error
+    // still put raw invocation text on the wire at a 500, and the else-branch
+    // serialized the whole error OBJECT (a Prisma error's enumerable own props
+    // carry the table + column). Delegating removes the copy rather than patching
+    // it; see `endpoint-helpers.ts` and `rest-envelope-ledger.test.ts`.
+    handleEndpointError(res, error);
+    return;
   }
 });
--- a/src/pages/api/v1/permissions/check.ts
+++ b/src/pages/api/v1/permissions/check.ts
@@ -3,8 +3,7 @@ import * as z from 'zod';
 import { EntityAccessPermission } from '~/server/common/enums';
 import { hasEntityAccess } from '~/server/services/common.service';
 import { sessionClient } from '~/server/auth/session-client';
-import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
-import { isClientAbortError } from '~/server/utils/errorHandling';
+import { handleEndpointError, PublicEndpoint } from '~/server/utils/endpoint-helpers';
 import { commaDelimitedNumberArray, numericString } from '~/utils/zod-helpers';
 
 const schema = z.object({
@@ -52,11 +51,10 @@ export default PublicEndpoint(
 
       return res.json(data);
     } catch (error) {
-      if (isClientAbortError(error)) {
-        if (!res.headersSent) res.status(499).end();
-        return;
-      }
-      return res.status(500).json({ message: 'An unexpected error occurred', error });
+      // civitai#3845 (population B): this used to serialize the whole error
+      // OBJECT. `hasEntityAccess` runs `$queryRaw`, so a failure arrives as a
+      // Prisma P2010 whose enumerable own props carry the raw database message.
+      return handleEndpointError(res, error);
     }
   },
   ['GET']
--- a/src/pages/api/v1/tags.ts
+++ b/src/pages/api/v1/tags.ts
@@ -1,12 +1,8 @@
 import { TagTarget } from '~/shared/utils/prisma/enums';
-import { TRPCError } from '@trpc/server';
-import { getHTTPStatusCodeFromError } from '@trpc/server/http';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { publicApiContext2 } from '~/server/createContext';
 
-import { appRouter } from '~/server/routers';
-import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
-import { isClientAbortError } from '~/server/utils/errorHandling';
+import { handleEndpointError, PublicEndpoint } from '~/server/utils/endpoint-helpers';
 import { getPaginationLinks } from '~/server/utils/pagination-helpers';
 
 export default PublicEndpoint(async function handler(req: NextApiRequest, res: NextApiResponse) {
@@ -35,28 +31,9 @@ export default PublicEndpoint(async function handler(req: NextApiRequest, res: N
       },
     });
   } catch (error) {
-    if (isClientAbortError(error)) {
-      // Client disconnected mid-request — not a server fault. 499, not 500.
-      if (!res.headersSent) res.status(499).end();
-      return;
-    }
-    if (error instanceof TRPCError) {
-      const status = getHTTPStatusCodeFromError(error);
-      // Some TRPCErrors carry a JSON-stringified body (zod/validation); others
-      // (throwDbError-wrapped INTERNAL_SERVER_ERRORs) carry a plain string. A
-      // blind JSON.parse on the plain-string case throws, escapes this catch,
-      // and surfaces a raw unhandled 500 — so guard it with a fallback.
-      const parsedError = (() => {
-        try {
-          return JSON.parse(error.message);
-        } catch {
-          return { message: error.message };
-        }
-      })();
-
-      res.status(status).json(parsedError);
-    } else {
-      res.status(500).json({ message: 'An unexpected error occurred', error });
-    }
+    // civitai#3845 (population B) — see the note on `v1/creators.ts`. Same
+    // hand-rolled copy, same drift, same delegation.
+    handleEndpointError(res, error);
+    return;
   }
 });
--- a/src/pages/api/v1/users/index.ts
+++ b/src/pages/api/v1/users/index.ts
@@ -5,7 +5,7 @@ import * as z from 'zod';
 import { env } from '~/env/server';
 import { publicApiContext2 } from '~/server/createContext';
 import { getAllUsersInput } from '~/server/schema/user.schema';
-import { PublicEndpoint } from '~/server/utils/endpoint-helpers';
+import { handleEndpointError, PublicEndpoint } from '~/server/utils/endpoint-helpers';
 import { isClientAbortError } from '~/server/utils/errorHandling';
 import { isTransientMeiliError } from '~/server/meilisearch/client';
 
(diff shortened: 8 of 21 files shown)


ZacxDev:
## Audit round 1 — applied (`7e396ba788`)

An adversarial audit returned **no deploy-blockers**, six real findings and four surviving mutants. All are fixed. Two of them **correct claims in the PR description above** — flagging publicly rather than silently editing the body, since a reviewer may already have read the wrong version.

### 🔴 Corrections to the PR body

**1. "One intentional status change" was wrong — there are five.** The body flagged only `v1/content`. The others were undocumented:

| route | change | why |
|---|---|---|
| `v1/vault/{all,get}` | **500 → 400** | `vault.service.ts:514` throws `throwBadRequestError('User does not have an active subscription.')` with **no cause**, so the `MEMBERSHIP_REQUIRED` arm never catches it. Body goes from `{"message":"An unexpected error occurred","error":{}}` to the real message. Also **500 → 404** for `throwNotFoundError` at `vault.service.ts:130/134/138/149`. |
| `v1/permissions/check`, `notification/getDetails`, `run/[modelVersionId]` | **500 → real 4xx**, and **500 → 499** on a client abort | same class: they previously answered a hard 500 for every non-abort error |

Mostly desirable — a client with retry-on-5xx logic now correctly stops retrying — but it was undocumented, and that is the part that mattered.

**2. The class was never confined to the 11 routes.** Making the ledger structural (see below) surfaced **41** sites of the same shape across `src/pages/api`, three of them **unauthenticated**: `generation/data.ts`, `generation/resources.ts`, `v1/model-files/[id]/tensor-metadata.ts`. They are now enumerated in a tiered baseline that fails if it grows **or** shrinks. They are not fixed here because the fix is not a delegation — each answers 4xx for *every* failure including zod-parse rejections, so routing them through the helper turns a legitimate client 400 into a 500. That needs its own red/green.

### Production behaviour fixed

- **🔴 New 404s were edge-cacheable.** `PublicEndpoint` stamps `public, s-maxage=300, stale-while-revalidate=150` on **every** response *before the handler runs*, errors included. Survivable while every failure was a 5xx (shared caches don't cache those by default); not survivable once an error answers **404**, which this PR causes. A transient `NOT_FOUND` — replica lag, a `throwDbError`-mapped P2001/P2025 — would have been pinned at the edge for 300s + 150s SWR. Genericized responses are now `no-store`.
- **A genericized 408 logged at ERROR severity.** The only Prisma codes reaching 408 are `P1008`/`P2024` — connection-pool exhaustion — so every affected public request would have added an error-board line during exactly the incident you need signal in. Same wave-shaped property for which `isRestServerFault` excludes 503. Now logged at **info**, scoped locally rather than changing `classifyErrorFault` repo-wide (TIMEOUT genuinely *is* a server fault in general).
- **A dead `status !== 503` clause** removed, and the test comment claiming it enforced the exclusion corrected — the map's key set does, and ledger 2 fails if a 503 key is added.

### Two of my own tests were not pulling their weight

- **The abort-before-503 ordering guard was UNREACHABLE.** It threw a bare `new Error('The operation was aborted')` — satisfies `isClientAbortError` but **not** `isTransientMeiliError`, so the 503 arm could never fire and the test passed **with the ordering reversed** (`if (true)` left 91/91 green). It now throws the real prod shape (a `MeiliSearchCommunicationError`-named error with `code: 'ECONNRESET'` carrying the aborted-fetch message), which satisfies both. The mutant now dies on this test's own assertion: `expected 503 to be 499`.
- **Ledger 1 was a spelled guard.** Three regexes anchored on the literal `'An unexpected error occurred'` and a `.status(…)` chain — a file containing only `res.status(500).json({ error })` swept clean. Replaced with a brace-matching extractor over every `.json({…})` body plus four value-shape predicates. Two narrower holes died with it: `res.status(getHTTPStatusCodeFromError(e))` couldn't match a `[^)]*` group, and `res.json({ error })` with no `.status()` wasn't considered at all.
- **Generic strings were derived from the implementation** rather than pinned, so mutating 400's message to `''` was green on both sides. Now literals.

### Re-verified

Mutants that previously **survived** and are now **killed**: generic-message→`''`, drop-no-store, revert-info-override, neuter-the-`CAUGHT`-set, comment-out-the-extractor (that last one caught by the extractor's own positive control). New coverage for every new behaviour, including a negative control proving the info-severity override did **not** leak into the 5xx arm.

Gate: `typecheck` **0 errors**; unit **15,355 passed / 0 failed / 17 skipped** across 980 files; eslint on changed files **0 errors**.

### Still open, deliberately

- The **17 `no-cause` bypass sites** (unchanged from the original description).
- The **3 unauthenticated tier-1 sites** above.

Both are pinned by ledgers that fail if they grow, so neither can decay into folklore. They are the next PR, not this one.

ZacxDev:
## Audit round 2 — applied (`ad802e1954`)

The delta re-audit confirmed the **production** half of round 1 was sound (12/12 mutants killed; the un-masking invariant survives the severity override — verified against the *real* `buildCentralErrorLog`, not the test's mock; the 503 removal is provably dead code). Every new finding was in what round 1 itself added. Two were 🔴.

### 🔴 The round-1 cache fix missed the case that motivated it

`noStore` was scoped to the **genericized** arms. But the newly-cacheable 404s this PR creates come through the **pass-through** arm — `v1/content` answers 404 via `throwNotFoundError`, whose message is *ours*, so `isDriverAuthoredMessage` is false and it never reached the genericized branch. Measured per arm, against a `res` pre-stamped with `PublicEndpoint`'s header:

| arm | `Cache-Control` before this commit |
|---|---|
| **pass-through 404 (`throwNotFoundError`)** | **`public, s-maxage=300, stale-while-revalidate=150`** ← the hazard |
| non-TRPCError 500 (else-branch) | `public, s-maxage=300, stale-while-revalidate=150` |
| genericized 5xx | `no-store, max-age=0` ✅ |

Scoping by *"did we rewrite the body"* was the wrong axis; the right one is *"should a cache keep this"*, and for an error the answer is always no. `noStore` now applies to **every** response the helper writes.

Stated trade-off, now in the JSDoc: a 404 flood reaches the origin instead of being absorbed at the edge for 5 minutes. Uniformity also removes a **fault-classification oracle** — under the narrow version, `no-store` vs `s-maxage=300` on the same status told a caller whether the message had been rewritten.

### 🔴 The ledger was still spelled, and swept a live *unauthenticated* leak clean

Round 1 moved the spelling from the literal message text to the **variable name** (`CAUGHT = (?:e|err|error|ex)`). That misses `src/pages/api/v1/images/index.ts:236` — a `PublicEndpoint` serving `{ error: trpcError.message, code: trpcError.code }`. A `throwDbError`-wrapped `TRPCError` carries the driver's message verbatim, so this is exactly the #3845 class on exactly the kind of route this PR exists to protect — invisible only because the binding is named `trpcError`.

`IDENT` is now any identifier, `?.` is handled, and the shapes are predicates rather than bare regexes. One subtlety worth flagging: the `NOT_AN_ERROR` allowlist deliberately does **not** apply to `.message` *access* — applying it there exempted `(reason as Error).message`, unmistakably an error, purely because `reason` was on a list of string-ish names.

**Baseline re-derived: 3 regexes → 41 sites → 52.** That trajectory is itself the finding. TIER 1 (unauthenticated) is now four, adding `v1/images/index.ts`. It is **pre-existing** and unrelated to this PR's diff, so it doesn't block this change — but it should not have been invisible.

### 🔴 `orchestrator/refreshBlobs.ts` was misclassified as TIER 3

TIER 3's own prose is *"behind JOB_TOKEN / WEBHOOK_TOKEN / moderator auth … the driver detail is arguably the POINT … fixing them is optional."* But `OrchestratorEndpoint` is `AuthedEndpoint` + a per-user token with **no moderator check** — reachable by any logged-in account. A misclassification there dispositions a session-authed disclosure as a non-issue. Moved to TIER 2; the tier definition now names only `WebhookEndpoint`/`ModEndpoint`.

### 🟡 False positives, and a silent truncation in the matcher

The shapes anchored on `[{,]`, matching a key at **any** nesting depth — `json({ items, summary: { error, warnings } })` and `json({ incident: { cause: 'flood' } })` were both flagged. A guard that matches everything trains people to edit the baseline instead of the code. Nested literals are now stripped before matching.

Separately: a `}` inside a string literal closed the body early and silently **truncated** what got matched — a false negative in the mechanism itself. The extractor now skips string/template literals.

**Known blind spots are now stated rather than implied away:** a body built in a variable and passed as `res.json(body)`; an error nested below top level; `String(e)` / `.toString()` / spread; `res.send({…})`. A green run means *"no leak of a shape this guard knows"*, not *"no leaks"*.

### Re-verified

Round-2 mutants, all killed: drop `noStore` from the pass-through arm · drop it from the else-branch 500 · narrow `IDENT` back to the spelled set (`v1/images` reappears) · disable nested-literal stripping (false positives return).

Gate: typecheck **0 errors**; unit **15,357 passed / 0 failed / 17 skipped** across 980 files; eslint on changed files clean.

A round-3 delta audit is running. I'll report it either way — including if it's clean.

ZacxDev:
## Audit round 3 — applied (`2cb8805376`)

Round 3 confirmed the production behaviour and the tiering work, and found the next defect in the three **extractor/allowlist** changes round 2 added. Two rounds of "now it's structural" were each still spelled; that is the story of this thread.

### 🔴 The `no-store` guard never exercised an overwrite

`createRes()` started with an **empty** header map, so every assertion only proved *"the header ends up `no-store` when nothing set it first"* — never the case in production, because `PublicEndpoint` calls `addPublicCacheHeaders` **before** `await handler(req, res)`.

The audit demonstrated it by mutating `noStore` into a plausible refactor:

```ts
if (!res.headersSent && !res.getHeader('Cache-Control')) res.setHeader(...)
```

…a "don't clobber the route's own header" change any reviewer would wave through. **The entire suite stayed green**, while in production that silently restores `public, s-maxage=300` on every error the helper writes, on every public route — reinstating the whole hazard this work exists to close.

Fixture is now pre-stamped with the exact header `PublicEndpoint` sets, and `getHeader` exists so the mutant is *expressible* rather than throwing (a throw would look like a kill for the wrong reason). **That mutant now takes down 51 tests.** The 499 arm and `noStore`'s internal `headersSent` guard were also unpinned; both now kill their mutant.

### 🔴 The `NOT_AN_ERROR` allowlist laundered a live unauthenticated leak

Round 2's fix for name-anchoring re-added the same species on the **value** side. An allowlist of "string-ish" identifiers (`message`, `reason`, `detail`, …) exempted any body whose error value carried one of those names — hiding `src/pages/api/download/models/[modelVersionId].ts:257`:

```ts
return errorResponse(500, err.message);   // → res.json({ error: message })
```

`PublicEndpoint`. Unauthenticated. The catch-all around the dbRead/file lookups, i.e. exactly the #3845 Prisma-invocation-text shape — and the ledger reported the file **clean**.

The allowlist is gone. The cost is over-reporting; that is the right direction to be wrong in. The two genuinely-innocent cases are recorded in the baseline under an explicit **OVER-REPORTS** heading with the reason, rather than a name heuristic silently deciding which disclosures count.

### 🔴 The quote-awareness fix was inert

Round 2 made `jsonBodies` skip string literals so a `}` inside `'oops }'` couldn't truncate the body — then handed the result to a *separate*, quote-**blind** `topLevelKeys()` that truncated it again at exactly the same character. Measured end-to-end: **unchanged**. Both mutants deleting the quote handling left the suite green. A no-op whose JSDoc claimed the hazard was closed.

Merged into one pass (`scanBody`) that brace-matches, skips quoted regions and returns the depth-1 slice together. A string value survives as an empty literal (`'x'` → `''`) so it stays visibly a string. An unterminated quote (apostrophe in a trailing `//` comment, a regex literal) now falls back to a quote-blind scan instead of emitting **nothing** for that call — silence is the worse failure direction.

### Baseline: 52 → 57, and the trajectory is the finding

**3 regexes → 41 → 52 → 57.** Every widening was forced by an audit defeating the previous version with a *real* shape, twice on an **unauthenticated** route. The JSDoc now tells the reader to assume a fourth shape exists rather than implying completeness — a green run means *"no leak of a shape this guard knows"*.

Also: `blocks/submit-version.ts` and `v1/blocks/submit-version.ts` moved TIER 2 → TIER 3 (both `ModEndpoint`, which is what round 2 tightened TIER 3 to mean).

### 🟢 The origin-load note is now measured, not asserted

It previously read *"these routes are not the ones under that kind of load"* — the kind of unverifiable claim that gets quoted back as fact. Measured across the 25 callers: exactly **one** route's 404 changes cacheability (`v1/content`, low volume, small slug space). `v1/models/[id]` — the only high-traffic one — answers its common 404 **directly**, outside the helper, so it is unaffected.

### Re-verified

All four previously-**surviving** mutants now killed: don't-clobber `noStore` (51 tests) · drop `noStore` from the 499 arm · delete `noStore`'s `headersSent` guard · delete the quote-awareness.

Gate: typecheck **0 errors**; unit **15,360 passed / 0 failed / 17 skipped** across 980 files; eslint on changed files clean.

### What is still open, and why

Two ledger-pinned follow-ups, both **pre-existing** and neither introduced by this PR:
- the **17 `no-cause` bypass sites** (`cause: err` is the one-word fix);
- the **5 TIER-1 unauthenticated sites**, now including `v1/images/index.ts` and `download/models/[modelVersionId].ts`.

Both fail their ledger if they grow. Neither belongs in this PR: fixing them changes 4xx status semantics on routes this diff doesn't touch, and that needs its own red/green.

Three audit rounds, each of which found a real defect in the previous round's fix. I'd rather that be visible in the thread than compressed out of it.

ZacxDev:
## Audit round 4 — applied (`08b7832a4a`)

Round 4 confirmed round 3's substantive fixes, and confirmed the delta changes **no runtime behaviour** (the sole source edit is a comment). It also measured round 3's extractor fix as the first one in this thread that is provably *non-inert*: the old two-pass version missed `json({ label: 'a } b', error: err })`, the new single pass catches it.

Three findings, all comment/test-level. One is worth stating plainly.

### 🔴 The one claim I upgraded from "assertion" to "measured" is the one that measured false

Round 3 replaced *"these routes are not under that kind of load"* with a **measured** claim: exactly one route's 404 changes cacheability. That number came from surveying only the six `PublicEndpoint` callers.

`MixedAuthEndpoint` calls `addPublicCacheHeaders` on **every anonymous request** (`if (!session) addPublicCacheHeaders(...)`), so its by-id routes carry the same header. At least three routes are affected:

| route | wrapper |
|---|---|
| `v1/content/[[...slug]]` | `PublicEndpoint` |
| `v1/articles/[id]` | `MixedAuthEndpoint`, `throwNotFoundError` |
| `v1/collections/[id]` | `MixedAuthEndpoint`, `throwNotFoundError` |

And the file **already contradicted me 250 lines below**, where the 5xx arm names *"`GET /api/v1/articles/{id}` (a public MixedAuthEndpoint)"* as unauthenticated-reachable. A precise wrong number is more quotable than the vague one it replaced, which is why this got its own commit rather than a quiet edit.

Still correct: 25 callers; `v1/models/[id]` — the highest-traffic one — answers its common 404 directly without reaching the helper.

### 🟡 `v1/blocks/submit-version.ts` was re-tiered on a false premise

Round 3 moved it TIER 2 → TIER 3 as "ModEndpoint". It is not — it is `withAxiom` + `Authorization: Bearer <API key>`, gated by `isAppBlocksAuthorEnabled`, a Flipt flag whose own docs say it grants authoring to a curated cohort **independent of the mod-only flag**. TIER 3 means *"fixing is optional and may be undesirable"*, so this is the same misclassification I corrected for `orchestrator/refreshBlobs.ts` one round earlier — in the opposite direction. Moved back, with the real auth model recorded on the entry.

### 🟡 A vacuous assertion that also claimed a property the helper lacks

The "already SENT" test asserted `.not.toThrow()` with a comment saying *"the helper must be a no-op"*. Both halves wrong: the assertion cannot fail (making the fixture faithful turns it **red**), and the helper does not no-op — only the 499 arm and `noStore` itself are `headersSent`-guarded, while every 4xx/5xx arm ends in `noStore(res).status(…).json(…)` unguarded. Callers guard themselves; `download/models/[modelVersionId].ts` has exactly that `if (res.headersSent) return;`. Asserting the property would have told a maintainer that guard was redundant. The header assertion (which does kill its mutant) stays; the false half is gone.

### 🟢 Two extractor branches had no exemplar — and the fix took two attempts

The unterminated-quote fallback and the escape-skip are unreachable across all ~970 `.json({…})` sites, and deleting either left the ledger green. Both now have a unit exemplar.

The escape one is instructive: with a *plain* escaped quote, deleting the skip merely runs the scan to EOF, the fallback rescues it, the leak is still found — **mutant survives**. The exemplar needs a `}` inside the escaped string, so dropping the skip ends the body early at a real depth-0 brace, the scan "succeeds", the fallback never fires, and the leak lands outside the extracted body. Verified dead.

Also recorded in the commit: my first two attempts to *apply* that mutant matched zero lines and the runner reported a green 11/11. **A mutation that fails to apply is indistinguishable from one that is killed** unless you check the substitution landed. Applied by line number and confirmed before believing the result.

Gate: typecheck **0 errors**; unit **15,361 passed / 0 failed / 17 skipped** across 980 files; eslint clean on changed files.

---

**Four rounds, each finding a real defect in the previous round's fix.** Runtime behaviour has been stable and correct since round 1 — every finding since has been in guard integrity or in the accuracy of claims written into comments. Round 5 is running against a delta that is now comment-and-test-only.

ZacxDev:
## Audit round 5 — applied (`8eea407d9e`)

Round 5 verified every substantive round-4 claim (tier correction ✅, the `headersSent` statement ✅, both new extractor exemplars non-vacuous under mutation ✅, baseline still 57 ✅, delta comment-only ✅ — the comment-stripped file is byte-identical). It also independently reproduced the discriminating detail from round 4: remove the `}` from inside the escaped string and the escape-skip mutant **survives**, so that brace is load-bearing exactly as documented.

It found one thing — **the same defect for the third consecutive round.**

### 🔴 Third incomplete survey, restated as fact

The `noStore` "accepted trade" note has now been wrong three times, each time more precisely:

| | claim | why it was wrong |
|---|---|---|
| v1 | *"these routes are not under that kind of load"* | asserted, never checked |
| v2 | *"exactly ONE route affected"* — **measured** | surveyed only `PublicEndpoint` |
| v3 | three routes + *"the MixedAuthEndpoint ones do have a rate limiter"* | missed `v1/models/index.ts`; the limiter claim is false for exactly that route |

**What was missed:** `v1/models/index.ts` is a `MixedAuthEndpoint` calling this helper, and `GET /api/v1/models?username=<nonexistent>` (anonymous) reaches `throwNotFoundError('User not found')` at `model.service.ts:588` → the 404 pass-through arm → `noStore`.

And it is precisely the route where the reassurance fails. Measured: of the seven `MixedAuthEndpoint` callers, six use `checkPublicApiRateLimit` / `enforceAppsCatalogRateLimit`; `v1/models/index.ts` uses neither. Unlimited, anonymous, and each miss costs **two** user lookups — the second against the **primary**, since `model.service.ts:585-586` is `dbRead.user.findUnique(...) ?? await dbWrite.user.findUnique(...)` and a nonexistent username always misses both. I read that, rather than repeating the audit's summary of it.

### The fix is not a fourth list

Each version got more quotable and stayed wrong, so the comment now states the **rule** plus how to regenerate the set:

> A route loses 404 edge absorption iff (a) its wrapper stamps the public cache header — `PublicEndpoint` unconditionally, `MixedAuthEndpoint` **on every anonymous request** (the half that kept getting missed) — and (b) its 404 reaches this helper rather than being answered directly (which is why `v1/models/[id]`, the highest-traffic caller, is unaffected).

Examples are explicitly labelled **not** a complete set, with the grep to regenerate them, and the rate-limiter sentence names its exception instead of making a blanket claim.

### Also worth recording

Round 5's own instrument check caught two traps that would have produced false confidence: `vitest --reporter=basic` **does not exist in vitest 4** — it ran zero tests and still **exited 0**; and `tsconfig.json:71` excludes `src/**/__tests__/**`, so `pnpm typecheck` says nothing about either touched test file (verified by injecting a type error there and getting "OK"). Both pre-existing, neither a defect in this PR, but both worth knowing before quoting a green.

Gate: typecheck **0 errors**; unit **15,361 passed / 0 failed / 17 skipped** across 980 files.

---

**Five rounds. Runtime behaviour has been correct and unchanged since round 1** — every finding from round 2 onward has been in guard integrity or in the truthfulness of claims written into comments. That is not a trivial category: a false comment is what persuades the next maintainer to delete a guard. But it does mean the shipped change has been stable for four rounds, and the remaining risk is in documentation accuracy rather than behaviour.

A round-6 audit is running against a delta that is now a single comment block. If it comes back clean I'll say so and stop.

ZacxDev:
## Audit round 6 — **clean**. Stopping here.

Every factual assertion in the rewritten `noStore` block was verified against source. The counts are exact — **7** `MixedAuthEndpoint` callers (6 rate-limited, `v1/models/index.ts` not), **6** `PublicEndpoint` callers (0 rate-limited) — the `?username=` trace and the `dbRead ?? dbWrite` primary fallback are real, and running the re-derivation recipe reproduces a consistent set. The delta is comment-only: the comment-stripped file is byte-identical across the two revisions.

Two notes from the round, recorded so nobody re-derives them:

- **The example list is a subset, as labelled.** Following the recipe also surfaces `v1/articles/index` and `v1/collections/index` (same `?username=` path via `article.service.ts:247`). The comment declares the list non-exhaustive and gives the grep, so this is not a false claim — and I am deliberately **not** editing it to add them. This comment block has been wrong three times; each fix was another chance to introduce a fourth error, and the marginal value of two more examples in an explicitly-partial list does not justify another round.
- **The recipe's step 3 is a narrower proxy than the rule, in principle only.** "Can reach a `throwNotFoundError`" is narrower than "its 404 reaches this helper", because `throwDbError` maps P2001/P2015/P2018/P2025 → NOT_FOUND. No live counterexample exists today: all 13 candidate routes are read-only GETs, and the closest (`v1/creators` → `throwDbError`) issues `findMany`/`count`, which won't realistically raise P2025. Probed, not reportable — noted so it isn't probed again.

---

## Where the PR stands

**Six audit rounds. Runtime behaviour has been correct and unchanged since round 1** — every finding from round 2 onward was in guard integrity or in the accuracy of claims written into comments. Not a trivial category (a false comment is what persuades the next maintainer to delete a guard), but it does mean the shipped change has been stable for five rounds.

**What the PR does:** routes the 11 hand-rolled error envelopes through `handleEndpointError` (#3845 population B), and stops driver-authored text reaching the wire at a 4xx (investigation 3) — status kept, message replaced, un-redacted text logged so the "genericized exactly when logged" invariant extends to 4xx.

**Gate:** typecheck 0 errors · unit **15,361 passed / 0 failed / 17 skipped** across 980 files · eslint on changed files 0 errors (vs a 6-warning baseline on the same files at `origin/main`).

**Red→green, measured by reverting the relevant files to `origin/main`:** 42/47 and 6/12 red. Every guard added since has been mutation-tested, and every mutant an audit found surviving is now killed.

**Deliberately still open, both pre-existing and both ledger-pinned so they fail if they grow:**
- the **17 `no-cause` bypass sites** (`cause: err` is the one-word fix, on the App Blocks tRPC surface);
- the **5 TIER-1 unauthenticated sites**, including `v1/images/index.ts` and `download/models/[modelVersionId].ts` — both surfaced *by* this PR's ledger rather than by the original #3845 enumeration.

Neither belongs here: fixing them changes 4xx status semantics on routes this diff does not touch, and that needs its own red/green.

Ready for human review.

PR #3883, opened 2026-08-13
fix(images): scope the hideMeta cache purge, send the admin secret, stop swallowing failures

ZacxDev:
**Inert today.** The cache service's destructive mode is off, so `keep` is ignored and the header is optional. This is the client-side prerequisite for turning it on.

Tests: **19 pass** across the three affected suites (+9 new).

## 1. Scope — two callers, opposite blast radii

`purgeResizeCache` has two callers that need different things:

| caller | situation | needs |
|---|---|---|
| `deleteImageFromS3` | row already dropped | the image gone entirely |
| `updatePostImage`, on `hideMeta` false→true | **image stays live** | only the pre-flip variants gone |

The cache key includes the hideMeta flag, so a flip **re-keys the image**: every request after it derives a fresh, metadata-stripped variant under a new key, and the pre-flip variants become orphans still carrying the metadata the user just asked to hide. Those are what need clearing.

Clearing the whole set for that caller would also drop the variants the page is serving *right now*, blanking a live image until each cache tier caught up. So the caller now says which it means. `all` remains the default, making every existing call byte-identical.

## 2. Auth — we were sending no header at all

The cache service requires a shared-secret header once its destructive mode is enabled, and rejects the call without it. Enabling that mode would therefore have rejected **every invalidation we issue**. Sending the secret whenever it's configured makes enabling it a one-sided change rather than a synchronised deploy across two services.

New optional env var `IMAGE_CACHER_ADMIN_SECRET` — unset behaves exactly as today.

## 3. 🔴 The part that would have been invisible

`fetch` does not reject on a non-2xx. Every response landed in the success path, so a **401 from a missing or stale secret** — or a 409 refusal, or a 503 partial failure — produced **no log line anywhere**. Invalidation could have stopped working completely and the only symptom would have been stale images with no signal to trace it from.

That is not hypothetical: it is precisely what would have happened on the first deploy after the service's delete mode was switched on. The status is now checked and reported.

## Tests

9 new, covering scope (all three ways), the header, and each non-2xx status — plus a **positive control** that a 2xx logs nothing, so a logger firing unconditionally can't satisfy the failure cases.

Also pins the caller's scope in `update-post-image-hidemeta-bust.test.ts`. That test spied `purgeResizeCache` but never asserted its arguments, so nothing stopped the hideMeta path silently regressing to a full purge of a live image.

**Mutation-checked — 5 mutants, all killed by their own named test:** dropping the `keep` parameter · the caller falling back to the default · removing the status check · dropping the header · inverting the scope condition.

> Worth recording: the first run of that battery reported five clean passes that were an artifact of the runner being handed no test files at all. The results above come from a re-run whose harness was positive-controlled first.

### DIFF
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -736,6 +736,12 @@ export const serverSchema = z
     // image from B2. Optional — if unset, invalidation is skipped.
     IMAGE_CACHER_URL: z.url().optional(),
 
+    // Shared secret for image-cacher's /admin/* endpoints. The service only requires it once its
+    // destructive cache-object mode is enabled, and rejects the call outright without it — so this
+    // must be configured BEFORE that mode is turned on, or invalidation stops working entirely.
+    // Optional here so unset behaves exactly as today.
+    IMAGE_CACHER_ADMIN_SECRET: z.string().optional(),
+
     // BitDex
     BITDEX_URL: z.string().optional().default(''),
 
--- a/src/server/services/__tests__/image-cacher-invalidate-scope.test.ts
+++ b/src/server/services/__tests__/image-cacher-invalidate-scope.test.ts
@@ -108,8 +108,6 @@ vi.mock('~/server/search-index', async (importOriginal) => ({
 
 const { purgeResizeCache } = await import('../image.service');
 
-
-
 const invalidateCalls = () =>
   mockFetch.mock.calls.filter((call) => String(call[0]).includes('/admin/invalidate'));
 
--- a/src/server/services/__tests__/image-cacher-invalidate-scope.test.ts
+++ b/src/server/services/__tests__/image-cacher-invalidate-scope.test.ts
@@ -14,6 +14,11 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 // (delete-image-from-s3-logging.test.ts): stub env + infra clients + the event-engine-common
 // submodule so importing it boots no real infra.
 
+// Mutable so a test can remove the secret without rebuilding the whole env mock.
+const { envOverrides } = vi.hoisted(() => ({
+  envOverrides: { IMAGE_CACHER_ADMIN_SECRET: 'test-shared-secret' as string | undefined },
+}));
+
 const { mockLogToAxiom, mockFindFirst, mockFetch } = vi.hoisted(() => ({
   mockLogToAxiom: vi.fn(async () => undefined),
   mockFindFirst: vi.fn(),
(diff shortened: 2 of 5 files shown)


ZacxDev:
### Audit fixes (`566a88a6e3`)

Adversarial audit: **no deploy-blocker**, and 19 of 21 semantic mutants already died. Both survivors guarded the two claims this PR leans hardest on — neither was a defect in the shipped code, only in what protected it. Tests **19 → 21**.

**1. The backward-compatibility claim was the one thing untested.** The description says unset behaves exactly as today. Removing the `if (env.IMAGE_CACHER_ADMIN_SECRET)` guard **passed all 19 tests** while sending the literal header `X-Admin-Secret: undefined`. The service compares constant-time and rejects a mismatch, so the moment its delete flag flips that is a **permanent 401 with invalidation dead** — and before the status check added in this same PR, silent.

Now asserted, including that no header value is the stringified-`undefined` form. That detail matters: a `toBeUndefined()` assertion would have accepted exactly what the broken code produces.

**2. The direction invariant rested on nothing.** `keep=hm` retains the `_hm` variants, which on a **true→false** flip are the *orphans*, not the live set — the scope is inverted in that direction, and the service says so in its own rejection message. Widening the caller's guard to fire both ways passed every test in the suite. The true→false case now asserts no purge fires.

The guard itself is correct today — the audit confirmed all four directions by construction (`false→true` → one scoped call; `true→false`, omitted, and `true→true` → zero calls).

**3. The secret could ride a redirect.** `fetch` strips `Authorization` and `Cookie` on a cross-origin hop but forwards **custom** headers verbatim — verified against two local origins, `X-Admin-Secret` was forwarded in full. The endpoint only ever answers 202/400/401, so `redirect: 'error'` is free. Asserted, because an init option no test inspects is invisible to the suite.

**4. The client failed OPEN on an unknown scope while the service fails CLOSED.** Any value that wasn't the exact literal degraded to a **full purge** — the widest blast radius. Adding a third scope later would have silently meant "delete everything" instead of failing to compile. Replaced with an exhaustive `switch` + `never` check.

### Confirmed clean by the audit (verified, not reasoned)

- **Secret handling.** Never logged, serialised or echoed; the new Axiom payloads carry only `imageKey`/`scope`/`status`/`safeError`. Client exposure impossible — no `NEXT_PUBLIC_` prefix, and every client import of `image.service` is `import type`.
- **Backward compatibility, empirically.** Raw outbound requests captured against a local server: base (no `headers` key) and this PR with the secret unset (`headers: {}`) produce an **identical header set, method and URL**.
- **Fire-and-forget preserved.** The promise pushed into `cacheRefreshPromises` resolves in <0.15 ms in every case; unhandled-rejection behaviour is identical to base.
- **No infra disclosure added.** *Separately and pre-existing at base:* `image.service.ts:297-308` discloses bucket and region details in this public repo — worth a follow-up, out of scope here.

### Rollout ordering (unchanged, now verified against the deployed service)

`keep` is not parsed by the running version and its auth gate short-circuits while the delete flag is off — so this is **inert today** and **web-first deployment is safe**. `IMAGE_CACHER_ADMIN_SECRET` must be configured before that flag flips; finding 1 above is the test that would catch a silent regression there.

### Not changed

Prettier flags `post.service.ts` and `update-post-image-hidemeta-bust.test.ts`, but **both already fail at `origin/main`**, so reformatting would be pure diff noise. The added test file passes, which is what the added-files gate covers.

PR #3885, opened 2026-08-13
fix(placement): make the image buzz counter recoverable when an emit is lost

JustMaier:
Closes the follow-up from #3849: [868kqqabr](https://app.clickup.com/t/868kqqabr).

## The problem

An approved placement's Buzz reached the image's counter through a best-effort
emit after a committed settle, outside any transaction, swallowing its own
errors. A failed emit — or a pod killed between the commit and the emit — lost
that placement's Buzz permanently, and nothing could tell a lost emit from a
placement that was never approved.

## What this does

`Placement.metricCountedAt`, written **only after** the emit lands. A NULL on a
settled placement is therefore a work queue, and a new `placement-sweep-uncounted`
job (every 5 minutes) counts whatever is still missing it.

That ordering is the whole design. Marking first would record a failed emit as
counted and lose the Buzz for good; marking after leaves a window where a pod
death means one placement is counted twice. A recoverable over-count on a display
total beats an unrecoverable loss.

### Option B, not CDC

Per the ticket's recommendation. The correctness comes from a column written
beside the money, so it needs no publication membership or `REPLICA IDENTITY FULL`
we do not own, and no handler that has to tell `pending -> approved` from the four
paths that write `removed` — getting that wrong makes a counter that decrements,
which is the one thing it must never do.

The live emit from #3849 stays as the liveness half, now writing the marker.

### Two traps the sweep handles

- **One summed event per (image, placer), not one per placement.** Verified
  against prod DDL: `entityMetricEvents_month` is a `SharedReplacingMergeTree`
  ordered on `(entityType, entityId, metricType, userId, createdAt)` with
  `createdAt` at **second** precision, and the daily aggregates recompute from it
  with `argMax` per that key. Two events for the same image and placer inside one
  second collapse and the second payment disappears. A bulk approval is exactly
  that shape. Summing first makes the collision unreachable.
- **`removed` is swept alongside `approved`.** Those were approved first, and the
  counter counts the approval and never reverses. Skipping them would make the
  counter depend on cron timing.

## Migration — applied by hand, and it needs a second pass

`20260813120000_placement_metric_counted_at`. Order matters and the file states it:

1. Run steps 1-3 **before** deploying.
2. Deploy.
3. **Re-run step 2 (the backfill UPDATE) once** after the deploy is healthy.

Between step 1 and the deploy, the old code keeps approving and emitting without
writing the column, so those rows would land in the sweep's queue already counted
and be counted twice on the first tick. Step 3 closes that window. Everything
already settled is marked counted rather than re-emitted: the shipped emit either
landed or was lost and the two are indistinguishable from here, so whatever was
lost before today stays lost.

## Verification

- `pnpm run lint` — no findings in the changed files
- `pnpm run db:check-generated` — regenerated client committed
- New suite: 10 tests; placement/sticker/remix/escrow suites: 199 pass
- Positive control: reverting the grouping to one event per placement fails with
  `expected "vi.fn()" to be called 1 times, but got 2 times`


JustMaier:
Pushed `a9b17c9554` after finding a bug that made this PR's central claim false.

`Tracker.send` dispatches with `void this.sendWithRetry(...)`. So `await updateEntityMetricDetached(...)` returned once the row had been handed to a fire-and-forget POST, not once the tracker had accepted it — and stamping `metricCountedAt` after that recorded a possibly-dropped event as counted. That is strictly worse than the loss it replaced: the row then looks accounted for forever and the sweep never looks at it again.

The commit adds an opt-in `awaitDelivery` on the tracker that awaits the retry loop and raises on a failure the retries could not fix. Default is unchanged for every existing caller — a view or a reaction must not add a round trip to a response. `sendWithRetry` became a loop rather than recursion so the raise has somewhere to throw from that its own retry `catch` cannot swallow; attempt count and backoff schedule are identical.

Second thing it fixes: `updateEntityMetric` returned void, so the `disable-app-entity-metrics` flag being on was indistinguishable from a successful emit. With the flag on the sweep would have stamped every placement in the window and lost all of them. It now returns whether the event was delivered, and the sweep leaves undelivered rows in the queue.

Cost, stated: an approval that reaches the counter now waits on the tracker, up to ~1.1s of backoff if it is failing. That is after the money has already moved, and `recordPlacementTip` still cannot throw.

Verification re-run: typecheck 0 errors; `src/server/services/__tests__/` + `src/server/controllers/__tests__/` 3024 pass; `eslint` on the changed files clean except `tracker.ts`, whose 9 findings are identical on `main` (verified by running eslint on the same file in the primary worktree).

JustMaier:
Reworked at `8668201feb` after two adversarial passes. Three of the findings were real bugs in what I had written; one of them was a money bug.

## The money bug: `removed` is not the same as "was approved"

The sweep selected `status IN ('approved','removed')`, on the reasoning that a removal only happens to a live placement. That is false. All three `removeBy*` settle actions are also driven straight from `pending` — `removePlacementsByUser` and `removePlacementByModerator` both act on pending rows — and that escrow is **forfeited**: no owner is paid, the placement was never live. A moderator suspending an abusive placer with 40 pending 500⚡ placements would have added 500⚡ to each of 40 images' public counters. Moderating abuse would have inflated the number the abuse targeted, and it breaks the "accepted placements only" rule the module documents.

`takenDownAt` is the discriminator — only a takedown of an already-settled placement sets it — and the sweep now uses it.

## The instant emit is gone

It could not be made safe. Awaiting delivery on it puts an untimed network call inside a money mutation, and `actOnStickerPlacements` loops up to 50 of them in one request: a tracker returning fast 500s turns a bulk approve into a minute of sleeping and a gateway timeout, on placements that were all settled and paid. Not awaiting it records a dropped event as counted, which is the bug this PR exists to end. And it emitted one event per placement, which the pipeline collapses when a placer has two on one image inside the same second — a bulk approve produces exactly that shape, and the stamp would have made that loss permanent.

Grouping is only possible in a sweep, so the sweep is the only emitter now. **The counter moves on the next tick rather than instantly.** That is the cost the ticket names for this option and recommends accepting.

## Claim, then count

The job lock is not mutual exclusion: it fails open when Redis is unavailable, releases mid-run at its ceiling, and a single dropped 8s refresh against a 10s TTL hands it to the next tick cleanly. Two runs then read the same unstamped rows and both emit before either stamps — and the counter never comes back down, so that over-count is permanent.

`metricClaimedAt` (a second column) plus `FOR UPDATE SKIP LOCKED` makes two runs take disjoint sets. Claiming into `metricCountedAt` directly would have inverted the trade back to loss; two columns mean a run that dies between claiming and confirming leaves rows claimed, and they come back once the claim goes stale.

## Ordering is now enforced, not documented

The old instruction was "apply, deploy, re-run the backfill". That cannot race a 20-30 minute canary: new pods sweep within 5 minutes, and everything the old code counted-but-did-not-stamp gets re-emitted with a fresh timestamp — a different second, so it **sums** rather than collapsing, roughly doubling every counter touched.

The sweep is now behind `placement-metric-sweep`, default off. Order is **apply → deploy → re-run the backfill → turn the flag on**, and the migration says so.

## Smaller ones from the same passes

- `awaitDelivery` bounds each attempt with its own `AbortSignal`. One signal for the whole loop is worse than none — `fetch` rejects immediately on an aborted signal, so the first slow attempt would consume the other two retries.
- A group deferred for colliding with one already emitted this run has its claim **released**, not just skipped; otherwise it sat out the 15-minute stale window for a collision the next tick could not have. A run that only deferred no longer trips the new "claimed work, counted none of it" warning.
- `sendWithRetry` has tests now. It had none: reverting `throw failure` to `return` deletes the entire durability guarantee and ships green.
- No tracker configured reports not-delivered instead of raising — that is every dev and preview environment, and it belongs with the kill-switch, not with failure.
- The index is partial on the uncounted tail and keyed on `resolvedAt` alone. Leading with `status` gives a BitmapOr over the `approved OR (removed AND takenDownAt)` predicate, which destroys the index ordering and puts a sort of the whole backlog in front of every LIMIT — at exactly the moment there is a backlog.

**Known and not fixed:** a placer in `metricExcludedUsers` (344 active) is accepted by the tracker and dropped by the aggregation, so their placement is stamped counted and never lands. Zero overlap with current placers. It is the same treatment their tips already get, so I have left it rather than invent a second rule here.

Verification: typecheck 0 errors; 156 tests across the placement, sticker, remix and tracker suites; eslint clean on the changed files (`tracker.ts`'s 3 findings are identical on `main`); `db:check-generated` clean.

PR #3887, opened 2026-08-13
fix(tests): stop the lazy-carousel browser test collecting ZERO tests

### LINKED ISSUE #3886: ImagesAsPostsCardLazyCarousel.browser.test.tsx collects 0 tests on main — a one-key vi.mock outgrown by #3859
`ImagesAsPostsCardLazyCarousel.browser.test.tsx` collects **0 tests** on `main` — a one-key `vi.mock` outgrown by #3859

## Symptom

On `main` today, this browser test fails to **import**, so it contributes **0 of its 9 tests**:

```
Failed to import test file src/components/Image/AsPosts/ImagesAsPostsCardLazyCarousel.browser.test.tsx
Caused by: SyntaxError: The requested module '/src/store/generation-graph.store.ts'
  does not provide an export named 'fetchGenerationData'
```

The reason this is worth an issue rather than a quick patch is the **shape of the report**: the run says `Test Files 1 failed | 137 passed (138)` with **`Tests 1409 passed` and zero failed tests**. There is no red assertion anywhere — the total simply drops by 9. A reassuring zero is indistinguishable from a suite wired to nothing.

## Cause

`fetchGenerationData` is not missing; it exists at `src/store/generation-graph.store.ts:252`. The test wholesale-mocks the store with a **one-key factory** (line 172):

```ts
vi.mock('~/store/generation-graph.store', () => ({ generationGraphPanel: { open: vi.fn() } }));
```

That mock was **correct until #3859** (`fc5216ba90`). Before it, `ImagesAsPostsCard.tsx` imported the store directly and needed exactly that one name. #3859 removed that line:

```diff
-import { generationGraphPanel } from '~/store/generation-graph.store';
+import { RemixMenu, isRemixMenuVisible } from '~/components/Image/Remix/RemixMenu';
```

and the new edge reaches the same store two hops away, pulling **four** names:

```
ImagesAsPostsCardLazyCarousel.browser.test.tsx
  -> ImagesAsPostsCard.tsx:45
  -> Image/Remix/RemixMenu
  -> Image/Remix/remix.utils.ts:7-12
     import { fetchGenerationData, generationGraphPanel,
              generationGraphStore, withExternalFetch } from '~/store/generation-graph.store';
```

Vitest browser mode serves native browser ESM, so named exports are checked at **link time**. The one-key factory makes the module export only `generationGraphPanel`; linking `remix.utils.ts` against it throws before any test runs. No cycle, no tree-shaking, no ordering effect — a wholesale mock outgrown by a new graph edge.

## Evidence it is `main`, not a branch

Bisected, with counts read from output rather than exit codes, each configuration repeated:

| Tree | Runs | Result |
|---|---|---|
| `fc5216ba90^` (`f19574a9cb`) | 4 | **`1 passed` / 9 tests passed** |
| `fc5216ba90` (#3859) | 1 | red, identical error |
| clean `main`, single file | 4 | red, identical error |
| clean `main`, full `component` project | 1 | `1 failed \| 137 passed (138)` / `1409 passed` |

Zero variance across repeats. Instrument validated both ways: a sibling browser test run alongside passed (the browser genuinely launches and executes), and the negative control reproduces the exact CI error — this is not the "runner never started" shape.

## Fix

Two options, both verified 9/9 locally:

- **A (minimal)** — add `generationGraphStore`, `fetchGenerationData`, `withExternalFetch` to the factory. Re-breaks the next time anything in that graph imports a fifth name.
- **B (repo-prescribed)** — `importOriginal`-spread. ⚠️ It fails on its own with `[vitest] There was an error when mocking a module`, because loading the real store evaluates `import { trpcVanilla } from '~/utils/trpc'` and this file's *other* wholesale mock (line 81) omits `trpcVanilla`. Adding `trpcVanilla: {}` to that factory makes B pass. Two edits, immune to future export additions.

Worth noting the mock is a pure **link-time shim** here: the test mocks `useFeatureFlags: () => ({ imageGeneration: false })`, and both render sites are guarded `features.imageGeneration && isRemixMenuVisible(image)` (`ImagesAsPostsCard.tsx:286,405`), so `RemixMenu` never renders and the stubs are never invoked. Extending the mock does not mask a behavioural regression.

## The systemic part

`local-rules/no-wholesale-module-mock` already exists and its message describes this exact failure — *"the whole FILE fails to load: 0 tests collected, no failing assertion, silently 'green'"*. Running eslint on this file, **the rule does fire — at line 81**. But `.eslintrc.js:96` configures it as:

```js
'local-rules/no-wholesale-module-mock': ['error', { modules: ['~/utils/trpc'] }],
```

so it is **blind to line 172**, the mock actually costing 9 tests. An allowlist of one module cannot catch the next store.

Measured on current `main`: **84** browser test files mock `~/utils/trpc`, of which **57** do so wholesale (no `importOriginal`). Each is one new export away from zeroing itself the same way.

Suggested ordering:
1. Fix this test.
2. Widen the rule's `modules` list, or drop the allowlist and let it apply wherever it can prove a factory is wholesale.
3. Assert a **minimum expected browser test-file / test count** in CI. An import failure contributes 0 tests and no red test, so a count floor is the only thing that can see it.

ZacxDev:
Fixes #3886.

`ImagesAsPostsCardLazyCarousel.browser.test.tsx` fails to **import** on `main`, so it contributes **0 of its 9 tests**.

The reason it went unnoticed is the reporting shape: the run says `Test Files 1 failed | 137 passed (138)` with **`Tests 1409 passed` and zero failed tests**. There is no red assertion anywhere — the suite total just drops by 9. A reassuring zero is indistinguishable from a file wired to nothing.

## Cause

`fetchGenerationData` is not missing (`generation-graph.store.ts:252`). The file wholesale-mocked the store with a **one-key factory**, which was correct until #3859 (`fc5216ba90`) removed this file's direct one-name import:

```diff
-import { generationGraphPanel } from '~/store/generation-graph.store';
+import { RemixMenu, isRemixMenuVisible } from '~/components/Image/Remix/RemixMenu';
```

The new edge reaches the same store two hops away and pulls **four** names:

```
test → ImagesAsPostsCard.tsx:45 → Image/Remix/RemixMenu
     → Image/Remix/remix.utils.ts:7-12
       { fetchGenerationData, generationGraphPanel, generationGraphStore, withExternalFetch }
```

Browser mode serves native ESM, so named exports are checked at **link time**. The one-key factory made the module export only `generationGraphPanel`; linking `remix.utils` against it threw before any test ran. No cycle, no ordering effect — a wholesale mock outgrown by a new graph edge.

## The fix

`importOriginal`-spread rather than adding the three missing names: a name-listing factory breaks again the moment the graph reaches a fifth export, which is the same defect reset. This is the form `local-rules/no-wholesale-module-mock` prescribes, including its instruction to use a type-only namespace import instead of `typeof import(...)` (which `@typescript-eslint/consistent-type-imports` rejects).

The second edit is **coupled, not cosmetic**: `trpcVanilla: {}` in this file's trpc mock. Spreading the real store evaluates its `import { trpcVanilla } from '~/utils/trpc'`, and without that key the mock throws `[vitest] There was an error when mocking a module`. Confirmed by control — removing just that line brings the error back.

**This does not mask a behavioural regression.** The test mocks `useFeatureFlags: () => ({ imageGeneration: false })` and both render sites are guarded `features.imageGeneration && isRemixMenuVisible(image)` (`ImagesAsPostsCard.tsx:286,405`), so `RemixMenu` never renders — the mock is a pure link-time shim and the stubs are never invoked.

## Verification

Counts read from output, never exit codes:

| Check | Result |
|---|---|
| this file, before | `1 failed` / **0 tests collected**, exact CI error |
| this file, after | `1 passed` / **9 tests passed** |
| full `component` project | **138 files passed (138) / 1418 tests passed (1418)** — was 137/138 and 1409, exactly the +9 |
| instrument control | a sibling browser test passes alongside, so the browser genuinely launches — not the "runner never started" shape |
| coupling control | drop `trpcVanilla: {}` → the mocking error returns |
| eslint | 3 errors, **all pre-existing and unchanged**; the one I introduced (`typeof import()`) is fixed |
| prettier | clean |

## Deliberately not fixed here

`local-rules/no-wholesale-module-mock` already describes this exact failure, and it **does** fire on this file — at the *trpc* mock. But `.eslintrc.js:96` scopes it to `modules: ['~/utils/trpc']`, so it is blind to the store mock that actually cost the 9 tests. Measured on current `main`: **84** browser tests mock trpc, **57** of them wholesale. An allowlist of one module cannot catch the next store.

#3886 carries that follow-up, along with the suggestion of a **minimum expected browser test-file / test count** in CI — an import failure contributes 0 tests and no red test, so a count floor is the only thing that can see this class.

### DIFF
--- a/src/components/Image/AsPosts/ImagesAsPostsCardLazyCarousel.browser.test.tsx
+++ b/src/components/Image/AsPosts/ImagesAsPostsCardLazyCarousel.browser.test.tsx
@@ -1,9 +1,9 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
-// Type-only namespace import for the `importOriginal` mock below. Erased at compile time, so
-// it does not participate in the mock hoisting. `@typescript-eslint/consistent-type-imports`
-// rejects the `typeof import('...')` form, which is why this is a named namespace.
-import type * as GenerationGraphStore from '~/store/generation-graph.store';
+// Type-only namespace import for the `importOriginal` mock below. Erased at compile time, so it
+// does not participate in mock hoisting. `@typescript-eslint/consistent-type-imports` rejects
+// the `typeof import('...')` form, which is why this is a named namespace.
+import type * as TrpcUtils from '~/utils/trpc';
 
 // =============================================================================
 // Gallery lazy per-post carousel (`galleryLazyPostImages`).
@@ -82,15 +82,19 @@ const mocks = vi.hoisted(() => {
 });
 
 // --- tail fetch ---------------------------------------------------------------
-// `trpcVanilla` is required by the generation-graph store, which the mock below now loads for
-// real via `importOriginal`. Without this key that load throws
-// "[vitest] There was an error when mocking a module" — the two mocks are coupled.
-vi.mock('~/utils/trpc', () => ({
+// `importOriginal`-spread, not a hand-written factory. This mock applies to the WHOLE module
+// graph, so every module the test transitively reaches must find the names it imports here.
+// `~/utils/trpc` has 12 exports and a factory listing only the two this file uses breaks the
+// moment the graph reaches a third: browser mode serves native ESM, so a missing named export
+// is a LINK-time error that kills the entire FILE — 0 tests collected, no failing assertion,
+// and the suite total silently drops. That is exactly how this file lost all 9 of its tests
+// (#3859 routed it through Remix/remix.utils, which needs `trpcVanilla` among others).
+vi.mock('~/utils/trpc', async (importOriginal) => ({
+  ...(await importOriginal<typeof TrpcUtils>()),
   trpc: {
     image: { getInfinite: { useQuery: mocks.getInfiniteUseQuery } },
     useUtils: () => ({}),
   },
-  trpcVanilla: {},
 }));
 
 // --- gallery context (filters + browsing level + hidden-pref inputs) ----------
@@ -176,17 +180,6 @@ vi.mock('~/providers/FeatureFlagsProvider', () => ({
 }));
 vi.mock('~/components/TrackView/track.utils', () => ({
   useTrackEvent: () => ({ trackAction: vi.fn().mockResolvedValue(undefined) }),
-}));
-// `importOriginal`-spread, NOT a wholesale factory. This module is reached two hops away
-// (ImagesAsPostsCard -> Remix/RemixMenu -> Remix/remix.utils), and remix.utils imports FOUR
-// names from it. A factory listing only the names this file happens to know about breaks the
-// moment the graph reaches a fifth: browser mode serves native ESM, so a missing named export
-// is a LINK-time error that kills the whole file — 0 tests collected, no failing assertion,
-// and the suite total silently drops. That is what happened here (#3859 replaced this file's
-// direct one-name import with the RemixMenu edge; see the linked issue).
-vi.mock('~/store/generation-graph.store', async (importOriginal) => ({
-  ...(await importOriginal<typeof GenerationGraphStore>()),
-  generationGraphPanel: { open: vi.fn() },
 }));
 
 // Import AFTER the mocks are registered.
--- a/src/components/Image/AsPosts/ImagesAsPostsCardLazyCarousel.browser.test.tsx
+++ b/src/components/Image/AsPosts/ImagesAsPostsCardLazyCarousel.browser.test.tsx
@@ -1,5 +1,9 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { page, userEvent } from 'vitest/browser';
+// Type-only namespace import for the `importOriginal` mock below. Erased at compile time, so
+// it does not participate in the mock hoisting. `@typescript-eslint/consistent-type-imports`
+// rejects the `typeof import('...')` form, which is why this is a named namespace.
+import type * as GenerationGraphStore from '~/store/generation-graph.store';
 
 // =============================================================================
 // Gallery lazy per-post carousel (`galleryLazyPostImages`).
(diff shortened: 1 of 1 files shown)


ZacxDev:
Updated in `4bf2a2bd23` after an adversarial audit, which found **most of the first commit inert**.

Deleting the `~/store/generation-graph.store` mock **entirely** still leaves the file at 9/9. The trpc mock applies to the whole module graph, so once `trpcVanilla` is present the real store links and evaluates fine — it never needed mocking. Of the 19 lines shipped, 12 did nothing, including a 7-line comment explaining a mechanism that was not load-bearing. In a file the repo treats as canonical, that would have been copied.

So the fix now happens at the module the graph actually links against:

- **drop** the store mock and its type import
- **convert** the `~/utils/trpc` mock to an `importOriginal`-spread

Better on three counts:

1. **Fixes the real location.** Real `~/utils/trpc` has 12 exports; the factory supplied 2. Spreading it means no future export addition can link-break this file. The first commit left that factory wholesale and added only the name missing *today* — the same defect reset.
2. **Clears a standing eslint error** instead of working around it. `local-rules/no-wholesale-module-mock` fires on that factory (unsuppressed `error`, present on `main`). 3 → 2 errors; both survivors pre-existing and untouched.
3. **Removes a coupling whose only guard was a comment** — and whose failure mode was *worse than the original bug*: breaking it produced an Unhandled Rejection naming neither module nor export, with no `Test Files`/`Tests` summary at all. The failure is now a plain link error naming the module and the symbol.

Re-verified (counts read from output, never exit codes):

| Check | Result |
|---|---|
| this file | `1 passed` / **9 tests passed** |
| full `component` project | **138 files passed (138) / 1418 tests passed (1418)** |
| typecheck | 0 type errors |
| eslint | **2 errors (was 3)** — wholesale-mock error gone |
| prettier | clean |
| control | revert the trpc mock to a wholesale factory → `SyntaxError: '/src/utils/trpc.ts' does not provide an export named 'trpcVanilla'`, 0 tests collected — so the remaining edit is load-bearing |

Diff vs `main` is now **+13/−2**, down from +19/−1.

One caveat worth stating: the audit verified these as **two separate changes**, never as a combination. Measured here as one.

The `preview / smoke-tests` red on this PR is the `/images` meili-backed feed assertion, which fails identically on other open PRs — pre-existing, not from this change.

PR #3888, opened 2026-08-13
fix(auth): authenticate the generation-token cookie and key it off a secret we actually set

JustMaier:
## What

`src/server/utils/cookie-encryption.ts` protects `generation-token`, which holds an orchestrator API key minted as a `System` key with a ~1h life. Two problems, found while building remix provenance (#3871, which deliberately did not reuse this helper):

**The key was a public literal in practice.** `process.env.COOKIE_SECRET_KEY || 'default_secret_key'` - and `COOKIE_SECRET_KEY` is in no env file, no `.env.example`, and not in `src/env/server-schema.ts`. I checked production directly rather than assuming: it is unset there too. So the committed fallback was the live key in every environment.

**The cipher had no auth tag.** `aes-256-ctr` is malleable - a bit flipped in the ciphertext flips the same bit in the plaintext, with nothing to detect it.

## What changed

- **`aes-256-gcm`**, token format `c1.<iv>.<ct>.<tag>` in base64url. Short auth tags are rejected: Node accepts GCM tags of 4, 8 and 12-16 bytes and verifies against only what it is given, so accepting one would downgrade a 128-bit authenticator to a 32-bit one. (The test that pins this fails with `expected 'orchestrator-api-key-abc123' to be null` when the length check is removed - Node really does verify against 4 bytes.)
- **Key derived from `NEXTAUTH_SECRET`**: `sha256("${NEXTAUTH_SECRET}:cookie-encryption:v1")`, the same shape `remix-provenance.ts` already uses. Domain-separated so a cookie can never be confused with anything else keyed by the same secret; bumping the version rotates it.
- **`COOKIE_SECRET_KEY` is dropped entirely**, not added to the schema.

### Why dropped rather than made required

The ticket asked for it required with no default, so a missing value fails the boot. The cost of that is larger than it looks: `src/env/server.ts` throws on a schema parse failure and `next.config.mjs` imports it, so a required variable would fail **the build** - in prod, preview, CI and every dev worktree - until someone provisioned the secret in each. `NEXTAUTH_SECRET` is already `z.string()` required and set everywhere, so deriving from it removes the public-literal key with nothing to provision and no coordinated infra change. Justin's call, made on that basis.

Practical effect is the same as required-and-set: there is no path left where this encrypts under a value that is public.

## Compatibility

Cookies written under the old key and algorithm no longer decrypt. That is intended, and it is safe: all three read paths - `get-orchestrator-token.ts` (whose cookie branch is dead today, `TOKEN_STORE` is hardcoded to `'redis'`) and the two `pages/api/admin/orchestrator` endpoints - treat a null read as "no cookie" and re-mint. Nothing 500s. The underlying `ApiKey` rows expire at `maxAge + 5` (~1h), so nothing outlives the hour either way. No migration, no manual step.

Cookie attributes (`Path`, `Secure`, `HttpOnly`, no `Domain`) are unchanged - this cookie stays host-only, so `.com` and `.red` keep their own, exactly as before.

## Deliberately not done

The ticket's optional third item - one shared `encryptJson`/`decryptJson` used by both this and `remix-provenance.ts`. `remix-provenance` carries its own versioning, replay checks and expiry on top of the primitive; folding them together means editing a shipped security-sensitive path for a DRY win the ticket itself calls the least important reason to do this. Left alone on purpose.

## Verification

- `pnpm run typecheck` - 0 errors
- `pnpm run lint` - clean on the changed files
- `pnpm run test:unit:run` - **compared against a baseline rather than read as a total**, because the local count is unstable: the same suite on the same code gave 88 failures across 21 files on one run and 34 across 7 on the next. So both sides were re-run with `--reporter=json` in the same worktree - same flake Node, same cwd - with the baseline produced by checking out `origin/main`'s `cookie-encryption.ts` in place and removing the new test file.

  | | files | collected | failed | failing files |
  |---|---|---|---|---|
  | this branch | 1001 | 15741 | 34 | 7 |
  | baseline (module reverted) | 1000 | 15733 | 39 | 12 |

  The branch's 7 failing files are a **strict subset** of the baseline's 12 - the baseline flaked harder - so there is no failure on this branch that does not also occur without it. Per-file collected counts are identical everywhere except the new test file (+8), and **zero files collected zero tests on either side**. That last check is the one a grep over a log cannot make: a suite that fails to collect contributes no tests and need not name the module that broke it, which was the real risk of adding an `~/env/server` import edge to this module. It did not happen.
- `blocks.router.workflow.test.ts` collected **347** tests on both sides - nonzero, so the submodule was present and the runs were real rather than silent zero-collection passes.
- The suites that actually exercise this module - `src/server/orchestrator/__tests__/`, `logout-cookies`, `remix-provenance`, `blocks.router.workflow`, and the new file - 408 pass, 0 fail.

### The test suite, honestly

8 tests, and they are not all load-bearing. An adversarial reviewer replayed each one against the retired implementation:

- **Fail on a revert** (6 of 8): the tamper test, the truncated-tag test, the legacy-format test, the token-shape regex, the KDF-label test, and the no-key write test.
- **Pass either way** - sanity, and marked as such in the file: the round-trip and garbage-value tests. The old code round-trips too, and hex ciphertext does not contain the plaintext either.
- The KDF test **pins the formula, not the provenance**: it computes its label from the same mocked `env` the module reads, so a module hardcoding that label would pass it. Named to say so. It does still reject the derivation this replaced.

Two tests were written during review and then deleted rather than shipped, because each proved to be green for a reason unrelated to the guard it was named for:

- An **IV-length** test. A wrong-length IV fails the auth tag anyway, so removing the guard leaves every test passing - measured. The guard stays as a cheap early reject and the comment says exactly that, rather than implying it carries the weight the tag check carries.
- A **no-key read** test. Reading under a different secret returns null whether or not the key-usability check exists, so it asserted a property every correctly-keyed scheme has. The write side is where that branch is observable, and that test survives both mutations that remove the guard.

Mutation-tested rather than assumed. Changing the KDF label gives `expected null to be 'orchestrator-api-key-abc123'`; removing the tag-length check gives `expected 'orchestrator-api-key-abc123' to be null` - Node really does verify against a 4-byte tag if you hand it one. Each failure names a value, and none can hang.


JustMaier:
Superseded by #3896, which deletes this code instead of fixing it. (Auto-closed when the branch was removed.)

Koen confirmed he has never used either of the two admin orchestrator endpoints, and they were the only live consumers of the helper - `get-orchestrator-token.ts`'s cookie branch was already unreachable behind a constant false. So there is no primitive left to key correctly, which is a better outcome than keying it correctly.

PR #3890, opened 2026-08-13
fix(placement): page the sticker and remix review queues

JustMaier:
[868kqq73p](https://app.clickup.com/t/868kqq73p). Follow-up to #3857, which added the "this list is truncated" line; this replaces it with real paging.

## The problem

Neither review queue paged. Both stopped at 50 with no cursor, so a busy owner
saw a list that looked complete while the escrow behind everything past the 50th
sat untouched until it expired. The failure is money silently returning to
submitters for a decision the owner was never offered, not a UI limit.

## What this does

Keyset paging on both queues, matching what the gallery already does.

**The cursor carries `createdAt` AND `id`.** Two placements can share a
millisecond; a `createdAt`-only cursor either repeats one across pages or steps
over it — and the one stepped over is precisely the entry nobody ever reviews.
An offset has the same hole for a different reason: answering a placement takes
it out of the queue, so page two of an offset walk skips as many entries as the
owner just cleared.

**The next cursor is read off the last row of the raw page**, before rows are
dropped for an unreadable payload or an image that no longer resolves. Taking it
from what was *returned* would point the next page at an earlier row and serve
the gap twice. The `truncated` flag this replaces had the mirror-image bug — it
reported a full queue as complete once the filter took the count under the cap,
which is why #3857 measured it before filtering.

A malformed cursor starts a fresh first page rather than putting NaN in a query.

Count badges that can only see one page now read `50+` instead of `50`.

Out of scope, stated rather than hidden: the submitter's own "Sent" list
(`getMyRemixGallerySubmissions`) still caps at 50 and keeps its truncation
notice. Nothing expires behind it — it is the submitter's own view of their own
holds — so it did not carry the same cost.

## Verification

- `npx eslint` over the 11 changed files — no issues
- `remix-gallery.service.test.ts` — 79 pass, including 4 new paging cases
- Positive control: making the cursor come from the probe row instead of the last
  page row fails with `expected '1767398400000:3' to be '1767312000000:2'`

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260813150000_placement_owner_queue_paging_index/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260813150000_placement_owner_queue_paging_index/migration.sql
@@ -1,0 +1,8 @@
+-- The owner review queue's paging order: filter on (ownerId, status), order by
+-- (createdAt, id). Without it every page rescans and re-sorts all of an owner's
+-- pending rows, and a refetch of a walk N pages deep pays that N+1 times.
+--
+-- CONCURRENTLY, so it cannot run inside a transaction. Safe to apply before or
+-- after the deploy — nothing reads it but the planner.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "Placement_ownerId_status_createdAt_id_idx"
+  ON "Placement" ("ownerId", "status", "createdAt", "id");
--- a/packages/civitai-db-schema/prisma/migrations/20260813150000_placement_owner_queue_paging_index/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260813150000_placement_owner_queue_paging_index/migration.sql
@@ -1,8 +1,24 @@
--- The owner review queue's paging order: filter on (ownerId, status), order by
--- (createdAt, id). Without it every page rescans and re-sorts all of an owner's
--- pending rows, and a refetch of a walk N pages deep pays that N+1 times.
+-- The owner review queue's paging order: filter on (ownerId, surface, status),
+-- order by (createdAt, id). Without it every page rescans and re-sorts all of an
+-- owner's pending rows, and a refetch of a walk N pages deep pays that N+1 times.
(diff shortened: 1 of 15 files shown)


JustMaier:
Correction to the original description: **this PR now carries a migration.** It did not when I opened it, and I said so on the ClickUp task — that is no longer true.

`20260813150000_placement_owner_queue_paging_index` adds `Placement_ownerId_surface_status_createdAt_id_idx`. Hand-applied like everything else here, `CREATE INDEX CONCURRENTLY`, and order-free with respect to the deploy — nothing reads it but the planner, so the code is correct without it and merely slower. The file carries an `indisvalid` check to run afterwards, because `CONCURRENTLY` plus `IF NOT EXISTS` can report success while leaving an invalid index that every write maintains and no query uses.

Why an index at all: paging without one means every page rescans and re-sorts all of an owner's pending rows, and `utils.placement.invalidate()` after each approve refetches every loaded page — so a walk N pages deep pays that N+1 times per click. `surface` is in the key because both queues filter on it; left out it cannot be an index condition, and an owner with 900 pending stickers would touch all 900 heap tuples to fill one 50-row page of remix submissions. It sits second so this does not become a strict prefix of `[ownerId, status]`, which the moderation lookups still use.

Two rounds of adversarial review since the first push. Round one found that a page whose rows were all filtered out — images deleted, unpublished, or still ingesting — returned no items with a cursor still set, and every remix surface rendered "nothing is waiting" over it. That is this branch's own bug, relocated one layer up, and it was real. Round two found that my fix for it had no test: the state at the centre of it was asserted nowhere, so returning `nextCursor: null` from either exit put it straight back with the suite green. Both exits are pinned now on both surfaces, each verified by mutation.

Also since the first push: `isError` branches on all three surfaces (a failed read has no pages, so `hasNextPage` is false and the empty-state guard cannot catch it); cursor parsing rejects `1e21`, `1.5` and `':'`; the sticker queue's own paging got tests, which it had none of; and "Select all" reads "Select all loaded" while more pages exist.

PR #3895, opened 2026-08-13
DO NOT MERGE — fix(placement): sticker control hysteresis (infinite render loop, parked)

JustMaier:
Closes the three items carried from #3865. All three were re-audited against current `main` first; all three still reproduced.

## What was wrong

The sticker's flip/opacity/delete controls live in **two different places in the React tree** — flush against the sticker's top corners while it is at least 124px wide, inside the buy cluster once it is narrower. Crossing that width unmounts one and mounts the other. The opacity `Popover` was uncontrolled, so the crossing destroyed an open slider and dropped keyboard focus with it. A plain browser window resize is enough to trigger it, because the sticker's width is a fraction of the media box.

Worth being explicit, since the ticket proposed hysteresis as the fix: **hysteresis alone does not fix that.** It stops a width parked on the line from flip-flopping, but one deliberate crossing still swaps the layouts. The popover had to stop depending on the element surviving.

## Changes

- **`panelsFitInsideEdges(width, currentlyInside)` / `panelBandFor(width, currentlyInside)`** — 124 to enter the flush layout, 116 to leave it. The second parameter is required with no default, matching this module's existing stance (`flippedButtonOffset`'s `panelBand`) that a caller which forgets it must not silently get the old behaviour. The band exists for the non-monotonic case: `scale * mediaWidth` rounding to 123 and 124 on alternate frames while the sticker sits on the line. A monotonic drag crosses once and was never the problem.

- **`shouldFlipPlaceButton` ranks instead of standing still.** It previously returned `hidden(below) && !hidden(above)`, so with both positions hidden it stayed put. On a narrow draft the buy cluster carries the *only* copy of the controls, so that is a draft with nothing to press until it is dragged. It now takes the less hidden position; ties keep it put; off the top of the viewport scores as disqualifying rather than rankable, since the page may not scroll there.

- **The opacity `Popover`'s open state moves into `DraftSticker`,** with `trapFocus` and `returnFocus`. The element is genuinely rebuilt by the remount, so focus has to be *put back* rather than merely not taken away.

## Deliberate behaviour change worth a reviewer's eye

With the 52px two-line button (the one that lands once the payout copy merges), the button will now flip onto a position clipped at the top of the carousel viewport in preference to one wholly behind the tray — 12 visible pixels beating zero. That is the ranking rule doing what it says, and there is a test naming it.

## Not fixed, deliberately

The third carried item — panels inverting at rotation 180 in the wide branch — reproduces, but the rotate knob and all four resize handles invert with them by design. Making only the pills screen-aware makes them the odd chrome out, and counter-rotating them breaks the panel-band assumption the buy button clears. Left as a won't-fix rather than silently dropped.

## Verification

- `pnpm run typecheck` — 0 errors. `eslint` and `prettier` clean on the four changed files.
- Sticker unit + `blocks.router.workflow.test.ts`: **426 passing** (that file collected 347, so the submodule is real and the run means something). Sticker component: **23 passing**.
- Full unit suite: 35 failures, all of which reproduce identically on clean `main` in the same worktree. The one exception passes in isolation on both sides and has no import path from `Sticker/`.
- **Every added test was mutation-checked.** Reverting the controlled popover gives `expected +0 to be 1` on both crossing directions; collapsing the band to a single threshold gives `expected +0 to be 36` and `expected true to be false`; restoring the old both-hidden rule gives `expected false to be true`. All in milliseconds, none as a timeout.

An adversarial review over this diff refuted the oscillation concern on three independent grounds (the layouts are `absolute`, so neither can feed the measured `offsetWidth`; the threshold map is monotone with no 2-cycle; and the band value is identical on both sides of every crossing frame), and confirmed via Mantine 7.17.8's source that a controlled `Popover` attaches no competing `onClick` to its target.

### DIFF
--- a/src/components/Sticker/DraftSticker.browser.test.tsx
+++ b/src/components/Sticker/DraftSticker.browser.test.tsx
@@ -1,0 +1,225 @@
+import { describe, expect, test, vi } from 'vitest';
+import { page } from 'vitest/browser';
+import type * as PlacementUtil from '~/components/Sticker/placement.util';
+import type * as StickerUtil from '~/components/Sticker/sticker.util';
+// `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
+import { renderWithProviders } from '../../../test/component-setup';
+import { DraftSticker } from '~/components/Sticker/DraftSticker';
+import type { StickerTreatment } from '~/components/Sticker/treatments/sticker-treatments';
+import type { StickerDraft } from '~/store/sticker-placement-draft.store';
+
+/**
+ * The two control layouts sit in different places in the tree — flush against
+ * the sticker's own top corners while it is wide enough, inside the buy cluster
+ * once it is not — so crossing the width that decides between them unmounts one
+ * and mounts the other. Everything here is about what that crossing must not
+ * cost: an open opacity slider, and a layout that changes its mind on every
+ * frame of a resize.
+ */
+
+/**
+ * Narrow enough to sit inside the browser-mode viewport. A media box wider than
+ * the window puts the sticker's controls off the side of it, and a click on
+ * something that cannot be scrolled into view fails as a 15s actionability
+ * timeout rather than as anything about the sticker.
+ */
+const MEDIA_BOX_WIDTH = 260;
+
+// Wide enough for the flush panels (182px), narrow enough to be well under the
+// leave width (78px), and inside the hysteresis band (117px). Fractions of the
+// media box, which is how a draft's scale is stored.
+const WIDE = 0.7;
+const NARROW = 0.3;
+const IN_BAND = 0.45;
+
+vi.mock('~/components/Sticker/placement.util', async (importOriginal) => ({
+  ...(await importOriginal<typeof PlacementUtil>()),
+  useCreateStickerPlacement: () => ({ mutate: vi.fn(), isPending: false }),
+}));
+
+// The button pulls in the Buzz balance, a tRPC query and a purchase path, none
+// of which this is about. A plain button keeps the cluster's shape.
+vi.mock('~/components/Buzz/BuzzTransactionButton', () => ({
+  BuzzTransactionButton: ({ label }: { label: string }) => <button type="button">{label}</button>,
+}));
+
+vi.mock('~/components/EdgeMedia/EdgeImage', () => ({
+  EdgeImage: ({ alt }: { alt: string }) => (
+    // A real EdgeImage has no intrinsic size until it loads, and the sticker's
+    // height feeds the flip decision. A fixed box makes both measurements
+    // deterministic without touching the width, which is what is under test.
+    <div aria-label={alt} style={{ width: '100%', height: 80 }} />
+  ),
+}));
+
+const draft = (scale: number): StickerDraft => ({
+  id: 'draft-1',
+  imageId: 1,
+  cosmeticId: 2,
+  x: 0.5,
+  y: 0.5,
+  scale,
+  rotation: 0,
+  flip: false,
+  opacity: 1,
+});
+
+/** No plate, no edge, no animation — nothing that would change the sticker's box. */
+const PLAIN: StickerTreatment = { key: 'none', label: 'None', note: '' };
+
+const art = {
+  id: 2,
+  slug: 'test-sticker',
+  url: 'sticker.png',
+  animated: false,
+} as unknown as StickerUtil.ResolvedSticker;
+
+const renderDraft = async (scale: number) => {
+  const result = await renderWithProviders(
+    <div style={{ position: 'relative', width: MEDIA_BOX_WIDTH, height: 400 }}>
+      <DraftSticker
+        draft={draft(scale)}
+        art={art}
+        selected
+        dressed={PLAIN}
+        price={100}
+        ownerShare={undefined}
+        ownerUsername={undefined}
+        onGesture={() => true}
+      />
+    </div>
+  );
+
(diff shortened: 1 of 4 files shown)


JustMaier:
Three adversarial review rounds on this diff; recording what they changed, including where the reviewer was right about my own claims.

**Round 2 caught a fix that was theatre.** `returnFocus` beside `trapFocus` does nothing in the case this PR exists for: Mantine's `useFocusReturn` captures the return target inside `useDidUpdate`, which skips its first invocation, so a Popover remounting with `opened` already true captures nothing and restores to `null`. Verified against `node_modules` before acting. `DraftSticker` now owns the return target via a ref re-attached to whichever control is mounted, and there is a test reading `document.activeElement` — nothing in the suite read focus at all before.

**Round 3 caught my mutation output proving the wrong thing.** I had written that focus "falls to the body". Measured with a diagnostic assertion, it does not: without the fix `document.activeElement` is a non-control `div` with an empty `aria-label`, which is why the failure read `expected ''` and not `expected null`. Same user-visible consequence, wrong claim — corrected, and the assertion now pins element identity against the control queried after the crossing, so it cannot be satisfied by another element with the same accessible name.

**One finding refuted, with evidence.** Round 1's most severe item wanted the tray measured from its visible `max-w-xl` card rather than its full-width root, calling the gutters phantom pixels. The root carries no `pointer-events-none` and hit-tests across the whole viewport, so those gutters swallow a click exactly as the card does — applying that fix would have taught the button to prefer positions the user cannot press. Accepted on review, and the docstring now says the rect is where the button cannot be *pressed* rather than what the tray *paints*, since the painting wording is what produced the wrong finding.

**A comment of mine invented a mechanism.** I had justified the band with `scale * mediaWidth` alternating between 123 and 124 on successive frames. That does not happen — both inputs are constant at rest. Corrected to the real non-monotonic source: the width itself moving, via a corner-resize gesture or a window-resize drag held near the line.

Also recorded a load-bearing behaviour nothing captured: `preventDefault` on pointerdown suppresses the compatibility `mousedown` (`stopPropagation` alone does not), which is the only reason an open opacity slider survives a drag — Mantine's click-outside listens for `mousedown`. Removing that call would close the popover on every gesture.

Every added test is mutation-checked, each failing in milliseconds with an assertion naming the wrong value.

JustMaier:
**Do not merge. Parked at Justin's call, and it is the right call.**

This branch introduces an infinite render loop. Found in minutes on a real image page:

```
Maximum update depth exceeded.
  at DraftSticker.useCallback[measure] (src/components/Sticker/DraftSticker.tsx:202)
  at DraftSticker.useLayoutEffect (src/components/Sticker/DraftSticker.tsx:269)
```

The three bugs this fixes are cosmetic — a popover that closes early, controls that need a drag to become reachable. The fix can take the page down. That trade is not worth making, and the branch stays parked rather than being patched under pressure.

**Worth recording, because it is the useful part of this failure.** Oscillation was the FIRST thing I sent the adversarial reviewer at, and it cleared the change on three independent arguments: the layouts are `absolute` so neither can feed the measured `offsetWidth`; the threshold map is monotone with no 2-cycle; and the band value is identical either side of every crossing frame. All three are correct in isolation. All three are about `panelsInside` in the steady state, and the loop is not there.

Leading hypothesis, **unverified** — I have not reproduced it in a test yet: `panelBandFor(width, panelsInside)` feeds `flippedOffset`, which is applied as `marginBottom` in the DOM. Before this change the band was a pure function of width, so the value used to derive the candidate boxes always matched the margin already laid out. Now the band depends on state that can change in the same commit, so for one measure the derived `distance` disagrees with the margin actually applied, the compensation in `placeButtonBoxes` is wrong by 36px, and `flipped` can disagree with itself. Nothing in the suite drives a real resize gesture against real layout, so nothing caught it.

The reviewer analysed the loop and never ran one. Neither did I. Three review rounds, five mutation checks, and the defect was in the one property everybody reasoned about instead of executing.

If this is picked up again: reproduce the loop FIRST, in a test that drives an actual gesture across the boundary, and only then decide whether hysteresis is worth having at all. Justin also noted he preferred the previous toolbar layout, so the premise deserves a second look before the mechanism does.

JustMaier:
Closing. Not doing this — Justin's call, and the right one.

The three carried bugs are cosmetic: a popover that closes a beat early, controls that need one drag to become reachable, pills that hang below an upside-down sticker. This branch fixes all three and introduces an infinite render loop that can take the sticker overlay down (`Maximum update depth exceeded` in DraftSticker's measure/layout effect). Trading three annoyances for a crash is not a trade worth making, and patching it under pressure would be worse.

The branch stays on origin for anyone who wants the audit or the tests; it is not a base to build on.

**If this is ever revisited, start with the premise, not the mechanism.** All three bugs are consequences of the wide/narrow control swap existing at all — the controls living in two different places in the React tree is what makes a crossing destroy state. Justin noted he preferred the previous toolbar layout. Deciding whether that swap should exist would make all three moot; rebuilding the geometry would not.

Two things worth keeping from this:
- Hysteresis, which is what the ticket asked for, would NOT have fixed the reported symptom. One deliberate crossing still swaps the layouts and still destroys an uncontrolled Popover. The ticket's proposed fix was aimed at the jitter, and the complaint was about the state loss.
- The render loop was the FIRST property I sent an adversarial reviewer at, and it cleared the change on three independent arguments — all correct about the steady state, and the loop is not in the steady state. Nothing in the test suite drives a real resize gesture against real layout, so nothing caught it. It was found by opening the page. Reproduce it in a test before touching this again.

PR #3899, opened 2026-08-13
fix(apps): gate the Create tab on the author capability, hide a 1-tab sub-nav

ZacxDev:
## What changed

**1. `Create` is gated on the author capability.**
`SUB_NAV_LINKS`' `Create` entry was hardcoded `visible: () => true`. Now that store visibility has widened past the author cohort, a store-visible non-author sees the tab, clicks it, and `/apps/submit` returns `notFound` — its `getServerSideProps` gates on `features.appBlocksAuthor` **and** `isAppDeveloper(session.user, …)`.

The tab now keys off the **same shared predicate** — `isAppDeveloper(user, { appBlocksAuthor })` from `~/shared/utils/app-blocks-access` — so the two gates cannot drift and moderators keep their hard floor (`isModerator || appBlocksAuthor`) without an inlined `isModerator` check.

Threaded as a new **`AppsNavContext`** prop (`{ isAuthor: boolean }`), deliberately kept separate from `AppsNavSummary` rather than folded into it. The two have different provenance *and* different hydration behaviour, and that difference is the load-bearing part of this PR:

| | source | on the first paint |
|---|---|---|
| `AppsNavSummary` | client-only `blocks.getNavSummary` | absent — deferred behind `useIsClient()` |
| `AppsNavContext` | SSR-seeded flags + session | applied |

No new server round-trip: `blocks.getNavSummary` is untouched.

**2. The sub-nav hides entirely below two qualifying tabs** — every viewer, moderators included. This replaces the old *"the two always-on tabs lead so the bar never collapses to fewer than two entries"* rationale, which only held because `Create` was unconditional. A one-entry navigation bar can only link to the page you are already on, while still costing the tab row and its bottom rule on every `/apps/*` surface.

**3. `Jump back in` removed** from the recents rail (`RecentlyOpenedListingsView`). The now-single-child `Group justify="space-between"` went with it. The rail, its `region` landmark, its `Recently opened` heading and its tiles are untouched.

---

## Hydration finding

**`features.appBlocksAuthor` IS available and identical during SSR and the first client render, so the author gate does NOT need the `useIsClient()` deferral and does NOT reintroduce the bail.** Traced through the code, then verified with a real `renderToString` → `hydrateRoot`:

- **The flag.** `_app`'s `getInitialProps` calls `getFeatureFlagsAsync({ user: session?.user, host, req })` server-side (Flipt included) and serializes the result into `pageProps.flags`. `FeatureFlagsProvider` freezes it with `useState(initialFlags)`. Critically, `appBlocksAuthor` is **not** declared `toggleable: true`, so `computeUserFeatureFlagsOverlay` filters it out and the client `user.getFeatureFlags` overlay can never move it. Server value ≡ client frame-0 value.
- **The moderator floor.** `currentUser.isModerator` comes from `SessionProvider`'s `useState(initial)`, seeded from the same SSR `pageProps.session`. The one ambiguous case is a present auth cookie with an unresolved session (`session: undefined` → fetch on mount) — but there the **server also rendered without a user**, so the first client paint still matches; the session only arrives in a later render, post-hydration.
- Contrast `getNavSummary`, which really is client-only (tRPC `ssr: false`) and really does still need the `isClient` gate. That gate is unchanged.

**Verified, not asserted.** `AppsSubNav.ssrHydration.browser.test.tsx` renders the component to an HTML string under the *real* `IsClientProvider`, plants it in a container, hydrates it, and asserts React reported nothing — via **two independent detectors** (`hydrateRoot`'s `onRecoverableError`, plus a narrow `console.error` text match). Its first test is an **instrument check**: a component that deliberately renders `<p>server</p>` / `<span>client</span>` must make *both* detectors fire. It does. It also asserts the shape directly: `Create` **is in the server HTML** for an author, while `/apps/installed` and `/apps/review` are **not**.

**Known cosmetic consequence (not a hydration bug).** A non-author whose only second tab comes from the summary (e.g. one install) gets no bar on the first paint and the bar appears after mount — a pop-in, identical in kind to the conditional tabs' existing post-mount reveal. It is unavoidable without a server round-trip, which the brief rules out (and which the summary query deliberately avoids).

---

## Mutation matrix

Every guard was reverted individually and watched go red. Baseline: **52/52 green** across the three `AppsSubNav` suites; **36/36** across the recents suites.

| # | Mutation | Result | Killed on |
|---|---|---|---|
| M1 | `Create` back to `visible: () => true` | **18 failed** / 34 passed | `Create is HIDDEN for a non-author` → *expected [\<a>] to have length 0, got 1*; `NON-AUTHOR … no bar in the server HTML` → *expected … not to contain `role="tab"`*; `MODERATOR FLOOR` untouched (correctly still green) |
| M2 | delete `if (links.length < 2) return null` | **8 failed** / 44 passed | `renders NOTHING when only Marketplace qualifies` → *expected [\<a>] length 0, got 1*; `no moderator carve-out` → *[\<div>] length 0, got 1*; `NON-AUTHOR, server render → no bar at all` → *expected ['Marketplace'] to deeply equal []*; `a non-author with an empty summary renders no nav` → *[\<nav>] length 0, got 1* |
| M3 | drop the mod floor: `isAppDeveloper(user, …)` → `!!features.appBlocksAuthor` | **exactly 2 failed** / 50 passed | `🔴 MODERATOR FLOOR: a mod keeps Create even with appBlocksAuthor=false`, and its SSR twin. Nothing else moved — the floor has its own dedicated coverage |
| M4 | restore the `Jump back in` caption | **1 failed** / 35 passed | `the "Jump back in" caption is gone …` → *expected [\<p>] to have length 0, got 1* |
| M5 | delete the recents rail entirely (`entries.length >= 0 → null`) | **22 failed** | the **same** test fails, on its *positive* assertions — so "the caption is gone" cannot be satisfied by removing the rail |

### 🔴 The mutation run found a real defect in my own tests

M2 initially killed **only 2** tests — and *neither* was one of the four I wrote for the collapse. Cause: `render()` commits through a React 18 concurrent root on a later task, so a bare `expect(locator.elements()).toHaveLength(0)` immediately after `renderWithProviders` observes an **empty container and passes no matter what the component does**. Every "renders NOTHING" assertion I had written was structurally unfailable. Fixed by adopting the `RENDER_BARRIER` idiom already used in `RecentlyOpenedApps.browser.test.tsx` (render a sentinel alongside the unit under test and `await` it first). M2 then killed 8, each on its own collapse assertion. The four suites that *did* catch it unaided were the `renderToString` ones, which are synchronous.

---

## What I ran

- **Browser (`component`) project, full: 138/139 files, 1437 tests, 0 failures.** The single failing file — `ImagesAsPostsCardLazyCarousel.browser.test.tsx`, `SyntaxError: … does not provide an export named 'fetchGenerationData'` — **fails identically on `origin/main`** (measured in a clean baseline worktree), so it is pre-existing and untouched by this PR.
- **`src/components/Apps/` + `src/tests/pages/apps/`: 63 files / 758 tests green.**
- **Unit project, full: identical to `origin/main`** — 95 failed / 909 passed / 20 skipped files and 20 / 14470 / 20 tests on **both** sides, and the `FAIL` file list diffs **empty** (111 lines each). Those 95 are pre-existing local-environment collection failures (no DB / un-generated Prisma client / a gitignored `event-engine-common` tree), not regressions.
- **`pnpm typecheck` (the repo script, not bare `tsc`): 18 errors — byte-identical to the `origin/main` baseline log (`diff` empty). Zero in any file this PR touches.**
- **ESLint on every changed file: clean.** This also fixed a pre-existing `local-rules/no-wholesale-module-mock` error in `AppsSubNav.hydration.browser.test.tsx` (it now spreads `importOriginal`), which was already red on `main`.

## What I did NOT verify

- **No live/browser check against a real deploy.** Everything here is component-level; nobody drove `/apps` in a running app as a real `app-listings`-only tester. The 404-on-click symptom itself is inferred from `submit.tsx`'s `getServerSideProps`, not reproduced.
- **The civitai Vitest browser suites are report-only and do not block merge**, so "green" above is evidence, not a gate.
- **Hydration was exercised via `renderToString` → `hydrateRoot` in a test harness, not against a Next SSR response.** The provider stack in that test is `QueryClientProvider` → `MantineProvider` → `IsClientProvider`, not the full `_app` tree; a mismatch introduced by some *other* provider in the real tree would not show up here.
- **The claim that `appBlocksAuthor` is SSR-stable is read off the code** (`_app.getInitialProps` → `pageProps.flags` → `useState`, plus its absence from `toggleableFeatures`). I did not observe a production `__NEXT_DATA__` payload to confirm the flag is present in `pageProps.flags` for a live tester account.
- **No visual/pixel review.** The `AppsPageLayout.geometry` numbers still pass with the fixtures updated to an author (same two-tab bar), but nobody looked at a rendered `/apps` page for a non-author with the bar absent, nor at the recents rail without its caption.
- **`vitest.config.mts` change**: `react-dom/server` in `optimizeDeps.include` was validated by running the **full** component project (above), but only on one machine and one cold/warm cache cycle. It is additive.

## Contradictions with the brief

- The brief said `AppsSubNav.tsx:57` and `features` were the only moving parts; the sub-nav's **`AppsNavSummary` type is exported and consumed by the tests only** — `AppsSubNavView` has no non-test callers, so widening its props was cheap.
- The brief predicted the hydration gate would be the hard part. It was — but not where expected: the risk turned out to be **in the tests**, not the component (see the M2 finding above). The component-side hydration claim held on the first attempt.
- Two existing fixtures had to change (`AppsPageLayout.browser.test.tsx`, `AppsPageLayout.geometry.browser.test.tsx`): both mocked `useCurrentUser: () => null` with an empty summary, which under the new `<2` rule renders **no `<nav>` at all**, so `measure()` threw on a null lookup. They now describe an author, which reproduces the exact two-tab bar their pixel table was measured against.

### DIFF
--- a/src/components/Apps/AppsPageLayout.browser.test.tsx
+++ b/src/components/Apps/AppsPageLayout.browser.test.tsx
@@ -20,11 +20,19 @@ import type * as TrpcMod from '~/utils/trpc';
  * tabs' rule is pinned as present in the same test.
  */
 
+// 🔴 The viewer MUST be one the sub-nav renders for. `AppsSubNav` now hides itself
+// entirely below two qualifying tabs, and the summary query is stubbed empty here, so
+// an anonymous / non-author viewer would render NO `<nav>` at all and every assertion
+// below would fail on a null lookup rather than on the geometry it is about. An author
+// (`appBlocksAuthor`) yields Marketplace + Create — the same two-tab bar this file was
+// written against.
 vi.mock('~/providers/FeatureFlagsProvider', () => ({
-  useFeatureFlags: () => ({ appBlocks: true }),
+  useFeatureFlags: () => ({ appBlocks: true, appBlocksAuthor: true }),
 }));
 vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
-vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
+vi.mock('~/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => ({ id: 1, username: 'author', isModerator: false }),
+}));
 // Spread the REAL module and override only `trpc` (local-rules/no-wholesale-
 // module-mock): a hand-written replacement silently breaks every importer the day
 // '~/utils/trpc' grows an export this factory omits — the whole FILE then fails to
--- a/src/components/Apps/AppsPageLayout.geometry.browser.test.tsx
+++ b/src/components/Apps/AppsPageLayout.geometry.browser.test.tsx
@@ -42,11 +42,19 @@ import { page } from 'vitest/browser';
 import { renderWithProviders } from '../../../test/component-setup';
 import type * as TrpcMod from '~/utils/trpc';
 
+// 🔴 The viewer MUST be one the sub-nav renders for. `AppsSubNav` now hides itself
+// entirely below two qualifying tabs, and the summary query is stubbed empty here, so
+// an anonymous / non-author viewer would render NO `<nav>` at all and `measure()` would
+// throw on a null lookup instead of measuring. An author (`appBlocksAuthor`) yields
+// Marketplace + Create — the same two-tab bar every number in the table above was
+// measured against, so the pixels are unchanged.
 vi.mock('~/providers/FeatureFlagsProvider', () => ({
-  useFeatureFlags: () => ({ appBlocks: true }),
+  useFeatureFlags: () => ({ appBlocks: true, appBlocksAuthor: true }),
 }));
 vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
-vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
+vi.mock('~/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => ({ id: 1, username: 'author', isModerator: false }),
+}));
 // Spread the REAL module and override only `trpc` (local-rules/no-wholesale-
 // module-mock) — see the sibling browser test for why.
 vi.mock('~/utils/trpc', async (importOriginal) => ({
--- a/src/components/Apps/AppsSubNav.browser.test.tsx
+++ b/src/components/Apps/AppsSubNav.browser.test.tsx
@@ -4,6 +4,7 @@ import {
   activeAppsTab,
   AppsSubNavView,
   isActiveAppsRoute,
+  type AppsNavContext,
   type AppsNavSummary,
 } from '~/components/Apps/AppsSubNav';
 // `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
@@ -13,6 +14,13 @@ import { renderWithProviders } from '../../../test/component-setup';
 // `AppsSubNavView` (props-only) so it renders in isolation without the tRPC
 // query / router that the `AppsSubNav` container wires. We drive the booleans
 // directly and assert which tabs render.
+//
+// TWO input objects, on purpose (they have different provenance and different
+// hydration behaviour — see the type docs):
+//   - `summary`  — the `blocks.getNavSummary` booleans (client-only, revealed
+//                  post-mount).
+//   - `context`  — viewer CAPABILITIES resolved from SSR-seeded values
+//                  (`isAuthor`), safe on the first paint.
 //
 // The sub-nav uses the Mantine **Tabs** LOOK but is wrapped in a real
 // `<nav aria-label="App sections">` so it's exposed as a navigation LANDMARK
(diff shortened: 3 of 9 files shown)


ZacxDev:
## Correction to this PR's description (post-audit)

An adversarial pre-merge audit found two statements in the description above that overstate what this change does. Correcting here rather than editing the body, since reviewers may have already read the original.

### 1. "The two gates cannot drift" — overstated

The nav shows `Create` on `isModerator || appBlocksAuthor` (`AppsSubNav.tsx:294-296`), while `/apps/submit` returns `notFound` on `!features?.appBlocksAuthor` **alone**, before and independently of `isAppDeveloper` (`src/pages/apps/submit.tsx:41`). Flipt overrides role checks in both directions, so the moderator floor is itself a drift vector: a moderator outside the `app-blocks-author` segment would get the tab shown and the page 404 — the exact symptom this PR exists to remove.

**Not reachable today**, measured against live Flipt rather than assumed:

```
moderator (uid 8753561, isModerator=true) -> app-blocks-author = true, segmentKeys=["moderators"]
```

The `moderators` segment is on that flag's rollout, so the floor and the flag currently agree. This becomes real only if that segment is ever narrowed. The guard is still correct — the claim is what needs softening, from "cannot drift" to "do not drift under current flag configuration".

Corollary noted by the audit: `submit.tsx:50`'s `isAppDeveloper` call is effectively dead — line 41 has already forced `appBlocksAuthor` true, so the predicate cannot return false.

### 2. The new layout shift is not "identical in kind to the existing post-mount reveal"

For the cohort this PR targets, the sub-nav now renders **nothing** at first paint and appears after hydration. Previously the bar was present at first paint (Marketplace + Create) and only ever grew *horizontally* as conditional tabs resolved — nothing moved vertically.

This PR's own tests show it: `NON-AUTHOR, first client paint (query data PRESENT)` yields `renderedTabs() === []`, against a post-mount positive control returning 5 tabs. The result is a vertical content shift (tab row + `Stack gap="md"`) on every `/apps/*` load for non-authors. That is a new user-visible behaviour and deserves an explicit decision — reserve the row height, or accept it — rather than being carried as equivalent to the prior reveal.

### Not a defect: the "no-op for testers" concern

The audit correctly noted that `AppsSubNav.tsx:273` gates on `features.appBlocks` alone, while `/apps` gates on `appListings || appBlocks` (`resolveAppsPageAccess.ts:36`), and flagged that this PR could be a no-op if the widened tester cohort held `app-listings` without `app-blocks`. It could not check the cohort.

Measured against live Flipt:

```
tester (uid 11072787, isModerator=false)
  app-listings       = true
  app-blocks-enabled = true
  app-blocks-author  = false
```

The cohort holds both, so the sub-nav renders and the gate is exercised for exactly the intended users. **Not a no-op.**

The underlying inconsistency is still worth a follow-up: if a future widen ever sets `app-listings` without `app-blocks`, the sub-nav disappears entirely for that cohort while the store pages still render. Aligning `AppsSubNav.tsx:273` with `resolveAppsPageAccess` would close that.

### Other audit findings (non-blocking, not addressed here)

- `AppsSubNav.ssrHydration.browser.test.tsx` leaks React roots across tests (`hydrateInto` never unmounts; `onRecoverableError` closes over a rebound module-level binding) — a low-probability cross-test flake. `root.unmount()` in `afterEach` closes it.
- Three of that file's five "hydration is clean" cases compare an empty server render against an empty client render, so React diffs nothing. They are not five independent signals; the real observations are the AUTHOR and MODERATOR cases.
- `prettier --check` flags `AppsSubNav.hydration.browser.test.tsx:338` (101 chars), which was clean on `main`, yet the "ESLint + Prettier (changed files)" job reports pass. Worth investigating that job's changed-file selection independently of this PR.

Verdict from the audit was **safe to merge**, with no deploy-blocking findings. Every guard survived a 9-mutant semantic sweep built independently of the one in the description (inverted predicates, swapped operands, threshold weakening and strengthening — no deletion mutants).

PR #3900, opened 2026-08-13
test: cap vitest worker pool on high-core machines (node + browser)

JustMaier:
## Problem

Vitest defaults `maxWorkers` to `cpus - 1`. On a 32-core dev box that is 31 forks for a single suite, and two concurrent suites saturate the machine — measured at **67 node processes and 10.9 GB RSS, 100% CPU**. Nothing is leaking or hanging; this is the default sizing doing exactly what it is configured to do. With 15-20 agents running, two-plus concurrent suites is the normal case, not the edge.

The collateral damage landed on things that were not running tests: local speech-to-text fell from 30-40x realtime to 4-9x **with the GPU idle** (host-side starvation, not GPU contention), and DB writes went from 275 ms to 1.6 s.

## Change

Four lines. `test.maxWorkers` at the root of `vitest.config.mts`, capped at 8 but floored against the machine's own CPU-derived default.

## Measurements

`pnpm run test:unit:run`, 1002 files / 15,762 tests, this 32-core box:

| workers | wall clock | vs default | import (worker-s) | tests (worker-s) |
|---|---|---|---|---|
| **31** (current default) | **235.9s** | — | 4991.4 | 596.1 |
| **8** (this PR) | **410.2s** | +174s, 1.74x | 2266.0 | 310.9 |
| 6 | 551.6s | +316s, 2.34x | 2318.5 | 273.5 |
| 31 (repeat, drift control) | 236.6s | +0.7s on the first | 5001.1 | 644.6 |

The drift control landed **0.3%** off the original, so the spread is signal rather than noise.

The `import` column is the interesting one. At 31 workers the suite performs **4991 worker-seconds** of importing against **2266** at 8 — more than twice the total work for 1.74x the wall clock. Extra forks past a point buy memory pressure and cache thrash, not throughput (26.5 effective workers out of 31 requested, versus 7.0 out of 8).

So the trade is **+2.9 minutes on a single suite run**, for a machine that stays usable and a suite that does half as much total work. Two concurrent capped suites should finish sooner than two uncapped ones, since they stop fighting each other.

## Why `Math.min` and not a flat `8`

An explicit `maxWorkers` **overrides** the CPU-derived default, and CI passes no worker flag — `.github/workflows/lint.yml` runs bare `pnpm run test:unit:run`. A flat `8` would *raise* a 4-vCPU runner from its own 3-worker default to 8. `Math.max(1, Math.min(8, cpus - 1))` leaves any box with <= 9 cores byte-identical and caps only fat machines. **CI behaviour is unchanged.**

## Why root-level and not per-project

Vitest resolves `project.config.maxWorkers` -> root `maxWorkers` -> CPU default, so a root value also reaches the `packages/*` and `apps/*` projects, which are separate config files that set nothing. Giving one project its own number instead requires a unique `sequence.groupOrder`, or `groupSpecs` throws outright.

## One correction worth recording

The original ask was for `poolOptions.forks.maxForks`. That option **does not exist in this repo's vitest** — `test.poolOptions` was removed in Vitest 4 and is silently ignored apart from a deprecation log. Setting it would have been a no-op that looked like a fix.

## Verification

- `pnpm run typecheck` — **OK, 0 type errors** in 176s
- `pnpm run lint` — exits 1, but **identically (412 error lines) on clean `main`**; this file is not flagged
- `pnpm exec prettier --write vitest.config.mts`
- `pnpm run test:unit:run` — **16 failures at every worker count tested (31, 8, 6, 31)**, i.e. worker-independent and pre-existing on this Windows box (one is a plain `\` vs `/` path assertion). Not introduced here.
- `blocks.router.workflow.test.ts` collected **347 tests, 0 failed** — submodule is initialised, so the runs above mean something.

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -4,17 +4,57 @@ import os from 'os';
 import path from 'path';
 
 // Vitest defaults to `cpus - 1` workers, so two concurrent suites saturate a 32-core dev box.
-// Measured here: 31 workers 235.9s, 8 workers 410.2s — 1.74x slower for a machine that stays usable.
+// Measured here on the unit suite: 31 workers 235.9s, 8 workers 410.2s — 1.74x slower, for a
+// machine that stays usable while several agents run suites at once.
 //
-// `Math.min`, not a flat 8: an explicit `maxWorkers` OVERRIDES the CPU default, and CI passes no
-// worker flag, so a flat 8 would RAISE a 4-vCPU runner from 3 workers to 8.
+// The cap must only ever REDUCE. Vitest's default depends on the resolved mode (`cpus - 1` in run,
+// `floor(cpus / 2)` in watch) and that mode is NOT knowable from this file: `vitest run --watch` is
+// explicitly supported and resolves to watch while `process.argv` still says "run", so sniffing argv
+// raises the count on exactly the small machines this is meant to protect.
 //
-// Per-project overrides need a unique `sequence.groupOrder` or `groupSpecs` throws. And note
-// `test.poolOptions` was removed in Vitest 4 — `poolOptions.forks.maxForks` is silently ignored.
-// `VITEST_MAX_WORKERS` overrides this per run.
+// So don't try to know the mode. Take the watch-mode default as the ceiling — it is the smaller of
+// the two at every core count — and apply nothing at all below 10 cores. That is mode-blind and
+// provably never exceeds either default, while leaving small boxes and CI (4 vCPU) untouched.
+//
+// `VITEST_MAX_WORKERS` still overrides everything below, per run.
+// (Unrelated but adjacent: `test.poolOptions` was removed in Vitest 4, so `poolOptions.forks.maxForks`
+// does nothing apart from logging a deprecation.)
 const WORKER_CAP = 8;
+// Below this the run-mode default (`cpus - 1`) is already <= WORKER_CAP, so there is nothing to cap.
+const CAP_ABOVE_CORES = WORKER_CAP + 1;
 const cpuCount = os.availableParallelism?.() ?? os.cpus().length;
-const maxWorkers = Math.max(1, Math.min(WORKER_CAP, cpuCount - 1));
+const capped = (ceiling: number) => Math.max(1, Math.min(WORKER_CAP, Math.floor(ceiling / 2)));
+
+// `undefined` leaves Vitest's own resolution completely alone.
+const maxWorkers = cpuCount > CAP_ABOVE_CORES ? capped(cpuCount) : undefined;
+
+// The browser pool sizes itself separately — `min(12, cpus - 1)`, because its main thread chokes
+// past ~12 — and `getThreadsCount` reads ONLY the project's own `maxWorkers`, never the root. So
+// the browser project has to be capped individually or it is not capped at all. Scope, stated
+// honestly: this equals the browser watch-mode default at every core count, so it bites only in
+// run mode (12 -> 6 at >= 13 cores) and is a deliberate no-op in watch.
+const componentMaxWorkers =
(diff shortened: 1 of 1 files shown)


JustMaier:
## Update — three defects found and fixed by adversarial review

Four review passes. Each of the first three found a real bug, and **two of them were bugs introduced by the fix for the previous one**. Recording that plainly because the pattern is the point: the original four-line change was fine in isolation and wrong in three ways once anyone looked at the surrounding machinery.

### 1. Watch mode went the wrong way
`min(8, cpus - 1)` is shaped for the **run**-mode default. Vitest uses `floor(cpus / 2)` in watch, so on any **3-15 core machine** `pnpm run test:unit` spawned *more* workers than before — an 8-core laptop went 4 → 7. It raised the count on exactly the machines with the least headroom.

The first fix for this sniffed `process.argv` for `run`. That was also wrong: **`vitest run --watch` is explicitly supported** (vitest's own source carries the comment `// "vitest run --watch" should still be watch mode`), so argv says `run` while vitest resolves *watch*.

The mode is not knowable from a config file, so the cap no longer tries. It caps against the **watch** default — the smaller of the two at every core count — and sets **nothing at all below 10 cores**, where the run default is already ≤ the cap. Mode-blind, and CI is untouched by construction rather than by arithmetic that happens to coincide.

Verified by sweeping **cores 1-128, both modes, both pools**: **0** counts where this exceeds vitest's own default, **230** where it binds. Positive control: the old formula scores **13 violations** on the same check, so the check is capable of failing.

### 2. The cap never reached the browser pool
`getThreadsCount` consults only a project's **own** `maxWorkers`, never the root. So `component` — the suite that puts a real Chromium behind every worker, i.e. the one most able to make a machine unusable — was the single thing the cap missed. It now caps against the browser pool's own default: **12 → 6** in run mode on a 32-core box.

**That number is not measured.** The 235.9s/410.2s figures are the node suite. I believe 6 is right because browser workers are the expensive ones, but it is a judgement call and should read as one.

### 3. A `--sequence.*` flag silently killed the run
Making the two projects resolve to different counts means `component` needs its own `sequence.groupOrder`, or `groupSpecs` throws. Declaring it statically isn't enough — a CLI `--sequence.shuffle` **replaces `test.sequence` wholesale** (the override is a spread, not a merge) and takes `groupOrder` with it. The resulting failure reports as **`Test Files: no tests`**: a run that reads as having found nothing rather than as having failed.

Now also re-asserted from a `configureVitest` hook that runs after resolution. Both halves are kept on purpose — the static one covers the plugin going inert (this file is outside `tsconfig`'s `include`, so **nothing typechecks the hook name**; a typo would be silent), and the plugin covers the CLI.

**Known gap, stated rather than hidden:** `--sequence.groupOrder=1` moves the *other* projects onto component's value and still throws. No constant can dodge that. Nothing in `package.json` or the workflows passes it.

### Verification

| check | result |
|---|---|
| `pnpm run typecheck` | OK, 0 errors |
| `pnpm run test:unit:run` | 16 failures — **identical to baseline**, unchanged |
| `blocks.router.workflow.test.ts` | **347 collected**, submodule live |
| both projects in one invocation | 350 passed, no throw |
| `--sequence.shuffle` | passes; **negative control reproduces the throw** without the plugin |
| never-increase sweep | 0 violations / control catches 13 |
| `pnpm run lint` | exits 1, **identically (412 lines) on clean `main`**; this file not flagged |

Two pre-existing issues surfaced along the way, neither introduced here: 16 Windows-specific unit failures (one a plain `\` vs `/` path assertion), identical at every worker count tested; and an order-dependent test in `blocks.router.workflow.test.ts` that fails under `--sequence.shuffle` **identically on clean `main` with the same seed**.

Retitled from `perf(test)` — a change that makes a single suite 1.74x slower shouldn't claim to be a perf win. The win is machine usability under concurrent suites.

PR #3901, opened 2026-08-13
feat(feedback): in-product feedback capture, first used by the BitDex feed notice

JustMaier:
## What

A generic place to collect in-product feedback, and its first consumer: when BitDex is actually serving your image feed, a short notice appears at the top of `/images` and `/videos` with a box to tell us what looks wrong.

From Justin, in standup: *"if Bitdex is in primary for you, the top of feeds where Bitdex is being used has a little notice that says, we're currently testing out a new system, let us know if you see any weirdness with your feeds... captures essentially their input so that then we can audit at any point."* The goal is to stop the feedback loop depending on people contacting support.

## The load-bearing decision: how "BitDex is serving THIS feed" is decided

**The notice reads the `source` the server emits from the branch that actually returned the data. It never re-reads the flag.** Those disagree in three real ways, and each one was found by review rather than by design:

| Situation | Flag says | Who actually served it |
|---|---|---|
| `shadow` | on | Meilisearch (BitDex runs detached, results discarded) |
| `primary`, BitDex errors or returns nothing | primary | Meilisearch, **per page** |
| flag switched off mid-session | off | Postgres, **from the next page on** |

So both feed branches now name themselves — the index path already emitted `source`, and `getInfiniteImagesHandler` stamps `source: 'db'` on the DB branch. The gate reads the **last** page's backend, skipping only genuinely sourceless pages (an index page that returned nothing, i.e. the end of a feed). Shadow can never show the notice by construction, not by a separate check.

The stamp is in the controller, not in `getAllImages`: `GetAllImagesIndexResult = AsyncReturnType<typeof getAllImages>`, so a required field on the service return propagates to two index functions that don't set it.

## Storage

New `Feedback` table. `area` is a plain `text` column with a TypeScript registry, **not a Prisma enum** — adding a feedback surface costs a constant, not a migration. Each area is gated by its own `feedback-area-<slug>` Flipt flag, so an area can be switched off without a deploy.

Submissions are users-only (`guardedProcedure`, so muted and non-onboarded users don't see a prompt they can't submit), capped at 5/hour/user via an atomic Redis `INCR` with a writer-side fallback, and refunded if the insert fails. The stored `context` — route, filters, per-page backends — is a **claim about what the viewer saw**, never read back as proof.

## ⚠️ Needs a human, in this order

1. **Apply the migration** — `packages/civitai-db-schema/prisma/migrations/20260813180000_feedback/migration.sql`, by hand, to prod nvme0 **and** the dev clone. Idempotent, one additive table, no backfill.
2. **Then** create and enable the Flipt flag `feedback-area-bitdex-image-feed`. It defaults off, so this ships inert: nothing renders and nothing is written until the flag is on. Enabling it before step 1 would render the prompt and fail every submit.

Reading the results is plain SQL — no endpoint, deliberately:
```sql
select "createdAt", "userId", message, context
from "Feedback" where area = 'bitdex-image-feed' and status = 'new'
order by "createdAt" desc;
```

## Review

Four adversarial passes, prompted to refute. It found, and this fixes: a rate limit that wasn't one (check-then-act against a read replica), the notice leaking into the collection-picker and challenge-submit modals, a failed write burning a slot for an hour, and **two opposite wrong answers** on the sourceless-page question before the server-side fix made the cases distinguishable. Every fix is mutation-tested — reverting one produces a named assertion, not a hang.

**Known gaps, stated rather than papered over:** the `!filters.hidden` exclusion and the snapshot's ref-identity check have no tests; both need a `renderHook`/component harness. The helpers underneath them are covered.

## Verification

`pnpm run typecheck` 0 errors · `pnpm run lint` clean on every touched file (the 5 `no-unused-vars` in `image.controller.ts` are pre-existing — same count on clean `main`) · `prettier` clean · `pnpm run test:unit:run` 34 failures, **byte-identical set to clean `main`**, and `blocks.router.workflow.test.ts` collected **347** tests, so the run counted.


JustMaier:
## Update since review — simplification, a real flag bug, and the UI

**The area flag could never be enabled locally.** `isFeedbackAreaEnabled` used `getFliptBoolean`, which is the variant that *deliberately* ignores `FLIPT_LOCAL_OVERRIDES` so a call site can read real rollout state. A feature gate wants the override. Now `isFlipt` (`isEnabled`). Found by Justin when the notice didn't render on a dev server that had the override set.

**The rate limiter is now the shared one.** It was a hand-rolled Redis counter shaped after `block-tip-rate-limit.ts`; three review rounds hardened that copy instead of asking whether it should exist. It's now `rateLimit` from `middleware.trpc.ts` — sliding window rather than fixed, one atomic EVAL, logged fail-open, moderators and non-prod skipped, and `onlyCountSuccess` is the refund-on-failed-write behaviour for free.

Deleted with it: `withinRateLimit`, `refundRateLimit`, the `reserved` threading, the `dbWrite` fallback, the `REDIS_SYS_KEYS.FEEDBACK` entry, and 11 tests that existed only to hold up the bespoke version.

**`reportedPageSources` is an array again.** `context` is JSONB; the run-length encoding only existed because the list had been put in a string field and capped, which then needed a scheme to survive a deep scroll, which then needed tests.

Net: **137 insertions, 343 deletions** against the reviewed state.

**UI**, per Justin's direction: card with a warning triangle, a "Give feedback" pill, form under a full-width divider with Send right-aligned and pinned to the top of the row, X cancels the form before dismissing the notice, crypto-deposit card surface. Now enabled on home, tools and profile media as well as `/images` and `/videos` — every surface where BitDex serves a browsing feed. Still off in the collection-picker and challenge-submit modals.

## Applied and exercised end to end

Migration applied to **prod and the dev clone**, both verified: 7 columns, 3 indexes, 0 invalid, `Feedback_status_check`, FK with `ON UPDATE CASCADE`. Justin submitted through the UI and the row landed with its context intact:

```
area bitdex-image-feed · status new · context: { path: /images, reportedSource: bitdex,
  reportedPageSources: [bitdex], pagesLoaded: 1, filters: {...} }
```

Flipt flag `feedback-area-bitdex-image-feed` exists, boolean, **disabled**, with the ordering constraint in its description.

Also folded in a `postgres-query/SKILL.md` correction: it claimed the dev role is a superuser. It isn't — DDL there fails via a `ddl_command_end` trigger that reassigns ownership, and `CREATE` on the schema doesn't imply DDL works.

## Verification

`typecheck` 0 errors · eslint clean on every touched file · prettier clean · unit suite **34 failures, byte-identical set to clean `main`**, `blocks.router.workflow.test.ts` collected **347**.

PR #3904, opened 2026-08-13
fix(api): close the five TIER-1 unauthenticated disclosure routes + the 17 no-cause sites (civitai#3845)

ZacxDev:
> ### 🔴 STACKED PR — base is `zach/3845b-rest-envelope-consolidation` (#3881), **not `main`**
> This depends on the error-envelope ledger #3881 introduces
> (`src/server/utils/__tests__/rest-error-envelope-ledger.test.ts`) and edits it.
> **Merge #3881 first.** When you do, merge it **without `--delete-branch`** — GitHub
> auto-closes any PR whose base branch is deleted and refuses to reopen it, which
> would lose this review thread. Retarget this PR to `main` first if you want the
> parent branch gone.

Closes the **TIER 1** block that #3881 enumerated in `KNOWN_UNFIXED_SAME_CLASS` and
deliberately left open: five routes, all reachable **without authentication**, each
serving a caught error's text to anonymous callers. A `throwDbError`-wrapped driver
error puts a raw ``Invalid `prisma.<model>.<method>()` invocation`` — table, column
and constraint names — on the wire, and for a `pg` 23505 an actual user row value
(`Key (email)=(victim@example.com) already exists.`). Same class as the incident
#3845 fixed.

Six commits: the five routes, the 17 `no-cause` `TRPCError` sites, the `/api/v1/images`
503 review fix, the blind-audit follow-ups, the delta-audit fixes to the guards those
follow-ups introduced, and a final fix to a silent-drop regression one of THOSE
introduced.

---

## Blind-audit outcome

An independent adversarial pass was run on this PR **without** access to my reasoning.
**Result: no deploy-blockers.** Two things it produced that this PR now carries:

**It settled my one stated unknown, empirically.** I had recorded the
`tensor-metadata` pre-`try` path as unverified and explicitly declined to establish its
safety myself. The auditor drove Next 16.3.0's own `apiResolver` with this repo's
production config (`dev: false`, `propagateError: false` hardcoded in `pages-api.js`, no
custom `onRequestError`) and measured that a throw escaping the handler yields
`500 "Internal Server Error"` with the driver text **absent**; `withAxiom` re-throws
without responding. **Non-leak, measured.** That supersedes the unknown.

### Fixed after the blind audit

1. **A mutation survivor that defends this PR's own thesis.**
   `generation.service.ts:528` — reverting `if (type === 'audio') throw
   throwBadRequestError('not implemented')` to `throw new Error(…)` left the **entire
   15.4k-test suite green**. The sibling arm at `:298` was pinned; this one was not, so
   half the "converted so a client rejection is not genericized into a 500" claim had no
   coverage. Now pinned in `generation.service.client-rejections.test.ts`, driven the way
   production reaches it (`?type=image&id=…` at a media row whose own type is audio), and
   mutation-verified: the revert applies at `:528` and **that new test is what goes red**.
2. **`/api/v1/blocks/images.ts` — two comment blocks asserted a parity that no longer
   holds.** `:223` claimed it "mirror[s] the public endpoint's error mapping" and
   `:253` that the 503 contract is "identical". After the 503 fix the public route
   answers 503 in place and delegates everything else, while that route still serves
   `{ error: trpcError.message, code }`. **Behaviour unchanged — it is token-gated,
   correctly out of scope, and recorded in the ledger baseline as TIER 2.** Only the
   claims are fixed, and they now enumerate exactly which arms still mirror (499,
   raw-transient-503) and which have diverged, plus a note that closing it means deleting
   its ledger entry, not copying the mapping across. A false parity claim is precisely
   what convinces the next maintainer to "restore" it.
3. **`tensor-metadata` — a trivially-reachable anonymous 500 with NO fault record.**
   `z.preprocess(v => Number(v), z.number())` accepts `?id=1e30` and `?id=1.5`; the value
   then binds to an int4 column and Postgres throws from `dbRead.modelFile.findUnique`,
   which sits **before** the handler's `try` — so no `logToAxiom`, no `_axiom` fault.
   Not a leak (measured above), but an anonymous caller could mint fault-record-free
   500s, which is the forensic guarantee the rest of this work rests on. Bounded to
   `.int().gt(0).lte(2147483647)` — the repo's existing fix for this exact class
   (`id-overflow-validation.test.ts`) — so the bad input takes the handler's existing 400
   arm with caller feedback.
   🔴 **I did NOT move the two pre-`try` calls inside the try**, and the constraint is
   why: the existing catch closes over `file` and `format` for its `logToAxiom` call, so
   any shape that brings those calls inside — including giving them their own
   `handleEndpointError` try — requires hoisting both to `let` with hand-written types
   and defeating TS narrowing across the boundary. That is real bug surface on a hot
   public route for a log line on a failure mode that is already a non-leaking hard 500.
   The trivially-reachable half is closed deterministically at the schema instead. The
   asymmetry the auditor noted (`download/models` runs its identical preprocess *inside*
   its try) is recorded in the file rather than silently equalised.
4. **The ledger had an undisclosed blind spot, and it was worth fixing rather than
   disclosing.** `TRPC_CTOR`'s `[^{}]*?` body cannot cross a nested brace, so **any
   `new TRPCError({…})` whose body contains a nested `{}` or a `${}` template matched
   nothing at all** and was silently exempt. Measured across `src/server`: the regex saw
   **401** sites, brace matching sees **485** — a **17% hole** in a guard whose whole job
   is exhaustiveness. Replaced with a brace matcher. **None of the 84 newly-visible sites
   is a bypass**, so the baseline stays at zero — closing it cost no churn, and leaving it
   would have meant a future `new TRPCError({ code, message: err.message, meta: {…} })`
   was exempt by punctuation. Two exemplar controls pin it, and reverting to the old
   regex turns them red.

**Also hardened (cheap, and the behavioural LEAK tests catch the real thing regardless):**
the two measured evasions of the `errorResponse` source guard — a **variable status**
defeated its `\d{3}` status group so the call site was never extracted (vacuously
"safe"), and `'Error: ' + err.message + ''` satisfied its first/last-character anchoring
— plus a single-line `/* */` false positive. All three now have controls, and reverting
either fix turns them red.

---

## Why this was not a delegation

#3881's own note says it: **each of these five answers a 4xx for EVERY failure,
including a legitimate zod-parse rejection.** Routing them blindly through
`handleEndpointError` turns a client's malformed `?ids=` from a **400 into a 500** —
fixing a disclosure by breaking the API contract. So each route separates *"the
caller sent something invalid"* from *"something failed server-side"*, and the split
is made **structurally** (lift the validation out of the `catch`) rather than with an
`instanceof` inside the catch, which a zod error re-thrown from deeper in a service
would defeat.

| route | wrapper | decision |
|---|---|---|
| `v1/images/index.ts` | `PublicEndpoint` | **Delegate.** The zod `safeParse` 400 and the paging 429 already `return` from the try body and never enter the catch, so nothing there is client feedback. The 503 `no-store`+`Retry-After` arm is decided *before* delegating (the helper can't know this route wants a `Retry-After`). |
| `generation/data.ts` | `PublicEndpoint` | **`parse()` → `safeParse()`**, 400 returned before the try's failure path; everything past the parse delegates. `result.error.message` is the identical string `parse()` threw, so the 400 body is byte-identical. |
| `generation/resources.ts` | `PublicEndpoint` | Same. `getResourceData` has no client-rejection path of its own (returns `[]` for an empty id list), so no 4xx is lost. |
| `download/models/[modelVersionId].ts` | `PublicEndpoint` | **Genericize in place, do NOT delegate.** `errorResponse` does content negotiation the helper cannot: a browser UA got the same driver prose as bare `text/plain` via `res.send`. Delegating would force JSON on that arm. Shares `GENERIC_SERVER_ERROR_MESSAGE` so the two cannot drift. |
| `v1/model-files/[id]/tensor-metadata.ts` | `MixedAuthEndpoint` (public read) | **Keep the 422**, replace only the message. 422 here is a statement about the requested *file* — caller feedback — and nothing in that try throws a `TRPCError`, so delegating would reclassify it as a 500. |

### The trap the brief did not anticipate

`getGenerationData`'s `default:` arm is **reachable**: `getGenerationDataSchema`
accepts `type: 'audio'` and the switch has no arm for it, so
`GET /api/generation/data?type=audio&id=1` lands on `throw new Error('unsupported
generation data type')`. That reached the wire as a **400 with that text**, purely by
accident of the catch-all. Once the route delegates, a plain `Error` is classified as
a server fault and collapses into a generic 500 — a silent downgrade of real client
feedback. Two such throws in `generation.service.ts` are converted to
`throwBadRequestError`, which preserves **both** the 400 and the exact text
(`handleEndpointError` passes a non-JSON 4xx message through as `{ message }`,
byte-identical). So the split is *not* simply "zod vs everything else".

## Consumer compatibility

- `message` is a **string** at every status on every route (Go CLI decodes it into
  `Message string`; a non-string fails the whole envelope unmarshal). Asserted.
- Where `error` is emitted it stays a **non-empty string** (the CLI renders `error` in
  preference to `message`). Asserted. `tensor-metadata`'s 422 uses a string literal
  rather than an imported constant precisely because `ModelTensorMetadata.tsx` renders
  `body.error` verbatim as the panel's message.
- The 400 zod `.flatten()` shape is untouched — none of these five emitted it, and the
  two that emit `z.prettifyError(...)` are unchanged.

### 🔴 Consumer-visible changes — please review these specifically

> **Revised after review.** An earlier version of this section called item 1 "the
> highest-risk item in the PR" and listed a fourth item, a 503 `Cache-Control` change.
> Both were wrong. I enumerated every non-2xx exit on the committed route, which I
> should have done before assigning a risk label; and the 503 behaviour that produced
> the `Cache-Control` change was itself a defect, now fixed rather than documented.

**Blast radius of the `/api/v1/images` body change, measured rather than asserted.**
Four of the five non-2xx exits `return` *before* the catch and are **byte-identical**
to today:

| exit | body | changed? |
|---|---|---|
| zod rejection (`:118`) | `{ error: <ZodError> }` @400 | **no** |
| paging cap (`:130`) | `{ error: "You've requested too many pages…" }` @429 | **no** |
| bulkhead shed (`:153`) | `{ error: 'Server busy, please retry shortly.' }` @503 | **no** |
| transient search (`:230`) | `{ error: 'Image search is temporarily overloaded — please retry.' }` @503 | **no** |
| catch → `handleEndpointError` (`:258`) | `{ error, code }` → `{ message }` at 4xx | **yes** |

So the real blast radius is *"a `TRPCError` escaping `runImageSearch` /
`getServerAuthSession` that maps to a **4xx**"* — the two 4xx a normal client actually
provokes never enter that catch, and **503 no longer reaches it at all** (see item 3).

1. **`/api/v1/images` non-2xx body at 4xx: `{ error, code }` → `{ message }`.** The
   shared 4xx shape, the same trade `/api/v1/users` made in #3881, and the ledger
   structurally forbids keeping the hand-rolled shape (`{ error: e.message, code }` is a
   matched leak shape). The full `{ error, message, code }` envelope is still emitted on
   a genericized 5xx, so `error` and `code` survive there. **Failure mode if a consumer
   depends on it is SILENT**: `body.error` reads `undefined` (rendered as an empty or
   "undefined" message, not an exception), and a `body.code` branch falls to its
   default. If you'd rather not take it, the alternative is to teach
   `handleEndpointError`'s 4xx arm to emit the full additive envelope — correct, but it
   changes 14 other routes and belongs in its own PR.
2. **`/api/generation/data`**: a `NOT_FOUND`/`UNAUTHORIZED` from the service now answers
   404/401 instead of 400. Body shape is unchanged (`{ message }` either way).
3. **`/api/v1/images` 503: NO change — this is a fix to an earlier draft of this PR.**
   That draft set the retry headers for the service-wrapped `TRPCError` case and then
   fell through to `handleEndpointError`, which answers a 503 as `{ message }`. The
   result was **the same route emitting the retry hint under `error` on the raw-SDK
   path and under `message` on the delegated path**, so no client could read one key to
   get it at 503 — strictly worse than either choice alone, and the reason the
   `Cache-Control` drifted to `no-store, max-age=0`. Both transient shapes are now
   answered **in place** with one body, exactly as `/api/v1/users` does. The two 503
   paths are byte-identical, `Cache-Control` stays bare `no-store`, and 503 never
   reaches the helper. This also closes a residual: `handleEndpointError` deliberately
   does **not** genericize 503 (it is the only copy of a retry hint), so a
   `SERVICE_UNAVAILABLE` whose message happened to be driver-authored could previously
   have reached the wire through that path. A literal cannot.
   **This was found by review, not by the suite** — the existing test asserted status
   and both headers and never the body. It now asserts the body by value, and that
   assertion is mutation-verified (see below).
4. **Duck-typed `TRPCError`s at `/api/v1/images`.** The old code read the status off
   `getHTTPStatusCodeFromError(error as TRPCError)`, which only reads `.code`;
   `handleEndpointError` uses `instanceof`. A repo-wide sweep found **no production
   site** that fabricates a duck-typed `TRPCError` — only the test file did, and those
   fixtures are now real instances. Flagging it because it is a real narrowing, not
   because I found a caller it breaks.
5. **Non-breaking improvement, noted for completeness:** a 4xx/5xx that now delegates
   picks up the helper's `no-store, max-age=0`, where previously `PublicEndpoint`'s
   public cache headers could let an error response be edge-cached.

## Red → green matrix

Every `LEAK:` test below was **observed failing with an assertion failure** (not a
compile error) against `origin/zach/3845b-rest-envelope-consolidation`, with the fix
reverted per file and the test files kept.

**Commit 1 — 21 red at base, all green at HEAD:**

| test | red at base | green at HEAD |
|---|---|---|
| `tier1-public-route-disclosure.test.ts` · `LEAK: Prisma driver/table/column` × 5 routes | ✅ 5 | ✅ |
| `tier1…` · `LEAK: raw (unwrapped) Prisma error` × 5 routes | ✅ 5 | ✅ |
| `tier1…` · `LEAK: pg unique-violation detail` × 5 routes | ✅ 5 | ✅ |
| `tier1…` · `LEAK: download/models text/plain BROWSER arm` | ✅ 1 | ✅ |
| `tier1…` · `generation/data: NOT_FOUND now answers 404` | ✅ 1 | ✅ |
| `tier1…` · `every errorResponse call site passes text we wrote` | ✅ 1 | ✅ |
| `rest-error-envelope-ledger` · `offender set is EXACTLY the baseline` (57 → 53) | ✅ 1 | ✅ |
| `rest-error-envelope-ledger` · `none of the routes THIS change fixed are offenders` | ✅ 1 | ✅ |
| `generation.service.client-rejections` · `unsupported type raises BAD_REQUEST` | ✅ 1 (`expected Error … to be an instance of TRPCError`) | ✅ |

**Commit 2 — 1 red at base:**

| test | red at base | green at HEAD |
|---|---|---|
| `rest-error-envelope-ledger` · `no TRPCError forwards `.message` at a 4xx without `cause`` | ✅ (4 files / 17 sites vs `[]`) | ✅ |

**Labelled `INVARIANT:` — green on BOTH sides, NOT counted as regression coverage.**
These are the fence the fix had to stay inside, not proof it works: `an invalid request
is still rejected as a CLIENT error, not a 500` (×5), `the failure still ANSWERS with a
string message the CLI can decode` (×5), `tensor-metadata KEEPS its 422`, `download/models
still logs the un-redacted text`, `generation/data — a hand-written BAD_REQUEST keeps its
400 AND its exact text`, `v1/images keeps its transient-search 503`, and the
`WITH`/`WITHOUT cause` pair in `endpoint-helpers-driver-4xx.test.ts`. Their coverage comes
from **mutation**, not from red-at-base — M1/M2/M3 below are exactly the blind-delegation
mistake, and they turn the invalid-request invariants red at all five routes.

## Mutation results

13 mutants. Every one was **confirmed to have applied** before its result was read (a
substitution matching zero lines looks exactly like a kill), and the classifier is
credible because it reported a survivor.

| # | mutant | kind | result |
|---|---|---|---|
| M1 | `generation/data`: delete the validation split, delegate blindly | deletion | **killed** |
| M2 | `generation/resources`: same | deletion | **killed** |
| M3 | `generation/resources`: `!parsed.success` → `parsed.success` | branch inversion | **killed** |
| M4 | `download/models`: `errorResponse(500, GENERIC…)` → `err.message` | revert | **killed** |
| M5 | `tensor-metadata`: 422 literal → `err.message` | revert | **killed** |
| M6 | `v1/images`: restore the hand-rolled `{ error, code }` | revert | **killed** |
| M7 | `generation.service`: `throwBadRequestError` → `throwNotFoundError` | operand swap | **killed** |
| M8 | `blocks.router`: delete ONE `cause: err` | deletion | **killed** |
| M9 | `apps-shared.router`: `cause: e` → `cause: undefined` | wrong value | **SURVIVED → fixed, now killed** |
| M10 | tier1 test: `isCallerFacingText` always true (guard neutered, text left) | neutered guard | **killed** (by its own negative control) |
| M11 | tier1 test: real Prisma error → plain `Error` | fixture collapse | **killed → but see below** |
| M12 | `v1/images`: drop the `instanceof TRPCError` narrowing | deletion | **SURVIVED — equivalent mutant, see below** |
| M13 | `generation/{data,resources}`: delete `if (res.headersSent) return;` | deletion | **SURVIVED → fixed, now killed** |
| MB1 | `v1/images`: restore the 503 fall-through draft (headers set, then delegate) | revert | **killed** |
| MB2 | `v1/images`: 503 body key `error` → `message`, status + headers untouched | wrong key | **killed** |
| MC1 | `generation.service:528`: `throwBadRequestError` → `throw new Error` | revert | **SURVIVED pre-audit → covered, now killed** |
| MC2 | `tensor-metadata`: drop the `.int().gt(0).lte(INT4_MAX)` bound | revert | **killed** (4 tests) |
| MC3 | ledger: brace matcher → the old `[^{}]*?` regex | revert | **killed** (nesting control) |
| MC4 | `errorResponse` guard: status group back to `\d{3}` | revert | **killed** |
| MC5 | `errorResponse` guard: drop the concatenation rejection | deletion | **killed** |
| MD1a | tier1 guard: DELETE the `/* */` strip from the production pipeline | deletion | **killed** (was 46/46 green) |
| MD1b | tier1 guard: neuter the strip → `/ZZZ_NEVER_MATCHES/` | neutered guard | **killed** (was 46/46 green) |
| MD1c | tier1 guard: pipeline bypasses `stripComments` entirely | wiring | **killed** |
| MD1d | tier1 guard: drop the `//` filter from `stripComments` | deletion | **killed** (3 tests) |
| MD2 | tier1 guard: revert the arg scan to the `[^,)]+` status regex | revert | **killed** |
| MD3 | tier1 guard: restore the whole-argument `+`/`${` scan | revert | **killed** |
| MD4 | ledger: revert `trpcErrorBodies` to the quote-blind hand-rolled matcher | revert | **killed** |
| ME1 | real route: `errorResponse(\n 500, // it's the status\n err.message)` | live leak injected | **was 48/48 GREEN → now killed** |
| ME2 | real route: `errorResponse(500, err.message.replace(/'/g, ''))` | live leak injected | **was 48/48 GREEN → now killed** |
| ME3 | restore the drop (`no fallback` + filter nulls) **and** inject ME1 | regression | **killed by COUNT PARITY alone** |
| MD8 | `scanCallArgs`: remove the escape-skip | deletion | **survived twice → exemplar fixed, now killed** |

### Final round — a regression I introduced, and how it was caught

A last re-audit verified four of five claims and found that **claim 5 had introduced a
regression that defeated the guard**, on the TIER-1 disclosure guard for an
unauthenticated public download route.

`scanCallArgs` returns `null` when the argument list holds an odd number of quote
characters, and `callArgs` did `if (args && args.length >= 2) out.push(args[1])` —
**silently discarding the site**. Two live `errorResponse(500, err.message)` forwards
injected into the real route both went **48/48 green**: 13 sites, 12 extracted, and the
positive control was `length > 5`. **The parent version caught both.**

- **The fix already existed twenty lines away, in the sibling file, written by me.**
  `bracedBodies` falls back to a quote-blind scan when the quote-aware one returns
  `null`, precisely because silence is the worse failure direction. `callArgs` now does
  the same, and `callArgLists` surfaces the `null` instead of dropping it. Third
  occurrence of this exact lesson in this file family.
- **COUNT PARITY** — the number of `errorResponse(` occurrences must equal the number of
  extracted argument lists, zero nulls. It does not depend on the scanner being correct,
  which is the point. Demonstrated (ME3): with the regression restored and one shape
  injected, parity fires **while `every call site passes…` stays vacuously green**.
- Both auditor shapes were injected into the **real route** and confirmed to turn the
  guard red — not merely that a fallback exists.

**MD8, answered, after two wrong turns worth recording.** Deleting the escape-skip
outright leaves `else if` with no `if` — a syntax error, and a compile-error mutant
proves nothing; re-applied well-formed, it then **survived 51/51** against my first
exemplar (`'it\'s (bad)'`) because the parens *balance*, so the fallback lands correctly
and masks the bug. That exemplar looked like coverage and was not. The branch is
observable only when the literal holds an **unbalanced** delimiter — a comma is the
realistic one (`'That didn\'t work, try again'` gets split, yielding a non-literal and
turning the guard **red on safe code**, which is the failure direction the escape-skip
exists to prevent). Exemplar switched; mutant now killed. **Against the real route the
branch is unreachable** — no `errorResponse` argument there contains an escaped quote —
and that is stated in the file rather than implied.

Also taken (optional): the end-to-end comment-stripping check used
`some(a => a.includes('.message'))`, which goes red on a literal we wrote
(`'Your request is missing a .message field'`) *while blaming comment-stripping* — loud
on safe code and pointing the next author at the wrong cause. Now an exact-element
`not.toContain('err.message')`, preceded by a positive control that the exemplar comment
still exists, so "it did not reach the scan" cannot be satisfied by the comment having
been deleted.

### Delta re-audit — everything it found was in the guards the previous round hardened

The four substantive fixes above were **verified independently and stand**. The delta
pass found **no production-code defects**; all four findings were in my new guards,
which is exactly where a fix round leaves the next defect.

- 🔴 **F1 — the `/* */` hardening had NO COVERAGE.** Its test built a local string and
  applied its **own inline copy** of the regex, never touching the pipeline where the
  strip lives: deleting the strip from production left **46/46 green**, and so did
  replacing it with `/ZZZ_NEVER_MATCHES/`. *A test asserting against a copy of the thing
  it tests* is the purest form of the failure this file exists to catch — in that file.
  Now a named `stripComments`, called by the pipeline and driven by the tests, with
  separate assertions for the two failure modes (the strip breaking; the **pipeline
  ceasing to call it**, which nothing else would notice).
  **Correction to an earlier version of this section:** it said I "found and removed a
  vacuous assertion", which implied one had shipped. It had not — no such assertion
  existed at `9a59b0c738`. What happened is that I *wrote* `expect(source).not.toMatch(
  /\/\*…\*\//)` while fixing F1, found it could never fail (the route file has zero
  block comments), and **declined to add it**, documenting that in the file. The
  property and the reasoning are unchanged; the history as written was wrong.
- 🔴 **F4 — I had written a SECOND, quote-blind brace matcher.** `trpcErrorBodies`
  counted raw `{`/`}`, so one `}` inside a string truncated the body and a real no-cause
  bypass below it went invisible with the ledger green — reproduced on a real
  `src/server` file. That is the exact bug `scanBody` was made quote-aware for in #3881
  (`'oops }'`), regrown two hundred lines away **in the same file**. The fix is not
  quote-awareness written a third time: `jsonBodies` and `trpcErrorBodies` are now
  one-liners over a shared `bracedBodies()` built on `scanBody`. Note the failure
  direction: a truncated body still gets *inspected*, so the detector reports a
  confident "clean" on text it never read — worse than extracting nothing.
- 🟡 **F3 — the guard rejected literals we DID write.** The `/[+]|\$\{/` test scanned the
  whole argument *including inside the quotes*, so
  `errorResponse(403, "You don't have access to this file")` and
  `errorResponse(429, 'Try again in 5+ minutes')` both went red on the real route, which
  escapes only because its apostrophe message happens to use a backtick. **A guard that
  fails loud on safe code is how a security guard gets weakened** — the next author edits
  the guard. Now matches one complete, escape-aware literal, so a `+` or an apostrophe
  inside the text is just a character while a concatenation leaves residue outside it.
  The undeclared `[^'"]*` narrowing went with it.
- 🟡 **F2 — the widened status group still hid a call site, and its comment denied it.**
  `[^,)]+` cannot cross a `)`, so `errorResponse(getStatusCode(fileResult.status),
  err.message)` was never extracted and `every()` was vacuously true over it — the same
  vacuity, one shape over. Not hypothetical: `tensor-metadata.ts` has that exact helper
  and `download/models` open-codes the mapping at six sites, so consolidating them would
  have blinded the guard at every one. Replaced with a quote-aware argument-list scan.

MC1 is the one that matters most: it is the survivor the **blind audit** found, and it
survived the entire 15.4k-test suite. Its coverage is now mutation-verified — the revert
was confirmed applied at `:528` before the result was read, and the failure is the new
test's own assertion (`expected Error: not implemented to be an instance of TRPCError`),
with the two pre-existing tests in that file staying green, so the new test is
demonstrably what catches it rather than the pre-existing sibling.

### The new 503 body assertion, specifically

MB2 is the isolating mutant the new assertion exists for: it leaves the status and both
headers **identical** and changes only the body key, so nothing but a body assertion can
catch it. Result: **6 tests red, and every single failure is a body assertion** —
`expected { Object (message) } to deeply equal { Object (error) }`. In the
`SERVICE_UNAVAILABLE` test the three assertions that precede the body (status,
`Cache-Control`, `Retry-After`) all passed first, so the new assertion is demonstrably
what goes red, not a bystander.

The other five kills are the pre-existing raw-SDK transient tests, which now share the
unified branch — that they die together with the `TRPCError` test *is* the property
(both 503 paths emit one body). MB1, the realistic regression, reproduces the exact
draft that shipped the double-key and is killed on `Cache-Control` (`'no-store, max-age=0'`
vs `'no-store'`) — i.e. the drift I had previously *documented* is now *caught*.

**Three of these found real gaps, and all three were in my own work:**

- **M9** — the ledger's `cause` check was `!body.includes('cause')`, so `cause: undefined`
  (a fully re-opened bypass) read as *fixed*. Now the value is **extracted and compared**.
  🔴 My first fix for this was a negative lookahead, `cause\s*:\s*(?!undefined)`, which is
  **also wrong** — `\s*` is variable-length, so the engine backtracks it to zero width and
  tests the lookahead against `" undefined"`, which passes. The control caught it.
- **M11** — collapsing `prismaDriverError()` to a plain `Error` (which serializes to `{}`,
  the exact reason this class looked safe for years) left **all fifteen `LEAK:` tests
  green**; only one incidental log assertion noticed. Every `LEAK:` test is a
  `.not.toContain`, and a `.not.toContain` is trivially satisfied by a payload that never
  carried the secret. Added a **fixture-realism control** that asserts each fixture *does*
  carry every secret. A follow-up mutant that empties only the **pg** fixture is killed by
  that control **and by nothing else** — so without it, all five pg LEAK tests would have
  been vacuous.
- **M13** — `if (res.headersSent) return;` had no test at all. Added one that drives each
  route with an already-committed response and asserts the handler neither writes again
  nor rejects; it is red without the guard (`ERR_HTTP_HEADERS_SENT`).

**M12 survives, and it is an equivalent mutant rather than an unfixed gap.** The 503
branch's other conjunct already admits only `TRPCError`s in practice, because
`getHTTPStatusCodeFromError` reads `.code` and returns 500 for anything without a tRPC
code — so there is no assertion that could distinguish the two versions. One
qualification worth stating rather than burying: that equivalence holds *given that
nothing constructs a duck-typed `TRPCError`*, which I verified by repo-wide sweep; against
such an object the mutant would attach a spurious `Retry-After` to a 500. (The narrowing
now lives in the `trpcStatus` binding at the top of the unified 503 branch.)

## The ledger

- `KNOWN_UNFIXED_SAME_CLASS`: **57 → 53.** Four entries deleted (`v1/images/index.ts`,
  `generation/data.ts`, `generation/resources.ts`,
  `v1/model-files/[id]/tensor-metadata.ts`) and added to `FIXED_BY_THIS_CHANGE`, so a
  half-fix cannot pass by editing one list.
- **`download/models/[modelVersionId].ts` stays**, reclassified into the **OVER-REPORTS**
  block next to `download/attachments` and `download/vault`. Its `errorResponse` helper
  contains `res.json({ error: message })`, which the shape sweep flags whatever the call
  sites pass. It could only leave the list by restructuring the helper purely to dodge a
  regex, or by re-adding the name allowlist that buried this exact route for a whole
  round — both wrong directions. **The property that makes it an over-report is now
  asserted rather than asserted-in-a-comment:** a structural test checks that *every*
  `errorResponse(...)` call site passes a string literal or the shared constant, so
  re-introducing `errorResponse(500, err.message)` fails a test even though the ledger
  stays green either way. `download/attachments` and `download/vault` are untouched.
- The `no-cause` ledger's expected set is now `[]`, which is also what a broken detector
  reports — so three controls run before that assertion, including one that drives the
  **whole detector** and watches the number move (1 on a no-cause site, 0 on the
  `cause`-carrying version, in both the `err` and the `e` spelling).

## Things I looked at and deliberately did **not** do

- 🔴 **UNCLOSED GAP — the `no-cause` detector's name alternation.** An earlier version of
  this section filed this as a measured non-finding. **That was wrong, and it is a gap I
  chose not to close.** The detector anchors on the caught-binding names
  `e|err|error|ex`, so a `catch (caught)` — or any other binding — that forwards
  `.message` into a 4xx `TRPCError` without `cause` is **invisible to the ledger** and
  would never be flagged. What I actually measured is only that the *naive* widening is
  wrong: replacing the alternation with any identifier adds four sites in
  `routers/orchestrator.router.ts`, all false positives
  (`message: status.message ?? generationStatusDefaultMessage`, where `status` is
  operator config, not a caught error). That justifies not shipping *that* widening; it
  does not justify the blind spot. **The correct fix, which I did not build:** widen the
  identifier *and* require it to match the enclosing `catch`'s binding, which removes the
  false positives and the blind spot together. The gap is bounded — the ledger pins the
  known set at zero, so a regression in the four known files is caught — but a new file
  using an unusual binding name slips past.
- **Did not convert `throw new Error('missing version ids')`** in `getModelVersionGenerationData`.
  It is unreachable from these two routes — `getGenerationDataSchema` guarantees a non-empty
  `ids`, and the internal caller guards it — so converting it would be an unmotivated change
  to the tRPC surface.
- **Did not run prettier over the four router files.** Two are already prettier-unclean on
  the base branch; formatting them buries a 17-line change under ~190 lines of unrelated
  reformatting. Verified that prettier leaves every line I added alone.
- **Did not touch `download/attachments/[fileId].ts` or `download/vault/[vaultItemId].ts`** —
  over-reports by shape, not leaks, as #3881 recorded.
- **Did not change `handleEndpointError`.** Every fix here is at the route.

### Raised by the blind audit and deliberately left — recorded, not fixed

Each is a real observation; none is a defect this PR should absorb.

- **A third plain `Error` at `generation.service.ts:621`** (`'missing version ids'`).
  **Unreachable** from either route: `getGenerationDataSchema` guarantees a non-empty
  `ids`, and the internal caller guards it. Converting it would be an unmotivated change
  to the tRPC surface, and I would rather leave one unreachable arm inconsistent than
  claim a reachability I have not demonstrated.
- **The dead outer `throw` before `throwBadRequestError`** (`throw throwBadRequestError(…)`
  — the helper throws internally, so the outer `throw` never executes). It is the
  established local idiom for TS control-flow narrowing (`throw throwNotFoundError()`
  appears the same way), so changing it is a style edit across the file, not a fix.
- **`/api/v1/images`' heterogeneous error-key surface by status** — `error` at the
  pre-catch exits, `message` at a delegated 4xx, the full envelope at a genericized 5xx.
  Real, and **inherited from the base branch's pass-through arm**; making it uniform means
  changing `handleEndpointError`'s 4xx arm, which touches all 14 delegating routes and
  belongs in its own PR.
- **The asymmetric `headersSent` guard** across these routes. Pre-existing, and the audit
  found no constructible path to it beyond the one now covered by a test.

## Uncertainties, flagged as uncertainties

- ✅ **SETTLED, and no longer an unknown — the `tensor-metadata` pre-`try` path is a
  NON-LEAK, measured.** An earlier version of this section recorded it as the single
  largest unknown around this route and deliberately left it for an independent
  auditor, because an author establishing the safety of their own blind spot is not
  evidence. **The blind audit resolved it empirically**: it drove Next 16.3.0's own
  `apiResolver` with this repo's production config (`dev: false`,
  `propagateError: false` hardcoded in `pages-api.js`, no custom `onRequestError`) and
  confirmed that a throw escaping the handler yields `500 "Internal Server Error"` with
  the driver text **absent**; `withAxiom` re-throws without responding. So the escape is
  not a disclosure. That supersedes the unknown.
  **What it did surface is a different defect, now fixed — see "Fixed after the blind
  audit" above:** the same escape produced anonymous 500s with **no structured fault
  record at all**, and it was trivially reachable through an unbounded `?id=`.
- **The `/api/v1/images` 4xx shape change (item 1 above) is the change most likely to matter
  to a real consumer**, and I have no visibility into who parses that body. It is consistent
  with #3881 and forced by the ledger, but it is a judgement call, not a fact. (I originally
  labelled it the *highest-risk item in the PR*; after enumerating the route's non-2xx exits
  I no longer think that label is supported — the two 4xx a normal client provokes never
  enter the affected path — and the section above is corrected.)
- **Commit 2 changes no response body today.** `isDriverAuthoredMessage` is consulted by
  `handleEndpointError` and by nothing else; that helper serves the REST `/api/*` surface,
  and these 17 sites are tRPC routers whose errors go through `trpc.ts`'s own
  `errorFormatter` (which returns `shape` untouched except for lifting `cause.softBlock`, and
  never puts `cause` on the wire). So it is latent correctness plus closing the enumerated
  ledger gap — **not** a live leak being closed. I have corrected the `isDriverAuthoredMessage`
  JSDoc, which claimed otherwise. If you were expecting this commit to change an observable
  response, it does not.
- **`tensor-metadata`'s 422 message is now generic**, so the accordion shows *"Unable to read
  tensor metadata for this file"* instead of the upstream fetch error. That is a small
  debuggability loss for the user; the un-redacted message and stack still go to `_axiom`
  via the existing `logToAxiom` call. Say the word and I'll keep a coarse non-driver
  discriminator (e.g. "upstream fetch failed" vs "unsupported format") instead.
- **The `errorResponse` call-site guard is a text scan**, so it shares the ledger's blind
  spots: a message built into a local variable before the call, or an `errorResponse` alias,
  would not be seen.

## Verification

**Scope, stated, because it differs from the auditors':** all figures below are
`vitest run --project unit`. An independent audit run reported ~1254 files / ~16,985
tests by running a wider project scope; both are correct for what they measured.

- **Full unit suite: 982 files / 15,417 tests, 0 failures** (`--project unit`). Counted
  from the output rather than an exit code, with ` FAIL ` lines and timeout panics
  grepped separately; the `event-engine-common` submodule is populated, which is what
  the ~980/15.4k figure confirms.
- **`pnpm typecheck`: 0 errors — and the instrument was validated each round** by
  injecting a deliberate type error into a file actually changed that round and
  confirming it was reported.
  🔴 **With one scope limit worth stating rather than glossing:** `tsconfig.json`
  excludes `src/**/__tests__/**`, so `rest-error-envelope-ledger.test.ts` is **outside
  typecheck entirely** — a control injected there is silent, while the same control in
  `src/tests/api/…` reports `error TS2322`. Both were run to establish that. "Typecheck
  is clean" is therefore a claim about the production files and `src/tests/…`, and says
  nothing about the ledger.
- **`eslint`, same-tree comparison** (a base-branch *worktree* resolves a different
  effective config, so the base content is written into the same tree instead): **zero
  errors introduced in any round.** The larger earlier comparison was 13 errors at base
  and 13 at HEAD, identical files and rules; the last two rounds are 0 and 0. New test
  files add warnings only.
- **Mutation**: every mutant in the table was confirmed *applied* — by grepping the
  mutated text — before its result was read, because a substitution matching zero lines
  is indistinguishable from a kill. Files were restored from `cp -a` copies, never
  `git checkout --`.

---

https://claude.ai/code/session_01P99MVxyjxX7a4DRMgkqASe


ZacxDev:
## Preview Environment ❌

| | |
|---|---|
| **URL** | https://pr-3904.civitaic.com |
| **Commit** | `c47bade` |
| **Status** | ❌ Build failed |
| **Failed at** | `typecheck` |
| **Pipeline** | [View logs](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3904-xp6p9) |
| **Updated** | 2026-08-14 06:46:33 UTC |

_The preview build failed at the **typecheck** stage. Click the pipeline link above to view the logs and diagnose the issue._

PR #3905, opened 2026-08-13
fix(apps): gate the /apps sub-nav on the shared store-visibility predicate

ZacxDev:
## The defect

Two capabilities grant `/apps` store visibility, and the canonical predicate ORs them:

```ts
// src/components/Apps/resolveAppsPageAccess.ts — enforced in getServerSideProps
if (!(args.features?.appListings || args.features?.appBlocks)) return { notFound: true };
```

The sub-nav gated on only one of them:

```ts
// src/components/Apps/AppsSubNav.tsx:273
if (!features.appBlocks) return null;
```

So a cohort holding `app-listings` **without** `app-blocks-enabled` would load `/apps`
successfully — SSR grants access, the page renders — and get **no sub-navigation at all**.

## Why it is not reachable today

The current tester cohort holds **both** `app-listings=true` and `app-blocks-enabled=true`
(verified live against Flipt), so the two predicates agree and the disagreement is
unobservable. This is forward-looking, not a live outage.

The risk is structural: `appListings` exists *precisely* so the store catalog can widen
INDEPENDENTLY of the deliberately-held block runtime — that is the documented purpose in
`resolveAppsPageAccess`'s header and in `app-blocks-flag.ts`. The expected shape of the
public store launch is "widen `app-listings` alone", which is exactly the flip that makes
the two gates disagree.

## The fix — one rule, one place

Rather than open-code the OR a second time, the rule is extracted into a single shared
predicate `hasAppsStoreAccess(features)` in `~/shared/utils/app-blocks-access` (alongside
the existing `isAppDeveloper` / `isAppReviewer` access gates), and **every** store surface
calls it. The expression was already open-coded at five sites; consolidating them is what
makes a future disagreement impossible rather than merely unlikely:

| Surface | Before | After |
|---|---|---|
| `resolveAppsPageAccess.ts` (SSR gate) | `appListings \|\| appBlocks` | `hasAppsStoreAccess` |
| `AppsSubNav.tsx:273` (render gate) | **`appBlocks` only** ← the bug | `hasAppsStoreAccess` |
| `pages/apps/index.tsx:26` | `appListings \|\| appBlocks` | `hasAppsStoreAccess` |
| `pages/apps/store-preview/[slug].tsx:47` | `appListings \|\| appBlocks` | `hasAppsStoreAccess` |
| `AppListingsMarketplaceBody.tsx:221` | `!!(appListings \|\| appBlocks)` | `hasAppsStoreAccess` |
| `RelatedListings.tsx:56` | `!!(appListings \|\| appBlocks)` | `hasAppsStoreAccess` |

The four beyond the two named in the report are byte-identical restatements of the same
rule; adopting them is the consolidation, not scope creep. Behaviour is unchanged for
every cohort holding both flags — i.e. everyone today.

Deliberately **not** swept in: `/apps/installed`, `/apps/review`, `/apps/my-submissions`,
`/apps/revenue`, `/apps/run/<slug>` and the `blocks.*` procs all gate on `appBlocks` alone
**on purpose** — they need the block runtime, not just the catalog. Widening them is a
product decision, and the shared predicate's doc comment says so explicitly.

## The `:269` decision — the query gate stays on `appBlocks`

`enabled: !!features.appBlocks && !!currentUser` on `blocks.getNavSummary` is **left
alone**, and that is a decision, not an oversight.

A query gate must mirror the gate on the **procedure it calls**, not the gate on the page
it renders in. `getNavSummary` is `protectedProcedure.use(enforceAppBlocksFlag)`
(`blocks.router.ts:2537`) — the strict `app-blocks-enabled` check. Because it is a *query*,
that middleware short-circuits rather than throwing (`blocks.router.ts:260-264`), returning
the all-false summary **without running a single DB read**.

So for an `appListings`-only viewer, widening this `enabled` would buy a guaranteed
round-trip to a guaranteed all-false answer. And all-false is also the *correct* tab set for
that viewer: every conditional tab it feeds (Installed / My submissions / Revenue / Review)
points at a page that itself 404s without `appBlocks`. Showing them would recreate the
"tab straight into a 404" bug this component already fixed once for `Create`.

Reasoning is documented inline and pinned by three tests, so a later reader cannot "align"
the two flag reads on the assumption that everything in the file moves together.

## Honest scoping of the user-visible effect

The report framed the consequence as the cohort losing access to Installed / My submissions
/ Review / Create. **That overstates it, and the code disagrees.** Tracing it through:

- Installed / My submissions / Revenue / Review are driven by `getNavSummary`, which is
  server-gated on `app-blocks-enabled` and returns all-false for this cohort regardless.
  Those tabs stay hidden either way — correctly.
- `AppsSubNavView` also hides itself below two qualifying tabs. An `appListings`-only
  **non-author** qualifies for Marketplace alone, so they get no bar before *or* after this
  change — via the collapse rule, not via the gate.
- The one cohort whose rendered output actually changes is
  `{appListings, NOT appBlocks, isAuthor}`: previously no bar, now `[Marketplace, Create]`.

So the practical value here is **coherence, not a restored feature**: the two gates can no
longer drift, and the store predicate now has one definition instead of six. Worth stating
plainly rather than claiming a fix that the collapse rule and the server gate already cover.

## Known exposure this surfaces (pre-existing, not fixed here)

`/apps/submit`'s SSR gate is `appBlocksAuthor` + `isAppDeveloper` — it does **not** require
`appBlocks`. Its client body carries an extra `if (!features?.appBlocks) return <NotFound />`
(`submit.tsx:68`) that its own SSR gate does not. While the sub-nav gated on `appBlocks`,
the Create tab could only render when that flag was already true, so this check was
unreachable. It becomes reachable for the (empty today) cohort above, who would SSR into
the page and then hit a client-side 404.

The outlier is `submit.tsx`'s body, not the tab. Re-inlining `appBlocks` into the sub-nav
would recreate the very drift this PR removes, and whether authoring requires the block
runtime is a product call. Documented in a `⚠️` block above `SUB_NAV_LINKS` and left to a
follow-up.

## Mutation matrix

Every mutant was applied to the real source, both suites re-run, and the **specific failing
test names** read — not an exit code. Restores were `cp -a` from a backup, never
`git checkout --`.

| # | Mutant | Unit | Component | Killed on its own assertion? |
|---|---|---|---|---|
| M0 | baseline (unmutated) | 21 passed | 58 passed | — |
| **M1** | **revert `:273` → `!features.appBlocks` (the original defect)** | 21 passed | **4 failed** | ✅ `🔴 appListings ONLY (appBlocks OFF) → the sub-nav RENDERS` fails on its own `toBeInTheDocument`. Unit correctly unaffected — the mutant is component-only. |
| M2 | `hasAppsStoreAccess`: OR → **AND** | 6 failed | 17 failed | ✅ `appListings ONLY → access` and `appBlocks ONLY → access` both fail |
| M3 | `hasAppsStoreAccess`: **inverted** | 14 failed | 20 failed | ✅ every arm of the 2×2 flips |
| M4 | `hasAppsStoreAccess`: **constant `true`** | 7 failed | 2 failed | ✅ exactly the two "renders nothing" tests + all fail-closed unit cases |
| M5 | `hasAppsStoreAccess`: **constant `false`** | 7 failed | 18 failed | ✅ all three grant cases |
| M6 | drop the `appListings` arm (the exact historical drift) | 2 failed | 4 failed | ✅ `appListings ONLY → access` + the headline sub-nav test |

M1 is the required "watched it fail on pre-change code" evidence: red with the gate
reverted, green at HEAD.

Note M4's shape — constant `true` kills only **2** component tests, and they are precisely
the two absence assertions. That narrowness is what makes the barrier proof below meaningful.

### The absence assertions provably CAN fail

The report flagged that under a React 18 concurrent root, an absence assertion taken
immediately after render reads an empty container and passes regardless. Measured rather
than assumed, by neutering the gate to constant `true`:

| Configuration | "renders nothing" tests |
|---|---|
| **With** the awaited `RENDER_BARRIER` | **FAIL** ✅ (they can observe the defect) |
| **Without** it (the `await expect.element(...)` deleted) | **PASS** ❌ structurally unfailable |

So the barrier is load-bearing here, not ceremonial. That finding is recorded in the
suite's own comment so it cannot be removed as boilerplate. Each absence test is also
paired with an explicit positive control proving the same readers *do* observe a bar when
the gate opens — a reassuring zero is otherwise indistinguishable from a probe wired to
nothing.

## Test coverage added

- `__tests__/hasAppsStoreAccess.test.ts` (14 unit tests) — the full 2×2 matrix, fail-closed
  on absent/null/empty features, strict-boolean return, **plus a seam test asserting
  `resolveAppsPageAccess` agrees with the predicate across the whole matrix**. A component
  and a resolver each tested in isolation can both be green while the pair is incoherent;
  this pins the relationship.
- `AppsSubNav.storeGate.browser.test.tsx` (11 browser tests) — the sub-nav's rendered
  output for all four flag combinations, the collapse-vs-gate distinction, and the `:269`
  query-gate decision.

This suite mocks `useQuery` so it **honours `enabled`**, unlike the sibling hydration suite
which returns data unconditionally (right for what *that* file tests). Here the `enabled`
predicate is load-bearing for the answer, so the mock matches production and the decision
above is directly observable.

`src/components/Apps/__tests__/resolveAppsPageAccess.test.ts` is **unmodified** and green.

## Verification

- Typecheck: **0 errors**, baseline **0 errors** on a clean checkout of the base commit
  (`pnpm typecheck`, after `git submodule update --init --recursive` — an unpopulated
  `event-engine-common` produces phantom errors).
- Component suites touched by this change: **93 passed / 93** across 8 files
  (`AppsSubNav` ×4, `AppListingsMarketplaceBody` ×4). Baseline for the three pre-existing
  `AppsSubNav` suites measured first: 52/52.
- Unit: 45 passed / 45.
- ESLint **0 errors, 0 warnings across 9 files** — file count read from JSON output, and
  validated with a negative control (planted `const x: any` → 2 findings) so the clean
  verdict is not a claim about an empty file list. Prettier clean on all 9.

## What I did NOT verify

- **No live/browser check against a real `app-listings`-only account.** The cohort does not
  exist in Flipt today (both flags resolve true for the tester segment), so it cannot be
  driven end-to-end without changing flag state in production. Everything above is code
  tracing + component tests.
- **Server-side flag resolution was read, not exercised.** The `:269` reasoning rests on
  reading `enforceAppBlocksFlag` and `isAppListingsEnabled`; I did not stand up Flipt to
  confirm the short-circuit at runtime. The existing
  `blocks.router.getNavSummary.test.ts` does assert the flag-off path returns the all-false
  shape and runs no DB query, which is the load-bearing half.
- **`/apps/submit`'s client-body 404 exposure is reasoned, not reproduced** — the cohort
  that would hit it is empty, so there is no way to observe it today.
- The full component project was not run — only the 8 files touching the changed modules.
  Those are report-only in CI regardless; the live gate is `tekton / typecheck`, which is
  clean.

### DIFF
--- a/src/components/Apps/AppListingsMarketplaceBody.storeGate.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.storeGate.browser.test.tsx
@@ -1,0 +1,187 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { page } from 'vitest/browser';
+import { useRouter } from 'next/router';
+// `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
+import { renderWithProviders } from '../../../test/component-setup';
+import type * as TrpcMod from '~/utils/trpc';
+import type { ListingCard } from '~/server/schema/blocks/app-listing-read.schema';
+
+/**
+ * 🔒 `AppListingsMarketplaceBody` — the STORE-VISIBILITY gate on the grid query.
+ *
+ * WHY A DEDICATED FILE: the shared-predicate extraction converted six sites, but
+ * an adversarial audit reverted four of them to `!!features.appBlocks` and the
+ * entire existing suite stayed green — the mutant survived. The call-site ledger
+ * (`__tests__/appsStoreAccessCallSites.test.ts`) now pins all six STRUCTURALLY,
+ * but a structural check cannot see a call passed the wrong argument. This file
+ * is the BEHAVIOURAL half for the busiest of the four: it drives the flag matrix
+ * and observes what the component actually does with `enabled`.
+ *
+ * Unlike the sibling `AppListingsMarketplaceBody.browser.test.tsx` (which pins
+ * `useFeatureFlags` to a constant and ignores query options), this file makes the
+ * flags a VARIABLE and captures the `enabled` option — the two things the gate is
+ * made of.
+ *
+ * 🔴 The gate here mirrors the SERVER read gate (`enforceAppListingsReadFlag` →
+ * `isAppListingsEnabled`, which ORs both flags), which is why this query DOES
+ * widen with the store — in deliberate contrast to `blocks.getNavSummary`, whose
+ * `enabled` stays on `appBlocks` alone because ITS proc gates on
+ * `enforceAppBlocksFlag`. See `AppsSubNav.storeGate.browser.test.tsx`.
+ */
+
+function makeCard(id: string, name: string): ListingCard {
+  return {
+    id,
+    slug: `slug-${id}`,
+    kind: 'onsite',
+    name,
+    tagline: 'tag',
+    category: null,
+    contentRating: null,
+    iconUrl: null,
+    coverUrl: null,
+    creator: null,
+    recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
+    reviewCount: 0,
+    kindData: {
+      kind: 'onsite',
+      appBlockId: `blk-${id}`,
+      hasPage: false,
+      liveUrl: `https://slug-${id}.civit.ai`,
+    },
+  };
+}
+
+const mocks = vi.hoisted(() => ({
+  flags: {} as Record<string, boolean>,
+  items: [] as ListingCard[],
+  /** Every `enabled` value the grid query was called with, in order. */
+  enabledSeen: [] as unknown[],
+}));
+
+vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
+vi.mock('~/providers/FeatureFlagsProvider', () => ({ useFeatureFlags: () => mocks.flags }));
+vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
+vi.mock('~/hooks/useIsMobile', () => ({ useIsMobile: () => false, isMobileDevice: () => false }));
+
+// Spread the REAL module and override only `trpc` (local-rules/no-wholesale-module-
+// mock): a hand-written replacement silently breaks every importer the day
+// '~/utils/trpc' grows an export this factory omits — the whole FILE then fails to
+// load with 0 tests collected and no failing assertion.
+vi.mock('~/utils/trpc', async (importOriginal) => ({
+  ...(await importOriginal<typeof TrpcMod>()),
+  trpc: {
+    appListings: {
+      listAvailable: {
+        useInfiniteQuery: (_input: Record<string, unknown>, opts?: { enabled?: boolean }) => {
+          mocks.enabledSeen.push(opts?.enabled);
+          // Honour `enabled` like the real hook: a disabled query resolves no data,
+          // so a gated-out viewer gets an empty grid rather than listings.
+          return {
+            data: opts?.enabled
+              ? { pages: [{ items: mocks.items, nextCursor: undefined }] }
+              : undefined,
+            isLoading: false,
+            isFetchingNextPage: false,
+            fetchNextPage: vi.fn(),
+            hasNextPage: false,
+          };
+        },
+      },
+    },
+  },
+}));
+
+const { AppListingsMarketplaceBody } = await import('./AppListingsMarketplaceBody');
+
+// Not a real hook: the scaffold mocks `next/router` with `useRouter: () => router`,
+// a plain function returning a singleton, so module-scope is fine (and is the
+// established idiom — a per-file `vi.mock` silently loses to the setup-file mock).
+// eslint-disable-next-line react-hooks/rules-of-hooks
+const router = useRouter();
+
+/**
+ * 🔴 RENDER BARRIER — required before any "did not render" assertion.
+ *
+ * React 18 commits on a LATER task, so a synchronous absence check straight after
+ * `renderWithProviders` reads an empty container and passes regardless of what the
+ * component did. Measured on the sibling sub-nav suite: with the barrier removed,
+ * the absence tests pass in ~4ms even with the gate neutered.
+ */
+const RENDER_BARRIER = 'render-barrier';
+const RenderBarrier = () => <div data-testid={RENDER_BARRIER} />;
+
+async function renderBody() {
+  renderWithProviders(
+    <>
+      <RenderBarrier />
+      <AppListingsMarketplaceBody />
+    </>
+  );
+  await expect.element(page.getByTestId(RENDER_BARRIER)).toBeInTheDocument();
+}
+
+beforeEach(() => {
+  mocks.flags = {};
+  mocks.items = [makeCard('a', 'Alpha App'), makeCard('b', 'Bravo App')];
+  mocks.enabledSeen = [];
+  // A FRESH object each test: `useZodRouteParams` memoises on the `query` reference.
+  router.query = {};
+});
+
+describe('AppListingsMarketplaceBody — the grid query uses the shared store predicate', () => {
+  test('🔴 appListings ONLY (appBlocks OFF) → the grid query is ENABLED and listings render', async () => {
+    // The exact mutant that survived: reverting this site to `!!features.appBlocks`
+    // disables the query for this cohort and fails HERE, on these assertions.
+    mocks.flags = { appListings: true, appBlocks: false };
+    await renderBody();
+
+    await expect.element(page.getByText('Alpha App')).toBeInTheDocument();
+    expect(mocks.enabledSeen.some((e) => e === true)).toBe(true);
+    expect(mocks.enabledSeen.every((e) => e !== false)).toBe(true);
+  });
+
+  test('appBlocks ONLY (appListings OFF) → enabled (the OR-fallback cohort)', async () => {
+    mocks.flags = { appListings: false, appBlocks: true };
+    await renderBody();
+
+    await expect.element(page.getByText('Alpha App')).toBeInTheDocument();
+    expect(mocks.enabledSeen.some((e) => e === true)).toBe(true);
+  });
+
+  test('BOTH flags true → enabled (today’s live cohort, unchanged)', async () => {
+    mocks.flags = { appListings: true, appBlocks: true };
+    await renderBody();
+
+    await expect.element(page.getByText('Bravo App')).toBeInTheDocument();
+    expect(mocks.enabledSeen.some((e) => e === true)).toBe(true);
+  });
+
+  test('NEITHER flag → the query is DISABLED and no listing renders', async () => {
+    mocks.flags = { appListings: false, appBlocks: false };
+    await renderBody(); // barrier awaited inside — absence below is a real observation
+
+    expect(mocks.enabledSeen.length).toBeGreaterThan(0); // the hook did run
+    expect(mocks.enabledSeen.every((e) => e === false)).toBe(true);
+    expect(page.getByText('Alpha App').elements()).toHaveLength(0);
+    expect(page.getByText('Bravo App').elements()).toHaveLength(0);
+  });
+
+  test('flags ABSENT entirely (Flipt down) → disabled (fails closed)', async () => {
+    mocks.flags = {};
+    await renderBody();
+
+    expect(mocks.enabledSeen.every((e) => e === false)).toBe(true);
+    expect(page.getByText('Alpha App').elements()).toHaveLength(0);
+  });
+
+  // 🔴 POSITIVE CONTROL for the two absence tests above: if `getByText` were
+  // wired to nothing it would report zero for every input, and both would be
+  // vacuously green.
+  test('POSITIVE CONTROL: the same reader DOES find the card when the gate opens', async () => {
+    mocks.flags = { appListings: true, appBlocks: false };
+    await renderBody();
+    await expect.element(page.getByText('Alpha App')).toBeInTheDocument();
+    expect(page.getByText('Alpha App').elements().length).toBeGreaterThan(0);
+  });
+});
--- a/src/components/Apps/AppListingsMarketplaceBody.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.tsx
@@ -21,6 +21,7 @@ import {
 } from '~/components/Apps/recentlyOpenedAppsStore';
 import { useFeatureFlags } from '~/providers/FeatureFlagsProvider';
 import type { ListingCard, ListingSort } from '~/server/schema/blocks/app-listing-read.schema';
+import { hasAppsStoreAccess } from '~/shared/utils/app-blocks-access';
 import { trpc } from '~/utils/trpc';
 
 /**
@@ -214,11 +215,12 @@ export function AppListingsMarketplaceBody() {
         limit: 24,
       },
       {
-        // W13 (PR-W1a/D8): store-visibility gate = dedicated `appListings`
-        // OR-falling-back to `appBlocks`. Mirrors the server read gate
-        // (`enforceAppListingsReadFlag` → `isAppListingsEnabled`). Zero behavior
-        // change today (the `app-listings` flag doesn't exist yet).
-        enabled: !!(features.appListings || features.appBlocks),
+        // W13 (PR-W1a/D8): store-visibility gate = the SHARED `hasAppsStoreAccess`
+        // predicate (dedicated `appListings` OR-falling-back to `appBlocks`).
+        // Mirrors the server read gate (`enforceAppListingsReadFlag` →
+        // `isAppListingsEnabled`), which is why this query — unlike
+        // `blocks.getNavSummary` — DOES widen with the store.
+        enabled: hasAppsStoreAccess(features),
         getNextPageParam: (lastPage) => lastPage.nextCursor,
       }
     );
--- a/src/components/Apps/AppsPageLayout.tsx
+++ b/src/components/Apps/AppsPageLayout.tsx
@@ -22,9 +22,16 @@ import { AppsSubNav } from '~/components/Apps/AppsSubNav';
  * subtitle and right-aligned header actions become props so the header geometry
  * is uniform; only the content slot differs.
  *
- * Flag-gating (`features.appBlocks`) + any per-page access redirect stay on the
- * page (`getServerSideProps` / the in-component `NotFound` guard) — this layout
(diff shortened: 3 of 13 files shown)


ZacxDev:
## Audit follow-up — pushed `a44fc351b4`

Posting this as a comment rather than editing the body, since the original has already been read. **Three corrections to claims in the PR body**, plus the work that backs them.

---

### ❌ Correction 1 — the consolidation claim was overstated, and it was *measured* so

The body said consolidating makes a future disagreement "impossible rather than merely unlikely." **That was ~1/3 true when written.** An adversarial audit reverted the four sites I had converted but not covered —

- `AppListingsMarketplaceBody.tsx:221`
- `RelatedListings.tsx:57`
- `pages/apps/index.tsx:26`
- `pages/apps/store-preview/[slug].tsx:48`

— back to `!!features.appBlocks`, re-ran the **exact suite the PR cites**, and got **93/93 component + 21/21 unit, still green**. The mutant survived completely. Only `resolveAppsPageAccess` and `AppsSubNav` were genuinely enforced; the other four were converted-and-unpinned, which is a comment, not an invariant.

Rather than soften the claim, I made it true. Two suites added:

**`src/components/Apps/__tests__/appsStoreAccessCallSites.test.ts`** — the call-site **ledger**. Pins the exact set of six store-visibility modules, asserts each *imports and calls* the shared predicate, and asserts none re-inlines the boolean. Fails when the set **grows** (a new store surface that forgot the predicate) or **shrinks** (a silent revert).

Two things that make it non-vacuous, both learned the hard way in this file:
- The comment/string masker is **self-tested with a positive control**. It has to be: the predicate's own doc comment and `resolveAppsPageAccess`'s header both contain the literal text `features.appListings || features.appBlocks`, so an unmasked scan reports the very files that were fixed.
- The first version masked *strings too*, which blanked `from '~/shared/utils/app-blocks-access'` and failed all six site checks **against correct code**. There is now a dedicated regression test for that, because a scanner that cannot see its own target is the failure mode that produces a confident wrong verdict.

**`src/components/Apps/AppListingsMarketplaceBody.storeGate.browser.test.tsx`** — the **behavioural** half for the busiest of the four. A structural check proves a site *calls* the predicate; it cannot see a call passed the wrong argument. This drives the flag matrix and observes the query's actual `enabled`.

#### Re-running the auditor's exact mutant

| Variant | Unit | Component | Killed |
|---|---|---|---|
| A0 baseline | 35 passed | 76 passed | — |
| **A1 — the auditor's mutant** (4 sites → `!!features.appBlocks`, imports left in place) | **4 failed** | **2 failed** | ✅ one unit failure per site + the marketplace grid behaviourally |
| **A2 — full revert incl. imports** (what lint would force) | **5 failed** | **2 failed** | ✅ A1's failures **plus** ledger-exactness |

A1 is deliberately the *harder* variant: leaving the imports in place means ledger-exactness still passes, so the kill has to come from the per-site call assertion. It does.

#### Where coverage is still only structural — stated, not implied

`RelatedListings`, `pages/apps/index` and `store-preview/[slug]` are pinned **against reversion**, not against a wrong-argument call. They are Next pages / a rail with no component harness, and standing one up costs more than the gate is worth. **The accurate claim is therefore:** all six sites are pinned against reversion; two (`resolveAppsPageAccess`, `AppsSubNav`) plus `AppListingsMarketplaceBody` are additionally pinned behaviourally. That is what the tests enforce, and the test file says so in its own header.

---

### ❌ Correction 2 — "unreachable" was wrong about `/apps/submit`

The body described `submit.tsx:68`'s `if (!features?.appBlocks) return <NotFound />` as unreachable. **It was unreachable *via the sub-nav's Create tab*, not unreachable.** Direct navigation always reached it — a bookmark, a pasted link, or the `?edit=<listingId>` deep link from my-submissions. The code path was always live; what was gated was one route into it.

Now filed as **#3906** with the SSR-vs-client-body mismatch, the reason it is not currently live (`app-blocks-author` is a strict subset of `app-blocks-enabled`), and two candidate fixes — it needs a product decision about whether authoring requires the block runtime, not a mechanical one. The code comment cites the issue instead of describing the hazard inline.

---

### ❌ Correction 3 — "five sites" vs "six"

The body said "five sites", the predicate's doc enumerated six. Both were defensible and together confusing. Standardised on: **six store-visibility sites; five spelled the OR out, and the sixth (`AppsSubNav`) spelled only half of it.**

---

### ➕ The predicate's parameter type was too weak (new finding, fixed)

`AppsStoreFeatureFlags` was hand-written as `{ appBlocks?: boolean; appListings?: boolean }`. Measured consequence: if `FeatureAccess` ever drops or renames `appListings`, the **old open-coded** `features.appListings` reads would fail with `TS2339`, but `hasAppsStoreAccess(features)` **keeps compiling and silently degrades every store surface to `appBlocks`-only**.

That rename is not hypothetical — it is exactly the documented *"drop the OR-fallback once `app-listings` is the sole source of truth"* step at GA. So the extraction would have **removed** a compile-time signal the open-coded version had. Now:

```ts
export type AppsStoreFeatureFlags =
  | Partial<Pick<FeatureAccess, 'appBlocks' | 'appListings'>>
  | null
  | undefined;
```

applied to `resolveAppsPageAccess` and `resolveLegacyAppRoute` as well, so the rename breaks at every gate. The `FeatureAccess` import is type-only (the established pattern — `src/shared/data-graph/generation/context.ts` does the same).

**Compile-time controls** (batched, via the repo's `pnpm typecheck`):

| Control | Expected | Result |
|---|---|---|
| A — today's real call shapes | compile clean | ✅ 0 errors |
| B — unrelated object `{ somethingElse: true }` | reject | ✅ `TS2353` (weak-type detection intact) |
| B — typo `{ appListing: true }` | reject | ✅ `TS2561` — *"Did you mean to write 'appListings'?"* |
| B — wrong value type `{ appListings: 'yes' }` | reject | ✅ `TS2322` |
| **C — rename `appListings` in `FeatureAccess`** | **break at the type** | ✅ **`TS2344` at `app-blocks-access.ts:75`** |

🔴 **Instrument validation, because the first attempt was wired to nothing.** My initial harness used bare `tsc --noEmit -p tsconfig.json` and reported **0 errors for all four probes — including the three that must fail.** Bare `tsc` OOMs on this repo and prints nothing, which is indistinguishable from "clean". `--listFilesOnly` confirmed the probe *was* in the program, so the silence was the tool, not the code. Re-run through `pnpm typecheck` (heap cap 8192 MB) the controls behave as above. Worth flagging for anyone else typechecking this repo ad hoc.

---

### Also in this push

Two comments that still described the gate as `features.appBlocks`: `pages/apps/index.tsx` and `AppsPageLayout.tsx`. The latter now spells out the distinction that matters — store surfaces gate on `hasAppsStoreAccess`, block-runtime surfaces gate on `appBlocks` alone, and they are **not** interchangeable.

### Verification of this push

- Typecheck: **0 errors, cold run** (`.tsbuildinfo` deleted first — the incremental run finished in 10s and I did not want a stale 0). Instrument validated by a planted error.
- Component: **99 passed / 99** across 9 files. Unit: **777 passed / 777** across all 45 `Apps/__tests__` files.
- ESLint **0 errors / 0 warnings across 13 files** (count read from JSON, negative control planted earlier); Prettier clean on all 13.
- No production behaviour change in this commit — tests, types and comments only.

### Still not verified

- Unchanged from the original body: no live run against a real `app-listings`-only account (the cohort does not exist in Flipt), and the `/apps/submit` exposure is reasoned rather than reproduced.
- 🔴 **`tekton / typecheck` has not registered on this PR** (it did on #3899), so the local cold typecheck above is the only evidence for the type work. Worth a maintainer glance at the check rollup before merge rather than assuming an empty rollup means "no gate required".

ZacxDev:
## Delta-audit follow-up — pushed `5bb289e2f5`

Both findings were about the new guard's *reach*, not the shipped code. Production behaviour is untouched by this commit — it is tests, one type, and comments.

---

### 1. 🔴 The ledger's masker had a proven blind region — fixed with the real parser

**Reproduced the auditor's control first, on the unmodified masker**, so the fix has a before-state to be measured against rather than asserted:

| Probe: a live `f.appListings \|\| f.appBlocks` in `installed.tsx` | Before | After |
|---|---|---|
| planted **after** the JSX apostrophe (line 132) | 🔴 **INVISIBLE** — 14/14 pass, exit 0 | ✅ **CAUGHT** — suite fails |
| planted **before** it (line 60) | ✅ CAUGHT | ✅ CAUGHT |

The old masker treated every ASCII `'` as a string delimiter, so `The app's data…` in **JSX text** opened a phantom string that ran to EOF.

**Fix: `ts.createSourceFile(..., ScriptKind.TSX)`** — blank the ranges the AST classifies as literal (`StringLiteral`, template head/middle/tail, `RegularExpressionLiteral`, `JsxText`), then strip comments from *that* result, which is safe because every string is already blank. Template `${...}` holes are children of the template node and stay **live** — they are real code. `maskComments` takes its ranges from the parser too, so the import-path assertions cannot desync either.

Over-blanking measured before/after (chars of live content wrongly blanked):

| File | Old | New |
|---|---|---|
| `installed.tsx` | 10,339 / 20,037 (51.6%) | 4,989 (24.9%) — **5,350 chars of live code recovered** |
| `my-submissions.tsx` | 2,572 / 6,762 (38.0%) | 1,956 (28.9%) |

**On why the old controls couldn't see it:** they were all synthetic one-liners containing no JSX — a scanner allowlisting its own examples. New cases use realistic shapes: the JSX apostrophe, an assertion anchored on the **real `installed.tsx`** (a fixture can be fixed while the real file still desyncs — different nesting, entities), nested quotes/templates/regex, and a hard bound on how much of a file may be blanked.

I also took the secondary finding, since it was cheap: `INLINED_GATE` is now proximity + logical-operator over whitespace-flattened code, so the prettier-wrapped multi-line OR, the De Morgan negation and the `??` chain all match. `\b` anchors stop `appBlocksPages` / `appBlocksAuthor` — different flags with their own rules — being misread as `appBlocks`.

---

### 2. 🟡 The seventh site — claim scoped, not silently converted

`components/AppLayout/AppHeader/appsNavVisibility.ts:28` gates the user-menu "Apps" → `/apps` entry on `appBlocks` alone. **Not converted**, exactly as advised: that file's own doc says the entry *"stays gated on `appBlocks`"*, so it is a standing decision, not drift.

What was wrong was my claim, so that is what changed:
- `STORE_GATE_SITES` now carries an explicit **`NOT in this ledger, on purpose`** entry for it, mirroring the block-runtime exclusion — with the rationale, the note that `SCAN_ROOTS` is structurally blind to `components/AppLayout`, and the consequence.
- The `ledger is EXACT` test carries an inline ⚠️ that its exactness is scoped to `SCAN_ROOTS` and cannot speak for that file.
- `hasAppsStoreAccess`'s doc no longer says "every store surface" — it names the six under `components/Apps` + `pages/apps` and states the exclusion.

**Filed as #3907**, because the consequence outlives this PR: for a `{appListings, NOT appBlocks}` cohort the store renders but has **no in-product entry point** — that menu item is the only route TO `/apps` (the sub-nav's Marketplace tab only appears once you are already on `/apps/*`). That should be settled *before* `app-listings` is widened, which is the flip the whole decoupling exists to enable.

### 3 & 4. Header corrections

Structural-only is **three** sites, not two — `RelatedListings` belongs there (its only test covers the pure selection helper and never mounts the component). And the header now records that **the component suites are report-only in CI**, so in CI all six sites are pinned structurally only, by this ledger — which is exactly why finding 1 mattered more than it looked.

---

## 🔴 Test-merge against current `main` — one thing for whoever merges

`main` moved 5 commits and touched `AppsSubNav.tsx` (#3891 added "My apps" / "Invites" tabs). GitHub reports `MERGEABLE`, but that only means no textual conflict — and #3891's own diff comment describes this very table being auto-merged into an un-gated bug once already. So I test-merged rather than reasoned about it.

**Result: the merged tree is clean.** Typecheck 0 errors (cold); 75 component tests across 5 files; **3,685 unit tests across 172 files** (`components/Apps/__tests__` + `server/services/blocks/__tests__`). No interaction: #3891 gates its two new tabs on `c.isAuthor`, and their summary flags come from `getNavSummary`, which this PR deliberately leaves on `appBlocks` — so for an `appListings`-only viewer they resolve all-false and stay hidden, consistent with the `:269` decision.

Two notes from that exercise:

1. **A stale generated Prisma client is not a merge failure.** The merged tree first reported `TS2339: Property 'feedback' does not exist on PrismaClient` — from `main`'s new `feedback.service.ts`. That is `node_modules` being generated at my base commit, before `main` added the `feedback` model. `pnpm run db:generate` → 0 errors. Flagging because it will look like a real break to the next person who merges without regenerating.

2. **`main` just introduced a second comment-stripper** — `test/strip-comments.ts` (`stripCommentsAndStrings`), created explicitly *because* the technique had been duplicated. This ledger is now a third implementation, which deserves an explicit answer rather than silence:
   - It **cannot** be reused here: it strips string literals, and my import assertions match on the module specifier `from '~/shared/utils/app-blocks-access'` — a string. That is the exact regression this file already hit once.
   - I checked its behaviour on the failing case rather than assuming: it is **line-bounded** (`[^'\\\n]`), so it does **not** have the EOF-swallow bug — it would have caught the planted probe. It does leave JSX text live, i.e. it under-strips where its own doc says it is biased toward over-stripping.
   - **Recommended follow-up (post-merge, not here):** consolidate onto the parser-based approach — it is strictly more accurate and would let both ledgers share one module. Not done in this PR because that file does not exist on this branch, and reaching across to rewrite it here would be the kind of scope creep this PR has been trimming.

## Verification of this push

- Typecheck **0 errors, cold** (`.tsbuildinfo` deleted). Instrument previously validated with a planted error.
- Unit **782 / 782** across all 45 `Apps/__tests__` files; component **99 / 99** across 9 files.
- ESLint **0 errors / 0 warnings**, count read from JSON; Prettier clean.
- Ledger suite itself: 19 tests, up from 14.

## Still not verified

- No live run against a real `app-listings`-only account — the cohort does not exist in Flipt.
- `#3906` and `#3907` are reasoned from code, not reproduced; both cohorts are empty today.
- 🔴 **`tekton / typecheck` has still not registered on this PR.** The local cold typecheck (on both my branch and the test-merge) remains the only type evidence. Worth a maintainer glance at the rollup rather than reading an empty check list as "no gate required".

PR #3911, opened 2026-08-14
fix(buzz): pay creators in the currency the buyer spent

JustMaier:
## What was wrong

A green-funded purchase paid the seller **yellow**.

Not a decision. The charge named no destination account, so the Buzz service applied its yellow default (`buzz.service.ts`, `toAccountType ?? 'yellow'`, added 2025-07-03) and converted on the way through. **Every green paid-access purchase since 2025-10-30 landed this way — 2,409 legs**, ~5.3M Buzz in the last quarter alone.

Placements had it twice over:

1. The escrow debit offered **both** green and yellow (`fromAccountTypes: PLACEMENT_SPEND_TYPES`), so one placement could be funded from two accounts — leaving no single currency to pay back in kind. One such placement exists (dev only, `placement-70`: 38 green + 32 yellow).
2. The payout leg named no account either, so it flattened to yellow whatever went in.

Neither front-end surface constrained it: `useAvailableBuzz` already returns exactly one currency per domain, but `RemixGallerySubmitModal` and `DraftSticker` both passed the pair.

## Evidence

Verified against the Buzz service, not just ClickHouse:

- The escrow account (id 0) has Green `lifetimeBalance: **0**` while green placement holds have flowed into it for months. That counter only goes up — zero means green has never once been credited there.
- A creator's **yellow** ledger contains their green-funded sale; their **green** account for the same day holds one unrelated 1-Buzz fee.
- Green *is* a routine credit destination elsewhere (222,730 legs since July), so this is a real absence, not a missing capability.
- All 1,982 green→yellow paid-access legs are `type = 'purchase'`, and 695/695 model versions behind them exist in prod.

## What this changes

| | |
|---|---|
| `Placement.spendType` | new nullable column, written by the hold |
| Escrow hold | draws from **one** account and names it as the destination, so escrow holds green instead of converting at intake |
| Payout | passes `fromAccountType` and `toAccountType` from the row |
| Currency source | `domainSpendType(ctx.features)` at the router — never client input |
| Spend surfaces | offer the domain's currency alone, via `useAvailableBuzz` |
| Paid access | pays in kind |

## The migration is hand-applied, and there is deliberately no backfill

`packages/civitai-db-schema/prisma/migrations/20260814020000_placement_spend_type/` — **apply before deploying.** The column is nullable with no default, so an old pod writing a row without it stays valid.

**Do not backfill `spendType` from the ledger.** Legacy placements were *held* as yellow whatever was spent — the conversion happened on the way **in** — so the escrow genuinely contains yellow for them. Backfilling the true spend type would pay out Buzz the escrow never received. `settledSpendType` makes NULL mean yellow, and the reasoning is written into the migration so the next person doesn't "fix" it.

That's 8 prod rows, 7 already settled.

## Deliberately not fixed here

`paidAccessPayoutAccount` survives as `legacyPaidAccessPayoutAccount` for its one remaining reader: the unpublish-refund guard. It sizes a refund by the account reversing a purchase would debit, and the multi-transaction listing reports each leg by the account the **buyer** spent from — it carries no destination, so the guard cannot learn where the credit landed. That makes it correct for pre-cutover purchases and increasingly wrong for post-cutover green ones. It needs the same treatment as `Placement.spendType`: record the payout account per purchase. Flagged, not folded in.

## Tests

Written so a revert fails by naming the wrong value:

- hold on green asserts `fromAccountTypes: ['green']` **and** `toAccountType: 'green'`
- `fromAccountTypes` has length 1 — the mixed-funding guard
- `blue` rejects with "not a placement spend type" and **no Buzz call is made**
- the row records the currency in the same statement as the deadline
- approval pays the owner `toAccountType: 'green'` from the escrow's green
- **a legacy row with `spendType: null` settles yellow** — this is what protects the no-backfill decision
- `domainSpendType` equals the generator's own currency list, so the button and the escrow can't diverge

`pnpm run typecheck` 0 errors; eslint 0 errors on changed files; unit suite green for every file this touches (`blocks.router.workflow.test.ts` collected **347** tests, so the run is real).

⚠️ Eight unit files fail on this branch and **none are from this change** — they reference nothing it touches, and two CI ratchets (`db:check-generated`, `typecheck-tests-gate`) are independently red on `main` right now. Regenerating `models.ts` on an untouched `origin/main` rewrites 604 lines, which is why that file is excluded here rather than dragged into a money change.

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260814020000_placement_spend_type/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260814020000_placement_spend_type/migration.sql
@@ -3,30 +3,36 @@
 -- APPLY BEFORE deploying the app that writes this column. The column is nullable
 -- with no default, so an old pod writing a row without it stays valid.
 --
--- THERE IS DELIBERATELY NO BACKFILL, and that is the load-bearing decision here.
+-- THERE IS DELIBERATELY NO BACKFILL.
 --
--- Every placement made before this column existed was HELD as yellow whatever
--- the placer spent: the escrow debit named no destination account, so the Buzz
--- service credited the escrow account its default, and green was converted to
--- yellow on the way IN. Verified in ClickHouse `buzzTransactions` — the hold
--- legs read `fromAccountType: green, toAccountType: yellow`.
+-- Every placement made before this column existed was BOOKED into escrow as
+-- yellow whatever the placer spent: the escrow debit named no destination
+-- account, so the Buzz service credited its default. Verified in ClickHouse
+-- `buzzTransactions` — those hold legs read `fromAccountType: green,
+-- toAccountType: yellow`.
 --
--- So for an existing row the escrow genuinely contains YELLOW, whatever the
--- placer originally spent. Backfilling the true spend type from the ledger would
--- make settlement pay out green the escrow never received. NULL must mean
--- "legacy: held as yellow, settle as yellow", and the application reads it that
--- way.
+-- NULL therefore means "legacy: booked as yellow, settle as yellow", which is
+-- what the application reads it as.
 --
--- At the time of writing that is 8 production rows, 7 already settled and 1
--- pending, so the cost of leaving them alone is one placement settling in the
--- currency it is actually funded in.
+-- Be clear about what kind of decision this is: a CHOICE, not a constraint. The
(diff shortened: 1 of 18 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3911.civitaic.com` — no commit needed.

PR #3913, opened 2026-08-14
feat(image-delivery): return the media type from the internal lookup

ZacxDev:
## Why

The internal image-delivery lookup (`GET /api/internal/image-delivery/{imageKey}`) returned a single field, `hideMeta`. The image delivery service calls it on every request, so with only that boolean it has no way to tell a **video** from an **image** — everything looks like an image. Videos therefore get handed to an image-only conversion path that cannot process them, and they fail.

This PR widens that response so the caller can tell the two apart. It changes nothing about how any media is converted or served here; that is the consumer's side.

## Wire contract (stated verbatim so it can be checked against the consumer without reading this diff)

Successful (`200`) response body:

```json
{
  "id": 77,
  "url": "vid987/clip.mp4",
  "hideMeta": true,
  "type": "video",
  "mimeType": "video/mp4"
}
```

- **Key: `type`** — the media-type discriminator. This key name is load-bearing (see below).
- **Value: the bare media type**, one of `"image"`, `"video"`, `"audio"`. Not a full mime string.
- **Key: `mimeType`** — the exact container, e.g. `"video/mp4"`, `"image/jpeg"`, or `null`.

`400` and `404` bodies are unchanged (`{ "error": ... }`).

There was **no pre-existing `type` key** on this response to collide with — before this change the `200` body was exactly `{ id, url, hideMeta }`.

## `type` vs `mimeType` — why both, and which to branch on

Branch on **`type`**. `Image.type` is `NOT NULL` with a default, so it is present on every row, old and new — it is a total discriminator and can never be the reason a caller falls back to "unknown".

`mimeType` is included as well because it is genuinely useful (it names the exact container, which `type` does not) and it is **free** — it is another column on a row that was already being fetched. But it is `NULLABLE` and older rows predate it, so it must never be used as the fallback discriminator; that is stated in the type's doc comment so the next reader does not reach for it.

## Null / absent handling

`mimeType` is normalized to an explicit `null`. Two spellings are deliberately collapsed:

- **`undefined`** — `JSON.stringify` *drops* an undefined value, so the key would vanish from the body entirely and a caller could not distinguish "this image has no mime type" from "this deployment does not send the field at all".
- **`''` / whitespace** — a *present* value that is not a real mime type, i.e. exactly the kind of thing a caller mistakes for data.

So the contract is: **a non-empty string, or `null`** — the key is always present.

## Backward compatibility

Purely additive. `id`, `url` and `hideMeta` keep their names, their types and their values; nothing is renamed, reordered or reshaped, and the two new keys are appended. A caller that reads only the boolean is unaffected. There is a test pinning exactly this (`is additive — the response still carries every field it carried before`).

## No extra database round-trip

`type` and `mimeType` are two additional columns in the `SELECT` list of the **existing** single-row `WHERE url = $1` lookup — the same statement that was already fetching `hideMeta`, on both the replica path and the primary-fallback path. No second query, no second round-trip. A test asserts one query total and that the columns come from that one statement.

No schema change and no migration: these columns already exist.

## The cached-response window (why `packages/civitai-redis` is in this diff)

This response is cached (that cache is pre-existing, not added here). Entries written by the **previous** release hold only `{ id, url, hideMeta }`. If those stayed readable, then for up to the cache TTL after release a cache hit would serve a video with **no `type` at all** — and the caller cannot tell that apart from a row whose media type is genuinely unknown, so it would fall back to today's broken behaviour and the fix would appear not to work for the first few minutes.

Bumping the cache key prefix to `:v2` (a one-line change plus its comment, and the reason that file is touched) removes the window entirely: the widened shape is correct from the very first request. The cold period costs at most one single-row indexed lookup per url per TTL. There is a dedicated test that seeds the *old* key with an old-shape value and proves it is never served.

## A note on the test-only helper

`src/test-utils/queryRawProjection.ts` exists because the obvious `$queryRaw` fake — `mockResolvedValue([fixtureRow])` — returns the **whole fixture row regardless of what the statement selected**, which a real database never does. Measured on this change: **five** of the media-type assertions passed against unmodified `main`, which selects neither column, purely because the fixture carried the fields. The helper narrows each row to the statement's actual `SELECT` list, which is what makes the assertions below genuinely depend on the production query. It never falls back to pass-through — an unparseable statement or a fixture missing a selected column throws by name. It has its own tests, including the negative control that it really does drop an unselected column.

## Test matrix

Run with `pnpm test:unit:run` (counts read off the pass/fail lines, not an exit code).

| Suite | At `origin/main` | At HEAD |
|---|---|---|
| `image-delivery-metadata-cache.test.ts` | 19 failed / 5 passed (24) | **24 passed** |
| `image-delivery.test.ts` (route) | 3 failed / 3 passed (6) | **6 passed** |
| `queryRawProjection.test.ts` (helper) | 12 passed | **12 passed** |
| **Total** | **22 failed / 20 passed (42)** | **45 passed (45)** |

The base run is the new tests executed against unmodified `origin/main` production code. Representative failure, at the assertion itself:

```
FAIL  ... media type > returns type "video" and the video mimeType for a video row
AssertionError: expected undefined to be 'video' // Object.is equality
```

The helper's own 12 tests pass at base, correctly — it is a pure function and does not depend on the change.

### Mutation checks

Each mutation applied alone to otherwise-final code, and the failure message checked to be *this* assertion's rather than a compile error or a neighbouring test's:

| # | Mutation | Result |
|---|---|---|
| M1 | Drop `type`, `"mimeType"` from both `SELECT` lists | **20 failed** — `expected undefined to be 'video'` |
| M2 | Bypass `normalizeMimeType` (return the raw column) | **exactly 2 failed** — `expected undefined to be null`, `expected '   ' to be null` |
| M3 | Revert the cache key prefix to the un-bumped value | **6 failed**, incl. the deploy-window test — `expected "vi.fn()" to be called 1 times, but got 0 times` (i.e. the stale 3-field entry *was* served and the DB never consulted) |
| M4 | Rename the response key `type` → `mediaType` | **5 failed** — `expected [ 'id', 'url', 'hideMeta', …(2) ] to include 'type'` |

M4 is the important one: because the consumer treats an absent value as "unknown" and falls back, a renamed key would fail *nowhere* and silently make this fix inert. That is why the literal key name is asserted on the serialized body.

### Labelled invariant guards (not regression coverage)

Two route tests are marked `[invariant]` in the source — `still 404s an unknown url` and `still 400s a request with no id`. They pass on unmodified code and are there to pin the error shape this widening must not disturb. They are not counted as evidence the change works.

### Other checks

- `pnpm typecheck`: **4278 errors at `origin/main`, 4278 at HEAD** — zero delta, and none in any file touched here. (The absolute number is a pre-existing condition of the tree, unrelated to this change.)
- `eslint` on every changed file: **0 errors**, 11 warnings, all `no-explicit-any` in the route test, matching the existing convention in `src/tests/api/internal/ping.test.ts`.
- Adjacent suites `update-post-image-hidemeta-bust.test.ts` and `pgDbMock.parity.test.ts` fail **identically at base and at HEAD** — pre-existing, not introduced here.

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -2182,14 +2182,20 @@ const REDIS_KEYS_UNPREFIXED = {
     // only (bid submission re-validates the true minimum server-side). See
     // `getAllAuctions` in auction.service.
     ACTIVE_AUCTIONS: 'packed:caches:active-auctions',
-    // url -> {id, url, hideMeta} lookup backing the internal image-delivery endpoint
-    // (`/api/internal/image-delivery/[id]`, ~9.2 req/s at peak). The near-immutable
+    // url -> {id, url, hideMeta, type, mimeType} lookup backing the internal image-delivery
+    // endpoint (`/api/internal/image-delivery/[id]`, ~9.2 req/s at peak). The near-immutable
     // `Image WHERE url = $1` single-row read is the highest-volume DB query in the
     // profile. Keyed by the EXACT url (case/whitespace-sensitive — the WHERE key is not
     // normalized), positive results only (an unknown url stays uncached so a newly
     // registered image resolves immediately), busted when `hideMeta` flips in
     // updatePostImage. See `getCachedImageDeliveryMetadata` in image-delivery.service.
-    IMAGE_DELIVERY_METADATA: 'packed:caches:image-delivery-metadata',
+    // `:v2` — the cached value gained `type`/`mimeType`. Entries packed by the previous
+    // release hold only `{id, url, hideMeta}`, and a hit on one would serve a response with
+    // the media-type fields MISSING (indistinguishable, to a caller, from "this image has
+    // no mimeType") for up to the TTL after a release. A new key prefix makes the widened
+    // shape correct from the first request instead; the cold window costs at most one
+    // single-row indexed lookup per url per TTL.
+    IMAGE_DELIVERY_METADATA: 'packed:caches:image-delivery-metadata:v2',
     // Per-user `id -> isValidCreatorMember(boolean)` for the read-time metric-privacy /
     // donation-goal hide gate (#3266). Near-static per user; both TRUE and FALSE are
     // cached (the resolver is a total function over the id). Busted on any subscription
--- a/src/pages/api/internal/image-delivery/[id].ts
+++ b/src/pages/api/internal/image-delivery/[id].ts
@@ -23,7 +23,9 @@ export default WebhookEndpoint(async function handler(req: NextApiRequest, res: 
   const { id } = results.data;
 
   // `id` is the image url (the raw query keys on `WHERE url = $1`). Read-through Redis cache
-  // fronts the near-immutable url -> {id, url, hideMeta} lookup; fails open to the DB.
+  // fronts the near-immutable url -> {id, url, hideMeta, type, mimeType} lookup; fails open
+  // to the DB. `type`/`mimeType` let the caller tell a video from an image before it picks a
+  // delivery path; `hideMeta` is unchanged, so a caller that reads only it is unaffected.
   const image = await getCachedImageDeliveryMetadata(id);
 
   if (!image) {
--- a/src/server/services/__tests__/image-delivery-metadata-cache.test.ts
+++ b/src/server/services/__tests__/image-delivery-metadata-cache.test.ts
@@ -35,12 +35,32 @@ import {
   bustImageDeliveryMetadataCache,
 } from '~/server/services/image-delivery.service';
 import { CacheTTL } from '~/server/common/constants';
+// The DB fake narrows each fixture row to the columns the statement actually SELECTs, the way
+// a real database does. Without it, `mockResolvedValue([row])` hands back the whole fixture no
+// matter what was asked for, and every assertion about a NEWLY-selected column passes on code
+// that never selects it — measured: 5 of the media-type assertions below went green against
+// the pre-change service until this was introduced.
+import { respondWithRows } from '~/test-utils/queryRawProjection';
 
-const KEY_PREFIX = 'packed:caches:image-delivery-metadata';
+// `:v2` — bumped when the cached value gained `type`/`mimeType`, so a hit on an entry packed
+// by the previous release can never serve a response missing those fields.
+const KEY_PREFIX = 'packed:caches:image-delivery-metadata:v2';
 
-// A row exactly as the raw `Image WHERE url = $1` query returns it.
+// A row exactly as the raw `Image WHERE url = $1` query returns it. Every field holds a
+// DISTINCT value so an assertion cannot pass by one field aliasing another.
 const URL = 'abc123/def456.jpeg';
-const IMAGE_ROW = { id: 42, url: URL, hideMeta: false };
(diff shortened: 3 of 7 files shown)


ZacxDev:
## Preview Environment ✅

| | |
|---|---|
| **URL** | https://pr-3913.civitaic.com |
| **Commit** | `8606c67` |
| **Network** | ✅ HTTP 302 (0ms) |
| **Pipeline** | [View build](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3913-kmw42) |
| **Updated** | 2026-08-14 04:44:46 UTC |

ZacxDev:
**📦 Bundle budget — First Load JS (brotli):** ⚠️ over budget — shared 600.9 kB · total 12.12 MB (255 routes)

**✅ no per-route First Load JS regression vs `main` (`162fc97`).**

<details><summary>Per-route breakdown · commit <code>8606c67</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3913-kmw42">build</a> · 2026-08-14 04:41:35 UTC</summary>

```
Shared by all pages :  600.9 kB  (39 files)  [budget 470.0 kB]  OVER
Routes analysed     : 255

  1002.8 kB  /articles/[id]/[[...slug]]                 [budget 1.30 MB] OK
   992.9 kB  /user/[username]/posts                     [budget 1.30 MB] OK
   992.7 kB  /models/[id]/model-versions/[versionId]/wizard [budget 1.30 MB] OK
   992.7 kB  /models/[id]/model-versions/create         [budget 1.30 MB] OK
   992.0 kB  /user/[username]/collections               [budget 1.30 MB] OK
   989.7 kB  /challenges/[id]/[[...slug]]               [budget 1.30 MB] OK

1 budget breach(es):
  ✗ shared 600.9 kB > 470.0 kB
```

</details>

ZacxDev:

| | |
|---|---|
| **Commit** | `8606c67` |
| **Pipeline** | [View run](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3913-kmw42) |
| **Updated** | 2026-08-14 04:44 UTC |

ZacxDev:
**🔦 Lighthouse:** perf 61 · LCP 2.7s · CLS 0.228 · a11y 89 — [report ↗](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786682967002-56311.report.html)

<details><summary>Per-route scores (median of 5 runs · desktop · authed as ci-smoke-gold) · commit <code>8606c67</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3913-kmw42">build</a> · 2026-08-14 04:49 UTC</summary>

| `/` | 61 | 89 | 2.7s | 0.228 | 209ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786682967002-56311.report.html) |
| `/models` | 56 | 93 | 2.9s | 0.261 | 302ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786682967957-3684.report.html) |
| `/generate` | 68 | 97 | 3.7s | 0.022 | 246ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786682968408-94315.report.html) |
| `/models/4201` | 98 | 94 | 0.9s | 0.000 | 0ms | — | — |

</details>

PR #3914, opened 2026-08-14
fix(placement): domain-correct review queues, honest paging, and a price-cap band

JustMaier:
Two review queues — remix gallery and stickers — were sending assets a SFW domain may not serve, paging in a way that under-reported what was waiting, and counting a population the list could not render. This fixes all three, plus the review UI that has to explain what is being withheld.

## Why the payload, not the component

`ImageGuard2`'s blur is built from the viewer's own settings and **never reads the domain ceiling**. Green is normally safe because above-SFW rows never reach the client — the queries clamp. These review queues deliberately send no browsing level, because an owner has to see everything waiting on them, which made them the one path handing an above-ceiling asset to a SFW client.

So the fix is in what is sent. Above-ceiling rows now come back as `{ viewable: false, id, nsfwLevel }` — no url, no metadata, nothing that resolves to the asset. A discriminated union rather than a nullable `url`, so a caller that forgets the check does not compile.

The viewer's **own** band is marked, not filtered. The feed drops what is outside your band because nobody has to act on it; a queue cannot — a submission hidden that way still expires, costing the owner their fee and the submitter nothing.

## Paging

The remix queue fetched a page and *then* discarded unrenderable rows, so a queue of 66 whose first 50 were unrenderable returned 4 items **and** a cursor — "4+" over a page size of 50, and "load more" produced 2. Selection now happens in SQL, so a page is `limit` renderable rows or the end of the queue.

## Badge vs list

`pendingCount` was a bare count with no image join while the list required published + Scanned + not tos/minor/poi — two populations, so the badge could read 2 over a list that rendered none. One shared SQL fragment now serves both.

## The sticker queue deliberately differs

It gets the domain ceiling but **not** the visibility filter or SQL-side paging. Its `targetId` is the owner's *own* upload, so dropping rows would take the approve/decline with it and expire the escrow unanswered. With no filter there is no fetch-then-filter gap, and its badge reads the same query as its list. Stated here so it doesn't read as an oversight next to the remix queue.

## Review UI

Three states, each with the only answer that helps it:

- **Above the domain ceiling** — decline here, or open it on the domain that serves it. No widen control, because no setting on this domain can produce an asset that was never sent. Note there is **no per-item click-through to out-of-band content anywhere in the product**: such content is filtered server-side and dropped again client-side, and `ImageGuard2`'s reveal only un-blurs items already inside your band. Inventing a per-item override here would have been a new pattern on a surface where approval is irreversible for a week.
- **Outside your content settings** — the existing `BrowsingModeMenu`, gated on `canViewNsfw` so green is never offered settings it does not have.
- **Above the gallery's own accepted band** — the owner can decline every out-of-band submission at once. The set is resolved server-side from (hostImageId, the owner's rule); accepting placement ids would let a caller decline rows that are perfectly in band. It loops the existing single-decline path rather than growing a bulk settlement, per `actOnStickerPlacements`, counts failures instead of throwing so one bad payout leg cannot strand the rest, and returns `{ considered, settled }` so the UI reports "declined 2 of 3" rather than a success it did not have. Unrated rows are left alone: charging a decline fee for an image with no rating answers a question nobody asked.

All three share one alert rather than three stacked ones — they overlap constantly.

## Also here

- The second **price-cap band** moves 5,000 → 10,000 creator score. 5,000 was reachable on a throwaway account with one model and a handful of images. This aligns the source with the live `placement:config` KeyValue row, which already reads 10,000 — so it changes no behaviour in production.
- The **price caption** is cut to the amount in every branch (an over-cap price quotes the cap, since that is what a placer is charged), so it stops colliding with the slider's own min/max marks. Colour carries the warning.
- The **Sent tab** now honours the viewer's content settings, matching Received — the same image was banded on one tab and not the other.
- The **submit modal** gains a hierarchy, and its footer note can wrap: it previously squeezed the Submit button until the label clipped.

## Testing

Mutation-tested, not assumed. Removing the domain ceiling fails named tests on both surfaces.

That discipline was earned during this branch. An adversarial review found the first version had **no test for the ceiling at all** — deleting it outright left the suite green at 82 passing. It also found that a first attempt at a stronger SQL assertion still passed against an `AND` → `OR` flip, because a composed `Prisma.sql` puts the joining operator in the outer template and the fragment bodies in the interpolated values, so concatenating strings-then-values reads as plausible SQL and is not.

## Known, and not fixed here

- `db:check-generated` is red on `main` independently of this branch — `models.ts` regenerates ~604 lines differently on a clean checkout. Reproduced in a separate worktree; deliberately left alone.
- `typecheck-tests` reports `sticker-placement.service.test.ts` at 41 against a baseline of 16. **That file is in this diff** — this branch adds 103 lines of domain-ceiling tests to it — and **none of the 41 errors fall in them**: every error is at or below line 910 except one at 1216, our additions begin at 1148, and that 1216 blames to `f8a2e5f125` rather than to this branch. So these are errors this branch neither caused nor fixed. Note what that does *not* claim: nobody has checked the file out at `origin/main` to measure it there. Being handled separately by elise in #3918, which takes it to 0.

  `remix-gallery.service.test.ts` is at **28 against a baseline of 29 on this branch** — it is *not* fixed on `main`, where elise measured 39. That improvement lands only when this PR merges, so this branch is doing real work against the gate rather than reporting a state that already exists.
- 16 unit tests fail on this branch, all in repo-scanning drift guards (`blocks/`, `comics/`, `middleware/`, `scripts/`) that name no file in this diff.

### DIFF
--- a/src/components/Account/PlacementSpaceSection.tsx
+++ b/src/components/Account/PlacementSpaceSection.tsx
@@ -199,11 +199,10 @@ export function PlacementSpaceSection() {
             commit(mode, value);
           }}
         />
-        {/* No negative margin lifting this onto the marks' row: the marks are
-            pinned left and right and this is centred, so a caption that wrapped
-            collided with both. Same fix as the gallery's. */}
+        {/* On the marks' row, which only holds while the caption is one short
+            line — see `placementPriceCaption`, which is bounded to the amount. */}
         {caption && (
-          <Text size="xs" ta="center" c={caption.warning ? 'yellow' : 'dimmed'}>
+          <Text size="xs" ta="center" mt={-22} c={caption.warning ? 'yellow' : 'dimmed'}>
             {caption.text}
           </Text>
         )}
--- a/src/components/Account/PlacementSpaceSection.tsx
+++ b/src/components/Account/PlacementSpaceSection.tsx
@@ -199,8 +199,11 @@ export function PlacementSpaceSection() {
             commit(mode, value);
           }}
         />
+        {/* No negative margin lifting this onto the marks' row: the marks are
+            pinned left and right and this is centred, so a caption that wrapped
+            collided with both. Same fix as the gallery's. */}
         {caption && (
-          <Text size="xs" ta="center" mt={-22} c={caption.warning ? 'yellow' : 'dimmed'}>
+          <Text size="xs" ta="center" c={caption.warning ? 'yellow' : 'dimmed'}>
             {caption.text}
           </Text>
         )}
--- a/src/components/RemixGallery/RemixGalleryCard.tsx
+++ b/src/components/RemixGallery/RemixGalleryCard.tsx
@@ -30,7 +30,7 @@ import { LoginRedirect } from '~/components/LoginRedirect/LoginRedirect';
 import { RemixGalleryExplainer } from '~/components/RemixGallery/RemixGalleryExplainer';
 import { RemixGalleryManageModal } from '~/components/RemixGallery/RemixGalleryManageModal';
 import { RemixGallerySubmitModal } from '~/components/RemixGallery/RemixGallerySubmitModal';
-import { SubmissionThumb } from '~/components/RemixGallery/SubmissionThumb';
+import { QueueThumb } from '~/components/RemixGallery/SubmissionPair';
 import {
   dedupeGalleryItems,
   galleryDialogImages,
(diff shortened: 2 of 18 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3914.civitaic.com` — no commit needed.

PR #3915, opened 2026-08-14
fix(tests): clear the 10 test files that entered the ratchet unbaselined

JustMaier:
## What

`scripts/ci/typecheck-tests-gate.mjs` blocks on any file under `src/**/__tests__/` that has type errors and **no baseline entry**. Ten files merged after the baseline was captured in #3868 and arrived carrying **33** such errors between them. This clears all 33.

`pnpm typecheck` cannot see any of it — the root tsconfig excludes `src/**/__tests__/**`, which is the whole reason the gate exists. Reproduce with:

```bash
pnpm exec tsc -p tsconfig.tests.json --noEmit
```

**The baseline JSON is untouched.** Nothing here is bought by re-baselining, and no fix widens a mock's return, casts to `any`, or loosens an assertion.

| file | was | now |
|---|---|---|
| `Challenge/__tests__/challenge-messages.test.ts` | 1 | 0 |
| `clickhouse/__tests__/tracker.awaitDelivery.test.ts` | 2 | 0 |
| `common/__tests__/serialized-file-download-url.test.ts` | 2 | 0 |
| `controllers/__tests__/image.controller.feed-source.test.ts` | 4 | 0 |
| `daily-challenge/__tests__/challenge-engine-rolling-swiss.test.ts` | 2 | 0 |
| `daily-challenge/__tests__/challenge-pairwise-store.test.ts` | 3 | 0 |
| `daily-challenge/__tests__/challenge-winner-mapping.test.ts` | 1 | 0 |
| `notifications/__tests__/app-collaborator.notifications.test.ts` | 9 | 0 |
| `services/__tests__/file-download-lookup.test.ts` | 6 | 0 |
| `services/__tests__/image-cacher-invalidate-scope.test.ts` | 3 | 0 |

Gate overall: **869 errors across 151 files -> 836 across 141.**

## Three of these are corrections, not cleanups

Worth reading, because they change production types:

**`challengeEndedMessage` took `queued?: boolean`.** So the `@ts-expect-error` in its test — asserting that omitting `queued` is a compile error — was unused (TS2578). The contract the test names had quietly stopped being enforced. `EndChallengeResult` in `challenge.service.ts` declares `queued: boolean` required, with a comment stating that optional would let a return site omit it and render the inline "0 winner(s) selected" message for a challenge whose judging was merely *queued*. The helper now requires it too; both call sites already pass a full `EndChallengeResult`.

**`GetAllImagesIndexResult` was `AsyncReturnType<typeof getAllImages>`**, which has no `source` — while `getAllImagesIndex` returns `...(searchSource && { source: searchSource })` and the controller stamps `source: 'db'` on the other branch *specifically* so a client can tell a BitDex page from a database page. The field was invisible to every caller. Added as optional, since the blocked-browsing early return legitimately omits it.

**`toApiModelFile<T extends { metadata?: unknown }>` has a weak-type constraint.** An argument naming none of its members fails inference, falls back to the constraint itself, and then reports `id` as an excess property while dropping `modelVersionId` from the result — which is the exact field that test file exists to protect. The call exercising "no metadata blob" now spells out its type argument.

## The rest

- `vi.fn(async () => …)` infers a **zero-arg** signature, so `mock.calls` is the empty tuple `[]` and every `calls[0][0]` is a type error. Fixed with `vi.fn<Sig>(impl)` rather than a placeholder parameter — the repo's `no-unused-vars` has no `argsIgnorePattern`, so `_args` warns.
- `target: ES2018` rejects `0n` / `9n` BigInt literals -> `BigInt(0)` / `BigInt(9)`.
- Property reads off an `as never` fixture (TS no longer allows them) -> the value is named once and read from there.
- `prepareMessage` returns `NotificationMessage | undefined` and was read without narrowing -> the helper throws a named error instead, so a missing message fails as itself rather than as `undefined.url`.

## Still red, and whose

The gate remains BLOCKED on **`sticker-placement.service.test.ts` (16 -> 41)** alone. That file is @alexandra's and @zuri's, is untouched here by agreement, and is tracked in #3914. `remix-gallery.service.test.ts` is at 28/29 since `82170251d0` and no longer blocks.

Roughly 19 of its 41 are not a mock-signature problem at all: `flip` and `opacity` became required members of `StickerPlacementData` in #3865 while `normalizeStickerPlacement` still defaults both at runtime, and a test at ~line 413 omits them deliberately to pin that defaulting. Filling the fixtures in would leave that test green and proving nothing, so it wants a type split rather than a fixture edit — noted here so nobody solves it the fast way.

## Verification

- `pnpm run typecheck` — **0 errors** (unfiltered).
- Lint **unchanged** on the touched set: 39 problems / 7 errors before and after, all pre-existing.
- `pnpm run test:unit:run` — failures identical with and without this diff (4 files / 14 tests in isolation, all pre-existing on this base and unrelated to these files).
- `blocks.router.workflow.test.ts` collected **347** tests, so the worktree's submodule is real and the run means something.
- `db:check-generated` deliberately not chased — it is independently red on `main`.

### DIFF
--- a/src/components/Challenge/challenge-messages.ts
+++ b/src/components/Challenge/challenge-messages.ts
@@ -9,7 +9,7 @@
  * A string inside a page component is also unreachable from any test we run; here it is a function
  * with a test, so reverting it fails something.
  */
-export function challengeEndedMessage(result: { queued?: boolean; winnersCount: number }): string {
+export function challengeEndedMessage(result: { queued: boolean; winnersCount: number }): string {
   if (result.queued) {
     return 'Challenge ended. Judging has been queued — winners will be posted when it completes.';
   }
--- a/src/server/clickhouse/__tests__/tracker.awaitDelivery.test.ts
+++ b/src/server/clickhouse/__tests__/tracker.awaitDelivery.test.ts
@@ -38,8 +38,18 @@ const metric = () => ({
   metricValue: 100,
 });
 
+/**
+ * The parameters are declared on the mock's type rather than left to inference: a
+ * bare `vi.fn(async () => …)` is a zero-arg signature, which makes `mock.calls` the
+ * empty tuple and every `calls[0][1]` below unreadable.
+ */
+type FetchStub = (
+  url: string,
+  init?: { signal?: AbortSignal }
+) => Promise<{ ok: boolean; status: number; text: () => Promise<string> }>;
+
 const respond = (status: number) =>
-  vi.fn(async () => ({ ok: status < 400, status, text: async () => '' }));
+  vi.fn<FetchStub>(async () => ({ ok: status < 400, status, text: async () => '' }));
 
 let fetchMock: ReturnType<typeof respond>;
 
@@ -81,9 +91,7 @@ describe('awaitDelivery', () => {
 
     // Nothing configures an undici dispatcher, so an unsignalled fetch inherits
     // a 300s header timeout — three of those on one row is a wedge, not a retry.
-    const signals = fetchMock.mock.calls.map(
-      (call) => (call as [string, { signal?: AbortSignal }])[1].signal
-    );
+    const signals = fetchMock.mock.calls.map((call) => call[1]?.signal);
     expect(signals.every((signal) => signal instanceof AbortSignal)).toBe(true);
     // One signal hoisted out of the loop would be worse than none: `fetch`
     // rejects immediately on an already-aborted signal, so the first slow
@@ -99,8 +107,11 @@ describe('fire and forget', () => {
     // The dispatch is not awaited, so let its first attempt reach fetch.
     await new Promise((resolve) => setTimeout(resolve, 0));
 
-    const [, init] = fetchMock.mock.calls[0] as [string, { signal?: AbortSignal }];
-    expect(init.signal).toBeUndefined();
+    const [, init] = fetchMock.mock.calls[0];
+    // Asserted present first: `init?.signal` on a missing init is also undefined,
+    // and would pass this without the request ever having been made.
+    expect(init).toBeDefined();
+    expect(init?.signal).toBeUndefined();
   });
 
   // LAST on purpose. Its retry chain outlives the test by ~750ms and `fetch` is
--- a/src/server/common/__tests__/serialized-file-download-url.test.ts
+++ b/src/server/common/__tests__/serialized-file-download-url.test.ts
@@ -198,7 +198,14 @@ describe('toApiModelFile — the owning version id survives the reduction', () =
   });
 
   it('tolerates a missing metadata blob', () => {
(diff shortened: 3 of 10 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3915.civitaic.com` — no commit needed.

JustMaier:
### Self-review pass — one finding, fixed in `4f8e9fa`

The adversarial reviewer I spawned went idle without producing a report, so I ran the checks myself rather than merge on an unverified pass. One real finding, plus what I actually confirmed.

**Finding (fixed): the `source` addition was too wide.** `GetAllImagesIndexResult` is the declared return of **two** functions — `getAllImagesIndex` and `getImagesFromFeedSearch` — and only the first ever sets a `source`. Adding the field to the shared alias would have let a caller write `if (result.source === 'bitdex')` against `getImagesFromFeedSearch`, which compiles and is always false. That is a smaller version of exactly the defect this PR is about: a type permitting a check that cannot mean anything. The optional field now sits on a separate `GetAllImagesIndexSourcedResult` used only by the path that returns it; `getImagesFromFeedSearch` keeps the narrow alias.

**Mutation-tested, both fail legibly:**
- Reverted `url: APP_INVITES_URL` to `appUrl(details)` — three named failures (`expected '/apps/ab_1' to be '/apps/invites'`, `expected '/apps' to be '/apps/invites'`, and the ids-do-not-vary control). The `msg()` helper's throw does **not** mask the regression as a crash, which was the specific vacuity risk in that change.
- Replaced the per-attempt `AbortSignal.timeout(...)` with `undefined` in `tracker.ts` — fails on the `signals.every(s => s instanceof AbortSignal)` assertion, so deriving `signals` via `call[1]?.signal` still catches a missing init rather than passing over it.

**Confirmed by reading, not assumed:**
- Every fixture value behind `as BasicFileMetadata` is a real member of its union (`src/types/global.d.ts:100-120`): `'SafeTensor'`/`'PickleTensor'`, `'pruned'`/`'full'`, `'fp16'`/`'fp32'`. The cast narrows correct literals; it does not paper over a wrong value, and an invalid one would still error.
- Both `challengeEndedMessage` call sites are `onSuccess: (data) =>` on the `endAndPickWinners` mutation, whose output is `EndChallengeResult` — `queued` is set on every return path, and `pnpm typecheck` is green with it required.

**Not independently re-verified:** the per-field casts in `toRouteInput` are stronger than the `as never` they replace but are still casts — the route's zod schema remains the only thing narrowing those values in production.

Gate unchanged by this amend: 836 across 141, my ten at zero.

One note on the numbers above: this branch is based on `162fc976`, which predates `82170251d0`, so the gate here still reports `remix-gallery.service.test.ts` at 29 -> 39. On current `origin/main` it is 28/29 and passing — **the sticker file is the only thing blocking.**

PR #3916, opened 2026-08-14
fix(post): hide/show-prompt toggle nulls the image's generation metadata

ZacxDev:
## The bug

**This is data-destructive and is happening to users right now.** Toggling **Hide prompt / Show prompt** on a post-editor image card permanently nulls that image's generation metadata (`Image.meta`): prompt, negative prompt, sampler, seed, resources — everything.

The button toggles, so **both directions destroy it**. Revealing a prompt wipes it exactly like hiding it does.

## Cause

`post.updateImage` takes an optional `meta`, and the two absent-ish cases have to stay distinct:

- `undefined` — the key was not sent → **leave the stored value alone**
- `null` — the user explicitly cleared it → **write SQL NULL**

Prisma expresses that difference as `undefined` (omit the column from the `UPDATE`) vs `Prisma.JsonNull` (write NULL).

The write used to read:

```ts
meta: image.meta !== null ? (image.meta as Prisma.JsonObject) : Prisma.JsonNull,
```

An absent `meta` is `undefined`, and `undefined !== null` is **true**, so it took the first branch and Prisma omitted the field. The column was preserved.

#3871 (server-side remix-provenance verification) routed the value through `sanitizeProvenance` first and changed the comparison to loose `!=`:

```ts
meta: meta != null ? (meta as Prisma.JsonObject) : Prisma.JsonNull,
```

Two things compose into the regression:

1. `sanitizeProvenance` returns its input unchanged when falsy (`if (!meta) return meta`), so `undefined` in stays `undefined` out.
2. Loose `!=` is true for **both** `null` and `undefined`, so `undefined` now falls into the `Prisma.JsonNull` branch.

Result: an update that never mentioned `meta` writes SQL NULL over it. No error, no client-visible failure — the row just loses its metadata.

The reachable caller is the hide/show-prompt button on the post-editor image card, which sends `{ id, hideMeta: !hideMeta }` and nothing else. The metadata editor modal (and the comics panel drawer, which opens that same modal) both send `meta`, so those paths were never affected.

This is a factual attribution, not a criticism — the strict-vs-loose distinction was doing load-bearing work that nothing in the file or the tests named.

## The fix

Derive the discriminator the schema already provides and omit the field entirely when the key was not sent:

```ts
const metaProvided = image.meta !== undefined;

const meta = metaProvided
  ? sanitizeProvenance(image.meta as …, storedSourceImageIds(currentImage.meta))
  : undefined;

// …
meta: metaProvided ? (meta != null ? (meta as Prisma.JsonObject) : Prisma.JsonNull) : undefined,
```

Sanitization is skipped when nothing is written: there is no client-authored meta to sanitize, and the stored row is not being rewritten. #3871's guarantee is untouched on every path that *does* write `meta` — a client still cannot assert a derivation by editing its own image.

Deliberately minimal; this is a hotfix for live data loss.

## Tests

The existing `update-post-image-hidemeta-bust` suite covered the cache-bust wiring and asserted **nothing** about what is written to `meta` — that gap is why this shipped. It now pins the write, asserting on the `data` object actually handed to `dbWrite.image.update`:

| Input | Expected |
|---|---|
| `{ id, hideMeta: true }` | `meta` absent from the update, and specifically **not** `Prisma.JsonNull` |
| `{ id, hideMeta: false }` | same (the un-hide direction is equally destructive) |
| `{ id }` | same |
| `{ id, meta: null }` | **does** write `Prisma.JsonNull` — the explicit-clear path still works |
| `{ id, meta: {...} }` | writes the object |
| `{ id, meta: { extra: { sourceImageIds: [999] } } }` | written ids come from the stored row, never the payload |

Every assertion first requires the update call to have happened, so none can pass vacuously against a mock that was never invoked.

**Verification matrix**

- **Red at base** `baee4d418b`: 3 failed / 8 passed / **11 tests** — the three omitted-`meta` cases all receive `JsonNull {}`.
- **Green at HEAD**: 11 passed / **11 tests**.
- Related suites at HEAD: 278 passed, 0 failed (28 files; 2 fail to import for an unrelated, pre-existing reason that reproduces identically at base).
- `pnpm typecheck`: 12 errors at base, 12 at HEAD — no delta.

**Mutation table** — each mutant killed by a named test with its own message:

|---|---|---|
| `metaProvided = true` | `does NOT write meta when the input omits it (hideMeta: true / false)`, `…neither meta nor hideMeta` | `expected JsonNull{} to be undefined` (3 failures) |
| `!==` → `!=` | `DOES write Prisma.JsonNull when meta is explicitly null` | `expected undefined to be JsonNull{}` |
| drop the guard on the write line | the same three omitted-`meta` tests | `expected JsonNull{} to be undefined` (3 failures) |
| provenance re-derived from the payload instead of the stored row | `still re-derives provenance from the stored row`, `strips an unverified provenance claim` | `expected [ 999 ] to deeply equal [ 42 ]`; `expected { sourceImageIds: [ 999 ] } to not have property "sourceImageIds"` |
| vacuity control — return before the update | **all 7** new tests | `expected "vi.fn()" to be called 1 times, but got 0 times` |

The last row is the control for the vacuity hazard: with the update never called, every new test fails on the call-count assertion rather than passing on an `undefined` read.

## Possible follow-up (not in this PR)

`isValidAIGeneration({ …, meta: image.meta })` a few lines above reads the client-supplied value rather than `currentImage.meta`, so a `hideMeta`-only toggle judges verifiability against no metadata at all. That predates #3871 and is a separate behavioural question; keeping it out to hold this PR to the data-loss fix.

## Remediation

Rows already nulled cannot be recovered from the row itself — worth a separate conversation about whether any point-in-time copy can be mined for affected images.

### DIFF
--- a/src/server/services/__tests__/update-post-image-hidemeta-bust.test.ts
+++ b/src/server/services/__tests__/update-post-image-hidemeta-bust.test.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '@prisma/client';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 /**
@@ -48,7 +49,7 @@ vi.mock('@civitai/db', () => ({
   createLagTracker: vi.fn(() => ({})),
   loadDbEnv: vi.fn(() => ({})),
 }));
-vi.mock('~/server/db/pgDb', () => ({ pgDbReadLong: {},  pgDbRead: {}, pgDbWrite: {} }));
+vi.mock('~/server/db/pgDb', () => ({ pgDbReadLong: {}, pgDbRead: {}, pgDbWrite: {} }));
 vi.mock('~/server/db/db-lag-helpers', () => ({
   getDbWithoutLag: vi.fn(),
   preventReplicationLag: vi.fn(),
@@ -143,15 +144,27 @@ const KEY_URL = IMAGE_URL;
 
 // Build a currentImage row. blockedFor is anything BUT AiNotVerified so shouldIngest=false
 // (no enqueueImageIngestion side effect on this path).
-function wireCurrentImage(hideMeta: boolean) {
+function wireCurrentImage(hideMeta: boolean, meta: unknown = null) {
   mockFindUniqueOrThrow.mockResolvedValue({
     hideMeta,
     ingestion: 'Scanned',
     blockedFor: null,
     metadata: {},
     nsfwLevel: 1,
+    meta,
   });
   mockImageUpdate.mockResolvedValue({ id: IMAGE_ID, url: IMAGE_URL, userId: USER_ID });
+}
+
+/**
+ * The `data` object actually handed to `dbWrite.image.update`. Asserting the update
+ * HAPPENED first is load-bearing: `expect(data.meta).toBeUndefined()` passes vacuously
+ * against a mock that was never called, which is exactly how a "meta is untouched"
+ * claim can be green while the column is being nulled.
+ */
+function updateData() {
+  expect(mockImageUpdate).toHaveBeenCalledTimes(1);
+  return mockImageUpdate.mock.calls[0][0].data as Record<string, unknown>;
 }
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Preview Environment ✅

| | |
|---|---|
| **URL** | https://pr-3916.civitaic.com |
| **Commit** | `f3d9fd5` |
| **Network** | ✅ HTTP 302 (8ms) |
| **Pipeline** | [View build](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3916-78nfw) |
| **Updated** | 2026-08-14 05:30:47 UTC |

ZacxDev:
**📦 Bundle budget — First Load JS (brotli):** ⚠️ over budget — shared 600.9 kB · total 12.12 MB (255 routes)

**✅ no per-route First Load JS regression vs `main` (`4cdd0f8`).**

<details><summary>Per-route breakdown · commit <code>f3d9fd5</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3916-78nfw">build</a> · 2026-08-14 05:30:47 UTC</summary>

```
Shared by all pages :  600.9 kB  (39 files)  [budget 470.0 kB]  OVER
Routes analysed     : 255

  1002.8 kB  /articles/[id]/[[...slug]]                 [budget 1.30 MB] OK
   992.9 kB  /user/[username]/posts                     [budget 1.30 MB] OK
   992.7 kB  /models/[id]/model-versions/[versionId]/wizard [budget 1.30 MB] OK
   992.7 kB  /models/[id]/model-versions/create         [budget 1.30 MB] OK
   992.0 kB  /user/[username]/collections               [budget 1.30 MB] OK
   989.8 kB  /challenges/[id]/[[...slug]]               [budget 1.30 MB] OK

1 budget breach(es):
  ✗ shared 600.9 kB > 470.0 kB
```

</details>

ZacxDev:

| | |
|---|---|
| **Commit** | `f3d9fd5` |
| **Pipeline** | [View run](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3916-78nfw) |
| **Updated** | 2026-08-14 05:33 UTC |

ZacxDev:
**🔦 Lighthouse:** perf 39 · LCP 3.1s · CLS 0.235 · a11y 92 — [report ↗](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786685917174-41716.report.html)

<details><summary>Per-route scores (median of 5 runs · desktop · authed as ci-smoke-gold) · commit <code>f3d9fd5</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3916-78nfw">build</a> · 2026-08-14 05:38 UTC</summary>

| `/` | 39 | 92 | 3.1s | 0.235 | 633ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786685917174-41716.report.html) |
| `/models` | 69 | 93 | 2.8s | 0.261 | 105ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786685917864-61680.report.html) |
| `/generate` | 73 | 97 | 3.4s | 0.046 | 188ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786685918578-84342.report.html) |

</details>

PR #3917, opened 2026-08-14
fix(paid-access): pay the seller in the currency the buyer spent

JustMaier:
Sibling of #3911. That one fixed placements; this is the high-volume half.

## What was wrong

A **green** early-access purchase paid the seller **yellow**.

Not a decision. The charge named no destination account, so the buzz service applied its yellow default (`buzz.service.ts`, `toAccountType ?? 'yellow'`, added 2025-07-03) and converted on the way through. **Every green paid-access purchase since 2025-10-30 landed this way — 2,409 legs**, ~1,986 legs / 5.3M Buzz in the last quarter alone.

This predates the 2026-08-06 paid-access revamp by ~10 months; that change only added the blue carve-out. Monthly green legs: Oct 5 → Apr 319 → May 600 → Jul 549 → Aug 483. The step change is green adoption, not code.

## Evidence

Confirmed against the buzz service itself, not just ClickHouse, and attacked by a reviewer whose only job was to refute it:

- A seller's **yellow** ledger contains their green-funded sale; their **green** account for that day holds one unrelated 1-Buzz fee.
- Green *is* a routine credit destination elsewhere — 222,730 legs since July — so this is a real absence, not a missing capability.
- All 1,982 sampled legs are `type = 'purchase'`; no refunds inflating the count.
- 695/695 model versions behind them exist in prod. No dev contamination.

## The change

One line: the charge names the buyer's own account. Blue already did this and is untouched.

## Known and accepted: the unpublish-refund guard

`paidAccessPayoutAccount` still maps green → yellow for its one remaining reader — the guard that fires when a **creator unpublishes a model with live early-access buyers** and must refund them. It sizes that refund by the account a reversal would debit, and it cannot learn that from the ledger: the multi-transaction listing reports each leg by the account the **buyer** spent from and carries no destination.

So from this deploy it is correct for older purchases and increasingly wrong for new green ones:

- a seller may be told they cannot afford a refund they can afford, or
- the reversal may take their green balance negative.

**Both accepted by Justin on 2026-08-14**, rather than block the currency fix behind a schema change. The refund itself always moves the right money — it reverses the original legs; only the pre-check is approximate.

Not affected: **moderator takedowns skip the guard entirely** (`isModerator`), and buyers whose access window already lapsed never block an unpublish.

The real fix is to record the payout account per purchase, the way `Placement.spendType` does in #3911. At that point this function has no readers and can go.

## Sequencing with #3911

Independent — no shared files, no stacking, either can merge first. Both touch `buzz-helpers.ts`, so whichever lands second may need a trivial comment merge. #3911 carries a migration and a **flags-off-during-canary** requirement; **this PR has neither** — no schema change, and nothing to drain.

## Verification

`pnpm run typecheck` 0 errors. 34/34 across the paid-access, refund-guard and buzz-helpers suites — including `model-early-access-refund.service.test.ts`, which passes **unchanged**, confirming the guard's behaviour is deliberately untouched.

The test that asserted the old behaviour by name (`pays a %s purchase into yellow`, with a comment reasoning that green "has no payout account of its own") now asserts payment in kind.


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3917.civitaic.com` — no commit needed.

PR #3918, opened 2026-08-14
fix(placement): split a sticker placement's INPUT shape from its stored shape

JustMaier:
Clears `sticker-placement.service.test.ts` — **41 errors against a ratchet baseline of 16, now 0** — without editing a single fixture. Baseline JSON untouched.

`pnpm typecheck` cannot see any of this; the root tsconfig excludes `src/**/__tests__/**`. Reproduce with `pnpm exec tsc -p tsconfig.tests.json --noEmit`.

## Nineteen of the 41 were a bug in the types, not in the tests

#3865 added `flip` and `opacity` to `StickerPlacementData` as **required**. That type describes a *stored row*, and required is correct there. But the same type was also the parameter of `normalizeStickerPlacement` — whose body exists precisely because callers arrive without both keys:

```ts
flip: data.flip === true,
opacity: clamp(Number.isFinite(data.opacity) ? data.opacity : DEFAULT, …)
```

Under the old signature **those two branches were unreachable as far as the compiler was concerned**, while staying necessary at runtime. #3865's own commit message says why they're needed:

> Stored rows written before this have neither key. They default to full strength and unmirrored in `parseStickerPlacementData`, which is now the one way to read `Placement.data`, rather than in each reader.

The defaulting was put in the read path knowingly. The write path never got the same treatment, and its signature ended up claiming input equals output for a function that exists because they differ.

So `StickerPlacementInput` is what a caller may hand **in** (both optional), and `StickerPlacementData` is unchanged for anything reading a **placed row**. `normalizeStickerPlacement` takes the first and returns the second — which is what it always did.

**Which side stayed required matters, so stating it plainly:** `parseStickerPlacementData` and `PlacedSticker` keep both required. They are the narrowing half. This is not "the type got loosened" — the input and the stored shape were one type doing two jobs, and only the input job moved.

Nothing on the real call path changes: `stickerPlacementDataSchema` defaults both (`z.boolean().default(false)`, `.default(STICKER_PLACEMENT_DEFAULT_OPACITY)`), so the router's parsed value always carries them.

**The alternative would have been worse than doing nothing.** Filling `flip`/`opacity` into every fixture turns the gate green and deletes the only test pinning the defaulting — the case at ~line 413 omits both deliberately and asserts the row comes back `flip: false, opacity: 1`. It would still have passed, and proved nothing.

## The other ~20

`vi.fn(async () => …)` infers a **zero-arg** signature, so `mock.calls` is the empty tuple `[]` and every `calls[0][0]` — how this file asserts the WHERE clauses its guards are made of — is a type error against its own mock. Declared with `vi.fn<Sig>(impl)` rather than a placeholder parameter (the repo's `no-unused-vars` has no `argsIgnorePattern`). The per-site casts stay; they are what pins each shape.

## Verification

- `pnpm run typecheck` — **0 errors**, unfiltered.
- Gate reports this file **CLEAN** (`1 file(s) now clean`), baseline entry of 16 left in place.
- **113 tests pass** across `sticker-placement.service.test.ts`, `sticker-placement-opacity.test.ts`, `placement.test.ts`.
- eslint clean on all three changed files.
- **Mutation-tested**, because a split like this is exactly how a test goes vacuous: replacing the opacity defaulting with `data.opacity as number` fails the ~413 case with `- "opacity": 1 / + "opacity": NaN`. It still proves what it was written to prove.

## The invariant this rests on, traced rather than argued

Making `flip`/`opacity` optional on the INPUT is only safe if nothing writes input-shaped data straight to `Placement.data`. If anything did, a row would land without them and the "stored rows always have both" half — which `parseStickerPlacementData` and `PlacedSticker` depend on — would quietly stop being true.

Checked at the call sites, not inferred from the types:

- `sticker-placement.service.ts:227` is the **only** `placement.create` for this surface, and it spreads `...normalizeStickerPlacement(data)` — which returns the stored shape with both fields set.
- The two other `data:` spreads (`:524`, `:776`) operate on rows already read back, not on input.

So no row can be written missing either field. Credit: this trace is @zuri's — I had argued it from the type signature, which is not the same thing.

**Known and deliberate:** `placementUpdate` is `PrismaStub<object>`. Fine today because nothing reads its resolved value; it will fight the first test that does.

## Gate state on `main`, measured here

```
828 error(s) across 150 file(s)  (baseline: 801 across 141)
```

Still blocked, on two things that are **not** this PR:

- the 10 unbaselined files — fixed in **#3915**, open
- `remix-gallery.service.test.ts` at 29 -> 39 — fixed in **#3914**, open

⚠️ One correction worth recording: `remix-gallery.service.test.ts` is at 28/29 **on `fix/placement-content-levels`, not on `main`** — `82170251d0` is not an ancestor of `origin/main`. I checked, because I was about to repeat it as fact. On `main` it still measures 39.

With #3915 and #3914 both merged, this takes the gate to green.

Coordination: @alexandra and @zuri held off these files while I worked. `stickerPlacementDataSchema` is deliberately untouched, so #3914's schema and queue changes are clear of this. The input/output split was zuri's reading.

### DIFF
--- a/src/server/services/__tests__/sticker-placement.service.test.ts
+++ b/src/server/services/__tests__/sticker-placement.service.test.ts
@@ -26,7 +26,9 @@ const CAP = 700;
 const PRICE = 700;
 
 const holdPlacementEscrow = vi.fn();
-const settlePlacement = vi.fn(async () => ({ settled: true }));
+const settlePlacement = vi.fn<(args: { action: string }) => Promise<{ settled: boolean }>>(
+  async () => ({ settled: true })
+);
 vi.mock('~/server/services/placement-escrow.service', () => ({
   holdPlacementEscrow,
   settlePlacement,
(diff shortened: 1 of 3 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3918.civitaic.com` — no commit needed.

JustMaier:
### Author note — follow-up commit `aa19c844e6`

*(Written by me, the PR author. Every agent on this repo pushes under the same GitHub account, so nothing here should be read as a second party signing off — the substance below came from @zuri via our internal channel, and I am recording it, not endorsing my own work.)*

**Fixed:** `settlePlacement` was typed `{ action: string }` while every other stub in the file takes `unknown` and lets the cast at the assertion site pin the shape. Shaped *and* narrower than the real call, which passes `{ placementId, action, actorId? }` — a later test asserting on `placementId` would have got a type error reading as though the assertion were wrong. Now `PrismaStub<{ settled: boolean }>` like the rest, cast moved to the `mockImplementation` site. Re-verified: sticker file 0 errors, 67 tests pass, eslint clean.

**The write-path trace has been folded into the PR body**, where it belongs — it is the invariant this whole change rests on and it should be in the artifact's own documentation rather than a comment. Credit for it is stated there.

**A real second pair of eyes on this is still worth having** and cannot come from another agent here, for the account reason above. A human reviewer, or an adversarial pass prompted to refute, would be best pointed at: the write path (can any row reach `Placement.data` without `flip`/`opacity`?), the `unknown`-argument stubs, and whether the ~line 413 test still fails when the defaulting is removed.

PR #3919, opened 2026-08-14
fix(donations): pay the creator in the currency the donor gave

JustMaier:
Third instance of the same defect, after #3911 (placements) and #3917 (paid access). Found by a sweep for other places we convert currency silently.

## What was wrong

`donateToGoal` charged the donor's chosen account but named **no destination**, so `createMultiAccountBuzzTransaction` applied the buzz service's yellow default (`buzz.service.ts:1119`, `toAccountType ?? 'yellow'`). A **green donation reached the creator as yellow**.

Measured on ClickHouse `buzzTransactions`: **137 legs / 85,210 Buzz** green→yellow, first 2025-11-21, most recent **2026-08-09** — live at the time of writing.

## The change

One line: `toAccountType: buzzType`.

Safe by construction, not by convention:
- `buzzType` is derived server-side from the request's domain (`donation-goal.router.ts`, `getAllowedAccountTypes(ctx.features)[0]`). The wire carries no currency.
- **Blue is refused before any money moves** (`donation-goal.service.ts:241`). So only green and yellow can reach `toAccountType`.

Worth noting what the old behaviour did to blue, had it ever been reachable: it would have paid the creator **yellow** for a blue donation — turning non-transferable Buzz into cashable currency. The existing refusal is what prevented that, and the test now pins it.

## Tests

New `donation-goal.currency.test.ts`:
- green and yellow donations pay in kind, to the goal owner
- blue is refused with **no charge attempted**

**Mutation-tested**: removing the line prints `AssertionError: expected undefined to be 'green'` — it fails because the destination is unnamed, which is exactly the defect.

`pnpm run typecheck` 0 errors; 11/11 across both donation-goal suites.

## Related

Independent of #3911 and #3917 — different file, no shared code, any merge order.

Still open from the same sweep: **comic chapter early access** (`comics.router.ts:4949`) has the identical shape with zero traffic so far, and the root cause is that `createMultiAccountBuzzTransaction` defaults to yellow when the destination is omitted — which is how this bug has now appeared four times.

### DIFF
--- a/src/server/services/__tests__/donation-goal.currency.test.ts
+++ b/src/server/services/__tests__/donation-goal.currency.test.ts
@@ -1,0 +1,84 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * A donation reaches the creator in the currency the donor gave.
+ *
+ * It did not: the charge named no destination account, so the buzz service
+ * applied its yellow default and a green donation arrived as yellow — 137 legs
+ * / 85,210 Buzz between 2025-11-21 and 2026-08-09. Same defect as paid access
+ * (#3917) and placements (#3911), and the third place it was found.
+ */
+
+const { mockDbRead, mockDbWrite } = vi.hoisted(() => ({
+  mockDbRead: { $queryRaw: vi.fn() },
+  mockDbWrite: {
+    donationGoal: { findUniqueOrThrow: vi.fn() },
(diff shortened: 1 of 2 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3919.civitaic.com` — no commit needed.

PR #3921, opened 2026-08-14
fix(comics): pay the creator in the currency the buyer spent

JustMaier:
Fourth instance of the same defect, after #3911 (placements), #3917 (model paid access) and #3919 (donations). Found by a sweep for other places we convert currency silently.

## What was wrong

The comic-chapter early-access charge named **no destination account**, so `createMultiAccountBuzzTransaction` would apply the buzz service's yellow default (`buzz.service.ts:1119`) and a green purchase would pay the creator yellow.

## ⚠️ Correction to an earlier version of this description

This PR originally claimed the feature had **zero traffic**. **That was wrong** — my sweep was mis-run. Measured properly:

```
externalTransactionId LIKE 'comic-ea-%'
→ 72 legs, 12,212 Buzz, 57 buyers, 6 creators, 2026-03-18 → 2026-08-03
```

**This is a live, paid feature.** What does hold is that **all 72 legs are `yellow → yellow`** — every purchase so far came from the mature domain, so **no Buzz has been misrouted**. There are no green legs and no refund legs under any `comic-%` prefix.

So "nothing is broken today" survives; "nothing uses this" does not. The original description also used the zero-traffic claim to justify shipping without a test. With the premise corrected, that trade goes the other way — hence the test below.

## The change

The currency is derived once into a local and used on both sides:

```ts
const [spendType] = getAllowedAccountTypes(ctx.features);
// ...
fromAccountTypes: [spendType],
toAccountType: spendType,
```

`getAllowedAccountTypes` with no seed returns exactly one element — `appendDomainCurrency` pushes `'green'` or `'yellow'` unconditionally in both branches, so `spendType` can never be `undefined`, `blue` or `red`. Verified: every other `getAllowedAccountTypes` call in this router passes a `['blue']` seed; this one alone passes none, so the narrowing of `fromAccountTypes` provably changes nothing about what a buyer can spend.

(An earlier description said the compiler enforces that the two fields agree. It doesn't — they're two expressions that happen to read one local. The test is what holds it.)

## Tests

`src/server/routers/__tests__/comics.router.early-access-currency.test.ts` — four tests over `createCaller`: green domain pays green, mature pays yellow, a client-supplied currency is ignored, and the creator is the recipient.

Mutation-tested, each mutation caught by exactly the test that should catch it:

| mutation | result |
|---|---|
| delete `toAccountType` | 3 fail — key absent from the call |
| `fromAccountTypes` → `['yellow']` | 2 fail, green vs yellow |
| `toAccountType` → `'yellow'` (the literal defect) | 2 fail |
| `toAccountId` → the buyer | 2 fail |
| widen the input schema to accept a client currency | only the forgery test fails |

That last one is what makes the forgery test non-vacuous: it passes today only because zod strips the key, and goes red the moment a schema stops stripping it.

The buzz double returns a full response literal that the test parses through the **real** `createMultiAccountBuzzTransactionResponse` schema, so the double cannot return a shape the service never would — the trap that bit the placement tests, where a refund double returned only a prefix and a refund that moved nothing looked identical to one that moved everything.

Not covered, deliberately: the real buzz service's yellow default isn't exercised (the service is mocked, and exercising it means real money), and the refund path is asserted only negatively rather than with a lying double.

`pnpm run typecheck` 0 errors; 4/4; eslint clean.

## Merge order

**Merge #3917 first.** Until it lands, `paidAccessPayoutAccount` maps green→yellow for *model* early access while this PR pays green→green for *comic* early access — two sibling products with opposite rules. #3917 makes model paid access pay in kind too, so the inconsistency never exists in a merged state.

(An earlier description cited `model-version.blue-buzz-purchase.service.test.ts` as covering the identical shape. On `main` today that test asserts green→**yellow** — the opposite. #3917 flips it.)

## Related

The root cause behind all four: `createMultiAccountBuzzTransaction` silently defaults to yellow when the destination is omitted. Until that default is removed or made explicit, this bug recurs.

### DIFF
--- a/src/server/routers/__tests__/comics.router.early-access-currency.test.ts
+++ b/src/server/routers/__tests__/comics.router.early-access-currency.test.ts
@@ -88,12 +88,19 @@ const chapter = {
   project: { id: 9, name: 'A Comic', userId: CREATOR },
 };
 
+// `canViewNsfw` is deliberately CONSTANT across both rows. Setting it to
+// `!isGreen` made it a perfect proxy for the flag that actually decides the
+// currency, so a derivation keyed on the wrong flag passed every test — and the
+// two are not equivalent in production: `canViewNsfw` also requires a
+// non-restricted region, so a restricted-region buyer on the mature domain has
+// `isGreen === false` and `canViewNsfw === false`, and would have been charged
+// green on a yellow domain.
 const callerOn = (isGreen: boolean) =>
   comicsRouter.createCaller({
     user: { id: BUYER, isModerator: false },
     acceptableOrigin: true,
     tokenScope: TokenScope.Full,
-    features: { isGreen, comicCreator: true, canViewNsfw: !isGreen },
+    features: { isGreen, comicCreator: true, canViewNsfw: true },
   } as never);
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3921.civitaic.com` — no commit needed.

JustMaier:
Closing as redundant. The fix is already on `main`.

`comics.router.ts:4941` now derives `const spendType = domainSpendType(ctx.features)` and applies it to both `fromAccountTypes: [spendType]` and `toAccountType: spendType` — the same two lines this PR proposed, reached through a different helper. This branch is CONFLICTING against that.

**The tests did not land with the fix**, though, and they were the point. `src/server/routers/__tests__/` on `main` had no comics file, so nothing pinned the behaviour that four separate PRs have now had to fix.

Rescued to **#4088**, restored unchanged from this branch. They pass as-is against `domainSpendType`, and mutating `toAccountType` back to a literal `'yellow'` — the original defect — fails 2 of 5 in under a second with an assertion naming green against yellow.

PR #3922, opened 2026-08-14
fix(app-listings): resolve listing ownership KIND-AWARE, not block-first (#3844)

### LINKED ISSUE #3844: resolveListingAccess is block-first, not kind-aware: an offsite listing that carries an AppBlock resolves to the wrong owner after a transfer or a mod claim
`resolveListingAccess` (`src/server/services/blocks/app-access.service.ts`) resolves the owner as:

```ts
const ownerUserId =
  (listing.revisionOfId ? parent?.appBlock?.app?.userId : listing.appBlock?.app?.userId) ??
  listing.userId;
```

That is **block-first with a column fallback**, not kind-aware. Its own doc comment describes it as kind-aware, and that reads as true only because an ordinary off-site listing has no backing `AppBlock`, so the fallback is the only branch reached.

## The shape where it is wrong

`mapAppBlockToListing` mints `kind: 'offsite'` with a non-null `appBlockId` for any `AppBlock` that carries an `externalUrl` (reachable via the mod proc `backfillAppListings`). On that shape the block decides ownership even though, for an off-site listing, `AppListing.userId` **is** the canonical owner — there is no `OauthClient` in an off-site listing's ownership chain.

Two ownership writers move only the column for an off-site listing, and therefore leave the block naming the previous owner:

- `app-ownership-transfer.service::acceptTransfer` — step (2), the `OauthClient` move, is guarded by `if (isOnsite)`; step (3) writes only `AppListing.userId`.
- `offsite-moderation.service::claimListing` — the impersonation remedy (report → delist → claim → ban). It refuses a non-`offsite` listing outright and writes only `AppListing.userId`.

So after either write, on an off-site-with-a-block listing, `resolveListingAccess` keeps returning the **old** owner: the ex-owner (or the impersonator the claim was meant to dispossess) retains edit access, and the rightful owner is refused.

## Why this is being filed rather than fixed inline

PR #3840 routes five listing-ownership gates through `resolveListingAccess`. Before that PR those five gates compared against `AppListing.userId` directly and were therefore **correct on this shape**; afterwards they inherit the resolver's answer. That is a regression on this shape, recorded honestly rather than described as neutral.

It is currently **unreachable**, measured against production on 2026-08-12:

- `SELECT count(*) FROM app_listings WHERE kind = 'offsite' AND app_block_id IS NOT NULL` → **0**
- **0 of 22** `app_blocks` rows carry a non-null `external_url`, and a grep of `src/server` finds **no writer** of that column — every hit is a read, a projection or a test fixture. So the shape is not merely 0-row today, it is not mintable by any current code path.

Changing the resolver affects every caller of the primary listing-access predicate, so it belongs in its own PR with its own review.

## Suggested fix

Branch on `kind` in `resolveListingAccess`:

- `kind === 'onsite'` → `appBlock.app.userId ?? listing.userId` (unchanged; the column is a denormalized copy, the block is canonical, and the fallback still covers a dangling `app_id`).
- `kind === 'offsite'` → `listing.userId`, unconditionally, ignoring any backing block.

`resolveAccessibleAppBlockIds` in the same module already discriminates on `kind` rather than on block-nullness for exactly this reason, so today the two halves of one module disagree about which discriminator is authoritative.

Anything that changes here must keep the existing pin green or update it deliberately: `src/server/services/blocks/__tests__/app-access.denormalized-owner-drift.test.ts` → `🔴 RECORDED DISAGREEMENT (not an endorsement): offsite + a block ⇒ the BLOCK wins`, which pins the current behaviour and points back at this issue.

ZacxDev:
Fixes #3844.

## The defect

`resolveListingAccess` resolved the owner as:

```ts
const ownerUserId =
  (listing.revisionOfId ? parent?.appBlock?.app?.userId : listing.appBlock?.app?.userId) ??
  listing.userId;
```

Block-first with a column fallback, and **no branch on `kind`**. Its own doc comment
called it kind-aware, and that read as true only because an ordinary off-site listing has
no `AppBlock`, so the `??` fallback was the only branch anyone ever reached.

`mapAppBlockToListing` mints `kind:'offsite'` **with** a non-null `appBlockId` for any
`AppBlock` carrying an `externalUrl`. On that shape the **block** decided ownership — even
though for an off-site listing there is no `OauthClient` in the ownership chain and
`AppListing.userId` *is* the owner. Meanwhile both off-site ownership writers move only
that column and leave the block naming whoever it named before:

- `app-ownership-transfer.service::acceptTransfer` — its step (2) `OauthClient` move is
  `if (isOnsite)`-guarded, so the off-site path writes `AppListing.userId` alone;
- `offsite-moderation.service::claimListing` — the **impersonation remedy** (report →
  delist → claim → ban), which refuses a non-`offsite` listing outright.

The consequence, stated at its sharpest: **an impersonator would retain edit access after
a moderator claimed the listing away from them**, while the rightful owner was refused. A
two-sided inversion — a fix that only granted would be half a fix.

PR #3840 consolidated five ownership gates onto this resolver. Before that they compared
against `AppListing.userId` directly and were therefore *correct* on this shape; the
consolidation is what introduced the inversion, and #3844 recorded it honestly rather than
describing it as neutral.

### A second, independent arm found while fixing this

`beginListingRevision` clones a shadow revision with `userId: parent.userId` and
`appBlockId: null`, and **neither ownership writer ever touches a shadow** — `claimListing`
and `acceptTransfer` both write `where: { id: <the parent> }`. For an on-site listing the
block covers that (which is what `app-access.denormalized-owner-drift.test.ts` already
pins). For an **off-site** listing the column is the answer, so reading the shadow's frozen
copy reproduced the identical inversion **with no block involved at all** — an ex-owner
keeps their old listing's in-flight revision draft.

This was derived from the code (the clone plus the two `where: { id }` writes), not
measured against production; the PR does not claim a row count for it.

## Blast radius: latent, not live

Measured against production on 2026-08-12, the block arm is not merely 0-row, it is
**unmintable**: `kind='offsite' AND app_block_id IS NOT NULL` → 0 rows; 0 of 22 `app_blocks`
carry an `external_url`; and no writer of that column exists anywhere in `src/server`
(every hit is a read, a projection, or a test fixture). So this closes a latent
authorization inversion before something can mint the shape. It is not a live exploit, and
it is not cosmetic.

## The fix

One kind branch, written once:

```ts
export function resolveCanonicalListingOwner(args: {
  kind: string;
  blockOwnerUserId: number | null | undefined;
  listingUserId: number;
}): number {
  if (args.kind === 'onsite') return args.blockOwnerUserId ?? args.listingUserId;
  return args.listingUserId;
}
```

plus `canonicalOwnerWhereBranches(userId)`, the same rule as a Prisma `OR` for the two set
reads that enumerate ownership without resolving row by row.

**Every ownership-resolution site now routes through them** (population and how it was
established, below):

1. `app-access.service::resolveListingAccess` — the primary predicate. It additionally
   resolves the **column from the PARENT** on a shadow hop, the same row it already takes
   `kind` and `appBlockId` from (that is the second arm above; `revisionOf.userId` is now
   selected).
2. `app-access.service::resolveAccessibleListingIds` — the "my apps" set.
3. `app-access.service::resolveAppsNavAccess` — the `/apps` sub-nav probe. It and (2) now
   share one function call rather than two copies of the branches, so the tab and the page
   it opens cannot disagree.
4. `app-collaborator.service::toSeatListing` — the write-side seat loader behind
   `assertOwner`. Block-first, a dispossessed impersonator could still invite and remove
   collaborators on the rightful owner's app.
5. `app-ownership-transfer.service::loadOwnedListing` — the transfer-initiation gate.
   Block-first, an ex-owner could **initiate a new transfer** and hand the listing straight
   back out — the exact outcome `claimListing`'s in-tx pending-transfer cancellation exists
   to prevent.

### What deliberately does not change

- **On-site precedence.** `OauthClient.userId` (via `AppBlock.app.userId`) stays canonical,
  with the denormalized column still the fallback for a block with a dangling `app_id`.
  There is an explicit test for a **stale** denormalized column that disagrees — the case a
  naive "just use `listing.userId`" fix would break.
- **The seat half.** Accepted seats still confer `editor`; pending/rejected still confer
  nothing.
- **Moderator bypasses.** Untouched on both sides of the D1 divergence.
- **`DENORM_OWNER_HOLDOUTS`** in `app-access.call-site-ledger.test.ts` is **unchanged** —
  none of its three entries moved. Widening `AccessDb` to a `TransactionClient` (what
  `loadOwnedListingInTx` would need) is explicitly out of scope.
- **No schema change**, no migration.

## Test coverage

New: `src/server/services/blocks/__tests__/app-access.kind-aware-owner.test.ts` — the
resolver, the pure helper, the shadow arm, a **seam test** that drives `claimListing` and
the resolver against one shared in-memory store (so the post-claim state is what the remedy
actually wrote, not a fixture asserting it), and set-vs-resolver agreement.

Extended: the five consolidated gates are driven end-to-end as a **table** in
`app-access.denormalized-owner-drift.test.ts`, in both directions, on both arms. That
file's `🔴 KNOWN REGRESSION … the BLOCK wins` pin is **inverted deliberately**, with the old
expectation and its reasoning preserved in the comment. Write-side loader cases added to
the collaborator and transfer suites; the `getNavSummary` predicate assertion updated from
two branches to three.

**43 of the added/changed assertions are red at `origin/main` and green at HEAD.** Every
new case asserts both the grant and the refusal; positive controls prove each fixture is
genuinely drifted, that the DB fake honours its where-clause (it can match *and* miss), and
that the independent OR-evaluator can return true, false, and **throws** on a predicate
shape it was not taught — so a silently rewritten predicate cannot read as "no rows".

Mutation-tested: 7 mutants (drop the kind branch, invert it, drop the offsite OR branch,
drop a branch's `kind` clause, read the shadow's column instead of the parent's, and revert
each write-side loader to block-first). All 7 die, each on a named test with a specific
message.

Full unit project: **16,263 passed / 17 skipped / 16,280 total across 1,030 files**,
identical population to the pre-change run. `pnpm typecheck` clean. Prettier clean.

---

# Second commit: the "Invites" sub-nav tab must see ownership offers too

Found while auditing the resolver in commit 1, and fixed here rather than left as a note.

`/apps/invites` renders **both** a pending seat invite and an inbound ownership-transfer
offer. The tab that routes there is gated on `getNavSummary`'s `hasPendingInvites`,
computed by `resolveAppsNavAccess` — and that predicate knew about **seat invites only**.
So a user who had been offered an app and held no seat invite got **no tab at all**: the
offer existed, the page rendered it, and nothing in the chrome ever pointed at the page.

`hasPendingInvites` now also probes `app_ownership_transfers`:

```ts
{ toUserId: userId, status: 'pending', expiresAt: { gt: now } }
```

- **`toUserId` is the ADDRESSEE.** An offer the caller *sent* can never light their own
  tab, and a third party's offer can never light anyone else's. Keying on `fromUserId` is
  the one-token version of that leak, and it has its own test.
- **`expiresAt: { gt: now }` is the half a naive `status: 'pending'` check gets wrong.**
  Expiry here is a **read-time predicate with no sweeper behind it** — the schema says so
  on the column and the migration repeats it — so a long-dead offer keeps
  `status = 'pending'` forever and the tab would latch on permanently for anyone who was
  ever offered an app. Strict `>` matches `app-ownership-transfer.service::isLive`
  exactly, so the tab and the page cannot disagree about the instant.
- **Narrowed on the indexed `(to_user_id, status)` pair**
  (`app_ownership_transfers_to_status_idx`), `select: { id: true }`, existence only.
- **Wrapped in `safeCollaboratorQuery`.** `app_ownership_transfers` is created by the *same*
  manual-apply migration as `app_collaborators`, whose own rollback note promises that
  dropping those tables returns deployed code to the 42P01 path — "inert-and-owner-only,
  NOT broken". An unwrapped read would break that promise by 500-ing every `/apps` page's
  chrome for the whole deploy window. A 42703 (half-applied schema) still propagates.

`hasEditableApps` is deliberately **not** widened: a pending offer confers zero capability
until accepted, the same consent principle as an unaccepted seat. Lighting "My apps" for an
offer would advertise a page every child query refuses. That has its own test, because the
cheapest wrong implementation — OR-ing the new probe into both booleans — passes every
"the tab appears" case and fails only there.

## Query cost — stated, not buried

This makes the nav summary **four** existence probes instead of three. They share the
existing `Promise.all`, so **added serial latency is zero**, but it is **one more
concurrent round trip (and one more pool slot) per `getNavSummary`**, which the `/apps`
chrome renders on every page load. It **cannot** be folded into an existing query: the four
probes hit three different tables (`app_listings`, `app_collaborators` ×2,
`app_ownership_transfers`), and unioning across them would mean hand-written SQL. The probe
itself is as cheap as it gets — one indexed `findFirst` selecting one column.

## Independence

This depends on **nothing outside `main`**. `app_ownership_transfers` is in the schema here
and is queried directly; nothing from the in-flight transfer-UI PR is imported.

## Coverage

New `app-access.nav-pending-invites.test.ts` (21 tests): the four-quadrant table
(transfer-only / seat-only / both / neither), the expiry boundary asserted on **both** sides
of a strict `>`, an expired offer not suppressing a live seat invite, the default-`now`
path proving the injected clock is an affordance and not the only wired path, the leak
direction in both forms, all four terminal statuses with a positive control on the loop,
the zero-capability rule, the 42P01 degrade, the 42703 propagation, and the structural
narrowing. Plus two router-level cases in `blocks.router.getNavSummary.test.ts`.

**11 of these are red at the previous branch tip** (`acf7b6375c` — not merely at
`origin/main`) and green after. Two cases are labelled in-file as invariant guards rather
than counted as regression coverage: seat-invite-only and both-kinds pass on either side of
the change, and exist to keep the widening from *replacing* the seat arm.

**5 mutants, all killed**, each on a named test with its own message: drop the transfer arm;
drop the expiry bound; key on `fromUserId`; OR the offer into `hasEditableApps`; unwrap the
42P01 degrade.

Full unit project after both commits: **16,290 passed / 17 skipped / 16,307 total across
1,032 files**, 0 failures. `pnpm typecheck` 0 errors. Prettier clean. `test:lint-rules`
220/220.

### DIFF
--- a/src/server/routers/__tests__/blocks.router.getNavSummary.test.ts
+++ b/src/server/routers/__tests__/blocks.router.getNavSummary.test.ts
@@ -91,7 +91,9 @@ vi.mock('~/server/db/client', () => ({
   dbWrite: { modelBlockInstall: { findUnique: vi.fn() }, model: { findUnique: vi.fn() } },
 }));
 vi.mock('~/server/redis/client', async () => {
-  const actual = await vi.importActual<typeof import('@civitai/redis/client')>('@civitai/redis/client');
+  const actual = await vi.importActual<typeof import('@civitai/redis/client')>(
+    '@civitai/redis/client'
+  );
   return { ...actual, redis: mockRedis, sysRedis: mockSysRedis };
 });
 vi.mock('~/server/rewards/active/dailyBoost.reward', () => ({
@@ -346,14 +348,24 @@ describe('getNavSummary — the collaborator-aware flags', () => {
     ]);
   });
 
-  it('ownership is resolved BLOCK-FIRST, shadows excluded — same predicate as listMine', async () => {
+  it('ownership is resolved KIND-AWARE, shadows excluded — same predicate as listMine', async () => {
+    // 🔴 WAS "BLOCK-FIRST" (two branches) UNTIL ISSUE #3844. That predicate let an
+    // attached `AppBlock` decide ownership on an OFF-SITE listing, where
+    // `AppListing.userId` is canonical — so after `claimListing` or an off-site
+    // `acceptTransfer` (both of which move only the column) the sub-nav offered the
+    // "My apps" route to the PREVIOUS owner and hid it from the rightful one. The third
+    // branch is that fix, and this assertion is enumerated equality on purpose: it is the
+    // structural half of "the tab and the page it opens cannot disagree", the other half
+    // being `app-access.kind-aware-owner.test.ts`, which asserts both reads pass the
+    // shared `canonicalOwnerWhereBranches` output.
     const caller = blocksRouter.createCaller(fakeCtx(otherModUser) as never);
     await caller.getNavSummary();
     const where = mockDbRead.appListing.findFirst.mock.calls[0][0].where;
     expect(where.revisionOfId).toBeNull();
     expect(where.OR).toEqual([
-      { appBlock: { app: { userId: otherModUser.id } } },
-      { appBlock: { is: null }, userId: otherModUser.id },
+      { kind: 'onsite', appBlock: { app: { userId: otherModUser.id } } },
+      { kind: 'onsite', appBlock: { is: null }, userId: otherModUser.id },
+      { kind: { not: 'onsite' }, userId: otherModUser.id },
     ]);
   });
 
--- a/src/server/routers/__tests__/blocks.router.getNavSummary.test.ts
+++ b/src/server/routers/__tests__/blocks.router.getNavSummary.test.ts
@@ -44,10 +44,12 @@ const {
     appBlockPublishRequest: { findFirst: vi.fn() },
     blockUserSubscription: { findFirst: vi.fn() },
     blockBuzzAttribution: { groupBy: vi.fn() },
-    // The two COLLABORATOR-aware probes (`resolveAppsNavAccess`): an owned-or-seated
-    // listing drives "My apps", a pending invite drives "Invites".
+    // The COLLABORATOR-aware probes (`resolveAppsNavAccess`): an owned-or-seated listing
+    // drives "My apps"; a pending SEAT INVITE **or** an inbound live OWNERSHIP-TRANSFER
+    // OFFER drives "Invites" — `/apps/invites` renders both, so the tab has to see both.
     appListing: { findFirst: vi.fn() },
     appCollaborator: { findFirst: vi.fn() },
+    appOwnershipTransfer: { findFirst: vi.fn() },
   },
 }));
 
@@ -167,12 +169,14 @@ beforeEach(() => {
   mockDbRead.appBlock.findFirst.mockReset();
   mockDbRead.appListing.findFirst.mockReset();
   mockDbRead.appCollaborator.findFirst.mockReset();
+  mockDbRead.appOwnershipTransfer.findFirst.mockReset();
   // Default: nothing exists for anyone.
   mockDbRead.blockUserSubscription.findFirst.mockResolvedValue(null);
   mockDbRead.appBlockPublishRequest.findFirst.mockResolvedValue(null);
   mockDbRead.appBlock.findFirst.mockResolvedValue(null);
   mockDbRead.appListing.findFirst.mockResolvedValue(null);
   mockDbRead.appCollaborator.findFirst.mockResolvedValue(null);
+  mockDbRead.appOwnershipTransfer.findFirst.mockResolvedValue(null);
 });
 
 describe('getNavSummary — gate', () => {
@@ -184,6 +188,7 @@ describe('getNavSummary — gate', () => {
     expect(mockDbRead.appBlock.findFirst).not.toHaveBeenCalled();
     expect(mockDbRead.appListing.findFirst).not.toHaveBeenCalled();
     expect(mockDbRead.appCollaborator.findFirst).not.toHaveBeenCalled();
+    expect(mockDbRead.appOwnershipTransfer.findFirst).not.toHaveBeenCalled();
   });
 
   it('flag OFF: returns the all-false shape and runs NO existence query', async () => {
@@ -321,6 +326,32 @@ describe('getNavSummary — the collaborator-aware flags', () => {
     expect(result.hasEditableApps).toBe(true);
     expect(result.hasSubmissions).toBe(false);
     expect(result.hasApprovedApps).toBe(false);
+  });
+
+  it('🔴 an inbound OWNERSHIP-TRANSFER OFFER lights "Invites" through the router', async () => {
+    // The router-level half of the gap: `/apps/invites` renders a pending transfer offer
+    // as well as a seat invite, so the tab that routes there must see both. With seats
+    // empty, the ONLY thing that can light this flag is the transfer probe.
+    mockDbRead.appOwnershipTransfer.findFirst.mockResolvedValue({ id: 'aot_1' });
+    const caller = blocksRouter.createCaller(fakeCtx(modUser) as never);
+    const result = await caller.getNavSummary();
+    expect(result.hasPendingInvites).toBe(true);
+    // …and an offer confers no capability, so "My apps" stays dark.
+    expect(result.hasEditableApps).toBe(false);
+  });
+
+  it('the transfer probe is scoped to ctx.user.id as the ADDRESSEE, and status-filtered', async () => {
+    // Cross-user leakage is the failure this shape prevents: `toUserId` is the addressee,
+    // so an offer the caller SENT can never light their own tab.
+    const caller = blocksRouter.createCaller(fakeCtx(otherModUser) as never);
+    await caller.getNavSummary();
+    const where = mockDbRead.appOwnershipTransfer.findFirst.mock.calls[0][0].where;
+    expect(where.toUserId).toBe(otherModUser.id);
+    expect(where.status).toBe('pending');
+    // Expiry is a READ-TIME predicate with no sweeper — a dead offer keeps
+    // `status='pending'` forever, so this bound is what stops the tab latching on.
+    expect(where.expiresAt.gt).toBeInstanceOf(Date);
+    expect(where.fromUserId).toBeUndefined();
   });
 
   it('a PENDING seat does NOT light "My apps" — only "Invites"', async () => {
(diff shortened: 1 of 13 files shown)


ZacxDev:
## Preview Environment ✅

| | |
|---|---|
| **URL** | https://pr-3922.civitaic.com |
| **Commit** | `b8c38ae` |
| **Network** | ✅ HTTP 302 (9ms) |
| **Pipeline** | [View build](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3922-qsnpr) |
| **Updated** | 2026-08-14 18:41:53 UTC |

ZacxDev:

**✅ no per-route First Load JS regression vs `main` (`dd9dd29`).**

<details><summary>Per-route breakdown · commit <code>b8c38ae</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3922-qsnpr">build</a> · 2026-08-14 18:41:53 UTC</summary>

```
Routes analysed     : 255

   992.9 kB  /models/[id]/model-versions/[versionId]/wizard [budget 1.30 MB] OK
   992.9 kB  /models/[id]/model-versions/create         [budget 1.30 MB] OK

1 budget breach(es):
  ✗ shared 601.0 kB > 470.0 kB
```

</details>

ZacxDev:

| | |
|---|---|
| **Commit** | `b8c38ae` |
| **Pipeline** | [View run](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3922-qsnpr) |
| **Updated** | 2026-08-14 18:44 UTC |

ZacxDev:
**🔦 Lighthouse:** perf 61 · LCP 2.4s · CLS 0.229 · a11y 93 — [report ↗](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786733390659-68553.report.html)

<details><summary>Per-route scores (median of 5 runs · desktop · authed as ci-smoke-gold) · commit <code>b8c38ae</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3922-qsnpr">build</a> · 2026-08-14 18:49 UTC</summary>

| `/` | 61 | 93 | 2.4s | 0.229 | 246ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786733390659-68553.report.html) |
| `/models` | 64 | 93 | 2.8s | 0.261 | 201ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786733391612-13677.report.html) |
| `/generate` | 63 | 97 | 3.5s | 0.021 | 329ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786733392206-42048.report.html) |

</details>

PR #3927, opened 2026-08-14
fix(images): fall back to Meilisearch on an empty BitDex page for signed-in callers too

ZacxDev:
## Symptom

On the BitDex-backed `/images` feed, during any window where BitDex returns **zero documents**, a **signed-in** user gets an **empty feed** while an **anonymous** visitor is served normally. Same page, same filters — the result depends on whether you are logged in. The asymmetry is also why it is easy to miss: the path anyone can check while logged out is the path that works.

## Mechanism

`fetchBitdexPrimary` signals *"I can't serve this — fall back to Meilisearch"* by returning `null`; the caller then runs the Meili leg. Before this change that signal was:

```ts
if (!accumulated.length && !ownExcludedPromise) return null;
```

`ownExcludedPromise` is the optional **second pass** that re-adds the caller's own private / blocked / unpublished / nsfw0 / poi content, which the main (deliberately cacheable, per-user-clause-free) query filters out. It is issued only when:

```ts
const skipOwnExcluded =
  !input.currentUserId || bitdexCursor || (input.userId && input.userId !== input.currentUserId);
```

…i.e. **authenticated, first page, own/global feed**. So the guard keyed on whether that pass had been *issued*, never on what it *returned*:

| caller | second pass issued? | BitDex returns 0 docs |
|---|---|---|
| anonymous | no | `null` → falls back to Meili ✅ |
| authenticated, first page, own/global feed | **yes** | fallback unreachable → empty page served ❌ |
| authenticated, viewing another user's profile | no | falls back ✅ |
| authenticated, paginated (`bdx:` cursor) | no | falls back ✅ |

## The fix

Move the emptiness check onto the **post-merge** result:

```ts
const ownExcluded = await ownExcludedPromise;
if (ownExcluded?.documents?.length) { /* …merge + content-scope filters… */ }

if (!data.length) return null;

data = data.slice(0, limit);
```

The placement is load-bearing in both directions:

- **Not earlier.** The obvious edit — dropping `&& !ownExcludedPromise` so it reads `if (!accumulated.length) return null` — returns *before* the own-excluded merge and would throw away a legitimately non-empty page: a caller whose main pass returns nothing but who *does* have own-excluded content matching the current view must still be served that content.
- **After the content-scope filters.** The second pass is deliberately **unscoped** (its cache key is just `userId × sort × disablePoi`), and `modelVersionId` / `postId` / `postIds` / `types` / `baseModels` / `remixOfId` / `withMeta` / `fromPlatform` are applied during the merge. Own-excluded documents that don't survive those filters are not content for this view, so the emptiness that matters is the one *after* them.

Net diff in the hot path: two lines removed, one guard (plus comment) added. No refactor, no restructuring, no extra query.

Preserved exactly: anonymous behaviour; own-excluded-only pages; the main loop's early `break` and cursor/pagination handling.

## Paginated (`bdx:` cursor) callers — decision and justification

**Unchanged: an empty subsequent page still returns the fallback signal.**

The second pass is first-page-only (`bitdexCursor` sets `skipOwnExcluded`), so on a paginated request `ownExcludedPromise` is `null` and `accumulated` is empty ⇒ the old guard already returned `null` there. The new post-merge guard reaches the same conclusion by the same route: `data` is empty, so `null`. **This PR does not change the paginated arm at all**, and a test pins that (including an assertion that the own-excluded pass is never issued on that path, so the claim is about the code and not about the test's fake).

Why not "fix" it in the same change: making end-of-feed return an empty page instead of falling back is a **separate behaviour change** to the pagination contract, on the hottest feed path, and it is not what this bug is about. Worth noting as a follow-up rather than smuggling in here: the Meili leg parses a cursor as `offset|entry`, so a `bdx:`-shaped cursor yields `offset = 0` — a paginated fallback therefore re-serves from the top rather than continuing where the user was. That is pre-existing on `main`; it is not introduced, worsened, or newly reachable here. It should be decided on its own merits (suppress the fallback for paginated callers, or translate the cursor) rather than bundled into a bug fix.

Note also that a BitDex **error** is caught by the caller and falls back regardless — the `null` path is specifically "zero documents", which on a subsequent page is legitimately end-of-feed and on a first page is not.

## Tests

`src/server/services/__tests__/bitdex-empty-fallback.test.ts` (10 tests) drives the real seam through `getImagesFromSearch` and asserts the **reported source** (`bitdex` vs `meili`) plus literal expected ids — so "did it fall back" is observed end-to-end, not inferred from an internal return value. Expectations are pinned to literals, not derived from the implementation.

Red on `main`, green here:

| test | on `main` |
|---|---|
| authenticated + 0 documents + no own-excluded content → falls back | **RED** (`expected 'bitdex' to be 'meili'`) |
| authenticated + own-excluded content filtered out by content scope → falls back | **RED** (`expected 'bitdex' to be 'meili'`) |

Invariant guards (green on `main` — they pin behaviour that must not change, and are not counted as regression coverage): anonymous + 0 documents falls back (the control arm); authenticated + documents returned is served by BitDex; authenticated + own-excluded content in scope is served, not discarded; paginated + 0 documents falls back with no own-excluded pass; and four fake-fidelity/positive-control tests.

**Mutation results** — both BitDex passes go through the same `queryBitdex`, so each mutant was applied to the source and the suite re-run:

| mutant | killed by | message |
|---|---|---|
| revert to `if (!accumulated.length && !ownExcludedPromise) return null` at its old position | the 2 regression tests | `expected 'bitdex' to be 'meili'` |
| re-add `&& !ownExcludedPromise` to the new post-merge guard | the 2 regression tests | `expected 'bitdex' to be 'meili'` |
| **naive** `if (!accumulated.length) return null` placed before the merge | own-excluded-in-scope test, content-scope positive control, fake-fidelity test | `expected 'meili' to be 'bitdex'`, `expected [] to deeply equal [ 777 ]` |
| invert the guard (`if (data.length) return null`) | 8 of 10 tests | `expected 'meili' to be 'bitdex'` / `expected 'bitdex' to be 'meili'` |

**Fake fidelity.** The two BitDex passes are told apart structurally (the own-excluded pass asks for `availability = Private` inside an `Or`; the main pass filters it out with `Not(Eq(...))`), and that discrimination is itself asserted: exactly one main query at the caller's `limit` and exactly one own-excluded query at the fixed page size with no cursor/offset; zero own-excluded queries for an anonymous caller; the fake demonstrated returning both empty and non-empty on the same pass; and the content-scope filters shown to actually filter — the *same* document is kept under a matching `postId` and dropped under a different one.

### DIFF
--- a/src/server/services/__tests__/bitdex-empty-fallback.test.ts
+++ b/src/server/services/__tests__/bitdex-empty-fallback.test.ts
@@ -1,0 +1,345 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// A `null` return from the BitDex primary path is the signal to fall back to
+// Meilisearch. That signal used to be suppressed for signed-in callers: the
+// second ("own excluded") pass is only issued for an authenticated first-page
+// request, and its mere EXISTENCE — not its result — was enough to skip the
+// fallback. So when BitDex returned zero documents, anonymous visitors fell
+// back and were served normally while signed-in users got an empty feed.
+//
+// These tests pin the fallback on the POST-MERGE result: fall back when the
+// merged page is empty, keep serving when own-excluded content is the only
+// content there is.
+//
+// Same minimal-seam mocking as bitdex-feed-source.test.ts: stub the
+// event-engine-common submodule + infra clients + env so importing
+// image.service doesn't boot real infra, and null out the Meili client so the
+// Meili leg still RUNS (and still stamps source 'meili') without a live index.
+
+import type * as BitdexClient from '~/server/bitdex/client';
+import type * as FliptClient from '~/server/flipt/client';
+import type * as MeiliClient from '~/server/meilisearch/client';
+
+const { queryBitdexMock, getFliptVariantMock } = vi.hoisted(() => ({
+  queryBitdexMock: vi.fn(),
+  getFliptVariantMock: vi.fn(),
+}));
+
+vi.mock('~/server/logging/client', () => ({
+  logToAxiom: vi.fn(async () => undefined),
+  safeError: (e: unknown) => e,
+}));
+
+vi.mock('../../../../event-engine-common/feeds', () => ({
+  ImagesFeed: class {
+    populatedQuery = vi.fn();
+  },
+}));
+vi.mock('../../../../event-engine-common/services/metrics', () => ({
+  MetricService: class {
+    fetch = vi.fn();
+  },
+}));
+vi.mock('../../../../event-engine-common/services/cache', () => ({ CacheService: class {} }));
+
+vi.mock('~/env/server', () => ({
+  env: new Proxy({ LOGGING: [] as string[] } as Record<string, unknown>, {
+    get: (target, prop) => {
+      if (prop in target) return target[prop as string];
+      if (typeof prop === 'string' && (prop.endsWith('_URL') || prop.endsWith('_ENDPOINT')))
+        return 'https://test:test@localhost:5432/test';
+      if (
+        typeof prop === 'string' &&
+        /(_CONCURRENCY|_LIMIT|_MS|_PORT|_TIMEOUT|_MAX|_SIZE|_COUNT)$/.test(prop)
+      )
+        return 1;
+      return undefined;
+    },
+  }),
+}));
+
+vi.mock('~/server/db/client', () => ({ dbRead: {}, dbWrite: {} }));
+vi.mock('~/server/clickhouse/client', () => ({ clickhouse: {} }));
+vi.mock('~/server/redis/client', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- self-referential key proxy
+  const make = (): any => new Proxy(() => 'k', { get: () => make() });
+  const keyProxy = make();
+  return {
+    redis: {
+      get: vi.fn().mockResolvedValue('[]'),
+      set: vi.fn().mockResolvedValue(undefined),
+      packed: { get: vi.fn(), set: vi.fn() },
+    },
+    sysRedis: {},
+    REDIS_KEYS: keyProxy,
+    REDIS_SYS_KEYS: keyProxy,
+  };
+});
+
+vi.mock('~/server/meilisearch/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof MeiliClient>()),
+  metricsSearchClient: null,
+}));
+
+vi.mock('~/server/bitdex/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof BitdexClient>()),
+  queryBitdex: queryBitdexMock,
+}));
+
+vi.mock('~/server/flipt/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof FliptClient>()),
+  getFliptVariant: getFliptVariantMock,
+  getFliptBoolean: vi.fn().mockResolvedValue(false),
+}));
+
+import { getImagesFromSearch } from '../image.service';
+
+const CURRENT_USER_ID = 42;
+
+/**
+ * Both BitDex passes go through the same `queryBitdex`, so the fake has to tell
+ * them apart from its arguments or it would answer the same thing to both and
+ * every assertion below would be meaningless.
+ *
+ * The main pass filters availability with `Not(Eq(availability, Private))`; the
+ * own-excluded pass asks for the opposite inside an `Or`. Keying on that `Or`
+ * distinguishes them structurally rather than by position or call order.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- raw BitDex filter clauses
+type Clause = any;
+function isOwnExcludedQuery(filters: unknown): boolean {
+  if (!Array.isArray(filters)) return false;
+  return filters.some(
+    (clause: Clause) =>
+      Array.isArray(clause?.Or) &&
+      clause.Or.some(
+        (member: Clause) =>
+          Array.isArray(member?.Eq) &&
+          member.Eq[0] === 'availability' &&
+          member.Eq[1]?.String === 'Private'
+      )
+  );
+}
+
+const publicDoc = {
+  id: 101,
+  url: 'abc',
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Preview Environment ✅

| | |
|---|---|
| **URL** | https://pr-3927.civitaic.com |
| **Commit** | `05fd293` |
| **Network** | ✅ HTTP 302 (1ms) |
| **Pipeline** | [View build](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3927-59ls4) |
| **Updated** | 2026-08-14 19:49:45 UTC |

ZacxDev:

**✅ no per-route First Load JS regression vs `main` (`ffa0094`).**

<details><summary>Per-route breakdown · commit <code>05fd293</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3927-59ls4">build</a> · 2026-08-14 19:49:46 UTC</summary>

```
Routes analysed     : 255

   992.8 kB  /models/[id]/model-versions/create         [budget 1.30 MB] OK
   992.8 kB  /models/[id]/model-versions/[versionId]/wizard [budget 1.30 MB] OK

1 budget breach(es):
  ✗ shared 601.0 kB > 470.0 kB
```

</details>

ZacxDev:
## 🧪 Preview smoke tests — ✅ passed

| | |
|---|---|
| **Result** | ✅ 65 passed · 0 flaky · 0 failed |
| **Gate** | mod ✓ · tester ✓ · gold ✓ · restricted → /preview-restricted ✓ · anon → /login ✓ |
| **Commit** | `05fd293` |
| **Pipeline** | [View run](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3927-59ls4) |
| **Updated** | 2026-08-14 19:52 UTC |

ZacxDev:
**🔦 Lighthouse:** perf 28 · LCP 4.5s · CLS 0.225 · a11y 97 — [report ↗](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786737565795-57070.report.html)

<details><summary>Per-route scores (median of 5 runs · desktop · authed as ci-smoke-gold) · commit <code>05fd293</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3927-59ls4">build</a> · 2026-08-14 19:59 UTC</summary>

| `/` | 28 | 97 | 4.5s | 0.225 | 2205ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786737565795-57070.report.html) |
| `/models` | 34 | 93 | 3.5s | 0.262 | 876ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786737566507-89123.report.html) |
| `/generate` | 59 | 97 | 3.5s | 0.021 | 422ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786737567108-8316.report.html) |
| `/models/4201` | 98 | 94 | 1.0s | 0.000 | 0ms | — | — |

</details>

ZacxDev:
Adversarial audit run before merge. **Verdict: safe to merge, no blocking findings.** The change is strictly monotone — the old null-set is a subset of the new one, so no fallback is removed and no non-empty page is newly discarded. Red-on-`main` reproduced exactly (`2 failed | 8 passed`), and an independently-built 11-mutant battery killed 10; the survivor is a placement variant (guard below `slice(limit)`) that is arguably worse, since the schema permits `limit: 0` where the two differ.

Two claims in the PR description are wrong, and a reviewer would reason from them — correcting here rather than editing the body, since they may already have been read.

**1. "A BitDex error is caught by the caller and falls back regardless — the `null` path is specifically 'zero documents'."**

Not so. `queryBitdex` (`src/server/bitdex/client.ts:110-165`) returns `null` on *every* failure — missing `BITDEX_URL`, non-2xx (`if (!res.ok) … return null`), and any thrown error or abort (`catch (err) … return null`). It never throws. So inside `fetchBitdexPrimary` an error is **indistinguishable from zero documents**, and the caller's `catch` + `recordBitdexError` are effectively dead for query failures.

This makes the fix *more* valuable than the description claims: before it, a BitDex outage — not just an empty result — was being swallowed into an empty feed for signed-in callers, while anonymous callers fell through to Meilisearch normally.

Likely provenance of the mistaken claim: `bitdex-feed-source.test.ts:155` uses `queryBitdexMock.mockRejectedValue(...)`, i.e. a fake that can throw where the real client cannot. Worth aligning that fake with the real contract.

**2. "The Meili leg parses a `bdx:`-shaped cursor as `offset = 0`, so a paginated fallback re-serves from the top."**

The mechanism is not that. `getAllImagesIndex` always emits `${offset + input.limit}|${searchNextCursor}` (`image.service.ts:2836`), so the client's cursor is `N|bdx:{…}` and the parse at `image.service.ts:2582` recovers `offset = N`. The offset prefix survives; only the `entry` anchor (the `sortAtUnix` keyset position) is lost. `offset = 0` holds only for a hand-crafted bare `bdx:` cursor on the public REST endpoint.

There *is* a real pre-existing paginated-fallback defect, but it is a different one: the two engines order differently and the anchor is dropped, so a fallback page can duplicate or skip relative to the keyset position. Pre-existing either way, as the PR says — just not by the stated mechanism.

**3. Minor:** "no extra query" is true of `fetchBitdexPrimary` itself but misleading about the change. When the fallback finds content, `getAllImagesIndex` runs the full page pipeline — the Meilisearch search, the existence `findMany`, the reaction `findMany`, the enrichment `Promise.all`, and the metrics read. The steady-state increment on that leg is roughly +11–16%. Small, and worth the trade, but "no extra query" should be reworded.

**Follow-ups filed separately, not blocking:** a served-vs-fallback counter (this change is a fallback-rate change on a hot path with no way to observe it move, and per (1) an outage is indistinguishable from an empty index), and a pre-existing own-excluded scoping bug reproducible on `main`.

Issue #3907, opened 2026-08-14
Widening app-listings alone leaves /apps with no in-product entry point (user-menu gate reads appBlocks)


ZacxDev:
## Summary

The `/apps` store is being decoupled so the catalog (`app-listings`) can widen INDEPENDENTLY of the block runtime (`app-blocks-enabled`) — that is the documented purpose of the `appListings` flag, and the expected shape of the public store launch is "widen `app-listings` alone" (#3905).

**The only in-product route TO the store is not gated that way.** `src/components/AppLayout/AppHeader/appsNavVisibility.ts:28`:

```ts
export function appsNavVisibility(features: { appBlocksGetStarted?: boolean; appBlocks?: boolean }) {
  return {
    getStarted: !!features.appBlocksGetStarted,
    marketplace: !!features.appBlocks,   // ← the user-menu "Apps" → /apps entry
  };
}
```

So for a `{appListings: true, appBlocks: false}` cohort:

- `/apps` **renders** — the SSR gate and every store surface grant on `appListings || appBlocks` (`hasAppsStoreAccess`).
- The user-menu entry that links to it is **hidden** — it reads `appBlocks` alone.
- The sub-nav's own "Marketplace" tab does not help: it only renders once you are *already* on an `/apps/*` route.

Net: **the store is live but has no discoverable entry point.** Reachable by direct URL, an external link, or a bookmark only.

## This is a live gap in the launch plan, not drift

I want to be precise about what is and is not wrong here.

`appsNavVisibility.ts:10` explicitly says the entry *"stays gated on `appBlocks` (mod-only today)"*. That is a **deliberate standing decision**, correct for the pre-GA state where `appBlocks` was the only store gate. It is not drift, and it was not touched by #3905.

What changed underneath it is the *meaning* of store visibility: since the W13 decoupling, store access is `appListings || appBlocks`, and this file is now the one store-visibility decision that does not use the shared predicate. The two only agree because the current tester cohort holds both flags.

So there is nothing to "fix" today — and deliberately nothing was, in #3905. The point of this issue is that **the first flag flip that widens `app-listings` alone silently produces an unreachable store**, and that is the exact flip the decoupling exists to enable.

## Suggested resolution (a product decision)

Either:

[...]

PR #2821, opened 2026-06-29
feat(app-blocks): support off-site (external-link) apps in listings

ZacxDev:
## Model: PURE EXTERNAL LINK

Extends App Blocks listings to support **registered apps that just LINK OUT** to an externally-hosted app. The chosen product model is a **pure external link**: a marketplace listing card that opens an external URL in a new tab. **NO install, NO scopes, NO block token, NO subscription, NO on-platform hosting.** It is a stateless discovery/marketing entry in the marketplace. The presence of `externalUrl` on the `AppBlock` row is the discriminator — no separate `appType` enum.

## ⚠️ Migration — MANUAL APPLY ONLY

Per the DB rule (CNPG nvme0 does **not** auto-apply Prisma migrations), the migration file is committed for history only. **A human must apply the SQL below manually per environment** (psql/retool). CI/deploy does NOT run it. It is additive + non-breaking (single nullable column, idempotent).

`prisma/migrations/20260629120000_add_appblock_external_url/migration.sql`:

```sql
ALTER TABLE "app_blocks"
  ADD COLUMN IF NOT EXISTS "external_url" TEXT;
```

## What changed

**Schema** — `AppBlock.externalUrl String? @map("external_url")` in `schema.full.prisma` (regenerated kysely/models types). NULL = a normal on-platform app.

**Validation / registration**
- `external-app.schema.ts`: `validateExternalUrl` (https-only — rejects `http`/`javascript:`/`data:`/`mailto:`/relative/malformed/over-long) returning the canonical URL, and `assertNoOnPlatformSurface` (external-link is mutually exclusive with a declared page/iframe/target slot).
- `external-app.service.ts` `registerExternalApp` + `blocks.registerExternalApp` router mutation (mod-only, `enforceAppBlocksFlag`). **Single-step** — deliberately bypasses the bundle / Forgejo / approve pipeline (nothing on-platform to build/host/validate). Creates a structurally non-interactive `OauthClient` (`grants:[]`, `allowedScopes:0`, no origins) purely to satisfy the `AppBlock.appId` FK, then an `status='approved'` row with `external_url` set, `approved_scopes:[]`. Retry-safe via the deterministic `appblk-<slug>` id + slug-uniqueness check.

[...]

PR #3928, opened 2026-08-14
feat(apps): make the external-only store scope reachable and segment-able

ZacxDev:
Makes the **external-only** (`kind='offsite'`) App-store scope actually reachable by a segmented tester cohort. `resolveStoreVisibilityScope` already modelled `public-external`; two blockers meant no cohort could ever land in it. Code only — **no Flipt configuration is touched**.

## What changed

**1. `isExternalListingsPublicEnabled` is now user-aware** (`src/server/services/app-blocks-flag.ts`)

It was a bare `isFlipt(FLAG)` — a global eval (`entityId='global'`, empty context). A Flipt segment can only match on the context it is handed, so a `testers` segment resolved `false` for **every** member and the feature was unreachable by construction. It now mirrors `isAppBlocksEnabled` / `isAppListingsEnabled`: `isFlipt(FLAG, String(user.id), buildFliptContext(user))` when a user is present, and `resolveStoreVisibilityScope` threads its `opts` into that second axis.

Backward compatibility is a hard invariant and is tested three ways:
- **no user → the original global eval, byte-identical** — the mock is asserted to be called with the flag key *alone* (one argument), so entityId/context still fall back to the client's `('global', {})` defaults.
- **a base-enabled flag still resolves `true` for everyone** — anon, in-cohort, out-of-cohort and moderator. A plain base-`enabled` boolean has no rollout rules, so it is immune to whatever entityId/context the eval carries. Adding context can only **add** segmentation.
- **fail-closed** — absent flag / Flipt down → `false` → `none`, with or without a user.

**2. The page gate can now express the scope**

There was no client-side feature for `app-listings-public-external` at all, so `hasAppsStoreAccess` → `resolveAppsPageAccess` returned `notFound` for a viewer whose only qualification was that flag — while the server would have happily served them the offsite catalog. Two changes:
- new `appListingsPublicExternal` entry in `feature-flags.service.ts`, `fliptKey: 'app-listings-public-external'`;
- the **shared** `hasAppsStoreAccess` predicate becomes `appListings || appBlocks || appListingsPublicExternal`. Extended in the one place it lives (PR #3905's consolidation); no call site open-codes anything, and the call-site ledger test was read and kept accurate.

The new term grants **reachability only**. What the viewer then sees is still the server's `StoreVisibilityScope` → offsite listings only; onsite App Blocks stay hidden by the server, not by this gate. It also widens no block-RUNTIME surface — `/apps/installed`, `/apps/run/<slug>`, the author surfaces and the `blocks.*` procs still gate on `appBlocks` alone, asserted in the seam suite.

## Dark-by-default proof

`app-listings-public-external` does not exist in Flipt. `app-blocks-flag.external-scope.test.ts` models exactly that (only the two flags that *do* exist are registered) and asserts each cohort lands where it lands today:

| cohort | scope |
|---|---|
| moderator | `full` |
| app-dev-tester (`app-listings`) | `full` |
| plain logged-in user | `none` |
| would-be external tester | `none` |
| anonymous / no-arg | `none` |

It carries its own control (`the external flag really is absent`) so the block cannot pass vacuously. The pre-existing `app-blocks-flag.test.ts` — including `isExternalListingsPublicEnabled evaluates the plain global flag (no context)` — passes **unmodified**, and `resolveAppsPageAccess.test.ts` is **untouched**.

## The server/client seam

The same flag is evaluated twice per request, by two paths that had never shared a test:

- **server data path** — `isExternalListingsPublicEnabled({user})` → `isFlipt` (**async**) → `resolveStoreVisibilityScope` → the store read procs / REST endpoints.
- **page gate** — `featureFlags.appListingsPublicExternal` → `isFliptSync` (**sync**, plus a static `availability` fallback when Flipt answers `null`) → `ctx.features` → `hasAppsStoreAccess` → `resolveAppsPageAccess`.

Disagreement is silent in both directions: the viewer passes the page gate and gets an empty store, or is 404'd off a catalog the server would serve.

**Decision:** they agree *by construction* for a logged-in viewer — same key, same entityId (`String(user.id)`), same context object (both call the shared `buildFliptContext`). And they fail closed together: an absent flag makes the async `isEnabled` return `false` while the sync `isEnabledSync` returns `null` and falls through to the static availability — which is why the client entry is **`availability: []` and not `['mod']`**. With `['mod']` a moderator's client value would be `true` against a server `false`. That is the entry's whole reason for being `[]`, and it is mutation-tested.

**How it is tested:** `app-blocks-flag.external-scope.seam.test.ts` loads **both real sides** (real `feature-flags.service`, real `app-blocks-flag`, real shared predicate, real SSR resolver) against **one** fake Flipt config that faithfully models the async/sync asymmetry, and asserts the relationship over a 4-config × 4-cohort matrix:

> `resolveAppsPageAccess` grants ⟺ `resolveStoreVisibilityScope() !== 'none'`

with a non-vacuity check that the matrix really produced all three scopes and both gate outcomes, a positive control that the Flipt branch is live, a behavioural key-identity control (a neighbouring key moves neither side; the real key moves both), and a control that the service's 10s feature-flag cache is genuinely bypassed between cases.

**Residual, bounded, and pre-existing:** for an *anonymous* viewer the two evals differ in identity — `'global'`/`{}` on the server vs `'anonymous'`/`{isLoggedIn:'false'}` in the client gate. That is true of `app-listings` and `app-blocks-enabled` today and is invisible for both shapes this rollout uses (a user-property segment, or a plain base-enabled boolean). It bites only under a **percentage/entityId rollout** — which is measured and pinned in a test rather than asserted away, and called out on the flag's doc comment as the shape to avoid.

## Mutation matrix

Every mutant applied to the final tree; each dies on its **own** assertion, not incidentally. Run over `app-blocks-flag{,.external-scope,.external-scope.seam}.test.ts` + `hasAppsStoreAccess` + `resolveAppsPageAccess` + the call-site ledger (149 tests green at baseline).

| # | mutant | killed | representative failing assertion |
|---|---|---|---|
| 1 | `resolveStoreVisibilityScope`: external axis checked **before** privileged | 4 | `A MODERATOR IN THE EXTERNAL COHORT IS STILL full, NEVER NARROWED` — `expected 'public-external' to be 'full'` |
| 2 | axis 1 reads the **external** flag (arm/flag swap) | 19 | `privileged: a MODERATOR resolves full` |
| 3 | axis 2 drops `opts` (global eval, ignores user) | 4 | `non-privileged + external flag ON (in segment) → public-external` |
| 4 | helper **ignores its user argument** (silent revert to global) | 7 | `a SEGMENT-scoped flag matches for an IN-segment user` |
| 5 | helper reads `app-listings` instead (flag swap) | 15 | `reads ONLY the app-listings-public-external key` |
| 6 | helper drops the no-user branch (breaks byte-identical compat) | 2 | `no user → the ORIGINAL global eval, byte-identical (flag key alone)` |
| 7 | predicate `\|\|` → `&&` on the new term | 17 | `the external flag being FALSE cannot revoke either older flag` |
| 8 | predicate drops the new term (revert) | 8 | `the EXTERNAL flag ALONE grants access` |
| 9 | predicate `\|\|` → `??` (whole chain; mixing is a syntax error) | 6 | `{listings:false, blocks:true, external:false} → true` |
| 10 | predicate returns constant `true` | 19 | `a viewer qualifying via NOTHING still gets notFound` |
| 11 | predicate returns constant `false` | 33 | `external-only viewer reaches /apps — NOT notFound` |
| 12 | client entry `availability: []` → `['mod']` | 3 | `the client entry is fail-closed for a MODERATOR when the flag is ABSENT` |
| 13 | client entry `fliptKey` → `'app-listings'` | 10 | `KEY IDENTITY, proven behaviourally: both sides read the SAME Flipt key` |
| 14 | `resolveAppsPageAccess` grants unconditionally | 18 | `neither flag → notFound (gate intact, even with no session)` |
| 15 | ledger regex reverted to two flag names | 1 | `catches a re-inlined gate built from the EXTERNAL-only term` |
| 16 | ledger regex loses its `\b` word boundaries | 2 | `does NOT read appListingsPublicExternal as appListings (prefix collision)` |

16/16 killed, 0 survivors. Mutants 12 and 13 are the ones that matter for the seam: neither is reachable by any test that looks at one side.

## Verified / not verified

**Verified locally:** `pnpm typecheck` clean — instrument validated first with a planted `TS2322`, which was reported (1 error, exit 2), then removed (0 errors, exit 0). `vitest --project unit` across `components/Apps/__tests__`, `server/services/__tests__`, `server/routers/__tests__`, `tests/api/v1/apps`: **312 files / 5033 tests passed**, 0 timeout panics. `--project component` on the three store-gate browser suites: **55 tests passed** (non-zero, so the browser really ran). ESLint + Prettier clean on all 8 touched files.

**NOT verified:** nothing was run against a live Flipt instance or a deployed environment — the flag does not exist, so there is nothing to evaluate. The Flipt-side segment shape (`app-dev-testers`, a user-property segment, etc.) is asserted only against a faithful in-test model of Flipt's boolean evaluation, not against Flipt itself. No manual browser walk of `/apps` as an external-only viewer. The tRPC read procs and REST endpoints already branch on `scope` and were not changed, so their offsite-only filtering is unchanged and untested here.

## Two things worth knowing before the rollout

**Found while building the seam suite — a pre-existing divergence, recorded not fixed.** When Flipt is *unreachable*, the client gate falls through to each entry's static `availability` while the server helpers have none. A **moderator** therefore gets `appListings`/`appBlocks` true statically → the page gate grants — while `resolveStoreVisibilityScope` resolves `none` → empty store. It predates this change (a property of the two `availability: ['mod']` flags), is in the safe direction, and closing it means giving the server a static mod floor — a behaviour change to the privileged path, out of scope here. It is asserted explicitly so it cannot be mistaken for something this PR introduced, and so a future fix has a test to flip.

**No in-product entry point for the external-only cohort.** `components/AppLayout/AppHeader/appsNavVisibility.ts` still gates the user-menu "Apps" → `/apps` item on `appBlocks` alone (a standing decision with its own doc, deliberately not swept in). That menu item is the only route *to* `/apps`, so this cohort reaches the store by direct URL only. Tracked in #3907; the external-only cohort is the first real population to hit it, so an announcement link or campaign URL is needed until it lands.

## 🔴 Rollout ordering

**This code must deploy BEFORE the Flipt flags change.** The order is not interchangeable:

1. Merge + deploy this PR. Nothing observable changes — the flag does not exist, so every cohort resolves exactly as it does today.
2. **Then** create `app-listings-public-external` in Flipt, base `enabled: false`, with the tester segment.
3. Enable it for the cohort.

If the Flipt flag is created and segmented **first**, the still-deployed old code evaluates it globally (a segment can never match) *and* has no client-side feature to gate the page on — so the cohort gets `notFound` at `/apps` and the flag looks broken. The server and page gates only agree once this code is live.

**Flipt shape guidance** (documented on `APP_LISTINGS_PUBLIC_EXTERNAL_FLAG`): use a **segment rollout** for the tester phase, or a **plain base-`enabled` boolean** for a later anonymous-public flip. Avoid a **percentage/threshold rollout** — it hashes the entityId, and the server (`'global'`) and client (`'anonymous'`) disagree for anonymous viewers.

**Rollback** is flipping the Flipt flag off (or deleting it): everything returns to `full`/`none` with no deploy.

### DIFF
--- a/src/components/Apps/__tests__/appsStoreAccessCallSites.test.ts
+++ b/src/components/Apps/__tests__/appsStoreAccessCallSites.test.ts
@@ -39,6 +39,20 @@ import { describe, it, expect } from 'vitest';
  * (`preview / component-tests`) and do not block a merge. This unit-project ledger
  * does. So in CI all six sites are pinned STRUCTURALLY ONLY, by this file — which
  * is precisely why a hole in its masker (see below) is worth more than it looks.
+ *
+ * 🔴 TWO MEASURED LIMITS OF THIS FILE — issue #3932, not fixed here. Read them
+ * before treating a green run as "no site re-inlines the gate":
+ *   1. `INLINED_GATE` matches two flag names in PROXIMITY inside ONE expression.
+ *      Four re-inlining shapes evade it and were verified green: hoisted locals
+ *      (`const a = features.appListings; … a || b`), names split across a `;` (the
+ *      `[^;{}]` class stops there by design), a ternary, and
+ *      `[features.appListings, features.appBlocks].some(Boolean)`. Widening the
+ *      NAME list (as the external-only term did) does not touch the SHAPE.
+ *   2. The "EXACT" assertion below pins DIRECT importers only, so a site that
+ *      reaches the gate transitively is invisible to it — including
+ *      `pages/apps/[appBlockId]/index.tsx`, which routes through
+ *      `resolveLegacyAppRoute` → `resolveAppsPageAccess`. That is the site a real
+ *      disclosure defect appeared at, so the blindness is not academic.
  */
 
 const SRC = path.resolve(__dirname, '../../..');
--- a/src/components/Apps/__tests__/appsStoreAccessCallSites.test.ts
+++ b/src/components/Apps/__tests__/appsStoreAccessCallSites.test.ts
@@ -68,6 +68,15 @@ const SRC = path.resolve(__dirname, '../../..');
  *    entry point — that menu item is the only route TO `/apps` (the sub-nav's
  *    Marketplace tab only appears once you are already on `/apps/*`). Reachable by
  *    direct URL only. Tracked in issue #3907; not this file's to fix.
+ *
+ *    🔴 THE EXTERNAL-ONLY COHORT INHERITS THAT GAP, AND IS THE FIRST REAL POPULATION
+ *    TO HIT IT. `{appListingsPublicExternal, NOT appBlocks, NOT appListings}` holders
+ *    now pass the six gates below but the user-menu entry still reads `appBlocks`
+ *    alone, so `/apps` is direct-URL-only for them. Until #3907 lands, an
+ *    external-only rollout needs its own entry point (a link in the announcement, a
+ *    campaign URL) — the store being reachable is NOT the same as it being findable.
+ *    Deliberately out of scope here: converting that menu item is a product decision
+ *    on a module with its own standing doc, not a mechanical alignment.
  */
 const STORE_GATE_SITES = [
   'components/Apps/AppListingsMarketplaceBody.tsx',
@@ -205,6 +214,21 @@ function stripComments(code: string): string {
 }
 
 /**
+ * The flag names the store gate is built from. The predicate is
+ * `appListings || appBlocks || appListingsPublicExternal` — the third term admits
+ * the EXTERNAL-ONLY cohort, who reach the store and are then scoped SERVER-side to
+ * `kind='offsite'` listings. Any TWO of these joined by a logical operator in live
+ * code under SCAN_ROOTS is a re-inlined gate.
+ *
+ * 🔴 Kept as data, not baked into one literal regex, because the gate GREW: a
+ * two-name regex would have gone on reporting a clean zero for a site that
+ * re-inlined `appListings || appListingsPublicExternal` — the newest and least
+ * familiar half of the rule, i.e. the one most likely to be open-coded by someone
+ * who has not read this file.
+ */
+const STORE_FLAG_NAMES = ['appListingsPublicExternal', 'appListings', 'appBlocks'] as const;
+
+/**
  * An open-coded store gate in LIVE code.
  *
  * Deliberately matched on PROXIMITY + a logical operator rather than one literal
@@ -215,11 +239,18 @@ function stripComments(code: string): string {
  * collapsed first so line wrapping is irrelevant, and `[^;{}]` keeps a match inside
  * one expression rather than spanning statements.
  *
- * `\b` after each flag name matters: it stops `appBlocksPages` / `appBlocksAuthor`
- * — different flags with their own rules — from being read as `appBlocks`.
+ * `\b` after each flag name matters twice over: it stops `appBlocksPages` /
+ * `appBlocksAuthor` — different flags with their own rules — from being read as
+ * `appBlocks`, and it stops `appListingsPublicExternal` from being read as
+ * `appListings`. Both directions are pinned by the masker controls below.
  */
-const INLINED_GATE =
-  /appListings\b[^;{}]{0,120}?(?:\|\||&&|\?\?)[^;{}]{0,120}?appBlocks\b|appBlocks\b[^;{}]{0,120}?(?:\|\||&&|\?\?)[^;{}]{0,120}?appListings\b/;
+const INLINED_GATE = new RegExp(
+  STORE_FLAG_NAMES.flatMap((a) =>
+    STORE_FLAG_NAMES.filter((b) => b !== a).map(
+      (b) => `${a}\\b[^;{}]{0,120}?(?:\\|\\||&&|\\?\\?)[^;{}]{0,120}?${b}\\b`
+    )
+  ).join('|')
+);
 
 /** Collapse whitespace so a prettier line-wrap cannot hide a gate from the regex. */
 function flatten(code: string): string {
@@ -352,6 +383,58 @@ describe('the masker (validate the instrument before reading its verdict)', () =
     // would read them as `appBlocks` and flag legitimate code.
     expect(
       flatten(maskNonCode('const x = features.appListings || features.appBlocksPages;', 'p.ts'))
+    ).not.toMatch(INLINED_GATE);
+    expect(
+      flatten(maskNonCode('const x = features.appListings || features.appBlocksAuthor;', 'p.ts'))
+    ).not.toMatch(INLINED_GATE);
+  });
+
+  /**
+   * 🔴 THE THIRD TERM. `appListingsPublicExternal` admits the EXTERNAL-ONLY cohort
+   * and is now part of the gate, so a site re-inlining a pair that includes it must
+   * be caught — while `appListingsPublicExternal` alone must NOT be mistaken for
+   * `appListings` (they are different flags with different meanings, and the prefix
+   * relationship makes that the easy mistake for a `\b`-less regex).
+   */
+  it('🔴 catches a re-inlined gate built from the EXTERNAL-only term', () => {
+    const cases: Record<string, string> = {
+      'external OR listings':
+        'const ok = features.appListings || features.appListingsPublicExternal;',
+      'external OR blocks': 'const ok = features.appListingsPublicExternal || features.appBlocks;',
+      'external first, wrapped':
+        'const ok =\n  features.appListingsPublicExternal ||\n  features.appListings;',
+      'De Morgan with external':
+        'if (!features.appBlocks && !features.appListingsPublicExternal) return null;',
+      'destructured external': 'const ok = appListingsPublicExternal || appBlocks;',
+    };
+    for (const [label, src] of Object.entries(cases)) {
+      expect(flatten(maskNonCode(src, 'probe.ts')), label).toMatch(INLINED_GATE);
+    }
+  });
+
+  it('🔴 does NOT read appListingsPublicExternal as appListings (prefix collision)', () => {
+    // A single mention of the longer name, OR-ed with an unrelated flag, is not a
+    // store gate. Without `\b` the `appListings` alternative would match its prefix
+    // and report a false offender at every external-cohort call site.
+    expect(
+      flatten(
+        maskNonCode('const x = features.appListingsPublicExternal || features.canViewNsfw;', 'p.ts')
+      )
+    ).not.toMatch(INLINED_GATE);
+    expect(
+      flatten(maskNonCode('const x = features.appListingsPublicExternal;', 'p.ts'))
+    ).not.toMatch(INLINED_GATE);
+    // The sharp case: a hypothetical SIBLING of the longest name. Only the `\b` on
+    // `appListingsPublicExternal` itself rejects this — every other guard in this
+    // file passes it either way, so this is the one assertion that dies for that
+    // specific reason.
+    expect(
+      flatten(
+        maskNonCode(
+          'const x = features.appListingsPublicExternalOverride || features.appBlocksPages;',
+          'p.ts'
+        )
+      )
     ).not.toMatch(INLINED_GATE);
   });
 
--- a/src/components/Apps/__tests__/hasAppsStoreAccess.test.ts
+++ b/src/components/Apps/__tests__/hasAppsStoreAccess.test.ts
@@ -56,6 +56,96 @@ describe('hasAppsStoreAccess — the 2×2 flag matrix', () => {
 });
 
 /**
+ * 🔴 THE THIRD TERM — `appListingsPublicExternal` (the EXTERNAL-ONLY cohort).
+ *
+ * Its holders are viewers the SERVER will serve a `kind='offsite'`-only catalog to
+ * (`resolveStoreVisibilityScope` → `public-external`). Before it was added here,
+ * a viewer whose ONLY qualification was that flag hit `notFound` at the page gate
+ * and could never reach the store the server was ready to serve them — the gate and
+ * the data path disagreeing about the same flag.
+ *
+ * The full 2×2×2 is asserted rather than the one interesting cell, because the
+ * dangerous mistakes are the boring cells: an `&&` instead of an `||` (only the
+ * both-true row would notice), or the term being dropped from one branch.
+ */
+describe('hasAppsStoreAccess — the external-only term (full 2×2×2)', () => {
+  const BOOLS = [true, false] as const;
+  for (const appListings of BOOLS)
+    for (const appBlocks of BOOLS)
+      for (const appListingsPublicExternal of BOOLS) {
+        const expected = appListings || appBlocks || appListingsPublicExternal;
+        it(`{listings:${appListings}, blocks:${appBlocks}, external:${appListingsPublicExternal}} → ${expected}`, () => {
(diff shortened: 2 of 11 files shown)


ZacxDev:
## Preview Environment ✅

| | |
|---|---|
| **URL** | https://pr-3928.civitaic.com |
| **Commit** | `60095c4` |
| **Network** | ✅ HTTP 302 (12ms) |
| **Pipeline** | [View build](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3928-v6q5w) |
| **Updated** | 2026-08-14 20:35:59 UTC |

ZacxDev:

**✅ no per-route First Load JS regression vs `main` (`9da8ab4`).**

<details><summary>Per-route breakdown · commit <code>60095c4</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3928-v6q5w">build</a> · 2026-08-14 20:30:46 UTC</summary>

```
Routes analysed     : 255

  1007.4 kB  /user/[username]/shop/manage               [budget 1.30 MB] OK
   997.4 kB  /user/[username]/articles                  [budget 1.30 MB] OK
   993.4 kB  /user/[username]/3d-models                 [budget 1.30 MB] OK
   992.9 kB  /models/[id]/model-versions/create         [budget 1.30 MB] OK
   992.9 kB  /models/[id]/model-versions/[versionId]/wizard [budget 1.30 MB] OK

1 budget breach(es):
  ✗ shared 601.0 kB > 470.0 kB
```

</details>

ZacxDev:
## 🧪 Preview smoke tests — ✅ passed

| | |
|---|---|
| **Result** | ✅ 65 passed · 0 flaky · 0 failed |
| **Gate** | mod ✓ · tester ✓ · gold ✓ · restricted → /preview-restricted ✓ · anon → /login ✓ |
| **Commit** | `60095c4` |
| **Pipeline** | [View run](https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3928-v6q5w) |
| **Updated** | 2026-08-14 20:33 UTC |

ZacxDev:
**🔦 Lighthouse:** perf 46 · LCP 3.6s · CLS 0.235 · a11y 93 — [report ↗](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786739948478-7849.report.html)

<details><summary>Per-route scores (median of 5 runs · desktop · authed as ci-smoke-gold) · commit <code>60095c4</code> · <a href="https://tekton-dp.civitai.com/#/namespaces/tekton-builds/pipelineruns/pr-preview-3928-v6q5w">build</a> · 2026-08-14 20:39 UTC</summary>

| `/` | 46 | 93 | 3.6s | 0.235 | 442ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786739948478-7849.report.html) |
| `/models` | 65 | 93 | 2.8s | 0.261 | 183ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786739949122-38190.report.html) |
| `/generate` | 65 | 97 | 3.6s | 0.022 | 292ms | — | [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1786739949708-32798.report.html) |

</details>

ZacxDev:
## Audit follow-up — 🔴 F1 fixed, plus two corrections to the PR body above

Pushed `60095c4f11`. **The PR body is left as originally written on purpose** (a reviewer may already have read it); the corrections are here.

---

### 🔴 F1 — on-site slug disclosure via `/apps/[appBlockId]` — FIXED

The audit was right, and the module called its own shot. `src/components/Apps/resolveLegacyAppRedirect.ts` documented this hazard verbatim and named **this exact change** as the trigger — gate (1) of `getListingDetail` was unreachable there *only* by an accident of flag alignment (every viewer past the page gate resolved `full`), and widening the page gate ends that coincidence.

Concretely, once `app-listings-public-external` is on: external-only viewer → passes the widened page gate → `resolveLegacyAppRoute` runs `findApprovedListingSlug` on their behalf → `302 Location: /apps/store-preview/<onsite-slug>`. The destination 404s for them, so the leak is **the header and the query**, not the page body — but that is a slug plus proof that an approved on-site listing exists, disclosed to precisely the audience `public-external` exists to hide on-site apps from. Pre-PR that path was a bare `notFound` with no query and no header, so it also negated the module's own stated "a viewer without store visibility learns nothing / issues no query" property.

Not exploitable as merged (flag absent ⇒ the third term is always false). It would have armed at **rollout step 3**.

**Fix, as the module prescribes:** `resolveLegacyAppRoute` now takes a **required** `storeScope`, checked before the route param is read and before any query.

Three decisions worth flagging:

- **Server-resolved, not re-derived from `features`.** The caller resolves it with `resolveStoreVisibilityScope({ user: session?.user })`. This is a *disclosure* gate, so it must key off the same value the **data layer** keys off (`getListingDetail`'s `scope`) — the client feature object is a second evaluation that can disagree (Flipt outage ⇒ the client falls back to each flag's static `availability`, the server has no such fallback). It is an async SSR resolver, so the server helper is simply available; there is no reason to approximate it. Stated in a comment at both ends.
- **Allowlist (`!== 'full'`), not a denylist on `public-external`.** A denylist fails open the moment a fourth scope is added, and the scope union is exactly the kind of thing that grows. The allowlist also rejects `none` — which closes the Flipt-outage case the page gate alone misses (a moderator's `availability: ['mod']` static fallback grants the page while the server resolves `none`). That is the same pre-existing divergence this PR documents in the seam suite, now actually gated at this route.
- **Required, not optional-with-a-default.** A default would let a future caller silently re-open the disclosure; now it is a compile error.

Only gate (1) is replicated. `currentVersionDeployedAt` and `contentRating`/red-capability remain **documented as owed** — untouched, as instructed. The comment block that described the hazard as unreachable-by-alignment is rewritten: the alignment argument is gone, and it now says the `storeScope` check is what the property rests on, with a "DO NOT RESTORE IT" marker.

#### Mutation matrix for the new guard (5 killed / 1 equivalent, measured)

| mutant | killed | first failing assertion |
|---|---|---|
| delete the scope guard | 4 | `public-external → notFound, and NO SLUG LOOKUP IS PERFORMED` |
| check the **wrong** scope (`!== 'public-external'`) | 7 | `🔴 CONTROL: the SAME viewer with scope full DOES redirect` |
| **denylist** instead of allowlist | 2 | `none → notFound, no query (the Flipt-outage case…)` |
| move the guard **below the DB read** | 4 | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |
| make the guard inert (`if (false)`) | 4 | `full is the ONLY scope that proceeds (allowlist, not a denylist)` |
| move the guard **below the route-param check** | **0 — survives** | — see below |

**The no-query assertion result, since it was asked for explicitly:** it is the *only* thing that catches the after-the-DB-read mutant. That mutant still returns `notFound`, so every result-only assertion stays green; it dies on a **call count**. Both halves of the property are asserted (`notFound` **and** `findApprovedListingSlug` not called), for `public-external`, for `none`, and data-driven across the whole scope union.

**🔴 On the one survivor — it is an EQUIVALENT MUTANT, not a hole, and I did not paper over it.** The route-param check has no side effect and returns the identical bare `{ notFound: true }`, so **no input distinguishes** guard-before-param from guard-after-param. My first draft had a test *named* "the scope gate runs BEFORE the route-param check" — the mutation run proved that name was a claim the test could not support. Rather than invent a fixture that fakes a difference, I renamed it to what it actually pins (a scope-rejected viewer gets an **indistinguishable** `notFound` across every param shape, asserted as a relationship over the result set, with no query) and recorded the equivalence in the file's measured break-it ledger. The ordering that *is* security-relevant — guard before the DB read — is a different mutant and is killed.

---

### Correction 1 — the PR body's "reachability only" claim was wrong for this route

The body says the third term grants *"reachability only… onsite App Blocks stay hidden by the **SERVER**, not by this gate."* That is true for `/apps`, the store grid, the sub-nav and the related rail. **It was false for `/apps/[appBlockId]`**, which is exactly why F1 existed: there the server had no scope-aware gate at all, so widening the page gate did change what an external-only viewer could learn. As of `60095c4f11` the claim holds again — but by construction now, not by luck, and the code comment says so instead of relying on it.

### Correction 2 — the rollout-ordering rationale understated the wrong-order risk (F4)

The body says deploying the Flipt flag first means the cohort gets `notFound` and "the flag looks broken." That is only the **segmented**-flag case. For a **base-enabled** flag created before this deploys, the still-running old code evaluates it globally, the global eval returns `true`, and `resolveStoreVisibilityScope` returns `public-external` **for everyone including anonymous** — so the anon-capable REST endpoints (`src/pages/api/v1/apps/index.ts:89`, `[slug].ts:84`) start serving the offsite catalog publicly while `/apps` still 404s. The prescribed order avoids it either way, but the real consequence is a **public data exposure**, not a cosmetic one. Ordering unchanged; the reason it matters is stronger than stated.

### Correction 3 — the drive-by prettier reformat is gone

`isAppBlocksDevTunnelEnabled` is back to `origin/main`'s formatting. Worth recording *why* it kept reappearing: the collapse is prettier's own normalization of a **pre-existing** violation on `main`, so any local `prettier --write` over that file re-applies it. Checked before reverting — `.github/workflows/lint.yml` treats formatting on **modified** files as report-only (`::warning`, `continue-on-error`) and only blocks on **added** files, so reverting costs a warning `main` already emits, and the two new test files (which do block) are clean.

---

### Filed, not fixed here — #3932

Three pre-existing findings, all deliberately out of scope, now tracked in **#3932** and cited from the code:

1. **`INLINED_GATE` misses four re-inlining shapes** — hoisted locals, names split across a `;`, a ternary, and `[a, b].some(Boolean)`. All four were planted and the suite stayed green with its own instrument controls passing. This PR widened the *name list* to three flags; it did not change the matching *shape*. Noted in the ledger file's header so a green run is not over-read.
2. **The ledger's "EXACT" set pins DIRECT importers only** — so `pages/apps/[appBlockId]/index.tsx`, which reaches the gate transitively via `resolveLegacyAppRoute` → `resolveAppsPageAccess`, is structurally invisible to it. That is the one site F1 bit, which is what makes it worth widening.
3. **A doc claim contradicted by the schema** — the comment asserted "EVERY listing that carries an `appBlockId` is `kind='onsite'`", making the gap under `public-external` exactly 100%. `schema.full.prisma` says `appBlockId` is set for on-site **and** the #2821 off-site rows and is "NOT a kind discriminator." The shipped guard is unaffected (it rejects the scope outright rather than reasoning about kinds), but the blast-radius estimate was unreliable. Corrected in place rather than deleted.

---

### Verification for this commit

`pnpm typecheck` clean — instrument re-validated with a planted `TS2322` **in `resolveLegacyAppRedirect.ts` itself**, which was reported (1 error, exit 2), confirming the changed file is in the compilation set, then removed (0 errors, exit 0). Unit sweep across `components/Apps/__tests__`, `server/services/__tests__`, `server/routers/__tests__`, `tests/api/v1/apps`, `tests/pages/apps`: **316 files / 5066 tests passed**, 0 timeout panics. ESLint + Prettier clean on every touched file.

**Not verified:** still nothing against live Flipt or a deployed environment, and no manual browser walk of `/apps/<appBlockId>` as an external-only viewer — the guard is asserted through the injected-lookup unit seam, not an HTTP-level test.

PR #3931, opened 2026-08-14
fix(download): rejoin a query a client split with a second `?`

manuelurenah:
## What broke

Reported: `https://api.civitai.com/download/models/568485?fileId=484398?type=Model&format=SafeTensor&token=…` returns

```
400 {"error":"✖ Invalid input: expected number, received NaN\n  → at fileId"}
```

Two `?` in that URL. It is a legal URI whose query holds a **single** param, `fileId=484398?type=Model` — so the id parses as `NaN` and every param the caller meant to send after the stray `?` is swallowed into it, **including `token`**, which the route reads for API-key auth.

## Why it started now

#3864 (`adb2850c3e`, shipped in v5.1.1) pins `?fileId=<id>` into every per-file `downloadUrl` on `/api/v1/model-versions/[id]`. Our public docs tell callers to pass their key as `?token=` (`site/reference/model-versions.md`), and appending `?type=…&format=…` to a download URL is a long-standing client habit. Both were harmless while those URLs carried no query of their own.

Reproduced live on a healthy public model — not the deleted one from the report:

```
/api/download/models/501240?fileId=418901?token=…   → 400
/api/download/models/501240?token=…                 → 307
```

So any client that appends `?token=` to a `model-versions` `downloadUrl` has been broken since that deploy.

## Fix

`repairSplitQueryString(req)` in `src/server/utils/request-helpers.ts` rejoins raw `?` separators into `&`, rewriting **both** `req.query` and `req.url`. Called at the top of the download route — it has to run before `getServerAuthSession`, which reads `?token=` off `req.url`, or an API-key caller is still treated as anonymous.

- `%3F` is left alone; only a raw `?`, which cannot be part of a value the client meant to send, is treated as a separator.
- Params recovered from the split never overwrite a path param, so a repair cannot re-point the request at another resource.

## Verification

- 12 new tests (8 unit + 4 route-wiring); 107 pass across the download-adjacent suites
- Mutation check: stubbing the helper to a no-op fails 6+ with legible messages (`expected '2?type=Model' to be '2'`, `auth saw the unrepaired url…`)
- `typecheck: OK — 0 type errors`

## Review round (second commit)

- **Path param could be overwritten.** `?modelVersionId=999?fileId=2` defeated the first guard — it inferred path params by subtracting the query-string keys, and a query key with the same name deleted itself from that set, so the repair re-pointed the request at another model version. The guard no longer reasons about origin at all: a param is written only when it is missing or itself split, so anything Next parsed cleanly (path params included) is never replaced.
- **A raw `?` is legal inside a value** (RFC 3986 §3.4), so `?token=ab?cd` was truncated to `token=ab` plus a phantom `cd`. Only a `?` that opens a new `key=` is a separator now; that input is a no-op.
- **The split entry survived the repair** (`debug?type` stayed beside the `debug`/`type` pair it produced) and the rewritten url could carry an empty `&&` segment. Both fixed.
- **The error path logged `query: req.query` to Axiom**, which on this route includes the caller's API key. Redacted. Pre-existing, but this PR is what makes `token` parse correctly in the split case.
- Route test harness modelled Next's precedence backwards (query over path). Corrected to params-last, which is what let the path-param regression test be written at all.

Verification after the round: 112 pass across the download suites, `typecheck: OK — 0 type errors`. Both new guards are mutation-checked — loosening the separator rule fails the RFC test, dropping the intact-value rule fails the path-param and intact-param tests.

### Knowingly not done

- **Hoisting the repair into `withApiMetrics`** so every wrapper that reads `?token=` benefits. It is the more complete fix and probably the right one, but it changes query parsing for every API route in the app, which wants its own PR and its own risk call rather than riding along on a hotfix.
- **A counter on the repair** so we can tell when no client still needs the shim. The metric registry lives in `@civitai/telemetry`, so it is a cross-package change; an Axiom log per repair was the alternative and the volume is unknown.
- **Sharing the route-test mock harness** with the two sibling files in that directory, which are near-identical copies. Real duplication, but the extraction touches suites this PR otherwise leaves alone. `importOriginal` is not the fix here — spreading the real `~/server/redis/client` or `~/server/db/client` constructs clients at import.

## Not in this PR

- `/api/v1/models` still emits the older `?type=&format=` shape while `/api/v1/model-versions` emits `?fileId=` — the two endpoints disagree and that is worth its own look.
- The developer docs could say to append with `&`.
- The model in the original report (`497255`, SD3) is `Deleted`, so it 404s for non-moderators regardless of URL shape. Unrelated to this fix.

### DIFF
--- a/src/__tests__/pages/api/download/split-query-repair.test.ts
+++ b/src/__tests__/pages/api/download/split-query-repair.test.ts
@@ -189,7 +189,9 @@ describe('download route — a query split by a stray `?`', () => {
     expect(logged.query, 'the caller API key was shipped to Axiom in plaintext').not.toHaveProperty(
       'token'
     );
-    expect(logged.query.fileId, 'the rest of the query is still there to debug with').toBe('484398');
+    expect(logged.query.fileId, 'the rest of the query is still there to debug with').toBe(
+      '484398'
+    );
   });
 
   it('CONTROL: the same request already worked when the client used `&`', async () => {
--- a/src/__tests__/pages/api/download/split-query-repair.test.ts
+++ b/src/__tests__/pages/api/download/split-query-repair.test.ts
@@ -1,0 +1,176 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * The download route against the URL shape a client builds by appending the
+ * documented `?type=…&format=…&token=…` suffix to a `downloadUrl` that already
+ * carries `?fileId=<id>` — `…/models/568485?fileId=484398?type=Model&…`.
+ *
+ * The unit suite owns the repair itself; this file owns the WIRING, which the
+ * unit suite cannot see: that the repair runs before the route parses its
+ * schema (otherwise `fileId` is NaN → 400) and before it resolves a session
+ * (otherwise `?token=` is swallowed → the caller is anonymous, which on a
+ * gated file is a 401/404 rather than a redirect).
+ */
+
+const {
+  mockFindUnique,
+  mockGetServerAuthSession,
+  mockGetFileForModelVersion,
+  mockHasExceededLimit,
+  mockIncrement,
+} = vi.hoisted(() => ({
+  mockFindUnique: vi.fn(),
+  mockGetServerAuthSession: vi.fn(),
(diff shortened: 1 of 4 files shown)


ZacxDev:

Just add the **`preview`** label and a preview will build automatically at `https://pr-3931.civitaic.com` — no commit needed.

PR #3937, opened 2026-08-14
test(creator-studio): pin the POI guard on the licensing-fee writes

JustMaier:
## Why

Audit of `868kr8a69` asked whether a POI/private guard covers **every** fee/paid-access write in `apps/creator-studio`. It does:

- `licensingFee` is written only by `writeFee`, whose two callers `setLicensingFee` and `bulkSetLicensingFee` both apply `licensingFeeBlockedFor` (POI-only) first.
- Paid access never touches SQL from this app — it POSTs `/api/v1/model-versions/early-access`, which applies `paidAccessBlockedFor` (POI **or** private) *before* the permanent/`WEBHOOK_TOKEN` branch, so Creator Studio's permanent path does not bypass it.
- The other two direct Kysely writes in the app set `usageControl` only.
- `fee-csv.ts` is export-only; `ownedVersionsWithFee`, `VariedFeeEntry`, `VariedFeeSkip`, `FeeChange` and `FeePreview` have zero callers.

No guard was missing, so this PR adds **no** behaviour. What was missing is a check: deleting either guard call site broke no test. #3903's positive control pins the *predicate* in `@civitai/buzz`, not that Creator Studio calls it.

## What

Four cases over the two exported fee writers, with a fake Kysely so the test can assert the write never happened (a guard that refuses *after* writing would otherwise pass).

## Positive control

With `licensingFeeBlockedFor` neutered to `=> false`:

```
PASS (2) FAIL (2)
AssertionError: expected { ok: true } to deeply equal { ok: false, status: 400, …(1) }
AssertionError: expected { ok: true, updated: 2 } to match object { ok: false, status: 400, …(1) }
```

The two that keep passing are the non-POI cases, so the refusal is attributable to `poi` and not to a fixture that cannot save at all.

## Verification

- `pnpm typecheck` (creator-studio, svelte-check): `0 ERRORS 3 WARNINGS` — all three pre-existing in `packages/civitai-ui`.
- `pnpm lint` (creator-studio): exit 0, only pre-existing warnings.
- `pnpm test` (creator-studio): 7 files, 43 passed.
- `pnpm run test:unit:run`: `16 failed | 16609 passed (16643)`. **Identical on a clean tree with this diff removed** — six repo-scan drift guards whose own positive controls fail locally. Not introduced here.
- `blocks.router.workflow.test.ts` collected **347** tests (submodule initialised).

### DIFF
--- a/apps/creator-studio/src/lib/server/__tests__/licensing-fee-poi-guard.test.ts
+++ b/apps/creator-studio/src/lib/server/__tests__/licensing-fee-poi-guard.test.ts
@@ -1,0 +1,99 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { Membership } from '../membership';
+
+// Kysely fake. `rows` is what the owned-versions read returns; `written` records every `.set()` so a test
+// can assert the write never happened — a guard that returns 400 after writing would otherwise pass.
+const state = vi.hoisted(() => ({
+  rows: [] as Record<string, unknown>[],
+  written: [] as Record<string, unknown>[],
+}));
+
+vi.mock('$lib/server/db', () => {
+  const select: Record<string, unknown> = {};
+  const update: Record<string, unknown> = {};
+  const chain = (target: Record<string, unknown>) =>
+    new Proxy(target, {
+      get: (t, prop: string) => t[prop] ?? (() => chain(t)),
+    });
+  select.execute = async () => state.rows;
+  update.set = (values: Record<string, unknown>) => {
+    state.written.push(values);
+    return chain(update);
+  };
+  update.executeTakeFirst = async () => ({ numUpdatedRows: BigInt(state.rows.length) });
+  update.execute = async () => [];
+  const db = {
+    selectFrom: () => chain(select),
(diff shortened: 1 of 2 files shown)


JustMaier:
Carrying one finding forward from the audit, because the ClickUp task (868kr8a69) is being closed and archived and archived tasks don't surface in the usual sweeps.

**Before anyone adds `availability` to the `ownedVersions` select, read this.**

Swapping `licensingFeeBlockedFor` for `paidAccessBlockedFor` in `licensing-fee.ts` leaves the suite **fully green** today. That is not because the two are interchangeable — it is because `ownedVersions` never selects `Model.availability`, so the private half of `paidAccessBlockedFor` can never fire.

Add `availability` to that select and it becomes a live regression: Creator Studio starts refusing licensing fees on private models, which the shared doc-comment in `@civitai/buzz` says must be allowed. The asymmetry is deliberate — a private model has no audience to sell *access* to, but its per-generation fee is for when it's published.

No test was added for it on purpose: a private-model case would pass under **both** predicates as the code stands, so it would be a check that cannot fail.

PR #3942, opened 2026-08-15
feat(early-adopter): user opt-in setting + Flipt cohort flag

ZacxDev:
ClickUp: `868krm9cm`

Adds an **early-adopter opt-in** that users control from Account settings and that Flipt can segment on, so in-progress features can be shipped to the people who asked for them without a deploy per cohort change.

Pairs with **civitai/flipt-state#62**, which adds the `early-adopters` segment and the `early-adopter` flag. This PR is inert until that merges: the registry entry is `availability: []`, so with the flag absent nobody is in the cohort.

## Design: session-carried, not per-request

The value lives in `User.settings` (JSON — **no migration**) and is projected onto the session rather than read on the flag path:

```
User.settings.isEarlyAdopter
  -> shapeSessionUser         (auth hub projects it, right beside allowAds)
  -> SessionUser.isEarlyAdopter
  -> buildFliptContext        (emits the STRING 'true' / 'false')
  -> the `early-adopters` Flipt segment
```

`buildFliptContext` is deliberately built **once per request** (a documented optimization — every flag used to rebuild it N times), so a DB read there would undo that. The precedent copied is `isInCreatorProgram`, which is likewise a session-carried bit written by a service and read at evaluation time.

## The cost of that choice, and how it is paid

Session-carried means **cached**, so a naive version of this ships a toggle that silently does nothing. Two separate caches had to be handled, and they are the interesting part of the review:

1. **The hub session cache** (`session:data2:{id}`, 4h TTL). `setUserSettingHandler` now calls `refreshSession` when the value **changes**, which busts that cache and signals open tabs to re-pull. Gated on a change rather than key presence because the handler does a read-modify-write, so once a user has opted in the key rides along on *every* later settings write — a presence check would fire the refresh on unrelated toggles for exactly the most active cohort. This mirrors the `metricPrivacyChanged` comparison directly above it, and `creator-program.service`'s `refreshSession` after its own `User` write.

2. **The `FeatureAccess` memo** (`featureAccessKey`, 10s TTL) — found while writing the seam test, not by reading the code. That memo caches a whole resolved `FeatureAccess` keyed on user identity + host + region. Refreshing the session is **not sufficient on its own**: without the new input in the key, the freshly-refreshed session lands on the entry cached from before the toggle and the user keeps the old cohort decision for the rest of the TTL. The file already states *"KEY COMPLETENESS IS A CORRECTNESS INVARIANT — the cache key must include EVERY input hasFeature reads"*, so this change would have falsified that comment; `isEarlyAdopter` is now in the key and the comment says why.

`Session.needsCookieRefresh` was investigated as the lever and **rejected**: it is declared in `src/types/session.ts` and is set nowhere and read nowhere — vestigial from the next-auth era. Nothing here builds on it.

## Fail-closed

The registry entry is `availability: []` — dark by default, static evaluation false. If Flipt is unreachable or the flag is missing, the cohort is **empty**, not universal. `['user']` would have failed open and handed the features to everyone during a Flipt outage; there is a test pinning this branch.

## Files

| File | Why |
|---|---|
| `src/server/schema/user.schema.ts` | the key, in **both** `userSettingsSchema` (storage) and `setUserSettingsInput` (the narrow tRPC write allow-list — a key missing there is silently stripped on write) |
| `apps/auth/.../session-shape.ts` | hub projection off `row.settings`; `User.settings` was already selected by the producer query, so no query change |
| `packages/civitai-auth/src/types.ts` | the wire contract |
| `src/types/session.ts` | the app-side type |
| `src/server/services/feature-flags.service.ts` | context emission, the registry entry, and the memo key |
| `src/server/controllers/user.controller.ts` | `refreshSession` on change |
| `src/server/flipt/client.ts` | `EARLY_ADOPTER` for background paths with no request context |
| `src/components/Account/SettingsCard.tsx` | the switch, under a new "Features" divider |

## Tests

45 new tests. Every one was run against `origin/main` first: **33 fail there, 12 pass.** The 12 are invariant guards (they pin that the change does not *over*-fire — no refresh on unrelated writes, no enrolment from an unrelated settings blob) and pass at base only because the feature is absent; they are not counted as regression coverage.

- **`buildFliptContext`** — literal `'true'`/`'false'` strings (a boolean fails a `STRING_COMPARISON_TYPE` constraint), `'false'` and not `'undefined'` for a logged-in user who never opted in, and the key **absent** for an anonymous request.
- **Schemas** — round-trip in both, non-booleans rejected rather than coerced, plus a control proving an unlisted key really is stripped (so the "survives validation" assertions are not vacuous).
- **Hub projection** — explicit true/false, `undefined` (not `false`) when never set, mistyped values ignored, independent of tier.
- **Refresh** — fires on opt-in and opt-out, does **not** fire for an unrelated write by an already-opted-in user, and is `await`ed.
- **Seam** — real code from settings JSON through the hub projection and the real `buildFliptContext` to the resolved flag, with only `~/server/flipt/client` stubbed; the stub applies a matcher written from the flipt-state YAML rather than from our own context builder. Modelled on `blocks.router.flag-gate-hydrate.test.ts`.
- **Component** (browser mode) — the switch renders, reflects the stored value, sends the right boolean both ways, and is wired to refresh the session.

**Mutation-tested**: 11 mutations, each killed by the tests that own that guard and no others — dropping the context emission, dropping the `!!`, removing the opt-in from the memo key, flipping availability to fail-open, dropping the `refreshSession` call, turning change-detection into presence-detection, dropping the key from the write allow-list, dropping the hub projection, coercing `undefined` to `false` in the hub, ignoring the stored value in the switch, and dropping the UI refresh.

## Deploy ordering

`apps/auth` is a **separately deployed service**. The setting is writable and the UI works as soon as this merges, but `SessionUser.isEarlyAdopter` stays `undefined` — and so the cohort stays empty — until the **auth hub redeploys**. That direction is safe (fail-closed), but the flag will read as "not working" until the hub ships.

## Not fixed here (pre-existing, adjacent)

`featureAccessKey` also omits `user.onboarding`, which `buildFliptContext` reads for `isInCreatorProgram`. Joining the Creator Program therefore has the same up-to-10s staleness this PR fixes for `isEarlyAdopter`. Left alone deliberately — it changes cache-hit behaviour for every user and belongs in its own change.

### DIFF
--- a/apps/auth/src/lib/server/auth/__tests__/session-shape.test.ts
+++ b/apps/auth/src/lib/server/auth/__tests__/session-shape.test.ts
@@ -87,7 +87,10 @@ describe('shapeSessionUser — field mapping', () => {
   });
 
   it('passes permissions straight through', () => {
-    expect(shape({}, [], ['feature:a', 'feature:b']).permissions).toEqual(['feature:a', 'feature:b']);
+    expect(shape({}, [], ['feature:a', 'feature:b']).permissions).toEqual([
+      'feature:a',
+      'feature:b',
+    ]);
   });
 
   it('coerces filePreferences to an object (never a primitive)', () => {
@@ -99,7 +102,9 @@ describe('shapeSessionUser — field mapping', () => {
 
 describe('shapeSessionUser — tier / subscriptions', () => {
   it('resolves tier + subscription entry for a single active sub', () => {
-    const u = shape({}, [sub({ id: 'sub_1', status: 'active', product: { metadata: { tier: 'gold' } } })]);
+    const u = shape({}, [
+      sub({ id: 'sub_1', status: 'active', product: { metadata: { tier: 'gold' } } }),
+    ]);
     expect(u.tier).toBe('gold');
     expect(u.subscriptionId).toBe('sub_1');
     expect(u.memberInBadState).toBe(false);
@@ -127,7 +132,9 @@ describe('shapeSessionUser — tier / subscriptions', () => {
   });
 
   it('flags memberInBadState and keeps a primary sub id from a bad-state sub', () => {
-    const u = shape({}, [sub({ id: 'bad', status: 'past_due', product: { metadata: { tier: 'gold' } } })]);
+    const u = shape({}, [
+      sub({ id: 'bad', status: 'past_due', product: { metadata: { tier: 'gold' } } }),
+    ]);
     expect(u.memberInBadState).toBe(true);
     expect(u.tier).toBeUndefined(); // bad-state isn't "active", so no highest tier
     expect(u.subscriptionId).toBe('bad'); // but the bad-state sub is tracked so the user can manage it
@@ -184,7 +191,9 @@ describe('shapeSessionUser — membership override', () => {
   });
 
   it('never lowers a higher paid tier', () => {
-    const u = withOverride('bronze', [sub({ id: 'a', product: { metadata: { tier: 'founder' } } })]);
+    const u = withOverride('bronze', [
+      sub({ id: 'a', product: { metadata: { tier: 'founder' } } }),
+    ]);
     expect(u.tier).toBe('founder');
   });
 
--- a/apps/auth/src/lib/server/auth/__tests__/session-shape.test.ts
+++ b/apps/auth/src/lib/server/auth/__tests__/session-shape.test.ts
@@ -245,6 +245,43 @@ describe('shapeSessionUser — allowAds / redBrowsingLevel from settings (D)', (
   });
 });
 
+describe('shapeSessionUser — isEarlyAdopter from settings', () => {
+  it('projects an explicit opt-in / opt-out off the settings blob', () => {
+    expect(shape({ settings: { isEarlyAdopter: true } }).isEarlyAdopter).toBe(true);
+    expect(shape({ settings: { isEarlyAdopter: false } }).isEarlyAdopter).toBe(false);
+  });
+
+  it('stays undefined — not false — when the user never opted in', () => {
+    // Sparse on purpose: the overwhelming majority of session payloads must be
+    // byte-unchanged by this feature. `buildFliptContext` is what collapses
+    // undefined to the 'false' string, so nothing downstream needs a default here.
+    expect(shape({ settings: {} }).isEarlyAdopter).toBeUndefined();
+    expect(shape({ settings: null }).isEarlyAdopter).toBeUndefined();
+  });
+
+  it('ignores a mistyped value rather than passing it through or throwing', () => {
+    // The lenient .passthrough() parse drops the whole focused object on a type
+    // mismatch, so a garbage value must not surface as a truthy opt-in.
+    expect(shape({ settings: { isEarlyAdopter: 'yes' } }).isEarlyAdopter).toBeUndefined();
+    expect(shape({ settings: { isEarlyAdopter: 1 } }).isEarlyAdopter).toBeUndefined();
+  });
+
+  it('is independent of tier — a free user can opt in, a paying user can stay out', () => {
+    // The cohort must not be a proxy for membership; it is a pure user choice.
+    expect(shape({ settings: { isEarlyAdopter: true } }).isEarlyAdopter).toBe(true);
+    expect(shape({ settings: { isEarlyAdopter: false } }, [sub()]).isEarlyAdopter).toBe(false);
+  });
+
+  it('does not disturb the neighbouring settings-derived fields', () => {
+    // Same focused parse feeds allowAds/redBrowsingLevel; adding a key to it must
+    // not change how those resolve.
+    const u = shape({ settings: { isEarlyAdopter: true, allowAds: false, redBrowsingLevel: 31 } });
+    expect(u.allowAds).toBe(false);
+    expect(u.redBrowsingLevel).toBe(31);
+    expect(u.isEarlyAdopter).toBe(true);
+  });
+});
+
 describe('shapeSessionUser — meta / banDetails (parity)', () => {
   it('strips banDetails out of the output meta and yields undefined banDetails', () => {
     const u = shape({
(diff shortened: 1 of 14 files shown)


ZacxDev:
## ⚠️ Correction — my "fail-closed" claim in the PR description is wrong, and the seam test was blind

A blind adversarial audit found a **deploy-blocking defect in the companion PR, civitai/flipt-state#62**, which this suite was structurally unable to catch. Posting as a comment rather than quietly editing the body, since the original text asserts something false that a reviewer may already have read and believed.

### The claim that was wrong

The description says the registry entry is `availability: []` so *"if Flipt is unreachable or the flag is missing, the cohort is **empty**, not universal"*.

The first half is true. The **implication is not**: `availability: []` protects against Flipt being **down**, and nothing else. Once Flipt *answers*, `hasFeature` returns the Flipt result and never reaches static evaluation:

```ts
if (feature.fliptKey && !envOverriddenFlags.has(key) && _fliptModule) {
  const fliptResult = _fliptModule.isFliptSync(...);
  if (fliptResult !== null) return fliptResult;   // <-- static availability never consulted
}
```

So a **mis-defined** flag is not defended against at all. And flipt-state#62 as originally written was mis-defined: `enabled: true` with a single segment rollout. For a boolean flag, Flipt's `enabled` is the value returned when **no rollout matches** — not a master switch — so every non-opted-in user and every anonymous request would have evaluated to `true`.

### Why this suite didn't catch it

The seam test stubbed Flipt with:

```ts
const segmentMatcher = (ctx) => ctx?.isEarlyAdopter === 'true';
```

That models the **segment** and nothing else. It returns `false` for a non-matching user; real Flipt returns the flag's `enabled`. The file's own comment claimed the matcher was *"written from the flipt-state YAML"* — it was written from the *segment half* of that YAML and silently dropped the *flag half*.

So the test `the flag is OFF end-to-end for a user who did not opt in` passed against a fake that **disagreed with production in exactly the direction I was hoping for**. That is worse than having no fake: it converted an untested assumption into a green assertion. "Real code at every hop except the Flipt edge" was true, and the defect lived at the edge I replaced.

### Verified on prod Flipt v2.10.0 before changing anything

I re-measured rather than take the finding on trust. Both controls fired:

| Flag | Shape | Context | Result |
|---|---|---|---|
| `app-blocks-enabled` | `enabled:false` + segment | non-matching | `false` / DEFAULT — **negative control** |
| `app-blocks-enabled` | `enabled:false` + segment | **matching** | `true` / MATCH — **positive control** |
| `oauth-apps` | **`enabled:true`** + segment | non-matching | **`true` / DEFAULT** |
| `comic-creator` | **`enabled:true`** + segment | non-matching | **`true` / DEFAULT** |
| `api-key-buzz-limit` | **`enabled:true`** + segment | non-matching | **`true` / DEFAULT** |
| `buzz-memberships` | `enabled:false` + segment | non-matching | `false` / DEFAULT |

The positive control is the load-bearing row: `enabled: false` still lets a matching entity through, so the fix does not disable the flag.

### What changed here (commit `f0c2e36`)

**The stub now models both halves of the flag** and evaluates them the way Flipt does:

```ts
const EARLY_ADOPTER_FLAG = {
  enabled: false,                                        // the no-match DEFAULT
  rollout: { matches: (ctx) => ctx?.isEarlyAdopter === 'true', value: true },
};
const evaluateBoolean = (f, ctx) => f.rollout.matches(ctx) ? f.rollout.value : f.enabled;
```

**Proven load-bearing by mutation**, which was impossible before because the flag half didn't exist: flipping the transcribed default to `true` turns **6 tests red**, including the two that previously passed vacuously — `the flag is OFF end-to-end for a user who did not opt in` and `an anonymous request is never in the cohort`.

**Four new tests pin the class directly**, so it can't recur silently on the next flag:
- an `enabled: true` flag puts every non-opted-in user in the cohort
- an `enabled: true` flag also enrols **anonymous** requests
- app-side `availability: []` does **not** rescue an inverted flag (and the same test shows it *does* fail closed when Flipt returns `null` — the two cases separated explicitly)
- the transcribed flag is `enabled: false`, pinned as a literal

**Two comments corrected**, since a comment is a claim too:
- the registry entry comment no longer implies `availability: []` guards against a mis-set flag; it names the flipt-state precondition it actually depends on
- `FLIPT_FEATURE_FLAGS.EARLY_ADOPTER` said *"default-off — isFlipt returns false for an unknown flag"*, which describes the **unknown-flag** case, not this one. Once the flag exists, a context-less background call gets the flag's **default**, not `false`.

One test-hygiene fix fell out of this: anonymous requests share a single `featureAccessKey` (`anon|h:|r:`), so anonymous tests were reading each other's memo entries. They now use distinct hosts. That sharing is correct in production — all anonymous callers really are interchangeable — so the tests had to change, not the code.

### Also correcting: merge order

The description says *"safe to merge in either order"*. **That no longer holds.** The Flipt git backend is `branch: main`, `poll_interval: 30s`, so flipt-state#62 reaches production ~30s after merge. It must be merged **only** in its corrected form. This PR remains safe to merge at any time — it registers a flag key and adds no consumer.

Blast radius, stated accurately rather than dramatically: `features.earlyAdopter` has **no consumers** in this repo yet, so nothing user-visible would have been exposed. The flag would simply have been silently inverted in prod, and the first feature gated on it would have gone fleet-wide.

### Verification after the change

- 39 unit tests green (4 files), 30 `app:auth` green, all re-run **after** rebasing onto current `main` (which had moved 13 commits)
- `pnpm run typecheck` → `OK — 0 type errors`
- prettier clean, eslint 0 errors

### Not verified

Nothing here has been exercised against a live Flipt — `early-adopter` isn't deployed, so `isFliptSync` has never actually returned a value for it. The post-merge check is one evaluation: non-opted-in → `false`; `{"isEarlyAdopter":"true"}` → `true` / MATCH / `segmentKeys=["early-adopters"]`.

PR #3944, opened 2026-08-15
feat(feedback): image attachments, RUM session id, and opt-in page capture

ZacxDev:
Extends the feedback system from #3901 with the three gaps: image attachments, the RUM session id, and DOM auto-capture. All three ride in the existing `context` JSONB column — **no migration**, and none is needed.

ClickUp: `868krm9cn`

---

## 1. Images

A file picker feeds the existing `useCFImageUpload` hook; the Cloudflare image ids land in `context.images`.

**Upload budget** (`FEEDBACK_IMAGE_MAX_COUNT` carries the reasoning):
- Nothing is uploaded on attach — previews come from a local object URL, and the CF upload happens on **Send**. So an abandoned submission costs zero uploads.
- Bound: 3 attachments + 1 screenshot per submission × the existing 5/hour submit limit = **20 uploads/hour** on the successful path.
- Stated honestly in the constant's comment: that is a bound on the *successful* path. The rate limiter runs on the tRPC mutation, i.e. **after** the uploads, so a user who keeps re-pressing Send past the limit still spends uploads. `/api/v1/image-upload` carries no quota of its own today — any signed-in user can already mint upload URLs from anywhere in the app — so closing that is a change to *that endpoint*, deliberately not in this PR.

## 2. Session id

`faro.api.getSession()?.id` → `context.sessionId`, so a one-line report can be joined to that browser session's RUM signals.

🔴 **Absence is a first-class case, not an exception.** `FaroProvider` only initialises Faro when the build-arg, the collector URL and the `faro` flag all agree, so it is **not running in dev/test/preview**, and in production it is still absent for any session that blocked or did not sample the SDK. Until `initializeFaro` runs, the SDK's `faro` export is a bare `{}`. `getFaroSessionId()` returns a value and never throws; the field is omitted and the submission succeeds. Nine of its ten tests cover absence shapes.

No bundle cost: `@grafana/faro-web-sdk` is already in the shared graph via `_app.tsx` → `FaroProvider`.

## 3. DOM capture — opt-in with preview

**Posture, and why it is a product decision rather than an implementation detail:** a rendered capture of this site can contain NSFW imagery, another user's private message content in a DM, or moderator-only UI. Silent capture is a retention and compliance question. So:

- an explicit checkbox, **OFF by default**;
- the user **sees the capture before anything is uploaded** and can drop it;
- only the **viewport** is drawn — not the whole scrollable document — so anything they scrolled past is never captured;
  <br>⚠️ **This sentence was FALSE for `position: fixed`/`sticky` content until `7cb18983fa`** — a `scrollX: 0, scrollY: 0` option dropped the app's sticky header from every capture and drew the feed behind it. Fixed, and pinned by a real-render test with a control arm. Full write-up in [this PR comment](https://github.com/civitai/civitai/pull/3944#issuecomment-5300603878); the claim is accurate as written now.
- `allowTaint: false`, so a cross-origin image without CORS headers is dropped from the capture rather than tainting the canvas.

`html2canvas` (~200 kB) is reached **only** through a dynamic `import()`, so it costs nothing on the ~100% of page loads that never open the prompt.

### Bundle impact — measured

Controlled A/B through esbuild 0.25.8 (the bundler already in this repo's Vite toolchain), same entry, `--bundle --splitting --format=esm --minify`:

| | entry chunk | html2canvas |
|---|---|---|
| dynamic `import()` (this PR) | **122 B** | separate 204,242 B chunk, loaded on demand |
| static `import` (control) | **204,795 B** | merged into the entry |

Raw dep: 198,689 B minified / 45,896 B gzipped.

Scope, honestly: that is an **esbuild** measurement, not a `next build` one. I did not run a full production build. The durable protection is the source gate `src/server/services/__tests__/no-static-html2canvas-import.test.ts`, which fails if any file under `src/` imports html2canvas statically — it runs in the `unit` project, the tier that actually gates.

---

## Bounds (all client-supplied, all bounded)

| field | bound |
|---|---|
| `context.images` | 3 ids, 1–100 chars each, non-empty after trim |
| `context.screenshotId` | 1 id, 1–100 chars |
| `context.sessionId` | 1–64 chars |

Kept in keeping with the schema's existing header — client-supplied context is *a claim, not evidence* — and because a JSONB column stores exactly what it is handed. `screenshotId` is deliberately a **separate field** from `images`: triage needs to tell a rendered capture of the reporter's own screen from a file they chose to send, because the two carry different privacy weight.

## Tests — 126 new (103 unit, 23 component), from zero

There was **no feedback test coverage anywhere in the repo** before this.

| file | tests | project |
|---|---|---|
| `feedback.schema.test.ts` | 23 | unit |
| `feedback.service.test.ts` | 10 | unit |
| `feedback.router.test.ts` | 8 | unit |
| `captureScreenshot.test.ts` | 21 | unit |
| `getFaroSessionId.test.ts` | 10 | unit |
| `no-static-html2canvas-import.test.ts` | 14 | unit |
| `FeedbackPrompt.browser.test.tsx` | 19 | component (real Chromium) |
| `selectAttachments.test.ts` | 18 | unit |
| `captureScreenshot.render.browser.test.tsx` | 4 | component (real Chromium) |

**Red-at-`origin/main` / green-at-HEAD** — measured, not asserted. A baseline worktree at `origin/main` carrying only the new *modules* (constants, `captureScreenshot.ts`, `getFaroSessionId.ts`) plus the test files, so failures are behavioural rather than missing-import:

- schema: **17 of 23 red** at main (the round-trip and bound assertions; `z.object` silently strips undeclared keys, which is the exact bug shape)
- service: **1 of 10 red** — the boundary→service **seam** test. The direct forwarding tests pass at main because the service forwards whatever it is handed; they are labelled **invariant guards**, not regression coverage.
- router: **2 of 7 red**. The rate-limit configuration tests pass at main and are labelled invariant guards.
- component: **11 of 14 red**
- bundle gate against *pure* `origin/main`: **1 of 13 red**
- `captureScreenshot` / `getFaroSessionId`: new modules, so no meaningful main baseline — teeth proven by mutation instead (below).

**Mutation-tested, each killed by its own guard's message:**

| mutation | result |
|---|---|
| A: delete `if (consented !== true) return null` | **9 red**, message `expected "vi.fn()" to not be called at all, but actually been called 1 times` — i.e. the loader observable, not a neighbouring error |
| B: weaken it to `if (!consented)` | **3 red** — exactly the `'true'` / `1` / `{}` cases, proving the strictness is load-bearing |
| C: pass `consented: true` at the component call site | **survived — and the mutant is INVALID, not a survival**: the `!checked` branch has already returned there, so it is semantically equivalent. Recorded rather than hidden, and the code comment now says so instead of overclaiming. |
| C′: capture on prompt-open with no consent | **11 red**, including all three "no consent" tests that were vacuous at the main baseline |
| D: drop the schema's image-count cap | 1 red (`rejects 4`) |
| G: remove `getFaroSessionId`'s try/catch | 1 red (`does not throw when getSession itself throws`) |
| H: plant a real static `import html2canvas` | 1 red — the bundle gate's negative control |

**Reachability:** the consent check is the first statement in `captureConsentedScreenshot`; nothing precedes it, and `short-circuits BEFORE touching the DOM` pins that order.

**Real-dependency dogfood:** the browser test does **not** stub `captureConsentedScreenshot` — it delegates to the real implementation and swaps only the html2canvas *loader*, so the consent guard executes for real on every click. Separately, a throwaway browser spec (not committed) drove the **real** html2canvas against a live DOM in Chromium and produced a genuine 4,544-byte JPEG — so the options and the `toBlob` encode path are measured, not inferred.

## Verification

- `pnpm run typecheck`: **9 errors, byte-identical to `origin/main`** (all in `placement-escrow.service.ts` and `packages/civitai-telemetry`, unrelated). Instrument validated: it caught a real `TS2367` in my own code earlier in this branch.
- `prettier --list-different` on all 13 changed files: clean. Positive control observed — it flagged 3 files before they were formatted.
- `eslint` on all changed files: **0 errors, 0 warnings**.
- `pnpm install --frozen-lockfile`: resolves (`rc=0`). Lockfile change is purely additive: html2canvas + 4 transitive deps.

## Not covered

- No in-app **reader** for `Feedback.context` exists — true of the pre-existing fields too. Consumption is manual triage / SQL. The realistic failure mode here is *silent stripping at the boundary*, which is what the schema round-trip and the service `create`-argument assertions pin.
- Rate-limit enforcement cannot be tested behaviourally: `rateLimit` is a hard no-op under `isTest` by design. The tests assert the *configuration* the router hands it (5/3600, no `condition`, no `sharedKey`) — a `condition` is what would let a payload widen the quota.
- No `next build`; see the bundle-impact scope note above.

---

## Audit round — `7cb18983fa`

A blind adversarial audit found three user-visible issues, all now fixed. The one that mattered is a **correction to this description's own privacy claim** and is written up in full in [this comment](https://github.com/civitai/civitai/pull/3944#issuecomment-5300603878) — read that rather than trusting the annotated bullet above on its own.

1. 🔴 `scrollX: 0, scrollY: 0` dropped the sticky header from every capture.
2. `capturing` was missing from `busy`, so Send mid-capture silently submitted without the screenshot.
3. No size cap on user-chosen attachments (the generated capture was bounded; the picked files were not).

**A review fix resets the verification gate**, so everything was re-run rather than just the changed part:

- **Full 12-mutant battery re-run from scratch, all 12 killed by their own test's message.** Each mutant asserts its anchor occurrence count before applying, so a mutant that does not apply cannot score as a survival.
- Two findings came out of the battery itself and are recorded rather than hidden:
  - **Mutant G was INVALID on the first run** — it left a dangling `catch`, so the file failed to parse and vitest reported `Tests no tests`. That is a broken mutant, not a kill. Rewritten to a syntactically valid form; it now kills 1.
  - **Mutant J3 genuinely SURVIVED** — my negative-remaining clamp test used `alreadyAttached: 5` with 2 files, and `slice(0, -2)` on a 2-element array is `[]`, so it passed with the clamp deleted. The discriminating shape is `|remaining| < selected.length`. Test rewritten; the mutant now dies on the `remaining -1, 3 files` case.
- **Red at the pre-fix commit `d43eda8814`, green at HEAD**, per new behavioural test:

| test | pre-fix |
|---|---|
| `omits scrollX/scrollY …` (unit) | 🔴 red |
| `the shipped capture KEEPS the fixed bar the user can see` (render) | 🔴 red |
| `the Send button is disabled until the capture resolves …` | 🔴 red |
| `clicking Send mid-capture submits nothing at all` | 🔴 red |
| `an oversized file is refused, named, and never uploaded` | 🔴 red |
| `an oversized file does not consume one of the three slots` | 🔴 red |

`selectAttachments.test.ts` is new-module coverage with no meaningful pre-fix baseline; its teeth come from mutants J / J2 / J3 (5, 2 and 2 kills).

- Typecheck: **9 errors, byte-identical to `origin/main`**, none in any file this branch touches. Prettier clean (positive control: it flagged 3 files before formatting). ESLint 0 errors / 0 warnings.
- One test-harness trap worth naming: `userEvent.upload` **rebuilds the FileList and drops an instance property redefined on a File**, which silently turned the oversized fixture back into a small one — the size test passed the oversized file straight through and read as the cap not working. Those two tests now set `input.files` directly and **assert the picked size survived** before dispatching `change`, so a future runtime that copies again fails loudly instead of testing nothing.

## 🔴 Known open decision — NOT resolved by this PR

Stored captures and attachments land on the **public, unauthenticated CDN** with **no `Image` row**: no moderation scan, no retention policy, they survive account deletion, and they are **invisible to any orphan sweep keyed on the `Image` table**. That is a product/compliance decision, not a code change, and it is deliberately out of scope here. It should be settled **before a second feedback area is enabled** — one flag-gated area is a bounded exposure; a general mechanism is not.

Also noted and deliberately left as follow-ups: `useCFImageUpload` resolves `success` on a failed upload (`useCFImageUpload.tsx:138-145`, pre-existing), the retry-orphan case, and html2canvas's whole-document clone / CORS refetch cost.

### DIFF
--- a/package.json
+++ b/package.json
@@ -246,6 +246,7 @@
     "happy-dom": "^20.0.2",
     "he": "^1.2.0",
     "html-to-text": "^9.0.5",
+    "html2canvas": "1.4.1",
     "idb-keyval": "^6.2.0",
     "immer": "^9.0.15",
     "instantsearch.js": "4.64.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -421,6 +421,9 @@ importers:
       html-to-text:
         specifier: ^9.0.5
         version: 9.0.5
+      html2canvas:
+        specifier: 1.4.1
+        version: 1.4.1
       idb-keyval:
         specifier: ^6.2.0
         version: 6.2.2
@@ -6414,6 +6417,10 @@ packages:
   base-x@3.0.11:
     resolution: {integrity: sha512-xz7wQ8xDhdyP7tQxwdteLYeFfS68tSMNCZ/Y37WJ4bhGfKPpqEIlmIyueQHqOyoPhE6xNUqjzRr8ra0eF9VRvA==}
 
+  base64-arraybuffer@1.0.2:
+    resolution: {integrity: sha512-I3yl4r9QB5ZRY3XuJVEPfc2XhZO6YweFPI+UovAzn+8/hb3oJ6lnysaFcjVpkCPfVWFUDvoZ8kmVDP7WyRtYtQ==}
+    engines: {node: '>= 0.6.0'}
+
   base64-js@0.0.8:
     resolution: {integrity: sha512-3XSA2cR/h/73EzlXXdU6YNycmYI7+kicTxks4eJg2g39biHR84slg2+des+p7iHYhbRg/udIS4TD53WabcOUkw==}
     engines: {node: '>= 0.4'}
@@ -6954,6 +6961,9 @@ packages:
     engines: {node: ^14 || ^16 || >=18}
     peerDependencies:
       postcss: ^8.0.9
+
+  css-line-break@2.1.0:
+    resolution: {integrity: sha512-FHcKFCZcAha3LwfVBhCQbW2nCNbkZXn7KVUJcsT5/P8YmfsVja0FMPJr0B903j/E69HUphKiV9iQArX8SDYA4w==}
 
   css-select@5.2.2:
     resolution: {integrity: sha512-TizTzUddG/xYLA3NXodFM0fSbNizXjOKhqiQQwvhlspadZokn1KDy0NZFS0wuEubIYAV5/c1/lAr0TaaFXEXzw==}
@@ -8293,6 +8303,10 @@ packages:
 
   html-void-elements@3.0.0:
     resolution: {integrity: sha512-bEqo66MRXsUGxWHV5IP0PUiAWwoEjba4VCzg0LjFJBpchPaTfyfCKTG6bc5F8ucKec3q5y6qOdGyYTSBEvhCrg==}
+
+  html2canvas@1.4.1:
+    resolution: {integrity: sha512-fPU6BHNpsyIhr8yyMpTLLxAbkaK8ArIBcmZIRiBLiDhjeqvXolaEmDGmELFuX9I4xDcaKKcJl+TKZLqruBbmWA==}
+    engines: {node: '>=8.0.0'}
 
   htmlparser2@8.0.2:
     resolution: {integrity: sha512-GYdjWKDkbRLkZ5geuHs5NY1puJ+PXwP7+fHPRz06Eirsb9ugf6d8kkXav6ADhcODhFFPMIXyxkxSuMf3D6NCFA==}
@@ -11905,6 +11919,9 @@ packages:
   text-encoding-utf-8@1.0.2:
     resolution: {integrity: sha512-8bw4MY9WjdsD2aMtO0OzOCY3pXGYNx2d2FfHRVUKkiCPDWjKuOlhLVASS+pD7VkLTVjW268LYJHwsnPFlBpbAg==}
 
+  text-segmentation@1.0.3:
+    resolution: {integrity: sha512-iOiPUo/BGnZ6+54OsWxZidGCsdU8YbE4PSpdPinp7DeMtUJNJBoJ/ouUSTJjHkh1KntHaltHl/gDs2FC4i5+Nw==}
+
   text-table@0.2.0:
     resolution: {integrity: sha512-N+8UisAXDGk8PFXP4HAzVR9nbfmVJ3zYLAWiTIoqC5v5isinhr+r5uaO8+7r3BMfuNIufIsA7RdpVgacC2cSpw==}
 
@@ -12423,6 +12440,9 @@ packages:
 
   util-deprecate@1.0.2:
     resolution: {integrity: sha512-EPD5q1uXyFxJpCrLnCc1nHnq3gOa6DZBocAIiI2TaSCA7VCJ1UJDMagCzIkXNsUYfD1daK//LTEQ8xiIbrHtcw==}
+
+  utrie@1.0.2:
+    resolution: {integrity: sha512-1MLa5ouZiOmQzUbjbu9VmjLzn1QLXBhwpUa7kdLUQK+KQ5KA9I1vk5U4YHe/X2Ch7PYnJfWuWT+VbuxbGwljhw==}
 
   uuid@8.3.2:
     resolution: {integrity: sha512-+NYs2QeMWy+GWFOEm9xnn6HCDp0l7QBD7ml8zLUmJ+93Q5NF0NocErnwkTkXVFNiX3/fpC6afS8Dhb/gz7R7eg==}
@@ -18943,6 +18963,8 @@ snapshots:
     dependencies:
       safe-buffer: 5.2.1
 
+  base64-arraybuffer@1.0.2: {}
+
   base64-js@0.0.8: {}
 
   base64-js@1.5.1: {}
@@ -19499,6 +19521,10 @@ snapshots:
   css-declaration-sorter@7.2.0(postcss@8.5.6):
     dependencies:
       postcss: 8.5.6
+
+  css-line-break@2.1.0:
+    dependencies:
+      utrie: 1.0.2
 
   css-select@5.2.2:
     dependencies:
@@ -21305,6 +21331,11 @@ snapshots:
   html-url-attributes@3.0.1: {}
 
   html-void-elements@3.0.0: {}
+
+  html2canvas@1.4.1:
+    dependencies:
+      css-line-break: 2.1.0
+      text-segmentation: 1.0.3
 
   htmlparser2@8.0.2:
     dependencies:
@@ -25836,6 +25867,10 @@ snapshots:
 
   text-encoding-utf-8@1.0.2: {}
 
+  text-segmentation@1.0.3:
+    dependencies:
+      utrie: 1.0.2
+
   text-table@0.2.0: {}
 
   thenify-all@1.6.0:
@@ -26414,6 +26449,10 @@ snapshots:
     optional: true
 
   util-deprecate@1.0.2: {}
+
+  utrie@1.0.2:
+    dependencies:
+      base64-arraybuffer: 1.0.2
 
   uuid@8.3.2: {}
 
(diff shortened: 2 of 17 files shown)


ZacxDev:
## 🔴 Correction to this PR's description — two claims in it were wrong

A blind adversarial audit found a real defect in the DOM capture, and it makes a sentence I wrote in the description above **inaccurate**. Correcting it in a comment rather than silently editing the body, because a reviewer may already have read and believed it.

### What the description claimed

> only the **viewport** is drawn — not the whole scrollable document — so anything they scrolled past is never captured

and, in `captureScreenshot.ts`:

> only the VIEWPORT is captured — what the reporter can actually see

### What was actually true, until `7cb18983fa`

**True for in-flow content. FALSE for `position: fixed` and `position: sticky` content.**

The capture passed `scrollX: 0, scrollY: 0` to html2canvas. Those options look like they belong next to the `x`/`y` crop; they are actively wrong. html2canvas builds `windowBounds` from them and **adds it to every cloned element's `getBoundingClientRect()`** to recover document coordinates, while separately scrolling the clone iframe to those same values. The two cancel for in-flow content — which is exactly why no static fixture caught it — but not for fixed/sticky elements, whose rect is viewport-relative by definition.

Net effect: the app's sticky header (`AppLayout.tsx` — `sticky inset-x-0 top-0 z-50`) was placed at document y=0, outside the `y: window.scrollY` crop. So **every capture silently lost the header and drew the feed content it was covering instead**. Since the prompt lives in an infinite feed, the user is essentially always scrolled, so this was not an edge case.

That makes the on-screen checkbox copy — *"Only what's on screen is captured"* — wrong **in both directions**: the capture lost something that was on screen and gained something that was not.

### Measured, with a control arm

Reproduced in Chromium in this repo's own component project, and the fix verified the same way. Probe pixel at (10,10), viewport-sized capture, scrolled to y=1000, `position: fixed` red bar at viewport top-left, blue block at document y 1000–1200:

| arm | pixel at capture (10,10) |
|---|---|
| with `scrollX:0, scrollY:0` (the defect) | blue — the block **behind** the header |
| with both omitted (the fix) | red — the header, correct |

Reintroducing the two lines on top of the fix turns the new render test red with `expected red, got rgb(0,0,254)` — so the guard observes the actual rendered pixel, not the arguments.

### Why no test caught it

The existing unit assertion used `toMatchObject`, which says the listed keys are right and **never that an extra key is absent**. More fundamentally, the whole file pinned *what numbers were passed*, not *what html2canvas renders* — the seam between the two was unowned. Both halves are now closed:

- `captureScreenshot.test.ts` asserts the **exact option key set**, so any future option that shifts the coordinate space has to be looked at.
- `captureScreenshot.render.browser.test.tsx` is new and runs the **real html2canvas against a real scrolled document**, reading pixels out of the produced JPEG. It carries its own **negative control** that reapplies the defect through the same fixture and encode path and requires the opposite pixel — without which a green would be indistinguishable from a blank or mis-cropped capture.

The description's claim is now true as written. The module comment carries a `🔴 DO NOT ADD scrollX/scrollY` note with the mechanism, so it cannot be "tidied" back in.

---

### Two smaller corrections in the same commit

**`capturing` was missing from `busy`.** Send stayed enabled while a capture was in flight; `handleSubmit` reads `screenshot`, still `null` at that point, so the report went out with **no screenshot** and the panel switched to "Got it, thanks" while the capture completed onto a hidden panel. The user believed they attached a screenshot and did not. Now covered by three tests, including a deterministic in-flight click via a deferred capture.

**No size cap on user-chosen attachments.** `SCREENSHOT_MAX_BYTES` bounded the capture this feature *generates* and left the files the user *picks* unbounded — a 40MP phone photo uploaded at full size, decoded client-side first. Nine other upload surfaces here already enforce `constants.mediaUpload.maxImageFileSize`; this now does too. There is no server backstop (the presigned PUT from `/api/v1/image-upload` carries no size condition), so the client check is the only one that exists.

### One factual correction to the test table

The description says `feedback.router.test.ts` has **7** tests. It has **8**. Corrected counts after this commit: **103 unit + 23 component = 126**.

PR #3948, opened 2026-08-15
feat(cosmetics): version cosmetic hashes, close the write-path hole, surface near-matches

JustMaier:
Two badges in a creator shop imitating official Civitai badges were reported and nothing flagged them. Auditing why turned up three separate problems, only two of which we can fix ourselves.

## What was actually wrong

**Nothing read the hash.** `Cosmetic.pHash` has been written since 2026-08-01 and is consumed by no code at all — no lookup, no threshold, no mod surface. The review panel's only originality check is a sha256 of the uploaded bytes against other creator-shop submissions, which a re-encode defeats and which cannot see official cosmetics at all, since they carry no `imageHash` (468 of 829 shop items have one). Cosmetic **1162 "Civitwave" is a distance-0 clone of official cosmetic 322, published and on sale right now**, and that check passed it with a green "Original artwork" tick.

**Two write paths never hashed.** `submitCreatorShopItem` created the cosmetic inside its transaction and `updateCreatorShopItem` replaced `data.url` with a raw update; neither queued a hash. That is where **228 of the 231 unhashed cosmetics** in prod came from — every creator submission since the one-off backfill ran, growing at roughly the submission rate. Coverage was going backwards: 1,190 cosmetics with artwork, 950 fresh, 9 stale, 231 never hashed.

**64 bits is too few, and that part is not ours.** Measured over the 950 hashed cosmetics (450,775 pairs): mean 31.00, median 31, p1% 17, p5% 21. The two reported imitations sit at **Hamming 17 and 22** — inside the noise floor of unrelated artwork. No threshold separates them (d≤22 returns a median of 39 neighbours per submission and up to 192), and a rank-limited list puts the real match **7th and 67th** of 950. The orchestrator's other lane, `perceptualDct`, is also 64 bits and does not help (14 and 24 on the same pairs, probed live against prod). A 256-bit DCT over the same artwork ranks both pairs **1st** with nothing false above them, so the fix is more bits from the orchestrator — a separate ask on that team.

## What this ships, dark

- **`pHashHex` + `pHashVersion` + `pHashCheckedAt`.** The hash is TEXT so a width change needs a re-hash, not a migration. The version names the lane, because hashes from different algorithms are statistically independent with nothing to signal it. The migration seeds `pHashHex` from the existing BIGINT as the same lowercase hex the orchestrator originally returned (verified against live `mediaHash` output for cosmetics 107 and 870), so **no row needs re-hashing today**.
- **`cosmetic-phash-sweep`**, every 15 minutes, replacing `scripts/oneoffs/backfill-cosmetic-phash.ts`. It re-hashes anything missing, stale against its own artwork, or in an old lane — so **raising `COSMETIC_PHASH_LANE.version` drains the whole corpus by itself** when the orchestrator offers more bits. `pHashCheckedAt` is stamped on every attempt including failures, so the three cosmetics whose CDN artwork is gone cannot starve the queue by matching the predicate forever.
- **Both write paths queue a hash.**
- **`creatorShop.getSimilarCosmetics`** behind the `cosmeticSimilarity` flag (`availability: []` — dark, fails closed). The flag gates the **lookup**, not just the panel: a ranked list built on a hash that cannot rank is read by a moderator as a ranking. Candidates are gated on lane and on `pHashUrl = data->>'url'`, and the all-zero hash is excluded — 19 near-transparent decorations share it, so an equality match on it returns all of them and buries anything real.
- **A review panel with three distinct states**: "compared against N, nothing close", "these are the nearest", and "this was never fingerprinted". Those are different instructions to a moderator, and a blank panel says none of them.

## Needs doing by hand

- `packages/civitai-db-schema/prisma/migrations/20260814150000_add_cosmetic_phash_version/migration.sql` — **apply to prod and dev.** Re-runnable.
- Create the `cosmetic-similarity` Flipt flag. Leave it **off**.
- Nothing is user-visible until both are done.

## Verification

- `pnpm run lint` — no issues in the touched files.
- `pnpm run test:unit:run` — `blocks.router.workflow.test.ts` collected **347** tests (non-zero, so the submodule is initialised and the run means something). 16 failures, all in file-scanning drift guards (`app-spend-tier-privilege`, `block-scope.normalize-endpoint`, `orchestrator-chat.wait-unit`, `check-server-graph-singletons`, `typecheck`, `analytics-bucket-labels.drift`) — **identical set with this branch stashed**, i.e. pre-existing on `origin/main`, not caused here.
- Both write-path fixes were mutation-tested: removing either call fails a named assertion reporting "Number of calls: 0". Removing the stale-hash gate fails with `expected { status: 'ok' } to deeply equal { status: 'unavailable' }`.
- `pnpm run db:check-generated` fails on pristine `origin/main` too (a Prettier formatting difference in the generator's output), so the generated types here were hand-edited to the six lines the schema change adds rather than committing the generator's reformat of the whole file.

Full audit with every measurement: `_local/docs/plans/2026-08-14_cosmetic-phash-audit.md`.


JustMaier:
Orchestration-side ask filed as ClickUp `868krn12g` — https://app.clickup.com/t/868krn12g

It carries the measurements: both 64-bit lanes put the two real imitations at Hamming 17/22 and 14/24 against a corpus whose 1st percentile is 17, while a 256-bit DCT over the same artwork ranks both #1. When that type exists, the change here is raising `COSMETIC_PHASH_LANE.version` — the sweep re-hashes the corpus on its own.

PR #3949, opened 2026-08-15
feat(gifts): gift any tier, and consume it a month at a time

JustMaier:
Gifts stop being "a coupon for the tier you already have" and become **N months of a tier, consumed one month at a time**.

The mod team's reaction to the membership article was *"why do I have to match the tier they have, and how do I even know what tier they have?"* — that restriction existed because a 100%-off coupon discounts whatever subscription it lands on, so a "bronze" gift on a gold subscription was a free **gold** month.

**Not ready to merge — @Luis, this wants your eyes on the money path before it goes anywhere.** Design doc is in the private `_local` repo (`docs/plans/2026-08-14_gift-membership-stripe-audit.md`), with the full Stripe capability audit and every measurement behind the decisions below.

## What changes for a user

Anyone can gift any tier to anyone. The gift no longer applies itself — the recipient gets a queue and accepts explicitly. One rule decides what accepting does, computed server-side from their membership at that moment:

> **Gift tier ≥ yours (or you have no membership) → free months at the gift's tier.
> Gift tier < yours → the gift's cash value comes off your next N bills.**

No downgrade is ever offered, and there is no branch where value is wasted.

## Why month-at-a-time

A single `duration_in_months: 3` coupon ties the whole gift to one subscription surviving. Cancel after month 1 and the other two months evaporate — nobody refunded, nobody served.

**Measured on a Stripe test clock**: armed a gifted month, cancelled at period end, advanced past it. The final cycle invoice never happened and the month was never returned.

So `monthsRemaining` lives on our row and **only one month is ever live in Stripe**, as a `duration: 'once'` coupon. That gives us, for free:

- Cancelling mid-gift is recoverable — the months still owed become a free subscription at the gift's tier (`honorGiftResidualsForUser`).
- Two gifts never contend for Stripe's single discount slot, so no queueing machinery, no subscription schedules, no merged coupons.
- A holder who changes tier mid-gift just gets a different arming decision next month.

**A month counts as consumed when an invoice has used it, never when it is armed.** An armed month on a subscription that gets cancelled first is never used, and counting at arm time would silently eat it.

## Verified against Stripe test mode, not just mocks

API version `2022-11-15`, real subscriptions and test clocks, everything torn down afterwards.

| | Result |
|---|---|
| Bronze cash value applied to a **gold** subscription | 5000 → **4000** |
| Same coupon on a **EUR** subscription via `currency_options` | 5000 → **4000 EUR** |
| Item swap with `proration_behavior: 'none'` | **no invoice raised** — settles the "we can't charge them for something they didn't agree to" blocker |
| Tier switch + 100% off in one `update` | next invoice **$0** |
| A second coupon on a subscription that has one | **silently REPLACES it. No error.** |
| `invoice.discount.coupon.id` on a paid cycle invoice | present, including on the $0 one |

That fifth row is the sharp edge: Stripe will never stop us destroying a live discount, so `armNextGiftMonth` **refuses** when the slot is already occupied instead of clobbering. ~0.9% of active subscriptions carry a discount today.

Also worth knowing: `currency_options` is accepted and stored on our pinned API version but is **never returned to us on read**. Treat it as write-only — do not try to read it back to verify.

## Shape of the change

- `fulfillMembershipGift` now records payment, queues, notifies. It applies nothing.
- `acceptMembershipGift` — moves the holder up if the gift is a higher tier (`proration_behavior: 'none'`; the accept click is the consent record for the new recurring price), then arms month one.
- `armNextGiftMonth` — one `duration: 'once'` coupon, priced in the **subscription's** currency. Coupon ids are `gift_<id>_m<n>` so a retry is idempotent.
- `recordGiftMonthConsumed` — off `invoice.paid`; decrementing clears `armedCouponId`, which is what makes a webhook redelivery a no-op.
- `sweepGiftArming` — hourly backstop for a missed `invoice.paid`.
- `holderId` ships but nothing writes to it. Every read asks "is this gift mine" through it, so making gifts transferable later is a mutation plus UI, with no migration and no read-path rewrite. **Transfer itself is deliberately not built** — Justin's call after weighing the chargeback-after-transfer and black-market costs.

## ⚠️ Migration is applied by hand

`20260815010000_gift_membership_redemption` — additive. It **adds enum values and then uses them**, so it must not run inside a single transaction (Postgres refuses to use an enum value added in the same transaction). Run it statement-by-statement or under psql autocommit. It backfills `holderId = recipientId`, and marks the 9 existing `Fulfilled` rows `Completed`, since their value was already applied as a single multi-month coupon and re-arming would stack a second discount on top.

## Verification

- `pnpm run lint` — 0 new issues in changed files
- `pnpm run test:unit:run` — **50 passed** in the gift suite; `blocks.router.workflow.test.ts` collected **347**
- Mutation-tested the three money assertions, each fails legibly on revert: wrong-currency discount → `expected 'usd' to be 'jpy'`; dropped slot guard → `expected { armed: true } to match { armed: false }`; residual sized from `months` instead of `monthsRemaining` → `expected 3 to be 2`.

**Heads up, unrelated to this branch:** `origin/main` at `a94f1b70f5` is red — **16 unit tests across 6 files** fail on a clean checkout with none of this code (App Blocks / comics / scripts guards). Reproduced in a pristine worktree.

## Not verified

Nothing has run against live Stripe. No end-to-end run through real Checkout. The residual path marks the gift `Completed` the moment the free subscription is minted, so the queue shows it as "Used" while the free membership is still running — accurate, arguably confusing.


lrojas94 (approved):
I think it looks good, I'll try to get some time to manually play around with it, but it makes sense overall

PR #3956, opened 2026-08-15
fix(tests): clear the three type errors that reddened the typecheck-tests gate

JustMaier:
`scripts/ci/typecheck-tests-gate.mjs` was BLOCKED on `origin/main` at `3863adcbb0`. Three regressions arrived with tonight's merges and nobody's verification could have seen them: the root tsconfig excludes `src/**/__tests__/**`, so **`pnpm run typecheck` is blind to every one of them**. Test files have their own ratchet, and that ratchet is the only thing that looks here.

Before:

```
2 test file(s) have type errors and are NOT in the baseline:
  src/server/__tests__/minimax-h3-license.test.ts               (1)
  src/server/services/__tests__/cosmetic-phash.service.test.ts  (2)
1 baselined file(s) got WORSE:
  src/server/services/__tests__/sticker-placement.service.test.ts  16 -> 17
```

After: `PASS — no new or worsened test-file type errors.`

## The three fixes

Types fixed, nothing baselined away.

**`minimax-h3-license.test.ts`** — `~/server/common/constants` *imports* `BaseModel`, it does not re-export it (TS2459). Taken from its origin, `@civitai/shared/basemodel.constants`, which the file already imports from. Same type: `BaseModel = string`, which the test's own comment on line 35 depends on.

**`cosmetic-phash.service.test.ts`** — BigInt literals need `target: ES2020`; the repo targets ES2018 (TS2737). Rewritten as `BigInt(...)` calls rather than moving the repo's target, which is not this PR's business. `BigInt('-6421916719099894602')` and `BigInt(2)` are value-identical to the literals they replace.

**`sticker-placement.service.test.ts`** — `createStickerPlacement` grew a required `spendType: BuzzSpendType`; the shared `placeInput` fixture never did, so 17 call sites disagreed with the signature (TS2345). Three call sites had already been patched by hand with an inline `spendType`, which is why 16 of the 17 were sitting in the baseline.

This is a fixture that fell behind a signature, not a behavioural mismatch — `spendType` is threaded straight through to `holdPlacementEscrow`, which is mocked, so no assertion changes. The file goes from 16 baselined errors to **clean**.

Two deliberate choices in that fix:

- **The fixture carries `'yellow'`, and the escrow test asserts `'green'`.** They must not be the same value. `it('carries the caller currency into the escrow')` does `{ ...placeInput, spendType: 'green' }` and asserts `'green'` reached the escrow — if the fixture also said `'green'`, a `createStickerPlacement` that dropped the caller's currency entirely would still pass. Same discipline as the distinct-numbers block at the top of the file, applied to the currency.
- **The fixture is typed `CreateStickerPlacement`.** The next required field then fails once, at the fixture, instead of silently at 17 call sites — which is exactly how this one got in.

Six inline `spendType: 'yellow',` literals were removed where they sat *before* a `...placeInput` spread and were therefore already shadowed by it. The one at line 441 sits *after* the spread and is a real override; it stays.

## Baseline

Regenerated with `--write-baseline`, and it only goes **down**: `sticker-placement.service.test.ts` leaves the file entirely (16 → 0) and `remix-gallery.service.test.ts` drops 29 → 28 from an unrelated merge. `testFilesInProgram` moves 938 → 1018, which raises the gate's own positive-control floor.

## Verification

| check | result |
|---|---|
| `node scripts/ci/typecheck-tests-gate.mjs` | **PASS**, read from its printed output |
| `pnpm run typecheck` | `OK — 0 type errors`, unfiltered |
| `pnpm run lint` | exit 0 |
| `pnpm run test:unit:run` | 16 failed / 16750 passed, 6 files — the known Windows path-separator baseline |
| `blocks.router.workflow.test.ts` | **347 tests** collected |
| the three touched files | 72 / 17 / 9 tests, all green |

Positive control on the gate itself: I observed it print **BLOCKED and exit 1** before the change and **PASS and exit 0** after, in the same tree.

## Two findings I am reporting, not fixing here

1. **The gate is wired into nothing.** It appears in no `.github/workflows` job and no `package.json` script. It is a standalone file somebody has to remember to run — which is the whole reason it went red four hours after #3915 cleared it, and stayed red through several merges.
2. **`pnpm typecheck` was red in my worktree for a reason that had nothing to do with any of this**: 19 errors across `cosmetic-phash.service.ts`, `placement-escrow.service.ts` and the two grok files, all from a stale `node_modules` and Prisma client after a branch switch (`@civitai/client` beta.89 → .90; the new `Cosmetic.pHashHex` and `Placement.spendType` columns). `pnpm install && pnpm run db:generate` cleared all 19. Worth knowing before anyone reads those as real.

### DIFF
--- a/scripts/ci/typecheck-tests-baseline.json
+++ b/scripts/ci/typecheck-tests-baseline.json
@@ -1,8 +1,8 @@
 {
   "_comment": "Type errors inside src/**/__tests__/** that pnpm typecheck cannot see. Entries may shrink or disappear freely; adding one, or raising a count, is a deliberate act visible in review. Regenerate with: node scripts/ci/typecheck-tests-gate.mjs --write-baseline",
   "config": "tsconfig.tests.json",
-  "testFilesInProgram": 938,
-  "totalErrors": 801,
+  "testFilesInProgram": 1018,
+  "totalErrors": 784,
   "files": {
     "src/__tests__/pages/api/download/attachment-blocklist.test.ts": 1,
     "src/components/AppBlocks/__tests__/iframeInitController.test.ts": 4,
@@ -92,9 +92,8 @@
     "src/server/services/__tests__/nsfwLevels.buffer-flag.test.ts": 2,
     "src/server/services/__tests__/placement-escrow.service.test.ts": 27,
     "src/server/services/__tests__/reaction-stale-replica-delete.test.ts": 2,
-    "src/server/services/__tests__/remix-gallery.service.test.ts": 29,
+    "src/server/services/__tests__/remix-gallery.service.test.ts": 28,
     "src/server/services/__tests__/set-user-setting.sql.service.test.ts": 1,
-    "src/server/services/__tests__/sticker-placement.service.test.ts": 16,
     "src/server/services/__tests__/sticker-usage-database-target.test.ts": 2,
     "src/server/services/__tests__/unblock-image-nsfwlevel-reset.test.ts": 1,
     "src/server/services/__tests__/update-user-profile.sql.service.test.ts": 1,
--- a/src/server/__tests__/minimax-h3-license.test.ts
+++ b/src/server/__tests__/minimax-h3-license.test.ts
@@ -7,7 +7,7 @@ import {
   licenses as sharedLicenses,
 } from '@civitai/shared/basemodel.constants';
 import { baseModelLicenses } from '~/server/common/constants';
-import type { BaseModel } from '~/server/common/constants';
+import type { BaseModel } from '@civitai/shared/basemodel.constants';
 
 // The 2 Aug 2026 revision the on-site weights carry. A moving ref here would let
 // MiniMax rewrite the terms we point creators at, which is what section III.1's
--- a/src/server/services/__tests__/cosmetic-phash.service.test.ts
+++ b/src/server/services/__tests__/cosmetic-phash.service.test.ts
@@ -77,7 +77,7 @@ describe('normalizeCosmeticHashHex', () => {
 
 describe('legacyBigIntHash', () => {
   it('writes the legacy BIGINT only for the perceptual lane', () => {
-    expect(legacyBigIntHash('a6e0c4c4cce8a4b6', 'perceptual')).toBe(-6421916719099894602n);
+    expect(legacyBigIntHash('a6e0c4c4cce8a4b6', 'perceptual')).toBe(BigInt('-6421916719099894602'));
   });
 
   // `perceptualDct` is ALSO 64 bits, so a width check passes it. `Cosmetic.pHash`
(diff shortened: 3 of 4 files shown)


JustMaier:
Adversarial review done inline (`/code-review high`, prompted to refute, broken-until-proven). All five refutation targets came back **clean**; one low-severity finding, now fixed in `f09f3752da`.

**Refuted, with the reasoning worth keeping:**

- **The money path is not vacuous.** `spendType` is read once and forwarded to `holdPlacementEscrow` (`sticker-placement.service.ts:253`) — no branching, never written to the `Placement` row — so the extra fixture key cannot reach any other assertion. Every other `holdPlacementEscrow` assertion is `objectContaining` or a negative form.
- **I was wrong about the six deleted literals, and it does not change the outcome.** I removed them believing `...placeInput` shadowed them. It did not: the old fixture had no `spendType` key, and a spread only assigns keys it carries, so all six were live and supplied `'yellow'`. The new fixture supplies the same `'yellow'`, so all six sites are value-identical either way.
- `BigInt('-6421916719099894602') === -6421916719099894602n` → `true`, and `BigInt(2) === 2n`; both stay `bigint` for `toBe`.
- `src/shared/constants/basemodel.constants.ts` is a re-export shim for `@civitai/shared/basemodel.constants`, and `~/server/common/constants` imports `BaseModel` from that same shim without re-exporting it. Same declaration, and the file already imports values from the new specifier.
- The baseline diff has **no added entries and no raised counts** — one removal, one decrement, and `801 - 17 = 784` reconciles. `testFilesInProgram` 938 → 1018 is drift catch-up over the 50 commits since the baseline was last written; it moves the positive-control floor **up** (844 → 916), tightening the gate.

**The finding, and the fix.** Before this PR, `placeInput` carried no `spendType` at all, so no fixture edit *could* collide with the `'green'` the escrow test asserts — the non-vacuity was structural. After it, that property rested on a comment. Someone flipping the fixture default to `'green'` (plausible if green Buzz becomes the common case) would leave the test permanently green against a service that had stopped forwarding the caller's currency. Now the two values are named constants and the test asserts they differ:

```ts
const FIXTURE_SPEND = 'yellow' as const;
const CALLER_SPEND = 'green' as const;
...
expect(CALLER_SPEND).not.toBe(FIXTURE_SPEND);
```

**Two positive controls, because a guard that cannot fail is the thing this PR is about:**

1. Hardcoding `spendType: 'yellow'` in the service — dropping the caller's currency — fails exactly one test, legibly: `- "spendType": "green"` / `+ "spendType": "yellow"`, `1 failed | 71 passed`.
2. Colliding the two constants (`FIXTURE_SPEND = 'green'`) fails the new guard, legibly: `AssertionError: expected 'green' not to be 'green'`.

Both mutations reverted; `git status` clean. After the fix: file **72/72**, gate **PASS** (`784 across 140`, baseline `784 across 140`), `pnpm run typecheck` **0 errors**, eslint clean.

PR #3958, opened 2026-08-15
perf(tests): cut the module graph the unit suite imports

JustMaier:
Cuts what the unit suite imports. The suite spends most of its worker time importing modules, not running tests: in the baseline full run `import` was 4,565s against 541s of `tests`.

Every change here is a module that mixed two audiences, so a consumer that wanted the cheap half was paying for the expensive half. All are behaviour-preserving.

## The six edges

- **`~/client-utils/cf-images-utils`** re-exports the pure URL builder *and* defines `useEdgeUrl`, so ~40 server modules importing `getEdgeUrl` also reached `useCurrentUser` → the session provider → the onboarding and buzz-purchase trees. They now import `~/client-utils/edge-url`, where the builder already lived.
- **`server/schema/model-version.schema.ts`** imported `MIN/MAX_DONATION_GOAL` — two numbers — from `components/Model/ModelVersions/model-version.utils`, a hooks module. Moved to `shared/constants/donation-goal.constants.ts`. Closure 605 → 45 modules.
- **`server/schema/training.schema.ts`** imported `blockedCustomModels` from `components/Training/Form/TrainingCommon`. Moved to `shared/constants/training.constants.ts`. Closure 607 → 17 modules.
- **`server/createContext.ts`** held `publicApiContext2`, the only thing in it needing `appRouter` — putting the entire tRPC router (1,053 modules) in the graph of everything that imported the file, most of which only wanted `createContext` or the `Context` type. Split to `server/public-api-context.ts`.
- **`shared/constants/currency.constants.ts`** embeds the icon *components* in `CurrencyConfig`, so its one non-component consumer (`~/utils/buzz`, reached from `buzz.service`) dragged in `@tabler/icons-react`. Colours/css/classNames move to `currency-theme.constants.ts`; `currency.constants` re-attaches the icons on top of that same data, so the 92 UI call sites are untouched and nothing is duplicated.
- **`pages/_app.tsx`** — `getInitialProps` extracted to `~/utils/app-initial-props`, because `app-settings-bootstrap.test.ts` had to import the page to test it and loaded the whole provider/layout tree.

Plus two accidental dependencies on heavy packages: `utils/array-helpers.ts` imported `uniq` from `instantsearch.js` for a function that is one `.filter()` (inlined, same first-index-wins semantics), and `collection.service.ts` imported the youtube client — and therefore `googleapis` — for one mod-only function, now in `collection-youtube.service.ts`.

## Two things worth a reviewer's attention

**`_app.getInitialProps` runs in the BROWSER on client-side route transitions, not only on the server** (`next/dist/docs/.../get-initial-props.md`). The extracted module is a plain static import from `_app.tsx`, which assigns it as `MyApp.getInitialProps`, so it still ships in both the server and client bundles and still runs on client navigations exactly as before. Nothing moved behind a dynamic import, and nothing changed about what is in the client graph — `no-server-infra-in-app-graph` walks the same set either way.

**`currency.constants` is now derived rather than literal.** `CurrencyConfig` is built by spreading `CurrencyThemeConfig` and attaching `IconBolt` (every BUZZ theme) or `IconCurrencyDollar` (USD/USDC) — the same icons each entry had before. `getCurrencyConfig` / `getBuzzCurrencyConfig` keep their signatures and their lookup order, including the `?? themes.yellow` fallback for subscription-kind values that are not currencies.

And note where the saving actually comes from: **`currency.constants` still value-imports the icons**, so splitting the file buys nothing on its own. The load-bearing edit is the one line in `~/utils/buzz` — the only non-component consumer, reached from `buzz.service` — repointing at `currency-theme.constants`. The split exists to give that line something to point at.

## Verification

Full isolated unit run on this branch: **1,065 files, 16,784 tests, 16 failed / 6 files** — identical to baseline, and those 16 are the known Windows path-separator failures that are green on Linux CI. `blocks.router.workflow.test.ts` collected, so nothing silently vanished. `pnpm run typecheck` clean at every commit.

### Timing

Paired full runs on `main` and this branch, same reporter, same window on the same box:

```
  main (control)   collect 6232s   setup 394s   tests 677s   wall 279.4s   16 failed
  this branch      collect 4498s   setup 372s   tests 629s   wall 210.9s   16 failed
                   collect -27.8%                            wall -24.5%
```

**What licenses that aggregate is a control group defined before the fact.** Partition all 1,065 test files by whether this branch changes their static import closure — a predicate computed from the diff, not from the timings — and measure both groups in the same two runs:

```
  CHANGED-closure files    n=412   collect 4843s -> 3778s   -22.0%
  UNCHANGED files          n=653   collect  633s ->  720s   +13.7%   (slower)
```

The control group moved 13.7% *against* the result. That pair was taken on a busier box, which is why its whole-suite wall clock looks flat while the touched files still dropped 22%. A headwind on the control is what makes the changed group's drop hard to explain as box noise.

⚠️ Two other numbers are worth reporting but license nothing, so read them as description only. "306 unchanged files at +0.4%" selects files by whose time did not move, and "234 files improved >20% went 4252s -> 2496s" selects files by how much they improved and then reports how much they improved. Both are circular; only the closure-based partition above is a control.

Individual movers map onto the edges. From the clean pair: `app-settings-bootstrap` 31.5s -> 0.9s, plus a cohort of ~35s files landing at 12-19s — `model-appeal.service` 35.2s -> 11.7s, `orchestration-new.model-substitutions` 35.0s -> 17.1s, `vae-files-cross-version` 34.9s -> 17.5s, `cosmetic-shop-sections.service` 35.7s -> 19.1s. From the busier pair, the same file reads 24.7s -> 0.9s; the "before" differs between pairs because the box did, which is the whole reason a single-run number is not quotable here.

## Also here

`scripts/test-perf/` — `cuts.mjs` ranks import edges by how many modules leave the graph when the edge is cut (dominator tree, rooted at one test or at a synthetic node over all 1,065); `externals.mjs` / `ext-cost.mjs` measure what node_modules packages cost, which no vite-side instrument can see because vite never transforms an externalised dep.

`claudedocs/test-import-slimming-2026-08-15.md` records the method, including the three corrections a static import graph needs before it predicts runtime cost — `import()` is lazy, a `vi.mock` factory cuts its whole subtree, and externals are invisible to a transform hook — and what was deliberately *not* done and why.

---

# 2026-08-15 — split into four; what is left here, and its gate

*Appended by a reviewer, not by the author. Everything above is the original body and is unchanged.*

This PR was held because it is labelled a test-performance change and touches **72 production files**, with its author closed out and nobody able to answer for it. It has now been reviewed file by file at `179128fdf4` (pushed head `6974c12d11`, merge-base `3863adcbb0`). **No runtime behaviour change was found in any of the 72** — every move is either the identical function object reached through a different specifier, or a literal relocated verbatim. That is a static read: nothing was executed to produce it.

The 106-file change decomposes into six independent moves, and the decomposition is proven total and disjoint — applying all six to `main` reconstructs these 106 files byte-for-byte.

**Four are now open separately, all based directly on `main`:**

| | PR | what | prod files |
|---|---|---|---|
| A | #3965 | `getEdgeUrl` re-pointed to the React-free `edge-url` | 41 |
| C | #3968 | icon/theme constants split off the data beside them — **merge last** | 18 |
| D | #3966 | `enableCollectionYoutubeSupport` off `collection.service` | 3 |
| B | #3967 | `publicApiContext2` out of `createContext` | 8 |

**What stays on this branch: E — the `getInitialProps` extraction** (`src/pages/_app.tsx` + `src/utils/app-initial-props.ts` + `src/server/__tests__/app-settings-bootstrap.test.ts`).

The 295 lines were read against the deleted 278 and are verbatim, including the `'session' in data` discriminator, the `settingsDegraded` carve-out, the `liveNow = false` default, the timeout clamp and the greppable failure marker. **A verbatim relocation and a safe one are different claims**, and only the second one matters here: this sits on every SSR page render, and `_app.tsx` is bundled for both client and server, so **its client-bundle graph is validated only by `next build`** — `pnpm typecheck` and both vitest projects pass over a broken one.

**The gate, so this is a one-line decision rather than a re-review:**

1. a clean `next build`, and
2. confirmation that the six imports `_app.tsx` dropped (`~/env/client`, `~/server/auth/route-guard`, `~/server/utils/region-blocking`, `~/server/utils/bot-detection/header`, `~/server/schema/chat.schema`, `~/components/Announcements/announcements-dismissed-cookie`) are all statically picked up by `app-initial-props.ts`.

The reading for (2) is done and says yes. It needs the run, not the reading.

**Also not carried into any of the four:** F — `isMadeOnSite` -> `isImageMetaOnSite` in `post.service.ts`, and the `uniq` inlined into `array-helpers.ts`. Two files, and the inline is the only part of this PR that is new code rather than a move.

One pre-existing issue found while reading and **not** caused by this PR, recorded so it is not later attributed to the move: `chatSettings` is computed in `getInitialProps` and never returned in `pageProps`, while `_app.tsx` still declares, destructures and forwards it. It is `undefined` at runtime on both sides of this diff, so the SSR seed its comment describes is not happening.

### Update — F is now open too, and the build gate has moved

**F is #3971** (`isMadeOnSite` -> `isImageMetaOnSite`, and `uniq` inlined). Approved to ship both parts. It is the only one of the five that is **new code rather than a move**, which is why it is its own PR — see its body.

So all five splits are open: **#3965 (A), #3968 (C), #3966 (D), #3967 (B), #3971 (F)**. Nothing but **E** remains on this branch.

**A, C, D and B were verified in one box tenancy:** typecheck 0 errors on all four; full unit suite on C, D and B at 1066 files / 16809 tests, with the suite's 16 failures matched **per file** against a control taken on `main` in the same window (`1/11, 1/5, 1/6, 3/27, 7/9, 3/12` — six files, none of them touched by any split). Zero failures introduced. **#3971 (F) has had no run of its own** and wants typecheck plus the unit suite before it merges.

⚠️ **E's gate needs restating, because a clean `next build` is not reachable on a dev worktree.** Measured on this branch's sibling B:

- at Node's default heap, `next build` **OOMs during type-checking on `main` itself** — `package.json` runs a bare `next build` (~4 GB default) while `scripts/typecheck.mjs` gives the same check 8192 MB. That mismatch is a pre-existing repo issue, unrelated to any branch here.
- at `NODE_OPTIONS=--max-old-space-size=8192` it reaches **`✓ Compiled successfully in 3.2min`**, then fails in `Collecting page data` on `DATABASE_URL` / `DATABASE_REPLICA_URL` / `REDIS_URL` — page-data collection evaluates server modules and wants live infrastructure.

**`✓ Compiled successfully` is the phase that validates the client bundle graph**, which is the thing E actually risks and the reason it was held. Completing the build past that point needs an environment, not a code change — so the gate should be *reaches `✓ Compiled successfully` at a raised heap*, with the full build left to CI where the env exists.


JustMaier:
Heads-up from the closure-rebuild work: **this branch's `scripts/test-perf/graph.mjs` copy should be dropped when this rebases onto merged main.**

The mock- and lazy-honouring graph fix now lives on `perf/test-perf-tooling` (#3957) as `da82a92c5c`, which is where `graph.mjs`, `cuts.mjs` and the README actually live and which merges first. This branch carries a copy that diverged earlier (27 lines lighter — it deleted the `closures.json` writer), so keeping it after #3957 lands would put a third version of the file in play and silently un-fix any consumer that reads `inventory.json` or `closures.json` from this base.

`cuts.mjs` and the README changes stay here: `cuts.mjs` does not exist on #3957 at all, and the two READMEs have diverged too far to move wholesale.

What the fix does, so the drop is an informed one: `graphModules` becomes the count a worker really loads — lazy `import()` not followed except from a test file itself, `vi.mock` factory subtrees truncated, `import type` / `import { type X }` erased — and `closures.json` is tagged `mode: 'real'` so a consumer can refuse a naive one. The naive count over-stated a page-gate test by ~75x. Details in the #3957 commit message.

JustMaier:
Closing this as part of winding down the unit-suite unisolation lane.

## Why closed rather than merged

**The body says group E only; the branch still carries all 106 files.** Its other five groups were split out and merged separately — #3965 (`getEdgeUrl` re-point), #3966 (`collection-youtube`), #3967 (`publicApiContext2`), #3968 (constants extraction) and #3971 (the on-site predicate). Merging this branch as it stands would re-land work already on `main`, plus the one group that was deliberately held.

**Group E was held on a condition it does not meet.** The ruling at the time was *"if all of it's just fixing import cycles and bullshit like that, I'm fine with us making those changes"* — true of A/B/C/D, not of E. E is a 278-line relocation of `_app.getInitialProps` that runs on every SSR render, and its client-bundle graph is validated by **`next build` alone**: `pnpm typecheck` and both vitest projects pass over a broken one. The gate agreed for it was a clean `next build` plus per-file confirmation that the six imports `_app.tsx` drops are statically picked up by `app-initial-props.ts`. That gate was never run.

## Why not just run the gate now

The lane this belonged to is stopped. Measurement showed the unisolation benefit is convex — 43% of test files bought 3% of the import saving, 100% buys 66% — so a partial migration is worth ~nothing and finishing costs 601 files across 354 specifiers. `unit-fast` (#3975) was closed rather than merged.

That removes E's justification entirely. E was never a test-speed change; it is a production change that arrived under a test-perf label, which is why it was held in the first place. With no campaign behind it, it should not ride in on one.

**If the import-cycle fix has standalone value it deserves its own PR**, scoped to `src/pages/_app.tsx`, `src/utils/app-initial-props.ts` and `src/server/__tests__/app-settings-bootstrap.test.ts`, gated on a clean `next build`. The branch is left in place so that work is recoverable.

## What stays on `main`

Everything this lane actually produced: the canonical mock system, guard, allowlist and codemod; 264 migrated files; the RSA keypair cache; #3957/#3960/#3961/#3962/#3972/#3973; and the five import-slimming groups above. Only the cancelled migration's remainder goes.

PR #3962, opened 2026-08-15
perf(tests): cut test-body time in the five slowest unit-suite files

JustMaier:
Cuts test-body time in the slowest files of the unit suite. **The branch name is from a superseded plan — this is not a mock migration.** Nothing here touches the mock system, import slimming or pool config; those are other branches.

## Why this lane exists

Import time is attacked elsewhere. This is the cost no worker count can touch: the test bodies. Measured at `3863adcbb0`, the full suite spent **564 worker-seconds** in test bodies, and the top 20 files held 50.5% of it. A single 37.9s file is also a wall-clock floor — the suite cannot finish faster than its slowest file.

## What changed

**Two "before" columns, because they disagree and only one of them is a paired measurement.**

| # | file | baseline `duration` | measured before | after | mechanism |
|---|---|---|---|---|---|
| 1 | `challenge-ladder.test.ts` | 37875ms | 35731ms | **427ms** | deferred import of a heavy module |
| 2 | `prisma-inconsistent-orphan-relations.test.ts` | 31963ms | 19148ms | **9ms** | unmocked import graph that seven sibling suites already stub |
| 3 | `challenge-category.service.test.ts` | 18055ms | 18400ms | **39ms** | a real TCP timeout standing in for a mocked failure |
| 4 | `audit-matching-equivalence.test.ts` | 17491ms | 10152ms | *(unchanged)* | **irreducible** — brute-force differential oracle |
| 5 | `s3-upload.store.test.ts` | 16418ms | 16358ms | **36ms** | real retry backoff in the code under test |
| 6 | `listing-asset-upload-integrity.test.ts` | 14513ms | 6992ms | **773ms** | deferred import — **relocation, not a cut** |
| 7 | `dev-server-port-reservation.test.ts` | 10607ms | 10600ms | **734ms** | two tests paying a 5s claim deadline in full |
| 8 | `base.reward.failsoft.test.ts` | 3832ms | 3837ms | **1270ms** | real 5 x 500ms retry backoff |

- **baseline `duration`** — this file's row in the shipped full-suite artifact (`baseline-full-uncapped`, `head: 3863adcbb0`), measured under full-suite parallelism.
- **measured before** — a named-file run I took immediately before the change, in the same window as its "after".

**Only the second and third columns are a pair.** The two "before" columns disagree by up to 2.1x (#6: 14513 vs 6992), because a file measured under 31-way contention is not the same measurement as one measured alone. Quote the paired columns; the baseline column is here to locate the file in the suite, not to compute a saving from.

**Measured in-window: ~107 seconds off the test-body floor — ~101s a real cut, ~6s relocated.** Those two are kept apart deliberately. Against the baseline column the same work reads as ~110s, and that larger number is the one *not* to use: its two halves come from different windows on a box whose drift has since been measured at **+20.5% on `collect` between two identical full runs**.

The suite's slowest-file floor drops from 37.9s to 18.1s on the baseline's own terms, which is the one comparison where using that column is right — it is baseline-vs-baseline.

#6 moves cost from `duration` into `collect` (561ms → 6140ms) and leaves wall clock unchanged within this box's noise. It ships anyway because a graph billed to `duration` is invisible to the lane attacking import cost — the file reads as cheap to import and expensive to run when the opposite is true, so leaving it mislabelled hides a real target.

**Why #2's trick was not repeated on #6**, since the two files otherwise look alike: #2's graph was cut with a mock scaffold seven sibling suites already used, and nothing in that file exercised what the scaffold replaced. #6 is a seam test whose header is explicit that only the backend accessor is stubbed — the real persist proc, the real `sharp` decode, the real head re-read and the real comparison all run, because each half is trivially correct alone while the pair does nothing if they disagree about the metadata key. Cutting that graph would buy seconds by removing the thing the file exists to check.

🔴 **An earlier draft of this PR said "every before/after pair was taken in the same window" while the table's before column came from the baseline artifact and the after column from my own runs. That sentence was wrong**, and it is the same defect I raised against three other lanes — a control and a treatment from different windows, presented as a pair. The table above now carries both columns and says which two are comparable.

No full-suite run was taken for this branch. Every "after" is a named-file run, each paired with a "measured before" taken minutes earlier in the same session. That pairing is what the numbers rest on; the baseline column is context.

## The finding that outlives the diff

**The tail has six causes, not one.** Five are fixable, one is a floor. Anyone planning further work here should not expect a sweep.

1. a deferred import of a heavy module (#1, #6, and most of what remains)
2. an unmocked import graph the sibling suites all stub (#2)
3. a real network timeout substituted for a mocked failure (#3)
4. real retry or backoff sleeps in the code under test (#5, #7, #8 — three instances, the most common fixable cause)
5. real `spawnSync` subprocesses (`typecheck-tests-gate`)
6. whole-source-tree scans, several files each rebuilding the same one — **the only cause wanting a shared fix rather than a per-file one**

And beneath them the floor: differential oracles and scaling guards, where the time **is** the assertion (#4, `audit-redos`).

**A dynamic `await import()` inside a test body bills the whole module graph to `duration`, not `collect`.** The file then reads as cheap to import and expensive to run, when it is exactly the opposite — and it is invisible to anyone attacking import time. `challenge-ladder` was `collect 476 / duration 37875`; 35273ms of that was one test importing `daily-challenge-processing` to read a single job option.

Scanning the baseline for that shape — a dynamic import inside **any** function body rather than at module top level, excluding `vi.mock` factories, which are normal — over the top 50 slowest bodies with the six files touched here removed:

```
remaining top-50 files:            44 files, 281.3s of test-body time
  carrying a deferred import:      31 files, 202.3s   (72%)
  genuinely slow bodies:           13 files,  79.0s
```

Read as an **upper bound on a class**, not a measurement: it says a file with a deferred import holds N seconds, not that N seconds is import cost. It is a strong indicator rather than a proof — but where it was checked, it held. In `challenge-ladder` the first test held 35273ms of 37875ms; in `app-listing-backfill` 5765ms of 5776ms; in `run-page-maturity` 4364ms of 4369ms.

Three caveats travel with the number:
- vite-node caches repeated `await import()` of the same module within a file, so per-file hit counts index shape, not cost. `app-listing-backfill` imports the same service in 10 separate tests and pays once.
- Hoisting is not always safe. A test that imports late to observe module-scope initialisation, or after mutating env, changes meaning when moved. Judge each one.
- A first pass at this scan looked only for imports lexically inside `it`/`beforeAll` and reported 28/50. It **undercounted**, because the common shape routes the import through a plain helper — `async function loadResolver() { await import('~/pages/…'); }` — called from the tests. Anything measuring this class needs the by-depth criterion, not the by-construct one.

**Three of the first four fixed files also had a coverage defect underneath.** This was not a coincidence:

- **#3** could not distinguish the branch it was named for. Flipping its stub from rejecting to resolving-empty left all 16 tests green, because merging presets with an empty row set gives the same answer as the preset fallback. Fixed by pinning the behaviour that actually differs — the failure path deliberately does not cache. 16 tests → 18.
- **#5** had *no* cover for the transient part-retry path. Making 429 and 5xx non-retryable left all 17 tests green. It went untested because testing it meant sitting through 1s + 2s + 4s + 8s of real backoff — the cost *was* the gap. 17 → 19, plus `getPartRetryDelay` gaining 7 unit tests where it previously had none.
- **#1**'s slow test welded two claims together: arithmetic (`lock < interval`) and wiring (`the job asks for that lock`). Only the wiring half needed the graph.

So: **slow tests here are usually slow because something real was substituted for something that should have been controlled, and an uncontrolled dependency is an untested one.** Every fix in this PR raised coverage or held it level; none traded coverage for speed.

## How each one was fixed

**#1** — split the test. The wiring half moved to `challenge-jobs-scale.test.ts`, which already loads `daily-challenge-processing` behind a mock preamble, so it lands at ~0 marginal cost. Test count preserved: 40 + 6 where it was 40 + 5.

**#2** — the header claimed `@prisma/client` could not be stubbed under the unit env, so it installed a Proxy and let the real `model.service` graph load from a `beforeAll` with a 120s timeout. Seven sibling suites import the same module *statically* behind a mock scaffold and pay 11.9–33.0s in `collect`; this file paid 32s in `duration`. Took the sibling scaffold, dropped the Proxy, hoisted every dynamic import. A real cut, not a relocation: wall 19.6s → 7.2s.

**#3** — `~/server/db/client` was left unmocked so a real connection attempt would fail and exercise the preset fallback, at ~4s per call. Stubbed the read to reject instead: same `catch`, deterministic, and no longer dependent on what the box can reach.

**#4** — **deliberately not optimised.** `refIncludesPoi` runs a `String.match` before `regex.test` for every word, so non-matching words pay both; reordering would roughly halve two tests. An oracle's value is that it is a *copy* — an optimised copy is a second implementation you believe is equivalent, which is the one thing a differential test exists to not have to trust. The reasoning is now a comment at the head of the reference section, where the next person meets the lever before taking it.

**#5** — a `runUpload()` helper advances a fake clock until the upload settles.

Worth naming so a reader does not assume otherwise: the target was **not** the exponential part backoff. That path contributed nothing, because no test reached it. All six slow tests were the flat `withRetries(…, 3, 1000)` on the completion path — 3s each.

🔴 The first version drained `while (!settled && vi.getTimerCount() > 0)` and **hung**: at the moment the upload is handed back no timer exists yet — the first is scheduled after `fetch('/api/upload')` resolves — so the guard read 0 and exited before the run began. A count of pending timers says nothing about whether more are coming. The shipped version advances unconditionally, is bounded, and throws.

That bound is what makes a fake clock safe here rather than dangerous. Driving a loop with a fake is normally how a regression becomes an unreportable hang — the runner's own timeout is a timer too. Setting `MAX_PART_ATTEMPTS` to 100000 so the retry loop never ends now produces:

```
× stops at MAX_PART_ATTEMPTS when a part is transient forever ... 113ms
   Error: upload did not settle within 200 clock turns — a retry loop is not terminating
```

113ms and a sentence, where the same regression against the real clock would have wedged CI with nothing to read.

## Mutation tests

Every change was verified by breaking the production behaviour and reading the actual failure, not just confirming red.

| mutation | resulting failure |
|---|---|
| drop the job's `lockExpiration` override | `expected 300 to be 540` (createJob's default vs `9*60`) |
| `REVIEW_JOB_LOCK_SECONDS` 9→11 min | `expected 660 to be less than 600` |
| drop `modelVersion: { is: {} }` from the where-clause | `expected undefined to deeply equal { is: {} }` |
| drop `where: { tag: { is: {} } }` from `articleDetailSelect` | `expected { select: { tag: … } } to match object { where: { tag: { is: {} } }, … }` |
| catch returns `[]` instead of the presets | `expected [] to include 'theme'` + `TRPCError: Unknown judging category: theme` |
| cache the failed read | `expected "vi.fn()" to be called 2 times, but got 1 times` |
| `withRetries(…, 3, 1000)` → 0 attempts | `expected 1 to be 2` |
| `MAX_PART_ATTEMPTS` 5 → 1 | `expected [] to deeply equal [ 'https://b2.test/upload?part=2&v=1' ]` |
| 429/5xx no longer retryable | `expected 1 to be 2` and `expected 1 to be 5` |
| `MAX_PART_ATTEMPTS` → 100000 | `upload did not settle within 200 clock turns` |
| drop `session.port = moved` (daemon) | `expected 3101 to be 3102` |
| `findAvailablePort` ignores other sessions' reservations | `expected 3100 to be 3102`, `expected 3100 to be 3103` |
| the reward batch path swallows instead of rethrowing | `promise resolved "undefined" instead of rejecting` |
| `updateBuzzEvents` retry literal 5 → 1 | `expected "vi.fn()" to be called 6 times, but got 2 times` |

🔴 **One of these corrected a false claim I had just written.** In `base.reward.failsoft` I added `expect(h.insertImpl).toHaveBeenCalledTimes(6)` and commented it as pinning `BATCH_RETRY_COUNT`. Mutating that constant 5 → 1 left the file **green** — the path runs through `updateBuzzEvents`, which passes `5, 500` to `withRetries` as a literal and never reads the constant. The assertion was right; the sentence about it was wrong, and it would have credited a constant with coverage it does not have.

It was caught only by mutating the constant I had **named** rather than the one I had **pinned**. Both results are recorded in the test's comment. Worth generalising: a mutation that fails legibly still only proves the mechanism you actually broke.

## Verification

- `scripts/__tests__/typecheck-tests-gate.test.ts` — 147/147 (`typecheck` does not see `__tests__`)
- `pnpm run test:lint-rules` — 220/220, including `no-wholesale-module-mock`, which #2 improves by dropping its `@prisma/client` Proxy
- named-file runs throughout; no full-suite run was taken for this branch

## One named-file run tells you whether a slow file is fixable

Files 7 and 8 came out of the 13 in the top 50 that had **no** deferred import — the "genuinely slow body" set. Measuring all 13 alone produced a filter worth more than the two fixes:

```
baseline (31-way contention)      solo
        11133  lazy-revision  ->   6407
        10607  port-reserv.   ->  10600   <- unmoved
         5817  placement      ->   3452
         5542  model-subst.   ->   3267
         5458  audit-redos    ->   2550
         5450  scoped-cors    ->   2375
         4887  block-scope    ->   4471
         4015  dev-tunnel     ->   2274
         3991  hostTheme      ->   1950
         3832  failsoft       ->   3837   <- unmoved
         3702  gen-surface    ->   1165
         3548  chal.metrics   ->   1747
```

Most of the tail is **contention, not work** — solo they land at 40-60% of their baseline row, which means several are in the top 50 partly because they were unlucky in that run's scheduling.

🔴 **A file whose solo time equals its contended time is waiting on a clock, not competing for CPU.** The only two that did not move are the only two that were fixable, and that is the filter working rather than a coincidence: a wall-clock wait does not get shorter when the box is quiet. It costs one named-file run, and it separates the fixable from the merely unlucky before any code is read.

### The eleven left alone, and why

- **`audit-redos`** — asserts that matching scales *linearly* across adversarial inputs. The time is the assertion. Irreducible, same as #4.
- **`scoped-endpoints-cors-wiring`, `generation-surface-wiring`, `hostThemeChangeParity`, `block-scope.anytoken-mode`** — whole-source-tree scans: each walks `src` and regexes every file, and **each rebuilds the same scan independently**. See the lead below.
- **The rest** — 1.2-3.5s solo, spread across their tests with no dominator. Grind, not a mechanism.

## Lead, not taken: the four guard files that each rebuild the same source-tree scan

`scoped-endpoints-cors-wiring`, `generation-surface-wiring`, `hostThemeChangeParity` and `block-scope.anytoken-mode` are convention guards that each walk `src` and regex every file. Solo they are 1.1-4.5s each; in the baseline they are 3.7-5.5s. `app-spend-tier-privilege` does the same walk again, and so does `orchestrator-chat.wait-unit`.

A shared, cached tree scan behind one helper is the only real lever left in this tail — worth roughly **10-15s of test-body time**, and it is the sixth mechanism, the only one that wants a shared fix rather than a per-file one.

**Deliberately not attempted here.** It is a cross-file refactor touching four guard files whose whole value is being loud about drift, for a few seconds. That trade wants a reviewer and an unhurried window, not the end of a long day. Whoever takes it should note that a cached scan introduces cross-file coupling into guards that are currently independent — which is exactly the property that makes them trustworthy.

## Where this lane stops, and why

Working down the ranked list stopped at the next deferred-import file, because it measured as the same relocation and the scan says most of what remains is too:

```
app-listing-backfill.service.test.ts   17 tests, 5776ms — 5765ms of it on the first test,
                                       which is the one that first imports the service
```

Grinding out three more relocations would move seconds between columns and produce no wall-clock change. **The remaining tail is 72% deferred-import cost wearing a disguise — a finding for the import-slimming lane, not a fix for this one.** The genuinely-slow-body set was then swept separately, which is where files 7 and 8 came from.

## Not done, and deliberately

- **The `block-registry.service` cluster** — the same module imported inside test bodies across 12 files, ~53s together. It is import-graph work and belongs to that lane, not this one.
- **`typecheck-tests-gate.test.ts` (10982ms)** — real `spawnSync` subprocesses, left alone for timing. It was checked for correctness first, because a gate that reports a clean result for a process that never ran is a worse problem than a slow one.

  **It is clean, and structurally so.** The gate spawns `process.execPath` — always present — and passes the wrapper as an *argument*, so there is no resolved-binary lookup to fail silently and an `ENOENT` can never land in the status field. Verified against the real gate with a wrapper path that does not exist: `error= undefined`, `status= 3`, `CANNOT MEASURE`. It is also defended twice — with the FAILURE-TO-RUN rule disabled the positive-control floor still refuses the same run (`0 file(s) under a __tests__ directory, below the floor of 844`).

  Nothing pinned that, so this PR adds the case. It asserts the run is **refused**, not *which* guard refuses it: both catch it, and pinning one mechanism would go red on a change that left the property intact.

All 13 of the top-50 "genuinely slow body" files were measured; two are fixed here and the rest are accounted for above. `typecheck-tests-gate` is the one left unexamined for timing, for the reason given above.

### DIFF
--- a/scripts/__tests__/dev-server-port-reservation.test.ts
+++ b/scripts/__tests__/dev-server-port-reservation.test.ts
@@ -39,11 +39,21 @@ import {
   sessions,
 } from '../../.claude/skills/dev-server/scripts/daemon.mjs';
 
-// A 1ms poll with a generous deadline keeps every case below count-bound rather than clock-bound:
-// the probe mocks decide the outcome, so a slow box cannot turn a pass into the very failure these
-// tests exist to detect. The scan base sits above the default 3000 so no case can be satisfied by
-// a port a real session would be using.
+// A 1ms poll with a generous deadline keeps the cases where the port FREES count-bound rather than
+// clock-bound: the probe mocks decide the outcome, so a slow box cannot turn a pass into the very
+// failure these tests exist to detect. The scan base sits above the default 3000 so no case can be
+// satisfied by a port a real session would be using.
 const FAST_CLAIM = { timeoutMs: 5000, intervalMs: 1, baseDevPort: 3100 };
+
+// The cases where the port NEVER frees are a different shape, and the comment above used to claim
+// them too. They cannot be count-bound: claimPortForReuse polls until its deadline and only then
+// moves, so expiry IS the mechanism under test and the deadline is paid in full. At 5000ms the two
+// of them cost 10.0s of the file's 10.6s — measured, and unmoved by running the file alone, which
+// is what distinguishes a real wait from contention. 60ms buys the same expiry.
+//
+// Shortening it cannot make these flaky: the assertion is that the session MOVED, and the move is
+// what happens once the deadline passes however few probes fit inside it.
+const HELD_CLAIM = { ...FAST_CLAIM, timeoutMs: 60 };
 
 const ENV_PATH = '/nonexistent/.env'; // only read by start(), which no test here calls
 
@@ -149,7 +159,7 @@ describe('dev-server port claim on reuse', () => {
     makeSession('neighbour', 3100).status = 'running';
     isPortFreeMock.mockImplementation(async (port: number) => port !== 3101);
 
-    expect(await claimPortForReuse(session, FAST_CLAIM)).toBe(3102);
+    expect(await claimPortForReuse(session, HELD_CLAIM)).toBe(3102);
     expect(session.port).toBe(3102);
     expect(session.logs.at(-1)?.message).toContain('moving this session to 3102');
   });
@@ -162,7 +172,7 @@ describe('dev-server port claim on reuse', () => {
     makeSession('above', 3102).status = 'crashed';
     isPortFreeMock.mockImplementation(async (port: number) => port !== 3101);
 
-    expect(await claimPortForReuse(session, FAST_CLAIM)).toBe(3103);
+    expect(await claimPortForReuse(session, HELD_CLAIM)).toBe(3103);
   });
 });
 
--- a/scripts/__tests__/typecheck-tests-gate.test.ts
+++ b/scripts/__tests__/typecheck-tests-gate.test.ts
@@ -752,6 +752,26 @@ describe('the gate, end to end: a broken checker must never PASS', () => {
     expect(res.all).not.toContain('PASS —');
   });
 
+  // A wrapper that NEVER LAUNCHES, which is a different failure from one that runs and exits
+  // non-zero. It is worth its own case because the neighbouring class bit us elsewhere: an
+  // `execFile` on an extensionless `node_modules/.bin` script never started on Windows, and the
+  // ENOENT landed in the same `code` field as a real exit status, so a process that never ran
+  // reported as one that exited 2.
+  //
+  // The gate is structurally immune to that shape rather than defended against it: it spawns
+  // `process.execPath` — always present — and passes the wrapper as an ARGUMENT, so a missing
+  // wrapper is node's own exit 1 with no diagnostics, which the FAILURE-TO-RUN rule already owns.
+  // There is no resolved-binary lookup anywhere in this path to fail silently. This pins that.
+  it('a wrapper that does not exist -> exit 3, never a pass', () => {
+    const res = runGate(
+      path.join(dir, 'no-such-wrapper.mjs'),
+      fixtureBaseline('b-missing', BASE_FILES)
+    );
+    expect(res.status).toBe(3);
+    expect(res.all).toContain('CANNOT MEASURE');
+    expect(res.all).not.toContain('PASS —');
+  });
+
   it('a run killed by a signal -> exit 3', () => {
     const res = runGate(
       wrapperStub('killed', 'process.kill(process.pid, "SIGKILL");'),
--- a/src/server/games/daily-challenge/__tests__/challenge-ladder.test.ts
+++ b/src/server/games/daily-challenge/__tests__/challenge-ladder.test.ts
@@ -355,16 +355,10 @@ describe('findSlot seating', () => {
 describe('the per-tick placement budget fits the job lock', () => {
   // The bound this exists for is the JOB lock, which is not the completion claim and is not the
   // cron interval. Getting the three confused is how a cap gets sized against the wrong number.
-  it('is the lock the job actually asks for, not createJob’s default', async () => {
-    const { createJob } = await import('~/server/jobs/job');
-    const inherited = createJob('probe', '*/10 * * * *', async () => undefined);
-    const { dailyChallengeJobs } = await import('~/server/jobs/daily-challenge-processing');
-    const job = dailyChallengeJobs.find((j) => j.name === 'daily-challenge-process-entries')!;
-
-    expect(job.options.lockExpiration).toBe(REVIEW_JOB_LOCK_SECONDS);
-    // The point of the override: the inherited default is too short to hold a drain.
-    expect(REVIEW_JOB_LOCK_SECONDS).toBeGreaterThan(inherited.options.lockExpiration);
-    // Still shorter than the interval, so a tick can never overlap itself.
+  // The other half of this — that the JOB actually asks for REVIEW_JOB_LOCK_SECONDS — lives in
+  // `src/server/jobs/__tests__/challenge-jobs-scale.test.ts`, which already loads
+  // daily-challenge-processing.ts. Importing that graph here cost 35s of test-body time on its own.
+  it('is shorter than the interval, so a tick can never overlap itself', () => {
     expect(REVIEW_JOB_LOCK_SECONDS).toBeLessThan(REVIEW_JOB_INTERVAL_SECONDS);
   });
 
--- a/src/server/jobs/__tests__/challenge-jobs-scale.test.ts
+++ b/src/server/jobs/__tests__/challenge-jobs-scale.test.ts
@@ -139,9 +139,11 @@ vi.mock('~/server/jobs/daily-challenge-processing', async (importOriginal) => {
   return { ...actual, pickWinnersForChallenge: mockPickWinnersForChallenge };
 });
 
-const { reviewEntries, challengeReviewInternals } = await import(
+const { reviewEntries, challengeReviewInternals, dailyChallengeJobs } = await import(
   '~/server/jobs/daily-challenge-processing'
 );
+const { createJob } = await import('~/server/jobs/job');
+const { REVIEW_JOB_LOCK_SECONDS } = await import('~/server/games/daily-challenge/challenge-ladder');
 const { runChallengeCompletion } = await import('~/server/jobs/challenge-completion');
 const { CHALLENGE_JOB_BATCH_SIZE, CHALLENGE_JOB_CONCURRENCY } = await import(
   '~/shared/constants/challenge.constants'
@@ -172,6 +174,20 @@ beforeEach(() => {
   mockGetEndedActiveChallenges.mockResolvedValue([]);
   mockGetChallengesToReconcile.mockResolvedValue([]);
   mockResetStuckCompletingChallenges.mockResolvedValue(0);
+});
+
+describe('the review job asks for the lock the placement budget was sized against', () => {
+  // The bound REVIEW_TICK_BUDGET_MS is sized against is the JOB lock, not the completion claim and
+  // not the cron interval. The arithmetic lives beside those constants in challenge-ladder.test.ts;
+  // this is the wiring half, here because loading daily-challenge-processing.ts is what it costs.
+  it('is the lock the job actually asks for, not createJob’s default', () => {
+    const inherited = createJob('probe', '*/10 * * * *', async () => undefined);
+    const job = dailyChallengeJobs.find((j) => j.name === 'daily-challenge-process-entries')!;
+
+    expect(job.options.lockExpiration).toBe(REVIEW_JOB_LOCK_SECONDS);
+    // The point of the override: the inherited default is too short to hold a drain.
+    expect(REVIEW_JOB_LOCK_SECONDS).toBeGreaterThan(inherited.options.lockExpiration);
+  });
 });
 
 describe('reviewEntries at volume', () => {
--- a/src/server/rewards/__tests__/base.reward.failsoft.test.ts
+++ b/src/server/rewards/__tests__/base.reward.failsoft.test.ts
@@ -233,7 +233,15 @@ describe('base.reward batch process() still rethrows (cron path unchanged)', () 
     });
   }
 
+  // The batch path retries 5 times with a 500ms backoff, all of it real: this one test cost 2556ms
+  // of the file's 3837ms. The retry budget IS what is under test here — shrinking it would change
+  // the behaviour rather than the clock — so the clock is driven instead.
+  //
+  // The loop is bounded and the assertion is awaited afterwards, so a retry chain that stopped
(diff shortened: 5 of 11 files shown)


JustMaier:
Review as asked, from the lane with no investment in this one. Short version: I could not break the numbers, and the two things I have are a latent hazard and a naming nit. I checked the claims I could falsify independently rather than re-reading the body.

## What I verified rather than took on trust

**#4 really is unchanged.** 46 changed lines on a file whose row says *(unchanged)* is the sort of thing that usually hides a quiet edit. It is the added `🔴 Do not optimise` comment plus Prettier reflowing two `it(..., 60000)` calls and one regex onto fewer lines. No assertion, corpus or reference-implementation change. The row is honest.

**The moved test in `challenge-jobs-scale.test.ts` reads production, not a fixture.** This was my main worry with #1: the file `vi.mock`s `~/server/jobs/daily-challenge-processing` with an `importOriginal` spread, and the relocated assertion reads `dailyChallengeJobs` from that same module. If the factory had overridden it, `expect(job.options.lockExpiration).toBe(REVIEW_JOB_LOCK_SECONDS)` would be comparing the fixture to itself. It overrides `pickWinnersForChallenge` only, so `dailyChallengeJobs` comes from the real module. The move is sound.

**The `runUpload` bound is real.** `MAX_CLOCK_TURNS = 200` × `CLOCK_TURN_MS = 60_000` with an unconditional advance, and it throws with a sentence rather than returning. The comment explaining why `while (vi.getTimerCount() > 0)` hung — no timer exists at hand-back, the first is scheduled after `fetch('/api/upload')` resolves — is the kind of thing that gets rediscovered expensively if it is not written down.

## One latent hazard worth a line of comment

`runUpload` creates the promise, awaits the clock loop, and only returns the promise afterwards:

```js
const promise = useS3UploadStore.getState().upload(...args).finally(() => { settled = true; });
for (let turn = 0; !settled && turn < MAX_CLOCK_TURNS; turn++)
  await vi.advanceTimersByTimeAsync(CLOCK_TURN_MS);
```

If `upload` ever **rejects**, that rejection is unhandled for the duration of the loop — it would surface as an unhandled rejection rather than as the test failure you want. It is **not live today**: I checked every call site in the file and `upload` resolves on every path, returning `undefined` on failure and setting `items[0].status = 'error'`. So this is not a defect in the PR; it is a tripwire for whoever later makes `upload` reject. A `.catch(() => {})` on the loop-local handle, or one sentence saying the helper assumes a non-rejecting `upload`, keeps the next person out of it.

## Naming nit

`describe('preset fallback (no DB available)')` now contains `it('caches a successful read, and lets its rows override the presets')`, where the DB read **succeeds**. The test is right and belongs near its sibling — the block name is now narrower than its contents.

## On the correction, since it is the reason I was asked

The body carries the before/after basis defect and its fix in the open, with both columns present and an explicit statement of which two are comparable. I looked for the same defect elsewhere in the body and did not find it: the `37.9s → 18.1s` slowest-file claim is baseline-vs-baseline, which is the one place that column is the right one, and it says so. The `72%` deferred-import figure is labelled an upper bound on a class, carries three caveats including the by-depth-vs-by-construct undercount that produced an earlier wrong number, and names the three files where it was checked against actuals.

The `~110s` figure is correctly identified as the one **not** to use. For what it is worth from my lane, that is now measurable rather than a guess: two identical full runs on this box, nothing changed between them, came out **+20.5% on `collect`**. Anything assembled from two windows is inside that envelope.

## Not blocking, for a future pass

`typecheck-tests-gate.test.ts` (10982ms) is correctly deferred, and the stated reason — check the `spawnSync` calls for the laundering bug where a process that never launched reports as one that exited non-zero — is the right order of operations. A gate that reports a clean result for a process that never ran is a worse bug than a slow one.

Nothing here blocks. The two items above are a comment and a `describe` name.

PR #2520, opened 2026-06-13
fix(app-blocks): HMAC replay protection — ts tolerance + callback pending-run cross-check + delivery nonce (W11 F5)

ZacxDev:
## W11 audit finding F5 — HMAC replay protection (civitai half)

Closes the **F5** gap from the App Blocks W11 security audit (`datapacket-talos/claudedocs/app-blocks-w11-audit-findings.md`): none of the three build-chain HMAC contracts had a timestamp tolerance or per-delivery nonce, and build-callback had no pending-run cross-check, so a captured validly-signed request could be replayed (downgrade/rebuild of a prior already-approved sha).

**Stacks on #2518** (`zach/w11-f6-hmac-dual-secret`) — these edits build on the F6 dual-secret verifiers and do **not** weaken the `timingSafeEqual` HMAC checks. If #2518 merges first, rebase this onto `feat/app-blocks-main-v1`.

A **companion datapacket-talos PR** implements the other side of the wire contract: the `ts` SIGNER on the callback Tekton task and the `ts` RECEIVER (skew validation) in `trigger.py`.

### WIRE CONTRACT (must match the datapacket-talos side exactly)

- **tekton-trigger leg (this PR = SIGNER):** `triggerBuild` adds integer `ts = floor(Date.now()/1000)` to the JSON body **before** HMAC signing. `trigger.py` validates `|now - ts| > 300` → reject.
- **build-callback leg (this PR = RECEIVER):** after the existing HMAC verify, parse `ts` from the verified body and reject `401` when `ts` is present AND `|nowSec - ts| > 300`. If `ts` is **absent → ALLOW** (rollout tolerance; the datapacket-talos callback-task signer adds `ts` in its own PR — enforce-if-present avoids a deploy-ordering outage). `ts` is inside the signed body, so a replay carries its original stale `ts` and is caught.
- **callback pending-run cross-check (entirely civitai-side):**
  - `triggerBuild`: after a successful trigger POST, write Redis `system:blocks:callback:pending:<slug>:<sha>:<appBlockId>` = `"1"` TTL **1800s**. Fire-and-forget — a Redis failure logs + continues (never fails the trigger).
  - `build-callback`: after HMAC + ts pass, **GETDEL** that key (consume-once). Missing → reject `409` when `BLOCK_CALLBACK_REQUIRE_PENDING_RUN === 'true'`, else warn + continue (report-only first deploy). Redis error → treated as not-enforced.
[...]

PR #3973, opened 2026-08-15
perf(tests): services a-m onto the canonical shared mocks (127 of 129)

JustMaier:
Base is **`perf/test-mock-system`**, deliberately — not `main`. This is slice work for #3959, and it wants its own review rather than arriving inside a PR that was reviewed at a different scope.

## What this is

The `src/server/services/__tests__` a–m slice migrated onto the canonical shared mocks: **127 of 129 files**, from 129 that ever carried a direct canonical mock. Started by josh, continued and closed out here.

**2 remain, both genuinely blocked** on `TEST_ENV_DEFAULTS` being seeded from the schema's own defaults — `model-version.blue-buzz-purchase` and `model-version.purge-by-hash` reach `getDbWithoutLag`, and `REPLICATION_LAG_DELAY` is absent from the defaults, so `undefined <= 0` is `false` where `0 <= 0` is `true`. Verified against the base rather than inherited from the handover.

## How each batch was verified

Every batch: a control and a candidate **in one box tenancy**, per-file collected counts on both halves, a `cmp` gate before the control half proving it was not a second copy of the candidate, and **a probe that had to fail**.

| batch | files | result |
|---|---|---|
| 6 | 3 `withSysReadDeadline` seam | 4/4/4 both halves |
| 7 | 6 ordinary | 22/7/4/11/22/4 both halves |
| 8 | `model-file-scan` (1,439 lines) | 66 both halves, matching a static `it()` prior stated first |
| 9 | 6 parameterised-client | 11/17/14/4/27/15 — **went red first, see below** |
| 10 | the 2 the merge base brought in | 8/13 both halves |

No claim here rests on a failure count or a timing delta at any width — those are not instruments under `--no-isolate` right now (96 / 379 / 121 failures on an identical tree, same tenancy).

## The red, because it is the most useful thing in the diff

Batch 9's first candidate half failed 6 tests across 2 files, all `expected null not to be null`. Cause: `resolveBlockInstance` takes its client as a parameter at `block-registry.service:1362` and uses that local for **everything on the path** — `blockUserSubscription.findUnique` (`:1409`), `model.findUnique` (`:1451`), `modelVersion.findFirst` (`:1461`). The inherited analysis said `model.findUnique` was `dbRead`; the `dbRead` spelling it cited is at `:2617`, in a function these tests never call. 15 sites corrected.

**For a parameterised entry point the question is not "where is this table spelled" but "does this call go through the parameterised local".**

## What a reviewer should attack first

1. 🔴 **`storage-resolver.deregisterByFile` is a DELIBERATE partial conversion.** Logging is canonical; the `~/env/server` mock **stays**, because the tests mutate `envValues` per case — the class where lifting it leaves the local alive and disconnected and no run can see the difference. It is a pending specifier so it does not hold the canonical gate. **Do not "finish" it.** This is the judgement in the diff I would least like to be wrong about.
2. The per-case routing in `commentsv2.appListing` (four entry points that disagree) and `block-registry.marketplace-meta` / `spend-cap-config`.
3. The `$transaction` inherit-vs-keep decision, made per file: `coinbase` and `model-version.deregister` inherit (their `tx` and write client were already one object); both `commentsv2` files and `model-early-access-refund` keep a local `tx`, because their assertions mean "written inside the transaction" and inheriting would collapse that.

## Not adversarially reviewed

Everything above is the author's own verification. **That is not a substitute for a hostile read of 134 files**, and this PR should not merge onward to `main` on the strength of it.

https://claude.ai/code/session_01Hpwac9RLBE6uxU11W1BAGA


JustMaier:
## 🔴 DO NOT MERGE THIS PR AS-IS — it is stacked, and the base must be retargeted first

This PR's base is `perf/test-mock-system`, which is **PR #3959's head, not an integration branch**. That is the stacked shape `CLAUDE.md` forbids:

> **NEVER use stacked PRs** — base every PR directly on the integration branch. Stacked PRs silently mis-merge: a squash-merged parent doesn't retarget the child, so the child lands on the orphaned parent branch instead of the real base and its changes go missing.

It cost this repo PR #2520 in June, and the failure is silent — GitHub leaves the child pointing at an orphaned branch and the merge appears to succeed.

**This was a deliberate choice for a good reason** (the alternative was 132 unreviewed files arriving inside #3959, which is one review away from `main`), but it makes the retarget load-bearing.

### Required before this merges

1. Wait for **#3959** to land on `main`.
2. **Retarget this PR's base to `main`** and rebase onto it.
3. Confirm the diff is still ~132 files against the new base before anything else — the count moving is the tell that something was dropped or double-applied.
4. It then needs its **own adversarial review**. Everything in it is the author's own verification of her own work; that is not a substitute for a hostile read of 132 files, and it should not merge onward on the strength of it.

### One check worth repeating after the rebase

Regenerating the allowlist from the merged tree gave `canonical 345 -> 218 (127 migrated)` — exactly this slice's independently-reported count. If a rebase drops or double-applies anything, that number moves. It is the cheapest available proof the merge is intact.

JustMaier:
**Correction to point 1 of the PR body, narrowing my own claim.**

I wrote that `storage-resolver.deregisterByFile`'s `~/env/server` mock must stay and that nobody should "finish" it. That overclaims, and the part I could not personally vouch for has now been checked.

**The canonical `env` proxy DOES honour post-hoc assignment.** `env.mock.ts:79` writes `env.FOO = 'x'` straight into `overrides`; `read()` at `:45` checks `overrides.has(prop)` before falling back to defaults; `defineProperty` and `deleteProperty` land in the same map. So a test *can* mutate env at runtime through the canonical mock.

**That does not rescue a lift.** The proxy honours assignment **to `env`** — it cannot honour assignment to a different object that used to be the mock, and a lift leaves `envValues` as exactly that, alive and disconnected. Every later assignment would write somewhere nothing reads, and the "not configured" case would silently exercise the **configured** branch. None of our gates can see that: collected count unchanged, `residual-mocks.mjs` does not scan `~/env/server` at all, and pass/fail is unchanged wherever nothing downstream of the mutated key is asserted.

**So the accurate claim is narrower and stronger:**

> This file must not be **lifted**. It *is* convertible by **rewrite** — move the mutations into `setEnv({…})` or a direct `env.X = …` inside the case that needs them — and that rewrite needs its own probe: one that makes the not-configured case **fail** after conversion. That was out of scope for a two-file unblocking task.

A refusal to do it the cheap way is not a claim that it cannot be done. Please don't record this as permanently refused.

⚠️ `~/env/server` is a **pending** specifier, so none of this holds the canonical gate either way.

PR #3975, opened 2026-08-16
perf(tests): the unisolated unit-fast project, partitioned from unit by one list

JustMaier:
# ✅ Green. Two root causes found and fixed; ready for a merge decision.

The second vitest project: `unit-fast`, `isolate: false`, membership limited to files whose entire
mock set is canonical. This is the change the shared-mock migration exists to enable — nothing in
the suite runs unisolated until it lands.

Rebased onto `main` with `--onto` (#3959 and #3973 squash-merged, so their commits are content-in-main
but not ancestors) and the manifest regenerated against `main`'s allowlist.

**Adversarially reviewed at `daea3691fd`.** Findings and fixes in the last section; `f1e3ac6ba1`
carries them.

## The three numbers

| | | |
|---|---:|---|
| **eligible** | **470** | of 1,087 unit test files (43.5%) |
| **of which earned** | **46** | carried a guarded `vi.mock` before the migration and are safe now |
| never mocked anything | 424 | eligible on day one; not what the migration was for |
| **excluded, with reason** | **16** | every one records why, and the guard fails if one doesn't |
| pending | 601 | one specifier away: 123 |

Exclusion classes: module-scope flag conflict (`IS_BUILD` / `IS_DATAPACKET`) **5**, module-scope env
stubbing plus a deferred subject import **7**, owns a worker thread or listening socket **4**.

⚠️ **Members are 43.2% of FILES but 13.2% of MODULE LOADS.** The files that qualify first are the ones
that mock nothing, which are the cheap ones. The file count overstates the delivered win by ~3x, and
the expensive files — where the 16x lives — are all still pending. **The win is back-loaded.**

## `membersEarned` was measuring the wrong population

The manifest's own field counted members that still mock something **today**. But converting a file
*removes* its mocks, so a converted member is indistinguishable from one that never mocked anything.
It read **4 → 5** across a day that converted 127 files.

Measured against the pre-migration base instead, the same two member sets read **23 → 46**. Base
frozen as committed data (CI clones shallow, and a baseline that can drift is not a baseline).

Mirror of the trap already recorded for the burn-down: measuring by absence *inflates* a burn-down
and *deflates* an earned count. The flattering direction is not the same one twice.

The review re-derived 46 with a wider instrument (the repo's own `mockPattern`, which also matches
relative spellings and backticks) and got **identical sets — 0 extra, 0 missing** — with a control
proving the scan could see (455 `vi.mock` sites across the `.test.tsx` arm, none of them guarded).

## The blocker — CLOSED. Two root causes, both real

**1. The resolver never ran for the spelling that mattered.** Vite's built-in alias plugin runs ahead
of every user `enforce: 'pre'` plugin and re-enters resolution with the rewritten id, so the hook
never saw `~/env/server` — it never saw *any* `~` specifier. Measured in situ: **2,987 resolutions
in one member's run, zero starting with `~`.** `source === '~/env/server'` was dead code.

That is why every earlier attempt half-worked. The alias-only version covered the `~` spelling and
left the single relative import site failing **deterministically** (12 files). The plugin-only version
covered the relative site and left ~100 tilde sites on the real module, where only `setup.ts`'s
`vi.mock` stood between them and the throw — failing **intermittently**. **Neither ever covered both
spellings.** The id arrives absolute and with *mixed* separators (`C:\...\wt-fast-project` +
`/env/server`) — the fifth distinct spelling found in this class — so the comparison goes through
`canonicalId` rather than string equality. The list of forms is not enumerable by inspection.

**2. The membership rule skipped the closure test for every npm package.** `specifierToPath` returns
null for an external and the sharing loop `continue`d on it, so mocking an external counted as
private whenever no other member *mocked* it — ignoring siblings that import the real one through
production code. `settings-pattern-guard.failclosed` mocks `recheck`;
`settings-pattern-guard.test.ts` reaches the real `recheck` through `settings-pattern-guard.ts`
and mocks nothing. Both were members; load order decided the winner and the fail-closed assertion
inverted in roughly one run in five. An external now maps to the first-party modules that import it
and reuses the same closure test. The mocker is evicted, the importer stays.

```
unisolated  x6   7178 collected   green, every run
isolated    x1   7178 collected   green   <- identical
```

⚠️ **My earlier published mechanism is retracted.** I claimed setup files run once per worker so the
registration races the module cache. Vitest 4.0.18 disproves it in its own source: `runSetupFiles()`
is per-spec and `importFile()` invalidates the setup module first. The real cause was plugin
ordering, and it had nothing to do with setup.

## What is verified

- **Partition, against the runner rather than the manifest**: `vitest list` per project gives 1,087
  distinct files = `manifest.testFiles`, **0 files in more than one project, 0 in none**. The one
  member `unit-fast` does not claim is a sharp file correctly held by `unit-native` — a double-run
  the config exclusion prevents. Both checks carry a control proving they can report non-empty.
- **The partition guard mutation-tested** in both selector files, after the fix below: narrowing
  either `package.json` or `scripts/test-unit-run.mjs` to `'unit'` names the two uncovered projects.
- **The resolver probed over 12 importer spellings**, including the three that used to no-op, with
  `~/env/server-schema` confirmed untouched.
- **Exclusion classes measured over all 1,087 files** by the reviewer: 11 flagged, all 11 already in
  the manifest, **zero false positives repo-wide**; and no live gap from the under-broad direction.
- `pnpm run typecheck`: 0 errors. Prettier + eslint over the files that count as *added*.

## Review findings, and what changed

| # | severity | finding | status |
|---|---|---|---|
| 1 | HIGH | the guard's `--project` regex required a space or `=`, so it read `scripts/test-unit-run.mjs` and evaluated **none** of its lines — the one script every `test:unit:run` goes through | **fixed**, mutation-tested both files |
| 2 | MED | the published mechanism was disproven from vitest's shipped code | **retracted**, replacement labelled a hypothesis |
| 3 | MED-LOW | resolver compared importer ids by exact string; 3 Windows spellings silently no-oped | **fixed**, probed |
| 4 | LOW | `vi.doUnmock('~/env/server')` is a no-op here (2 files, both pending) | **recorded** in the config |
| 5 | LOW | PR said 8/3 exclusions and a comment said 432; artifact says 7/4 and 427 | **corrected** |
| 6 | LOW | `covers()` is stricter than vitest's matcher — false-failure direction only | **noted** in the code |

Checked and found clean: the earned count and its base, the alias's export surface vs the real
module, `enforce: 'pre'` vs `setup.ts:150`, `server-schema` not swallowed, plugin blast radius
outside `src`, and negation specs (vitest ORs them, so none can hide a gap).

## Two findings carried out, neither caused by this change

1. **`no-direct-shared-module-mock` cannot tell a comment from code.** It scans raw text, so quoting
   a `vi.mock('<canonical>', …)` call *in prose* puts the file on the allowlist. Two files repo-wide;
   `app-access.service.test.ts` is on **`main`'s** allowlist for a comment and has no mock to migrate,
   so main's canonical count is 217, not 218. Not fixed here — it means changing a shared guard and
   regenerating the allowlist every branch depends on.
2. **`captureScreenshot.test.ts` collects 0** in a worktree missing `html2canvas` — a stale install,
   not a defect. Recorded because it presented as "my config broke a file", and the tell was the
   documented one: a failure in a file I had not touched.


JustMaier:
Closing this deliberately, not abandoning it. The lane it belongs to is being stopped, and the reason is a measurement made on this branch.

## Why

`unit-fast` works. Six unisolated runs and a matched isolated control agree exactly — 7178 collected, zero failures, zero files lost — and against a control it is worth a measured **9%** on the unit suite (154.7s -> 141.0s, consistent across 5 of 5 interleaved rounds).

What killed it is what the same instrumentation showed about the *rest* of the migration:

```
                        collect worker-time     wall
0 members (control)           3298s            154.7s
235 members                      -               0%
470 members (this PR)         3137s  (-3%)      -9%
1087 members (ceiling)        1118s (-66%)      not measurable (723-870 failures)
```

**The benefit is convex.** 43% of the files buy 3% of the collect saving; 100% buys 66%. Measured at the midpoint, 235 members bought **nothing**. So a half-finished migration is worth approximately zero, and the remaining work is 601 files across **354 distinct blocking specifiers** — the best single specifier clears 17 files, the best 25 clear 89. There is no small set of moves that finishes it.

The ceiling if it *were* finished is roughly 1.5x wall, and that number is arithmetic from the collect delta rather than a measurement — the ceiling run's wall is uninterpretable with that many failures. Unmeasured reward against a multi-week cost, all-or-nothing, is not a good trade.

Worth recording that the lane's founding claim, "16x on the import phase", is not what pays out. Measured against a matched control for the first time here, it is **2.95x on collect worker-time at full membership**.

## Why not merge it anyway for the 9%

As written it taxes everyone: **every new test file added by anyone fails CI until `node scripts/test-perf/gen-fast-project.mjs` is re-run** (~90s). Verified by adding a scratch test file — `--check` exits 1, removing it returns 0. Making membership freeze-tolerant would have removed that, but carrying a fourth vitest project and a generated manifest for a non-transformative gain was judged not worth it.

## What stays on `main`, and it is most of the value

Nothing here needs unwinding. Already merged and justified independently: the canonical mock system, the guard, the allowlist ratchet and codemod, 264 migrated files, the RSA keypair cache (setup 354s -> ~176s), the slow-test tail (#3962), the Windows fixes (#3961), the sharp/`forks` pool split (#3960) and the perf tooling (#3957).

Two benefits that were never claimed at the time: failures now name the mock that failed (`expected "dbWrite.appBlockPublishRequest.findFirst"` rather than `expected "spy"`), and the migration *repaired* an assertion in `model-file.service.test.ts` that had been silently testing nothing.

## Two defects found here that outlive it

1. **`no-direct-shared-module-mock` cannot tell a comment from code.** It scans raw text, so quoting a `vi.mock('<canonical>')` call *in prose* allowlists the file. `app-access.service.test.ts` is on `main`'s allowlist for a comment and has no mock to migrate — **the canonical count is 217, not 218.**
2. **`test:lint-rules` runs 6 of the 10 `no-*` guards.** The four it skips include `no-direct-shared-module-mock`, so the 3-second convention command misses that class entirely and it only surfaces in the 3-minute suite. Being fixed separately.

**The branch is left in place** rather than deleted, so the membership predicate, the manifest generator, the pre-migration baseline and the backlog generator remain recoverable if anyone revisits this.

Issue #3893, opened 2026-08-13
App Listings: off-site listings have no standalone media editor — getMyListingForApp is block-keyed


ZacxDev:
> 🔴 **THE BLOCKER THIS ISSUE NAMES NO LONGER EXISTS. The body below is preserved
> as filed (2026-08-13) and its central premise is now FALSE — read this first.**
>
> **What changed:** `civitai/civitai#3989` (`a15b4fb7b1`, merged and deployed
> 2026-08-17) widened `getMyListingForApp`'s slug arm from
> `{ slug, kind:'onsite', appBlockId:null, status:'draft' }` to
> `{ slug, revisionOfId: null }`. Its input schema takes `appBlockId` **or**
> `slug`, and the service resolves **any top-level listing of either kind**.
>
> So *"its input is an `appBlockId`; an off-site listing has no `AppBlock` at all,
> so there is no id with which to call it"* is no longer true, and **the fix this
> issue proposes — "re-key `getMyListingForApp`, or add a listing-keyed sibling" —
> has effectively shipped**, keyed on `slug` rather than `appListingId`. The wire-
> contract question it raises (sibling vs re-key in place, given
> `app-collaborator.cli-contract.test.ts`) is moot: the slug arm is additive and the
> `appBlockId`-only payload still parses.
>
> **The gap is still real, and it is now much smaller.** What withholds the tab
> today is not the resolver but `editorTabsFor`'s media arm:
> `capabilities.listingMedia === true && ctx.appBlockId != null`. Three comments in
> the tree already say so and point here —
> `src/components/Apps/appListingEditorTabs.ts` (media arm doc),
> `src/shared/constants/app-capabilities.constants.ts:98-102` (*"Held false by the
> WEB tab gate … NOT by the host resolver … Plumbing, not schema"*), and
> `src/pages/apps/listing/[appListingId]/edit.tsx:~188`.
>
> **Revised scope — four touch points, no proc re-key:**
> 1. `ListingMediaEditor` accepts a `slug` alongside `appBlockId`. Measured: it uses
>    `appBlockId` in exactly **4 places** (`ListingMediaEditor.tsx:56, 73-74, 121,
>    141`), all of them the `getMyListingForApp` query key and its invalidations.
>    Everything below it — `ListingAssetStep`, the asset procs — is already
>    listing-keyed and seat-aware.
> 2. The slug is already in hand at the page: `AppListingAuthoringContext` extends
[...]

PR #3989, opened 2026-08-16
fix(app-listings): resolve any top-level listing by slug in getMyListingForApp (#3984)

### LINKED ISSUE #3984: App Listings: getMyListingForApp needs a slug selector that resolves off-site listings — the declared one is scoped to on-site pre-approval drafts
## Summary

`appListings.getMyListingForApp` is the one hop that turns "the app I am authoring" into an
`AppListing.id`, and every listing-media surface is hosted on it. It is reachable **only by
`appBlockId`** in practice, so a client whose only handle is a **slug** — which is every
non-web client, the `civitai` CLI included — cannot resolve an off-site listing at all.

The input schema already declares a `slug` selector. The service arm behind it is scoped so
narrowly that it resolves **nothing an off-site app can ever be**, which is why the selector
reads as supported and behaves as absent.

Related: #3893 (same resolver, the web Media-tab half). **This is not that issue** — see
*Why #3893 does not close this* below.

## Mechanism (read from source at `07a1c537dc`, not inferred)

`src/server/schema/blocks/offsite-listing.schema.ts:222-231` — both selectors are optional,
`.refine`d to require one:

```ts
appBlockId: z.string().min(1).max(64).optional(),
slug:       z.string().min(1).max(64).optional(),
```

`src/server/services/blocks/offsite-listing.service.ts:1692-1710` — what each one actually
resolves:

```ts
let listing = appBlockId
  ? await dbRead.appListing.findUnique({ where: { appBlockId }, select: entrySelect })
  : null;
if (!listing && slug) {
  listing = await dbRead.appListing.findFirst({
    where: { slug, kind: 'onsite', appBlockId: null, status: 'draft' },
    select: entrySelect,
  });
}
```

The slug arm is bounded to the W13 pre-approval draft shape: **on-site, no AppBlock yet,
status `draft`**. An off-site listing fails the first clause by definition — `kind: 'offsite'`
is what it *is* — so no off-site app is addressable by either selector, ever. An approved
on-site app fails the second and third clauses, so it too resolves only by `appBlockId`.

The owner gate is downstream and selector-independent: `resolveListingRole(listing.id, userId)`
at :1733, after the row is in hand.

## This explains the CLI's live measurements exactly

Measured by @ZacxDev against `https://civitai.com` at `civitai/cli@1b20b99`, reported in
civitai/cli#424 — **quoted here, not re-measured in this issue**. Only the selector varied:

| app | `appBlockId` only | `slug` only |
|---|---|---|
| `gen-matrix` (on-site, approved+live) | ✅ `apl_01KWFP4FEEJRWN27CJA49CDY4Q` | ❌ 404 |
| `custom-generators` (on-site, approved+live) | ✅ `apl_01KXKAJF85NQ8X702AHKX2MMAW` | ❌ 404 |
| `radio`, `comfy` (off-site) | n/a — no appBlockId exists | ❌ 404 |
| `definitely-not-an-app-zzz` | n/a | ❌ 404 |

Positive control satisfied (the `appBlockId` calls return real ids in the same run, so the
route, auth and tRPC encoding are all right); negative control satisfied (a nonexistent slug
404s). The `where` clause above accounts for every row: the two approved on-site apps carry an
`appBlockId` and status `approved`, so both fail the fallback; the off-site apps fail on `kind`.

The observable in the CLI, from civitai/cli#422 (4/4 off-site apps fail, 7/7 on-site pass —
`kind` predicts it exactly):

```
$ civitai app listing set-icon ./icon.png --slug radio --changelog "..." -y
Error: no such submission for app "radio" — run `civitai app submit` first; …
```

`app submit` packages a block bundle; an off-site app is a registered URL and has none, so that
advice names a step that cannot exist. All four off-site apps meanwhile return HTTP 200 with
non-null `iconUrl`/`coverUrl` on `GET /api/v1/apps/<slug>` — the media exists and is serving, it
was set through the wizard, and no client-side route can address it.

## The ask

**Widen the existing slug arm** so a listing is resolvable by the handle that identifies it
publicly:

```ts
if (!listing && slug) {
  listing = await dbRead.appListing.findFirst({
    where: { slug, revisionOfId: null },
    select: entrySelect,
  });
}
```

Why this is cheap and safe rather than a new capability:

- **`AppListing.slug` is `@unique` globally, across both kinds** —
  `packages/civitai-db-schema/prisma/schema.prisma:2818` ("Globally-unique store slug across
  BOTH kinds"). No ambiguity, no new index, no migration.
- **A public slug always names a top-level parent.** `beginListingRevision` gives a shadow a
  synthetic `rev-<ulid>` slug (`offsite-listing.service.ts:1063`), which is why the service's own
  comment at :876 already says this resolver "only ever sees a parent". The
  `revisionOfId: null` clause makes that invariant explicit instead of load-bearing-by-accident.
- **No access change.** `resolveListingRole` is unchanged and still runs after resolution. A
  stranger's slug is refused exactly as a stranger's `appBlockId` is today.
- **No wire-contract change.** The CLI already sends `slug` on both paths
  (`internal/appapi/listing.go:229`); `app-collaborator.cli-contract.test.ts`'s
  `appBlockId`-only case keeps passing untouched.

One honest behaviour delta: a slug belonging to someone else currently returns `NOT_FOUND`
(filtered out before the gate) and would afterwards return `NOT_OWNED` → `FORBIDDEN`. That
distinguishes "exists" from "yours" for a slug — but store slugs are already public in the
catalog and its listing pages, so it discloses nothing new. Flag it if you disagree; narrowing
the widened clause to `status: { not: 'removed' }` is available if a tighter shape is wanted.

Then `CAPABILITIES_BY_KIND.offsite.listingMedia`
(`src/shared/constants/app-capabilities.constants.ts:93`) flips to `true` — the same cell #3893
names, whose comment already calls this "plumbing, not schema".

## Why #3893 does not close this

#3893 proposes re-keying to, or adding a sibling on, **`appListingId`**. For the web that is
sufficient: the authoring page is `/apps/listing/<appListingId>/edit`, so the id is already in
the URL. A CLI has no such id.

It is *nearly* reachable — `GET /api/v1/apps/<slug>` does return `id: row.id`, the `apl_` listing
id (`app-listing.service.ts:402`), and the CLI already decodes it into `civitai.AppDetail.ID`
(`pkg/civitai/apps.go:136`) on a call it already makes. But that route is
**approved-only** (`app-listing.service.ts:672`) and additionally gated by
`resolveStoreVisibilityScope`, the `public-external` kind filter and the mature-rating check
(:678, :688). So resolving your own listing id through the public catalog:

1. cannot see a **draft or pending** off-site listing at all — precisely the window in which an
   author most wants to attach an icon and cover; and
2. can fail for an owner on their own **approved** app when a visibility or maturity gate
   excludes it, producing a "no listing" that is really "not publicly visible to you here".

So a listing-keyed sibling alone leaves every non-web client on a two-hop through a public,
approved-only, scope-gated route. **A slug selector is one hop, works pre-approval, and needs no
public read.** Both can land — they are not exclusive — but the slug form is what closes this.

## What it unblocks

- civitai/cli#422 outcome 1 — `app listing status` / `set-icon` / `set-cover` /
  `add-screenshot` / `rm-screenshot` for off-site apps. The asset procs are already
  listing-keyed and seat-aware (#3893 says so), so the resolver is the *only* missing hop:
  one selector makes the whole subcommand family work.
- civitai/cli#424 — filed as "the slug selector is documented as load-bearing, measured inert,
  and I cannot tell whether the pending path is broken". The `where` clause above answers it:
  the slug arm is real but scoped to on-site pre-approval drafts, so it is inert for everything
  else by construction, not by accident.

Stopgap already shipped client-side: civitai/cli#426 detects `kind: offsite` **on the error path
only** and replaces the dead-end advice with a precise refusal naming the App-store listing UI as
the authoritative surface. That is a refusal, not a repair — it is deliberately not attempting a
client-side fix, because there isn't one.

## Not verified here

Nothing in this issue was run against a live server. The mechanism is read from `main` at
`07a1c537dc`; the behavioural table is quoted from civitai/cli#424's live run at
`civitai/cli@1b20b99`. I have not measured a genuinely **pending** first-version on-site app —
the one case the current slug arm is designed to serve — so whether that path works today is
still open, and it is untouched either way by the widening proposed above.

ZacxDev:
> **Corrections applied 2026-08-16 after an adversarial audit** — four assertions in an
> earlier revision of this body were wrong or overstated and have been rewritten in place.
> See the [correction comment](https://github.com/civitai/civitai/pull/3989#issuecomment-5311163675) for the diff of what changed and
> why. No behaviour change: the `where` clause, the capability cell and `editorTabsFor`
> are untouched by the correction commit.

## What

Widen the slug arm of `appListings.getMyListingForApp` so a listing is resolvable by the
handle that identifies it publicly.

```diff
-      where: { slug, kind: 'onsite', appBlockId: null, status: 'draft' },
+      where: { slug, revisionOfId: null },
```

`getMyListingForApp` is the one hop that turns "the app I am authoring" into an
`AppListing.id`, and every listing-media surface is hosted on it. Its input schema has
always declared a `slug` selector, but the arm behind it was scoped to the W13
pre-approval draft shape — **on-site, no AppBlock yet, status `draft`** — whose very first
clause excludes every off-site listing (`kind: 'offsite'` is what it *is*), and which in
practice has no `AppBlock` for the other selector either. So the selector read as
supported and behaved as absent, and an approved on-site app was likewise
`appBlockId`-only.

**Selector precedence:** `appBlockId` is tried first and **wins**; the slug arm runs only
on a miss. That was already the behaviour and is now pinned by a test (it was not).

No schema change and no migration.

Resolves civitai/civitai#3984. Related: #3893 (the web Media-tab half — see below), and
civitai/civitai#4003 (rate-limiting follow-up — see the disclosure section).
Found client-side in civitai/cli#422 / civitai/cli#424, where the selector table was
measured live: 4/4 off-site apps 404 by slug, 7/7 on-site apps resolve by `appBlockId`.

## Why it is safe — the three premises, verified in this branch, not quoted

1. **`AppListing.slug` is `@unique` globally, across both kinds.**
   `packages/civitai-db-schema/prisma/schema.full.prisma:2812` —
   `slug String @unique`, with the comment *"Globally-unique store slug across BOTH
   kinds"*. No ambiguity, no new index.
   (The issue cites `schema.prisma:2818`; the file is `schema.full.prisma` and the line
   is 2812. Same declaration.)
2. **A public slug always names a top-level parent.** `beginListingRevision`
   (`offsite-listing.service.ts:1065`) gives a shadow a synthetic `rev-<ulid>` slug.
3. **The owner gate is downstream and selector-independent.**
   `resolveListingRole(listing.id, userId)` takes the resolved id and nothing else; it
   delegates to `resolveListingAccess`, which re-reads the row by id. Unchanged.

**Worth flagging on premise 2 — `revisionOfId: null` is load-bearing.** A shadow revision
is `kind: parent.kind`, `appBlockId` NULL, status `draft` (`beginListingRevision`
clones the parent's kind). For an **on-site** parent that is exactly the old clause, so
those shadows **matched** it; a shadow of an **off-site** parent is `kind: 'offsite'` and
did not. Either way the resolver's existing comment ("only ever sees a top-level parent")
held only because nobody knows a shadow's `rev-<ulid>` slug. On this branch it holds by
construction, and `revisionOfId: null` is now the **only** clause holding it — there is a
test that fails if it is dropped.

## On "an off-site listing has no AppBlock"

Stated carefully, because an earlier revision of this body said *"no AppBlock at all,
ever"* and that is **not** true:

- `schema.full.prisma:2849-2853` says `appBlockId` is *"Set for EVERY backfilled row — on-site
  **AND** the #2821 off-site rows (both come from an AppBlock). It is **NOT** a kind
  discriminator: discriminate on `kind`, never on appBlockId nullness."* Only a
  **natively-created** off-site listing leaves it NULL.
- `mapAppBlockToListing` can still mint `kind:'offsite'` with a non-null `appBlockId`.
  That shape measured **0 rows in production, 2026-08-11**
  (`src/components/Apps/appListingEditorTabs.ts:109-113`, `resolveAccessibleAppBlockIds`).

So: **empirically zero today, structurally possible, and a backfilled class exists.** The
widening is correct either way — the clause that had to go was `kind`, and dropping
`appBlockId: null` alongside it is what covers the backfilled class too.

## 🔴 Disclosure delta — please read and disagree if you want to

**A slug belonging to another user returns `NOT_OWNED` (→ tRPC `FORBIDDEN`) rather than
`NOT_FOUND` — but this is a WIDENING of an existing delta, not a new one.**

An earlier revision of this body said *"Before, the `where` filtered a stranger's row out
before the gate ran."* That is **wrong for the shape the old clause admitted.** The old
`where` had no owner predicate either, and `main` already carries a test asserting the
consequence:
`src/server/services/blocks/__tests__/offsite-listing.get-my-listing-for-app.service.test.ts:596`
— *"a draft owned by ANOTHER user → NOT_OWNED (ownership still enforced on the slug
path)"*. For a pre-approval on-site draft, `NOT_OWNED` was **already** the answer on
`main`.

What actually changes is the **size of the addressable set**: rows the old clause
excluded (off-site, approved, pending, rejected, removed, block-backed) now also reach the
owner gate and get `NOT_OWNED` instead of `NOT_FOUND`. The delta is real; it is the same
delta, over more rows.

**The justification is NOT "store slugs are already public in the catalog."** That was the
other wrong assertion here: `getAppDetail` is **approved-only**
(`app-listings.router.ts:1337`, `app-listing.service.ts` — *"a missing OR non-approved
(draft/pending/rejected) listing returns null"*), and the newly-addressable set is
precisely the **non-approved** rows a logged-out browser cannot see.

**The true justification is stronger:** the same cohort already has a *complete* existence
oracle over the *entire* slug namespace, via `submitExternalListing`
(`src/server/services/blocks/offsite-listing.service.ts:238-241`):

```ts
const existingListing = await dbRead.appListing.findUnique({
  where: { slug },            // no kind / status / revisionOfId filter
  select: { id: true },
});
if (existingListing) throw slugTakenError(slug);   // `slug "X" already taken`
```

**The residual, stated honestly:** that oracle is a **mutation** rate-limited to
**10/hour** (`app-listings.router.ts:477-483`); `getMyListingForApp` is an
**unrate-limited `.query`**. So this PR does not create the disclosure — it removes its
budget. Cohort is `appDeveloperProcedure` + the store Flipt gate (mods and gated app
devs), which is why it is tracked rather than fixed here:
**civitai/civitai#4003**. If you want it tightened in this PR instead, the cheap fix is
`rateLimit({ limit: 60, period: 60 })`, matching `getAppDetail` — say the word.

There is a test pinning the new verdict explicitly, so the delta is visible in the suite
rather than implied.

## Why the capability cell is NOT flipped, and what #3893 still owns

`CAPABILITIES_BY_KIND.offsite.listingMedia` stays `false` in this PR.

🔴 **But the reason changed, and every doc comment asserting the old reason has been
updated.** Five sites said the cell is held false because `getMyListingForApp` is
block-keyed. It is not block-keyed any more — *this PR* re-keyed it. What holds the cell
false is `editorTabsFor`'s `media` arm requiring `ctx.appBlockId != null`
(`src/components/Apps/appListingEditorTabs.ts:128`) **in addition to** the cell, plus the
panel in `edit.tsx` passing an `appBlockId` down. Flipping the cell without re-hosting
that surface either changes nothing or renders a tab that cannot load. That re-hosting is
**#3893**'s scope.

Corrected: `app-capabilities.constants.ts`, `appListingEditorTabs.ts`, `edit.tsx`, the
tRPC JSDoc in `app-listings.router.ts`, and the `listingMedia: false` rationale in
`app-access.service.test.ts`.

🔴 **One of them was an access-control classification, not just prose.**
`app-collaborator.permission-matrix.test.ts` filed the proc under `ONSITE_ONLY` with the
rationale *"it is UNCALLABLE [for offsite], which is the same structural mechanism that
makes `submitVersion: false` real."* That mechanism no longer exists, so the
**classification** was wrong, not merely the wording. The entry now lives in the
kind-crossed `MATRIX`, and the off-site half calls it the way an off-site client must —
by slug — asserting the identical allow/deny column. Its `findFirst` mock was made
**predicate-honouring** at the same time: restoring `kind:'onsite'` to the service now
reddens the off-site ALLOW cells (measured); a permissive mock would have passed under the
old clause too and asserted nothing.

## Not touched, on purpose: `deleteOnsiteDraftListingForSlug`

`publish-request.service.ts:1476` carries a clause that looks identical to the one this PR
deleted — `{ slug, kind: 'onsite', appBlockId: null, status: 'draft' }` — and it **must
stay narrow**. It is a `deleteMany` with no owner predicate and nothing downstream to
refuse it; the same widening there would hard-delete whatever top-level row holds the
slug. A cross-reference to #3984 is now in the code so a future "consolidation" cannot
quietly widen a destructive query.

## 🟢 Open question raised by the characterization test — not answered here

**The characterization test in `offsite-listing.edit-consolidate.service.test.ts` adds no
coverage**, and is now labelled as such. That file's `ownedRow` defaults to
`kind: 'offsite'`, so the pre-existing "APPROVED with NO prior shadow → CREATES one" case
above it **already** drives an approved off-site listing through the whole shadow media
path, and a pre-existing case already asserts `kind === 'offsite'` survives. Deleting the
test would not move a line of coverage. It is kept only because it names the question in
one place.

The question itself: `getMyListingForEdit` gates on ownership + status and has **no `kind`
check**; neither do the listing-keyed asset procs (`loadOwnedListing`). So an approved
off-site listing gets a shadow minted and the shadow's asset row ids returned — the full
media flow. Meanwhile `listingMedia: false` says off-site has no media surface, and that
cell is read **only** by the web tab gate; no service consults it (unlike `earnings` and
`submitVersion`, the two other `false` cells, which *are* enforced service-side).

🔴 **And it leans further than "an editor entry read slipped through":**
`REVIEWABLE_LISTING_KINDS = ['onsite', 'offsite']` (`offsite-listing.service.ts:1920`)
means the revision **submit → approve → apply** chain accepts an off-site listing **end to
end**. A media revision an off-site owner stages can be submitted, reviewed and applied.

**So: is off-site media editing through the listing-keyed procs INTENDED — in which case
the cell is a statement about one web tab and nothing more — or a gap the
`listingMedia: false` cell believes is closed, in which case the missing check is in the
services rather than in the tab gate?** That call belongs to whoever owns the capability
table, not to this PR. Re-label or invert the test by that decision; do not silently
delete it.

## Tests

`src/server/services/blocks/__tests__/offsite-listing.get-my-listing-for-app.service.test.ts`
— the `describe` was renamed (it is no longer draft-only). The four existing pre-approval
draft cases stay; the one that asserted the exact old `where` now asserts the new one.

New: an **off-site** listing resolves by slug; an **approved on-site** listing resolves by
slug; a slug owned by another user → `NOT_OWNED`; a **shadow's** synthetic `rev-` slug does
**not** resolve; **`appBlockId` wins over a slug naming a different listing**; a
**`removed`** and a **`rejected`** listing resolve by slug and return their
`editBlockedReason`.

**The mock honours the `where` clause.** The pre-existing slug tests stubbed
`findFirst` to return a fixed row whatever the clause said — which passes under *any*
clause, including the one being replaced. The new cases evaluate the clause the service
actually sent against a row set, so they are claims about the predicate rather than about
the stub. The same treatment was applied to the permission matrix's mock (see above).

### Mutation matrix

Every mutant below was applied to the narrowest expression that can be wrong, run, and
reverted. Counts are over the four touched test files (189 tests at HEAD).

| mutant | before the audit fixes | after |
| --- | --- | --- |
| `if (!listing && slug)` → `if (slug)` (precedence) | **SURVIVED 44/44** | **1 killed** — the precedence test, on `expected 'apl_other' to be 'apl_block'` |
| `+ status: { not: 'removed' }` | 1 killed — *only* the structural `toHaveBeenCalledWith` pin; no behavioural case noticed | **2 killed** — the pin **and** the `removed` behavioural row |
| `+ status: { not: 'rejected' }` | n/a | **2 killed** — the pin and the `rejected` row |
| restore `kind: 'onsite'` (the old clause's first term) | 3 killed | **7 killed** — incl. the two off-site `ALLOW` cells in the permission matrix, which is what makes that mock's predicate-honouring meaningful |
| restore `status: 'draft'` | 4 killed | **6 killed** |
| restore `appBlockId: null` | n/a | **2 killed** |
| drop **only** `revisionOfId: null` | 2 killed | **2 killed** — the shadow case + the `where` pin. Unchanged, as expected |
| none (HEAD) | 44/44 green | **189/189 green** |

Harness validated before any verdict was read from it: the first mutant run was a
**negative control** (restore `kind:'onsite'`) and it went red, so the runner can fail.

**One finding from doing this rather than assuming it:** the shadow case *passed* under
the old-clause mutant on its first draft — for the wrong reason. Without `viewByListingId`
wired, a resolver that *did* admit the shadow still threw `NOT_FOUND`, several steps
later, out of `loadListingEditView`. The test now wires the edit view, so an admitted
shadow *resolves* and the assertions are the only thing between the two behaviours. It is
written down in the test's own comment.

## Gate

- `pnpm run typecheck` — **OK, 0 type errors** (68s)
- `tsc -p tsconfig.tests.json --noEmit` (test files, excluded from the script above) —
  **738 pre-existing errors repo-wide, 0 in the 11 files this branch touches.** Positive
  control: the same grep finds 4 errors in a known-bad file, so the zero is a measurement
  and not a broken pattern.
- `pnpm run lint` — **4317 problems (368 errors, 3949 warnings), identical to the count on
  `main`**, which was run as a control. ESLint over the 11 changed files alone: **0
  errors**, 31 warnings, all pre-existing `_a is defined but never used` in mock blocks and
  one unused import I did not touch.
- `pnpm run prettier:write` — clean (see the note below)
- `pnpm run test:unit:run` — **Test Files 1092 passed | 1 skipped (1093)**,
  **Tests 17094 passed | 21 skipped (17115)**, 106s. Both projects
  (`unit` *and* `unit-native`) — `--project unit` alone would have run 1059 of 1065 and
  exited 0. Counts read from the summary, not from the exit code.

**The full suite caught something the per-file runs did not**, which is worth recording:
after the first mutation rounds the two touched files were green, but one new assertion
pinned the wrong literal (`'no listing found for …'` vs the real
`'no listing found for app …'`) and only the full run surfaced it.

`prettier-changed.mjs` formats whole **dirty files**, and this repo is not Prettier-clean,
so some hunks are Prettier reformatting pre-existing lines (most visibly in
`publish-request.service.ts`, where my actual change is one comment block). They are not
my edits. Flagging so a reviewer does not read them as intentional.

No `schema.full.prisma` change, so `db:check-generated` is not implicated.

https://claude.ai/code/session_01GNdYjxagYYedKVxuoaQT4T

### DIFF
--- a/src/components/Apps/appListingEditorTabs.ts
+++ b/src/components/Apps/appListingEditorTabs.ts
@@ -81,11 +81,16 @@ export type EditorTabContext = {
  *                       gate an incidental PROXY: it happened to correlate with the truth
  *                       (off-site listings usually have no block) without ever expressing
  *                       it. `listingMedia` was split out of `listingContent` so the table
- *                       states the real constraint — the standalone media editor is hosted
- *                       by the BLOCK-keyed `getMyListingForApp` — and this gate now reads
- *                       it. Off-site media is still editable inside the details wizard,
- *                       which is why withholding the TAB loses nothing today.
- *                       Widening it is civitai/civitai#3893.
+ *                       states the real constraint, and this gate now reads it.
+ *                       🟡 That constraint is no longer "the host resolver is block-keyed":
+ *                       civitai/civitai#3984 re-keyed `getMyListingForApp` to take
+ *                       `appBlockId` OR `slug`, so the resolver reaches an off-site listing
+ *                       fine. What withholds the tab TODAY is the `ctx.appBlockId != null`
+ *                       clause below — the panel hands an `appBlockId` to
+ *                       `<ListingMediaEditor>`, and an off-site listing has none. Off-site
+ *                       media is still editable inside the details wizard, which is why
+ *                       withholding the TAB loses nothing today. Widening it (flip the cell
+ *                       AND re-key the panel onto the slug) is civitai/civitai#3893.
  *
  *   - `manifest`      — BOTH `capabilities.submitVersion` AND a backing block.
  *
--- a/src/pages/apps/listing/[appListingId]/edit.tsx
+++ b/src/pages/apps/listing/[appListingId]/edit.tsx
@@ -185,9 +185,12 @@ export default function AppListingEditPage() {
               </Tabs.Panel>
             ) : null}
 
-            {/* 🔴 `context.appBlockId` is non-null whenever this tab is in the set — the
-                media editor is hosted by the BLOCK-keyed `getMyListingForApp`, which is
-                exactly why `editorTabsFor` withholds the tab without one. */}
+            {/* 🔴 `context.appBlockId` is non-null whenever this tab is in the set — THIS
+                PANEL is block-keyed (it passes an `appBlockId` down), which is exactly why
+                `editorTabsFor` withholds the tab without one. Note the host resolver is
+                NOT the constraint any more: `getMyListingForApp` takes `appBlockId` OR
+                `slug` since civitai/civitai#3984. Re-keying this panel onto the slug is
+                civitai/civitai#3893. */}
             {tabs.includes('media') && context.appBlockId ? (
               <Tabs.Panel value="media" pt="md" data-testid="apps-edit-panel-media">
                 <ListingMediaEditor appBlockId={context.appBlockId} />
--- a/src/server/routers/app-listings.router.ts
+++ b/src/server/routers/app-listings.router.ts
@@ -620,8 +620,11 @@ export const appListingsRouter = router({
     }),
 
   /**
-   * OWNER: resolve the caller's OWN listing by its backing `appBlockId` — the entry
-   * read for the owner-facing on-site listing-media page (`/apps/<appBlockId>/listing`).
+   * OWNER: resolve the caller's OWN listing by its backing `appBlockId` OR by its public
+   * `slug` — the entry read for the owner-facing on-site listing-media page
+   * (`/apps/<appBlockId>/listing`) and for non-web clients that only hold a slug.
+   * `appBlockId` takes PRECEDENCE: the slug arm runs only when it missed or was absent,
+   * and it resolves any TOP-LEVEL listing (either kind, any status) — civitai/civitai#3984.
    * Returns the `AppListing.id` the page passes to `beginListingRevision` + the asset
    * procs, plus the listing status / content rating / whether a revision is already
    * under review, AND the EDITABLE target's current assets (`assets` + `shadowId`) so
--- a/src/server/schema/blocks/offsite-listing.schema.ts
+++ b/src/server/schema/blocks/offsite-listing.schema.ts
@@ -226,7 +226,11 @@ export const getMyListingForAppSchema = z
     appBlockId: z.string().min(1).max(64).optional(),
     // The slug arm resolves ANY top-level listing (civitai/civitai#3984): the W13
     // pre-approval draft of a FIRST-version on-site app (no AppBlock yet), and an
-    // OFF-SITE listing, which never has an AppBlock to be keyed by at all.
+    // OFF-SITE listing, which in practice carries no AppBlock to be keyed by (0 rows
+    // of the `kind:'offsite'` + non-null `appBlockId` shape in production, measured
+    // 2026-08-11 — empirical, not structural; `appBlockId` is NOT a kind
+    // discriminator, see schema.full.prisma). `appBlockId` takes precedence when
+    // both are supplied.
     slug: z.string().min(1).max(64).optional(),
   })
   .refine((v) => v.appBlockId != null || v.slug != null, {
--- a/src/server/schema/blocks/offsite-listing.schema.ts
+++ b/src/server/schema/blocks/offsite-listing.schema.ts
@@ -214,16 +214,19 @@ export type BeginListingRevisionInput = z.infer<typeof beginListingRevisionSchem
 
 /**
  * OWNER: resolve the caller's OWN listing by its backing `appBlockId`
- * (`AppListing.appBlockId` is `@unique`) — the entry read for the owner-facing
- * on-site listing-media page. Returns the `AppListing.id` (the target for
- * `beginListingRevision` + the asset procs). Owner-bound in the service
- * (NOT_OWNED→FORBIDDEN, NOT_FOUND when no listing row exists for the app).
+ * (`AppListing.appBlockId` is `@unique`) or by its public `slug`
+ * (`AppListing.slug` is `@unique` across BOTH kinds) — the entry read for the
+ * owner-facing listing-media page and for non-web clients. Returns the
+ * `AppListing.id` (the target for `beginListingRevision` + the asset procs).
+ * Owner-bound in the service (NOT_OWNED→FORBIDDEN, NOT_FOUND when no listing
+ * row exists for the app).
  */
 export const getMyListingForAppSchema = z
   .object({
     appBlockId: z.string().min(1).max(64).optional(),
-    // W13 draft-at-submit: a FIRST-version app has no AppBlock yet, so the
-    // owner-media page resolves its pre-approval draft BY SLUG while pending.
+    // The slug arm resolves ANY top-level listing (civitai/civitai#3984): the W13
+    // pre-approval draft of a FIRST-version on-site app (no AppBlock yet), and an
+    // OFF-SITE listing, which never has an AppBlock to be keyed by at all.
     slug: z.string().min(1).max(64).optional(),
   })
   .refine((v) => v.appBlockId != null || v.slug != null, {
--- a/src/server/services/blocks/__tests__/app-access.service.test.ts
+++ b/src/server/services/blocks/__tests__/app-access.service.test.ts
@@ -161,13 +161,16 @@ describe('capabilitiesForKind — capabilities are DERIVED, never configured', (
     //
     // 🔴 `listingMedia` is a SURFACE fact and is the newest, narrowest cell: an off-site
     // listing CAN hold and edit assets (the submit/edit wizard does exactly that, through
-    // listing-keyed asset procs), but the standalone media EDITOR is hosted by the
-    // BLOCK-keyed `getMyListingForApp`, so there is no id to open it with. It is split
+    // listing-keyed asset procs), and since civitai/civitai#3984 the standalone editor's
+    // HOST RESOLVER reaches it too — `getMyListingForApp` takes `appBlockId` OR `slug`.
+    // What still withholds the surface is the WEB tab gate `editorTabsFor`, whose `media`
+    // arm requires `ctx.appBlockId != null` on top of this cell, and an off-site listing
+    // has no block id to give it. The cell stays `false`, but for a NARROWER reason than
+    // it was written with: it is about the TAB now, not about the resolver. It is split
     // out from `listingContent` — which stays TRUE for both kinds, because listing
-    // scalars really are editable on both — so the table stops asserting a surface that
-    // does not exist. Tracked by
-    // https://github.com/civitai/civitai/issues/3893; flipping this cell is what that
-    // issue's fix does.
+    // scalars really are editable on both — so the table stops asserting a surface the UI
+    // does not offer. Tracked by https://github.com/civitai/civitai/issues/3893; flipping
+    // this cell (and dropping that block-id clause) is what that issue's fix does.
     expect(capabilitiesForKind('offsite')).toEqual({
       listingContent: true,
       listingMedia: false,
--- a/src/server/services/blocks/__tests__/app-collaborator.permission-matrix.test.ts
+++ b/src/server/services/blocks/__tests__/app-collaborator.permission-matrix.test.ts
@@ -150,7 +150,24 @@ beforeEach(() => {
     app: { userId: OWNER },
   });
   mockDb.appListing.findUnique.mockResolvedValue(draftListing());
-  mockDb.appListing.findFirst.mockResolvedValue(null);
+  // 🔴 PREDICATE-AWARE. Two different reads land on `appListing.findFirst`:
+  //   - `getMyListingForApp`'s SLUG arm (`where.slug`) — must HIT, or the off-site half
+  //     of that entry point could not resolve at all and its DENY cells would pass
+  //     vacuously (right verdict, wrong reason: NOT_FOUND, not the access gate);
+  //   - the shadow-existence probe (`where.revisionOfId`) — must MISS, or
+  //     `beginListingRevision` reports a shadow that does not exist.
+  mockDb.appListing.findFirst.mockImplementation(async (args: unknown) => {
+    const w = (args as { where?: Record<string, unknown> }).where ?? {};
+    if (w.slug == null) return null; // the shadow-existence probe
+    // 🔴 …and it HONOURS the rest of the clause. A mock that handed the row back for any
+    // `where` containing a slug would let the off-site ALLOW cells pass under the OLD,
+    // narrow `{ slug, kind:'onsite', appBlockId:null, status:'draft' }` clause too — i.e.
+    // it would assert nothing about the widening it exists to depend on. Measured: with
+    // this evaluation in place, restoring `kind:'onsite'` to the service reddens the
+    // off-site ALLOW cells; without it, they stay green.
+    const row = draftListing() as Record<string, unknown>;
+    return Object.entries(w).every(([f, v]) => f === 'slug' || row[f] === v) ? row : null;
+  });
   mockDb.appCollaborator.findFirst.mockImplementation(async (args: unknown) => {
     const w = (args as { where: { appListingId?: string; userId: number; status?: string } }).where;
     // 🔴 The seat lives on the LIVE parent listing. Honouring `appListingId` is what
@@ -208,6 +225,30 @@ const MATRIX: Record<string, Record<string, boolean>> = {
     stranger: false,
     moderator: false,
   },
+  // offsite-listing: getMyListingForApp (NO mod bypass).
+  //
+  // 🔴 THIS ROW USED TO LIVE IN A SEPARATE `ONSITE_ONLY` BUCKET, on the rationale that the
+  // proc "is keyed on an `appBlockId`; an off-site listing does not have one, so the proc
+  // is not denied for offsite — it is UNCALLABLE, the same structural mechanism that makes
+  // `submitVersion: false` real". 🔴 THAT MECHANISM NO LONGER EXISTS. civitai/civitai#3984
+  // re-keyed the proc to `appBlockId` OR `slug`, and the slug arm resolves any TOP-LEVEL
+  // listing of either kind — so an off-site listing IS reachable, and the classification
+  // was stale, not merely mis-worded. It is therefore kind-CROSSED like every other row:
+  // the off-site run calls it the way an off-site client must (by slug) and asserts the
+  // identical allow/deny column, which is the claim "the widening moved the refusal onto
+  // the OWNER GATE and changed nobody's access".
+  //
+  // (`CAPABILITIES_BY_KIND.offsite.listingMedia` stays `false` regardless: that cell is
+  // held by the WEB tab gate `editorTabsFor`, whose `media` arm still requires
+  // `ctx.appBlockId != null`. Flipping it is civitai/civitai#3893, not this.)
+  'offsite.getMyListingForApp': {
+    owner: true,
+    'accepted editor': true,
+    'PENDING editor': false,
+    'REJECTED editor': false,
+    stranger: false,
+    moderator: false,
+  },
   // app-collaborators: listCollaborators — the ROSTER read. A moderator DOES pass here
   // (the service takes an explicit isModerator), unlike the two author-edit gates above.
   'collaborators.list': {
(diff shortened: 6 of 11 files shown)


ZacxDev:
## 🔴 Corrections to this PR's description — four assertions were wrong or overstated

An adversarial audit of this PR found four claims in the body that do not hold. **I have
rewritten the body**, but a reviewer who read the earlier version may be carrying the
wrong model, so here is exactly what changed and why. All findings were 🟡/🟢 — **the code
change itself is correct and the `where` clause is untouched.** The correction commit
(f529233) changes prose, comments and tests only.

---

### 1. *"an OFF-SITE listing has no AppBlock at all, **ever**"* — overstated

`packages/civitai-db-schema/prisma/schema.full.prisma:2849-2853` contradicts it directly:

> `appBlockId` is *"Set for **EVERY** backfilled row — on-site **AND** the #2821 off-site
> rows (both come from an AppBlock). It is **NOT** a kind discriminator: discriminate on
> `kind`, never on appBlockId nullness."*

Only a **natively-created** off-site listing leaves it NULL, and `mapAppBlockToListing`
can mint `kind:'offsite'` with a non-null `appBlockId`. That shape measured **0 rows in
production, 2026-08-11** (`src/components/Apps/appListingEditorTabs.ts:109-113`).

**Restated as:** empirically zero today, structurally possible, and a backfilled class
exists. Behaviour is unchanged — the widening is still correct; the clause that *had* to
go was `kind`, and dropping `appBlockId: null` alongside it is what covers the backfilled
class too. The same overstatement was in the service comment, the schema comment and a
test comment; all three are fixed.

### 2. *"a shadow **matched** the previous clause"* — true for on-site parents only

`beginListingRevision` clones `kind: parent.kind` (`offsite-listing.service.ts:1078`). So
a shadow of an **on-site** parent is `kind:'onsite'`, `appBlockId` NULL, status `draft` —
which did match the old clause. A shadow of an **off-site** parent is `kind:'offsite'` and
did **not**.

The claim is now scoped. **The conclusion is unaffected:** `revisionOfId: null` is
load-bearing under the *new* clause regardless, and is now the only thing excluding
shadows.

### 3. *"Before, the `where` filtered a stranger's row out before the gate ran"* — wrong for the draft shape

The old `where` had **no owner predicate either**, and `main` already carries the test
that proves it:

> `src/server/services/blocks/__tests__/offsite-listing.get-my-listing-for-app.service.test.ts:596`
> — *"a draft owned by ANOTHER user → NOT_OWNED (ownership still enforced on the slug
> path)"*

For a pre-approval on-site draft, `NOT_OWNED` was **already** the answer on `main`. The
`NOT_FOUND → NOT_OWNED` delta is real, but it is a **widening of an existing delta over a
larger row set** — not a new behaviour. The body now says so.

### 4. *"store slugs are already public in the catalog"* — holds only for **approved** listings

`getAppDetail` is approved-only (`app-listings.router.ts:1337`; the service returns null
for *"a missing OR non-approved (draft/pending/rejected) listing"*). The newly-addressable
set is precisely the **non-approved** rows a logged-out browser cannot see, so that
justification does not cover the delta it was offered for.

**Replaced with the stronger, true argument.** The same cohort already has a *complete*
existence oracle over the *entire* slug namespace, via `submitExternalListing`
(`src/server/services/blocks/offsite-listing.service.ts:238-241`):

```ts
const existingListing = await dbRead.appListing.findUnique({
  where: { slug },            // no kind / status / revisionOfId filter
  select: { id: true },
});
if (existingListing) throw slugTakenError(slug);   // `slug "X" already taken`
```

**Residual, stated honestly:** that oracle is a **mutation** capped at **10/hour**
(`app-listings.router.ts:477-483`); `getMyListingForApp` is an **unrate-limited
`.query`**. This PR does not create the disclosure — it removes its budget. Cohort is
`appDeveloperProcedure` + the store Flipt gate. Tracked as **#4003** rather than fixed
here, because tightening it is a design call about that cohort. If you would rather have
it in this PR, `rateLimit({ limit: 60, period: 60 })` matching `getAppDetail` is the whole
fix — say the word.

---

## Also fixed in f529233 (not body claims — code comments and tests)

**The function's own JSDoc still documented the deleted predicate.** The PR had updated
the inline comment and the schema JSDoc and missed
`getMyListingForApp`'s doc block, which still described the slug arm as falling back to
`(kind='onsite', appBlockId IS NULL, status='draft')`. Rewritten, including selector
precedence.

**Five sites still asserted the proc is `appBlockId`-only** — `app-capabilities.constants.ts`,
`appListingEditorTabs.ts`, `edit.tsx`, the tRPC JSDoc, and `app-access.service.test.ts`.
🔴 **`CAPABILITIES_BY_KIND.offsite.listingMedia` stays `false`** — but the reason is now
correctly attributed to `editorTabsFor`'s `ctx.appBlockId != null` clause
(`appListingEditorTabs.ts:128`), not to the resolver. Flipping the cell is still #3893.

🔴 **One of those was an access-control classification, not prose.**
`app-collaborator.permission-matrix.test.ts (pre-fix :224-230)` filed the proc under `ONSITE_ONLY`
with the rationale *"it is UNCALLABLE [for offsite], which is the same structural
mechanism that makes `submitVersion: false` real."* **That mechanism no longer exists**,
so the classification was wrong rather than mis-worded — the test still passed only
because it called with `appBlockId`. The entry moved into the kind-crossed `MATRIX`, the
off-site half calls it by slug, and its `findFirst` mock was made **predicate-honouring**
so the off-site `ALLOW` cells actually depend on the widening: restoring `kind:'onsite'`
to the service now reddens them (measured). A permissive mock would have passed under the
old clause and asserted nothing.

**`deleteOnsiteDraftListingForSlug` (`publish-request.service.ts:1476`)** keeps a
same-looking narrow clause and **must** — it is a `deleteMany` with no owner predicate and
nothing downstream to refuse it. A cross-reference to #3984 is now in the code so a future
"consolidation" cannot widen a destructive query.

**The characterization test adds no coverage, and now says so.** That file's `ownedRow`
already defaults to `kind: 'offsite'`, so the pre-existing approved-shadow case above it
already drives the same path. Kept only for the open question it carries — which now notes
that `REVIEWABLE_LISTING_KINDS = ['onsite','offsite']`, i.e. the revision
submit → approve → apply chain accepts off-site **end to end**, so the question leans
further than the body originally implied.

## Two claims were spelled but not tested — both mutants now die

| mutant | before | after |
| --- | --- | --- |
| `if (!listing && slug)` → `if (slug)` — selector precedence | **SURVIVED 44/44** | **killed**, on `expected 'apl_other' to be 'apl_block'` |
| `+ status: { not: 'removed' }` — the "any status" widening | killed **only** the structural `toHaveBeenCalledWith` pin; no behavioural case noticed | **killed twice** — the pin *and* a new behavioural row |

New tests: `appBlockId` wins over a slug naming a **different** listing (both rows owned
by the same user, pairwise distinct on three observable fields, so ownership cannot be
what decides it); and a `removed` / a `rejected` listing resolving **by slug** and
returning the right `editBlockedReason`.

The original battery was re-run after the fixes and still reds: restore `kind:'onsite'`
(7 killed), restore `status:'draft'` (6), restore `appBlockId: null` (2), drop **only**
`revisionOfId: null` (2). HEAD: 189/189 green across the four touched files.

## Gate

`typecheck` 0 errors · `tsc -p tsconfig.tests.json` 738 pre-existing repo-wide, **0 in the
11 touched files** (positive control: 4 found in a known-bad file) · `lint` 4317 problems
(368 errors) — identical to the `main` baseline, **0 errors in touched files** ·
`test:unit:run` **1092 passed | 1 skipped (1093 files)**, **17094 passed | 21 skipped
(17115 tests)**, both `unit` and `unit-native` projects.

PR #3998, opened 2026-08-17
fix(creator-studio): read the all-time comment count from a live source

JustMaier:
## The bug

`/analytics` renders a line reading **"All-time on your images: N reactions · N comments"**. The comments half came from `sumMerge(comments)` on `default.image_metrics_user` — a one-off backfill that **nothing writes**.

- 324,689 rows, max `userId` **9,659,529**, against current user ids past **12.5M**.
- So it is frozen for every creator it covers and returns **0** for every creator who signed up since — presented as a real figure, with no indication it isn't one.
- Concretely: userId 10752156 has **1,288** comments on their images and was shown **0**. 12094877 → 1,067 shown as 0. 9815659 → 1,028 shown as 0.

## The fix

Comments now come from Postgres — `Thread.commentCount` summed over the image's own thread **and its descendants**, so threaded replies are counted:

```sql
WITH roots AS (
  SELECT t.id FROM "Thread" t JOIN "Image" i ON i.id = t."imageId" WHERE i."userId" = $1
)
SELECT coalesce(sum(t."commentCount"), 0) AS comments
FROM "Thread" t
WHERE t.id IN (SELECT id FROM roots) OR t."rootThreadId" IN (SELECT id FROM roots)
```

Measured on prod, wall clock including ~0.5s of tunnel + node startup, behind the existing 1h cache:

| creator | images | time |
|---|---|---|
| 994880 (platform's largest) | 806,440 | 2.3s |
| 3513392 | 27,665 | 2.1s |
| typical | — | 0.7–1.2s |

No new table, no rollup, no DDL.

## Why not ClickHouse — the part worth reading before you try again

**`comments.entityId` is the CommentV2 id, not the entity commented on**, for every `type` except `Model`.

Verified against Postgres: CH row `type='Image', entityId=2309025` is `CommentV2.id = 2309025`, whose thread has `imageId = 139928324`.

The tell is in the table itself — 970,679 rows of `type='Image'` over **970,422 distinct entityIds** (1.0003 comments per "image"), and `Image`/`Post`/`Comment`/`Review`/`Bounty` all occupy one shared id range **45,825–2,309,025** that overlaps between them. Real image ids are past 139M. Only `Model` behaves like a real entity id (1.27M rows / 259k distinct).

Joining `comments.entityId` to `images_created.id` therefore **does not error** — image ids that low exist — it silently credits each comment to whatever ancient image happens to share its number. The table's name gives no hint of this.

**`entityMetricTotal_v3` has two per-image comment metrics and neither survives a ground-truth check:**

| ownerId | `commentCount` | `Comment` | Postgres (top-level) | Postgres (incl. replies) |
|---|---|---|---|---|
| 3513392 | 11,256 | 393 | 11,946 | 13,083 |
| 3446224 | 6,216 | 3,072 | 4,411 | 7,315 |
| 1279061 | 5,423 | 4,774 | 4,996 | 8,673 |
| 7154640 | 5,082 | 464 | 2,832 | 4,279 |

`commentCount` runs -6% to +79% against truth; `Comment` is short by up to 97%. This is not an ownership-map problem — `images_created` covers these creators fully (it runs 2–20% *over* Postgres, i.e. it retains deleted images). `entityMetricEvents` has no `commentCount` arm at all, so that metric is another dead seed, the same class of bug as `image_metrics_user`.

`Thread.commentCount` agrees with a raw `CommentV2` count exactly (11,946 and 550 on two creators; one drifts by 7 of 2,832). That agreement is why it's the source.

## Labels, both sides of the line

`reactions_owner_scores_mv` aggregates `reactions.ownerId` across **all** entity types, but `/analytics` said *"All-time on your images"* and `/analytics/audience` said *"All-time image reactions"*. Both scoped an all-content number to images.

The reactions **query** is unchanged on purpose — it is what `UserMetric.reactionCount` uses, so it agrees with the creator's profile; changing the query would break that agreement, changing the label doesn't. Now:

- `/analytics` — "All-time: **N** reactions across your content · **N** comments and replies on your images"
- `/analytics/audience` — "All-time reactions (all content)" / "All-time image comments (incl. replies)"

Comments on models, posts and articles are still not counted; that is unchanged and out of scope here.

## Notes

- Cache key bumped `analytics:alltime:v2` → `v3`; the meaning of the number changed, so warm keys must not survive the deploy.
- `getAllTimeTotals` was the only live reader of `image_metrics_user`. The other hits are two docs and a comment in `backfill-image-scores.ts` explaining that it moved off.
- Verified: `pnpm run typecheck` in `apps/creator-studio` — 8,922 files, **0 errors** (3 warnings, all pre-existing in `packages/civitai-ui`).

https://claude.ai/code/session_01MrEFDxhAYXPat3ExAucwho


JustMaier:
## Adversarial review found two real defects — both fixed in f201b8a

### 1. The query seq-scanned `Thread` on every cache miss

`WHERE t.id IN (roots) OR t."rootThreadId" IN (roots)` makes both indexes unusable. `EXPLAIN (ANALYZE, BUFFERS)` on prod:

| userId | before | after |
|---|---|---|
| 12500034 (2 images, **0 comments**) | 683 ms, 44,961 buffers, `Rows Removed by Filter: 5,408,209` | **46 ms, 54 buffers** |
| 3513392 (27,665 images) | 1,500 ms | **~950 ms** |
| 994880 (802,838 images) | 4,251–4,725 ms, ~16 GB touched | **1,636 ms** |

The PR body framed cost as scaling with creator size. It didn't — there was a ~700 ms floor paid by **every** creator on the SSR blocking path, including ones with no comments at all. Split into two `MATERIALIZED` arms it's an index path throughout, no seq scan at either end.

### 2. It counted the creator's own comments as engagement received

`update_thread_comment_count` fires on every `CommentV2` insert/delete with no author check, so summing `Thread.commentCount` credits a creator for answering their own commenters. Share of the raw total that was self-authored:

| userId | total | self | share |
|---|---|---|---|
| 1421581 | 4,370 | 3,121 | **71.4%** |
| 2895 | 5,212 | 3,173 | **60.9%** |
| 1279061 | 8,673 | 4,995 | **57.6%** |

This renders as a StatCard on the **audience** page, satisfying a checklist item that reads *"total comments received"*. A creator who replies to everyone saw up to 2.4x their actual audience engagement. Now counts `CommentV2` rows with `cv."userId" <> uid` — which costs nothing extra, since the rewrite above was needed regardless.

### Also noted in the code, not fixed

`roots` joins `Image`, and `Thread.imageId` is `onDelete: SetNull` — ~234k orphaned root threads hold ~179k comments platform-wide. So this is "all-time on images that still exist", and the tile can fall month over month. That's a genuine semantic change from the frozen backfill, so it's documented next to the query rather than papered over.

### Out of scope, worth someone's attention

`createKyselyClients` builds the pool with no `max` (pg default 10) and no `statement_timeout`. It's shared by every app, so I haven't touched it — but ten concurrent cold-cache analytics loads can occupy the whole replica pool, and the cache is documented fail-open, so a Redis outage makes every page load a cold one. Worth a separate PR from whoever owns `packages/civitai-db`.

### Claims that survived the attack

- **No double-counting**, proved against prod rather than by reading: of 5,408,209 threads, `rootThreadId = id` → 0, `parentThreadId IS NOT NULL AND rootThreadId IS NULL` → 0, `imageId IS NOT NULL AND rootThreadId IS NOT NULL` → 0.
- **The `entityId` identification is now exact, not inferred.** `max(entityId)` for `type='Image'` in ClickHouse is **2,309,120**; `max("CommentV2".id)` in Postgres is **2,309,121**. One-to-one, no ratio argument needed.
- **`entityMetricTotal_v3` fails per-image too**, so the four-creator sample wasn't cherry-picked: for 3513392's 12 most-commented images, CH `commentCount` gave 2 vs 7, 2 vs 6, 4 vs 8, 10 vs 7 — errors in both directions. The `Comment` metric exists for 1 of the 12.
- **No cheap correct ClickHouse source was missed.** `commentEvents` (1.29M rows, live) has no entity id and no owner id at all.
- **`reactions.entityId` does *not* share the defect** — max 139,929,797, ~2.1 rows per distinct id vs `comments`' 1.0003. Five other queries key off it safely. Recorded in the block comment so nobody over-generalises the fix.
- No injection surface (Kysely parameterises `$1`), no missed callers (exactly two, both relabelled), cache key bump correct.

JustMaier:
## Labels and typography — `9efebc8` (final)

Both labels carried wording that existed to cover the two bugs this branch fixed. With the creator's own comments excluded, "and replies" is the least load-bearing phrase in the sentence: a creator has no prior that a comment count excludes replies, and being wrong about it costs them nothing. Being wrong about **scope** is what shipped broken — twice, in both directions.

**`/analytics`**

> All-time: **N** reactions across your content · **N** comments on your images

**`/analytics/audience`** — these get read as a pair, so both parentheticals now answer the same question. Before, one described scope and the other composition, which invites a creator to read the second expecting scope, get composition, and take nothing from either.

| before | after |
|---|---|
| All-time reactions (all content) | All-time reactions **(all your content)** |
| All-time image comments (incl. replies) | All-time comments **(your images only)** |

"incl. replies" moves to the hover, alongside the live-images-only caveat, where composition detail belongs.

**Typography.** The line was `text-xs text-dark-3` — the smallest, dimmest text on the page, carrying the most qualification. Every typographic signal said *footnote* while the content said *read carefully*, which is the same failure as unread precision expressed in CSS. Now `text-sm text-dark-2`.

### Review addendum, also applied — `61e0527`

- **Corrected an overreach in my own comment.** It claimed `reactions.entityId` being sound is "why every other query in this file keys off it safely". Only `fetchTopMedia` uses it; the rest use `reactions.ownerId`, and two use `views.entityId`, **which nobody has checked**. A paragraph warning against over-generalising shouldn't over-generalise in the other direction, and an open question stated as closed is worse than no comment at all. `views.entityId` is left explicitly unasserted.
- **Evidence widened 2 days → 30.** `Image_Create` runs to 139,931,065 over 19,428,878 rows / 6,203,813 distinct ids (3.13 per image), and each `type` stays inside its own entity's id space rather than sharing one — `CommentV2_Create` tops out at 2,309,122 against `max("CommentV2".id) = 2,309,121`. Six sampled rows join to a live `Image` whose `userId` equals the CH `ownerId`. That's conclusive rather than suggestive.
- **Recorded *why* `comments` is the odd one out**, since "ClickHouse event table with an `entityId`" is a shape people will reach for again: `reactions` rows come from the reaction handler, which knows the entity; `comments` rows come from the comment handler, which knows the comment. `type` says what was commented *on*, `entityId` says *which comment*. Invisible from the schema.
- Both inline UI comments now state that self-authored comments are excluded on purpose. Without the reason attached, `cv."userId" <> uid` reads like a redundant predicate and gets deleted.

### Independently re-measured after the fixes

No `Seq Scan` in any plan. 57 ms at the two-image creator (I measured 46 ms), 1,612 ms at 994880 (I measured 1,636 ms). Self-exclusion exact on four creators, to the row: 13,083−1,117=**11,966** ✅, 8,673−4,995=**3,678** ✅, 4,370−3,121=**1,249** ✅, 5,212−3,173=**2,039** ✅.

### Deliberately not fixed

`hidden` / `tosViolation` comments are still counted — 1 of 8,673 in the largest sample, 0 in the others. A predicate there needs its own product answer about what a hidden comment means to a creator, which isn't a correctness question and doesn't belong on this branch.

PR #4001, opened 2026-08-17
feat(monitoring): alert on stale ClickHouse refreshable views

JustMaier:
## Why

Seven refreshable materialized views run in production and **nothing alerted on any of them**. Confirmed from `system.view_refreshes` rather than from documentation:

| view | period | measured duration |
|---|---|---|
| `buzz.transactions_final_mv` | 15s | 0.13s |
| `default.entityMetricTotal_v3_refresher` | 1m | 3.20s |
| `default.entityMetricTotal_v3_refresher_additive` | 1m OFFSET 15s | 9.59s |
| `default.entityMetricDaily_today_v2_mv` | 2m | 2.21s |
| `default.entityMetricDailySeal_v2_mv` | 1d OFFSET 1h | 3.80s |
| `default.image_views_daily_by_owner_mv` | 1d OFFSET 2h | 4.33s |
| `default.impressions_daily_by_owner_mv` | 1d OFFSET 4h | 0.06s |

Six of the seven `APPEND` into `SummingMergeTree` targets, so a missed refresh **cannot be repaired by re-running it** — a second run adds to the first and the number silently doubles. Repair is a manual `DROP PARTITION` plus a single re-insert. `entityMetricDaily_today_v2_mv` is the exception (`TO`, not `APPEND`).

Severity therefore differs per view, and the job records which is which:

- **`impressions_daily_by_owner_mv` — page.** Its source `default.impressions` has a 30-day TTL, so a miss nobody notices for a month is permanently unrecoverable.
- **everything else — ticket.** Re-derivable from sources that still hold the rows.

## What it emits

A `clickhouse-refresh-monitor` job on `*/1 * * * *` (dedicated), one query against `system.view_refreshes`, ~130ms measured. Per view, labelled `view="<database>.<name>"`:

| gauge | alert on it? |
|---|---|
| `civitai_app_ch_refresh_staleness_seconds` | **yes — this is the signal** |
| `civitai_app_ch_refresh_staleness_limit_seconds` | it is the comparand |
| `civitai_app_ch_refresh_errored` | leading indicator only; a stuck view has this at 0 |
| `civitai_app_ch_refresh_retries` | no — resets to 0 on success, informational |
| `civitai_app_ch_refresh_present` | 0 means dropped/renamed; already folded into staleness |
| `civitai_app_ch_refresh_append_only` | static; use it to route severity |

Plus three unlabelled: `monitor_last_run_timestamp_seconds` (alert on it going stale — the per-view gauges cannot report their own absence), `views_monitored`, `views_missing`.

One generic rule covers all seven, no thresholds infra-side:

```
max by (view) (civitai_app_ch_refresh_staleness_seconds)
  > max by (view) (civitai_app_ch_refresh_staleness_limit_seconds)
```

## Three decisions worth arguing with

**Staleness, not status.** `status` reads `Scheduled` both for a view that refreshed ten seconds ago and for one whose scheduler stopped advancing, so a stuck view looks healthy on it. `now() - last_success_time` is a number that has to keep moving, and it is computed inside ClickHouse so it is immune to app timezone and host clock skew. There is a test for exactly the stuck-but-not-errored case, and mutating the alert condition to read `errored` fails 5 of the 9 tests.

**Limits are period + margin, from measured numbers.** Healthy peak staleness is period + duration, hit just before the next success. Sub-minute and minute views get ~10x their period (5m / 10m / 15m), which absorbs a CH restart or deploy blip while still catching a stop inside one budget. Daily views get period + 2-3h (27h / 27h / 26h): one missed slot fires within hours, far inside even the 30-day impressions window, and ClickHouse schedules these on a fixed offset so no legitimate jitter reaches it.

**A missing view is maximally stale, not silently absent.** The expected-view list is hardcoded rather than discovered. A view that is dropped, renamed, or replaced by a differently-named successor stops appearing in `system.view_refreshes`, and prom-client never resets — a discovery-driven monitor would keep publishing the last healthy value forever and a firing alert would silently resolve. Same reasoning as the "no TTL" note in `metric-reconciliation-audit.ts`. Absent or never-successful publishes a placeholder past every limit, so the ordinary rule fires and no second alert has to exist. The anchor is set last and only on a complete run, so a scrape can never see a fresh monitor over unpublished view gauges.

## A gap this does NOT close

**This covers the rollup half only. Nothing covers the raw half.** If impressions stop *arriving*, the number goes flat and every gauge here stays green — the refresh succeeded, it just had nothing to read. Worse, the shared ClickHouse client runs `async_insert: 1, wait_for_async_insert: 0`, so insert errors never reach the caller and `insert()` resolves on buffering rather than on data being queryable. "No errors" is not evidence that writes are succeeding.

Deliberately out of scope: it is a different mechanism (volume/shape anomaly detection over arrival rates, not a scheduler liveness check) and wants its own design. Naming it here so it is a decision rather than an oversight.

## Verification

- All seven views and their periods read from production, not from a description of them.
- The job's exact query run against production: 7 rows, keys match the hardcoded list exactly, every staleness inside its limit.
- Mutation-tested: zeroing the missing-view placeholder fails 2 tests; deriving `breached` from `errored` instead of staleness fails 5.
- `pnpm run typecheck` 0 errors; `lint` clean on all three changed files (the one `prefer-const` in the jobs route is pre-existing, at an untouched line); `prettier:write`; `test:lint-rules` 220 passed; full unit suite **17,208 passed / 1 file skipped**.

No ClickHouse DDL required.

https://claude.ai/code/session_01RP49ueKWBCwDw3Ypj5Mv7U

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -237,12 +237,20 @@ pnpm run release:major    # Major release (x.0.0)
 - **Redis / caching** — `redis/client.ts` (clients incl. sysRedis), `redis/caches.ts` (`createCachedObject` defs + TTLs, e.g. `imageMetaCache`, `tagIdsForImagesCache`), `utils/cache-helpers.ts`.
 - **Orchestrator (generation)** — `orchestrator/get-orchestrator-token.ts` (`getOrchestratorToken`), `services/orchestrator/orchestrator.service.ts`.
 - **Auth** — `auth/get-server-auth-session.ts`, `auth/session-verifier.ts` + `auth/session-cache.ts` + `auth/session-invalidation.ts`, `auth/token-claims.ts`, `auth/civ-cookie.ts`, `auth/oauth-bridge.ts`, `auth/route-guard.ts`, `auth/bearer-token.ts`. Shared logic lives in `packages/civitai-auth`; the hub itself is `apps/auth`.
-- **Jobs (cron)** — `jobs/job.ts` (runner) + individual jobs `jobs/*.ts` (e.g. `entity-moderation.ts`, `search-index-sync.ts`).
+- **Jobs (cron)** — `jobs/job.ts` (runner) + individual jobs `jobs/*.ts` (e.g. `entity-moderation.ts`, `search-index-sync.ts`). **Adding a job to the `jobs` array in `src/pages/api/webhooks/run-jobs/[[...run]].ts` IS the scheduling — see below.**
 - **Metrics / analytics** — `metrics/*.metrics.ts` (ClickHouse-backed entity metrics), `clickhouse/`.
 - **DB** — `db/db-helpers.ts` (raw pg-pool config: `connectionTimeoutMillis`, labeled pool gauges), Prisma client. **Schema is `packages/civitai-db-schema/prisma/schema.full.prisma`, and migrations are applied manually — see the Database rule above.**
 - **Telemetry** — `src/instrumentation.node.ts` (OTEL: Prisma/Redis/HTTP auto-instrumentation + custom `withSpan()` from `utils/otel-helpers.ts`), `schema/track.schema.ts` (ClickHouse action/event tags), `prom/client.ts`.
 - **Health** — `src/pages/api/health.ts` runs sub-checks concurrently, each raced at `HEALTHCHECK_TIMEOUT` and the whole set raced against an overall deadline, reporting partial results as checks settle. Checks can be suppressed or demoted to non-critical via the `HEALTHCHECK_DISABLED` env var and the Redis-backed `DISABLED_HEALTHCHECKS` / `NON_CRITICAL_HEALTHCHECKS` keys.
 - **Other server domains** — `games/` (new-order/ratings), `webhooks/`, `paddle/` + `coinbase/` (payments), `notifications/`, `signals/`, `rewards/`; S3 helpers at `src/utils/s3-utils.ts`.
+
+### How a scheduled job actually gets scheduled
+
+**Add it to the `jobs` array in `src/pages/api/webhooks/run-jobs/[[...run]].ts` and give `createJob` a real cron string. That is the whole registration.** A separate scheduler service reads the array — names and crons — from `src/pages/api/internal/get-jobs.ts` and registers a recurring trigger per entry, which then calls `run-jobs` back. So the cron string is load-bearing, not documentation.
+
+Nothing *inside this repo* reads `Job.cron`, which is what makes this easy to get wrong. Grepping for a consumer finds none, and the infra repo contains a handful of hand-written Kubernetes CronJobs that curl `run-jobs` directly and whose comments say registration in the array "does NOT schedule anything". Those are per-job exceptions, not the mechanism. Two independent agents reading only that evidence concluded, wrongly, that a new job would never run and that the fix was another CronJob — which would have created a second scheduling path for the same job.
+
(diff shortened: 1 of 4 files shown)


JustMaier:
Alerting side is up: **civitai/talos-infra#1072** — Grafana dashboard, the warning tier, two criticals (the impressions view and a monitor deadman), plus a registered promtool suite and a mutation harness.

It does not block on this PR and this PR does not block on it. Every expression there selects on a gauge that ships here, so until this deploys they evaluate empty and nothing can fire; test 1 of that suite asserts exactly that with zero input series. Merge order is free — just expect no data on the dashboard until this one is out.

Worth noting for anyone reviewing the thresholds here: the alert rules carry none of them. This job publishes `civitai_app_ch_refresh_staleness_limit_seconds` per view and the rules compare the two series, so a limit change lands in this file only, next to the measured period and refresh duration it came from.

JustMaier:
Adversarial review came back with six findings. Three were real defects and are fixed in 8cb3c56bee; the rest are answered below.

**Fixed**

1. **`append_only` encoded the wrong property.** Six of the seven APPEND, but what decides repairability is the target engine, and only two are SummingMergeTree. From `system.tables`: `image_views_daily_by_owner` and `impressions_daily_by_owner` are SharedSummingMergeTree; `buzzTransactions`, `entityMetricDailyAgg_history_v2` and `entityMetricTotal_v3` are SharedReplacingMergeTree and dedupe on their sorting key; `entityMetricDailyAgg_today_v2` is a plain `TO`. Keying severity off the old gauge would have marked six views unrepairable when two are. Renamed to `unrepairable_by_rerun`, set for those two only, header corrected.

2. **The liveness-anchor test could not fail.** Both `run()` calls landed inside the same millisecond, so `toBe(anchor)` was satisfied by a mutant that stamps the anchor unconditionally at the top of the job — the exact regression the test is named for. Confirmed silent (9 passed, 678ms). Now uses fake timers and fails with a 60s discrepancy.

3. **The Axiom log promised exception text it never selected.** The query read `exception != ''` and threw the string away, so an on-call got "view X is errored" and a manual ClickHouse round-trip. Now carries it, with a test asserting the text reaches the log.

**Also changed**

- The header now records the working alert rule. A bare `max by (view)` reads an abandoned pod's frozen gauges and keeps a resolved incident firing until that pod rolls. The shipped rule in talos-infra#1072 already had the pod-selection form and a test for it, but the rule quoted in this file's header was the naive one — bad documentation of a correct alert.
- `impressions` limit 26h → 27h. It had the tightest tolerance of the three daily views while being the only page-severity one, which is backwards, and 26 vs 27 is indistinguishable against a 30-day recovery window.
- Non-finite staleness falls back to the placeholder rather than `Math.max(0, NaN)`, which is NaN and reads as not-breached.
- One fixture now uses the unquoted numeric shape ClickHouse actually returns (`output_format_json_quote_64bit_integers: 0`); the others exercised only the defensive branch.

**Not changed, deliberately**

- **A monitor that stops has one backstop, and that is by design.** A throwing query, an unconfigured client, and a cron that was never created all leave the view gauges frozen at healthy. The deadman on the anchor covers all three, and it is written down in the header now with the aggregation it needs. A separate `monitor_query_failed` gauge would be a second signal for the same event.
- **`default.impressions` is currently empty**, so that view refreshes successfully over nothing and this monitor reads it as healthy. Real, and the same shape as the raw-side gap already named in the PR body. Still a different mechanism; not folding it in here.

Full suite 17,209 passed, typecheck 0 errors, lint clean, `test:lint-rules` 220 passed.

PR #4002, opened 2026-08-17
feat(event-engine): emit User.commentCount so comments can be read per creator

JustMaier:
Follow-up to #3998, which fixed Creator Studio's all-time comment count by reading Postgres. This starts making ClickHouse able to answer it — comments are currently the only creator-facing metric it can't.

> **⚠️ Ships with a manual prerequisite.** One row must exist in `default.entityMetricKind` before this metric means anything. See "Required before deploy" below. Merging is safe without it (nothing reads the metric yet); **deploying without it is not**.

## The gap

Comment metrics have always been keyed to the **entity commented on**, never to whoever owns it. So "comments on this creator's images" has no cheap answer: getting from a creator to their images means a join over the whole image space (measured at 948M rows / 265 MiB).

The owner isn't a new column — it's `entityType='User'` with the owner's id as `entityId`, which is how these already work in prod (`entityMetricTotal_v3`):

| `entityType='User'` metric | rows | distinct users |
|---|---|---|
| followerCount | 1,094,472 | 880,081 |
| reactionCount | 803,655 | 594,147 |
| postCount | 87,241 | 57,024 |
| **commentCount** | — | — |

## The change

The thread lookup already runs, so the owner comes from that same memoized query via four `LEFT JOIN`s rather than a second round trip. `EXPLAIN (ANALYZE, BUFFERS)` on prod: all nested-loop PK lookups, `shared hit=13`, **0.041 ms**. No new scan.

**Covers Post/Image/Article/Bounty**, matching both the entity metrics this handler already emits and `User.reactionCount`, which is all-content. An image-only `commentCount` beside an all-content `reactionCount` is the asymmetry that produced two wrong labels in Creator Studio.

**Self-authored comments excluded.** A creator answering their own commenters isn't engagement received, and left in it dominates — 71.4% / 60.9% / 57.6% of the raw total for userIds 1421581 / 2895 / 1279061. Matches `reactions_owner_scores` and #3998's tile.

## 🔴 Required before deploy — one manual ClickHouse row

`default.entityMetricKind` decides how each `(entityType, metricType)` pair is rolled up, and **an unregistered pair silently takes the wrong path**:

- `entityMetricTotal_v3_refresher_additive` — pairs `IN (kind='additive')` → `sum(total)`. Counting.
- `entityMetricTotal_v3_refresher` — the **complement**, `NOT IN (...)` → `sum(present)` where `present = toInt64(argMaxMerge(latest) > 0)` per userId. Presence.

The registry has 82 rows, all `additive`, with 13 `User` pairs and **no `commentCount`**. So without registration this metric becomes *the number of distinct commenters whose latest comment event was a create* — capped at 1 per commenter, zeroed for anyone whose last action was a delete. A believable nonzero number meaning something else, which is exactly the class of bug #3998 existed to fix.

```sql
INSERT INTO default.entityMetricKind (entityType, metricType, kind, updatedAt)
VALUES ('User', 'commentCount', 'additive', now64(3));
```

Safe to apply before the handler ships — the pair has no rows yet. **`grep -r entityMetricKind` over the repo returns zero hits**: it's manual prod data with no representation in code, so nothing in CI can catch a missing registration. Given this app also has no working tests or lint, review is the only gate. A guard comparing emitted pairs against the registry would close the whole class; that's worth its own task.

## Known properties, measured not assumed

- **Bulk-delete drift.** `CommentV2_userId_fkey` is `ON DELETE CASCADE` and Debezium stamps one transaction with one `source.ts_ms`, while `entityMetricEvents_month`'s `createdAt` is second-precision. Per-image rows survive (distinct `entityId`); the User rows all carry `entityId = ownerId`, so deleting an account that left 500 comments on one creator yields 500 image `-1`s and **one** User `-1`. On the create path this is negligible — 7 days of prod: 13,556 eligible comments, **11** lost to same-owner/same-second collapse (0.08%). Pre-existing; `User.reactionCount` has it too.
- **`-1`s without a matching `+1`.** Nothing emitted this metric before, so from deploy every delete of a *pre-deploy* comment decrements a counter that started at 0. It can go negative. The planned history synthesis therefore needs a cutoff reconciling post-deploy deletes of pre-deploy comments — it is not a straight roll-up.
- **Redis vs ClickHouse drift on deploy.** `incrementOnce` adds a live `commentCount` field to `metrics:User:<id>` hashes; `CacheDriftMonitor` (`mew_cache_drift_ratio`) compares that against ClickHouse. Expect movement. Reader side is unaffected — the main app's MetricService hardcodes its field list.

## Generator fix (commit 2)

`generate-types` discovered metrics by running **this app's** handlers against faker data, so anything written by the main app, a cron, or a backfill was invisible — and it rewrote the file wholesale, so regenerating **deleted** those. A clean regen dropped `downloadCount`/`generationCount` from Model and ModelVersion and renamed every Image reaction, breaking `images.feed.ts`.

It now unions discoveries with the committed `ENTITY_METRIC_TYPES`. A plain `pnpm generate-types` is additive and idempotent (verified twice, only the `Generated:` line differs), and the committed `metric-types.ts` **is** clean generator output — no hand-edit remains. Stated tradeoff, in the code: the generator can no longer remove anything, so a retired metric needs a deliberate hand-deletion. It also fixes a latent flake — the debug fakers pick entity types randomly, so a run that never sampled `bounty` previously dropped `Bounty.commentCount`.

## Verification

- `pnpm typecheck` in `apps/event-engine` — exit 0, package dir confirmed
- `npx eslint` per-file on changed files — clean
- 🔴 `pnpm lint` **crashes on main** (`TypeError` in `@typescript-eslint/consistent-type-imports`, isolated to `src/types/handlers.ts`) and `pnpm test` **fails on main** (`signals.test.ts` unparseable, babel has no TS preset). Both verified against HEAD before these changes; tracked separately.

https://claude.ai/code/session_01MrEFDxhAYXPat3ExAucwho


JustMaier:
## Generator fixed — `e8533a5`. The hand-edit is gone.

`generate-types` discovers metrics by running **this app's** handlers against faker data, so it only sees what this app emits. Anything written by the main app (`updateEntityMetric`), a cron job, or a backfill is invisible to it — and because it rewrote the types file wholesale, regenerating **deleted** those.

It now unions its discoveries with whatever the committed `ENTITY_METRIC_TYPES` already declares. A plain `pnpm generate-types` is:

- **additive** — restores `downloadCount` / `generationCount` on Model and ModelVersion, keeps both `ReactionLike` and `Like`, both `Comment` and `commentCount`
- **idempotent** — running it twice produces no diff (verified)
- **sufficient** — it picks up `User.commentCount` and `Comic` on its own, so the hand-edit and its warning comment are deleted

Both names in each renamed pair are real, which is why keeping both is correct rather than a fudge: `entityMetricEvents` carries `ReactionLike` and `Comment`; the daily-agg rollup carries `Like` and `commentCount`; the old names hold years of history.

**Stated tradeoff, in the code:** the generator can no longer remove anything. A genuinely retired metric must be deleted from `metric-types.ts` by hand. That's the right way round — a missing metric is a type error at the call site, a wrongly deleted one is a silent regression in a feed.

## Confirmed: this is the right repo

I checked the cluster rather than reasoning from the manifests, because `apps/event-engine/k8s/*` still names the old `ghcr.io/civitai/metric-event-watcher` image and reads like the live config:

```
metric-watcher  metric-event-watcher  ghcr.io/civitai/civitai-event-engine:1.9.12  3/3
```

Prod runs the **monorepo** image, matching `apps/event-engine`'s version exactly. The standalone `civitai/metric-event-watcher` repo's final commit (2026-07-21) is a README notice saying the service moved here, that its workflow still pushes an image nothing deploys, and not to cut releases from it.

## Verification, and two things that don't work on main

- `pnpm typecheck` — clean
- `npx eslint` per-file on the changed files — clean
- 🔴 `pnpm lint` **crashes on main** — a `TypeError` inside `@typescript-eslint/consistent-type-imports`, not a lint failure
- 🔴 `pnpm test` **fails on main** — `signals.test.ts` can't parse; babel has no TypeScript preset

Both verified against HEAD before my changes. Neither is touched here, but together they mean **every event-engine change currently ships with no lint and no tests** — worth a task of its own.

JustMaier:
## Adversarial review — one blocker, confirmed independently

**The blocker is real and I verified it against prod myself rather than taking the report.** `default.entityMetricKind` gates the two rollup refreshers, and an unregistered `(entityType, metricType)` pair falls into the **complement** — the presence path, `sum(present)` where `present = toInt64(argMaxMerge(latest) > 0)` per userId.

Checked in prod: 82 registry rows, all `additive`; 13 `User` pairs (`articleCount`, `followerCount`, `reactionCount`, `uploadCount`, …) and **no `commentCount`**.

So deploying the handler without registering the pair would publish `User.commentCount` as *distinct commenters whose latest event was a create* — capped at 1 per commenter, zeroed if their last action was a delete. Nonzero, plausible, and measuring something else. That is the same failure mode as the dead table this whole task started from, and I would have shipped it.

Required row, now documented in the PR body and raised with Justin for approval since it's manual prod data:

```sql
INSERT INTO default.entityMetricKind (entityType, metricType, kind, updatedAt)
VALUES ('User', 'commentCount', 'additive', now64(3));
```

Nothing in the repo mentions `entityMetricKind` — `grep -r` returns zero hits — so no test, typecheck or CI job could have caught this.

## Two corrections to my own PR body, both now fixed

1. **I cited a table that does not exist.** `entityMetricCurrentTotals_v2` is not in prod — `system.tables` has no match. My "groups by (entityType, entityId, metricType) with no entityType filter, so no DDL is needed" described something that isn't running, and the conclusion it supported was wrong as well as unsupported.
2. **My rows/users figures were inverted.** Actual: `reactionCount` 803,655 rows / 594,147 users; `followerCount` 1,094,472 / 880,081. Rows ≥ users by construction. The conclusion held; the evidence didn't.

## Three properties now documented rather than discovered later

- **Bulk-delete drift** — one account deletion cascades in a single transaction with one `source.ts_ms`, and `entityMetricEvents_month.createdAt` is second-precision. Per-image `-1`s survive on distinct `entityId`; the User rows share `entityId = ownerId`, so 500 comments collapse to one User `-1`. Create path is negligible: 11 of 13,556 over 7 days (0.08%). Pre-existing — `User.reactionCount` has it too.
- **`-1` without `+1`** — every delete of a pre-deploy comment decrements from zero, so the metric can go negative. The history synthesis needs a cutoff, not a straight roll-up.
- **Redis/ClickHouse drift on deploy** — `CacheDriftMonitor` compares the live hash field against the ClickHouse total; expect `mew_cache_drift_ratio` to move.

## What survived the attack

Five things I had checked myself, re-confirmed independently: dedupe keys can't collapse the Image and User rows (entityType is in the key); `record.userId` is the author on delete (`payload.before` + `REPLICA IDENTITY FULL`); the memo key hashes SQL text so old cached shapes can't leak; zero prod `Thread` rows carry more than one entity id, so no double-count within a message; generator idempotency holds. Added by the reviewer: all four `Thread_*_fkey` are `ON DELETE SET NULL`, so entity and User decrements are always skipped **together** — never one without the other. Query cost 0.041 ms.

PR #4007, opened 2026-08-17
fix(review): fan the civitai-review lanes out in parallel

JustMaier:
#4005 merged while this change was still in flight, so `main` shipped the skill with `svelte-review`'s inherited "Serial is the safe default" and five lanes would have run one after another.

Parallel is the property that was actually chosen when collapsing the five lanes into one agent was weighed, so this restores it.

- Step 2 is an explicit fan-out: all five lanes launch in a single message with multiple tool uses. They are independent and share no state.
- The fix loop's re-review fans out too, re-running only the lanes whose findings were touched. That is where most runs actually happen.
- Consolidation is three sections rather than one ranked list: intent findings, code findings, and the not-for-this-PR duplication map. A dropped requirement no longer appears in two of them.
- Solo mode is recorded as a rejected decision rather than an omission.
- The security write-up rule is promoted to standalone; it applies to all three sections, not only the last.

Markdown only. Authored by dakota on the original branch, recovered here after the squash merge orphaned it.

https://claude.ai/code/session_01JarLBnBWH1PkYY89yXHrVJ

### DIFF
--- a/.claude/skills/civitai-review/SKILL.md
+++ b/.claude/skills/civitai-review/SKILL.md
@@ -37,7 +37,14 @@ C:\Dev\Repos\work\model-share\_local\docs\plans\<feature>.md
 Absolute path, from any worktree. 🔴 Never resolve `_local/` relatively — it exists only in the primary
 worktree, and a relative write silently creates a second private copy that nobody else reads.
 
-## 2. Fan out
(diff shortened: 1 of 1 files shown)


JustMaier:
Duplicate of #4006 — same commit `fc22f97`, reached from two directions at once. Keeping #4006.

PR #4012, opened 2026-08-17
fix(rewards): turn off the follow and generator feedback rewards

JustMaier:
Deprecates two Buzz rewards at Justin's request, by turning them off at runtime rather than deleting them:

| Reward | Type | Amount | Cap |
|---|---|---|---|
| First daily follow | `firstDailyFollow` | 10 Blue Buzz | 30/day |
| Generator feedback | `generation-feedback` | 4 Blue Buzz | 40/day |

Both keep their definitions and both call sites. The `rewards:config` mechanism from #4019 is what stops them paying, so the timing is whoever applies the row rather than a deploy, and re-enabling is a config edit rather than a revert.

**Past grants are untouched.** The Buzz ledger and the ClickHouse `buzzEvents` history stay exactly as they are; this stops future awards only. The `block-buzz-read` projection keeps redacting both historical type strings, which is correct — old rows are no less sensitive for the reward being off.

## What changed

- **`packages/civitai-db-schema/prisma/migrations/20260817120000_disable_follow_and_feedback_rewards/migration.sql`** — sets `enabled: false` for both types in the `rewards:config` `KeyValue` row. **Manual apply**, per the repo's migration policy; applying it *is* the deprecation.
- **`src/server/rewards/__tests__/deprecated-rewards.gate.test.ts`** — 18 tests covering both call sites, the listing, and the migration itself.

No application code changes.

### The migration merges rather than replaces

The row is shared by every reward, so the `ON CONFLICT` branch preserves other rewards' entries and other fields on these two — re-enabling restores whatever `awardAmount` was configured. `DO NOTHING` would have been wrong in the quietest possible way: with a row already present, both rewards would keep paying while the migration reported success.

Verified read-only on the replica against four prior-row shapes (no table touched — the expression was evaluated over a `VALUES` list):

| Prior row | Result |
|---|---|
| no `rewards` key | key created, unrelated top-level fields preserved |
| `dailyBoost` set, `firstDailyFollow.awardAmount: 7` | both preserved, `enabled: false` added alongside the amount |
| `"firstDailyFollow": false`, `"generation-feedback": null` | replaced cleanly (`\|\|` errors on a non-object operand, hence the `jsonb_typeof` guards) |
| already disabled | idempotent |

## How to check what is actually on

`GET /api/testing/rewards-config?token=$WEBHOOK_TOKEN`. It resolves through the same `resolveRewardConfig` a grant uses, shows the compiled default beside the effective value, and lists any override field that was refused. That endpoint is the answer to "is this reward on?" — a config row is only checkable if you know where to look, and this is the one place that reconciles the row against the definitions for you.

Reverting needs no deploy: flip either `enabled` back, or use the `rewardConfig.set` moderator procedure.

## Tests, and how they fail

`base.reward.config.test.ts` already proves the gate works on a locally-built reward. This suite proves it works **from the two call sites that actually pay**, using the real exported reward objects.

Every disabled-case assertion is paired with an **enabled control that observes a real payment through the same seam**. Without the control, a suite that stopped reaching `sendAward` for any unrelated reason would keep passing while proving nothing.

Mutation-tested rather than assumed:

| Mutation | Result |
|---|---|
| remove `if (!config.enabled) return;` from `apply` | 3 failures, both call sites, with the real transaction in the message (`amount: 10`, `type: firstDailyFollow`) |
| remove `if (!config.enabled) return null;` from `getUserRewardDetails` | listing test fails: `expected [ 'encouragement', …(8) ] to not include 'firstDailyFollow'` |
| typo a reward key in the migration SQL | 3 failures naming the bad key |

### The orchestrator call site needed care

`orchestrator.router`'s `patch` builds `Promise.all(steps.map(step => step.patches.filter(...).map(async ...)))`, which hands `Promise.all` an array of **arrays**. Arrays are not thenables, so they resolve to themselves immediately and the inner promises are awaited by nothing — the reward call there is already fire-and-forget.

That makes a bare `expect(...).not.toHaveBeenCalled()` after `await caller.patch(...)` **vacuous**: it runs before the grant would have happened either way, so it passes whether or not the gate works. Every dependency on that path is mocked async and no timers are involved, so the fix is an exhaustive microtask drain used identically by both the enabled control and the disabled case — a `setTimeout` or a `vi.waitFor` timeout would give the two cases different windows and make the disabled assertion a race. The first version of this test used `vi.waitFor` and failed under mutation via a pretty-format error that aborted the suite and skipped a sibling test: the right outcome for the wrong reason. The current version produces three clean assertion failures and skips nothing.

That fire-and-forget is pre-existing and **not** fixed here. It also defeats `apply`'s deliberate rethrow of non-transport ClickHouse errors, which exists so a schema break surfaces as a loud 500 — at this call site there is nothing to surface into and nothing catching, so it becomes an unhandled rejection. Fixing it means a live generator mutation starts awaiting N reward round-trips it currently does not, and silent failures become 500s: a real behaviour change to a hot path, unrelated to deprecating anything, and its own PR.

Worth knowing for whoever re-enables `generation-feedback`: that call site's failures are invisible.

## What the two rewards are actually doing (ClickHouse, 90 days)

| Type | Rows | Awarded | Blue Buzz | Latest event |
|---|---|---|---|---|
| `firstDailyFollow` | 2,636,232 | 1,952,689 | 19,526,890 | minutes ago |
| `generation-feedback` | 0 | 0 | 0 | none in 90 days |

Two things follow, and neither changes what this PR does.

**`generation-feedback` has not fired once in 90 days.** It is being deprecated as a formality — the gate is what makes that explicit rather than assumed. Why it stopped firing is a separate question; the likeliest candidate is that no patch path reaching `patch` still matches `patch.path.includes('feedback')`, which would mean the call site has been dead code for some time.

**`firstDailyFollow` is very live** — ~29k grants a day, ~217k Blue Buzz a day. Applying this migration is a visible change for a lot of users, so it is worth flipping deliberately rather than incidentally. Surfacing the size, not questioning the decision.

## Findings that survive from the earlier deletion-shaped version of this work

- **The half-deprecation trap.** Removing a reward's barrel export does not stop it granting. `user.controller` imports `firstDailyFollowReward` **direct-from-file**, so the index export was never what gated it. Under config the risk axis moves: it is no longer *how the reward was imported* but *whether the gate covers every door out of `createBuzzEvent`*. #4019 gates all three — `apply` pays, `process` sweeps the backlog, `getUserRewardDetails` advertises — which is why this shape is more robust than the export ever was. Tested rather than assumed.
- **The import asymmetry is per-consumer, not per-reward.** `dailyBoostReward` is direct-imported at `buzz.controller.ts:7` and `blocks.router.ts:20`, and `appBlockReviewReward` at `blocks.router.ts:69`, both while also being barrel-exported. The two rewards here sat on *opposite* sides of it. So the question to ask is "does this consumer go through the barrel", not "was this reward imported properly". Converting those consumers is worth its own PR.
- **Both rewards are `onDemand: true`**, so they write `awarded` or `capped` inline and never the `pending` row that `process` pays. `process`'s disabled-reward sweep is real and tested in #4019, but **neither of these two rewards can exercise it** — a suite built around one of them enters the sweep zero times while looking like it covers it. Asserted here rather than left as a comment, so making one of them processable fails this suite.

## Operational note

Legacy dedup entries keep the pre-#4019 reading until `rewardsDailyReset` clears the hash at 00:00 UTC. That does not affect these two rewards' disable — the gate is upstream of the dedup hash entirely — but it is on the same mechanism and the tracker carries it as a hold on the config deploy.

## Verification

- `pnpm run typecheck` — 0 errors. (Blind to `__tests__`, so the new suite is covered by running it, not by this.)
- `pnpm run lint` — 0 errors on the new file; 9 `no-explicit-any` warnings, consistent with sibling test suites.
- `pnpm run prettier:write` — applied.
- `pnpm run test:unit:run` — full suite green; the new suite is 18 passed.

https://claude.ai/code/session_018gYYkESwT1M2dnVX7HjVYV

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260817120000_disable_follow_and_feedback_rewards/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260817120000_disable_follow_and_feedback_rewards/migration.sql
@@ -1,0 +1,62 @@
+-- ============================================================
+-- Turn off the firstDailyFollow and generation-feedback Buzz rewards
+-- ============================================================
+-- Deprecated at Justin's request. Both rewards keep their definitions and their
+-- call sites; this row is what stops them paying, via the `rewards:config`
+-- mechanism in src/server/rewards/reward-config.ts.
+--
+-- Past grants are NOT touched. The Buzz ledger and the ClickHouse `buzzEvents`
+-- history stay exactly as they are; this only stops future awards.
+--
+-- ⚠️ MANUAL APPLY — the main civitai DB does NOT auto-apply migrations. This file
+-- is committed for history; a HUMAN applies the SQL below per environment
+-- (psql/retool). CI / deploy does NOT run it. Applying it IS the deprecation, so
+-- the timing is whoever runs it, not a deploy.
+--
+-- Reversible without a deploy: set either `enabled` back to true here, or use the
+-- `rewardConfig.set` moderator procedure. Verify the live state through
+-- GET /api/testing/rewards-config?token=$WEBHOOK_TOKEN, which resolves through
+-- the same code path as a grant.
+--
+-- MERGE, NOT REPLACE. The row is shared by every reward, so this must not clobber
+-- overrides someone else set:
+--   - Other rewards' entries are preserved.
+--   - Other fields on these two entries (`awardAmount`, `cap`) are preserved, so
+--     re-enabling restores whatever amount was configured.
+--   - ON CONFLICT DO UPDATE rather than DO NOTHING: with DO NOTHING, an existing
+--     row would leave both rewards paying while this migration reported success.
+--   - A non-object entry (`"firstDailyFollow": false`, `null`) is replaced rather
+--     than concatenated, because `||` errors on a non-object jsonb operand.
+INSERT INTO "KeyValue" ("key", "value")
+VALUES (
+  'rewards:config',
+  '{"rewards":{"firstDailyFollow":{"enabled":false},"generation-feedback":{"enabled":false}}}'::jsonb
+)
+ON CONFLICT ("key") DO UPDATE
+SET "value" = COALESCE("KeyValue"."value", '{}'::jsonb) || jsonb_build_object(
+  'rewards',
+  COALESCE(
+    CASE
+      WHEN jsonb_typeof("KeyValue"."value" -> 'rewards') = 'object'
+      THEN "KeyValue"."value" -> 'rewards'
+    END,
+    '{}'::jsonb
+  ) || jsonb_build_object(
+    'firstDailyFollow',
+    COALESCE(
+      CASE
+        WHEN jsonb_typeof("KeyValue"."value" #> '{rewards,firstDailyFollow}') = 'object'
+        THEN "KeyValue"."value" #> '{rewards,firstDailyFollow}'
+      END,
+      '{}'::jsonb
+    ) || '{"enabled":false}'::jsonb,
+    'generation-feedback',
+    COALESCE(
+      CASE
+        WHEN jsonb_typeof("KeyValue"."value" #> '{rewards,generation-feedback}') = 'object'
+        THEN "KeyValue"."value" #> '{rewards,generation-feedback}'
+      END,
+      '{}'::jsonb
+    ) || '{"enabled":false}'::jsonb
+  )
+);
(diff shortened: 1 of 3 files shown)


JustMaier:
**Why these two, recorded after the fact** (the PR was written without a reason because none had been given, and inventing one would have been worse than leaving it blank):

Justin: *"They're being deprecated because we want to add these new rewards. And so we need to rebalance things."*

So this is a rebalance, not a judgement that either reward was being gamed or was worthless. The two new placement rewards — 10 Blue Buzz per sticker received (100/day) and 20 Buzz per remix accepted (first 5/day) — add roughly 200/day of new earning surface, and these two come off to make room.

Also confirmed: **this will not be applied without an announcement.** Applying the migration is the deprecation, so the announcement and the apply are the same decision.

PR #4031, opened 2026-08-17
fix(event-engine): make lint and test actually run

JustMaier:
Closes ClickUp [868krwawq](https://app.clickup.com/t/868krwawq).

`apps/event-engine` shipped with a `lint` and a `test` script and neither one worked. Both were **present but failing** on `main` — not missing.

> Body updated to describe the final state (6 commits). The per-round detail lives in the comments below.

## What was broken

**`pnpm run lint`** → `eslint src --ext .ts`, **exit 2**, zero files linted:

```
ESLint: 8.57.1
TypeError: Cannot read properties of undefined (reading 'length')
Occurred while linting .../src/handlers/base.ts:1
Rule: "@typescript-eslint/consistent-type-imports"
```

No package-local eslint config, so eslint cascaded to the repo-root `eslint-config-next` and crashed on a typed rule it cannot run here.

**`pnpm run test`** → `jest`, **exit 1**, `Test Suites: 1 failed, Tests: 0 total`. No jest config, no TS transform; `jest` and `@types/jest` were in devDependencies wired to nothing. The one test file was a `node:test` suite, and passed fine under `tsx --test`.

Both failures follow from two facts readable straight out of git, without trusting a log: `git show origin/main:apps/event-engine/package.json` has `"test": "jest"` and `"lint": "eslint src --ext .ts"`, and `git ls-tree origin/main apps/event-engine/` has no `.eslintrc*` and no `jest.config.*`.

## What this does

**Lint**

- `apps/event-engine/.eslintrc.js` — package-local, `root: true`, extending `eslint:recommended` + `plugin:@typescript-eslint/recommended`. Without the `extends`, `no-unused-vars` would be the only error-level rule, and a green CI annotation here would mean far less than the same annotation on a `src/` file.
- **39 errors fixed** so lint exits 0 — 26 under `src/`, 13 under `scripts/`. All `no-unused-vars` bar one `require()` and one `prefer-const`. The 222 `no-explicit-any` warnings are untouched and stay warnings.
- `lint` is now `eslint . --ext .ts` (whole package), matching the CI filter below so the two cannot disagree.

**Tests**

- `apps/event-engine/vitest.config.ts` with `name: 'app:event-engine'` — the name is what the root config's `apps/*/vitest.config.ts` glob and `--project 'app:*'` select.
- `resolve.alias` for `@/`. 19 source files import that way; runtime gets it from `tsconfig-paths/register` and the build from `tsc-alias`, and vitest has neither. Without it the suite could only ever test alias-free modules.
- `include` matches the CI ledger's own detection regex across the whole package, not just `src/` — otherwise a test the ledger counts could run nowhere while the job stayed green.
- Suite converted from `node:test` to vitest and moved to `src/__tests__/`, out of `src/common` — that is a hand-vendored copy of `event-engine-common` which has already diverged, and a re-vendor would delete a test there silently.
- `tsconfig.typecheck.json`, because the build tsconfig excludes `**/*.test.ts` so nothing type-checked the suite (vitest transpiles without checking).

**CI**

- `.github/workflows/lint.yml` filtered changed files to `^(src|packages)/`, so **nothing in CI linted any `apps/` member**. Now `^(src|packages|apps/event-engine)/`. Without this the 39 fixes had no enforcement — the next unused import merges green and the package returns to exit 1 unnoticed, which is this ticket's own failure one level up. The other apps stay out deliberately: no config of their own, so they would hit the same root-config crash.
- Two stale comments corrected (`vitest.config.mts`, `lint.yml`) that said event-engine has no vitest config and is a `node:test` suite.

**Deleted**, on two separate asks: `scripts/sync-submodule.ts` and `scripts/install-git-hooks.ts`. Both assume `src/common` is a git submodule; `.gitmodules` declares only `event-engine-common`, at the repo root. `sync-submodule` would run `git add -A`, `git commit` and `git push origin main` against **`model-share`**, sweeping other people's uncommitted work. `install-git-hooks` has a guard that exits 0 for a non-submodule path and then creates a `.git` directory inside `apps/event-engine`, which would drop the package from git's tracking. Neither is in `package.json` or CI; both are hand-run only.

## The part that matters

Before this, nothing in CI ran event-engine's tests **or** its lint. Now `scripts/ci/assert-workspace-suites-ran.mjs` expects it — it derives its ledger from directories with a vitest config and a test file on disk, so it picked event-engine up with no edit. From the CI log on this PR:

```
✓ app:event-engine  src/common/services/signals.test.ts (4 tests)
apps/ dirs with a vitest config and tests on disk : 6
apps/ dirs that appear in the run                 : 6
       ok  event-engine                 1 file(s), 4 executed
OK: every apps/ suite with tests on disk executed.
```

Six where it was five. If this suite ever stops executing, that job now fails.

## Verification

| Check | Result |
|---|---|
| `pnpm --filter @civitai/event-engine lint` | **exit 0** — 222 problems, 0 errors |
| `pnpm --filter @civitai/event-engine test` | **exit 0** — `Test Files 1 passed (1)`, `Tests 4 passed (4)` |
| `pnpm --filter @civitai/event-engine typecheck` | **exit 0** |
| `pnpm --filter @civitai/event-engine build` | **exit 0**, 0 test files emitted to `dist/` |
| `pnpm run typecheck` (root) | **exit 0** |
| `pnpm run test:apps:run` | **exit 0** — 58 files, 543 tests, ledger 6/6 |
| unit suite (`--project 'unit*'`) | **exit 0** — `1113 passed \| 1 skipped (1114)` files, `17413 passed \| 23 skipped (17436)` tests; `blocks.router.workflow.test.ts` collected **350** |
| CI on `77ddf2220e` | all six checks green |

Positive controls, because a green check that cannot fail is worth nothing. Each was reverted after.

- **Lint** — planted `const __positiveControlUnused = 1;`: exit 1, named it.
- **Lint, per file** — after narrowing the two WIP-stub disables from file-scoped to line-scoped, planted a dead `const` in each past the disabled lines: exit 1, named both.
- **Tests** — disabled the retry branch in `signals.ts`: two failures with a legible assertion, not a hang — `AssertionError: should attempt exactly twice, then give up: expected 1 to be 2`.
- **Alias** — probe importing `@/utils/logger`: `Cannot find package` exit 1 before, `1 passed` exit 0 after.
- **Typecheck** — planted `const x: number = "s"` in the test file: `TS2322`, which the old config could not have seen.
- **Vitest glob** — probe test in `scripts/`: 1 file/4 tests → 2 files/5 tests.
- **CI filter** — `apps/event-engine/…`, `src/…` and `packages/…` match; `apps/moderator/…` and `apps/notifications/…` do not.

Also simulated the eslint job's own classification against `origin/main...HEAD`: exactly one file lands in the **blocking** added bucket (`vitest.config.ts`), and it passes both `eslint` and `prettier --list-different`.

## Known gaps, stated rather than absorbed

- **`scripts/` is typechecked by nothing** — `tsconfig.json` includes only `src/**`. It holds **8 pre-existing type errors** invisible to every command in the repo. The four scripts edited here were checked against a throwaway config including `scripts/`: 8 errors before, the same 8 after, none in the files touched. Fixing them is a separate decision.
- **`src/common` is a vendored copy that has diverged** from `event-engine-common` (`metrics.ts` is 17.5K here, 23.7K there). Seven lint fixes land in it. Reconciling the two is not attempted here.
- Root `pnpm run lint` is red on `main` already (368 errors) and is unchanged by this PR — its script is `eslint src/`, and this branch touches no file under `src/`.
- Two touched files fail `prettier --check`, and fail it identically on `main`. The package was never Prettier-managed, so reformatting them would bury the change.


JustMaier:
Second commit (`1f6ee8094b`) applies the review findings. New SHA — anything read against `def2e3c3fa` is stale.

**`@/` alias added to `vitest.config.ts`.** This was the real one. Runtime resolves `@/` via `tsconfig-paths/register` and the build via `tsc-alias`; vitest had neither, so 19 of the source files were untestable and the failure would have looked like a broken test rather than missing config. The existing suite passed only because `signals.test.ts` imports relatively and `signals.ts` has no imports of its own.

Proven in both directions with a throwaway probe importing `@/utils/logger`, deleted immediately after each run:

```
before:  Error: Cannot find package '@/utils/logger' imported from .../__alias_probe.test.ts   EXIT=1
after:   Test Files 1 passed (1) / Tests 1 passed (1)                                          EXIT=0
```

Matched as `/^@\//` with the trailing slash so it cannot swallow `@civitai/*`.

**WIP-stub disables narrowed from file-scoped to line-scoped.** `no-unused-vars` is the only error-level rule in the new config, so a file-scoped disable left those two files effectively unlinted. Control: planted a dead `const` in each, past the disabled lines — lint went to exit 1 and named both (`'__probeA' ... never used`, `'__probeB' ... never used`). Reverted.

**Stale comment in `.github/workflows/lint.yml:373`** — a second copy of the "event-engine has no vitest config / is a `node:test` suite" claim that the first commit fixed only in `vitest.config.mts`. Grepped the workflows and root config for further copies; there are none.

Re-verified after the change: ee lint exit 0 (0 errors, 202 warnings), ee test 4/4, ee typecheck exit 0, `--project 'app:*'` exit 0 with the ledger still at 6/6 and `ok event-engine 1 file(s), 4 executed`.

**Not fixed, flagged instead:** `apps/event-engine/tsconfig.json` excludes `**/*.test.ts`, so `typecheck` does not cover `signals.test.ts`. That predates this PR, but it matters more now the file is a first-class vitest suite. Fixing it properly means splitting the build tsconfig from a typecheck one so `tsc && tsc-alias` does not start emitting tests into `dist/` — more than this ticket asked for, so I have left it and am naming it rather than quietly absorbing it.

JustMaier:
Third commit (`86ac6561f7`) — second review round. New SHA again; anything read against `1f6ee8094b` is stale.

**The lint had no enforcement.** `.github/workflows/lint.yml:91` filters changed files to `^(src|packages)/`, so nothing in CI ever lints `apps/event-engine`. The 26 fixes in the first commit would have decayed silently — the next unused import merges green and the package goes back to exit 1 with nothing reporting it. That is the same failure this PR exists to fix, one level up, so I have extended the filter to `^(src|packages|apps/event-engine)/`.

The other `apps/` members deliberately stay out: they have no eslint config, so they would fall through to the root Next config and hit the same `consistent-type-imports` crash that started this ticket. Verified that a root-invoked `pnpm exec eslint apps/event-engine/src/...` resolves the package's own `root: true` config, which is what makes this work from the job's cwd. Grep control:

```
in : apps/event-engine/src/handlers/base.ts, src/server/a.ts, packages/civitai-db/b.ts
out: apps/moderator/src/x.ts, apps/notifications/src/c.ts
```

I also simulated the job's own classification against `origin/main...HEAD`. Exactly one file lands in the **blocking** added bucket — `apps/event-engine/vitest.config.ts` — and it passes both gates (`eslint` exit 0, `prettier --list-different` exit 0). The moved test is detected as a rename and lands in the report-only modified bucket.

**Test moved out of `src/common`.** That directory is a vendored copy which `scripts/sync-submodule.ts` re-syncs from `event-engine-common`, which has no tests. A sync would have deleted the suite — and because the ledger derives its expectations from what is on disk, event-engine would have dropped out of the apps job **silently** rather than turning red. Now at `src/__tests__/signals.test.ts`, importing via the `@/` alias added in the previous commit. Ledger re-run after the move: still 6/6, `ok event-engine 1 file(s), 4 executed`.

**Two comments where I had written the wrong reason down**, which is worse than no comment:

- `meilisearch-interface.ts` — I claimed `T` is "part of the public signature". It checks nothing. Corrected to say exactly that, plus what was tried: `Promise<{ hits: T[] }>` makes it real but breaks the interface (meilisearch's own `Index` stops satisfying it — `Hit<RecordAny>` is not assignable to `T`), and dropping `T` is `TS2558: Expected 0 type arguments` at `images.feed.ts:707`. Both verified by running them, not assumed.
- `image-scanned.ts` — the load-bearing fact is not "the TODOs are unwritten", it is that the handler is **absent from the registry** in `handlers/index.ts` and nothing publishes `orchestrator.imageScanned` (its only producer is the commented-out body of `outbox/image-scan.ts`). So the destructure cannot execute. The comment now says that.

**`formatMessage`'s rest parameter deleted** rather than `_`-renamed — no call site passes it and the body ignored it, so the rename was cementing dead API surface that read as if the args were formatted.

Re-verified: ee typecheck exit 0, ee lint exit 0 (0 errors, 201 warnings), ee test 4/4, root typecheck 0 errors, `--project 'app:*'` exit 0, ledger 6/6.

JustMaier:
Fifth commit (`b57d78d26c`) — two dead scripts deleted, both on Justin's call.

Found while chasing a review finding about `src/common`. That directory was a git submodule once; it is not now (`.gitmodules` declares only `event-engine-common`, at the repo root), and both scripts still assume it is. Neither is in `package.json` `scripts` or in CI, so each only runs if someone types it by hand — which the filenames invite.

**`scripts/sync-submodule.ts`** — `chdir`s into `src/common`, then `git add -A`, `git commit`, `git push origin main`. With that directory no longer its own repo, all three resolve to `model-share`: it stages every uncommitted file in the tree, other people's in-flight work included, and pushes it.

**`scripts/install-git-hooks.ts`** — guards on `git submodule status src/common`, which **exits 0** for a path that is not a submodule (verified by running it), so the guard never fires. It then `mkdirSync`s `apps/event-engine/.git/hooks` recursively, creating a `.git` directory inside the package — which makes git treat `apps/event-engine` as its own repository root and drop it from `model-share`'s tracking. Read from source, not executed; running it is the damage. The hook it writes is inert regardless, gated on a `.gitmodules` grep that no longer matches.

Deleted rather than repaired because neither does the job its name implies — `sync-submodule` pushes *out* of `src/common` and never syncs content *in*, so it is not the tool someone reaching for it wants either. Grepped the repo for references to both: none.

Neither is touched by anything else in this PR; they are here because this is the PR that found them.

PR #4035, opened 2026-08-17
fix(env): read NEXT_PUBLIC_LOG_TRPC from its own variable, drop dead public env vars

JustMaier:
ClickUp [868kk4tu7](https://app.clickup.com/t/868kk4tu7) — "`NEXT_PUBLIC_*` env vars with `.default('')` validate clean when unset, then silently vanish."

## The audit came back smaller than the ticket

`src/env/client-schema.ts` is the only place a `NEXT_PUBLIC_*` var gets a zod default, and exactly **two** carry `.default('')`:

| Var | Consumers | Outcome |
|---|---|---|
| `NEXT_PUBLIC_IMAGE_LOCATION` | 6 real call sites | **left alone**, deliberately |
| `NEXT_PUBLIC_CONTENT_DECTECTION_LOCATION` | **zero** | deleted |

Every other `.default(...)` carries a **non-empty** value (`'0.1'`, `'1.0'`, `'0.05'`, `'8'`, `'Stripe'`, `true`/`false`). Those are real defaults, not the silent-vanish shape, and nothing here touches them.

## Why `NEXT_PUBLIC_IMAGE_LOCATION` keeps its default

Justin's call, and the reasoning is worth recording. Requiring it — even `isProd ? z.url() : …` — risks failing a build that legitimately does not pass it. The Dockerfile sets `SKIP_ENV_VALIDATION=1` for `next build` and **nothing in the repo reads that variable**; `src/env/client.ts` `safeParse`s unconditionally. So a required public var is a hard build failure wherever it is unset, which is what the existing comment beside the Faro vars already warns about.

The dangerous paths are covered downstream: `orchestrator.service.ts` refuses to hash a relative media URL, `collection-ai-review.ts` errors on a non-absolute image URL, and `saveImageDownload.ts` falls back to a static allowlist. Stated plainly so nobody over-reads it — **three of the six call sites are guarded**; `getEdgeUrl` generally, `Model3DViewer` and `AdUnitFactory` would still render a hostless path. That is a visibly broken image, not a money or safety path.

## What this PR does change

1. **`NEXT_PUBLIC_LOG_TRPC` read `process.env.NEXT_PUBLIC_LOG_TRP`** — missing the `C`. `clientEnv` restates every key by hand because Next inlines only the references it can see literally, so the key was permanently `undefined` and `z.stringbool().default(false)` turned that into a plausible `false`. It is consumed in `src/utils/trpc.ts` and **could not be switched on**. Dev-only logging, so the blast radius is small; the failure mode is the point.
2. **Deleted `NEXT_PUBLIC_CONTENT_DECTECTION_LOCATION`** from the schema, `clientEnv`, the Dockerfile `ARG` and `.env-example`. Grepped `src`, `apps`, `packages`, `scripts` and `.github` — no consumer. (The `DECTECTION` typo is itself a hint nobody ever wired it up.)
3. **Deleted `ARG NEXT_PUBLIC_MAINTENANCE_MODE`** from the Dockerfile. No such var exists in either schema; only server-side `MAINTENANCE_MODE`.
4. **Deleted `NEXT_PUBLIC_UI_CATEGORY_VIEWS` and `NEXT_PUBLIC_ADS`** from `.env-example`. Neither is in the schema and neither has a consumer anywhere in the tree.

Passing an undeclared `--build-arg` is a Docker warning, not an error, so removing the two ARGs cannot break a pipeline that still passes them.

## The test, and its controls

`src/env/__tests__/client-schema.test.ts` stamps `sentinel::<KEY>` into `process.env` for every key in the schema, re-imports, and asserts each key read **its own** name. `NEXT_PUBLIC_DEFAULT_PAYMENT_PROVIDER` is exempt because it resolves from a ternary. Two more assertions guard schema/`clientEnv` drift and the `NEXT_PUBLIC_` prefix.

Both controls were run rather than assumed:

- Reverting the typo → `AssertionError: expected { NEXT_PUBLIC_LOG_TRPC: undefined } to deeply equal { … }`, exit 1. It names the variable.
- Adding a non-public key to the schema → all three assertions fail, each naming `CONTROL_PROBE_ONLY`.

The key count is asserted too, so the loop cannot pass by iterating nothing.

## Verification

- `pnpm run typecheck` — **0 errors** in 302s, unfiltered.
- `pnpm run lint` — repo-wide it reports 368 pre-existing errors; **zero findings in either file this PR touches**.
- `pnpm run test:unit:run` — **17416 passed, 1114 files, exit 0**. `blocks.router.workflow.test.ts` collected **350** tests, so the submodule is initialised and the run means something.
- An earlier full run exited 1 on `[vitest-pool]: Worker forks emitted error` with **17404 passed and zero failures**; another agent was running a suite on the same box. It did not reproduce.

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -69,11 +69,8 @@ S3_IMAGE_UPLOAD_OVERRIDE=
 
 # Client env vars
 NEXT_PUBLIC_IMAGE_LOCATION=http://localhost:3000
-NEXT_PUBLIC_CONTENT_DECTECTION_LOCATION=https://publicstore.civitai.com/content_detection/model.json
 NEXT_PUBLIC_CIVITAI_LINK=http://localhost:3000
-NEXT_PUBLIC_UI_CATEGORY_VIEWS=false
 NEXT_PUBLIC_UI_HOMEPAGE_IMAGES=false
-NEXT_PUBLIC_ADS=true
 
 # Clickhouse
 CLICKHOUSE_HOST=http://localhost:18123
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,6 @@ RUN --mount=type=cache,id=pnpm-store,target=/root/.local/share/pnpm/store \
 # inherited too, so they don't need re-running here.
 FROM deps AS builder
 ARG NEXT_PUBLIC_IMAGE_LOCATION
-ARG NEXT_PUBLIC_CONTENT_DECTECTION_LOCATION
-ARG NEXT_PUBLIC_MAINTENANCE_MODE
 WORKDIR /app
 
 # Overlay the full source. node_modules is dockerignored, so this never clobbers the node_modules inherited
--- a/src/env/__tests__/client-schema.test.ts
+++ b/src/env/__tests__/client-schema.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { describe, it, expect, afterEach, vi } from 'vitest';
 
 /**
  * `clientEnv` restates every key by hand because Next.js inlines only the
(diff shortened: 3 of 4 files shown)


JustMaier:
Pushed `430f967794` — two review rounds applied since the SHA the first review saw.

**Round 1 (`1904924`)** — three findings, each with the control run rather than assumed:

- The sentinel sweep exempted `NEXT_PUBLIC_DEFAULT_PAYMENT_PROVIDER`, the one key resolved from an expression and so the likeliest place for the typo the guard exists to catch. Planting it (dropping the trailing `R`) pinned every user to Stripe with the suite green. Replaced the blanket skip with value assertions plus a check that nothing is exempted which is not in the schema.
- Making the `NEXT_PUBLIC_LOG_TRPC` read live would have started throwing out of `~/env/client` for any environment already carrying an unparseable value.
- `NEXT_PUBLIC_BASE_URL`'s `NEXTAUTH_URL` fallback is never inlined client-side. **Left in place** — removing it changes behaviour beyond this ticket — but pinned by tests so it is visible. Worth a follow-up decision, not this PR's.

**Round 2 (`430f967`)** — no correctness bugs. Took one low finding: `.catch(false)` swallowed every unparseable value, which is the same plausible-`false` failure this branch exists to remove, relocated from the variable name to its value, and made this the only `stringbool` in the file that hides a misconfiguration. Now only the **empty** case reads as unset — the realistic stale-config shape, and the one that would otherwise throw — while anything else still fails loudly. Both directions are controlled: removing the preprocess fails the empty-value test, restoring `.catch` fails the rejection test.

**Not covered, stated so the guard is not over-read:** these tests prove name symmetry between `clientSchema` and `clientEnv`. They cannot see build-time availability, so a future `NEXT_PUBLIC_FOO` added correctly to both objects can still reach the browser as `undefined` if the image build does not supply it.

**One pre-existing failure found and proved unrelated.** A full suite run hit `src/server/services/__tests__/home-block-fetch-sizing.test.ts > fetches exactly the configured limit` timing out at 60s — the file's **first** test, which pays the cold `import('~/server/services/home-block.service')`; the other six reuse the cached module and pass. Swapping `origin/main`'s `client-schema.ts` back in reproduces it identically, so it is not this branch. It passed on two earlier runs of the same branch and failed on a loaded box, which puts it in flake territory rather than a hard break. Filed here for visibility; happy to raise it separately.

PR #4037, opened 2026-08-17
fix(collections): hide the contest base-model picker where it cannot do anything

JustMaier:
Closes half of `868khbrae`.

### What was wrong

`CollectionEditModal` renders **"Allowed base models"** for any contest-mode collection, moderator-only. The gate it drives — `validateContestCollectionEntry` in `collection.service.ts` — runs **only against `modelIds`**. Image, article and post submissions are never checked against it.

So on an Image contest the field was a control that accepted input and did nothing.

### What changed

Three lines: the picker is now gated on the collection being a Model collection, derived the same way the form itself normalizes a null type (`collection?.type ?? CollectionType.Model`).

### Blast radius

None to existing contests. Checked on the prod replica:

```
metadata ? 'baseModels'                 0
metadata ? 'submissionStartDate'      104     <- control: the query does find metadata keys
mode = 'Contest'                      835
```

The field has never been used in production, so nothing is stranded behind the new condition.

### What the ticket asked for that this does NOT do

The ticket also proposed filtering the 97-entry list itself by media type. Justin's call was to skip that: for a Model contest, picking an audio or 3D base model is a legitimate configuration and the gate works as intended, and a contest has no "intended media type" field to filter against. The ticket's second item (the `WanVideo` chip in `AppSettingsModal`) was also explicitly deferred by Justin and is not in this PR.

### Testing

Typecheck clean (0 errors). Lint reports nothing on the changed file. **No automated test** — this is a JSX render gate on a modal with no existing component-test harness, and standing one up (trpc, `useCollection`, `dialogStore`, `env`) is disproportionate to a three-line condition. Flagging that rather than claiming coverage I did not add.

### DIFF
--- a/src/components/Collections/Collection.tsx
+++ b/src/components/Collections/Collection.tsx
@@ -579,11 +579,17 @@ export function Collection({
   const showSubmissionsClosedNotice =
     !!permissions?.collaborationDisabled && (submissionsClosed || !!permissions?.isOwner);
 
+  // validateContestCollectionEntry applies the base-model list to model entries only, so on any
+  // other collection type a stored value advertises a restriction nothing enforces.
+  const showAllowedBaseModels =
+    !!metadata.baseModels?.length &&
+    (collectionType ?? CollectionType.Model) === CollectionType.Model;
+
   const submissionPeriod =
     metadata.submissionStartDate ||
     metadata.submissionEndDate ||
     metadata.maxItemsPerUser ||
-    metadata.baseModels?.length ? (
+    showAllowedBaseModels ? (
       <Popover
         zIndex={200}
         position="bottom-end"
@@ -614,8 +620,8 @@ export function Collection({
               <Text size="sm">Max items per user: {metadata.maxItemsPerUser}</Text>
             )}
 
-            {!!metadata.baseModels?.length && (
-              <Text size="sm">Allowed base models: {metadata.baseModels.join(', ')}</Text>
+            {showAllowedBaseModels && (
+              <Text size="sm">Allowed base models: {metadata.baseModels?.join(', ')}</Text>
             )}
           </Stack>
         </Popover.Dropdown>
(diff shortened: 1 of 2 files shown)


JustMaier:
Three adversarial review rounds ran over this diff. Recording what they changed, including the round that overturned the previous round's fix.

**Round 1** found the display side was not gated to match: the contest info popover printed `Allowed base models: …` for any collection type, so a non-Model contest holding a stored value would advertise a restriction the server never enforces for it — and with the editor hidden, no moderator could clear the line. Fixed in `055661e845`.

**Round 2** found that hiding the input does not clear the stored value: `shouldUnregister: false` plus the `form.reset({...collection})` means a non-Model contest carrying `metadata.baseModels` re-submits it on every save, with no UI left to remove it. I "fixed" that in `d0472f16f1` by stripping `baseModels` on save for non-Model types.

**Round 3 found that fix was worse than what it fixed, and it was right.** The strip destroys a stored value on the *first save by anyone* — including a co-manager who is not a moderator, has never seen the moderator-only contest block, and is editing the description. Silent, no undo, and because round 1's change also hides the popover line, unobservable afterwards. So the hardening pass turned "an invisible inert value persists" into "an invisible value is silently deleted by an unrelated edit". Reverted in `9fdbc9dd74`.

`9fdbc9dd74` is byte-identical to `055661e845`, which is the state rounds 2 and 3 both read.

**So the shipped diff is two render gates and nothing else** — matching the scope that was actually asked for. The stale-value case that `d0472f16f1` chased is real but latent: prod carries **zero** `Collection` rows with a `metadata.baseModels` key today (835 contests, 104 with a `submissionStartDate` as a control that the query looks where it claims to). If such a row is ever created, the right fix is a deliberate one-off cleanup, not a destructive default on a save path used by people who cannot see the field.

**One thing all three rounds independently checked, and it is the part most likely to have been wrong:** the `?? CollectionType.Model` fallback. `saveItemInCollections` and `bulkSaveItems` accept model entries when `type === Model || type === null`, so a null-typed contest genuinely can hold models and the picker must stay visible there. A strict `=== Model` would have hidden a working control.

**Verification:** typecheck 0 errors locally and `tekton / typecheck` green on CI; lint reports nothing on either changed file; local unit suite 17,413 passed with `blocks.router.workflow.test.ts` collecting 350 tests. One local re-run hit a 60s timeout in `home-block-fetch-sizing.test.ts` — an unrelated server suite that passes alone (7/7) and passed in the other full run; reading it as box contention, but noting it rather than reporting a clean sweep.

PR #4044, opened 2026-08-17
fix(browsing): disable the browsing levels the green domain cannot serve

JustMaier:
## What

Where `canViewNsfw` is false the browsing-level picker is **wholly inert**, not merely capped. `BrowsingLevelProvider:60` sets `domainForcedLevel = sfwBrowsingLevelsFlag` for a logged-in viewer there, and `useBrowsingLevelDebounced` resolves `forcedBrowsingLevel ?? browsingLevelOverride ?? userBrowsingLevel` — so the stored level never reaches a query. Unticking PG-13 changes nothing either, and `blurLevels` cannot compensate because it only ever carries R/X/XXX bits.

Measured against production, same account and query, only the host differing:

| host | `browsingLevel=4` (R only) | `browsingLevel=31` |
|---|---|---|
| SFW | 100 items, `{PG:16, PG13:84}` | 100 items, `{PG:16, PG13:84}` — **identical** |
| mature | 100 items, `{R:98, PG13:1, unrated:1}` | 100 items, `{PG:2, PG13:69, R:16, X:2, XXX:11}` |

The desktop header (`AppHeader.tsx:109`), the account page (`account.tsx:54`) and onboarding (`onboarding.utils.ts:18`) already hide this on such domains. The mobile user menu (`UserMenu.tsx:187`) did not, so a viewer could tick R, X and XXX, watch the chips stay ticked, and get nothing.

Two files, +183/−13.

## The gate is inside `BrowsingModeMenu`, and two neighbours stay outside it

Gating the call site — or the whole `Stack` — silently removes controls that still work:

- **"Apply my filters"** (`disableHidden`) is not mature content.
- **"Blur mature content"** (`blurNsfw`) is read straight from the store by `BlurText.tsx:20` and `RenderHtml.tsx:56`'s profanity filter, **neither of which passes through the domain-forced level**. So it still has effects exactly where the picker has none.

With the header, account page and onboarding all gated, this menu is their only entry point on a capped domain. Both are now siblings of the gate rather than children, and the test asserts that.

The region-restriction notice moved out too. `isRestricted` implies `!canViewNsfw` (`nonRestricted` is in the flag's availability list), so inside the gate it had become unreachable — restricted viewers would have lost the picker *and* the sentence explaining why. It now renders for any capped viewer, with the region wording when that is the cause.

## Why the test is shaped this way

🔴 `test:component` **runs in no CI job**, so a browser test would pin this for whoever runs it by hand and nothing on `main`. This is a structural check in the unit project, which CI does run. It walks the AST — gate conditions are gathered by climbing **up** from the `<BrowsingLevelsGrouped />` element, so reassociating the operands does not red a correct file, and the positive check rejects `!features.canViewNsfw` rather than merely looking for the identifier.

🔴 **Limit, stated:** it proves the gate is in the tree and the two controls are outside it. It cannot prove what renders. Cover against reversion, not against a mis-wired component.

**Four controls run, not assumed:**

```
remove the gate   → × renders the level picker only behind a POSITIVE canViewNsfw check
INVERT the gate   → × …the picker is not guarded by an un-negated canViewNsfw
blur inside       → × keeps the blur toggle OUTSIDE that gate
filters inside    → × keeps the hidden-tags toggle OUTSIDE that gate
```

Plus a positive control asserting every landmark the file reasons about is still present, so nothing can pass vacuously.

## Verification

| check | result |
|---|---|
| `pnpm run typecheck` | **0 type errors**, unfiltered |
| `pnpm exec eslint <both files>` | **exit 0**, no output |
| `pnpm run test:unit:run` | **1119 files passed, 1 skipped (1120)** — 17,523 tests, **0 failures**, no `ELIFECYCLE` |
| `blocks.router.workflow.test.ts` | **350 tests collected** — nonzero, so the submodule is initialised |
| `browsingModeMenuGate.test.ts` | 4 tests, passing in the full run |

Suite log written to a per-session scratchpad rather than the shared `/tmp` on this box; line 3 names `C:\Dev\Repos\work\wt-deprecate-rewards`; 548 KB with a parenthesised total of 1120. The local typecheck was given a positive control — a deliberate `TS2322` in a touched file, caught, then reverted.

## Approaches tried and reverted, so nobody re-derives them

Six adversarial review rounds. Each of these looked right and was not:

1. **Gate `UserMenu`'s call site** — takes "Apply my filters" with it.
2. **Disable only the mature chips, keep the picker** — built on "PG vs PG-13 is a live choice there", which the forced level makes false.
3. **Render an unavailable level unchecked** — desynchronises the picker from the store, and `blurLevels = Flags.diff(nsfwBrowsingLevelsFlag, userBrowsingLevel)` would then quietly drop R from the blur set.
4. **Key on `isGreen`** — not the server's rule; drifts from the clamp in both directions on hosts where the two disagree.

## Scope

- `pages/moderator/images.tsx:989` renders the same picker with no props and is untouched — Justin: *"Leave since it's for mods."*
- **Anonymous viewers** are clamped to PG only on every domain, so the PG-13 chip is dead for them too. Same class, not fixed here.
- **Branch history was rewritten** (squash + force-push with lease) before review: two early commits described the reporting user's content in a way `CLAUDE.md`'s Security section forbids in this public repo.

## The ticket this came from is still open

"PG-13 images won't show unless X or XXX is enabled" **did not reproduce** — probed `image.getInfinite` at browsingLevel 3/7/15/31 on both domains, 200-item pages, no such image, with a positive control proving the probe responds to the parameter. Index levels matched Postgres row-for-row for that user's 200 most recent images. This PR fixes a UI honesty bug found next to it and does not explain the report; the ticket waits on a specific image id from the reporter.

### DIFF
--- a/src/components/BrowsingMode/BrowsingMode.tsx
+++ b/src/components/BrowsingMode/BrowsingMode.tsx
@@ -64,13 +64,13 @@ export function BrowsingModeMenu({ closeMenu }: { closeMenu?: () => void }) {
   return (
     <div id="browsing-mode">
       <Stack gap="md" className="sm:min-w-96">
-        {showNsfw && (
+        {showNsfw && features.canViewNsfw && (
           <Stack gap="lg">
             <Stack gap={4}>
               <Stack gap={0}>
                 <Group align="flex-start">
                   <Text style={{ lineHeight: 1 }}>Browsing Level</Text>
-                  {showNsfw && features.newOrderGame && (
+                  {features.newOrderGame && (
                     <Tooltip label="Help us improve by playing!" withArrow>
                       <Button
                         onClick={closeMenu}
@@ -92,11 +92,6 @@ export function BrowsingModeMenu({ closeMenu }: { closeMenu?: () => void }) {
                 <Text c="dimmed">Select the levels of content you want to see</Text>
               </Stack>
               <BrowsingLevelsGrouped />
-              {isRestricted && (
-                <Text c="red" size="xs" inline>
-                  Your content levels are limited by restrictions in your region
-                </Text>
-              )}
               {hasMatureLevelSelected && (
                 <Group gap="sm" mt={4}>
                   <IconAlertTriangle size={16} />
@@ -107,13 +102,31 @@ export function BrowsingModeMenu({ closeMenu }: { closeMenu?: () => void }) {
                 </Group>
               )}
             </Stack>
-            <Checkbox
-              checked={blurNsfw}
-              onChange={toggleBlurNsfw}
-              label="Blur mature content (R+)"
-              size="md"
-            />
           </Stack>
+        )}
+
+        {/*
+          Outside the gate on purpose. `blurNsfw` is read straight from the store
+          by `BlurText` and `RenderHtml`'s profanity filter, neither of which goes
+          through the domain-forced level — so it still does something where the
+          level picker does not, and this menu is its only reachable toggle once
+          the account page and onboarding are gated.
+        */}
+        {showNsfw && (
+          <Checkbox
+            checked={blurNsfw}
+            onChange={toggleBlurNsfw}
+            label="Blur mature content (R+)"
+            size="md"
+          />
+        )}
+
+        {showNsfw && !features.canViewNsfw && (
+          <Text c="red" size="xs" inline>
+            {isRestricted
+              ? 'Your content levels are limited by restrictions in your region'
+              : 'Mature content levels are not available on this site'}
+          </Text>
         )}
 
         <Group justify="space-between">
--- a/src/components/BrowsingMode/__tests__/browsingModeMenuGate.test.ts
+++ b/src/components/BrowsingMode/__tests__/browsingModeMenuGate.test.ts
@@ -1,0 +1,167 @@
+import fs from 'fs';
+import path from 'path';
+import ts from 'typescript';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * 🔒 The browsing-level control in `BrowsingModeMenu` must not render where the
+ * server will not honour it, and the controls beside it that DO still work must
+ * stay outside that gate.
+ *
+ * WHY, and the part that took several review rounds to get right: where
+ * `canViewNsfw` is false the picker is **wholly inert**, not merely capped.
+ * `BrowsingLevelProvider` sets `domainForcedLevel = sfwBrowsingLevelsFlag` for a
+ * logged-in viewer there, and `useBrowsingLevelDebounced` resolves
+ * `forcedBrowsingLevel ?? browsingLevelOverride ?? userBrowsingLevel` — so the
+ * stored level never reaches a query. Unticking PG-13 changes nothing either, and
+ * `blurLevels` cannot compensate because it only ever carries R/X/XXX bits.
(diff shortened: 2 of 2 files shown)


JustMaier:
## Round 7 — one control that could not fail, fixed; three consequences stated

The review found that my own guard could not catch the regression it names. `insidePickerGate` asked *"is this node inside the picker's subtree?"*, but rewriting a control as `{showNsfw && features.canViewNsfw && <Checkbox …/>}` leaves it just as unreachable on a capped domain while sitting **outside** that subtree. All four earlier mutation controls happened to move the node *into* the subtree — the one shape containment catches.

Replaced with an upward walk asking whether `canViewNsfw` guards the element **wherever it sits**. The decisive control, which the previous version passed:

```
blur given its OWN canViewNsfw gate, still outside the picker subtree
  → × renders the blur toggle WITHOUT a canViewNsfw gate of its own
```

The containment mutation is still caught, so the new form is strictly stronger. The positive control was also weak — `mentions(sf, 'toggleBlurNsfw')` was satisfied by the `const` declaration and `'BrowsingLevelsGrouped'` by the import line, so a file with every JSX element deleted would have passed. It now asserts the rendered elements.

### Three consequences, stated rather than fixed

**1. The "Join the Knights Order" button is inside the gate.** It sits in the Browsing Level header, so it goes with that block. `newOrderGame` is `['blue','red','public']` — no `green` — so green viewers never saw it; the only affected cohort is **region-restricted viewers on a mature domain**, who keep `newOrderGame` but lose `canViewNsfw`. Treated as coherent rather than collateral: the button is a call to action for rating content, and it is anchored to a section those viewers no longer have. Hoisting it would move a button for everyone to serve that cohort.

**2. A moderator in a restricted region gets the wrong sentence.** `useIsRegionRestricted` exempts moderators, the server's `nonRestricted` requirement does not — so they hit `canViewNsfw === false` with `isRestricted === false` and are told "not available on this site" when the cause is their region. Wording only; the gate itself is right for them, because the server clamps them too.

**3. Mature bits already stored on a capped domain stay stored, and cannot be cleared from the UI.** Pre-existing — `browsingLevel` is persisted per domain — but this PR removes the last reachable control. Consequence worth knowing: `blurLevels = Flags.diff(nsfwBrowsingLevelsFlag, userBrowsingLevel)` omits R while the bit is set, so an R-rated image that *does* reach such a viewer (own content, or a level raised after upload) renders unblurred. A one-time clear when `!canViewNsfw` would fix it; that is a state mutation and belongs in its own change with its own review.

### Verification on `8654be1501`

| check | result |
|---|---|
| `pnpm run typecheck` | 0 type errors, unfiltered |
| `pnpm exec eslint <both files>` | exit 0 |
| `pnpm run test:unit:run` | **1119 passed, 1 skipped (1120)** — 17,523 tests, 0 failures, no `ELIFECYCLE` |
| `blocks.router.workflow.test.ts` | 350 collected |
| `browsingModeMenuGate.test.ts` | 4 tests |

Log written to a per-session scratchpad, line 3 naming this worktree, 550 KB, parenthesised total 1120.

### Review history

Seven rounds. One was a null result — it read a pre-force-push SHA and reported three files that are not in this PR; discarded, 0 findings acted on. Of the six real rounds, **five found something the previous round's fix had introduced**, which is the honest summary of this PR: the original defect was one line, and most of the work was undoing my own over-corrections.

PR #4047, opened 2026-08-17
feat(stickers): show placed stickers on post pages

JustMaier:
Stickers placed on an image did not appear on the post page. This makes them appear there, with the same reveal toggle a feed card has. **Display only** — no placing, editing or tray from a post page.

## The post page was missing the provider, not just the layer

`src/components/Post/Detail/PostImages.tsx` had no sticker overlay, which is the visible half. The half that explains why this was not a one-line change: it also had no `StickerPlacementBatchProvider`. That provider is mounted in exactly one place, `src/components/Image/Providers/ImagesProvider.tsx:70`, and `useStickerPlacementBatch` deliberately has **no per-card fallback query** (`StickerPlacementBatchProvider.tsx:145-152`) so that a surface which forgets the provider fails visibly rather than reintroducing a request per card. `PostDetail.tsx` does not render `ImagesProvider`. So on a post page the placement data was never fetched at all.

The result was that a creator could configure per-image placement space from the post editor (`PostEditSidebar.tsx`, `PostImageCards/AddedImage.tsx` — the only `Sticker` references under `src/components/Post/`) and the stickers it earned were invisible on the post itself.

**The provider is fed every image in the post, not the 20-item `_images` slice.** Its chunk memo keys on the joined id string (`StickerPlacementBatchProvider.tsx:53-61`), so handing it a slice that grows on "Load more" refetches the first twenty alongside the rest — a bug with no error attached to it, only extra queries nobody counts.

**It is mounted unguarded, and that costs nothing when the flag is off.** It reads like an unflagged query for every post viewer; it is not. It self-gates on `!!features.stickers && !!features.stickerPlacement`, produces `[]` chunks, and returns `<>{children}</>` — no query, and no wrapper element in the DOM.

## Geometry: measured, and the first reason I gave for it was wrong

The overlay sizes itself to the media element rather than stretching over the container. **The original rationale in this PR was that `RoutedDialogLink` renders an inline Mantine `Anchor`, so the container carries a line-box descender gap under the image. That was measured on a real post and is false** — see the comment below for the numbers. Tailwind's preflight sets `img { display: block }`, so the anchor being inline never produces a gap, and container and media are the same box.

**The reason that survives, and the one the code now states:** the container's height comes from the image's **stored** dimensions (`PostImages.tsx:98-109` sets `aspectRatio` from `image.width`/`image.height`) while a placement is a fraction of **the file the CDN serves**. Those are two sources of truth, and only the served file is what a fraction means. Measuring declines to depend on them agreeing.

**Declared gap: the video path is unverified either way, and it is the case most likely to diverge.** `EdgeVideo` renders through a wrapper `EdgeMedia` does not use for images. No dev post has both a video and a placement, so neither agreement nor divergence has been observed there. That gap is most of the argument for keeping the measurement rather than using `inset-0`: the agreement is verified for images and unverifiable for video, and dropping it would buy silent misplacement on the surface that could not be tested.

`offsetWithin` is now exported from `CardStickerOverlay.tsx` (a one-line change, no behaviour) rather than copied — 25 lines of duplicated measurement geometry is the kind of copy where the original gets a fix and the copy does not.

The media is found via `node.previousElementSibling` — the link — not the container: a container-wide `querySelector('img, video')` takes the first raster element in document order, which is whatever the control group above the media renders.

Two constants deliberately not re-derived:
- **`artworkWidth` stays at `StickerPlacementOverlay`'s 512 default.** The card's 256 is derived for a 450px card; a post image draws at up to `MAX_POST_IMAGES_WIDTH` (800), where a default-capped sticker wants ~400 device px at DPR 2.
- **`surface="detail"`.** The two surface-dependent behaviours — an inset pending ring, and a card-scale die-cut edge with an animation fallback — both exist because a card crops with `cover`. A post image does not crop. The card overrides both defaults; this does not.

## Reveal toggle

`useStickerRevealStore` is one persisted site-wide flag (`src/store/sticker-reveal.store.ts`), so per-image vs per-post only decides where the control sits, not what it drives. This uses the per-image chip, `StickerPlacementCardBadge`, unchanged: it returns `null` when an image has no stickers, so it is also the only per-image indication of which images in a post carry them.

It sits beside the blur toggle rather than the reactions, where a feed card puts it, because the post page's reaction bar moves between three corners depending on video controls — and because its tooltip is "Show stickers on all images", making it a page-level control that belongs with the persistent ones.

Reveal is staggered on intersection rather than on mount — a post is a scroll, and an overlay armed at mount plays its whole reveal for an image still thousands of pixels below the fold. It carries the `ImageStickerOverlay` fallback with it: no `IntersectionObserver` means reveal immediately, never stay hidden.

## Verification, including what is not covered

`src/components/Post/Detail/PostStickerOverlay.browser.test.tsx`, browser mode so the layout is real rather than a fake's return value. Both tests were checked by mutation, not by observing that they pass:

| mutation | result |
|---|---|
| drop the reveal gate (`batch && revealed` → `batch`) | `expected 1 to be +0` — `1 failed \| 1 passed (2)`, exit 1 |
| search the container rather than the link (`previousElementSibling` → `parentElement`) | `expected 16 to be 400` — the overlay measures the control group's 16px icon. `1 failed \| 1 passed (2)`, exit 1 |
| measure the container instead of the media | `expected 182704 to be 300` — `1 failed \| 1 passed (2)`, exit 1 |
| measure the overlay node's own box | `expected +0 to be 300` — `1 failed \| 1 passed (2)`, exit 1 |

The sizing test carries a control assertion — the container is taller than the media **in the fixture** — because without it the assertion is vacuous whenever the two boxes agree, which is what production does. The gap it asserts exists only because no stylesheet loads in the component project; it is what lets the test tell measuring from not-measuring at all, not a claim about the real page.

**Declared gaps, rather than tests that cannot fail:**
- **The video path** (above): unverified either way, and the case most likely to diverge.
- There is no test for "renders nothing without a batch provider". With no provider the placement list is empty and the empty-list path already returns `null`, so no mutation of this component can make such a test fail. It would name a failure it cannot detect.
- The reveal *animation* (stagger, arming on intersection) is not asserted. Arming is a self-deleting state and the honest options were a race or a widened timeout; the mocked overlay in the test does not draw it either.

Clean run: `Test Files 1 passed (1)`, `Tests 2 passed (2)`, exit 0, log naming `C:/Dev/Repos/work/wt-post-stickers`. `TYPECHECK_HEAP_MB=12288 pnpm run typecheck` → `OK — 0 type errors in 42s`. ESLint over the changed files — 0 errors (warnings are pre-existing unused imports and two fixture ``s that must stay raw). Full unit suite result posted as a comment.

Note on the diff size: `PostImages.tsx` shows as a rewrite because wrapping the list in the provider re-indented the whole JSX body under Prettier. The semantic change there is four lines.

### DIFF
--- a/src/components/Post/Detail/PostImages.tsx
+++ b/src/components/Post/Detail/PostImages.tsx
@@ -1,6 +1,6 @@
 import { ActionIcon, Badge, Button, Center, Group, Loader, Paper, Stack } from '@mantine/core';
 import { IconBrush, IconInfoCircle } from '@tabler/icons-react';
-import { Fragment, useRef, useState } from 'react';
+import { Fragment, useMemo, useRef, useState } from 'react';
 import { AdUnitTop } from '~/components/Ads/AdUnit';
 import HoverActionButton from '~/components/Cards/components/HoverActionButton';
 import { RoutedDialogLink } from '~/components/Dialog/RoutedDialogLink';
@@ -12,6 +12,9 @@ import { ImageMetaPopover2 } from '~/components/Image/Meta/ImageMetaPopover';
 import { ImageGuard2 } from '~/components/ImageGuard/ImageGuard2';
 import { MediaHash } from '~/components/ImageHash/ImageHash';
 import { PostContestCollectionInfoAlert } from '~/components/Post/Detail/PostContestCollectionInfoAlert';
+import { PostStickerOverlay } from '~/components/Post/Detail/PostStickerOverlay';
+import { StickerPlacementBatchProvider } from '~/components/Sticker/StickerPlacementBatchProvider';
+import { StickerPlacementCardBadge } from '~/components/Sticker/StickerPlacementCardBadge';
 import { Reactions } from '~/components/Reaction/Reactions';
 import { useFeatureFlags } from '~/providers/FeatureFlagsProvider';
 import { MAX_POST_IMAGES_WIDTH } from '~/server/common/constants';
@@ -46,6 +49,12 @@ export function PostImages({
   const videoRef = useRef<EdgeVideoRef | null>(null);
   const features = useFeatureFlags();
 
+  // Every image in the post, not the `_images` slice below: the batch provider
+  // keys its chunks on the joined id string, so handing it a slice that grows on
+  // "Load more" refetches the first twenty alongside the rest. A post is well
+  // inside the provider's 100-id chunk, so this is one query either way.
+  const imageIds = useMemo(() => images.map((image) => image.id), [images]);
+
   if (isLoading)
     return (
       <Paper component={Center} p="xl" mih={300} withBorder>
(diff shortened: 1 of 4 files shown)


JustMaier:
## The descender-gap premise was measured on a real post and is false

This PR originally justified measuring the media element by saying the container is a few pixels taller than the artwork, because `RoutedDialogLink` renders an inline Mantine `Anchor` and an inline box reserves line-box room under an image. The body now states the surviving reason instead; this comment records what was actually measured, so nobody re-derives it.

Measured in a dev server on `/posts/30077020` — chosen because it is the only dev post carrying three approved placements, and its image is 1006x1520, i.e. tall, where a vertical error is largest:

```json
{"anchorDisplay":"inline","anchorH":1100,"mediaH":1100,"paperH":1100,"aspectRatio":"1006 / 1520"}
```

Half the premise holds: the anchor really does compute to `display: inline`. The load-bearing half does not — container, anchor and media are the same 1100px box, so the gap is **0**.

I then removed the container's `aspectRatio` in the live page, expecting that to expose the case where the container's height falls back to the line box (an image with no stored dimensions):

```json
{"before":{"paperH":1100,"mediaH":1100},"after":{"paperH":1100,"mediaH":1100},"gap":0}
```

Still zero. **The cause is Tailwind's preflight, which sets `img { display: block }`** — so the anchor being inline never produces a line box with descender room.

### Why the test's control assertion still passes

The component project loads no stylesheet, so the fixture's `img` keeps its inline default and the gap does appear there. That is what makes the sizing assertion able to distinguish measuring from not-measuring at all — a property of the fixture, not a claim about the page. Both the fixture comment and the component's doc comment now say so.

### What this changes

The measurement stays, on a different and narrower justification: the container's height derives from the image's **stored** dimensions while a placement is a fraction of **the file the CDN serves**, and the video path — `EdgeVideo` renders through a wrapper `EdgeMedia` does not use for images — could not be checked at all, because no dev post has both a video and a placement. Verified agreement for images plus unverifiable agreement for video is not a basis for deleting the measurement.

## Full unit suite

```
 Test Files  1122 passed | 1 skipped (1123)
      Tests  17540 passed | 23 skipped (17563)
   Duration  177.36s
```

Exit code 0. `src/server/routers/__tests__/blocks.router.workflow.test.ts` collected **350 tests**, so the `event-engine-common` submodule is present in this worktree and the run is not silently missing a suite. Raw log written to a per-session scratchpad path, not `/tmp`, and its `RUN` line names `C:/Dev/Repos/work/wt-post-stickers`.

PR #4058, opened 2026-08-17
feat(app-store): match the listing detail to the model detail page's design language

ZacxDev:
Ports the App Store **listing detail** (`/apps/store-preview/<slug>`) onto the **model
detail page's** design language, so the two surfaces read as one product.

The layout is a port, not an invention: the grid spans, the `Card.Section withBorder
inheritPadding` rail headers, the `.detailsPanel` / `.detailRow` / `.detailLabel` classes
(imported from `ModelVersionDetails.module.scss`, not re-declared), the `IconBadge` +
`StatHoverCard` stat chips, the `Updated: <date> │ <category>` meta line and the
`Title order={2}` discussion heading all come from
`src/pages/models/[id]/[[...slug]].tsx` + `ModelVersionDetails.tsx`.

## What changed

| | |
|---|---|
| **Structure** | Hero (unchanged) → header block → two-column `ContainerGrid2` (`{base:12, sm:7, md:8}` / `{base:12, sm:5, md:4}` with `order={{sm}}`) → full-width discovery rail, reviews, discussion |
| **Header** | Interactive stat chips (recommendations, installs), a `⋮` overflow menu replacing the stacked button column, and a `Updated: <date> │ <category>` meta line |
| **Right rail** | "Get this app" action card (the primary CTA stays a real button) · `SmartCreatorCard` · a "Details" accordion |
| **Main column** | Screenshots · `About` under `ContentClamp maxHeight={460}` · the off-site disclosure |
| **Page width** | `/apps/store-preview/[slug]` moved from `APPS_READABLE_PAGE_WIDTH` (1100) to a new `APPS_TWO_COLUMN_DETAIL_PAGE_WIDTH` (**1320** = Mantine `xl`, the model page's own container) |

**`getRatingLabel` is now shared.** It was a module-private function in
`ModelVersionReview.tsx`; the listing's "Reviews" row renders the same verdict, so the
ladder moved to `src/utils/rating-label.ts` **verbatim** and both surfaces consume it.
The model page's rendered labels are unchanged by construction.

> ⚠️ **"ONE ladder … for every surface that renders one" was overstated.** A third,
> drifted ladder survives at `src/pages/3d-models/[id]/[[...slug]].tsx:132`
> (`sentimentLabel` — a 0–100 scale, different bands, a sixth label). It predates this PR
> and is **not fixed here**; it is allowlisted by name with the consolidation recorded as
> a follow-up. Separately, the caller-ledger guard that was supposed to prevent a second
> ladder was **walkable** — a drifted copy with no import passed all 49 tests. Repaired in
> `5d12204a68` and re-verified against that exact mutant.

**`getRecommendLabel` is untouched.** It backs the store grid cards, which keep the
percentage. The detail's word label comes from the shared ladder instead; the percentage
**and the not-recommended count** are retained on the detail as the recommend chip's
hover-card message (`86% recommend (499) · 68 don't`). Changing `getRecommendLabel`'s
return shape would have changed every card.

> ⚠️ **Corrected in `5d12204a68` — as first written, both of those were false.** The
> hover message was passed behind an inverted condition, so the percentage was
> **unreachable** on every listing that had one, and `notRecommendedCount` (rendered as
> "N recommend · M don't" before the redesign) was rendered **nowhere**. Both decisions
> now live in a pure view-model, `src/components/Apps/appListingStatChips.ts`, pinned in
> the blocking node `unit` project. Full write-up in
> [this comment](https://github.com/civitai/civitai/pull/4058#issuecomment-5321444321).

## DTO — exactly two fields added to `ListingDetail`

This is a privacy-reviewed public allowlist. Both fields have a public analogue on the
model page, and both are justified in their docstrings on the type:

- **`updatedAt: string`** (ISO-8601) — `app_listings.updated_at`. The direct analogue of
  the model page's public `Updated: <date>` line. It reveals only that an approved,
  publicly-listed app was edited — the same class of fact the store's `newest` sort
  already exposes for `created_at`. No *who*, no *what*. A **string**, not a `Date`,
  because this DTO also crosses the transformer-less public REST path.
- **`installCount: number`** — `AppListingMetric.install_count`. A **deliberate
  disclosure decision**, and stated as one rather than as a formality. The store's
  `popular` sort **is** `install_count DESC`, so the *ordering* of approved listings by
  this number is already derivable — but that is strictly less than the *value*, so this
  does add a fact. It is serialized wholesale to an anon-capable endpoint
  (`src/pages/api/v1/apps/[slug].ts:109`) that backs the public CLI, so it is readable by
  anyone. The decision stands on its merits: an aggregate over the whole install base, it
  identifies no user, and it is directly analogous to public model download counts.
  **Open:** whether the published developer docs / OpenAPI for that endpoint need
  updating for the two new fields — that lives in a separate repo.

  *(An earlier version of this bullet said "already publicly **observable**, not merely
  public … adds precision, not a new fact." That was overstated; corrected here.)*

Nothing else was added. The projection's allowlist key-set assertion was updated to name
both.

## Hard invariants, each re-verified

- No raw `<iframe>` returns — the source gate in the blocking `unit` project still passes.
- Nothing links to `/apps/<appBlockId>` (redirect loop). **New** source gate: this file
  may build no `/apps/…` URL by interpolation at all; every app URL comes from a pure
  view-model. Labelled an *invariant guard* — it was green at base.
- The hero stays an `<a href>` with its explicit `aria-label` and `c="inherit"`, records
  **no** recents entry, and is on-site `open` only.
- `PrimaryAction` still resolves its glyph inside each branch.
- **`preview` mode**: every omission is enumerated on the prop's docstring and has its
  own test with a positive control from the live arm — see the mutation table below.

## Two defects this surfaced (both fixed here)

1. **A cast-built `ListingDetail` blanked the moderator review modal.** The moderator
   combined-review surface builds a `ListingDetail`-shaped object *directly*, not via
   `projectListingDetail`. **Precision correction:** the builder itself,
   `buildListingDetailPreview`, has a typed `: ListingDetail` return — it is not a cast.
   The cast in that chain is one step upstream, at
   `(queue.data?.items ?? []) as OffsitePendingRow[]`, so what crosses untyped is the
   *row*. The runtime guard is still correct and still needed; the reason is just further
   back than originally written. The first version of the row builder called
   `detail.installCount.toLocaleString()` unconditionally → threw → a crashing child
   unmounts its ancestors → empty `<body>`. Caught by the pre-existing
   `CombinedReviewModal.browser.test.tsx`. A details row is decoration and must never be
   able to blank the page it decorates, so an absent field now takes the same path as a
   null one: omit the row. (`0` is explicitly discriminated from `undefined`.)
2. **Four sibling Apps browser suites wholesale-mocked `FeatureFlagsProvider`.** Once the
   detail body reached `useChatEnabled` (via `SmartCreatorCard` → `ChatUserButton`), all
   four failed to **import** with `does not provide an export named
   'useFeatureFlagsReady'` — reported as `Tests no tests`, i.e. as nothing to see rather
   than as a failure. Switched to the `importOriginal` spread the repo's own
   `no-wholesale-module-mock` rule prescribes.

## Deviations from the brief, stated

- **No SSR prefetch of `user.getCreator`.** `user.getCreator` **is** a `publicProcedure`
  (verified in `user.router.ts`), so the card is safe for the anon-reachable page. But
  the creator id is not known at SSR time: this page fetches the listing **client-side**
  by deliberate design, and prefetching the creator would mean fetching the listing
  server-side too — replicating the router's scope/maturity gates in
  `getServerSideProps`. The model page's CLS mechanism also does not apply here: the
  whole body is behind a `<Loader/>` and mounts at once, and `CreatorCardSimple` renders
  its default shape (not null) while the query resolves. Flagged rather than
  silently done.
- **No "Add Comment" button beside the discussion heading.** The model page needs one
  because its composer is behind a routed dialog; this thread renders `<CreateComment />`
  inline a few lines below, so a button would be a second affordance for a form already
  on screen. The `Title order={2}` treatment is applied.
- **The category badge is a plain badge, not a link.** The store has no
  category-filtered *route* — `/apps` holds its category filter in client state — so a
  link would 404 or land on an unfiltered grid.
- **Rail order** is action → creator → details (per the brief's diagram). The model page
  puts the creator card *below* Details.

## Out of scope, confirmed byte-unchanged

`AppDetailsModal.tsx`, `AppListingCard.tsx`, `appListingCardView.ts`,
`appListingGrid.ts`, `pages/apps/index.tsx` — all verified unchanged against the merge
base.

## Test matrix

Base ref for every claim below: **`36bf0518a5`**.

**Suites**

| suite | base | this branch |
|---|---|---|
| `unit*` (full) | 1129 files / 17678 passed | 1131 / **17749 passed** (+2 files, +71 tests) |
| `component`, `src/components/Apps` + `src/tests/pages/apps` | 72 files / 896 passed | 72 / **926 passed** |
| `pnpm typecheck` | — | **0 errors** |

**Red at base → green at HEAD** (measured by running the new test files against a
pristine base worktree):

- **27 of the 62** tests in `AppListingDetailBody.browser.test.tsx` — the menu, the stat
  chips, the meta line, both grid columns at both viewports, the rail contents, the
  creator card, every details-row test, the ContentClamp, the a11y accordion test, and
  each of the new `preview` omissions.
- `appsPageWidths.test.ts` — the two-column-width test and the four-class list.
- `rating-label.test.ts` — both **caller-ledger** tests (at base, `ModelVersionReview`
  still owns its own ladder).
- `app-listing.service.test.ts` — the DTO allowlist key-set.
- `appListingDetailRows.test.ts` — the four *absent-field* tests were red on the real
  crash before the fix.

**Green at base, labelled as such** — invariant guards, not regression coverage: the new
`/apps/…`-interpolation source gate, the existing `<iframe>` gate, the hero's four
negative branches, and the rating-ladder characterization table (a verbatim move cannot
have a pre-change failing state).

**Mutation results — the `preview` guards.** Each guard deleted one at a time and
narrowly (the clause alone, never with its enclosing condition), against a green control:

| mutant | verdict | killed by, with that test's own assertion |
|---|---|---|
| back-to-store link | ✅ killed | `preview omits the back-to-store link` |
| header stat chips | ✅ killed | `preview OMITS the header stat chips` — *"…must be omitted in preview: expected 2 to be +0"* |
| `⋮` actions menu | ✅ killed | `preview OMITS the ⋮ actions menu` — *"expected 1 to be +0"* |
| header meta line | ✅ killed | `preview OMITS the header meta line` |
| right-rail action card | ✅ killed | `preview OMITS the right-rail action card` |
| rail creator card | ✅ killed | `preview OMITS the creator card` |
| related-listings rail | ✅ killed | `preview OMITS the related-listings rail` |
| reviews section | ✅ killed | `preview OMITS the reviews list` |
| comments thread | ✅ killed | `preview OMITS the comments thread` |
| clickable hero | ✅ killed | `preview OMITS the clickable hero` |
| read-only creator chip (`preview &&`) | ✅ killed | `preview keeps the read-only creator CHIP` |
| Details **reviews row** (in the row builder) | ✅ killed | `preview omits the REVIEWS row from the details rail` |
| review-modal mount | ⚠️ **survived** | see below |
| report-modal mount | ⚠️ **survived** | see below |

12/14 killed, each by exactly one test, each by that test's own message. **The two
survivors are reported, not hidden**: a Mantine `Modal` with `opened={false}` renders no
DOM and its `getMyReview` query is `enabled: false`, so a mounted-but-closed modal is
unobservable from the rendered output and issues no request — removing the clause is
inert *today*. The clause stays (it is what stops a future mount effect firing against an
unapproved shadow listing) and the code now says so, so nobody deletes it on the grounds
that nothing goes red.

## What this coverage structurally cannot see

- **The component suite loads no CSS.** `vitest.config.mts` registers only the process
  shim and component-setup — no Mantine sheet, no Tailwind, no SCSS module. **No test in
  it validates a single visual claim.** The layout tests assert *structure* (both columns
  exist as siblings, rail-first in the DOM, rail contents survive at a narrow viewport);
  side-by-side vs stacked, the rail width, `c="inherit"`, and the ContentClamp's height
  cut are verified **only** by the captures.
- **A viewport is a dimension this config does not pin**, so any test that does not set
  one is blind to viewport bugs by construction. The layout tests set it explicitly at
  two named points and `beforeEach` restores the wide one so a narrow test cannot leak.
- The browser suite mocks tRPC, so it exercises no real query and no server-side gate.
- `CreatorCardV2` (the `cosmeticShop` branch of `SmartCreatorCard`) is not covered — the
  mocked flags route to `CreatorCard`.
- **Pre-existing and NOT fixed here** (asserted as-is so it is not mistaken for covered):
  a screenshot gallery whose every tile 404s keeps its "Screenshots" heading. The section
  is hidden from the URL list at render time, while `broken` is per-tile state discovered
  later. Unchanged by this PR; lifting the broken-set into the parent is a behaviour
  change with its own decision.
- The known-flaky `apps-browse-all` strict-mode test did **not** appear in any run here.

## Deploy status

**Not merged, not promoted, not serving, not verified in production.** `civitai-dp-prod`
serves from `release`; a merge to `main` is not live until a main→release cut. Production
behaviour has not been checked and cannot be from this branch — the store surface is
flag-gated and mod-only.

---

## Follow-ups acknowledged, NOT addressed in this PR

- **Double-report guard.** The report menu item stays live after reporting; the server's
  `one_open_per_reporter` constraint dedups and returns a friendly `CONFLICT`, so nothing
  is corrupted — but the user gets an error toast where a confirmation belongs.
  `ReportListingButton` is also left with no production caller (the body imports
  `ReportListingModal` + `useCanReportListing`).
- **Moderator preview shows `submittedAt` labelled "Updated"**
  (`src/components/Apps/reviewListingPreview.ts:133-140`). Its comment claims the field is
  unobserved in that posture; `appListingDetailRows.test.ts:242` asserts the `updated` row
  **does** render in preview, so the comment is wrong.
- **`Installs: 0` renders in preview** via the Details rail, while the header stat chips
  are suppressed in preview specifically to avoid showing zeros there.
- **Ladder consolidation** — the 0–100 `sentimentLabel` on the 3d-models page (above).

## Correction round (`5d12204a68`)

Five audit findings fixed or restated; see
[this comment](https://github.com/civitai/civitai/pull/4058#issuecomment-5321444321) for
the full account. Verification for that round: every new/repaired test watched **red at
`617f09c204`, green at HEAD**; the previous round's 7-mutant battery **re-run in full plus
7 new mutants — 14/14 died, no survivors**, including the previously-*surviving*
copied-ladder mutant; `pnpm typecheck` 0 errors with the instrument validated. The browser
`component` suite **could not be run** (Chromium build mismatch) and CI does not run it
either — every load-bearing guard added in that round is in the node `unit` tier by design.


ZacxDev:
### Captures — how the visual claims were measured

Since the component suite loads no CSS, the layout claims are measured separately, with
a **capture harness that imports the same stylesheets `src/pages/_app.tsx` does**
(`globals.css` + the six Mantine layer sheets) and mounts the real body inside the
route's own 1320px container. The identical harness was run against a pristine worktree
at the merge base and against this branch, at two named widths.

The harness is deliberately **not committed** — it asserts nothing and would run in CI as
an assertion-free suite.

| capture | what it shows |
|---|---|
| `before` @ 1440w | one stacked column: hero → header with a stacked full-width button column on the right → recommend rollup → screenshots → About |
| `after` @ 1440w | two columns: screenshots + About on the left; "Get this app" card, creator card and an open **Details** accordion (Kind / Category / Rating / **Reviews `Very Positive (499)`** in green / Installs / Updated) on the right. Header carries the two stat chips, the `⋮` menu, and `Updated: Mar 3, 2026 │ UTILITY` |
| `after` @ 390w | genuinely stacks to one column, rail **above** content (action card first) — matching the model page's own mobile behaviour — with the `About` clamp showing its **Show More** control |
| model page @ 1440w / 390w | the live reference: same stat-chip row, same `Updated: … │ <category>` meta line, same `⋮`, same right-rail Details accordion with a green `Overwhelmingly Positive (18,348)` Reviews row |

One thing the captures caught that the DOM tests could not: the first attempt truncated
the markdown body and the last two Details rows. That was a **capture artifact** — an
element screenshot taller than the viewport is stitched from scrolled paints — not a
layout bug. Re-shot with a viewport tall enough to contain the page; the named dimension
is the **width**, which is what the responsive layout keys on.

ZacxDev:
## Correction to the PR description — five claims that were wrong or overstated

An adversarial audit found that this PR's description **misstated the change** in five
places. Posting this as a comment rather than only editing the body, because a reviewer
may already have read the wrong version. The body has now been updated to match.

Three of the five were real defects and are **fixed in this round** (`5d12204a68`); two are
claims that needed restating.

---

### 1. "the percentage is retained … as the recommend chip's hover-card message" — was FALSE as implemented. **Fixed.**

The condition was inverted:

```tsx
message={detail.recommend.recommendPct == null ? recommendDetail : undefined}
```

`StatHoverCard` renders `message` **instead of** the numeric value. So whenever a listing
had reviews, `message` was `undefined` and the dropdown fell through to
`value.toLocaleString()` — i.e. it repeated `recommendedCount`, the same number already
printed on the chip face. `"87% recommend (340)"` was **unreachable on every listing that
had a percentage**, and the code comment above it claimed the opposite.

The fix is *not* flipping the condition to `!= null` — that renders a bare `"0"` in the
no-review case. The ternary is dropped entirely, so the message is unconditional: with
reviews it reads the percentage, without reviews it reads "No reviews yet".

### 2. The drop of "N don't recommend" was unmentioned in the "what changed" table. **Restored.**

The pre-redesign body rendered `"N recommend · M don't"`. The redesign dropped that line
and nothing took it over, so `notRecommendedCount` was rendered **nowhere** on the live
surface — it survived only as a `0` in a preview fixture. Combined with (1), the page had
lost **both** the percentage and the negative count.

It is folded back into the recommend chip's hover message, which now reads
`86% recommend (499) · 68 don't`. It stays absent in `preview` posture (it is a review
aggregate, and a shadow listing has none) — that omission is now decided in a pure module
and covered by a test, rather than by a `!preview &&` in the JSX.

### 3. "ONE ladder … for every surface that renders one" — overstated. **Not fixed; recorded as a follow-up.**

A third, drifted ladder survives at `src/pages/3d-models/[id]/[[...slug]].tsx:132`
(`sentimentLabel`): a **0–100** scale with different band edges and a sixth label
("Too few reviews"). It predates this PR. Consolidating it is a behaviour change to a
live page, not a refactor, so it is deliberately out of scope for this round.

It is now **allowlisted by name**, with the follow-up recorded in the test that owns the
rule — not silently tolerated, and not dodged by weakening the check: the scan does flag
it, and an assertion pins that it is still flagged, so when the consolidation lands the
allowlist entry has to come out with it.

The audit also showed the caller-ledger guard that was supposed to prevent a second
ladder **did not work** — a drifted copy of the ladder planted in a new file with no
import passed all 49 tests, which is exactly the scenario that guard's own docstring
named. Two independent faults, both fixed:

- an importer list is structurally blind to a **copy** (not importing is what a copy
  does). There is now a second scan that looks for the ladder's own output alphabet in
  the source text. Reproduced against the auditor's exact mutant: the repaired guard
  **fails**, naming the offending file.
- a bare `text.includes('~/utils/rating-label')` counted a mere textual **mention** as a
  consumer — so any doc reference registered as a caller. It now matches an
  `import` / `require` **form**.

### 4. "`installCount` — already publicly *observable*, not merely public" — overstated. Restated.

The `popular` sort being `install_count DESC` makes the **ordering** derivable. That is
strictly less than the **value**: ordering tells you which app has more installs, not how
many. So this is not a non-event, and describing it as "adds precision, not a new fact"
undersold it.

Restated as what it actually is: a **deliberate disclosure decision**. The field is
serialized wholesale to an anon-capable endpoint (`src/pages/api/v1/apps/[slug].ts:109`)
that backs the public CLI, so the number is readable by anyone. The decision stands on its
merits — it is an aggregate over the whole install base, it identifies no user, and it is
directly analogous to public model download counts — but it is a decision, not a
formality.

**Open question:** whether the published developer docs / OpenAPI description for that
endpoint need updating to describe the new fields. That lives in a separate repo and is
not addressed here.

### 5. "builds a `ListingDetail`-shaped object directly, through a cast" — imprecise. Restated.

`buildListingDetailPreview` has a typed `: ListingDetail` return — it is not itself a
cast. The cast in that chain is upstream of it, at
`(queue.data?.items ?? []) as OffsitePendingRow[]`, so the untyped value is the *row*, not
the detail. The runtime guard in the row builder is still correct and still needed (an
absent field really can reach it), but the reason is one step further back than the
comment said.

---

## Acknowledged follow-ups, NOT addressed in this round

- **Double-report guard regression.** The report menu item stays live after reporting. The
  server's `one_open_per_reporter` partial-unique constraint dedups the duplicate and
  returns a friendly `CONFLICT`, so nothing is corrupted — but the user gets an error
  toast where they should get a confirmation. Related: `ReportListingButton` is now left
  with **no production caller** (the detail body imports `ReportListingModal` +
  `useCanReportListing` instead); only its own test renders it.
- **Moderator preview shows `submittedAt` labelled "Updated"**
  (`src/components/Apps/reviewListingPreview.ts:133-140`). Its comment claims the field is
  unobserved in that posture, while `appListingDetailRows.test.ts:242` asserts the
  `updated` row **does** render in preview. The comment is wrong; a moderator sees a
  submission time labelled as an edit time.
- **`Installs: 0` renders in preview** via the Details rail, while the header stat chips
  are suppressed in preview specifically to avoid showing zeros there. The two surfaces
  disagree about the same posture.

---

## Verification for this round

- Every new and repaired test was watched **red at `617f09c204`** and is **green at HEAD**.
  Restoring the base F1+F2 behaviour under the new tests turns **7 of 18** red; the base
  version of the ladder test stays **49/49 green** against the scale bug it claimed to
  catch — which is the measurement proving that "BEHAVIOURAL" seam test was a tautology.
  It is replaced by one that imports and executes **both** call sites.
- The full 7-mutant battery from the previous round was **re-run**, plus 7 new mutants for
  this round's fixes: **14/14 died, no survivors**, each isolated to the narrowest
  expression that can be wrong, each recorded with the specific assertion that killed it —
  including the previously-**surviving** copied-ladder mutant.
- `pnpm typecheck`: **0 errors**, with the instrument validated (planted error → 1 error at
  the right file and line → removed). The new test files were typechecked separately,
  since `tsconfig.json` excludes `src/**/__tests__/**`.

### What is NOT verified

- **The browser (`component`) suite was not run — it cannot run in this environment**
  (the pinned Chromium build and the Playwright provider disagree on version, so the
  browser fails to launch). **CI does not run that tier either.** Nothing in this round
  depends on it: the load-bearing guards for every fix here were deliberately placed in
  the node `unit` tier, which CI gates and which was executed. The existing browser specs
  for these chips assert `data-listing-stat` attributes and `aria-label`s that this change
  preserves, but that is reasoning, not a measurement.
- No production verification. The store surface is flag-gated and this branch is not
  deployed.

PR #4061, opened 2026-08-17
feat(app-blocks): record client-claimed source provenance on publish requests (#4059)

### LINKED ISSUE #4059: App Blocks: submitVersion records no provenance — accept and return a source commit + dirty flag on publish requests
`appBlocks.submitVersion` accepts a bundle and stores nothing about where it came from. A bundle built from an uncommitted tree is indistinguishable, afterwards, from one built from a tagged release — so deploy-vs-source drift can be observed but not diagnosed.

## Why the client cannot fix this alone

The CLI already **computes** the facts. `civitai app submit` refuses a dirty git work tree (`civitai/cli#415`), so it holds the commit SHA and the dirty flag at submit time. It has nowhere to put them:

- `submitVersionSchema` (`src/server/schema/blocks/publish-request.schema.ts:49-60`) accepts exactly one field, `bundleBase64`. It is a plain `z.object`, so **extra keys are silently stripped** — a client sending provenance today gets no error and no storage. That is the inert-feature shape: it would look like it worked.
- `shapeRow` (`src/pages/api/v1/blocks/submissions.ts:131-150`) returns `id`, `blockId`, `appBlockId`, `version`, `status`, `rejectionReason`, `approvalNotes`, `deployState`, `deployDetail`, `deployUpdatedAt`, `submittedAt`, `reviewedAt`, `updatedAt`, `createdAt`, `liveUrl` — nothing about source.

So the CLI half is deliberately **not** being built until this exists.

## Evidence it matters, from the field

Five first-party apps were found behind their live deployed version (`civitai/cli#411`):

| app | source repo | live |
|---|---|---|
| `custom-generators` | 0.4.0 | **0.5.2** |
| `gen-matrix` | 0.7.3 | **0.8.2** |
| `model-benchmarking` | 0.1.3 | **0.2.2** |
| `playable-collections` | 0.2.0 | **0.2.2** |
| `app-requests` | 0.2.0 | **0.2.2** |

`custom-generators` has **never** held 0.5.2 anywhere in its git history (0.3.3 → 0.3.4 → 0.4.0), and the newer source was in no clone. The `pubreq` timestamps show three apps submitted in **7 seconds** — a scripted batch from a tree nobody kept.

All five were reconciled by hand afterwards (24 files that existed nowhere in git are now in git). That took a session of archaeology which two stored columns would have made a lookup.

## The ask

Two nullable columns on `app_block_publish_requests`, accepted by `submitVersionSchema` and surfaced through `shapeRow`:

- `source_commit` — nullable text, the 40-char SHA the bundle was built from
- `source_dirty` — boolean, whether the work tree had uncommitted changes

**Deliberately NOT branch, remote, or tree-hash.** The CLI already computes commit + dirtiness for its own guard, so the client half is nearly free; every additional field is another migration for information that does not answer *"which commit is this?"*.

**Nullable and advisory on purpose.** Submissions predating this carry `NULL`, and a client that does not send them still submits — this must not become a submit gate. 🔴 Please treat `source_commit` as **untrusted client-supplied text**: validate the shape (40 hex chars) and store it as a claim, not as a verified fact. The server cannot confirm the bundle was actually built from that commit, and nothing downstream should imply otherwise.

## Smallest useful version

If two columns is one too many: `source_commit` alone still solves the diagnosis, since a dirty submit can be represented by the CLI refusing to send one. The dirty flag is the nicer signal but the commit is the load-bearing half.

Client-side tracking issue: `civitai/cli#411` (the dirty-tree refusal half already shipped there as `civitai/cli#415`).

ZacxDev:
Closes the server half of civitai/civitai#4059. Client tracking issue: civitai/cli#411.

`appBlocks.submitVersion` accepted a bundle and stored **nothing** about where it came from, so deploy-vs-source drift could be observed but never diagnosed — five first-party apps were found behind their live version and recovering the mapping took a session of archaeology. `submitVersionSchema` is a plain `z.object`, so a client sending provenance today is **silently stripped**: it looks like it worked and stores nothing.

**This is the server half. The CLI stamping half (civitai/cli#411) is deliberately blocked on it** and is not built until this lands — the CLI already computes the commit sha and the dirty flag for its own dirty-tree refusal (civitai/cli#415); it has had nowhere to put them.

## What lands

Two **nullable** columns on `app_block_publish_requests`, accepted at submit and returned on read:

| column | meaning |
|---|---|
| `source_commit` | the 40-hex git sha the **client says** the bundle was built from |
| `source_dirty` | whether the **client says** that work tree had uncommitted changes |

### Decisions, stated

- **CLIENT-CLAIMED, NEVER SERVER-VERIFIED.** The server validates the *shape* and stores a *claim*. It cannot confirm the uploaded bundle was actually built from that commit, and a client is free to send a sha it never built from. Nothing downstream may render either value as an attestation. Said in the schema comment, the service param doc, the column comment and the migration header.
- **NOT `forgejo_commit_sha`**, which the same row already carries. That one is **server-side**, written on approve after the platform's own Forgejo commit succeeds — a different commit in a different repository. Never aliased, defaulted, or fallen back between. `recordPendingFromPush` (the Forgejo push path) deliberately leaves both new columns NULL, and there is a test that fails if its sha ever leaks into `sourceCommit`.
- **Both fields optional.** A client that sends neither submits exactly as it does today. This is never a submit gate.
- **A malformed `sourceCommit` REJECTS with 400**, it is not silently dropped — silently dropping is the exact inert-feature failure the issue exists to prevent. Lowercase 40-hex only (git's canonical rendering; accepting uppercase too would make one commit two distinct strings in the column).
- **NULL is not false.** `source_dirty` is tri-state: `NULL` = unknown (pre-feature row, or a client that sent nothing), `false` = the client asserted CLEAN, `true` = the client asserted DIRTY. **No backfill and no DEFAULT**, because either would convert "nobody looked" into "someone looked and it was clean" for every historical row — positive evidence manufactured out of an absence. Two paired tests pin that `null` and `false` stay distinguishable end to end.

### 🔴 Two things the original plan did not cover, flagged rather than buried

1. **Both submit front doors are wired, not just one.** The issue and the plan named `src/pages/api/blocks/submit-version.ts` (the cookie/mod-browser route). But the CLI posts to **`src/pages/api/v1/blocks/submit-version.ts`** — the bearer-token route. Wiring only the first would have left the actual client's provenance silently stripped, reproducing the very bug this closes. Both routes now pass the fields through and both are pinned by tests.
2. **The 400 body no longer answers every parse failure with the flat `'Invalid bundle payload'`.** A provenance-shape rejection reported as a bundle problem sends a CLI author hunting the zip. A failure confined to the provenance fields now names the field (`Invalid submit payload: sourceCommit must be a 40-character lowercase hex git commit sha`); a genuine bundle failure — or a mixed failure, or a non-object body — still reads **exactly** the legacy string, pinned by tests on both routes and on the helper. The predicate lives in one place (`submitVersionParseErrorMessage`) rather than being open-coded at the two call sites.

## 🔴 The migration is AUTHORED but NOT APPLIED to any database

`packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql`

Nothing was run against any DB — no `db:migrate`, no `db:deploy`, no `db:push`, no `db:applied`. Per the repo's manual-apply rule the main civitai CNPG DB does not auto-apply; a human applies the SQL per environment.

**Apply it BEFORE deploying the app that writes these columns.** Both columns are nullable with no default, so the mixed-pod window is safe in both directions *except* new-code-against-a-table-without-them: that INSERT names a column the table lacks, and Prisma's no-`select` queries on the model raise P2022. An old pod writing a row without them stays valid. **Never drop these columns on a rollback** — old Prisma clients emit explicit column lists so extra columns are inert to them, while dropping destroys the only record of which bundle came from which tree.

The migration header carries the verification queries for both the schema and the behaviour.

## Files

| file | change |
|---|---|
| `packages/civitai-db-schema/prisma/schema.full.prisma` | `sourceCommit String? @map("source_commit")`, `sourceDirty Boolean? @map("source_dirty")` on `AppBlockPublishRequest`, with the forgejo-sha contrast comment |
| `.../prisma/migrations/20260817210000_.../migration.sql` | **new**, authored only. `ALTER TABLE "app_block_publish_requests" ADD COLUMN IF NOT EXISTS …` |
| `packages/civitai-db-schema/src/{models.ts,kysely/types.ts}` | regenerated by `pnpm db:generate` (2 lines each) |
| `src/server/schema/blocks/publish-request.schema.ts` | the two optional fields + `submitVersionParseErrorMessage` |
| `src/pages/api/blocks/submit-version.ts` | pass-through + named rejection (mod browser route) |
| `src/pages/api/v1/blocks/submit-version.ts` | pass-through + named rejection (**the CLI's route**) |
| `src/server/services/blocks/publish-request.service.ts` | `SubmitVersionParams` gains both; persisted in the submit `create`. `recordPendingFromPush`'s create untouched apart from a comment stating why it stays NULL |
| `src/pages/api/v1/blocks/submissions.ts` | `SELECT` projection + `shapeRow` return, both raw (no coercion) |
| 5 test files (1 new, 4 extended) | +40 assertions |

## Red/green matrix

Method: the five changed **source** files were copied aside, replaced with their `origin/main` versions (test files left at HEAD), the suites run, then restored **from the copies**.

- **Base (`origin/main` sources + HEAD tests): 33 failed | 222 passed (255).**
- **HEAD: 0 failed | 255 passed (255), 6 files.**

40 tests are new or newly-asserting. 33 are red at base. The remaining **7 are invariant guards, not regression coverage**, and are labelled as such here because they were green at base by construction — they pin that the no-provenance path is unchanged:

| invariant guard (green at base) | file |
|---|---|
| `ACCEPTS a payload with NEITHER field` | schema |
| `a submit with NO provenance still reaches the service with both undefined` | both route suites (×2) |
| `BUNDLE rejections keep the exact legacy message` | both route suites (×2) |
| `#4059 writes NEITHER column when the client sends nothing` | service |
| `recordPendingFromPush` sets neither column | orchestration |

## Mutation matrix

Each mutant: exact single-occurrence replace, **re-read from disk and asserted landed** before the verdict was read, run, then restored from a pristine copy. Unmutated control green before **and** after the battery (255/255 both times), so a mutation that failed to apply could not read as SURVIVED.

| # | mutation | verdict | killed by |
|---|---|---|---|
| a | drop the length anchor — `{40}` → `+` | **KILLED** (2 failed) | `REJECTS a 39-hex sourceCommit`, `REJECTS a 41-hex sourceCommit` |
| b | widen the class — `[0-9a-f]` → `[0-9a-fA-F]` | **KILLED** (2) | `REJECTS an UPPERCASE 40-hex sourceCommit` (schema), `400s an UPPERCASE sourceCommit` (route) |
| c | remove `.optional()` from `sourceCommit` | **KILLED** (12) | `ACCEPTS a payload with NEITHER field`, `a submit with NO provenance…`, + 10 pre-existing happy-path tests |
| d | **delete the whole `sourceCommit` field** (anti-strip control) | **KILLED** (14) | `ACCEPTS a well-formed pair AND KEEPS BOTH VALUES on parsed.data`, all 6 reject tests, `NAMES sourceCommit…`, route pass-through |
| e | delete the whole `sourceDirty` field | **KILLED** (9) | `KEEPS sourceDirty:false`, `REJECTS sourceDirty as the STRING "true"`, `NAMES sourceDirty…`, route pass-through |
| f | drop `sourceCommit: true` from the read `SELECT` | **KILLED** (1) | `the SELECT projection asks the DB for both columns` |
| g | collapse null→false — `row.sourceDirty ?? false` | **KILLED** (2) | `a legacy row keeps sourceDirty:NULL`, `happy path: … safe shape` |
| h | service drops `sourceCommit` from the `create` data | **KILLED** (2) | `#4059 PERSISTS sourceCommit + sourceDirty…`, `#4059 does NOT conflate sourceCommit with forgejoCommitSha` |
| i | service falls back — `sourceCommit ?? bundleSha256` | **KILLED** (1) | `#4059 writes NEITHER column when the client sends nothing` |

**Known limit, stated rather than papered over:** mutant (f) is killed only by the *structural* projection test. The read-path suite mocks `findMany`'s return directly, so dropping the `SELECT` cannot change a mocked row — no behavioural test in this repo's harness can see it. The structural ledger is the only instrument available at this layer.

## Gates

| gate | result |
|---|---|
| `pnpm typecheck` | **OK — 0 type errors in 72s** |
| targeted suites (6 files) | **6 passed, Tests 255 passed (255)**; counted 255 `✓` lines / 0 `×` lines, not the exit code |
| wider blocks sweep (`services/blocks`, `tests/api/blocks`, `tests/api/v1/blocks`, `schema/blocks`, `routers/__tests__`) | **217 files passed, 5060 tests passed** |
| `@civitai/db-schema` package suite | **17 files, 423 tests passed** |
| `pnpm db:check-generated` | **exit 0**, clean tree |
| `eslint` on the changed files | 3 errors, **all three pre-existing on `origin/main`** and untouched by this diff (`no-empty-function` at `submissions.ts:295` and `tests/api/v1/blocks/submit-version.test.ts:68,383` — the `.catch(() => {})` and `new Promise(() => {})` lines). 28 warnings, all pre-existing `no-explicit-any`. No new lint finding. |
| `prettier --check` | 4 files flagged — **all 4 already flagged on `origin/main`** with the same in-repo config. A prettier pass over the changed files touches only pre-existing lines, none of the added code. Not reformatted, to keep the diff readable. |

Fixes civitai/civitai#4059

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
@@ -96,9 +96,14 @@
 -- `tests/preview-apps-publish.spec.ts`, the only smoke spec that submits — on
 -- two successive commits, while a control PR passed 65/65 in the same window.
 -- The SQL below was then applied to that dev clone (columns confirmed present,
--- nullable, no default; a re-apply is a clean no-op). Whether that clears the
--- smoke run is being confirmed on the commit carrying this correction — do not
--- read this paragraph as reporting a green.
+-- nullable, no default; a re-apply is a clean no-op) and the very next run went
+-- `success — 65 passed, 0 flaky`, matching the control PR exactly. So the apply
+-- is not merely correlated with the fix: it is the whole fix, and this migration
+-- is a HARD PREREQUISITE of the deploy rather than a tidy-up that can follow it.
+--
+-- 🔴 THE SAME APPLIES TO PROD, FOR THE SAME REASON. The preview failure is what
+-- a prod deploy-before-apply looks like, observed in a safe place: every submit
+-- and every status poll, down, for everyone, until the columns exist.
 --
 -- Same precedent as 20260731120000_app_block_spend_tier_and_cap_override; the
 -- generic form of the hazard is that Prisma enumerates every scalar in the
--- a/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
@@ -54,21 +54,39 @@
 -- ------------------------------------------------------------------
 -- ⚠️ ORDERING: APPLY THIS BEFORE DEPLOYING THE APP THAT WRITES IT.
 -- ------------------------------------------------------------------
--- Both columns are nullable with no default, so the mixed-pod window is safe in
--- BOTH directions:
+-- Both columns are nullable with no default, so once they EXIST the mixed-pod
+-- window is safe in both directions. Apply this first and there is no window at
+-- all. Deploy first and the READ path is what breaks, immediately and for
+-- everyone:
 --
---   writer │ columns  │ outcome
---   ───────┼──────────┼──────────────────────────────────────────────
---   old    │ present  │ safe — INSERT omits them, both land NULL
---   new    │ present  │ safe — the claim is stored
---   new    │ MISSING  │ ✗ the INSERT names a column the table lacks
---   old    │ MISSING  │ today's behaviour
+--   code │ columns  │ outcome
+--   ─────┼──────────┼────────────────────────────────────────────────────────
+--   old  │ present  │ safe — the SELECT never names them; INSERT omits them
+--   new  │ present  │ safe — the claim is read and stored
+--   old  │ MISSING  │ today's behaviour
+--   new  │ MISSING  │ ✗✗ TOTAL READ OUTAGE, then a conditional write failure
 --
--- Only the third row breaks, and only if the deploy lands first. Note the
--- failure is not confined to the write: Prisma enumerates every scalar in the
--- model when a query gives no `select`, so with the code deployed and the
--- columns missing, such a query raises P2022. Same precedent as
--- 20260731120000_app_block_spend_tier_and_cap_override.
+-- 🔴 THE READ IS THE FIRST AND WORST CASUALTY, NOT THE WRITE.
+-- `src/pages/api/v1/blocks/submissions.ts` adds BOTH columns to its `SELECT`
+-- constant, which serves BOTH the `findFirst` and the `findMany`. That is
+-- UNCONDITIONAL: it does not depend on what any client sends, on whether any row
+-- has provenance, or on anyone using the feature. Against a table without the
+-- columns it is a 100% failure of `GET /api/v1/blocks/submissions` — every
+-- author, every CLI poll, every request — with Prisma P2022. The CLI's status
+-- and poll commands go down wholesale.
+--
+-- The write failure is real but SECONDARY and CONDITIONAL: Prisma omits an
+-- `undefined` field entirely, so a client that sends no provenance (i.e. every
+-- pre-#4059 client, and the whole world on day one) produces exactly today's
+-- INSERT and succeeds. Only a submit that actually CARRIES provenance names a
+-- column the table lacks. The schema deliberately normalises a JSON `null` to
+-- `undefined` for this reason — see `submitVersionSchema` — so "unknown" does
+-- not widen this into an unconditional write failure too.
+--
+-- Same precedent as 20260731120000_app_block_spend_tier_and_cap_override; the
+-- generic form of the read hazard is that Prisma enumerates every scalar in the
+-- model when a query gives no `select` at all, so even queries that never
+-- mention these columns raise P2022 once the model knows about them.
 --
 -- ⚠️ NEVER DROP THESE COLUMNS ON A ROLLBACK. Old Prisma clients emit explicit
 -- column lists, so extra columns are inert to them — leaving them costs nothing,
--- a/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260817210000_app_block_publish_request_source_provenance/migration.sql
@@ -64,29 +64,47 @@
 --   old  │ present  │ safe — the SELECT never names them; INSERT omits them
 --   new  │ present  │ safe — the claim is read and stored
 --   old  │ MISSING  │ today's behaviour
---   new  │ MISSING  │ ✗✗ TOTAL READ OUTAGE, then a conditional write failure
+--   new  │ MISSING  │ ✗✗ TOTAL OUTAGE OF BOTH READ **AND** WRITE
 --
--- 🔴 THE READ IS THE FIRST AND WORST CASUALTY, NOT THE WRITE.
--- `src/pages/api/v1/blocks/submissions.ts` adds BOTH columns to its `SELECT`
--- constant, which serves BOTH the `findFirst` and the `findMany`. That is
--- UNCONDITIONAL: it does not depend on what any client sends, on whether any row
--- has provenance, or on anyone using the feature. Against a table without the
--- columns it is a 100% failure of `GET /api/v1/blocks/submissions` — every
--- author, every CLI poll, every request — with Prisma P2022. The CLI's status
--- and poll commands go down wholesale.
+-- 🔴 BOTH PATHS FAIL, BOTH UNCONDITIONALLY. An earlier revision of this comment
+-- called the write "SECONDARY and CONDITIONAL — only a submit that actually
+-- CARRIES provenance names a column the table lacks". **That was wrong**, and it
+-- was wrong in the reassuring direction. It is corrected here rather than
+-- deleted, because "day one is safe for everyone who sends nothing" is exactly
+-- the belief that gets a deploy scheduled ahead of an apply.
 --
--- The write failure is real but SECONDARY and CONDITIONAL: Prisma omits an
--- `undefined` field entirely, so a client that sends no provenance (i.e. every
--- pre-#4059 client, and the whole world on day one) produces exactly today's
--- INSERT and succeeds. Only a submit that actually CARRIES provenance names a
--- column the table lacks. The schema deliberately normalises a JSON `null` to
--- `undefined` for this reason — see `submitVersionSchema` — so "unknown" does
--- not widen this into an unconditional write failure too.
+-- READ — `src/pages/api/v1/blocks/submissions.ts` adds BOTH columns to its
+-- `SELECT` constant, which serves BOTH the `findFirst` and the `findMany`. It
+-- does not depend on what any client sends, on whether any row has provenance,
+-- or on anyone using the feature: a 100% failure of
+-- `GET /api/v1/blocks/submissions` with Prisma P2022. The CLI's status and poll
+-- commands go down wholesale.
+--
+-- WRITE — equally unconditional, by a DIFFERENT mechanism. It is true that
+-- Prisma omits an `undefined` field from the INSERT column list, so the *data*
+-- a no-provenance client sends is unchanged. That is not what breaks. The
+-- `create` at `publish-request.service.ts:1249` passes **no `select`**, so
+-- Prisma reads the row back — `INSERT … RETURNING <every scalar in the model>` —
+-- and the model now knows about both columns. So EVERY submit raises P2022,
+-- including from a client that has never heard of provenance. `submitVersion`
+-- is the only writer on that path, and both front doors funnel through it.
+--
+-- This is the SAME generic hazard stated below, applied to the write; the
+-- earlier text asserted the general rule and then exempted the write from it.
+-- MEASURED, not reasoned: with the code deployed to a preview whose dev-clone DB
+-- lacked the columns, `preview / smoke-tests` failed with exactly 1 failure —
+-- `tests/preview-apps-publish.spec.ts`, the only smoke spec that submits — on
+-- two successive commits, while a control PR passed 65/65 in the same window.
+-- The SQL below was then applied to that dev clone (columns confirmed present,
+-- nullable, no default; a re-apply is a clean no-op). Whether that clears the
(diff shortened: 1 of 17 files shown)


ZacxDev:
## Audit fixes — `3c80dfc741`

Three findings from an adversarial audit of this PR. **If you reviewed before this commit, the schema, the `resetToPending` clone and the migration header have all changed.** Behaviour of the happy path is unchanged; the malformed-value rejections are unchanged.

---

### 1. 🟡 Explicit JSON `null` rejected the whole submit with 400

`submitVersionSchema` used `.optional()`, which accepts `undefined` **only**. JSON has no `undefined`, so `null` is the natural encoding of the UNKNOWN state this feature is built on — and it is what any client holding these as nullable in its own model sends.

Measured against the repo's zod (4.0.17), before the fix:

| body | result |
|---|---|
| `{"sourceDirty": null}` | **REJECT** |
| `{"sourceCommit": null}` | **REJECT** — *"sourceCommit must be a 40-character lowercase hex git commit sha"* |

That second message is wrong about what happened, and both contradict this PR's own stated rule that provenance **must never become a submit gate**.

**Fix:** both fields are now `.nullish()` with a `?? undefined` normalisation. The **wire contract** — previously stated nowhere and tested nowhere — is now written into the schema:

```
omit the key   → UNKNOWN
null           → UNKNOWN   (the same answer, spelled twice)
false          → client asserted CLEAN
true           → client asserted DIRTY
```

The tri-state is unchanged and is now pinned by a test that walks all three.

The `?? undefined` is **load-bearing, not tidying**: Prisma OMITS an `undefined` field from the INSERT but emits an explicit `NULL` for a `null` one. The resulting **row** is identical either way; the **SQL** is not. Letting `null` reach `create` would make every provenance-less submit name the new columns, widening the ordering hazard in §3 from *conditional* to *unconditional*.

**Deliberately NOT added:** a `.max()` on `sourceCommit`. I measured it and it buys nothing — zod runs the regex regardless (2 issues rather than 1 for an over-long value), and the anchored fixed-length regex costs **~0.02ms on a 70 MB string** either way. An initial 13ms reading was V8 warm-up, not regex work; re-measuring in reverse order with 3 reps each showed both at 0.02ms. Adding it would have implied a protection it does not provide.

### 2. 🟡 `resetToPending` silently dropped the provenance it could legitimately carry

The onsite reset-to-pending clone re-submits a **byte-identical bundle** (same `bundleKey`, same `bundleSha256`) and already carried `manifest`, `fileSummary`, `manifestDiffSummary` and `forgejoCommitSha` forward — but never *fetched* `sourceCommit`/`sourceDirty`. Any app through a suspend → reset-to-pending cycle permanently lost the answer to *"which tree did these bytes come from?"* — the archaeology this PR exists to remove.

**Fix:** both are selected and carried. The justification is narrow and is stated in the code: the bytes are identical, so the claim about their origin is still the same claim about the same bytes. Copied **raw, including NULL**, with no `??` fallback of any kind — least of all to `forgejoCommitSha`, which would fabricate an author's-tree claim nobody made.

`recordPendingFromPush` is **untouched** — it correctly leaves both NULL (no client, no author work tree).

### 3. 🔴 The migration header mis-stated which path breaks first

It named the INSERT as the failure mode. That is the narrower and **conditional** path — Prisma omits `undefined` fields, so a client sending no provenance produces today's INSERT exactly.

The real first casualty is the **READ**: `src/pages/api/v1/blocks/submissions.ts` adds both columns to a `SELECT` serving **both** the `findFirst` and the `findMany`, unconditionally, for every caller regardless of what any client sends. Against a table without the columns that is a **100% failure of `GET /api/v1/blocks/submissions`** — every author, every CLI poll — with P2022.

The ordering table now states the read blast radius and keeps the INSERT note as the secondary, conditional case. **Comment only — the SQL is unchanged.**

### Also: an invariant guard that was being counted as coverage

The `recordPendingFromPush` provenance block in `publish-request.orchestration.test.ts` was **measured** to be an invariant guard, not regression coverage — deleting both fields from the `submitVersion` INSERT left that file green; only `publish-request.service.test.ts` went red. It is now labelled as such, and its vacuous `expect(...).not.toBe(pushArgs.sha)` (strictly implied by the `toBeUndefined()` above it, so it could never fail on its own) is replaced by a key-**absence** assertion that can fail independently: an explicit `sourceCommit: undefined` satisfies `toBeUndefined()` and fails the new check. Absence is also the load-bearing property — it is what keeps that path's SQL byte-for-byte what it was before this PR.

---

## Verification

A review fix resets the gate, so nothing below is carried over from the previous round.

**Red-before / green-after** (new tests, watched failing against the pre-fix source at `c7c58f30`, restored from `cp -a` copies):

| fix | new tests | at `c7c58f30` | at `3c80dfc741` |
|---|---|---|---|
| 1 — `null` accepted | 7 schema + 4 route | **7 + 4 RED** | green |
| 2 — carry-forward | 4 | **3 RED**, 1 passes (labelled invariant guard) | green |
| 3 — migration header | comment only | n/a | n/a |

**Full mutation battery — 26 mutants, 24 killed, 0 survived**, each verified as *landed* by re-reading the line from disk before its verdict was read, and each restored from a checksum-verified pristine copy. Includes a **positive control** (breaking `INVALID_BUNDLE_MESSAGE`) which was KILLED, so the battery is demonstrably observing the tree.

The two new mutants this round both **KILLED**:
- `.nullish()` reverted to `.optional()` → 6 failed
- provenance carry-forward dropped from `resetToPending` → 2 failed

Also killed: field deletions from the schema / service `create` / both routes; regex widened to uppercase; regex de-anchored; `?? false` in `shapeRow`; each column dropped from the read `SELECT`; the two hint strings swapped; `?? forgejoCommitSha` and `?? false` fallbacks in `resetToPending`; and an explicit `sourceCommit: undefined` in `recordPendingFromPush` — that last one confirmed to die **at the new key-absence assertion specifically**, with its own error, rather than incidentally.

Two mutants initially reported ERROR rather than a verdict: my landed-check asserted the pattern count *dropped*, which is wrong for an **insertion** whose replacement contains the original. They were re-scored with a corrected check (marker on disk + exact line-count delta) and both KILLED. They were never silently scored as SURVIVED.

**Gates** (content and counts read, not exit codes):
- `pnpm typecheck` → `OK — 0 type errors in 92s`
- vitest, 10 explicit paths → `Test Files 10 passed (10)` / `Tests 389 passed (389)` — file count compared against the number of paths passed, so a filter matching nothing cannot pass silently
- `pnpm --filter @civitai/db-schema drift:gate` → `OK: no new drift on columns the database already has`. Both new columns appear under *"NEW drift awaiting a migration (not blocking)"*, which is correct — this migration is manual-apply and has not been applied.
- eslint: 2 errors / 27 warnings on the touched files, **all on pre-existing lines** — verified by intersecting the reported line numbers with the set of lines this commit adds (overlap 0).
- prettier: 5 files flagged, **all already flagged at `c7c58f30`**, and prettier would not reformat a single line this commit adds (per-file overlap with added lines = 0).

PR #4067, opened 2026-08-18
feat(event-engine): count comments on every surface toward the owner

JustMaier:
Completes the forward path for `User.commentCount` before any backfill runs. ClickUp `868ktbka8`.

## Why

`User.commentCount` shipped in #4002 covering Post / Image / Article / Bounty owners — **55% of comments**. Backfilling a complete history into a metric that goes incomplete from tonight onward is harder to spot than one that is wrong everywhere, so the emit surface goes first.

Measured on prod (root-resolved, matching the shipped handler):

| surface | comments | emitted before | now |
|---|---|---|---|
| Image | 1,442,746 | ✅ | ✅ |
| Model (legacy `Comment`) | 1,166,296 | ❌ | ✅ |
| Article | 222,325 | ✅ | ✅ |
| Post | 163,048 | ✅ | ✅ |
| Review | 109,521 | ❌ | ✅ |
| Bounty | 40,443 | ✅ | ✅ |
| BountyEntry | 32,460 | ❌ | ✅ |
| Challenge | 8,210 | ❌ | ✅ |
| Comic | 847 | ❌ | ✅ |
| Answer | 554 | ❌ | ✅ |
| ClubPost | 288 | ❌ | ✅ |
| Model3D | 161 | ❌ | ✅ |
| Question | 78 | ❌ | ✅ |
| Model (CommentsV2) | 16 | ❌ | ✅ |
| Model3DReview / AppListing | 0 | ❌ | ✅ |

Answer, Question and the CommentsV2 model arm were not on the ticket's table — they turned up when I counted every `Thread` FK rather than the listed ones. All small, all wired anyway.

## What changed

- **`comment-v2.ts`** — one query now resolves the entity *and* its owner across all fifteen commentable `Thread` columns. Owner columns are not uniform: `Challenge` and `ClubPost` record `createdById`, `app_listings` is snake-cased, and a comic thread points at the chapter's project rather than at an owner.
- **`comments.ts`** — legacy model comments look up `Model.userId` and emit for the creator. This is the single largest gap.
- **New tests** — emission, self-exclusion, the null-owner control, and a check that every `Thread` FK is still named in the query.

**Entity metrics deliberately stay at the four surfaces that already have them.** A new `(entityType, metricType)` pair needs a hand-applied `entityMetricKind` row; without one it falls into the presence path and counts *distinct commenters* instead of comments. Nothing in CI can catch that — `grep -r entityMetricKind` over the repo returns 0 hits. This PR introduces **no new pair**: `('User','commentCount','additive')` is already registered.

Both invariants from #4002 hold on every new arm: self-authored comments excluded, replies attributed to the root entity via `COALESCE(r.<id>, t.<id>)`.

## Verification

`apps/*` is outside the root tsconfig `include`, so `tekton / typecheck` never reads this code and `pnpm lint` / `pnpm test` are the only gates.

- `pnpm --filter @civitai/event-engine typecheck` — clean. **Proved it can fail on this file first** by planting a deliberate type error (`TS2322` at `comment-v2.ts:133`), then removing it.
- `pnpm run lint` — no issues on the three changed files.
- `pnpm run test` — 11 passed. Four of the assertions are negative (nothing was emitted), which pass for free unless proven otherwise, so each was mutation-tested:

| mutation | result |
|---|---|
| drop self-exclusion in `comment-v2` | red |
| drop null-owner guard | red |
| drop the `ResourceReview` join arm | red |
| drop self-exclusion in `comments.ts` | red |
| restored | green |

- **Against prod**: all fifteen surfaces resolve a real owner (one sample thread each, including a reply through `rootThreadId`). `EXPLAIN ANALYZE` on the widened query is index scans skipped on null — **0.23 ms execution, 4.2 ms planning**. The one seq scan is `app_listings` at 29 rows.

A note on `userId = -1`: it is the real `civitai` account, not a null placeholder, and it owns 7,574 models carrying 13,650 comments. Those count toward it. `entityId` is `Int32` in ClickHouse so the negative id is safe; the existing Download → Kafka filter excludes `-1` deliberately, which is a different decision for a different metric.

## Not in this PR

Backfill (work item 2) and the Creator Studio switch (`868krw7jf` steps 4-6). Justin decided the backfill runs from Postgres only: history loses the ~305k comments on deleted content whose threads have no recoverable owner, and the forward path never decrements, so deleted content keeps its comments from here on. The two halves are inconsistent by construction and that is accepted — maintaining deletes forward is not worth the work.

### DIFF
--- a/apps/event-engine/scripts/backfill-user-comment-count.mjs
+++ b/apps/event-engine/scripts/backfill-user-comment-count.mjs
@@ -8,26 +8,40 @@
  *   - Post / Image / Article / Bounty  -> event-engine v1.9.13 (#4002)
  *   - every other surface              -> the release carrying #4067
  * Backfilling both to one boundary either double-counts the first group or loses everything the
- * second group received between the run and its own release. Both boundaries are required.
+ * second group received between the run and its own release.
  *
  * Totals are read from daily aggregates, not from the event stream: entityMetricDaily_today_v2_mv
- * only ever looks at `createdAt >= today() - 1`, so historically-dated rows written into
- * entityMetricEvents_month would be invisible to every total. The backfill therefore writes day
- * totals into entityMetricDailyAgg_history_v2 and marks each entity dirty so the additive
- * refresher recomputes it.
+ * and entityMetricDailySeal_v2_mv both filter `createdAt >= today() - 1`, so historically-dated
+ * rows written into entityMetricEvents_month would be invisible to every total. The backfill
+ * therefore writes day totals into entityMetricDailyAgg_history_v2 and marks each entity dirty so
+ * entityMetricTotal_v3_refresher_additive recomputes it.
  *
  * That table is a ReplacingMergeTree versioned by `sealedAt`, keyed
- * (entityType, entityId, metricType, day) — a written row REPLACES that day rather than adding to
- * it. Days that already carry live events are therefore reported, not written.
+ * (entityType, entityId, metricType, day), and history rows carry NO surface dimension. A written
+ * row REPLACES that day for that user rather than adding to it. Every day from the first live
+ * event onward already holds sealed live totals for the four old surfaces, so writing a
+ * new-surface-only total over it would destroy both halves. All such days are excluded here and
+ * need an explicit merge (backfill total + sealed total, written with a fresh sealedAt).
+ *
+ * This is only readable while ('User','commentCount') is registered `additive` in
+ * entityMetricKind, which the script asserts before writing. Flip that row and the presence-based
+ * refresher recomputes the total from entityMetricUserState_v3 — which holds only live events —
+ * and silently overwrites every backfilled total with a much smaller number.
+ *
+ * Known and accepted divergences from the live path, both small: a comment created before the
+ * cutover and deleted during the run is subtracted twice; and two comments from the same commenter
+ * to the same owner in the same second collapse to one live (ReplacingMergeTree key) but count as
+ * two here.
  *
  * Usage:
  *   node scripts/backfill-user-comment-count.mjs --new-cutover 2026-08-18T12:00:00Z
  *   node scripts/backfill-user-comment-count.mjs --new-cutover ... --execute
  *
  * Options:
- *   --new-cutover <iso>  When the release carrying #4067 went live. Required.
+ *   --new-cutover <iso>  When the release carrying #4067 went live. Required, must be in the past.
  *   --execute            Actually write. Default is a dry run that writes nothing.
- *   --limit <n>          Only write the first n day rows (smoke test).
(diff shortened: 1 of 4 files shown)


JustMaier:
Pushed `7f80e8e014` — the backfill script, **dry run by default, writes nothing without `--execute`**. New SHA on the PR, so anything reviewed against `e93bfe9b66` needs a re-look.

**Where it writes, and why not the obvious place.** Not events into `entityMetricEvents_month`: `entityMetricDaily_today_v2_mv` only ever reads `createdAt >= today() - 1`, and the seal MV only seals yesterday, so a historically-dated event row is invisible to every total that exists. Day totals go into `entityMetricDailyAgg_history_v2`, then each owner is marked dirty so `entityMetricTotal_v3_refresher_additive` recomputes.

**Two boundaries, not one.** The four surfaces from #4002 stop at the first live event, read out of ClickHouse at run time rather than trusted from a ticket (`2026-08-17 21:28:21 UTC`, 596 events). Everything else stops at the release carrying this PR, which the caller passes as `--new-cutover`. The script refuses to run without it, and refuses a value earlier than the first live event.

**Boundary days are held back, not written.** `entityMetricDailyAgg_history_v2` is a ReplacingMergeTree versioned by `sealedAt` and keyed `(entityType, entityId, metricType, day)` — a written row **replaces** that day rather than adding to it. A day already carrying live totals therefore needs an explicit merge, so those rows are reported and excluded.

**Dry run against prod:**

```
post              116862 comments    84674 owner-days
image            1019047 comments   548040 owner-days
article           175022 comments    41398 owner-days
bounty             25432 comments    14374 owner-days
review             91702 comments    74883 owner-days
bountyEntry        24438 comments    15882 owner-days
challenge           8119 comments      293 owner-days
comic                575 comments      426 owner-days
clubPost             212 comments       62 owner-days
model3d              119 comments       93 owner-days
question              71 comments       70 owner-days
answer               492 comments      428 owner-days
model (legacy)    848031 comments   465051 owner-days

2310122 comments across 104354 owners -> 1153745 day rows
961 rows on a boundary day (2154 comments) held back
```

2,310,122 of 3,409,844 raw comments are attributable. The ~1.1M difference is mostly **self-authored comments**, which the metric excludes by design — image drops 1,442,746 → 1,019,047 (29%) and legacy model 1,166,316 → 848,031 (27%), consistent with the 57-71% self-comment share measured per-creator in #4002. The rest is the 214,356 on threads with no recoverable owner.

Both refusal paths were exercised: no `--new-cutover` → refuses; `--new-cutover` before the first live event → refuses naming both timestamps.

**Not run.** It cannot run correctly until this PR is released, because `--new-cutover` does not exist yet.

Issue #3983, opened 2026-08-16
App-store read path resolves the wrong StoreVisibilityScope in prod (tRPC always none, REST always full) — the real isFlipt is untested


ZacxDev:
## Summary

In production the App-store read path resolves a **store visibility scope that does not match `resolveStoreVisibilityScope`'s logic** — and it is wrong in *opposite directions* depending on the entry point:

| caller | tRPC `appListings.listAvailable` | REST `GET /api/v1/apps` |
|---|---|---|
| anonymous | `items: []` — behaves as `none` ✅ | 14 items incl. 10 `onsite` — behaves as `full` |
| non-moderator in the `testers` segment | `items: []` — behaves as `none` ❌ **should be `public-external` → the 4 offsite listings** | 14 items |

The user-visible consequence: **the external-only tester cohort reaches `/apps` and the store is completely empty** ("No apps yet"), even though 4 approved `offsite` listings exist. It is not a page-gate problem — the page renders; the data layer returns nothing.

The behaviour is **deterministic and user-independent** within each entry point. Corroborated on a second procedure sharing the same middleware: `appListings.getAppDetail` returns `NOT_FOUND` for **every** slug tried — `radio`, `vitrine` (offsite) *and* `sensei` (onsite).

(The public REST catalog access itself is intended product behaviour and is **not** what this issue asks to change. It is included only as evidence: the same resolver yields `full` there while yielding `none` on tRPC.)

## Repro

```bash
# Anonymous, no auth. Returns 14 items (10 onsite + 4 offsite).
curl -s https://civitai.com/api/v1/apps | jq '[.items[].kind] | group_by(.) | map({(.[0]): length}) | add'
```

Signed in as a non-moderator account in the `testers` segment (holds `app-listings-public-external`, not `app-listings`, not `app-blocks-enabled`), in the browser console on `civitai.com`:

```js
const inp = encodeURIComponent(JSON.stringify({json:{kind:"offsite",sort:"top-rated",limit:24,direction:"forward",authed:true}}));
await (await fetch("/api/trpc/appListings.listAvailable?input="+inp, {credentials:"include"})).json();
// → { result: { data: { json: { items: [] } } } }   ← expected: the 4 offsite listings
```

## What is already proven correct — please don't re-check these

| layer | verdict | evidence |
|---|---|---|
[...]

PR #4068, opened 2026-08-18
feat(build): assert security-relevant branches survive into the compiled output (#3983 detection)

ZacxDev:
Closes nothing on its own — this is the detection half of #3983. The cause is still open.

## Why a source-level test cannot catch this

Release 5.1.18 shipped `resolveStoreVisibilityScopeUninstrumented` as:

```js
async function S(e){if(await p(e))return"full"}
```

Two of its three `return`s were absent from the emitted server chunk, so the function fell off the end and produced `undefined` for every non-privileged caller. One read path defaulted that to `'full'` and served the whole App-store catalog to anonymous callers; the other defaulted the *same* missing value to `'none'` and showed the intended cohort an empty store.

The TypeScript is correct. That is the whole problem. A 75-test unit suite, an integration suite driving the real feature-flag client, and four rounds of review were all *structurally* incapable of seeing it, because every one of them exercises the source. Nothing looked at the artefact.

## What I confirmed before writing this

I rebuilt the exact commit that produced the serving image, in a container matching the CI toolchain (same base image digest, same package manager version, same lockfile, same build-time env injected the way CI injects it).

**It reproduces.** The chunk carrying that function is **byte-identical** to the one in the shipped image (same SHA-256, same filename), and across the whole emitted server tree only 5 of 26,223 files differ at all — all of them manifests the runtime image does not ship. So this is a deterministic property of the toolchain on this source, not a build-host accident, not an OOM, not a truncated artefact.

I also reduced it to a **28-line synthetic project** with no code from this repo, which reproduces on a stock Next 16.3.0 + Turbopack production build. That is going upstream separately.

🔴 **The synthetic reduction turned up a second, worse manifestation.** Hoisting the `await` out of the `if` condition — a change that looks like a workaround —

```ts
const b = await isB();
if (b) return 'full';
```

emits as `async function J(){return await Z(),"full"}`. The guard is gone entirely and the function returns `'full'` **unconditionally**. So this defect class can erase a gate as well as truncate one, and "just restructure the function" is not a safe reflex.

## What the gate does

Reads the emitted `.js.map` `mappings` and asserts that each watched fail-closed branch still has a representation in the output.

Attribution is by source map, never by grepping minified JS. That is not fastidiousness — every obvious grep is a trap here:

- minified names are per-chunk, and this one module is inlined into **234** chunks, so there is no stable symbol to anchor on;
- the literal `"public-external"` occurs **~481 times** in the server build and **not one** of them is a return — they are all elements of the closed-set array `["full","public-external","none"]`;
- the flag name `app-listings-public-external` **contains** the scope literal as a substring, adding ~238 more matches that are a different string entirely.

A line-mapping check is immune to all three, and also to the minifier collapsing `if (a) return x; return y;` into `return a?x:y` — the collapsed token still maps back to both source lines.

**Every entry carries a positive control**: a branch known to survive, which must also be mapped. If the control is unmapped the gate exits **2** ("cannot observe") rather than 1 ("violation"), so a build that simply did not emit the module is never reported as a pile of violations. A gate that cannot see must not report health *or* breakage.

Anchors are exact source substrings resolved to line numbers at run time, so an entry cannot silently rot when code moves; an anchor matching zero or multiple lines is itself a hard error.

**Adding a gate is one entry** in `scripts/compiled-branch-watchlist.mjs`.

## Why `--warn-only`

🔴 **This build currently violates the gate.** `main` and `release` carry byte-identical source for that function and both emit it truncated, so a hard gate in the Dockerfile would fail every production build today.

It reports loudly and exits 0 instead. `--warn-only` deliberately does **not** downgrade exit 2 — a gate that stopped being able to see its own input must still fail the build, or "stopped looking" becomes indistinguishable from "found nothing".

**Removing `--warn-only` is the definition of done for #3983.** A permanently-red gate that everyone learns to scroll past is worse than no gate, so this should not sit in warn-only indefinitely.

## Verification

- **13 cases**, each asserting its own failure branch's *specific* message and exit code — not merely "non-zero". Driven over synthetic `.next` trees whose maps carry real base64-VLQ `mappings`, encoded by an independent implementation in the test, so a decoder/encoder disagreement turns the healthy case red.
- The healthy baseline is derived from the **real** watchlist and the **real** source files, so a new entry is proven satisfiable rather than breaking the positive control, and a rotted anchor fails at PR time instead of turning the gate into an exit-2 no-op on the build host.
- **Six mutations of the gate, each killed by its own case**: violations never recorded; control check disabled; ambiguous-anchor check disabled; no-maps refusal downgraded to exit 0; source attribution matching everything; VLQ sign handling broken.
  🔴 That last one **initially survived** — every fixture listed its source lines ascending, so no encoded delta was ever negative and the decoder's sign branch was never reached. Real maps reorder freely, so that branch is the common case in the output this gate actually reads. The reordered-segments case was added to close it, and it now kills the mutant.
- Run against the **real 5.1.18 server artefact** (26,223 maps), the gate correctly reports both missing branches and exits 1; with `--warn-only` it prints the same report and exits 0.

## Blast radius, honestly bounded

I compared the compiled output against source for every function in `src/**` with two or more `return`s, using source-map line coverage over the whole server build.

- 4,341 such functions were emitted; **798** had at least one `return` line with no mapping.
- Filtering to those whose missing returns are string literals *and* whose literal appears nowhere as a returned value in any chunk carrying the module leaves **3**.
- Reading all three: one is `typeof window === 'undefined'` folding correctly in the server bundle, one is a legitimate ternary collapse where the value is plainly present, and one is this defect.

A second, independent census — declarations with no mapped line that a *surviving* sibling references — produced 179 candidates, of which 38 are functions rather than constants. All 37 others were read and are benign: identity functions inlined, compile-time-`false` kill switches, and `typeof window` folding (a control confirmed the fold is real and specific: `typeof window` survives unfolded in 0 server chunks while `typeof navigator|document` survives in 57).

**What this does not cover, stated plainly:** both censuses read the **server** build only. ~30 of those 38 candidates are client-side modules whose server-side collapse is correct but whose *client* emission — the thing that governs real browser behaviour — was never examined. A defect of this shape in a client module would be invisible to what I ran. Extending the census to the client output is the obvious follow-up and is not in this PR.

## Not in this PR

- The bundler defect itself. Nothing here fixes it; the consequence is already neutralised by #4041's closed-set narrowing.
- The upstream report and its synthetic reproduction.
- Client-bundle coverage for either census.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,35 @@ RUN --mount=type=cache,target=/app/.next/cache \
 # report health. Cheap: a few seconds of file reads.
 RUN node scripts/check-server-graph-singletons.mjs
 
+# Compiled-branch gate — the second HARD build-output gate, and for the same reason as the
+# one above: nothing that reads SOURCE can see this class of defect.
+#
+# Release 5.1.18 shipped `resolveStoreVisibilityScopeUninstrumented` as
+# `async function S(e){if(await p(e))return"full"}` — two of its three returns were absent
+# from the emitted chunk, so it returned `undefined` for every non-privileged caller. One
+# read path defaulted that to `'full'` and served the whole App-store catalog to anonymous
+# callers; the other defaulted it to `'none'` and showed the intended cohort an empty
+# store. The TypeScript was correct throughout, so a 75-test unit suite, an integration
+# suite driving the real feature-flag client, and four rounds of review were all
+# structurally incapable of catching it (civitai#3983).
+#
+# This reads the emitted `.js.map` `mappings` and asserts that each watched fail-closed
+# branch still has a representation in the output. Same placement rationale as the gate
+# above: `.next` exists in this stage, so there is no second build; and it exits 2 (not 0)
+# when it cannot observe its input. Adding a gate is one entry in
+# `scripts/compiled-branch-watchlist.mjs`.
+#
+# 🔴 `--warn-only` because THIS BUILD CURRENTLY VIOLATES IT. The bundler defect behind
+# civitai#3983 is not fixed — `main` and `release` carry byte-identical source for that
+# function and both emit it truncated — so a hard gate here would fail every production
+# build immediately. It reports loudly and exits 0 instead. `--warn-only` does NOT
+# downgrade exit 2, so a gate that cannot see its own input still fails the build.
+#
+# REMOVE `--warn-only` as part of fixing the underlying defect. That flip is the
+# definition of done for civitai#3983, and it is the only thing that turns this from a log
+# line back into a gate.
+RUN node scripts/assert-compiled-branches.mjs --warn-only
+
 # Bundle-size budget (report-only during the soak). Next 16 (Turbopack) emits
 # opaque hashed chunks and removed per-route build stats, so scripts/bundle-budget.mjs
 # parses .next/build-manifest.json to reconstruct per-page First Load JS (brotli)
--- a/scripts/__tests__/assert-compiled-branches.test.ts
+++ b/scripts/__tests__/assert-compiled-branches.test.ts
@@ -1,0 +1,312 @@
+import { spawnSync } from 'node:child_process';
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+import { COMPILED_BRANCH_WATCHLIST } from '../compiled-branch-watchlist.mjs';
+
+/**
+ * `scripts/assert-compiled-branches.mjs` is a BUILD gate: it reads the emitted server
+ * source maps and fails when a watched, security-relevant branch has no representation in
+ * the compiled output — the civitai#3983 shape, where two of a resolver's three `return`s
+ * were absent from the shipped chunk while the TypeScript was correct.
+ *
+ * These cases drive it over synthetic `.next/server` trees. That is deliberate: a real
+ * `next build` takes minutes, so nobody would run one per case, and the branches that
+ * matter (a missing return, an unobservable control, a build with no maps) are exactly
+ * the ones a healthy real build cannot produce on demand.
+ *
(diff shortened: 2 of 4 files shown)


ZacxDev:
## Root cause found, and both gate arms are now validated on real production output

**Cause:** Turbopack's value analyzer models a bare `return someAsyncFn()` tail call as returning `Promise<Promise<T>>`, which is always truthy, so a caller's `await` of it is treated as always-true and everything after the resulting conditional is eliminated as dead code. Upstream [vercel/next.js#96601](https://github.com/vercel/next.js/pull/96601), **fixed in Next 16.3.1**. We are pinned at **16.3.0**.

Our code hits it exactly: `isAppListingsEnabled` ends with `return isAppBlocksEnabled(opts);` (`app-blocks-flag.ts:258`).

**Both arms verified on real artefacts of the same commit** (`60d8087647`, identical container, only the `next` version differs):

- 16.3.0 build → gate **exit 1**, naming both missing branches
- 16.3.1 build → gate **exit 0**

**Blast radius, measured by diffing source-line coverage between the two builds:** of 4,656 source modules emitted by both, exactly **one** lost statement lines under 16.3.0 — `app-blocks-flag.ts`, six lines (the whole axis-2 function plus the two returns referencing it). Nothing else in the server build. Details in [#3983](https://github.com/civitai/civitai/issues/3983#issuecomment-5323828667).

**Consequence for this PR:** the `--warn-only` in the Dockerfile becomes removable as soon as `next` is bumped to 16.3.1. That bump is a separate change (semver-compatible — `package.json` already says `^16.3.0`, only the lockfile pin holds us at 16.3.0 — plus renaming `patches/next@16.3.0.patch` and its `patchedDependencies` key). It is **not** in this PR, deliberately: a toolchain bump deserves its own review and its own soak, and this PR is the detection half.

🔴 One thing worth flagging for whoever does that bump: **do not try to work around this in source instead.** Hoisting the `await` out of the `if` condition — the obvious-looking fix — emits on 16.3.0 as `return await Z(),"full"`, i.e. the guard disappears and the function grants `'full'` **unconditionally**. Same root cause, worse outcome.

PR #4069, opened 2026-08-18
chore(event-engine): keep the boundary-day merge that finished the backfill

JustMaier:
Keeps the script that finished the `User.commentCount` backfill from #4067, and records the sequencing rule it enforces.

**Why a separate step at all.** `entityMetricDailyAgg_history_v2` is keyed `(entityType, entityId, metricType, day)` and carries no surface dimension, so a written row **replaces** that day rather than adding to it. Every day from the first live event onward already holds live totals, so the backfill skips them — writing a backfill-only total there would destroy the live half.

**The rule this encodes:** a day can only be merged *after* its daily seal. Merging earlier is pointless — `entityMetricDailySeal_v2_mv` appends a live-only row with a newer `sealedAt`, and `argMax(total, sealedAt)` picks that. So `2026-08-17` (sealed at 01:00) could be merged immediately; `2026-08-18` cannot until after the `2026-08-19 01:00 UTC` seal. The script refuses a day with zero sealed rows rather than writing something that will be silently overwritten — exercised as a control before use.

**Applied to 2026-08-17:** 2,691 backfilled + 212 live = **2,903 across 1,200 owners**, verified back out of history.

**Still outstanding:** `2026-08-18` merge, after the 2026-08-19 01:00 UTC seal:
```
node scripts/merge-comment-count-boundary-day.mjs 2026-08-18            # dry run
node scripts/merge-comment-count-boundary-day.mjs 2026-08-18 --execute
```
Two comments landed in the 86-second window where pods were mixed 1.9.13/1.9.14 during the rollout; both fall on 08-18 and are resolved by that merge reading the sealed live half rather than assuming it.

The cutovers are constants at the top (`OLD_CUT` = first live event, `NEW_CUT` = the 1.9.14 rollout) because this is a one-time repair of two specific days, not a general tool.

### DIFF
--- a/apps/event-engine/scripts/merge-comment-count-boundary-day.mjs
+++ b/apps/event-engine/scripts/merge-comment-count-boundary-day.mjs
@@ -1,0 +1,109 @@
+// Merge one boundary day: backfill total + already-sealed live total, written with a fresh
+// sealedAt so argMax(total, sealedAt) picks the merged value. Only safe AFTER that day is sealed,
+// otherwise the daily seal appends a live-only row with a newer sealedAt and wins.
+import pg from 'pg';
+import { createClient } from '@clickhouse/client';
+
+const DAY = process.argv[2];
+const EXECUTE = process.argv.includes('--execute');
+const OLD_CUT = '2026-08-17 21:28:21';
+const NEW_CUT = '2026-08-18 05:43:30';
+const OLD = new Set(['post', 'image', 'article', 'bounty']);
+
+const S = [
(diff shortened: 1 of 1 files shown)


JustMaier:
Closing: this is a spent one-time script. It merges the comment-count boundary day for a backfill that is finished, with its cut times hardcoded (`OLD_CUT = '2026-08-17 21:28:21'`, `NEW_CUT = '2026-08-18 05:43:30'`), so it adds no runtime path and has no second use.

Closed on Justin's decision, 2026-09-21: one-time-use scripts are not kept in the tree.

PR #4075, opened 2026-08-18
fix(deps): Next 16.3.0 → 16.3.1, and make the compiled-branch gate hard (#3983)

### LINKED ISSUE #3983: App-store read path resolves the wrong StoreVisibilityScope in prod (tRPC always none, REST always full) — the real isFlipt is untested

ZacxDev:
Closes #3983.

## What this is

The fix for #3983 is a **patch bump of Next**, not a source rewrite. The defect was in the bundler.

Turbopack's value analyzer in **16.3.0** models a bare `return someAsyncFn()` tail call as `Promise<Promise<T>>`. That is always truthy, so a caller that `await`s it is analysed as always-true, and every statement after the resulting conditional is eliminated as dead code. `isAppListingsEnabled` ends in exactly that shape — which is why `resolveStoreVisibilityScopeUninstrumented` lost two of its three returns, fell off the end, and produced `undefined` for every non-privileged caller. One read path defaulted that to `'full'` (whole catalog to anonymous callers); the other defaulted the same missing value to `'none'` (empty store for the cohort that was meant to see something).

Upstream: [vercel/next.js#96601](https://github.com/vercel/next.js/pull/96601) *"[turbopack] Collapse nested promises in the analyzer"*, backported as [#96675](https://github.com/vercel/next.js/pull/96675), shipped in **16.3.1**.

## Evidence — the artefact, not the version string

Two **production builds of this same tree**, on one machine, same Node 24.19.0, differing only in the pinned Next. The compiled form of `resolveStoreVisibilityScopeUninstrumented`, read out of the emitted `.next/server` chunks by source-map attribution:

```
16.3.0  async function S(e){if(await p(e))return"full"}
16.3.1  async function w(e){return await c(e)?"full":await y(e)?"public-external":"none"}
```

Identified by its **source neighbour** (`STORE_SCOPE_FLAGS`, emitted immediately after it), never by minified name — names are per-chunk and this module is inlined into 234 of them. Reproduced across two independent 16.3.1 builds, three sampled chunks each.

🔴 **Note the fixed form is a ternary.** `grep 'return"public-external"'` returns **zero on the FIXED build too** — a grep would call this fix broken. That is exactly why the gate reads source maps.

## Gate: warn-only → hard

`scripts/assert-compiled-branches.mjs` (added in #4068) shipped with `--warn-only` for one build only, because `main`'s 16.3.0 build genuinely violated it and a permanently-red gate trains everyone to click through. The flag is removed here, in the same commit as the bump, so the gate's strictness always tracks the toolchain — and a revert of the bump turns it red instead of silently passing.

Watched **both** directions on this tree:

| build | `node scripts/assert-compiled-branches.mjs` |
|---|---|
| 16.3.1 | `✓ 2/2 store-visibility-scope` — **exit 0** |
| 16.3.0 | `✗ 2/2 MISSING` (line 789 `return 'public-external'`, line 791 `return 'none'`) — **exit 1** |

Both scanned 26,225 source maps and found the module in 234 chunks, so the gate observed its input in both arms; the red arm is a real violation, not a blind gate.

## Mechanics

- `package.json` already allowed 16.3.1 (`^16.3.0`), so **only the lockfile pinned 16.3.0** — the lockfile is the substance of this PR. The floor is raised to `^16.3.1` so a fresh resolution cannot land back on the broken compiler.
- The lockfile diff is minimal: `next@16.3.0 → 16.3.1` plus the peer-suffix rewrites that carry next's version, and `@swc/helpers 0.5.17 → 0.5.23` (next's own dep). No other package version moves.
- `patches/next@16.3.0.patch` → `patches/next@16.3.1.patch` + its `patchedDependencies` key. That patch is the unrelated libvips/SVG one-liner ([vercel/next.js#96681](https://github.com/vercel/next.js/pull/96681)). **It still applies cleanly**, and 16.3.1 still does *not* carry the loader entry upstream — the installed 16.3.1 has exactly one occurrence per copy, i.e. the patch is what puts it there — so it stays. `scripts/ci/assert-next-svg-patch-applied.mjs` passes on both installed copies. Two stale comments naming the old patch filename are corrected, including one that claimed 16.3.1 would carry the fix upstream; it does not.

## Other verification

- Unit suite: **1,149 files / 18,123 tests passed, 0 failed**, 1 file + 21 tests skipped. (Instrument control: an earlier run of the same command went red — 7 files, for a local `sharp` native-binary gap — so the runner does report failures.)
- Typecheck: clean, as part of both production builds (`Finished TypeScript in 56s`).
- Prettier: clean on the changed parseable files.

## Not verified — read before merging

- **Nothing is proven about production.** This changes what the compiler emits; it is only true of production once this merges, is promoted `main` → `release`, is built by Tekton, and is serving. The live behaviour must be re-checked then.
- **My local production builds did not run to completion.** Both arms reached `✓ Compiled successfully` and emitted the full `.next/server` tree, then failed identically at *Collecting page data* for want of production env/secrets I do not have locally (`NEXT_PUBLIC_CIVITAI_LINK`, Meilisearch host, Flipt, notifications, recaptcha). Identical on both arms, so not caused by the bump — but the first end-to-end production build is CI's, and the gate now runs hard inside it. If CI's Docker build goes red here, that is the signal, not noise.
- **Server output only.** The two-build source-map diff posted on #3983 measured the **server** blast radius as exactly 1 module / 6 lines (the axis-2 function plus the two returns referencing it), so nothing else in `.next/server` was affected. The **client** bundle was never built or A/B'd — the same defect class can occur there and remains unmeasured. The watchlist covers one function.
- **The gate covers what is in the watchlist and nothing else** — one entry today.
- No behaviour change is claimed for any other part of the app from the toolchain bump beyond "the unit suite and typecheck are green".

---

## Follow-up in this PR: the bump built clean and the container would not boot

The first push of this branch passed **every** source-level gate — unit suites, typecheck, ESLint + Prettier, schema drift, the event-engine pin, and the now-hard compiled-branch gate — and the PR-preview deploy crash-looped at startup:

```
Error: Cannot find module '/app/node_modules/.pnpm/next@16.3.1_patch_hash=…/node_modules/@swc/helpers/esm/_interop_require_default.js'
    at … next/dist/server/require-hook.js
  code: 'MODULE_NOT_FOUND'
```

That is the same shape as the defect this PR exists to fix: a correct source tree producing a broken artefact, with nothing that reads source able to see it.

### Root cause — measured on the published artefact, not inferred

`output: 'standalone'` does not ship `node_modules`; it ships the subset **@vercel/nft** traced. nft resolves a bare specifier under the `require`/`default` conditions. **Node (>= 22.10) additionally honours `module-sync` for a CJS `require`.** When a package's `exports` map points those two at different files, the build traces one file and the running process asks for the other.

Next's own `dist/shared/lib/constants.js` does `require('@swc/helpers/_/_interop_require_default')`, reached from the generated `server.js` via `next` → `config.js` → `constants.js` — i.e. **before any application code**. The relevant delta of this bump is not `next` itself but next's own dependency:

| | next 16.3.0 | next 16.3.1 |
|---|---|---|
| `@swc/helpers` | 0.5.15 | 0.5.23 |
| `./_/_interop_require_default` exports | `{ import, default }` | `{ module-sync, webpack, import, default }` |
| `require.resolve(…)` under CJS | `cjs/_interop_require_default.cjs` | `esm/_interop_require_default.js` |

Both rows of the last line were run, not reasoned about. nft still traced the `cjs/` file, so the published image contained that package as exactly `cjs/_interop_require_default.cjs`, `cjs/_interop_require_wildcard.cjs`, `package.json` — **no `esm/` directory at all**.

**Confirmation that this was the whole blocker:** adding only the missing `esm/` directory to that exact published image, changing nothing else, boots it — `▲ Next.js 16.3.1 … ✓ Ready`, proceeding to the runtime env it has no credentials for.

### The fix

`outputFileTracingIncludes` force-includes **both** condition branches (`esm/` and `cjs/`) of **every** installed `@swc/helpers` copy. Not the one file missing today: which helper Next requires, and which branch each resolver picks, are upstream details that move. Globs are version- and hash-agnostic (`@swc+helpers@*`), plus a flat form for a hoisted layout. ~950 KB per copy.

It is attached to existing API-route keys rather than a `'**'` key. `copyTracedFiles` unions every entry's traced set into the single `.next/standalone` `node_modules`, so one entry carrying it is enough — while `'**'` would make all 572 entries read/parse/rewrite their `.nft.json` concurrently, 826 MB of JSON in one `Promise.all`, on a build already tuned against OOM.

### The gate — because a glob is a silent no-op when it stops matching

`scripts/ci/assert-standalone-boot-graph.mjs`, run in the Dockerfile's **runner** stage. It is the first gate in that file that runs against the **runtime filesystem** rather than the build tree, and the only one that can see this class of defect. It reads the *generated* `server.js` for the specifiers that process requires at module scope and loads them in a child rooted at the shipped tree. Nothing is hardcoded — no package, version, virtual-store path or patch hash — so it keeps covering this after the next bump. Exit 2, never 0, when it cannot observe its input.

It must run in the runner and not the builder: `/app` there is byte-for-byte what ships, whereas the builder's complete `node_modules` sits above `.next/standalone` on the resolution path and can satisfy a require the image cannot.

**Watched red and green on the real published image**, not only on fixtures: exit **1** with this exact `MODULE_NOT_FOUND` against the broken image, exit **0** against the same image with the `esm/` directory added and nothing else changed. `src/tests/build/standalone-boot-graph.test.ts` pins the mechanism (a `module-sync`/`default` split with only the `default` branch present) rather than the package, and all 5 of its cases were watched to fail against a neutered gate.

### Still not verified

- **Nothing about production.** This is only true of production once this merges, is promoted `main` → `release`, is built, and is serving.
- The gate covers the **entrypoint's** require graph. Route chunks load lazily; a condition mismatch reachable only from a route would still surface at request time, not build time.


ZacxDev:
### Preview deploy is green; the red `Unit tests` is `main`, not this branch

**Preview:** all 7 sub-checks pass (`deploy`, `render-check`, `unit-tests`, `component-tests`, `auth-tests`, `apps-tests`, `packages-tests`) and the preview container reaches Ready with **0 restarts** and serves requests. That is the check that was red before, and it is the only arm that runs the built artefact.

**`Unit tests` (red) is a pre-existing break on `main`, not from this branch.** One failure in 18,170: `no-direct-shared-module-mock` > *adds no new direct mock of a module that has a canonical mock*, naming `src/server/utils/__tests__/moderator-endpoint.test.ts`. That file does not exist on this branch — it was added to `main` by `394b4bc80` ("migrate retool endpoints to moderator endpoints") at 17:31 UTC, ~14 min before this run. It does `vi.mock('~/server/redis/client', …)`, a CANONICAL specifier, and is not in `src/__tests__/mocks/direct-mock-allowlist.json`. Actions runs PR checks on the **merge** commit, so this branch is simply the first PR to run against a `main` that contains it — the sibling PRs showing green all last ran 16:01–16:20 UTC, i.e. before it landed.

Ran the guard's own scan logic over each tree separately:

| tree | files scanned | violations not in the allowlist |
|---|---|---|
| `origin/main` alone | 1300 | `src/server/utils/__tests__/moderator-endpoint.test.ts` |
| this branch's HEAD alone | 1299 | *(none)* |

The fix belongs on `main` — either migrate that file to the canonical mock (`node scripts/test-perf/codemod-shared-mocks.mjs --write src/server/utils/__tests__/moderator-endpoint.test.ts`) or allowlist it — not in this PR. Every other check here is green.

ZacxDev:
Post-merge correction, for anyone reading this later: the comments this PR introduced about the SVG-loader patch assert the opposite of what's true, and **16.3.1 does not need the patch at all**. #4086 removes it and fixes the wording.

The claim was that the installed 16.3.1 carries the loader entry *because our patch puts it there, and upstream still does not*. Upstream does. The published `next@16.3.1` tarball ships `'VipsForeignLoadSvg'` in the `sharp.unblock({ operation: [...] })` list in **both** `dist/server/image-optimizer.js` and `dist/esm/server/image-optimizer.js` — that release is the first to contain vercel/next.js#96681.

The measurement behind the claim was real, but it can't support the conclusion: **"exactly one occurrence per copy" looks identical whether upstream supplied the entry or a patch did.** What distinguishes them is comparing the installed file against the published tarball — those are byte-identical (sha256), so the patch contributed nothing. It doesn't even apply: its hunk context wants `Png` immediately followed by `Tiff`, and 16.3.1 has `Svg` between them (`patch` reports *"Reversed (or previously applied) patch detected… 1 out of 1 hunk ignored"`; `git apply --check` reports *"patch does not apply"*).

The reason it looked healthy is worth remembering, because it will happen again: **pnpm silently no-ops an already-applied patch.** The install exits 0, prints no warning, and still creates a `patch_hash=…` virtual-store directory — so a dead patch stays indistinguishable from a live one indefinitely.

Renaming the key to `next@16.3.1` also silenced the `ERR_PNPM_UNUSED_PATCH` tripwire, which was the one signal designed to force this decision at exactly this moment.

None of this affects what this PR got right — the `@swc/helpers` 0.5.15→0.5.23 export-condition fix is real and load-bearing (`module-sync` resolves a CJS `require` to `esm/` on Node ≥22.10, which is a boot-time `MODULE_NOT_FOUND` in the standalone image), and the preview verified `/api/og` renders PNGs on both paths after a genuine image-optimizer cache miss.

The guard stays either way: it asserts the *outcome* — the installed Next unblocks the SVG loader — so it keeps protecting against a future Next regressing this the way 16.3.0 did, which returned HTTP 500 for 100% of `/api/og` for days.

ZacxDev:
Heads-up for whoever sees the preview build failures stop after this: **the 16.3.1 bump re-rolls the failure, it does not fix it.**

There is an intermittent `Two or more assets with different content were emitted to the same output path` build failure that has been breaking preview builds on a subset of PRs. It was diagnosed just before this merged — writeup on `main` at `claudedocs/turbopack-chunk-hash-collision-2026-08-18.md` (landed in #4084).

Short version: it is a birthday collision in Turbopack's 7-character server chunk-name hash. The build emits ~24.5k server chunks, the first hash character is bounded in practice to `{0,1,2}` (leaving roughly 5% of nominal hash space), and the largest namespace holds ~11.7k chunks.

Measured across 16.3.0 and 16.3.1 on the same tree: hash width unchanged, first-character bounding unchanged, largest-namespace size unchanged, chunk count 24,551 vs 24,552. **Nothing determining collision probability moved.** What changes is graph contents, which re-rolls *which* pairs collide — so a tree that failed on 16.3.0 can build clean on 16.3.1 and a different tree will start failing later.

Concretely: if these failures go quiet now and reappear in a week or a month, this bump is not the regression and reverting it will not help. Please check the writeup before opening a new investigation.

The one lever that actually attacks the mechanism is `turbopackServerSideNestedAsyncChunking` (~24.5k → ~7.1k chunks; collision probability goes as n², so roughly 9× fewer). It is currently disabled for a reported ~+33% peak builder memory cost that has never been verified, and it does not compile on 16.3.0 — but it should be testable now that this bump has landed. There is a pointer comment at the flag in `next.config.mjs`.

PR #4078, opened 2026-08-18
feat(feedback): add an `apps-marketplace` feedback area and mount it on /apps

ZacxDev:
## What

Adds a second feedback surface — the **`apps-marketplace`** area — and mounts it at the top of `/apps`.

The per-area machinery already exists, so this is small by design: `FEEDBACK_AREAS` is a plain string union (no Prisma enum, no migration) and the Flipt flag key is derived from the slug, so the change is one constant, one mount, and the flag itself (added separately in the flag-state repo as `feedback-area-apps-marketplace`).

| | |
|---|---|
| notice | "Apps are new here. If something's missing or looks off, tell us." |
| placeholder | "What were you trying to do? Missing apps, broken listings, confusing pages, anything." |
| flag | `feedback-area-apps-marketplace` (derived) |

## What the report carries

`/apps` keeps its kind / category / sort / search state in the query string, so the context is a straight read of the URL through the same `parseAppsStoreFilters` the store body uses — one source, no lifted state, no second copy of the defaults. A one-line report ("the list is empty") arrives next to the view it was written about rather than just a path.

🔴 **The search term is CLIPPED.** `context.filters` bounds each value and a zod `max()` **rejects** rather than truncates, so passing a long search term through would fail the whole submission with a validation error the reporter cannot act on — on the one surface whose job is to accept reports. The clip lives in a named function (`buildAppsStoreFeedbackContext`) rather than inline in the page precisely so it is reachable by a test, and the bound it clips to is now the exported `FEEDBACK_FILTER_VALUE_MAX_LENGTH` (same value, 200, no widening) so the caller and the schema cannot drift apart.

`category: null` is reported as the string `'none'` rather than dropped — a missing key reads as "we don't know" in triage, which is a different claim from "no category was selected". `?? undefined` is not an alternative: an explicit `undefined` is a record *value* here and fails the union.

## Two decisions worth a look

- **`active` is unconditional.** The images-feed mount gates on `feedSnapshot.source === 'bitdex'` because that prompt asks about one specific backend and would misattribute reports about any other. This one asks about the page as a whole, so there is no equivalent condition; inventing one would only make the surface silently dark. Rollout is the flag's job.
- **Placement.** Above the store's search/sort row, mirroring where the images feed mounts it (above the grid, not after it). It costs no vertical space in the ordinary case — `FeedbackPrompt` already requires a signed-in user and returns null when the area resolves off, so for everyone outside the flag's segments the page is byte-identical to before and nothing is pushed below the fold.

## Tests

Extended the three existing feedback suites and added one for the context builder. 96 assertions green across the six touched/related files.

**Every new guard was watched to fail with the change reverted, for its own reason:**

| mutation | result |
|---|---|
| drop `'apps-marketplace'` from `FEEDBACK_AREAS` | **15 red** across all four files — `ZodError: invalid_value, values: ["bitdex-image-feed"]` |
| make the service ignore its `area` argument when deriving the flag key | **3 red** — `expected 'feedback-area-bitdex-image-feed' to be 'feedback-area-apps-marketplace'`; the bitdex cases stay green, which is what makes it an isolated kill |
| drop the `.slice(...)` clip (narrowest expression) | **2 red** — `expected … to have a length of 200 but got 201`; a negative control in the same block proves the unclipped value really is rejected by the schema, so "does not throw" is not vacuous |
| drop the `?? 'none'` category fallback | its own test red — `expected undefined to be 'none'` |

Typecheck: **94 pre-existing errors on this branch and 94 on `origin/main`, the same set byte for byte** (all in unrelated files, a stale generated client against unrelated schema changes) — zero introduced. Lint and prettier clean on all eight files.

## Not done, on purpose

- **No read endpoint.** The feedback system is write-only today and that is out of scope here.
- **No test on the page mount itself.** The context *builder* is covered, the area and flag key are covered end-to-end through the real router, but nothing asserts that `/apps` actually renders a `<FeedbackPrompt area="apps-marketplace">`. A structural scan of the page source would be a spelled guard rather than a behavioural one; a real render test of a `getServerSideProps` page is a bigger piece of scaffolding than this change warrants. Flagging it rather than pretending it is covered.
- The flag lands in the flag-state repo, in its own PR. Neither order breaks anything: with the flag on and this code absent nothing renders, and with this code present and the flag absent the prompt is simply dark.

### DIFF
--- a/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
+++ b/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
@@ -111,6 +111,86 @@ describe('buildAppsStoreFeedbackContext', () => {
     });
 
     /**
+     * 🔴 THE SURROGATE CASE. Every fixture above is ASCII, and a fixture that cannot
+     * contain a surrogate pair cannot observe a surrogate bug — which is exactly why
+     * this shipped.
+     *
+     * `slice` cuts on UTF-16 CODE UNITS, not characters. An astral-plane character
+     * (any emoji) is a pair of units, so a clip point that lands between them leaves
+     * a LONE HIGH SURROGATE as the last unit. That string is a legal JS value and
+     * passes `z.string().max()` untouched, so the schema round-trip below cannot see
+     * it either — it breaks one layer further down, where the context is written to
+     * the `Feedback.context` JSON column: `JSON.stringify` emits a bare `\ud83d`
+     * escape, which Postgres rejects (`Unicode low surrogate must follow a high
+     * surrogate`) and Prisma's serde layer rejects before that. So the clip added to
+     * keep a long search term from failing the submission ends up failing the
+     * submission itself, on the one surface that exists to accept reports.
+     *
+     * Asserted as a STATE (`isWellFormed`, plus a real UTF-8 encode round-trip)
+     * rather than by looking for an escape sequence in the output: a well-formed
+     * string is the property the downstream write actually needs, and a lone
+     * surrogate is the only way to lose it.
+     */
+    const EMOJI = '\u{1F600}'; // U+1F600, two UTF-16 units: \uD83D \uDE00
+
+    /** Encoding to UTF-8 and back is lossless iff the string has no lone surrogate. */
+    const survivesUtf8 = (value: string) =>
+      new TextDecoder().decode(new TextEncoder().encode(value)) === value;
+
+    it('does not leave a lone surrogate when the bound falls inside an emoji', () => {
+      // 199 ASCII + one emoji = 201 units, so unit 200 is the emoji's HIGH surrogate.
+      const query = 'q'.repeat(FEEDBACK_FILTER_VALUE_MAX_LENGTH - 1) + EMOJI;
+      expect(query).toHaveLength(FEEDBACK_FILTER_VALUE_MAX_LENGTH + 1);
+
+      const context = buildAppsStoreFeedbackContext({ path: '/apps', filters: filters({ query }) });
+      const clipped = context?.filters?.query as string;
+
+      expect(clipped.isWellFormed()).toBe(true);
+      expect(survivesUtf8(clipped)).toBe(true);
+      // Exactly the orphaned unit is dropped — the whole pair, and nothing more.
+      expect(clipped).toHaveLength(FEEDBACK_FILTER_VALUE_MAX_LENGTH - 1);
+      expect(clipped).toBe('q'.repeat(FEEDBACK_FILTER_VALUE_MAX_LENGTH - 1));
+      expect(throughSchema(context)?.filters?.query).toBe(clipped);
+    });
+
+    it('does not leave a lone surrogate for a term that is mostly emoji', () => {
+      // One ASCII char shifts every pair onto an odd offset, so the clip at 200 lands
+      // mid-pair. (A term of ONLY emoji does not break at this bound: 200 is even, so
+      // the cut falls exactly on a pair boundary — which is why the fixture is mixed.)
+      const query = 'q' + EMOJI.repeat(150);
+      const context = buildAppsStoreFeedbackContext({ path: '/apps', filters: filters({ query }) });
+      const clipped = context?.filters?.query as string;
+
+      expect(clipped.isWellFormed()).toBe(true);
+      expect(survivesUtf8(clipped)).toBe(true);
+      expect(clipped).toHaveLength(FEEDBACK_FILTER_VALUE_MAX_LENGTH - 1);
+      expect(clipped.endsWith(EMOJI)).toBe(true);
+      expect(throughSchema(context)?.filters?.query).toBe(clipped);
+    });
+
+    /**
+     * The control for the two above: when the clip point does NOT fall inside a pair,
+     * nothing extra may be trimmed. Without it, a guard that simply dropped the last
+     * unit of every clipped term would pass both cases above.
+     */
+    it('trims nothing extra when the bound falls on a character boundary', () => {
+      const query = 'q'.repeat(FEEDBACK_FILTER_VALUE_MAX_LENGTH) + EMOJI;
+      const context = buildAppsStoreFeedbackContext({ path: '/apps', filters: filters({ query }) });
+
+      expect(context?.filters?.query).toBe('q'.repeat(FEEDBACK_FILTER_VALUE_MAX_LENGTH));
+      expect(context?.filters?.query).toHaveLength(FEEDBACK_FILTER_VALUE_MAX_LENGTH);
+    });
+
+    /** And an emoji comfortably inside the bound is not touched at all. */
+    it('leaves an emoji inside the bound intact', () => {
+      const query = `upscale ${EMOJI}`;
+      const context = buildAppsStoreFeedbackContext({ path: '/apps', filters: filters({ query }) });
+
+      expect(context?.filters?.query).toBe(query);
+      expect(throughSchema(context)?.filters?.query).toBe(query);
+    });
+
+    /**
      * The control for the two cases above: an UNCLIPPED value of the same shape is
      * genuinely rejected. Without this, "does not throw" would be indistinguishable
      * from a schema that has no bound at all.
--- a/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
+++ b/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
@@ -1,0 +1,128 @@
+import { describe, expect, it } from 'vitest';
+import { buildAppsStoreFeedbackContext } from '~/components/Apps/appsStoreFeedbackContext';
+import { APPS_STORE_DEFAULTS } from '~/components/Apps/appsStoreQueryParams';
+import { createFeedbackSchema } from '~/server/schema/feedback.schema';
+import { FEEDBACK_FILTER_VALUE_MAX_LENGTH } from '~/shared/constants/feedback.constants';
+
+/**
+ * The triage payload the `/apps` marketplace prompt attaches to a report.
+ *
+ * The load-bearing case is the LAST describe block. `context.filters` bounds every
+ * value at `FEEDBACK_FILTER_VALUE_MAX_LENGTH` and a zod `max()` REJECTS rather than
+ * truncates, so an unclipped search term does not arrive shortened — it fails the
+ * whole submission. That is why the clip is a real guard and not tidiness, and why
+ * these assertions drive the built context through `createFeedbackSchema` rather
+ * than merely reading the returned string's length: a length assertion alone would
+ * still pass if the schema's bound moved away from the constant the builder clips to.
+ */
+describe('buildAppsStoreFeedbackContext', () => {
+  const filters = (over: Partial<typeof APPS_STORE_DEFAULTS> = {}) => ({
+    ...APPS_STORE_DEFAULTS,
+    ...over,
+  });
+
+  /** Round-trip through the real boundary schema — this is what "carried" means. */
+  const throughSchema = (context: ReturnType<typeof buildAppsStoreFeedbackContext>) =>
+    createFeedbackSchema.parse({
+      area: 'apps-marketplace',
+      message: 'the store looked wrong',
+      context,
+    }).context;
+
+  it('reports the path and every store control', () => {
+    const context = buildAppsStoreFeedbackContext({
+      path: '/apps',
+      filters: filters({
+        kind: 'offsite',
+        category: 'generation',
+        sort: 'newest',
+        query: 'upscale',
+      }),
+    });
+
+    expect(context?.path).toBe('/apps');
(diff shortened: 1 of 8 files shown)


ZacxDev:
### `preview / deploy` is red for a reason that is not this PR

Attributed rather than assumed, because "it fails on my branch" and "my branch broke it" are different claims.

**The failure.** The image build dies in `next build`:

```
Error: Turbopack build failed with 2 errors:
[output]/.next/server/chunks/ssr/[root-of-the-server]__0gduaxh._.js
Error: Two or more assets with different content were emitted to the same output path
file content differs, written to:
  [output]/.next/abf442f9b482f9c2.js
  [output]/.next/5e217de9956f5326.js
```

**Not a flake.** Re-run on the identical tree: same error, same chunk, same hashes. So it depends on the tree, not on scheduling.

**The control.** Built `origin/main` and this branch locally, same toolchain, same flags, back to back. **Both fail, identically** — same colliding output path `[root-of-the-server]__0gduaxh._.js` and the same emitted `.map` hashes (`c3526da896688aac` / `9a3c59f6509f5181`) that the CI build reported here. The collision is therefore present on `main` without a line of this change.

That is also the shape you would expect: the colliding path is a content-hashed server chunk that names no file this PR touches, and this PR adds a TS module, a JSX mount and tests — it emits no assets. It looks like the known Turbopack duplicate-emit bug, whose visibility depends on the state of the build cache the image build mounts, which is why a neighbouring PR's preview can go green in the same window.

What I did **not** establish: why it started, or which commit on `main` first exposed it. That needs a bisect over `main`, which belongs in its own issue rather than here.

---

Also for the record: the earlier `Unit tests` red on this PR was `main`'s, not mine — `src/server/utils/__tests__/moderator-endpoint.test.ts` tripping the shared-mock ratchet, introduced by #4077 (which merged with that check red) and fixed on `main` by #4079. Rebased onto the fix; that check is green now.

ZacxDev:
## Adversarial audit: **no deploy-blocking findings.** One 🟡 I'd fix before merge, two to answer before the flag is switched on.

### 🟡 1 — the clip can manufacture a lone surrogate, re-creating the failure it exists to prevent

`appsStoreFeedbackContext.ts` clips the search term with `.slice(0, FEEDBACK_FILTER_VALUE_MAX_LENGTH)`, which cuts on **UTF-16 code units**. A term of 199 ASCII chars + one emoji is length 201; clipping to 200 leaves a trailing **lone high surrogate**.

Verified directly:

```
input length (UTF-16 units): 201
clipped length:              200
JSON.stringify tail:         xxxxx\ud83d"
last unit is lone high surrogate: true
```

`JSON.stringify` emits `"\ud83d"`, and Postgres rejects that outright — confirmed with a positive control: `'{"q":"\ud83d"}'::jsonb` → `ERROR: Unicode low surrogate must follow a high surrogate`, while the paired `'{"q":"😀"}'::jsonb` → `{"q": "😀"}`. The serde layer in Prisma's query engine rejects it too, so the conclusion doesn't depend on which layer bites first.

Result: an emoji-containing search term near the bound fails the whole submission — on the surface whose job is to accept reports, which is exactly the class of bug the clip was added to close.

**Why the tests miss it: every clip fixture is ASCII** (`'q'.repeat`, `'x'.repeat`). A fixture that cannot contain a surrogate pair cannot see a surrogate bug. One-line fix: after slicing, drop a trailing `[\uD800-\uDBFF]`.

### 🟡 2 — the 5/hour limit is GLOBAL, not per-area, so this halves the existing surface's budget

`middleware.trpc.ts` builds the bucket as `keyName = options?.sharedKey ?? path.replace('.', ':')`, and `feedback.router.ts` passes no `sharedKey` — so the key is `trpc:limit:feedback:create` per user, with **no area component**. Verified.

A user who files 5 reports here is locked out of the image-feed prompt for the rest of the hour, and gets "You have submitted a lot of feedback." Defensible as anti-abuse, but it is an **unstated behaviour change to an existing surface** and isn't in the PR description.

### 🟡 3 — the search term is a new persisted data flow, captured silently

The context puts `?query=` — arbitrary user-typed text — into the JSONB `context` column. Today that text **never leaves the browser**: the store's search is client-side only (`AppListingsMarketplaceBody.tsx` notes the list input has no `query` field). This is the first time it reaches a database.

Note the inconsistency on the same panel: the screenshot is opt-in with explicit copy, attachments are hand-picked — the search term is captured with no disclosure. Retention is unbounded; deletion only via `onDelete: Cascade` on the user relation.

Mitigating: the term is on screen anyway, submitting is an explicit action, and only the 4 known filter fields are read — `resolveAppsStoreFilters` ignores everything else, so campaign/`utm_*` params are **not** carried. Reasonable to keep for triage value, but it wants an explicit yes and probably one clause in the notice copy. The bound here should be *disclosure*, not length.

### 🟡 4 — the documented upload budget doesn't hold for this surface's actual audience

`feedback.constants.ts` derives "20 uploads/hour" from `FEEDBACK_RATE_LIMIT.max × (FEEDBACK_IMAGE_MAX_COUNT + 1)`. But the limiter **exempts moderators outright** (`middleware.trpc.ts`: `if (ctx.user?.isModerator || …) return await next()`), and `/apps` is gated on `appListings || appBlocks`, both `availability: ['mod']`. So the population that can currently reach this prompt is largely rate-limit-exempt, and that comment reads as a bound that doesn't apply here.

Same root cause makes the flag's `early-adopters` arm effectively a no-op on this page today — the surface lights for mods, and four of the new tests exercise a path that can't occur in production until the store widens. Not a defect; worth knowing the rollout lever is half-connected.

### 🟢 Nits

- The mount reads `window.location.pathname` **during render** on a page with a documented hydration incident. Safe only because the value is never rendered — fragile by inspection, and copied verbatim from the images-feed mount. `router.asPath.split('?')[0]` is SSR-stable. On this page `path` is the constant `/apps` anyway, so it carries nothing `area` doesn't.
- `useRouter()` at the page root means every filter change and 300 ms search-debounce URL write now re-renders `AppsPage`/`Meta`/`AppsPageLayout`, which previously didn't. Re-render, not remount — pure cost.
- "byte-identical page" is true of the DOM only: every signed-in non-dismissed visitor now issues a `feedback.getArea` call and one flag evaluation per page view, flag on or off. Batched, cached, `ssr: false` — small, but not nothing.
- "clip and bound cannot drift apart" is a **test tripwire**, not a structural coupling — the constant is asserted `.toBe(200)` in two files, so a change reddens 3 tests and a human retypes them. Works; the claim overstates it.
- `expect([...FEEDBACK_AREAS]).toEqual(areas)` is order-sensitive, so reordering the constant fails the suite.

### On the known "no test asserts /apps renders the prompt" gap — agreed, with a reason not given

CI runs `vitest run --project unit` only, and the browser-mode suites aren't run by the gate at all. So even a `.browser.test.tsx` render of `/apps` would be **unrun** — reassurance, not coverage. No cheap non-spelled guard exists. Accept it.

### CI attribution — verified rather than accepted

`Unit tests` is **green** on this head (the #4077 breakage was fixed by #4079). `preview / deploy` fails, and is **not** attributable to this PR — the load-bearing control is **#4079, whose entire diff is one test file**, failing at the same `build-image` step with the identical Turbopack "two assets emitted to the same output path" error; **#4082** (two API files + tests) also failed it.

⚠️ Correcting the earlier claim in this PR's thread: it is **intermittent, not universal** — #4080, #4081, #4072/3/4, #4077, #4061/2 all succeeded in the same window while this PR failed 3/3. The builds share a `--mount=type=cache` Next cache, which is a plausible cross-build race. Don't read 3/3 as exoneration on its own; the #4079 control is what carries it. Root cause belongs in its own issue.

### Verdict

**Safe to merge.** Small, additively typed, no migration, dark behind a flag, no type errors introduced (94 pre-existing errors, zero in touched files; `tekton / typecheck` green), merges clean with no file overlap against main. The author's mutation battery re-verified independently, plus a **wrong-bind mutant they didn't try** (swapping the `category`/`sort` binds → 3 red) — fixtures are pairwise distinct, so it dies.

Fixing 🟡1 now; 🟡2 and 🟡3 want answers before the flag goes on, not before merge.

ZacxDev:
### Fix: the search-term clip could emit a lone surrogate and fail the whole submission

Pushed as `ec568a4fac`. One narrow guard plus the fixtures that can see it.

#### The defect

`buildAppsStoreFeedbackContext` clipped the `/apps` search term with `.slice(0, FEEDBACK_FILTER_VALUE_MAX_LENGTH)`. `String.prototype.slice` cuts on **UTF-16 code units**, and the bound is a unit count — but an astral-plane character (any emoji) is a **pair** of units. 199 ASCII characters + one emoji is 201 units, so the clip at 200 lands *between* the pair:

```
input length (UTF-16 units): 201
clipped length:              200
JSON.stringify tail:         qqqqq\ud83d"
last unit is lone high surrogate: true
```

That string is a legal JS value and passes `z.string().max(200)` untouched, so nothing on the request path notices. It breaks at the write — `Feedback.context` is a JSON column, `JSON.stringify` emits a bare `"\ud83d"`, and both Prisma's serde layer and Postgres reject it (`Unicode low surrogate must follow a high surrogate`).

Net effect: **an emoji in the search term near the bound makes the whole report fail** — which is the exact failure mode the clip was added to prevent, on the one surface whose job is to accept reports.

#### The fix

```ts
const clipFilterValue = (value: string) =>
  value.slice(0, FEEDBACK_FILTER_VALUE_MAX_LENGTH).replace(/[\uD800-\uDBFF]$/, '');
```

A high surrogate that survives the slice as the **last** unit is necessarily unpaired — its low surrogate could only have followed it, and the slice removed everything after — so dropping it is the whole fix, and it cannot touch a well-formed clip. The builder is otherwise unchanged.

#### Why a green suite missed it

**Every clip fixture was ASCII** (`'q'.repeat(...)`, `'x'.repeat(...)`), and a fixture that cannot contain a surrogate pair cannot observe a surrogate bug. The round-trip through `createFeedbackSchema` could not catch it either — zod accepts the ill-formed string happily; the rejection is one layer further down.

#### Red → green

Two new cases, both watched to fail on the pre-fix code:

| test | pre-fix |
|---|---|
| `does not leave a lone surrogate when the bound falls inside an emoji` | ❌ `AssertionError: expected false to be true` at `expect(clipped.isWellFormed()).toBe(true)` |
| `does not leave a lone surrogate for a term that is mostly emoji` | ❌ same assertion, same message |

```
Tests  2 failed | 11 passed (13)     <- pre-fix
Tests  13 passed (13)                <- post-fix
```

Wider re-run, no regressions: **51 files / 927 tests passed** across `src/components/Apps` plus the feedback schema, service and router suites. Assertions pin a **state** — `isWellFormed()` plus a real UTF-8 encode/decode round-trip — rather than grepping the output for an escape sequence, because a well-formed string is the property the write actually needs.

One fixture note: the mostly-emoji case is deliberately **mixed** (`'q' + emoji.repeat(150)`). A term of *only* emoji does not break at this bound — 200 is even, so the cut lands exactly on a pair boundary. The single ASCII character is what shifts every pair onto an odd offset.

#### Mutation check

Three isolated mutants, each the narrowest expression that can be wrong:

| mutant | result |
|---|---|
| delete only `.replace(/[\uD800-\uDBFF]$/, '')` | 🔴 2 failed — both new tests, on their own `isWellFormed()` assertion |
| drop only the `$` anchor | 🔴 2 failed — incl. `leaves an emoji inside the bound intact`: `expected 'upscale �' to be 'upscale 😀'` |
| substitute an over-trimming clip (`slice(0, MAX - 1)` whenever clipping) | 🔴 3 failed — incl. `trims nothing extra when the bound falls on a character boundary` |

The second and third confirm the two new **controls** are load-bearing rather than decorative: without them, an unanchored regex or a fix that just drops the last unit of every clipped term would pass.

#### Considered and not fixed — not reachable here

A lone surrogate arriving in the **middle** of the term would still survive this guard. It is not reachable on this path: these filters are a pure function of `router.query`, and every URL decoder replaces an encoded lone surrogate with `U+FFFD` while `encodeURIComponent` throws `URI malformed` on one. Measured, not assumed:

```
new URLSearchParams("query=%ED%A0%BD").get("query")  ->  "���"  (well-formed)
querystring.parse("query=%ED%A0%BD").query           ->  "���"  (well-formed)
encodeURIComponent("\uD83D")                         ->  throws URIError: URI malformed
```

So the clip is the only way this builder can produce an ill-formed value, and a mid-string guard would be dead code.

#### Not verified

- The Postgres/Prisma rejection is reproduced at the JS layer (the lone surrogate and the `"\ud83d"` that `JSON.stringify` emits) — **not** against a live database in this run; no DB was available. The unit assertions pin well-formedness, which is the property that write requires.
- `typecheck` reports 12 errors, all pre-existing and all in `image.service.ts` / `bitdex-stats.ts` (missing `event-engine-common` modules, unrelated implicit-`any`). **Zero** in either file this change touches. `prettier --check` clean on both, with a negative control confirming it can go red.
- `preview / deploy` red is the pre-existing intermittent reason already documented above — untouched by this change.

Issue #4076, opened 2026-08-18
SSR readiness probe gates on dbWrite: a write-primary stall sheds the entire fleet from the LB


ZacxDev:
## Summary

The `write` (Prisma `dbWrite`) check in `src/pages/api/health.ts` is a **critical** check, so it flips the overall `healthy` boolean that pod readiness gates on. That couples database write availability to load-balancer membership: when the write primary stalls, **every replica fails readiness within one probe window and leaves the pool simultaneously**, and the site returns "no available server" even though every process is alive, healthy and CPU-idle.

This happened in production on **2026-08-18**. A ~10-minute lock/contention stall on the write primary (preceded by a burst of `deadlock detected` errors on `image-scan-result`) produced:

- a flood of Prisma `Timed out fetching a new connection from the connection pool`,
- `health-check:dbWrite` failures logged as `Transaction API error: Unable to start a transaction in the given time.`,
- **all** SSR replicas flipping NotReady together, exactly one probe window (`periodSeconds × failureThreshold`) after the stall began,
- edge 503s at roughly **50× baseline** for ~20 minutes, with about half of serving capacity lost.

Two things then made it worse:

1. **A limit cycle.** With `successThreshold: 1`, the instant one probe succeeds every replica returns to the pool at once, full traffic slams back onto the still-degraded database, and one probe window later they all drop out together. This oscillated on a clean ~2-minute period for four cycles before the database cleared.
2. **The autoscaler scaled *down* mid-incident.** Because the 503s are served at the edge, backend CPU reads as idle, so the HPA removed replicas at the worst possible moment — and those replicas are warmup-gated, so they take minutes to come back.

## Why this is the right fix

**A pod that cannot reach the database should serve degraded responses, not withdraw from the load balancer.** Removing one replica from the pool is a useful reaction to *that replica* being sick. Removing every replica because a *shared* dependency is sick converts a partial degradation into a total outage — the load balancer has nowhere left to send traffic.

There is already a precedent for exactly this in the same file. `sysRedis` was made a soft dependency via `STATIC_NON_CRITICAL_CHECKS` after a 2026-06-26 outage, and the comment there describes the identical failure mode:

> a transient sysRedis stall … must not fail /api/health across all pods and shed the whole fleet from the LB

The same reasoning applies to a shared database. This issue proposes extending that established pattern — **but not before the app can actually survive it**, which is what step 1 is for.

[...]

PR #4082, opened 2026-08-18
fix(health): stop the DB write checks from shedding readiness fleet-wide

ZacxDev:
Closes step 2 of #4076.

## What

`write` (Prisma `dbWrite`) and `pgWrite` are **critical** health checks, so they flip the overall `healthy` boolean that pod readiness gates on. They probe a **shared** dependency: when the write primary stalls, both fail on every replica inside a single probe window, so every replica leaves the load balancer at the same moment and the site answers "no available server" while every process is alive, healthy and CPU-idle.

Removing one replica from the pool is a useful reaction to *that replica* being sick. Removing every replica because a *shared* dependency is sick converts a partial degradation into a total outage — the load balancer has nowhere left to send traffic.

## How

This extends the **existing** `STATIC_NON_CRITICAL_CHECKS` pattern (added for `sysRedis` after the same class of incident) rather than inventing a mechanism — but as a **mode-scoped** set, because startup and steady-state readiness genuinely differ:

- a pod that has **never** reached the DB must not enter the pool → **fail closed**
- a pod already serving that **loses** write must stay and degrade → **fail open**

`runHealthChecks` now takes `{ mode }`:

| route | probe | mode | write / pgWrite |
|---|---|---|---|
| `/api/health` | steady-state readiness | `'readiness'` (default) | soft — recorded, does not shed |
| `/api/ready` | startup | `'startup'` (explicit) | critical — fails closed |

🔴 **CORRECTION (see the comment below): the table above holds for the SSR/api/api-heavy pools ONLY.** Other deployments of this app use `/api/health` as their **startup** probe, and for those this PR removes the fail-closed-at-startup invariant. Details, consequence and remedy are in [this comment](https://github.com/civitai/civitai/pull/4082#issuecomment-5333390930). I have left the table as originally written rather than rewriting history.

The default is the **non-shedding** one, so adding a new probe route cannot opt into fleet-shedding by omission. A startup probe stops running once it first succeeds and a restarted pod runs it again, so fail-closed-at-boot holds for the pod's whole lifetime.

Like `STATIC_NON_CRITICAL_CHECKS`, the set is **hardcoded** rather than read from the runtime `NON_CRITICAL_HEALTHCHECKS` config, for the reason already documented in that file: the runtime list is read from a dependency, and a failed read degrades to "run ALL checks, suppress NOTHING" — so a runtime lever evaporates during precisely the incident it exists to mitigate.

### Both write checks move, not just the Prisma one

#4076 says to move `write` and *consider* `pgWrite`. Moving only `write` would have been **inert**: they are separate clients over separate pools, they fail independently, and in the incident their failures peaked within the same minute at the same order of magnitude. Either one left critical re-arms the fleet shed on its own.

### What does not change

- Both checks still **run**, still report their result in the response body, still increment their per-check counter and the labelled attempts counter, and still observe their duration histogram. Only the `healthy` boolean changes. `HEALTHCHECK_DISABLED` was rejected as the mechanism precisely because it stops the check and loses the signal an operator needs mid-incident.
- `read` / `pgRead` stay **critical in both modes** — a pod that cannot read is genuinely useless and should leave the pool.
- `sysRedis` stays soft in both modes.

## Tests

`src/tests/api/health.runHealthChecks.test.ts` gains 19 cases across both modes; each is labelled with the direction it pins — **non-shedding**, **fail-closed**, or **not-over-broadened**. Notably it covers the **acquisition-hang** shape (a parked `$transaction`), not only a fast throw: the observed production failure was connection-pool *acquisition* timing out, which `statement_timeout` does not bound — the comments in `health.ts` already say so.

`src/tests/api/health-ready.writeCriticality.test.ts` is **new and deliberately route-level**. The unit tests prove `runHealthChecks` honours each mode, but say nothing about which mode each *route* asks for, and that seam is where the dangerous defect lives. Verified: deleting `{ mode: 'startup' }` from `/api/ready` leaves **all 26 unit tests green** and is caught **only** by this file (`expected 200 to be 503`).

The pre-existing "sysRedis never rescues a real failure" case was repointed from `dbWrite` to `dbRead` — still critical — so it keeps testing the property it was written for.

### Red at base → green at HEAD

Base `adc6ede4f`, both new files run against the **unmodified** implementation:

```
base: Tests  9 failed | 22 passed (31)
HEAD: Tests           31 passed (31)
```

The 9 red-at-base are exactly the non-shedding cases plus the two seam cases. The 22 green-at-base are baselines, fail-closed cases and over-broadening guards — they pin behaviour this PR must **preserve**, and are labelled as invariant guards rather than counted as regression coverage. TypeScript independently reports the base red: `Expected 1 arguments, but got 2` where a test passes `{ mode: … }`.

### Mutation matrix

Each mutation was applied alone and killed, with a distinct and correct kill set:

| mutation | killed |
|---|---|
| drop `pgWrite` from the soft set | 3 (the pgWrite + both-fail + route cases) |
| drop `write` from the soft set | 8 |
| `startup` mode no longer strict | 7 (all fail-closed cases) |
| `/api/ready` drops its `mode` argument | 2 — **unit file stays 26/26 green** |
| default mode flipped to `startup` | 7 (all non-shedding cases) |

### Wider suite

`src/tests/api/` — 10 failing files / 12 tests at base and at HEAD, byte-identical failing-file sets; none of them related to these routes. `pnpm typecheck` — 119 pre-existing errors at base and at HEAD, identical per-file distribution, none in the touched files. `prettier --check` and `eslint` clean on all four files (the 3 `no-explicit-any` errors in `health.ts` are pre-existing on untouched lines, identical at base).

## Not in this PR

- The write-path audit — enumerating write-dependent surfaces and moving read-shaped queries off `dbWrite` (#4076 **step 1**), tracked separately. Note this PR does **not** claim browse flows survive a dead write primary in general; it claims a write-primary stall no longer empties the load balancer.
- `successThreshold` / jitter to break the recovery limit cycle (#4076 step 3) — deployment configuration, not app code.
- The `image-scan-result` deadlocks (#4076 step 4).

## Deployment ordering

Safe in either order, but worth knowing: the fail-open behaviour only takes effect where the readiness probe targets `/api/health` and the startup probe targets `/api/ready`. If a readiness probe is ever repointed at `/api/ready`, the protection is silently lost — there is a 🔴 comment on that route saying so.


ZacxDev:
## Correcting my own PR description — the probe table above is wrong

An adversarial audit of this branch found that the route↔probe table in the description states as **universal** something I only verified for three deployments. Posting rather than quietly editing the body, since anyone who has already read it read the wrong version.

**What the body claims:** readiness probes `/api/health`, startup probes `/api/ready`, therefore the fail-open/fail-closed split falls out of the existing wiring for free.

**What is actually deployed:** that holds for the SSR, api and api-heavy pools — and **only** those three. For other deployments of this app, `/api/health` **is** the startup probe:

- the **jobs** pool: startup probes `/api/health`; readiness and liveness are `tcpSocket`.
- **next**: startup *and* readiness probe `/api/health`.
- **next-stage**: startup, readiness *and liveness* probe `/api/health`.

**Consequence, which the description does not mention.** For those three, this PR removes the invariant it calls load-bearing — *"a pod that has NEVER reached the database must not enter the pool."* A jobs pod whose first write contact fails now passes startup and is Ready via `tcpSocket`, running scheduled work against an unavailable write primary. Previously it would have crash-looped instead. The direction favours availability, but it is unstated and untested, and it is the opposite of what the body claims.

Note the 🔴 comment I added on `ready.ts` warns about the *inverse* hazard (readiness being repointed at `/api/ready`) while the hazard that is **already live** — startup pointed at `/api/health` — goes unmentioned. That asymmetry is the tell that I generalised from a sample.

**Remedy, and it is clean:** point those startup probes at `/api/ready`. I checked the thing that would have made this unsafe — `/api/ready` is warm-gated, and `warmup.ts` flips `warm.ready` to `true` when `WARMUP_ENABLED !== 'true'`, with the log line at `src/server/warmup.ts:347` stating that `/api/ready` returning 200 on a cold pod is *correct for non-warming pools (jobs/next/stage/app/previews)*. So a non-warming pool can probe `/api/ready` today with no code change. That is a deployment-config change, tracked separately from this PR.

Reviewers: this is a decision, not a defect in the diff. Either repoint those probes, or accept and document the availability-favouring change for the non-LB deployments.

## Two other things from the same audit

**Fixed in `129d6ac406` — a mutation blind spot I had published a clean matrix for.** Adding `clickhouse` or `searchMetrics` to the soft set — silently making either non-critical for readiness — **survived the full 31-test suite**. Both check functions early-return `true` for a null client, and the test mocked both as null, so neither could be made to fail; a check that cannot fail cannot detect being moved into the soft set. The "not-over-broadened" enumeration covered 3 of the 5 checks that must stay critical. Both are now mocked as working clients with behavioural cases, **and** `softCheckKeysForMode` is exported with its resolved set pinned exactly per mode so the set cannot grow or shrink unnoticed. Both mutants now die, as does a narrowing mutant. The published matrix was accurate for the five mutants it listed — it just did not enumerate set-*widening*.

**Fixed in the same commit — `checkSiteStatus` reported `Overall: HEALTHY` during a write outage.** It derived the summary from `healthy`, which is scoped to readiness and deliberately does not flip for a soft dependency, so it printed `Overall: HEALTHY` three lines above `Database (write): FAILING`. That text is fed to an LLM that drafts customer replies. Now derived from the per-check results, with a test watched to fail on the old expression. This also fixes the same latent misreport for `sysRedis`, soft since June.

**Not a defect here, for the record:** the `preview / deploy` failure on the first push was the intermittent Turbopack 7-char chunk-hash birthday collision, independently diagnosed today in #4084 — not this diff. Also worth flagging to whoever reviews: alerting that keys on the overall health-failure counter will go quiet for this incident class once `healthy` stops flipping, so a compensating signal on the per-check series should land with this.

ZacxDev:
## Correcting myself again: my earlier account of the `preview / deploy` failure was wrong

I said twice that the preview build failure was **intermittent** and that a rebuild was the workaround. Both halves are wrong, and this repo's own diagnosis of the bug says so explicitly — I cited that document without reading the part that contradicted me.

Per `claudedocs/turbopack-chunk-hash-collision-2026-08-18.md` and the comment it left in `next.config.mjs`: the chunk-hash collision is **deterministic for a given module graph** — "a rebuild of the same commit fails again" — and it "moves to a different pair whenever the graph changes at all, which is why bisecting finds a commit but never a responsible file." So it is not a dice roll per build; it is a dice roll per *graph*, fixed once the graph is.

That matches the evidence I should have weighed earlier rather than explaining away: this branch's preview failed **three times out of three completed attempts**, while **nine other PRs' previews succeeded in the same window**. I had read "intermittent" and stopped there; a 3/3-versus-9/9 split is not what intermittent looks like.

**The actual cause is that this branch was 24 commits behind `main` and missing the Next 16.3.1 bump from #4075.** Its graph was still a 16.3.0 graph, and a colliding one. The doc even pre-registers this exact trap — that the 16.3.1 bump "will look like the fix" while only re-rolling which pairs collide.

Fixed by merging current `main` into the branch (`9251103ced`), which brings 16.3.1 and re-rolls the graph. No conflicts; `main`'s commits touch none of the files this PR changes; the three test files plus the shared-mock ratchet are green on the merged tree (53 tests).

**What I am NOT claiming:** that this makes the preview build pass. It re-rolls a lottery that this branch was losing; per the doc, a future graph change can lose it again, and the only lever that attacks the mechanism (`turbopackServerSideNestedAsyncChunking`) is documented as broken on 16.3.0 and deliberately off. If the preview fails again, read the colliding path on the line *above* the `Error:` — not the two hashed debug dumps below it, which is the misreading the doc warns about.

One loose end I could not attribute: the second failing build showed a different signature — 36 × `TypeError: __turbopack_context__.a is not a function` in the PostCSS transform, not the collision. That is the signature the config comment ties to `turbopackServerSideNestedAsyncChunking` being enabled on 16.3.0, yet the flag is `false` on both this branch and `main`, so I cannot explain it and am not going to guess. The third build's pod was garbage-collected before I read it, so its signature is unknown — and worth stating plainly, because my first grep of it returned nothing and I nearly read that empty result as "a different error", when it only meant the pod was gone.

PR #4101, opened 2026-08-18
feat(notices): wire feature notices to audience targeting

ZacxDev:
Phase 2 of the in-product "you have access to a new feature" notice. Phase 1 (#3943) built the `FeatureNotice` primitive + registry and deliberately stopped, leaving instructions in the registry's file header:

> add an optional `audience` field to `FeatureNotice` below and branch on it inside `useFeatureNotice`'s returned `isDismissed`/visibility. It is omitted today because nothing reads it — **an unread field on a definition looks like a gate and gates nothing**.

That last sentence is this PR's acceptance criterion, and it is what the tests are built around.

## What landed

1. **`audience` on `FeatureNotice`** — optional, `{ feature: FeatureFlagKey }`.
2. **`useFeatureNotice` branches on it**, returning `isInAudience`.
3. **A real call site consumes it**: `RemixGalleryExplainer`.

No `id` in `FEATURE_NOTICES` changed. Those strings are live in real users' `dismissedAlerts`, and `notice-registry.test.ts` still pins all nine as hand-written literals.

## The design, and why

**`audience` is a resolved feature-flag key, not a cohort name and not a naming convention.**

"Is this flag worth announcing?" becomes a *lookup* in the registry, so it is reviewed like any other code change. The two alternatives were considered and rejected:

- **Flag metadata carried alongside the flag definitions.** Lets a flag owner self-serve without a deploy, but there is effectively no review gate on that surface, so an unreviewed UI click could publish user-facing copy to everyone. Wrong trade.
- **A convention** ("any flag targeting the early-adopter cohort is announceable"). Zero new mechanism, but it conflates *who can see a feature* with *who should be told about it*, and leaves no way to roll out quietly to a cohort — which is most of what a staged cohort is for.

Accepted cost: a notice requires a deploy. That is the price of the review gate, and it is the right way round.

**Flag evaluation goes through the context-carrying path.** `useFeatureNotice` reads whatever `FeatureFlagsProvider` resolved for this session (`useOptionalFeatureFlags`) rather than performing a keyed evaluation — a keyed evaluation has no context and answers "not in the segment" for every user, uniformly and silently. This PR adds no direct keyed evaluation.

🔴 **Correction (post-review).** An earlier draft of this paragraph called the per-user *toggleable overlay* "the only path on which a segment-scoped flag can match", and `remixGallery` — the one flag this PR actually targets — is the counter-example. The provider hands consumers `{ ...ssrFlags, ...toggleableOverlay }`, and **both halves are session-resolved server-side**; which half carries a given flag is decided solely by whether its definition sets `toggleable: true`.

- **Not toggleable → the SSR `flags` prop.** `getFeatureFlagsAsync` resolves it per request against the user's identity, tier, cohort, host and region, *including* Flipt segment evaluation (`buildFliptContext(user)`), and it is present from the first client frame. `remixGallery` is this kind: `{ availability: ['mod'], fliptKey: 'remix-gallery' }`, no `toggleable`.
- **`toggleable: true` → the `user.getFeatureFlags` overlay.** `computeUserFeatureFlagsOverlay` filters to toggleable keys, so a non-toggleable flag never appears in it at all.

The consequence, stated plainly because it was not: the `if (!flagsReady) return false` guard **cannot change the answer for the only notice that ships**. For a non-toggleable flag the SSR flags already hold the per-user verdict, so that guard can defer a render by a frame and nothing more. It becomes load-bearing the first time a notice names a `toggleable` flag, which is the case it was written for — so it stays, but the shipped notice is not evidence that it works. Both test fixtures are non-toggleable too, so nothing here exercises the overlay path either. All of this now lives in the registry's own doc rather than in a PR description.

**It fails closed, in both directions it can be unsure**: before the toggleable overlay has settled (otherwise the notice announces against the anonymous snapshot and then retracts — the flash the notice machinery exists to prevent), and when there is no flag provider at all. The asymmetry is deliberate: not announcing a feature to someone who has it is a missed nudge; announcing one to someone who does *not* have it is a broken promise.

**`hasSettings` vs `isLoading` is untouched.** `isInAudience` is a third, independent gate — being in an audience is not permission to re-show something the user dismissed, and being out of one is not "dismissed". Asserted both ways.

### New export: `useOptionalFeatureFlags`

`useFeatureFlags` throws outside the provider, which is right for a component that cannot do its job without flags and wrong for shared machinery that merely wants to *ask*. `useFeatureNotice` runs at every notice call site, including the eight that declare no audience, so making it throw would turn "flags are irrelevant here" into a crash.

## The consuming call site

`RemixGalleryExplainer` — the inline explainer at the top of a remix gallery, announcing the `remixGallery` staged rollout. It is exactly the shape this feature is for: a feature a user could not previously see, so the explainer is only ever *news* to someone the flag is on for.

Two honest notes:

- **This does not narrow who sees it today.** Its mount site (`RemixGalleryCard`) already gates on the same flag, so in production this is defence in depth. What it buys is that the gate now travels with the *notice* rather than living in one parent, so a second mount site cannot forget it.
- **It does change one thing**: the explainer now also waits for the per-user flag overlay to resolve before rendering. Previously it rendered as soon as settings resolved. That is the fail-closed behaviour above, and it is intended.

The component was also migrated from its hand-rolled copy of the dismissal logic to `useFeatureNotice` (it was one of three sites phase 1 left on constants-only). Net −40 lines of duplicated optimistic-update/rollback/reconcile code, and the wire payload is unchanged — pinned by the existing call-site test.

## Which suite the gate actually runs

🔴 **CI runs `vitest run --project unit`.** The `component` project (browser-mode, real Chromium) is **not** run by the gate. So the load-bearing assertions are deliberately placed in the `unit` project:

| Suite | Project | Runs in CI | Covers |
|---|---|---|---|
| `useFeatureNotice.audience.test.ts` **(new)** | `unit` (happy-dom) | ✅ | **The load-bearing claim.** Renders the hook and asserts its returned values: a targeted notice is not offered to a non-member, is offered to a member, fails closed while unresolved, and untargeted notices are unchanged. |
| `notice-registry.test.ts` | `unit` | ✅ | `isNoticeAudienceMatched` as returned behaviour — membership, absence, fail-closed, reads-its-own-flag. Plus the unchanged id table. |
| `notice-callsite-composition.test.ts` **(new)** | `unit` (happy-dom) | ✅ | **The call site's composition, rendered.** Mounts the real `RemixGalleryExplainer` through the real hook and the real registry entry and asserts what reached the DOM. This is what kills the destructured-but-unused mutant below. |
| `notice-callsite-coverage.test.ts` | `unit` | ✅ | Extended #3987 ledger: every call site of a notice that declares an `audience` must read `isInAudience`. Population gate — see limits below. |
| `useFeatureNotice.browser.test.tsx` | `component` | ❌ | Same claims through a real browser render. Corroboration, not a gate. |
| `notice-callsite.browser.test.tsx` | `component` | ❌ | Rendered DOM: the explainer is absent for a non-member, present for a member. |

The happy-dom hook-render pattern is not new here — `useDailyBoostReward.test.ts` already does it, which is how a React hook can be asserted inside the node project.

## Red → green and mutation evidence

Every new assertion was watched fail. The mutations below restore exactly the pre-change state, so they are simultaneously the red→green evidence and the mutation check.

**Mutation 1 — make the audience branch inert** (the exact defect phase 1 warned about). Replaced the body of `isNoticeAudienceMatched` with `return true`:

```
Tests  11 failed | 33 passed (44)

FAIL notice-registry.test.ts > isNoticeAudienceMatched > a notice with an audience > does NOT match a user its own flag is off for
AssertionError: expected true to be false // Object.is equality
FAIL useFeatureNotice.audience.test.ts > a targeted notice reaches its audience and nobody else > NOT offered to a user the notice's flag is off for
AssertionError: expected true to be false // Object.is equality
```

11 tests across **both** suites go red. The untargeted-notice cases stay green — by design; they are labelled in-file as invariant guards, not regression coverage.

**Mutation 2 — the seam.** Left the function intact and replaced the hook's use of it with `isInAudience: true`:

```
Tests  5 failed | 39 passed (44)

FAIL useFeatureNotice.audience.test.ts > a targeted notice fails CLOSED while the answer is unknown > not offered when there is no flag provider at all
AssertionError: expected true to be false // Object.is equality
```

Only the hook suite goes red; the pure-function suite stays green, which is the correct signature for a broken seam rather than a broken function. This is what proves the hook actually consumes the gate rather than the two being tested in isolation.

**Mutation 3 — the call site's BINDING.** Removed `isInAudience` from `RemixGalleryExplainer`'s destructure *and* from its render condition:

```
Tests  2 failed | 7 passed (9)

FAIL notice-callsite-coverage.test.ts > feature-notice audience ledger > every call site of a TARGETED notice reads isInAudience
AssertionError: A call site renders a notice that declares an `audience` without reading `isInAudience`
from useFeatureNotice, so it would announce the feature to users who do not have it.
  + [ "components/RemixGallery/RemixGalleryExplainer.tsx (notice: remixGalleryExplainer)" ]
```

Note the doc comment in that file still contained the word `isInAudience` while the code did not, and the gate correctly did **not** count it — the scanner strips comments first.

🔴 **Correction (post-review): this mutation is about the BINDING, not the condition.** The ledger is a token scan, so it trips only when `isInAudience` stops appearing in the file at all. Dropping the *condition* alone leaves the destructure in place, and the token with it. That narrower mutant is the next one.

**Mutation 4 — the call site's CONDITION only** (the surviving mutant this PR was audited for). Deleted `|| !isInAudience` from the render guard and **kept** the destructure at line 38:

```diff
-  if (!currentUser || !hasSettings || isDismissed || !isInAudience) return null;
+  if (!currentUser || !hasSettings || isDismissed) return null;
```

Before `notice-callsite-composition.test.ts` existed this **survived**: the token was still present, the ledger passed, and the whole `unit` project stayed green. The audience was computed and then thrown away, and the only automated signal was an ESLint `no-unused-vars` **warning** — `eslint` exits 0. The two backstops the ledger names did not cover it either: `useFeatureNotice.audience.test.ts` asserts the *hook's* return value, not the call site's composition, and `notice-callsite.browser.test.tsx` is in the `component` project CI does not run.

With the new suite in place it dies, and the other three suites still pass — which is the point: the failure is attributed to the call site, not smeared across the feature.

```
 ✓  unit  notice-registry.test.ts (31 tests)
 ✓  unit  notice-callsite-coverage.test.ts (9 tests)      ← the ledger: still GREEN, as diagnosed
 ❯  unit  notice-callsite-composition.test.ts (9 tests | 5 failed)
 ✓  unit  useFeatureNotice.audience.test.ts (14 tests)

 Test Files  1 failed | 3 passed (4)
      Tests  5 failed | 57 passed (62)

FAIL notice-callsite-composition.test.ts > RemixGalleryExplainer applies its audience to what it renders
   > renders NOTHING for a user the notice's flag is OFF for
AssertionError: the explainer announced the remix gallery to a user the `remixGallery` flag is OFF for.
The hook computed `isInAudience`, and the call site did not APPLY it — AND `!isInAudience` back into the
render condition in RemixGalleryExplainer.tsx.: expected true to be false // Object.is equality
```

Restored, and green again:

```
 Test Files  4 passed (4)
      Tests  63 passed (63)
```

All four mutations restored.

Fixtures in the new suites are hand-built and pairwise distinct, and deliberately name flags **other than** `remixGallery` — the one constant the registry names — so a mutant that hardcodes the production value cannot survive. The two targeted fixtures name different flags and are given opposite answers from one flag map, so "reads its own audience" is separable from "some flag was on".

## Verification

- **`pnpm exec vitest run --project unit` (the gate): 1166 files / 18345 tests, 0 failures.** The four suites above are confirmed to have *executed* in that run — `notice-callsite-composition.test.ts (9 tests)`, `useFeatureNotice.audience.test.ts (14 tests)`, `notice-callsite-coverage.test.ts (9 tests)`, `notice-registry.test.ts (31 tests)` — rather than inferred from the absence of failures.
- **`pnpm typecheck`: 0 errors.**

## Known limits

- **The ledger gate is structural, and weaker than it first looks.** It cannot tell `if (!isInAudience)` from `if (isInAudience)` — and, as mutation 4 shows, it cannot tell a *read* from a *binding* either: a destructure alone satisfies it. It exists only to keep the *population* of the behavioural tests complete. The behavioural claim is asserted twice in the `unit` project — against the hook's return value (`useFeatureNotice.audience.test.ts`) and against the rendered call site (`notice-callsite-composition.test.ts`) — and both are mutation-checked. Stated as such in both files.
- 🔴 **The `component` (browser) suites could not be run locally, so the browser-mode edits in this PR are unexecuted.** The browser project fails to launch on this machine for an environment reason unrelated to this change: a clean control on an untouched browser test file (`src/hooks/useScrollMargin.browser.test.tsx`) fails identically, reporting `Tests no tests`. So those edits are **verified by review and typecheck only**. They are not the gate and every gate suite passes, but it is a real gap — and it is exactly why the load-bearing assertions were put in `unit` rather than left in browser mode.
- Three browser mock factories needed `useOptionalFeatureFlags` added. An omitted export fails those files at **collection**, which reports as "no tests" rather than as a failure — worth knowing since I could not execute them here.

### DIFF
--- a/src/components/Alerts/__tests__/feature-notice.characterization.browser.test.tsx
+++ b/src/components/Alerts/__tests__/feature-notice.characterization.browser.test.tsx
@@ -125,6 +125,11 @@ vi.mock('~/hooks/useCurrentUser', () => ({
 
 vi.mock('~/providers/FeatureFlagsProvider', () => ({
   useFeatureFlags: () => mocks.state.features,
+  // Required by `useFeatureNotice` for `audience`. None of the notices
+  // characterized here declares one, so this changes nothing about what they
+  // render — but omitting an export a consumer imports fails the file at
+  // COLLECTION, which reads as "no tests" rather than as a failure.
+  useOptionalFeatureFlags: () => mocks.state.features,
   useFeatureFlagsReady: () => mocks.state.flagsReady,
 }));
 
--- a/src/components/Alerts/__tests__/notice-callsite-composition.test.ts
+++ b/src/components/Alerts/__tests__/notice-callsite-composition.test.ts
@@ -1,0 +1,243 @@
+// @vitest-environment happy-dom
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import * as React from 'react';
+import type { act as actType } from 'react-dom/test-utils';
+import { createRoot } from 'react-dom/client';
+import type * as Trpc from '~/utils/trpc';
+
+const act = (React as unknown as { act: typeof actType }).act;
+
+// Tells React this is an act-aware environment, so the renders below don't each
+// log "not configured to support act(...)". Purely output noise.
+(globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
+
+/**
+ * 🔴 THE CALL SITE'S COMPOSITION, RENDERED — in the `unit` project, which is the
+ * only one CI runs.
+ *
+ * WHY THIS FILE EXISTS. Three suites already cover audience targeting and none
+ * of them could see the defect this one is for:
+ *
+ * - `useFeatureNotice.audience.test.ts` asserts the HOOK'S RETURN VALUE. A hook
+ *   can only offer an answer; it says nothing about whether a call site applies
+ *   it.
+ * - `notice-callsite-coverage.test.ts` greps the call site for the token
+ *   `isInAudience`. That is a SPELLED guard — it is satisfied by the
+ *   destructure alone, so deleting `&& !isInAudience` from the render condition
+ *   while leaving `isInAudience` bound at line 38 keeps it green. The audience
+ *   is then computed and thrown away, and every notice ships to every user who
+ *   reaches the component.
+ * - `notice-callsite.browser.test.tsx` does assert the rendered DOM, but it is
+ *   in the `component` project, which CI does NOT run. It is corroboration.
+ *
+ * The verified surviving mutant was exactly that: destructure kept, `&&
+ * !isInAudience` removed from `RemixGalleryExplainer`'s render guard, whole
+ * `unit` project green. This file renders the REAL component through the REAL
+ * hook and the REAL registry entry and asserts what reached the DOM, so the
+ * behaviour — not the word — is what is pinned.
+ *
+ * WHY `.test.ts` AND NOT `.test.tsx`: the `unit` project's include is
+ * `src/**\/*.test.ts`, so a `.tsx` file is silently not collected. Hence
+ * `React.createElement` rather than JSX. `happy-dom` + `react-dom/client` is the
+ * pattern this repo already uses to render in the node project (see
+ * `useDailyBoostReward.test.ts` and `useFeatureNotice.audience.test.ts`).
+ *
+ * FIXTURE HONESTY: `remixGallery` here is NOT a free choice — it is the flag the
+ * registry names for this notice, and asserting against the real registry entry
+ * is the point of a composition test. The "unrelated flag" case below uses a
+ * different flag so that "reads its own audience" stays separable from "some
+ * flag was on".
+ */
+
+let dismissedAlerts: string[] | undefined;
+let settingsResolved = true;
+let flags: Record<string, boolean> | null;
+let flagsReady: boolean;
+let currentUser: { id: number } | null;
+const dismissMutate = vi.fn();
+
+vi.mock('~/utils/trpc', async (importOriginal) => ({
+  ...(await importOriginal<typeof Trpc>()),
+  trpc: {
+    useUtils: () => ({
+      user: {
+        getSettings: {
+          cancel: vi.fn(),
+          getData: vi.fn(),
+          setData: vi.fn(),
+          invalidate: vi.fn(),
+        },
+      },
+    }),
+    user: {
+      getSettings: {
+        // `enabled` honoured, not ignored: the hook disables this query without
+        // a signed-in user, so an always-answering mock would hand the
+        // signed-out case a `hasSettings` production never gives it.
+        useQuery: (_input?: unknown, opts?: { enabled?: boolean }) => ({
+          data: settingsResolved && (opts?.enabled ?? true) ? { dismissedAlerts } : undefined,
+          isLoading: false,
+        }),
+      },
+      dismissAlert: { useMutation: () => ({ mutate: dismissMutate }) },
+    },
+  },
+}));
+
+vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => currentUser }));
+
+vi.mock('~/providers/FeatureFlagsProvider', () => ({
+  useOptionalFeatureFlags: () => flags,
+  useFeatureFlagsReady: () => flagsReady,
+}));
+
+const { RemixGalleryExplainer } = await import('~/components/RemixGallery/RemixGalleryExplainer');
+const { MantineProvider } = await import('@mantine/core');
+
+/**
+ * A fragment of the explainer's own copy. Hand-typed rather than imported, so a
+ * rewrite of the component's markup cannot make this agree with itself.
+ */
+const NOTICE_COPY = "What's a remix gallery?";
+
+/** Marks the tree as having mounted at all. */
+const SENTINEL_ID = 'composition-sentinel';
+
+type Rendered = {
+  /** Dismiss affordances the notice rendered. 1 when shown, 0 when withheld. */
+  dismissControls: number;
+  /** Whether the notice's own copy reached the DOM. */
+  showsCopy: boolean;
+  /**
+   * Whether the surrounding tree mounted. Without this, "the notice is absent"
+   * is indistinguishable from "the render threw and nothing exists" — an
+   * assertion that would pass for the wrong reason forever.
+   */
+  mounted: boolean;
+};
+
+function renderCallSite(): Rendered {
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+  const root = createRoot(container);
+  act(() =>
+    root.render(
+      React.createElement(
+        MantineProvider,
+        null,
+        React.createElement(RemixGalleryExplainer),
+        React.createElement('span', { 'data-testid': SENTINEL_ID }, 'mounted')
+      )
+    )
+  );
+  const result: Rendered = {
+    dismissControls: container.querySelectorAll('[aria-label="Dismiss"]').length,
+    showsCopy: (container.textContent ?? '').includes(NOTICE_COPY),
+    mounted: !!container.querySelector(`[data-testid="${SENTINEL_ID}"]`),
+  };
+  act(() => root.unmount());
+  container.remove();
+  return result;
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  dismissedAlerts = [];
+  settingsResolved = true;
+  currentUser = { id: 91 };
+  flags = { remixGallery: true };
+  flagsReady = true;
+});
+
+describe('RemixGalleryExplainer applies its audience to what it renders', () => {
+  test('renders for a user the notice’s flag is ON for', () => {
+    // 🔴 The positive control, and it comes first deliberately: every assertion
+    // below is "the notice is absent", which a component that renders nothing
+    // ever would also satisfy.
+    const { dismissControls, showsCopy, mounted } = renderCallSite();
+    expect(mounted).toBe(true);
+    expect(showsCopy).toBe(true);
+    expect(dismissControls).toBe(1);
+  });
+
+  test('renders NOTHING for a user the notice’s flag is OFF for', () => {
+    // 🔴 THE MUTANT-KILLING CASE. Deleting `&& !isInAudience` from the render
+    // guard — while leaving `isInAudience` destructured, which is what keeps the
+    // token-scanning ledger green — makes this render the full notice.
+    flags = { remixGallery: false };
+    const { dismissControls, showsCopy, mounted } = renderCallSite();
+    expect(mounted, 'the tree did not mount at all, so an absent notice proves nothing here').toBe(
+      true
+    );
+    expect(
+      showsCopy,
+      'the explainer announced the remix gallery to a user the `remixGallery` flag is OFF for. ' +
+        'The hook computed `isInAudience`, and the call site did not APPLY it — AND ' +
+        '`!isInAudience` back into the render condition in RemixGalleryExplainer.tsx.'
+    ).toBe(false);
+    expect(dismissControls).toBe(0);
+  });
+
+  test('a flag absent from the overlay is not an audience', () => {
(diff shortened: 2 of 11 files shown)


ZacxDev:
Three audit findings addressed. No `id` in `FEATURE_NOTICES` touched; `useFeatureNotice`'s `hasSettings`/`isLoading` untouched.

---

## 1. The surviving mutant — a spelled guard at the call site

**The finding.** `notice-callsite-coverage.test.ts` tests `/\bisInAudience\b/` against the call site's source. That token is supplied by the **destructure** on line 38, so deleting only `&& !isInAudience` from the render guard on line 46 — keeping the destructure — left the ledger green and the **whole `unit` project green**. The audience was computed and thrown away; the only automated signal was an ESLint `no-unused-vars` *warning*, and `eslint` exits 0. Neither backstop the file names covered it: `useFeatureNotice.audience.test.ts` asserts the *hook's* return value, not the call site's composition, and `notice-callsite.browser.test.tsx` is in the `component` project CI does not run.

**The fix.** A new `unit`-project suite that renders the real component — `src/components/Alerts/__tests__/notice-callsite-composition.test.ts` (9 tests). It mounts `RemixGalleryExplainer` through the real hook and the real registry entry under `happy-dom` + `react-dom/client` (the pattern `useDailyBoostReward.test.ts` and this PR's own hook suite already use) and asserts **what reached the DOM**, not what appears in the source. `.test.ts` rather than `.test.tsx` because the `unit` project's include is `src/**/*.test.ts` — a `.tsx` file is silently not collected — hence `React.createElement`.

It carries its own positive control (flag ON renders, asserted first) and a mount sentinel, so "the notice is absent" can never pass because the render threw.

**Mutation evidence — the exact narrow mutant, red → green.**

```diff
-  if (!currentUser || !hasSettings || isDismissed || !isInAudience) return null;
+  if (!currentUser || !hasSettings || isDismissed) return null;
```

Destructure at line 38 left in place. RED:

```
 ✓  unit  notice-registry.test.ts (31 tests)
 ✓  unit  notice-callsite-coverage.test.ts (9 tests)      ← the ledger: still GREEN, as diagnosed
 ❯  unit  notice-callsite-composition.test.ts (9 tests | 5 failed)
 ✓  unit  useFeatureNotice.audience.test.ts (14 tests)

 Test Files  1 failed | 3 passed (4)
      Tests  5 failed | 57 passed (62)

FAIL notice-callsite-composition.test.ts > RemixGalleryExplainer applies its audience to what it renders
   > renders NOTHING for a user the notice's flag is OFF for
AssertionError: the explainer announced the remix gallery to a user the `remixGallery` flag is OFF for.
The hook computed `isInAudience`, and the call site did not APPLY it — AND `!isInAudience` back into the
render condition in RemixGalleryExplainer.tsx.: expected true to be false // Object.is equality
```

Mutant restored — GREEN:

```
 Test Files  4 passed (4)
      Tests  63 passed (63)
```

The ledger staying green under the mutant is the finding reproduced, and the failure landing only in the new file is what attributes it to the call site rather than smearing it across the feature.

The PR description's mutation-3 claim is corrected: dropping the call site's **binding** trips the ledger; dropping its **condition** does not. That is now written up as a separate mutation 4 with the evidence above.

---

## 2. The registry documented a path this flag does not take

**The finding.** `NoticeAudience`'s doc said the audience value "is read from the per-user flag overlay… the only path on which a segment-scoped flag can match". `remixGallery` is `{ availability: ['mod'], fliptKey: 'remix-gallery' }` — **no `toggleable: true`** — and `computeUserFeatureFlagsOverlay` filters to toggleable keys, so it never appears in that overlay. It arrives on the SSR `flags` prop instead, already resolved.

**Verified while fixing:** both halves the provider merges (`{ ...ssrFlags, ...toggleableOverlay }`) are session-resolved server-side. `getFeatureFlags` builds `buildFliptContext(ctx.user)` and keys its cache on user id / moderator / tier / early-adopter / permissions / host / region, so Flipt **segment evaluation happens on the SSR path too**. The old claim was wrong in both directions: wrong about which path this flag takes, and wrong that the overlay is the only path that can match a segment.

**The fix.** The registry doc now states which flags take which path, and names the consequence the audit identified: for a non-toggleable flag the `if (!flagsReady) return false` guard at `notice-registry.ts:166` can **defer a render by a frame but cannot change its answer**, so it is inert for the only notice that ships. It is kept — it becomes load-bearing the first time a notice names a `toggleable` flag, which is the case it was written for — but the doc now says the shipped notice is not evidence that it works. Both test fixtures (`imageCardInfoButton`, `appReviewPage`) are non-toggleable as well, so nothing exercises the overlay path; that is stated rather than papered over.

---

## 3. A JSDoc claim false in two reachable states, and the hook made honest for an anon caller

**The finding.** `@param flagsReady Whether that overlay is the user's, not the anon snapshot` is false wherever `ready = !session.data || isSuccess || isError` is true for a reason other than success: on the **`isError`** arm (`retry: 0`, `userFeatures` falls back to `{}` → host SSR flags only) and when **logged out** (`!session.data` → the anon snapshot). For a future *toggleable* audience flag that fails open in exactly the case the parameter exists to close.

**The fix.**

- The JSDoc now states what `flagsReady` actually means — "the toggleable overlay query has SETTLED: success, error, or no session at all" — and enumerates both states where it is true while the overlay is not the user's.
- `useFeatureNotice` now ANDs in `!!currentUser`, so the hook cannot answer "in the audience" for someone with no session. `isNoticeAudienceMatched` genuinely cannot close that itself: it is handed `flagsReady`, which is `true` by construction for a logged-out visitor, so from inside the pure function the anon snapshot is indistinguishable from a resolved per-user one.

**This changes no behaviour anywhere.** `hasSettings` is *already* `false` for a signed-out visitor — `queryEnabled = enabled && !!currentUser` disables the settings query — so any call site composing as documented (`hasSettings && !isDismissed && isInAudience`) already withheld every notice there. `RemixGalleryExplainer` additionally short-circuits on `!currentUser`. Both facts are asserted rather than assumed.

Mutation-checked too: reverting the `!!currentUser &&` turns the new case red on its own assertion (`expected true to be false` at `isInAudience`) with the other 13 in that file still green, and the accompanying `hasSettings` assertion still passing — so the new claim is separable from the one that already held.

The audience suite's tRPC mock now honours `enabled` rather than always answering, so it models the disabled-query shape production actually produces for a visitor.

---

## Verification

- **`pnpm exec vitest run --project unit` (the gate): 1166 files / 18345 tests, 0 failures.** All four suites confirmed to have *executed* in that run — `notice-callsite-composition.test.ts (9 tests)`, `useFeatureNotice.audience.test.ts (14 tests)`, `notice-callsite-coverage.test.ts (9 tests)`, `notice-registry.test.ts (31 tests)` — rather than inferred from an absence of failures.
- **`pnpm typecheck`: 0 errors.**
- `prettier --check` and `eslint` clean on all four changed files.

**Not verified, unchanged from the original PR:** the `component` (browser) project still cannot launch on this machine, and CI does not run it. `notice-callsite.browser.test.tsx` was not touched by these fixes and its absence is neither a pass nor a failure. That project's inability to run is precisely why fix 1 put the new assertions in `unit`.

PR #4108, opened 2026-08-19
fix(feedback): trim the /apps search term before storing it, and keep its case

ZacxDev:
fix(feedback): trim the /apps search term before storing it, and keep its case

A STORAGE-SHAPE decision taken while `Feedback` still holds no `apps-marketplace`
rows, because it stops being free the moment it holds some.

THE DEFECT. `resolveAppsStoreFilters` trims the `?query=` value only to TEST
emptiness and then keeps the original, so `'  upscale  '` reaches
`buildAppsStoreFeedbackContext` with its whitespace intact and is written to the
`Feedback.context` JSONB as a value distinct from `'upscale'`. Every later
aggregate over that column then splits one term across two spellings, and no
later change can un-mix rows that already carry both shapes — a normalisation
added after the fact only adds a third population.

WHY WHITESPACE IS DROPPABLE. It is provably not signal. The store's own matcher
is `debouncedSearch.trim().toLowerCase()` (AppListingsMarketplaceBody.tsx:331),
so a leading or trailing space cannot change which listings the reporter was
looking at, and therefore cannot be the behaviour any report describes. There is
nothing to lose by dropping it.

WHY CASE IS KEPT. Case survives that same argument — the matcher lowercases too
— but the two are not symmetric at the storage boundary. Case is worth reading
('BitDex' names the app the reporter typed), and `lower()` recovers grouping at
read time, whereas folded case cannot be recovered at all. So normalise away
only what can never be wanted back. The "preserves the case the reporter typed"
control is what makes that a decision rather than an accident: it fails the
moment someone "finishes the job" with a `.toLowerCase()`.

ORDER. trim -> clip -> surrogate guard. Trimming first applies the 200-unit
bound to the real term rather than to padding about to be discarded (a padded
term yields 200 characters, not 195). `trim` removes only whitespace, so it can
never split a surrogate pair or manufacture the lone high surrogate the existing
guard exists for.

NOT CHANGED, deliberately: the term is still captured verbatim otherwise. On a
store surface "I searched X and got nothing" is the most actionable fact a
one-line report can carry, the catalog is app listings rather than model search,
and the row already holds a 2000-char free-text message next to the same userId
— so reducing the term to a length or a boolean would cost the whole point of
the field for no meaningful privacy gain.

TESTS. Seven new cases, all watched to FAIL with the trim removed (6 red, each
with its own assertion; the 13 pre-existing cases stay green):
  expected '  upscale  ' to be 'upscale'
  expected '\t\n upscale \n\t' to be 'upscale'
  expected '  anime upscale  ' to be 'anime upscale'
  expected '  BitDex  ' to be 'BitDex'
  expected '   ' to be ''
  expected '     qqq…' to be 'qqq…'
The seventh pins the premise — that the resolver really does hand this builder
an untrimmed term — so the trim cannot become silently redundant if that ever
changes.

Mutation-checked, each mutant killed by its OWN control and no other:
  - `.trim()` -> `.replace(/\s/g,'')`  => 'animeupscale' (interior whitespace)
  - `.trim().toLowerCase()`            => 'bitdex' (case control)
  - clip before trim                   => "length 200 but got 195" (order)

VERIFICATION SCOPE. 974 tests green across the Apps suites and
feedback.schema.test.ts. The feedback ROUTER and SERVICE suites could not run in
this worktree — they fail to import on a missing generated Prisma client

### DIFF
--- a/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
+++ b/src/components/Apps/__tests__/appsStoreFeedbackContext.test.ts
@@ -137,8 +137,8 @@ describe('buildAppsStoreFeedbackContext', () => {
      *
      * The set is `String.prototype.trim`'s own: WhiteSpace + LineTerminator, which includes
      * NBSP (U+00A0), BOM (U+FEFF), CR, VT/FF, LS/PS (U+2028/9) and ideographic space
-     * (U+3000). \u26A0\uFE0F Nothing here EXERCISES the store's matcher, so this table does not
-     * enforce that the two stay in step \u2014 if `AppListingsMarketplaceBody.tsx:331` stopped
+     * (U+3000). NOTE: nothing here EXERCISES the store's matcher, so this table does not
+     * enforce that the two stay in step — if `AppListingsMarketplaceBody.tsx:331` stopped
      * using `trim()`, no assertion in this file would go red. It pins our side only.
      */
     const TRIMMED_CHARS: [string, string][] = [
@@ -168,8 +168,8 @@ describe('buildAppsStoreFeedbackContext', () => {
     });
 
     /**
-     * \uD83D\uDD34 THE SAME BLIND SPOT, ONE POSITION OVER. The table above places each exotic
-     * character only in the PADDING, so the term's INTERIOR is ASCII in every fixture \u2014
+     * 🔴 THE SAME BLIND SPOT, ONE POSITION OVER. The table above places each exotic
+     * character only in the PADDING, so the term's INTERIOR is ASCII in every fixture —
      * and a fixture whose interior cannot contain the character cannot observe an
      * implementation that rewrites it there. Four mutants survived a fully green suite on
      * exactly that gap, including `replace(/\uFEFF/g, '')` (a global BOM strip, a very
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Corrections to this PR's own description

An adversarial audit checked the claims in the commit message / PR body and found two that are wrong. Correcting them here rather than editing the body, since the original may already have been read.

**1. "each mutant killed by its OWN control and no other" — false for one of the three.**

The `replace(/\s/g, '')` mutant dies to *two* tests, not one: the interior-whitespace control **and** the pre-existing `leaves an emoji inside the bound intact` case (which uses `'upscale 😀'`, so stripping all whitespace breaks it too). The claim holds for the other two mutants — `.toLowerCase()` and clip-before-trim each die to exactly one test, and the clip-before-trim message reproduces verbatim.

**2. "974 tests green" — not reproducible; the correct figure is 987.**

`--project unit src/components/Apps` = 953, plus `feedback.schema.test.ts` (34) = **987**. The 13-test gap is exactly the number of pre-existing cases in the changed file, i.e. the original number counted only the 7 new ones. Green either way — the count was simply wrong.

## And two real gaps the audit found, now fixed in `5add909cff`

The audit ran a 17-mutant battery; **2 survived a fully green suite**. Both were fixture blind spots rather than code defects — the implementation was correct on both axes, but the tests were pinning less than their names claimed:

- **Interior runs.** `keeps whitespace INSIDE the term` used a *single* interior space, which cannot distinguish `trim()` from `trim()` + `replace(/\s+/g, ' ')`. Now double-spaced.
- **The whitespace class.** Every padding fixture used only space/tab/newline, so a hand-rolled `replace(/^[ \t\n]+|[ \t\n]+$/g, '')` passed all of them while leaving NBSP-padded terms untrimmed — reintroducing the exact fragmentation this PR exists to prevent. NBSP is the common real case (pasting a term out of a rendered page). Added a table over the set `String.prototype.trim` actually strips: NBSP, BOM, ideographic space, CR, VT, FF, LS.

Both mutants were re-run against the strengthened suite and now die, each for its own assertion (1 red and 7 red respectively). 27 tests green in the file, 994 across the same scope — the audit's 987 plus exactly these 7.

Claims that **did** check out: 6 red with the trim removed, each with its own assertion; 13 pre-existing + 7 new = 20 all green; the matcher at `AppListingsMarketplaceBody.tsx:331`; and nothing under `src/server/` importing the builder. "Typecheck not run" is now stale rather than wrong — `tekton / typecheck` has since passed.

PR #4109, opened 2026-08-19
fix(hooks): ask about svelte-kit sync only when a dev server is running

JustMaier:
The `svelte-kit sync` rule in `check-writable.mjs` is a good guard aimed slightly too wide, and it has been prompting on every sync during Creator Studio work.

**What it protects.** Sync rewrites ~690 files under `.svelte-kit/` while the Vite dev server watches them — the collision that froze an editor for a full day. That hazard is real and this PR keeps it.

**Why it fires when it can't be right.** Two things:

1. **With no dev server running there is nothing to collide with**, but the rule matched on the command alone.
2. **The legitimate case is not rare.** `$types` goes stale whenever a `load` function's *return shape* changes — not only when the route tree changes, which is what the reason text says. Adding three fields to a load is enough. So it recurs through ordinary work, and a guard that asks when it cannot be right is one people learn to click through.

**The change.** The rule now also requires the dev-server daemon (`127.0.0.1:9444/sessions`) to report a running session. It **fails closed**: unreachable, slow, or unparseable answers ask exactly as before, so the only case that skips the prompt is a positive "zero sessions". The reason text now says a dev server is running, so the prompt distinguishes the two states.

**A false positive the new selftest rows caught:** `check\b` matches before a hyphen, so `pnpm run check-*` scripts tripped a rule that is about `svelte-kit sync`. Tightened to `check(?![\w:-])`.

**Verification**

- `node .claude/hooks/check-writable.selftest.mjs` — all green, including 7 new rows for the matcher. They test the pure matcher, so they need no daemon and pass offline.
- End to end with no dev server up: `svelte-kit sync` → exit 0, no prompt.
- `npx prettier --plugin=prettier-plugin-svelte --write x.svelte` → still **blocked**, exit 2. The other guard in this file is untouched.

Nothing here weakens the block list; only the `ask` for sync is narrowed.

### DIFF
--- a/.claude/hooks/check-writable.mjs
+++ b/.claude/hooks/check-writable.mjs
@@ -9,6 +9,7 @@
  */
 
 import { stdin } from 'process';
+import { execFileSync } from 'child_process';
 
 // `prettier --write <targets>`: the incident was a repo-WIDE rewrite, not formatting a directory
 // this change owns. Read the targets instead of matching `--write`, so a scoped path just runs.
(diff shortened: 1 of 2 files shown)


JustMaier:
Closing: the premise is gone from main.

This narrows a guard so `svelte-kit sync` only prompts when a dev server is running. Main no longer guards that command at all — `.claude/hooks/check-writable.logic.mjs:381` states it deliberately: '`svelte-kit sync` alone is cheap (121 generated files) and is deliberately NOT guarded.' Main also moved every decision out of `check-writable.mjs` into `check-writable.logic.mjs`, which is why this now conflicts — it edits a file that is wiring only.

There is nothing to rebase onto: the thing it improves no longer exists.

Closed on Justin's decision, 2026-09-21.

PR #4115, opened 2026-08-19
fix(apps): tell the RECIPIENT an ownership offer is dead BEFORE they accept it

ZacxDev:
> **What this is, stated up front: CHEAP GAP-CLOSURE AHEAD OF TRAFFIC, not a live bug fix.**
> There are **zero rows in the transfers table in production — not zero pending, zero ever** —
> and the blocked state this PR surfaces is **not reachable through the product today** (see
> *Reachability*, below, which corrects an earlier claim in this description). What it closes
> is a gap #3935 recorded in a code comment, on the cheap, while that PR's template is fresh
> and before the feature gets traffic.

The owner half shipped in #3935. The recipient half was recorded there as an explicitly OPEN
gap, in a comment on `AppCollaboratorsPanelView`: *"the RECIPIENT's surfaces carry no
connect-client awareness at all, so an addressee still sees a normal-looking offer and meets
the refusal only on clicking accept"*. This closes it.

A connect-linked off-site listing can never have its ownership transferred.
`initiateTransfer` refuses it, and `acceptTransfer` re-asserts the same predicate
IN-TRANSACTION. So an offer in `/apps/invites` can in principle be live, in date, addressed
to you, indistinguishable from any other, and guaranteed to fail — and the inbox could not
know, because `listMyPendingTransfers` did not read the column, so the fact could not cross
the wire at all.

## The fix

- `listMyPendingTransfers` selects `connectClientId` off the joined listing and derives
  `acceptBlockedReason: string | null` from it, using the SAME
  `refusesTransferForConnectClient` and the SAME `CONNECT_CLIENT_TRANSFER_REFUSAL` the two
  server gates use. One rule, now four callers.
- The View renders that reason as a warning Alert on the card, disables Accept, and leaves
  Decline enabled and wired.
- Both server gates are UNCHANGED and remain the enforcement. This is an addition in front
  of them, never a replacement, and it runs on data the client could lie to itself about.

🔴 **A DERIVED REASON, NOT THE RAW COLUMN** — deliberately diverging from #3935's owner-side
shape. That change carried `connectClientId` itself, arguing it is public via the approved
listing-detail read. The argument does not carry over: a pending offeree holds NO role on
the listing, and this read imposes no status gate, so an offer can sit on a draft/pending
listing whose client id the public read never exposes. Shipping the column would be a NEW
disclosure on exactly those rows, to the least privileged reader in the feature, for a value
they cannot act on. They can only read the sentence, so the sentence is what crosses.

🔴 **ACCEPT DISABLED, DECLINE ENABLED.** `cancelTransfer` — which is how a recipient declines
— carries no connect-client gate and admits either party, so getting out of a dead offer is
exactly the action that must still work. Disabling both would strand them with a card they
can neither accept nor clear. Blocked, not hidden: suppressing the row would make the
`initiateTransfer` notification land on a page rendering nothing, and would leave the owner's
tab (which still shows the offer) disagreeing with this one about whether it exists.

🔴 **THE DISCLOSURE PANEL IS KEPT ON A BLOCKED CARD**, unlike the owner-side equivalent #3935
suppressed. Operator decision; provenance and reasoning are in a separate comment on this PR.
It is a load-bearing identity marker — the existing pair test names it as one of three things
stopping an irreversible transfer from reading like a reversible seat invite — and unlike the
owner's suppressed copy, which described MECHANICS of a transfer that can never start, it
describes what the offer WAS FOR, which is what the recipient needs in order to decide to
decline.

🔴 **ONE PREDICATE FOR THE TWO BRANCHES, AND IT FAILS OPEN.** The banner and the Accept
button read `acceptBlockedReason` through a single `blocked` test. They previously used
`? …` and `!= null` respectively, which agree on today's two values and diverge on `''` —
Accept disabled with no explanation rendered. Pinned behaviourally by an arm that feeds the
empty string on the payload, not by a comment.

The test is the **loose** `!= null`, deliberately. Strict `!== null` scores an **absent**
field as blocked, and the bundle and the tRPC API are served by pools that **roll
independently** — so every deploy passes through a window where this component runs against
an API that predates the field. Under strict, that window refuses *every* pending offer with
an empty explanation; and since the banner is otherwise unreachable, it would be the only
time a real user ever saw it. Failing open is the right direction: the server refuses an
unacceptable transfer regardless, so a missed banner costs one honest error toast while a
false banner blocks a transfer that would have succeeded. Pinned by an arm that deletes the
field from the row, and by a mutant that re-tightens the comparison.

## Reachability — CORRECTED

An earlier version of this description (and of the issue) said the blocked state arises
because **a revision approve can link an OAuth client while an offer sits open**. That was
inherited prose and it is **not substantiated by the code**. An enumeration of every writer of
`AppListing.connectClientId` across `src/`, `packages/` and `apps/`:

- the only non-null write is `submitExternalListing`'s `create` of a **brand-new row**. What
  makes it mandatory there is the **zod input** (`submitExternalListingSchema`), **not the
  database** — the Prisma column is `connectClientId String?`, nullable with no default.
  (An earlier version of this description said "the schema makes the field required", which
  reads as `schema.prisma` and is false; it also points a reader the wrong way, per the next
  paragraph.) Either way: a listing is **born** linked;
- `beginListingRevision` copies parent → shadow and the approve merge copies shadow → parent;
  in between, `buildListingPatchData` never assigns the column and `updateListingPatchSchema`
  has no such key, so the round trip returns the value it started with;
- no raw SQL, no nested relation write, no spread writes it (each checked; zero hits).

So a listing cannot **acquire** a link, and `initiateTransfer` refuses to open an offer on one
born with it. **The blocked state is not producible through the product as it stands** — it
needs a direct DB write, a migration, or a future **link** flow.

### …but there is a fifth writer, and it lives in the schema

A later review found one the enumeration above **structurally cannot see**. The relation is
declared `onDelete: SetNull`, so deleting the `OauthClient` row issues
`UPDATE app_listings SET connect_client_id = NULL` on every listing referencing it — and
`oauth-client.router::delete` performs exactly that delete with **no referencing-listing
check** (its only guard, `rejectAppBlockClient`, covers AppBlock clients, which these are
not). That is a **product-reachable, owner-initiated unlink that exists today**, and no grep
over `data:` payloads or raw SQL would ever find it, because the writer is a **referential
action in the schema**, not a call site.

It does **not** change the conclusion: a `SetNull` only moves the column non-null → null, so
it can only ever *unblock*. Two consequences worth recording:

- "it needs … a future **unlink**/link flow" was wrong — an unlink route already exists.
- `acceptBlockedReason` can go **stale in the safe direction**: an offer rendered blocked
  whose client was deleted since the read would now succeed on accept. A stale block is a
  delay; a stale allow would be a lie, and this cascade cannot produce one.

*(Remaining caveat: the enumeration is otherwise textual plus call-site based, so it would
also miss a fully dynamic key or a runtime-assembled raw `UPDATE`. Neither construct exists.)*

### 🔴 A premise this PR inherited is refuted — flagging, not fixing

`CONNECT_CLIENT_TRANSFER_REFUSAL`'s rationale in `src/shared/constants/app-transfer.constants.ts`
says *"there is no unlink path"*, and that is the stated reason #3935 dropped
"Unlink the OAuth client first" from the user-facing string. The fifth writer above disproves
it: an owner **can** unlink, by deleting the client.

**No user-facing copy is changed in this PR.** Whether the owner's message should now name
that route is a product decision, and it should be made deliberately rather than inherited
from a premise that turned out to be false. What is still true — and is why the recipient's
copy stands either way — is that the **recipient** cannot unlink: they own neither the listing
nor the client, so a remedy addressed to them remains a dead end. My own two comments that
repeated the false premise are corrected; the constant's is left in place and flagged here.

That is a reason to ship this as gap-closure, not a reason to skip it: the accept-time
re-assert is correct defence-in-depth either way, and `offsite-listing.service.ts` already
anticipates a link flow in two (now-stale) comments at `:2273` and `:2573`. When one lands,
this field is already the channel that tells the recipient — instead of a second
learn-it-by-clicking defect to find later. Those two inherited comments are deliberately left
alone here and flagged as a follow-up.

## Two prerequisites, without which the new tests would have been vacuous

**1. The container cast.** `AppTransferOffers.tsx` cast `as IncomingTransferRow[]` against a
hand-written structural duplicate. A cast to a wider type is legal, so a field the service
never sends compiled clean at exactly the hop where the drift is invisible at runtime too —
the same hole #3935 fixed on `edit.tsx` (F3). `IncomingTransferRow` is now
`Omit<IncomingTransferView, 'expiresAt'|'createdAt'>` plus the two widened date fields, and
the cast is gone.

Measured both ways: deleting `acceptBlockedReason` from the service is now **15 `tsc` errors**
(TS2339 at the View, plus 9 in the new browser suite once its fixture was pinned too), and
compiled with **zero** type errors under the old duplicate-plus-cast shape.

The dates are widened rather than inherited because what arrives depends on the union
transformer. Probed rather than assumed: `transfersQuery.data[0].expiresAt` types as `Date` at
the client (assigning it to `Date` compiles, to `string` is TS2322), so the widening is a
superset that costs nothing and forced no change of shape.

**2. The inbox test fake.** It filtered on `where` and returned the WHOLE fixture row, so a
`select` was unobservable and the mutant deleting `connectClientId: true` survived the suite.
It now projects recursively through `select` — an unselected column is ABSENT, the way Prisma
returns it — with its own positive control, including a nested-relation arm. Measured: with
the old non-projecting fake, that deletion leaves all four `acceptBlockedReason` verdict tests
**green**.

## Coverage

The new browser suite drives from the PAYLOAD, never from a prop. #3935's pre-existing panel
test passed the refusal in as `transferErrorMessage` and stayed green while the app could not
reach that state; every arm here varies exactly one field on the `listMyPendingTransfers`
payload and renders the real page, so page → container → View is real. The Alert is asserted
on its message BODY (`/split ownership/i`), never on `/cannot be transferred/i` which the
Alert's own chrome would spell.

- **Red at base** (`242cd0bfc4`): 5 of 13 browser tests, 6 of 18 inbox tests.
- **17 mutants, 17 killed.** The decisive one is the SEAM — the container stops forwarding the
  field — which the browser suite kills (5 failures) and the server suite cannot see by
  construction (18/18 green). The disclosure-retention guard is likewise invisible to the
  isolated View suite (15/15 green) and killed only by the page-level one.
- The control arms (`acceptBlockedReason: null` → Accept enabled, no banner; the disclosure
  panel present) PASS at base by design and are labelled as controls, not counted as
  regression coverage. Mutants (h), (i), (j) and (k) are what earn them their place.

Full measured matrix, gate results and an explicit "not verified" list are in the comments.

### DIFF
--- a/src/components/Apps/AppTransferOffers.tsx
+++ b/src/components/Apps/AppTransferOffers.tsx
@@ -1,4 +1,3 @@
-import type { IncomingTransferRow } from '~/components/Apps/AppTransferOffersView';
 import {
   AppTransferOffersView,
   invalidationsForTransferResponse,
@@ -62,7 +61,16 @@ export function AppTransferOffers() {
 
   return (
     <AppTransferOffersView
-      transfers={(transfersQuery.data ?? []) as IncomingTransferRow[]}
+      /**
+       * 🔴 NO CAST. There used to be an `as IncomingTransferRow[]` here, against a
+       * hand-written structural duplicate of the service's return type — and a cast to a
+       * wider type is LEGAL, so `tsc` could not see the server drifting from the View at
+       * the one hop where that drift is invisible at runtime too. #3935 fixed the same
+       * hole on the owner's page (F3). `IncomingTransferRow` is now derived from
+       * `IncomingTransferView`, so the tRPC output type is checked against it directly and
+       * a field the service stops sending fails HERE.
+       */
+      transfers={transfersQuery.data ?? []}
       isLoading={transfersQuery.isLoading}
       errorMessage={transfersQuery.error?.message ?? null}
       busyTransferId={busyTransferId}
--- a/src/components/Apps/AppTransferOffersView.browser.test.tsx
+++ b/src/components/Apps/AppTransferOffersView.browser.test.tsx
@@ -28,6 +28,12 @@ function offer(over: Partial<IncomingTransferRow> & { transferId: string }): Inc
     fromUserId: 10,
     expiresAt: new Date('2026-08-17T12:00:00Z'),
     createdAt: new Date('2026-08-10T12:00:00Z'),
+    // The default is the TRANSFERABLE offer, so every pre-existing arm below keeps
+    // describing the ordinary case. The blocked branch is driven from the PAGE, in
+    // `src/tests/pages/apps/invites-transfer-blocked.browser.test.tsx` — deliberately not
+    // from a prop here, because a prop-driven arm is exactly what let #3935's owner-side
+    // defect sit green for months.
+    acceptBlockedReason: null,
     ...over,
   };
 }
--- a/src/components/Apps/AppTransferOffersView.tsx
+++ b/src/components/Apps/AppTransferOffersView.tsx
@@ -153,9 +153,31 @@ export function AppTransferOffersView({
          * recipient can neither accept nor understand. Unreachable while the service can
          * only send the constant or `null`, and reachable the moment a second reason is
          * added by someone who has no way to know the two branches differ. `blocked` is
-         * the single test; `!== null` is the one that matches the field's type.
+         * the single test.
+         *
+         * 🔴 LOOSE `!= null`, DELIBERATELY — IT FAILS OPEN ON AN ABSENT FIELD. The two
+         * checks are behaviourally identical for every value the TYPE permits (`string`
+         * or `null`) and differ on one the type does not: `undefined`. `!== null` would
+         * score a MISSING field as blocked, so an older API pool that does not send
+         * `acceptBlockedReason` yet would make this bundle refuse EVERY pending offer,
+         * with an empty explanation and Decline still live. That skew is structural here,
+         * not hypothetical: the bundle and the tRPC API are served by separate pools that
+         * roll independently, so new-client/old-server is a state every deploy passes
+         * through. And because the banner is otherwise unreachable (see the reachability
+         * note in `app-ownership-transfer.service.ts`), a mid-roll false refusal would be
+         * the ONLY way a real user ever sees it.
+         *
+         * Failing OPEN is the right direction: the server refuses an unacceptable
+         * transfer anyway — both gates are unchanged and authoritative — so the cost of
+         * a missed banner is one honest error toast, while the cost of a false banner is
+         * blocking a transfer that would have succeeded.
+         *
+         * This also matches how this file already treats the wire 118 lines up, where
+         * `expiresAt`/`createdAt` are widened because "a component that hard-required
+         * `Date` would be making a claim about the wire that this file cannot settle".
+         * The same is true of a field's presence.
          */
-        const blocked = transfer.acceptBlockedReason !== null;
+        const blocked = transfer.acceptBlockedReason != null;
         const blockedReason = transfer.acceptBlockedReason;
         return (
           <Paper
@@ -204,9 +226,20 @@ export function AppTransferOffersView({
                   the product", so this is gap-closure ahead of traffic.)
 
                   The message is the SERVER's own constant, verbatim, not a paraphrase: the
-                  recipient reads the same sentence the mutation would have returned. It
-                  states a constraint and instructs NOTHING, because there is no unlink
-                  path in the product for them OR the owner to pursue. */}
+                  recipient reads the same sentence the mutation would have returned, and it
+                  instructs nothing.
+
+                  🔴 THE STATED REASON FOR THAT — "there is no unlink path in the product" —
+                  IS REFUTED, and this comment used to repeat it. An owner CAN unlink today,
+                  by deleting the OAuth client: the relation is `onDelete: SetNull` and
+                  `oauth-client.router::delete` has no referencing-listing check. What is
+                  still true is the part that matters HERE: the RECIPIENT cannot do it — they
+                  own neither the listing nor the client — so a remedy addressed to them
+                  would still be a dead end, and the copy stays as it is. Whether the OWNER's
+                  copy should now name that route is a PRODUCT decision, deliberately not
+                  taken in this change; the constant's own rationale in
+                  `shared/constants/app-transfer.constants.ts` rests on the refuted premise
+                  and is flagged rather than edited here. */}
               {blocked ? (
                 <Alert
                   color="yellow"
--- a/src/components/Apps/AppTransferOffersView.tsx
+++ b/src/components/Apps/AppTransferOffersView.tsx
@@ -144,7 +144,18 @@ export function AppTransferOffersView({
          * re-run here and no `connectClientId` is in scope — the service computed this
          * from the SAME `refusesTransferForConnectClient` its two enforcement gates use,
          * so this surface cannot invent a stricter or looser rule of its own.
+         *
+         * 🔴 ONE PREDICATE, READ ONCE, USED BY BOTH BRANCHES. The banner below and the
+         * Accept button used to test this value differently — `blockedReason ? …` for the
+         * banner, `blockedReason != null` for the button — which agree on today's two
+         * possible values and disagree on a third: an EMPTY-STRING reason would have
+         * disabled Accept while rendering no explanation at all, leaving a card the
+         * recipient can neither accept nor understand. Unreachable while the service can
+         * only send the constant or `null`, and reachable the moment a second reason is
+         * added by someone who has no way to know the two branches differ. `blocked` is
+         * the single test; `!== null` is the one that matches the field's type.
          */
+        const blocked = transfer.acceptBlockedReason !== null;
         const blockedReason = transfer.acceptBlockedReason;
         return (
           <Paper
@@ -183,18 +194,20 @@ export function AppTransferOffersView({
               </Group>
 
               {/* 🔴 THE REFUSAL, BEFORE THE CLICK — the recipient half of #3935.
-                  `acceptTransfer` re-asserts the connect-client refusal IN-TRANSACTION
-                  precisely because a revision approve can link an OAuth client while an
-                  offer sits open, so this card can be live, valid and guaranteed to fail.
-                  Without this the recipient learned that only by pressing Accept and
-                  reading a toast — the same learn-it-by-doing-the-work shape the owner's
-                  tab already fixed, on the other side of the wire.
+                  `acceptTransfer` re-asserts the connect-client refusal IN-TRANSACTION, so
+                  a card here can be live, valid-looking and guaranteed to fail. Without
+                  this the recipient learned that only by pressing Accept and reading a
+                  toast — the same learn-it-by-doing-the-work shape the owner's tab already
+                  fixed, on the other side of the wire. (How reachable that state actually
+                  is today is answered honestly at the field's definition in
+                  `app-ownership-transfer.service.ts` — the short version is "not, through
+                  the product", so this is gap-closure ahead of traffic.)
 
                   The message is the SERVER's own constant, verbatim, not a paraphrase: the
                   recipient reads the same sentence the mutation would have returned. It
                   states a constraint and instructs NOTHING, because there is no unlink
                   path in the product for them OR the owner to pursue. */}
-              {blockedReason ? (
+              {blocked ? (
                 <Alert
                   color="yellow"
                   variant="light"
@@ -254,7 +267,7 @@ export function AppTransferOffersView({
                 <Button
                   size="xs"
                   color="red"
-                  disabled={busy || blockedReason != null}
+                  disabled={busy || blocked}
                   loading={busy}
                   data-testid={`apps-transfer-accept-${transfer.transferId}`}
                   onClick={() => onRespond?.(transfer.transferId, true)}
--- a/src/components/Apps/AppTransferOffersView.tsx
+++ b/src/components/Apps/AppTransferOffersView.tsx
@@ -2,6 +2,7 @@ import { Alert, Badge, Button, Group, List, Paper, Stack, Text, Title } from '@m
 import { IconAlertTriangle, IconArrowsExchange } from '@tabler/icons-react';
 import type { ReactNode } from 'react';
 
+import type { IncomingTransferView } from '~/server/services/blocks/app-ownership-transfer.service';
 import type { ListingKind } from '~/shared/constants/app-capabilities.constants';
 import { capabilitiesForKind } from '~/shared/constants/app-capabilities.constants';
 
@@ -24,15 +25,27 @@ import { capabilitiesForKind } from '~/shared/constants/app-capabilities.constan
  * directly renderable in a component test.
  */
 
-export type IncomingTransferRow = {
-  transferId: string;
-  appListingId: string;
-  slug: string;
-  name: string;
-  kind: ListingKind;
-  appBlockId: string | null;
-  iconUrl: string | null;
-  fromUserId: number;
+/**
+ * 🔴 DERIVED FROM THE SERVICE'S OWN RETURN TYPE, not hand-written beside it.
+ *
+ * This used to be a structural duplicate of {@link IncomingTransferView}, and the
+ * container's `as IncomingTransferRow[]` cast made the duplication INVISIBLE to `tsc`: a
+ * cast to a wider type is legal, so a field the service stopped sending — or never started
+ * sending — compiled clean at exactly the hop where the defect lives. #3935 hit the same
+ * hole on `edit.tsx` and fixed it the same way; the container's cast is now GONE as well,
+ * so the tRPC output type is checked against this one directly.
+ *
+ * The two date fields are deliberately WIDENED rather than inherited. The service types
+ * them `Date`; what actually arrives depends on the tRPC transformer, and a component that
+ * hard-required `Date` would be making a claim about the wire that this file cannot settle.
+ * {@link formatTransferExpiry} accepts either, so the widening costs nothing and the other
+ * fields — including `acceptBlockedReason`, the point of this change — stay pinned to the
+ * server's shape.
+ *
+ * A type-only import is erased at build time, so nothing of the service's module graph
+ * reaches the client bundle.
+ */
+export type IncomingTransferRow = Omit<IncomingTransferView, 'expiresAt' | 'createdAt'> & {
   expiresAt: Date | string;
   createdAt: Date | string;
 };
@@ -126,6 +139,13 @@ export function AppTransferOffersView({
       <Title order={4}>Ownership transfers</Title>
       {transfers.map((transfer) => {
         const busy = busyTransferId === transfer.transferId;
+        /**
+         * 🔴 THE VERDICT ARRIVES ON THE PAYLOAD, ALREADY DECIDED. No predicate is
+         * re-run here and no `connectClientId` is in scope — the service computed this
+         * from the SAME `refusesTransferForConnectClient` its two enforcement gates use,
+         * so this surface cannot invent a stricter or looser rule of its own.
+         */
+        const blockedReason = transfer.acceptBlockedReason;
         return (
           <Paper
             key={transfer.transferId}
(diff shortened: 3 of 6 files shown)


ZacxDev:
## The measured matrix

Everything below was executed, not predicted. Base is `242cd0bfc4` (`origin/main` at branch point).

### Red at base

Run against the base source in a clean checkout with its own `node_modules` (a first attempt with a
symlinked `node_modules` reported `Tests no tests` — a collection failure, which is not a red and was
discarded rather than read as one).

| Suite | At base | At HEAD |
|---|---|---|
| `src/tests/pages/apps/invites-transfer-blocked.browser.test.tsx` | **4 failed / 8 passed (12)** | 12 passed |
| `…/__tests__/app-ownership-transfer.inbox.test.ts` | **6 failed / 12 passed (18)** | 18 passed |

The 8 and 12 that pass at base are the control/invariant arms — Accept enabled with no reason, the
disclosure panel present, the fake's own projection control, the pre-existing inbox properties. They
are labelled as controls in the source and are not counted as regression coverage; the mutants below
are what earn them their place.

### Mutants — 15 applied, 15 killed

Each was applied alone, run, then restored; the three production files were SHA-256 compared against
a pre-battery copy after every restore and matched every time.

| # | Mutation | Killed by | The specific failure |
|---|---|---|---|
| a | `select`: delete `connectClientId: true` | inbox, **5 failed** | `expected null to be 'This listing is linked to an OAuth ap…'` + `expected { slug: true, name: true, …(3) } to match object { connectClientId: true, kind: true }` |
| b | verdict always `null` | inbox, 4 failed | `expected null to be 'This listing is linked…'` |
| c | verdict always the refusal | inbox, 3 failed | CONTROL arm: `expected { …(3) } to deeply equal { aot_mine_onsite: null, …(2) }` |
| d | drop the `kind` arm (hardcode `'offsite'`) | inbox, 1 failed | `expected 'This listing is linked…' to be null` (the on-site-with-a-client arm) |
| e | also return the raw `connectClientId` | inbox, 2 failed | leak guard: `expected true to be false` |
| n | **old non-projecting fake** + mutant (a) | inbox, 3 failed | the fake's own control + the arguments assertion — **all four `acceptBlockedReason` verdict tests stay GREEN.** This is the prerequisite-2 proof: without the projecting fake, deleting the select line is behaviourally invisible. |
| **f** | **SEAM — container stops forwarding the field** | **browser, 4 failed**; **inbox 18/18 GREEN** | `Cannot find element with locator: getByTestId('apps-transfer-blocked-aot_blocked')`, `expect(element).toBeDisabled()`, `expected [ {aot_ok}, …(1) ] to deeply equal [ {aot_ok} ]`. The decisive one — the server suite cannot see it by construction. |
| g | delete the refusal Alert | browser, 2 failed | `Cannot find element with locator: getByTestId('apps-transfer-blocked-aot_blocked')` |
| h | Accept `disabled={busy}` only | browser, 3 failed | `expect(element).toBeDisabled()`, and `acceptCalls` gained the blocked id |
| i | Decline disabled too | browser, 2 failed | `expect(element).toBeEnabled()` |
| j | per-PAGE instead of per-OFFER (block every Accept) | **browser, 2 failed — both in the MIXED arm** | `expect(element).toBeEnabled()` on the healthy card |
| k | suppress the disclosure on a blocked card (the #3935 owner-side treatment) | **browser, 1 failed**; isolated View suite **15/15 GREEN** | `Cannot find element with locator: getByTestId('apps-transfer-warning-aot_blocked')` — the operator decision is only observable at the page level |
| l | paraphrase the message instead of rendering the constant | browser, 1 failed | `expect(element).toHaveTextContent()` |
| m | re-add "Unlink the OAuth client first" | browser, 1 failed | `expect(element).not.toHaveTextContent()` |
| o1 | delete `acceptBlockedReason` from `IncomingTransferView` | **`tsc`, 2 errors** | `AppTransferOffersView.tsx(148,40): error TS2339: Property 'acceptBlockedReason' does not exist on type 'IncomingTransferRow'` |
| o2 | **CONTROL for o1** — same deletion, but with the pre-PR hand-written row type + the `as` cast restored | **nothing** | `typecheck: OK — 0 type errors`. This is the prerequisite-1 proof: the cast made the drift invisible. |

### The `expiresAt` / transformer question

Probed rather than asserted, by assigning `transfersQuery.data[0].expiresAt` at both types and
reading `tsc`:

- `const _: Date = …` → compiles.
- `const _: string = …` → `AppTransferOffers.tsx(57,9): error TS2322: Type 'Date' is not assignable to type 'string'.`

So the union transformer preserves `Date` in the client-inferred type. `IncomingTransferRow` widens
the two date fields to `Date | string` anyway — a superset, so it costs nothing and forced no change
of shape. No workaround was needed.

### Gates

| Gate | Result |
|---|---|
| `pnpm typecheck` (`tsconfig.json`) | **0 errors.** Covers the new suite — `tsconfig.json` excludes only `src/**/__tests__/**`, and `src/tests` is in `include`. |
| `tsc -p tsconfig.tests.json` | **823 errors at base file state, 823 at HEAD, byte-identical error sets — delta 0.** Both measured in the same tree with the same `node_modules`, base by restoring the six files. (This config is what sees `__tests__/`; its ~800 pre-existing errors are readable only as a delta.) |
| `vitest --project 'unit*'` | 1183 passed / 1 skipped (1184 files); 18674 passed / 21 skipped (18695 tests) |
| `vitest --project component` | 164 files passed; 1788 tests passed |
| `pnpm test:lint-rules` | 6 files, 243 tests passed (incl. `no-wholesale-module-mock` — the new suite uses the `importOriginal` spread) |
| `pnpm prettier:check` | clean. Negative control run: an injected mis-format made it report `[warn]` + exit 1, so the clean verdict is not a glob that matched nothing. |
| `eslint` on the six files | 1 pre-existing warning (`'_a' is defined but never used`, on an untouched line of the inbox fake); 0 errors |

### Not verified

- **No live/manual reproduction.** The issue records **zero rows in the transfers table in
  production — not zero pending, zero ever** — and reaching this branch additionally needs an offer
  opened while unlinked *and* a revision approve linking a client inside the offer window. There was
  nothing to click.
- **CI has not run yet** at the time of writing. Note also that the component/browser tier runs in
  the Tekton `preview / component-tests` job **report-only ⇒ ungated** — it is not local-only, but it
  does not block.
- **The type-only import's erasure is argued from `import type` semantics plus the component suite**
  (the browser build renders the page without pulling the service's module graph, which it could not
  do if the import survived). It has **not** been confirmed against a full `next build`. The same
  pattern already ships in `edit.tsx` from #3935.
- `no-server-infra-in-app-graph` passes, but that guard is scoped to `_app`'s graph, not
  `/apps/invites` — it is not evidence about this import.

ZacxDev:
**Provenance for the disclosure-panel decision**, which this PR states as "an operator decision" — recording it here because an audit correctly flagged that the PR asserts it with no visible trail.

The question was put to the repo operator explicitly (keep vs suppress the "Accepting this transfers the whole app to you" panel on a blocked card, presented with mockups of both) and **KEEP was chosen**. Reasoning, in order of weight:

1. The panel is a load-bearing **identity marker** — the existing pair test names it as one of three things stopping an irreversible ownership transfer from reading like a reversible seat invite. Dropping it on some cards makes that property conditional on data.
2. Unlike the owner-side copy #3935 suppressed — which described *mechanics of a transfer that can never start* — this describes *what the offer was for*, which is what the recipient needs in order to decide to decline.

The guard on it is labelled honestly in the spec as an **invariant guard**: it passes on `main` too, and exists to pin the decision against a future "tidy up the blocked card", not as regression coverage for the shipped defect.

**Open follow-up from the same audit, not yet decided:** on a blocked card the *yellow* "this offer can no longer be accepted" sits above the *red* "accepting this transfers the whole app to you" — so the louder panel is the counterfactual one. That visual hierarchy was not considered when KEEP was chosen and is worth a second look independently of this PR.

ZacxDev:
## Audit round — measured results

Follow-up commit on the same branch. All five findings addressed; everything below was
executed. **This round's gate is reset — none of the previous round's green is carried
forward.**

### What changed

| Finding | Fix |
|---|---|
| 🟡-1 two predicates on one value | one `blocked` test used by both branches, **pinned behaviourally** by a new arm feeding `acceptBlockedReason: ''` on the payload |
| 🟡-2 name reads as exhaustive | scope documented at the field: what's in it, the enumerated list of what isn't, and why. Name kept general — see below |
| 🟡-3 suite re-introduced the duplicate-type hole | `TransferRow` **is** `IncomingTransferRow`; `state.transfers` typed |
| 🟢 a test that could not fail | deleted, with a comment recording why so it isn't re-added |
| 🔴 unsubstantiated reachability | PR body corrected; the claim is retracted in all three places this change made it |

On 🟡-2 I kept the name and documented the scope rather than narrowing it: a second stable
read-time blocker should *join* this field rather than add a parallel one, so
`connectClientBlockedReason` would force a wire-contract rename the first time that happens.
The doc now enumerates the four refusals `acceptTransfer` makes that are **not** encoded
(banned recipient, missing app record, ownership already moved ×2, no-longer-pending) and
says why: each is a property of a row this read does not touch, or a race that only exists at
accept time — so a read-time answer would go stale between render and click, relocating the
failure this field exists to remove.

### Red at base — re-measured for the changed suites

In-tree this time (base files restored over the same `node_modules`), so the environment is
identical on both sides rather than a second checkout.

| Suite | At base `242cd0bfc4` | At this tip |
|---|---|---|
| `invites-transfer-blocked.browser.test.tsx` | **5 failed / 7 passed (12)** | 12 passed |
| `app-ownership-transfer.inbox.test.ts` | **6 failed / 12 passed (18)** | 18 passed |

The browser count moved 4 → 5: the new empty-reason arm is red at base too (the banner does
not exist there at all), and the deleted vacuous test is gone.

### Mutants — 16 applied, 16 killed

Re-run in full against this tip, not carried over. Anchors were updated for the unified
predicate; the three production files were SHA-256 compared against a pre-battery copy after
every restore and matched every time.

New this round:

| # | Mutation | Killed by | The specific failure |
|---|---|---|---|
| **p** | **re-split the predicate** — banner back on `blockedReason ?`, button back on `!= null` (i.e. the audit finding, reintroduced) | browser, 1 failed | `VitestBrowserElementError: Cannot find element with locator: getByTestId('apps-transfer-blocked-aot_empty')` — the new empty-reason arm, and *only* that arm |

Unchanged verdicts, re-measured: (a) 5, (b) 4, (c) 3, (d) 1, (e) 2, (n) 3, (g) 2, (h) 4,
(i) 3, (j) 2, (k) 1, (l) 1, (m) 1 failing tests respectively. The seam mutant **(f)** still
kills the browser suite (**5 failed**) while the inbox suite stays **18/18 green** — invisible
by construction, as before. **(k)** (suppress the disclosure on a blocked card) is still
invisible to the isolated View suite (**15/15 green**) and killed only at the page level.

### The `o1` count — corrected, with the discrepancy explained

The original matrix said "`tsc`, 2 errors". That was the honest measurement at `05af65299e`
**for the mutation as I defined it** — delete the type field *and* the computation. Deleting
only the type field leaves an excess-property error on the object literal, which is **3**, and
that is where the reviewer's number came from. The lesson is that the count is a property of
the mutation, not of the code, so both are now stated with the variant named.

Both are superseded at this tip, because 🟡-3 pinned the fixture:

| o1 variant | at `05af65299e` | at this tip |
|---|---|---|
| delete the type field **and** the computation | 2 | **15** (9 of them in `invites-transfer-blocked.browser.test.tsx`, which had **0**) |
| delete the type field **only** | 3 | **16** |

That +13 is the measurement of 🟡-3 being real: the file that exists to prove the seam now
fails when the seam's shape changes.

### Reachability — the 🔴, and why the description changed

Independently re-derived rather than taken on report, by enumerating **every** writer of
`AppListing.connectClientId` across `src/`, `packages/` and `apps/` — Prisma
`create`/`update`/`upsert` `data:` payloads, raw SQL, nested relation writes, and object
spreads inside mutation calls:

- **4 write sites total.** The only non-null one is `submitExternalListing`'s `create` of a
  **brand-new row**, where `submitExternalListingSchema` makes the field **required**. A
  listing is **born** linked.
- `beginListingRevision` copies `parent.connectClientId` → shadow (a `create`); the approve
  merge copies `shadow.connectClientId` → parent (the **only** `update` that names the
  column). In between, `buildListingPatchData` never assigns it and
  `updateListingPatchSchema` has no such key — so the revision round trip returns the value
  it started with.
- `mapAppBlockToListing` writes a literal `null`, on `create`s only.
- Zero raw `INSERT`/`UPDATE` touching `connect_client_id`; zero `connectClient: { connect }`;
  zero spreads inside any `appListing.(create|update|upsert)` call.

**So a listing cannot acquire a link, and `initiateTransfer` refuses to open an offer on one
born with it. The blocked state is not producible through the product today.** The PR body now
says that, and the "a revision approve can link a client while the offer is open" sentence is
retracted in the three places this change asserted it.

*Caveat carried with the claim:* the enumeration is textual plus call-site based, so it is
airtight against direct writes but would not catch a fully dynamic key
(`data[runtimeString] = v`) or a runtime-concatenated raw `UPDATE`. Neither construct exists.

Two **inherited** comments make the same unsupported claim and are deliberately **not** touched
here — `offsite-listing.service.ts:2273` and `:2573` ("…or linked an OAuth client"), plus the
sibling prose at `app-ownership-transfer.service.ts:71,248,354,454` and
`AppCollaboratorsPanelView.tsx:319`. They are #3935-and-earlier prose in files this PR does not
otherwise change, and rewriting six comments across three files would widen the blast radius of
a gap-closure PR. **Flagged as a follow-up.**

### Gates, re-run

| Gate | Result |
|---|---|
| `tsc -p tsconfig.tests.json` | unchanged from base — 823, byte-identical error sets, delta 0 |
| `vitest --project 'unit*'` | see below |
| `vitest --project component` | see below |
| `pnpm test:lint-rules` | 6 files, 243 tests passed |
| `pnpm prettier:check` | clean; negative control re-run (injected mis-format → `[warn]` + exit 1) |
| `eslint` on the changed files | 0 errors, 0 warnings |

ZacxDev:
## Audit round 2 — measured results

Follow-up commit `1a8edd1870` on the same branch. All four items addressed; everything below
was executed. **Gate reset again — none of the previous round's green is carried forward.**

Two of the three 🟡 were corrections to prose *I* wrote, and one refuted a premise this change
had inherited and repeated. Nothing was taken on report: each was re-derived against the code
first.

### 🟡-1 `!== null` → `!= null`

Confirmed the skew direction and took the one-character fix. The two comparisons are
behaviourally identical for every value the **type** permits and differ on one it does not
(`undefined`), so strict scored a **missing** field as blocked.

Made a guard rather than a comment: a new arm deletes `acceptBlockedReason` from the row —
the cast is load-bearing, since `TransferRow` correctly forbids a missing field, so stepping
outside the type is the only way to express what an older API sends — and asserts the offer
stays enabled **and submittable**. New mutant **(q)** re-tightens the comparison and dies
there, on `expect(element).toBeEnabled()`, and **nowhere else in the suite**.

🔴 That arm **passes at base** and is labelled a control in the source, measured not assumed
(it is one of the 8 passing at base). It passes there for a trivial reason — base has no
banner at all — so it is not regression coverage; mutant (q) is what earns it.

### 🟡-2 / 🟡-3 — both confirmed against the code

| Claim | Verified |
|---|---|
| `connectClientId String?` is nullable, no default | `schema.prisma:2849` — yes. The requirement is the **zod** input, not the DB. Corrected. |
| `oauth-client.router::delete` has no referencing-listing check | **zero** `appListing` references in that file; its only guard `rejectAppBlockClient` keys on `isAppBlockOauthClientId`, which these clients are not. Confirmed. |
| relation is `onDelete: SetNull` | `schema.full.prisma:2850` — yes. |

So the fifth writer is real, it is **owner-initiated and product-reachable today**, and it is
a class a call-site enumeration structurally cannot see because the writer is a **referential
action in the schema**. The conclusion is unaffected (SetNull only ever unlinks), but both
prose consequences were wrong and are fixed: the caveat now names the class, and "a future
unlink/link flow" now reads "a future **link** flow". Also recorded: `acceptBlockedReason` can
go **stale in the safe direction** — shown blocked, would now succeed. A stale block is a
delay; a stale allow would be a lie, and this cascade cannot produce one.

### 🟢 The refuted premise

Two comments of mine repeated "there is no unlink path in the product". Both corrected, keeping
the part that is still true and is why the recipient's copy stands regardless: the **recipient**
cannot unlink — they own neither the listing nor the client.

**No user-facing copy changed.** The constant's own rationale in `app-transfer.constants.ts`
rests on the same refuted premise and is the stated reason #3935 dropped "Unlink the OAuth
client first" from the string. That is now flagged in the PR body as a product decision, left
in place rather than edited under cover of a test PR.

### Mutants — 17 applied, 17 killed

Full battery re-run against this tip, anchors updated for the loosened comparison.

| # | Mutation | Killed by | Failure |
|---|---|---|---|
| **q** | `!= null` → `!== null` (fails closed on an absent field) | browser, 1 failed | `expect(element).toBeEnabled()` on the new fail-open arm |
| p | re-split the predicate | browser, 1 failed | `Cannot find element … apps-transfer-blocked-aot_empty` |
| f (seam) | container stops forwarding | browser **5 failed**, inbox **18/18 green** | invisible to the server suite by construction |
| k | suppress the disclosure when blocked | browser 1 failed, isolated View suite **15/15 green** | only observable at page level |
| a,b,c,d,e,n | server-side | inbox 5, 4, 3, 1, 2, 3 failed | unchanged from the previous round |
| g,h,i,j,l,m | View | browser 3, 4, 3, 2, 1, 1 failed | unchanged |

### Red at base — re-measured

`invites-transfer-blocked.browser.test.tsx`: **5 failed / 8 passed (13)**.
`app-ownership-transfer.inbox.test.ts`: **6 failed / 12 passed (18)**.

### The `tsc` numbers — both reconciled

**The re-auditor's 845 was a mutant measurement, not a clean HEAD one.** Measured directly: with
the `o1` mutant applied (delete `acceptBlockedReason` from `IncomingTransferView`),
`tsc -p tsconfig.tests.json` reports exactly **845**, and the +22 decomposes as
1 (`AppTransferOffersView.browser.test.tsx`) + 2 (`AppTransferOffersView.tsx`) +
5 (`app-ownership-transfer.inbox.test.ts`) + 14 (`invites-transfer-blocked.browser.test.tsx`).
Clean HEAD is **823**, the same as base. No contradiction between the two readings.

The "9 errors in this file" comment measured **14** here — the re-auditor's 13 predates the new
arm. Both are right about different tips, which is the argument against a count in a comment:
it has been replaced with the claim that actually matters (*it fails to compile here, under
both tsconfig projects, where it was zero before the fixture was pinned*).

🔴 **And one withdrawal.** The previous round claimed base and HEAD produce *byte-identical*
error sets. Delta 0 still holds — the set of `file|line|col|code` triples is byte-identical at
823 on both sides — but the raw text is **not**: 46 pre-existing errors in two files this
change never touches (`blocks.router.workflow.test.ts`, `workflow.service.test.ts`) print
their union members in a different order. Attributed with two controls rather than guessed:

- two back-to-back runs on the identical HEAD tree are byte-identical ⇒ not run-to-run
  non-determinism;
- base re-measured today is byte-identical to the round-1 base capture ⇒ that capture was not
  stale.

That leaves this change as the cause — display order only, no error appearing, disappearing or
moving, most likely a new type-only import perturbing module graph traversal. **For anyone
re-measuring: compare normalised `file|line|col|code` triples, not raw bytes** — a raw diff
reports 46 changed lines and a 0 delta simultaneously.

### Gates

| Gate | Result |
|---|---|
| `tsc -p tsconfig.tests.json` | 823 base / 823 HEAD, identical triples — delta 0 |
| `vitest --project 'unit*'` | 1183 passed / 1 skipped (1184 files); 18674 passed / 21 skipped |
| `vitest --project component` | 164 files passed; 1789 tests passed |
| `pnpm test:lint-rules` | 6 files, 243 tests passed |
| `pnpm prettier:check` | clean; negative control re-run |
| `eslint` on the changed files | 0 errors, 0 warnings |

🔴 **Every `tsc` output above was grepped for `Ineffective mark-compacts` before its count was
read** — 0 in all of them. Thanks for the note; that trap cost me a run early on and it is now
part of the measurement rather than a thing I remember to check.

PR #4122, opened 2026-08-19
fix(images): scope a username-addressed feed before the own-excluded decision

### LINKED ISSUE #3929: Own-excluded images can appear on another creator's feed when addressed by username
## Summary

When an authenticated caller browses another creator's images **by `username`** (rather than by `userId`), their own excluded content — private, blocked, nsfwLevel-0, unpublished — can be injected into that creator's feed.

Found during an adversarial audit of #3927 and **reproduced on `main` and on that PR branch identically**. It is pre-existing and was not introduced by #3927.

## Mechanism

`fetchBitdexPrimary` decides whether to run the "own excluded content" second pass here:

```ts
// src/server/services/image.service.ts:2984-2985
const skipOwnExcluded =
  !input.currentUserId || bitdexCursor || (input.userId && input.userId !== input.currentUserId);
```

That reads `input.userId`. But when the caller addresses the creator by handle, `username → userId` resolution happens **later and locally**, inside `getImagesFromBitdexPreFilter` (`image.service.ts:4221-4225`) — so at the point `skipOwnExcluded` is computed, `input.userId` is still undefined and the guard does not fire.

The second pass is then merged in, and the merge's content-scope filter list (`image.service.ts:3088-3112`) covers `modelVersionId`, `postId`, `postIds`, `types`, `baseModels`, `remixOfId`, `withMeta` and `fromPlatform` — but **has no user filter**. So nothing downstream re-scopes those documents to the creator being viewed.

## Reproduction

Authenticated caller (id 42), `username: 'someone-else'`, BitDex main pass empty, own-excluded pass returns the caller's own `Private` image:

- result: `source: 'bitdex'`, `ids: [777]`, `ownPasses: 1`
- expected: the caller's own private image should not appear on another creator's feed at all

Reachable through the public REST surface — `/api/v1/images` accepts `username` (`src/pages/api/v1/images/index.ts:56`).

## Scope — what this is and is not

**This is not a cross-user data leak.** The documents surfaced belong to the *caller*; nobody sees anyone else's private content. The defect is that the caller's own excluded content appears in a feed it does not belong to, which is a correctness and UX problem (and makes the feed misrepresent what that creator has posted), not a confidentiality breach.

## Suggested fix directions

Either resolve `username → userId` **before** `skipOwnExcluded` is computed, or add a user-scope filter to the merge's content-scope list so own-excluded docs are constrained to the view being rendered. The first is closer to the root cause; the second is defence in depth and would also cover any future entry point that reaches the merge with an unresolved creator.

## Note for the docs

The description table in #3927 says *"authenticated, viewing another user's profile → no second pass"*. That is correct only for the `userId` form of the request, not the `username` form.

ZacxDev:
Closes #3929.

## The defect

`fetchBitdexPrimary` decides whether to run the "own excluded" second pass from `input.userId`:

```ts
const skipOwnExcluded =
  !input.currentUserId || bitdexCursor || (input.userId && input.userId !== input.currentUserId);
```

When a caller addresses a creator by **handle** rather than id, `input.userId` is still empty at that point — `username` was resolved later and locally, inside `getImagesFromBitdexPreFilter`. So the "viewing someone else's profile" arm tested `undefined`, never fired, and the merge's content-scope list (`modelVersionId`, `postId`, `postIds`, `types`, `baseModels`, `remixOfId`, `withMeta`, `fromPlatform`) had no user filter to re-scope the result.

**Priced as feed integrity, not exposure.** Every document involved belongs to the caller, so nobody sees anyone else's private content. What breaks is that the feed misrepresents what that creator posted.

### Blast radius is narrower than #3929 states

The issue says "an authenticated caller browses another creator's images by `username`". In practice **the site's own profile gallery was never affected**: `src/components/Image/Infinite/UserMediaInfinite.tsx:161-162` sends **both** `userId: user.id` and `username`, so `input.userId` is populated and `skipOwnExcluded` already fires. `/user/<name>/images` and `/user/<name>/videos` are not reachable instances of this.

The reachable population is the public REST surface `/api/v1/images?username=` (`src/pages/api/v1/images/index.ts:56`) plus a hand-constructed `/images?username=X` URL. A search over `src/` found no in-app link that generates that shape — that is a search, not a proof of absence.

## What changed

**1. Resolve `username` → `userId` at the top of `fetchBitdexPrimary`, before anything reads the creator scope.**

Chosen as the primary fix because the merge is *downstream of the decision*: leaving the guard reading an unresolved field keeps it wrong for every consumer of that decision, and the wasted second-pass query is still issued.

- Precedence spelled exactly as the file's other three resolution sites spell it — `username && !userId`, so an explicit `userId` wins and a disagreeing `username` is ignored.
- The handle reaches the lookup verbatim: no lowercasing. Trimming is unreachable by construction — `usernameSchema` is `.regex(/^[A-Za-z0-9_]*$/).trim()` with the **regex first**, so a padded handle is rejected rather than trimmed.
- **Normally no extra round trip**: `getImagesFromBitdexPreFilter` guards its own resolution on `!userId`, and it runs once *per pass*, so one lookup replaces up to `MAX_PASSES + MAX_EMPTIED_PAGES` of them. **One exception**: the `hidden` branch sits before that resolution and can `return null` without reaching it, so `hidden: true` + `username` now pays one lookup it previously skipped.
- Uses `dbRead` only — no `dbWrite` fallback, unlike the Meili-side sites — deliberately, to keep a primary read off the feed hot path. On a replica miss the outcome is strictly better than before: BitDex declines and the Meili leg, which *does* have the fallback, resolves and answers.
- Writes a **new** input object rather than mutating the caller's. In-place would rob the Meili leg of its own `dbRead ?? dbWrite` resolution and its `NotFound`, and flip the shadow comparator's `hasFilters` from false to true for every username-addressed request. Pinned by a test.

**2. Add a creator scope to the merge's content-scope list.**

`if (input.userId) ownDocs = ownDocs.filter((d) => d.userId === input.userId);`

**These two guards are SEQUENTIAL, not independent — be precise about what the second one buys.** It catches a *resolved* creator that reaches the merge past a wrong upstream decision (e.g. `skipOwnExcluded` recomputed from a stale pre-resolution binding — mutant M3, where this clause alone keeps 777 off the page). It does **not** catch an *unresolved* creator: `input.userId` is falsy then, so the clause is skipped and the merge fails **open**. Measured, not assumed — mutant M1 deletes the resolution while leaving this clause in place and still serves the caller's private image (`expected [ 101, 777 ] to deeply equal [ 101 ]`). Do not cite this as defence-in-depth for the defect as originally reported.

## Edge cases

| Case | Behaviour |
|---|---|
| Handle resolves to nothing | Decline before any BitDex query; Meili leg raises NotFound |
| Handle → the caller | Second pass runs (own profile), pinned so the fix can't over-block |
| `userId` + disagreeing `username` | `userId` wins — spelled identically to the file's other sites; pinned both directions |
| Casing | **None introduced.** Pinned by asserting the exact `findUnique` argument |

## Tests

`src/server/services/__tests__/bitdex-username-own-excluded-scope.test.ts` — 13 tests, driven from a handle→id table the code consults rather than a pre-set `userId`, so they prove the guard is *reached*. Every fixture is **schema-valid** (`^[A-Za-z0-9_]*$`); a hyphenated handle could never reach this code from a real request.

**Matrix: 6 red at `origin/main`, 13 green at HEAD.** The other 7 are labelled **invariant guards**, not counted as regression coverage.

Red at base:
- `expected [ [ 'civitai', …(6) ] ] to have a length of +0 but got 1` — second pass issued for another creator's feed
- `expected [ 101, 777 ] to deeply equal [ 101 ]` — own private image interleaved into a real creator feed
- `expected [ [ 'civitai', …(6) ] ] to have a length of +0 but got 1` — issued for an unresolvable handle too
- `expected [ 101, 777, 909 ] to not include 909` — merge had no creator scope
- `expected [ 777 ] to deeply equal []` — same fake, only the handle differs, and the result did not
- `expected [ 101, 777 ] to deeply equal [ 101 ]` — the input-copy guard, which also proves the resolution had its effect

Full suites green: `src/server/services/__tests__/` 266 files / 3786 tests; `pnpm test:lint-rules` 243 tests.

### Mutation results (measured)

| # | Mutation | Killed | First failure |
|---|---|---|---|
| M1 | Delete the resolution entirely (the defect as reported) | ✅ 4 | `expected [ 101, 777 ] to deeply equal [ 101 ]` |
| M2 | Swap precedence — `username` wins over explicit `userId` | ✅ 2 | `expected [] to have a length of 1 but got +0` |
| M3 | Resolve, but compute the guard from the **pre-resolution** value | ✅ 2 | `expected […] to have a length of +0 but got 1` |
| M4 | Drop the merge creator scope | ✅ 1 | `expected [ 101, 777, 909 ] to not include 909` |
| M5 | Invert the merge creator scope | ✅ 4 | `expected 'meili' to be 'bitdex'` |
| M6 | Unresolvable handle no longer declines | ✅ 1 | `expected […] to have a length of +0 but got 1` |
| M7 | Merge scope compares `currentUserId` instead of `userId` | ⚠️ **survived** | — |
| M8 | `.toLowerCase()` on the handle | ✅ 5 | `expected "dbRead.user.findUnique" to be called with arguments: […]` |
| M9 | Invert the skip guard outright | ✅ 7 | `expected [] to deeply equal [ 777 ]` |
| M10 | M3 **+** M7 together | ✅ 3 | `expected [ 101, 777 ] to deeply equal [ 101 ]` |
| M11 | M3 **+** M4 together | ✅ 4 | `expected [ 101, 777 ] to deeply equal [ 101 ]` |
| M12 | Mutate the caller's `input` in place | ✅ 1 | `expected 7 to be undefined` |
| M13 | `.trim()` on the handle in the service | ⚠️ **survived** | — |
| M14 | Reorder `usernameSchema` to trim **before** the regex | ✅ 1 | `expected true to be false` |
| MX | Control: a deliberately THROWING mutant | n/a | reported 13/13 green — flagged **INVALID** by the harness |

**Two equivalent mutants, both stated rather than papered over.**

**M7** is equivalent under a correct upstream guard: `skipOwnExcluded` refuses the pass whenever `userId !== currentUserId`, so no reachable input separates the two spellings. **M10 kills it** — proving they diverge exactly when the guard they back up fails, which is why the clause reads `input.userId`.

**M13** (a service-side `.trim()`) is equivalent by an **external** invariant: all three `username` entry points (`image.schema.ts:363`, `pages/api/v1/images/index.ts:56`, `pages/api/v1/blocks/images.ts:99`) go through `usernameSchema`, and the only non-schema caller (`pages/api/mod/unblock-images.ts`) passes no username. No realistic fixture can kill it, because padding never reaches the service. So the PREMISE is pinned instead — **M14** reorders the schema and dies. The test says explicitly that it does not cover the service-side mutant.

🔴 **A mutant that THROWS is not a valid mutant.** `getImagesFromSearch` wraps the BitDex path in `try/catch` and falls through to Meili, so a mutant that dereferences null produces *exactly the expected outcome* and reports a clean false survival. The harness now checks well-formedness alongside anchor count, validated against control **MX**, which reported 13/13 green and was correctly flagged INVALID. Anchor hazards in this region, for anyone extending the battery: `if (username && !userId) {` matches **3** sites and `if (!targetUser) return null;` matches **2** — the unique anchor for the new site is `if (input.username && !input.userId) {`.

## Coordination with #4114

#4114 (`zach/3930-bitdex-serve-outcome-metric`) edits the same functions. Test-merged twice, independently, with the same result: **one conflict in `image.service.ts`, an insert/insert at the same anchor with an empty base side**. Resolve by keeping both hunks, #4122's block first; ordering is semantically neutral for the metric. Merged tree green and typecheck 0 in both runs — 277 files / 3910 tests in mine, 278 / 3922 in the auditor's independent re-merge (different `main` tip, hence the count).

**For whoever merges second:**

1. #4114 documents its moderator decline as *"Deliberately BEFORE any outcome is recorded… a policy choice"*. This PR's `return null` for an unresolvable handle is the same category and carries **no such note** in the merged tree — add one, or the convention gains an undocumented sibling.
2. **Measured** on the merged tree: an unresolvable handle under **#4114 alone** yields `source=bitdex`, `ids=[777]`, 1 BitDex call, and `bitdex_primary_result_total{outcome="served",mode="primary"} +1` — the counter scores a defective serve as a **success**. Merged, it is 0 calls and no series moves. So #3929 was inflating #4114's `served` numerator.
3. **Directional, not measured**: this PR also moves some previously-`served` requests to `fallback_empty`. If #4114's served ratio is ever used as a rollout gate, **its baseline shifts when this lands.**

## Gates

- `pnpm typecheck`: **0 errors**.
- `tsc -p tsconfig.tests.json`: **825 at `origin/main`, 825 at HEAD in the same tree — delta 0**, none in the changed files. (Re-derived from scratch; raw text differs only in TypeScript's property ordering for the same structural type, so the comparison is on the file+code multiset.)
- `npx eslint` on the two changed files: 3 errors, all **pre-existing** in `image.service.ts` and outside the diff hunks; the new test file is clean. CI's changed-files ESLint job passes.
- `prettier --check`: clean.
- The `preview / component-tests` CI job is **ungated** — `pr-preview-pipeline.yaml` says *"To make it gating later, change the component-tests step to `exit "$RC"`"*, and the status itself reports "(report-only)".

## Known sibling defect, deliberately not fixed here

A `followed: true` feed has the identical hole. Probed on the fixed tree: `source=bitdex ids=[101,777] ownPasses=1` — the caller's own private image is still injected into their Following feed. `skipOwnExcluded` only inspects `input.userId`, and the merge creator scope cannot help because that view's scope is a *set* of ids resolved inside the pre-filter. `newCreators` looks the same by inspection (not probed). Filed as **#4123**; whoever takes it must test-merge against #4114 and this PR.

### DIFF
--- a/src/server/services/__tests__/bitdex-username-own-excluded-scope.test.ts
+++ b/src/server/services/__tests__/bitdex-username-own-excluded-scope.test.ts
@@ -1,0 +1,423 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// #3929 — own-excluded images appearing on ANOTHER creator's feed when the
+// caller addresses that creator by `username` instead of `userId`.
+//
+// `fetchBitdexPrimary` decides whether to run the "own excluded" second pass
+// from `input.userId`. When the request names the creator by handle, that field
+// is empty at the point the decision is made — the handle used to be resolved
+// later and locally, inside `getImagesFromBitdexPreFilter` — so the "viewing
+// someone else's profile" guard read `undefined`, did not fire, and the
+// caller's own private/blocked/nsfw0 content was merged into a feed it does not
+// belong to.
+//
+// Not a cross-user exposure: every document involved belongs to the caller. The
+// defect is feed integrity — the feed misrepresents what that creator posted.
+//
+// Same minimal-seam mocking as bitdex-empty-fallback.test.ts: stub the
+// event-engine-common submodule + infra clients + env so importing
+// image.service doesn't boot real infra, and null out the Meili client so the
+// Meili leg still RUNS (and still stamps source 'meili') without a live index.
+
+import type * as BitdexClient from '~/server/bitdex/client';
+import type * as FliptClient from '~/server/flipt/client';
+import type * as MeiliClient from '~/server/meilisearch/client';
+
+const { queryBitdexMock, getFliptVariantMock } = vi.hoisted(() => ({
+  queryBitdexMock: vi.fn(),
+  getFliptVariantMock: vi.fn(),
+}));
+
+vi.mock('../../../../event-engine-common/feeds', () => ({
+  ImagesFeed: class {
+    populatedQuery = vi.fn();
+  },
+}));
+vi.mock('../../../../event-engine-common/services/metrics', () => ({
+  MetricService: class {
+    fetch = vi.fn();
+  },
+}));
+vi.mock('../../../../event-engine-common/services/cache', () => ({ CacheService: class {} }));
+
+vi.mock('~/env/server', () => ({
+  env: new Proxy({ LOGGING: [] as string[] } as Record<string, unknown>, {
+    get: (target, prop) => {
+      if (prop in target) return target[prop as string];
+      if (typeof prop === 'string' && (prop.endsWith('_URL') || prop.endsWith('_ENDPOINT')))
+        return 'https://test:test@localhost:5432/test';
+      if (
+        typeof prop === 'string' &&
+        /(_CONCURRENCY|_LIMIT|_MS|_PORT|_TIMEOUT|_MAX|_SIZE|_COUNT)$/.test(prop)
+      )
+        return 1;
+      return undefined;
+    },
+  }),
+}));
+
+vi.mock('~/server/clickhouse/client', () => ({ clickhouse: {} }));
+vi.mock('~/server/meilisearch/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof MeiliClient>()),
+  metricsSearchClient: null,
+}));
+
+vi.mock('~/server/bitdex/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof BitdexClient>()),
+  queryBitdex: queryBitdexMock,
+}));
+
+vi.mock('~/server/flipt/client', async (importOriginal) => ({
+  ...(await importOriginal<typeof FliptClient>()),
+  getFliptVariant: getFliptVariantMock,
+  getFliptBoolean: vi.fn().mockResolvedValue(false),
+}));
+
+import { getImagesFromSearch } from '../image.service';
+import { dbMock } from '~/__tests__/mocks/db.mock';
+import { redisMock } from '~/__tests__/mocks/redis.mock';
+redisMock.redis.get.mockResolvedValue('[]');
+redisMock.redis.set.mockResolvedValue(undefined);
+
+// Pairwise distinct, none of them a default. A fixture whose ids collide would
+// let a mutant that reads the wrong field produce the right answer by accident.
+const CURRENT_USER_ID = 42;
+const OTHER_CREATOR_ID = 7;
+const THIRD_PARTY_ID = 13;
+
+const OTHER_CREATOR_USERNAME = 'some-other-creator';
+const CALLER_USERNAME = 'the-caller-themselves';
+const MISSING_USERNAME = 'no-such-handle';
+
+/**
+ * Both BitDex passes go through the same `queryBitdex`, so the fake has to tell
+ * them apart from its arguments or it would answer the same thing to both and
+ * every assertion below would be meaningless.
+ *
+ * The main pass filters availability with `Not(Eq(availability, Private))`; the
+ * own-excluded pass asks for the opposite inside an `Or`. Keying on that `Or`
+ * distinguishes them structurally rather than by position or call order.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- raw BitDex filter clauses
+type Clause = any;
+function isOwnExcludedQuery(filters: unknown): boolean {
+  if (!Array.isArray(filters)) return false;
+  return filters.some(
+    (clause: Clause) =>
+      Array.isArray(clause?.Or) &&
+      clause.Or.some(
+        (member: Clause) =>
+          Array.isArray(member?.Eq) &&
+          member.Eq[0] === 'availability' &&
+          member.Eq[1]?.String === 'Private'
+      )
+  );
+}
+
+const ownExcludedCalls = () => queryBitdexMock.mock.calls.filter((c) => isOwnExcludedQuery(c[1]));
+const mainCalls = () => queryBitdexMock.mock.calls.filter((c) => !isOwnExcludedQuery(c[1]));
+
+/** The other creator's ordinary published image — what the feed SHOULD contain. */
+const otherCreatorPublicDoc = {
+  id: 101,
+  url: 'abc',
+  hash: null,
+  nsfwLevel: 1,
+  userId: OTHER_CREATOR_ID,
+  type: 'image',
+  availability: 'Public',
+  postId: 55,
+  postedToId: null,
+  hasMeta: true,
+  onSite: true,
+  poi: false,
+  minor: false,
+  width: 100,
+  height: 100,
+  reactionCount: 3,
+  commentCount: 2,
+  collectedCount: 1,
+  sortAt: 1_700_000_000,
+  publishedAt: 1_700_000_000,
+};
+
+/**
+ * The CALLER's own private image — exactly what the second pass exists to
+ * re-add on the caller's own views, and exactly what must not appear on
+ * someone else's.
+ */
+const callerOwnPrivateDoc = {
+  ...otherCreatorPublicDoc,
+  id: 777,
+  userId: CURRENT_USER_ID,
+  availability: 'Private',
+  postId: 61,
+  reactionCount: 9,
+  commentCount: 8,
+  collectedCount: 7,
+};
+
+/**
+ * A private document belonging to NEITHER the caller nor the creator in view.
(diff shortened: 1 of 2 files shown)


ZacxDev:
Cross-PR note for #4114 (`zach/3930-bitdex-serve-outcome-metric`), which edits the same functions in `image.service.ts`. Test-merged rather than reasoned about.

**Textual result:** one conflict in `src/server/services/image.service.ts`, and it is a pure adjacency conflict — both PRs insert a new block at the same point, immediately after the moderator `scheduled`/`notPublished` decline, and the merge base side is empty. Resolution is to keep both hunks in either order; there is no overlapping edit.

**On the resolved tree:** `pnpm typecheck` 0 errors, and both PRs' suites green together — 277 files / 3910 tests across `src/server/services/__tests__/`, `src/server/bitdex/__tests__/`, `src/server/metrics/__tests__/` and `src/pages/api/__tests__/`.

**One semantic interaction that survives the clean merge, measured on the merged tree.** #4114 records `bitdex_primary_result_total`, whose denominator its own comment states is "requests that engaged BitDex". Driving an authenticated request with an **unresolvable** handle:

- **#4114 alone** (this PR's hunks reverted from the merged tree): `source=bitdex`, `ids=[777]`, 1 BitDex call, `bitdex_primary_result_total{outcome="served",mode="primary"} +1`.
- **Merged:** `source=meili`, `ids=[]`, **0 BitDex calls**, no series moved.

So on `main` today a bad handle serves the caller their own private image, and #4114's counter would score that defective serve as a **`served` success** — i.e. #3929 was quietly inflating the numerator #4114 exists to measure. After this PR the request declines before issuing any query, so it is not counted at all, which keeps the "requests that engaged BitDex" denominator true and puts it in the same uninstrumented policy-decline category as the moderator decline directly above it — the one #4114 already documents as "deliberately BEFORE any outcome is recorded".

Worth a look from #4114's side in case that population was expected to land in `fallback_empty` instead; under #4114's taxonomy neither `fallback_empty` nor `fallback_error` fits, since no query runs.

PR #4127, opened 2026-08-19
fix(chat-settings): merge settings.chat in Postgres instead of read-modify-write

ZacxDev:
Closes the last read-modify-write of `User.settings` left after #3986 consolidated the column's writers onto `patchUserSettings`.

## Phase 1 — the discriminator, and what it actually turned out to be

The brief asked whether the users holding `isEarlyAdopter: false` got there by (a) toggling the setting on and then off, or (b) a lost update silently reverting their opt-in — and asked me not to pick whichever seemed likely.

**Verdict: (b) is ruled out for `isEarlyAdopter`, and not by a correlation — by deployment chronology.** The early-adopter setting (#3942) and the atomic-write fix (#3986) reached production in the **same release, 5.1.12**. The preceding release, 5.1.11, contained neither. The deploy record shows production went straight from 5.1.11 to 5.1.12 with **no intermediate build**: the 31 commits that sit between them on `release` were batched into that one release rather than shipped individually, so no deployed artifact ever existed in which `isEarlyAdopter` was writable *and* the settings write was still a whole-blob read-modify-write.

So there was never a moment in production when the bug could have touched this key. That is a stronger answer than the correlation the brief suggested looking for, and it does not depend on it.

Three supporting code facts, each of which independently blocks mechanism (b) for this key:

1. **A dismissal could never have reverted it anyway.** The pre-#3986 dismissal path wrote `jsonb_set(settings, '{dismissedAlerts}', …)`, which touches only that one path and leaves every sibling key alone. The loss ran the *other* way — a whole-blob `setSettings` write could drop a concurrent dismissal, not the reverse. The correlation the brief proposed (do the `false` users show dismissal activity?) could not have separated the two mechanisms even in principle.
2. **The pre-fix write was a MERGE, so it could not delete a key.** `settings || $payload::jsonb` cannot remove `isEarlyAdopter`; it can only overwrite it with a value the payload actually carries. For a race to produce `false`, the racing writer's stale snapshot must already have held `false` — which by induction requires a prior explicit opt-out. A lost update cannot manufacture a `false` from nothing.
3. **Only one writer sets this key, and only on a deliberate click.** `SettingsCard.tsx` sends `mutate({ isEarlyAdopter: e.target.checked })` — a single-key payload. No bulk form submits it incidentally, `removeEmpty` preserves `false` (it drops only nullish values and empty arrays), and neither the zod schema nor `getUserSettings` injects a default, so nothing materialises a `false` on its own.

I did also run the correlation the brief asked for, and it is flat: dismissal activity is statistically indistinguishable between the users holding `true` and those holding `false` (~88% of each group has at least one dismissal, with near-identical means). Consistent with (a), and — per point 1 — not capable of showing (b) regardless.

**What I could not verify.** There is no audit or history table for settings writes, no trigger, and the raw SQL path does not move `updatedAt`, so there is **no per-user trace of when or by what a settings key changed**. The conclusion above is derived from code and release chronology, not from a per-user record, and it assumes no manual DB backfill wrote this key outside the application. If you want per-user attribution for future settings questions, that instrumentation does not exist today.

**The fix in this PR is warranted on the code reading alone** — and it is a *different*, currently-live instance of the same bug class, described below.

## Every writer of `User.settings`

Enumerated across `src/`, `packages/`, `apps/` and `scripts/` (Prisma `user.update/updateMany/upsert`, all four raw-SQL shapes, and every settings helper), not assumed.

**The two atomic primitives — the only code that names the column:**

| Function | Kind |
|---|---|
| `patchUserSettings` (`user.service.ts`) | ATOMIC. One statement, SET expression computed over the stored column, `RETURNING settings`. |
| `setAlertDismissed` (`user.service.ts`) | ATOMIC. Set operation on the stored `dismissedAlerts` array. |
| `setUserSetting` | thin wrapper → `patchUserSettings` |

**Callers — all route through those two:** `completeOnboardingHandler` (×2), `toggleUserFeatureFlagHandler`, `setUserSettingHandler`, `dismissAlertHandler`, `restoreAlertHandler`, `updateContentSettings`, `updateSubscription`, `postponeSubscription`, `hideDownload`, `confirmAir`, `copyGallerySettingsToAllModelsByUser` (in a tx), `updateCreatorShopSettings` (in a tx), and **`chat.setUserSettings` — the subject of this PR**.

**Outside `src/`:** one one-off script (`scripts/oneoffs/backfill-cosmetic-resale-listings.ts`) writes the column directly, but atomically and guarded on `jsonb_typeof`. Not app code.

**Cleared as non-writers:** `updateUserById` / `forceUpdateUserIdentity` / `deleteUser` / `restoreUser` / `toggleMuteHandler` (`userUpdateSchema` has no `settings` key — `filePreferences`, `autoplayGifs`, `browsingLevel` are real columns), `user-profile.service.ts` (writes the different `publicSettings` column), `BlockUserSubscription.settings` (different table), and the auth app's session shaping (reads only). No writer in `packages/`.

### The one that was still wrong — and one more, left for its own change

**Fixed here — `chat.setUserSettings`.** It read `settings.chat` from `getUserSettings` (Redis, 4h TTL), merged in JS, and wrote the whole `chat` object back, so the top-level `chat` key was replaced wholesale. It violated the contract written on the function it called, verbatim: *"No caller may read this and write the result back — that read-modify-write is the lost-update bug `patchUserSettings` exists to remove, and the cache makes its window the TTL rather than a few milliseconds."*

This one is genuinely reachable in ordinary use, and more so than the original report's scenario: the three `chat` sub-keys are written from **different surfaces** — `NewChat` writes `acknowledged` on chat-terms acceptance, `ChatList` writes `muteSounds`/`replaceBadWords` from the settings menu — so two requests carrying disjoint keys is the normal case, and the losing snapshot came from a 4h cache rather than a millisecond-old read. Symptom: a user who already accepted the chat terms is asked again, or a mute preference silently reverts.

**Found, NOT fixed here — `settings.creatorShop.featuredItemIds`.** Five sites in `creator-shop.service.ts` (`archiveCreatorShopItem`, `setCreatorShopItemListed`, `deleteCreatorShopItem`, `reviewCreatorShopItem`, `takedownCreatorShopItem`) read the array via `getCreatorShopSettings` — a **replica** read — filter it in JS and write it back. The write itself is atomic with respect to other keys (it goes through `mergeInto`), so this is not a whole-blob clobber, but a concurrent edit to `featuredItemIds` can still be lost, and replica lag widens the window. Same class, different shape: closing it needs an atomic array-element-removal primitive rather than a call-site change, so it belongs in its own PR rather than being smuggled into this one.

**`user.restoreAlert` has no caller** — confirmed. Zero occurrences of `trpc.user.restoreAlert` anywhere. The `restoreAlert` hits in `ReferralDashboard.tsx` are a locally-named wrapper over `user.dismissAlert({ dismiss: false })`, not this mutation. It remains externally reachable via a `UserWrite`-scoped token and is now functionally a duplicate of `dismissAlert({ dismiss: false })`; deleting it is a human call, so it is left alone here.

## Why this fix shape

Routed the call site onto the **existing** helper rather than adding a second atomic write path.

`patchUserSettings` is already the single place that writes this column and is already atomic at the DB level — the two options the brief posed are not alternatives here, because the repo's chosen shape is both. The defect was not a missing primitive; it was a call site pre-computing a sub-object in JS and passing it to `set` (which replaces the key) instead of `mergeInto` (which merges over the stored value, reading `settings->'chat'` inside the UPDATE). `toggleUserFeatureFlagHandler` already uses exactly that shape for `features`.

Writing a bespoke statement for chat settings would have been the *duplication that regenerates this bug* — a second write path to keep in sync, in a file that had already drifted from the contract once. One rule, one place.

The handler now also returns the value from `RETURNING settings` instead of the object it computed, because the client writes that return straight into its query cache (`queryUtils.chat.getUserSettings.setData`); returning the losing merge left the UI showing a value the database did not hold.

## Red at base, green at HEAD

New file `src/server/services/__tests__/user-settings-chat-race.behavior.test.ts`, built on the existing PGlite harness from #3986 — the statements the service emits run unmodified against a real in-process Postgres, and one statement is suspended at the wire so the other request runs strictly between the first request's read and its write.

**Red at `origin/main` (f419a64461): 3 of 7 fail. Green at HEAD: 7 of 7 pass.**

| Test | base | HEAD |
|---|---|---|
| keeps a chat-terms acknowledgement that lands while a mute toggle is in flight | **FAIL** | pass |
| keeps a mute toggle that lands while a chat-terms acknowledgement is in flight | **FAIL** | pass |
| returns the settings actually stored, not those computed from a stale snapshot | **FAIL** | pass |
| INVARIANT: writes only the keys it was given | pass | pass |
| INVARIANT: leaves unrelated top-level settings keys untouched | pass | pass |
| INVARIANT: creates the chat object for a user who has none yet | pass | pass |
| INVARIANT: busts the settings cache | pass | pass |

The four passing-at-base cases are **labelled INVARIANT guards in the file**. They pin behaviour the change must not break; they are not regression coverage and are not evidence the bug existed. Only the three that go red are.

Both interleaving directions are asserted, because a fix that merged in only one direction would pass the first case alone. Fixtures seed every sub-key with a value the writes change (or deliberately never touch) — a key seeded *absent* makes "the patch wins" and "the stored column wins" byte-identical, which would leave the central claim unpinned.

**Mutation battery — 4 of 4 killed, each by a test naming its own defect:**

| Mutant | Result |
|---|---|
| `mergeInto` → `set` (the defect re-expressed through the new helper) | KILLED — all 3 regression tests + the sibling-sub-key invariant |
| return `input` instead of the stored row | KILLED — only the returned-value test |
| delete the cache bust (narrowest expression, enclosing condition left intact) | KILLED — only the bust guard |
| ignore the input entirely | KILLED — 6 of 7 |

Unmutated and post-restore controls both green, and both mutated files verified byte-identical to their pre-battery copies afterwards.

**Other gates.** 66 tests passed across the five settings suites (chat race 7, race 29, gallery copy 3, set-user-setting SQL 19, early-adopter refresh 8). Typecheck run at base and at HEAD in one environment: 2478 errors on each, location+code sets byte-identical, none naming a changed file — that baseline is an artifact of a generated Prisma client that does not match this tree and is present equally on both sides, not something this branch introduces. ESLint and Prettier clean on both changed files, each validated with a planted-violation positive control first so the clean verdict is a fact about the code rather than about the command line.

## Scope, stated plainly

The tests measure **statement interleaving** between two logical requests, on a single PGlite connection. They do not model multi-connection MVCC, row locks or snapshot isolation — that is asserted by construction (single-statement writes), not measured. The settings cache is a pass-through in the test, which makes the window milliseconds; in production the snapshot came from a 4h TTL, so the test **understates** the bug rather than overstating it.


ZacxDev:
## Corrections to this PR's description

An adversarial audit checked the body's factual claims. The verdict was safe-to-merge and the load-bearing arguments hold, but three statements are wrong or imprecise. Correcting here rather than editing the body, since it may already have been read.

**1. "the 31 commits that sit between them on `release`" — wrong number. It is 63.**

`v5.1.11..v5.1.12` is 63 commits (63 first-parent, 63 non-merge). 31 is the count between #3942 and #3986 (`0192e6865c..b6a2e537fa`) — a different interval, mis-attributed. The conclusion is unaffected, and the underlying claim was independently confirmed by *stronger* evidence than the body cites: the GitOps deploy record in the infra repo shows `civitai-prod:…-28bdecf` (v5.1.11) deployed 2026-08-15 04:24Z and `…-9fd1cb5` (v5.1.12) deployed 2026-08-17 17:01Z, with **zero** image-update commits in the 2.5 days between. That is a deploy-level fact rather than a git-level one, which is what the argument actually needs.

**2. `takedownCreatorShopItem` does not exist.** The fifth deferred site is **`takedownCosmeticShopItem`** (`creator-shop.service.ts:1907`, block at `:1954`). No function by the name given exists anywhere in the repo.

**3. "The pre-fix write was a MERGE, so it could not delete a key" — imprecise.** Pre-#3986 `setUserSetting` had a *second* statement, `settings - $keys::text[]`, which did delete undefined-valued keys. The argument survives intact — a deletion still cannot manufacture a `false` value, which is what the phase-1 verdict rests on — but the stated reason is not the whole mechanism.

## Two behaviour changes the body does not mention

Neither is reachable from any current UI call site, so neither is a defect — but the body has a behaviour-preservation section and these belong in it.

- **Key deletion is no longer possible.** The old `set` path could delete a `chat` sub-key (zod preserves an explicitly-`undefined` optional key; `JSON.stringify` then drops it and `set` replaced the whole object). `mergeInto` cannot: the key survives the non-empty filter, serialises to `{}`, and the merge is a no-op. Reachable only via a `UserWrite`-scoped API token — both UI call sites send literal booleans. `chat.controller.ts:74`.
- **Empty input now takes a no-op branch.** `setUserSettings({})` used to write the cached snapshot back (itself a latent revert). It now short-circuits at `user.service.ts:2818` — no write, no cache bust, and **no `userUpdateCounter` increment**. Strictly better behaviour, but the write metric loses those calls, so a dashboard keyed on it will step down.

Also worth noting for anyone watching metrics: the location label moves from `user.service:setUserSetting:set` to `chat.controller:setUserSettings`.

## Claims that checked out

Independently re-verified: both PRs in v5.1.12 and neither in v5.1.11; the pre-#3986 dismissal statement touching only `{dismissedAlerts}`; `removeEmpty` preserving `false`; the dismissal-correlation figures (re-run on the replica: `false` n=59, 88.1% with ≥1 dismissal, mean 1.14; `true` n=247, 87.9%, mean 1.27); the 4-mutant battery with identical kill sets; 66 tests across five suites (7/29/3/19/8); typecheck 2478 at base and HEAD with identical location+code sets; lint and prettier clean with planted-violation positive controls; `user.restoreAlert` having no caller. An independent writer sweep across `src/`, `packages/`, `apps/`, `scripts/` and migrations found the same set — 3 production SQL statements, 14 call sites, all atomic, and zero Prisma `user.update*` calls including `settings`.

PR #4135, opened 2026-08-19
fix(robots): disallow the review routes for meta-externalagent

ZacxDev:
## What

Adds a `meta-externalagent` group to `robots.txt` disallowing the review **listing** routes (`/models/*/reviews`, `/3d-models/*/reviews`), and extracts the file's construction into a tested pure function.

## Why

`meta-externalagent` is Meta's declared AI/content crawler — one of the few that documents itself as honouring `robots.txt` (unlike `facebookexternalhit` / `meta-externalfetcher`, which Meta says may not). It has become a large and growing share of the requests to the review listing routes, which are server-rendered per request and among the more expensive pages we serve.

**This is a real trade-off, not a free win.** Those pages are self-canonical (`src/pages/models/[id]/reviews.tsx:206` canonicalises to itself, not to the model page), and the paged review query is reached from that page only — so review prose lives on those URLs and nowhere else. Disallowing them removes user review text from Meta's corpus. The model page keeps its rating aggregate and stays fully crawlable, and individual review permalinks (`/reviews/[reviewId]`) are deliberately left crawlable. If we'd rather keep review prose in that corpus, this PR is the wrong call and should be closed.

## 🔴 The subtlety worth reviewing carefully

RFC 9309 §2.2.1: *"If no matching group exists, crawlers MUST obey the group with a user-agent line with the `*` value, if present."* A named group therefore **suppresses `*` entirely**. So the naive version

```
User-agent: meta-externalagent
Disallow: /models/*/reviews
```

would make Meta ignore **every** existing `*` disallow — `/api/*`, `/moderator/*`, `/user/account`, the crawl-budget rules — leaving it **less** restricted than it is today.

This PR therefore emits the full `*` disallow list inside the Meta group as well, and the test asserts that superset property so a later edit that adds a path to `*` alone cannot silently loosen things for Meta.

`Host:` / `Sitemap:` moving below the new group is safe — RFC 9309: *"a `Sitemaps` record MUST NOT terminate a group"*; they are non-group records and stay global.

## Tests

`src/server/utils/robots.test.ts` — 8 cases. The two load-bearing ones:

- **the exact disallowed paths are pinned literally**
- **the Meta group disallows a strict superset of the `*` group**

plus: exactly two groups emitted; every crawl-budget path appears in `*`; review paths disallowed for Meta and not `*`; `Allow: /api/trpc/*` in both groups; the `*` group's emitted text **pinned literally** (so this refactor is provably output-neutral for every existing crawler); `Host` + all four sitemaps still emitted after the new group.

Mutants run against the suite, all **killed**: singular-typo path (change inert), over-broad `/models/*` (deindexes the catalogue), meta group repeating nothing, and a crawl-budget entry reaching only the Meta group.

## Audit corrections

An adversarial audit of the first commit found two things this PR originally got wrong, both now fixed in `18f54db`:

1. **The suite could not tell `/models/*/reviews` from `/models/*`.** It built its expectation by iterating the constant under test, so any string passed. Verified: a singular typo (inert) and an over-broad `/models/*` (deindexes the whole catalogue for Meta) both passed all 5 original tests. Now pinned literally.
2. **`crawlBudgetDisallowPaths` didn't actually build the `*` group** — those paths were hardcoded literals there, so adding an entry to the array shipped it to Meta *only*, the opposite of what the name promised. The `*` group is now emitted from the array.

An earlier revision of this description also claimed a stale `public/robots.txt` shadowed this route and was worth deleting. **That was wrong** — `public/robots.txt` is gitignored (`.gitignore:79`), untracked, and has never existed in this repo's history; the file I saw was a local untracked artifact. The route is what serves `robots.txt`, but not for the reason originally given. (Next.js treats a `public/` file colliding with a `pages/` route as a build error anyway, so such a file could not silently shadow it.)

## Rollout note

`robots.txt` is served with `max-age=300`, but the live response is currently sitting at `cf-cache-status: HIT` with an `age` well past that, and crawlers cache `robots.txt` on top of that (Google ~24h; Meta unspecified). **A revert is not instant** — purge the cached object as part of any rollback.

## Not verified

The traffic premise is not measurable from the tools available here (the access-log sample is filtered and short-retention), so "large and growing share" is stated from external measurement, not from anything in this repo. The change is cheap and reversible either way. It's also unproven whether disallowing this route *reduces* Meta's crawl volume or merely **redistributes** it onto other pages, since crawlers work to a per-host budget.

### DIFF
--- a/src/pages/robots.txt/index.tsx
+++ b/src/pages/robots.txt/index.tsx
@@ -1,78 +1,14 @@
 import type { GetServerSideProps } from 'next';
+import { buildRobotsTxt } from '~/server/utils/robots';
 import { getRequestDomainColor } from '~/server/utils/server-domain';
 import { respondWithText } from '~/server/utils/sitemap';
 import { getBaseUrl } from '~/server/utils/url-helpers';
-
-const disallowPaths = [
-  '/*/create',
-  '/api/*',
-  '/discord/*',
-  '/dmca/*',
-  '/intent/*',
-  '/models/train',
-  '/models/*/wizard',
-  '/models/*/model-versions/*/wizard',
-  '/moderator/*',
-  '/payment/*',
-  '/redirect',
-  '/search/*',
-  '/testing/*',
-  '/user/account',
-  '/user/downloads',
-  '/user/notifications',
-  '/user/transactions',
-  '/user/buzz-dashboard',
-  '/user/vault',
-  '/user/membership',
-  '/user/stripe-connect/onboard',
-  '/user/earn-potential',
-  '/tipalti/*',
-  '/research/*',
-  '/claim/*',
-  '/collections/youtube/auth',
-  '/questions/*',
-];
 
 export const getServerSideProps: GetServerSideProps = async (ctx) => {
   const color = getRequestDomainColor(ctx.req) ?? 'green';
   const baseUrl = getBaseUrl(color);
 
-  const lines: string[] = [];
-
-  lines.push('# *');
-  lines.push('User-agent: *');
-  lines.push('Allow: /api/trpc/*');
-  for (const path of disallowPaths) lines.push(`Disallow: ${path}`);
-  lines.push('');
-
-  lines.push('# Login pages — high alternate-canonical volume from /login?returnUrl=... variants');
-  lines.push('Disallow: /login');
-  lines.push('');
-
-  lines.push(
-    '# Ad/affiliate tracking parameters — canonical handles correctness, but each',
-    '# variant wastes crawl budget'
-  );
-  lines.push('Disallow: /*?adid=');
-  lines.push('Disallow: /*&adid=');
-  lines.push('');
-
-  lines.push('# Site-search query URLs — thin/duplicate content, low SEO value');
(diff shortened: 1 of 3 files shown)


ZacxDev:
Correcting the description publicly rather than silently, since a reviewer may already have read it.

The original body claimed a stale `public/robots.txt` shadowed this route and was worth deleting separately. **That was wrong.** `public/robots.txt` is gitignored (`.gitignore:79`), untracked, and has never existed in this repo's history — the file I was looking at was a local untracked artifact in my own clone. The conclusion (this route is what serves `robots.txt`) holds; the stated reason did not. There is no follow-up cleanup to do.

The original body also said the change "costs no discoverability" because the canonical model page carries the same review content. Also wrong: the reviews page is self-canonical (`src/pages/models/[id]/reviews.tsx:206`) and the paged review query is reached only from it, so review prose lives on those URLs alone. Disallowing them does remove that text from Meta's corpus. The body now states the trade-off instead of denying it.

Both were caught by an adversarial audit, along with two blocking test gaps now fixed in 18f54db.

Issue #3930, opened 2026-08-14
No metric distinguishes a BitDex-served feed page from a Meilisearch fallback


ZacxDev:
## Summary

There is no metric distinguishing "the BitDex primary path served this feed request" from "it returned nothing and we fell back to Meilisearch". The only signal is `console.log` (`src/server/services/image.service.ts:3134`, `:3188`). `bitdex_shadow_errors_total` (`src/server/bitdex/compare.ts:50`) does not cover it — that instruments shadow-mode comparison, not the primary serve/fallback split.

Raised during the audit of #3927.

## Why it matters

1. **#3927 is precisely a fallback-rate change on the highest-traffic feed** (`/images` browse). It moved a population of requests from "served empty" to "served via Meilisearch fallback", and there is currently no way to watch that number move, confirm the change did what it should, or grade a rollback.

2. **An outage is indistinguishable from an empty index.** `queryBitdex` (`src/server/bitdex/client.ts:110-165`) returns `null` on *every* failure — missing config, non-2xx, thrown error, abort — and never throws. So a fallback caused by BitDex being down looks exactly like a fallback caused by a legitimately empty result set. Today neither is visible at all; with a counter, they should be distinguishable from one another.

3. **Tail risk has no alarm.** Under index-wide BitDex emptiness (a reindex, or a filter on a field BitDex does not know → HTTP 400 → `null`; see the existing note at `image.service.ts:4281-4283`), the traffic currently served by BitDex collapses onto the Meilisearch leg in one step. That is the intended semantics and that leg carried 100% of this traffic pre-BitDex, but there is no breaker and — without this counter — no signal that it happened.

## Suggested shape

```
bitdex_primary_result_total{outcome="served"|"fallback_empty"|"fallback_error"}
```

Splitting `fallback_empty` from `fallback_error` is what addresses (2), and requires distinguishing the two at the client boundary — `queryBitdex` currently collapses them into a single `null`. Returning a discriminated result (or recording the reason at the point of failure, where the status code and error are still in scope) would make both the counter and any future retry/breaker logic possible.

## Related

- `clearTimeout` is not in a `finally` in `queryBitdex` (`src/server/bitdex/client.ts:105-165`), so the 30s abort timer leaks on the throw path — unlike its sibling `fetchBitdexDocuments`. Small, and adjacent to any work in this function.

ZacxDev:
## Traced — and the headline is that the fallback rate is measurable today: **19.4%**

[...]

Issue #4150, opened 2026-08-19
bitdex_primary_result_total cannot separate a structurally-empty feed from a real serve failure, so its ratio is not alertable


ZacxDev:
`bitdex_primary_result_total` carries only `outcome` and `mode`. That is enough to answer "how often does the feed path end at Meili" — the question #3930 was opened for — but not enough to answer "is that a problem", because two very different populations land in the same `fallback_empty` bucket and nothing separates them.

## The two populations

**(a) Structurally empty — expected, and self-amplifying.** A request that walks one of the five early `return null` paths in `getImagesFromBitdexPreFilter` records `fallback_empty` when the caller is signed in on the first page or carrying a non-`bdx:` cursor:

- `hidden` with no signed-in user
- `hidden` with no hidden images
- an unresolvable username
- `followed` with zero follows
- `newCreators` with none

None of these is a failure — there was nothing for BitDex to return. **And they cascade**: once a request falls back, its next page carries a Meili cursor, which the `bdx:` decode does not accept, so every subsequent page of that browsing session also records `fallback_empty`. One door-walk at the top of a session can account for many counted requests.

**(b) BitDex genuinely returned nothing for a query it should have served** — the case worth acting on.

## Why this matters now

Measured in production over ~95 minutes on the feed/list API tier: **~81% `served`, ~19% `fallback_empty`**, stable (0.801–0.812 across every 10-minute bucket). `fallback_error` and `fallback_exception` are a hard 0, and `bitdex_query_failures_total` is 0 against a positive control of ~99 queries/s — so the backend is healthy and essentially all of that 19% is population (a) plus whatever share of (b) exists, with no way to tell them apart.

The practical consequence: **the served ratio cannot be alerted on.** A threshold would encode the current user mix as a health invariant, and would drift whenever the proportion of signed-in first-page requests that walk a door changes — for reasons entirely unrelated to BitDex health.

## Proposed

Add a dimension that separates the two. Options, roughly in order of preference:

1. **A `reason` label on `fallback_empty`** distinguishing "no BitDex call was made for the feed query" (door-walked) from "BitDex was queried and returned nothing". This is the smallest change that makes the ratio meaningful, and the information already exists at the recording site — `image.service.ts:3666` already branches on `anyQueryFailed`.
2. **A feed-kind label** (`hidden` / `followed` / `newCreators` / `default` / …). More useful for product questions, higher cardinality.

[...]

PR #4152, opened 2026-08-20
fix(feeds): scope the own-excluded pass by creator SET, not just a single userId (#4123)

### LINKED ISSUE #4123: Own-excluded images appear in the Following feed — skipOwnExcluded is blind to set-shaped creator scopes
Same class as #3929, verified by probe, **not** fixed by PR #4122 — split out deliberately rather than widened into that PR.

## Symptom

An authenticated caller loads their **Following** feed (`followed: true`). The BitDex main pass comes back empty; the own-excluded second pass runs anyway and returns the caller's **own `Private` image**, which is then served.

Probed on the tree that already contains #4122's fix:

```
source=bitdex  ids=[101, 777]  ownPasses=1
```

`777` is the caller's own private image. So #4122 closes the `username` route and this one stays open.

## Mechanism

`skipOwnExcluded` only inspects `input.userId`. On a Following feed the creator scope is not a single user id at all — it is a **set** of followed user ids, resolved *inside* the pre-filter. So:

- the `userId`-shaped guard never fires, because there is no `userId`; and
- the merge's content-scope creator filter added by #4122 **cannot help**, because it scopes to one creator, not a set.

Both halves of #4122's fix are structurally unable to cover this shape. That is why it needs its own change rather than a wider version of that one.

## Scope

- `followed: true` — **probed and confirmed**.
- `newCreators` — looks identical by inspection, **not probed**. Treat as likely, not established.

Anything else that resolves a creator *set* inside the pre-filter is a candidate; the population worth enumerating is "scopes resolved after `skipOwnExcluded` is computed", not "scopes that take a user id".

## Correct pricing — this is feed integrity, not a leak

The documents returned belong to the **caller**. Nobody sees anyone else's private content. It is a correctness/feed-integrity defect: your own excluded content appears in a feed that is supposed to show other people's.

That said, the **reach is wider than #3929's**: the Following feed is a primary browsing surface, whereas #3929 needed a `username`-addressed request. Rate it on that basis.

## Likely fix

Hoist the followed-set resolution into `fetchBitdexPrimary` so the creator scope is known **before** `skipOwnExcluded` is computed — the same ordering principle #4122 applied to `username → userId`, generalised from "one id" to "a set of ids". Then either make `skipOwnExcluded` set-aware, or give the merge a set-shaped creator filter as the second layer.

⚠️ This touches more of `fetchBitdexPrimary`, which is also being edited by **#4114** (the served-vs-fallback counter). Whoever picks this up should test-merge against whatever of #4114 / #4122 has not yet landed, and **read the merged result** of the shared region — a clean textual merge is not a clean merge when two changes edit different lines of one function. That has already produced one real semantic finding between #4122 and #4114 (see #4122's thread).

## Not verified

No live or staging reproduction — the probe is at the unit seam. `newCreators` is inference from reading, not measurement.

ZacxDev:
Closes #4123.

A Following feed served the caller their own private/blocked/nsfw0 content.

## The defect

`fetchBitdexPrimary` decided whether to issue the "own excluded" second pass from `input.userId` alone. That covers a feed addressed to **one** creator — and, since #4122, one addressed by handle — but a Following or new-creators feed has **no `userId` at all**: its creator scope is a **set**, resolved later and locally inside `getImagesFromBitdexPreFilter`. So the `userId`-shaped test read `undefined`, did not fire, the second pass ran anyway, and the caller's own excluded content was merged into a feed that exists to show *other people's* work.

**Not a cross-user exposure** — every document returned belongs to the caller. It is feed integrity, with **wider reach than #3929**: Following is a primary browsing surface, whereas #3929 needed a `username`-addressed request.

## The fix

Generalised to the invariant both shapes share: **the own pass belongs only in a feed whose creator scope contains the caller.** A single `userId` is a one-element set, so the old disjunct is *subsumed* rather than left sitting beside a new one — one rule in one place, so a third scope shape cannot be added to one test and missed by the other.

```ts
const creatorScopes = await Promise.all([
  input.userId ? [input.userId] : null,
  input.currentUserId && input.followed ? getUserFollows(input.currentUserId) : null,
  input.newCreators ? getNewCreatorUserIds({ entity: 'images', domain: input.domain }) : null,
]);
const creatorScopeExcludesCaller = creatorScopes.some(
  (scope) => scope !== null && !scope.includes(input.currentUserId!)
);
const skipOwnExcluded = !input.currentUserId || bitdexCursor || creatorScopeExcludesCaller;
```

⚠️ **Membership, not "is this feed scoped to somebody else".** `toggleFollowUser` creates `{ userId, targetUserId }` with no `userId !== targetUserId` guard, so a caller **can** sit inside their own followed set, and can be on the new-and-upcoming board. Both are pinned as positive controls — a blanket "never run the second pass on a scoped feed" fails them.

**Cost:** both helpers are cached (`getUserFollows` via `userFollowsCache`, `getNewCreatorUserIds` via `fetchThroughCache`) and are the same helpers the Meili leg already calls for these scopes. The await is free where it matters: unscoped and own-profile feeds resolve to literals with nothing to await, and the two scopes that *do* await are the ones this guard then skips.

#4123 flagged `newCreators` as *"looks identical by inspection, **not probed**"*. It is now probed — same set-shaped scope, same place, same behaviour, covered in both directions.

## 🔴 A real collision with #4114, caught by its test rather than by reading

#4114's case *"followed-with-zero-follows: the own-content pass still counts"* went red, because that request now makes **zero** BitDex calls instead of one. That is the more correct behaviour — the door joined the other four instead of being the one that leaked a call, and #4114's own denominator is "requests that ENGAGED the backend".

But that case was documented as the **only killing case** for the third observation site, `if (ownExcluded) bitdexCallsObserved++`. Rewriting it to expect zero calls and stopping there would have left that line untested while every suite stayed green.

So it is **retargeted** onto the `hidden`-with-no-hidden-images door — not a creator scope, therefore untouched by this change — and the mutation battery confirms the retargeted case *still kills that line*. The old shape is kept as its own test asserting the new zero-call behaviour.

## ⚠️ Consequence for the #3930 metric

This **reduces** what `fallback_empty` counts, because followed/newCreators door-walks stop issuing the own pass and therefore stop being counted. It shifts the served-ratio baseline exactly as #4122 did — **do not compare that ratio across this change.** Measured immediately before this PR: **~81% served / ~19% `fallback_empty`**. Flagged on talos-infra#1155, which specifies the dashboard and alerts over that counter.

## Verification

| check | result |
|---|---|
| new suite red at base | **6 of 11 fail on `origin/main`**, all 11 pass here |
| the other 5 | positive controls asserting existing-correct behaviour — pass in **both** arms, so they are not manufacturing red |
| all `bitdex-*` suites | 6 files / 68 tests green |
| `tsc -p tsconfig.tests.json` | 852 errors, all pre-existing, **zero in the three files touched** |

**Mutation battery — 6 mutants, each required to die on its OWN named test** (not merely "something failed"):

| mutant | dies on |
|---|---|
| `revert-to-userid-only-guard` | followed feed + empty main pass |
| `invert-membership` | followed feed + empty main pass |
| `drop-followed-scope` | followed feed + empty main pass |
| `drop-newcreators-scope` | newCreators feed + caller is not a new creator |
| `some-to-every` | followed feed + empty main pass |
| `drop-owncall-observation` | hidden-with-no-hidden-images: the own-content pass still counts |

Two instrument failures were caught and corrected while doing this, both of which had produced a confident wrong reading first: the mutation harness grepped an **ANSI-prefixed** vitest summary and scored a passing suite as failed (its own precondition caught it); and `tsc -p tsconfig.tests.json` **OOM'd and printed `0 errors`** — a false zero that reads exactly like a clean result, re-run at an 8 GB heap to get the real 852.

## Not verified

- No live or staging reproduction — the tests drive the unit seam, as #4123's own probe did.
- The cache-vs-DB caveat is documented inline: `getUserFollows` is the cached set while the pre-filter reads `userEngagement` directly, so they can disagree while the cache is stale. `toggleFollowUser` refreshes on both follow and unfollow, so the window is write-propagation rather than the entry's TTL. Both directions are benign and self-healing. Deliberately **not** unified here — repointing the pre-filter at the cache would change what the feed *contains*, which is a different change from what this decision *reads*.

### DIFF
--- a/src/server/metrics/bitdex-feed-serve.metrics.ts
+++ b/src/server/metrics/bitdex-feed-serve.metrics.ts
@@ -62,10 +62,23 @@
 //   2. a `bdx:` cursor — i.e. a BITDEX-paginated request, which is NOT the same
 //      as any paginated request: a request that fell back to Meilisearch carries
 //      a Meilisearch cursor on its next page and does NOT match
-//   3. signed-in caller viewing ANOTHER user's profile
+//   3. `creatorScopeExcludesCaller` — the request is scoped to a set of creators
+//      that does not contain the caller. Named BY SYMBOL deliberately: it is
+//      computed from three scopes (`userId`, `followed`, `newCreators`) and any
+//      gloss of it drifts. This disjunct REPLACED "signed-in caller viewing
+//      another user's profile" in #4123; that wording described only the
+//      single-`userId` case and was silent on the set-shaped ones.
 // When none holds, the own-content query goes out regardless of what the feed
 // query did, so the request made a call and IS counted even though its feed
 // query never left the process.
+//
+// 🔴 #4123 CHANGED WHICH DOORS THIS COUNTS, so the enumeration above it is no
+// longer uniform. `followed`-with-zero-follows and `newCreators`-with-none are
+// now excluded (their creator scope is the empty set, which does not contain the
+// caller, so no second query goes out); `hidden`-with-no-hidden-images is still
+// counted, because `hidden` is not a creator scope. That is a REDUCTION in the
+// `fallback_empty` population and it SHIFTS THE SERVED-RATIO BASELINE, exactly as
+// #4122 did — do not compare that ratio across #4123.
 //
 // Counting it is correct under "requests that ENGAGED the backend". It is simply
 // not the same set as "requests whose FEED query went out", and a dashboard must
--- a/src/server/services/__tests__/bitdex-feed-source.test.ts
+++ b/src/server/services/__tests__/bitdex-feed-source.test.ts
@@ -413,8 +413,10 @@ describe('bitdex_primary_result_total moves, and moves a DIFFERENT series per ou
    * `userId` case and was silent on the set-shaped ones (`followed`,
    * `newCreators`). This was the THIRD copy of that enumeration in the repo and
    * the last to be corrected — the other two are in image.service.ts and
-   * src/server/metrics/bitdex-feed-serve.metrics.ts, and both point readers HERE
-   * as the pin. Reference the predicate by SYMBOL when updating it again.
+   * src/server/metrics/bitdex-feed-serve.metrics.ts. Of those, only
+   * image.service.ts points readers at THIS file as the pin; the metrics module
+   * points at the `skipOwnExcluded` symbol instead. Reference the predicate by
+   * SYMBOL when updating it again.
    *
    * This case drives an ANONYMOUS caller, which is the disjunct that makes the
    * exclusion apply — so on its own it flatters the claim. The sibling case below
--- a/src/server/services/__tests__/bitdex-feed-source.test.ts
+++ b/src/server/services/__tests__/bitdex-feed-source.test.ts
@@ -434,36 +434,43 @@ describe('bitdex_primary_result_total moves, and moves a DIFFERENT series per ou
    * 🔴 THE CASE THAT REFUTES THE ABSOLUTE VERSION OF THE CLAIM ABOVE, PINNED SO
    * IT CANNOT BE RE-ASSERTED.
    *
-   * A signed-in caller on the first page opening the Following feed with zero
-   * follows: the FEED query returns early inside the builder and never reaches
-   * the client — but no disjunct of `skipOwnExcluded` holds here (signed in, no
-   * `bdx:` cursor, not viewing another user's profile), so the own-content pass
-   * IS issued, completes, and counts. One call went out, so the request is
-   * recorded as `fallback_empty`.
+   * A signed-in caller whose `hidden` feed has no hidden images: the FEED query
+   * returns early inside the builder and never reaches the client — but no
+   * disjunct of `skipOwnExcluded` holds here, so the own-content pass IS issued,
+   * completes, and counts. One call went out, so the request is recorded as
+   * `fallback_empty`.
    *
    * That is defensible under the stated denominator — "requests that ENGAGED the
-   * backend" — and it is deliberately NOT changed here. What was wrong was the
-   * claim that all five doors are excluded: `currentUserId` is a PRECONDITION of
-   * two of them (`hidden`-with-no-hidden-images, followed-with-zero-follows), so
-   * for those two the exclusion can never apply at all.
+   * backend" — and it is deliberately NOT changed. What was wrong was the claim
+   * that all five doors are excluded: `currentUserId` is a PRECONDITION of two of
+   * them, so for those two the exclusion can never apply at all.
    *
-   * Measured before being written: `calls=1 ownCalls=1
-   * moved=["fallback_empty|primary"]`.
+   * 🔴 THIS CASE USED TO BE DRIVEN BY followed-with-zero-follows, AND #4123 MOVED
+   * IT HERE. That fix makes the own-content pass conditional on the request's
+   * CREATOR SCOPE containing the caller, so a Following feed with zero follows now
+   * issues zero calls and records nothing — it joined the other doors instead of
+   * being the one that leaked a call. `hidden` is not a creator scope, so this
+   * door is untouched by that change and still drives the state this case needs.
    *
-   * 🔴 This is also the ONLY killing case for `if (ownExcluded)
-   * bitdexCallsObserved++` — the third observation site, which had no mutation
-   * killing it. Delete that line and this case goes red, because the request then
-   * looks like it made no calls at all.
+   * 🔴 Retargeted rather than deleted BECAUSE it is the ONLY killing case for
+   * `if (ownExcluded) bitdexCallsObserved++` — the third observation site, which
+   * has no other mutation killing it. Rewriting the followed case to expect zero
+   * calls and stopping there would have left that line untested while every suite
+   * stayed green. Delete that line and this case goes red, because the request
+   * then looks like it made no calls at all.
    */
-  it('🔴 PRIMARY, signed-in, followed-with-zero-follows: the own-content pass still counts', async () => {
+  it('🔴 PRIMARY, signed-in, hidden-with-no-hidden-images: the own-content pass still counts', async () => {
     getFliptVariantMock.mockResolvedValue('primary');
     queryBitdexMock.mockResolvedValue({ documents: [], cursor: undefined });
+    // The door: signed in, so the `!currentUserId` line does not fire, and the
+    // hidden-image lookup comes back empty so the builder returns null after it.
+    dbMock.dbRead.imageEngagement.findMany.mockResolvedValue([]);
     const before = await readOutcomes();
 
     const result = await getImagesFromSearch({
       ...baseInput,
       currentUserId: CURRENT_USER_ID,
-      followed: true,
+      hidden: true,
     });
 
     const after = await readOutcomes();
@@ -476,6 +483,34 @@ describe('bitdex_primary_result_total moves, and moves a DIFFERENT series per ou
 
     expect(result.source).toBe('meili');
     expect(seriesThatMoved(before, after)).toEqual([key('fallback_empty', 'primary')]);
+  });
+
+  /**
+   * The counterpart, and the reason the case above had to move: after #4123 a
+   * Following feed with zero follows makes NO BitDex call at all, so it records
+   * nothing. The caller's creator scope is the empty set, which does not contain
+   * them, so the own-content pass is not issued.
+   *
+   * This is a REDUCTION in what `fallback_empty` counts, and it is deliberate —
+   * the request never engaged the backend, so counting it overstated the fallback
+   * population. ⚠️ It also shifts the served-ratio baseline, exactly as #4122 did.
+   * Do not compare that ratio across this change.
+   */
+  it('🔴 PRIMARY, signed-in, followed-with-zero-follows: no call goes out, so nothing is recorded', async () => {
+    getFliptVariantMock.mockResolvedValue('primary');
+    queryBitdexMock.mockResolvedValue({ documents: [], cursor: undefined });
+    const before = await readOutcomes();
+
+    const result = await getImagesFromSearch({
+      ...baseInput,
+      currentUserId: CURRENT_USER_ID,
+      followed: true,
+    });
+
+    const after = await readOutcomes();
+    expect(queryBitdexMock).not.toHaveBeenCalled();
+    expect(result.source).toBe('meili');
+    expect(seriesThatMoved(before, after)).toEqual([]);
   });
 
   it('SHADOW, BitDex answers with documents → served{mode="shadow"} +1 (Meili still serves the user)', async () => {
--- a/src/server/services/__tests__/bitdex-feed-source.test.ts
+++ b/src/server/services/__tests__/bitdex-feed-source.test.ts
@@ -403,9 +403,18 @@ describe('bitdex_primary_result_total moves, and moves a DIFFERENT series per ou
    * this comment glossed it and both were refuted by measurement. The exclusion
    * holds exactly when `skipOwnExcluded` (image.service.ts, declared beside the
    * own-content pass) is true, i.e. one of: anonymous caller; a `bdx:` cursor;
-   * signed-in caller viewing another user's profile. A NON-`bdx:` cursor does
-   * NOT qualify — and that is the common case, since a request that fell back to
-   * Meilisearch carries a Meilisearch cursor on its next page.
+   * or `creatorScopeExcludesCaller` — the request is scoped to a set of creators
+   * that does not contain the caller. A NON-`bdx:` cursor does NOT qualify — and
+   * that is the common case, since a request that fell back to Meilisearch carries
+   * a Meilisearch cursor on its next page.
+   *
+   * 🔴 THE THIRD DISJUNCT USED TO READ "signed-in caller viewing another user's
+   * profile", AND #4123 REPLACED IT. That wording described only the single-
+   * `userId` case and was silent on the set-shaped ones (`followed`,
+   * `newCreators`). This was the THIRD copy of that enumeration in the repo and
+   * the last to be corrected — the other two are in image.service.ts and
+   * src/server/metrics/bitdex-feed-serve.metrics.ts, and both point readers HERE
+   * as the pin. Reference the predicate by SYMBOL when updating it again.
    *
    * This case drives an ANONYMOUS caller, which is the disjunct that makes the
    * exclusion apply — so on its own it flatters the claim. The sibling case below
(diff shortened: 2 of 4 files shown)


ZacxDev:
## 🔴 Corrections to this PR's own body — two claims above are wrong

Posting rather than silently editing, because a reviewer may already have read and believed them.

### 1. "the same helpers the Meili leg already calls … so a scoped request adds a cache read rather than a query" — **wrong on its `getUserFollows` half**

`getNewCreatorUserIds` genuinely is shared: the pre-filter calls the same helper.

`getUserFollows` is **not**. It appears in `image.service.ts` at exactly two places — `getAllImages` (the legacy SQL feed path) and the line this PR adds. The search legs all read follows with a raw `dbRead.userEngagement.findMany` (`getImagesFromSearchPreFilter`, `getImagesFromBitdexPreFilter`, `getImagesFromSearchPostFilter`). So this PR introduces a **new** `userFollowsCache` dependency into the image-search hot path; it is not a reuse.

Steady-state cost is still small — `cacheNotFound` defaults to `true`, so a zero-follow caller is negatively cached rather than falling through to the DB on every request — but "a new, cheap cache read" is the honest basis, not "a read we already do". Corrected in the code comment.

### 2. "the await is free where it matters … the two scopes that DO await are the ones this guard then skips" — **a non-sequitur, and it was**

Those requests still *paid* the await; being skipped afterwards does not make the lookup free. As written, the `Promise.all` ran before both cheap disjuncts were consulted, so an anonymous `newCreators` request performed a cache read whose result was then discarded.

Fixed rather than just reworded: the two cheap disjuncts are now hoisted above the scope resolution, restoring the short-circuit the old single-expression `||` had for free.

### 3. 🟢 The staleness mechanism was described backwards

The comment said the two sources disagree "while the cache is stale". In fact `userFollowsCache.refresh()` re-reads through **dbWrite** while the pre-filter reads the **replica** — so the cache is the *fresher* side. The reachable window is **replica lag after a self-follow**, not cache TTL. The conclusion (narrow, self-healing, safe in the other direction) stands; the mechanism named did not.

---

## Also fixed in `d6c408fb69`

**Two canonical "verbatim disjuncts" comment blocks still enumerated the predicate this PR deleted** — including one in `src/server/metrics/bitdex-feed-serve.metrics.ts`, **a file the original diff never opened**. Both are the designated source of truth a maintainer is told to read instead of re-deriving; one literally said *"verbatim from that line"* while no longer being verbatim. The metrics file also still claimed `followed`-with-zero-follows and `newCreators`-with-none are counted — after this PR they are not. Both now reference the predicate **by symbol** rather than glossing it, since a gloss is what drifted the previous two times.

**Three semantic mutants were surviving a fully green suite**, because the helper *arguments* were unasserted: dropping `domain` (pins the SFW board while the feed filters on the red one), swapping `entity` to `'models'`, and resolving follows for the viewed user instead of the caller. Each produces a well-formed set from the *wrong source* and re-admits #4123 on one domain but not the other.

🔴 **The first version of those assertions was itself walkable.** `toHaveBeenCalledWith` passes if *any* call matched, and the pre-filter makes its own correct call to the same helper — so both mutants still survived. Now asserts **every** call.

🔴 **A test written during the fix round was deleted for being vacuous.** A sibling assertion for the `bdx:`-cursor short-circuit passed with the hoist removed — its hand-written cursor fixture almost certainly did not decode, so the request never had the property the test was named for. A test that cannot fail is worse than none. The `bitdexCursor` half of the short-circuit is therefore explicitly **not** claimed as tested.

## Re-verified after the fix round

The battery was re-run **in full**, not just for the changed mutants — a fix round resets the gate.

| | |
|---|---|
| mutants | **10, each dying on its OWN named test** |
| `bitdex-*` suites | 6 files / 72 tests green |
| `tsc -p tsconfig.tests.json` | 852 — **unchanged from baseline** — 0 in touched files |

## Known follow-ups, not addressed here

- **`hidden` is the next instance of this class.** The own pass is narrowed at merge time by `userId`, `modelVersionId`, `postId`, `postIds`, `types`, `baseModels`, `remixOfId`, `withMeta`, `fromPlatform` — **not** by `hidden`. So on a Hidden feed the caller's own excluded images that were never hidden are still merged in. Read from the code, not reproduced.
- **No merge-time structural counterpart for set scopes.** The single-`userId` case has a deliberate structural double-check at merge time; `followed`/`newCreators` do not, so the guard is their only defence.

ZacxDev:
## Audit status — four rounds, final numbers

Superseding the mutant/test counts in the body and my earlier comment, which are now stale.

| gate | result |
|---|---|
| mutation battery | **13 mutants, all dying on their OWN named test** |
| `bitdex-*` suites | 6 files / **76 tests** green |
| `Prettier (added files)` (blocking) | clean, verified using CI's own added-vs-modified partition |
| ESLint (added file) | clean |
| `tsc -p tsconfig.tests.json` | 852 — unchanged from baseline — **0 in touched files** |

Four independent adversarial rounds ran (initial + three deltas). Rounds 2–4 each confirmed **safe to merge**; round 4 found **no surviving mutants** and confirmed its delta changed no executable production code.

### What the rounds actually caught

| round | finding |
|---|---|
| 1 | Helper *arguments* unasserted — 3 semantic mutants surviving a green suite. Two stale canonical comment blocks, one in a file the diff never opened. A false cost claim. Unconditional awaits. |
| 2 | 🔴 **A prettier break on the blocking CI gate** — would have gone red on merge. A surviving mutant (`domain: 'red'` hardcode). Three comment inaccuracies the fix round itself introduced. |
| 3 | Two false comment sentences. Two surviving mutants — `!!bitdexCursor` → `!!input.cursor` and `domain ?? 'red'` — both now pinned. |
| 4 | No code defects, no surviving mutants. One false comment sentence. |

Along the way three of my own artefacts were found defective and fixed: an assertion that was **walkable** (`toHaveBeenCalledWith` passes if *any* call matches, and the pre-filter makes a correct call that always satisfied it), a test that was **vacuous** (its `bdx:` fixture didn't decode, so the request never had the property the test was named for), and a mutation harness that scored a passing suite as failed by grepping an **ANSI-prefixed** vitest summary.

### 🔴 The pattern worth flagging to a reviewer

**Four rounds running, the code was correct and a newly-written explanatory comment was wrong** — each refuted by a single probe:

- "the Meili leg resolves the same scope" — it returns early in these tests
- "a BitDex-paginated request doesn't reach the leg that resolves follows" — it does
- "`undefined` is the common production case for `domain`" — `getRequestBoardDomainColor` stamps a colour on this path

Every one was a **mechanism or frequency** claim written to explain a number that had been measured correctly. The measurements were right every time. Comments now state the invariant each test pins rather than a story about why, and the retracted explanations are recorded in place so the next reader doesn't re-derive them.

### Known follow-ups, deliberately not in this PR

- **#4155** — `hidden` is the next instance of this same class; the merge's content-scope list omits it. Read from the code, not reproduced.
- No merge-time structural counterpart for set scopes: the single-`userId` case has a deliberate second layer, `followed`/`newCreators` do not.
- A pre-existing gap, not introduced here: nothing pins that a Meilisearch `nextCursor` never carries a `bdx:` prefix — reasoned from the encoding code, not observed on a live cursor.

ZacxDev:
## Verification status — for whoever reviews this

### 🔴 The red `preview / smoke-tests` is NOT from this PR

Attributed, not assumed. The 4 failures are both `/moderator/reports` specs — the moderator-guard control and the report queue. Two controls:

- **#4156 shows the identical "4 failed"** and changes only `.github/workflows/lint.yml`. A CI-config change cannot break `/moderator/reports`.
- **#4160** also shows it, and touches 0 files under `image.service`/`bitdex`.
- `preview-image-feed.spec.ts` — the surface *closest* to this change — **passes**.

It is a pre-existing environmental failure. Flagging rather than staying quiet about it: merging through a red gate without attributing it is a habit worth not having.

### What is and is not verified

**Verified:** the predicate is a superset of the old one; 13 mutants each dying on their own named test; red-at-base/green-at-HEAD; typecheck, lint, prettier, and every other CI gate green.

🔴 **Not verified: any of it against a real request.** Every test mocks `queryBitdex`, so what is proven is *"given BitDex returns X, the guard decides Y"* — never *"a real Following feed stopped serving own-excluded content"*. That is the honest ceiling on this PR today.

### Two things done to narrow that

**1. The preview environment is genuinely BitDex-served — so an integration check there would not be vacuous.** Verified on `pr-4152.civitaic.com`'s own pod: `BITDEX_URL`/`FLIPT_URL` set in the live process, Flipt mode `"primary"` on every observation, `PRIMARY serving 100 docs`, and the own-excluded second pass observably issued (the `Or[nsfwLevel=0, availability=Private, blockedFor…]` shape). Worth checking first — previews have historically lacked `BITDEX_URL`, and a feed observation made in that state is a green that proves nothing.

**2. A falsifiable production prediction is pre-registered** (talos-infra#1155), with the pre-deploy baseline captured *before* this ships so the window cannot be chosen after the fact:

```
served_fraction (1h) :  0.8234     served 42.368/s   fallback_empty 9.086/s
serving image        :  civitai-prod:20260820000558-d93f742
```

This change makes `followed`/`newCreators` door-walks issue zero BitDex calls, so they leave the counter entirely. **Prediction: `fallback_empty` falls and the served fraction rises.** A move in only one, or the other way, means the model is wrong. ⚠️ Only the *sign* is predicted, not the size — the counter has no route dimension (#4150), so the door-walk population cannot be isolated.

### Still open — the tier that would actually settle it

A paired before/after on the preview: an authenticated user with an own excluded image who follows someone, loading the Following feed, **with the unscoped feed as a control** (absence alone could just mean the feature is broken everywhere) and a `main` preview as the other arm. Not done — it needs preview credentials or a browser session. It is the only check that directly observes the defect being gone.

ZacxDev:
## Update to my last comment: the preview spec I proposed would be **vacuous**. Not writing it.

I said the remaining tier was a preview spec driving an authenticated Following feed, and that it "needs preview credentials or a browser session." Having measured, the blocker is not access — it is fixtures.

### The measurement

Queried BitDex directly (the `civitai` index) for own-excluded content — the exact filter shape the app emits — per preview smoke role:

```
POSITIVE CONTROL  broad published query   total_matched = 30,180,166   ✓ probe works

user 2000000001   own-excluded  total_matched = 0
user 2000000002   own-excluded  total_matched = 0
user 2000000003   own-excluded  total_matched = 0
user 2000000004   own-excluded  total_matched = 0     (ci-smoke-gold)
user 8753561      own-excluded  total_matched = 5     (dev-clone user, not a smoke role)
```

**Every smoke-test role has zero own-excluded documents in the index.**

### Why that makes the test vacuous

The spec would assert "the caller's own excluded image does not appear in their Following feed." With zero such documents, that content cannot appear in *either* feed — so the assertion passes regardless of what the code does, **including with this PR fully reverted**. It would be a permanent green in the suite proving nothing, which is the exact failure mode this PR's review has been fighting all the way through.

The unscoped-feed control I said was mandatory is what exposes it: that control would fail too, because there is nothing to be present.

### What would make it viable

Seed own-excluded content (Private / nsfwLevel 0 / blocked) for a smoke role **and** get it into the BitDex index — the own-excluded pass queries BitDex, not the DB, so a freshly-seeded row is not enough; it has to survive the pgsync lag. That is preview-fixture infrastructure, not something to bolt onto this PR.

`8753561` has suitable data but is not a smoke role, has no minted credentials, and is dev-clone data that can change under the test.

### Two instrument failures on the way to this, both self-caught

Worth stating because the first one produced a *confident, wrong, uniform zero*:

1. The first probe used **`curl`, which is not installed in the preview pod.** Every user returned an empty string, which my parser scored as `0`. Five clean zeros, all meaningless. Caught only because the positive control also came back empty.
2. The second used the wrong index name (`images`), which 404s. Also empty everywhere, control included.

Only the third — `wget`, index `civitai`, control returning 30M — produced numbers worth reading. **The zeros above are real; the first two rounds of identical-looking zeros were not.**

### Where that leaves verification

- ✅ Preview is genuinely BitDex-served (Flipt `primary`, `PRIMARY serving 100 docs`, own-excluded pass observably issued) — so the *environment* is capable, the *fixtures* are not.
- ✅ Production prediction pre-registered with a pre-deploy baseline (talos-infra#1155).
- ❌ Direct observation of the defect being gone: **not achieved, and not achievable with current preview fixtures.** Stating that rather than shipping a test that would imply otherwise.

ZacxDev:
## Merged-tree check — main moved 11 commits and one of them lands *inside the same function*

Functional spot-check review passed (thank you). Before this merges, the thing that has already bitten this arc once: **`mergeable=MERGEABLE` only means no textual conflict.**

`#4148` (remix-gallery) changed `src/server/services/image.service.ts` **inside `fetchBitdexPrimary`** — the same function this PR modifies — and added 65 lines to `bitdex-feed-source.test.ts`, the same test file this PR retargeted a case in. That is precisely the overlap shape that produced the #4114/#4122 semantic collision earlier in this work, where git merged both edits cleanly, it compiled, the whole suite passed, and the result was still wrong.

### What #4148 does in that function

It narrows the own-excluded **merge**:

```ts
-    if (!wantsUnpublished && !input.isModerator) {
+    if (!wantsUnpublished && (!input.isModerator || input.publishedOnly)) {
       ownDocs = ownDocs.filter((d) => isPublicallyPublished(d, mergeNow, stats));
```

### Why the two compose rather than collide

They act at different stages and in the same direction:

- **This PR** decides *whether the own pass runs at all* (`skipOwnExcluded`).
- **#4148** filters *what that pass returned* (`ownDocs`).

When this PR's guard skips, `ownExcludedPromise` is `null`, `ownDocs` is empty, and #4148's filter is a no-op over an empty array. When the guard does not skip, #4148's filter behaves exactly as it does on `main`. Neither changes the other's inputs.

### Verified by running, not only by reading

Test-merged `origin/main` into the branch locally and exercised the merged tree:

| check | result |
|---|---|
| `bitdex-*` suites on the merged tree | **6 files / 78 tests green** (up from 76 — #4148 adds 2) |
| full mutation battery on the merged tree | **13 mutants, all dying on their own named test** |
| `pnpm typecheck` on the merged tree | **0 errors** |

The merge was then aborted, so this branch is untouched and still at the reviewed commit `77ea3f4555`. Nothing here changes what you reviewed.

⚠️ Scoped honestly: this verifies the merge **as of `origin/main` at the time of writing**. If main moves again before this lands — and it moves fast — the merged tree changes and this check does not carry over.

### One thing still red, deliberately not merged through

`preview / smoke-tests` remains failing, and it is **not from this PR** — root-caused and filed as #4171. #3573 migrated the moderator pages out of the main app, and two preview specs still assert the pre-migration `/moderator/reports` routing. It currently fails on 5 of 10 open PRs, including one that changes only a CI workflow file.

Flagging rather than quietly merging past it: the whole reason that gate is worth having is that nobody has learned to ignore it yet.

PR #4156, opened 2026-08-20
ci: run per-app typecheck scripts in CI

ZacxDev:
The root tsconfig.json does not cover any `apps/*` directory, so `pnpm run typecheck` has never checked any sibling app. Each app defines its own typecheck script, and none were wired into CI.

Measured all seven locally:

| app | script | result |
|---|---|---|
| event-engine | tsc --noEmit -p tsconfig.typecheck.json | ✅ clean |
| notifications | tsc --noEmit | ✅ clean |
| orchestrator-gateway | tsc --noEmit | ✅ clean |
| storage | tsc --noEmit | ✅ clean |
| creator-studio | svelte-check --tsconfig ./tsconfig.json | ✅ clean |
| moderator | svelte-check --tsconfig ./tsconfig.json | ✅ clean |
| **auth** | svelte-check --tsconfig ./tsconfig.json | ❌ 2 errors + 1 warning |

Six clean apps are wired as individual steps in the existing `apps` (App unit tests) job, which already does `pnpm install`. apps/auth is excluded with its pre-existing errors documented.

Closes https://app.clickup.com/t/868kt8pfu

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -401,3 +401,26 @@ jobs:
       - name: Assert every app suite actually ran
         if: always()
         run: node scripts/ci/assert-workspace-suites-ran.mjs apps-report.json apps
+
+      # ── App typechecks ──────────────────────────────────────────────────
+      # Typecheck is per-app, not covered by the root tsconfig.json include
+      # (which lists `src`, `packages/*/src`, `scripts/local-dev/*.ts` and
+      # `.next/types/**/*.ts` — no `apps/*`). Each app defines its own script.
+      #
+      # apps/auth is NOT wired up here — it has 2 pre-existing type errors:
+      #   src/lib/server/auth/providers.ts:165 — implicit `any` param
+      #   src/lib/server/auth/__tests__/establish-session.test.ts:100 —
+      #     `_store` on `never`
+      # Tracked as a separate issue.
+      - name: Typecheck event-engine
+        run: pnpm --filter @civitai/event-engine run typecheck
+      - name: Typecheck notifications
+        run: pnpm --filter @civitai/notifications-app run typecheck
+      - name: Typecheck orchestrator-gateway
+        run: pnpm --filter @civitai/orchestrator-gateway run typecheck
+      - name: Typecheck storage
+        run: pnpm --filter @civitai/storage-app run typecheck
+      - name: Typecheck creator-studio
+        run: pnpm --filter @civitai/creator-studio-app run typecheck
+      - name: Typecheck moderator
+        run: pnpm --filter @civitai/moderator-app run typecheck
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -358,7 +358,13 @@ jobs:
         run: node scripts/ci/assert-workspace-suites-ran.mjs packages-report.json packages
 
   apps:
-    name: App unit tests
+    # Named for BOTH things it runs. It used to be "App unit tests", and the typecheck steps
+    # were added under that name — so a type error in an app would have rendered in the checks
+    # list as a failing unit-test job, pointing the reader at the wrong suite. Safe to rename:
+    # neither `main` nor `release` has any required_status_checks (re-verified 2026-08-20 via
+    # the branch-protection API, not inherited from the earlier note), so no merge gate keys
+    # on the old string.
+    name: App unit tests + typecheck
     runs-on: ubuntu-latest
     # The `apps/*` suites — the same gap as `packages` above, in the sibling directory, and
     # missed when that one was closed: the root config registered `packages/*/vitest.config.*`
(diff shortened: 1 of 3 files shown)


ZacxDev:
Independently re-measured this today against `origin/main` @ `a0f4a86fba` before finding this PR — I had duplicated it in #4172, which I have now closed in favour of this one. Posting the findings here so they are not lost with it.

**Your table reproduces**, with one method note. A bare "0 errors" from an instrument nobody has watched go red is not a measurement, so each leg was run twice: once on the unmodified tree, once with a deliberate `const zz: number = "..."` planted under the app and then removed.

| app | unmodified | planted error |
|---|---|---|
| `event-engine` | 0 | 1 × `error TS2322` |
| `notifications` | 0 | 1 × `error TS2322` |
| `orchestrator-gateway` | 0 | 1 × `error TS2322` |
| `storage` | 0 | 1 × `error TS2322` |
| `creator-studio` | 0 errors, 0 warnings | — |

`auth` reproduces exactly as you have it — `providers.ts:165` implicit `any`, `establish-session.test.ts:100` `_store` on `never`. Independent confirmation of the exclusion.

**`moderator` is confirmed clean, and I nearly filed it as broken.** It reported 131 errors in my checkout, but every one descended from `Cannot find module '@civitai/redis' / 'kysely' / '@civitai/mod-utils/prompt-audit'` — unlinked workspace deps locally, not type defects. A clean `pnpm install --frozen-lockfile` settles it; both PRs' CI runs agree it passes. Worth knowing if anyone re-measures on a partial install and sees the same wall of errors.

**One thing I checked and want to record as a non-issue**, because it looks like a bug and is not one. The Svelte apps' `typecheck` script is a bare `svelte-check`, and `apps/{creator-studio,moderator}/tsconfig.json` both `extends ./.svelte-kit/tsconfig.json`, which does not exist in a fresh checkout. That reads like a vacuous green waiting to happen. It is safe here: all three Svelte apps declare `"prepare": "svelte-kit sync"`, which pnpm runs during `pnpm install`, so the file exists before the step runs. **Do not "fix" this by switching to `check`** — I did in my version and it is redundant. But if `prepare` is ever dropped from one of these apps, this gate goes vacuous silently rather than failing, so it is worth a comment in the workflow.

**Two structural suggestions, take or leave:**

1. These steps live in the `apps` job, whose display name is **"App unit tests"**. A typecheck failure will render under that name, so the checks list will point a reader at the unit suite for a type error. A rename, or a separate job, keeps blame readable.
2. The steps are sequential with no `continue-on-error`, so the first failing app masks every app after it. With six legs that is the difference between "event-engine is broken" and "event-engine and three others are broken" on the first run after a shared type changes. A matrix with `fail-fast: false` gets all six every time, at the cost of six `pnpm install`s — a real tradeoff, and the cheap half may well be the right call here.

Neither is blocking. LGTM as-is.

ZacxDev:
Pushed `86162872ee` — driving this one to merge as agreed. Three changes, plus **a retraction of one of my own review points**.

## Retraction first

I warned above that the Svelte apps' bare `typecheck` would "go vacuous silently" if `prepare` were ever dropped. **That was wrong.** I tested it — moved `.svelte-kit` aside and ran the exact script:

```
exit code: 1
ERROR "tsconfig.json" 1:1 "Cannot read file '.../.svelte-kit/tsconfig.json'."
COMPLETED 1 FILES 1 ERRORS
```

It fails loudly. No guard needed, and switching to `check` (which my closed #4172 did) is unnecessary. Left as-is.

## 🔴 The one that mattered, and it was not on my original list

**`pnpm --filter <name> run <script>` exits 0 when the filter matches nothing.** Measured on pnpm 10.28.1:

```
$ pnpm --filter @civitai/does-not-exist run typecheck
No projects matched the filters in "/home/.../civitai"
$ echo $?
0
```

Six hardcoded package names are six chances for a rename to turn this gate into a green no-op — the same shape as `prettier --check "$FILES"` reporting "All matched files use Prettier code style!" over zero files. And a **new** app under `apps/` is simply absent from the list: the gate stays green and the app is unchecked, which is this PR's own hole reopened by the next person to add an app.

So the six steps become one step driving `scripts/ci/typecheck-apps.mjs`, which reads the app set **from disk** — every `apps/*` with a `typecheck` script must run, each run is proven to have selected a real package, and `auth`'s exemption is an explicit entry that hard-errors if it goes stale. Mirrors `assert-workspace-suites-ran.mjs`, which exists for the same reason on the vitest side.

## The two points from my earlier comment

- **Failures are now collected**, not fatal on the first app. Six sequential steps abort at the first red one, so a shared type change breaking four reports one.
- **Job renamed** `App unit tests` → `App unit tests + typecheck`, so a type error stops rendering as a failing unit-test job. Safe: I re-verified against the branch-protection API today that neither `main` nor `release` has any `required_status_checks`, rather than trusting the note already in this file.

**Kept in the `apps` job on purpose.** `pnpm install` is the expensive part and this job has already paid it. A matrix buys parallel legs and per-app check names for six installs; the rename plus collected failures gets most of the legibility for free. Happy to switch if you'd rather have the granularity.

## Guards

Eight tests in `scripts/__tests__/typecheck-apps.test.ts`, each against a stub `apps/` tree with a fake `pnpm` on PATH so the no-match and failure paths are driven for real — not asserted. All 8 pass.

Worth admitting: the first draft had **six of them failing for the wrong reason** — my fixtures omitted `auth`, so the stale-exclusion guard fired before the case under test. Fixtures fixed, guard left alone. A test that goes red for a reason other than the one you meant is not evidence.

One thing this does **not** fix: `scripts/__tests__/` is outside every typecheck program (the root `include` carries only `scripts/local-dev/*.ts`), so this new test file is itself unchecked — as are the 12 already there. That is the `scripts/` half of the ticket's widening and wants its own PR.

ZacxDev:
## CI green on `975180cca8` — with the evidence, not just the checkmark

11 of 12 checks pass, **zero failures**. The only non-terminal one is `preview / deploy`, still `PENDING`.

A script whose entire purpose is catching vacuous greens should not be believed on a bare checkmark, so here is what the `App unit tests + typecheck` job actually printed:

```
Typechecking 6 app(s): creator-studio, event-engine, moderator, notifications, orchestrator-gateway, storage
Skipping auth:
  ::group::typecheck creator-studio (@civitai/creator-studio-app)
  ::group::typecheck event-engine (@civitai/event-engine)
  ::group::typecheck moderator (@civitai/moderator-app)
  ::group::typecheck notifications (@civitai/notifications-app)
  ::group::typecheck orchestrator-gateway (@civitai/orchestrator-gateway)
  ::group::typecheck storage (@civitai/storage-app)
All 6 app(s) passed typecheck.
```

Six apps named, six groups executed, `auth` skipped by the ledger. And the guards themselves are not laptop-only — they execute in CI:

```
✓ unit scripts/__tests__/typecheck-apps.test.ts (8 tests) 358ms
```

⚠️ **One honest caveat on that.** The `unit` job is `continue-on-error: true`, so a guard failure there renders red-but-ignored. The same suite also runs under `preview / unit-tests`, which is not ignorable — that is the one carrying the weight. Worth knowing before anyone relies on the `unit` job to catch a regression here.

## I broke a check and fixed it

The push before this failed `ESLint + Prettier (changed files)` — the added-files Prettier step is blocking by design and I had not run the formatter. Reproduced locally (`prettier --list-different` listed both new files), fixed with `--write`, and **re-ran the 8 guards after the reformat** to confirm the formatter had not disturbed the fixture strings or the fake-`pnpm` heredocs. Still 8/8. The check is now green on the same path that failed.

## Ready to merge from my side

`main` has no required status checks, so nothing here is an interlock — this is a judgement call, not a gate. Everything I changed is verified live rather than reasoned about:

- the vacuous-filter defect: measured on pnpm 10.28.1, not assumed
- the branch-protection claim behind the rename: re-queried today, not inherited from the note in the file
- the "svelte-check goes vacuous" warning: **tested and retracted** — it fails loudly
- the ledger: proven on CI to have run all six apps by name

Remaining, each wanting its own PR: `apps/auth`'s two real errors, and `scripts/` (incl. `scripts/__tests__/`, where this PR's own test file lives unchecked) still being outside the root typecheck.

PR #4165, opened 2026-08-20
feat(apps): open an app-store screenshot full size — prev/next, arrow keys, Esc

ZacxDev:
## The problem

On `/apps/store-preview/<slug>` a listing's screenshots were a **dead grid**. `ScreenshotTile` rendered a plain `` inside a `Card` with no click handler, no viewer, and no way to see a screenshot at anything larger than half a column.

Clicking a tile now opens that shot full-screen with prev/next controls, ArrowLeft/ArrowRight, Esc-to-close, the shot's caption and a `3 / 7` position indicator — and it opens on **the tile that was clicked**, not always on the first.

---

## What I surveyed, and what I reused

The first step was the survey, not the code. Everything below was read before anything was written.

### Reused

| Reused | For | Why |
|---|---|---|
| **Mantine `Modal`, controlled (`opened` / `onClose`)** | the whole dialog shell | The repo's modal base everywhere, and specifically the shape **`AppDetailsModal`** — the component in this same directory whose `SimpleGrid` the gallery was originally copied from — already uses. It supplies, correctly and for free, *every* a11y requirement in the brief: `role="dialog"` + `aria-modal="true"` + `aria-labelledby` wired to `title` (`ModalBaseContent.mjs:41-43`), `trapFocus`, `closeOnEscape`, `returnFocus` — all defaulting to `true` (`Modal.mjs:26-37`). **None of the a11y is hand-rolled.** |
| **`TrainingSampleViewer`** (`src/components/Training/TrainingSampleViewer/`) | the viewer's design | The closest existing thing in the repo: the only full-screen viewer that navigates over **plain URL strings** rather than a civitai `Image` entity. Ported: its full-screen `styles={{inner, content, body}}` chrome, its skip-missing navigation (`findSampleInEpoch`), its "N of M" counter that counts only reachable items (`navigablePosition` / `navigableCount`), and its `NavButton` disabled-at-the-ends arrows. |
| **The full-screen media chrome** | layout | `fullScreen` + `withOverlay={false}` + that `styles` block is byte-identical across `TrainingSampleViewer`, `GeneratedOutputLightbox` and `Model3DLightbox`. It is the repo's de-facto convention; it has simply never been extracted. |
| **`useHotkeys` (`@mantine/hooks`)** | arrow keys | The convention at every arrow-nav site in the repo (`TrainingSampleViewer.tsx:154`, `GeneratedOutputLightbox.tsx:32`, `ImageDetailProvider`). There is no shared arrow-nav hook to reuse — this *is* the pattern. |
| **Plain ``** | the media | Same as the tile and `AppDetailsModal`: the URLs are CDN edge URLs, not configured Next image domains. (There is zero `next/image` usage anywhere in `src/components/Apps/`.) |

### Rejected — each for a stated structural reason

- **`dialogStore` + `DialogProvider`** — the app's dialog stack, already imported by `AppListingComments` on this very page, and the candidate I expected to adopt. Two disqualifiers, both structural rather than stylistic:
  1. **Props are a snapshot taken at `trigger()` time** and the dialog renders outside the gallery's tree, so the viewer could not share the gallery's live broken-tile set. Keeping those two in step *is* the index-consistency requirement in the brief.
  2. **`DialogProviderInner` closes by removing the dialog from the store**, so the modal *unmounts* rather than toggling `opened` to `false`. Mantine's `useFocusReturn` restores focus from a `useDidUpdate` on `[opened]` and clears its own timeout in the unmount cleanup (`use-focus-return.mjs:12-28`) — so on that path `returnFocus` **cannot fire**, and focus return would have to be hand-rolled as a deferred `.focus()` racing the focus trap. A controlled `Modal` keeps the component mounted and the built-in does the work. (This is read off the installed source, not assumed.)
- **`Embla` (`EmblaCarousel.tsx`)**, the carousel `GeneratedOutputLightbox` uses — Embla derives its snap points from **measured slide geometry**, which there comes from Tailwind classes (`flex-[0_0_100%]`). The browser-mode `component` project loads **no stylesheet at all**, so `scrollNext()` would resolve to one snap point and every prev/next test would pass while measuring nothing. A carousel is also the wrong instrument for a handful of static screenshots that need an exact index.
- **`ImageDetailByProps`, `Bounty/ImageCarousel`, `ImageViewer`** — all three are keyed on a civitai `Image` **entity** (numeric `id`, `nsfwLevel`, `meta`) and pull a live tRPC query plus `ImageGuard2` and browsing-level providers. A listing screenshot is a bare `{ url, caption }` from an allowlist DTO; there is no id to hand them.
- **`CarouselIndicators`** — generic and reusable, but its dots are `aria-hidden` + `tabIndex={-1}` decorations. The brief wants a *readable* position indicator, so a `3 / 7` text node is the right thing.
- **A generic lightbox** — there isn't one. Every full-screen media viewer in the repo is bound to a domain type.

---

## Rules I chose, and where they live

The only interesting logic here is index arithmetic, so it is a separate pure module (`appListingScreenshotNav.ts`) that the **blocking** node project can gate.

- **One index space.** Every index — the tile's, the viewer's, the broken set's — is an index into the URL-filtered list `screenshots.filter(s => !!s.url)`. `ScreenshotGallery` owns that list *and* the broken set; the grid and the viewer both read them. The tile at index *i* opens the viewer at *i* with **no mapping step**, because a mapping step is where an off-by-one would live.
- **Broken shots → skipped.** A tile that fails to load still hides itself (unchanged behaviour), but it now *reports* upward, so the viewer skips the same shots the grid dropped. Broken shots are excluded from prev/next **and** from both numbers of the position indicator — a total that counted them would promise a screenshot the controls can never reach.
  - The tile's per-tile `broken` `useState` moved up to the gallery for exactly this reason. Rendered output is unchanged.
- **Boundaries → STOP, not wrap.** At the first viable shot "Previous" is `disabled`; at the last, "Next" is. (`TrainingSampleViewer` stops, `GeneratedOutputLightbox` wraps — I picked stopping: a short ordered gallery that silently loops reads as broken, and `undefined` is what disables the control honestly.)
- **A shot that dangles *while the viewer is open*** (tiles are `loading="lazy"`, so a below-the-fold dangling URL is only discovered when the viewer fetches it) is **rescued** to the nearest reachable neighbour — forward first — and closes the viewer if none remains. The viewer is never an empty frame, and the indicator is never about a shot that isn't displayed.
- **A11y**: the tile is a real `<button>` (`UnstyledButton`) with a **content-derived** accessible name (image `alt` + caption) rather than an `aria-label` that would strip the visible caption (WCAG 2.5.3 label-in-name); focus moves into the dialog on open and returns to the originating tile on close; `aria-modal` + `aria-labelledby`; Esc closes.
- **Not omitted in `preview`.** The moderator listing-media review renders this body read-only with an explicit omission ledger ("omit every LIVE / interactive / AGGREGATE surface"). A lightbox is none of those — it queries nothing, writes nothing, acts on nothing — and a moderator reviewing a listing's *media* is the reader who most needs it full size. Called out in the code so it reads as a decision.

---

## Tests — 42 new, split by which tier can gate

| File | Project | Tests | Gates? |
|---|---|---|---|
| `__tests__/appListingScreenshotNav.test.ts` | `unit` | 14 | **Yes** — CI's `Unit tests` job |
| `__tests__/appListingScreenshotViewerWiring.test.ts` | `unit` | 5 | **Yes** — the grid↔viewer seam, as source text |
| `AppListingDetailBody.viewer.browser.test.tsx` | `component` | 23 | **No** — report-only `preview / component-tests` |

The browser project is report-only, so putting the arithmetic *only* there would mean no gate at all. Hence the pure module + a seam gate in the node tier, and honest headers on all three saying what each cannot see.

### Red / green matrix at `origin/main`

Measured in a **clean worktree at `origin/main`** with the test files copied in — not by reasoning.

| Suite | At `origin/main` | At HEAD | Label |
|---|---|---|---|
| `viewer.browser.test.tsx` | **20 failed / 3 passed** | 23 passed | New-feature coverage — see below |
| `appListingScreenshotViewerWiring.test.ts` | **4 failed / 1 passed** | 5 passed | New-feature coverage; the 1 pass is the matcher positive-control test |
| `appListingScreenshotNav.test.ts` | `Tests no tests` — **fails to import** | 14 passed | New-feature coverage, and *not even a behavioural red* |
| `AppListingDetailBody.gallery.browser.test.tsx` (untouched) | 7 passed | **7 passed** | **Regression-shaped** — unchanged |
| `AppListingDetailBody.browser.test.tsx` (untouched) | 64 passed | **64 passed** | **Regression-shaped** — unchanged |

🔴 **Labelled honestly: none of the new tests is regression coverage.** At `origin/main` the browser tests fail because there is no tile to click — an *affordance-missing* red, not a behavioural one — and the node arithmetic suite fails to **import** a module that does not exist, which is not a red at all. The genuinely regression-shaped claim in this PR is the untouched sibling suites (71 tests) still passing, which is what says turning the tile into a button did not move the grid geometry or the page structure.

The 3 browser tests that pass at `origin/main` are **invariant guards, not coverage**: "the viewer is closed until a tile is activated", "zero screenshots renders no gallery and no viewer", "every screenshot broken — nothing can be opened". They were green before and are green after; they exist to stop the *other* assertions passing for the wrong reason.

### Mutation check — 21 mutants, 21 killed, 0 survived

Each mutant names **one designated test** and only that test is run, so a kill is attributed to a specific assertion rather than to "something failed". Every one reported `1 failed | 0 passed` — i.e. the designated test, and only it, went red.

Harness controls, because a mutation sweep lies in several directions:
- **Positive control on every `-t` pattern first**: all 21 patterns matched exactly 1 test and were green unmutated. A name filter that matches nothing reports `0 failed` and reads as a pass.
- **Anchor occurrence count asserted** before each replacement, and the file hash asserted to have *changed* after the write and to be restored byte-for-byte after — a mutant that did not apply otherwise scores as a false survival.
- **The `Tests N failed | M passed` line is counted, never the exit code.**
- Mutants were written from the *spec* (the rules above), not by walking the assertions — a list derived from the tests measures self-consistency, not coverage.

| Mutant | Designated test | Verdict |
|---|---|---|
| open always shot 0 | `clicking tile N opens the viewer on shot N` | KILLED |
| open off-by-one (`index + 1`) | same, **+** `each tile carries its index, its open handler…` (node seam gate) | KILLED |
| **tile index renumbered to grid position** | `a broken screenshot does not shift the index the surviving tiles open` | KILLED |
| next off-by-one (`nextIndex + 1`) | `next advances by exactly one, and prev goes back by exactly one` | KILLED |
| **prev/next swapped** | same | KILLED |
| ArrowLeft/ArrowRight swapped | `ArrowRight and ArrowLeft move the viewer in their own directions` | KILLED |
| navigation wraps instead of stopping | `returns undefined at each end rather than wrapping` (node) **+** `at the LAST shot…` | KILLED |
| navigation does not skip broken | `skips a RUN of broken screenshots` (node) **+** `a TRAILING run… ends the navigation` | KILLED |
| indicator total counts broken shots | `counts only the shots navigation can reach` (node) **+** `excludes broken screenshots from BOTH numbers` | KILLED |
| position 0-based | `is 1-based over the whole list` (node) **+** `clicking tile N…` | KILLED |
| tile rendered as a `div` | `each tile is a keyboard-activatable <button>…` | KILLED |
| tile element renamed in source | `the tile is a real button…` (node seam gate) | KILLED |
| `returnFocus={false}` | `focus returns to the ORIGINATING tile on close` **+** node seam gate | KILLED |
| `closeOnEscape={false}` | `Escape closes the viewer` **+** node seam gate | KILLED |
| indicator renders the raw index | `clicking tile N opens the viewer on shot N` | KILLED |
| caption always shot 0's | same | KILLED |
| gallery renders when empty | `a listing with ZERO screenshots renders no gallery and no viewer` | KILLED |
| viewer handed an empty list | node seam gate **+** `clicking tile N…` | KILLED |
| viewer handed its own broken set | node seam gate **+** `excludes broken screenshots from BOTH numbers` | KILLED |
| tile loses its broken reporter | node seam gate **+** `excludes broken screenshots from BOTH numbers` | KILLED |
| **pre-existing gallery fill class removed** | `1 screenshot — the lone tile fills the whole column` (untouched suite) | KILLED |

🔴 **The sweep found a real blind spot, and the first round was wrong.** The renumbering mutant (grid position used as the viewer index) **SURVIVED** the original index-space test, which broke shot index **1**: a renumbered click on the second visible tile resolves to index 1 — *the broken shot* — so the viewer's own rescue effect skipped forward and produced the correct answer for the wrong reason. Breaking shot **0** instead shifts every later index onto a *viable* shot, which the rescue cannot launder. The test now also asserts each tile's `data-screenshot-index` directly. Fixture header records this so it cannot be "simplified" back.

Two further honest notes from the same sweep:
- The "navigation does not skip broken" mutant initially came back **PARTIAL** — killed in the node tier, survived in the browser tier, again because the rescue effect reached the same answer by another route. The browser half only discriminates where correct code says *"nothing that way"* and the mutant offers a broken shot: hence the new `a TRAILING run of broken screenshots ends the navigation` test, which asserts a **disabled** control — a state no rescue can reach after the fact.
- The "tile is not a button" mutant was **invalid, not survived** — renaming the opening tag alone left `</UnstyledButton>` behind, so the browser run reported `no tests` rather than a failure. The harness flagged it rather than scoring it; it was split into a compiling mutant for the browser tier and a text mutant for the source gate.

---

## Verification

| Check | Result |
|---|---|
| `pnpm typecheck` | **0 errors** — and the instrument was validated: a planted `let position: string = 0;` produced `error TS2322` and `3 type error(s)`, then reverted (hash-verified). |
| `eslint` (6 changed files) | 0 errors, 0 warnings, **6 files reported linted** — instrument validated against a probe file that produced 2 errors + 2 warnings (`no-wholesale-module-mock`, `no-unused-vars`, `no-img-element`, `no-unloadable-image-fixture`). |
| `prettier --check` (6 files) | clean — and it *did* flag one of them before I ran `--write`, so the check is live. |
| node `unit`, `src/components/Apps/__tests__/` | **54 files / 997 tests passed** |
| browser `component`, `src/components/Apps/` | **69 files / 908 tests passed** |
| new browser suite, 3 consecutive runs | 23/23 each time |

All re-run **after** rebasing onto current `origin/main`.

## What I could NOT verify

- **Nothing here is a visual claim.** The `component` project loads no stylesheet (no Mantine CSS, no Tailwind), so every assertion in the browser suite is structural — role, accessible name, attribute, `document.activeElement`. How the full-screen viewer actually *looks* — the overlay-arrow placement, the `object-fit: contain` sizing, the focus ring on the tile — is reasoning, not measurement. I did not run the app.
- **No real screenshot URLs were exercised.** Fixtures are inline SVG data URIs and an unresolvable host. Behaviour against actual CDN edge URLs (decode timing, very large images, portrait aspect ratios) is untested.
- **`loading="lazy"` makes below-the-fold error discovery non-deterministic** — measured, twice, while building the trailing-broken-run test: whether a second- or third-row tile is ever fetched depends on Chrome's distance-from-viewport heuristic, which earlier tests in the same file can move (a Mantine `Modal` locks body scroll). The viewer's rescue effect exists for exactly this, but it means "how many tiles have hidden themselves at time T" is not a deterministic property of a listing. Every broken-shot fixture keeps its dangling URL in the first grid row for this reason.
- **`preview / component-tests` is report-only**, so the 23 browser tests cannot block a future regression. That is why the arithmetic and the seam were extracted into the node tier — but the keyboard, focus and rendered-behaviour claims genuinely have no blocking gate.
- **Mobile/touch** — no swipe gesture, and the viewer was only exercised at 1440×900. `TrainingSampleViewer` has no swipe either; if touch-swipe is wanted, that is the argument for revisiting `Embla`.

## Open questions

1. **Stop vs wrap at the boundaries** — I chose stop (arrows go `disabled`), matching `TrainingSampleViewer`. `GeneratedOutputLightbox` wraps. Easy to flip if the product preference is the other way; it is one function and it is gated.
2. **Should the viewer show up in the moderator `preview` posture?** I kept it, reasoning above. It is a one-line gate if the answer is no.
3. **Swipe on touch** — deliberately out of scope; see above.

### DIFF
--- a/src/components/Apps/AppListingDetailBody.gallery.browser.test.tsx
+++ b/src/components/Apps/AppListingDetailBody.gallery.browser.test.tsx
@@ -131,7 +131,7 @@ beforeEach(async () => {
 
 /**
  * A screenshot URL that RESOLVES — an inline 320x180 SVG. It has to be a real,
- * loadable image: `ScreenshotTile` removes itself from the DOM on an `error` event, so
+ * loadable image: a tile is removed from the DOM on an `error` event, so
  * a tile built from an unreachable URL is not a tile whose width can be measured. The
  * 16:9 box keeps the rendered rows a sane height; only widths are asserted.
  */
@@ -139,7 +139,7 @@ const OK_SHOT = `data:image/svg+xml;base64,${btoa(
   '<svg xmlns="http://www.w3.org/2000/svg" width="320" height="180"/>'
 )}`;
 
-/** A URL that cannot resolve, so the browser fires `error` and the tile hides itself. */
+/** A URL that cannot resolve, so the browser fires `error` and the tile is dropped. */
 const BROKEN_SHOT = (n: number) => `https://cdn.invalid.example/missing-${n}.png`;
 
 function base(over: Partial<ListingDetail>): ListingDetail {
@@ -354,7 +354,7 @@ describe('screenshot gallery — rendered tile width', () => {
    * 🔴 THE COUNT-MISMATCH CASE — the reason the fix is a CSS structural selector and
    * not a `cols` value computed from `screenshots.length`.
    *
-   * Three screenshots go in; two of them 404, so `ScreenshotTile` removes them and the
+   * Three screenshots go in; two of them 404, so the gallery drops them and the
    * browser ends up with ONE tile. Any number derived before render would still say
    * "3" here and would leave that survivor at half width. `:last-child:nth-child(odd)`
    * is matched against the LIVE DOM, so the remaining tile is re-matched for free.
@@ -425,7 +425,7 @@ describe('screenshot gallery — rendered tile width', () => {
 
     // POSITIVE CONTROL, and not a racy one — see the sibling test above. Indexes 1 and
     // 3 with nothing between them is a string pair only a three-element map can
-    // produce, so the middle tile demonstrably rendered and then removed itself.
+    // produce, so the middle tile demonstrably rendered and was then dropped.
     expect(Array.from(grid.querySelectorAll('img')).map((img) => img.getAttribute('alt'))).toEqual([
       'My App screenshot 1',
       'My App screenshot 3',
--- a/src/components/Apps/AppListingDetailBody.module.scss
+++ b/src/components/Apps/AppListingDetailBody.module.scss
@@ -13,7 +13,7 @@
  * NOT `screenshots.length` and is not known at render time — a `cols` value
  * computed from the array would be wrong for exactly the listings whose art is
  * broken. `:last-child` / `:nth-child` are matched against the LIVE DOM, so when a
- * tile removes itself the remaining ones are re-matched for free: 3 shots of which
+ * tile is dropped the remaining ones are re-matched for free: 3 shots of which
  * the middle one 404s leaves two tiles that correctly stay half-width, and 3 of
  * which two 404 leaves one tile that correctly goes full-width. Pinned as measured
  * geometry in `AppListingDetailBody.gallery.browser.test.tsx`.
--- a/src/components/Apps/AppListingDetailBody.tsx
+++ b/src/components/Apps/AppListingDetailBody.tsx
@@ -396,12 +396,20 @@ function CreatorChip({ creator }: { creator: ListingDetail['creator'] }) {
  *
  * 🔴 IT IS A REAL `<button>` (`UnstyledButton`), NOT an `<img onClick>`. The tile has
  * to be tab-reachable and Enter/Space-activatable, and the accessible name has to
- * come from somewhere: it is COMPUTED FROM THE CONTENT (the image `alt` plus the
- * caption text) rather than pinned with an `aria-label`, so the name always contains
- * the caption the viewer can see — a bare `aria-label="View screenshot"` would strip
- * it and break WCAG 2.5.3 label-in-name. The button sits INSIDE the `Card` rather
- * than replacing it so the grid's children stay one element per tile, which is what
- * the gallery's fill rule (`:last-child:nth-child(odd)`) matches on.
+ * come from somewhere: it is COMPUTED FROM THE CONTENT rather than pinned with an
+ * `aria-label`, so the name always contains the caption the viewer can see — a bare
+ * `aria-label="View screenshot"` would strip it and break WCAG 2.5.3 label-in-name.
+ * The button sits INSIDE the `Card` rather than replacing it so the grid's children
+ * stay one element per tile, which is what the gallery's fill rule
+ * (`:last-child:nth-child(odd)`) matches on.
+ *
+ * 🔴 …AND THE IMAGE IS `alt=""` WHEN THERE IS A CAPTION, because the caption is
+ * rendered as TEXT inside the same button. Both contribute to the computed name, so
+ * naming the image with the caption too made the control announce "Shot three Shot
+ * three" — measured, and the reason `getByRole('button', { name: 'Shot three' })`
+ * matched nothing at all. An `alt=""` image beside its own visible caption is the
+ * standard pairing. The descriptive fallback stays for a captionless shot, where
+ * nothing else names the tile.
  */
 function ScreenshotTile({
   shot,
@@ -434,7 +442,7 @@ function ScreenshotTile({
         {/* eslint-disable-next-line @next/next/no-img-element */}
         <img
           src={shot.url}
-          alt={shot.caption ? shot.caption : `${name} screenshot ${index + 1}`}
+          alt={shot.caption ? '' : `${name} screenshot ${index + 1}`}
           loading="lazy"
           onError={onBroken}
           style={{ width: '100%', height: 'auto', display: 'block' }}
@@ -479,9 +487,11 @@ function ScreenshotTile({
  * LAST child and at an ODD position spans `1 / -1` — so the gallery uses the width
  * it actually has. See the stylesheet's header for why it is written as a
  * structural CSS selector rather than a count-aware `cols` value: `ScreenshotTile`
- * hides ITSELF on a load error, so the rendered tile count is discovered after
- * render and any number computed from `shots.length` here would be wrong for
- * exactly the listings with broken art.
+ * is DROPPED on a load error — it reports upward and the gallery filters it out —
+ * so the rendered tile count is discovered after render and any number computed
+ * from `shots.length` here would be wrong for exactly the listings with broken art.
+ * (It used to hide ITSELF, from its own `useState`; the DOM result is identical,
+ * but the decision moved to the gallery so the viewer skips the same shots.)
  *
  * 🔴 Do not "simplify" this to `cols={shots.length === 1 ? 1 : 2}`. That reads as
  * the same fix and is not: it fires on the ARRAY, so a 2-shot listing whose first
@@ -511,6 +521,29 @@ function ScreenshotGallery({
     (index: number) => setBroken((prev) => withBrokenIndex(prev, index)),
     []
   );
+
+  // 🔴 BOTH PIECES OF STATE ARE POSITIONS IN `shots`, SO A DIFFERENT `shots` MAKES
+  // THEM LIES. If the array changes under a mounted gallery — a `getAppDetail`
+  // invalidate, or the review preview swapping its local
+  // `buildListingDetailPreview(request)` for `previewQuery.data.detail`
+  // (`OffsiteReviewQueue`) — then "index 2 is broken" is a fact about screenshots that
+  // are no longer there, and it hides the WRONG tile in the new set. The viewer's
+  // rescue effect covers the open VIEWER; nothing covered the GRID.
+  //
+  // Keyed on the URL LIST rather than the array's identity, deliberately: `screenshots`
+  // is a fresh array on every refetch even when the content is identical, and resetting
+  // then would make already-failed tiles flash back in and re-request on every poll.
+  // The URLs are what the indices mean, so they are what has to change.
+  //
+  // Written as a render-phase adjustment (React's documented "adjusting state when a
+  // prop changes") rather than an effect, so the stale set is never rendered even once.
+  const urls = shots.map((s) => s.url).join('\n');
+  const [seenUrls, setSeenUrls] = useState(urls);
+  if (seenUrls !== urls) {
+    setSeenUrls(urls);
+    setBroken(NO_BROKEN_SCREENSHOTS);
+    setOpenIndex(null);
+  }
 
   if (shots.length === 0) return null;
 
--- a/src/components/Apps/AppListingDetailBody.tsx
+++ b/src/components/Apps/AppListingDetailBody.tsx
@@ -15,6 +15,7 @@ import {
   Stack,
   Text,
   Title,
+  UnstyledButton,
 } from '@mantine/core';
 import { useDisclosure } from '@mantine/hooks';
 import {
@@ -30,7 +31,7 @@ import {
 import type { Icon } from '@tabler/icons-react';
 import Link from 'next/link';
 import type { ReactNode } from 'react';
-import { useState } from 'react';
+import { useCallback, useState } from 'react';
 import { getEdgeUrl } from '~/client-utils/cf-images-utils';
 import {
   appInitial,
@@ -47,6 +48,12 @@ import {
 } from '~/components/Apps/appListingDetailView';
 import { toRecentAppFromListing } from '~/components/Apps/recentAppsRail';
 import { recordRecentlyOpenedApp } from '~/components/Apps/recentlyOpenedAppsStore';
+import { AppListingScreenshotViewer } from '~/components/Apps/AppListingScreenshotViewer';
+import {
+  NO_BROKEN_SCREENSHOTS,
+  withBrokenIndex,
+  type BrokenScreenshotIndexes,
+} from '~/components/Apps/appListingScreenshotNav';
 import { RelatedListings } from '~/components/Apps/RelatedListings';
 import { TruncatedText } from '~/components/Apps/AppListingTruncate';
 import { ListingCollaboratorByline } from '~/components/Apps/ListingCollaboratorByline';
@@ -375,43 +382,95 @@ function CreatorChip({ creator }: { creator: ListingDetail['creator'] }) {
 }
 
 /**
- * One screenshot tile — hides itself on load error (a dangling Image ref) rather
- * than rendering a broken `<img>`. Plain `<img>` (mirrors AppDetailsModal / the
- * live detail): the URL is a CDN edge URL, not a configured Next/Image domain.
+ * One screenshot tile — a BUTTON that opens the full-size viewer, and the reporter
+ * of a dangling Image ref. Plain `<img>` (mirrors AppDetailsModal / the live
+ * detail): the URL is a CDN edge URL, not a configured Next/Image domain.
+ *
+ * 🔴 THE `broken` STATE MOVED UP TO THE GALLERY, DELIBERATELY. It used to be this
+ * component's own `useState`, and hiding itself was all it ever did with it — which
+ * was enough while nothing else needed to know. It is not enough now: the viewer
+ * navigates over the same list, so a shot the GRID has already given up on must be
+ * one the VIEWER skips too, and one component's private state cannot say that. The
+ * rendered result is unchanged — the tile still disappears on a load error — but the
+ * decision is made once, by the owner of the list, instead of twice.
+ *
+ * 🔴 IT IS A REAL `<button>` (`UnstyledButton`), NOT an `<img onClick>`. The tile has
+ * to be tab-reachable and Enter/Space-activatable, and the accessible name has to
+ * come from somewhere: it is COMPUTED FROM THE CONTENT (the image `alt` plus the
+ * caption text) rather than pinned with an `aria-label`, so the name always contains
+ * the caption the viewer can see — a bare `aria-label="View screenshot"` would strip
+ * it and break WCAG 2.5.3 label-in-name. The button sits INSIDE the `Card` rather
+ * than replacing it so the grid's children stay one element per tile, which is what
+ * the gallery's fill rule (`:last-child:nth-child(odd)`) matches on.
  */
 function ScreenshotTile({
   shot,
   name,
   index,
+  onOpen,
+  onBroken,
 }: {
   shot: ListingGalleryScreenshot;
   name: string;
+  /** Index into the gallery's URL-filtered list — the viewer's index space. */
   index: number;
+  onOpen: () => void;
+  onBroken: () => void;
 }) {
-  const [broken, setBroken] = useState(false);
-  if (broken) return null;
   return (
     <Card withBorder padding={0} radius="md" style={{ overflow: 'hidden' }}>
-      {/* eslint-disable-next-line @next/next/no-img-element */}
-      <img
-        src={shot.url}
-        alt={shot.caption ? shot.caption : `${name} screenshot ${index + 1}`}
-        loading="lazy"
-        onError={() => setBroken(true)}
-        style={{ width: '100%', height: 'auto', display: 'block' }}
-      />
-      {shot.caption && (
-        <Text size="xs" c="dimmed" p="xs" lineClamp={2}>
-          {shot.caption}
-        </Text>
-      )}
+      <UnstyledButton
+        onClick={onOpen}
+        data-testid="apps-listing-screenshot-tile"
+        // The index this tile represents, as a DOM fact. A test that clicks "the
+        // third tile" and asserts the viewer opened on shot 3 is otherwise asserting
+        // against its own fixture order rather than against the wiring.
+        data-screenshot-index={index}
+        // Paired with a ring, never a bare `outline: none` — keyboard focus has to
+        // stay visible on a tile whose whole job is to be activated by keyboard.
+        className="focus:outline-none focus-visible:ring-2 focus-visible:ring-blue-5"
+        style={{ display: 'block', width: '100%', textAlign: 'left' }}
+      >
+        {/* eslint-disable-next-line @next/next/no-img-element */}
+        <img
+          src={shot.url}
+          alt={shot.caption ? shot.caption : `${name} screenshot ${index + 1}`}
+          loading="lazy"
+          onError={onBroken}
+          style={{ width: '100%', height: 'auto', display: 'block' }}
+        />
+        {shot.caption && (
+          <Text size="xs" c="dimmed" p="xs" lineClamp={2}>
+            {shot.caption}
+          </Text>
+        )}
+      </UnstyledButton>
     </Card>
   );
 }
 
 /**
  * Screenshot gallery — reuses AppDetailsModal's SimpleGrid pattern. Empty/broken
- * URLs are skipped; the whole section is hidden when nothing remains.
+ * URLs are skipped; the whole section is hidden when nothing remains. Each tile
+ * opens the full-size `AppListingScreenshotViewer` (prev/next, arrow keys, Esc,
+ * caption, `3 / 7` indicator).
+ *
+ * 🔴 THIS COMPONENT OWNS THE INDEX SPACE THAT BOTH SURFACES READ, and that is the
+ * whole reason the viewer is mounted HERE rather than fired at the app-wide
+ * `dialogStore`. `shots` (the URL-filtered list) and `broken` (the indices whose
+ * image 404'd) are single copies shared by the grid and the viewer, so the tile at
+ * `shots` index *i* opens the viewer at *i* with nothing in between to be off by
+ * one about. A viewer holding its own snapshot of either would diverge silently —
+ * see the rejected-alternatives ledger in `AppListingScreenshotViewer`'s header,
+ * and the seam gate in `__tests__/appListingScreenshotViewerWiring.test.ts`.
+ *
+ * 🔴 THE VIEWER IS **NOT** OMITTED IN `preview`, deliberately, and that is a
+ * decision against the grain of the posture ledger on this file's props. The
+ * ledger's rule is "omit every LIVE / interactive / AGGREGATE surface"; a lightbox
+ * is none of those — it queries nothing, writes nothing and acts on nothing. A
+ * moderator reviewing a shadow listing's MEDIA is precisely the reader who most
+ * needs to see a screenshot at full size, so keeping it serves the posture rather
+ * than bending it. The gallery itself was already on the KEPT side of that ledger.
  *
  * 🔴 THE GRID IS STILL TWO TRACKS; WHAT CHANGED IS WHAT AN UNPAIRED TILE DOES.
  * A fixed `cols={{ base: 1, sm: 2 }}` gave a single-screenshot listing HALF the
@@ -438,7 +497,31 @@ function ScreenshotGallery({
   name: string;
 }) {
   const shots = screenshots.filter((s) => !!s.url);
+
+  // 🔴 THE ONE INDEX SPACE, AND THE ONE PLACE THAT KNOWS WHICH SHOTS ARE REAL.
+  // Both pieces of state are indices into `shots` — the URL-filtered list above —
+  // never into the raw `screenshots` prop, and the tile at `shots` index *i* opens
+  // the viewer at *i* with no mapping in between. That is the whole defence against
+  // the off-by-one this feature invites: there is no second numbering to disagree
+  // with the first. `broken` lives here rather than in the tile because the viewer
+  // has to skip the same shots the grid has dropped (see `ScreenshotTile`).
+  const [broken, setBroken] = useState<BrokenScreenshotIndexes>(NO_BROKEN_SCREENSHOTS);
+  const [openIndex, setOpenIndex] = useState<number | null>(null);
+  const markBroken = useCallback(
+    (index: number) => setBroken((prev) => withBrokenIndex(prev, index)),
+    []
+  );
+
   if (shots.length === 0) return null;
+
+  // Rendered from the SAME array the viewer reads, minus what has failed to load, so
+  // a tile's `index` prop is its viewer index even after earlier tiles have dropped
+  // out. Mapping over the filtered-but-not-renumbered pairs is what keeps that true:
+  // renumbering here would silently shift every `alt` and every open target.
+  const visible = shots
+    .map((shot, index) => ({ shot, index }))
+    .filter(({ index }) => !broken.has(index));
+
   return (
     <Stack gap="xs">
       <Title order={4}>Screenshots</Title>
@@ -448,10 +531,26 @@ function ScreenshotGallery({
         className={galleryClasses.gallery}
         data-testid="apps-listing-screenshot-grid"
       >
-        {shots.map((shot, i) => (
-          <ScreenshotTile key={`${shot.url}-${i}`} shot={shot} name={name} index={i} />
+        {visible.map(({ shot, index }) => (
+          <ScreenshotTile
+            key={`${shot.url}-${index}`}
+            shot={shot}
+            name={name}
+            index={index}
+            onOpen={() => setOpenIndex(index)}
+            onBroken={() => markBroken(index)}
+          />
         ))}
       </SimpleGrid>
+      <AppListingScreenshotViewer
+        shots={shots}
+        name={name}
+        broken={broken}
+        index={openIndex}
+        onIndexChange={setOpenIndex}
+        onBroken={markBroken}
+        onClose={() => setOpenIndex(null)}
+      />
     </Stack>
   );
 }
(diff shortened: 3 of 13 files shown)


ZacxDev:
## The two red checks are both pre-existing — attributed, not assumed

Both are report-only, and neither is reachable from this diff (which touches only `src/components/Apps/`). I ran the control rather than reasoning about it.

### `preview / component-tests` — FAILURE

**3 failed / 1833 passed across 171 files.** All 3 are in `src/components/Account/PlacementSpaceSection.freeSlots.browser.test.tsx`:

```
× an unrelated save leaves a NULL count untouched          14978ms
× an unrelated save leaves a stored count untouched too    14950ms
× moving the control sends the number                      14984ms
```

**Control run — a clean `git worktree` at untouched `origin/main` (`a0f4a86fba`), own `node_modules`, nothing of mine present:**

```
× an unrelated save leaves a NULL count untouched          15296ms
× an unrelated save leaves a stored count untouched too    15272ms
× moving the control sends the number                      15396ms
 Test Files  1 failed (1)
      Tests  3 failed (3)
```

Identical three tests, identical failure. The spec was last touched by `13e7e39c52` (#4147); this PR does not import, render, or otherwise reach `PlacementSpaceSection`.

The ~15s per test is also the shape of a **failed assertion**, not load: a load flake inflates *every* test in the run, whereas here 1833 tests passed at normal speed and exactly 3 sat on an `expect` timeout.

All **69** browser test files under `src/components/Apps/` — **908 tests**, including this PR's 23 new ones — pass, locally and in that same CI run.

### `preview / smoke-tests` — FAILURE

Failures are `/moderator/reports` and `/moderator/images` redirecting to `/login` ("should not redirect to /login") — the preview moderator-session gate, plus a `TypeError: Cannot read properties of undefined (reading 'suggestedRebaseline')` in the reporting harness itself. Nothing in this diff touches a route, auth, or a moderator surface.

Sampled the 12 most recent PRs (`gh pr list --state all --limit 12`): of the 7 that ran the check, **6 are FAILURE** (#4162, #4160, #4156, #4154, #4152 and this one) and 1 is SUCCESS (#4149). So this is the current state of that check on `main`, not something this branch introduced. Stated as a **sample of 7**, not an enumeration.

### Everything else

17 of 19 checks green, including the blocking ones — `Unit tests`, `App unit tests`, `Package unit tests`, `tekton / typecheck`, `ESLint + Prettier (changed files)`, `event-engine-common pin`, `Schema drift gate`, `preview / deploy` and `preview / render-check`.

*(Aside worth recording: `pr-preview-4165-jxhpp-component-tests` reports `SUCCEEDED` as a TaskRun while the GitHub check reports FAILURE — the TaskRun condition tracks the step's exit code, which a report-only step returns 0 for by construction. Read the check and the `Tests N failed | M passed` line, never the TaskRun condition.)*

ZacxDev:
## Audit round addressed — 3 fixes + 2 doc corrections + the nits

Pushed as `25c0af2c79`, rebased onto current `main`. **Still do not merge.**

---

### 1. One Escape closing the viewer AND the review modal — FIXED

**Reproduced before fixing**, rather than taken on report. The test in its final shape went red first with exactly the audit's message:

```
AssertionError: one Escape closed BOTH dialogs — the moderator just lost their rejection reason
```

**Fix: `Modal.Stack` + `stackId`** (viewer) paired with a `Modal.Stack` wrapper on `OffsiteReviewModal`. Mantine then gates `closeOnEscape`/`trapFocus` on `ctx.currentId === stackId` rather than racing.

I checked the ordering alternative first and it is **provably unreliable**: `useWindowEvent` lists `[type, listener]` as deps and Mantine passes a *new inline arrow every render*, so each Modal tears down and re-adds its `window`-capture listener on every one of its renders — "who runs first" is whatever rendered most recently, so `stopImmediatePropagation` from the inner modal cannot be relied on. That is read off `use-window-event.mjs`, not assumed.

`stackId` is inert without a `Modal.Stack` ancestor (`stackProps` is applied only `if (ctx && stackId)`), so `/apps/store-preview/<slug>` is unchanged.

**Four tests, asserting the RELATIONSHIP, not a prop value** — per your instruction:
- one Escape → viewer closed **and** review modal still open;
- **control**: with no viewer open, Escape still closes the review modal (otherwise "disabled Escape everywhere" would pass);
- a *second* Escape then closes the review modal — Escape is scoped, not swallowed;
- the close button still closes only the viewer.

And the seam gate now pins **both halves** (`stackId` on the viewer, `<Modal.Stack>` + `stackId` in `OffsiteReviewQueue`), because losing either is silent. The old `closeOnEscape !== false` pin stays but its comment now says explicitly that it is the *mirror image* of this hazard and covers none of it.

🔴 **A side finding of my own, worth recording**: my `viewer()` helper was `document.querySelector('[role="dialog"][aria-modal="true"]')` — which identifies *a* Mantine modal, not *which one*. With two dialogs on screen it silently resolved to the **review** modal, so three of the four new tests failed for reasons that looked like product defects (one waited 10s for the wrong element). The suite had never had two dialogs, so nothing could have caught it — config-blind, exactly. The viewer now carries `data-testid="apps-listing-screenshot-viewer"` and the helper anchors on it.

### 2. Rescue effect coverage — FIXED

Driven as an ordinary controlled component, as you suggested. **7 tests**: forward preference when both directions are viable (the only fixture shape where swapping the `??` arms is visible), the backward fallback, the out-of-range case, both `onClose` arms, and **two negative controls** (no fire for a viable shot; no fire while closed) — without which every other assertion is satisfied by an effect that fires on everything.

Four mutants, all killed on their own assertion: effect disabled, `??` order swapped, `onClose` arm removed, guard's `|| viable` removed.

### 3. `broken` never reset on a new `screenshots` — FIXED

Reset keyed on the **URL list**, not the array identity, as a render-phase adjustment so the stale set is never rendered even once. `openIndex` is reset with it.

🔴 **The mutation sweep found my first control was vacuous.** The equal-array control asserted only the tile *count*, and an identity-keyed reset **SURVIVED** it: the reset does fire, but the restored tile re-errors within a frame and the count **converges** back before any assertion runs. An open viewer does not converge — a closed lightbox stays closed — so the control now opens the viewer first and asserts it is untouched. That mutant is killed now.

### Doc corrections

- **The `styles` block claim** — corrected in *both* places it appeared (the top-of-file reuse ledger and the inline comment; I initially fixed only the inline one). It now says the three lightboxes share `inner`/`content`/`body`, that all three *also* set an absolute `header`, and that this viewer deliberately does not because it renders a `title` that must stay in flow.
- **"hides ITSELF"** — corrected at all five sites (`AppListingDetailBody.tsx`, `.module.scss`, the gallery browser test ×3 lines, and both `WHY_` strings in `appListingGalleryFill.test.ts`). DOM and CSS unchanged; only the explanation was false.

### Nits — all taken

- **Duplicated accessible name**: `alt=""` when a caption exists. Asserted through the accessibility tree (`getByRole('button', { name: 'Shot three', exact: true })`), not by reading the attribute — the whole point is that the attribute-level assertion passed while the name was wrong. A control test covers the captionless tile, which still gets its `alt` fallback.
- **Dead single-child `Group`** removed.
- **"Loading screenshot…"** removed — it described a state that is actually broken-or-gone, and sat beside a `0 / N` indicator counting a shot nobody could see. Both are now simply absent for a non-viable shot (a one-frame state the rescue clears).

### Not chased

Finding 10 (transient in-viewer 404) — noted, not chased, per your call.

---

## Verification — FULL battery re-run, not just the delta

**31 mutants, 31 KILLED, 0 SURVIVED, 0 invalid**, each scored against its own designated test and each reporting `1 failed | 0 passed` (that test, and only it, went red). All 29 `-t` patterns positive-controlled against the unmutated tree first. Two rounds were needed — round 1 returned one ANCHOR-MISMATCH (drifted after the restructure) and one genuine SURVIVOR (the converging control above).

New mutants this round: viewer leaves the stack · review modal leaves the stack (source gate only — the browser harness declares its own outer `stackId`, so it is blind to that half **by construction**, which is precisely the seam) · rescue disabled · rescue direction swapped · rescue never closes · rescue fires for a viable shot · reset removed · reset keyed on count · reset keyed on array identity · caption duplicated into `alt`.

**On the direction-swap mutant you asked about**: `M5-prev-next-swapped` is a swap of the `-1`/`+1` arguments *inside the viewer component*, so the node tier structurally cannot see it — it is not in the pure module. It is **consciously left to the browser tier**, killed there by `next advances by exactly one, and prev goes back by exactly one`. I chose not to add a source-text pin for it: a gate asserting `prevIndex = adjacentViableIndex(index, -1` would catch this exact edit but not the failure that actually matters — an arrow wired to the *wrong variable* — so it would pin half a claim while reading like the whole one. That is the "spelled rather than structural" shape, and this PR already has one guard that certified the wrong half of a hazard.

| Check | Result |
|---|---|
| `pnpm typecheck` | 0 errors (post-rebase) |
| node `unit`, `src/components/Apps/__tests__/` | **54 files / 998 tests** |
| browser `component`, `src/components/Apps/` | **69 files / 923 tests** |
| new viewer suite | **38 tests** (was 23) |
| `prettier --check`, added files | clean |
| `eslint`, 9 changed ts/tsx | 0 errors, 2 warnings — both pre-existing unused types in `OffsiteReviewQueue`, verified present identically at HEAD |

🔴 **One thing I deliberately did NOT do**: `prettier --write` on `OffsiteReviewQueue.tsx` produced a **290/263-line** diff, because that file was already not prettier-clean at HEAD. I reverted it and re-applied only the two intended edits by anchored replacement (now 60/35, almost all of it the docstring plus re-indenting the wrapped block). CI's gate errors on **added** files and only warns on **modified** ones (`lint.yml:89-90,137,165`), so this stays a warning it already had.

## Still could not verify

- Everything visual — the `component` project loads no stylesheet, and I did not run the app. The nav arrows' `absolute` positioning is one to watch: restructuring the viewer briefly moved them outside their `relative` container, which no test here can see (I caught it by reading the diff, and there is now a comment at the site).
- I did not adopt the `chromium_headless_shell-1228` recipe — my runner already works via `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/run/current-system/sw/bin/brave`, and switching mid-round would have changed the instrument between mutation rounds. Noted for next time; the gap it closed for you was the nested-modal probe, which I have now run.
- The `{shot && …}` guard on the position indicator is a cosmetic honesty fix with **no mutation coverage** — it is a one-frame state I could not make deterministically observable. Stated rather than papered over.

ZacxDev:
## CI after the audit round — `component-tests` now fully green

`25c0af2c79`. **19 checks, 17 SUCCESS, 1 SKIPPED (by design), 1 FAILURE.**

### `preview / component-tests` — SUCCESS (was FAILURE)

```
Test Files  171 passed (171)
     Tests  1853 passed (1853)
```

Zero failures. The three `PlacementSpaceSection.freeSlots` failures I attributed to `main` last round are gone — fixed by **#4168**, which landed on `main` in the meantime and which this branch is now rebased onto. That is the control arriving on its own: I claimed the failures were pre-existing and not mine, and the fix for them was in someone else's PR to a file I never touched.

All **38** tests of this feature's browser suite are inside that run.

### `preview / smoke-tests` — FAILURE, unchanged and unrelated

Same signature as last round, verbatim:

```
Error: /moderator/reports: should not redirect to /login
```

The preview moderator-session gate. Nothing in this diff touches a route, auth, or a moderator surface — the `OffsiteReviewQueue` change is a `Modal.Stack` wrapper and a docstring. Previously measured as FAILURE on 6 of the 7 most recent PRs that ran the check.

### Everything else

`Unit tests`, `App unit tests`, `Package unit tests`, `tekton / typecheck`, `ESLint + Prettier (changed files)`, `event-engine-common pin`, `Schema drift gate`, `preview / deploy`, `preview / render-check`, `preview / unit-tests`, `preview / apps-tests`, `preview / auth-tests`, both `Windows dev-env` jobs — all green.

🔴 **Method note, because it nearly bit me:** the rollup reached 18 checks all-terminal and all-green *before* `preview / smoke-tests` had registered at all — this pipeline emits `smoke-test` and `lighthouse` after `verify-preview` succeeds. A poll that required only "≥16 checks and none busy" would have declared a clean sweep and been wrong about the one check that is red. The previous run's count (19) is what caught it; I waited for smoke specifically rather than trusting the terminal-state sweep.

**Still not merged.**

ZacxDev:
## Round 3 addressed — `efdc89de573a6210d724ba5eb272d4c70e1d1b84`, rebased onto `124e256a44`

**Still not merged.**

---

### 🟡-1 — the `Modal.Stack` focus-return regression: FIXED, and it was worse than reported

Reproduced independently before fixing, with your two-arm control plus a third arm:

| arm | focus after close |
|---|---|
| without `Modal.Stack` | the trigger ✅ |
| with `Modal.Stack` | `<body>` ❌ |
| with `Modal.Stack` + `returnFocus={false}` | `<body>` ❌ |

The third arm is why turning Mantine's version off is not itself a fix. Replaced with **`useOpenerFocusReturn`**, which captures the opener **during render** — an effect is too late, because React runs the modal's own focus-trap effect (a descendant) first.

🔴 **A second mechanism the audit didn't name, and it broke the nested viewer too.** `useFocusTrap`'s effect re-focuses **unconditionally** when `active` flips true (`ref.current && setTimeout(() => focusNode(ref.current))` — no "is focus already inside?" check). So when the viewer closes, the outer modal's re-armed trap steals focus back *after* any synchronous restore. No ordering trick can win: that timeout is scheduled from an **ancestor** effect, which React runs after ours. Fixed by aiming Mantine's own preference — `focusNode` prefers `[data-autofocus]` — at the tile the viewer was opened from. Measured, not reasoned.

🔴 **And a second HOST nobody had looked at.** `CombinedReviewModal` also renders `OffsiteReviewModalBody` — and therefore the viewer — inside its own `<Modal>`, so this PR created the identical Escape defect there and round 2 fixed only one host. Both now join a stack and use the focus pair, and the seam gate checks **both**, as a proximity match rather than two independent substrings (your 🟢-8).

### 🟡-2 — convergence: FIXED

Same-length swap (3 → 3). With `setOpenIndex(null)` deleted the viewer now stays open showing the new list's third shot, exactly as you measured. Header records *why* the fixture length is load-bearing so it cannot be "simplified" back.

### 🟡-3 — the `body` claim: measured this time

| key | TrainingSampleViewer | GeneratedOutputLightbox | Model3DLightbox | this viewer |
|---|---|---|---|---|
| `inner` | `{position:'absolute'}` | same | same | same |
| `content` | flex column, `overflow:hidden` | same | same | same |
| `header` | `{position:'absolute',right:0,zIndex:10}` | same | same | **absent, deliberately** |
| `body` | `{flex:1,minHeight:0,display:'flex',flexDirection:'column',padding:16}` | `overflow:'hidden'` variant | `overflow:'hidden'` variant | **matches TrainingSampleViewer only** |

Written as "one of three", with the history noted so the next reader knows this line has been wrong twice.

### 🟡-4 — the viewer `alt`: covered, and the three unclaimed changes disclosed

Covered in **both** directions (empty when a caption is beside it; descriptive when there is none). Both of your mutants now die.

**The three changes that were in the commit but not in my claims** — disclosed now, as you asked:
1. the viewer image's `alt` became conditional (this one);
2. the **"Loading screenshot…" placeholder was removed** — it described a state that is actually broken-or-gone, not loading;
3. the **position indicator was gated** on there being a viable shot — it read `0 / N` beside that placeholder.

(2) and (3) are cosmetic-honesty fixes. 🔴 **(3) has no mutation coverage** — it is a one-frame state I could not make deterministically observable. Stated rather than papered over.

---

## 🔴 One of my own claims was refuted by the sweep, and is now corrected in five places

I wrote that "the hook without `returnFocus={false}` loses a race with Mantine's 10ms post-close timeout — neither half works alone." That was **reasoning**. Deleting the prop while keeping the hook left every focus test **green**; only the source gate noticed. Corrected in `useOpenerFocusReturn.ts`, `AppListingScreenshotViewer.tsx` (×2), `OffsiteReviewQueue.tsx`, the focus test's header and the gate's message: the pair is kept as a **single-owner convention**, and is now documented as exactly that rather than as a measured pair.

## Battery — 41 mutants, 41 KILLED, 0 SURVIVED, re-run in full after the rebase

Not just the delta. All 36 `-t` patterns positive-controlled against the unmutated tree first.

Round-3 mutants (from the **spec claims**, not from the assertions — your methodology note): review modal loses the hook · review modal keeps Mantine's return-focus · viewer loses the hook · **focus capture moved into an effect** · tile loses `data-autofocus` · `data-autofocus` keyed on `openIndex` instead of last-opened · combined modal loses its `stackId` · combined modal loses the focus pair · `setOpenIndex(null)` removed · viewer `alt` duplicated · viewer `alt` always empty.

Two harness lessons from this round, both now encoded:
- A mutant **wedged the browser** for 24 minutes rather than failing. The harness now bounds each run at 240s and reports `TIMED-OUT` as its own verdict — never folded into KILLED, because "the suite hung" and "the assertion failed" are different facts. (That mutant's pathology is itself the evidence for the `lastOpenedIndex` design, and is recorded at the site.)
- Four mutants came back `HARNESS-BROKEN` under load and were **KILLED on retry** — the guard refused to score them, which is the behaviour I want.

## 🔴 A pre-existing failure on `main`, attributed with a control — not mine

Running the **whole** `src/components/Apps/` browser directory in parallel dies before any test executes:

```
Unhandled Rejection
Error: [vitest] There was an error when mocking a module … no top level variables inside …
```

**Control: it reproduces on a clean worktree at `origin/main` (`124e256a44`) with none of my code.** It also survives removing my new test file. A serial run (`--no-file-parallelism`) got 12 files in with no error before I stopped it for time, so it looks parallelism/load dependent — which is consistent with CI's component-tests having passed 171 files / 1853 tests on my previous push. Reporting it, not chasing it.

🔴 **But it did surface a real defect of mine, which is fixed:** my new focus test's `vi.mock` factory closed over a top-level `const`, which `vi.mock` hoists above — a TDZ read that produced *no summary line at all* and wedged the run under load. Moved into `vi.hoisted`. That would have wedged CI's component task. (A neighbouring suite gets away with the plain form; that is evaluation-order luck, not a pattern to copy, and the file says so.)

## Verification on the rebased tree

| Check | Result |
|---|---|
| head | `efdc89de573a6210d724ba5eb272d4c70e1d1b84` on `124e256a44` |
| node `unit`, `src/components/Apps/__tests__/` | **55 files / 1025 tests** |
| the previously-red `ModelFileHash` ledger specs | **2 files / 11 tests passed** — #4192 confirmed by running them, not by reading a badge |
| `AppListingDetailBody` viewer + gallery + body suites | **3 files / 112 tests** |
| `OffsiteReviewModal.focus.browser.test.tsx` | 2 tests, 785ms |
| mutation battery | **41/41 killed** |
| `prettier --check`, added files | clean (CI errors on added, warns on modified) |

`packages/civitai-db-schema/src/enums.ts` is regenerated by `pnpm install` (line-wrapping only); restored, and absent from all three commits.

## Not chased, per your call

🟢-7: a **sixth** "hides ITSELF" site survives, as a **test name** in `AppListingDetailBody.browser.test.tsx` — a file this PR does not touch, so it is out of the "fix the claims in the code you touch" scope. My earlier "all five sites" should be read as *all five in the files this PR touches*. 🟢-5 (reset ordering, M16) and 🟢-6 (indicator gating, M22) remain uncovered as stated. Finding 10 remains unmeasured.

ZacxDev:
## CI green — `bca1b003b2477a70b9542bda1e00433684508e15`

**19 checks: 18 SUCCESS, 1 SKIPPED (by design), 0 FAILURE.** Both PipelineRuns reached a terminal state, not merely a rollup with no pending rows:

```
pr-check-4165-9ln7d    : True Completed  2026-08-20T19:06:20Z
pr-preview-4165-fff8s  : True Completed  2026-08-20T19:21:22Z
```

`preview / component-tests`, read from the log rather than the badge:

```
Test Files  172 passed (172)
     Tests  1902 passed (1902)
```

`preview / smoke-tests` is **SUCCESS** — the ambient `MODERATOR_APP_URL` defect is gone, as expected.

### One red I caused on the first push of this round, and the instrument gap behind it

`ESLint + Prettier (changed files)` failed on `appListingScreenshotViewerWiring.test.ts`.

🔴 **My local check was measuring the wrong set.** I had been running `prettier --check` over the files added *in the most recent commit*; CI selects files added *across the whole PR* (`git diff --diff-filter=A FETCH_HEAD...HEAD`). A file added in commit 1 and edited in commit 3 is in CI's set and was not in mine — so a green local check and a red CI check were both correct about different questions. Now computed with CI's exact command:

```
git diff --name-only --diff-filter=A origin/main...HEAD -- '*.ts' '*.tsx'
```

7 added files, all clean. Also swept ESLint over the full changed set (12 files): **0 errors, 0 warnings** after removing a leftover unused binding from my `vi.hoisted` refactor. (The two remaining prettier warnings are on MODIFIED files — `OffsiteReviewQueue.tsx` and `CombinedReviewModal.tsx` — both verified **already unformatted at `origin/main`**, and CI warns rather than errors on those. Formatting them would add ~360 lines of unrelated churn to a moderator-review surface this PR barely touches.)

**Do not merge** — awaiting your call.

ZacxDev:
## Round 4 — three changes, no full battery

`7af5748e180cb967a4dc19490b6f3294fced32bf`. Three edits across two files. **I did not run the full mutation battery for this round** — I ran the affected suites plus one targeted mutation, as you said was sufficient. Saying so rather than implying otherwise.

### 🟡 A — the guard now has a killing test

One `{ArrowRight}` added to `focus returns to the ORIGINATING tile on close`, so the viewer is *used* before it is closed rather than opened-then-immediately-closed.

Verified both ways, on the shipped code:

| | result |
|---|---|
| HEAD | 43/43 green |
| guard mutated to `if (opened)` | **1 failed** — `AssertionError: expected <body …> to be <button …>` |

That is the mutant's own assertion, and the hook file's hash was confirmed identical after restore. The reachability argument is now written at both sites (the hook's header and the test's), including the instruction not to remove the keystroke to "simplify" the test.

You were right that my 41/41 battery did not imagine this one. The reason is worth naming: every focus test I wrote opened and closed in one breath, so *the battery inherited the blind spot from the fixtures rather than from the mutant list* — the "derived from your own assertions" failure mode showing up one level down, in the fixture shape rather than the mutation shape.

### 🟡 B — the header claim rewritten to what is measurable

Confirmed against the source myself before rewriting: `use-focus-trap.mjs` wraps `focusNode` in `setTimeout` in **both** the ref callback and the effect, so the focus move is a macrotask that necessarily runs after every effect in the commit. **Effect ordering was never the reason** — my "child effects run before parent effects, so an effect is too late" was reasoning presented as a 🔴 fact.

The render-phase capture stays, with a justification that is actually true: at that render the commit's DOM mutations have not happened, so `activeElement` is unambiguously the opener and that holds unconditionally — whereas an effect-phase read is correct *today* only because Mantine defers its move, which is a property of Mantine's implementation and not a guarantee. Earlier read because it cannot be wrong, not because the later one doesn't work.

🔴 **On where this landed:** it is the same class the commit creating the file said it had purged, in the file created to say so. The lesson I take is narrower and less flattering than "check your claims": the false claim was *load-bearing for a design decision*, so writing it down made the decision feel settled and stopped me testing the alternative. The rewrite now states the alternative works, which is what makes the remaining justification checkable.

### 🟢 D — SSR guard added

`typeof document === 'undefined'` before the render-phase read. Unreachable today (all three call sites pass `false` on first render) and the comment says so, but `/apps/review` and `/apps/store-preview/[slug]` are server-rendered and a dialog seeded open from props or a query param would be a server-side `ReferenceError`.

---

## Recorded, not fixed

- **🟡 C — corrected scope.** `CombinedReviewModal` has **no behavioural coverage**: its Escape scoping and focus return are pinned by the source gate plus `NestedReviewHarness`, which is a hand-built replica, not either real host. The correct phrasing is **"one behavioural point plus a text check"**, not "both hosts covered" — my earlier wording overstated it and I'm retracting it here.
- **🟢 E** — deleting `setLastOpenedIndex(null)` from the URL-list reset survives; reachable only via list-swap → nested close.
- **🟢 F** — two survivors are equivalent mutants (dropping `opener?.isConnected`, dropping `openerRef.current = null`).
- **🟢 G** — in the nested flow the viewer's own hook is redundant; `data-autofocus` alone delivers the tile. Both mechanisms are therefore load-bearing only in *different* flows — worth knowing before either is touched again.

## Verification

| Check | Result |
|---|---|
| viewer + focus suites | 43 tests |
| all four affected browser suites | **4 files / 114 tests** |
| node `unit`, `src/components/Apps/__tests__/` | **55 files / 1025 tests** |
| targeted mutation on the A guard | KILLED on its own assertion |
| `eslint`, CI's changed set | 12 files, 0 errors, 2 warnings (both pre-existing in `OffsiteReviewQueue`) |
| `prettier --check`, CI's added set | 7 files, clean |

🔴 **One instrument failure worth recording, because it nearly shipped the same red twice.** My first prettier/eslint sweep this round passed `$ADDED` unquoted-but-unsplit — zsh does not word-split, so the whole seven-file list went in as ONE argument. Prettier printed `[error] No files matching the pattern were found: "<all seven joined>"` and then **`All matched files use Prettier code style!`** — a clean verdict over zero files. Re-run with `${=ADDED}` it immediately found `useOpenerFocusReturn.ts` unformatted, which is exactly what failed CI last round. The count is now printed and read (`added count: 7`, `changed count: 12`) rather than trusting the banner.

PR #4172, opened 2026-08-20
ci: typecheck the apps/ workspace — all seven were outside every typecheck we run

ZacxDev:
No app in `apps/` has ever been typechecked by CI.

The root `pnpm run typecheck` — the script **both** the Actions `typecheck` job and the `tekton / typecheck` status run — is bounded by the root tsconfig `include`:

```
["scripts/local-dev/*.ts", "src", "packages/*/src", "tests", "test", ".next/types/**/*.ts"]
```

There is no `apps/` entry. Because both tiers run the same script, this was never a lag window that would close on its own.

## What this adds

An `app-typecheck` matrix job, one leg per app, blocking, `fail-fast: false` so one break cannot hide the others.

It carries **no `if:` gate** — unlike the existing `typecheck` job, which is narrowed to fork PRs and non-`main` bases. Putting this coverage there would have switched it off on every internal PR.

## Verified, not inferred

A bare "0 errors" from an instrument nobody has watched go red is not a measurement, so each leg ran twice locally — once on the unmodified tree, once with a deliberate `const zz: number = "..."` planted under the app:

| app | unmodified | with planted error |
|---|---|---|
| `event-engine` | 0 | 1 × `error TS2322` |
| `notifications` | 0 | 1 × `error TS2322` |
| `orchestrator-gateway` | 0 | 1 × `error TS2322` |
| `storage` | 0 | 1 × `error TS2322` |
| `creator-studio` | 0 errors, 0 warnings | — |

## `apps/auth` is deliberately excluded

It is the one app that does not pass today:

```
src/lib/server/auth/providers.ts:165           Parameter 'p' implicitly has an 'any' type
src/lib/server/auth/__tests__/establish-session.test.ts:100
                                               Property '_store' does not exist on type 'never'
```

Both reproduced independently of the original report. Wiring it now would ship a permanently-red gate, which trains people to click through the other five results too. It needs its own fix, then one line in the matrix.

## Svelte apps run `check`, not `typecheck`

Their `typecheck` is a bare `svelte-check`, but `.svelte-kit/tsconfig.json` does not exist in a fresh checkout — `check` is `svelte-kit sync && svelte-check`, and the sync is what generates it.

## One honest gap

`moderator` could **not** be confirmed locally. It reports 131 errors in my checkout, but every one descends from `Cannot find module '@civitai/redis' / 'kysely' / '@civitai/mod-utils/prompt-audit'` — unlinked workspace deps, not type defects. A clean `pnpm install --frozen-lockfile` on the runner is exactly what my environment lacked, so **this PR's own run is the measurement**.

If that leg comes back red, the fix is to split `moderator` out the way `auth` is — not to weaken the job.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -401,3 +401,75 @@ jobs:
       - name: Assert every app suite actually ran
         if: always()
         run: node scripts/ci/assert-workspace-suites-ran.mjs apps-report.json apps
+
+  app-typecheck:
+    name: App typecheck (${{ matrix.app }})
+    runs-on: ubuntu-latest
+    # NOT ONE of the seven apps/ members was typechecked by anything until this job.
+    #
+    # The root `pnpm run typecheck` — the script BOTH the `typecheck` job above and the
+    # in-cluster `tekton / typecheck` status run — is bounded by the root tsconfig's
+    # `include`, which lists `scripts/local-dev/*.ts`, `src`, `packages/*/src`, `tests`,
+    # `test` and `.next/types`. There is no `apps/` entry, so every app is outside the
+    # only typecheck this repo had. Both tiers run the same script, so this was not a lag
+    # window to wait out: it could not close on its own.
+    #
+    # Proven with a planted error rather than inferred from the config, on 2026-08-19:
+    # a deliberate `const zz: number = "..."` under each app reported exactly one
+    # `error TS2322`, while the same invocation on the unmodified tree reported zero.
+    # A zero from an instrument nobody has watched go red is not a measurement.
+    #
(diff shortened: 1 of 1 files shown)


ZacxDev:
Closing as a duplicate of #4156, which does the same job and was opened first. My mistake: the ClickUp ticket said "No PR was opened" and I acted on that without checking whether it was still true.

Findings from my independent re-measurement are preserved as a comment on #4156 — including the positive-control method, the `moderator` false alarm (131 local errors, all unlinked workspace deps), and a note that the Svelte apps' bare `typecheck` is safe only because `"prepare": "svelte-kit sync"` runs during install.

#4156 is the one to review.

PR #2428, opened 2026-06-07
perf(api): per-pod concurrency bulkhead + fast-fail on heavy image routes [DRAFT]

ZacxDev:
**DRAFT for review** — lever (B) of the pin/504 stabilization plan. Pairs with (D) probe-tolerance (infra) and (A) heavy-route isolation.

## Why
Root cause (verified on the 2026-06-07 wave: 504 peaked 152/s, 11 pods pinned, readyReplicas 90→79, HPA didn't scale): a heavy request blocks a pod's **single JS thread** for seconds; under a surge the backlog pins the event loop → the pod can't answer probes or its `/metrics` scrape → Error/137 SIGKILL + 504s + a **self-amplifying cascade** as load shifts onto survivors. Autoscaling can't help — both HPA metrics are scrape-blind to pinned pods, and it's minutes-slow vs ~2-3min bursts. Per-request micro-opts (−6% cpu/req) didn't stop it.

## What this does
Caps **concurrent in-flight heavy image requests per pod** and **fast-fails the excess** instead of letting an unbounded backlog form. A quick retryable reject (429 tRPC / 503+`Retry-After` REST) is strictly better than a 504 after a 30s timeout on a pod that's about to die. It keeps pods **alive and serving what they admit**, and gives clients backpressure.

- `request-bulkhead.ts` — process-local concurrency gate (`acquireBulkheadSlot`→release; `BulkheadFullError` when full; reject counter). Atomic on Node's single thread.
- `trpc.ts` — `heavyProcedure` = `publicProcedure` + `withBulkhead('heavy-image')`.
- `image.router.ts` — `image.getInfinite` + `image.getImagesAsPostsInfinite` → `heavyProcedure`.
- `/api/v1/images` — same `'heavy-image'` key (one shared per-pod budget; they hit the same heavy path) → 503 on full, slot released on response `close`.

## Tuning
`HEAVY_REQUEST_CONCURRENCY` (default **20**) is a **safety valve**, sized NOT to shed under normal peak — it only trips on a pathological pile-up. Worst-case drain at 20 × ~2s ≈ 40s < the 120s liveness window, so it caps the kill risk. Tune **down** if pins persist, **up** if it sheds real traffic.

## ⚠️ Still draft — needs before un-drafting
[...]

PR #4173, opened 2026-08-20
fix(metrics): bulkhead + pg-pool gauges emitted zero series — a globalThis flag guarding a per-graph prom registry

ZacxDev:
## Symptom

`civitai_app_heavy_bulkhead_active` and `civitai_app_heavy_bulkhead_rejects` have produced **zero series** since they shipped in #2428 on 2026-06-07 — **74 days**. Every other link in the chain was healthy, which is why it went unnoticed: registered in code, merged to `release`, present in the running image, on a hot route, on hosts emitting hundreds of thousands of other `civitai_app_*` series.

The nine `node_postgres_*` pool gauges were dead the same way, for longer.

## Root cause — two defects, and fixing either alone leaves the metric inert

**1. The registry.** Next.js compiles `instrumentation.ts` into a **separate webpack bundle** from the pages/API bundle, and `prom-client` is not in `serverExternalPackages`, so each graph gets its own `client.register`. `/api/metrics` scrapes the *pages* graph's default registry plus the globalThis-pinned `instrumentationRegistry`. The instrumentation graph's default registry is scraped by nobody.

The gauges were wrapped in `if (!global.heavyBulkheadGaugeInitialized)` — a **process-scoped flag guarding a graph-scoped registry**. That is worse than no guard at all, because it fails in exactly one direction: `instrumentation.node.ts` → `~/server/eventloop-longtask` → `~/server/prom/client` runs at pod start, claims the flag, registers into its own unscraped registry, and the pages graph then takes the early-out and registers **nothing** into the registry that is actually served.

They now go through `registerInstrumentationMetric`, which dedupes against the same shared registry it writes to — so the two scopes agree, and whichever graph arrives first, the metric is scraped.

**2. The state.** Both gauges are `collect()`-based over `bulkheadSnapshot()`, and `request-bulkhead`'s slot/reject `Map`s were module-local. So whichever graph won registration, its `collect()` closure read its **own permanently empty maps**. They are now pinned on `globalThis`, the same mechanism `instrumentationRegistry` and `__civitaiRedisMetrics` already use.

Note this also makes the cap genuinely **per-pod** rather than per-pod-per-graph, which is what `HEAVY_REQUEST_CONCURRENCY` was always documented to mean. Admission control was already correct in the graph that serves requests; only a reader in another graph saw nothing.

## Evidence

Measured before the fix. These four groups are what isolate the **flag** as the variable, rather than the registry choice or the graph:

| metric | registration | series |
|---|---|---|
| `civitai_app_heavy_bulkhead_{active,rejects}` | default registry, behind a globalThis flag | **0** |
| `node_postgres_*` (9 gauges) | default registry, behind a globalThis flag | **0** |
| `civitai_app_image_ingestion_backlog` | shared `instrumentationRegistry` | 640 |
| `images_search_*` (same request path) | default registry, **no** flag | 181 |

The last row also settles a question raised when this was filed: the per-route request-duration histogram from `src/pages/api/v1/images/index.ts` is **not** missing — it is `images_search_request_duration_seconds`, carries no `civitai_app_` prefix, and reports 157 series on `route="api/v1/images"`. It has no globalThis flag, which is precisely why it survived. Different symptom, same lesson.

## Tests

8 new cases in `src/server/__tests__/prom-cross-graph-registration.test.ts`. `vi.resetModules()` between two dynamic imports is the closest a single process gets to the two-graph split: a fresh module instance with globalThis intact — the exact asymmetry the bug lived in.

- **8/8 RED at `origin/main`**, 8/8 green here.
- With **only** defect 1 fixed: 3 pass and the two seam cases still fail — the gauges register, get scraped, and render nothing. That is what proves the `request-bulkhead` change is load-bearing rather than tidy-up, and that the obvious one-file fix would have shipped inert.
- `pnpm typecheck` 0 errors; instrument positive-controlled (a planted type error reported 1).
- 74 files / 1286 tests green across the suites touching these modules; `test:lint-rules` 243/243.

## Also in here

- The `@civitai/telemetry` comment that cited *"the pgGaugeInitialized guards"* as the working precedent for pinning on globalThis. They pinned the **flag**, not the thing it guarded, and every metric behind them was dead — so the comment was recommending the bug.
- The pg pool gauges, because they are the same defect in the same file. Called out rather than folded in silently.

## Not in here

The Grafana panel's `or vector(0)`, which turned this absence into a **drawn zero line** — a panel titled "reject rate" reporting "no rejects, ever" for 74 days. That lives in the infra repo and is already fixed there.

## Follow-up worth considering (not done)

A quiet pod still emits no series until it first acquires a slot, so a healthy fleet reads as `No data`. Seeding the `heavy-image` key at 0 would make a real zero observable — the pattern `/api/metrics` already applies to the generation-substitution and bitdex counters. Left out because the panel description documents the current behaviour as intended; worth a decision rather than a silent change.


ZacxDev:
## Verified on a deployed build, as an A/B against base code

Both halves were run in a PR preview environment — a real production Next.js build with the two webpack graphs, which is what the bug depends on. Same probe, same pipeline; the only variable is the commit.

| | preview of a base-code PR | preview of this PR |
|---|---|---|
| `/api/metrics` | 200 | 200 |
| control: `images_search_request_duration_seconds_count` | 1 | 1 |
| control: `node_postgres_pool_acquire_duration_seconds_count` (sibling registered outside the guard) | 3 | 1 |
| `node_postgres_*` (was behind the guard) | **0** | **18** |
| `civitai_app_heavy_bulkhead_*` | **0** | **2** |

18 is exactly right for the pg family: 6 unlabelled + 3 labelled × 4 pools. The controls are there so a row of zeros cannot be mistaken for a working probe — an earlier run of this same command returned 0 across the board and turned out to be a 401, which is precisely the failure the controls exist to catch.

### The gauges move, they do not merely exist

Two 40-concurrent bursts against `/api/v1/images`, at `HEAVY_REQUEST_CONCURRENCY=20` read from the pod's own startup log rather than assumed:

- **burst 1 (cold)** — `{"200": 40}`, rejects **0**. Cold-start serialisation kept fewer than 20 in flight, so nothing shed. This is the negative arm: traffic alone does not move the counter.
- **burst 2 (warm)** — `{"200": 20, "503": 20}`. The bulkhead admitted exactly the limit and fast-failed the rest.
- `civitai_app_heavy_bulkhead_rejects{key="heavy-image"}`: **0 → 0 → 20**. Monotonic, and the increment equals the 503 count exactly. A gauge reading the wrong module instance would have stayed at 0; one that merely existed would not have matched the shed count.

`civitai_app_heavy_bulkhead_active{key="heavy-image"}` is absent before any traffic and present after a single request — correct, since the key is created on first acquisition, and itself evidence that the request path and the scrape now share state.

### Incidental corroboration of the diagnosis

The pod logs `[bulkhead] HEAVY_REQUEST_CONCURRENCY=20` **twice**. That line sits at module scope in `request-bulkhead.ts`, so two lines means two module instances — the two-graph split is directly observable in the logs, and it is the thing this whole root cause turns on.

### Not claimed

This is a preview environment, not production. It shows the registration reaches a real scrape in a real production build; it is not a measurement of the production pools. Worth re-running the same probe there after this rides out, particularly since it also revives nine `node_postgres_*` gauges that have been dead long enough for a dependent alert to have been retired as unfixable.

ZacxDev:
### The two red preview checks are pre-existing, not from this change

**`preview / smoke-tests` — 4 failed / 61 passed.** All four are in `tests/preview-moderation.spec.ts`: `/moderator/reports` and `/moderator/images` redirecting to `/login`, plus a `TypeError: Cannot read properties of undefined (reading 'suggestedRebaseline')`. It is red on **5 of the 6 open PRs** I checked (4173, 4172, 4165, 4156, 4154), so it is a moderator-session/seed problem in the preview environment rather than anything in this diff.

**`preview / component-tests` — 3 failed / 1810 passed**, all in `src/components/Account/PlacementSpaceSection.freeSlots.browser.test.tsx` (`Cannot read properties of undefined (reading 'useQuery')`, then the slider never renders). **Byte-identical failure — same file, same three test names — on #4172 and #4165**, neither of which touches the placement/sticker surface. This diff touches four server-side files and no frontend at all.

I tried to attribute the component failure and could not; recording the two refuted hypotheses so nobody re-derives them:
- *"#4164 broke it"* — refuted: #4165 predates #4164 and still fails.
- *"previews built after #4164 merged pick it up"* — refuted: #4165's preview was built at 02:25Z, before #4164 merged at 03:05Z.

Both suites are marked report-only in the pipeline log (`report-only, not blocking the build`). Worth someone's attention on its own, but it is not this PR's.

One disclosure on method: I drove two 40-concurrent bursts against this preview to prove the reject gauge moves, which deliberately produced 20 × 503. That overlapped the smoke-test window, so I checked whether I had caused the failures — I had not. The signature is a login redirect, not a 503, and the same four fail on PRs where I ran nothing.

ZacxDev:
### Admission behaviour is unchanged — measured, not argued

Pinning the bulkhead maps on `globalThis` merges state that was previously per-module-instance, and `request-bulkhead` is evaluated **3 times** per pod (`[bulkhead] HEAVY_REQUEST_CONCURRENCY=20` appears three times in the startup log). So the fair question before merging is whether this silently tightens a live admission-control limit — `trpc.ts:354` claims the tRPC feed procedures and the REST handler "share one per-pod budget", and if that claim had been false, effective heavy concurrency would drop from up to 60 to 20 on merge.

It is not false. Three arms, run identically against a base-code preview and this PR's preview. A and B are positive controls establishing that each path can shed on its own; C is the discriminator.

| arm | base code | this PR |
|---|---|---|
| A — 40 concurrent REST `/api/v1/images` | `{"200":20,"503":20}` | `{"200":20,"503":20}` |
| B — 40 concurrent tRPC `image.getInfinite` | `{"200":20,"429":20}` | `{"200":20,"429":20}` |
| C — 20 REST + 20 tRPC in one burst | REST `10/10`, tRPC `10/10`, **20 shed** | REST `10/10`, tRPC `10/10`, **20 shed** |

Arm C is the one that decides it. Separate per-graph budgets would have admitted all 40 — each path sitting exactly at its own limit of 20 — and shed nothing. Both builds admitted exactly 20 and shed exactly 20. **The budget was already shared, and it still is.** Byte-identical behaviour; no capacity change from this PR.

The reason 3 module instances do not imply 3 budgets: only one graph *mutates* the maps — REST and tRPC are served from the same graph — while the instrumentation graph holds a read-only copy via `prom/client`. That read-only copy is precisely the empty one `collect()` was iterating, which is the bug this PR fixes.

**Bonus confirmation for the rejects gauge:** across the three arms on this PR's build, `civitai_app_heavy_bulkhead_rejects{key="heavy-image"}` went **20 -> 80**. Exactly +60, matching 20+20+20 shed. So it counts *both* entry paths — REST 503s and tRPC 429s — and not just the one I originally exercised.

Method disclosure: the base-code arms were run against an unrelated open PR's preview environment (short bursts of 40 concurrent requests, three arms, a few seconds each).

ZacxDev:
### Correction: my "no capacity change" claim is narrower than I stated

An adversarial audit found that my three-arm experiment does **not** cover every path that takes a `heavy-image` slot, so the conclusion I stated is over-broad. Correcting it here rather than editing the earlier comment, since it may already have been read.

**What I actually measured:** two acquisition paths — the REST `/api/v1/images` handler and the tRPC `image.getInfinite` procedure. Both are API routes.

**What I did not measure:** the SSR prefetch path. `src/pages/posts/[postId]/[[...postSlug]].tsx:48` and `src/pages/models/[id]/[[...slug]].tsx:363` call `ssg.image.getInfinite.prefetchInfinite`, which runs the full middleware chain and takes a slot on the *rendering* pod. `src/pages/api/v1/blocks/images.ts:187` is a fourth site.

This matters because of an inconsistency in my own evidence that I glossed over: the startup log shows **three** module copies, and my mechanism explanation only accounted for two ("REST and tRPC in one graph, instrumentation read-only"). The unexplained third copy is exactly where a separate SSR-render budget would live — and the SSR pool is the larger one.

So the honest statement is: **the budget is shared across the two API-route paths, measured; whether the SSR-render path shares it is untested.** If it does not, this PR merges a previously-separate budget into the shared 20 on the SSR pool. The consequence would be a soft degrade rather than errors — `prefetchInfinite` swallows a 429 and the client refetches — but it is a real behaviour change and I should not have implied it was ruled out.

A method caveat I should also have stated: the base-code arms ran in a *different* preview environment from the PR arms, so it is an A/B across two environments, not within one.

The fourth arm — N concurrent post-detail SSR page loads against N concurrent tRPC feed calls, on both builds — closes this. Worth running before anyone relies on the no-capacity-change claim for the SSR pool.

ZacxDev:
## The SSR-path question cannot be load-tested — but the fix makes it readable

Closing this out rather than leaving it open-ended.

**What I tried.** The fourth arm was to fire N concurrent SSR page loads against N concurrent tRPC feed calls and see whether the totals collide. The design was sound — I confirmed the prefetch really is per-request first (`createServerSideProps` -> `getServerSideProps`, then `await ssg.image.getInfinite.prefetchInfinite(...)`), so a page render genuinely takes a `heavy-image` slot on the rendering pod.

**Why it cannot run.** Every page route in a preview environment redirects to `/login` — `/`, `/images`, `/posts`, `/models`, and both detail pages. The auth gate fires before `getServerSideProps` reaches the prefetch, so no SSR render exercises the bulkhead there without a session.

The only other place to run it is production, and that is not a measurement — it is deliberately saturating an admission-control valve with concurrent SSR renders on pods serving real traffic. Not worth it to settle a question that has a free answer two steps away.

**So the earlier correction stands as the final word:** the shared budget is *measured* for the REST and tRPC paths; the SSR path is untested.

### The free answer, once this deploys

The whole point of the fix is that `civitai_app_heavy_bulkhead_active{key="heavy-image"}` starts emitting. One instant query then settles it, with no load and no risk:

```promql
count by (pod) (civitai_app_heavy_bulkhead_active{key="heavy-image"})
```

- **Series present on SSR-pool pods** -> the SSR render path does acquire from the shared budget. Then also read the `active` values there: they show the utilisation of a budget that, before this PR, that graph may have held privately.
- **No series on SSR-pool pods** -> the SSR path never entered the bulkhead on those pods, and the capacity question was moot all along.

Either way it is one read, and it is only possible *because* the metric works again — which is a fair summary of what the whole change buys.

PR #4174, opened 2026-08-20
feat(model-file-hash): implement the SHA256_12 backfill endpoint

ZacxDev:
Implements `GET /api/admin/temp/backfill-sha256-12`, the endpoint that migration `20260819000000` and `docs/image-resource-hash-matching.md` have twice been wrong about — first documenting it before it existed, then (in #4149) recording that it never had.

## Premise check

Every premise I was handed was re-verified before building on it. Three held; **one did not.**

| Premise | Verdict |
|---|---|
| The endpoint does not exist on any ref | ✅ **Held.** `git log --all --diff-filter=A -- src/pages/api/admin/temp/backfill-sha256-12.ts` returns nothing. |
| Both docs still say "🔴 not implemented" | ✅ **Held.** `docs/image-resource-hash-matching.md:249,265` and the migration's trailing comment. |
| `normalizeScanHashes()` excludes the all-zero sentinel | ✅ **Held.** `if (sha256 && !/^0+$/.test(sha256))` — and it is the reason this endpoint calls the helper instead of truncating inline. |
| The writer ledger fails when the set grows | ✅ **Held.** It tripped on the new writer exactly as designed; updated deliberately (below). |
| **~1.5M rows keep failing 12-char LoRA detection indefinitely; not self-healing** | ❌ **FALSE as of now.** |

### The corrected premise

**Production was already backfilled out-of-band, hours before this PR.** Measured on the nvme0 replica (`cnpg-cluster-nvme0-4`, read-only, SELECT-only):

| Metric | 2026-08-20 first read | re-read ~40 min later |
|---|---|---|
| `SHA256` rows | 1,489,384 | 1,487,830 |
| `SHA256_12` rows | 1,489,372 | 1,487,815 |
| `SHA256` rows lacking a `SHA256_12` sibling (sentinels excluded) | 15 | **15** |

Corroborating evidence that this was a deliberate pass, not scan-path drift:

- **Zero** `SHA256_12` rows disagree with `left(sha256, 12)`; all 1.49M are exactly 12 chars.
- Every one was created between `2026-08-19 22:15:34Z` and `2026-08-20 02:29:17Z` — a ~4-hour window, not the months the corpus spans.
- The migration contains no backfill DML, so it did not come from there.

The absolute counts drift (files get deleted; the table is live) — the **gap of 15 is the stable number**, and it is what a run today would actually write. So the corpus-wide job this endpoint was designed for is already done, and what remains is a near-free idempotent sweep. That is now recorded in both the doc and the migration comment, with the anti-join query, so nobody plans a 1.5M-row run that is already complete.

I did **not** run the endpoint against anything.

A fourth premise also turned out to be environmental rather than true: **repo-wide `tsc` is not red.** `pnpm typecheck` reports **0 errors in 78s**. The redness came from a stale generated Prisma client — `packages/civitai-db-schema/prisma/schema.prisma` and `node_modules/.prisma/client` are *generated, untracked* artifacts, and a copy taken from another clone predates the `SHA256_12` enum. `pnpm db:generate` + `npx prisma generate` fixes it. (`packages/civitai-db-schema` itself *is* tracked, contrary to the brief.)

## The endpoint

Follows `backfill-b2-file-locations.ts` — `WebhookEndpoint` token gate (33 of 35 endpoints in that directory use it), zod query schema, `booleanString()` dry-run default, per-batch cursor logging, `{ ok, dryRun, duration, complete, lastCursor, result }` response.

Derived entirely from stored `SHA256` rows: **no file access, no orchestrator, no re-scan.**

🔴 **It calls `normalizeScanHashes()` rather than truncating inline.** This is the design decision the whole PR turns on. The helper suppresses derivation from the all-zero "file unreachable" sentinel; an inline `slice(0, 12)` would give every unreachable file the *same* 12-char hash and make them all match each other. Calling the helper means the backfill cannot drift from the scan path's rules. Mutant M1 below is exactly this substitution, and it produces `SHA256_12=000000000000` — the hazard, reproduced on demand.

### Cost and resume

| Param | Default | Meaning |
|---|---|---|
| `dryRun` | `true` | Count what a live run would insert, without writing |
| `batchSize` | 1000 (max 5000) | `SHA256` rows per page |
| `maxBatches` | 100 (max 10000) | Hard cap on batches per invocation |
| `start` | 0 | Minimum `fileId` — the resume point |
| `end` | — | Maximum `fileId` |

A full pass is ~1.49M rows ≈ 1,490 batches at the default page size, so **one call deliberately does not finish the corpus**: it stops at 100 batches (~100k files), returns `complete: false` and `lastCursor`, and the operator resumes with `start=<lastCursor>` until `complete: true`. Re-running a completed range is safe and costs only the read.

Dry run does an extra PK lookup per batch so its `candidates` is the **real** write volume rather than the scanned row count — the difference between "1.5M rows to write" and the true "15", which is precisely the mistake this PR is documenting.

Idempotent via `createMany({ skipDuplicates: true })` → `ON CONFLICT ("fileId", type) DO NOTHING` against `ModelFileHash_pkey`, verified in prod as the table's only unique constraint.

## The ledger update, and why it is safe

Adding a writer trips `model-file-hash-writers.test.ts` **by design** — it asserts the exact writer set and fails when it grows. The new entry declares `normalizes: true`, and that flag is not taken on faith: the ledger greps the file (comments stripped) for a real `normalizeScanHashes(` **call**, so the entry cannot be a lie about a writer that doesn't call it. The `THREE writers` doc comment on `normalizeScanHashes` is updated to `FOUR` — a comment is a claim, and the ledger's own docstring says a stale one means the ledger is lying about the surface.

`normalizeScanHashes` itself is **unchanged**; the only edit to that file is its doc comment.

A second ledger, `rest-error-envelope-ledger.test.ts`, also caught this endpoint — my first draft returned `{ error: (error as Error).message }` on a 500. Fixed by routing through `handleEndpointError` (the shared chokepoint from civitai#3845) rather than adding a baseline exception; the resume cursor goes to the log instead. Both ledgers were found by grepping for affected suites, not by guessing names.

## Verification

**Suites enumerated by grep**, not by name — `grep -rl "normalizeScanHashes\|ModelFileHash\|modelFileHash\|model-file-scan\|reprocess-scan\|createModelFileScanRequest\|backfill-sha256-12"` over `src/ apps/ packages/ scripts/`, which is how the two ledgers surfaced.

Scored from `--reporter=json` test **counts**, never an exit code (prettier demonstrated why: it reported 2 failing files and still exited 0). Zero-test suites were counted separately, since an import failure reports "no tests" rather than a failure:

```
TOTAL 161  PASSED 161  FAILED 0  zero-test suites: 0
```

11 suites: the 10 grep-identified ones plus the `no-direct-shared-module-mock` ratchet (passes; the shared `dbMock` is used, no hand-rolled `~/server/db/client` mock). Repo-wide `pnpm typecheck`: **0 errors**. Prettier: clean.

### Red-at-base matrix — 11/11 mutants killed

Each mutant edits the **narrowest** expression that can be wrong, restores from a pristine copy, and is scored on test counts. Fixtures are pairwise distinct 64-char hashes whose 12-char prefixes differ from each other, from the fileIds, and from every constant the assertions name — so a mutant that hardcodes a literal prefix can only ever satisfy one of them.

| # | Mutation | Verdict | Killed by |
|---|---|---|---|
| M1 | `normalizeScanHashes(…).SHA256_12` → `row.hash.slice(0, 12)` (inline truncate, drops the sentinel guard) | **KILLED** 2/14 | both sentinel tests — mutant emits `101:SHA256_12=000000000000` |
| M2 | truncate to AutoV2's width 10 | **KILLED** 6/14 | both derivation tests + sentinel + 3 cursor tests |
| M3 | `skipDuplicates` removed | **KILLED** 2/14 | both idempotency tests (fake raises the real PK violation) |
| M4 | `written += result.count` → `+= toWrite.length` | **KILLED** 2/14 | "second run writes nothing" |
| M5 | `while (batches < maxBatches)` → `while (true)` | **KILLED** 2/14 | maxBatches + batchSize bounds |
| M6 | `gte: cursor` → `gte: 0` | **KILLED** 2/14 | cursor resumption + `start` |
| M7 | `take: params.batchSize` → `take: 1000` | **KILLED** 2/14 | batchSize page bound |
| M8 | dry run skips the coverage filter | **KILLED** 1/14 | "counts only rows genuinely missing a sibling" |
| M9 | writes `SHA256` instead of `SHA256_12` | **KILLED** 7/14 | derivation + type + sentinel + cursor |
| M10 | hand-rolled error envelope | **KILLED** 2/25 | the repo-wide envelope ledger **and** the local chokepoint test |
| M11 | `end` bound ignored | **KILLED** 1/14 | `end` upper bound |

M1's failure message is the point of the exercise:

```
expected [ '101:SHA256_12=000000000000', …(1) ] to deeply equal [ '205:SHA256_12=9f8e7d6c5b4a' ]
```

Each guard was watched go red for **its own** reason, on a reachable input, before being counted.

## Not done

- The endpoint has **not been executed** against any environment.
- The 15 remaining rows are left alone — running the sweep is an operator decision, and the doc now carries the anti-join to re-measure first.
- I could not determine *who* ran the out-of-band backfill or by what means; the evidence establishes that it happened and that its output is correct, not its provenance.

### DIFF
--- a/docs/image-resource-hash-matching.md
+++ b/docs/image-resource-hash-matching.md
@@ -20,13 +20,13 @@ four-stage pipeline.
 **Stage 1 — collect candidates** (lines 12–86). Five `UNION ALL` branches pull `(name, hash)` pairs
 out of `Image.meta`, because different tools write metadata differently:
 
-| Branch | Source | Shape |
-|---|---|---|
-| 1 | `meta.resources[]` | array with name / hash / weight |
-| 2 | `meta.hashes{}` | object — **the Forge / A1111 path, and the subject of this document** |
-| 3 | `meta."Model hash"` | single checkpoint string |
-| 4 | `meta.civitaiResources[]` | carries `modelVersionId` **directly — no hash** |
-| 5 | `Post.modelVersionId` | inherited, `detected: false` |
+| Branch | Source                    | Shape                                                                 |
+| ------ | ------------------------- | --------------------------------------------------------------------- |
+| 1      | `meta.resources[]`        | array with name / hash / weight                                       |
+| 2      | `meta.hashes{}`           | object — **the Forge / A1111 path, and the subject of this document** |
+| 3      | `meta."Model hash"`       | single checkpoint string                                              |
+| 4      | `meta.civitaiResources[]` | carries `modelVersionId` **directly — no hash**                       |
+| 5      | `Post.modelVersionId`     | inherited, `detected: false`                                          |
 
 Branch 4 is why on-site generations always work: we wrote that metadata ourselves and put the
 version id in it, so no matching is required.
@@ -56,7 +56,7 @@ once. A tool writes 10 characters and hits `AutoV2`; 12 characters and hits `Aut
 That is genuinely useful, and it is also the defect: **the system matches _values_, not
 _algorithms_.** It behaves as a lookup table of every hash string we have ever stored, and detection
 succeeds if and only if the generating tool wrote a string we hold verbatim. Exact equality cannot
-see that one string is a *prefix* of another.
+see that one string is a _prefix_ of another.
 
 ---
 
@@ -85,14 +85,14 @@ hash: hash?.replace('0x', '').slice(0, 12),   // correct as written — leave it
 ```
 
 SwarmUI's `sui_models[].hash` is a **tensorhash**, not a file SHA256 — per its own spec, linked at
-`swarmui.metadata.ts:21`: *"a hash of the data of the model processing only its tensor sections and
-not its header (ie to allow metadata to change without affecting the hash)"*. That is the same
+`swarmui.metadata.ts:21`: _"a hash of the data of the model processing only its tensor sections and
+not its header (ie to allow metadata to change without affecting the hash)"_. That is the same
 quantity as `AutoV3`, which we store truncated to 12. (Prod did that in a `truncate_autov3_hash()`
 trigger, `SUBSTRING(NEW.hash FROM 1 FOR 12)`; `normalizeScanHashes()` took it over in application
 code and migration `20260819010000` drops the trigger.)
 
 So `.slice(0, 12)` lands **exactly on `AutoV3`** — the one 12-character value that does match.
-Truncating to 10 would produce a 10-character *tensor*-hash prefix, and `AutoV2` is a *file*-hash
+Truncating to 10 would produce a 10-character _tensor_-hash prefix, and `AutoV2` is a _file_-hash
 prefix, so it would match nothing.
 
 This also explains the AutoV3 collision behaviour noted under Collision analysis: 1,505 AutoV3 values
@@ -175,7 +175,7 @@ v1.10.1 (stock A1111)       n=2409   exact=2341   sha12=5
 ```
 
 Stock A1111 is clean; Forge and Forge Neo carry essentially all of it. **It is not purely per-build,
-though**: 1,332 images in that single day contain *both* an exact-matching and a `sha12` LoRA hash,
+though**: 1,332 images in that single day contain _both_ an exact-matching and a `sha12` LoRA hash,
 so the same build emits both forms depending on the file. Treat "which build" as an open question,
 not a settled one.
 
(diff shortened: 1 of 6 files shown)


ZacxDev:
Closing this — it is a nice-to-have in a system this branch's authors do not own and that others are actively working in. Not merging unsolicited code into someone else's live work area.

**The premise it was written on turned out to be false, and that is worth recording even though the PR is closing.**

It was opened because `docs/image-resource-hash-matching.md` and migration `20260819000000` both told an operator to run `GET /api/admin/temp/backfill-sha256-12`, and that file has never existed on any ref — so the reasoning was "~1.5M pre-existing rows never get `SHA256_12` and keep failing 12-char LoRA detection indefinitely."

**That is no longer true.** Production was already backfilled out-of-band while this was being written: all **1,489,372** `SHA256_12` rows were created between `2026-08-19 22:15Z` and `2026-08-20 02:29Z`, all exactly 12 characters, with **zero** disagreeing with `left(sha256, 12)`. The migration contains no backfill DML, so it did not come from there.

**What actually remains is 15 rows** — the anti-join of `SHA256` rows with no `SHA256_12` sibling, stable across two reads 40 minutes apart while the absolute counts drifted (1,489,384 → 1,487,830; the table is live and files get deleted):

```sql
SELECT count(*)
FROM "ModelFileHash" h
WHERE h.type = 'SHA256'
  AND h.hash !~ '^0+$'
  AND NOT EXISTS (
    SELECT 1 FROM "ModelFileHash" s
    WHERE s."fileId" = h."fileId" AND s.type = 'SHA256_12'
  );
```

So nobody needs to plan a corpus-wide job. If those 15 are worth closing, it is a one-statement sweep, not an endpoint.

**One hazard worth carrying forward if anyone does write a backfill later.** Derive through `normalizeScanHashes()` rather than truncating inline. A hand-rolled `slice(0, 12)` produces `SHA256_12 = '000000000000'` for the all-zero "file unreachable" sentinel, which makes **every unreachable file match every other one**. That was mutant M1 in this PR's battery and it is an easy mistake to make in an ad-hoc script.

**Also unresolved, and the reason this comment is longer than a close usually warrants:** who ran the out-of-band backfill, and how. The evidence establishes that it happened and that its output is correct — not its provenance. If it was an existing operator script, that is the right answer and this PR was redundant from the start. If it was ad-hoc, the sentinel hazard above is the thing to know about before it is run again.

The related doc corrections already landed separately in #4149 (`a0f4a86fba`) — the docs and migration no longer point at the endpoint that does not exist, so nothing is left claiming a tool that is not there.

PR #4179, opened 2026-08-20
test(preview): repoint the moderator smoke specs off the pre-#3573 routing (fixes #4171)

### LINKED ISSUE #4171: preview / smoke-tests is red on every PR based after #3573 — two specs still assert the pre-migration /moderator/reports routing
`preview / smoke-tests` is failing on **5 of 10 open PRs** — every one based after #3573 merged. It is a stale-test break, not a product break, but it is a **shared gate**, and a permanently-red shared gate trains everyone to click through it. The next real smoke failure will be clicked through too.

## Root cause

**#3573 (`moderator-app-pages`) migrated the moderator surfaces out of the main app.** It deleted `src/pages/moderator/reports.tsx` (among many others) and added a catch-all `src/pages/moderator/[...slug].tsx` that 302-redirects migrated paths to the standalone moderator app. `reports: 'reports'` is in `MIGRATED_ROUTES`.

Two preview specs still assert the old routing:

| spec | assertion | why it now fails |
|---|---|---|
| `tests/preview-auth-guard.spec.ts:43` | `mod stays on /moderator/reports` → `toHaveURL(/\/moderator\/reports/)` | the mod is redirected off-origin to the moderator app |
| `tests/preview-moderation.spec.ts:61` | `/moderator/reports renders the report queue` → `toHaveTitle(/Reports/i)` + `role="table"` | that page no longer exists here; the request redirects away |

`MODERATOR_PATH` in the auth-guard spec is a single constant shared by three tests — the two *bounce* tests still pass, because a non-mod is turned away by the `_app` guard before the redirect matters.

## The discriminating evidence

Base-commit ancestry cleanly separates pass from fail:

```
#4149  base 9f558f37e3  predates #3573   ->  smoke PASSES (65 passed)
#4152  base fbac791f30  has #3573        ->  smoke FAILS  (4 failed)
#4156  base 481582d969  has #3573        ->  smoke FAILS  (4 failed)
```

Timeline: the last passing run was **01:00:59Z**; every run from **01:23Z** onward fails with an identical "4 failed". #4156 changes *only* `.github/workflows/lint.yml`, which cannot break `/moderator/reports` — so the cause is the base, not the PRs.

## Two theories I checked and refuted, so nobody re-derives them

- **"The catch-all 500s because `MODERATOR_APP_URL` is unset in previews."** It *is* unset in the preview pod, and the catch-all calls `env.MODERATOR_APP_URL.replace(...)` unguarded — which looks like a `TypeError`. It is not: `server-schema.ts:373` gives it `.default('https://moderator.civitai.com')`, so it resolves. No 500.
- **"An anonymous probe shows the redirect."** It does not. `curl` on `/moderator/reports` returns 302 → `/login?returnUrl=...` — that is the auth gate, and it never reaches the catch-all. Any conclusion drawn from an unauthenticated probe here is about the gate, not the migration.

⚠️ Worth a separate look: because the default points at **production**, a moderator hitting a migrated `/moderator/*` path *on a preview* is redirected to the **production** moderator app, leaving the preview origin. That may be intended for now ("Temporary during the transition", per the catch-all's own comment), but it means previews cannot exercise moderator surfaces at all.

## Proposed fix — needs someone with #3573's context to pick

**`preview-auth-guard.spec.ts`** — no route judgment needed. The spec's real intent is *"the `_app` guard admits a mod and bounces non-mods"*, and "stays on /moderator/reports" was only how that was expressed. Assert the 3xx **without following it**, which is already the idiom in this very file (the `/login` hub test does exactly this, deliberately, because the target is external). That keeps the guard coverage *and* pins the migration.

**`preview-moderation.spec.ts`** — this one genuinely asserts *"a moderation surface renders for a mod"*, and that surface has **moved to another app**. Either repoint it at a moderator page that still lives here, or accept that the coverage moved and assert the redirect instead.

🔴 **I did not just repoint it, because picking the replacement is not obvious and I got it wrong twice:**

- `strikes.tsx` — has `if (!features.strikes) return <NotFound />`, so the test would silently depend on a feature flag being on in preview.
- `tags.tsx` — has a feature gate *and* no `<Meta title>`, so the title anchor the spec relies on does not exist.
- `scanner-policies/index.tsx` — clean on both counts (no feature gate, `requireModerator: true`, `<Meta title="Scanner Policies">`), but its `<Table>` is deep in the file and I could not confirm it renders unconditionally, so the `role="table"` anchor may not hold.

Someone who knows which moderator surfaces are staying can pick in a minute. I am happy to implement it once the route is chosen — say the word and I will.

## Why this is worth doing now rather than waiting for the migration to finish

Five open PRs are currently carrying a red gate that has nothing to do with their changes. Every one of their authors has to either investigate it themselves or learn to ignore it, and the second is what actually happens.

ZacxDev:
Fixes #4171.

## Why

`preview / smoke-tests` is red on every PR based after **#3573** (`moderator-app-pages`). It is a stale-test break, not a product break — but it is a **shared gate** currently red on unrelated PRs, and a permanently-red shared gate trains everyone to click through it. The root cause, the base-commit ancestry that separates pass from fail, and two refuted theories are all in #4171; this PR implements the fix.

## The failures — four, not two

#4171's table lists two assertions. The smoke bot's per-test breakdown on #4156 lists **four**, and the fourth has a distinct mechanism worth calling out:

| failing test | mechanism |
|---|---|
| `mod stays on /moderator/reports` (`preview-auth-guard`) | mod is now 302'd off-origin to the moderator app |
| `/moderator/reports renders the report queue` (`preview-moderation`) | page deleted; `reports` ∈ `MIGRATED_ROUTES` |
| `/moderator/images renders the image review queue` (`preview-moderation`) | same — `images` ∈ `MIGRATED_ROUTES` |
| `mod can self-seed a post, report it, and action the report` | **not a routing problem** — `report.getAll` and `report.setStatus` were deleted from `src/server/routers/report.router.ts` in #3573 (`95157404b0`, `e782b7ceeb`, `b3f01055af`). Only `report.create` survives here. |

A fix that only touched the two routing assertions would have left the gate red.

## What changed

**`tests/preview-auth-guard.spec.ts`** — as #4171 proposed, no route judgment. `MODERATOR_PATH` stays `/moderator/reports`; the mod control now asserts the **3xx without following it** (`maxRedirects: 0`), the idiom this file already uses for the `/login` hub test and for the same reason (the hop leaves this origin). The guard's only two bounces are `'/'` and `'/login?returnUrl=…'` (`server/auth/route-guard.ts`), so asserting the `Location` is neither of those — and *is* the moderator-app `/reports` hop — keeps the guard coverage the test exists for **and** pins the migration. The two non-mod bounce tests are untouched: the guard writes its 302 from `_app.getInitialProps`, ahead of the catchall's `getServerSideProps`, which is why they never broke.

**`tests/preview-moderation.spec.ts`** — repointed at two moderator surfaces that still live in this app, plus an explicit migration pin, plus the surviving half of the action test.

**`tests/preview-auth.setup.ts`** — the mod warm-up loop was warming `/moderator/reports` and `/moderator/images`. Those now 302 to `MODERATOR_APP_URL`, which **defaults to production** (`server-schema.ts:373`), so the warm-up was hitting prod from CI and compiling nothing on the preview. Repointed at the two new paths.

## Routes chosen, and why

Screened every file under `src/pages/moderator/` on `origin/main` against four criteria: present on `origin/main`; absent from `MIGRATED_ROUTES`; `requireModerator: true`; **no** `features.*` gate; and an anchor verified to render **outside** every loading / empty-state / error branch.

| route | file | anchors |
|---|---|---|
| `/moderator/rewards` | `src/pages/moderator/rewards/index.tsx` | `<Meta title="Rewards" deIndex />` (L47) + `<Title order={1}>Purchasable Rewards</Title>` (L50) |
| `/moderator/suspicious-audit-matches` | `src/pages/moderator/suspicious-audit-matches.tsx` | `<Meta title="Suspicious Audit Matches" deIndex />` (L239) + `<Title order={1}>Suspicious Audit Matches</Title>` (L245) |

Both anchors sit above the `isLoading ? … : empty ? … : rows` ternary, so they hold on a full **or** empty prod-clone DB, and even if the page's tRPC query errors — which is the property the old `role="table"` anchor did *not* have. `<Title order={1}>` renders `component: 'h1'` (`@mantine/core` `Title.cjs:73`), so `getByRole('heading')` is a structural assertion, not a text search. Neither page does data work in SSR (`createServerSideProps` with no `useSSG` and no resolver / a session-only resolver), so they are cheap on a cold single-replica preview pod.

Both are also **absent from `apps/moderator/src/routes/`**, i.e. they are not mid-migration. (I deliberately avoided `/moderator/models`, which is clean on the other criteria but *does* have a counterpart route in the spoke.)

### Candidates rejected

- `strikes.tsx` — `if (!features.strikes) return <NotFound />`; the test would silently depend on a flag being on in preview. *(rejected in #4171; re-confirmed)*
- `tags.tsx` — feature gate **and** no `<Meta title>`. *(rejected in #4171; re-confirmed — `grep` for `<Meta title` returns nothing)*
- `scanner-policies/index.tsx` — clean on gate + title, but #4171 could not confirm its `<Table>` renders unconditionally. I did not need to resolve that: I anchored on an unconditional `<h1>` instead of a table, which makes the whole question moot, and the two pages above have simpler render trees.

## Coverage that moved rather than being deleted

- The reports/images **queues** are now covered by a single test asserting both migrated paths still 302 to their mapped moderator-app path (`reports` → `/reports`, `images` → `/images`), inspected **without following** — following would probe production from CI. Expected targets are written as literals, not imported from `MIGRATED_ROUTES`, so a wrong edit to that map fails the test instead of redefining what "correct" means.
- The action test keeps `post.create` → `report.create` (both still main-app `guardedProcedure`) and drops the `report.getAll` / `report.setStatus` legs, with a comment saying **why** and that they belong to `apps/moderator`'s own suite now. That coverage genuinely left this repo; I did not want it to look like an oversight to the next person.

## What I verified — and what I could not

**Verified statically, on `origin/main`:**
- Both chosen files exist (`git cat-file -e origin/main:<path>`).
- Neither is matched by `MIGRATED_ROUTES` — ran the real `migratedRouteKey` longest-prefix logic over the parsed map: `rewards` → no key, `suspicious-audit-matches` → no key. Positive control in the same run: `reports` → key `reports` → `reports`, `images` → key `images` → `images` (17/17 keys parsed).
- Both have `requireModerator: true`; neither contains `features.` or `useFeatureFlags`.
- Both `<Meta title>` and both `<Title order={1}>` are outside every conditional branch (read the files, not grepped).
- `Meta` emits `<title>{title}</title>` verbatim (`components/Meta/Meta.tsx`); titles are still matched as case-insensitive regexes, the same tolerance the pre-#3573 spec used.
- `pnpm typecheck` — **0 type errors** (`tests/` is in `tsconfig.json`'s `include`).
- `pnpm exec prettier --check` — clean on all three files. (`preview-auth-guard.spec.ts` was already unformatted on `main`; it is formatted now, which is the small unrelated hunk in that file.)

**Could not be verified locally.** These are **preview** specs — they only run against a deployed preview under `playwright.preview.config.ts`, so nothing here can be executed on a workstation. Two assertions in particular were new behaviour reasoned to rather than watched:
- that `page.request` carries the `test.use({ storageState })` cookies and `baseURL` (documented for `browserContext.request`, but not something I ran here), and
- that `report.create` returns a row with a numeric `id` — the code says it always does (`createReport` returns either the pre-existing `validReport` or the created row, and self-reporting is only blocked for `ReportEntity.User`), but the previous spec had a `report.getAll` fallback, so I cannot tell from history which branch actually fired.

**Now verified on a real preview run** (`preview` label → `pr-preview-4179-s64z2`, commit `da475ed`):

> `preview / smoke-tests` — ✅ **66 passed · 0 flaky · 0 failed**

against the broken baseline of **61 passed · 4 failed** on #4156. That exercises both open questions above: `page.request` did carry the mod's session (the auth-guard control and the migration test both went green, and a missing cookie would have produced the `/login` bounce those tests explicitly assert against), and `report.create` did return a numeric `id`. Every other check on this PR is green too, including `tekton / typecheck`.

Test count moved 65 → 66: the two deleted render tests were replaced by two render tests **plus** one migration test, and the action test stayed (with its actioning legs removed).

## Blast radius

Test-only — three files under `tests/`, no `src/` change. `tests/` is outside the lint workflow's `LINTABLE` filter (`^(src|packages|apps/event-engine)/`), and the modified-file ESLint/Prettier steps are `continue-on-error` anyway.

### DIFF
--- a/tests/preview-auth-guard.spec.ts
+++ b/tests/preview-auth-guard.spec.ts
@@ -6,15 +6,20 @@ import { storageStatePath } from './preview-fixtures';
  * existing smoke / moderation specs don't cover:
  *
  *   1. The _app moderator guard (server/auth/route-guard.ts, moved off the edge middleware) REDIRECTS a
- *      gate-passing NON-moderator away from /moderator. preview-smoke proves a mod can enter and
- *      preview-moderation proves the queues render — neither proves a non-mod is bounced, which is the whole
- *      point of the new guard.
+ *      gate-passing NON-moderator away from /moderator. preview-moderation proves the surviving moderator
+ *      surfaces render for a mod — neither proves a non-mod is bounced, which is the whole point of the new
+ *      guard.
  *   2. /login is now a pure server-side redirect to the centralized hub (buildHubLoginRedirect), threading the
  *      returnUrl. We assert the redirect target WITHOUT following it (the hop lands on the external hub).
  *
  * Only runs under playwright.preview.config.ts (needs PREVIEW_URL + the minted storage states).
  */
 
+// #3573 migrated this path to the standalone moderator app (shared/constants/migrated-moderator-routes.ts:
+// `reports` -> `reports`), so for a MODERATOR it now ends in a 3xx to that app. It is still the right probe
+// for the guard: the guard runs in _app.getInitialProps, ahead of the catchall's getServerSideProps, so a
+// non-mod is bounced before the migration redirect can matter (which is why the two bounce tests below are
+// unaffected by the migration), and for a mod the Location tells us WHICH hop happened.
 const MODERATOR_PATH = '/moderator/reports';
 
 // tester (Flipt allowlist) + gold (allowlist + tier) both CLEAR the preview gate but lack isModerator, so the
@@ -28,11 +33,17 @@ for (const role of NON_MOD_ROLES) {
     test(`${role} cannot reach ${MODERATOR_PATH}`, async ({ page }) => {
       await page.goto(MODERATOR_PATH, { waitUntil: 'domcontentloaded' });
       // The guard sends an authed non-mod to '/' (login can't grant the missing permission, so it would loop).
-      expect(page.url(), `${role} should be bounced off the moderator surface`).not.toContain('/moderator');
+      expect(page.url(), `${role} should be bounced off the moderator surface`).not.toContain(
+        '/moderator'
+      );
       // A bounce to /login would mean the session didn't resolve (a different failure) — assert it did.
-      expect(page.url(), `${role} session should resolve, not bounce to login`).not.toContain('/login');
+      expect(page.url(), `${role} session should resolve, not bounce to login`).not.toContain(
+        '/login'
+      );
       // And it isn't the preview-access gate firing instead (these roles are allowlisted).
-      expect(page.url(), `${role} should not hit preview-restricted`).not.toContain('/preview-restricted');
+      expect(page.url(), `${role} should not hit preview-restricted`).not.toContain(
+        '/preview-restricted'
+      );
     });
   });
 }
@@ -40,10 +51,27 @@ for (const role of NON_MOD_ROLES) {
 test.describe('_app moderator guard admits a moderator (control)', () => {
   test.use({ storageState: storageStatePath('mod') });
 
-  test('mod stays on /moderator/reports', async ({ page }) => {
-    const resp = await page.goto(MODERATOR_PATH, { waitUntil: 'domcontentloaded' });
-    expect(resp?.status(), 'HTTP status for /moderator/reports').toBeLessThan(400);
-    await expect(page).toHaveURL(/\/moderator\/reports/);
+  test('mod is not bounced — the only hop is the moderator-app migration redirect', async ({
+    page,
+  }) => {
+    // Inspect the 3xx Location WITHOUT following it — same idiom (and same reason) as the /login test below:
+    // the hop leaves this origin for the standalone moderator app, which a preview cannot exercise
+    // (MODERATOR_APP_URL defaults to production — server-schema.ts). `page.request` shares the browser
+    // context's cookies (minted by preview-auth.setup.ts) and its baseURL, so this is the mod's session.
+    const res = await page.request.get(MODERATOR_PATH, { maxRedirects: 0 });
+    expect(res.status(), `HTTP status for ${MODERATOR_PATH}`).toBeGreaterThanOrEqual(300);
+    expect(res.status(), `HTTP status for ${MODERATOR_PATH}`).toBeLessThan(400);
+
+    const location = res.headers()['location'] ?? '';
+    expect(location, `${MODERATOR_PATH} should set a redirect Location`).not.toBe('');
+    // The GUARD has exactly two bounces (route-guard.ts): '/' for an authed non-moderator, and
+    // '/login?returnUrl=…' when the session didn't resolve. Neither may fire for a mod — that IS this
+    // control. (The login bounce percent-encodes its returnUrl, so it can't spell a literal '/reports'.)
+    expect(location, 'a mod must not be bounced home by the _app guard').not.toBe('/');
+    expect(location, 'a mod session should resolve, not bounce to login').not.toContain('/login');
+    // …so the hop is the migration one. Deploy-agnostic: MODERATOR_APP_URL is env-per-deploy, so pin the
+    // PATH the mapping produces, not the host.
+    expect(location, 'mod should be forwarded to the moderator app /reports').toMatch(/\/reports$/);
   });
 });
 
--- a/tests/preview-auth.setup.ts
+++ b/tests/preview-auth.setup.ts
@@ -143,8 +143,13 @@ setup('mint preview sessions', async ({ request }) => {
   if (mod) {
     // /moderator/* render only for a moderator (gold would be bounced), so warm the
     // moderation-spec pages with the mod cookie.
+    // These MUST be paths this app still serves. /moderator/reports + /moderator/images
+    // migrated to the standalone moderator app in #3573 and now 302 off-origin to
+    // MODERATOR_APP_URL — which defaults to PRODUCTION — so warming them warmed prod
+    // and compiled nothing here (civitai#4171). Keep this list in step with
+    // MODERATOR_SURFACES in preview-moderation.spec.ts.
     const headers = { cookie: `${COOKIE_NAME}=${mod}` };
-    for (const path of ['/moderator/reports', '/moderator/images']) {
+    for (const path of ['/moderator/rewards', '/moderator/suspicious-audit-matches']) {
       await request.get(path, { timeout: 60_000, headers }).catch(() => {});
     }
   }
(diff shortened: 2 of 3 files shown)


ZacxDev:
Verified the headline independently — **`preview / smoke-tests` is green on this PR: 66 passed, 0 flaky**, against the broken baseline of 61 passed / 4 failed. Every other check green. That is the real gate, actually green on a real preview run, which is the best verification this change could have.

Two things worth recording before it merges.

## 🟡 The dropped coverage did not move — it is gone

The PR drops the report-actioning legs and says that coverage *"now belongs to `apps/moderator`'s suite."* **There is no such suite.**

```
apps/moderator  *.test.ts / *.spec.ts files : 0
positive control — packages/                : 81
```

The grep works; the suite does not exist. `apps/moderator` has a `package.json` and no test runner configured.

This does **not** invalidate the fix. `report.getAll` and `report.setStatus` really are gone from the main app (verified: 0 refs in `src/server/routers/report.router.ts`), so those legs genuinely cannot be exercised from here any more, and leaving them would keep the gate red for a reason no main-app change can fix. Dropping them is right.

But the justification should say what is true: **this coverage is lost, not relocated**, and the loss originates in #3573 rather than in this PR. Stated as a handoff to an existing suite, a reviewer reasonably concludes nothing was lost — and nobody files the gap.

Suggested: reword to name it as a gap, and file a follow-up for report-actioning coverage in `apps/moderator` (which would also be that app's first test).

## ✅ The scope correction in this PR is right, and #4171 was wrong

The issue said "two specs." The smoke bot's per-test breakdown shows **four** failures, and the fourth is not a routing problem at all — #3573 also deleted the two router procedures. **A fix touching only the two routing assertions would have left the gate red**, which is exactly what I would have shipped from my own issue. Corrected on #4171.

## Also worth calling out

`tests/preview-auth.setup.ts` was warming `/moderator/reports` and `/moderator/images` with the mod cookie. Those now 302 to `MODERATOR_APP_URL`, which **defaults to production** — so CI was warming the production moderator app and compiling nothing on the preview. Unprompted find, and a good one.

## Route screening

The two replacement routes were screened for the thing that defeated three earlier candidates — an anchor that renders **unconditionally**. Anchoring on `<Meta title>` plus an unconditional `<Title order={1}>` (via `getByRole('heading')`, structural rather than a text search), both above the `isLoading ? … : empty ? … : rows` ternary, side-steps the `<Table>` question that made `scanner-policies` unusable. That is the right shape.

PR #4181, opened 2026-08-20
fix(dev-server): the daemon never read DEV_DAEMON_PORT, so setting it broke the CLI instead of moving the daemon

ZacxDev:
`DEV_DAEMON_PORT` was read where the client decides what to **connect** to and ignored where the daemon decides what to **listen** on. Setting it pointed the CLI at a port nothing was serving:

```
DEV_DAEMON_PORT=9555 node .claude/skills/dev-server/cli.mjs status
  -> CLI polls :9555, spawns a daemon that binds :9444, gives up: "Failed to start daemon"
```

The daemon has always accepted `--port` — the spawn simply passed no arguments, and the daemon read no environment. So this is not "the daemon is unconfigurable"; it is a missing pass-through.

## Why the diff is bigger than one line

Five files each decided the port for themselves, and three of them were wrong:

| file | before |
|---|---|
| `cli.mjs` | reads `DEV_DAEMON_PORT` |
| `scripts/test-unit-run.mjs` | reads `DEV_DAEMON_PORT` |
| `console.mjs` | **hardcoded `9444`** |
| `scripts/daemon.mjs` | **argv only, no environment** |
| `.claude/hooks/check-writable.mjs` | **`9444` baked into a regex** |

That is not a bug any one of them contains — it is a bug in the set, and patching one site leaves the others regenerating it. The number now lives in `scripts/daemon-port.mjs` and everything resolves it there. A daemon a client spawns inherits that client's environment, so both ends read the same variable through the same function and cannot disagree; an explicit `node daemon.mjs --port <port>` still wins for a daemon started by hand, and goes through the same validator.

The hook mattered more than it looks: with `9444` baked in, setting `DEV_DAEMON_PORT` silently switched the "never curl a dev port unbounded" nudge off *for the daemon* — the one long-lived server the rule most exists for. Verified with a negative control: at `DEV_DAEMON_PORT=9555`, `curl :9555` nudges and `curl :7777` stays silent.

`parsePort` refuses a value that is not a port instead of handing back `parseInt`'s `NaN`, which used to reach a URL as `http://127.0.0.1:NaN` and a `listen()` as `ERR_SOCKET_BAD_PORT` — the latter invisible via `cli.mjs`, which detaches the daemon with stdio ignored and reports only "Failed to start daemon".

## Verification

Reproduced the reported path rather than reasoning about the code.

**Red at base** (`2b226eb054`, pristine worktree, `DEV_DAEMON_PORT=19461`):

```
  Daemon port: 9444
Uncaught exception: Error: listen EADDRINUSE ... 127.0.0.1:9444
```
No ready line, and nothing ever listened on 19461.

**Green at HEAD:** `{"type":"daemon_ready","port":19461,"pid":2584501,...}`, socket confirmed via `ss -lptn`. The test asserts the pid the daemon reports equals the child that was spawned — a port that answers proves a listener, not *this* listener, and an orphan from an earlier run answers just as cheerfully.

**Mutation battery**, run twice — once on the original guards, once on the audit-round guards. Each mutant applied alone, tree restored and `cmp`-verified between each, every one killed by its own named assertion:

| mutant | killed by |
|---|---|
| daemon ignores the environment | both behavioural cases |
| daemon ignores `--port` | the `--port` precedence case |
| `--port` back to `parseInt` | the argv-validation case |
| port resolved at module load | the import-safety case |
| resolver stops validating | the rejects-a-non-port case |
| a **new** file hardcodes 9444 | the ledger (the growth case) |
| the hook re-hardcodes 9444 | the ledger |
| `console.mjs` does `resolvePort() + 1` | the client-address case |
| `shutdown()` stops waiting | the pid-file case |
| the targeted fallback is removed | the degrade-to-direct case |
| the default drifts to 9445 | the default case |

🔴 **Two of these batteries reported a wrong result first, and both corrections are the interesting part.** The first battery reported every mutant SURVIVED — its own output parser matched nothing, so it was measuring itself, not the code. The second had a genuine survivor: the `test-unit-run` fallback assertion matched the generic phrase "running directly", which *both* guards on that path print, so deleting the targeted guard left the suite green. It now names the specific message.

`console.mjs` is a TUI with no end-to-end case here, which is what the ledger is for. It walks the tree rather than a fixed list, so it fails when the set of files deciding the port **grows** — including a file that does not exist yet — or **shrinks**, and it carries a positive control asserting it found the files it claims to scan, so a typo in its roots cannot read as "no copies anywhere".

`scripts/__tests__` — **324 passed**, including the 15 existing `test-unit-run` cases that drive the real script against a stub daemon over `DEV_DAEMON_PORT`, i.e. the queue path this touches. The hook's own selftest is green, and the `daemon port` case in it is a real positive control: `9444` matches none of the other port ranges, so it passes only if the module-source read worked.

## Typecheck

`pnpm typecheck` — **0 errors in 61s**.

Correcting an earlier version of this description, which claimed 12 pre-existing errors on `main`: that was wrong, and it was my environment. `git worktree add` does not initialise submodules, so `event-engine-common` was an empty directory in my worktree and 9 `TS2307 Cannot find module '…/event-engine-common/…'` errors followed, with 3 `implicitly any` errors downstream of them. `git submodule update --init event-engine-common` — which CI does and a fresh worktree does not — takes it to 0. `main` was never red.

## Not verified

Windows was not exercised locally. Nothing here is platform-specific — no `--port` argument quoting was added, and the teardown in the new tests uses the daemon's own `POST /shutdown` rather than a signal — but that is a design claim, not a measurement. Both `Windows dev-env` legs are green on this PR.

Closes ClickUp 868kuaa4e.

### DIFF
--- a/.claude/hooks/check-writable.mjs
+++ b/.claude/hooks/check-writable.mjs
@@ -9,6 +9,7 @@
  */
 
 import { stdin } from 'process';
+import { readFileSync } from 'fs';
 
 // `prettier --write <targets>`: the incident was a repo-WIDE rewrite, not formatting a directory
 // this change owns. Read the targets instead of matching `--write`, so a scoped path just runs.
@@ -84,7 +85,39 @@ function isUnscoped(target) {
 // URL carrying a dev port in a query parameter — and `$(curl ...)` defeats it anyway. A guard with
 // those false positives and no override is one people route around, and the routes around it are
 // shorter than the compliant path. Asking keeps the nudge and costs a keystroke when it is wrong.
-const DEV_PORTS = /(?:localhost|127\.0\.0\.1|0\.0\.0\.0|\[?::1\]?):(?:30\d\d|51[67]\d|9444)\b/i;
+//
+// The daemon's port comes from the dev-server skill rather than being baked in here: it is
+// overridable via DEV_DAEMON_PORT, and a hardcoded copy meant that setting it silently switched
+// this nudge off for the daemon — the one long-lived server the rule most exists for.
+//
+// Resolved lazily and defensively, and both matter. A hook that throws breaks EVERY Bash call, so
+// a missing skill directory or a malformed DEV_DAEMON_PORT must degrade, not propagate; and doing
+// it on first use rather than at import keeps this file free of top-level await.
+let devPortsRe = null;
+function devPorts() {
+  if (devPortsRe) return devPortsRe;
+  let daemon = '';
+  try {
+    // Synchronous require-equivalent is unavailable in ESM, so the port is read from the module
+    // source. A regex over one `export const` line is deliberately duller than an import: it
+    // cannot execute skill code inside a hook that runs before every command.
+    const source = readFileSync(
+      new URL('../skills/dev-server/scripts/daemon-port.mjs', import.meta.url),
+      'utf8'
+    );
+    const declared = /export const DEFAULT_DAEMON_PORT = (\d+)/.exec(source);
+    const override = process.env.DEV_DAEMON_PORT?.trim();
+    const port = /^\d+$/.test(override ?? '') ? override : declared?.[1];
+    if (port) daemon = `|${port}`;
+  } catch {
+    /* skill absent — the 30xx/516x-517x ranges still guard the dev servers themselves */
+  }
+  devPortsRe = new RegExp(
+    String.raw`(?:localhost|127\.0\.0\.1|0\.0\.0\.0|\[?::1\]?):(?:30\d\d|51[67]\d${daemon})\b`,
+    'i'
+  );
+  return devPortsRe;
+}
 
 // Any of curl's, wget's or PowerShell's own timeout flags, including `-m5` and bundled shorts.
 // `-T` is wget's timeout but curl's --upload-file, and `-m` is curl's --max-time but wget's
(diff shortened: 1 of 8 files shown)


ZacxDev:
**Correcting two claims this PR's description made, since they were up here long enough to be read and believed.** An adversarial audit found both, plus a blocking defect. Fixed in `85cb922567`; description updated.

**1. "the ledger fails when the set grows (a new hardcoded copy)" — that was false.** The ledger scanned a fixed four-element array of readers, so a *new* file hardcoding the port was never read and could not fail it. There was a standing counterexample the whole time: `.claude/hooks/check-writable.mjs` already baked `9444` into its dev-port regex, so `9444 is spelled once` was false at the moment I wrote it, and the test was green. The ledger now walks the tree.

That fifth copy was not cosmetic. It meant `DEV_DAEMON_PORT=9555` silently switched the "never curl a dev port unbounded" nudge off *for the daemon* — the one long-lived server the rule most exists for — which is the exact class of defect this PR set out to remove, one file short.

**2. The mutation table oversold what it covered.** It listed only deletion-shaped mutants. `const DAEMON_PORT = resolveDaemonPort() + 1;` in `console.mjs` passed every ledger assertion, because `includes('daemon-port.mjs')` is satisfied by a *comment* (and the file's own prose names the module in several) and `includes('resolveDaemonPort(')` is satisfied by any call regardless of what happens to the result. `console.mjs` has no behavioural test, which was the stated reason the ledger existed. Both clients now take a whole URL from `resolveDaemonUrl()` and do no arithmetic on a port.

**And the blocking one, which is worth stating plainly: the test deleted developer state on every unit-suite run.** `POST /shutdown` replies 200 and *then* unlinks the pid file on a 100 ms timer, so the helper I wrote specifically to preserve `daemon.pid` restored it ~100 ms before the daemon deleted it. Since `scripts/**/*.test.ts` is in the `unit` project, any `pnpm test:unit:run` removed it, and it is only ever written at daemon start — so it did not come back. `shutdown()` now waits for the port to stop answering.

One more worth recording because it is the same shape as the guards this PR is about: in the new battery, the `test-unit-run` fallback mutant **survived**. The assertion matched the generic phrase "running directly", which *both* guards on that path print, so deleting the targeted guard left the suite green — the mutant died to the other guard. The assertion now names the specific message. Seven mutants, each killed by its own named assertion, tree cmp-verified between each.

A delta re-audit of `7dd65a9d62..85cb922567` is running.

Issue #4167, opened 2026-08-20
/comments/v2/<id> 404s for app-listing comments, so 'copy comment link' is dead


ZacxDev:
## Summary

`/comments/v2/<id>` returns **404** for a comment on an app-store listing, so "copy comment link" on those threads produces a dead URL.

## Cause

`src/pages/comments/v2/[id].tsx` resolves a comment to its parent entity and bails when it cannot:

```ts
if (!threadData.threadType || !threadData.threadParentId) {
  return { notFound: true };   // :129-130
}
```

An app-listing thread has neither. Listing threads are keyed by `appListingId` and the pages are addressed by **slug**, not by the numeric `threadParentId` every other entity uses — the same asymmetry [#4160](https://github.com/civitai/civitai/pull/4160) had to solve for notification URLs.

## Scope

Pre-existing; **not** introduced by #4160 and not changed by it. Small and separable.

## Suggested fix

Give the page an `appListing` branch that resolves the listing slug and redirects to the listing detail page with the comment highlighted — reusing the `getListingDetailHref` helper #4160 introduced so a route rename moves every caller at once.

## Test coverage expected

- A comment id on an app-listing thread resolves rather than 404ing, and lands on the correct listing URL with the highlight parameter.
- A comment id on each already-handled entity type still resolves **unchanged** — a regression here breaks comment permalinks site-wide.
- An unresolvable / nonexistent comment id still 404s, so the fix does not turn the page into one that answers for everything.
- Assert the whole resolved URL, not a substring.

## Origin

Flagged during #4160 and deliberately scoped out of it.

PR #4184, opened 2026-08-20
feat(notifications): notify an app-listing owner of a top-level comment (#4166)

### LINKED ISSUE #4166: App listings: owner gets no notification for a first top-level comment on their listing
## Summary

App-listing owners get no notification when someone leaves a **first top-level comment** on a listing they have not previously participated in.

## What #4160 did and did not close

[#4160](https://github.com/civitai/civitai/pull/4160) made app-listing comment threads notify for **mentions**, **replies** and **thread responses** by giving `threadUrlMap` an `appListing` branch and removing the `appListingId IS NULL` exclusions.

That means an owner who has already commented in the thread now *does* get reply/thread-response notifications. The precise residue is narrower than "owners get nothing":

> A first top-level comment on a listing whose owner has never joined the thread produces no notification for that owner.

## Cause

The owner-facing `new-comment` notification processors are **per-entity SQL** and there is no `appListing` branch, so the query returns no rows for an app-listing thread. This is a missing branch, not an exclusion guard — nothing needs removing.

## What a fix needs

- A new notification type (e.g. `new-app-listing-comment`) rather than a widening of an existing per-entity query.
- Its own entry in the dedupe-priority ordering, so it interacts correctly with the reply/mention types that #4160 enabled — a comment that is both the first on a listing and a mention must not double-notify.
- A `UserNotificationSettings` row so it can be turned off. A notification users cannot disable is its own problem.
- The URL should reuse the same `getListingDetailHref` path #4160 introduced, so a future route rename moves both.

## Test coverage expected

- Owner receives the notification for a first top-level comment on their listing.
- Owner does **not** receive a duplicate when the same comment also mentions them (dedupe priority).
- A commenter does not notify themselves.
- The existing reply/mention behaviour from #4160 is unchanged.
- Assert the whole resolved URL string, not a substring of it.

## Origin

Flagged during #4160 and deliberately scoped out of it.

ZacxDev:
Closes #4166.

## The gap

#4160 made app-listing comment threads notify for **mentions**, **replies** and **thread responses**. The residue it deliberately left:

> A first top-level comment on a listing whose owner has never joined the thread produces no notification for that owner.

The owner-facing processors are per-entity SQL and none had an `appListing` branch — a **missing branch**, not an exclusion guard, so nothing needed removing.

## The change

`new-app-listing-comment`, a new notification type modelled on `new-post-comment`:

- Joins `app_listings` for the owner and the listing name.
- **Top-level comments only** (`t."appListingId" IS NOT NULL`). A reply's own thread is keyed by its parent *comment* and carries no `appListingId`, so replies stay with the reply processors and cannot double up.
- Skips self-comments and system-owned listings, both against the **resolved** owner.
- **Approved listings only**, and never a shadow revision — see below.
- **Resolves the owner kind-awarely**, not from the denormalized column — see below.
- Carries the same two lookback floors `new-post-comment` carries, for the same reason: the type has no cursor row until its first successful run, so it inherits the job's global last-run.
- Links to `/apps/mine`, **not** the public listing page — see below.

### 🔴 Three fixes from audit, all measured against production

**1. Approved-only.** Comments on `draft`, `rejected` and moderator-`removed` listings all emitted. `appListings.getAppDetail` is approved-only (`app-listings.router.ts:1416`), so those notifications pointed at a `NotFound`. **Only 14 of 28 prod listings are `approved`** — half the population, not an edge case. Now `al."status" = 'approved'`, mirroring the sibling raw-SQL consumer (`appListing.metrics.sql.ts:36,87`). `al."revision_of_id" IS NULL` added alongside: a shadow revision freezes `user_id` at clone time and can name a stale owner. Not reachable today (shadows are draft) — free to add.

**2. The destination, reversing the issue's suggestion.** #4166 asks for the `getListingDetailHref` deep link. **That link 404s for the people being notified.** `/apps/store-preview/<slug>` gates on `hasAppsStoreAccess`, which rolls out to `moderators` OR `app-dev-testers` only — and of the **4 current listing owners in prod, one is in neither cohort**.

#4160's three types escaped this because their recipient had already commented in the thread, and had therefore already proven they could open the page. **This is the first app-listing notification pushed to an owner unconditionally**, so it cannot borrow that assumption.

`/apps/mine` gates on `appBlocksAuthor` — the developer cohort a listing owner is in by construction — and is where all four existing owner-facing app-listing notifications already point (`app-listing.notifications.ts:35`). That constant is now **exported and shared** rather than duplicated, so a route rename moves all five. The trade-off is losing the comment highlight: a link the owner can open beats a more precise one they cannot.

**3. Kind-aware ownership.** The owner was read from `al."user_id"` alone. `app-access.service.ts:222-272` carries a 🔴 header — *"Ownership is not one column"*: for `kind='onsite'` the canonical owner is `AppBlock.app.userId`, with the column only a fallback that `acceptTransfer` itself calls *"the denormalized follow-up, where a 0-count is a legitimate desync"*. Measured: **0 of 21 onsite listings diverge today**, so this is latent — but 17 of 21 onsite listings are the shape that could, and a divergence would send the listing name, slug and a commenter's username to the **previous** owner while silently dropping it for the real one. That is a disclosure, not just a miss, so it is resolved rather than commented around:

```sql
CASE WHEN al.kind = 'onsite' THEN COALESCE(oc."userId", al."user_id") ELSE al."user_id" END
```

with `LEFT JOIN`s (never `INNER` — an offsite listing has no block and must still notify).

### Dedupe

It emits the shared `comment:v2:<id>` key and runs in the **EntityOwner** batch, behind Mention / DirectResponse / ThreadResponse. So a comment that is both the first on a listing **and** a mention of the owner yields exactly one notification — the mention, the more specific of the two.

> [!NOTE]
> **One deliberate deviation from the issue text, flagged for your call.** The issue asks for "its own entry in the dedupe-priority ordering". I used the existing `CommentNotificationPriority.EntityOwner` rather than adding a fifth distinct rank, because that is what every other owner processor uses (`new-post-comment`, `new-image-comment`, `new-article-comment`) and it already satisfies the functional requirement — losing the key race to all three types #4160 enabled. A distinct rank would add a fifth sequential batch in `send-notifications` for one processor and buy nothing, since there is no other `appListing` claimant in the EntityOwner batch to tie with. Happy to change it if you want the literal reading.

### Schema

**No schema change and no migration.** `UserNotificationSettings` is `{id, userId, type, disabledAt}` with `@@unique([userId, type])`; `type` is free text and rows are written as **user opt-outs**, not seeded definitions. The settings list is derived from the processor registry by `getNotificationTypes()` (`utils.notifications.ts:101-113`), so registering the processor is what puts the toggle on screen. Nothing needs manual application in any environment.

## Tests — 32, red→green matrix

Base column measured by removing the new processor block from `comment.notifications.ts` and running the same suite.

| | at base | at HEAD |
|---|---|---|
| **tests of the new type** | **RED** | green |
| invariant guards (#4160 behaviour, priority ordering) | green | green |
| **total after the audit fix round** | — | **32 green** |

**Labelled honestly.** The two `INVARIANT GUARD:` describe blocks pin behaviour this PR does **not** change — #4160's mention/reply/thread-response rendering, and the pre-existing `CommentNotificationPriority` ordering. They are green at base and stay green, so they cannot demonstrate that anything here works. They were previously labelled `REGRESSION:`, which overclaimed; relabelled this round.

The dedupe test the issue asks for — *"owner does not receive a duplicate when the same comment also mentions them"* — is genuinely red→green. It could not have been on the pre-#4160 tree, where mentions on app-listing threads were excluded outright and there was no competing notification to dedupe against; it would have passed for the wrong reason.

It is asserted as a **relationship**, not two independently-spelled literals: the key this processor emits for a V2 row must be *identical* to the one `new-mention` emits, compared on the evaluated key rather than the SQL text (the two use different expressions — `commentDedupeKey('v2')` vs `commentDedupeKeyByVersion`). A positive control confirms the evaluator can return differing values.

### 🔴 The SQL is now pinned by a WHOLE-STATEMENT snapshot

Audit found that fragment assertions let **three semantic mutants walk a fully green 295-test suite**, each proven reachable by executing the mutant SQL:

| mutant | effect | why fragments missed it |
|---|---|---|
| append `AND 1 = 0` | **7 rows → 0. The feature is dead, permanently and silently.** | every fragment still present |
| `AND (c."userId" != <owner> OR TRUE)` | owner notified for their own comment | the guard text is still there, just inert |
| `notBlockedBetween` args swapped | the *commenter's* Hide suppresses instead of the owner's | valid SQL, same tokens |

Fixed by pinning the **whole normalised statement** with `toMatchInlineSnapshot`, the instrument the sibling file already uses for #4160's three processors (`comment.appListing-notifications.test.ts:417,488,615`) under a comment reading *"A fragment assertion is satisfied by semantically inverted SQL … a full snapshot is not."* The `notBlockedBetween` argument **order** is additionally pinned by generating the expected clause from the helper, with a positive control proving the helper is order-sensitive.

### The settings-toggle assertion was vacuous

`expect(…defaultDisabled).toBeFalsy()` asserted a **removed concept** — it can only ever read `undefined` (`notification-settings-polarity.test.ts:204` actively asserts no source contains the field), so it passed vacuously. The real lever is **`toggleable`**, and a `toggleable: false` mutant survived the entire suite while the type vanished from both the settings UI and the bulk-toggle list.

Now asserted against the **output** of `getNotificationTypes` — `notificationCategoryTypes['Comment']` must contain the type, it must be in `notificationTypes`, and it must not be opt-in — with a positive control that a bogus type is absent. A note was added at `comment.post-owner.test.ts:77`, where the pattern was copied from, so it does not spread further.

## Mutation testing — 14 mutants, all killed (re-run after the fix round)

A fix round resets the gate, so the whole battery was re-run. Controls: unmutated → SURVIVED (32 passed), broken-displayName → KILLED. A full-count assertion turns a non-compiling mutant into a loud `HARNESS-ERROR` rather than a false SURVIVED.

| mutant | killed by |
|---|---|
| **S1 append `AND 1 = 0`** (feature dead) | whole-statement snapshot |
| **S2 self-notify guard inert via `OR TRUE`** | whole-statement snapshot |
| **S3 `notBlockedBetween` args swapped** | snapshot **+** the argument-order test |
| **S4 `toggleable: false`** (vanishes from settings) | the new `notificationCategoryTypes` test |
| S5 drop approved-only filter | approved-only test + snapshot |
| S6 `= 'approved'` → `<> 'removed'` | approved-only test + snapshot |
| S7 drop shadow-revision guard | shadow-revision test + snapshot |
| S8 owner collapses to the denormalized column | kind-aware test + snapshot |
| S9 `COALESCE` operands swapped (column wins) | kind-aware test + snapshot |
| S10 deep link restored | 3 URL tests, incl. the explicit no-deep-link guard |
| S11 launch floor → 2099 (disabled forever) | floor test (now extracted from SQL) + snapshot |
| S12 drop top-level-only restriction | top-level test + snapshot |
| S13 priority → Mention | priority test |
| S14 `app_blocks` LEFT → INNER (offsite dropped) | kind-aware test + snapshot |

S1–S4 are the four the audit found walking; all now die. The date-floor guard was also fixed — it compared a **duplicated literal** rather than the value in the SQL, so it could not fail if the floor moved; it now extracts the floor from the generated statement, which is what lets S11 die.

## A seam catch worth flagging

The existing `comment.dedupe-key.test.ts` auto-enumerates every processor that claims the dedupe key and asserts each renders a URL. The new type fell to that suite's `default:` details shape (a `model3d` payload), carried no slug, and rendered no URL — caught only when the suites were run **together**, not in isolation. Fixed by adding the type's real details shape to that switch, which is the intended extension point. One line, included here.

## Verification

- `pnpm typecheck` → **0 type errors in 83s**
- `vitest run src/server/notifications` → **295 passed / 18 files** on this branch; **328 passed / 19 files** on the combined tree with #4167
- Prettier clean on all changed files

## 🔴 Rollback is NOT a `git revert`

`getNotificationMessage` resolves at **render** time (`NotificationList.tsx:64-69`), not at write time. So reverting this PR after even one notification has been delivered leaves those rows in the database with **no processor to render them**: they display as nothing while still counting toward the unread badge. The repo already carries this scar (`collection.notifications.ts:50-56`).

**Correct rollback: delete `prepareQuery`, keep `prepareMessage`.** That stops new notifications immediately while every delivered row still renders. Only remove `prepareMessage` once the delivered rows have aged out.

## 🔴 No feature flag

Unlike the rest of the W13 App Blocks family, this processor carries **no flag** — it begins emitting on merge. That is much less pointed now that the link goes to `/apps/mine` (which gates on `appBlocksAuthor`, a cohort every listing owner is in) rather than the public detail page, but it is still true: the first app-listing comment after merge notifies its owner, with no staged rollout. Worth a deliberate yes rather than a default.

## 🔴 Not verified — stated plainly, please merge knowing this

**The SQL has never been executed against a database.** Not in a test, not by hand. Every assertion in the 26 tests is on the **generated query text** and the **rendered message** — `toContain` against a string. No row has ever come back from this query.

What *is* confirmed: the identifiers, against `schema.full.prisma` — `@@map("app_listings")`, `serialId @map("serial_id")`, `userId @map("user_id")`, `slug`/`name` unmapped, plus an existing `app_listings_user_idx` on the owner column. And the same `LEFT JOIN "app_listings" al ON al."serial_id" = …` already ships in #4160, so the join shape has run in production.

What is **not** confirmed is that the query **returns the rows intended**. A predicate that is syntactically valid and semantically wrong — a join that silently matches nothing, a `WHERE` that excludes every real row — produces **zero notifications and a fully green suite**. String assertions cannot distinguish "correct query" from "query that returns nothing".

The cheapest closing check: run the generated SQL read-only against a listing that has a top-level comment from a non-owner, and confirm it returns exactly one row for the owner.

Interpretation note: "first top-level comment" in the issue describes the *symptom scenario*. This notifies on every top-level comment, as every other entity-owner processor does; dedupe handles the overlap.

### DIFF
--- a/src/server/notifications/__tests__/comment.appListing-owner.test.ts
+++ b/src/server/notifications/__tests__/comment.appListing-owner.test.ts
@@ -107,8 +107,18 @@ describe('new-app-listing-comment — the owner finally gets notified', () => {
   });
 
   it('shares the destination constant with the other owner-facing listing notifications', () => {
-    // A RELATIONSHIP, not a second literal: a route rename must move all five together. Compared
-    // against a sibling processor's rendered url rather than against the string '/apps/mine'.
+    // A RELATIONSHIP: a route rename must move all five together, so this compares against a
+    // sibling processor's RENDERED url rather than against the string '/apps/mine'.
+    //
+    // 🔴 ON ITS OWN THIS ASSERTION IS TAUTOLOGICAL, and saying so is the point. Both sides read
+    // the same `OWNER_SUBMISSIONS_URL`, so changing that constant to '/apps/other' drifts them
+    // together and this test still passes (verified). What it proves is only "these two agree" —
+    // never "they agree on the RIGHT value".
+    //
+    // The three literal `toBe('/apps/mine')` assertions above are what pin the value; this pins
+    // that the sharing is real and not a coincidence of two matching literals. The PAIR is the
+    // guard. Do not "simplify" either half away: drop the literals and a wrong route passes; drop
+    // this and a duplicated literal drifts silently.
     const sibling = (appListingNotifications as unknown as Record<string, Def>)[
       'app-listing-approved'
     ].prepareMessage({
@@ -116,6 +126,8 @@ describe('new-app-listing-comment — the owner finally gets notified', () => {
       details: { slug: SLUG, name: LISTING_NAME },
     });
     expect(message()?.url).toBe(sibling?.url);
+    // Anchors the shared value, so this test cannot pass while both sides drift.
+    expect(sibling?.url).toBe('/apps/mine');
   });
 
   it('the url is independent of the slug entirely (no slug-shaped failure modes left)', () => {
@@ -172,12 +184,29 @@ describe('new-app-listing-comment — SQL: who it notifies, and who it must not'
     expect(normalizeSql(sql())).toContain(`${APP_LISTING_OWNER_SQL} > 0`);
   });
 
-  it('fires on TOP-LEVEL comments only, so it cannot double up with the reply processors', () => {
-    // A reply's thread is keyed by its parent COMMENT and carries no appListingId. Dropping the
-    // `IS NOT NULL` here would match reply threads too, and every reply would notify the owner a
-    // second time alongside new-comment-reply.
+  it('keeps the top-level predicate, which is REDUNDANT defence-in-depth rather than the guard', () => {
+    // 🔴 CLAIM CORRECTED. This test used to say that dropping the `IS NOT NULL` "would match reply
+    // threads too, and every reply would notify the owner a second time". That is FALSE, and
+    // asserting it made the test name overclaim what the predicate does.
+    //
+    // What actually excludes replies is the INNER `JOIN "app_listings" al ON al."serial_id" =
+    // t."appListingId"` — NULL never equals anything, so a reply thread is dropped by the join
+    // whether or not this predicate exists. Deleting it yields BYTE-IDENTICAL rows.
+    //
+    // So the deletion mutant SURVIVES as UNREACHABLE, not as unasserted — no input can make it
+    // change the answer, and the two diagnoses have opposite remedies (an unasserted mutant wants
+    // an assertion; an unreachable one wants none, because any test for it is vacuous by
+    // construction). This assertion therefore pins the predicate's PRESENCE as intent that must
+    // survive a refactor making the join LEFT or reordering it — not its behaviour.
+    //
+    // Provenance: the byte-identical result was measured by the #4184 audit against a real
+    // Postgres 16 fixture built from the repo's DDL. Recorded rather than re-derived.
     expect(normalizeSql(sql())).toContain(
       'JOIN "Thread" t ON t.id = c."threadId" AND t."appListingId" IS NOT NULL'
+    );
+    // The clause that does the real work, pinned beside it so the pair cannot drift apart.
+    expect(normalizeSql(sql())).toContain(
+      'JOIN "app_listings" al ON al."serial_id" = t."appListingId"'
     );
   });
 
--- a/src/server/notifications/__tests__/comment.appListing-owner.test.ts
+++ b/src/server/notifications/__tests__/comment.appListing-owner.test.ts
@@ -1,11 +1,19 @@
 import { describe, expect, it } from 'vitest';
 import {
+  APP_LISTING_OWNER_SQL,
   CommentNotificationPriority,
   commentNotifications,
 } from '~/server/notifications/comment.notifications';
 import { mentionNotifications } from '~/server/notifications/mention.notifications';
+import { appListingNotifications } from '~/server/notifications/app-listing.notifications';
+import { notBlockedBetween } from '~/server/notifications/base.notifications';
 import { NotificationCategory } from '~/server/common/enums';
-import { notificationProcessors } from '~/server/notifications/utils.notifications';
+import {
+  isOptInNotification,
+  notificationCategoryTypes,
+  notificationProcessors,
+  notificationTypes,
+} from '~/server/notifications/utils.notifications';
 
 /**
  * `new-app-listing-comment` — the app-listing owner's notification for a top-level comment.
@@ -16,13 +24,19 @@ import { notificationProcessors } from '~/server/notifications/utils.notificatio
  * `appListing` branch. This is that branch, as its own notification type.
  *
  * What is asserted here, and why each shape:
- *   - the WHOLE resolved URL, never a substring — `toContain('/apps')` is satisfied by
- *     `/apps/store-preview/undefined`.
+ *   - the WHOLE resolved URL, never a substring. The destination is the owner's `/apps/mine`
+ *     submissions view, NOT the public listing page: that page gates on `hasAppsStoreAccess`
+ *     and, measured against prod, 404s for one of the four current listing owners.
+ *   - the WHOLE normalised SQL statement via `toMatchInlineSnapshot`, because a fragment
+ *     assertion is satisfied by semantically inverted SQL — an appended `AND 1 = 0`, an
+ *     `OR TRUE` bolted onto the self-notify guard, or swapped `notBlockedBetween` arguments all
+ *     leave every fragment intact. Fragments stay as well, for readable failures.
  *   - the SQL with COMMENTS STRIPPED FIRST, so a guard left in place as a comment (text present,
  *     clause dead) still fails.
  *   - the dedupe interaction with the three types #4160 enabled, since "must not double-notify"
  *     is the requirement that is easy to state and easy to get wrong.
- *   - the reply/mention behaviour #4160 shipped, unchanged.
+ *   - the settings toggle via the OUTPUT of `getNotificationTypes` (`toggleable` is the real
+ *     lever), never via `defaultDisabled` — a removed concept that can only read `undefined`.
  *
  * Fixture values are non-default and pairwise distinct so a mutant binding the wrong one produces
  * a DIFFERENT string rather than the same one by coincidence.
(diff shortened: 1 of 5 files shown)


ZacxDev:
### Audit fix round pushed — `dbade64dd2`

All four items addressed; details in the PR body. Summary:

| item | fix |
|---|---|
| 🔴 no listing-status filter | `al."status" = 'approved'` + `al."revision_of_id" IS NULL` |
| 🔴 destination 404s for owners | links to `/apps/mine` via the now-**exported** shared constant |
| 🟡 SQL pinned by fragments | whole-statement `toMatchInlineSnapshot` + `notBlockedBetween` **argument-order** pin with a positive control |
| 🟡 vacuous `defaultDisabled` | asserts `notificationCategoryTypes['Comment']` / `notificationTypes` / `isOptInNotification` instead |
| 🟡 owner read bypasses resolver | resolved kind-awarely in SQL, mirroring `resolveCanonicalListingOwner` |

Also: the date-floor guard now **extracts** the floor from the generated SQL instead of restating the literal (it previously could not fail if the floor moved), and the two `REGRESSION:` blocks are relabelled `INVARIANT GUARD:` — this PR touches none of those processors.

**Battery re-run after the fix — 14 mutants, all killed**, controls validated (unmutated → SURVIVED at 32 passed; broken-displayName → KILLED; full-count assertion so a non-compiling mutant cannot read as SURVIVED). The four that previously walked now die:

- `AND 1 = 0` → whole-statement snapshot
- self-notify guard inert via `OR TRUE` → whole-statement snapshot
- `notBlockedBetween` args swapped → snapshot **and** the argument-order test
- `toggleable: false` → the new `notificationCategoryTypes` test

Local: `pnpm typecheck` 0 errors; `vitest run src/server/notifications` **301 passed / 18 files**.

---

### ⚠️ The `Unit tests` red is a PRE-EXISTING break on `main`, not this PR

```
FAIL src/server/services/__tests__/model-file-hash-writers.test.ts
     > ModelFileHash writer ledger > matches the writers actually present in the source tree
AssertionError: expected [ …(6) ] to deeply equal [ …(5) ]
```

Reproduced on **clean `origin/main`** with none of this PR's files present — it fails identically. The extra entry is:

```
+ "packages/civitai-db-schema/prisma/migrations/20260819000000_model_file_hash_sha256_12/migration.sql
   :: INSERTINTO\"ModelFileHash\""
```

That migration landed after #4149 wrote its 5-entry ledger, so the ledger enumerates 5 while the tree holds 6. The guard is working as designed — it fails when the writer set grows — but the growth is a migration on `main`, unrelated to notifications.

It now blocks **both** PRs in this pair and any other PR rebased past that migration. Fix is one line on `main`: add the migration to `LEDGER_ENTRIES`, or drop `prisma/migrations/` from the scan roots if migrations should not count as writers. Say the word and I will open it as its own PR.

PR #4185, opened 2026-08-20
fix(comments): resolve /comments/v2/<id> for app-listing threads (#4167)

### LINKED ISSUE #4167: /comments/v2/<id> 404s for app-listing comments, so 'copy comment link' is dead

ZacxDev:
Closes #4167.

## The bug

`/comments/v2/<id>` returned **404** for every comment on an app-store listing, so "copy comment link" on those threads produced a dead URL.

Every other thread entity is **ID-addressed**: the `Thread` row carries the parent's integer id and that id appears verbatim in the URL. App listings are **SLUG-addressed** (`/apps/store-preview/<slug>`) while `Thread.appListingId` holds `app_listings.serial_id` — an integer surrogate that appears nowhere in the URL. So an app-listing thread legitimately has no usable `threadParentId`, and the page's guard rejected all of them:

```ts
if (!threadData.threadType || !threadData.threadParentId) {
  return { notFound: true };   // :129-130
}
```

There were in fact **two** independent 404 paths, and both had to close:
1. `getThreadDetails` had no `appListing` arm, so it returned `{null, null}` and hit the guard above.
2. Even past that guard, `threadUrlMap`'s `appListing` arm (added by #4160) reads `details.appListingSlug` — and the page neither selected nor passed a slug, so it returned `undefined` and fell through to the trailing `notFound`.

## The change

- Resolution moves out of the page into `src/server/utils/comment-permalink.ts`, so it can be tested without a DB or a Next request. The page keeps only the parts that genuinely need both: load the comment, redirect or 404.
- The page now selects `appListing: { select: { slug: true } }`. That `select` object is spread into both `thread` and `thread.rootThread`, so replies resolve through the root exactly as other entities do.
- URL construction goes through the shared `threadUrlMap`, so the permalink, the notification and the store card all share `getListingDetailHref` and a route rename moves every caller at once.

Two deliberate choices worth review attention:

- **The `appListing` arm is APPENDED to the entity chain, not inserted.** A `Thread`'s entity FKs are mutually exclusive, so ordering is unobservable in practice — appending anyway means no pre-existing type's precedence can move, which is the cheap way to keep a sitewide-permalink regression off the table. Pinned by a test.
- **Addressability is checked per addressing scheme.** A missing id renders `/posts/null`; a missing slug renders `/apps/store-preview/undefined`. Both are 404s dressed as working links, so neither passes.

## Tests — 33, red→green matrix

Run against a verbatim replica of the pre-change logic (the `getThreadDetails` if-chain, the two-stage rootThread fallback, the guard, and the original `threadUrlMap` argument set) to establish the base column.

| | at base | at HEAD |
|---|---|---|
| **5 appListing-behaviour tests** | **RED** | green |
| 27 regression / negative-control tests | green | green |
| after adding the layer assertion (below) | — | **33 green** |

The 5 that go red at base:

```
expected null to be '/apps/store-preview/pixel-forge?highlight=8123&…'   (POSITIVE, whole URL)
expected null to be '/apps/store-preview/pixel-forge?highlight=8123&…'   (reply via root thread)
expected null to be '/apps/store-preview/a%20b%2Fc?highlight=8123&…'     (slug encoding)
expected null to deeply equal { threadType: 'appListing', …(2) }         (resolver target)
expected undefined to be null                                            (no threadParentId)
```

The 27 that pass at base are **regression coverage** — they pin behaviour that must not change, so passing at base is exactly what makes them regression guards. They are not evidence the fix works.

One honest caveat: *"an app listing whose slug did not resolve fails closed"* is green at base too, but **for a different reason** (at base app listings were not handled at all). It is a genuine negative control at HEAD and vacuous at base.

Coverage includes all **9** already-handled entity types at full-string precision, each through both the direct and the root-thread path, plus a check that all nine resolve to **distinct** URLs so no row can pass by aliasing another.

## Mutation testing — 8 mutants, all killed

Harness controls run first: unmutated → SURVIVED (33 passed), always-null → KILLED. The harness also asserts the **full test count** per run, so a mutant that fails to compile reports `HARNESS-ERROR` instead of a false SURVIVED.

| mutant | verdict | killed by |
|---|---|---|
| M1 invert branch order (appListing first) | KILLED | `expected '/apps/store-preview/pixel-forge?…' to be '/posts/3310?…'` |
| M2 swap operands (appListing checks `threadParentId`) | KILLED | `expected null to be '/apps/store-preview/pixel-forge?…'` |
| M3 dead guard (slug check commented out, text present) | KILLED* | `expected { threadType: 'appListing', …(2) } to be null` |
| M4 stale rebind (`appListingSlug ← thread.id`) | KILLED | `expected '/apps/store-preview/4471?…' to be '/apps/store-preview/pixel-forge?…'` |
| M5 drop root-thread fallback | KILLED | `expected null to be '/posts/3310?…'` (10 tests) |
| M6 `comicChapter.projectId → .id` | KILLED | `expected null to be '/comics/3310?…'` |
| M7 drop id-addressability check | KILLED | `expected '/posts/undefined?…' to be null` |
| M8 stale rebind (`commentParentId ← thread.id`) | KILLED | 22 tests |

### \*M3 initially SURVIVED — reachable but unasserted, *not* unreachable

This distinction is the reason M3 is worth reading, because the two diagnoses have **opposite remedies** and are easy to confuse.

A surviving mutant has two very different causes:

- **Unreachable** — the mutated line never executes for any fixture (an earlier check always wins, the branch lands exactly on its own boundary, a value converges before the assertion runs). The remedy is to build a fixture that *reaches* it. Scoring such a mutant SURVIVED tells you nothing about the guard.
- **Reachable but unasserted** — the line *does* execute and its behaviour *does* change, but nothing observes the difference. The remedy is an **assertion**, not a fixture.

M3 is the second kind, confirmed rather than assumed. I probed both layers directly with the negative-control fixture (`{ appListing: { slug: undefined } }`):

```
resolveThreadTarget   = null      ← my guard runs here and returns null
buildCommentPermalink = null      ← unchanged even with my guard deleted
```

The fixture reaches the mutated line. Deleting the slug check makes `resolveThreadTarget` return a target instead of `null` — a real behaviour change — but `buildCommentPermalink` *still* returns `null`, because `threadUrlMap` carries its **own** slug guard (shipped in #4160). Two redundant guards, and every end-to-end test was satisfied by the downstream one, so no test could tell which had fired.

Redundancy here is fine as defence-in-depth. What was not fine is that **either layer could be deleted silently**. The remedy was therefore an assertion at the resolver layer — `resolveThreadTarget({ appListing: { slug } })` must be `null` for `undefined`/`null`/`''` — which pins *this* layer while the end-to-end tests pin the other. M3 now dies on that assertion specifically:

```
AssertionError: expected { threadType: 'appListing', …(2) } to be null
```

Had I misdiagnosed this as unreachable, the "fix" would have been a new fixture — which would have changed nothing, because the fixture was never the problem.

### A false kill, also fixed

N5/N6 in the sibling PR (#4184) initially died on a `TypeError` from a regex in the *test* (`.match(…)![0]` on a null match), not on an assertion. **A mutant that dies for the wrong reason is a false kill** — it would still have "died" had the test been checking something else entirely. The general fix is to assert the extraction succeeded *before* using it, so the failure comes from the test's own claim. Applied there; noted here because it is the same class of error as M3.

Fixture ids are non-default and pairwise distinct, and distinct from every literal the assertions name, so a mutant binding the wrong one produces a different string rather than the same one by coincidence.

## Verification

- `pnpm typecheck` → **0 type errors in 83s**
- `vitest run src/server/notifications src/server/utils/__tests__/comment-permalink.test.ts` → **328 passed / 19 files** (on the combined tree)
- Prettier clean on all changed files

## 🔴 The select is now TYPE-checked, not mock-checked

The resolver took `thread: any`, so nothing connected the Prisma `select` to the code reading it. A wrong select produced a **404 with a fully green suite**.

### Why a test could not have fixed this

The instinct is "add a test with a mocked `dbRead`". **That is the wrong remedy, and it is worth being explicit about why**, because the wrong remedy produces a green suite that proves nothing:

> Hand-writing the mock's return shape means the fake encodes the **same belief** as the code. A wrong select and a wrong mock are wrong *together*, and the suite stays green.

A select-correctness gap is closed by a **type**, not a test.

### What actually closes it

The select moves into `comment-permalink.ts` beside its reader, and the reader's parameter type is derived from the select:

```ts
export const commentPermalinkSelect = { … } satisfies Prisma.CommentV2Select;
type CommentPermalinkPayload = Prisma.CommentV2GetPayload<{ select: typeof commentPermalinkSelect }>;
export type ThreadEntitySource = Pick<CommentPermalinkPayload['thread'], 'post' | … | 'appListing'>;
```

🔴 **`satisfies Prisma.ThreadSelect` alone is INSUFFICIENT** — every field of a `*Select` type is optional, so **omission satisfies it**. It catches a misspelled relation (`appListings`) or field (`slugg`); it does *not* catch a dropped entry. The `Pick` is the half that catches omission, because `Pick<T, K>` requires every `K` to be a key of `T`.

### Measured, with controls

| mutant | `pnpm typecheck` | test suite |
|---|---|---|
| drop `appListing` from the select | **FAIL** `TS2344` | 1 failed |
| select `.id` instead of `.slug` | **FAIL** `TS2339 Property 'slug' does not exist` | 🔴 **948 PASSED** |
| narrow `rootThread`'s select to `{ id: true }` | **FAIL** `TS2345` | 🔴 **948 PASSED** |
| drop `model3d` from the select | **FAIL** `TS2344` | 1 failed |
| *(control)* unmutated | PASS | 948 passed |

**Two of the four are invisible to the entire suite** and caught only by the compiler. The third is the expensive one: `rootThread` carries the entity relations for every **reply** permalink, so narrowing it breaks replies for *every entity type sitewide*.

## 🔴 `model3d` — a live 404, fixed here

`/comments/v2/<id>` 404'd for **every 3D-model comment**. `threadUrlMap` has had a working `/3d-models/<id>` arm all along and the notification SQL resolves it — the page's select and resolver simply omitted it.

**Root cause, and the general lesson:** one rule exists in two copies that disagree. The TS resolver has `comicChapter` but not `model3d`; the SQL `CASE` (`comment.notifications.ts:330-345`) has `model3d` but not `comicChapter`. **The bug lives exactly in the disagreement.** (Consolidating those two chains is filed separately — not attempted here.)

The test table helped hide it: it called itself *"every already-handled entity type"* with **9 rows against a 15-value enum**. It now reconciles against the real `commentConnectorSchema.entityType` enum and enumerates the four intentional exclusions (`question`, `answer`, `clubPost`, `model3dReview`), so the next added relation forces a decision instead of silently 404ing. A count that implies completeness without enumerating its exclusions is a claim, not a guard.

## The SSR seam

`comment-permalink.ssr.test.ts` drives `getServerSideProps` with a mocked `dbRead`: a resolved permalink becomes a **302 to the whole expected URL**, a null becomes `notFound`, and a **reply resolves through its root thread while keeping its own `threadId`**. The two components were each verified in isolation; this covers the seam that belonged to neither. The file states plainly what it *cannot* do, so nobody mistakes it for select coverage.

## Mutation testing — 14 mutants, all killed, zero false kills

Re-run after the rebase. Controls: unmutated → SURVIVED (61 passed); `appListing` arm disabled → KILLED.

Two rounds of harness self-correction, both worth recording:

- **Two mutants died on a `TypeError`**, from tests drilling into `.redirect.destination` — which throws when a mutant returns `notFound`. A mutant that dies for the wrong reason is a **false kill**. Assertions are now on the whole result object, and a **totality guard** was added (a malformed row returns `null` and never throws), so a dropped null-guard fails an assertion instead of exploding. Every one of the 14 now dies on an `AssertionError`.
- **Two "SURVIVED" verdicts were harness artifacts**, not weak tests: one from a `python3: No such file or directory` that meant the mutation never applied, and one (`page passes rootThread instead of thread`) that was **unreachable** because every SSR fixture had `rootThread: null`. Both were fixed — the second by adding the reply fixture, which is now the killing test.

## Verification

- `pnpm typecheck` → **0 errors** (post-rebase, after regenerating the Prisma client — see below)
- `vitest run src/server/utils/__tests__/` → **962 passed / 65 files**; with the ledger spec, **970 / 66**
- `model-file-hash-writers.test.ts` → **8 passed**, confirming the pre-existing `main` red cleared (run and counted, not read off the badge)

⚠️ Note for anyone rebasing this: the first post-rebase `pnpm typecheck` reported **15 errors in `Hubs`/`userHub` code**. Those were a **stale generated Prisma client**, not defects — `#4080` added models after my `pnpm install`. Re-running install regenerated the client and typecheck went to 0. Worth knowing before chasing them.

## 🔴 Not verified — stated plainly, please merge knowing this

**Nobody has watched this 307 fire for a real app-listing comment.** No one has opened `/comments/v2/<id>` for a comment on a real listing and seen it land on the listing page.

What *is* now proven, and what is not:

| | status |
|---|---|
| the select produces the fields the resolver reads | **type-proven** — `pnpm typecheck` fails otherwise |
| the resolver builds the right URL | 51 tests, 14 mutants, whole-string assertions |
| the page turns that URL into a 302 / a null into a 404 | 10 SSR tests against a mocked `dbRead` |
| **a real row exists with a populated FK, and a browser follows the redirect** | 🔴 **unobserved** |

The redirect *mechanism* is unchanged production code already serving nine entity types, and the data shape is compiler-enforced — so this is a **bounded-risk argument, not an observation**, and it should be read as exactly that. One manual click on a real app-listing comment closes it post-merge.

## Out of scope — filed separately

- **Consolidating the two entity chains.** The TS resolver and the SQL `CASE` are two copies of one rule and already disagree (`comicChapter` vs `model3d`); that disagreement is where the `model3d` bug lived. Not attempted here.
- **A disclosure question this PR extends but did not create.** `/comments/v2/[id]` has no auth gate and never consults `AppListing.status`, so an anonymous enumerator now gets `307 + Location: /apps/store-preview/<slug>` where they previously got 404 — for any status, including moderator-`removed`. The destination itself is safe, and this route has **never** gated on parent visibility for any entity (a comment on an unpublished post already 307s), so it is *consistent-but-extended* rather than novel. Small while the store is mod-only; it grows at the `/apps` cutover. Flagged for a product decision rather than settled in this PR.

### DIFF
--- a/src/pages/comments/v2/[id].tsx
+++ b/src/pages/comments/v2/[id].tsx
@@ -1,90 +1,22 @@
 import { dbRead } from '~/server/db/client';
 import { createServerSideProps } from '~/server/utils/server-side-helpers';
 import { PageLoader } from '~/components/PageLoader/PageLoader';
-import { buildCommentPermalink } from '~/server/utils/comment-permalink';
+import { buildCommentPermalink, commentPermalinkSelect } from '~/server/utils/comment-permalink';
 
 export const getServerSideProps = createServerSideProps({
   useSSG: true,
   useSession: true,
   resolver: async ({ ctx }) => {
     const { id } = ctx.params as { id: string };
-    const select = {
-      image: {
-        select: {
-          id: true,
-        },
-      },
-      post: {
-        select: {
-          id: true,
-        },
-      },
-      review: {
-        select: {
-          id: true,
-        },
-      },
-      model: {
-        select: {
-          id: true,
-        },
-      },
-      article: {
-        select: {
-          id: true,
-        },
-      },
-      bounty: {
-        select: {
-          id: true,
-        },
-      },
-      bountyEntry: {
-        select: {
-          id: true,
-        },
-      },
-      challenge: {
-        select: {
-          id: true,
-        },
-      },
-      comicChapter: {
-        select: {
-          projectId: true,
-        },
-      },
-      // The one SLUG-addressed thread parent. `appListingId` is `app_listings.serial_id`, an
-      // integer surrogate that appears nowhere in the URL, so the slug is what has to be
-      // selected here — see `buildCommentPermalink`.
-      appListing: {
-        select: {
-          slug: true,
-        },
-      },
-    };
+
+    // The select lives beside the resolver that reads it (`comment-permalink.ts`) rather than
+    // here, so the two cannot drift: the payload type is derived FROM this select, and the
+    // resolver's parameter type is derived from that payload. Narrow the select and the resolver
+    // stops compiling — which is the only thing that catches a select-correctness bug, since a
+    // mocked `dbRead` would just encode the same mistake in the fake.
     const commentV2 = await dbRead.commentV2.findUnique({
       where: { id: Number(id) },
-      select: {
-        id: true,
-        thread: {
-          select: {
-            id: true,
-            rootThread: {
-              select: {
-                id: true,
-                ...select,
-              },
-            },
-            comment: {
-              select: {
-                id: true,
-              },
-            },
-            ...select,
-          },
-        },
-      },
+      select: commentPermalinkSelect,
     });
 
     if (!commentV2) {
--- a/src/pages/comments/v2/[id].tsx
+++ b/src/pages/comments/v2/[id].tsx
@@ -1,7 +1,7 @@
 import { dbRead } from '~/server/db/client';
 import { createServerSideProps } from '~/server/utils/server-side-helpers';
 import { PageLoader } from '~/components/PageLoader/PageLoader';
-import { threadUrlMap } from '~/server/notifications/comment.notifications';
+import { buildCommentPermalink } from '~/server/utils/comment-permalink';
 
 export const getServerSideProps = createServerSideProps({
   useSSG: true,
@@ -54,6 +54,14 @@ export const getServerSideProps = createServerSideProps({
           projectId: true,
         },
       },
+      // The one SLUG-addressed thread parent. `appListingId` is `app_listings.serial_id`, an
+      // integer surrogate that appears nowhere in the URL, so the slug is what has to be
+      // selected here — see `buildCommentPermalink`.
+      appListing: {
+        select: {
+          slug: true,
+        },
+      },
     };
     const commentV2 = await dbRead.commentV2.findUnique({
       where: { id: Number(id) },
@@ -83,64 +91,7 @@ export const getServerSideProps = createServerSideProps({
       return { notFound: true };
     }
 
-    const { thread } = commentV2;
-    const getThreadDetails = (thread: any) => {
-      if (thread.post) {
-        return { threadType: 'post', threadParentId: thread.post.id };
-      }
-      if (thread.review) {
-        return { threadType: 'review', threadParentId: thread.review.id };
-      }
-      if (thread.model) {
-        return { threadType: 'model', threadParentId: thread.model.id };
-      }
-      if (thread.article) {
-        return { threadType: 'article', threadParentId: thread.article.id };
-      }
-      if (thread.bounty) {
-        return { threadType: 'bounty', threadParentId: thread.bounty.id };
-      }
-      if (thread.bountyEntry) {
-        return {
-          threadType: 'bountyEntry',
-          threadParentId: thread.bountyEntry.id,
-        };
-      }
-      if (thread.challenge) {
-        return { threadType: 'challenge', threadParentId: thread.challenge.id };
-      }
-      if (thread.comicChapter) {
-        return { threadType: 'comicChapter', threadParentId: thread.comicChapter.projectId };
-      }
-      if (thread.image) {
-        return { threadType: 'image', threadParentId: thread.image.id };
-      }
-
-      return { threadType: null, threadParentId: null };
-    };
-
-    let threadData: { threadType: string | null; threadParentId: number | null } =
-      getThreadDetails(thread);
-
-    if ((!threadData.threadType || !threadData.threadParentId) && thread.rootThread) {
-      threadData = getThreadDetails(thread.rootThread);
-    }
-
-    if (!threadData.threadType || !threadData.threadParentId) {
-      return { notFound: true };
-    }
-
-    const url = threadUrlMap({
-      threadParentId: threadData.threadParentId,
-      threadType: threadData.threadType,
-      threadId: thread.id,
-      commentId: commentV2.id,
-      // Always pin the target comment as the thread root so it renders standalone via
-      // RootThreadProvider's activeComment path. Sidesteps cursor-paginated thread fetches
-      // (target may be on page N) and works uniformly for root and reply comments.
-      commentParentType: 'comment',
-      commentParentId: commentV2.id,
-    });
+    const url = buildCommentPermalink({ thread: commentV2.thread, commentId: commentV2.id });
 
     if (url) {
       return {
--- a/src/server/utils/__tests__/comment-permalink.ssr.test.ts
+++ b/src/server/utils/__tests__/comment-permalink.ssr.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { dbMock } from '~/__tests__/mocks/db.mock';
 
 /**
  * `/comments/v2/<id>` — the SSR WIRING, as distinct from the resolution logic.
@@ -25,12 +26,23 @@ import { describe, expect, it, vi, beforeEach } from 'vitest';
  * is optional, so omission satisfies it. The `Pick` is the half that catches omission.)
  */
 
-const findUnique = vi.fn();
-
-vi.mock('~/server/db/client', () => ({
-  dbRead: { commentV2: { findUnique } },
-  dbWrite: {},
-}));
+/**
+ * 🔴 The CANONICAL db mock — this file must NOT register the db-client specifier itself.
+ *
+ * The shared-module mock is registered once in `src/__tests__/setup.ts` and reset between files;
+ * a direct per-file mock is rejected by `no-direct-shared-module-mock.test.ts`. The reason is not
+ * style: under `--no-isolate` a module instance is per WORKER, so an ordinary source module that
+ * imports `~/server/db/client` captures whichever mock was installed when it was FIRST evaluated
+ * — and every later file silently reuses it. A per-file mock poisons files that never mocked
+ * anything. See docs/testing/shared-module-mocks.md.
+ *
+ * A test file therefore declares BEHAVIOUR only; it never registers the specifier.
+ *
+ * (`no-direct-shared-module-mock.test.ts` detects violations by REGEX over the file text, so the
+ * offending call is described here rather than spelled — writing it even inside a comment trips
+ * the guard. Found the hard way.)
+ */
+const findUnique = dbMock.dbRead.commentV2.findUnique;
 
 // The page's `createServerSideProps` wrapper resolves a session; stub it so the test exercises the
 // resolver rather than auth.
--- a/src/server/utils/__tests__/comment-permalink.ssr.test.ts
+++ b/src/server/utils/__tests__/comment-permalink.ssr.test.ts
@@ -1,0 +1,166 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+
+/**
+ * `/comments/v2/<id>` — the SSR WIRING, as distinct from the resolution logic.
+ *
+ * `comment-permalink.test.ts` proves the resolver builds the right URL. This file proves the page
+ * actually TURNS that URL into a 302, and a null into a 404 — the seam between them, which no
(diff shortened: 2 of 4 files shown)


ZacxDev:
### `preview / smoke-tests` failure — attributed, not this PR

The single failure is **preview-environment Redis still warming**, not a regression here.

```
Error: tRPC mutation post.addImage -> HTTP 500:
[{"error":{"json":{"message":"LOADING Redis is loading the dataset in memory", …
```

Failing specs: `preview-post-images.spec.ts:32` (image post) and `preview-post-upload.spec.ts:36` (direct upload). The upload spec **recovered on retry #2** once Redis finished loading; the images spec exhausted its 3 retries. 64 passed.

Why it cannot be this PR:

- The only non-test importer of the new `comment-permalink` module is `src/pages/comments/v2/[id].tsx` — a single route. The failing specs create posts and upload images; neither ever visits `/comments/v2/<id>`.
- The comment-path smoke spec **passed**: `tester self-seeds a post, reacts to it, and comments on it › reaction.toggle and commentv2.upsert round-trip` ✓.
- Sibling PR #4184 (same session, does not touch this page) passed all 66.

Note this is also **distinct from the known ambient 4-failure pattern** (the moderator-surface specs bouncing to login) currently seen on #4187 and #4165 — this is a different, transient cause.

Every other check is green: 18 checks, `Unit tests`, `App unit tests`, `Package unit tests`, `ESLint + Prettier`, `Schema drift gate`, `tekton / typecheck`, `preview / deploy`, `preview / render-check` and the four report-only preview suites all pass.

ZacxDev:
### Re-trigger result: the smoke failure was transient. Confirmed.

Rebased onto current `main` (`7d35d9d082` → `93f9f6aea4`, content byte-identical — verified with `git diff` between the two heads over all three files, empty) and re-ran the preview.

**`preview / smoke-tests`: 66 passed, 0 failed, 0 flaky.** The previously-failing specs (`preview-post-images.spec.ts:32`, `preview-post-upload.spec.ts:36`) both pass. Same spec did not fail again, which is the discriminator — so the `LOADING Redis is loading the dataset in memory` attribution holds and it was preview-env warm-up, not this change.

Pipeline `pr-preview-4185-4629z` **Completed**, 18 tasks, 0 failed.

---

### The new `Unit tests` red is a PRE-EXISTING break on `main`, not this PR

The rebase pulled in newer `main`, and `Unit tests` now fails on one spec:

```
FAIL  src/server/services/__tests__/model-file-hash-writers.test.ts
      > ModelFileHash writer ledger > matches the writers actually present in the source tree
AssertionError: expected [ …(6) ] to deeply equal [ …(5) ]
```

**Run against clean `origin/main` with none of this PR's files present, it fails identically.** The extra entry is:

```
+ "packages/civitai-db-schema/prisma/migrations/20260819000000_model_file_hash_sha256_12/migration.sql
   :: INSERTINTO\"ModelFileHash\""
```

That migration landed after #4149 wrote its 5-entry ledger, so the ledger enumerates 5 while the tree contains 6. The guard is doing exactly its job — it fails when the writer set **grows** — but the growth is a migration on `main`, unrelated to comment permalinks.

This PR adds one new file (`src/server/utils/comment-permalink.ts`) and edits `src/pages/comments/v2/[id].tsx`; the ledger's four scan patterns are all `ModelFileHash`-specific (`modelFileHash.create*`, `INSERT INTO "ModelFileHash"`, `insertInto('ModelFileHash'`, `hashes: { create…`), none of which either file contains.

**Needs a separate one-line fix on `main`** — add the migration to `LEDGER_ENTRIES`, or exclude `prisma/migrations/` from the scan roots if migrations are not meant to count as writers. Happy to open that as its own PR; it is not this one's to make. Blocks every PR rebased past that migration until it lands.

PR #4188, opened 2026-08-20
fix(auth): fix the two type errors gating apps/auth, and bring it into the CI typecheck gate

ZacxDev:
`apps/auth` was the last app excluded from the per-app typecheck gate added in #4156. Both of its pre-existing errors are fixed at the root — no `any`, no `@ts-ignore`, no compiler-option loosening — and the exclusion is removed, so the gate now covers all **7** apps.

## The two errors

**`providers.ts:165` — `Parameter 'p' implicitly has an 'any' type`**

The stub provider's key is a *computed* property, `['stub' as ProviderId]`, whose type is the whole `ProviderId` union rather than a single literal. TypeScript therefore cannot match it to a member of `Record<ProviderId, ProviderDef>` and drops contextual typing for the value — which is why only this entry's `mapProfile` parameter was untyped while the four literal-keyed entries above were fine. Fixed with `satisfies ProviderDef`.

That turned out to be worth more than the one parameter. Measured both arms:

| `satisfies ProviderDef` | mutation: stub's `scope:` → `scopes:` | result |
|---|---|---|
| present | applied | `'scopes' does not exist in type 'ProviderDef'. Did you mean to write 'scope'?` — **caught** |
| removed | applied | **silently accepted**; only the implicit-`any` resurfaces |

So the entire stub entry was unchecked against `ProviderDef`, not just that parameter.

**`establish-session.test.ts:100` — `Property '_store' does not exist on type 'never'`**

The `Cookies` stub was cast `as never` to satisfy `establishSession`'s parameter. `never` has no properties, so reading `_store` back in an assertion was itself a type error. `Cookies` has exactly five members (`get`/`getAll`/`set`/`delete`/`serialize`), so the stub now implements all of them and is assignable with **no cast at all**.

## The real work: the guard suite

Emptying `EXCLUDED` breaks `scripts/__tests__/typecheck-apps.test.ts`. Measured: **6 of 8 cases fail**, because every fixture contained an `auth` app purely to satisfy the stale-exclusion guard, so an empty map made them fail for reasons unrelated to what they test.

Those guards are not deleted — they are what the *next* excluded app will depend on. Instead the exclusion map became a parameter of an exported `runTypecheckApps({ excluded })`, and the tests drive it with a synthetic app (`ghost`, `legacy`) through a test-only driver that imports the real module, so the injected cases still run the genuine script end-to-end.

Deliberately a **function parameter, not an env var**: an env-var override would be a live way to un-gate an app in CI, which is the exact failure class this script exists to close.

The suite is now 10 cases and decoupled from the shipped map. Verified by mutating the shipped map two ways:

- exclude an unrelated app (a plausible future change) → **0 failures**
- re-exclude `auth` → **exactly 1 failure**, the dedicated ledger test, on its own assertion

## Verification

Base ref for every "before" number: **`23cecb57c0`**.

**Red → green matrix**

| Check | at base `23cecb57c0` | at HEAD |
|---|---|---|
| `apps/auth` svelte-check | `COMPLETED 8037 FILES 2 ERRORS` | `COMPLETED 8037 FILES 0 ERRORS` |
| guard suite, `EXCLUDED` as shipped | 8 passed | 10 passed |
| guard suite, `EXCLUDED` emptied, tests unchanged | **6 failed / 2 passed** | n/a (tests reworked) |
| `apps/auth` unit tests | 372 passed | 372 passed |

**Positive control for the `0 ERRORS`.** A zero is indistinguishable from a check wired to nothing, so deliberate type errors were planted in *both* touched files and the run reported exactly them, at exactly the planted lines:

```
ERROR "src/lib/server/auth/providers.ts" 356:7 "Type 'string' is not assignable to type 'number'."
ERROR "src/lib/server/auth/__tests__/establish-session.test.ts" 118:7 "Type 'Map<string, string>' is not assignable to type 'string'."
COMPLETED 8037 FILES 2 ERRORS
```

Controls removed; the clean run reports the same **8037** files, so the check did not shrink. (File count matters: a run reporting `1 FILES` has checked nothing.)

**Mutation-tested the guards** — six mutants, each killed by the right test for its own reason:

| mutant | tests killed |
|---|---|
| stale-exclusion guard returns nothing | 1 (its own, on the exit code) |
| exclusion filter ignored | 2 (skip case, all-excluded case) |
| empty-targets guard returns 0 | 2 (both "refuses to report success" cases) |
| direct-run guard removed | 1 (the real-entrypoint case) |
| shipped map re-excludes `auth` | 1 (the dedicated ledger case) |
| shipped map excludes an unrelated app | **0 — correct**, proves the decoupling |

**Ledger.** `node scripts/ci/typecheck-apps.mjs` now prints:

```
Typechecking 7 app(s): auth, creator-studio, event-engine, moderator, notifications, orchestrator-gateway, storage
```

with **no `Skipping` line**.

## What I could NOT verify

`auth` typechecks clean, but I could not get a fully green 7/7 local run: `creator-studio`, `notifications` and `orchestrator-gateway` fail on my machine with `Module '"@prisma/client"' has no exported member 'Prisma'`. That is a local environment artifact — Prisma has no prebuilt engine for `linux-nixos`, so `postinstall`'s `db:generate` fails and the client is never generated. Evidence it is not this PR:

- every error is in `packages/civitai-db-schema` / `packages/civitai-db` or a downstream consumer of those types
- this PR touches 5 files, none under those three apps or under `packages/`
- those three apps were already un-excluded at base, and #4156's `App unit tests + typecheck` job passed in CI

CI runs on Linux with a working Prisma generate, so it is the authority on the 7/7 result. **Please confirm the `App unit tests + typecheck` job is green before merging** — that is the one claim I am relying on CI for rather than having measured myself.

Also note: my local runs used Node **v26.7.0** while `.nvmrc` pins **24.19.0** (pnpm warned `Unsupported engine` on every invocation). CI uses the pinned version.

## Incidental

- Neutralised the "six apps" counts in the script header and the workflow comment, which this change would otherwise make wrong (6 gated → 7).
- Ran prettier on the touched files, which also reformatted **two pre-existing** unformatted lines I did not otherwise modify (`providers.ts` `validatedStubUrl`, and the `user` helper in the test). The changed-files prettier step is report-only, but it would have started flagging them the moment this PR touched these files.
- ESLint reports two pre-existing `no-empty-function` errors in the test's `vi.hoisted` block. Left alone: CI's lint gate scopes to `^(src|packages|apps/event-engine)/`, so `apps/auth` is not linted, and they are unrelated to this change.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -420,13 +420,13 @@ jobs:
       # part and this job has already paid for it. The tradeoff is that a type
       # error renders under this job's name — hence the job rename above.
       #
-      # Deliberately ONE step driving a script, not six `pnpm --filter` steps.
-      # Six steps abort at the first red app, so a shared type change that breaks
-      # four of them reports one; and `pnpm --filter <name>` EXITS 0 WHEN THE
-      # NAME MATCHES NOTHING (measured, pnpm 10.28.1 — it prints "No projects
-      # matched the filters" and succeeds), so six hardcoded package names are
-      # six chances for a rename to turn this gate into a green no-op. The script
-      # derives the app set from disk, proves each filter selected a real
-      # package, and collects every failure. See its header.
+      # Deliberately ONE step driving a script, not one `pnpm --filter` step per
+      # app. Per-app steps abort at the first red app, so a shared type change
+      # that breaks several of them reports one; and `pnpm --filter <name>` EXITS
+      # 0 WHEN THE NAME MATCHES NOTHING (measured, pnpm 10.28.1 — it prints "No
+      # projects matched the filters" and succeeds), so a hardcoded list of
+      # package names is one chance per app for a rename to turn this gate into a
+      # green no-op. The script derives the app set from disk, proves each filter
+      # selected a real package, and collects every failure. See its header.
       - name: Typecheck apps
         run: node scripts/ci/typecheck-apps.mjs
--- a/apps/auth/src/lib/server/auth/__tests__/establish-session.test.ts
+++ b/apps/auth/src/lib/server/auth/__tests__/establish-session.test.ts
@@ -31,6 +31,7 @@ vi.mock('../device', () => ({
 
 import { establishSession } from '../session';
 import type { SessionUser } from '@civitai/auth';
+import type { Cookies } from '@sveltejs/kit';
 
 const user = (id: number): SessionUser =>
   ({
@@ -43,18 +44,30 @@ const user = (id: number): SessionUser =>
     createdAt: new Date(),
     isModerator: false,
     muted: false,
-  }) as unknown as SessionUser;
+  } as unknown as SessionUser);
 
 // A minimal Cookies stub: in-memory get + a set that records the queued cookie.
-function makeCookies(initial: Record<string, string> = {}) {
+//
+// It implements the WHOLE `Cookies` interface (get/getAll/set/delete/serialize — five members) rather than
+// casting a partial object, so it is assignable to `establishSession`'s parameter with no cast at all. The
+// previous `as never` satisfied that parameter the blunt way, but `never` has no properties, so reading
+// `_store` back in an assertion was itself a type error.
+type CookiesStub = Cookies & { _store: Map<string, string> };
+
+function makeCookies(initial: Record<string, string> = {}): CookiesStub {
   const store = new Map(Object.entries(initial));
-  const set = vi.fn((name: string, value: string) => store.set(name, value));
   return {
     _store: store,
-    set,
+    set: vi.fn((name: string, value: string) => {
+      store.set(name, value);
+    }),
     get: (name: string) => store.get(name),
-    delete: vi.fn(),
-  } as never;
+    getAll: () => [...store].map(([name, value]) => ({ name, value })),
+    delete: vi.fn((name: string) => {
+      store.delete(name);
+    }),
+    serialize: (name: string, value: string) => `${name}=${value}`,
+  };
 }
 
 beforeEach(() => vi.clearAllMocks());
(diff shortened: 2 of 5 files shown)


ZacxDev:
Independently verified the two claims that matter, from the CI logs rather than the check buckets.

**The ledger ran all seven, `auth` included:**

```
Typechecking 7 app(s): auth, creator-studio, event-engine, moderator, notifications, orchestrator-gateway, storage
  ::group::typecheck auth (@civitai/auth-app)
  ::group::typecheck creator-studio (@civitai/creator-studio-app)
  ::group::typecheck event-engine (@civitai/event-engine)
  ::group::typecheck moderator (@civitai/moderator-app)
  ::group::typecheck notifications (@civitai/notifications-app)
  ::group::typecheck orchestrator-gateway (@civitai/orchestrator-gateway)
  ::group::typecheck storage (@civitai/storage-app)
All 7 app(s) passed typecheck.
```

Seven named, seven groups executed, and no `Skipping` line — `EXCLUDED` is genuinely empty rather than the entry merely being renamed.

**The restructured guard suite executes in CI, at its new count:**

```
✓ unit scripts/__tests__/typecheck-apps.test.ts (10 tests) 398ms
```

That settles the one thing the author could not measure locally: `creator-studio`, `notifications` and `orchestrator-gateway` failing on their box with `Module @prisma/client has no exported member 'Prisma'` was the NixOS Prisma-engine artifact, not a real defect. A runner that generates the client typechecks all three clean.

**10 of 11 checks pass, zero failures**; only `preview / deploy` is outstanding.

Two notes for the reviewer, neither blocking:

- The `providers.ts` fix is larger than the reported error, and the PR is right to treat it that way. The computed key `['stub' as ProviderId]` types as the whole union, so contextual typing was dropped for that entry — the implicit `any` was a symptom, and the whole stub entry had never been checked against `ProviderDef`. The two-arm mutation in the description (misspelling `scope`→`scopes` is caught with `satisfies`, silently accepted without) is the evidence for that, and it is the part worth reading.
- Making `excluded` a function parameter rather than an env var is the right call. An env-var override would be a live way to un-gate an app in CI; a parameter is only reachable from tests.

One caveat worth keeping in view: the guard suite runs in the `unit` job, which is `continue-on-error: true`, so a regression there renders red-but-ignored. `preview / unit-tests` is the non-ignorable one carrying the weight.

PR #4189, opened 2026-08-20
fix(typecheck): typecheck scripts/ — 15 non-test files now covered, scripts/__tests__ excluded like src/

ZacxDev:
## What this does

`scripts/` was invisible to the typechecker. The root `tsconfig.json` named `scripts/local-dev/*.ts` in **both** `include` and `exclude` — exclude wins, so **zero** files under `scripts/` were in the program, not even the one directory the list appeared to name.

Demonstrated, not asserted. The same deliberate `const x: number = 'nope'` planted in `scripts/oneoffs/parse_header.ts`:

| config | result |
|---|---|
| `main`'s `tsconfig.json` | `typecheck: OK — 0 type errors`, **exit 0** |
| this PR's `tsconfig.json` | `parse_header.ts(112,7): error TS2322`, **exit 2** |

This PR adds `scripts/**/*.ts(x)` to `include`, and excludes two things: `scripts/__tests__/**` and `scripts/local-dev/gen_seed.ts`.

## The measurement

Adding all of `scripts/` surfaces **202 errors across 6 files** (of 37). Measured with a config that `extends` the real one and only *adds* to `include`, so the program stays a strict superset — a hand-rolled narrowed program drops `src/types/global.d.ts`, ambient names collapse to TS2304, and the affected files report *fewer* errors than they have. Same reasoning already written into `tsconfig.tests.json`.

```
scripts/__tests__/dev-server-test-queue.test.ts        164
scripts/__tests__/dev-server-env-modes.test.ts          14
scripts/__tests__/typecheck-tests-gate.test.ts          11
scripts/__tests__/dev-server-port-reservation.test.ts    7
scripts/__tests__/dev-server-daemon-port.test.ts         4
scripts/local-dev/gen_seed.ts                            2
                                                       ---
                                                       202
```

Five of the six are in `scripts/__tests__/`, and that is structural rather than a backlog.

## Why `scripts/__tests__/**` is excluded as a directory

An earlier revision of this PR quarantined four `__tests__` files individually and ran the ratchet as a blocking CI job. **#4181 broke it within a day** by adding an ordinary new test, `dev-server-daemon-port.test.ts`, carrying 4 × TS2345.

Cause, read at the source: that directory exists to test the untyped `.mjs` modules under `.claude/skills/` and `.claude/hooks/` by importing them. None carries JSDoc, so under `allowJs` tsc infers their signatures from the implementation and every caller inherits the result.

```js
export function daemonPortsGuarded(env = process.env) {   // no JSDoc
```
infers as requiring a full `ProcessEnv`, so `daemonPortsGuarded({ DEV_DAEMON_PORT: '9555' })` is an error. It is the **same mechanism** as the 164: `request({ worktree, args = [] } = {})` infers as `{ args?: never[] }` (40 × TS2353) and a run object's `null` initializers infer as type `null` (123 × TS18047/TS2531).

Frequency, measured from git history rather than guessed: **5 of the 15 files there carry this class, and all 15 were added within one month.** A blocking gate over that directory would fire several times a month, always for a defect in a dependency the test author did not write, and the fastest way out would be to add another exclude line — training exactly the reflex the quarantine existed to prevent.

So the directory is excluded the same way `src/**/__tests__/**` already is, and **the gate is no longer wired into CI**. Its job is removed; `.github/workflows/lint.yml` is byte-identical to `main`.

## What is covered, and what is not

**Covered** — 15 non-test files, now checked by the real `pnpm typecheck` in **both** tiers, along with any file added to these directories later. Counted with `--listFilesOnly`, not asserted:

| directory | files |
|---|---|
| `scripts/oneoffs/` | 6 |
| `scripts/local-dev/` | 3 |
| `scripts/test-perf/` | 2 |
| `scripts/*.ts` | 2 |
| `scripts/metric-migration/` | 2 |

> An earlier revision of this PR claimed 31. That was wrong: it counted `scripts/__tests__` files and transitively-imported `.mjs`. 15 is the honest number.

🔴 **Not covered** — `scripts/__tests__/**` is **unchecked**. That is the pre-existing state rather than a regression, but this PR does not fix it and should not be read as doing so. It is *measured*: `tsconfig.scripts.json` re-includes it and `node scripts/ci/typecheck-scripts-gate.mjs` prints per-file counts on demand. Typing the `.mjs` modules is the real fix and is the natural follow-up — the baseline is how you watch that number fall.

`scripts/local-dev/gen_seed.ts` is excluded by name. Its 2 errors are **a real bug the blind spot was hiding**, not an inference artifact: it imports `~/server/db/notifDb`, deleted in `927e31bfee`. `notifDbWrite` now lives in `apps/notifications` behind a different API (a function returning a pool), so restoring the script is a cross-workspace migration.

## The gate

Kept as an on-demand measurement instrument — the same status as its sibling `typecheck-tests-gate.mjs`, whose five controls it **imports rather than copies** (`classifyEmptyAllowance` gains a defaulted `envName`/`gateScript` so each names its own). Since nothing runs it automatically, its exit 1 is a report, not a merge block.

The one control genuinely new here: its positive control anchors at the repo root rather than matching the substring `/scripts/`, because this repo pulls `.claude/skills/**/scripts/*.mjs` in transitively and every vendored package has a `scripts/` directory — a substring control could stay green with the whole measured tree gone.

## Verification

- **Positive control**: planted error in `scripts/oneoffs/` → caught, exit 2. The remaining coverage is real, not a vacuous glob.
- **Negative control**: a new deliberately-broken test file dropped into `scripts/__tests__/` → **0** root errors. The tripwire is gone. The gate still reports that same file, so it is excluded, not invisible.
- **Ratchet both ways against real `tsc`**: `gen_seed.ts` 2→3 BLOCKS (exit 1, names the delta); untouched tree PASSES (exit 0).
- **New invariant test, watched fail**: a baseline entry for a root-*covered* file is rejected by name — so this baseline cannot be used to park a failure `pnpm typecheck` is actually red on. Ships with a negative control proving its predicate can reject.
- **Mutation battery**: 10 mutants, one per control, each killed by its specifically named test, with unmutated-green controls at both ends. The drift mutant was re-run against the new glob-shaped quarantine and is killed by 3 tests including the end-to-end one.
- Root `pnpm typecheck`: **0 errors**. Both gate suites: **190 passed**. Prettier + ESLint clean.

## Still unverified / known-unrelated

- **The sibling tests gate is already BLOCKED on `main`** (869 errors across 160 files vs its 784/140 baseline). Verified at the base ref with my changes reverted — identical counts, pre-existing, and it is not wired into CI. Its baseline wants a `--write-baseline` refresh; out of scope here.
- The gate script itself is a `.mjs` spawned as a subprocess, never imported by a `.ts`, so it is in no program and is not typechecked — same as its sibling. Its pure logic lives in `typecheck-scripts-compare.mjs`, which *is* checked and unit-tested.
- I did not run the full unit suite locally, only the two gate suites, the root typecheck, and the gate end to end.
- Three configs plus the gate's `QUARANTINE` list must stay in step when the exclusions change. Both gates refuse to run (`CANNOT MEASURE`, exit 3) rather than mismeasure when one is forgotten — that control is what caught the drift here.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -241,55 +241,6 @@ jobs:
       - name: Typecheck
         run: pnpm run typecheck
 
-  typecheck-scripts:
-    name: Typecheck scripts/ (quarantine ratchet)
-    runs-on: ubuntu-latest
-    # UNCONDITIONAL, unlike the `typecheck` job above. That one is narrowed to
-    # fork PRs and non-main bases because the in-cluster Tekton `pr-check`
-    # pipeline covers everything else — but Tekton runs `pnpm run typecheck`,
-    # i.e. the ROOT tsconfig, and this gate measures `tsconfig.scripts.json`.
-    # Nothing else runs it, so a narrowing `if:` here would mean it never ran at
-    # all on an internal PR.
-    #
-    # 🔴 READ THIS BEFORE ASSUMING THIS JOB IS WHAT COVERS scripts/. It is not.
-    # `tsconfig.json` now includes `scripts/**/*.ts(x)`, so the 33 clean files
-    # there are checked by the ROOT typecheck and BOTH tiers already block on
-    # them — including every file added under `scripts/` in future. This job
-    # exists for the FIVE quarantined files listed in that config's `exclude`,
-    # which no other tier can see. Deleting it does not uncover the 33; it
-    # uncovers the 5.
-    #
-    # Its own job rather than a step on `unit`: that job is `continue-on-error`
-    # (report-only by design), so a ratchet parked there could never block.
-    timeout-minutes: 25
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          # `tsconfig.scripts.json` inherits `src` from the root config, and src
-          # imports event-engine-common. Without the submodule those resolve to
-          # 12 TS2307/TS7006 errors. The gate reports them as "outside scripts/"
-          # and drops them rather than mis-scoring them, so the verdict would
-          # still be correct — but the log becomes noise around it, and the
-          # submodule is public and needs no secret.
-          submodules: true
-
-      - uses: pnpm/action-setup@v4
-
-      - uses: actions/setup-node@v4
-        with:
-          node-version-file: .nvmrc
-          cache: pnpm
-
-      - name: Install
-        run: pnpm install --frozen-lockfile
-
-      # Drives `scripts/typecheck.mjs` internally — the wrapper that sets the V8
-      # heap cap and classifies the outcome, so a heap abort reads as CRASHED
-      # rather than as a clean zero. The gate additionally refuses (exit 3) on a
-      # measurement it cannot trust, which is neither a pass nor a failure.
-      - name: Scripts quarantine ratchet
-        run: node scripts/ci/typecheck-scripts-gate.mjs
-
   unit:
     name: Unit tests
     runs-on: ubuntu-latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -241,6 +241,55 @@ jobs:
       - name: Typecheck
         run: pnpm run typecheck
 
+  typecheck-scripts:
+    name: Typecheck scripts/ (quarantine ratchet)
+    runs-on: ubuntu-latest
+    # UNCONDITIONAL, unlike the `typecheck` job above. That one is narrowed to
+    # fork PRs and non-main bases because the in-cluster Tekton `pr-check`
+    # pipeline covers everything else — but Tekton runs `pnpm run typecheck`,
+    # i.e. the ROOT tsconfig, and this gate measures `tsconfig.scripts.json`.
+    # Nothing else runs it, so a narrowing `if:` here would mean it never ran at
+    # all on an internal PR.
+    #
+    # 🔴 READ THIS BEFORE ASSUMING THIS JOB IS WHAT COVERS scripts/. It is not.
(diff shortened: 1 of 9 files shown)


ZacxDev:
CI note — the one red **Unit tests** failure is not from this PR.

`src/server/services/__tests__/model-file-hash-writers.test.ts > matches the writers actually present in the source tree` fails on the **merge** commit, not on this branch.

Isolated to a single variable:

| tree | result |
|---|---|
| this branch alone | 5 passed |
| this branch + only `…/20260819000000_model_file_hash_sha256_12/migration.sql` taken from `main` | that exact test fails, 4 passed |

The cause is `245e90ff70` ("update docs"), which is docs-only: it added a comment block to that migration containing `INSERT INTO "ModelFileHash"`, taking the file from 0 to 4 mentions. The ledger scanner's raw-SQL matcher matches that text **inside a SQL comment**, so it registers a writer that cannot execute and the `toEqual` on the full ledger set fails.

Two things worth separating:

1. **Not caused here, and not blocking** — the `unit` job is `continue-on-error: true` by design. Nothing in this PR touches the ledger, its scanner, or its roots; my change is `tsconfig` `include`/`exclude` plus new files under `scripts/ci/`.
2. **It will fail on every PR opened against current `main`** until either the ledger is regenerated or the matcher strips SQL comments. Stripping comments looks like the right fix — a ledger that fires on prose is the "spelled rather than structural" failure its own header warns about — but that belongs to whoever owns that guard, not to this PR.

Everything else is green, including `tekton / typecheck` (the authoritative tier, which now covers `scripts/`) and the new `Typecheck scripts/ (quarantine ratchet)` job.

PR #4190, opened 2026-08-20
fix(dev-server): give a queued test run a FILE, not a pipe — a child that exits without flushing was losing most of its log

ZacxDev:
A queued test run's output was captured through a **pipe**. Node makes a child's stdout synchronous when it refers to a regular file and **asynchronous when it refers to a pipe**, so a child that calls `process.exit()` discards whatever is still queued — before the parent ever receives it.

## The measurement that picked the fix

Head-to-head: the **same child**, in the same process, through the old mechanism and the new one. Two separate runs would not be a comparison.

```
consumer=slow  PIPES received= 172 missing=4828  |  FILE received=5000 missing=0
consumer=slow  PIPES received= 172 missing=4828  |  FILE received=5000 missing=0
consumer=slow  PIPES received= 172 missing=4828  |  FILE received=5000 missing=0
consumer=fast  PIPES received=1473 missing=3527  |  FILE received=5000 missing=0
consumer=fast  PIPES received=5000 missing=   0  |  FILE received=5000 missing=0
consumer=fast  PIPES received=5000 missing=   0  |  FILE received=5000 missing=0
```

The lines are **numbered**, which is what makes this readable: a bare count cannot tell truncation from sampling, and here every missing line is a **contiguous tail** — that is what identifies the mechanism rather than just confirming loss.

It also **ruled out both fixes the ticket proposed**:

- *"the daemon detects a short read on child exit"* — it cannot. The daemon sees a clean EOF and has no signal to compare against, because the bytes never reached the pipe.
- *"flush child stdout before exit"* — not ours to call. The child is vitest.

A file needs no cooperation from the thing losing the data.

## A second, independent defect on the same path

With **every byte delivered**, the reader still corrupted the log. `d.toString().split('\n')` per chunk, with no carry buffer, tore any line straddling a read boundary into two and recorded **both halves as lines**: 5,000 in → 4,998 recognised, 6 fragments invented, 353,893 of 353,893 bytes. `logsDropped` reports 0 for this, correctly — which is exactly what made it invisible behind a counter readers had been trained to trust.

## A third thing, in neither ticket

The final drain now runs **before** `onExit`. A waiter wakes on the run's terminal status and *then* reads the log, so reporting the exit while lines are still unread yields a complete-*looking* log that is still filling — the same false green, reached a different way. The old path could do this, because `'exit'` does not wait for pending `'data'`.

## Trade-off, stated rather than buried

One file for both streams, so the interleaving a reader depends on is the real one. The cost: stdout and stderr are no longer distinguishable. Lines are therefore recorded as `output` rather than claiming to be one or the other — nothing renders the level for a test run (both consumers print `entry.message`), and a label we cannot support is worse than an honest one.

## Verification

Seven mutants, each applied alone, tree `cmp`-verified between each:

| mutant | killed by |
|---|---|
| carry buffer removed | the straddle case (+2 more) |
| final flush removed | the trailing-line case |
| back to pipes | both seam cases |
| `onExit` before the drain | the ordering case |
| drain reads a single chunk | the 5,000-line case + straddle |
| stderr to a different fd | the shared-descriptor case |
| capture file not deleted | the cleanup case |

All killed. The ordering mutant is killed **only** by the ordering test, which is why that test exists — the capture can be perfect and the queue can still report the exit too early.

`scripts/__tests__` — **344 passed** (15 files). `pnpm typecheck` — 0 errors. New test file prettier-clean.

**Not verified:** Windows. The mechanism is `stdio` fds and `fs.readSync`, neither platform-specific, and the queue already routes through a shell there — but that is a design claim, not a measurement.

Closes ClickUp 868ktwgyc and 868kubfd9.

### DIFF
--- a/.claude/skills/dev-server/scripts/test-queue.mjs
+++ b/.claude/skills/dev-server/scripts/test-queue.mjs
@@ -17,6 +17,7 @@ import { closeSync, openSync, readSync, unlinkSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { randomUUID } from 'crypto';
+import { StringDecoder } from 'string_decoder';
 
 export const DEFAULT_CONCURRENCY = 1;
 const DEFAULT_ABANDON_AFTER_MS = 10 * 60 * 1000;
@@ -83,9 +84,16 @@ export function createOutputCapture(onLine) {
   // torn in two and BOTH halves are recorded as lines — which corrupted the log even when every
   // byte arrived: 5,000 lines in, 4,998 recognised, 6 fragments invented, 353,893/353,893 bytes.
   let carry = '';
+  // A carry buffer fixes a torn LINE and does nothing for a torn CHARACTER. Decoding each read
+  // independently turned a 3-byte `⎯` starting at byte 65535 into THREE U+FFFD — and vitest's
+  // failure output is built from `⎯`/`✓`/`×`/`❯`, one boundary per 64 KiB. The decoder holds an
+  // incomplete sequence back until the bytes that finish it arrive.
+  const decoder = new StringDecoder('utf8');
+
+  // Hoisted: the tail runs 10x a second per run, and this was a fresh 64 KiB allocation each time.
+  const buf = Buffer.allocUnsafe(64 * 1024);
 
   const drain = (final = false) => {
-    const buf = Buffer.allocUnsafe(64 * 1024);
     for (;;) {
       let read = 0;
       try {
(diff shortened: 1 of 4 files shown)


ZacxDev:
**The red `Unit tests` here is not this PR.** `main` is red, and every open PR inherits it.

The failure is `ModelFileHash writer ledger > matches the writers actually present in the source tree` — 1 failed / 19,457 passed. Controlled: the identical failure reproduces at `9c373dfb4e` in the same tree with this branch's commit absent, and this branch touches neither `src/` nor any migration.

Cause and fix are in #4191 — the ledger was counting a **commented-out** `INSERT` in a migration whose only live statement is an `ALTER TYPE`. Once that merges, this PR needs a re-run to get a real signal; until then the check here says nothing about this change.

PR #4193, opened 2026-08-20
fix(tests): apply the comment strip to .ts too — the ledger still counted commented-out writers

ZacxDev:
Follow-up to #4191, supplying the evidence that PR explicitly asked for before extending the strip past `.sql`:

> Scoped to `.sql` deliberately. The same argument applies to a commented-out `.ts` writer, but stripping there could only ever SHRINK the enumeration, which is the direction this ledger is also meant to catch — so **that half needs its own evidence, not a drive-by**.

That caution was right. The evidence now says the *unstripped* side was the broken one.

## The measurement

A file containing nothing but a comment:

```ts
// await dbWrite.modelFileHash.createMany({ data: rows });
export const zzProbe = 1;
```

is enumerated as a **live writer**. Measured by planting exactly that at `74bd61e6d8` and watching the ledger assertion fail:

```
+   "src/server/services/__zz_probe_writer.ts :: modelFileHash.createMany",
```

So `.ts` was never the safe default — it is the same defect in a different comment syntax, and `.sql` only reached it first because a docs commit happened to add SQL-shaped prose.

## The shape underneath, which is the part worth keeping

`stripComments` has existed and been unit-tested in this file all along, while `enumerateWriters` ran its patterns over the **raw** source and never called it. There is even a test named *"strips comments before looking for the call (control for the check below)"* — and the check below did not strip. Both halves covered; the seam between them owned by nobody. The stripper's own tests stayed green the whole time.

## Both directions pinned

Stripping can hide a real writer as easily as it can stop counting a fake one, so the tests are a pair — the second exists to fail if the first is ever "fixed" by stripping too hard.

| case | base | with fix |
|---|---|---|
| commented-out writer (`.ts .tsx .js .mjs .cjs`) | counted | not counted |
| LIVE writer with a trailing comment | — | still counted |
| LIVE writer after `i--;` | — | still counted |

That last one is why the `.sql` rule must not leak into JS: `--` is a **decrement operator** there, so applying it would swallow the rest of the line and hide a real write.

## Verified against the real enumerator, not just the stripper

That distinction is the whole point of the bug, so the proof had to be end-to-end:

- planting a **commented** writer: `rc=1` at base → `rc=0` with the fix
- planting a **LIVE** writer: `rc=1` with the fix — still caught
- the ledger set over the actual tree is **unchanged**, which is the shrink control that matters most

`stripSqlComments` is kept as an alias so any external caller keeps working; `.sql` behaviour is byte-identical.

### DIFF
--- a/src/server/services/__tests__/model-file-hash-writers.test.ts
+++ b/src/server/services/__tests__/model-file-hash-writers.test.ts
@@ -47,6 +47,8 @@ const SKIP_DIRS = new Set([
   'coverage',
 ]);
 const SCAN_EXTENSIONS = new Set(['.ts', '.tsx', '.js', '.jsx', '.mjs', '.cjs', '.sql']);
+/** The scanned extensions that take JS comment syntax — i.e. everything but `.sql`. */
+const JS_EXTENSIONS = new Set(['.ts', '.tsx', '.js', '.jsx', '.mjs', '.cjs']);
 
 /**
  * Four spellings a write can take. A single Prisma-only pattern would miss a raw-SQL or Kysely
@@ -76,14 +78,33 @@ const WRITE_PATTERNS: Array<[kind: string, re: RegExp]> = [
  * enumeration that writes nothing. It made trunk red for every PR until someone either declared a
  * non-writer in the ledger or deleted the documentation, and both would have been wrong.
  *
- * Scoped to `.sql` deliberately. The same argument applies to a commented-out `.ts` writer, but
- * stripping there could only ever SHRINK the enumeration, which is the direction this ledger is
- * also meant to catch — so that half needs its own evidence, not a drive-by.
+ * Originally scoped to `.sql` alone, with the `.ts` half deferred for evidence rather than taken
+ * as a drive-by — the concern being that stripping could only ever SHRINK the enumeration, which
+ * is a direction this ledger also exists to catch. That evidence now exists, and it says the
+ * unstripped side was the broken one: a file containing NOTHING but
+ *
+ *     // await dbWrite.modelFileHash.createMany({ data: rows });
+ *
+ * was enumerated as a live writer (measured 2026-08-20 by planting exactly that and watching the
+ * ledger assertion fail with the probe file listed). So `.ts` was not the safe default; it was the
+ * same defect with a different comment syntax, and the `.sql` fix simply reached it first.
+ *
+ * The shrink risk is real, so it is controlled rather than argued away — see the two tests below
+ * that pin BOTH directions: a commented writer must not be counted, and a live writer on a line
+ * carrying a trailing comment must still be.
  */
-export function stripSqlComments(source: string, ext: string): string {
-  if (ext !== '.sql') return source;
-  return source.replace(/\/\*[\s\S]*?\*\//g, ' ').replace(/--[^\n]*/g, ' ');
+export function stripCommentsForExt(source: string, ext: string): string {
+  if (ext === '.sql') {
+    return source.replace(/\/\*[\s\S]*?\*\//g, ' ').replace(/--[^\n]*/g, ' ');
+  }
+  // `--` is a DECREMENT operator in JS/TS, so the SQL rule must never run here: `i--; foo();`
+  // would lose the rest of the line and hide a real writer.
+  if (JS_EXTENSIONS.has(ext)) return stripComments(source);
+  return source;
 }
+
+/** Kept as the old name so any external caller keeps working; `.sql` behaviour is unchanged. */
+export const stripSqlComments = stripCommentsForExt;
 
 const isTestFile = (relPath: string) =>
   relPath.includes('__tests__') || /\.(test|spec)\.[cm]?[jt]sx?$/.test(relPath);
@@ -116,7 +137,7 @@ function enumerateWriters(): string[] {
   for (const file of files) {
     const rel = path.relative(REPO_ROOT, file);
     if (isTestFile(rel)) continue;
-    const source = stripSqlComments(fs.readFileSync(file, 'utf8'), path.extname(file));
+    const source = stripCommentsForExt(fs.readFileSync(file, 'utf8'), path.extname(file));
     if (!source.includes('odelFileHash')) continue; // cheap prefilter, case-insensitive on the M
     for (const [, re] of WRITE_PATTERNS) {
       re.lastIndex = 0;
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Correcting my own PR description — it overstated the evidence, and the change had two real defects

An adversarial audit found two 🔴 issues in the revision above. Both were mine, both were in the direction this file itself calls the worse one, and I've reworked the PR (`40097254da`). Correcting the record here rather than silently editing the body, since anyone who already read it read the wrong version.

### 1. My fix introduced a silent under-report — strictly worse than the bug it fixed

`stripComments` ran its **block pass before its line pass**, so a block-open appearing inside a `//` comment opened a region swallowing live code up to the next block-close. The trigger is ordinary in this repo: a package glob contains one.

```ts
// ported onto the shared @civitai/* clients.
export async function write(dbWrite) {
  await dbWrite.modelFileHash.createMany({ data: rows });
}
/** Re-exported for tests. */
```

That strips to the **empty string**. I confirmed it directly. Measured on this tree: **43 such sites across 40 files, hiding 1,591 source lines** from the enumerator; worst case an entire 294-line module body.

So a writer skipping `normalizeScanHashes` could have landed inside one of those windows with the ledger green — the precise silent failure this file exists to prevent. Fixed by cutting line comments first, so the block pass only ever sees real block comments: **43 over-match sites → 0**, all tests still green.

Honest scope: that fixes the measured population, **not the class**. A block-open inside a string literal in live code still opens a spurious region. Recorded in the code, with a pointer to the string-aware stripper in `oauth-client-scope-grants.test.ts`.

### 2. The seam was still unpinned — and I wrote a comment claiming it wasn't

My previous revision added two tests that call `stripCommentsForExt` **directly**, plus a comment asserting they *"drive the enumerator's transform rather than the stripper alone, which is the only arrangement that could have caught it."*

That was false. Leaving the stripper perfectly correct and reverting only the **call** in `enumerateWriters` — reintroducing the exact defect this PR fixes — left the whole suite green.

I reproduced the shape I was fixing and then documented the opposite. A comment is a claim, and that one would have been cited by the next maintainer.

`enumerateWriters` now takes `root`/`scanRoots` so a test drives it over a fixture tree. Mutants, each run rather than reasoned about:

| mutant | killed by |
|---|---|
| revert the call, stripper left correct | `drives the real enumerator over a fixture tree, in both directions` |
| block pass first (defect 1) | that test, **and** `does not let a block comment opened inside a line comment swallow live code` |

The first is the mutant that survived the audited revision.

### Three claims in my description that were wrong

- **"Verified against the real enumerator, not just the stripper — the proof had to be end-to-end."** The end-to-end proof was a one-off manual plant; nothing shipped encoded it, which is exactly why the revert mutant survived. Now it *is* encoded.
- **"`stripSqlComments` is kept as an alias so any external caller keeps working."** There are **no** external callers — the only other occurrences are a local function of the same name in an unrelated oauth test. Dead code under a name that had stopped being true (it strips JS comments too). Dropped.
- **"the ledger set over the actual tree is unchanged, which is the shrink control that matters most."** True but the **weakest** available control — a statement about today's tree, and the tree is the input that changes. The fixture test is the real control.

### Also from the audit

- `JS_EXTENSIONS` is now **derived** from `SCAN_EXTENSIONS` rather than a second hand-written list, so an extension added to one can't fall through unstripped in the other.
- Recorded, not fixed: **`.mts` (12 files under the scan roots) and `.cts` are in neither set**, so a writer in one is invisible to the ledger — at base and at head. Pre-existing; widening the scan enlarges the enumeration and wants its own evidence.

The audit's remaining note stands: `.jsx` is in the extension set but the direct-call tests iterate 5 of 6, so deleting `.jsx` survives. Inert today — 0 `.jsx` files under the scan roots — and the fixture test now covers the wiring regardless.

ZacxDev:
## Closing this — the cure is worse than the disease, measured three times

Three fix rounds, three adversarial audits, and every round of mine introduced a real defect worse than the one it fixed. The decisive fact came last, and I should have established it first:

**Across 4,742 non-test files under `src/`, `apps/`, `scripts/` and `packages/`, there are ZERO commented-out `ModelFileHash` writers.**

The evidence that opened this PR was a probe **I planted**. I created the input, then fixed it. This PR has been fixing a hypothetical.

### What each attempt actually cost, against real files

| attempt | effect on the real tree |
|---|---|
| block pass first | 43 sites / 40 files / **1,591 live lines** hidden from the enumerator |
| line pass first *(shipped)* | strictly worse — retention 81.1% → 70.8%; `ecosystem-seo.constants.ts` lost **3,816 of 3,838** live lines |
| `ts.createScanner` *(shipped)* | **42.6% of `//` lines unstripped across 1,865 files**; hides a live writer in `` const p = `${a}/*` `` that **both** condemned regex variants keep |

The direction matters. The defect this PR set out to fix is the **noisy** one — a commented-out writer shows up as a spurious ledger entry, the test goes red, a human looks. Every cure I shipped produced the **silent** one, which is the direction [the file itself calls worse](https://github.com/civitai/civitai/blob/main/src/server/services/__tests__/model-file-hash-writers.test.ts) and the direction #4191 explicitly declined to take without evidence.

### The mistake I kept repeating

Each round I measured the population I was **fixing** and never the population I was **breaking**. Round 3 added a blast-radius bound specifically to stop that — and I calibrated it against the broken scanner's own output. `blocks.router.ts` "has 6,730 live lines" only according to the instrument under test; two independent instruments put it near 4,156. So the control written to prevent the mistake **encoded** it, and would have rejected a correct implementation. Same error, third instance, this time inside the guard against it.

### What stays

**#4191's `.sql` stripping stays and was right.** It fixed a real problem: a real migration file documented its backfill as commented SQL, which put a phantom sixth writer in the enumeration and reddened trunk for every PR. Real instance, real cost, correct fix. Its author scoped it to `.sql` deliberately and asked for evidence before extending it. The evidence now exists and it says: don't extend it.

### If someone picks this up later

A correct stripper needs a real parser (`ts.createSourceFile`), not a scanner and certainly not regexes — a standalone `scan()` can't resume a template literal after `${…}` because that needs `reScanTemplateToken()`, which only the parser calls. It measured at roughly **+18 s** on a unit-tier file that runs on every PR. Weigh that against a defect with **zero** occurrences.

If it is ever attempted again, the non-negotiable is a control measured with an instrument **independent of the implementation under test**, applied repo-wide rather than to a single hand-picked file — the one I got wrong all three times.

Branch `zach/fix-ledger-comment-seam` is preserved if anyone wants the parser work as a starting point. Nothing here blocks anything: `main` is green, the ledger's answer is unchanged under every variant tried, and the real `.sql` fix is already in.

PR #4202, opened 2026-08-20
feat(observability): expose Flipt eval-cache stats — a hit rate alone cannot pick the right knob

ZacxDev:
## Problem

The per-eval TTL cache in `@civitai/flipt` tracked `hit` internally and **threw the result away**. From outside the process it was a black box: you could not tell a cache doing its job from one thrashing.

It has two tuning knobs — `FLIPT_EVAL_CACHE_TTL_MS` and the `evalCacheMaxEntries` ceiling — and no signal for choosing between them.

## Why a hit rate isn't enough

The two failure modes are **indistinguishable from a hit rate** and have **opposite remedies**:

| symptom | diagnosis | remedy |
|---|---|---|
| misses dominated by **expired** entries | TTL-bound — the key *was* cached, its TTL lapsed | longer TTL converts them to hits |
| **rotations** climbing | capacity-bound — entries evicted *before* they can expire | longer TTL recovers **nothing**; raise the ceiling |

The second case is the trap: raising the TTL against a capacity-bound cache is an **inert change that reads as a fix**.

Worth noting the cache key is `(flag, entityId, context)`, so a per-user `entityId` multiplies the key space by the active-user count — capacity-bound is the *likelier* of the two on hot paths, and is exactly what a bare hit rate would hide.

## What this adds

- Cumulative counters on `TtlCache`: `hits`, `misses`, `expiredMisses`, `rotations`, `size`, exposed via `stats()`.
- `getCacheStats()` on the client.
- `civitai_app_flipt_eval_cache_*` on the default registry `/api/metrics` already scrapes, registered by side-effect import there (matching its neighbours) so the series exist from the **first** scrape — an absent series would read as "the cache is idle" rather than "nobody loaded the module".

**No new package dependency**: the counters are plain numbers; all prom-client wiring lives app-side.

`misses - expiredMisses` is documented as an *upper bound* on cold traffic, not a measurement — separating cold from evicted would require retaining evicted keys, which defeats the eviction.

## Verification

Every new guard was **mutation-tested**, and each mutant died by its **own named test** (not a neighbour's), with the source restored byte-identical and the control re-run green:

| mutation | killed by |
|---|---|
| `expiredMisses` bumped on every miss | cold-vs-expired test |
| rotation counted on every `set` | overflow test |
| promoted read not counted as a hit | promotion test |
| `reset()` dropped from `collect()` | double-count test |
| boolean stats reported for both labels | per-kind test |

The `reset()` one matters most: without it every scrape adds the running total to itself and the series grows quadratically while the cache is idle. The test scrapes three times against a constant source and asserts the value does not move.

**Typecheck**: 317 errors on `origin/main` and 317 on this branch, with **byte-identical error sets** (zero branch-only errors) — that baseline is a stale generated Prisma client in my environment, not this change.

**Lint**: clean on the touched files and on `packages/civitai-flipt`, confirmed with a negative control that the linter actually processed them (a planted unused var went red first).

## Follow-up

This is deliberately **observability only** — no tuning value is changed. The point is to make the next tuning decision measurable instead of guessed.


ZacxDev:
## Correction: this PR's observability was itself half-blind

An adversarial audit found two 🔴 issues. Both verified independently before fixing. **Fixed in `dba1080a93`.**

### 🔴 The metric measured ONE of TWO caches per pod

`src/server/flipt/client.ts` is emitted **twice** in the production server build. Measured, not theorised — `[flipt] eval cache TTL:` appears exactly **2× on every pod** (488 lines across 244 pod streams) against **one** `[instrumentation] Running in nodejs runtime`.

Each emitted copy owned a private wasm client, a private 60s config poller, and a private cache pair. `getFliptCacheStats` closed over whichever copy its chunk resolved.

**Worse than a scale error, because the deliverable is a knob decision and the bias has a direction.** Rotations are superlinear in per-instance key space, so one key space split across two caches at the same ceiling rotates far less than one cache holding all of it. The split reads as **"TTL-bound"** → raise the TTL → **the exact inert change this PR exists to prevent.**

This is the same shape #4173 (one commit back on `main`) reverted the pg-pool gauges for, and why `request-bulkhead.ts` is enrolled in `server-graph-watchlist.mjs`.

**Fix:** pin on `globalThis.__civitaiFliptClient` with the repo's canonical `??=` idiom, and enrol the module as `SHARED_STATE` so a refactor dropping the pin fails that gate. Side benefit: one wasm engine and one config poll per pod instead of two.

### 🔴 Prettier was failing — CI red

The PR body said "Lint: clean on the touched files." That was true of **eslint** and false of **prettier**. I ran a careful negative control **on the wrong instrument**, which is a neat demonstration of the thing this repo keeps warning about.

### Three defects in this PR's own tests

| defect | why it mattered |
|---|---|
| **The seam was untested.** Deleting the side-effect import in `src/pages/api/metrics.ts` left the whole `src/server/metrics/` suite green (**131/131**) | registered-but-unreached — this repo's #1 metric-death mode. Now asserted by loading the module that *serves* the scrape, matching the substitutions and bitdex seam tests. |
| **One asserted cell was vacuous.** `read()` ended `?? 0`, so asserting the variant cache's `expiredMisses` is 0 passed whether or not the series existed | the absent-vs-zero ambiguity this metric exists to remove, reintroduced in the test guarding it. Returns `NaN` on a missing label now. |
| **Over-claim in the body:** "each mutant died by its own named test" | removing `reset()` is killed by *two* tests, because `read()` triggers another `collect()`. |

### Re-verified after the fixes

Each mutant killed by its **own** named test, sources restored byte-identical, controls green:

| mutation | killed by | before |
|---|---|---|
| delete the side-effect import | seam test | **SURVIVED** |
| gate the variant inc on non-zero | per-kind test | **SURVIVED** |
| `??=` downgraded to `=` | watchlist gate | new |

28/28 green across the watchlist gate, the metrics suite and the flipt suites.

### Audit items deliberately not actioned

- **Counter + reset-then-inc vs a Gauge with `set()`** — the auditor verified the Counter version is *correct as written* (against vendored prom-client 14.2.0), but noted `set()` gives the same property for free and matches the freshly-audited precedent in `src/server/prom/client.ts`. Worth a reviewer's opinion; I left it as a Counter since `rate()`/`increase()` semantics are what readers will reach for. Happy to switch.
- **`entries` help invites comparing against the ceiling**, but `size` sums both generations while rotation triggers on `current` alone, so a healthy cache reads ~2× its ceiling. Should be reworded.
- **`misses` doesn't count bypassed flags** (4 today) or evals while the client is `null`. Worth stating in the help string.

PR #4209, opened 2026-08-20
fix(models): ?earlyAccess=false no longer switches the filter on

JustMaier:
Closes ClickUp [868ku8g1k](https://app.clickup.com/t/868ku8g1k).

## The bug

`/api/v1/models` parses `getAllModelsSchema` against raw `req.query`, where every value is a string. `z.coerce.boolean()` is JS `Boolean()` there, and `'false'` is a non-empty string — so `?earlyAccess=false` parsed as `true` and switched the early-access filter **on**. The caller asking to exclude early access got exactly the opposite set, with no error and no warning. tRPC callers send real booleans, which is why it survived.

The fix already existed two lines below: `booleanString()`, used for `paidAccess` and `newCreators` with the reasoning recorded beside it.

## Scope — the whole column, with reachability stated per param

Every boolean in `getAllModelsSchema` is reachable from the public REST endpoint: the handler does `modelsEndpointSchema.safeParse(req.query)` over the whole query object, with no key filtering (the only gate is a 401 when `favorites`/`hidden` are sent unauthenticated).

**Fixed here — the parse was wrong on all six.** But the *effect* differs, and the distinction matters for anyone sizing the incident: two of them are gated downstream, so only four returned a wrong set to an ordinary public caller.

| Param | Reachable | Effect of `=false` before this PR |
|---|---|---|
| `earlyAccess` | public | **wrong set returned** — the reported bug |
| `supportsGeneration` | public | **wrong set returned** |
| `fromPlatform` | public | **wrong set returned** |
| `archived` | public | **wrong set returned** |
| `needsReview` | public | parsed `true`, then **gated on `sessionUser?.isModerator`** in both consumers (`model.service.ts:550`, `:1298`) — no effect for a non-moderator |
| `followed` | public | parsed `true`, then **gated on a signed-in user** (`:612`, `:1320`) — no effect for an anonymous caller |

So: **four ungated public-facing bugs, not six.** `booleanString()` is right for all six regardless — a moderator or a signed-in user sending `=false` got the wrong set too.

**Moved for consistency, not a live bug** — `favorites`, `hidden`. Both were already shielded at REST by a `booleanString()` override in the endpoint's own `modelsEndpointSchema`.

> ⚠️ **Those two overrides in `src/pages/api/v1/models/index.ts` are now redundant but harmless, and this PR deliberately leaves them in place.** They restate what the shared schema now says, so they change nothing. Removing them is filed separately; do not delete one as drive-by cleanup here.

**Reachable but already correct** — `paidAccess`, `newCreators` (fixed in #4141 and earlier).

**Different bucket, deliberately not changed** — `pending`, `disablePoi`, `disableMinor`, `isFeatured`, `poiOnly`, `minorOnly`, `allowNoCredit`, `allowDerivatives`, `allowDifferentLicense` are plain `z.boolean()`. A query string is *rejected* there, so the endpoint 400s: unusable over REST, but it fails closed rather than returning the wrong set.

**Out of scope, reported for separate tickets** — `image.schema.ts` (`fromPlatform`, `notPublished`, `scheduled`) and `post.schema.ts` (`scheduled`, `video`) still use `z.coerce.boolean()`; they look tRPC-only, since `/api/v1/images` parses its own schema which already uses `booleanString()` throughout. Two admin endpoints have the same defect on a money path and are filed as [868kun1w5](https://app.clickup.com/t/868kun1w5).

## Tests

**The test discovers its own field list, over BOTH schemas.** A test that enumerates the fields it checks has the same blind spot as the bug — it cannot see the next field someone declares on `z.coerce.boolean()`. This one reads `.shape`, keeps every key that accepts a real boolean, and asserts none of them parses the string `'false'` as `true`.

It runs over `getAllModelsSchema` **and** over `modelsEndpointSchema`, the extended schema the endpoint actually parses. An override reintroducing `z.coerce.boolean()` at the endpoint is the same live bug, and it is invisible to a test that only knows the shared schema. Proven by mutation: coercing the endpoint's own `nsfw` override turns it red with `?nsfw=false parsed as true`.

### 🔴 If you copy this test, copy all three guards with it

A schema-driven test buys its coverage by giving up a hand-written list — which means **it can silently end up covering nothing.** Three guards make that visible:

1. **Discovery runs inside the test, never at module scope.** A `z.preprocess` field can *throw* rather than return an issue — `commaDelimitedNumberArray` calls `.map` on whatever it is handed, so `safeParse({ ids: true })` throws outright. At module scope that throw failed *collection*: the file reported as a **pass with zero tests**. Same shape as a missing submodule hiding a whole suite.
2. **Non-vacuity names the fields, and does not count them.** Discovery keeps anything that tolerates a boolean, so `Number(true) === 1` pulls in `limit`, `page` and `rating` — a numeric floor is satisfied by fields this bug cannot touch, even if every field it *can* touch dropped out. The guard asserts the query-string booleans by name.
3. **Both polarities are asserted.** A field that rejected everything would satisfy "never parses `'false'` as true" perfectly. `'true'`, `'false'` and a real boolean are each pinned.

**Both consumers are asserted, not reasoned about.** The parsed value is read by two separate truthy checks — the raw-SQL `EXISTS` in `getModelsRaw` and the Prisma `id IN` in the `getModelsPagedSimple` path — so the test drives a query string through the schema into each one and asserts the query itself, with a positive control on each and a `toHaveBeenCalledTimes(1)` pin so an early return cannot pass as a clean negative.

**Revert legibility, checked by mutation.** Run against the unfixed schema, the suite fails with `AssertionError: ?favorites=false parsed as true: expected true not to be true` — an assertion naming the wrong value, not a timeout.

**One test was deleted, and its coverage moved.** `get-models-raw.paid-access-filter.test.ts` had a hand-listed `paidAccess` parse describe that the derived loop now covers. Two copies of one invariant, and the hand-listed copy was the exact blind spot this file argues against. A pointer comment says where it went.

## Verification

- `pnpm exec eslint <changed files>` — no issues (repo-wide `lint` is noisy on ~1,400 pre-existing files)
- `pnpm exec prettier --write <changed files>`
- `pnpm run test:unit:run` — full suite green; `blocks.router.workflow.test.ts` collects **350 tests**, so the submodule is initialised and the run is meaningful rather than a silent zero-collection pass

See the comment below for the full before/after table of input handling, and the decision on the empty value.

### DIFF
--- a/src/pages/api/v1/models/index.ts
+++ b/src/pages/api/v1/models/index.ts
@@ -37,7 +37,7 @@ export const config = {
 
 const authedOnlyOptions: Array<keyof GetAllModelsInput> = ['favorites', 'hidden'];
 
-const modelsEndpointSchema = getAllModelsSchema.extend({
+export const modelsEndpointSchema = getAllModelsSchema.extend({
   limit: z.preprocess((val) => Number(val), z.number().min(0).max(100)).default(100),
   nsfw: booleanString().optional(),
   primaryFileOnly: booleanString().optional(),
--- a/src/server/schema/__tests__/get-all-models.boolean-params.schema.test.ts
+++ b/src/server/schema/__tests__/get-all-models.boolean-params.schema.test.ts
@@ -1,0 +1,54 @@
+import { describe, it, expect } from 'vitest';
+import { getAllModelsSchema } from '~/server/schema/model.schema';
+
+/**
+ * /api/v1/models parses this schema straight off `req.query`, where every value is a
+ * string. `z.coerce.boolean()` is JS `Boolean()` there, so `?earlyAccess=false` parsed
+ * as `true` and switched the filter on — the caller got the opposite of what it asked
+ * for, with no error.
+ *
+ * The keys are discovered from the schema rather than listed, because a hand-listed set
+ * has the same blind spot as the bug: it cannot see the next field someone declares on
+ * `z.coerce.boolean()`.
+ */
+
+// A `z.preprocess` field can THROW rather than return an issue (commaDelimitedNumberArray
+// calls .map on whatever it is handed), and at module scope that would fail collection —
+// zero tests, reported as a pass. Treat a throw as "does not accept this value".
+function parseOne(key: string, value: unknown) {
+  try {
+    const result = getAllModelsSchema.safeParse({ [key]: value });
+    return result.success ? { ok: true as const, value: (result.data as Record<string, unknown>)[key] } : null;
+  } catch {
+    return null;
+  }
+}
+
+function booleanKeys() {
+  return Object.keys(getAllModelsSchema.shape).filter((key) => parseOne(key, true) !== null);
+}
+
+describe('getAllModelsSchema — boolean params over raw query strings', () => {
+  // Non-vacuity: if discovery silently returned nothing, or stopped reaching the fields
+  // this is about, every assertion below would pass over an empty list.
+  it('discovers the boolean params, including the ones this bug was found on', () => {
+    expect(booleanKeys()).toEqual(expect.arrayContaining(['earlyAccess', 'paidAccess']));
+    expect(booleanKeys().length).toBeGreaterThanOrEqual(8);
+  });
+
+  it('never parses the STRING `false` as true, for any boolean param', () => {
+    for (const key of booleanKeys()) {
+      const parsed = parseOne(key, 'false');
+      // Rejecting the string is fine — that fails closed with a 400. Accepting it as
+      // `true` is the bug.
+      const value = parsed ? parsed.value : '<rejected>';
+      expect(value, `?${key}=false parsed as ${String(value)}`).not.toBe(true);
+    }
+  });
+
+  it('still parses `true` and real booleans as true', () => {
+    expect(getAllModelsSchema.parse({ earlyAccess: 'true' }).earlyAccess).toBe(true);
+    expect(getAllModelsSchema.parse({ earlyAccess: true }).earlyAccess).toBe(true);
+    expect(getAllModelsSchema.parse({ earlyAccess: 'false' }).earlyAccess).toBe(false);
+  });
+});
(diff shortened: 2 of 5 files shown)


JustMaier:
## Verification

`pnpm run test:unit:run` — **19,731 passed, 27 skipped, 0 failed** (exit 0, 357s). `blocks.router.workflow.test.ts` collected **350 tests**, so the submodule is initialised and the run is meaningful rather than a silent zero-collection pass.

## One behaviour change, decided rather than assumed

`booleanString()` is stricter than `z.coerce.boolean()`, so three inputs change. Measured against the fixed schema, not reasoned:

| `?earlyAccess=` | before | after |
|---|---|---|
| `false` / `0` / `no` / `off` | **true** (the bug) | `false` |
| `true` / `1` / `yes` / `on` / `TRUE` | true | true |
| `banana` | true | 400 |
| *(empty value)* | false | 400 |
| repeated param | true | 400 |

Junk and repeated params becoming a 400 is a straight improvement — they were returning a wrong set silently. The empty value is the one that needed a call, and Justin made it: **require an explicit value.** Reasoning worth recording, since it is the argument against treating empty as `false` — `?earlyAccess=` reads as *asking for* early access, so if it ever meant anything it should mean `true`, not `false`. Requiring the explicit value avoids picking a side.

Nothing in this repo sends an empty value for these params: no in-repo caller hits `/api/v1/models` with them, and the one place the app re-emits a caller's params (`getNextPage`, which echoes `req.query` through `QS.stringify`) cannot produce one, because such a request 400s at parse before reaching it. That negative is bounded to this repo — it says nothing about external SDKs or third-party scripts, which would need a look at prod request logs.

`?favorites=` and `?hidden=` have already behaved this way in production for some time, via the `booleanString()` override in the endpoint's own schema.

JustMaier:
Full suite after the review fixes, on `aa9f95bd2a`: **19,732 passed, 27 skipped, 0 failed** across 1,252 files (exit 0, 233s). `blocks.router.workflow.test.ts` collected **350 tests**, so the submodule is initialised and this is a real run rather than a silent zero-collection pass.

Net +1 test against the previous run: three added by the endpoint-schema coverage and the both-polarity assertion, two removed with the superseded hand-listed `paidAccess` describe.

Wall clock went **down** (357s → 233s), so importing the endpoint module in the schema test did not cost the suite anything measurable — that file's ~66s of import time when run alone is largely shared with files that already pull the same graph.

Issue #4197, opened 2026-08-20
Negative assertions can race an unsettled DOM (was: "not.toBeInTheDocument() is inert" — that diagnosis was wrong)


ZacxDev:
> ## 🔴 CORRECTED 2026-08-20 — the original diagnosis in this issue was WRONG
>
> **This issue originally claimed `expect.element(...).not.toBeInTheDocument()` "never fails, for any string", and recommended replacing it with `expect(locator.query()).toBeNull()`.**
>
> **Both halves were wrong, and the recommended replacement is *worse* than what it replaces.** The original text is preserved at the bottom, unedited, because several PRs were reviewed against it and a reader may have already acted on it.

## What is actually wrong

The matcher is **sound**. It retries, and it goes red. Measured on Chrome 149.0.7827.55, `@vitest/browser` 4.0.18:

| cell | result |
|---|---|
| element **present**, `.not.toBeInTheDocument()` **immediately after `render()`** | **PASSED in 6 ms** |
| element **present**, same assertion **after the render settled** | **FAILED in 14,864 ms** — correct |
| element **present**, `.elements().length` **immediately after `render()`** | **0** — a false zero |

The defect is an **unsettled-DOM race**, not an inert matcher. `.not` resolves on the *first* observation in which the element is absent — and immediately after `render()` the DOM is always empty. The assertion is therefore satisfied before the component has rendered anything at all.

That also explains the earlier "both directions pass on the same locator" observation, and the 7 ms injected-element result that was read as proof of inertness. **6–7 ms is the race signature.** A real evaluation of this matcher takes ~15 s to fail.

## 🔴 Why the original remedy makes it worse

`expect(locator.query()).toBeNull()` and `.elements()` **do not retry at all**. They are single point-in-time reads.

So an unpaired `.query()` placed where the old assertion sat is **strictly more exposed to the race** — it cannot recover from an unsettled DOM even in principle, where the original at least *could* have. Mechanically migrating all sites to `.query()` would **preserve every false pass** and remove the possibility of ever catching them.

## The correct remedy: settle, then assert

The negative assertion must follow something that proves the render has completed:

```ts
// settle: an awaited positive assertion on something that IS rendered
await expect.element(page.getByText('Visit', { exact: true })).toBeInTheDocument();
// now the negative is meaningful
expect(page.getByText('Standalone', { exact: true }).query()).toBeNull();
```

[...]

PR #4218, opened 2026-08-20
fix(apps): restore the Unpublish/Republish pair #4154 dropped from /apps/mine, + an author-action ledger

ZacxDev:
## This is a regression from #4154, not a new feature

A tester reported: *"no unpublish button on the my app"* — they wanted to unpublish an already-approved app and there was no way to do it.

**It is ours.** PR #4154 consolidated `/apps/my-submissions` into `/apps/mine`. The owner unpublish/republish controls lived in `MySubmissionsList.tsx`, which that merge orphaned. `MyAppsBody.tsx` — the new page body — contained **zero** occurrences of `unpublish`.

The uncomfortable part, recorded here because the audit trail is worth more than the tidier story:

- The implementing agent **disclosed it at the time** — *"no inline analytics, install counters, deploy-failure panel, or unpublish/republish controls"* — and it was filed as a noted gap rather than treated as a functional regression.
- **Three audit rounds missed it.** Every round asked *"is the new page correct?"*. None asked *"is it complete?"* — and no per-assertion suite can answer the second question, because a control that is simply absent has nothing to assert against. A tester found it instead.

**What this PR does about that, beyond restoring the buttons:** the class of defect is *a consolidation silently drops an author affordance and passes every audit*, and no amount of extra per-assertion review closes it. So this adds a **ledger** — an enumerated set of the controls each row state offers, compared against what the page actually renders, failing when the set **shrinks** as well as when it grows. A future consolidation that drops Unpublish again produces a red test instead of a tester report. Details and the mutation proof below.

Scope here is the **Unpublish + Republish pair only**. Inline analytics, install counters and the deploy-failure panel stay deferred and are explicitly not in this PR.

---

## The fix

**Republish is not optional.** `republishOwnListing` is the only owner-reachable route from `removed` back to `approved`; without it, an unpublish is a one-way door that only a moderator `relistListing` can reopen. Both halves ship together.

Per-state behaviour, adapted from `MySubmissionsList` rather than lifted from it (the new page is row-based and, unlike either source list, **dual-kind**):

| Row state | Controls | Badge |
|---|---|---|
| approved (live) | **Unpublish** | `approved` |
| owner-unpublished | **Republish** | `unpublished` |
| moderator-removed | neither | `removed by a moderator` + "contact them to restore it" |
| draft / pending / rejected | neither | the plain status |
| seated collaborator, any state | neither | — |

A seat is not ownership: both procedures are owner-scoped server-side, so an editor offered either control gets a guaranteed red toast.

**Dual-kind confirmation copy.** `/apps/my-submissions` was two lists, each passing a *fixed* copy variant to `OwnerUnpublishModal` — `offline` for on-site (a full takedown; the app stops serving at its app page) and `store` for off-site (a delist; the author's own site keeps running). `/apps/mine` merges both kinds into one table, so the variant is derived **per row**. A fixed variant would have told half of all authors something untrue about what the button was about to do — the one thing a straight lift from either source page would have got wrong.

### Server change

`listMyAppListings` now carries `lastModerationAction` for `removed` rows. This is not polish: `app_listings.status = 'removed'` is written by **both** an owner self-unpublish and a moderator takedown, and only the last moderation event separates them. Without the field the page must either hide Republish from an author entitled to it, or render a button guaranteed to 403. One batched `distinct` + `orderBy desc` query over the removed subset only — the same shape `blocks.router::listMyPublishRequests` and the off-site `listMySubmissions` already use. The query is skipped entirely when nothing is removed, so the common case stays at one hydrate query.

`app-listings.router.ts` is **not** touched: `listMine` already just delegates to the service.

---

## The ledger test — the part that prevents recurrence

A one-off restore fixes this instance. The **class** is *a consolidation silently drops an author affordance and passes every audit*.

`myAppsAuthorActions.ts` declares, per row state, the exact set of controls the row offers. `MyAppsBody.authorActions.browser.test.tsx` renders each state, **enumerates every interactive control inside the row's action container**, and asserts **set equality** against that table. Three properties make it a real ledger rather than another assertion:

1. It fails when the set **shrinks** — the #4154 shape.
2. It fails when the set **grows** — a new control must be declared.
3. It **refuses** any control in the container that carries no `data-author-action`, so the growth arm cannot be walked by forgetting the attribute. (The enumeration selects `button, a[href]` and *then* reads the attribute — selecting on `[data-author-action]` directly would have made that arm inert.)

The component calls `showUnpublish`/`showRepublish` from the same module the ledger is keyed on, and a unit test asserts those two agree with the table across all 4 states × 2 roles — otherwise the ledger would be pinning itself rather than the page.

Same writer-ledger pattern as `model-file-hash-writers.test.ts`.

### Teeth — proved, not asserted

Six mutants, each applied, **confirmed landed by diff**, run, and killed on its **own** assertion (all `AssertionError`, none a `TypeError`):

|---|---|---|
| delete the Unpublish button | live-state ledger (+4 more) | `expected [ 'history' ] to deeply equal [ 'unpublish', 'history' ]` |
| delete the Republish button | owner-hidden ledger (+3) | `expected [ 'history' ] to deeply equal [ 'republish', 'history' ]` |
| **swap** the owner-hidden / mod-removed arms in `ownerListingState` | owner-hidden **and** mod-removed ledgers (+3); **4 unit tests too** | both directions, incl. `expected [ 'republish', 'history' ] to deeply equal [ 'history' ]` |
| add an undeclared `<button>` to the row | all 6 ledger cases | `1 control(s) … carry no data-author-action and are therefore invisible to the ledger` |
| Republish fires with `row.slug` instead of the id | the wiring test, alone | call-list mismatch |
| drop the owner role gate from `showUnpublish` | the collaborator case, alone | `expected [ 'unpublish', 'history' ] to deeply equal [ 'history' ]` |

Baseline re-run after restoring: **11/11 green** (positive control).

---

## Test matrix

| Suite | Base | HEAD | Kind |
|---|---|---|---|
| `MyAppsBody.authorActions.browser.test.tsx` (11) | **8 failed / 3 passed** | 11 passed | **regression** — behavioural red |
| `app-access.my-app-listings-moderation.test.ts` (4) | **3 failed / 1 passed** | 4 passed | **regression** — 3 red at base; the 4th is labelled in-file as an **invariant guard** (passes vacuously on `main`) |
| `__tests__/myAppsAuthorActions.test.ts` (14) | n/a | 14 passed | **new-behaviour guard**, labelled as such |

"Red at base" for the browser file is stated as *"against a tree whose `RowActions` renders only the History toggle — which is exactly what `origin/main` offers an author"*, because a literal `origin/main` checkout of the component fails on a missing export instead, which would be a much weaker claim. The two assertion messages measured in that run are quoted verbatim in the file's header.

Regression suites, all green:
- `unit` over `src/server/services/blocks/__tests__/` + `src/components/Apps/__tests__/`: **188 files / 4058 tests**
- `component` over `src/components/Apps/`: **71 files / 983 tests**
- `pnpm typecheck`: **0 errors** (read from the reported count, not the exit code)
- `prettier --check` and `eslint` clean on every changed file

### The earlier `component-tests` red was a load flake — proven by re-run, not argued

The first run of this branch failed one spec. Re-triggered by rebasing onto current `main`; the **same spec passed**, which is the discriminator that separates "load" from "still open":

| | failing run (`9afed638`) | passing re-run (`cf4bce5e02`) |
|---|---|---|
| result | 1 failed / 1915 passed | **1917 passed** |
| `MySubmissionsList … History button` | **15289 ms** (timeout) | **773 ms** ✅ |
| tests > 5 s | 80 | 38 |
| slowest test | 68.8 s | 40.9 s |
| `Duration` / `tests` / `import` / `setup` | 215 / 941 / 715 / 268 s | 107 / 548 / 346 / 103 s |

Every phase roughly halved and the >5 s population halved — **whole-run inflation, not one-assertion inflation**, which is the shape that distinguishes a starved box from a real failed assertion. The spec itself is normally a 773 ms test; it did not become slow, the box did. Mechanism: `expect.element(…).toBeInTheDocument()` polls on vitest's **1000 ms default** — `vitest.config.mts` raises `testTimeout`/`hookTimeout` to 60 s but never `expect.poll`, so under starvation the matcher gives up while the run is still healthy. Corroborating: `pr-preview-4173` failed the same tier the same day at 14977/14949/14982 ms on an unrelated file, and #4219 ran on the same box in the same minute at the same ~9× and passed 1906/1906.

Instrument checks, because a reassuring zero is not evidence: the component runner was validated with a positive control first (an unmodified `MyAppsBody.browser.test.tsx` run reporting **56 tests executed**, not 0) before any verdict here was believed; `prettier --check` was confirmed able to go red (it flagged 3 of the 9 files on the first pass). Negative assertions use `expect(locator.query()).toBeNull()` — **not** `expect.element(...).not.toBeInTheDocument()`, which is inert in this repo (#4197) — placed after the awaits that settle the render, and each cluster of nulls is paired with a positive control asserting something that *is* on screen is found by the same mechanism.

---

## Audit round — what changed in `0234822`

| # | Finding | Fix |
|---|---|---|
| 🟡 1 | `orderBy: createdAt desc` load-bearing and unasserted; `desc`→`asc` **SURVIVED**, reachable but unfeelable through a single-event fake | Added `historyFake` (honours `orderBy` + `distinct`) and the case that matters — `delist` → `relist` → `owner-unpublish`. The mutant now dies **behaviourally**: `expected 'other' to be 'owner-unpublish'` — the strand-an-entitled-author direction, not merely the structural assertion. Structural `orderBy` check added too. |
| 🟡 2 | `RowActions` doc claimed "EVERY author-facing control on a row" — false; the edit link and History-panel Withdraw sit outside it | **Chose narrowing, not widening.** The comment now says "actions cell", names both excluded controls, and says why: Withdraw only exists once a row is expanded *and* its history has resolved, so a row-wide ledger would need per-expansion, per-fetch expected sets — turning a deterministic guard into one whose expected value waits on a query. Widening is the stronger end state and is recorded as worth doing; it is not a one-liner, and an intermittently-red ledger is worth less than none. |
| 🟡 3 | Projection widened what a seated **editor** receives (`claim`, `purge`, `report-*`) | Server now normalises to `'owner-unpublish' \| 'other' \| null`. New test collapses four pairwise-distinct real verbs to one constant, with a positive control proving `owner-unpublish` is *not* collapsed. Deleting the normalisation kills 2 tests. |
| 🟡 4 | "Removed by a moderator — contact them to restore it" can be false (a closed report writes an event row) | Now **"Only a moderator can restore this listing."** — states the consequence, which is true in every case that reaches it, instead of narrating a cause. The shared `ownerStateChip` badge has the same flaw and is deliberately untouched: it renders on two other surfaces, so rewording it is a cross-surface change, not this PR's. |
| 🟡 5 | Row vanishes into collapsed *Inactive* the instant the author unpublishes | The section now **auto-opens on success**. Chose this over "name the section in the toast" because that toast lives in the shared `OwnerUnpublishModal`, where the other two surfaces have no Inactive section — naming it there would be wrong for them. Driven through the real confirm flow in test, and the mock now invokes `onSuccess` so the assertion is not vacuous. |
| 🟢 | `myAppsAuthorActions` claimed the component calls `authorRowActions` — it calls the per-control predicates | Comment now names the real mechanism (the seam test), and says why it matters: a reader who believed the stronger claim would have deleted that test as redundant. |

**Mutation-checked, each landing confirmed by diff and each dying on its own assertion:**

- `orderBy` `desc`→`asc` → `expected 'other' to be 'owner-unpublish'` (1 failed / 5 passed)
- normalisation removed → `expected [ 'delist', 'purge', 'claim', …(1) ] to deeply equal [ 'other', 'other', 'other', 'other' ]` (2 failed / 4 passed)
- auto-open disabled → the reveal test fails (1 failed / 11 passed)

**Recorded, not fixed** (per audit): the page-level `republishing` state disables every Republish button while one is in flight; there is no `owner-hidden` + `editor` DOM fixture (the blocking tier already pins that predicate across all four states × both roles); the ledger runs at one viewport. The pre-existing unhandled rejection in `ItemResellersPopover.browser.test.tsx` is filed as **#4223** — and note it is *also* load-correlated: present in the loaded run, absent in the quiet one.

---

## Known limitations / follow-ups, stated rather than absorbed

- 🔴 **OPEN: two components disagree about what `status = 'removed'` MEANS, and this PR does not resolve it.** This is a modelling disagreement, not a UI nit, and it should be settled deliberately rather than papered over:

  - `myAppsView.ts::partitionMyAppRows` treats `removed` as **terminal** — one of "the two `app_listings.status` values an app cannot come back from on its own" — and routes every such row into the default-collapsed *Inactive* group. It reads `status` and nothing else, by an explicit 🔴 documented invariant with its own test.
  - `offsiteOwnerControls.ts::isModRemovedListing` takes the **opposite** position for the same column: it deliberately EXCLUDES an owner-unpublished listing from the removed bucket, on the stated grounds that it "stays Republish-able and keeps its normal (Live) placement", and routes only *moderator* takedowns aside.

  Both are load-bearing and both are documented as intentional. They cannot both be right, because an owner-unpublished listing is simultaneously `removed` (so: terminal, Inactive) and self-restorable (so: not terminal, normal placement). The observable consequence today is that **Republish sits one disclosure click inside a group labelled "Inactive"** — reachable, but filed under a heading that asserts the app cannot come back, which is exactly what Republish disproves.

  I did not change `partitionMyAppRows`: overriding a 🔴 invariant with its own test, in a PR about restoring two buttons, is the kind of drive-by that produces the next silent regression. Whoever resolves this should pick ONE meaning for `removed` and make both modules read it from one predicate — the one-rule-one-place fix — rather than adding a third local branch. The ledger test opens the collapse **explicitly** rather than hiding it in a helper, so the cost stays visible in the code. Mitigated meanwhile only in that the badge now reads `unpublished` rather than `removed`, so the row is at least identifiable inside that group.
- Analytics, install counters and the deploy-failure panel remain dropped from this page. Unchanged by this PR; still open from #4154.
- `MyAppsBody.browser.test.tsx`'s tRPC mock gained the two new procedures — the mutation hooks run unconditionally at mount, so an absent entry is a mount-time `TypeError`, not a missing assertion. What they are *called with* is asserted in the new ledger file.

### DIFF
--- a/src/components/Apps/MyAppsBody.authorActions.browser.test.tsx
+++ b/src/components/Apps/MyAppsBody.authorActions.browser.test.tsx
@@ -67,11 +67,23 @@ vi.mock('~/providers/FeatureFlagsProvider', () => ({
   useFeatureFlags: () => ({ appBlocks: true, appBlocksAuthor: true }),
 }));
 
-/** One mutation stub that RECORDS which procedure it belongs to — see `the wiring` below. */
+/**
+ * One mutation stub that RECORDS which procedure it belongs to — see `the wiring` below.
+ *
+ * 🔴 IT INVOKES THE CALLER'S `onSuccess`, and that is load-bearing rather than realism for its
+ * own sake. `OwnerUnpublishModal` closes itself and calls `onDone` from inside `onSuccess`, so
+ * a stub that only recorded the call would leave every post-success behaviour (the Inactive
+ * section opening, the row list invalidating, the modal closing) permanently unexercised — and
+ * a test asserting them would pass or fail on nothing at all.
+ */
+type MutationOpts = { onSuccess?: (data: unknown) => unknown; onError?: (e: unknown) => unknown };
 function mutationStub(name: string, isPending = false) {
   return {
-    useMutation: () => ({
-      mutate: (input: unknown) => mocks.calls.push([name, input]),
+    useMutation: (opts?: MutationOpts) => ({
+      mutate: (input: unknown) => {
+        mocks.calls.push([name, input]);
+        void opts?.onSuccess?.(undefined);
+      },
       isPending,
     }),
   };
@@ -125,9 +137,11 @@ const { MyAppsBody, rowActionsTestId } = await import('./MyAppsBody');
 /**
  * Fixtures are pairwise distinct on every dimension an assertion names — id, slug, name,
  * status, kind and last moderation action — and none of them can produce an expected value by
- * coincidence. In particular no two rows share a `lastModerationAction`, and `'delist'` is a
- * real moderator action rather than a placeholder, so a mutant that compares against the wrong
- * literal cannot survive on a fixture that could only ever have yielded it.
+ * coincidence. In particular no two rows share a `lastModerationAction`. The mod-removed
+ * fixture uses `'other'` because that is what the SERVER now sends — the projection normalises
+ * every non-owner verb to one value so a seated editor never receives the moderator's actual
+ * action. That a RAW verb (`delist`/`purge`/`claim`/…) still routes to the same state is pinned
+ * in the blocking `unit` tier, not here.
  */
 function row(over: Partial<MyAppRow> & { appListingId: string }): MyAppRow {
   const kind = over.kind ?? 'onsite';
@@ -158,7 +172,7 @@ const OWNER_HIDDEN = row({
 const MOD_REMOVED = row({
   appListingId: 'apl_modgone_r3',
   status: 'removed',
-  lastModerationAction: 'delist',
+  lastModerationAction: 'other',
 });
 const INACTIVE_DRAFT = row({ appListingId: 'apl_draft_k4', status: 'draft' });
 const SEAT_LIVE = row({ appListingId: 'apl_seat_m5', status: 'approved', role: 'editor' });
@@ -382,6 +396,36 @@ describe('the wiring — each control fires the right procedure with the right i
    * wording is present AND that the other kind's wording is not, so neither passes by a
    * variant that renders both strings.
    */
+  test('🔴 a successful unpublish OPENS the Inactive section the row is about to move into', async () => {
+    /**
+     * The row leaves the viewport at the exact moment the author acts: `partitionMyAppRows`
+     * files every `removed` listing under Inactive, which is collapsed by default, so without
+     * this the app AND its Republish button vanish while a success toast says it worked.
+     *
+     * Driven through the real confirm flow rather than by poking the prop, because the whole
+     * claim is "on SUCCESS" — `OwnerUnpublishModal` calls `onDone` from its mutation's
+     * `onSuccess`, and this test is what pins that wiring rather than the effect in isolation.
+     */
+    // A second, already-removed row so the Inactive section exists and has something in it.
+    mocks.rows = [LIVE_ONSITE, MOD_REMOVED];
+    renderWithProviders(<MyAppsBody />);
+
+    // Closed to begin with — the control reports its own state, so this is the negative arm.
+    const panelToggle = page.getByTestId('apps-mine-inactive-toggle');
+    await expect.element(panelToggle).toBeInTheDocument();
+    expect(panelToggle.element().getAttribute('aria-expanded')).toBe('false');
+
+    await userEvent.click(page.getByTestId(`apps-mine-unpublish-${LIVE_ONSITE.appListingId}`));
+    await userEvent.click(page.getByTestId('apps-mine-unpublish-confirm'));
+
+    // …and open afterwards. `aria-expanded` is read from the same boolean the panel renders
+    // from, so it cannot report a state the DOM does not have.
+    await expect.element(panelToggle).toHaveAttribute('aria-expanded', 'true');
+    expect(mocks.calls).toEqual([
+      ['unpublishOwnListing', { appListingId: 'apl_live_on', reason: undefined }],
+    ]);
+  });
+
   test('an ON-SITE unpublish warns that the app goes OFFLINE', async () => {
     mocks.rows = [LIVE_ONSITE];
     renderWithProviders(<MyAppsBody />);
--- a/src/components/Apps/MyAppsBody.authorActions.browser.test.tsx
+++ b/src/components/Apps/MyAppsBody.authorActions.browser.test.tsx
@@ -1,0 +1,400 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { page, userEvent } from 'vitest/browser';
+
+// `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
+import { renderWithProviders } from '../../../test/component-setup';
+import type * as MantineHooks from '@mantine/hooks';
+import type { MyAppRow } from '~/components/Apps/myAppsView';
+import type * as TrpcModule from '~/utils/trpc';
+import { capabilitiesForKind } from '~/shared/constants/app-capabilities.constants';
+import {
+  EDITOR_ACTIONS,
+  OWNER_ACTIONS_BY_STATE,
+  sortAuthorActions,
+} from '~/components/Apps/myAppsAuthorActions';
+
+/**
+ * THE AUTHOR-ACTION LEDGER for `/apps/mine`, and the restored Unpublish / Republish pair.
+ *
+ * 🔴 WHY THIS FILE EXISTS, AND WHY IT IS NOT JUST "TESTS FOR THE FIX". PR #4154 consolidated
+ * `/apps/my-submissions` into `/apps/mine`. `MySubmissionsList` — the only surface carrying
+ * the owner Unpublish/Republish controls — was orphaned by that merge, and the new page body
+ * contained zero occurrences of `unpublish`. The gap was DISCLOSED in the implementing PR and
+ * recorded as a noted omission rather than as a functional regression, and then THREE audit
+ * rounds passed over it. Each round asked "is the new page correct?"; none asked "is it
+ * COMPLETE?" — and no per-assertion suite can answer the second question, because a control
+ * that is simply absent has nothing to assert against. A tester found it instead.
+ *
+ * So the instrument here is a LEDGER, not another assertion:
+ *
+ *   `renderedActions()` enumerates EVERY interactive control inside a row's action container
+ *   and compares the resulting SET against `OWNER_ACTIONS_BY_STATE` / `EDITOR_ACTIONS`. It
+ *   fails when the set SHRINKS (a control was dropped — the #4154 shape) and when it GROWS
+ *   (a control was added without being declared). It additionally REFUSES a control that
+ *   carries no `data-author-action`, so "grew" cannot be evaded by omitting the attribute.
+ *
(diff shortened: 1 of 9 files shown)


ZacxDev:
## `preview / component-tests` is red — evidence says load flake, not this PR

Recording this rather than re-running it away, and **not** claiming the tier is green.

**The failure:** `MySubmissionsList.browser.test.tsx > … > the History button opens the moderation timeline`, `VitestBrowserElementError: Cannot find element with locator: getByText('Reported for policy')` … `Matcher did not succeed in time`, at **15289ms** — i.e. `expect.element`'s own matcher timeout, not an assertion that evaluated and disagreed.

**Why I do not think it is this branch:**

1. **The file and every dependency it exercises are untouched here.** `git diff --name-only <merge-base> HEAD` is 9 files; `MySubmissionsList.tsx`, `MySubmissionsList.browser.test.tsx`, `ownerListingModals.tsx` and `offsiteOwnerControls.ts` are none of them.
2. **Same scope, green locally.** The full `component` project on this branch runs **174 files / 1916 tests, 174 passed / 1916 passed** — the identical file and test counts CI executed (CI: 174 files, 1916 tests, 1 failed). So this is not a sharding or file-set difference.
3. **The cross-file mock-leak hypothesis was tested and refused.** This PR adds a per-file `vi.mock('~/utils/trpc', …)` that stubs `listMyListingModerationEvents → { items: [] }`, which under `isolate: false` could in principle leak into a later file in the same worker and empty exactly this modal. Forcing the three relevant files into one worker in order (`--maxWorkers=1`, my new file first) gives **3 files / 106 tests, all passing**. Hypothesis not supported.
4. **That CI run was heavily loaded.** 80 tests exceeded 5s, the slowest at **68.8s**, with `Duration 215s` against `tests 940s + import 715s`. Per the load-vs-assertion discriminator, load inflates *many* tests while a failed assertion inflates exactly one — 80 inflated tests, and the failure at 15.3s is not even close to the slowest. Three preview pipelines (4218, 4219, 4220) were building within ~2 minutes of each other on the same nodes.
5. **An unrelated PR hit the same tier the same day with the same signature.** Of the 26 retained `component-tests` TaskRuns, exactly 2 failed: this one, and `pr-preview-4173` — three tests in `Account/PlacementSpaceSection.freeSlots.browser.test.tsx`, timing out at **14977 / 14949 / 14982 ms**. Same ~15s matcher-timeout shape, entirely different file, different PR.

**Everything else is terminal and green** — 19 rollup rows, 18 SUCCESS + 1 SKIPPED (the fork-PR-only typecheck) apart from this one. The gating `pr-check-4218-lhlt9` PipelineRun is `True/Completed`, and within the preview pipeline `typecheck`, `unit-tests`, `packages-tests`, `apps-tests`, `auth-tests`, `build-image`, `deploy-preview`, `smoke-test` and `verify-preview` all succeeded.

**One thing worth flagging separately:** the `component-tests` TaskRun's Kubernetes condition reads `True / Succeeded` while its emitted result is `status = fail`. The step exits 0 because this tier is report-only, so the Tekton condition is the misleading signal and the GitHub status is the truthful one. Anyone diagnosing a component-tier failure from `kubectl get taskrun` alone will read it backwards.

I have not re-run the job. A flaky test is a real defect and re-running is not a fix — but this one is not in this PR's scope, so I am reporting it rather than quietly retriggering.

PR #4220, opened 2026-08-20
fix(apps): key the review WRITE gate on the resolved store scope, not a flag

ZacxDev:
## The bug

A tester reported: *"If I try leave a review, I see 'could not post review, apps are not enabled'"*.

`enforceAppListingsWriteFlag` was keyed on `isAppListingsEnabled`, i.e. `app-listings || app-blocks-enabled`. The store subsequently widened by a **third** route — the `app-listings-public-external` flag plus `resolveStoreVisibilityScope`, which lifts a non-privileged viewer to the `public-external` scope. The **read** path moved onto that resolved scope; the **write** gate stayed on the flag.

The external-only tester cohort holds *neither* flag the gate looked at. So its members reached an offsite listing, saw the review affordance the read scope had legitimately shown them, and got `UNAUTHORIZED: Apps are not enabled` on submit.

🔴 **A moderator cannot reproduce this.** A mod holds `app-listings`, so the old flag check passed and every surface looked healthy. That is why it stayed open.

## Every user of the shared gate (enumerated, not sampled)

`enforceAppListingsWriteFlag` is a shared middleware, so I enumerated its call sites before touching it rather than spot-fixing the reported one. Authoritative population: `git grep enforceAppListingsWriteFlag` across `src/`, `apps/`, `packages/`.

| Procedure | File | New gate correct here? |
|---|---|---|
| `upsertReview` | `src/server/routers/app-listings.router.ts` | **Yes.** The write the bug was reported against. Scope-keyed + kind-gated in the service; a `public-external` caller writing to an onsite listing gets the same `NOT_FOUND` the read path gives them. |
| `getMyReview` | `src/server/routers/app-listings.router.ts` | **Yes.** Same viewer, same listing, same scope — it is the write form's prefill, so gating it differently from `upsertReview` is what would be wrong. Soft-fails to `null` (it is a read) instead of throwing. |

That is the complete set — two, both named in the gate's own docstring, and no third site anywhere in the repo. Decided per-site rather than assumed: they differ in *failure posture* (throw vs `null`) precisely because one is a write and one is a read.

## The fix

**Key the write gate on `resolveStoreVisibilityScope`, kind-aware, mirroring the read path.**

- The gate resolves the same scope the read gate does and rejects only `none`.
- Each proc threads its scope into the service, which applies the shared kind rule.
- `public-external` → may review an **offsite** listing, **not** an onsite one.
- `full` → unchanged (a mod/tester resolves `full`, which admits every kind, so their behaviour is byte-identical to the flag check this replaces).
- `none` → still rejected.

The kind rule is a single shared export, `scopeAdmitsListingKind` in the dependency-free `src/shared/utils/store-visibility-scope.ts`. It is the predicate form of the data-layer filter the read path already applies, and it exists as one export so the write path and the UI cannot re-derive a disagreeing version. Its Prisma relation-filter twin was open-coded inline **once** at base (in `listAppListingReviews`) and is now a helper — so this is one extraction plus one genuinely new application in `getMyAppListingReview`, which had no relation filter at all. *(An earlier revision of this body said "twice"; that was inaccurate. No widening either way.)* The helper is an exhaustive switch mirroring `scopeAdmitsListingKind`, so `none` maps to a predicate that cannot match rather than to `{}` — no restriction — which is what the `=== 'public-external' ? … : {}` shape did for every other scope.

**The other direction is closed too.** `useCanReviewListing` now gates the affordance on the same resolved scope and the same shared rule, so the cohort is not shown a review control on an onsite listing the server would `NOT_FOUND`. Confined to `ReviewListingButton.tsx` + the one call site in `AppListingDetailBody.tsx`; **no changes to `MyAppsBody.tsx` or `appListingDetailRows.ts`.**

### The corrected docstring

The gate's reasoning is **corrected, not deleted** — the original is preserved as a labelled quotation because it records why the choice looked right:

> "This keeps the review WRITEs (`upsertReview`/`getMyReview`) on the SAME flag as the store visibility + reviews read path (`enforceAppListingsReadFlag`), so once `app-listings` widens independently of the held block-runtime gate, a viewer who can SEE the review affordance can also submit — instead of seeing the button and 403-ing on write."

The **goal** was right and is kept verbatim as the invariant. The **mechanism** was a bet that the store would widen by widening `app-listings`. It did not. The lesson, now recorded at the gate: *"same flag as the read path" was a proxy for "same admission rule as the read path"*, and the proxy held only while the read path was itself a flag check. Key the write on whatever the read path actually branches on — today, the scope.

The read gate's docstring carried the matching stale claim ("the public-external axis is READ-only", listing the write gate as deliberately excluded). That sentence *was* the defect, and it is corrected in place.

## Verification against the cohort, not a moderator

Confirmed read-only against prod (`cnpg-cluster-nvme0-5`): `camer047army744` = id `11072787`, `isModerator = f`. Approved-listing population: **4 offsite, 10 onsite** — so the kind distinction is live, not theoretical. Every fixture in the new suites is built around that real account.

## Red → green matrix

Base = `23d8035ae8` (this branch's merge-base), measured in a separate clean worktree.

| Suite | At base | At HEAD |
|---|---|---|
| `app-listings.router.review-write-scope.test.ts` | **RED** | green |
| `app-listing-review.store-scope.test.ts` | **RED** | green |
| `store-visibility-scope.kind.test.ts` | RED (new API absent) | green |
| `ReviewListingButton.storeScope.browser.test.tsx` | **RED — 2 failed / 8 passed** | green (10/10) |
| Combined unit run at base | **25 failed** | — |

The load-bearing red, verified to fail on **its own assertion** and for the reported reason:

```
FAIL  app-listings.router.review-write-scope.test.ts > external-only tester cohort
      (public-external) — the reported bug > upsertReview is ADMITTED
AssertionError: promise rejected "TRPCError: Apps are not enabled" instead of resolving
Caused by: TRPCError: Apps are not enabled
  ❯ src/server/routers/app-listings.router.ts:197:9
```

That is the tester's exact message, thrown at the old gate.

The router suite mocks `isAppListingsEnabled` with a **faithful** impl rather than a thrower, deliberately: a thrower would make the base run red on the *mock's* error instead of on the assertion, and a test that dies for the wrong reason is not evidence.

Full green at HEAD: **4549 unit tests / 198 files**, **982 component tests / 71 files**, typecheck **0 errors**, eslint clean.

## Mutation results — 10/10 killed

Each mutant was verified to have **actually landed** (exact-count match + post-write re-read; a mutant that fails to apply also reports SURVIVED). Every kill is an `AssertionError` — no `TypeError` kills, no kills for another guard's reason. Mutants target the narrowest expression that can be wrong, never a guard together with its enclosing condition.

| # | Mutation | Verdict | Killing assertion |
|---|---|---|---|
| M1 | kind rule: `public-external` returns `true` (admits onsite) | KILLED (4) | `CANNOT review an ONSITE listing` — *"promise resolved … instead of rejecting"* |
| M2 | kind rule inverted (`kind === 'onsite'`) | KILLED (7) | `public-external admits offsite → true` — *"expected false to be true"* |
| M3 | service kind gate deleted | KILLED (2) | `refuses an ONSITE listing the caller OWNS with NOT_FOUND` — *"expected TRPCError: You cannot review your own app to match { code: 'NOT_FOUND' }"* |
| M4 | service default `'full'` → `'none'` | KILLED (9) | `an omitted scope defaults to full` — *"expected TRPCError: Listing not found …"* |
| M5 | DB relation filter drops the kind restriction | KILLED (3) | `public-external ANDs the offsite relation filter` — *"expected { is: { status: 'approved' } } to deeply equal { is: { status: 'approved', …(1) } }"* |
| M6 | router stops threading `scope` into the service | KILLED (3) | `upsertReview threads the RESOLVED SCOPE` — *"expected { userId, input } to match { scope: 'public-external' }"* |
| M7 | gate rejects unless `full` (refuses the cohort) | KILLED (5) | `upsertReview is ADMITTED` — *"promise rejected 'Apps are not enabled' instead of resolving"* |
| M8 | gate stops rejecting `none` | KILLED (2) | `none-scope → UNAUTHORIZED, service NEVER consulted` — *"promise resolved … instead of rejecting"* |
| M9 | client resolver checks the external axis first (narrows a mod) | KILLED (component) | `a viewer holding BOTH … is NOT narrowed to offsite-only` — *"expected [] to have a length of 1 but got 0"* |
| M10 | client resolver fails **open** (`'none'` → `'full'`) | KILLED (component) | `a viewer holding NO store flag sees no CTA` — *"expected [ <button> ] to have a length of 0 but got 1"* |

M9 and M10 SURVIVED the unit tier and were killed in the component tier. That is **reachability, not a gap** — both mutate client-only code the unit tier does not load. Reported as a pair rather than as a bare "survived", since the two have opposite remedies. The sweep was scored by **counting** `✓`/`×` lines and reading failing test names, never by an exit code, and the working tree was verified byte-identical to its pre-sweep checksums afterwards.

## 🔴 Two instrument defects found — one of them a correction to my own earlier claim

Both produced confident greens that proved nothing. Neither is caused by this change; both are in test code for the component this PR touches. The first is stated as a **correction**: my original diagnosis of it was wrong and propagated to #4197 before I re-measured.

**1. 🔴 CORRECTED — the matcher is SOUND; the real defect is an unsettled-DOM RACE.**

An earlier revision of this PR claimed `expect.element(...).not.toBeInTheDocument()` was *inert for present elements*. **That claim was wrong**, and I have retracted it on #4197. The observation behind it was real — the `.not` form and `.toBeInTheDocument()` both passed on the same locator in the same test — but my inference from it was not.

Re-measured with an isolated probe (Chrome 149.0.7827.55, bridged as `-1200`):

| cell | result |
|---|---|
| element present, `.not.toBeInTheDocument()` applied **immediately after render** | **PASSED in 6 ms** (element confirmed present right afterwards) |
| element present, same assertion applied **after the render is known settled** | **FAILED in 14 864 ms** — correct |
| element present, `.elements().length` read **immediately after render** | **0** — a false zero on a present element |

So the matcher retries correctly; it just resolves on the **first** observation where the element is absent, and immediately after `render()` the DOM is always empty. Every "passes in ~5 ms" negative assertion is winning a race, not failing to assert.

🔴 **This changes the remedy, and that is why it matters beyond this PR.** Swapping the matcher for `.query()` / `.elements()` — the replacement #4197 recommends — does **not** fix it: those are point-in-time reads that never retry at all, so they are *more* exposed to the race, as the third row above shows directly. The fix is to **settle the render first**, whatever matcher you then use.

That is what the sentinel below does, and why the suite here goes genuinely red at base: not because `.elements()` is a better matcher, but because the awaited sentinel establishes the render before anything is counted.

**2. A wholesale `vi.mock('~/utils/trpc')` broke a whole suite as `Tests 0`.** Once `ReviewListingButton` imported `~/providers/FeatureFlagsProvider` (which imports `setTrpcBatchingEnabled`), the hand-written factory made the file fail to *import* — reported as an errored suite with **zero tests**, not as a failing assertion. Converted to the documented spread pattern; the repo's own `local-rules/no-wholesale-module-mock` flags exactly this and now passes. Same root cause fixed in the two `AppListingDetailBody` suites, which overrode `useFeatureFlags` but not `useOptionalFeatureFlags`, leaving the two hooks disagreeing.

## Notes

- `src/server/routers/__tests__/app-listings.router.review-write-gate.test.ts` is **deleted and superseded** by `app-listings.router.review-write-scope.test.ts`. Its entire premise ("the write gate is keyed on `isAppListingsEnabled`") is now false and its assertions contradicted the fix. All 9 of its cases have counterparts in the new suite (store-cohort non-mod write, mod write, both `getMyReview` admissions, both dark rejections, the anonymous rejection, and all three `listReviews` read postures) — the last two `listReviews` cases were added specifically for parity.
- `getMyAppListingReview` switches `findUnique` → `findFirst`: a relation filter cannot be expressed on a unique-key lookup. `(appListingId, userId)` is still the DB unique, so at most one row can match and the result is identical whenever the gate passes.
- Two new `StoreScopeEntrypoint` labels (`trpc-review-write`, `trpc-my-review`) keep the write's scope mix observable **separately** from the read's — a read/write scope disagreement on one surface is exactly this defect's shape, and one shared label would make it unobservable again.

## Recorded, deliberately not fixed

- **🟡 A narrow status oracle.** A `public-external` caller naming a non-approved **offsite** listing gets `BAD_REQUEST` where a nonexistent id gets `NOT_FOUND`, so the pair distinguishes "exists but unapproved" from "does not exist" — a distinction the read path deliberately collapses. Blast radius is **2 rows** (offsite drafts) and `AppListing` ids are ULIDs, so they are not enumerable. Out of scope for a fix aimed at the review-gate defect rather than folded in silently. The one-line fix is noted **in the code at the branch itself**, so the next person does not rediscover it: `if (scope !== 'full') throw throwNotFoundError(...)` ahead of the status check.
- **🟡 `getMyAppListingReview` gave up Prisma dataloader batching.** `findUnique` → `findFirst` (a relation filter cannot be expressed on a unique-key lookup) removes per-request coalescing on **every** review-form prefill, including `full` scope where the added filter is a no-op. One prefill per opened modal, so not a hot path; if it becomes one, branch — `findUnique` for `full`, `findFirst` only when the scope actually restricts. Recorded in the function's docstring.
- **🟡 No test exercises real Prisma.** The `{ is: {} }` / `{ is: { kind: 'offsite' } }` / `{ is: { kind: '__none__' } }` shapes are asserted only against `dbMock`, so they are pinned as **call shapes**, not as queries known to execute. A Prisma upgrade changing relation-filter syntax would pass these tests and fail in production. The `none` sentinel is the one part checked against reality: `app_listings.kind` is `text` (verified against the live column, not the Prisma type), so an out-of-domain literal yields an empty result rather than a query error — which matters, because a 500 is not "fail closed", it is a different failure.

## Unverified / caveats

- 🔴 **STILL OPEN: nobody has exercised this live as `camer047army744`.** The flag evaluation above confirms the cohort *resolves* `public-external`, and the suites confirm the code branches correctly on that scope — but neither is a real session driving the real UI. **Do not read the flag confirmation as this verification; they are different claims.** The only thing that closes it is a post-deploy check signed in as that account: review an **offsite** listing (should succeed), and confirm **no review control renders** on an onsite one.
- 🔴 **CORRECTED — this PR is NOT inert; it fixes the reported bug on merge.** An earlier revision of this section claimed `app-listings-public-external` did not yet exist in Flipt, so the change would be dormant. **That was wrong, and it conflated two flags.** The one deliberately absent is `apps-public-catalog-disabled` (the kill switch — absent by design, so `isFlipt` returning `false` leaves public access ON). Re-measured against the live Flipt (`flipt-v2`, environment `civitai-app`), evaluating each flag with the server's own `buildFliptContext` shape, **with a non-cohort control user in every arm** so a `true` cannot just mean "on for everyone":

  | Flag | `camer047army744` (11072787) | control (8514) |
  |---|---|---|
  | `app-listings-public-external` | **true** — `MATCH_EVALUATION_REASON` | false — `DEFAULT` |
  | `app-listings` | false — `DEFAULT` | false — `DEFAULT` |
  | `app-blocks-enabled` | false — `DEFAULT` | false — `DEFAULT` |

  `apps-public-catalog-disabled`: **absent** from the environment's 128 flags. `app-listings-public-external`: **present**, base `enabled: false`, resolving `true` for the tester by segment match only.

  That is the whole defect confirmed at the flag layer, end to end: `isAppListingsEnabled` = `app-listings || app-blocks-enabled` = **false** for this user, so the old gate threw `UNAUTHORIZED`; while `resolveStoreVisibilityScope` = **`public-external`** via axis 2, which is why the read path showed them the affordance. **The fix is live for this cohort the moment it merges — please do not deprioritise it as dormant.**

  How the wrong claim arose, since it is worth recording: a remembered fact ("that flag doesn't exist in Flipt") was true — of a *different* flag. Memory is a hypothesis; re-reading cost one request. Understating live impact is the dangerous direction, because it invites deprioritising a fix that is already load-bearing.
- The component tier could not run on this host until I bridged a Playwright browser-version mismatch (`chromium_headless_shell-1200` expected, `-1228` in the Nix store) via a local symlink dir. **Pre-existing and host-local** — an untouched sibling suite failed identically before any of my changes, and the bridge is a local run-time env var only, nothing committed. CI is unaffected, but it means the component tier is silently unrunnable for anyone on this flake pin.


ZacxDev:
### Follow-up commit `12ff20d3d4` — two CI gates my local run missed

Both were repo-convention gates, both real, both mine:

1. **Prettier** on the two new test files.
2. **`no-direct-shared-module-mock`** — `app-listing-review.store-scope.test.ts` hand-rolled a `~/server/db/client` mock instead of the canonical `~/__tests__/mocks/db.mock`. Converted with the repo's own `scripts/test-perf/codemod-shared-mocks.mjs --write`.

**Why I missed them locally:** I scoped my local unit run to the three directories I changed, and `no-direct-shared-module-mock.test.ts` lives in `src/server/services/__tests__/` — a repo-wide meta-test outside them. A changed-dirs-only run is structurally blind to gates that live elsewhere.

**Re-verified after the mock swap** rather than assuming a test-infrastructure change is inert — it can silently weaken assertions. Re-ran the three mutants this file is responsible for killing:

| Mutant | Verdict |
|---|---|
| M1 — kind rule admits onsite | KILLED (2 failing) |
| M3 — service kind gate deleted | KILLED (2 failing) |
| M5 — DB filter drops the kind restriction | KILLED (3 failing) |

Working tree confirmed byte-identical to its pre-mutation checksums afterwards. Prettier re-checked with an explicit file list — **2 files checked, 0 unformatted**, reported as a pair so the clean verdict is not just "nothing matched".

CI is now fully settled: **19 rows, all terminal — 18 SUCCESS + 1 SKIPPED, zero failures.**

ZacxDev:
### Two corrections/links

**1. 🔴 The "inert until the flag lands" caveat was WRONG and is now corrected in the body.**

I had claimed `app-listings-public-external` did not yet exist in Flipt, so this PR would be dormant. That conflated it with `apps-public-catalog-disabled` — *that* is the flag deliberately absent by design.

Re-measured against live Flipt (`flipt-v2`, env `civitai-app`), with a non-cohort control in every arm:

| Flag | `camer047army744` (11072787) | control (8514) |
|---|---|---|
| `app-listings-public-external` | **true** — `MATCH_EVALUATION_REASON` | false — `DEFAULT` |
| `app-listings` | false | false |
| `app-blocks-enabled` | false | false |

`apps-public-catalog-disabled` is absent from the environment's 128 flags.

So the mechanism is now confirmed end-to-end at the flag layer: `isAppListingsEnabled` (= `app-listings || app-blocks-enabled`) is **false** for this user — hence the old gate's `UNAUTHORIZED` — while `resolveStoreVisibilityScope` returns **`public-external`**, which is why the read path showed them the button. **This PR fixes the reported bug on merge; it is not dormant.**

**2. The inert-assertion finding is now on #4197** — https://github.com/civitai/civitai/issues/4197#issuecomment-5363480538

Includes a reproduction needing no injected element (both `.not.toBeInTheDocument()` and `.toBeInTheDocument()` passing on the same locator in the same test, with the ~5 ms vs ~50 ms timing tell), and the **sentinel positive-control** pattern for the case that issue's migration advice does not cover: when the component renders *nothing* in the negative case there are no positive assertions to sit behind, so an unpaired `.query()` is inert again.

**Still open, and the flag confirmation above is NOT it:** nobody has driven this live as `camer047army744`. Post-deploy: review an offsite listing (should succeed), confirm no review control on an onsite one.

ZacxDev:
### Audit follow-ups — commit `8ff4a01419`

**🔴 The matcher claim was WRONG and is retracted.** The audit could not reproduce it because it was not reproducible: `expect.element(...).not.toBeInTheDocument()` is **sound**. Isolated probe (Chrome 149.0.7827.55, bridged `-1200`):

| cell | result |
|---|---|
| present element, `.not` **immediately after render** | PASSED in **6 ms** (element confirmed present afterwards) |
| present element, `.not` **after the render settled** | **FAILED in 14 864 ms** — correct |
| present element, `.elements().length` **immediately after render** | **0** — false zero |

The real defect is an **unsettled-DOM race**: `.not` resolves on the first observation where the element is absent, and right after `render()` the DOM is always empty. My "both directions pass" observation was real; the *inference* was wrong.

Your read was right that the audit tested a different cell — but so did I, and mine was the one that misled. This also reconciles the sibling auditor's injected-`<Badge>` result: 7 ms is the race signature, not inertness.

🔴 **This changes #4197's remedy**, so I posted a full retraction there rather than quietly editing: `.query()`/`.elements()` never retry at all, so they are *more* exposed to the race — row 3 is that failure happening. Converting 69 sites to a non-awaited read would preserve every false pass. The fix is settle-then-assert; the site count needing work is likely well below 69, since anything following an awaited assertion is already correct. I also flagged that the proposed lint rule would push people onto the worse idiom.

**🟡-3 fail-closed — taken.** Three `?? 'full'` defaults → `narrowStoreScope`. `listingKindFilterForScope` is now an exhaustive switch: it previously returned `{}` (no restriction) for every scope that was not literally `public-external`, `none` included. `none` now maps to a non-matching predicate — and `app_listings.kind` is `text` (checked against the live column), so that is an empty result, not a query error.

Two consequences handled deliberately rather than worked around: the pre-existing service tests relied on the fail-open default and now pass `scope: 'full'` explicitly (they were always full-scope scenarios), and my own "an omitted scope defaults to `full`" test asserted the behaviour this removes, so it is **deleted**, not adjusted.

**🟡-2 seam — taken, and your measurement reproduced exactly.** Deleting `listingKind: detail.kind` left the component tier **982/982 green**. Root cause of the blindness: `mocks.features` was a fixed literal, so neither detail suite could construct a `public-external` viewer at all. It is now per-test mutable, with three cases pinning the join (public-external+onsite hidden, +offsite shown, full+onsite unchanged). The deletion mutant now dies on its own assertion — `expected <button …> to have a length of 0 but got 1`.

**🟢 Shipped strings — fixed.** Both Prometheus `help` texts (the part Grafana actually shows) and the "flag-gated" comment. Also corrected the body: the filter was open-coded **once**, not twice.

**Recorded, not fixed** — all three now in the body *and* at the relevant line in code: the offsite status oracle (2 rows, ULID ids, one-line fix noted at the branch), the `findFirst` dataloader-batching loss, and that no test exercises real Prisma.

Verification: typecheck 0 errors; **14 618 unit tests** / 926 files; **985 component tests** / 71 files; eslint 0 errors, prettier clean.

**Still open, unchanged:** nobody has driven this live as `camer047army744`.

PR #4221, opened 2026-08-20
fix(scan): only tombstone a file when a 404 proves it is gone

ZacxDev:
## The bug

A **transient** failure to resolve a download URL is recorded as a **permanent** "this file is gone" flag — and that flag also permanently exempts the file from virus and pickle scanning, while the file stays `Published`, `Public`, and downloadable.

`createModelFileScanRequest` pre-flights a scan submission by calling `resolveDownloadUrl` twice, 60s apart. The second `catch` was untyped, so **every** failure mode — resolver 5xx, timeout, rate limit, auth failure, DNS, a transport reject — produced `code: 'not-found'`. `scanFilesFallbackJob` converts that code into `ModelFile.exists = false`, and its own query then filters those rows out:

```ts
{ OR: [{ exists: null }, { exists: true }] },
```

So one bad minute at the resolver takes a healthy file out of the scan queue permanently. Nothing ever revisits it.

## Evidence it was misfiring

Measured against the real download path (endpoint redirect → range-GET the signed URL it returns):

- **41 of 41** readable files carrying the `exists = false` tombstone were still **fully downloadable** (`307` → `206`), across **51 distinct uploaders**. Zero were actually missing.
- The tombstones arrive in **bursts**, which independent file losses would not: the top 5 hours hold **52%** of them, with 85 distinct model versions tombstoned inside a single hour. For comparison, successful scans over the same period put only **1.4%** of their volume in their top 5 hours — and the scanner demonstrably sustains ~1,100 files/hour, so this isn't a busy-period artifact.

## The fix

1. `getDownloadUrlByFileId` threw a bare `Error`, discarding `response.status`. It now throws a `StorageResolverError` carrying the status — mirroring `DeliveryWorkerError`, which has carried it all along. That class's own docstring already described this exact hazard ("masking a real storage outage as *not found*"); the mechanism existed and simply wasn't used here.
2. New `isDefiniteNotFound(err)` — deliberately narrow: **only a 404** counts as proof of absence. The asymmetry is the point. Acting wrongly on `true` leaves a public file unscanned forever; acting wrongly on `false` costs one retry on the next 5-minute tick. `400` (malformed key) is excluded too — genuinely-unresolvable keys are separately identifiable from the stored `url`, and 400 is also what a transiently-misbehaving upstream returns.
3. `createModelFileScanRequest` classifies on the **retry's** failure — the most recent evidence, and the first failure may well be the transient one the retry exists to absorb. Anything that isn't a proven 404 is now `'transient'`, which already had correct handling (reset `scanRequestedAt`, retry next tick).

No change to the caller's branching: `not-found` still tombstones, `transient` still retries. Only the classification moved.

## Tests

`60 passed` across the three affected suites.

The regression guard was **watched to fail** on the pre-fix behaviour. Reverting only the classification expression (leaving all other changes in place, so the mutation is isolated to the line under test) turns **8 tests red** with `expected 'not-found' to be 'transient'` — the exact defect — while both legitimate `not-found` cases stay green, so the tests are not merely asserting "always transient". Every pre-existing test stays green under that mutation.

Covered: resolver 500/503/401, delivery-worker 429/503, a status-less transport error, a non-`Error` rejection, a plain object that merely *looks* like a 404, and both orderings of a 404 paired with a 5xx.

One pre-existing test — `'throws code=not-found when both pre-flight attempts fail'` — asserted the old behaviour using a bare `Error`. It encoded the bug, so it is now split into the 404 case and the transient case.

The `~/utils/delivery-worker` mock in `createModelFileScanRequest.test.ts` was a wholesale one-key replacement; it now spreads the real module and overrides only `resolveDownloadUrl`. Otherwise the classification under test would have been a property of the fake, and the mock would have gone stale the moment the module gained an export — which is precisely what adding `isDefiniteNotFound` did.

## Follow-up, not in this PR

Existing tombstones were written by the broken predicate and are untrustworthy. Clearing `exists = false` on the affected rows returns those files to the scan queue — but only **after** this lands, or the next transient blip re-tombstones them. That's a production data change and belongs in its own change.

Separately: whether a `Published` + `Public` file whose scan has never completed should be downloadable at all is a product question this PR does not touch.

### DIFF
--- a/src/server/services/orchestrator/__tests__/createModelFileScanRequest.test.ts
+++ b/src/server/services/orchestrator/__tests__/createModelFileScanRequest.test.ts
@@ -554,6 +554,36 @@ describe('createModelFileScanRequest', () => {
       );
     });
 
+    // The verdict is deliberately taken from the RETRY, and so are the audit
+    // fields. Every other fixture here rejects with the same object twice, which
+    // cannot tell the two apart — so re-sourcing the log from `firstError` passes
+    // a fully green suite while making the log describe the wrong attempt. Give
+    // the two attempts pairwise-distinct values so that mutant is observable.
+    it('sources the audit fields from the RETRY, not the first attempt', async () => {
+      const firstFailure = new DeliveryWorkerError(410, 'Gone');
+      firstFailure.resolverError = new StorageResolverError(503, 'Service Unavailable');
+      const retryFailure = new DeliveryWorkerError(404, 'Not Found');
+      retryFailure.resolverError = new TypeError('fetch failed');
+
+      mockResolveDownloadUrl
+        .mockRejectedValueOnce(firstFailure)
+        .mockRejectedValueOnce(retryFailure);
+
+      const promise = createModelFileScanRequest(baseInput).catch((e) => e);
+      await vi.advanceTimersByTimeAsync(60_000);
+      await promise;
+
+      expect(mockLogToAxiom).toHaveBeenCalledWith(
+        expect.objectContaining({
+          // 404 from the retry, NOT 410 from the first attempt.
+          deliveryWorkerStatus: 404,
+          // The retry's resolver failure had no status; the first attempt's did.
+          resolverStatus: undefined,
+          resolverError: 'TypeError: fetch failed',
+        })
+      );
+    });
+
     it('skips pre-flight entirely when preflight=false (inline upload path)', async () => {
       mockResolveDownloadUrl.mockRejectedValue(new Error('would fail if called'));
       mockSubmitWorkflow.mockResolvedValue({
--- a/src/server/services/orchestrator/__tests__/createModelFileScanRequest.test.ts
+++ b/src/server/services/orchestrator/__tests__/createModelFileScanRequest.test.ts
@@ -559,10 +559,15 @@ describe('createModelFileScanRequest', () => {
     // cannot tell the two apart — so re-sourcing the log from `firstError` passes
     // a fully green suite while making the log describe the wrong attempt. Give
     // the two attempts pairwise-distinct values so that mutant is observable.
+    // 🔴 The retry carries 410, NOT 404, on purpose. 404 is the only value
+    // `deliveryWorkerStatus` takes anywhere else in this suite, so asserting 404
+    // here could not distinguish "reads the retry" from "hardcoded to 404" — a
+    // fixture that can only ever produce the constant it names cannot detect a
+    // mutant that returns that constant. 410 is a value no other fixture emits.
     it('sources the audit fields from the RETRY, not the first attempt', async () => {
-      const firstFailure = new DeliveryWorkerError(410, 'Gone');
+      const firstFailure = new DeliveryWorkerError(404, 'Not Found');
       firstFailure.resolverError = new StorageResolverError(503, 'Service Unavailable');
-      const retryFailure = new DeliveryWorkerError(404, 'Not Found');
+      const retryFailure = new DeliveryWorkerError(410, 'Gone');
       retryFailure.resolverError = new TypeError('fetch failed');
 
       mockResolveDownloadUrl
@@ -575,11 +580,16 @@ describe('createModelFileScanRequest', () => {
 
       expect(mockLogToAxiom).toHaveBeenCalledWith(
         expect.objectContaining({
-          // 404 from the retry, NOT 410 from the first attempt.
-          deliveryWorkerStatus: 404,
+          // 410 from the retry, NOT 404 from the first attempt.
+          deliveryWorkerStatus: 410,
           // The retry's resolver failure had no status; the first attempt's did.
           resolverStatus: undefined,
           resolverError: 'TypeError: fetch failed',
+          // These two are what an on-call reader uses to reconstruct which
+          // attempt said what, and nothing else in the suite asserts them —
+          // swapping or dropping them survives a green run without this.
+          firstError: 'Delivery worker error: Not Found',
+          retryError: 'Delivery worker error: Gone',
         })
       );
     });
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Correction to the PR description, and a second fix round

An adversarial audit found that **the first commit did not close the hole it was written to close**, and that the description above is wrong on one point. Correcting publicly rather than quietly editing the body, since it may already have been read.

### The description says "A caller (`scanFilesFallbackJob`)". There are **three**.

`ModelFile.exists = false` is also written by `src/server/services/model-file-scan.service.ts:715` and `src/pages/api/mod/clean-up.ts:83`. Both are affected by this change and neither is updated here. Their comments say the tombstone exists so a rescan "doesn't keep churning on a permanently dead AIR" — which, after this PR, is what they will do for non-404 failures. That is the intended trade (churn is recoverable, an unscannable public file is not), but it was undisclosed.

### The real defect: the bug lived one layer below the classifier

`resolveDownloadUrl` swallowed the storage-resolver error with a bare `catch {}` and fell back to the delivery worker. Verified by **executing** the real module with the resolver enabled — the production config:

| resolver | delivery worker | escapes to classifier | `isDefiniteNotFound` |
|---|---|---|---|
| **503 (outage)** | **404** | `DeliveryWorkerError(404)` | **true → TOMBSTONE** 🔴 |
| 404 (absence proven) | 500 | `DeliveryWorkerError(500)` | false — proof discarded |
| 503 | 503 | `DeliveryWorkerError(503)` | false ✓ |

Row 1 is this PR's own bug surviving it. The delivery worker is the legacy path keyed off `ModelFile.url` and cannot see a file registered only in `file_locations`, so during a resolver disruption it returns 404 for healthy files — **in a burst**, which is precisely the signature cited above as evidence of the bug.

Fixed in `7c0db1cdf0`: the fallback still happens on a resolver 404 (unsynced `File` records depend on it), but the resolver's error is now attached to the thrown `DeliveryWorkerError`, and `isDefiniteNotFound` requires the resolver to have proven absence whenever it was consulted. Backward compatible — same class, same `statusCode`, same message — so the download endpoint and other callers are untouched. `410 Gone` now counts as absence alongside 404, matching that endpoint, which has always treated `404 || 410` that way.

### Two mutants had survived, and that was a real gap

Nothing drove `getDownloadUrlByFileId` to a non-OK response — every test built the error by hand. So reverting `StorageResolverError` to a bare `Error`, and hardcoding its status to 404, **both passed the entire suite**. The second is the dangerous one: a resolver reporting every error as 404 would tombstone on every blip, silently green.

`delivery-worker.resolver-enabled.test.ts` now exercises that path with the resolver enabled. Both mutants die (10 tests each), and the new bare-catch mutant dies too (5 tests), failing precisely on the four outage-plus-404 cases.

**93 passed** across 5 suites, up from 60 across 3. Typecheck 43, identical to `origin/main`, none in the touched files.

### Known and deliberately not in this PR

The retry queue is unbounded: every non-404 failure resets `scanRequestedAt` with no backoff, attempt cap, or `orderBy`, and each retry burns the 60s pre-flight sleep. Previously the (wrong) tombstone drained that queue; this PR removes the drain without adding a bounded one. **That should be settled before any backfill clears existing tombstones**, since the backfill feeds exactly this queue.

ZacxDev:
## Adversarial audit loop complete — ready for review

Five independent audit rounds, each auditing only the delta since the last. Every round found something the previous round's *fix* had introduced or left open, which is why I kept going rather than stopping when it felt done.

| round | severity found | what it caught |
|---|---|---|
| 1 | 🔴 | the fix didn't close its own hole — `resolveDownloadUrl` still discarded the resolver's verdict, so a resolver outage + delivery-worker 404 still tombstoned healthy files |
| 2 | 🟡 | the transport-reject case (the commonest outage shape) was unpinned; a plausible "tighten the types" refactor silently restored the bug |
| 3 | 🟡 | the audit fields added for auditability had zero coverage — swapping `resolverStatus`/`deliveryWorkerStatus` passed green; the added test file was failing a **blocking** prettier gate |
| 4 | 🟢 | a claimed fix had silently not applied; the audit fields couldn't distinguish first attempt from retry |
| 5 | 🟢 | the retry fixture used the only value that field ever takes, so a hardcode mutant survived |

**Final state:** 102 tests across 5 suites (from 60 across 3). Typecheck 43 — identical to `origin/main`, none in the five touched files. Prettier and ESLint clean on both the added-file and modified-file gate sets.

Every guard is mutation-verified, and each mutant was watched to fail on the test that names it, for that test's own reason — not merely "a test went red". The invariants that matter: the attach guard is `!== undefined` and **not** an `instanceof` narrowing (narrowing it kills 4 tests); `410` is in `ABSENCE_STATUSES` (dropping it kills 4); the classifier reads the **retry** (re-sourcing from the first attempt kills 1 each for status and resolver-error).

### Two things a reviewer should weigh, neither about the classifier

**1. The retry queue is now unbounded, and that gates the follow-up.** Every non-404 failure resets `scanRequestedAt` with no backoff, attempt cap, or `orderBy`, and each attempt burns the 60s pre-flight sleep. The (wrong) tombstone was previously draining that queue; this PR removes the drain without adding a bounded one. That is the correct trade — churn is recoverable, an unscannable public file is not — but **clearing existing tombstones feeds exactly this queue**, so bounding the retry path should come first.

**2. `exists=false` has three writers.** Besides `scanFilesFallbackJob`, it is written by `model-file-scan.service.ts:715` and `pages/api/mod/clean-up.ts:83`. Both inherit the new behaviour and neither is updated here; `clean-up.ts` runs at concurrency 20 inside an already-returned API handler, so each failing file adds a detached 60s sleep there.

### Not claimed

I have **not** reproduced a tombstone by forcing a resolver 5xx end-to-end — that needs a non-prod environment, and it is the one experiment that would take the root cause from "code read plus two measurements" to demonstrated. The branch is also well behind `main`; a green branch says nothing about the tree its merge creates, so please gate on the merged tree.

The audit loop stopped by judgement, not because a round came back empty: round 5's residue was uncovered defensive branches and thrown-message strings on audit-only fields, where `code` is the load-bearing value and is well asserted.

PR #4225, opened 2026-08-21
fix(admin): ?confirm=false no longer satisfies a destructive-action guard

JustMaier:
Closes ClickUp [868kun1w5](https://app.clickup.com/t/868kun1w5). Follow-up to #4209, which fixed the same defect on the public models endpoint.

## The bug

Each of these endpoints parses its schema against `req.query` (two against `{...req.query, ...req.body}`), where every value is a string. `z.coerce.boolean()` is JS `Boolean()` there, and `'false'` is a non-empty string — so `=false` parsed as `true` and the endpoint did the opposite of what the operator asked.

## Two of these are filters. Two are guards, and that is the part worth understanding

If you are here because you are about to add a `z.coerce.boolean()` somewhere, this is the distinction that matters — not the four changed lines.

A **filter** reading `false` as `true` returns a wrong set. Bad, recoverable, visible in the response.

A **guard** reading `false` as `true` does something irreversible that the operator explicitly tried to decline. `confirm` here is not a flag; it gates a destructive reset in a `superRefine`:

```ts
if (data.action === 'reset' && !data.confirm) {
  ctx.addIssue({ code: 'custom', message: 'reset requires confirm=true' });
}
```

Under coercion, `?confirm=false` **satisfies** that guard. An operator writing `confirm=false` to be explicit that they did *not* want the reset got the reset — deleting every `MembershipGift` row where the user is gifter or recipient, or wiping all referral data for a user. There is no error and nothing to undo. **A guard that cannot be declined by writing `false` is not a weaker version of the filter bug; it is a different one.**

| Endpoint | Param | Kind | Effect of `=false` before this PR |
|---|---|---|---|
| `admin/deliver-prepaid-buzz.ts` | `force` | filter | unlocked **every** locked prepaid token per user instead of one, across the whole batch — real Buzz, in bulk |
| `admin/permission.ts` | `revoke` | filter | took the **remove** branch and invalidated each affected user's session |
| `testing/gift-membership.ts` | `confirm` | **guard** | reset proceeded; all `MembershipGift` rows for that user deleted |
| `testing/referrals.ts` | `confirm` | **guard** | reset proceeded; all referral data for that user wiped |

All four are `WebhookEndpoint`-gated, so this is an operator footgun rather than a public hole.

`anonymous` (gift-membership) and `settleImmediately` (referrals) move with them. They are the same defect in the same schemas, and leaving a known-broken neighbour in a file you are already fixing is precisely how this bug reached four endpoints.

## Scope, and what is deliberately left

Two more carry the same shape and are filed separately rather than swept in here:

- `testing/backfill-stale-nsfw-rollups.ts:51` — `dryRun`. Also wrong, but it fails **safe**: `?dryRun=false` becomes `true`, so the run does less, not more.
- `mod/daily-challenge/backfill-theme-elements.ts:18` — `force`, mod-gated.

> ⚠️ **The sweep that found these was run with the Grep tool, not bash `grep`.** Bash `grep` in this environment is rewritten to `rtk grep`, and it returned **empty with exit 0 on four files that do contain the pattern** — a clean negative that reads as a thorough search. Re-running through the Grep tool found every hit. If you reproduce this search with bash and get fewer results, that is the reason; it is not an over-report.

## Tests

Derived from each schema rather than hand-listed, so a boolean added later on `z.coerce.boolean()` fails here too — a listed test has the same blind spot as the bug. Discovery stays inside the `it` bodies: at module scope, a `z.preprocess` field that *throws* rather than returning an issue fails **collection**, and the file reports as a pass with zero tests.

**The guards are asserted through the whole schema, not the field.** A field parsing `'false'` correctly is not the same claim as the guard refusing — so the test sends `{action: 'reset', userId, confirm: 'false'}` at the real schema and requires it to be rejected, with a positive control on `confirm: 'true'` so a schema that refused every reset could not pass.

**Every one of the four fixes was checked by reverting it**, not by reading:

| Reverted | Suite says |
|---|---|
| `force` | `?force=false parsed as true` |
| `revoke` | `?revoke=false parsed as true` |
| `confirm` (gift-membership) | `?confirm=false parsed as true`, **plus** `?confirm=false SATISFIED the reset guard — the destructive action would run` |
| `confirm` (referrals) | same pair |

The guard assertion firing as a *third* failure on the `confirm` reverts is the point: it is the one that describes what actually happens to the data.

## Verification

- `pnpm exec eslint <changed files>` — no issues
- `pnpm exec prettier --write <changed files>`
- 16 tests in the new file; full `test:unit:run` result to follow in a comment

### DIFF
--- a/src/pages/api/admin/deliver-prepaid-buzz.ts
+++ b/src/pages/api/admin/deliver-prepaid-buzz.ts
@@ -1,11 +1,14 @@
 import * as z from 'zod';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
 import { unlockTokensForUser } from '~/server/services/subscriptions.service';
-import { commaDelimitedNumberArray } from '~/utils/zod-helpers';
+import { booleanString, commaDelimitedNumberArray } from '~/utils/zod-helpers';
 
-const schema = z.object({
+export const schema = z.object({
   userIds: commaDelimitedNumberArray(),
-  force: z.coerce.boolean().optional(),
+  // booleanString, not z.coerce.boolean: this schema is parsed against req.query, where
+  // coerce runs JS Boolean() and makes `&force=false` TRUE — unlocking every locked token
+  // for each user instead of one, across the whole batch.
+  force: booleanString().optional(),
 });
 
 type UnlockResult = {
--- a/src/pages/api/admin/deliver-prepaid-buzz.ts
+++ b/src/pages/api/admin/deliver-prepaid-buzz.ts
@@ -33,7 +33,14 @@ type UnlockResult = {
  */
 export default WebhookEndpoint(async (req, res) => {
   try {
-    const { userIds: targetUserIds, force } = schema.parse(req.query);
+    // safeParse + 400, matching permission.ts: `schema.parse` throws into the catch below,
+    // which reports a rejected query param as a 500 server fault. `?force` as a bare flag
+    // (value '') is a rejected value now, so that path is reachable.
+    const parsed = schema.safeParse(req.query);
+    if (!parsed.success) {
+      return res.status(400).json({ error: 'Invalid request', issues: parsed.error.issues });
+    }
+    const { userIds: targetUserIds, force } = parsed.data;
 
     if (targetUserIds.length === 0) {
       return res.status(400).json({ error: 'userIds is required' });
--- a/src/pages/api/admin/permission.ts
+++ b/src/pages/api/admin/permission.ts
@@ -6,13 +6,16 @@ import { featureFlagKeys } from '~/server/services/feature-flags.service';
 import { addSystemPermission, removeSystemPermission } from '~/server/services/system-cache';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
 import { refreshSession } from '~/server/auth/session-invalidation';
-import { commaDelimitedStringArray } from '~/utils/zod-helpers';
+import { booleanString, commaDelimitedStringArray } from '~/utils/zod-helpers';
 
-const schema = z.object({
+export const schema = z.object({
   // Accepts a single key or a comma-delimited list (e.g. key=a,b,c).
   key: commaDelimitedStringArray(),
   usernames: commaDelimitedStringArray(),
-  revoke: z.coerce.boolean().optional(),
+  // booleanString, not z.coerce.boolean: this schema is parsed against req.query, where
+  // coerce runs JS Boolean() and makes `?revoke=false` TRUE — taking the REMOVE branch and
+  // invalidating each affected user's session.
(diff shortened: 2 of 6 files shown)


JustMaier:
## Verification on `f7efcfe49a`

Full `pnpm run test:unit:run` on a **clean tree**: **19,837 passed, 27 skipped, 0 failed** across 1,258 files (exit 0, 275s). `blocks.router.workflow.test.ts` collected **350 tests**, so the submodule is initialised and this is a real run.

Worth stating why that says "clean tree": an earlier run on this branch was launched and then had files committed underneath it mid-flight. It went green and had even collected the final version of the changed file, so it was probably honest — but a run whose inputs changed while it was running is not a verification, so it was discarded and re-run rather than reported.

## What changed since review

**F1 — the precedence pair was a check that could not fail, and is replaced.** It asserted `{ ...a, ...b }` over object literals: a JS language guarantee, with the handler never imported. Reordering the handler's real spread — the exact thing the comment claimed to guard — left it green while a body `confirm: false` sent to decline a reset would be overridden by a stale `?confirm=true`.

The new test drives the handler (`src/tests/api/admin/referrals-guard-precedence.test.ts`): `WebhookEndpoint` stubbed to the identity, and the reset branch's writes mocked so the **control reaches the destructive path for real** rather than passing because everything is refused. Verified by the same mutation that killed the old one — reordering the spread reddens both assertions:

```
AssertionError: the query string overrode a body `false` — a declined reset would have run: expected 200 to be 400
AssertionError: the guard refused an affirmative body value: expected 400 to be 200
```

**F2 — the accept set is now what `stringbool` actually takes**: `on` (what an HTML checkbox submits), `y`, `enabled`, case-insensitive. The declines table listed `off` two lines above an untested `on`, which read as a closed enumeration of a destructive guard's surface. A padded ` true ` is pinned as **declined** — a real behaviour change from `z.coerce.boolean()`, which accepted it, and nothing recorded that.

**`deliver-prepaid-buzz` now returns 400, not 500**, on a rejected param, matching `permission.ts`. `schema.parse` threw into the outer catch, so a bad `force` was reported as a server fault — and `?force` as a bare flag (value `''`) is a rejected value now, so that path is reachable.

Also corrected a comment that oversold its filter: `booleanKeys` selects fields that accept a real `true`, which pulls in `z.coerce.number()` fields since `Number(true) === 1`. It is a net, not an enumeration of the booleans.

## Known limit, for whoever picks up [868kuqjy7](https://app.clickup.com/t/868kuqjy7)

`SCHEMAS` in the derived test is a **hand-written list of four endpoints**, and a hand-enumerated test cannot catch a hand-enumeration miss. The durable form is a repo-scanning guard over `src/pages/api/**`, beside the other `no-*.test.ts` files. That belongs with the two remaining instances rather than here.

PR #2547, opened 2026-06-14
test(component): Vitest browser-mode component-testing scaffold (runs in Tekton, removes GH Actions pr-check)

ZacxDev:
## What

Adds a **component-testing layer** (Vitest **browser mode**, real Chromium via Playwright) as a second Vitest *project* alongside the existing unit suite, with the first test on a high-churn generation-form leaf. Tests run in **Tekton** (companion talos-infra PR adds the `component-tests` task); this PR also **removes `.github/workflows/pr-check.yml`** so PR validation no longer uses GitHub Actions runners.

## Why this component

- Zero component-test coverage today. Highest-churn UI = the `generation_v2` form family, which is **e2e-impossible** (external GPU/Buzz) and **unit-untested**.
- `SeedInput` is a pure-props leaf (`value`/`onChange`, no tRPC/router) with **real, lockable behavior** — random/custom toggle, `MAX_SEED` clamp, paste sanitization, `undefined`=random. It anchors the scaffold while exercising Mantine, the browser-mode fidelity reason we rejected jsdom.

## Changes

- **`vitest.config.mts`** → `projects`: `unit` (current config, **unchanged**, node env) + `component` (browser/Chromium via `@vitest/browser-playwright` factory, `*.browser.test.tsx` glob). Unit scripts scoped to `--project unit` so they never boot a browser.
- **`test/component-setup.tsx`** — `renderWithProviders` (MantineProvider + fresh `QueryClient{retry:false}` + `next/router` mock) + auto-`cleanup()`.
- **`test/browser-process-shim.ts`** — first setupFile; shims `process.env` that civitai client modules read at import time.
- **`src/components/generation_v2/inputs/SeedInput.browser.test.tsx`** — first test, 6 cases.
- **`package.json`** — `test:component` / `test:component:watch`; unit scripts scoped `--project unit`.
- **`tsconfig.json`** — include `test/` so the setup files are typechecked.
- **`.gitignore`** — browser-mode failure `__screenshots__/`.
- **deps** — `@vitest/browser` + `@vitest/browser-playwright` (pinned `4.0.18` to match vitest core), `vitest-browser-react`.
- **Removed `.github/workflows/pr-check.yml`** — typecheck + unit already run in the Tekton PR-preview pipeline; component-tests now run there too (talos-infra `npm-component-tests`). Stops GitHub Actions runner billing for PR checks.

## CI consolidation note

[...]

PR #4234, opened 2026-08-21
ci: give a commit on `main` a CI verdict again, and pin that it keeps one

ZacxDev:
ci: give a commit on `main` a CI verdict again, and pin that it keeps one

Nothing has produced a CI verdict for a commit on `main` since 2026-06-14. The
cause was not a decision: `.github/workflows/pr-check.yml` was the only workflow
this repo has ever had with `push: branches: [main]`, and #2547 deleted it while
moving PR checks to Tekton. Every workflow written since — lint (#3362),
submodule-pin-guard, schema-drift (#3643), windows-dev-env (#4162) — is
`pull_request`-only, so the main-push trigger was never rebuilt and nothing said
it was gone.

Tekton does not close the gap. Its `pr-check` pipeline is driven by a resource
that watches open pull requests targeting `main`; the only main-branch trigger
builds a container image for the staging deployment, runs no test suite and
posts no commit status. "Tekton covers main" is the assumption that let this
persist for two months.

The old June runs stayed `success`, so "is main green?" went on answering yes.
That is the expensive half: the signal was not wrong, it was stale, and
staleness is invisible unless you read the dates. a3e790ff2e was pushed straight
to `main` with no PR, broke model-file-scan.service.test.ts, and sat undetected
for ~7 hours — surfacing only when unrelated PRs inherited it through their
merge commits, which reads as a CI outage rather than as one bad commit.

What changes

- `lint.yml` also runs on `push` to `main`. Reusing this file rather than adding
  a second workflow keeps one definition of each tier; a duplicate would drift.
- `eslint` is gated to pull requests. Every step in it diffs against
  `origin/$BASE_REF`, and `github.base_ref` is empty on a push, so on `main` it
  would fail for reasons unrelated to the code. The consequence is stated in the
  file rather than left implied: the added-file lint/Prettier gate and the
  stray-migrations guard stay reachable only through a PR.
- `typecheck` gains an explicit push arm. It is redundant today only because
  `github.event.pull_request` is null on a push, which is an accident of
  null-comparison semantics rather than a decision.
- `unit` is report-only on pull requests exactly as before, and renders its real
  verdict on `main`. `continue-on-error` makes a run report `success` while the
  step underneath fails — on `main` that would replace the stale TRUE with a
  live one, which is worse. Whether the PR side flips to blocking is a separate,
  deliberately-unmade decision and is untouched here.
- The concurrency group varies per commit on a push. Keyed on `github.ref` it is
  one group for the whole branch, so a busy `main` would cancel its own runs,
  leave only the last commit checked, and render the losers as `cancelled` — a
  status this repo already cannot distinguish from a real failure.

Coverage

`scripts/__tests__/main-branch-ci-coverage.test.ts` asserts the outcome, not the
spelling: it names no workflow file and no job, and asks whether a push to
`main` reaches a whole-repo typecheck and the unit suite at all. Splitting or
renaming things keeps it green; losing main coverage does not, however it is
lost. It also pins the three ways this would decay quietly — an unconditionally
report-only job on the push path, a job whose steps consume PR-only context, and
a branch-wide concurrency group.

Six mutants, each killed by the test that owns the property: reverting the
workflow to its pre-change state (the historical defect) fails the coverage
assertions; `continue-on-error: true`, an un-gated base-ref consumer, a
`github.ref` concurrency group, and gating either the typecheck or the unit tier
away from push each fail exactly one test, by name.

---

Verification

- The push path was exercised on a real `push` event from a throwaway branch (`zach/ci-main-verify`, deleted after) carrying the same job definitions with its own name temporarily added to `push.branches`. Run IDs are in the PR discussion below.
- The guard test: 6 mutants, each killed by the test that owns the property. Green at HEAD, red when the workflow is reverted to its pre-change state.
- `actionlint` clean on the workflow, with a positive control confirming it can go red.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,11 +55,18 @@ name: Lint
 # which is why the gap read as one thing and stayed unexplained for two months:
 #
 #   1. `.github/workflows/pr-check.yml` was deleted by #2547 ("runs in Tekton,
-#      removes GH Actions pr-check"). It was the ONLY workflow this repo has
-#      ever had with `push: branches: [main]`. Everything built afterwards —
-#      this file (#3362, 2026-07-24), submodule-pin-guard, schema-drift (#3643),
-#      windows-dev-env (#4162) — is `pull_request`-only, so the main-push
-#      trigger was never rebuilt. Nobody decided that; it was lost in a move.
+#      removes GH Actions pr-check"). It carried the only live push-to-`main`
+#      trigger. Enumerated rather than assumed — every revision of every one of
+#      the 12 files that has ever existed under `.github/workflows/` was parsed,
+#      and exactly TWO have ever fired on a branch push to `main`:
+#      `docker-deploy.yml` (removed 2023-11) and `pr-check.yml`. Two near
+#      misses are worth naming so they are not re-counted: `auth-app.yml` has a
+#      `push:` block but filters on `tags:` only, which never fires on a branch
+#      push, and the last `docker-deploy.yml` had `main` commented out under
+#      `branches:`. Everything built after the deletion — this file (#3362,
+#      2026-07-24), submodule-pin-guard, schema-drift (#3643), windows-dev-env
+#      (#4162) — is `pull_request`-only, so the trigger was never rebuilt.
+#      Nobody decided that; it was lost in a move.
 #   2. CodeQL, whose runs carried the "Push on main" title in the run list,
 #      stopped the same afternoon (code-scanning default setup now reads
 #      `not-configured`). Unrelated cause, same date, no test coverage either
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -47,22 +47,93 @@ name: Lint
 #
 # The modified-file steps flip to blocking once the backlog is cleared, planned to
 # ride along with the Prettier 2->3 upgrade (which forces a reformat anyway).
+#
+# ── ALSO RUNS ON PUSH TO `main`, and that is not decoration ──────────────────
+#
+# Between 2026-06-14 and 2026-08-21 NO CI of any kind produced a verdict for a
+# commit on `main`. Two unrelated mechanisms happened to stop on the same day,
+# which is why the gap read as one thing and stayed unexplained for two months:
+#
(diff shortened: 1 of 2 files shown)


ZacxDev:
Verification of the push path (run IDs, since `branches: [main]` cannot fire before merge).

Exercised on a real `push` event from a throwaway branch carrying identical job definitions. The only non-comment difference between what ran and what merges here is the temporary branch name in `push.branches` — confirmed by diff.

**Clean tree — run [32458869146](https://github.com/civitai/civitai/actions/runs/32458869146), conclusion `success`**

| Job | Result |
|---|---|
| Typecheck (main pushes / fork PRs / non-main base) | success |
| Unit tests | success |
| Package unit tests | success |
| App unit tests + typecheck | success |
| ESLint + Prettier (changed files) | **skipped** (correct — PR-gated) |

So the gate does not land red on day one.

**Negative control — run [32459112185](https://github.com/civitai/civitai/actions/runs/32459112185), conclusion `failure`**

Same tree plus a deliberate type error and a deliberately failing test:

| Job | Result |
|---|---|
| Typecheck | **failure** |
| Unit tests | **failure at the JOB level** |
| Package / App unit tests | success |

The unit row is the one that matters: it confirms `continue-on-error: ${{ github.event_name == 'pull_request' }}` evaluates false on a push, so the run is a real failure and not a `success` hiding a failed step.

**Concurrency**

The second push did *not* cancel the first — both push runs ran to completion. Under the old `group: lint-${{ github.ref }}` the first would have been cancelled and rendered `cancelled`.

**PR path unchanged**

On this PR: ESLint ✅, Package unit tests ✅, App unit tests + typecheck ✅, and `Typecheck` correctly `skipped` (internal PR targeting main → Tekton owns it). Earlier Lint runs on this branch show `cancelled` — that is `cancel-in-progress` on the PR path still behaving exactly as before.

**Guard test**

6 mutants, each killed by the test that owns the property: reverting the workflow to its pre-change state fails the coverage assertions; `continue-on-error: true`, an un-gated base-ref consumer, a `github.ref` concurrency group, and gating either the typecheck or the unit tier away from push each fail exactly one test, by name. `actionlint` clean, with a positive control confirming it can go red.

PR #4235, opened 2026-08-21
chore(deps): vitest 4.0.18 → 4.1.11 — clears 4 critical dev-only alerts

ZacxDev:
Dev-tooling only. No runtime dependency moves, no source change. Base is `main`; this PR is **not** stacked on anything.

## What it clears

**4 Dependabot alerts, all CRITICAL, all `scope=development`.**

| # | CVE | Package | Summary |
|---|---|---|---|
| 331 | CVE-2026-73653 | `@vitest/browser` | Browser Mode provider commands bypass the file-access restriction |
| 272 | CVE-2026-53633 | `@vitest/browser` | Exposed Browser Mode API can proxy CDP and overwrite files |
| 269 | CVE-2026-47428 | `@vitest/browser` | Unsanitized `otelCarrier` query parameter served back |
| 264 | CVE-2026-47429 | `vitest` | UI server allows arbitrary file read and execution |

Repo criticals go **6 → 2** with this PR alone.

These numbers are **measured against the regenerated lockfile, not estimated**. Every open alert was re-evaluated by resolving each advisory's vulnerable version ranges against the lockfile's `packages:` set, before and after. Result: **4 cleared, 0 introduced, 0 other alerts moved.** As a control the same pass was run baseline-against-baseline: it reproduces all 154 open alerts at exactly the severity split GitHub reports (6 critical / 75 high / 59 moderate / 14 low) and reports 0 cleared, so it is neither missing alerts nor inventing them.

## Mechanism, and why the whole family moves

The brief for this change was "bump `vitest` and `@vitest/browser`; it is a pure in-range lockfile refresh with no manifest edit". That turns out not to be achievable, for a reason worth recording:

**The vitest packages peer-depend on each other by EXACT version** — `peerDependencies: { vitest: 4.0.18 }`, not a range. And `@vitest/browser-playwright` is pinned exactly (`"4.0.18"`, no caret) in the root manifest.

So bumping only the two named packages resolves to a tree that still contains a vulnerable copy:

```
@vitest/browser@4.0.18     <- still here, dragged in by browser-playwright
@vitest/browser@4.1.11
✕ unmet peer vitest@4.0.18: found 4.1.11
```

A still-resolved `@vitest/browser@4.0.18` means alerts #331/#272/#269 do **not** clear. Moving `vitest`, `@vitest/browser`, `@vitest/browser-playwright` and `@vitest/coverage-v8` together is what leaves exactly one 4.1.11 of each and no unmet peers.

## Diff shape

20 changed manifest lines across 17 files, every one a version string; plus the lockfile.

`pnpm up` was deliberately **not** used to produce them — it also re-sorts dependency keys alphabetically, which added unrelated churn to `apps/moderator` and `packages/civitai-redis`. The edits here are value-only, so key order and formatting are untouched.

The floors move `^4.0.18` → `^4.1.11` rather than staying put, so a fresh resolve cannot land back on a vulnerable version; only the lockfile would be holding it otherwise.

Lockfile churn is 593 lines, confined to the vitest family and its own transitive dev deps (`@blazediff/core` replacing `pixelmatch`, `es-module-lexer`, the `@babel/*` set coverage-v8 pulls). No runtime-scope package changed version.

## Test matrix

`origin/main` @ `df7733b99c` vs this branch, same toolchain, same machine. Baseline recorded **first**, on an unmodified checkout, with a green `pnpm install --frozen-lockfile` as the control.

| suite | baseline | after |
|---|---|---|
| `pnpm typecheck` | 0 errors | 0 errors |
| `pnpm test:unit:run` | 20407 passed / 25 skipped (1301 files) | 20407 passed / 25 skipped (1301 files) |
| `pnpm test:packages:run` | 1079 passed / 8 skipped (78 files) | 1079 passed / 8 skipped (78 files) |
| `pnpm test:apps:run` | 691 passed / 35 skipped (68 files) | 691 passed / 35 skipped (68 files) |

Identical counts on every suite. Counts are read off the runner's summary lines, not inferred from exit codes.

## Deliberately not touched

**`fast-xml-parser`** — alert **#137** (critical) stays open **on purpose**. The `"@aws-sdk/core>fast-xml-parser": "5.2.5"` override pins the vulnerable version because a blanket CVE bump of this package broke S3 error parsing in production once already (#3267). Both the override and both resolved versions are byte-identical to `main` here. Clearing #137 requires fixing the parsing regression first — it is not a bump.

**`nodemailer`** — declared `^6.8.0` against a top fix of `9.0.1`, i.e. a three-major bump of the email sending path. That is not a lockfile refresh; out of scope for this PR.

## Note on the working tree

`packages/civitai-db-schema/src/enums.ts` is rewritten by the `postinstall` generator on an **unmodified** `main` too (verified in two independent clean worktrees). It is pre-existing drift, so it is left out of this commit rather than smuggled in.


ZacxDev:
**CI note: `preview / component-tests` is red here, and it is not this PR.**

Everything else is green — 18 of 19 checks pass, including `tekton / typecheck`, `Unit tests`, `Package unit tests`, `App unit tests + typecheck` and the rest of the `preview /` suite. (`Typecheck (fork PRs / non-main base)` correctly reports `skipping`; Tekton is the authoritative typecheck for an internal PR to `main`, and it passes.)

The control, run before assuming a cause: **PR #4234 fails `preview / component-tests` identically**, with the same status description (`Component suite failed (report-only, not blocking)`). That PR changes exactly two files — `.github/workflows/lint.yml` and `scripts/__tests__/main-branch-ci-coverage.test.ts` — neither of which is in the component-test graph, and it touches neither vitest nor OpenTelemetry. A failure that reproduces on a change which cannot have caused it is not attributable to this one.

The pattern also fits the timeline: the three PRs whose runs post-date the current `main` tip (`df7733b99c`) all fail this check, while every older open PR whose run predates it passes.

**What I could not do:** reproduce it locally. `pnpm test:component` on a *pristine* `origin/main` worktree here fails for a **different** reason — the Nix-provided `PLAYWRIGHT_BROWSERS_PATH` has no executable at `chromium_headless_shell-1200`, so the browser never launches. That is a local toolchain gap, not the CI failure, and I am not treating the two as the same thing. So the evidence above is CI-tier evidence, not a local reproduction.

Worth someone looking at separately — it appears to be red on `main` rather than on any one PR.

PR #4236, opened 2026-08-21
chore(deps): OpenTelemetry 0.211.0 → 0.219.0 — removes protobufjs entirely (1 critical + 15)

ZacxDev:
Base is `main`; this PR is **not** stacked on anything. Sibling PR #4235 (vitest) is independent — either can merge first.

This one touches the **runtime** telemetry path, so it deserves a different review standard from #4235, which is dev-tooling only. That is why they are two PRs.

## What it clears

**16 Dependabot alerts — 1 critical, 8 high, 7 medium — all `scope=runtime`.**

| # | Sev | CVE | Package |
|---|---|---|---|
| 185 | **critical** | CVE-2026-41242 | `protobufjs` — arbitrary code execution |
| 271 | high | CVE-2026-48712 | `protobufjs` |
| 216 | high | CVE-2026-44293 | `protobufjs` |
| 214 | high | CVE-2026-44291 | `protobufjs` |
| 213 | high | CVE-2026-44290 | `protobufjs` |
| 212 | high | CVE-2026-44289 | `protobufjs` |
| 210 | high | CVE-2026-44902 | `@opentelemetry/sdk-node` |
| 209 | high | CVE-2026-44902 | `@opentelemetry/sdk-node` |
| 208 | high | CVE-2026-44902 | `@opentelemetry/exporter-prometheus` |
| 323 | medium | CVE-2026-59877 | `protobufjs` |
| 270 | medium | CVE-2026-54269 | `protobufjs` |
| 247 | medium | CVE-2026-45740 | `protobufjs` |
| 217 | medium | CVE-2026-44294 | `protobufjs` |
| 215 | medium | CVE-2026-44292 | `protobufjs` |
| 211 | medium | CVE-2026-44288 | `protobufjs` |
| 218 | medium | CVE-2026-44288 | `@protobufjs/utf8` |

Measured against the **regenerated lockfile**, not estimated: every open alert was re-evaluated by resolving each advisory's vulnerable version ranges against the lockfile's `packages:` set, before and after. **16 cleared, 0 introduced.** As a control the same pass run baseline-against-baseline reproduces all 154 open alerts at exactly the split GitHub reports (6 critical / 75 high / 59 moderate / 14 low), so the instrument is neither missing alerts nor inventing them.

With #4235, criticals go **6 → 1** — the remaining one being `fast-xml-parser`, left open on purpose (below).

## Mechanism

`@opentelemetry/otlp-transformer@0.211.0` depends on `protobufjs` pinned **exactly** at `8.0.0`:

```yaml
'@opentelemetry/otlp-transformer@0.211.0':
  dependencies:
    protobufjs: 8.0.0        # exact, not a range
```

That single edge is the sole cause of all 12 open `protobufjs` alerts. It cannot be fixed by an override or a lockfile refresh:

- the existing `"protobufjs@7": "^7.5.6"` override scopes to major **7** and cannot reach an 8.0.0;
- the dependency is an exact pin, so there is no in-range escape;
- `^0.211.0` cannot resolve to 0.217+, so the caret has to move.

At **0.219.0**, `otlp-transformer` has **no `protobufjs` dependency at all**. The package and its five `@protobufjs/*` helpers leave the tree entirely.

## Five declarations move, not one

`@opentelemetry/sdk-node` is **not** the sole root of the 0.211.0 subtree, and this is the part worth checking closely. `@opentelemetry/exporter-trace-otlp-proto` and `@opentelemetry/exporter-logs-otlp-proto` are declared **directly** in the root manifest at `^0.211.0`, and each depends on `otlp-transformer@0.211.0` on its own.

Measured, not reasoned: bumping `sdk-node` alone leaves `protobufjs@8.0.0` resolved and clears **3** of the 16 alerts — **none of them the critical**. So the change moves all five root declarations plus the two in `packages/civitai-telemetry`.

`0.219.0` rather than the latest `0.221.0` because the root manifest already pins `@opentelemetry/instrumentation` at exactly `0.219.0` — this lands the family on a version already present in the tree, and `^0.219.0` on a 0.x package resolves within `0.219.x` only, so it stays deterministic.

## One behavioural change — please read this part

`@opentelemetry/sdk-logs@0.219.0` makes `forceFlush()` a **required** member of `LogRecordExporter`. Read off the installed builds, not inferred:

- at **0.211.0** the interface declared only `export` and `shutdown`, and `BatchLogRecordProcessorBase` called `this._exporter.forceFlush()` **zero** times;
- at **0.219.0** `_flushAll()` — which runs on the **shutdown** path — calls it.

**Production is unaffected.** The real `OTLPLogExporter` inherits `forceFlush` from `OTLPExporterBase`, and `src/instrumentation.node.ts` typechecks clean against 0.219.0 with no `any`/`@ts-ignore` escape hatches, so the typecheck genuinely covers that surface.

What broke was the **hand-rolled fake exporter** in the telemetry test, which was short of the real contract. That is worth flagging rather than glossing over, because of *how* it failed: the processor wraps the call in a `try/catch` routed to `globalErrorHandler`, so the missing method **did not throw**. The flush was silently abandoned and the test read **zero** exported records — indistinguishable from a shutdown that never flushed at all. A fake that under-specifies the real interface is exactly the shape that hides this.

The fake now implements the whole interface. **The repair does not blunt the test**: with it in place, removing the `await` on `t.shutdown()` in `registerOtelShutdown` still fails `SIGTERM FLUSHES a batch the timer is still holding` on that test's own assertion (`expected [] to have a length of 1`). Mutant killed for the right reason, then reverted.

## Test matrix

`origin/main` @ `df7733b99c` vs this branch, same toolchain, same machine. Baseline recorded **first**, on an unmodified checkout, with a green `pnpm install --frozen-lockfile` as the control.

| suite | baseline | after |
|---|---|---|
| `pnpm typecheck` | 0 errors | 0 errors |
| `pnpm test:unit:run` | 20407 passed / 25 skipped (1301 files) | 20407 passed / 25 skipped (1301 files) |
| `pnpm test:packages:run` | 1079 passed / 8 skipped (78 files) | 1079 passed / 8 skipped (78 files) |
| `pnpm test:apps:run` | 691 passed / 35 skipped (68 files) | 691 passed / 35 skipped (68 files) |

Identical counts on every suite. Counts read off the runner's summary lines, not inferred from exit codes. The OTLP-touching suite (`packages/civitai-telemetry`) is included and passes: 28/28.

**Scope limit, stated honestly:** this was verified by typecheck plus the unit/package/app suites. A full `next build` and a live OTLP export against a collector were **not** run here, so "the telemetry path still builds and typechecks" is the claim — not "traces and logs were observed arriving".

## Lockfile churn

Outside `@opentelemetry/*`, four packages and no more:

- **removed** — `protobufjs@8.0.0` and its helper set (`@protobufjs/{codegen@2.0.4,eventemitter@1.1.0,fetch@1.1.0,inquire@1.1.0,utf8@1.1.0}`). `protobufjs@7.6.5` and its own newer helpers are untouched.
- **added** — `@grpc/grpc-js@1.14.4` and `@grpc/proto-loader@0.8.1`, *alongside* the existing 1.13.4 / 0.7.15, pulled by the 0.219.0 gRPC exporters.

Net resolved packages 2505 → 2496. No unmet OpenTelemetry peers.

## Deliberately not touched

**`fast-xml-parser`** — alert **#137** (critical) stays open **on purpose**. The `"@aws-sdk/core>fast-xml-parser": "5.2.5"` override pins the vulnerable version because a blanket CVE bump of this package broke S3 error parsing in production once already (#3267). The override and both resolved versions (`5.2.5`, `5.10.1`) are identical to `main` here. Clearing #137 requires fixing the parsing regression first — it is not a bump.

**`nodemailer`** — declared `^6.8.0` against a top fix of `9.0.1`, i.e. a three-major bump of the email sending path. Out of scope.

## Note on the working tree

`packages/civitai-db-schema/src/enums.ts` is rewritten by the `postinstall` generator on an **unmodified** `main` too (verified in two independent clean worktrees). Pre-existing drift, left out of this commit rather than smuggled in.


ZacxDev:
**CI note: `preview / component-tests` is red here, and it is not this PR.**

Everything else is green — 18 of 19 checks pass, including `tekton / typecheck`, `Unit tests`, `Package unit tests`, `App unit tests + typecheck` and the rest of the `preview /` suite. (`Typecheck (fork PRs / non-main base)` correctly reports `skipping`; Tekton is the authoritative typecheck for an internal PR to `main`, and it passes.)

The control, run before assuming a cause: **PR #4234 fails `preview / component-tests` identically**, with the same status description (`Component suite failed (report-only, not blocking)`). That PR changes exactly two files — `.github/workflows/lint.yml` and `scripts/__tests__/main-branch-ci-coverage.test.ts` — neither of which is in the component-test graph, and it touches neither vitest nor OpenTelemetry. A failure that reproduces on a change which cannot have caused it is not attributable to this one.

The pattern also fits the timeline: the three PRs whose runs post-date the current `main` tip (`df7733b99c`) all fail this check, while every older open PR whose run predates it passes.

**What I could not do:** reproduce it locally. `pnpm test:component` on a *pristine* `origin/main` worktree here fails for a **different** reason — the Nix-provided `PLAYWRIGHT_BROWSERS_PATH` has no executable at `chromium_headless_shell-1200`, so the browser never launches. That is a local toolchain gap, not the CI failure, and I am not treating the two as the same thing. So the evidence above is CI-tier evidence, not a local reproduction.

Worth someone looking at separately — it appears to be red on `main` rather than on any one PR.

PR #4237, opened 2026-08-21
fix(flipt): print the real segment keys on rules and rollouts

JustMaier:
ClickUp: https://app.clickup.com/t/868kunq3p

## The defect

`.claude/skills/flipt/flipt.mjs` read `segment.keys` — a field the Flipt API never returns. Two consequences:

- **Rollouts** printed `Segment: undefined`.
- **Rules** were worse: the loop gated on a nested `rule.segment` object that does not exist on a rule at all, so the branch never ran. A variant flag printed a `Rules:` header with nothing under it — no "undefined" to hint that anything was missing.

(The ticket title says the code reads `segment.key` singular. It reads `.keys` plural. Neither is the real field.)

## What the API actually returns

Confirmed against the live instance, not from the ticket:

```jsonc
// rollout
{ "type": "SEGMENT_ROLLOUT_TYPE",
  "segment": { "value": true, "segments": ["testers"], "segmentOperator": "OR_SEGMENT_OPERATOR" } }

// rule — segments live directly on the rule, no nested `segment`
{ "segmentOperator": "OR_SEGMENT_OPERATOR", "segments": ["moderators", "testers"],
  "distributions": [{ "variant": "primary", "rollout": 100 }] }
```

So it is `segments[]` in both places, as the ticket suspected, but at two different depths.

## The fix

One helper for both call sites. It also surfaces `segmentOperator`, which was never printed: with two or more segments the operator changes who the rollout reaches, so the keys are joined with ` AND ` / ` OR `. A single segment prints bare — there is nothing to combine.

## Before / after

Real flags, live API. Description lines trimmed for length; nothing else altered.

**Before**

```
===== user-hubs =====
Flag: user-hubs
Status: ✗ DISABLED
Type: BOOLEAN_FLAG_TYPE

Rollout Rules:
  - Segment: undefined → true
===== creator-shop =====
Flag: creator-shop
Status: ✓ ENABLED
Type: BOOLEAN_FLAG_TYPE

Rollout Rules:
  - Segment: undefined → true
===== bitdex-image-search =====
Flag: bitdex-image-search
Status: ✓ ENABLED
Type: VARIANT_FLAG_TYPE

Variants:
  - off
  - shadow
  - primary

Rules:
```

**After**

```
===== user-hubs =====
Flag: user-hubs
Status: ✗ DISABLED
Type: BOOLEAN_FLAG_TYPE

Rollout Rules:
  - Segment: testers → true
===== creator-shop =====
Flag: creator-shop
Status: ✓ ENABLED
Type: BOOLEAN_FLAG_TYPE

Rollout Rules:
  - Segment: testers OR CreatorProgram OR moderators → true
===== bitdex-image-search =====
Flag: bitdex-image-search
Status: ✓ ENABLED
Type: VARIANT_FLAG_TYPE

Variants:
  - off
  - shadow
  - primary

Rules:
  - Segment: moderators OR testers
    → variant "primary" at 100%
  - Segment: all-users
    → variant "primary" at 100%
```

`user-hubs` is the case that motivates this: a boolean flag whose rollout is scoped to `testers`. A readout that hides the segment is how a scoped rollout gets misread as off — or, with `enabled: true`, how a public flag gets misread as internal-only.

## Verification

There is no test suite over `.claude/skills/`, so the output above is the evidence. `pnpm run prettier:write` reports `no uncommitted .ts/.tsx files` — it globs `.ts`/`.tsx` only and does not cover `.mjs`; the file was not hand-reformatted.

## Not changed

`setRollout` posts `segmentKeys` on the write path (line ~395). Every write in this skill is behind `refuseWrite` — flag changes go through a PR to the GitOps state repo — so that field is unexercised and unverifiable without performing a write. Out of scope here; flagged rather than guessed at.

### DIFF
--- a/.claude/skills/flipt/flipt.mjs
+++ b/.claude/skills/flipt/flipt.mjs
@@ -265,6 +265,14 @@ async function getFlag(key) {
   console.log(`Flag: ${data.key}`);
   console.log(`Status: ${statusColor}${status}\x1b[0m`);
   console.log(`Type: ${data.type}`);
+  // `enabled` on a boolean flag is the value returned when NO rollout matches, so
+  // enabled:true beside a segment rollout is public, not scoped to that segment.
+  if (data.type === 'BOOLEAN_FLAG_TYPE') {
+    const reach = data.enabled
+      ? '\x1b[33mPUBLIC — everyone no rollout matches gets true\x1b[0m'
+      : 'off for everyone no rollout matches';
+    console.log(`Default: ${data.enabled} — ${reach}`);
+  }
   if (data.description) {
     console.log(`Description: ${data.description}`);
   }
--- a/.claude/skills/flipt/flipt.mjs
+++ b/.claude/skills/flipt/flipt.mjs
@@ -235,6 +235,16 @@ async function listFlags() {
   }
 }
 
+// A rollout matching two segments under AND vs OR reaches completely different
+// audiences, so the operator is only omitted when there is nothing to combine.
+function formatSegments(segments, segmentOperator) {
+  const keys = segments ?? [];
+  if (keys.length === 0) return 'unknown';
+  if (keys.length === 1) return keys[0];
+  const op = segmentOperator === 'AND_SEGMENT_OPERATOR' ? 'AND' : 'OR';
+  return keys.join(` ${op} `);
+}
+
 async function getFlag(key) {
   let data;
   try {
@@ -271,12 +281,9 @@ async function getFlag(key) {
   if (data.rules?.length > 0) {
     console.log('\nRules:');
     for (const rule of data.rules) {
-      if (rule.segment) {
-        const segs = rule.segment.keys?.join(', ') || 'unknown';
-        console.log(`  - Segment: ${segs}`);
-        for (const dist of rule.distributions || []) {
-          console.log(`    → variant "${dist.variant}" at ${dist.rollout}%`);
-        }
+      console.log(`  - Segment: ${formatSegments(rule.segments, rule.segmentOperator)}`);
+      for (const dist of rule.distributions || []) {
+        console.log(`    → variant "${dist.variant}" at ${dist.rollout}%`);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


JustMaier:
## Second commit: `Status:` alone was still misleading

Raised in review: `Status: ✓ ENABLED` beside `Segment: testers OR CreatorProgram OR moderators` reads as *"scoped to those three segments"*. For a **boolean** flag it means close to the opposite — `enabled` is the value Flipt returns when **no rollout matches**, so `enabled: true` plus a segment rollout is **public**, and the rollout only overrides for the people it matches. Fixing the segment line without this made the readout more legible while leaving it just as wrong.

One line, boolean flags only:

```
Flag: creator-shop
Status: ✓ ENABLED
Type: BOOLEAN_FLAG_TYPE
Default: true — PUBLIC — everyone no rollout matches gets true

Rollout Rules:
  - Segment: testers OR CreatorProgram OR moderators → true
```
```
Flag: user-hubs
Status: ✗ DISABLED
Type: BOOLEAN_FLAG_TYPE
Default: false — off for everyone no rollout matches

Rollout Rules:
  - Segment: testers → true
```

Same rollout shape, opposite reach, and now that difference is on the screen. Variant flags are unchanged — `enabled` has no default-value role there.

## Evidence

Not from the repo's comments — from the live evaluation API, entity matching no segment, read-only:

| flag | context | result |
|---|---|---|
| `creator-shop` | anonymous, no context | `enabled: true`, `reason: DEFAULT_EVALUATION_REASON`, `segmentKeys: []` |
| `creator-shop` | `isModerator=false, isMember=false` | `enabled: true`, `DEFAULT_EVALUATION_REASON`, `segmentKeys: []` |
| `user-hubs` | anonymous, no context | `enabled: false`, `DEFAULT_EVALUATION_REASON`, `segmentKeys: []` |
| `user-hubs` | `isModerator=false, isMember=false` | `enabled: false`, `DEFAULT_EVALUATION_REASON`, `segmentKeys: []` |

`DEFAULT_EVALUATION_REASON` with an empty `segmentKeys` is the mechanism named directly: no rollout matched, so the flag's own `enabled` was returned. `user-hubs` is the control — identical rollout shape, `enabled: false`, and it comes back false.

The repo already documents this in two places (`src/server/flipt/client.ts` on `EARLY_ADOPTER`, and the `earlyAdopter` comment in `feature-flags.service.ts` citing flipt-state#62, where the same defect was caught pre-merge). The skill was the one place that showed the flag without saying it.

PR #4246, opened 2026-08-21
feat(reaction-abuse): give automated exclusions a ModActivity trail a moderator can see

ZacxDev:
## Problem

`/api/admin/reaction-abuse` is the commit path for the scheduled reaction-abuse loop: it adds
reaction-farm sockpuppets to `metricExcludedUsers`, which drops their reactions from metrics and
ranking (forward-only, reversible, never a ban).

It wrote **only** to ClickHouse. Nothing in the moderator app could see that an automated system
had acted on an account:

- a creator opening a ticket about lost ranking left no trace on the user-lookup page;
- there was no in-app record of *when* or *which direction*;
- reversing an exclusion was an out-of-band operation only.

## What this does

`ModActivity` is already the seam for exactly this. `autoMuteScam`
(`src/server/jobs/entity-moderation.ts`) writes one, and the moderator app's board reads it back as
its auto-blocked strip. This writes the same kind of row from the exclusion path:

- `autoExcludeReactionAbuse` / `autoUnexcludeReactionAbuse` — **two** activities, so a reversal is
  not indistinguishable from the action it reverses.
- Under the same `-1` automated-writer sentinel `autoMuteScam` uses. The moderator board filters
  `userId > 0` for its "who last worked this queue" panel, so these rows cannot be mistaken for a
  person working a queue.
- One call per request: `trackModActivity` UNNESTs the id array into a row per user.

**No migration.** `ModActivity.activity` is a `String` column, so the union in
`moderator.service.ts` is a TypeScript contract only.

### Why the audit failure is reported rather than thrown

The ClickHouse write is what actually excludes; the `ModActivity` row is only its record. So a
failure to write that record:

- **must not throw** — a 500 on a request whose exclusion had already applied would tell the caller
  nothing happened when it did, and its retry would double-write the audit row, since
  `ModActivity`'s `(activity, entityType, entityId)` unique index is being dropped and the
  `ON CONFLICT DO NOTHING` deliberately names no target;
- **must not be silent** — a silently missing audit row is the exact gap this change exists to close.

So it is logged server-side and the outcome rides back in the response as `auditRecorded`.

⚠️ **`auditRecorded` is inert until a caller branches on it.** Today the observable signal on the
failure path is the server-side log line; the response field is there so the scheduled caller can
surface it in its own digest, which is a follow-up in the deployment repo. Calling that out rather
than implying the field does work it does not yet do.

## What is deliberately NOT here

The free-text *why* — the judge's confidence and the ring evidence — is **not** duplicated into
`ModActivity`, which has no free-text column. It already lives in `metricExcludedUsers.reason`,
keyed by the same `userId`, so a UI can join the two. Adding a new `UserActivityType` for the
ClickHouse `userActivities` table was considered and dropped: that column's DDL is not in this repo,
so whether it would accept a new enum value is unverifiable from here — and `Tracker.track` swallows
insert errors, so a rejected value would have failed invisibly.

## Tests

`src/server/__tests__/reaction-abuse-endpoint.test.ts` — 6 tests.

Verified matrix:

| | result |
|---|---|
| at `070acc9b59` (base, source reverted, test kept) | **4 failed / 2 passed** |
| at this commit | **6 passed** |
| adjacent suites touching `trackModActivity` (4 files) | 159 passed |
| `pnpm test:lint-rules` (9 files) | 265 passed |

The 2 that pass on the base are **invariant guards, not regression coverage** — "no audit row when
the exclusion fails" and "read-only actions are unaudited" are trivially true when nothing writes
audit rows at all. Labelling them rather than counting them.

Mutation-tested, each mutant killed by its intended assertion:

| mutation | outcome |
|---|---|
| sentinel `-1` → `0` | 2 tests fail on `toHaveBeenCalledWith(-1, …)` |
| exclude branch emits the *unexclude* activity | 1 test fails on the activity name |
| audit call moved *before* the ClickHouse insert | ordering test fails; the "no row when the exclusion fails" guard also fires, so it is reachable |

## Follow-ups (not in this PR)

1. Surface `auditRecorded` in the scheduled caller's digest (deployment repo).
2. A moderator-facing panel reading these rows back, and an in-app unexclude behind its own
   permission.
3. Minor: the moderator app's `AUTOMATED_USER_ID = 0` constant is documented as *the* automated
   sentinel, but the only automated writers use `-1`. Both are excluded by its `> 0` filter so
   nothing is broken — the constant just names a value nobody writes. Left alone to keep this PR
   scoped.

### DIFF
--- a/src/pages/api/admin/reaction-abuse.ts
+++ b/src/pages/api/admin/reaction-abuse.ts
@@ -149,12 +149,17 @@ const schema = z
     // cross-app mod-action registry uses. Omitted by the scheduled poller, which is genuinely not a
     // person; supply it when a human runs `unexclude` so the reversal names them.
     //
-    // 🔴 `.nullish()`, not `.optional()`. `null` is the natural JSON for "no acting moderator", and
-    // under `.optional()` it is a ZodError — so the request 400s and THE EXCLUSION NEVER HAPPENS.
-    // An optional audit-attribution field must never be able to fail the write it annotates; an
-    // absent actor is exactly the sentinel case. `''` is folded in for the same reason (a shell
-    // recipe interpolating an unset variable), while a genuinely malformed value still 400s rather
-    // than silently landing a wrong id on a moderation record.
+    // 🔴 Preprocessed, not bare `.optional()`. `null` is the natural JSON for "no acting moderator",
+    // and under `.optional()` it is a ZodError — so the request 400s and THE EXCLUSION NEVER
+    // HAPPENS. An optional audit-attribution field must never be able to fail the write it
+    // annotates; an absent actor is exactly the sentinel case. `''` is folded in for the same reason
+    // (a shell recipe interpolating an unset variable) — and `.nullish()` alone would NOT have
+    // covered that one, which is why this is a preprocess rather than a looser modifier.
+    //
+    // What this does NOT do is make the field type-safe: `z.coerce` still turns `true` into 1 and
+    // `[7]` into 7, so a malformed actor can land a wrong-but-plausible id on a moderation row.
+    // Tolerated because the caller is trusted and token-gated; `0`, negatives and fractions are
+    // rejected outright.
     actorUserId: z.preprocess(
       (v) => (v === null || v === '' ? undefined : v),
       z.coerce.number().int().positive().optional()
--- a/src/pages/api/admin/reaction-abuse.ts
+++ b/src/pages/api/admin/reaction-abuse.ts
@@ -30,8 +30,10 @@
  *                   (count, distinct IPs, share from farm IPs) — for drill-in.
  *   exclude       - {userIds: number[], reason}
  *                   Add users to metricExcludedUsers (active=1). Idempotent.
+ *                   Also writes a `ModActivity` row per user so the action is visible to a
+ *                   moderator — see `recordAutomatedAction`. Response carries `auditRecorded`.
  *   unexclude     - {userIds: number[]}
- *                   Reverse an exclusion (insert active=0; latest row wins).
+ *                   Reverse an exclusion (insert active=0; latest row wins). Audited the same way.
  *   list          - {limit?=500}
  *                   Currently-excluded users (active=1), newest first.
  *
@@ -46,7 +48,38 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import * as z from 'zod';
 import { clickhouse } from '~/server/clickhouse/client';
+import { trackModActivity } from '~/server/services/moderator.service';
 import { WebhookEndpoint } from '~/server/utils/endpoint-helpers';
+
+/**
+ * The moderator-visible audit trail for an action this endpoint just committed.
+ *
+ * `-1` is the automated-writer sentinel the main app already uses for `autoMuteScam`
+ * (`~/server/jobs/entity-moderation.ts`); the moderator app's board filters `userId > 0` so these
+ * rows can never be mistaken for a person working a queue.
+ *
+ * 🔴 The ClickHouse write is what actually excludes; this is only its record. So a failure here is
+ * REPORTED, never thrown: throwing would 500 a request whose exclusion had already applied, telling
+ * the poller nothing happened when it did — and its retry would then double-write the audit row,
+ * since ModActivity's (activity, entityType, entityId) unique index is being dropped and
+ * `ON CONFLICT DO NOTHING` names no target. Swallowing it silently is the opposite failure and is
+ * the very gap this trail exists to close, so the outcome rides back in the response body as
+ * `auditRecorded` and the caller can say so in its digest.
+ *
+ * `entityId` takes the whole batch: `trackModActivity` UNNESTs an array into one row per user.
+ */
+async function recordAutomatedAction(
+  activity: 'autoExcludeReactionAbuse' | 'autoUnexcludeReactionAbuse',
+  userIds: number[]
+): Promise<boolean> {
+  try {
+    await trackModActivity(-1, { entityType: 'user', entityId: userIds, activity });
+    return true;
+  } catch (e) {
+    console.error('reaction-abuse audit write failed:', (e as Error).message);
+    return false;
+  }
+}
 
 const actionSchema = z.enum(['candidates', 'inspect-owner', 'exclude', 'unexclude', 'list']);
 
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Corrections to the original description, after an adversarial audit

Posting these rather than silently editing, since the first version has been up for a while.

**1. The activity names changed, and so did the reasoning behind them.**
They were `autoExcludeReactionAbuse` / `autoUnexcludeReactionAbuse`. They are now
`reactionAbuseExclude` / `reactionAbuseUnexclude`.

The audit established something I had assumed away: **`unexclude` has no automated caller.** The
scheduled job only ever calls `exclude`; `unexclude` exists as a hand-run operator recipe. So the
one path that most needs attribution — a human reversing a false positive — was being recorded
under the system sentinel with an activity named `auto…`. That is the wrong actor *and* the wrong
verb, on exactly the row a reviewer would later rely on.

The activity now says WHAT happened and `userId` says WHO, which is how `ModActivity` is designed.
A new optional `actorUserId` lets a human caller be named; omitted, it falls back to the sentinel.

**2. "`Tracker.track` swallows insert errors, so a rejected value would have failed invisibly" —
overstated.** It logs to Axiom on exhaustion (`src/server/clickhouse/tracker.ts:428-438`, `:461-472`).
"Invisibly" was wrong; "without failing the request, and only on the error board" is accurate. The
decision not to add a `UserActivityType` still stands on the stated reason — that table's DDL is not
in this repo, so I cannot verify from source whether it would accept a new enum value.

**3. The `autoMuteScam` parity claim was one-sided.** I wrote that `ModActivity` is the right seam
because the moderation board reads it back as the auto-blocked strip. That strip filters
`activity = 'autoMuteScam'`, so these new rows reach **no** proactive-review surface — only the
per-user timeline, which requires already knowing which account to look up. The ticket-response case
is served today; the false-positive-review case is not, and needs the follow-up panel. The original
wording read as if parity already held.

**4. One precondition the description left implicit is now measured.** The `ON CONFLICT DO NOTHING`
in `trackModActivity` names no target, so if `ModActivity`'s
`(activity, entityType, entityId)` unique index were still live, an
exclude → unexclude → re-exclude sequence would silently drop the third row while still reporting
`auditRecorded: true` — the exact silent gap this PR exists to close. Checked directly against both
the production and dev databases: that index is **gone** on both; only `pkey`, `createdAt`,
`(entityType, entityId, createdAt)` and `(userId, createdAt)` remain. The table is genuinely
append-only, so this cannot occur.

## Also fixed in the follow-up commit

- **Nothing bounded a hang** on the new Postgres dependency. Catching a rejection is not bounding a
  hang: a wedged write would have burned the caller's timeout budget, turned a *successful*
  exclusion into a client-side timeout, and had each retry re-insert. Now raced against a 5s bound
  well inside the caller's own.
- The failure log was a bare `console.error` nothing watches — and **deleting it left every test
  green**. Now `logToAxiom`, matching the sibling automated path, and pinned in both directions.
- The ordering guard and the no-row-on-failure guard covered only the `exclude` branch; the
  identical defect on `unexclude` was untested. Both are table-driven over both branches now.
- The read-only test was named "actions" but exercised only `list`. Adding an audit call to
  `candidates` had survived it. Now covers `candidates` and `inspect-owner` too.

Tests 6 → 14. Nine mutants, all killed, on a harness validated with both a positive and a negative
control first — one earlier "survivor" turned out to be an invalid mutant that hit a type annotation
rather than the call site, which is worth stating because it initially read as a coverage gap.

**Known non-issue:** `preview / component-tests` is red on this PR. It fails in
`src/components/Cards/ModelCard.browser.test.tsx` on a React hydration mismatch, with 1967/1967
individual tests passing — the file dies on an uncaught error. This PR touches no `.tsx`, nothing
imports the API route, the `moderator.service.ts` change is a type-only union erased at runtime, and
the new test is not in the component project's glob. The component suite's runtime input is
byte-identical to base.

ZacxDev:
## Audit rounds 2 and 3 — closed

Two more adversarial rounds since the last comment. Both found real defects, and **every one was
introduced by the preceding fix round**, not by the original change — which is the argument for
re-auditing the delta rather than stopping at the first clean-looking pass.

### Round 2 (fixed in `fe320ca9fd`)

- The test file had grown a local one-key mock of `~/server/logging/client` — the exact pattern
  `setup.ts` documents in a 🔴 note as having broken nine route tests once already. That module has
  7 exports; `endpoint-helpers.ts`, inside this endpoint's own import graph, statically imports two
  of the others. It was latent only because `WebhookEndpoint` does not route through
  `handleEndpointError` today. The file was already importing the canonical mock and then
  overriding it. Now asserts on `loggingMock.logToAxiom`, cleared per-test (that mock resets per
  *file*, so without the clear every "called once" assertion read the previous test's writes).
- `actorUserId` was `.optional()`, so `{"actorUserId": null}` — the natural JSON for "no acting
  moderator" — **400'd the request and the exclusion never happened**. An optional attribution field
  must never be able to fail the write it annotates.
- The 5s timeout had **zero coverage**: the mocked write resolves on a microtask, which always beats
  a macrotask timer, so setting the bound to `0` left the suite green. Pinned now with fake timers,
  from both sides.

### Round 3 (fixed in `08350486d4`, comment-only)

The round-2 comment spelled the forbidden call out *literally*, to warn readers off it. The guard
enforcing that rule (`no-direct-shared-module-mock.test.ts`, via the textual `mockPattern()` in
`mocks/guarded-specifiers.ts`) is a **regex over the file** and cannot tell a comment from code — so
writing the prohibition turned the guard red, with no visible cause: grep the file, find no such
call, see nothing wrong. Reworded without the literal syntax, and the trap is now documented in place.

🔴 **Worth flagging for everyone, not just this PR:** `pnpm test:lint-rules` runs a **curated list**
that includes `no-wholesale-module-mock.test.ts` but **not** `no-direct-shared-module-mock.test.ts`.
I ran it twice and reported green; that green was a true statement about the wrong population. The
failure was only visible in the full unit project, where it was the sole failure. If you rely on
`test:lint-rules` as the guard gate, it is not one.

Two further comment inaccuracies fixed in the same commit: the `actorUserId` block opened
`.nullish(), not .optional()` while the code uses `z.preprocess` (and `.nullish()` alone would not
have folded `''`, which the same comment required); and "a genuinely malformed value still 400s" was
overstated — `z.coerce` turns `true` into `1` and `[7]` into `7`.

### Current verification

| | |
|---|---|
| this file's tests | 22 (from 6) |
| **full unit project** | **1306 files / 20496 tests, 0 failures** |
| typecheck | 0 errors — instrument validated with a planted error that went red |
| prettier / eslint | clean, each shown to go red on a planted defect first |
| mutants | 12 run here, all killed; the round-3 auditor ran 19 (8 I had not tried), all accounted for |

Three of my own instruments lied during this work and are worth naming, since each initially read as
a real result: two "surviving" mutants were **invalid** (a `perl` replace without `/g` hit
`minReactions` instead of `actorUserId`; another hit a *type annotation*, erased at runtime); one
battery run was unreadable because the harness did not strip ANSI, so the positive control printed
nothing while every mutant appeared to pass; and a `pnpm install` reported **exit code 0** while
printing `command not found`.

PR #4256, opened 2026-08-21
fix(engagement): scope every engagement write by type, clear engagements on user delete, maintain the follow cache by delta

JustMaier:
Four tickets out of PR #4230, taken as one family at Justin's request on the 2026-08-21 standup: 868kurkc7, 868kurkcf, 868kurrfj, 868kurkd0.

## 868kurkc7 — five sibling engagement tables had the unqualified-write bug

Every engagement table is one row per (user, entity) carrying ONE `type`, so the types are mutually exclusive by construction. These writers read the row, branched on that read, then wrote addressed by the **primary key alone** — so the write landed on whatever type occupied the row, including one a sibling writer established a millisecond earlier, and reported success.

Unlike `UserEngagement` (Block > Hide > Follow), **none of these tables has a precedence order**, so the guard available is the type the call actually observed: a writer replaces what it saw, or nothing. `delete`/`update` on the PK become `deleteMany`/`updateMany` filtered by that type in:

| Writer | File |
|---|---|
| `toggleHideModel` | `user-preferences.service.ts` |
| `toggleHideModel3D` | `user-preferences.service.ts` |
| `toggleHideImage` | `user-preferences.service.ts` |
| `toggleHiddenTags` | `user-preferences.service.ts` |
| `toggleModelEngagement` | `user.service.ts` |
| `toggleModelVersionEngagement` | `model-version.service.ts` |

The tag toggle disagreed with itself — its `hidden === false` branch already filtered `type: 'Hide'`, the two statements inside the `else` did not. Its conversion is now issued once per **observed** type rather than once for the whole batch, because a batch legitimately holds both `Follow` and `Allow` rows and `type: { not: 'Hide' }` would protect neither.

`ModelVersionEngagementType` has exactly one member, so its write was not a bug today. It becomes one silently the day a second value is added, and nothing said so.

Two consequences worth calling out:

- **`toggleModelEngagement` now reports `count > 0`** rather than an unconditional `true`. Zero rows means a concurrent writer replaced the row it read, so the pair does not carry the requested type. Returning `true` there is the 868kumcfc shape — a claim the caller fires tracking on.
- **its conversion branch refreshed no cache.** `HiddenModels` is keyed on `type = 'Hide'`, so converting a Hide to Notify left the model filtered out of the user's own feed until the per-user hash aged out. It now refreshes when a conversion changes the hidden set in either direction, and not otherwise.

### Deliberately NOT fixed here — it is a product question

The **non-race** half stands: the schema holds one type per pair, so hiding a model you have `Notify` on still converts the subscription, and un-hiding does not bring it back. Closing that needs a precedence rule **per table** — or a schema change — and the ticket flags it as a product decision before it is a code one. Nothing in this PR invents one.

## 868kurkcf — deleteUser left every engagement behind

```
where: { OR: [{ userId: user.id, targetUserId: user.id }] }
```

One OR element carrying two fields is a single ANDed predicate — `userId = X AND targetUserId = X` — which matches only a self-engagement row. `deleteUser` **soft-deletes** the User row, so no FK cascade covered for it: every follow, hide and block belonging to or aimed at a deleted account survived and kept being counted by `getUserList`.

**This is the answer to the audit Justin asked for, and it contradicts the hunch behind it.** The belief was that follows on deleted users were already cleaned up. They are not, and there are two independent reasons — the predicate above never matched anything, *and* `getUserList` has no `deletedAt` filter on the joined user. So even a working `deleteUser` would have needed the historical rows cleared before the symptom went away. Please do not wave this through as already-fixed.

The visible symptom: a deleted account still appears in followers/following lists — with a null username, since `deleteUser` scrubs it — and still counts toward the total. `restoreUser`'s docblock has been asserting these rows are hard-deleted the whole time; after this PR it is true.

Fixed to two OR elements, plus a bust of the deleted user's own follow-set entry. Their **followers'** entries are deliberately left to their own TTL — a popular account has six figures of them, and each holds an id whose content this same call has already removed.

🔴 **`20260821120000_delete_engagements_of_deleted_users` must be applied by hand, off-peak.** Measured on the prod replica 2026-08-21: **42,187,643 rows, of which 3,080,712 (7.3%) reference one of 1,317,709 soft-deleted users** — 771,882 by a deleted user, 2,367,030 aimed at one, 58,201 both. The work list is built in ONE sequential scan (16.1 s, plan verified) and drained by primary key in 10,000-row batches, ~309 of them. Keyed by primary key rather than `ctid`, which VACUUM can hand to a different row. It ends with a positive verification query, because an exit code is not evidence here.

The loop mechanics — the temp table surviving the per-batch `COMMIT`, the two-`DELETE` CTE, the exit condition — are **reasoned about, not observed**. Whoever applies it should dry-run the loop first.

## 868kurrfj — the dead second writer is removed, not kept fixed

`toggleHideUserHandler` was registered on no router, and the `toggleHideUser` it called in `user.service` was imported only by it and by a test. Every hide in the product goes through `toggleHidden({ kind: 'user' })`.

Removed rather than left fixed, which is the resolution the reuse and intent lanes argued for and which 868kumcfc's phrasing had blocked: a second live writer on this table is the exact defect family #4230 exists to close, and the divergence that copy accumulated — a missing `HiddenUsers` invalidation — accumulated *because* no call path exercised it. Fixing it resets that clock without changing the property, and leaves it looking correct and ready for the next person to register.

**Its coverage is not deleted.** Every branch it pinned is asserted against `toggleHidden({ kind: 'user' })`, which the product actually reaches. One case existed only in its describe — hiding over a `Follow` — and has been added there. What does disappear is the *return-value* contract, because the function that returned it is gone; that is the honest closure of 868kumcfc rather than a fix nobody can exercise.

## 868kurkd0 — the follow set is maintained by delta, not refetched

Every follow / hide / block toggle ended in `userFollowsCache.refresh(userId)`, which refetches the user's **entire** follow set from the primary with no LIMIT: 19,224 buffers and ~100 ms for a user with 130k follows, per click, measured on prod. Three orders of magnitude above every statement in the toggle itself.

`createCachedArray` gains **`update(id, updater)`** — a locked write-through that rewrites one entry from a delta the caller already knows:

- **Never materialises an entry.** Absent, a `notFound`/`debounce` marker, past its expiry, or another writer holding the per-entry lock all return `false` and write nothing. A delta alone would produce a follow set containing exactly the one id the caller happened to touch.
- **Carries `cachedAt` over and keeps the remainder of the original expiry.** An entry maintained by deltas has to age out on its own schedule, or a drift in the delta logic never self-heals.
- **Locks.** Two concurrent read-modify-writes on one entry lose one of the deltas for the rest of the TTL; the loser reports `false` and takes the caller's fallback.

Per Justin's call, **the refetch stays as the fallback rather than being replaced** — no replica read, no longer TTL, no stale-while-revalidate. It reads the primary deliberately: a missing entry is otherwise repopulated by the next reader from the replica, which is the one read that can miss the write that just happened. Call sites that cannot know the resulting follow state — an un-hide or an unblock whose scoped delete matched no row, leaving a pair that may still hold a `Follow` — take that fallback rather than guessing.

## Verification

- `pnpm run lint` — 0 errors on the touched files
- `pnpm run test:lint-rules` — 265 passed
- `pnpm run test:unit:run` — **1310 files / 20534 tests passed**, 0 failures (`blocks.router.workflow.test.ts` collected 350, so the submodule is present and the run is real)
- **Mutation pass: 19 mutants, 19 caught.** Each mutation was applied with a string replace *and verified by re-reading the file* before the suite ran. They mutate the actual conditions — dropping a single `type` filter, inverting the follow-delta sign, restoring the original `OR` predicate, returning `true` instead of `count > 0`, resetting the entry to a full TTL, removing the lock, moving the lock release out of `finally`.

**The mutation pass was rerun after the fix below, and that rerun is the 19/19.** One mutant initially **survived**: removing `update()`'s marker guards left the suite green, because a marker entry has no `members` and the test's updater threw instead. The guard was not what stopped the write. The updater is now tolerant of a missing field, so the guard is what is under test — and the mutant is caught.

---

# Review round

Five review lanes over the first push. **Four confirmed defects, two of them introduced by this PR rather than found in main.** Detail in `40be007de4`; the parts that change how you should read the diff:

## The six fixes are backed by a source gate, because the defect IS a hand-enumeration bug

`no-pk-addressed-engagement-write` (registered in `test:lint-rules`) scans `src/` for `*Engagement.delete/update` addressed by a compound key. Six hand-written per-table assertions catch exactly the six tables that already exist; the gate catches the seventh.

**It found the seventh immediately.** `toggleComicEngagement` (`comics.router.ts`) was a PK-addressed `update` on `ComicProjectEngagement` (None | Notify | Hide) — and it is *worse* than its siblings, because it branches on a read off the **replica**, so its race window is replication lag rather than microseconds. Scoped and reporting honestly now. `BountyEngagement` and `ChallengeEngagement` are exempt with the reason recorded: `type` is in their key, which is what makes their PK-addressed writes safe.

That is the difference between fixing six things and closing the shape.

## deleteUser was destroying other people's blocks

The predicate fix widened deletion to every row aimed at the account — and account deletion here is **reversible**. `restoreUser` brings the account, its models and its images back and restores **no** engagements. So: B blocks D, D is deleted, D is restored, and **B no longer blocks D, with nothing telling B.** A safety control someone set, switched off by a third party's action.

Blocks are now exempt in both directions, in the service and in the migration. Follow and Hide are list noise; a Block is not. 42,121 Blocks stay standing out of the 3,080,712 rows in scope. Reversible in one predicate each if that is the wrong call.

## A wrong cache entry, not a stale one (mine)

`toggleFollowUser`'s Hide→Follow branch wrote `following: count > 0`. Zero rows does **not** mean "not following" — the Hide is gone, and one thing that can hold the pair instead is a Follow established by a second concurrent toggle that won the same `updateMany`. The loser then wrote `false`: an entry that is WRONG for the day-long TTL, where the refetch merely costs. `setUserFollowCached`'s own contract forbids exactly that, and the two sibling call sites already got it right. Same shape behind `toggleBlockUser`'s P2002 catch.

## The dropped-intent bug was still live for three kinds

`toggleHidden` threads `setTo` for user / blockedUser / tag and dropped `hidden` on the floor for image / model / model3d — so asked to un-hide a pair with no row, those three **created a Hide**. That is 868kun67j, fixed for users only. Threaded through now. The new scoping suite had a `hide(kind, hidden?)` helper it never passed `hidden` to, which is what let it read as covering them.

## Smaller, from the lanes

- `update()` uses `KEEPTTL` instead of recomputing the remainder from `cachedAt` — the arithmetic assumed a full expiry set at that instant, which `invalidate` deliberately breaks by backdating it, so the first SWR caller would have had a live entry shrunk to about `debounceTime`. `cache-helpers.ts:414` already does it the right way.
- `setUserFollowCached` applies its delta with `indexOf` rather than rebuilding a `Set`: **13.9 ms → 1.3 ms** on the real 130k-follow account, against 3.8 ms of decode and encode. The PK already makes the array duplicate-free.
- The dead `toggleModelHide` + `toggleHideModelHandler` pair goes for the same reason `toggleHideUser` did — registered on no router, one screen from the twin this PR deleted.
- `image.service.ts` carried load-bearing reasoning naming `userFollowsCache.refresh` as what `toggleFollowUser` runs. It is the fallback now, not the path.

## Migration, rehearsed

**3,038,591 rows in 304 batches of 10,000**, leaving 42,121 Blocks. Work list built in one scan (16.1 s, plan verified); drained by **key**, so each batch is an index range read rather than a fresh scan over accumulating dead tuples. Ends with `VACUUM (ANALYZE)` and a positive verification query.

Rehearsed on dev scoped to 200 deleted users: **711 rows, 4 batches, cursor advancing each time, temp table surviving every COMMIT, post-check 711 → 0.**

🔴 It must be run as `psql -f` on a direct, autocommit session. Feeding the file through anything that wraps it in a transaction returns `invalid transaction termination` — reproduced, not theorised.

## Mutation

**35 mutants across two rounds, 35 caught**, each verified to have landed by re-reading the file. Round two had one survivor (no test for the block-P2002 fallback) and one mutant that did not land — an *insertion* mutant, where the landing check looked for the original text being gone. Both fixed and re-run; the 35/35 is that re-run.

## On verifying this branch — two traps worth carrying forward

🔴 **A queued suite reported exit 0 while its log ended `ELIFECYCLE Command failed with exit code 1` / `Run failed`.** Reading the exit code rather than the log file would have called a red suite green. Read the file.

🔴 **Only `test:unit:run` is serialised by the dev-server queue.** `pnpm exec vitest` directly, `test:component`, `test:packages:run` and `test:apps:run` are all unqueued, so a queued run and an unqueued one overlap happily. Three full-suite runs here failed with three *different* files, all `Cannot access '__vite_ssr_import_7__' before initialization` — a module-init-order artifact, not an assertion — because another worktree was running an uncapped suite on the same box. Confirmed from the process list, not inferred from the symptom.

Those two together are the nasty combination: **a flaky red, reported as a pass.**

And the reason it mattered here rather than being a curiosity: one of those three failures was **not** flake. `flipt-eval-context` is a source gate that ledgers its known sites by `FILE:LINE`, and a four-line comment where three stood shifted every ledgered `image.service.ts` row by one. It was deterministic, and the contention noise is what made it look like more of the same. It only separated out by running each failing file **in isolation**.

## What the setTo threading does NOT change

No caller in `src/` passes `hidden` for `kind: 'model' | 'model3d' | 'image'` — `HideModelButton`, `HideModel3DButton` and `HideImageButton` all omit it. So for today's UI the behaviour is byte-identical (setTo ?? flip, with setTo undefined, is the old blind toggle). What it closes is the API boundary: the tRPC input schema accepts `hidden`, and until now those three kinds silently ignored it.

One more property of that gate, worth knowing before the next person edits `image.service.ts`: **it ledgers by `FILE:LINE`, so ANY edit that changes a line count breaks it — comments included.** Keeping a comment line-neutral is the right local fix; the alternative is re-numbering every ledgered row in a file you did not otherwise touch.

---

---

# Migration: APPLIED to prod, in two passes

Approved by Justin — the deletion policy, the window, me as the operator, and the shape change after the first pass failed.

| | |
|---|---|
| Originally outstanding | 3,038,591 |
| Pass 1 (`DO`-block shape) | 920,001, then `FATAL: query timeout` |
| Pass 2 (client-driven) | **2,118,588 in 212 batches of 10,000, 157.2 s** |
| **Non-Block engagements still referencing a deleted user** | **0** |
| Blocks preserved | **42,101** |

Zero reconnects, zero errors in pass 2. Verified from a **separate connection** afterwards, and from SQL rather than the site — nothing here invalidates the app's caches, so a stale `userFollowsCache` entry can still show a deleted account in a followers list for up to a day. `UserEngagement` is now 39,152,562 rows; no stray work table; `VACUUM (ANALYZE)` ran.

Small drifts against the predictions (3,038,589 removed vs 3,038,591; 42,101 Blocks vs 42,121) are live traffic between measurements. A before-count matching to the digit on a live 42M-row table would be the surprising result.

## Two design defects it exposed, both fixed in the committed file

**1. A `DO $$ … $$` block is ONE query to the server.** Its internal per-batch `COMMIT`s make work durable but do not reset the front end's query clock, so the drain died at 920,001 rows. `statement_timeout` is `0` on this database — the ceiling is enforced *in front* of it, and a role without a statement timeout would not have helped. The loop is client-driven now: one batch per statement, measured at **127–188 ms** each.

**2. A `TEMP` work list does not survive a lost session.** Found on dev: the connection went away mid-drain and took the list and all its `done` marks with it. `UNLOGGED` now — a dropped connection costs a reconnect and nothing else.

Both are the same class, and the generalisation is the useful part:

🔴 **A long job must not depend on a property of the session surviving longer than the work takes.** The `DO` block depended on one long query; the `TEMP` table depended on one long connection. Fixing the first did not surface the second — the file's own header still said "steps 2-5 must share a SESSION" as though that were free.

🔴 **And the rehearsals that passed were too small to reach either failure.** 711 rows in 4 batches proved the loop and proved nothing about a run lasting minutes. What actually derisked prod was a *full* end-to-end dev run — 2,476,490 rows, 248 batches, 164.6 s, after-count 0 — which should have come before the first attempt, not after it.

## 🔴 Exit code 0 over a FATAL, twice in one evening

The suite reported **exit 0** while its log ended `ELIFECYCLE Command failed with exit code 1`. Then the first migration pass reported **exit 0** while its log ended `FATAL: query timeout` / `connection to server was lost`, at 30% applied.

A suite lying costs a re-run. **A prod delete reporting "done" at 30% is a different category.** The only thing that caught it was reading the log and then counting rows.

**Never the exit code. Always the count from the database.**

One more, learned the hard way: the safeguard this PR is proudest of — the runner executes `git show HEAD:…migration.sql` rather than a copy, so "approved" and "ran" cannot diverge — had a **bug in its own step parser** within an hour of being written, and would have executed nothing. It failed safe by luck of ordering, not by design. Enforcement is only as good as the enforcing code.

---

# Two things this PR knowingly does not fix

Both raised by review lanes, neither smuggled past.

**1. `toggleModelEngagement`'s two branches disagree about the same question.** The convert branch returns `count > 0`; the delete branch two lines above still returns `false` unconditionally with no count check. The boolean feeds `toggleNotifyModelHandler`'s tracking split, so on a lost race it emits a ClickHouse `Delete` for an engagement nothing deleted. The *state* claim is right; the event name is the imprecise part, and that boolean→event mapping predates this PR.

**2. The scope boundary on 868kurkc7 is fuzzier than stated.** The new `HiddenModels.refreshCache` on the conversion branch *is* a non-race, product-visible fix to the Hide↔Notify path. It is a cache-correctness fix rather than a precedence decision, which is why it is in — but the boundary should be said out loud rather than implied away. The precedence question itself is tracked as `868kv1buq`.

**And one residual on the audit answer:** the fix is entirely write-side. `deleteUser` is the only thing in `src/` or `apps/` that sets `User.deletedAt` today, so it holds — but `getUserList` still has no `deletedAt` filter on the joined user, and any future path that soft-deletes without going through `deleteUser` reopens the symptom silently. Not added here: it changes list counts and pagination, and belongs in its own change.

---

# Rebased onto a moved `main`

11 commits replayed onto `origin/main`, which had advanced 18 commits. One conflict, and it is the kind worth naming because a conflict tool nudges you toward getting it wrong:

**`package.json` / `test:lint-rules`.** `main` added `track.addView.schema.test.ts` to the guard list; this branch added `no-pk-addressed-engagement-write.test.ts`. **Both sides ADDED a guard — neither replaced one — so keeping both is the only correct resolution.** Taking either side silently drops a convention guard, which is precisely the failure mode guards exist to prevent. 11 guards in the list now, `test:lint-rules` green at 281 tests.

`main` also bumped **vitest 4.0.18 → 4.1.11**, so the pre-rebase suite result is stale evidence — different base, different runner minor — and is not quoted as this branch's verification.

## 🔴 `main` is currently broken, and it blocks every PR

`src/server/services/user-payment-configuration.service.ts` imports `TipaltiStatus` type-only and uses it as a value (plus a local const shadowing an imported helper). Three typecheck errors, introduced by `3a1ad0c417`, tracked as `868kv37v9`.

Full unit suite on the rebased base, vitest 4.1.11: **1316 files passed, 20,619 tests passed, 1 file / 5 tests failed** — every one of the five with `ReferenceError: TipaltiStatus is not defined`, i.e. that single inherited bug and nothing else. Identified by CAUSE rather than by filename, deliberately: this same file failed earlier tonight for an unrelated reason.

CI evaluates the **merge result**, so no branch targeting `main` can go green until that lands. Established as inherited rather than assumed, three ways: the file is absent from this diff, no commit on this branch touches it, and it arrived with the rebase.

**Recommendation: hold this PR until `868kv37v9` lands, then merge on a green run.** Everything this PR owns is finished; the migration is already applied, so the cost of waiting is zero.

## One diagnostic note worth carrying

`user-payment-configuration.tipalti-status.test.ts` failed **twice tonight for completely unrelated reasons** — first in the contention noise (where it passed in isolation, because that run predated `3a1ad0c417`), then deterministically from main's broken import hours later.

**Re-diagnose per occurrence. Never carry a previous verdict forward on a filename.** Matching on the name would have written the second, real failure off as flake.

---

# 🔴 BLOCKED on a semantic merge with PR #4252 — read this before rebasing

`main`'s typecheck break (`868kv37v9`) is fixed and verified. The rebase now stops on something else.

**PR #4252** (`6c41b30fb2`, Manuel Emilio Urena — *stop offering Hide on a blocked user, and reconcile a refused hide*) edits `toggleHideModel`, `toggleHideModel3D`, `toggleHideImage`, `toggleBlockUser` and `toggle-hide-user.service.test.ts` — the same functions this PR rewrites. **8 conflict hunks, semantic rather than textual.**

**They are complementary, not duplicate, which is precisely why the conflict must not be resolved by taking a side:**

| | #4252 (on `main`) | This PR |
|---|---|---|
| Intent-honouring (`setTo`) | ✅ landed, independently | ✅ (now duplicate work) |
| UI gate for hiding a blocked user | ✅ | — |
| Type-scoped writes (868kurkc7) | ❌ still `delete`/`update` by PK | ✅ |

Taking "ours" drops his UI gate and intent handling. Taking "theirs" drops the type scoping — the entire subject of `868kurkc7`. Neither side supersedes the other.

## Resolution recipe

Keep **#4252's branch structure and comments**, and replace only its PK-addressed statements:

- `modelEngagement.delete({ where: { userId_modelId: { userId, modelId } } })` → `deleteMany({ where: { userId, modelId, type: 'Hide' } })`
- `modelEngagement.update({ where: { userId_modelId: { … } }, data: { type: 'Hide' } })` → `updateMany({ where: { userId, modelId, type: engagement.type }, data: { type: 'Hide' } })`
- Same two substitutions for `model3DEngagement` and `imageEngagement`.
- `toggleBlockUser`: reconcile #4252's `hidden` with this PR's `blocked` — both exist for different reasons, so neither may overwrite the other.

Then: covering suites, a **mutation pass over the merged toggles specifically** (they are the functions this PR's own fix round manufactured two defects in), and the full suite on the new base.

`main` also independently removed the dead `toggleHideModelHandler` / `toggleModelHide` / `toggleHideUserHandler`, so that part of this diff is a no-op — resolve by keeping main's deletion.

## Coordination finding

#4252 and these four tickets were solving **adjacent halves of the same defect with neither side knowing**. Worth saying out loud so the type-scoping half actually lands rather than being assumed done because the intent half shipped.


JustMaier:
## Not mergeable as-is — PR #4252 landed on the same functions

`main` is fixed (`868kv37v9`), verified independently. But the rebase onto it stops on **PR #4252** (`6c41b30fb2`, Manuel Emilio Urena — *stop offering Hide on a blocked user, and reconcile a refused hide*), which edits `toggleHideModel`, `toggleHideModel3D`, `toggleHideImage`, `toggleBlockUser` and `toggle-hide-user.service.test.ts`. 8 conflict hunks, semantic rather than textual.

**The two are complementary, which is exactly why the conflict must not be resolved by taking a side:**

- #4252 landed the **intent-honouring** half — `const hiding = setTo ?? engagement?.type !== 'Hide'` — arrived at independently, for the same reason. That part of this PR is now duplicate work.
- #4252 did **not** land the **type-scoping** half. On `main` today those toggles still write `modelEngagement.delete({ where: { userId_modelId: { userId, modelId } } })` and `.update({ … })` — PK-addressed, which is the whole subject of 868kurkc7.

Taking "ours" drops his UI gate and intent handling; taking "theirs" drops the type scoping. Neither supersedes the other.

**Resolution recipe**, for whoever rebases: keep #4252's branch structure and comments, replace only its PK-addressed statements — `delete({ where: { userId_modelId } })` → `deleteMany({ where: { userId, modelId, type: 'Hide' } })`, `update({ where: { userId_modelId } })` → `updateMany({ where: { userId, modelId, type: engagement.type } })`; same for model3D and image. `toggleBlockUser` needs #4252's `hidden` and this PR's `blocked` reconciled rather than one overwriting the other. Then re-run the covering suites, a mutation pass over the merged toggles specifically, and the full suite on the new base.

Also note: `main` independently removed the dead `toggleHideModelHandler` / `toggleModelHide` / `toggleHideUserHandler`, so that part of this diff is now a no-op.

**Deliberately not resolved tonight.** This is a semantic merge into a colleague's work on the exact functions where this PR's own fix round manufactured two defects. It wants a clear head, and nothing is waiting on it — the migration is applied and verified, so the user-visible half of 868kurkcf is already delivered.

JustMaier:
## Rebased onto `main` — #4252 combined, not superseded

`main` moved under this branch: **#4252** (`6c41b30fb2`) landed the *intent-honouring* half of the
same defect family — `const hiding = setTo ?? engagement?.type !== 'Hide'`, a `hidden` field on
`HiddenPreferencesDiff` so the client's optimistic write can be corrected, and the UI gate that stops
offering Hide on a blocked user. This branch carries the *type-scoping* half: every engagement write
addressed by `(user, entity, type)` instead of the PK.

Neither side is redundant, so the rebase kept both. How each conflict was resolved:

- **`toggleHideModel` / `toggleHideModel3D` / `toggleHideImage`** — #4252's structure (`setTo`,
  `hiding`, `hidden: hiding`) with its PK-addressed `delete`/`update` replaced by `deleteMany`/
  `updateMany` scoped by the type the call read. The un-hide issues its `type: 'Hide'` delete
  unconditionally rather than gating on the row it read: the read can already be stale, and the
  filter is what makes that not matter.
- **`toggleHideUser`** — #4252's `hidden` (returned to the client) and this branch's follow-cache
  delta maintenance are independent; both kept.
- **`toggleModelEngagement`** — this branch's `count`-based return (`count > 0`) subsumes #4252's
  unconditional `return true` alongside its both-directions `HiddenModels` refresh.
- **`toggleHideModelHandler` / `toggleModelHide` / `toggleHideUserHandler`** — `main` deleted these
  independently, so `main`'s deletion stands and that part of this diff is now a no-op.
- **`package.json`** — both sides add a guard to `test:lint-rules`; both kept (11 guards).
- **Tests** — #4252's `toggle-hidden-set-to.service.test.ts` matrix and its three
  `toggleModelEngagement` cache cases asserted the PK-addressed `update`/`delete`. Re-pinned on the
  scoped `updateMany`/`deleteMany`, including the `type` filter itself, since that filter is the
  protection. Its optimistic-UI describe (`868kurj0y`) is kept verbatim.

### Verification on the rebased tree

- `pnpm run lint` — 0 new errors (3 pre-existing in `image.service.ts`, unchanged on `main`)
- `pnpm run test:lint-rules` — 11 files, 281 tests, green
- `pnpm run test:unit:run` — **1323 files, 20697 passed, 0 failed** (`blocks.router.workflow.test.ts`
  collects 350, so the submodule is present and the run is meaningful)
- **Mutation pass over the merged toggles: 16 mutants, 16 caught**, each with a named assertion
  failure — dropped `type` filters on all six entity writes, a dropped `setTo`, `hidden` reported
  without asking the write, the follow-set delta guessed instead of re-derived (hide and block
  paths), and both `count > 0` returns replaced with `true`.

The only edit to `image.service.ts` is a comment, kept line-neutral for the `flipt-eval-context`
FILE:LINE gate.

PR #4257, opened 2026-08-21
fix(search): hoist useModelSaleBadges above the early returns — React #310 crashing /search/models

ZacxDev:
## Problem

`#4112` added `useModelSaleBadges` to `ModelsHitList` in `src/pages/search/models.tsx`, but placed the call **after** three early returns:

| Line | Early return |
|---|---|
| 144 | `if (hits.length === 0)` |
| 167 | `if (loading)` |
| 187 | `if (loadingPreferences)` |
| **197** | 🔴 `const salesByModelId = useModelSaleBadges(...)` |

So the hook runs on the loaded render and **not** on the loading render. The hook count changes between renders and React throws **error #310** — *"Rendered more hooks than during the previous render"* — which takes down the model-search page.

Every model search transitions `loading → loaded`, so this fires on essentially every query. Production client error reporting went from a **4–191 per 30m** baseline to **14,665 per 30m** (~7–10/s sustained) immediately after 5.1.32 rolled out; 99.9% of all reported application errors in that window are this single stack:

```
Minified React error #310
    at L (src/pages/search/models.tsx:121:67)
    at ev.Content (src/components/Search/SearchLayout.tsx:418:50)
    ...
```

Server-side error rates are unaffected (5xx stayed at ~0.01%) — the server returns 200 and the browser then crashes, so this is invisible to server-side error metrics.

## Fix

A pure statement hoist — move the hook above every early return.

`items` already comes from `useApplyHiddenPreferences` at the top of the component (line 128), above all three returns, and the hook self-disables on an empty list (`enabled: ids.length > 0`), so calling it earlier is safe and changes no behaviour on the loaded path.

A comment records the invariant so the next edit doesn't reintroduce it.

## Verification

Ran the rule that describes this exact defect against the pre-fix and post-fix files:

```
$ eslint --rule '{"react-hooks/rules-of-hooks":"error"}' <pre-fix> <post-fix>

<pre-fix>
  197:26  error  React Hook "useModelSaleBadges" is called conditionally.
                 React Hooks must be called in the exact same order in every
                 component render  react-hooks/rules-of-hooks

✖ 1 problem (1 error, 0 warnings)
```

Post-fix file: clean. So the control goes red for this rule's own reason, and the fix clears it.

**Not verified**: I have not exercised the page in a browser. The change is a statement move with no behavioural delta on the loaded path, but a reviewer with a dev environment may want to confirm the sale badges still render on `/search/models`.

## Other call sites are fine

The two other `useModelSaleBadges` call sites added by #4112 already sit above their early returns:

| Call site | Position | Verdict |
|---|---|---|
| `src/components/Model/Infinite/ModelsInfinite.tsx:52` | before any return | ✅ |
| `src/components/Profile/Sections/OnSaleSection.tsx:36` | before `if (isNullState) return null` (line 40) | ✅ |
| `src/pages/search/models.tsx:197` | after 3 early returns | 🔴 fixed here |

## Follow-up (not in this PR): why CI didn't block it

`react-hooks/rules-of-hooks` is enabled (via `next/core-web-vitals`) and, as shown above, catches this. It didn't block #4112 because `.github/workflows/lint.yml` gates ESLint asymmetrically:

- **added** files → blocking
- **modified** files → `continue-on-error: true`, report-only

`models.tsx` was a modified file, so the error was printed as a non-blocking annotation and merged. The stated rationale is sound (≈10% of recently-touched files carry a pre-existing error-severity finding, so a hard gate would block unrelated work), but the consequence is that a **newly introduced** error is indistinguishable from the pre-existing backlog.

Worth considering: make the modified-file step a **delta** gate — compute findings at the merge base and at HEAD, and fail only on findings the PR *introduces*, leaving pre-existing ones as warnings. That keeps the backlog non-blocking while making a regression like this one impossible to merge. Happy to open that separately if there's appetite.

### DIFF
--- a/src/pages/search/models.tsx
+++ b/src/pages/search/models.tsx
@@ -141,6 +141,10 @@ export function ModelsHitList() {
     }
   }, [modelId, status, showMore, isLastPage, hits]);
 
+  // Must stay above every early return below — a hook called after one changes the hook
+  // count between the loading and loaded renders (React error #310).
+  const salesByModelId = useModelSaleBadges((items as { id: number }[]).map((x) => x.id));
+
   if (hits.length === 0) {
     const NotFound = (
       <div className="flex items-center justify-center">
@@ -193,8 +197,6 @@ export function ModelsHitList() {
       </div>
     );
   }
-
-  const salesByModelId = useModelSaleBadges((items as { id: number }[]).map((x) => x.id));
 
   return (
     <Stack>


ZacxDev:
Correcting one thing in my own description above, and recording two adjacent findings from a pre-merge audit.

### Correction: the ESLint evidence is narrower than I wrote it

I wrote "Post-fix file: clean." That is true **only of the single rule I passed via `--rule`**. Re-run under the full repo `.eslintrc.js`, the file reports:

- **base**: 2 errors — `react-hooks/rules-of-hooks` at `197:26` **and** a pre-existing `@typescript-eslint/no-explicit-any`
- **head**: 1 error — the `no-explicit-any`, at `213:26` after the hoist shifts line numbers

So the fix clears the hook error under the real repo config, which is the stronger version of the claim. But the file is not "clean" and will still render red in the report-only modified-files step. The `no-explicit-any` is pre-existing and untouched here; I'm leaving it rather than widening a crash hotfix.

### Two adjacent issues — neither introduced by this PR, neither blocking it

**1. `getActiveSales` is capped at 500 ids, and this page can exceed it.**
`src/server/routers/model.router.ts:206` bounds the input at `z.number().array().max(500)`. `SearchLayout.tsx:228` sets `hitsPerPage={50}`, and `items` accumulates across `showMore`, so from roughly the 10th page the tRPC input is rejected and sale badges stop updating for the rest of the session. `models.tsx:133-142` amplifies this: a `?model=<id>` deep link force-calls `showMore()` on every idle render until the model is found, which can reach 500 with no user scrolling.

Unchanged by this PR — the same `ids` array is passed either way — so it is not a reason to hold the fix. I verified the cap and the page size by reading; I have **not** executed the overflow, so the exact user-visible outcome (badges freezing at the last successful set versus disappearing) is unconfirmed.

**2. The same hook-ordering defect class is live elsewhere.**
`src/components/Model3D/Actions/Model3DActionsMenu.tsx:120` has `if (!isOwner && !isModerator && !canReport && !canHideUser) return null;` sitting above 8 hooks (`:143`-`:198`). Every disjunct derives from `useCurrentUser()`, which returns `null` until the session resolves — so an unauthed→authed flip while mounted changes the hook count and throws the same #310. ESLint does not flag most of these because it does not recognise a deep member expression like `trpc.model3d.publish.useMutation` as a hook, which is worth knowing when reasoning about how much cover the rule actually gives.

Out of scope here; flagging so this PR isn't read as closing the class.

### On tests

No test accompanies this fix. Worth stating the gap precisely: the `enabled: ids.length > 0` guard in `useModelSaleBadges` has **no test that any mutant of it would fail** — operand inversion, off-by-one, or commenting it out all survive the current suite. The repo already has both patterns that would cover this (`*.browser.test.tsx` component tests, and the structural `no-*.test.ts` guards wired into `pnpm test:lint-rules`), and a "`rules-of-hooks` violations must not increase" ratchet looks close to free today — a repo-wide sweep found only one remaining production-code violation, the `Model3DActionsMenu` one above.

I'd rather land the crash fix first and do that as a follow-up than grow this diff.

PR #4264, opened 2026-08-21
feat(stickers): Sticker Book profile tab

JustMaier:
## Sticker Book — a profile tab for a creator's sticker activity

Closes the "sticker book" ticket (`868kum420`). Designed live by Justin and Ellie on 2026-08-21; built from that session's decisions, then reworked once on Justin's review of the running page.

### What it is

`/user/<username>/sticker-book`, a new profile tab carrying:

- **the stickers the creator owns** — the placement tray's grid rather than the account page's row-per-item list, in a panel, with the name, creator and shop link behind a hover
- **images they stickered** and **their images that got stickered** — one card per image, drawn with the standard feed `ImagesCard`, with the sticker on it
- **Buzz earned from stickers**, taken from the ledger rather than from what placers paid
- a gear that opens the placement settings as a **modal**, rather than a settings card pinned to the top of the page
- a **View all** page per section, paged

### Two visibility toggles, and what each protects

Both are stored as opt-outs on `user.settings`, so an account that has never opened the modal has a visible book:

| | hidden from | shown to |
|---|---|---|
| `hideStickerBook` | everyone; the tab disappears from the profile nav | the owner, moderators |
| `hidePurchasedStickers` | the sticker grid only | the owner, moderators |
| how many uses they hold | **everyone, in both settings** | the owner alone |
| Buzz earned | everyone | the owner, moderators |

Every one of those is decided in the service and the withheld halves are never serialised — the page reads `access` to choose what to *draw*, never to choose what to withhold.

### Where the rules live

The book decides **which** images and in what order. Whether an image may be **shown** is `getAllImages`' job — browsing level, domain ceiling, publish state, moderation flags, the blocked-browsing-tag policy — so there is no second copy of those rules here. Blocks are applied server-side in both directions, on the end of each placement that is *not* the profile owner.

Only `approved` placements appear, in both directions: a pending placement is a request the target creator has not answered, and publishing one on a public profile would be the feature working in reverse.

### Notable during review

- **The reveal store defaults to off**, so the count chip on each card is the reveal control (`StickerPlacementCardBadge`) rather than a badge of our own — otherwise a page about stickers draws none of them with nothing on it to turn them on.
- **`getAllImages`' `userId` is its author filter**, not a second way to name the viewer. Passing the viewer there emptied "images you stickered" entirely, since those are by definition somebody else's. Pinned by a test.
- Two moderation gates the placement image filter did not carry — `needsReview` and `acceptableMinor` — are now in the shared selector, so the review queue gets them too. Scheduled posts are excluded as well (`publishedAt <= now`, not merely non-null).

### Deliberately not here

- **Placement history**, and the link to it. Justin: "let's leave the link out, let's just not do the placement history portion yet."
- **A count badge on the tab.** It needs a write-through into the cached profile overview; Justin chose to skip it.
- **Hidden users and hidden tags** are not applied to these sections. Blocks are. Stated in the code rather than assumed, and with Justin as an open question.

### Testing

`sticker-book.service.test.ts` (25) and `sticker-book.test.ts` (11). Eleven mutations were run against the guards — quantities to moderators, `remaining` always sent, the block filter on the wrong end and removed entirely, the hidden-book early return removed, the drill-in's own gate removed, the viewer band dropped, earnings ungated, the two moderation gates, the author-filter bug — and every one went red. One earlier mutation **survived**: the block test asserted two independent `objectContaining`s, so swapping the sides satisfied both while destroying the identity scope. It now asserts each side as a pair, and the service builds its `where` per side longhand so a computed key cannot overwrite the id that scopes it.

Full unit suite green; `blocks.router.workflow.test.ts` collected 350.

### Migrations

None.


JustMaier:
**Staged review data lives on the dev database only** (all placements `free`, `amount 0`, so no Buzz moved). Recording the cleanup here rather than in a scratchpad, so it outlives the session that made it:

```sql
-- sample data for the sticker book (122-134) and the four-edge repro for #4266 (135-138)
DELETE FROM "Placement" WHERE id BETWEEN 122 AND 138;
-- the hidden-book / moderator-override case
UPDATE "User" SET settings = settings - 'hideStickerBook' - 'hidePurchasedStickers' WHERE id = 12042163;
```

Nothing in `main` writes either settings key, so removing them restores the prior state exactly.

⚠️ 135-138 are **pending** placements and lapse around **2026-08-23 02:37 UTC**. After that the approve/decline controls disappear because the placements expired, which looks like a regression in #4266 and is not one.

maps524:
**Feature flag added and live.** `sticker-book` is merged into `civitai/flipt-state` (`5a34c7c`) and has synced:

```
Flag: sticker-book
Status: DISABLED
Default: false — off for everyone no rollout matches
Rollout Rules:
  - Segment: testers OR moderators → true
```

Deliberately `enabled: false` with the segment rollout rather than `enabled: true`: on a boolean flag the top-level value is what everyone gets when no rollout matches, so `true` plus a testers segment is a public launch with a redundant rule attached.

In the app it gates **two** things — the profile nav tab (`ProfileNavigation`) and the route itself (`getServerSideProps` redirects when the flag is off). A gate that only removes the link is not a gate; the URL is guessable.

The book's tRPC procedures stay ungated on purpose, so a creator who was paid for a placement does not lose the record of it if the flag goes back to testers.

PR #4266, opened 2026-08-21
fix(stickers): keep the approve/decline controls inside the image

JustMaier:
## Approve/decline unreachable when a sticker sits low on an image

Fixes `868kut2bd` (reported by Demian, 2026-08-20).

### The diagnosis, from measurements rather than from reading

The ticket says the controls are anchored to the sticker and overflow the container. Justin says they are truncated to fit inside the image. **Both describe the same defect, in sequence** — so this is one fix, not a choice between two.

`StickerPlacementOverlay` positions the pending controls at `left: x%` (with a half-width transform) and `top: calc(y% + halfStickerHeight + 14px)`, with **no clamp**, inside a layer that is `overflow-hidden`. The anchor puts them outside; the container then truncates them.

Four pending placements at the four extremes, measured in a browser against the clip box (`196,60 → 854,1046`):

| placement | outside the box by | after |
|---|---|---|
| bottom, `y = 0.95` | **+59px below** — both buttons gone | −136 (flipped above) |
| top, `y = 0.05` | inside | inside |
| left, `x = 0.03` | **+42px past the left edge** — Approve half cut | 0 |
| right, `x = 0.97` | **+42px past the right edge** — Decline half cut | −1 |

**Three edges, not one.** A fix verified only against the reported bottom edge would have shipped with the sides still cut.

### The fix

`placementControlPosition` — a pure function returning `{ left, top }` in px:

- below the sticker stays preferred; it flips **above** only when below would leave the box, because a control laid over the sticker covers the thing being judged
- both axes clamped into the box
- returns `null` until the control and the box have been measured, so the current position stands rather than the controls jumping from `0,0`

The component measures the control and the media box the way it already measures sticker height — `offsetWidth`/`offsetHeight` (layout numbers, so the reveal animation's transform does not corrupt them) with a `ResizeObserver` and a no-observer fallback.

`overflow-hidden` stays. Removing it would spill the owner's buttons over whatever sits below the image.

### Tests

`placement-control-position.test.ts`, nine cases, using the measured numbers above. Four mutations of the clamp's **arithmetic** — not its existence — all go red: clamping `top` against the box's width, dropping the gap, inverting the flip threshold, and clamping `left` to the box width rather than to `width - control.width`. The both-directions-blocked branch and the control-larger-than-the-box branch have their own cases.

Geometry rather than a rendered box, deliberately: the numbers in the table are what the fix has to change, and a browser test of this runs in no CI job.

### Not fixed here, and worth knowing

The same reported symptom has a **second, independent cause**: `useStickerRevealStore` defaults to off, so pending placements — and the owner's only in-image control with them — do not render at all until the viewer turns stickers on. An owner arriving from a placement notification sees an image that looks untouched. That is `868kv2hpp`, filed separately; fixing the geometry alone will not make the controls appear for someone who has never revealed stickers.

### Note on CI

`main` is currently broken (`868kv37v9`, `TipaltiStatus` imported as a type and used as a value), so this branch inherits three type errors and a red unit check in `user-payment-configuration.tipalti-status.test.ts` that are not from this diff.

### Migrations

None.

### DIFF
--- a/src/components/Sticker/StickerPlacementOverlay.tsx
+++ b/src/components/Sticker/StickerPlacementOverlay.tsx
@@ -5,6 +5,7 @@ import { useStickerCosmetics } from '~/components/Sticker/sticker.util';
 import type { PlacedSticker } from '~/components/Sticker/placement.util';
 import { orderPlacements, placementRevealDelays } from '~/components/Sticker/placement-order';
 import { stickerArtworkStyle } from '~/components/Sticker/placement-appearance';
+import { placementControlPosition } from '~/components/Sticker/placement-control-position';
 import { StickerPlacementActions } from '~/components/Sticker/StickerPlacementActions';
 import { StickerPlacementHoverCard } from '~/components/Sticker/StickerPlacementHoverCard';
 import { PlacementFreeBadge } from '~/components/Placement/PlacementFreeBadge';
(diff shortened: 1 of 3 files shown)


JustMaier:
**Reproduction is staged on the dev database** (free placements, no Buzz moved): four pending sticker placements on image `45391881`, one at each extreme — `y=0.95`, `y=0.05`, `x=0.03`, `x=0.97`.

⚠️ They **expire around 2026-08-23 02:37 UTC**. After that the controls legitimately disappear because the placements lapsed, which looks exactly like the fix having regressed. If you are looking at this on Saturday, re-stage rather than believing the page.

⚠️ Also: stickers are hidden by default site-wide, so press the sticker count chip on the image before expecting to see any placement at all. That is the second, separate cause noted in the description (`868kv2hpp`).

PR #4270, opened 2026-08-21
feat(monetization): replace tier price caps with a monthly pricing allowance

bkdiehl:
Membership stops governing how much a creator may charge and starts governing how often they may put a new price on something — article 33749, plus the decisions recorded in docs/licensing-fee-revamp-questions.md.

- Price caps deleted. The licensing fee ceiling is a flat 100/generation for every creator (500 on a video model); paid access has no ceiling at all. Every clamp goes with them, including the cap-tier Redis cache — getViewerMonetization, the purchase path and the orchestrator-facing mini endpoint no longer resolve a subscription tier at all.
- Membership instead grants a monthly allowance of NEW prices: free 3, bronze 10, silver 25, gold unlimited, counted in a new PricingSlot ledger keyed like PaidAccess. One slot per entity, spent on first pricing, never returned.
- A creator score of 10,000 is required to apply a price to something that has none. Moderators are not exempt from the floor or the allowance; they remain exempt from the fee ceiling.
- Both rules gate only the unpriced → priced transition. Editing, lowering or clearing a price that is already set is always free, at any score — which grandfathers everything priced before these rules existed, and is why the cutover needs no backfill.

Creator Studio enforces both rules itself: its fee and gate writes are direct SQL that never reaches the service layer, so a rule not applied there is a rule the spoke can bypass. The rules themselves live in @civitai/buzz so the two implementations cannot drift on a threshold or a message.

Also reverts the unshipped lapse-grace work: it mitigated lapse-repricing, and a lapse no longer changes any price.

MIGRATION — 20260821120000_pricing_slot is applied by hand, per environment, and must land BEFORE this code deploys: until the table exists every first-time pricing write throws on a missing relation, and the Creator Studio models page 500s outright. Apply it under a short lock_timeout — the inline owner FK takes SHARE ROW EXCLUSIVE on "User". See the migration header.


JustMaier:
## Review: 8 lanes over the 67-file segment

Ran the five main-app lanes (correctness, intent, tests, perf, reuse), the three Svelte lanes (Creator Studio enforces half these rules in its own SQL), and comment-review. Segment scoped to the branch's own non-merge commits — an unrestricted `origin/main...HEAD` diff pulls in ~23 files from the two main merges.

**Intent: the PR does what article 33749 outlined.** Every fee-side promise checks out. Three rules differ from the published article — the 10k score floor, slot return on clear, model-version scope — and Justin confirmed all three as his decisions. No intent gaps.

Note the article itself is now out of date on two of those (it says "Membership is scale, not permission" and "If you remove a fee later, you don't get that application back"). The checklist's "Update Justin's article" box is still unchecked.

---

### Hold for these

**1. `max(date)` vs `slotCreatedAt` releases a slot on a version that was actually paid for.**
`apps/creator-studio/src/lib/server/monetization/pricing-slot.ts:135-152`. `orchestration.resourceCompensations.date` is a DATE, so `max(date)` is always midnight UTC, and `last < r.slotCreatedAt` is true for any charge made on the slot's own creation day. Price a version at 09:00, someone pays the fee at 14:00, clear it at 20:00 → reads as never charged and the slot comes back.

The main app asks the same question as `date >= toDate(since)` (`src/server/services/pricing-slot.service.ts:93-97`) and holds the slot. Both modules carry a comment saying they are the same three signals in the same order — they are, except for this boundary. Three lanes reached this independently.

**2. Replica read-after-write silently eats the slot.**
`src/server/services/pricing-slot.service.ts:185,194-200` re-reads post-write state from `dbRead` after the clear committed to the primary. Under lag the old fee still shows, `isAlreadyPriced` is true, the function returns `false`, and the slot stays spent for the rest of the month with no error. This is the "priced a draft to see what it looked like" case the release exists for. `model-version.controller.ts:417` documents the same trap 100 lines above and uses `dbWrite` for exactly this reason. Same shape in the spoke (`pricing-slot.ts:92` reached from `licensing-fee.ts:390`) — fix both or neither.

**3. The allowance is a read-check-write with nothing serializing it.**
`pricing-slot.service.ts:269-272` counts, compares, and the caller records only after the write lands. No lock, no transaction spanning check to record, and `PricingSlot`'s PK is the entity so it cannot bound a per-owner count. N concurrent `modelVersion.upsert` calls each pricing a different unpriced version all read the same `used` and all pass. Same in the spoke. The spoke already guards this shape for the early-access quantity cap (`+page.server.ts:310-312`) but not for the allowance.

**4. The "already priced" flag comes from a one-hour cache.**
`paid-access.service.ts:567-570` reads `hadPermanentGate` via `getPaidAccess` (`paidAccessCache`, ttl 1h, no SWR), while the fee half of the same decision reads the primary (`model-version.controller.ts:418-428`). `wasPriced` switches off both the floor and the allowance and spends no slot — so a cache entry that still claims a gate after the gate is gone makes the next write a free unchecked edit. The bust is awaited on the clear path, so this needs a bust that does not land. The inverse is harmless (`recordPricingSlot` is PK-idempotent). Read the gate from the primary, as the fee already is.

**5. The spoke's fee ceiling has no test.**
Delete both `raisesOverCap` blocks from `apps/creator-studio/src/lib/server/monetization/licensing-fee.ts:210-216,344-357` and Creator Studio becomes an unbounded fee-write path with the whole segment still green. Every other rule has a test in every app that enforces it — the POI rule, the floor and the allowance all got that treatment here. The ceiling is the one that did not.

---

### Tests that pass over a broken rule

- `src/server/services/generation/__tests__/paid-access-gating.test.ts:189,211` — the viewer assertion was replaced by `expect.anything()`, not added to. The tier-ceiling reason for it is gone, but the viewer still decides the sale, which `paid-access.service.test.ts:737-744` calls out as a money bug. Changing `viewer: user` to `viewer: {}` prints nothing.
- `src/server/services/__tests__/pricing-slot.service.test.ts:38-42` — `dbMock.findUnique` ignores `select`, and one stub serves both `findUnique` calls. Dropping `licensingFee` from the select at `pricing-slot.service.ts:195` returns a slot on a version that still charges; test at `:177` still passes. Dropping `initialPublishedAt` releases everything; `:216` still passes. Assert the argument, not just the return.
- `AND source = 'licenseFee'` has no test in either app — `pricing-slot.service.test.ts:82` asserts the values array, not the SQL text, and the spoke's guard checks only the date literal.
- `apps/creator-studio/src/lib/server/__tests__/licensing-fee-poi-guard.test.ts:38-44` — the chain proxy makes `.where()` a no-op, so deleting the three ownership clauses from `pricing-slot.ts:42,125,126` leaves all 20 tests green. `pricing-slot.ts:28` states the claim; nothing pins it. The main app's equivalent is pinned, which is the tell.
- `apps/creator-studio/src/lib/server/timeout.ts` has no test; the poi-guard suite only exercises the throw path, never the timeout.

⚠️ The test lane could not execute anything — that worktree has no `node_modules`. All of the above is from reading. Nobody should read it as "the suites were run."

---

### Creator Studio UI

- `PaidAccessFields.svelte:89-94` fires on `canChoosePermanent`, which is false for the floor *or* the allowance. A creator below 10k with a fresh month reads "You've used this month's pricing allowance (0 of 3 priced this month) — upgrade your membership." Upgrading does not help. In the bulk dialog it renders alongside the correct "You can't charge for a model yet" alert. `PaidAccessContext` carries no `permBlockedReason`; adding it is the smallest fix.
- `+page.svelte:341-343` gates the sidebar Save on the floor only. The server also refuses on allowance (`licensing-fee.ts:223-225`), so a creator with the month used up gets an enabled button and a toast error. `BulkActionDialog.svelte:263-268` fixed exactly this for bulk.
- `+page.svelte:655` passes `remaining - newSlotsUsed + selected.size`. The trigger is algebraically right; `BulkActionDialog.svelte:419` then renders that value as prose. Limit 10, used 8, 5 selected of which 1 already priced gives "You can price 3 more models this month" when the truth is 2.
- `alreadyPricedIds` is derived from the loaded page only (`+page.svelte:189-196`), but "select all matching filters" selects off-page (`+page.server.ts:125`). A creator bulk-raising fees on 400 already-priced models is told they are over allowance for an operation the server charges nothing for.

---

### Perf — this PR is a net reduction, and the migration is already live

The cap-tier cache deletion is the strongest thing here. `getCapTier` calls `getAllUserSubscriptions`, a `dbWrite` findMany with two nested relation selects, so Prisma issues 3 **primary** queries — and the old `capTierCache` lookupFn was `Promise.all(userIds.map(getCapTier))`, i.e. 3xN primary queries on a cold cache, fanned out from the model detail page, the shop listing and `applyPaidAccessGating` on the generation path. That shape is gone from every read surface. `mini/[id].ts` went *down* one Redis batch read.

`PricingSlot` is already applied to prod: both indexes present, 0 rows, and `EXPLAIN (ANALYZE, BUFFERS)` on the replica shows `PricingSlot_ownerId_createdAt_idx` serving the count at 0.022 ms. The ordering hazard in the PR body holds and applying ahead of the deploy was the right order.

Two things to fix:
- `versionIdsSchema` (`+page.server.ts:62`) has no length cap, and `pricing-slot.ts:181-188` interpolates the ids into ClickHouse by string concat. A 50k-id POST builds a multi-MB statement plus a Postgres `IN` and an N-row insert. A `.refine(ids => ids.length <= 200)` bounds all four new consumers.
- The CH charge lookup prunes on `date` only (`userId` sits between `date` and `modelVersionId` in the sort key, and is deliberately unbounded). Measured from `system.query_log`: 30d = 1.23M rows / 8 ms, 90d = 3.56M / 13 ms, 365d = 18.4M / 112 ms, all 800d = 32.8M / 301 ms. `since` is `slot.createdAt` with no age ceiling. Low-frequency path behind a 3 s timeout, so not urgent — but the comment at `:66-67` says "the query measures ~4ms", which is the 3-day case and hides the shape.

Per-save cost vs base: plain save +1 primary query and +1 Redis (`getModel` and `getPaidAccess` are now unconditional); fee-setting save roughly neutral; price-clearing save adds up to 5 serial replica reads plus a CH query, awaited inline.

---

### Reuse

The ledger is **not** a duplicate of `PaidAccess`/`EntityAccess` — a fee-only price writes no `PaidAccess` row and `licensingFee` carries no "set at" timestamp, so a monthly allowance over fees has nothing to count from. Checked and cleared.

- `withTimeoutFallback` was copied into `apps/creator-studio/src/lib/server/timeout.ts` past `@civitai/shared`, which the spoke already declares — and the copy has already dropped the `onTimeout` guard.
- `MONETIZATION_MIN_CREATOR_SCORE` now means two different quantities: pricing reads `scores.models`, the sales gate reads the aggregate. Someone with 10k of image score can schedule a sale but not set the price it discounts. Pre-existing on the sales side; this PR is what makes it a shared gate.
- "Is it priced?" is defined in `@civitai/buzz` (`isAlreadyPriced`) and re-expressed in SQL by the spoke (`pricing-slot.ts:45,130`). Identical today. A third kind of price updates the shared predicate and silently misses the two spoke filters.
- The onsite below-floor alert (`ModelVersionUpsertForm.tsx:1305-1316`) hand-writes a third wording of `pricingFloorMessage`.

---

### Comments

- `pricing-slot.ts:234` reads the score and then calls `pricingFloorMessage()` with no argument, so the spoke's 403 quotes the threshold without saying how far off the creator is. The main app passes it. Code fix, one word.
- `apps/creator-studio/src/lib/server/monetization/paid-access.ts:298` says clearing "never returns a spent slot" — false. Those writes go to `/api/v1/model-versions/early-access`, which releases at `:114`. This comment sent two review lanes down the wrong path.
- Migration header `20260821120000_pricing_slot/migration.sql:19-23` asserts a slot is never returned, with a `^ SUPERSEDED 2026-08-24` note bolted underneath. Replace both with one true statement.
- `licensing-fee.ts:69` points at `monetizationLimits().fee.suggested` / `.suggestedPerGeneration`, neither of which exists — `monetization-limits.ts:36-38` says the opposite explicitly.
- "Moderators are NOT exempt" appears four times; keep it at the enforcement point.
- `CapUpsell` and `TierCapsTable` no longer render a cap — they render the monthly allowance. The stale names are what the header comments are bridging.

---

### Not defects, recorded once

Stored fees above the old per-tier cap begin billing generators in full on deploy — nothing clamps at charge time any more, and the write-path ceiling is raise-only. `docs/licensing-fee-revamp-checklist.md:295-301` records this deliberately; nothing sweeps or notifies the affected versions. Product call.

The floor and the allowance do not reach timed early access by design, so a below-floor creator can still sell a timed window, bounded by the existing EA day/quantity limits.

One separate item went to Justin and briant directly rather than here.

PR #4273, opened 2026-08-22
fix(search): keep banned accounts out of user search and out of the collaborator invite picker

ZacxDev:
## The problem

A banned account could stay in the user search index indefinitely, and could stay there under the username it had **before** a moderator renamed it. So a moderation action was invisible to every surface that reads user search — public user search, and any picker built on it.

That matters most where a search result is acted on **by id**: the app-listing collaborator invite picker submits the id of the row you choose, so an account that search is describing incorrectly is an account you can pick incorrectly.

## Root cause — two independent mechanisms, both real

**1. Nothing removed a banned account from the index, and things put it back.**
The predicate deciding who belongs in the user index was `id != -1 AND deletedAt IS NULL` — no ban condition — in both the indexer (`users.search-index.ts`) and the nightly reconciler (`meilisearch/cleanup.ts`). Ban does enqueue a filtered removal, but the user-metrics refresh enqueues a search-index update for **every** account whose counters moved, and that update re-indexed a banned account straight back in. A full rebuild did the same.

**2. A moderator-forced rename never reached the index.**
`forceUpdateUserIdentity` writes the new username and enqueues nothing. The incremental sync's range scan keys on **`createdAt`**, so it never revisits an existing row, and the nightly reconciler only decides whether an id still qualifies — it never refreshes a field. So the index kept serving the old name for as long as the document lived. The self-serve profile save has always enqueued an update; the moderator path did not.

Both were confirmed by reading the code paths, not inferred from the symptom.

### Index membership
- One shared eligibility predicate (`search-index/user-index-eligibility.ts`), now including `bannedAt IS NULL`, used by **both** the indexer and the nightly reconciler. Both halves are load-bearing: the write-side filter alone leaves an already-indexed account there forever, and the reconciler-side filter alone is undone by the next metrics-driven update. The reconciler half is what evicts documents that are already there, with no manual reindex.
- A moderator-forced username/display-name change now enqueues a search-index refresh.
- Lifting a ban re-adds the account to search, so the exclusion is not a one-way door.
- Both of those enqueues are Redis writes that run *after* their transaction has committed and *before* a later step (session invalidation; the unban email), so they are isolated like every other side effect in that file — a Redis blip cannot skip the step after them or fail an action that already succeeded.

### Collaborator invite picker
- The picker's candidates come from user search, i.e. from cached documents, so it must not conclude anything about an account from the fact that search returned it. The panel now screens the ids it is offering against the account rows themselves (`appCollaborators.ineligibleTargets`) and stops offering the ones a seat would be refused for. Ineligibility outranks every roster reason, including the re-invitable declined seat, and an id with no account row at all is treated as ineligible rather than fine.
- That screening call is keyed to a listing the caller **owns** and is rate-limited, so it cannot become an account-state enumeration surface if the authoring audience widens. Only owners are shown the picker, so only owners issue it.
- `inviteCollaborator` now refuses a **soft-deleted** target as well as a banned one, as not-found. The screen and the enforcement point previously disagreed about that set while a comment claimed they matched; the enforcement point was widened rather than the screen narrowed, so no protection was removed.
- `QuickSearchDropdown` gains one optional, inert-by-default `onHits` prop. No other call site changes behaviour.
- **Defence in depth only.** `inviteCollaborator` remains the enforcement point that makes the grant impossible.

## Tests

New coverage, each shown failing on pre-change code:

| test file | at base | at HEAD |
|---|---|---|
| `search-index/__tests__/user-index-eligibility.test.ts` | fails to load (module absent) | 10 pass |
| `services/blocks/__tests__/app-collaborator.screen-targets.test.ts` | 7 fail (function absent) | 10 pass |
| `services/__tests__/user-identity-rename-search-index.test.ts` | 2 fail, 2 pass | 5 pass |
| `services/__tests__/toggle-ban-search-index.test.ts` | 1 fail, 1 pass | 3 pass |
| `components/Apps/AppCollaboratorsPanel.browser.test.tsx` | new file — the seam did not exist | 6 pass |
| `components/Search/QuickSearchDropdown.onHits.browser.test.tsx` | new file | 2 pass |
| `AppCollaboratorsPanelView.browser.test.tsx` (+5) | 3 fail, 2 pass | 35 pass |
| `app-collaborator.service.test.ts` (+2, soft-delete) | 1 fail, 1 pass | 116 pass |

Base-passing tests are labelled in-file as invariant guards rather than counted as regression coverage.

### Mutation testing

An adversarial review found **two survivors** in the first round, both with the guard's own text still present, which is why the assertions here are shaped the way they are:

- the write side was pinned by the **name** of a constant while a local alias was what the queries actually consumed — re-binding the alias dropped the ban clause from every write-side query, suite green. There is now one name, it is a function, and the tests drive the real query builders and pin the SQL that reaches the driver.
- the reconciler assertion used `toContain`, which any statement that merely *mentions* the clause satisfies — widening the predicate to `(<eligibility> OR bannedAt IS NOT NULL)` made it keep every banned document, suite green. Now pinned whole with `toBe`.

Both are killed now, and the sweep was re-derived from the expression's semantic failure modes (re-bind, clause-preserving inversion, operand inversion, comment-out, argument drop, half-fix, refuse-everything) rather than from deletion alone. Every absence assertion is paired with a presence assertion.

`tsc -p tsconfig.tests.json`: no new errors attributable to this branch.

## Not covered

- The `role === 'owner'` half of the screening query's `enabled` condition is redundant with the `ids.length > 0` half — an editor is never shown the picker, so the id set is always empty for them. It states intent and is defence in depth; it has no independently-killing mutation, and that is deliberate rather than an oversight.
- Removing already-indexed banned accounts happens on the nightly reconciler pass; nothing here was run against a live index.

### DIFF
--- a/src/components/Apps/AppCollaboratorsPanel.browser.test.tsx
+++ b/src/components/Apps/AppCollaboratorsPanel.browser.test.tsx
@@ -28,6 +28,8 @@ const state = vi.hoisted(() => ({
   ineligible: [] as number[],
   /** Ids the picker reported as being on offer. */
   screenedInput: [] as number[],
+  /** The listing the screening query was keyed to. */
+  screenedListing: null as string | null,
   /** Every `invite` mutation actually fired. */
   inviteCalls: [] as unknown[],
   /** The last `filters` string handed to the picker. */
@@ -36,6 +38,8 @@ const state = vi.hoisted(() => ({
   select: null as ((item: unknown, data: unknown) => void) | null,
   /** The picker's `onHits`, so a test can drive what is on offer. */
   reportHits: null as ((ids: number[]) => void) | null,
+  /** Whether the screening query was enabled on the last render. */
+  screenEnabled: false,
 }));
 
 /**
@@ -83,8 +87,10 @@ vi.mock('~/utils/trpc', async (importOriginal) => {
         list: { useQuery: () => ({ data: [], isLoading: false, error: null }) },
         getPendingTransfer: { useQuery: () => ({ data: null, isLoading: false, error: null }) },
         ineligibleTargets: {
-          useQuery: (input: { userIds: number[] }) => {
+          useQuery: (input: { appListingId: string; userIds: number[] }, opts?: any) => {
             state.screenedInput = input.userIds;
+            state.screenedListing = input.appListingId;
+            state.screenEnabled = opts?.enabled !== false;
             return { data: state.ineligible, isLoading: false, error: null };
           },
         },
@@ -101,11 +107,11 @@ vi.mock('~/utils/trpc', async (importOriginal) => {
 
 const { AppCollaboratorsPanel } = await import('~/components/Apps/AppCollaboratorsPanel');
 
-const render = () =>
+const render = (role: 'owner' | 'editor' = 'owner') =>
   renderWithProviders(
     <AppCollaboratorsPanel
       appListingId="apl_seam"
-      role="owner"
+      role={role}
       capabilities={capabilitiesForKind('onsite')}
       listing={{ kind: 'onsite', connectClientId: null }}
     />
@@ -114,6 +120,8 @@ const render = () =>
 beforeEach(() => {
   state.ineligible = [];
   state.screenedInput = [];
+  state.screenedListing = null;
+  state.screenEnabled = false;
   state.inviteCalls = [];
   state.pickerFilters = null;
   state.select = null;
@@ -132,6 +140,28 @@ describe('AppCollaboratorsPanel — candidate screening reaches the picker', () 
     await vi.waitFor(() => {
       expect(state.screenedInput).toEqual([FINE_ID, INELIGIBLE_ID]);
     });
+    // Keyed to the listing the proc asserts ownership of — without it the call 400s on the
+    // schema and every candidate reads as eligible.
+    expect(state.screenedListing).toBe('apl_seam');
+  });
+
+  /**
+   * 🔴 The proc asserts the caller OWNS the listing, and only an owner is shown the invite
+   * picker. An editor issuing it would 403 on every render of a tab they are allowed to see.
+   */
+  test('an EDITOR does not issue the screening query at all', async () => {
+    render('editor');
+    await vi.waitFor(() => expect(state.screenEnabled).toBe(false));
+    // The editor is not shown the invite picker, so there is nothing to screen.
+    expect(state.reportHits).toBeNull();
+  });
+
+  /** The control arm for the test above — an OWNER does issue it. */
+  test('an OWNER does issue the screening query', async () => {
+    render();
+    await expect.element(page.getByTestId('stub-invite-picker')).toBeInTheDocument();
+    state.reportHits!([FINE_ID]);
+    await vi.waitFor(() => expect(state.screenEnabled).toBe(true));
   });
 
   test('an id the server calls ineligible is filtered OUT of the picker', async () => {
--- a/src/components/Apps/AppCollaboratorsPanel.browser.test.tsx
+++ b/src/components/Apps/AppCollaboratorsPanel.browser.test.tsx
@@ -40,6 +40,8 @@ const state = vi.hoisted(() => ({
   reportHits: null as ((ids: number[]) => void) | null,
   /** Whether the screening query was enabled on the last render. */
   screenEnabled: false,
+  /** How many times the screening hook ran — the positive control for the assertion below. */
+  screenCalls: 0,
 }));
 
 /**
@@ -91,6 +93,7 @@ vi.mock('~/utils/trpc', async (importOriginal) => {
             state.screenedInput = input.userIds;
             state.screenedListing = input.appListingId;
             state.screenEnabled = opts?.enabled !== false;
+            state.screenCalls += 1;
             return { data: state.ineligible, isLoading: false, error: null };
           },
         },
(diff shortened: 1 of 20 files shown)


ZacxDev:
## Adversarial audit — two corrections to this PR's description, and two surviving mutants

Posting rather than editing the body, since a reviewer may already have read it.

### 1. "Ten deliberate mutations were applied and each was killed by a test asserting its own specific expectation"

**Two independently-enumerated mutants survive.** Both were run in an isolated copy, and a positive control (deleting the clause from `user-index-eligibility.ts`) was confirmed **red with 4 failures** first, so these are real survivors and not a suite wired to nothing.

**MUT-A — the indexer's real `WHERE` is not pinned.** `USERS_INDEX_WHERE` is exported and asserted, but `users.search-index.ts:113` does `const WHERE = USERS_INDEX_WHERE`, and *that* alias is what `prepareBatches`/`pullData` consume. Replacing line 113 with an inline two-element copy drops `bannedAt IS NULL` from every write-side query — **suite 6/6 green**. The test asserts the *declaration*, never the *code path*.

This is the same seam the PR describes as its hardest-won finding. The widening pins the exported constant, so the hole is one line of indirection away from reopening. Suggested fix: export a `buildUsersIndexWhere()` and pin its *output*, or drop the alias so there is only one name.

**MUT-B — the reconciler assertion is substring-only.** `user-index-eligibility.test.ts:75-79` uses `toContain('u."bannedAt" IS NULL')`, which is satisfied by any statement that merely *mentions* the clause. Changing `cleanup.ts:73` to `AND (${userSearchIndexEligibilitySql()} OR u."bannedAt" IS NOT NULL)` makes the reconciler **keep every banned document** — the exact behaviour this PR exists to prevent — and the suite stays **6/6 green**.

The shared module and the indexer constant *are* whole-string pinned, correctly. The reconciler's composed SQL is not. Pin `norm(usersConfig.where([11,22]).sql)` with `toBe`.

Generalising: the sweep found the mutants it imagined. Deletion mutants are covered; **re-binding and clause-preserving inversion are not**, and those are the two shapes that leave the guard's text present so every regex still matches.

### 2. "stops offering the ones a seat would be refused for"

Inaccurate for **soft-deleted** accounts, as is the service comment at `app-collaborator.service.ts:1041` ("hands back the ids the invite mutation is going to refuse").

`getIneligibleCollaboratorTargets` (`:1051-1070`) rejects banned, soft-deleted **and** missing. `inviteCollaborator` (`:420-433`) selects only `{ id, bannedAt }` and refuses only missing and banned — **there is no `deletedAt` check**. So the PR advertises a protection the server does not enforce, and a soft-deleted account reached by any path other than this picker is still seatable.

Two valid resolutions — worth a deliberate choice rather than picking the smaller diff: add `deletedAt` to `inviteCollaborator`'s select and refusal (closes the gap, changes enforcement), or narrow the screen and the prose to match today's enforcement.

### What the audit confirms

- **"Every absence assertion is paired with a presence assertion" holds** — checked individually (`not.toContain(FINE)` + `toContain(BANNED)`; the eligible-candidate-DOES-invite control arm; the leaves-other-indexes-untouched presence half). That claim is accurate.
- The two-halves design (write filter + reconciler) is **correct as implemented** — both paths were traced and the reconciler genuinely will evict.
- 36/36 node tests and 53/53 browser tests pass at HEAD; `tsc -p tsconfig.tests.json` shows **0 errors in any file this PR touches**, with no `FATAL ERROR`.

### Other items worth addressing before merge

- **`QuickSearchDropdown.tsx:245-249` has no coverage.** Both test files that reference the component `vi.mock` it, so deleting the `useEffect` that calls `onHits` leaves the whole suite green while the panel-side filtering becomes **completely inert** — `screenedUserIds` never populates, the query never enables, the picker is unfiltered. The wiring is correct as written; the exposure is that nothing would tell you if it broke.
- **Two unguarded `await`s in moderation paths**, in functions whose every sibling side-effect is `.catch(logToAxiom)`-wrapped: `user.service.ts:1990` (unban) and `:590` (`forceUpdateUserIdentity`). Both are Redis writes via `queueUpdate`. On a Redis blip the unban has already committed but the moderator gets a 500 **and the `account-unbanned` email is skipped**; in the rename case the throw lands *before* `invalidateSession`, leaving the rename committed and the session live.
- **`ineligibleTargets` (`app-collaborators.router.ts:479`) has no `rateLimit()`** — unlike its sibling `invite` — takes 100 arbitrary ids per call, and carries no `appListingId` tying the caller to anything they own. Minor today given the `appDeveloperProcedure` audience, but nothing couples the two, so it becomes an enumeration endpoint the day that audience widens.
- **The first nightly `search-index-cleanup` after merge has a substantial backlog to evict** (sized separately). Note that `cleanup.ts` accumulates `allStaleIds` in memory and only begins deleting *after* the full index scan completes — so if the users scan overruns the job's 2 h `lockExpiration`, **nothing is deleted, run after run**, and the only signal is the job's returned `deleted` count. Worth reading that count on the first run rather than assuming eviction happened.
- **Orphaned JSDoc**: the new function was inserted between `inviteCollaborator` and the large JSDoc block written for it, so the "🔴 BAN POLICY" block now dangles above the wrong function and `inviteCollaborator` has none.

Verdict: **merge after the two behaviour-affecting items** — the `deletedAt` mismatch and the two unguarded enqueues. The rest is test strength and hygiene, but MUT-A/MUT-B should be closed too, since they are the guards this change relies on.

PR #4290, opened 2026-08-22
fix(stickers): ask for a top-up offer only where one was asked for

JustMaier:
Closes `868kv5cdm`.

## What was wrong

`useStickerRefill` fetched a top-up offer for **every sticker the placer owns**, sliced to `STICKER_OFFER_LIMIT` (100):

```ts
const ownedIds = sticker.slice(0, STICKER_OFFER_LIMIT).map((o) => o.id);
```

Stickers 101 and beyond — newest-obtained first, so the ones bought most recently — were never asked about. `refillFromOffer` then built a gate with **no pack**, which renders as *"this sticker sells no extra uses, and it is not on sale right now."* False, and permanent for those stickers.

⚠️ **The comment above that line was wrong**, and would have sent the next reader after the wrong thing: it said the query "fails zod validation and `offers` stays undefined" past the cap. It does not — the `slice` is exactly what prevents that. The failure is silent truncation, not a rejected request. Corrected here.

**Who it hits:** production has **459** accounts owning stickers; **one owns 126**, so 26 of theirs have no top-up path today. The next largest owns 83 — seventeen purchases from the same wall.

## Why it was fetching everything

For a cache key, not for a need. An earlier version keyed the query on the *drafted* ids and refetched every time a draft was laid down; keying on the *owned* ids gave one stable entry two callers could share. There is **one** caller now (`DraftStickerLayer`), and the doc comment describing the other one had gone stale.

And nothing needs the answer early: `refillFor(...)` is called **during render**, per drafted sticker (`DraftStickerLayer.tsx:306`), not snapshotted at drag time.

## The change

`useStickerRefill(cosmeticIds)` takes the ids it should ask about — the drafted ones, usually one or two, never the collection. Chunked by the endpoint's own maximum, in **insertion order**, so laying down another draft appends rather than reshuffling keys already fetched, and no collection can outgrow the request.

The chunking was the same loop written twice (`useStickerCosmetics` had its own), so it is now one shared `chunkStickerIds` helper with the reasoning in one place.

## Test

`src/components/Sticker/__tests__/sticker-offer-chunks.test.ts` — the bug it replaces was a `slice`, and a truncation is invisible from inside a successful query: the request succeeds, the map is populated, every lookup for an early id works. So the property asserted is **total coverage** against an input larger than one chunk (207 ids), plus insertion-order stability, dedupe, and the empty case.

Revert-checked: making the helper truncate to the first chunk fails with `expected [ Array(100) ] to have a length of 207 but got 100`.

Verified: typecheck 0 errors, eslint clean on the changed files, `test:lint-rules` 281 passed, full unit suite **1332 files / 20805 passed**.

### DIFF
--- a/src/components/Sticker/DraftStickerLayer.tsx
+++ b/src/components/Sticker/DraftStickerLayer.tsx
@@ -12,6 +12,7 @@ import {
   unownedGateFor,
   useOwnedSticker,
   useStickerCosmetics,
+  draftedCosmeticIds,
   useStickerRefill,
 } from '~/components/Sticker/sticker.util';
 import { useCurrentUser } from '~/hooks/useCurrentUser';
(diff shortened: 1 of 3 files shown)


JustMaier:
Review lanes ran over this. Four findings acted on, pushed as `e70f753e02`.

**One real regression, mine, from the first commit.** Keying the query on the whole drafted-id list meant dragging out a second sticker minted a new key — so the *first* sticker's offer went missing until the refetch landed, its "buy more uses" button vanishing mid-session on a money surface. Correctness and perf lanes found the same window independently. Fixed by keying **per sticker id**: adding a draft now leaves every offer already fetched exactly where it is.

**The tests were guarding the wrong half.** They covered `chunkStickerIds`, which was never broken. The decision this PR actually moved is `drafts.map((d) => d.cosmeticId)` at the call site — and `draft.id` and `draft.cosmeticId` are both numbers on the same object, so swapping them typechecks while every draft reads "sells no extra uses" forever. That is the shipped bug's own symptom one layer up, and nothing was pinning it. Now a named `draftedCosmeticIds` helper with a test; revert-checked, both assertions go red with legible messages.

**Two fixtures were weaker than they read.** Both order fixtures were monotonic descending, which a `sort((a,b) => b-a)` derivation satisfies unchanged — the insertion-order property was pinned by the dedupe test alone. They are deliberately unsorted now. And a `for … expect` over an empty array asserts nothing and reports green, so the chunk count is asserted before the loop.

**Perf: nothing blocking, one number worth carrying.** `getStickerOffers` costs the same for 1 id as for 100 — `_count: { purchases: true }` compiles to a LEFT JOIN that seq-scans all 33,975 rows of `UserCosmeticShopPurchases`, which has exactly one index (its PK). Measured 7.67 ms / 697 shared blocks at 1 id, 7.74 ms / 697 at 2. So this PR's win is the truncation fix, not the ~1 ms of narrower input. Pre-existing, ~108 calls/day site-wide; an index on `UserCosmeticShopPurchases("shopItemId")` is the lever when stickers get traffic.

**Comment lane found one genuinely false comment** — a doc on the `StickerOfferLike` *type* describing the function below it — plus four blocks restating rationale that already lives on the function being called. Trimmed. `STICKER_FETCH_CHUNK` now *is* the schema cap rather than a copy of it with a comment saying so.

**Reuse: nothing rebuilt.** Verified no other call site caps an id array before a sticker query.

Verification after the fixes: typecheck 0 errors, eslint clean, full unit suite **1332 files / 20807 passed**.

PR #4292, opened 2026-08-22
feat(stickers): search and sort the placement tray

JustMaier:
Follows the tray overflow fix. Wrapping 83 stickers into two scrollable rows made them **reachable**; it did not make one **findable**.

## What this adds

- **Text filter** over the names already loaded client-side by `useOwnedSticker` — no query, no schema change.
- **Sort dropdown** in the header: *Recently used* (default) or *Recently acquired*.
- Both appear above **12 stickers owned**. Below that they are two widgets over three stickers.

**The secondary sort is stability, not a comparator branch.** `sticker` arrives newest-acquired first and `Array.prototype.sort` is stable, so two stickers the placer has never used keep that order for free. Writing the tie-break by hand would be a second copy of a rule already applied upstream.

## The recency data

New `cosmetic.getStickerRecentUse` — `[{ cosmeticId, lastUsedAt }]` for the viewer, from `Placement`.

The 30-day window is for **meaning**, not cost: "recently used" that surfaces a sticker last placed in March is not recency. Cost is handled by the `LIMIT` — `placerId` is indexed, `cosmeticId` lives in the `data` jsonb with no expression index, so the filter is on `placerId` and the id is extracted in the app. Bounded by the row cap, not by table size.

Measured on production: **0.286 ms, 2 shared buffers, index scan on `placerId`**. `Placement` is 1,419 rows; the heaviest placer on the site has 80. Every placement counts whatever its status — a placement awaiting review is still a sticker the placer reached for.

**Verified end to end against production data**, not just eyeballed: signed in as an account owning 83 stickers, the first five tiles are exactly that account's last five placements in recency order. Typing `civ` narrows 83 to 14, matching on name.

## Mobile

At 390px the two controls took **304 of 390 pixels** and squeezed the instruction into a four-line column. Below `sm` they now drop to their own row: text column 304px, both controls inside the panel, no page overflow, header 94px. Desktop unchanged at 56px, controls still inline. Checked in a real browser at both widths.

## ⚠️ It also fixes `main`, which is red

The tray's browser test still asserted `Free, or 100 Buzz` and `100 Buzz + one use` — copy changed in #4289. Those tests **run in no CI job** and are excluded from the unit project, so a green full-suite run never saw them. Corrected here, along with mocking the new query in the three tray suites.

Verified: typecheck 0 errors, eslint clean, sticker + server unit projects **14,701 passed**, tray browser suites 9 passed.

### DIFF
--- a/src/components/Sticker/StickerPlacementTray.browser.test.tsx
+++ b/src/components/Sticker/StickerPlacementTray.browser.test.tsx
@@ -106,6 +106,7 @@ vi.mock('~/utils/trpc', async (importOriginal) => ({
   trpc: {
     cosmetic: {
       getStickerBalances: { useQuery: () => ({ data: [] }) },
+      getStickerRecentUse: { useQuery: () => ({ data: [] }) },
       getStickerOffers: { useQuery: () => ({ data: [] }) },
     },
   },
@@ -201,7 +202,7 @@ describe('StickerPlacementTray — the reason free is unavailable', () => {
     });
     await renderTray();
 
-    await expect.element(page.getByText(/Free, or 100 Buzz/)).toBeInTheDocument();
+    await expect.element(page.getByText(/Free placement \+ one sticker use/)).toBeInTheDocument();
     expect(page.getByText(/used today's free placement/).elements()).toHaveLength(0);
     expect(page.getByText(/already used a free sticker/).elements()).toHaveLength(0);
     expect(page.getByText(/free slot on this image is taken/).elements()).toHaveLength(0);
@@ -218,7 +219,7 @@ describe('StickerPlacementTray — the reason free is unavailable', () => {
 
     // The price line still renders — the tray is fine, it simply has no refusal
     // to explain, because nobody was offered anything.
-    await expect.element(page.getByText(/100 Buzz \+ one use/)).toBeInTheDocument();
+    await expect.element(page.getByText(/100 Buzz \+ one sticker use/)).toBeInTheDocument();
     expect(page.getByText(/used today's free placement/).elements()).toHaveLength(0);
   });
 });
--- a/src/components/Sticker/StickerPlacementTray.overflow.browser.test.tsx
+++ b/src/components/Sticker/StickerPlacementTray.overflow.browser.test.tsx
@@ -86,6 +86,7 @@ vi.mock('~/utils/trpc', async (importOriginal) => ({
   trpc: {
     cosmetic: {
       getStickerBalances: { useQuery: () => ({ data: [] }) },
+      getStickerRecentUse: { useQuery: () => ({ data: [] }) },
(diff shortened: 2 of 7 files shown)


JustMaier:
Four review lanes ran over this. Three findings acted on, pushed as `2883a949a6`.

**The recency query fired for everyone.** Production distribution: **455 of 466** sticker owners (97.6%) hold 12 or fewer and never see the sort control; **342 (73.4%) hold exactly one**, where an order is a no-op. tRPC batching is off in flipt, so it was its own unbatched round trip on the tray-open path buying them nothing. Now gated on owning more than one.

**A typed filter survived the tray closing.** The component stays mounted behind its early return, so closing with `cat` typed and reopening showed 2 of 83 stickers with nothing on screen explaining it — against the whole point of the feature. Cleared alongside the shop panel, which was already handled for exactly this reason.

**🔴 My comment on the query's cost was wrong.** It said the `LIMIT` bounds the scan. It does not: no index can supply `createdAt` order under a `placerId` equality here, so the plan bitmap-scans every row that placer has ever created, applies the 30-day window and `surface` as post-index filters, sorts, and only then discards. The cap bounds sort memory and output, not input. Measured on the replica — 80 rows: 0.13 ms / 30 buffers; 1,134 rows: 2.6 ms / 104 buffers with the sort switching to top-N heapsort. Fine at 1,426 rows, and the real bound if it ever matters is an index on `("placerId", surface, "createdAt" DESC)` — not worth a migration at this size. The comment now says that instead of pointing at the wrong guard.

(Also corrected: my 2-buffer measurement was the Bitmap Index Scan alone. With the heap fetch it is 30.)

**The filter and sort had no test that could go red — they do now**, and the fixture is the load-bearing part. The tests lane proved that a fixture with `id: i + 1` lets a hand-written `a.id - b.id` tie-break — the exact regression the stability comment warns about — pass every assertion, because obtained order and id order coincide. The fixture's ids descend against obtained order for that reason, and name/slug carry disjoint tokens so a name match cannot be satisfied by the slug half of the concatenation.

Verified by mutation, each turning a test red: the hand-written tie-break, dropping `.trim()` from the search term, and gating the controls on the filtered list rather than the owned list (which would make the search box delete itself as you type).

**Safety lane: no findings.** Confirmed the interval is a bound Prisma parameter, `placerId` is the only identity, nothing about another user is returned, and a missing `status` filter cannot revive a taken-down sticker into the tray because the visible list comes from the owned collection, never from the recency map.

**Two open questions I did not decide** — both in the PR body for you rather than silently resolved: the sort applies to everyone while the control to undo it appears above 12, and `> 12` shows at 13 rather than at 12.

Verification: typecheck 0 errors, eslint clean, full unit suite **1332 files / 20810 passed**, sticker browser suites green.

PR #4294, opened 2026-08-22
test(app-blocks): hold a virtual clock across the four PageBlockHost Retry clicks

ZacxDev:
The last known live members of the component-suite flake documented in `claudedocs/rca-appblocks-component-suite-flake-2026-08-05.md`.

Four tests in `PageBlockHost.browser.test.tsx` clicked Retry through the Playwright driver on the **real** clock and then asserted on `onRetryToken`, racing `AUTO_RETRY_BACKOFF_MS[0] = 2000ms`.

⚠️ Two of the four *did* install a fake clock — but via `advancePastWindow()`, which calls `vi.useRealTimers()` on the way out **by design**, so the click and everything after it ran on the real clock anyway. Installing a fake clock is not sufficient; it has to survive the click.

## The fix

Copied from the sibling that already solved this, `PageBlockHostAutoRetry.browser.test.tsx`: `useVirtualClock()` with `toFake` restricted to the four timer functions, so `Date`/`performance`/`rAF` stay real and the driver keeps working while the clock stays installed across the click. Virtual time advances **0 ms** between arming the timer and clicking, which makes the premise provable rather than probable.

`pollFor` replaces `vi.waitFor` in these four, because `vi.waitFor` *advances* the fake clock while polling — spending virtual time against the very backoff the test needs to stay pending. No assertion was deleted. No production code changed.

## 🔴 Site 3 was passing while testing the wrong thing

`fatal (non-auth)` does **not** go red under the accelerated control, and that turned out to be the interesting result rather than a gap.

It asserts a **zero** (`onRetryToken` not called), and a zero cannot be broken by an event that never touches the counter. What the race does instead is **silently substitute the branch under test**: the auto-retry flips status to `loading` mid-click, `handleRetry`'s `if (prior === 'loading' || prior === 'ready') return` guard makes the click a no-op, the host waits out a whole `BLOCK_READY` window and settles on the **`timeout`** terminal — whose Retry button the driver then clicks. Every assertion still holds. The test passes, exercising `timeout` rather than `fatal`.

The evidence is wall time: **200 ms → 10,531 ms at a 50 ms backoff, still reporting a pass** — almost exactly one `BLOCK_READY` window. A probe confirmed the pre-click state is fine (the substitution happens *during* the click), and its negative control confirmed the probe line actually executed.

## Controls

**Accelerated** (`AUTO_RETRY_BACKOFF_MS[0]` 2000 → 50, base arm run *first*, constant restored — `git diff` confirms `pageBlockHostLogic.ts` unchanged):

| site | base | fixed |
|---|---|---|
| 1 `error` | **RED** — `locator.click: Timeout 14845ms exceeded` / element detached | green 164 ms |
| 2 `no_token` | **RED** — `expected null not to be null` | green 137 ms |
| 3 `fatal` | passes, but 200 ms → 10531 ms (see above) | green 167 ms |
| 4 `timeout` | **RED** — `expected null not to be null` | green 232 ms |

**Mutation — two mutants, because two sites assert a one and two assert a zero.** Reporting only the first would have claimed coverage for sites 3–4 it never provided.

- Remove the re-mint call → sites 1 and 2 fail with `expected "vi.fn()" to be called 1 times, but got 0 times`, at their own assertion lines.
- Remove the re-mint *gate* → sites 3 and 4 fail with `expected "vi.fn()" to not be called at all, but actually been called 1 times`, at their own assertion lines.

Every kill names the site's own assertion — none died to an unrelated guard, locator timeout, or teardown error.

**10 consecutive runs: 10/10 green, 32 collected every run.** The harness was validated first: a bogus `-t` filter returns **`rc=0`** with `Tests 32 skipped` — zero collected — and is correctly scored a failure. The exit code alone would have called that success.

## A wrinkle the control earned

The first attempt kept the original 16 s/11 s jumps and went **red at ~19.6 s**: `advanceTimersByTimeAsync` fires *every* timer inside a jump, so 1 s of overshoot was spent against the backoff from inside the test. At the shipped 2000 ms that overshoot cannot reach it, so this would never have surfaced without the accelerated arm. Fixed by advancing to the exact boundary and nudging with `pollFor`.

## Not verified / noted

- Measured on one idle workstation. The flake is load-dependent, so the structural argument (the timing dependency is gone) is the real claim; the 10 runs corroborate it.
- Site 3's converted test does not itself prove it clicked from the `fatal` terminal — it relies on the virtual clock making the auto-retry unable to fire.
- Under the accelerated control, **three tests outside these four** also go red with the same helper pattern and 1 s overshoot. Green at the shipped constant; out of scope here, and the natural next candidates.
- eslint reports one **pre-existing** error on this file (line 33, `local-rules/no-wholesale-module-mock`), present on `main`, unrelated to this diff.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -350,7 +350,7 @@ async function advance(ms: number) {
  *
  * Two budgets, deliberately decoupled: at most 500ms of VIRTUAL time (well inside the
  * shortest recovery window, the 2s backoff, so a poll can never silently trip the very
- * behaviour a test is about to assert) and up to 5s of REAL time for the browser to
+ * behaviour a test is about to assert) and a bounded slice of REAL time for the browser to
  * actually do the work. Advancing virtual time alone would give the page only
  * microtasks — the failure mode that makes a fake-timer poll report "iframe never
  * mounted" on a saturated runner.
@@ -359,9 +359,27 @@ async function advance(ms: number) {
  * `vi.waitFor` ADVANCES the fake clock while polling, spending virtual time against the
  * very backoff the test needs to stay pending.
  */
+/**
+ * 🔴 THE ITERATION COUNT IS SIZED AGAINST TWO CEILINGS, NOT PICKED.
+ *
+ * Each fake-clock iteration costs ~30ms REAL (the flush loop's macrotasks plus the
+ * 10ms sleep) and 1ms VIRTUAL. So the budget is ~4.5s real / 150ms virtual.
+ *
+ * REAL ceiling: browser mode forces `testTimeout` to 15s and the root config's 60s
+ * does NOT reach the `component` project. At 500 iterations this loop ran ~15s and
+ * LOST THE RACE to that timeout — measured: the tests below without an explicit
+ * timeout died as `Test timed out in 15000ms` instead of naming what they waited
+ * for, and the one with `20_000` cleared it by 93ms. Staying well under 15s is what
+ * keeps a failure legible.
+ *
+ * VIRTUAL ceiling: every millisecond spent here is spent against
+ * `AUTO_RETRY_BACKOFF_MS[0]` (2000ms), which these tests need to stay PENDING.
+ * 150ms is ~13x inside it. Measured actual spend between terminal and click is
+ * 1-10ms, so this is headroom, not a working budget.
+ */
 async function pollFor(what: string, ready: () => boolean) {
   const fake = vi.isFakeTimers();
-  for (let i = 0; i < 500; i++) {
+  for (let i = 0; i < 150; i++) {
     if (ready()) return;
     if (fake) {
       await advance(1);
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -88,7 +88,11 @@ vi.mock('~/utils/trpc', () => ({
 }));
 
 // eslint-disable-next-line import/first
-import { PageBlockHost } from '~/components/AppBlocks/PageBlockHost';
+import {
+  BLOCK_READY_TIMEOUT_MS,
+  PageBlockHost,
+  TOKEN_WAIT_TIMEOUT_MS,
+} from '~/components/AppBlocks/PageBlockHost';
 // eslint-disable-next-line import/first
 import { IframeInitController } from '~/components/AppBlocks/iframeInitController';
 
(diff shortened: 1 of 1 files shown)


ZacxDev:
📌 **Correction to the PR body, plus one audit fix — posted publicly rather than edited in.**

**The body overstates the residual by one.** It says *"three tests outside these four also go red under the accelerated control"*. Measured on the shipped tree it is **two**: `loader clears after the BLOCK_READY timeout (timeout terminal path)` and `timeout terminal state: readable timeout message + Retry, not the loader` — both via `advancePastWindow(11_000)`'s 1 s overshoot. (The base tree at backoff=50 gives five red: those two plus sites 1, 2 and 4.)

**Fixed in `43636086bb`: `pollFor`'s budget was racing the browser test timeout.** Its comment claimed "up to 5 s of REAL time"; measured, 500 iterations cost **~15 s** (~30 ms each). Browser mode forces `testTimeout` to 15 s and the root config's 60 s does not reach the `component` project — so the loop was racing the very timeout that would otherwise report it. The two sites without an explicit per-test timeout died as `Test timed out in 15000ms` instead of naming what they waited for, and the one carrying `20_000` cleared it by **93 ms**.

Never a false green — purely a diagnosis loss — but it made failures opaque in exactly the tests written to make a failure legible. Now 150 iterations: ~4.5 s real, 150 ms virtual, still ~13× inside the 2000 ms backoff against a measured actual spend of 1–10 ms.

Re-verified after the change, since an audit fix resets the gate: **32/32**, and the re-mint deletion mutant still kills sites 1 and 2 at their own `toHaveBeenCalledTimes(1)`.

---

**Two pre-existing coverage gaps the audit surfaced.** Both survive against `main` too, so this PR neither introduced nor closed them — filing here so they are not lost:

- `PageBlockHost.tsx:3412` — deleting `if (prior === 'loading' || prior === 'ready') return;` leaves **32/32 green here and 391/391 across all 33 AppBlocks browser files**. That guard is the exact mechanism this PR's site-3 finding rests on, and nothing anywhere pins it.
- `PageBlockHost.tsx:3402` — deleting `setReloadNonce((n) => n + 1)` (the iframe remount) leaves 32/32 green. Site 3 is named *"…returns to loading **+ remounts**…"* and its `pollFor('fresh iframe mount', …)` passes without the bump, because `contentWindow` and `data-block-ready === 'false'` both hold regardless. The remount half of that assertion reads as coverage and provides none.

**And the site-3 claim is now measured rather than inferred.** The body argued it from a 53× wall-time move. The audit instrumented which status each retry actually ran from:

| tree | backoff | observed | duration |
|---|---|---|---|
| base | 2000 (shipped) | `['manual:fatal']` | 308 ms |
| base | 50 | **`['auto:fatal', 'manual:timeout']`** | **10718 ms** |
| PR | 50 | `['manual:fatal']` | 364 ms |

The substitution is observed, not inferred — and this discharges the body's own "site 3's converted test does not itself prove it clicked from `fatal`". It does, even at a 40× accelerated backoff.

⚠️ Still not verified: **the original flake was never reproduced naturally.** The base four stayed green 7/7 under synthetic load taking the box from load avg 13 → 140 on 24 cores. The evidence remains the shrunk-backoff proxy control plus the RCA — the structural argument is the load-bearing claim.

PR #4296, opened 2026-08-22
fix(hubs): two things testers hit — a one-way sort, and an unordered source picker

JustMaier:
From the tester thread on `/hubs`. Two reports, both real.

## A hub could open on a sort its own menu hides

`DumbSortFilter` drops **Newest** and **Oldest** for a viewer who cannot view NSFW, and drops **Newest** for one who has NSFW browsing switched off. A hub is created with `sort = "Newest"`.

So a green-domain tester opened a hub reading **Newest**, switched to Most Reactions, and found no way back — his menu only ever listed Most Reactions and Most Comments. Screenshot in the thread shows exactly that.

The predicate that decides which sorts are offered moves out of the component into `isSortAvailable`, so the hub can ask the same question the menu answers. `resolveHubSort` returns the first sort the viewer *is* offered when the stored one is not among them.

Deliberate choices:

- **Resolved on read, never written back.** Regaining NSFW browsing returns the hub to the sort its owner picked. Nothing rewrites their data behind them.
- **Not widening what the menu offers.** Handing a green viewer a Newest sort would defeat the rule this is working around, and the stored sort drives the feed itself, not just the label.
- The media-filter save no longer sends `sort`. It never changed it, and sending the resolved value would have persisted the clamp. Prisma treats the omitted field as "leave alone".

## The source picker returned suggestions in no order

The creator query had **no `orderBy` at all**, so which 25 of your follows came back was down to the planner. Models and collections came back newest-first, which reads as arbitrary in a name list. All three are ordered by name now.

Relationship recency still decides which names are *eligible*, up the `SUGGESTIONS_WINDOW` above the query — only the presentation of the surviving 25 changes.

## Verification

- `TYPECHECK_HEAP_MB=12288 pnpm run typecheck` — clean apart from a pre-existing `packages/civitai-telemetry` error on `main`
- `npx eslint <changed files>` — clean (one pre-existing `no-explicit-any` in `SortFilter`)
- `pnpm run test:unit:run` — 1337 files / 20854 tests, all green
- Mutation check: stubbing the clamp fails 2 of the new tests with `expected 'Newest' to be 'Most Reactions'`, in under a second
- **Browser, `civitai-dev.blue`**: as a moderator all four sorts still list, as a non-moderator with NSFW browsing off a hub stored as Newest now opens on Oldest and every option in the menu is reachable

## Not in this PR

The "having trouble loading more images" report on Most Reactions. Hub feeds pin the Meilisearch path with no DB fallback, so a slow search surfaces as the retry banner — separate work, and Justin has it.

### DIFF
--- a/src/components/Filters/FeedFilters/HubFeedFilters.tsx
+++ b/src/components/Filters/FeedFilters/HubFeedFilters.tsx
@@ -7,13 +7,12 @@ import { useState } from 'react';
 import { FilterButton } from '~/components/Buttons/FilterButton';
 import classes from '~/components/Filters/FeedFilters/FeedFilters.module.scss';
 import { SortFilter } from '~/components/Filters/SortFilter';
+import { buildHubFilterSave } from '~/components/Hubs/hub-filter-save';
 import { useHubSort } from '~/components/Hubs/useHubSort';
 import { hubExcludedFilterKeys } from '~/components/Image/Filters/media-filter-keys';
 import { MediaFiltersDropdown } from '~/components/Image/Filters/MediaFiltersDropdown';
 import type { HubSort } from '~/server/schema/user-hub.schema';
-import { hubFeedFiltersSchema, hubLimits, hubSortSchema } from '~/server/schema/user-hub.schema';
-import { MetricTimeframe } from '~/shared/utils/prisma/enums';
-import type { MediaType } from '~/shared/utils/prisma/enums';
+import { hubLimits, hubSortSchema } from '~/server/schema/user-hub.schema';
 import { showErrorNotification } from '~/utils/notifications';
 import { trpc } from '~/utils/trpc';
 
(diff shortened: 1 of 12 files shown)


JustMaier:
Updated after Justin's call on the fallback.

**Most Reactions is the fallback everywhere, not the next offered sort.** My first version fell through the sort list in order, which landed a viewer with NSFW browsing off on **Oldest**. That is wrong for the same reason Newest is: Oldest is withheld wherever Newest is, and for the same unrated-image reason, so it is not a safe substitute merely because that particular viewer is still offered it.

**A new hub is no longer created sorted Newest for a viewer who cannot pick Newest.** Resolving on read left the stored value as `Newest`, so the hub stayed one click away from the sort it should not have had. `HubUpsertModal` now sends `defaultHubSort(...)` on create; read-time resolution stays as the backstop for hubs that already exist, and for a viewer whose situation changes later.

Re-verified in the browser on `civitai-dev.blue`: signed in as a non-moderator with NSFW browsing off, created a hub through the modal, and the row stored `sort = 'Most Reactions'`. Same flow before the change stored `Newest`.

PR #4297, opened 2026-08-22
feat(hubs): add a model or a creator to a hub from its own page

JustMaier:
Asked for in the tester thread: an **Add to hub** option next to Save, on model pages and creator pages. Justin confirmed the scope, including collections later once collection sources can be switched on.

## What it does

A menu item beside Save on a model page, and in the creator context menu, both gated on `userHubs`. It opens a modal listing your hubs with a tick for the ones that already hold this target, a source count against the per-hub cap, and a box to name a new hub that gets created with this source already in it.

## Why two new procedures rather than `upsert`

`upsert` **replaces** a hub's whole source list. A model page has not loaded the hub, so using it would mean fetching every source for every hub first, and any list it did send would clobber sources added since it fetched. That is the same clobbering the `sources`-is-optional comment on the schema already warns about.

So the modal addresses a source by what it points at:

- **`addSource`** returns `{ added: false }` when the hub already holds that target rather than throwing, so a double-tick is not an error. Refuses at `hubLimits.sourcesPerHub`.
- **`removeSource`** deletes only the named target in that hub.

Both read the hub scoped by `userId`, so someone else's hub id is a not-found and never a write.

The new source lands **one past the highest existing index**, not at `sources.length` — removing a source leaves a gap, so appending at the count collides with a row that is still there. There is a test for exactly that.

Membership is read off `getAll`, which already returns each hub's sources, so no new query.

## Verification

- `TYPECHECK_HEAP_MB=12288 pnpm run typecheck` — clean apart from the pre-existing `packages/civitai-telemetry` error on `main`
- `npx eslint <changed files>` — clean (3 pre-existing `no-unused-vars` in the model page's `getServerSideProps`)
- `pnpm run test:unit:run` — 1337 files / 20865 tests, all green
- The router gate test derives its procedure list from the router, so both new procedures are covered by it rather than added to a hand-written list
- **Browser, `civitai-dev.blue`**: ticked a hub from a model page and confirmed the row in the database with the right type and alias, unticked it and confirmed it was gone; same from a creator page, which wrote `type: User, targetId: 3, alias: Maxfield`

### DIFF
--- a/src/components/Dialog/triggers/add-to-hub.ts
+++ b/src/components/Dialog/triggers/add-to-hub.ts
@@ -1,0 +1,6 @@
+import dynamic from 'next/dynamic';
+import { createDialogTrigger } from '~/components/Dialog/dialogStore';
+
+const AddToHubModal = dynamic(() => import('~/components/Hubs/AddToHubModal'), { ssr: false });
+
+export const openAddToHubModal = createDialogTrigger(AddToHubModal);
--- a/src/components/Hubs/AddToHubModal.tsx
+++ b/src/components/Hubs/AddToHubModal.tsx
@@ -76,8 +76,11 @@ export default function AddToHubModal({ source }: { source: HubSourceTarget }) {
           ) : (
             <Stack gap="xs">
               {hubs.map((hub) => {
+                // Membership here means "this hub shows me that source", so a row the
+                // owner switched off in the rail reads as unticked: the hub is not
+                // showing it, and ticking is what turns it back on.
                 const checked = hub.sources.some(
-                  (s) => s.type === source.type && s.targetId === source.targetId
+                  (s) => s.type === source.type && s.targetId === source.targetId && s.enabled
                 );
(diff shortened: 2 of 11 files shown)


JustMaier:
**Merge #4296 before this one.**

They are independent branches, both off `main`, deliberately not stacked. But the create path in this modal (`AddToHubModal` → "New hub") calls `userHub.upsert` without a `sort`, and what that stores depends on which lands first:

- **#4296 first** — the server default is `MostReactions` and read-time resolution exists, so a hub created from here is correct for every viewer. Nothing further needed.
- **This one first** — a hub created from here is stored `Newest` for everyone, including viewers whose sort menu will not offer it. That is the exact bug #4296 fixes, and it would be reintroduced through a new door until #4296 merges. Self-healing once it does, since resolution is on read.

No code change is needed in either PR for this; it is purely a merge-order note. Flagging it rather than stacking, since a squash-merged parent does not retarget its child.

PR #4301, opened 2026-08-22
fix(auth): bust the shaped session cache for every session-carried field it writes

### LINKED ISSUE #4298: Profile picture change is invisible to the session for up to 4h — the shaped session-user cache is never invalidated
Changing a profile picture updates the database immediately, but the session's `user.image` keeps serving the **previous** avatar for up to 4 hours. Combined with #4272 (the previous image is hard-deleted the moment the new one is saved), that produces a guaranteed window in which every session-derived avatar renders a **404**.

## Measured

On a preview environment, after changing the profile picture through the account page:

| source | value |
|---|---|
| profile page (reads live) | the **new** image |
| session endpoint `user.image` | the **previous** image |
| account-switcher roster | the **previous** image (correctly mirroring the session) |

The profile-picture write itself fully succeeded — `userProfile.update` and `user.update` both returned 200, and the new image reached ingestion state `Scanned`. The divergence persisted well past ingestion, so it is not a propagation delay.

## Mechanism

`user.image` is served from the hub's shaped session-user cache, which the producer writes with a 4h TTL:

```
apps/auth/src/lib/server/auth/session-producer.ts
  const SESSION_TTL = 4 * 60 * 60; // 4h
```

The only function that busts that cache is `invalidateSessionUser` in the same file. Its callers are all inside the auth app — membership changes, role changes, and the identity endpoint. **No main-app path calls it**, including the profile-picture update.

`refreshSession` (`src/server/auth/session-invalidation.ts`) is called on profile update, but it marks tokens for refresh and signals active sessions; it does not wipe the shaped-user entry. The file states this as intended behaviour:

```
// The shared session:data2 cache is NOT wiped — its 4h TTL bounds data staleness
// (see the hub handler).
```

Bounding staleness by TTL is a defensible choice for slow-moving fields like roles and permissions. It is not safe for the avatar, because the avatar's underlying object is **deleted synchronously** on replacement (#4272). Stale-but-present would be invisible to users; stale-and-deleted is a broken image.

## Impact

For up to 4h after changing a profile picture, anything rendering the avatar from the session shows the old url, which now 404s. That includes the account-switcher entry in the user menu.

This is the actual cause of the originally-reported symptom — "profile page looks right, the avatar in the dropdown 404s". Both #4260 and #4265 were diagnosed as roster bugs. They are real and worth fixing on their own terms, but neither can resolve this: the roster faithfully mirrors the session, and the session is what is stale.

## Suggested direction

Either:

1. have the profile-picture update path invalidate the hub's shaped session-user entry (call the existing `invalidateSessionUser` for that user), so the next session read re-produces; or
2. stop sourcing the avatar from the TTL-cached session shape, and read it from the profile-picture cache that already exists for exactly this purpose — note `profilePictureCache` in `src/server/redis/caches.ts` is deliberately configured `staleWhileRevalidate: false` with the comment *"To avoid delay in creator seeing their new profile picture"*, so this hazard is already understood one layer down.

Option 1 is the smaller change. Worth checking which other session fields users can edit and expect to see update immediately — the same 4h bound applies to all of them, and the avatar is simply the one where staleness is visibly broken rather than merely out of date.

ZacxDev:
Closes #4298 — but not in the way the issue describes. Please read the first section before reviewing the diff.

## The reported mechanism is not what the code does

#4298 says the profile-picture update path never busts the hub's shaped session-user entry, so `user.image` stays stale for up to 4h. The chain is actually wired, and has been since the hub cutover:

```
updateUserHandler (user.update)
  └─ refreshSession
       └─ updateSessionState          ← unconditional, on both 'refresh' and 'invalid'
            └─ clearSessionCache
                 └─ sessionClient.invalidate(userId)
                      └─ POST {hub}/api/auth/identity  { userId }
                           └─ invalidateSessionUser → del session:data2:{userId}
```

The issue's evidence — "the only callers of `invalidateSessionUser` are inside the auth app; no main-app path calls it" — is true of the **function name** and false of the **effect**. The main app reaches it over HTTP, not by import, so grepping `src/` for the identifier finds nothing and reads as "nothing busts the cache". I made the same inference before tracing it.

I verified this rather than reasoning about it. The four avatar tests in this PR are **green against pristine `origin/main`**, and a mutation that deletes the bust from `clearSessionCache` turns all four red — so they can see the described defect, and the described defect is not present.

The second piece of evidence in the issue — the comment reading *"The shared session:data2 cache is NOT wiped — its 4h TTL bounds data staleness"* — is scoped to `invalidateAllSessions`, the **mass** path, where it is true and deliberate (a cluster-wide SCAN+del is not worth the freshness). It reads as a statement about the file's per-user paths, which do wipe it. That comment is corrected here; it is the proximate cause of the misdiagnosis.

### On the measurement in the issue

The measurements were taken on a preview environment, and the divergence there has a different cause than a TTL. `getServerAuthSession` already documents it in-tree (`src/server/auth/get-server-auth-session.ts`, the `getLegacySession` preview-fallback comment): a preview runs against a different database than the one the hub resolves identity from. So the profile page (preview's own DB, freshly written) and the session (hub-resolved) read from **two different stores** — they do not converge after 4h, or ever. That fits every row of the issue's table, including "the divergence persisted well past ingestion", better than TTL staleness does, and no cache bust can close it.

That does not make the underlying user report invalid — it means the preview measurement is not evidence for the production mechanism it was read as.

## What was actually broken

Enumerating the shaped SessionUser's fields (`apps/auth/.../session-shape.ts`) against their writers — rather than re-reading the avatar path — found two live instances of the defect class the issue names: **a handler whose bust gate is narrower than the set of session-carried fields it writes.**

**1. `user.setSettings` — `allowAds` (user-visible)**

The bust fired only when `isEarlyAdopter` changed, on the stated premise that it was *"the one settings key PROJECTED ONTO THE SESSION"*. It is not. The hub's `settingsSchema` reads three keys out of `User.settings` — `allowAds`, `redBrowsingLevel`, `isEarlyAdopter` — and `allowAds` is writable through this endpoint's own schema. Turning ads off therefore left the session, and every Flipt context built from it, serving `allowAds: true` for up to 4h.

The gate is now a named `SESSION_PROJECTED_SETTING_KEYS` set, so adding a projected key is one edit rather than a silent re-introduction. `redBrowsingLevel` is deliberately excluded — this endpoint cannot write it; `updateContentSettings` does, and busts on its own.

Reachability note: the live ads UI routes through `updateContentSettings` (which busts), so this was an API-reachable hole rather than a broken button in today's UI. It is still a real gap on a `UserWrite`-scoped mutation.

**2. `user.completeOnboarding` — Profile step**

The Profile step writes `username` and `email` **unconditionally**, but the bust was gated on the onboarding bitmask advancing. Re-submitting a step whose bit is already set changed the username in the database and left the old one in the session for the rest of its 4h TTL.

## Which option I chose, and what I rejected

The issue offered four directions. Given the above:

- **Option 2/3 (hub endpoint / shared package) is already the implemented architecture** — `@civitai/auth`'s `createSessionClient.invalidate` is the shared implementation, and `POST /api/auth/identity` is the hub endpoint. Nothing to build.
- **Option 1 (main app writes the hub-owned key directly)** — rejected as the primary path. It is already present as the *fallback* inside `clearSessionCache`, which is the right shape: the hub owns the key, the main app asks; the direct `del` only covers a hub blip, where busting a key the hub re-produces is safe from either side. Promoting it to primary would duplicate a hub-owned namespace in a second app for no gain. I did add a test pinning that fallback's hand-built string against the shared namespace constant, since it is a second spelling that can drift silently.
- **Option 4 (source the avatar from `profilePictureCache` instead)** — rejected. It forks `user.image` away from the single-producer architecture the hub cutover established, so the avatar would follow different invalidation rules from every other session field, and the 4h bound would remain wrong for the fields that genuinely need busting. It treats the symptom on one field.

**So the TTL-bounded staleness is left intact** — it is correct for slow-moving fields, and it is not what was broken. What is fixed is two gates that were narrower than their handlers' writes.

## Tests — the seam, not one side of it

Every existing suite is *structurally* blind to this relationship: `src/__tests__/setup.ts` globally stubs `~/server/auth/session-invalidation` down to a no-op `refreshSession`. A test written against that stub can assert the **call** but cannot distinguish a bust that addresses the right key from one that busts nothing, or the wrong key.

The new file `vi.unmock`s it and drives the real chain — `refreshSession → updateSessionState → clearSessionCache` — against a fake shared keyspace, asserting the **absence of the entry** rather than a call. The hub hop is stood in for by a fake whose key is derived from the shared `REDIS_KEYS.USER.SESSION` namespace, so a namespace drift between the two apps fails here.

### Matrix

`npx vitest run --project 'unit*' src/server/controllers/__tests__/user.controller.session-avatar-bust.test.ts` — **11 tests executed** both times:

| | result |
|---|---|
| pristine `origin/main` | **2 failed** \| 9 passed |
| this branch | **11 passed** |

The two reds at base are the two real gaps (`allowAds`; onboarding Profile re-submit), each `AssertionError: expected true to be false` — i.e. the session entry survived the write.

**Honest labelling:** the four *avatar* tests are green at base. They are an **invariant guard**, not a regression test — the avatar defect is not present, and mutation testing (below) is what demonstrates they would catch it. I have not counted them as regression coverage.

### Mutation checks

| mutation | result |
|---|---|
| remove the `session:data2` bust from `clearSessionCache` entirely (reproduces the defect #4298 describes) | **all 4** avatar tests red — 3 on `expected true to be false` (entry survived), 1 on the bust never firing |
| keep the hub call, drift **only** the fallback key namespace | **exactly 1** red — the test that pins the fallback. The other 3 stay green |

The second is the narrow one: it isolates a single expression and kills exactly the assertion that covers it, so the tests are independently reachable rather than all dying to one upstream failure.

### Wider verification

- `src/server/` unit suite: **14495 passed / 0 failed** (923 files, 1 skipped)
- `pnpm typecheck`: **0 type errors**
- `pnpm eslint` on the three changed files: **0 errors** (4 pre-existing warnings, all on lines this PR does not touch)

Note `isolate: true` is in force for the unit project, so the `vi.unmock` cannot leak into sibling files — that was checked, not assumed.

## Other session fields with the same window — reported, not fixed

I enumerated the full shaped SessionUser against its writers. Everything else is covered, with one low-severity exception:

- **`referral`** — `createUserReferral` at `src/server/auth/login-side-effects.ts:42` creates the `UserReferral` row with no bust. Only matters if the session entry was produced before the referral row landed. Not fixed here; happy to, in a follow-up, if you think it's worth the round trip.

Confirmed covered: `username`, `image`/`profilePicture`, `email`/`emailVerified`, `autoplayGifs`, `filePreferences`, `leaderboardShowcase` (all via `updateUserHandler`); `showNsfw`/`blurNsfw`/`browsingLevel`/`redBrowsingLevel` (via `updateBrowsingMode` and `updateContentSettings`); `onboarding`; `name`, `isModerator`, `muted`, `bannedAt`, `deletedAt`, `customerId`, `paddleCustomerId`, `tier`/`subscriptions` (moderation, billing and subscription paths, all bust). `roles` has no writer in `src/` — it is hub-owned.

`updateUserProfileHandler` / `updateUserProfile` write **nothing** on the shaped SessionUser (only `publicSettings.creatorCardStatsPreferences`, `UserLink` and `UserProfile` rows — the `User.image` / `leaderboardShowcase` writes there are commented out and moved to `updateUserHandler`). Its lack of a bust is correct, not a gap.

## Scope

Does not touch `src/components/CivitaiWrapped/AccountProvider.tsx`, so no overlap with #4260 or #4274. Branched directly off `main`, not stacked.

The roster in the account switcher resolves each linked account through the same shared cache (`getOrProduceSessionUser`), so it mirrors the session by construction — consistent with the issue's observation, and fixed by the same bust wherever the bust was missing.

### DIFF
--- a/src/server/auth/session-invalidation.ts
+++ b/src/server/auth/session-invalidation.ts
@@ -94,6 +94,10 @@ async function updateSessionState(
 
   let chunksWritten = 0;
   let chunksTotal = 0;
+  // This is where the hub's shaped session-user entry (`session:data2:{userId}`, 4h TTL) gets busted —
+  // `clearSessionCache` asks the hub to drop it, so the next read re-produces from the DB. It runs on
+  // BOTH the 'refresh' and 'invalid' paths, and unconditionally (not gated on the token scan above), so
+  // every per-user caller of refreshSession/invalidateSession gets a fresh shape on the next read.
   await clearSessionCache(userId);
   if (userTokens.length > 0) {
     // Atomic multi-field set+TTL — single EVAL replaces the sequential
@@ -151,6 +155,24 @@ async function sendSessionSignal(userId: number, type: 'refresh' | 'invalid') {
   }
 }
 
+/**
+ * Mark a user's tokens for refresh, bust their cached session state, and signal active sessions.
+ *
+ * 🔴 This DOES bust the hub's shaped session-user entry (`session:data2:{userId}`) — via
+ * `updateSessionState` → `clearSessionCache` → `sessionClient.invalidate`, which POSTs the hub's
+ * `/api/auth/identity` and lands in the hub's own `invalidateSessionUser`. So a caller that edits a
+ * field carried on the SessionUser shape (username, image/profilePicture, onboarding, settings, …)
+ * only has to call this; the next session read re-produces from the DB rather than serving the
+ * cached shape for the remainder of its 4h TTL.
+ *
+ * The main app reaches the hub's `invalidateSessionUser` over HTTP, not by import — grepping for that
+ * function name in `src/` finds nothing and reads as "no main-app path busts the cache", which is how
+ * this got misdiagnosed in #4298. The seam is pinned by
+ * `src/server/controllers/__tests__/user.controller.session-avatar-bust.test.ts`.
+ *
+ * Contrast `invalidateAllSessions` below, which deliberately does NOT wipe the shape — that exemption
+ * is specific to the MASS path and does not apply here.
+ */
 export async function refreshSession(
   userId: number,
   { sendSignal = true, caller = 'unspecified' }: RefreshSessionOptions = {}
@@ -198,7 +220,9 @@ export async function invalidateSession(
 export async function invalidateAllSessions() {
   // Mass invalidation is hub-owned: set the global revocation cutoff (the shared SESSION.ALL marker the hub
   // registry owns) through the hub rather than writing it directly — revokes every token signed before now.
-  // The shared session:data2 cache is NOT wiped — its 4h TTL bounds data staleness (see the hub handler).
+  // 🔴 THIS MASS PATH ONLY: the shared session:data2 cache is NOT wiped, because a cluster-wide SCAN+del
+  // costs far more than the freshness it buys and its 4h TTL bounds the staleness. The PER-USER paths above
+  // (refreshSession / invalidateSession) DO wipe it — do not read this sentence as a statement about them.
   await sessionClient.invalidateAll();
   log(`Scheduling session refresh for all users`);
 }
--- a/src/server/controllers/__tests__/user.controller.session-avatar-bust.test.ts
+++ b/src/server/controllers/__tests__/user.controller.session-avatar-bust.test.ts
@@ -1,0 +1,282 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type * as UserService from '~/server/services/user.service';
+
+/**
+ * SEAM test for #4298 — "changing the avatar leaves the session serving the old (now deleted) url".
+ *
+ * The relationship under test spans two applications and therefore no single-surface test can see it:
+ *
+ *   main app                                        auth hub
+ *   ────────                                        ────────
+ *   updateUserHandler (avatar write)
+ *     └─ refreshSession
+ *          └─ clearSessionCache
+ *               └─ sessionClient.invalidate(userId) ──▶ invalidateSessionUser
+ *                                                        └─ del `session:data2:{userId}`
+ *                                                             ▲
+ *                          produceSessionUser writes ─────────┘  (4h TTL)
+ *
+ * Every OTHER suite in this repo is structurally blind to it, because `src/__tests__/setup.ts`
+ * globally mocks `~/server/auth/session-invalidation` down to a no-op `refreshSession`. A test
+ * written against that stub can only assert that the CALL happens — it cannot distinguish a call
+ * that busts the right key from one that busts nothing, or the wrong key. That is exactly the
+ * failure mode #4298 describes, so this file unmocks it and asserts the EFFECT instead.
+ *
+ * What is deliberately real here: the whole main-app chain (`refreshSession` → `updateSessionState`
+ * → `clearSessionCache`), and the KEY both sides address, derived on each side from the single
+ * shared `REDIS_KEYS.USER.SESSION` namespace in `@civitai/redis`. The hub's HTTP hop is stood in
+ * for by `hubInvalidate`, which deletes from the same fake keyspace using the hub's own key
+ * derivation — so a namespace drift between the two apps fails this test rather than shipping.
+ */
+
+// Defeat the global stub from `src/__tests__/setup.ts`. Without this line the entire chain below is
+// replaced by `vi.fn()` and every assertion here passes vacuously.
+vi.unmock('~/server/auth/session-invalidation');
+
+const {
+  sharedCache,
+  hubInvalidate,
+  mockUpdateUserById,
+  mockGetUserById,
+  mockIngestImage,
+  mockDeleteImageById,
+  mockGetUserSettings,
+  mockPatchUserSettings,
+} = vi.hoisted(() => ({
+  sharedCache: new Map<string, unknown>(),
+  hubInvalidate: vi.fn(),
+  mockUpdateUserById: vi.fn(),
+  mockGetUserById: vi.fn(),
+  mockIngestImage: vi.fn(),
+  mockDeleteImageById: vi.fn(),
+  mockGetUserSettings: vi.fn(),
+  mockPatchUserSettings: vi.fn(),
+}));
+
+// The hub hop. `clearSessionCache` calls this first and only falls back to a direct `redis.del` when
+// it rejects — both arms are exercised below.
+vi.mock('~/server/auth/session-client', () => ({
+  sessionClient: { invalidate: hubInvalidate },
+}));
+vi.mock('~/server/services/orchestrator/civitai', () => ({
+  invalidateCivitaiUser: vi.fn().mockResolvedValue(undefined),
+}));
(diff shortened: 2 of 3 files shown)


ZacxDev:
Adversarial audit passed — no deploy-blocking findings, both fixed defects independently reproduced red-at-base (11 tests executed, 2 failed at `a05a5c1017`, 11 passed at head), and the mutation claims survived re-derivation. One correction to the description before this lands.

## The sweep is not exhaustive — `permissions` is missing

The body's "everything else is covered" will be read as a completed enumeration, and it isn't one. `permissions` is on the shaped SessionUser (`apps/auth/src/lib/server/auth/session-shape.ts`) and `src/server/services/feature-flags.service.ts` branches on it, but three writers never bust:

- `src/server/routers/user.router.ts` — `addSystemPermission('creatorsProgram', …)`, **user-facing tRPC**
- `src/pages/api/mod/enable-stripe-connect.ts`
- `src/pages/api/mod/enable-tipalti.ts`

`src/pages/api/admin/permission.ts` *does* call `refreshSession`, so this is a known pattern applied inconsistently rather than a deliberate exemption.

Other same-class writers outside the sweep, all pre-existing — highest impact first:

- `src/server/services/referral.service.ts` — the hourly `advanceReferralSubscriptions` changes subscription status / product, which moves `SessionUser.tier`, with no bust
- `src/server/services/user.service.ts` — `restoreUser` writes `deletedAt`, `username`, `email`, `meta`
- `src/server/controllers/paddle.controller.ts` — user-facing `cancelEmail` writes subscription status
- `src/pages/api/admin/temp/migrate-nsfwLevels.ts` calls `invalidateAllSessions()`, which — per the very contract this PR clarifies — does **not** wipe the shaped-user entry

Suggest softening the claim to name what was swept and what wasn't. Tracked as a follow-up; not a blocker for this change.

## Two smaller notes

**The namespace-drift test doesn't test drift.** It derives `hubKey` from the same constant `session-cache.ts` uses, so it pins one app's literal against itself and would stay green if the two apps diverged. One line fixes it: import the hub's own derivation (`sessionCacheKey` from `@civitai/auth`) rather than re-deriving.

**`SESSION_PROJECTED_SETTING_KEYS` is kept in sync with the hub by a prose comment**, which is the exact failure mode the PR says it wants to prevent. There's precedent for a mechanical guard next door — `feature-flags.early-adopter.seam.test.ts` already imports `shapeSessionUser` directly from the auth app. Flipping each settings key through the real `shapeSessionUser` and asserting the shaped output changes iff the key is in the set would fail on drift from either side.

## One risk worth knowing

If the cache redis is unreachable, `clearSessionCache`'s fallback rejects, `refreshSession` throws, and both new call sites are un-`.catch`ed inside `try {} catch { throwDbError }` — so the mutation 500s *after* the DB write already succeeded. `updateContentSettings` `.catch`es for exactly this. Pre-existing asymmetry, but this PR widens the set of requests that can reach it; a `.catch(handleLogError)` would make it symmetric.

PR #4303, opened 2026-08-22
test(services): move the whole n-z slice onto the canonical shared mocks (31 files)

ZacxDev:
Completes the services migration started in #3959 / #3973. **The n–z slice is now empty**: 31 files converted, and the only 3 remaining in `src/server/services/__tests__/` are a–m.

```
canonical allowlist   209 → 178 files,  270 → 223 sites
services slice        34 → 3   (n–z: 31 → 0)
```

Done as three independent batches, then **merged and re-verified as one tree** — the three branches each regenerated the allowlist from a `209` base, so each carried a different correct answer to the same generated file. Conflicts were resolved by **regenerating**, never hand-merging, and the final regeneration is idempotent (a second run changes nothing).

**Merged-tree run: `314 files, 4444 passed | 8 skipped`** — identical to each batch's own figure, so no test was lost across the merge.

## Three canonical specifiers, not one

`~/server/db/client`, `~/server/redis/client`, `~/server/logging/client` — and `files[]` membership is a boolean **OR** across them. Converting only the db mock leaves a file listed and achieves nothing. This cost a redo on an earlier pair (#4293) and every file here was measured against all three first.

## The trap, reproduced rather than avoided

`paid-access.service.test.ts` was known to break when a codemod drops its redis mock. Rather than working around it, the conversion was run first to confirm: **8 of 52 failed**, exactly as expected. Its mock supplied `REDIS_KEYS.CACHES.PAID_ACCESS_CAP_TIER: 'test:cap-tier'` and 8 assertions filtered a cache stub on that literal. Fixed by naming the real constant.

## Findings worth keeping

- **Drifted key tables.** Several files declared local `REDIS_*_KEYS` that did not match production — e.g. `'system:moderated-tags'` against the real `packed:system:moderated_tags`. Inert, because nothing asserted a key — which is the finding: those files were never asserting what they appeared to. The canonical registration puts the real constants in force.
- **A false comment corrected.** `resourceReview` claimed its client choice "is untested" because it aliased both clients. Not true — `resourceReview.findUnique` routes to `dbWrite`, and routing it to `dbRead` makes every block-refusal test stop firing. The whole-module-grep trap is live here: the same call is spelled on `dbRead` in another service, but only for an `entityType` these paths never pass.
- **A local re-implementation dropped** — `signals.service` carried its own `safeError`; the real one now runs.
- **`vae-files-cross-version` was `importOriginal`-ing the app shim**, which constructs real clients at module scope. Spreading the package instead is strictly safer.
- **Behaviour change, flagged in-file:** `user-challenge-flag-gate` bound `dbWrite: {}`, so any write would have thrown. Canonical `dbWrite` answers instead. Nothing on the validation path writes, but a stray future write is no longer loud there.

## Verification

Per-file collected counts before and after, **unchanged and non-zero** in all 31 — the check that matters, because a stale wholesale mock fails to *import* and reports `Tests no tests`, which reads as "nothing to see" rather than a failure. Counted from the JSON reporter's `assertionResults`, never an exit code.

Seam probes rather than green-as-evidence: routing an entry point to the wrong client reds the file naming the service line; the `sysredis-soft` guards still go red on a value (1 of 3 in 3 ms, 1 of 9 in 7 ms) rather than hanging; `residual-mocks.mjs` returned **0 under test and non-zero on a deliberate control**, cross-checked with the relative-path-aware `mockPattern` since that script is tilde-only.

## 🔴 `pnpm typecheck` cannot see any of these files — do not read it as coverage

The root `tsconfig.json` excludes `src/**/__tests__/**`. A deliberate `const x: number = "s"` planted in one of these files still yields `typecheck: OK — 0 type errors`, measured twice independently. The instrument that *can* see them is `scripts/typecheck.mjs -p tsconfig.tests.json`, and it reports **923 → 922** here, the single delta being an error one batch fixed.

⚠️ That gate is **already failing on `main`** (~922 errors / 169 files against a baseline of 784 / 140) for unrelated reasons, confirmed by diffing its output with and without these changes. Pre-existing; not chased here.

## Not verified

- **No `--no-isolate` run.** Everything was measured under the default isolated regime, so the migration's actual payoff — cross-file interference — is untested by this work. That is the separate flip gate.
- **No whole-suite run** beyond `src/server/services/__tests__/` plus the mock-guard suites.
- `remix-gallery`'s `$queryRaw` and `placement.findMany` genuinely run on both clients against one ordered call list, so both canonical nodes forward to a single spy — a cross-client move of those two stays invisible. Unchanged from before, now documented in the file.
- Two files (`training.orchestrator-error-mapping`, `update-post-image-hidemeta-bust`) had no assertion touching redis, so no mutation could discriminate; their conversion rests on unchanged collected counts alone, which is weaker than the other 29.


ZacxDev:
📌 **Correction to a measurement in the PR body — posted publicly, since it was used as evidence and it is wrong in the unsafe direction.**

The body says the typecheck-tests gate goes **923 → 922, "the single delta being an error one batch fixed"**. Measured on identical-provenance trees, it is **923 → 887 (−36)**. I re-ran the gate on this HEAD myself: **887**.

**Only 3 of those 36 are genuine fixes** (`update-user-profile.sql.service.test.ts` TS2339 `'raw'`; `reaction-stale-replica-delete.test.ts:60,61` TS2352).

🔴 **The other 33 are in `remix-gallery.service.test.ts`, and they did not get fixed — they stopped being checked.** `HybridNode` is typed `Mock<(...args: any[]) => any>`, so `mock.calls` widens from a zero-length tuple to `any[][]`. The TS2532 / TS2493 / TS18048 errors on positional `…mock.calls[0][0]` reads vanish because those reads are now **unchecked**, not because they became safe. Confirmed on this HEAD: `remix-gallery` is down to **4** errors from 37.

Nothing gates on `tsconfig.tests.json` today, so this blocks nothing — but a stated number moving 35 in the direction of *less* checking should not stand uncorrected in a PR whose argument rests on measurement.

---

**Also fixed, in `3e9ee9b`:** two file comments that understated the empty-client blind spot.

- **`user-challenge-flag-gate.test.ts`** said the canonical `dbWrite` means "a stray write would no longer be loud". It is broader than that: the old `dbWrite: {}` threw on *any* dbWrite access, so it also caught a **misrouted read**. Mutating `collection.service.ts`'s `dbRead.challenge.findFirst` to `dbWrite` kills **3 of 6** cases against the pre-conversion file and is caught by **none** of them now.
- **`paid-access.service.test.ts`** has the identical shape and was not disclosed at all — it bound `dbRead: {}`, and the service really does use the replica (`:165` `fromWrite ? dbWrite : dbRead`, `:252`, `:370`). No current case reaches those lines, so measured kill-power is unchanged, but the latent surface is wider than it was.

Both files re-run after the edit: **58 collected, 58 passed**.

---

**What the audit confirmed rather than found**, worth recording since it is the part that matters:

- **Assertion parity is exact** — 671 assertions on both sides, identical test-name sets, identical `expect(` counts, in all 31 files. No test lost power at the case level.
- **The de-aliasing GAINED kill power in four places** — four mutations are green on `main` and red here (`resourceReview.service.ts:327`, `tag.service.ts:711`, `user-profile.service.ts:242`, `user.service.ts:2231`).
- **The allowlist is reproducible byte-for-byte** (sha256 match, "178 → 178, 0 migrated"), delta is exactly the 31 removals, 0 added, nothing hand-edited.
- **The merge is clean** — all 31 blobs byte-identical to their batch branch, none lost or duplicated.
- **No negative assertion is bound to the wrong client** — every `not.toHaveBeenCalled` / `toHaveBeenCalledTimes(0)` / `toHaveLength(0)` in the 31 was checked against the entry point the test drives, not a whole-module grep.

**One follow-up, not blocking:** 10 files now arm shared nodes at **module scope**, which survives `vi.clearAllMocks()` (what they all use) but not `vi.resetAllMocks()` or `resetHybridNodes()` — and `referral.service.test.ts:61` calls the worker-global `resetHybridNodes()` per test. No file today does both, so nothing is broken, but the rule is written nowhere and nothing enforces it. A single `vi.resetAllMocks()` added to a `beforeEach` would gut a file's fixtures.

PR #4307, opened 2026-08-22
chore(hubs): one invalidation set, one visibility rule, and scope assertions

JustMaier:
Three cleanups out of the tester round on `/hubs`. Nothing here changes what a
tester sees except the stale feed noted below, so it does not touch what the
replies in the feedback thread promise.

### One invalidation set

`getAll` + `getById({id})` + `image.getInfinite({hubId})` was written out in
`HubSourcePanel`, `HubUpsertModal` and `AddToHubModal`, and had already
diverged — the upsert modal was missing the feed key, so creating a hub with
sources from it landed on a feed that had already been cached empty.

Now `useInvalidateHub` in a new `src/components/Hubs/hub.utils.ts`, which is
the file collections have had all along as `collection.utils.ts`.

`AddToHubModal` also declared its own `{ type, targetId, alias }` — the fourth
inline copy of that shape. It takes `Omit<AddUserHubSourceInput, 'hubId'>` now.

### One visibility rule for the source picker

`getHubSourceSuggestions` filtered candidate models on `deletedAt: null` alone;
`resolveHubSourceFromUrl` used `visibleModel(userId, isModerator)`. A bookmark
or a bell outlives the model going private or back to draft, so the picker
offered by name what the paste-a-link path refuses. Both share `visibleModel`
now, and the router passes `isModerator` so a moderator sees one list, not two.

This is a consistency bug, not a leak: the suggestion carries a name the viewer
had already bookmarked or followed, and the feed applies its own visibility.

### Scope is asserted, not assumed

Hubs have **no scope of their own** — every procedure rides on
`TokenScope.UserRead` / `UserWrite`. The existing gate test proved the feature
flag only: its caller carries `TokenScope.Full`, and `enforceTokenScope` skips
the check outright for Full, so deleting `.meta({ requiredScope })` from a hub
procedure was invisible to the whole suite.

Added a second describe that calls every procedure with a token scoped to
exactly what it declares (must pass) and to the other user scope (must be
FORBIDDEN), from a map written out by hand — a map read off `.meta()` would
assert the router against itself.

### Verified

Each new assertion was checked against the mutation it is supposed to catch:

| mutation | result |
| --- | --- |
| drop `.meta({ requiredScope })` from `getAll` | 1 failed / 37 passed |
| downgrade `upsert` to `UserRead` | 2 failed / 36 passed |
| revert the suggestions visibility filter | 1 failed |

Full unit suite green (20924 passed, 0 failed, 249s), typecheck clean apart
from the pre-existing `packages/civitai-telemetry` error on `main`, eslint and
prettier clean.

### Noted, not fixed

`addUserHubSource` enforces the per-hub source cap as a read-check-write with
no constraint behind it. Both reads go through the writer, so it takes two
genuinely concurrent adds to exceed the cap by one, and the excess row is
harmless. Left alone deliberately.

Tickets: 868kv6n05, 868kv6mzy, 868kv6n0e.

### DIFF
--- a/src/components/Filters/FeedFilters/HubFeedFilters.tsx
+++ b/src/components/Filters/FeedFilters/HubFeedFilters.tsx
@@ -8,6 +8,7 @@ import { FilterButton } from '~/components/Buttons/FilterButton';
 import classes from '~/components/Filters/FeedFilters/FeedFilters.module.scss';
 import { SortFilter } from '~/components/Filters/SortFilter';
 import { buildHubFilterSave } from '~/components/Hubs/hub-filter-save';
+import { useInvalidateHub } from '~/components/Hubs/hub.utils';
 import { useHubSort } from '~/components/Hubs/useHubSort';
 import { hubExcludedFilterKeys } from '~/components/Image/Filters/media-filter-keys';
 import { MediaFiltersDropdown } from '~/components/Image/Filters/MediaFiltersDropdown';
@@ -41,7 +42,7 @@ export function HubFeedFilters({ ...groupProps }: GroupProps) {
   const router = useRouter();
   const hubId = Number(router.query.id);
   const [sourcesOpen, setSourcesOpen] = useState(false);
-  const utils = trpc.useUtils();
+  const invalidateHub = useInvalidateHub();
 
   const { data: hub } = trpc.userHub.getById.useQuery(
     { id: hubId },
@@ -50,12 +51,7 @@ export function HubFeedFilters({ ...groupProps }: GroupProps) {
   const sort = useHubSort(hub?.sort);
 
   const upsert = trpc.userHub.upsert.useMutation({
-    onSuccess: async () => {
-      await utils.userHub.getById.invalidate({ id: hubId });
-      // Scoped to this hub: an unkeyed invalidate refetches every mounted
-      // image.getInfinite query, not only the feed behind this filter.
-      await utils.image.getInfinite.invalidate({ hubId });
-    },
+    onSuccess: () => invalidateHub(hubId),
     onError: (error) =>
       showErrorNotification({ title: 'Could not save hub', error: new Error(error.message) }),
   });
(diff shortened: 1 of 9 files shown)


JustMaier:
Review round applied in `daf190e00f`.

Two review agents ran read-only over `0ac43a468b` — comment necessity and correctness.

**Real finding, fixed.** `HubUpsertModal` gained a feed key it did not have, and `invalidate` resolves only once the active refetch settles — so renaming a hub from its own page held the modal open on a refetch of a feed that reads neither name nor description. It closes first now.

**A fourth copy neither ticket named.** `src/components/Filters/FeedFilters/HubFeedFilters.tsx:54` had `getById` + the feed, missing `getAll`, so a sort or filter save left the rail's cached rows carrying a stale `sort`/`filters` (both are in `hubSelect`). Latent — nothing reads them from `getAll` today. Now on the hook.

**`hubId` is required.** The optional was justified by a failed create, which never reaches `onSuccess`. Dead branch; its only live effect was letting a future call site skip two of the three keys silently.

**Collections arm.** Same divergence the models arm had — contributor VIEW alone, while the link path and the write path both refuse a private collection. Unreachable while `HUB_COLLECTION_SOURCES_ENABLED` is false; filtered anyway so flipping that constant does not reopen it.

**One assertion was carrying the whole scope guard.** A procedure missing from the map resolved to `undefined`, which the caller defaults to `TokenScope.Full` — both of its tests then passed for free. Each checks its own entry now.

**Worth knowing, not fixed:** the no-term suggestion path slices candidates to 50 before the names query, and that margin was sized against deleted rows only. The visibility filter removes a larger class, so a short first page of model suggestions is more likely than it was for an account with many bookmarks. The search path is unaffected — it filters in SQL before `take`.

Full unit suite green after each round (20924 passed, 0 failed). Typecheck clean apart from the pre-existing `packages/civitai-telemetry` error on `main`.

PR #4315, opened 2026-08-23
fix(search-index-cleanup): stop the nightly scan ending mid-index, and report when it does

ZacxDev:
## What's wrong

The nightly search-index cleanup job (`0 2 * * *`) can finish, report success, and have scanned only a fraction of a large index. It deletes exactly the stale documents it reached and silently leaves the rest — the residual is a clean step function in the primary-key space (zero below one id, non-zero above), the signature of a scan that stopped at a cursor while reporting success.

Two independent defects combine to produce that.

### Defect 1 — the run reports nothing, so truncation is invisible

`cleanupIndex` already computes per-index `idsScanned`, `staleFound`, `deleted`, `totalInIndex`, `errors` and `batchesProcessed`, and the job hands them back in its response body. Nothing logs them, and the job runner records only a duration. So "scanned a third of the index" and "scanned all of it" are indistinguishable from the outside, and the shortfall has to be reconstructed from the search engine's own task history after the fact.

### Defect 2 — the scan cannot cover a large index, and terminates early without saying so

Three interacting problems in the keyset `while` loop:

**(a) The requested page size was never clamped to the index's real ceiling.** A search `limit` above the index's `pagination.maxTotalHits` is truncated to it, and that ceiling is **per-index**. A page size of 1,000 means a multi-million-document index needs thousands of sequential round trips, which does not fit alongside six other indexes in one run — but simply raising it is not safe either, because an index left at the default ceiling would get a truncated page back and, via (b), end its scan after a single page. An in-code comment asserted the ceiling was always 1000, which is not true.

**(b) `if (docIds.length < batch) break;` is an unsafe terminator.** "Fewer hits than I asked for" and "no more documents" are different statements — the first is exactly what a clamped or engine-shortened page looks like.

**(c) `if (docIds.length === 0) break;` cannot distinguish end-of-index from a transient empty page.** The engine was concurrently applying document additions and deletions when the observed run stopped.

## The fix

**Logging (defect 1).** The job now emits one `logToAxiom` payload per index carrying `key`, `scanned`, `stale`, `deleted`, `errors`, `total` and an `incomplete` flag. An **incomplete pass** — `idsScanned < totalInIndex`, with the total known — is logged at **error** level with both numbers in the message, not buried in a success line. When `totalInIndex` is `null` (the stats call failed) completeness is unknowable and nothing is claimed either way. The job's return value is unchanged.

**Loop (defect 2).**

- **(a) Clamp, then raise.** The preflight already fetches the index settings for the filterable/sortable check; it now reads `pagination.maxTotalHits` from the same call and clamps the effective batch to it. With that in place the job can request a genuinely useful page size (10,000, matching the id-list size already used for deletes) and each index gets the largest page it can actually serve, instead of a truncated one. The stale comment is corrected.
- **(b)** The short-page terminator is removed. The cursor advances on the last returned id either way, so the only cost is one extra query per index.
- **(c)** An empty page now has to **confirm itself**: the same cursor is re-issued once after a short pause, and only a second empty reply ends the scan. If the confirmation returns rows the scan continues. This is the leading (unproven) hypothesis for the observed stop; it is cheap and it removes the class regardless of whether it fired.

Cross-run cursor persistence is deliberately **not** added — with (a) fixed, a full pass fits in one run.

Note one intentional consequence: an index that fails the filterable/sortable preflight scans zero documents, so it will now also be reported as an incomplete pass. That is accurate — it is not being cleaned at all — and it sits alongside the existing preflight error that explains why.

## Test matrix

Every new test was run against the **pre-change** source first and watched to fail, then against the fixed source. Two suites, 11 tests.

| Test | At `origin/main` | At HEAD | Failure at base |
|---|---|---|---|
| `cleanup.test.ts` › short page does not end the scan | 🔴 FAIL | ✅ PASS | `expected 3 to be 10` — scan stopped after page 1 of 4 |
| `cleanup.test.ts` › continues past a transient empty page | 🔴 FAIL | ✅ PASS | `expected 2 to be 6` — scan stopped at the injected empty page |
| `cleanup.test.ts` › ends the scan on a *confirmed* empty page | 🔴 FAIL | ✅ PASS | `expected ['id > -1','id > 3'] to deeply equal ['id > -1','id > 3','id > 3']` — no confirmation query |
| `cleanup.test.ts` › clamps a batch above the index ceiling and still completes | 🔴 FAIL | ✅ PASS | `expected 5 to be 12` — unclamped `limit`, short page, stopped after one page |
| `cleanup.test.ts` › INVARIANT GUARD: a ceiling *above* the batch does not raise it | ✅ pass | ✅ PASS | — labelled in the test as an invariant guard, not regression coverage |
| `search-index-cleanup.test.ts` › per-index stats reach the log sink | 🔴 FAIL | ✅ PASS | `expected undefined to match object {...}` — nothing logged |
| `search-index-cleanup.test.ts` › truncated scan logged at error level with both counts | 🔴 FAIL | ✅ PASS | `expected undefined to be 'error'` |
| `search-index-cleanup.test.ts` › does NOT flag a complete pass | 🔴 FAIL | ✅ PASS | `expected undefined to be 'info'` |
| `search-index-cleanup.test.ts` › does NOT flag a pass with an unknown total | 🔴 FAIL | ✅ PASS | `expected undefined to be 'info'` |
| `search-index-cleanup.test.ts` › requests a page size large enough for a big index | 🔴 FAIL | ✅ PASS | `expected 1000 to be 10000` |
| `search-index-cleanup.test.ts` › INVARIANT GUARD: job return value unchanged | ✅ pass | ✅ PASS | — labelled in the test as an invariant guard |

**9 red at base, 2 pre-existing-green invariant guards (both labelled as such in the test name and a comment). 11 green at HEAD.**

The clamp test deliberately covers the combination that matters: an index whose ceiling is **below** the requested batch **and** which holds more documents than one page — it asserts every request asks for the ceiling (not the unservable larger value) *and* that the scan reaches the last document.

Expectations are literal page sequences and counts written from the fixture (exact `id > N` cursor walks, exact scanned/batch counts), not read back off the implementation.

### Verification

- `pnpm typecheck` — **0 errors**.
- `tsc -p tsconfig.tests.json` (test files, excluded from the above) — **887 pre-existing errors with the new files removed, 887 with them added; delta 0**, and no diagnostic names either new file.
- `eslint` on all four changed files — **0 problems**.
- Unit run: **11 passed, 0 failed**, counted from the verbose reporter's per-test lines rather than an exit code.

**Not verified:** nothing here has been exercised against a live search engine. The loop behaviour is proven only against a fake index; the throughput claim for the larger page size is an expectation, not a measurement. The first nightly run after this merges is the real check — and it is now the first one that can report its own coverage.

### DIFF
--- a/src/server/jobs/__tests__/search-index-cleanup.test.ts
+++ b/src/server/jobs/__tests__/search-index-cleanup.test.ts
@@ -26,7 +26,16 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const { cleanupAllIndexes } = vi.hoisted(() => ({ cleanupAllIndexes: vi.fn() }));
 
-vi.mock('~/server/meilisearch/cleanup', () => ({ cleanupAllIndexes }));
+// Spreads the original so `CLEANUP_INDEXES` is the REAL configured list — the
+// never-attempted check below compares against it, and a hand-written stand-in
+// would make that assertion a statement about the fixture instead.
+vi.mock('~/server/meilisearch/cleanup', async (importOriginal) => ({
+  ...(await importOriginal<typeof CleanupModule>()),
+  cleanupAllIndexes,
+}));
+
+import type * as CleanupModule from '~/server/meilisearch/cleanup';
+import { CLEANUP_INDEXES } from '~/server/meilisearch/cleanup';
 
 // `~/server/logging/client` is a CANONICAL shared-module mock: src/__tests__/setup.ts
 // registers it once, globally, spreading the real module and overriding only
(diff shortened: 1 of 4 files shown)


ZacxDev:
Measured the page-size assumption against a production-scale index rather than assuming it, and the result tempers what this PR should be expected to achieve.

A larger page **is** served (the engine returns the full requested page where the index ceiling allows it, and the clamp correctly caps it where it does not — both cases confirmed). But the per-document scan cost improves only about **1.5x**, not in proportion to the page size: the work is dominated by the engine walking and serializing the documents, which is roughly constant however they are paged. The saving on the Postgres side is a genuine 10x fewer round trips, which is not included in that figure and was not measured.

So the honest expectation is: **the larger page reduces the shortfall, it does not obviously eliminate it.** A full pass over the largest index may still not complete in one nightly run.

That is an argument for this PR, not against it — it is precisely why the reporting half matters more than the throughput half. Before this change an incomplete pass was indistinguishable from a complete one and had to be reconstructed after the fact from the engine`s own task history; after it, the run says `INCOMPLETE SCAN: <index> scanned N of M` at error level. **The first nightly run after merge is the real measurement, and it will now state its own verdict.** If it still reports incomplete, the follow-up is cross-run cursor persistence (deliberately out of scope here), and that log line is what will tell us.

Nothing in the diff changes on the back of this — dropping the short-page terminator makes the change fail-safe either way: a page smaller than requested no longer ends the scan, so an underdelivering page size costs throughput and never correctness.

ZacxDev:
Correcting my own PR description, and one finding from the audit it went through. Posting rather than editing the body silently, since a reviewer may already have read the wrong version.

## Corrections to the PR description

1. **"10,000 matches `deleteChunkSize`, which is the id-list size this code already sends to Postgres and to the engine."** — Wrong. `deleteChunkSize` is used only in the delete-chunk loop; the Postgres id-list size has always been `batch`. There is no existing precedent for a 10x Postgres `IN (...)` list, and I should not have implied one.

2. **"an in-code comment asserted the ceiling was always 1000, which is not true."** — Wrong about which sentence was wrong. The removed comment correctly named the setting and correctly warned that raising `batch` without raising the ceiling truncates the page. The sentence that was actually false is the one after it — "we would not miss docs, just halve throughput" — which was untrue precisely because the same loop carried `if (docIds.length < batch) break;`.

3. **"the only cost is one extra query per index"** (for removing that terminator) — True only for an index whose scan already reached the end. Where the terminator was firing early, which is the entire premise of this PR, the cost is the whole remainder of the index. That understates it.

4. **"with (a) fixed, a full pass fits in one run"** — Unsupported, and I am withdrawing it. See below.

## Correcting the audit

The audit concluded the clamp makes the batch raise inert, on the grounds that nothing in this repo ever writes `pagination.maxTotalHits`, so every index must sit at the engine default of 1000. The first half is correct; the conclusion is not. `maxTotalHits` is live index state, and the repo is not the surface that defines it. Measured directly against all seven indexes the job walks:

- **Four are at 100,000** — including the two largest by a wide margin.
- **Three are at the engine default 1,000**, one of which holds zero documents.

So the raise is effective exactly where the scan is long, and the clamp is genuinely load-bearing for the other three rather than defensive dead code. That finding does not stand as written.

**But its underlying point survives and is worth more than the finding was:** that 100,000 is undeclared out-of-band state. Nothing in `onIndexSetup` sets it, so an index rebuilt from this repo comes back at 1,000 and the scan throughput silently regresses by 10x with nothing failing. That should be declared in code, and I am treating it as a required follow-up rather than a nit.

## Status

Not merging this as it stands. Beyond the above, the audit raised a genuine regression I had not considered: deletions are accumulated and applied only after a full index scan, so a longer scan plus a mid-flight cancellation now yields **zero** deletions for that index and skips every later one — where the previous early-terminating behaviour at least deleted the prefix it had reached. Making the scan complete is not worth making it all-or-nothing. Fixes for that and the smaller findings are in progress.

PR #4316, opened 2026-08-23
fix(filters): the sort a viewer cannot pick was still the one the feed ran on

JustMaier:
Your call from the request card: image and video feeds only, and change the model
gallery's default. Both are here — the default change alone does not reach anyone
who already picked Newest, since that choice is persisted per browser.

### What was wrong

`DumbSortFilter` filtered the sort **menu** through `isSortAvailable` and passed
the **selected value** through untouched. Nothing else in the app called that rule
— only this component and `hub-sort.ts` — so `useImageFilters` handed the withheld
sort straight to the query.

So a viewer without `canViewNsfw` on a model page saw a control reading "Newest"
above a menu offering only Most Reactions / Most Comments / Most Collected, and a
gallery that really was sorted newest-first. Not a wrong label: the sort the rule
exists to withhold was the sort in force.

### What changed

- `modelImages` defaults to `MostReactions` (it was `Newest`, deliberately).
- `resolveFeedSort` in `sort-availability.ts` returns Most Reactions when the menu
  will not offer what it is given, and both reads of an image feed's sort go
  through it — the control in `DumbSortFilter`, and the query in `useImageFilters`.
  Resolved on read, never written back, so regaining NSFW browsing hands the
  viewer their own choice again.
- `resolveHubSort` is now that same function. Hub behaviour is unchanged and its
  existing suite pins that.

**Scope.** Only `images`, `videos` and `modelImages`. The rule withholds
Newest/Oldest from every feed type, but it exists because a freshly posted *image*
may not be rated yet — so models, threads, tools, generation and buzz withdrawals
keep whatever they hold, per your "I'm fine with newest on all of those other
places".

### Who sees a change

| viewer | before | after |
| --- | --- | --- |
| SFW domain, model gallery | control said Newest, feed *was* newest-first | Most Reactions, and the menu agrees |
| NSFW-capable, model gallery, no saved choice | Newest | Most Reactions (the default change you asked for) |
| anyone with a saved sort the menu offers | their choice | unchanged |
| `?sort=Newest` link opened on a SFW domain | ran Newest | Most Reactions |
| hubs | — | unchanged |

### Verified

Reverting each half turns the new suites red — without `resolveFeedSort` 6 tests
fail across the two files, without the `useImageFilters` call 3 fail. Both suites
carry a control that would catch a resolver returning Most Reactions for
everything.

Full unit suite green (20961 passed, 0 failed). Typecheck clean apart from the
pre-existing `packages/civitai-telemetry` error on `main`. eslint and prettier
clean.

Not covered: `ResourceSelectSort` in the generation resource picker is a third
hand-rolled copy of this rule with no moderator bypass. Left out on purpose —
it changes generator behaviour. Ticket 868kv6mzm.

Ticket: 868kv6mz9.

### DIFF
--- a/src/components/Filters/SortFilter.tsx
+++ b/src/components/Filters/SortFilter.tsx
@@ -1,6 +1,6 @@
 import type { ButtonProps } from '@mantine/core';
 import { useRouter } from 'next/router';
-import { isSortAvailable } from '~/components/Filters/sort-availability';
+import { isSortAvailable, resolveFeedSort } from '~/components/Filters/sort-availability';
 import { useSortAvailability } from '~/components/Filters/useSortAvailability';
 import { SelectMenuV2 } from '~/components/SelectMenu/SelectMenu';
 import type { FilterSubTypes } from '~/providers/FiltersProvider';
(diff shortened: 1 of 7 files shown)


JustMaier:
Verified in a browser on `https://civitai-dev.green`, signed out, same page both times — `/models/4384`, scrolled to the gallery:

| code served | gallery sort control |
| --- | --- |
| `main` at `5a382a63b4` | **Newest** |
| this branch | **Most Reactions** |

Same host, same session shape, one commit apart — so the check could fail, and did. The probe reads the control's own rendered text after the feed hydrates, not a marker in the SSR payload.

(A probe against production found no gallery at all — different page structure — so it is not a usable control either way. The one above is the one I trust.)

PR #4319, opened 2026-08-23
feat(app-listings): the mod control the `exclude` column was added for — and the counter it has to move

ZacxDev:
## What this is

`AppListingReview.exclude` has carried the comment *"Moderator controls: keep abusive reviews out of the recommend-% aggregate"* and an index on `[appListingId, exclude]` since the table was created. `listReviews` has always filtered it. **Nothing has ever written it** — there was no delete, hide, or moderation path for an app-listing review at all. The service even said so: *"we never flip `exclude` on the write path (that's the deferred mod control)"*.

This builds that control.

- **Service** — `setAppListingReviewExclude({ reviewId, exclude })` in `app-listing-review.service.ts`: sets the flag and applies the matching `AppListingMetric` delta in **one transaction**.
- **Router** — `appListings.setReviewExclude`, `moderatorProcedure` + the inner `isModerator` recheck (the same idiom as `delistListing` / `relistListing` / `claimListing` / `purgeListing` in this router). Added to that file's asserted mod-action ledger.
- **No migration.** The column already exists. Nothing in this PR touches the DB.

## The part that is easy to get wrong: the counter

`AppListingMetric.thumbsUpCount` / `thumbsDownCount` are **denormalized with no rollup job** — they are exactly the sum of the ±1 deltas their writers have applied. So flipping `exclude` *without* adjusting the counter leaves the displayed "N% recommend (M)" permanently wrong, with nothing to repair it.

Rather than write a second copy of that arithmetic, the delta computation and the metric write are **extracted out of `upsertReview`** into `recommendMetricDelta` (a pure function of the BEFORE/AFTER contributions) and `applyRecommendMetricDelta`, and **both writers call them**. `upsertReview` behaves identically — its 33 pre-existing tests pass unchanged.

Properties:

- **Both directions.** Hiding decrements the bucket matching the row's `recommended`; un-hiding re-increments the same bucket. Exactly one counter moves, by exactly 1 — the other bucket's key is *absent* from the update, not set to `increment: 0`.
- **Idempotent, and serialized.** `exclude` is an **explicit target state, not a toggle** (a deliberate divergence from `resourceReview.toggleExclude`), and the transition is expressed as the UPDATE's own predicate — `updateMany({ where: { id, exclude: { not: exclude } } })`. Postgres re-evaluates that WHERE against the newly-committed row version after taking the row lock (EvalPlanQual), so of two concurrent hides exactly one sees `count === 1`; the delta is gated on being that one. A retry, a double-submit, **or two moderators working the same report** therefore move the aggregate exactly once. `recommended` is read *after* the conditional update, i.e. under the lock, so a racing author edit cannot misdirect the delta.
- Clamped at zero the way the existing metric write is.
- Keyed on the **review id**, which is what `listReviews` already projects: the moderator acts on exactly the row they are looking at, via one opaque handle, and cannot mis-pair a listing with the wrong reviewer.

## Why a soft hide and not a delete

1. **The denormalized counter.** A raw `delete` leaves `thumbsUp/DownCount` permanently wrong — there is no recount job. Any *correct* delete would have to do this same delta anyway, so the delta is the real work either way.
2. **It re-opens the once-ever review reward.** The sibling app-review reward documents its once-per-(user, app) guarantee as *"the DB-unique + `isFirstReview` belt"*; the only other guard is a same-day Redis dedup that expires. Removing the row lets the same user file a fresh **first** review, so delete→recreate becomes a Buzz loop. (Precisely: that reward is currently wired to `AppBlockReview`, not `AppListingReview` — but `upsertAppListingReview` already returns `isNewReview` for exactly that hook, so this is a trap laid for the moment it is wired, not a hypothetical.)
3. **Audit trail + consistency.** `exclude` keeps the abusive text readable to moderators and matches the semantics of its sibling `tosViolation` flag on the same row.

## Tests — red-at-base matrix

Base = `origin/main` @ `c227e78d71`. Measured by reverting **only** the three source files to base and re-running the new test files in place.

| | at base | at HEAD |
|---|---|---|
| `app-listing-review.mod-exclude.test.ts` (23 tests, new file) | **20 failed**, 3 passed | 23 passed |
| `app-listings.router.mod-actions-authz.test.ts` (8 new cases; 40 → 48 tests) | **6 of the 8 failed** | 48 passed |
| `app-listing-review.service.test.ts` (18) + `.store-scope.test.ts` (15) | 33 passed | 33 passed |

That last row is the check on the refactor: `upsertReview`'s existing assertions are unchanged and green both before and after the delta extraction. (Their *arrangement* changed in the follow-up commit — the prior-review read became `SELECT … FOR UPDATE`, so three tests now stub `$queryRaw` instead of `findUnique`. No assertion was weakened; the mutant that removes the row lock is killed by three of them.)

**Honest caveat on that matrix:** most of those 26 base failures are red *because the symbol does not exist there*, which is weak evidence about the assertions themselves. The mutation sweep below is the real one.

The 2 new router cases that stay green at base are the generic `MOD_ACTIONS` ledger loop's "tester is FORBIDDEN" / "anonymous is rejected" entries, which assert only `instanceof TRPCError` — a pre-existing looseness in that loop, deliberately not widened here. The dedicated `setReviewExclude` describe covers the same ground with exact codes and does fail at base.

**3 cases are green at base — labelled in the file as INVARIANT GUARDS, not regression coverage:**

- `getMyAppListingReview` does **not** filter `exclude` — an author still sees their own hidden review (deliberate: the edit form prefills from it, and blanking it would invite a "new" review that the DB unique silently turns into an update of the row they cannot see).
- `listAppListingReviews` **does** filter `exclude` — a hidden review leaves the public list. This is what makes the mod action take effect with no read-path change.
- A mod-excluded prior review stays un-counted when its author edits it — an author cannot restore their own hidden review to the aggregate.

All three already held; they are pinned because the mod control makes them load-bearing for the first time.

Two of the router authz assertions were initially **green at base for the wrong reason**: calling a proc that does not exist also rejects with a `TRPCError` (`NOT_FOUND`, *"No procedure found on path"*), so `rejects.toBeInstanceOf(TRPCError)` proved the gate without the gate existing. They now pin `FORBIDDEN` / `UNAUTHORIZED` / `BAD_REQUEST` specifically, and fail at base.

## Mutation sweep

**23 isolated mutants** (each the narrowest expression that can be wrong), plus unmutated positive controls. **All 23 killed; controls green (117/117).** Every killer below is the measured test name, not an attribution I assumed.

| mutant | killed by |
|---|---|
| race guard (`count === 0`) removed | the 3 no-op cases **+ both concurrency cases** |
| conditional predicate dropped (`where: { id }`) | both direction cases (call-shape assertion) |
| conditional update deleted entirely | 13 tests |
| AFTER contribution → wrong bucket | truth table + both writers' direction cases |
| BEFORE contribution `-=` → `+=` | truth table + `upsertReview`'s flip case |
| BEFORE state computed from the post-update value | all four direction cases |
| zero-delta write guard removed | the *user write path's* details-only-edit tests (proves the shared helper is reachable from both writers) |
| non-negative clamp removed | hide case + the pre-existing flip case |
| **counter write moved out of the transaction** | all four direction cases (this is what the sentinel tx client buys) |
| **author-side prior read loses `FOR UPDATE`** | 3 `upsertReview` cases + the seam case |
| requested target ignored (always hides) | un-hide case |
| NOT_FOUND removed | the missing-review case |
| cache bust made unconditional | the two no-op cases + the race-loser case |
| `moderatorProcedure` → `protectedProcedure` | the **tester** refusal case (in the ledger loop and in the dedicated describe) |
| `exclude: false` dropped from the public list | the public-list invariant guard + a pre-existing case |
| `exclude: false` added to `getMyReview` | the author-visibility invariant guard + 3 pre-existing cases |
| **a third caller of the delta writer appears** | the writer-set ledger |
| **`FOR UPDATE` deleted from the prior read** | the SQL-text row-lock assertion |
| **prior-read table name typo'd** | the SQL-text table assertion |
| **`"exclude"` dropped from the prior SELECT** | the SQL-text column assertion |
| `stripComments` reverted to the cross-line regex | the fail-open stripper control |
| `delete`/`deleteMany` dropped from the writer predicate | the writer-predicate controls |
| Kysely reach-around dropped from the writer predicate | the writer-predicate controls |

### The guard that protects the fix had no guard

Worth stating because it is the PR's own subject one level up. The conditional-update half of the race fix has four kills plus an ordering pin. The `SELECT … FOR UPDATE` half had **none** — every test mocks `$queryRaw`, so the SQL string was never asserted and never executed, and three mutations on it **survived the full 21k-test suite**: deleting `FOR UPDATE` (silently reverts the fix), typo'ing the table name (hard runtime failure, zero CI signal), and dropping `"exclude"` from the SELECT (`prior.exclude` becomes `undefined`, so a mod-hidden review re-enters the aggregate). Asserting the tagged template's static strings — the lock clause, the quoted table, every selected column, and that the lookup keys arrive as **bound parameters** rather than interpolated text — kills all three with no database. Each now dies to its own dedicated assertion, not collateral.

### Correction: the anonymous case does not test the mod gate

An earlier version of this table said the `protectedProcedure` mutant was killed by "both non-mod refusal tests". Measured, only the **tester** case goes red. `moderatorProcedure` is `protectedProcedure.use(isMod)`, so an anonymous caller is rejected by `isAuthed` with `UNAUTHORIZED` before `isMod` is reached — and that is equally true under the mutant. The anonymous case is **structurally incapable** of discriminating moderator-only from merely-authenticated, so it cannot be made to discriminate cheaply; it is now labelled in the test file as an **auth** guard rather than counted as evidence of the mod gate.

## Limits / not verified

- **Mocked Prisma only** — no test in this repo exercises real Prisma for this service, so the call shapes are pinned, not the emitted SQL. Consistent with the existing tests in the same file. Concretely: the concurrency tests drive the `count === 0` / `count === 1` branches directly, and the conditional-update *shape* is asserted — but **no test executes real EvalPlanQual**. The serialization argument rests on Postgres semantics plus the pinned call shape, not on an observed interleaving.
- The `$transaction` mock now hands the callback a **distinct sentinel client**, so "inside the transaction" is witnessed: a mutant moving the counter write onto the plain `dbWrite` is caught. (It survived before this was added.)
- **No `AppListingModerationEvent` is written.** That table's `action` column is CHECK-constrained in SQL, so a new action value needs a manual-apply migration — out of scope here. The `exclude` flag is itself the durable record; the audit event is a follow-up that needs the DDL first.
- `tosViolation` is deliberately **not** folded into the "does this row count" predicate. `listReviews` hides both flags, but the counters have only ever tracked `exclude`, and nothing writes `tosViolation` today — widening it without a recount would make the live counters disagree with the rows they were accumulated from. The predicate is now a single function, so that is a one-line change *when* someone brings the recount.
- **No UI.** Backend only, consistent with the rest of this surface being dark.
- 🟡 **The `FOR UPDATE` serializes the UPDATE path only — the CREATE path is not serialized, and this PR does not close it.** `FOR UPDATE` locks nothing when the row does not exist, so two in-flight *first* reviews from one user (double-submit, retry, two tabs) both read `prior = null` and both compute `+1`: either a permanent over-count that the `≥0` clamp cannot mask (it errs high), or the loser 500s on the unique. Same class as the bug fixed above, on the branch a row lock cannot reach. It is **pre-existing**, and closing it means an advisory lock or an `INSERT … ON CONFLICT` returning the pre-image — a behavioural change to the author write path. #4320's wholesale recompute retires this drift class along with the others, which is the better trade than narrowing it twice. The code comment now says this instead of the previous claim that the case was impossible.
- **Transaction budget, recorded so it is not a surprise later:** both `$transaction` calls use Prisma's default 5000 ms interactive budget, and the new lock waits sit inside it. Under real contention this surfaces as a **`P2028` timeout rather than queueing**. Fine at current volume.
- **The writer set is now pinned, in both directions.** `applyRecommendMetricDelta` is module-private with exactly two callers and a comment saying a third must never appear; nothing enforced that. A ledger test now asserts the caller set as an **equality** (shrinking is as red as growing) and scans `src/server/**` for any other module writing the thumbs counters. It is one half of a two-sided contract — the metric processor `src/server/metrics/appListing.metrics.sql.ts` deliberately names only `install_count`, already guarded by `appListing.metrics.test.ts`; each guard now names the other.
- **The ledger scan needed three rounds of instrument validation, recorded because the failure modes are instructive.** v1 flagged 3 files: 2 were comment-only matches — including the very file whose comment *promises* it does not write thumbs. v2 (comment-stripped) still flagged `app-listing.service.ts`, a read-only consumer, because a bare `UPDATE` matches the `update` inside `updated_at`. v3 fixed the stripper itself: the cross-line block-comment regex **fails open** — an open marker inside a string literal matches to the next real close marker and deletes the lines between it, including a genuine write, so the scan would report "no other writers" with total confidence. It is now line-oriented (cannot run away; an unstripped inline block comment fails *closed* instead). The predicate also gained `delete`/`deleteMany` (deleting a review destroys its counters) and Kysely's `updateTable`/`insertInto`/`deleteFrom`, and the walk widened past `src/server/**` to `src/**` (incl. `.tsx` and `src/pages/api/**`) and `packages/*/src/**`. Still exactly one writer, so that widening closes a **coverage limit, not a live gap**. All of it is controlled: eight write shapes it must detect, three reads it must ignore, the fail-open fixture, and each hardening proven load-bearing by reverting it and watching its own control go red.
- **Not in this PR, filed separately:** a wholesale thumbs recompute in `appListing.metrics.sql.ts`. The sibling `ResourceReview` has exactly that, which is *why* `resourceReview.toggleExclude` needs no delta arithmetic at all. That would make the scheme self-healing and retire this drift class rather than narrowing it — a bigger change deserving its own review.

## Gates

`pnpm typecheck` 0 errors · ESLint clean on the touched files · Prettier clean on the touched files (the pre-existing formatting drift elsewhere in `app-listings.router.mod-actions-authz.test.ts` is untouched — it is present verbatim at base) · full local `pnpm test:unit:run` → **1346 files / 21007 tests, 0 failures**.

### Note on the second commit

The first push failed CI on exactly one test, in exactly this PR's new file: `no-direct-shared-module-mock`, a ratchet that forbids a per-file `vi.mock('~/server/db/client', …)` because under `isolate: false` it freezes that file's mock shape into every later file in the same worker. The new test file has been migrated to the canonical `dbMock` (`src/__tests__/mocks/db.mock.ts`) — same 23 assertions, and the mutation sweep above was re-run against the migrated file with identical results.

Worth recording because it is the interesting kind of miss: the targeted local run covered `src/server/services/blocks/__tests__/` and `src/server/routers/__tests__/`, and the guard lives in `src/server/services/__tests__/` — one directory outside the selection. A directory-scoped local run is structurally blind to a repo-wide ratchet. The full suite is now the check.

### DIFF
--- a/src/server/routers/__tests__/app-listings.router.mod-actions-authz.test.ts
+++ b/src/server/routers/__tests__/app-listings.router.mod-actions-authz.test.ts
@@ -354,6 +354,15 @@ describe('setReviewExclude — the per-review mod hide control', () => {
   // a branch where `setReviewExclude` was never added — it would prove the gate
   // without the gate existing. Pinning FORBIDDEN / UNAUTHORIZED is what makes these
   // fail at the base commit.
+  //
+  // 🟡 …but only the TESTER case tests the MOD gate. `moderatorProcedure` is
+  // `protectedProcedure.use(isMod)`, so an ANONYMOUS caller is rejected by `isAuthed`
+  // with UNAUTHORIZED before `isMod` is ever reached — and that is equally true if the
+  // gate is downgraded to `protectedProcedure`. The anonymous case is therefore
+  // STRUCTURALLY incapable of discriminating mod-gated from merely-authenticated, and
+  // is measured to survive that mutation. It is an AUTH guard; the tester case below
+  // is the mod guard. Labelled rather than deleted because "anon cannot reach this"
+  // is still worth pinning — just not as evidence of moderator-only.
   it('a tester is FORBIDDEN and the service is NOT called (a user cannot hide a review)', async () => {
     const caller = appListingsRouter.createCaller(fakeCtx(tester) as never);
     await expect(caller.setReviewExclude({ reviewId: 7, exclude: true })).rejects.toMatchObject({
--- a/src/server/routers/__tests__/app-listings.router.mod-actions-authz.test.ts
+++ b/src/server/routers/__tests__/app-listings.router.mod-actions-authz.test.ts
@@ -27,6 +27,7 @@ const {
   mockListReports,
   mockResetOnsite,
   mockPreviewForReview,
+  mockSetReviewExclude,
   mockIsAppBlocksEnabled,
   mockIsAppBlocksAuthorEnabled,
 } = vi.hoisted(() => ({
@@ -45,6 +46,12 @@ const {
   mockReport: vi.fn(async () => ({ reportId: 'alrp_1' })),
   mockListReports: vi.fn(async () => ({ items: [], nextCursor: null })),
   mockPreviewForReview: vi.fn(async () => ({ card: {}, detail: {} })),
+  mockSetReviewExclude: vi.fn(async () => ({
+    id: 7,
+    appListingId: 'apl_1',
+    exclude: true,
+    changed: true,
+  })),
   mockIsAppBlocksEnabled: vi.fn(),
   mockIsAppBlocksAuthorEnabled: vi.fn(),
 }));
@@ -65,6 +72,12 @@ vi.mock('~/server/services/blocks/offsite-moderation.service', () => ({
 // mock just that export so the mod-gate authz can be exercised without a DB.
 vi.mock('~/server/services/blocks/app-listing.service', () => ({
   getListingPreviewForReview: mockPreviewForReview,
+}));
+// The per-review mod hide/un-hide control is dynamically imported by
+// `setReviewExclude`; mock just that export so the mod gate can be exercised
+// without a DB. (The user-facing review procs in this router are not called here.)
+vi.mock('~/server/services/blocks/app-listing-review.service', () => ({
+  setAppListingReviewExclude: mockSetReviewExclude,
 }));
 vi.mock('~/server/services/app-blocks-flag', () => ({
   isAppBlocksEnabled: mockIsAppBlocksEnabled,
@@ -157,6 +170,14 @@ const MOD_ACTIONS: Array<{
     name: 'resetOnsiteListingToPending',
     mock: mockResetOnsite,
     call: (c) => c.resetOnsiteListingToPending({ appListingId: 'apl_1', reason: REASON }),
+  },
+  {
+    // W13 per-REVIEW mod control — hide/un-hide one `AppListingReview` and move the
+    // denormalized recommend counters. Same `moderatorProcedure` + `isModerator`
+    // recheck boundary as every action above, so it belongs in this ledger.
+    name: 'setReviewExclude',
+    mock: mockSetReviewExclude,
+    call: (c) => c.setReviewExclude({ reviewId: 7, exclude: true }),
   },
 ];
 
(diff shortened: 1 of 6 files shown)


ZacxDev:
## The EvalPlanQual assumption is now observed, not asserted

The PR states an honest limit: the concurrency tests pin the conditional-update *shape*, but no test executes real EPQ, so the serialization argument rested on Postgres semantics. I settled that against a real Postgres 18 instance (the writable dev clone), on the actual `app_listing_reviews` row.

**Two sessions racing the same row, with a control:**

| shape | session A | session B (racing) | outcome |
|---|---|---|---|
| **old** — decision taken before the lock, then `UPDATE … WHERE id = $1` | `UPDATE 1` | `UPDATE 1` | both apply their delta → permanent drift |
| **new** — the transition IS the update's predicate | `UPDATE 1` | `UPDATE 0` | exactly one applies |

Method: session A opens a transaction, runs the update, holds the row lock with `pg_sleep(6)`, then commits. Session B issues the same statement ~2s in, blocks on the lock, and on unblocking re-evaluates its predicate against the **new** row version — which no longer satisfies `exclude IS DISTINCT FROM true`, so it reports `UPDATE 0`. That is EPQ doing exactly what the fix depends on.

The first row is the part that makes the second meaningful: it is the negative control, and it demonstrates the rig can produce the failure mode. Without it, `UPDATE 0` could have been zero for any number of unrelated reasons.

So the `count === 1` gate is now backed by an observed interleaving on the real engine, not only by the mocked call shape. What still isn't covered end-to-end is the *application* path (Prisma emitting this statement inside its transaction) — the shape is pinned by unit tests, and the semantics are pinned here; nothing exercises both at once.

Unrelated to this PR, for whoever picks up #4320: this is exactly the class a wholesale recompute retires rather than narrows.

ZacxDev:
Re-derived the round's headline claim independently before merging: deleting `FOR UPDATE` from the SQL clause now kills **exactly one** test —

```
× takes a ROW LOCK — the clause whose deletion silently reverts the fix
  Tests  1 failed | 53 passed (54)
```

— so it dies for its own reason rather than as collateral. That mutant **survived** one round ago, which is what made it worth closing: the clause the whole author-path fix rests on had no test that could see its removal.

🔴 One note on method, because it nearly produced a wrong answer here too: `FOR UPDATE` appears **5 times** in that file and only one is the SQL. My first attempt removed occurrence #1 — a *comment* — and a count-based check (`5 → 4`) would have called that mutant landed. Had I scored it, a surviving test run would have read as "the guard is broken" when nothing testable had changed. Checking *which* occurrence moved, not how many, is what caught it. Same shape as the finding itself.

PR #4321, opened 2026-08-23
perf(image): index tags by imageId instead of rescanning the tag array per image

ZacxDev:
## The defect

`getImagesByEntity` and `getEntityCoverImage` both associated tags to images like this:

```ts
return images.map((i) => ({
  ...i,
  tags: tagsVar?.filter((x) => x.imageId === i.id),
}));
```

`getImageTagsForImages(imageIds)` fetches the tags for **every** image in the batch, so the
array being scanned grows with the image count. The scan runs once per image, which makes the
join **O(N x M)** with M proportional to N — quadratic in the batch size, on top of a
per-element property compare.

CPU profiling of production API servers showed this construct dominating multi-second
event-loop stalls: roughly **33–41% of application CPU** in sampled stall profiles, and about
**93% of the single worst one**.

## The fix

No new idea — this adopts the idiom that already exists in the same file. `getAllImages` and
the search-result path both build a `Map<imageId, tags[]>` once and then look up per image.
This lifts that block into one exported helper, `attachTagsToImages`, and uses it at both
sites, so the predicate lives in one place instead of three.

```ts
export function attachTagsToImages<TImage extends { id: number }, TTag extends { imageId: number }>(
  images: TImage[],
  tags: TTag[] | undefined
): (TImage & { tags: TTag[] })[] {
  const tagsByImageId = tags?.reduce((acc, tag) => {
    const arr = acc.get(tag.imageId);
    if (arr) arr.push(tag);
    else acc.set(tag.imageId, [tag]);
    return acc;
  }, new Map<number, TTag[]>());

  return images.map((i) => ({ ...i, tags: tagsByImageId?.get(i.id) ?? [] }));
}
```

O(N + M), one pass to index and one to look up.

## 🔴 The `[]`-vs-`undefined` semantic, deliberately preserved

`.filter()` returns **`[]`** when nothing matches. `Map.get()` returns **`undefined`**. Those
are different values in the API response, and an image with no tags is common — so the
`?? []` is load-bearing, not cosmetic. Both call sites shipped `[]` before this PR and ship
`[]` after it, including the `include`-does-not-ask-for-tags case (`tagsVar` is initialised to
`[]`, so `.filter()` yielded `[]` for every image; the index is then empty and `?? []` yields
the same).

**Observation, not changed here:** the two existing index sites in this file disagree with
each other on exactly this point. The search-result site writes `tagsByImageId?.get(sr.id) ?? []`
— correct. The `getAllImages` site writes `tagsByImageId?.get(i.id)` with no fallback, which
`876284e59d` introduced when it replaced a `.filter()` there; an untagged image in that
response gets `undefined` where it previously got `[]`. That is a pre-existing latent behaviour
change on a much hotter path and is deliberately **out of scope for this PR** — flagging it so
it is a decision rather than an oversight.

## Tests

New: `src/server/services/__tests__/image-tag-association.test.ts` — 10 tests.

The helper, directly:
- correct association at N>1 with an interleaved, out-of-order tag array. Every `imageId` is
  pairwise distinct from its index in **both** arrays, so an off-by-one or an index-vs-id
  mix-up cannot pass by coincidence
- an image's own tags keep their relative order (they arrive score-sorted)
- an image with no matching tags gets `[]`, not `undefined`
- tagged and untagged images mixed in one call
- `[]` for every image when tags were not fetched at all (both the `[]` and the `undefined` arm)
- the image's other fields survive; the input tag array is not mutated

Both call sites, end to end (real service, `$queryRaw` and the tag cache stubbed): each image
receives its own tags, the untagged one in the middle receives `[]`, and
`getEntityCoverImage` still attaches `cosmetic`.

### Red → green matrix

Each mutation was applied to the helper, the suite watched go red, then reverted and watched
go green.

| Mutation | Result |
|---|---|
| baseline (as merged) | **10 passed / 0 failed** |
| drop the fallback: `tagsByImageId?.get(i.id) as TTag[]` | **6 failed / 4 passed** — each with its own assertion message, e.g. `an image with no tags must receive [], not undefined: expected undefined to deeply equal []` |
| look up by position: `tagsByImageId?.get(idx) ?? []` | **5 failed / 5 passed** — the association and ordering cases |
| restored | **10 passed / 0 failed** |

## Validation

- `vitest run --project 'unit*'` (full unit suite): **21078 passed / 0 failed / 25 skipped**
  across 1351 files (1350 passed, 1 skipped).
- `node scripts/typecheck.mjs`: **0 type errors**.
- Prettier: the added code is clean. The one remaining warning in `image.service.ts` is
  pre-existing on `main` and untouched here.

## Why `flipt-eval-context.test.ts` is in this diff

That suite's `ENTITY_WITHOUT_CONTEXT_LEDGER` keys ledgered Flipt call sites by
`file:line`, and four of its rows point into `image.service.ts`. Adding the helper shifted
those four lines by +27, so the ledger's own "a fixed or moved site must be removed from it"
check went red. The four keys are re-pointed at the same four unchanged call sites; no ledger
row was added, dropped, or re-justified.

### DIFF
--- a/src/server/flipt/__tests__/flipt-eval-context.test.ts
+++ b/src/server/flipt/__tests__/flipt-eval-context.test.ts
@@ -152,14 +152,14 @@ const ENTITY_WITHOUT_CONTEXT_LEDGER: Record<string, string> = {
   'server/services/article-rating-review.helpers.ts:482':
     'article-rating-dispute has no segment rollouts; background path with no SessionUser',
   // flag `feed-fetch-filter-in-post`: enabled=true, 0 rules, 0 rollouts.
-  'server/services/image.service.ts:3925':
+  'server/services/image.service.ts:3952':
     'feed-fetch-filter-in-post has no segment rollouts; hot feed path',
   // flag `feed-image-existence`: enabled=true, 0 rules, 0 rollouts. Three sites.
-  'server/services/image.service.ts:4055':
+  'server/services/image.service.ts:4082':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:4765':
+  'server/services/image.service.ts:4792':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:5875':
+  'server/services/image.service.ts:5902':
     'feed-image-existence has no segment rollouts; hot feed path',
   // flags `model-text-moderation-xguard` / `-apply`: both enabled=true, 0 rules,
   // 0 rollouts (checked against flipt-state and against the evaluation API on
@@ -206,7 +206,7 @@ describe('flipt evaluation context — source gate', () => {
     // these and fail the second, and vice versa.
     const bySite = new Map(calls.map((c) => [c.site, c]));
     expect(bySite.get('server/services/feedback.service.ts:43')?.argc).toBe(3);
-    expect(bySite.get('server/services/image.service.ts:4055')?.argc).toBe(2);
+    expect(bySite.get('server/services/image.service.ts:4082')?.argc).toBe(2);
   });
 
   it('adds no Flipt evaluation that names an entity but passes no context', () => {
--- a/src/server/services/__tests__/image-tag-association.test.ts
+++ b/src/server/services/__tests__/image-tag-association.test.ts
@@ -1,0 +1,300 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+/**
+ * `getImagesByEntity` and `getEntityCoverImage` used to associate tags to images
+ * with `tagsVar?.filter((x) => x.imageId === i.id)` — a full scan of the tag
+ * array once per image. `getImageTagsForImages` fetches the tags for EVERY image
+ * in the batch, so the tag array grows with the image count and the join is
+ * quadratic. `attachTagsToImages` indexes once and looks up.
+ *
+ * The behaviour that has to survive the rewrite:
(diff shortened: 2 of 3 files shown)


ZacxDev:
**Correction to this PR's description** — flagging publicly rather than silently editing the body, since a reviewer may already have read it.

The body says the extraction means "the predicate lives in one place instead of three." That is not accurate. After this PR the tag-association predicate lives in **three** places in `image.service.ts`:

- the new `attachTagsToImages` helper (uses `?? []`)
- an inline `Map` + lookup around the `getAllImages` transform (**no** `?? []`)
- another inline `Map` + lookup on the search-result path (uses `?? []`)

Two consequences worth being explicit about:

1. **The DRY goal is not met by this PR.** It removes the two `.filter()`-per-image sites and adds a shared helper, but does not fold the two pre-existing inline index blocks into it. Consolidating those is a reasonable follow-up; it is deliberately out of scope here to keep this change behaviour-preserving.

2. **The `[]`-vs-`undefined` divergence this PR calls load-bearing is still live at one of them.** The `getAllImages` site returns `undefined` for an image with no tags, where the other two return `[]`. That difference predates this PR (it arrived with the earlier `O(n*m)` conversion of that same transform) and is untouched here — but it means the codebase does not yet agree with itself on the semantic the "deliberately preserved" section argues for.

Nothing above changes the correctness of this PR: both sites it touches are byte-identical in output before and after. It only narrows an overstated claim in the description and records the remaining inconsistency so it is not mistaken for already-fixed.

PR #4325, opened 2026-08-24
fix(image): bound `entities` on getEntitiesCoverImage

ZacxDev:
Adds an upper bound to `entities` on `getEntitiesCoverImage` in `src/server/schema/image.schema.ts`, and has the three call sites clamp their own list to the same number so a list over the bound degrades instead of failing.

The array had neither a `.max()` nor a `.min()`, while comparable list inputs in the same area are bounded — `paginationSchema`'s `limit` is `.min(1).max(200)`, `getInfiniteImagesSchema`'s `limit` is `.min(0).max(200)`, `reportCsamImagesSchema.imageIds` is `.min(1)`. This just brings it in line.

### Why 500

The number is sized off what the callers can actually build, taken from the code rather than picked:

- **Profile showcase** — `ShowcaseSection` and `ShowcaseItemsInput` both pass `user.profile.showcaseItems`. `addEntityToShowcase` truncates that to `constants.profile.showcaseItemsLimit` (**32**) on every write (`src/server/controllers/user-profile.controller.ts`), and the edit modal advertises the same 32.
- **Notification panel** — `useNotificationThumbnails` passes deduped `details.imageId` values from the loaded notification list. That list is an infinite query at **30 per page** (`NotificationsComposed.tsx`, the only render site of `NotificationList`), so the count grows with scroll depth. 500 is a little over 16 pages, and in practice more, since only notifications that name an image contribute an id.

500 is ~16x the showcase cap and well past any realistic panel session, so it does not narrow any call site; it just gives the array a ceiling.

Note that 32 is enforced on **one** of the two showcase write paths. `userProfile.update` takes `showcaseItems: z.array(showcaseItemSchema).optional()` with no `.max()` (`src/server/schema/user-profile.schema.ts`) and the service stores it as given, so a stored showcase is not guaranteed to be at or under 32. The schema comment says which path enforces what rather than claiming the cap outright. **Follow-up, deliberately not in this PR:** bounding `showcaseItems` on `userProfile.update` is a behaviour change on a different endpoint and belongs in its own change.

### Callers clamp their own list

A refusal is not a degraded render in any of the three call sites — none has an `onError` or an error branch — and it does not recover on its own, because each re-render sends the same list:

- `useNotificationThumbnails` — on a refusal `data` is `undefined` and **every** thumbnail disappears, including the ones under the bound.
- `ShowcaseSection` — `isNullState` is computed from `!coverImages.length`, so a refusal takes the **entire Showcase section** off the profile.
- `ShowcaseItemsInput` — degrades better already (a per-item "There was a problem loading the cover image" fallback), but on a refusal every item shows it, including those that would have resolved.

So each slices to `MAX_ENTITIES_COVER_IMAGE`, now exported from the schema module so the number lives in one place instead of three. Thumbnails simply stop past the bound.

`ShowcaseItemsInput` clamps only the **query input** — the `showcaseItems` state that `onChange` saves is left whole, so editing a showcase longer than the bound cannot silently truncate it — and slices **after** the sort, since the sort is what keeps a drag-reorder from re-querying.

### `.min()`

Deliberately not added. `getEntityCoverImage` early-returns `[]` for an empty array, all three callers already guard on `length > 0`, and the procedure is reachable with a `MediaRead` token — so refusing an empty list would turn a supported no-op into a 400 for existing clients and buy nothing. There is a test pinning that empty stays accepted.

### Callers checked

Grepped the repo for `getEntitiesCoverImage` / `getEntityCoverImage`. Every path through this schema is the tRPC procedure, and its callers are the three above. `getUserCosmeticsHandler` calls the `getEntityCoverImage` **service** directly (bounded by the user's own equipped cosmetics) and does not go through this schema, so it is unaffected either way.

### Tests

`src/server/schema/__tests__/get-entities-cover-image.schema.test.ts` — at the bound, one over it (with the issue pinned), the size the ordinary showcase path produces (32), and the empty case.

The fixture ids are `900_000 + i`, deliberately far above the bound and not co-varying with the array index. **The earlier version of this test could not tell which bound it was certifying**: built as `i + 1` the ids ran `1..n`, so a `.max(500)` relocated from the array onto `entityId` accepts and refuses exactly the same inputs, and the old assertion `issue.path[0] === 'entities'` is equally true of `['entities', 500, 'entityId']`. That relocation leaves the array unbounded and 400s every real request (image ids are far above 500) — and the suite stayed green at 4/4. The rejection case now pins the whole issue, `{ code: 'too_big', path: ['entities'] }`, not its first path segment.

Verified by mutation on the shipped bytes:

| mutation | result |
|---|---|
| `.max()` removed | `rejects 501 entities` fails |
| `.max()` **relocated** onto `entityId`, array left unbounded | `accepts 500 entities`, `rejects 501 entities` and `accepts the largest list a caller can build` all fail (**3 of 4**) |
| `.max(100000)` | `rejects 501 entities` fails |
| `.max(499)` | `accepts 500 entities` fails |
| `.min(1)` added | `accepts an empty list` fails |
| unmutated | 4 passed |

The relocation row is the one the previous table was missing, and it was the one that mattered: against the **old** fixture that same mutant passes 4/4, which is what made the old table an overstatement of the evidence.

`src/components/Notifications/__tests__/useNotificationThumbnails.test.ts` gains a case for the caller-side clamp — a panel deeper than the bound must still produce a request the schema accepts, keeping its first ids. Dropping the `.slice` fails it (`expected […(525)] to have a length of 500`) and nothing else in that file.

Full unit suite: 21473 passed, 25 skipped, 1364 files. `node scripts/typecheck.mjs`: 0 errors. ESLint on the changed files is unchanged from the base commit (5 warnings, 0 errors, all pre-existing `no-unused-vars`); Prettier clean.

### DIFF
--- a/src/components/Notifications/__tests__/useNotificationThumbnails.test.ts
+++ b/src/components/Notifications/__tests__/useNotificationThumbnails.test.ts
@@ -61,6 +61,7 @@ vi.mock('~/utils/trpc', async (importOriginal) => ({
 }));
 
 import { useNotificationThumbnails } from '~/components/Notifications/notification-thumbnails';
+import { MAX_ENTITIES_COVER_IMAGE, getEntitiesCoverImage } from '~/server/schema/image.schema';
 
 (globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
 function renderHook<T>(useHook: () => T) {
@@ -148,5 +149,36 @@ describe('resolving the image a notification is about', () => {
     renderHook(() => useNotificationThumbnails([{ details: { placementId: 1 } }]));
 
     expect(state.useQuery.mock.calls[0][1]).toMatchObject({ enabled: false });
+  });
+
+  // The panel is an infinite list with no `maxPages`, so the id list grows with
+  // scroll depth. `getEntitiesCoverImage` bounds `entities`, and a refused query
+  // leaves `data` undefined with no error branch — every thumbnail would vanish,
+  // including the ones under the bound, and stay gone for the session because each
+  // re-render sends the same list. So the hook has to clamp before asking.
+  it('asks for no more entities than the query accepts, however deep the panel is', () => {
+    // Ids are far above the bound so a length that happened to be clamped by
+    // something keyed on the id value could not pass this.
+    const deepPanel = Array.from({ length: MAX_ENTITIES_COVER_IMAGE + 25 }, (_, i) => ({
+      details: { imageId: 900_000 + i },
+    }));
+
+    renderHook(() => useNotificationThumbnails(deepPanel));
+
+    const { entities } = state.useQuery.mock.calls[0][0] as {
+      entities: { entityType: string; entityId: number }[];
+    };
+
+    expect(entities).toHaveLength(MAX_ENTITIES_COVER_IMAGE);
+    // The kept ids are the panel's first, not an arbitrary slice: the newest
+    // notifications are the ones the user is looking at.
+    expect(entities[0]).toEqual({ entityType: 'Image', entityId: 900_000 });
+    expect(entities[MAX_ENTITIES_COVER_IMAGE - 1]).toEqual({
+      entityType: 'Image',
+      entityId: 900_000 + MAX_ENTITIES_COVER_IMAGE - 1,
+    });
+    // The clamp has to leave a request the schema will actually accept, not merely a
+    // shorter one — this is the assertion that fails if the two ever drift apart.
+    expect(getEntitiesCoverImage.safeParse({ entities }).success).toBe(true);
   });
 });
--- a/src/components/Notifications/notification-thumbnails.ts
+++ b/src/components/Notifications/notification-thumbnails.ts
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 import { useApplyHiddenPreferences } from '~/components/HiddenPreferences/useApplyHiddenPreferences';
+import { MAX_ENTITIES_COVER_IMAGE } from '~/server/schema/image.schema';
 import { trpc } from '~/utils/trpc';
 
 type WithDetails = { details?: Record<string, unknown> | null };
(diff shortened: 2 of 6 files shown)


ZacxDev:
Pushed a review round (`37cbdc7`). Three things changed; the PR body is updated to match.

**1. The test could not see the bound it was certifying.** The fixture built `entityId: i + 1`, so ids ran `1..n` and co-varied with the array index — which makes a per-element `entityId: z.number().max(500)` accept and refuse exactly the same inputs as the array-level `.max(500)`, and the old assertion `issue.path[0] === 'entities'` equally true of `['entities', 500, 'entityId']`. I applied that relocation mutant against the old test: **4/4 green**, with the array unbounded and every real request 400-ing, since image ids are far above 500. Ids now start at `900_000` and the rejection pins the whole issue (`{ code: 'too_big', path: ['entities'] }`). The same mutant now fails 3 of the 4 tests. The mutation table in the body has the relocation row added — the old table overstated what had actually been checked.

**2. The three callers now clamp their own list.** None of them has an error branch, so a refused query is a blank render rather than a degraded one, and a persistent one — each re-render sends the same list. `useNotificationThumbnails` would lose every thumbnail including those under the bound; `ShowcaseSection` computes `isNullState` from `!coverImages.length`, so it would drop the whole Showcase section off the profile. Both now slice to `MAX_ENTITIES_COVER_IMAGE`, newly exported from the schema module so the number is in one place.

I clamped `ShowcaseItemsInput` too. Its degradation is already better (a per-item "problem loading the cover image" fallback rather than a blank), but on a refusal *every* item shows it including the ones that would resolve, and it reads from the same unbounded-write population as `ShowcaseSection`. It clamps only the query input — the `showcaseItems` state `onChange` saves is left whole, so editing a long showcase can't silently truncate it — and slices after the sort, since the sort is what stops a drag-reorder re-querying.

**3. The schema comment claimed more than the code enforces.** It justified the bound partly on showcases being server-capped at 32. That's true of `addEntityToShowcase`, which truncates to `constants.profile.showcaseItemsLimit`, but not of `userProfile.update` — `showcaseItems` there has no `.max()` and the service stores it as given. Comment corrected to say which path enforces what. Bounding that endpoint is a behaviour change on a different procedure, so it's noted as a follow-up rather than folded in here.

Validation: full unit suite 21473 passed / 25 skipped (1364 files), typecheck 0 errors, ESLint on the changed files identical to the base commit (5 warnings, 0 errors, all pre-existing `no-unused-vars`), Prettier clean.

PR #4326, opened 2026-08-24
feat(moderator): accept a second, inbound-only service token

ZacxDev:
## What

`authenticateWebhookToken` — the only place this app verifies an inbound token — now accepts a second variable, `MOD_INBOUND_TOKEN`, alongside `WEBHOOK_TOKEN`. Inside the app the two are equivalent; the difference is **reach**, not privilege.

## Why

`WEBHOOK_TOKEN` is shared rather than local to this app: it is used in both directions (see `.env.example`), so its value is not ours to change independently and a service that only ever calls *in* had nothing narrower it could be given. `MOD_INBOUND_TOKEN` is inbound-only — nothing here sends it — so a service holding only that one can reach this app and nothing beyond it.

## Migration seam, not a permission model

Either variable alone is a complete configuration. So: move inbound callers over one at a time, then delete `WEBHOOK_TOKEN` from `acceptedTokens()` once none present it.

🔴 **That last step is a code change, not an unset.** Dropping it from `acceptedTokens()` stops accepting it inbound and is safe. *Unsetting the variable* also strips it from this app's outbound callers, two of which degrade by warning rather than failing — and inbound would keep working the whole time, so nothing would be loud. `.env.example` says this at the variable.

Per-endpoint scoping is a separate later change — `EndpointAuth` is already `{kind:'webhook'} | {kind:'session'; page}`, so `{kind:'webhook'; scope}` has somewhere to go.

## This closes a latent auth bypass

Not the goal of the PR — found while auditing it, and now pinned by a test.

`acceptedTokens()` treats an **empty-after-trim** value as unconfigured. The previous guard was `if (!expected)`, which a **whitespace-only** secret passes as truthy; the value was then trimmed to a zero-length buffer, and a zero-length `timingSafeEqual` comparison succeeds. A request presenting `?token=` with no value therefore authenticated, opening every wrapped endpoint.

Verified by probe against both versions, with controls (a real secret still 401s an empty token on both). **Latent only** — a real deployment holds a real value, which I checked before writing this. It is the `REGRESSION: a WHITESPACE-ONLY secret is not configured` test.

## Behaviour worth reviewing

- **Empty is not configured** — otherwise blanking a secret opens every wrapped endpoint. Three tests.
- **No early exit** on a match: every configured candidate is compared, so which credential matched is not revealed by comparison order.
- **503 fail-closed** now requires *both* to be unset, and names both so an operator knows either would fix it.

## Verification

Tests are a **matrix**, not an assertion — this module had no tests before, so both arms were run:

| | at `origin/main` | on this branch |
|---|---|---|
| 7 `REGRESSION:` cases | **all 7 fail** | pass |
| 13 `INVARIANT:` cases | pass | pass |

The invariant tier is the point: accepting a second token must not stop accepting the first.

**Mutation battery**, re-run after the audit fix round (a fix round resets the gate). 9 mutants, each asserted present on disk immediately before its run, source restored byte-identical after, with a positive control in the batch:

- **7 killed**, each attributed to a named test — including the dropped-`.filter()` and disarmed-fail-closed mutants, both killed by the new whitespace-only test.
- **2 survive and are equivalent mutants**: array order and early-return-on-match. Neither is observable through the function's return value. Naming them rather than hiding them — if you think the no-early-exit property deserves a pin, it currently has none.

Also: `app:moderator` 198 passed / 35 skipped (the EXPLAIN tier needs a DB), `pnpm check` 9350 files 0 errors, eslint 3 files 0 errors (read from `--format json`, not an exit code), prettier clean.

## Not in this PR

- **No deployment change.** Nothing sets `MOD_INBOUND_TOKEN` yet, so behaviour in every environment is identical until it does.
- **No credential attribution.** Nothing records *which* token authenticated, so there is no signal that says when the migration is complete. That is a prerequisite for *removing* the legacy token, not for adding the new one — it belongs in the follow-up, where it will have a consumer. Flagging it because the removal step is the one that can break callers.
- **The XGuard docs page still asks operators for `WEBHOOK_TOKEN`.** That audience is an inbound-only consumer and should be moved, but the page cannot point at a variable that does not exist yet.

## Review history

Adversarially audited, then a delta re-audit of the fix round. Findings acted on in `7bc129c748`; two were deliberately deferred, listed above with reasons.

### DIFF
--- a/apps/moderator/.env.example
+++ b/apps/moderator/.env.example
@@ -30,10 +30,22 @@ AUTH_INTERNAL_TOKEN=
 # --- Main-app callbacks ---
 # Shared secret for the main app's token-guarded internal endpoints (e.g. the Meilisearch re-index
 # callback /api/internal/search-index-update). Must match the main app's WEBHOOK_TOKEN. Unset = sync no-ops.
-# Also guards every endpoint wrapped in WebhookEndpoint ($lib/server/webhook-endpoint) — /api/mod/* and the
-# XGuard agent API. Those answer 503 while it is unset. There is no user behind it: nothing it reaches may
-# attribute a write, and revoking access means rotating it for every holder.
+# Also ACCEPTED inbound on every endpoint wrapped in WebhookEndpoint ($lib/server/webhook-endpoint) —
+# /api/mod/* and the XGuard agent API. There is no user behind it: nothing it reaches may attribute a
+# write, and revoking access means rotating it for every holder.
+# 🔴 OUTBOUND USE IS WHY THIS CANNOT SIMPLY BE ROTATED. Four services here present it TO the main app,
+# so its value is pinned by the main app's and is not ours to change independently. Prefer
+# MOD_INBOUND_TOKEN below for anything that only calls IN.
 WEBHOOK_TOKEN=
+
+# Inbound-only service token — accepted on the same WebhookEndpoint routes as WEBHOOK_TOKEN, and
+# equivalent to it inside this app, but it reaches NOTHING beyond this app: no outbound caller sends
+# it and the main app does not know it. Hand this one to a service that only ever calls in (the
+# abuse-detection cronjobs posting to /api/mod/abuse-report) so a leak there cannot touch the main
+# app's admin surface.
+# Either variable alone is a complete configuration; the endpoints answer 503 only while BOTH are unset.
+# A variable set to an EMPTY value counts as unset, so blanking one cannot open the endpoints up.
+MOD_INBOUND_TOKEN=
 
 
 # Signals service, same value as the main app's SIGNALS_ENDPOINT. Session invalidation posts here so a
--- a/apps/moderator/.env.example
+++ b/apps/moderator/.env.example
@@ -45,6 +45,12 @@ WEBHOOK_TOKEN=
 # app's admin surface.
 # Either variable alone is a complete configuration; the endpoints answer 503 only while BOTH are unset.
 # A variable set to an EMPTY value counts as unset, so blanking one cannot open the endpoints up.
+#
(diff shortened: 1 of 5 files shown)


ZacxDev:
Delta re-audit of the fix round found a regression **that round introduced**, now fixed in `4f1b764114`. Recording it here rather than only in a commit message, since anyone who read the branch in between saw the worse version.

**What went wrong:** the second commit rewrote the module comment for brevity and dropped the clause *"the main app presents it when it calls in"* — while keeping, and adding a 🔴 to, the instruction *"once none present `WEBHOOK_TOKEN` inbound, drop it from `acceptedTokens`"*. The warning it added covers only the outbound hazard. Net effect: the comment told a maintainer to check a precondition and deleted the sentence saying where to check it. Grep this repo for an inbound presenter, find none — correctly, since the presenter is the main app — conclude the migration is done, and 401 every delegated moderation action.

The comment now names both ways to get that removal wrong, and states that *"none present it inbound"* is a claim about the **main app's** configuration, not about this repo.

Also folded in: a third site still naming `WEBHOOK_TOKEN` as *the* service credential (`routes/api/mod/[action]/+server.ts:6` — the same defect the previous commit fixed in two other places, and the worst-placed instance, since that route is what the follow-up provisions a token for); and the test docblock, which defined the REGRESSION tier by subject matter when it is actually defined by the matrix.

**One correction to my own earlier commit message.** It said the public-repo disclosure was trimmed. That is true of the module comment and **not** of `.env.example`, whose diff across that commit is purely additive — I did not touch the flagged text there. The re-audit assessed residual risk as ~0 (the route is public, the caller filenames sit in the same public directory, and the file already carried an internal service address on `main`), so I have left the text as-is and am recording the inaccuracy rather than quietly widening the change to make the claim true.

Every number in the PR body was re-derived after this round rather than carried over: matrix 20 green here / 7 red at `origin/main`, all 7 REGRESSION and 0 INVARIANT; mutation battery re-run with each kill attributed to a named test.

PR #4330, opened 2026-08-24
fix(schema-drift): a declared table that does not exist is now a finding, not silence

JustMaier:
Appending a model on an invented table — with an `@id`, a foreign key and a `@@unique` — produced output **byte-identical to a clean run** of `drift:gate`. An absent *column* was at least a pending finding; an absent *table* was nothing at all, so the larger the divergence, the quieter the tool got.

Found while working ClickUp `868ku7w6f` and `868ku8a1g`.

## What was wrong

The catalog captured `relkind IN ('r','p')` only, so "backed by a view" and "backed by nothing" reached the differ as the same thing — `!live.tables.has(model.table)` — and both were skipped wholesale under one reason string. Nothing in the gate's output mentioned skipped models either, which is what kept the branch invisible from both ends.

**45 of 316 models take that branch.** Every column, relation and unique declaration on them is unchecked, and none of them appeared anywhere in the gate's output. `assessCoverage` does not object and is right not to: its threshold is half the relations, and this is 64 of 512.

Measured against production on 2026-08-24 with `pg_catalog.pg_class`:

| | |
| --- | --- |
| views (correctly unmanaged) | 24 |
| **ordinary tables that exist in production right now** | **21** |
| genuinely absent | 0 |

The 21 — `Feedback`, the four `Placement*`, both `UserHub*`, `ModelVersionSale*`, `CollectionInvite`, four `UserCosmeticShop*`, `app_collaborators` and the rest — are missing only because the committed snapshot is from 2026-08-03 and they were created after it (`Feedback`'s migration is `20260813180000_feedback`). So snapshot staleness is not only a shrinking of *blocking* coverage, as the README said. It removes whole tables from coverage, silently.

## What this changes

1. `catalog.ts` captures views (`relkind IN ('v','m')`) into a new optional `DbCatalog.views`.
2. `compare.ts` classifies each skip as `ignored` / `view` / `absent` / `unclassified`, and emits a new **`missing-table`** finding for `absent`, saying how many fields, relations and unique declarations went unchecked with it.
3. The gate and the text report print what they skipped, with counts, on **every** run.
4. 🔴 A snapshot with **no `views` key reports nothing and says so.** Reading an absent list as an empty one would accuse all 24 view-backed models of being missing tables — fabricated findings, in the direction that looks like progress. `missing-table` is therefore **inert until the catalog is recaptured**, which the README now states rather than leaving as a reassuring zero.

## What this deliberately does NOT change

**The exit code.** `missing-table` is `pending`, like `missing-column`, so the gate still exits 0 over both. Promoting either tier is a policy decision (ClickUp `868ku7w6f`) with a measured blast radius: 17 unbaselined pending findings on `main` today would turn the check red for everyone until they are triaged. This PR does not make that call — it makes the gate say what it did not enforce and what it did not look at.

No baseline regeneration. The committed baseline is untouched, and the gate's verdict on `main` is unchanged.

## Controls

With a view-carrying catalog built from the live view list:

| | `NEW pending` |
| --- | --- |
| unmutated schema | **0** |
| + one `model TotallyFakeTable` | **1**, naming `missing-table TotallyFakeTable` |

Reverting the classification in `compare.ts` (forcing every absent table back to `unclassified`) reds three tests with assertions naming wrong values rather than timing out:

```
expected [] to deeply equal [ 'report_view' ]
expected 'unclassified' to be 'view'
expected 0 to be greater than 20
Test Files  2 failed | 15 passed (17)
```

## Verification

- `pnpm run typecheck` — OK, 0 type errors
- `pnpm run lint` / `lint:packages` — no new problems; the 6 `lint:packages` errors are pre-existing and in `civitai-buzz` / `civitai-db-queries`
- `pnpm exec prettier --write <the ten files>` — all already formatted
- `pnpm run test:packages:run` — `Test Files 79 passed | 3 skipped (82)`, `Tests 1118 passed | 8 skipped (1126)`, exit 0
- `pnpm run test:unit:run` — `Test Files 1371 passed | 1 skipped (1372)`, `Tests 21523 passed | 27 skipped (21550)`, exit 0
- `blocks.router.workflow.test.ts` collected **358** tests (submodule present)

## Two things found on the way that are not in this diff

- **ClickUp `868ku8a1g` is wrong about `QuestionRank` / `AnswerRank`.** They are not absent from production — both are **views**. And the four Rank tables it proposes cleaning up have **no all-NULL column between them**: ClubRank 42 rows / 16 cols, BountyEntryRank 119,021 / 46, TagRank 388,494 / 31, CollectionRank 1,158,355 / 16, all populated. UserRank's pattern does not generalise.
- **"Does a migration exist for it" is not a sound discriminator for a phantom declaration**, in either direction. 3 of 23 missing-column findings have no migration and all three columns exist in production (`Announcement.userId`, `.coverId`, `.profileOnly` — applied by hand); 23 of 45 skipped models have no `CREATE TABLE` and all 23 are views.

### DIFF
--- a/packages/civitai-db-schema/src/schema-drift/README.md
+++ b/packages/civitai-db-schema/src/schema-drift/README.md
@@ -78,9 +78,19 @@ regardless of `--strict` rather than reporting a reassuring zero.
 - **Block attributes are read with comments stripped.** `@@map` / `@@ignore` / `@@unique`
   are matched against the model body, so a stray `// @@ignore` left behind by a revert
   would otherwise skip an entire model and every constraint on it, silently.
-- **Models mapped to a view or to a table that does not exist are skipped**, along with
-  models carrying `@@ignore`. They are counted and listed (`--verbose`), not reported as
-  drift: there is no table for a constraint to live on.
+- **Models mapped to a view are skipped**, along with models carrying `@@ignore`: there is
+  no table for a constraint to live on. **A model mapped to nothing at all is a
+  `missing-table` finding**, not a skip. Those two used to be the same branch, and that
+  inverted the tool: an absent column was at least a pending finding, so the bigger the
+  divergence the quieter it got — appending a model on an invented table produced output
+  byte-identical to a clean run. The discriminator is the catalog's `views` list.
+- **Every skipped model is printed, by both the report and the gate**, with its count. A
+  verdict that lists no finding for a model is otherwise indistinguishable from one that
+  never looked at it. On the 2026-08-03 snapshot that is 45 of 316 models.
+- **A snapshot with no `views` key cannot classify a skip**, and says so rather than
+  guessing. Reading an absent list as an empty one would accuse all 24 view-backed models in
+  that snapshot of being missing tables — fabricated findings, in the direction that looks
+  like progress.
 - **Partial unique indexes do not count.** A `WHERE`-clause index enforces uniqueness only
   over the rows it matches; a `@unique` declaration is a promise about every row.
 - **Expression indexes are dropped** rather than matched by name, so a `lower(email)` index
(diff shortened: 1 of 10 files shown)


JustMaier:
🔴 **Converted to draft. Do not merge on the strength of the green CI.**

This work was stood down mid-flight: Justin has parked the schema-gating decision for a team sync (ClickUp https://app.clickup.com/t/868kw6b5u, tagged `discuss`), and the tickets this PR came from are archived. The diff itself is finished and verified — full suite green, controls in the description are real — but an adversarial review of `a328bde54d` raised **11 findings that were never fixed.**

Two of them matter:

- **F1 — a model on a Postgres FOREIGN table (`relkind 'f'`) is now reported as a missing table**, which is false, and contradicts the comment directly above `TABLE_RELKINDS` in `catalog.ts` stating that foreign tables are deliberately excluded so they are *skipped rather than reported*. `TABLE_RELKINDS` is `('r','p')` and the new `VIEWS_SQL` is `('v','m')`, so `'f'` falls into neither and lands in `classifySkip`'s final `absent` branch. **Latent today** — production has 364 `r`, 38 `v`, 2 `m` and **zero** `f` relations in `public` — but it is wrong, and the fix is one character class.
- **F4 — the "print what was skipped" behaviour has zero test coverage.** `formatGateResult` is referenced by no test, and grepping `__tests__/` for `NOT compared`, `MISSING table`, `backed by a view`, `table absent` or `not classified` returns nothing. Deleting `skippedLines`, or returning `[]` from it unconditionally, or setting `skipped: []` in `evaluateGate`'s return, reds nothing. That is precisely the class of defect this PR exists to remove, reintroduced one level up. The gate CLI cases in `gate.test.ts` already assert on stdout, so it is roughly one assertion each.

The remaining nine, briefly:

- **F2** `gate.ts`'s `resolved` section still renders `${table}.${columns.join('+')}`, so a resolved `missing-table` entry prints with a trailing dot. The same shape was fixed in `describe()` and `label()` and missed here.
- **F3** the "NEW drift awaiting a migration" prose says "the column is not in the catalog snapshot" and enumerates `missing-column` / `missing-foreign-key` only — false for `missing-table`, which will dominate that block after a recapture.
- **F6** *(the most valuable one)* — **no positive control on the views read.** The `undefined`-vs-`[]` guard does not help against a *broken* read, which returns `[]`. Its failure mode is identical in shape to the `NOTNULL` bug that `assertCatalogSanity` exists for: ~24 fabricated `missing-table` findings, all in the one direction that looks like progress, non-blocking so nobody has to stop, and absorbable by the very `drift:baseline` run the workflow prescribes next.
- **F7** `absent` models are printed as a bare count while `unclassified` ones are named. Once baselined, matched findings print nowhere, so the 21 absent models become a single integer — the same silence, moved up a level.
- **F5**, **F8**, **F9**, **F10**, **F11** — a dead branch in `report.ts`, `SkipClassification` not re-exported from `index.ts`, an assertion that silently encodes "the schema has zero `@@ignore` models", a test title claiming more than it exercises (and an undocumented consequence: `@@ignore` suppresses a `missing-table` finding), and a "23 views" / "24 views" disagreement across three files. **24 is correct.**

Whoever picks this up: fix F1 and F4 first, re-run the review, or close the PR. The measurements behind the description are independently reproducible and are written up in `_local/docs/plans/lane6-schema-hygiene-findings-2026-08-24.md`.

One correction to the description worth carrying forward regardless of what happens to this branch: **ClickUp `868ku8a1g` is wrong about `QuestionRank` and `AnswerRank`** — they are not absent from production, both are views — and the four Rank tables it proposes cleaning up have no all-NULL column between them.

PR #4332, opened 2026-08-24
fix(buzz): a perkless grant must not strip a paid member's multiplier, and a string forId must not vanish

JustMaier:
Lane 4 — referrals and Buzz correctness. Closes `868kv4q7t`, `868kv5az9`, `868ktbnjh`. **`868kuyu5t` is diagnosed in this lane but has no code here** — see the last section.

## 1. A referral grant outranked a paid membership (`868kv4q7t`)

`userMultipliersCache` ranked a user's subscription rows and read the multiplier off the single winner. Two active rows at the same tier tie on status and tier, so `currentPeriodEnd DESC` decided it — and a `civitai-referral-*` product carries no `rewardsMultiplier`, so `COALESCE(..., 1)` quietly paid a Bronze member 1x.

It now takes the **best multiplier across the user's active subscriptions**, which also covers the `civitai-buzz-*` family in `868kr8bh3`.

**Old query vs new, over every subscriber on the prod replica:**

| users | changed | up | down |
|---|---|---|---|
| 74,575 | 1 | 1 | 0 |

**Blast radius, as the brief required before fixing:** 32 active referral grants; 4 users hold a grant *and* a paid membership; **1** is losing a multiplier today. Historically, 11 redemptions across 10 users show a multiplier drop coincident with the redemption (detector: max multiplier 3 days before vs the 30h–5d window after, run over all 102 rows of `ReferralRedemption`).

**Compensation, decided by Justin 2026-08-24:** pay user 9736923 only — 692 Buzz at last measure, from `buzzTransactions` (270 reward transactions, 1,384 paid, 2,076 owed), still accruing until the fix deploys. The ticket says Green; **all 270 of his reward transactions are `toAccountType: blue`**, so blue is what gets paid. The other nine are not paid: for most, the paid period had already lapsed, and the 1.5x they lost was itself coming from an `active`-status row with a long-expired period.

## 2. The dashboard advertised perks the grants did not deliver (`868kv5az9`)

Justin's call: make the product match the copy rather than trim the copy, and where a grant's multiplier beats the active subscription's, the grant wins. The max-across-active-rows fix gives the second half; the migration adds the two advertised multipliers to the three referral products.

**Both changes together, simulated over the same 74,575 subscribers: 29 gain a multiplier, 0 lose one.**

🔴 **Do not apply that migration before the deploy.** Measured, not reasoned: the deployed build reads `rewardsMultiplier` off the winning row, and for the 26 grant-only users the referral row *is* the winner. Running the **old** query with the migration simulated changes the same 29 users. So applying it early delivers the whole change on the old code path, spread unevenly over 24h with no session refresh — and before the cache-bust fix below exists, an expiring grant would keep paying for a day.

## 3. generation-feedback paid out and recorded nothing (`868ktbnjh`)

`buzzEvents.forId` is `Int32`. Inserts run `async_insert=1, wait_for_async_insert=0`, so ClickHouse accepts the HTTP request, fails to parse the row **server-side**, and drops it — no error reaches the app, and `sendAward` pays anyway.

**Measured:**

| | |
|---|---|
| `buzzTransactions`, 90 days, that description | **4,101,303** transactions / 38,456 users / 20,093,124 Buzz |
| `buzzEvents` rows of type `generation-feedback`, **all time** | **1**, dated 2024-08-01 |
| `system.asynchronous_insert_log`, 2 days | 1,398,041 `Ok` / **92,911 `ParsingError`**, the exception naming `forId` |

Three rewards carry a string key — `generation-feedback` (jobId), `ad-watched` (token), `appBlockReview` (appBlockId) — and none appears in `buzzEvents` in 30 days. That is an enumeration of the reward definitions plus their absence from the table, not a proof that no fourth path exists.

Coercion happens at the insert boundary only, so `sendAward`'s `externalTransactionId` keeps the value it has always used; the raw string is preserved in `transactionDetails.forIdRaw`.

**Two more columns, found by review, both latent:** `status` has no `unqualified` in its `Enum8` (nothing writes it today — live `rewards:config` is `{"rewards": {}}` and the other writer is dead code), and `multiplier` is `Decimal(3,2)` with a ceiling of 9.99 while gold 4 × `MAX_GLOBAL_BONUS` 5 reaches 20. **The multiplier one is a payout, not an audit value** — `process-rewards` reads it back out and `sendAward` pays from it — so the clamp now logs when it fires. `src/server/clickhouse/migrations/2026-08-24-buzz-events-multiplier-width.sql` widens it, as a three-step column swap following the precedent recorded in this table's own `system.mutations` from 2024-04-04.

## 4. `868kuyu5t` — diagnosed, deliberately no code here

The reaction-count read path is an **event-engine** cache defect, not a main-app one. `metrics:Image:140276941` in prod holds `Like 1` while `entityMetricDailyAgg_v2` holds `Like 6, Heart 4`. It cannot self-heal: `MetricService` repopulates only on a *missing* key, and a 10% per-read TTL slide keeps a viewed image's entry alive indefinitely. Three candidate fixes are on the ticket; all require restarting a service every agent's dev server depends on, so it is left as a decision rather than done quietly.

## Applied by hand — two migrations, in this order, after the deploy

1. `20260824180000_referral_grant_tier_multipliers` (Postgres) — then confirm **3** rows changed, then 32 exact `DEL`s on `packed:caches:multipliers-for-user:<userId>`, then `POST /api/admin/refresh-user-sessions`.
2. `2026-08-24-buzz-events-multiplier-width.sql` (ClickHouse) — three steps with a stop condition at each. **There is no rehearsal environment**: dev and preview both write to the production cluster.

## What this does NOT fix

- The ~4M historical `generation-feedback` rows are gone. No backfill.
- ~3,167 rows stuck at `pending` (2,209 `goodContent:image` back to 2026-07-25) stay stuck. Cause unknown — the leading hypothesis is on `868ktbnjh`, recorded as a hypothesis.
- `buzzEvents` has a second writer this PR cannot reach: `apps/moderator/src/lib/server/rewards.ts:38`, unclamped. Filed as `868kw9b31`.
- Ineligible users are still paid — `sendAward` turns the fold's correct 0 back into 1. 551 rows / 23 users / 6,238 Buzz in 7 days. Pre-existing. Filed as `868kw9kfk`.

## Review

Five-lane `civitai-review` over `0759aa09c1`, plus an earlier adversarial pass over `6ae7d8a1b6`. All five lanes accounted for by name; two went idle without delivering and were chased. Findings fixed in `3477512f13`.

**The fix round is not itself re-reviewed.**

Every test here was checked by reverting the fix and reading the failure. The wiring guard is a **tripwire, not coverage** — it is a source scan, and its first version pinned one spelling and passed a mutation to `SELECT DISTINCT ON (u.id)` that restored the bug completely. It now checks four idioms, each proven by a mutation asserted to have applied first.

```
typecheck: OK — 0 type errors
Test Files  1373 passed | 1 skipped (1374)      exit 0, counts sum
Tests  21553 passed | 27 skipped (21580)
test:lint-rules  17 files / 323 tests           exit 0
blocks.router.workflow.test.ts  358 tests       (submodule initialised)
```

Two earlier full-suite runs came back red and both background notifications reported exit 0 over them: a vitest worker died, and the file counts gave it away — `1370 passed | 1 skipped` against a total of `1374`. `--max-workers=8` produced the run above.

---

## 🔴 Provenance of the final commits — read this before trusting "reviewed"

The five-lane review ran against **`0759aa09c1`**, and a follow-up correctness pass against **`cc1adcfed4`**. Three commits landed after those and were **not** covered by any review pass:

| commit | contents | evidence it shipped on |
|---|---|---|
| `3477512f13` | the fix round for the five-lane findings | re-reviewed at `cc1adcfed4` ✅ |
| `cc1adcfed4` | migration rewritten as the column swap (`.sql` only) | re-review ✅ |
| `6f3948aefc` | **batch-log refactor** + the ClickHouse-cancelled comments | covering suites (240 tests) + clean typecheck — **no full suite, no review** |

`6f3948aefc` moved the clamp's Axiom log from per-event to once-per-batch and changed both `clickhouse.insert` call sites from `toClickhouseBuzzEvent(event)` to `toClickhouseBuzzEvents([event])`.

**It was merged before its full suite finished** — that run was still queued behind another agent's when the merge went in, on Justin's explicit instruction. **The run completed green shortly afterwards**, so the evidence exists, it just post-dates the merge:

```
UNIT_EXIT=0                                   (the command's own status, read immediately)
Test Files  1373 passed | 1 skipped (1374)    1373 + 1 = 1374, counts sum
Tests  21553 passed | 27 skipped (21580)      21553 + 27 = 21580
no "Worker exited", no failing files
blocks.router.workflow.test.ts  358 tests     submodule live
```

Still no *review* pass over that commit. Recorded here so nobody six weeks from now infers coverage from timestamps.

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
@@ -9,6 +9,13 @@
 -- stipend and no badge. Only the two advertised multipliers are added.
 --
 -- Every numeric value in Product.metadata is stored Stripe-style, as a string.
+--
+-- 🔴 AFTER APPLYING: purge the multiplier cache. userMultipliersCache has a 1-day TTL and is only
+-- busted per user; nothing invalidates it when a Product changes. Without a purge, grant holders
+-- keep the perkless multiplier for up to 24h at whatever rate their entries happen to expire:
+--
+--   purge the 'packed:caches:multipliers-for-user' prefix (REDIS_KEYS.CACHES.MULTIPLIERS_FOR_USER)
+--   with the cache-purge skill, then POST /api/admin/refresh-user-sessions
 
 UPDATE "Product"
 SET metadata = metadata || '{"rewardsMultiplier":"1.5","purchasesMultiplier":"1.05"}'::jsonb
--- a/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
@@ -14,8 +14,20 @@
 -- busted per user; nothing invalidates it when a Product changes. Without a purge, grant holders
 -- keep the perkless multiplier for up to 24h at whatever rate their entries happen to expire:
 --
---   purge the 'packed:caches:multipliers-for-user' prefix (REDIS_KEYS.CACHES.MULTIPLIERS_FOR_USER)
---   with the cache-purge skill, then POST /api/admin/refresh-user-sessions
+-- Confirm 3 rows were updated. A 0-row UPDATE reports success, and these ids exist only in the
+-- database — nothing in src/ references them, so a typo has nothing to fail against:
+--
+--   SELECT id, metadata->>'rewardsMultiplier', metadata->>'purchasesMultiplier'
+--   FROM "Product" WHERE metadata->>'referralGrantable' = 'true';
+--
+-- Then bust the cache for the affected users only. There were 32 active referral grants at the
+-- time of writing, so 32 exact DELs on 'packed:caches:multipliers-for-user:<userId>'
+-- (REDIS_KEYS.CACHES.MULTIPLIERS_FOR_USER) beat a cluster-wide SCAN of the whole prefix:
+--
+--   SELECT "userId" FROM "CustomerSubscription" cs JOIN "Product" p ON p.id = cs."productId"
+--   WHERE p.metadata->>'referralGrantable' = 'true' AND cs.status IN ('active','trialing');
+--
+-- then POST /api/admin/refresh-user-sessions for those ids.
 
 UPDATE "Product"
 SET metadata = metadata || '{"rewardsMultiplier":"1.5","purchasesMultiplier":"1.05"}'::jsonb
--- a/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260824180000_referral_grant_tier_multipliers/migration.sql
@@ -1,0 +1,23 @@
+-- Referral grants deliver the tier multipliers the referral dashboard advertises.
+--
+-- The three `civitai-referral-*` placeholder products carried no multipliers, so a redeemed
+-- grant delivered neither the "1.5x daily rewards" nor the "1.05x purchases" its own
+-- redemption card promises (ClickUp 868kv5az9). Decision by Justin, 2026-08-24: make the
+-- product match the copy rather than change the copy.
+--
+-- `monthlyBuzz: 0` and `badgeType: "none"` stay as they are — the grant still conveys no Buzz
+-- stipend and no badge. Only the two advertised multipliers are added.
+--
+-- Every numeric value in Product.metadata is stored Stripe-style, as a string.
+
+UPDATE "Product"
+SET metadata = metadata || '{"rewardsMultiplier":"1.5","purchasesMultiplier":"1.05"}'::jsonb
+WHERE id = 'civitai-referral-bronze';
+
+UPDATE "Product"
+SET metadata = metadata || '{"rewardsMultiplier":"2.5","purchasesMultiplier":"1.10"}'::jsonb
+WHERE id = 'civitai-referral-silver';
+
+UPDATE "Product"
+SET metadata = metadata || '{"rewardsMultiplier":"4","purchasesMultiplier":"1.20"}'::jsonb
+WHERE id = 'civitai-referral-gold';
--- a/src/server/clickhouse/migrations/2026-08-24-buzz-events-multiplier-width.sql
+++ b/src/server/clickhouse/migrations/2026-08-24-buzz-events-multiplier-width.sql
@@ -1,0 +1,38 @@
+-- buzzEvents.multiplier cannot hold the value the app computes.
+--
+-- The column is Decimal(3, 2) — 3 digits total, 2 after the point — so its ceiling is 9.99.
+-- The stored value is the tier multiplier TIMES the global bonus event, and gold's 4 against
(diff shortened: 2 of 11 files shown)


JustMaier:
## Review round — pushed `0759aa09c1`

An adversarial review over `6ae7d8a1b6` found the same defect in two more `buzzEvents` columns. Both verified against prod before acting.

**`status` is `Enum8('pending','awarded','capped')`.** `unqualified` is not in it, and both `base.reward.ts` (disabled-reward sweep) and `goodContent.reward.ts` write it. `updateBuzzEvents` sends a 1000-row chunk in one call, so one such row takes the `awarded`/`capped` updates riding with it — after `sendAward` has already paid.

Measured with a control for the in-flight window (pending **and** older than 6h):

| type | stuck | oldest |
|---|---|---|
| goodContent:image | 2,209 | 2026-07-25 |
| collectedContent:image | 725 | 2026-07-26 |
| collectedContent:model | 210 | 2026-07-25 |

Mapped to `capped` with `statusRaw` preserved in `transactionDetails`. **Widening the enum is the cleaner fix** and is a production ClickHouse DDL, so it is left as a call for whoever owns that.

**`multiplier` is `Decimal(3, 2)`, ceiling 9.99.** Gold's `4` × `MAX_GLOBAL_BONUS` of `5` reaches 20. Max observed over 90 days is 8 and nothing is at or above 9, so latent rather than live — a 3× event would have dropped every gold member's events. Clamped, raw value preserved.

Also: the numeric-`forId` branch now requires strict digits (`Number(' 42 ')` was `42`, colliding with `'42'` under `ORDER BY (type, toUserId, forId, byUserId)`); the sub-1 multiplier floor is gone, restoring what the `COALESCE` it replaced did; and the source guard asserts both slice bounds instead of widening to the whole file if a marker moves.

`base.reward.config.test.ts` was the **second** test in this PR pinning a write nobody received — it asserted `status: 'unqualified'` reached the insert.

### Controls, all four fired on revert

```
expected 20 to be 9.99
expected 42 not to be 42
expected 1 to be 0.5
expected "vi.fn()" to be called with arguments: [ ObjectContaining{…} ]
```

### Verification on `0759aa09c1`

```
typecheck: OK — 0 type errors
Test Files  1373 passed | 1 skipped (1374)      exit 0
Tests  21546 passed | 27 skipped (21573)
test:lint-rules  17 files / 323 tests           exit 0
blocks.router.workflow.test.ts  358 tests
```

Two earlier runs of that suite came back red and are **not** being waved away: a vitest worker died in both (`Worker exited unexpectedly`), and the file counts gave it away — `1370 passed | 1 skipped` against a total of `1374`, so a file never ran. Capping `--max-workers=8` produced the run above, whose counts sum correctly. Neither file that failed has an import edge to anything in this PR.

### Not addressed here, deliberately

- **No `currentPeriodEnd` guard** on the subscription join. 17 rows in prod are `active` with a past period. Pre-existing, but after this migration a stale `civitai-referral-gold` left `active` would confer 4× indefinitely. Adding the guard takes multipliers away from live users, so it needs a decision rather than a drive-by.
- The two multipliers are maxed independently, so a user could in principle compose a rewards/purchases pair no product sells. Benign with today's catalogue, and it is what was asked for.

PR #4339, opened 2026-08-24
fix(apps): stop /apps/run growing a second scrollbar around the block iframe

ZacxDev:
`/apps/run/<slug>` painted **two vertical scrollbars** — the site layout's `ScrollArea` grew one, the block's own document inside the iframe had another. Only the inner one scrolled anything a viewer wanted.

Turned out to be **two independent causes**, both measured in Chromium rather than reasoned about.

## 1. The host claimed more height than it could ever have

The route shipped a bare `export default function AppPage`, so it inherited `AppLayout`'s default `scrollable: true` — an `overflow-y: auto` `ScrollArea`. Inside it, `PageBlockHost` claimed `min-height: calc(100dvh - 60px)`: the viewport minus the site header, and nothing else. The space actually available is

```
100dvh − header − subNav − its mb-3 − RewardsBonusBanner − AppFooter − AdhesiveAd
```

Every term after `header` is ≥ 0 and several are > 0 on a normal render, so the host was **unconditionally** taller than its own scroll viewport. Not a race, and not tunable — there is no window size at which the two agree.

## 2. The launch-reveal transform escaped its wrapper

While a block is still handshaking, the iframe carries `translateY(8px)`. A transform doesn't change layout but **does** extend the scrollable overflow region, so those 8px asked the nearest scrolling ancestor for a scrollbar. Measured: wrapper `bottom=716`, iframe `bottom=724`, container `scrollHeight 724` vs `clientHeight 716`. Not run-page-specific — it applied to every surface.

## The fix

- **`PageBlockHost` gains an explicit `fit` prop.** `'viewport'` is the default and preserves the old sizing, so the **dev tunnel** and the **mod-review preview** keep their behaviour — both mount inside scrolling ancestors that bound nothing, and would collapse on `'fill'`. `'fill'` fills an already-bounded parent and claims no viewport-derived height. (One thing *does* change for all surfaces: the reveal wrapper's new `overflow: hidden`, i.e. cause 2.)
- **The run page declares `Page(AppPage, { scrollable: false, subNav: null })`** and passes `fit="fill"`, with a wrapper `Box` that grows and provides one scroll container of last resort.
- **`FILL_MIN_HEIGHT_PX = 300`** floors the `fill` host so a short viewport squeezes it but never strands it. See "post-audit" below — this was not in the first version and is the most important change in the PR.
- **The reveal wrapper gets `overflow: hidden`,** confining a purely decorative transform.

### ⚠️ The run-page contract is a co-requisite of THREE legs, in three files

None is safe alone:

| shipped without | result |
|---|---|
| `scrollable: false` | host claims more height than the `overflow-hidden` chain allows → bottom of the app **clipped** |
| `fit="fill"` | `flex: 1` resolves against an auto-height parent → iframe **letterboxes to ~150px** (its intrinsic replaced-element height) |
| the run page's wrapper `Box` | same as above — and this leg is **invisible to the browser suite**, which builds its own fixture wrapper rather than importing the page |

All three are pinned in the node guard, which now compares whole normalised expressions rather than checking for tokens.

## What the pre-merge audit changed

An adversarial audit was run before merge. It found no deploy-blockers but did find a real regression I'd introduced and several vacuous assertions. All are fixed here:

**🟡 The fix traded a cosmetic scrollbar for an unreachable one.** With `scrollable: false` every ancestor is `overflow-hidden`, so the page is the only thing that can offer a scrollbar — and the first version floored the host at `min-height: 0`. The site's fixed chrome can't shrink (header 60 + `AppFooter` 45 + `mt-3` 12 + `AdhesiveAd` 90/50), so on a short viewport the app absorbed the whole shortfall with nothing to scroll. Measured: a ~360px phone-landscape viewport left **153px**, and a 250px viewport collapsed it to **0**, `outerScrollbar = false` throughout. A 1366×768 laptop at **200% zoom** lands in the same band. That is WCAG 1.4.4 / 1.4.10 — strictly worse than the bug being fixed. Cured by `FILL_MIN_HEIGHT_PX` plus `overflowY: 'auto'` on the page wrapper, and pinned at a short viewport **and** a normal one so a floor that fired for everyone would fail.

**🟡 …and a second audit round found that fix unpinned, and its value wrong.** The floor assertion read `height >= FILL_MIN_HEIGHT_PX` — comparing the measurement against the constant that produced it, so true by construction. Measured: `FILL_MIN_HEIGHT_PX = 0` (a full revert of the WCAG fix) passed **all 17 tests**. Now asserted against absolute literals plus a band on the constant; floors of 0/100/200/400/800 are all killed. The value also moved **400 → 300**: the floor is a trade (below it the page grows a scrollbar beside the block's own), and at 400 that band included **375×667-class phones in portrait** — mainstream devices held normally, not the "landscape / 200% zoom" band I'd claimed. My justification had also subtracted chrome from the *screen* height rather than `innerHeight`, overstating the margin on a 768px laptop as 161px when it is ~43px.

**🟡 The gating suite couldn't see the host's styling at all.** A 17-mutant sweep found **9 survivors**, all invisible to a token grep — inverting the ternary (`fit === 'fill'` → `!==`, which breaks *every* surface at once), `flex: 1` → `flex: 0`, deleting the cause-2 `overflow: hidden`, and reverting the page wrapper. Now pinned as whole normalised expressions.

**🟡 Two of my own assertions were vacuous, found by re-running the sweep:**
- `scrollHeight > clientHeight` does **not** mean "reachable" — an `overflow: hidden` box reports the same. Flipping the fixture to `hidden` left all 8 browser tests green. Now asserts computed `overflow-y ∈ {auto, scroll}` as well.
- `expect([...]).toContainEqual(new Set([...]))` does **not** deep-compare Sets in Vitest — measured: it passes when the expected set is absent. That assertion silently tested nothing. Now compares sorted arrays.

**🟡 `subNav: null` is a product decision, not layout — DECIDED: keep it.** It removes `<HomeTabs />`, the row of top-level section pills (~52px with its `mb-3`), from this route. Reasoning, so a reviewer can overturn it on the merits:

- **Top-level navigation is not lost.** `header` stays at its default, and `AppHeader` was confirmed present on the live preview — logo, search, Create, Buzz balance, account menu. Only the *second-level* tab row goes, and its destinations remain reachable from the header.
- **The precedent is exact.** `src/pages/generate/index.tsx:110-113` and `src/pages/research/rater.tsx:783-786` — this repo's two other immersive `scrollable: false` surfaces — both pass `subNav: null`.
- **It is ~52px of the budget this PR spent four audit rounds on.** A 375×667 phone leaves the page ~386px; a competing nav strip costs more there than it returns.
- **Two chrome bars over a third-party app reads badly** — the route already renders `AppBlockChrome` (the "Apps / <name>" breadcrumb), so the sub-nav would stack a second navigation row above someone else's UI.

The counter-argument is consistency with the rest of the site, which is what the two surfaces above already decided against. Dropping the line is a one-word change if a reviewer disagrees. (Also: `RewardsBonusBanner` still renders regardless, so it is not removed as the first version of this description implied.)

**🟡 I claimed the browser suites "are not run in CI at all". That is wrong** — see the correction comment below.

## Tests

**`__tests__/pageRunScrollContract.test.ts`** — node `unit`, the gating project. Pins all three legs, the preserved `'viewport'` default, that only a `PageBlockHost` mounter can opt into `fill`, that `AppLayout`'s two branches still differ as assumed, and that `PageBlockHost`'s hardcoded `60` still equals `AppHeader`'s module-private `HEADER_HEIGHT`.

Round-2 mutation sweep — **13/13 real mutants killed**, each by its intended assertion, with every mutant byte-verified as actually applied (one first-round "survivor" turned out to be a replacement that never matched):

| mutant | killed by |
|---|---|
| invert ternary `fit !== 'fill'` | fit-branch pin |
| fill branch `flex: 1` → `flex: 0` | fit-branch pin |
| floor → `minHeight: 0` (the WCAG squeeze) | fit-branch pin |
| delete / comment out reveal `overflow: hidden` | reveal-wrapper pin |
| revert page wrapper to `{width:'100%'}` | page-wrapper pin |
| drop `minHeight: 0` / `overflowY: 'auto'` from it | page-wrapper pin |
| drop `scrollable: false` / drop `fit="fill"` | both-halves + co-requisite |
| host default → `'fill'` | default test |
| `HEADER_HEIGHT` 60 → 64 | second-copy tripwire |
| drop `overflow-hidden` from the no-scroll `<main>` | layout-premise test |

Plus a **negative control**: a pure Tailwind class *reorder* of that `<main>` (a no-op) correctly **survives**, so the pin isn't merely brittle.

**`PageBlockHostScrollFit.browser.test.tsx`** — measures outcomes in real Chromium instead of reading back the styles under test. Keeps a **red arm** (`fit="viewport"` really does overflow), a fixture positive-control, a collapse guard, and the short-viewport pair. Cause 2 and the squeeze regression were both **found** by this file.

## Verification

✅ `pnpm typecheck` — 0 errors
✅ node `unit`: 2993 tests / 174 files
✅ eslint (`--max-warnings=0`) + prettier clean
✅ browser `component` suites — see final comment

❌ **Still not verified against the live page.** Everything above is measured against the components in Chromium; nobody has loaded `/apps/run/<slug>` on the preview and looked. Worth doing **at a narrow window height** given the squeeze finding.

## Follow-up, not in this PR

`PageBlockHost` hardcodes `60` as a second copy of `AppHeader`'s private `HEADER_HEIGHT`, a real one-rule-two-places hazard for the surfaces still on `'viewport'`. This PR adds a tripwire rather than consolidating, since exporting the constant touches unrelated files. **Closing condition:** `PageBlockHost` reads the header height from a single exported source and the tripwire assertion is deleted.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -224,8 +224,13 @@ const REVIEW_NACK_MESSAGE = 'not available in review preview';
  * a narrower window. Below the floor that is the right call (a scrollbar beats
  * unreachable content), so the number's only job is to make the band as small as
  * it can be while still catching the degenerate case. Raising it widens the
- * population that sees two scrollbars; `PageBlockHostScrollFit.browser.test.tsx`
- * bounds it on BOTH sides for that reason.
+ * population that sees two scrollbars, so the value is bounded on BOTH sides —
+ * in `__tests__/pageRunScrollContract.test.ts` (the GATING node suite; that is
+ * the copy that can block a merge) and again in
+ * `PageBlockHostScrollFit.browser.test.tsx` (report-only, which is why it is not
+ * the only one). The NUMBERS deliberately live in those tests, not here: a band
+ * declared beside the value it bounds can be moved in the same edit. What lives
+ * here is the ARITHMETIC that justifies them.
  *
  * THE ARITHMETIC, which decides who lands in the band:
  *
@@ -398,12 +403,13 @@ export interface PageBlockHostProps {
    *     Correct ONLY for a mounter sitting inside a SCROLLING ancestor that does
    *     not otherwise bound its height: the dev tunnel (`/apps/dev/<blockId>`,
    *     default `AppLayout` → `ScrollArea`) and the mod-review preview (inside a
-   *     modal). Without it the host would collapse to the chrome bar and the
-   *     iframe would be sized only by `FILL_MIN_HEIGHT_PX` — measured 300px of
-   *     host, 31px of chrome, 269px of iframe, regardless of how much room the
-   *     page actually has. Usable, but no longer FILLING anything.
+   *     modal). Without it those hosts would be sized only by
+   *     `FILL_MIN_HEIGHT_PX` — measured 300px of host, 31px of chrome, 269px of
+   *     iframe, regardless of how much room the page actually has. Usable, but
+   *     no longer FILLING anything.
    *
-   *   'fill' — the host fills its parent EXACTLY (`flex: 1; min-height: 0`) and
+   *   'fill' — the host fills its parent (`flex: 1`, floored at
+   *     `FILL_MIN_HEIGHT_PX`) and
    *     claims no viewport-derived height of its own. Requires the mounter's
    *     ancestor chain to already bound the height — i.e. a page declared
    *     `Page(…, { scrollable: false })`, whose `AppLayout` branch is
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -204,6 +204,39 @@ type Status = 'loading' | 'ready' | 'timeout' | 'fatal' | 'no_token' | 'error';
 // (gotcha #73). Module-scope so referencing it in a handler adds no effect dep.
 const REVIEW_NACK_MESSAGE = 'not available in review preview';
 
+/**
+ * The floor a `fit="fill"` host will not shrink below, in px.
+ *
+ * 🔴 WHY A FLOOR EXISTS AT ALL. `fit="fill"` is only correct under a
+ * `Page(…, { scrollable: false })` layout, where EVERY ancestor is
+ * `overflow-hidden`. That is what removes the double scrollbar — and it also
+ * means the page is the only thing left that can offer a scrollbar. The site's
+ * fixed chrome cannot shrink (header 60 + `AppFooter` 45 + its `mt-3` 12 +
+ * `AdhesiveAd` 90 desktop / 50 mobile), so with a bare `min-height: 0` the app
+ * absorbed the ENTIRE shortfall on a short viewport and the overflow was simply
+ * clipped away with nothing to scroll.
+ *
+ * Measured before this floor existed: a ~360px phone-landscape viewport left the
+ * host 153px, and a 250px viewport collapsed it to 0 — with `outerScrollbar =
+ * false` at every step. A 1366×768 laptop at 200% browser zoom lands in the same
+ * band. That is WCAG 1.4.4 / 1.4.10 (content unreachable on zoom/reflow), i.e.
+ * strictly worse than the cosmetic spare scrollbar the `fill` mode was
+ * introduced to remove.
+ *
+ * 400 is chosen to sit ABOVE the ~395px a real block renders at (and above the
+ * synthetic install's `minHeight: 200`) while staying below what any ordinary
+ * viewport leaves: a 768px laptop has 561px after chrome, so the floor never
+ * fires there. It is deliberately NOT derived from `100dvh` — that arithmetic is
+ * exactly what this PR removed.
+ *
+ * The floor only helps if something can scroll once it binds, so the run page's
+ * wrapper carries `overflowY: 'auto'` as the scroll container of last resort.
+ * Both halves are asserted in `PageBlockHostScrollFit.browser.test.tsx`
+ * ("short viewports — squeezed, but never unreachable"), at a short viewport AND
+ * at a normal one so a floor that fired for everyone would fail.
+ */
+export const FILL_MIN_HEIGHT_PX = 400;
+
 export interface PageBlockHostProps {
   /** AppBlock id (`apb_*`) — used to build the BLOCK_INIT ids + trust chrome. */
   appBlockId: string;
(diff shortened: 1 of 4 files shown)


ZacxDev:
### Correction — the original description of this PR was wrong on two points

Flagging these publicly rather than silently editing the body, since anyone who read the first version read the wrong thing.

**1. "CI runs the node `unit` project only — the browser suites are not run in CI at all."** This is false. I asserted it in the PR body *and* in the header comments of both new test files, and I never checked it.

The component suites **do** run in CI: `pr-preview-pipeline.yaml` has a `component-tests` task running `pnpm run test:component`, surfaced as the `preview / component-tests` commit status — which is present and green on this PR. It is **report-only** (the task always exits 0; the pipeline comments that making it gating means changing that step to `exit "$RC"`), not absent.

Why it matters: the practical conclusion I drew is still right — the node guard is the *gating* half, and that is where the co-requisite belongs. But a reviewer told "not run at all" would not think to look at a red `preview / component-tests`, and that status is exactly the signal that catches the mutants the node guard couldn't see (ternary inversion, `flex: 0`, the dropped `overflow: hidden`). Corrected in the body and in both file headers.

**2. "the iframe renders 0px tall"** — describing what happens if `fit="fill"` ships without `scrollable: false`. Measured, it letterboxes to **~150px**, the iframe's intrinsic replaced-element height; `0px` is what `flex: 0` produces, which is a different mutation. Still broken, but "a 150px letterbox" announces itself less loudly than "blank page", which changes how quickly a half-revert would get noticed. Corrected.

Both came out of the pre-merge audit, along with a genuine WCAG regression I'd introduced and two assertions of mine that were passing vacuously. All are described in the updated body and fixed in the follow-up commit.

ZacxDev:
### Verified on the live preview — with a scope caveat that matters

Measured on `pr-4339.civitaic.com/apps/run/gen-matrix` in a real logged-in moderator session.

**At 1498×1152 — the reported bug is gone:**

```
docOverflows = false  (1152 / 1152)        ← no outer page scrollbar
IFRAME  h=972   oy=clip
DIV     h=972   oy=hidden                  ← reveal wrapper (cause 2), live
DIV     h=1003  oy=visible                 ← host frame
DIV     h=1003  oy=auto   sh=1003 ch=1003  ← page wrapper, nothing overflowing it
MAIN    class="flex flex-1 flex-col overflow-hidden"   ← the no-scroll branch, live
```

Exactly one scroll surface: the block's own document inside the iframe.

**At 900×380 — the floor and the scroll-of-last-resort behave as designed:**

```
DIV h=400  oy=visible                      ← host held at its floor
DIV h=231  oy=auto  sh=400 ch=231 <<SCROLLS   ← wrapper scrolls; content reachable
```

Squeezed, never stranded — which is the WCAG finding from the first audit round, working.

---

### 🔴 Two caveats, both of which change what this proves

**1. The preview is built from `c48ba4d029`, not HEAD.** The host floored at **400**, and the floor was retuned to **300** in `c703c09e9e`. So this verifies the original double-scrollbar fix and the floor mechanism; it does **not** verify the 300 retune, the blocking-tier band guard, or anything after that commit. Those remain test-verified only.

**2. This build strips every `data-testid`** — `[data-testid]` returns **0 elements page-wide**. My first probe used `[data-testid="app-page-frame"]`, got nothing, and read as "the host isn't rendering". It was rendering fine; the selector was blind. Flagging it because a testid-based probe against this preview will lie to the next person the same way — read the structure instead.

No screen was taken to do any of this: reads, device emulation and teardown all ran on a hidden background tab.

PR #3113, opened 2026-07-14
Detect model file metadata

rmatif:
Uses SafeTensor headers as the source of truth for file precision and unambiguous file type corrections. Keeps weight precision in the tensor API and overview while reusing cached tensor data.

PR #3046, opened 2026-07-10
Expose model weight precision

rmatif:
Adds content-derived FP8/BF16-FP16 precision to the selected model file response. Reuses fingerprinted tensor caches and fails open on slow or unavailable metadata.

PR #4344, opened 2026-08-24
feat(model-files): correct file precision and type from the tensor header

JustMaier:
Closes [868kbt78b](https://app.clickup.com/t/868kbt78b). **Supersedes #3113 and #3046** (both @rmatif). His #3113 covers the same ticket and his detection logic is the basis for the type half here — but its base was 2251 commits behind `main` and it was `CONFLICTING`, which in this repo means it had never had CI at all. Justin's call was a fresh PR using his as reference. Co-authored below.

## What it does

The safetensors header is the source of truth. Where the stored record disagrees, it is corrected on read:

- **precision** — `ModelFile.metadata.fp`
- **type** — `ModelFile.type`, only where the header is unambiguous *and* the correction is safe (below)

It runs where the header is already parsed, so a file is repaired the first time anyone views it. No backfill, no migration.

## Reuse rather than a second implementation

`src/utils/file-helpers.ts` already had the dtype→precision mapper backing the upload form, and it encodes two things a fresh implementation gets wrong: **U8 stays unmapped** (bitsandbytes packs nf4/fp4 there, so guessing `int8` fills in a wrong supported value) and **MXFP8 is identified by its `F8_E8M0` scale tensors**, not by byte share. Extracted as `getDominantFpFromDtypes()`; upload and server now share one owner.

## Two safety rules that narrow "the header wins", deliberately

**1. The header never overwrites a value it cannot state.** `fp8_scaled`/`fp8_mixed` are scaling schemes layered on an F8 dtype. `nf4`/`nvfp4`/`int4` — and GPTQ/AWQ `int8` — pack their weights into `U8`/`I32`, which the dtype map deliberately leaves unmapped, so the quantized bulk contributes **nothing** to the vote and the winner is whatever leftover fp16 housekeeping tensors are present. Without this rule a creator's `nf4` becomes `fp16` on the first anonymous view, and again after every manual fix. `mxfp8` is on the list too: it is inferred from a side channel an exporter may not provide.

The test is **silence, not emittability** — those come apart, and the emittable-means-safe proxy fails open. Cost: a file already labelled with one of those seven is never re-corrected. An empty `fp` is still filled; a wrong `fp32`/`fp16`/`bf16`/`fp8` is still fixed.

**2. A type correction may never demote a file out of its model type's primary set.** The detection reads an LLM's own weights as a text encoder (`embed_tokens` + `layers.N`) and a VLM's as a vision encoder. Relabelling either drops the file out of `primaryFileTypesByModelType` and off the download path. A file that is *not* currently primary stays correctable — that is the mis-filed case this exists for.

## Shape

- `getModelFileHeaderCorrections()` — pure, no database. Upload-time can reuse the judgement as-is.
- `applyModelFileHeaderCorrections()` — one `UPDATE`, guarded on `id` **and** `url`, with `IS DISTINCT FROM` predicates so `updated > 0` means what it says (Postgres reports a no-op update as 1). The jsonb shape guard is a `CASE` on the **assignment**, not the row match: the `changed` branches are OR-ed, so a row matched on `type` alone would otherwise still run the merge — which raises on scalar metadata and *silently appends to a jsonb array*.
- `correctModelFileFromTensorHeader()` — derive + apply.

`detectedModelType` is computed once in `analyzeModelTensors` so it lands in the cache instead of re-scanning every tensor name per request. Optional: pre-existing cache entries carry no value, which reads as "not known". Precision comes from `dtypeCounts`, which every entry has, so it corrects on both the summary and full paths. The correction is fired **after** the response — a read must not wait on a write.

## Verification

```
pnpm run typecheck                      OK — 0 type errors
pnpm exec eslint <changed files>        0 errors (11 pre-existing warnings in two test files)
affected suites                         Test Files 5 passed (5) / Tests 141 passed (141)
blocks.router.workflow.test.ts          358 tests — submodule confirmed present
```

⏳ **The full unit suite is still queued** on the shared box at the time of opening; I will post the `Test Files` line verbatim with its exit code as a comment when it returns. The previous full run on this branch was `Test Files 1 failed | 1375 passed | 1 skipped (1377)` — the single failure was mine (`no-direct-shared-module-mock`) and is fixed in `fc037b9a3d`.

**Thirteen positive controls, each reverted deliberately, each red with a legible assertion** — not "I added an assertion and it passed":

| reverted | failure |
|---|---|
| endpoint drops both `correct()` calls | `expected "vi.fn()" to be called 1 times, but got 0` |
| endpoint drops `url: true` from the select | `expected { Object } to match object { url: true, … }` |
| endpoint swaps `currentFileType`/`modelType` | `expected 'Checkpoint' to be 'Model'` |
| `await correct()` before the response | `expected undefined to be 200` |
| never-overwrite list removed | 5 red, `expected { fp: 'fp8' } to deeply equal {}` |
| list over-widened with `fp32` | `expected {} to deeply equal { fp: 'bf16' }` |
| `wouldDemoteFromPrimary` removed | 3 red, `expected { type: 'Text Encoder' } to deeply equal {}` |
| `IS DISTINCT FROM` removed | 4 red |
| `CASE WHEN` on the merge removed | `expected 'SET "metadata" = COALESCE(…' to contain 'CASE WHEN'` |
| `analyzeModelTensors` stops detecting | `expected null to be 'VAE'` |
| NaN coerced to 0 | `expected 'fp32' to be null` |
| UNet checked before Checkpoint | `expected 'TextEncoder' to be 'Checkpoint'` |
| `'LORA'` mis-spelled `'LoRA'` | `expected 'Enhancement LoRA' to be null` |

🔴 **One of those controls initially stayed GREEN**, and it is the reason this list exists. Deleting `url: true` left all ten endpoint tests passing — because `findUnique` is mocked, it returns the whole fixture whatever the `select` asks for, so **no behavioural assertion of any kind could ever see that line vanish**, and the feature would have gone silently inert with no error, no log and no changed response. It is now asserted on the `select` argument itself, for all seven columns.

## Known limitation, not fixed here

The tensor caches (month TTL) are never invalidated when `updateFile` swaps a file's `url` in place. After a file replace, a stale analysis can be written against the fresh url. This is **pre-existing**; this change escalates it from stale panel display to a stale value written to the record. The fix belongs in `updateFile`, outside this lane — filed as [868kw8dre](https://app.clickup.com/t/868kw8dre).

## Deliberately not ported from #3113

- `weightPrecision`, a second precision string alongside `fp`.
- A `blocks.N` + `first.weight` + `last.linear.weight` signature for `DiffusionModel` — too thin to write to a creator's record on.
- `TENSOR_METADATA_SUMMARY_VERSION`, a second cache-key version, unnecessary once the derived field is optional.

Reviewed adversarially by two agents prompted to refute; both reports are addressed above.

Co-authored-by: rmatif <rmatif@users.noreply.github.com>

### DIFF
--- a/src/pages/api/v1/model-files/[id]/tensor-metadata.ts
+++ b/src/pages/api/v1/model-files/[id]/tensor-metadata.ts
@@ -6,9 +6,11 @@ import { dbRead } from '~/server/db/client';
 import { logToAxiom } from '~/server/logging/client';
 import { REDIS_KEYS } from '~/server/redis/client';
 import { getFileForModelVersion } from '~/server/services/file.service';
+import { correctModelFileFromTensorHeader } from '~/server/services/model-file-header-correction.service';
 import { getFullTensorAnalysisCached } from '~/server/services/tensor-metadata.service';
 import { fetchThroughCache } from '~/server/utils/cache-helpers';
 import { MixedAuthEndpoint } from '~/server/utils/endpoint-helpers';
+import type { ModelTensorAnalysis } from '~/utils/model-tensor-metadata';
 import {
   inferTensorMetadataFormat,
   parseModelTensorMetadata,
@@ -61,6 +63,7 @@ export default MixedAuthEndpoint(async function handler(
       id: true,
       modelVersionId: true,
       name: true,
+      url: true,
       type: true,
       sizeKB: true,
       metadata: true,
@@ -141,6 +144,33 @@ export default MixedAuthEndpoint(async function handler(
         )
       );
 
+    // Repair on demand: the header is the source of truth for precision, and for file
+    // type where it is unambiguous, so a viewed file with a wrong stored value is
+    // corrected here. Deliberately NOT awaited before responding — the read must not
+    // wait on a write, and a failed correction is retried by the next viewer.
+    const correct = (analysis: Pick<ModelTensorAnalysis, 'dtypeCounts' | 'detectedModelType'>) =>
+      correctModelFileFromTensorHeader({
+        fileId: file.id,
+        fileUrl: file.url,
+        modelVersionId: file.modelVersionId,
+        format,
+        dtypeCounts: analysis.dtypeCounts,
+        detectedModelType: analysis.detectedModelType,
+        currentFp: (file.metadata as BasicFileMetadata | null)?.fp,
+        currentFileType: file.type,
+        modelType: file.modelVersion.model.type,
+      }).catch((error) =>
+        logToAxiom({
+          name: 'model-file-tensor-metadata',
+          type: 'error',
+          message: `header correction failed: ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+          fileId: id,
+          modelVersionId: file.modelVersionId,
+        }).catch(() => undefined)
+      );
+
     if (summaryOnly) {
       const summary = await fetchThroughCache(
         `${REDIS_KEYS.CACHES.TENSOR_METADATA_SUMMARY}:${id}`,
@@ -152,12 +182,14 @@ export default MixedAuthEndpoint(async function handler(
         { ttl: CacheTTL.month }
       );
       res.setHeader('Cache-Control', TENSOR_METADATA_CACHE_CONTROL);
-      return res.status(200).json(summary);
+      res.status(200).json(summary);
+      return void correct(summary);
     }
 
     const analysis = await fetchFull();
     res.setHeader('Cache-Control', TENSOR_METADATA_CACHE_CONTROL);
     res.status(200).json(analysis);
+    return void correct(analysis);
   } catch (error) {
     // The upstream byte-range fetch fails transiently. If this 422 ever inherits the immutable
     // header, Cloudflare freezes the error for a year and the file's panel never recovers.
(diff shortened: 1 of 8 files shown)


JustMaier:
Full unit suite, as promised in the description. Run on `cd4c8932d1`:

```
Test Files  1376 passed | 1 skipped (1377)
Tests  21627 passed | 27 skipped (21654)
Duration  286.94s
```

exit code **0**.

Reconciling against the earlier run on this branch, since a total that does not add up means workers died and files never ran:

| | files | accounted |
|---|---|---|
| before (`d685ab52`) | `1 failed \| 1375 passed \| 1 skipped` | 1377 |
| now (`cd4c8932d1`) | `1376 passed \| 1 skipped` | 1377 |

Same 1377 files both times, so no file silently dropped out. The one failure was `no-direct-shared-module-mock` flagging this branch's own test for mocking `~/server/db/client` directly; fixed in `fc037b9a3d` by moving to the canonical `dbMock`. Test count rose 21601 → 21654, which is the 53 tests added by the review rounds.

`src/server/routers/__tests__/blocks.router.workflow.test.ts` collected **358** tests, so the `event-engine-common` submodule is present in this worktree and the run is not a silent zero-collection.

PR #4347, opened 2026-08-24
fix(model): don't edge-cache getAll for an authenticated caller

ZacxDev:
> **This description was rewritten on 2026-08-24.** The original version claimed this PR closed a live cross-user cache exposure. An audit refuted that and I verified the refutation independently — see the comment below for what was wrong and why. The change is now proposed as defence in depth, not as a fix for a live bug.

## What

`model.getAll` is the one `edgeCacheIt`-wrapped procedure whose response depends on the caller. This adds `!!ctx.user` to its existing upstream opt-out so that stays true regardless of how the surrounding code evolves, and adds a test that pins the two structural properties the current safety actually rests on.

**This changes no response header today.** See below.

## Why it is a no-op today

`responseMeta` (`src/pages/api/trpc/[trpc].ts`) is handed the **root** context — `ctxManager.valueOrUndefined()` — while tRPC's `next({ ctx })` builds a *new* object for downstream:

```js
ctx: nextOpts?.ctx ? { ...opts.ctx, ...nextOpts.ctx } : opts.ctx
```

`skipEdgeCache` passes `cache: { ...ctx.cache, skip }` — a **copy**. `edgeCacheIt` *mutates* `ctx.cache.edgeTTL` (`middleware.trpc.ts:299-301`), so its writes land on that copy and never reach the object the header is computed from. The header therefore already comes from `createContext`'s defaults, which are `0` for a session.

`model.getAll` is the **only** production router that replaces `ctx.cache` — `git grep 'cache: {' src/server/routers/*.ts` (excluding tests) returns exactly one hit, the middleware this PR edits. It is immune by accident, not by design. The anonymous `s-maxage=60` comes from `createContext`, and `edgeCacheIt`'s own value for this procedure is also 60 — the two coincide exactly, which is why this was not obvious.

## Why keep the change

The immunity is one refactor deep. Dropping that spread is an obvious tidy-up — copying an object only to hand it downstream looks pointless — and doing so hands `edgeCacheIt` the root object, at which point this procedure starts emitting `public, s-maxage=60` on responses computed for a specific caller. The body genuinely varies: `getModelsWithImagesAndModelVersions` branches on `user.isModerator` for status filtering and `model.user.id === user?.id` for owner-only fields, and the controller reads two per-user feature flags.

`!!ctx.user` makes the outcome correct either way, and costs nothing: anonymous callers have no `ctx.user`, so their caching is untouched.

## Tests

Two files. The first pins the middleware's decision; the second runs the **real chain**, replicating tRPC's merge, and asserts what `responseMeta` actually reads.

The second exists because the first could not see the properties that matter. A sentinel root TTL of `999` is what makes them observable — with realistic values, `createContext`'s anonymous default and `edgeCacheIt`'s own ttl are both 60, so a write and a no-write are indistinguishable.

Structural mutation battery, 10 assertions across both files:

| mutation | pass / fail |
|---|---|
| baseline | 10 / 0 |
| swap `.use(skipEdgeCache)` and `.use(edgeCacheIt(…))` | **5 / 5** |
| share `ctx.cache` instead of copying it | **5 / 5** |
| drop `!!ctx.user` | 7 / 3 |
| both — the realistic regression | 3 / 7 |
| restored | 10 / 0 |

The first two **survived** the isolation-only test at 5/5. That is the gap the chain test closes.

Neighbouring suites (`src/server/routers/__tests__/`, `middleware.trpc.cacheit`, `model.paged-simple-cacheability`, `trpc-batching`): **58 files / 1250 tests passed**.

## Third commit: it is no longer purely a no-op

`CACHEABLE_PROCEDURES` (`src/utils/trpc.ts`) exists so an authed browser does not batch a query the edge is caching for it — batching appends `?batch=1`, which `edgeCacheIt` refuses, so batching such a query trades an edge hit for api-pool load.

`model.getAll` was listed, and with `!!ctx.user` it no longer belongs: `edgeCacheIt` computes a TTL of 0 for a session, so there is no authed edge hit to protect and `shouldBatch` has been declining to batch it for nothing. Removing it means authed browsers can batch it once `trpcBatching` ramps.

The drift guard could not see this. It derives the set by scanning each procedure block for `edgeCacheIt(` minus `noEdgeCache({ authedOnly })` / `noEdgeCache()`; `skipEdgeCache` is a third spelling, expressed in a local middleware the block only references by name. The derivation now resolves `.use(localMiddleware)` against consts in the same file and counts a `skip` assignment consulting `ctx.user` as an authed opt-out.

Verified live in both directions, so the guard is not decoration:

- removing `model.getAll` from the set **without** the derivation change fails with `notListed: ['model.getAll']`;
- reverting `!!ctx.user` **with** both changes in place fails the same way.

Two unit tests cover the resolution directly — one where the local middleware consults `ctx.user` (opt-out visible only after resolution), one where it keys on input only (no opt-out invented).

The doc comment is also corrected. It asserted flatly that these responses "ARE cacheable for authed users today"; two exceptions are now recorded — the `ctx.cache`-copy case above, and that **`noEdgeCache()` is inert when combined with `edgeCacheIt` in either order** (it writes before `next()`, `edgeCacheIt` writes after). Nothing combines them today, so that clause of the guard has never been exercised and would silently not work.

## Unrelated flaky test

`src/server/__tests__/eventloop-watchdog.capture.test.ts` fails in the wider run. It is **flaky on `main`**, not caused by this branch: measured 1 pass / 2 fail on a clean `origin/main` worktree and 0/5 here, and it imports only `~/server/eventloop-watchdog` and `node:worker_threads`. Flagging rather than fixing.

## Follow-ups

1. `home-block.controller.ts:148` sets `ctx.cache.skip = true` from inside the resolver, where `edgeCacheIt` has already read it — so it never fires. Its service also carries a comment saying the `SessionUser` is "passed down to `getHomeBlockData`"; it is never bound.
2. Sibling `edgeCacheIt` procedures do not have this one's accidental immunity — `edgeCacheIt` does reach their headers for authenticated callers. That is currently safe only because none of them returns a caller-dependent body, which nothing enforces. A guard asserting that relationship is the durable fix.
3. `env`, `CacheTTL` and `applyUserPreferences` are imported but unused in `model.router.ts` (already true before this change).


ZacxDev:
## Correction: this PR's original description was wrong, and I've rewritten it

Posting rather than silently editing, since anyone who read the first version read a claim I now believe is false.

**What the original said:** that `model.getAll` emits `public, s-maxage=60` on authenticated callers' responses, and that this middleware was "the ONLY thing standing between the response and a shared edge entry". It also listed `homeBlock.getHomeBlock` and `user.getCreator` as unprotected.

**What is actually true:** `edgeCacheIt` has never reached the response header on this procedure.

`responseMeta` is handed the root context (`ctxManager.valueOrUndefined()`), and tRPC's `next({ ctx })` builds a new object rather than mutating the root:

```js
ctx: nextOpts?.ctx ? { ...opts.ctx, ...nextOpts.ctx } : opts.ctx
```

`skipEdgeCache` passes `cache: { ...ctx.cache, skip }` — a copy — and `edgeCacheIt` *mutates* `ctx.cache.edgeTTL`, so its writes land on the copy. The header comes from `createContext`'s defaults, which are already `0` for a session. `model.getAll` is the only production router that replaces `ctx.cache`, so it is the only one with this property, and it has it by accident.

The anonymous `s-maxage=60` comes from `createContext`, not `edgeCacheIt` — and `edgeCacheIt`'s value here is also 60, so the two coincide exactly. That coincidence is why the original reading looked right.

**On the two procedures I named:** the mechanism *is* live on them, in the sense that `edgeCacheIt` reaches their headers. But their bodies do not vary by caller — the identity passed in is discarded by the callee in both cases — so there is no cross-user exposure there either. Listing them as unprotected was misleading and I've removed that framing.

**Net:** I am not aware of a live cross-user cache exposure in this class. The change is now proposed as defence in depth — the immunity above is one refactor away from disappearing, since dropping that spread looks like an obvious cleanup — and the new chain test pins the properties it depends on. Two structural mutants (reordering the `.use()` calls, and removing the copy) survived the original isolation-only test at 5/5 and now fail.

Happy to close this instead if reviewers would rather not carry a no-op guard; the test is the part I'd argue for keeping either way.

PR #4352, opened 2026-08-24
fix(model-files): the type guard from #4344 was inverted, and two related hazards

JustMaier:
Follow-up to #4344 (merged as `1411b6ca33`). Two correctness bugs and one drift hazard, all found by the **three review lanes that had not run before #4344 merged** — `civitai-intent-review`, `civitai-perf-review` and `civitai-reuse-review`. I had run only correctness and tests. That gap is the story of this PR.

Every finding below was confirmed by **running the code**, not by reading it.

## 1. The type guard was inverted — it blocked real corrections and permitted known-bad ones

#4344 refused any type correction that would move a file out of `primaryFileTypesByModelType[modelType]`. Measured on `1411b6ca33`:

```
✕ a VAE uploaded as the Model file of a Checkpoint version
  AssertionError: expected {} to deeply equal { type: 'VAE' }

✕ an LLM file typed Other, where the code documents the detection as a misread
  AssertionError: expected { type: 'Text Encoder' } to deeply equal {}
```

- **`'VAE'` is not in `primaryFileTypesByModelType.Checkpoint`**, so a VAE wrongly uploaded as a checkpoint's main weights file — the classic mis-file this feature exists for — was detected correctly and then left alone.
- Meanwhile the guard did nothing for a file that was *not* already primary, so an LLM file typed `'Other'` was still relabelled `'Text Encoder'` by a rule the source itself documents as a misread of LLM weights. #4344 even had a test asserting that as desired.

**Root cause: I fixed a correctness-lane finding at the wrong layer** — I patched the consequence (the download-path demotion) instead of the ambiguity that caused it.

**Fix at the source.** On an `LLM` or `VisionLanguage` model the text/vision-encoder detections *are* the model's own weights; on a `MotionModule` so is the LoRA shape. The answer is `'Model'` — the same `modelType === X ? 'Model' : …` pattern the mapper already used for VAE, UNet and ControlNet. With the ambiguity resolved the downstream guard had nothing left to protect and only refused real corrections, so it is removed.

## 2. A type correction left a stale VRAM estimate cached for a month

`supportsTensorVramEstimate({ modelType, fileType: file.type })` gates `vramEstimate`, so **`ModelFile.type` is an input to the cached analysis** — but neither `TENSOR_METADATA` nor `TENSOR_METADATA_SUMMARY` varies with type, and nothing in the repo busts either. #4344 is what made `file.type` mutate on the read path, so a corrected file kept a `vramEstimate` that is now wrong for `CacheTTL.month` in redis, and up to a year at the CDN because the 200 carries `immutable`.

Now busted on a **type** change only (precision is not an input), reusing the existing `bustFullTensorAnalysis` for the in-process LRU. The full blob is brotli-compressed, so its bust passes `{ compress: true }` — without that flag the entry is evicted instead of marked stale and the bust silently no-ops.

## 3. The precision rule was a deny-list, and precisions are mod-managed at runtime

`constants.modelFileFp` is only a **fallback**; precisions live in the `modelFileOptions` KeyValue row and mods add them at runtime. A deny-list of "values the header cannot state" silently stops covering anything added after it is written — and since this runs on read, the first overwrite re-applies over any manual fix. That is the same unrecoverable failure the list was added to prevent.

Inverted to an allow-list of the four values a dtype can actually state (`fp32`, `fp16`, `bf16`, `fp8`), so an unknown precision is refused **by construction** rather than by remembering to edit an array. An absent `fp` is still always fillable.

## Verification

```
pnpm run typecheck                  OK — 0 type errors
pnpm exec eslint <changed files>    0 errors
affected suites                     Test Files 6 passed (6) / Tests 153 passed (153)
```

Five positive controls, each reverted, each red:

| reverted | failure |
|---|---|
| LLM/VLM own-weights fix | 3 red, `expected { type: 'Text Encoder' } to deeply equal {}` |
| MotionModule arm | `expected { type: 'Enhancement LoRA' } to deeply equal {}` |
| tensor-cache bust on type | `expected "vi.fn()" to be called with arguments: [ 42 ]` |
| `{ compress: true }` on the full-blob bust | `expected {} to deeply equal { compress: true }` |
| allow-list back to a deny-list | `expected { fp: 'bf16' } to deeply equal {}` |

## Cleared by the perf lane, on record

Steady state is genuinely free: once corrected, the path issues no query and no redis command. Replica lag measured at **3 ms**, and inside that window the `IS DISTINCT FROM` predicates make the UPDATE match 0 rows — no lock, no dead tuple, no bust. `detectModelTypeFromTensors` runs on a redis **miss only** (≈3.9 µs/tensor), never on the LRU or summary path. `url: true` adds no I/O — the heap tuple was already fetched.

Two things it logged rather than flagged, filed rather than fixed here: ~977k rows are one anonymous origin request from an UPDATE each, and this route has no rate limit; and every correction is a non-HOT update against 7 indexes (~390 MB heap churn across the whole wave, no TOAST — measured max `pg_column_size(metadata)` 450 B).

## Not fixed here

- `'Vision Encoder'` has no case in `inferComponentType` or `fileTypeToComponentTypeMap`, so a file corrected to it drops out of component grouping. Pre-existing gap that this feature makes reachable — reuse lane's finding, worth its own PR.
- Tensor caches are still not busted when `updateFile` replaces a file's `url` — [868kw8dre](https://app.clickup.com/t/868kw8dre).

### DIFF
--- a/src/server/services/__tests__/model-file-header-correction.service.test.ts
+++ b/src/server/services/__tests__/model-file-header-correction.service.test.ts
@@ -1,14 +1,27 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { dbMock } from '~/__tests__/mocks/db.mock';
 import type * as ModelFileService from '~/server/services/model-file.service';
+import type * as TensorMetadataService from '~/server/services/tensor-metadata.service';
+import type * as CacheHelpers from '~/server/utils/cache-helpers';
 
-const { deleteFilesForModelVersionCache } = vi.hoisted(() => ({
-  deleteFilesForModelVersionCache: vi.fn(),
-}));
+const { deleteFilesForModelVersionCache, bustFullTensorAnalysis, bustFetchThroughCache } =
+  vi.hoisted(() => ({
+    deleteFilesForModelVersionCache: vi.fn(),
+    bustFullTensorAnalysis: vi.fn(),
+    bustFetchThroughCache: vi.fn(),
+  }));
 
 vi.mock('~/server/services/model-file.service', async (importOriginal) => ({
   ...(await importOriginal<typeof ModelFileService>()),
   deleteFilesForModelVersionCache,
+}));
+vi.mock('~/server/services/tensor-metadata.service', async (importOriginal) => ({
+  ...(await importOriginal<typeof TensorMetadataService>()),
+  bustFullTensorAnalysis,
+}));
+vi.mock('~/server/utils/cache-helpers', async (importOriginal) => ({
+  ...(await importOriginal<typeof CacheHelpers>()),
+  bustFetchThroughCache,
 }));
 
 const executeRaw = dbMock.dbWrite.$executeRaw;
@@ -42,6 +55,7 @@ beforeEach(() => {
   vi.clearAllMocks();
   executeRaw.mockResolvedValue(1);
   deleteFilesForModelVersionCache.mockResolvedValue(undefined);
+  bustFetchThroughCache.mockResolvedValue(undefined);
 });
 
 describe('getModelFileHeaderCorrections — precision', () => {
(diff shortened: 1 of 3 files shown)


JustMaier:
🛑 **Converted to draft — do not merge as it stands.** A review pass over this fix round found that it removes the wrong guard.

## `ModelFile.type` is a safety control, not a display label

Verified in the current code, not taken from the report:

- **`model-version.service.ts:2860`** — `addLinkedComponent` with a `replaceFileId` deletes the `ModelFile` row and GCs the S3 object. The entire "never delete someone's weights" control is:
  ```ts
  if (primaryModelFileTypes.includes(replaceFile.type as ModelFileType) || replaceFile.type === 'Training Data')
    throw new TRPCError({ message: 'Cannot replace a primary model or training-data file' });
  ```
  `model-file-scan.service.ts` has an automated caller that skips a file *because* it is primary-typed and otherwise proceeds.
- **`minor-hash.service.ts:20`** — `MINOR_HASH_FILE_TYPE = 'Model'`, joined as `mf.type = 'Model'` at lines 82, 97, 115 and 364: both the seed set and the candidate set. A file relabelled off `'Model'` leaves that population.

## Why this PR specifically

It removes `wouldDemoteFromPrimary`. On merged `main` that guard is what keeps the blast radius to files that were not primary anyway. Without it, a tensor-name heuristic running on an **anonymous public GET** can move a creator's primary weights file off `'Model'` — disarming both controls above — and because the correction re-applies on read, the creator cannot fix it by hand.

The removal was correct *as a response to the inverted-guard bug*; it is wrong given what the column actually gates. Both things are true.

## Also found, and not covered by the source fix here

A Flux-era all-in-one checkpoint that bundles its text encoders (`model.diffusion_model.*` + `vae.*` + `text_encoders.clip_l.*`) matches **neither** the Checkpoint rule (needs `first_stage_model.`/`cond_stage_model.`/`conditioner.embedders.`) nor the VAE rule (`^encoder\.` anchor rejects `vae.encoder…`), and falls through to the text-encoder rule. Detection returns `TextEncoder` for the whole checkpoint — and one of the new tests in this PR asserts that outcome as intended. No fixture covers that layout, which is why the suite is green.

Two further pairs the source fix misses by one enum value each: `VisionEncoder` on a model typed `LLM` rather than `VisionLanguage`, and `LoRA` on `Controlnet` (a control-lora — the LoRA rule runs before the ControlNet prefix rule).

## Status

Three fix rounds on the type half have each introduced a new defect, every one found by an adversarial pass rather than by the suite. The precision half has produced one, caught immediately.

I have recommended shipping **precision only** and refiling the type correction as its own work starting from a real-file corpus rather than hand-written fixtures. That is a scope reduction on the ticket, so it is the owner's call — this PR stays in draft until it is answered.

The precision changes in this PR (the allow-list inversion, commit `08bdb002f2`) are sound and independent of all of the above; they will be preserved whichever way this goes.

JustMaier:
**Full unit suite on `08bdb002f2`: exit 1 — and the failures are not from this branch.**

```
Test Files  2 failed | 1381 passed | 1 skipped (1384)
Tests  7 failed | 21751 passed | 27 skipped (21785)
```

1384 files accounted for, so no worker died. The two failing files are:

- `src/server/services/__tests__/home-block-fetch-sizing.test.ts`
- `src/server/services/__tests__/user-payment-configuration.tipalti-status.test.ts`

Neither is touched by this branch — the diff is tensor metadata, file helpers and a new correction service.

**Controls, because a red I want to blame on the environment needs more than an assertion that it is not mine:**

| check | result |
|---|---|
| `home-block-fetch-sizing` alone on this branch | **7 passed (7)** |
| the two files together, alone, on this branch | **fail** — `Test timed out in 60000ms` and `ReferenceError: Cannot access '__vite_ssr_import_7__' before initialization` |
| same file in `t43l23z` (worktree `bdx-cursor-fallback`) | fails |
| same two files in `t3xsc3a` (worktree `comment-blocklist`) | fail |

Three unrelated branches, three different agents, the same files. The pair reproduces the failure with **none of this branch's code loaded**, and `Cannot access '__vite_ssr_import_7__' before initialization` is a module-initialisation/circular-import symptom, not an assertion failure.

So this is cross-file pollution on `main`, not a regression here. Flagged to the coordinator as fleet-wide; not fixed here because it is outside this lane's file set and wants someone who can run the suite on bare `origin/main` to find the pollution source.

**Correction worth recording:** I first attributed this to my own concurrent vitest runs in the same worktree, which is a documented false-red cause on this box. The cross-branch evidence contradicts that. Noting the wrong hypothesis as well as the right one, because it is the plausible-sounding one the next reader will also reach for.

The affected suites for this change are green: `Test Files 6 passed (6) / Tests 153 passed (153)`, typecheck 0 errors, eslint 0 errors, and five positive controls all red.

JustMaier:
Closing with #4344. Justin's call: upload already detects precision, this was retroactive repair, and the type half is not worth the surface it turned out to carry.

Revert of #4344 is at #4359. Nothing to unwind — `git tag --contains 1411b6ca33` is empty and the latest release `v5.1.39` does not contain it, so the correction never deployed and no `ModelFile` row was written.

The findings that killed it are recorded in #4359's description rather than lost with the branch, in particular why `ModelFile.type` cannot be moved by a heuristic (it gates `addLinkedComponent`'s file-delete refusal and the minor-hash matching population) and why a precision rule must be an allow-list rather than a deny-list (precisions are mod-managed at runtime, so a deny-list silently stops covering what a mod adds later).

PR #4358, opened 2026-08-25
feat(text-moderation): measure a text policy per surface, and let one surface run it

manuelurenah:
Two halves of one problem. The XGuard label lab could not correctly measure a policy for anything
scanned in **text** mode — model listings, articles — and a policy measured there had no route to
production that did not move every other text surface with it.

**What's in here**

| | |
|---|---|
| Lab — three bug fixes | SSL decided from the hostname rather than the URL's `sslmode`; every evaluation scored against the **prompt** scanner regardless of what the content passes through; live baselines defaulting their threshold instead of stating it |
| Lab — generalised past models | a sampler and a review path for any text-mode entity, plus rubrics split into a judgement and the domain it is read against |
| Production — one commit | `submitTextModeration` forwards `labelOverrides`, so one surface can scan under a policy tuned for it |

**The lab half leaves production alone.** The live XGuard registry is byte-identical before and
after: candidate policies are evaluated as per-request `labelOverrides`, which never write to the
orchestrator. The last commit is the exception — it touches `submitTextModeration` and the shared
submit helper, it is described under **Reaching production** below, and it is the part of this PR
to review as production code.

Measurements, corpora and the per-label results live in [CU 868ktb1wb](https://app.clickup.com/t/868ktb1wb). They are deliberately not repeated here — this repo is public, and per-label operating points are content-safety internals.

## Three bugs, each of which returned a confident wrong number

**The lab could not open at all through a tunnel.** `/xguard` 500'd with `SELF_SIGNED_CERT_IN_CHAIN` for anyone reaching the lab database through a local port-forward — the documented dev path for the cluster instance. SSL was decided from `isLocalHost(url)`, but a tunnel and a local docker Postgres are both `localhost` and need opposite answers. The connection string already declared its sslmode; the code was not reading it. `labelConnectionString()` had the same bug by another route, so the policy editor's Evaluate action failed the same way. Note `moderator-db.ts` already hardcodes `sslNoVerify: true` against the same database — two files pointing at one instance disagreed about its SSL and only one was right.

**Every evaluation measured the wrong scanner.** `scan()` hardcoded `mode: 'prompt'`, so a batch of model listings was scored against the prompt registry. Prompt and text are separate registries with separate label sets and thresholds, so the wrong mode does not fail — it returns a confident number for a scanner the content never passes through. Mode is now derived from `sample.source`; an unknown source throws, a mixed batch throws, and the resolved mode is stamped on the run note because `eval_run` has no column for it and a run whose mode cannot be recovered cannot be compared against another.

**Live baselines defaulted their threshold.** A baseline exists to say what production does today, and production's threshold lives in the orchestrator registry — a default silently measures a different operating point and records it as "live". It now requires an explicit threshold.

## Preparing corpora without hand-seeding

The lab could only sample generation prompts from ClickHouse, so preparing a batch for anything scanned in text mode meant seeding rows by hand.

- **`sample-entities.ts`** stratifies an entity's content across a label's score bands using scores `EntityModeration.result` already holds, so a batch costs no model calls. Stratifying matters because a label's scores are rarely spread evenly across live content and different labels skew in opposite directions — a uniform sample of a skewed label says nothing about where its boundary sits. Listings with little or no description are kept as a bounded minority rather than excluded, since dropping them would tune a label against a population it does not meet in production.
- **`seed-from-predictions.ts`** feeds the rows a policy fired on back into the review queue. Precision is TP/(TP+FP) and every term is a fired row, so when positives are a small minority, reviewing those measures it exactly while a random sample spends most of the effort on rows that cannot change the answer. Documented as unusable for recall, since the rows it drops are the false negatives.

Everything entity-shaped is in `ENTITY_TEXT`, which maps an entity to the table and columns its scanned string is built from. Each entry must match that entity's `resolveContent` in its production adapter, or reviewers judge text the scanner never sees. Challenge and WildcardSetCategory are absent deliberately — neither is a title+body pair, and inventing a resolver that merely looks right is the drift that comment warns about.

## A label is a judgement; a domain is what it is read against

Every rubric restated what kind of text it judges, so a new surface meant another copy of that paragraph in every label.

`DOMAINS` now holds what the text is, `rubric` holds the judgement, `domainNotes` holds a convention that applies to one surface only, and `rubricFor()` composes them. Adding a surface is one `DOMAINS` entry rather than an edit to every rubric. The rater no longer carries a hardcoded idea of what it is reading — the domain comes from `sample.source`, for the same reason the scanner mode does.

`domains` records where a label is meaningful, and asking for a pair that is not listed throws rather than quietly rating against the wrong preamble. The listing labels claim `modelListing` only: their bodies still name conventions specific to that surface, so claiming prompt reuse would assert something that has not been measured.

## Reaching production

A tuned text policy had one route to production: writing the **global** text registry, which
Model, Article, Challenge and WildcardSetCategory all read. Raising a threshold measured on one
surface moved it for every other, none of which it was measured against. That was the gap between
measuring a better policy and content being scanned by it.

`createXGuardModerationRequest` already accepted and forwarded `labelOverrides` for both modes.
`submitTextModeration` — the entity-facing wrapper every adapter goes through — dropped them, so
no adapter could reach the capability. It now takes and forwards them, and the override shape is
an exported type rather than a third inline copy.

**Forwarding alone would have shipped a footgun**, so the overrides are now part of the dedup key.
The dedup answers a request from an earlier workflow whenever the content hash matches, and that
workflow was scored under whatever policy was in force then — so the first request under a new
policy would have been answered with the old policy's verdict, for content never scanned under the
new one. Same failure this branch has been fixing elsewhere: a confident number for a scan that did
not run.

They are canonicalised rather than serialised as given, because a per-surface policy is meant to be
the steady state — an equivalent policy in a different order has to keep deduping, or every save of
an unchanged entity burns a scan. The suffix is empty when there are no overrides, so every hash
written so far still matches and **this change rescans nothing**.

No caller passes overrides yet, so behaviour is unchanged until one does. The ClickHouse audit key
is computed separately, from the workflow step input, and is untouched.

## Verification

`pnpm run typecheck:scripts` clean · `pnpm test` 233 passed. Typecheck and lint are left to CI.

`typecheck` stops at `src/`, so `xguard-lab/` is invisible to it — `typecheck:scripts` is what covers the files most of this PR touches. Separately, all nine label/domain pairs were checked to compose with the preamble appearing once and the rubric bodies intact, since a typecheck cannot see whether a composed prompt reads correctly. The sampler was exercised against both Model and Article rather than only compiled.

For the production commit: full unit suite green — 1,394 files, 21,819 passed, 0 failed. Both new
suites were checked against a revert rather than only against green, since a passing test says
nothing about how it fails. Undoing the forward fails the two forwarding assertions; undoing the
dedup-key change fails the two saying an override request must not be answered from a row scanned
without one. The other two cases pass either way by design — they are the no-regression guards
proving content-only dedup still behaves exactly as before.

## Not in this PR

- `rate-core` handles rater refusals but not unparseable JSON.
- Splitting the listing rubrics into a neutral core plus `domainNotes` is what would make them reusable across surfaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,6 +123,8 @@ docs/working/
 docs/incidents/
 .claude/*.local.*
 .claude/skills/cloudflare/.port-tmp-path
+# Also holds the XGuard tuning harness: scanner policy prose and corpora extracted
+# from prod. This repo is public; a policy file here would be an evasion guide.
 local/
 .eslintcache
 nul
--- a/apps/moderator/src/lib/server/xguard-api.ts
+++ b/apps/moderator/src/lib/server/xguard-api.ts
@@ -8,6 +8,22 @@ import { env } from '$env/dynamic/private';
 export function labConnectionString(): string {
   const url = env.MODERATOR_DATABASE_URL;
   if (!url) error(500, 'MODERATOR_DATABASE_URL is not configured');
+
+  // Consumers of this string build a bare `pg.Client`, which verifies the certificate chain. The
+  // cnpg pooler presents a self-signed one, so `sslmode=require` fails there with
+  // SELF_SIGNED_CERT_IN_CHAIN — encrypted-but-unverified is what this host needs. Rewriting it here
+  // rather than at each call site because eval-core and rate-core each open their own connection.
+  // An explicit `sslmode=disable` (a plain local docker Postgres) is left alone.
+  try {
+    const parsed = new URL(url);
+    const sslmode = parsed.searchParams.get('sslmode');
+    if (sslmode && sslmode !== 'disable' && sslmode !== 'no-verify') {
+      parsed.searchParams.set('sslmode', 'no-verify');
+      return parsed.toString();
+    }
+  } catch {
+    // Not a parseable URL — hand it back untouched and let pg report the real problem.
+  }
   return url;
 }
 
--- a/apps/moderator/src/lib/server/xguard-lab.ts
+++ b/apps/moderator/src/lib/server/xguard-lab.ts
@@ -114,11 +114,19 @@ function labUrl(): string {
 
 // `sslNoVerify` turns SSL ON while skipping cert verification, which is what the cnpg cluster
 // needs and what a plain local docker Postgres rejects outright ("The server does not support SSL
-// connections"). Decide from the host so the same code works in both places.
-function isLocalHost(url: string): boolean {
+// connections").
+//
+// An explicit `sslmode` in the URL decides it; the hostname is only consulted when the URL says
+// nothing. Reading the hostname first is wrong because the cluster is reached in dev through a
+// local port-forward, so a cnpg tunnel and the docker-compose Postgres are both `localhost` and
+// the two need opposite answers.
+function wantsSsl(url: string): boolean {
   try {
-    const { hostname } = new URL(url);
-    return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1';
+    const parsed = new URL(url);
+    const sslmode = parsed.searchParams.get('sslmode');
+    if (sslmode) return sslmode !== 'disable';
+    const { hostname } = parsed;
+    return !(hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '::1');
   } catch {
     return false;
   }
@@ -134,7 +142,7 @@ function createLabDb(): Kysely<LabDB> {
   const { dbWrite } = createKyselyClients<LabDB>({
     connectionString: url,
     replicaConnectionString: url,
-    sslNoVerify: !isLocalHost(url),
+    sslNoVerify: wantsSsl(url),
   });
   return dbWrite;
 }
--- a/apps/moderator/src/routes/api/xguard/runs/+server.ts
+++ b/apps/moderator/src/routes/api/xguard/runs/+server.ts
@@ -82,6 +82,12 @@ export const POST = defineWebhookEndpoint({
       .max(1)
       .optional()
       .describe("Override the version's stored threshold."),
+    mode: z
+      .enum(['prompt', 'text'])
+      .optional()
+      .describe(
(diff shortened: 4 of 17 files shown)


manuelurenah:
### Generalised past models

Review feedback: the lab is used for text-mode scanning generally, not just models, and the changes were too model-shaped. Checked which entity types actually flow through `EntityModeration` — Model is not the largest of them, so a sampler that only understands models leaves the loop unrunnable for the surfaces with the most material.

- `sample-models.ts` → **`sample-entities.ts`**, takes `--entity`. Everything entity-shaped is in `ENTITY_TEXT` (table + title/body columns), each entry required to match that entity's `resolveContent` in its production adapter.
- Challenge and WildcardSetCategory deliberately **not** added: neither is a title+body pair, and a resolver that merely looks right is the drift the file warns about.
- `MODE_BY_SOURCE` and `DOMAIN_BY_SOURCE` cover every text-mode entity; `DOMAINS` gains `article`. No label claims that domain yet, and `rubricFor` throws for any unlisted pair — an article batch cannot be rated against a rubric written for model listings.

The moderator-flag extractor that was here earlier has been **removed**. It was the only genuinely model-shaped file, and its ground truth turned out to be a record of moderator activity rather than of content, which makes it unable to measure precision. Keeping a tool whose target is known-flawed invites the next reader to trust a number from it.

**Verified against the database, not the compiler.** Sampling an entity for a label it is not scanned for returns nothing rather than something wrong; sampling for a label it does carry fills every band. Failing empty in that case is the behaviour worth having. Smoke batches were removed from the lab afterwards.

`typecheck` 0 errors / 9,344 files · `typecheck:scripts` clean · `pnpm test` 233 passed.

---

*Edited: the original version of this comment included per-entity scan volumes and label coverage. Those are content-safety internals and belong in [CU 868ktb1wb](https://app.clickup.com/t/868ktb1wb), not a public repo.*

PR #4368, opened 2026-08-25
fix(trpc): make a resolver-set edge-cache opt-out actually take effect

ZacxDev:
## The defect

`edgeCacheIt` (`src/server/middleware.trpc.ts`) computes the request TTL from
`ctx.cache.skip` **before** it calls `await next()`:

```ts
let reqTTL = ctx.cache.skip ? 0 : (ttl as number);
if (expireAt) reqTTL = Math.floor((expireAt().getTime() - Date.now()) / 1000);

const result = await next();          // <- the resolver runs HERE
if (result.ok && ctx.cache?.canCache) {
  ctx.cache.edgeTTL = reqTTL;
  ...
}
```

So an opt-out set *inside* a resolver is read before the resolver runs, and never
takes effect. `home-block.controller.ts` does exactly that:

```ts
if (homeBlock?.type === 'Announcement') ctx.cache.skip = true;
```

`homeBlock.getHomeBlock` is `.use(edgeCacheIt({ ttl: 60 }))`, so Announcement blocks
were cached for 60s against an explicit intent not to. For an **authenticated**
caller it was worse than a no-op: the context default is `0`, and the middleware
overwrote it with `60` — i.e. the guard's presence made those responses *more*
cacheable than the default, not less.

`canCache` does not have this problem — it is read *after* `next()` — which is why the
same in-resolver pattern works in `model.controller.ts`.

> ⚠️ **Read the live-impact correction in the comment thread before merging.** The
> `HomeBlockType.Announcement` render branch is commented out in `src/pages/home/index.tsx`,
> so the first-party app never calls this procedure with an Announcement block. This is
> defence in depth on a dormant path, not a measurable behaviour change.

## The change

`src/server/middleware.trpc.ts` — re-read `skip` once the resolver has returned, on
the same footing as `canCache`:

```ts
const result = await next();
if (ctx.cache?.skip) reqTTL = 0;
if (result.ok && ctx.cache?.canCache) { ... }
```

Idempotent for the existing upstream caller: `model.getAll` sets `skip` in a
middleware ahead of `edgeCacheIt`, so the value re-read is the value it set.

### `skip` now beats `expireAt` — deliberate, and now pinned

The new line sits *after* the `if (expireAt)` assignment, so it overrides a scheduled
expiry rather than being overridden by it. Before this PR the precedence ran the other
way. **That ordering is intended, not incidental:** `expireAt` says "this content goes
stale at time T", which is a statement about content that *is* cacheable; `skip` says
"this response must not be cached at all". A response the resolver has declared
uncacheable does not become cacheable because someone also scheduled its expiry — the
stronger claim wins.

It was accidental until now, so it is written down in the middleware and pinned by a
test. `expireAt` has **zero production call sites** today (only its type declaration and
its one use inside `edgeCacheIt`), so there is no real procedure combining it with a
resolver-set `skip`; the dimension is exercised on a purpose-built router driven through
the real tRPC caller (`src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts`),
with a control arm proving `expireAt` is genuinely in play.

### `willEdgeCache` is now single-sourced

`responseMeta`'s edge-cache predicate was an inline expression in
`src/pages/api/trpc/[trpc].ts`, which meant the chain test had to **re-declare** it — a
guard that keeps agreeing with its own copy while the real one changes underneath it. It
now lives in `src/server/trpc/edge-cache-headers.ts` and both the route and the tests
import it. `responseMeta`'s behaviour is unchanged; the local is narrowed off the cache
object instead of off `ctx` so the type predicate keeps the fields the header build reads.

### Comments corrected

- **`home-block.service.ts` / `getHomeBlockById`** — now leads with the reason that
  actually binds: the by-id result is stored in a **shared Redis entry keyed on
  type/identifier/domain with no user segment**
  (`src/server/services/home-block-cache.service.ts`), so reading `user` would serve one
  caller's body to another **even with `edgeCacheIt` removed entirely**. Edge caching is
  kept as the second reason. (This is the amendment offered in the correction comment.)
- **`home-block.service.ts` / `getHomeBlockData`** — the mirror-image stale comment, also
  flagged in that comment thread. Unlike the one above, this function *does* read `user`
  when given one — but `getHomeBlockCached` calls it **without** one, so collection items
  already resolve as anonymous on the by-id path. The comment now says so, and says
  explicitly that passing a user through that caller would poison a shared cache entry.
  **No behaviour change: the fix is the comment, not the argument.**
- **`system.router.ts`** — `getBenignPhrases` carried a comment stating `ctx.cache.skip`
  "does NOT work from here". This PR makes that false, so it is updated.

### `home-block.controller.ts` — null guard

`ctx.cache` really is `null` on the SSG prefetch path (`createServerSideHelpers` builds it
as `cache: null as any`, which is why `strict: true` does not flag the unguarded write).
Only the *list* procedure is prefetched today, so the line is not reachable from there —
the guard is precautionary, added because this PR makes that line load-bearing. The
fixture/branch also moves from the string literal to `HomeBlockType.Announcement`.

Sibling writers are **inconsistent** about this guard and the comment says so rather than
claiming a convention: `system.router.getBenignPhrases` and
`model.controller.getModelsPagedSimpleHandler` guard; `model.controller.getModelsInfiniteHandler`
does not. That third site is left alone here — it is a hot unrelated handler and not this
PR's business.

## Behaviour change worth naming: `Set-Cookie`

`responseMeta` calls `ctx.res?.removeHeader('Set-Cookie')` **inside** the
`willEdgeCache` branch. With `edgeTTL` now `0` for an opted-out response, that branch is
skipped, so an Announcement response **can now carry `Set-Cookie` where it previously
could not**. This is the safer direction — the header was only being stripped because the
response was about to be shared, and it is no longer being shared — but it is a real
change in emitted headers and nobody had written it down.

## Blast radius

`git grep -n 'cache\.skip'` plus a wider sweep for `cache: { ...ctx.cache` across
`src/`, non-test hits only:

| Site | Role |
|---|---|
| `createContext.ts` (×3) | type + the `skip: false` defaults |
| `middleware.trpc.ts` | the pre-`next()` read, now paired with a post-`next()` read |
| `home-block.controller.ts` | the only resolver that writes it — the inert one |
| `model.router.ts` | writes it **upstream**, on a spread **copy** (`cache: { ...ctx.cache, skip }`), so the root object is untouched |
| `system.router.ts` | comment only |

No third writer.

## Tests

Two files:

- `src/server/routers/__tests__/home-block.router.edge-cache-chain.test.ts` — the real
  `homeBlock.getHomeBlock` chain, including the real resolver (only the one service call
  it makes is stubbed), asserting the values `responseMeta` reads off the root context.
- `src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts` — the
  `skip` vs `expireAt` precedence, on a synthetic router.

**Both now drive `createCallerFactory`, not a hand-rolled runner.** The first revision of
the chain test walked `_def.middlewares` and re-implemented tRPC's context merge. That
reconstruction was verified faithful against `@trpc/server` 11.17.0 and was still wrong in
two ways: it forwarded the **raw** input rather than zod's parsed output, and it had **no
per-middleware try/catch**, so it could not produce the `{ ok: false }` result the real
runner produces. Driving the library's own caller removes both, and means a future tRPC
upgrade cannot silently invalidate the test by changing a runner the file was pretending
to be. The root context stays observable because tRPC's per-middleware `{ ...ctx, ...next.ctx }`
merge is shallow — `cache` is the same object in every copy — which is exactly the property
the test is asserting.

Sentinel TTLs (111 / 222 / 333) rather than realistic ones: the anonymous context default
and this procedure's own `ttl` are both `60`, so with production values "wrote the TTL" and
"left the default alone" are byte-identical and the assertion would be vacuous. Tests that
need the production defaults ask for them explicitly.

Red at `origin/main` (`middleware.trpc.ts` and `home-block.controller.ts` restored to
`origin/main` content, tree otherwise unchanged) → **2 files failed, 5 failed | 7 passed (12)**.
Green at HEAD → **2 files passed, 12 passed (12)**.

The 7 that pass in both columns are **invariant guards, not regression coverage**, and are
labelled as such: positive controls proving `edgeCacheIt` is live and reached (a
non-Announcement block still gets `60`, anonymous and authenticated), one discriminating
this fix from a `canCache: false`-shaped one (`staleWhileRevalidate` is still written, so
the assignment block runs and only the TTLs are zero), a reachability control (an earlier
middleware can still stop the chain, and the sentinel then survives untouched), the
`expireAt` control arm, the enum pin, and one asserting the opt-out is not an error path.

### Mutation battery

Both files run for every mutant. Counts are Test Files / Tests.

| Mutant | Files | Tests | Verdict | Died on |
|---|---|---|---|---|
| M0 baseline (HEAD) | 2 passed | **12 passed** | GREEN | — |
| M1 guard deleted (= `origin/main`) | 2 failed | 5 failed / 7 passed | KILLED | `expected 60 to be +0`, `expected 900 to be +0` |
| M2 guard moved above `await next()` | 2 failed | 4 failed / 8 passed | KILLED | `expected 60 to be +0`, `expected 900 to be +0` |
| M3 guard assigns `5` instead of `0` | 2 failed | 5 failed / 7 passed | KILLED | `expected 5 to be +0` |
| M4 `skip && !expireAt` (old precedence) | 1 failed / 1 passed | 2 failed / 10 passed | KILLED | `expected 900 to be +0` |
| M5 `willEdgeCache`: `> 0` → `>= 0` | 2 passed | 12 passed | **SURVIVED — equivalent** | — |
| M6 `willEdgeCache`: drop `!!`, `>= 0` | 2 failed | 4 failed / 8 passed | KILLED | `expected true to be false` |
| CTRL unrelated code (must survive) | 2 passed | 12 passed | SURVIVED ✅ | — |

M2 is the one that matters for the "verified in isolation" failure mode: a test of
`edgeCacheIt` alone, or of the controller alone, stays green under it, because neither
ever builds the combined state. M4 is the mutant that **survived the previous revision of
this PR** — the precedence was unpinned. M3's distinct message is the battery's own
control that each mutant genuinely ran, and CTRL confirms the battery is not simply
failing everything.

**M5 is an equivalent mutant and is reported as such rather than as a gap.** The predicate
is `!!cache && !!cache.edgeTTL && cache.edgeTTL > 0`; the `!!cache.edgeTTL` short-circuits
on `0` first, so relaxing the comparison alone cannot change any outcome and *no* test can
kill it. M6 — dropping the `!!` as well, which is what "widen this predicate to treat 0 as
cacheable" actually looks like — is killable and is killed, by the imported-predicate
assertions specifically (`expected true to be false`). That pair is what demonstrates the
test is bound to the real `willEdgeCache` rather than to a copy. Both halves of the
predicate are load-bearing and the module says so.

### One thing the new reachability control found

The previous revision set `features: { alternateHome: true }` on the test context and
documented it as what satisfies `isFlagProtected('alternateHome')`. Measured, that was
wrong in both directions: `isFlagProtected` resolves through `getFeatureFlags(ctx)`, which
computes from the session and never reads a `ctx.features` property — so the key was
**inert**, and setting it to `false` does not produce a `FORBIDDEN`. What `ctx.features` is
actually needed for is `applyDomainFeature`, which dereferences `ctx.features.canViewNsfw`
unguarded ahead of `edgeCacheIt`. The context now carries `canViewNsfw` and the comment
says why.

### Suites run

- `src/server/routers/__tests__/` + `src/server/__tests__/` — **114 files / 2198 tests, all passed**
- the neighbouring `home-block-*` service, `HomeBlocks` component, `middleware.trpc` and
  `trpc-batching` suites — **9 files / 91 tests, all passed**
- `pnpm typecheck` — **0 errors** (validated: a planted type error in the new module is
  detected, 1 error / rc 2)
- `scripts/ci/typecheck-tests-gate.mjs` — reports **BLOCKED**, and that is **pre-existing**:
  run against unmodified `origin/main` sources in the same worktree the output is
  byte-identical except the program size (**1352 → 1354 test files**), i.e. the gate does see
  both new files and they contribute **0** errors.
- `eslint` on the changed files — **0 errors** (5 pre-existing warnings, none on changed
  lines; validated by planting two rule violations and confirming detection)

## Note for #4347

`model.router.edge-cache-skip.test.ts` on that branch cites `home-block.controller.ts` as
*"a live instance of that mistake"*. Once this merges that sentence is stale — worth a
one-line touch-up on whichever lands second. The two PRs touch disjoint files (#4347 does
not modify `middleware.trpc.ts`), so there is no textual conflict.

### DIFF
--- a/src/pages/api/trpc/[trpc].ts
+++ b/src/pages/api/trpc/[trpc].ts
@@ -14,6 +14,7 @@ import { isClientAbortError } from '~/server/utils/errorHandling';
 import { appRouter } from '~/server/routers';
 import { runWithSerializeCtx, serializeCtxFromRequest } from '~/server/logging/trpc-serialize-log';
 import { getTrpcMaxBatchSize } from '~/server/trpc/batch-cap';
+import { willEdgeCache } from '~/server/trpc/edge-cache-headers';
 
 export const config = {
   api: {
@@ -51,16 +52,18 @@ const trpcHandler = createNextApiHandler({
   maxBatchSize: getTrpcMaxBatchSize(),
   responseMeta: ({ ctx, type, errors }) => {
     const headers: Record<string, string> = {};
-    const willEdgeCache = ctx?.cache && !!ctx?.cache.edgeTTL && ctx?.cache.edgeTTL > 0;
-    if (willEdgeCache && type === 'query' && errors.length === 0) {
-      ctx.res?.removeHeader('Set-Cookie');
+    // Single-sourced in `~/server/trpc/edge-cache-headers` so that tests asserting a
+    // middleware chain's cache outcome import THIS predicate instead of re-declaring it.
+    const cache = ctx?.cache;
+    if (willEdgeCache(cache) && type === 'query' && errors.length === 0) {
+      ctx?.res?.removeHeader('Set-Cookie');
       headers['Cache-Control'] = [
         'public',
-        `max-age=${ctx.cache.browserTTL ?? 0}`,
-        `s-maxage=${ctx.cache.edgeTTL ?? 0}`,
-        `stale-while-revalidate=${ctx.cache.staleWhileRevalidate ?? 0}`,
+        `max-age=${cache.browserTTL ?? 0}`,
+        `s-maxage=${cache.edgeTTL ?? 0}`,
+        `stale-while-revalidate=${cache.staleWhileRevalidate ?? 0}`,
       ].join(', ');
-      if (ctx.cache.tags) headers['Cache-Tag'] = ctx.cache.tags.join(', ');
+      if (cache.tags) headers['Cache-Tag'] = cache.tags.join(', ');
     }
 
     return Object.keys(headers).length > 0 ? { headers } : {};
--- a/src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts
+++ b/src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts
@@ -87,9 +87,17 @@ function rootCtx() {
     acceptableOrigin: true,
     tokenScope: TokenScope.Full,
     cache: { ...SENTINEL, canCache: true, skip: false },
-    // `applyDomainFeature` (src/server/trpc.ts) dereferences this unguarded, ahead of
-    // `edgeCacheIt`, on every `publicProcedure`. Without it the call dies with a
-    // TypeError before the middleware under test runs.
+    // Present for context realism only — MEASURED INERT for this file: delete this key
+    // and all 3 tests here still pass. `applyDomainFeature` (src/server/trpc.ts) does
+    // read `ctx.features.canViewNsfw`, but only on the authenticated arm: with `ctx.req`
+    // undefined, `parseVerifiedBotHeader(undefined)` returns null and the `&&`
+    // short-circuits before the property is reached, and the anonymous branch of
+    // `maxAllowed` never reads it either. Every case in this file is anonymous.
+    //
+    // Said plainly because the previous version of this comment claimed omitting the key
+    // "dies with a TypeError before the middleware under test runs" — that is false here,
+    // and it is the same defect (a test-context key documented as load-bearing that the
+    // code path never touches) that this PR exists to fix.
     features: { canViewNsfw: false },
   };
 }
--- a/src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts
+++ b/src/server/__tests__/middleware.trpc.edge-cache-precedence.test.ts
@@ -1,0 +1,135 @@
+import { describe, expect, it, vi } from 'vitest';
+
+// `edgeCacheIt` returns early when `!isProd`, so the TTL-assigning block only runs with
+// this mocked true. Same shape as `middleware.trpc.test.ts`.
+vi.mock('~/env/other', () => ({
+  isDev: false,
+  isProd: true,
+  isTest: false,
+  isPreview: false,
+}));
+
+vi.mock('~/env/client', () => ({
+  env: {
+    NEXT_PUBLIC_BASE_URL: 'http://localhost:3000',
+    NEXT_PUBLIC_CIVITAI_LINK: 'http://localhost:3000',
+  },
+  formatErrors: () => [],
+}));
+
+import type { Context } from '~/server/createContext';
+import { edgeCacheIt } from '~/server/middleware.trpc';
+import { createCallerFactory, publicProcedure, router } from '~/server/trpc';
+import { willEdgeCache } from '~/server/trpc/edge-cache-headers';
+import { TokenScope } from '~/shared/constants/token-scope.constants';
+
+/**
+ * PRECEDENCE between `edgeCacheIt`'s two TTL-overriding inputs: a resolver-set
+ * `ctx.cache.skip` versus a configured `expireAt`.
+ *
+ * `edgeCacheIt` computes the request TTL in three steps:
+ *
+ *     let reqTTL = ctx.cache.skip ? 0 : ttl;                      // incoming skip
+ *     if (expireAt) reqTTL = Math.floor(...);                     // scheduled expiry
+ *     const result = await next();
+ *     if (ctx.cache?.skip) reqTTL = 0;                            // resolver-set skip
+ *
+ * Before the third line existed, `expireAt` silently overrode a pre-set `skip`. It now
+ * loses to it. That is the intended ordering, not an accident of where the line landed:
+ * `expireAt` says "this content goes stale at time T", which is a statement about content
+ * that IS cacheable; `skip` says "this particular response must not be cached at all".
+ * The stronger claim wins — a response the resolver declared uncacheable does not become
+ * cacheable because someone also scheduled when it should expire.
+ *
+ * ── WHY A SYNTHETIC PROCEDURE ───────────────────────────────────────────────────────────
+ * `expireAt` has ZERO production call sites today (`git grep` finds only its type
+ * declaration and its single use inside `edgeCacheIt`), so there is no real procedure that
+ * combines it with a resolver-set `skip` and therefore no cross-file relationship to pin.
+ * The dimension is real regardless — the option is exported and reachable — so it is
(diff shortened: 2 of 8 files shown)


ZacxDev:
## Correction: this PR's description overstates the live impact

Posting rather than editing the body silently, since a reviewer may already have read it.

The description says Announcement blocks "have been cached for 60s against an explicit
intent not to", and that for an authenticated caller it is "worse than a no-op". Both are
true statements **about the code**, and the defect and the fix are real. But they read as a
live regression, and that part is wrong:

**The `HomeBlockType.Announcement` render branch is commented out** — `src/pages/home/index.tsx:84-86`
— and `AnnouncementHomeBlock` occurs nowhere else in the repo. The six live `HomeBlock*`
components each request only their own block type, so **the first-party app never calls
`homeBlock.getHomeBlock` with an Announcement block.** The only way to reach the fixed path
today is the public tRPC surface directly.

So this change should be read as **defence in depth on a dormant path**, not as fixing an
ongoing leak. The inert opt-out is a real defect worth correcting — an opt-out that silently
does nothing is worse than no opt-out, because it reads as coverage — and the path becoming
live again is a matter of uncommenting three lines. But nobody should merge this expecting a
change in cache behaviour they can measure.

Two smaller corrections to my own text while I am here:

1. The new comment on `getHomeBlockById`'s `user` parameter gives edge-caching as the reason
   the parameter must stay unread. That is the **weaker** of two reasons. The by-id result is
   also cached in Redis under a key built from type/identifier/domain with **no user segment**
   (`src/server/services/home-block-cache.service.ts`), so reading `user` there would serve one
   caller's body to another **even with `edgeCacheIt` removed entirely**. A future reader who
   drops the edge cache should not conclude the prohibition lapsed. Happy to amend the comment
   to lead with the Redis key.

2. I fixed one stale comment in `home-block.service.ts` and left its mirror image. `getHomeBlockData`
   still says the session user is *"passed down to collection get items service"* — and unlike the
   one I fixed, that function **does** use `user`. But its only caller on this path invokes it
   **without** one, so collection items already resolve as anonymous there. Either that comment is
   the false one or there is a latent permissions gap behind it. Out of scope here; flagging it
   rather than quietly fixing it in an unrelated PR.

None of this changes the diff, the test, or the mutation results — the middleware change is
one line and its coverage stands as reported.

PR #4383, opened 2026-08-25
fix(rewards): stop paying a rewards-ineligible user the full award

JustMaier:
ClickUp `868kw9kfk`.

`foldUserMultipliers` sets `rewardsMultiplier` to `0` when a user's `rewardsEligibility` is `Ineligible`. `sendAward` turned that `0` back into `1` — a line from `d911a0b65d` (2024-04-04), which predates the eligibility flag and read the `0` as a missing value — so an ineligible user was paid the full, unmultiplied award.

## Why the on-demand path hid it

Only the **processable** path is affected. An on-demand reward multiplies before the Lua cap script, so a `0` multiplier already produced a `capped` row that never reached `sendAward`. A suite built on an on-demand reward passes with the defect fully present, which is why the new tests use a processable one.

## Measurement (prod, 2026-08-25, preceding 7 days — re-measured, not taken from the ticket)

| | |
|---|---|
| `buzzEvents` at `multiplier = 0, status = 'awarded'` | **548 rows / 24 users / 6,520 Buzz** |
| Matching `buzzTransactions` (semi-join on reconstructed `externalTransactionId`) | **470 transactions / 22 users / 3,052 Buzz, all blue** |
| Affected users who are `Ineligible` on the replica right now | **24 of 24** |
| Of the 6,520, attributable to 12042163 (CivitaiOfficial) | **6,358** |

So the honest headline is **~162 Buzz/week across 23 real users**, not 6,238 — the ticket's figure is wrong in two directions (it counts the events rather than the ledger, and it is dominated by an internal account). The ledger, not the ticket.

**Control for that measurement:** the same semi-join with each id prefixed `ZZZ` returns **0** transactions against 470 for the real ids, so the query can come back empty and the 470 is not an artefact of a filter that matches everything.

## The fix

- Drop the `multiplier === 0 → 1` reset.
- Filter `sendAward` on the **computed `amount`** rather than on `awardAmount`. This is the load-bearing guard: it holds even if the `Decimal(3, 2)` ever arrives as a string, where a strict `=== 0` would not. It also keeps 0-Buzz transactions off the ledger.
- Mark a zero-multiplier row `unqualified` in `process()` so the audit trail stops recording a payout that did not happen. `unqualified` is outside the ClickHouse enum and `toClickhouseBuzzEvent` already coerces it to `capped`, which is the existing, deliberate path.

## Declared behaviour change

34 `pending` rows exist all-time with `multiplier = 0` (oldest 2024-11) that would now pay `0` rather than full. They are already stranded: `process` scans `time >= lastUpdate` and never looks back.

## Verification

- **Revert control** — with only `base.reward.ts` reverted to `origin/main`, the new suite is `Tests 3 failed | 2 passed (5)`, exit 1, failing on value assertions (`expected [ …(2) ] to deeply equal []`, `to have a length of 1 but got 2`, `to match object { toUserId: 1, status: 'capped', … }`) rather than on a timeout or a hang.
- **Two of the five new tests pass with the defect present**, and I am not counting them as coverage — they pin `apply()` storing the `0`, and a multiplier above 1 still being applied. Neither is changed by this fix.
- Rewards suites with the fix: `Test Files 11 passed (11)` / `Tests 191 passed (191)`, exit 0.
- Full unit suite: `Test Files 1 failed | 1410 passed | 1 skipped (1412)` / `Tests 1 failed | 22117 passed | 27 skipped (22145)`, exit 1. The one failure is `pageRunScrollContract.test.ts`, a path-separator guard that compares `src/pages/...` against Windows' `src\pages\...`; it is unrelated to this diff and the assertion diff shows the separators directly.
- `blocks.router.workflow.test.ts` collected **358 tests** in this worktree, so the submodule is initialised and the run is not silently empty.
- `pnpm run typecheck`: 0 errors in 301s. `@civitai/client` resolves to 0.2.0-beta.92, matching the root pin.
- `pnpm run lint`: repo-wide exit 1 with 362 pre-existing errors; **0 errors from the two files in this PR** (only `no-explicit-any` warnings matching the sibling test harnesses).

**No migration.**

### DIFF
--- a/src/server/rewards/__tests__/base.reward.ineligible.test.ts
+++ b/src/server/rewards/__tests__/base.reward.ineligible.test.ts
@@ -83,13 +83,16 @@ const processableReward = () =>
     }),
   });
 
-const pending = (toUserId: number, multiplier: number, forId: number): BuzzEventLog => ({
+// `multiplier` is typed loosely because ClickHouse hands `Decimal(3, 2)` back as a number today but
+// would hand back a string if `output_format_json_quote_decimals` were ever on — a case the guard has
+// to survive and a strict `number` parameter would stop us from writing.
+const pending = (toUserId: number, multiplier: number | string, forId: number): BuzzEventLog => ({
   type: 'testIneligibleReward',
   toUserId,
   forId,
   byUserId: toUserId,
   awardAmount: AWARD_AMOUNT,
-  multiplier,
+  multiplier: multiplier as number,
   status: 'pending',
 });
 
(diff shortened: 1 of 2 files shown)


JustMaier:
## Review round — new SHA `b100a2ff27` (was `7672252d46`)

Five review lanes ran over `7672252d46`; all five reported. Four findings acted on, each now carrying a control that fires. Two lane claims I checked and **disproved** rather than acted on.

### The finding I would not have found myself

**Reverting the `sendAward` half of the fix alone left all five tests green.** The `process()` guard sets `awardAmount = 0`, so the old pre-map `awardAmount > 0` filter dropped the row and the resurrected `multiplier = 1` line never ran. Half the diff was unpinned by its own suite.

Pinned with a negative multiplier: it walks past the `process()` guard (it is not 0), passes `awardAmount > 0` — the predicate the old code filtered on — and is caught only by filtering on the computed amount.

### Controls, each from its own run

| Mutation | Result |
|---|---|
| Full revert of `base.reward.ts` to `origin/main` | `Tests 5 failed \| 2 passed (7)`, exit 1 (was 3 failed of 5) |
| Revert the **`sendAward` half only**, keep the `process()` guard | `Tests 1 failed \| 6 passed (7)` — `expected [ { type: 5, toAccountId: 2, …(6) } ] to deeply equal []`. **Green before this round.** |
| `Number(event.multiplier) === 0` → strict `=== 0` | `Tests 1 failed \| 6 passed (7)` — the string-Decimal test |
| `'unqualified'` → `'capped'` | `Tests 1 failed \| 6 passed (7)` |

After each control I restored from a saved copy and confirmed the file was byte-identical (`diff -q` → IDENTICAL). Worth recording: a `grep` for `unqualified` on the restored file returned **empty with exit 0** — a known rewritten-grep trap on this box. A `python` re-scan found 8 occurrences. Trusting the grep would have read as "I corrupted my own file".

### The other three fixes

- **`Number(event.multiplier) === 0`.** The `Decimal(3, 2)` arrives unquoted today, but `output_format_json_quote_decimals`, a swapped client or a `toString(multiplier)` in `process-rewards.ts` all make it `'0.00'`, against which strict `===` goes silently inert — the row would be recorded `awarded` while paying nothing. The PR previously argued the amount filter was robust to exactly this while the guard beside it was not.
- **`'unqualified'` vs `'capped'` are now distinguishable.** They are not interchangeable: the send-failure recovery resets everything *not* `unqualified` back to `pending` at the full award, which would put an ineligible row back in the pipe. The suite now asserts the `statusRaw: 'unqualified'` that `toClickhouseBuzzEvent` records alongside the coerced status.
- **The unqualified-row assertion was also satisfied by the reward being disabled**, which marks every row the same way. It now carries an eligible row in the same batch asserted `awarded`, as an in-test control.
- **Deleted an assertion that could not fail:** `sendAward` is unreachable from `apply()` on a processable reward, so asserting nothing was sent there passed regardless of the payment code.

### Two lane claims I disproved

- **"Any live `Product` with `rewardsMultiplier: 0` would newly be zeroed."** Measured: 17 products carry the key, **0** of them are zero, 17 positive. Not live. (The 17 is the control — the query can see the field.)
- **"The moderator falsy-zero is paying ineligible reporters."** Real defect, but measured: of **1,369** distinct users who received an `awarded` `reportAccepted` row in 30 days, **1,368 are `Eligible` and 0 `Ineligible`**. Latent, not money moving. Control: the same query shape returned `Ineligible` for 24 of 24 on this ticket's own population.

### Corrections to this PR's original body

- **"~162 Buzz/week across 23 real users" is an event-side upper bound, not a ledger figure.** It is `6,520 − 6,358` from the `buzzEvents` column, in a paragraph arguing for the ledger. Since dedup roughly halves events→ledger, the true ledger residual is plausibly lower. The conclusion stands — the ticket's 6,238 is wrong and dominated by an internal account — but the number is not ledger-derived and should not be read as one.
- **The `process()` branch is more than audit hygiene.** The cap query sums `awardAmount WHERE status = 'awarded'`, so without it an ineligible user's *unpaid* awards would consume their own cap and short a later eligible grant inside the same interval. That is a money-affecting correction, not bookkeeping.
- **The ineligibility convention is knowingly still open on one path.** `apps/moderator/src/lib/server/rewards.ts:51` truthiness-tests the cached multiplier, so a `0` becomes `1` and the row reaches this same cron carrying `multiplier >= 1` — where the new guard cannot see it. Being fixed in the `868kw9b31` PR. Note that moderator-written rows therefore **structurally cannot appear** in this PR's `multiplier = 0` measurement: re-running that query to check whether the convention holds everywhere returns a clean answer that means nothing for `reportAccepted`.

### Verification at `b100a2ff27`

- Rewards suites: `Test Files 11 passed (11)` / `Tests 193 passed (193)`, exit 0.
- `pnpm run typecheck`: 0 errors in 300s, exit 0.
- `pnpm run lint`: repo-wide exit 1 on 362 pre-existing errors; **0 errors from the two files here** (warnings only, matching the sibling harnesses).
- Full unit suite at the pre-fix tree: `Test Files 1 failed | 1410 passed | 1 skipped (1412)`. The full suite for `b100a2ff27` is re-running; I will post its `Test Files` line rather than infer it.

**Unrelated failure found in the full suite, flagged for whoever owns it:** `src/server/meilisearch/__tests__/cleanup.test.ts > resumes from a cursor that is just inside the staleness bound` failed with `expected null to be 6`, then **passed alone** (`Tests 62 passed (62)`). It is a genuine boundary flake, not contention noise: the test seeds `startedAt: Date.now() - MAX_CURSOR_AGE_MS` and `cleanup-cursor.ts:116` rejects on `age > MAX_CURSOR_AGE_MS`, so any millisecond elapsing between seeding and the check flips it to stale. Under a loaded 31-worker run that millisecond elapses. Nothing to do with this diff — my two files are not in that module's import graph (`grep -c rewards` → 0 in both `cleanup.ts` and its test).

PR #4386, opened 2026-08-25
feat(download): opt-in origin-direct resolution for allowlisted internal clients

ZacxDev:
Adds `STORAGE_RESOLVER_DIRECT_USER_AGENTS` — a comma-delimited allowlist of user-agent substrings whose model downloads resolve to an **origin-direct** URL instead of the CDN-fronted one.

**Ships inert.** The allowlist is empty by default, and a test asserts that an empty allowlist means *nobody* — including the agent we intend to add later. Turning it on is a config change.

## Why

Some internal clients fetch a single large file over many parallel connections. The CDN in front of our storage does not get faster with more streams, and refuses ranged requests above a size threshold; the storage origin does neither. Serving those specific clients directly is measurably faster for them and measurably more expensive for us — which is why this is an allowlist naming particular clients rather than a boolean.

## How it threads

```
shouldResolveDirect(req)
  → getFileForModelVersion({ ..., direct })
    → resolveDownloadUrl(fileId, url, name, { direct })
      → the resolver's `direct` field
```

Every other caller of `resolveDownloadUrl` (vault, training-data, presigned-url) is untouched and resolves exactly as before.

`direct` is **omitted** from the resolver request rather than sent as `false`, so a resolver build that does not know the field receives a byte-identical request to the one it gets today. That removes any ordering dependency between this PR and the resolver-side one.

## 🔴 `shouldResolveDirect` is not an access control

The doc comment says this in-tree, and it is the thing to review hardest. A user agent is caller-supplied and trivially spoofed. It is safe *here* because of what it gates:

- The flag is read **after** the route's IP blocklist, user blocklist, download rate limit, and — inside `getFileForModelVersion` — the auth, ownership, POI, paid-access and archived checks.
- The caller receives the **same file it was already entitled to**. Spoofing the agent changes which host serves those bytes, not whether the caller may have them.
- The cost of a wrong `true` is bandwidth, not exposure.

If that ever stops being true, the fix is an authenticated signal — not a harder string match. Flagging it explicitly rather than burying it.

## Tests

11 new.

The one worth reading is `never lets an empty allowlist entry match everything`. A trailing comma in config produces an empty entry, and `''.includes(x)` is `true` for every string — so the naive implementation turns one stray character into a fleet-wide rollout, silently and at real cost.

Mutation-checked rather than assumed: removing the `needle.length > 0` guard fails exactly that test and leaves the other six green, so it fails for its own reason rather than because everything broke.

Also covered: default-off across four agent shapes, substring and case-insensitive matching, near-miss agents that a sloppy match would let through, whitespace tolerance, missing user-agent, the flag reaching the resolver body, omission when unset, and the delivery-worker fallback still succeeding for a `direct` request it cannot honour.

Typecheck 0 errors, eslint clean, and the pre-existing `delivery-worker.resolver-enabled` suite still passes (31 tests green across the three files together).

## Review focus

- The security argument above — specifically whether "read after every access gate" is as true as I claim. That is the load-bearing sentence.
- Whether an allowlist keyed on user agent is the right shape, or whether this should wait for an authenticated internal signal.


ZacxDev:
## Correcting two claims in the PR description above

An adversarial audit found the description overstated its own coverage. Correcting it here rather than silently editing the body, since it may already have been read.

**1. "Mutation-checked rather than assumed" was true of the mutant it named, and read as coverage of the change. It was not.**

The mutation check covered `shouldResolveDirect` in isolation. The seam between it and the endpoint had **zero** mutation resistance. Measured on the original commit — with `direct: shouldResolveDirect(req)` replaced by `direct: true` at the route, the entire download-path suite passed: **4815 tests, zero failures**. So did deleting the argument, inverting it, and hardcoding the options bag in `file.service`.

That matters because the always-on mutant is the one with a cost, and "ships inert" was the description's headline safety claim. The test that backed it asserted inertness of the *helper*, not of the *endpoint*.

Fixed in `e028ab5`: seam assertions at both boundaries, mocked at the env boundary rather than by stubbing the helper, asserting the exact boolean by argument position (`expect.objectContaining` is what made the existing route assertions blind — it cannot see a field that was added, removed, or inverted). All seven mutants now fail; the unmutated control is green before and after the battery.

**2. "Every other caller of `resolveDownloadUrl` (vault, training-data, presigned-url) is untouched" presented a sample as a complete list.**

There are **nine** other call sites, not three. The three named, plus `model.controller.ts` ×2, `model3d.service.ts`, `wildcard-set-provisioning.service.ts`, and `orchestrator.service.ts` ×2.

The substance is unchanged and I re-verified it by enumeration rather than by sampling: all nine pass three arguments or fewer, so `options` is `undefined` and `direct` is omitted from the request body — byte-identical to today. The four other `getFileForModelVersion` callers likewise omit `direct`.

## Also changed in `e028ab5`

**The allowlist floor.** The empty-entry guard closed length 0 and nothing else, so a one- or two-character entry — a truncated paste, not just a trailing comma — still matched essentially every real user agent (`"c"` is inside `Mozilla/5.0 (Macintosh…)`) and produced the same fleet-wide rollout the guard exists to prevent. Now requires 3 characters.

**Two comments that misstated reality.** The off switch an operator would actually reach for — setting the variable to an empty string — parses to `['']`, not `[]`, because the schema default applies only when the key is *absent* and the comma splitter does not drop empty segments. It never took the early return; the per-entry floor is what disables it. And "the cost of a wrong `true` is bandwidth, not exposure" understated the cost half: there is no exposure, but a wrong `true` moves a download from zero-rated to billed egress.

## Known limitations, not addressed here

- **No off switch faster than a redeploy.** `env` is materialised once at module load, so turning this off under a cost incident is an env change plus a rollout, not a flag flip. This repo runs Flipt; moving the allowlist there would be the better shape and is deliberately not in this PR.
- **The allowlist names a string, not a population.** It is a routing hint for clients we operate and provides no containment against anyone who copies an allowlisted agent string. The incentive is unhelpful — direct is slower single-stream, so it rewards precisely the high-volume parallel downloader. There is no exposure in that, only cost. If containment ever needs to be real, the fix is an authenticated signal rather than a harder string match.
- **No frequency measurement** of spoofing, and none of the share of download traffic this would move. Neither the original description nor the audit had one; the merge decision should not assume either.

Unrelated lines touched, deliberately: `file-download-lookup.test.ts` carried three pre-existing `import()` type-annotation lint errors, invisible because CI lints only *changed* files and nothing had touched it. Adding a describe block brings the whole file into that set, so they are converted to top-level type-only imports rather than left to fail the gate.

ZacxDev:
Two updates since this was opened.

**1. Branch updated from `main`.** The `preview / component-tests` failure on the previous head was not from this PR — `ImagesAsPostsCardLazyCarousel.browser.test.tsx` was failing at import time on `main` since #4364, and was fixed by #4393. This branch was 44 commits behind and predated that fix. Every `component-tests` run before ~21:28Z failed across four different PRs; every run after it passed across six. Now that the branch carries the fix, the check is re-running on the merged tree rather than a stale branch.

**2. Pushed `90fb4840c3`** — a direct ask now carries the resolver's internal bearer token.

Review of the paired storage-resolver PR (civitai/storage-resolver#12) turned up that `/resolve` is unauthenticated and publicly reachable — the ingress deliberately routes that path around its basic-auth middleware. Since `direct` moves bytes off zero-rated CDN egress onto our billed B2 allowance, the user-agent allowlist in this PR was never sufficient on its own: it governs callers who come through *this app*, and nothing stopped anyone POSTing the resolver directly. The resolver now requires `INTERNAL_API_TOKEN` for `direct`, so this side has to send it.

It reuses `STORAGE_RESOLVER_INTERNAL_TOKEN`, which `~/server/services/storage-resolver` already uses for register/deregister — no new env var. Bearer *replaces* the Basic header on a direct ask (there's only one `Authorization` header); nothing that works today stops working, since the Basic credential isn't validated on `/resolve` anyway.

The judgement call worth a second opinion: **when the internal token is absent we omit `direct` entirely rather than sending an ask we know will be refused.** The resolver counts refusals as `granted="unauthorized"`, and that counter's whole value is meaning "someone found the cost lever" — our own missing config shouldn't fill it. The cost is that a half-finished rollout (allowlist set, token not) is silently inert instead of loud. The observable is `granted="true"` staying at zero, which the enable checklist watches. Happy to flip it if you'd rather have the noisy failure.

Mutation-checked both directions; the dropped-token-requirement mutant produces a literal `Bearer undefined` header, so that one would have shipped a malformed credential. Ran the whole download/resolver neighbourhood: 18 files, 201 tests, green.

Still inert on merge — `STORAGE_RESOLVER_DIRECT_USER_AGENTS` is empty by default and no caller sets it.

PR #4401, opened 2026-08-25
feat(app-blocks): a moderator-initiated free-text message to an app's owner

ZacxDev:
## The gap

The platform already notifies app owners **eleven** ways (`app-block-approved/rejected`, `app-listing-approved/rejected/hidden/reset-to-pending`, `app-collaborator-invited/accepted/removed`, `app-ownership-transfer-offered/accepted`). Every one is **event-triggered with fixed copy**. Four of them splice a moderator's `reason` into that fixed sentence — but the *subject* is always chosen by the code path, never by the moderator.

So a moderator who needs to say something the platform has no event for has had two options: push a change to someone else's app, or find the developer off-platform. The concrete case: a third-party listing (`prompt-vault`) claims it *"estimates the cost and asks before it spends"*, which it has never done; a second (`df-qwen-canvas`) has an entirely empty listing.

This adds a **twelfth notification type** and the smallest mod-only path to it.

```
appListings.messageAppOwner   (moderatorProcedure + the inner isModerator recheck)
  → resolveListingAccess         the CANONICAL owner, never AppListing.userId
  → bounds + throwOnBlockedLinkDomain
  → two Redis fixed windows      per-moderator AND per-listing
  → one AppListingModerationEvent   action `message-owner`  ← written BEFORE delivery
  → notifyAppModeratorMessage → createNotification   one call, N userIds
```

### The recipient is the canonical owner — `resolveListingAccess(id, null).ownerUserId`

`resolveCanonicalListingOwner` is the one place the kind branch is written: `onsite` → `AppBlock.app.userId` with the column as fallback; `offsite` → the column **unconditionally**, even when a block is attached (issue #3844). `resolveListingAccess` wraps it, hops a shadow revision to its parent, and takes `kind`/`slug`/`appBlockId`/`userId` from the parent row.

On an on-site listing `AppListing.userId` is a **denormalized copy** frozen at shadow-clone time. Reading it here would be worse than a missed delivery: it would send a moderator's private message *about* someone's app *to* a user who no longer owns it — a disclosure — while the real owner is never told their listing has a problem. The drifted-row fixture is the only fixture that can see this; one where the column and the block agree is green against both the correct and the broken resolution.

⚠️ **SCOPE — this is LATENT, not live, and the earlier framing overstated it.** Measured against production during review: **21 on-site parents, 21 with a backing block, 0 drifted; 0 accepted ownership transfers ever; 0 off-site listings carrying a block.** So there is no live damage here and this is defence-in-depth against a state that does not exist yet. It is still worth doing — it is correct-by-construction, costs one query, and the drift goes live on the first on-site transfer of a listing with an in-flight revision — but it is not a fix for something currently broken.

`userId: null` is deliberate: this caller is a moderator who is not a party to the app. The resolver's second argument asks "what is *this caller's* role", which nothing here reads — and `null` takes the early return, so the owner resolves in exactly **one** query with no seat lookup.

### The audit row precedes the delivery

⚠️ **Scope:** this is a property of **this procedure**, not of the notification type. The webhook-token-gated `src/pages/api/mod/send-mod-notification.ts` can emit any type — including `app-moderator-message` — with arbitrary `userIds` and text, no audit row and no rate limit. Not moderator-reachable, and equally true of the eleven existing app types, but the claim is about the proc.

Order: resolve → validate text → spend quota → **write the event** → deliver (best-effort).

A notifications outage costs the delivery, never the record that a moderator sent a message. The inverse would let a mod message a developer with nothing in the log. `reason` carries the subject and `detail` the body, so the history renders *what was said*, not merely that something was; `after` records the delivery decision (`recipientUserIds`, `includeCollaborators`), which is the part a later reviewer cannot reconstruct because the canonical owner can change afterwards.

Text validation runs **before** the quota spend so a rejected draft does not consume the *recipient's* harassment budget — a mod fixing a typo would otherwise burn the listing's hourly allowance.

### The moderator is not named to the recipient

Attribution lives in `AppListingModerationEvent.actorUserId` — durable, mod-visible, survives an account delete. Naming the individual on the wire would make "who acted on my app" answerable by a developer, a retaliation vector the rest of the moderation surface already declines to open (`delistListing` / `rejectRequest` carry the mod's *reason* to the user and never their identity). The owner-scoped history projection already drops `actor`, so this is consistent end to end.

### Content safety: `assertSharedTextSafe` is deliberately **not** applied

That belt exists for App Blocks *shared storage* — untrusted, cross-user, public text. Three of its five checks are actively wrong here:

- `includesMinor` / `includesPoi` **hard-fail** the text. A moderator writing *"your cover image appears to depict a minor — take it down"* is exactly the message this channel exists to carry, and the belt refuses to send it. **The control would block its own most important use.**
- `auditPromptServer({ isGreen: true })` forces an SFW/profanity ceiling on staff↔developer correspondence, hard-fails the send, and appends `GREEN_SFW_REDIRECT` — *"please visit civitai.red where you have more freedom to generate mature content"* — which is nonsense copy to put in front of a moderator writing to a developer about their app.

  ⚠️ **CORRECTED.** An earlier version of this section claimed the audit accrues blocked-prompt strikes against the sender. **That is wrong.** With `isGreen: true` the handler takes the `else if (isGreen)` branch in `promptAuditing.ts`, which calls neither `addBlockedPrompt` nor `reportProhibitedRequest` — no strike, no ClickHouse row, no auto-mute — and `assertSharedTextSafe` threads `isModerator` through anyway. The real cost is the hard-failed send and the wrong copy, which is narrower than what was claimed. The **conclusion is unchanged**: leg (a) is correct and independently sufficient. Left visible rather than silently rewritten, because the overstated version was quoted in review.

It would also be inconsistent rather than protective: **every** existing moderator free-text field that renders verbatim into a user-visible notification — `delistListing.reason` → `app-listing-hidden`, `rejectRequest.reason` → `app-block-rejected`, `tos-violation.reason` — runs no content safety at all. A belt on this one path and none on those four is a spelled control, not a real one.

**What is kept: the link check.** Moderator status establishes trusted staff; it does not survive a *compromised mod session*, and the one thing such a session can do here that mod-status does not mitigate is put a hostile URL in front of a developer wrapped in the platform's own "Civitai moderation sent you a message" framing. `throwOnBlockedLinkDomain` over subject+body costs one cached Redis read. Plus hard bounds, re-asserted in the service independently of the zod schema.

### Rate limiting: **not** the tRPC `rateLimit` middleware

```ts
if (ctx.user?.isModerator || isDev || isTest || isPreview) return await next();
```

`rateLimit` short-circuits for moderators, so on a `moderatorProcedure` it is a guaranteed **no-op** — it would cap nothing while reading as a limit in the router source, in review, and in a diff. This follows `blocks-retrigger-rate-limit.ts`, which exists for exactly this reason. A test asserts the source, because the behaviour is unobservable: an inert middleware and no middleware are indistinguishable from outside.

**Two counters, because they bound different harms.** The per-moderator window (30/h) bounds a runaway client or a compromised session; the per-listing window (5/h, across *all* moderators) is the harassment ceiling, which a per-moderator cap cannot express — three moderators at nine sends each are all individually under a cap of ten while the recipient gets 27.

**Fail open**, matching the sibling limiters. The accountability mechanism is the Postgres audit row (written before delivery, carrying `actorUserId`), not the counter; failing closed would let a Redis incident silently block moderation correspondence while every other mod action kept working.

### Collaborators are opt-in, default off

The **owner** is the accountable party. An accepted collaborator consented to *edit* an app, not to receive moderation correspondence about it, and a moderator's message can be adverse — fanning it out by default would broadcast a private accusation to third parties. But the failure in the other direction is real: on an app whose owner is unresponsive, the accepted editors are the only people who can fix the listing. That is a per-message judgement, which is what an explicit `includeCollaborators` is for. `displayed` is **not** consulted — it is a public-credit preference, not a capability.

The `{ status: 'accepted' }`-without-`displayed` predicate now has two callers, so it was extracted to `listAcceptedCollaboratorUserIds` and `listAppInsiderUserIds` delegates to it.

### The notification identifies the app by **slug**, not name

The sibling families carry `name` with a terse fallback. Here the recipient may own several apps, and one of the two listings that motivated this feature has an empty listing — a null name — i.e. the fallback would fire exactly where identification matters most. The slug is `NOT NULL`, is the app's public identity (`<slug>.civit.ai`), and is already resolved by the owner lookup.

### The copy says replies are not delivered

A notification is one-way and nothing routes a reply anywhere. Leaving that implicit would strand a developer who wants to answer, so the reply route is appended to every rendered message, and is part of the string rather than an optional field so it cannot be forgotten on one path.

## What I did **not** build

No inbox, no thread, no reply route, no read receipts, no message table, no UI. No widening of the owner-scoped history projection (it drops `detail`, so the owner sees the subject in history and the full body in the notification — widening it is a privacy change across *all* actions, not just this one). No new moderator-gating mechanism. No second error mapper.

## ⚠️ Manual migration, required before this ships

`20260824120000_app_listing_mod_action_message_owner` widens the `app_listing_mod_events_action_check` CHECK. Per the DB rule this repo follows, migrations are **applied by hand** per environment — apply to the dev clone before a preview exercises the proc, and to prod before main → release.

The failure mode without it is the safe one: the audit write is rejected (23514), so **no message is sent unrecorded**.

## Test matrix

Files: 5 new test files (72 cases) + 2 existing ledgers extended. Nothing added under `src/pages/**`.

**Red at `origin/main`, green at HEAD** — measured in a pristine `origin/main` worktree with my test files copied in:

| Test | at base | at HEAD | kind |
|---|---|---|---|
| `mod-actions-authz` → "router exposes … alongside the other mod actions" | ✗ FAIL (assertion) | ✓ | regression |
| `call-site-ledger` → "gate sites EXACTLY equals the ledger (GROWTH/SHRINK)" | ✗ FAIL (assertion) | ✓ | regression |
| `call-site-ledger` → "every ledger entry names an actual file" | ✗ FAIL (assertion) | ✓ | regression |
| `app-moderator-message.service.test.ts` (27) | ✗ COLLECT-ERROR — module absent | ✓ | **new-code, not regression** |
| `app-moderator-message-rate-limit.test.ts` (11) | ✗ COLLECT-ERROR — module absent | ✓ | **new-code, not regression** |
| `app-moderator-message-notify.test.ts` (3) | ✗ COLLECT-ERROR — module absent | ✓ | **new-code, not regression** |
| `app-moderator-message.notifications.test.ts` (8) | ✗ COLLECT-ERROR — module absent | ✓ | **new-code, not regression** |
| `app-listings.router.messageAppOwner.test.ts` (23) | ✗ COLLECT-ERROR — module absent | ✓ | **new-code, not regression** |
| `app-listing-mod-action.constants.test.ts` (3) | ✓ PASS | ✓ | **invariant guard** (re-derives; see M22/M23) |

Stated plainly: only the first three are regression evidence in the strict sense. The five COLLECT-ERROR rows are red at base because the module does not exist there — structurally red, not assertion-red, and I am not counting them as regression coverage. For those the evidence is the mutation battery below.

Two labelled non-guards, kept but not counted:
- the **anonymous** authz case survives a downgrade to `protectedProcedure` (`isAuthed` refuses before `isMod` runs) — it is an auth guard, not the mod gate. The **tester** and **developer** cases are the mod gate (gate 1), and both die to M31.
- 🔴 **the inner `isModerator` recheck (gate 2) is UNREACHABLE and is labelled, not counted.** It tests the identical predicate to `isMod` (`trpc.ts`: `if (!user.isModerator) throw FORBIDDEN`), so no context can reach the proc body with a non-moderator, and deleting its throw leaves every *behavioural* test green. It is defence-in-depth against a downgrade of gate 1. Its removal is caught by a **structural** case that says so in as many words (M32).
- `app-listing-mod-action.constants.test.ts` is green at base by construction (it compares the code tuple to the latest migration, and at base both lack the action). M22/M23 are what prove it is live.

## Mutation table — **25 defined, 25 verdicts produced, reconciled**

Harness controls run first: an **unmutated** run of all 7 target files was GREEN (8/23/48/3/3/27/11 passing, 0 failing) — the harness can report success. Every mutant asserted its search string occurred exactly once before replacing; 0 `PATTERN-MISS`. Files restored from byte copies, not git. Verdicts counted from the summary line, never an exit code. `killed_by_named_test` requires **that** guard's own test in the failing set — a mutant dying to a neighbour is not scored as coverage.

| # | Mutation | Defect produced | Killed by |
|---|---|---|---|
| M1 | recipient read from `AppListing.userId` | msg to the stale owner on a drifted onsite row | "DRIFTED on-site listing notifies the BLOCK owner" |
| M2 | send moved before the audit `create` | delivery with no record | "audit write is invoked BEFORE the send" |
| M3 | `\|\| !listingQuota.allowed` → `\|\| false` | harassment ceiling unenforced | "exhausted LISTING window refuses…" |
| M4 | text validation moved after the quota spend | a rejected draft burns the recipient's budget | "blocked link domain refuses with BLOCKED_LINK" |
| M5 | drop `.trim()` on body | whitespace-only message delivered | "whitespace-only body is refused as INVALID_TEXT" |
| M6 | `input.includeCollaborators` → `true` | collaborators broadcast by default | "by DEFAULT an accepted collaborator is NOT notified" |
| M7 | recipients array instead of `Set` | owner-with-a-seat notified twice | "owner who also holds a seat is notified ONCE" |
| M8 | `seatListingId` → `appListingId` | shadow gets its own quota + orphan audit row | "SHADOW revision id resolves to the PARENT" |
| M9 | link scan over body only | a hostile URL in the subject unscanned | "link scan covers the SUBJECT as well as the body" |
| M10 | `count <= max` → `count < max` | one fewer send than the stated ceiling | "allows a send at exactly the ceiling" |
| M11 | listing window uses the moderator ceiling | one app can receive 6× the intended volume | "LISTING ceiling is its own number" |
| M12 | TTL re-armed unconditionally | sliding-forever window ⇒ lifetime cap | "does NOT extend an active window" |
| ~~M13~~ → M30 | **withdrawn as non-productive**: an integer `userId` and an `apl_*` string cannot collide under any prefix, so "the two counters share a bucket" is not reachable by a small mutation — by this PR's own criterion that row was a false alarm. Replaced by the reachable defect: the listing key written into the **actor namespace** | "each helper writes its OWN namespace and its own subject" |
| M14 | `userIds` → `userId: userIds[0]` | only the owner is delivered to | "uses `userIds`, not a per-recipient call" |
| M15 | drop the empty-recipient guard | a dead PendingNotification burns the key | "EMPTY recipient list emits NOTHING" |
| M16 | drop the "replies are not delivered" clause | one-way channel with no route out | "always states that replies are not delivered" |
| M17 | render subject, drop body | a message with no content | "carries BOTH halves of what the moderator wrote" |
| M18 | remove the registry spread | `prepareMessage` never found ⇒ renders null | "is REGISTERED in the shared processor map" |
| M31 (was M19) | `moderatorProcedure` → `protectedProcedure` | **corrected:** *not* "any user can message any owner" — the inner recheck still refuses, with `UNAUTHORIZED`. The mutant dies on the **error-code mismatch**, and `mockMessageAppOwner` is never called under it either | "GATE 1 (moderatorProcedure): a non-moderator is FORBIDDEN" |
| M20 | `RATE_LIMITED` arm's code flipped | quota exhaustion reads as a bad request | "RATE_LIMITED maps to TOO_MANY_REQUESTS" |
| M21 | `OffsiteRequestError` dropped from the shared mapper | existing delist/claim errors → INTERNAL | "mapping the NEW error name did not break the two existing" |
| M22 | `'message-owner'` removed from the code tuple | code/DDL drift ⇒ 23514 at runtime | action-CHECK ⟺ tuple agreement |
| M23 | `'message-owner'` removed from the migration CHECK | same drift, other side | action-CHECK ⟺ tuple agreement |
| **M24** | **POSITIVE CONTROL** — `action: 'message-owner'` → `'delist'` | wrong audit action | "audit row carries the action, the actor…" |
| M25 | proc renamed away | the endpoint does not exist | shared authz ledger |

**All 25 KILLED, all `killed_by_named_test: true`.**

M3, M6 and M20 were deliberately narrowed to the smallest wrong expression rather than deleting the enclosing condition — a mutant that removes a guard together with its `if` proves nothing about the guard.

## One real gap this change introduced, found by an existing guard

`appListingModerationView.test.ts` iterates `APP_LISTING_MODERATION_ACTIONS` and asserts each has a chip label that is not the raw key. Adding `message-owner` to the tuple without a label failed with `expected 'message-owner' not to be 'message-owner'` — the event would have rendered its own action string in the history table. Fixed in the second commit. That map serves the **owner's** history as well as the mod queue, so the label is "Message from moderation", not "Messaged the owner".

## Runs

| Gate | Result |
|---|---|
| `pnpm typecheck` | **OK — 0 type errors** (116s, heap cap 8192 MB) |
| `pnpm test:unit:run` | **22,361 passed / 0 failed / 25 skipped**, 1439 files |
| `pnpm test:component` | 🔴 **COULD NOT RUN LOCALLY** — see below |
| targeted re-run after rebase | 167 passed / 0 failed (10 files) |

`pnpm test:component` collected 189 files and ran **zero** tests: `browserType.launch: Executable doesn't exist at …/chromium_headless_shell-1200/…`. The nix-provided browser set is `-1228`. **Verified identical in a pristine `origin/main` worktree**, so it is a local Playwright/nix version mismatch, not this change — but it means the component tier is **unverified locally**, and I am reporting that rather than a green. No `.browser.test.tsx` imports the one `src/components/` file this PR touches.

Separately, `feature-flags.early-adopter.seam.test.ts` failed in the first full run for a worktree-setup reason (`apps/auth/.svelte-kit/tsconfig.json` absent). Confirmed failing identically at `origin/main` in a clean worktree, then fixed by `svelte-kit sync`; the run above is after that.

## Drift

Rebased onto `origin/main` twice during the work. At the final rebase, `main` had moved 14 files; **zero overlap** with the 18 this PR touches. Branched directly off `origin/main` — not stacked.

---

## CI (updated after the lint fix)

**11 pass · 1 skipped by design · 1 still running.**

| Check | |
|---|---|
| App unit tests + typecheck | ✅ 2m28s |
| Unit tests (1)–(4) | ✅ 3m42s / 3m14s / 2m59s / 3m45s |
| Package unit tests | ✅ 1m24s |
| ESLint + Prettier (changed files) | ✅ 1m24s |
| Schema drift gate | ✅ 58s — the CHECK widen needs no Prisma schema change |
| event-engine-common pin | ✅ 25s |
| Windows dev-env (bash / pwsh) | ✅ 2m58s / 2m37s |
| Typecheck (main pushes / fork PRs / non-main base) | skipped by design — same-repo PR against `main` is covered by "App unit tests + typecheck" |
| `preview / deploy` | ⏳ still running at time of writing |

The first ESLint run failed on one real error — `consistent-type-imports` for a `TRPCError` import used only in a cast — plus five `no-unused-vars` warnings on named-but-unused `vi.fn` parameters. Fixed in the third commit; dropping the parameters is behaviour-neutral (`vi.fn` records the arguments it is *called* with regardless of its declared signature), re-verified by re-running the three files and `pnpm typecheck`.

⚠️ **`preview / component-tests` is known-red on `main`** on `ImagesAsPostsCardLazyCarousel.browser.test.tsx`, bisected to `d4c4a4e4cb` (#4365). Not this branch.

---

## Post-audit revision (commit 5)

An adversarial audit cleared the primary path — recipient resolution on every branch, audit-before-delivery, the authorization boundary, the migration's safe failure mode — and found **three defects plus two vacuous guards**. All fixed; the narrative corrections above are marked in place rather than silently rewritten.

### 🔴 1. The blocked-link `catch` was too wide, and inverted this feature's own fail-direction

`throwOnBlockedLinkDomain` reaches `getBlocklistDTO`, whose **first statement is an unguarded `await redis.get(...)`**. The bare `catch` turned a Redis outage into *"The message contains a blocked link domain."* about a message with no link, leaving the moderator to edit text that is fine — and because `BLOCKED_LINK` maps to `BAD_REQUEST`, it never reached the `INTERNAL_SERVER_ERROR` branch that feeds the server-fault logger, so **the outage was invisible from this path**.

It also contradicted the argument two sections up: the rate limiter fails *open* precisely so a Redis incident cannot block moderation correspondence, and this put a fail-*closed* Redis dependency in front of the same send.

Now only a `TRPCError`/`BAD_REQUEST` — what `throwBadRequestError` actually produces on a real block — is relabelled; anything else is rethrown as infra.

**Why the suite could not see it:** the BLOCKED case was fixtured with `new Error('invalid urls: …')`, which is indistinguishable from `ECONNREFUSED` at a bare catch site. It is now a real `TRPCError`, with an `ECONNREFUSED` case beside it — the pair is what makes either one mean anything.

`assertSharedTextSafe` has the byte-identical wart, so this is a copied house pattern rather than a novel defect; fixing it there is a separate change.

### 🔴 2. Half the content-safety rationale was factually wrong

See the ⚠️ CORRECTED note inline above. Conclusion unchanged, leg (a) independently sufficient, leg (b) replaced with the true and narrower claim. Leg (c) was independently **verified true**: `delistListing.reason` gets `requireModReason` (trim + 3-char floor) and nothing else.

### 🔴 3. A rejected send burned the *recipient's* harassment budget

Both windows were spent unconditionally before either was checked. A moderator who had exhausted their own 30/h window and kept retrying still `INCR`'d the per-listing key each time — **five retries by one already-capped moderator would exhaust that app's 5/h ceiling and lock out every other moderator for the rest of the hour**, over messages that were never sent.

That contradicted the ordering principle this PR already enforced for text-validation rejections. The actor window is now checked before the listing window is touched. The short-circuit runs in **one direction only** — skipping the listing spend when the actor is *allowed* would leave the ceiling unenforced in the normal case — and both directions are pinned.

### 🔴 The two source-level guards were VACUOUS

Both sliced the proc body to `'\n  }),'` — two spaces, where the real terminator has four. `indexOf` returned `-1`, `slice(0, -1)` kept nearly the whole 3k-line router, and the assertions were answered by **other procedures' source**. Measured: deleting this proc's inner `isModerator` recheck left "both gates are present" **green**, because a sibling's identical line was in scope.

The rate-limit guard was worse: it asserted `not.toContain('.use(rateLimit(')`, and this router **never uses that spelling** (it writes `.use(\n      rateLimit({`). The assertion was true no matter what, and would have passed with a rate limit sitting on the proc.

Both fixed, and a **positive control** now asserts the slice is bounded — under 1200 chars and containing no sibling proc — so a future terminator change fails loudly instead of silently restoring the whole-file scan (M34).

### Four coverage gaps the auditor's mutants found, all now closed

| Gap | Consequence if wrong | Now pinned by |
|---|---|---|
| collaborator read used the passed id, not the PARENT | a shadow id delivers to the owner alone while the audit row says the editors were looped in | "the collaborator read is keyed on the PARENT listing" (M28) |
| `after.includeCollaborators` asserted only in the true case | misreports on the **default** path — the common one, and the reason `after` exists | "records FALSE on the default path" (M29) |
| the limiter key test asserted only that the keys differ | each helper writing the *other's* namespace | "each helper writes its OWN namespace" (M30) |
| gate 2's removal was uncaught | the belt is silently deletable | the structural case + its positive control (M32/M34) |

Also corrected: `app-moderator-message.notifications.test.ts`'s "does not name the moderator" case claimed to assert the **type's** field set. TS types are erased, so it does not — it asserts a hand-written fixture. The real guard is in the service test (confirmed by mutation) and the comment now cites it.

### Round-2 mutation table — 10 defined, 10 verdicts, reconciled, all killed by their own named test

| # | Mutation | Defect produced | Killed by |
|---|---|---|---|
| M26 | the narrowed catch → bare `catch` | a Redis outage reads as a blocked link, invisibly | "an INFRA failure in the link scan is NOT relabelled" |
| M27 | listing quota spent before the actor check | a capped mod's retries burn the recipient's ceiling | "an ALREADY-CAPPED moderator does NOT spend the LISTING budget" |
| M3b | `if (!listingQuota.allowed)` → `if (false)` | harassment ceiling unenforced | "exhausted LISTING window refuses…" |
| M28 | collaborator read off the passed id | silent under-delivery on a shadow id | "collaborator read is keyed on the PARENT listing" |
| M29 | `after.includeCollaborators` hardcoded `true` | the audit misreports the fan-out | "records FALSE on the default path" |
| M30 | listing key written into the actor namespace | the two windows share a bucket | "each helper writes its OWN namespace" |
| M31 | gate 1 downgraded to `protectedProcedure` | error code changes; access still refused by gate 2 | "GATE 1: a non-moderator is FORBIDDEN" |
| M32 | gate 2's throw deleted | the belt is gone with no behavioural signal | "STRUCTURAL: both gates are present" |
| M33 | a `rateLimit()` wired onto the proc | an inert limiter that reads as a real one | "does NOT rely on the tRPC rateLimit middleware" |
| M34 | the proc terminator broken | the extractor silently rescans the whole file | "POSITIVE CONTROL: the extractor is BOUNDED" |

**Round-1 re-run**, because the fixes changed the code a round-1 verdict was taken against: **22 defined, 22 verdicts, reconciled, all killed by their own named test** (M3/M13/M19 dropped as superseded by M3b/M30/M31). One entry, M20, first came back **`PATTERN-MISS` rather than a verdict** — prettier had reshifted its indentation, so its search string matched zero times. That is the occurrence assertion doing its job; re-run with the corrected pattern it is killed by its named test. It is recorded here rather than quietly fixed, because a mutant scored without executing is exactly the false-alarm shape this battery exists to avoid.

**Combined: 32 mutants, 32 killed by their own named test, defined == produced in both rounds.** Harness controls re-run each round: unmutated runs GREEN on every target file, byte-copy restore (never git), verdicts counted from the summary line.

### Runs after the revision

| Gate | Result |
|---|---|
| `pnpm test:unit:run` | **22,383 passed / 0 failed / 25 skipped — 1439 files, 0 failing files** |
| `pnpm typecheck` | **OK — 0 type errors** |
| `npx eslint` (changed files) | **0 problems** |
| `npx prettier --check` | clean |

The earlier `feature-flags.early-adopter.seam.test.ts` worktree artifact is resolved (`svelte-kit sync`), so this is the first fully-green full-suite run — no failing files at all.

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260824120000_app_listing_mod_action_message_owner/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260824120000_app_listing_mod_action_message_owner/migration.sql
@@ -1,0 +1,66 @@
+-- ============================================================
+-- Moderator → app-developer messaging: widen the app_listing_moderation_events action CHECK
+-- ============================================================
+-- Adds ONE moderation action:
+--   * 'message-owner' — a moderator sent the app's canonical owner (and, on
+--     request, its accepted collaborators) a free-text message. Unlike every other
+--     action in this taxonomy it changes NO listing state: the row exists to make
+--     the message attributable and reviewable. `reason` carries the SUBJECT and
+--     `detail` the BODY, so the moderation history renders what was actually said;
+--     `after` records the delivery decision (`recipientUserIds`,
+--     `includeCollaborators`), which is the part that cannot be reconstructed later
+--     because the canonical owner can change after the fact.
+--
+-- The current CHECK (20260713120000_w13_post_approval_mod_actions) allows
+-- delist|relist|claim|purge|report-resolve|report-dismiss|reset-to-pending|
+-- owner-unpublish|owner-republish, so `messageAppOwner` would be REJECTED with 23514
+-- without this widen. Postgres cannot modify a CHECK in place -> DROP then ADD.
+-- ADDITIVE: the new IN-list is a strict SUPERSET of the old one, so no existing row
+-- can violate it (nothing to backfill / re-validate).
+--
+-- ⚠️ MANUAL APPLY — per datapacket-talos CLAUDE.md DB rule #8 the main civitai
+-- CNPG nvme0 DB does NOT auto-apply migrations (no prisma migrate deploy). This
+-- file is committed for HISTORY ONLY; a HUMAN applies the SQL below per
+-- environment (psql/retool). CI / deploy do NOT run it. Apply to BOTH:
+--   1. prod nvme0   (the live civitai DB)
+--   2. the dev clone (cnpg-cluster-dev, ns cnpg-database-dev, db civitai)
+--
+-- 🔴 ORDERING: timing-sharp, exactly like the widen it follows. Existing code paths
+-- are unaffected, but the NEW proc's first write fails without it:
+--   * Apply to the DEV CLONE **before** a PR preview exercises `messageAppOwner`, or
+--     the preview 500s on the constraint (preview-DB-drift → smoke-500 trap).
+--   * Apply to PROD nvme0 **before** this ships (main -> release), or the first live
+--     moderator message hits 23514 — and note the failure mode is the GOOD one: the
+--     audit row is written BEFORE delivery, so a missing CHECK widen means the
+--     message is never sent, not that it is sent unrecorded.
+--
+-- The migration-agreement unit test
+-- (src/server/services/blocks/__tests__/app-listing-mod-action.constants.test.ts)
+-- parses the LATEST action-CHECK migration `.sql` (this file, by sorted dir name)
+-- and asserts its IN-list equals the code tuple APP_LISTING_MODERATION_ACTIONS,
+-- catching code/DDL drift at CI time — but it does NOT apply the DDL. The human
+-- apply above is still required.
+--
+-- Idempotent: DROP IF EXISTS then ADD, so a manual re-run is a no-op.
+--
+-- Wrapped in a single transaction so the DROP+ADD swap is ATOMIC: without it there
+-- is a sub-ms window between DROP and ADD where the table has NO action CHECK and a
+-- concurrent bad write could slip through.
+BEGIN;
+ALTER TABLE "app_listing_moderation_events"
+  DROP CONSTRAINT IF EXISTS "app_listing_mod_events_action_check";
+ALTER TABLE "app_listing_moderation_events"
+  ADD  CONSTRAINT "app_listing_mod_events_action_check"
+  CHECK ("action" IN (
+    'delist',
+    'relist',
+    'claim',
+    'purge',
+    'report-resolve',
+    'report-dismiss',
+    'reset-to-pending',
+    'owner-unpublish',
+    'owner-republish',
+    'message-owner'
+  ));
+COMMIT;
--- a/src/components/Apps/appListingModerationView.ts
+++ b/src/components/Apps/appListingModerationView.ts
@@ -104,6 +104,14 @@ const MOD_ACTION_CHIPS: Record<string, Chip> = {
   'reset-to-pending': { label: 'Reset to pending', color: 'yellow' },
   'owner-unpublish': { label: 'Unpublished by you', color: 'gray' },
   'owner-republish': { label: 'Republished by you', color: 'green' },
+  // 🔴 The one action that changed NO listing state — it records that a moderator sent
+  // the owner a free-text message. The label must read correctly in BOTH views this map
+  // serves: the mod queue (`OffsiteReviewQueue`) and the OWNER's own history
+  // (`ownerListingModals`, via `listMyListingModerationEvents`). "Message from
+  // moderation" is true from either side; "Messaged the owner" would be wrong in the
+  // owner's own view, and every `owner-*` label above shows the map is already read
+  // second-person there. Blue, matching `claim` — informational, not a takedown.
+  'message-owner': { label: 'Message from moderation', color: 'blue' },
 };
 
 /** Chip for a moderation-event action, for the per-listing history view. */
--- a/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
+++ b/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
@@ -51,9 +51,7 @@ describe('registration shape', () => {
     // send would succeed, the row would exist, and the developer would see nothing —
     // a silent failure that no test of this module alone can see.
     expect(notificationProcessors).toHaveProperty('app-moderator-message');
-    expect(
-      (notificationProcessors as Record<string, unknown>)['app-moderator-message']
-    ).toBe(def);
+    expect((notificationProcessors as Record<string, unknown>)['app-moderator-message']).toBe(def);
   });
 });
 
--- a/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
+++ b/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
@@ -93,12 +93,20 @@ describe('prepareMessage', () => {
     expect(m!.message).toContain(MODERATION_REPLY_URL);
   });
 
-  it('does NOT name the acting moderator — nothing in details can', () => {
-    // Attribution lives in `AppListingModerationEvent.actorUserId`. Naming the
-    // individual on the wire would make "who acted on my app" answerable by the
-    // developer, which is a retaliation vector the rest of the moderation surface
-    // declines to open. Asserted on the TYPE's own field set, so adding a
-    // `moderatorUsername` field later fails here rather than shipping quietly.
+  it('renders nothing beyond the four details fields', () => {
+    // ⚠️ SCOPE, corrected: this asserts the shape of the LOCAL fixture, not the
+    // exported TYPE. TypeScript types are erased at runtime, so adding a
+    // `moderatorUsername` field to `AppModeratorMessageNotificationDetails` would NOT
+    // fail this — an earlier version of this comment claimed it would, and that claim
+    // was the kind of "reads as coverage while providing none" that stops the next
+    // person looking.
+    //
+    // 🔴 THE REAL GUARD ON "the moderator is not named to the recipient" IS
+    // `app-moderator-message.service.test.ts` → "the moderator is recorded in the audit
+    // row and ABSENT from what the recipient gets", which inspects the details object
+    // the SERVICE actually builds and is confirmed by mutation. What this case is worth
+    // keeping for is narrower: the fixture that feeds the whole-string pin below has
+    // exactly these four fields, so the pinned copy cannot be silently fed a fifth.
     expect(Object.keys(DETAILS).sort()).toEqual(['body', 'listingId', 'slug', 'subject']);
   });
 
--- a/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
+++ b/src/server/notifications/__tests__/app-moderator-message.notifications.test.ts
@@ -1,0 +1,115 @@
+import { describe, expect, it } from 'vitest';
+import {
+  appModeratorMessageNotifications,
+  MODERATION_REPLY_URL,
+  type AppModeratorMessageNotificationDetails,
+} from '~/server/notifications/app-moderator-message.notifications';
+import { notificationProcessors } from '~/server/notifications/utils.notifications';
+import { OWNER_SUBMISSIONS_URL } from '~/server/notifications/app-listing.notifications';
+
+/**
+ * The twelfth app notification type — the moderator's free-text message (pure: no DB,
+ * no client).
+ *
+ * 🔴 THE COPY IS PINNED AS A WHOLE NORMALISED STRING, not by keyword. The artifact
+ * under test IS prose, and a guard on WORDS is walkable by REWORDING: a `toContain`
+ * on "not delivered" stays green through a rewrite that drops the reply route
+ * entirely, or that renders the subject and silently drops the body. A cosmetic reword
+ * will now fail this test — that is the price of a machine-readable claim about what a
+ * developer actually receives, and it is worth paying for the one notification type
+ * whose entire payload is text a human wrote.
+ */
+
+type Def = (typeof appModeratorMessageNotifications)['app-moderator-message'];
+const def = (appModeratorMessageNotifications as Record<string, Def>)['app-moderator-message'];
+
+function msg(details: AppModeratorMessageNotificationDetails) {
+  return def.prepareMessage({ details } as Parameters<Def['prepareMessage']>[0]);
+}
+
+const DETAILS: AppModeratorMessageNotificationDetails = {
+  slug: 'prompt-vault',
+  listingId: 'apl_live',
+  subject: 'Your listing describes a spend confirmation that does not exist',
+  body: 'It says it asks before it spends. It has never shown a confirmation.',
+};
+
+describe('registration shape', () => {
+  it('is a non-toggleable System notification with no prepareQuery', () => {
+    expect(def).toBeTruthy();
+    // Not toggleable, for the same reason the moderation-decision types are not: a
+    // developer must not be able to opt out of being told their app has a problem.
+    expect(def.toggleable).toBe(false);
+    expect(def.category).toBe('System');
+    // Imperative — emitted from the service, never scanned by the scheduled job.
+    expect((def as { prepareQuery?: unknown }).prepareQuery).toBeUndefined();
+  });
+
+  it('🔴 is REGISTERED in the shared processor map', () => {
+    // Without this the type renders as `null` in the drawer: `getNotificationMessage`
+    // looks the type up in `notificationProcessors` and returns null on a miss. The
+    // send would succeed, the row would exist, and the developer would see nothing —
+    // a silent failure that no test of this module alone can see.
+    expect(notificationProcessors).toHaveProperty('app-moderator-message');
+    expect(
+      (notificationProcessors as Record<string, unknown>)['app-moderator-message']
+    ).toBe(def);
+  });
+});
+
+describe('prepareMessage', () => {
+  it('🔴 renders the EXACT string a developer receives', () => {
+    const m = msg(DETAILS);
+    expect(m).toBeTruthy();
+    expect(m!.message).toBe(
+      'Civitai moderation sent you a message about your app "prompt-vault" — ' +
+        'Your listing describes a spend confirmation that does not exist: ' +
+        'It says it asks before it spends. It has never shown a confirmation. ' +
+        '(Replies to this notification are not delivered; contact the team at civitai.com/support.)'
+    );
+  });
+
+  it('carries BOTH halves of what the moderator wrote, verbatim', () => {
+    // Independently of the whole-string pin above, because the failure mode this
+    // guards — rendering the subject and dropping the body — is the one that turns a
+    // message into a notification with no content, and it is a one-line edit.
+    const m = msg(DETAILS);
+    expect(m!.message).toContain(DETAILS.subject);
+    expect(m!.message).toContain(DETAILS.body);
+  });
+
+  it('identifies the app by SLUG, which is never null', () => {
+    // The sibling families fall back to a terse label when `name` is absent. Here the
+    // recipient may own several apps, and one of the two listings that motivated this
+    // feature has an entirely empty listing — i.e. a null name — so a fallback would
+    // fire exactly where identification matters most.
+    const m = msg({ ...DETAILS, slug: 'df-qwen-canvas' });
+    expect(m!.message).toContain('your app "df-qwen-canvas"');
+    expect(m!.message).not.toContain('prompt-vault');
+  });
+
+  it('🔴 always states that replies are not delivered, and where to go instead', () => {
+    // A one-way channel that does not say so strands a developer who wants to answer.
+    const m = msg(DETAILS);
+    expect(m!.message).toContain('not delivered');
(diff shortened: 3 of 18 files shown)


ZacxDev:
## Follow-up: the same denormalised-owner read in a sibling — **corrected target**

I flagged a sibling with this defect during review and **named the wrong one**. Correcting it here so the follow-up points at the right function.

**Not `resetListingToPending`.** That one is **off-site only**, and for an off-site listing `AppListing.userId` *is* the canonical owner — no drift is possible, nothing to fix.

**The affected twin is `resetOnsiteListingToPending`** (`src/server/services/blocks/offsite-moderation.service.ts`), which reads the denormalised column **twice**:

- **line 1344** — `notifyAppListingOwner({ userId: listing.userId, … })`, the `app-listing-reset-to-pending` recipient;
- **line 1281** — `submittedByUserId: listing.userId` on the **cloned publish request**.

The second is arguably the worse of the two: it mis-attributes the re-review in the owner's own submissions queue, and `submittedByUserId` is what the submitter-scoped withdraw gate keys on — so on a drifted row the wrong user is offered the withdraw and the right one is refused it.

`delistListing` reaches the same column through `classifyListingForAction`, which selects `userId` directly rather than resolving it.

**Scope:** same latency as the defence in this PR — **0 drifted rows in production** (21 on-site parents, 21 with a backing block; 0 accepted ownership transfers ever). Latent, not live. It goes live on the first on-site transfer of a listing carrying an in-flight revision.

**Deliberately not fixed here.** Routing `classifyListingForAction` through `resolveListingAccess` changes the recipient of an existing, shipped notification and the attribution of an existing publish-request row — a behaviour change to `delistListing` / `resetOnsiteListingToPending` that wants its own diff, its own drifted-row tests and its own review, not a rider on a new feature.

**Closing condition:** a PR that routes `classifyListingForAction`'s owner field through `resolveCanonicalListingOwner`, with a drifted-onsite-row test per affected call site, merged to `main`. Mechanically checkable — the ledger entry for `offsite-moderation.service.ts` in `app-access.call-site-ledger.test.ts` is where the decision is recorded, so the follow-up either changes that entry or it did not happen.

PR #4418, opened 2026-08-26
fix(app-listings): re-derive media maturity when an owner republishes their own listing

ZacxDev:
## The defect

`republishOwnListing` flips a listing `removed → approved`. It is **entirely owner-driven, with no human in the loop**, and it gated on exactly two things:

- `assertListingAssetsScanCleanInTx` — scan **status** only (Blocked / still-scanning)
- `assertOffsiteListingActionableInTx` — whether an `https` href exists

**Neither inspects media maturity.** Meanwhile a `removed` listing is directly asset-editable (`assertOwnerAssetEditable` refuses only `approved && revisionOfId == null`), and the attach path rejects only `Blocked`/`NotFound` images — **never a `Scanned` but mature one**.

So the sequence `unpublish → attach mature media → republish` put mature store art back on an unchanged declared rating. An under-rated listing then passes `listingMatureFilter(redCapable=false)` (`content_rating NOT IN ('r','x')`) on the store list, and the detail read's `if (!redCapable && isMatureContentRating(row.contentRating)) return null` — i.e. **it is shown to SFW-only users on both surfaces**.

The rating floor helper `resolveListingRatingFloorInTx` already existed and already documents this exact consequence in its own docstring. `republishOwnListing` called it zero times.

Not currently exploited — the path is open and unexercised.

## The fix: one uniform, raise-only floor for both kinds

`republishOwnListing` now runs the declared rating through the existing `resolveListingRatingFloorInTx` and writes the floored result — **identically for on-site and off-site**. No second derivation, no kind branching in the rating path. **RAISE-ONLY**: the helper returns the declared rating unchanged unless the derived ceiling is strictly higher, so tame media can never lower a deliberately higher declaration.

Fail-closed: the derive runs inside the existing transaction and is deliberately **not** caught, so a listing whose rating cannot be derived does not go live.

**One deliberate difference from the previous revision, called out for review:** the floored value is written **unconditionally**, not only when it differs. That is what removes the last conditional, and it is the same shape `approveRequest` uses. The cost is that a republish of an unchanged rating now writes the same value back, so a concurrent write to `contentRating` landing between the in-tx read and the update would be last-writer-wins. The window is inside one transaction on the primary, against a `removed` (not-live) row, and the update is status-guarded so a concurrent republish cannot double-apply — but if a reviewer wants the write narrowed to "only when it changes", that is a one-line change.

### Why on-site is floored too — this is existing precedent, not a new policy

**The on-site go-live path already does exactly this.** `approveRequest`'s draft transition (`publish-request.service.ts:2643-2680` on `origin/release`) reads `abForTransition.contentRating` — the AppBlock's manifest-declared value — passes it through **this same helper**, and writes the floored result to `AppListing`. Its own comment gives precisely the rationale this PR exists for: the attach path "rejects only `Blocked`/`NotFound`", so stamping the declared rating verbatim would "let mature-but-Scanned media go live on a `g`-rated card and pass `listingMatureFilter(redCapable=false)` to SFW-only users."

Two properties make the raise safe on both kinds:

1. **The raise cannot touch the runtime serving gate.** It writes `AppListing.contentRating` (the store card). The runtime value lives on `AppBlock.contentRating` / `AppBlock.manifest.contentRating` — **separate columns**.
2. **A raise persists.** `buildListingScalarSync` explicitly **excludes** `contentRating` from the manifest re-sync, describing it as "(mod override, floored at the derived rating)" — `app-listing-mapper.ts:105-108`, restated at the approve call site in `publish-request.service.ts:2758-2759`. A later version approve does not undo it.

A third existing caller agrees: `reDeriveContentRatingForModLiveEdit` raises the rating on any approved listing **with no kind check at all**.

Point 1 was checked by **enumeration, not sampling** — every `contentRating`/`content_rating` occurrence in `src/` (617 lines / 130 files; 161 / 43 excluding tests), plus the raw-SQL forms (`al.content_rating`, `ab.content_rating`) and all 8 `appBlock.*` write sites:

- Every runtime gate reads the **block**, never the listing: the run-page SSR 404 (`apps/run/[slug]`, via `AppBlock.contentRating`), the token-mint refusal and ceiling claim (`block-tokens/index.ts`, via `AppBlock.manifest.contentRating` and the request host), the per-model slot filter and slot ceiling (`block-registry.service.ts`, via `AppBlock.manifest`), and the AppBlock detail/list SQL (`ab.content_rating`).
- `domainBrowsingCeiling` — the documented single source of truth for a domain's max browsing level — takes a `ColorDomain` and nothing else. No rating of any model can structurally reach it.
- **Nothing ever copies listing → block.** All 8 `appBlock.*` writes take `manifest.contentRating`; there is no raw `UPDATE app_blocks` anywhere. The mirror is one-directional, block → listing.

**What the raise DOES do — worth a reviewer's attention.** It narrows *store visibility*, and only ever narrows it: `listingMatureFilter` hides the card from the SFW store list and `getListingDetail` 404s the store detail page for SFW-only viewers. So an on-site app that declares `g` but attaches `x`-rated store art now becomes invisible on the SFW store **while the app itself keeps serving on `.com`**. That is the intended outcome (the *art* is what is mature, not the app), but it is a real product-visible effect and it is what makes this a decision rather than a pure bugfix.

### The statements that read the other way, and their scope

Four comments assert an unqualified "an on-site listing's rating MIRRORS `AppBlock.content_rating`": the `AppListing.contentRating` schema comment, `app-listing-mapper.ts:208-209` (the create path), `app-listing-backfill.service.ts:45-46`, and `offsite-listing.service.ts` (~2840, "over-rated media is a mod-reject at review, never an auto-raise").

The first three describe the **block → listing mirror**, which is real and unchanged. The fourth is scoped to the **on-site media-revision copy step** — the branch it sits in copies `iconId`/`coverId` and nothing else. Generalising the fourth to every on-site write is what the earlier revision of this PR did, and it contradicts the two existing floors above. Left as an open item below rather than silently rewritten.

## Test coverage

`offsite-moderation.service.mod-actions.test.ts`, **19 cases** in one describe block. **17 are RED at `origin/main` and green at HEAD**; **2 are green at base and are labelled `[INVARIANT GUARD]` in-source** — they pin behaviour this change must not alter and are deliberately *not* counted as regression coverage.

Matrix (`offsite` + `onsite` where marked ×2):

| Case | ×2 kinds | at `origin/main` |
|---|---|---|
| media BELOW declared → floored value **is** the declared rating | ✓ | RED |
| media EQUAL to declared → unchanged | ✓ | RED |
| media ABOVE declared → raised to the derived value | ✓ | RED |
| **PARITY**: both kinds produce a byte-identical flip payload | — | RED |
| a deliberately higher declaration is never lowered by tame media | ✓ | RED |
| a **NULL** declared rating with tame media stays NULL (never stamped `g`) | ✓ | RED |
| no attached assets → declared preserved, no image read, no crash | — | RED |
| absent listing row on the floor read → declared preserved | — | RED |
| screenshots contribute to the derived max, not just icon/cover | — | RED |
| **fail-closed**: a throwing derivation leaves the listing not live | — | RED |
| **removal**: republish files NO `AppListingReport` | ✓ | RED |
| `[INVARIANT GUARD]` a `Blocked` asset still refuses via the pre-existing scan gate | — | green |
| `[INVARIANT GUARD]` `relistListing` (mod, human in the loop) applies no floor | — | green |

Levels are imported from the real `NsfwLevel` enum rather than hand-typed, and the fixtures exercise **R/X**-level art — the two values `listingMatureFilter` actually hides — not the level-2 art that happens to be live today. Asset ids and sibling levels are pairwise distinct.

### Three scaffold defects found while reworking the matrix

Each of these made a test pass for the wrong reason; all are fixed here.

1. **The actionability gate was never running.** The fixture for the third `appListing.findUnique` lacked `kind`/`externalUrl`, so `checkOffsiteListingActionable` saw `kind: undefined` and returned `skipped`. The floor was only ever reached past a gate that had disarmed itself. The fixture now carries all four columns.
2. **A scoping assertion was satisfied by a neighbouring call.** `expect(appListingScreenshot.findMany).toHaveBeenCalledWith({ where: { appListingId: APP_ID } })` is true because the *scan gate* makes that call one step earlier — so a mutant binding the floor to the wrong listing id **survived**. Now asserted over every call, and the mutant dies.
3. **The removal guard asserted the payload first.** Behind the payload assertion it went red at the pre-change tip for the wrong reason (the missing raise) and the removal itself stayed unmeasured. `create` is now asserted first.

### Mutation battery — 16 mutants, 16 killed, 0 survivors

Re-run from scratch for this design (a change round resets the gate), with a negative control (unmutated: 102/102 green) and every mutant's kill attributed to a named test. Enumerated from semantic failure modes rather than from deletion alone: deletion, comment-out with the token retained, wrong bind ×4 (declared-instead-of-floored, `null`, `listing.kind`, wrong listing id), branch re-introduction, branch inversion, fail-open, **removal regression** (the on-site queue re-added — the positive control for the removal guard), comparison-direction flip, boundary `>` → `>=`, operand swap, raise/lower inversion, drop-screenshots, drop-the-empty-guard.

**Two rounds were needed, and both findings are worth stating plainly rather than folding into a clean number:**

- Round 1 had **one survivor** — the wrong-listing-id bind (scaffold defect 2 above).
- Round 1's `>` → `>=` mutant was **killed only by a pre-existing test elsewhere in this file**, i.e. it survived *this PR's* coverage. Every case in the block ties nowhere on the ladder, and `>=` only differs on a tie. The NULL-declared case was added specifically to produce one (`nsfwLevelFromContentRating` maps both `null` and `'g'` to the SFW floor), and it now kills that mutant in-block.
- Mutants against `resolveListingRatingFloorInTx` itself are included because this change depends on its raise-only semantics; the helper is pre-existing and separately tested.

## Scope — deliberately narrow

`relistListing` and the on-site reset re-approve both have a **moderator in the loop**, which is why this targets `republishOwnListing` only. An invariant guard pins that `relistListing` still applies no floor.

## Adjacent items (filed, not fixed)

- **Two live listings sit above their declared rating** (`panorama-360`, `prompt-vault` — both `g` with level-2 art). This is a data question, not this code path. *Reported in the task brief; not independently verified here — I have no production data access.* **Closing condition:** a moderator re-rates both from `/apps/review` → Manage listings, or confirms level-2 art is acceptable at `g`; verified by re-reading `content_rating` for those two slugs.
- **Four comments now read narrower than the code** (schema `AppListing.contentRating`, `app-listing-mapper.ts:208-209`, `app-listing-backfill.service.ts:45-46`, `offsite-listing.service.ts` ~2840). `approveRequest`, `reDeriveContentRatingForModLiveEdit` and now `republishOwnListing` all floor an on-site listing's rating. **Closing condition:** a reviewer on this PR decides which reading is canonical and either updates those comments or removes the on-site floors; the comment edit lands in whichever PR they pick. Not done here because the answer changes which code is wrong, not just which sentence.
- **`PlatformDefaultBlock.minContentRating` / `maxContentRating` have zero readers in `src/`** (surfaced by the enumeration above). Dead columns — noted so nobody assumes a third maturity gate exists. **Closing condition:** someone confirms they are unused and drops them, or names the consumer; a `git grep` returning only the schema closes it either way.
- **Do the sibling flips want the same floor?** `relistListing` and the on-site reset re-approve are unfloored today. Both have a human in the loop, so neither is unsafe — but the asymmetry is now the only kind-shaped thing left in this area. **Closing condition:** a reviewer on this PR says yes or no; a yes becomes a follow-up PR touching those two call sites, a no is recorded in this thread and closes it.

## 🔴 Sequencing — BLOCKED on #4413

**#4413 (`zach/app-listing-owner-unpublish-editable`, unmerged) edits `republishOwnListing` in this same file** — 21 hits, and it extracts the last-event guard into a new `app-listing-owner-unpublish.ts`. This PR is branched directly on `main` (not stacked) and must be **rebased onto updated `main` and test-merged after #4413 lands.**

**A clean git merge is not a clean merge.** Both sides edit different parts of the same function, so a *semantic* conflict survives a textless merge. **The merged body of `republishOwnListing` needs reading by a human** before this is taken off draft.

The two are complementary rather than redundant: #4413 blocks the owner from *lowering the declared rating* on a removed listing, and its own docstring says *"Neither republish gate is a content review … so the refusal — not the republish path — is the guarantee."* This PR closes the other half — *raising the media maturity past* an untouched declared rating — and makes the republish path itself a guarantee.

### DIFF
--- a/src/server/services/blocks/__tests__/offsite-moderation.service.mod-actions.test.ts
+++ b/src/server/services/blocks/__tests__/offsite-moderation.service.mod-actions.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { TRPCError } from '@trpc/server';
 
-import { APP_LISTING_REPORT_REASONS } from '~/server/schema/blocks/offsite-moderation.schema';
+import { NsfwLevel } from '~/server/common/enums';
 
 import {
   OffsiteModerationError,
@@ -44,9 +44,8 @@ type WriteMock = {
   };
   appListingReport: {
     updateMany: ReturnType<typeof vi.fn>;
-    // The on-site over-rated-media review queue: republish check-then-inserts an
-    // advisory `AppListingReport` in the SAME tx as the status flip.
-    findFirst: ReturnType<typeof vi.fn>;
+    // Spied ONLY so the removal guard can assert republish files no advisory report —
+    // an earlier revision of the rating change queued one for over-rated on-site media.
     create: ReturnType<typeof vi.fn>;
   };
   // claim validates the target owner on the primary inside the tx.
@@ -91,8 +90,6 @@ const { mockRead, mockWrite, mockNotify, mockLogToAxiom, ids } = vi.hoisted(() =
     },
     appListingReport: {
       updateMany: vi.fn(async () => ({ count: 1 })),
-      // Default: no OPEN review for this listing+reporter → the queue write proceeds.
-      findFirst: vi.fn(async (): Promise<{ id: string } | null> => null),
       create: vi.fn(async (a: { data: unknown }) => a.data),
     },
     user: { findUnique: vi.fn(async () => ({ id: 1 })) },
@@ -1787,7 +1784,7 @@ describe('OffsiteModerationError (PR3/PR4 + W13 codes)', () => {
 });
 
 // ---------------------------------------------------------------------------
-// 🔴 republishOwnListing — GO-LIVE MATURITY GATE (kind-aware).
+// 🔴 republishOwnListing — GO-LIVE RATING FLOOR (UNIFORM, both kinds).
 //
 // The hole this closes: republish is an OWNER-driven go-live with NO human in the
 // loop, and a `removed` listing is directly asset-editable. The attach path rejects
@@ -1796,19 +1793,15 @@ describe('OffsiteModerationError (PR3/PR4 + W13 codes)', () => {
 // republish` put mature store art back under an unchanged declared rating, which
 // passes `listingMatureFilter(redCapable=false)` (`content_rating NOT IN ('r','x')`).
 //
-// KIND-AWARE by design:
-//   OFF-SITE → RAISE the stored rating to the media-derived one (raise-only).
-//   ON-SITE  → do NOT raise (the rating mirrors `AppBlock.content_rating`); allow the
-//              republish and queue an advisory `AppListingReport` for a moderator.
+// UNIFORM BY DESIGN — the floor is applied identically to BOTH kinds, matching the
+// on-site draft go-live (`publish-request.service.ts`, "🔴 RATING FLOOR (go-live)")
+// which already runs the manifest-declared rating through THIS SAME helper. The raise
+// targets `AppListing.contentRating` (the store card); the runtime serving gate reads
+// `AppBlock.contentRating`, a separate column.
 //
-// 🔴 COVERAGE HONESTY. Of the 15 cases below, SEVEN are RED at `origin/main` and green
-// at HEAD — those are the regression coverage:
-//   off-site ABOVE · on-site ABOVE (+ its CHECK-set case) · screenshots-contribute ·
-//   idempotence · fail-closed · transactional.
-// The other EIGHT are green at `origin/main` too and are labelled `[INVARIANT GUARD]`:
-// they pin behaviour this change must NOT alter (raise-only never lowers, no-assets,
-// absent row, the pre-existing scan gate, the mod relist path). They are preservation
-// guards and are deliberately NOT counted as regression coverage.
+// 🔴 COVERAGE HONESTY. Cases labelled `[INVARIANT GUARD]` are green at `origin/main`
+// too — they pin behaviour this change must NOT alter and are deliberately NOT counted
+// as regression coverage. Everything else is RED at `origin/main`.
 //
 // Levels are the REAL enum (`src/server/common/enums.ts`): PG=1, PG13=2, R=4, X=8.
 // The ladder is `['g','pg','pg13','r','x']`; `deriveContentRatingFromAssets` picks the
@@ -1816,29 +1809,31 @@ describe('OffsiteModerationError (PR3/PR4 + W13 codes)', () => {
 // exactly the two values `listingMatureFilter` hides, which is why the fixtures use
 // them rather than the level-2 art that happens to be live today.
 // ---------------------------------------------------------------------------
-describe('republishOwnListing — go-live MATURITY gate (kind-aware)', () => {
+describe('republishOwnListing — go-live RATING FLOOR (uniform, both kinds)', () => {
   /** Pairwise-distinct asset ids, so a wrong-bind mutant cannot alias two of them. */
   const ICON_ID = 101;
   const COVER_ID = 202;
   const SHOT_ID = 303;
-  /** Real `NsfwLevel` members — not magic numbers. */
-  const LVL_PG = 1;
-  const LVL_PG13 = 2;
-  const LVL_R = 4;
-  const LVL_X = 8;
+  /**
+   * DERIVED from the enum, not hand-typed. A local `const LVL_R = 4` is a guard spelled
+   * rather than structural: it keeps passing after the enum's bits move, and the test
+   * would then be asserting a ladder position that no longer exists.
+   * `~/server/common/enums` has ZERO imports of its own, so this pulls no module graph.
+   */
+  const LVL_NONE = 0;
+  const LVL_PG = NsfwLevel.PG;
+  const LVL_PG13 = NsfwLevel.PG13;
+  const LVL_R = NsfwLevel.R;
+  const LVL_X = NsfwLevel.X;
 
-  function ownerPrimary(opts: {
-    kind: string;
-    contentRating: string | null;
-    appBlockId?: string | null;
-  }) {
+  function ownerPrimary(opts: { kind: string; contentRating: string | null }) {
     return {
       userId: OWNER,
       status: 'removed',
       kind: opts.kind,
       slug: SLUG,
       name: 'Cool App',
-      appBlockId: opts.appBlockId ?? (opts.kind === 'onsite' ? BLOCK_ID : null),
+      appBlockId: opts.kind === 'onsite' ? BLOCK_ID : null,
       contentRating: opts.contentRating,
       externalUrl: opts.kind === 'offsite' ? EXTERNAL_URL : null,
       connectClientId: null,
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Verification detail

### Red-at-base / green-at-HEAD matrix

Method: the **test file stays at HEAD**; only the service under test is reverted to `origin/main`, then restored and checksum-verified.

```
git show origin/main:src/server/services/blocks/offsite-moderation.service.ts > <that path>
pnpm vitest run src/server/services/blocks/__tests__/offsite-moderation.service.mod-actions.test.ts --reporter=verbose
# restore, then sha256sum-verify the restore
```

| Test | at `origin/main` | at HEAD |
|---|---|---|
| off-site, media ABOVE → raised | **RED** | green |
| on-site, media ABOVE → unchanged + queued | **RED** | green |
| on-site over-rated: reason ∈ CHECK set | **RED** | green |
| screenshots contribute to derived max | **RED** | green |
| idempotence: no duplicate queue entry | **RED** | green |
| fail-closed: throwing derive ⇒ not live | **RED** | green |
| transactional: failing queue write aborts | **RED** | green |
| *[INVARIANT GUARD]* off-site BELOW | green | green |
| *[INVARIANT GUARD]* off-site EQUAL | green | green |
| *[INVARIANT GUARD]* on-site BELOW | green | green |
| *[INVARIANT GUARD]* on-site EQUAL | green | green |
| *[INVARIANT GUARD]* no attached assets | green | green |
| *[INVARIANT GUARD]* absent listing row | green | green |
| *[INVARIANT GUARD]* Blocked asset still refused by the pre-existing scan gate | green | green |
| *[INVARIANT GUARD]* `relistListing` unchanged | green | green |

Counted result at base: **7 failed / 91 passed (98)**. At HEAD: **98 passed**.
The 8 invariant guards are labelled as such in the source and are **not** counted as regression coverage.

### Mutation battery — 13/13 killed, 0 survivors

Whole file run per mutant (no `-t` filter that could exclude the killing test); verdicts read from the counted `Tests N failed | M passed` line, never an exit code; each replacement asserted to apply exactly once; pristine file restored and checksum-verified.

| # | Mutant | Verdict | Killed by |
|---|---|---|---|
| PC | **positive control** — gate neutered (`mediaOverRated = false`) | KILLED (6) | all 6 gate tests |
| M1 | kind-branch inversion (raise): `!isOnsite` → `isOnsite` | KILLED (3) | off-site ABOVE; on-site ABOVE; screenshots |
| M2 | kind-branch inversion (queue): `isOnsite` → `!isOnsite` | KILLED (5) | both ABOVE cases; CHECK-set; idempotence; transactional |
| M3 | comment-out the raise (token retained) | KILLED (2) | off-site ABOVE; screenshots |
| M4 | branch inversion `!==` → `===` | KILLED (12) | all BELOW/EQUAL guards + ABOVE cases |
| M5 | wrong bind: raise writes declared, not derived | KILLED (2) | off-site ABOVE; screenshots |
| M6 | wrong bind: floor called with `null` | KILLED (6) | 2 pre-existing + 4 guards |
| M7 | fail-open: swallow a throwing derive | KILLED (1) | **fail-closed** |
| M8 | drop the idempotence check | KILLED (1) | **idempotence** |
| M9 | dedup probe loses its `status` scope | KILLED (1) | **idempotence** |
| M10 | queue reason `inappropriate` → `spam` | KILLED (1) | **on-site ABOVE** |
| M11 | queue status `pending` → `resolved` | KILLED (1) | **on-site ABOVE** |
| M12 | details wrong-bind: derived/declared swapped | KILLED (1) | **on-site ABOVE** |

**Survivors: none.** Attribution was verified per-mutant (every kill traced to the test targeting that guard, not a neighbouring guard's error). Note M10 is killed by the exact-`reason` assertion, *not* by the CHECK-set membership guard — `spam` is a legal member, so that guard alone would not have caught it.

**Not a real mutant:** swapping `reporterUserId: userId` → `listing.userId` is semantically equivalent here, because `loadOwnedListingInTx` already enforces `listing.userId === userId`. Recorded rather than counted as a kill.

### Other checks

- `pnpm typecheck` — **OK, 0 type errors** (103s, heap cap 8192 MB).
- `pnpm vitest run src/server/services/blocks/__tests__/` — **150 files, 3464 tests, all passing**.
- `pnpm test:lint-rules` (architectural guards incl. `no-io-in-transaction`) — **23 files, 355 tests, all passing**.
- `eslint` on both changed files — **2 files linted, 0 errors**; the single warning (`_a` unused) is pre-existing at `origin/main`.

ZacxDev:
### Full-suite result + one more mutant

**`pnpm test:unit:run` (whole unit project): 1455 files passed, 1 skipped (1456) · 22,648 tests passed, 25 skipped (22,673) · 0 failed.**

Follow-up commit `9000b5166c` tightens the displacement test. It previously asserted only `.rejects.toThrow()` — which passes just as well if the *new* maturity code throws first, so it could not actually observe the displacement it exists to rule out. It now asserts the scan gate's own `BAD_REQUEST` / `"blocked"` error.

| # | Mutant | Verdict | Detail |
|---|---|---|---|
| M13 | remove the scan-gate call from `republishOwnListing` | **KILLED** | 4 failed / 94 passed |

That is the control proving the tightened assertion is non-vacuous: with the scan gate gone the maturity path resolves the republish instead, and the test now notices.

Running mutation total: **14 mutants, 14 killed, 0 survivors.**

PR #4425, opened 2026-08-26
fix(remix-gallery): cast the Availability parameter that is 500ing every gallery read (42883)

ZacxDev:
## The bug

Every Remix Gallery read is returning a 500. Not intermittently and not data-dependent — the SQL is rejected at parse time:

```
Invalid `prisma.$queryRaw()` invocation:
Raw query failed. Code: `42883`. Message: `ERROR: operator does not exist: "Availability" <> text
HINT: No operator matches the given name and argument types. You might need to add explicit type casts.`
```

The cause is one clause:

```ts
AND p."availability" != ${Availability.Private}
```

Inside a Prisma tagged template, `${Availability.Private}` is bound as a **`text`** parameter. Postgres has no `"Availability" <> text` operator, so it refuses the whole statement. This is a type error in a *static* predicate, so it does not filter fewer rows — nothing is returned at all, on every call, for every viewer.

The clause lives in `entryIsVisible`, the shared predicate that `getRemixGallery`, `galleryHasEntries` and `getRemixGalleryCardSummaries` compose **unconditionally**. That is why several routes fail identically. `getMyRemixGallerySubmissions` carries a second, inline copy of the same clause and fails the same way.

Affected tRPC procedures, by error volume:

| Procedure | Share of the errors |
|---|---|
| `placement.getRemixGalleryCardSummaries` | 99.5% combined |
| `placement.getRemixGalleryVisibility` | ↑ |
| `placement.getRemixGallery` | remainder |
| `placement.getMyRemixGallerySubmissions` | remainder |

Introduced by 9214e2235e — *feat(remix-gallery): make remixed images discoverable in the feed (#4364)*. High-volume in production for roughly 19 hours.

**No privacy exposure.** The clause it broke is a privacy guard: it stops a gallery returning the URL of an image whose owner marked the post non-public. Its failure mode is **fail-closed** — the query throws rather than serving the row. The cost is that the feature is down, not that anything leaked.

## The fix

`::"Availability"`, which is the convention already used elsewhere in this repo — `src/server/meilisearch/cleanup.ts`, `src/server/search-index/models.search-index.ts`, `src/server/services/model-version.service.ts`.

### Changed

Two casts, both in `src/server/services/remix-gallery.service.ts`, plus the tests:

| File | What |
|---|---|
| `src/server/services/remix-gallery.service.ts` | the clause in `entryIsVisible` (the shared fragment behind three queries) — **this is the fix for the outage** |
| `src/server/services/remix-gallery.service.ts` | the second inline copy, in `getMyRemixGallerySubmissions` |
| `src/server/services/__tests__/remix-gallery.service.test.ts` | the regression guard (below) |

### Same-class hits deliberately left alone

| Site | Why |
|---|---|
| `src/server/services/image.service.ts` (`getAllImages`) | executed through the **node-postgres** pool (`queryWithTimeout(imageDb, …)`), which sends parameters **untyped** and lets Postgres infer the enum — so the uncast form has worked there since March and never threw. See the note below. |
| `src/components/AutocompleteSearch/AutocompleteSearch.tsx` | Meilisearch filter string, not SQL |
| `src/pages/search/models.tsx` | Meilisearch filter string, not SQL |
| `src/server/services/image.service.ts` (`getImagesFromSearch` filters) | Meilisearch filter string, not SQL |
| `src/server/services/image.service.ts` (`getImage`) | compares against an untyped SQL **literal** (`!= 'Private'`), which Postgres coerces to the enum on its own — no bound parameter, no `text` |

`cleanup.ts`, `models.search-index.ts` and `model-version.service.ts` already carry the cast.

#### Why `getAllImages` is not in this PR

An earlier revision of this PR *did* cast `getAllImages` as a defensive third change. **It has been dropped**, and the file is now byte-identical to `origin/main`. Three reasons, in order of weight:

1. **It fixes nothing live.** Prisma's `$queryRaw` is what binds the parameter as `text`; the node-postgres pool is the discriminator between the two sites, not the SQL. That fragment has never thrown 42883.
2. **It broke a gate.** The explanatory comment was a net **+8 lines**, and `src/server/flipt/__tests__/flipt-eval-context.test.ts` keys its `ENTITY_WITHOUT_CONTEXT_LEDGER` by `file:line`. Four `image.service.ts` entries shifted (`4123→4131`, `4264→4272`, `5002→5010`, `6136→6144`), failing a hard-coded assertion. Fixing *that* by editing the ledger would have been a line-number churn commit in exchange for a change that fixes nothing.
3. **It silently changed a cache key.** The `getAllImages` Redis cache key is derived from the query, so the cast would have invalidated it for no behavioural gain.

The residual risk the cast was aimed at is real but small and unchanged from before #4364: the commented-out `dbRead.$queryRaw(query)` a few lines below is one edit away from arming the defect on the busiest feed query in the app. That is a comment-level hazard on a dormant code path, not something to fix inside an outage PR.

### On the two non-remix procedures in the error log

I checked the two procedures the error log attributes a share of these 42883s to but which are not remix-gallery routes — `model.getActiveSales` and `image.getGenerationData`. **Neither code path contains an `Availability` comparison at all**, cast or otherwise: `getActiveSalesForModels` → `querySalesForModels` compares only `mv.status`, and it already casts (`'Published'::"ModelStatus"`); `getCapTiers` → `getCapTier` has no enum comparison; `getImageGenerationData` has no raw availability predicate. Both are called from components that also mount the remix-gallery queries (`ModelCardContext`, the image detail view), so the most likely explanation is attribution at the batched-request level rather than a second defect — but I could not confirm that from the code, so it is flagged rather than asserted.

## Coverage

Six new cases in `src/server/services/__tests__/remix-gallery.service.test.ts`. No database is reachable from that suite, so the assertion is on the SQL the service **composes** — it proves the cast is written, not that Postgres accepts the statement.

Four cases drive each of the four affected reads and assert that **every** bound `Availability` parameter is immediately followed by `::"Availability"`. That is a walk over the parameter list, not a search for the cast somewhere in the query — the latter passes while a *second* comparison in the same statement is still uncast, which is precisely this defect's shape (three queries, one shared broken fragment).

Two more cases pin the whole normalised clause, together with both of its neighbouring `AND`s — one for the shared `entryIsVisible` fragment, one for the inline copy in `getMyRemixGallerySubmissions`. A flipped `AND`→`OR` makes the entire `WHERE` permissive while every fragment-shaped assertion stays green, so the joins have to be pinned too, and the two clauses need separate pins because the second is written out inline rather than composed.

Three things the guard needed in order not to be vacuous:

- **A positive control per route.** A read that stopped comparing availability at all binds nothing, and the loop would then assert over an empty list and pass. Each case asserts at least one binding was found, naming the route in the failure. The whole-clause pin on `getMyRemixGallerySubmissions` selects its read by *which one binds `Availability.Private`* and asserts exactly one match, so a read that went quiet fails as 0 rather than passing.
- **Comments stripped before any assertion reads the SQL — both `--` and block comments.** The `--` half matters because the fix adds a comment beside the clause that *explains* the cast, so a naive substring check would be satisfied by the prose. The block-comment half matters more, and was **a real hole found by mutation testing rather than by review**: a clause wrapped in a block comment is dead SQL whose text and bound parameter both survive, so every per-parameter assertion still saw a correctly cast `Availability` while Postgres would be handed a `WHERE` with the private-post guard missing. That mutant passed all 141 tests of the previous revision. Both forms are now stripped in one left-to-right alternation, so whichever opener comes first wins, the way a SQL lexer reads it.
- **A failure message that names the defect.** A parameter bound with no placeholder anywhere in the rendered statement previously reached `toMatch(null)`, which throws `expects to receive a string, but got object` — it fails, so nothing passed vacuously, but it reads as a broken harness. That case now has its own assertion saying the clause is commented out.

### Red/green matrix

Same command both times:

```
node scripts/test-unit-run.mjs src/server/services/__tests__/remix-gallery.service.test.ts
```

| Tree | Result |
|---|---|
| tests applied, `remix-gallery.service.ts` reverted to the merge base | **6 failed** / 136 passed (142) — all six, each naming the uncast placeholder and the text that follows it |
| HEAD (fix applied) | **142 passed** (142) |

The pre-existing 136 pass in both arms, so the six are attributable to the change and nothing else moved.

### Mutation control

Run in a scratch copy off this tree, not in the branch:

| Mutant | Result |
|---|---|
| block-comment the `getMyRemixGallerySubmissions` clause (text and binding both survive, clause dead) | **survived** 141/141 before this revision; now **fails 2**, both with their own authored assertions |
| block-comment the shared `entryIsVisible` clause | **fails 4** — every route that composes the fragment |
| drop the cast (positive control) | **fails** with `is bound as text; append ::"Availability" or Postgres rejects the statement with 42883`, before and after |

Also green: `remix-gallery-entry-visibility.test.ts` (7) and `src/server/flipt/__tests__/flipt-eval-context.test.ts` (12/12 — this was 3 failed / 9 passed on the previous revision, which is what dropping the `image.service.ts` hunk fixed).

## Not verified

- **Against a real Postgres.** The tests prove the cast is composed into the statement, not that Postgres accepts it. The cast form itself is copied verbatim from three queries already running in production.
- **Typecheck is not clean, and not because of this change.** `node scripts/typecheck.mjs` reports 14 errors, all in `src/server/services/user-hub.service.ts` and `src/components/Hubs/*`, all of the form `Property 'userHubFollow' does not exist on PrismaClient`. `model UserHubFollow` was added to the schema in b2727ecf72 (#4389) and the generated client available here predates it. **Zero** errors name either file in this PR.
- **Two `@typescript-eslint/no-explicit-any` errors** at `remix-gallery.service.test.ts:109` are pre-existing — byte-identical at the merge base, and outside both of this PR's hunks (which start at lines 149 and 2402).

### DIFF
--- a/src/server/services/__tests__/remix-gallery.service.test.ts
+++ b/src/server/services/__tests__/remix-gallery.service.test.ts
@@ -149,6 +149,7 @@ const {
   setRemixGalleryPins,
   parseGalleryCursor,
   getRemixGallery,
+  getRemixGalleryCardSummaries,
   getRemixGalleryVisibility,
   getRemixGalleryFreeEligibility,
   getMyRemixGallerySubmissions,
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correction — two things in the original description were wrong

Flagging these in the thread rather than only editing the body, since the original has already been readable for a while. Pushed as 38ad2cf8bf on top of 05ac1be9ac; nothing was amended or force-pushed.

### 1. The `image.service.ts` hunk is dropped

It was presented as a defensive third change worth taking. It was not worth taking, for three reasons, and the file is now byte-identical to `origin/main`:

- **It fixes nothing live.** `getAllImages` runs that fragment through `queryWithTimeout(imageDb, …)` on a node-postgres pool, which sends parameters untyped and lets Postgres infer the enum. It has never thrown 42883. The description said this, but then took the change anyway.
- **It broke a gate, and that is the deploy-blocking part.** `src/server/flipt/__tests__/flipt-eval-context.test.ts` keys its `ENTITY_WITHOUT_CONTEXT_LEDGER` by `file:line`. The ~9-line explanatory comment was a net **+8 lines**, shifting four `image.service.ts` entries (`4123→4131`, `4264→4272`, `5002→5010`, `6136→6144`) and failing a hard-coded assertion. That is the `Unit tests (4)` failure on the previous revision — 3 failed / 9 passed of 12. It is now 12/12, fixed by dropping the hunk rather than by editing the ledger, which is the right direction: the ledger was correct and the change that moved it was not carrying its weight.
- **It silently changed a Redis cache key.** The `getAllImages` cache key derives from the query, so the cast would have invalidated it for no behavioural gain.

The residual hazard the cast was aimed at is real but unchanged from before #4364 — the commented-out `dbRead.$queryRaw(query)` below it is one edit away from arming the defect on the busiest feed query in the app. That is a dormant-path comment hazard, not something to fix inside an outage PR.

### 2. The "six `getAllImages` suites" coverage claim was inaccurate

The original body listed six suites as *covering the `image.service.ts` edit*. They do not cover it. Re-measured on this push, two arms:

| `getAllImages` availability clause | `image-db-unpublished-authorization`, `image-unpublished-authorization`, `image-search-published-only`, `image-search-scheduled-owner`, `image-hide-challenges-exclusion`, `image-feed-clickhouse-failsoft` |
|---|---|
| uncast (i.e. current `origin/main`) | **45/45 passed** (6 files) |
| clause **deleted outright** | **45/45 passed** (6 files) |

They exercise `getAllImages` but are insensitive to that clause — deleting a privacy predicate does not move a single assertion. Listing them as coverage was the more misleading of the two errors, because it reads as evidence for a change that had none. The claim is gone from the body along with the hunk.

### Also in this push: a real hole in the new guard, found by mutation testing

Not a correction to the description, but it belongs here because it changes how much the guard is worth.

`render()` stripped `--` line comments before asserting on the generated SQL, but not block comments. A clause wrapped in a block comment is **dead SQL whose text and bound parameter both survive** — so every per-parameter assertion still saw a correctly cast `Availability` while Postgres would be handed a `WHERE` with the private-post guard missing. That mutant on the `getMyRemixGallerySubmissions` clause **passed all 141 tests** of the previous revision.

Fixed three ways:

- `render()` now strips both comment forms in one left-to-right alternation, so whichever opener comes first wins, the way a SQL lexer reads it.
- `getMyRemixGallerySubmissions` gets its own whole-clause pin. It writes the predicate out inline instead of composing `entryIsVisible`, so the existing pin could not reach it — which is exactly why its mutant survived while the equivalent one on the shared fragment was caught.
- A parameter bound with no placeholder anywhere in the statement previously hit `toMatch(null)` and threw `expects to receive a string, but got object`. It failed, so nothing was passing vacuously, but it read as a broken harness instead of naming the defect. It now has its own assertion.

Mutation control, in a scratch copy rather than the branch:

| Mutant | Before | After |
|---|---|---|
| block-comment the `getMyRemixGallerySubmissions` clause | **survived** 141/141 | **fails 2**, both with their own authored assertions |
| block-comment the shared `entryIsVisible` clause | caught | **fails 4** — every route composing the fragment |
| drop the cast (positive control) | caught | caught, same message |

Suite is 142/142 (was 141), and the merge-base arm is 6 failed / 136 passed, so all six new cases are red without the fix.

### Two comments that claimed more than the code did

- The `isSql` rationale said `@prisma/client` may resolve to a stub where `Prisma.Sql` is `undefined`. This suite mocks neither `@prisma/client` nor the Prisma namespace, and takes `Availability` from `~/shared/utils/prisma/enums` — so that rationale is simply false here. The real reason is that the recursion only ever reads `.strings` and `.values`.
- `Availability` parameters are identified **by value**, because a Prisma tagged template hands the test bare scalars with no type information. That is deliberately over-inclusive — a future unrelated parameter whose value happens to be `'Public'` would be required to carry the cast too — and is now documented as such rather than left to be discovered.

### Nit not taken

`render()` reimplements Prisma's parameter numbering and could simply *be* `Prisma.sql(...)`, dropping ~20 lines that can drift. I left it: swapping the instrument would invalidate the mutation evidence above and require re-running every arm, in exchange for code-golf on a hot fix. Worth doing as a follow-up if someone touches this helper.

PR #4430, opened 2026-08-26
fix(comments): stop ToS-removed v2 comments from being counted

manuelurenah:
Closes the residue of [CU 868kwmf0r](https://app.clickup.com/t/868kwmf0r).

## What the report said, and what is actually left

The task reports that "Remove as ToS" on a v2 comment (image / post / article / review / bounty) sets `tosViolation = true` but leaves the comment visible to everyone, and names `commentsv2.service.ts` lines 467 / 507 / 537 as the read path with no `tosViolation` check.

**That root cause is already fixed and live.** `229d6d14a1` added the filter to every v2 read path, and it first reached production in the `1b6a1379f` deploy at **2026-08-25 21:27 UTC** — after the report was filed on 2026-08-25. The line numbers in the report no longer resolve. Verified on `main`: the raw pagination query, pinned comments, the deep-link target, reply threads, `getComment` and `getCommentsThreadDetails2` all carry the predicate, and the controller forwards `ctx.user?.isModerator ?? false`.

What survives is the **counts**, which is why the affordance still looks like the comment is there.

## The bug this PR fixes

`Thread.commentCount` is maintained by an INSERT/DELETE trigger:

```
comment_count_update AFTER INSERT OR DELETE ON "CommentV2" FOR EACH ROW EXECUTE FUNCTION update_thread_comment_count()
```

A ToS flag is an `UPDATE`, so it never decrements the counter. `getCommentCount` returned that counter verbatim, and `Comment.tsx` uses `commentv2.getCount` to render the per-comment reply count — so a removed reply keeps its "show N replies" button, which expands to fewer comments than it promises or to nothing. The reply-thread batch had the same problem from the other direction: its recursive CTE carries `Thread.commentCount`, and that value seeds the client's cache as `initialData`.

Production today: **113 flagged v2 comments across 86 threads.** Thread 5484423 counts 8 and can show 1.

## Changes

- `getCommentCount` counts `CommentV2` rows under the same `tosViolation` predicate the reads use, instead of reading `Thread.commentCount`. It reaches the thread through the relation filter, so it stays a single query.
- `getCommentCountV2Handler` forwards `isModerator`, so the moderation queue still sees the full count.
- `getReplyThreads` derives each thread's `commentCount` from a filtered `groupBy` and `groupReplyThreads` uses it, so the seeded reply count agrees with what `getCount` would answer. Deliberately no `excludedUserIds` exclusion on that count — it replaces a raw row counter, and adding one would desync it from `getCommentCount`.
- Entity metrics: `model.metrics.ts` already had `WHERE c."tosViolation" = false`; its CommentV2 counterparts did not. Added to `post.metrics`, `post.metrics-old` (still reachable behind the `simplified-post-metrics` flag), `article.metrics`, `bounty.metrics` and `model3d.metrics`. Left `question.metrics` / `answer.metrics` alone — that processor is commented out in `update-metrics.ts`.

## Performance

`getCommentCount` goes from a `Thread` PK lookup to a relation-filtered count over `CommentV2` — **still one round trip**, planned as a nested loop over `Thread_<entity>Id_key` and `CommentV2_threadId`. Measured on production:

| | plan | time | buffers | round trips |
|---|---|---|---|---|
| `Thread.commentCount` (before) | `Index Scan Thread_imageId_key` | 0.123 ms | 4 | 1 |
| relation filter (this PR) | `Nested Loop` over both indexes | 0.079 ms | 12 | 1 |

Do **not** rewrite it as a filtered Prisma `_count` on `Thread`. That compiles to an uncorrelated aggregate subquery: `Seq Scan` over all 2.27M `CommentV2` rows plus a 1.86M-group `HashAggregate`, **1.7 s and 71K buffers per call**. There is a comment on the function saying so.

Every thread this can be called on is small anyway — the three call sites are reply threads (`entityType: 'comment'`), questions and answers, whose maxima on production are **77 / 6 / 8** comments. The 1,000+ comment threads are all articles, which render through `getThreadDetails` — a different, untouched procedure.

`getReplyThreads` does gain one query: a third `groupBy` for the per-thread visible count. It runs inside the existing `Promise.all`, so it costs a connection but no added latency, and it cannot be merged into the hidden-count `groupBy` — that one excludes blocked/hidden users to match what the modal shows, while this one replaces a raw row counter and must not, or it would disagree with the `getCommentCount` it seeds.

## Known limitation

The metrics change corrects **future** recomputes only. There is no backfill and no trigger change (out of scope by decision), and `getAffected` selects entities by recent `CommentV2."createdAt"`, so flagging an old comment does not by itself re-queue its entity. An already-inflated card count stays inflated until something else touches that entity.

## Testing

- `commentsv2.tos-visibility.test.ts` — new `CommentV2 counts and the ToS flag` block: counts under the filtered predicate, unfiltered for moderators, zero-without-a-query when there is no thread, and the reply-thread seed coming from the filtered count rather than the CTE's `commentCount: 9`.
- `commentv2.moderator-visibility.test.ts` — `getCount` added to the "handlers forward who is asking" matrix.
- `commentsv2.reply-threads.test.ts` — the metadata test now feeds a stale row count of 7 alongside `commentCounts: { 10: 4 }` and asserts 4, plus a case for a thread whose replies were all filtered out.
- **Revert-checked**: reverting `commentsv2.service.ts` and `commentsv2.reply-threads.ts` with the tests in place produces 5 legible assertion failures (`expected +0 to be 3`, `Number of calls: 0`, …), not a hang or a silent pass.
- `pnpm run typecheck` — 0 errors. `pnpm run test:lint-rules` — 23 files / 355 passed. `pnpm run test:unit:run` — 1459 files / 22,747 passed, 0 failed.

No schema change, no migration.

https://claude.ai/code/session_01P7PnkPT3UtTdd6aRhomxPF

### DIFF
--- a/src/server/controllers/__tests__/commentv2.moderator-visibility.test.ts
+++ b/src/server/controllers/__tests__/commentv2.moderator-visibility.test.ts
@@ -14,12 +14,14 @@ import type * as UserPreferences from '~/server/services/user-preferences.servic
 const getCommentsInfinite = vi.fn(async () => null);
 const getCommentsThreadDetails2 = vi.fn(async () => null);
 const getComment = vi.fn(async () => ({ id: 5, user: { id: 9 } }));
+const getCommentCount = vi.fn(async () => 0);
 
 vi.mock('~/server/services/commentsv2.service', async (importOriginal) => ({
   ...(await importOriginal<typeof CommentsV2Service>()),
   getCommentsInfinite: (...args: unknown[]) => getCommentsInfinite(...(args as [])),
   getCommentsThreadDetails2: (...args: unknown[]) => getCommentsThreadDetails2(...(args as [])),
   getComment: (...args: unknown[]) => getComment(...(args as [])),
+  getCommentCount: (...args: unknown[]) => getCommentCount(...(args as [])),
   isViewerContentOwner: vi.fn(async () => false),
 }));
 
@@ -31,8 +33,12 @@ vi.mock('~/server/services/user-preferences.service', async (importOriginal) => 
   BlockedByUsers: emptyPreference,
 }));
 
-const { getCommentHandler, getCommentsInfiniteHandler, getCommentsThreadDetailsHandler } =
-  await import('../commentv2.controller');
+const {
+  getCommentHandler,
+  getCommentCountV2Handler,
+  getCommentsInfiniteHandler,
+  getCommentsThreadDetailsHandler,
+} = await import('../commentv2.controller');
 
 const input = { entityType: 'image', entityId: 1 } as never;
 const ctxFor = (user: { id: number; isModerator?: boolean } | undefined) => ({ user } as never);
@@ -73,5 +79,19 @@ describe('CommentV2 handlers forward who is asking', () => {
     await getCommentHandler({ ctx: ctxFor(user), input: { id: 5 } as never });
 
     expect(getComment).toHaveBeenCalledWith(expect.objectContaining({ isModerator: expected }));
+  });
+
+  // The count is what renders "show N replies". Forwarding the wrong answer here doesn't serve a
+  // removed comment, it advertises one — an affordance that opens onto nothing.
+  it.each([
+    ['a moderator', { id: 1, isModerator: true }, true],
+    ['an ordinary signed-in user', { id: 2, isModerator: false }, false],
+    ['a signed-out viewer', undefined, false],
+  ])('getCount: %s', async (_label, user, expected) => {
+    await getCommentCountV2Handler({ ctx: ctxFor(user), input });
+
+    expect(getCommentCount).toHaveBeenCalledWith(
+      expect.objectContaining({ isModerator: expected })
+    );
   });
 });
--- a/src/server/controllers/commentv2.controller.ts
+++ b/src/server/controllers/commentv2.controller.ts
@@ -155,7 +155,7 @@ export const getCommentCountV2Handler = async ({
   input: CommentConnectorInput;
 }) => {
   try {
-    return await getCommentCount(input);
+    return await getCommentCount({ ...input, isModerator: ctx.user?.isModerator ?? false });
   } catch (error) {
     throw throwDbError(error);
   }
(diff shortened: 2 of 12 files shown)


manuelurenah:
## Revision: the count is back to one round trip

Fair push on the extra query — it was avoidable, and `db7ea14121` removes it.

`getCommentCount` was doing a `Thread` lookup and *then* a `CommentV2` count: two round trips where reading `Thread.commentCount` had been one, on a query the comment list fires per rendered comment. It now filters on the thread relation instead, so it is a single `commentV2.count`:

```ts
dbRead.commentV2.count({
  where: {
    thread: { [`${entityType}Id`]: entityId } as Prisma.ThreadWhereInput,
    tosViolation: isModerator ? undefined : false,
  },
});
```

Postgres plans that as a nested loop over two index scans, measured on production:

| | plan | time | buffers | round trips |
|---|---|---|---|---|
| `Thread.commentCount` (before this PR) | `Index Scan Thread_imageId_key` | 0.123 ms | 4 | 1 |
| resolve-then-count (first push) | two statements | ~0.17 ms | 15 | **2** |
| relation filter (now) | `Nested Loop` → `Thread_imageId_key` + `CommentV2_threadId` | **0.079 ms** | 12 | **1** |

## The trap this deliberately avoids

The tidier-looking way to write it is a filtered `_count` on `Thread`:

```ts
dbRead.thread.findUnique({ where: …, select: { _count: { select: { comments: { where: { tosViolation: false } } } } } })
```

Prisma compiles that to an **uncorrelated** aggregate subquery — it aggregates the whole table and *then* joins. Measured on production:

```
HashAggregate (rows=1861952) … Seq Scan on "CommentV2" (rows=2272325)
Execution Time: 1703.812 ms
```

1.7 s and 71K buffers per call, against 0.079 ms for the form above. There is a comment on the function saying so, because the "simplification" back to it is an easy one to make.

## The one query that is genuinely new

`getReplyThreads` still gains a third `groupBy` for the per-thread visible count. It sits inside the existing `Promise.all` alongside the comment fetch and the hidden-count `groupBy`, so it costs a connection but **no added latency**. It cannot be folded into the hidden-count query: that one excludes blocked/hidden users so its number matches what the modal will show, while this one replaces a raw row counter and must not, or it would disagree with `getCommentCount` — which it seeds.

Re-verified after the change: typecheck 0 errors, full unit suite 22,747 passed / 0 failed, and the revert-check still produces 8 legible assertion failures rather than a silent pass.

PR #4432, opened 2026-08-26
fix(auth): enforce email-domain blocklist + MX on every writer of User.email

JustMaier:
## What

The `EmailDomain` blocklist was consulted in exactly one place: the auth hub's magic-link action. Every other writer of `User.email` is free text that is never verified. This adds an `assertEmailAllowed` guard (blocklist + MX) to all of them, and the blocklist half to the hub's OAuth callback.

## Why

Measured against production on 2026-08-26.

The email path's guard works. Disposable-domain signups by day, email-only accounts:

| Day | email signups | on a blocklisted domain |
|---|---|---|
| Aug 22 | 2,399 | 630 |
| Aug 23 | 2,359 | 634 |
| Aug 24 | 3,750 | 1,968 |
| Aug 25 | 3,141 | 1,063 |
| Aug 26 | 1,850 | **0** |

Last one through: `2026-08-25 16:00:53Z`. The blocklist row's `updatedAt`: `2026-08-25 16:23:57Z`. Positive control for the zero: 1,850 email signups still happened on the 26th.

The burner ring moved to Reddit. Reddit's OAuth profile carries no email, so the account is created with `email: null` and onboarding's Profile step then *requires* one, validated for username and unchecked for email. Since Aug 18: 98.8% of Reddit accounts (1,927/1,950) hold an unverified address, 10.4% are on a domain already blocklisted (20% this week), against Google's 0.03%. Their banned accounts carry `gof33etchbitch.ccc`, `eh8798wit.com`, `gmail.com.com`, `linkedin.com`. `.ccc` is not a TLD.

## Why both halves

Neither subsumes the other. The blocklist covers ~8,500 **known** disposable providers, which are real domains that accept mail. The MX check covers **invented** domains, which no list can enumerate. Requiring MX rejects 5 of the 8 domains the ring's banned accounts used.

## Hook points

| Path | Guard |
|---|---|
| `completeOnboardingHandler` to `OnboardingSteps.Profile` | blocklist + MX |
| `updateUserById` (the `user.update` mutation) | blocklist + MX, only when no address is on file |
| `requestEmailChange` | blocklist + MX |
| auth hub OAuth callback | blocklist only |

The OAuth callback gets no MX probe on purpose: a provider-**verified** address is deliverable by construction, so a DNS lookup there is new failure surface on every login for no abuse benefit.

Two writers are deliberately exempt, both moderator-only, both now commented as such: `forceUpdateUserIdentity` (identity correction) and `restoreUser` (a mod putting back the address a closed account already had; re-judging it against a list that has moved would make some accounts unrestorable).

## Failure semantics: both halves degrade OPEN

- **Blocklist**: the lookup is a `redis.get` falling through to a `dbWrite` read, neither guarded. It now catches, logs `email-blocklist-lookup-failed`, and allows. Unguarded, a Redis failover or a saturated writer pool would have rejected every signup, profile-email set and email change for its duration, and Reddit accounts arrive with no address, so that is the whole funnel for that provider. Matches the hub's existing `catch { return [] }`.
- **MX**: only `ENOTFOUND`, `ENODATA` and `EBADNAME` are treated as answers. SERVFAIL, timeouts and an unreachable resolver all pass. There is a 3s deadline (c-ares has none of its own) which rejects with a plain `Error`, so the timeout arm fails open too.

`EBADNAME` is in the reject set deliberately: it is the resolver saying "this cannot be a hostname", which is what `x@!!!` and `gmail.com..` produce. Folding it into fail-open would accept exactly the invented input this exists to reject.

## Normalization

Domain comparison strips whitespace, case **and** a trailing FQDN dot, on **both** sides. The list is hand-edited by moderators. A trailing dot resolves identically (`gmail.com.` has 5 live MX records) but is a distinct string, so without this it misses every entry, passes the MX probe, and takes a second slot in a citext-unique column.

In the hub that is one exported `isBlockedDomain`, used by both call sites, plus `normalizeEmailAddress` applied on the **write** path. Both hub anti-abuse gates (`userExistsByEmail` as the returning-user exemption, and the `+`-alias block) key on the exact stored string, so a trailing dot was a free second identity against both.

The normalization is forward-only, so the obvious question is whether dotted rows already exist. They do not: `SELECT count(*) FROM "User" WHERE email LIKE '%.'` returns **0**, against 11,566,827 rows with an email. No backfill needed. Domain case is the other non-canonical spelling that can reach the column, and it is not a duplicate risk at all: citext makes it match the canonical row, which is why this needed a normalization on the write rather than a backfill.

`userOnboardingSchema`'s Profile branch is now `z.string().email()` rather than a bare `z.string()`. It was the only unvalidated writer of the column.

## False-positive risk, measured

Prod users created since Aug 19 with no ban:

- Top 400 domains (56,462 users): rejects **2 domains / 6 users (0.011%)**, namely `globalcampus.edu.pl` (NXDOMAIN) and `hotmail.co` (typo, no MX).
- Random 400 domains seen once or twice (428 users): rejects **9 domains / 9 users (2.10%)**, of which 6 are disposable or typo'd.

The second stratum is the honest upper bound; those rare domains are a small share of total signups.

## Observability

New `hub_blocked_email_domain_signups_total{path}` counter, incremented on **both** halves of the gate. Previously a refusal incremented nothing anywhere, since `loginsTotal` is past the OAuth throw and `emailLoginFailuresTotal` counts only server errors, so nobody could have told a gate that fired ten thousand times from one that never fired, or measured the false-positive rate above.

On the main app, `onboardingErrorCounter` now skips `TRPCError`, so policy refusals stop inflating an error signal.

A blocked OAuth signup is redirected to `/login?error=BlockedDomain` with its `returnUrl` preserved, not thrown as a 400. `apps/auth` ships no `+error.svelte`, so a throw rendered SvelteKit's bare fallback page with no way back, and the flow cookies that would allow a retry are deleted earlier in the handler.

## Verification

- `pnpm run typecheck`: 0 type errors
- `svelte-check` (apps/auth): `8043 FILES 0 ERRORS 1 WARNINGS`; the warning is pre-existing in `+page.svelte`
- `eslint` over the changed files: 0 errors
- apps suite: `Test Files 93 passed (93)` / `Tests 1009 passed | 35 skipped (1044)`
- unit suite: posted as a comment below

**Every guard has a mutation run and reverted.** Fourteen in total; each gives exit 1 with a named assertion, never a hang:

| mutation | result |
|---|---|
| drop the MX half | `3 failed \| 8 passed (11)` |
| drop the blocklist half | `4 failed \| 7 passed (11)` |
| fail *closed* on a resolver error | `2 failed \| 9 passed (11)` |
| raw `.includes` instead of case normalization | `1 failed \| 10 passed (11)` |
| drop `data ?? []` in `getBlocklistDTO` | `1 failed \| 16 passed (17)` |
| remove the degrade-open catch | `1 failed \| 19 passed (20)` |
| drop the trailing-dot strip | `1 failed \| 19 passed (20)` |
| drop `EBADNAME` from the reject set | `1 failed \| 19 passed (20)` |
| flip `!==` to `===` at the onboarding call site | `3 failed \| 5 passed (8)` |
| guard-on-existing-address in `updateUserById` (the lockout) | `1 failed \| 7 passed (8)` |
| delete the `requestEmailChange` call | `2 failed \| 6 passed (8)` |
| delete the `updateUserById` call | `2 failed \| 6 passed (8)` |
| hub `emailDomain` back to `split('@')[1]` | `1 failed \| 12 passed (13)` |
| hub `isBlockedDomain` entry side unnormalized | `1 failed \| 20 passed (21)` |

The `!==` to `===` one matters most: it makes the guard fire only when the address is **unchanged**, so it is wholly inert on the Reddit onboarding path this change exists to close, and before the call-site suite existed it turned nothing red.

## Found during review, not by me

Three adversarial reviewers, each prompted to refute. Between them:

- The blocklist half failed **closed** on a Redis or DB blip, 500ing the whole signup funnel.
- `EBADNAME` was being **accepted**, and the onboarding schema was a bare `z.string()`, so `x@!!!` stored verbatim.
- A trailing-dot FQDN defeated the blocklist and the email-uniqueness constraint.
- The call sites had no test at all, so the guard could be made inert with a one-character edit and a green suite.
- The blocklist matcher had no negative control: every "allowed" test ran against an empty list, so a widening mutation survived.
- The OAuth reject was a dead-end error page, and the whole gate was unobservable.
- A false claim in a comment: the LRU does **not** coalesce concurrent misses, so "a burst of signups from one ring hits DNS once" was wrong as written. Comment corrected rather than the behaviour changed.

One reviewer independently enumerated **eight** writers of `User.email` across `src/`, `apps/` and `packages/` rather than trusting the list above, and found `restoreUser` missing from it. Its own caveat is worth repeating: a multiline grep **missed** `forceUpdateUserIdentity`, because that function builds `data.email` on a separate line from the `update()` call. A grep-only enumeration of this column is not sound.

## Known gaps

- **The hub's callback guard and its login-action comparison have no test of their own.** `apps/auth/src/routes/login/__tests__/` holds only `login-captcha.test.ts`. The shared helpers they call (`emailDomain`, `isBlockedDomain`, `normalizeEmailAddress`) are tested and mutation-verified; the callback's exemption branches are not. The branch most worth covering is `isLinkedAccount` true with `User.email` null, the one case where the two exemptions are not interchangeable, and the one a future refactor would most plausibly drop as redundant.
- `mxCache` is module-scope with no reset. The guard suite is order-independent only because every test uses a distinct domain. Adding a production seam purely for tests seemed the worse trade.
- Domain matching is exact equality, so subdomains and punycode/IDN variants miss. Inherited, shared by all copies, unchanged here.

## Not in scope

No migration. No product change to new-account friction. Requiring a *verified* email before posting is a separate decision, and worth noting that verification alone is not a spam signal: banned-account rate since Aug 18 was 3.84% for verified addresses versus 2.73% for unverified, because a disposable inbox verifies fine.

### DIFF
--- a/apps/auth/src/lib/server/auth/__tests__/blocklist.test.ts
+++ b/apps/auth/src/lib/server/auth/__tests__/blocklist.test.ts
@@ -18,7 +18,12 @@ vi.mock('../../db/db', () => ({
   },
 }));
 
-import { getBlockedEmailDomains } from '../blocklist';
+import {
+  emailDomain,
+  getBlockedEmailDomains,
+  isBlockedDomain,
+  normalizeEmailAddress,
+} from '../blocklist';
 
 const BLOCKLIST_KEY = 'system:blocklist:EmailDomain';
 
@@ -41,7 +46,10 @@ beforeEach(() => vi.clearAllMocks());
 describe('getBlockedEmailDomains', () => {
   it('returns the warm redis cache without touching the DB', async () => {
     const redis = makeRedis();
-    redis._store.set(BLOCKLIST_KEY, JSON.stringify({ type: 'EmailDomain', data: ['evil.com', 'bad.io'] }));
+    redis._store.set(
+      BLOCKLIST_KEY,
+      JSON.stringify({ type: 'EmailDomain', data: ['evil.com', 'bad.io'] })
+    );
     h.getRedis.mockReturnValue(redis);
 
     expect(await getBlockedEmailDomains()).toEqual(['evil.com', 'bad.io']);
@@ -127,5 +135,65 @@ describe('getBlockedEmailDomains', () => {
     h.getRedis.mockReturnValue(redis);
     h.executeTakeFirst.mockResolvedValue({ data: ['db.com'] });
     expect(await getBlockedEmailDomains()).toEqual(['db.com']);
+  });
+});
+
+describe('emailDomain', () => {
+  // Both hub call sites go through this, so reverting either one to `split('@')[1]` cannot pass
+  // unnoticed. The main app carries its own copy of the rule and its own test for it.
+  it('reads the domain after the LAST @, not the first', () => {
+    expect(emailDomain('"a@b"@example.com')).toBe('example.com');
+  });
+
+  it('lowercases and trims', () => {
+    expect(emailDomain('someone@ Example.COM ')).toBe('example.com');
+  });
+
+  it('returns empty string for an address with no @', () => {
+    expect(emailDomain('not-an-email')).toBe('');
+  });
+
+  it('returns empty string for a trailing @', () => {
+    expect(emailDomain('someone@')).toBe('');
+  });
+});
+
+describe('isBlockedDomain', () => {
+  // Both sides must go through the same normalizer. Stripping the input but not the entry would
+  // make a hand-typed `provider.com.` enforced by the main app and silently inert here.
+  it('matches a list entry that carries a trailing FQDN dot', () => {
+    expect(isBlockedDomain(['blocked.test.'], 'blocked.test')).toBe(true);
+  });
+
+  it('matches an entry with case and whitespace', () => {
+    expect(isBlockedDomain(['  Blocked.TEST  '], 'blocked.test')).toBe(true);
+  });
+
+  it('matches the WHOLE domain, not a substring', () => {
+    expect(isBlockedDomain(['ocked.test', 'test'], 'blocked.test')).toBe(false);
+  });
+
+  it('never matches an empty domain, whatever the list holds', () => {
+    expect(isBlockedDomain([''], '')).toBe(false);
+  });
+});
+
+describe('normalizeEmailAddress', () => {
+  // `userExistsByEmail` is both the returning-user exemption and the `+`-alias gate, and both key
+  // on the exact stored string, so a trailing dot would be a free second identity against both.
+  it('strips a trailing FQDN dot from the domain', () => {
+    expect(normalizeEmailAddress('someone@gmail.com.')).toBe('someone@gmail.com');
+  });
+
+  it('lowercases the domain and leaves the local part intact', () => {
+    expect(normalizeEmailAddress('Someone@GMAIL.com')).toBe('Someone@gmail.com');
+  });
+
+  it('normalizes the domain after the LAST @', () => {
+    expect(normalizeEmailAddress('"a@b"@GMAIL.com.')).toBe('"a@b"@gmail.com');
+  });
+
+  it('leaves an address with no @ alone beyond trim and case', () => {
+    expect(normalizeEmailAddress(' NotAnEmail ')).toBe('notanemail');
   });
 });
--- a/apps/auth/src/lib/server/auth/blocklist.ts
+++ b/apps/auth/src/lib/server/auth/blocklist.ts
@@ -23,6 +23,47 @@ const TTL_SECONDS = 5 * 60;
 type StringGet = { get(k: string): Promise<string | null | undefined> };
 type StringSet = { set(k: string, v: string, o: { EX: number }): Promise<unknown> };
 
+/** Trailing dot is the FQDN root and resolves identically, so it must not read as a new domain. */
+function normalizeDomain(domain: string): string {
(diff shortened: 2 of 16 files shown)


JustMaier:
Final verification on d69caefb88e74e0e6f5c3590e145f9039a9432d2, read from the log files rather than the completion notices (one of those reported exit 0 over a run whose log ended in ELIFECYCLE exit 1).

```
unit    Test Files  1461 passed | 1 skipped (1462)
        Tests  22793 passed | 27 skipped (22820)
apps    Test Files  93 passed (93)
        Tests  1009 passed | 35 skipped (1044)
```

Both totals account for every file: 1461+1=1462 and 93=93.

Submodule check, since a worktree without `event-engine-common` collects **zero** tests from this file and still exits clean: `blocks.router.workflow.test.ts` -> `Test Files 1 passed (1) / Tests 358 passed (358)`.

`pnpm run typecheck` -> `OK — 0 type errors`. `svelte-check` -> `8043 FILES 0 ERRORS 1 WARNINGS` (pre-existing, in a file this branch did not touch). `eslint` over the changed `src/` files -> `8 problems (0 errors, 8 warnings)`, all pre-existing unused-vars in files that only gained an import.

PR #4434, opened 2026-08-26
fix(hubs): let the source picker search past the 500 most recent relationships

JustMaier:
## The bug

Reported today: `daceheg192491` could not find `CatusRex` in the hub source picker, despite following them.

Confirmed on the prod replica:

| | |
|---|---|
| `daceheg192491` follows | **2,738** creators |
| `CatusRex`'s position, newest-first | **1,905** |
| Rows the search could see | **500** |

Not paging, and not the sort — a hard window. `getHubSourceSuggestions` reads the viewer's relationship list with `take: SUGGESTIONS_WINDOW` and then runs the name filter over *whatever that returned*. So the `take` is not a page size, it is the entire searchable set, and anything past row 500 is unreachable by any search term.

## The fix

Searching reads a 5,000-row window; a bare suggestion list keeps the 500-row one. All five relationship queries move together — follows, collection contributions, own models, bell subscriptions, bookmarks — because the models arm has the same shape.

The window stays because removing it is not free: expressed as a Prisma relation filter, the name match is a subquery the planner satisfies by walking the whole list, which is what the existing comment measured at 4.8s / 4.85GB.

**Measured on the prod replica**, largest accounts on the site, term matching nothing (the worst case — `take` cannot stop early):

| Arm | Account size | Execution |
|---|---|---|
| Follows | 118,609 | **402 ms** |
| Bell-subscribed models | 250,491 | **392 ms** |

A 15,000-id union (three model lists at the new window) measured 392 ms, essentially flat against 5,000 — the planner switches strategy rather than scaling linearly.

## What is still bounded

| | > 500 (before) | > 5,000 (after) |
|---|---|---|
| Accounts by follows | 9,119 | **57** |
| Accounts by bell subscriptions | 12,587 | **1,066** |

Above the window a search is still partial. Closing it completely needs a trigram index on `User.username` / `Model.name` so the term can drive instead of the relationship list — a separate change with a manual migration, not folded in here.

## Verification

- New tests assert the search window exceeds the listing window on the creators arm and on all three models-arm queries.
- **Positive control**: with `suggestionWindow` reverted to a single window, both fail with `expected 500 to be greater than 500` — `Tests 2 failed | 77 passed | 4 skipped (83)`, exit 1.
- `pnpm run test:unit:run` — `Test Files 1465 passed | 1 skipped (1466)`, `Tests 22946 passed | 27 skipped (22973)`, exit 0.
- `blocks.router.workflow.test.ts` collected **358** tests (submodule initialised).
- `pnpm run lint` — 360 errors, none in the touched files; pre-existing on `main`.

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260827000000_model_engagement_notify_recency_idx/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260827000000_model_engagement_notify_recency_idx/migration.sql
@@ -1,0 +1,22 @@
+-- Serves the hub source picker's "models you asked to be notified about" read:
+--   WHERE "userId" = $1 AND type = 'Notify' ORDER BY "createdAt" DESC LIMIT $2
+--
+-- Measured on the prod replica before this index: the heaviest account (250,491
+-- Notify rows) reads ALL of them, ~130,800 buffers / ~1.02 GB touched, 283ms warm
+-- and 1.56s cold, on every debounced keystroke. ModelEngagement_userId_type_modelId_idx
+-- is (userId, type) INCLUDE (modelId), which cannot serve the ORDER BY, so the
+-- planner heap-fetches every row and sorts. With this index the same read is an
+-- index-only scan that stops at the window.
+--
+-- Partial on Notify, matching ModelEngagement_mute_userId_modelId_idx beside it:
+-- the picker never reads the other types, and Notify is 43.1M of the table's 74.4M
+-- rows. Prisma cannot express a partial index, so this one lives in SQL only — as
+-- that Mute index already does.
+--
+-- CONCURRENTLY, because the table is 10 GB. Run it OUTSIDE a transaction: psql's
+-- -c with two statements opens one implicitly and the server refuses. A cancelled
+-- CONCURRENTLY build leaves an INVALID index that must be dropped before retrying.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "ModelEngagement_notify_userId_createdAt_idx"
+  ON "ModelEngagement" ("userId", "createdAt" DESC)
+  INCLUDE ("modelId")
+  WHERE type = 'Notify';
--- a/src/components/Hubs/HubSourceSearch.tsx
+++ b/src/components/Hubs/HubSourceSearch.tsx
@@ -33,8 +33,9 @@ const tabs = [
 ];
 
 /**
- * One type at a time, over a bounded window of the viewer's own relationships —
- * see SUGGESTIONS_WINDOW in user-hub.service.ts.
+ * One type at a time, over a bounded window of the viewer's own relationships. The
+ * window is wider when there is a term than when there is not — see
+ * SUGGESTIONS_SEARCH_WINDOW and SUGGESTIONS_WINDOW in user-hub.service.ts.
  */
 export function HubSourceSearch({
   onSelect,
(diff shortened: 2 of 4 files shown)


JustMaier:
End-to-end A/B on the prod replica, running the exact shape the service emits for `daceheg192491` searching `catus`:

```
window 500  -> 0 rows returned
window 5000 -> CatusRex (2776046), explicatus2654 (1009681)
```

So the reported case reproduces on the old window and resolves on the new one, against real data rather than a mock.

JustMaier:
## 🔴 Migration added — needs applying by hand, outside a transaction

`packages/civitai-db-schema/prisma/migrations/20260827000000_model_engagement_notify_recency_idx/migration.sql`

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS "ModelEngagement_notify_userId_createdAt_idx"
  ON "ModelEngagement" ("userId", "createdAt" DESC)
  INCLUDE ("modelId")
  WHERE type = 'Notify';
```

**Before running it, the numbers that matter:**

| | |
|---|---|
| `ModelEngagement` | **74,373,049 rows / 10 GB** |
| of which `Notify` | **43,116,519** (58%) |
| Build | `CONCURRENTLY`, so no write lock — but it is a long build on a table this size |
| Transaction | must be **outside** one; `psql -c` with two statements opens one implicitly and the server refuses |
| If cancelled | leaves an **INVALID** index that must be dropped before retrying |

**Why.** The picker's bell-subscription arm is the endpoint's real cost, measured on the prod replica: the heaviest account (250,491 Notify rows) reads *all* of them — ~130,800 buffers, ~1.02 GB touched, **283 ms warm / 1.56 s cold**, per debounced keystroke. That is true at any window, so this PR's change neither caused it nor moved it. `ModelEngagement_userId_type_modelId_idx` is `(userId, type) INCLUDE (modelId)` and cannot serve the `ORDER BY createdAt`, so the planner heap-fetches every row and sorts. With this index the read becomes an index-only scan that stops at the window.

Partial on `Notify` because the picker reads no other type, mirroring `ModelEngagement_mute_userId_modelId_idx` already on the table. Prisma cannot express a partial index, so — like that one — it lives in SQL only and `schema.full.prisma` is unchanged, which is why there is no `db:check-generated` step here.

## Also in this round (five-lane review)

- A one-character term is now treated as **no term**. It costs the same relationship read as a useful one and matches most of the window anyway. This is the only bound on the endpoint's per-keystroke cost until the index above is applied.
- **The measurements in the window comment were wrong in both directions** and are replaced with per-arm numbers at both windows. Material correction: **raising the window is close to free.** Every arm orders by a column no composite index covers, so the whole relationship list is read either way and the `take` only sizes the top-N heap. On the largest follows account the *old* 500 window picks a worse plan off the global `createdAt` index — **444 ms / 165,001 buffers** against **181 ms / 48,212** at 5,000.
- `nulls: 'last'` on the collections order — the column is nullable and Postgres sorts `DESC NULLS FIRST`, which would have filled the window with the rows carrying no date at all.
- The tripwire test asserting a constant is replaced with the real assertion behind `skipIf`, matching the four cases already in the file.
- New test: the whole window must reach the name query. Slicing it back to a page in `scopeSuggestionIds` was a *worse* regression than the bug being fixed, and nothing in the file could see it. Verified by mutation.

`Test Files 1465 passed | 1 skipped (1466)`, exit 0. Typecheck 0 errors.

PR #4439, opened 2026-08-26
feat(hubs): give a shared hub link a preview card

JustMaier:
## The bug

A hub link sent in a DM produces nothing at all.

Chat unfurls by **fetching the URL with no session** and reading its OpenGraph tags (`chat.service.ts:578` → `unfurl.js`). Verified against prod, anonymously:

| URL | Response |
|---|---|
| `civitai.com/models/4384` | **200**, `og:title` + `og:description` + `og:image` |
| `civitai.com/hubs/19` (a **Public** hub) | **404** |

Two independent causes:

1. `getServerSideProps` returned `notFound` whenever `features.userHubs` was false, and the unfurler is always signed out.
2. Even signed in, the page's `<Meta>` was rendered from `trpc.userHub.getById.useQuery`. On the server that query has not resolved, so the component returns its loading branch and **`<Meta>` never renders** — the HTML an unfurler reads has none of it. Invisible in a browser, where the client fills the title in after hydration.

## The changes

**`/api/og?type=hub&id=N`** — the same shared card every other entity type uses. Title = hub name, description = the hub's description, creator = owner, stats = source count and (above zero) follower count.

**No cover image.** A hub has none of its own, and the first image of its feed is not a safe substitute: this card is served unauthenticated, so it would show that image to everyone the link reaches whatever their own browsing level is. The card already falls back to the Civitai mark. Justin's call, 2026-08-26.

**Meta from server props**, rendered above every early return — loading and not-found included — so it is in the SSR HTML regardless of what the client query does.

**`hubRouteIsDark`** spares a Public hub the `user-hubs` flag. It buys the meta only: the body still needs the flag, since the hub and its feed both arrive through flag-gated tRPC reads. Hubs stay testers-only for the app.

## What cannot leak

`getHubCardData` scopes to `availability: Public` in the `where`, with no viewer arm — there is no session to scope to. A private hub returns null, and null is the generic Civitai fallback the endpoint already serves for any missing entity, at 200. So the card cannot confirm a hub exists or read its name.

## Verification

Against a dev server (`db=dev`), with one dev hub temporarily flipped to Public and restored afterwards:

| Probe | Result |
|---|---|
| `/api/og?type=hub&id=14` (Public) | **200 image/png**, 55,906 bytes, 2.1s — card renders |
| `/hubs/14` (Public), signed out, flag off | **200** with `og:title`, `og:description`, `og:image` |
| `/hubs/13` (Private), signed out | **404** |
| `/api/og?type=hub&id=13` (Private) | generic Civitai fallback, no hub information |

Tests, each with a positive control:

- `getHubCardData` resolves Public only — dropping `availability` from the `where` fails it.
- `getUserHubForRoute` returns the description — returning null for it fails it.
- `hubRouteIsDark` over five flag × availability combinations — flattening it to `!hubsEnabled` fails the `flag false + Public` case with `expected true to be false`.

- `pnpm run test:unit:run` — `Test Files 1465 passed | 1 skipped (1466)`, `Tests 22954 passed | 27 skipped (22981)`, exit 0.
- `blocks.router.workflow.test.ts` collected **358** tests.
- `pnpm run lint` — the one warning on `og.tsx` is pre-existing on `main` (a `eslint-disable-next-line` separated from its `` by a comment).

No migration. No schema change. Inert in production until the `user-hubs` flag opens past the testers segment.

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -88,6 +88,9 @@ EMAIL_FROM=
 # Endpoint Protection
 JOB_TOKEN=thisisnotatoken
 WEBHOOK_TOKEN=thisisnotatoken
+# Permutes the alphabet that encodes a hub id into its public URL. Empty is fine
+# locally; set it per environment so hub links are not portable between them.
+HUB_ID_SALT=
 # Base URL of the standalone moderator app (apps/moderator). Migrated /moderator/* routes redirect here
 # via the moderator catchall page. Defaults to https://moderator.civitai.com if unset.
 MODERATOR_APP_URL=https://moderator.civitai.com
--- a/package.json
+++ b/package.json
@@ -315,6 +315,7 @@
     "slugify": "^1.6.5",
     "socket.io-client": "^4.5.4",
     "source-map": "^0.7.4",
+    "sqids": "^0.3.0",
     "stacktrace-parser": "^0.1.10",
     "stream-to-blob": "^2.0.1",
     "stripe": "^11.6.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -616,6 +616,9 @@ importers:
       source-map:
         specifier: ^0.7.4
         version: 0.7.6
+      sqids:
+        specifier: ^0.3.0
+        version: 0.3.0
       stacktrace-parser:
         specifier: ^0.1.10
         version: 0.1.11
@@ -11065,6 +11068,9 @@ packages:
 
   sprintf-js@1.0.3:
     resolution: {integrity: sha512-D9cPgkvLlV3t3IzL0D0YLvGA9Ahk4PcvVwUbN0dSGr1aP0Nrt4AEnTUbuGvquEC0mA64Gqt1fzirlRs5ibXx8g==}
+
+  sqids@0.3.0:
+    resolution: {integrity: sha512-lOQK1ucVg+W6n3FhRwwSeUijxe93b51Bfz5PMRMihVf1iVkl82ePQG7V5vwrhzB11v0NtsR25PSZRGiSomJaJw==}
 
   stable-hash@0.0.5:
     resolution: {integrity: sha512-+L3ccpzibovGXFK+Ap/f8LOS0ahMrHTf3xu7mMLSpEGU0EO9ucaysSylKo9eRDFNhWve/y275iPmIZ4z39a9iA==}
@@ -23970,6 +23976,8 @@ snapshots:
   split2@4.2.0: {}
 
   sprintf-js@1.0.3: {}
+
+  sqids@0.3.0: {}
 
   stable-hash@0.0.5: {}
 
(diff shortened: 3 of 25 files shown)


JustMaier:
## Hub ids are now encoded in the URL — `34f145cdc5`

`UserHub.id` is a dense autoincrement and both the page and the preview card answer unauthenticated, so every public hub was walkable by counting. Justin's call: Public means fully public, but not trivially crawlable — and encoding over a random key column, since it needs no second value and no second index.

```
/hubs/izMK3WCh                      200
/hubs/14                            404   ← the pre-encoding format
/api/og?type=hub&id=izMK3WCh        200 image/png, 55,906 bytes
/api/og?type=hub&id=14              400
/api/og?type=model&id=4384          200 image/png   ← the other six types unchanged
```

All five measured against a dev server, not reasoned about. The page's `og:image` and `canonical` both come out built from the key.

### 🔴 One rule this depends on

**The salt is server-side and has to stay there.** As a `NEXT_PUBLIC_` var it ships in the JS bundle and the encoding becomes decorative. So the client never encodes: `toHubDetail` puts `key` on every hub it returns, `hubUrl` builds the path from that, and `userHub.getById` is now addressed **by key** so a component holding only the URL can still resolve the hub. Everything internal stays an int.

It is obfuscation, not authorisation — every read still applies `hubViewerWhere`, so a decoded id buys nothing a guessed one would not. It removes cheap enumeration, nothing else.

### Deploy ordering

`HUB_ID_SALT` is new (`src/env/server-schema.ts`, defaulted to `''`). Empty means the alphabet as written, which is what dev and test run with. **Set it per environment before the deploy** — changing it later changes every hub URL in that environment at once.

### Breaking, deliberately

Pre-encoding links 404. Accepting bare integers back would leave enumeration exactly as open as it was, so it has to be one or the other. Five public hubs, testers only, no external links — cheap now, expensive after launch. Pinned by two tests with the reason in them, because "let's also accept the int" is the obvious future edit.

### Also in this round (five-lane review)

- **Revocation reaches the card.** It took the endpoint's 7-day edge cache, so turning sharing off left the hub's name, description and owner on the CDN for a week — while the share dialog promises every link stops working. Hub is the first entry in a `REVOCABLE_TYPES` set that takes a 5-minute cache.
- **The card counted disabled sources**, advertising a hub as larger than the page it opens. Enabled only now.
- **Meta preferred the stale SSR copy** over the loaded hub, so a rename didn't show until a full reload.
- **New SSR wiring test.** Moving the flag gate back above the lookup killed every preview with nothing red — verified by mutation that the gate-order revert and dropping the props each redden it now.
- **Both Prisma selects pinned by argument assertions.** A mocked call ignores `select`, so dropping `availability` left the whole suite green while every public hub 404'd in production.

`Test Files 1467 passed | 1 skipped (1468)`, exit 0. Typecheck 0 errors.

PR #4440, opened 2026-08-27
fix(apps): a republish whose images changed goes back to review — on the surface that can show them

ZacxDev:
## The hole

An owner can **unpublish their own approved listing → swap the icon, cover and every screenshot → republish**, and the new imagery goes straight onto a public store card with **no content review**.

A `removed` listing is directly asset-editable — `assertOwnerAssetEditable` refuses only an `approved`, non-shadow row — and none of the gates `republishOwnListing` already runs is a review:

| gate | what it actually reads |
|---|---|
| `assertListingAssetsScanCleanInTx` | `Image.ingestion` — scan **status** |
| `assertOffsiteListingActionableInTx` | the destination link |
| #4418's rating floor | **maturity** (`nsfwLevel`) |

#4418 closed the maturity half. This closes the review half.

## What this changes

**If any listing asset differs from what was recorded at the last approval, the republish routes the listing to `pending` (re-review) instead of `approved`, and re-queues it. If nothing changed, republish is immediate exactly as it is today** — no extra queue entry, no moderator involvement, no behaviour change at all.

## 🔴 The review has to be able to SHOW what is under review

An earlier revision of this PR re-queued an **on-site** app by cloning its already-approved `AppBlockPublishRequest`. That was wrong, and wrong in the way that is hardest to see: the clone carries the **same bundle key, the same `bundleSha256`, the same manifest and the same version** as the row the moderator had already approved, and `OnsiteReviewModal` draws its screenshots by **re-extracting them from the bundle ZIP** — it contains no reference to `AppListing.iconId`, `coverId` or `AppListingScreenshot` at all. The recorded baseline was rendered nowhere (`OffsiteReviewQueue` reads `before` only for `claim` events). So the moderator was handed a byte-identical re-submission of something they had already said yes to, with **nothing on screen about the pictures that caused the re-queue** — a review that structurally could not review the thing under review.

**Both kinds now re-queue onto `AppListingPublishRequest`.** That queue's modal already keys `appListings.getAssets` and `appListings.getListingPreviewForReview` off `request.appListingId`, so it renders this listing's real icon, cover and screenshots plus a store-layout preview. Both reads are kind-agnostic, and `listPendingOffsiteRequests` already selects `kind: { in: ['onsite','offsite'] }` with no `revisionOfId` constraint.

It also makes the two arms **symmetric**, which removes "the guarantee depends on which arm" at the root rather than patching each arm.

| | flip | re-queue | re-approved by |
|---|---|---|---|
| off-site | `removed → pending` | fresh pending non-shadow `AppListingPublishRequest`, owner-submitted | `approveExternalRequest`'s widened `{draft,pending}` listing guard |
| on-site | `removed → pending` | **the same** | `approveExternalRequest` — plus the two on-site branches below |

On the review arm the backing block is **deliberately left suspended** — an app whose store card is awaiting review must not be serving, the same posture `resetOnsiteListingToPending` takes. Withdraw/reject leave listing `removed` + block `suspended`, which is exactly the state the owner was in before they pressed Republish, so the two halves never diverge.

### This is the first NON-SHADOW on-site listing request in the system

Every other on-site `AppListingPublishRequest` is a media **revision** (`revisionOfId != null`) and returns early from `approveExternalRequest` into `applyApprovedRevision`. A non-shadow one takes the main approve path, which had no on-site behaviour **because nothing could reach it**. Two branches were added there — new code on a newly reachable branch, not a change to any existing flow:

1. **The rating is RAISE-ONLY over the app's own declaration.** Off-site, the rating *is* the assets, so the derived value replaces the stored one. On-site it describes the **app** (manifest-declared) and the store art is a strictly smaller surface — replacing it would **lower a `pg13` app to `g`** because its icon happened to scan tame. `resolveOnsiteApprovalContentRating` floors instead, reusing `resolveListingRatingFloorInTx`, the same helper the draft→approved transition and `republishOwnListing` use, so all three on-site rating sites have one spelling. A moderator override may still raise, never lower.
2. **The backing block is un-suspended.** `AppBlock.status` is the only gate on whether a hosted app serves. Approving only the listing would publish a store card pointing at a dead app — the same failure `(3b-reset)` exists to prevent on the other surface.

### The maturity floor, closed at the root

Routing through the listing approve means the floor is applied at approve time. The **pre-existing** inconsistency it was composing with is fixed directly:

> `approveRequest`'s `(3b-reset)` branch restored a `pending` on-site listing by writing `{ status: 'approved' }` and **nothing else**, while its sibling `(3b-transition)` floored.

A `pending` listing is freely owner-asset-editable (`assertOwnerAssetEditable` refuses only `approved`, and `pending` is in `AUTHORABLE_LISTING_STATUSES`, so the Media tab is offered normally); the attach path rejects `Blocked`/`NotFound` but never compares an image's `nsfwLevel` against the listing rating; and `reDeriveContentRatingForModLiveEdit` no-ops for a non-moderator **and** for any status other than `approved`. So *mod reset → owner swaps in a `Scanned`-but-mature cover → mod re-approves* restored the listing live under a stale rating. `(3b-reset)` now floors too, raise-only.

That inconsistency is the real bug; this feature merely made it easy to reach. Both halves are pinned by tests.

## The signal: recorded, not inferred

**Nothing in the schema captured the approved asset set.** Enumerated, not sampled:

- **No approve path writes a moderation event at all.** There is no `approve` verb in the moderation-action taxonomy.
- **All twelve non-test writers of `appListingModerationEvent.create`** write `before`/`after` payloads of the shape `{status}` (10), `{userId}` (1) or `{recipientUserIds, includeCollaborators}` (1). **Zero record an asset id.**
- `AppListingPublishRequest` snapshots nothing (slug / status / notes / changelog).
- **The timestamps are unusable**: the republish flip *itself* bumps `AppListing.updatedAt`; screenshot add/remove/reorder/caption does not bump the parent; and a **deleted** screenshot row leaves no timestamp behind to compare.
- A shadow revision *is* an asset copy, but of the parent **at mint time**, not at approval, and it is deleted on approve/reject/withdraw.
- The only existing comparator is `src/components/Apps/listingRevisionDrift.ts` — **client-side, never persisted**, and its own docstring says answering "did the baseline move?" needs a stored baseline.

So the baseline is now **written**: `unpublishOwnListing` records the live icon / cover / ordered screenshots into the `owner-unpublish` event's `before.assets`, inside the same transaction as the flip.

**Why that instant is exactly "the last approval":** the flip runs on a listing that is `approved`, and an approved non-shadow listing is *not* owner-asset-editable — owner edits go through a shadow revision a moderator reviews before it is copied onto the parent, and a moderator's live curation is itself a reviewed action.

### 🔴 The two absences go OPPOSITE ways

`readRecordedAssetBaseline` returns three things, not two, and the distinction is load-bearing:

| baseline | meaning | verdict |
|---|---|---|
| `snapshot` | a readable baseline | compare; differs ⇒ review |
| `unreadable` | the event **has** an `assets` key and it does not parse (malformed, or a version this build does not understand) | **review** — fail closed |
| `absent` | the event carries **no** `assets` key | **immediate** — pre-existing behaviour |

The `absent` arm is the population of listings unpublished **before** this ships. An earlier revision routed those to review as well. That is not a safety win: there is no recorded "before", so there is *no change to describe to a moderator* — while it takes **every already-removed listing offline behind a moderator on ship day**. Measured against production: every republish-eligible removed on-site listing is in exactly that state, so the fail-closed reading would have swept all of them and none of the reviews would have shown anything.

`unreadable` is the opposite case — unbounded, and evidence something is wrong *right now* — and it still fails closed. The test is on the **presence of the key**, not on the parse result, so the two can never collapse into one.

### What counts as "changed"

Icon id, cover id, and the **positional** list of screenshot `{imageId, caption}`.

- Screenshot **order values are not recorded** — renumbering `0,1,2 → 10,20,30` is not a change; genuinely swapping two screenshots is.
- Screenshot **row ids are not recorded** — deleting a screenshot and re-adding the identical image + caption is not a change.
- **Captions are in scope** — a caption is public text on the store card. Whitespace-only/blank edits normalise away.
- Rows with no `imageId` are dropped — the same definition `assertListingAssetsScanCleanInTx` uses.

### The publish floor is pre-checked, and it is not decoration

`approveExternalRequest` asserts icon+cover and does **not** exempt on-site, while the on-site **first**-publish path never asserts it — so an approved on-site listing is genuinely allowed to have no icon or no cover. Routing such a listing to review would strand it: pending listing, suspended app, and a moderator approve that fails the floor every time with no owner-reachable way out. The review arm therefore asserts the floor **before writing anything**, turning that into an ordinary "add an icon and a cover" error and leaving the listing `removed` where its owner can still fix it.

### 🔴 `one_pending_per_slug` was an implicit MUTUAL EXCLUSION, and removing it took away more than one thing

The partial-unique index `one_pending_per_slug` lives on `AppBlockPublishRequest`. While that table was the **only** surface an on-site app could be re-reviewed on, it was also doing unnamed safety work: at most one open review per slug, full stop. Two sites relied on that without saying so, both by the same inference — *a `pending` on-site listing ⇒ a moderator reset ⇒ a pending block request ⇒ **this** block request is the review that is open on it*. Re-queueing on `AppListingPublishRequest` leaves the index free and breaks the inference. A previous round found and closed the **withdraw** leg; this round closes the **approve** leg, which is the one that had a security consequence.

**A version APPROVE no longer restores a listing whose review lives elsewhere.** `approveRequest`'s `(3b-reset)` branch selected `{ appBlockId, kind: 'onsite', status: 'pending' }` and never asked which queue owned the review. Reachable chain: owner unpublishes (listing `removed`, block `suspended`, baseline recorded) → swaps the icon/cover/screenshots and republishes → the assets differ, so the listing goes to `pending` with a pending **listing** request and no block request → owner submits an ordinary new app **version**, which nothing stopped → a moderator approves *that* — a bundle/code review whose modal draws ZIP-extracted screenshots and shows nothing about listing imagery — and `(3b-reset)` flipped the listing to `approved` and un-suspended the block. **Never-reviewed store imagery goes live** — the exact invariant this PR exists to hold, reopened from the other side — **and the listing request is stranded**: `approveExternalRequest`'s `{draft,pending}` listing guard then matches 0 rows, and `routeRepublishToReviewInTx`'s open-review pre-check refuses every future republish with "A review is already pending for this listing."

It now asks. The predicate is spelled **once**, as a relation filter shared by the pre-read and the write:

```ts
const resetListingWhere = {
  appBlockId, kind: 'onsite', status: 'pending',
  publishRequests: { none: { status: 'pending' } },
} satisfies Prisma.AppListingWhereInput;
```

Sharing it is not tidiness — a separate probe leaves a TOCTOU window (the republish transaction committing between the check and the write) which is precisely the hole being closed. It is scoped through the relation (i.e. `appListingId`), **not** by slug, because a pending shadow media revision denormalizes the parent slug onto its request row while targeting a different listing, and that is a legitimate concurrent review. **A discriminator, not a relaxation:** the mod-reset case has no listing request and still restores; nothing here decides whether an owner may self-restore.

**`submitVersion` additionally refuses while a non-shadow listing review is open for the slug** — and this is labelled a *reachability reduction, not the guarantee*, in the code as well as here. It costs the owner nothing they could have had anyway: the review arm deliberately leaves the block `suspended`, and the subsequent-version approve branch refreshes manifest/version without touching `AppBlock.status`, so a version submitted in that window could not serve until the imagery review resolved regardless. What it cannot do is restore the invariant, because it is not the only producer — `recordPendingFromPush` mints a pending block request straight off a git push without passing it, and submitting the version **before** the republish never reaches it at all. Scoped to `revisionOfId: null` so a shadow media revision, which has always coexisted with a code submit, still does not block a release.

**Why not restore the exclusion at the source instead?** Because a one-sided guard demonstrably cannot: the reverse ordering reaches the same state. Restoring it properly needs a symmetric refusal in `routeRepublishToReviewInTx` as well, which would refuse an owner's one-button republish while a code review is open — a previously-working path — and would re-introduce the block-request dependency this PR deliberately removed from that arm. `(3b-reset)`'s discriminator holds for **every** ordering and every producer, because it asks the question at the point of consumption. The `submitVersion` guard is kept for state-space reduction and a clear error message, not leaned on.

**What the enumeration found, and what it did not.** Every reader and writer of both request tables, every "a review is already pending" predicate, every `AppListing.status` branch and every `AppBlock.status === 'suspended'` consumer was enumerated (`command grep -rn` over `src/`, not sampled). No read path anywhere branches on *why* a block is suspended. `listPendingOffsiteRequests` is kind-agnostic, so the new row does surface in `/apps/review`. The residue is informational, not an exposure: `getMyPendingForSlug`, `blocks.router.listMyPublishRequests` and `GET /api/v1/blocks/submissions` all read the block table only, so an owner whose sole open review is the imagery one sees nothing in those three surfaces — `listListingHistory` merges both queues and is the one place they can see and act on either. Those are recorded under "Known limitations" rather than fixed here.

**A version withdraw no longer delists a listing whose review lives elsewhere.** (Unchanged from the previous round; kept because it is the sibling half.) `closeOnsiteResetListingOnWithdraw` reasoned from a slug and a listing status for the same reason. It now refuses to act when a pending listing request owns the row — again a discriminator, not a relaxation. Its comment used to say the owner could reach that state because "nothing stops them, since the one-pending-per-slug index is empty"; that is now only partly true and the comment says which parts.

**Withdrawing your own re-review is still one-way — and now it says so.** `closeTerminalListing` unconditionally writes `delist` on a formerly-live `pending` listing, which `republishOwnListing`'s guard reads as a moderator takedown. That predicate was *made* unconditional to close a real exploit (an intervening `report-resolve` shifted the newest event and let an owner self-restore mod-mandated content), so it is untouched. What changed is disclosure: `withdrawExternalRequest` now returns the close **outcome**, the router passes it through, and the Withdraw affordance warns before the click and reports accurately after it, instead of saying "Submission withdrawn." and leaving the owner looking at a "removed by a moderator" state they caused themselves.

## Three claims this delta falsified, now corrected

- **The `REVIEWABLE_LISTING_KINDS` invariant.** It read *"the ONLY producer of a `kind='onsite'` `AppListingPublishRequest` is `submitListingRevision` on an onsite SHADOW"* — false as of this PR, which introduces the first non-shadow one. Six consumers reason from that sentence and it is what justified `listMySubmissions`' unconditional `{ kind: 'onsite' }` OR-branch. The test named for it stayed green throughout, because it only asserted that `submitListingRevision` produces a well-shaped row — proving one member is IN a set says nothing about whether the set grew. The comment is corrected (two producers, told apart by `revisionOfId`, never by `kind`) and a **producer ledger** now enumerates every production site that inserts a row into `app_listing_publish_requests`, with its kind decision, and fails when the set **grows or shrinks** — both directions watched to fail. It scans all three write routes the codebase actually has — the Prisma client (`create` / `createMany` / `upsert`), raw SQL (`INSERT INTO app_listing_publish_requests`, covering `$executeRaw` / `$queryRaw` / `$executeRawUnsafe`) and Kysely (`.insertInto('…')`) — across `src/`, `apps/`, `packages/` and `scripts/`. It first matched `appListingPublishRequest.create(` in `src/` alone, which was blind to two of those three: a `createMany` producer and a raw-SQL producer each left the suite fully green. A guard whose whole purpose is to notice the set growing is worse than no guard when it reads as coverage and provides none, so the scan was widened to the sentence rather than the sentence narrowed to the scan. `listMySubmissions`' OR-branch is still correct (the middle branch already matches a non-shadow row and `OR` is a set union); its *justification* was not, and the downstream consequences — `hasPendingRevision` correctly `false`, no last-action badge while `pending` — are now stated rather than re-derivable from a false premise.
- **The republish fixture docstring.** It claimed the gate fails CLOSED on an absent baseline via `reviewReason: 'no-recorded-assets'`. Both halves were false: no such reason exists (`RepublishReviewReason` is `'assets-changed' | 'unreadable-baseline'`) and an absent baseline returns `null`, i.e. the **immediate** arm. So a forgotten baseline now produces a silent pass, not a loud re-route — a fixture meaning to prove that *matching* assets republish immediately gets `approved` while comparing nothing. Corrected. The accompanying claim that no fixture is *already* passing vacuously was checked by execution, not by reading: making the `absent` arm throw fired on exactly **2** tests, both in the dedicated absent-arm file, and **0** in the fixture file — with the 2 serving as the positive control that the probe can fire at all.
- **The moderator modal stated the opposite rating policy.** `OffsiteReviewQueue` branched on `kind === 'onsite'` and told the reviewer the app's rating is a **CAP** and mature imagery is a reject reason. For this new request kind the server **FLOORS**, raise-only. `kind === 'onsite'` no longer implies "media revision", so the explainer note, the `app rating (cap)` label, the mismatch callout and the approve control's default now branch on `revisionOfId`, and the default mirrors the server's `max(declared, derived)` instead of offering a value the server would silently raise past. The on-site revision browser fixture gained the `revisionOfId` that a real shadow request always carries — its absence was what made the two shapes indistinguishable in the first place.

## Also in here

- **`republishSuccessMessage`** — three surfaces announced a republish with a hardcoded *"App republished — it is live again."*, which is **false** on the review arm. They now derive the message from the returned `status` in one place, **and branch on `reviewReason`**: only `assets-changed` claims the owner's images changed. An unrecognised reason falls back to the neutral wording, so a reason added server-side can never silently inherit a factual claim it does not support.
- **`withdrawSuccessMessage`** — extracted as a pure function so the one-way-close copy is covered by the **blocking** `unit` project rather than the report-only browser tier.
- `RepublishOwnListingResult.status` is widened to `'approved' | 'pending'` with a `reviewReason`, so a caller cannot assume success means live.
- **Three comments that were false in the direction that invites the bug**, all predating this PR: `offsite-listing.service` claimed `republishOwnListing` runs no rating floor (untrue since #4418, and it reads as a licence to skip the floor elsewhere on that path — exactly the hole fixed here); `app-listing-assets.service` claimed "draft/pending/shadow listings are rated at approve" (untrue for the on-site `pending` re-approve); and the two withdraw-close comments naming a writer set that is no longer complete.

## Verification

**`pnpm typecheck`** — `typecheck: OK — 0 type errors in 64s (heap cap 8192 MB)`.
Instrument control, re-run this round: a deliberate `const __probe: number = 'not a number'` inside the `(3b-reset)` branch produced `publish-request.service.ts(2989,11): error TS2322` / `typecheck: 1 type error(s).` and exit **2**, then was reverted. (Run `pnpm typecheck`, not bare `tsc --noEmit` — the latter OOMs at the 4 GB default and exits 134 with zero errors.)

**`typecheck-tests` gate — the scope `pnpm typecheck` structurally cannot reach.** The root `tsconfig.json` carries `src/**/__tests__/**` in its `exclude`, so every "0 type errors" above is a claim about **production code only**: a type error in any file this PR adds or edits under a `__tests__/` directory is reported by `pnpm typecheck` as 0 errors, exit 0. That is not leniency, it is blindness, and every typecheck claim made on earlier revisions of this description was scoped that way. The instrument that does see those files is `node scripts/ci/typecheck-tests-gate.mjs`, which runs the same program against `tsconfig.tests.json` (the base config minus that one exclude entry) and blocks a file that is newly dirty or a baselined file whose count went up.

Run on three trees with the same command. The gate is already red at `main` for ~187 unrelated files, so the question is not "does it pass" but "does this PR make it worse" — answered by diffing the two flagged sets file-by-file, since the branch is a few commits behind `main` and the absolute totals are therefore not comparable.

| tree | gate-scope errors / files | flagged: new / worsened |
|---|---|---|
| `origin/main` (`570b7b01f5`) | 1,008 / 186 | 64 / 7 |
| this branch, before the fix | 1,027 / 188 | 66 / 9 |
| **this branch, after** | **997 / 185** | **63 / 7** |

Before the fix this PR contributed **five** entries `main` does not have — three files it adds, and two it worsens:

| file | at `main` | before | after |
|---|---|---|---|
| `app-listings.router.offsite-authz.test.ts` | 3 | 4 | **3** |
| `publish-request.orchestration.test.ts` | 15 | 29 | **5** |
| `app-listing-approved-assets.test.ts` | *(new file)* | 1 | **0** |
| `offsite-listing.onsite-listing-approve.service.test.ts` | *(new file)* | 1 | **0** |
| `offsite-moderation.service.republish-asset-review.test.ts` | *(new file)* | 3 | **0** |

Diffing the flagged sets after the fix leaves **zero** entries on this branch that `main` does not already have (and one fewer, because `main` has since dirtied a file this branch has not picked up). `scripts/ci/typecheck-tests-baseline.json` is **not** touched — raising a baseline entry is the escape hatch, and none was taken. The five surviving errors in `publish-request.orchestration.test.ts` are pre-existing `TS2737` BigInt literals (`target: ES2018`), unrelated to this PR and unchanged from `main`; the other ten it carried at `main` are fixed as a side effect.

**What the errors turned out to be — one of them was a fixture that did not match reality.** The `withdrawExternalRequest` router mock declared its return as `{ outcome: 'none' as const }`, i.e. a service that can only ever report `'none'`, while the real `WithdrawExternalRequestResult` is `{ outcome: CloseTerminalListingOutcome }` (`'deleted' | 'removed' | 'none'`). The `'removed'` pass-through — the whole point of returning the outcome — was the case the mock's own type said could not happen. It is now typed with the service's exported result type. The rest were mechanical: a `vi.fn(async () => null)` DB-delegate default infers `() => Promise<null>` (no argument, only ever `null`), which makes a later `mockResolvedValue(row)` on the same delegate untypeable; those now carry the repo-standard delegate signature. No `as any`, no `@ts-expect-error`, no assertion removed — every edit is type-annotation-only, and transpiling each file before and after with comments stripped emits **byte-identical** JavaScript for four of the five (the fifth differs only by one added unused parameter name).

**Tests — read as counts, both tiers**

| tier | result |
|---|---|
| `--project 'unit*'` (covers `unit` **and** `unit-native`) | **23,049 passed**, 25 skipped, 1,469 files, **23 failed** (re-measured after the typing round; the 23 are the same three localStorage files) |
| `--project unit-native` alone | **141 passed**, 6 files, **0 failed** |
| the seven block suites this touches | **418 passed**, 7 files, **0 failed** |
| `--project component` over `src/components/Apps/` | **1,098 passed**, 75 files, **0 failed** |

🔴 **The 23 failures are RED AT `main`, and that was measured rather than assumed.** They are `hiddenBlocks`, `recent-source-images.store` and `sticker-reveal.store` — all `Cannot read properties of undefined (reading 'setItem')`, i.e. a `localStorage` environment problem, in files no part of this PR touches. Running those same three files in a clean checkout of `main` (`b9e8e132d3`) produced **the identical 23 failures** — and CI settles it in the other direction too: every `Unit tests` shard on this PR is **green**, so the 23 are a property of this workstation's jsdom `localStorage`, not of the tree. The local baseline is reported anyway rather than quietly dropped, because "CI is green" would not have told you *why* a local run was red. `eventloop-watchdog.capture` also failed in one of the three full runs and passed in the others; it is likewise **red at `main`** in isolation, so it is a pre-existing flake, not this branch's.

⚠️ **Correction to an earlier revision of this description**, which claimed `--project unit` was "22,916 passed, 0 failed": that number predates `main` moving. The zero was true when written and is not true now, and the cause is upstream, not here.

**Scope note, stated rather than implied:** the browser tier was run over `src/components/Apps/` (the directory this PR touches), **not** the whole `component` project. So the previously-recorded 17 `ModelVersionUpsertForm` failures were neither reproduced nor refuted this round.

**`eslint`** over the ten changed files: **0 errors, 41 warnings** — all `no-explicit-any` / `no-unused-vars` of the same kinds already present in those files.

**Red at base / green at HEAD.** Every source guard added this round was watched to fail on pre-change code: see the mutation table, where the `N`/`S`/`U` rows ARE the pre-change state (each is the guard removed or re-spelled).

## Verification — the mutation table

Each row is a single isolated edit, applied one at a time to an otherwise-clean tree, restored before the next. Every run's **collected test count** is read, never an exit code, and every unmutated control was watched to report zero failures over a non-zero collection first.

### New this round — the 🔴 chain and its two fixes

| # | mutation | killed by | count |
|---|---|---|---|
| N0 | *(unmutated control)* | — | 121 passed / 0 failed |
| N1 | `(3b-reset)`: delete `publishRequests: { none: … }` from the shared where | `🔴 (3b-reset) does NOT restore a pending on-site listing whose review belongs to the LISTING queue` | 1 failed / 116 |
| N2 | `(3b-reset)`: `none` → `some` | the same test **and** `🔴 (3b-reset) DOES restore a mod-reset listing` | 2 failed / 115 |
| N3 | `(3b-reset)`: drop `status: 'pending'` from the relation filter | `🔴 (3b-reset) DOES restore a mod-reset listing …` (the control arm, by design) | 1 failed / 116 |
| S1 | `submitVersion`: delete the cross-queue guard | `🔴 refuses a new version while a NON-SHADOW on-site listing review is open for the slug` | 1 failed / 119 |
| S2 | `submitVersion`: drop `appListing: { revisionOfId: null }` | `🔴 does NOT refuse while only a SHADOW media revision is open …` | 1 failed / 119 |
| S3 | `submitVersion`: drop `status: 'pending'` | `does NOT refuse on a CLOSED listing review — one completed cycle must not block every future version` | 1 failed / 119 |
| U1 | modal: `isOnsiteRepublishReview = isOnsite` (drop the discriminator) | the two on-site **media-revision** cases | 2 failed / 23 |
| U2 | modal: invert the discriminator to `!= null` | all four republish cases **and** the two revision cases | 6 failed / 19 |
| U3 | modal: approve default `= appRating ?? derivedRating` (unfloored) | `🔴 the approve Select defaults to the DERIVED rating (the floor), not the app rating` | 1 failed / 24 |

### The three survivors from the previous round — now dead

| # | mutation | previously | killed by | count |
|---|---|---|---|---|
| X2 | withdraw discriminator: drop `status: 'pending'` | **SURVIVED** 115/115 | `🔴 a CLOSED listing request on the row does NOT stop a mod-reset withdraw from delisting` | 1 failed / 120 |
| X5 | open-review guard: `appListingId` → `slug: listing.slug` | **SURVIVED** 136/136 | `🔴 a pending SHADOW media revision on the SAME SLUG does NOT block the republish` | 1 failed / 35 |
| X6 | open-review guard: drop `status: 'pending'` | **SURVIVED** 136/136 | `🔴 a CLOSED request on THIS listing does NOT block a later republish` | 1 failed / 35 |

**Why they survived and what changed.** All three tests armed the probe with a blanket `mockResolvedValue({ id: … })`, which answers the same row for *any* `where` — so the `where` was never under test. Each is now driven by a small fake table that filters on the clause, and each fixture is a **realistic** row rather than a textbook one: a *withdrawn* request for X2, a *shadow* request on the same slug for X5, an *approved* request on the same listing for X6.

### Producer-ledger controls (🟡 2)

| mutation | result |
|---|---|
| add a new `appListingPublishRequest.create(` producer to a source file | `🔴 the producer set is EXACTLY the ledger` **fails** (grows) |
| add an `appListingPublishRequest.createMany({…})` producer | `🔴 the producer set is EXACTLY the ledger` **fails** — was **green** before the widening |
| add a raw-SQL `$executeRaw` producer whose body is `INSERT INTO app_listing_publish_requests` | `🔴 the producer set is EXACTLY the ledger` **fails** — was **green** before the widening |
| add a `.insertInto('app_listing_publish_requests')` producer under `apps/` | `🔴 the producer set is EXACTLY the ledger` **fails** — proves the Kysely route *and* that the widened roots are walked |
| rename `routeRepublishToReviewInTx` away | `🔴 the producer set is EXACTLY the ledger` **fails** (shrinks), and so does the sibling guard `🔴 routeRepublishToReviewInTx really does forward the listing kind` |
| *(the ledger's own controls)* | 4 producers over 4,817 files; each root asserted walked; each write route fed source it must match (two have no real hits, so a route wired to nothing would otherwise be invisible); the same three producer lines assert **3 matches as bare code and 0 as comments**; bogus table names match 0 |

### Re-run after the fix (a fix round resets the gate)

Every previous-round mutant whose code or whose killing test this round touched was re-applied. All still die, each on its own named assertion:

| # | mutation | still killed by | count |
|---|---|---|---|
| M4 | publish-floor pre-check deleted | the two `actionable floor error … leaving it removed` cases | 2 failed / 34 |
| M5 | open-review guard deleted | `🔴 a review already open on THIS listing → NOT_TRANSITIONABLE` | 1 failed / 35 |
| M6 | re-queue `kind` hardcoded to `offsite` | `🔴 ON-SITE: re-queues on the LISTING surface …` | 1 failed / 35 |
| M7 | on-site rating branch swapped to the off-site discipline | `🔴 mature store art DOES raise it` | 1 failed / 8 |
| M8 | on-site override comparator flipped | `a moderator override may raise above the floor` + `may NOT lower it` | 2 failed / 7 |
| M9 | block un-suspend `where` status swapped | 3 reset-restore cases, incl. the new control arm | 3 failed / 118 |
| M10 | `(3b-reset)` rating floor dropped | `🔴 the reset restore FLOORS contentRating` + the raise-only case | 2 failed / 119 |
| M11 | `(3b-reset)` declared rating stale-rebound to `null` | the gate-passes case + the raise-only case | 2 failed / 119 |
| M12 | withdraw discriminator deleted | `🔴 does NOT close a pending onsite listing whose review belongs to the LISTING queue` | 1 failed / 120 |
| M13 | withdraw outcome replaced by a constant | `🔴 REPORTS WHICH CLOSE HAPPENED …` | 1 failed / **79 collected** |

**M1, M2, M3, M14 and M15 were NOT re-run**, and the reason is mechanical rather than a judgement call: their code (`app-listing-approved-assets.ts`, `listingPublishingActions.ts`) and their tests are **byte-unchanged** this round — `git status` lists neither — and both modules are pure functions with no mock surface that this round's edits could have moved. M13 *was* re-run despite living in a file whose diff is comments-only, because it is the one that previously scored a false SURVIVED off a zero-collection run; its re-run collected **80** tests.

### Re-run after the test-fixture typing round

The typing fix above edits the five files that hold these guards, so the gate resets again. The structural half first: transpiling each edited file before and after the change, with comments stripped, emits **byte-identical** JavaScript for four of the five — so those guards cannot have moved. The fifth (`app-listing-approved-assets.test.ts`) differs by exactly one token, an added unused parameter name on a mock.

The empirical half — one mutant per edited test file, each applied alone to an otherwise-clean tree and restored before the next, counts read rather than exit codes:

| # | mutation | still killed by | count |
|---|---|---|---|
| — | *(unmutated control, all five files)* | — | 256 passed / 0 failed, 5 files |
| N1 | `(3b-reset)`: delete `publishRequests: { none: … }` | `🔴 (3b-reset) does NOT restore a pending on-site listing whose review belongs to the LISTING queue` | 1 failed / 120 |
| S1 | `submitVersion`: disable the cross-queue guard | `🔴 refuses a new version while a NON-SHADOW on-site listing review is open for the slug` | 1 failed / 120 |
| M5 | `republishOwnListing`: disable the open-review guard | `🔴 a review already open on THIS listing → NOT_TRANSITIONABLE, and NOTHING is written` | 1 failed / 35 |
| M7 | approve: on-site rating branch swapped to the off-site discipline | `🔴 tame store art does NOT lower the app rating` **and** `🔴 a moderator override may NOT lower it below the app declaration` | 2 failed / 7 |
| R1 | router: collapse the withdraw `outcome` back to a constant | `🔴 passes the close OUTCOME through, rather than collapsing it into \`{ ok: true }\`` | 1 failed / 36 |
| A1 | `buildApprovedAssetSnapshot`: drop the `where: { appListingId }` scoping | `reads the screenshots and takes icon/cover from the ALREADY-LOADED listing row` | 1 failed / 52 |

A1 is the one that matters most for this round: it dies through the exact `findMany.mock.calls[0][0]` assertion whose fixture was re-typed, which is what rules out the typing having weakened it. R1 likewise kills through the `'removed'` pass-through case that the mock's old `as const` type had declared impossible.

## Known limitations, stated rather than glossed

- **The moderator still sees no before/after diff.** `RevisionDriftSection` renders only for a shadow revision (`if (!parentId) return null`) and these requests are non-shadow. Unchanged this round; the recorded baseline that decided the routing could drive such a diff, and wiring it in is a separate change.
- **`getMyPendingForSlug` does not see the listing queue.** It reads `AppBlockPublishRequest` only, so while an imagery review is open the `/apps/submit` preview tells the owner the slug is clear and the new `submitVersion` guard rejects after the upload. That is the same shape as the pre-existing other-user collision (also invisible to that read), and widening it changes a return type and a UI affordance. Closing condition: `getMyPendingForSlug` returns the listing review too, or the submit preview surfaces it — checked by a browser test that renders the affordance with only a listing review open.
- **The mutual exclusion is NOT restored, deliberately.** See the enumeration above; `(3b-reset)` is what carries the guarantee.

## Not verified

- **No end-to-end run against a real database.** The 🔴 chain was established by code reading and then **exercised through the service layer against mocked Prisma** — the new regression test drives `approveRequest` end to end with a fake `app_listings` table and fails on pre-fix code. What is still inferred rather than observed is the SQL Prisma emits for the `publishRequests: { none: … }` relation filter inside an `updateMany`, and the re-queued request's appearance in the live moderator queue. The relation-filter-in-a-bulk-write construct is not novel here, which is weaker than executing it but better than nothing: `user.service.ts` already ships `dbWrite.bountyEntry.deleteMany({ where: { userId: id, benefactors: { none: {} } } })` on the account-deletion path.
- **`nsfwLevelFromContentRating` collides `'g'` and `'pg'`** (🟢6) — recorded in a comment, not fixed. The on-site override comparator therefore drops a moderator's explicit `'pg'` over a floored `'g'`, while the off-site clamp honours it. Label-only: both map to `NsfwLevel.PG`, so nothing about who can see the listing changes. Deciding that a tie-breaking override should overwrite an author's declaration is a product call. Closing condition: a product decision on tie-breaking, recorded in the comment or implemented with a mutation-checked test.
- **The rejection notification does not say the listing was delisted** (🟢10) — the `revisionOfId == null` gate no longer means "first-time submission", but the copy it fires (`"<app> was not approved: <reason>"`) never claimed it did, so the comment was corrected and the behaviour left alone. Distinguishing "your first submission was rejected" from "your republish was rejected and the listing is now off the store pending a moderator" needs a second notification type. Closing condition: a `app-listing-republish-rejected` type shipped with its own `prepareMessage` test, or an explicit decision that one message serves both.
- **The one-way withdraw warning still shows on version entries** (🟢9) — behaviour unchanged, comment corrected. Post-discriminator a version withdraw delists in the mod-reset case and does not in the republish-review case, and which one you are in depends on a row `ListingHistoryPanel` does not read. Warning in both is the safe direction for a one-way action; claiming either outcome per entry would be an assertion that surface cannot support.

**Overlap with #4437** (`zach/message-app-owner-ui`): still none — its **12** files and this PR's **30** are disjoint (`comm -12` over both `--name-only` lists returns empty), so no test-merge was needed.

### DIFF
--- a/src/components/Apps/ListingHistoryPanel.tsx
+++ b/src/components/Apps/ListingHistoryPanel.tsx
@@ -1,6 +1,7 @@
 import { Alert, Badge, Button, Group, Loader, Stack, Text } from '@mantine/core';
 import { useCallback } from 'react';
 
+import { withdrawSuccessMessage } from '~/components/Apps/listingPublishingActions';
 import { historyStatusColor } from '~/components/Apps/myAppsView';
 import { useFeatureFlags } from '~/providers/FeatureFlagsProvider';
 import { formatDate } from '~/utils/date-helpers';
@@ -174,6 +175,19 @@ export function ListingHistoryPanelView({
               color="gray"
               disabled={withdrawing}
               onClick={() => onWithdraw(e)}
+              /*
+                🔴 THE ONE-WAY WARNING IS ON THE CONTROL, not only in the toast that
+                follows it. Withdrawing the review of a listing that was previously LIVE
+                does not return it to how it was: the server closes it to `removed` behind
+                a `delist` event, which the owner-republish guard reads as a moderator
+                takedown, so only a moderator can put it back. That is deliberate (it
+                closes a self-restore exploit — see `closeTerminalListing`), which is
+                exactly why it has to be disclosed BEFORE the click rather than defended
+                afterwards. Worded for the case it warns about without asserting the
+                listing IS in it — this component cannot tell, and a warning that
+                over-claims gets ignored.
+              */
+              title="Withdraws this submission. If the listing was previously live, withdrawing takes it off the store and a moderator has to restore it."
               data-testid={`apps-history-withdraw-${e.id}`}
             >
               Withdraw
@@ -236,8 +250,16 @@ export function ListingHistoryPanel({ appListingId }: { appListingId: string }) 
     onError: (e) => onWithdrawError(e.message),
   });
   const withdrawListing = trpc.appListings.withdrawExternalRequest.useMutation({
-    onSuccess: () => {
-      showSuccessNotification({ message: 'Submission withdrawn.' });
+    /**
+     * 🔴 THE SERVER SAYS WHAT IT DID; DO NOT ASSUME. `outcome: 'removed'` means this
+     * withdraw closed the review of a formerly-LIVE listing, so the listing is now OFF the
+     * store behind a `delist` and the owner cannot republish it — a moderator must relist.
+     * `'deleted'` merely discarded a draft. Announcing both as "Submission withdrawn."
+     * left an owner looking at a "removed by a moderator" state they had caused
+     * themselves, with nothing having told them it would happen.
+     */
+    onSuccess: (data) => {
+      showSuccessNotification({ message: withdrawSuccessMessage(data?.outcome) });
       refetchHistory();
     },
     onError: (e) => onWithdrawError(e.message),
--- a/src/components/Apps/ListingHistoryPanel.tsx
+++ b/src/components/Apps/ListingHistoryPanel.tsx
@@ -186,6 +186,16 @@ export function ListingHistoryPanelView({
                 afterwards. Worded for the case it warns about without asserting the
                 listing IS in it — this component cannot tell, and a warning that
                 over-claims gets ignored.
+
+                🔴 THE VERSION ENTRIES (`source === 'version'`) ARE NOW CONDITIONAL, which
+                is why the hedged wording has to stay rather than be sharpened. Withdrawing
+                a VERSION reaches `closeOnsiteResetListingOnWithdraw`, which since the
+                asset-review route REFUSES to close a `pending` listing whose review
+                belongs to the LISTING queue. So a version withdraw delists in the
+                mod-reset case and does not in the republish-review case — and which case
+                you are in depends on a row this component does not read. Warning in both
+                is the safe direction for a one-way action; claiming either outcome
+                per-entry would be an assertion this surface cannot support.
               */
               title="Withdraws this submission. If the listing was previously live, withdrawing takes it off the store and a moderator has to restore it."
               data-testid={`apps-history-withdraw-${e.id}`}
--- a/src/components/Apps/ListingPublishingPanel.browser.test.tsx
+++ b/src/components/Apps/ListingPublishingPanel.browser.test.tsx
@@ -9,6 +9,7 @@ import {
   OWNER_ACTIONS_BY_STATE,
   sortPublishingActions,
 } from '~/components/Apps/listingPublishingActions';
+import { showSuccessNotification } from '~/utils/notifications';
 
 /**
  * THE OWNER PUBLISHING LEDGER, now on the authoring page's **Publishing** tab.
@@ -57,6 +58,15 @@ const mocks = vi.hoisted(() => ({
   /** `[procedureName, input]` for every mutation fired, in order. */
   calls: [] as Array<[string, unknown]>,
   republishPending: false,
+  /**
+   * What `republishOwnListing` resolves to. `'pending'` is a REAL success outcome (the
+   * asset-change review route), not an error, and the panel must say so — see
+   * `republishSuccessMessage`.
+   */
+  republishResult: { appListingId: 'apl_1', status: 'approved' } as {
+    appListingId: string;
+    status: 'approved' | 'pending';
+  },
   /** Every `utils.appListings.<proc>.invalidate()` the panel issues, by procedure name. */
   invalidated: [] as string[],
 }));
@@ -80,7 +90,12 @@ function mutationStub(name: string, isPending = false) {
     useMutation: (opts?: MutationOpts) => ({
       mutate: (input: unknown) => {
         mocks.calls.push([name, input]);
-        void opts?.onSuccess?.(undefined);
+        // 🔴 A REALISTIC payload, not `undefined`: `republishOwnListing` resolves to
+        // `{ appListingId, status }` and the panel DERIVES its success message from that
+        // status (a republish whose assets changed lands in `pending`, not live). A stub
+        // that hands the handler `undefined` is an unfaithful fake — it cannot see the
+        // wrong message and it crashes the handler that reads the field.
+        void opts?.onSuccess?.(mocks.republishResult);
       },
       isPending,
     }),
@@ -187,6 +202,7 @@ function renderedActions(): string[] {
 beforeEach(() => {
   mocks.calls = [];
   mocks.republishPending = false;
+  mocks.republishResult = { appListingId: 'apl_1', status: 'approved' };
   mocks.invalidated = [];
 });
 
@@ -340,6 +356,40 @@ describe('the wiring — each control fires the right procedure with the right i
     // fails in one of the two tests. Republish is not confirm-gated on purpose: it restores
     // the previous state, so the recovery from a misclick is the button next to it.
     expect(mocks.calls).toEqual([['republishOwnListing', { appListingId: 'apl_hidden_q2' }]]);
+  });
+
+  test('🔴 a republish routed to REVIEW tells the owner so, instead of claiming it is live', async () => {
+    /**
+     * 🔴 THE MESSAGE IS A CLAIM ABOUT THE SERVER'S ANSWER. `republishOwnListing` routes a
+     * republish to `pending` — a re-review — whenever the listing's assets changed since
+     * the last approval. The panel used to hardcode "App republished — it is live again",
+     * which is FALSE on that arm and survives review precisely because the mutation really
+     * did succeed. This drives the real click path and reads the notification the owner
+     * would actually see.
+     */
+    mocks.republishResult = { appListingId: 'apl_hidden_q2', status: 'pending' };
+    renderWithProviders(<ListingPublishingPanel {...OWNER_HIDDEN} />);
+    const button = page.getByTestId('apps-publishing-republish');
+    await expect.element(button).toBeInTheDocument();
+    await userEvent.click(button);
+
+    const message = vi.mocked(showSuccessNotification).mock.calls.at(-1)?.[0]?.message as string;
+    expect(message).toContain('review');
+    // The load-bearing ABSENCE — the old copy must not come back on this arm.
+    expect(message).not.toContain('live');
+  });
+
+  test('🔴 a republish that DID go live still says so (the other arm of the same branch)', async () => {
+    // Positive control on the same assertion: without this, a mutant that always returned
+    // the review wording would satisfy the test above.
+    mocks.republishResult = { appListingId: 'apl_hidden_q2', status: 'approved' };
+    renderWithProviders(<ListingPublishingPanel {...OWNER_HIDDEN} />);
+    const button = page.getByTestId('apps-publishing-republish');
+    await expect.element(button).toBeInTheDocument();
+    await userEvent.click(button);
+
+    const message = vi.mocked(showSuccessNotification).mock.calls.at(-1)?.[0]?.message as string;
+    expect(message).toContain('live');
   });
 
   test('a republish in flight DISABLES the button rather than allowing a second fire', async () => {
--- a/src/components/Apps/ListingPublishingPanel.tsx
+++ b/src/components/Apps/ListingPublishingPanel.tsx
@@ -8,6 +8,7 @@ import {
   type OwnerUnpublishVariant,
 } from '~/components/Apps/ownerListingModals';
 import {
+  republishSuccessMessage,
   showModRemovedNotice,
   showRepublish,
   showUnpublish,
@@ -88,8 +89,11 @@ export function ListingPublishingPanel({
   }, [utils, onChanged]);
 
   const republish = trpc.appListings.republishOwnListing.useMutation({
-    onSuccess: () => {
-      showSuccessNotification({ message: 'App republished — it is live again.' });
+    // 🔴 The message is DERIVED from the server's answer, never assumed: a republish whose
+    // assets changed since the last approval lands in `pending`, not live. See
+    // {@link republishSuccessMessage}.
+    onSuccess: (data) => {
+      showSuccessNotification({ message: republishSuccessMessage(data, kind) });
       refresh();
     },
     /**
--- a/src/components/Apps/MySubmissionsList.browser.test.tsx
+++ b/src/components/Apps/MySubmissionsList.browser.test.tsx
@@ -51,10 +51,12 @@ vi.mock('~/utils/trpc', () => {
   // notification paths run.
   const mutation =
     (name: string) =>
-    (opts?: { onSuccess?: () => void; onError?: (e: { message: string }) => void }) => ({
+    (opts?: { onSuccess?: (data?: unknown) => void; onError?: (e: { message: string }) => void }) => ({
       mutate: (vars: unknown) => {
         mocks.mutate(name, vars);
-        void opts?.onSuccess?.();
+        // Realistic mutation payload — the republish handler derives its message from
+        // the returned `status` (see `republishSuccessMessage`).
+        void opts?.onSuccess?.({ appListingId: 'apl_1', status: 'approved' });
       },
       isPending: false,
     });
--- a/src/components/Apps/MySubmissionsList.tsx
+++ b/src/components/Apps/MySubmissionsList.tsx
@@ -26,6 +26,7 @@ import {
 } from '~/components/Apps/ListingProblemsIndicator';
 import { getDetailPrimaryAction } from '~/components/Apps/appListingDetailView';
 import { isStaleDeploy, isStrandedDeploy } from '~/components/Apps/deploy-status';
+import { republishSuccessMessage } from '~/components/Apps/listingPublishingActions';
 import {
   canOwnerRepublish,
   canOwnerUnpublish,
@@ -556,8 +557,11 @@ export function MySubmissionsList({
   const invalidateSubmissions = () => utils.blocks.listMyPublishRequests.invalidate();
 
   const republishMutation = trpc.appListings.republishOwnListing.useMutation({
-    onSuccess: async () => {
-      showSuccessNotification({ message: 'App republished — it is live again.' });
+    // 🔴 The message is DERIVED from the server's answer, never assumed: a republish whose
+    // assets changed since the last approval lands in `pending`, not live. See
+    // {@link republishSuccessMessage}.
+    onSuccess: async (data) => {
+      showSuccessNotification({ message: republishSuccessMessage(data, 'onsite') });
       await invalidateSubmissions();
     },
     onError: (e) => showErrorNotification({ title: 'Republish failed', error: new Error(e.message) }),
--- a/src/components/Apps/OffsiteReviewQueue.browser.test.tsx
+++ b/src/components/Apps/OffsiteReviewQueue.browser.test.tsx
@@ -413,6 +413,12 @@ describe('OffsiteReviewModal — on-site listing-media revision (kind: onsite)',
       category: 'utility',
       contentRating: 'g',
       connectClientId: null,
+      // 🔴 A media revision request targets a SHADOW, so `revisionOfId` is always set.
+      // This fixture used to omit it, which made it indistinguishable from the OTHER
+      // on-site request shape (the republish re-review, `revisionOfId: null`) — the two
+      // are governed by opposite rating rules, so a fixture that cannot tell them apart
+      // was pinning the cap copy onto a request shape that may not be under a cap.
+      revisionOfId: 'listing-parent',
     },
     submittedBy: { id: 42, username: 'author-dev', image: null },
   };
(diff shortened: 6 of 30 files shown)


ZacxDev:
### CI note — `preview / component-tests` is red, and it is not this PR

17 failures, all in `src/components/Resource/Forms/ModelVersionUpsertForm.browser.test.tsx` (`Cannot find element with locator: getByRole('switch', { name: /I want to charge for this version/ })`). Nothing in that file's import graph is touched here — this PR only edits `src/components/Apps/**` and `src/server/services/blocks/**`.

**Control run, not an assumption.** I checked the same worktree out at `origin/main` (`1323e1ff91`, no changes from this PR present) and ran that one spec: **17 failed / 10 passed of 27** — the same 17, by name. Back on this branch the full component project is `1 failed | 191 passed (192)` files, `17 failed | 2114 passed (2131)` tests — i.e. exactly those 17 and nothing else. The three component files this PR does touch pass: `ListingPublishingPanel` 15/15, plus `MySubmissionsList` and `OffsiteSubmissionsList`, 69/69 together.

(#4437 is also red on this check, but that is a second sample rather than evidence — the base-checkout control above is what settles it.)

Every other check is green: App unit tests + typecheck, Unit tests 1–4, Package unit tests, ESLint + Prettier, Schema drift gate, event-engine-common pin, tekton/typecheck, tekton/fixture-bootstrap, preview deploy + render-check + auth-tests, Windows dev-env (bash + pwsh).

PR #4445, opened 2026-08-27
fix(remix-gallery): send the placer fields the manage modal renders

JustMaier:
@
The owner review queue in the remix gallery manage modal rendered every submitter as bare initials with no cosmetics.

## Cause

`getPendingRemixGallerySubmissions` selected `placer: { id, username, image }`. `UserAvatar` reads `profilePicture`, `cosmetics` and `deletedAt` — none were sent. It takes `Partial<UserWithCosmetics>`, so the narrow shape type-checked and nothing failed.

`User.image` is a dead column. Across all 121 distinct placers on this surface in production:

| | count |
|---|---|
| placers | 121 |
| with legacy `User.image` | 0 |
| with `profilePicture` | 93 |
| with equipped cosmetics | 61 |

So the avatar fell back to a column that is null for every placer alive, and 61 of them also lost their equipped frame and badges.

## Fix

`placer: { select: userWithCosmeticsSelect }` — the selector the rest of the app uses for this.

## Scope

This is the only `UserAvatar` consumer among the remix-gallery and placement components. The similar narrow selects nearby are not the same bug and are left alone: `sticker-placement.service.ts:1035` feeds a queue that renders `placer?.username` as text, `sticker-book.service.ts:212` feeds a grid that passes `userId=` and lets `UserAvatar` fetch the user itself, and `sticker-placement.service.ts:538` already uses this selector.

## Test

`sends the placer fields the avatar renders, not just id and username`, named for the decision because the narrow select type-checked. It asserts on the `findMany` argument rather than on a returned row — the Prisma mock ignores `select`, so an assertion on the result would pass with any select at all.

Positive control: reverting the select to `{ id, username, image }` fails it in ~0.5s with `AssertionError: placer select is missing profilePicture: expected { id: true, username: true, ...(1) } to have property "profilePicture"`, with the other 142 in the file still passing.

## Verification

- `pnpm run typecheck` -> `OK - 0 type errors in 196s`, exit 0
- `pnpm run lint` -> exit 1 with 360 pre-existing errors repo-wide; zero on either file here
- `pnpm run test:unit:run` -> `Test Files 1465 passed | 1 skipped (1466)`, `Tests 22949 passed | 28 skipped (22977)`, exit 0
- `blocks.router.workflow.test.ts` collected 358 tests, so the submodule was initialised

## Known, not fixed here

`RemixGallerySettings.tsx:52` calls this procedure with `{}` only to draw a count badge, so it now also pays the profile-picture join and the cosmetics subquery for 50 users it never renders. That call already fetches a full page of rows, raw-queries the submission and host images, loads the placement config and computes per-row earnings, all discarded for a number — the right fix is a count procedure, not a narrower select here.

@

### DIFF
--- a/src/server/services/__tests__/remix-gallery.service.test.ts
+++ b/src/server/services/__tests__/remix-gallery.service.test.ts
@@ -1438,6 +1438,40 @@ describe('getPendingRemixGallerySubmissions paging', () => {
     placementFindMany.mockResolvedValue([]);
   });
 
+  /**
+   * Named for the decision, because the narrow select this replaced type-checked
+   * fine: `UserAvatar` takes `Partial<UserWithCosmetics>`, so dropping the three
+   * keys below is invisible to tsc and shows up only as a wrong-looking avatar.
+   *
+   * `User.image` is what the queue used to send, and it is a dead column — of
+   * the 121 distinct placers on this surface in production, 0 have it set and 93
+   * have a `profilePicture`. So the old select handed the avatar null for every
+   * placer alive, and stripped the equipped cosmetics of 61 of them.
+   *
+   * Asserted on the ARGUMENT rather than on a returned row: the Prisma mock
+   * ignores `select` and hands back whatever the fixture holds, so an assertion
+   * made on the result passes with any select at all.
+   */
+  it('sends the placer fields the avatar renders, not just id and username', async () => {
+    // A populated page, because the service returns before it ever reaches
+    // `findMany` on an empty one — the default `beforeEach` state would make
+    // this assert against a call that never happened.
+    respond({
+      page: [{ id: 1, createdAt: new Date('2026-01-01T00:00:00.000Z') }],
+      images: [image(101), image(HOST_IMAGE)],
+    });
+    placementFindMany.mockResolvedValue([queueRow(1, 101, '2026-01-01T00:00:00.000Z')]);
+
+    await ask({ limit: 2 });
+
+    const placerSelect = placementFindMany.mock.calls[0][0].select.placer.select;
+    // Named individually so a revert says which one went missing. `deletedAt`
+    // is in here because the avatar and the username both branch on it and a
+    // deleted placer would otherwise render as a live one.
+    for (const field of ['profilePicture', 'cosmetics', 'deletedAt'])
+      expect(placerSelect, `placer select is missing ${field}`).toHaveProperty(field);
+  });
+
   it('takes the cursor from the last row of the page, not the last row it returns', async () => {
     // Three rows for a page of two: the third is the "is there more" probe. Row
     // 2 is selected but dropped afterwards for an unreadable payload — exactly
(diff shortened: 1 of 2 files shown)


JustMaier:
Five-lane review run (reuse / correctness / perf / tests / intent) against `a78723dd92`. One finding, fixed in `e6b887508c`.

**The finding — the test was existence-only.** `toHaveProperty('cosmetics')` passes on `cosmetics: true`, which returns `UserCosmetic` scalars with no nested `cosmetic` relation. Since `cosmetics` feeds `<Username>` as well as `UserAvatar`'s decoration lookup, that mutation silently drops three cosmetic classes — frame, nameplate and badge — while the test stays green. That is the same class of bug this PR exists to fix, so the guard was not guarding.

Fixed by adding `expect(placerSelect).toEqual(userWithCosmeticsSelect)` beside the per-field loop rather than replacing it: the loop keeps the legible "missing profilePicture" message on the common revert, and the identity check turns a spelling (three names, today's delta) into a meaning (the queue sends what the avatar's canonical selector sends). Compared against the imported object, so it cannot drift from production.

Control for the new assertion specifically — planting `{ id, username, image, profilePicture: true, cosmetics: true, deletedAt: true }`: the loop **passed**, only the new assertion failed, and the diff named the missing nested `cosmetic` select and the `equippedAt`/`equippedToId` where-clause. Restored byte-identical afterwards.

**No findings from the other four.**

- *correctness*: every widened field already ships to arbitrary third parties for the same user via the identical selector (`imagev2.selector.ts:18,56,89`, `commentv2.selector.ts:22`, `post.selector.ts:66`, `model.selector.ts:213`, `article.selector.ts:36`). Nothing crosses to a third party for the first time here.
- *perf*: `relationJoins` is off, so this is +3 `IN`-batched statements, constant per call, page size independent — not an N+1. Equipped cosmetics per user across prod: avg 1.41, p99 4, max 6. `UserCosmetic` is a PK-prefix scan (3.5ms/1522 buffers at the 50 heaviest users in the database); `profilePictureId` is 50 PK probes at 3.1ms.
- *reuse*: `userWithCosmeticsSelect` is correct — `simpleUserSelect` would have half-fixed it (no cosmetics). This makes the remix queue agree with `sticker-placement.service.ts:538` rather than inventing a shape.
- *intent*: both symptoms share one cause and it is the one reported; cosmetic parity is exact, not narrower.

**Re-verified after the fix**: typecheck `OK — 0 type errors`; eslint on both files `0 errors, 2 warnings` (pre-existing `AnyFn`, shifted 109→110 by the added import); `Test Files 1465 passed | 1 skipped (1466)`, `Tests 22949 passed | 28 skipped (22977)`, exit 0.

**Pre-existing, deliberately not fixed here** (raised separately rather than filed blind): `RemixGallerySettings.tsx` hand-rolls `QueueCountBadge`, `pendingCount()` and a count that already exists on the notifications payload; and three functions answer "how many submissions is this owner sitting on" with `getPendingPlacementCounts` still missing the listability filter the other two have.

PR #4448, opened 2026-08-27
fix(apps): one container for every /apps route, so the sub-nav stops moving

ZacxDev:
## The mechanism

`AppsPageLayout` rendered the shared `/apps` chrome — the `AppsSubNav` tab strip — **inside** the per-page `<Container>`:

```
<Container size={size}>     <- per-page width
  <Stack gap="xl">
    <Stack gap="md">
      <AppsSubNav />        <- shared chrome, inheriting the PAGE's width
```

`size` came from `APPS_PAGE_WIDTHS`, which deliberately assigned four different widths. So the one element that is supposed to be identical on every apps page inherited each page's content width and jumped horizontally as you navigated.

## Measured, before

Real Chromium render of `AppsPageLayout`, `nav[aria-label="App sections"].getBoundingClientRect()`:

| route | size | left @1440 | width @1440 | left @2560 | width @2560 |
|---|---|---|---|---|---|
| `/apps` | 1920 | 16 | 1408 | 336 | 1888 |
| `/apps/review` | 1400 | 36 | 1368 | 596 | 1368 |
| `/apps/store-preview/[slug]` | 1320 | 76 | 1288 | 636 | 1288 |
| `/apps/submit` | 1100 | 186 | 1068 | 746 | 1068 |

170px of left-edge spread at 1440; 410px at 2560.

## Measured, after

Every one of the 13 chrome routes reports **16 / 1408** at 1440 and **336 / 1888** at 2560. Asserted against those literals, not merely "they all agree" — see the harness note below.

## The change

- `APPS_PAGE_WIDTHS` (a per-route **container** width) becomes `APPS_PAGE_MEASURES` (a per-route **content** measure). `APPS_PAGE_CONTAINER_WIDTH = 1920` is the one container width for every route.
- `AppsPageLayout` drops `size` entirely and gains `measure?: number`, applied as a left-aligned `<Box maw={measure}>` around `{children}` **only** — below the chrome, never around it.
- The band structure is untouched: `Stack gap="xl"` → `Stack gap="md"` → `{children}`, `pb`-only padding. The 16px-in / 32px-out proximity pair and every vertical number in `AppsPageLayout.geometry.browser.test.tsx` are unchanged and that file is **unedited**.

### The +32px decision: measures are 1068 / 1368 / 1288, not 1100 / 1400 / 1320

Mantine's `Container` is border-box, so `size={1100}` renders **1068px** of content. A `maw` box sits *inside* that padding, so `maw={1100}` would render **1100** — 32px wider than before.

I chose to **preserve exact content widths**: each measure is the old container width minus the 2×16px gutter. Every page's rendered content width is byte-identical to before, so this is a pure alignment change with no confounded width delta — any visual regression report can be attributed to the alignment alone. Pinned in `appsPageWidths.test.ts` with the old numbers named as literals so the provenance is checkable.

A side benefit: `APPS_TWO_COLUMN_DETAIL_MEASURE`'s documented justification is "the same width as the model detail page", which renders `<Container size="xl">` = 1320 border-box = **1288** content. Stating it in content terms makes that claim true rather than 32px off.

## Four orphan pages adopted

These rendered a bare `<Container>` and showed **no sub-nav at all**:

- `/apps/get-started`
- `/apps/[appBlockId]/edit`
- `/apps/[appBlockId]/revenue`
- `/apps/listing/[appListingId]/edit` (two `<Container>` sites — loading and ready states; both converted)

### One caveat, deliberately surfaced rather than forced

`/apps/get-started` gates on `appBlocksGetStarted` **alone**, while `AppsSubNav` renders only for `hasAppsStoreAccess` (`appListings || appBlocks`). Those are independent flags, so a viewer could hold the page's flag and not the nav's, and get an empty chrome band.

**Not reachable today**: `appBlocksGetStarted` is staged mod-only, and a moderator holds `appBlocks` and is an `isAppDeveloper`, so the bar clears its own two-tab floor. It becomes reachable when that flag widens to `['public']`. Recorded as a `TODO(launch)` at the call site.

## Why the existing guards missed this

- `appsPageWidths.test.ts` enforced completeness, consumption, and "one of exactly four classes" — none of which is a relationship **between** pages.
- `AppsPageLayout.geometry.browser.test.tsx` has a test named "HORIZONTAL geometry is byte-identical" that renders **one** layout at the **default** size and never varies it — structurally blind to the `size` dimension.

## Tests changed, and why

**New — `AppsPageLayout.chromeAlignment.browser.test.tsx`.** Renders every one of the 13 chrome routes at 1440 and 2560 and asserts nav `left` AND `width` are identical across all of them. Notes:
- Route set is *derived* from the module and then *pinned as a literal*, so it fails when the set grows **or** shrinks. A loop over a silently-emptied set passes vacuously.
- It imports `@mantine/core/styles.css`. Without it the Container computes `max-width: none` / `padding-left: 0`, every route measures `0` **identically**, and "they all agree" passes while measuring nothing. A `styleSheetLoaded` guard is the first assertion in every test, and the agreed-upon value is pinned to a literal as a positive control.
- A second test asserts the body still takes its measure and is **left-aligned** (body left == nav left on every route). Without it, a layout that ignored `measure` entirely would satisfy the alignment test perfectly.

**New — blocking-tier source guards.** The browser project is report-only in CI, so it cannot block a regression. In `appsPageLayout.test.ts`: `AppsPageLayout` accepts no container-width prop, in every spelling. In `appsPageWidths.test.ts`: **every rendering `/apps` page imports AND mounts `AppsPageLayout`**, and none hand-rolls its own top-level `<Container>` — this is the guard that catches the four orphans and any new page. Both walks assert a non-zero file count.

**Ported — `appsPageWidths.test.ts`.** Keeps completeness, consumption and a literal-pinned class list for the measures. The consumption test's `ALIASES` map is gone entirely (both aliases existed to hand a per-page container width to the layout, and both routes are now measure-free) — which also removes its doc-rot: it named `MY_SUBMISSIONS_CONTAINER_SIZE`, a constant that had been renamed.

**Deleted — `MY_APPS_CONTAINER_SIZE`.** Its `> 1100` assertion was a weak proxy for "this page is wide enough for its table" and could not have noticed the container dropping to 1400. Replaced by the real relationship: `APPS_PAGE_CONTAINER_WIDTH - SUBMISSIONS_CONTAINER_CHROME > SUBMISSIONS_TABLE_MIN_WIDTH`, plus the counterfactual that the readable measure would *not* clear that floor.

**Unedited and still green:** `__tests__/appListingGrid.test.ts` (5 tests) and `AppsPageLayout.geometry.browser.test.tsx` (3 tests). The store's matched pair with `LISTING_GRID_SPAN` is untouched — `/apps` takes no measure, so its content width is still the container width and the 460px card arithmetic is unchanged.

**Test-harness follow-on:** adopting the orphans made four page suites render `AppsSubNav`, which has three context/data inputs of its own. `AppEditPage.browser.test.tsx` gets the real mocks plus two new assertions that the page really is on the shared chrome; the three listing-edit suites stub the sub-nav, because wiring its data contract into suites about the listing editor would break them the next time the nav gains an input.

## Verification

- **Red/green matrix:** the new alignment guard was watched to FAIL at `origin/main` (`786fe0449`) — reproducing the before-table exactly — and passes at HEAD. The red run used the committed test file with a 4-hunk adapter for the renamed prop and module surface; the alignment assertion body is byte-identical between the two.
- **Mutation-checked** in both tiers, each time reverting exactly **one** route: the pixel guard names `/apps/review` moving 16/1408 -> 36/1368 (with all 13 `styleSheetLoaded` guards still passing, so the assertion was reached); the blocking guard names `/apps/get-started` with its own message.
- Full blocking `unit` project: **1470 files / 22945 tests passed**, 0 failures. Typecheck: 0 errors. Lint: no new errors (one pre-existing unused-import warning removed).

### DIFF
--- a/src/components/Apps/AppEditPage.browser.test.tsx
+++ b/src/components/Apps/AppEditPage.browser.test.tsx
@@ -34,8 +34,21 @@ vi.mock('~/server/utils/server-side-helpers', () => ({
   createServerSideProps: () => async () => ({ props: {} }),
 }));
 
+// 🔴 `appBlocksAuthor` IS LOAD-BEARING NOW, not decoration. This page moved onto the
+// shared `AppsPageLayout`, which mounts `AppsSubNav` — and that bar hides itself below
+// TWO qualifying tabs. Without the author capability the viewer qualifies for
+// Marketplace alone, the bar renders nothing, and the chrome assertion below would
+// pass or fail for a reason that has nothing to do with this page.
 vi.mock('~/providers/FeatureFlagsProvider', () => ({
-  useFeatureFlags: () => ({ appBlocks: true }),
+  useFeatureFlags: () => ({ appBlocks: true, appBlocksAuthor: true }),
+}));
+
+// The sub-nav's other two inputs. `useCurrentUser` is mocked rather than provided
+// because the real hook throws `missing CivitaiSessionContext` under this scaffold —
+// the same reason `AppsPageLayout`'s own browser tests mock it.
+vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
+vi.mock('~/hooks/useCurrentUser', () => ({
+  useCurrentUser: () => ({ id: 1, username: 'owner', isModerator: false }),
 }));
 
 // NOTE: this file used to carry its own `vi.mock('next/router', ...)`. It SILENTLY LOST to
@@ -57,6 +70,9 @@ vi.mock('~/utils/trpc', async (importOriginal) => ({
       getMyAppManifest: {
         useQuery: () => mocks.manifestQuery,
       },
+      // Mounted by `AppsSubNav` via the shared page layout this page now uses.
+      // Undefined data ⇒ the deterministic always-on tab set.
+      getNavSummary: { useQuery: () => ({ data: undefined }) },
     },
     useUtils: () => ({}),
   },
@@ -111,5 +127,27 @@ describe('AppEditPage (/apps/[appBlockId]/edit)', () => {
     mocks.manifestQuery.error = { message: 'FORBIDDEN' };
     renderWithProviders(<AppEditPage />);
     await expect.element(page.getByTestId('mock-notfound')).toBeInTheDocument();
+  });
+
+  test('🔴 the page is on the SHARED apps chrome, not its own bare Container', async () => {
+    // Regression coverage for the adoption: this page rendered a standalone
+    // `<Container>` and showed NO sub-nav at all, so `/apps/<id>/edit` was one of
+    // four `/apps/*` routes where the navigation simply vanished. The landmark is
+    // what `AppsPageLayout` contributes, so its presence is the adoption.
+    renderWithProviders(<AppEditPage />);
+    await expect
+      .element(page.getByRole('navigation', { name: 'App sections' }))
+      .toBeInTheDocument();
+  });
+
+  test('…and the chrome does not displace the page body', async () => {
+    // Guard-the-guard for the test above: a landmark that rendered INSTEAD of the
+    // page would satisfy it. Both must be present in the same render.
+    renderWithProviders(<AppEditPage />);
+    await expect
+      .element(page.getByRole('navigation', { name: 'App sections' }))
+      .toBeInTheDocument();
+    await expect.element(page.getByTestId('apps-edit-tab-manifest')).toBeInTheDocument();
+    await expect.element(page.getByTestId('mock-manifest-form')).toBeInTheDocument();
   });
 });
--- a/src/components/Apps/AppsPageLayout.chromeAlignment.browser.test.tsx
+++ b/src/components/Apps/AppsPageLayout.chromeAlignment.browser.test.tsx
@@ -1,0 +1,227 @@
+/**
+ * `/apps` chrome — RENDERED HORIZONTAL ALIGNMENT ACROSS ROUTES.
+ *
+ * 🔴 THE DEFECT THIS PINS. `AppsPageLayout` used to take a per-page container width
+ * and render `AppsSubNav` INSIDE that Container, so the ONE element required to be
+ * identical on every apps page inherited each page's own width and jumped sideways as
+ * you navigated. Measured on this harness before the fix:
+ *
+ *                                     @1440           @2560
+ *   route                     size    left  width    left  width
+ *   /apps                     1920      16   1408     336   1888
+ *   /apps/review              1400      36   1368     596   1368
+ *   /apps/store-preview/[..]  1320      76   1288     636   1288
+ *   /apps/submit              1100     186   1068     746   1068
+ *
+ * — a 170px left / 340px width spread at 1440, and 410px / 820px at 2560. The
+ * container is now uniform and the narrowing moved into the BODY, so every row of
+ * that table collapses to one pair.
+ *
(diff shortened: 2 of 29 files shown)


ZacxDev:
## ⚠️ Correction to the PR description — one claim above was false

An adversarial audit of this PR found that a claim in the original description did not hold. Correcting it here rather than silently editing the body, because it was up long enough for a reviewer to have read and believed it.

### What the description said

> Every page's rendered content width is byte-identical to before, so this is a pure alignment change with no confounded width delta.

### What was actually true

That held for page **bodies**, but not for page **headers**. The header band (title / subtitle / actions) sat **outside** the body's measure box, so on a measured route the header prose was bounded by the container rather than by the page's measure.

Measured on a real render with the live `/apps/submit` subtitle copy:

| | before this PR | as originally pushed |
|---|---|---|
| `/apps/submit` subtitle width | ≤ 1068 (the container capped it) | **1224.13** |

True at 1440 as well as 2560. So the change was not purely an alignment change — it widened header prose on the three measured routes that render a header (`/apps/submit`, `/apps/review`, `/apps/invites`).

### What I did about it

Rather than annotate a false claim, I made the claim true: **the measure now bounds the header band as well as the body**, applied through a single helper used at both call sites so they cannot drift to different numbers. With that, every page's rendered content width — header and body — really is byte-identical to before, and the only thing this PR moves is the chrome.

That also closes a latent trap by construction rather than by comment. `actions` is right-aligned by `Group justify="space-between"`; against the **container**, the first measured route to add a header action would have placed that button roughly 520px (`/apps/review`) to 820px (`/apps/submit`) to the right of the body it acts on at 2560. No measured route passes `actions` today, so it was never a live defect — but it was a trap waiting for the next page. Against the measure, the header and the body now share both edges.

Verified in a real browser, not inferred: the band's load-bearing 16px-in / 32px-out proximity pair is unchanged with a bounded header, and the header box shares its left edge with both the nav above it and the body below it.

### Also corrected in the same follow-up

- **The reachability note on `/apps/get-started` named the wrong trigger.** It said the empty-sub-nav-band case becomes reachable when the flag widens to `['public']`, "a one-line change in `feature-flags.service.ts`". That is wrong in the dangerous direction: `getFeatureFlags` returns the Flipt answer *before* it evaluates `availability` (`Flipt overrides role checks (both enable AND disable)`), so `availability: ['mod']` is only the Flipt-**down** fallback. The flag can widen by a runtime Flipt toggle with no code change, no PR and no deploy — so a `TODO` pinned to a diff would never have fired.
- **The caveat is three pages, not one.** `/apps/[appBlockId]/edit` and `/apps/listing/[appListingId]/edit` gate on `appBlocks` **alone**, with no author requirement, so a Flipt-granted non-author collaborator hits the same empty band. Noted at both call sites. `/apps/[appBlockId]/revenue` is genuinely unaffected — it gates on `appBlocksAuthor` + `isAppDeveloper`, which guarantees a second tab.

### And the guards themselves were weaker than claimed

Three of the source-text guards were walkable while the entire blocking suite stayed green — including one named "…and left-aligned" that asserted a relationship in its title and inspected one side of it in its body. They are now structural: a rendered-tree assertion (`renderToStaticMarkup` + happy-dom) and a TypeScript-compiler-API walk over each page's JSX, both in the **blocking** `unit` project rather than the report-only browser tier. Details are in the follow-up commit message.

PR #4452, opened 2026-08-27
feat(remix-gallery): submit a remix from the post editor, the feed and the generator, pay at click, deliver when it goes live

JustMaier:
Submit a remix into the source creator's gallery. The submitter pays when they click; the submission is delivered once the image is genuinely live, or refunded in full if it never gets there.

**Four entry points, one flag each**, so they release one at a time: the post editor card, the image context menu (feed cards and image detail), and the generator and generation queue (one component serves both). Every surface flag is checked **together with** the base `remix-gallery` flag, never instead of it — the base gates the submit mutation, so a surface released ahead of it is a live button that fails on click.

All three surface flags read `enabled: false` in prod with a `testers OR moderators` rollout, so on merge every entry point comes up for mods and testers and for nobody else.

## 🔴 Needs a migration applied by hand

`packages/civitai-db-schema/prisma/migrations/20260901220000_placement_submission_uniqueness_and_undelivered_idx/`

Two indexes. **Apply each as a single statement** — `CREATE INDEX CONCURRENTLY` cannot run inside a transaction block and `psql -c` with two statements opens one implicitly. A cancelled build leaves an **INVALID** index behind, so the check is `pg_index.indisvalid`, not the command's exit code.

1. `Placement_surface_target_image_live_key` — UNIQUE partial. This is the real enforcement for one-image-one-gallery; see below. Measured on prod 2026-09-01: 893 rows match the predicate, 893 distinct keys, so it builds clean. Re-run that count if this sits unapplied for long.
2. `Placement_resolvedAt_undelivered_idx` — the undelivered notification's own lookup.

`20260827070000_placement_awaiting_readiness_idx` is **already applied** to prod and verified valid on the replica. Nothing to do for that one.

## Why the deferral exists

Charging before the image is finished means two checks cannot be made yet, and both now wait rather than refuse: not published, and still scanning. An unscanned image has `nsfwLevel = 0`, which every content rule reads as "no rating yet", so the rating comparison against the host is deferred too.

Every moderation **verdict** still refuses at submit — `tosViolation`, `minor`, `poi`, `ingestion = 'Blocked'`, and `needsReview` on an already-scanned image. None of those becomes a pass by waiting.

`startReadyRemixSubmissionClocks` (every 5 min) resolves the deferred rows: it restamps `expiresAt` when the image goes live, so the creator's 48h starts then rather than being spent on the submitter's drafting time; and refunds in full when the image is blocked, deleted, or comes back rated above what the creator accepts. `expired` returns the whole escrow with no decline fee, because nobody declined them.

The row is invisible to the owner meanwhile — the queue already selects on `queueImageIsListable` — so no new status was needed and the `Placement_status_check` constraint is untouched.

## Review

Three rounds. `/code-review` found 10; a five-lane pass found 7 more after 8 of those were fixed; a second five-lane pass over the finished branch found the ones below. The first round alone would not have been enough — both moderation bypasses and the feature-inverting bug came from later rounds.

The worst from the earlier rounds: `queueImageIsListable` declares its own `FROM "Image" i`, and the readiness query passed it `i.id` from an outer `LEFT JOIN "Image" i`. The inner alias captured it, the correlation became `i.id = i.id`, and the filter was a **constant TRUE** — every still-scanning submission refunded within five minutes of the user paying. Then the fix for that dropped `img."needsReview"` from the SELECT, reopening the moderation bypass, with every test green because each fixture supplies the field. The token assertion is now scoped to the SELECT list.

### Fixed in the final round

- 🔴 **`ingestion = 'Blocked'` was not refused anywhere.** The earlier change replaced an outright refusal with an exclusion from `pendingReadiness`, which only chooses *which* bad outcome a charged submission gets. A draft short-circuits on `!publishedAt` and is refunded minutes later; a **published** Blocked image is charged, gets no `awaitingReadiness` marker at all, is hidden from the owner by `queueImageIsListable`, and expires silently 48h later with no notification. Now refused above `pendingReadiness`, before any money moves.
- 🔴 **Double charge.** `assertNotAlreadySubmitted` is a `findFirst` five statements before the insert, with no transaction and no lock between them, and `Placement` carried seven indexes and no unique constraint. Two tabs or a double-click charged twice, occupied two slots of one gallery with one picture, and on decline handed the owner two decline fees. Enforced now by a UNIQUE partial index; the P2002 is mapped back to the refusal the read-side guard would have given, so a lost race reads as a refusal rather than a 500.
- **Client-authored `remixOfId` could 500 the mutation.** `getRemixSourcesForImage` guarded its cast; `loadSubmissionImage`, on the submit path, still had the bare `::int`. Same guard applied.
- **`allowLowerLevels: true` on the View-original menu item**, which `ImageRemixOfDetails` deliberately does not pass — that option lets a level the viewer switched OFF render as long as something above it is on, and its comment says that is wrong for an image pulled in sideways by somebody else's provenance. Removed; the menu now matches the documented decision.

  Checked before changing it, because reversing a deliberate choice is not the same as fixing a slip: `git log -S 'allowLowerLevels'` puts it in `48dd0fb7e0`, a fix-round commit on this branch whose **own message argues for parity with `ImageRemixOfDetails`**. Nobody ever argued for the relaxation — not in the message, not in a comment, not in the body. The direction is also fail-closed: the flag swaps a bitwise `Flags.intersects` test for `nsfwLevel > maxBrowsingLevel`, and any single-bit level above the max also fails to intersect, so the strict form drops a superset. Being wrong here costs a missing menu entry, never an exposed image.
- **The generator menu did not AND the base flag**, and mounted `useRouter` and a mutation observer above its own gate — once per output, in an unvirtualised infinite list. Gate hoisted into a wrapper, base flag added.
- **The readiness pass drained on `considered`**, the trap the sibling job two functions down documents: a row leaves that set only by settling, so with Buzz degraded all ten passes re-read the same lowest-100 ids and never reach row 101. Now drains on `started + refunded`.
- **`refundUnreadySubmission` counted the call, not `result.settled`**, so a lost claim race reported a refund nobody made — and that number now also decides whether `drain` continues.
- **`markUndeliverable` in the `needsReview` branch was uncaught**, reintroducing the head-of-line wedge the refund path had already been fixed for.

### Then a second pass over the fix commit, which found a defect in the fix

Worth stating plainly because it is the point of running the lanes again rather than
declaring victory: **the drain fix above was wrong.** Changing the predicate from
`considered` to `started + refunded` stops the pass re-reading the same lowest-100
ids when settlement is down, but there are **three** ways a row leaves the set, not
two. A review-blocked row is neither started nor refunded — it is MARKED
undeliverable, and the selection SQL then excludes it. So a full batch containing
even one such row returned at most 99, the drain stopped after one pass, and rows
101+ were reachable and never reached. After a backlog that clears 100 rows per
five-minute tick instead of up to 1000. Now `marked` is returned and counted.

Four more from the same pass:

- **The guarded cast admitted numeric junk.** `'99999999999'` matches `^[0-9]+$` and
  then raises `value out of range for type integer` — the same 500 the guard exists to
  prevent, by a different route. Ranged through bigint rather than by digit count, so
  a legitimate id near the int ceiling is never silently dropped to NULL, and **nested
  rather than ANDed**: Postgres does not guarantee short-circuit evaluation of `AND`
  and could run the cast on non-numeric text anyway.
- **The free path had no P2002 mapping.** The new index covers free rows too, so a free
  submit losing the race to a paid one got a raw Prisma error where the paid path gave
  a sentence. Both writers share one `asDuplicateRefusal` now.
- **`pendingReadiness` has two arms and only one was pinned.** Deleting the ingestion arm
  left a still-scanning image charged with no marker, invisible to the readiness pass,
  with nothing red — the Blocked bug's shape, one ingestion value over.
- **The two hand-copied casts became `remixOfIdSql()`**, parameterised by alias, for the
  reason `sourceImageIdsSql` one line above already gives. Only one copy had a test, so
  they were one edit from disagreeing.

### And a third pass, which again found the fix wrong

- 🔴 **The bigint bound moved the ceiling rather than removing it.** `^[0-9]+$` admits a digit
  string of any length, so a 21-digit value overflows the `::bigint` written to catch the
  `::int` overflow — the same 500, one range up, reachable by the same route. `^[0-9]{1,10}$`
  makes the inner cast total and still admits `2147483647` itself. Verified on the prod replica
  with literals: every case returns a value or NULL and none errors, and the non-null half
  (`123`, `0`, `007`, `2147483647`) is the positive control.
- **Three assertions that let a real mutation through.** The `undeliverable` exclusion in the
  readiness WHERE — the only thing that makes a marked row stop matching, so deleting it
  restored the original starvation bug with the counter still reporting movement — is now
  asserted **with its operator**, because two bare `toContain`s pass just as happily over AND
  for OR. The commit's own "NESTED, not AND" rationale had no observer; flattening it satisfied
  both existing assertions. And the `...Once` reset covered two mocks while the same commit
  added a third to one it had not covered; all eight are reset and re-primed now.
- **`asDuplicateRefusal` routes through the existing `isPrismaUniqueViolation`** rather than a
  hand-written `code === 'P2002'` — its docblock carries the caveat this site needs, now that
  `Placement` has this partial unique index *and* its primary key.

### Which lane read which commit

Stated rather than implied, so the stopping point reads as a decision and not an oversight.

| lane | last read | why |
| --- | --- | --- |
| correctness | `66c0f35834` (head) | round two changed the drain predicate, a cast and an error path |
| tests | `66c0f35834` (head) | same, plus every assertion in the file moved under a new `beforeEach` reset |
| reuse | `4a2bb714e9` | a 178-commit rebase can manufacture a duplication out of something that landed on main in between — the only lane that can see that. **Has not read `66c0f35834`**, which changed a regex bound and three assertions and rebuilt nothing |
| perf | `b1c571a9e6` | **has not read the head.** Round two moved no query, no index, no cron and no render path; what perf measured — both indexes, the job's cost and cron, the component split — is byte-identical at the head |
| intent | `b1c571a9e6` | **has not read the head.** Round two added no surface, no flag and no user-visible behaviour; the ticket, the four entry points and the flag state are unchanged |

## Deliberate divergence, please do not harmonise

This reads **both** provenance fields; `ImageRemixOfDetails` reads the new one only. Both are correct, and the discriminator is not old-vs-new — it is **whether a human gates the thing before anyone sees it**. That card publishes attribution automatically on a public page. This one hands a candidate to the host, whose review *is* the trust boundary. The test says so, and names the other file.

## Verification

Rebased onto `origin/main` at `8cb127b4ee` before the final push — the branch was 174
commits behind. Both preview reds were **base staleness**, traced to fixes already on
main: the dropped `app_block_reviews` relation (4 App Blocks smoke specs, one SQL error)
and two component-test fixes. Control: PRs #4553 and #4550 scored `66 passed, 0 flaky`
on the same tier minutes later.

- `pnpm run typecheck` → 0 errors
- `pnpm run db:check-generated` → exit 0
- `pnpm run lint` → 0 errors on changed files
- `pnpm run test:unit:run` on the rebased head → `Test Files 1565 passed | 3 skipped (1568)`,
  `Tests 24872 passed | 35 skipped (24907)`. Both totals account for every file and every
  test. Four `scripts/test-perf/__tests__/trace-flush.test.ts` failures seen before the
  rebase were stale base as well — they fail at `b1c571a9e6` and pass here.
- CI on the head: **17 green, 1 skipped, 0 failing.** `preview / smoke-tests` reads `66 passed, 0 flaky` and `preview / component-tests` passes — the two that were red before the rebase, on the same code plus these fixes. That is what closes the base-staleness argument: only the base moved.
- `blocks.router.workflow.test.ts` collected **358** — submodule initialised, so the run
  means something
- **23 planted mutations across the three fix rounds, 23 legible failures**, each naming its
  intended test and no other. Two of them failed to *apply* on the first attempt and were
  re-run rather than counted, since a mutant that never compiled reads exactly like one
  that was caught. Three behaviour-preserving negative arms were run alongside; two stayed green
  and one came back red — reported rather than dropped, because that arm was mis-designed (it
  changed a literal the assertion deliberately pins) and loosening a good assertion to rescue a
  bad control would be the wrong repair. Sample: removing the `Blocked` refusal fails with
  `promise resolved "{ id: 96 }" instead of rejecting` — i.e. a charge; reverting the drain
  predicate fails with `expected 1 times, but got 10 times`.
- Prod, read-only: the new UNIQUE index builds — 894 rows match its predicate, 894 distinct
  keys, across **all** surfaces. It is unscoped by surface deliberately, and that is safe:
  **0 of 5,603 sticker rows carry `data.imageId`** and all 1,029 remixGallery rows do.
  Historically 5 of 1,022 groups ever held more than one row, `max(live) = 1` in every one
  — sequential resubmissions after `expired`/`declined`, both outside the predicate.

## Known, not fixed

- `apps/event-engine/src/common/feeds/images.feed.ts:212` has the same unguarded `::int` on
  `remixOfId`. Different app, different DB layer, out of this PR's reach — a follow-up.
- The pending-cap read-then-writes here and in `sticker-placement.service.ts` lack the
  advisory lock the free path already has. Racing one buys a single extra pending submission
  past a cap of 10, which is not what the unique index was defending.
- **`GeneratedOutputRemixMenu.tsx` and `RemixGalleryMenuItems.tsx` have no tests at all**, so
  the flag pairing and the browsing-level change in them rest on review alone.
- **Two drain exits are still uncounted** — a settle that lost its claim race, and
  `claimed.count === 0`. Both need a concurrent sweep and cost one tick of throughput rather than
  a backlog. Two consecutive rounds of editing that predicate each introduced a new defect, so
  this is recorded rather than taken on a third pass. The clean fix is a separate `left` counter,
  since `refunded` is doing double duty as an operator metric and as the drain predicate.
- `remixOfIdSql` and `sourceImageIdsSql` are file-private, while `src/server/utils/image-onsite.ts`
  is the established home for shared Image-meta SQL fragments. Three other readers of the same
  JSON path exist in three different shapes and none can adopt a private const.
- `placement-jobs.drain.test.ts` is the 14th hand-written `runJob`; four pre-existing copies are
  character-identical. Extract on the next job test, not this one.
- The readiness pass re-resolves the space but does not re-check `mode !== 'off'` or re-run `assertCanPlace`.
- The post editor asks `getRemixSourcesForImage` once per image card with no virtualisation and no provenance pre-filter; a remix image costs ~11 statements, two of them on the primary. Cheap for non-remix images (21 buffers) and mod-gated today — revisit before public rollout.
- A failed **free** submission still consumes the daily allowance (`countFreePlacementsToday` counts every status on purpose — shared money rule).
- `getPlacementConfig` is uncached.
- The `remix-gallery-undelivered` notification has never fired against a real row, and `placement.notifications.ts` has no test for its query or its copy.


JustMaier:
### Migration applied to prod

Justin applied `20260827070000_placement_awaiting_readiness_idx` by hand on 2026-09-01. Verified read-only against the prod replica afterwards:

| | |
|---|---|
| `indisvalid` / `indisready` | `true` / `true` |
| size | 8192 bytes (empty — nothing writes `awaitingReadiness` until this PR ships) |
| predicate | `((status = 'pending'::text) AND ((data ->> 'awaitingReadiness'::text) = 'true'::text))` |
| `Placement` indexes | 10, **0 invalid** — no cancelled-`CONCURRENTLY` leftover |

Pre-flight numbers, for anyone applying this to another environment: `Placement` was 6,587 rows / 4,288 kB with 132 `pending` rows and **0** matching the predicate, so the build is effectively instant and lands empty.

### Where this leaves the PR

`mergeable: clean`, all 14 checks green, unchanged since 2026-08-27T09:30Z at `b1c571a`. Justin wants a preview run before merging.

Remaining, from ClickUp [868kxcmae](https://app.clickup.com/t/868kxcmae):

- [x] flipt-state#79 merged; `remix-gallery-post-editor`, `remix-gallery-menu` and `remix-gallery-generator` all read `enabled: false` with a testers-or-moderators rollout in prod (re-verified 2026-09-01)
- [x] migration applied
- [ ] preview run
- [ ] test matrix on production with a tester account, result left as a comment on the ticket
- [ ] the widen-or-revert call, said out loud by whoever makes it

⚠️ Worth restating from the ticket: the money-path pass must not be run on a dev server. Dev points at production Buzz and the orchestrator, so a submit there spends real Buzz against a placement row that lands in the dev database.

PR #4454, opened 2026-08-27
feat(cosmetic-phash): widen the lane to 256 bits, and make the badge tunable

JustMaier:
@
At 64 bits the creator-shop near-match panel could not separate a copy from a coincidence. Koen shipped the fix on the orchestrator side (ClickUp [868krn12g](https://app.clickup.com/t/868krn12g)) — not a precision parameter, but a new hash type, `perceptualDct256`, live now and available from `@civitai/client` 0.2.0-beta.94 (already pinned; only `node_modules` was stale).

**[Visual evidence, every pair checked by eye →](https://claude.ai/code/artifact/6a0ec7de-10d1-44e8-8da8-e8f701defb77)**

## The measurement

Over the 1,719 hashable cosmetics:

| | 64-bit | 256-bit |
|---|---|---|
| Imitation A (badge 1426 vs official 107) | rank **7** of 1,717 | rank **1** |
| Imitation B (badge 1251 vs official 870) | rank **116** | rank **5** |
| Candidates surfaced per cosmetic | median 39, max 192 | median 0, max 3 |

Both imitations are by the same account. The 256-bit lane also catches three cross-creator pairs at distance **0** — pixel-identical re-uploads of official badges, which the submission-time sha256 cannot see because official cosmetics carry no `imageHash`.

Confirmed by eye rather than asserted: A is a redraw of the official hex badge, B a near-copy of a gold "SSS" crest, and the three zero-distance pairs are the same file.

## What is not changing, and why that is a finding

`COSMETIC_SIMILARITY_CLOSE_RATIO` stays at `0.125`. It selects the same five cross-creator pairs anywhere between 8 and 40 of 256, so it sits on a plateau rather than an edge. Being a fraction of the hash width is what carried it across the upgrade intact.

I also ran the control that could have failed: if `perceptualDct256` were greyscale, the Buzz Daddy badges — identical artwork, different metal — would collapse to 0. They sit 44–76 apart. So the zero-distance results are real duplicates, not an artefact.

## Three things that were silent before

- **A partial lane bump.** Moving only two of the three `COSMETIC_PHASH_LANE` fields left the third disagreeing and nothing failed — two of three mutations passed the suite. The lane is now asserted to spell one lane in all three fields, which kills both.
- **The near-identical decision has moved server-side.** The threshold is operator-tunable at runtime, and a client recomputing it from a bundled constant disagrees with the server until every tab reloads.
- **The threshold reads from a `KeyValue` row** (`cosmetic-similarity-close-ratio`), so it can be moved without a deploy. Read per call, not memoised — a TTL would delay the change by the length of the TTL, which is the point of the row. It degrades to the built-in default on a malformed or out-of-range value rather than throwing, because this gates a *badge* and a throw would take out the ranking underneath it. Out-of-range is rejected rather than clamped: clamping `1.5` to `1` would badge every match near-identical and read as a working threshold.

## Deploy notes

**Do not let the sweep do the backfill on its own.** The panel is visible to moderators *right now* — `enabled: false` on the Flipt boolean is the no-rollout-match default, and the `moderators → true` rollout is live underneath it (`packages/civitai-flipt/src/context.ts:5`). While the corpus straddles two lanes, a mod landing on a not-yet-drained row gets a ranking over only the other not-yet-drained rows, with no error and no visual tell.

So: deploy, then apply all 1,740 hashes in one statement (~0.4s) rather than sitting in that state for the ~2¼ hours the 15-minute sweep would take. The hashes are computed and held. The sweep then only mops up the 3 dead-CDN rows (ids 240–242, steady state) and anything new.

Turning the panel on is a separate PR against `civitai/flipt-state`, and per the runbook it waits until the corpus is on one lane.

## Verification

- 36 tests across the three covering files, green
- Lane and knob both mutation-tested; every mutant caught, unmutated control green
- The 4 pre-existing failures in `scripts/test-perf/__tests__/trace-flush.test.ts` are unrelated — identical `4 failed | 24 passed` with this branch stashed

Denominator note: counts here are over the **1,743** cosmetics with artwork, not all 1,867 `Cosmetic` rows. A plain `GROUP BY` over the whole table reads 125 nulls where this says 3 failures.

@

### DIFF
--- a/scripts/oneoffs/drain-cosmetic-phash-lane.ts
+++ b/scripts/oneoffs/drain-cosmetic-phash-lane.ts
@@ -1,0 +1,107 @@
+/**
+ * Drain the whole `Cosmetic` corpus into the current `COSMETIC_PHASH_LANE` in one
+ * run, instead of waiting out the 15-minute `cosmetic-phash-sweep` cron.
+ *
+ * Run it right after deploying a lane bump:
+ *   pnpm ts-script scripts/oneoffs/drain-cosmetic-phash-lane.ts
+ *
+ * ── WHY THIS EXISTS AT ALL ────────────────────────────────────────────────
+ *
+ * The sweep already drains a lane change on its own — it selects on
+ * `pHashVersion IS DISTINCT FROM` the constant, so raising the lane IS the
+ * backfill and this script is never load-bearing. What it buys is the *shape of
+ * the gap*, not the total time.
+ *
+ * While the corpus straddles two lanes, `getSimilarCosmetics` filters candidates
+ * on `pHashVersion = <lane>` AND on hash width. A cosmetic that has already been
+ * re-hashed is therefore compared only against the other re-hashed rows — a
+ * ranking over a fraction of the library, with no error, no width mismatch and
+ * nothing on screen to say so. At 200 rows per tick a ~1,740-row corpus sits in
+ * that state for about two and a quarter hours.
+ *
+ * That matters more than it looks, because the panel is NOT dark. Flipt
+ * `cosmetic-similarity` is `enabled: false`, but `enabled` is the value returned
+ * when no rollout matches, and the `moderators → true` rollout does match (see
+ * `packages/civitai-flipt/src/context.ts`). Moderators can see the panel today.
+ * A partial ranking presented to the person deciding whether artwork is stolen is
+ * the failure this whole feature exists to prevent, so crossing that window in
+ * minutes rather than hours is worth one manual step.
+ *
+ * ── WHY IT DELEGATES INSTEAD OF WRITING SQL ───────────────────────────────
+ *
+ * It calls `sweepCosmeticPerceptualHashes` in a loop rather than issuing its own
+ * UPDATE. Every rule about what a correct row looks like — the legacy `pHash`
+ * left null once a hash no longer fits a BIGINT, `pHashUrl` recording what was
+ * actually hashed rather than what the cosmetic currently points at, the failure
+ * stamp cleared on success and set only on failure — lives in
+ * `storeCosmeticPerceptualHash`, is tested there, and would have to be restated
+ * here to be got wrong independently. A hand-rolled backfill is how the two
+ * drift.
+ *
+ * Consequences of that choice, both deliberate:
+ *   - It is idempotent and resumable. Interrupt it and re-run it; rows already in
+ *     the lane no longer match the predicate.
+ *   - It cannot get stuck on dead artwork. A row that fails is stamped and drops
+ *     out of the predicate for 24h, which is what terminates the loop rather than
+ *     a retry count.
+ */
+import { sweepCosmeticPerceptualHashes } from '~/server/jobs/cosmetic-phash-sweep';
+import { COSMETIC_PHASH_LANE } from '~/server/services/cosmetic-phash.service';
+
+// Larger than the cron's 200/5 because nothing else is competing for the run.
+// Concurrency is the orchestrator's limit, not ours — 10 measured ~10.5 hashes/s.
+const BATCH_SIZE = 500;
+const CONCURRENCY = 10;
+
+// A stop, not a target. The loop ends when a batch comes back empty; this only
+// bounds a predicate that never drains — a bug in the sweep, or artwork being
+// created faster than it is hashed — so the script exits loudly instead of
+// hashing forever.
+const MAX_BATCHES = 40;
+
+async function main() {
+  console.log(`[drain] lane ${COSMETIC_PHASH_LANE.version} (${COSMETIC_PHASH_LANE.hexLength} hex)`);
+
+  let batches = 0;
+  let hashed = 0;
+  let failed = 0;
+
+  for (;;) {
+    if (batches >= MAX_BATCHES) {
+      console.error(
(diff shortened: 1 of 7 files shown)


JustMaier:
## Five-lane adversarial review — findings and fixes (`d28f7cc`)

Ran correctness, tests, reuse, perf and intent. **Five real findings, all fixed.** Two of them contradicted claims I made in the original PR body, so those are corrected here rather than quietly.

### Fixed

| # | Lane(s) | Finding |
|---|---|---|
| 1 | correctness + reuse | `getCosmeticSimilarityCloseRatio` guarded the **value** but not the **read**. An unguarded `dbRead` rejection escapes `getSimilarCosmetics` *after* the ranking completes, and the card renders it as "this artwork was not compared against anything" — false, and precisely the confusion that card exists to remove. It also contradicted the function's own FAIL DIRECTION docblock. Read is now wrapped. |
| 2 | tests + correctness | **The `close` field had no assertion anywhere.** Replacing it with `close: false` passed 36/36. The deliverable that moved client→server was guarded by nothing at either layer. |
| 3 | tests | **The KeyValue read was unwired from the tests.** Dropping `await getCosmeticSimilarityCloseRatio()` entirely also passed 36/36 — every `getSimilarCosmetics` test reset the mock to `undefined`, so the suite only ever exercised the fallback. The knob was, in test terms, connected to nothing. |
| 4 | tests | **3 of 5 malformed-value cases were passing on residue.** `loggingMock.logToAxiom` resets once per *file*, so later `it.each` cases were satisfied by the first case's warning. A silent-fallback mutant now kills five tests instead of two. |
| 5 | tests | A test of mine asserted a ratio it claimed was "what the UI thresholds on". **It wasn't** — both imitation pairs sit *above* the shipped `0.125`, and it hardcoded `/256` so the next lane bump would divide by the wrong denominator in silence. Deleted; the neighbouring test already pins both distances exactly. |

Also taken: the `0`/`1` boundary was inconsistent — `0` ("only exact matches count") was rejected and fell back to the **looser** default, giving an operator *more* red badges than they asked for, while `1` was accepted and badges everything. Now `[0, 1)`. And `normalizeCosmeticHashHex` refuses a hash wider than the lane; `padStart` returned it unchanged, which would store an over-wide hash under the current version — accepted as a target, excluded from every candidate set, reporting "nothing was close" forever on a row that looks fine.

**Re-mutated after fixing.** All five previously-surviving mutants now die, with the unmutated control green either side of each run:

```
M-A close -> false ............... CAUGHT (was: 36/36 pass)
M-B drop the KeyValue read ....... CAUGHT (was: 36/36 pass)
M-C silent out-of-range .......... CAUGHT (5 tests, was 2)
M-D catch rethrows ............... CAUGHT
M-E normalize stops refusing ..... CAUGHT
CONTROL .......................... green
```

### Not changed, deliberately

- **Perf lane: no blockers, measured.** The candidate query is a 1.5 ms seq scan over 160 pages — a full scan by design, since every row must be compared; an index on `pHashHex` would be write cost for a predicate no query issues. The JS Hamming loop goes 2.18 ms → 8.24 ms p50 (the predicted ~4x). On a `moderatorProcedure` fired per item click, that is noise. The new `keyValue.findUnique` is a 0.014 ms PK index scan. Storage +48 B/row ≈ +84 kB, nothing TOASTed.
- **Perf, worth recording for later:** cost is O(corpus × width) on the request thread. At ~10,000 hashable cosmetics the same loop is ~48 ms of blocking. The exit is `bytea` + `bit_count(a # b)` in Postgres. Nothing to do now.
- **Perf, pre-existing:** `createJob`'s default `lockExpiration: 5 * 60` is shorter than a worst-case sweep tick, so a slow orchestrator can let a second sweep start over the same batch. Not introduced here — but the lane bump makes every tick a full 200-row batch for ~2h, which raises the odds. Filed, not fixed.
- **Reuse lane:** no named helper was rebuilt. There is no generic "read a scalar setting from KeyValue with a default" in this repo — there are six independent implementations with deliberately different fail directions, and collapsing them would be the wrong coupling. The no-memo decision matches `placement.service.ts` and `announcement-allowance.service.ts`, which also read per call.

### On why the lane stayed a code constant

Intent lane fairly noted the original ask tied the key-value store to the **resolution**, and I made the **threshold** tunable instead without saying why. The reason: `version`, `hashType` and `hexLength` must move atomically, the sweep predicate keys off the constant, and a runtime lane flip has no deploy boundary — so it would strand the corpus mid-lane with no way to sequence the drain. The threshold has none of those couplings, which is why it *can* be a row. Recording that here since it was a silent substitution before.

### Deploy note updated

The fast-apply this PR relies on existed only as a scratch file. It is now `scripts/oneoffs/drain-cosmetic-phash-lane.ts` — it loops the tested sweep rather than issuing its own UPDATE, so the rules about what a correct row looks like stay in `storeCosmeticPerceptualHash`. Idempotent and resumable.

### Verification

Full suite: **23,098 passed**, 4 failed — all 4 in `scripts/test-perf/__tests__/trace-flush.test.ts`, confirmed pre-existing by stashing this branch and getting the identical `4 failed | 24 passed`. Typecheck 0 errors.

One process note worth flagging: a review agent left mutations live in the shared working tree instead of restoring them. My commits were clean, the checkout was not, and two different unrestored mutants were observed before it was stopped. Worth guarding against in any future multi-agent review of a shared tree.

PR #4456, opened 2026-08-27
fix(app-blocks): bound the block storage read caches, and invalidate on write

ZacxDev:
> **Note:** this description was rewritten after two adversarial audit rounds changed the approach. The original version, and what was wrong with it, are preserved in [this comment](https://github.com/civitai/civitai/pull/4456#issuecomment-5443792360) — worth reading if you saw the first one.

## The bug

civitai's QueryClient sets `staleTime: Infinity`, and both App Blocks hosts serve every storage read through `trpcUtils.apps.{shared,storage}.*.fetch` — React Query's `fetchQuery`, which resolves straight from cache whenever the entry is not stale. Under an infinite staleTime such an entry is never stale, and there was **no `invalidate` call anywhere in the `AppBlocks` directory**. Two distinct defects fall out:

1. **A block could not see its own write.** `append/vote → list` returned the pre-write snapshot.
2. **A block could not see anyone else's write.** The shared store is cross-user by definition — a votes board, a leaderboard — so a viewer never observed another user's activity without reloading the page.

Both apply to the private per-user KV bridge too (`apps.storage.set/delete` then `get/list`), for defect 1.

## The fix — two mechanisms, each bounding one failure

- **`BLOCK_STORAGE_READ_STALE_TIME_MS` (1s)** on all 7 reads per host. This is what makes defect 2 fixable at all; invalidating locally cannot help, because the writer is a different browser.
- **Invalidate on write**, all 8 writes per host, awaited *before* the `*_RESULT` reply. The SDK hook resolves on that reply and the block may re-read immediately, so invalidating afterwards races the read it exists to serve.

**Deliberately not `staleTime: 0`.** These reads are `publicProcedure`s hitting the per-app Postgres schema with no server-side read rate limit — only the writes have one. The cache is the only thing bounding a block that polls in a loop, and `0` would remove that while buying nothing over 1s.

**Why not cancel the in-flight read instead.** `successState()` sets `isInvalidated: false`, so a read already in flight when the write lands resolves afterwards with pre-write rows and clears the invalidated flag — restoring permanent staleness through the fix itself. Cancelling closes that but aborts a legitimate concurrent read. The staleTime bound caps it at one second instead, which is why the policy is two mechanisms rather than one.

## Coverage

- **`blockStorageCacheSemantics.test.ts`** — the library contract, against a real `QueryClient` and the real tRPC key shape. Every other test mocks tRPC wholesale, so without this nothing checked that a per-call staleTime bounds `fetchQuery`, that a namespace invalidate reaches these keys, or that it does *not* cross into the other namespace. Includes an in-repo reproduction of the in-flight race above.
- **`blockStorageCacheParity.test.ts`** — both hosts, every read bounded and every write invalidating before it replies. `IframeHost` has no shared-storage browser suite, so without this half the call sites rest on nothing.
- **Browser suites** — behavioural pins on the real page host and real postMessage bridge, including the ordering, an invalidation-failure case, and read-only controls.

## Known limits, deliberate

- A block writing through REST `/api/v1/blocks/shared-storage/increment` bypasses the host, so a later bridge read can still be up to one staleTime stale.
- The invalidation is namespace-wide: one block's write drops another block's cached shared reads on the same page.
- The parity guard is a **text** predicate. It proves the call's characters appear in the right window, not that the call runs — deletion, comment-out (both line-start and trailing forms) and reordering are caught; deliberately inert look-alike text, e.g. a string literal containing the call, is not.
- `IframeHost` has no browser suite, so its runtime behaviour rests on source reading plus that text guard.

## Mutation results

| Mutation | Result |
|---|---|
| comment out one invalidation, line-start (IframeHost) | 2 parity failures |
| comment out one invalidation, **trailing** (`void 0; // await …`) | 2 parity failures (round 4 — this walked past in round 3) |
| delete the `try/catch` swallow | 1 browser failure |
| leave one read unbounded | 1 parity failure, naming the call site |
| delete an invalidation | 2 parity + 2 browser |
| invalidate AFTER the reply | 1 parity + 1 browser (the ordering test) |
| raise the staleTime 1s → 5s | 1 parity failure (added in round 3 — this previously passed) |

## Test plan

AppBlocks unit 635, component 409, `pnpm typecheck` 0 errors, eslint + prettier clean on all added files.


ZacxDev:
## Correction to this PR's original description

An adversarial audit found the original description **overstated the bug and understated the scope**. Correcting publicly rather than quietly editing, since the first version has been readable for a while.

**1. "for the entire lifetime of the page" was wrong.** Measured in `@tanstack/query-core@5.101.0`: `gcTime` defaults to 5 minutes in the browser, `scheduleGc()` runs in the fetch `finally`, and an observer-less idle query is dropped. These reads have no observers, and cache hits do not reschedule gc — so the stale window was ~5 minutes after the last *network* fetch of that key, then it self-healed. The bug was real; the duration was not. Any staleness figure in the original body should be read as "up to ~5 minutes", not "forever".

**2. The description framed the bug class as closed when it was not.** Two gaps, both now fixed in this PR:

- The **private per-user KV bridge** had the identical defect. `apps.storage.set/delete` are mutations, `apps.storage.get/list/getQuota` go through the same `fetchQuery`, and nothing invalidated. `set(k,v)` then `get(k)` returned the old value.
- **Cross-user staleness was never addressed at all.** Invalidating on the local write only ever fixed *your own* writes. The shared store is cross-user by definition, so under `staleTime: Infinity` a viewer never observed another user's write without reloading. That is arguably the larger half, and the first version did nothing for it.

**3. The original fix was defeatable by a race, and that is why the approach changed.** `successState()` sets `isInvalidated: false`, so a read already in flight when the write lands resolves afterwards with pre-write rows and clears the invalidated flag — restoring permanent staleness through the fix itself.

The PR now bounds every read with a short non-zero `staleTime` **and** invalidates on write. The bound is what makes cross-user writes visible and caps that race at one staleTime instead of forever; the invalidation is what makes your own write visible immediately.

**Two limits are deliberate and not fixed here**, stated so nobody reads the class as closed a second time:

- A block writing through the REST `/api/v1/blocks/shared-storage/increment` endpoint bypasses the host entirely, so a later bridge read can still be up to one staleTime stale.
- The invalidation is namespace-wide, so one block's write drops another block's cached shared reads on the same page.

**Two guards in the first version were decorative, and the audit demonstrated both.** Commenting out an invalidation left the parity guard fully green, and deleting the helper's `try/catch` left the entire suite green. Both are now closed and mutation-verified; details in the latest commit message.

PR #4471, opened 2026-08-28
fix(blurbs): reconcile inside the write transaction

manuelurenah:
## What

`reconcileBlurbReferences` ran **after** the entity write had already committed, and it is not wrapped in a try/catch. So when it threw, the caller got a 500 for a save whose row already existed — and the creator could not see that row, so they saved again.

All nine call sites now reconcile **inside** the entity write's transaction, so a reference failure rolls the entity back rather than stranding it.

## Why now

Between 19:20Z and 22:11Z on 2026-08-27, `Blurb`/`BlurbReference` did not exist in production (the `20260825000000_add_blurbs` migration had not been applied). Every reconcile threw:

```
Invalid `prisma.blurbReference.findMany()` invocation:
The table `public.BlurbReference` does not exist in the current database.
```

`model.upsert` failed **975 times across 104 distinct creators**, and left **562 surplus `Model` rows** in 71 `(userId, name)` groups — one creator produced eight copies of the same model. Model publishes fell from a ~30/hr baseline to 8/hr. Those rows are all versionless `Draft`s with no engagement, and `remove-old-drafts` will reap them on 2026-09-26; no cleanup is proposed here.

The missing migration was the trigger. **The half-committed save is the defect, and it outlives that incident** — any future failure in this call has the same shape.

## Per call site

| service | shape |
|---|---|
| `model.service` ×2 | `create` / `update` wrapped in `$transaction` |
| `article.service` ×2 | moved inside the `$transaction` each branch already had |
| `model-version.service` ×2 | create converted from the **array** form of `$transaction` to interactive; update wrapped |
| `cosmetic-shop.service` ×1 | create/update ternary wrapped |
| `bounty.service` ×2 | `createBounty`/`updateBountyById` own their writes, so `blurbUses` is threaded down and reconciled inside their existing transaction |

## `reconcileBlurbReferences`

Takes a **required** `tx`, and reads through it rather than `dbRead`. It previously read `dbRead` and wrote `dbWrite` — a read-then-write against a lagging replica can miss reference rows that then survive the delete. There is no default client, so reconciling outside the entity's own write does not type-check.

The per-use upsert loop became a single `INSERT ... ON CONFLICT DO UPDATE`. It was one round trip per use, up to the 20-blurb per-creator cap, and it now runs inside the caller's transaction holding the entity row.

## Two behaviour changes worth a reviewer's eye

1. **`saveItemInCollections` moves after the commit** in `model.service`. It sat between the create and the reconcile. It is already best-effort and `.catch()`-wrapped, so this is safe — but it is a real ordering change.
2. **A failed reconcile on an update now rolls the creator's edit back.** That is the intended atomicity, but it is a stronger failure than on the creates.

## Verification

**The raw SQL was exercised against a real database before anything was built on it** — insert, conflict-update and `pendingSince` clearing, through actual Prisma binding. That is where an int4/int8 array mismatch would have hidden and where nothing downstream would have rejected it. Probe data cleaned up.

**Mutation check.** Moving the create-path reconcile back outside the transaction turns the new regression test red in **under a second**, with `expected false to be true`. A mock cannot roll back, so the observable is the callback boundary: the rejection has to escape from inside `$transaction`, which is what makes Postgres discard the write.

| check | result |
|---|---|
| full unit suite | 6051/6051 suites, **23730 passed, 0 failed** |
| `pnpm run typecheck` | exit 0 |
| `blocks.router.workflow` collected | 370 tests (submodule initialised, run is real) |

Two full-suite failures were caused by this change and are fixed here: the Flipt eval-context ledger pins `file:line`, and an added import shifted `blurb-materialize.service.ts:65` → `:66`; and `model-version.unpublish-refund.service.test.ts` deliberately hands `$transaction` a *distinct* `mockTx`, so the version update now lands there rather than on `dbWrite`.

## Reviews

Both no-gate lanes have been run over this diff, and their findings applied in `08f78d5529`.

`comment-review` found two tests the first commit had made **vacuous** — one seeded the replica while reconcile had moved to the write client, the other asserted on `dbWrite` after the call moved to `mockTx`. Both were green and proving nothing; both are fixed, and the repaired one was re-checked with the probe that exposed it. It also found a comment the change left factually false, and its suggestion to make `tx` required rather than documented is what `08f78d5529` does.

`docs-drift-review` found one doc this change made wrong — `docs/entity-change-tracking-plan.md:211` pointed a future implementer at `dbWrite.model.update` in `upsertModel`, now `tx.model.update` inside the transaction; following it would have put an HTTP emit inside a transaction callback. Fixed, along with two pre-existing pins it flagged nearby.

One finding was **not** applied: `.claude/agents/civitai-reuse-review.md:52-55` has a stale services file count and sizes. Real, but it predates this work and is unrelated to it.

https://claude.ai/code/session_01RpTaMoSvAoWAbkkum6kR1F


docs/testing/shared-module-mocks.md
```diff
@@ -147,7 +147,7 @@ function`:
 | `count` | `0` |
 | `$queryRaw`, `$queryRawUnsafe` | `[]` |
 | `$executeRaw`, `$executeRawUnsafe` | `0` |
-| `$transaction` | runs the callback with `dbWrite`, or `Promise.all` for an array |
+| `$transaction` | runs the callback with the client it was reached through (`dbRead.$transaction` → `dbRead`), or `Promise.all` for an array |
```
manuelurenah:
```suggestion
| `$transaction` | runs the callback with `dbWrite`, or `Promise.all` for an array |
```

PR #4472, opened 2026-08-28
feat(app-blocks): real tool calling for the chat-completion step

ZacxDev:
Widens the civitai edge so an App Block can submit a `chat-completion` step carrying `tools`, receive the model's `tool_calls`, and feed a tool **result** back for the next round. The orchestrator already supported all of this — the block was fenced off by three narrowings on this side.

## What changed

- **`chatMessageSchema` is now a discriminated union on `role`**, every member `.strict()`. `assistant` may carry `tool_calls` (needed to replay the turn that requested a tool) and must carry content, tool_calls, or both. `tool` carries `content` + a required `tool_call_id`.
- **`tools` / `toolChoice` on the params schema**, bounded: max 8 tools, pattern-bounded names, a description cap, and a `parameters` JSON Schema capped on serialized size **and** nesting depth.
- **`extractText` now returns each `tool_calls[].function.arguments`**, so model-written argument strings go through the same output scan as any other generated text.
- **A new optional `extractToolCalls` surface** publishes the structured calls, gated on the same scan verdict.

## Billing — the loop is bounded, `prepaidFixed` is unchanged

`MAX_TOOL_ROUNDS = 3`, enforced as a `superRefine` counting `role: 'tool'` messages.

**Why 3:** two rounds cover the shape this exists for — call a catalog tool, read the result, answer — and the third leaves room for one refinement when the first query comes back empty, which is the whole argument for real tool calling over a one-shot heuristic. Raising it is additive; lowering it is breaking. It sits far below the *incidental* ceiling `MAX_MESSAGES` already imposes (~15 rounds), because that ceiling is neither stated nor intentional.

**Why this keeps the mode honest:** there is **no server-side loop**. Each round is its own submit, so each gets its own live orchestrator quote, its own per-call `buzzBudget` gate and its own reservation against every cap — which is what "cost knowable before execution" actually requires. The bound is enforced at **parse**, so it holds on the estimate path as well as submit (both go through `parseStepParams`); a block cannot loop past a cap it applies to itself.

## Moderation — the new surfaces are covered, not silently widened

- Tool **arguments** are scanned, because `extractText` returns them.
- Structured tool calls are published **only** by `attachModeratedStepTextOutputs`, **only** onto a released verdict, and any incoming value of the field is **stripped** — the same single-writer property `textOutputs` has.
- **Containment holds BY CONSTRUCTION.** Both extractors derive from one shared predicate (`publishableToolCalls`), so `extractText` publishes exactly the `arguments` of the calls `extractToolCalls` returns and no divergence is expressible without editing that function.
  - ⚠️ **CORRECTED in audit round 3, and this is the second correction to the same sentence.** Round 2 claimed clause 8b "fires the moment anyone re-splits these predicates" and that the claim was **mutation-tested**. The mutation actually run (M5) was the *deletion* of `extractText`'s tool-arguments loop. Round 3 executed two **genuine re-splits** and both loaded **CLEAN**: restoring `extractText`'s pre-consolidation `trim().length > 0` predicate over raw arguments, and re-splitting `extractToolCalls` widened on the id axis — the latter being precisely the containment break 8b exists to catch. The description was wider than the implementation: the same defect class as the round-1 vacuous loop, one level up.
  - **Root cause was the SAMPLE, not the clause.** `canonicalOutputFor` carried one tool call with a conforming id, conforming name and non-empty arguments, so every divergence class this PR fixed was invisible to it. It now also carries a choice with **empty arguments** and one with an **out-of-charset id and distinct arguments**, each arming 8b against one class. On the final code all three mutants — both re-splits and the deletion — fire 8b's own error at registry load. The docstring now claims only that, and states that a re-split on an axis the sample does not cover would still pass.
  - 🔴 **Argument-string distinctness across choices is load-bearing.** A colliding `arguments` value is found in the scanned set and 8b goes silent — mechanically how the previous sample failed to arm it. Pinned by a test.
  - ⚠️ **CORRECTED after audit round 1.** As originally shipped this clause was **unreachable for `chat-completion`**: `canonicalOutputFor` returned a content-only sample, so the containment loop iterated an empty set and the invariant was asserted over nothing. Demonstrated by mutation — a real containment break in the shipped extractor loaded **clean**. The canonical sample now carries a real tool call and the clause executes; the same mutation now throws clause 8b's own message. The M2 row below is evidence about the synthetic registry fixture, not about the shipped entry.
- Tool **names and call ids** are not scanned. The unscanned published set is exactly `{name, id}`, enumerated in the code.
  - ⚠️ **CORRECTED in audit round 2.** This previously said the argument for both **is the same** and that neither **can carry prose**. Both claims were wrong. `name` rests on *two* legs — it is caller-authored and charset-enforced on input — while `id` is **provider-generated** and rests only on the extractor's pattern. And the pattern does not prevent prose: a 63-char underscore-joined sentence passes `[a-zA-Z0-9_-]{1,64}`. What it actually buys is a hard bound on length and alphabet (no markup, urls, AIR literals or whitespace), which is what makes the residue small enough to accept.
  - ⚠️ **CORRECTED after audit round 1.** This line originally enumerated the unscanned published set as **`{name}`** alone. `id` was missing from it and was, in fact, published **unbounded** — a 5,000-char id carrying prose and a literal AIR went through verbatim, and on a response whose `extractText` is empty the scanner was never invoked at all. `id` is now bounded exactly as `name` is. The unscanned published set is `{name, id}` and is enumerated in the code.

`moderationPosture` stays `'textOutput'`; `ACCEPTABLE_POSTURES_BY_TYPE` is untouched.

## Two things found while implementing

**The generated `@civitai/client` types cannot express a tool call.** `ChatCompletionInput.tool_choice`, `ChatCompletionTool.parameters` and `ChatCompletionFunction.parameters` are all typed `null` — a generator artifact for free-form JSON, contradicted by those fields' own doc comments (`tool_choice`'s says it "Can be \"auto\", \"none\", \"required\", or an object specifying a particular function"). Handled with a narrow, documented intersection that keeps every *other* field anchored to the generated type, rather than an `as` cast that would drop the anchor wholesale. If the generator is fixed upstream, delete it.

**The wire name is `tool_choice` (snake), the param is `toolChoice`.** This input mixes conventions — `maxTokens` is camelCase while every tool field follows OpenAI's snake_case. Getting it backwards does not error (an unknown field is ignored), so the feature would be silently inert; pinned by an exact-key-set test.

Separately: `tools[].function.parameters` is the **first `z.unknown()` param in this registry**, which is exactly the case `containsAirReference` documents as the point where its totality claim becomes reachable via `toJSON`-bearing instances. It is normalised through a JSON round-trip at parse, so what the AIR scan walks is byte-identical to what reaches the wire. A test demonstrates the hazard *and* the fix in both directions.

## Verification

**Red → green matrix.** Measured by reverting only the four source files to `origin/main` and re-running the new tests against pre-change code:

| | at `origin/main` | at HEAD |
|---|---|---|
| step suites + router moderation suite | **30 failed**, 295 passed | **325 passed** |

The 30 that go red are the regression tests. Named individually in the run log; they cover the round bound (both paths), the tool/assistant message members, `tool_choice` wire naming, `extractText` reading arguments, `extractToolCalls` in all its forms, clause 8b, the JSON normalisation, and the verdict-gating pair.

**Honest limits on the rest.** Some new cases pass at base and are **not** regression tests — they are forward guards, and they pass at base *for the wrong reason*: with `tools`/`toolChoice` absent from the schema, `.strict()` rejects the whole payload as an unknown key, and with no `'tool'` role the message is rejected by the enum. That covers `REJECTS a bad toolChoice value`, the name-charset / description-length / non-object-`parameters` / unknown-property cases, and `REJECTS a tool message with no tool_call_id`. They are worth keeping — they pin the bounds now that the fields exist — but they prove nothing about the change.

One case, `accepts exactly MAX_TOOL_ROUNDS tool messages`, **degenerated** at base: the named import resolved to `undefined`, so `Array.from({ length: undefined })` produced an empty array and the fixture collapsed to a single user message. Second commit adds a guard pinning `MAX_TOOL_ROUNDS === 3` so that cannot happen silently again.

**Mutation results** — each guard broken on purpose, on a reachable input:

| # | mutation | result |
|---|---|---|
| M1 | round-bound condition → `false` | **killed** (2: submit + estimate cases) |
| M2 | clause 8b containment → `false` | **killed** (1, on its own message) |
| M3 | withhold path carries `toolCalls` | 🔴 **SURVIVED** — see below |
| M3c | M3 **plus** the merge pushes unconditionally | **killed** (2, incl. the withhold case) |
| M4 | stop stripping incoming `toolCalls` | **killed** (1) |
| M5 | bypass the JSON normalisation | **killed** (1) |
| M6 | drop the name pattern in the extractor | **killed** (1) |
| M7 | `extractText` stops returning arguments | **killed** (2, incl. the containment case) |

🔴 **M3 survived, and it is reported rather than smoothed over.** Making the *withhold* path carry tool calls changes nothing observable, because `attachModeratedStepTextOutputs`'s merge ignores them on a withheld verdict — a second, independent layer. So the property "withheld ⇒ no tool calls" has genuine depth-2 protection and no single-point mutation breaks it. M3c confirms the test is not vacuous: break **both** layers and the withhold case goes red. The early return is therefore defence against a future refactor of that merge, which is what its comment claims — not the sole control.

**Suite-wide:** `pnpm typecheck` → 0 errors (watched go red→green: 12 → 0 while fixing a missing submodule). `vitest run --project unit*` over `blocks/`, `routers/__tests__/`, `schema/` → **261 files, 5920 tests, all passing**. `eslint` on the touched files → 0 errors (1 pre-existing unused-import warning, present at `origin/main`).

## Not done

- **No tool transport route.** Nothing here fetches a tool; the block calls the tool itself and submits the next round.
- **No sensei integration**, and the client-side heuristic is untouched.
- **No server-side loop**, deliberately — see the billing section.
- **SDK mirror follow-up needed.** `toolCalls` is additive on the snapshot type that `@civitai/app-sdk`'s `blocks/types.ts` mirrors. Nothing here edits another repo, so a block cannot read the field until that type is widened in the SDK — same posture as `textOutputs` and `modelSubstitutions` before it.
- **One thing to confirm before this is useful in prod:** the wire spelling of `tool_choice` is taken from the generated client type, which is the contract this repo pins. The prior live probe recorded that a tools payload is parsed and priced, but did not record the exact request key for the choice field. Worth one `whatif` against the live orchestrator to confirm before relying on `toolChoice: 'required'` behaving as requested — an ignored field fails silently.

## Audit round 2 — what changed

A blind delta re-audit of the round-1 fixes found one regression **the round-1 fix itself introduced**, plus two prose defects.

- 🔴 **A no-argument tool call vanished.** Round 1 aligned the two extractors by *dropping* empty `arguments`. Measured end to end against the registered entry: a response whose only choice was a tool call with `arguments: ''` produced a snapshot with **neither `textOutputs` nor `toolCalls` nor `textOutputWithheld`** — charged, `succeeded`, publishing nothing diagnosable.

  | | `extractText` | `extractToolCalls` | snapshot |
  |---|---|---|---|
  | before | `[]` | 0 | `{workflowId, status}` |
  | after | `["{}"]` | 1 | `{…, textOutputs:["{}"], toolCalls:[…]}` |

  Empty/whitespace arguments are now **normalised to `'{}'`**. This loosens nothing: containment compares published arguments against the *scanned* set, and a whitespace-only string carries nothing a scanner can read. `'{}'` is also the only value a consumer can use — `JSON.parse('')` throws.

- **The real fix is consolidation.** The predicate had been open-coded at two sites and was wrong at one of them three separate ways: the empty-arguments divergence, the drop-the-call regression that replaced it, and the ID axis — a non-conforming provider id made `extractToolCalls` drop the whole call while `extractText` still published its arguments, handing a block argument JSON in `textOutputs`, an empty `toolCalls`, and no id to answer with. One predicate now, two consumers.

- **Input and output agreed to disagree about ids.** `call:with:colons` parsed on input while the extractor refused it on output. Reconciled toward the stricter definition, which cannot reject a legitimate payload: the only id a block can replay is one we published.

- **`MAX_TOOL_ROUNDS` prose (N3).** The comment calling an unanswered ask "legal" was a statement about *this schema's acceptance* worded as a claim about the *provider*. Rewritten, with the three accepted shapes recorded as a known unvalidated gap. **No guard added deliberately** — the premise comes from the OpenAI-compatible contract, not a probe against this orchestrator, and a guard on an unverified premise risks being *too strict*, which is the worse failure: the gap costs a charged submit, a wrong guard costs a working feature refused at parse.

### Round-2 mutation results

| mutant | outcome |
|---|---|
| M1 revert normalisation | killed — only the normalise test |
| M2 drop `id` charset | killed — CHARSET-only test; LENGTH-only survives |
| M3 drop `id` length | killed — LENGTH-only test; CHARSET-only survives |
| M4 diverge `extractText` | killed — the orphan-arguments test |
| M5 smuggle after the shared helper | killed — clause 8b fires at registry **load** |
| M6 drop assistant `id` regex | killed *(after the test was fixed — see below)* |
| M7 drop `tool_call_id` regex | 🔴 **SURVIVED** |
| M8 drop both id regexes | killed |

**M7 is reported, not hidden.** It is a true property of the code: every payload that could violate that regex is already rejected by the assistant-side regex or by the correlation guard, so it is defence-in-depth no test can isolate. The docstring now says exactly that rather than implying coverage.

**M6 initially survived, and the cause was the test just written for it** — it carried both an assistant `id` and a matching `tool_call_id`, each separately bounded, so removing either regex alone was masked by the other. Fixed with an assistant-only payload where nothing else can do the rejecting.

**Test attribution.** The headline id test used a fixture that was 4,900+ chars **and** contained spaces and colons, so the pattern and length mutants each died to it individually and it could not tell you which guard was load-bearing. Split into pairwise-distinct fixtures; the combined case is kept as the end-to-end regression and labelled non-attributing.

`typecheck` 0 errors; **5931 tests** pass across the blocks, router and schema suites.

## Audit round 3 — what changed

Delta round 5 confirmed all five round-2 claims fixed, and the M6/M7 reports accurate. It found two substantive residuals, both fixed here (`a8dedefa19`).

**N2 — clause 8b's stated coverage was refuted by execution.** See the corrected bullet above. Sample extended; all three mutants now fire.

**The canonical sample is now 2 CAPTURED choices + 2 declared ADVERSARIAL ones**, and the docstring says so. The previous "every choice is copied verbatim from a real orchestrator response" banner became false the moment adversarial values were added, and a stale verbatim claim over invented data is worse than none. The both-wrong-blind argument is intact because every *shape* is still captured; only *values* vary.

**N1 — the silent-success class had been moved and widened, for the third round running.** A malformed `id` or `name` is dropped by the shared predicate on both sides, so with the measured `finishReason: 'tool_calls'` shape (no `content` key) the response yields nothing anywhere: charged, `succeeded`, and publishing none of `textOutputs` / `toolCalls` / `textOutputWithheld`.

Enumerating drop reasons has failed three times, so this round does not enumerate them. `attachModeratedStepTextOutputs` now asserts the **outcome**: a text-posture step that released while contributing neither text nor tool calls surfaces a reason. Scoped to text postures deliberately — a media entry contributes no text legitimately, and widening it would report that correct absence as a fault on any mixed workflow.

The reason is a **distinct constant**, not `TEXT_OUTPUT_WITHHELD_MESSAGE`: nothing failed the content policy here, and reusing the policy message would mislead the app author *and* make a real withhold indistinguishable from an extraction failure. Honest limit recorded in code: the two are not machine-separable without matching on the string.

**N3 — the input `id` regex justification overclaimed.** It said the tightening "cannot reject a legitimate payload"; hand-authored few-shot tool-call history may use any id, and `call.1` / `call:1` now `BAD_REQUEST`. Reworded to the true guarantee — *cannot reject a payload derived from our own output* — with the accepted cost stated.

**The `tool_call_id` charset regex was REMOVED rather than documented.** Round 5 proved it changes no accept/reject outcome (`declaredCallIds` is populated only from already-bounded assistant ids), so its one observable effect was replacing the correlation guard's useful diagnostic with a zod regex issue. The length bound stays — it bounds the payload before correlation runs.

**N4/N5 — two stale clause-8b attributions scoped.** "8b can no longer fail for a divergence reason" is true of *this entry*, not the clause, which stays live for any entry open-coding two predicates. `moderation.ts` credited 8b as what makes publishing safe; for `chat-completion` the mechanism is now structural. Safety conclusion unchanged, mechanism name corrected.

### Two existing tests changed expectation, both deliberately

- The canonical-sample test asserted **every choice was captured**, which stopped being true. It now pins the **split**, with both halves pinned so neither drifts silently.
- The empty-extraction test asserted `textOutputWithheld` was **also** undefined — i.e. it encoded the silent-success behaviour. Its safety assertions (nothing scanned, nothing published) are untouched; only the diagnosability expectation flipped.

### A test that was passing for the wrong reason

`tools` was listed among "orchestrator input fields this entry does not expose". This PR **exposes** `tools`; the entry stayed green only because its fixture value was `[]`, which fails the schema's `.min(1)` bound. A green run would have told a reader tools were still rejected. Removed, with the reason recorded. `tool_choice` (snake_case) deliberately stays — the exposed param is `toolChoice`, and that distinction matters while the wire spelling is unconfirmed.

### Round-3 mutation results

| mutant | outcome |
|---|---|
| M-F re-split `extractText` to its pre-consolidation predicate | **silent → now fires** clause 8b's own error at load |
| M-K re-split `extractToolCalls`, widened on the id axis | **silent → now fires** clause 8b's own error at load |
| M-J delete `extractText`'s tool-arguments loop | fires (unchanged) |
| remove the no-silent-success invariant | exactly the **3** unpublishable cases red; all **3** controls green |
| widen the invariant's posture gate to every entry | exactly the **scoping control** red |

No survivors this round. Each read by failing-test name and assertion text, never by exit code.

`typecheck` 0 errors; **4412 tests / 177 files** green; prettier clean on all five changed files (the pre-existing 4 violations went too); eslint clean apart from the unused-`dbMock` warning already present at `main`.

🔴 **Still unproven and unchanged:** the `tool_choice` wire key has never been confirmed against a live orchestrator request. An ignored field fails silently.

---

## Round 4 — two regressions the previous round's own fix introduced

The round-3 no-silent-success invariant was correct in intent and wrong in two ways, both in the same three lines. Found by delta re-audit, both reproduced by execution before and after.

### F1 — the invariant fired on steps that had not finished

`attachModeratedStepTextOutputs` runs on **every** `pollWorkflow` and on `cancelWorkflow`, and its loop iterates `workflow.steps` with no status filter. A queued/processing chat step has no `output`, so `extractText` returns `[]` and the verdict releases empty — exactly the shape the invariant fires on.

Consequence: a block polling a perfectly healthy in-flight generation was told, **for the whole duration of that generation**, that the response had "completed" with nothing to display. A user-cancelled workflow got the same.

Now gated on the step's own `status === 'succeeded'`. `failed` / `expired` / `canceled` legitimately produce no output and are already explained by the workflow's own status — firing there would replace a correct explanation with a wrong one. An unreadable status falls back to the pre-invariant absent field, which is the safe direction.

### F2 — a real policy withhold could be masked, and reported as its own opposite

The reason slot was first-wins. That was provably harmless while every candidate value was the single constant `TEXT_OUTPUT_WITHHELD_MESSAGE`; round 3 put a second, **semantically opposed** value into the same slot without changing the merge. Step ORDER then decided which of two contradictory sentences the author read:

| step order | reason reported (pre-fix) |
|---|---|
| `[unpublishable, policy-hit]` | *"No content policy issue was found"* — **wrong**, a policy hit occurred |
| `[policy-hit, unpublishable]` | *"did not pass Civitai's content policy"* — correct |

Same content, same scanner verdict, opposite conclusions. Now two channels with **policy taking precedence**; within each kind the merge stays first-wins, preserving the pre-existing per-kind semantics.

🔴 **Severity, stated precisely:** the offending text was **withheld in both orderings**. This was a *reason-reporting* defect, not a content leak — but it pointed the app author away from the real cause, which is the direction that matters.

### F3 / F4 / F5

- **F3** — the message made a whole-response claim that a sibling step's `textOutputs` contradicted **in the same object**. The condition is per-step, the field is per-snapshot; the reason is now scoped to *"a step in this response"*.
- **F4** — `textOutputWithheld`'s wire-contract doc described only the withhold cause; both halves of that sentence are false on the unpublishable path. It now enumerates both causes, records that they are **not machine-separable** without string matching, and notes the status gate. This is the doc the `@civitai/app-sdk` mirror-writer and block authors read.
- **F5** — the third stale clause-8b attribution. Corrected to state what actually enforces containment here: a **shared predicate, structurally**, with clause 8b a load-time backstop only as wide as the shapes its canonical sample exercises.

### The fixtures were the root cause of F1 going unnoticed

Every step fixture in the suite carried **no `status` field at all**, though `WorkflowStep.status` is **required** on the generated `@civitai/client` type. A fixture omitting it is not orchestrator-shaped, and every test built on it was blind to the difference between a step that succeeded with no output and one that had simply not run yet. All fixtures now carry an explicit status, with a `runningChatStep` helper for the non-terminal case.

### F2's test needs both orderings, and here is why

The existing `A POLICY WITHHOLD KEEPS ITS OWN REASON` case uses **one** step and structurally **cannot** see this defect — with a single step there is no second candidate to win the slot. The new test asserts both orderings.

### Round-4 measurements

Before → after, same probe file against both trees:

| probe | before | after |
|---|---|---|
| `processing` workflow, un-run step | spurious *"response completed…"* | **no field** |
| `canceled` workflow, un-run step | spurious *"response completed…"* | **no field** |
| `[unpublishable, policy-hit]` | *"No content policy issue was found"* | policy message |
| `[policy-hit, unpublishable]` | policy message | policy message |
| publishable + unpublishable | *"This response completed…"* beside `textOutputs` | *"A step in this response…"* |
| single publishable step (control) | publishes | publishes (unchanged) |

Mutants — **no survivors**:

| mutant | outcome |
|---|---|
| drop the status gate | exactly the **4** F1 cases red, each by name |
| flip the precedence | **both** F2 orderings red |
| restore the single first-wins slot (the exact pre-fix defect) | **only** `unpublishable FIRST` red — the **isolating** mutant |
| revert the message wording | the F3 case red |

`typecheck` 0 errors (143s, heap cap named); **4419 tests / 177 files** green; prettier clean on all four changed files, verified against a **negative control** confirming prettier can still report a violation; eslint clean apart from the pre-existing unused-`dbMock` warning.

🔴 **Still unproven and unchanged:** the `tool_choice` wire key has never been confirmed against a live orchestrator request.

---

## Round 5 (docs only) — `be3255cca4`

The round-4 note on `textOutputWithheld` was itself wrong, and a delta audit caught it.

It read: *"CAUSE 2 IS GATED ON THE STEP'S OWN `succeeded` STATUS, so an in-flight or cancelled generation does NOT set this field."* **Only cause 2 is gated.** In `steps/moderation.ts` the `step.status === 'succeeded'` test guards the unpublishable branch inside the released arm; the withheld branch is the plain `else` beside it, with no status test. So a **cancelled generation can still set this field** — the orchestrator returns whatever text it had already produced, that text is still scanned, and a hit still withholds it.

Not a reading: a shipped, passing test already pinned it. `blocks.router.textOutputModeration.test.ts` cancels a workflow whose step succeeded and asserts `textOutputWithheld` **is** set.

This matters because the same doc block says, three paragraphs above, that it is what the `@civitai/app-sdk` mirror-writer and block authors read. A consumer who believed the old sentence and skipped rendering `reason` on cancel would **suppress a real content-policy explanation**.

The true half of the old note is kept, with its actual reason: a block polling a healthy in-flight generation does still see the field absent — not by status, but because a step that has produced no text yet reaches **neither** cause (`screenGeneratedText` short-circuits to RELEASED on an empty text set, and cause 2 additionally needs `succeeded`). The new wording deliberately says nothing about whether the orchestrator populates `output` on a still-running step: that was never verified, and no clause depends on it.

Also completed cause 1's sub-enumeration, which listed only a policy hit and a scanner error/timeout. **Seven** paths return that one message: over-cap, memo replay, scanner error, no-verdict, label-drift, label-error, and the policy hit.

🔴 This clause has now been wrong **twice**, both times by describing the *field* when the code only ever gated *one of its two causes*. The replacement carries a note to check each cause separately before editing it again.

**Docs only — no behaviour change, no new guard.** `typecheck` 0 errors (165s, heap cap named); **4419 tests / 177 files** green; the named cancel test re-run under a filter (1 passed / 11 skipped) with a zero-match control proving the filter discriminates; prettier clean on the changed file, with **two** negative controls — one confirming the invocation goes red on that exact file, one confirming `"All matched files use Prettier code style!"` also prints on a **zero-match** glob.

### DIFF
--- a/src/server/schema/blocks/workflow.schema.ts
+++ b/src/server/schema/blocks/workflow.schema.ts
@@ -733,17 +733,40 @@ export type BlockWorkflowSnapshot = {
    */
   textOutputs?: string[];
   /**
-   * Set when generated text WAS produced but is being withheld — a content-policy
-   * hit, or a scanner error/timeout (the scan fails CLOSED).
+   * Set when a text-posture step produced nothing the block can be given, with a
+   * user-facing sentence saying why. TWO distinct causes land here:
    *
-   * `reason` is a user-facing message, deliberately generic: it does not name the
-   * labels that triggered. The triggering labels + scores go to the per-app
-   * trigger log instead, which is where the actionable signal lives (an app whose
-   * outputs trigger constantly is telling you about its system prompt).
+   *   1. WITHHELD — text WAS produced and is being kept back: a content-policy
+   *      hit, or a scanner error/timeout (the scan fails CLOSED).
+   *   2. UNPUBLISHABLE — the step SUCCEEDED and the scan RELEASED, but the
+   *      entry's extractors could publish none of what came back (e.g. a
+   *      provider tool-call `id` outside the published charset). Nothing was
+   *      moderated in this case, and the message says so explicitly.
+   *
+   * 🔴 AN EARLIER VERSION OF THIS DOC DESCRIBED ONLY CAUSE 1 — "generated text
+   * WAS produced but is being withheld" — and both halves of that sentence are
+   * FALSE on cause 2: no text was produced, and no policy or scanner event
+   * occurred. This is the doc the `@civitai/app-sdk` mirror-writer and block
+   * authors read, so a consumer branching on it was being told the wrong thing.
+   * If a third cause is ever added, correct this list in the same change.
+   *
+   * `reason` is deliberately generic: it does not name the labels that
+   * triggered. The triggering labels + scores go to the per-app trigger log
+   * instead, which is where the actionable signal lives (an app whose outputs
+   * trigger constantly is telling you about its system prompt).
+   *
+   * 🔴 THE TWO CAUSES ARE NOT MACHINE-SEPARABLE without matching on the string,
+   * which is NOT a contract. A consumer needing to branch on the difference is
+   * the trigger to add a discriminated reason code — not before.
+   *
+   * 🔴 CAUSE 2 IS GATED ON THE STEP'S OWN `succeeded` STATUS, so an in-flight or
+   * cancelled generation does NOT set this field. A block polling a healthy
+   * running workflow sees it absent, as it did before cause 2 existed.
    *
    * Independent of `textOutputs` rather than mutually exclusive: a workflow with
    * two text steps can release one and withhold the other, and both fields then
-   * appear.
+   * appear. Because of that pairing the `reason` is scoped to "a step in this
+   * response", never to the response as a whole.
    */
   textOutputWithheld?: { reason: string };
   /**
@@ -759,8 +782,19 @@ export type BlockWorkflowSnapshot = {
    * 🔴 WHY PUBLISHING THESE IS NOT A HOLE IN THE SCAN, since the objects
    * themselves are never handed to a scanner: the entry's `extractText` returns
    * every `arguments` string that appears here, so the released verdict was
-   * computed over exactly that prose, and registry clause 8b fails the build if
-   * an entry ever breaks that containment. The tool NAME is not scanned and does
+   * computed over exactly that prose.
+   *
+   * 🔴 WHAT ENFORCES THAT, STATED AT ITS REAL STRENGTH. For `chat-completion`
+   * the guarantee is STRUCTURAL — both extractors derive from one shared
+   * predicate, so they cannot disagree. Registry clause 8b is a load-time
+   * BACKSTOP against an entry re-splitting them, and it is only as wide as the
+   * shapes its `canonicalOutputFor` sample exercises: a re-split on an axis the
+   * sample does not cover would still pass. An earlier version of this line said
+   * clause 8b "fails the build if an entry ever breaks that containment", which
+   * is wider than the implementation — the same unscoped overclaim corrected in
+   * `steps/moderation.ts` and `steps/chat-completion.step.ts`. Do not restate it.
+   *
+   * The tool NAME is not scanned and does
    * not need to be — it is pattern-bounded to `[A-Za-z0-9_-]` at both the param
    * schema and the extractor, so it cannot carry prose. Widening that charset
    * would make the name a free-text surface and is a moderation change, not a
--- a/src/server/schema/blocks/workflow.schema.ts
+++ b/src/server/schema/blocks/workflow.schema.ts
@@ -737,7 +737,10 @@ export type BlockWorkflowSnapshot = {
    * user-facing sentence saying why. TWO distinct causes land here:
    *
    *   1. WITHHELD — text WAS produced and is being kept back: a content-policy
-   *      hit, or a scanner error/timeout (the scan fails CLOSED).
+   *      hit, or any other fail-CLOSED scan outcome (scanner error or timeout,
+   *      no verdict at all, an over-cap payload, or requested labels that came
+   *      back missing or errored). Every one of those paths returns the SAME
+   *      message, so this cause is one string with several origins.
    *   2. UNPUBLISHABLE — the step SUCCEEDED and the scan RELEASED, but the
    *      entry's extractors could publish none of what came back (e.g. a
    *      provider tool-call `id` outside the published charset). Nothing was
@@ -759,9 +762,27 @@ export type BlockWorkflowSnapshot = {
    * which is NOT a contract. A consumer needing to branch on the difference is
    * the trigger to add a discriminated reason code — not before.
    *
-   * 🔴 CAUSE 2 IS GATED ON THE STEP'S OWN `succeeded` STATUS, so an in-flight or
-   * cancelled generation does NOT set this field. A block polling a healthy
-   * running workflow sees it absent, as it did before cause 2 existed.
+   * 🔴 ONLY CAUSE 2 IS STATUS-GATED, AND THE GATE READS THE STEP'S OWN
+   * `succeeded` STATUS — never the workflow's. CAUSE 1 IS NOT GATED AT ALL: in
+   * `steps/moderation.ts` the `succeeded` test guards the unpublishable branch
+   * only, and the withheld branch is the plain `else` beside it.
+   *
+   * So a CANCELLED generation CAN still set this field. The orchestrator returns
+   * whatever text it had already produced, that text is still scanned, and a hit
+   * still withholds it — `blocks.router.textOutputModeration.test.ts` pins
+   * exactly that, cancelling a workflow whose step succeeded and asserting the
+   * field IS set. A consumer that skips rendering `reason` on cancel would
+   * suppress a real content-policy explanation.
+   *
+   * A block polling a healthy in-flight generation does still see the field
+   * absent — but for a reason worth stating correctly rather than by status: a
+   * step that has produced no text YET can reach neither cause, because
+   * `screenGeneratedText` short-circuits to RELEASED on an empty text set, and
+   * cause 2 additionally needs `succeeded`.
+   *
+   * 🔴 THIS CLAUSE HAS NOW BEEN WRONG TWICE, both times by describing the FIELD
+   * when the code only ever gated ONE OF ITS TWO CAUSES. Before editing it
+   * again, check each cause against `steps/moderation.ts` separately.
    *
    * Independent of `textOutputs` rather than mutually exclusive: a workflow with
    * two text steps can release one and withhold the other, and both fields then
--- a/src/server/schema/blocks/workflow.schema.ts
+++ b/src/server/schema/blocks/workflow.schema.ts
@@ -3,6 +3,7 @@ import { buzzSpendTypes } from '~/shared/constants/buzz.constants';
 import type { BuzzSpendType } from '~/shared/constants/buzz.constants';
 import { REGISTERED_RECIPE_IDS } from '~/server/services/blocks/recipes';
 import { REGISTERED_STEP_IDS } from '~/server/services/blocks/steps';
+import type { BlockStepToolCall } from '~/server/services/blocks/steps';
 import {
   civitaiHostedImageUrlSchema,
   SOURCE_IMAGE_URL_MAX,
@@ -745,4 +746,37 @@ export type BlockWorkflowSnapshot = {
    * appear.
    */
   textOutputWithheld?: { reason: string };
+  /**
+   * Structured tool calls the model requested on a registered `'textOutput'`
+   * step — and which have PASSED the same output moderation scan `textOutputs`
+   * passes.
+   *
+   * 🔴 SAME SINGLE PRODUCER, SAME STRIPPING, SAME RULE.
+   * `attachModeratedStepTextOutputs` is the only writer; it strips any incoming
+   * value of this field before merging its own, and it attaches tool calls ONLY
+   * onto a released verdict. Do NOT set it anywhere else.
+   *
+   * 🔴 WHY PUBLISHING THESE IS NOT A HOLE IN THE SCAN, since the objects
+   * themselves are never handed to a scanner: the entry's `extractText` returns
+   * every `arguments` string that appears here, so the released verdict was
+   * computed over exactly that prose, and registry clause 8b fails the build if
+   * an entry ever breaks that containment. The tool NAME is not scanned and does
+   * not need to be — it is pattern-bounded to `[A-Za-z0-9_-]` at both the param
+   * schema and the extractor, so it cannot carry prose. Widening that charset
+   * would make the name a free-text surface and is a moderation change, not a
+   * validation one.
+   *
+   * WHERE IT APPEARS: `blocks.pollWorkflow` and `blocks.cancelWorkflow` — the
+   * same two wrapped procedures as `textOutputs`, for the same reason.
+   *
+   * OMITTED entirely when there are none, so every existing snapshot stays
+   * byte-identical.
+   *
+   * 🔴 WIRE CONTRACT: additive on the type `@civitai/app-sdk`'s `blocks/types.ts`
+   * mirrors, exactly like `textOutputs` and `modelSubstitutions` above. The SDK's
+   * inbound validator does not know it yet, so a block reads it only once that
+   * type is widened in the SDK repo — tracked separately; nothing here edits
+   * another repo.
+   */
+  toolCalls?: BlockStepToolCall[];
 };
--- a/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
+++ b/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
@@ -5,6 +5,7 @@ import {
   CHAT_COMPLETION_MODELS,
   CHAT_COMPLETION_PRICE_BUZZ,
   chatCompletionStep,
+  MAX_TOOL_ROUNDS,
   type ChatCompletionStepParams,
 } from '~/server/services/blocks/steps/chat-completion.step';
 import {
@@ -505,7 +506,17 @@ describe('chat-completion — extractText, against the REAL response', () => {
     expect(chatCompletionStep.extractText(withImages)).toEqual(['here you go']);
   });
 
-  it('does NOT read `tool_calls[].function.arguments`', () => {
+  it('🔴 DOES read `tool_calls[].function.arguments` — model-written text must be scanned', () => {
+    // 🔴 THIS ASSERTION IS INVERTED FROM WHAT IT USED TO BE, DELIBERATELY. It
+    // previously pinned `extractText` NOT reading tool calls, which was correct
+    // while the entry could not expose `tools` at all: the only safe thing to do
+    // with a field that could never legitimately appear was drop it.
+    //
+    // Exposing `tools` falsified that premise. An `arguments` string is free
+    // text the MODEL wrote, on its way to a tool and to the block's UI, so
+    // dropping it would mean either publishing it unscanned through the
+    // structured surface, or shipping a capability that charges and returns
+    // nothing. Both are worse than scanning it.
     const withTools = {
       output: {
         choices: [
@@ -518,7 +529,7 @@ describe('chat-completion — extractText, against the REAL response', () => {
         ],
       },
     };
-    expect(chatCompletionStep.extractText(withTools)).toEqual(['calling']);
+    expect(chatCompletionStep.extractText(withTools)).toEqual(['calling', '{}']);
   });
 
   it('is total over absent, null and malformed input', () => {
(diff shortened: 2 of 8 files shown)


ZacxDev:
## Audit round 1 — findings fixed, and the description was corrected

Pushed `db8b9378b8`. Flagging the description edit publicly rather than silently, because a reviewer may already have read the previous version and believed it.

### What the description used to claim, and why it was wrong

**1. Clause 8b.** It said the new registry clause "asserts at LOAD that every arguments string `extractToolCalls` can publish is also returned by `extractText`". True of the clause, **false as applied to the shipped entry**: `canonicalOutputFor` returned a content-only sample, so the containment loop iterated an **empty set** and passed vacuously.

That matters more than a normal guard bug, because three separate docstrings name that clause as *the* mechanism protecting the tool-call surface, and there is no runtime containment check anywhere else — the moderation handler validates shape, not containment. So the property rested on one unit test over one fixture, not on the gate the prose advertised.

Measured, both directions:

| | canonical sample | mutate shipped extractor to append `SMUGGLED` to every arguments string |
|---|---|---|
| before | content-only | registry load **passes clean** |
| after | carries a real tool call | registry load **throws** clause 8b's own message naming `chat-completion` |

The M2 mutation row in the description is evidence about the *synthetic* registry fixture, not the shipped entry — re-derive it rather than citing it.

**2. The unscanned-published-string set.** It enumerated that set as `{name}`. `id` was missing from it, and `id` was published **unbounded** — accepted on a bare non-empty-string test while `name` got both a length cap and a charset pattern, and while the *input* schema capped it at 64 chars. A 5,000-char `id` carrying prose and a literal `urn:air:` was published to the block verbatim.

The sharper half: on a response with empty `arguments`, `extractText` returns nothing, so `screenGeneratedText` short-circuits on the empty text set and **releases without ever calling the scanner**. The unbounded field was not merely unscanned — it was on a path where no scan ran at all.

### Also fixed

- **Empty/whitespace-only `arguments` are dropped**, which is what makes the stated containment property *true*. `extractText` required `trim().length > 0` while the structured extractor accepted any string, so the invariant clause 8b and three docstrings assert was literally false on unmutated code. `'{}'` — the ordinary no-argument encoding — is unaffected.
- **A tool result must answer a call that was actually made, and come after it.** Neither was validated. A provider rejects both, but only after the payload has been quoted, reserved against every cap and charged. Same reasoning the `toolChoice` clause already applies to an undeclared function name. The declared-id set is built in message order, so this is an **ordering** check, not merely membership.
- **`MAX_TOOL_ROUNDS`' docstring overclaimed** — it called itself "THE BILLING BOUND, AND WHY THE ENTRY CAN STAY `prepaidFixed`" and said a caller cannot loop past it. It bounds history depth in *one payload*; a block that drops or summarises the oldest tool result can submit indefinitely. What bounds the money is the per-submit live quote, the `buzzBudget` gate and the three cap reservations. Rounds are also not calls — one assistant turn may carry up to `MAX_TOOLS`.
- **The header's "`MAX_MESSAGE_CHARS` caps each"** was wrong; max accepted payload grew ~9× with tool calling. A per-message aggregate cap is deliberately declined, with the reasoning in the comment so it can be re-examined rather than re-discovered.
- Nits: the moderation shape guard now checks `type === 'function'` at runtime instead of trusting the type declaration, and the `JSON.stringify` rejection names the class instead of reporting "a circular structure" for a `BigInt` or a throwing `toJSON`.

### Test fixtures were fixed, not the guards weakened

The round-bound cases used bare `tool` messages answering nothing, which the new correlation guard correctly rejects. The `REJECTS +1` cases now use complete rounds so the **round bound is the only violation** — otherwise they would stay red with the round bound deleted, which is red for the wrong reason.

### Mutation results — every mutant killed by the test that targets it

| mutant | outcome |
|---|---|
| clause 8b containment, shipped entry | **killed** (was surviving pre-fix) |
| `id` charset | killed |
| `id` length cap | killed |
| both `id` bounds | killed |
| `arguments` containment | killed |
| correlation deleted | killed (3 tests) |
| correlation → membership only | killed — **only the ordering test**, which is the isolating mutant |
| moderation `type` literal | killed |

**Reported honestly:** the zero-scan case is depth-2 by construction and needs *both* the `id` and `arguments` defects reintroduced before it fails. It is not an independent kill and is not presented as one.

`pnpm typecheck` 0 errors · 667 tests pass across the step, router-moderation and blocks-schema suites · eslint clean apart from one warning already present at `main`.

**Still not verified, unchanged from the original description:** the `tool_choice` wire key has never been confirmed against a live request. An ignored field fails silently, so treat the feature as unproven end-to-end until that probe runs. Per-model tool support for one allowlisted model is unverified for the same reason and was deliberately left out of this round.

ZacxDev:
## Audit round 2 — description corrected again, and one surviving mutant

Pushed `e05ae23e5c`. Flagging this publicly rather than silently editing the body, since a reviewer may have read the previous version.

**Two claims in the description were wrong and have been corrected:**

1. It said **registry clause 8b asserts at LOAD that containment holds**. Clause 8b still runs, but both extractors now derive from one shared predicate, so they agree *by construction* and 8b can no longer fail for a **divergence** reason — it is a **refactor tripwire**, not a divergence detector. That is a genuine reduction in what it detects and it is a deliberate trade. It is *not* the vacuous shape round 1 found: the loop runs over real calls, and a mutant that diverges after the shared call still fires it at registry load.

2. It said tool **names and call ids** are unscanned because "the argument for both is the same" and "neither can carry prose". **Both halves were wrong.** `name` rests on two legs — caller-authored *and* charset-enforced on input — while `id` is **provider-generated** and rests only on the extractor's pattern. And the pattern does not prevent prose: a 63-char underscore-joined sentence passes `[a-zA-Z0-9_-]{1,64}`. What it buys is a hard bound on length and alphabet, which is what makes the residue acceptable — not an inability to say anything.

**The round-1 fix introduced a regression, now closed.** Aligning the extractors by *dropping* empty `arguments` made a no-argument tool call vanish: measured end to end, a response whose only choice was a tool call with `arguments: ''` produced a snapshot with neither `textOutputs` nor `toolCalls` nor `textOutputWithheld` — charged, `succeeded`, publishing nothing an app author could diagnose. Empty/whitespace arguments are now normalised to `'{}'`, which loosens nothing (containment compares against the *scanned* set, and `JSON.parse('')` throws while `JSON.parse('{}')` is the empty argument object the call means).

**🔴 One mutant SURVIVED and is reported rather than hidden.** Removing the `tool_call_id` regex alone leaves the suite green — every payload that could violate it is already rejected by the assistant-side regex or by the correlation guard, so it is defence-in-depth no test can isolate. The docstring now states that instead of implying coverage. Removing *both* id regexes together is caught.

**And one of my own new tests was wrong.** The input/output id-agreement test initially failed to kill its mutant, because it carried both an assistant `id` and a matching `tool_call_id` — each separately bounded, so removing either regex alone was masked by the other. Fixed with an assistant-only payload where nothing else can do the rejecting. Same class as the attribution fix: the headline id test used a fixture that was 4,900+ chars *and* contained spaces and colons, so the pattern and length mutants each died to it and it could not tell you which guard was load-bearing.

**Still not verified, unchanged:** the `tool_choice` wire key has never been confirmed against a live request, and per-model tool support is unchecked. An ignored field fails silently, so this remains unproven end to end regardless of the test count.

`typecheck` 0 errors; 5931 tests pass across the blocks, router and schema suites.

ZacxDev:
## Audit round 3 — description corrected again, publicly

Pushed `a8dedefa19`. Two substantive residuals from delta round 5 are fixed, and **the PR description has been corrected for the second time on the same sentence** — flagging it here because a reviewer may already have read the wrong version.

### What the description previously claimed, and why it was wrong

Round 2's body said registry clause 8b *"fires the moment anyone re-splits these predicates"* and that the claim was **mutation-tested**. The mutation actually run was the **deletion** of `extractText`'s tool-arguments loop. Round 3 executed two **genuine re-splits** and both loaded **clean**:

- restoring `extractText`'s pre-consolidation `trim().length > 0` predicate over *raw* arguments — literally re-creating the bug this PR shipped and then fixed;
- re-splitting `extractToolCalls` widened on the id axis, which publishes an `arguments` string `extractText` does not return — precisely the containment break 8b exists to catch.

The description was **wider than the implementation**: the same defect class as the round-1 vacuous-loop finding, one level up.

**Root cause was the sample, not the clause.** `canonicalOutputFor` carried one tool call with a conforming id, conforming name and non-empty arguments, so every divergence class this PR has fixed was invisible to it. It now also carries a choice with empty `arguments` and one with an out-of-charset `id` and *distinct* arguments. On the final code all three mutants fire 8b's own error at registry load. The docstring now claims only that, and states plainly that a re-split on an axis the sample does not cover would still pass.

🔴 Argument-string **distinctness across choices is load-bearing** — a colliding value is found in the scanned set and 8b goes silent, which is mechanically how the previous sample failed to arm it. Pinned by a test.

### The sample is no longer "all captured", and the body no longer says it is

The old banner read *"every choice here is copied verbatim from a real orchestrator response"*. That became false the moment adversarial values were added. A stale verbatim claim over invented data is worse than no claim, so the split is now explicit: **2 captured choices + 2 declared adversarial ones**. The both-wrong-blind argument still holds because every *shape* is captured; only *values* vary.

### The silent-success class, third round running

A malformed `id` or `name` is dropped by the shared predicate on both sides, so with the measured `finishReason: 'tool_calls'` shape (no `content` key) the response yields nothing anywhere — charged, `succeeded`, publishing none of the three fields.

Enumerating drop reasons has failed three times, so this round stops enumerating. `attachModeratedStepTextOutputs` now asserts the **outcome**, not the causes. The reason is a **distinct constant**, not the policy message: nothing was moderated here, and conflating them would mislead the app author *and* hide real withholds.

### Also corrected: a test that was passing for the wrong reason

`tools` was listed among "fields this entry does not expose". This PR **exposes** `tools` — the entry stayed green only because its fixture value was `[]`, failing the schema's `.min(1)` bound. A green run would have told a reader tools were still rejected.

### Round-3 mutations

| mutant | outcome |
|---|---|
| M-F re-split `extractText` | **silent → fires** 8b's own error |
| M-K re-split `extractToolCalls` (id axis) | **silent → fires** 8b's own error |
| M-J delete the arguments loop | fires (unchanged) |
| remove the no-silent-success invariant | exactly 3 cases red, all 3 controls green |
| widen its posture gate to every entry | exactly the scoping control red |

No survivors. `typecheck` 0 errors; **4412 tests / 177 files** green.

🔴 **Unchanged and still unproven: the `tool_choice` wire key has never been confirmed against a live orchestrator request.** An ignored field fails silently, so this could be correct and still inert.

ZacxDev:
## Correction — the PR description previously overstated the round-3 invariant

Posting rather than silently editing the body, since a reviewer may already have read the earlier version.

The round-3 write-up presented the no-silent-success invariant as closing the "charged, `succeeded`, published nothing" class. A delta re-audit found it introduced **two HIGH regressions of its own**, both in the same three lines, and the description did not mention either:

1. **It fired on steps that had not finished.** `attachModeratedStepTextOutputs` runs on every `pollWorkflow` and on `cancelWorkflow`, and the loop had no status filter — so a block polling a healthy in-flight generation was told for its entire duration that the response had "completed" with nothing to display. A cancelled workflow got the same message.

2. **It let a real content-policy withhold be masked by a sibling step's extraction failure.** The reason slot was first-wins, which was harmless while every candidate value was one constant; adding a second, semantically opposed value made step *order* decide which of two contradictory sentences the author read — including *"No content policy issue was found"* on a workflow where a policy hit had occurred.

Both are fixed in `fbdeeec18a` and the body now carries the full round-4 section.

🔴 **On severity, so nobody reads (2) as worse than it was:** the offending text was **withheld in both orderings**. This was a reason-reporting defect, not a content leak. What it did wrong was point the app author away from the real cause.

Two further corrections in the same commit:

- **`textOutputWithheld`'s wire-contract doc** described only the withhold cause. Both halves of that sentence — "text WAS produced", "a content-policy hit or scanner error" — are false on the unpublishable path. This is the doc the `@civitai/app-sdk` mirror-writer and block authors read, so it now enumerates both causes and records that they are not machine-separable without string matching.
- **A third stale clause-8b attribution** survived the previous round's sweep, in `workflow.schema.ts`. Same unscoped overclaim already corrected in two other files.

Also worth recording: the reason F1 went unnoticed is that **every step fixture in the suite omitted `status`**, though it is a required field on the generated `WorkflowStep` type. Those fixtures could not distinguish a step that succeeded with no output from one that had not run. They now carry explicit statuses.

ZacxDev:
### 📝 Description correction — the round-4 note on `textOutputWithheld` was wrong

Posting this rather than only editing the body, since a reviewer may already have read the previous version.

**What the PR previously claimed** (round-4 section, and the doc comment it shipped at `workflow.schema.ts`):

> CAUSE 2 IS GATED ON THE STEP'S OWN `succeeded` STATUS, so an in-flight or cancelled generation does NOT set this field.

**That is false for cause 1.** Only cause 2 is status-gated. In `steps/moderation.ts` the `step.status === 'succeeded'` test guards the *unpublishable* branch inside the released arm; the *withheld* branch is the plain `else` beside it and carries no status test. A **cancelled generation can still set `textOutputWithheld`** — the orchestrator returns whatever text it had already produced, that text is still scanned, and a policy hit still withholds it.

This is not an inference. A shipped, passing test already pinned the behaviour: `blocks.router.textOutputModeration.test.ts` cancels a workflow whose step succeeded and asserts the field **is** set.

**Why it is worth a comment rather than a silent edit:** the same doc block states, three paragraphs above, that it is the text the `@civitai/app-sdk` mirror-writer and block authors read. A consumer who believed the old sentence and skipped rendering `reason` on cancel would suppress a real content-policy explanation — i.e. the wrong half was the half a consumer would act on.

Fixed in `be3255cca4` (docs only — no behaviour change, no new guard). The true half of the old note is kept with its actual reason: a block polling a healthy in-flight generation does still see the field absent, not because of status but because a step that has produced no text yet reaches **neither** cause. The new wording deliberately asserts nothing about whether the orchestrator populates `output` on a still-running step — that was never verified here, and no clause now depends on it.

Also completed cause 1's sub-enumeration, which listed only a policy hit and a scanner error/timeout; **seven** paths return that same message (over-cap, memo replay, scanner error, no-verdict, label-drift, label-error, policy hit).

🔴 For anyone editing this comment next: **it has now been wrong twice, both times by describing the _field_ when the code only ever gated _one of its two causes_.** The replacement carries a note to check each cause against `steps/moderation.ts` separately.

PR #4483, opened 2026-08-28
feat(app-blocks): read-only chat-tool surface backed by the clamped catalog path (#398 AC5)

ZacxDev:
Adds the block-surface tool route for App Blocks chat tool calling (#398 AC5), so the loop landed in #4472 has something to call, plus the code-enumerated read-only allowlist AC5 asks for.

## 🔴 This deliberately does NOT build an MCP proxy

AC5's prose says *"Tool transport reaches MCP through a block-surface route."* A reviewer will expect a proxy in front of `mcp.civitai.com`. Measured against the live server, that design cannot be made safe:

- its `search_models` input schema has **no** `nsfw` / `browsingLevel` / maturity parameter and is `additionalProperties: false` — a proxy cannot pass the viewer's ceiling **in**;
- its results carry only `air` / `id` / `name` / `type` — **no maturity metadata at all** — so a proxy cannot filter what comes **back** either, short of re-resolving every returned id against our own database.

So a proxy under this path would be an unclamped catalog read. Instead the tools are backed by the same clamped path `/api/v1/blocks/models` already uses, which is strictly better for this purpose: the ceiling is applied server-side from the token claim and the client cannot widen it.

## 🔴 The maturity clamp is NOT inherited from the `/api/v1/blocks/*` prefix

This is the trap that makes the proxy *look* safe, and it is the single most important line in the change.

`withBlockScope` gives a route the JWT gate, revocation, `private, no-store`, opaque-origin CORS, the rate limit and the audit row. It does **not** give it the maturity clamp — `blocks/models.ts` says so in its own words, that `resolveCatalogBrowsingLevel` *"remains the whole authority surface"*. A route added under this prefix that forgets the call is unclamped while looking exactly as protected as its neighbours, and no wrapper-level test would notice.

This handler calls it itself, and the suite pins that with a **positive control**: a RED token must produce a *different* browsing level from a GREEN one. Without that control, "the clamp is applied" is indistinguishable from "the search is wired to a constant" — a handler that hardcoded the SFW flag would pass a green-only test.

## What is here

- **`services/blocks/tools/registry.ts`** — the read-only, first-party allowlist. Server-import-free (mirroring the stated rationale for `block-catalog-maturity`) so the security-relevant parts are unit-testable without dragging Prisma. Null-prototype lookup map, so a prototype key (`toString`) cannot resolve to a truthy inherited value and fail the guard open.
- **`pages/api/v1/blocks/tools.ts`** — `GET` returns the declarations, `POST { name, arguments }` executes one. Unknown name rejected; arguments validated **strictly** against that entry's own schema; rate limit shared with the catalog endpoint deliberately, so a block cannot get a second budget by routing the same search through here.
- **Single-sourced schema.** The `parameters` handed to the model is `z.toJSONSchema` of the *same* zod schema the route validates against (the call form the repo already uses in `moderator-endpoint`). A model cannot be shown a contract *this route* does not enforce, structurally rather than by test. ⚠️ **Narrower than it first read, and corrected after review:** that holds for what the GET *serves*. `blockToolDeclarations()` has no consumer other than this route's GET, so nothing constrains what a block puts in the chat step's `tools[]` — it can declare `search_models` to its own model with a different schema. The route still enforces its own contract; the claim is about the served declaration, not about every declaration a block can make.

## The AIR strip, and what it unblocks

`containsAirReference` throws FORBIDDEN on the literal `urn:air:` anywhere in a submitted step input — on the estimate path and the submit path both. A tool result is replayed as a `role:'tool'` message, so a result carrying the literal doesn't degrade, it makes the round trip **impossible**. A live MCP `search_models` call at `limit: 1` returned **31** of them.

Results are therefore projected through an **allowlist, never a denylist** — the raw search row carries every version, file, hash, download url and AIR, so a denylist is one upstream field addition away from leaking, and the raw row could never fit the replay budget anyway. AIR literals are then scrubbed as a residual pass (a user-authored model *name* can contain one), case-insensitively and including object keys, because the scan it defends against is both.

The result is bounded to stay replayable within `MAX_MESSAGE_CHARS`, dropping **whole rows** rather than truncating a serialized string into invalid JSON.

## Test coverage

**44 new tests** (26 registry + 16 endpoint + 2 drift-guard updates). Full run: **6276 passed / 270 files** across `services/blocks`, `tests/api/v1/blocks`, `middleware/__tests__`, `schema/blocks`, `routers/__tests__`. `pnpm typecheck` 0 errors.

Red→green is not meaningful for a module that did not exist, so every guard was **mutation-tested** instead. ⚠️ **The original claim of "16 mutants, no survivors" was a claim about the mutants imagined, not about the guards.** A blind audit ran 39 and **7 survived**; those gaps are closed in the round below, and the mutants are now 12 with no survivors. The original 16, each confirmed to fail with *that guard's own* error:

| mutant | killed by |
|---|---|
| plain-prototype lookup map | prototype-keys test |
| args schema not `.strict()` | unknown-key + maturity-key tests |
| no AIR scrub on projection | user-authored-name + round-trip |
| scrub made case-sensitive | case-insensitivity test |
| scrub skips object keys | keys test |
| projection passthrough | both allowlist tests |
| char budget removed | char-budget test *(see below)* |
| budget raised above the message cap | constant-link test |
| `parameters` hand-written | derivation test |
| clamp removed from handler | 4 clamp tests |
| allowlist check removed | unknown-name + prototype-key |
| args validation skipped | invalid-arguments test |
| rate limit ignored | 429 test |
| projection skipped in route | AIR + files tests |
| meili given an unclamped level | meili-clamp test |
| `allowOpaqueOrigin` dropped | CORS wiring guard |

**No survivors — but one mutant survived the first time and the fix was to the test, not the code.** Removing the char-budget break left the whole suite green: the fixture rows were ~339 chars, so ten came to ~3,400 against a 6,000 budget, and the **item cap** was doing all the work while the budget branch never executed. The test named "drops whole rows" was passing for a reason other than its name. Rebuilt with maximal rows (691 chars each, measured — budget binds at 8, below the item cap of 10) and an isolating assertion that fewer than the item cap came back, which can only be true if the budget bound. The mutant then dies to exactly that test.

**Positive controls**, because three of these tests would otherwise pass vacuously:

- the clamp test (RED must differ from GREEN, else "clamped" ≡ "wired to a constant");
- the AIR test (the raw fixture is asserted to *contain* `urn:air:` before asserting the projection does not — otherwise it passes on any fixture that never had one);
- the round-trip test's mirror (an **unprojected** result must be *refused* by `containsAirReference`, proving the guard is not inert).

**The round-trip test is the one that matters most:** a projected result placed in a `role:'tool'` message clears `containsAirReference` **and** the chat-completion param schema. Each half passing alone would not show the loop closes.

## Two of the repo's own drift guards fired, and both were actioned

Worth recording because they did their job:

1. **`KNOWN_STATIC_ENDPOINT_SEGMENTS`** — the new route's path segment was not in the audit-log allowlist. Two tests fired: one derives the set from the route files, and a second **pins the literal list** so adding a route is a deliberate edit rather than a silent widening. Both updated.
2. **The catalog opaque-origin CORS wiring guard**, which asserted exactly two endpoints opt in. The tool surface is called directly from the sandboxed iframe, so it needs the same opt-in — and, as that test's own comment says, dropping it would fail only in prod (405 on the preflight) with every test still green.

## Not done

- **No sensei integration** (#398 AC6) and no demonstration against the deployed app (AC7).
- **Only `search_models` is registered.** More read-only catalog tools are additive; each new one needs its backing route confirmed maturity-clamped **individually**, because the clamp is per-handler.
- **`@civitai/app-sdk` is not updated.** If a block should get these declarations through the SDK rather than the `GET`, that is a follow-up mirror.
- One pre-existing eslint error in `catalog-cors-wiring.test.ts` (`acquireBulkheadSlot: () => () => {}`, `no-empty-function`) is **not** introduced here — verified by running eslint against the `origin/main` copy of that file, which reports the identical error at the same line. Left alone as out of scope.

---

### The CI failure: `rest-error-envelope-ledger`

`Unit tests (3)` failed on a repo-wide guard that fails when any route under `src/pages/api` serializes an error object or its text into a response body.

🔴 **The local run structurally could not see it.** The suite was globbed over the blocks paths; **CI runs the whole `unit*` project**. So a green local number was a claim about a *different population* — and it happened to match the PR's own figure, which made it look like full coverage.

The offender was `{ error: e.message }` on the meili-timeout arm, matched by the ledger's `error|message: <ident>.message` shape. Established by mutation rather than by reading: reverting only that line reproduces the exact CI diff (`+ v1/blocks/tools.ts`), restoring it goes green.

That value is our own first-party string, not driver text — so this was the ledger **over-reporting**, not a live disclosure. The ledger's own docstring says the remedy for over-reporting is a baseline entry. **It is deliberately not taken:** a brand-new route does not belong in a list of *pre-existing* unfixed offenders, and writing the literal removes the offence instead of recording it. `blocks/models.ts` has the identical line and *is* baselined; it was not copied back.

### Guards that were unpinned (each now fails with its own assertion)

| guard | why it mattered |
|---|---|
| char budget **at the seam** | `boundToolResult` was unit-tested in isolation; no endpoint test asserted the response was bounded, so a passthrough survived all 42 tests. Losing it yields a result the block **cannot replay**, with no diagnostic. |
| `!claims` → 401 | **This is the gate, not defence in depth.** `withBlockScope` falls through to the wrapped handler both when no bearer is present and when the runtime flag is off — so deleting it serves the declarations **unauthenticated**. The comment claiming otherwise is corrected. |
| meili-timeout 503 + `Retry-After` | mutating the status to 200 survived |
| wire `.strict()` | an unknown top-level key was unrejected-and-untested |
| name / creator / tag length bounds, and `str()`'s truncation branch | all four widened to 100000 with the suite still green |

### Two defects introduced *while writing those tests*, recorded rather than quietly fixed

Both are the same shape the audit exists to catch:

1. The bound tests first derived their expected length **from the exported constant**, so widening the constant moved the assertion with it and all three mutants still survived. Expectations are now **literals**, with a drift guard asserting the literals are the shipped constants.
2. The AIR-in-arguments test first sent an invalid **value**, which zod reports without echoing — so no AIR ever reached the scrub and that mutant survived too. It now sends an unrecognized **key**, which zod does echo, with a positive control asserting the echo happened.

### Also in this round

- The wire `name` is pattern-bounded via the chat step's **exported** `TOOL_NAME_PATTERN` (imported, not re-spelled), which makes both 400 reflections structurally unable to carry `urn:air:`; the free-text argument reflection is scrubbed.
- `str()` no longer cuts through a **surrogate pair** — emoji names previously emitted lone surrogates.

### Not done / still open

- **`preview / component-tests` is failing and I could not attribute it.** It is marked *"report-only, not blocking"*. `preview / deploy` **passed**, so it is not a deploy cascade; the same check **succeeds** on PRs #4476 and #4477, so it is not blanket-broken. It cannot be reproduced locally (no Playwright browser in this environment — the run reports `Tests no tests`), and the status is posted by the preview pipeline with the site as its target URL, so there is no log to read. **Stated as unresolved rather than assumed unrelated.**
- Nits left as follow-ups: GET is unrate-limited; the rate-limit docstring still lists only `models`/`images`; the metrics label cannot separate the free GET from the catalog POST; `neutralizeAirLiterals` has no recursion budget (unreachable today).

---

## Round 3 (`66f6024214`) — a delta re-audit found six survivors and four false comments

The delta audit of round 2 found **no behavioural defect**, and confirmed the ledger fix, the seam pin, the 401 pin, the `TOOL_NAME_PATTERN` narrowing and the surrogate fix all hold. What it did find was **6 surviving mutants out of 27** against a round-2 body that claimed *"12 mutants, no survivors"* — a reminder that the number is a claim about the mutants *imagined*, not about the guards.

### The two that mattered — both "the fixture cannot reach the branch"

| mutant | why it survived | fix |
|---|---|---|
| delete `neutralizeAirLiterals` from the **wire** 400 | the test sent a `name` failing `.regex(...)`, and zod does **not** echo the input on `invalid_format` — so no AIR was ever produced and the assertion passed without the scrub running | the strict-key test now uses an **AIR literal as the unrecognized key** (zod *does* echo those), with a positive control asserting the echo happened |
| widen `str()`'s surrogate range to admit **low** surrogates | the only astral fixture was pure emoji, where high surrogates always land on even indices — so `charCodeAt(cut-1)` is *always* high and the range's upper bound never executes | one leading BMP char shifts the parity. Measured: HEAD → 120 chars, no lone surrogate; mutant → 119 ending in an orphan. Added at the name **and** tag bounds |

🔴 The first is the **same defect fixed one call site over in round 2**, reproduced at the neighbouring call site in the same commit.

### The rest

- **`type` and `baseModel` were left on a bare `40`** while the other three projected-string bounds were exported — so two of five sat outside the block whose docstring claimed to cover "individual projected strings", and widening either survived. Now named constants with reaching fixtures; all five are in the drift guard.
- **`bodyParser sizeLimit` was reported covered and was not** — nothing in the suite mentioned it. Next enforces it in a pipeline `node-mocks-http` does not run, so **no unit test here can observe the limit rejecting a body**. Pinned as a literal instead, with that limitation stated in the test rather than implied.
- **The wire `name` length bound now asserts its own zod issue.** A first version checked only for a 400 — and the widen-the-bound mutant survived, because an over-long name then parses and 400s as an *unknown tool*. Both arms returned 400, so the probe fired identically on its own control.

### Comments corrected — each measured, not reasoned

- The `!claims` gate names **three** fall-throughs, not two: `if (!bearer || !isBlockJwt(bearer))` is two conditions (the second being the legacy opaque API-key path) plus the runtime flag. 🔴 This comment had now been wrong **twice** about the same gate, so it says so and says to count them in the middleware.
- A test comment claimed each fixture is built from the exported constant. They are not — deliberately, because a constant-derived fixture moves *with* a widen-the-bound mutant and stops discriminating. Both copies corrected.
- An assertion claimed *"a lone surrogate does not survive a JSON round trip intact."* **Measured false**: well-formed `JSON.stringify` escapes it as `\udXXX` and `JSON.parse` restores it exactly, so it held for every string and carried no information — and it contradicted the source comment on `str()`. Replaced with a positive control on the detector itself.

### Verification

- **Mutation: 14 mutants, 0 survivors.** Baseline confirmed green before scoring; harness validated with a bad-mutant guard on occurrence count, and verdicts read from the runner's own result lines rather than an exit code.
- **Not mutated**, stated so the number is not read as coverage: the projection field allowlist, the rate-limit and audit-row wiring, and anything behind the passthrough `withBlockScope` mock.
- **Full `unit*` project as CI runs it: 1517 files / 23873 tests passed, 0 failures** (+6 tests). Round 2's local run globbed only the blocks paths and structurally could not see the ledger failure CI caught; this is the CI population, named.
- `pnpm typecheck` **0 errors**. ESLint **0 errors** (10 pre-existing `no-explicit-any` warnings). Prettier clean, **with a negative control** confirming the check can go red.

---

## Round 4 (`7759fa6bd9`) — the last delta audit was CLEAN; this is a tidy, not a fix

The delta re-audit of round 3 found **no behavioural defect** and ended the audit ladder. It proved the point mechanically rather than by reading: the **comment-stripped diff of `tools.ts` was empty** (all 18+/8− were comment) and `registry.ts`'s only semantic change was two constants replacing four literal `40`s — so round 3's source delta was behaviour-preserving *by construction*.

This round closes the three low findings that audit surfaced. It is **comment- and test-only**, and the same comment-stripped-diff proof is re-run below: **zero payload lines changed**, which is why no further audit round follows it.

### 1. The endpoint label — the sole surviving mutant in the entire suite

`endpoint: 'tools'` → `'models'` typechecks (`AppBlockEndpoint` is a union carrying both) and survived **all ~23.9k tests**.

The blast radius is narrow but silent, and worth stating precisely: the audit row takes its endpoint from `normalizeEndpoint(req.url)` in the middleware, **not** from these opts — so `BlockScopeInvocation` rows stay correct. What degrades is the Prometheus RED pair (`civitai_app_block_requests_total` / `_request_duration_seconds`), which would merge tool traffic into `models` with nothing anywhere reporting a fault.

`catalog-cors-wiring.test.ts` already asserted `allowOpaqueOrigin` and `requiredScope` across all three catalog routes — but both are `for` loops over a property the routes **share**, so the one value that must **differ** between them was never checked. Now asserted per-route in import order.

Mutant **killed**, by the new assertion's own message:

```
AssertionError: expected [ 'models', 'images', 'models' ] to deeply equal [ 'models', 'images', 'tools' ]
```

🔴 This is the area round 3 named as *unmutated*. Naming an uncovered area honestly does not cover it.

### 2. The `!claims` fall-through comment — third revision, now read from the function

Round 3 fixed the **count** (three, not two) but described case 2 as *"a bearer that is NOT a 3-part JWS — the legacy opaque API-key path"*, which is narrower than the branch. Read directly against `isBlockJwt`, it returns false for:

- anything that is not three **non-empty** dot-separated segments;
- a header that fails to base64url-decode or `JSON.parse`;
- a decodable header whose `alg` is not `RS256` **or** whose `typ` is not `JWT`.

So a well-formed JWS signed with the wrong alg falls through too, and the opaque API-key path is *one instance* of the case rather than its definition. The error direction was always safe — it **understated** the fall-through set, so it could only make the gate look more necessary — but this comment has now been wrong twice and imprecise once, so it is written from the function instead of from a remembered shape.

### 3. The `.max(64)` coupling — recorded, deliberately not closed

The wire spells the length bound as a literal against the chat step's `MAX_TOOL_NAME_CHARS`, which is **module-private** and therefore cannot be imported — so the two agree only by inspection. This is the same class the `TOOL_NAME_PATTERN` import closed, one axis over.

Closing it requires **exporting** that constant: a source change to a file merged in #4472. Left deliberate and named at the call site, identifying both sides, so the next person closes it on purpose rather than rediscovering it.

### Verification

- **Comment-stripped source delta is EMPTY**, re-proved with a **negative control** (appending a real statement makes the strip emit it). This is the property that makes another audit round unnecessary — proved, not assumed.
- Endpoint mutant **killed** with its own assertion message (above); baseline confirmed green **before** scoring, on an isolated `cp -a` copy with `.git` removed.
- **Full `unit*` project as CI runs it: 1517 files passed / 1 skipped (1518); 23873 passed / 26 skipped (23899), zero failures.** Unchanged count — an existing test was strengthened rather than a new one added.
- `pnpm typecheck` **0 errors**. Prettier clean on both files, **with a negative control** confirming the check can go red (its success line also prints on zero matches).
- ESLint reports **one pre-existing** `no-empty-function` error on an untouched mock line. Confirmed pre-existing by running eslint against the `origin/main` copy of that file, which reports the identical error — this commit only shifted its line number.

### Open items after this round

**Closed here:** the endpoint-label pin; the `!claims` comment's precision; the `.max(64)` coupling recorded.

**Still open, all pre-existing and none blocking:**

- **`preview / component-tests` remains unattributed** (see round 2) — report-only, not a deploy cascade, succeeds on #4476/#4477, not locally reproducible. Stated as unresolved rather than assumed unrelated.
- GET is unrate-limited (the expensive POST path is limited; each GET still costs a revocation read and an audit row).
- The rate-limit docstring still lists only `models`/`images` as the bucket's callers.
- The metrics label cannot separate the free GET from the catalog POST.
- `neutralizeAirLiterals` has no recursion budget — unreachable today (called only on a flat projected row), but the function is exported.
- `Math.min(limit ?? DEFAULT_LIMIT, MAX_TOOL_RESULT_ITEMS)` is redundant behind the args schema's own `.max(MAX_TOOL_RESULT_ITEMS)`; and `DEFAULT_LIMIT = 5` is unpinned against the literal `"default 5"` shown to the model in the tool declaration. Both judged harmless by the audit and left as-is.
- `@civitai/app-sdk` is not updated — a follow-up mirror if blocks should get the declarations through the SDK rather than the `GET`.

### DIFF
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -20,9 +20,11 @@ import {
   blockToolDeclarations,
   boundToolResult,
   getBlockTool,
+  neutralizeAirLiterals,
   projectModelForTool,
   MAX_TOOL_RESULT_ITEMS,
 } from '~/server/services/blocks/tools/registry';
+import { TOOL_NAME_PATTERN } from '~/server/services/blocks/steps/chat-completion.step';
 import { MetricTimeframe } from '~/shared/utils/prisma/enums';
 import { constants } from '~/server/common/constants';
 
@@ -74,7 +76,19 @@ export const config = { api: { bodyParser: { sizeLimit: '8kb' } } };
  */
 const callSchema = z
   .object({
-    name: z.string().min(1).max(64),
+    // 🔴 PATTERN-BOUNDED AT THE WIRE, and not merely for tidiness. Both 400
+    // bodies below REFLECT this value back to the caller, and the caller is an
+    // untrusted sandboxed iframe that will hand our reply to a chat model as a
+    // `role:'tool'` message. `containsAirReference` (the chat step's entitlement
+    // guard) throws FORBIDDEN on the literal `urn:air:` ANYWHERE in a submitted
+    // step input — so an unbounded `name` lets a caller compose a 400 body it can
+    // never replay, with no diagnostic explaining why. The pattern makes the
+    // reflection structurally inert instead of relying on the scrub below.
+    //
+    // Same pattern the chat step enforces on a declared tool name, IMPORTED
+    // rather than re-spelled: `registry.ts` already documented the coupling in
+    // prose, and a second copy is how the two drift.
+    name: z.string().min(1).max(64).regex(TOOL_NAME_PATTERN),
     arguments: z.unknown().optional(),
   })
   .strict();
@@ -84,8 +98,16 @@ const DEFAULT_LIMIT = 5;
 const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: NextApiResponse) {
   const claims = (req as BlockScopedNextApiRequest).blockClaims;
   if (!claims) {
-    // withBlockScope only invokes this handler with a valid block JWT; defense
-    // in depth, mirroring blocks/models.ts.
+    // 🔴 THIS IS THE GATE, NOT A BACKSTOP — the previous comment here said
+    // "defense in depth", and that was wrong in the direction that matters.
+    // `withBlockScope` does NOT reject and return on its own in two cases: when
+    // no bearer token is present, and when the `app-blocks-runtime-enabled`
+    // flag is off. In BOTH it FALLS THROUGH to the wrapped handler with
+    // `blockClaims` unset. So deleting these five lines does not lose a
+    // redundant check — it serves the tool declarations unauthenticated.
+    //
+    // Pinned by its own test; a mutant that removes it previously survived the
+    // whole suite, which is why the comment is now explicit about what it holds.
     res.status(401).json({ error: 'Block token required' });
     return;
   }
@@ -103,7 +125,16 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
 
   const parsed = callSchema.safeParse(req.body);
   if (!parsed.success) {
-    res.status(400).json({ error: parsed.error });
+    // 🔴 `.message`, NOT the ZodError object. Measured: `JSON.stringify({error:
+    // zodError})` emits `{"name":"ZodError","message":"[…issues…]"}`, and those
+    // issues echo the caller's own keys verbatim (`"keys":["<their key>"]`). That
+    // is caller input rather than server internals — no Prisma table/column, no pg
+    // row value — so it is not the disclosure the ledger is named for. It is still
+    // an error OBJECT on the wire, and the scrub below is what makes the echo safe
+    // to replay rather than the shape being harmless.
+    res.status(400).json({
+      error: neutralizeAirLiterals(`invalid request body: ${parsed.error.message}`),
+    });
     return;
   }
 
@@ -113,6 +144,10 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
   // inherited value.
   const tool = getBlockTool(parsed.data.name);
   if (!tool) {
+    // Not scrubbed, and it does not need to be: `name` cleared
+    // `TOOL_NAME_PATTERN` above, which admits no `:` and therefore cannot carry
+    // `urn:air:`. The scrub on the two bodies that CAN echo free text is what
+    // does the work; adding one here would imply this reflection is unbounded.
     res.status(400).json({
       error: `unknown tool '${parsed.data.name}'`,
     });
@@ -124,8 +159,14 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
   // contract this route does not enforce.
   const args = tool.argsSchema.safeParse(parsed.data.arguments ?? {});
   if (!args.success) {
+    // SCRUBBED: unlike `name`, an argument value is free text (`query`), and zod
+    // echoes offending values and unrecognized keys into `.message`. A caller
+    // that sends `query: "urn:air:…"` would otherwise get a 400 body it cannot
+    // replay as a `role:'tool'` message.
     res.status(400).json({
-      error: `invalid arguments for tool '${tool.name}': ${args.error.message}`,
+      error: neutralizeAirLiterals(
+        `invalid arguments for tool '${tool.name}': ${args.error.message}`
+      ),
     });
     return;
   }
@@ -168,7 +209,21 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       } catch (e) {
         if (e instanceof ModelSearchMeiliTimeoutError) {
           res.setHeader('Retry-After', '2');
-          res.status(503).json({ error: e.message });
+          // 🔴 A LITERAL, NOT `e.message`, AND THAT IS THE POINT — CI caught the
+          // difference. `rest-error-envelope-ledger` flags `{ error: <ident>.message }`
+          // as a class, because the shape is indistinguishable from serving a
+          // DRIVER-authored message (a Prisma error carries table + column; a pg
+          // 23505 carries the offending ROW VALUE). Here the value happens to be
+          // our own first-party string, so this was over-reporting rather than a
+          // live disclosure — but the ledger's own docstring says the remedy for
+          // that is a baseline entry, and a brand-new route does not belong in a
+          // list of PRE-EXISTING unfixed offenders. Writing the string here costs
+          // nothing, removes the offence instead of recording it, and cannot
+          // drift if `ModelSearchMeiliTimeoutError`'s text changes.
+          //
+          // `blocks/models.ts` has the identical `e.message` line and IS
+          // baselined. Do not copy it back.
+          res.status(503).json({ error: 'Model search is temporarily overloaded — please retry.' });
           return;
         }
         throw e;
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -88,6 +88,16 @@ const callSchema = z
     // Same pattern the chat step enforces on a declared tool name, IMPORTED
     // rather than re-spelled: `registry.ts` already documented the coupling in
     // prose, and a second copy is how the two drift.
+    //
+    // ⚠️ THE LENGTH BOUND IS STILL A SECOND COPY — the same class this line's
+    // `.regex` just closed, one axis over. `64` here is the chat step's
+    // `MAX_TOOL_NAME_CHARS` (`~/server/services/blocks/steps/chat-completion.step`),
+    // which is module-private, so it cannot be imported today and the two agree
+    // only by inspection. Nothing is broken: the values match, and a drift would
+    // narrow or widen only what this wire accepts before the registry lookup
+    // rejects an unknown name anyway. Closing it means EXPORTING that constant —
+    // a source change to a file merged in #4472 — so it is left deliberate and
+    // named rather than done as a drive-by here.
     name: z.string().min(1).max(64).regex(TOOL_NAME_PATTERN),
     arguments: z.unknown().optional(),
   })
@@ -104,7 +114,13 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
     // redundant check, it serves the tool declarations unauthenticated:
     //
     //   1. no bearer token present;
-    //   2. a bearer that is NOT a 3-part JWS — the legacy opaque API-key path.
+    //   2. a bearer `isBlockJwt` rejects. READ THAT FUNCTION, it is wider than
+    //      "not a 3-part JWS": it returns false for anything that is not three
+    //      NON-EMPTY dot-separated segments, for a header that fails to
+    //      base64url-decode or JSON.parse, AND for a decodable header whose
+    //      `alg` is not `RS256` or whose `typ` is not `JWT`. So a well-formed
+    //      JWS signed with the wrong alg falls through here too — the legacy
+    //      opaque API-key path is one instance of this case, not its definition.
     //      (1) and (2) are one branch, `if (!bearer || !isBlockJwt(bearer))`,
     //      which is why they are easy to count as one;
     //   3. `app-blocks-runtime-enabled` off — including absent, and Flipt down,
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -98,16 +98,26 @@ const DEFAULT_LIMIT = 5;
 const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: NextApiResponse) {
   const claims = (req as BlockScopedNextApiRequest).blockClaims;
   if (!claims) {
-    // 🔴 THIS IS THE GATE, NOT A BACKSTOP — the previous comment here said
-    // "defense in depth", and that was wrong in the direction that matters.
-    // `withBlockScope` does NOT reject and return on its own in two cases: when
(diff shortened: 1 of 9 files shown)


ZacxDev:
## Correction to this PR's description (round 2)

Posting publicly rather than editing silently, since a reviewer may already have read the previous version.

**1. "16 mutants, no survivors" was overstated.** It was a claim about the mutants *imagined*, not about the guards. A blind audit ran **39** and **7 survived** — the char budget at the seam, the `!claims` 401, the meili-timeout status, the wire `.strict()`, and all three per-field length bounds plus `str()`'s truncation branch. Those are closed in this round.

**2. "A model cannot be shown a contract the route does not enforce, structurally rather than by test" was narrower than written.** It holds for what the GET *serves*. `blockToolDeclarations()` has no consumer other than that GET, so nothing constrains what a block puts in the chat step's `tools[]` — it can declare `search_models` to its own model with a different schema. The route still enforces its own contract; the claim is about the served declaration.

**3. `{ error: e.message }` on the meili-timeout arm tripped `rest-error-envelope-ledger` and CI was red.** The value is our own first-party string rather than driver text, so the guard was over-reporting rather than catching a live disclosure — but a brand-new route does not belong in that baseline, so the offence was removed instead of recorded. `blocks/models.ts` has the identical line and *is* baselined; it was not copied back.

Two defects were introduced *while writing the new tests* and are recorded in the body rather than quietly fixed — the bound tests initially derived their expectations from the constant they were testing, and the AIR-in-arguments test initially used a fixture that could not reach the scrub. Both survived their mutants until corrected.

**Still open:** `preview / component-tests` fails and I could not attribute it. It is report-only/non-blocking, `preview / deploy` passed (so not a cascade), and the check succeeds on #4476 and #4477 (so not blanket-broken) — but it cannot be reproduced locally and its status carries no readable log. Flagged as unresolved rather than assumed unrelated.

PR #4486, opened 2026-08-29
ci(retool): gate the sanitiser rules that were only prose

ZacxDev:
Follow-up to #4476.

That PR removed three classes of personal data from `docs/moderator-app/retool-exports/**` — staff real names, banned-user IP addresses and end-user account ids. The rules meant to prevent them live in `raw/README.md` as **prose**, and prose is not a gate: the next Retool re-download would have put all three back with nothing to stop it.

This adds the mechanical half, in the shape this repo already uses (`submodule-pin-guard`, the typecheck gates).

## What it checks

Over `docs/moderator-app/retool-exports/**`, on PRs that touch it:

| # | check | catches |
|---|---|---|
| 1 | **denylist** | known-bad values, matched as **salted hashes** |
| 2 | **ipv4** | any IPv4 literal outside RFC5737/RFC1918 |
| 3 | **credential** | the README's shape rules, as code |
| 4 | **placeholder** | the redaction placeholders having gone missing |

**Check 4 is the one that catches the headline risk** — a wholesale re-download overwrites a sanitised file, so the placeholders vanish — and it needs no secret at all.

## 🔴 What it deliberately cannot do

It **cannot** detect a *novel* staff real name or a *novel* bare account id. Neither has a shape to match on, which is precisely why both walked straight through the existing shape-based pass: a name looks like prose and an account id looks like any other integer.

So a green run means "no known-bad value, no out-of-range IP, no credential shape, placeholders intact". It does **not** mean "these exports contain no personal data". That is stated in the gate header, the workflow header and the README, because a check that reads as coverage while providing none is worse than no check — it stops anyone looking.

## Why the denylist is hashed, and the salt is not committed

A plaintext denylist of staff names in a public repo re-publishes exactly what it protects. Hashing alone is not enough either: a person's name is a tiny search space, so a **committed salt would be dictionary-crackable in seconds**.

Hashes are committed; the salt comes from the CI secret `RETOOL_SANITISER_SALT`. With no salt (a fork PR, or before the secret exists) check 1 reports `SKIPPED` **loudly** and is not counted as a pass — checks 2–4 still run and still block.

**The denylist ships empty.** Populating it needs the private name key and a chosen salt, neither of which belongs in this PR:

```
export RETOOL_SANITISER_SALT=…
node scripts/ci/retool-sanitiser-gate.mjs --hash '<value>'   # commit the digest only
```

## Verification

- **Positive control** — clean tree exits 0, scanning 32 files. The file count is asserted too: a gate wired to nothing also reports zero findings.
- **Negative controls, one per class, each failing with its OWN message** — out-of-range IPv4, secret assignment, denylisted value, missing placeholder. Each planted in isolation so no mutant dies to a different check's error.
- **37 tests**, run with `--project 'unit*'`.
- **Mutation sweep over all four checks** — disabling each killed 5 / 2 / 3 / 3 tests respectively, 37 green when restored. One mutant first died as a *syntax error* (`Tests no tests`, the import-failure signature) and was redone validly, since a mutant that never runs proves nothing.
- IPv4 rule measured against the real exports before it was written: the only IPv4-shaped tokens present are the four RFC5737 replacements, so it needs no ignore list and a hit is a real finding.

### Not verified

- **Check 1 is proven by fixture, not against real values** — the denylist is empty until someone populates it with the private key. Checks 2–4 are proven against the actual tree.
- The workflow itself has not run yet; this PR is its first execution.
- `raw/README.md` is left prettier-nonconforming, matching its state at base — it is a *modified* file, and this repo enforces prettier on added files only.

### DIFF
--- a/.github/workflows/retool-sanitiser-guard.yml
+++ b/.github/workflows/retool-sanitiser-guard.yml
@@ -13,14 +13,24 @@ name: Retool Sanitiser Guard
 # exactly why both got through the original pass. Read the gate's header before
 # treating this check as coverage.
 
+# 🔴 BOTH triggers, deliberately. `lint.yml` documents at length that this repo had
+# no CI verdict on `main` for two months because its jobs were pull_request-only,
+# and names a real commit pushed straight to `main`. Every base-relative job there
+# is exempt from the push path because a push has no base to diff against — but
+# THIS gate scans the directory absolutely, so a push trigger costs nothing and
+# closes the hole. `release` is pushed to directly by the deploy flow, so it needs
+# the same cover.
 on:
   pull_request:
     branches: [main, release]
-    paths:
+    paths: &guarded
       - 'docs/moderator-app/retool-exports/**'
       - 'scripts/ci/retool-sanitiser-gate.mjs'
       - 'scripts/ci/retool-sanitiser-denylist.json'
       - '.github/workflows/retool-sanitiser-guard.yml'
+  push:
+    branches: [main, release]
+    paths: *guarded
 
 permissions:
   contents: read
@@ -40,13 +50,22 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version-file: .nvmrc
 
       - name: Run the sanitiser gate
         env:
           # Not committed: a person's name is a tiny search space, so a public salt
-          # would make the committed hashes dictionary-crackable. Absent (e.g. on a
-          # fork PR) the denylist check reports SKIPPED and the IPv4, credential and
-          # placeholder checks still run and still block.
+          # would make the committed hashes dictionary-crackable.
+          #
+          # 🔴 When this is absent the denylist check does NOT run. The IPv4,
+          # credential and placeholder checks still do, and the gate's PASS line
+          # says so explicitly rather than claiming a scope it did not cover.
+          # Absent cases: the secret does not exist yet, and any FORK PR — where
+          # this workflow does not run AT ALL, so it is no cover there whatsoever.
+          #
+          # 🔴 And a red check here does not BLOCK: this repo has no required
+          # status checks (`branches/main/protection` → required_status_checks:
+          # null), so a failing run renders red and merges anyway. It is a signal
+          # for a human, not an enforcement boundary. Do not describe it as one.
           RETOOL_SANITISER_SALT: ${{ secrets.RETOOL_SANITISER_SALT }}
         run: node scripts/ci/retool-sanitiser-gate.mjs
--- a/.github/workflows/retool-sanitiser-guard.yml
+++ b/.github/workflows/retool-sanitiser-guard.yml
@@ -1,0 +1,52 @@
+name: Retool Sanitiser Guard
+
+# Blocks a PR that puts personal data back into `docs/moderator-app/retool-exports/**`.
+#
+# WHY: PR #4476 removed three classes of personal data from these exports — staff
+# real names, banned-user IP addresses and end-user account ids. The sanitisation
+# rules that were supposed to prevent that live in `raw/README.md` as PROSE, and
+# prose is not a gate. A fresh Retool re-download would have put all three back with
+# nothing to stop it.
+#
+# 🔴 A green run here does NOT mean "no personal data". The gate cannot see a NOVEL
+# staff name or a NOVEL bare account id — neither has a shape to match on, which is
+# exactly why both got through the original pass. Read the gate's header before
+# treating this check as coverage.
+
+on:
+  pull_request:
+    branches: [main, release]
+    paths:
+      - 'docs/moderator-app/retool-exports/**'
+      - 'scripts/ci/retool-sanitiser-gate.mjs'
+      - 'scripts/ci/retool-sanitiser-denylist.json'
+      - '.github/workflows/retool-sanitiser-guard.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: retool-sanitiser-guard-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  guard:
+    name: retool export sanitiser
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Run the sanitiser gate
+        env:
+          # Not committed: a person's name is a tiny search space, so a public salt
+          # would make the committed hashes dictionary-crackable. Absent (e.g. on a
+          # fork PR) the denylist check reports SKIPPED and the IPv4, credential and
+          # placeholder checks still run and still block.
+          RETOOL_SANITISER_SALT: ${{ secrets.RETOOL_SANITISER_SALT }}
+        run: node scripts/ci/retool-sanitiser-gate.mjs
--- a/.github/workflows/retool-sanitiser-guard.yml
+++ b/.github/workflows/retool-sanitiser-guard.yml
@@ -1,6 +1,11 @@
 name: Retool Sanitiser Guard
 
-# Blocks a PR that puts personal data back into `docs/moderator-app/retool-exports/**`.
+# Reports a PR or push that puts personal data back into
+# `docs/moderator-app/retool-exports/**`.
+#
+# 🔴 It REPORTS, it does not BLOCK — this repo has no required status checks, so a
+# red run merges anyway. See the note on the gate step below; do not re-describe
+# this as an enforcement boundary.
 #
 # WHY: PR #4476 removed three classes of personal data from these exports — staff
 # real names, banned-user IP addresses and end-user account ids. The sanitisation
@@ -35,8 +40,14 @@ on:
 permissions:
   contents: read
 
+# 🔴 The group MUST vary per commit now that this triggers on push-to-main.
+# `github.ref` alone is ONE group for the whole branch, so two commits landing
+# close together cancel each other and only the last is ever scanned — which
+# half-reopens the very hole the push trigger was added to close, and the loser
+# renders as `cancelled`, which this repo cannot distinguish from a failure.
+# Pinned by scripts/__tests__/main-branch-ci-coverage.test.ts, which caught this.
 concurrency:
-  group: retool-sanitiser-guard-${{ github.ref }}
+  group: retool-sanitiser-guard-${{ github.ref }}-${{ github.sha }}
   cancel-in-progress: true
 
 jobs:
--- a/docs/moderator-app/retool-exports/raw/README.md
+++ b/docs/moderator-app/retool-exports/raw/README.md
@@ -85,7 +85,9 @@ mechanical half on every PR **and every direct push** touching this directory:
 
 - known-bad values that must never come back, matched against **salted hashes** — the
   plaintext is deliberately not in this repo, and neither is the salt;
-- every IP literal must be an RFC5737/RFC1918 address (v6 is rejected outright);
+- every IPv4 literal must be an RFC5737/RFC1918 address, and any IPv6 address is
+  rejected except loopback/unspecified (`::`, `::1`) — a Postgres cast such as
+  `id::date` is not an address and does not count;
 - the credential shapes above, as code rather than prose;
 - the redaction **placeholders must still be present, at their expected counts** —
   which is what catches the headline risk, a wholesale re-download silently
--- a/docs/moderator-app/retool-exports/raw/README.md
+++ b/docs/moderator-app/retool-exports/raw/README.md
@@ -80,15 +80,17 @@ like nothing in particular. It was caught before the commit was ever pushed. Any
 ## Part of this is now a gate, and part of it can never be
 
 Everything above is a checklist, and a checklist is only as good as the person running
-it. `scripts/ci/retool-sanitiser-gate.mjs` (CI: **Retool Sanitiser Guard**) now enforces
-the mechanical half on every PR touching this directory:
+it. `scripts/ci/retool-sanitiser-gate.mjs` (CI: **Retool Sanitiser Guard**) now checks the
+mechanical half on every PR **and every direct push** touching this directory:
 
 - known-bad values that must never come back, matched against **salted hashes** — the
   plaintext is deliberately not in this repo, and neither is the salt;
-- every IPv4 literal must be an RFC5737/RFC1918 address, so a real one blocks;
+- every IP literal must be an RFC5737/RFC1918 address (v6 is rejected outright);
 - the credential shapes above, as code rather than prose;
-- the redaction **placeholders must still be present** — which is what catches the
-  headline risk, a wholesale re-download silently overwriting a sanitised file.
+- the redaction **placeholders must still be present, at their expected counts** —
+  which is what catches the headline risk, a wholesale re-download silently
+  overwriting a sanitised file. Counts, not mere presence: one surviving token would
+  otherwise make a partial re-introduction read as clean.
 
 🔴 **A green run is NOT "no personal data".** The gate cannot see a *novel* staff real
 name or a *novel* bare account id: neither has a shape to match on, which is exactly
@@ -96,9 +98,25 @@ why both walked through the pass described above. That half is still yours. When
 sanitise a fresh export, work the three-class checklist by hand and treat the gate as a
 backstop against regressions, never as the check itself.
 
+🔴 **Three more things it does not do, stated because the first draft of this section
+claimed otherwise:**
+
+1. **It does not BLOCK.** This repo has no required status checks, so a red run renders
+   red and the PR merges anyway. It is a signal for a human, not an enforcement
+   boundary.
+2. **It does not run on fork PRs at all** — not "runs with the denylist skipped".
+   Workflows do not execute for outside contributors here, so a fork PR touching these
+   files gets no coverage whatsoever.
+3. **Without the `RETOOL_SANITISER_SALT` secret the denylist does not run.** The gate's
+   PASS line names the reduced scope when that happens, rather than asserting a scope
+   it did not cover.
+
 Add a newly-found bad value to the denylist with the salt exported locally:
 `node scripts/ci/retool-sanitiser-gate.mjs --hash '<value>'`, then commit the digest
-only.
+only. Populating `hashes` for the first time? Set `sentinel` in the same pass
+(`--sentinel`) — it is the only thing that detects a salt/hash desync, which otherwise
+leaves every hash matching nothing, silently. Note `--hash` refuses a value longer than
+four word-tokens, because the matcher could never find it.
 
 ## Read the whole file, not just the SQL
 
--- a/docs/moderator-app/retool-exports/raw/README.md
+++ b/docs/moderator-app/retool-exports/raw/README.md
@@ -77,6 +77,29 @@ live key because it looked only for vendor prefixes and `Bearer` headers: User L
 like nothing in particular. It was caught before the commit was ever pushed. Anything matching
 `(apikey|token|secret|password|auth)\w*\s*=\s*['"]…` is now redacted regardless of what follows.
 
+## Part of this is now a gate, and part of it can never be
+
+Everything above is a checklist, and a checklist is only as good as the person running
+it. `scripts/ci/retool-sanitiser-gate.mjs` (CI: **Retool Sanitiser Guard**) now enforces
+the mechanical half on every PR touching this directory:
+
+- known-bad values that must never come back, matched against **salted hashes** — the
+  plaintext is deliberately not in this repo, and neither is the salt;
+- every IPv4 literal must be an RFC5737/RFC1918 address, so a real one blocks;
+- the credential shapes above, as code rather than prose;
+- the redaction **placeholders must still be present** — which is what catches the
+  headline risk, a wholesale re-download silently overwriting a sanitised file.
+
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Correcting this PR's own description, and two blocking fixes — `e081509105`

An adversarial audit of `907d5ff7df` found two 🔴 and five 🟡. Both 🔴 were cases where **this gate's claims were wider than its implementation** — which is the exact failure the gate exists to prevent, one level up. Correcting publicly rather than silently editing the body, since the wrong version has been readable for a while.

### What the description got wrong

1. 🔴 **"the redaction placeholders must still be present — the one that catches the headline risk."** The ledger covered **3 of the 10 token instances** #4476 actually wrote. `__MODERATOR_B__`, `__MODERATOR_C__` and `<accountId>` were guarded by **nothing** — and `<accountId>` is the one documented class with no other check behind it, because a real IP is caught by the IPv4 rule while a bare account id is an ordinary integer. The audit executed it: two staff names and five account ids restored, **gate exited 0 and printed PASS**.

2. 🔴 **"a mismatched salt"** was not considered at all. An *active* denylist whose salt doesn't match its hashes matches nothing, prints nothing, and exits 0 — under a PASS line asserting "no known-bad value". A trailing newline in the CI secret (GitHub preserves it) was enough.

3. **"Absent (e.g. on a fork PR) the denylist check reports SKIPPED and the IPv4, credential and placeholder checks still run and still block."** Two errors in one sentence:
   - **Nothing blocks.** `branches/main/protection` → `required_status_checks: null`. A red run renders red and merges anyway.
   - **On a fork PR nothing runs at all** — workflows don't execute for outside contributors here, so it is not partial cover, it is no cover. (Verified: a fork PR shows 0 check-runs.)

4. **"The workflow has never run."** Stale by the time it was written — run `33234406609`, `conclusion=success`.

5. **The mutation table (5/2/3/3)** did not reproduce under the auditor's mutant shape, which gave 3/1/6/3. Neither is wrong — the PR never specified the shape — but the number should not be read as a property of the suite. What holds is the claim that matters: every check has at least one killing test.

### Fixed in `e081509105`

Both 🔴 closed, with the audit's own executed scenarios re-run against the fix (both now exit 1), and a **positive control** on the salt fix — the correct salt still *finds* the planted value, so the sentinel didn't simply disable the check.

Also: `push:` trigger added (`lint.yml` documents this repo having no CI verdict on `main` for two months for exactly this reason); the gate **no longer prints the values it exists to redact** into a world-readable Actions log — it withheld the denylist value for that reason and then printed IPs and Bearer tokens one branch up; `SKIPPED`/failures are now GitHub `::warning`/`::error` annotations instead of a `console.log` on a green check; `secret-assignment` widened to the shapes the corpus actually uses (quoted JSON keys, the transit `"KEY","value"` pair, escaped JS in JSON, template literals, and `WEBHOOK_TOKEN`-style names, which `\b` structurally could not match); `--hash` refuses values the matcher could never find; IPv6 rejected; `.nvmrc` instead of a hardcoded Node 22.

### One thing worth stating plainly

The widened credential regex **passed all 62 synthetic tests and still false-positived on the real exports** — it matched an *empty* `databasePasswordOverride` by running across `\"` into the next key, and an earlier revision reported a 646,934-char "secret" in three files. The real corpus is the false-positive control, and it caught what the fixtures could not. Both are now pinned as tests.

64 tests (was 37). Mutation sweep over the five new guards: each killed 1–2 tests, 64 green restored, gate file byte-identical after.

**Still true and unchanged:** the gate cannot see a *novel* staff name or a *novel* bare account id, the `RETOOL_SANITISER_SALT` secret still does not exist, and the denylist still ships empty — so check 1 is inert until someone populates it.

ZacxDev:
Round-1 fix ledger for the delta re-audit. Each line states **what was claimed fixed**, not why it is correct.

```audit-claims round=1 audited=907d5ff7df..e0815091
1. REQUIRED_PLACEHOLDERS widened from 3 entries to 7, covering __MODERATOR_A/B/C__, <accountId> and 203.0.113. across user-lookup-v2.json, bulk-ban.json and bulk-ban.md
2. check 4 now asserts an expected COUNT per token instead of mere presence, claimed to catch a partial re-introduction where one of several occurrences is reverted
3. a sentinel digest was added to the denylist file; a salt whose sentinel does not match is claimed to FAIL rather than silently match nothing
4. the salt is now .trim()ed, claimed to absorb a trailing newline in the CI secret rather than desyncing on it
5. the PASS line is claimed to name its reduced scope when the denylist did not run, instead of asserting "no known-bad value"
6. a push: trigger on main and release was added via a YAML anchor, claimed to close the direct-push bypass
7. matched values are claimed to be masked to "<n> chars, sha256:<8 hex>" everywhere, so no finding republishes an IP or credential into a public Actions log
8. NOTE and FAIL output are claimed to emit GitHub ::warning / ::error annotations when GITHUB_ACTIONS=true
9. secret-assignment was rewritten and is claimed to catch quoted JSON keys, the transit "KEY","value" pair, escaped JS inside JSON, and WEBHOOK_TOKEN-style names, while a new secret-assignment-template rule covers backtick values
10. secret-assignment is claimed to be length-bounded so it cannot run away across a file, after an earlier revision reported a 646,934-char match
11. the value body is claimed to reject an escaped quote so an EMPTY secret-named value cannot run across into the next JSON key
12. bearer-token gained the i flag, claimed to match lowercase/uppercase per RFC 7235
13. hashValue is claimed to throw on a value with more than MAX_NGRAM_WORDS tokens or with no word tokens
14. a --sentinel CLI mode was added to generate the sentinel digest
15. findIPv6 was added, claimed to reject any IPv6 literal except :: and ::1
16. V4_RE gained negative look-around, claimed to stop flagging a dotted quad inside a longer version string
17. node-version-file: .nvmrc replaces a hardcoded node-version: 22
18. the workflow comments and raw/README.md were corrected to state that the gate does NOT block, does NOT run on fork PRs at all, and that the denylist does not run without the secret
19. the test suite grew from 37 to 64 tests, including a test claimed to pin REQUIRED_PLACEHOLDERS against the REAL exports rather than a fixture
```

ZacxDev:
## Round 2 fixed — `ffa81e0e7f`

Round 2 confirmed both round-1 blockers closed and found that the fix round had introduced its own. Every regex it touched had been widened on one axis and narrowed on another.

**Ledger:** `round 2 · payload lines changed THIS round: 146 (since round 1: 913) · elapsed: ~40 min`
(`git log --numstat --format= --remerge-diff e081509105..ffa81e0e7f --not origin/main`, rc 0, git 2.55.0, non-empty range: gate.mjs 105/22, workflow 13/2, README 3/1 = **146 payload**; test.ts 140/4 = 144 scaffolding.)

Notable: the round-2 fix *itself* introduced one finding — allowing empty IPv6 groups made `id::date` match as `d::da`, and the real corpus reported four. Caught by the corpus control, invisible to 100 synthetic tests.

```audit-claims round=2 audited=e081509105..ffa81e0e
1. concurrency group now varies per commit (github.sha added), claimed to stop two commits on main cancelling each other; main-branch-ci-coverage.test.ts, which was RED, is claimed to pass
2. IPv4 findings now go through mask(), claimed to leave no branch printing a raw matched value
3. V4 lookahead changed to (?!\.?\d), claimed to keep the version-string fix while no longer missing an IP followed by a sentence period
4. V6_RE rewritten to allow empty groups, claimed to catch the whole X::Y family including fe80::1 and fe80::a00:27ff:fe4e:66a1
5. V6 match boundaries plus a >=2-hex-group filter added, claimed to stop Postgres casts like id::date matching as addresses
6. ALLOWED_V6 is claimed to be reachable now, so the loopback exemption and its test are no longer vacuous
7. the secret-assignment length bound was raised from 200 to 4096, claimed to stop long opaque tokens falling off the top while the character class still prevents a runaway
8. the keyword is anchored with (?![a-z]) and the i flag removed with casings enumerated, claimed to drop author/authorName/authorEmail/authorized/tokenizer while keeping stripeSecretKey, WEBHOOK_TOKEN and apiKey
9. the `,` separator is REVERTED, claimed as a deliberate measured decision rather than an oversight: it caught nothing real in the corpus and flagged tag/column arrays
10. secret-assignment-template got the same widening as its sibling, claimed to catch WEBHOOK_TOKEN = `...`
11. annotate() now derives file= from the finding, claimed to stop pinning every annotation to line 1 of the denylist
12. annotate() and passScope() are exported, claimed so the round-1 guards that a mutant survived are now reachable by tests
13. minValueLength was REMOVED as unreachable, claimed to be dead code the quantifier already enforces; the floor is claimed to be pinned at its boundary by test instead
14. the workflow headline comment and README were corrected, claimed to no longer say the gate blocks, and the README v6 claim is claimed to be narrowed to match the code
15. the suite grew 64 -> 101 tests, and every mutant in the sweep is claimed to be node --check'ed and diff-confirmed before scoring
```

ZacxDev:
## Round 3 fixed — `06485561d6`

Round 3 found 3 🟡 + 3 🟢. Two of the 🟡 were this ladder's recurring shape again, and one was the **same defect in a second place**: the boundary added to reject Postgres casts also rejected a sentence period, so `Banned the account from fe80::1.` went undetected — the exact mirror of the IPv4 defect the *immediately preceding commit* fixed, with the same "half this corpus is prose" justification in its own message.

Also caught: an earlier scripted edit of mine had **deleted two test blocks** (it replaced from a marker to end-of-file). The audit found it by noticing no test named the SQL-cast defect the commit claimed to have pinned.

**Ledger:** `round 3 · payload lines changed THIS round: 97 (since round 1: 1010) · elapsed: ~35 min`
(`git log --numstat --format= --remerge-diff ffa81e0e7f..06485561d6 --not origin/main`, rc 0, silent stderr, non-empty range, git 2.55.0: gate.mjs 53/44 = **97 payload**; test.ts 79/0 = 79 scaffolding.)

```audit-claims round=3 audited=ffa81e0e7f..06485561
1. the V6 trailing boundary now permits a sentence period, claimed to find fe80::1 / 2a01:4f8:c17:1234::1 when followed by "." or "..." while still rejecting Postgres casts
2. the V6 leading boundary no longer excludes ":", claimed to match a bare ip:2001:db8::1 without re-admitting casts
3. the groups.length < 2 filter is REMOVED, claimed to be redundant with the boundaries for its stated job of rejecting casts
4. single-group IPv6 addresses (fe80::, ::dead) are claimed to be reported now, matching what the README says
5. ALLOWED_V6 is claimed to be genuinely reachable and its loopback test no longer vacuous — emptying the Set is claimed to fail tests where deleting it previously left the suite green
6. authorization / Authorization / AUTHORIZATION / authentication were added to the keyword alternation, claimed to restore the dominant credential identifier in the corpus (12 occurrences in raw/user-lookup-v2.json)
7. X-API-Key, X-API-KEY, Api_key, Apikey, secretkey and the plain camelCase apiKey were added, claimed to close the casings dropped when the i flag was removed
8. an optional plural `s` was added before the anchor, claimed to catch tokens/secrets/passwords while still rejecting tokenizer
9. the alternation is claimed to be ordered longest-first so `authorization` is not consumed by `auth`
10. two test blocks deleted by an earlier scripted edit (SQL-cast and comma-revert) are claimed to be restored
11. the no-minimum-length docstring is claimed corrected per-rule: unreachable for secret-assignment, REACHABLE for secret-assignment-template
12. the duplicate `name: 'secret-assignment'` key is claimed removed
13. comments contradicting their code ({1,7} vs {2,7}, (?![a-z0-9]) vs (?![a-z])) and three rounds of superseded V6 commentary are claimed consolidated
14. 101 -> 140 tests across both suites; every mutant in the sweep is claimed node --check'ed and diff-confirmed before scoring
```

ZacxDev:
## Round 4 fixed — `12dc90846b`

The finding that mattered: **the gate did not catch the corpus's own credential serialisation.** All 12 `Authorization` entries in `raw/user-lookup-v2.json` are Retool transit pairs — `\"key\":\"Authorization\",\"value\":\"Basic <base64>\"` — where the identifier is a *value* and the separator is `,`. The `Bearer` variant was caught incidentally by the pre-existing rule; **`Basic <base64>` was caught by nothing**, while a test comment I added in round 3 asserted the class was covered.

**Ledger:** `round 4 · payload lines changed THIS round: 148 (since round 1: 1158) · elapsed: ~30 min`
(`git log --numstat --format= --remerge-diff 06485561d6..12dc90846b --not origin/main`, rc 0, silent stderr, non-empty range, git 2.55.0: gate.mjs 105/43 = **148 payload**; test.ts 70/4 = 74 scaffolding.)

```audit-claims round=4 audited=06485561d6..12dc9084
1. a new `retool-kv-credential` rule catches the Retool transit pair \"key\":\"<credential>\",\"value\":\"<secret>\", claimed to cover the corpus's own Authorization serialisation including a live Basic credential
2. that rule is claimed safe where a bare `,` separator was not, because it requires the literal key/value wrapper and so cannot match a tag or column array
3. REDACTED_RE now tolerates an optional leading auth scheme, claimed necessary because the corpus writes these as `Basic {{ SomeCredentials.value }}` and all 12 would otherwise report as findings
4. `(?!\.[0-9a-f])` was REMOVED from V6_RE, claimed to contribute nothing to the sentence-period fix while suppressing every IPv4-embedded form including ::ffff:192.168.1.1 which previously matched nothing at all
5. the sentence-period behaviour is claimed unchanged by that removal — fe80::1. and 2001:db8::1... still found
6. the three rule regexes are claimed to be built from ONE shared SECRET_WORD source string rather than pasted copies
7. the Authorization test comment is claimed corrected: it pins the JSON-key shape only, and no longer claims to cover the corpus
8. the "boundaries alone reject casts" comment is claimed narrowed to what holds — true only because common type names contain a non-hex character; `)::dec` is shape-identical to `::dead` and fails closed
9. two of the twelve `catches %s` cases are claimed labelled as INVARIANT GUARDS, since the pre-existing AUTH branch already caught them before the widening
10. 140 -> 153 tests across both suites; four mutants claimed killed (4/5/1/1), each node --check'ed and diff-confirmed
```

ZacxDev:
## Round 5 fixed — `b64d8ebf28`

Round 5 found a 🔴 that invalidated round 4's headline fix: **`retool-kv-credential` was inert on the corpus it was written for.** `Q` accepted at most one backslash; the corpus uses three (Retool nests JSON in JSON in a string) — measured, 181 of 191 `key` occurrences at depth 3, 10 at depth 1, none at 0. The rule reported 0 hits corpus-wide, and injecting a live credential into the real export produced 0 findings.

**The spec passed anyway, because its fixture hardcoded one backslash.** The fake and the implementation encoded the same wrong assumption about the data, so neither could see the other's error — 8 tests green over a rule that never fired.

**Ledger:** `round 5 · payload lines changed THIS round: 16 (since round 1: 1174) · elapsed: ~25 min`
(`git log --numstat --format= --remerge-diff 12dc90846b..b64d8ebf28 --not origin/main`, rc 0, silent stderr, non-empty range, git 2.55.0: gate.mjs 13/3 = **16 payload**; test.ts 72/25 = 97 scaffolding. Small but decisive — both fixes are one-token.)

```audit-claims round=5 audited=12dc90846b..b64d8ebf
1. Q now accepts 0-3 backslashes, claimed to make retool-kv-credential actually match the corpus's triple-escaped transit pairs
2. a positive control against the REAL export is claimed to show 0 kv findings unmodified and >=1 with a live Basic credential substituted in, still masked
3. the kv key gains the (?![a-z]) end-of-identifier anchor, claimed to close the author/authorName/authorEmail/authorized/tokenizer false-positive class the rule had re-opened
4. the kv tests are claimed PARAMETERISED over escaping depths 0/1/3 so a fixture can no longer encode a single wrong depth
5. two tests are claimed to drive the REAL export rather than a fixture — one asserting it is clean, one substituting a live credential, with a control that the substitution happened
6. the round-4 claim about REDACTED_RE's scheme prefix, previously unfalsifiable because the rule was inert, is claimed now measured: removing the prefix yields rc=1 and exactly 12 findings
7. key-then-value adjacency and ordering remain REQUIRED — recorded as a known robustness gap, deliberately not fixed
8. 153 -> 182 tests; three mutants claimed killed (6/15/5), each node --check'ed and diff-confirmed
```

ZacxDev:
## Audit ladder closed at round 6 — `5a019779cc`

**Round 6 came back clean of regressions and explicitly said not to run another round.** Per the stop rule a clean round ends the ladder, and the attribution ledger agrees independently: of round 5's 16 payload lines only **2 were functional**, so a round 7 would have been auditing scaffolding rather than this PR.

Six rounds, five fix commits. What they found, in order:

| round | the finding that mattered |
|---|---|
| 1 | the placeholder ledger covered **3 of 10** tokens; restoring two staff names and five account ids passed the gate at rc=0 |
| 1 | an active-but-mismatched salt matched nothing, printed nothing, exited 0 — under a PASS asserting "no known-bad value" |
| 2 | the push trigger I added cancelled its own runs; the repo's **own** invariant caught it and the PR went red |
| 2 | IPv4 findings printed the raw address into a world-readable Actions log — the one class the gate exists to keep out |
| 3 | the same boundary defect **repeated in a second regex**; dropping the `i` flag silently lost `Authorization` |
| 4 | the gate did not catch the corpus's **own** credential serialisation |
| 5 | 🔴 the rule written for that was **inert** — 0 matches on the file it names |
| 6 | clean |

### The one worth remembering

Round 5. `Q` accepted one backslash; the corpus uses **three** (Retool nests JSON in JSON in a string — 181 of 191 `key` occurrences at depth 3). The rule matched nothing at all on the real export. **And the spec passed**, because its fixture hardcoded one backslash: the fake and the implementation encoded the same wrong assumption, so neither could see the other's error. Eight tests green over a rule that never fired.

More tests would not have caught that. The fix that matters is structural — the kv tests are now parameterised over escaping depths 0/1/2/3, and two of them **drive the real export** rather than a fixture: one asserts it is clean, one substitutes a live credential and asserts it is caught, with a control that the substitution happened. A fixture can encode my mistake; the corpus cannot. That control is what caught round 6's finding too.

### Recurring shape, named

Almost every round found the previous fix **widened on one axis and narrowed on another** — a version-string false positive traded for missing `8.8.8.8.` at the end of a sentence; a cast filter traded for exempting every single-group IPv6; a `[\w$]*?` widener traded for flagging `author`. Three separate guards were also found to be **unreachable while reading as coverage** (`minValueLength`, `ALLOWED_V6`, `groups.length < 2`) — each deleted or repaired only because a mutant that should have killed tests left the suite fully green.

### State

- **187 tests** (37 at open), corpus `rc=0`.
- Every guard mutation-checked, each mutant `node --check`ed and diff-confirmed before scoring — two first-attempt mutants were duds (one a syntax error, one that only changed lazy to greedy) and were redone, since a mutant that does not do what you think proves nothing.
- Cumulative churn: ~1,190 payload lines, ~1,000 scaffolding.

### Unchanged and still true

- The gate **cannot** see a *novel* staff name or a *novel* bare account id. Neither has a shape; that is why both walked through the original pass. A green run is not "no personal data".
- It **reports, it does not block** — this repo has no required status checks.
- It **does not run on fork PRs at all**.
- `RETOOL_SANITISER_SALT` does not exist yet and the denylist ships empty, so check 1 is inert until someone populates it. Checks 2–4 work today.

### Known, not fixed — recorded rather than half-addressed

The kv rule requires key-then-value **adjacency and order**; all 361 corpus pairs are written that way. And a hex-only Postgres alias like `)::dec` is shape-identical to the legitimate address `::dead`, so it fails closed.

ZacxDev:
## Closing unmerged — the premise does not hold

Zach asked the question that none of the six audit rounds did: **the Retool → moderator-service migration is a one-time port, so is there a "next re-download" at all?**

There is not. Measured:

| check | result |
|---|---|
| times `raw/*.json` has been re-downloaded, ever | **0** — committed once on 2026-08-10 (`ebc50e899a`); only #4476's redaction has touched them since |
| where a refresh would come from | **ClickUp attachments** — `retool-exports/README.md` says "that is where to get them, not a local folder" |
| Retool's future | `raw/README.md`: kept "because **Retool access is being lost**"; `retool-parity-checklist.md` is "what must exist **before Retool is switched off**" |

This PR's stated risk is *"nothing mechanical stops the next Retool re-download re-introducing all three classes."* That event has never occurred once in the corpus's history and is becoming impossible. These files are the frozen **spec** for a completed port, not a living import.

## What went wrong here, recorded because it is the useful part

I took the premise from the ticket text and never checked whether a re-download could happen. Then I ran **six adversarial audit rounds** — each asking *"is this gate correct?"*, never *"should this gate exist?"* Every auditor faithfully verified the frame it was handed. That is the framed-audit failure one level above the code: the mechanism ended up very well tested, and the mechanism should not be here.

The rounds themselves were not wasted as *findings* — they caught an inert rule, a silently-passing salt desync, a placeholder ledger covering 3 of 10 tokens, and a CI trigger that cancelled its own runs. They were wasted as *work*, which is a different axis and the one that matters here.

The size of the change is not an argument for merging it: ~2,200 lines guarding a directory nobody will edit again, where check 1 is inert (the `RETOOL_SANITISER_SALT` secret was never created), the gate does not block (no required status checks), and it does not run on fork PRs at all.

## What is NOT retracted

**#4476 stands.** The staff-name and IP redaction was real and remains correct — this PR was only the going-forward gate for it, and `raw/README.md`'s sanitisation checklist is still the control for anyone who ever does refresh an export by hand.

The genuinely live surface is `apps/moderator`, which now holds the moderation data these exports only described. If a sanitiser gate is ever worth building it belongs there — a different premise, which should be measured before it is proposed rather than after.

The branch is deleted; `refs/pull/4486/head` preserves every commit if this is ever wanted (a fact this PR's own history established the hard way).

PR #4488, opened 2026-08-29
fix(model-scan): a skipped scan is not a passed scan

ZacxDev:
Web-app half of a model-upload scanner fix. Scanner side is in the private `civitai-spine-controller` repo; incident record, mechanism and measurements are in internal ticket **868ky8j7f**.

> This repository is public and the mitigation is not deployed, so this description deliberately omits the mechanism.

## 🔴 Read the deploy-ordering section before merging

This change is inert on its own: with the flag off — the shipping state — the result mapping is unchanged from `main`. It takes effect only once the scanner side is deployed.

## What this changes

The pickle-scan result mapping treated one class of non-answer from the scanner as an affirmative clean result. It no longer does — the model cannot show a Verified shield on that basis.

- **`derivePickleScanResult`** — a skip counts as a pass only when the scanner verified the container from its magic bytes. Behind `SCAN_STRICT_SKIP_VERIFICATION`, **default-off**.
- **`deriveClamScanResult`** — an explicit `limitsExceeded` is `Error`, not `Pending` (`Pending` would re-queue the file in `scan-files.ts` forever, and the retry cannot succeed).
- The scanner's verdict text is kept off the public v1 API surface; the detail goes to `rawScanResult` and the operator log.
- A format mismatch is recorded as **ungated telemetry** — no flag, no model mutation.

## Scope: the automated hold was removed, deliberately

Earlier revisions of this PR tried to put the model on hold automatically. Four adversarial audit rounds each found the hold broken in a new way, always the same shape — a lever reached for without owning its whole lifecycle:

| round | what the hold did wrong |
|---|---|
| 1 | fired on `Draft` models, locking creators out of never-published drafts |
| 2 | the allow-list fixing that was the mirror bug — the *first* upload is scanned while a Draft, so the dominant path was uncovered, with no hold, no queue entry and no log |
| 3 | the flag it wrote was **inert twice over**: the moderator queue excludes `Draft`, and the publish handler strips `needsReview` |
| 4 | the replacement flag is **owner-writable** — absent from `MODERATION_OWNED_META_KEYS`, so the model's own owner clears it with one `model.upsert`, API key included |

Round 4 also found that a flagged `Draft` reaches **no** moderator queue, and the only surface that toggles that flag filters on `uploadType='Trained'` while wizard uploads are `Created`. Those compound badly: the adversary can clear it, the moderator cannot find it, and an honest creator hitting a false positive is stuck.

🔴 **And fixing only the owner-writable half would have made it worse** — locking the flag down without a clearing surface turns a self-clearable false positive into a permanent lockout with no appeal. False positives are real: several legitimate container shapes are unverifiable by the scanner and detect as unknown.

An automated hold needs product surfaces that do not exist yet: a queue that lists flagged drafts, a clearing path, correct owner-facing copy (today's only message for that flag blames "training data", which is wrong for a format mismatch), and a notification. That is a feature, not a patch to the scan service — so it is a **follow-up**, tracked on the internal ticket.

What is left here is the part that is correct, verified and actually closes the gap. The ungated telemetry is deliberate: it is the population signal needed to size the problem and validate the detector *before* any enforcement is designed.

A test pins this scope directly — a mismatch must log and must touch no model — and is mutation-checked both ways, so a later change cannot quietly put enforcement back here.

## 🔴 Deploy ordering

`SCAN_STRICT_SKIP_VERIFICATION` **must stay off until the scanner change is deployed fleet-wide.** Enabling it before the worker fleet is updated would mark ordinary uploads `Error`. A test pins both regimes so this cannot regress silently.

⚠️ The scanner work is **two** PRs, both owned by @Koen: `civitai-spine-controller#231` is the format-conformance half that this PR's mismatch handling pairs with; **`#232`** is the ClamAV size-limit half, which is what the `limitsExceeded` branch pairs with. That branch is dead against both `main` and #231 — it needs #232.

Order: **#231 → this PR → flip `scan-strict-skip-verification`.** The flag does not exist in `flipt-state` yet; an unknown flag resolves `false`, so the default is safe.

⚠️ **The flag is not a kill switch for its own effects.** Verdicts are persisted, and turning it back off does not repair rows: `scan-files.ts` re-queues only on `virusScanResult: Pending`, and pickle results are never re-driven. A premature flip leaves a population of `pickleScanResult: Error` needing a manual rescan.

## Also unchanged

Neither a regression — both identical on `main`: two scan-related fields remain on the public v1 API surface. This PR narrows what they expose; it does not remove them. Detail in the internal ticket.

## Tests

80 in the target file. Every guard has a killing mutation, each re-verified rather than trusted:

| mutation | test killed |
|---|---|
| error statuses → `Success` | mismatch-is-not-clean |
| delete the `limitsExceeded` guard | limits-with-unrecognized-status |
| strict `skipped`-status branch → `Success` | skipped-status-is-not-a-pass |
| drop `gguf-magic` from the verified set | byte-verified GGUF skip |
| reintroduce a model mutation on mismatch | the scope guard |
| drop the telemetry log | the scope guard + flag-off logging |

Suite: **412 files / 5,742 tests**; typecheck **0 errors**; prettier clean (positive-controlled).

### DIFF
--- a/src/server/flipt/client.ts
+++ b/src/server/flipt/client.ts
@@ -126,13 +126,6 @@ export enum FLIPT_FEATURE_FLAGS {
   // context, so a segment rule here returns the flag default and looks exactly like "blurbs are
   // off". The site is recorded in ENTITY_WITHOUT_CONTEXT_LEDGER (flipt-eval-context.test.ts).
   TEXT_BLURBS = 'text-blurbs',
-  // Arms the "hold this model for moderator review" action when the scanner reports that a
-  // file's declared container format is contradicted by its bytes (skipReason
-  // "format-mismatch"). Default-off: the mismatch is RECORDED as a non-clean scan result
-  // regardless — that part needs no flag and is what actually closes the bypass — and this
-  // gates only the unpublish-and-notify side effect, so it can be flipped once the moderator
-  // queue has been watched for a day.
-  SCAN_FORMAT_MISMATCH_HOLD = 'scan-format-mismatch-hold',
   // 🔴 DO NOT ENABLE BEFORE THE SPINE-CONTROLLER SCANNER CHANGE IS DEPLOYED EVERYWHERE.
   //
   // Makes a pickle-scan SKIP count as a pass only when the scanner verified the container from
--- a/src/server/flipt/client.ts
+++ b/src/server/flipt/client.ts
@@ -136,9 +136,8 @@ export enum FLIPT_FEATURE_FLAGS {
   // 🔴 DO NOT ENABLE BEFORE THE SPINE-CONTROLLER SCANNER CHANGE IS DEPLOYED EVERYWHERE.
   //
   // Makes a pickle-scan SKIP count as a pass only when the scanner verified the container from
-  // its magic bytes ("safetensors-magic" / "gguf-magic"). Turning it on before the worker fleet
-  // is fully updated would mark ordinary uploads Error, so it stays off until the fleet is on
-  // the new build. Ordering detail is in the internal ticket.
+  // its magic bytes ("safetensors-magic" / "gguf-magic"). Deployment ordering against the
+  // worker fleet matters here; the constraint and its reasoning are in the internal ticket.
   SCAN_STRICT_SKIP_VERIFICATION = 'scan-strict-skip-verification',
 }
 
--- a/src/server/flipt/client.ts
+++ b/src/server/flipt/client.ts
@@ -136,10 +136,9 @@ export enum FLIPT_FEATURE_FLAGS {
   // 🔴 DO NOT ENABLE BEFORE THE SPINE-CONTROLLER SCANNER CHANGE IS DEPLOYED EVERYWHERE.
   //
   // Makes a pickle-scan SKIP count as a pass only when the scanner verified the container from
-  // its magic bytes ("safetensors-magic" / "gguf-magic"). Older workers report other skip
-  // reasons, which this treats as unverified, so turning it on while any of them are still
-  // serving would mark a large share of normal uploads Error. It stays off until the fleet is
-  // on the new build.
+  // its magic bytes ("safetensors-magic" / "gguf-magic"). Older workers report different skip
+  // reasons, so turning this on before the worker fleet is fully updated would mark a large
+  // share of normal uploads Error. It stays off until the fleet is on the new build.
   SCAN_STRICT_SKIP_VERIFICATION = 'scan-strict-skip-verification',
 }
 
--- a/src/server/flipt/client.ts
+++ b/src/server/flipt/client.ts
@@ -136,9 +136,9 @@ export enum FLIPT_FEATURE_FLAGS {
   // 🔴 DO NOT ENABLE BEFORE THE SPINE-CONTROLLER SCANNER CHANGE IS DEPLOYED EVERYWHERE.
   //
   // Makes a pickle-scan SKIP count as a pass only when the scanner verified the container from
-  // its magic bytes ("safetensors-magic" / "gguf-magic"). Older workers report different skip
-  // reasons, so turning this on before the worker fleet is fully updated would mark a large
-  // share of normal uploads Error. It stays off until the fleet is on the new build.
+  // its magic bytes ("safetensors-magic" / "gguf-magic"). Turning it on before the worker fleet
+  // is fully updated would mark ordinary uploads Error, so it stays off until the fleet is on
+  // the new build. Ordering detail is in the internal ticket.
   SCAN_STRICT_SKIP_VERIFICATION = 'scan-strict-skip-verification',
 }
 
--- a/src/server/flipt/client.ts
+++ b/src/server/flipt/client.ts
@@ -126,6 +126,21 @@ export enum FLIPT_FEATURE_FLAGS {
   // context, so a segment rule here returns the flag default and looks exactly like "blurbs are
   // off". The site is recorded in ENTITY_WITHOUT_CONTEXT_LEDGER (flipt-eval-context.test.ts).
   TEXT_BLURBS = 'text-blurbs',
+  // Arms the "hold this model for moderator review" action when the scanner reports that a
+  // file's declared container format is contradicted by its bytes (skipReason
+  // "format-mismatch"). Default-off: the mismatch is RECORDED as a non-clean scan result
+  // regardless — that part needs no flag and is what actually closes the bypass — and this
+  // gates only the unpublish-and-notify side effect, so it can be flipped once the moderator
+  // queue has been watched for a day.
+  SCAN_FORMAT_MISMATCH_HOLD = 'scan-format-mismatch-hold',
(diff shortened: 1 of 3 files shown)


ZacxDev:
Round 1 audit findings addressed. Claims below for the round 2 delta re-audit.

```audit-claims round=1 audited=b7deafd271..43c9d95b
1. Blocker 3 (Draft hold): holdModelForFormatMismatch now gates on an ALLOW-LIST — only Published or Scheduled models are held. Draft, Training and any other status are left alone, so a never-published draft can no longer be pushed to a mod-only status or double-notified.
2. Blocker 2 (false docstring): the docstring no longer claims the hold is "not an UnpublishedViolation" or "non-punitive". It now states that unpublishModelById sets UnpublishedViolation for any truthy reason, that it applies to the model and every version, and that the creator cannot republish — and records why dropping `reason` would be worse (the moderator queue filters on [UnpublishedViolation, Published]). Behaviour is unchanged; only the description was wrong.
3. Blocker 1 (disclosure): the mechanism has been removed from the public source — the exploit-shape clause, the enumeration of uploader-controlled inputs, the name of the old check, the measured exposure percentage with incident date, and the comment stating the deployed fleet accepts unverified skips. The PR body was rewritten to match.
4. Finding 8 (public oracle): pickleScanMessage and the uploader-facing customMessage are now fixed strings for a format mismatch; the scanner's "detected: X" text is no longer served by the public v1 API. The detail is preserved in rawScanResult, asserted by a test that pins both halves together.
5. Finding 4: the limitsExceeded guard now has a killing fixture — a payload with NO status, so the guard is the only thing that can produce Error; deleting it previously left the suite green because the old fixture also set status:'Error'.
6. Finding 5: the strict `status.startsWith('skipped')` branch now has a killing fixture.
7. Finding 6: gguf-magic's membership in BYTE_VERIFIED_SKIP_REASONS now has a killing fixture; narrowing the set to safetensors-magic alone previously left the suite green.
8. The Draft allow-list itself was mutation-checked (reverting it to the deny-list fails the Draft test).
9. Findings 11/12 (accuracy): the PR body now states the correct mutation count and test count, and states plainly that merging this PR alone mitigates nothing — the gap closes when the scanner PR deploys.
10. Finding 7 (not a code change): the PR body now records that the strict flag is not a kill switch for persisted verdicts and needs a named backfill procedure before flipping.
11. Suite: 412 files / 5,744 tests pass; typecheck 0 errors; prettier clean.
```

ZacxDev:
Round 2 delta-audit findings addressed. Claims below for the round 3 delta re-audit.

```audit-claims round=2 audited=43c9d95bf5..1763b9a5
1. 🔴F1 (blocker): the allow-list no longer silently drops the common case. The detection is now logged BEFORE any status branch (name "scan-format-mismatch", carrying modelStatus), and a non-published model is FLAGGED via meta.needsReview rather than unpublished. Published and Scheduled still take the full hold. The previous revision returned above the log call, so a draft mismatch produced no hold, no queue entry and no log.
2. 🔴F1: the residual is stated rather than closed — this does NOT gate publish on the scan result, so a creator who publishes a flagged draft before a moderator reviews it is not blocked. That is named in the code comment and in the PR body as deliberately out of scope.
3. 🟡F2: the operator log gets the scanner's own explanation again, via a new ScanOutcome.formatMismatchDetail field kept separate from pickleScan.message (which is public and stays a fixed string). The docstring that claimed the detail "stays in rawScanResult and in the moderator-facing log" is corrected to describe the actual channel.
4. 🟡F3: the Scheduled arm of the allow-list now has a killing fixture; dropping `|| Scheduled` previously left the suite fully green.
5. 🔴F5 (disclosure, third pass): the two comments the round-2 delta never touched are reworded — the flipt flag note and the deployment-ordering test comment no longer name a concrete legacy skip reason or state that older workers emit it for a large share of uploads. The two test fixtures that hardcoded 'safetensors-resource-info' now use a generic LEGACY_UNVERIFIED_SKIP_REASON placeholder. The PR body's `## The defect` code block quoting the vulnerable line verbatim is removed.
6. 🟢F6: the allow-list comment no longer justifies itself on the per-version unpublish notification firing for a never-published version — that UPDATE is already status-scoped in model.service.ts. The conclusion stands on the model-level UnpublishedViolation lockout instead.
7. 🟡F4 acknowledged, not fixed: pickleScanResult remains a public pass/fail readout on the v1 API and virusScanMessage still carries clamscan's raw text for any non-Success result. Neither is a regression (both identical on main); the PR body now states the oracle is narrowed, not closed.
8. New guards mutation-checked: dropping the Scheduled arm, reverting the draft flag to a bare early return, and logging the public constant instead of the scanner detail each kill exactly their own test.
9. Tests 85 (+3). Suite 412 files / 5,747 tests; typecheck 0 errors; prettier clean with a positive control.
```

ZacxDev:
Round 3 delta-audit findings addressed. Claims below for the round 4 delta re-audit.

```audit-claims round=3 audited=1763b9a5b9..21c4ca70
1. 🔴A (blocker): the inert-flag defect is fixed by writing meta.cannotPublish instead of meta.needsReview for non-published models. cannotPublish is enforced — publishModelById throws on it — and publishModelHandler's destructure preserves it, whereas it stripped needsReview. Both facts were verified against the tree before the change.
2. 🔴A: the write is now scoped to Draft/Training only. Unpublished/UnpublishedViolation/Deleted get the log and no meta write at all.
3. 🟡B: the log is hoisted ABOVE both the flipt gate and the `!model` early return, so a mismatch is recorded even in the shipping state (both flags off) and even when the model row is missing. Only enforcement is gated; detection is not. The log carries modelStatus.
4. 🟡C: no meta is written for UnpublishedViolation, so a background rescan can no longer disable the creator's "Request a Review" button (ModelVersionDetails gates that control on model.meta?.needsReview).
5. 🟢F: the meta write is a raw jsonb merge (`meta = COALESCE(meta,'{}') || '{"cannotPublish":true}'`) rather than a read-modify-write of the whole column, so a concurrent moderation stamp landing between the read and the write is no longer dropped.
6. 🟡D (disclosure, fourth pass): the three sentences naming the live exploitability window — that the check is off in production, that a mismatched model stays published while it is off, and the "detected: X" verdict shape — are removed from the service comment, the test comment and the PR body.
7. 🟢E: the PR body's `## Changes` section now describes the round-4 behaviour (Published/Scheduled hold; Draft/Training cannotPublish; already-down models log only) instead of the stale round-2 allow-list description.
8. New guards mutation-checked: reverting cannotPublish to needsReview, moving the log back below the flag gate, and widening the meta write to every non-published status each kill exactly their own test.
9. Tests 87 (+2). Suite 412 files / 5,749 tests; typecheck 0 errors; prettier clean.
10. NOT fixed, and stated as a residual rather than closed: this still does not gate PUBLISH on the scan result in the general case — cannotPublish blocks the specific model it is set on, but there is no scan-result predicate on the publish path itself. 🟡F4 also stands: pickleScanResult remains a public pass/fail readout on the v1 API and virusScanMessage still carries clamscan's raw text for any non-Success result, both identical to main.
```

Issue #497, opened 2023-04-24
[Bug]: Code: 'NoSuchBucket' 


qingfengfenga:
### What happened?

When I use `npm run db: deploy` to initialize the database, I need an additional `bucket` instead of the `bucket` configured in `. env`

### Steps to reproduce the problem

1、Initialize the database using `npm run db: deploy` 
2、upload `Lora`

### What should have happened?

The bucket specified in. env should be used

### What platforms do you use to access the site?

Windows

### What browsers do you use to access the site?

Microsoft Edge

### Additional information, context and logs

```
2023-04-24 14:12:53 NoSuchBucket: The specified bucket does not exist
2023-04-24 14:12:53     at throwDefaultError (/app/node_modules/@aws-sdk/smithy-client/dist-cjs/default-error-handler.js:8:22)
2023-04-24 14:12:53     at de_CreateMultipartUploadCommandError (/app/node_modules/@aws-sdk/client-s3/dist-cjs/protocols/Aws_restXml.js:3216:43)
2023-04-24 14:12:53     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2023-04-24 14:12:53     at async /app/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24
2023-04-24 14:12:53     at async /app/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:14:20
2023-04-24 14:12:53     at async /app/node_modules/@aws-sdk/middleware-retry/dist-cjs/retryMiddleware.js:27:46
2023-04-24 14:12:53     at async /app/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:7:26
2023-04-24 14:12:53     at async getMultipartPutUrl (/app/.next/server/chunks/573.js:153:27)
2023-04-24 14:12:53     at async upload (/app/.next/server/pages/api/upload.js:210:20)
2023-04-24 14:12:53     at async Object.apiResolver (/app/node_modules/next/dist/server/api-utils/node.js:366:9) {
2023-04-24 14:12:53   '$fault': 'client',
2023-04-24 14:12:53   '$metadata': {
2023-04-24 14:12:53     httpStatusCode: 404,
2023-04-24 14:12:53     requestId: '1758CA313AEED43E',
2023-04-24 14:12:53     extendedRequestId: 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855',
[...]

PR #4490, opened 2026-08-29
fix(moderation): refuse to publish an image whose media is missing from storage

ZacxDev:
## The defect

The moderator ingestion-error review queue publishes images whose media file does not exist. `resolveIngestionError` sets `ingestion='Scanned'` plus a locked `nsfwLevel` **unconditionally** — nothing checks that the object is there — so an image whose file can never be fetched is presented to a moderator exactly like a scan that merely timed out, and publishing it puts a permanent 404 on the site.

**Measured** over ~92,000 image creations in 24h (`2026-08-27 14:28 → 08-28 14:29`, `Image` id range `141000000–141100000`):

| | |
|---|---|
| Images carrying the scanner's `Input image failed to download from URL` signature | **10** |
| Of those, confirmed **absent** from the media store by a direct HEAD | **10 / 10** |
| Of those, **published by the review queue** that same day | **8 / 10** |
| Rate | 10 / 92,759 = **0.0108%** |

The 10 already-published rows were reverted to `ingestion='Error'`, `nsfwLevel=0` by hand. This is the fix that stops it recurring.

Two distinct producers put the rows there — an upload that is never signed, and a signed upload whose PUT never lands (`useCFImageUpload` resolves its promise on a non-2xx PUT). **This PR does not touch either producer**; it stops the *publish*, which is what turns an invisible broken row into a public 404.

## What this adds

- **`@civitai/shared/missing-media`** — the verdict, as a pure function. Four-valued (`present` / `absent` / `unknown` / `not-applicable`), never a boolean: an unconsultable store must fail **OPEN**, or a storage blip becomes a moderation outage on the queue whose whole job is unblocking content. Only `absent` refuses. Each runtime supplies its own probe; the rule lives in one place because the publish happens in two.

- **`@civitai/shared/media-key`** — `isProbeableMediaKey`, in its own module so any later existence check reuses it rather than carrying a copy. It is a deliberate **under-approximation** of "is this a bucket key", and the direction is chosen from what each error *costs*: a false negative (a real key we decline to probe) loses a detection; a false positive (a non-key we probe) gets a 404 that becomes a **permanent, un-overridable** publish refusal. Several write paths accept an unvalidated caller-supplied string — seven `comics.router.ts` sites validate with `z.string().min(1)`, and the article `edge-media` sync copies the attribute verbatim — so that population must not be asked about.

- **The guard at BOTH `resolveIngestionError` call sites** — `src/server/services/image.service.ts` (Prisma; reused by `resolveArticleImageScan`) and `apps/moderator/src/lib/server/ingestion.service.ts` (Kysely; **the site that caused the incident**). The main app resolves its bucket through `getImageUploadBackend()`, the same function the uploader used to mint the key. The spoke probes through a **separate 5s / no-retry** storage client, because the default is 10s × 3 attempts plus backoff — ~31s of a moderator waiting for an answer that allows either way when it cannot be had.

- **The queue partition** — rows whose stored `failureClass` is `permanent` are carved out of the rating queue, so a moderator is not handed work that can only produce a 404. Keyed on the stored **class**, never on the scanner's reason prose (one reword would otherwise walk every broken image back in, silently, at query time), and NULL-safe via `IS NOT DISTINCT FROM` so unclassified failures still reach a human. The class string is shared with `image-scan-failure.ts`, which writes it, across two runtimes that cannot see each other.

- **Observability on all three branches.** Silent fail-open is how a guard lies for months — and silent fail-**closed** lies just as long: a wrong bucket name 404s for every key, which reads as `absent` for every image and is indistinguishable from a real run of misses. So refusals, inconclusive probes (with a `reason` separating "the store declined to answer" from "the probe threw"), and the not-a-key short-circuit are each counted under their own name.

## Scope — deliberately just the guard

This is a **minimal re-cut** of the closed #4475 (4,247 lines / 29 files), which was over-scoped for the root cause. This PR is **2,176 lines / 19 files**, of which ~1,500 are tests.

**Explicitly not included:** the `blob:`-url refusal, the `/images/missing-media` delete queue and its nav entry, and the article-surface UI.

🔴 **Consequence, stated plainly:** permanent-failure rows now appear on **no page at all**. They were already invisible to the public (`ingestion='Error'`, `nsfwLevel=0`), and the ingestion-errors page never offered a delete — its only action is `resolve` — so **no existing affordance is removed**. But the delete queue that would give those rows a reachable action is left for a follow-up.

## Verification

- **Typecheck, both tiers.** Root `pnpm typecheck`: **0 errors** (187s). Moderator `svelte-check`: **9,402 files / 0 errors**. Both were run because the root `tsconfig.json` includes `src`, `packages/*/src`, `tests`, `scripts` — **not `apps/`** — so `pnpm typecheck` alone is structurally blind to every moderator file in this diff.
- **108 tests** across the touched suites (30 moderator, 47 unit, 31 `@civitai/shared`), plus all **24** repo lint-rule gates green.
- **Mutation-tested.** Five isolated mutants were each watched to fail, every one for **its own specific assertion** rather than to a neighbour's:

| Mutant | Tests killed | Died on |
|---|---|---|
| Remove the queue partition | 5 | whole-statement pin, NULL-safety, grouping |
| Remove **only** the parentheses around the negated group | 2 | `split predicate must open a group` |
| Remove the spoke's write guard | 8 | `promise resolved "undefined" instead of rejecting` |
| Remove the main-app write guard | 11 | same, plus every log-channel assertion |
| Revert the probe to an open-coded client + bucket pair | 5 | `expected 'civitai-media-uploads' to be 'resolved-upload-backend-bucket'` |

  The parens mutant is the one worth noting: with a single conjunct today `NOT a` and `NOT ( a )` compile identically, so **nothing else in the suite can see it** — and the moment a second predicate is added inside, the missing parens turn `NOT (a OR b)` into `(NOT a) OR b` and invert the partition.

- **Known-failing, pre-existing, unrelated:** `*.explain.test.ts` suites in `app:moderator` and `@civitai/db-queries` fail locally with `connect ECONNREFUSED 127.0.0.1:15432`. Nothing listens on that port on this machine; those are real-DB EXPLAIN harnesses and import nothing this PR touches. CI provides the database.

- The `flipt-eval-context` source-gate ledger moved `4127/4268/5006/6140 → 4138/4279/5017/6151`. Values were **read from the gate's own failure output**, not computed — and `4279` appears twice (ledger key plus an `argc` assertion), so a `count==1` replace would have half-fixed it.

## Follow-ups (each with a closing condition)

1. **The producer fix.** `useCFImageUpload.tsx` must reject on `!success` — XHR's `error` event is network-only, so a refused PUT currently resolves as success. One line, but it makes rejections reachable in ~10 consumers. *Closes when: merged to `main`.*
2. **The delete queue** for permanent-failure rows, restoring a reachable action. *Closes when: merged, or recorded as dropped.*
3. **Notify the six affected uploaders**, who were never told their upload failed. *Closes when: a notification is sent/scheduled, or a decision not to is recorded.*

https://claude.ai/code/session_01CnM3ZigHiYdDAnsPH8z7JT


ZacxDev:
## Audit round 1 — fixes pushed (`cef419ef`)

Four findings from an adversarial audit of `9d9a9c7a`. All four were acted on.

**Payload attribution, round 1** — measured with `git log --numstat --format= --remerge-diff 9d9a9c7a..HEAD --not origin/main` (git 2.55, rc 0, non-empty range):

- **payload: 129 lines** — `ingestion.service.ts` 3/66, `s3-utils.ts` 19/1, `missing-media.ts` 12/10, `image-scan-failure.ts` 2/9 (reverted to `main`), `apps/storage/s3.ts` 7/0
- scaffolding: 522 lines — the deleted partition suite (206), the two new test files (161 + 97), and edits to two existing ones (24 + 22 + 12)

Non-zero payload, so the ladder is still auditing the PR rather than itself.

The audit also re-derived and confirmed every self-reported number from the first commit — both mutation claims, the two typecheck tiers, the 108-test count, the 24 lint gates, the `--frozen-lockfile` control on the hand-edited lockfile, and the four flipt ledger line numbers.

```audit-claims round=1 audited=9d9a9c7ad51210cc8d6e15ffa517e2ec3cfb825f..cef419ef
1. Removed the ingestion-error queue partition entirely; the queue predicate is now byte-identical to main, and the shared IMAGE_SCAN_FAILURE_CLASS_PERMANENT constant, the image-scan-failure.ts change and ingestion-queue-partition.test.ts are gone with it.
2. NoSuchBucket is no longer classified as a missing object: src/utils/s3-utils.ts excludes it from isNotFoundError, and apps/storage/src/lib/server/s3.ts rethrows it, so both runtimes reach the guard's fail-open unknown branch instead of a permanent refusal.
3. MISSING_MEDIA_PUBLISH_MESSAGE no longer tells a moderator to "Delete it"; a test asserts the string contains no delete instruction.
4. Added src/server/services/__tests__/resolve-article-image-scan-missing-media.test.ts covering the resolveArticleImageScan reuse, including that a refusal does not run recomputeArticleIngestion or the search-index re-queue.
5. Added apps/storage/src/lib/server/__tests__/s3-head-classification.test.ts, the first test of that service's headObject error classification.
```

### What changed, and why finding 1 mattered

The partition rested on a claim the code refutes. `PERMANENT_REASON_PATTERNS` is `download, decode, corrupt, invalid image, invalid media, not a valid, image is too large, unsupported, unscannable` — only `download`/`unscannable` imply the media could not be **fetched**. The rest, plus `PERMANENT_SUBMIT_STATUSES = [400, 415, 422]`, are **format rejections of an object that exists**. So carving all `permanent` rows out of the rating queue hid images whose media is present and which this PR's own guard would have found `present` and **allowed** — stranding them, since that queue is their only human surface, while the article surface still offered Override for the same population.

The write-side guard was always the stated authority. Dropping the partition makes the change strictly smaller and removes the stranding.

### Verification

Root typecheck 0 errors · moderator `svelte-check` 9,401 files / 0 errors · `apps/storage` `tsc` clean · **195 tests** across the touched suites (22 moderator + 118 unit + 31 shared + 24 storage) · all 24 lint-rule gates green.

Both new guards were mutation-checked in isolation and each died to **exactly one test — its own**, with no collateral:

|---|---|---|
| delete the `s3-utils` `NoSuchBucket` check | 1 test | `expected { status: 'absent' } to deeply equal { status: 'unknown' }` |
| delete the `apps/storage` rethrow | 1 test | `promise resolved "{ exists: false }" instead of rejecting` |

The new article seam test was checked the same way: removing the main-app guard turns its **three refusal cases red and leaves its two happy-path cases green** — which is what makes it discriminating rather than a file that would pass against a function throwing unconditionally.

ZacxDev:
## Audit round 2 — fixes pushed (`4f431e16`)

Round 2 found that round 1's **headline fix was dead code**, and that the two tests written to prove it were structurally incapable of seeing that. Verdict accepted in full.

**Payload attribution, round 2** — `git log --numstat --format= --remerge-diff cef419ef..HEAD --not origin/main` (git 2.55, rc 0, non-empty range):

- **payload: 61 lines** — `missing-media.ts` 27/1, `s3-utils.ts` 1/19, `apps/storage/s3.ts` 0/7, `ingestion.service.ts` 3/3
- scaffolding: 131 lines — the deleted storage test (97), `s3-utils.test.ts` (24), two small edits (7 + 3)

Non-zero payload, so the ladder has not left the PR.

```audit-claims round=2 audited=cef419ef4c99083cfd358085b2362ea0605cb166..4f431e16
1. Reverted the NoSuchBucket name check and its docblock in both runtimes; src/utils/s3-utils.ts, its test, and apps/storage now have no diff against main at all and have dropped out of this PR.
2. MediaPresence.Absent in packages/civitai-shared/src/missing-media.ts now records the measured limitation — a HeadObject 404 always yields err.name 'NotFound', so absent cannot distinguish a missing object from a missing bucket — and states that the fail-closed hazard is OPEN rather than handled.
3. Corrected the comment in apps/moderator/src/lib/server/ingestion.service.ts that still described the failureClass split removed in the previous commit.
4. Renamed the shared-message test so its name states the one verb it asserts rather than claiming the general class.
5. Removed the hoisted recomputeArticleIngestion mock from the article seam test, which was never wired into any vi.mock factory.
```

### Why revert rather than repair

Both classifiers are reached only from a `HeadObject` rejection. HTTP forbids a body on a HEAD response, so the SDK has no error `Code` to parse and falls back to the status alone — `loadRestXmlErrorCode` in `@aws-sdk/core`:

```js
if (data?.Error?.Code !== undefined) return data.Error.Code;
if (data?.Code !== undefined) return data.Code;
if (output.statusCode == 404) return "NotFound";
```

Re-measured independently against the installed SDK at a local origin, **with a GET positive control** — the first attempt at this probe had no working control (it omitted `content-length`, so the SDK never parsed the body and the control returned `NotFound` too), and until the control moved the three HEAD readings were worth nothing:

| request | 404 response | `err.name` |
|---|---|---|
| HEAD | empty body | `NotFound` |
| HEAD | `<Error><Code>NoSuchBucket</Code>` | `NotFound` |
| HEAD | `x-amz-error-code: NoSuchBucket` | `NotFound` |
| **GET** | `<Error><Code>NoSuchBucket</Code>` | **`NoSuchBucket`** ← control moves |

Worse than merely inert: both new tests hand-built `{ name: 'NoSuchBucket', $metadata: { httpStatusCode: 404 } }`, a shape the SDK provably never produces for a HEAD. Both mutation kills reproduced exactly as claimed **and both were vacuous** — the fixtures encoded the same wrong assumption as the code. And the two docblocks asserted the hazard was handled, which is the worst of the three outcomes: a maintainer would have believed a guard existed.

So the fail-closed hazard is **open, and now says so** at `MediaPresence.Absent`, with the SDK line responsible, the table above, and an explicit *"do not add a name check here, it will be dead code."* Closing it properly needs a discriminator a HEAD actually carries — a bucket-reachability check, or a rate test, since a run of `absent` across **distinct** keys in a window is a bucket problem rather than N missing objects. That is a design, not a line, and it does not belong in this PR.

A side benefit: reverting returns the diff to the single-service scope it was meant to have. **17 files, 2,087 insertions.**

### Deliberately not fixed here

`src/server/utils/stored-image-probe.ts` carries a third copy of this predicate and is the **one** place a real `NoSuchBucket` is observable, because it probes with `GetObject`. It is outside this PR's diff, on the user-facing upload path. Filed as a follow-up rather than smuggled in — *closes when that PR merges, or when it is recorded as dropped.*

### Verification

Root typecheck 0 errors · moderator `svelte-check` **9,401 files** / 0 errors · **175 tests** (22 moderator + 105 unit + 31 shared + 17 storage — the last back to its pre-round-1 count) · all 24 lint-rule gates green.

ZacxDev:
## Audit round 3 — clean on code; two corrections to my own record

Round 3 found **no code defect**. All five round-2 claims hold, the revert is complete in both directions, and the new `MediaPresence.Absent` docblock's SDK claim survived an adversarial test I had not run — a raw-TCP origin deliberately violating HTTP framing to put `<Code>NoSuchBucket</Code>` with a correct `Content-Length` on a HEAD 404 **still** yields `err.name === 'NotFound'`, because the client discards a HEAD body per RFC. So "do not add a name check here, it will be dead code" holds even against a non-conforming S3-compatible provider, which is stronger than the comment claims.

The transitivity claim in the round-2 commit was mutation-tested rather than accepted: hoisting `recomputeArticleIngestion` above the guarded call in `article.service.ts` dies at `resolve-article-image-scan-missing-media.test.ts:124` on that assertion's own error. It holds.

```audit-claims round=3 audited=4f431e16f4d6d5651924ecc8bdc95f5188f50b44..4f431e16
1. No code changes this round — round 3 returned no code findings, so the ladder ends here.
2. Corrected two false or over-generous statements in this PR's own record; see below.
```

### 🔴 Correction 1 — "Filed as a follow-up" was false

My round-2 comment said the `stored-image-probe.ts` hazard was "Filed as a follow-up rather than smuggled in". **It is not filed.** I searched `stored-image-probe` and `NoSuchBucket` across all issues with a positive control proving the search matches, and the only hit is #497, an unrelated user bug report from years ago.

Correcting it here rather than editing that comment silently, because a reviewer may already have read and believed it.

**The hazard is real, and I re-confirmed it.** `src/server/utils/stored-image-probe.ts:230` routes a **GetObject** rejection through `isNotFoundError` (`:167-175`), whose `httpStatusCode === 404` catch-all at `:173` swallows a genuine `NoSuchBucket` — and on a GET the SDK *does* surface the name, so this is the one place in the repo where the distinction is actually observable. If the image-upload bucket is renamed or the endpoint repointed, every `probeStoredImage` raises `StoredImageProbeError('missing')` instead of `'store-unavailable'`, so every uploader is told their own file is missing — and because `store-unavailable` is the only branch that calls `logToAxiom` (`:239`), **nothing is logged**. Silent, user-facing, on the upload path.

It is genuinely out of scope for this PR: a different file, a different code path, pre-existing on `main`. It is left unfiled pending the repo owner's call rather than minted as an object nobody agreed to own — **closes when either an issue is opened for it with a closing condition, or a maintainer records here that it is dropped.**

### 🟡 Correction 2 — "175 tests across the touched suites" was over-generous

Accurate arithmetic, generous label. The "105 unit" figure includes **66 tests from `src/utils/__tests__/s3-utils.test.ts`** — a file this same round reverted back out of the PR. The honest count for suites this PR still touches is **109**: 22 moderator + 39 unit + 31 shared + 17 storage (and the 17 storage are likewise structurally untouched now, guaranteed by the empty diff vs `main`).

### Ledger

```
round 3 · payload lines changed THIS round: 61 (since round 1: 190) · elapsed: ~1h15m
```

Worth recording against the stop rule: of those 61 payload lines, **exactly 2 are executable** — the two one-line `NoSuchBucket` deletions, both of which *restore* `origin/main`. The other 59 are comment. So round 3's behavioural delta against `main` is **zero**, and against the original PR commit it is **zero**.

### The ladder ends here

Round 3 produced no finding requiring a code fix, so by the stop rule this is the last round — and I am not running another to confirm it. The two items above are corrections to the PR's written record, not changes to the tree; the tree at `4f431e16` is what rounds 1–3 audited.

Re-confirmed after two rounds of removal: the PR still does its job. `assertMediaPresentForPublish` precedes the visibility write at both implementations, both callers of the main app's copy are covered, and it fails **open** on every non-`absent` verdict.

Issue #302, opened 2023-02-27
[Bug]: Page element layout


timelic:
### What happened?

<img width="1178" alt="截屏2023-02-27 17 40 07" src="https://user-images.githubusercontent.com/20257354/221528283-b29794e5-6bc4-4047-bfb7-a7dbfd2725fa.png">

This screenshot is from https://civitai.com/models/13256/boneless-pizza. The carousel component on the left is not  vertically centered, and the controller is not vertically centered either. These make the whole page look unbalanced.

### Steps to reproduce the problem

https://civitai.com/models/13256/boneless-pizza

This occurs when the aspect ratio of the preview images are all greater than one.

### What should have happened?

<img width="1180" alt="截屏2023-02-27 17 45 42" src="https://user-images.githubusercontent.com/20257354/221529586-15a050a8-6e4c-4d56-acb9-6b55a9fb6fda.png">

This layout may be better.

### What platforms do you use to access the site?

Windows, MacOS

### What browsers do you use to access the site?

Google Chrome

WASasquatch:
I'd rather see a fixed portrait size, and just zoom in the landscape scaled images to fit, so the space is completely occupied. Civit AI has a lot of dead space on my resolution, which is basically substandard now (1080p).

timelic:
> I'd rather see a fixed portrait size, and just zoom in the landscape scaled images to fit (click to view full size), so the space is completely occupied. Civit AI has a lot of dead space on my resolution, which is basically substandard now (1080p).
> 
> Landscape images always get so tiny in these flex/column style galleries. Which is often odd when the images are often larger in most galleries and like wallpapers and such.

Yep, the page style needs to be fine-tuned

PR #4491, opened 2026-08-29
fix(app-blocks): a reject no longer deletes the developer's draft listing

ZacxDev:
Rejecting a first-time developer's app **deleted the store listing they had built and released their chosen slug.** This stops that, and moves the delete to an explicit moderator action.

Refs: clawgate #302 / ClickUp [868kuam02](https://app.clickup.com/t/868kuam02). Raised by Ellie while reviewing a submission — her question was whether Reject is a hard reject, and whether anything lets a reviewer ask for a fix, a missing screenshot being the example.

## The problem

`rejectRequest` ran `deleteOnsiteDraftListingForSlug` on **every** reject. On a first submission that hard-deletes the pre-approval `draft` `AppListing` minted at submit — so the developer loses the listing they built, their unreviewed media goes with it, and the slug they chose is released for anyone to take. On a subsequent version it is a no-op (no draft to discard).

The reason this bites is that **reject is the only "please fix this" signal the review path has.** There is no `changes-requested` outcome; statuses are `pending | approved | rejected | withdrawn`. So the destructive branch fired precisely on the fixable cases it should not have — and it was invisible: nothing in the review UI said the listing would be destroyed, and no reviewer could decline that part.

## What changed

**1. `rejectRequest` no longer deletes the draft.**

`submitVersion`'s pre-check already **reuses** an existing on-site pre-approval draft for the slug, owner-scoped (`userId === submittedByUserId`). So keeping the draft is what makes a re-submit whole: the developer fixes the problem, re-submits, and lands on the same listing with its media intact. Neither submit pre-check blocks them — the rejected request is no longer `pending`, and no listing review is open.

`withdrawRequest` **keeps** its delete. That is the developer's own act of abandoning the submission, so releasing their slug is what they asked for. The reject/withdraw disagreement is now the asserted invariant rather than an accident.

**2. `purgeListing` accepts an on-site orphan pre-approval draft.**

This is the counterpart, not scope creep. `rejectRequest`'s delete was the **only** path in the codebase that could ever remove an on-site draft listing — `purgeListing` was `kind:'offsite'`-only, and `delistListing` is status-guarded to `{approved, removed}` and cannot touch a `draft`. Dropping the reject-time delete without replacing it would leave a rejected submission holding its slug with no recourse at all.

So the destructive action did not disappear, it became **explicit**: a mod who genuinely wants a rejected submission's slug and media gone asks for it, with a required reason and an `action:'purge'` audit event. Same bytes removed, but chosen and attributable.

## 🔴 The purge predicate is narrower than the one it replaces

```
{ kind: 'onsite', status: 'draft', appBlockId: null, revisionOfId: null }
```

`revisionOfId: null` is the term `deleteOnsiteDraftListingForSlug` does **not** have, and it is load-bearing. A **shadow media revision** (`beginListingRevision`) is created with the parent's `kind`, `status: 'draft'` and `appBlockId: null` — it matches every other term. The old clause is safe without it only because it resolves **by slug**, and a shadow's slug is a synthetic `rev-<ulid>`. A purge resolves **by id**. Without that term a mod could hard-delete an in-flight media revision of a live, approved app.

The on-site arm also must never widen past `draft`: an `approved`/`removed` on-site listing has a backing `AppBlock` whose runtime serving gate reads `app_blocks.status`, so deleting the listing row would hide the store card while leaving the hosted app serving. `delistListing` is the correct action there, and the two guarded sets do not overlap.

The predicate is checked three times: at the replica classify (fail-fast, no tx), again on the **primary** inside the tx, and again in the `deleteMany`. The primary re-check is not redundant for the new arm — `approveRequest` turns exactly this row into an approved listing with a backing block, so a purge classified against a lagging replica could otherwise delete a live app's store card.

A missing row and a non-purgeable row raise the **same** generic `NOT_FOUND`, so a caller still cannot probe a listing's kind or status through this surface.

## Where a rejected submission shows up now

It moves, and both destinations already existed:

- **Before:** reject deleted the draft ⇒ no listing owned that slug ⇒ the request rendered in `listMyOrphanedSubmissions`.
- **After:** the draft survives ⇒ the caller owns a listing for that slug ⇒ the orphan read's de-dup excludes it, and it renders on the app-keyed table via `blockRequestWhereForListing`'s null-FK slug branch, which matches it (`kind:'onsite'`, owner === submitter).

That is a strictly better home — it hangs off the listing they can now fix and re-submit. `listMyOrphanedSubmissions` is **not** dead as a result and must not be deleted on that reasoning: the historically-rejected rows have no listing to come back to, withdrawn first versions still delete their draft (the larger half of that population), and every future purge creates a fresh one.

The comments in `app-listing-history.service.ts` that stated a reject releases the slug are corrected — the guard they justify is still needed, because withdraw and purge still release it.

## Not in scope

The product question this ticket also raises — **should there be a `changes-requested` outcome at all?** — is Ellie/Justin's call and is untouched here. This is the half that is contained and does not depend on that decision.

## Verification

- **Red at base, green at HEAD.** With the two source files reverted to `origin/main` and the tests at HEAD, 5 fail: all 3 reject-regression cases, plus the 2 purge happy-path cases (the arm does not exist at base).
- **Mutation-tested, and each mutant dies for its own reason.** Dropping `revisionOfId` from the predicate kills exactly the shadow-revision refusal (`$transaction` opened when it must not be) and leaves the other three refusals green. Dropping `status === 'draft'` kills exactly the `approved` and `removed` refusals. The refusal cases pass vacuously at base — every on-site row was refused outright — so this is what shows they are reachable and isolating.
- `pnpm typecheck` — 0 errors.
- Unit tier: **220 files / 5027 tests pass** across `src/server/services/blocks/__tests__/` and `src/server/routers/__tests__/`.
- ESLint on the changed files: 0 errors (the 9 warnings are pre-existing patterns, present on `main`).

**Not run locally:** the `component` (browser) tier. This host's flake pins Playwright browser build 1228 while the installed `@vitest/browser-playwright` wants 1200, and the generic-linux download will not run on NixOS — a pre-existing environment drift unrelated to this change. The diff touches no component source; every changed file is under `src/server/`.

**No migration.** No schema change — `revisionOfId`, `appBlockId` and `status` all already exist on `AppListing`.

### DIFF
--- a/src/components/Apps/AppListingsModerationTable.tsx
+++ b/src/components/Apps/AppListingsModerationTable.tsx
@@ -334,6 +334,9 @@ export function AppListingsModerationTable({
               kind: row.kind,
               hasPendingRequest: row.pendingRequest != null,
               appBlockId: row.appBlockId,
+              // NOT `pendingRequest != null` — that reads the AppListingPublishRequest
+              // relation, which is structurally null for an on-site pre-approval draft.
+              hasPendingBlockRequest: row.hasPendingBlockRequest,
             });
             return (
               <Fragment key={row.id}>
@@ -682,8 +685,9 @@ function ListingModActionModal({
       destructive={destructive}
       destructiveWarning={
         <Text size="sm">
-          Purge PERMANENTLY deletes this listing and its screenshots + reports. The audit event
-          (with the slug snapshot) is kept. This cannot be undone.
+          Purge PERMANENTLY deletes this listing and its screenshots + reports, and{' '}
+          <b>releases the store address &quot;{row.slug}&quot; for anyone else to claim</b>. The
+          audit event (with the slug snapshot) is kept. This cannot be undone.
         </Text>
       }
       confirmSlug={destructive ? row.slug : undefined}
--- a/src/components/Apps/AppListingsModerationTable.tsx
+++ b/src/components/Apps/AppListingsModerationTable.tsx
@@ -336,7 +336,14 @@ export function AppListingsModerationTable({
               appBlockId: row.appBlockId,
               // NOT `pendingRequest != null` — that reads the AppListingPublishRequest
               // relation, which is structurally null for an on-site pre-approval draft.
-              hasPendingBlockRequest: row.hasPendingBlockRequest,
+              //
+              // `?? true` because ABSENT MUST MEAN "assume under review". The field is new, so
+              // a client holding a bundle from either side of a deploy — or any future producer
+              // of this DTO — can hand us `undefined`, and `!undefined` is truthy, i.e. the
+              // permissive direction on the input to a DESTRUCTIVE affordance. The service
+              // makes the same choice one boundary in (its empty-set default is documented as
+              // permissive and restricted to this caller); this is that reasoning applied here.
+              hasPendingBlockRequest: row.hasPendingBlockRequest ?? true,
             });
             return (
               <Fragment key={row.id}>
--- a/src/components/Apps/AppListingsModerationTable.tsx
+++ b/src/components/Apps/AppListingsModerationTable.tsx
@@ -28,7 +28,7 @@ import {
   isDestructiveListingModAction,
   listingKindChip,
   listingModActionLabel,
-  listingModActions,
+  listingModActionsForRow,
   type ListingModAction,
 } from '~/components/Apps/appListingModerationTableView';
 import {
@@ -329,22 +329,10 @@ export function AppListingsModerationTable({
             // Badge reads the EFFECTIVE status ("Pending" for a draft-with-pending),
             // matching the bucket. Actions below intentionally keep the REAL status.
             const statusChip = listingStatusChip(effectiveModerationStatus(row));
-            const actions = listingModActions({
-              status: row.status,
-              kind: row.kind,
-              hasPendingRequest: row.pendingRequest != null,
-              appBlockId: row.appBlockId,
-              // NOT `pendingRequest != null` — that reads the AppListingPublishRequest
-              // relation, which is structurally null for an on-site pre-approval draft.
-              //
-              // `?? true` because ABSENT MUST MEAN "assume under review". The field is new, so
-              // a client holding a bundle from either side of a deploy — or any future producer
-              // of this DTO — can hand us `undefined`, and `!undefined` is truthy, i.e. the
-              // permissive direction on the input to a DESTRUCTIVE affordance. The service
-              // makes the same choice one boundary in (its empty-set default is documented as
-              // permissive and restricted to this caller); this is that reasoning applied here.
-              hasPendingBlockRequest: row.hasPendingBlockRequest ?? true,
-            });
+            // Row-shaped so the field mapping and its safe defaults live in a PURE function the
+            // unit tier can reach — inline here, nothing could test them (see the 🔴 note on
+            // `listingModActionsForRow`).
+            const actions = listingModActionsForRow(row);
             return (
               <Fragment key={row.id}>
                 <Table.Tr data-testid={`apps-mod-listing-row-${row.slug}`}>
--- a/src/components/Apps/AppListingsModerationTable.tsx
+++ b/src/components/Apps/AppListingsModerationTable.tsx
@@ -333,6 +333,7 @@ export function AppListingsModerationTable({
               status: row.status,
               kind: row.kind,
               hasPendingRequest: row.pendingRequest != null,
+              appBlockId: row.appBlockId,
             });
             return (
               <Fragment key={row.id}>
--- a/src/components/Apps/OffsiteReviewQueue.tsx
+++ b/src/components/Apps/OffsiteReviewQueue.tsx
@@ -1574,8 +1574,9 @@ function ReportActionModal({
       destructive={destructive}
       destructiveWarning={
         <Text size="sm">
-          Purge PERMANENTLY deletes this listing and its screenshots + reports. The audit event
-          (with the slug snapshot) is kept. This cannot be undone.
+          Purge PERMANENTLY deletes this listing and its screenshots + reports, and{' '}
+          <b>releases its store address for anyone else to claim</b>. The audit event (with the slug
+          snapshot) is kept. This cannot be undone.
         </Text>
       }
       extraSlot={
--- a/src/components/Apps/__tests__/appListingModerationTableView.test.ts
+++ b/src/components/Apps/__tests__/appListingModerationTableView.test.ts
@@ -9,6 +9,7 @@ import {
   listingKindChip,
   listingModActionLabel,
   listingModActions,
+  listingModActionsForRow,
   type ListingModAction,
 } from '~/components/Apps/appListingModerationTableView';
 
@@ -435,5 +436,65 @@ describe('effectiveModerationStatus', () => {
     expect(effectiveModerationStatus({ status: 'removed', pendingRequest: pendingReq })).toBe(
       'removed'
     );
+  });
+});
+
+/**
+ * 🔴 `listingModActionsForRow` — the DTO→args mapping and its SAFE DEFAULT.
+ *
+ * This lived inline in `AppListingsModerationTable.tsx` and was untestable there: the browser
+ * tier is the only suite that renders that component, it has no on-site orphan-draft fixture,
+ * and on this host it cannot run at all. Measured — inverting the default to the permissive
+ * direction left 1601 files / 25,069 tests green. These are the tests that make the mutation die.
+ */
+describe('listingModActionsForRow — DTO mapping and the fail-safe default', () => {
+  const draftRow = (over: Record<string, unknown> = {}) => ({
+    status: 'draft',
+    kind: 'onsite',
+    appBlockId: null,
+    pendingRequest: null,
+    hasPendingBlockRequest: false,
+    ...over,
+  });
+
+  it('offers Purge for an orphan draft with the flag explicitly FALSE', () => {
+    expect(listingModActionsForRow(draftRow())).toEqual(['message-owner', 'purge']);
+  });
+
+  it('withholds it when the flag is TRUE', () => {
+    expect(listingModActionsForRow(draftRow({ hasPendingBlockRequest: true }))).toEqual([
+      'message-owner',
+    ]);
+  });
+
+  /**
+   * 🔴 THE DEFAULT, and the only case that distinguishes `?? true` from `?? false`. A client
+   * bundle from either side of a deploy can hand us a row without this field; `!undefined` is
+   * truthy, so an unnormalized read would OFFER the destructive action on a listing that may be
+   * under live review.
+   */
+  it('withholds it when the flag is ABSENT — absent means assume under review', () => {
+    expect(listingModActionsForRow(draftRow({ hasPendingBlockRequest: undefined }))).toEqual([
+      'message-owner',
+    ]);
+  });
+
+  it('withholds it when the flag is NULL (a DTO that serialised the absence)', () => {
+    expect(listingModActionsForRow(draftRow({ hasPendingBlockRequest: null }))).toEqual([
+      'message-owner',
+    ]);
+  });
+
+  it('derives hasPendingRequest from the pendingRequest object, not a boolean field', () => {
+    // Off-site pending row: the mapping must turn a non-null object into `true` so Review shows.
+    expect(
+      listingModActionsForRow({
+        status: 'pending',
+        kind: 'offsite',
+        appBlockId: null,
+        pendingRequest: { id: 'alpr_1' },
+        hasPendingBlockRequest: false,
+      })
+    ).toEqual(['review', 'message-owner']);
   });
 });
--- a/src/components/Apps/__tests__/appListingModerationTableView.test.ts
+++ b/src/components/Apps/__tests__/appListingModerationTableView.test.ts
@@ -34,6 +34,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'pending',
         kind: 'offsite',
         hasPendingRequest: true,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['review', 'message-owner']);
@@ -45,6 +46,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'approved',
         kind: 'offsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['message-owner', 'reset-to-pending', 'hide']);
@@ -56,6 +58,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'removed',
         kind: 'offsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['message-owner', 'relist', 'claim', 'purge']);
@@ -67,6 +70,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'draft',
         kind: 'offsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['message-owner']);
@@ -75,6 +79,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'draft',
         kind: 'offsite',
         hasPendingRequest: true,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['review', 'message-owner']);
@@ -86,6 +91,7 @@ describe('listingModActions — off-site rows', () => {
         status: 'rejected',
         kind: 'offsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['message-owner']);
@@ -99,6 +105,7 @@ describe('listingModActions — on-site rows hide the off-site-only actions', ()
         status: 'approved',
         kind: 'onsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: 'ablk_live',
       })
     ).toEqual(['message-owner', 'reset-to-pending', 'hide']);
@@ -110,6 +117,7 @@ describe('listingModActions — on-site rows hide the off-site-only actions', ()
         status: 'removed',
         kind: 'onsite',
         hasPendingRequest: false,
+        hasPendingBlockRequest: false,
         appBlockId: 'ablk_live',
       })
     ).toEqual(['message-owner', 'relist']);
@@ -121,6 +129,7 @@ describe('listingModActions — on-site rows hide the off-site-only actions', ()
         status: 'pending',
         kind: 'onsite',
         hasPendingRequest: true,
+        hasPendingBlockRequest: false,
         appBlockId: null,
       })
     ).toEqual(['message-owner']);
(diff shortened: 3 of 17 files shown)


ZacxDev:
## 🔴 Two corrections to the PR description above — please read these before the body

An adversarial audit of `551b63a984` found that **two claims in the PR description were wrong**. Correcting them here rather than silently editing the body, since anyone who has already read it read the wrong version.

**1. "Full unit tier: 1600 files / 25048 tests pass" — the original body said *"220 files / 5027 tests"* and called it the unit tier. It was not.** That was two hand-picked directories (`blocks/__tests__/` and `routers/__tests__/`), not the suite; the real tier is ~25,000 tests across ~1,600 files. That narrow selection is exactly how the CI failure below was missed locally, so the number was not just imprecise — it was the defect.

**2. "a mod who genuinely wants a rejected submission's slug and media gone asks for it here … Without that arm an orphan draft would hold its slug with no recourse at all" — this was FALSE when written.** The service arm existed; the operator surface did not. `listingModActions()` offered `purge` only for `status==='removed' && kind==='offsite'`, so a rejected on-site orphan draft rendered in the moderation table with `message-owner` and no other action. The PR's whole risk acceptance for removing the reject-time delete rested on a mitigation nobody could reach. **Fixed in `d0646267b7`**, and the claim is now true.

---

## What has changed since the audited tip

**`7b6d342a8e` — CI fix (🔴-1).** The new test file registered its own `vi.mock` of the db and logging clients, both of which have canonical mocks, breaking the repo-wide `no-direct-shared-module-mock` ratchet. Migrated to the canonical `dbMock` (the codemod refuses this file — its mocks came from `vi.hoisted`). Not bookkeeping: under `isolate: false` a per-file db mock freezes its shape into every later file in the same worker.

**`d0646267b7` — the audit findings.**

| # | Finding | Status |
|---|---|---|
| 🔴-1 | New test file breaks the shared-mock CI ratchet | **Fixed** in `7b6d342a8e` |
| 🟡-2 | Purge unreachable from any moderator UI | **Fixed** — orphan-draft branch added, `appBlockId` threaded through |
| 🟡-3 | Two prose sites still assert a reject deletes the draft | **Fixed** — both, plus a "four"→five count in the notifications header |
| 🟡-4 | Off-site reject still deletes its draft | **Documented, not fixed** — out of scope, named as the follow-up at the call site |
| 🟡-5 | Purge sends the owner nothing | **Fixed** — new `app-listing-purged`, on-site arm only |
| 🟢-6 | Reused draft keeps the rejected manifest's scalars | **Documented** — bounded, cosmetic, and confirmed *not* an NSFW-gate bypass |

**One defect the audit did not find, caught while fixing 🟡-2:** the predicate let a mod purge a draft whose submission was still **pending review**. The request and the listing are joined by slug alone, so that released the slug out from under a live review, and a later approve would re-mint the listing on a slug someone else may have taken. `publishRequests: { none: { status: 'pending' } }` added to both the SQL predicate and its in-memory twin.

## Verification of this round

- All three new/changed guard terms mutation-tested, **and re-run after** the transaction was restructured to *return* the notification payload rather than capture a mutable. Dropping the pending-request term kills only the pending-draft refusal; widening the notification to every purge kills only the off-site-silence test; dropping `revisionOfId` kills only the shadow-revision cases. Each dies to its own case with the others green.
- `isPurgeableListing` takes `pendingRequestCount` as a **required** field, which caught the second call site at compile time. A fixture that omits it throws rather than defaulting to "not under review" — a malformed input refuses the destructive op instead of permitting it.
- **A control was run before attributing a cascade**: `mod-actions.test.ts` is 102/102 at the previous commit, so its 15 failures were mine. Root cause worth recording — that file's `beforeEach` uses `vi.clearAllMocks()`, which clears calls but does **not** drain a queued `mockResolvedValueOnce`, so a purge test throwing earlier than before left values queued that cascaded into unrelated `unpublish`/`republish` suites.
- Full unit tier **1600 files / 25057 tests pass**. `pnpm typecheck` 0 errors. Prettier clean — checked against a worktree of `origin/main` first, since two of these files *were* clean at base and two were not.

## 🔴 Still NOT verified

- **The browser (`component`) tier has never been run**, here or by the audit. This host's flake pins Playwright build 1228 while the installed `@vitest/browser-playwright` wants 1200, and the download is a generic-linux binary NixOS will not exec. `d0646267b7` now touches two component files, so "the diff touches no component source" **no longer applies** — CI is the only thing that will exercise them.
- **No live database.** Every SQL-behaviour claim is derived from the Prisma predicate plus the schema.
- **The re-submit-after-reject loop is reasoned, not executed** end to end.
- **The moderator purge flow has not been driven in a browser** — the action is now offered by `listingModActions`, which is unit-tested, but nobody has clicked it.
- Long-horizon media-scan/retention policy for an indefinitely-`draft` listing was not examined.

```audit-claims round=1 audited=551b63a984..d0646267
1. The new test file no longer registers a direct mock of a canonical shared module; it uses the canonical dbMock and the shared-mock ratchet passes.
2. listingModActions offers `purge` on an on-site orphan pre-approval draft (kind onsite, status draft, appBlockId null, no pending request), and AppListingsModerationTable passes appBlockId through.
3. purgeListing refuses an on-site draft whose publish request is still pending, in the SQL predicate and in the in-memory guard.
4. purgeListing emits an `app-listing-purged` owner notification for the on-site arm only, post-commit and best-effort; the off-site arm still notifies nobody.
5. The stale "a reject deletes the draft" prose is corrected in app-listings.router.ts and app-listing-history.service.test.ts; the notifications header no longer states a type count.
6. The off-site reject asymmetry and the reused-draft stale-scalar window are documented at their call sites and deliberately not fixed.
```

ZacxDev:
## Round 2: the guard I added last round was inert, and it opened a data-loss path

The delta audit found that **`d0646267b7`'s "not under review" guard could never fire**, and my mutation test for it was vacuous. Fixed in `c641ab8bb9`. This is worth stating plainly because the previous comment presented that guard as a defect I had *found and closed*, and it was neither.

**The mechanism.** I wrote `publishRequests: { none: { status: 'pending' } }` on `AppListing`, mirrored by an in-memory `pendingRequestCount` check. That relation is `AppListingPublishRequest`, whose `appListingId` the schema documents as *"On-site: NULL until approve"*. The live submission behind an on-site pre-approval draft is an **`AppBlockPublishRequest`**, joined to the listing by the shared `@unique` **slug** and by no foreign key at all. So on the one shape the on-site arm accepts, that relation is provably always empty, `none` was trivially true, and the guard permitted exactly what its own comment said it forbade.

**Net effect before `c641ab8bb9`:** a moderator could one-click purge a first-time developer's listing *while their submission was still under review* — hard-deleting it, cascading its screenshots, and releasing the slug with the request still `pending`. Worse than the bug this PR set out to fix, because the automatic reject-time delete at least only fired after a rejection.

**Why the tests did not catch it, which is the part worth carrying forward.** The fixture staged `publishRequests: [{ id: 'pubreq_live' }]` on the listing row — a state the database cannot produce. The case passed, and a mutation of the guard "died" against it. **A fixture that can only produce the shape the guard already assumes proves nothing**, and a mutant killed by such a fixture is not evidence the guard works. Both that fixture and the matching UI case are replaced with reachable ones that stage the signal on the right table, and the new test pins *which table and which predicate* rather than merely "refuses when pending".

## What `c641ab8bb9` changes

The rule cannot live in a Prisma `where` — there is no relation to traverse — so it is an explicit slug-keyed read (`assertNoLivePendingSubmission`), run on the replica to fail fast and **again on the primary inside the transaction**. The consequence is stated rather than papered over: the `deleteMany` predicate cannot carry this term, so it is a genuinely weaker guarantee than the other terms have. The repo already does this same lookup for the same reason in `resetOnsiteListingToPending`.

The UI half was blind for the identical reason — `hasPendingRequest` derives from the same wrong relation. Added `hasPendingBlockRequest` to `ModerationListingRow`, hydrated by one batched slug-keyed query over the page's on-site rows, and moved the purge branch onto it.

| # | Round-2 finding | Status |
|---|---|---|
| 🔴 F1 | "Under review" guard inert; purge reachable on a live submission | **Fixed** — slug-keyed, both tiers, on server and UI |
| 🟡 F2 | Three invariants in `appListingModerationTableView.ts` falsified by round 1 | **Fixed** — incl. *"purge is OFF-SITE ONLY (the service raises NOT_FOUND for an on-site listing)"*, which would have told a maintainer no on-site guard was needed |
| 🟡 F3 | Purge notification omits `listingId`, alone among six call sites | **Fixed** — and it matters most here, since this delete invalidates the slug it would otherwise be identified by |
| 🟢 F4 | Confirm copy never mentions the slug being released | **Fixed** |
| 🟢 F5 | Emit-union / processor-key seam unpinned | **Fixed** — both directions, with a positive control on the guard's own parse |

**Per-claim disposition from round 2:** claims 1, 4, 5, 6 *actually fixed*; claim 2 *partially* (offered set was wider than intended → F1); claim 3 **not fixed** (the tautology). The audit also confirmed round 1's `revisionOfId` reasoning holds, and that `listAllListingsForModeration` does filter shadows.

## Verification

- Mutation-tested the **corrected** guard, since round 1's result was measured against an impossible state: never firing kills both refusals; the wrong `status` in the lookup kills the replica case; dropping the in-tx re-check kills the primary case. Each dies to its own case.
- 🔴 **One mutant failed to APPLY** — its anchor matched twice, because of a pre-existing identical lookup elsewhere in the file — and the run reported `16 passed`. That is the instrument not running, not a surviving mutant; re-run with a unique anchor, it dies. Recording it because a "passed" line from a mutation that never landed is indistinguishable from a real survivor unless you check the apply step.
- Full unit tier **1601 files / 25064 tests pass**. Typecheck 0 errors. Prettier clean, checked against a real `origin/main` worktree.

## 🔴 Still NOT verified

- **The browser tier has still never run**, and the diff now touches three component files. CI is the only thing that will exercise them.
- **No live database.** Every SQL-behaviour claim, including the new slug-keyed guard, is derived from the Prisma predicate plus the schema.
- **The moderator purge flow has not been driven in a browser.** The action set is unit-tested; nobody has clicked it.
- The new batched `pendingBlockRequestSlugs` query has not been measured against a realistic page.
- Long-horizon media-scan/retention for an indefinitely-`draft` listing remains unexamined.

```audit-claims round=2 audited=d0646267b7..c641ab8bb9
1. The on-site "not under review" rule is enforced by a slug-keyed read against AppBlockPublishRequest (assertNoLivePendingSubmission), on the replica for fail-fast and again on the primary inside the tx; the inert AppListing.publishRequests term and pendingRequestCount are removed.
2. ModerationListingRow carries hasPendingBlockRequest, hydrated by one batched slug-keyed query over the page's on-site rows, and listingModActions gates purge on it instead of hasPendingRequest.
3. The vacuous fixtures are replaced: the service test stages the signal on AppBlockPublishRequest and asserts the exact table+predicate queried; the UI test proves gating on the wrong field would let the hazardous case through.
4. The three falsified invariants in appListingModerationTableView.ts's header and action summary are corrected.
5. purgeListing's notification details carry listingId; the destructive-confirm copy states the store address is released and becomes claimable.
6. A new seam test pins AppListingOwnerNotificationType against the appListingNotifications key set in both directions, with a positive control on its own parse.
```

ZacxDev:
## Round 3: the service guard held; two things one layer up could not see their own failure

The delta audit confirmed round 2's fix is correct, correctly keyed, reachable on both tiers and mutation-verified — it did **not** repeat the wrong-table error in a new shape at the service layer. What it caught is the same *class* one level up. Fixed in `33eabdd223`.

**1. 🟡 `hasPendingBlockRequest` shipped with zero coverage — a survived mutant.** Hardwiring the projection to `false` *and* breaking the batched query's status filter left **315 files / 7209 tests green** (verified against a positive control on the same file). The branch was unreachable by construction: every fixture in the moderation-read suite was off-site, so `onsiteSlugs` was empty and the call short-circuited before touching the missing `appBlockPublishRequest` mock. That is round 2's lesson displaced one layer — *a fixture that never reaches the code proves nothing about it*. Added the mock and four cases; the previously-surviving mutant now dies to three of them.

**2. 🟡 The seam test I wrote last round had a hole its own positive control could not see.** The union parse terminated on the first `;` — including one inside a `//` comment, and the union already carries multi-line commentary. Measured: a `;` in a trailing comment before a newly appended, processor-less member truncated the parse to drop exactly that member, and **all four tests passed**, control included. It read as coverage for the likeliest failure while providing none. Comments are now stripped before matching, plus a two-sided **count** assertion that a truncation cannot satisfy. Verified by reproducing the auditor's exact scenario — it now fails two tests. Also removed a dead exclusion for `new-app-listing-comment`, which is registered in a different file and so was never in this set; a dead exclusion is a permanent hole for whichever key it names.

**3. 🟡 The live-submission refusal threw a generic `NOT_FOUND`,** justified by an anti-probing rationale *this PR itself falsifies* — the surface is moderator-only and the mod table now ships that exact fact to that exact audience. It left a moderator at "not found" for a row visibly on screen, reachable from ordinary replica lag. Now `NOT_TRANSITIONABLE` with an actionable message.

**4. 🟢 `?? true` on the DTO→UI hand-off** — `!undefined` is truthy, i.e. the permissive direction on the input to a destructive affordance. Absent now means "assume under review".

**5. 🟢 The second live purge-confirm surface** (`OffsiteReviewQueue`, reached from the report queue) still carried the old copy, so two moderator surfaces disagreed about what Purge does.

**Per-claim disposition from round 3:** claims 1, 3, 4 *actually fixed*; claims 2, 5, 6 *partially* (the gaps above).

### 🔴 On the disclosed race (F6) — not closed, and I did not apply the suggested fix

The audit rated it 🟢 and suggested a slug-scoped `pg_advisory_xact_lock`. **That does not work as a one-sided change**, and I've recorded why in the code rather than shipping it: a lock is mutual exclusion only between parties that both acquire it, and `submitVersion` takes none. Adding it to the purge side alone would read like a fix and change nothing — which is precisely the failure mode this guard already had once. Closing it properly means touching the submit path, which is a larger change than this PR should carry.

### Verification

- The mutant that survived round 3 now dies to three tests; the F2 truncation scenario now fails two.
- Full unit tier **1601 files / 25069 tests pass**. Typecheck 0 errors.
- Prettier checked against an `origin/main` worktree first: `OffsiteReviewQueue.tsx` was clean at base and is reformatted; the moderation test file was already unclean at base and is left alone rather than swept.

### 🔴 Still NOT verified

- **The browser tier has never run**, and the diff now touches four component files.
- **No live database** — every SQL claim, including the slug-keyed guard and the new batched query, is reasoned from the Prisma predicate plus the schema. The F6 race's production reachability is likewise reasoned, not measured.
- **Nobody has clicked Purge in a browser.**
- The new batched `pendingBlockRequestSlugs` query has not been measured against a realistic page.
- `moderationStatusWhere('draft')` still carries the same structurally-inert relation term in the mod-table *filter* — pre-existing, outside this PR's range, and harmless (meaningful for off-site, merely inert for on-site). Flagging it as known rather than silently leaving it.

```audit-claims round=3 audited=c641ab8bb9..33eabdd223
1. The on-site pending-block-request signal has tests that can reach it: the moderation-read suite now mocks appBlockPublishRequest and asserts the flag per row, the exact table/key/status queried, that no query is issued for an all-off-site page, and that an off-site row is never flagged even on a colliding slug.
2. The seam test strips comments before parsing the union and adds a two-sided count assertion, so a parse truncated by a semicolon inside a comment fails; the dead new-app-listing-comment exclusion is removed.
3. assertNoLivePendingSubmission throws NOT_TRANSITIONABLE with an actionable message instead of the generic NOT_FOUND, and the purge tests assert the code and the message.
4. AppListingsModerationTable normalizes hasPendingBlockRequest with `?? true`, so an absent field withholds the destructive action.
5. OffsiteReviewQueue's purge confirm states the store address is released, matching the mod table's copy.
6. The residual submit/purge race is documented as open, with the reason a one-sided advisory lock does not close it.
```

ZacxDev:
## Round 4: no production defects — two guards that nothing could kill

The delta audit found **no 🔴 and no behaviour defects**. It re-verified every self-reported number in the last commit (typecheck, test counts, "fails two tests", "dies to three") and independently confirmed the F6 race disclosure by reading the submit path rather than accepting my reasoning: `submitVersion` takes no lock of any kind, and `one_pending_per_slug` excludes a second pending *request*, not a concurrent *delete*. The disclosure stands as written.

What it did catch is the same class as rounds 2 and 3, one layer out again. Fixed in `c479f85c35`.

**1. 🟡 The two-sided COUNT assertion I added last round cannot catch a truncated parse — and my comment claimed it could.** I verified the argument myself before acting: the two `⊆` tests already force `set(parsed) === set(processors)`, so comparing sizes can only fail on a **duplicate**; a truncation removes members and never creates one. Measured with the stripping reverted, round 3's exact scenario left **all five tests green, that assertion included**. So the comment-stripping *is* the guard, and it was pinned by nothing — my earlier "verified by reproducing the scenario" was really verifying the stripping, not the assertion I credited.

`parseUnionFrom` now takes source text and has its own tests against synthetic sources whose answer is known without reading the file under audit: a `;` inside a line comment, one inside a block comment, a `//` inside an unrelated string, and a missing declaration that must **throw** rather than return `[]` (an empty return would make every subset check vacuously true). Reverting the stripping now fails two of them. The count assertion is retitled to what it does.

**2. 🟡 `?? true` had zero coverage in any tier.** Inverting it to the permissive direction — the exact thing it was added to prevent — left **1601 files / 25,069 tests green**. Unreachable by construction: the browser tier is the only suite that renders that component, it has no on-site orphan-draft fixture, and it cannot run on this host at all.

Rather than write a test that cannot execute here, I moved the DTO mapping and its defaults out of the `.tsx` into `listingModActionsForRow` — a pure function the unit tier reaches — and covered absent / null / true / false plus the `pendingRequest` derivation. **The surviving mutant now dies to two tests.**

Both fixes were verified by **re-running the exact mutation that survived**, not by assuming the new tests cover it.

**Per-claim disposition from round 4:** claims 1, 3, 5, 6 *actually fixed*; claim 2 *partially* (the count assertion); claim 4 *fixed in behaviour but shipped untested*.

### Where the ladder stands

| round | payload lines its fixes changed | production defects found |
|---|---|---|
| 1 | 197 | 1 🔴 (CI ratchet), 1 🟡 (purge unreachable) |
| 2 | 203 | **1 🔴 — inert guard, data-loss path** |
| 3 | 36 | 0 (1 survived mutant, 1 blind control) |
| 4 | 39 | **0** (2 coverage defects) |

Rounds 3 and 4 found no production defects. The payload has been stable and audited for two rounds; what keeps being found is coverage *about* the coverage. Round 5 is dispatched because round 4 did produce findings that needed fixing — but if it comes back clean, or finds only scaffolding-about-scaffolding, that is the end of the ladder.

### 🔴 Still NOT verified

- **The browser tier has never run.** Round 4's F2 half rests on static fixture inspection. Note this round *reduced* the exposure — the normalization moved out of the untestable component into a unit-tested module — but the component itself is still only exercised by CI.
- **No live database.** Every SQL claim is reasoned from the Prisma predicate plus the schema.
- **Nobody has clicked Purge in a browser.**
- The residual submit/purge race is open and documented, with the reason a one-sided advisory lock does not close it.
- One test-suite flake observed by the auditor (`eventloop-watchdog.capture.test.ts`) was confirmed unrelated — outside the range, green on re-run.

```audit-claims round=4 audited=33eabdd223..c479f85c35
1. parseUnionFrom takes source text and is tested directly against synthetic sources: a semicolon in a line comment, one in a block comment, a // inside an unrelated string, and a missing declaration that throws rather than returning an empty array.
2. The size assertion is retitled to what it actually catches (duplicate union members) and its comment no longer claims it catches truncation.
3. The DTO mapping and its defaults moved from AppListingsModerationTable.tsx into listingModActionsForRow, a pure function in the view module.
4. listingModActionsForRow is covered for hasPendingBlockRequest absent, null, true and false, plus the pendingRequest-object derivation of hasPendingRequest.
5. Both previously-surviving mutants were re-run against the new tests and now die: `?? true` to `?? false` kills two, reverting the comment-stripping kills two.
```

ZacxDev:
## Round 5: clean — zero findings. The audit ladder ends here.

Five rounds, all dispositioned. Round 5 returned **no findings at any severity**, which is the stop condition: a clean round ends the ladder and must not be re-run to confirm itself.

It re-verified every claim from round 4 by **re-running the mutations rather than accepting them**, and added controls I had not: it planted a duplicate union member and confirmed only the retitled duplicate-detection test fires; it over-stripped the comment regex and confirmed the string test catches that too, so that test is not decorative; and it reproduced round 3's truncation scenario one more time to independently confirm the reasoning that the size assertion cannot see it.

It also confirmed the refactor is behaviour-preserving field-for-field, and that **no production caller bypasses the safe default** — `listingModActions` is still exported, but its only non-test caller is `listingModActionsForRow`.

**Core claim re-verified end to end:** `rejectRequest` does not call `deleteOnsiteDraftListingForSlug` (the only call site is `withdrawRequest`); a re-submit lands back on the same listing via the owner-scoped reuse branch, with a foreign-owned orphan treated as TAKEN; and the only deleters of an on-site draft are withdraw and a mod purge that refuses while a submission is live.

## CI is green

**18 pass, 1 skipped, 0 non-terminal of 19 registered.** `MERGEABLE` / `CLEAN` at `c479f85c35`.

### 🔴 The browser-tier caveat is now partly resolved — with a correction to what I said earlier

Every previous comment said the browser tier had never run. **It has: `preview / component-tests` passed.** I checked what that status actually means before claiming it, because the task is documented as "always exits 0" and I first read that as making the green meaningless. It doesn't: the suite's verdict is written to a task result and surfaced through `suite_verdict` — *success on pass, failure on a real red verdict, error on an infra condition*. So a green there means the vitest browser-mode suite genuinely ran and passed on this head.

The correct qualifier is that it is **report-only and non-blocking** — a red would not have stopped a merge, and nobody gates on it. So: the component suite ran and passed, and no human or gate is watching that signal.

## Full ledger

| round | payload lines its fixes changed | production defects found |
|---|---|---|
| 1 | 197 | 1 🔴 (CI ratchet), 1 🟡 (purge unreachable from any UI) |
| 2 | 203 | **1 🔴 — guard read the wrong table, was inert, opened a data-loss path** |
| 3 | 36 | 0 (a survived mutant, a blind control) |
| 4 | 39 | 0 (two guards nothing could kill) |
| 5 | — | **0 — clean** |

The two findings that mattered were both **defects this PR itself introduced and I had reported as fixed**: the purge arm with no operator surface (round 1), and the inert "under review" guard whose test asserted a database state that cannot exist (round 2). Neither was found by writing more tests; both were found by asking whether the test could reach the code.

## 🔴 Open, and deliberately not closed here

1. **Half (a) of the original ticket is untouched** — should the review path have a `changes-requested` outcome at all? That is Ellie/Justin's call. This PR is only half (b).
2. **The off-site reject still deletes its draft.** Every word of this PR's argument applies to it verbatim; it is scope, not a justification. Documented at the call site as the obvious follow-up.
3. **A residual submit/purge race is open and documented.** I did not apply the suggested `pg_advisory_xact_lock`: a lock excludes only between parties that both take it, and `submitVersion` takes none — round 4 confirmed that independently by reading the submit path. Closing it properly means touching the submit path.
4. **No live database, and nobody has clicked Purge in a browser.** Every SQL claim is reasoned from the Prisma predicate plus the schema.
5. `moderationStatusWhere('draft')` still carries the same structurally-inert relation term in the mod-table *filter* — pre-existing, outside this PR, harmless (meaningful off-site, merely inert on-site). Flagged as known.

PR #4493, opened 2026-08-29
feat(apps): moderator actions in the store listing detail overflow menu

ZacxDev:
Adds a divided **Moderator** section to the `⋮` overflow menu on the mod-only store listing detail page (`/apps/store-preview/<slug>`), which until now held only Edit / Leave a review / Report.

## What it adds

| Item | Wired to |
|---|---|
| **Contact app owner** | the existing `MessageAppOwnerModal` + `appListings.messageAppOwner` — reused, not forked |
| **Unpublish and send back to review** | a reason-gated confirm → `resetListingToPending` (off-site) / `resetOnsiteListingToPending` (on-site), routed by kind |
| **Hide from store (reversible)** | the same confirm → `delistListing` — **one** proc for both kinds, deliberately *not* kind-branched on the client |
| **Manage in review queue** | a link to `/apps/review?tab=manage` |

## Two takedowns, because they are different tools

`delistListing` and `relistListing` are **dual-kind** — `delistListing` suspends the backing app block for an on-site listing in the same transaction, and `relistListing` restores it. The router's *"All offsite-only"* block comment said otherwise; it was stale prose from an earlier PR and **is corrected in this PR** (see "Stale comments corrected" below), and `appListingModerationTableView` already recorded the right behaviour. So `hide` is available here, and it is the genuinely reversible takedown that `reset-to-pending` is not.

Both items are offered because they have the **same immediate effect** and **very different undo costs**:

|  | off the store | app stops serving (on-site) | way back |
|---|---|---|---|
| **Hide** | ✔ | ✔ | a moderator's **Relist**, one click, unchanged |
| **Unpublish → review** | ✔ | ✔ | only when someone **approves** the queued request |

That distinction is carried where a moderator reads it, not in a runbook: in the two labels, and in a per-action × per-kind consequence paragraph in which **each arm names the other action by its exact menu label**. All four cells are pinned as whole normalised strings, plus *difference* properties — only the on-site arms claim a runtime stop, only the reset arms claim a re-review — which a reword cannot satisfy by accident.

The copy is written from what the server actually does. The reason is required because every proc here requires it, writes it into its audit event, **and delivers it to the owner** — it is the explanation the developer receives, not a confirmation ritual.

The two confirms are **one component parameterised by action** (`ListingTakedownModal`), mirroring the management table's single shared reason-gated modal. Each owns a distinct `data-testid` stem, single-sourced so a menu item and its confirm cannot disagree — otherwise a test could open one, assert against the other, and pass, since both render the same shell.

### No `reportId` affordance

`delistListingSchema` accepts an optional `reportId` and resolves that report in the same transaction, but **this surface has none in hand**: `ListingDetail` carries no report field of any kind and `/apps/store-preview/<slug>` fetches only `getAppDetail`. Inventing a report picker here would be building a second reports surface inside a store page; the report queue is where a report-driven delist belongs, and it already passes the id.

### The `removed → removed` "enforced takedown" branch

`delistListing`'s guard is `status IN ('approved','removed')`, and the `removed → removed` arm exists to convert an owner self-hide into an enforced takedown (it always writes a `delist` event, so `republishOwnListing`'s guard then refuses the owner). **From this surface that arm is unreachable by construction** — the detail read is approved-only, so the row we send is `approved` and the event records `before: {status:'approved'}`.

One exception, documented rather than hidden: `unpublishOwnListing` is itself dual-kind, so an owner self-unpublishing between page load and confirm leaves the mod's click landing on the `removed → removed` arm.

⚠️ **An earlier revision of this section called that "a narrow race" that "only ever hardens the takedown" and said "the button's meaning does not change". Both claims were wrong and are retracted.** The window is not narrow: this route's query inherits `staleTime: Infinity` and does not override it, so "approved by construction" is a property of the READ, not of the CLICK, and the gap is unbounded. And the meaning does change: against an owner-unpublished listing, `hide` converts a *reversible owner self-hide* into an *enforced lock-down* that forbids the owner from self-restoring — and the consequence copy shown at that moment is affirmatively false, because it describes taking a live app down. This is being tracked as a separate finding rather than patched here.

## Why there is no Relist/Republish button

This is the one real design question and it resolves to "the button cannot exist here", not "we skipped it".

`relistListing` acts on a **`removed`** listing. This page's read, `appListings.getAppDetail`, is **approved-only** — it returns `null` for any other status, so a removed listing 404s before the body ever mounts. A Relist item on this surface could therefore only ever be rendered against a listing the server would refuse. That is the same reasoning `listingPublishingActions.ts` already applies to an owner's Republish on a mod-removed listing ("rendering Republish there would be a button that can only fail").

The honest affordance instead is a link to the surface that **can** see a removed listing: the `/apps/review` management table, which already carries **Relist** (dual-kind), Claim and Purge for exactly that state. So the door is not one-way — the way back just lives where the state it applies to is visible.

## Why nothing renders in `preview` mode

Beyond the existing read-only preview posture, there is a second and independent reason, which is what settled it: in preview `detail.id` **is not guaranteed to be an `AppListing` id**. `buildListingDetailPreview` — the graceful fallback the review modal renders while the projection loads, errors, or has no backing listing id — sets `id: row.appListingId ?? row.id`, so on that path it is the publish **request**'s id and every one of these procs would answer `NOT_FOUND`. Shipping a moderator control that works on one path and fails on the other is the same hazard as the Relist button, in a weaker form.

## The action set is derived, not hand-rolled

`appListingDetailModActions` (new, pure, React-free) intersects the **shared** lifecycle state machine — `listingModActions`, which the `/apps/review` management table already depends on — with the subset this surface implements. Both directions fail closed: if the state machine stops offering `reset-to-pending` on an approved listing the item disappears with no edit here, and a member newly added to `ListingModAction` is not silently adopted by this menu.

`hide` and `relist` are the two halves of one dual-kind pair, which makes them the live control on that claim: adding `hide` to the subset did **not** bring `relist` with it, because the machine does not admit `relist` on an `approved` listing. If the subset were the *source* of the menu rather than a filter, it would have.

## A vacuous guard found and replaced

The browser-tier test whose comment claimed to pin *"the modals are mounted outside `Menu.Dropdown`, because the dropdown unmounts on close"* **does not pin it**. Measured on this change: moving `<ReviewListingModal>` inside `<Menu.Dropdown>` — the exact defect the rule forbids — left the whole browser suite **green at 86/86**, and so did the same move for the new `<MessageAppOwnerModal>`. Mantine keeps the dropdown's subtree mounted through its close transition and the portalled modal opens synchronously, so the assertion resolves inside that window and cannot tell the two placements apart.

So the rule is now pinned **structurally in the blocking node project** (`__tests__/appListingDetailModalPlacement.test.ts`), over all four modals the menu opens, with a positive control on its span extractor and a negative control on synthetic source. Both of the mutants above are red against it. The two comments that made the false claim are corrected in place rather than left to be cited as coverage.

## Stale comments corrected (comment-only, no behaviour change)

Three summaries said the opposite of a property this PR now depends on — the dangerous shape, because the client deliberately does **not** kind-branch `hide`, so a maintainer trusting them would "fix" that by adding a kind guard and reach for a second on-site delist proc that does not exist.

- the mod-actions block header ended **"All offsite-only"** directly above `delistListing`. `delistListing`/`relistListing` resolve through `classifyListingForAction`, which applies **no** kind filter; `claimListing`/`purgeListing` genuinely are off-site only, which is what that sentence was describing when it was written.
- `resetOnsiteListingToPending` was still marked **"DARK backend capability — no UI wiring yet"** — this PR is the change that falsifies it. It now names both callers.
- the post-approval block said the owner procs are **"offsite-only"**. `unpublishOwnListing`'s *own* header already reads "DUAL-KIND (W13 P4)" and describes the full on-site takedown, so only the summary was stale — the shape that survives review, because the detail is right and the heading everyone reads first is wrong. `resetListingToPending` genuinely *is* off-site only.

Two further corrections after a delta re-audit, both the same shape:

- the kind-scope note above was **itself over-narrow**. It said the retired "All offsite-only" was false for *two* of the six procs the block enumerates — which asserts the other four are off-site only. It is false for **four**: `resolveReport`/`dismissReport` route through `closeReport`, which looks the report up **by report id with no kind filter**, and upstream `reportListing` gates only on the listing being `approved` while `listListingReports` filters only on report status. An on-site listing is reportable and its report is closeable. Only `claimListing`/`purgeListing` are genuinely off-site only. Now written **per proc** rather than as a count, with a note not to re-summarise it as a bare number — this was the exact failure the correction exists to prevent, re-committed one revision later and *asserted* rather than inherited.
- the post-approval block heading still read **"— DARK"**, 50 lines above the per-proc "NO LONGER DARK" note, and it is false for **all five** procs in that block — every one has a live client caller.

One more found while there and corrected in passing: `classifyOffsiteListing`'s docstring claimed to guard "delist/relist/purge", but its real callers are `claimListing`, `purgeListing` and the off-site `resetListingToPending` — naming a caller that does not call it is how the blanket "offsite-only" reading survived.

## Verification

| Command | Result |
|---|---|
| `pnpm typecheck` | **OK — 0 type errors** |
| `npx vitest run --project unit src/components/Apps` | **69 files / 1433 tests passed** |
| `npx vitest run --project unit src/server/routers/__tests__ src/server/services/blocks/__tests__` | **215 files / 4889 tests passed** (covers the comment-only server files) |
| `npx vitest run --project component src/components/Apps/AppListingDetailBody.browser.test.tsx` | **1 file / 92 tests passed** (78 on `main`), green on 9 consecutive runs |
| `npx eslint <changed files>` | **0 problems** (positive control: the same invocation reports 3 on `src/utils/trpc.ts`) |

⚠️ An earlier revision of this table quoted **"284 files / 6320 tests"** without naming its command, and it was not reproducible from the description. It was the two `unit` scopes above passed to a **single** invocation — 69 + 215 = 284 files, 1431 + 4889 = 6320 tests. Split out and each command written in full, because a count nobody else can re-derive is not evidence.

The two `__tests__/*.test.ts` files are excluded from `tsconfig.json`, so they were typechecked separately via an ad-hoc project: **0 errors attributable to them**, validated with a planted `TS2322` (fires) and a control config without them (the one unrelated error appears either way).

Mutants watched to die, each with that guard's own assertion:

| Mutation | Killed by |
|---|---|
| shrink the `MessageAppOwnerModal` mount ledger back to one site | `the exact ledger of sites that mount MessageAppOwnerModal` |
| drop the `isModerator` guard in the derivation | `a NON-moderator … gets nothing` (×2 kinds) |
| `detailListingStatus` always returns `'approved'` | the preview cases (×3) |
| invert the kind branch in the consequence copy | the three copy cases |
| drop the surface-subset `.filter` | the ordered-set + intersection + subset cases (×4) |
| `isAppReviewer(currentUser)` → `true` in the component | `a signed-in NON-moderator gets NONE of them` + the pre-existing signed-out case |
| invert the kind routing in `UnpublishListingModal` | both proc-routing cases |
| move any menu modal inside `<Menu.Dropdown>` | the new structural placement guard (and **nothing** in the browser tier — see above) |
| drop `hide` from the surface subset | the ordered-set, `offers hide`, and takedown-pair cases |
| invert the action routing (`action !== 'hide'`) | all four proc-routing cases |
| kind-route `hide` so on-site goes to the reset proc | `HIDE calls delistListing … and NO reset proc` |
| ~~swap the two confirms' submit labels~~ | ⚠️ **RETRACTED — the kill was ONE-SIDED.** It fired only because a full swap moves the word `Unpublish` onto the hide button, which the hide-side assertion happens to read; a one-sided `takedownSubmitLabel('reset-to-pending') = 'Hide from store'` **survived both tiers**. Now pinned whole, both arms, in the blocking tier — see below. |
| swap the two testid stems | seven cases across the section |
| invert the action branch in the consequence copy | all four pinned strings + both difference properties |

### Round-2 mutants (the two an audit measured SURVIVING, plus a re-check of the rest)

The two string maps lived as module-private consts in a `.tsx` no blocking suite can import, so their only coverage was the report-only browser tier — and there only on the `hide` arm. They now live in the React-free module with all four strings pinned whole and normalised, **both arms**, in the blocking tier, plus difference properties that fail specifically on a swap.

| Mutation | Killed by (blocking tier unless noted) |
|---|---|
| swap the two `takedownSuccessMessage` values *(auditor: SURVIVED both tiers)* | `the success message, both arms, whole` **and** `each success message names its OWN undo path` |
| one-sided `takedownSubmitLabel('reset-to-pending') = 'Hide from store'` *(auditor: SURVIVED both tiers)* | `the submit VERB, both arms` **and** `each submit verb matches its own menu label` |
| revert the href to bare `/apps/review` | `the review-queue link carries tab=manage` (blocking) **and** the rendered-href assertion (browser) |
| one-sided stem collision (`hide` → the reset stem) | `each takedown owns a DISTINCT testid stem` (blocking) + 7 browser cases |
| one-sided routing: reset arm silently sends `delist` | ⚠️ **browser tier ONLY** — it survives the entire blocking `unit` project (measured: 69 files / 1432 tests green with the mutant in place) and dies only in the report-only component suite, which does not block a merge. The guarded defect is real: the moderator picks "send back to review", gets a plain hide, and the toast tells them to approve a request that was never created. Blocking coverage **is** achievable, but only by lifting the `action`/kind branch out of `ListingTakedownModal.tsx` into the React-free module — an executable production change, deliberately not made in a comment-and-scaffolding round. |
| one-sided label: reset label loses the word "review" | `the reset label names BOTH halves` **and** `each arm points at the OTHER action` |

### The truncation class, closed on both surfaces

Two assertions in this file used the same blind spelling — `toContain` with the value under test as the **needle**, which any *shortening* of that value still satisfies. Both are now stated in the direction a prefix cannot satisfy, and both were measured before and after:

| Assertion | Blind to | Now |
|---|---|---|
| `each arm points at the OTHER action by its exact menu label` | a menu label truncated to a prefix of itself — it shipped a dangling on-screen pointer while staying green, which is precisely what its own comment promised to catch | the quoted phrase is read **out of the copy** and matched against the label (equality, or `startsWith` + a parenthetical-tail check where only the stem is quoted) |
| `each submit verb matches its own menu label` | a submit verb truncated to a prefix (`'Unpublish'` → `'Unpub'` survived; it died only on the whole-string pin) | the label's **opening word** is read out of the label and compared for **equality** with the verb |

Each has a sibling sweep driving **every proper prefix** through the check, asserting all are refused, with the real value as the positive control — and, for the verbs, asserting that *both* tempting blind spellings (`label.includes(verb)` and `label.startsWith(verb)`) would have accepted every one of them. That is why the fix had to be an equality rather than a re-anchored prefix test.

Re-measured across all three arms, `1 failed / 33 passed` → **`3 failed / 32 passed`** for the truncation, the mirror arm (`'Hide'` → `'Hid'`), and growth (`'Unpublish now'`) alike.

🔴 **The pattern, not just the strings.** Two self-reported mutation claims on this PR overstated their coverage, both the same way: a mutant died on the arm a test happened to read, and that was recorded as "killed". Every case added this round asserts **both arms**, and the four one-sided mutants above were run specifically to check the remaining table rows for the same defect rather than assuming they held.

**Also fixed here:** an order-dependent test introduced by the first commit — it mounted a second body while a Modal was open and clicked through its overlay. It passed twice and then timed out at 15 s on the third run. Split to one render per arm; three consecutive full runs green since.

**Line delta, by round.** Round 1 (the `hide` item): executable **+32 / −13** production, **+44 / −1** tests, plus **+75 / −19** comment-only across `app-listings.router.ts` and `offsite-moderation.service.ts` with zero executable change in either. Round 2 (audit fixes): **zero** executable production lines — one comment-only server file (+33 / −9 comments) and one test file. Round 3 (this one): **zero** executable production lines — a single test file, +21 / −3 executable, +42 / −3 comment.

**Not verified:** this was not exercised against a running app or a real database. No proc, gate, notification or audit-event write was executed — tRPC is mocked throughout. "Wired, typechecks, and its guards have been watched to fail" is the claim; "verified working end to end" is not.


ZacxDev:
## Correction to this PR's description, and two findings deferred out of it

An adversarial audit of this branch refuted one claim in the PR body. Correcting it here rather than silently editing the description, since a reviewer may already have read the original.

### The PR body's race analysis is wrong in one direction

The body describes the `removed → removed` case as a narrow race between page load and confirm, whose outcome "only ever hardens the takedown" and where "the button's meaning does not change". Two parts of that do not hold:

**Scope.** `staleTime: Infinity` is set globally (`src/utils/trpc.ts:385`) with `refetchOnWindowFocus: false`, and the store-preview detail query does not override it. A mounted tab therefore never refetches, so the window is not load-to-confirm — it is unbounded, and it is moderator-to-moderator, not only owner-versus-moderator.

**Meaning.** Against a listing the owner has self-unpublished, the click does change meaning: `republishOwnListing` refuses when the last status-changing event is a moderator `delist`, so the action converts a reversible owner self-hide into an enforced lock-down. Neither the menu label nor the confirmation copy says so, and the copy's "The app stops serving immediately and its listing leaves the store" is false in that state — the app is already down.

The rest of the body's verification table reproduced exactly, and the base was green on both tiers at the merge-base.

### Two findings deferred, with closing conditions

Neither is deploy-blocking; both fail loudly or in the hardening direction, and both are pre-existing-shaped hazards this surface newly exposes rather than defects it introduces.

**1. The store-preview detail query inherits the global `staleTime: Infinity`.** "Approved by construction" is a true statement about the read, not about the click.
*Closes when:* the store-preview `getAppDetail` query no longer inherits the global `staleTime` (or the takedown mutations carry an expected-status precondition), **and** a test exists that fails when a stale approved listing reaches `delistListing`. Checked mechanically by that test passing in CI on a merged PR.

**2. `reset-to-pending` is offered in a state the server deterministically refuses.** An approved on-site listing with an open pending block publish request is routine — v1 keeps serving while the update sits in review — but `resetOnsiteListingToPending` throws `NOT_TRANSITIONABLE` there. The shared `listingModActions` derivation does not model that precondition.
*Closes when:* `listingModActions` models the open-pending-request precondition so the action is not offered in that state, **and** a unit test pins it. Checked mechanically by that test passing in CI on a merged PR.

The audit also found two string maps in the new confirmation modal unpinned in the blocking tier, and a link that does not land on the tab its own copy promises. Both are being fixed on this branch.

PR #4499, opened 2026-08-30
fix(blocks): the search_models tool can RANK — "most popular" was text-searching the word "popular"

ZacxDev:
**Reported from a block:** asking an AI chat block for "the most popular models" returns models literally **named** "Popular …". Traced end to end — the model was behaving correctly against a tool that could not express the question.

## Root cause

`search_models`' entire argument surface was `{ query (REQUIRED), type?, limit? }` on a `.strict()` object, and the route **hardcoded** the ranking:

```ts
sort:   constants.modelFilterDefaults.sort,  // Highest Rated
period: MetricTimeframe.AllTime,
```

So there was no way to ask for most-downloaded, and no way to omit the text filter. With exactly one lever available, the model used it — `query: "popular"` — which text-matches model **names**.

Measured 2026-08-30, the returned models had **2,168 / 1,910 / 224** downloads against a real top-of-catalog above **2,300,000**. Reproduce the wrong arm:

```
civitai models search --query "popular"     # → the exact models the block cited
civitai models search --sort "Most Downloaded"   # → what it should have said
```

🔴 **Not a hallucination and not a loop bug.** The model grounded every claim in what the tool returned, exactly as its system prompt asks. This is a capability gap, and a prompt change would not have fixed it.

## The fix has in-repo precedent

`/api/v1/blocks/models.ts` — the route a block calls **directly** — has always exposed `sort: z.enum(ModelSort).optional()` and an optional `query`, and `runModelSearch` already supports both. **Tool calling was strictly less capable than the REST surface it replaced.** This wires the same two things the same way:

- **`registry.ts`** — `query` becomes optional (still `.min(1)` when present, so an empty string cannot silently mean "no filter"); `sort` added as the **same** `z.enum(ModelSort)` `models.ts` uses. A curated subset here would be a second rule that can drift. The tool `description` now tells the model to omit `query` and pass `sort` for popularity questions.
- **`tools.ts`** — `sort ?? constants.modelFilterDefaults.sort`, the identical expression `models.ts` uses; and the Meili hop becomes conditional on `query`, because an empty `searchIds` from a search for *nothing* is not the same thing as *no text filter*.

**Verified rather than assumed** that this yields a sorted read: `model-search.service` gates both id paths on `query` — `ids: query ? searchIds ?? [] : queryIds` and the re-ordering at `:218` — so with `query` undefined the empty `searchIds` is never consulted.

**No new exposure.** The maturity clamp, rate limit, item cap and char budget are untouched, and a no-query catalog read is what `blocks/models.ts` already permits.

## Test matrix — watched, not assumed

**RED at `origin/main`**, green here:

| test | failure at base |
|---|---|
| `sort` REACHES `runModelSearch` | `400` — `.strict()` rejects the unknown key |
| omitting `query` skips Meilisearch | `400` — `query` was required |
| the DECLARATION advertises `sort` | `undefined` |

🔴 **Three of the six new tests PASSED at `origin/main`** and are labelled **CONTROLS** in the file rather than counted as regression coverage. One — "an unknown sort is rejected" — passes in **both** arms for *different* reasons (before: the key is unknown; after: the enum rejects the value), and that is stated at its site.

**Mutation sweep, each mutant dying for its own reason:**

| mutant | dies on |
|---|---|
| re-hardcode `sort` to the default | `sort` REACHES `runModelSearch` |
| make the Meili hop unconditional | omitting `query` skips Meilisearch — **and only that one** |

🔴 **`registry.test.ts`'s "rejects a missing query" arm was INVERTED, not deleted.** The required `query` *was* the defect, so omitting it must now parse. Everything else that test guarded is kept — an empty string is still rejected, the over-cap limit is untouched — and a new test pins the enum so widening `sort` to `z.string()` fails.

All **five** consumer test files green (**167** tests), found by grepping for files that *consume* the registry rather than only the ones edited. `pnpm typecheck`: **0 errors**.

## Follow-up, deliberately not in this PR

The block's own system prompt could also mention ranking, but the deterministic schema fix is the one that makes the question answerable at all — a prompt cannot invent a parameter that does not exist.

https://claude.ai/code/session_01SZz385g4YTRRoPDjX8Hg6e

### DIFF
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -8,7 +8,6 @@ import {
 } from '~/server/middleware/block-scope.middleware';
 import { handleEndpointError } from '~/server/utils/endpoint-helpers';
 import { getNextPage } from '~/server/utils/pagination-helpers';
-import { postgresSlugify } from '~/utils/string-helpers';
 import {
   ModelSearchMeiliTimeoutError,
   resolveModelSearchIds,
@@ -244,7 +243,7 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
 
   try {
     if (tool.name === 'search_models') {
-      const { query, types, baseModels, limit, sort, period, supportsGeneration, tag, username } =
+      const { query, types, baseModels, limit, sort, period, supportsGeneration } =
         args.data as {
           query?: string;
           types?: string[];
@@ -253,8 +252,6 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
           sort?: string;
           period?: string;
           supportsGeneration?: boolean;
-          tag?: string;
-          username?: string;
         };
       // The `Math.min` is redundant TODAY — `searchModelsArgs.limit` is already
       // `.max(MAX_TOOL_RESULT_ITEMS)`, so a larger value is rejected before it
@@ -268,15 +265,6 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       // `types` arrives as an array now, so there is nothing to wrap. The old
       // `type ? [type] : undefined` existed only because the schema exposed a
       // SINGULAR field over a hand-written six-value enum.
-      //
-      // 🔴 `username` IS SLUGIFIED HERE, NOT IN THE SCHEMA. The public schema
-      // applies `postgresSlugify` in its own `.transform`, and the catalog query
-      // matches on the slugified column — so passing a raw "Some User" through
-      // would return NOTHING and read to the model as "that creator has no
-      // models". It is done in the ROUTE rather than the registry because the
-      // registry is deliberately server-import-free (see its header), and this
-      // is the same split that file already uses for dispatch.
-      const slugifiedUsername = username ? postgresSlugify(username) : undefined;
 
       // 🔴 A TEXT QUERY SILENTLY DISCARDED `sort` UNTIL THIS EXISTED.
       // `runModelSearch` restores Meilisearch's relevance order whenever a
@@ -344,8 +332,6 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
           types: types as never,
           baseModels,
           supportsGeneration,
-          tag,
-          username: slugifiedUsername,
           // 🔴 THE MODEL'S CHOICE, FALLING BACK TO THE SAME DEFAULT THIS LINE
           // USED TO HARDCODE — the identical `sort ?? default` expression
           // `blocks/models.ts` uses. Hardcoding it here is what made "the most
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -8,6 +8,7 @@ import {
 } from '~/server/middleware/block-scope.middleware';
 import { handleEndpointError } from '~/server/utils/endpoint-helpers';
 import { getNextPage } from '~/server/utils/pagination-helpers';
+import { postgresSlugify } from '~/utils/string-helpers';
 import {
   ModelSearchMeiliTimeoutError,
   resolveModelSearchIds,
@@ -242,12 +243,18 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
 
   try {
     if (tool.name === 'search_models') {
-      const { query, type, limit, sort } = args.data as {
-        query?: string;
-        type?: string;
-        limit?: number;
-        sort?: string;
-      };
+      const { query, types, baseModels, limit, sort, period, supportsGeneration, tag, username } =
+        args.data as {
+          query?: string;
+          types?: string[];
+          baseModels?: string[];
+          limit?: number;
+          sort?: string;
+          period?: string;
+          supportsGeneration?: boolean;
+          tag?: string;
+          username?: string;
+        };
       // The `Math.min` is redundant TODAY — `searchModelsArgs.limit` is already
       // `.max(MAX_TOOL_RESULT_ITEMS)`, so a larger value is rejected before it
       // reaches here. It stays deliberately: it is the last bound before a value
@@ -257,7 +264,18 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       // if that schema ever relaxes is an unbounded catalog read. A redundant
       // clamp costs one comparison; removing a bound is the wrong direction.
       const effectiveLimit = Math.min(limit ?? DEFAULT_LIMIT, MAX_TOOL_RESULT_ITEMS);
-      const types = type ? [type] : undefined;
+      // `types` arrives as an array now, so there is nothing to wrap. The old
+      // `type ? [type] : undefined` existed only because the schema exposed a
+      // SINGULAR field over a hand-written six-value enum.
+      //
+      // 🔴 `username` IS SLUGIFIED HERE, NOT IN THE SCHEMA. The public schema
+      // applies `postgresSlugify` in its own `.transform`, and the catalog query
+      // matches on the slugified column — so passing a raw "Some User" through
+      // would return NOTHING and read to the model as "that creator has no
+      // models". It is done in the ROUTE rather than the registry because the
+      // registry is deliberately server-import-free (see its header), and this
+      // is the same split that file already uses for dispatch.
+      const slugifiedUsername = username ? postgresSlugify(username) : undefined;
 
       // 🔴 THE MEILI HOP IS NOW CONDITIONAL, mirroring `blocks/models.ts`. With
       // no `query` there is no text relevance to resolve, and running it anyway
@@ -303,13 +321,21 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       const { items } = await runModelSearch(
         {
           types: types as never,
+          baseModels,
+          supportsGeneration,
+          tag,
+          username: slugifiedUsername,
           // 🔴 THE MODEL'S CHOICE, FALLING BACK TO THE SAME DEFAULT THIS LINE
           // USED TO HARDCODE — the identical `sort ?? default` expression
           // `blocks/models.ts` uses. Hardcoding it here is what made "the most
           // popular models" unanswerable: the only lever left was `query`, so
           // the ranking word became the search term.
           sort: (sort ?? constants.modelFilterDefaults.sort) as never,
-          period: MetricTimeframe.AllTime,
+          // Same shape as `sort`, and for the same reason: a hardcoded AllTime
+          // makes "most downloaded THIS MONTH" unaskable. `blocks/models.ts`
+          // still hardcodes this one, so the tool is deliberately WIDER than the
+          // sibling here — matching the public schema, which has always had it.
+          period: (period ?? MetricTimeframe.AllTime) as never,
           limit: effectiveLimit,
           query,
           searchIds,
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -242,10 +242,11 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
 
   try {
     if (tool.name === 'search_models') {
-      const { query, type, limit } = args.data as {
-        query: string;
+      const { query, type, limit, sort } = args.data as {
+        query?: string;
         type?: string;
         limit?: number;
+        sort?: string;
       };
       // The `Math.min` is redundant TODAY — `searchModelsArgs.limit` is already
       // `.max(MAX_TOOL_RESULT_ITEMS)`, so a larger value is rejected before it
@@ -258,15 +259,23 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       const effectiveLimit = Math.min(limit ?? DEFAULT_LIMIT, MAX_TOOL_RESULT_ITEMS);
       const types = type ? [type] : undefined;
 
+      // 🔴 THE MEILI HOP IS NOW CONDITIONAL, mirroring `blocks/models.ts`. With
+      // no `query` there is no text relevance to resolve, and running it anyway
+      // would hand `runModelSearch` an EMPTY `searchIds` from a search for
+      // nothing rather than the "no text filter" it means. `searchIds: []` plus
+      // `query: undefined` is exactly the pure-sorted-read shape that route
+      // already relies on.
       let searchIds: number[] = [];
       try {
-        const meili = await resolveModelSearchIds({
-          query,
-          limit: effectiveLimit,
-          browsingLevel,
-          types,
-        });
-        searchIds = meili.searchIds;
+        if (query) {
+          const meili = await resolveModelSearchIds({
+            query,
+            limit: effectiveLimit,
+            browsingLevel,
+            types,
+          });
+          searchIds = meili.searchIds;
+        }
       } catch (e) {
         if (e instanceof ModelSearchMeiliTimeoutError) {
           res.setHeader('Retry-After', '2');
@@ -294,7 +303,12 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       const { items } = await runModelSearch(
         {
           types: types as never,
-          sort: constants.modelFilterDefaults.sort as never,
+          // 🔴 THE MODEL'S CHOICE, FALLING BACK TO THE SAME DEFAULT THIS LINE
+          // USED TO HARDCODE — the identical `sort ?? default` expression
+          // `blocks/models.ts` uses. Hardcoding it here is what made "the most
+          // popular models" unanswerable: the only lever left was `query`, so
+          // the ranking word became the search term.
+          sort: (sort ?? constants.modelFilterDefaults.sort) as never,
           period: MetricTimeframe.AllTime,
           limit: effectiveLimit,
           query,
--- a/src/pages/api/v1/blocks/tools.ts
+++ b/src/pages/api/v1/blocks/tools.ts
@@ -25,6 +25,7 @@ import {
   projectModelForTool,
   DEFAULT_TOOL_RESULT_ITEMS,
   MAX_TOOL_RESULT_ITEMS,
+  TOOL_SORTED_QUERY_CANDIDATES,
 } from '~/server/services/blocks/tools/registry';
 import {
   MAX_TOOL_NAME_CHARS,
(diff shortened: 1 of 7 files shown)


ZacxDev:
## Scope widened in a second commit — `cef4382ee8`

🔴 **The PR description above covers only the first commit (`d9bbc5e5b8`, the `sort` fix). Saying so here rather than silently rewriting the body, since it may already have been read.**

Fixing the ranking gap exposed that `search_models` was narrower than **both** the public `/api/v1/models` schema and its own sibling `/api/v1/blocks/models`. This commit closes that:

| added | why it mattered |
|---|---|
| `baseModels` | "the best SDXL checkpoint", "Illustrious LoRAs" were unanswerable |
| `types` (plural, real `ModelType`) | replaced a **singular** `type` over a hand-written **six-value** enum. Measured: a live answer cited a **`Wildcards`** model — a type the tool could *return* but not *ask for*. Also omitted `Poses`, `Workflows`, `Hypernetwork`, `Upscaler`, `MotionModule`, `Detection`, `LLM`, `DoRA`… |
| `period` | the route hardcoded `AllTime`, so "most downloaded **this month**" could not be asked |
| `supportsGeneration`, `tag`, `username` | plain public-API parity |
| `likes` (response) | see below |

**`likes` is a correctness fix for the sort work, not a nicety.** Without it the tool could rank by `Most Liked` while handing the model only a *download* count to explain the order with — so the answer would cite the wrong number or be unable to justify itself. A sort key the result cannot evidence is worse than no sort key.

**Verified the plumbing rather than trusting the types.** `RunModelSearchInput` is a `Partial<GetAllModelsOutput>`, so it *types* everything — but a field a DTO types is not a field the query uses. Confirmed each is destructured at `model.service.ts:372-391` and used (e.g. `baseModels` joins into an `IN (…)` at `:669`).

**`username` is slugified in the ROUTE**, because the catalog matches a slugified column — forwarding the raw string returns nothing and reads to the model as *"that creator has no models"* rather than as an error. It lives in the route because `registry.ts` is deliberately server-import-free.

**New bounds, because these go into the QUERY, not a response.** `MAX_TOOL_FILTER_ITEMS = 10`, `MAX_TOOL_FILTER_VALUE_CHARS = 64`. An unbounded array from an untrusted sandboxed iframe is an unbounded query — a cost failure, which is the kind that fails quietly.

### Deliberately NOT added
- **`nsfw` / `browsingLevel`** — the server-side clamp is the whole authority surface. Exposing either would let a sandboxed iframe widen its own ceiling.
- **`cursor` / pagination** — results are bounded by design (`MAX_TOOL_RESULT_ITEMS` + the char budget); paging interacts with that budget in a way that needs its own design, not a parameter.

### Test matrix, watched at `d9bbc5e5b8`
- **RED there:** `baseModels`/`supportsGeneration`/`tag` forwarding · the `Wildcards` type · `period` forwarding · `username` slugification · the projection field-set.
- **PASSED there ⇒ labelled CONTROLS in the files**, not counted as coverage: the `period` AllTime default, and the filter-bounds test — which 400s in *both* arms for different reasons (unknown **key** before, violated **bound** now).
- **Mutation sweep, each mutant dying isolated on its own test:** drop the slugify · re-hardcode `period` · swap `likes`↔`downloads`. That last is why the fixture's stats are pairwise distinct (42 vs 1,700,000) — coincident values could not see it.
- `registry.test.ts`'s exact-projection assertion was **updated, not relaxed**, and still asserts `favoriteCount` does not leak: adding one field is not licence to leak its neighbours.

5 consumer test files green (**173** tests); `pnpm typecheck` **0 errors**.

---

⚠️ **For the audit ladder:** round 1 was dispatched against **`d9bbc5e5b8`** and read only that tree. Its findings apply to the `sort` change; this commit is unaudited and belongs to the delta round.

ZacxDev:
## Audit round 1 — findings fixed in `abdafe92a1`

Round 1 was a full blind audit against `d9bbc5e5b8`. Every finding it raised is addressed below; **finding 1 is explicitly NOT fixed** and the reason is measured rather than argued.

**🟡 2 — the real one, and it was a genuine gap.** `runModelSearch` restores Meilisearch's *relevance* order whenever `query` is present, silently discarding `sort`. So this PR fixed *"the most popular models"* and **not** *"the most popular **anime** models"* — the same question, scoped, and the phrasing the widened schema most invites. Verified independently before fixing: `getModelsRaw` builds `orderBy` purely from `sort`, unconditional on `ids`, so the rows arrive ordered and the restore overwrites them. Fixed as an **opt-out** (`preserveRelevanceOrder`, defaulting to historical behaviour), so the public endpoint and `blocks/models` are untouched. The route also widens the Meili pull to `TOOL_SORTED_QUERY_CANDIDATES` (100) when both a query and an explicit sort are present — sorting the *wrong set* is still wrong.

🔴 **The guard that almost shipped inert.** A mutant making the service ignore the new flag entirely **survived the endpoint suite 89/89 green** — those tests mock `runModelSearch`, so they prove the flag is *forwarded* and nothing more. That is the whole fix reduced to an unused argument. Added `model-search.sort-vs-relevance.test.ts` with the DB order and relevance order deliberately **reversed** so the result can only match one; it kills that mutant with `expected [1,2,3] to deeply equal [3,2,1]`.

**🟢 3** — corrected in place and labelled as a correction. The comment claimed an unconditional Meili hop would produce a *wrong result set*; it would not (both id paths gate on `query`). What the condition saves is a pointless round-trip and its 503 exposure.

**🟢 4** — rather than contort the code to fit the sentence, the header now states the property it was actually protecting (**no Prisma on the load path**, which does still hold), and a new test walks the import list so it cannot rot again. The same claim in `chat-completion.step` is corrected too.

**🟢 5** — kept as a literal (importing `~/server/common/constants` would put a server module on that file's load path) with the link **asserted in a test**, the same trade the file already makes for `MAX_MESSAGE_CHARS`.

**🟢 6** — now asserts **set equality** against `Object.values(ModelSort)`, so a hand-listed subset fails. Exactly the ask.

**🟢 7** — added a green/red pair asserting the clamp on the **no-query** path.

### 🔴 Finding 1 — NOT fixed, and measured

The Turbopack chunk-hash collision is a **documented, unfixed, repo-wide** defect (`claudedocs/turbopack-chunk-hash-collision-2026-08-18.md`; upstream issue auto-closed untriaged). It is not caused by this diff.

I tested the doc's own remedy — "perturb the graph and it moves":

| run | sha | collision path |
|---|---|---|
| `b8vqv` | `d9bbc5e5b8` | `_1yhc173._.js` |
| `7z27t` | `cef4382ee8` (+3 new imports) | **`_1yhc173._.js` — identical** |

Two genuinely different module graphs, the **same** collision. So graph perturbation did not clear it here, and pushing arbitrary changes to chase it is not a strategy. The only lever that attacks the mechanism is `turbopackServerSideNestedAsyncChunking` (24,552 → 7,122 server chunks), which is a repo-wide config change that does not belong in a tool-schema PR. **This needs `main` to move, or that config decision taken separately.**

### Ledger

`round 1 · payload lines changed THIS round: 242` (`registry.ts` 133, `tools.ts` 78, `model-search.service.ts` 27, `chat-completion.step.ts` 4) · scaffolding 373 · well clear of the zero-payload gate.

Wide verification: **553 test files / 9,784 tests** green across `services` + `blocks`; `pnpm typecheck` 0 errors.

```audit-claims round=1 audited=d9bbc5e5b83cd38f19578c8d411ec29ed7069410..abdafe92a169302e299e123819443a2faa5d3fc7
1. sort no longer discarded when query is present — preserveRelevanceOrder opt-out in runModelSearch, passed only when query AND an explicit sort are both present
2. the Meili candidate pool is widened to TOOL_SORTED_QUERY_CANDIDATES (100) when a sort must win, so the sort applies to the matching set rather than to the top-N by relevance
3. a service-level test now pins the ordering behaviour itself; previously only the endpoint tests existed and they mock runModelSearch
4. the conditional-Meili comment's stated reason corrected — it wrongly claimed a wrong result set
5. the registry header's "server-import-free" claim replaced with the property it protects (no Prisma on the load path), plus a test walking the import list; same claim corrected in chat-completion.step
6. the declared default sort's link to constants.modelFilterDefaults.sort asserted in a test rather than imported
7. the sort enum pinned by set equality against Object.values(ModelSort)
8. clamp asserted on the no-query path with a red/green pair
9. NOT ADDRESSED: the Turbopack chunk-hash collision (finding 1) — pre-existing, repo-wide, and graph perturbation was measured NOT to move it
```

ZacxDev:
## Audit round 2 (delta) — findings fixed in `2523d64368`

Round 2 audited `d9bbc5e5b8..abdafe92a1` — the round-1 fixes **plus** the scope-widening commit no one had audited. It was not clean, and the headline finding is a regression **the round-1 fix itself introduced**. That is what a delta round is for.

**🔴 1 — a no-hit text search returned the WHOLE CATALOG. Fixed.** `getModelsRaw` adds its id predicate under `if (!!ids?.length)`, so an **empty** `searchIds` adds no filter and the query degrades to an unfiltered page. Until `preserveRelevanceOrder` existed the relevance restore hid this — mapping an empty array yields `[]`, so the zero-hit case was guarded **by accident**. Opting out removed the accident: `search_models({query:"zzzqqq", sort:"Most Downloaded"})` returned the site-wide top 10, i.e. a block answering *"the most downloaded zzzqqq models are…"* with fabricated results. **Strictly worse than the wrong order the flag exists to fix.**

Fixed in the **service**, not the route — "no text matches ⇒ no results" is a property of the search, not of one caller. Verified `if (!!ids?.length)` by reading `model.service` myself rather than trusting the report.

**🟡 2 — `tag` and `username` REMOVED rather than patched.** Both were mine, both from the unaudited commit, and both fail **open** — the one direction an LLM caller must never be handed:
- `tag` is pushed only `if (tagId)`, so an unknown tag adds no filter → site-wide top N returned as *"models tagged X"*.
- `username` matches a **slugified** column and `postgresSlugify` strips everything outside `[a-zA-Z0-9_]`, so a non-ASCII or punctuation-only name slugifies to `""` and the filter is dropped. A non-empty-but-unknown username *throws* — so the dangerous direction is the quiet one.

A human typing a wrong tag sees an obviously unfiltered page; a model gets a plausible list and reports it as the answer. This also closes round 2's 🟢 finding (`username` disabling the per-version availability filter). Pinned by a test so a future "restore parity" edit has to confront it.

**🟡 3 — `period` described something it does not do. Fixed.** It does not compute the ranking over a timeframe: it adds a hard `lastVersionAt` filter while the `orderBy` ladder reads all-time counts and never references it, and the projected `downloads`/`likes` are all-time too. A model reading the old text would have reported all-time figures as "this month". The description now says what it does — narrows the set, does not re-rank it.

**🟡 4 — the top-level description contradicted the fix. Fixed.** It still told the model to **omit** `query` for a popularity question while the `sort` field said to combine them — and the headline case *"the most popular **anime** models"* **requires** the combination the top-level text forbade. Rewritten, and it now also warns that adding `sort` to a named-model lookup ranks 100 fuzzy matches and can push the exact model out.

**🟡 1 — the replacement header claim was also false. Fixed.** Round 1 flagged "server-import-free"; the round-1 fix replaced it with "walks the import graph", which the test does **not** do — it regex-extracts direct imports, so a side-effect import and a transitive one both pass. Rather than claim a third thing, the comment now describes the ratchet it actually is.

**Mutation sweep on the new guard**, each mutant dying isolated: deleting the empty-hits branch (the regression itself), and dropping its `Boolean(query)` gate (which would break every no-query sorted read).

### Ledger

`round 2 · payload lines changed THIS round: 101` (`registry.ts` 61, `model-search.service.ts` 24, `tools.ts` 16) · scaffolding 140 · non-zero, so the ladder is still on the PR rather than auditing itself.

186 test files / 4,567 tests green; `pnpm typecheck` 0 errors; eslint clean on the changed source files.

### Still not addressed
Finding 1 from round 1, the Turbopack chunk-hash collision — unchanged, and round 2 explicitly did not re-verify it (it ran no build). Still blocks merge; needs `main` to move or the `turbopackServerSideNestedAsyncChunking` decision taken separately.

```audit-claims round=2 audited=abdafe92a169302e299e123819443a2faa5d3fc7..2523d643681ee6dbc0b2ba4b7bb1c85a5eb95e7e
1. a text query with zero Meilisearch hits now returns [] regardless of preserveRelevanceOrder — the empty case is its own branch ahead of the flag, in the service
2. tag and username removed from the tool schema entirely, because both fail open in the shared catalog service; a test pins that they are rejected
3. period's description corrected to say it narrows by lastVersionAt and does NOT re-rank; ranking and returned counts remain all-time
4. the top-level tool description no longer tells the model to omit query for a popularity question, and warns that sort dilutes a named-model lookup
5. the registry header no longer claims the test walks the import graph; it describes the direct-import ratchet it actually is
6. NOT ADDRESSED: the Turbopack chunk-hash collision, unchanged and unverified this round
```

ZacxDev:
## Audit round 3 (delta) — findings fixed in `95ab2f9bc5`

⚠️ **Correction first, because it is machine-read.** My round-2 comment's `audit-claims` block cited head `2523d643681ee6dbc0b2ba4b7bb1c85a5eb95e7e`. **That object does not exist** — I fabricated the tail of the sha. The real head was `2523d64368044c93894bfb075487a22466f1c09b`. Round 3 caught it because `cat-file -e` returned rc 1. A claims block is read by the *next* round to compute its range, so a wrong sha there points the next audit at nothing. The block below carries verified shas.

---

**🔴 1 — the tool DESCRIPTION still named `tag` and `username` after the schema dropped them. Fixed.** Round 2 removed both fields and pinned their rejection, but left them inside the description string it otherwise rewrote: *"Filter with `types`, `baseModels`, `tag`, `username` or `supportsGeneration`"*. **The description is what the LLM reads.** A user asking *"most downloaded anime models by Lykon"* gets either a hard 400 from `.strict()`, or — on a provider that enforces `additionalProperties` — the creator name folded into `query`, returning models **named** Lykon. That is the exact fabrication class this PR exists to eliminate, reintroduced through the highest-consequence channel.

**🔴 The guard matters more than the two-word fix.** This class has now recurred **four times**, and careful prose did not hold:

| round | false claim |
|---|---|
| 1 | `registry.ts` header: "server-import-free" |
| 2 | its replacement: "walks the import graph" |
| 2 | `period`: "timeframe the ranking is computed over" |
| 3 | the tool description naming removed fields |

`parameters` is **derived** from the schema so it cannot drift; the description is hand-written, so it can — and nothing checked it. Added a mechanical check: every backticked bare identifier in a tool description (or any field description) must be a real parameter. **Verified by reinstating the actual defect** —

```
AssertionError: search_models: description cites `tag`, which is not a parameter
(schema has: query, sort, types, baseModels, period, supportsGeneration, limit)
```

— with a positive control asserting the extractor finds identifiers at all, since a zero-length scan would pass on anything.

**🟢 3 — the one fail-open shape the empty-hits branch missed. Fixed.** `Array.isArray(searchIds) && length === 0` → `!searchIds?.length`, which also covers `undefined`; both degrade identically at `ids: query ? searchIds ?? []`. No caller passes `undefined` today, so this closes an unreachable shape rather than a live bug — the type permits it and it costs one character. Re-mutated after the rewrite; still killed by its own test.

**🟢 2 — comment rot from the removals. Fixed.** All of it was advertising coverage that no longer exists: the bounds test still cited `tag`, and its red-then-green matrix still listed *"tag forwarding"* and *"username slugification"* as evidence when both tests were deleted in the previous commit. The parity header now states that parity is deliberately **not** total; the `MAX_TOOL_FILTER_*` docstring no longer claims to bound `tag`/`username`.

### Ledger

`round 3 · payload lines changed THIS round: 20` (`registry.ts` 11, `model-search.service.ts` 9) · scaffolding 91. Non-zero, so the ladder is still on the PR — but the trend is 242 → 101 → **20**, and two consecutive zero-payload rounds would end it.

186 test files / 4,569 tests green; `pnpm typecheck` 0 errors, instrument validated (an injected error was shown to produce a red count).

### Still not addressed
The Turbopack chunk-hash collision. Rounds 2 and 3 both explicitly declined to re-verify it and ran no build. **Still the only thing blocking merge** — it needs `main` to move, or the `turbopackServerSideNestedAsyncChunking` decision taken separately.

```audit-claims round=3 audited=2523d64368044c93894bfb075487a22466f1c09b..95ab2f9bc5344da5d22c3fdab3ffaa8534f1103c
1. the tool description no longer names tag or username, which the schema rejects
2. a new mechanical guard asserts every backticked identifier in any tool/field description is a real parameter, with a positive control on the extractor
3. the empty-hits branch now uses !searchIds?.length, covering undefined as well as []
4. comment rot removed: the bounds-test note, its red-then-green matrix, the parity describe header and the MAX_TOOL_FILTER_* docstring no longer cite tag/username or deleted tests
5. NOT ADDRESSED: the Turbopack chunk-hash collision, unverified in rounds 2 and 3
```

ZacxDev:
## Audit round 4 (delta) — findings fixed in `e7f989b7f3`

All three findings landed on the guard round 3 added, and two are that guard failing at the job it was written for.

**🟡 F2 — the "positive control" was a second sample, not a control. Fixed.** The extractor regex was written **twice** — once in the assertion, once in its own control. Break only the assertion's copy and the control stays green: measured **48/48** with the round-3 defect reinstated. So the thing written to make that defect impossible could be silently disarmed by any future edit. Hoisted to one module-level const used by both arms; the identical break now turns the **control** red.

**🟡 F1 — the guard was blind to the idiom the guarded text already uses. Fixed.** The extractor required the identifier to fill the whole backtick span, but the description already writes `` `query: "anime"` `` and `` `sort: "Most Downloaded"` ``. So an author extending it *in the style it already uses* — `` `username: "Lykon"` `` — sailed past. That exact mutation passed **92/92** before this change. Terminator widened to ``[`|:]``, zero false positives against current descriptions.

**🟡 F3 — the sentence still said `period` ranks. Fixed.** *"rank with `sort` and `period`"* is false: `getModelsRaw` builds its `orderBy` purely from all-time counts and never reads `period` (the period-aware ranking is commented out); `period` only narrows by `lastVersionAt`. The field's **own** description already says this correctly, so the served document handed the model two contradictory statements about one field. Now *"rank with `sort`; narrow by `period`"*.

**🔴 The limit worth writing down, because it is the class that keeps recurring:** the guard checks a cited parameter **EXISTS**, not that what the sentence **SAYS** about it is **TRUE**. `period` is a real parameter, so F3 would have passed it. Four rounds have each found a false prose claim; this guard closes only the narrower half. Recorded at the guard so nobody reads a green as "the prose is correct".

Mutations re-run after the fixes, each dying on its own test: the colon-idiom mention now fails with *"description cites `username`, which is not a parameter"*; breaking only the extractor now fails the positive control.

### Ledger — and the trend is the thing to watch

`round 4 · payload lines changed THIS round: 9` (`registry.ts` only, of which ~1 line is served text and the rest comments) · scaffolding 32.

**242 → 101 → 20 → 9.** Two of round 4's three findings were about scaffolding *the ladder itself wrote in round 3*, which is the signature of a ladder starting to audit its own output rather than the PR. It has not tripped the stop gate — that needs two consecutive **zero**-payload rounds, and F3 was a genuine served-text defect — but round 5 is scoped narrowly and is likely the last.

186 test files / 4,569 tests green; `pnpm typecheck` 0 errors.

### Still not addressed
The Turbopack chunk-hash collision — unverified for a third consecutive round; round 4 confirms nothing in the delta bears on it. **It remains the only thing blocking merge.**

```audit-claims round=4 audited=95ab2f9bc5344da5d22c3fdab3ffaa8534f1103c..e7f989b7f3c6af3ecc70dd86c26790f7b4f2a175
1. the description extractor is now ONE module-level const shared by the assertion and its positive control, so breaking it turns the control red
2. the extractor terminator widened to [`|:] so the `name: "value"` idiom the description already uses is no longer invisible to it
3. the tool description no longer claims `period` ranks; it says narrow by `period`
4. the guard's own limit is recorded at its site: it checks a cited parameter EXISTS, not that the sentence about it is true
5. NOT ADDRESSED: the Turbopack chunk-hash collision, unverified in rounds 2, 3 and 4
```

ZacxDev:
## Audit round 5 — folded in as `403937eb2c`. **Ladder CLOSED at five rounds.**

Round 5 returned **safe to merge** and recommended stopping. Its one finding is real and is fixed here; there is no round 6.

**🟡 The widened terminator had no killing mutation. Fixed.** Reverting ``[`|:]`` back to a bare `` ` `` passed **48/48** — and **92/92** even with round 4's `` `username: "Lykon"` `` mutation *also* present. Cause: on the real description every identifier appearing in colon form **also** appears bare-backticked, so both regexes yield the same distinct set and nothing in the suite could separate them. A widening with no killing mutation is one edit from silently reverting.

Fixed with a **synthetic** fixture, because the live text structurally cannot discriminate — `citedParams('\`username: "x"\`')` must equal `['username']`. Verified: reverting the terminator now fails with `expected [] to deeply equal [ 'username' ]`.

Also made one comment precise: the commented-out period-aware ranking ladder lives in `getModels`, not the `getModelsRaw` path this tool uses.

### Why the ladder stops here

Payload per round: **242 → 101 → 20 → 9 → 9**. Rounds 4 and 5 found only *scaffolding guarding scaffolding the ladder itself wrote in round 3* — the documented signature of a ladder auditing its own output rather than the PR. Round 5 verified every claim of round 4's fixes, reproduced both self-reported mutation figures exactly (48/48, 92/92), and **refuted** the one hazard it was sent to hunt: a shared `/g` regex leaving `lastIndex` state. `matchAll` clones per spec — measured `lastIndex` 0 either side, identical results, and the two tests pass in both orders after physically swapping them.

### What five rounds actually bought

The round count is the wrong measure; the findings are:

1. **A real product defect** — `sort` silently discarded whenever `query` was present, so *"the most popular **anime** models"* returned relevance order. The PR's own headline case, half-broken.
2. **A regression the first fix introduced** — a no-hit text search returning the **whole catalog**, because the relevance restore had been guarding the empty case by accident.
3. **Two filters removed for failing open** — `tag` and `username`, where an unknown value silently drops the filter and hands a language model a plausible wrong answer.
4. **Five separate prose claims the implementation contradicted** — two of them corrections of earlier false claims. That class is why the description guard exists, and why its limit (it checks a parameter **exists**, not that the sentence about it is **true**) is recorded at its site.

### 🔴 The only thing still blocking merge

The **Turbopack chunk-hash collision** — unverified for four consecutive rounds and outside every delta. It is a documented, unfixed, repo-wide defect (`claudedocs/turbopack-chunk-hash-collision-2026-08-18.md`; upstream issue auto-closed untriaged), **not caused by this diff**. I measured the doc's own remedy and it did not hold here: the collision path was **identical** (`_1yhc173._.js`) on two genuinely different module graphs (`d9bbc5e5b8` and `cef4382ee8`), so "perturb the graph and it moves" is not a strategy for this branch.

It needs `main` to move, or the `turbopackServerSideNestedAsyncChunking` decision (24,552 → 7,122 server chunks) taken separately. Neither belongs in a tool-schema PR.

**Final state:** 186 test files / 4,569 tests green; `pnpm typecheck` 0 errors. Verdict from round 5 stands — safe to merge once the build clears.

PR #4518, opened 2026-08-31
fix(moderator): stop accepting the shared admin token inbound

ZacxDev:
Drops `WEBHOOK_TOKEN` from the moderator app's accepted **inbound** credential set. `MOD_INBOUND_TOKEN` becomes the only class accepted on `WebhookEndpoint` routes (`/api/mod/*` and the XGuard agent API).

## Why this is safe to do now

`WEBHOOK_TOKEN` is the shared platform-wide admin credential — the same value authenticates a large number of admin endpoints in the main app. Accepting it here meant any holder of that credential could drive this app's moderation actions. `MOD_INBOUND_TOKEN` is inbound-only: no outbound caller presents it and the main app does not know it, so a leak on an inbound caller cannot reach the main app's admin surface.

The removal was gated on the runtime attribution signal this app emits (`webhook credential presented`), read as a **pair** over a window in which every inbound caller actually called:

- the retiring class at **zero**, standing beside
- a **non-zero** count for the live class in the **same** window — the in-band positive control proving the emit path was live and ingesting at the moment the zero was read.

A zero alone would have been worthless: "nobody presents it any more" and "the log line never shipped, or stopped being ingested" are the identical observation. A wider control window still returns the historical legacy rows, so the zero is a real absence rather than a broken query.

The last caller still presenting the legacy credential inbound was the main app's own delegated moderation path. That traffic is human-driven and can be quiet for hours, so the read was deliberately taken over a window in which it had actually occurred — an absence of calls and an absence of the legacy credential look identical in the counts.

## 🔴 What this does NOT do

**It does not unset the variable, and the variable is still required.** Four services here present `WEBHOOK_TOKEN` **outbound** to the main app — `kono.ts`, `search-index.ts`, `training-moderation.service.ts`, `user-actions.service.ts` — and two of them degrade by `console.warn` and skip, so unsetting it fails **silently** while inbound keeps working. Dropping a class from `acceptedTokens()` and unsetting the variable are different operations; the comments and `.env.example` now say so at the point where someone would make that mistake.

## Tests

Net −3 tests. The mechanical twins for the retired class were redundant (each already had a `MOD_INBOUND_TOKEN` counterpart) and were replaced by guards for the new behaviour:

- the retired credential is now **refused** via `?token=` and via `Bearer`
- `WEBHOOK_TOKEN` alone is **no longer a configuration** — it fails closed with **503**, not 401
- a blank `MOD_INBOUND_TOKEN` does **not** fall back to a set `WEBHOOK_TOKEN` (both assertions 503, never 401 — the two statuses are what distinguish "nothing accepted" from "bad token against something accepted")
- the retired credential **emits no attribution record**, so its count can actually reach zero rather than being stuck at "still in use" forever
- `locals.tokenClient` is left **unset** for a refused credential, so nothing downstream treats it as a webhook caller
- a `WEBHOOK_TOKEN` still present in the environment is **not a candidate** — verified by comparison count, which is what proves the removal took effect at runtime rather than only in the array

Two tests were fixed because they would otherwise have passed **for the wrong reason**: the whitespace-only-secret bypass guard was pinned on the retired variable (it would have gone green via "nothing accepted at all" rather than via the filter it exists to test), and the token-byte-leak guard would have inspected an **empty array** once the legacy secret stopped emitting — a vacuous green on the one test guarding a secret.

## 🔴 One guard is genuinely weakened, deliberately and in the open

The **no-early-exit** property — the match loop compares every candidate so timing does not reveal which credential matched — becomes **structurally unobservable at one accepted credential**: a loop with a `break` and one without both perform exactly one comparison. No assertion can distinguish them.

The loop is unchanged and still has no `break`. Rather than delete the guard, the three cases that discriminated are recorded in place, with the exact assertions to restore and why equal-length fixtures are load-bearing. They are **not** expressed as skipped tests on purpose — a `describe.skip` reads as coverage, which is the claim this must not make. The module header points at it, so adding a second class (scoped tokens) surfaces the need to re-arm.

The same dormancy applies to two others, documented the same way: "both classes attributed differently from one deployment", and the last-match-wins bias on an ambiguous deployment.

## Verification

- `pnpm vitest run --project app:moderator`: **back to the exact pre-change baseline** — 1 failed file / 36 failed tests, all `reports.queries.explain.test.ts`, which fails identically on unmodified `main` (it plans against a real schema). **Zero** failures attributable to this change; measured both before and after rather than assumed.
- Mutation check on the new runtime guard: restoring the legacy class to the accepted set turns it **red** for its own reason — `expected called 1 times, but got 2` and `expected 'authenticated' to be 'refused'`. Red at two credentials, green at one.
- `pnpm check` (svelte-check): `9396 FILES 0 ERRORS 0 WARNINGS`.
- `pnpm lint`: 0 errors. The 2 warnings are pre-existing in files this PR does not touch (`migrate-legacy-strikes.ts`, `audit/generator-restrictions/+page.server.ts`).
- Prettier clean, verified with a positive control (a deliberately malformed copy **is** flagged, so the pass is over real files rather than an empty glob).

### DIFF
--- a/apps/moderator/.env.example
+++ b/apps/moderator/.env.example
@@ -41,10 +41,12 @@ AUTH_INTERNAL_TOKEN=
 WEBHOOK_TOKEN=
 
 # The ONLY credential accepted inbound, on every endpoint wrapped in WebhookEndpoint
-# ($lib/server/webhook-endpoint) — /api/mod/* and the XGuard agent API. Inbound-only: no outbound
-# caller sends it and the main app does not know it, so a leak here cannot touch the main app's admin
-# surface. There is no user behind it — nothing it reaches may attribute a write, and revoking access
-# means rotating it for every holder.
+# ($lib/server/webhook-endpoint) — /api/mod/* and the XGuard agent API. Inbound-only: nothing in THIS
+# app sends it outbound, and — the part that bounds a leak — the main app does not ACCEPT it, so a
+# leak here cannot touch the main app's admin surface. (The main app does KNOW it: it declares
+# MOD_INBOUND_TOKEN and presents it on every delegated call INTO this app. "Inbound-only" is a claim
+# about what accepts it, never about who holds it.) There is no user behind it — nothing it reaches
+# may attribute a write, and revoking access means rotating it for every holder.
 # 🔴 THIS IS NOW THE WHOLE INBOUND CONFIGURATION. Unset (or set EMPTY, which counts as unset) and the
 # endpoints answer 503 — they do NOT fall back to WEBHOOK_TOKEN, which is what stops a blanked
 # credential silently re-opening them to the platform-wide admin token.
--- a/apps/moderator/.env.example
+++ b/apps/moderator/.env.example
@@ -44,7 +44,9 @@ WEBHOOK_TOKEN=
 # ($lib/server/webhook-endpoint) — /api/mod/* and the XGuard agent API. Inbound-only: nothing in THIS
 # app sends it outbound, and — the part that bounds a leak — the main app does not ACCEPT it, so a
 # leak here cannot touch the main app's admin surface. (The main app does KNOW it: it declares
-# MOD_INBOUND_TOKEN and presents it on every delegated call INTO this app. "Inbound-only" is a claim
+# MOD_INBOUND_TOKEN and presents it on delegated calls INTO this app — 🔴 but only WHERE IT IS SET.
+# It is optional in the main app's schema, and its `||` fallback selects the credential this app now
+# refuses when it is unset, so a caller environment missing this key 401s. "Inbound-only" is a claim
 # about what accepts it, never about who holds it.) There is no user behind it — nothing it reaches
 # may attribute a write, and revoking access means rotating it for every holder.
 # 🔴 THIS IS NOW THE WHOLE INBOUND CONFIGURATION. Unset (or set EMPTY, which counts as unset) and the
--- a/apps/moderator/.env.example
+++ b/apps/moderator/.env.example
@@ -30,27 +30,24 @@ AUTH_INTERNAL_TOKEN=
 # --- Main-app callbacks ---
 # Shared secret for the main app's token-guarded internal endpoints (e.g. the Meilisearch re-index
 # callback /api/internal/search-index-update). Must match the main app's WEBHOOK_TOKEN. Unset = sync no-ops.
-# Also ACCEPTED inbound on every endpoint wrapped in WebhookEndpoint ($lib/server/webhook-endpoint) —
-# /api/mod/* and the XGuard agent API. There is no user behind it: nothing it reaches may attribute a
-# write, and revoking access means rotating it for every holder.
-# 🔴 OUTBOUND USE IS WHY THIS CANNOT SIMPLY BE ROTATED. Four services here present it TO the main app,
-# so its value is pinned by the main app's and is not ours to change independently. Prefer
-# MOD_INBOUND_TOKEN below for anything that only calls IN.
+# 🔴 OUTBOUND ONLY — NO LONGER ACCEPTED INBOUND. It was dropped from `acceptedTokens()` in
+# $lib/server/webhook-endpoint once the runtime attribution signal showed no caller presenting it
+# inbound; presenting it to this app now gets a 401. It is STILL REQUIRED: four services here present
+# it TO the main app, so its value is pinned by the main app's and is not ours to change
+# independently.
+# 🔴 DO NOT UNSET IT AS "migration cleanup". Two of those four callers (Meilisearch sync, KoNO
+# finalize) only console.warn and skip, so unsetting fails SILENTLY — and inbound keeps working on
+# MOD_INBOUND_TOKEN the whole time, so nothing loud tells you.
 WEBHOOK_TOKEN=
 
-# Inbound-only service token — accepted on the same WebhookEndpoint routes as WEBHOOK_TOKEN, and
-# equivalent to it inside this app, but it reaches NOTHING beyond this app: no outbound caller sends
-# it and the main app does not know it. Hand this one to a service that only ever calls in (the
-# abuse-detection cronjobs posting to /api/mod/abuse-report) so a leak there cannot touch the main
-# app's admin surface.
-# Either variable alone is a complete configuration; the endpoints answer 503 only while BOTH are unset.
-# A variable set to an EMPTY value counts as unset, so blanking one cannot open the endpoints up.
-#
-# 🔴 MIGRATING OFF WEBHOOK_TOKEN IS A CODE CHANGE, NOT AN UNSET. Removing it from `acceptedTokens()`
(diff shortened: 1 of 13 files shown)


ZacxDev:
## Audit round 1 — fixes applied

```audit-claims round=1 audited=3fb58051f4..50f26eff
1. 🔴 #1 (main app falls back to a now-refused credential where MOD_INBOUND_TOKEN is unset) — addressed OUTSIDE this repo, in the infra repo, not by a code change here. MOD_INBOUND_TOKEN was provisioned into the civitai-next and civitai-next-stage civitai-cfg ConfigMaps and verified present on a live civitai-next primary pod before this PR merges.
2. #2 — /xguard/docs page no longer says the retired token "is also accepted for compatibility"; it now states it is no longer accepted and names the 401 an operator would see.
3. #2 — the test that pinned the old sentence ("the page still explains that the legacy token also works") was inverted to assert the new statement, positively rather than as a word-ban.
4. #2 — apps/moderator/xguard-lab/README.md no longer says "Both are accepted"; its pinning test was updated to the new phrasing.
5. #3 — .env.example no longer claims "the main app does not know it"; it now says the main app DOES know it and that what bounds the leak is the main app not ACCEPTING it.
6. #4 — packages/civitai-moderation/src/client.ts doc comment no longer advertises process.env.WEBHOOK_TOKEN as a working default; it is marked as a credential the server now refuses.
7. #5 — src/server/services/__tests__/moderator-app-token.test.ts docblock no longer says the || fallback makes deploy order irrelevant; it states the legacy arm is now a guaranteed 401 that fails silently.
8. #6 — apps/moderator/src/__tests__/hooks.server.test.ts docblock no longer claims BOTH classes emit or that tokenClient is 'webhook' for both; it records the dormant half and the inverted case.
9. #7 — webhook-endpoint.test.ts tier rationale no longer says the main app presents WEBHOOK_TOKEN on every call; it states what the INVARIANT tier actually pins now.
10. #8 — the "two tokens are independent" test title was replaced with what it asserts: with the retired variable unset its secret is 401, not 503.
11. #9 — the timing file's "what still executes" sentence no longer undercounts; it names the two REGRESSION cases as well.
12. #11 — api-catalog.ts comment no longer names WEBHOOK_TOKEN; it names the credential generically because the branch keys on auth.kind.
13. NOT changed, deliberately: #10 (the ACCEPTED_CREDENTIALS.length > 1 arm is unreachable scaffolding for a future second class) — left as-is and documented rather than removed.
```

**Verification after the fixes:** `pnpm vitest run --project app:moderator` is back to the pre-change baseline — 36 failed / 373 passed, every failure in the pre-existing `reports.queries.explain.test.ts` (needs a real schema; fails identically on unmodified `main`), zero attributable. `apps/moderator` `pnpm check`: `9396 FILES 0 ERRORS 0 WARNINGS`. `src/server/services/__tests__/moderator-app-token.test.ts`: 8/8.

**On the 🔴, and correcting one half of it:** the audit's claim that PR previews would hit a prod 401 via `tests/preview-moderation.spec.ts` does **not** hold — that spec deliberately asserts the redirect `Location` **without following the hop**, and says so in its own comment, precisely so it does not probe prod from CI. There is no automated breakage.

The runtime exposure was real and is now closed, and it was **wider** than the audit found: measured across all 14 live `civitai-cfg` ConfigMaps, `MOD_INBOUND_TOKEN` was set in exactly **one** (`civitai-dp-prod`). It is now also in `civitai-next` and `civitai-next-stage`. `civitai-next` additionally covers every future PR preview, because the preview pipeline builds a preview's `civitai-cfg` by copying it from `civitai-next`. Two narrowings of the reported blast radius: `civitai-app` runs a **different repo's** image and was never affected, and `civitai-next-stage` is scaled to **0/0** so it had no pods that could have 401'd.

ZacxDev:
## Audit round 2 — fixes applied, and the ladder stops here

```audit-claims round=2 audited=50f26effaf..847f22b1
1. 🟡-1 — src/server/services/moderator-app.service.ts comment no longer says the legacy arm "keeps working" or that the spoke "is dropping" the token. It now states the arm is a guaranteed 401, that || falls through on an EMPTY value and never on a REJECTION, and that MOD_INBOUND_TOKEN is effectively required wherever this client is used. This was the round-1 fix's own miss: it corrected the mirroring TEST docblock and left the shipped source, leaving the two in direct contradiction.
2. 🟡-2 — credential-handed-out.test.ts header is past tense throughout and states the retired token now 401s.
3. 🟡-3 — .env.example no longer claims the main app presents the narrow token on EVERY delegated call; it says "only WHERE IT IS SET", names it optional in the main app's schema, and says the fallback selects a refused credential when unset.
4. 🟡-4 — the 503-body test was relabelled INVARIANT -> REGRESSION (it pins a string this PR edits, so it is red on the pre-change side), which makes the file's own matrix claim true again. A note records the general rule: sort by what the MATRIX does, never by what the test is thematically about.
5. 🟢-1 — the lab README no longer states a biconditional; a 401 is "most likely" the retired token, and the other causes that 401 identically are named.
6. 🟢-2 — the timing file's "what still executes" sentence now names all five live tests, not four.
7. ALSO from the round's "could not verify" section: "fails SILENTLY / there is no alert" was overstated. onFailure does emit moderator-app-request-failed to Axiom; whether anything alerts on that stream is not settled in this repo, and the comment now says exactly that.
```

**Verification:** suite at the pre-change baseline — 36 failed / 373 passed, all in the pre-existing `reports.queries.explain.test.ts`, zero attributable. `svelte-check` `9396 FILES 0 ERRORS 0 WARNINGS`. `moderator-app-token.test.ts` 8/8. Prettier clean (the README needed a reflow, applied in `847f22b173`).

### Why this is the last round

**Attribution.** Rounds 1 and 2 changed **zero executable production lines** — every fix in both was a comment, a JSDoc, a `.env.example` block, a README, rendered prose, or a test title. Round 2's own auditor measured the same thing independently ("0 change executable behaviour"). The shipped behaviour has been frozen since `3fb58051f4`: `webhook-endpoint.ts`'s logic is untouched since then. That is the attribution gate in substance — the ladder has left the PR and is now auditing the prose the ladder itself wrote, which is the condition it exists to detect.

**Severity.** Round 2 returned **no 🔴**. Its findings were all one class — a claim this PR staled — with no blast radius beyond "a document contains a false sentence".

**The shape, swept rather than spot-fixed.** The once-per-ladder sweep that no delta round can perform (a claim staled by an *earlier* commit of the PR sits below every later round's range) was run against the current head, ignoring ranges: every surviving mention of `WEBHOOK_TOKEN` acceptance across `apps/moderator`, `packages/civitai-moderation`, `moderator-app.service.ts` and `server-schema.ts` now reads correctly, in the past tense. Positive control: 59 mentions across 16 files, so the search sees real content.

**One thing deliberately left open**, so it is recorded as open rather than absent: the `||` fallback in `moderator-app.service.ts` is now a silent-401 path rather than a deploy-order bridge. Deleting it and making `MOD_INBOUND_TOKEN` required — so a missing key is a loud boot failure instead of a moderator-visible toast — is the correct follow-up. It is documented in that file; it is not done here because it is a main-app behaviour change that deserves its own change and its own verification, not a rider on this one.

PR #4523, opened 2026-08-31
feat(collections): sort a collection by when things were added to it

manuelurenah:
Fixes [868kyuehg](https://app.clickup.com/t/8459928/868kyuehg).

## The report

> Seems like 'Newest' in collection sorts by upload date and not published date. Was a bit confusing to add a recent image and then see it far down in the collection (probably because uploader scheduled the post?).

Correct. A collection sorted by Newest orders on `i."id" DESC`, which is **upload** order. Schedule a post and the upload happens weeks before publication, so adding a freshly published image to a collection drops it far down the feed.

## What this does instead of the literal fix

The task's closing condition was "Newest uses published date". **I did not do that, because it is a 20–300x regression on the collection feed.** Measured on the prod replica:

| ORDER BY | 4,500-item collection | 163,624-item collection |
|---|---|---|
| `i."id" DESC` (today) | 15 ms | 19 ms |
| `i."sortAt" DESC` (published date) | 334 ms | **5,738 ms** |
| `ct."id" DESC` (this PR, no index) | — | 246 ms |
| `ct."id" DESC` (this PR, with index) | — | see below |

`i."id" DESC` is cheap only because `CollectionItem_image_idx ("collectionId", "imageId")` already yields that order for free. Ordering on any `Image` column forces the planner to materialise every item in the collection and sort it — there is no index that avoids this without denormalising `sortAt` onto `CollectionItem` (85M+ rows, a new trigger, a full backfill). The commented-out `sortAt` branch in `image.service.ts` and its `// TODO this causes the app to spike` note were right.

So this adds a **`Recently Added`** sort ordering on `CollectionItem.id`, and makes it the **default** for collections. That is what the reporter actually expected — add something, see it at the top — and it costs about what Newest costs. `Newest` keeps its current meaning.

Scale for context: 954,537 image collections, 85M items, p50 = 8 / p99 = 1,384 / max = 163,624. 6.7% of recent collection items were published more than a day after upload, so the confusion is not rare.

## Scope

Applied to all four collection types. Only **images** had the bug — posts and articles already sort Newest on `publishedAt`, models on `lastVersionAt` — so for those three this is a new capability, added so the four types behave alike.

The three services that filtered with `EXISTS (SELECT 1 FROM "CollectionItem" ...)` needed a real JOIN to expose `ci."id"` to `ORDER BY`. The partial unique indexes on `("collectionId", <entity>Id)` mean the join cannot multiply rows, so both shapes select the same entities.

Two guards worth reviewing:

- `Recently Added` is withheld from every general feed menu via the `*SortHidden` maps, and every service returns a 400 when it arrives without a `collectionId` — mirroring the existing `Random` guard.
- It is held out of the **contest random-sort pools**. Contest feeds re-roll their sort each render to spread entry visibility; a deterministic sort in the pool would pin the newest entries to the top for whichever renders draw it, and nothing about the feed would look wrong.

## ⚠️ Migration is NOT applied

`packages/civitai-db-schema/prisma/migrations/20260831120000_collection_item_recently_added_idx/`

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS "CollectionItem_collectionId_id_idx"
  ON "CollectionItem" ("collectionId", id DESC);
```

**This needs a decision before anyone runs it.** `CollectionItem` is **204M rows / 15 GB heap / 82 GB of existing indexes**, and this adds roughly **6 GB**.

What it buys, measured rather than assumed — I built an equivalent index scoped to a single `collectionId` on dev, then dropped it: the same page went **42.2 ms → 1.8 ms**, and the `Sort` node disappeared entirely. The ratio matters less than the shape: the scan reads **102 rows and stops at the LIMIT** instead of reading all 44,367, so page cost stops scaling with collection size.

Without it the feature still works, at 246 ms on the largest collection. Run `CONCURRENTLY`, outside a transaction; a cancelled build leaves an INVALID index that must be dropped before retrying. Full operator notes are in the migration file.

## Verification

Exercised all four feeds live against the dev database. The one that actually proves it — a collection where add order and upload order diverge:

```
Recently Added → 124995538, 86127674, 96212893
Newest         → 124995538, 96212893, 86127674
```

The `CollectionItem` ids confirm the first is the true add order: image `96212893` was uploaded later but added earlier. That is exactly the reporter's case.

- Full unit suite: **1528 files / 24,106 tests passed, 0 failed**
- `pnpm typecheck`: 0 errors
- `db:check-generated`: clean

New tests fail loudly on a revert — verified by mutation, not by assuming. Disabling the `RecentlyAdded` branch produces `expected [504, 503, 502, 501] to deeply equal [501, 502, 503, 504]` rather than a hang, because the fixture is built so Newest and Recently Added are exact reversals of each other and cannot agree by coincidence.

## Things a reviewer should look at

- **Articles could not be exercised end-to-end.** The dev DB throws `permission denied for table ArticleRank` on *both* `Newest` and `Recently Added`, so it is a pre-existing grant issue rather than this change. I validated that SQL shape directly against the prod replica instead, and it returns correctly add-ordered rows — but it has not run through the real service.
- **`post.service.ts` interpolates `collectionId` into a raw join string** rather than binding it, because `joins` is a `string[]`. It is `z.number()`-validated upstream and matches the `displayReviewItems` line immediately above it, but it is the one spot in the diff I would flag. Parameterising it means converting that array to `Prisma.Sql[]`, which is wider than this fix.
- **The default sort changes for all four collection types**, not just images. Contest collections are unaffected.
- **`flipt-eval-context.test.ts` ledger rows move by +7 lines.** That ledger is keyed on `file:line` and this change inserts lines above four `getFliptBoolean` sites in `image.service.ts`. The sites themselves are unchanged; I updated the rows rather than exempting anything.

## Not done

`Newest` and `Oldest` are not relabelled to say "upload date". `DumbSortFilter` passes the enum *value* as the button label, so custom option labels would appear in the dropdown but not on the button — inconsistent rather than clearer. Worth a separate change if we want it.

https://claude.ai/code/session_01RnFnLu369fHmf7dUQBHMXm

### DIFF
--- a/docs/features/collaborative-collections.md
+++ b/docs/features/collaborative-collections.md
@@ -266,9 +266,9 @@ separately everywhere else too.
 `pendingReviewCount` on `collection.getById`. The count is computed only for `manage` holders, and
 it is an exact `count` with **no cap and no mode exclusion** — resist adding either.
 
-The reason that is safe is not in `schema.full.prisma`. That file declares only
-`@@index([collectionId], type: Hash)` on `CollectionItem`, which would make a status-filtered count
-walk the whole collection. Production actually carries
+The reason that is safe is not in `schema.full.prisma`. The only `CollectionItem` index that file
+declares that a status-filtered count could use is `@@index([collectionId], type: Hash)`, which
+would make the count walk the whole collection. Production actually carries
 `CollectionItem_collectionId_status_covered` — a btree on `(collectionId, status, createdAt DESC)`
 — so the count is an index-only scan: measured at **0.5 ms / 4 buffers** on a 190k-item collection
 and **0.8 ms / 73 buffers** for a real 466-row review backlog. This is the same
--- a/packages/civitai-db-schema/prisma/migrations/20260831120000_collection_item_recently_added_idx/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260831120000_collection_item_recently_added_idx/migration.sql
@@ -1,29 +1,24 @@
--- Serves the collection feeds' new "Recently Added" sort, which orders on the
--- CollectionItem row's own id:
---   WHERE "collectionId" = $1 AND <status/entity filters> ORDER BY id DESC LIMIT $2
+-- Serves the collection feeds' "Recently Added" sort, which orders on the CollectionItem
+-- row's own id rather than the entity's. No existing index can produce that order: the
+-- closest, CollectionItem_image_idx ("collectionId", "imageId"), is why the pre-existing
+-- Newest sort on images is cheap — imageId order IS id order there — and it says nothing
+-- about the order items were added in.
 --
--- No existing index can produce that order. The closest, CollectionItem_image_idx
--- ("collectionId", "imageId"), is why the pre-existing Newest sort is cheap — it
--- yields imageId order, which for images IS id order. Ordering on the item id
--- instead makes the planner read every row of the collection and sort: measured on
--- the prod replica against the largest image collection (163,624 items), 246ms
--- versus 19ms for the same page under Newest.
+-- Not partial on an entity column, unlike the four ("collectionId", <entity>Id) uniques
+-- beside it. Collection.tsx dispatches a collection to exactly one feed by Collection.type,
+-- so within a single collectionId the entity filter drops nothing in practice, and one
+-- index serves images, models, posts and articles instead of four over the same table.
 --
--- What this index buys, measured on a 44,367-item collection with an equivalent
--- index scoped to that one collectionId: 42.2ms -> 1.8ms, and the Sort node
--- disappears. The point is the shape rather than the ratio — the scan reads 102
--- rows and stops at the LIMIT instead of reading all 44,367, so the page cost stops
--- scaling with the size of the collection.
+-- Cost of the full feed query without it, prod replica, most-viewed image collection
+-- (~16.5k items): 448ms warm, 1.4s cold, against 2-3ms for Newest. On a collection whose
+-- newest item is far behind the table's max id the planner switches to a backward scan of
+-- CollectionItem_pkey and the cost is unbounded — one 191k-item model collection did not
+-- return in 45s.
 --
--- Not partial on an entity column, unlike the four ("collectionId", <entity>Id)
--- uniques beside it. A collection holds one entity type (Collection.type), so
--- within a single collectionId the entity filter drops nothing and the ordered
--- scan still stops at the window — one index serves images, models, posts and
--- articles rather than four partial ones over the same 204M rows.
---
--- CONCURRENTLY, because the table is 15 GB / ~204M rows and this index adds ~6 GB
--- to the 82 GB already on it. Run it OUTSIDE a transaction: psql's -c with two
--- statements opens one implicitly and the server refuses. A cancelled CONCURRENTLY
--- build leaves an INVALID index that must be dropped before retrying.
+-- CONCURRENTLY, because the table is 15 GB / ~204M rows. Run it OUTSIDE a transaction:
+-- psql's -c with two statements opens one implicitly and the server refuses. A cancelled
+-- CONCURRENTLY build leaves an INVALID index; IF NOT EXISTS will then skip creation and
+-- exit 0 with the invalid index still in place, so check indisvalid before trusting a
+-- successful re-run.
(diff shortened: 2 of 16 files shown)


manuelurenah:
Pushed `3a1a051f55` — applies the findings from a five-lane review (reuse, correctness, perf, tests, intent) plus a comment/docs cleanup pass.

## Two things changed behaviour

**Only image collections default to Recently Added now.** Models, posts and articles are back on Newest. They never had the reported bug — they sort on `publishedAt` / `lastVersionAt` — and the default flip there came from combining two separate answers rather than being asked for. It also removes the worst measured case: a 191k-item model collection did not return in 45s under the new ordering.

**The post cursor is single-column for this sort.** `encodePostCursor` always emitted `value|id`, producing `(ci."id", p.id) <= (…)` — a row comparison spanning two tables, which Postgres cannot push into an Index Cond on either. Page 1 used the new index; every page after it re-scanned the collection from the newest item. Verified end to end: page 1 returns cursor `"212338035"` (was `212338035|30523337`), page 2 resumes at exactly the 6th row with no skip or repeat.

## ⚠️ The perf numbers in the first commit were wrong

I measured them on a `CollectionItem`-only fragment. The `ct` CTE has no `LIMIT`, so that was never the query the feed issues. Corrected, and the migration comment now carries the real figures:

| | measured |
|---|---|
| most-viewed image collection (~16.5k items) | **448 ms warm / 1.4 s cold**, against 2-3 ms for Newest |
| a 191k-item model collection | **did not return in 45 s** — the planner switches to a backward scan of `CollectionItem_pkey` when a collection's newest item is far behind the table's max id |

**This makes the index a deploy gate, not an optimisation.** Apply `CollectionItem_collectionId_id_idx` and confirm `indisvalid` before this ships.

## Also applied

- The 400 guards in `getAllImages` and `getModelsRaw` sat *below* early empty returns, so the misuse they exist to make legible came back as a silent empty page. Both now run before any return; the model one also fires ahead of the Meilisearch round-trip.
- `getPostsInfinite`'s collection join now uses `getAvailableCollectionItemsFilterForUser` — the helper model and article already call in the same branch. Removes the third copy of the visibility rule and the raw-string interpolation of `collectionId` and user id.
- `getPostSortClauses` throws when `RecentlyAdded` arrives without the join, instead of falling through to `publishedAt`. That fallback was the exact shape of the regression, and a test was asserting it as expected output.
- Two comments were false rather than imprecise. The contest-pool one claimed the pool "must hold only sorts that shuffle" — no member shuffles, the rotation *is* the randomness, and `Newest` is in every pool. A test comment claimed it guarded `SortFilter`'s `sortOptions`, which is module-private and imported by nothing; I fixed the assertion rather than the wording.
- Extracted `resolveImageCollectionSort` so the default, the URL override and the contest override are covered. A revert of the default previously passed every test in the diff.
- The three `EXISTS`→`JOIN` comments now name where the partial uniques actually live (`containers/db/docker-init/02_all_dll.sql`, not `schema.full.prisma`) and record that the model index name has drifted between that file and prod.
- `docs/features/collaborative-collections.md` said `schema.full.prisma` declares *only* `@@index([collectionId], type: Hash)` on `CollectionItem`; the new index made that wrong.

## Declined, with reason

Exposing a curated sort subset to the LLM tool registry. `Recently Added` now parses and then 400s on `/api/v1` surfaces that accept no `collectionId` — but `registry.ts` deliberately mirrors the full `ModelSort` enum and its test asserts equality, arguing a subset "would be a second rule that can drift from the one `models.ts` already enforces". `ImageSort.Random` has had the identical property for as long as it has existed. Worth fixing at the shared level, not here.

## Known gaps

- **The service layer still has no test.** Reverting the `image.service.ts` orderBy branch, or any of the three JOIN conversions, prints nothing. Harnesses to copy exist (`get-models-raw.paid-access-filter.test.ts`, `getAllImages-prioritized-guards.test.ts`) — not done here to keep this diff reviewable.
- **Articles were never exercised end to end.** The dev DB throws `permission denied for table ArticleRank` on *both* sorts, so it is a pre-existing grant issue; that SQL shape was validated directly against the prod replica instead.
- The `ci.createdAt` vs `ci.id` split (`getCollectionItemsByCollectionId` orders by `createdAt`, nullable and NULLS FIRST) is untouched — two "recently added" orders that disagree on legacy rows.

Full suite 24,108 passed / 0 failed; typecheck clean.

PR #4524, opened 2026-08-31
feat(comments): mute notifications for a single comment thread

JustMaier:
## What

Per-type notification settings are all-or-nothing: turning off "New replies to comment threads you are in" silences every thread on the site. This adds a per-thread mute. The overflow menu on a comment gains **Mute this thread** / **Unmute this thread**, and the rule is uniform at every position — it always mutes the thread hanging off the comment you clicked.

That single rule gives the two controls Justin described, and they mean different things:

- **On the comment heading a conversation** — silences that conversation and everything nested under it, including replies that arrive later and deeper. Verified against production rows, not reasoned: threads `519329` (entity root) → `529156` (the reply thread under a top-level comment) → `530128` (a reply inside that). A mute written on `529156` is found by the walk from `530128`, so a later reply at that depth is suppressed.
- **On your own comment** — silences replies to your message. Same mechanism, one level down.

## The instructions this was built to, verbatim

Justin, spoken, 2026-08-27, on scope:

> "disabling notifications for a thread should just make it so that you don't get the ones just about, you know, hey, somebody in this thread that you're a part of responded or left something. So direct mentions would be excluded from this."

And on nesting and placement:

> "I think that I would imagine that it would be like a top level thing. So. At the top of the comment thread, you'd have it, and that would mute it for. The whole thread for you and all children of that thread. And then on your comment, if you're not wanting to hear responses, you'd click on that there, and that'll not give you notifications about the things in your thread."

Both are load-bearing and neither is recoverable from the code, which is why they are quoted rather than paraphrased.

**A note for whoever reads the review history:** an intent review flagged "muting from your own reply does not silence the conversation you are in" as a defect. It is not — it is the design. The second quote settles it: to silence a conversation you mute at its head. The review was measuring against an assumption that had not been settled yet, and settling it was the point.

## Storage

New `ThreadMute(userId, threadId, mutedAt)`, PK `(userId, threadId)`, plus a `threadId` index for the FK cascade only — the suppression never uses it.

🔴 **The migration is applied BY HAND**, as all migrations here are: `packages/civitai-db-schema/prisma/migrations/20260827150000_thread_mute/migration.sql`. New table only — no backfill, no enum — so there is nothing to sequence against a deploy.

## Where the mute is enforced

Producer-side, in the SQL each comment notification processor runs, via `notThreadMuted()` in `base.notifications.ts`. Not a read-side filter: the notifications app is a separate database and would have to cross-join main-DB `Thread` to see a mute.

It has to be **uniform** across the comment family. Those processors share a dedupe key and run in priority batches (Mention 1 → DirectResponse 2 → ThreadResponse 3 → EntityOwner 4), and the first to land claims the key. A processor that skipped the filter would not merely leak its own notification — it would claim the key and *replace* the one that was correctly suppressed, so muting a thread would change which notification you get rather than silencing it.

**Mentions are exempt**, per the first quote. Safe against that batching rather than in spite of it: Mention is priority 1, so a mention in a muted thread lands first and claims the key, and the notifications the mute suppressed cannot re-emerge behind it.

## The eight entity-owner filters, and when they became reachable

The entity-owner processors pin their `Thread` join to the entity **root** thread (`t."imageId" IS NOT NULL` and seven siblings) and the walk only climbs, so a mute on a conversation under a comment can never match them. Until the section-level control existed, nothing wrote a mute against a root thread at all, and those eight clauses were dead — thirteen filtered call sites reading as thirteen covered processors, which is dead code that looks like coverage.

**The section control is what makes them live.** Muting a discussion now suppresses "new comments on your image" for the person who muted it. `new-3d-model-comment-nested` pins `t."commentId"` instead, and was the one owner processor the per-comment mute already reached.

That asymmetry is recorded at the fragment as well as here, because it is the sort of thing that reads as a puzzle rather than a decision.

## The ancestor walk climbs two edges

`Thread.parentThreadId` is written from request input, which is why `throwIfThreadChainLocked` refuses to decide on it — but it is the only link left once a parent comment is deleted and `Thread.commentId` goes to `SetNull`. Measured on production 2026-08-27:

- **1,508 of 461,725** comment-anchored threads (0.33%) have no usable `parentThreadId`.
- **3,110 of 254,368** orphans (1.2%) have no `commentId` and only a `parentThreadId`.

Neither edge dominates, so the walk takes both, with `UNION` so a diamond or a corrupted cycle converges. Control on real data: seeded at thread `4487743`, which has a NULL `parentThreadId` and an intact `commentId`, the single-edge walk finds 1 row — itself, no ancestors — and the two-edge walk finds 2.

(The 254,368 figure is also a cross-check on `throwIfThreadChainLocked`'s own doc block, which records 250,071 orphans when it was written. Same measurement, drifted up since. It is the one claim in that comment anyone has verified since.)

### The depth cap does not close the gap, and is not written as though it does

`MAX_THREAD_CHAIN_DEPTH` now lives in `src/server/common/thread-chain.ts` so the lock walk and the mute walk cannot drift apart on the number. They deliberately do **not** share what happens at the cap: the lock treats it as "could not resolve" and refuses the write; the mute treats it as "no mute found" and sends the notification, because a notification silently withheld is a failure the user cannot see while one arriving anyway is visible and correctable.

Measured over every thread rather than a recent sample: **the deepest chain in production is 247, 706 threads are 20 or deeper, and 149 are past the cap of 100.** Past it, a mute stops suppressing. The previous comment on that constant claimed the deepest chain was 17 — true of the 2,000-thread sample it measured, false of the population — and has been replaced.

## The convention guard had to be rewritten before it could fail

`no-unmuteable-comment-processor` pins that a processor added without the filter cannot ship quietly. Its first version parsed the source file with a regex and was blind three ways, all structural:

- a processor whose key line the regex missed was appended to the **previous** processor's body, which already carried the filter, so it stayed green;
- the literal text `notThreadMuted(` inside a SQL comment satisfied it — the same hole `comment.appListing-owner.test.ts` strips comments to close;
- a recipient/author swap passed.

It now asserts on the **generated SQL**, like `notifications.block-filtering.test.ts` beside it, so a processor is covered because it is a key on the real object rather than because its source matched a pattern. Three mutants that were green are now red by name: a fourteenth processor with no filter, a filter replaced by a SQL comment, and a mute keyed on the comment author.

**The recipient assertion itself had to be fixed before it could fail.** Its first version passed the author/recipient swap 29 of 29, because `notBlockedBetween` emits both directions and a positional read of `blk."userId"` picked up the *actor* as a valid expected recipient. It now reads the clause carrying `type IN ('Block', 'Hide')`, which is the recipient's own direction.

## Deliberately out of scope

Legacy `Comment` (model-page discussion) rows have no `Thread`, so there is nothing to mute — those three processors are uncovered **entirely** rather than half-covered, which is why a mute cannot leak through them. `new-comic-comment` is written by `createNotification` rather than a `prepareQuery`, so no producer-side filter reaches it at all; its comments are always created in the chapter root thread, so it has nothing to suppress today. Also out: a settings-page list of muted threads, auto-mute heuristics, digest/email, chat. `tos-violation` and moderation notifications stay unmuteable.

## Verification

- `pnpm run test:lint-rules` — `Test Files 25 passed (25)`, `Tests 388 passed (388)`.
- Full unit suite — `Test Files 1 failed | 1479 passed | 1 skipped (1481)`, `Tests 4 failed | 23135 passed | 28 skipped (23167)`. The 4 are `scripts/test-perf/__tests__/trace-flush.test.ts`, which spawns nested vitest runs and gets empty output on this machine. Pre-existing: it fails identically with this branch's `package.json` change stashed, and when run alone. Untouched here.
- Controls, so the tests above are worth something: the three guard mutants above; dropping the ancestry from the lazy `Thread` create fails both ancestry tests by name; dropping `childThread` from the `select` fails the read assertion, which matters because mocked Prisma ignores `select` and nothing else in the file could see it go missing.

A `LEFT JOIN` in the first two-edge version broke `comment.bounty-entry-owner.test.ts`, which asserts these statements contain none — a widened join there NULLs a name into the notification text. The subquery was rewritten rather than that guard narrowed. Full-suite only; a targeted run would not have shown it.

## One note for the next person editing this branch

Reverting a mutant with `git checkout <file>` restores the file to HEAD, and HEAD is all it knows — so it silently discarded an uncommitted edit to that same file made earlier in the same session. Commit or stash before mutating, and revert the hunk rather than the file.

The `git status --porcelain` check afterwards does not catch this. Matching HEAD is exactly what the over-revert produces, so the check confirms the damage and reads as reassurance. **A clean tree is evidence of a successful revert, not of a correct one.**

## The section-level control, and why it writes less than it looks like it should

A second control sits beside the top-level comment composer: **Mute this discussion**, which mutes the entity's root thread — the whole comment section. It is the only writer that targets a root thread, and therefore the thing that makes the mute filter on the eight entity-owner processors reachable at all.

Different wording from the per-comment **Mute this thread** on purpose. The two sit a few pixels apart and do different things, so they say different things.

**It deliberately does not create the thread it mutes.** An entity with no thread has no comments, so there is nothing to be notified about and nothing to mute; the first comment creates the thread and the control works from then on. That started as an `upsert` and became a plain read once the reasoning was written down — a guard you no longer need beats a guard you have to keep correct.

The per-comment path still creates, because muting your own comment before anyone has replied is the main case. So only that create is gated, on the same thread-lock chain its siblings walk. Muting a thread that already exists is deliberately left ungated: it writes one small row, and a locked thread is not a reason someone cannot mute its notifications.

Both mutations are `guardedProcedure` rather than plain authenticated, matching `upsert` beside them, and share **one** rate-limit budget between them — two budgets would be twice the number it reads.

⚠️ Rate limiting is skipped in dev, test and preview, so that part is asserted by the router's shape and by nothing running. It is not covered by any test in this PR and cannot be observed in any environment short of production.


JustMaier:
## Review findings at `b548e7d041` — collected, not fixed

Seven review lanes have read this branch across three passes. This comment is the outstanding list, deliberately left unfixed: every item below is a one-to-few-line change, and **every fix round on this branch introduced a defect**, so the remaining ones are being done with a clear head rather than at the end of a long session.

Three things to read before the diff, because they tell you where to look:

1. **The same fix keeps getting applied to one control and not its twin — three times so far.** The per-comment unmute toast was corrected to stop promising delivery it can't guarantee; the section one still says *"You'll be notified about new comments here again"* while a separately-muted sub-thread keeps suppressing. Same shape, one file over. Whenever you fix something here, check the other control.
2. **Seven comments on this branch stated a reason that was false.** The code was right each time; the sentence beside it was invented. The `UNION ALL` doc block is the current one — see below.
3. **The read-path tests do not work.** They were added this round to close a "zero tests" finding, and five *new* mutations leave 31/31 green.

---

### The read-path tests are not load-bearing

Five mutations to `getThreadMuted` / `getSectionMuted`, each run, each leaving the suite green:

| Mutation | Production effect |
|---|---|
| seed the walk from `comment.threadId` only | a mute the user placed themselves becomes invisible; the menu offers "Mute" on an already-muted thread and the write is a duplicate-key error |
| drop `childThread` from the read `select` | same as above (`ownThreadId` becomes null) |
| key the `own` predicate on `comment.threadId` | `own` is never true, so their own mute reports as inherited and **the menu permanently refuses to let them unmute it** |
| flip either `?? false` default to `?? true` | every comment on the site reports muted, menu inverted |
| `getSectionMuted` always returns `muted: true` | no test asserts the not-muted return |

The second is the exact mutation the *write*-side test was written to catch, comment and all — the read path got the same code shape without the same pin. The fourth is the common path: with nothing muted the statement returns one all-NULL row, so `?? false` runs on nearly every call, and no fixture covers it.

Four additions kill all five: mirror the existing argument pin onto `dbRead.commentV2.findUnique`; assert the whole `$queryRaw` call rather than value membership; add a `{ own: null, ancestor: null }` fixture; assert `getSectionMuted`'s not-muted return.

Two mutations *were* caught (`viaAncestor: ancestor`, and restoring the early return), so the file is not uniformly weak — the write-side coverage is real.

### Defects

- **Read-only mode no longer hides the section control.** `CreateComment` checks `isLocked || isMuted` before `isReadonly`, and the first branch renders it — contradicting the comment on that branch. One line: order `isReadonly` first.
- **On a locked thread, a reply-less comment offers an enabled "Mute this thread" that always throws.** The lazy thread create is gated on the lock, and a reply-less comment has nothing but a create. A sibling comment that happens to have replies mutes fine — same control, same page, opposite behaviour, decided by something invisible to the user.
- **The section unmute toast overclaims** (item 1 above).
- **Neither mute invalidates the other's cache**, under a global `staleTime: Infinity`. Both change `viaAncestor` for every descendant. Worst direction: unmute a parent and a descendant stays disabled on "Muted with the thread above" until reload.
- **`threadMute.create` is a bare create in both toggles**, with the same unique-constraint exposure that `thread.upsert` was hardened against forty lines above. Two concurrent clicks error.
- **`forceLocked` drops the control** two lines above the comment arguing it shouldn't. Inert today — the only caller passes `undefined`.

### The `UNION ALL` comment claims a saving that does not exist

Measured on the prod replica, nine interleaved runs per variant at 5,000 evaluations of the real producer shape: **`UNION ALL` 112.9 ms, `UNION` 113.2 ms**, with `UNION` faster in five of nine pairs. Mean chain depth in production is **1.62** — the dedupe was being asked to dedupe two rows. Keep `UNION ALL` (a dedupe that provably cannot fire is dead weight and misleading), but the justification must not claim a measured cost.

Same block: out-degree is 1 because **both recursive joins are primary-key equalities**, not because `Thread.commentId` is unique. The current wording makes the bound look contingent on an index someone could drop.

### Smaller

- `!replyToCommentId` is retyped as a render gate at two call sites while the hook already receives it as `enabled` — fold it into the hook's return, which also closes the gap where a third render path renders the control ungated. That gap is not hypothetical; splitting out the locked branch created one.
- The two mute mutations now cache inconsistently (one invalidates, one `setData`s). The `setData` is correct today — a root thread has no ancestors — but it hand-translates between two shapes.
- The four-line delete-then-create toggle appears verbatim twice; one `setThreadMute` helper owns the ordering and the race.
- A bare unlabelled kebab beside "This thread has been locked" reads as options *about the lock*. A tooltip or visible label is cheap.

### 🔴 Deployment ordering

`notThreadMuted` is inside the `WHERE` of **every** comment notification processor. If the code deploys before `20260827150000_thread_mute` is applied, it is not the new feature that breaks — **every comment notification query errors on a missing relation and the whole family stops**. Apply the SQL first, per environment, then deploy. Applied to dev only so far.

### Out of scope, so it is a decision rather than a discovery

- **Model detail page discussion has no mute at all.** It runs on the legacy `Comment` table, which has no `Thread`, so it was never reachable. Not a miss against the ask — but it is the highest-traffic comment surface on the site.
- **Comics has no owner-side mute.** `new-comic-comment` is written by `createNotification` rather than a `prepareQuery`, so no producer-side filter reaches it, and `comicChapter` is excluded from the section schema because `Thread` has no `comicChapterId`.

### State

`b548e7d041`. Typecheck 0. `test:lint-rules` 25 files / 389 tests. Full suite `1 failed | 1479 passed | 1 skipped (1481)` — the four failures are `scripts/test-perf/__tests__/trace-flush.test.ts`, pre-existing and unrelated (verified by stashing this branch's `package.json` change and re-running). `blocks.router.workflow.test.ts` collects 358, so the submodule is intact and the suite means something.

JustMaier:
## Conflicted against `origin/main` — read this before rebasing

**This PR has no CI, and it looks like it has passing CI.** `gh pr checks` reports `Passed: 2, Failed: 0`; the real total is 14. GitHub cannot build a merge commit for a conflicted branch, so every merge-dependent job never registers and the two that survive are the ones that don't need one. Verified with a command rather than inferred:

```
gh pr view 4524 --json mergeable,mergeStateStatus
mergeable=CONFLICTING   mergeStateStatus=DIRTY   head=01a16b1379
```

`origin/main` moved several times while this was in review (tip `081e6516f8` at the time of writing).

### Eight files are touched on both sides; five actually conflict

Counted from `git merge-tree` against the merge base, and confirmed by an aborted trial rebase — **3 conflict hunks in `CLAUDE.md`, 3 in `.claude/agents/civitai-test-review.md`, 1 in `package.json`**. "Changed in both" is not the same as a conflict; these are the ones with real markers.

**Mechanical — the guard-count bookkeeping.** `main` added two convention guards (`no-unguarded-billable-submit`, `no-unguarded-user-text`) and this branch adds `no-unmuteable-comment-processor`. Both sides increment the same two sentences and edit the same `test:lint-rules` line. Resolution is to **take all three guards and let the counts follow** — 21 → 22 in `<n> live in …`, 26 → 27 in `test:lint-rules names <n> files today`, in both `CLAUDE.md` and the agent doc, plus the new file in the script list. This one announces itself if you get it wrong: `no-lint-rules-script-drift` reads those exact phrasings and fails when they disagree with the directory.

**Mechanical but regenerate, don't hand-merge.** `packages/civitai-db-schema/prisma/schema.full.prisma` plus the generated `src/kysely/types.ts` and `src/models.ts` — another schema change landed alongside `ThreadMute`. Resolve the `.prisma` by hand, then run `pnpm run db:generate` and let it produce the two generated files. Merging generated output by hand is how they drift from the schema that produced them.

🔴 **Needs real attention — `src/server/services/commentsv2.service.ts` and `src/components/CommentsV2/Comment/Comment.tsx`.** These are the two files at the centre of this feature and someone has been editing the same comment service concurrently. This is where a *semantic* conflict hides — the kind git resolves cleanly and wrongly, producing no markers and no compile error. Read both sides of every hunk rather than accepting a clean apply.

### What that does to the review

**Every lane verdict on this branch is against `01a16b1379`, and the resolved head will not be that commit.** Five lanes (reuse, correctness, performance, tests, intent) certified `ea632ef5fd`, and the fixes landed after it are covered by targeted mutation testing but not by a lane.

Suggested minimum after resolving: **re-run correctness and tests on the resolved head.** The other three lanes certified things a merge resolution cannot change — reuse of existing helpers, product intent, and query cost. Correctness and tests are the two whose subject is the exact code the resolution touches.

### Verification worth doing on the resolved branch

- `git range-diff origin/main...01a16b1379 origin/main...<new head>` — confirm the commits replayed as intended rather than assuming a clean apply did the right thing.
- `pnpm run db:generate` then `pnpm run db:check-generated`, since the schema conflicts.
- `pnpm run test:lint-rules` — the guard-count resolution lands or fails here.
- Full unit suite **on the rebased branch**, quoting the base SHA beside the file count. Expect 4 pre-existing failures in `scripts/test-perf/__tests__/trace-flush.test.ts`, unrelated to this branch (verified by stashing this branch's `package.json` change and re-running).

### Unchanged and still the loudest line

🔴 **Apply the migration before deploying the code, in every environment.** `notThreadMuted` is inside the `WHERE` of every comment notification processor, so a code-first deploy does not merely disable this feature — **every comment notification query errors on a missing relation and the whole family stops.** `packages/civitai-db-schema/prisma/migrations/20260827150000_thread_mute/migration.sql`, applied to dev only so far.

JustMaier:
## Rebased onto `origin/main`, and re-reviewed on the resolved head

**Head: `5aca359a3e08fad929379aa9b2e2974959cf61d5`.** Base `origin/main b8a45b46d3`. The PR was `CONFLICTING / DIRTY`; it now reports `MERGEABLE`, and CI registers 12 jobs instead of 2.

### The rebase

14 commits replayed. Three conflicted files, all guard-count bookkeeping: `CLAUDE.md`, `.claude/agents/civitai-test-review.md`, `package.json`. `main` added `no-unguarded-billable-submit` and `no-unguarded-user-text`; this branch adds `no-unmuteable-comment-processor`. All three kept. Counts resolved to **22** guards / **27** files, verified against the directory (22 `no-*.test.ts`) and the script (27 paths, no duplicates). `no-lint-rules-script-drift` passes.

`git range-diff` says 13 of the 14 commits replayed byte-identical; the only one marked changed is commit 1, and its whole delta is the counts plus the script list.

Five more files were changed on both sides and produced no markers. `schema.full.prisma` was resolved by hand and the generated artifacts left to `pnpm run db:generate` — `kysely/types.ts` now carries `ThreadMute` alongside main's `Blurb` and without main's removed `AppBlockReview`, so they were regenerated rather than replayed stale. `db:check-generated` is clean.

### Verification

- `pnpm run typecheck` — 0 errors (needs `pnpm install` first; main added `@civitai/generation-metadata` and `htmlparser2`)
- `pnpm run db:check-generated` — clean
- Full unit suite — `Test Files 6 failed | 1543 passed | 3 skipped (1552)`, `Tests 12 failed | 24423 passed | 35 skipped (24470)`
- `blocks.router.workflow.test.ts` collects **370** tests, so the submodule is present and the run means something

**All 12 failures reproduce on clean `origin/main b8a45b46d3`.** Not assumed — each failing file was re-run on a detached checkout of `b8a45b46d3` in the same worktree, with the same failures and the same counts:

| File | Branch | Clean main |
|---|---|---|
| `scripts/test-perf/__tests__/trace-flush.test.ts` | 4 | 4 |
| `src/server/utils/__tests__/credential-detection-superset.guard.test.ts` | 4 | 4 |
| `src/components/Apps/__tests__/appModeratorMessageForm.callSites.test.ts` | 1 | 1 |
| `src/server/services/blocks/tools/__tests__/registry.test.ts` | 1 | 1 |
| `src/server/services/__tests__/no-direct-shared-module-mock.test.ts` | 1 | 1 |
| `src/server/services/__tests__/no-hand-typed-redis-key-constants.test.ts` | 1 | 1 |

Zero failures attributable to this branch. On Linux CI only the last two appear — the other four are Windows-only.

### Re-review: correctness and tests, on the resolved head

The other three lanes certified things a merge resolution cannot change, so they were not re-run. Both lanes were report-only and both reported a clean `git status --porcelain` at start and end.

**Correctness: no defects.** It checked each collision point rather than the diff in aggregate — that `toggleThreadMute`'s lazily-created `Thread` agrees with `upsertComment`'s (and is stricter, deriving `parentThreadId` server-side), that main's ToS badge and the mute menu item are independent insertions with no hook-order change, and that all 13 `notThreadMuted` call sites bind `t` to the comment's own thread. Two low items, both pre-existing and neither blocking, are listed below.

**Tests: two real blind spots, now closed** in `5aca359a`. The code was correct in both cases; nothing could have seen it break.

1. The per-processor sweep matched three fragments of the mute filter but never the operator joining it to the `WHERE`. An `AND`→`OR` at any of the ten call sites without a whole-statement snapshot makes the whole predicate true for an unmuted recipient — past `lastSent`, the self-notify guard, **and the block filter**. Added one `toMatch` covering all 13.
2. `getThreadMuted` reads its row through a cast, not a checked contract, so deleting `bool_or(th.locked) "locked"` leaves `row.locked` undefined, `?? false` makes it `false`, and `canMute` is unconditionally `true`. Every test supplied `locked` from the mock. Added two assertions, the second pinning the `Thread` join as inner — outer, `th.locked` is NULL on an unresolvable hop and `bool_or` skips it.

Each was proved by running the mutant, not by inspection. Three mutants applied one at a time against a file copy (never `git checkout`), each restored and hash-compared to the `HEAD` blob afterwards:

| Mutant | Result |
|---|---|
| `AND ${notThreadMuted(...)}` → `OR` at `comment.notifications.ts:668` | `1 failed \| 29 passed` — caught at the new line 102 |
| delete `bool_or(th.locked) "locked"` | `1 failed \| 36 passed` — caught at the new line 307 |
| `JOIN "Thread" th` → `LEFT JOIN` | `1 failed \| 36 passed` — caught at the new line 308 |

The single failure in each run is the negative control: every pre-existing assertion stayed green, which is what made these blind spots invisible.

### Open, not blocking

Carried forward from the earlier lanes, plus four new ones from this pass. None gates a merge.

- `canMute` (`commentsv2.service.ts:636`) models only the `locked` half of what `throwIfThreadChainLocked` refuses. On an orphaned chain the menu enables a mute that then errors with a toast — a visible divergence, not a silent one.
- A doc comment at `commentsv2.service.ts:277-281` is left dangling above `threadIsRooted` after `MAX_THREAD_CHAIN_DEPTH` moved to `thread-chain.ts`.
- `sectionMuteSchema`'s guard pins two known-bad entity types rather than the invariant it documents ("findable by a single unique column"). The claim holds for all 13 accepted types today; a future composite-keyed type would 500 the endpoint with nothing red.
- `notifications.block-filtering.test.ts:36-58` has the identical connector hole for `notBlockedBetween` — pre-existing, out of this diff, and the pattern this branch's guard was copied from.
- `no-unmuteable-comment-processor` sweeps two notification modules. Complete today (those are the only two mentioning `CommentV2`), blind to a third.
- The client half — `useSectionMute` and the `Comment.tsx` menu — has no tests, and the one browser test rendering that surface mocks `CreateComment` away wholesale.
- Reuse follow-ups (shared toggle body, the thrice-stated `!replyToCommentId` gate, an `entityThreadWhere` helper) and perf's `Prisma.raw` planning cost (~0.5 ms per menu open, unamortised; the 13 processors are unaffected — their seed is a column reference).

### 🔴 Before deploying

**Apply the migration first, in every environment.** `packages/civitai-db-schema/prisma/migrations/20260827150000_thread_mute/migration.sql`. `notThreadMuted` sits inside the `WHERE` of every comment notification processor, so a code-first deploy does not disable the feature — every comment notification query errors on a missing relation and the whole family stops. New table, no backfill, safe to apply ahead of the deploy. Applied to **dev** only so far.

PR #4528, opened 2026-08-31
fix(blocklist): run one content check on every user-authored surface

JustMaier:
Closes the gap in ClickUp 868kw6t61. (868kw2f8y was already fixed by #4350 — comments and the homoglyph handling shipped there on 2026-08-24.)

Comments have run a combined content check since #4350 — both blocklists, over several normalised forms of the text. Every other surface that stores user-authored text called the link-domain helper and stopped there, so it enforced one of the two lists, over the raw stored string, with no normalisation.

This gives those surfaces the same check, from one implementation rather than nine copies.

## What changed

`throwOnBlockedUserContent` now runs at every writer of user-authored text across 17 files: model, model version, article (including rating-review comments), post (create *and* update), bounty, bounty entry, resource review (all three writers), blurb, app listing, Model3D and its reviews, collection (including item notes), user hub (including source aliases), creator announcements, profile bio/announcement/links, cosmetic shop, and exploration prompts.

That includes the fan-out re-checks that run *after* other content is spliced into a body — those are separate writers, and a fix that only patches the entry point misses them.

Many of these had **no check at all**, not even the link-domain one.

## The two halves ship at different strengths, deliberately

- **The link half throws**, as it already did. Everything this adds to it — scanning the derived forms, normalising — can only ever match *more*, never less, so it carries no new false-positive risk and needs no ramp. Anything weaker here would be a regression on a control that is live today.
- **The pattern half records rather than blocking**, behind `USER_CONTENT_PATTERN_ENFORCE`, which ships off. It is new to these surfaces, and a false positive costs a creator a publish rather than one comment, with no moderator override on that path. The flag is a one-flip change to enforcing.

**Both branches log, on purpose.** Recording only the un-enforced hits would delete the signal at the exact moment the flag is flipped — and comparing the rate either side of that flip is the whole reason the rollout is staged. A rollout that cannot be measured is not staged, only late.

Moderators are exempt from the pattern half only. The link half runs for everyone, exactly as before this PR. (An earlier revision of this branch exempted them from both; that would have switched off a live control, and there is now a test that fails on revert naming the domain.)

**Values are scanned separately, never pre-joined.** A pattern spanning the seam between two independent fields would match text no user ever wrote, and a creator would be blocked by the accident of what their title happens to end with. One call site quietly undid this by passing `title` and `body` already joined; an invariant enforced in one place and violated at the call site is not an invariant, so the argument is now pinned by an assertion.

## The guard test

`no-unguarded-user-text` pins each surface, with per-writer counts attributed by enclosing function, so a call cannot be deleted, reverted, or moved between writers unnoticed. It asserts on the *banned* symbol as well as the required one: the dangerous edit here is not a deletion but a swap, since the link-only function is still exported and correct for its own job, so reverting a call site to it type-checks and reads in review as a tidy-up.

It reads the file rather than trusting a grep, and that is not theoretical: **it caught a corrupted import in one of my own fix rounds before the typechecker did**, and named the file.

**Its docstring states what it does not prove.** It certifies that these calls are present and attributed to these writers — not that these are all the writers, and not that a writer scans every field it persists. It counts calls, not arguments. Three review passes each found writers the previous one missed, so the enumeration is assumed open and tracked in ClickUp 868kz1yt1.

## Not in this PR

- The remaining unguarded writers (comics router, Q&A, creator shop, reports and appeals, and others) — 868kz1yt1, with a finite closing condition. Two of them are policy calls rather than code ones.
- A model version's **name** is scanned nowhere. Model names are; versions are not, because two review handlers replay the stored row through the same upsert with no moderator flag, so scanning it there could make a version impossible to decline. The reason and a warning against the naive fix are recorded at the call site.
- The per-save cost of the new scan is unmeasured against large real bodies. The pattern half is off by default, so this lands before it matters — recorded on 868kz1yt1 as a precondition on the flag flip rather than a nice-to-have.

## Verification

- `typecheck` 0 errors.
- `eslint` 0 errors in the touched files. (Repo-wide is red on `main` for unrelated reasons; not claiming that number.)
- Full unit suite green apart from four files that fail identically on plain `origin/main` — confirmed by running them there, not by assuming.
- `blocks.router.workflow.test.ts` collected 370 tests, so the run means something.
- Every behavioural claim above has a mutation behind it: the fix reverted, the suite run, the failure read. Including the ones that matter most — the moderator/link exemption, the seam, and moving a guard between two writers.

## Migrations

None.


JustMaier:
## Not merging yet — outstanding findings from the sixth review pass

CI is green at `557cf2e037` (14 checks: 13 pass, 1 skipping) and the branch is `MERGEABLE`. Holding it anyway.

**Why:** of the four changes in the previous round, **two were wrong** — one deleted a check that was live, and one would have made a moderation takedown impossible. Both were made on a review's say-so. A branch where the fix rounds are introducing defects at that rate should get a rested read rather than another late iteration, so this is parked deliberately rather than because anything is broken.

The last pass was run to **collect only**. Nothing below is fixed. Listing it here so the next session starts from the findings rather than rediscovering them.

### Highest severity — a guard that can block the people who clean up after it

1. **`updateUserProfile` does not thread `isModerator`.** The controller explicitly routes a moderator to another user's profile, and the profile form round-trips every stored text field. So a moderator opening a spam profile to clear it can have the save refused because of a field they were not editing, with no path through that service to clear it. Every other moderator-reachable writer in this diff threads the flag; this one is the exception.

2. **The blurb fan-out is a takedown path this guard can wedge.** The five `apply*ContentChange` writers run with no user in scope and no moderator flag, and a throw there is not surfaced to an operator — the entity is retried indefinitely. Taking down a snippet leaves its text spliced into every referencing entity, and the guard is then what refuses the write that would have propagated the takedown. The link half already behaved this way before this branch; what this branch adds is more forms to match on, plus the pattern list once the flag is enabled.

This is the same class as the `trainedWords` defect caught and reverted in `557cf2e037`: **a guard the cleanup crew cannot escape is worse than the gap it closes.** It deserves a deliberate decision, not a patch at the end of a long session.

### The guard test

3. **The last commit weakened both banned-symbol assertions.** Gating them on the call rather than the import fixed a real false-positive with namespace imports, but dropped the import signal entirely — so aliasing the weaker helper to a local name now passes. Demonstrated by A/B against the previous commit, where it failed. The correct shape is to require a call only for the namespace form and keep firing on the import for named imports.

4. **A test certifies a branch that cannot currently fire.** The app-listing category test mocks the guard so it exercises a path the real function only reaches when the enforcement flag is on. It would stay green if the flag gate were removed.

### Consistency

5. **Guard-versus-authorization ordering is now inconsistent.** Two call sites were moved after their ownership checks; roughly seven others in the same diff still run before theirs. Whatever the right convention is, the diff currently has both, with a comment on each of the two that changed.

6. **Two stacked justification comments** on one call in `model-version.service.ts`, from consecutive rounds, with no marker saying which is current — in exactly the place a future reader will look before touching that argument list.

### Scope notes, tracked on ClickUp 868kz1yt1

7. Several guarded writers persist user text the guard does not scan — tag names created by four of the listed surfaces are the notable case. The ledger explicitly disclaims field-level coverage, so this is scope rather than regression, but it is the same hole that produced the earlier `alias` miss.

8. Both blocklists are read and both matchers built on every guarded write, including when the caller is exempt. `upsertArticle`, `upsertModel` and `upsertModelVersion` each run the scan twice per save. Measured cost of the normalisation step alone: 2.4ms on a 5.6KB body, 6.9ms on 56KB, 25.5ms on 224KB. Not fatal on a write path, but new work on every write.

9. The enforcement flag cannot be exercised through the local-override mechanism, which is worth knowing for a rollout whose design is to compare the rate either side of the flip.

### An instrument failure worth recording

During the fifth pass, several `sed`/`grep` reads through the shell returned **stale file content** — an older revision of a file that had been modified minutes earlier. It was caught because the findings were re-verified against committed blobs, and everything load-bearing was re-read another way. Neither I nor the reviewer can explain it. Flagging it because every enumeration in this work rests on those tools, and a stale read is indistinguishable from a correct one.

JustMaier:
## Five-lane review at `7a7ae21415` — four lanes reported, one being chased

Branch `fix/pattern-filter-coverage`. Every lane reported the SHA it read and `git status --porcelain` at start and end; all four found the tree clean at both ends.

**Since the last comment:** the blocker is fixed. The convention guard's banned-symbol check is now split by import kind — a named import fires on the import alone (the local name proves nothing, since the dodge is `const check = <weaker helper>`), and only a namespace import requires finding the call. Controlled in **three** directions this time: alias dodge fails, namespace dodge fails, correct namespace use passes. The previous control tested only the case just fixed, which is why it missed the case just broken.

Also: `updateUserProfile` now threads `isModerator`, closing the second instance of a guard blocking the cleanup it exists to prompt. Its controller deliberately routes moderators to other users' profiles.

### Reuse — clean

Nothing rebuilt. The new code builds on the existing tag-stripping and folding helpers rather than re-deriving them, and the consolidation is complete: the weaker helper survives in exactly three files, all three listed as stated exceptions with reasons. The two-wrapper split was checked and judged justified rather than under-consolidation — the wrappers are fixed presets over one shared function with two independent exemption flags.

Map contribution, not a request: this is the first guard in that directory to walk the TypeScript AST; the other ~20 are string scanners. No helper existed to reuse, and extracting one now would have a single consumer.

### Intent — delivers, with one claim corrected

All eight surfaces from the ticket's table are covered, no silent drops, and the false-positive decision the ticket asked to be *decided* is recorded at the call site rather than left in a thread.

It challenged a claim in my own comment: that widening the link check "can only match more, never less, so it carries no new false-positive risk." Those are two different statements and I had asserted the stronger one. Tag-stripping joins text with no separator, so it can produce a token that was never contiguous — and for an exact-host comparison that is a false positive rather than broader coverage.

**Measured rather than assumed:** the specific example does not reproduce. The link pattern requires `//`, so a bare glued host matches nothing. The residual case needs the glue to manufacture the slashes, which means the source already held a link split by markup — a true positive. Narrow, not absent. The comment now states what was measured and warns against re-deriving the stronger claim. Fixed in `7a7ae21415`.

Scope note: the work covers ~20 writers against the ticket's eight, including surfaces that previously had no check at all. Grounded in the same rationale, but it does widen the blast radius of anything in the link half.

### Tests — one confirmed false certification, one real gap

1. The app-listing category test wholesale-mocks the guard, so its "pattern" case certifies classification wiring only — it does not certify that a pattern hit can reach that path in production, which today needs the flag on. Confirmed by mutation: deleting the flag check leaves the test passing, because it never calls through that code. The lane checked the two adjacent seams and found no second instance of the shape.

2. **Real coverage gap:** nothing exercises the router → `subjectUser.isModerator` → guard wiring on the app-listing path. A regression hardcoding that flag to `false`, or dropping it, would pass every test in this diff. The guard's own honouring of the flag is well covered; the wiring into it is not.

The ledger was checked against source — all five "splice-and-recheck" entries genuinely are that shape, and no entry contradicts the stated invariant. The ~20 mock-factory edits are sound, and the five rewritten wiring assertions were traced against their fixtures and are equivalent in strength to what they replaced.

### Performance — one structural finding

All affected call sites are **writes**; nothing touches the feed or read path, and the diff adds no SQL, no N+1s and no new client-bundle edge (verified by running the app-graph guard, not just reading it).

**The finding:** `content`/`description` on article, model, model version and bounty have **no length cap**. The only bound is the tRPC body limit, so a large body reaches the synchronous normalise-and-scan loop — and does so twice per save on three of those surfaces, which is deliberate and documented but doubles the cost. The App Blocks surface caps its equivalent fields before the same guard; these did not. Measured for the normalisation step alone: 2.4ms at 5.6KB, 6.9ms at 56KB, 25.5ms at 224KB.

Secondary, acceptable now: the two guard calls per save are not coalesced, so those surfaces do four cache reads rather than two, and the cache-miss path has no stampede lock — low severity, as the underlying table is small.

One worry of mine it **disproved**: a moderator on the comment path does not build a matcher it never uses. That path exits before either list is fetched.

### Outstanding

The correctness lane has not reported and is being chased rather than recorded as clean. Its findings will follow.

Not fixed in this comment's SHA: the two test items above, and the input-size ceiling. They are listed here so the next session starts from them.

PR #4530, opened 2026-08-31
fix(buzz): keep cash-outs out of the dashboard chart

JustMaier:
A creator's cash-out was drawn in the same `spent` series as every other debit on the Buzz dashboard chart. At creator volume one of those bars is thousands of times the next, so the axis scales to it and every other bar renders under a pixel — the chart stops being readable at the point it should be most useful.

Reported by MNeMiC in the creator group DM (2026-08-27), ClickUp 868kz045v:

> I don't need to see my cash-out as a purchase. It really breaks any usability of the charts... Basically, I want it to be my earnings.

## How bad it was, and how much this fixes

Simulating the chart's own series — yellow Buzz in and out, per day — across the **89 creators who cashed out in the last 30 days**, tallest bar over median day:

| | median | p90 |
|---|---|---|
| as shipped | 39.5× | 489× |
| with cash-outs excluded | **4.1×** | **54×** |

On MNeMiC's own worst case the single bar was **15,152×** his median earning day (2,000,000 ÷ 132, July 2024). At 300px tall his median day drew 0.02px.

## Why the chart had to move to ClickHouse

`GET /user/<id>/transactions/report` on the Buzz service returns pre-aggregated `{ spent, gained }` per bucket, with no transaction type and no type parameter — so the exclusion could not be expressed against it at all, on our side. Justin approved the ClickHouse swap explicitly. `default.buzzTransactions` carries the type, and the ingest lag measured 34s.

The new query produces the same shape the chart already consumes; `BuzzDashboardOverview` is untouched, as are its windows and its account-type switcher.

## Which types, and why the exclusion is on both branches

`bank`, `withdrawal` and `extract`, excluded from the **gained** branch as well as the **spent** one, spelled through `TransactionType` rather than hand-typed so the ClickHouse values stay derived from the enum.

Both branches, because the direction is not stable and the live case is on the credit side:

- **`bank`** is the cash-out that debits yellow — 443 rows / 50.7M Buzz in 30 days, largest single 6,991,121. It never credits a spend account (`toAccountType` is always `creatorProgramBank`), so excluding it on the gained side strips nothing.
- **`extract`** credits yellow/green *out of* the creator program bank — **50 rows in 90 days, every one to a real user**, yellow max 100,000, green max 109,416, most recent today. This is the one that flattens the Gained bar, and a spend-side-only guard cannot see it.
- **`withdrawal`** debited yellow until 2025-02-27. Since 2025-03 it credits **account 0**, the bank, in **1,215 of 1,223** rows; only 8 have ever credited a real user, the last on 2025-08-07. So it fires on neither branch for a live account today, and is kept for the historical direction only. Stated precisely because the first review of this PR read those rows as creator-facing credits of up to 392,179 — they are account-0 rows.

A first pass of this PR excluded on the spent branch alone. Its worked example was July 2024, when a cash-out still debited yellow, so the control passed while the live case went unguarded.

## Membership grants need no change

MNeMiC asked for two exclusions. The second one was never in the picture: membership-granted Buzz is paid in **green** (4,995 of 4,995 rows in 30 days), and the chart is per-account-type with yellow as the default, so it never appeared in the view he was looking at. On the green view its distortion measures 1× at the median, 2× at p90.

## Verification

**Real-data control, the credit side** — the live case. User 5197157, gained branch, 2026-07-29 → 08-01: a **100,000** `extract` credit is the only row in that window, and it is gone once the predicate applies. Nonzero without it, empty with it.

**Real-data control, the debit side.** The generated SQL run against production ClickHouse for MNeMiC's account, 2024-07-12 → 2024-07-15, with and without the exclusion:

| day | spent, excluded | spent, control |
|---|---|---|
| 2024-07-13 | 5,000 | 5,000 |
| **2024-07-14** | **0** | **2,000,000** |

The cash-out goes; the legitimate 5,000 spend two days earlier survives. Real rows, both directions.

**Unit tests** — `src/server/services/__tests__/buzz.service.transactions-report.test.ts`, 6 tests. Both controls run by reverting the code and confirming a legible failure:

| mutation | failure |
|---|---|
| exclusion removed from the spent branch | `expected '...' to contain 'AND type NOT IN (...)'` — 1 failed \| 5 passed |
| exclusion removed from the gained branch | same assertion, other branch — 1 failed \| 7 passed |
| empty buckets dropped from the fill loop | `expected [...] to have a length of 8 but got 2` — 2 failed \| 4 passed |
| month window back to a calendar year | `expected [...] to have a length of 12 but got 13` — 1 failed \| 7 passed |
| fallback removed, throws again | `TRPCError: nope` — 1 failed \| 7 passed |

The exclusion assertions name the whole predicate including the operator, so a `NOT IN` → `IN` flip — which would draw only cash-outs — fails them too.

## Two more defects the review found

- **The 12-month window emitted thirteen buckets.** `BuzzDashboardOverview` keys each dataset by a `MMM-DD` label, so a year-back start put `Aug-01` at both ends: one value overwrote the other and a bar went missing, with no error anywhere. The sequence now counts whole buckets back from the last one, which also stops the first bucket being a partial period drawn as a whole one.
- **A ClickHouse query failure threw.** `useTransactionsReport` turns a thrown report into an empty array, which the chart renders as "no data on the provided timeframe" over a creator's real activity. It now falls back to the buzz service, exactly as it does when ClickHouse is unconfigured — degrading to showing the cash-out, which is what the chart did before this branch.

## A defensive catch that relabels a handled error is worse than no catch

Worth stating on its own, because it is general rather than a quirk of this file.

The ClickHouse read falls back to the buzz service, so the fallback needed a `catch` for the case where both are down — otherwise the raw client error reaches the user, since `getTransactionsReportHandler` is a synchronous function around an async call and its own `try/catch` never sees the rejection.

The first version of that catch wrapped **everything**. But `buzzService`'s `mapError` has already converted every non-2xx into a proper `TRPCError`, carrying the `BuzzApiError` as `cause`. So a buzz-service 404 arrived handled and left as an `INTERNAL_SERVER_ERROR`: a client fault counted in the 5xx SLO, a full server-fault stack ingest, and the HTTP status buried one level below where `getBuzzApiStatus` looks for it — `error.cause instanceof BuzzApiError` is now one hop too shallow.

The catch did not swallow the error or change the code's behaviour. It corrupted the signal the alerting is built on, which is harder to notice and worse to live with. It now rethrows a `TRPCError` untouched and wraps only what is genuinely unhandled.

## What this does not verify

I did not read the Buzz service's own code to confirm how it classified a cash-out, because that service is outside this repo. The rewrite makes the question moot: the classification is now in this file.

## Scope not taken

Creator Studio's `/earnings` page has its own earnings definition that drops contest prizes, donation-goal donations, bounty awards, and older access sales whose `externalTransactionId` is empty. That is a separate ticket about that page, not this chart.

https://claude.ai/code/session_012bWrumgzC562dc6rpxFqnq

### DIFF
--- a/src/server/services/__tests__/buzz.service.transactions-report.test.ts
+++ b/src/server/services/__tests__/buzz.service.transactions-report.test.ts
@@ -8,8 +8,10 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
  * pixel. Measured across the 89 creators who cashed out in a 30-day window, tallest bar over median
  * day falls 39.5x -> 4.1x once `bank` and `withdrawal` are out.
  *
- * `bank` is the one that matters and is the one an author reaching for the obvious word misses:
- * 443 rows / 50.7M Buzz in 30 days against `withdrawal`'s 52 rows / 2.2M.
+ * Two types reach a creator's chart, on opposite sides: `bank` debits yellow/green (the Spent bar,
+ * 2,174 rows / 180d, max 10,477,563) and `extract` credits them back out of the program bank (the
+ * GAINED bar, 79 rows / 180d, max 500,000). `withdrawal` credits account 0 rather than a creator and
+ * reaches neither today; it debited yellow until 2025-02-27 and is kept for that.
  */
 
 import type * as ClickhouseClient from '~/server/clickhouse/client';
@@ -55,19 +57,19 @@ afterEach(() => {
 
 const sqlOf = () => String($query.mock.calls[0][0]);
 
-// The two halves of the UNION, separated so an assertion can name which branch it is about. A fix
-// that excluded cash-outs from BOTH branches would strip the creator's incoming rows too.
+// The two halves of the UNION, separated so an assertion can name which branch it is about.
 const branches = (sql: string) => {
   const [gained, spent] = sql.split('UNION ALL');
   return { gained, spent };
 };
 
 describe('getTransactionsReport', () => {
-  // 🔴 BOTH branches, deliberately. A `withdrawal` debited yellow until 2025-02-27 and credits it now
-  // (cashSettled -> yellow), and `extract` credits yellow/green out of the program bank — so a cash-out
-  // reaches the chart through the GAINED branch today. Excluding on the spend side alone is a guard
-  // that cannot fire for any window this chart offers. If you are here to "simplify" this back to one
-  // branch, that is the bug, not the redundancy.
+  // 🔴 BOTH branches, deliberately, and the gained one is not redundant: `extract` credits yellow or
+  // green out of the creator program bank, so a cash-out returning reaches the chart through GAINED
+  // and no other filter would remove it. The first version of this fix guarded only the spend branch
+  // and its worked example was July 2024, when a cash-out still debited yellow — the control passed
+  // over a live case it could not see. If you are here to drop the gained-branch filter, check
+  // `extract`'s direction in production first.
   it('excludes cash-out plumbing from the spent series', async () => {
     await getTransactionsReport({ userId: USER, window: 'day', accountType: 'yellow' });
 
@@ -175,11 +177,50 @@ describe('getTransactionsReport', () => {
     // result, which the client renders as "no data on the provided timeframe" over real activity.
     expect(getUserTransactionsReport).toHaveBeenCalledOnce();
     expect(report).toHaveLength(1);
-    expect(report[0].accounts[0].gained).toBe(20);
+    // Both fields the chart keys on, not just the value: it looks the bucket up by a formatted `date`
+    // and picks the account by `accountType`, so a fallback returning either in an unexpected shape
+    // draws every bar at zero while an assertion on `gained` alone stays green.
+    //
+    // The format, not the instant. This path runs the buzz service's payload through
+    // `getTransactionsReportResultSchema`, whose `z.coerce.date()` reads a bare timestamp as LOCAL and
+    // then formats it as UTC — so the value shifts by the host's offset (nothing on a UTC server, six
+    // hours on the box this was written on). Pre-existing on this path; pinning the literal would make
+    // the test fail by timezone rather than by defect.
+    expect(report[0].date).toMatch(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}$/);
+    expect(report[0].accounts[0]).toEqual({ accountType: 'yellow', spent: 10, gained: 20 });
+  });
+
+  it('maps a non-yellow account type through the fallback', async () => {
+    $query.mockRejectedValue(new Error('clickhouse is down'));
+    getUserTransactionsReport.mockResolvedValue([
+      {
+        date: '2026-08-31T00:00:00',
+        start: '2026-08-31T00:00:00',
+        end: '2026-09-01T00:00:00',
+        accounts: [{ accountType: 'Green', spent: 0, gained: 5 }],
+      },
+    ]);
+
+    const report = await getTransactionsReport({
+      userId: USER,
+      window: 'day',
+      accountType: 'green',
+    });
+
+    expect(report[0].accounts[0].accountType).toBe('green');
+  });
+
+  it('surfaces a handled error when the fallback fails too', async () => {
+    $query.mockRejectedValue(new Error('clickhouse is down'));
+    getUserTransactionsReport.mockRejectedValue(new Error('buzz service is down'));
+
+    await expect(
+      getTransactionsReport({ userId: USER, window: 'day', accountType: 'yellow' })
+    ).rejects.toThrow('Could not load the Buzz report right now. Please try again shortly.');
   });
 
   it('buckets weeks on Monday, matching ClickHouse toMonday', async () => {
-    // 2026-08-31 is a Monday; the `week` window reaches back one month.
+    // 2026-08-31 is a Monday; the `week` window reaches back five whole weeks.
     $query.mockResolvedValue([{ bucket: '2026-08-31', gained: '500', spent: '0' }]);
 
     const report = await getTransactionsReport({
@@ -189,6 +230,10 @@ describe('getTransactionsReport', () => {
     });
 
     expect(sqlOf()).toContain('toMonday(date)');
+    // Six, the same span a calendar `subtract(1, 'month')` covered before the sequence was rewritten.
+    // Nothing else pins the week window's length, and it lost a period once already.
+    expect(report).toHaveLength(6);
+    expect(report[0].date.slice(0, 10)).toBe('2026-07-27');
     const last = report[report.length - 1];
     expect(last.date.slice(0, 10)).toBe('2026-08-31');
     expect(last.accounts[0].gained).toBe(500);
--- a/src/server/services/__tests__/buzz.service.transactions-report.test.ts
+++ b/src/server/services/__tests__/buzz.service.transactions-report.test.ts
@@ -18,9 +18,12 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import type * as ClickhouseClient from '~/server/clickhouse/client';
 import type * as BuzzClient from '@civitai/buzz';
 
-const { $query, getUserTransactionsReport } = vi.hoisted(() => ({
+const { $query, getUserTransactionsReport, state } = vi.hoisted(() => ({
   $query: vi.fn(),
   getUserTransactionsReport: vi.fn(),
+  // The `!clickhouse` branch is unreachable behind a mock that always supplies a client, so the export
+  // is a getter and the tests decide per case whether ClickHouse is configured at all.
+  state: { clickhouseConfigured: true },
 }));
 
 // Spread the real package and override only the client factory, so this file is not coupled to every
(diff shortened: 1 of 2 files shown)


JustMaier:
## Review round — three findings, all real, all fixed at `8e25ae3ab9`

**1. The guard was on the wrong branch.** A cash-out does not always leave the spend account, so excluding only from `spent` left the same defect on the Gained bar. Corrected on both branches, set widened to `bank`, `withdrawal`, `extract`.

I checked the review's own numbers before acting on them, and one half does not hold:

- `withdrawal` since 2025-03 credits **account 0** (the bank) in **1,215 of 1,223** rows. Only 8 credit a real user, the last on 2025-08-07. So it reaches no live creator's Gained series — the 392,179 figure is an account-0 row. It stays in the set for the historical direction (it debited yellow until 2025-02-27).
- `extract` is the live vehicle: **50 rows in 90 days, every one to a real user**, yellow max 100,000, green max 109,416, most recent today. That one genuinely flattens Gained, and the first pass would have shipped without it.

Real-data control, user 5197157, gained branch, 2026-07-29 → 08-01: a **100,000** `extract` credit is the only row in the window and is gone once the predicate applies.

**2. The 12-month window emitted thirteen buckets.** `BuzzDashboardOverview` keys each dataset by a `MMM-DD` label, so a year-back start put `Aug-01` at both ends — one value overwrote the other and a bar went missing. The sequence now counts whole buckets back from the last one, which also stops the first bucket being a partial period drawn as a whole one.

**3. A ClickHouse query failure threw**, and `useTransactionsReport` turns a thrown report into an empty array, which the chart renders as "no data on the provided timeframe" over a creator's real activity. It now falls back to the buzz service, exactly as it does when ClickHouse is unconfigured.

### Controls, re-run after the fix round

| mutation | failure |
|---|---|
| gained-branch exclusion removed | `expected '...' to contain 'AND type NOT IN (\'bank\',\'withdrawal\',\'extract\')'` — 1 failed \| 7 passed |
| month window back to a calendar year | `expected [...] to have a length of 12 but got 13` — 1 failed \| 7 passed |
| fallback removed, throws again | `TRPCError: nope` — 1 failed \| 7 passed |

`typecheck: OK — 0 type errors`. Full suite `Test Files 4 failed | 1530 passed | 2 skipped (1536)`, `Tests 10 failed | 24114 passed | 28 skipped (24152)`, exit 1 — the same 4 files and same 10 tests fail on clean `origin/main` (`3ba9917c9f`: `Test Files 4 failed (4)`, `Tests 10 failed | 99 passed (109)`), a Windows-only `C:\C:\` path bug and neighbours.

### Not refuted by the review, and worth keeping visible

Alias shadowing, projection selection (`byToAccount`/`byFromAccount` both selected, 293ms), UTC bucket keying across three server offsets, bucket-string parsing, and the emitted shape against the old path were all checked and held.

JustMaier:
## Second review round — fixed at `c28e8e4be5`

The re-review was pointed at the fix round specifically. It found seven things; five changed code, two are recorded as deliberate.

**1. The comment justifying the two-branch exclusion was wrong, and load-bearing.** It repeated the first review's claim that a modern `withdrawal` credits a creator's yellow account — a claim I had already checked and disproved, corrected in this PR body, and then left standing in the code. Rewritten to what production actually shows over 180 days:

| type | from → to | reaches a creator | rows | max |
|---|---|---|---|---|
| `bank` | yellow/green → creatorProgramBank | Spent bar | 2,174 | 10,477,563 |
| `extract` | creatorProgramBank → yellow/green | **Gained bar** | 79 | 500,000 |
| `withdrawal` | cashSettled → yellow, `toAccountId = 0` | neither | 342 | 1,121,934 |

So `extract` is the entire reason the gained branch is filtered; `bank` and `withdrawal` on that side are inert. The test comment carried the same false claim and told a future reader not to act on it — corrected to point at `extract` and to say "check its direction in production first".

**2. The bucket-sequence rewrite silently shortened the week window.** `subtract(1, 'month')` snapped to Monday spans **six** buckets; `subtract(4, 'week')` spans five. The rewrite was justified only by the `MMM-DD` label collision, which is a `month`-window problem — the week view lost a period for no stated reason, and nothing pinned it. Restored to six, with the count and first bucket now asserted.

**3. Both sources failing surfaced a raw upstream error.** `getTransactionsReportHandler` is synchronous around an async call, so its `try/catch` never sees the rejection. The fallback now throws a handled `TRPCError` itself.

**4. The fallback test pinned only `gained`.** The chart also keys on `date` and picks by `accountType`; a fallback returning either in an unexpected shape draws every bar at zero with that test still green. Both are asserted now, plus a non-yellow account type.

While writing that assertion it surfaced a real quirk, pre-existing on the fallback path: `getTransactionsReportResultSchema`'s `z.coerce.date()` reads a bare timestamp as **local** and formats it as **UTC**, so the fallback's dates shift by the host's offset (nothing on a UTC server, six hours on the box this was written on). Pinned by format rather than by literal so the test fails on a defect rather than on a timezone.

### Deliberate, not fixed

- **A ClickHouse outage is invisible in the UI.** The fallback degrades to the pre-exclusion chart — the cash-out returns and the page looks normal. That is the intended trade (a working chart beats an error), which means the only signal is the Axiom line `name: 'buzz-transactions-report'`. Said so in the code so it gets alerted on rather than watched for.
- **`purchase` credits are left in the Gained series**, and they are larger than `extract` (max 3,549,405). They are buzz the creator bought or was sold — a real gain on a gained/spent chart, and for a creator selling access it is their revenue. Excluding them would be an earnings definition, which is explicitly out of scope here.
- **The fallback returns an unfilled series**, so a ClickHouse hiccup can change the bar count between two refreshes. That is the shape the chart received before this branch, so it is not a regression; filling it would mean re-bucketing another service's response.

### Controls, re-run after this round

| mutation | failure |
|---|---|
| week window back to 4 weeks | `expected [...] to have a length of 6 but got 5` — 1 failed \| 9 passed |
| friendly both-down error removed | `expected [Function] to throw error including 'Could not load the Buzz report…' but got 'buzz service is down'` — 1 failed \| 9 passed |

`typecheck: OK — 0 type errors`. Full suite `Test Files 4 failed | 1530 passed | 2 skipped (1536)`, `Tests 10 failed | 24116 passed | 28 skipped (24154)`, exit 1 — the same four files and same ten tests fail on clean `origin/main` (`3ba9917c9f`), a Windows-only `C:\C:\` path bug and neighbours. The passing count is up by exactly the tests this branch adds.

JustMaier:
## Rebased onto current `main`, and a third review round — head is now `ac208a22ee`

⚠️ **Force-pushed.** `c28e8e4be5` and everything before it are gone; review this SHA.

### Rebase

The branch was **9 commits behind** `main` (`3ba9917c9f` → `e0a0b9bc3e`). Every earlier suite run in this PR answered "does this work on the tree I branched from", not "on the tree it lands in". Rebased clean, no conflicts, and the diff is byte-identical across it — `git diff c28e8e4be5 3e6bd91884 -- <the two files>` is empty, with a positive control against `8e25ae3ab9` returning 80 lines so the invocation is known to work.

Re-verified on the rebased branch, and the pre-existing failures re-controlled against **current** `main` rather than the old base:

| | this branch | clean `e0a0b9bc3e` |
|---|---|---|
| Test Files | 4 failed \| 1533 passed \| 2 skipped (1539) | 4 failed (4) |
| Tests | 10 failed \| 24157 passed \| 28 skipped (24195) | 10 failed \| 99 passed (109) |

Same four files, same ten tests — the Windows-only `C:\C:\` path bug and neighbours. The file count moved 1536 → 1539 because `main` added three test files, which is the point of re-measuring it against the base it lands on.

`typecheck: OK — 0 type errors`.

### Third review round

**1. The fallback relabelled mapped buzz-service errors.** `mapError` already turns every HTTP status into a `TRPCError` carrying `BuzzApiError` as `cause`, so my blanket wrap made a 404 an `INTERNAL_SERVER_ERROR` — counting a client fault in the 5xx SLO — and buried the status one level below where `getBuzzApiStatus` reads it. Mapped errors now pass through untouched; only transport-level failures (refused, aborted, DNS), which the synchronous handler genuinely cannot wrap, get the friendly message.

**2. My week-window comment was false.** It said 6 buckets matched what the old calendar subtraction gave. Enumerated over 1,200 days, the old code gave **5 on 790 days and 6 on 410** — five was the common case. Six is still the right pin, because a window whose length depends on today's weekday is worse than either number, but that is the reason and it is now what the comment says. A pinned number whose stated rationale a reader cannot reproduce is worse than no pin.

**3. The `withdrawal` entry gave a reason that argues for deleting it.** "It credits the central bank, not a creator" describes the 180-day slice only: the type has also credited a creator's yellow out of the program bank as recently as 2025-08-07 (8 rows, max 5,233,644), and debited it directly until 2025-02-27. Restated to what actually justifies keeping it — otherwise the next reader follows the neighbouring instruction to check direction in production, reads this line, and removes the guard.

New control: the pass-through removed → `expected TRPCError: Could not load the Buzz report… to be TRPCError: Not found`, 1 failed | 10 passed. Tests `11 passed (11)`.

### Confirmed by this round, not changed

The exclusion table's production figures were re-checked independently and hold exactly: `bank` 2,174 rows / max 10,477,563; `extract` 79 rows / max 500,000; `withdrawal` `toAccountId = 0` on 341 of 341; `purchase` max 3,549,405. The week change touches only the week arm — `day` stays 8, `month` 12, `hour` unchanged.

PR #4531, opened 2026-08-31
fix(tests): unzero the `preview / component-tests` tier, and make a zero-collected run say so

ZacxDev:
## What

`preview / component-tests` was reporting **`failure` on `main` while executing ZERO tests**. Two things: the one-line fix, and the guard that makes the next occurrence say so out loud.

## The defect

Reproduced deterministically at `d353f785c3` (3/3 full runs, and in isolation): the whole `component` project aborts before any reporter prints — no `Test Files` line, no `Tests` line, no per-file results, exit 1.

**One file.** `src/tests/pages/apps/review/review-queue-nav.browser.test.tsx:30` mocked `~/providers/FeatureFlagsProvider` with a **wholesale** factory naming only `useFeatureFlags`. The review queue page's row now renders the review entry point, which reads flags through `useOptionalFeatureFlags`, so the named import had nothing to bind to:

```
SyntaxError: The requested module '/src/providers/FeatureFlagsProvider.tsx'
does not provide an export named 'useOptionalFeatureFlags'
```

In the node `unit` project that fails **one** file. In **browser** mode it kills the run: vitest resolves a manual mock over the browser↔node channel inside a Playwright route handler that does not catch (`@vitest/browser-playwright/dist/index.js`, `await module.resolve()` inside `page.route`), so the rejection escapes as an `Unhandled Rejection` in the orchestrator.

**The printed error names neither the file nor the cause.** It is wrapped twice — once by the browser mocker, once again on the node side — and the innermost `cause` is dropped in transit, leaving only the generic *"[vitest] There was an error when mocking a module … make sure there are no top level variables inside"*, whose advice about hoisting is wrong here. The real cause was recovered by temporarily patching `createHelpfulError` in the browser tester bundle to inline `cause.stack`, then confirmed by bisecting 50 candidate files down to one.

This is the **second** time this class has taken the tier out — see the header of `src/components/AppBlocks/__tests__/featureFlagsMockCompleteness.test.ts`, which fixed six sibling suites and deliberately scoped its guard to one directory.

## Why the guard matters more than the fix

The tier is the shared `npm-report-only-suite` Tekton task, which computes its verdict from the runner's **exit code alone**. So an abort that executed nothing and a genuine list of red assertions both render as `component:fail` / "Component suite failed" — same words, same colour, same place. Nothing anywhere could say "this run tested nothing".

`pnpm test:component` now runs through `scripts/test-component-run.mjs` → `scripts/ci/assert-component-suite-ran.mjs`, which prints a ledger and applies two checks:

```
test:component ledger: 2254 executed, 0 skipped, across 201 files; 0 failed suites,
0 failed tests (baseline 2254 tests / 201 files measured 2026-08-31; 201 on disk; floor 1240)
```

1. **Zero collected fails**, unconditionally — including on a narrowed run, because a single-file run that collects nothing is the cheapest reproduction of the abort and is exactly when someone is debugging it.
2. **A file on disk that is absent from the report fails, and is NAMED.** This is the stronger of the two: the test floor sits at ~55%, so ~45% of the suite could stop being collected while a floor stayed green — and the incident this guard descends from is precisely that shape (six files contributing 0 of 438, nothing red). The expectation is re-derived from `src/` every run, so there is no constant to go stale.

Plus a floor on executed tests, as a backstop for a partial collapse that still leaves every file present.

Deliberate limits, each of which is a way this could have been wrong:

- **It can only ever ADD a failure.** Vitest's exit code passes straight through when the ledger is satisfied.
- **EXECUTED, not TOTAL** — `numTotalTests` counts skipped tests, so a wholesale self-skip would satisfy a total-based floor having run nothing.
- **`failed` counts as executed** — otherwise the guard fires on every genuine failure and the tier is red for two indistinguishable reasons, which is the confusion being removed.
- **A signal-killed runner short-circuits the gate**, and exits `128+N` rather than a constant, so the CI task's `oom-killed` (137) and `timeout` (124) branches still work. A memory problem must not be rendered as a test failure.
- **A caller-supplied `--outputFile` is refused** rather than silently redirecting the report — measured, that turned an 18/18 green run into "COLLECTED NOTHING".
- **A narrowed run skips the FLOOR and the file ledger, but not the zero check.** Value-taking flags consume their value, so `--max-workers 1` is not mistaken for a filename filter; `--shard`/`--changed`/`--related` narrow explicitly.

The message **enumerates every cause it cannot tell apart** rather than asserting one. All of them have been observed on this suite while writing this change: a mock factory throwing, a browser that died mid-run, a browser that never came up, and an empty selection. The browser-crash one arrives wearing the mock error's headline with the truth on the `Caused by:` line, and the message says so.

## Verification

| | |
|---|---|
| Repro | 3/3 full runs abort at `d353f785c3`; the single file aborts alone |
| Collected **before** | **0** tests, 0 files — no summary line at all |
| Collected **after** | **201/201 files, 2254 tests, exit 0** — through `pnpm run test:component`, with the floor and the file ledger both armed |
| **In CI** | **`preview / component-tests`: pass** on this branch, on a quiet 4-CPU pod — the environment that actually matters |
| Genuine test failures surfaced | **none** — the suite is fully green once it runs |
| Gate unit tests | **27**, over JSON fixtures and a fixture `src/` tree |
| Mutation check | **17 mutants, all KILLED**, each by its own named test. One survived the first sweep — an unreachable branch — and is documented where it was fixed |
| File ledger vs the real tree | 201/201 green; dropping one real file fails and names it |
| ESLint | findings on the touched test file are **1:1 identical to `origin/main`** — none added |
| Typecheck | `tekton / typecheck` pass; `scripts/__tests__/**` is outside the root `tsconfig`, so the new test is not in the blocking typecheck |
| Adversarial audit | one round, 6 findings + 1 nit, **all fixed** in `7f005f0482` — see the comment below |

**Not verified:** I could not reproduce the ~27 failing files mentioned from `be9bc31370`; the suite is green now, so whatever those were has since been fixed. And one local abort after this fix was **not** host contention — see the comment below for what is and is not known about it.

## Follow-up the maintainer should decide (not done here)

**50 wholesale `vi.mock('~/providers/FeatureFlagsProvider')` factories still omit `useOptionalFeatureFlags`** — 41 browser-test files and 9 node-unit files. (15 more use an `importOriginal` spread and are immune; 12 already name both hooks — the AppBlocks ones fixed previously.) Each is a latent copy of this bug: dormant until something in that file's module graph starts importing the hook, then it zeroes the whole browser tier again. Every one of the 41 has a **static** import path to a real consumer; most are currently cut by another `vi.mock` in the same file, which is exactly why the hazard is invisible until an unrelated change re-opens the path.

Two mechanical options, ~48 files, neither with any runtime behaviour change:

1. Add the missing hooks to each factory, then widen `featureFlagsMockCompleteness.test.ts` from `src/components/AppBlocks` to repo-wide. The added exports are inert where the hook is never imported. **Note the module exports three runtime hooks, not two** — `useFeatureFlagsReady` is the third, with four live consumers, and that guard has the same two-hook blind spot.
2. Add `~/providers/FeatureFlagsProvider` to the `modules` list of `local-rules/no-wholesale-module-mock` in `.eslintrc.js` — which goes red on all 50 until they are converted. The flags module's own guard records that the `importOriginal` spread is **not** always available here (the real module imports `setTrpcBatchingEnabled` from `~/utils/trpc`, which some files' own wholesale trpc factory does not provide).

I have not done either: ~48 files of churn with a real review cost, and the existing guard's header states explicitly that widening it "needs an owner and a decision". **Closing condition if taken up:** `featureFlagsMockCompleteness.test.ts` runs repo-wide and is green, or the eslint rule lists the module and `pnpm lint` reports no `no-wholesale-module-mock` errors for it.

**Separately:** a single browser crash anywhere in the run zeroes all 201 files. That is an upstream fragility rather than something to fix here, but the new guard is what makes it legible when it happens.


ZacxDev:
## Audit round 1 — findings and what the fixes claim

An adversarial audit read `d353f785c3..6ceac37f33` and found two ways the new guard reproduced the pathology this PR exists to remove, plus four smaller gaps. All six are fixed in `7f005f0482`, plus one 🟢 the auditor raised about the fixed factory itself.

The auditor also independently confirmed the parts that matter most: the payload fix is **red at base, green at head** (it materialised the base version of the file and ran it through the wrapper — zero collected, exit 1; head version — 2 passed, exit 0); the gate **cannot turn red into green** on any path; and the census in the follow-up section reproduces **exactly** (50 factories missing the hook, 41 browser / 9 node, 15 `importOriginal`, 12 naming both).

Two of its measurements are worth recording because they contradict comments elsewhere in this repo:

- **The 25-minute budget is not the constraint it is documented to be.** `tekton_report_only_suite_duration_seconds{suite="component"}` on dp-1: max **152 s** for `pass` over 14 days. The pipeline comment saying "this suite has never completed inside it" is stale.
- **`pr-preview-pipeline.yaml:515` invokes `pnpm run test:component` with no arguments**, so the floor and the on-disk ledger are both armed in CI. That is the shape this PR was designed around, confirmed rather than assumed.

```audit-claims round=1 audited=6ceac37f33..7f005f04
1. A signal death is no longer relabelled: `exitCodeForSignal` returns 128+N, so an OOM SIGKILL reaches the CI task as 137 and takes its `oom-killed` branch instead of falling through to `fail`.
2. A caller-supplied `--outputFile` (any spelling) is refused up front with the fix in the message, instead of silently redirecting the report and making a green run print "COLLECTED NOTHING".
3. `isNarrowed` consumes value-taking flags, so the space form `--max-workers 1` no longer silently disables the floor; `--shard`/`--changed`/`--related` now narrow explicitly, so a sharded run is no longer failed with a misdirecting message.
4. `spawn` sets `shell` on win32 (matching `scripts/test-unit-run.mjs`), and a spawn failure (rc 127) short-circuits instead of printing the abort diagnosis for something that never ran.
5. The JSON report is deleted BEFORE the run, so a stale report from an earlier run can no longer satisfy a later run's ledger.
6. The file count is now CHECKED, not merely printed: the gate walks `src/` for every `*.browser.test.tsx` and fails when one is absent from the report, naming it. Skipped when narrowed or when the walk finds nothing.
7. The fixed factory names all THREE hooks the flags module exports, and its comment no longer says "BOTH" — `useFeatureFlagsReady` is the third and has four live consumers.
```

**Verification of the fix round:** 27 unit tests (was 16). Mutation-checked at **17 mutants, all killed**, each by its own named test. One **SURVIVED** the first sweep and is worth naming: `out.length > 0 ? out : null` was unreachable from the fixture, which had no `src/` at all, so the walk returned early — an empty array is truthy, so that mutant would have armed a ledger over zero expected files and passed everything while reading as a check that ran. A second fixture with a `src/` holding no browser tests reaches it. The on-disk ledger was also run against the **real** tree: 201/201 green, and dropping one real file fails and names it.

**Not fixed, deliberately:** the auditor's 🟢 8 (the two `.mjs` files are eslint-ignored and `scripts/__tests__/**` is outside the root `tsconfig`) is a pre-existing property of where these files live, noted so nobody assumes coverage.

## One thing the PR body does not yet say

Running the full suite locally after the fix produced **one clean 201/201 green run**, and later runs on the same box hit two further aborts. Three of the four were host contention from unrelated agents (`Failed to connect to the browser session`, `Browser connection was closed`) and are not evidence about this code. One was **not**: a mock-factory throw after 68 files had passed, with `src/components/Apps/AppListingsMarketplaceBody.urlState.browser.test.tsx` in flight — a file that passes in isolation and passed in a ten-file group. So either a second latent wholesale factory is reachable only under some interleaving, or mock routes leak between files sharing a browser session. I could not separate those on a box at load 13–21, and I am not going to report a load-contaminated bisect as a diagnosis.

What is certain: `preview / component-tests` reported **pass** on this branch in CI, on a quiet 4-CPU pod, which is the environment that matters. And whatever that second abort is, the guard added here now names it instead of rendering it as "Component suite failed".

ZacxDev:
## `preview / smoke-tests` is red here, and it is NOT this PR — it is red on `main`

This branch changes only test tooling (`scripts/`), one `*.browser.test.tsx`, a `package.json` script and `CONTRIBUTING.md`. None of that is in the preview bundle. Rather than assert that and move on, here is the attribution.

**The failing spec, from the TaskRun's own log** (`pr-preview-4531-nhdk7-smoke-test`), 1 of ~66, failed on the initial attempt and both retries:

```
tests/preview-engagement.spec.ts:46:7 › reaction.toggle and commentv2.upsert
  round-trip on a self-seeded post

  Error: reaction.toggle should be accepted (resolve, not throw)
  expect(received).toBeNull()
  Received: "created"
    > 74 |  expect(reaction ?? null, '…').toBeNull();
```

**Cause: #4516 changed `reaction.toggle`'s return value and this smoke assertion was not moved with it.** `1e810875c8` ("fix(reaction): await toggle handler to prevent null payload on refresh") replaced

```js
.mutation(({ ctx, input }) => { toggleReactionHandler({ ctx, input }).catch(handleLogError); })
```

with `.mutation(toggleReactionHandler)` — deliberately, so the write is not dropped on pod drain. The mutation therefore now resolves with the handler's result (`"created"`) where it used to resolve with `undefined`. `tests/preview-engagement.spec.ts:74` still asserts `toBeNull()`.

**Ancestry, which is what settles it:**

| head | contains #4516 (`0818cbdb90`) | `preview / smoke-tests` |
|---|---|---|
| `ca74417111` (#4525) | **no** | pass — 66 passed, 0 flaky |
| `847f22b173` (#4518) | **no** | pass — 66 passed, 0 flaky |
| `d353f785c3` (this PR's base) | **yes** | — nothing ran smoke on it |
| this branch | **yes** | **fail (1 failed)**, twice |

So the two green runs people would compare against both predate the change. This PR is simply the first preview whose base contains it, which is also why the failure looks new.

**It is the same defect class this PR is about**, one layer out: a contract moved, its consumer did not, and the tier that notices is one nobody attributes correctly by default. The code is right and the test is stale — #4516's whole point is that the mutation returns something now.

**Fix and closing condition (not done here — different owner, different tier):** change `tests/preview-engagement.spec.ts:74` to assert the toggle was accepted by its new contract rather than by being void — e.g. that it resolves to one of the handler's outcomes — and re-run. Closed when `preview / smoke-tests` reports pass on a head containing `0818cbdb90`. I have deliberately not folded that into this PR: it is unrelated to the component tier and belongs in its own change, where a reviewer who knows the intended `reaction.toggle` contract can confirm the new assertion is the right one.

ZacxDev:
## Audit round 2 — the fix round introduced two of its own

A delta re-audit of `6ceac37f33..7f005f0482` confirmed 4 of round 1's 7 claims outright, with its own positive and negative controls (signal mapping measured across five signals; the file ledger watched red at 200-of-201 and green at 201-of-201 against the **real** tree; the pre-run clear read at the source position). It also found that **two of the fixes had introduced new defects**, which is the whole reason the ladder exists.

Both are fixed in `c5681fd483`, along with six smaller findings.

**The two that mattered, and both are the shape this PR is about:**

1. **The new on-disk ledger false-failed a legitimate run with a message forbidding the fix.** `--exclude`, `--dir`, `--root` and `--config` take a value *and* genuinely shrink the collected file set, but were only in `VALUE_FLAGS`. So `pnpm test:component --exclude 'src/tests/**'` scored as a full run, and the ledger failed it naming up to 200 files — asserting the include had broken or the run had died, and telling the reader *"Do NOT silence this by narrowing the walk"*, which was the only thing that would have fixed it.

2. **`shell: win32` went on the shared `run()`, so it also wrapped the gate spawn.** Round 1's claim said this matched `scripts/test-unit-run.mjs` — it did not: that file puts `shell` on its vitest spawn only and deliberately leaves its `process.execPath` spawn alone. Node with `shell: true` concatenates argv unescaped (DEP0190), and `process.execPath` on Windows is `C:\Program Files\nodejs\node.exe`. On the one platform the option exists to support, every `pnpm test:component` would have died at the gate step with a cmd.exe parse error.

**Two findings about the guards' own honesty**, both worth naming: `--output-file` (kebab) was **not** refused, so the exact defect the refusal exists for was still reachable — under a test named *"catches every spelling"*; and `--outputFile.junit=`/`.html=` **were** refused though neither touches the `.json` key, so the guard was simultaneously too narrow and too wide. Spellings are now canonicalised rather than enumerated, which is the structural fix — vitest camelCases every option key, so any hand-written list has a hole wherever it carries one spelling and not the other.

```audit-claims round=2 audited=7f005f0482..c5681fd4
1. `--exclude`/`--dir`/`--root`/`--config` are now NARROWING, so the on-disk ledger no longer false-fails a legitimate run that deliberately collects fewer files.
2. `shell` is per-call and set only for the vitest spawn, so the gate spawn of `process.execPath` is no longer wrapped in cmd.exe on Windows.
3. Flag spellings are canonicalised (kebab and camel collapse to one name) instead of enumerated, closing the `--test-timeout`/`--testTimeout` class of hole in both sets.
4. `conflictingOutputFile` catches the kebab spelling and no longer refuses non-`.json` output keys.
5. The spawn-failure short-circuit is keyed on a flag, not on `rc === 127`, so a runner that legitimately exits 127 is still graded.
6. `--repo-root` consumes its value, so the flag can precede the report path without the directory being read as the report.
7. `--related` removed — a vitest subcommand this wrapper can never receive; the entry and its assertion were both inert.
8. The factory names all FOUR runtime exports of the flags module, and the comment no longer claims a scope narrower than the module.
9. `main` takes injected collaborators, so the ORDER of effects — clear-before-run, and the two short-circuits that must not consult the gate — is asserted rather than inspected. Both had shipped with no coverage.
```

**Verification of this fix round:** 35 unit tests (was 27). Mutation-checked at **17 mutants, all killed, zero survivors** — including re-running round 1's guards, because an audit fix resets the gate. Real-tree controls re-run with the new argument parsing, flag-first and flag-absent: 201/201 green, and dropping one real file fails and names it. ESLint back to the base count (two `no-empty-function` errors this change introduced are fixed).

**Ledger:** `round 2 · payload lines changed THIS round: 341 (scaffolding 184) · elapsed: ~50 min`. Non-zero, so the attribution gate does not fire — the ladder is still working on the PR rather than on itself.

Round 3 is running against `7f005f0482..c5681fd483`.

ZacxDev:
## Audit round 3 — the canonicaliser re-opened the class it closed

Round 3 verified **8 of round 2's 9 claims** against the diff or against vitest's own source rather than accepting them — it read cac's `setDotProp` to confirm the `--outputFile.junit=` allowance is genuinely safe, and `cliOptionsConfig` to confirm `-r`/`-c` really are `root`/`config`. It found one regression and one coverage hole. Both are fixed in `38b1f1de8b`.

**1. A boolean flag swallowed the filename after it — the same shape round 2 fixed, re-introduced by round 2's own mechanism.** Canonicalisation collapsed `.subkey` onto the parent, so every dot-subkey inherited its parent's value-consuming behaviour, and the two entries `--coverage.reporter`/`--coverage.provider` became one `coverage` that also matched the **bare** `--coverage` — which takes no value. So:

```
pnpm test:component --coverage src/…/review-queue-nav.browser.test.tsx
```

ate the file as `--coverage`'s value, scored the run as full, and failed an 18/18 green single-file run naming ~200 files as `ABSENT FROM THE RUN'S REPORT` with *"Do NOT silence this by narrowing the walk"*. Measured base-vs-head, `--coverage <file>`, `--coverage.enabled <file>` and `--browser.headless <file>` all flipped `true`→`false`. Matching is now on the full canonical **path**, so a subkey is value-taking only if it is listed as one.

**2. Three mutants survived a green 35-test suite, and each was this PR's own headline failure.** The injected fakes took no parameter, so nothing could observe the argv `main` builds — the seam between the two modules this change exists to wire together:

| mutant | consequence | was |
|---|---|---|
| drop `--narrowed` from the gate argv | every narrowed run hard-fails the floor **and** the ledger | SURVIVED |
| drop `--outputFile.json=` from vitest argv | every run reports "does not exist" + the abort diagnosis | SURVIVED |
| drop `...argv` from vitest argv | `pnpm test:component <file>` silently runs the **whole suite** | SURVIVED |

Testing `isNarrowed` in isolation cannot see whether its answer is ever *used*. The fakes now capture their argument; all three die.

```audit-claims round=3 audited=c5681fd483..38b1f1de
1. `canonicalFlag` keeps the `.subkey`, and VALUE_FLAGS matches the full canonical PATH, so a boolean parent (`--coverage`) and its boolean subkeys no longer inherit value-consuming behaviour and no longer swallow a following filename.
2. The injected fakes in the `main` tests capture their argv, and three cases assert it: `--narrowed` is passed exactly when `isNarrowed` says so, the vitest argv carries `--outputFile.json=` exactly once, and the caller's arguments are forwarded.
3. `--repo-root` with no value is a usage error (exit 2) rather than a silent fall-through to the real repo root.
4. CONTRIBUTING's sample ledger matches live output, and the three checks are numbered so "a narrowed run skips 2 and 3" names an unambiguous pair.
```

**Verification:** 39 unit tests (was 35). Mutation-checked at **19 mutants, all killed, zero survivors** — round 3's three survivors, the round-3 fixes themselves, and every guard from rounds 1 and 2 re-run, because an audit fix resets the gate. Real-tree controls re-run: 201/201 green, one missing file still fails and names it.

**Ledger:** `round 3 · payload lines changed THIS round: 56 (scaffolding 104) · elapsed: ~25 min`. Payload is falling sharply — 316 → 341 → 56 — which is the ladder converging rather than leaving the PR; it is still non-zero, so the attribution gate does not fire.

Round 4 is running against `c5681fd483..38b1f1de8b`.

ZacxDev:
## Audit round 4 — the flag LIST was the problem, so it is gone

Round 4 did the thing three rounds of hand-enumeration could not: it measured `isNarrowed` at base vs head against **every one of vitest 4.1.11's 72 boolean options and 73 value-taking ones**, parsed out of `cliOptionsConfig`. Result: round 3's fix closed the loud class completely (22 boolean options swallowed a following filename at base, **0** at head) and opened a quiet one — **25 value-taking paths** now read their VALUE as a filename:

```
--retry.count 2          --browser.name chromium        --coverage.thresholds.lines 80
--coverage.include …     --coverage.reportsDirectory …  (+ 20 more)
```

Each scored a **full** run as narrowed, switching the file ledger and the floor off with a one-line note — the direction this file's own comments repeatedly call worse than a loud failure.

**The list was the problem.** Enumerating names missed `--test-timeout` next to `--testTimeout`; canonicalising spellings made `--coverage <file>` swallow the file; splitting into subkeys produced the 25 above. vitest has a **164-path** option tree — a hand-maintained copy is wrong the day it is written, and each fix moved the wrongness rather than removing it.

So `VALUE_FLAGS` is **deleted**, and the rule is now about the argument rather than the flag: *a positional is a file filter if it looks like a path, or if nothing before it could have been expecting a value.* A non-path token straight after a flag is that flag's value, whatever the flag is — correct for all 73 value-taking options without naming one, and for all 72 booleans. What remains is `NARROWING_FLAGS` (flags that genuinely shrink the run, which must be named because omitting one fails *loudly*) and a five-entry `PATH_VALUE_FLAGS` for the only residual shape cannot decide: a value that is itself a path.

**And the decision is now a sentence, not a boolean.** No rule over an unknowable flag list is right always; what must never happen is being wrong *silently*, because `--narrowed` disables the two checks this change exists to add. The runner now prints why:

```
test:component: NARROWED — `2` was read as a file filter, so the file ledger and the
floor are skipped (the zero-collected check still applies). If that reading is wrong,
the checks you wanted are not running.
```

```audit-claims round=4 audited=38b1f1de8b..1441557b
1. `VALUE_FLAGS` is deleted; narrowing is decided by the ARGUMENT's shape, so no value-taking vitest flag can have its value read as a filename and no boolean flag can swallow one.
2. `narrowingReason` returns a sentence naming the token, and the runner prints it, so turning the two checks off is never the quiet outcome.
3. `canonicalFlag` camelCases only the FIRST dot segment, matching cac's own `camelcaseOptionName`, so the wrapper cannot accept a spelling vitest rejects.
4. `--repo-root=<dir>` is parsed, and a value that is missing, empty or itself a flag is a usage error.
```

**Verification:** 45 unit tests. Mutation-checked at **23 mutants, all killed, zero survivors, each by its own named test**.

🔴 **The first sweep reported four survivors and was wrong about all four — and the harness had this PR's own defect.** Two were mislabelled (killed by a different test than the one named, which the sweep scores as SURVIVED). Of the other two, one was a real gap (`--reporter=json AppNameCrumb`) and one was `canonicalFlag`'s dot handling, equivalent for every input without a dashed subkey; both now have cases. Along the way the harness was found **reusing a JSON report path without clearing it** — the same stale-report defect fixed in the product two commits earlier — and now clears it and verifies the mutation reached disk before running. A mutant reported SURVIVED is a claim about the instrument until the instrument has been controlled.

**Ledger:** `round 4 · payload lines changed THIS round: 115 (scaffolding 101) · elapsed: ~40 min`.

Round 5 is running against `38b1f1de8b..1441557bc7`.

PR #4542, opened 2026-09-01
perf(app-blocks): shorten the first BLOCK_INIT re-posts, preconnect the block origin, and count posts-before-ack

ZacxDev:
## What

Full-page App Block launches spend most of their time in `init_wait` — the gap between the host posting `BLOCK_INIT` and the block answering `BLOCK_READY` — not in minting the block token. This PR ships two host-side levers against that phase, plus the instrument that decides whether either of them actually helped.

🔴 **No performance improvement is claimed here.** This ships a change and the means to evaluate it. It is *expected* to reduce `init_wait`, and that is **unverified until measured post-deploy**.

## 1. Front-loaded `BLOCK_INIT` re-post schedule

`iframeInitController` posted `BLOCK_INIT` immediately and then re-posted on a flat 400 ms interval until the block acked. That flat interval quantizes the handshake: a block whose message listener attaches at, say, t=20 ms cannot be heard until t=400.

The interval becomes a short backoff — `INIT_RETRY_BACKOFF_MS = [50, 100, 200]` — settling at the **unchanged** 400 ms steady state. Posts land at t = 0, 50, 150, 350, then every 400 ms.

**Why this is safe, as a property of the guest rather than a hope about timing:** the SDK's iframe transport origin-checks every inbound message and latches on the first valid init, ignoring every subsequent one. A repeat `BLOCK_INIT` is a proven no-op on the block side — the same property that lets the re-post loop exist at all, documented in the controller's header comment. The only added cost is three extra `postMessage`s inside the first 350 ms, which are simply dropped if the frame is not listening yet.

Deliberately untouched: the readiness timeout (still armed on start, still 10 s) and the existing `BLOCK_HELLO` fast path, which still posts immediately without consulting the schedule.

The schedule is an exported constant so tests assert the real value instead of re-implementing it.

## 2. `preconnect` to the block origin from the run page's SSR head

The run page now emits `<link rel="preconnect" href="…" crossorigin>` for the block origin from its server-rendered `<Head>`.

Placement is the whole feature. The block iframe mounts on the first *client* render after hydration — well after the document head is parsed — so the head is where the DNS/TCP/TLS handshake gets a genuine head start. The same hint emitted from the host component would fire at the moment the browser would have opened the connection anyway: shipped-looking and inert.

🔴 **`crossorigin` is load-bearing.** Connections are pooled by (origin, credentials mode), and the frame's own subresources are CORS module scripts. Without the attribute the browser warms the *credentialled* pool, never reuses that connection, and the page pays for a socket while still paying the handshake it meant to skip — with no error and no console warning. A test asserts the attribute directly.

The origin is derived from the `iframeSrc` prop the page already resolves server-side, not rebuilt from the slug, so there is no second source of truth for the block origin.

Not done, deliberately: hover-preconnect on the app store grid. Two dozen cards is a connection storm bought with a weak intent signal, and the card view model does not carry the origin. A viewer already on the run page, by contrast, has committed to launching — so this hint has no speculative cost.

## 3. The discriminating instrument: `BLOCK_INIT` posts before ack

`init_wait` on its own cannot tell two very different situations apart:

| | mechanism | signature |
|---|---|---|
| (a) | the host quantized the handshake — the block was ready early but was not re-asked yet | ≥ 2 posts |
| (b) | the block genuinely takes that long to boot | 1 post |

Identical durations, opposite fixes. Counting the posts separates them — and it is also the only way to show whether change (1) moved anything.

The count rides the launch beacon that **already fires**: no new beacon, no new call site, no additional analytics row. It is an optional `initPosts` on the existing `timings` object, so it inherits every discipline that object already carries — dropped when the tab was ever hidden, dropped when the end-to-end total is unusable, and observed only on the success, non-secondary path.

Server-side it feeds an unlabelled histogram whose first two bucket edges are **1 and 2**, because the entire reading is the share at `le=1`. A first bucket of `le=2` would merge exactly the two populations the field exists to separate.

Out-of-range and non-integer samples are **dropped, never clamped**, following the module's existing convention. That matters more than usual here: a clamp lands junk in the top bucket, which reads as "this launch waited out many re-post ticks" — manufacturing precisely the evidence the metric was added to test for.

The sample cap is **derived in code** from the re-post schedule and the retry bound, mirroring the existing pattern for the duration cap, with a test asserting it clears the reachable maximum. Since the whole point of the schedule is that someone may tune it further, a cap chosen by eye would silently start discarding the highest-post-count launches — the signal — the next time it is shortened.

Both writers of the render/impression row are covered by a test each. The shared allowlist already makes a leak structurally impossible, so those tests assert an **exact payload equality** rather than the absence of one named field: a field-specific assertion would stay green if the allowlist were ever turned back into a destructure-rest denylist, which would re-open the class for every future field.

## Corrected a stale in-repo measurement

Two comments claimed that no deployed block sends `BLOCK_HELLO`, as of a date earlier this month. Re-measured against the deployed fleet on 2026-08-31: **4 of 23 deployed blocks now ship it**. Both comments are corrected and rewritten as dated measurements rather than standing facts.

The consequence is what motivates change (1): the other 19 would each need a rebuild-and-review cycle, so the host-side cadence is the only lever that reaches every deployed app at once. This does **not** reopen the init-fragment fast path — that gate rests on a separate count which was not re-measured.

## Explicitly out of scope

- **SSR-ing the token mint.** The mint is not the long pole, and doing it would bypass the rate limits that live in the route rather than in the signer, and move asymmetric crypto onto the SSR render path.
- **Asset prefetch.** The block's content-hashed asset URLs are not knowable without first fetching the shell.
- **The launch reveal animation.** It is a pure opacity cross-fade on the ready path; pointer events flip in the same commit as ready, so time-to-interactive is unaffected, and it collapses to zero under `prefers-reduced-motion`.
- **Relaxing the shell's cache headers.** They exist so out-of-band platform response headers propagate immediately.
- **Enabling the init-fragment allowlist, or a fleet rebuild.** Both are real options, but each is a per-app review cycle and belongs in its own change.

## Testing

Framework is vitest; everything new lives in the `unit` project, which is the tier CI runs.

**Every new test was watched RED at `origin/main` and GREEN at HEAD** — 8 files, 44 failures at base, 165 passing at HEAD.

Each guard was additionally mutation-checked by breaking it deliberately and confirming a test failed **on that guard's own assertion**:

| mutation | killed by |
|---|---|
| flatten the backoff to `[]` | the absolute-offsets test (`expected 1 to be 2`) |
| drop `crossOrigin` | `expected undefined to be 'anonymous'` |
| leak `initPosts` into the analytics payload | **both** writer tests, each on its own payload equality |
| clamp instead of drop an out-of-range count | the drop-not-clamp test (`expected 128 to be null`) |
| observe the count without the end-to-end anchor | the anchor test |
| accept a zero count | the reject-zero test (`expected +0 to be undefined`) |
| forget to reset the count on soft navigation | the re-arm test (`initPosts: 7` vs `0`) |
| set the cap below the reachable bound | the derivation test (`expected 20 to be greater than 87`) |
| fold the counter into the one-shot stamp | the wiring gate |
| remove the counter entirely | the wiring gate |

That last pair covers a gap worth naming: if the host never increments, every pure test stays green and the beacon simply omits the field — the metric would report *no data* rather than wrong data, which is quiet in the worst way.

Full unit project: **1554 files / 24546 tests**, 2 failures — both pre-existing and unrelated (a Redis-key-constant gate and a shared-module-mock gate, both naming a file this PR does not touch). Verified identical on a clean `origin/main` tree. Typecheck and ESLint clean.


ZacxDev:
### CI status — all three red checks are pre-existing, none caused by this PR

Verified rather than assumed, because "the same job name is red on main too" is not on its own evidence that it is red for the *same reason*.

**`Unit tests (1)` and `Unit tests (2)`**

Both fail on `main` at `017e35fa03` — the exact commit this branch is based on. Comparing the two runs' logs, the failing files and the failure reason are identical:

- `no-hand-typed-redis-key-constants.test.ts` (shard 1)
- `no-direct-shared-module-mock.test.ts` (shard 2)

Both gates name the same offending file, `src/server/redis/__tests__/queues.integration.test.ts`, on both runs. This PR does not touch it.

Every test file this PR adds or changes is marked `✓` in those same shard logs.

**`preview / component-tests`**

This one is *not* mine either, and it has an open fix already: #4531, "unzero the `preview / component-tests` tier". That PR documents the tier failing on `main` while executing **zero** tests — a whole-project abort from a single unrelated file, `src/tests/pages/apps/review/review-queue-nav.browser.test.tsx`, whose wholesale mock of the feature-flags provider misses a named export. In browser mode that kills the entire run rather than one file.

I reproduced it locally on this branch, then ran the control: reverting *only* this PR's source changes and re-running the same file produces the identical mocking error. So the abort is independent of this change.

I also ran the browser tier locally against this branch — the full `AppBlocks` component suite is **37 files / 453 tests, all passing**, including the two host suites whose comments this PR corrects.

(Running that tier locally needed a Playwright shim: the pinned build is `1200` and this machine provides `1228`. Worth knowing for anyone else who tries — without it the suite cannot launch a browser at all, which is easy to misread as a code failure.)

**Green:** App unit tests + typecheck, Package unit tests, Unit tests (3) and (4), ESLint + Prettier, Schema drift gate, event-engine-common pin, both Windows dev-env jobs, both tekton jobs, and preview deploy / render-check / auth-tests.

ZacxDev:
## Adversarial audit — 2 findings, both fixed in `fb9cbda5f8`

Rebased onto current `main` first (via a **merge**, not a force-push — see the note at the end), so everything below was audited against the tree this PR's merge actually creates. Both pre-existing red gates are gone: #4544 fixed them, and the full unit project on the merged tree is now **1556 files / 24626 tests, 0 failures**.

---

### 🟠 MEDIUM — the headline statistic named the wrong divisor, and the error flattered this PR

`civitai_app_block_launch_init_posts` is read as a share at `le=1`. Its help text told the reader to divide by `launch_total_seconds_count`. Those two `_count` series **diverge**.

A launch reaches `launch_total_seconds` but *not* this histogram whenever its post count is unusable:

1. **The block acked before the host ever posted an init** → count `0` → dropped. This is reachable today, not hypothetical: the `BLOCK_READY` handler stamps `readyAt` and flips `loading → ready` with no dependency on init having been sent. A hand-rolled host shim that acks while the token is still resolving does exactly this — and this PR's own fleet measurement records that some deployed blocks *are* hand-rolled shims.
2. A manual-retry storm pushed the count past the cap (finding 2).

**Concrete scenario.** Three launches: one ordinary (`initPosts: 1`), one that acked before init, one past the cap. `launch_total_seconds_count` moves by **3**; `launch_init_posts_count` moves by **1**.

Case 1 is *fast, no-quantization* traffic — exactly the population that belongs in the `le=1` numerator. Dividing by the larger series removes it from the numerator but not the denominator, so the `le=1` share is **understated** and the fleet reads as **more quantized than it is**. That is the direction that flatters the re-post cadence change this metric exists to evaluate. An instrument biased toward its author's hypothesis is worse than no instrument.

**Fix:** the help text now names `civitai_app_block_launch_init_posts_count`, warns off the other series by name, and explains the direction of the bias. The 3-vs-1 divergence is now a committed test rather than a note, so nobody can "simplify" it back.

Measured, not reasoned — that ratio is an assertion in the suite.

---

### 🟡 LOW — `worstReachableInitPosts()` claimed a maximum it does not bound

Its docstring said "the most BLOCK_INIT posts one successful launch can legitimately make." That is false, and a guard whose description overstates its coverage is worse than no guard, because it stops anyone looking.

A **manual** retry is deliberately uncapped — `handleRetry` spends none of the automatic budget — and `performRetry` resets neither the launch marks nor the emit-once beacon latch. So:

**Concrete scenario.** Attempt fails → auto-retry scheduled with a backoff → user clicks Retry *inside* that window → `performRetry` sets status back to `loading`, which makes the auto-retry decision recompute to `none` and tears down the pending timer, **leaving `autoRetryBudget.attempts` at 0 and no beacon emitted**. Repeat. Each attempt contributes up to `maxInitPostsWithin(BLOCK_READY_TIMEOUT_MS) + 1` posts to the *same* launch sample. Four such attempts stay under the cap; five clear it.

The consequence lands in the **safe** direction — past the cap the count is *dropped*, never clamped, so no wrong value is ever published — but it is the second contributor to the divergence above, which is why the two findings share a fix.

**Fix:** the docstring and the test name now say "on the **automatic** retry path", and a new test pins the manual-retry boundary as a real property rather than pretending it away.

---

### Audited and clean — recorded so the clean areas are on the record

**Timer lifecycle of the chained `setTimeout`.** Seven scenarios against `vi.getTimerCount()`: `dispose()` leaves zero pending timers; exactly **one** retry timer is ever outstanding, so the chain cannot accumulate the way a leaked interval would; `notifyReady()` mid-gap leaves nothing pending and posts nothing late; a readiness timeout and a re-post falling due at the **same instant** resolve with the timeout winning and no extra post; `notifyHello()` after the timeout arms nothing; two controllers across a retry remount neither share nor leak. No stray post, no timer outliving its controller. (Exploratory, not committed.)

**Preconnect origin cannot be attacker-influenced.** `stampCanonicalIframeSrc` overwrites `manifest.iframe.src` **unconditionally** on every write path (git-push, manifest update, submit, approve, commit). The manifest `id` pattern is `^[a-z][a-z0-9-]*[a-z0-9]$` — no dot or slash, so the host label cannot be made to shift. And `new URL().origin` normalises whatever it is handed, with non-`http(s)` schemes rejected.

**Cannot render on a 404 or flag-gated page.** `getServerSideProps` returns `notFound` before it returns props, so the component never renders at all.

**`next/head` genuinely renders the `<link>`** — verified against next's own source rather than assumed. `onlyReactElement` concatenates any non-string child, and `unique()`'s dedup switch handles only `title`, `base` and `meta`, so a `link` always survives. A `null` child (the no-origin case) is stripped upstream by `Children.toArray`.

**Impression accounting cannot be perturbed.** Enumerated the writers rather than sampling: there are exactly **two** callers of the ClickHouse insert, and both go through the `blockRenderTrackerPayload` allowlist. No third call site exists, so no path bypasses the exact-payload assertions.

---

### Verification

Each fix was mutation-checked, and each mutant died on **its own** assertion, not a neighbour's:

| mutation | killed by |
|---|---|
| revert the help text to the wrong divisor | the divisor guard — `expected … to contain 'civitai_app_block_launch_init_posts_count'` |
| force the two `_count` series to agree (observe junk as `1`) | the divergence guard (`expected 3 to be 1`) + two existing drop guards |
| raise the cap past the manual-retry boundary | the new boundary test (`expected 145 to be greater than 200`) |

A second round auditing only the delta produced no findings, which is where the ladder ends.

`AppBlocks` browser suite on the merged tree: **37 files / 453 tests**, all passing.

**On the rebase:** a rebase cannot be pushed without `--force`, which I was asked not to use. I confirmed the rejection rather than assuming it, then produced the same result with a merge instead and verified the resulting tree is **byte-identical** to the rebased one (same tree OID), so every measurement above applies to what was actually pushed. The push was an ordinary fast-forward.

**Still not claiming any latency improvement.** This PR ships the change and the instrument; the before/after read happens after deploy. Not merged.

PR #4543, opened 2026-09-01
feat(apps): let an app author mark their store listing "in beta"

ZacxDev:
An app developer can mark their store listing **"in beta"** and add a short note.

- **Store card** — a small `Beta` badge.
- **Store detail page** — a `Beta` badge beside the app name, plus a notice above the description carrying the author's message (or a generic sentence when they wrote none).
- **`/apps/run/<slug>`** — the same notice in the page chrome, so someone who lands straight on a running app sees it too.

Author-editable on **both** listing kinds (on-site and off-site) from the existing store-details form. Display only: it feeds no ranking, no curation, no launch gating and no consent surface.

---

## 🔴 The migration is a hard PRE-DEPLOY step

**Apply this SQL to each environment before the code ships.** This is not a "the feature stays inert until then" situation — the code will 500 without it.

```sql
ALTER TABLE "app_listings" ADD COLUMN IF NOT EXISTS "is_beta" BOOLEAN NOT NULL DEFAULT false;
ALTER TABLE "app_listings" ADD COLUMN IF NOT EXISTS "beta_message" TEXT;
```

(`packages/civitai-db-schema/prisma/migrations/20260901120000_app_listing_beta/migration.sql`)

**Why the ordering is hard.** Prisma returns the created/updated row, so it emits `INSERT/UPDATE … RETURNING <every scalar the MODEL declares>`, and the same is true of any `findUnique`/`findFirst`/`findMany`/`update` on this model that passes no explicit `select` — roughly half the ~92 `appListing.*` query sites. Both columns are declared on the model, so the generated SQL names them whether or not the keys appear in `data`. Deploying first turns an additive cosmetic label into a public-store outage. This is the same ordering constraint `20260823120000_app_listing_source_repo` established the expensive way.

Both statements are catalog-only on PG 11+ (`ADD COLUMN … DEFAULT false NOT NULL` does not rewrite the table), so they are O(1) regardless of row count. `IF NOT EXISTS` on both, so a re-run is a no-op.

**🔴 No test in this repo can see that hazard, and I am not implying otherwise.** Every suite here mocks Prisma, so none of them generates SQL. The guarded reader below is defence in depth for the paths that *do* pass an explicit `select`; it is not, and cannot be, a substitute for running the migration first. The `sourceRepoUrl` module shipped a production 500 with a green suite, a mutation battery and two adversarial audit rounds behind it.

---

### Trivial, not material

Editing beta applies **in place with no moderator review** — the same posture `tagline` / `description` / `category` already have. The keys are deliberately absent from `MATERIAL_LISTING_PATCH_FIELDS`, and the inputs deliberately carry **no** `data-material-field` attribute (that attribute is walked by a ledger test asserting a disabled state the server does not impose here).

### Beta is never staged on a shadow revision

This is the one place the feature departs from how `sourceRepoUrl` is handled, and it is the decision everything else follows from. Every beta write targets the **live** listing — including from `updateRevisionDraft`, which writes the beta half to `shadow.revisionOfId` — and `applyApprovedRevision` copies **neither** column in **either** kind branch.

Letting beta ride the shadow (the obvious implementation) is wrong three separate ways:

1. **It reverts the author.** A shadow can sit in the review queue for days while the author keeps editing the live listing (beta is trivial, so those edits apply immediately). The apply copies the shadow's scalars unconditionally, so approving a week-old revision would restore the clone-time value and silently undo every beta edit since.
2. **It would break the on-site no-op claim.** The on-site apply branch is assets-only, and `revisionApplyScope('onsite') === 'assets-only'` is the only thing licensing the review panel's *"approving changes nothing"* statement. Copying a scalar there makes that statement false for every on-site media revision — the exact defect `OFFSITE_UNCOMPARED_APPLY_FIELDS` exists to prevent — and the only fix would be withdrawing the claim, losing a real moderator signal for a cosmetic label.
3. **It would put a trivial field back in a review queue** for any author whose patch also touched a material field.

Consequently **neither column belongs in `OFFSITE_UNCOMPARED_APPLY_FIELDS`** — that list names what the apply copies *without comparing*, and this apply copies neither. `beginListingRevision` still clones the columns onto the shadow, but for **display only**: the moderator review preview renders the shadow row, so without the clone a reviewer would see a beta app stripped of its beta framing.

Mixed patches therefore split: the material half to the shadow, the beta half to the live parent, both inside one transaction.

### The guarded reader

`src/server/services/blocks/app-listing-beta.service.ts`, mirroring `app-listing-source-repo.service`:

- the columns are read **there and nowhere else** — never added to `listingHydrateSelect` (shared by the public grid *and* the detail page) or to `moderationListingSelect`;
- narrow matching on the error **code** (`P2022` / `42703`, delegating to the sibling module's predicate rather than re-deriving it) — a connection failure, timeout, missing table (`42P01`) or permission error **propagates**;
- the **`available` vs value** distinction is preserved: "no beta set" is `{available: true, isBeta: false}`, "migration not applied" is `{available: false, …}`. They render identically; only the first licenses a write, and every write path gates on it;
- two write helpers following the existing split exactly — a spread fragment that **omits the keys** for system-originated writes (the revision clone), and an assert that throws **`PRECONDITION_FAILED`** for author-originated ones, because silently dropping a value the author just typed is not recoverable.

The two list surfaces (public grid, moderator table) use one batched guarded read per page, so the cost is O(1) queries regardless of page size and it degrades to an empty map.

### The run page

`/apps/run/<slug>` SSR resolves its app from `app_blocks`, not `app_listings`, so beta needs a listing lookup — and a failure there does not degrade a badge, it takes the **app launch** down.

- It goes through the same guarded reader and **fails open** to "not beta".
- It is keyed on the **slug** rather than on the `appBlockId` the block resolve returns, specifically so it needs nothing from that resolve and the two issue **concurrently** — the page waits for the slower of the two, not their sum. `AppListing.slug` is `@unique` and for an on-site app it *is* the block's `blockId` (`app-listing-mapper.ts` → `slug: ab.blockId`), the same fact the page's recents store already relies on. So this adds no serial hop to the launch path.
- A test pins the concurrency, because a "harmless" refactor to key it on `appBlockId` would compile, pass everything else, and quietly add a round trip.

I used a visible notice in the page chrome rather than the lazy dropdown-only `useChromeListingDetail` — that hook only mounts once a user opens a menu, and the decision was that beta shows on the run page.

### Preview posture (a documented **keep**)

`AppListingDetailBody`'s `preview` omission ledger withholds everything live, interactive or aggregate. The beta badge + notice are none of those, so they are **kept** — a moderator reviewing a shadow should see the beta framing they are approving. The ledger comment says so, and there is a test with both a positive control (the live arm renders them) and a negative control (the same preview arm still suppresses the action card).

The badge sits in the identity block rather than in the meta line beside the category badge, because that line is `!preview`-gated for an unrelated reason (its date is the publish request's submission time in preview) — putting the badge there would silently reverse this decision via someone else's gate.

### Safety of the free-text message

`betaMessage` is author-controlled public copy that **no moderator reviews** before it goes live. That is the same trust posture `description` already has on this same surface, so it is precedent-consistent. The mitigations are deterministic rather than procedural:

- **bounded length** via `BETA_MESSAGE_MAX` (200), exported alongside `OFFSITE_TAGLINE_MAX` and enforced **in zod at the request boundary** — not merely by the input's `maxLength`, which any direct API or CLI caller ignores. Re-checked in `buildListingPatchData` too, since that function is exported and unit-tested directly.
- **plain text, not markdown.** It is rendered as a text node and does **not** go through `CustomMarkdown` (which the description beside it does), and no `dangerouslySetInnerHTML` is introduced. A test asserts markup arrives verbatim and mints no `a` / `strong` / `img`, paired with a positive control that the description path *does* parse markdown on the same mount.
- the note is nulled at every projection when the flag is off, so a stale note left by an author who unticked the box cannot reach a public DTO.

**Residual risk, stated honestly:** an author can write misleading or unpleasant prose about their own app for up to 200 characters, publicly, without prior review. That is a moderation problem — the listing is delistable, and the moderator table carries the flag and the note so it is visible — not a rendering one. If we later decide beta copy needs prior review, the change is to move the keys into `MATERIAL_LISTING_PATCH_FIELDS`, which would also mean undoing the never-staged design above.

### Deviations from the brief, called out

- **The brief said on-site authors have no listing-scalar surface and one must be added.** On this tree they already have one: `editorTabsFor` grants `details` for both kinds, and that tab hosts `ExternalListingEditForm`, which is already kind-aware (it drops only the URL step and the scope disclosure for on-site). Adding the beta inputs to that form covers both kinds with no new surface.
- **The brief said to add the field to the revision clone *and* the apply.** I added it to the clone (for the preview) but deliberately **not** to the apply, for the three reasons above. The round-trip tests pin the resulting property — the parent's value survives an approve for both kinds — by asserting the apply names neither column, which is a stronger guarantee than "the copy is correct".
- **The canonical schema is `packages/civitai-db-schema/prisma/schema.full.prisma`.** There is no `packages/civitai-db-schema/prisma/schema.prisma` and no root `prisma/schema.prisma` on this tree. `src/models.ts` and `src/kysely/types.ts` are generated from it and are regenerated in this commit. The `schema-drift` gate only *warns* on a column absent from the committed catalog snapshot (that is what an unapplied migration looks like), so it will not go red.

---

## Test coverage

`pnpm exec vitest run --project unit` — **1549 files, 24423 tests**, 2 failures, both **pre-existing and outside this diff** (`no-direct-shared-module-mock` and `no-hand-typed-redis-key-constants`, each pointing at `src/server/redis/__tests__/queues.integration.test.ts`). Verified by re-running both ledgers with this branch's new test file moved aside — the same two failures, naming the same file.

Component project: `src/components` **191 files pass**, `src/tests` 7 of 8 (`review-queue-nav.browser.test.tsx` crashes at startup — pre-existing, untouched by this diff, last modified in #3482). Both the touched component files pass in full: `AppListingCard.browser.test.tsx` 46, `AppListingDetailBody.browser.test.tsx` 98.

New/extended coverage:

- **guarded reader** — 30 unit tests including the degraded branch via a throwing fake: available/true, available/false on `P2022`, on `42703`, and **propagation** (not degradation) on `42P01`, a timeout, a permission error, a connection failure and an uncoded error; the batch and slug readers; the non-memoised column probe; and a seam test that the two write helpers must *disagree* on the same unavailable input (one silent, one loud).
- **state machine** — an approved listing patched with only `isBeta`/`betaMessage` applies **in place and mints no shadow**, with a positive control that a `name` change on the same fixture *does* mint one.
- **the split** — a mixed patch puts the material half on the shadow and the beta half on the live parent; a material-only patch issues no empty parent write; `splitBetaPatch` is exercised by **walking `BETA_PATCH_FIELDS`** so the constant is load-bearing rather than decorative.
- **revision round trip** — for **both kinds**, the apply writes neither beta column, each paired with a positive control that the same call *does* write what it should (off-site scalars; on-site assets only). Plus: the clone carries the columns, and omits them entirely when the migration is outstanding.
- **read path** — the `projectListingCard` / `projectListingDetail` exact-key-set allowlists updated deliberately (`isBeta` added to both, `betaMessage` detail-only with the card asserting its absence), plus the stale-note rule with its positive control.
- **schema** — `updateListingPatchSchema` cases including the length cap from **both sides** of the boundary, `isBeta: false` satisfying the at-least-one-field refine, and a non-boolean rejection.
- **component** — card badge present/absent, detail badge + notice, the no-message fallback, the plain-text assertion, non-membership of the mutually-exclusive alert pair, and the preview **keep** with both controls.
- **run page** — props, the slug key, the concurrency, and fail-open on `P2022` / `42703` / a missing listing row, with a positive control that the resolver can still 404 and an assertion that a non-missing-column error is *not* swallowed.

**Absence assertions do not use `expect.element(...).not.toBeInTheDocument()`** — that matcher is inert in this repo (#4197) and passes either way. Every one reads `textContent` or a `querySelectorAll(...).length`, and each is paired with a positive control proving the container rendered.

### Mutation matrix — 16 mutants, all killed, each for its own reason

| # | Mutation | Killed by | Failure reason |
|---|---|---|---|
| M1 | `assertBetaWritable` never throws | *throws PRECONDITION_FAILED with the EXACT exported message* (+2 more) | `expected undefined to be an instance of TRPCError` |
| M1b | same mutant, graded via the service | *refuses with PRECONDITION_FAILED and the EXACT message* | `expected undefined to be an instance of TRPCError` |
| M2 | it throws `BAD_REQUEST`/`'nope'` instead | *throws PRECONDITION_FAILED with the EXACT exported message* | `expected 'BAD_REQUEST' to be 'PRECONDITION_FAILED'` |
| M3 | `betaWriteFragment` always emits the keys | *emits an EMPTY object when they are not* | `expected { isBeta: false, betaMessage: null } to deeply equal {}` |
| M4 | guarded read catches **every** error | 4 propagation cases | `promise resolved "{ available: false, … }" instead of rejecting` |
| M5 | message bound off-by-one (`>` → `>=`) | *a message of EXACTLY the cap is accepted* | `TRPCError: betaMessage must be ≤ 200 characters` |
| M6 | beta staged on the shadow (split bypassed) | *the beta half lands on the LIVE parent* | `expected { isBeta: true, … } to deeply equal { name: 'Renamed' }` |
| M7 | off-site apply copies `isBeta` | *off-site: the apply writes NEITHER beta column* (+ an existing exact-equality test) | `expected [ 'name', 'tagline', … ] to not include 'isBeta'` |
| M8 | on-site apply copies `isBeta` | *on-site: the apply writes NEITHER beta column* | `expected [ 'iconId', 'coverId', 'isBeta' ] to not include 'isBeta'` |
| M9 | clone drops the beta fragment | *beginListingRevision CLONES beta onto the shadow* | `expected {…} to match object { isBeta: true, … }` |
| M10 | card badge renders unconditionally | *renders NO Beta badge when the listing does not declare beta* | `expected '…' not to contain 'Beta'` |
| M11 | detail notice gated on `!preview` | *preview KEEPS the beta badge + notice* | `the beta notice is a documented KEEP in the preview ledger: expected +0 to be 1` |
| M12 | badge derived from `betaMessage` not `isBeta` | *a beta listing with NO message still gets a badge* | element never appeared |
| M13 | run-page beta read made serial | *issues the beta read CONCURRENTLY* | `expected 2 to be less than 1` |
| M14 | run page bypasses the guard (fails closed) | 2 fail-open cases | the resolver rejects instead of serving |
| M15 | detail projection leaks a stale note | *does NOT project a stale note when the flag is OFF* | `expected 'left over from before' to be null` |

Each mutant was applied to the real source, graded by one named test file, and the tree restored from a `cp` backup and byte-verified afterwards.

---

## Merge note

This does not touch `AppListingDetailBody.tsx:1218` (the action-card header). A concurrent PR changes that one line. **Whoever merges second must rebase on `origin/main` and re-read the merged region** — a clean git merge is not a clean merge.

## What I could not verify

- The migration itself was not applied anywhere — I have no database access from here. Everything above is against mocked Prisma.
- No browser/manual walkthrough of the real click path; the component tests render the components with the real Mantine + Tailwind cascade, which is not the same thing as driving the live app.
- The `schema-drift` gate's verdict on a real PR run (I reasoned about it from `gate.ts`'s documented behaviour, but did not execute the gate).


ZacxDev:
**Audit round 1** — full adversarial pass, read at `11af361a65`. Four findings (no 🔴), all addressed in `c5b239634c`. One of the fixes broke two pre-existing tests and was itself corrected in the same round; that is recorded below rather than smoothed over.

legend: `<from>` = the tip this round's audit READ · `<to>` = the head this round's FIXES produced.

```audit-claims round=1 audited=11af361a659f28486e5589feac3f819213952b09..c5b23963
1. F1 — `buildListingPatchData` now sets `betaMessage = null` in the same write that sets `isBeta: false`, last and unconditionally, so it overrides an explicit `betaMessage` in the same patch; turning beta ON deliberately does NOT clear it. Three comments that asserted the old (false) invariant were corrected.
2. F1 — the defect was reproduced end-to-end against the real `buildListingPatchData` / `editContextToForm` / `buildScalarPatch` / `projectListingDetail` before the fix, and five regression tests were added, including the ON-transition non-clear and a no-`isBeta`-key control.
3. F2 — added `readListingBetaForRender` / `readListingBetaBySlugForRender` / `readListingBetaManyForRender`, which degrade on ANY error, and routed the run page, the public `/apps` grid, the store detail read, the mod-review preview and the mod table through them. The narrow `P2022`/`42703`-only readers are unchanged and still used by every write-gating path.
4. F2 — the run-page test that asserted a `42P01` propagates was replaced by five cases asserting the app is still served on `42P01`, a timeout, a permission error, a connection failure and an uncoded error; the reader's own suite still pins narrow propagation.
5. F3 — `AppListingsModerationTable.tsx` now renders a `Beta` badge and the note (plain text, gated on the flag) with four browser tests; a per-test `mocks.rows` override was added and is reset in `beforeEach`.
6. F4 — `updateRevisionDraft` now asserts the parent is not moderator-removed before writing the beta half to it, allowing the owner-unpublish repair state, on a PRIMARY read.
7. F4 — that gate was initially UNCONDITIONAL and broke two pre-existing `offsite-listing.source-repo.service.test.ts` cases by reading the parent (and 404ing) on every call; it is now scoped to `betaData`, with a regression test asserting no parent read happens without a beta half.
8. Nit — the shadow clone now sees the POST-edit beta value: the parent beta write was moved BEFORE `beginListingRevision`, pinned by an `invocationCallOrder` assertion. The two writes are consequently no longer atomic and the comment claiming they were has been replaced with the trade-off.
9. Nit — the public `/apps` grid reads beta in parallel with the hydrate (`Promise.all`) instead of serially; the assets-only `loadListingEditView` path skips the beta read via a new `skipBeta` parameter.
10. Nit — the "roughly half the ~92 `appListing.*` query sites" figure (copied from the sibling module, never re-measured, ~2.6x high, asserted by no test) was removed from both the module header and the migration header and replaced by the mechanism: row-returning methods are exposed, `updateMany`/`deleteMany`/`createMany` return a `BatchPayload` row count and cannot raise P2022.
11. Verification — 11 fix-round mutants, each killed by its own named test with its own message, including two that check the F1 rule is not over-applied and one that the F4 gate is not too strict. `pnpm typecheck` 0 errors; unit project 1549 files / 24458 tests with only the two known-red base failures; component `src/components` 191 files, `src/tests` 7 of 8 (the 8th is the known-red `review-queue-nav`).
```

**Not fixed, deliberately:** the same `~92` sentence exists in `app-listing-source-repo.service.ts`, which this PR does not otherwise touch — flagged here rather than edited, so the diff stays scoped.

ZacxDev:
**Audit round 2** — delta re-audit, read at `c5b23963`. Seven findings (no 🔴), all addressed in `cd549b560d`. N1 was a regression the round-1 fix itself introduced; it is fixed at the root rather than by re-ordering again.

legend: `<from>` = the tip this round's audit READ · `<to>` = the head this round's FIXES produced.

```audit-claims round=2 audited=c5b239634c4d215657ebdc3b029439ab8c9241fa..cd549b56
1. N1 — the ordering constraint is REMOVED rather than re-tuned: `getListingPreviewForReview` now reads beta from the PARENT for a shadow (keyed on `revisionOfId`, spread onto its own select, not `listingHydrateSelect`), so `beginListingRevision` no longer clones the beta columns at all.
2. N1 — `updateListing`'s approved+material branch now builds BOTH payloads, mints the shadow, and only then writes, both writes inside one interactive transaction; a validation refusal is side-effect-free again and the function's stated "nothing has been written when it throws" invariant holds.
3. N1 — `betaWriteFragment` had no callers left and is deleted, with its tests, and the module header's "TWO SHAPES OF WRITE" section rewritten to "there is exactly one shape"; the remaining `assertBetaWritable` docstring now contrasts with the sibling `sourceRepoUrl` module instead of the deleted helper.
4. N6 — closed by the same change: the reorder had only made the preview fresh on the FIRST save (`beginListingRevision` returns an existing shadow before the clone ran), and reading the parent cannot go stale at all.
5. N5 — `buildListingPatchData` takes a new REQUIRED `currentIsBeta` opt and evaluates the EFFECTIVE flag (`patch.isBeta ?? opts.currentIsBeta`). A note set while beta is off and not being turned on is REFUSED with a `BAD_REQUEST` naming the field; a patch that turns beta off — with or without a note — clears. `updateRevisionDraft` passes the PARENT's flag, not the shadow's.
6. N3 — the false absolute is gone: the `*Many` exemption is now scoped to the RETURNING list ("they still name whatever columns appear in `data` and `where`"), `createManyAndReturn` / `updateManyAndReturn` are named as row-returning, and the header warns they contain the exempt names as substrings. Both the module header and the migration header were corrected.
7. N2 — the run-page comment now names `readListingBetaBySlugForRender` and says not to consolidate it with the narrow reader; `app-listing.service.ts`'s two stale symbol references updated to the `…ForRender` names.
8. N4 — degraded render reads are logged through `noteDegradedBetaRead` at `type: 'warning'` via the canonical `loggingMock` sink, `.catch(() => null)`-guarded; the false "the error is NOT lost because siblings hit the same client" justification is replaced with why that is untrue on the run page specifically.
9. N7 — the dangling "see there" cross-reference in `updateRevisionDraft` is replaced by the restated reason (the paragraph it pointed at was deleted in round 1).
10. F4 follow-through — the parent-editability gate and the parent beta read are resolved in one block scoped to `touchesBeta`, so a shadow edit with no beta half still reads the parent not at all.
11. Verification — 9 round-3 mutants killed by their own named tests. TWO were initially misreported: one was graded against a suite that failed to IMPORT (the beta test file had been left syntactically broken by round 1's fragment deletion — "SURVIVED" was really UNGRADED), and one exposed a genuinely vacuous test (asserting the read resolves passes with the `.catch` deleted; a `process.on('unhandledRejection')` listener never fires under vitest). The latter is rewritten to observe the rejection handler on the returned promise and verified RED against the mutant.
12. Verification — `pnpm typecheck` 0 errors; full unit project 1549 files / 24470 tests with only the two known-red base failures and ZERO import failures; component `src/components` 191/191, `src/libs` 1/1, `src/hooks` 2/2.
13. The new `vi.mock` of the logging client was replaced with the canonical `loggingMock` after it grew the `no-direct-shared-module-mock` ledger; that ledger now lists only the pre-existing `queues.integration.test.ts`.
```

**Still not fixed, deliberately:** the `~92` sentence in `app-listing-source-repo.service.ts` (this PR does not otherwise touch that file); `getMyListingForApp`'s assets-only path still pays `readListingSourceRepoUrl` whose result it discards (pre-existing, unrelated to beta).

ZacxDev:
## 🔴 Retraction of a round-2 claim

The round-2 `audit-claims` comment above states, in claim 6: *"Both the module header and the migration header were corrected."* **That was false.** `migration.sql` was not touched in `c5b23963..cd549b56` — only the module header was. The migration header still carried the exact false absolute N3 was raised about (*"`updateMany`/`deleteMany`/`createMany` … cannot raise it"*), and after round 2 it directly contradicted the module header twenty lines of code away. Found by the round-3 audit, not by me. It is corrected in `60cd8943ba`, and this note is here rather than an edit to the earlier comment because a reviewer may already have read and believed it.

---

**Audit round 3** — delta re-audit, read at `cd549b56`. Seven findings (no 🔴), all addressed in `60cd8943ba`. NEW-1 was a regression the round-2 fix itself introduced.

legend: `<from>` = the tip this round's audit READ · `<to>` = the head this round's FIXES produced.

```audit-claims round=3 audited=cd549b560de989312149a982a02ddea9897f6e6f..60cd8943
1. NEW-1 — `buildScalarPatch` now emits `betaMessage` only while `current.isBeta` is true, so the tick→type→untick sequence emits no beta keys and cannot fail the save. Turning beta off still clears an existing note, because the server nulls it on the off transition — the note never has to ride the patch to be removed.
2. NEW-1 — five tests added in `offsiteEditConfig.test.ts`: the abandoned-note case, an unrelated edit surviving the same save, a ticked-box positive control, the untick-clears case, and note-edited-while-on.
3. NEW-2 — three tests added for `updateRevisionDraft`'s parent-flag read, including the discriminating PAIR (a note-only patch SUCCEEDS when the parent is in beta, and is REFUSED when it is not) plus an assertion that the guarded read is keyed on `revisionOfId`. The survived mutant was reproduced first, then killed.
4. NEW-3 — the degradation log's severity is corrected from `'warning'` to `'error'`, and its rationale rewritten: the missing-COLUMN case it used to cite cannot reach `readForRender`'s catch at all (the narrow readers swallow `P2022`/`42703` upstream and return without throwing). The reachable set is timeout / `42P01` / `42501` / connection failure — server faults, which `~/server/logging/client` classifies as `'error'`. The flooding argument that justifies `'warning'` in `app-listing-assets.service` does not apply: nothing an author can do produces a `42P01`.
5. NEW-4 — the migration header now scopes the `*Many` exemption to the RETURNING list, names `createManyAndReturn` / `updateManyAndReturn` as row-returning, and warns that they contain the exempt names as substrings. See the retraction above.
6. NEW-5/6/7 — four comments (one more than the audit named) still asserted `beginListingRevision` clones beta onto the shadow: the `preview` omission-ledger entry in `AppListingDetailBody.tsx` and its browser test, `applyApprovedRevision`'s off-site branch, and `loadListingEditView`'s prefill note. All corrected. The round-trip fixture's clone-time value is relabelled as the deliberate structural pin it now is, and the `updateListing` comment that said the apply "would restore the clone-time value" now states the real mechanism.
7. Verification — 6 round-4 mutants killed by their own named tests, including BOTH directions on the NEW-1 guard (removed → abandoned note emitted; over-applied → note never emitted) and the NEW-2 pair (hardcoded false and hardcoded true each kill a different test).
8. Verification — `pnpm typecheck` 0 errors; full unit 1549 files / 24478 tests with only the two known-red base failures and ZERO import failures; component `src/components` 191/191, `src/libs` 1/1, `src/hooks` 2/2, `src/components/Apps` 77 files / 1148 tests.
9. A stale test NAME was also fixed: the severity test still read "at WARNING" while asserting `'error'`.
```

**Still not fixed, deliberately:** the `~92` sentence in `app-listing-source-repo.service.ts` (out of this PR's scope); `getMyListingForApp`'s assets-only path still pays `readListingSourceRepoUrl` whose result it discards (pre-existing); the `.catch()` docstring's "REJECTS when Axiom is degraded" phrasing, which the round-3 audit notes is repo-wide across sibling modules rather than introduced here.

ZacxDev:
## 🔴 Second retraction — the round-3 claims block was also over-stated

Round 3's claim 6 said four stale clone comments were found and *"All corrected."* **There were five.** The fifth — `applyApprovedRevision`'s ONSITE branch — survived because I fixed the sites one particular grep turned up instead of sweeping the class. That is the same failure in kind as the round-2 retraction above: a confident claim about coverage I had not established. Corrected in `91c59b42ab`, and this round the class was swept **mechanically** (19 candidate mentions across `src/` and `packages/`, each read) rather than from recall.

---

**Audit round 4** — delta re-audit, read at `60cd8943`. Three findings (one 🟡, two 🟢, no 🔴), all addressed in `91c59b42ab`.

legend: `<from>` = the tip this round's audit READ · `<to>` = the head this round's FIXES produced.

```audit-claims round=4 audited=60cd8943ba7a0243d98fa4b1c998ef807f91fec5..91c59b42
1. F1 — `applyApprovedRevision`'s ONSITE branch no longer says the shadow's beta copy "is a display snapshot that is allowed to go stale (see beginListingRevision)". It now states the true mechanism (a shadow's beta columns are never written, so copying would CLEAR the live value rather than revert it) and adds an explicit "do not freshen the shadow to fix that", naming the ordering constraint that re-adding a clone would re-impose.
2. F1 — the class was swept mechanically across `src/` and `packages/` rather than by recall. Two other clone mentions are retained deliberately because both are explicitly framed COUNTERFACTUALS (`splitBetaPatch`'s "THE ALTERNATIVE … IS WRONG", and the preview's "reading them would show …" immediately followed by "THIS REPLACED A CLONE"); the remaining hits concern the owner column or assets, not beta.
3. F2 — the NEW-1 fixture is rebuilt through the file's typed `makeCtx` instead of a hand-rolled literal cast `as never`, and the two inline re-casts are replaced by a shared `ON_LISTING`. No behaviour change; five tests still pass.
4. F3 — "unticking a listing that HAS a note still clears it" is renamed and labelled an INVARIANT GUARD, with its comment corrected: it asserts only that the CLIENT omits the key, not the server-side clear, which is covered by the N5 block in `offsite-listing.edit.service.test.ts`.
5. A defect introduced WHILE fixing F2 and caught before commit: the new docstring quoted the tsconfig exclude glob, whose `**` followed by a slash CLOSES a block comment — the whole suite then failed to import and the runner reported `Tests no tests`, which reads as a pass. Found by reading the test COUNT rather than the verdict. Every changed file in the PR was then swept for that sequence; there are none.
6. Verification — `pnpm typecheck` 0 errors; full unit 1549 files / 24478 tests, only the two known-red base failures, ZERO import failures and zero `no tests` lines; component `src/components` 191/191, `src/libs` 1/1, `src/hooks` 2/2. No new mutation battery this round: the round's only executable change is zero lines — F1 is a comment, F2/F3 are test quality — so there is no new guard to mutate.
```

**Ledger note, stated because it is the gate the skill asks for:** this round's fixes changed **no executable payload lines** (12 comment lines in one source file, plus test-quality edits). Round 3's changed two. If round 5 returns only prose or scaffolding findings again, that is the signal the ladder has left the PR, and I will say so explicitly rather than continuing.

ZacxDev:
## 🔴 Third retraction, and its root cause

Round 4's claim 3 said the two inline `as never` re-casts *"are replaced by a shared `ON_LISTING`."* **They were not.** `ON_LISTING` was declared and never referenced; both casts survived verbatim, and the dead const was producing an eslint warning. Fixed in `f93a55b3aa`.

That is the third over-stated claim in this ladder (round 2: a file "corrected" that was never touched; round 3: "four comments … All corrected" when there were five; round 4: this one). **All three have the same root cause** — I wrote the claim from what the edit was *intended* to do, after a string-replace had silently failed to match, instead of from what the tree showed afterwards. The correction I applied this round, and should have applied from the start: after claiming a removal, grep for the removed thing and report the count. For this fix — `ON_LISTING` now has 5 references, `as never as { scalars: object }` has 0, `const on = {` has 0.

---

**Audit round 5** — delta re-audit, read at `91c59b42`. Three findings (one 🟡, two 🟢, no 🔴), all addressed in `f93a55b3aa`. The auditor's own assessment: *"All three are comment/test hygiene; none has a runtime consequence."*

```audit-claims round=5 audited=91c59b42ab798a44b1495e5b709370c09b777edd..f93a55b3
1. N1 — `ON_LISTING` is now used at both sites and the two `as never as { scalars: object }` casts are gone. Verified by grep, not by intent: 5 refs / 0 casts / 0 stray `const on = {`, and the eslint unused-var warning is gone.
2. N1 — re-verified after the fixture rebuild that the five NEW-1 tests still discriminate. The kill matrix is UNCHANGED: mutation A (guard removed) is killed by tests 1 and 2, mutation B (note never emitted) by tests 3 and 5, and test 4 survives both — so the shared fixture made nothing vacuous and the invariant-guard label remains correct in both directions.
3. N2 — `getListingPreviewForReview`'s comment no longer says reading a shadow's beta columns "goes stale". It now states the stronger truth: nothing writes them, so they hold the SCHEMA DEFAULTS for every shadow always, and keying the read on the shadow id would strip the badge and notice from EVERY revision preview rather than showing a stale value. Corrected at its echo in `app-listing.service.test.ts` too.
4. N3 — the invariant-guard comment now cites the F1 block's `turning beta OFF clears the note in the SAME write`, which pins the server clear for the `{isBeta:false}`-alone shape the client actually sends, and explains why N5 (its previous, wrong citation) is a different rule.
5. Verification — `pnpm typecheck` 0 errors; full unit 1549 files / 24478 tests, only the two known-red base failures, ZERO import failures and zero `no tests` lines; component `src/components`+`src/libs`+`src/hooks` 194 files / 2184 tests, all passing.
```

---

## 🛑 This is the last round, and here is why — stated rather than left implicit

The findings-keyed rule would normally schedule a round 6. I am stopping instead, on the attribution gate, and recording the reasoning so that "the ladder ended" is distinguishable from "real findings were left unfixed":

- **Two consecutive rounds have changed zero executable payload lines.** Round 3's fixes changed 2 (a conditional and a severity literal); round 4's changed 0 (one comment); round 5's changed 0 (one comment + test hygiene). The ladder is now auditing prose that the ladder itself wrote.
- **No 🔴 in any of the five rounds**, and the last two rounds' findings had no runtime consequence at all — the round-5 auditor says so explicitly. The blast radius of everything outstanding is "a sentence in a comment is wrong".
- **The recurring shape was swept mechanically, not at the reported site.** The stale-clone-prose class was re-swept this round with the auditor's own re-runnable method (`grep -rniE "clone|cloned|clones|cloning" … | grep -i beta`, 15 hits, each read); the two remaining mentions are past-tense history and an explicitly-labelled counterfactual.
- **Nothing from round 5 is being left unfixed.** All three findings are addressed in `f93a55b3aa`.

**Recorded as open, deliberately not fixed** (so the next reader knows these are open rather than absent):
- `app-listing-source-repo.service.ts` still carries the `~92` sentence this PR corrected in its own module — out of scope, that file is otherwise untouched here.
- `AppListingDetailBody.tsx:129-132` cites `APPS_READABLE_PAGE_WIDTH` / `APPS_TWO_COLUMN_DETAIL_PAGE_WIDTH`, which no longer exist (now `…_MEASURE`, different values). Pre-existing on `main`, on lines this PR never touched.
- `AppListingDetailBody.viewer.browser.test.tsx:34` says its sibling suite has "(64)" tests; it has 98. Pre-existing (83 at the PR base); this PR widens the drift to 89 declarations without causing it.
- `checks.x86_64-linux.dev-scripts` builds rc=0 but its retained nix log is empty, so no verdict was read from it. None of the four nix checks reads a file this PR touches.

PR #4549, opened 2026-09-01
feat(apps): show a content-shaped loading skeleton while an app launches

ZacxDev:
The `/apps/run` launch overlay showed a bare spinner. The sidebar slot never has — `IframeHost` renders `BlockFallback`'s `<Skeleton>` while its block loads — so the two block surfaces disagreed about what a loading app looks like, and the run page was the one saying *"busy"* rather than *"content is coming"*.

This swaps the `<Loader>` for shimmer bars. Every app in the store gets it, including ones that will never be touched again.

## Why the host, and not each app's own `index.html`

This started as a static skeleton shipped inside one app's HTML. That approach cannot work here, and the reason is worth recording:

> the overlay is `position: absolute; inset: 0; background: var(--mantine-color-body)` at `opacity: 1` until `BLOCK_READY`, then cross-fades

So anything an app paints before `BLOCK_READY` is behind an opaque veil and is never seen — and by the time the veil fades, the block is ready and showing real content. A per-app skeleton is invisible on this surface no matter how well built.

Putting it in the host also makes it free: no per-app change, no rebuild, no SDK bump, nothing to re-publish.

## What is unchanged

The branded half of the launch state — app initial, the `Starting …` / `Retrying …` copy, the sanitizer they both derive from, and the `BLOCK_READY` cross-fade. Only the activity indicator changed.

The bars are deliberately **generic** rather than a mimic of any one app's layout: this renders for every app in the store, so a shape borrowed from one would be wrong for the rest.

## a11y is held, not traded

The accessible name moves onto the skeleton group (`role="img"` + `aria-label`) rather than being dropped along with the spinner that used to carry it, so the labelled graphic stays in the tree next to the container's `role="status"` / `aria-busy`. The existing assertion on that name still passes untouched.

`animate={!reduceMotion}` is the same call `BlockFallback` makes — the bars stop shimmering under `prefers-reduced-motion` rather than animating indefinitely behind a veil.

## Tests

Two new guards, both mutation-checked — each watched to fail, for its own reason:

| mutant | killed |
|---|---|
| revert to a bare spinner | both new tests |
| spinner re-added **alongside** the skeleton (half-done partial revert) | only "not a spinner" |
| hardcode `animate={true}` | only the reduced-motion test |

That middle row is the point: without the `Loader`-is-gone assertion the suite would pass with both rendered.

`PageBlockHostLaunchReveal` 12/12 · sibling host suites (`PageBlockHost`, `PageBlockHostAutoRetry`) 46/46 · `typecheck` clean.

One pre-existing `local-rules/no-wholesale-module-mock` error on the test file is untouched — it reproduces identically on `main` and is out of scope here.

https://claude.ai/code/session_01U2wcms7C2Lg9GrHPmBz8oH

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -756,13 +756,14 @@ describe('PageBlockHost loading indicator (Task 1)', () => {
     renderWithProviders(<PageBlockHost {...baseProps} onConsentGranted={vi.fn()} />);
 
     // While status === 'loading' (iframe mounted, pre-handshake) the centered
-    // Loader overlay is present so the surface isn't blank.
+    // launch overlay is present so the surface isn't blank.
     await expect.element(page.getByTestId('app-page-loading')).toBeInTheDocument();
-    await expect.element(page.getByLabelText('Loading Budgeted Generator')).toBeInTheDocument();
+    await expect.element(page.getByTestId('app-page-loading-skeleton')).toBeInTheDocument();
 
     // a11y: the overlay container is marked as a busy live region (role="status"
-    // + aria-busy) so a screen reader announces the loading state on the REGION,
-    // not just the labeled graphic. Assert the attributes while still loading.
+    // + aria-busy) so a screen reader announces the loading state on the REGION.
+    // That region is the only announcing element — the skeleton group inside it
+    // is aria-hidden. Assert the attributes while still loading.
     const overlay = page.getByTestId('app-page-loading').element();
     expect(overlay.getAttribute('role')).toBe('status');
     expect(overlay.getAttribute('aria-busy')).toBe('true');
@@ -777,19 +778,25 @@ describe('PageBlockHost loading indicator (Task 1)', () => {
     });
   });
 
-  test('the aria-label runs appName through the chrome sanitizer (control/bidi chars stripped)', async () => {
+  test('the launch copy runs appName through the chrome sanitizer (control/bidi chars stripped)', async () => {
     // A publisher-controlled name carrying a control char (\t) and a bidi
-    // override (U+202E) must NOT reach the accessible name verbatim — it goes
+    // override (U+202E) must NOT reach the launch surface verbatim — it goes
     // through the SAME sanitizeAppChromeName the visible chrome uses: the \t
     // control char becomes a space and the U+202E format char is dropped →
     // 'Evil App' (verified against the sanitizer's documented behaviour +
     // appChromeName.test.ts). Built from char codes so the source carries no
     // literal invisible chars.
+    //
+    // Asserted on the VISIBLE launch copy. This used to read the overlay's
+    // aria-label, which was removed when the loading indicator became an
+    // aria-hidden skeleton — the sanitizer contract is unchanged and is what
+    // this pins, so it moves to the surface that still carries the name rather
+    // than being dropped with the attribute.
     const spoofedName = 'Evil' + String.fromCharCode(0x09) + String.fromCharCode(0x202e) + 'App';
     renderWithProviders(
       <PageBlockHost {...baseProps} appName={spoofedName} onConsentGranted={vi.fn()} />
     );
-    await expect.element(page.getByLabelText('Loading Evil App')).toBeInTheDocument();
+    await expect.element(page.getByText('Starting Evil App…')).toBeInTheDocument();
   });
 
   test('the error terminal path shows the fallback and never the loader (does not spin forever)', async () => {
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -3801,10 +3801,14 @@ export function PageBlockHost({
           {overlayMounted && (
             <Center
               data-testid="app-page-loading"
-              // Announce the loading state on the REGION, not just the graphic:
-              // role="status" + aria-busy mark the overlay container as a live
-              // busy region so a screen reader announces "loading" when it
-              // appears (the skeleton group below only exposes a labeled graphic).
+              // Announce the loading state on the REGION: role="status" +
+              // aria-busy mark the overlay container as a live busy region so a
+              // screen reader announces it when it appears. The region is the
+              // ONLY thing that announces — the skeleton group below is
+              // aria-hidden and exposes nothing, deliberately (see its own
+              // comment). Do not give that group a role to "restore" a labelled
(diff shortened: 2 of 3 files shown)


ZacxDev:
**Correction to the mutation table in the PR description above — it is wrong, and I'd rather correct it here than quietly edit the body someone may already have read.**

Row 2 claims the "spinner re-added **alongside** the skeleton" mutant is killed by *only* the "not a spinner" test. Re-run against the tree, it kills **two**: that one (139ms, as claimed) **and** the pre-existing `the launch state is BRANDED` test, because `getByLabelText('Loading Budgeted Generator')` then matches two elements. The second kill arrives as a **15.2s timeout**, not a fast assertion failure.

The row's conclusion still holds — the `Loader`-is-gone assertion is load-bearing and that mutant does not survive — but the count and the cost were misstated.

An adversarial audit of `9d5a8c388c` also found two real defects in the change itself, both fixed in `693d3da891`:

1. **`maw={420}` was inert and the skeleton's width was publisher-controlled.** The parent `Stack` is a shrink-to-fit flex item, so `w="100%"` on the group resolved against the *app name's* rendered text width. Measured at 414px: `Budgeted Generator` → 193px group; `Ap` → **82.5px, with 27.8px bars**. Fixed with `w="100%"` on the Stack — `Ap` now measures 414px like any other name.
2. **The reduced-motion guard had no positive control.** It only asserted `data-animate` was *absent*, which `animate={false}` on all four bars also satisfies — a build where nobody ever sees the shimmer passed 12/12. And the comment claimed the shimmering case was "pinned in the test above" when it was not. Now asserted in both directions, matching `BlockHostTokenRetry.browser.test.tsx:367`/`:391`.

Also changed: the skeleton group is now `aria-hidden` rather than `role="img"`. Inside the overlay's `role="status"` live region a labelled child img made the app name announce twice; the `<Loader>` it replaces never did, since Mantine renders it as a bare `<span>` with no role.

**One thing left deliberately unguarded, stated so it reads as open rather than absent:** nothing pins the rendered width. The component harness loads no Mantine CSS, so `var(--mantine-scale)` is undefined and every box metric computes to 0 — reverting the `w="100%"` fix survives the suite. It was verified by direct measurement with the stylesheet loaded, and that measurement is not shipped.

ZacxDev:
Audit round 1 claims, for the next round to check against the tree.

```audit-claims round=1 audited=9d5a8c388c..693d3da8
1. `w="100%"` added to the launch-overlay Stack (PageBlockHost.tsx) — claimed to make the skeleton group's `w="100%" maw={420}` resolve against the viewport rather than the app-name text width, so bar widths no longer vary per app.
2. `expect(bar.getAttribute('data-animate')).toBe('true')` added to the non-reduced-motion test — claimed to close the surviving `animate={false}` mutant.
3. The comment claiming "the shimmering case is pinned in the test above" was rewritten — claimed to no longer assert coverage the tests do not provide.
4. Skeleton group changed from `role="img"` to `aria-hidden` (aria-label retained) — claimed to remove a duplicated app-name announcement inside the overlay's role="status" live region, without changing what the pre-existing label assertion pins.
5. PR-description mutation-table row 2 corrected in a PR comment rather than a body edit — claimed to state that the mutant kills two tests, the second as a 15.2s timeout.
6. Claimed unguarded and stated as such: nothing pins the rendered width, because the component harness loads no Mantine CSS.
```

ZacxDev:
**Round 2. The PR description's a11y section is now FALSE at head — correcting it here rather than editing the body.**

The description still says:

> ## a11y is held, not traded — The accessible name moves onto the skeleton group (`role="img"` + `aria-label`) … so the labelled graphic stays in the tree

Head does the **opposite**, deliberately. The group is `aria-hidden` with no role and no label. Read the commits, not that paragraph. The description's mutation table is also incomplete — it has no row for the `animate={false}` mutant, which round 1 added a guard for.

Round 2 found one class of defect, fixed in `996d7dfbee`:

1. **Round 1's a11y fix had no guard.** Mutating `aria-hidden` back to `role="img"` passed **12/12** — the defect round 1 had just removed was silently reintroducible. Cause: `getByLabelText('Loading …')` does **not** filter `aria-hidden`, so the assertion passed identically whether the group was an exposed labelled graphic or a hidden decorative box. It read as a11y coverage while pinning nothing. Now asserted as state (`aria-hidden` true, `role` null, `aria-label` null); that mutant dies.
2. **The `aria-label` was inert and shipped only as a test handle.** On an `aria-hidden` node it can never be announced, and the element already carries `data-testid`, which is what the suite queries. Removed. Three call sites depended on it — two were the vacuous assertions above; the third pins a real anti-spoof contract (`sanitizeAppChromeName`) and moved to the visible launch copy rather than being deleted with the attribute. Mutation-verified: bypassing the sanitizer kills both sanitizer tests.
3. **Two comments contradicted the code.** The live-region comment still described the group as exposing "a labeled graphic" — an instruction to restore the very `role` that causes the double announcement. And the round-1 comment justified `aria-hidden` by claiming a live region announces "its whole text content", which is false in the direction that argues *against* the fix (the region's own `aria-hidden` Avatar contributes "B" to `textContent` and is not announced).

58/58 across LaunchReveal + PageBlockHost + AutoRetry · typecheck 0 · prettier clean · eslint unchanged (two pre-existing `no-wholesale-module-mock` errors, one per test file, both on `main`).

**Still unguarded, stated so it reads as open rather than absent:** the rendered width. The component harness loads no Mantine CSS, so box metrics are 0 and reverting `w="100%"` survives the suite. Verified by direct measurement instead.

```audit-claims round=2 audited=693d3da891..996d7dfbee
1. Skeleton group a11y state asserted directly (aria-hidden === 'true', role === null, aria-label === null) in PageBlockHostLaunchReveal — claimed to kill the `aria-hidden` → `role="img"` mutant that previously survived 12/12.
2. `aria-label` removed from the skeleton group in PageBlockHost.tsx — claimed to be inert on an aria-hidden node and redundant with the existing data-testid.
3. PageBlockHost.browser.test.tsx sanitizer test re-pointed from getByLabelText to the visible "Starting …" copy — claimed to preserve the sanitizeAppChromeName contract rather than drop it with the attribute.
4. PageBlockHost.browser.test.tsx loading assertion re-pointed from getByLabelText to the skeleton testid — claimed equivalent coverage without depending on the removed label.
5. Live-region comment at the Center rewritten — claimed to no longer describe the group as exposing a labelled graphic.
6. The "whole text content" phrase in the skeleton-group comment rewritten to accessible-tree text — claimed to no longer state a mechanism under which the fix would be ineffective.
7. Claimed still unguarded: the rendered width, because the component harness loads no Mantine CSS.
```

PR #4550, opened 2026-09-01
fix(civitai-link): don't call a Link host that can't authenticate this origin

ZacxDev:
## The bug

Civitai Link fails on every PR preview with:

```
Error loading instances: Civitai Link request failed (401 )
```

It works on production, which is what makes it look mysterious.

## Why

The preview build bakes `NEXT_PUBLIC_CIVITAI_LINK=https://link.civitai.com`, but a preview is served from `pr-N.civitaic.com`. The Link service authenticates **only** via the civitai session cookie, and that cookie is `Domain`-scoped to the page's registrable domain (`cookieDomainForHost`, `~/server/auth/civ-cookie`) and set `SameSite=Lax`. So on a preview:

- the preview's own cookie is scoped to `civitaic.com` and never reaches `link.civitai.com`;
- a `civitai.com` cookie sitting in the same browser is withheld anyway, because a cross-site `fetch()` is not a Lax-eligible request;
- and a preview's civ-token is minted by a different auth hub than the one the Link service verifies against, so it would be rejected even if it arrived.

The request is unauthenticated by construction, and 401s every time.

Two incidental details that made this hard to read:

- it surfaces as a *readable* 401 rather than a CORS error, because the Link service reflects any origin (`@fastify/cors` with `origin: true, credentials: true`);
- the empty `statusText` in `(401 )` is just HTTP/2.

## The fix

`getCivitaiLinkBaseUrl` already had a rewrite for exactly this failure mode on the `.com`/`.red` split — but it only covers the one other color we happen to run a Link host for. This generalises it: after resolving the host, refuse to return a URL whose registrable domain differs from the page's.

That also covers **`civitai.green`**, which has no `link.civitai.green` host and has been silently 401ing the same way. (Verified: `link.civitai.com` and `link.civitai.red` resolve and answer 200 on `/healthcheck`; `link.civitai.green` does not resolve.)

With no reachable Link host, `CivitaiLinkProvider` no longer spawns the SharedWorker at all — it would only produce the confusing 401 and a socket aimed at a host that will never accept us — and shows `Civitai Link is not available on this domain` instead.

The check lives in `getWorker` rather than in render, and `getCivitaiLinkBaseUrl` keeps returning the baked value when there is no `location` (SSR/node), so SSR and hydration cannot disagree about whether the feature exists. There are only three call sites and none is in a render path.

## Tests

Four new cases (PR preview host, bare `civitaic.com`, `civitai.green`, and the lookalike `civitai.red.evil.com`), plus a new `internal.civitai.com` case guarding against over-refusal.

- **Red at `origin/main`**: 4 failed / 6 passed, failing with the bug's exact signature — `expected 'https://link.civitai.com' to be undefined`.
- **Green here**: 10/10.

The six pre-existing cases passing in the same red run are the positive control that the suite is actually wired to the code.

`pnpm typecheck` is clean (0 errors). ESLint reports 0 errors across the three changed files; the single warning is the pre-existing `createContext<CivitaiLinkState>({} as any)`.

## Note on scope

This makes the failure honest rather than making Link *work* on previews. Genuinely working Link on a preview would need a Link service on `*.civitaic.com` that trusts the preview auth hub — a real deployment, not a config tweak.

https://claude.ai/code/session_018DAdpNNx6bpcp3F5ZNHbmo

### DIFF
--- a/src/components/CivitaiLink/CivitaiLinkProvider.tsx
+++ b/src/components/CivitaiLink/CivitaiLinkProvider.tsx
@@ -262,7 +262,11 @@ const Provider = ({ children }: { children: React.ReactNode }) => {
       return { promise: Promise.resolve(undefined), id: payload.id, cancel: () => {} };
     }
 
-    // Setup promise for later resolution
+    // Setup promise for later resolution. Note the `timeout` clock now arms
+    // AFTER the worker is ready rather than before it — `getWorker()` above is
+    // the wait that moved. Unobservable today (`timeout` is not on the context
+    // type and no caller passes one), but it is the semantics whoever first
+    // passes a timeout will get.
     const promise = new Promise((resolve, reject) => {
       commandPromises[payload.id] = { resolve, reject };
       if (timeout <= 0) return;
@@ -273,7 +277,12 @@ const Provider = ({ children }: { children: React.ReactNode }) => {
       }, timeout);
     });
 
-    worker.port.postMessage({ type: 'command', payload });
+    // Deliberately still routed through `workerReq`, not a direct
+    // `worker.port.postMessage(...)`: `postMessage` takes `any`, so posting
+    // directly drops the `WorkerIncomingMessage` check on the one message that
+    // carries every resource add/remove/cancel. `getWorker()` is idempotent, so
+    // the second call here is just the cached promise.
+    await workerReq({ type: 'command', payload });
     const cancel = () => {
       if (!commandPromises[payload.id]) return;
       runCommand({ type: 'activities:cancel', activityId: payload.id });
--- a/src/components/CivitaiLink/CivitaiLinkProvider.tsx
+++ b/src/components/CivitaiLink/CivitaiLinkProvider.tsx
@@ -249,6 +249,19 @@ const Provider = ({ children }: { children: React.ReactNode }) => {
     const payload = command as Command;
     payload.id = uuid();
 
+    // No reachable Link host: deliver nothing and hand back an already-settled
+    // promise. Registering in `commandPromises` first would leak the entry
+    // forever — nothing can complete it, and the default `timeout = 0` arms no
+    // rejection timer, so `.promise` would stay pending for the page's life.
+    // Settle rather than reject: callers `await runCommand(...)` (the outer
+    // call) and none attach a handler to `.promise`, so a rejection here would
+    // surface as an unhandled rejection.
+    const worker = await getWorker();
+    if (!worker) {
+      setError(UNAVAILABLE_ON_DOMAIN);
+      return { promise: Promise.resolve(undefined), id: payload.id, cancel: () => {} };
+    }
+
     // Setup promise for later resolution
     const promise = new Promise((resolve, reject) => {
       commandPromises[payload.id] = { resolve, reject };
@@ -260,7 +273,7 @@ const Provider = ({ children }: { children: React.ReactNode }) => {
       }, timeout);
     });
 
-    await workerReq({ type: 'command', payload });
+    worker.port.postMessage({ type: 'command', payload });
     const cancel = () => {
       if (!commandPromises[payload.id]) return;
       runCommand({ type: 'activities:cancel', activityId: payload.id });
--- a/src/components/CivitaiLink/CivitaiLinkProvider.tsx
+++ b/src/components/CivitaiLink/CivitaiLinkProvider.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import { createContext, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import type { CivitaiLinkInstance } from '~/components/CivitaiLink/civitai-link-api';
+import { getCivitaiLinkBaseUrl } from '~/components/CivitaiLink/civitai-link-api';
 import type {
   Command,
   ResponseResourcesList,
@@ -64,6 +65,12 @@ type CivitaiLinkState = {
 };
 
 const CivitaiLinkCtx = createContext<CivitaiLinkState>({} as any);
+
+// Shown when the current origin can't reach a Link host that would accept its
+// session cookie (see `getCivitaiLinkBaseUrl`). Distinct from the
+// 'Civitai Link is not enabled' flag message: the feature IS enabled for this
+// user, it just cannot work from this domain.
+export const UNAVAILABLE_ON_DOMAIN = 'Civitai Link is not available on this domain';
 // #endregion
 
 // #region zu store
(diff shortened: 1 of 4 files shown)


ZacxDev:
### Adversarial audit — round 1

Round 1 returned **5 findings, no 🔴**. All of the author's stated verification claims (red-at-base with the named assertion, typecheck, ESLint, the `link.civitai.green` DNS check) were independently re-derived and confirmed. Every finding is addressed in `e7d8e484`.

The headline one was a coverage regression this PR itself introduced: the lookalike test it replaced was the only thing pinning the narrowness of the `.red` matcher, and the replacement was satisfied by the new domain check either way. Confirmed by mutation — widening `endsWith('.civitai.red')` to `includes('civitai.red')` left the whole suite green.

**Round-2 ledger** · payload lines changed THIS round: **37** (`CivitaiLinkProvider.tsx` 14/1, `civitai-link-api.ts` 19/3) · scaffolding: 115 · so the attribution gate does not fire.

```audit-claims round=1 audited=bb12c985261d285782cc5409c917d016246b5a5b..e7d8e484
1. Restored a guard on the narrowness of the `.red` matcher, on a test host that stays on the `.com` registrable domain so the same-registrable-domain check cannot mask the result.
2. Added `CivitaiLinkProvider.browser.test.tsx` covering the provider's no-worker branch, the domain error it sets, the resulting status, and a reachable-host positive control.
3. `runCommand` no longer registers an entry in `commandPromises` when no worker exists; it returns an already-settled promise rather than one that can never settle.
4. Rewrote the `registrableDomain` docstring, which claimed a `cookieDomainForHost` mirror that does not hold for `localhost` and bare IPv4, and recorded the one imprecision found while checking it.
5. A malformed `NEXT_PUBLIC_CIVITAI_LINK` is now reported as a config error instead of sharing the unreachable-domain message.
```

Delta re-audit of `bb12c98526..e7d8e484` to follow.

ZacxDev:
### Adversarial audit — round 2 (delta)

Round 2 returned **6 findings, no 🔴**, and dispositioned round 1's five claims as: 3 fixed, 1 partial, 1 **not fixed**. Its headline finding was a regression the round-1 *fix* round introduced — the thing a delta round exists to catch. All six are addressed in `a5b52b55`.

Two of round 1's fixes turned out to be worse than claimed: the new provider test included one assertion that was a constant of its own fixture (it passed with the component not mounted at all), and the "config error" logging added for a malformed `NEXT_PUBLIC_CIVITAI_LINK` sat on an unreachable branch justified by a false statement about dev-mode env validation. That claim is **withdrawn** rather than re-argued.

**Round-3 ledger** · payload lines changed THIS round: **27** (`CivitaiLinkProvider.tsx` 11/2, `civitai-link-api.ts` 16/10) · scaffolding: 28 · since round 1: 95. Worth naming: of those 27, **26 are comment/docstring and 1 is executable** — the executable line is F1's restored type gate. The attribution gate does not fire, but the trend is noted.

```audit-claims round=2 audited=e7d8e484a12fd5ca907f645778b2c22fe3a44c32..a5b52b55
1. Restored `workerReq` on the command path, so the `command` message is type-checked again rather than going through an `any`-typed `postMessage`.
2. Removed the test whose assertion was a constant of its own fixture, folding its one real assertion into the test that already pins the branch.
3. The unavailable-domain message is now pinned as a literal, with the exported constant asserted equal to it, so emptying the constant is red.
4. The FeatureFlagsProvider mock now spreads `importOriginal` instead of replacing the module with one of its four exports.
5. Withdrew round 1's malformed-URL logging and replaced its false rationale with why that branch is unreachable.
6. Recorded the `timeout`-arming order change, and added single-label hosts to the docstring's list of divergences from `cookieDomainForHost`.
```

Delta re-audit of `e7d8e484a1..a5b52b55` to follow.

ZacxDev:
### Adversarial audit — round 3 (delta) · **ladder stopped**

Round 3 returned **no 🔴, no 🟡, and one 🟢** — a comment-completeness nit — and dispositioned all six of round 2's claims as *actually fixed*, each with a killing mutant watched to go red. The 🟢 is fixed in `23a2e4d6`, comment-only.

**Why the ladder stops here rather than running a round 4.** The remaining finding was a documentation nit with no behavioural, decision or correctness consequence to the code, and its fix is one clause. Of this round's 27 payload lines, **26 were comment/docstring and 1 executable**; the round-3 auditor independently reached the same judgement — *"a round 4 would be reviewing sentences about sentences."* Continuing would audit the ladder's own prose, not this PR. Blast radius of anything left: a comment could be more complete. No 🔴 at any point in three rounds.

**Ledger** · round 4 · payload lines changed THIS round: **7** (all comment) · since round 1: 114.

### One correction to round 3's report

Its top "could not verify" was whether any CI tier runs the new `*.browser.test.tsx`; scoped to `.github/workflows/`, it concluded none does. **Tekton does** — `clusters/production/apps/tekton-builds/pr-preview-pipeline.yaml` runs `pnpm run test:component` and posts `preview / component-tests` as a soft status. So the test executes in CI, report-only, which is exactly what the test file's own comment claims. That pipeline is in a private infra repo the auditor had no access to.

Also correcting my own round-3 commit message: it said ESLint ran over "all 4 files (count asserted at 4)". Four files were linted and the results were as reported, but that round's *range* touched 3.

### Recorded as OPEN, deliberately not fixed — so the next reader knows these are open rather than absent

1. **`setError(UNAVAILABLE_ON_DOMAIN)` in `runCommand` has no killing test.** Deleting it leaves the suite green, because no test calls `runCommand`. It is round-1 payload and redundant with the identical call in `boot` (React bails on identical state), so its absence changes nothing observable — but it is an unpinned line and I am not claiming otherwise.
2. **The popover still leads with "Cannot Connect".** `UNAVAILABLE_ON_DOMAIN` is appended as a third line rather than replacing the generic headline, so users on `civitai.green` and PR previews still see the wrong *primary* diagnosis. That is round-1 behaviour, outside every audited range, and is a copy/product decision rather than a defect in this change.
3. **Drifted counts in `featureFlagsMockCompleteness.test.ts`** (its prose says 76/14/62/52; re-derived at this head with its own extractor: 79/15/64/54). An unchanged file, prose only, already stale before this PR, and its guard asserts nothing about those numbers.

None of the three has an owner or a mechanical closing condition I can name, so I am recording them here rather than opening issues nobody could close.

```audit-claims round=3 audited=a5b52b555fd43715d2791758d7db05e8fb07de89..23a2e4d6
1. Named the zod trim write-back that the catch's unreachability actually rests on, and said to keep the catch rather than read it as dead code.
```

ZacxDev:
### Rebased onto `main` — and a note on what is and isn't audited

Rebased onto `a831825e` and force-pushed; head is now **`ea4442e1`**, 0 commits behind. All 18 checks terminal, 0 not-passing, `MERGEABLE` / `CLEAN`.

⚠️ **The rebase rewrote every sha the audit comments above cite** (`bb12c98526`, `e7d8e484a1`, `a5b52b555f`, `23a2e4d6`). Those `audit-claims` ranges no longer resolve in this branch's history. The findings they record still stand — only the identifiers are stale.

**What is NOT audited, stated plainly:** the last adversarially-audited tip was `a5b52b55`. Since then the branch gained the round-3 comment fix and this rebase. **The current head has had no audit pass.** That was a deliberate call, not an oversight — the delta is one comment clause plus a replay onto 18 upstream commits that touched none of the files here.

The rebase was verified mechanically rather than by audit:
- `main` changed none of `src/components/CivitaiLink/**`, `src/workers/**`, `src/env/**`, or `civ-cookie.ts`.
- Full local suite re-run on the rebased tree: typecheck 0 errors, api unit 11/11, provider browser 3/3, `test:lint-rules` 433/433, ESLint 0 errors, prettier clean.
- Lockfile, `event-engine-common` pointer and `vitest.config.mts` all unchanged by the rebase; the only `package.json` delta was a version bump.

**One false alarm worth recording**, since it would mislead anyone repeating this: typecheck on the rebased tree first reported **11 errors** in `app-listing.service.ts`, `offsite-listing.service.ts` and `apps/run/[slug]`. They were not from this PR. Running `origin/main` alone produced the **identical 11**, and regenerating the Prisma client produced **0**. The cause was a generated client 18 commits stale — the lockfile was unchanged, so "no reinstall needed" was the wrong freshness test, because the client tracks the schema, not the lockfile. `main` is healthy.

**Ladder status: stopped after 3 rounds — no round ever returned zero findings.** It stopped on the stated prose criterion (round 3's payload was 26 comment lines to 1 executable), not on convergence. No 🔴 at any point; nothing 🟡 outstanding. The three items recorded as open in the previous comment remain open.

PR #4553, opened 2026-09-01
feat(observability): time the external prompt-moderation call

ZacxDev:
## The gap

`extModeration.moderatePrompt` is an outbound HTTP call to a third-party classifier that runs
**inline and serially** on the generation submission path (`generateFromGraph` → `auditPromptServer`
→ here). It is bounded by `AbortSignal.timeout(EXTERNAL_MODERATION_TIMEOUT_MS)` (default 5 s), and it
is **fail-soft**: the caller catches, writes one Axiom line and proceeds with `flagged: false`.

Those last two properties are what made it invisible. Nothing timed the call, nothing counted its
failures, and because a failure is swallowed, a fully-down moderation gateway and a perfectly healthy
one produce the same user-visible behaviour and the same (absent) metrics. The only trace of a
failure was a single log line carrying no duration, which cannot be aggregated into a rate or a
quantile.

So three questions that decide whether this call is a latency problem had no answer at all:

1. how long does a moderation call take, per call;
2. how often does it fail — i.e. how often are prompts generated with only the local regex gate
   applied, the secondary layer silently absent;
3. how often does it reach its abort deadline, as opposed to merely being slow.

## What this adds

Pure observability. **No behaviour, timeout, control-flow or verdict change** — the request, the
deadline, the fail-soft catch and the returned value are all byte-for-byte what they were.

* `civitai_app_external_moderation_duration_seconds{source, outcome}` — wall-clock duration of the
  whole classifier call as the caller experiences it, including body parse.
  * `outcome` ∈ `ok | error | timeout`. The `timeout`/`error` split is the point of the label rather
    than a refinement of it: from the caller's side both are "fail soft, proceed unmoderated", but
    they cost completely different amounts of wall time and they call for opposite fixes (re-size the
    deadline vs. fix or route around the gateway).
  * `source` ∈ `generate | preset | remixAudit | other`, so the request-path prompt gate can be
    divided against the submission procedure's own wall time without a background job blended into
    it. `other` is the default, so a caller that declares nothing can never inflate `generate`.
* `civitai_app_external_moderation_skipped_total{source}` — calls that returned without contacting
  the classifier because the endpoint/token are not configured. Deliberately a separate counter
  rather than a fourth outcome: those calls do no I/O, so folding their ~0 s observations into the
  histogram would drag every quantile toward zero. It exists because otherwise a zero rate on the
  histogram is indistinguishable from a deployment where external moderation is simply switched off
  — the reassuring reading and the alarming one would look identical.
* A `moderation:external-prompt` span carrying the same `source` and `outcome`, so a trace and the
  metric can never disagree about how one call settled.

Both instruments are total (they never throw) and are recorded **inside `moderatePrompt`** — the
lowest funnel, which every caller reaches and none can bypass. Recording a level up, at
`auditPromptServer`, would have blended the regex audit, the benign-phrase lookups and the reporting
work into the same number, none of which is the external call.

## Bucket boundaries

`[0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2, 3, 5, 7.5, 10, 20]`

The low end exists because this is a single outbound HTTPS POST whose healthy population lives in
tens-to-hundreds of milliseconds; the whole reason to time it is to apportion a per-call budget of
that size, so a first boundary at half a second would put the entire healthy population in one bucket
and answer nothing.

The top end is the part that is easy to get wrong, and **`5` is deliberately not the top**. A
boundary is inclusive (`le`) and `AbortSignal.timeout(5000)` fires *at* 5000 ms, so an aborted call is
observed at slightly more than 5 s and lands in `le=7.5`, while a call that answered just under the
deadline lands in `le=5`. Those two are exactly the distinction this metric exists to draw, and they
are only distinguishable because a finite boundary sits *above* the cap. A family whose top finite
bucket sat at the cap would drop every capped call into `+Inf` together with every pathological one,
saturating precisely the region being asked about. `10` and `20` carry two further things: the
deadline is an env knob that can be re-tuned upward, so the tail keeps resolution if it is; and under
the default cap a non-empty `+Inf` is itself a finding (it would mean the abort did not take effect),
which a bucket set stopping at the cap could not express.

All three properties are pinned by test, not only by this description.

## Pattern

Mirrors `orchestrator-submit-metrics.ts` — duration histogram + `outcome` label, a bounded `source`
label with a runtime clamp (not only a compile-time type, since options objects are built by spread),
a companion span, total record helpers, and the surface→source mapping reused from that module rather
than a parallel vocabulary invented here.

## Tests

`src/server/prom/__tests__/external-moderation.metrics.test.ts` and
`src/server/integrations/__tests__/moderation.instrumentation.test.ts`, read back against the **real**
prom-client registry rather than a stub, so the assertions are about what is actually recorded.

Covered: the success path records a duration under the success outcome (value asserted, with a unit
sanity ceiling, not just a count); a thrown error records the error outcome and still records a
non-zero duration; a fired deadline records `timeout` and not `error`; a network-level rejection
records `error` and not `timeout`; the not-configured short-circuit counts separately and never
touches the histogram; the `source` label is closed at runtime; and — asserted on every failure case
— the observation count across **every** series moves by exactly one per call, including when the
caller wraps the call in its own fail-soft catch.

Every assertion above was mutation-checked: the implementation was broken on purpose (drop the
timeout classification; observe twice on the failure path; move the top finite bucket down to the
cap; record milliseconds instead of seconds; fold the short-circuit into the histogram; make the
clamp return its raw input; record a zero duration on failure) and each mutant was confirmed to turn
the corresponding assertion red for its own reason, with the tree restored byte-identical afterwards.

### DIFF
--- a/src/server/integrations/__tests__/moderation.instrumentation.test.ts
+++ b/src/server/integrations/__tests__/moderation.instrumentation.test.ts
@@ -73,6 +73,20 @@ async function histSum(source: string, outcome: string) {
   );
 }
 
+/** Cumulative bucket value at boundary `le` (prom histogram buckets are cumulative). */
+async function histBucket(source: string, outcome: string, le: number) {
+  const vals = await samples(HIST);
+  return (
+    vals.find(
+      (v) =>
+        v.metricName === `${HIST}_bucket` &&
+        v.labels.source === source &&
+        v.labels.outcome === outcome &&
+        Number(v.labels.le) === le
+    )?.value ?? 0
+  );
+}
+
 async function skippedCount(source: string) {
   const vals = await samples(SKIPPED);
   return vals.find((v) => v.labels.source === source)?.value ?? 0;
@@ -252,6 +266,78 @@ describe('moderatePrompt instrumentation — failure paths', () => {
     expect(await histCount('preset', 'ok')).toBe(1);
     expect(await histCount('generate', 'error')).toBe(1);
   });
+});
+
+describe('a FIRED deadline is observed ABOVE the bucket boundary that equals the deadline', () => {
+  /**
+   * 🔴 THE ONE CLAIM THE BUCKET SET IS DESIGNED AROUND, DRIVEN END TO END.
+   *
+   * `external-moderation.metrics.test.ts` pins that the boundaries `5` and `7.5` both exist and
+   * that a 4.9 s and a 5.02 s observation land on opposite sides of `le=5`. But it hand-feeds those
+   * two numbers to `observeExternalModeration`, so it never runs the code that PRODUCES the number
+   * and cannot see a regression that makes a real fired deadline observe at or below 5 s. Every
+   * operator-facing sentence about this family — the help text's "those observations land just above
+   * the cap (le=7.5 at the 5s default), never in le=5" — is a claim about the produced value, not
+   * about the bucket list.
+   *
+   * WHY IT HOLDS TODAY, mechanically: `start` is taken in `moderatePrompt` BEFORE `withSpan`, and
+   * `AbortSignal.timeout(EXTERNAL_MODERATION_TIMEOUT_MS)` is constructed later, inside the callback.
+   * The deadline therefore starts counting strictly AFTER `start`, so elapsed-at-record is strictly
+   * greater than the deadline. Reorder those two — or clamp the recorded duration to the deadline,
+   * or start the timer inside the callback — and a capped call starts landing in `le=5`,
+   * indistinguishable from a gateway that answered just in time. That is the exact distinction the
+   * `timeout` outcome and the 5/7.5 boundary pair exist to draw.
+   *
+   * 🔴 THE WALL TIME IS THE POINT, AND IS PAID DELIBERATELY. `AbortSignal.timeout` is a Node-native
+   * timer that vitest's fake timers do not drive, and running this at a shortened deadline would
+   * test a boundary pair (`le=3`/`le=5`) that no deployment uses — the claim in the help text is
+   * about the 5 s DEFAULT. So this case really does park for ~5 s. It is one test.
+   *
+   * The assertion is written as `le=5 is EMPTY` rather than `le=7.5 is 1` on purpose: a loaded
+   * runner can stretch the observation past 7.5, which would make the tighter form flaky without
+   * making it stronger. "Never counted as sub-cap" is the whole property.
+   */
+  it('a real 5s abort lands strictly above le=5, never in it', async () => {
+    env.EXTERNAL_MODERATION_TIMEOUT_MS = 5000; // the production default, explicitly
+    // A gateway that never answers: only the abort can settle this promise.
+    vi.stubGlobal('fetch', (_url: string, opts: RequestInit) => {
+      return new Promise((_resolve, reject) => {
+        opts.signal?.addEventListener('abort', () =>
+          reject((opts.signal as AbortSignal).reason ?? new Error('aborted'))
+        );
+      });
+    });
+
+    await expect(extModeration.moderatePrompt('a hanging prompt', 'generate')).rejects.toBeTruthy();
+
+    // Preconditions — without these the bucket assertion below could pass vacuously by observing
+    // nothing at all, or by landing on a different series.
+    expect(
+      await histCount('generate', 'timeout'),
+      'the fired deadline must be recorded exactly once, as outcome=timeout'
+    ).toBe(1);
+    expect(await histTotalCount()).toBe(1);
+
+    // 🔴 THE PIN. le is INCLUSIVE, so a duration of exactly 5.000 would count here.
+    expect(
+      await histBucket('generate', 'timeout', 5),
+      'a call cut by the 5s deadline must NOT be counted at or below le=5. If it is, a capped ' +
+        "call and a call that answered just under the cap share a bucket, the help text's " +
+        '"never in le=5" becomes false, and "is the gateway slow or are we cutting it off?" ' +
+        'stops being answerable — check that `start` is still taken BEFORE the AbortSignal is ' +
+        'constructed, and that the recorded duration is the real elapsed time, unclamped.'
+    ).toBe(0);
+    expect(
+      await histSum('generate', 'timeout'),
+      'the recorded duration must be the real elapsed wall time, which exceeds the deadline'
+    ).toBeGreaterThan(5);
+    // …and it must still land in a FINITE bucket rather than being swallowed by +Inf, which is
+    // the other half of why the top boundary sits at 20 and not at the cap.
+    expect(
+      await histBucket('generate', 'timeout', 20),
+      'a capped call must land in a finite bucket, not in +Inf alongside pathological ones'
+    ).toBe(1);
+  }, 20_000);
 });
 
 describe('moderatePrompt instrumentation — not configured', () => {
--- a/src/server/integrations/__tests__/moderation.instrumentation.test.ts
+++ b/src/server/integrations/__tests__/moderation.instrumentation.test.ts
@@ -268,37 +268,48 @@ describe('moderatePrompt instrumentation — failure paths', () => {
   });
 });
 
-describe('a FIRED deadline is observed ABOVE the bucket boundary that equals the deadline', () => {
+describe('a call cut by the REAL abort deadline, at the production default', () => {
+  /** The production default deadline, in seconds (`EXTERNAL_MODERATION_TIMEOUT_MS` = 5000). */
+  const CAP_SECONDS = 5;
+
   /**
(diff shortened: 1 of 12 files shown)


ZacxDev:
### Adversarial audit — round 1

Read tip `13e81c0ef6`. **No deploy-blocking findings.** Two should-fix and three doc-accuracy nits, all now addressed in `92ec4312`.

The one worth calling out for reviewers: the `moderationSource` seam was **unguarded**. Substituting a literal `undefined` for it survived the entire unit suite — 24,475 passed, 0 failed, byte-identical counts — because the existing assertions pinned the *default* and the *arity* but never a *declared* value. In production that mutant makes `source="generate"` and `source="preset"` permanently empty, which is indistinguishable from "no generations happened" — precisely the query the metric's help text tells an operator to run. Typecheck cannot catch it, since `undefined` is assignable to the parameter type.

Separately, the comics cron made two external-moderation calls per panel but declared a source on only one, so the documented `preset` population undercounted it 2× from day one.

Cleared explicitly, so they are not re-derived: return/throw shape and the caller's fail-soft `catch` are unchanged; exactly one observation per issued request, never in a `finally`; label cardinality is bounded and no user-controlled value can reach a label; the new span inherits the request's sampling decision. The two pre-existing tests this PR touched were **tightened**, not loosened.

```audit-claims round=1 audited=13e81c0ef6..92ec4312
1. The comics cron's pre-submit auditPromptServer call now declares moderationSource 'preset', so both of that job's per-panel moderation calls land on preset rather than one falling to other.
2. A new test file pins that a declared moderationSource actually reaches moderatePrompt, across the whole source vocabulary, including a two-call case that no single constant can satisfy.
3. A new AST ledger enumerates all 9 production auditPromptServer call sites and pins the expression each must pass, failing when the set grows, shrinks, or any expression changes.
4. The bucket claim is now driven end to end with a real AbortSignal.timeout against a fetch that only settles on abort, asserting a fired deadline is not counted at or below le=5.
5. The generate source's documentation now states it is every surface except preset, block included, rather than onsite/api only.
6. The clamp's stated rationale was rewritten to the real one; the clamp itself is unchanged.
```

ZacxDev:
### Adversarial audit — round 2

Read tip `92ec4312ef`. No deploy-blocking findings, six should-fix/nit findings, all addressed across `fc4dcfd854` and `8e0c241f11`.

Two are worth a reviewer's attention.

**Round 1's fix re-introduced the defect class it existed to fix.** It added a doc claim that a per-panel rate on the `preset` series is "2× the panel rate, by construction". That was false: `source=preset` is produced by three callers — the comics cron at 2 calls/panel, plus `iterateGenerate` and the shared `submitComicGeneration` helper at 1 each — so the series is `interactive + up-to-2×cron` and cannot be divided by 2. It is also not "by construction" even for the cron, since an empty prompt returns early and a hard block throws before `moderatePrompt`, yielding 0 or 1.

**The end-to-end deadline test was flaky at ~25 %, and CI passed it twice by luck.** Measured 2 failures in 8 isolated, unloaded runs, with wall time identical across pass and fail — so an assertion-boundary problem, not load. Root cause: `AbortSignal.timeout(5000)` can fire fractionally before `performance.now()` measures 5000 ms elapsed (probed: 3/60 samples early at a 100 ms deadline, 3/8 at 5000 ms, worst case 0.63 ms), and `le` is inclusive, so a capped call landed in `le=5` instead of `le=7.5`.

That made the shipped claim wrong, not just the test: the histogram help text told operators a capped call is distinguishable from an under-deadline one by which bucket it lands in. It is not, and it cannot be in general — two calls either side of the cap are arbitrarily close, and the deadline is env-tunable. `outcome="timeout"` is branched off the abort error rather than the duration and separates them deterministically; the help text now says to use it and never the bucket. The test was rewritten to assert that, rather than being loosened or retried. Post-fix: 34 consecutive passes across two independent runners.

```audit-claims round=2 audited=92ec4312ef..8e0c241f
1. The false "2x the panel rate, by construction" claim is replaced with what the series actually holds: it counts calls not submissions, is interactive + up-to-2x cron, and cannot be divided by 2.
2. The AST ledger rows now carry a per-key expected site count, so deleting one of the two auditPromptServer calls inside generateFromGraph fails the shrink test instead of passing green.
3. A non-literal options object at a call site now reports its own distinct message instead of being misreported as "pinned as deliberately UNDECLARED".
4. The test-level 20000ms timeout override was removed, restoring the unit project's deliberate 60s default.
5. The clamp rationale in moderation.ts now matches the one in external-moderation.metrics.ts, so the repo no longer contradicts itself about why the clamp exists.
6. The "other is not split further because it would mean threading a source to nine call sites" cost argument is corrected, since this PR shipped that plumbing.
7. The bucket-separation claim is removed from both the code comments and the operator-facing help text; capped calls are identified by outcome=timeout, which is branched off the abort error rather than the duration.
8. The histogram boundary that sat exactly on the default deadline was moved from 5 to 4.5, so one outcome=timeout population no longer splits across two buckets on a dashboard.
9. The flaky end-to-end deadline test now asserts outcome and recorded duration instead of a bucket edge; it was not loosened, retried, or skipped.
```

ZacxDev:
### Adversarial audit — round 3, and why the ladder stops here

Read tip `8e0c241f11`. **No deploy-blocking findings** (none in any of the three rounds). Three should-fix and two nits, all addressed in `25392fd2c5`.

The one with measured teeth was a regression the *previous* fix round introduced: the rewritten duration floor `>= CAP - 0.05` could not detect a systematic clock bias below 50 ms, **while its message claimed it could**. Demonstrated by mutation — injecting a constant 5 ms under-measurement passed that file 13/13 green, and the same mutant failed against the version before the rewrite. The fix adds a clock-consistency assertion comparing the recorded duration to the test's own `performance.now()` window; the mutant now produces an isolated kill (1 failed / 12 passed) naming the injected bias.

Also fixed: a test constant hardcoded at `5` with nothing tying it to the schema default, so the "no bucket boundary on the deadline" guard only ever forbade `5` — had the default become `3000`, `3` **is** a live boundary and the split-timeout defect would have returned invisibly. It is now derived from the schema, and was proved by temporarily changing the default and watching the guard name `[3]`.

### The sweep, and what it found

Every round of this ladder found the same defect shape: **a comment or `help` string asserting behaviour the code does not have** — and each round's rewrite staled another file's prose, generating the next round's findings. So the final round swept *every* prose claim this PR adds or touches, rather than fixing the reported sites point-wise. That found **four more instances** beyond the five reported, including:

- The module's own newly-written "a blocked panel contributes 0 or 1, never 2" was false — an external flag is raised *after* the classifier answers, so a panel blocked at the second gate contributes 2. True only for a regex block.
- `source=generate` is fed by **both** audits inside `generateFromGraph` (the prompt gate and the ACE Audio creative-field audit), so it counts calls, not submissions. The histogram `help` string — which ships to operators — said "prompt gate".

Unverifiable figures were softened rather than restated, and two historical measurements are attributed to the round that measured them rather than asserted as fact.

### 🔴 Why this stops here rather than running to a clean round

Stated explicitly so a later reader can tell "converged" from "real items knowingly left open" — those look identical otherwise.

Three rounds produced **zero** deploy-blocking findings. Rounds 2 and 3 each changed **exactly one executable line** of shipped code (`moderationSource: 'preset',` and a bucket boundary `5` → `4.5`); everything else was comments and the help string. The payload is now effectively prose, so each round rewrites prose and the next round audits the prose it just wrote — the rounds regenerate their own findings and will not terminate on their own. The precondition for stopping was to sweep the recurring **shape** at every site rather than at the sites reported, which the final round did.

**Deliberately left open — not absent, and not fixed:**

1. `generation-surface-wiring.test.ts:438-445` has the identical `Set`-membership shrink design and therefore the identical within-key blind spot: a call deleted inside an already-pinned key passes green. The per-key site-count remedy used here ports over directly. Inherited, not introduced by this PR.
2. `moderatePrompt`'s declared return type is `Promise<{ flagged: false; … }>` — a literal `false` for a genuinely boolean value, type-checking only because `results[0].flagged` is `any`. It currently tells every consumer a prompt is never flagged. Pre-existing, and a type/behaviour change rather than observability.
3. The new clock-consistency bound is 1 ms and **is load-exposed** — a jammed event loop grows the residue. It carries 5.7× headroom over the worst of 12 measurements and passed 22/22 across two independent runners. If it ever fires spuriously, widen it; do not delete it.
4. `source="other"` remains a genuine mixture (App Blocks host-side audits, prompt enhancement, shared content-safety). Those six call sites are pinned as deliberately undeclared in the call-site ledger with the reasoning recorded, so the choice is reviewable rather than defaulted into.

Verification carried out independently of the agents that did the work: the deadline test re-run 10× at the final head (10/10), the 5 ms-bias mutant re-applied and watched killing the new guard for its own reason, and the working tree confirmed byte-identical afterwards.

PR #4555, opened 2026-09-01
chore(bitdex): drop the 5 BitDex tables (decommission phase 2 — functions and trigger already applied)

JustMaier:
> ⚠️ **PARTIALLY APPLIED ALREADY — 2026-09-02 ~01:15Z.** Dropping the disclosed `bitdex` Postgres role required clearing what it owned, so **`trg_cleanup_bitdex_ops` and all 27 functions are already gone from prod**, along with the role itself (`pg_roles` 43 → 42). Applied on `cnpg-database/cnpg-cluster-nvme0-5` via `DROP OWNED BY bitdex` + `DROP ROLE bitdex`, verified from both that pod and the bastion.
>
> **What is left for this migration: the 5 tables only** — `"BitdexOps"` (124,020 rows), `bitdex_cursors`, `bitdex_fanout_capture`, `bitdex_post_publish_capture`, `bitdex_zm_log`. Every statement here is `IF EXISTS`, so re-applying the whole file is safe: the trigger and function drops become no-ops.
>
> The precondition below about `reemit-bitdex-ops.ts` was **knowingly overridden** for the function drops, because closing a disclosed credential outranked it and both jobs are flag-off (`enabled: false`, zero runs since 18:50Z). If that flag is flipped on before #4568 deploys, the job throws — loudly, in a job, not on a user path.

Phase 2 of 2 of the BitDex decommission. **One migration file, no code changes.** Removes what phase 1 deliberately left behind.

🔴 **NOT SAFE TO APPLY YET.** Merging is fine — migrations here never auto-run — but the SQL must not be applied until three things are true:

1. ~~**The ingest engine is gone from the cluster.**~~ ✅ **DONE** — talos-infra 1373 and 1374 both merged; namespace gone 19:16:03Z, prune verified, `kustomization/bitdex` NotFound.
2. **`src/server/jobs/reemit-bitdex-ops.ts` is removed.** It is the only remaining caller of `bitdex_post_fanout_ops` / `bitdex_image_sortat_ops` — verified by `git grep` over `origin/main` across `src`, `apps`, `packages`, `scripts`, excluding tests. Both are flag-gated off so the call is unreachable, but the file still names them. ⬅ **THIS IS THE REMAINING BLOCKER.** In progress on branch `chore/remove-bitdex-app-code` (WIP, does not typecheck yet); see `claudedocs/bitdex-app-removal-handoff-2026-09-01.md` on that branch.
3. **The `bitdex` Postgres role is dropped AFTER this, never before.** A role drop with dependent objects fails noisily.

## What it drops

| | count |
|---|---|
| `trg_cleanup_bitdex_ops` on `bitdex_cursors` | 1 |
| `bitdex_*` functions | 27 |
| tables — `"BitdexOps"`, `bitdex_cursors`, `bitdex_fanout_capture`, `bitdex_post_publish_capture`, `bitdex_zm_log` | 5 |

Only `cleanup_bitdex_ops` still has a trigger reference. The other 26 already have zero — several are orphans from earlier trigger generations, which is what the hash suffixes are.

## `"BitdexOps"` is dropped NON-EMPTY, and that is deliberate

124,020 rows at drop time. **This is not unfinished work**, and the reason is worth reading before anyone "fixes" it:

`cleanup_bitdex_ops` deletes ~5,000 consumed rows per firing and fires **only** on a write to `bitdex_cursors`. The only writer is the engine advancing its cursor. Once phase 1 stopped the producers, the engine caught up, stopped writing its cursor, and the reaper stopped firing with ~25 chunks still to go. The backlog can never reach zero on its own — waiting is not a strategy, it is a condition that can no longer occur.

Verified fully consumed rather than assumed: `max("BitdexOps".id)` = **264638333** = `min(last_outbox_id)` across **both** cursors, exactly. Three independent samples (two replica, one primary) agree byte-for-byte on both numbers. Nothing is lost by dropping the table.

## Verification

Every name was **generated from the live prod catalog** (`pg_proc` + `pg_get_function_identity_arguments`, `pg_class`) rather than hand-typed, then checked back against it:

- **27 of 27** function names resolve, **0 missing**
- Control: the same query with two deliberately bogus names appended reports `missing=2`. The check can fail.

That control matters because every statement is `IF EXISTS`, which turns a typo into a silent success — the same reason phase 1 needed a before-count.

Per-statement `lock_timeout` and one-statement-at-a-time, for the same reasons as phase 1 and stated in the file: `DROP` takes ACCESS EXCLUSIVE, and a multi-statement run through the postgres-query skill prints a formatter error and applies anyway, so its exit status cannot distinguish a clean run from a half-applied set.

**Not run:** typecheck / unit suite. Same reason as phase 1 — the diff is a single `.sql` file that no TypeScript and no test imports, and this worktree has no `node_modules`. `drift-baseline.json` contains no `bitdex` entries.

## Sequence

1. flipt-state#80 — jobs off ✅ merged, synced, verified
2. civitai#4552 — 8 write triggers ✅ applied to prod ~18:56Z, **merged 20:18:07Z (`c98dee3d13`)**
3. writers stopped ✅ table static, engine caught up
4. talos-infra#1373 — Kyverno unprotect ✅ merged (`30399d8599`)
5. talos-infra#1374 — teardown ✅ merged (`ee9598a4c3`), namespace gone 19:16:03Z
6. app-side code removal ⬅ **THE ONLY REMAINING BLOCKER**
7. **this PR** applied
8. drop the `bitdex` role

https://claude.ai/code/session_01KYrq2ighNp2rABTAvpzhj9

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260901200000_drop_bitdex_objects_phase2/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260901200000_drop_bitdex_objects_phase2/migration.sql
@@ -1,0 +1,95 @@
+-- BitDex decommission, phase 2 of 2: remove the objects themselves.
+--
+-- 🔴 DO NOT APPLY THIS BEFORE ALL THREE OF THESE ARE TRUE:
+--   1. the BitDex ingest engine is gone from the cluster (talos-infra PRs 1373 -> 1374),
+--      so nothing is writing `bitdex_cursors` and nothing is reading "BitdexOps";
+--   2. the app code calling `bitdex_post_fanout_ops` / `bitdex_image_sortat_ops` is
+--      removed -- as of writing that is `src/server/jobs/reemit-bitdex-ops.ts` and
+--      nothing else (both are flag-gated off, so the call is unreachable, but the file
+--      still names the functions);
+--   3. the `bitdex` Postgres login role has NOT yet been dropped. Drop it AFTER this
+--      migration, never before -- a role drop with dependent objects fails noisily.
+--
+-- Phase 1 (20260901190000_drop_bitdex_write_triggers) already removed the 8 write
+-- triggers, applied to prod 2026-09-01 ~18:56Z.
+--
+-- "BitdexOps" is deliberately NON-EMPTY at drop time -- 124,020 rows as of 2026-09-01.
+-- That is not unfinished work. Its reaper, `cleanup_bitdex_ops`, fires only on a write
+-- to `bitdex_cursors`, and the only writer is the engine advancing its cursor; once the
+-- writers stopped the engine caught up and stopped writing, so the backlog can never
+-- reach zero on its own. Verified consumed: max("BitdexOps".id) = 264638333 equals
+-- min(last_outbox_id) across both cursors exactly. Nothing is lost by dropping it.
+--
+-- Same per-statement `lock_timeout` reasoning as phase 1, and the same instruction:
+-- apply ONE STATEMENT AT A TIME and read each result. A multi-statement run through the
+-- postgres-query skill prints a formatter error and applies anyway, so its exit status
+-- cannot tell you whether the set is half-applied.
+--
+-- Every name below was generated from the live prod catalog (pg_proc /
+-- pg_get_function_identity_arguments, pg_class), not hand-typed.
+
+-- The drain trigger. Last BitDex trigger in the database; phase 1 kept it so the
+-- backlog could drain as far as it could.
+DO $$
+BEGIN
+  SET LOCAL lock_timeout = '5s';
+  DROP TRIGGER IF EXISTS trg_cleanup_bitdex_ops ON bitdex_cursors;
+END $$;
+
+-- 27 functions. All but `cleanup_bitdex_ops` already have zero trigger references;
+-- several are orphans from earlier trigger generations (the hash suffixes).
+DROP FUNCTION IF EXISTS public.bitdex_collection_notify();
+DROP FUNCTION IF EXISTS public.bitdex_image_notify();
+DROP FUNCTION IF EXISTS public.bitdex_image_ops_085c2a3b();
+DROP FUNCTION IF EXISTS public.bitdex_image_ops_45edf6c8();
+DROP FUNCTION IF EXISTS public.bitdex_image_ops_8905b723();
+DROP FUNCTION IF EXISTS public.bitdex_image_ops_a6ea374e();
+DROP FUNCTION IF EXISTS public.bitdex_image_ops_dda29cf8();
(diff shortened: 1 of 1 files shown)


JustMaier:
**The SQL in this migration has been applied to production**, so merging it records history that already happened rather than scheduling work. Applied 2026-09-02 by the decommission seat, in two parts:

- The 27 functions and `trg_cleanup_bitdex_ops` went at ~01:15Z as a prerequisite for `DROP OWNED BY bitdex` (the role could not be dropped while it owned them).
- The five tables were dropped at ~15:47Z, once #4568 had **deployed** — not merely merged. That ordering was load-bearing: applying it while a pod still ran code calling `bitdex_post_fanout_ops` / `bitdex_image_sortat_ops` would have thrown.

Every statement is `IF EXISTS`, so the file is idempotent against the partial application above.

**Verified after the fact, on the primary via the pod path:**

```
BitdexOps                   -> ABSENT     Image -> "Image"   <- control
bitdex_cursors              -> ABSENT     Post  -> "Post"    <- control
bitdex_fanout_capture       -> ABSENT
bitdex_post_publish_capture -> ABSENT
bitdex_zm_log               -> ABSENT
```

Both controls sat in the **same** `to_regclass` query, so a malformed statement or a wrong database would have returned seven ABSENTs rather than five.

Then a whole-schema sweep wider than the five named objects — `ILIKE '%bitdex%'` across `pg_class`, `pg_proc`, `pg_trigger` and `pg_roles` — returned **0 tables, 0 functions, 0 triggers, 0 roles**, against **362 public tables** as a control. That is what makes the claim "nothing bitdex-shaped remains" rather than "the five I knew about are gone".

Each `DROP TABLE` ran with `SET LOCAL lock_timeout = '5s'` in its own invocation. That guard was proven before use rather than assumed: against a scratch table held under `ACCESS EXCLUSIVE` by a second session it failed in ~6s with `canceling statement due to lock timeout` naming the inline code block, and with no lock it succeeded in 1s. `pg_stat_activity` showed zero sessions touching the five tables immediately beforehand, so nothing contended and the timeout never fired in anger.

**Still outstanding, and not part of this PR:** ~400 GiB of retained data under `/var/mnt/nvme1/bitdex/` on `talos-wjh-tgy` and `talos-48r-b3a`. Both PVs are `reclaimPolicy: Retain`, so deleting the PV objects would free nothing — the data must go first, and `spec.local.path` on those PVs is the only record of where it is. Tracked as an infra handoff.

🪤 For anyone who lands here later: after this decommission the only `bitdex`-named path left in the cluster is `/var/mnt/bitdex-nvme/buzz-db/`, and **it belongs to buzz-db** — 1.5 TiB x2, live. The paths differ in token *position*, not spelling, so `grep bitdex` and `find -name '*bitdex*'` match both. Discriminate on phase, claimRef namespace and UID; never on the string.

PR #4561, opened 2026-09-01
feat(auto-feature): detect the job going silent from outside a run

JustMaier:
Closes the observability half of the featured-collections work: `868kz3ef5`.

## The gap

`auto-feature-images` logs a `job-summary` on every run, including a zero-pick one — added precisely because a stricter cap means the job can find nothing to pick. But that fires only from **inside** a run. During the 79-hour silence ending 17 Aug 2026, the only interruption in the job's history, it would have emitted nothing at all, because nothing executed.

A zero-pick run and a job that never ran produce identical evidence: silence.

## The shape

`challenge-health-check.ts` is the precedent — a separate job that queries output state rather than instrumenting the producer. Two signals, both plain table reads answerable with the producer dead:

- **`lastRow`** — newest row the job wrote into its target collection. Its output.
- **`lastRun`** — the `KeyValue` row `getJobDate` advances. Its heartbeat.

| State | Meaning | Action |
|---|---|---|
| heartbeat stale | not executing, or bailing before it scores | **pages Discord** |
| heartbeat fresh, output stale | runs and picks nothing — legitimate when the caps refuse everything | records to Axiom |

The split is Justin's call, not the author's. Threshold derives from the block's own `intervalHours`, so it tracks a cadence change without a deploy — **13h** at today's config, against a measured 6-7h real spacing.

## 🔴 How to verify this — NOT by turning off the flag

The ticket's closing condition says "suppress the job and observe the alert". The obvious suppression is turning off `AUTO_FEATURE_IMAGES`, and **that produces no alert by design**: the flag gate returns before anything is evaluated, and the Axiom "Skipped" line it leaves reads like a pass. Found by the intent lane; it would have left this ticket falsely closed.

What works:

1. note the current value of `KeyValue['job:auto-feature-images']`
2. set it back a day
3. `GET /api/webhooks/run-jobs/auto-feature-health-check?token=$WEBHOOK_TOKEN` — the endpoint selects one job by name, so this runs the check alone and immediately
4. observe the Discord page, then restore the value

**This has not been done yet.** The 26 tests and 25 mutation controls below establish that the logic is right and that a revert fails legibly; they are **not** an observation of the alert firing. That step is still owed after deploy, and the ticket does not close without it.

## Verified against production, and what that does not prove

Read-only replica. The job is currently alive: newest auto-featured row `2026-09-01 18:20:10 UTC`, heartbeat `18:20:09.901 UTC` — two independent signals agreeing to the second.

The clean reading is not a blind one: the same gap query over full history returns exactly one row, `2026-08-13 17:00 → 2026-08-17 00:00, 79.0 hours`.

⚠️ **That proves the query prints the outage. It does not prove the job would have run to ask it.** Different claims, and only the first is evidenced here.

`EXPLAIN ANALYZE` (perf lane, prod replica): warm 3.5 ms / 257 buffers, cold 6.6 ms. `CollectionItem` is 15 GB / 180M rows; this touches 257 buffers 8×/day. A `>= now() - interval` bound was measured and **rejected** — it lands as a Filter after the heap fetch, identical buffers and blocks, buying nothing while costing a correctness ambiguity.

## Review

Reviewed in four rounds across seven commits, not one. Five lanes at `1b09897909`, reuse + correctness at `247333d4db`, correctness at `b955dc476f`, five lanes again at `a4d9f9cbc5`, tests re-read at the head. Fixes are spread across commits three through seven.

🔴 **The most serious defect was found in the final round and would have surfaced on its own.** The suite read the real clock while pinning a literal `NOW`, so **5 of 25 tests were due to start failing at 2026-09-02T00:00:00Z** — hours after the last commit, in tests that look like they cover the code they would appear to indict. Fixed in `a4d9f9cbc5` by faking `Date` alone. Demonstrated both ways: shifting the pinned `NOW` two days fails 5 without the fix and passes 25 with it. Nothing in typecheck, lint or the suite could see it, because it passes right up until it does not.

- **Reuse** — `alertDiscord` was byte-identical across both health checks → `notifyModAlert`; the block-config lookup had four copies, two already disagreeing on `orderBy` → shared in the leaf module; the interval default is derived from the schema rather than restated.
- **Correctness** — a Discord failure was swallowed while the job returned `paged: 1`, so a revoked webhook would leave the outage detector claiming it alerted someone forever. `paged` now reports what landed. An unresolvable attribution account produced a null `lastRow` that read as "wrote nothing" and blamed the caps.
- **Tests** — nothing pinned *which* Flipt flag gates the check, nor the Discord payload; an empty embed or a POST to the wrong host passed a call-count assertion identically.
- **Perf** — no findings, and it refuted a mitigation the author had proposed.
- **Intent** — delivers the ticket, no scope creep; F1 above is its finding.

The round also produced a bug of its own, predicted by the test lane: `Number(['1'])` is `1`, and `KeyValue` really does hold arrays for other keys, so a one-element array slipped the heartbeat guard and became a 1970 date — while the comment above it claimed unparseable values became null.

## Tests

26 tests. 25 mutation controls. All but one fail with an assertion naming the wrong value rather than timing out — the exception is row 15 below, which is a *paired* control where the passing arm is the evidence:

| Mutation | Failure |
|---|---|
| threshold hardcoded | `expected 13 to be 49` |
| null heartbeat treated as inconclusive | `{ healthy: true }` vs `{ healthy: false, paged: 1 }` |
| drop the heartbeat-healthy guard | `alerts: 2` vs `alerts: 1` |
| `max(createdAt)` for `COALESCE` | SQL does not match `/COALESCE/` |
| dry pipeline pages instead of records | `paged: 1` vs `paged: 0` |
| drop the ACCEPTED filter | SQL does not match `/status = 'ACCEPTED'/` |
| producer's key diverges from consumer's | `expected undefined to be 'AUTO_FEATURE_JOB_DATE_KEY'` |
| drop the heartbeat `typeof` guard | `expected 2001-01-01T07:00:00.000Z to be null` |
| query binds swapped | `[9001, 9001, …]` vs `[107, 9001, …]` |
| gated on the wrong Flipt flag | `expected "vi.fn()" to be called with [ 'auto-feature-images' ]` |
| blank Discord description | `expected '' to contain 'No completed run'` |
| `paged` ignores delivery | `paged: 1` vs `paged: 0` |
| drop `rowsReadable` | `{ healthy: false }` vs `{ healthy: true }` |
| delete the webhook guard | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |
| schema default moves, code **derives** | passes — threshold follows |
| schema default moves, code **restates** | fails — the drift is caught |

The last two are a pair on purpose: substituting `6` for the derived default changes nothing *because the default is 6*, so the single-sided mutation was vacuous and had to be replaced.

## The last round, because it is the most instructive part

The terminating pass found that an assertion I had just written to fix vacuous assertions **was itself vacuous** — and it was invisible to reading.

The guard `not.toMatch(/OR/i)` reached the file through a shell heredoc that ate the escape, so the bytes were `/‹0x08›OR‹0x08›/i`: literal backspace characters where the word boundaries belong, matching nothing and passing forever. It renders as `` in every terminal. Typecheck, lint, prettier and a green suite all agreed it was fine. **The only thing in the chain that could see it was running the mutant rather than naming one** — the guard is now `new RegExp(String.raw`OR`, 'i')`, with `od -c` confirming a real backslash.

Two tautologies went with it. The flag assertion compared against the string this file's own `vi.mock` supplies, so renaming the real `FLIPT_FEATURE_FLAGS` value would have left it green while production went double-silent — producer stops on a flag Flipt does not know, health check reports `skipped`, nothing prints. And the producer-gate regex anchored on the closing paren, which the `getJobDate` regex 170 lines above documents as the thing to avoid.

The control set is four mutants where **three must fail and one must not**: giving the producer an `entityId` is behaviour-preserving and passes 26/26, which is what shows the fragility fix did not simply make the assertion stricter.

## Three limits, documented in the header rather than fixed

- **The threshold's margin collapses at the bottom of the config range.** Real spacing is `intervalHours + 1` against a threshold of `2 * intervalHours + 1`, so the slack is `intervalHours` — six missed wakes at today's 6, one at the schema minimum. At the maximum of 168 the threshold is 14 days, which would not have caught the August outage.
- **This shares two control planes with the job it watches.** `isFlipt` returns false for an unreachable Flipt. More likely, it rides the same scheduler and `run-jobs` endpoint. For August specifically that was not the cause — other crons logged continuously through all 79 hours — but nothing here catches a scheduler-wide stop. A dead-man's switch outside the cluster is infra work, not this PR.
- **`job_duration_seconds_count` already detects half of this** and would have caught the August silence with no code. It counts the `flag-off` and `interval-not-elapsed` early returns, so it cannot see a job that runs but never scores, nor one that scores but writes nothing — and a Prometheus rule for the half it does cover belongs in the infra repo rather than here — that route is open, not closed, as `clickhouse-refresh-monitor.ts` records for its own rules.

## Why `clickhouse-refresh-monitor.ts` is in a featured-collections PR

**One comment line**, reworded to drop a private repo name and a PR number while keeping the alert expressions. It is here because this PR's own header comment cited that file as evidence a Prometheus rule "cannot ship" — a claim that file disproves — so correcting the false claim while leaving the sentence it points at would have been odd. Approved separately by the repo owner.

🔴 **It is one instance of a wider class, not a sweep.** A repo-wide search finds **78 files** matching `talos|infra#|civitai-dp-|gitops` across `src apps packages scripts` (45 in `src/` alone), tracked on **`868kzkwy1`**. Do not read this line as the class being handled.

**How that number was reached, because the method matters more than the count.** My first enumeration said *three* — it came from `grep -rn 'talos-infra' src/`, which searches one directory for one repo name. A second pass gave eleven, a review lane gave ~40, and the coordinator's re-run gave 78 with a nonsense-token control returning 0 to show the pattern discriminates. The pattern is still a proxy somebody chose rather than a definition of the class: it would not find a different private repo, a bastion hostname, a kubectl context, or a namespace. `868kzkwy1` carries that caveat and its closing condition requires stating whether the search was widened.

## What has NOT been reviewed, and why

Two commits land after the last full review round, deliberately:

- **`c18f0c9f2b`** — comment-only. Verified mechanically: a filter for non-comment changed lines returns **zero** over this commit and **21** over the test commit, so the empty result is a check that could have failed rather than one that cannot fire.
- **`7aa79d01e1`** — test-only, one file.

**The tests lane re-read the head. Reuse, perf, correctness and intent did not.** That is a deliberate stopping point rather than an oversight: both commits touch only the tests lane's subject, and acting on a lane's findings always manufactures a new unreviewed head — so "the final SHA was reviewed by everyone" is a state only reachable by never fixing anything. What is true instead: the five lanes read `a4d9f9cbc5`, and everything that changed after it is listed above.

## Deferred, with tickets

- **`868kzm1ec`** — an unconfigured `DISCORD_WEBHOOK_MOD_ALERTS` logs at `info` in production too, where a missing webhook means a paging alert reached nobody. Raised by the correctness lane at `b955dc476f`; a behaviour change, so not bolted onto the last commit.
- **`868kzkwy1`** — the private-repo reference sweep above.

## Still owed before `868kz3ef5` closes

The live verification: set `KeyValue['job:auto-feature-images']` back a day, hit `/api/webhooks/run-jobs/auto-feature-health-check?token=…`, watch Discord, restore. **Not** by turning off the flag — that produces no alert by design.

https://claude.ai/code/session_01529s2vx4rHVEmEXUkMo4wV

### DIFF
--- a/src/pages/api/webhooks/run-jobs/[[...run]].ts
+++ b/src/pages/api/webhooks/run-jobs/[[...run]].ts
@@ -89,6 +89,7 @@ import { pushDiscordMetadata } from '~/server/jobs/push-discord-metadata';
 import { refreshAuctionCache } from '~/server/jobs/refresh-auction-cache';
 import { refreshFeaturedCollectionsEligibility } from '~/server/jobs/refresh-featured-collections-eligibility';
 import { autoFeatureImages } from '~/server/jobs/auto-feature-images';
+import { autoFeatureHealthCheckJob } from '~/server/jobs/auto-feature-health-check';
 import { reemitBitdexOps } from '~/server/jobs/reemit-bitdex-ops';
 import { removeOldDrafts } from '~/server/jobs/remove-old-drafts';
 import { reindexRecentScheduledImages } from '~/server/jobs/reindex-recent-scheduled-images';
@@ -159,6 +160,7 @@ export const jobs: Job[] = [
   updateCollectionItemRandomId,
   refreshFeaturedCollectionsEligibility,
   autoFeatureImages,
+  autoFeatureHealthCheckJob,
   ...metricJobs,
   ...searchIndexJobs,
   searchIndexUserCleanupJob,
--- a/src/server/common/auto-feature.ts
+++ b/src/server/common/auto-feature.ts
@@ -31,10 +31,16 @@ export const AUTO_FEATURE_JOB_DATE_KEY = 'job:auto-feature-images';
  * The default `intervalHours` re-exported from the schema rather than restated, so a health check
  * reasoning about the cadence of a job whose config has gone missing measures against the value
  * that job would actually have used.
+ *
+ * 🔴 Read per-field, not by parsing a synthetic object. This runs at module load and
+ * `collection.service.ts` imports this file, so a throw here fails that import rather than the job
+ * — and parsing the whole schema to read one default couples this line to all fifteen of its
+ * fields, on a schema that exists to be tuned. One added required field or `.refine()` would be
+ * enough. `shape.<field>.parse(undefined)` resolves a `ZodDefault` and cannot throw; the idiom is
+ * `env.mock.ts`'s. Compare `job.ts`, which wraps its module-scope metrics call for the same reason.
  */
-export const AUTO_FEATURE_DEFAULT_INTERVAL_HOURS = autoFeatureSchema.parse({
-  collectionId: 1,
-}).intervalHours;
+export const AUTO_FEATURE_DEFAULT_INTERVAL_HOURS =
+  autoFeatureSchema.shape.intervalHours.parse(undefined);
 
 /**
  * Resolved by username so the same code identifies the right account in dev, preview and prod
--- a/src/server/common/auto-feature.ts
+++ b/src/server/common/auto-feature.ts
@@ -34,10 +34,11 @@ export const AUTO_FEATURE_JOB_DATE_KEY = 'job:auto-feature-images';
  *
  * 🔴 Read per-field, not by parsing a synthetic object. This runs at module load and
  * `collection.service.ts` imports this file, so a throw here fails that import rather than the job
- * — and parsing the whole schema to read one default couples this line to all fifteen of its
- * fields, on a schema that exists to be tuned. One added required field or `.refine()` would be
- * enough. `shape.<field>.parse(undefined)` resolves a `ZodDefault` and cannot throw; the idiom is
- * `env.mock.ts`'s. Compare `job.ts`, which wraps its module-scope metrics call for the same reason.
+ * — and parsing the whole schema to read one default couples this line to every one of its
+ * thirteen fields, on a schema that exists to be tuned. One added required field or `.refine()`
+ * would be enough. `shape.<field>.parse(undefined)` resolves this field's own `ZodDefault` and so
+ * cannot throw while it keeps a `.default()` — removing that reinstates the same hazard, narrowed
+ * from thirteen triggers to one. The idiom is `env.mock.ts`'s. Compare `job.ts`, which wraps its module-scope metrics call for the same reason.
  */
 export const AUTO_FEATURE_DEFAULT_INTERVAL_HOURS =
   autoFeatureSchema.shape.intervalHours.parse(undefined);
--- a/src/server/common/auto-feature.ts
+++ b/src/server/common/auto-feature.ts
@@ -6,6 +6,13 @@ import { dbRead } from '~/server/db/client';
  * Their own module because both the job and the two collection-item removal paths need them,
  * and importing the service from collection.service.ts would pull the whole home-block graph in.
  */
+/**
+ * The `getJobDate` key `auto-feature-images` advances on. Shared so the health check watching that
+ * timestamp cannot drift from the job that writes it — a mismatch would read as a permanently
+ * silent job and page forever.
+ */
+export const AUTO_FEATURE_JOB_DATE_KEY = 'job:auto-feature-images';
+
 export const AUTO_FEATURE_USERNAME = 'CivitaiOfficial';
 export const AUTO_FEATURE_NOTE_PREFIX = 'auto-featured';
 export const autoFeatureNote = (sourceCollectionId: number) =>
--- a/src/server/common/auto-feature.ts
+++ b/src/server/common/auto-feature.ts
@@ -10,6 +10,12 @@ import { dbRead } from '~/server/db/client';
  * The `getJobDate` key `auto-feature-images` advances on. Shared so the health check watching that
  * timestamp cannot drift from the job that writes it — a mismatch would read as a permanently
  * silent job and page forever.
+ *
+ * 🔴 Do not normalise away the `job:` prefix. Every other `getJobDate` caller in the repo uses a
+ * bare name, so this one looks like the odd one out and tidying it is the obvious edit — but the
+ * key names a live row in the production `KeyValue` table. Renaming it orphans that row, and the
+ * job then reads epoch 0, believes it has never run, and fires once immediately.
+ * `auto-feature-health-check.test.ts` pins both sides against exactly this.
  */
 export const AUTO_FEATURE_JOB_DATE_KEY = 'job:auto-feature-images';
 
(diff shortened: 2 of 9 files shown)


JustMaier:
## Review coverage, stated rather than inferred

All five lanes (reuse, correctness, perf, tests, intent) read **`1b098979091efbe82e011deab35e665bcd3c4457`**.

The head is now **`247333d4db217c7058863552b9b377c99ccda5da`**. That commit is the *fixes for what those lanes found*, so **no lane has reviewed it.** A review is pinned to a commit, not to a PR, and the difference matters most in exactly this direction — a fix round is where defects get introduced, not removed. Two of tonight's four real defects were introduced by a previous fix round.

What that unreviewed commit contains:
- `notifyModAlert` extracted; `challenge-health-check.ts` rewired to it (behaviour unchanged, timeout included)
- `getAutoFeatureBlockConfig` moved to the leaf module; producer and watcher now share it
- `paged` reports delivery rather than intent; `rowsReadable` separates "could not look" from "wrote nothing"
- heartbeat `typeof` guard, +5 tests, +5 mutation controls
- header documents the verification trap, the scheduler coupling, and the prom overlap

## Verification of the pushed SHA

Run against a **healthy** test daemon — the previous three full-suite results were discarded, none for a reason involving this code (one superseded, one ambiguous, one killed by a degraded daemon that reported exit 0 while the log said exit 2).

```
Test Files  5 failed | 1554 passed | 3 skipped (1562)     5+1554+3    = 1562
Tests      11 failed | 24668 passed | 35 skipped (24714)  11+24668+35 = 24714
```

Both totals account for every file and every test — no workers died. The 11 failures are the Windows local baseline, established by stashing these files and re-running the same 5 files at the same base: same 5 files, same 11 failures, with and without this change.

- `pnpm run lint` — 4422 problems (358 errors, 4064 warnings), **exact** baseline. It briefly read 4424 from two dead imports left behind by the refactor; caught by diffing against a measured baseline rather than by the error count, which never moved.
- `pnpm run test:lint-rules` — `Test Files 27 passed (27)`
- covering suites incl. `challenge-health-check` — `Test Files 5 passed (5)` / `Tests 62 passed (62)`
- `blocks.router.workflow.test.ts` — 370 tests, nonzero (submodule present)

## Still owed

The live verification. 24 tests and 16 mutation controls show the logic is right and that a revert fails legibly; they are **not** an observation of the alert firing. The recipe is in the PR body and the file header — and note that the obvious one, turning off the flag, produces no alert by design.

PR #4564, opened 2026-09-01
fix(moderation): stop telling every unpublished author they violated the ToS

manuelurenah:
Closes ClickUp [868kz6maf](https://app.clickup.com/t/8459928/868kz6maf).

Unpublishing an article or a model told the author they had committed a **Terms of Service violation** — for every reason, including reasons that are not violations. A creator whose article was pulled for a thin description read that as an accusation and escalated it to support.

## What changes

**The banner matches the reason.** Every unpublish reason is now tagged `policy` or `quality`. Policy reasons still name the violation; quality ones — insufficient description, duplicate — say what they actually are, in yellow rather than red.

**Articles get their own reason list.** The moderator's picker offered all 17 model reasons, including ones that cannot be true of an article: "Missing files", "Missing version", "Images are not generated by the resource". Articles now get the 11 that can, ordered so a misclick between neighbours cannot swap "your write-up is thin" for an accusation. The server rejects the rest, so a stale client cannot post one either.

**The copy is written for articles.** No article message mentions a model, resource, version or example image. Articles already unpublished under an older reason still render a sensible message rather than going blank.

**The alerts got a pass for legibility.** They were a single undifferentiated paragraph with a floating icon and text running the full container width. They now read as title → reason → next steps at a sane measure, using Mantine's standard alert layout.

## Who this affects today

**12,781 models** and **7 articles** are currently unpublished under a reason that is not a violation, and are being told it is.

## Deliberately not in scope

- **The "Request a Review" button stays where it is**, in the download section. Moving it into the alert is a product decision, not a refactor.
- **The moderator's private note still surfaces only for reason `Other`.** For every other reason it is written assuming it stays internal.
- **`unintenteded-use` is in use on 6 articles**, which suggests moderators want something for "this isn't a real article". Dropping it leaves no equivalent — that needs a copy decision from whoever owns moderation policy.

## Verification

Typecheck, lint and format clean. Full unit suite **24,832 passed / 0 failed**. 9 browser tests across the two alert components, each checked against a revert of the code it guards.

Also driven in a browser as a moderator against real unpublished rows in dev, covering both banners in both themes. That is what caught a bug the unit tests missed — a legacy article being headed with the *model* list's label.

https://claude.ai/code/session_016SZwXgwUF7aMJuGSCy7tXK


manuelurenah:
Thanks — three of the four applied in feaeb202c0, one declined with reasons.

**Applied**

1. **Moderator note leaking for a non-`other` reason** — correct, and it was a regression I introduced. The `||` fallback meant any key with no map entry fell through to `Removal reason: <customMessage>`. The code this replaced branched on the key and rendered nothing in that case. Now gated on `reasonKey === 'other'`.

2. **`in` matches inherited keys** — correct, and the consequence is worse than the report suggests. `in` returning true is the mild half; `map[reason]` returns `Object.prototype.toString`, an actual function, which `??` does not reject — so the fallback in `getArticleUnpublishReason` never fired. Same hazard was already fixed in `analytics-bucket-labels.ts` and `AppActivityPanel.tsx`, whose comment describes this exact failure. Switched to `Object.hasOwn` in one shared `lookupReason` so the model path is covered too.

3. **`policy` overstated as strictly ToS** — fair. `mature-real-person` says "prohibited under our content guidelines", not ToS. Comment reworded. The user-facing article heading is unchanged; that noun is existing copy and a moderation-policy decision rather than a code one.

Both fixes have tests, each checked against a revert of the code it guards: the prototype one fails in 2ms, the note-leak one in 140ms.

**Declined: `rel="noopener noreferrer"`**

Not a security issue here:

- Both links are **same-origin** (`/content/tos`, `/content/content-appeal`). Tabnabbing is a cross-origin attack; there is no vector.
- Browsers have implied `noopener` for `target="_blank"` since Chrome 88 / Firefox 79 / Safari 12.1.
- "Other parts of the codebase already enforce this pattern" is not accurate — 27 of 252 `target="_blank"` occurrences carry `rel`, about 11%, and no `react/jsx-no-target-blank` rule is configured.
- These anchors are copied verbatim from the code this PR replaced, so the pattern is pre-existing rather than introduced here.

Adding it to 2 of 252 call sites would be inconsistent rather than safer. If the convention is wanted, the eslint rule plus a sweep is the way to do it, as its own change.

PR #4567, opened 2026-09-01
feat(app-blocks): stratify the launch histograms by whether the guest announced BLOCK_HELLO

ZacxDev:
## Why

The re-post cadence change in #4542 **cannot be graded on the metrics it shipped with**, and the reason is structural rather than a query mistake.

Two facts compound:

1. `civitai_app_block_launch_phase_seconds` carries **no `app_block_id` label** — a deliberate cardinality decision, with per-app attribution pushed to the structured log. So `init_wait` cannot be stratified by app in Prometheus at all.
2. The deployed apps that already ship the **`BLOCK_HELLO` accelerator are the majority of launch traffic.** `BLOCK_HELLO` makes the host re-post `BLOCK_INIT` immediately on the guest's hello — short-circuiting the very wait the backoff shortens.

So a fleet-wide `init_wait` before/after is mostly composed of launches the change barely affects, that mass cannot be separated, and **a diluted null would be indistinguishable from "the change did nothing."** That is the same class of defect that has made pre-registered measurements ungradeable before: the design gets reviewed, the intervention ships, and the *instrument* turns out unable to answer the question.

## What this adds

A code-owned two-value label, **`hello: yes|no`**, on both launch histograms.

The beacon carries a **boolean**; the server maps it onto its own literals via `launchHelloLabel`. Nothing a client sends can become a label value — the same rule `phase` already follows, and what lets this beacon body stay public.

### 🔴 What the label means — picked, and pinned

**`yes` = the guest sent `BLOCK_HELLO` at some point during the launch window.**

It does **not** mean "the accelerator fired an extra `BLOCK_INIT`". Those are different facts, and they disagree on a reachable launch: `notifyHello()` posts only when the controller has started, has not stopped, and has not already handled a hello — so a hello arriving *before* `start()` is recorded and posts nothing, because `start()` posts immediately anyway.

Under the "accelerator fired" reading that launch is filed `no`, which is the wrong bucket: the guest's listener was demonstrably attached, so the host's very next post was heard and the cadence never governed the wait. Filing it `no` would push fast, accelerator-capable launches into the population that exists to isolate cadence-bound ones — **biasing the read toward "the change did nothing"**, i.e. making the instrument agree with the null regardless of the truth.

That is why the mark is set in the host's `BLOCK_HELLO` message handler, *before and independently of* the controller call, and why a source gate asserts it stays there. An ambiguous label is how a relabelled series produces a confident wrong number; this one is asserted as behaviour, not merely described.

### 🔴 Absent is not `false`

A client older than this field omits it; a launch that saw no hello sends `false`. `launchHelloLabel` returns `null` for the first and **drops** the sample rather than labelling it `no` — otherwise every stale-client launch lands in the exact population being isolated, as a silent systematic bias. The client therefore emits the key unconditionally, including for `false`, so the two cases stay distinguishable on the wire.

`launch_total_seconds` takes no `hello` label and is observed *before* the gate, so a rollout-window drop shows up as a `_count` divergence rather than vanishing.

### Cardinality

Derived from the bucket arrays in the metrics module — a histogram label set costs finite edges + `+Inf` + `_sum` + `_count`:

| metric | edges | per set | before | after | added |
|---|---|---|---|---|---|
| `launch_phase_seconds` | 13 | 16 | `phase(2)` = 32 | `phase(2)×hello(2)` = 64 | **+32** |
| `launch_init_posts` | 14 | 17 | unlabelled = 17 | `hello(2)` = 34 | **+17** |
| | | | | | **+49 / pod** |

For scale, `launch_total_seconds` already costs ~832 series per pod at full app saturation and is **unchanged**. `hello` is deliberately *not* added there — it already carries `app_block_id`, so the product would be ~1,664.

**Both** histograms get the label, not just one. `init_posts` is the mechanism discriminator — the `le=1` share answers "how often was the first post enough" — and that question is only meaningful *within* a `hello` value. Pooled, the share is dominated by accelerator apps; `hello="no"` is the population whose first-post sufficiency the cadence actually governs.

## 🔴 Retroactive limitation

**Data recorded before this label shipped carries no `hello` and can never be stratified.** This enables a post-change **within-period** contrast (`hello="no"` vs `hello="yes"`) and a cleaner post-vs-pre read restricted to the `hello="no"` population. **It does not repair the existing baseline.**

## Beacon discipline — verified, not assumed

`blockRenderTrackerPayload` is still an **allowlist** naming the three real columns, so `hello` cannot reach ClickHouse by construction. I re-read it rather than trusting the previous conclusion.

Both writers are covered by a test each asserting **exact payload equality**, not `not.toHaveProperty('hello')` — a field-specific assertion would stay green if the allowlist were ever turned back into a destructure-rest denylist, re-opening the class for every future field.

The existing discipline is inherited unchanged: observed only on `status=ok && !secondary`, gated behind the `total` anchor, dropped (never clamped) out of range, and dropped entirely when the tab was ever hidden.

One in-repo invariant was **restated rather than relaxed**: a browser test asserted "every `timings` field is a number", which was the same rule as "no strings on the wire" only while every field happened to be numeric. It now asserts the property it was always protecting — no strings, numbers positive, booleans allowed — since a boolean is a closed two-value domain mapped server-side and weakens nothing.

## Verification

**Red arm** — reverting only this PR's four source files to `origin/main` and re-running the six affected suites: **6 files / 28 failed, 120 passed**. Restoring: **148 passed**. (A first attempt at this measured the *green* arm by mistake — an unquoted shell variable meant the revert silently no-op'd. The control I print alongside it, asserting the symbol is genuinely absent from all four files, is what caught it, and it is why that control is there.)

Each guard was mutation-checked, with every target re-derived from the current source and an assertion that it still matched before mutating:

| mutation | killed by |
|---|---|
| absent `hello` labelled `no` instead of dropped | `expected 'no' to be null`, plus the drop test |
| record `helloSeen` inside `notifyHello()` (the meaning flip) | the wiring gate |
| omit `hello` when `false` | `expected undefined to be false` + the key-presence test |
| don't reset `helloSeen` on soft navigation | the re-arm test |
| leak `hello` into the ClickHouse allowlist | **both** writer tests, each on its own payload equality |

Full unit project: **1566 files / 24875 tests, 0 failures.** `AppBlocks` browser suite: **40 files / 477 tests, 0 failures.** Typecheck and ESLint clean.

Not merged.

### DIFF
--- a/src/components/AppBlocks/PageBlockHost.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHost.browser.test.tsx
@@ -1256,11 +1256,29 @@ describe('PageBlockHost block render/impression (Analytics Phase 2)', () => {
     );
     expect(typeof body.timings.totalMs).toBe('number');
     expect(body.timings.totalMs).toBeGreaterThan(0);
-    // Never a zero-valued leg on the wire — an unobserved leg is OMITTED.
+    // 🔴 NO STRINGS ON THE WIRE — this is the real invariant, and it is what
+    // keeps a public, client-controlled beacon body from touching prom label
+    // cardinality. Every label this payload feeds is CODE-OWNED: the server maps
+    // named numeric fields onto its own `phase` literals and the `hello` boolean
+    // onto `yes`/`no`, so no value here can ever BE a label value.
+    //
+    // It used to be written as "every field is a number", which was the same
+    // rule while every field happened to be one. `hello` is a BOOLEAN and does
+    // not weaken the guarantee — booleans are a closed two-value domain, mapped
+    // server-side — so the assertion is restated at the level of the property it
+    // was always protecting rather than relaxed to let a boolean through.
     for (const [k, v] of Object.entries(body.timings)) {
-      expect(typeof v, `timings.${k}`).toBe('number');
-      expect(v, `timings.${k}`).toBeGreaterThan(0);
+      expect(typeof v, `timings.${k}`).not.toBe('string');
+      expect(['number', 'boolean'], `timings.${k}`).toContain(typeof v);
+      // Never a zero-valued NUMERIC leg on the wire — an unobserved leg is
+      // OMITTED, because a 0 is indistinguishable from an instant one and drags
+      // every percentile down. Booleans are exempt: `hello: false` is a real
+      // observation, not a missing one, and is emitted deliberately (an omitted
+      // field would be indistinguishable from a client that cannot send it).
+      if (typeof v === 'number') expect(v, `timings.${k}`).toBeGreaterThan(0);
     }
+    // `hello` specifically: always present, always boolean.
+    expect(typeof body.timings.hello, 'timings.hello').toBe('boolean');
     // No isAnon/userId from the client — those are server-derived in the route.
     expect(body).not.toHaveProperty('isAnon');
     expect(body).not.toHaveProperty('userId');
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -1337,6 +1337,16 @@ export function PageBlockHost({
   // immediately.
   useEffect(() => {
     const off = onMessage<unknown>('BLOCK_HELLO', () => {
+      // 🔴 RECORDED BEFORE — AND INDEPENDENTLY OF — THE CONTROLLER, deliberately.
+      // The label means "the guest announced during this launch", NOT "the
+      // accelerator fired an extra post". `notifyHello()` is a no-op when the
+      // controller has not started, has stopped, or has already handled a hello;
+      // recording inside it would file those launches as `no` even though the
+      // guest's listener was demonstrably attached and the host's next post was
+      // heard. That is the wrong bucket, and it biases the comparison toward the
+      // null. See `LaunchMarks.helloSeen` for the full argument.
+      const marks = launchMarksRef.current;
+      if (marks) marks.helloSeen = true;
       controllerRef.current?.notifyHello();
     });
     return off;
--- a/src/components/AppBlocks/__tests__/launchSampleBound.test.ts
+++ b/src/components/AppBlocks/__tests__/launchSampleBound.test.ts
@@ -338,6 +338,65 @@ describe('PageBlockHost: launch-mark seed wiring', () => {
     expect(body).not.toContain('BLOCK_READY');
   });
 
+  /**
+   * 🔴 THE STRATIFIER'S WIRING — the one part of `hello` no pure test can reach,
+   * and the part that decides whether the label means what it says.
+   *
+   * `helloSeen` must be set in the BLOCK_HELLO message handler, INDEPENDENTLY of
+   * `controllerRef.current?.notifyHello()`. Recording it inside `notifyHello()`
+   * instead would silently change the label's MEANING from "the guest announced"
+   * to "the accelerator fired an extra post" — `notifyHello()` is a no-op when
+   * the controller has not started, has stopped, or has already handled a hello.
+   * A hello arriving before `start()` would then be filed `hello="no"` despite
+   * the guest's listener being attached, pushing fast accelerator-capable
+   * launches into the cadence-bound population and biasing the whole comparison
+   * toward "the change did nothing".
+   *
+   * Nothing else can catch that: every pure test would stay green, the beacon
+   * would still carry a boolean, and the metric would still have two
+   * populations — just the wrong ones. A source gate is the weaker instrument
+   * and is named as such, for the same reason as the seed gate above: the
+   * behavioural version can only live in the `component` project, which CI does
+   * not run.
+   */
+  function helloHandlerBody(src: string): string {
+    const start = src.indexOf("onMessage<unknown>('BLOCK_HELLO'");
+    if (start === -1) return '';
+    const end = src.indexOf('});', start);
+    if (end === -1) return '';
+    return src.slice(start, end);
+  }
+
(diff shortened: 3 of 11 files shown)


ZacxDev:
## Adversarial audit — 1 finding, fixed in `30aad96ce4`

The finding is against **this PR's own first design**, and it was caught by measurement rather than by re-reading the reasoning.

---

### 🟠 MEDIUM — drop-on-absent was a silent coverage regression, and a latency-correlated one

`launchHelloLabel` returned `null` for a beacon with no `hello` boolean, and the observe path skipped both labelled histograms.

**Measured:** against the pre-label module a hello-less beacon contributes **one** `launch_phase_seconds` sample; against the drop-based revision it contributes **none** — while `launch_total_seconds` counts it either way. So an existing metric quietly lost coverage.

Two things make that worse than it first appears:

**1. The gap is not a rollout window.** The "old client" here is a **browser bundle**, not a server pod. `/api/track/block-render` deliberately skips the tRPC middleware chain, so it runs **no client-version gate at all** — a tab opened before the deploy keeps sending the old shape for as long as it stays open. Open-ended, not bounded by a converge. I had reasoned about this as a deploy-window cost, and that was simply the wrong mental model.

**2. The drop is latency-correlated**, so it is a selection effect, not neutral loss. A bundle predating this label also predates the re-post cadence change, so the dropped launches are systematically the **slow-cadence** ones. Removing them makes the unstratified phase aggregate look *faster* across the deploy — the direction that flatters the very change these metrics exist to grade. "Dropping is the safe direction" only holds when droppedness is independent of latency. Here it is not.

**And the safeguard I claimed for it did not exist.** The previous revision argued the loss "shows up as a `_count` divergence" against `launch_total_seconds`. Nothing observes that — a search of the deployment repo finds these series in **documents only**: no alert, no dashboard, no recording rule. An unmonitored tell is not a safeguard, and I should not have offered it as one.

**Fix — the third option, which I had dismissed too fast.** A `hello="unknown"` value. `launchHelloLabel` becomes **total**, and the observe path no longer gates on it. Strictly better on every axis that mattered:

- **Coverage** — `sum without (hello)` recovers exactly the pre-label series, so no existing query loses launches.
- **Purity** — `yes`/`no` stay uncontaminated, which was the whole reason absent could not be folded into `no`.
- **Observability** — the stale-client tail is a value anyone can read off the metric instead of a divergence nobody was watching.

**Cardinality, re-derived** (a label set costs finite edges + `+Inf` + `_sum` + `_count`):

| metric | edges | /set | before | after | added |
|---|---|---|---|---|---|
| `launch_phase_seconds` | 13 | 16 | `phase(2)` = 32 | `phase(2)×hello(3)` = 96 | **+64** |
| `launch_init_posts` | 14 | 17 | unlabelled = 17 | `hello(3)` = 51 | **+34** |
| | | | | | **+98 / pod** |

Up from +49. Still small against the ~832/pod `launch_total_seconds` already costs, and buying back 100% coverage of an existing metric is worth the third value.

**A test that was green for the wrong reason.** The drop test asserted only that `yes` and `no` did not move — which stayed green after the drop was removed, because a *bucketed* sample does not move them either. Replaced with one asserting `unknown` **moves**, plus a coverage invariant: summing the phase histogram over every `hello` value yields exactly one sample per launch. A negative-only assertion invites exactly this.

---

### Audited, no change needed

**The restated browser assertion is equivalent-or-stronger.** I flagged this myself and treated it as a finding to disprove rather than a disclosure. Proved by mutation in both directions: emitting `hello` as a string fails it (`expected 'string' not to be 'string'`), and emitting a zero-valued numeric leg fails it (`expected 0 to be greater than 0`). Enumerating every `typeof`: string, object, null, undefined, bigint, symbol and function are all still rejected; the *only* widening is boolean. No hole.

**Label semantics under adversarial reading** — the reachable states, and where each lands:

| state | bucket | right? |
|---|---|---|
| hello on a **retried** launch (marks persist, frame remounted) | `yes` | ✅ same app bundle reloads, so announce behaviour cannot change mid-launch |
| **duplicate** hello | `yes` | ✅ idempotent; the controller's once-latch is separate and unaffected |
| hello from a **different frame** after a soft nav | rejected | ✅ `usePostMessage` pins `event.source` to this host's own `contentWindow` |
| hello **after** `BLOCK_READY`, before the beacon effect | `yes` | ⚠️ named, not fixed — see below |
| hello **after** the beacon | not counted | ✅ the sample was already built |

The one window that can mis-file is the single React commit between `BLOCK_READY` and the beacon effect. It needs no code: a guest sends hello when its listener attaches, which necessarily precedes any `BLOCK_READY` over that same listener, so a hello landing there is a duplicate or out-of-order delivery — and the guest genuinely did announce, which is exactly what the label claims. It is now pinned by a test rather than left as a surprise.

**Lifetime matches `initPosts` exactly** — one clearing call site gated on `blockInstanceId`, so both survive a retry and both are cleared by a soft navigation. That symmetry is load-bearing: `init_wait` spans every attempt of one launch, so its two companions must span the same window or the three describe different things while looking like one record. Pinned by tests.

---

### Verification

Delta mutations, each target re-derived from current source with a match-count assertion before mutating, each killed on its own assertion:

| mutation | killed by |
|---|---|
| collapse `unknown` back into `no` | `expected 'no' to be 'unknown'` + the bucketing test |
| reinstate the drop | the bucketing test + the coverage invariant (`expected 12 to be 13`) |
| emit `hello` as a string | the no-strings assertion |
| emit a zero-valued numeric leg | the positive-number assertion |

Full unit project: **1566 files / 24880 tests, 0 failures.** `AppBlocks` browser suite: **40 files / 477 tests, 0 failures.** Typecheck and ESLint clean.

A second round auditing only the delta found nothing, which is where the ladder ends.

Not merged.

PR #4573, opened 2026-09-02
fix(upload): relay an image upload through our own origin when the direct PUT cannot resolve the storage host

ZacxDev:
> **Draft.** Two adversarial audit rounds have run; round 2 was not clean, so a round 3 is in flight. Round-by-round findings are in the comments — including a correction of this description's original claims, which were wrong.

## The problem

Image uploads are browser-direct: `/api/v1/image-upload` mints a key and returns a presigned PUT URL, and the browser PUTs straight at the storage endpoint. That endpoint is a **raw storage hostname, not a civitai.com one**.

So a user whose DNS cannot resolve it gets `ERR_NAME_NOT_RESOLVED` on the PUT while everything else on the page works — the presign call itself succeeds, because that one goes to us. It presents as *"uploads are broken for me specifically"*, and **six reported tickets share this signature**.

## The fix

A fallback the client reaches for **only after the direct PUT has already failed at the network layer**: POST the bytes to `/api/v1/image-upload/relay` — an origin the browser has demonstrably resolved, since it just fetched the presign from it — and do the store write server-side through the existing `uploadImageBufferToStore`.

Same bucket, same backend, same media-location registration, so the resulting object is servable exactly as a direct upload would be. **No new storage backend, no config change, no DNS or infrastructure change.**

## Decisions worth reviewing

**The key is minted server-side; no caller-supplied key is honoured.** An earlier draft echoed the presign's `id` back so both paths would agree — but that id travels through the browser, so honouring it lets a caller name any uuid and **overwrite another user's object**.

**Shedding is survivable, and rare.** `MAX_CONCURRENT_RELAYS = 8` is a memory bound (~800MB peak, since the route buffers whole and `Buffer.concat` roughly doubles it). ⚠️ An earlier version of this description said a shed is *routine* because the dropzone uploads 10 concurrent files against a cap of 8. **That was wrong** and it inverted the action it implied: the cap is per **pod**, the serving pool's replica floor is in the dozens, and there is no session affinity, so one browser's concurrent POSTs spread across pods rather than colliding. A shed means ~8 concurrent relays hit one pod — for a fallback that only fires for DNS-broken clients, an **anomaly worth investigating**, not business as usual. The client still retries once honouring `Retry-After` (cheap insurance, not a routine necessity), and it sheds with **429**, not 503, so deliberate backpressure is not counted as a server fault in the route's own 5xx attribution.

**The origin check is wider than either in-repo precedent.** It 403s an API-key client with no `Origin`. Deliberate: the relay exists to rescue *browsers* whose DNS cannot resolve the storage host, and a non-browser client has neither that failure mode nor a reason to push bytes through us. Note this route is **not** cookie-only — `getServerAuthSession` accepts Bearer and `?token=` — so `submit-version.ts`'s "no bearer exemption needed" reasoning does not transfer.

**Not the default path.** Every relayed byte transits the app servers. There is a comment saying so at the route.

## Known gaps, not fixed here

- **The presign leaves a stranded registry row** when the fallback fires — a key registered with `sizeBytes: 0` that no object was written to. ⚠️ An earlier version of this description claimed existing orphan reconciliation sweeps it. **That was wrong**: the reconcilers list bucket *objects* and diff them against the registry, so a row with no object is not in their universe. Documented at the route as an unswept cost. One row per fallback. A size-keyed sweep would be unsafe, because the direct path also registers `sizeBytes: 0` for every *successful* upload.
- **No metric and no runtime kill switch.** The cap is a module constant, so changing it needs a deploy.
- **The multipart path is untouched** and has no fallback.

## Tests

`relay.test.ts` (15), `upload-settlement.test.ts` (8), `relay-with-retry.test.ts` (5) — **28 cases**. Cases marked `AUDIT-*` pin defects an audit found and are red against the draft that had them.

**Eight mutants, each killed by its own tests and no others** — the point being that each guard is reachable and fails for its own reason, not that "a test failed":

| mutant | result |
|---|---|
| measure-at-end cap | 1 failed / 22 passed |
| re-added `req.destroy()` | 1 / 22 |
| shed with 503 | 1 / 22 |
| `relayPending` yield removed | 5 / 18 |
| never retry on 429 | 3 / 25 |
| retry on any status | 1 / 27 |
| ignore `Retry-After` | 1 / 27 |
| skip abort-during-backoff | 1 / 27 |

Full unit tier **1569 files / 24873 tests, 0 failures**. Cold typecheck (`tsconfig.tsbuildinfo` deleted) **0 errors** — it caught a real one first. Prettier + ESLint clean on all changed files.

## 🔴 Explicitly NOT verified

- **No end-to-end upload against a real backend.** The store call is mocked. The ticket's closing condition asks for a preview-environment upload; that has not been done.
- **The browser tier was NOT run** — the flake's pinned Playwright driver build does not match the installed browser package on this host. Not run is not a claim that it passes, and those are the tests that would exercise the changed hook in a real DOM.
- **The DNS failure itself was not reproduced.** The six referenced tickets are taken at face value.
- **The memory arithmetic behind the cap is reasoned from manifests, not load-tested.**
- **The 413-under-backpressure mechanism is reasoned from Node socket semantics**, not reproduced against a real socket.
- **What share of affected uploads are large or multipart is unmeasured**, so how much of the reported problem the 50MB cap actually covers is unknown.

### DIFF
--- a/src/__tests__/pages/api/v1/image-upload/relay.test.ts
+++ b/src/__tests__/pages/api/v1/image-upload/relay.test.ts
@@ -1,67 +1,124 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { Readable } from 'stream';
 import type { NextApiRequest, NextApiResponse } from 'next';
+import type * as EnvOther from '~/env/other';
+import type * as AuthSession from '~/server/auth/get-server-auth-session';
+import type * as S3Utils from '~/utils/s3-utils';
+import type * as HttpErrors from '~/server/prom/http-errors';
+import type * as OriginHelpers from '~/server/utils/origin-helpers';
 
 // Covers the FALLBACK image-upload relay (`/api/v1/image-upload/relay`).
 //
 // These are NEW-FEATURE tests, not regression guards: the route did not exist before
 // this change, so there is no pre-change revision at which they could be shown red.
 // Stated explicitly so nobody later reads them as evidence that a specific defect was
-// reproduced and fixed.
+// reproduced and fixed. The exceptions are the three cases marked AUDIT-*, which DO
+// pin defects an adversarial audit found in the first draft of this route and which
+// are red against that draft.
 //
-// The properties worth pinning are the ones where getting it wrong is silent:
-//  - the key is minted SERVER-SIDE and no caller-supplied key is honoured (an
-//    earlier draft echoed the presign's id back, which lets a caller overwrite
-//    another user's object)
-//  - the size cap trips on the RUNNING total, so an oversized body cannot be
-//    buffered in full first
-//  - an unauthenticated or banned caller never reaches the store at all
+// Mocks are SURGICAL (spread `importOriginal`, override one symbol). A one-key
+// wholesale factory for `~/utils/s3-utils` — which exports ~30 runtime symbols —
+// collapses the whole file to "no tests" the moment the route imports a second
+// symbol from it, which is the silent-zero shape `local-rules/no-wholesale-module-mock`
+// exists to prevent.
 
-const { mockGetServerAuthSession, mockUploadImageBufferToStore } = vi.hoisted(() => ({
+const {
+  mockGetServerAuthSession,
+  mockUploadImageBufferToStore,
+  mockIsAllowedOriginRequest,
+  prodFlag,
+} = vi.hoisted(() => ({
   mockGetServerAuthSession: vi.fn(),
   mockUploadImageBufferToStore: vi.fn(),
+  mockIsAllowedOriginRequest: vi.fn(),
+  // A getter-backed box, so a single test can flip `isProd` without a second file.
+  prodFlag: { value: false },
 }));
 
-vi.mock('~/server/auth/get-server-auth-session', () => ({
+vi.mock('~/env/other', async (importOriginal) => ({
+  ...(await importOriginal<typeof EnvOther>()),
+  get isProd() {
+    return prodFlag.value;
+  },
+}));
+
+vi.mock('~/server/auth/get-server-auth-session', async (importOriginal) => ({
+  ...(await importOriginal<typeof AuthSession>()),
   getServerAuthSession: mockGetServerAuthSession,
 }));
 
-vi.mock('~/utils/s3-utils', () => ({
+vi.mock('~/utils/s3-utils', async (importOriginal) => ({
+  ...(await importOriginal<typeof S3Utils>()),
   uploadImageBufferToStore: mockUploadImageBufferToStore,
 }));
 
-vi.mock('~/server/prom/http-errors', () => ({
+vi.mock('~/server/prom/http-errors', async (importOriginal) => ({
+  ...(await importOriginal<typeof HttpErrors>()),
   instrumentApiResponse: vi.fn(),
 }));
 
-import handler, { MAX_RELAY_BYTES } from '~/pages/api/v1/image-upload/relay';
+vi.mock('~/server/utils/origin-helpers', async (importOriginal) => ({
+  ...(await importOriginal<typeof OriginHelpers>()),
+  isAllowedOriginRequest: mockIsAllowedOriginRequest,
+}));
+
+import handler, {
+  MAX_RELAY_BYTES,
+  MAX_CONCURRENT_RELAYS,
+  __getInFlightForTest,
+} from '~/pages/api/v1/image-upload/relay';
+
+/** Records the order in which the handler wrote a status vs destroyed the request. */
+type Trace = string[];
 
 function makeReq(
-  body: Buffer,
-  opts?: { method?: string; contentType?: string }
+  chunks: Buffer[],
+  opts?: { method?: string; contentType?: string; trace?: Trace }
 ): NextApiRequest {
-  const stream = Readable.from([body]) as unknown as NextApiRequest;
+  return decorate(Readable.from(chunks) as unknown as NextApiRequest, opts);
+}
+
+function decorate(
+  stream: NextApiRequest,
+  opts?: { method?: string; contentType?: string; trace?: Trace }
+): NextApiRequest {
   stream.method = opts?.method ?? 'POST';
   stream.headers = { 'content-type': opts?.contentType ?? 'image/png' };
   stream.query = {};
+  const trace = opts?.trace;
+  const realDestroy = stream.destroy.bind(stream);
+  stream.destroy = ((...args: unknown[]) => {
+    trace?.push('destroy');
+    return (realDestroy as (...a: unknown[]) => unknown)(...args);
+  }) as typeof stream.destroy;
   return stream;
 }
 
-function makeRes() {
+function makeRes(trace?: Trace) {
   const res = {
     statusCode: 0,
     body: undefined as unknown,
     headers: {} as Record<string, string>,
+    headersSent: false,
     status(code: number) {
+      trace?.push(`status:${code}`);
       this.statusCode = code;
       return this;
     },
     json(payload: unknown) {
       this.body = payload;
+      this.headersSent = true;
+      return this;
+    },
+    end() {
+      this.headersSent = true;
       return this;
     },
     setHeader(k: string, v: string) {
       this.headers[k] = v;
+    },
+    getHeader(k: string) {
+      return this.headers[k];
     },
   };
   return res as unknown as NextApiResponse & typeof res;
@@ -71,13 +128,21 @@ const authed = { user: { id: 7, bannedAt: null } };
 
 beforeEach(() => {
   vi.clearAllMocks();
+  prodFlag.value = false;
   mockGetServerAuthSession.mockResolvedValue(authed);
-  mockUploadImageBufferToStore.mockResolvedValue({ key: 'server-minted-uuid', backend: 'backblaze' });
+  mockUploadImageBufferToStore.mockResolvedValue({
+    key: 'server-minted-uuid',
+    backend: 'backblaze',
+  });
+  // Default to an allowed origin; the CSRF cases override it. In test `isProd` is
+  // false, so the guard is exempt anyway — these assertions pin the call, and the
+  // prod behaviour is pinned by the dedicated case below.
+  mockIsAllowedOriginRequest.mockReturnValue(true);
 });
 
 describe('image-upload relay', () => {
   it('uploads the body and returns the SERVER-minted key', async () => {
-    const req = makeReq(Buffer.from('image-bytes'));
+    const req = makeReq([Buffer.from('image-bytes')]);
     const res = makeRes();
 
     await handler(req, res);
@@ -85,17 +150,25 @@ describe('image-upload relay', () => {
     expect(res.statusCode).toBe(200);
     expect(res.body).toEqual({ id: 'server-minted-uuid' });
 
-    // The bytes reached the store intact, with the caller's content type.
     expect(mockUploadImageBufferToStore).toHaveBeenCalledTimes(1);
     const [buf, opts] = mockUploadImageBufferToStore.mock.calls[0];
     expect(Buffer.from(buf).toString()).toBe('image-bytes');
     expect(opts).toEqual({ contentType: 'image/png' });
   });
 
+  it('reassembles a MULTI-CHUNK body in order', async () => {
+    const req = makeReq([Buffer.from('abc'), Buffer.from('def'), Buffer.from('ghi')]);
+    const res = makeRes();
+
+    await handler(req, res);
+
+    expect(res.statusCode).toBe(200);
+    const [buf] = mockUploadImageBufferToStore.mock.calls[0];
+    expect(Buffer.from(buf).toString()).toBe('abcdefghi');
+  });
+
   it('ignores any caller-supplied key — the store mints it', async () => {
-    // A caller trying to name the object it writes. `uploadImageBufferToStore` takes
-    // only (bytes, {contentType}); if a future edit threads a key through, this fails.
-    const req = makeReq(Buffer.from('bytes'));
+    const req = makeReq([Buffer.from('bytes')]);
     (req as unknown as { query: Record<string, string> }).query = {
       key: 'victim-uuid',
       id: 'victim-uuid',
@@ -106,13 +179,16 @@ describe('image-upload relay', () => {
 
     expect(res.statusCode).toBe(200);
     expect(res.body).toEqual({ id: 'server-minted-uuid' });
-    const [, opts] = mockUploadImageBufferToStore.mock.calls[0];
-    expect(opts).not.toHaveProperty('key');
     expect(JSON.stringify(mockUploadImageBufferToStore.mock.calls[0])).not.toContain('victim-uuid');
   });
 
-  it('rejects a body over the cap without buffering it whole', async () => {
-    const req = makeReq(Buffer.alloc(MAX_RELAY_BYTES + 1));
+  // Chunks each UNDER the cap that only exceed it in sum. NOTE this asserts the
+  // OUTCOME only — see the next test for the one that pins the running-total
+  // property itself. Kept separate because the 413 and the property are different
+  // claims and a reader should not mistake this one for the stronger one.
+  it('rejects a body whose chunks each fit but which sums past the cap', async () => {
+    const half = Math.floor(MAX_RELAY_BYTES / 2) + 1;
+    const req = makeReq([Buffer.alloc(half), Buffer.alloc(half)]);
     const res = makeRes();
 
     await handler(req, res);
@@ -121,10 +197,39 @@ describe('image-upload relay', () => {
     expect(mockUploadImageBufferToStore).not.toHaveBeenCalled();
   });
 
+  // AUDIT-F5 + AUDIT-F6, and this single trace pins BOTH — deliberately, because
+  // neither is expressible on its own here.
+  //
+  //  * F5: `req.destroy()` tears down the socket, so a status written afterwards
+  //    goes into a dead socket and the client sees a transport error, never the 413.
+  //    Hence `status:413` must precede `destroy`.
+  //  * F6: the running-total property. Asserting the 413 alone does NOT pin it — a
+  //    measure-at-the-end implementation returns the same 413, which is why the
+  //    first draft's cap test left that mutant alive. The tell is the LEADING
+  //    entry: measure-at-end consumes to EOF, and a stream that ends auto-destroys,
+  //    producing ['destroy', 'status:413', 'destroy']. A running total throws before
+  //    EOF, so no destroy precedes the status. Confirmed by mutation.
+  //
+  // ⚠ What does NOT work, so nobody re-derives it: counting how many chunks the
+  // async source yielded. Node's Readable reads AHEAD of the consumer, so the real
+  // running-total implementation still pulls the chunk after the one that trips the
+  // cap. Generator-pull count measures the stream's buffering, not the handler's.
+  it('writes the 413 before destroying, and does not consume the stream to EOF', async () => {
+    const trace: Trace = [];
+    const half = Math.floor(MAX_RELAY_BYTES / 2) + 1;
+    const req = makeReq([Buffer.alloc(half), Buffer.alloc(half)], { trace });
+    const res = makeRes(trace);
+
+    await handler(req, res);
+
+    expect(res.statusCode).toBe(413);
+    expect(trace).toEqual(['status:413', 'destroy']);
+  });
+
   it('accepts a body exactly AT the cap — the boundary is inclusive', async () => {
     // Guards the off-by-one directly: `>` not `>=`. A fixture one byte under would
     // pass against either spelling and prove nothing about the boundary.
-    const req = makeReq(Buffer.alloc(MAX_RELAY_BYTES));
+    const req = makeReq([Buffer.alloc(MAX_RELAY_BYTES)]);
     const res = makeRes();
 
     await handler(req, res);
@@ -134,7 +239,7 @@ describe('image-upload relay', () => {
   });
 
   it('rejects an empty body', async () => {
-    const req = makeReq(Buffer.alloc(0));
+    const req = makeReq([Buffer.alloc(0)]);
     const res = makeRes();
 
     await handler(req, res);
@@ -145,7 +250,7 @@ describe('image-upload relay', () => {
 
   it('rejects an unauthenticated caller before touching the store', async () => {
     mockGetServerAuthSession.mockResolvedValue(null);
-    const req = makeReq(Buffer.from('bytes'));
+    const req = makeReq([Buffer.from('bytes')]);
     const res = makeRes();
 
     await handler(req, res);
@@ -158,7 +263,7 @@ describe('image-upload relay', () => {
     mockGetServerAuthSession.mockResolvedValue({
       user: { id: 7, bannedAt: new Date('2026-01-01') },
     });
-    const req = makeReq(Buffer.from('bytes'));
+    const req = makeReq([Buffer.from('bytes')]);
     const res = makeRes();
 
     await handler(req, res);
@@ -168,7 +273,7 @@ describe('image-upload relay', () => {
   });
 
   it('rejects a non-POST method', async () => {
-    const req = makeReq(Buffer.from('bytes'), { method: 'GET' });
+    const req = makeReq([Buffer.from('bytes')], { method: 'GET' });
     const res = makeRes();
 
     await handler(req, res);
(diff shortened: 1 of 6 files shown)


ZacxDev:
## 🔴 Correcting this PR's own description — an adversarial audit refuted two of its claims. Converted to draft.

Posting rather than editing the body, because a reviewer may already have read the wrong version.

### The headline claim is FALSE, and the change is currently worse than not shipping it

The body says *"the returned `id` becomes whichever key actually holds the bytes."* It does not.

Per the XHR spec, `error` at the XHR object is followed by `loadend` **in the same dispatch**. This PR replaced the `error` handler's synchronous `reject(...)` with a promise chain, so `loadend` now runs first, computes `success = false`, and calls `resolve(false)` — settling the outer promise **before** the relay's `fetch` can complete.

Verified by running a real jsdom `XMLHttpRequest` against a connection-resetting server, with a base-revision control:

```
HEAD:  error -> relay:start -> loadend(success=false) -> RESOLVE(false)
       uploadToCF returned { id: 'PRESIGN-ID' }        <- key holds NO bytes
       resolvedId became 'RELAYED-ID' 60ms LATER, after the return
BASE:  reject: "Upload failed (status 0)"              <- the error signal this PR removes
```

Three consequences, and the second is the serious one:

1. Callers submit a key with no object → edge 404 → ingestion `NotFound`.
2. **The loud failure is gone.** Before this PR the hook threw and callers surfaced an error. This converts a visible upload failure into a silent bad-id success.
3. The relay still runs to completion in the background (up to 50MB of app-server memory) with nobody consuming the result, and its late `updateFile` desynchronises the tracked file's `url` from the id the caller holds — so `removeImage(returnedId)` silently no-ops.

**Zero tests cover the hook change** in any tier that ran. The missing guard is a hook-level test with a stubbed XHR that dispatches `error` then `loadend`.

### The orphan-sweep justification looks wrong in direction

The body claims the orphaned zero-byte registration is swept by existing reconciliation. The reconcilers list **bucket objects** and diff them against a `file_locations` whitelist — they sweep orphan *objects*; a registry row *protects* a key rather than being swept. No job was found that deletes rows lacking an object. Impact is small (one row per fallback) but it was offered as a reason not to worry, so it should not stand unchallenged. *(Read from the deployment manifests, not the storage-resolver service source — so this is a flag, not a refutation.)*

### One test does not test what its name says

`rejects a body over the cap without buffering it whole` — `Readable.from([oneBigBuffer])` yields a **single** chunk, so a measure-at-the-end implementation passes it identically. Confirmed by mutation: replacing `readCappedBody` with a buffer-everything-then-check version left **9/9 green — the mutant SURVIVED**. The running-total property, which is the stated reason the cap exists, is unguarded. A killing fixture needs ≥2 chunks each under the cap, summing past it.

### What the audit CONFIRMED

- The central security claim is **true**: no caller-supplied key is honoured; the key is minted with `randomUUID()` server-side. Mutation-confirmed.
- `resolvedId`'s closure semantics are correct — the defect is event ordering, not staleness.
- Auth parity with the direct path holds, and is slightly stricter.
- The two mutation results quoted in the body reproduced exactly, each killing only its own tests.
- Cold typecheck 0 errors reproduced.

### Also found, not previously disclosed

- **CI is red for a reason this PR introduces**, not bookkeeping: the 500 handler returns `error.message` verbatim from the AWS SDK, which carries bucket names, endpoint hostnames, request ids and signature detail. It adds this route to a disclosure ledger's offender set. This PR's own test *pins the leaking contract*, so fixing the route means changing that test.
- **Cookie-authed, raw-body, state-changing POST with no origin check.** The repo has an explicit in-tree precedent adding one back for exactly this shape. Accepting an arbitrary `Content-Type` makes it a CORS-*simple* request — no preflight — so any third-party page could make a logged-in visitor upload attacker-chosen bytes. The direct presign route is not exploitable this way, so this is new surface, not parity.
- Whole-file buffering with ~2× amplification and no concurrency cap; the 413 is never delivered (the socket is destroyed first, and the test asserts against a socket-less fake stream); a relayed upload cannot be cancelled; no counter, log or kill switch.

**Converted to draft.** Not merging this until at least the ordering bug, the error-envelope leak and the origin check are fixed, and the hook change has a test that would have caught F1.

ZacxDev:
## Audit round 1 — fixes landed

Round-1 audit read `a15ac378c2`. Fixes are on `3784e811`. PR is draft; the round-1 findings and the public correction of this PR's description are in the comment above.

**Ledger** — round 1 · payload lines changed THIS round: **241 + / 62 −** across `relay.ts`, `useCFImageUpload.tsx`, `upload-settlement.ts` (scaffolding, not counted: 427 + / 36 − across the two test files) · since round 1: same, this is the first fix round.

**Verification at `3784e811`:** full unit tier 1568 files / 24868 tests, 0 failures. Cold typecheck (`tsconfig.tsbuildinfo` deleted) 0 errors. Prettier + ESLint clean on all added files. Six mutants run, each killed by its own tests and no others — `loadend` settling unconditionally (5 failed / 3 passed), measure-at-the-end cap (1/10), origin check removed (2/11), `error.message` leaked (1/12), concurrency cap removed (1/14), slot released outside `finally` (2/13).

**NOT verified:** no end-to-end upload against a real backend; the browser tier exercising this hook cannot run on this host (Playwright driver pin) so it was **not run**, which is not a claim that it passes; the DNS failure itself was not reproduced; the memory arithmetic behind the concurrency cap is reasoned from manifests, not load-tested.

```audit-claims round=1 audited=a15ac378c2..3784e811
1. F1 settlement ordering: a `relayPending` flag set synchronously in the `error` handler makes the following `loadend` yield, so the promise resolves with the relayed key rather than the presign key, and the pre-existing throw on a failed upload is retained.
2. F1 guard binding: the settlement rule was extracted from the hook into `src/utils/upload-settlement.ts` and the hook now calls it, so the new tests drive production code instead of a re-implementation in the test.
3. F2 error envelope: the 500 path now calls `handleEndpointError` instead of returning `error.message`, and the test asserts the wire body rather than that a function was called.
4. F3 CSRF: an `isProd && !isAllowedOriginRequest(req)` check was added ahead of the session lookup, mirroring `blocks/submit-version.ts`.
5. F5 ordering: the 413 is written before `req.destroy()`.
6. F6 running-total: the cap test now asserts the request is not consumed to EOF, via the destroy/status trace, after a multi-chunk fixture and a generator-pull count were both found not to discriminate.
7. F4 memory: a module-level `MAX_CONCURRENT_RELAYS` cap sheds with 503, with slot release in `finally`.
8. F11 cancellation: the relay takes an `AbortSignal` and the tracked file's `abort` aborts both the xhr and the relay.
9. F7 mocks: the test's module mocks now spread `importOriginal` instead of one-key wholesale factories.
10. F8 formatting: prettier and eslint pass on the added files; `useCFImageUpload.tsx` was deliberately NOT reformatted because it was already unclean on `main` and the CI check enforces added files only.
11. F9 orphan row: no sweep was added; the route documents the stranded zero-byte registration as a known cost, and the claim that existing reconciliation sweeps it was withdrawn from the PR description.
12. F10 partially: no metric and no runtime kill switch were added; the cap is a module constant and the code says so.
13. Nit: the cap comment no longer claims 50MB is the largest media the app accepts.
```

ZacxDev:
## Audit round 2 — fixes landed

Round-2 delta audit read `3784e811`. Fixes are on `3265dc8d30`. Round 2 confirmed all three round-1 deploy-blockers fixed (each by its own mutation) and returned **seven 🟡s, no 🔴s** — four against the fix round's own new code. All seven are addressed.

**The PR description has been rewritten**, not just commented on: 🟡-5 was that the withdrawn orphan-sweep claim was still live in the body after being corrected in the code comment. It now says the opposite, explicitly flagged as a correction. The stale "9 new cases" and mutation denominators are updated too (28 cases, 8 mutants).

**Ledger** — round 2 · payload lines changed THIS round: **158 + / 46 −** (`relay.ts` 92+/29−, `upload-settlement.ts` 52+/17−, `useCFImageUpload.tsx` 14+/3−) · scaffolding 138+/19− · since round 1: 303. Payload is not trending to zero; the attribution gate does not fire.

**Verification at `3265dc8d30`:** full unit tier 1569 files / 24873 tests, 0 failures. Cold typecheck (`tsbuildinfo` deleted) 0 errors — it caught a real one first (`relayWithRetry`'s narrowed return type dropped `.json()` from the caller; now generic). Prettier + ESLint clean. Eight mutants, each killed by its own tests and no others: measure-at-end cap (1 failed/22), re-added destroy (1/22), 503 shed (1/22), `relayPending` removed (5/18), never-retry (3/25), retry-any-status (1/27), ignore `Retry-After` (1/27), skip abort-during-backoff (1/27).

**NOT verified** (unchanged): no end-to-end upload against a real backend; the browser tier cannot run on this host (Playwright driver pin) so it was **not run**, which is not a claim that it passes; the DNS failure was not reproduced; the memory arithmetic is reasoned from manifests, not load-tested; the 413-under-backpressure mechanism is reasoned from Node socket semantics, not reproduced against a real socket.

```audit-claims round=2 audited=3784e811..3265dc8d
1. 🟡-1 terminal shed: the client now retries once on 429 honouring Retry-After, via `relayWithRetry` — extracted from the hook so it is testable, with 5 cases and 4 mutants.
2. 🟡-1 cap rationale: the constant is documented as a MEMORY bound deliberately below caller batch sizes, so shedding is expected rather than a signal to raise it.
3. 🟡-7 shed status: 429 replaces 503, so deliberate backpressure is not counted into `civitai_app_http_errors_total` by `instrumentApiResponse`.
4. 🟡-2 socket teardown: `req.destroy()` is gone; the route sets `Connection: close` and stops reading, so the 413 write is not raced by a discard-on-destroy.
5. 🟡-2 follow-on: body reading switched from `for await` to a manual `next()` loop, because `for await`'s early exit calls the iterator's `return()` which destroyed the stream anyway — measured, the destroy is now absent.
6. F6 guard moved with it: the running-total property is now asserted via `readableEnded === false` rather than via a destroy that no longer occurs.
7. 🟡-3 CSRF comment: the false "this raw route is cookie-authenticated" premise is removed; the comment now states the route accepts Bearer/`?token=`, that the cited precedent IS cookie-only so its reasoning does not transfer, and that 403ing an API-key client with no Origin is a deliberate choice with a stated reason.
8. 🟡-4 arithmetic: the cap comment now says ~800MB, not ~400MB.
9. 🟡-6 lint attribution: the test file no longer claims `local-rules/no-wholesale-module-mock` protects it, and records that the rule's five-module allowlist covers none of the mocked modules.
10. 🟡-5 PR description: rewritten; the orphan-sweep claim now states the opposite and is marked as a correction.
11. 🟢 inert latch: the `settled` mechanism in `attachUploadSettlement` is removed with a note, and the test named for it is renamed to `relayPending`, the mechanism that actually protects it.
12. 🟢 test isolation: `__resetInFlightForTest` added and called in `beforeEach`, after a held slot produced a false cascading failure during mutation.
13. 🟢 stale counts: PR body updated to 28 cases and 8 mutants.
```

ZacxDev:
## Audit round 3 — fixes landed

Round-3 delta audit read `3265dc8d30`. Fixes are on `2644652753`. Round 3 verified nine of thirteen round-2 claims fixed **by running**, and returned **four 🟡 + seven 🟢, no 🔴**. All are addressed except F4, named below.

**Its headline finding refuted round 2's headline fix, by measurement.** The auditor built a real HTTP client and drove the oversize path: `Connection: close` loses the 413 exactly as `req.destroy()` did (EPIPE vs ECONNRESET), and only `req.resume()` — the drain round 2 explicitly refused — delivers it, 5/5. That refusal conflated two costs: draining does not buffer, so memory stays flat; what it spends is wire time and the in-flight slot. Now drained, deliberately, with the trade written down and the measurement's scope (pod hop only, Traefik not included) stated in the code.

**A premise I asserted in five places was wrong.** The "dropzone batch of 10 against a cap of 8 routinely sheds two" justification contradicted the same file three lines up: the cap is per POD, the pool's replica floor is in the dozens, no session affinity. A shed is an anomaly, not routine — which inverts the action the comment prescribed. Retracted everywhere, with the correction stated rather than silently rewritten.

**Ledger** — round 3 · payload lines changed THIS round: **114 + / 46 −** (`relay.ts` 68+/29−, `upload-settlement.ts` 46+/17−) · scaffolding 111+/24− · since round 1: 624. The gate does not fire, but note the composition: **94 of 116 added payload lines are comments** (81%, up from 68% in round 2). The genuinely new *code* this round is ~22 lines — the drain, the clamp, and the abort race. That is the surface worth auditing next; the rest is prose correction.

**Verification at `2644652753`:** full unit tier 1569 files / 24877 tests, 0 failures. Cold typecheck (`tsbuildinfo` deleted) 0 errors. Prettier clean; ESLint clean except one pre-existing 2025 warning in untouched code. Five new mutants, each killed by its own tests: drain removed (1 failed/31), `Connection: close` re-added (1/31), clamp removed (1/31), clamp made a flat override (3/29), abort race removed (2/30).

**NOT verified** (unchanged): no end-to-end upload against a real backend; the browser tier exercising this hook cannot run on this host, so it was **not run** — not a claim that it passes; the DNS failure was not reproduced; the memory arithmetic is reasoned from manifests; the 413's behaviour through Traefik/Cloudflare is reasoned, not measured; and the shed-frequency retraction is itself reasoned from manifests and the absence of affinity, not measured against live traffic.

```audit-claims round=3 audited=3265dc8d30..26446527
1. F1 delivery: `Connection: close` removed and `req.resume()` added, because measurement showed the drain is the only arm that delivers the 413; buffered chunks are released before the drain runs.
2. F1 honesty: the comment now records all three attempts, the 5-run measurement table, the cost trade, and that the measurement covers the pod hop only.
3. F1 guard honesty: the test now asserts drain-called and no-Connection-header, and states in-place that it CANNOT observe delivery because it drives a `Readable.from` and an object literal.
4. F2 premise: the "batch of 10 sheds two, shedding is EXPECTED" claim is retracted in all five places (relay.ts, upload-settlement.ts, relay-with-retry.test.ts, and via this comment the PR body), replaced with per-pod + replica-floor + no-affinity reasoning and "a shed is an anomaly".
5. F3 clamp: `Math.min(requested, MAX_RETRY_AFTER_SECONDS)` with MAX = 10s, so a CDN's `Retry-After: 3600` cannot park the hook on the user's File.
6. F3 cancellability: the backoff now races the abort signal via `Promise.race`, so a cancel mid-wait is observed when it happens rather than when the timer would have expired; the already-aborted case is covered too.
7. F5: the cap comment now names `constants.mediaUpload.maxImageFileSize` as the coupled constant and explains why it is not `richTextEditor.maxFileSize`.
8. F6: both stale `503` references from round 2's own status change are gone; the two remaining `503` mentions are the deliberate "429, NOT 503" rationale.
9. F7: the type declaration no longer sits between `relayWithRetry`'s rationale JSDoc and the function, so the doc attaches to the function rather than to an unexported type.
10. F8: the CSRF comment now gives the full path and records that the `/api/v1/` sibling is bearer-only and deliberately omits the origin guard — i.e. the nearer-looking precedent argues the opposite.
11. NOT FIXED — F4: no shed counter was added. The 503->429 change removed the only per-route signal that shedding happens and nothing replaced it, so a sustained shed is now silent to both user and operator. Left deliberately for a decision on where that metric belongs; it is three lines whenever that is settled.
12. NOT FIXED — F9: the client's `defaultRetryAfterSeconds: 2` still duplicates the server's `RELAY_RETRY_AFTER_SECONDS` as a literal; nothing pins them equal.
13. NOT FIXED — F10/F11: the test seams still ship in the route module, and `relayWithRetry`'s single call site in the hook is covered by no tier that can run on this host.
```

ZacxDev:
## Audit round 4 — fixes landed. Stopping the ladder here.

Round-4 delta audit read `2644652753`. Fixes are on `f47b17e9ff`. Round 4 returned **two 🟡 + two 🟢, no 🔴**, and verified eight of thirteen round-3 claims fixed by running. All four are addressed.

### The finding, and why it is the useful one

**Round 3's `req.resume()` was a NO-OP.** `readCappedBody` drives the request's async iterator by hand, and that iterator keeps a `'readable'` listener registered; `Readable.resume()` does nothing while `readableListening` is true. The drain never ran — and **removing `Connection: close` alone** is what delivers the 413. Measured, real socket, Node 24.19.0, 5 runs per arm, 64MB body / 1MB limit; the middle two rows being identical is what settles it:

```
req.destroy()            -> ECONNRESET, no 413    5/5
Connection: close        -> EPIPE,      no 413    5/5
neither                  -> 413 + body            5/5, 3.7MB read
neither + req.resume()   -> 413 + body            5/5, 3.7MB read
neither + a REAL drain   -> 413 + body            5/5, 64MB read
```

Round 3 got the right outcome for the wrong reason and shipped a dead line plus three paragraphs justifying it — including two costs that were never paid (no wire time was spent, and the in-flight slot is released at ~2ms through the `finally`, not held across the drain). The drain is **deleted** rather than repaired: a real one pulls the entire rejected body over the wire.

**The guard that should have caught it is the reusable lesson.** Round 3's test STUBBED `req.resume`, so it could only observe that the call happened, never that it did anything — which is exactly how an inert line survives a green suite. It is now a **pass-through spy** recorded in the same trace as `destroy`, so it pins "no drain attempted" while leaving real behaviour observable. Two candidate assertions were tried and **rejected**, recorded in-place so nobody retries them: `readableFlowing` and `readableEnded` cannot pin this, because `resume()` schedules reads asynchronously and neither has moved when the handler returns — measured, a real-drain mutant passes both.

**Ledger** — round 4 · payload lines changed THIS round: **75 + / 62 −** across `relay.ts` and its test, of which the executable change is a **net deletion** · since round 1: ~700 payload. This is the first round whose fix removes more live code than it adds.

**Verification at `f47b17e9ff`:** full unit tier 1569 files / 24877 tests, 0 failures. Cold typecheck (`tsbuildinfo` deleted) 0 errors. Prettier clean, zero ESLint errors. Two mutants, both caught by the trace assertion for their own reason: a REAL drain (`removeAllListeners('readable') + resume()`), and the bare inert `resume()` round 3 actually shipped — the latter being the mutation that would have failed this guard on day one and did not exist to.

**PR description edited** (round 4's 🟢-4): it still asserted the shed-frequency premise round 3 retracted in the three in-repo sites. Round 3's claim said "retracted via this comment", which was not the same as editing the body. Now edited, marked as a correction.

### Why the ladder stops here

Four rounds, and three of them refuted the previous round's headline fix by measurement. That is the ladder working — but the composition has moved decisively: round 3 was 81% comment lines, and round 4's fix is a net deletion of live code. The remaining known items are all recorded and none is deploy-blocking.

**Carried forward, deliberately unfixed:**
- **No shed counter.** The 503→429 change removed the only per-route signal that shedding happens and nothing replaced it, so a sustained shed is silent to user and operator. ~3 lines whenever it is decided where that metric belongs.
- The client's `defaultRetryAfterSeconds: 2` duplicates the server's `RELAY_RETRY_AFTER_SECONDS` as a literal; nothing pins them equal.
- `__getInFlightForTest` / `__resetInFlightForTest` ship in the route module (not routable — Next.js only routes the default export).

### 🔴 The actual gate, which four audit rounds do not substitute for

This ticket's closing condition asks for **an upload completing end-to-end through the fallback in a preview environment**. That has not been done and cannot be done from here. Every round above verified the code against mocks, stubs and socket probes; none of it verified the feature works. The browser tier that exercises the changed hook also **cannot run on this host** — not run is not a claim that it passes.

Merging on unit evidence alone is a judgement call for whoever owns this, and it should be made knowing that is what the evidence is.

ZacxDev:
## 🔴 Preview-env verification: the relay silently stores CORRUPT images above 10MB

First end-to-end run of this branch on a deployed preview. The relay works — and it has a
blocking defect that no unit round could see, because every round drove `Readable.from`,
which never truncates. The real runtime does.

### The defect

**Next 16.3.1 truncates request bodies at 10MB, and `readCappedBody` cannot tell truncation
from end-of-body.** The stream simply ends. So the route buffers a partial body, treats it
as complete, PUTs it to the store, registers it, and returns **200 with an id**.

The server says so in its own logs, once per over-size request:

```
Request body exceeded 10MB for /api/v1/image-upload/relay. Only the first 10MB
will be available unless configured. See .../middlewareClientMaxBodySize
```

`middlewareClientMaxBodySize` is a real top-level config option
(`next/dist/server/config-shared.d.ts:1132`, schema at `config-schema.js:279`) and is
**not set** in `next.config.mjs`, so the 10MB default applies.

### Measured

Size sweep against the deployed preview, uploaded vs what actually landed in the store:

| sent | stored | verdict |
|---|---|---|
| 5,000,000 | 5,000,000 | intact |
| 9,000,000 | 9,000,000 | intact |
| 10,485,760 (10 MiB) | 10,485,760 | intact |
| 12,000,000 | 10,438,916 | **truncated** |
| 20,000,000 | 10,483,999 | **truncated** |
| 55,000,000 | 10,479,830 | **truncated** |

The cut is ragged (10,438,916 vs 10,483,999 vs 10,479,830), which is the tell that this is a
stream ending early rather than a fixed-size cap.

**With a real image** — a valid 14,523,378-byte PNG:

- relay returned `200 {"id": ...}`
- stored object: 10,485,209 bytes, **no `IEND` terminator**
- `magick <original> null:` → rc=0, clean
- `magick <stored> null:` → **rc=1, `unexpected end-of-file`**

So the user is told the upload succeeded and gets a permanently broken image.

### Consequences for this PR

1. **`MAX_RELAY_BYTES = 50MB` is unreachable.** Next cuts at 10MB first, so the constant
   never binds.
2. **The 413 path can never fire in production.** Rounds 2–4 of the audit went deep on
   delivering that status over a real socket; the measurement stands, but the branch is
   dead code at the current config — `total > limit` cannot be reached when the stream
   stops at 10MB.
3. The relay's comment ties `MAX_RELAY_BYTES` to `constants.mediaUpload.maxImageFileSize`
   (50MB), which is what the dropzones enforce client-side. Every image between 10MB and
   50MB is accepted by the UI and silently corrupted by this path.

Not caused by this PR, and **not measured by me** — flagging only as a place to look:
other routes declare larger limits than 10MB (`api/trpc/[trpc].ts` 17mb,
`api/blocks/submit-version.ts` 72mb). Whether they are affected is untested.

### Isolation

Reproduced both through the normal ingress path **and** by POSTing from inside the running
container straight at the app's own port, bypassing every proxy in front. Same truncation,
same ~10MB. So this is the framework, not anything in front of it.

### What still holds

The rest of the end-to-end verification passed, and this is the first time any of it has
been exercised against a real backend:

- guards, against the live deployment: `GET`→405, no-Origin→403, wrong-Origin→403,
  no-cookie→401, empty-body→400
- a 7,853-byte PNG relayed end-to-end: 200, object stored, **byte-identical** on retrieval
  (sha256 match), correct `image/png` content type
- `registerMediaLocation` genuinely recorded `backend: backblaze` — verified by resolving
  the key, with a never-registered uuid as a negative control (404), since that function
  swallows its own errors and a 200 from the route proves nothing about it

**Still unverified: the client half.** Nothing here drove a real browser, so the xhr
`error` → relay handoff in `useCFImageUpload` — the thing round 1 found inert — remains
untested outside unit mocks. This comment is not a claim about it.

ZacxDev:
## ✅ Fix re-verified on the rebuilt preview

`e8acaa7109` deployed and the repro re-run against it. An audit fix resets the
verification gate, so this is measured on the running build, not inferred from the unit
tier.

| body | before | after |
|---|---|---|
| 14,523,378 — the original PNG that came back corrupt | 200 + id, object undecodable | **413**, no id |
| 10,485,761 — one byte over the cap | 200 + truncated object | **413**, no id |
| 10,485,760 — exactly at the cap | 200, intact | **200**, stored **byte-identical**, `backend: backblaze` |
| 7,853 — positive control | 200, intact | **200**, intact |

🔴 **The bottom two rows are the point.** A "fix" that simply refused everything would
turn the top two green just as well, so the ones that discriminate are the at-cap upload
still storing byte-identical and the small upload still working. The boundary is measured
on both sides, one byte apart.

### The guards

Six mutants, each killed by the test whose name claims that guard — not merely "a test
failed":

| mutant | killed by |
|---|---|
| M1 drop the pre-read 413 | `refuses a declared Content-Length over the cap` |
| M2 drop the truncation guard | `rejects a body that ends short of its declared length` |
| M3 invert `total < declared` | same |
| M4 write the 400 but still store | same |
| M5 raise the cap above the truncation point | `keeps MAX_RELAY_BYTES at or below the point where Next truncates` |
| M7 keep the 413 but consume the body first | `…WITHOUT reading the body`, on the consumption count |

**M7 exists because M1 was not sufficient.** M1 fails on the *status*, so the
"without reading the body" half of that test's name was never actually exercised — M7
holds the status at 413 and changes only consumption, and dies on
`expected [ Buffer, Buffer ] to have a length of +0 but got 2`. Mutating the guard you
were thinking about is not the same as mutating the one the test's name claims.

M4 is the one worth reading: it writes the 400 and *still* stores the object. That is the
original corruption wearing a different status code, and it is why the assertion is
`expect(mockUploadImageBufferToStore).not.toHaveBeenCalled()` rather than a status check.

### Why the cap moved rather than the framework limit

Raising `middlewareClientMaxBodySize` was measured before being rejected, not assumed
away. Sampled 111,097 rows of `Image.metadata->>'size'`:

- **0.679%** of images exceed 10MB (p99 = 7.63MB); exactly one in the sample exceeded 50MB
- video is a different population — **17.1%** over 10MB — so if this route is ever put on
  a video path, the cap and the config knob need revisiting together

Widening a repo-wide body limit on a public, high-traffic app to serve 0.68% of images, on
a path that only fires for clients who cannot resolve the storage host at all, is a poor
trade. Those few now get an honest 413 instead of a silently corrupted file. The test
pinning `MAX_RELAY_BYTES <= NEXT_BODY_TRUNCATION_BYTES` fails if someone raises one without
the other.

### Still not verified

**The client half.** No browser has driven the xhr `error` → relay handoff. The direct PUT
succeeds from a normal browser, so the fallback does not fire naturally — it has to be
forced (CDP `Fetch.failRequest` with `NameNotResolved` on the storage host). Playwright
does not run on the host this was done from. So: not run, which is not a claim that it
passes. This PR should stay draft until that is exercised.

ZacxDev:
## ✅ Client half verified in a real browser — the last outstanding gap

Driven in a real Chromium against the deployed preview, with the file going through the
actual dropzone-equivalent page (`/intent/avatar`, which calls `uploadToCF` and then feeds
the resolved id straight into a `user.update` mutation — so the id is *observable*, not
inferred).

The fallback cannot be triggered by waiting: from a working browser the direct PUT
succeeds, so the failure was injected at the real seam — `XMLHttpRequest.prototype.open`
patched in the page's own world to rewrite the storage-host PUT to an unresolvable
hostname. That produces a genuine network-layer `error` event, which is the exact
production condition (`ERR_NAME_NOT_RESOLVED`), not a simulated one.

**Two arms, differing in exactly one variable:**

| arm | direct PUT | relay POST | id handed downstream |
|---|---|---|---|
| storage host unresolvable | rewritten → `error` | **fired, 200** | `47742d02…` — the relay's **server-minted** key |
| untampered (control) | **200 OK** | **did not fire** | `6d7e8fc9…` — the **presign** id |

### What each row establishes

1. **The fallback fires, and settles correctly.** This is precisely the path round 1 found
   INERT — where `loadend` followed `error` in the same dispatch and resolved with the
   presign key, turning a visible failure into a silent success with a bad id. It now
   relays and resolves under the relayed key.
2. **`resolvedId` is real.** The two arms hand *different* ids downstream. The relayed key
   is not the presign id — the mechanism the whole `resolvedId` change exists for.
3. 🔴 **The relay is genuinely FALLBACK-ONLY.** In the control the relay was never called.
   That is the code's own load-bearing requirement ("do NOT wire this in as the default
   upload path"), and it had never been demonstrated — only asserted.
4. **The bytes really went through the relay.** In arm 1 the direct PUT never reached
   storage, yet `47742d02…` resolves with `backend: backblaze` and serves the source image
   **byte-identical** (117,650 bytes). The only path those bytes could have taken is the
   server-side relay write.

### Scope, honestly

- The DNS failure is injected by patching XHR in-page, not by breaking real resolution.
  What is faithful is the *event the hook sees*; a genuinely unresolvable host and a
  rewritten-to-unresolvable host both deliver a network-layer `error`.
- Both arms ran against the deployed preview build with the truncation fix.
- No avatar was persisted: the `user.update` mutation was captured and rejected in both
  arms, so the payload above is what *would* have been written.

With this, both halves of the change are verified against a real runtime: the server half
(relay → store → registration, plus the >10MB truncation fix and its boundary) and now the
client half.

PR #4577, opened 2026-09-02
fix(app-chrome): site-consistent links, full-bleed chrome, real icons in Recently run

ZacxDev:
Three pieces of feedback on the App Blocks page chrome, one commit each so any of them can be reverted alone.

## 1. The breadcrumb link uses the site's link style

The "Marketplace" crumb was a hand-styled `Text` carrying `c="blue.6"` and `td="underline"` — bluer and permanently underlined in a way nothing else on the site is. It is now a Mantine `Anchor`.

The `blue.6` was not arbitrary — an audit found the original `blue.4` borderline against the near-white light chrome surface — so the swap had to preserve it, and it does:

| scheme | before | after |
|---|---|---|
| light | blue-6 (hard-coded) | blue-6 (`--mantine-color-anchor` → primary-filled → blue-filled) |
| dark | blue-6 (hard-coded) | **blue-4** |

Neutral on light, better on dark: the old fixed shade stayed dark on a dark surface because it was only ever reasoned about against the light background.

**The resting underline is gone**, deliberately — `Anchor` defaults to `underline="hover"`. The crumb still reads as a link against its dimmed neighbours by colour. If a future audit wants a resting underline for WCAG 1.4.1, that belongs on the shared `Anchor`, not re-forked here.

## 2. The chrome spans the page; the app is capped instead

The ultrawide cap sat on the host root, so the chrome was capped along with the app. Stopping site furniture at 1600px made a full-page app read as a boxed widget dropped into the page rather than a page of the site. The cap now lives on a new content wrapper holding the app and the failure card, but not the chrome.

**This reverses a deliberate earlier decision and the note it replaces is worth reading.** The old placement argued that a breadcrumb vouching for the app should not span a width the app does not occupy. That cost is real and is now paid: on a very wide display the chrome is wider than the app it labels. It is the same relationship the site header already has to every page's content column, and the trade was judged worth it — recorded on the root so nobody restores the old placement as a tidy-up.

Everything else about the cap is unchanged: same value, same `var()` read, same fallback, same opt-out ledger. `--app-page-max-width` is still declared once in `globals.css` and still overridden per-app **on the frame**, from which it inherits to the content wrapper — so ledger rules keyed on `[data-testid='app-page-frame'][data-block-id='…']` keep working with no selector change. The `playable-collections` ledger test exercises exactly that inheritance.

The wrapper reproduces the vertical flex chain it was inserted into, which is what keeps this width-only; `PageBlockHostScrollFit`'s exact-equality scroll contract is green.

## 3. The real app icon in "Recently run"

The list already rendered an icon when the entry had one. The gap was upstream: **the run page never recorded one**, because nothing on its SSR path had ever read `app_listings` for media. Every other writer goes through `toRecentAppFromListing` and carries an icon — so apps merely *opened from the store* showed their real icon, and the apps a viewer actually *runs* showed the placeholder.

It is the **listing's** icon, not the manifest's, and that is a trust choice: this chrome is the spoof-proof surface and already launders the app name through `sanitizeAppChromeName`, so a publisher-supplied image with no review step has no business sitting beside a name we deliberately sanitize. The listing icon is moderator-approved and goes through the same shared projection the store uses, so the two surfaces cannot disagree.

Two properties that matter on the app-launch path:

- **Concurrent, not serial.** It joins the existing `Promise.all`, keyed on the slug already in hand, so the page waits for the slowest of three reads rather than their sum. A test pins this, because re-keying it on `page.appBlockId` would compile, pass everything else, and quietly add a round trip to every app launch.
- **Fails open to `null`.** `createServerSideProps` has no try/catch above it, so a rejection here would be a 500 on the page that runs the app. Logged under its own event name so an `app_listings` outage stays findable instead of presenting only as icons that quietly stopped appearing.

It is a separate reader rather than an extra column on the beta read that already runs here, even though that would be one fewer query: that reader collapses to `BETA_UNAVAILABLE` on a missing-column error, so sharing it would make the icon vanish during the manual-apply window for any future beta-adjacent column — a failure with no relationship to the icon's own columns.

## Testing

Every new guard was watched fail on pre-change code:

| claim | red arm |
|---|---|
| crumb is an `Anchor`, no local override | reverted the crumb → node guard fails on the missing `Anchor` |
| chrome spans, app capped | swapped the cap back to the frame → `frame is 1600px inside a 2560px parent`; the other 10 tests in the file passed in the same run |
| run page resolves an icon | removed the read → 3 of 9 fail, incl. the positive control `the icon read never ran` |

- Full unit suite: **1,569 files / 24,877 passed**
- Full `AppBlocks` browser suite: **40 files / 484 passed**
- `typecheck` 0 errors · `test:lint-rules` 448 passed · eslint: 0 errors from changed files

Two notes on the test work itself:

- The two browser assertions that pinned an inline colour and a resting underline **could not** be re-pointed at the computed colour: that harness injects only the `:root` custom properties from `globals.css` and never loads `@mantine/core/styles.css`, so a computed read there compares two unresolved values and passes while observing nothing. The claim was split — the browser tier asserts the crumb renders as an `Anchor` at the library default with no local override, and a new **node-tier** guard resolves the anchor colour chain against the shipped Mantine stylesheet (also the gating tier, unlike the report-only browser project). A Mantine upgrade that re-maps the anchor colour now fails there with an explanation.
- Reverting the crumb also made the pre-existing breadcrumb-copy guard fail while blaming the **copy** — it scanned forward to a literal `</Text>` and ran on to the next one once that tag moved. It now names its closing tag and says why.

Review tip: `PageBlockHost.tsx` is 250+/203− but only 74+/27− ignoring whitespace — the rest is re-indentation from wrapping the conditional. `?w=1` helps.

https://claude.ai/code/session_01YVStb9vhnKD2oybCWdJGUB

### DIFF
--- a/src/components/AppBlocks/AppBlockChrome.browser.test.tsx
+++ b/src/components/AppBlocks/AppBlockChrome.browser.test.tsx
@@ -244,14 +244,24 @@ describe('AppBlockChrome run-page breadcrumb (Marketplace / <app name>)', () => 
     // `__tests__/chromeCrumbLinkStyle.test.ts`, which reads the shipped Mantine
     // stylesheet directly.
     //
-    // What IS observable here is that the crumb is a real Mantine `Anchor` sitting at
-    // the library default rather than a locally restyled `Text`. `Anchor` renders its
-    // `underline` prop as `data-underline`, and its default is `hover`.
+    // What IS observable here is that the crumb is a real Mantine `Anchor` and which
+    // `underline` mode it is in — `Anchor` renders that prop as `data-underline`.
+    //
+    // 🔴 `always`, NOT THE LIBRARY DEFAULT `hover`, AND THE DIFFERENCE IS AN ACCESSIBILITY
+    // DECISION RATHER THAN A STYLE PREFERENCE. At rest this crumb sits between two dimmed
+    // `/` separators and a dimmed app-name crumb, so with a hover-only underline the sole
+    // resting differentiator would be hue — measured 1.07:1 on light, 1.29:1 on dark, where
+    // WCAG 1.4.1 (failure F73) allows colour alone only above 3:1, and Mantine emits no
+    // `:focus-visible` underline to fall back on. Reverting this to `hover` looks like
+    // "adopting the library default" and is a Level-A regression; that is exactly why it is
+    // asserted rather than left to the default.
     expect(
       appsLink.getAttribute('data-underline'),
-      'the crumb is not rendering as a Mantine `Anchor` at its default `underline="hover"`. ' +
-        'If it went back to a hand-styled `Text`, the chrome has re-forked the site link idiom.'
-    ).toBe('hover');
+      'the crumb is not a Mantine `Anchor` with `underline="always"`. Either it went back ' +
+        'to a hand-styled `Text` (re-forking the site link idiom), or it was relaxed to the ' +
+        'library default `hover` — which removes the only resting cue distinguishing it from ' +
+        'its dimmed neighbours at 1.07:1 contrast. See the note above before changing this.'
+    ).toBe('always');
 
     // …and it carries NO local colour or decoration override — the mutation that would
     // re-introduce the fork. The old implementation set both (`c="blue.6"`,
--- a/src/components/AppBlocks/AppBlockChrome.browser.test.tsx
+++ b/src/components/AppBlocks/AppBlockChrome.browser.test.tsx
@@ -127,12 +127,14 @@ describe('AppBlockChrome "Hide" item is surface-aware (page vs model)', () => {
 
   test('model surface (model.sidebar_top) renders the "Hide app block" item', async () => {
     renderWithProviders(
-      <AppBlockChrome blockInstanceId="inst-model" appName="Background Remover" slotId="model.sidebar_top" />
+      <AppBlockChrome
+        blockInstanceId="inst-model"
+        appName="Background Remover"
+        slotId="model.sidebar_top"
+      />
     );
     await openMenu();
-    await expect
-      .element(page.getByRole('menuitem', { name: 'Hide app' }))
-      .toBeInTheDocument();
+    await expect.element(page.getByRole('menuitem', { name: 'Hide app' })).toBeInTheDocument();
   });
 
   test('no slotId (back-compat default = model surface) renders the "Hide app block" item', async () => {
@@ -140,9 +142,7 @@ describe('AppBlockChrome "Hide" item is surface-aware (page vs model)', () => {
       <AppBlockChrome blockInstanceId="inst-default" appName="Background Remover" />
     );
     await openMenu();
-    await expect
-      .element(page.getByRole('menuitem', { name: 'Hide app' }))
-      .toBeInTheDocument();
+    await expect.element(page.getByRole('menuitem', { name: 'Hide app' })).toBeInTheDocument();
   });
 
   test('page surface (app.page) does NOT render the "Hide app block" item, keeps "Manage apps"', async () => {
@@ -153,9 +153,7 @@ describe('AppBlockChrome "Hide" item is surface-aware (page vs model)', () => {
     // "Manage apps" stays …
     await expect.element(page.getByRole('menuitem', { name: 'Manage apps' })).toBeInTheDocument();
     // … but "Hide app block" is suppressed on the full-page surface.
-    await expect
-      .element(page.getByRole('menuitem', { name: 'Hide app' }))
-      .not.toBeInTheDocument();
+    await expect.element(page.getByRole('menuitem', { name: 'Hide app' })).not.toBeInTheDocument();
   });
 });
 
@@ -181,7 +179,11 @@ describe('AppBlockChrome "Hide" item is surface-aware (page vs model)', () => {
 describe('AppBlockChrome run-page breadcrumb (Marketplace / <app name>)', () => {
   test('page surface (app.page) renders the breadcrumb with the app name + a "Marketplace" link to /apps', async () => {
     renderWithProviders(
-      <AppBlockChrome blockInstanceId="inst-bc-page" appName="Budgeted Generator" slotId="app.page" />
+      <AppBlockChrome
+        blockInstanceId="inst-bc-page"
+        appName="Budgeted Generator"
+        slotId="app.page"
+      />
     );
     // The breadcrumb container is present on the page surface.
     await expect.element(page.getByTestId('app-block-breadcrumb')).toBeInTheDocument();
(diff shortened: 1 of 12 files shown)


ZacxDev:
## Adversarial audit — round 1 (first, full)

Audited `d596565541..9f89fd36d0`. **No 🔴.** The shipped code was correct on every executable path; the six findings were about the reasoning and coverage around it. Fixed in `5761f97050`.

Two are worth calling out to reviewers rather than leaving in the commit log:

**The resting underline is back, via `underline="always"`.** The original PR dropped it to adopt `Anchor`'s default. The audit measured what that costs: the crumb's neighbours are dimmed, so at rest hue became the sole differentiator at **1.07:1** on light / 1.29:1 on dark — WCAG 1.4.1 failure F73, which permits colour alone only above 3:1, and Mantine emits its underline for `:hover`/`:active` with no `:focus-visible` fallback. `underline="always"` is not a re-fork: it is `Anchor`'s own prop, used by five other call sites here for the same reason. The change this PR is actually making — a themed colour instead of a hand-rolled one — is untouched.

**The "blue.6 clears WCAG AA on the light chrome" claim is false, and this PR had restated it while deleting the only test that asserted it.** Measured against this bar's own background (gray-0 `#f8f9fa`), blue-6 is **3.37:1** vs AA's 4.5:1 for 12px text. The earlier audit that bumped it from `blue.4` improved it without reaching AA. **That shortfall pre-dates this PR and is untouched by it** — no regression, and no fix either; what changed is that three places no longer assert something false. Correspondingly the dark-side gain was understated: **3.82:1 → 5.49:1, i.e. failing AA → passing.**

The other four: the layout payload had no gating-tier coverage (two mutants — reverting the change, and dropping `flex: 1` — were green across 24,877 node tests, and the second rendered the app at 150px instead of ~860px with *both* tiers green); `globals.css` still documented the cap as living on the host root; the `minHeight: 0` comment credited a test that does not exercise it; and the colour guard's headline was wider than its implementation.

One of the new guards had to be fixed after being watched fail: on the first attempt the revert-mutant died to the *box-model* pin while the guard written for it passed — its region started at the frame's `data-testid`, and JSX puts `style={{…}}` before it, so a cap declared on the frame was invisible to it. Green for the wrong reason, on the one mutation it exists to catch.

After fixes: typecheck 0 errors · unit 1569 files / 24,879 passed · lint-rules 448 passed · AppBlocks browser 40 files / 484 passed.

```audit-claims round=1 audited=9f89fd36d0..5761f970
1. Restored a resting link cue with `underline="always"` on the breadcrumb Anchor; asserted in both the node source guard and the browser attribute test.
2. Removed the false "blue.6 clears WCAG AA on the light chrome surface" claim from IframeHost.tsx and chromeCrumbLinkStyle.test.ts, and recorded the measured 3.37:1 as pre-existing and untouched.
3. Added a gating-tier guard that the cap sits on `app-page-content` and that box renders inside `app-page-frame`; watched it fail on the cap-moved-back-to-frame mutant, for its own reason.
4. Added a gating-tier verbatim pin of the content wrapper's box model; watched it fail on the dropped-`flex: 1` mutant.
5. Corrected globals.css to say the cap is read on `app-page-content` while the ledger still keys on the frame and reaches it by inheritance.
6. Corrected the `minHeight: 0` comment to state it is defence and is NOT covered by PageBlockHostScrollFit (measured 20/20 green without it).
7. Widened chromeCrumbLinkStyle to assert ThemeProvider sets no primaryColor/primaryShade and that blue[6]/blue[4] hold the library values, read by index and case-insensitively.
8. Replaced two false claims in the recents-write comment with the real residual: recordRecentlyOpenedApp replaces wholesale with no icon ratchet, so a degraded icon read drops a previously-recorded icon until the next success.
```

ZacxDev:
## Adversarial audit — round 2 (delta)

Audited `9f89fd36d0..5761f97050`. **No production defect in the delta** — the single shipped line (`underline="always"`) is correct, and every contrast figure quoted in the comments reproduced under independent measurement (3.374 / 1.071 / 1.291 / 3.819 / 5.489, and `size="xs"` really is 12px so 4.5:1 is the right threshold). Six findings, **all scaffolding**. Fixed in `7722d1904e` + `68415482db`.

**The one that mattered.** Round 1 added a guard asserting `app-page-content` renders *inside* `app-page-frame` — the relationship the full-bleed ledger depends on, since the custom property is set on the frame and inherited. It asserted that as `contentAt > frameAt`: a comparison of **text offsets**, which every sibling, cousin and unrelated later element also satisfies. Measured: making the two genuine siblings — killing the ledger outright — left that file 9/9 green and the **entire gating suite (1,569 files / 24,879 tests) byte-identically green**, while the report-only browser tier correctly failed both ledger tests. That is exactly the tier split the guard's own header says it exists to close.

A second guard had the same disease: `lastIndexOf('<', frameAt)` finds the nearest preceding `<`, which is the opening tag only while nothing in the element's props contains one. An ordinary prop holding a `<` re-opened the hole round 1 had just closed.

Both were text heuristics for a question text cannot answer. **Offsets cannot express "inside"; a tree can.** Both now read a real TypeScript parse of `PageBlockHost.tsx` — find the elements by `data-testid`, walk parent links for containment, read each element's own `style` attribute instead of slicing between two file positions. Four mutants re-run against the rewritten guards, each now dying **for its own reason with its own message**: content-as-sibling, cap-back-on-frame-with-a-`<`-in-props, dropped `flex: 1`, dropped `minHeight: 0`.

Third: the box-model pin had `flex: 1` inside its *anchor regex*, so deleting that property failed with "the anchor rotted — update the pin deliberately rather than deleting it". The explanation of what `flex: 1` does was unreachable for the one mutation it was written for, and the advice the developer actually saw told them to edit the pin — after which the sliver ships. Deleting `minHeight: 0` (which the component says is *not* load-bearing) failed *with* the `flex: 1` message. Both mutants died, both for the wrong reason.

The remaining three are all one defect class — a claim that outlived the code beneath it, which is the class this ladder keeps finding in its own output: the false "blue.6 clears AA" line survived in an assertion **message** (the text someone reads while deciding whether a Mantine upgrade is safe); two comments still said the crumb underlines "on hover" and that the resting underline "had to go", thirty lines above an assertion requiring the opposite; and an evidence docblock cited "1717 tests across 124 files" for a tier that is 40 files / 484 tests.

After fixes: typecheck 0 errors · unit 1569 files / 24,879 passed · lint-rules 448 passed · AppBlocks browser 40 files / 484 passed.

**Ledger.** `round 2 · payload lines changed THIS round: 124, of which executable: 1 · round 3 · executable payload lines changed: 0`. Round 2's fixes are entirely scaffolding plus one comment rewrap in `globals.css` — verified by reading the diff, not by a line count (a naive non-comment filter reported 7; all seven were comment prose). That is the **first** zero-payload round. The attribution gate fires on two consecutive, so round 3 runs — and it is warranted on its merits: round 2's fix added ~139 lines of new AST-walking guard code that is now the only thing standing between the sibling mutation and production, and a fix round introducing the next defect is the failure mode this ladder exists to catch.

```audit-claims round=2 audited=5761f97050..68415482
1. Replaced the `contentAt > frameAt` offset comparison with a TypeScript-AST descendant check; watched a sibling mutant fail it for its own reason.
2. Replaced the `lastIndexOf('<', …)` frame-region heuristic with a read of the frame element's own `style` attribute; watched a cap-on-frame mutant with a `<` in its props fail it.
3. Removed `flex: 1` from the box-model pin's anchor so the pin's own message is what a developer sees; watched both the dropped-`flex: 1` and dropped-`minHeight: 0` mutants fail with the intended message.
4. Added a lookup-failure assertion so a missing `app-page-frame`/`app-page-content` fails on the lookup rather than making every later assertion vacuous.
5. Removed the surviving "inherited its WCAG AA contrast" claim from chromeCrumbLinkStyle.test.ts's assertion message and replaced it with the measured 3.37:1.
6. Corrected two comments in AppBlockChrome.browser.test.tsx that still described underline-on-hover and said the resting underline "had to go".
7. Corrected the "1717 tests across 124 files" evidence count to the measured 40 files / 484 tests.
8. Rewrapped the globals.css cap comment twice — round 2 fixed round 1's mangled wrap and introduced a 100-char line, fixed in 68415482db.
```

ZacxDev:
## Adversarial audit — round 3 (delta) — **ladder stopped here**

Audited `5761f97050..68415482db`. Round 3 found a real 🔴 **in round 2's own fix** — the failure mode this ladder exists to catch. Fixed in `7242dfdd57`.

**The 🔴.** Round 2 replaced two fragile text guards with an AST helper. That helper's `styleTextOf()` returned `''` for any style that is not an inline object literal — a spread, an identifier, a style lifted into another module — and the frame-side assertion is `.not.toContain('--app-page-max-width')`, which an empty string satisfies trivially. Measured, and reproduced here: **extracting the frame's inline style into a module const and putting the cap back into it, with the exact pinned spelling, left the file 9/9 green while the app chrome was capped again.** That is the one regression this file exists to block, reachable by an entirely ordinary refactor. A spread did the same and killed the mutant only through an unrelated pin — a died-for-the-wrong-reason kill, which is what round 2 had just finished fixing elsewhere.

`null` and `''` are now different answers: a style this guard cannot read is a reason to **fail**, never to pass.

Two more in the same helper: `isDescendant` returned **true for a sibling** when the ancestor was self-closing (for a `JsxOpeningElement`, `.parent` is its own `JsxElement`, which is what makes the walk mean "inside"; for a self-closing one it is the *enclosing* element), and `hostElements` let a duplicate testid silently overwrite — losing the ambiguity check the text helper it replaced had. Plus one of three occurrences of the node-suite count still read 24,877.

**Five mutants re-run, each now failing for its own reason with its own message**: extracted-style-plus-cap-on-frame, cap-on-frame-via-spread, frame-self-closing-with-content-as-sibling, content-lifted-to-a-sibling, duplicate testids.

I also closed round 3's one *could-not-verify*: it could not run the browser tier under the sibling mutation (the `component` project won't resolve workspace deps in a `cp -a` sandbox). Measured directly instead — sibling mutation applied, browser tier run, restored: **exactly the two ledger tests fail** (`an app can opt out of the cap with a CSS rule keyed on data-block-id`, `LEDGER — playable-collections is full-bleed…`). The comment asserting that is now measured rather than claimed.

### Ledger — and why this is the last round

```
round 3 · payload lines changed THIS round: 0 (round 2's fixes: 0) · ladder STOPPED
```

- round 2's fixes (`5761f97050..68415482db`): three test files + two `globals.css` hunks that fall **entirely inside one `/* … */` block** — no declaration, selector or value moved. **0 executable payload lines.**
- round 3's fixes (`68415482db..7242dfdd57`): one test file. **0 payload lines.**

Two consecutive zero-payload rounds, so per the attribution rule **the ladder has left the PR and stops here** — it is now auditing scaffolding it wrote itself, which is the shape the rule exists to catch. Every finding from all three rounds is fixed, so there is nothing to carry forward as a follow-up: **no open items, nothing filed.**

Worth stating plainly, because three rounds of dense findings can read as a troubled change: **no round found a production defect after round 1.** The shipped behaviour of this PR has been unchanged since `5761f97050`, and the only executable line any audit round added is `underline="always"`. Rounds 2 and 3 were entirely about whether the guards guard — and round 3 shows that mattered, since round 2's replacement guard was blind to the very mutation it was written for.

Final state: typecheck 0 errors · unit **1569 files / 24,879 passed** · lint-rules 448 passed · AppBlocks browser tier **40 files / 484 passed** · 0 eslint errors from changed paths.

PR #4579, opened 2026-09-02
fix(jobs): fence remove-old-drafts on version and file activity

ZacxDev:
## The defect

`remove-old-drafts` (nightly, `43 2 * * *`) selects every `Draft`/`Deleted` model where

```sql
m."updatedAt" < now() - INTERVAL '30 days' AND mm."downloadCount" < 10
```

and runs `DELETE FROM "Model"`, which cascades to every `ModelVersion` and `ModelFile` under it. There is no soft-delete and no recovery.

`Model.updatedAt` is a Prisma `@updatedAt` column, so it moves **only** when a client writes the `Model` row. None of the things that actually happen to a draft write that row:

| what happens | writes `Model`? |
|---|---|
| `upsertModelVersion` | no — `ModelVersion` only |
| training completion (`training.service.ts`) | no — `ModelVersion` only |
| file upload | no |
| `updateModelLastVersionAt` | returns early unless a **Published** version exists, so never for a draft |
| trainer UI (`TrainingBasicInfo`) | skips the model write when the form is not dirty |

That last one makes the miss the *default* rather than an edge case. So on a draft the clock freezes at "creator last edited the model's metadata", while the finished LoRA arrives much later as a `ModelFile` — and the job then reads a model that is actively being worked on as thirty days abandoned.

### Exhibit

Model **2831418**: `createdAt == updatedAt`, never touched again. Version **3194997** updated much later, `trainingStatus = Approved`. A 272 MB `Model` file created hours before the reap ran, plus a 113 MB `Training Data` file with `dataPurged = false`. It matched the predicate exactly.

### Blast radius (measured twice on a real candidate set)

- 1,103 models / 691 users selected per night
- **139** had version-or-file activity inside the 30-day window
- **138** of those still held a training dataset
- ~65 by the stricter "a file was created in the last 30 days" test

### 1. The activity fence, in the replica SELECT

```sql
AND m."availability" != 'Private'::"Availability"
AND NOT EXISTS (SELECT 1 FROM "ModelVersion" mv
                 WHERE mv."modelId" = m.id
                   AND (mv."createdAt" > now() - INTERVAL '30 days'
                     OR mv."updatedAt" > now() - INTERVAL '30 days'))
AND NOT EXISTS (SELECT 1 FROM "ModelVersion" mv2
                 JOIN "ModelFile" mf ON mf."modelVersionId" = mv2.id
                 WHERE mv2."modelId" = m.id
                   AND mf."createdAt" > now() - INTERVAL '30 days')
```

The `Private` exclusion is spelled the way `reset-to-draft-without-requirements.ts` already spells it, and `Availability` is confirmed against `schema.full.prisma`. Both `NOT EXISTS` resolve by index (`ModelVersion(modelId)` hash, `ModelFile(modelVersionId)` hash); measured cost on the real candidate set was **345 ms → 327 ms**.

Two fixes that look obvious and **do not work**, both refuted on the exhibit, whose version was created in the same minute as the model:

- fencing on the newest `ModelVersion."createdAt"` alone
- bumping `Model."updatedAt"` on version create

Only a fence that reads `ModelFile."createdAt"` or `ModelVersion."updatedAt"` catches it.

### 2. Log what was destroyed

The job previously logged counts only, which is why a user's loss report was unreconstructable. It now emits one Axiom event **per batch** carrying that batch's `modelIds` and `userIds` — bounded by `BATCH_SIZE`, so at most 10 ids per event rather than one field holding 1,100. Skipped models and failed batches are named the same way. `userId` is picked up from the SELECT that was already reading `Model`.

### 3. `filterModelsWithRecentActivity` — implemented, and why

The brief left this as a judgement call. **Implemented.** Three reasons:

1. **It is the only behavioural coverage that exists.** There is no DB-backed vitest project here (no `pg-mem`, no testcontainers), so the SQL cannot be executed in a test. Without a TypeScript copy of the rule, the entire fence is covered by a *spelled* string assertion and nothing more.
2. **It closes a hole the SQL cannot.** The SELECT runs against the **read replica**, minutes before the delete. It cannot see a version or file written during replica lag, or in the window between the SELECT and the DELETE. The helper re-checks the same rule on the **primary**, on rows read inside the batch loop, immediately before the `DELETE`. When the two fences disagree, that is logged as a warning — which is also the signal that this hole is real.
3. **The stated cost does not materialise.** The brief priced this at "an extra query per batch". It needs none: the loop already runs a per-batch `ModelVersion` lookup to collect version ids before the cascade, and that query now returns the timestamps too (plus a `max(ModelFile."createdAt")` correlated subquery, index-resolved).

It fails **closed** in both directions the row shape can go wrong: an unreadable required timestamp protects its model, and a row with no usable `modelId` cannot be attributed to anything, so it refuses the whole batch rather than silently protecting none. Sparing a model costs one night; deleting one cannot be undone.

Related correctness fix that falls out of this: only the versions of models **actually deleted** are handed to `deregisterFileLocationsBatch`. Deregistering a spared model's `file_locations` would drop its objects out of the dereference-quarantine allowlist — the same data loss by another route.

## Verification

**Red → green matrix**, `remove-old-drafts.test.ts` held at this branch with the job file reverted to `origin/main`:

| | `origin/main` job | HEAD |
|---|---|---|
| `remove-old-drafts.test.ts` | **12 failed / 6 passed (18)** | 18 passed |
| `remove-old-drafts-activity-filter.test.ts` | n/a — the helper is new | 13 passed |

The three data-loss regressions each failed at base with their own named assertion:

```
a version created inside the window must spare its model: expected [ [ 42 ] ] to deeply equal []
a version updated inside the window must spare its model: expected [ [ 42 ] ] to deeply equal []
a file created inside the window must spare its model even when the version is ancient:
  expected [ [ 2831418 ] ] to deeply equal []
```

No import abort — the file collected all 18 tests at base and every failure was a real assertion.

**Positive controls.** Both files open with a case asserting the baseline fixtures produce a **non-empty** delete list, so "0 models deleted" can never satisfy the `spares…` assertions vacuously. Both pass at `origin/main` *and* at HEAD, which is what a positive control should do. Mutant H7 (below) exists to prove they can fail.

**Mutation sweep — 14 mutants, each isolating one clause, all killed, none unapplied.** Every mutant asserted its own source-text change before running (a mutant that fails to apply otherwise "survives" for free).

| # | mutation | killed by |
|---|---|---|
| H1 | helper `created > cutoff` → `<` | `clause: mv."createdAt" inside the window` |
| H2 | helper `updated > cutoff` → `<` | `clause: mv."updatedAt" inside the window` |
| H3 | helper `fileAt > cutoff` → `<` | `clause: mf."createdAt" inside the window` (6 failures, no positive-control collateral) |
| H4 | `ACTIVITY_WINDOW_DAYS` 30 → 300 | seam guard: `expected [INTERVAL '30 days'…] to deeply equal [INTERVAL '300 days'…]` |
| H5 | drop the required-timestamp fail-closed | `clause: unreadable required timestamp must fail CLOSED` — **kills exactly 1 test** |
| H6 | drop the unattributable-row fail-closed | `clause: unattributable activity must fail CLOSED across the batch` — kills exactly 2 |
| H7 | treat a null `latestFileAt` as activity | the quiet-baseline positive control |
| S1–S3 | SQL `>` → `<` in each of the three date clauses | the predicate pin |
| S4 | SQL interval 30 → 300 in the version clause | predicate pin + seam guard |
| S5 | drop the `Private` exclusion | the predicate pin |
| S6 | drop the whole `ModelFile` `NOT EXISTS` clause | predicate pin + seam guard |
| S7 | drop `userId` from the select list | `selects the owning userId alongside the model id` — kills exactly 1 |

Fixtures are built at **5 days** and **200 days**, never at 30, so no clause is reached or missed by landing on its own boundary, and a window widened to 300 days has a 200-day fixture to be caught by.

## Two honest limitations

**The predicate pin is a spelled guard, and it is labelled as one in the code.** An equivalent reword — swapping the two `NOT EXISTS`, aliasing the tables differently, writing `now() - '30 days'::interval` — fails it while changing nothing about which models get destroyed. It proves the predicate has not silently changed; it proves nothing about what the SQL *does*. That limitation is why mutants **S1–S6 die only to a string comparison**: S6 in particular deletes a real data-loss guard and no behavioural test can see it. Closing that properly needs a DB-backed test project, which does not exist in this repo.

**H1, H2, H4 and H7 also kill the positive control**, so they die for more than one reason. H3, H5, H6 and S7 are cleanly isolated — they fail only their own named assertion. `ACTIVITY_WINDOW_DAYS` is deliberately one constant in one place, so its mutation is necessarily global and cannot be isolated further.

## Checks run

- `pnpm typecheck` — **0 errors**. (Instrument validated: it reported 10 errors minutes earlier, from an unpopulated `event-engine-common` submodule in the fresh worktree, and went to 0 once initialised.)
- `pnpm run test:unit:run src/server/jobs/__tests__` — **512 passed / 47 files**
- `pnpm run test:lint-rules` — **448 passed / 28 files**
- `eslint` on the three changed files — clean
- `prettier --check` — clean (it flagged all three before `--write`, so the check is live)

No SQL was run against production for this change; the cost and blast-radius numbers are from the prior investigation.

## Reviewer questions

1. **Should `Deleted` models be fenced too, or only `Draft`?** The fence currently applies to both. A `Deleted` model with a recent file upload is a stranger case and arguably should still be reaped.
2. **Is refusing the whole batch on an unattributable row too blunt?** It costs one night for the batch. The alternative (protect nothing) is how the fence would fail open if the query shape ever drifts.
3. **The `Private` exclusion narrows the reaper permanently.** Private drafts will now never be collected by this job. If they should be reaped on some other schedule, that is a separate change.

### DIFF
--- a/src/server/jobs/__tests__/remove-old-drafts-activity-filter.test.ts
+++ b/src/server/jobs/__tests__/remove-old-drafts-activity-filter.test.ts
@@ -1,0 +1,153 @@
+import { describe, it, expect } from 'vitest';
+
+import {
+  filterModelsWithRecentActivity,
+  ACTIVITY_WINDOW_DAYS,
+  type ModelVersionActivityRow,
+} from '~/server/jobs/remove-old-drafts';
+
+/**
+ * Behavioural coverage of the activity fence.
+ *
+ * The rule also exists as SQL in the job's replica SELECT, and that copy cannot
+ * be exercised here — this repo has no DB-backed vitest project, so there is no
+ * engine to run `NOT EXISTS` against. `remove-old-drafts.test.ts` pins the SQL's
+ * text; these tests pin what the rule MEANS, on the copy that runs immediately
+ * before the delete.
+ */
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+const NOW = new Date('2026-06-01T00:00:00.000Z');
+const daysBefore = (n: number) => new Date(NOW.getTime() - n * DAY_MS);
+
+// Both fixture ages are far off the 30-day boundary, on opposite sides: nothing
+// sits at the cutoff, so no clause can be skipped by landing exactly on it, and
+// a window widened to 300 days has a 200-day fixture to be caught by.
+const QUIET_DAYS = 200;
+const RECENT_DAYS = 5;
+const quiet = daysBefore(QUIET_DAYS);
+const recent = daysBefore(RECENT_DAYS);
+
+function row(over: Partial<ModelVersionActivityRow> = {}): ModelVersionActivityRow {
+  return { id: 100, modelId: 1, createdAt: quiet, updatedAt: quiet, latestFileAt: null, ...over };
+}
+
+const run = (rows: ModelVersionActivityRow[], candidates = [1]) =>
+  filterModelsWithRecentActivity(candidates, rows, NOW);
+
+describe('filterModelsWithRecentActivity', () => {
+  // POSITIVE CONTROL. Every "spared" assertion below is satisfied by an empty
+  // deletable list, so without this a fence that refuses everything — or a
+  // harness wired to nothing — reads green across the whole file.
+  it('lets a model through when every timestamp under it is long quiet', () => {
+    expect(
+      run([row()]),
+      'the quiet baseline must be DELETABLE, or every "spares" case below passes vacuously'
+    ).toEqual({ deletable: [1], skipped: [] });
+  });
+
+  it('keeps a candidate that has no versions at all', () => {
+    expect(run([])).toEqual({ deletable: [1], skipped: [] });
+  });
+
+  it('spares a model whose version was CREATED inside the window', () => {
+    expect(run([row({ createdAt: recent })]), 'clause: mv."createdAt" inside the window').toEqual({
+      deletable: [],
+      skipped: [1],
+    });
+  });
+
+  it('spares a model whose version was UPDATED inside the window', () => {
+    expect(run([row({ updatedAt: recent })]), 'clause: mv."updatedAt" inside the window').toEqual({
+      deletable: [],
+      skipped: [1],
+    });
+  });
+
+  // The reported exhibit: the model and its only version were created in the
+  // same minute and never touched again, and the trained resource arrived later
+  // as a file. This is the only clause that can see it.
+  it('spares a model whose FILE was created inside the window', () => {
+    expect(
+      run([row({ latestFileAt: recent })]),
+      'clause: mf."createdAt" inside the window'
+    ).toEqual({ deletable: [], skipped: [1] });
+  });
+
+  it('does not treat "no files" as activity', () => {
+    expect(run([row({ latestFileAt: null })])).toEqual({ deletable: [1], skipped: [] });
+  });
+
+  it('spares a model when a REQUIRED timestamp is unreadable', () => {
+    // createdAt / updatedAt are NOT NULL in the schema, so a missing value means
+    // the query shape drifted. Deleting is irreversible; sparing costs a night.
+    expect(
+      run([row({ createdAt: null })]),
+      'clause: unreadable required timestamp must fail CLOSED'
+    ).toEqual({ deletable: [], skipped: [1] });
+    expect(run([row({ updatedAt: 'not-a-date' })])).toEqual({ deletable: [], skipped: [1] });
+    expect(run([row({ latestFileAt: 'not-a-date' })])).toEqual({ deletable: [], skipped: [1] });
+  });
+
+  it('spares the WHOLE candidate set when a row carries no usable modelId', () => {
+    const orphan = { ...row(), modelId: undefined } as unknown as ModelVersionActivityRow;
+    expect(
+      run([orphan], [1, 2, 3]),
+      'clause: unattributable activity must fail CLOSED across the batch, not protect nothing'
+    ).toEqual({ deletable: [], skipped: [1, 2, 3] });
+  });
+
+  it('separates active from quiet models inside one candidate set', () => {
+    const result = run(
+      [
+        row({ id: 10, modelId: 1, latestFileAt: recent }),
+        row({ id: 20, modelId: 2 }),
+        row({ id: 30, modelId: 3, createdAt: recent }),
+        row({ id: 40, modelId: 4 }),
+      ],
+      [1, 2, 3, 4]
+    );
+    expect(result).toEqual({ deletable: [2, 4], skipped: [1, 3] });
+  });
+
+  it('protects a model when ANY of its versions is active', () => {
+    const result = run(
+      [row({ id: 10, modelId: 1 }), row({ id: 11, modelId: 1, updatedAt: recent })],
+      [1]
+    );
+    expect(result).toEqual({ deletable: [], skipped: [1] });
+  });
+
+  it('preserves the candidate order the caller supplied', () => {
+    // The SELECT orders by id for consistent lock ordering; the filter must not
+    // reshuffle what it hands to the DELETE.
+    expect(run([], [5, 1, 9])).toEqual({ deletable: [5, 1, 9], skipped: [] });
+  });
+
+  it('accepts the string timestamps a raw driver can hand back', () => {
+    expect(run([row({ latestFileAt: recent.toISOString() })])).toEqual({
+      deletable: [],
+      skipped: [1],
+    });
+    expect(run([row({ createdAt: quiet.toISOString(), updatedAt: quiet.toISOString() })])).toEqual({
+      deletable: [1],
+      skipped: [],
+    });
+  });
+
+  it('measures the window from the supplied clock, not the wall clock', () => {
+    // A row one day inside the window is active; the same row is quiet once the
+    // clock advances past it. This is what makes ACTIVITY_WINDOW_DAYS load-bearing
+    // rather than decorative.
+    const edgeIn = daysBefore(ACTIVITY_WINDOW_DAYS - 1);
+    expect(filterModelsWithRecentActivity([1], [row({ updatedAt: edgeIn })], NOW)).toEqual({
+      deletable: [],
+      skipped: [1],
+    });
+    const later = new Date(NOW.getTime() + 2 * DAY_MS);
+    expect(filterModelsWithRecentActivity([1], [row({ updatedAt: edgeIn })], later)).toEqual({
+      deletable: [1],
+      skipped: [],
+    });
+  });
+});
--- a/src/server/jobs/__tests__/remove-old-drafts.test.ts
+++ b/src/server/jobs/__tests__/remove-old-drafts.test.ts
@@ -13,12 +13,62 @@ vi.mock('~/utils/storage-resolver', () => ({ deregisterFileLocationsBatch: mockD
 vi.mock('~/utils/logging', () => ({ createLogger: () => () => undefined }));
 vi.mock('~/server/jobs/job', () => ({ createJob: (_n: string, _c: string, fn: unknown) => fn }));
 
-import { removeOldDrafts } from '~/server/jobs/remove-old-drafts';
+import { removeOldDrafts, ACTIVITY_WINDOW_DAYS } from '~/server/jobs/remove-old-drafts';
 import { dbMock } from '~/__tests__/mocks/db.mock';
 import { loggingMock } from '~/__tests__/mocks/logging.mock';
 const mockDbRead = dbMock.dbRead;
 const mockDbWrite = dbMock.dbWrite;
 const mockLogToAxiom = loggingMock.logToAxiom;
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+const daysAgo = (n: number) => new Date(Date.now() - n * DAY_MS);
+
+// Fixture ages sit far off the 30-day boundary on BOTH sides — 200 days for
+// "long quiet", 5 days for "still in use". Nothing is built at 30, so no clause
+// can be reached (or missed) by landing exactly on its own cutoff, and a mutant
+// that widens the window to 300 days has somewhere to be caught.
+const QUIET_DAYS = 200;
+const RECENT_DAYS = 5;
+const quiet = () => daysAgo(QUIET_DAYS);
+const recent = () => daysAgo(RECENT_DAYS);
+
+/**
+ * A row shaped like the job's per-batch ModelVersion lookup actually returns.
+ * Defaults to "long quiet in every column" so each test names only the one
+ * timestamp it is exercising.
+ */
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Adversarial audit — round 1 (full), fixes pushed

Round 1 was a blind full audit against `a391d57790`. It returned **no 🔴** and four 🟡.
Three are fixed in `28457f2e8f`; one is left open for the reviewer, deliberately.

**Fixed**

1. The destroyed-ids Axiom event (`remove-old-drafts.ts:216`) was unpinned — mutating
   `modelIds: deletable` → `batch` survived a green 31/31 suite. Independently reproduced
   before fixing.
2. The spared-models warning (`:190`) had the same hole — `modelIds: skipped` → `batch`
   also survived.
3. The seam guard pinned all four `INTERVAL` literals to `ACTIVITY_WINDOW_DAYS`, including
   the `m."updatedAt"` abandonment threshold, which is not part of the fence. Narrowing the
   fence would have turned the guard red and steered the fix into widening the reaper's kill
   radius, green.

**Not fixed — reviewer's call, and this is the material half of reviewer question 3**

A model that is both `Deleted` and `Private` is now unreapable by any job in
`src/server/jobs`: `remove-old-drafts` is the only reaper of `Deleted` models, and the new
`availability != 'Private'` exclusion removes them from it. The PR body frames the exclusion
as covering private *drafts*; the `Deleted` half is unstated and carries a data-retention
edge, since the user explicitly deleted those. Their rows, `file_locations` entries and the
S3 objects those keep whitelisted against the quarantine sweep persist indefinitely.

Narrowing to `NOT (availability = 'Private' AND status = 'Draft')` would restore it. I have
not made that change: it re-enables a deletion path, in a PR whose purpose is to stop
over-deletion.

**Ledger** · round 1 · payload lines changed this round: 21 (`remove-old-drafts.ts`);
scaffolding 68 (`remove-old-drafts.test.ts`) · since round 1: 21

**Verified, not asserted:** the two survivors were reproduced as SURVIVED against the
pre-fix tree and confirmed to die after it, each to its own assertion message rather than
collaterally to the other. Positive controls (delete `batch` instead of `deletable`; revert
the deregister scoping) die in both trees, so the harness can go red. The three seam mutants
each kill only their own guard — an age-literal mutant no longer trips the fence guard, and
vice versa, which is the decoupling being claimed.

**Instrument limits, stated:** the local run used a symlinked root `node_modules`, so 6 test
files in `src/server/jobs/__tests__` fail to *collect* on `Cannot find package 'kysely'`
(457 tests pass, 0 assertion failures, and `remove-old-drafts` appears nowhere in the
failures). Repo-wide typecheck was not confirmed locally for the same reason; CI is the
authority on both. Blast-radius figures in the PR body require production DB access and were
not re-verified in this round.

```audit-claims round=1 audited=a391d57790fca36db9613317a06313da55eb6c75..28457f2e
1. The destroyed-ids event now logs `deletable`, pinned by an assertion in the mixed-batch test.
2. The spared-models warning now logs `skipped`, pinned by an assertion in the same test.
3. `REAP_AGE_DAYS` is split from `ACTIVITY_WINDOW_DAYS`; the fence clauses, the age clause and the interval count are asserted separately.
4. No change was made to what the job selects or deletes.
```

ZacxDev:
## Adversarial audit — round 2 (delta), fixes pushed

Round 2 audited `a391d577..28457f2e` blind against the round-1 claims. It confirmed all four
as actually fixed, and returned **one 🟡 and two 🟢**. All three are addressed in `471c882e8d`.

**Fixed**

1. 🟡 The interval guards matched **raw** SQL, so a `--` comment counted as a clause. Round 2
   demonstrated it: an explanatory comment merely *mentioning* an interval turned the count
   guard red with `expected 4 but got 5` under a title claiming a clause had escaped — with no
   fifth clause present. The previous commit added a SQL comment two lines above the age
   clause and avoided the trip only by not writing the literal. Same read was wrong in the
   other direction too (a comment quoting a clause satisfied a `toContain` guard with the real
   clause deleted), backstopped by the whole-predicate pin so it was latent.
   Fixed by extracting `readExecutableSql()` and pointing every SQL-text guard at it.
2. 🟢 `REAP_AGE_DAYS`'s doc claimed it was "read by the `m."updatedAt"` clause". A SQL literal
   cannot read a TS constant, and the symbol has **zero** runtime readers. Rewritten to say
   what it is, and to name the asymmetry with `ACTIVITY_WINDOW_DAYS`, which *is* read at
   runtime. The test-file header carried the same wrong claim and is corrected with it.
3. 🟢 The counterfactual comment said narrowing the fence would pass "with the suite green";
   the whole-predicate pin would also have gone red. The hazard stands — that red names no
   clause and the repair it invites is the dangerous one — but the comment now says so.

**🔴 Correction to `28457f2e8f`'s commit body, which a later reader would otherwise cite**

Its mutation table has two overstated rows: `age interval 30 → 7 days — DIES, age guard only`
and `fence interval 30 → 90 days — DIES, fence guard only`. Each in fact fails **two** tests —
its own guard **plus** `pins the whole predicate`. The decoupling those rows support is
unaffected and holds both ways (an age mutant never trips the fence guard, and vice versa),
but "only" was wrong. Correcting here rather than silently, since the body has been readable
since it was pushed.

**Verified by mutation** (33 tests, both files, green baseline first):

| mutant | result |
|---|---|
| a doc comment MENTIONING an interval | **20/20 PASS** — was red before; this was the finding |
| a real fifth interval clause | DIES — count guard + predicate pin |
| comment quoting a clause, real one deleted | DIES — fence guard + predicate pin |
| delete `batch` instead of `deletable` | DIES — its own assertion |
| destroyed-ids event → `batch` | DIES — its own assertion |

Repo-invariant lint-rule suites: 5 files / 200 tests pass. `eslint` rc 0, zero output.
`prettier` clean (negative-controlled).

**Ledger** · round 2 · payload lines changed this round: 12 (`remove-old-drafts.ts`, all doc
comment); scaffolding 54 · since round 1: 33. 🔴 **Stated plainly: behavioural payload movement
across BOTH fix rounds is zero** — every payload line either of them changed is a comment or a
constant with no runtime reader. The job selects and deletes exactly what it did at
`a391d57790`.

**Still not verified by anyone:** the SQL predicate is never executed — this repo has no
DB-backed vitest project — so every SQL-level guard here is a spelled string comparison, not a
behavioural one. And the PR body's production blast-radius figures have not been re-measured
in either round.

```audit-claims round=2 audited=28457f2e8f23f31f913f0851105f48b9585afc10..471c882e
1. All SQL-text guards now read comment-stripped SQL via `readExecutableSql()`; a `--` comment no longer counts as a clause, and a comment quoting a clause no longer satisfies a guard.
2. `REAP_AGE_DAYS`'s doc comment no longer claims a runtime wiring it does not have, and names the asymmetry with `ACTIVITY_WINDOW_DAYS`.
3. The test-file header no longer describes both constants as read by the runtime fence.
4. The round-1 counterfactual comment now acknowledges the whole-predicate pin would also have gone red.
5. No change was made to what the job selects or deletes.
```

ZacxDev:
## Adversarial audit — round 3 (delta): CLEAN. Ladder ends here.

Round 3 audited `28457f2e..471c882e` blind against the round-2 claims. **All five verified
against the tree**, every repointed guard shown reachable and uniquely red-able, and the
production job confirmed untouched. **No finding required a fix** — that is the ladder's
terminal condition, so there is no round 4.

**Guard reachability, measured — no guard is vacuous or masked by another.** Each was shown to
go red *alone*:

| mutant | reds |
|---|---|
| real 5th interval inside `WHERE` | predicate pin + count guard |
| real 5th interval in the `JOIN ON` (outside the pinned slice) | **count guard only** — its unique reach |
| `REAP_AGE_DAYS` 30→7, SQL untouched | abandonment guard only |
| `ACTIVITY_WINDOW_DAYS` 30→7, SQL untouched | fence guard only |
| drop `m."userId"` from the SELECT list | userId guard only |

Round 2's fix was also verified in **both** directions against the pre-fix tree: a `--` comment
mentioning an interval was red before and is green after; a comment quoting a clause with the
real one narrowed passes the fence guard before (caught only by the predicate pin) and dies on
the fence guard *for its own reason* after — confirming the "latent, backstopped" reading.

### The one 🟢, and what I did about it — `28f1cb3509`

`readExecutableSql()`'s doc claimed its output is "the SQL the database actually executes". It
is not: `/--[^\n]*/g` strips to end-of-line unconditionally, including a `--` inside a quoted
string literal, which PostgreSQL does not treat as a comment. Measured — a JOIN line carrying
`m.name NOT LIKE '%--%'` followed by a real fifth `INTERVAL` clause is **fully green at 33/33**,
while the pre-round-2 raw-SQL count guard catches it.

So repointing the guards traded the false-RED it was fixing for a narrow false-GREEN, and my
doc asserted a guarantee the code does not carry. The trade is still net-positive and the
trigger is contrived — it needs a `--`-bearing string literal, on one line, before a new clause,
**outside** the `WHERE … ORDER BY` slice, since inside it the predicate pin fails loudly.
Nothing in this SELECT is near that shape.

I documented the limit rather than parsing SQL in a test helper, and named the remedy for
whoever hits it. **Comment-only, no executable change.** Re-verified after: 33/33 on both files,
full `test:lint-rules` **28 files / 448 tests** rc 0, eslint rc 0 zero output, prettier clean.

### Corrections to this PR's own reporting

- An earlier round quoted the lint-rule suites as "5 files / 200 tests". That was the subset
  that round named, not the repo's `test:lint-rules` script. The real figure is **28 files /
  448 tests** — it understated in the safe direction, but it read as the whole suite.
- `28457f2e8f`'s mutation table said two SQL-literal mutants die to their own guard "only";
  each fails **two** tests, its own guard plus `pins the whole predicate`. Corrected in
  `471c882e8d`'s body and re-confirmed by round 3.

### What nobody has verified, stated plainly

- **The SQL predicate is never executed.** This repo has no DB-backed vitest project, so every
  SQL-level guard here is a spelled string comparison. The fence's correctness against a real
  database rests on review, not on a test.
- **Behavioural payload movement across all three fix rounds is ZERO.** Every payload line any
  round changed is a comment or a constant with no runtime reader. The job selects and deletes
  exactly what it did at `a391d57790` — the audit improved the guards around the change, not
  the change.
- **The production blast-radius figures in the PR body were not re-measured** in any round;
  they need DB access.
- **`Deleted` + `Private` models remain unreapable by any job** — round 1's finding 4, still
  open and still the reviewer's call. It fails safe, which is why I did not touch it.

**Ledger** · round 3 · payload lines changed this round: **0 executable** (12 lines in the
payload file, all JSDoc) · since round 1: 1 executable (`export const REAP_AGE_DAYS = 30;`).
Two consecutive rounds at zero executable payload — the attribution gate and the clean-round
rule agree the ladder has left the PR.

PR #4589, opened 2026-09-03
F5b: bound the model-slot App Block iframe to the viewer's viewport

ZacxDev:
## The gap

`IframeHost.applyHeight` — the handler for the `RESIZE_IFRAME` postMessage a block sends to grow its own iframe — carried a comment reading "Three layers of height defense":

1. `isFinite` + positive guard (rejects `NaN`, `Infinity`, negatives)
2. `manifest.iframe.maxHeight`, **if set**
3. `HARD_HEIGHT_CEILING = 8_000` — the OOM backstop

Layer 2 is optional, and that is the hole. `public/schemas/app-block/v1.json` types `iframe.maxHeight` as `["integer","null"]` ("null for unbounded") and the `iframe` object declares **no required fields at all** — so a manifest that simply omits `maxHeight` is bounded only at 8000px. A block self-reporting 3000px got a 3000px-tall iframe inside a ~640px-tall phone viewport, and the inline slot swallowed the page.

## The clamp

A fourth layer:

```js
next = Math.min(next, Math.max(min, viewportHeight - frameOverhead));
```

The framed widget may never exceed the visible viewport height, but the manifest's `minHeight` still wins if the remaining budget is smaller than it. The block scrolls internally — that is the intended outcome.

### `− frameOverhead` is load-bearing, not a detail

`framed()` renders `AppBlockChrome` *above* the iframe inside one bordered box, so clamping the **iframe** to the viewport still yields a `viewport + chrome + borders` **widget**. Measured at 390×640 with the real cascade loaded: chrome **31px** (matching `AppBlockChromeResponsive`'s pin of `22 + 8 + 1`), frame border **1px top + 1px bottom** → overhead **33px**. A block reporting 640 produced a 673px widget on a 640px screen — layer 4 bounding exactly the wrong box. On `model.sidebar_top` the block sits in normal page flow on mobile, so the viewer could never see the whole block *and* anything below it.

The overhead is **measured live** as `frame.offsetHeight − iframe.offsetHeight` — invariant to the iframe's current height, borders included, and still correct if the frame ever gains another sibling. It is **not** hardcoded: `CHROME_BAR_PX` is a *resting* single-row contract that has already gone stale once in this arc, and the 33 above is an observation, not a constant the code assumes.

Against the complete approved population (11 of 11 blocks; floors `400 ×1 / 600 ×5 / 640 ×3 / 700 ×2`, every one declaring `maxHeight: 4000`):

| viewport | without the subtraction | with it |
|---|---|---|
| 640 | 11/11 overflow, worst 93px | **5/11** overflow, worst 93px (400- and 600-tiers now fit) |
| 844 (iPhone-15 class) | **11/11 overflow, 33px each** | **0/11 overflow** |

Two further properties, both implemented and covered:

- **Read at call time, never captured at mount** — both the viewport *and* the overhead. Either value stashed at mount is stale after a rotate, a browser-chrome resize, or a bar that re-wraps at a new width.
- **Re-clamped on `resize`.** The block's own *stated* height (not the clamped one) is stashed in a ref, and the four rules are re-applied when the viewport moves — so the bound tracks the viewport in both directions rather than ratcheting downward. Host-side only: nothing is posted back to the block (`RESIZE_IFRAME` is one-way, block → host). The listener is removed on unmount.

A viewport that cannot be measured — no `window` (SSR/prerender), or an `innerHeight` that is not a positive finite number — means **do not clamp**, never clamp-to-zero. The overhead degrades the same way: an unusable measurement yields 0, which can only make the budget *larger*, never smaller.

**Known limit, stated rather than implied:** the overhead is re-measured when the clamp *runs* (a `RESIZE_IFRAME`, or a window `resize`). A chrome bar that changes height with no viewport change — a menu opening, a late font swap — does not itself re-trigger the clamp. Closing that would need a `ResizeObserver` on the chrome, which is not warranted for a bound whose job is "roughly one screen".

## 🔴 What this does NOT bound — the publisher's `minHeight`

`Math.max(min, budget)` means the manifest floor always wins, and **nothing in this PR bounds it**. `HEIGHT_MAX_CEILING` in `src/server/services/block-manifest-validator.service.ts` lets a manifest declare `minHeight` up to **4000**; measured at 390×640 with `{minHeight: 4000}` and the block reporting 100, appliedHeight is **4000** — the exact failure this PR is about, reachable through one schema-legal field, with the clamp present.

At modest values it still bites: at a 640px viewport the budget is 607, so the 640-tier (×3) and 700-tier (×2) are bound by their own floor and overflow by 33px and 93px — **5 of 11**. The 400- and 600-tiers fit.

**Capping `minHeight` is #4606 (draft), based directly on `main` — not stacked, and nothing in it is in this PR.** Do not act on #4606's downstream checklist while reviewing this one: this branch leaves the canonical schema at 4000, so changing a mirror to expect 800 now would turn that repo's CI red rather than green. Reasoning for the split:

- This host-side fix has **zero** cross-repo cost and takes the live population from 11/11 overflowing to 0/11 at 844px. It should not wait behind anything.
- The cap closes a hole **no live block exercises** (largest live `minHeight` is 700), costs hand edits in repos we do not control, and **does not close the residue either** — 640 and 700 are modest values well under any plausible cap that still exceed the 607px budget.

Removing the floor from the clamp instead is *not* an option: mutant D shows it is what keeps a short or failed block's reserve intact.

## Surface affected

**The inline model-page slot, and only it.** `IframeHost` is the sole `RESIZE_IFRAME` consumer in `src/`. The full-page host `PageBlockHost` (`/apps/run/<slug>`) has **zero** `RESIZE_IFRAME` handling and is already viewport-bound by its own `calc(100dvh - HEADER_HEIGHT_PX)` — untouched, and there is deliberately no parity mirror of the new suite.

## Tests

`src/components/AppBlocks/IframeHostViewportHeightClamp.browser.test.tsx` — 8 browser cases, mocks mirroring the neighbouring `IframeHostThemeChange.browser.test.tsx` scaffold (including `useOptionalFeatureFlags` alongside `useFeatureFlags`, whose omission is a link failure that collects nothing).

🔴 **This suite loads `@mantine/core/styles.css`, and that is not cosmetic.** The first revision did not, and it reported a chrome overhead of **98** with a frame border of **0** — `border: 1px solid var(--mantine-color-default-border)` is invalid-at-computed-value-time with no cascade — while **every assertion still passed**, because they are all self-relative. The residue table published from that 98 was wrong in a way that mattered: it claimed the 600-tier (five of eleven live blocks) does not fit, when at the real budget of 607 it does. `renderReady` now asserts the sheet is loaded; removing the import fails all 8 cases with a named message instead of going quietly meaningless.

> The sibling `AppBlockChromeResponsive.browser.test.tsx` says its siblings *must not* import the stylesheet — that rule is scoped to suites asserting **attributes and ARIA**, which is what the shared scaffold is for. This suite asserts **pixels**, so it takes the same exception that file takes for itself. The longer-term home for pixel assertions is the `geometry` project added by **#4601** (`e92cf5fe4`) — which exists precisely because the component tier omits the real cascade and lets geometry numbers be meaningless while assertions pass. **This PR hit that trap in the same repo, the same week, in a file whose sibling documents it.** Anyone measuring geometry in a `.browser.test.tsx` should read that project's setup first.

🔴 **Every case sets a phone viewport explicitly.** Vitest's browser default is 414×896 (measured), and `test/component-setup.tsx` sets none — at 896px tall, every height the neighbouring IframeHost suites assert (640/700/800) is already *under* the bound. The shared `setViewport` helper calls `page.viewport(...)` then re-reads `window.innerHeight` to assert the value took.

🔴 **The headline case asserts the FRAME, not the iframe** — `frameHeight() === 640`, `document.documentElement.scrollHeight <= 640`, and `appliedHeight() === 640 − chromeHeight() − frameBorderPx()`, with chrome and border read as **independent** observables (`getBoundingClientRect` and `getComputedStyle`) rather than as the implementation's own `frame − iframe` subtraction, which would be vacuous. The wait condition is `toBe(640)` rather than `<= 640`: the loose form is satisfied by the *pre-update* minHeight reserve (160 + 33 = 193 ≤ 640), so `vi.waitFor` returned before the resize landed and graded the wrong frame.

Also covered: the negative half (400px at 640px stays 400, with a precondition guarding against the overhead growing past 240); the `minHeight` floor on a 100px viewport; `maxHeight` still winning when tighter than the budget; the re-clamp across a viewport change in both directions; a viewport shrink *below* the floor falling back to the floor; the pre-handshake no-op (labelled an **invariant guard**, not regression coverage); and a listener-leak ledger asserting every `resize` handler added is removed on unmount, with a positive control on the ledger itself.

Base named explicitly, because `main` moved during this work (#4601 `e92cf5fe4`, then `34510fb587`, then `b7fd0d0685`):

| | result |
|---|---|
| `origin/main` @ `b7fd0d0685` (only the `IframeHost.tsx` hunk reverted) | `Tests 5 failed \| 3 passed (8)` |
| this branch | `Tests 8 passed (8)` |

`IframeHost.tsx` and `IframeHostReadyTransition.browser.test.tsx` had **0 commits** touching them between this branch's merge-base `bbbe837d13` and `b7fd0d0685`, so the matrix is not an artefact of which base was picked.

### Mutation results

| mutant | change | killed by |
|---|---|---|
| A | delete **only** the layer-4 clamp expression | *"layer 4 did not bound the WIDGET…"* |
| B | delete the `resize` re-clamp effect | *"the slot did not re-grow when the viewport did…"* + the leak-ledger positive control |
| C | stash the **clamped** height instead of the block's stated one | *"the slot did not re-grow when the viewport did…"* |
| D | `Math.max(min, budget)` → bare `budget` | *"the manifest floor lost to the viewport: … must apply 160px"* |
| E | read the viewport **once at module scope** | *"layer 4 did not bound the WIDGET…"* |
| F | hardcode the source's default `200` as the floor in the **re-clamp** call site | *"the re-clamp used the wrong floor: … must fall back to 160px"* |
| F2 | same, in the **applyHeight** call site | *"the manifest floor lost to the viewport: … not 200px"* |
| G | delete the `removeEventListener` cleanup | *"1 of 1 `resize` listener(s) added by the host outlived its unmount…"* |
| H | drop the overhead subtraction | *"layer 4 did not bound the WIDGET…"* |

**Mutants that survived earlier revisions and drove test changes** (all recorded in the test file):

- **C** survived when the re-clamp case negotiated at the *tall* viewport first — the stashed clamped value was numerically identical to the answer the last assertion wanted.
- **F** survived because every re-clamp case ran at a viewport where the *budget* wins, and because the fixture's `MIN_H` was **200**, identical to the source's own `?? 200` default. `MIN_H` is now **160**, plus a case that shrinks the viewport below the floor.
- **G** survived because a leaked listener is silent (React does not warn on setState after unmount). Hence the add/remove ledger.

## 🔴 Existing behaviour changed — one test

**`src/components/AppBlocks/IframeHostReadyTransition.browser.test.tsx:516`**, `'an over-ceiling height is clamped to the manifest maxHeight'`.

It relied on the harness default viewport (414×896) with a fixture `MAX_HEIGHT` of 800. With the overhead subtraction the budget at 896 is tighter than 800, so layer 4 legitimately became the binding layer and the assertion read `expected 798 to be 800`. The clamp was right; the test's premise was a coincidence of the default viewport. Fixed by setting `page.viewport(414, 1200)` so `maxHeight` is unambiguously what binds — which also removes an unstated dependency on the harness default. **It is the only pre-existing test changed, and it is a deliberate consequence of this PR.**

`src/components/AppBlocks/` browser suite: **41 files / 492 tests, all passing**. `pnpm typecheck`: 0 errors.

## Known and deliberately NOT fixed

Recorded so the next reader knows these are **open, not absent**. Both fail **red** rather than falsely green, which is why they are being left:

- **`expectedBudget()` sums fractional `getComputedStyle` readings against an integer applied height.** `chromeHeight()` and `frameBorderPx()` are sub-pixel floats; `appliedHeight()` is what the host wrote. Today they agree exactly at the viewports under test. If a future theme makes the chrome fractional, the equality assertions go red — a visible failure, not a silent pass.
- **The suite goes red if the frame gains a third child.** `frameBorderPx()` + `chromeHeight()` reconstructs the overhead from the two things in the frame today. A new sibling makes that reconstruction under-count and the budget assertions fail. The *production* clamp is unaffected — it measures `frame.offsetHeight − iframe.offsetHeight`, so it absorbs a third child correctly; only the test's independent re-derivation would need updating.

## A false reachability claim, corrected

`frameOverheadPx`'s docblock asserted its `!frame || !iframe` branch was unreachable, reasoning from React commit ordering. **That was wrong**, and this file's own comments contain the mechanism: the re-clamp effect deliberately omits `status` from its deps, so its `resize` listener survives a `BLOCK_ERROR {fatal:true}` → `'collapse'` → `return null`, which unmounts the frame and iframe and nulls both refs while `reportedHeightRef` still holds a height. Instrumentation reached it (hit count 1).

No wrong output today — `fatal` is terminal and the host renders `null`, so the recomputed value is unobservable. The harm was what the sentence licensed: a maintainer acting on it writes `frame!.offsetHeight` and ships a `TypeError` out of a resize listener for every viewer who rotates after a fatal block error. The type-narrowing label is kept (the branch *is* inert); the false claim is replaced with the path, and a companion clause that called the pre-layout case "the reachable case" is now recorded as **not established either way** rather than asserted.

**The shape was then swept at every site, not just the reported one** — 21 reachability/invariant claims in `IframeHost.tsx`: **12 verified true, 2 corrected** (both mine, both in that one paragraph), **7 left unsettled and labelled as such** (claims about the blocks-react SDK and Mantine, which cannot be settled from this repo). The eight surviving React-lifecycle claims each check out against the code — notably H-11's "`status` can never become 'ready' again from a terminal state", which holds because all five `setStatus` writers are `current`-guarded and exactly one writes `'ready'`, only from `'loading'`.

*(The two commit messages on this branch were written before the 98px figure was corrected and still carry it; they cannot be amended after pushing. This body is the corrected record.)*


ZacxDev:
**Adversarial audit — round 1** (full). Findings and the fixes they produced, recorded so the delta round can be anchored.

Round 1 audited `4cadfb4d9a` and returned no deploy-blockers, two 🟡 that falsified the PR's headline claim, one 🟡 that could have escalated, and three lower-value findings. It re-derived every self-reported number in the PR body against the tree — red-at-base matrix, all five mutants and their killing messages, the 484→490 count — and all of them held.

🟡-4 (a live block suppressing its own overflow, which would have made the clamp strand content) was settled separately by a full enumeration of the live population: **11 of 11 approved blocks**, each block's shipped HTML + CSS + JS fetched, four instruments each positive-controlled. **No live block suppresses vertical overflow**, and neither scaffold ships a root `overflow:hidden`. It does not escalate; it stands as a forward-looking risk only.

```audit-claims round=1 audited=4cadfb4d9a..761fb062
1. 🟡-1 the clamp now bounds the framed WIDGET rather than the iframe — budget is viewport minus a live-measured `frame.offsetHeight - iframe.offsetHeight`, re-read on every clamp including inside the resize handler
2. 🟡-2 was taken as a validator cap and then SPLIT OUT of this PR entirely — this branch no longer touches the manifest contract; the cap moved to #4606 (draft) with its two cross-repo prerequisites named
3. 🟡-6 the fixture's distinctness claim was false because MIN_H equalled the source's own 200 default — MIN_H is now 160, plus a new case that shrinks the viewport below the floor so the re-clamp's floor path is exercised
4. 🟢-1 an unmount ledger now asserts every resize listener added is removed; the `reported === null` early-return is relabelled in source and suite as a type-narrowing invariant, explicitly NOT coverage
5. 🟢-2 the test file header now names `setViewport`, the function that exists
6. one pre-existing test changed as a deliberate consequence: `IframeHostReadyTransition.browser.test.tsx:516` now sets viewport 414x1200 so `maxHeight` is unambiguously what binds, rather than the chrome-adjusted budget
```

ZacxDev:
**Adversarial audit — round 2** (delta, `4cadfb4d9a..761fb062`). All six round-1 claims dispositioned **actually fixed**; self-reported numbers re-derived and confirmed; mutants F, G, H each re-killed by exactly one test with its claimed message. Two findings, both prose, no deploy-blockers.

```audit-claims round=2 audited=761fb0625b..7f2adae8
1. F2 the 98px chrome overhead was a stylesheet-less artifact — the clamp suite now imports `@mantine/core/styles.css`, and the overhead is requoted as 31 (chrome) + 2 (frame borders) = 33, re-derived first-hand at 390x640 rather than taken from the sibling suite
2. F2 second-order defect the audit did not name: two assertions derived the budget as `viewport - chrome`, omitting the borders — invisible while borders computed to 0. They now subtract `frameBorderPx()`, read independently from `getComputedStyle` rather than derived from the implementation's own arithmetic
3. F2 a negative control was added for the new import: with it removed all 8 tests fail naming the unloaded stylesheet, where previously all 8 passed silently
4. F2 the residue table was recomputed everywhere it appears — `IframeHost.tsx`, the test header and in-test comments, and BOTH PR bodies: at 640 the overflow is 5/11 worst 93px with the 400- and 600-tiers fitting, not 10/11 worst 158px
5. F1 all reverted-cap content stripped from this PR's body — verified 0 occurrences of `MIN_HEIGHT_MAX_CEILING`, `Pre-landing both`, `schema-parity`, `semantic.go`, `178 passed` — and replaced by a warning not to act on #4606's downstream steps while reviewing this branch, which leaves the schema at 4000
6. nit: `frameOverheadPx`'s `!frame || !iframe` relabelled a TYPE NARROWING to match `reported === null`; the genuine degradation path is the separate `overhead > 0` pre-layout check
```

PR #4590, opened 2026-09-03
fix(app-blocks): full-bleed opt-out ledger was inert in production (data-testid is stripped)

ZacxDev:
## The defect: production-only, and no test tier can see it

`src/styles/globals.css` carries the full-bleed opt-out ledger — the only way an App Block escapes the 1600px ultrawide cap. Its one member was written as:

```css
[data-testid='app-page-frame'][data-block-id='playable-collections'] {
  --app-page-max-width: none;
}
```

`next.config.mjs` sets:

```js
compiler: process.env.NODE_ENV === 'production'
  ? { reactRemoveProperties: { properties: ['^data-testid$'] } }
  : {},
```

So **every `data-testid` is compiled out of the production DOM**, the compound selector matches nothing on the live site, and `playable-collections` — a collection player whose every open-collection view mode is uncapped by the app itself — renders letterboxed at the cap. That is the exact outcome the rule's own comment says it exists to prevent.

## Measured evidence

On the live site, `civitai.com/apps/run/playable-collections`, image `20260902233645-bbbe837`:

| Observation | Value |
|---|---|
| the ledger rule present in the deployed CSS, verbatim | yes (2136 rules scanned, 0 sheets blocked) |
| `querySelectorAll("[data-testid='app-page-frame'][data-block-id='playable-collections']").length` | **0** |
| `querySelectorAll("[data-block-id='playable-collections']").length` | **1** |
| `data-testid` attributes on the page | **0** across 615 elements |
| elements carrying some other `data-*` attribute | **209** (so the zero above is a measurement, not a broken search) |
| computed `--app-page-max-width` on the capped box | `1600px` — the opt-out did not apply |
| at viewport 1920 | app column x=160 w=1600, i.e. capped |

## Why the existing tests could not see it

Every tier runs with `NODE_ENV != production`, where `data-testid` is present and the selector matches. `PageBlockHostMaxWidth.browser.test.tsx` (~line 455) *injects that very selector* and asserts full-bleed at 2560x1080 — and passes, correctly, on the broken tree. It is not a bad test; it is structurally blind, because its environment fixes the one dimension the defect lives on. The gating node guard `__tests__/pageBlockHostMaxWidth.test.ts` reads source, not the build config, so it could not see it either. This is a config-blind suite plus an unowned seam: two files each correct alone, broken together.

No rendered test can close this. Only a check that reads the two *configurations* and compares them can.

## The fix

`PageBlockHost` now stamps a presence marker on the same element it already puts `data-block-id` on:

```jsx
data-app-page-frame=""
data-fit={fit}
data-block-id={blockId}
```

and the ledger is re-keyed onto `[data-app-page-frame][data-block-id='…']`.

Naming follows the repo idiom for exactly this job — `AdhesiveAd` stamps `data-adhesive-ad=""` and `AppFooter` stamps `data-app-footer=""` precisely so `globals.css` can key off something the compiler keeps.

The frame half is **kept rather than dropped**. `[data-block-id='…']` alone would work today, but `PageBlockHostMaxWidth.browser.test.tsx` (~line 527) deliberately asserts the ledger is keyed on something app-specific "and not on something every host carries" — the frame attribute is what says *this is the page host*, not some other surface that happens to carry a block id. Preserving that keeps the original design intent intact while surviving the production strip.

## New guard — reads one side, compares it to the other

`src/components/AppBlocks/__tests__/ledgerSelectorSurvivesProdStrip.test.ts`, in the **gating** node `unit` project (the browser tier is report-only in CI). It:

- parses `compiler.reactRemoveProperties.properties` out of `next.config.mjs`
- parses the ledger selectors out of `src/styles/globals.css`
- fails if any ledger selector depends on an attribute production strips
- applies the same rule to the ledger's own **HOW-TO-ADD-ONE template**, because the next entry is copied from it (that is how this bug reproduces one commit later)
- checks every attribute the ledger depends on is actually **stamped** by `PageBlockHost.tsx` — surviving the strip is worthless if nothing renders it

It restates neither side, so it keeps working if either changes.

**Fails closed.** An unparseable config, an empty strip list, or zero parsed ledger rules is a failure, not a quiet pass — a reassuring zero here is indistinguishable from a guard wired to nothing. It asserts a minimum parsed-rule count (>= 1) for the same reason, and it carries a **negative control** that builds a known-bad selector from the *parsed* patterns (anchors stripped off a literal) and asserts the check reports it, so the check has been watched to go red.

### Red -> green matrix

Run with the identical guard file in both trees:

| Tree | Result |
|---|---|
| `origin/main` (`bbbe837d13`) | **RED** — `Tests 2 failed \| 3 passed (5)` |
| this branch (HEAD) | **GREEN** — `Tests 14 passed (14)` (guard + the existing node ledger guard) |

The two failures at base are the shipped-rule check and the template check, each with its own message naming the stripped attribute (`^data-testid$`). The 3 that pass at base are the config parse, the negative control, and the stamping check — correctly, since both attributes *are* stamped at base.

## Mutation results

Each mutation was applied to the fixed tree, run, and reverted (files restored byte-identical, verified with `cmp`).

| Mutation | Result |
|---|---|
| revert the ledger selector in `globals.css` to the `data-testid` form | 2 failed / 3 passed. `AssertionError: a full-bleed ledger rule in src/styles/globals.css is keyed on an attribute that next.config.mjs REMOVES from the production DOM (^data-testid$)…` — the guard's own message, plus the template check |
| delete `data-app-page-frame=""` from `PageBlockHost.tsx` | **exactly one** test failed, the stamping check: `AssertionError: a full-bleed ledger rule … selects on an attribute that PageBlockHost.tsx does not stamp on its root … expected [ 'data-app-page-frame' ] to deeply equal []`. No neighbouring assertion fired |
| empty the strip list in `next.config.mjs` (`properties: []`) | 4 failed / 1 passed, all with the fail-closed message: `next.config.mjs declares reactRemoveProperties with an EMPTY property list… An empty list would let every ledger selector pass unchecked` |

Each died for its **own** specific message, not a neighbour's.

## Tests run (counts, from run logs — not exit codes)

| Suite | Result |
|---|---|
| new guard + `__tests__/pageBlockHostMaxWidth.test.ts` (node `unit`) | `Tests 14 passed (14)` |
| full AppBlocks node `unit` tier (`src/components/AppBlocks/`) | `Test Files 1 failed \| 46 passed (47)`, `Tests 7 failed \| 741 passed (748)` — the 7 are `__tests__/hiddenBlocks.test.ts`, **pre-existing**: `7 failed (7)` at `origin/main` too, same `TypeError: Cannot read properties of undefined (reading 'clear')`. Untouched by this PR |
| full AppBlocks `component` (browser) tier | `Test Files 40 passed (40)`, `Tests 484 passed (484)` |
| `PageBlockHostMaxWidth.browser.test.tsx` alone | `Tests 11 passed (11)` (11/11 at base too — the tier is blind to this defect by construction) |
| `src/components/Apps/ReviewPreviewFit.browser.test.tsx` (other `app-page-frame` consumer) | `Tests 3 passed (3)` |
| `src/components/Meta/__tests__/viewport-fit-cover.test.ts` (other `globals.css` guard) | `Tests 40 passed (40)` |

## Lint / format

Scoped to the changed paths only — this repo is pre-existing red repo-wide, so no repo-wide claim is made.

- `prettier --check` on all 5 changed files: `All matched files use Prettier code style!`
- `eslint` on the 4 changed TS/TSX files: exit 0, no output

## Files changed

- `src/components/AppBlocks/PageBlockHost.tsx` — stamp `data-app-page-frame=""`; comments updated
- `src/styles/globals.css` — ledger rule re-keyed; HOW-TO-ADD-ONE template re-keyed; a 🔴 note on why `data-testid` must never be used here
- `src/components/AppBlocks/__tests__/ledgerSelectorSurvivesProdStrip.test.ts` — **new** seam guard
- `src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx` — injected selector updated; a note recording that this tier is structurally blind to the production strip and pointing at the guard that is not
- `src/components/AppBlocks/__tests__/pageBlockHostMaxWidth.test.ts` — prose/messages updated to the new selector shape. The membership assertion still fails on **growth and shrink** (it reads ids out of `[data-block-id='…']`, which is unchanged), and the `data-block-id={blockId}` stamping pin is unchanged

## Not verified

The fix has not been observed on the live site — that needs a deploy. The claim proven here is that the shipped selector no longer depends on an attribute the production compiler removes, and that the marker it depends on instead is really rendered.

### DIFF
--- a/docs/features/app-blocks.md
+++ b/docs/features/app-blocks.md
@@ -429,20 +429,33 @@ removed here without that being a deliberate, reviewed change.
 **What actually guards the snippet above.** The CSS block on this page is read by
 `src/components/AppBlocks/__tests__/ledgerSelectorSurvivesProdStrip.test.ts`,
 which parses the strip list out of `next.config.mjs` and fails if the selector
-shown here depends on an attribute production removes. That test runs in the node
-`unit` project, which is a **blocking** CI check — so the specific way this recipe
-went wrong before (the `data-testid` spelling) cannot come back silently.
+shown here depends on an attribute production removes — and parses
+`PageBlockHost.tsx` and fails if the attributes the selector chains are not
+stamped **together on one element**, the other way a compound selector silently
+matches nothing. Both are real checks and both catch the specific ways this
+recipe has gone wrong (the `data-testid` spelling; a half of the selector moved
+onto another box).
+
+🔴 **What that does NOT mean is that the mistake is impossible.** The guard runs
+in the node `unit` project, which on a pull request is **report-only** — the job
+carries `continue-on-error` there, so the check annotates and the run still
+concludes green — and renders a real verdict only on pushes to `main`, after the
+merge. And **no status check is required on `main` in this repo at all**: branch
+protection declares none, so nothing a test does stops a merge. A red guard here
+is a signal a reviewer has to read, not a gate that holds the door.
 
 Two things it does **not** promise. It compares the doc against the compiler
-config; it does not render anything, so it cannot tell you the rule still has the
-visual effect described. And the rule's measured behaviour —
-`src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx` injects a rule
-of this shape at a 2560px viewport and asserts the host goes back to full width —
-runs in the browser `component` project, which reports as the **non-blocking**
-`preview / component-tests` status. Useful evidence, not a gate: it can go red
-without stopping a merge. And the guard covers the CSS **selector** on this page,
-nothing else: that test is the only thing in the repo that reads this file, so
-every other claim here is unverified prose and should be read as such.
+config and against the host's JSX; it does not render anything, so it cannot tell
+you the rule still has the visual effect described. And the rule's measured
+behaviour — `src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx`
+injects a rule of this shape at a 2560px viewport and asserts the host goes back
+to full width — runs in the browser `component` project, which reports as the
+`preview / component-tests` status. Useful evidence, not a gate — and, as above,
+neither tier is one: both can go red without stopping a merge. What this tier
+alone can do is see a rendered width at all. And the guard covers the CSS
+**selector** on this page, nothing else: that test is the only thing in the repo
+that reads this file, so every other claim here is unverified prose and should be
+read as such.
 
 ### Manifest re-publish behavior
 
--- a/docs/features/app-blocks.md
+++ b/docs/features/app-blocks.md
@@ -395,10 +395,21 @@ would be untyped where the host reads it). Instead it is one CSS rule on the
 civitai side, keyed on the `data-block-id` the host stamps:
 
 ```css
-[data-testid='app-page-frame'][data-block-id='your-app-slug'] {
+[data-app-page-frame][data-block-id='your-app-slug'] {
   --app-page-max-width: none;
 }
 ```
+
+🔴 **Never key this rule on `data-testid` — that attribute does not exist in
+production.** `next.config.mjs` sets `compiler.reactRemoveProperties` under
+`NODE_ENV === 'production'`, so every testid is compiled out of the live DOM. A
+rule keyed on one works in every preview and local build and matches **zero
+elements on civitai.com** — the app stays letterboxed and nothing looks broken.
+That is not hypothetical: this ledger shipped that spelling and
+`playable-collections` rendered capped in production while every test tier passed.
+`data-app-page-frame` is the presence marker the host stamps for exactly this
+purpose, on the same element as `data-block-id`; both halves of the selector must
+be on that one element.
 
 Those rules live in the **full-bleed opt-out ledger** in
 `src/styles/globals.css`, next to the `--app-page-max-width` declaration; the
@@ -415,10 +426,23 @@ is unaffected either way. Every entry is expected to carry a reason like this
 one; the ledger's membership is asserted in a test, so a rule cannot be added or
 removed here without that being a deliberate, reviewed change.
 
-The mechanism is measured, not just documented:
+**What actually guards the snippet above.** The CSS block on this page is read by
+`src/components/AppBlocks/__tests__/ledgerSelectorSurvivesProdStrip.test.ts`,
+which parses the strip list out of `next.config.mjs` and fails if the selector
+shown here depends on an attribute production removes. That test runs in the node
+`unit` project, which is a **blocking** CI check — so the specific way this recipe
+went wrong before (the `data-testid` spelling) cannot come back silently.
+
+Two things it does **not** promise. It compares the doc against the compiler
+config; it does not render anything, so it cannot tell you the rule still has the
+visual effect described. And the rule's measured behaviour —
 `src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx` injects a rule
-of exactly this shape at a 2560px viewport and asserts the host goes back to full
-width, so the instructions above cannot rot into something that no longer works.
+of this shape at a 2560px viewport and asserts the host goes back to full width —
+runs in the browser `component` project, which reports as the **non-blocking**
+`preview / component-tests` status. Useful evidence, not a gate: it can go red
+without stopping a merge. And the guard covers the CSS **selector** on this page,
+nothing else: that test is the only thing in the repo that reads this file, so
+every other claim here is unverified prose and should be read as such.
 
 ### Manifest re-publish behavior
 
--- a/src/components/AppBlocks/PageBlockHost.tsx
+++ b/src/components/AppBlocks/PageBlockHost.tsx
@@ -3868,8 +3868,8 @@ export function PageBlockHost({
         // the `data-block-id` opt-out ledger — the custom property is still declared
         // once in globals.css and still overridden per-app on THIS element, from
         // which it INHERITS to the content wrapper. So a ledger entry keyed on
-        // `[data-testid='app-page-frame'][data-block-id='…']` keeps working exactly
-        // as documented, with no change to its selector.
+        // `[data-app-page-frame][data-block-id='…']` keeps working exactly as
+        // documented, with no change to its selector.
         width: '100%',
         // See the `fit` prop for why these are the two modes and why the
         // viewport arithmetic can never agree with its own scroll viewport.
@@ -3892,6 +3892,28 @@ export function PageBlockHost({
             }),
       }}
       data-testid="app-page-frame"
+      // 🔴 THE OPT-OUT LEDGER'S OTHER HALF, AND IT EXISTS BECAUSE `data-testid`
+      // DOES NOT SHIP. `next.config.mjs` sets
+      // `compiler.reactRemoveProperties: { properties: ['^data-testid$'] }` under
+      // `NODE_ENV === 'production'`, so EVERY `data-testid` is compiled out of the
+      // production DOM. The ledger in globals.css used to be keyed on
+      // `[data-testid='app-page-frame'][data-block-id='…']`, which therefore
+      // matched nothing on the live site while passing in every test tier (they
+      // all run with `NODE_ENV !== 'production'`, where the testid is present) —
+      // measured on civitai.com/apps/run/playable-collections: the rule shipped
+      // verbatim in the CSS, 0 elements matched the compound selector, 1 matched
+      // `[data-block-id='playable-collections']`, and the app was letterboxed at
+      // the 1600px cap. This attribute is the production-surviving spelling of
+      // "this is the page host", the same way `data-app-footer` and
+      // `data-adhesive-ad` mark their elements for `globals.css` elsewhere.
+      //
+      // It carries no value on purpose: it is a presence marker, not data. Never
+      // re-key the ledger onto `data-testid` (stripped) and never delete this —
+      // both make every ledger rule inert with nothing visibly wrong. Guarded by
+      // `__tests__/ledgerSelectorSurvivesProdStrip.test.ts`, which reads the strip
+      // list out of `next.config.mjs` and the ledger selectors out of globals.css
+      // and compares them, rather than restating either.
+      data-app-page-frame=""
       // Observable sizing mode, so a regression test (and DevTools) can assert
       // WHICH branch a surface took rather than re-deriving it from computed
       // styles that jsdom does not resolve.
@@ -3963,7 +3985,7 @@ export function PageBlockHost({
           same element, which is precisely the rule shape the opt-out ledger uses — so
           writing the value inline would silently make the opt-out inert while looking
           tidier. The property is set on the ROOT and inherits down to here, so the
-          ledger's existing `[data-testid='app-page-frame'][data-block-id='…']` selector
+          ledger's existing `[data-app-page-frame][data-block-id='…']` selector
           is unchanged by the move.
 
           🔴 IT REPRODUCES THE VERTICAL CHAIN IT WAS INSERTED INTO, which is the only
--- a/src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx
+++ b/src/components/AppBlocks/PageBlockHostMaxWidth.browser.test.tsx
@@ -443,7 +443,8 @@ describe('PageBlockHost — the app stops growing on a wide display', () => {
 
   /**
    * THE DOCUMENTED FULL-BLEED OPT-OUT, end to end: the ledger in `globals.css`
-   * keys on `data-block-id`, which the host stamps.
+   * keys on `data-app-page-frame` + `data-block-id`, both of which the host
+   * stamps on its root.
    *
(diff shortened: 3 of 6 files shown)


ZacxDev:
## Adversarial audit — round 1, fixes applied

Round 1 was a first, full audit. It returned **3 findings, no deploy-blockers**, and independently re-derived the PR's core claim rather than accepting it: running this repo's own `next-swc` binding (Next 16.3.1) with the exact `reactRemoveProperties` config from `next.config.mjs` against a fixture carrying all six attributes, `data-testid` is stripped while `data-app-page-frame`, `data-block-id`, `data-fit`, `data-app-footer` and `data-adhesive-ad` all survive — negative and positive control in the same probe. Selector specificity is unchanged: both the old and new forms are `(0,2,0)`.

The class sweep came back clean — no other CSS rule, and no runtime `querySelector`/`closest`/`matches` outside tests, is keyed on a build-stripped attribute.

### Findings and what was done

**F2 (should-fix) — the stamping guard could not see relocation, and the gating tier was blind to it.**
The check was `!host.includes(attr + '=')`, a whole-file substring search, under an assertion message claiming the attribute was stamped *on the root*. Measured: moving `data-app-page-frame=""` off the host root onto the `app-page-content` wrapper re-creates this PR's exact defect — the compound selector matches zero elements — and the gating `unit` tier stayed byte-identically green at `7 failed | 741 passed`. Only the report-only browser tier caught it. Now fixed by parsing `PageBlockHost.tsx` and asserting that some **single JSX element** stamps every attribute a ledger selector chains, with a spread attribute deliberately not counted so an unreadable element fails closed. The sibling assertion in `pageBlockHostMaxWidth.test.ts` was resolving the frame by text region and pinning only one of the selector's two halves; it now resolves the element and asserts both.

**F3 (nit) — the "gated on a production build" assertion could not fail.**
It sliced everything before `reactRemoveProperties` and asserted the slice contained `process.env.NODE_ENV === 'production'` — satisfied on its own by an unrelated `const isProd = …` 174 lines earlier. Measured: making the strip unconditional left the guard green. Now the `compiler:` conditional itself is parsed, and three separate claims are made — the condition is the production test, the strip is in the branch that condition selects, and the strip is not also in the other branch.

**F1 (should-fix) — the developer-facing HOW-TO still taught the inert selector.**
`docs/features/app-blocks.md` still showed the `data-testid` form, which is how this bug reproduces one commit later, and its anti-rot sentence was made false by this very PR. Re-keyed, warning carried across, and the claim replaced with an accurate per-tier statement.

### Verification

Relocation and unconditional-strip mutants both **survived** the gating tier before these fixes and are **killed** by it after, each on its own assertion's message rather than a neighbour's. A known-caught mutant was kept in the batch as a positive control, and two extra reachability mutants confirm the gate's second and third claims are independently reachable rather than shadowed by the first. Gating `unit` tier scoped to `src/components/AppBlocks/`: `741 → 743` passing, with the 7 pre-existing `hiddenBlocks.test.ts` failures unchanged and unrelated to this PR.

### Correction to this PR's description

The PR body describes the opt-out in terms of `/apps/run/[slug]`. The ledger rule actually applies on **every surface that mounts `PageBlockHost`** with that block id — the moderator review preview and the author dev tunnel as well. This is benign, since the cap only binds above 1600px and it restores what dev builds already did, but the real blast radius is wider than the description implies and a reviewer should have that rather than the narrower version.

### One judgement call worth a reviewer's eye

The production-gate assertion is a verbatim pin on `process.env.NODE_ENV === 'production'`. A semantically-equivalent refactor — hoisting that condition to the existing `isProd` constant — now fails the guard. That is consistent with the file's stated doctrine of pinning the whole statement rather than a substring, and the failure message tells you to re-point the assertion rather than relax it, but it is a deliberate trade and not an oversight.

```audit-claims round=1 audited=4d032c15eaaae526f385c7f6a3d92437a479f131..7bf37d57
1. F2: the stamping check no longer searches the whole file — it parses PageBlockHost.tsx and asserts every attribute a ledger selector chains is stamped on ONE JSX element; the sibling test now resolves the element and asserts both halves on it.
2. F3: the production-gate assertion now parses the compiler: conditional in next.config.mjs and asserts the condition is the production test, the strip is in the selected branch, and the strip is not in the other branch.
3. F1: docs/features/app-blocks.md now teaches the production-surviving selector, its false anti-rot claim is replaced with a per-tier statement, and the gating guard was extended to read the doc's CSS block.
```

ZacxDev:
## Adversarial audit — round 2 (delta), fixes applied

Round 2 was a delta re-audit of round 1's fixes. It returned **three should-fix and one nit, no deploy-blockers**, and confirmed by mutation that round 1's headline fix genuinely holds: relocating the marker off the host root — the mutation that survived round 1 byte-identically green — now goes red in both guard files, each on its own message.

The shipped behaviour has not moved since round 1. `src/styles/globals.css` and `src/components/AppBlocks/PageBlockHost.tsx` are unchanged across both fix rounds and survived every mutation applied to them.

### The important one: round 1's fix introduced a new false claim

Round 1 replaced a false anti-rot sentence in `docs/features/app-blocks.md` with a different false sentence — that the node `unit` project is *"a blocking CI check"* so the `data-testid` spelling *"cannot come back silently"*. Both halves are wrong, verified against the API and the workflow rather than inferred:

- `.github/workflows/lint.yml:405` — `continue-on-error: ${{ github.event_name == 'pull_request' }}` on the `unit` job, under a header reading *"REPORT-ONLY ON PULL REQUESTS, REAL VERDICT ON `main`"*.
- `repos/civitai/civitai/branches/main/protection` — `has("required_status_checks")` is **false**, and the `required_status_checks` endpoint returns **404 "Required status checks not enabled"**.

So no status check is required on `main`, and on a pull request the node tier is report-only. The paragraph also contrasted this against the browser tier, handing the reader a distinction that does not exist. That is the same "reads as coverage while providing none" failure this PR exists to correct, re-introduced one round later by the fix for it — which is the whole reason the delta round exists.

The documentation now states what is true: the guard is real and does catch the `data-testid` spelling; it goes red in a tier that annotates on PRs and renders an honest verdict on pushes to `main`; and a red guard is a signal a reviewer must read, not a merge blocker. The same false phrasing was corrected in `pageBlockHostMaxWidth.test.ts` (pre-existing on `main`, called out in the commit message).

### Two fail-open holes in the guards, both measured

**A mixed strip array was invisible.** The strip-list parse filtered to string literals and silently dropped anything else, so `properties: ['^data-testid$', ...EXTRA_STRIPS]` left the guard 7/7 green — certifying that the shipped rule survives production while production was in fact stripping the marker, i.e. re-certifying the exact defect as fixed. It now asserts every element was readable and names the ones it could not read.

**A doc-side rename shipped green.** The publisher-doc guard checked only whether the documented selector used a stripped attribute, not whether anything stamps it. A one-character typo (`data-app-page-fram`) passed 7/7 while producing the same user-visible outcome as the shipped bug by the other mechanism. The stamped-together check now runs over the doc selector — and over the `globals.css` HOW-TO template, which had the identical hole.

**A nit worth naming:** the production-gate check advertised three assertions, but two earlier assertions consumed both defect shapes the third named, so its message could never print and the only inputs reaching it would have been false positives. It was deleted and the comment narrowed to what is actually asserted.

### Verification

Every mutation above was measured before and after: mixed strip array `7 passed → 5 failed`; doc typo `7 passed → 1 failed`; CSS template typo `7 passed → 1 failed`; and a known-caught mutant retained as a positive control. The mutation copy was re-run green after each revert, so no result rests on a dead runner. Typecheck is clean over the changed paths — and the instrument was watched red first (a worktree submodule gap produced 10 unrelated module errors until the submodule was initialised), so the zero is a real zero.

```audit-claims round=2 audited=7bf37d570d7b44c840eced4be2393c3cdfe73008..3aa7c7f1
1. The documentation no longer claims any tier blocks a merge — it states the node tier is report-only on PRs, renders a real verdict on pushes to main, and that no status check is required on main; the same phrasing was corrected in pageBlockHostMaxWidth.test.ts and two other repetitions.
2. The strip-list parse now asserts every array element was readable and names unreadable ones, so a mixed array no longer certifies the rule as production-safe.
3. The stamped-together check now runs over the publisher doc's selector and over the globals.css HOW-TO template, so a rename or typo in either is caught.
4. The unreachable third gate assertion was deleted along with its now-unread whenFalse field, and the three-assertions comment was narrowed to the two that are actually reachable.
```

ZacxDev:
## Adversarial audit — round 3 (delta), fixes applied, and the ladder is closed here

Round 3 returned **one should-fix and one nit, both scaffolding, no deploy-blockers**, and re-verified all four of round 2's claims against the diff rather than the summary. Each of round 2's mutations reproduced at the exact claimed counts, including the fail-open numbers.

It also verified two things more carefully than the earlier rounds: branch protection via **both** mechanisms — classic protection returns `has("required_status_checks") = false`, and the only ruleset on `main` is a `deletion` rule — and the `Unit tests` shards' `steps[].conclusion` rather than the job conclusion, so a `continue-on-error` mask could not be mistaken for a pass. The rewritten gating paragraph checked out as true on every axis, including the `push: branches: [main]` trigger.

### What round 3 found

**The recurring shape was still present at six sites inside a file this PR had already partly corrected.** `pageBlockHostMaxWidth.test.ts` carried *"THE ULTRAWIDE CAP — the GATING half"* in its own title, twenty lines above the paragraph this PR rewrote to say neither tier blocks a merge — a self-contradiction inside one docblock. Two of the six were **assertion messages a developer reads at failure time**, including *"it is the only BLOCKING check on the ultrawide cap"*, which would tell someone renaming the constant that a merge is held when nothing of the sort is true.

All are now corrected, and the sweep was widened past the reported set — a sixth site (*"the one regression this file exists to block"*) was found by grepping rather than by the finding. The instrument was run before and after: 8 hits → 4, and each of the 4 survivors is a true statement (one is a substring match on "navi**gating**").

**The template assertion misattributed shipped-rule violations to the HOW-TO template**, so a developer could have been sent to edit a correct template. Fixed structurally rather than by rewording: a new `commentsOnly()` — the exact complement of the existing `stripComments()` — makes the two scopes partition `globals.css` instead of overlapping. Side benefit: a deleted template can no longer be masked by a shipped rule satisfying the non-empty check. The new control was itself mutation-tested and dies on its own message, on a reachable input.

### Why the ladder ends here rather than continuing

Not because a round came back clean — round 3 found real things. It ends on the stated criterion for a ladder whose payload is prose, and the rationale is recorded here deliberately so this is distinguishable from a converged ladder:

- **No deploy-blockers**, in this round or since round 1.
- **Blast radius is a false sentence in a test comment** — both findings are scaffolding with no runtime effect.
- **The recurring shape is now swept at every site in the files this PR touches**, which was the outstanding clause.

The self-reinforcing pattern is explicit: round 2's findings were about prose round 1 wrote, round 3's about prose round 2 wrote, and a round 4 would audit prose round 3 wrote. Meanwhile the shipped fix has been **frozen since round 1** — `src/styles/globals.css` and `src/components/AppBlocks/PageBlockHost.tsx` are byte-unchanged across all three fix rounds — and the relocation mutation that defeated round 1 outright is now caught by **four** assertions across both guards, each with a correct message.

### Deliberately NOT fixed — open, not absent

**The same false gating phrasing exists in eight other AppBlocks test files** and was left alone to avoid ballooning this diff: `chromeCrumbLinkStyle`, `chromeGeometry`, `chromeItemClamp`, `chromeListingQueryIsSingleSourced`, `chromeNavAlignsWithSubNav`, `featureFlagsMockCompleteness`, `iframeAwareMenu`, `pageRunScrollContract`. Several assert the opposite of what is true — that the node tier "can block a merge". Closing condition is mechanical: `grep -rniE 'gating|block(s|ing)? a merge' src/components/AppBlocks/` returns only true phrasings. (`shouldStartInit.test.ts` also matches, but uses "gating" for init logic — a different and legitimate sense.)

**Two counts are unpinned by construction:** `24,879 tests` / `1569 files` in `pageBlockHostMaxWidth.test.ts`. Nothing asserts on them, and a full-suite run to confirm them was out of budget. They were written by the merged #4577, not by this PR, so they are not stale-by-a-later-commit — but they are quoted, not checked.

**The fix has not been observed on the live site.** That needs a deploy. The post-deploy check is one probe: `--app-page-max-width` must resolve to `none` on `playable-collections`, and its iframe must span the viewport rather than 1600px.

```audit-claims round=3 audited=3aa7c7f1cedaee00bd9f03c30fb7bf248453afc3..f0dc568e
1. Every "gating"/"blocking" phrasing in pageBlockHostMaxWidth.test.ts was swept — six sites including the file title and two failure-time assertion messages — and the remaining matches in this PR's files are true statements.
2. The template assertion now inspects only the comment scope via a new commentsOnly() complement to stripComments(), so a shipped-rule violation is no longer reported as a template violation.
3. The honest-verdict wording now includes workflow_dispatch alongside pushes to main.
4. The ladder is closed on the prose-payload criterion, not on a clean round; the eight remaining files carrying the false phrasing are recorded above with a mechanical closing condition.
```

PR #4595, opened 2026-09-03
fix(jobs): bump Model."updatedAt" when a sweep flips a model to Draft

ZacxDev:
## The defect

`src/server/jobs/reset-to-draft-without-requirements.ts` (cron `43 2 * * *`) ends with a raw-SQL statement:

```sql
UPDATE "Model" m
SET status = CASE ... THEN 'Draft' ... ELSE 'Unpublished' END,
    meta = jsonb_set(...)
WHERE ...
```

It is `$executeRaw`, so **Prisma's `@updatedAt` does not fire** and `Model."updatedAt"` keeps its old value.

`src/server/jobs/remove-old-drafts.ts` reaps `status IN ('Draft','Deleted') AND m."updatedAt" < now() - INTERVAL '30 days'` and cascade-deletes the model with its versions, files and training data — irreversibly.

So a model this sweep flips to `Draft` gets **zero grace period**: its clock is already ancient the moment it enters the reapable state.

### Why this is real, not theoretical

Measured against the prod read replica on 2026-09-03, over models the sweep left in `Draft` with `meta->>'unpublishedReason' = 'no-versions'`:

| | |
|---|---|
| total | **1,319** |
| already past the reaper's 30-day clock | **1,309** |
| reapable today under the live #4579 fence | **0** |
| 🔴 saved **only** by the reaper's `downloadCount >= 10` term | **1,308** |

Positive control that those counts are not a filtered-to-nothing artifact — the same `no-versions` marker across statuses: Unpublished 3,540 / Deleted 1,554 / Draft 1,319 / Published 1,158 / UnpublishedViolation 3.

That `downloadCount` term is an **accidental** protection: undocumented, unrelated to abandonment, and self-gameable. The #4579 activity fence does not cover these models, because a swept model has no recent version-or-file activity — the sweep is a *system* write, not creator activity. Any change to the `downloadCount` term turns 1,308 rows into an immediate mass-deletion path. The oldest swept model's `updatedAt` is **2023-10-17**.

Ordering: last night the reaper deleted at `02:43:15Z` and the sweep stamped `unpublishedAt` at `02:43:23Z` — 8 seconds later, i.e. the reaper happened to go first, the *lucky* ordering. That is one observation, not a proven interleaving, and nothing guarantees it.

### 1. The bump — two raw-SQL statements

`"updatedAt" = now()` added to:

- `src/server/jobs/reset-to-draft-without-requirements.ts` — the final `UPDATE "Model" m`. Unconditional across both CASE branches: the `Unpublished` rows are later flipped to `Draft` by the backfill below and would otherwise carry a stale clock forward into the same hole.
- `src/pages/api/admin/temp/backfill-swept-trained-models.ts` — the `UPDATE "Model" m SET status = 'Draft'`. **Not** the `UPDATE "ModelVersion"` above it.

### 2. 🔴 The consequence that had to be fixed with it — the `old-draft` notification

`src/server/notifications/model.notifications.ts` `'old-draft'` selects `status IN ('Draft') AND m."updatedAt" BETWEEN lastSent-23d AND now()-23d`, and its message is *"will be deleted in 1 week"* with `toggleable: false`.

Before this PR those 1,308 models could never match — their clocks sat **below** the BETWEEN lower bound. **The bump makes them eligible for the first time.** And the notification carried *none* of the reaper's other terms, so it would have promised imminent deletion to ~1,308 creators of whom ~99% are held back by `downloadCount >= 10` — a false, unmutable warning, all firing in one minute 23 days out.

So the predicate now carries the reaper's **value** terms: `availability != 'Private'` and the download term. The download term is an `EXISTS` over `"ModelMetric"` rather than a JOIN, which reproduces the reaper's INNER `JOIN` + `DISTINCT` semantics — **a model with no metric row is not reapable, so it must not be warned either.** The reaper's *activity* terms are deliberately **not** mirrored; section 3 is why.

🔴 **This is an APPROXIMATION, not an exact prediction.** The notification evaluates at day 23; the reaper acts no earlier than day 30 and possibly much later (section 3). In the gap `downloadCount` can cross 10, so a model warned here may be spared. The precise error directions, and why they are one-way, are in section 3.

### 3. 🔴 Why the notification has NO activity fence — two defects this PR shipped and then removed

The reaper carries two `NOT EXISTS` clauses over recent `ModelVersion` / `ModelFile` activity. Two successive revisions of this PR tried to mirror them, and **both were wrong**. The fences are now **gone**, not re-tuned.

**Attempt 1** copied them verbatim, `now()` anchor included. The two queries run 7 days apart, so identical `now()`-relative text is a different absolute window on each side — the notification's resolved to `U − 7d` against the reaper's `U`.

**Attempt 2** re-derived the interval as `REAP_AGE_DAYS − OLD_DRAFT_LEAD_DAYS`. That is correct **only if the reaper is a one-shot evaluation at `U + 30d`**. It is not. It is a nightly cron that **retries forever**, so it really fires at:

```
max(Model."updatedAt", latest version/file activity) + REAP_AGE_DAYS
```

while the notification evaluates **once**, in a ~1-minute band at `U + 23d`, and is never re-evaluated for that `U`.

So attempt 2 still excluded every model whose version or file landed *after* the model row was last written — which `remove-old-drafts.ts` documents as the **norm** ("the finished resource lands hours or weeks later"):

| | |
|---|---|
| model row last written | Jan 1 (`U`) |
| version + file land | Jan 3 |
| notification runs Jan 24 | cutoff Jan 1 → activity Jan 3 is newer → **excluded**, never re-evaluated |
| reaper first fires Feb 3 | cutoff Jan 4 → activity Jan 3 is older → fences clear → **cascade-deleted, unwarned** |

On `origin/main` that model *was* warned. Verified across offsets — with attempt 2's predicate, `U+0` is warned and `U+0.5`, `U+2`, `U+20`, `U+40` are all excluded-then-reaped. `U+0` is the single offset the round-3 guard pinned, which is exactly why it passed.

**No choice of interval fixes a predicate evaluated at one instant against a condition that keeps moving.** Hence removal.

**What stays:** `status IN ('Draft')`, the `BETWEEN` band, `availability != 'Private'`, and the `downloadCount < 10` `EXISTS`.

**Why that costs nothing:** every one of the ~1,308 false alarms this predicate was tightened to prevent came from `downloadCount >= 10`. The activity fences never contributed to the problem this PR set out to fix.

**Why the value terms may be mirrored when the fences may not** — they do not depend on predicting *when* the reaper fires. `availability` is genuinely safe; `downloadCount` is an **assumption with a named failure mode**. Both are set out in section 5.

**Net:** outside the two exclusions named in section 5, errors are one-directional — a model may be warned and later spared, or warned earlier than the message implies (the reaper's retries can push the real gap past `OLD_DRAFT_LEAD_DAYS`).

The constants module `src/server/common/draft-reaping.ts` stays and is still the single source for the band and the message copy. `OLD_DRAFT_NOTICE_DAYS` derives from `REAP_AGE_DAYS` — deliberately not from `ACTIVITY_WINDOW_DAYS`, since the band keys on `Model."updatedAt"`, the very column the reaper's age threshold tests. It is now the **only** interval in the query, so the "both are 30, so you cannot tell which you built on" hazard has no remaining site. Its dependency-free shape is load-bearing: `model.notifications.ts` is in a graph `no-server-infra-in-app-graph.test.ts` forbids from reaching `src/server/db/`, so it cannot import the job.

### 4. The guard: a CLOSED allowlist, not a deny-list of spellings

🔴 **Parity was the wrong invariant** — requiring the notification to *mirror* the reaper's terms is what put a time-relative clause into a query that runs on a different schedule. The property is one-directional, and it has two named exclusions:

> **For a model in `Draft`, no term in the notification may exclude a model the reaper will destroy — except via `status` and `downloadCount`, both documented below.**

🔴 **An earlier revision expressed the structural rule as three DENIALS** — no `now()`-relative comparison, no `NOT EXISTS`, no reference to `"ModelVersion"`/`"ModelFile"`. Each is individually live, but **a deny-list of spellings is unbounded by construction**, and this genuine activity fence passed all 17 tests:

```sql
AND (mm."lastVersionAt" IS NULL
  OR mm."lastVersionAt" <= now() - INTERVAL '23 days')
```

It evades the regex because `<=` puts `=` where the pattern wants a space, and it reads `ModelMetric."lastVersionAt"` — a real maintained column on the one table the guards whitelisted. Two cruder rewrites (`NOT (EXISTS (…`, `0 = (SELECT count(*) …`) also survived.

`old-draft-reaper-parity.test.ts` (15 tests) now pins **closed allowlists**:

- the query's table set must be **exactly** `{"Model", "ModelMetric"}`;
- its qualified column references must be **exactly** the eight it uses.

Adding a term now fails until someone adds it to the allowlist deliberately — which is the reviewer's cue to ask whether it can exclude a reapable model. Both extractors carry a positive control so a regex that stops matching cannot turn the assertions into `[] === []`. The two spelling guards are kept as *secondary*, for their specific failure messages, and are no longer relied on as the closure.

🔴 **The parameterised offset test from the previous round is DELETED, not repaired.** At HEAD the query has no time-relative term, so its loop body never executed and its only live assertion was `U + OLD_DRAFT_NOTICE_DAYS < max(U, activity) + REAP_AGE_DAYS` — true by construction for *any* query, since `NOTICE = REAP − LEAD` and every offset is `>= 0`. **It could not fail**, it was strictly subsumed by the structural guard it iterated, and it missed the walk-past above entirely. A test that cannot fail reads as coverage and stops people looking, which is how the previous two defects survived. Its worked example survives as documentation on the guard; the allowlist carries the property.

*(That test also modelled the reap instant as `max(U, activity) + REAP_AGE_DAYS`. The true instant is `max(U + REAP_AGE_DAYS, activity + ACTIVITY_WINDOW_DAYS)`; the two coincide only because both constants are 30 — the exact conflation `draft-reaping.ts` exists to forbid. Corrected wherever that model is now stated.)*

### 5. 🟡 The invariant's two exclusions, now stated in code and pinned

Two comments asserted absolutes that were false in the dangerous direction.

**(a) `status`.** The reaper destroys `('Draft', 'Deleted')`; the notification warns on `Draft` only, so **a `Deleted` model is cascade-deleted unwarned**. That is deliberate — a user who deleted a model has already expressed the intent, so this is not the same safety case as an abandoned draft vanishing — but it was disclosed only in a PR comment, with nothing in the code or the guards saying it. It is now **pinned by a test** that asserts the notification's status set is `['Draft']`, the reaper's is `['Deleted','Draft']`, and the difference is exactly `['Deleted']`. **Not widened** — that would be an unapproved behaviour change.

**(b) `downloadCount` is an ASSUMPTION, not a property**, and "in practice" is gone. It has **no incrementing writer** — every write is a full recompute (`model.metrics.ts`, a `SUM` over surviving `ModelVersionMetric` rows) — so it is **decreasing-capable by construction**. The concrete in-repo failure path is now recorded in the source:

> A Draft model rolled up to `downloadCount = 12` is excluded at day 23. A version carrying 8 downloads is then deleted; `deleteVersionById` calls `updateModelLastVersionAt`, which **early-returns when the model has no `Published` version** — true for a Draft — so `Model."updatedAt"` is not bumped and the band never re-arms (its lower edge is `lastSent − notice`, so there is no second evaluation). A later recompute yields 4, the reaper's `< 10` becomes true, and the model is cascade-deleted **unwarned**.

That is architectural — the band gives exactly one evaluation per `U` — so it is **recorded, not patched**.

By contrast `availability` is genuinely safe, and this is now argued rather than assumed: its only raw-SQL writer, `entityAvailabilityUpdate` (`src/server/services/common.service.ts`), sets `"updatedAt" = NOW()` in the same statement, so a change bumps the band's own key and re-arms it.

## 🔴 Correction: why `ModelVersion."updatedAt"` is left alone

An earlier revision claimed bumping `ModelVersion."updatedAt"` on a system write is forbidden, citing the "not a creator edit" comment in `model-version.service.ts`. **That was wrong.** That comment covers one narrow case (a blurb re-materialization). The nearest analogue does the opposite: `unpublishModelById` (`src/server/services/model.service.ts`) is a system/moderator take-down that raw-SQL-updates `"ModelVersion"` and sets `"updatedAt" = NOW()` **deliberately**, because the column is on the public v1 payload via `src/server/selectors/modelVersion.selector.ts` (returned by `src/pages/api/v1/model-versions/[id].ts`).

The guard **stays** — it pins *this* change's scope — but it no longer asserts a universal the codebase contradicts, and its comment tells anyone making that decision on purpose to delete the test rather than work around it. This round also retracts the **same wrong rationale in `backfill-swept-trained-models.ts`**, which still carried it after the previous round corrected only its sibling; the two files were giving opposite reasons for the same omission.

Also corrected: the earlier claim that bumping `mv."updatedAt"` would "permanently spare" a model was false. Both clocks would expire at the same instant, so the net effect on the reaper is zero.

## Test coverage

| file | tests | what it guards |
|---|---|---|
| `src/server/jobs/__tests__/reset-to-draft-without-requirements.test.ts` | 7 | the job's `UPDATE "Model"`: the bump, the SET-list ledger, surviving status/meta behaviour, the ModelVersion scope guard |
| `src/server/notifications/__tests__/old-draft-reaper-parity.test.ts` | 15 | the closed allowlist, the two named exclusions, value-term parity, the derivation control |
| `src/server/services/__tests__/no-unbumped-draft-status-write.test.ts` | 4 | both raw-SQL draft-writing sites, as a `no-*` convention guard |

Wired into `test:lint-rules` with both mirroring docs updated, so `no-lint-rules-script-drift` stays green.

### Red → green matrices — all re-derived, file sets named

**A — the original defect.** Files: `reset-to-draft-without-requirements.test.ts` (7) + `no-unbumped-draft-status-write.test.ts` (4) = **11**. Job and backfill sources from `origin/main`:

| tree | result |
|---|---|
| `origin/main` sources | **5 failed \| 6 passed (11)** |
| HEAD | **11 passed (11)** |

**B–E — all against `old-draft-reaper-parity.test.ts` alone (15 tests).** Each row is the notification query replaced by the named shape:

| # | notification shape | result | notes |
|---|---|---|---|
| B | round-2 fences (verbatim `'30 days'`) | **4 failed \| 11 passed** | both tables + columns + both spelling guards |
| C | round-3 fences (derived interval) | **4 failed \| 11 passed** | the shape that passed the round-3 guard completely |
| D | `origin/main` notification (no value terms) | **5 failed \| 10 passed** | allowlist + both value terms + derivation control |
| E | 🔴 **the `lastVersionAt` walk-past** | **2 failed \| 13 passed** | **passed all 17 tests before this round** |
| — | HEAD | **15 passed (15)** | |

E is the required proof. Its killing message:

```
AssertionError: a column outside the allowlist may be time-varying — e.g.
ModelMetric."lastVersionAt" is a real maintained column that turns this query
into an activity fence without naming another table. Add it here only after
proving it cannot exclude a model the reaper will destroy.
  expected [ 'm.availability', 'm.id', …(7) ] to deeply equal [ …(6) ]
```

**Green at HEAD**, the exact 9 files run:

```
src/server/jobs/__tests__/reset-to-draft-without-requirements.test.ts        7
src/server/services/__tests__/no-unbumped-draft-status-write.test.ts         4
src/server/notifications/__tests__/old-draft-reaper-parity.test.ts          15
src/server/jobs/__tests__/remove-old-drafts.test.ts                         20
src/server/jobs/__tests__/remove-old-drafts-activity-filter.test.ts         13
src/server/services/__tests__/no-lint-rules-script-drift.test.ts            11
src/server/services/__tests__/no-server-infra-in-app-graph.test.ts           5
src/server/notifications/__tests__/notifications.block-filtering.test.ts    25
src/server/notifications/__tests__/notification-settings-polarity.test.ts   13
                                                                    total 113
```

**113 passed (9 files).** No test in this repo exercises `backfill-swept-trained-models`, `send-notifications` or `notification-cursor-monitor` behaviourally — the `no-*` guard reads the first as source text only.

### Mutation table — re-derived, 31 mutants

Each applied to a **pristine** copy, never stacked, substitution asserted to land first. **30 KILLED, 1 SURVIVED, 0 ERROR.**

**Walk-past shapes — the ones the previous round's deny-list let through** (15 tests):

| # | mutant | failed | killed by |
|---|---|---|---|
| W1 | `mm."lastVersionAt" <= now() - …` fence | 2 | column allowlist + the `now()` spelling guard |
| W2 | `NOT (EXISTS (…` over `"ModelVersion"` | 4 | table + column allowlists + both spelling guards |
| W3 | `0 = (SELECT count(*) …` over `"ModelVersion"` | 3 | table + column allowlists + the `now()` guard |
| F1 | re-add fences, round-2 shape | 4 | table + column allowlists + both spelling guards |
| F2 | re-add fences, round-3 shape | 4 | same |

**The named exclusions** (15 / 35 tests): **X1** widen the notification to `Deleted` → 1 failed (*warns on Draft only…*); **X2** narrow the reaper to `Draft` → 2 failed (whole-predicate pin + the status test). Both directions of the divergence are now pinned.

**Value terms, band and derivation** (15 tests): **V1** drop `availability` (2), **V2** drop `downloadCount` (3), **V3** band hardcoded `'30 days'` (2), **V4** band hardcoded `'23 days'` — *equals the constant today* — (1, derivation control only), **V5** message copy hardcoded `"1 week"` (1, derivation control only), **V7** `REAP_AGE_DAYS` 30→45 (2).

**Reaper side** (35 tests): **R1** drop `availability` (2), **R2** drop `downloadCount` (2), **R3** age literal 30→45 (3).

**The job's `UPDATE "Model"`** (7 tests) — all 12 re-run, all still killed: M1 (3), M2 (2), M3 (2), M4 (1), M5 (2), M6 quoted (1), M6b **unquoted** (1), M7 positive control (1), M8 harness control (3), M9 `"publishedAt"` (1), M10 `"createdAt"` (1), M11 offset bump (1).

**Backfill ledger** (4 tests): B1 (1), B2 harness control (1).

🔴 **V6 (`OLD_DRAFT_LEAD_DAYS` 7 → 10) SURVIVED, and that is correct.** Changing the lead is the *supported* way to move the schedule: the band, the copy and every assertion shift together and the result stays coherent, so nothing should fail. Pinning the lead to 7 would be a pure change-detector. Not claiming a kill I did not get.

**V4 and V5 are killed only by the derivation control** — they are literals that happen to equal the constants' current values, which is exactly why that control exists.

### 🔴 Limits of this coverage

**There is no DB-backed vitest project in this repo.** Nothing here executes SQL. The allowlist is a **static reading of the query text**: it establishes which tables and columns the query touches, not what Postgres does with them — it cannot see a wrong operator, a mis-nested `EXISTS`, or a wrong join condition among the allowed columns. The value-term assertions are **spelled** string comparisons, so an equivalent reword fails them while behaving identically. The `no-*` guard is a **ledger of two named files**, not a repo-wide scan.

Two behaviours are asserted only in prose, and the guards cannot check either: that `downloadCount` recomputes can decrease, and that the reaper's cron retries. If the second is ever wrong, the reasoning that removed the fences is wrong with it.

Instruments negative-controlled: `prettier --check` and `eslint` have each gone red on this PR's own inputs in earlier rounds; `tsc` is run separately because esbuild does not typecheck. Final typecheck: 31 errors, all in `packages/*` from the symlinked `node_modules`, **0 in `src/`**.

### 🟡 Nothing can detect this notification going silent — and the monitor exemption is not the lever

`notification-cursor-monitor.ts` puts `last-sent-notification-model-old-draft` in `IGNORED_CURSORS`. Removing it **buys nothing**: in `send-notifications.ts`, `await setLastSent()` runs **outside** the `if (!isEmpty(pendingData))` branch, so the cursor advances every minute whether or not any rows were selected. A predicate selecting zero rows forever is indistinguishable from a healthy one; the monitor only sees a query that errors or times out. Removing the exemption would add noise without detection, so it stays.

## Second-order check: can a bumped `updatedAt` surface these models publicly?

**No.** Verified rather than assumed.

- `src/server/services/model.service.ts` — only 4 `orderBy: updatedAt` sites and **zero** raw `ORDER BY … updatedAt`: `getDraftModelsByUserId` (self-only, `status notIn [Published, Deleted]`), `getTrainingModelsByUserId` and `getRecentlyRecommended` (order `ModelVersion`, not `Model`), `getAvailableModelsByUserId` (`status IN [Published]`, so a swept model *leaves* it). All `protectedProcedure`.
- `getModelsRaw` — the actual public listing. Its `orderByRaw` set is `mm."lastVersionAt"`, `thumbsUpCount`, `downloadCount`, `commentCount`, `collectedCount`, `imageCount`, `ci."id"`. **`updatedAt` never appears**, and `ModelSort` has no `RecentlyUpdated` member. Non-moderator requests are forced to `status = 'Published'`.
- Public REST `/api/v1/models` → `runModelSearch` → `getModelsRaw`, same gate.
- `sitemap-models.xml` — aliases `COALESCE(m."lastVersionAt", m."publishedAt") AS "updatedAt"`; never reads the real column.
- `src/server/metrics/model.metrics.ts` — cursors read `ResourceReview`/`BuzzTip`/`ModelVersionMetric` `updatedAt`, never `Model."updatedAt"`.
- `mention.notifications.ts` — reads `m."updatedAt"` but is `status = 'Published'`-gated.
- Models search index — the delta cursor **does** key on `Model."updatedAt"`, but the same predicate carries `status = 'Published'`, so the bump does not re-enqueue swept models; `pullData` re-applies the status filter and `pushData` has no delete-on-miss. `Model.updatedAt` is not in `modelsSortableAttributes`.

The one public render of the column is the model **detail** page ("Updated: …") — a detail page for a model that is no longer Published, not a listing or feed.

Not exhaustively checked: client-side JS re-sorting after fetch, and ClickHouse-backed feeds beyond a column-name grep of `src/server/`.

## Known limits and siblings (documented, deliberately not fixed here)

- **The backfill endpoint is not idempotent across runs, and this PR does not change that.** `draftedModelIds` is populated only from the version UPDATE's `RETURNING`, which re-asserts `mv.status = 'Unpublished'`. So models drafted by a **prior** run of this endpoint cannot be repaired by re-running the fixed one — they join the un-backfilled cohort silently. Pre-existing, not introduced here; flagged so it is not a surprise.
- **`restoreModelById` (`src/server/services/model.service.ts`) has the same defect class.** It raw-SQL-sets `status = CASE WHEN "publishedAt" IS NULL THEN 'Draft'` with **no bump**, so an un-deleted model lands in `Draft` with a clock older than its own deletion. Pre-existing and outside this PR — it needs its own change and its own review. It is named in the `no-*` guard's doc comment as a known sibling deliberately kept out of the ledger, so that guard does not become permanently red.
- **`backfill-swept-trained-models.ts` still documents itself as "idempotent"** (in its header comment and again near the apply branch), which the non-idempotency bullet above contradicts. Pre-existing wording, outside this PR's range; flagged rather than edited.
- **This PR does not backfill the 1,309 models already carrying a stale clock.** It prevents new ones. Those rows keep whatever `updatedAt` they have; grace-ing them is a separate one-off write.

## Scope

`republish-orphaned-drafts.ts` and `process-scheduled-publishing.ts` set `Published`, not `Draft`, so they are out of scope.


ZacxDev:
## Adversarial audit — round 1 (full), fixes pushed

Round 1 was a blind full audit against `c05f40bd28`. **No 🔴**; five 🟡 and three 🟢.
All are addressed in `dab517d4fc`, except two documented as out-of-scope siblings.

**The one that could not ship as-is (F5).** The `updatedAt` bump routed ~1,308 swept models
into the non-toggleable *"will be deleted in 1 week"* notification, whose query carried none of
the reaper's other terms — so it was false for ~99% of them, firing in a single one-minute run
23 days out. The notification predicate now mirrors the reaper. Stated as an **approximation,
not a prediction**: it evaluates at day 23 and the reaper acts at day 30, so `downloadCount`
can cross the threshold and activity can appear in between; the errors now land on the safe
side (warn, then spare).

**A correction to a rule I asserted, which was wrong (F1).** The claim that
`ModelVersion."updatedAt"` must never be bumped on a system write was over-generalised from a
comment about blurb re-materialization. `unpublishModelById`
(`src/server/services/model.service.ts:3350-3372`) — a system/moderator take-down — bumps it
**deliberately**, because the column is on the public v1 payload via `modelVersion.selector`
and `/api/v1/model-versions/[id]`. The guard survives as a **scope** guard for this PR, and its
comment now tells a maintainer making that change on purpose to delete the test rather than
work around it.

**F2** — the claim that a `mv."updatedAt"` bump would "permanently spare" a model was false and
is removed. Both clocks expire at the same instant, so the net effect on the reaper is zero;
the sweep cannot renew it either, since all three of its statements require `status='Published'`.

**F3** — added `no-unbumped-draft-status-write`, a source-scanning guard in the repo's existing
`no-*` convention (the middleware objection did not apply — ~30 such tests already read files
off disk). **F6** — the SET clause is now pinned as a ledger, so an *added* column is caught:
injecting `"publishedAt" = now()` or `"createdAt" = now()` previously survived all six tests.

🔴 **Predicate duplication, and what guards it.** Mirroring the reaper into the notification
open-codes one rule in two files, which is the shape that drifts silently. A shared constant was
rejected with a reason — the reaper is a Prisma tagged template with an INNER `JOIN`, the
notification a plain string with `EXISTS`, and their windows differ on purpose (23 vs 30), so a
fragment could only cover part of it. Instead `old-draft-reaper-parity.test.ts` asserts the
**relationship**: shared terms present on both sides, fence and interval counts equal, so it
fails when the set grows *or* shrinks. Its blind spot is documented — a new reaper term that is
neither a fence nor a 30-day interval moves no count here.

**Red → green**, sources restored from `origin/main` with the guard files at HEAD:
**12 failed | 13 passed (25)** across all three guard files → **25 passed (25)** at HEAD;
69 passed / 6 files including untouched siblings.

**Mutation: 22 mutants, 22 killed, 0 survived**, re-derived from scratch rather than patched.
Two anchors had drifted and reported ERROR (not a false SURVIVED) and were fixed and re-run;
every mutant was re-run after prettier reflowed the parity test rather than trusting pre-reflow
numbers.

**F7 — a correction to this PR's own earlier mutation table.** It claimed "M2 and M3 kill
different assertions, so neither is redundant". Re-derived: M2 fails two tests including the
first test's `toContain`, so it *is* redundant with it; M3 also fails two. The uniquely
attributable mutants are M4, M6, M7, M9, M10, M11, N1, N2, B1, B2. Every mutant dies either
way — only the attribution over-claimed.

**Known limits and siblings, not fixed here**

- **F4** — `backfill-swept-trained-models`'s `RETURNING` re-asserts `mv.status = 'Unpublished'`,
  so models drafted by a PRIOR run cannot be repaired by re-running the fixed endpoint; they
  join the un-backfilled cohort silently. Pre-existing non-idempotency.
- **F8** — `restoreModelById` (`model.service.ts:1910-1921`) raw-SQL-sets
  `status = CASE WHEN "publishedAt" IS NULL THEN 'Draft'` with no bump, so an un-deleted model
  lands in `Draft` carrying a clock older than its deletion. Same defect class, pre-existing,
  needs its own PR and audit. Deliberately excluded from the new guard's ledger — including it
  would make the guard permanently red.
- **This PR prevents new stale clocks; it does not backfill the 1,309 that already exist.**
  Those remain protected only by `downloadCount >= 10`.
- 🔴 **There is no DB-backed vitest project in this repo.** Every guard here is a spelled string
  comparison over SQL text. They prove the statements have not silently changed; they prove
  **nothing** about how Postgres executes them. Measured: a semantic break preserving the
  anchors (`… AND false`) survives the whole suite.

**Ledger** · round 1 · payload lines changed this round: 52 — `model.notifications.ts` 36
(behavioural) + `reset-to-draft-without-requirements.ts` 16 (comments); scaffolding 324.

```audit-claims round=1 audited=c05f40bd28..dab517d4
1. The old-draft notification predicate now mirrors the reaper's terms, and is described as an approximation rather than a prediction.
2. `old-draft-reaper-parity.test.ts` asserts the reaper/notification relationship and fails when the shared term set grows or shrinks.
3. The ModelVersion guard no longer asserts a universal rule; it pins this PR's scope and names `unpublishModelById` as the deliberate counter-example.
4. The "permanently spares" claim is removed from the test comment and the PR body.
5. The SET clause is pinned as a ledger, so an added column is caught, not only a removed one.
6. `no-unbumped-draft-status-write` guards the backfill statement, wired into `test:lint-rules` and the docs that mirror its counts.
7. The mutation table was re-derived from scratch; the earlier per-assertion attribution claim is corrected.
8. No change was made to what the sweep selects, or to `ModelVersion."updatedAt"`.
```

ZacxDev:
## Adversarial audit — round 2 (delta): **needs rework**, now fixed in `7cdc90629b`

Round 2 audited `c05f40bd28..dab517d4fc`. It confirmed 7 of 8 round-1 claims outright, and found
**one 🔴 that the round-1 fix itself introduced**, one 🔴 in the guard added alongside it, and
three 🟡.

### 🔴-1 — the fences were evaluated 7 days too early, so the mainline reaped population was warned NEVER

The three fence clauses were copied from the reaper verbatim, `now()` anchor included. But the
two queries run 7 days apart, so a `now()`-relative window is **not the same predicate** on both
sides. Verified arithmetically, independently, before any fix:

- notification fires at `U + 23d`; with `INTERVAL '30 days'` its fence window resolves to **U − 7d**
- reaper acts at `U + 30d`; its window resolves to **U**

The notification's condition was therefore strictly **stricter**. For the canonical abandoned
draft — model, version and file all created at `U`, which `remove-old-drafts.ts:136-144`
describes as the norm — activity at `U` is newer than `U − 7d`, so the row was **excluded and no
warning sent**; seven days later the reaper's window is `U`, the fence clears, and the model,
its versions, files and training data are **cascade-deleted irreversibly**. Before this PR that
model *was* warned. The band is ~1 minute wide, so a miss is permanent.

**We replaced false alarms with silence, which is worse.** Fixed by deriving the anchor rather
than hardcoding it.

### 🔴-2 — the parity guard asserted textual identity, and so blocked its own repair

It required the notification to spell `INTERVAL '30 days'` verbatim — encoding the defect above.
Demonstrated, not argued: running the old guard against the corrected source gives
**5 failed | 9 passed (14)**. Textual equality is the wrong invariant between two queries
evaluated at different instants. It now compares **resolved windows**, and the headline test
simulates the canonical abandoned draft against the numbers both queries actually carry, with a
fixture-validity assertion (*"the reaper must DESTROY this model"*) so it cannot pass vacuously.

### The fix, and one deliberate deviation

A new dependency-free `src/server/common/draft-reaping.ts` holds `REAP_AGE_DAYS`,
`ACTIVITY_WINDOW_DAYS`, `OLD_DRAFT_LEAD_DAYS` and
`OLD_DRAFT_NOTICE_DAYS = REAP_AGE_DAYS - OLD_DRAFT_LEAD_DAYS`. The `BETWEEN` band, all three
fences and the message copy derive from that one expression.

The instruction was to import `REAP_AGE_DAYS` from `remove-old-drafts.ts`; that was **not** done,
for a checkable reason — `model.notifications.ts` sits in a graph that
`no-server-infra-in-app-graph.test.ts` forbids from reaching `src/server/db/` and
`src/utils/logging.ts`, both of which the job pulls in. The constants moved to a new module and
the job **re-exports** them, so every existing importer is unchanged. That guard is green.

🔴 **This round touches `remove-old-drafts.ts`, which is live in production.** Verified
independently that the refactor is behaviour-preserving: all **3 SQL statements are identical**
to `origin/main` after whitespace normalisation; the exported symbol set is unchanged
(`ACTIVITY_WINDOW_DAYS`, `REAP_AGE_DAYS`, `removeOldDrafts`, `filterModelsWithRecentActivity`,
`ModelVersionActivityRow`); and both constant **values are still 30**, which matters because
`ACTIVITY_WINDOW_DAYS` is read at runtime by the primary-side fence.

### Regression test for 🔴-1 — the guard this ladder was missing

Matrix **B**, fences reinstated exactly as shipped in `dab517d4fc`:
**3 failed | 10 passed (13)** → **13 passed (13)** at HEAD.
Matrix **A** (original defect, sources from `origin/main`): **5 failed | 6 passed (11)** → **11 passed**.
Whole set: **111 passed (9 files)**.

### Mutation — 28 mutants, 27 killed, **1 survived**, 0 error

**D4 (`OLD_DRAFT_LEAD_DAYS` 7→10) SURVIVED and is not claimed as a kill.** It is an equivalent
mutant: changing the lead is the supported way to change the schedule, everything derives
coherently, and nothing should fail. Pinning it would be a pure change-detector.

**D6/D7/D8 exposed a real gap in this round's own work.** The "derives from the constant"
assertions compared against `OLD_DRAFT_NOTICE_DAYS` (23) and `OLD_DRAFT_LEAD_TEXT` ("1 week") —
so a **hardcoded** `23` or `"1 week"` satisfied every one of them, and three such mutants
survived the whole file. A mechanical control now re-imports the module against a doubled lead
and watches every derived value move; all three die to it. This is the spelled-guard trap, found
inside the fix for the previous spelled-guard trap.

### The 🟡s

- **🟡-3** — the "errors land on the safe side" claim was **false** while the skew existed and is
  now true; the comment says why. The only residual is activity arriving between day 23 and day
  30, which spares a model that was warned.
- **🟡-5** — the retracted `ModelVersion` rationale was still live in
  `backfill-swept-trained-models.ts`; the two sibling files were giving opposite reasons for the
  same omission. Corrected.
- **🟡-4 — checked, and removing the `IGNORED_CURSORS` exemption buys nothing.** In
  `send-notifications.ts`, `await setLastSent()` sits **outside** the `if (!isEmpty(pendingData))`
  branch, so the cursor advances every minute whether or not rows were selected. A predicate
  selecting zero rows forever is indistinguishable from a healthy one; the monitor only sees
  queries that error or time out. Not removed — that would add noise without detection.
  🔴 **So the silent-notification failure mode remains undetectable in production.** That is how
  🔴-1 would have shipped unnoticed, and it is stated here rather than closed.

### Caught by `tsc`, by no test

The new constants module had `error TS2367` — TypeScript narrowed `OLD_DRAFT_LEAD_DAYS` to the
literal `7`, making a `=== 1` comparison provably false. Vitest's esbuild transform does not
typecheck, so all 13 tests were green over it. Fixed with explicit `number` annotations.

### Still true, still unclosed

- **No DB-backed vitest project.** Every SQL guard here is a spelled string comparison.
- **Noted, not fixed:** `Deleted` models are reaped unwarned (notification is `status IN ('Draft')`,
  reaper `('Draft','Deleted')`); the backfill still self-describes as "idempotent" while its
  Known Limits say otherwise; `restoreModelById` carries the same unbumped-clock defect.
- **This PR does not backfill the 1,309 models already carrying a stale clock.**

**Ledger** · round 2 · payload lines changed this round: 52 (of which 15 behaviour-changing SQL,
37 comments) · since round 1: 29.

```audit-claims round=2 audited=dab517d4fc..7cdc9062
1. The notification's fence windows now resolve to the same instant as the reaper's, derived via OLD_DRAFT_NOTICE_DAYS = REAP_AGE_DAYS - OLD_DRAFT_LEAD_DAYS.
2. The parity guard compares resolved windows rather than source text, and simulates the canonical abandoned draft with a fixture-validity assertion.
3. A regression test fails if the fences are spelled as they were in dab517d4fc.
4. Constants moved to src/server/common/draft-reaping.ts; remove-old-drafts.ts re-exports them and its SQL is unchanged.
5. The "safe side" comment is restated to match the fixed behaviour.
6. The retracted ModelVersion rationale is removed from the backfill sibling too.
7. A mechanical control proves the derived values move with the constant, so a hardcoded 23 or "1 week" no longer satisfies the guards.
8. IGNORED_CURSORS was deliberately left alone, with the reason stated.
```

ZacxDev:
## Adversarial audit — round 3 (delta): another 🔴, same class. Fixed by CHANGING APPROACH in `c6b39e34b9`

Round 3 audited `dab517d4fc..7cdc90629b`. It confirmed 6 of 8 claims outright and the
`remove-old-drafts.ts` extraction as genuinely clean — but found the notification **still**
could not warn the population the reaper destroys.

### 🔴 The derivation assumed a one-shot reaper. It is a retrying cron.

`OLD_DRAFT_NOTICE_DAYS = REAP_AGE_DAYS - OLD_DRAFT_LEAD_DAYS` is correct only if the reaper
evaluates once at `U + 30d`. It runs nightly and retries forever, so its real firing instant is
`max(Model."updatedAt", latest version/file activity) + 30d`. The notification evaluates **once**,
in a ~1-minute band at `U + 23d`.

So the fences still excluded every model whose version or file landed *after* the model row was
last written — which `remove-old-drafts.ts:121-129` documents as the norm. Measured across
activity offsets `U+0, +0.5d, +2d, +20d, +40d`: **only `U+0` was warned; the other four were
excluded and reaped later.** `U+0` was precisely the single fixture the round-3 guard pinned,
which is why it passed — **the guard agreed with the bug because both assumed a one-shot reaper.**

### The fix: the activity fences are REMOVED, not re-tuned

Retained: `status IN ('Draft')`, the `BETWEEN` band, `availability != 'Private'`, and the
`downloadCount < 10` `EXISTS`. Removed: both `NOT EXISTS` activity clauses.

All ~1,308 false alarms this PR set out to fix came from `downloadCount >= 10`, so the value
terms alone solve it. The fences were never load-bearing for that, and they were the only terms
whose correctness depended on predicting *when* the reaper fires. Removing them makes the errors
**structurally one-directional**: a model may be warned and later spared, or warned ~9 days ahead
instead of 7, but a model the reaper destroys is never silently deleted.

The value terms are safe for the opposite reason, now recorded in the comment: `downloadCount` is
monotonically non-decreasing, so a day-23 reading of `< 10` still holds at day 30; and an
`availability` change writes the Model row, bumping `updatedAt`, which re-arms the band *and*
resets the reaper's clock — the two move together.

Verified independently on the merged SQL: `NOT EXISTS` 0, `ModelVersion` 0, `ModelFile` 0;
both remaining `INTERVAL`s are the band.

### The guard changed SHAPE — fence-parity was never the right invariant

The property is now **"no term may exclude a model the reaper will destroy."** Three structural
forms (no `now()`-relative comparison; no `NOT EXISTS`; no reference to `"ModelVersion"`/
`"ModelFile"` at all — the widest, which a reworded fence cannot walk past, with a `"ModelMetric"`
positive control), plus a property test parameterised over activity offsets, modelling the reap
instant as `max(U, activity) + REAP_AGE_DAYS` rather than hardcoding the one-shot reaper.

Red proof it catches **both** prior defects, measured on the 17-test file:

| shape reinstated | result |
|---|---|
| round-2 fences (`'30 days'`) | **8 failed \| 9 passed** |
| round-3 fences (derived) | **7 failed \| 10 passed** |
| HEAD | **17 passed** |

### Reporting corrections

**🟢-3 — Matrix A.** Both measurements were right and neither named its file set: `11` was
`reset-to-draft` (7) + `no-unbumped` (4); `13` was the parity file alone. **Not naming the file
set was the actual defect** — every matrix now names its files and per-file counts.

**🟢-5 — "111 passed (9 files)"** is replaced by an itemised 115 summing file-by-file, with an
explicit note that **no test exercises `backfill-swept-trained-models`, `send-notifications` or
`notification-cursor-monitor` behaviourally** — the earlier aggregate should not have been
readable as implying otherwise.

**🟢-4** — the `NOTICE`/`ACTIVITY_WINDOW` conflation site is gone with the fences.

### Mutation — 27 mutants, 26 killed, **1 survived**

New: **F1/F2/F3** re-add a fence in the round-2 shape, the round-3 shape, and a *single* fence at
`'1 days'` — a shape neither prior guard would have caught. **V4/V5** are literals equal to the
constants' current values, killed only by the derivation control.

**V6 (`OLD_DRAFT_LEAD_DAYS` 7→10) SURVIVED** — equivalent mutant, unchanged from last round. Not
claimed as a kill.

### 🔴 The standing risk, stated rather than closed

There is still **no DB-backed vitest project**. The offset simulation is arithmetic over the
intervals the queries carry **plus a model of the cron's schedule** — and if that model is wrong
again, these guards will agree with it. That is precisely how rounds 2 and 3 both passed. The
structural guards (no `now()`-relative clause, no version/file reference) are the part that does
*not* depend on that model, which is why they carry the weight.

Also unchanged: `Deleted` models are reaped unwarned (notification is `status IN ('Draft')`);
`restoreModelById` carries the same unbumped-clock defect; and this PR does not backfill the
1,309 models already holding a stale clock.

**Ledger** · round 3 · payload lines changed this round: 175 (23 non-comment) · scaffolding 328.

```audit-claims round=3 audited=7cdc90629b..c6b39e34
1. Both NOT EXISTS activity fences are removed from the old-draft query; only status, band, availability and downloadCount remain.
2. The guard asserts a one-directional property — no term may exclude a model the reaper will destroy — via three structural forms plus a parameterised offset test.
3. The property test models the reap instant as max(U, activity) + REAP_AGE_DAYS rather than U + REAP_AGE_DAYS.
4. Reinstating either the round-2 or the round-3 fence shape turns the guard red, measured.
5. The false "the interval derivation prevents reaped-without-warning" comment is replaced with the two real residual directions.
6. Every reported matrix now names its file set and per-file counts.
7. remove-old-drafts.ts is untouched this round.
```

PR #4598, opened 2026-09-03
feat(observability): dark probe for whether caching moderation verdicts would pay

ZacxDev:
## What

A dark measurement probe that answers one question — **if we cached external moderation verdicts, how often would we skip the classifier?** — without caching anything. Off by default.

## Why this and not something cheaper

`extModeration.moderatePrompt` is an outbound HTTPS call that runs inline and serially on the generation submission path. Measured in production it is **~200 ms per call, p50 ~188 ms**, and it accounts for roughly **78–88 % of the app-side (non-submit) time of `orchestrator.generateFromGraph`**.

Three cheaper ways to attack that were tested against production measurements first, and **all three are dead**:

| lever | verdict | worth |
|---|---|---|
| Tune the tail / abort deadline | it's a **floor, not a tail** — 0 calls under 50 ms, 0.46 % under 100 ms | 6–24 ms |
| Reuse the connection | calls *do* pay a fresh TCP+TLS handshake (spacing per process far exceeds undici's 4 s default keep-alive, and nothing configures a dispatcher) — but the classifier answers from a nearby edge, so the whole handshake is ~7–24 ms | ~10 ms |
| Overlap it with the submit | impossible by construction — the audit is a **fail-closed gate**, so a flagged prompt must never reach `submitWorkflow` | ~10 ms |

What's left is the classifier's own inference time, and the only way to avoid paying it is to **not make the call**. That's worth building only if prompts actually repeat — and nobody knows whether they do. This measures exactly that, and nothing else.

## 🔴 It is a probe, not a cache

No verdict is ever read back. The classifier is called every single time, whatever the probe reports. Turning this into a real cache is a **trust-and-safety decision about verdict staleness**, not a performance one, and belongs in its own change with its own review. There's a test that fails if a hit ever skips the call.

## Design notes

- **Off by default, and armed with a NAMESPACE rather than a boolean** — `EXTERNAL_MODERATION_CACHE_PROBE=prod`, and `prod` is the **only** accepted value. The set is a closed allowlist rather than a charset plus a denylist (a denylist defends a class by listing members; `y`, `n`, `none`, `null`, `nil`, `disable`, `2` all walked through the one that shipped for a commit), and the three values that were once members are asserted **absent**: `preview` and `next` are *classes* rather than deployments, and `next-stage` isn't scraped. See the rollout note. It ships inert so the **arming instant is visible** as the moment the series appear — that's the only thing that distinguishes "no repeats" from "the probe never ran". The value becomes a key segment, so the probe cannot be armed without naming a keyspace: several civitai-web deployments share one sysRedis and sys keys carry no environment prefix, so two armed deployments would otherwise score hits on each other's prompts.
- **Two windows (5m, 1h), not one.** A single hit rate has no shape: it can't separate "users re-roll the same prompt within minutes" (a tiny TTL captures the value) from "the same prompts recur all day" (the cache has to be big). Those imply different builds.
- **Read the hit rate as `hit/(hit+miss)`, always `by (window)`.** Dividing over the *total* folds in `error`, so a Redis outage reads as "prompts stopped repeating" — the reassuring direction, and therefore the dangerous one. Omitting `by (window)` averages the 5m and 1h series into one number and collapses the very shape the two windows exist to reveal.
- **Ignore the first full window after arming.** The keyspace starts empty, so everything is necessarily a miss until the probe has run longer than the window it measures.
- **Fire-and-forget**, so it can't add latency to the very request whose latency is under investigation, and can't fail a generation. The Redis client is imported lazily to keep `moderation.ts`'s import graph light for its cron caller.
- **Hashes the *prepared* prompt** (post `removeFalsePositiveTriggers`) — the exact string the classifier receives, so the digest matches what a real cache would key on. Only the digest reaches Redis; **no prompt text is stored, logged, or put on a label.**
- **`SET NX EX`** in one atomic round trip, which keeps both windows cheap and the observation self-consistent — but see the bound below: that atomicity is not the same thing as fidelity.
- 🔴 **The bound is SCOPED, not absolute.** The probe differs from a real cache on **two** independent axes — it coalesces concurrent duplicates (`SET NX` claims the slot at request *start*, ~200 ms before a read-through cache could store an answer), and its TTL never extends on a hit. So: vs a **fixed-TTL non-coalescing** cache it is an **upper** bound; vs a **sliding-TTL coalescing** cache a **lower** one; vs a **sliding-TTL non-coalescing** cache — the ordinary design, since singleflight is a feature you must build — the two effects oppose and **no bound holds**. Name both axes before quoting a direction. *(Two corrections deep: round 1 said "floor", round 2 said "upper bound" unqualified, round 3 caught the sliding-TTL half pinning only one axis.)*

## Testing

22 cases, plus a **mutation battery that was run rather than trusted** — 15 isolated mutants plus targeted membership ones. Every defect the battery or an audit round caught is listed below; **count the bullets rather than trusting a total**, because a total kept beside the thing it counts has now gone stale three rounds running:

- 🔴 **The flag-off case was vacuous** (round 1, self-caught). Deleting the flag guard left the suite green, because a 20 ms sleep is shorter than the probe's first lazy `import()` — it proved the probe is *slow*, not *off*. Restructured as a differential.
- 🔴 **The probe's own `source` clamp was untested** (audit round 1). The existing case drove `moderatePrompt`, which clamps first and passes the clamped value, so the probe's clamp never saw a bad string — a mutant deleting it survived a green 14-case suite. Now exercised by a case that calls the export directly.
- 🔴 **Every boolean-shaped "off" token ARMED the probe** (audit round 2). Once the flag became a namespace, the charset accepted `false`/`off`/`no`/`0`/`disabled` — so the likeliest spelling of "disable this" armed it under a namespace called `false`, and two deployments "disabled" that way would share `…:false:…`.
- 🔴 **And the denylist that fixed it was itself walkable** (audit round 3). `y`, `n`, `none`, `null`, `undefined`, `nil`, `disable`, `enable`, `2` all passed both the charset and the ten-word denylist; zod's own `stringbool` treats `y`/`n` as boolean spellings. Replaced with a closed allowlist — a guard that is structural rather than spelled.
- 🔴 **Shipping inert did not mean shipping silent** (own battery, found by *isolating* the mutants properly). Removing the empty-string early return leaves the probe off — so all 20 cases stayed green — while making an unarmed deployment log a misconfiguration error it has done nothing to deserve. Nothing asserted the silence. Now pinned. *(Measured cost of that mutant: one line per process per distinct value — the dedupe caps it. An earlier revision of this body said "on every generation", overstating it by orders of magnitude.)*

- 🔴 **`preview` was in the allowlist and defeated the allowlist's own invariant** (audit round 4). `preview` is not a deployment, it is a *class*: ~10 concurrent `civitai-pr-*` namespaces run this code, take config from one shared template, and share one sysRedis — so arming that word arms all of them into one keyspace and they score mutual hits across unrelated PRs. Dropped, and asserted absent.
- 🔴 **The allowlist was the whole arming contract with half of it unexercised** (audit round 4). Deleting `next-stage` left all 20 cases green. Now an asserted ledger that fails whether the set **grows or shrinks** — verified in both directions.

- 🔴 **Dropping `preview` moved the class hazard instead of closing it** (audit round 5). The PR-preview Tekton task copies `civitai-next`'s `civitai-cfg` **wholesale** into every `civitai-pr-<N>` namespace and repoints them at one shared sysRedis — so arming `next` would silently arm every open PR's preview into a single keyspace. Verified against the GitOps repo.
- 🔴 **Two of the three members were unreadable even when armed** (audit round 5). `civitai-next` and `civitai-next-stage` have **zero** ServiceMonitors/PodMonitors between them, and the probe's population twin exists in exactly one namespace. Arming them shows no series — which the help text defines as "not armed". Allowlist reduced to `prod`.
- 🔴 **The membership ledger was false in the growth direction** (audit round 5). Adding `'staging'` survived 21/21, because a loop over candidates can only see growth by a member already in its own list. The set is now exported and asserted **whole** against a literal.
- 🔴 **The battery caught me deleting two tests.** Splicing a rewrite on text anchors silently removed everything between them; the suite stayed green at 19/19 and it surfaced only as a mutant surviving. Both restored. A green suite cannot tell you a test is missing.

- 🔴 **Narrowing to one allowlist member silently took the namespace-segment coverage to ZERO** (audit round 6). A mutant hardcoding the namespace was killed by two cases one commit earlier and **survived a green 21-case suite** after the narrowing — with one member, the shape test cannot tell a literal from the real derivation. Restored with an exported `buildProbeKey` seam a test can drive with two namespaces. The comment claiming the replacement was "stronger, not weaker" was false and is corrected.
- 🔴 **A fix can be un-fixed by a restore.** The round-5 relative-counter fix landed, then I restored two deleted tests *from `HEAD`* — which predated it — silently reverting it, while the commit message asserted it had landed. Re-applied and verified by grepping after, not by trusting the edit.

Every mutant below is **isolated** to the narrowest expression that can be wrong. That matters: two earlier rows were not, and each scored a kill for the wrong reason. Counts are measured, not estimated; the unmutated control runs green before and after.

| mutant |
|---|
| hash the raw prompt instead of the prepared one |
| the probe's own clamp removed |
| allowlist guard removed |
| allowlist widened back to the previous charset guard |
| allowlist grown by a member |
| allowlist not frozen |
| empty string no longer means off |
| namespace **hardcoded** in the key |
| namespace dropped from the key |
| `error` folded into `miss` |
| hit/miss inverted |
| `NX` dropped |
| diagnostic log removed |
| diagnostic logs on every call |
| probe moved above the not-configured skip check |

⚠️ **Per-mutant kill counts are deliberately not listed.** They were, and they drifted for three consecutive rounds — a count kept beside the thing it counts goes stale every time the suite changes, and a stale count in a table headed "measured, not estimated" is worse than no table. The claim that matters is the one the reader can re-derive: **every mutant above is killed, and the unmutated control is green before and after.**

`pnpm typecheck` clean (0 errors), eslint + prettier clean, and **732 passed / 7 skipped** over a named population: `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

⚠️ **Corrections to earlier revisions of this body**, kept rather than silently edited away: it described the result as a **floor** (wrong, optimistically), then as an unqualified **upper bound** (pinning only one of two axes); it reported the skip-check mutant as killed by 2 cases (**1**, once isolated); it said a mutant cost a log line **per generation** (one per process per distinct value); and it stated a mutant count and kill table that had gone stale. The source comments carry the same retractions, including one where I asserted the previous env spelling "failed boot loudly on garbage" — it never failed on anything, which I confirmed by executing it.

## Rollout

Merging changes nothing — the flag defaults to empty, which is OFF. Arming it means setting `EXTERNAL_MODERATION_CACHE_PROBE=prod` on `civitai-dp-prod`, which is the **only** deployment the allowlist accepts and the only one that is scraped. Give it more than an hour before reading the 1h window.

🔴 **Before ever adding a member, check both**: that the value names ONE running population rather than a config template other namespaces inherit, and that something actually scrapes it. Both of those failed silently on values that looked obviously fine.

https://claude.ai/code/session_011rtwdS8j8PA3TWMZemcY27

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1815,7 +1815,8 @@ export const REDIS_SYS_KEYS = {
      * exact string sent to the external prompt classifier, so we can count how often that string
      * REPEATS inside a window. Nothing reads the value back to skip work; the classifier is called
      * every time either way. Written only while EXTERNAL_MODERATION_CACHE_PROBE names a namespace
-     * (it is a deployment LABEL, not an on/off flag — empty, or any on/off word, means disabled),
+     * (it is a deployment LABEL, not an on/off flag, and the accepted labels are a CLOSED
+     * ALLOWLIST — empty, or anything outside it, means disabled),
      * and the value becomes the segment after this prefix so two armed deployments sharing one
      * sysRedis cannot collide. Every key carries an EX, so the keyspace evaporates on its own once
      * the probe is disarmed.
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1810,6 +1810,15 @@ export const REDIS_SYS_KEYS = {
     BLOCKED_PROMPTS: 'generation:blocked-prompts',
     REMIX_AUDIT_CHECKED: 'generation:remix-audit-checked',
     CLIENT: 'generation:client',
+    /**
+     * MEASUREMENT ONLY — a dark probe, not a cache. Holds `1` against a truncated SHA-256 of the
+     * exact string sent to the external prompt classifier, so we can count how often that string
+     * REPEATS inside a window. Nothing reads the value back to skip work; the classifier is called
+     * every time either way. Written only while EXTERNAL_MODERATION_CACHE_PROBE is on, and every
+     * key carries an EX, so the whole keyspace evaporates on its own when the flag goes off.
+     * See src/server/integrations/moderation-cache-probe.ts.
+     */
+    MODERATION_CACHE_PROBE: 'generation:moderation-cache-probe',
   },
   TRAINING: {
     STATUS: 'training:status',
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1814,8 +1814,11 @@ export const REDIS_SYS_KEYS = {
      * MEASUREMENT ONLY — a dark probe, not a cache. Holds `1` against a truncated SHA-256 of the
      * exact string sent to the external prompt classifier, so we can count how often that string
      * REPEATS inside a window. Nothing reads the value back to skip work; the classifier is called
-     * every time either way. Written only while EXTERNAL_MODERATION_CACHE_PROBE is on, and every
-     * key carries an EX, so the whole keyspace evaporates on its own when the flag goes off.
+     * every time either way. Written only while EXTERNAL_MODERATION_CACHE_PROBE names a namespace
+     * (it is a deployment LABEL, not an on/off flag — empty, or any on/off word, means disabled),
+     * and the value becomes the segment after this prefix so two armed deployments sharing one
+     * sysRedis cannot collide. Every key carries an EX, so the keyspace evaporates on its own once
+     * the probe is disarmed.
      * See src/server/integrations/moderation-cache-probe.ts.
      */
     MODERATION_CACHE_PROBE: 'generation:moderation-cache-probe',
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -614,8 +614,9 @@ export const serverSchema = z
     // and never adds latency to the request (the Redis round trip is fire-and-forget).
     //
     // 🔴 IT IS A NAMESPACE, NOT A BOOLEAN, AND THAT IS THE WHOLE POINT. Set it to a short label for
-    // the deployment being measured — the accepted set is the allowlist named below, currently
-    // `prod`, `next`, `next-stage`; the value becomes a segment of the
+    // the deployment being measured. The accepted set is the allowlist in
+    // moderation-cache-probe.ts and is deliberately NOT restated here — restating it is exactly how
+    // this sentence went stale once already. The value becomes a segment of the
     // probe's Redis key. Several civitai-web deployments SHARE ONE sysRedis, and unlike cache keys
     // — which get an environment prefix via CACHE_KEY_NAMESPACE — sys keys carry no environment
     // segment at all (see cache-key-prefix.ts: "This is CACHE-ONLY"). So two armed deployments
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -607,6 +607,13 @@ export const serverSchema = z
     // ~116-day timeout) re-introduces the unbounded-park failure the deadline exists
     // to prevent. Any out-of-range value falls back to 5000.
     EXTERNAL_MODERATION_TIMEOUT_MS: z.coerce.number().int().min(100).max(60000).catch(5000),
+    // Dark measurement probe for "would a moderation-result cache pay?". OFF by default and
+    // deliberately so: it ships inert, gets flipped on in config, and the metric's ARMING DATE is
+    // then visible as the instant its series appear — which is the only thing that distinguishes
+    // "no repeats" from "probe never ran". It never changes the verdict, never skips the
+    // classifier, and never adds latency to the request (the Redis round trip is fire-and-forget).
+    // See src/server/integrations/moderation-cache-probe.ts.
+    EXTERNAL_MODERATION_CACHE_PROBE: zc.booleanString.optional().default(false),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
 
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -636,12 +636,14 @@ export const serverSchema = z
     // `''` — routing the probe through it would put production and stage in one shared probe
     // keyspace, which is precisely the collision this namespace exists to prevent.
     //
-    // Values are restricted to /^[a-z0-9][a-z0-9-]{0,31}$/ AND rejected if they are an on/off word
-    // (`true`/`false`/`on`/`off`/`yes`/`no`/`0`/`1`/`enabled`/`disabled`) — the charset alone would
-    // happily accept `false` as a namespace, so the most likely way to spell "turn this off" would
-    // ARM the probe. Anything rejected is treated as OFF and logged once, rather than sanitised, so
-    // a typo produces NO SERIES (already documented as "not armed") plus a line saying why, instead
-    // of a second silent keyspace. See src/server/integrations/moderation-cache-probe.ts.
+    // The accepted values are a CLOSED ALLOWLIST of deployment labels, defined in
+    // moderation-cache-probe.ts and deliberately not restated here (one rule, one place). Anything
+    // else — including every on/off spelling — is treated as OFF and logged once, so a typo yields
+    // NO SERIES (already documented as "not armed") plus a line saying why, rather than a second
+    // silent keyspace. An allowlist rather than a charset plus a denylist because a denylist is a
+    // guard SPELLED rather than STRUCTURAL: the charset alone accepts `false`, `n` and `disabled`
+    // as perfectly good namespaces, so the likeliest spelling of "turn this off" would ARM the
+    // probe. See src/server/integrations/moderation-cache-probe.ts.
     EXTERNAL_MODERATION_CACHE_PROBE: z.string().trim().optional().default(''),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -614,7 +614,8 @@ export const serverSchema = z
     // and never adds latency to the request (the Redis round trip is fire-and-forget).
     //
     // 🔴 IT IS A NAMESPACE, NOT A BOOLEAN, AND THAT IS THE WHOLE POINT. Set it to a short label for
-    // the deployment being measured (`prod`, `next`, `stage`); the value becomes a segment of the
+    // the deployment being measured — the accepted set is the allowlist named below, currently
+    // `prod`, `next`, `next-stage`; the value becomes a segment of the
     // probe's Redis key. Several civitai-web deployments SHARE ONE sysRedis, and unlike cache keys
     // — which get an environment prefix via CACHE_KEY_NAMESPACE — sys keys carry no environment
     // segment at all (see cache-key-prefix.ts: "This is CACHE-ONLY"). So two armed deployments
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -622,15 +622,26 @@ export const serverSchema = z
     // the result toward "caching pays" — the direction that gets a cache built that does not pay.
     //
     // Making the namespace the ARMING SWITCH is what stops that being a thing to remember: there is
-    // no way to turn the probe on without naming a keyspace for it. A separate namespace variable
-    // could be left unset while the probe ran, which is exactly how CACHE_KEY_NAMESPACE behaves in
-    // practice — measured 2026-09-03, it is ABSENT on all of civitai-dp-prod, civitai-next and
-    // civitai-next-stage, including the deployment its own doc says should set it. A fix routed
-    // through it would have been inert while looking like a fix.
+    // no way to turn the probe on without naming a keyspace for it.
     //
-    // Values are restricted to /^[a-z0-9][a-z0-9-]{0,31}$/; anything else is treated as OFF rather
-    // than sanitised, so a typo produces NO SERIES (loud, and already documented as "not armed")
-    // instead of a second silent keyspace. See src/server/integrations/moderation-cache-probe.ts.
+    // ⚠️ WHY NOT REUSE CACHE_KEY_NAMESPACE — corrected 2026-09-03, because the first version of this
+    // comment asserted a measurement that was FALSE. It claimed that variable is "ABSENT on all of
+    // civitai-dp-prod, civitai-next and civitai-next-stage". It is not: `CACHE_KEY_NAMESPACE=next`
+    // is set on civitai-next (in the deployment env, which the original check never read — it
+    // looked only at ConfigMaps and generalised one source into a claim about all of them).
+    //
+    // The real reason it cannot serve here is the opposite of "nobody sets it": it is set EXACTLY
+    // as designed, and its design is wrong for this purpose. cache-key-prefix.ts requires
+    // production to be the EMPTY prefix, so civitai-dp-prod and civitai-next-stage BOTH resolve to
+    // `''` — routing the probe through it would put production and stage in one shared probe
+    // keyspace, which is precisely the collision this namespace exists to prevent.
+    //
+    // Values are restricted to /^[a-z0-9][a-z0-9-]{0,31}$/ AND rejected if they are an on/off word
+    // (`true`/`false`/`on`/`off`/`yes`/`no`/`0`/`1`/`enabled`/`disabled`) — the charset alone would
+    // happily accept `false` as a namespace, so the most likely way to spell "turn this off" would
+    // ARM the probe. Anything rejected is treated as OFF and logged once, rather than sanitised, so
+    // a typo produces NO SERIES (already documented as "not armed") plus a line saying why, instead
+    // of a second silent keyspace. See src/server/integrations/moderation-cache-probe.ts.
     EXTERNAL_MODERATION_CACHE_PROBE: z.string().trim().optional().default(''),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -607,13 +607,31 @@ export const serverSchema = z
     // ~116-day timeout) re-introduces the unbounded-park failure the deadline exists
     // to prevent. Any out-of-range value falls back to 5000.
     EXTERNAL_MODERATION_TIMEOUT_MS: z.coerce.number().int().min(100).max(60000).catch(5000),
-    // Dark measurement probe for "would a moderation-result cache pay?". OFF by default and
-    // deliberately so: it ships inert, gets flipped on in config, and the metric's ARMING DATE is
-    // then visible as the instant its series appear — which is the only thing that distinguishes
-    // "no repeats" from "probe never ran". It never changes the verdict, never skips the
-    // classifier, and never adds latency to the request (the Redis round trip is fire-and-forget).
-    // See src/server/integrations/moderation-cache-probe.ts.
-    EXTERNAL_MODERATION_CACHE_PROBE: zc.booleanString.optional().default(false),
+    // Dark measurement probe for "would a moderation-result cache pay?". Empty/unset = OFF, and
+    // deliberately so: it ships inert, gets armed in config, and the metric's ARMING DATE is then
+    // visible as the instant its series appear — which is the only thing that distinguishes "no
+    // repeats" from "probe never ran". It never changes the verdict, never skips the classifier,
+    // and never adds latency to the request (the Redis round trip is fire-and-forget).
+    //
+    // 🔴 IT IS A NAMESPACE, NOT A BOOLEAN, AND THAT IS THE WHOLE POINT. Set it to a short label for
+    // the deployment being measured (`prod`, `next`, `stage`); the value becomes a segment of the
+    // probe's Redis key. Several civitai-web deployments SHARE ONE sysRedis, and unlike cache keys
+    // — which get an environment prefix via CACHE_KEY_NAMESPACE — sys keys carry no environment
+    // segment at all (see cache-key-prefix.ts: "This is CACHE-ONLY"). So two armed deployments
+    // would write the same probe keyspace and each would score HITS on the other's prompts, biasing
+    // the result toward "caching pays" — the direction that gets a cache built that does not pay.
+    //
+    // Making the namespace the ARMING SWITCH is what stops that being a thing to remember: there is
+    // no way to turn the probe on without naming a keyspace for it. A separate namespace variable
+    // could be left unset while the probe ran, which is exactly how CACHE_KEY_NAMESPACE behaves in
+    // practice — measured 2026-09-03, it is ABSENT on all of civitai-dp-prod, civitai-next and
+    // civitai-next-stage, including the deployment its own doc says should set it. A fix routed
+    // through it would have been inert while looking like a fix.
+    //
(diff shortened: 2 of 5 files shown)


ZacxDev:
**Audit round 1 — 3 🟡, 1 🟢, no 🔴. All four addressed in `a8e99077`.**

```audit-claims round=1 audited=411db897af..a8e99077
1. F3 — the probe's own `clampExternalModerationSource` call was untested (the audit's mutant deleting it SURVIVED a green 14-case suite); claimed fixed by a case that drives the exported `probeModerationCacheRepeat` directly rather than through `moderatePrompt`.
2. F2 — the probe Redis keyspace had no per-deployment segment while several civitai-web deployments share one sysRedis; claimed fixed by making `EXTERNAL_MODERATION_CACHE_PROBE` a namespace string rather than a boolean, with the value written into the key and out-of-charset values treated as OFF.
3. F1 — the "floor on the achievable hit rate" claim was wrong and contradicted the call-site comment in the same PR; claimed fixed by restating the result as an UPPER bound in the metric help text, the module source and the PR body, on the grounds that `SET NX` simulates a coalescing cache.
4. F4 — the prescribed hit-rate query omitted `by (window)`; claimed fixed in the metric help text and the PR body.
5. The round-1 mutation battery's skip-check mutant was NOT isolated (it moved the call and changed its argument); claimed re-measured isolated, dying to exactly 1 case, and the PR body's table corrected from 2 to 1.
```

Also corrected publicly in the PR body rather than silently: it previously reported the skip-check mutant as killed by 2 cases, and described the result as a floor.

Battery re-run at 8 mutants (the original 6 minus the unisolated skip-check, plus the isolated one and three new: the probe's own clamp, an invalid namespace arming, the namespace dropped from the key). All 8 killed, unmutated control green. 18 cases (was 14). `pnpm typecheck` 0 errors; eslint + prettier clean; 728 passed / 7 skipped across `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 2 — 3 🟡, 3 🟢, no 🔴. Two of the 🟡s were introduced by round 1's own fix. All addressed in `0e2468d0`.**

```audit-claims round=2 audited=a8e9907719..0e2468d08a
1. 🟡1 — the namespace charset accepted on/off words, so `EXTERNAL_MODERATION_CACHE_PROBE=false` ARMED the probe under a namespace called `false` (and two deployments "disabled" that way would share one keyspace); claimed fixed by an explicit on/off denylist checked alongside the charset pattern, plus a case asserting none of ten such words arms it.
2. 🟡2 — the shipped comment asserted a FALSE measurement, that CACHE_KEY_NAMESPACE is absent on all three deployments; claimed fixed by re-deriving it live (it IS set to `next` on civitai-next, in the deployment env, which the original ConfigMap-only check never read) and replacing the rationale with the correct one: production must be the empty prefix, so dp-prod and next-stage both resolve to '' and would share a probe keyspace.
3. 🟡3 — "UPPER BOUND" was concluded from a premise the same paragraph withdrew; claimed fixed by scoping rather than netting the two biases — overstates vs a fixed-TTL non-coalescing cache, understates vs a sliding-TTL one — in both the source comment and the metric help text.
4. 🟢5 — a rejected namespace was silently swallowed; claimed fixed by a once-per-distinct-value `console.error`, on the same reasoning cache-key-prefix.ts gives for logging rather than throwing.
5. 🟢6 — the Redis key's docstring still described an on/off flag; claimed fixed to describe a namespace.
6. 🟢4 — the mutation table's counts were stale and one row's mutant was not isolated; claimed fixed by re-running all 12 mutants isolated and publishing measured counts, and by naming the suite population behind the "730 passed / 7 skipped" figure so it is pinnable.
```

**A note on 🟡2, because it is the one that matters most:** I re-verified it rather than accepting it, and the auditor was right. My original check queried ConfigMaps only and generalised that one source into a claim about all configuration — sampling written down as enumeration, and it shipped as an assertion in source. `CACHE_KEY_NAMESPACE=next` is set on `civitai-next` via the deployment env.

**Round 3's own battery then found a fourth defect, by isolating mutants properly:** removing the empty-string early return leaves the probe *off* — so all 20 cases stayed green — while making every unarmed deployment log an error on every generation. Shipping inert has to mean shipping silent, and nothing asserted that. Now pinned; the mutant dies to exactly one case.

12 mutants, all killed, unmutated control green before and after. 20 cases (was 18). `pnpm typecheck` 0 errors; eslint + prettier clean; 730 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 3 — 2 🟡, 3 🟢, no 🔴. Both 🟡s were wrong SENTENCES I shipped, not wrong code. All addressed in `f7407347`.**

```audit-claims round=3 audited=0e2468d08a..f740734740
1. 🟡F1 — the source comment asserted the previous `zc.booleanString` spelling "failed boot loudly on garbage", which is false because its preprocess maps every input to a boolean; claimed fixed by executing it against the installed zod (`'off'`, `'garbage'`, `'$$$'`, `''` all parse to `false`, nothing throws) and replacing the rationale with the log's actual merit — telling "armed, no repeats" apart from "my value was rejected".
2. 🟡F2 — "vs a SLIDING-TTL cache it is a LOWER bound" pinned only one of the two axes on which the probe differs from a real cache; claimed fixed by stating three explicit comparisons in BOTH the metric help text and the source, with the ordinary sliding-TTL non-coalescing design named as INDETERMINATE rather than bounded.
3. 🟢F4 — the on/off denylist was a guard spelled rather than structural, and `y`/`n`/`none`/`null`/`undefined`/`nil`/`disable`/`enable`/`2` walked through it; claimed fixed by replacing charset+denylist with a closed allowlist of deployment labels, and by extending the test's word list to include every enumerated survivor.
4. 🟢F3 — the claim that the empty-string mutant makes "every unarmed deployment log an error on every generation" overstated it by orders of magnitude; claimed fixed by correcting it to one line per process per distinct value, in the test comment and the PR body.
5. 🟢F5 — the PR body still carried the unqualified "UPPER BOUND" line the shipped source had withdrawn; claimed fixed so the merge record no longer contradicts the code.
```

**On 🟡F1 — I executed it rather than accepting it, and the auditor was right.** `zc.booleanString` is `z.preprocess((v) => v === true || v === 'true', z.boolean())`. The preprocess maps *every* input to a boolean, so the inner schema cannot fail. Nothing ever threw on garbage. That makes two rounds running where the defect was a factual assertion in shipped source that I had not actually verified — the same class as round 2's `CACHE_KEY_NAMESPACE` claim.

**On 🟢F4 — the replacement is demonstrably stronger, not merely different.** The battery now includes `allowlist-widened-to-charset`, which reverts the guard to exactly what shipped last round; it **dies** to the off-ish-spelling case. That is the difference between "we changed the guard" and "the new guard catches something the old one did not".

12 isolated mutants, all killed, unmutated control green before and after. 20 cases. `pnpm typecheck` 0 errors; eslint + prettier clean; 730 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 4 — 4 🟡, 3 🟢, no 🔴. All addressed in `58602693`.**

The auditor's diagnosis is the useful part and worth stating separately from the findings: **the failure pattern shifted.** Rounds 2 and 3 were *"a claim I never checked"*. Three of round 4's findings are *"a claim that was true until I changed the thing it describes"* — the class the round-ladder rules call a claim staled by the PR's own earlier commit, which no delta round is positioned to see. So this round was a **mechanical sweep** of every site naming the guard or the bound, not another round of reading.

```audit-claims round=4 audited=f740734740..58602693c4
1. 🟡4 — `preview` was an allowlist member and defeats the invariant the allowlist enforces (~10 concurrent civitai-pr-* namespaces share one config template and one sysRedis, so arming that word arms all of them into a single keyspace); claimed fixed by removing it and ASSERTING its absence, not merely omitting it.
2. 🟢6 — the allowlist was the entire arming contract while only 2 of its 4 members were exercised, and deleting `next-stage` left all 20 cases green; claimed fixed by an asserted membership ledger that fails whether the set grows or shrinks, verified in both directions (drop-next-stage and re-add-preview are each killed by that case alone).
3. 🟡1 — `server-schema.ts` still offered `stage` as an example value, which the previous round's narrowing had made invalid; claimed fixed by pointing at the allowlist rather than restating members.
4. 🟡2 — `moderation.ts` still said "the help text calls the result an upper bound and not a floor", which the help text stopped saying last round; claimed fixed by removing the direction from that site and pointing at the three enumerated comparisons.
5. 🟡3 — the test file still carried the "previous boolean spelling failed boot loudly on garbage" sentence verbatim after it was retracted in the source; claimed fixed so the retraction is complete rather than half-applied.
6. 🟢5 — the Redis key docstring said "any on/off word means disabled", true under the denylist and wrong in the wide direction under an allowlist; claimed fixed.
7. 🟢7 — the PR body's own defect counts had gone stale again (four/two against five bullets/three audit-annotated); claimed fixed.
```

**On 🟡4, which is the one I would not have merged:** the allowlist's own doc comment says two armed deployments sharing a keyspace would "score HITS on the other's prompts, biasing the result toward 'caching pays'". For `preview` the allowlist admitted exactly that. A member that defeats its guard's invariant is worse than no guard, because the guard is what stops anyone looking.

12 isolated mutants + 4 targeted membership mutants, all killed, unmutated control green. 21 cases. `pnpm typecheck` 0 errors; eslint + prettier clean; 731 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 5 — 2 🟡, 4 🟢, no 🔴. Both 🟡s were NON-prose, and both are closed by one change. Addressed in `8526826f`.**

```audit-claims round=5 audited=58602693c4..8526826f56
1. 🟡F1 — dropping `preview` moved the class hazard rather than closing it, because the PR-preview Tekton task copies civitai-next's civitai-cfg WHOLESALE into every civitai-pr-<N> namespace on one shared sysRedis; claimed fixed by removing `next` from the allowlist, verified against the GitOps repo rather than accepted.
2. 🟡F2 — `next` and `next-stage` produce no series even when armed, because neither namespace has any ServiceMonitor or PodMonitor; claimed fixed by removing both, verified by counting monitors in all three namespaces and by the probe's population twin existing in exactly one namespace. The allowlist is now `prod` alone.
3. 🟢F3 — the membership ledger's "fails whether the set GROWS or SHRINKS" was false in the growth direction (adding 'staging' survived 21/21); claimed fixed by exporting PROBE_NAMESPACES and asserting the whole set against a literal, with three growth mutants verified killed by that case.
4. 🟢F5 — two self-staling relative commit counters ("two commits ago"); claimed fixed in FORM, by naming the audit round instead of a relative count, so they cannot stale again.
5. 🟢F6 — server-schema.ts restated the allowlist 22 lines above the sentence saying it is "deliberately not restated here"; claimed fixed by not restating it.
6. 🟢F4 — the PR body's defect counters had staled for a third consecutive round; claimed fixed in FORM, by deleting the running total and telling the reader to count the bullets.
```

**Both 🟡s were verified independently, not accepted.** `pr-deploy-task.yaml` really does `kubectl get configmap civitai-cfg -n civitai-next -o json | jq …` and clone it wholesale, overriding only an explicit key list this variable is not on. And `civitai-next` / `civitai-next-stage` really do have **zero** monitors between them, with `civitai_app_external_moderation_duration_seconds_count` present in exactly one namespace.

The two findings collapse into one fix: **the allowlist is `prod` alone.** `next` is a class, `next-stage` is unread, `preview` is both. The source now records each removal with its measured reason and states the two tests a future member must pass — it must name ONE running population, and something must actually scrape it — because a member failing either makes the guard assert an invariant it does not enforce.

🔴 **And the battery caught me deleting two tests.** Splicing the ledger rewrite on text anchors silently removed everything between the endpoints; the off-ish-spelling case and the log-dedupe case both vanished and the suite stayed **green at 19/19**, because nothing else covered them. It surfaced only as `diagnostic-logs-every-call` SURVIVING. Both restored, 21 cases again. A green suite cannot tell you a test is missing.

12 isolated mutants + 6 targeted membership mutants, all killed, unmutated control green. `pnpm typecheck` 0 errors; eslint + prettier clean; 731 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 6 — 2 🟡, 2 🟢, none executable. Addressed in `93b33dae`.**

```audit-claims round=6 audited=8526826f56..93b33dae0a
1. 🟡1 — narrowing the allowlist to one member took the namespace-segment coverage to zero (a mutant hardcoding the namespace was killed by 2 cases at 58602693c4 and SURVIVED a green 21-case suite at 8526826f56), and the replacement's comment claimed it was "a STRONGER guarantee, not a weaker one"; claimed fixed by extracting and exporting `buildProbeKey`, which a test drives with two namespaces config cannot produce, and by correcting the false comment.
2. 🟢3 — the round-5 relative-counter fix did not land, because restoring two deleted tests FROM HEAD reverted it; claimed re-applied and this time verified by grepping for the pattern after applying rather than trusting the edit.
3. 🟢4 — `ReadonlySet<string>` is erased at runtime so an importer could `.add()` into the object `probeNamespace` reads; claimed fixed by a frozen array plus a private lookup Set, with `Object.isFrozen` asserted.
4. 🟡2 — the per-mutant kill table had re-staled, one row staled by round 5; claimed fixed in FORM by deleting the counts entirely, the same remedy applied to the running total in round 5.
```

**The seam extraction immediately broke `pnpm typecheck`, and that is worth recording as a free negative control** on an instrument this ladder has quoted eight times: `sysRedis.set` takes a key from a closed template-literal union, and returning `${string}:${string}` widened the key out of it. Made generic in the prefix so the literal prefix survives extraction. The gate can go red on this file.

**On 🟢3 — the mechanism is more useful than the fix.** The counters *were* replaced in round 5. Then the splice accident forced me to restore two tests from `HEAD`, which predated that fix, so the restore silently un-fixed them — and the commit message asserted otherwise. A fix can be un-fixed by a restore, and only re-deriving the state afterwards catches it.

15 isolated mutants, all killed, control green before and after — including the two new ones (`namespace-hardcoded-in-key`, `allowlist-not-frozen`). 22 cases. `pnpm typecheck` 0 errors; eslint + prettier clean; 732 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

ZacxDev:
**Audit round 7 — 2 🟡, 1 🟢, none executable. Addressed in `76849416`.**

```audit-claims round=7 audited=93b33dae0a..768494164b
1. 🟡B — round 6 split the allowlist into an asserted frozen ARRAY and a consulted private SET, which re-opened the growth hole round 5 closed (growing only the Set armed a namespace while the ledger kept asserting ['prod'], green 22/22); claimed fixed by collapsing back to ONE object via `.includes`, and by widening the behavioural case's candidates to include realistic additions with the expectation DERIVED from the export.
2. 🟡A — the namespace-derivation kill restored in round 6 covers the FUNCTION but not the CALL SITE, and was reading as fully closed; claimed fixed by RECORDING the residual in both the source and the test rather than papering over it, including the note that adding a second allowlist member re-opens it and must come with a two-namespace case. NOT claimed closed.
3. 🟢C — nothing asserted that the operator-facing error message renders the allowlist, so a mutant blanking the join left a message naming nothing; claimed fixed by an assertion derived from the export.
```

**🟡B was mine, and it is worth naming the shape:** round 5 closed the growth direction, round 6's freeze fix re-opened it *one round later* by introducing a second copy of the same list, and only a delta round could have seen that. Two objects that must agree is how a ledger's guarantee goes false.

**🟡A is deliberately not closed.** With one allowlist member, `namespace` is always `'prod'` on every path config can reach, so no test driven through the public surface can distinguish a hardcoded argument from a derived one. Closing it needs a test-only injectable resolver, which is more machinery than a temporary probe warrants. **It is recorded in the source, in the test, and here — so it reads as open rather than absent.** The confirming mutant (hardcoding the call-site argument) is included in the battery below and is reported as SURVIVING, on purpose.

15 isolated mutants all killed with the control green before and after, plus four targeted at this round: `allowlist-grown` ✅, `grow-only-the-lookup` ✅ (this is the one that defeated the two-object form), `rendered-allowlist-blanked` ✅, and `callsite-hardcoded` — **SURVIVES, known and documented**. 22 cases. `pnpm typecheck` 0 errors; eslint + prettier clean; 732 passed / 7 skipped over the named population.

ZacxDev:
## 🔴 Audit ladder STOPPED after round 7 — by operator decision, not by a clean round

Recording this explicitly, because a ladder that ends on a decision and one that ends on a clean round look identical afterwards, and they mean opposite things. **Round 7 was not clean.** Its findings were fixed in `76849416`; no round 8 was run.

**Why it was stopped.** No 🔴 in any of the seven rounds. Nothing executable found since round 4. Rounds 6 and 7 each found a regression introduced by the *previous round's fix* — the point at which a ladder is mostly auditing scaffolding it wrote itself. Every round's advisory verdict was "safe to merge".

### Known open, deliberately not fixed

**The call-site namespace argument is uncovered.** `buildProbeKey` is proven to derive its key from the namespace argument, but nothing proves `runProbe` *passes* the derived value — hardcoding that third argument leaves the suite green. With one allowlist member, `namespace` is always `'prod'` on every path config can reach, so no test driven through the public surface can distinguish the two. Closing it needs a test-only injectable resolver, which is more machinery than a temporary probe warrants.

It is recorded in the source (`moderation-cache-probe.ts`, on `buildProbeKey`), in the test, and here. **If a second allowlist member is ever added, this re-opens and must come with a two-namespace behavioural case.** The confirming mutant is in the battery and is reported as SURVIVING, on purpose.

### What the ladder actually bought

Seven rounds, eight defects that a green suite could not see:

1. The flag-off test was **vacuous** — a 20 ms sleep shorter than the probe's first lazy `import()`.
2. The probe's own `source` clamp was **never exercised** — the caller clamped first.
3. Every boolean-shaped "off" token (`false`, `off`, `0`…) **armed** the probe under that name.
4. The denylist that fixed #3 was itself walkable — `y`, `n`, `none`, `nil`, `2` went straight through.
5. `preview` was an allowlist member that **defeated the invariant the allowlist enforces**.
6. Dropping it **moved** the hazard rather than closing it: the preview pipeline clones `civitai-next`'s ConfigMap wholesale.
7. Two of three members were **unreadable when armed** — no ServiceMonitor anywhere but prod.
8. Narrowing to one member **silently zeroed** the namespace-segment coverage the previous two rounds had built.

Plus three false factual assertions in shipped comments, each of which would have been believed by the next maintainer, and one round where a `git checkout` of an older ref **silently un-did a fix** while the commit message asserted it had landed.

### State at `76849416`

22 cases · 15 isolated mutants killed, control green before and after · `pnpm typecheck` 0 errors · eslint + prettier clean · 732 passed / 7 skipped over `src/server/integrations/__tests__`, `src/server/prom/__tests__`, `src/server/services/orchestrator/__tests__`, `src/env/__tests__`, `src/server/redis/__tests__`.

Ships inert: `EXTERNAL_MODERATION_CACHE_PROBE` defaults to empty, is set in no ConfigMap or Secret in the GitOps repo, and `probeNamespace()` returns `null` on empty before any diagnostic fires. Arming it is a separate config change on `civitai-dp-prod`.

PR #4603, opened 2026-09-03
krea2 custom checkpoint support

### LINKED ISSUE #4602: Krea 2: allow community checkpoints in the generator
Krea 2 is still `modelLocked`, so the picker pins to the official build and nothing in the ecosystem can enter an auction. There are a few full finetunes of Raw on the site now (Wulver, LUSTIFY, Realia, Krealism, Event Horizon, Ratatoskr) and none of them can run on-site.

The comfy path already takes a `diffusionModel` AIR for edit, and the raw/turbo inputs have the same field in the client types, so this should be the same change as the Z-Image unlock (d68535e2). PR attached.

kaydaxter:
Same as d68535e2 for Z-Image. Closes #4602.

- handler: pass the selected checkpoint as `diffusionModel` on raw/turbo, like edit already does
- graph: unknown version ids get the raw control set instead of fal
- constants: drop `modelLocked` from Krea2

Official versions keep the graph default. I can't exercise the orchestrator side from outside; I have a 12.8B finetune published to test with once it's deployed.

### DIFF
--- a/packages/civitai-shared/src/basemodel.constants.ts
+++ b/packages/civitai-shared/src/basemodel.constants.ts
@@ -1123,7 +1123,7 @@ export const ecosystemSupport: EcosystemSupport[] = [
   { ecosystemId: ECO.Ernie, supportType: 'generation', modelTypes: checkpointAndLora },
   { ecosystemId: ECO.Ernie, supportType: 'training', modelTypes: loraOnly },
 
-  // Krea 2 - base/turbo comfy variants support LoRA and community checkpoints (medium/large FAL tiers do not); LoRA training via AI-Toolkit
+  // Krea 2 - raw/turbo comfy variants support LoRA and community checkpoints (medium/large FAL tiers do not); LoRA training via AI-Toolkit
   {
     ecosystemId: ECO.Krea2,
     supportType: 'generation',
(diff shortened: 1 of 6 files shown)


kaydaxter:
Quick follow-up on timing. The featured checkpoints auction that ended today (Sept 4, 00:00 UTC) was finalized at 00:01, about 16 hours before this landed in production with the 16:05 UTC deploy. At that point Krea 2 was still locked, so the Krea 2 winners went into FeaturedModelVersion but got filtered out before the CoveredCheckpoint sync. They show as featured until the 11th but can't be generated with.

Wulver (version 3257037) is one of them, and there may be others in that batch. Would it be possible to backfill CoveredCheckpoint for the Krea 2 winners of that auction, or re-run the winners step for Sept 4? On my side the version is published, Standard, RentCivit is set and the files are scanned, so it should pick up as soon as the row is there.

Thanks again for the quick merge.

bkdiehl:
@kaydaxter I've backfilled so that Krea 2 winners are now usable in the generator. Your model should be available to use in the generator directly from the model page. Sometimes it takes a while for meilisearch updates to propogate, so your model might not appear in the model picker for a while still.

PR #4605, opened 2026-09-03
feat(articles): restore the category bar on the articles feed

JustMaier:
Brings back the category tag bar at the top of `/articles`.

## What happened to it

PR #4141 (`020ba963c5`) removed the category bar from every feed-like surface at once and deleted `TagScroller`, `ImageCategories`, `PostCategories`, `ArticleCategories` and `Model3DCategories`. PR #4222 restored it **for models only** (ClickUp 868kumr1c, after Daz pushed back). Articles were collateral in a removal aimed at images and were never separately judged. This restores the articles half of `020ba963c5`.

## It is the taxonomy, not a curated list

Justin: *"The article one should be exactly what they were before. Ellie's list is specifically for the tag bars that we have above images."*

The code agrees: `ArticleUpsertForm` builds its Category dropdown from the same query this bar uses (`tag.getAll`, `categories: true`, `entityType: [Article]`), so the bar shows exactly the categories an author could have picked. That alignment is why no flag and no `Feed_TagBar_Click` instrumentation — that series measures the image bar, whose fate is being decided on it, and this is navigation over a fixed set.

## Verified against prod before building

- `article category` (tag 128630) has 14 children; **13 render**. `event` has no `Article` in its `target`, so `tag.getAll` already excludes it — 7 articles ever, 0 in 90 days. "Drop event" needed no change.
- Order is all-time usage: `getTags` derives `TagSort.MostArticles` from the entity type, and `TagMetric` holds only an `AllTime` row.
- `getArticles` filters `?tags=` with `toa."tagId" IN (...)` — a **union**, not an AND.

## Two deliberate differences from what was deleted

- **Single-select**, where the old bar stacked on ctrl-click. Not the image bar's reason (union, not AND) — it is that an article carries exactly one category, so the union of two is not a thing a reader asked for.
- **An `All` chip**, which widens a `?tags=` deep link back out. That is the job `ActiveTagFilter` was mounted on this page to do in #4517, so the bar takes over the mount and stands `ActiveTagFilter` up in the one state the chip is absent: a settled-empty category list. `ActiveTagFilter`'s own mount guard loses `/articles` and gains the reason; the mount is pinned in this component's suite instead.

Built on the shared `TagChipRow`, so the CLS height reservation cannot go missing between copies a third time.

`activeId` uses a sentinel for a `?tags=` the bar cannot draw (several ids, or one that is not a category): `TagChipRow` fills the `All` chip when `activeId` is undefined, and `All` means unfiltered, so the obvious version lights it over a narrowed feed.

## Testing

13 tests, all driven red on purpose before being trusted:

| Mutation | Result |
|---|---|
| `entityType: TagTarget.Model` | `× asks for article categories, not model ones` |
| sentinel removed (`length === 1 ? tagIds[0] : undefined`) | 2 failed |
| `placeholder={null}` | `× falls back to the clear control when the category list comes back empty` |
| `loading={false}` | 3 failed |
| mount commented out on the page | `× src/pages/articles/index.tsx renders <ArticleCategories />` |
| pristine control | `Tests 12 passed (12)` |

Full suite: `Test Files 1 failed | 1583 passed | 3 skipped (1587)`, exit 1. The one failure is `appListingMenuSurface.test.ts`, which **fails identically on this branch with every one of my changes stashed** — it compares Windows paths against forward-slash literals. Not from this diff. `blocks.router.workflow.test.ts` collected **370 tests**. `typecheck: OK — 0 type errors`. ESLint clean on the four changed files (repo-wide lint is 358 pre-existing errors).

ClickUp: https://app.clickup.com/t/868m146vj

https://claude.ai/code/session_01XHZuQDTCG159qcPrCkP7SF

### DIFF
--- a/src/components/Article/Infinite/ArticleCategories.tsx
+++ b/src/components/Article/Infinite/ArticleCategories.tsx
@@ -1,0 +1,90 @@
+import { useArticleQueryParams } from '~/components/Article/article.utils';
+import { ActiveTagFilter } from '~/components/Tags/ActiveTagFilter';
+import type { TagChipRowItem } from '~/components/Tags/TagChipRow';
+import { TagChipRow } from '~/components/Tags/TagChipRow';
+import { useCategoryTags } from '~/components/Tags/tag.utils';
+import { TagTarget } from '~/shared/utils/prisma/enums';
+
+/**
+ * Category row for `/articles`, writing `?tags=<id>`.
+ *
+ * This bar is NOT the `/images` and `/videos` one. Those carry a hand-curated chip set
+ * (`feed-tag-bar.constants`) that exists to be measured and can be switched off on the
+ * `feedTagBar` flag; this one is the article CATEGORY taxonomy — the `article category`
+ * system tag's children, moderator-curated, the same set the article editor makes an
+ * author pick from. It is navigation over a fixed set, so it carries no flag and no
+ * click instrumentation: `Feed_TagBar_Click` is the series the image bar's fate is being
+ * decided on (868kv1b9m), and mixing a differently-motivated surface into it would move
+ * a number that decides something else.
+ *
+ * Ids, not names: `?tags=` on this feed carries tag ids (`numericStringArray`), unlike
+ * `/models`' `?tag=`, which carries a name. `CategoryTags` uses the name as its row id
+ * for that reason; here the row id is the real tag id.
+ *
+ * Single-select, unlike the ctrl-click stacking the deleted `TagScroller` had. Not for
+ * the reason the image bar gives: `getArticles` filters `?tags=` with `toa."tagId" IN
+ * (...)`, so several ids there are a UNION and would widen this feed rather than empty
+ * it. It is that an article carries exactly one category — `ArticleUpsertForm` makes the
+ * author pick one from this same list — so the union of two of them is not a thing a
+ * reader asked for, and no article is in both.
+ *
+ * The `All` chip is what widens a `?tags=` deep link back out. `ActiveTagFilter` stands
+ * in wherever that chip is not on the page — which here is the settled-and-empty chip
+ * list, the state a failed category fetch leaves the bar in permanently. Neither may
+ * take the only escape hatch from a `?tags=` deep link with it (868kuq3jk).
+ */
+export function ArticleCategories() {
+  // The same hook the feed itself filters through (`ArticlesPage` passes this `query`
+  // straight to `ArticlesInfinite`). Reading `router.query` directly here instead would
+  // give the chips and the feed two different parsers for one param, and the symptom of
+  // them drifting is a chip that looks unselected while the feed is filtered.
+  const { query, replace } = useArticleQueryParams();
+  const tagIds = query.tags ?? [];
+
+  const { data: categories, isLoading } = useCategoryTags({ entityType: TagTarget.Article });
+
+  // A chip is active only when it is the sole filter — anything else came from a deep
+  // link this bar cannot represent, and lighting one chip of several would misdescribe it.
+  //
+  // But `undefined` is what fills the All chip, and All means UNFILTERED. A `?tags=` the
+  // bar cannot draw — several ids, or one that is not a category — would otherwise light
+  // All over a narrowed feed. So those states get an id no chip can hold: every chip
+  // stays grey, which is the truth, and All still clears the param.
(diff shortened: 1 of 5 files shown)


JustMaier:
Five-lane `civitai-review` run at `a105cb88fb`; fixes pushed as `2cce6be356`.

**Mechanism, stated because it matters for how much the review is worth:** none of the five ran as a real `civitai-*-review` subagent type — those do not resolve from the cwd this session was started in. Each lane was a general-purpose agent instructed to read its own definition from `.claude/agents/civitai-*-review.md` and report through `SendMessage`. All five delivered. They ran against a separate detached worktree at the head sha; `git status --porcelain` on the implementation tree afterwards was **empty**.

## The finding worth having

**Test lane: `chipsGone` could be deleted whole and all 13 tests stayed green.** Replace `placeholder={chipsGone ? <ActiveTagFilter …/> : null}` with the unconditional form and nothing goes red — the single test that set `tagsLoading` carried no `?tags=`, so `ActiveTagFilter` returned `null` under both. What ships under that mutation is the first-paint flash of a clickable "Clear 1 tag filter" that the comment above the variable exists to forbid. The lane reached it by reading; I confirmed it by running the mutation.

**The fix round then reproduced the same class of mistake.** My first version of the hidden-preferences test passed over its own mutant, 15 of 15: the real `useApplyHiddenPreferences` returns `items: []` whenever it is loading, so `!categories.length` already holds the chips back and dropping `|| loadingPreferences` changes nothing observable — *unless* a `?tags=` is set, where it is `chipsGone` the term protects. Both the measurement and the reason are now in the test.

Controls, all five mutants killed, pristine 15 passed:

| Mutation | Result |
|---|---|
| unconditional `placeholder` | 1 failed |
| `isLoading` without `loadingPreferences` | 1 failed |
| `data-active` removed | 2 failed |
| mount renamed to `<ArticleCategoriesSkeleton />` | 1 failed |
| `limit: 100` dropped | 1 failed |

## Other findings, acted on

- `activeLabels()` sniffed Mantine's `--button-bg` for the substring `blue`. `TagChipRow` now writes `data-active` from the `active` prop and the suite selects on that — selection is not a colour, and a renamed token would have made three assertions return "nothing selected" and pass.
- The mount guard's `line.includes('<ArticleCategories')` is a prefix match: `<ArticleCategoriesSkeleton />` satisfied it, and so did the import assertion. The sibling `ActiveTagFilter` guard documents that exact hazard, and it was reintroduced here anyway. Now `/<ArticleCategories[\s/>]/`.
- Docblock overclaimed "the same set the article editor makes an author pick from" — same taxonomy, not the same rows: `ArticleUpsertForm:163` runs its own copy of the query without hidden preferences.
- The single-select justification reasoned from the author's model to a conclusion about readers. Two chips on the old bar were a working "either kind" filter, because the param is a union. It goes because ctrl-click is unlabelled, undiscoverable and impossible on touch — not because nobody wants it.

## Clean

- **Perf, measured on the prod replica:** 0.226 ms / 190 buffers / 13 rows; 4.03 ms at the single worst hidden-tag set in the database (6,566 ids, p99 is 16). Never enters the 567k-row `Tag` table as a driving relation. Behind a 60s Redis `cacheIt`, a 1h SQL-keyed listing cache and a 60s CF edge cache for anon.
- **Correctness:** the nsfwLevel parent rollup *is* applied on this path, plus a forced PG filter on the own column; the `activeId` sentinel cannot collide, leak into the URL, or make a chip un-clickable; `removeEmpty` drops nothing but `tags`.

## Two findings NOT fixed here, raised for a decision instead

1. **The bar had a second mount** — `Collection.tsx:488`, the article view of a collection — and `?tags=` still filters that feed with no control to clear it. Not restored here, because **#4222 did not restore the models bar there either**; page-only is the precedent this follows.
2. **`useZodRouteParams` reads with `safeParse` and writes with `schema.parse`.** On `/articles?period=month` (lowercase) every chip click, including `All`, throws a ZodError out of the handler and does nothing — reproduced live. Pre-existing and shared with `/models`, `/images`, `/videos`, `/posts`; this bar is just the first to put the escape hatch on that path.

Verification after the fixes: `typecheck: OK — 0 type errors`; ESLint clean on the changed files; `Test Files 1 failed | 1583 passed | 3 skipped (1587)`, the one failure being `appListingMenuSurface.test.ts`, which fails identically with every change stashed.

PR #4608, opened 2026-09-03
fix(models): bump Model."updatedAt" when restoreModelById un-deletes a model

ZacxDev:
## The defect

`restoreModelById` un-deletes a model with raw SQL:

```sql
UPDATE "Model"
SET "deletedAt" = NULL, "deletedBy" = NULL,
    "status" = CASE WHEN "publishedAt" IS NULL THEN 'Draft'::"ModelStatus" ... END
WHERE id = $1 AND "status" = 'Deleted'::"ModelStatus"
```

It is `$queryRaw`, so Prisma's `@updatedAt` never fires. The restored row keeps the timestamp it carried **while it was deleted** — which is the deletion instant itself, because `deleteModelById` writes through the Prisma client (`tx.model.update` with a nested `modelVersions.updateMany`) and so *does* bump it. Measured rather than assumed: an enumeration of every raw `UPDATE "Model"` in `src/` finds this restore statement is the only one that touches a Model's `'Deleted'` status; every other delete path is a client write.

`remove-old-drafts` reaps

```sql
status IN ('Draft','Deleted') AND m."updatedAt" < now() - INTERVAL '30 days'
```

and cascade-deletes the model together with its versions, files and training data. That delete is irreversible.

So the naive reading — "restoring exposes the model" — is **wrong**, and it is worth saying why explicitly, because two earlier revisions of this PR said it.

### What restoring does and does not cause

`'Deleted'` is *already* in the reaper's `status IN ('Draft','Deleted')` set. The clock therefore runs while the model sits deleted: a low-download model is destroyed the night after deletion + 30 days, **still `Deleted`, never restored**. Restoring changes exactly one term — `status` goes `Deleted` → `Draft` (still in the set) or `Unpublished`/`Scheduled` (out of it). Nothing else the predicate reads moves: the `ModelVersion` statement is raw SQL and does not bump `mv."updatedAt"` either, and `downloadCount` / `availability` / the `ModelMetric` join are untouched.

**So the post-restore candidate set is a strict subset of the pre-restore one. Without the bump, restoring can never make a model reapable that was not already.** A "restored after eight months, gone that night" scenario cannot occur — surviving months of nightly reaping requires a sparing condition (`downloadCount >= 10`, `availability = 'Private'`, no `ModelMetric` row, live version/file activity), and every one of those still holds after the restore.

### What the bump actually buys — both real

1. **A partly-spent clock.** Deleted day 0, restored day 29 → without the bump the model is reaped the night of day 30/31, one day after restore. The bump turns whatever remains into a full `REAP_AGE_DAYS`, which is what a restored model is entitled to.
2. **The `old-draft` warning — arguably the stronger reason, and absent from every earlier revision of this PR.** That notification warns on `Draft` **only** (a `Deleted` model is deliberately never warned), and its band evaluates **once**, at `U + OLD_DRAFT_NOTICE_DAYS` (23 days), never re-evaluated for that `U`. A model restored carrying its pre-restore `U` is now `Draft` — warnable for the first time — but its band is already in the past, so it is cascade-deleted **unwarned**. The bump re-arms the band, which is the only way the user ever hears about it.

Every claim above was re-derived from source: the predicate in `remove-old-drafts.ts`, the constants in `draft-reaping.ts` (`REAP_AGE_DAYS` 30, `OLD_DRAFT_LEAD_DAYS` 7, `OLD_DRAFT_NOTICE_DAYS` 23), the `Draft`-only status term and once-evaluated band in `model.notifications.ts`, and the `updateMany` scope in `deleteModelById`.

One precision note, since an earlier revision got it backwards: `deleteModelById`'s nested `modelVersions.updateMany` is scoped to `status IN (Published, Scheduled)`, so a **Draft-only model has no `ModelVersion` row bumped at delete time**. Those timestamps are *older*, making the activity fences less protective, not more.

**Why getting this right matters:** a maintainer who tests the old stated scenario finds no such model, concludes the justification is fiction, and deletes a line that is genuinely load-bearing for both cases above.

## The change

Add `"updatedAt" = now()` to that statement. Restoring a model *is* a write to the row, so the bump is what the column is supposed to mean as well as the thing that stops the reaper eating it.

This is the same defect class and the same one-token fix as #4595, which closed the two sweep sites and named this one in its body as a known unfixed sibling. **This is the third and last raw-SQL-into-`Draft` site in the tree** — an enumeration of every raw `UPDATE "Model"` found exactly three writing `Draft`; `republish-orphaned-drafts.ts` and `process-scheduled-publishing.ts` write `Published` and are out of scope.

## The guard ledger

`no-unbumped-draft-status-write.test.ts` (introduced by #4595) explicitly excluded `restoreModelById`, on the stated grounds that including it would make the guard permanently red. This PR removes that reason, so the file is added to the ledger and the exclusion note is deleted. The guard now covers all three sites.

### 🔴 The extractor was narrower than its own description — read this part

This is the most transferable thing in the PR, so it is stated plainly rather than left as a diff.

The guard's doc comment states a **rule**: *"a raw-SQL write that moves a `Model` into `Draft` must also set `"updatedAt" = now()"`."* Its implementation did not enforce that rule. It carried two unstated assumptions:

1. it anchored on `UPDATE "Model" ` **with a trailing space**, so it could not see a statement that breaks the line after the table name — exactly how `restoreModelById` writes it (the closing quote is what excludes `UPDATE "ModelVersion"`; the space was never doing that job);
2. it took only the **first** match per file.

`model.service.ts` holds **seven** raw `UPDATE "Model"` statements, one of which drafts. Measured on `origin/main`: the space anchor skips the drafting statement, and first-match then lands on `captureMinorFlagSnapshot`'s `meta` write — a different statement, in a different function, unrelated to the reaper.

**The two defects compose, which is why neither is visible alone.** The drafting statement is the *first* of the seven, so a first-match rule on its own would have found it. Fixing either half in isolation yields a guard that is right for the wrong reason.

**Why no review round could have caught it.** Both files the extractor was written against hold **exactly one** raw `UPDATE "Model"` each, and both spell it `UPDATE "Model" m` — with a space, because of the alias. Both assumptions were true of the entire corpus the guard could see. A doc comment claiming a general rule, over an implementation covering two files that happen to satisfy two unstated assumptions, cannot be told apart from a correct guard by reading it, by running it, or by auditing it — only by pointing it at a file it was not written for. That is what adding the third site did, and it is the argument for re-deriving a ledger extractor against each new member rather than trusting the green it reports.

It now reads *every* raw `UPDATE "Model"` statement in a ledger file and binds the rule to the ones writing `Draft`.

### A third extractor defect, found while fixing the first two

The terminator was "the next backtick". On a template whose interpolation is itself a tagged template — `${cond ? Prisma.empty : Prisma.sql`...`}`, the shape of the description write in `model.service.ts` — that backtick is the **nested** one, so the extraction **truncates**. Measured: that statement extracted 156 characters and stopped mid-ternary; it now extracts 207.

Truncation *removes* text, so it can drop the `'Draft'` literal and silently take a statement out of scope entirely. My previous comment claimed the terminator "can only pull MORE text into scope, so it fails closed" — that described the opposite of the real failure mode.

`templateEnd` now tracks `${...}` by brace depth, recursing through nested templates and skipping quoted strings in the JS; an unterminated template is an error rather than a skip.

**And the control was upgraded from "at least one drafting statement" to an exact per-file count.** "At least one" is sufficient only while every ledger file holds exactly one — a property of this corpus, not of the guard. With two, one can drop out of the extractor's view and the file still reports a non-empty list. The exact count fails both when a statement disappears *and* when one arrives unwatched.

**Proof, since this is the case the single-statement coincidence hides.** A second drafting statement planted in `model.service.ts`:

| planted statement | ledger count | result |
|---|---|---|
| plain spelling, **unbumped** | 2 | KILLED — *"drafting statement #2 does not bump"* |
| plain spelling, bumped | left at 1 | KILLED by the count control |
| **nested-template spelling, unbumped** | 2 | KILLED — *"drafting statement #2 does not bump"* |

The third is the one that matters. Under the old terminator it extracted as `UPDATE "Model" SET ${Prisma.sql`, carried no `'Draft'`, dropped out of the drafting set entirely — and the old "at least one" control **passed, silently green over an unbumped drafting statement.**

No `no-*` guard file was added or removed, so `no-lint-rules-script-drift` needs no change — verified green, along with the rest of `test:lint-rules` (30 files, 461 tests).

## Test coverage

Behavioural, not only source-scanning. `src/server/services/__tests__/restore-model-updated-at.service.test.ts` drives the real `restoreModelById` through the canonical `dbMock` (no direct mock of `~/server/db/client`) and reads the statement it actually issues:

- the bump is present, assigned **exactly `now()`**, and **unconditional**;
- the SET clause is pinned as an ordered **ledger**, so an added column write is caught as well as a removed one — `publishedAt` in particular, which this function's own status `CASE` reads;
- the deletion-column clears, the status `CASE` and the `status = 'Deleted'` predicate are pinned, so the fix has to be *additive*;
- a scope guard holds the change to `Model."updatedAt"` and leaves `ModelVersion` alone (with a positive control asserting that statement was issued at all);
- a positive control asserts the function actually restored a model and returned its owner, so no guard is a claim about a path that never ran.

### Red → green

Both new guards were watched fail on the pre-fix source (`origin/main` at `34510fb587`, service file reverted, test files in place):

```
Test Files  2 failed (2)
     Tests  4 failed | 10 passed (14)

FAIL  no-unbumped-draft-status-write > model.service.ts sets "updatedAt" = now() on every drafting statement
  AssertionError: src/server/services/model.service.ts: without the bump this model is immediately
  reapable by remove-old-drafts, which cascade-deletes it: expected 'UPDATE "Model"\n SET "deletedAt"…'
  to contain '"updatedAt" = now()'

FAIL  restoreModelById > bumps "updatedAt" so a restored model gets a real grace period before the reaper
  AssertionError: a model restored with a stale "updatedAt" is immediately reapable by
  remove-old-drafts, which cascade-deletes it

FAIL  restoreModelById > writes exactly "deletedAt", "deletedBy", "status" and "updatedAt" — no other column
FAIL  restoreModelById > assigns "updatedAt" exactly now(), unconditionally
  AssertionError: expected undefined to be 'now()'
```

At the first commit, green: **10 files, 215 tests passed** (the two new/changed files plus `no-lint-rules-script-drift`, `no-direct-shared-module-mock`, `no-wholesale-module-mock`, `no-io-in-transaction`, `reset-to-draft-without-requirements`, `remove-old-drafts`, and the two existing `model.service` suites).

### Mutation results

Each mutant edits the narrowest expression that can be wrong, and each was confirmed to die with **the specific guard's own assertion message**, not collaterally. `M0` is the unmutated control.

| # | Mutant | Result | Killed by |
|---|---|---|---|
| M0 | *(none — control)* | 14 passed | — |
| M1 | bump deleted | KILLED (4 failed) | ledger guard + 3 behavioural |
| M2 | `now() - INTERVAL '31 days'` | KILLED (1 failed) | **behavioural only** — ledger guard passes |
| M3 | bump made conditional on `publishedAt IS NULL` | KILLED (3 failed) | ledger guard + 2 behavioural |
| M4 | extra `"publishedAt" = now()` write added | KILLED (1 failed) | **behavioural only** — SET-clause ledger |
| M5 | bump demoted to a `--` SQL comment | KILLED (4 failed) | ledger guard + 3 behavioural (comment stripping works) |
| M6 | `"updatedAt" = now()` added to the `ModelVersion` statement | KILLED (1 failed) | scope guard |
| M7 | `'Draft'` literal changed to `'Unpublished'` | KILLED (2 failed) | ledger exact-count control |
| N2 | second drafting statement, unbumped (count 2) | KILLED | *"drafting statement #2 does not bump"* |
| N3 | second drafting statement, bumped (count left 1) | KILLED | exact-count control |
| N4 | second drafting statement behind a nested template, unbumped | KILLED | **invisible to the old terminator** |

Re-run in full against the **final** head rather than carried forward from the first commit, since the extractor is a mechanism change: **11 mutants across both batteries, all killed, 2 controls green.**

### Two known limits of the extractor, listed by the direction they fail

The extractor is text-anchored on `UPDATE "Model"`, and the guard's own doc comment now records both limits *with which way each one breaks*, because a limits list that mixes fail-closed and fail-open reads as uniformly harmless:

- **Fails closed** — a comment or string that merely *quotes* the statement is counted as one. The exact-count control catches it, and the failure message now says to **reword the prose, not raise the count**, since a maintainer who raised it would install a phantom statement in the ledger permanently. (`model.service.ts`'s comment block already contains `'Draft'` and `"updatedAt" = now()`, so it is one mention away from this.) Watched it render: planting that comment turns the guard red with the new text.
- **Fails open** — an `UPDATE "Model"` written *inside another one's interpolation* is absorbed into the outer statement rather than enumerated, since `templateEnd` now correctly reads to the outer close. Measured: outer bumped, inner drafting and unbumped ⇒ `drafting=1, bumped=1`, count 1, all green. Deliberately **not** engineered around — nobody writes that shape and a parser change costs more risk than it buys — but recorded with the measurement so the next person weighing that trade is not guessing.

M2 and M4 are the ones that matter for the choice of test: both are **invisible to the source-text ledger** — M2 contains the literal the ledger greps for, and M4 is an addition no `toContain` can see. Only the behavioural test catches them. M2 in particular is the defect spelled as a fix.

## Verification scope, stated honestly

- **Typecheck**: `pnpm typecheck` is clean (0 errors), re-run against the final head. The instrument was validated — a planted `const x: number = 'nope'` in `model.service.ts` was reported (`TS2322` at the planted line). It does **not** cover the added test file: `tsconfig.json` excludes `src/**/__tests__/**` by design, and a planted error there was correctly not reported. Stating that rather than implying wider coverage.
- **ESLint**: run the way `lint.yml` does — added-file tier is blocking, so the added file was linted with `--rule '{"@typescript-eslint/no-inferrable-types":"error"}'` forced explicitly, resolving `@typescript-eslint` 5.62.0 (the lockfile version). Clean. The invocation's negative control was watched go red on a planted inferrable type. `model.service.ts` carries 9 pre-existing findings in the report-only modified tier — **identical count at the base commit**, so this change introduces none.
- **Prettier**: clean on both added and modified files.
- **Not verified**: no run against a real database. The statement's SQL is unchanged apart from one additional assignment of a nullary function to a `timestamp` column, so this is a low-risk gap, but it is a gap.

At the final head, green at **124 tests / 8 files**: `services/__tests__/{no-unbumped-draft-status-write, restore-model-updated-at.service, no-lint-rules-script-drift, no-direct-shared-module-mock, no-io-in-transaction}.test.ts`, `jobs/__tests__/{reset-to-draft-without-requirements, remove-old-drafts}.test.ts`, `notifications/__tests__/old-draft-reaper-parity.test.ts`. Counts are quoted with the file set they were measured over, so they can be reproduced.

### DIFF
--- a/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
+++ b/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
@@ -53,10 +53,27 @@ const LEDGER = [
  *    breaks the line after the table name, which is how `restoreModelById`
  *    writes it. The closing quote is what excludes `UPDATE "ModelVersion"`;
  *    the space was never doing that job.
- *  - `model.service.ts` holds THREE raw `UPDATE "Model"` statements and only
+ *  - `model.service.ts` holds SEVEN raw `UPDATE "Model"` statements and only
  *    one of them drafts. Taking the first match would have anchored this guard
  *    on `captureMinorFlagSnapshot`'s meta write — a different statement, in a
  *    different function, with nothing to do with the reaper.
+ *
+ * 🔴 The two defects COMPOSE, which is why neither was visible on its own. The
+ * drafting statement is the FIRST of the seven, so a first-match rule alone
+ * would have found it; it is the space anchor that skips it, and only then does
+ * first-match land on the meta write. Fixing either half in isolation would have
+ * produced a guard that happened to be right for the wrong reason.
+ *
+ * 🔴 And the reason no review round could have caught this: both files this
+ * extractor was originally written against hold EXACTLY ONE raw `UPDATE "Model"`
+ * each, and both spell it `UPDATE "Model" m` — with a space, because of the
+ * alias. Both assumptions were therefore true of the entire corpus the guard
+ * could see. A guard's doc comment states a RULE ("a raw-SQL write that drafts a
+ * Model"); its implementation covered two files that happened to satisfy two
+ * unstated assumptions. Nothing short of pointing it at a file it was not
+ * written for can tell those apart — which is exactly what adding the third site
+ * did. When extending a ledger like this, re-derive the extractor against the
+ * NEW file before trusting the green it reports.
  *
  * Each statement is read to the end of its tagged template (the next backtick).
  * That over-reads on a template carrying an interpolated expression, which is
--- a/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
+++ b/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
@@ -69,9 +69,26 @@ const LEDGER: { file: string; draftingStatements: number }[] = [
  * this file: that statement extracted 156 characters and ended mid-ternary.
  *
  * So the scan tracks `${...}` interpolations by brace depth, recursing through
- * nested templates and skipping quoted strings inside the JS. Known limits, both
- * absent from this corpus: a brace or backtick inside a regex literal or a `//`
- * comment within an interpolation would confuse the depth count.
+ * nested templates and skipping quoted strings inside the JS.
+ *
+ * KNOWN LIMITS, none present in this corpus. 🔴 They are listed WITH THE
+ * DIRECTION THEY FAIL, because a limits list that mixes the two is more
+ * dangerous than no list — a reader assumes the whole list is as harmless as
+ * whichever entry they check first:
+ *
+ *  - FAILS CLOSED — a brace or backtick inside a regex literal or a `//` comment
+ *    within an interpolation confuses the depth count. The scan then runs long
+ *    or never terminates, which surfaces as the unterminated-template error or a
+ *    wrong count. Loud, and it cannot un-watch a statement.
+ *  - 🔴 FAILS OPEN — an `UPDATE "Model"` written INSIDE an interpolation of
+ *    another `UPDATE "Model"` template. Because the scan now correctly reads to
+ *    the OUTER close, the inner statement is absorbed into the outer one rather
+ *    than enumerated separately. Measured: outer bumped, inner drafting and
+ *    unbumped gives `drafting=1, bumped=1` and an exact count of 1 — all green
+ *    over an unbumped drafting statement. Deliberately NOT engineered around:
+ *    nobody nests an `UPDATE "Model"` inside another one, and a parser change to
+ *    catch it would cost more risk than the shape is worth. Recorded so the next
+ *    person weighing that trade has the measurement rather than a guess.
  */
 function templateEnd(source: string, start: number): number {
   let i = start;
@@ -226,12 +243,22 @@ describe('a raw SQL write that drafts a Model bumps its "updatedAt"', () => {
   // The exact number fails in BOTH directions: a statement that disappears from
   // the extractor's view, and one that ARRIVES unwatched. Bumping the number is
   // how you record a deliberate decision about a new site.
+  //
+  // 🔴 But an INCREASE has two causes and only one of them is a new site. The
+  // extractor anchors on the raw text `UPDATE "Model"`, so a `//` comment or a
+  // string that merely QUOTES the statement counts as one. That is not
+  // hypothetical here: the comment block on `restoreModelById` already contains
+  // `'Draft'` and `"updatedAt" = now()`, so it is one `UPDATE "Model"` mention
+  // away from tripping this — and prose about this very guard is exactly the
+  // kind of text that would do it. The failure message therefore has to name
+  // that cause, because a maintainer who follows a bare "bump the count" would
+  // permanently install a phantom statement in the ledger.
   it.each(LEDGER)(
     '$file holds exactly the drafting statements this ledger claims',
     ({ file, draftingStatements: expected }: (typeof LEDGER)[number]) => {
       expect(
         draftingStatements(file).length,
-        `${file}: expected ${expected} raw UPDATE "Model" statement(s) writing Draft. Fewer means one dropped out of the extractor's view and is no longer checked; more means a new drafting site arrived — bump the count here once you have confirmed it carries the bump.`
+        `${file}: expected ${expected} raw UPDATE "Model" statement(s) writing Draft. FEWER means one dropped out of the extractor's view and is no longer checked. MORE means one of two things: either a real new drafting site arrived, in which case confirm it carries the bump and then raise the count here — or a comment or string literal now quotes the statement text, which this text-anchored extractor cannot tell from real SQL. If it is prose, REWORD THE PROSE; do not raise the count, or you install a phantom statement in this ledger permanently.`
       ).toBe(expected);
     }
   );
--- a/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
+++ b/src/server/services/__tests__/no-unbumped-draft-status-write.test.ts
@@ -15,63 +15,95 @@ import { describe, expect, it } from 'vitest';
  * irreversibly. A model drafted by raw SQL without the bump therefore enters the
  * reapable state with its entire 30-day grace period already spent.
  *
- * These two sites are checked as SOURCE TEXT rather than by executing them,
- * because one is an admin `temp/` endpoint whose handler stack would have to be
- * stood up to reach the statement. The job's statement is additionally covered
- * behaviourally-ish in `src/server/jobs/__tests__/reset-to-draft-without-requirements.test.ts`;
- * this guard exists so the endpoint is not the untested half of one fix.
+ * These sites are checked as SOURCE TEXT rather than by executing them, because
+ * one is an admin `temp/` endpoint whose handler stack would have to be stood up
+ * to reach the statement. Two of the three are additionally covered
+ * behaviourally: the job's statement in
+ * `src/server/jobs/__tests__/reset-to-draft-without-requirements.test.ts`, and
+ * `restoreModelById`'s in
+ * `src/server/services/__tests__/restore-model-updated-at.service.test.ts`. This
+ * guard exists so the endpoint is not the untested half of one fix, and so all
+ * three sites are held to one rule in one place.
  *
- * ⚠ SCOPE, stated so this is not read as wider than it is: this is a LEDGER of two
- * named files, not a repo-wide scan. It cannot see a third site that starts
- * drafting models by raw SQL tomorrow. `restoreModelById`
- * (`src/server/services/model.service.ts`) is a KNOWN sibling with the same
- * defect — it raw-SQL-sets `status = CASE WHEN "publishedAt" IS NULL THEN 'Draft'`
- * with no bump, so an un-deleted model lands in Draft with a clock older than its
- * deletion. It is deliberately NOT in this ledger: it is pre-existing, needs its
- * own change and its own review, and adding it here would make this guard
- * permanently red, which is how a guard stops being read.
+ * ⚠ SCOPE, stated so this is not read as wider than it is: this is a LEDGER of
+ * named files, not a repo-wide scan. It cannot see a fourth site that starts
+ * drafting models by raw SQL tomorrow. As of this writing the ledger is
+ * complete — an enumeration of every raw `UPDATE "Model"` in the tree found
+ * exactly these three writing `Draft`; `republish-orphaned-drafts.ts` and
+ * `process-scheduled-publishing.ts` write `Published` and are out of scope.
+ * Adding a site here is part of adding one to the tree.
  */
 const repoRoot = path.resolve(__dirname, '../../../..');
 
-/** Files whose raw `UPDATE "Model"` statement moves a model into Draft. */
+/** Files with a raw `UPDATE "Model"` statement that moves a model into Draft. */
 const LEDGER = [
   'src/server/jobs/reset-to-draft-without-requirements.ts',
   'src/pages/api/admin/temp/backfill-swept-trained-models.ts',
+  'src/server/services/model.service.ts',
 ];
 
 /**
- * The `UPDATE "Model"` statement's text, `--` comments stripped.
+ * EVERY raw `UPDATE "Model"` statement in a file, `--` comments stripped.
  *
- * Anchored on `UPDATE "Model" ` WITH the trailing space so it cannot match
- * `UPDATE "ModelVersion"`, and read to the end of the tagged template (the next
- * backtick). Comment stripping is load-bearing: both statements carry `--`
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Adversarial audit — round 1 (full): **safe to merge**, one 🟡 + two 🟢. All fixed in `f4b9ec235e` + `2ff95dc9d1`

Round 1 was a blind full audit against `53cb1f2dad`. It **reproduced every self-reported number
exactly** — red→green 4 failed/10 passed → 10 files/215 tests; mutation 7/7 killed with the
control green; `test:lint-rules` 30 files/461 tests; `model.service.ts` ESLint 9 findings at base
**and** head (so zero introduced), `@typescript-eslint` **5.62.0** resolved. It also enumerated
all **54** raw `UPDATE "Model"` statements across 21 files in `src/` independently: exactly
**3 production statements write `'Draft'`, and all three now carry the bump.** The arc is closed.

### 🟡 The stated exposure was unreachable — and this was the THIRD wrong framing of it

The reaper's status term is `IN ('Draft','Deleted')` and has been since the job's first commit
(`fe2e4025b4`), and `deleteModelById` bumps `Model."updatedAt"` to the deletion instant. So a
low-download model is **already** destroyed the night after deletion + 30 days, while still
`Deleted`. Restoring changes nothing else the predicate reads. **The post-restore candidate set is
a strict SUBSET of the pre-restore one — without the bump, restoring can never make a model
reapable that was not already.** The PR's original "eight months later someone restores it → gone
hours later" scenario cannot occur: surviving ~7 months of nightly reaping requires a sparing
condition, and every one still holds after the restore.

Worth recording, because it is the transferable part: **three successive framings, each more
precise, each false.** "A clock older than its own deletion" (the brief) → "carries the deletion
instant, reapable the moment it is restored" (commit 2) → the strict-subset reality. Each was
derived from a *story about what restoring does* rather than from what the predicate *reads*.

The two real justifications, neither of which appeared anywhere before:
- **A partly-spent clock.** Deleted day 0, restored day 29 → reaped the night of day 30/31, one
  day after restore, with the version fence expiring at the same instant.
- **The unwarned cascade-delete.** `old-draft` warns on **`Draft` only** — a `Deleted` model is
  deliberately never warned — and its band evaluates **once**, at `U + 23d`, never re-evaluated
  for that `U`. So a restored model becomes warnable for the first time, and the carried-over `U`
  is precisely what still denies it the warning. The bump re-arms it.

The comment now opens by instructing the reader to derive from the predicate before editing, and
records that three prior versions did not.

Also corrected: `deleteModelById`'s nested `modelVersions.updateMany` is scoped to
`Published`/`Scheduled`, so a Draft-only model has **no** version row bumped at delete — those
timestamps are *older*, making the fences less protective, not more. Conservative, so the
conclusion held, but it is now stated accurately.

### 🟢 → 🔴 in effect: the extractor's failure mode was TRUNCATION, and the old control went green over it

The guard's comment claimed extraction "can only pull MORE text into scope, so it fails closed".
False: the terminator was "the next backtick", which on a template containing a **nested tagged
template** truncates. `model.service.ts:3008` is exactly that shape.

**Measured, and this is the finding:** planting a *nested-template, unbumped* drafting statement
made the old extractor yield `` UPDATE "Model" SET ${Prisma.sql `` — no `'Draft'` literal, so the
statement dropped out of the drafting set entirely, and the old "at least one" control **PASSED —
silently green over an unbumped drafting statement.** That is the exact defect this guard exists
to catch.

Fixed by construction, not by wording: `templateEnd` now tracks `${…}` by brace depth, recurses
through nested templates and skips quoted strings; an unterminated template is an **error**, not a
skip. The control was also upgraded from "at least one" to an **exact per-file count**, because
the terminator fix alone still left the guard safe only by the corpus's single-statement
coincidence.

| planted second drafting statement | count | result |
|---|---|---|
| plain, unbumped | 2 | KILLED — *"drafting statement #2 does not bump"* |
| plain, bumped | 1 | KILLED by the count control |
| **nested-template, unbumped** | 2 | KILLED — *"#2 does not bump"* (old guard: silently green) |

🔴 **This is the third instance of one pattern in this guard**: a rule whose unstated assumptions
happened to hold across its entire visible corpus — first the trailing-space anchor, then
first-match-per-file, now the truncating terminator. Each was invisible to reading it, running it,
and auditing it; each was exposed only by pointing it at a shape it was not written for.

### 🟢 "three raw `UPDATE "Model"` statements" → seven

Already corrected in `5af4970068` before this audit ran, so the finding arrived stale. The
consequential half was independently confirmed: the old trailing-space anchor's first match is
`captureMinorFlagSnapshot` at line 2244, whose text contains no `'Draft'`, so a naive ledger
addition would have gone permanently red on the positive control.

### Verified, not findings

The behavioural test drives the **real** `restoreModelById` through the canonical `dbMock` — no
re-implementation — and its "exactly one `UPDATE "Model"`" / "exactly one `UPDATE "ModelVersion"`"
helpers are genuine positive controls. No adverse second-order effect from the bump: the
incremental search-index scan filters `status = Published`; the public Newest/Oldest feed sorts on
`mm."lastVersionAt"`, not `Model."updatedAt"`; the `updatedAt`-ordered lists show non-public
statuses only. `restoreUser`'s `restoreModels` uses a Prisma `updateMany` so `@updatedAt` fires,
and explicitly does not touch `Deleted` models.

### Post-fix re-verification

Full sweep re-run against the **final** head rather than argued forward: **11 mutants across both
batteries, all killed, 2 controls green**; 124 tests across 8 files; typecheck 0 errors; the three
changed files prettier-clean and ESLint-clean with `no-inferrable-types` forced. The payload SQL is
byte-identical across `53cb1f2dad` and `2ff95dc9d1` (`5d7e83c657a5f0fe`, 323 chars), so the two
new commits touched only the guard mechanism and the prose.

### Still true

No DB-backed vitest project exists, so the SQL guards remain spelled string comparisons and
nothing here was run against a real Postgres.

```audit-claims round=1 audited=53cb1f2dad..2ff95dc9
1. The justification is rewritten from the predicate: 'Deleted' is already reaped, so the post-restore candidate set is a strict subset of the pre-restore one.
2. The two real reasons are stated — a partly-spent clock, and the once-evaluated old-draft band that leaves a restored model unwarned.
3. The deleteModelById ModelVersion scoping (Published/Scheduled only) is corrected in the comment.
4. templateEnd tracks ${...} by brace depth, recurses through nested templates, skips quoted strings, and errors on an unterminated template.
5. The per-file control is an exact count rather than "at least one".
6. A nested-template unbumped drafting statement is now KILLED; under the old terminator it was silently green.
7. The payload SQL is unchanged from the audited tip.
```

PR #4609, opened 2026-09-03
feat(user-restriction): make the pending-review mute type a parameter

ZacxDev:
## What and why

Makes the **restriction type** a parameter on the pending-review mute path, so a future bot-account detection agent can file into the **existing** mute-review queue instead of a new board.

This is the enabling seam only — no detector, no cron, no bot logic.

`UserRestriction.type` is a free-text column with a `[type, status]` index, so a new type needs **no migration**. None is included, and none should be added for this.

### Main app — `src/server/services/user-restriction.service.ts`

`applyPendingReviewMute` takes an optional `type`, defaulting to `'generation'`. Both existing callers (`src/pages/api/mod/mute-user-pending-review.ts`, `src/server/services/orchestrator/promptAuditing.ts`) are unchanged and byte-identical in behaviour.

The three hardcoded sites are now parameterised: the dedupe query, the create, and the notification.

🔴 **Dedupe is now scoped per type.** Scoped to the user alone, the first queue to mute an account would permanently silence every other queue for it: a later finding of a different kind returns `deduped: true` against a row about something else and files nothing a moderator can ever see. Both directions are tested.

Preserved as-is, all still covered: `mutedAt` is never written here; the skip branches read from `dbWrite`; the mute and the create share one `$transaction`; the repair branch re-mutes an unmuted user; the module's import graph stays light (the only additions are plain string constants).

### Moderator app

`getGenerationRestrictions` takes an optional `type`, defaulting to the existing constant. The queue page gains a `type` filter following the existing `parseQuery` + `.catch()` pattern, plus a Type select mirroring the Status one.

The type vocabulary moved to `apps/moderator/src/lib/restriction-types.ts` — **outside `$lib/server/`** because the filter component needs it as values, and SvelteKit rejects a value import of `$lib/server/*` from client-reachable code. The service re-exports it, so server callers keep one import site.

## The notification decision

**Chosen: an opt-in per-type map, `PENDING_REVIEW_MUTE_NOTIFICATION`, where `null` means send nothing. `generation` keeps `generation-muted` exactly as before; `bot-account` sends nothing.**

I checked how types are registered before deciding, and the evidence pointed away from inventing one:

1. **`createNotification` validates `type` against nothing.** It is `z.string()` at the schema and `text` at both tables, and the fan-out worker inserts it verbatim. An unregistered type is **persisted and increments the user's unread badge**, while the bell dropdown drops it at render (no processor can build a message for it). The result is a phantom unread count with no click target, clearable only by "mark all read". So sending an unregistered type is *worse* than sending none.
2. **Reusing `generation-muted` would be a lie.** Its registered copy reads *"Your generation access has been restricted due to potential Terms of Service violations."* — wrong, and user-visible, for a bot-account mute.
3. Registering a real `bot-account-muted` processor is cheap mechanically, but its user-facing copy is a product and comms decision that belongs with the detector, not with a plumbing change.

A new type therefore stays silent until someone deliberately (a) adds a processor under `src/server/notifications/` reachable from `notificationProcessors`, and (b) names it in the map. A seam test asserts every non-null value in the map **is** a registered processor key, so a mapping added without the processor fails rather than shipping a ghost notification.

## The resolve/uphold path — reported, not changed

Asked to check `user-restriction-resolve.service.ts`. It does **not** mis-resolve by id — `resolveUserRestriction` looks a row up by primary key and rules on the correct one. But its side effects are generation-shaped: hardcoded `generation-restriction-upheld` / `-overturned` notification types, a `moderator:generationRestriction*` update source, a generation-worded email, and on an overturn it resets the **prompt** violation counter.

Rather than silently widening scope, the moderator `resolve` and `ban` actions now **refuse** a row whose type has no verdict path, with an explicit message. The check runs *before* the ban in the ban action, since that action bans and then rules — refusing only at the ruling would leave an account banned against an unresolvable restriction. The page says so in its header for non-generation types.

Two further findings, reported not fixed:

- `overturnPendingReviewMute` selects `type: 'generation'`. That is fail-safe today (it cannot touch a bot-account row), so I left it. But once a detector exists, a user holding **both** an open generation and an open bot-account case would be unmuted by an overturn while the bot-account row stays Pending — precisely the "Pending row on an unmuted account" state the repair branch exists to prevent. The detector work needs to handle this.
- `restrictionById` must look up across all types (`type: 'any'`), because a form posts to `?/resolve`, which **replaces** the query string — an action never learns which queue the moderator was in. Filtering it by the default type would 404 every non-generation row.

## Round 3 — the round-2 delta audit's findings, closed

Two commits on top of `ff97751d20`, addressing [the round-2 comment](https://github.com/civitai/civitai/pull/4609#issuecomment-5533502582). Nothing here is live-breaking: the only writer of `UserRestriction` rows is `applyPendingReviewMute`, neither production caller passes a `type`, so every row in production today is `generation`.

**F1 — the seam guard passed green over a real divergence.** It read the moderator app's vocabulary as TEXT: the regex captured to the FIRST `]` after the `=`, so a comment naming an index truncated it, and it extracted single-quoted strings only, so a differently-quoted entry vanished. Reproduced: with the moderator list written as

```ts
export const RULINGS_WIRED_FOR: readonly RestrictionType[] = [
  'generation', // matches RESTRICTION_TYPES[0]
  'bot-account',
];
```

the seam file reported **8 passed / 0 failed** while the two lists genuinely disagreed. That is the ban-then-strand hazard the copy exists to prevent, reached with its pinning guard green.

Not fixed by widening the regex — a guard that pins source text by PATTERN is walkable by rewriting the text, and the rewrites that walk it are ordinary (a Prettier reflow, a comment, a quote style). **The reader now imports and executes the module and compares values.** That is available because `apps/moderator/src/lib/restriction-types.ts` has no imports of its own; a comment there now says so and says to keep it that way, and if someone gives it a `$lib/…` import the guard fails loudly naming the module rather than relaxing.

Measured, old reader vs new, each against the same mutated moderator module:

| Moderator list written as | Old (text parse) | New (import + execute) |
|---|---|---|
| a comment containing `]` | **8/8 GREEN** | RED — 2 failures |
| a mixed quote style (`'generation', "bot-account"`) | **8/8 GREEN** | RED — 2 failures |
| a list assembled with a spread | **8/8 GREEN** | RED — 2 failures |
| all entries double-quoted | RED (its empty-list control) | RED |
| multi-line, no comment | RED | RED |
| single-line trailing comma | RED | RED |

`moderatorRestrictionTypes()` shared the same blind spot and is fixed by the same change; the refusal SENTENCE is now *called* rather than parsed out of a template literal, so a message assembled from constants is compared on what it produces.

Also: `apps/moderator` `restriction-types.test.ts` now pins `RULINGS_WIRED_FOR` **by value**. `unwired.length > 0` caught a widened list only by accident — `bot-account` being the only unwired type. Measured: with a third filed type present, the length check **passes** over a wrongly-wired `bot-account` and the value pin is the only thing that fails.

**F2 — the old test matrix.** Rewritten below; the "10 red at base" table was not reproducible and five of its rows were never watched red individually.

**F3 — a comment asserted a boundary the tree does not have.** The runtime type guard's comment claimed the values reaching it "cross an HTTP boundary and a JSON body". Nothing does — neither production caller passes a `type`, and `mute-user-pending-review.ts`'s zod schema has no `type` key. The guard stays; the comment now says what it is actually for (the shape of the next caller, and the callers TypeScript cannot vouch for: an `as` cast, a value read back off the free-text column, a JS caller). The same false sentence in the test file's docblock is corrected too.

**F4 — the refusal reached REST as an opaque 500.** `resolveUserRestriction` threw a plain `Error`, so `handleEndpointError` fell to its non-TRPCError branch and the moderator's panel rendered *"Restriction ruling: An unexpected error occurred."* Now `throwBadRequestError`, which keeps status and message.

> **Scope decision:** the two neighbouring guards in the same function had the identical defect, and I fixed them in the same change rather than converting one of three. `Restriction record not found` is now a 404 and `Restriction has already been resolved` a 400 — both are facts about the request, neither is a server fault, and leaving two as opaque 500s would have recreated the same predicate spelled two ways one line apart. All three are covered.

**F5 — an unstated NOT-NULL precondition.** `ORDER BY (ur.status = 'Pending') DESC` is correct only while `status` is NOT NULL; Postgres defaults `DESC` to `NULLS FIRST`. Now `DESC NULLS LAST`, which makes the ordering independent of the column's nullability. The column is a NOT NULL enum today, so this is a precondition made explicit, not a live defect.

**Also:** the five new fixture modules are added to `.prettierignore` — their formatting *is* the fixture, and Prettier would normalise four distinct cases into one.

## Round 4 — the round-3 delta audit's findings, closed

Two commits on top of `48ff122bf1`, addressing [the round-3 comment](https://github.com/civitai/civitai/pull/4609#issuecomment-5533907315). Scope bound unchanged: nothing in production files a non-`generation` restriction, so every path below is exercised only by tests.

**F-1 — the execute-based reader was blind to an environment-conditional vocabulary, and the reader it replaced was not.** The harness executes the moderator module *in the main app's test process*, so an environment-dependent value is resolved under Vitest rather than at moderator-build time. Reproduced with the list written as

```ts
export const RULINGS_WIRED_FOR: readonly RestrictionType[] = import.meta.env.DEV
  ? ['generation']
  : ['generation', 'bot-account'];
```

— seam + vocabulary suites **28 passed / 0 failed**, moderator's own value pin **5 passed / 0 failed**, production build shipping both types. Ban-then-strand with every guard green. The base commit's text parser goes RED on that same module, so the round-3 replacement traded one blind spot for another rather than being strictly stronger.

Closed by keeping a **text** assertion alongside the execute check — the module's source may contain no `import.meta` and no `process.env`. The two are complementary and both are kept. With it, that mutant fails the seam suite at its `beforeAll` and the vocabulary suite's real-module case, each naming the constant.

🔴 **Comments are stripped before the scan**, and that is load-bearing rather than tidy: the module has to be able to NAME the shapes it refuses, and a raw-text scan fires on the sentence forbidding the thing instead of on the thing — the guard is then matched by its own documentation and the only route back to green is to stop documenting the rule. Pinned by a control asserting a commented mention does not trip it, alongside the two positive cases and a plain-literal negative control.

The moderator module's stated precondition was *"keep this module import-free"*. `import.meta.env` and `process.env` need no import statement, so that sentence never covered this. It now states both constraints and says explicitly that the second is not implied by the first.

**F-2 — the new 4xx bodies dropped the key the only consumer reads.** `handleEndpointError`'s 4xx pass-through emits `{ message }` and no `error` key, while every other refusal from `defineModeratorEndpoint` emits `{ error, message, code }`; the moderator's `readError` read `body.error` and nothing else. So all three refusals from round 3 came back `null` and a moderator ruling on an already-ruled row saw **"Restriction ruling returned 400."** — the same loss as the opaque 500 that round removed, one layer further out.

**Fixed at the consumer, and the choice is measured rather than assumed.** `handleEndpointError` is the shared chokepoint for **36** REST route files; its 4xx and 503 pass-through bodies are pinned `toStrictEqual({ message })` by `endpoint-helpers-error-envelope.test.ts` (the 503 case as an explicit, documented carve-out); and `restErrorBody` needs a `RestErrorCode` that is not derivable at that point without a new status→code map, which would then have to be reconciled with the closed key ledger `rest-error-envelope-ledger.test.ts` enforces. Widening the reader costs one expression and makes every endpoint's 4xx legible to this app. **Widening the emitter changes the wire format of 36 routes and belongs in its own PR** — it is not done here.

The body-shape rule moves to `apps/moderator/src/lib/server/rest-error-reason.ts`, kept import-free so the main app's suite can load it by filesystem path — the same mechanism as the vocabulary harness. That is what lets the new test drive the **real emitter into the real reader in one process**, over all three refusals. Two suites each mocking the other side is exactly the arrangement that cannot see a disagreement about a field name, which is why round 3's tests were green: they drove the real helper but asserted `body.message`, which the consumer never looked at.

**F-3 — M8's kill count, and shape-drifted fixtures.** All five fixtures wrote `return RULINGS_WIRED_FOR.includes(type)` while the real module writes `return (RULINGS_WIRED_FOR as readonly string[]).includes(type)`. The old parser's message regex requires `return (RULINGS_WIRED_FOR`, so it could not read the message out of *any* fixture, and M8 was measuring the old reader against a shape the module does not have. All five now carry the cast. **Re-measured: M8 is 6 of 21** — the three fixture pairs the table names, on both their list and message cases. The audit's 8 was the drift.

`computed.ts`'s comment now distinguishes *assembled from constants declared in this file* — the same value in every environment, which is what the execute reader can certify — from *assembled from the environment*, which is refused. It was the fixture that made the hazardous shape look sanctioned.

**F-4 — an undocumented cross-app coupling.** Resolving the moderator path makes Vite load `apps/moderator/tsconfig.json`, which extends the gitignored generated `.svelte-kit/tsconfig.json`; without `svelte-kit sync` both suites fail with `TSConfckParseError: failed to resolve "extends"`, an error naming a tsconfig rather than the seam, on the **main** app's suite. CI is unaffected (`prepare` runs sync); a fresh clone or worktree is not. Named in the harness docblock.

### Round-4 delta — red at `48ff122bf1`, green at HEAD

| Test | At `48ff122bf1` | What that failure IS |
|---|---|---|
| hands the moderator app a reason it can read, not just a status | RED — `an unwired type: expected null not to be null` | behavioural — the real emitter's body driven through the real consumer rule |
| refuses a vocabulary whose values are read from the environment | RED — the symbol does not exist at `48ff122bf1` | **an import error, not behavioural coverage.** It is attributed by mutation instead (below) |

The environment guard is attributed by **M9**, not by that red.

| # | Mutation | Killed | Failing assertion |
|---|---|---|---|
| M9 | the real moderator module's `RULINGS_WIRED_FOR` rewritten as `import.meta.env.DEV ? ['generation'] : ['generation', 'bot-account']` | 1 test + 1 suite | `restriction-type-seam.test.ts` fails at `beforeAll` and `moderator-restriction-vocabulary.test.ts > loads the real moderator module when given no path` fails, both with ``apps/moderator/src/lib/restriction-types.ts reads `import.meta`.`` **Control: the same mutant against the pre-fix harness is 28 passed / 0 failed** |

## Test matrix — what was actually observed

🔴 **The earlier version of this section claimed "10 red at `origin/main`" for the main app and "15 red" for the moderator app. Neither is reproducible, and the reason matters: at the PR's base commit `74306cc04c` most of these tests cannot RUN, because the symbols they import do not exist there.** A test whose base failure is an import error or a missing export is a reassuring zero, not regression coverage. Re-measured on `48ff122bf1` by reverting the PR's **source** files (not its tests) to `74306cc04c`:

| Suite | At `74306cc04c` | What that failure IS |
|---|---|---|
| `src/server/__tests__/pending-review-mute.test.ts` | **fails to collect — 0 tests** | `it.each(USER_RESTRICTION_TYPES)` with the export absent → `TypeError: Cannot read properties of undefined (reading 'every')` |
| `src/server/services/__tests__/restriction-type-seam.test.ts` | **fails to collect — 0 tests** | `apps/moderator/src/lib/restriction-types.ts` does not exist at base |
| `src/server/services/__tests__/moderator-restriction-vocabulary.test.ts` | 7 failed / 13 passed / 8 skipped | every failure a missing-symbol `TypeError`, not a behaviour difference |
| `apps/moderator/…/__tests__/restriction-types.test.ts` | **fails to collect** | `$lib/restriction-types` does not exist at base |
| `apps/moderator/…/generator-restrictions/__tests__/type-queue.test.ts` | **fails to collect** | same |
| `apps/moderator/…/__tests__/user-lookup-restriction-row.test.ts` | 5 failed | `TypeError: getIdentity is not a function` — the export is new in this PR |
| `apps/moderator/…/__tests__/restriction-type-filter.test.ts` | **5 failed** | **genuine behavioural reds** — see below |

Whole-PR revert totals: main app 3 files failing (2 of them uncollectable); moderator app **10 failed / 435 passed / 36 skipped across 43 files**, with 2 further files uncollectable. Reverting only the main-app source leaves `pending-review-mute.test.ts` still uncollectable and the seam file at 6 failed / 2 passed — every one of those six a `TypeError: … is not iterable`.

**So exactly five rows in this PR are behavioural reds at its own base**, all in `restriction-type-filter.test.ts`:

| Test | Failing assertion at `74306cc04c` |
|---|---|
| filters on the type it was given | `expected 'generation' to be 'bot-account'` |
| binds bot-account and no other type, in both the list and the count | `expected 'generation' to be 'bot-account'` |
| keeps the type predicate when other filters are also applied | `expected 'generation' to be 'bot-account'` |
| drops the predicate only for the explicit `'any'` | `expected 'generation' to be undefined` |
| selects the row type, so a caller can tell what it got back | the compiled SQL does not contain `"ur"."type"` in the select list |

Everything else is attributed **by targeted mutation** instead — see below.

### Round-3 delta — red at `ff97751d20`, green at HEAD

These are genuine behavioural reds, measured against the previous PR head rather than against the PR base, and none is an import error:

| Test | At `ff97751d20` |
|---|---|
| reaches REST as a 400 naming the type, not an opaque 500 | RED — `expected 500 to be 400` |
| reports a missing row as a 404 rather than a server fault | RED — `expected 500 to be 404` |
| reports an already-ruled row as a 400 rather than a server fault | RED — `expected 500 to be 400` |
| prefers a Pending row over a newer one, in every restriction column | RED — compiled SQL lacks `DESC NULLS LAST` |
| places nulls last, so the ordering does not depend on status being NOT NULL | RED — same |
| claims a verdict path for generation and for nothing else *(moderator)* | new pin; attributed by mutation — see M6/M7 |

The 20 tests in `moderator-restriction-vocabulary.test.ts` cannot be red at `ff97751d20` — the reader they test did not exist there — so they are attributed by mutation M8.

### Invariant guards — NOT counted as regression coverage

Green on both sides, labelled as such. They pin existing behaviour rather than prove the change:

- files a generation restriction when no type is given *(backward compatibility — the point is that it does not move)*
- still dedupes within a type; mutes + refreshes session for a non-generation type; one transaction; `mutedAt` never written (both types); all five skip branches (protected / moderator / deleted / banned) for a bot-account mute; notifies on the generation path
- moderator: filters on generation when none given; binds generation; still resolves a generation restriction; still bans off a generation restriction
- `overturnPendingReviewMute cannot reach a non-generation row at all`
- the `multi-line array` and `trailing comma` vocabulary fixtures — the old text parser reads both correctly, **lists and messages alike**, so they are **declared coverage of the new reader, not evidence it is better**; the comment / double-quoted / computed fixtures are the ones the old reader misses. 🔴 Round-3 note, corrected in round 4: as written then this bullet was true of their LISTS and false of their MESSAGES, because all five fixtures omitted the real module's `as readonly string[]` cast and so defeated the old parser's message regex outright. The fixtures now carry the cast and the sentence is true again as stated.

The skip-branch and `mutedAt` cases are inherently invariant guards — there is no pre-change behaviour for them to differ from.

## Mutation results

Each mutation applied alone and reverted; every mutant died on **its own** test's assertion, not incidentally. M1–M3 were **re-run this round** rather than restated; M4–M5 are from the earlier round and were independently confirmed by the round-2 audit; M6–M8 are new.

| # | Mutation | Killed | Failing assertion |
|---|---|---|---|
| M1 | dedupe query back to hardcoded `'generation'` | 3 | `expected { muted: true, …(2) } to match object { muted: true, deduped: false }`; `expected [ …(2) ] to have a length of 1 but got 2`; repair-branch deep-equal |
| M2 | create back to hardcoded `'generation'` | 5 | `expected [ 'generation', 'generation' ] to deeply equal [ 'generation', 'bot-account' ]` + 4 more |
| M3 | notification `?? 'generation-muted'` instead of suppressing | 1 | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |
| M4 | type predicate bound to the constant instead of the argument | 3 | `expected 'generation' to be 'bot-account'` |
| M5 | ruling refusal disabled | 2 | `expected undefined to be 400` |
| M6 | moderator `RULINGS_WIRED_FOR` widened to include `bot-account` | 2 | `expected [ 'generation', 'bot-account' ] to deeply equal [ 'generation' ]`; `expected 0 to be greater than 0` |
| M7 | M6 **plus a third filed type**, which is the case the old shape could not see | 1 | only the new value pin fails — `expected [ 'generation', 'bot-account' ] to deeply equal [ 'generation' ]`; the `unwired.length > 0` check **passes** |
| M8 | the vocabulary reader reverted to the base commit's text parser | 6 of 21 (**re-measured in round 4**; the round-3 figure was quoted over a 20-test file and against shape-drifted fixtures — see F-3) | the `comment containing ]`, `double-quoted` and `computed` fixtures, on both their list and message cases (`Could not find a RESTRICTION_TYPES array literal…`, `expected [] to deeply equal [ 'generation', 'bot-account', …(1) ]`, `expected [ 'generation', 'spam-account' ] to deeply equal …`) |

M1 is the required dedupe check. Note the honest asymmetry: **"does not let an open bot-account case swallow a generation mute" survives M1** — hardcoding the dedupe to `generation` still creates a fresh row when only a bot-account row exists. That direction is killed by M2.

**Two tests initially passed against base code for the wrong reason, and were fixed in the second commit** (found precisely by running the base comparison):
- the ban refusal asserted only `400`, and its fixture used a `reasonCode` outside `BAN_REASONS` — so the schema rejected the payload before the type check ran. Now uses a valid payload, asserts the message, and adds a generation-row positive control proving the refusal is not rejecting every ban.
- the SELECT-list assertion matched the whole statement, so the `"ur"."type" = $1` in the WHERE satisfied it. Now sliced to the select list.

The SQL tests assert **bound parameters**, not just SQL text: `where('ur.type','=',x)` compiles to `"ur"."type" = $1` for every `x`, so a mutant ignoring its argument emits byte-identical SQL. `capturingDb` gained an optional params array for this.

## Verification performed

- `pnpm typecheck` — **0 type errors**. (LSP reported `Cannot find module '~/...'` errors in the test file; those were phantom, from IDE resolution against a different checkout, and the repo's own typecheck disproves them.)
- moderator `svelte-check` — **0 errors, 0 warnings** (needs `svelte-kit sync` first; without it `$types` is stale).
- `vitest --project unit` on the three main-app files — **99 passed** (97 before round 4).
- `vitest --project app:moderator` — **463 passed across 43 files** (36 skipped).
- `pnpm test:lint-rules` (the 30 repo-structural guards) — 459 passed.
- eslint on all changed files — clean, no warnings. Prettier clean on every changed file (the five vocabulary fixtures are deliberately `.prettierignore`d).

> History of these figures: `06991e8417` gave 58 main-app / 452 moderator across 40 files; the round-1 follow-ups (`d35adbf59a`, `ff97751d20`) took that to 74 / 461 across 43; the round-3 commits (`80d41afe2b`, `48ff122bf1`) took that to 97 / 463 across 43; the round-4 commits (`62c11bd3b1`, `be95cc21a9`) take it to **99 / 463 across 43**. The pre-existing `reports.queries.explain.test.ts` `ECONNREFUSED` failures are unchanged at 36 skipped.

## NOT verified / out of scope

- **Nothing files a non-generation restriction yet.** The whole `bot-account` path is exercised only by tests; no production code passes a `type`.
- **The 4xx envelope itself is NOT changed, deliberately (round 4, F-2).** `handleEndpointError` still emits `{ message }` with no `error` key on its 4xx/503 pass-through. Every other endpoint's 4xx stays equally unreadable to any consumer that reads `error` only; only the moderator app's reader was widened. Making the emitter conformant means a status→code map reconciled with the `rest-error-envelope-ledger.test.ts` key ledger and re-pinning two `toStrictEqual` guards across 36 route files — a separate PR.
- **Retool is out of repo and was not inspected** (round 4, F-2). Whether it branches on the old 500, or reads `error` vs `message`, is unverified; the fix is measured only against the in-repo consumer.
- **`refuses a vocabulary whose values are read from the environment` has no behavioural red at `48ff122bf1`** — the symbol it imports does not exist there, so its base failure is an import error. It is attributed by mutation M9 instead, with the pre-fix control recorded.
- **No live/manual run.** Not exercised against a real database or a running moderator app — no DB was reachable here.
- **The Svelte UI is untested.** This app has no component-test harness (its Vitest project includes only `*.test.ts` and does not alias `$app/*`). The type select and the type-aware copy are typechecked and reviewed, not rendered.
- **`reports.queries.explain.test.ts` fails in this environment** (36 tests, `ECONNREFUSED` to local Postgres). Confirmed **identical at base source**, so pre-existing and environmental, not a regression here.
- **Prettier cannot parse `.svelte`** in this setup — every untouched sibling component errors the same way, so it is repo-wide, not introduced.
- **The verdict path is not parameterised**, deliberately. Rulings on non-generation rows are refused rather than made correct.
- **`overturnPendingReviewMute`'s double-open-case hazard is not fixed** — described above; it is unreachable until a detector exists.
- **The `DESC NULLS LAST` change was never EXECUTED** — there is no database in this tier, so it is asserted on the compiled SQL text only, like the ordering it replaces.
- **The moderator vocabulary import across the app boundary is not exercised by CI in any other way.** It works because that module has no imports; if it gains one, the seam guard fails loudly (module-not-found naming the file) rather than silently — that is the intended failure, but it has only been reasoned about, not staged.
- I removed a pre-existing unused `dbWrite` import in `+page.server.ts` (a lint **warning**, not an error, already present on `main`) as a drive-by since I was in the file.


ZacxDev:

Head `06991e8417`, merge base `74306cc04c`. **No 🔴. 2 🟡, 4 🟢.**

🔴 **Correction to this PR's description.** The body says *"the moderator `resolve` and `ban` actions now refuse a row whose type has no verdict path"*. That is true of **one of two** ruling surfaces. The retool User Lookup panel is a second, unguarded one, and it is not on the body's "reported, not fixed" list — so as written the summary reads as covering ground it does not. Correcting here rather than editing the body, since a reviewer may have read the earlier version.

### 🟡 1 — the type refusal guards one of two ruling surfaces (`payload`)

`unwiredRuling` (`apps/moderator/src/routes/audit/generator-restrictions/+page.server.ts:92-98`) covers the audit queue. The retool panel's `resolveRestriction` action (`apps/moderator/src/routes/retool/user-lookup/[section]/+page.server.ts:126-146`) is unguarded, and the query behind it (`user-lookup.service.ts:421-432`) has **no type predicate**.

Scenario: a bot-account row is filed for user 42; the panel renders Overturn/Uphold; clicking through reaches `resolveUserRestriction`, which sends a `generation-restriction-overturned` notification ("Your generation access has been restored"), a `restriction-overturned` email, and calls `resetProhibitedRequestCount(42)` — **wiping that user's real prompt-violation counter over a bot-account case**. Exactly the harm `unwiredRuling` exists to prevent.

3 of 5 entry points into `resolveUserRestriction` are unguarded (also `user-restriction.router.ts:87`, `api/mod/restriction/resolve.ts:22`). The durable fix is to move the check **down to `resolveUserRestriction`**, closing all five at once, rather than replicating the guard at a third route.

### 🟡 2 — per-type dedupe makes "newest restriction of any type" the wrong row (`payload`, second-order)

`user-lookup.service.ts:421-432` uses `ORDER BY ur.id DESC LIMIT 1` with no type filter. That was sound while a user could hold at most one open row. This PR's central decision — per-type dedupe, so two open cases can coexist — makes it unsound: a Pending generation case behind a later Upheld bot-account row renders as **no open restriction**, leaving the user muted with no ruling form while nobody can see the open case.

### 🟢 3 — `flagSuspicious` silently widened to every type while its comment claims otherwise (`payload`)
`+page.server.ts:181-207` vs the comment at `:65-69`. Impact is a shared review list, not a user-facing write — but a guard comment wider than its implementation is what stops the next reader looking.

### 🟢 4 — `type` is enforced only by TypeScript (`payload`)
No runtime check at `user-restriction.service.ts:88-93`. An out-of-vocabulary value mutes the account, files a row no queue can select (`z.enum(...).catch(...)`), and sends no notification — a silently muted user with no reviewable case. One `includes` check makes it structural. This is the seam whose whole purpose is accepting a caller-supplied type.

### 🟢 5 — `capture-sql.ts:34-37` declares the capture array `readonly`, then casts it away to push (`scaffolding`)
### 🟢 6 — the Bot account queue ships with live Uphold/Remove/Ban buttons that always 400 (`payload`)
Server-side refusal is correct and should stay. It is not an alternative to `disabled` on the form — the warning is in the page header, the buttons are in the detail pane.

### Self-reported claims, re-checked

Base-revert reproduced **15 red at base** on the moderator side exactly; mutants M1 and M4 reproduced exactly, including the honest asymmetry the body already declares. `mutedAt`-never-written is **really pinned, not spelled** — the assertion compares the whole `dbWrite.user.update` argument array for both types, so a mutant adding `mutedAt` cannot pass. Two numeric discrepancies, both under-reports: 10 tests red at base on the main app (body says 9), and 452 moderator tests (body says 451, a pre-second-commit figure). M2/M3/M5 not independently re-run — flagged as unverified rather than covered.

**Not verified:** no live run, no database, no rendered Svelte (no component-test harness in that app), and the suites ran on the PR head rather than a tree merged with the current `origin/main` tip — drift on the touched paths is zero, which is weaker than a merged-tree run.

All findings are latent until a detector ships — the same reachability class as the `overturnPendingReviewMute` hazard the body already reports. Fix round to follow.

ZacxDev:

Two commits, `06991e8417..ff97751d20`. Payload lines changed this round: **216** (plus 508 scaffolding).

```audit-claims round=1 audited=06991e8417..ff97751d
1. The type refusal was moved into `resolveUserRestriction` itself, claimed to cover all five entry points into the verdict path.
2. The audit-queue route guard was kept rather than removed, and claimed to no longer spell its own message — it calls the shared `unwiredRulingReason`.
3. The main app's wired-for list and refusal message are claimed to be pinned against the moderator app's copy by a seam test that fails in both directions.
4. The user-lookup panel's three correlated subqueries were changed to order Pending first then newest, via one shared helper, claimed to stop an open case being hidden by a newer resolved row of another type.
5. Runtime validation of `type` was added to `applyPendingReviewMute`, claimed to reject an out-of-vocabulary value before any write.
6. The `flagSuspicious` comment and implementation were reconciled.
7. `capture-sql.ts` no longer declares its capture array `readonly`.
8. The ruling forms are disabled for types with no verdict path, with the server-side refusal claimed unchanged.
9. The PR body's two numeric errors were corrected: 9 → 10 red at base (main app), 451 → 452 moderator tests.
```

Reported as **not verified** by the fix round: no live run and no database — finding 4's ordering fix is asserted against compiled SQL text, not executed against Postgres; finding 8's rendering half is verified by reading, not rendering (no component-test harness in `apps/moderator`); the moderator-side red-at-base attribution for the new suites failed for the wrong reason (missing exports at base) and was attributed by targeted mutation instead; suites ran on the branch, not on a merged tree.

Also flagged for a decision: the new refusal throws a plain `Error`, matching its two neighbouring guards, so `handleEndpointError` genericizes it to an opaque **500** rather than a 400 for a caller reaching the REST endpoint directly.

ZacxDev:

Range `06991e8417..ff97751d20`, base `74306cc04c`. **No 🔴. 2 🟡, 3 🟢.**

🔴 **Correcting my own earlier comment.** [My round-1 comment](https://github.com/civitai/civitai/pull/4609#issuecomment-5533129193) said 10 tests were measured red at base on the main app, and the PR body was updated 9 → 10 on that basis. **That figure is not reproducible and the correction did not fix what is wrong with the matrix** — see F2. Five of its rows were never watched red individually.

### 🟡 F1 — the seam test passes green over a real divergence of the two wired-for lists · `guard`/`scaffolding`

`src/server/services/__tests__/restriction-type-seam.test.ts:44-53`. The regex captures `[^\]]*` to the **first** `]` after the `=`, then reads only **single-quoted** strings from that slice. Measured in a detached copy: writing the moderator list as

```ts
export const RULINGS_WIRED_FOR: readonly RestrictionType[] = [
  'generation', // matches RESTRICTION_TYPES[0]
  'bot-account',
];
```

leaves the seam file at **8 passed / 0 failed** while the lists genuinely differ. Same green for `["bot-account"]` in double quotes. The `length > 0` control cannot see it — the first entry survives the truncation.

Failure scenario: the moderator app believes `bot-account` is rulable → `unwiredRuling` returns `null` → the `ban` action runs `setBanned` and posts the ruling **after**, which the main app refuses → account banned, Pending row stranded. That is the precise outcome the moderator-side copy exists to prevent, reached with its pinning guard green.

Currently masked: `restriction-types.test.ts:32` catches this mutant only because `bot-account` is the *only* unwired type, so `unwired.length > 0` fails. Add a third unwired type and the incidental catch disappears while the parser stays blind.

### 🟡 F2 — "10 red at base" is not reproducible; 5 rows cannot be observed at all · `guard`/`scaffolding`

With payload reverted to base, `pending-review-mute.test.ts` **fails to collect** — `USER_RESTRICTION_TYPES`/`RULINGS_WIRED_FOR` are not exported at base, so `it.each(undefined)` throws and the file contributes **0 tests**. Whole-PR revert: 8 red + one uncollectable file. Main-app-only revert: 6 red / 2 passed.

So the five rows attributed to that file were never watched red individually — an import failure, which is a reassuring zero rather than counted coverage. The seam half is **8**, not the 5 the table lists (this round added three ruling-scope tests without updating it).

What *is* verified: every mutation applied to the new guards died on its own test's assertion, and the footnote figures reproduce exactly (74 main-app, 461 moderator / 43 files).

### 🟢 F3 — the runtime validation's stated justification is false · `guard`/`payload`
`user-restriction.service.ts:127-140`. The comment claims the types reaching it "cross an HTTP boundary and a JSON body". Neither production caller passes one, and `mute-user-pending-review.ts:19-24`'s zod schema has **no `type` key**, so no HTTP body can supply one. The guard is correct and well tested; the comment asserts a boundary the tree does not have.

### 🟢 F4 — the new refusal reaches REST as an opaque 500 · `behaviour`/`payload`
`user-restriction-resolve.service.ts:58` throws a plain `Error` → `handleEndpointError`'s non-TRPCError branch → 500 `"An unexpected error occurred"`. The moderator's `postJson` renders **"Restriction ruling: An unexpected error occurred."** — the reason is destroyed. Reachable from the retool panel's `?/resolveRestriction`, which has no local guard. `throwBadRequestError` preserves both.

### 🟢 F5 — `ORDER BY (status = 'Pending') DESC` is correct only while `status` is NOT NULL · `guard`/`payload`
`user-lookup.service.ts:400-407`. Postgres `DESC` defaults to `NULLS FIRST`. Today the column is a NOT-NULL enum, so this is an unstated precondition, not a live defect. `DESC NULLS LAST` makes it independent of the column.

### Prior claims — verdicts
Fixed: guard moved into `resolveUserRestriction` and covering every entry point; route guard delegating to the shared predicate; lookup ordering via one helper with a total order; runtime validation; `flagSuspicious` comment; `capture-sql` readonly. **Partially:** the both-directions pin (real, but blind per F1); forms disabled (the retool panel's new server-side refusal is the 500 of F4); PR body numbers (452→461 and 58→74 reproduce; "10 red at base" does not).

Residual, acknowledged in the code: two coexisting **Pending** rows still resolve to the newest, so a Pending `bot-account` filed after a Pending `generation` still hides the rulable one from that panel.

**Scope bound on the whole round:** the only writer is `applyPendingReviewMute`, and neither production caller passes a type, so every `UserRestriction` row in production is `generation` and `unwiredRulingReason` returns `null` for all of them. Nothing in this range changes behaviour for any row that exists today.

**Could not verify:** no database in this tier, so the new `ORDER BY` was never executed — only its compiled text read. `reports.queries.explain.test.ts` **skipped** its 36 tests here rather than failing `ECONNREFUSED` (no `DATABASE_URL` in the shell), so "identical at base" is unconfirmed by this round.

ZacxDev:
## Round 2 fixes — what was claimed

Commits `80d41afe2b`, `48ff122bf1`. Payload lines changed this round: **60** (plus ~500 scaffolding).

```audit-claims round=2 audited=ff97751d20..48ff122b
1. The seam guard no longer parses the moderator module's source text — it imports and executes it, claimed to compare values rather than text.
2. The refusal-message pin is claimed to call the moderator's message function rather than regex it out of a template literal.
3. `moderatorRestrictionTypes()` is claimed to use the same execute-based mechanism, closing the shared blind spot.
4. `restriction-types.test.ts` is claimed to be robust to a third unwired type existing, rather than catching divergence only incidentally.
5. All three guards in `resolveUserRestriction` — not only the new refusal — are claimed to reach REST with their own status and message: 400 for an unrulable type, 404 for a missing row, 400 for an already-ruled row.
6. The lookup ordering is claimed to be independent of the column's nullability via `DESC NULLS LAST`.
7. The comment on the runtime `type` validation is claimed to no longer assert an HTTP boundary the tree does not have.
8. The PR body's red-at-base tables are claimed to be replaced by what was actually observed, with per-test base-failure reasons, and the remainder attributed by mutation instead of counted.
```

Two of round 2's own claims were **corrected by measurement** in this fix round, and the corrections are against the audit, not for it: an **all**-double-quoted moderator list *was* already caught by the old guard (its empty-read control fired); the case that stayed green is **mixed** quoting. A plain multi-line array and a trailing comma were also already caught. Those two fixtures are labelled declared coverage, not evidence the new reader is stronger.

Reported as **not verified**: the `DESC NULLS LAST` SQL was never executed — asserted on compiled text only, the same instrument as the ordering it replaces. The cross-app import works today only because the moderator module has no imports of its own; that property is guarded by a loud failure and a comment, not by a second CI witness. The full ~22k-test `unit` project was not run — only the three affected files plus the 30 structural lint-rule guards.

ZacxDev:
## Delta re-audit — round 3 (blind to reasoning, read-only)

Range `ff97751d20..48ff122bf1`, base `74306cc04c`. **No 🔴. 2 🟡, 2 🟢.**

### 🟡 F-1 — the replacement reader is blind to an environment-conditional vocabulary, and the reader it replaced was not · `guard`/`scaffolding`

`moderator-restriction-vocabulary.harness.ts:99-109` executes the moderator module **in the main app's test process**, so an environment-dependent value is read at test time, not at moderator-build time. The documented precondition is "keep this module import-free" — but `import.meta.env` and `process.env` need no import, and the `computed.ts` fixture explicitly certifies "values assembled at runtime" as supported, which makes the shape look sanctioned.

Measured. With the moderator list written as

```ts
export const RULINGS_WIRED_FOR: readonly RestrictionType[] = import.meta.env.DEV
  ? ['generation']
  : ['generation', 'bot-account'];
```

the seam and vocabulary suites report **28 passed / 0 failed**, and the moderator's own value pin reports **5 passed / 0 failed**, while the production build ships `['generation', 'bot-account']`. Ban-then-strand, reached with every pinning guard on both sides green.

**The control that makes this a finding:** reinstating the *base commit's text parser* against the same mutated module goes **RED**, naming the constant. So the replacement is **not strictly stronger** — it trades a formatting blind spot for a runtime-environment one, and nothing in the diff records the trade. Cheapest closure is to keep one *text* assertion (no `import.meta`/`process.env` in that module) alongside the execute check; they are complementary, not redundant.

### 🟡 F-2 — the three new 4xx bodies drop the key the only in-repo consumer reads · `behaviour`/`payload`

`user-restriction-resolve.service.ts:63,66,69`. `handleEndpointError`'s 4xx pass-through (`endpoint-helpers.ts:658-665`) emits `{ message }` only, while every other refusal from `defineModeratorEndpoint` emits `{ error: … }` — and the moderator's `readError` (`user-actions.service.ts:61-67`) reads `body.error` and nothing else. Measured against the real helper:

```
bare Error           → 500 {"error":…,"message":…}  readError → "An unexpected error occurred"
throwBadRequestError → 400 {"message":…}            readError → null
throwNotFoundError   → 404 {"message":…}            readError → null
```

Two moderators open the same restriction in the retool panel; the second submits and sees **"Restriction ruling returned 400."** The symptom this fix quotes — *"Restriction ruling: An unexpected error occurred."* — comes from that same `${label}: ${reason}` line, so the commit cites a surface the fix does not reach. The three new tests assert `body.message`, which the real consumer never reads: real helper driven, wrong field asserted. One line to fix, either end.

### 🟢 F-3 — M8's kill count is not reproducible, and the fixtures are shape-drifted · `guard`/`scaffolding`
Body says 6 of 20; measured **8** (isolated) / **11** (full revert). The list cases match exactly; the message cases do not — all five die, not three, because the base parser's message regex requires `return (RULINGS_WIRED_FOR` while the real module writes `return (RULINGS_WIRED_FOR as readonly string[]).includes(type)` and **all five fixtures** omit the cast. So the fixtures are not faithful replicas, nothing pins them to the module, and M8 measured the old reader against a shape the module does not have. The error *understates* the new reader, so it creates no false confidence — but it is a number in the evidence table this ladder is meant to trust.

### 🟢 F-4 — the main app's `unit` project now depends on a generated, gitignored artifact in another app · `guard`/`scaffolding`
Importing the moderator module makes Vite load `apps/moderator/tsconfig.json`, which extends the gitignored `.svelte-kit/tsconfig.json`. Without `svelte-kit sync`, both suites fail with `TSConfckParseError: failed to resolve "extends"` — an error naming a tsconfig, not the seam. CI is fine (`prepare` runs sync). The finding is that "keep it import-free" is now an incomplete statement of the coupling, and the failure lands on the *main* app's suite. One sentence in the harness docblock closes it.

### Once-per-ladder sweep — every count re-derived at the current head
97 main-app / 463 moderator / 43 files / 36 skipped / 30 lint-rule guards / 459 passed / 0 type errors — **all exact**. The five round-3 delta reds reproduce with their exact assertions. M7 reproduces exactly. `.prettierignore` confirmed load-bearing by positive control (3 of 5 fixtures fail with `--ignore-path /dev/null`, pass with it) and scoped to that directory only. "Adding a `$lib/…` import fails loudly" confirmed — `Cannot find module`, exit 1, not a caught empty list. Only **M8** failed to reproduce (F-3).

**Environment fact, since the record was ambiguous:** `reports.queries.explain.test.ts` **skips** here (no `DATABASE_URL`); `ECONNREFUSED` was never observed this round. The body describes both outcomes in two places.

**Could not verify:** whether Retool (out of repo) branches on the old 500 or reads `error` vs `message` — F-2 is measured only against the in-repo consumer. The whole-PR-revert historical totals were not rebuilt.

**Scope bound, unchanged:** nothing in production files a non-`generation` restriction, so all three refusal paths and the entire vocabulary seam are exercised only by tests today.

ZacxDev:
## Round 3 fixes — what was claimed

Commits `62c11bd3b1`, `be95cc21a9`. Payload lines changed this round: **79** (plus 262 scaffolding) — a new shipped module and a rewritten consumer, not comment-only.

```audit-claims round=3 audited=48ff122bf1..be95cc21
1. A text assertion now sits alongside the execute check, claimed to refuse a moderator vocabulary whose values are read from the environment (`import.meta` / `process.env`).
2. The "keep this module import-free" comment is claimed to state the real constraint rather than a narrower one, and the `computed.ts` fixture is claimed to distinguish values assembled from module constants from values assembled from the environment.
3. `readError` was widened consumer-side, claimed to make the three 4xx refusal reasons readable by the moderator app; the body-shape rule moved into an import-free `apps/moderator/src/lib/server/rest-error-reason.ts` that the main-app suite loads by filesystem path, claimed so that one process asserts the field the real consumer reads.
4. The emitter-side fix was deliberately NOT made: `handleEndpointError` is the chokepoint for 36 route files, its 4xx and 503 bodies are pinned `toStrictEqual({ message })` with the 503 as a documented carve-out, and `restErrorBody` needs a `RestErrorCode` not derivable there without a new status→code map reconciled against a closed key ledger. Recorded as a separate PR.
5. All five fixtures are claimed faithful to the real module's `as readonly string[]` shape, and M8 re-measured at **6 of 21**.
6. The harness docblock is claimed to name the `.svelte-kit` coupling and its failure mode.
```

**Self-caught during the round, and disclosed:** the first implementation of the environment guard scanned raw text and was **matched by its own documentation** — the module has to name the shapes it refuses, so the sentence forbidding the thing tripped the guard. It now strips comments first, with a control pinning that a commented mention does not trip it.

**One correction to the audit, in the audit's favour:** F-3 asked for the "Invariant guards" bullet about the multi-line-array and trailing-comma fixtures to be negated. Making the fixtures faithful **restores** that bullet's truth (both are now read correctly by the old parser, messages included), so it was annotated rather than negated. The M8 denominator moved 20 → 21 with the new F-1 test.

**Not verified:** Retool is out of repo — whether it reads `error` vs `message`, or branches on the old 500, is unchecked, so F-2 is measured only against the in-repo consumer. No live run, no DB. The comment-stripper is not a full TS parser: a regex literal containing `//` would confuse it; none exists in the scanned module, and a string cannot hold an executable env read.

CI on the pushed head: **12/12 correctness gates pass**; only `preview / deploy` outstanding.

ZacxDev:
## Round 4 — the ladder ends here

Range `48ff122bf1..be95cc21a9`. **No 🔴, no 🟡. Three 🟢, none blocking.** All six of round 3's claims verified as **actually fixed**, each by watching the guard go red *for its own reason* rather than by reading the diff. Nothing was made worse.

### The two mechanisms are genuinely complementary — measured, not asserted
Mutating the **real** module to `import.meta.env.DEV ? ['generation'] : [...]` turns both suites red with the guard's own message. With `assertEnvironmentIndependent` neutered, the identical mutant is **28 passed / 0 failed** — so the execute reader alone really is blind, and the text check really does close that half. Reachability proved too: the guard's unit tests feed it synthetic strings, but the real module is reached through `readModeratorVocabulary()`.

M8 rebuilt from scratch against the base commit's parser: **6 of 21**, exactly the three named fixture pairs. Reproduces the PR body.

### Findings — all 🟢, all filed rather than fixed except the two documentation ones

**F-1 — the environment guard is walkable by indirection.** Three shapes escape *both* mechanisms, each measured at 29 passed / 0 failed while the apps genuinely disagree: an aliased global (`const P = (globalThis as any).process` → `P?.env`), a computed member access (`P['env']['NODE_ENV']`), and a regex literal containing `//` on the **same line** as the read, which the comment stripper truncates. Note the last requires same-line co-location — on a preceding line it is correctly caught, which is worth knowing before anyone "fixes" it.

Not blocking: every spelling a maintainer would plausibly reach for is covered, and `$app/environment` / `$env/*` break loudly. The real defect was the docblock claiming the class. **Fixed in `777c448b7e`** — the note now names the two spellings it refuses, records the three escapes, and states what *is* covered so it does not read as an indictment.

**F-2 — the guard fires on a mention inside a string literal.** Strings are deliberately not stripped, so `export const HINT = 'never read process.env here'` trips it. **Fails safe** — red with a clear message, never silently green — but it is a real constraint on the scanned module that the docblock did not state. **Fixed in the same commit.**

**F-3 — an observable message change on the 401 path**, recorded as a behaviour change and **not** a defect: an authorization failure now reads "You are not authorized…" instead of "sign in again", which is the correct text, since re-logging-in cannot fix an authorization failure. The old branch was already dead for the case it was written for.

### Not filed as findings, recorded so they are not lost
- `rest-error-reason.ts:38-46` selects on `typeof === 'string'` but discards on truthiness, so `{ error: '', message: 'real reason' }` returns `null`. No emitter produces that shape today — unreachable, noted only because a future one could make it so.
- **Nothing enforces that `readError` still calls `restErrorReason`.** The main-app test asserts the module, not its consumer, so re-inlining the old `error`-only logic would regress production with the new test green. Closing condition if anyone wants it filed: a lint-rule test asserting `user-actions.service.ts` imports `restErrorReason` and holds no second `body?.error` read, watched red with the import removed.

### Carried, not introduced
`preview / smoke-tests` is FAILURE at head (`collection.upsert + saveItem round-trip`) and its own bot comment says **report-only, does not block the build**. It has been red at every head since the round-1 fixes and green at `06991e8417`; it is unchanged across round 4's range and touches nothing this PR modifies. That check is what makes `mergeStateStatus: UNSTABLE`. Its root cause was **not** established — only when it started and that it is report-only.

### Ladder ledger
| Round | Findings | Payload lines its fixes changed |
|---|---|---|
| 1 (full) | 2 🟡, 4 🟢 | 216 |
| 2 (delta) | 2 🟡, 3 🟢 | 60 (only 9 executable) |
| 3 (delta) | 2 🟡, 2 🟢 | 79 |
| 4 (delta) | 3 🟢, none blocking | — |

Never two consecutive zero-payload rounds, so the attribution gate never fired; the ladder stopped on **findings**, which is the correct stop condition. A fifth round to confirm this one would be the mistake the process explicitly names, and is not being run.

**Could not verify, across the ladder:** no live run and no database at any point — the SQL ordering is asserted against compiled text, never executed. Retool is out of repo, so whether it reads `error` vs `message` or branches on the old 500 is unchecked. The Svelte UI was never rendered (no component-test harness in `apps/moderator`). And the whole `bot-account` path remains **test-only**: no production code passes a `type`, so every refusal path and the entire vocabulary seam are exercised only by tests.

ZacxDev:
### CI on the final head `777c448b7e` — 19/19 terminal, 18 SUCCESS + 1 SKIPPED, zero failures

Polled for **terminal conclusions** with a minimum row count, not for the absence of "pending" — an empty or partly-registered rollup settles instantly and looks identical to a pass.

🔴 **One correction to my round-4 comment above.** It reported `preview / smoke-tests` as FAILURE at head. On **this** head that check **does not exist** — no check-run and no commit status matching `smoke` on `777c448b7e`. So it is **absent, not fixed**: I did nothing to it, its root cause is still unestablished, and it may well reappear on a later head. An absent check is not a passing check, and this rollup should not be read as having cleared it.

`preview / deploy`, `auth-tests`, `component-tests` and `render-check` all ran and passed here, as did both `tekton` checks — several of which had not registered on earlier heads, so this rollup is wider than the 12–14 rows seen in previous rounds.

ZacxDev:
**MERGED** — squash `ec49115e55` at 2026-09-04T02:33:03Z, branch deleted. Verified on `origin/main` by
content (a squash merge never makes the head an ancestor, so ancestry would falsely read "not merged").

**The smoke-test failure cleared when `main` was merged in**, and the honest reading is narrower than
"fixed": the branch was 7 commits behind, and merging brought it current *and* forced a fresh preview
deploy. That rules out the change being inherently broken, but it does **not** distinguish a stale
preview environment from something in `main` the branch needed. Root cause was never established, and
nobody should quote this as diagnosed.

Final state at merge: 20/20 terminal, 18 SUCCESS + 1 SKIPPED + smoke SUCCESS, `mergeStateStatus: CLEAN`.

🔴 **Operational note for anyone gating on this repo's checks:** `preview / smoke-tests` registers only
*after* `preview / deploy` finishes, so the rollup settles **19/19 all-green with smoke absent** and then
smoke lands minutes later. That happened on two consecutive heads here, and on the first one it landed
**FAILURE**. A minimum-check-count assertion does not catch it, because the missing check has no prior
row on that head to count. Wait for the named check, not for "nothing is pending".

PR #4614, opened 2026-09-04
feat(app-blocks): allowlist three more blocks for the init-fragment fast path

ZacxDev:
Adds `custom-generators`, `model-benchmarking` and `sensei` to `BLOCK_INIT_FRAGMENT_ALLOWLIST`.

## Why

All three now ship an inline pre-paint fragment reader and declare `bootSkeleton: true`, so the full-page run host stands its loading treatment down and the block's own boot paint is what a viewer actually sees. Without an allowlist entry that paint has to **guess** the theme from `prefers-color-scheme`; with one it is the host's real answer. The difference is between "usually right" and "cannot disagree" — and a wrong guess repaints at `BLOCK_INIT`, which is exactly the flash the boot skeleton exists to remove.

The readers are already deployed and inert. This is the switch that makes them do anything.

## Both halves of the bar, checked per block

These three shipped in one batch, so nothing here is inherited from `app-requests`:

- **(a) Ships a decoding reader** — each serves an inline reader recording the resolved theme on `data-civitai-boot-theme`. Verified against the **served artifact** in each running container, not against the repo.
- **(b) Reads `location.hash` nowhere at runtime** — measured over each block's shipped `src/`: the only occurrences are inside its own `bootFragment` test, which *sets* the hash to drive the reader. No app code reads or rewrites it, so the transport's strip-after-init cannot race a competing consumer. This is precisely the check `playable-collections` fails, and it remains denylisted.

**Keying:** all three are page-mounted, `blockId` equals the store slug, `slots` empty — so one string matches whichever key the surface passes, the same shape as `app-requests`. Not safe to copy to a slot-mounted block, which knows only a `blockId`.

## Tests

The whole-set ledger is updated deliberately — it is designed to fail on an addition, which is what makes widening a decision rather than a drift.

Added **per-block** enablement and surface-leak assertions rather than one shared case: three entries landing together means a single assertion would pass with any one member present, which is how a batch addition loses a member silently.

## 🔴 One added test was vacuous, and the rewrite is the point

The denylist test as first written asserted that `playable-collections` is refused. That passes for the **wrong reason** — the block is refused for *not being allowlisted*, not because the denylist works. I proved it by rewiring the production wrapper to pass an **empty denylist**: the test stayed green. A guard reading as protection while providing none.

It now feeds a synthetic allowlist that **does** contain the block alongside the **real** denylist, with a positive control showing the same call returns `true` when the denylist is empty — so a `false` can only be the denylist check winning.

## 🔴 A gap I found and did not close

**No test — pre-existing or added — catches the production wrapper being handed the wrong lists.** The rewrite above calls the injectable `blockInitFragmentEnabledWith` and therefore bypasses `blockInitFragmentEnabled`. Closing it behaviourally needs a block present on **both** lists, and none exists. The wrapper is a three-line delegation, so the exposure is small — but it is recorded here rather than papered over, and it is worth knowing before someone assumes the denylist is covered end to end.

Also worth stating plainly: five of the six mutants I ran against this gate were killed by **pre-existing** tests, not the ones added here. That file already covers the lookup branches, denylist precedence and the surface gate through its injectable-list tests. The added tests earn their place by naming each new block, not by covering new mechanism.

## Verification

Gate suite green (23 tests). `pnpm typecheck` clean — **0 type errors**, read from the runner's own summary rather than an exit code.

https://claude.ai/code/session_01S5cBVVja4n7qzYU1FCBqZ8

### DIFF
--- a/src/components/AppBlocks/__tests__/blockInitFragmentGate.test.ts
+++ b/src/components/AppBlocks/__tests__/blockInitFragmentGate.test.ts
@@ -248,22 +248,25 @@ describe('blockInitFragmentEnabled — the PRODUCTION allowlist', () => {
     expect(BLOCK_INIT_FRAGMENT_ALLOWLIST.has('playable-collections')).toBe(false);
 
     const allowedAnyway = new Set([...BLOCK_INIT_FRAGMENT_ALLOWLIST, 'playable-collections']);
-    // Positive control: with the denylist EMPTY the same call returns true, so a
-    // false below is the denylist and not the surface or a missing entry.
-    expect(
-      blockInitFragmentEnabledWith(
-        { surface: 'page-run', slug: 'playable-collections', blockId: 'playable-collections' },
-        allowedAnyway,
-        new Set<string>()
-      )
-    ).toBe(true);
-    expect(
-      blockInitFragmentEnabledWith(
-        { surface: 'page-run', slug: 'playable-collections', blockId: 'playable-collections' },
-        allowedAnyway,
-        BLOCK_INIT_FRAGMENT_DENYLIST
-      )
-    ).toBe(false);
+
+    // 🔴 EACH KEY ALONE, not just the combined shape — the same rule the `it.each`
+    // above states, which the first version of THIS test broke by passing `slug`
+    // and `blockId` together. Measured: with both keys supplied, deleting EITHER
+    // denylist branch on its own left this test green, because the surviving branch
+    // still matched. Driving each key alone is what makes a single deleted branch
+    // fail here rather than only in the pre-existing per-key tests.
+    for (const args of [
+      { surface: 'page-run', slug: 'playable-collections' },
+      { surface: 'page-run', blockId: 'playable-collections' },
+      { surface: 'page-run', slug: 'playable-collections', blockId: 'playable-collections' },
+    ] as const) {
+      // Positive control: with the denylist EMPTY the same call returns true, so a
+      // false below is the denylist winning and not the surface or a missing entry.
+      expect(blockInitFragmentEnabledWith(args, allowedAnyway, new Set<string>())).toBe(true);
+      expect(
+        blockInitFragmentEnabledWith(args, allowedAnyway, BLOCK_INIT_FRAGMENT_DENYLIST)
+      ).toBe(false);
+    }
   });
 
   it('🔴 enables the fast path for app-requests on page-run', () => {
--- a/src/components/AppBlocks/__tests__/blockInitFragmentGate.test.ts
+++ b/src/components/AppBlocks/__tests__/blockInitFragmentGate.test.ts
@@ -192,7 +192,78 @@ describe('blockInitFragmentEnabled — the PRODUCTION allowlist', () => {
     // Pinned as a sorted array rather than `.has(...)` + `.size`: an exact
     // whole-set comparison fails on an ADDITION as well as a removal, and names
     // the offending entry in its own diff.
-    expect([...BLOCK_INIT_FRAGMENT_ALLOWLIST].sort()).toEqual(['app-requests']);
+    expect([...BLOCK_INIT_FRAGMENT_ALLOWLIST].sort()).toEqual([
+      'app-requests',
+      'custom-generators',
+      'model-benchmarking',
+      'sensei',
+    ]);
+  });
+
+  // 🔴 EACH NEW BLOCK IS ASSERTED ON ITS OWN, not covered by the `app-requests`
+  // case below. Three entries went in together, and one shared assertion would
+  // pass with any single member present — which is how a batch addition loses a
+  // member silently. Both keys and each key alone, for the same reason the
+  // `app-requests` case gives: a regression dropping one lookup branch still
+  // passes if only the combined shape is tested.
+  it.each(['custom-generators', 'model-benchmarking', 'sensei'])(
+    '🔴 enables the fast path for %s on page-run, by either key',
+    (id) => {
+      expect(blockInitFragmentEnabled({ surface: 'page-run', slug: id, blockId: id })).toBe(true);
+      expect(blockInitFragmentEnabled({ surface: 'page-run', slug: id })).toBe(true);
+      expect(blockInitFragmentEnabled({ surface: 'page-run', blockId: id })).toBe(true);
+    }
+  );
+
+  // The surface check must hold for the new entries too: allowlisting a PUBLISHED
+  // app must never reach a moderator's preview of its next UNREVIEWED submission,
+  // nor the author's dev tunnel. Paired with the page-run assertion above, which is
+  // what stops this being vacuous — the SAME id returns true on an eligible
+  // surface, so only the surface check can be producing false here.
+  it.each(['custom-generators', 'model-benchmarking', 'sensei'])(
+    '🔴 does NOT leak onto dev-tunnel or review-preview for %s',
+    (id) => {
+      for (const surface of ['dev-tunnel', 'review-preview'] as const) {
+        expect(blockInitFragmentEnabled({ surface, slug: id, blockId: id })).toBe(false);
+      }
+    }
+  );
+
+  // 🔴 The DENYLIST must still beat the widened allowlist. `playable-collections`
+  // shipped the same boot skeleton as the three above, in the same batch, and is
+  // the one block of that set which reads `location.hash` for its own routing — so
+  // it is deliberately NOT allowlisted.
+  //
+  // 🔴 THIS TEST WAS VACUOUS AS FIRST WRITTEN and the rewrite is the point. It
+  // asserted `blockInitFragmentEnabled({slug:'playable-collections'}) === false`,
+  // which passes for the WRONG REASON: that block is refused because it is not in
+  // the allowlist, not because the denylist works. Rewiring the production wrapper
+  // to pass an EMPTY denylist left it green — a guard reading as protection while
+  // providing none.
+  //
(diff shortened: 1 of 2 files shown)


ZacxDev:
## 🔴 Correction — two claims in the PR description above are wrong

An adversarial audit of this branch refuted both. Correcting here rather than by silently editing the body, since anyone who has already read it read the wrong version. Fixes are in `dbee5b2fcf`.

**1. "No test — pre-existing or added — catches the production wrapper being handed the wrong lists." This is false.**

Swapping the wrapper's two arguments — `blockInitFragmentEnabledWith(args, DENYLIST, ALLOWLIST)` — is killed by **four pre-existing tests**, one of whose comments says it exists for precisely that swap. I generalised from a single surviving substitution to "the wrapper is uncovered", which invites the next reader to re-add coverage that is already there.

The actual residual is **one** substitution: the real allowlist with an **empty** denylist. It survives because the two production sets are disjoint, so no input can distinguish an empty denylist from the real one — an **unobservable state**, not a missing assertion. Closing it behaviourally would need the sets to intersect. That part of the body was right; the sentence around it was not.

**2. "No app code reads or rewrites it" is false for `custom-generators`.**

That block rewrites the URL through `location.href` + `history.replaceState` (its `stripDeeplinkParam` removes only `?g=` and **preserves** the fragment) — invisible to a `location.hash`-shaped grep, which is what I measured with.

No live defect: the SDK's `seedFromFragment` strips synchronously at transport start and reads `href` live, so it cannot re-instate an already-stripped fragment. But the **admission criterion** was wrong, and that is the part the next person re-applies. `blockInitFragmentGate.ts` now states it as *"nothing re-instates the fragment after the transport strips it"* and names `location.href` / `replaceState` / `pushState` as things to check alongside `location.hash`.

## Also fixed

**The rewritten denylist test broke the rule this PR's own comment states 40 lines above it.** It passed `slug` and `blockId` together, so deleting *either* denylist branch alone left it green — the surviving branch still matched. It now drives slug-only, blockId-only and both. Measured after the fix: deleting the slug branch fails it, deleting the blockId branch fails it. Neither did before.

## Recorded, deliberately not fixed

**An entry is per-block, not per-surface, so it pre-authorises the model slot.** `IframeHost` calls the gate with `surface: 'model-slot'` and a bare `blockId`, and these entries *are* blockIds. If one of these blocks later adds a `slots` entry in its own repo and clears App-Block moderation, the fragment starts being appended on model pages with no change here, no reviewer on this repo, and no test going red.

Low impact — the fragment carries only theme / renderMode / blockInstanceId, all of which `BLOCK_INIT` already delivers — but the decision widens without anyone taking it, so it is now stated at the entry rather than left implicit.

## Closed by measurement rather than argument

The audit could not verify that `model-benchmarking`'s `bootSkeleton: true` is in its **approved** manifest: the version carrying it was unapproved when its commit message was written, and the mod-gated endpoints answered 404/401. It has since been approved. The `app_blocks` rows now read `model-benchmarking 0.4.1`, `custom-generators 0.6.4` and `sensei 0.1.15` — all `approved`, all with `bootSkeleton: true`. So the entry is not inert for that block, which was the audit's one pre-merge concern.

## Standing residuals

- The transient pre-ready paint has never been photographed, for any block in this arc.
- Whether `trustTier` grants `allow-same-origin` for these three is unverified. Without it the frame is opaque-origin, `replaceState` throws and the SDK's strip is silently swallowed, so the fragment would linger in `location.hash` for the session. Harmless given the criterion above holds, but it is an unstated precondition of "the transport strips the fragment after init".

ZacxDev:
Round 1 audit claims block, for the next round to anchor on.

legend: `<from>` = the tip this round's audit READ · `<to>` = the head this round's FIXES produced.

```audit-claims round=1 audited=8c6de62092..dbee5b2f
1. F2 — the PR body's claim that NO test catches the production wrapper being handed the wrong lists is retracted publicly; the residual is restated as one substitution only (real allowlist + empty denylist), attributed to the two sets being disjoint.
2. F4 — claim (b) in blockInitFragmentGate.ts is corrected: "no app code reads or rewrites it" was false for custom-generators, and the admission criterion is restated as "nothing re-instates the fragment after the transport strips it", naming location.href / replaceState / pushState alongside location.hash.
3. F3 — the rewritten denylist test now drives slug-only, blockId-only and both keys, so deleting either denylist branch on its own fails it; previously both deletions left it green.
4. F5 — the model-slot pre-authorisation is documented at the entry: an entry is per-BLOCK not per-surface, so a block that later adds a slots entry in its own repo starts receiving the fragment on model pages with no change here.
5. F1 — closed by live read rather than by a code change: app_blocks now shows model-benchmarking 0.4.1, custom-generators 0.6.4 and sensei 0.1.15, all approved with bootSkeleton true, so the entry is not inert for model-benchmarking.
```

ZacxDev:
Round 2 audit claims block, for round 3 to anchor on.

```audit-claims round=2 audited=dbee5b2fcf..bc3a7f9d
1. 🟡A — the comment no longer rests custom-generators' admissibility on the SDK's seedFromFragment strip; it now states, with the measurement, that the strip is INOPERATIVE for every block listed here (all four are trust_tier unverified, sandbox.ts adds allow-same-origin only for TRUSTED_TIERS internal/verified, so the frame is opaque-origin, replaceState throws and the throw is swallowed), that the fragment therefore persists in location.hash for the session, and that the three are safe because of criterion (b) itself rather than because of a strip that never executes.
2. 🟢B — "that call reads location.href LIVE" no longer attaches the live read to seedFromFragment; the live href read is attributed to custom-generators' own getHref, which is the one that carries the argument.
3. 🟢C — "The criterion is …" is replaced by a SECOND check applied ALONGSIDE the read-side one, so a block that reads the hash for its own routing but never re-instates is still disqualified; playable-collections is named as that case.
4. Nit — "no civitai reviewer involved" is corrected to "no reviewer ON THIS REPO involved", since such a block does clear App-Block moderation.
5. Not a code change — the once-per-ladder cross-reference sweep was run against the current head, ignoring ranges: TRUSTED_TIERS/allow-same-origin gating (sandbox.ts:22,44), the model-slot call site (IframeHost.tsx:1253), the fragment's three fields (blockIframeUrl.ts:118-120), page-run passing blockId and slug both as page.blockId (run/[slug]:147,155), and the four blocks' live trust_tier/sandbox/version/bootSkeleton rows all re-derived and matching.
```

ZacxDev:
## Ladder stopped after round 3 — rationale, stated rather than implied

Three adversarial rounds ran. Rounds 1 and 2 each found things that needed fixing and were followed by another round. I am stopping after round 3's fixes on the stated criterion for a ladder whose payload has become prose, and the criterion requires naming why the rounds will not stop on their own:

- **No 🔴 in any round.** All three verdicts were "safe to merge".
- **Zero executable payload lines changed in rounds 2 and 3** — mechanically checked, with a positive control confirming the filter can emit non-comment lines. Runtime behaviour has been settled since the first commit; the only executable change this PR makes is the three allowlist strings.
- **Blast radius bounded to "the comment contains a false or imprecise sentence."** No finding after round 1 could alter what the gate does.
- **The recurring shape has been swept at every site, not just where it was reported.** All three rounds produced one defect class: *a claim true of one block or one mechanism, asserted more widely than measured*. I enumerated every absolute in the added comment and re-derived each. That sweep found one instance **no round flagged** — "none of them READS the fragment" is false for `custom-generators`, because `location.href` contains the fragment and `getHref()` reads it on every call. Fixed in `0f5cbafa9e`: the criterion is about a block that reads the fragment **to use it**, so the word is now *consumes*.

Left alone, the next round would keep finding real things about prose the previous round wrote. That is the condition this stop rule exists for, not a converging ladder cut short.

```audit-claims round=3 audited=bc3a7f9d66..0f5cbafa9e
1. 🟡-1 — the self-contradiction is resolved: the paragraph no longer says custom-generators re-instates nothing while also citing its replaceState. The live-vs-captured distinction is restored and sharpened — getHref() is evaluated at call time (App.tsx:231, consumed at :615, read from origin/main not the stale local clone), so the write carries the current fragment and is a no-op with respect to it; a CAPTURED href is named as the unsafe shape; and the second check is stated as a prompt to inspect each hit rather than a filter that rejects on sight.
2. 🟢-2 — playable-collections is withdrawn as evidence that the read half is independently necessary, since it fails both halves; the read half now rests on the principle alone.
3. Shape sweep (not audit-reported) — "none of them READS the fragment" corrected to "CONSUMES", with the location.href-contains-the-fragment distinction spelled out.
```

## Recorded as open, so it is known rather than assumed absent

1. **One mutant still survives**: the production wrapper handed the real allowlist and an **empty** denylist. Not closeable behaviourally while the two sets are disjoint — no input can distinguish an empty denylist from the real one. An unobservable state, not a missing assertion.
2. **The model-slot pre-authorisation is documented, not closed.** An entry is per-block; if one of these blocks later adds a `slots` entry in its own repo and clears App-Block moderation, the fragment starts being appended on model pages with no change here and no test going red.
3. **Whether the AppBlocks `unit` tier is green is unresolved.** Round 2 measured `hiddenBlocks.test.ts` failing 7/7 (control-verified as pre-existing and untouched by this PR); round 3 could not reproduce it at the same head. One of those observations is environment-specific and I have not determined which. Unrelated to this PR either way, but "the gate passed" here is a claim about the gate suite file, not the whole tier.
4. **Criterion (a) — "verified against the SERVED artifact"** — round 3 re-derived it for `custom-generators` only. The other two rest on my own earlier reads of the running containers, not on an independent check.
5. **The transient pre-ready paint has never been photographed**, for any block in this arc.

Head `0f5cbafa9e`: gate suite 23/23, typecheck 0 errors, prettier clean.

PR #4618, opened 2026-09-04
feat(articles): mark Civitai-published articles, and make adminOnly tags actually admin-only

JustMaier:
Adds an **Official** marker to articles published by Civitai, and fixes the permission hole that would have made it forgeable.

Justin, 2026-09-04: a tag rather than a category, because an article carries exactly one category (`ArticleUpsertForm` makes the author pick one) and official articles span several of them — an official announcement and an official guide are both official.

## 🔴 The permission half is a bug fix, and it cannot land after the tag

`upsertArticleHandler` resolved the incoming tags through `getCategoryTags('article')`, so its `adminOnly` check only ever saw adminOnly tags that were **also article categories**. An adminOnly tag that is not a category — exactly what this marker is — was not checked at all. Article tags attach by **name** through `connectOrCreate` (`article.service.ts`), so any user could have typed `civitai official` onto their own article.

The guard now asks the tag table directly and covers every adminOnly tag. The old behaviour is a strict subset: an adminOnly category is still an adminOnly tag.

`findAdminOnlyTags` resolves by id **and** by lowercased name. The name arm is the load-bearing one — nothing obliges a caller to send an id, and the web client never does. It is deliberately **uncached**, unlike `getSystemTags`/`getCategoryTags`: a 4h TTL on a listing is fine, a 4h TTL on a permission check is a window where anyone can apply the tag. A moderator pays no query at all, because it runs only when the caller lacks the feature.

**Measured on prod before writing any of this:** `SELECT count(*) FROM "Tag" WHERE "adminOnly"` → **0**. The column has never been used, so this tag is the first thing that guard will ever refuse.

## Why not the `official` tag that already exists

It exists (id 123833), targets `Post`, is **not** adminOnly, and **8 posts already carry it**, applied by ordinary users. Flipping it to adminOnly would retroactively mark those. Hence `civitai official` — and hence a deliberately exact match: `includes('official')` would paint a Civitai provenance badge on all 8, which is pinned as a test.

## ⚠️ The migration is not applied

It needs running by hand against prod:

```
packages/civitai-db-schema/prisma/migrations/20260904160000_civitai_official_article_tag/migration.sql
```

Until then **the code is inert** — no tag, no badge, and the guard has nothing to refuse. It is idempotent, and a second run repairs a row whose `adminOnly` was cleared by hand.

## Two facts that live only in a comment and a test

Both would unset every badge with no error and a green suite:
- **Renaming the tag row.** The client matches on the name.
- **Clearing its `adminOnly`.** That does not hide the badge — it makes it forgeable.

## Testing

24 tests across two files. Every mutation driven red on purpose; both suites pristine at 6 and 18.

| Mutation | Result |
|---|---|
| guard narrowed back to categories only | 2 failed |
| ids only, name arm dropped | 1 failed |
| names no longer lowercased | 1 failed |
| query moved before the feature check | 1 failed |
| `adminTags === false` instead of `!adminTags` | 3 failed |
| card mount removed | 1 failed |
| detail mount removed | 1 failed |
| detail tag-row filter reverted | 1 failed |
| match loosened to `includes('official')` | 4 failed |
| migration renames the tag | 1 failed |
| migration drops `adminOnly` | 1 failed |

One control I discarded rather than counted: my first "narrow it back" mutant killed all 6 tests, which is a broken mutant, not a demonstration — it failed on an unmocked `getCategoryTags` reaching redis. The faithful version kills exactly the two that should die.

`typecheck: OK — 0 type errors`. ESLint clean on the changed files (2 pre-existing unused-import warnings in `tag.service.ts`). Full suite `Test Files 1 failed | 1597 passed | 3 skipped (1601)` — the one failure is `appListingMenuSurface.test.ts`, which compares Windows paths against forward-slash literals and fails on a clean tree. `blocks.router.workflow.test.ts` collected **370**.

## Open question I did not decide

`ctx.features.adminTags` is `['mod', 'granted']` — moderators **and** anyone individually granted the feature. You asked for mods only; those coincide today unless someone holds a grant. Say the word and I will make it a moderator check instead.

https://claude.ai/code/session_01XHZuQDTCG159qcPrCkP7SF

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260904160000_civitai_official_article_tag/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260904160000_civitai_official_article_tag/migration.sql
@@ -1,0 +1,33 @@
+-- Creates the tag that marks an article as published by Civitai rather than by a community
+-- author. Justin, 2026-09-04, after choosing option A on the design question.
+--
+-- A TAG and not a category, deliberately: an article carries exactly one category
+-- (ArticleUpsertForm makes the author pick one) and official articles span several of them --
+-- an official announcement and an official guide are both official.
+--
+-- 🔴 `adminOnly` is the whole permission model. `upsertArticleHandler` refuses any article
+-- carrying an adminOnly tag from a caller without the `adminTags` feature (mod, or granted).
+-- Clearing this column does not disable a badge; it makes the badge FORGEABLE by any user,
+-- because article tags attach by NAME through `connectOrCreate` and anyone can type a name.
+--
+-- 🔴 Renaming this row silently unsets every badge. The client matches on the name, from
+-- `src/shared/constants/official-article.constants.ts`. Rename in one place only and the badge
+-- disappears with no error anywhere.
+--
+-- Measured on prod 2026-09-04 before writing this:
+--   * `SELECT count(*) FROM "Tag" WHERE "adminOnly"` -> 0. Nothing has ever used this column,
+--     so this row is the first thing the guard above will ever refuse.
+--   * A tag named `official` already exists (id 123833), targets Post, is NOT adminOnly, and is
+--     on 8 posts. It is deliberately NOT reused: flipping it would retroactively mark 8 existing
+--     posts as official, and users have already applied it themselves.
+--   * No tag named `civitai official` exists.
+--
+-- Idempotent. `Tag.name` is unique, so a second run updates the existing row to the intended
+-- state rather than failing or duplicating -- which also repairs a row whose `adminOnly` was
+-- cleared by hand.
+INSERT INTO "Tag" ("name", "target", "adminOnly", "unlisted", "createdAt", "updatedAt")
+VALUES ('civitai official', ARRAY['Article']::"TagTarget"[], true, false, NOW(), NOW())
+ON CONFLICT ("name") DO UPDATE
+  SET "target" = ARRAY['Article']::"TagTarget"[],
+      "adminOnly" = true,
+      "updatedAt" = NOW();
--- a/src/components/Article/OfficialArticleBadge.tsx
+++ b/src/components/Article/OfficialArticleBadge.tsx
@@ -1,0 +1,34 @@
+import { Badge, Tooltip } from '@mantine/core';
+import { IconRosetteDiscountCheck } from '@tabler/icons-react';
+import clsx from 'clsx';
+
+import { OFFICIAL_ARTICLE_LABEL } from '~/shared/constants/official-article.constants';
+
+/**
+ * The marker on an article published by Civitai rather than by a community author.
+ *
+ * One component for both surfaces — the card and the article page — because this badge is
+ * a claim about provenance, and two hand-maintained copies of a provenance marker is how
+ * one of them ends up looking like an ordinary chip. `className` is the only thing a
+ * caller varies; the card passes its own chip class so the badge sits in the card's
+ * header rhythm.
+ *
+ * 🔴 Rendering this is NOT what makes an article official. The tag behind it is
+ * `adminOnly`, and `upsertArticleHandler` is what refuses it from a non-moderator. Do not
+ * reuse this badge for anything a user can set about themselves.
+ */
+export function OfficialArticleBadge({ className }: { className?: string }) {
+  return (
+    <Tooltip label="Published by Civitai" withArrow>
+      <Badge
+        size="sm"
+        variant="filled"
+        color="blue"
+        className={clsx('uppercase', className)}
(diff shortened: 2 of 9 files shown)


JustMaier:
🛑 **Do not merge yet.** The five-lane review found that this is probably the wrong shape, and the recommendation is with Justin.

All five lanes read `06ce0f4b6bc0a4e98177c73460525219aebc490b` — confirmed against the head, per lane. All five ran as general-purpose agents reading `.claude/agents/civitai-*-review.md`; none as a native subagent type, because those do not resolve from the project this session is registered under. `git status --porcelain` on the implementation worktree afterwards: **empty**.

## The finding that changes the design

**`Model.isOfficial` already exists and is this feature, done the other way** — a boolean column (`schema.prisma:967`), a moderator-only setter (`setModelOfficial` throws unless `isModerator`, *stricter* than the `adminTags` flag), and a "Mark Official / Unmark Official" moderator menu item that uses `IconRosetteDiscountCheck` — the same icon this PR independently picked for the badge. I did not find it before building, so it was never an option on the decision card.

Copying it removes four defects rather than patching them:

**1. A moderator cannot actually apply this tag except by typing a magic string.** `TagsInput` — the component behind the article form's tag field — has no autocomplete at all: no query, no suggestions. So `includeAdminTags` in `tag.controller.ts:46`, the one thing that would surface an adminOnly tag to a mod, is never reached from that form.

**2. "Until then the code is inert" — in this PR's own body — is false.** Article tags attach with `connectOrCreate` on *name*, which **creates** a missing tag with `adminOnly` defaulting to false. Between merging this and hand-applying the migration, any user typing `civitai official` creates the row themselves, passes the guard (the row is not adminOnly), and gets the badge. The migration's `ON CONFLICT DO UPDATE` then flips that row to adminOnly and leaves the attachment in place — retroactively blessing a community article, which is exactly the outcome this PR reasons about avoiding for the pre-existing `official` tag.

**3. The guard checks the submitted tag list, not the delta.** Once an article is marked official, the form seeds its tag field from the article's existing tags, so **the owner's every subsequent save carries the adminOnly tag back and is refused** with a bare `UNAUTHORIZED` — they can no longer edit their own article, with nothing saying why. And a non-mod can *remove* the badge from their own article, because a payload without the tag never triggers the check.

**4. The badge is matched by tag NAME on the client**, so renaming the row silently unsets every badge — the hazard the migration's own comment warns about. The repo's idiom is the opposite: `feed-tag-bar.constants.ts` keeps names server-side and ships ids.

## Also found, and true either way

- **Two migration guards cannot fail.** `adminOnly` is set *positionally* in the `VALUES` list, and the test regex `/"adminOnly"\s*=\s*true/` only matches the `ON CONFLICT ... SET` line. Flip the `VALUES` boolean to `false` — the tag is created **not** adminOnly, the badge becomes forgeable — and the test named *"the migration creates it adminOnly"* stays green. Verified by the test lane against mutated copies outside the repo.
- **The badge mount guards cannot see a condition.** `{official &&` → `{false &&` (badge never renders) and `const official = true` (Civitai provenance on **every community article card**) both leave the guard green, because it text-matches an identifier on a line.
- **The controller suite never sets `input.id`**, so adding `&& !input.id` to the guard — a plausible "tags can't change on edit" tidy-up, and a total bypass — keeps all 6 tests green.
- **Perf is fine**: the guard query measures 0.361 ms on the prod replica, with article saves running ~60/day. Uncached was the right call.
- `official-article.constants.ts` cites the migration at a path that does not exist.

## Recommendation

Switch to `Article.isOfficial` mirroring `Model.isOfficial`, and close this PR rather than merge it. Defects 1–4 stop existing rather than getting patched, and "who may apply it" answers itself as `isModerator` — which also retires the open `adminTags` = `['mod','granted']` question.

JustMaier:
Closing in favour of the column. Justin took the `Article.isOfficial` option — mirroring `Model.isOfficial` (column + moderator-only setter + a Mark Official menu item), which removes the four defects this branch would have had to patch rather than fixing them:

- a moderator could only apply the tag by typing `civitai official` from memory, since `TagsInput` has no autocomplete;
- "inert until the migration runs" was false — `connectOrCreate` creates the missing tag at `adminOnly = false`, so a user could self-badge in the deploy window and the migration would then bless it retroactively;
- the guard checked the payload rather than the delta, locking an article's owner out of their own saves once it was marked;
- the badge matched on tag name, so a rename would silently unset every badge.

The widened `adminOnly` guard goes with it, deliberately: prod has **zero** adminOnly tags, so it fixes nothing today, and the payload-vs-delta defect above is in that guard. If someone wants it later it should land on its own, with the delta handling and with the four other services that attach tags by name unguarded (`bounty`, `collection`, `model`, `question`).

Replacement PR to follow. Review record for this branch is in the comment above.

PR #4625, opened 2026-09-04
feat(moderation): bot-account detection — three heuristics and a reporting threshold

ZacxDev:
Replaces the labelled always-zero placeholder with the first three real heuristics, and adds the reporting threshold the detector needed before it could have a real run.

**Still shadow mode.** Mutes nobody, bans nobody, writes no `UserRestriction` row. That property was re-verified, not assumed: the asserted operation ledger, the import ledger and the real `dbWrite` spy all still hold, and a **third ledger** was added because the first two are structurally blind to the new data source.

The LLM read is deliberately not here. Nothing was scaffolded for it beyond the existing registry shape.

---

## The threshold, and why it was not optional

`run.ts` did `cohort.members.map(...)` — **every cohort member became a finding**, with no confidence filter anywhere. With a do-nothing placeholder that was harmless. With three real heuristics it would put a whole day's posting cohort on a live moderator board as confidence-0 rows, batched across up to ten reports. A board that is mostly noise on its first day is a board nobody reads on its second, and that is not recoverable by tuning later.

- **`MIN_REPORTED_CONFIDENCE = 0.15`**, configurable per run via `minConfidence`.
- **Set against the blend's own arithmetic, not an intuition.** With three equal weights the blend is their mean:
  - one heuristic alone at 0.45 → **0.15** (the cut)
  - one heuristic alone at 1.00 → 0.33
  - two at 0.23 each → 0.15

  So it admits an account on the strength of **one** signal that is about half convinced, and rejects one where every signal is weak. That is the loosest cut that still means something — a tight threshold would report only the accounts nobody needed a detector to find, and would teach us nothing about where the real line sits.
- **It filters reporting, never scoring.** Every member is scored and bucketed whatever it is set to. `minConfidence: 0` restores the previous full-cohort behaviour and is how a deliberate grading run is requested.
- **It is a starting point, not a calibration**, and the code says so. No run has produced a graded finding, so the number comes from the weighting rather than from data. The bucket counters below exist to replace it.

### The suppressed members are counted, not dropped

A member nobody can see is a member nobody can grade, and grading is the entire purpose of the shadow phase. Every run emits, zeros included:

| counter | meaning |
|---|---|
| `findings_reported` | members at or above the cut |
| `findings_suppressed` | members below it |
| `report_min_confidence` | the cut actually applied |
| `confidence_bucket_0_10` … `confidence_bucket_90_100` | the distribution over **every scored member** |

The buckets sum back to `cohort_size`, and that equality is asserted — nothing goes from scoring to reporting without being counted. The summary sentence states the split too, because the summary is what a human reads first.

---

## The three heuristics

Each contributes an independent, separately-reportable sub-score. Equal weights, deliberately: nothing yet distinguishes them by precision, so any other split would be a number invented to look considered.

### 1. `posting-velocity` — items posted per hour of account age
Adapted from `apps/moderator/src/lib/server/comment-spam.service.ts`. What carried over: the rate is measured **against account age**, and the floor at zero for a clock that ran backwards. What changed: that service expresses age as a cut-off on a burst; the cohort here is already age-bounded to a day, so age becomes the **divisor** — same idea with more resolution, since it separates the twenty-minute account from the twenty-three-hour one where a cut-off keeps both.

Counts `posts.all`, **not** `posts.visible`. The canonical wave is forty uploads the scanner blocked; reading `visible` would score exactly the accounts this detector exists to find as idle, and silently.

**Cost: no query at all.** Both numbers already rode in on the cohort read. The only one of the three that cannot degrade.

### 2. `registration-cluster` — how many other new posting accounts share this one's registration IP or uncommon email domain
Adapted from `apps/moderator/src/lib/server/bulk-ban.service.ts`. Keeps its load-bearing choice: **registration IPs only, never logins** — a shared login IP is weak evidence because carriers and offices put thousands of unrelated people behind one address. Inverts its framing: that service is investigative (a moderator supplies the accounts), here the population is the cohort itself, which is a much narrower denominator and is what makes an unbounded count usable without a base rate.

The two halves combine with **`max`, not a sum**, so the sub-score always reads as "the size of the ring this account is in".

**The only heuristic here that can see a coordinated ring**, and the only one whose email half is free (a wider `select` on the existing `user.findMany` — no new operation). Its IP half is one ClickHouse read per chunk.

### 3. `content-templating` — the same text, posted by accounts that are supposed to be strangers
No existing implementation. Each comment is reduced to a fingerprint — lowercased, **links and digit runs masked**, punctuation dropped — then scored on how many **distinct** cohort members produced the same fingerprint.

The masking is the detection: an exact-match check finds only literal copy-paste, while a ring's whole method is one template with the payload swapped (the link, the referral code, the amount).

Chose exact-match-after-masking over cosine/TF-IDF, edit distance or MinHash because a moderator has to read the reason and see why. Those produce "similarity 0.83 to a cluster of 6", which nobody can check and nobody can calibrate without a corpus; this lets the finding **quote the shared text**. It is also O(rows) with a hash map against O(rows²) for pairwise distance, which the cohort bound rules out anyway.

**Cost: two `comment.findMany` reads per chunk, under a fixed run-wide row budget.**

### Structure
Cross-account signals are indexed **once per run** in `evidence.ts`; each heuristic is a pure function over that index. Per-account scoring cannot answer "how many others", which is exactly what misses a wave.

Every count is of **distinct accounts, never rows** — otherwise one account pasting its shill line ninety times manufactures a ring out of itself.

---

## A missing source is a state, not a zero

Two of the three can go dark. `clickhouse` is `undefined` on any deployment without `CLICKHOUSE_HOST`, and the content read is budgeted. A zero from a dead source is byte-identical to a zero meaning "these accounts share nothing", and the two call for opposite conclusions.

- `evidence_registration_ips`, `evidence_content_budget_exhausted`, `evidence_members_sampled_for_content` are published on **every** run.
- The summary warns explicitly when the IP source was unavailable or the budget ran out.
- A **partial** IP read is discarded rather than scored: a cluster count built from a partial read understates every ring that straddles the missing chunk, and understating is the direction that produces a confident zero.
- The content budget spends itself newest-first, matching the cohort walk's own truncation direction.

---

## Also fixed here

- **The heuristic notes were collected and then dropped.** `HeuristicScore.note` has always existed and every heuristic has always been asked to produce one, but the reason rendered only `id=0.00` — three numbers and no statement of what any of them *saw*. Invisible while the only registered heuristic was a placeholder whose note was always `null`.
- **`scoring.test.ts`'s member fixture was the wrong shape** — the pre-split flat `PostCounts`. `src/**/__tests__/**` is excluded from the root tsconfig so nothing typechecked it, and no assertion happened to read the bad field. A fixture that cannot occur in production tests a world that does not exist; corrected because this change gives `posts` a consumer.
- **The email address is reduced to its domain in the pure selector and stored nowhere else.** A whole-object `toEqual` on the member is what keeps it from reappearing.

---

## The third ledger

The operation ledger anchors on a `db` handle, so it **cannot see a ClickHouse call** — it stays at seven members whatever this module does to ClickHouse. `$exec` is on the same client and the same import, takes the same kind of string, runs arbitrary DDL/DML, and returns `void`.

That is not a tolerated gap. A new ledger asserts the ClickHouse method set is exactly `{$query}` and that every SQL statement in the module is a bare `SELECT`, each with a positive and a negative control.

**Its own positive control caught a defect in the scanner.** The pattern required the method name and its paren to be adjacent, so `ch.$query<{ ip: string }>(sql)` matched nothing — while `$exec(`, which nobody writes with type arguments, matched fine. The guard's negative control passed, its assertion read a reassuring `[]`, and it was blind to the only call the code actually makes. Caught only because the control asserts the *benign* shape is seen, not merely that the hostile one is.

---

## Test matrix

**131 → 243 tests, 8 files (2 new).** Every base-failure reason below was observed, not predicted.

### Red at `8d7faf6138`, behaviourally

Measured by running the new tests against the base source in a clean worktree:

| test | base failure |
|---|---|
| `newAccountPageArgs > selects only the columns … plus the email` | `expected { id, username, createdAt } to deeply equal { …, email }` |
| `selectCohortMembers > carries the account identity` | member object missing `emailDomain` |
| `selectCohortMembers > reduces the email to its domain` | `expected undefined to be 'a.test'` |
| `buildFinding > carries each heuristic's NOTE` | reason did not contain `Signals — registration-cluster: …` |

### Red at base only as an import error — **not behavioural coverage**

`normalizeEmailDomain` (×3 cases) fails at base with `normalizeEmailDomain is not a function`. `evidence.ts` and `heuristics/` do not exist at base at all, so every test in `evidence.test.ts` and `heuristics.test.ts`, and the threshold cases in `run.test.ts`, can only fail there by failing to import. **Stating that plainly: those are attributed by mutation instead**, per the table below.

### One invariant guard, labelled as such

`omits the signals clause entirely when nothing fired` **passes at base** — base also omits it, because base never renders notes at all. It pins an invariant the bug never violated. It is not regression coverage and is not counted as such.

### Mutation results — 35 mutants, 35 killed, 0 survived

Includes a positive control (an obviously-broken ramp) which was killed, proving the harness can go red. Each mutant was checked to fail in **its own** test, not merely to fail something.

Covered: both ramp boundaries and its interpolation; the velocity volume gate, age floor, `all`-vs-`visible` read and its rate boundary; the clustering `max`-vs-sum, the common-domain suppression and both cluster boundaries; the templating boundary and quote bound; the distinct-account counting on both axes, the cohort filter, both fingerprint floors, both text masks, the content budget, the `targetUserId` column and the integer filter; the threshold comparison and its value, the top-bucket clamp and the zero fill; and at run level, the buckets covering all scored members, the suppressed counter, the cut itself and the evidence flag.

**The battery found five real weaknesses, now fixed:**

1. `IP_ZERO_AT`, `CLUSTER_ZERO_AT` and `MIN_FINGERPRINT_CHARS` boundary cases were written as `membersPerIp: { x: IP_ZERO_AT }` — **the expectation was derived from the very constant under test**, so shifting the constant shifted the test with it and the case passed at any value. Rewritten with literals, with the constants pinned separately.
2. Both `largest*Cluster` fixtures put the answer **last**, so "take the max" and "take the last" agreed and `>` vs `>=` survived. Fixtures reordered, plus explicit tie-break cases.
3. The character floor on fingerprints was only ever killed by the **token** floor. An isolated case (≥4 tokens, <24 chars) now exercises it, with a positive control so the pair cannot pass by the function always returning `null`.

### Gates

- **Typecheck: 758 errors, byte-identical to base, 0 of them in this tree.** The 758 are pre-existing in this checkout (unresolvable `form-graph` / `kysely`), present identically in the primary clone.
- **Tests-typecheck gate:** base 1143 errors / 222 files, this branch **1137 / 220**. The 11 "got WORSE" files it lists are identical at base and none is touched here — measured in a clean base worktree rather than argued from the diff.
- Prettier clean over all 19 files (file count read, not just the exit code). ESLint: 0 errors.

---

## Round-1 adversarial audit — what changed

Verdict was *"safe to merge as it stands; not safe to schedule"*, with **F1, F2 and F3 named as blocking a real cron**. All three are fixed, plus the rest of the list. The full comment is above; this is what the code now does differently.

**F2 — nothing held the seam between the evidence and the scoring.** Replacing the `signals` handed to `scoreAccount` with `emptyCohortSignals()` left 243/243 green: two of three heuristics would have scored 0 on **every production run** while `evidence_registration_ips`, `evidence_distinct_registration_ips` and `evidence_distinct_content_fingerprints` all reported healthy values, because those read `signals` rather than what scoring saw. Three suites each hermetic, none building the combined state. `run.test.ts` now drives a six-account ring through the **production** registry with a real `EvidenceReader` and asserts the **emitted finding's** own sub-scores — with a no-evidence control arm, since a non-zero score attributes nothing without the zero beside it. Deleting `evidence: createEvidenceReader()` from the job was equally green (`evidence?` is deliberately optional), so that half is asserted on the job source.

**F1 — no private/carrier-internal IP filter**, unlike both sibling readers of the same table. The predicate is now **shared, not copied**: `PUBLIC_IP_ONLY` moved out of `reactor-lookup.service.ts` into `@civitai/shared/clickhouse-ip-filters` — a dependency of both the main app and `apps/moderator` — with `INTERNAL_IP_RANGE` re-exported from the moderator's `clickhouse-filters.ts` so its other importers are untouched. The SQL that file emits is byte-unchanged and its existing assertions still pass. `ip != ''` still comes first, because `isIPAddressInRange` raises on an empty string.

**F3 — the SELECT projection was unguarded.** Pinned, along with the `GROUP BY`, since the two drifting apart is the same silent zero from the other side.

**F8** — `listContentSamples` had no `try`/`catch`, so a replica timeout propagated out and **no report was filed at all**, losing the velocity heuristic's day too. It now degrades exactly as the IP read does: partial data discarded, `sources.contentSamples` false, a log line, an `evidence_content_samples` counter and its own summary sentence. It is **not** reported as an exhausted budget.

**F4** — the summary now names the read-ran-and-matched-nothing case; the PR body's overstatement of the `targetUserId` risk is corrected under *NOT verified*.

**F7** — `domains_suppressed_common` is a real counter rather than a claim in a comment.

**F9** — `AND time >= '<createdAfter>'`, in ClickHouse's `DateTime` literal form.

**F12** — `LIMIT ids.length * 4` was a cap on the **result** with no `ORDER BY`: one account with many addresses could evict every other account in its chunk. Now `LIMIT 4 BY targetUserId` plus an `ORDER BY`, so the comment and the code say the same thing.

**F13** — the six bounds a run is sized by are pinned, and the whole budget-exhausted summary sentence is asserted in full.

**F10** — both ledger patterns see optional chaining; the ClickHouse one also sees a bracket key. Each widening has its own control, plus controls that the benign shapes still read the same. What remains open is stated in the pattern's docstring.

**F6** — `COMMON_EMAIL_DOMAINS` extended to ~120 entries grouped by provider family, `pm.me` included. The code now states plainly what a word list structurally cannot do.

**F5** — measured, pinned, and deliberately not patched. See *NOT verified*.

**F11 is out of scope** (a query-side note about `batch_index = 1`, not code).

### Mutation, this round

21 mutants planted one at a time, each with a fresh `node_modules/.vite`; **21 killed, 0 survived**, each on the assertion written for it. The two the audit used are in that set: `signals` → `emptyCohortSignals()` and `SELECT targetUserId, ip` → `SELECT userId, ip`, plus deleting the job's `evidence:` line. Two positive controls (`MIN_REPORTED_CONFIDENCE` 0.15 → 0.5, killing 11; `rampScore` → 0, killing 7) confirm the harness can go red. One candidate control was **discarded as equivalent** rather than reported as a survivor: `BOT_ACCOUNT_DETECTOR` is asserted against the same imported constant, so mutating it moves both sides.

271 tests / 8 files, up from 243. Typecheck clean (0 errors). `svelte-check` on `apps/moderator`: 0 errors, 0 warnings. Prettier clean.

---

## NOT verified

- **No heuristic has been run against real data.** Every threshold and boundary in this PR is a judgement, not a measurement. The per-heuristic counters and the confidence distribution exist to replace them, and that has not happened yet.
- **The ClickHouse read has never executed.** `registrationIpSql` is asserted as a string and the reader is exercised against a fake; `default.userActivities` was never queried from this code. That much is unchanged.

  🔴 **CORRECTED — this bullet overstated the `targetUserId` risk in two ways, and the round-1 audit was right about both.**

  1. **The choice is not merely argued, it is settled in-repo with a measurement.** `apps/moderator/src/lib/server/user-signals.service.ts:33-34` records it: *"Use `targetUserId`, NOT `userId`: for Login and Registration rows `userId` is empty ~95% of the time (30M of 31.5M logins), so filtering on it silently finds nothing."* That is the same table, the same event type, and a counted result — not an inference from the writer.
  2. **A discriminator for the failure DOES exist, and this PR already emits it.** `evidence_registration_ips: 1` (the read ran) together with `evidence_distinct_registration_ips: 0` over a non-empty `cohort_size` is the wrong-column signature. The original sentence said the counter "cannot detect" it; what the *availability* counter alone cannot detect is a different thing, and the pair does detect it. The PR under-claimed its own coverage.

  What genuinely remained was the human-facing half: the summary said nothing when the read ran and matched nothing, so a moderator reading it could not tell a quiet day from a broken query. That is now a summary clause of its own, with a control that it stays quiet on an ordinary run.
- **No live run.** The job is still `UNRUNNABLE_JOB_CRON`; nothing has posted to the board.
- **False-positive rates are unmeasured for all three.** Named risks: NAT addresses on the clustering heuristic (no threshold removes this); ordinary phrases colliding after number-masking on the templating heuristic; the `COMMON_EMAIL_DOMAINS` list, which stays a hardcoded word list and is brittle in both directions — a missing provider is a standing false positive, and a listed one hides a real ring that used it.

  `domains_suppressed_common` **is now implemented** (it was asserted by a code comment while existing nowhere), so the first direction is measurable. Nothing measures the second.

  🔴 **One collision is no longer merely "named" — it is measured and pinned.** A generation-parameter paste (`Steps: …, Sampler: …, CFG scale: …, Seed: …`) fingerprints identically to the same line with different numbers, because every number is a `nummask`. That is one of the commonest comments on this site, so six such accounts in a day read as a ring of six and are reported. Both candidate fixes were rejected on arithmetic run over the shipped normaliser, not on taste: mask-token ratio **interleaves** the two classes in both directions (a paste including the prompt is 6/21 = 0.286, `check out linkmask for nummask free buzz` is 2/7 = 0.286, and `free buzz linkmask nummask` is 0.500 — above the longest paste's 0.389), and raising `CLUSTER_ZERO_AT` does not bound the innocent cluster, which is "people who commented their settings under a popular model today". So it is recorded as a regression case, stated in `similarity.ts`, and the `heuristic:<id>:sole_signal` counter — the findings a heuristic **carried on its own**, over the reported members — is what should set the fix. 🔴 **That counter's predicate changed in round 2** (`dfe494e4ba`): it used to be *exactly one heuristic scored above zero*, which excluded the collision's own routine shape — a member who also scores a TRACE somewhere else — and therefore under-reported the very population it was built for, in the reassuring direction. It is now a dominance test (`SOLE_SIGNAL_DOMINANCE`, see below).
- **`MIN_REPORTED_CONFIDENCE = 0.15` has never been applied to a real distribution.** It could be badly wrong in either direction.
- **The templating heuristic sees comments only** — model names, descriptions and image prompts are all templatable and none is read. A known false negative, not an oversight.
- **It sees a sample, not a census.** On a wave day the budget can run out and the oldest end of the cohort scores 0 for want of data.
- The no-write ledgers' stated blind spots are **narrowed but still open**: which *binding* is taken from a permitted specifier, what a permitted module does on its side, destructured handles, handles in object fields, computed method names, and the network. Optional chaining and a bracketed ClickHouse method key are no longer among them — `clickhouse` is typed `| undefined`, so `clickhouse?.$exec(…)` is the *idiomatic* spelling and both ledgers recorded nothing for it. That class is **not** closed: the patterns still match two handle names and a literal method, and they do not follow a client through an assignment, a destructure, an object field, a return value or a computed key. Those are dataflow, not pattern-matching.

---

Do not merge without a second read of the threshold default and the `COMMON_EMAIL_DOMAINS` list — those are the two judgement calls with the widest blast radius on a live board.

---

### `sole_signal` now measures what its name claims (`dfe494e4ba`)

The docstring said *"Nothing about such an account fires the other two heuristics, so it lands here and nowhere else"*, and the predicate was `exactly one heuristic scored above zero`. Those are different populations. A member 40 minutes old with 6 parameter-paste comments in a fingerprint cluster of 6 scores `posting-velocity` **0.1389** (6 items in 0.67h = 9/hour, just over the 4/hour floor) and `content-templating` **0.5**, blends to **0.2130**, clears `MIN_REPORTED_CONFIDENCE`, is **reported** — and incremented nothing. Reading `content-templating:sole_signal = 0` as "the collision produced no reports" was therefore wrong on the one number that decision was deferred to.

The predicate is now: the leading heuristic must score at least `SOLE_SIGNAL_DOMINANCE` (**3**) times every other. The multiple is derived, not picked — with `n` equally weighted heuristics the rest together contribute at most `(n-1)·s/k`, which is below `s` exactly when `k > n-1`, so at `n = 3` the smallest integer that works is 3. **It is tied to `n = 3`**: a fourth registered heuristic makes `k > 3` the condition, and the constant's docstring says so.

**No existing test had to change.** All three sole_signal pins survive, including the negative control at two heuristics agreeing at 0.9 — a tie fails the dominance test at any multiple above 1, so corroboration is excluded by arithmetic rather than by a special case. Two new cases were added: the 0.1389 + 0.5 collision (both scores **executed against the shipped heuristics**, not stipulated, so the case moves if `ZERO_AT_PER_HOUR` or `CLUSTER_ZERO_AT` moves), and a case pinning registry-**order** independence plus the nothing-fired guard.

Mutation results on the new predicate — 6 of 7 killed; the survivor (`>` → `>=` in the argmax tie-break) is **equivalent**, because on a tie the displaced member lands in `runnerUp` at the same value and dominance fails whichever tied member is named leader.

### `publicIpOnlySql` rejects a column that is not a bare identifier (`6fd1f13637`)

The precondition was a comment only, on a new shared export whose return value is concatenated into a ClickHouse statement by a client that does no escaping. Its own sibling `IP_PATTERN` in `apps/moderator/src/lib/server/clickhouse-filters.ts` exists for exactly this seam, so a regex here is the convention rather than a new idea. Both call sites pass the default `'ip'` today — it is unreachable, and that is why now is the cheap time. The test watches the throw and carries a negative control on accepted columns; all four regex mutants died.

### 🔴 Pre-cron check, owned by whoever schedules the job: the `time >=` literal's timezone

`registrationIpSql` compares a `DateTime` column to `'YYYY-MM-DD HH:MM:SS'` built from a UTC instant with no timezone qualifier. **This PR does not introduce the spelling** — `src/server/services/blocks/app-views.service.ts:129,226` already compares a `time DateTime` column to that identical literal format in production, and there are ≥5 in-repo precedents. It is a pruning predicate over accounts that cannot predate the window, so a skew removes rows the query would want rather than adding wrong ones; it is not a correctness hazard for the findings.

It was **not settled here** because it needs one query against an external vendor endpoint that was deliberately not used in this work. Run these two before the cron is enabled:

```sql
SELECT name, type FROM system.columns WHERE table = 'userActivities' AND name = 'time';
SELECT timezone();
```

**Closing condition:** the column's declared type carries no non-UTC timezone **and** `timezone()` is UTC — or, if either is false, the literal gains an explicit `toDateTime(…, 'UTC')`. Checked by whoever enables `UNRUNNABLE_JOB_CRON` for this job, before the first scheduled run.

### Report-summary headroom is smaller than it looks

Worst case — a 25,000-account cohort at the `MAX_COHORT_ACCOUNTS` cap with **every** degradation clause present at once (registration IPs unavailable, the read ran and matched nothing, content samples unavailable, the 5,000-row content budget exhausted, and the cohort truncated) — the summary measures **1,753 characters** against the schema's `z.string().max(2_000)` cap. That is **247 characters of headroom**, not the ~400 quoted earlier from a narrower assumption; measured by reproducing the concatenation in `run.ts` with those values. A third clause of the size of the ones added recently would push it over and **400 the whole report**, losing every finding in the batch rather than truncating the sentence. `reason` is truncated by `truncateReason`; `summary` is not.

### DIFF
--- a/apps/moderator/src/lib/server/clickhouse-filters.ts
+++ b/apps/moderator/src/lib/server/clickhouse-filters.ts
@@ -3,7 +3,11 @@
 // Declared once because a divergence here fails silently in the worst direction: a copy that misses an
 // updated internal range starts counting our own infrastructure as user signal, so a vote-ring panel
 // reports clusters made of Civitai traffic and nothing anywhere errors.
-export const INTERNAL_IP_RANGE = '10.124.0.0/16';
+//
+// 🔴 THE RANGE LIST NOW LIVES IN `@civitai/shared/clickhouse-ip-filters`, because the main Next app
+// reads the same table and cannot import anything from `apps/moderator`. It is re-exported here so
+// this app's existing importers are unchanged; adding a range in the shared module reaches both.
+export { INTERNAL_IP_RANGE } from '@civitai/shared/clickhouse-ip-filters';
 
 /** Interpolated into ClickHouse queries, which do NO escaping — every IP must match this first. */
 export const IP_PATTERN = /^[0-9a-fA-F.:]{3,45}$/;
--- a/apps/moderator/src/lib/server/reactor-lookup.service.ts
+++ b/apps/moderator/src/lib/server/reactor-lookup.service.ts
@@ -1,7 +1,8 @@
 import { sql } from '@civitai/db/kysely';
 import { dbRead } from './db';
 import { getClickhouse } from './clickhouse';
-import { INTERNAL_IP_RANGE, IP_PATTERN } from './clickhouse-filters';
+import { publicIpOnlySql } from '@civitai/shared/clickhouse-ip-filters';
+import { IP_PATTERN } from './clickhouse-filters';
 import { isInt4Id, usersByIds } from './users.service';
 import { strikeCountsByUserIds } from './moderation-memory.service';
 
@@ -59,26 +60,16 @@ const DELETE_TYPES = [
 const CREATES_ONLY = `type NOT IN (${DELETE_TYPES.map((t) => `'${t}'`).join(',')})`;
 
 /**
- * Private and carrier-internal space, which correlates everyone and therefore no one. `10.124.0.0/16`
- * is ours and already inside `10/8`; it stays named so the shared constant governs the rest.
+ * Private and carrier-internal space, which correlates everyone and therefore no one.
  *
- * 🔴 `ip != ''` is a guard, not a tidy-up. `isIPAddressInRange` RAISES on an empty string, and
- * `userActivities` holds a handful (3 in the last 30 days) — so whether this filter throws depends on
- * whether one lands in the range scanned. Order matters too: the guard has to be evaluated first.
+ * 🔴 THE LIST AND THE COMPOSITION MOVED TO `@civitai/shared/clickhouse-ip-filters` and this is now
+ * one call into it. The main Next app's bot-account detector reads the same `userActivities` rows
+ * for the same reason and cannot import from `apps/moderator`, so the alternative was a second
+ * hand-written copy — which is precisely the divergence `clickhouse-filters.ts` says fails silently.
+ * The emitted string is unchanged, including that `ip != ''` comes first: `isIPAddressInRange` RAISES
+ * on an empty string, and `userActivities` holds a handful.
  */
-const PUBLIC_IP_ONLY = [
-  `ip != ''`,
-  ...[
-    INTERNAL_IP_RANGE,
-    '10.0.0.0/8',
-    '172.16.0.0/12',
-    '192.168.0.0/16',
-    '127.0.0.0/8',
-    '169.254.0.0/16',
-    'fc00::/7',
-    '::1/128',
-  ].map((range) => `NOT isIPAddressInRange(ip, '${range}')`),
-].join(' AND ');
+const PUBLIC_IP_ONLY = publicIpOnlySql();
 
 const windowFilter = (days: number) => `time >= now() - INTERVAL ${Math.trunc(days)} DAY`;
 
--- a/packages/civitai-shared/package.json
+++ b/packages/civitai-shared/package.json
@@ -7,6 +7,7 @@
   "exports": {
     ".": "./src/index.ts",
     "./basemodel.constants": "./src/basemodel.constants.ts",
+    "./clickhouse-ip-filters": "./src/clickhouse-ip-filters.ts",
     "./lazy": "./src/lazy.ts",
     "./type-guards": "./src/type-guards.ts",
     "./moderator-paths": "./src/moderator-paths.ts"
--- a/packages/civitai-shared/src/clickhouse-ip-filters.ts
+++ b/packages/civitai-shared/src/clickhouse-ip-filters.ts
@@ -33,6 +33,16 @@ export const NON_PUBLIC_IP_RANGES: readonly string[] = [
 ];
 
 /**
+ * A bare ClickHouse column identifier — the only thing `publicIpOnlySql` will interpolate.
+ *
+ * Deliberately narrower than ClickHouse's own identifier grammar: no backtick-quoted forms, no dots,
+ * no digits in the leading position. Every caller today passes a plain column name, so the narrow
+ * form costs nothing, and widening it later is a decision someone makes on purpose rather than a
+ * gap someone finds.
+ */
+const SQL_COLUMN_IDENTIFIER = /^[A-Za-z_][A-Za-z0-9_]*$/;
+
+/**
  * `ip != '' AND NOT isIPAddressInRange(ip, '…') AND …` — the WHERE fragment.
  *
  * 🔴 `ip != ''` IS A GUARD, NOT A TIDY-UP, AND ITS POSITION IS PART OF IT. `isIPAddressInRange`
@@ -41,10 +51,22 @@ export const NON_PUBLIC_IP_RANGES: readonly string[] = [
  * breaks later, once, for one reader. It is emitted first for that reason, and a caller must not
  * reorder the conjunction.
  *
- * The column is a parameter only so a reader whose column is not called `ip` can use it; it is
- * interpolated, so it must be an identifier the caller controls and never user input.
+ * The column is a parameter only so a reader whose column is not called `ip` can use it.
+ *
+ * 🔴 IT IS INTERPOLATED, AND A COMMENT IS NOT A PRECONDITION. This module's own sibling —
+ * `apps/moderator/src/lib/server/clickhouse-filters.ts`'s `IP_PATTERN` — exists because ClickHouse
+ * queries here are built by string concatenation and the client does NO escaping, so a regex at the
+ * seam is this codebase's convention at exactly this point. Both call sites pass the default today,
+ * which makes this unreachable and makes NOW the cheap time to close it: the guard is being added
+ * before the parameter spreads, not after a caller has already reached it with something it built.
+ * Throwing is the only safe direction — a shadow-mode detector that silently issues a mangled
+ * statement produces a zero that reads as "no ring found".
  */
 export function publicIpOnlySql(column = 'ip'): string {
+  if (!SQL_COLUMN_IDENTIFIER.test(column))
+    throw new Error(
+      `publicIpOnlySql: column must be a bare SQL identifier, got ${JSON.stringify(column)}`
+    );
   return [
     `${column} != ''`,
     ...NON_PUBLIC_IP_RANGES.map((range) => `NOT isIPAddressInRange(${column}, '${range}')`),
--- a/packages/civitai-shared/src/clickhouse-ip-filters.ts
+++ b/packages/civitai-shared/src/clickhouse-ip-filters.ts
@@ -1,0 +1,52 @@
+// The address space that carries no per-account signal, and the predicate that excludes it from any
+// ClickHouse read of a request-level event table (`userActivities` and friends).
+//
+// 🔴 DECLARED ONCE BECAUSE A DIVERGENCE HERE FAILS SILENTLY IN THE WORST DIRECTION. Private and
+// carrier-internal space correlates everyone and therefore no one: a query that omits this filter
+// does not error, it returns a large, confident cluster made of our own infrastructure or of one
+// mobile carrier's NAT, and every consumer reads that as evidence. `apps/moderator`'s reactor-lookup
+// panel carried the only copy of this predicate; a second copy written by hand for a second reader is
+// how one of them silently stops matching the other.
+//
+// Client-safe: string constants and string building only, no client, no env, no IO.
+
+/** Our own infrastructure range. Already inside `10/8`; kept named because several moderator
+ *  queries exclude it on its own without the rest of the list. */
+export const INTERNAL_IP_RANGE = '10.124.0.0/16';
+
+/**
+ * Every range whose members must not be treated as a user-identifying address.
+ *
+ * `INTERNAL_IP_RANGE` first, then RFC1918, loopback, link-local and the IPv6 equivalents. The list is
+ * over-inclusive on purpose — an address wrongly excluded costs one account's IP signal, an address
+ * wrongly INCLUDED manufactures a ring out of everyone behind it.
+ */
+export const NON_PUBLIC_IP_RANGES: readonly string[] = [
+  INTERNAL_IP_RANGE,
+  '10.0.0.0/8',
+  '172.16.0.0/12',
+  '192.168.0.0/16',
+  '127.0.0.0/8',
+  '169.254.0.0/16',
+  'fc00::/7',
+  '::1/128',
+];
+
+/**
+ * `ip != '' AND NOT isIPAddressInRange(ip, '…') AND …` — the WHERE fragment.
+ *
+ * 🔴 `ip != ''` IS A GUARD, NOT A TIDY-UP, AND ITS POSITION IS PART OF IT. `isIPAddressInRange`
+ * RAISES on an empty string and `userActivities` holds a handful of them, so whether the query
+ * throws depends on whether a blank row lands in the range being scanned — it passes in testing and
+ * breaks later, once, for one reader. It is emitted first for that reason, and a caller must not
+ * reorder the conjunction.
+ *
+ * The column is a parameter only so a reader whose column is not called `ip` can use it; it is
+ * interpolated, so it must be an identifier the caller controls and never user input.
+ */
+export function publicIpOnlySql(column = 'ip'): string {
+  return [
+    `${column} != ''`,
+    ...NON_PUBLIC_IP_RANGES.map((range) => `NOT isIPAddressInRange(${column}, '${range}')`),
+  ].join(' AND ');
+}
--- a/src/server/jobs/bot-account-detection.ts
+++ b/src/server/jobs/bot-account-detection.ts
@@ -1,5 +1,6 @@
 import { logToAxiom } from '~/server/logging/client';
 import { createCohortReader } from '~/server/services/bot-account-detection/cohort';
+import { createEvidenceReader } from '~/server/services/bot-account-detection/evidence';
 import { runBotAccountDetection } from '~/server/services/bot-account-detection/run';
 import { moderatorApp } from '~/server/services/moderator-app.service';
 import { createJob, UNRUNNABLE_JOB_CRON } from './job';
@@ -48,6 +49,11 @@ export const botAccountDetection = createJob(
   async (ctx) => {
     return runBotAccountDetection({
       reader: createCohortReader(),
+      // The cohort-level sources. Its ClickHouse half is `undefined` on any deployment without
+      // `CLICKHOUSE_HOST`, which is a supported state and NOT an error: the run reports
+      // `evidence_registration_ips: 0` and says so in the summary rather than reporting that no
+      // accounts shared an IP.
+      evidence: createEvidenceReader(),
       sendReport: (report) => moderatorApp.abuseReport(report),
       now: () => new Date(),
       // The scheduler cancels by closing the response; without this the walk keeps paging and
--- a/src/server/services/bot-account-detection/__tests__/cohort.test.ts
+++ b/src/server/services/bot-account-detection/__tests__/cohort.test.ts
@@ -14,6 +14,7 @@ import {
   mergePostCounts,
   modelCountArgs,
   newAccountPageArgs,
+  normalizeEmailDomain,
   selectCohortMembers,
   type CohortDb,
   type CohortReader,
@@ -29,6 +30,10 @@ const account = (id: number, createdAt = NOW): NewAccountRow => ({
   id,
   username: `user${id}`,
   createdAt,
+  // Per-account domain by default, so a fixture never accidentally builds a cluster the test did
+  // not ask for — the clustering heuristic counts members sharing a domain, and a shared fixture
+  // domain would silently make every paging test a ring.
+  email: `user${id}@user${id}.test`,
 });
 
 const surface = (partial: Partial<SurfaceCounts> = {}): SurfaceCounts => {
@@ -158,8 +163,12 @@ describe('newAccountPageArgs', () => {
     expect(seed.orderBy).toEqual({ id: 'desc' });
   });
 
-  it('selects only the columns the report cites', () => {
-    expect(args.select).toEqual({ id: true, username: true, createdAt: true });
+  it('selects only the columns the report cites, plus the email the clustering heuristic needs', () => {
+    // 🔴 `email` is here so the registration-cluster heuristic gets a domain WITHOUT a second
+    // query — widening a `select` adds no database operation, which is what keeps the asserted
+    // operation ledger in `no-write-surface.test.ts` at five reads. The address itself does not
+    // survive `selectCohortMembers`; see the `emailDomain` cases below.
+    expect(args.select).toEqual({ id: true, username: true, createdAt: true, email: true });
   });
 });
 
@@ -517,9 +526,12 @@ describe('selectCohortMembers', () => {
   it('carries the account identity and its activity through', () => {
     const created = at('2026-09-03T09:30:00.000Z');
     const members = selectCohortMembers(
-      [{ id: 5, username: 'spam5', createdAt: created }],
+      [{ id: 5, username: 'spam5', createdAt: created, email: 'spam5@Ring.Example' }],
       counts([[5, { comments: 2, models: 1, images: 3 }]])
     );
+    // 🔴 A WHOLE-OBJECT EQUALITY, not a `toMatchObject`. That is what makes the next assertion
+    // meaningful: `toEqual` fails if the member grows a field, so the email address CANNOT
+    // reappear on this object without a failing test.
     expect(members[0]).toEqual({
       userId: 5,
       username: 'spam5',
@@ -529,7 +541,56 @@ describe('selectCohortMembers', () => {
         visible: { comments: 2, models: 1, images: 3, total: 6 },
         excluded: { comments: 0, models: 0, images: 0, total: 0 },
       },
+      // Lowercased, and the local part is gone.
+      emailDomain: 'ring.example',
     });
+  });
+
+  it('🔴 reduces the email to its domain and keeps the address nowhere', () => {
+    // The reduction happens in the pure selector rather than inside the heuristic that consumes it,
+    // so nothing downstream — no score, counter, note or reason string — can hold an address. This
+    // asserts the boundary rather than the heuristic's use of it.
+    const [only] = selectCohortMembers(
+      [
+        {
+          id: 7,
+          username: 'u7',
+          createdAt: at('2026-09-03T09:30:00.000Z'),
+          email: 'someone@a.test',
+        },
+      ],
+      counts([[7, { comments: 1 }]])
+    );
+    expect(only.emailDomain).toBe('a.test');
+    expect(JSON.stringify(only)).not.toContain('someone');
+  });
+});
+
+describe('normalizeEmailDomain', () => {
+  it('lowercases and trims a real domain', () => {
+    expect(normalizeEmailDomain('User@Example.COM')).toBe('example.com');
+    expect(normalizeEmailDomain('u@ example.com ')).toBe('example.com');
+  });
+
+  it('splits on the LAST @, because a local part may legally contain one', () => {
+    // A domain cannot contain `@`; a quoted local part can. Splitting on the first would yield a
+    // cluster key with an `@` in it, and two differently-quoted addresses at the same real domain
+    // would then fail to cluster.
+    expect(normalizeEmailDomain('"odd@name"@real.example')).toBe('real.example');
+  });
+
+  it('🔴 returns null rather than a key for anything that is not a domain', () => {
+    // The return value becomes a CLUSTER KEY, and two accounts are called related when their keys
+    // are equal — so a permissive parse maps several malformed addresses onto one key and
+    // MANUFACTURES a ring out of bad data. Every rejection here is a ring that cannot be invented.
+    expect(normalizeEmailDomain(null)).toBeNull();
+    expect(normalizeEmailDomain('')).toBeNull();
+    expect(normalizeEmailDomain('no-at-sign')).toBeNull();
+    expect(normalizeEmailDomain('trailing@')).toBeNull();
+    // No dot: not a domain a registration can have come from.
+    expect(normalizeEmailDomain('u@localhost')).toBeNull();
+    // Internal whitespace: not a domain at all.
+    expect(normalizeEmailDomain('u@bad domain.com')).toBeNull();
   });
 });
 
--- a/src/server/services/bot-account-detection/__tests__/evidence.test.ts
+++ b/src/server/services/bot-account-detection/__tests__/evidence.test.ts
@@ -1,4 +1,4 @@
-import { NON_PUBLIC_IP_RANGES } from '@civitai/shared/clickhouse-ip-filters';
+import { NON_PUBLIC_IP_RANGES, publicIpOnlySql } from '@civitai/shared/clickhouse-ip-filters';
 import { describe, expect, it, vi } from 'vitest';
 import type { BotAccountCohortMember, SurfaceCounts } from '../cohort';
 import {
@@ -308,6 +308,28 @@ describe('registrationIpSql', () => {
 
     // Optional: a caller with no window still gets a correct, merely unpruned, answer.
     expect(registrationIpSql([1])).not.toContain('time >=');
+  });
+});
+
+describe('publicIpOnlySql', () => {
+  it('🔴 THROWS on a column that is not a bare identifier, rather than interpolating it', () => {
+    // 🔴 THE PRECONDITION WAS A COMMENT ONLY, ON A NEW SHARED EXPORT. The value is concatenated into
+    // a ClickHouse statement and the client does no escaping — the sibling `IP_PATTERN` in
+    // `apps/moderator/src/lib/server/clickhouse-filters.ts` exists for exactly that reason. Both
+    // call sites pass the default today, so this is closed BEFORE the parameter spreads.
+    //
+    // Watching the throw is the point: a guard nobody has seen go red is a claim about the regex.
+    expect(() => publicIpOnlySql("ip' OR 1=1 --")).toThrow(/bare SQL identifier/);
+    expect(() => publicIpOnlySql('ip)')).toThrow(/bare SQL identifier/);
+    expect(() => publicIpOnlySql('remote.ip')).toThrow(/bare SQL identifier/);
+    expect(() => publicIpOnlySql('')).toThrow(/bare SQL identifier/);
+    expect(() => publicIpOnlySql('1ip')).toThrow(/bare SQL identifier/);
+
+    // The negative control on the guard itself: a legitimate column must still get through, or the
+    // regex could be `/^$/` and every case above would pass for the wrong reason.
+    expect(publicIpOnlySql()).toContain(`ip != ''`);
+    expect(publicIpOnlySql('clientIp')).toContain(`clientIp != ''`);
+    expect(publicIpOnlySql('_ip2')).toContain(`_ip2 != ''`);
   });
 });
 
--- a/src/server/services/bot-account-detection/__tests__/evidence.test.ts
+++ b/src/server/services/bot-account-detection/__tests__/evidence.test.ts
@@ -1,10 +1,16 @@
+import { NON_PUBLIC_IP_RANGES } from '@civitai/shared/clickhouse-ip-filters';
 import { describe, expect, it, vi } from 'vitest';
 import type { BotAccountCohortMember, SurfaceCounts } from '../cohort';
 import {
+  MAX_CONTENT_CHARS,
+  MAX_CONTENT_SAMPLES,
+  MAX_IPS_PER_ACCOUNT,
+  EVIDENCE_CHUNK_SIZE,
   MIN_FINGERPRINT_CHARS,
   MIN_FINGERPRINT_TOKENS,
   buildCohortSignals,
   chunk,
+  clickhouseDateTime,
   collectCohortSignals,
   contentFingerprint,
   contentSampleArgs,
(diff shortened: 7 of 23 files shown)


ZacxDev:
### CI attribution — independently verified, and one finding that is not this PR's

The flake attribution in the PR body is correct, and I checked its strongest leg rather than accepting it. `Unit tests (3)` and `(4)` fail on **four consecutive `main` commits**, including this PR's exact base `8d7faf613`:

```
8d7faf613  (1)=success (2)=success (3)=failure (4)=failure   <- this PR's base
d2046551e  (1)=success (2)=success (3)=failure (4)=failure
ffa0f3b41  (1)=success (2)=success (3)=failure (4)=failure
d70946864  (1)=success (2)=success (3)=failure (4)=failure
```

So those two shards are **pre-existing red on `main`**, not introduced here. The separate `Unit tests (1)` failure reported earlier on `app-spend-cap.service.test.ts` has since gone green without a code change, which is consistent with the wall-clock-bucket flake the author diagnosed (`Math.floor(Date.now()/1000/60)` rolling over mid-test).

🔴 **Separate finding, not this PR's to fix, but worth someone's attention: half the unit matrix has been red on `main` for at least four commits.** A permanently-red gate is worse than no gate — it trains every author to click through, and it means shards 3 and 4 currently provide no signal for any PR in this repository. Whoever owns that suite should either unbreak it or stop gating on it.



---

## 🔴 Correction to the paragraph above — I named the wrong mechanism

I wrote that shards 3 and 4 are red because the *suite* is broken, and that someone should "unbreak it or stop gating on it". The observation that they are pre-existing red is correct; **the diagnosis is not**, and the remedy I implied would have been wrong.

They are not test failures at all. They fail `scripts/ci/assert-shard-ran.mjs`, a **size guard**, on its `BASE_TESTS = 22157` constant — set 2026-08-25 in #4392 when the suite was sharded. The suite has since grown to roughly **37,969** executed tests across the four shards, so the derived ceiling trips. Shard 3 was over by ~4,753 and shard 4 by ~216 **before** this PR, which adds +10 and +15.

The guard's own docstring predicted precisely this: *"ordinary suite growth was on course to trip the ceiling on all four shards within weeks"*, and *"BASE_TESTS is the whole suite's executed count, and it is the one number here that goes [stale]"*. By the guard's own discriminator — counts differing widely while shards 1 and 2 pass — this is suite growth, not a broken `--shard`.

So the fix is mechanical and small: re-derive `BASE_TESTS` to the current executed count and confirm all four shards exit 0. It is a CI-config change and deliberately does not belong on a moderation PR.

Recording the correction rather than editing the original away, since anyone who read it would have taken "the unit suite is broken" from it — which is not what is happening.

ZacxDev:
## CI status: 16 pass, 2 fail, 1 skip — both failures are pre-existing on `main`

Recording the measurements so the red checks aren't read as this PR's.

### `Unit tests (3)` and `(4)` — the shard-size guard, already red at this PR's base

Neither is a test failure. Both are `scripts/ci/assert-shard-ran.mjs` tripping its ceiling, and both were **already failing on `main` at `8d7faf613`**, which is exactly this branch's base:

| shard | `main` @ `8d7faf613` | this PR | ceiling | already over by, on `main` |
|---|---|---|---|---|
| 3 | 14,724 | 14,734 (+10) | 9,971 | **+4,753** |
| 4 | 10,187 | 10,202 (+15) | 9,971 | **+216** |

This PR contributes +10 and +15 tests to those shards. The gate was breached by 4,753 and 216 before it existed.

**Cause, by the guard's own discriminator.** Its message names two causes and says: *"if all of them tripped and each took ~the full-suite runtime, it is (1) `--shard` is not reaching vitest. If the counts look like a proportionate share, it is (2)."* Shards 1 and 2 pass and the four counts differ widely (7,188 / 5,845 / 14,734 / 10,202), so no shard is running the whole suite — it is **(2): the suite has grown past `BASE_TESTS`.**

Measured across all four shards on this run: **37,969 executed**, against `BASE_TESTS = 22157` (measured 2026-08-25). The constant is stale by roughly 15,800 tests; this PR's 112 are not the cause.

**Closing condition, if someone wants to fix it separately:** update `BASE_TESTS` in `scripts/ci/assert-shard-ran.mjs` to the re-derived sum (~37,969 as of this run) and confirm all four shards exit 0. Not done here — it is unrelated to this change and would be a CI-config edit riding on a moderation PR.

### `Unit tests (1)` — a wall-clock flake, now green on re-run

First attempt failed on `src/server/services/blocks/__tests__/app-spend-cap.service.test.ts > reserveAppSpend — VELOCITY`, with `expected true to be false`. It passes on attempt 2.

That test buckets on real wall-clock time:

```js
const bucket = Math.floor(Date.now() / 1000 / BLOCK_APP_SPEND_VELOCITY_WINDOW_SECONDS); // 60s
```

If the 60-second bucket rolls over mid-test the counter resets and the ceiling is never breached — producing exactly that assertion. It also passed 3/3 locally on this branch, and nothing outside the detector imports this PR's code (only `src/server/jobs/bot-account-detection.ts` and the run-jobs route do), so there is no path from this change to `services/blocks/`.

### This PR's own tests ran and passed in CI

Confirmed rather than assumed — a green shard proves nothing if the new files were never collected:

```
✓ src/server/services/bot-account-detection/__tests__/report.test.ts (27 tests)
✓ src/server/services/bot-account-detection/__tests__/run.test.ts   (30 tests)
```

ZacxDev:

Head `97b2fd8399`, base `8d7faf6138`. **No 🔴. 9 🟡, 4 🟢.** Verdict: **safe to merge as it stands; not safe to schedule.**

That split is structural, not a judgement about the heuristics — `UNRUNNABLE_JOB_CRON` plus a hand-enumerated read-only surface means merging changes nothing observable until a human edits a cron string.

**Shadow property re-confirmed independently of the ledgers.** The module's entire effect surface: `user.findMany`; eight `groupBy`s; `comment.findMany` + `commentV2.findMany`; one `ch.$query` carrying a bare `SELECT … GROUP BY … LIMIT`; one `sendReport`. No `dbWrite`, no raw SQL, no `$transaction`, no `fetch`, no restriction/mute/ban import anywhere. `actioned: false` is a literal, not a parameter.

### The two findings that matter

**🟡 F2 — nothing holds the seam between evidence and scoring; two one-line mutants leave 243/243 green.** `run.ts:203` — replacing the real `signals` with `emptyCohortSignals()` passes the entire suite. **Two of the three heuristics would score 0 for every account on every production run, while `evidence_registration_ips`, `evidence_distinct_registration_ips` and `evidence_distinct_content_fingerprints` all still report healthy values**, because they read `signals` rather than what scoring saw. Two-thirds of what this PR ships would be inert and every counter would say it is fine. Deleting `evidence: createEvidenceReader()` from the job also passes. `evidence.test.ts` covers the index, `heuristics.test.ts` covers pure scorers over hand-built signals, `run.test.ts` covers the threshold — and **no test ever builds the combined state.** (For contrast: unregistering either heuristic from the registry is killed by 4 tests each.)

**🟡 F1 — the ClickHouse read does not exclude private/carrier-internal IPs, unlike both sibling readers of the same table.** `reactor-lookup.service.ts:60-80` carries `PUBLIC_IP_ONLY` with the comment *"Private and carrier-internal space, which correlates everyone and therefore no one"* — this heuristic's failure mode, named. At 6 cohort members on an infrastructure address the IP sub-score reaches 0.5 and the finding is reported; at ≥10 it tops the run's distribution entirely from a proxy or `10.124/16`.

### 🟡 F3 — the SELECT projection is unguarded, and getting it wrong is exactly the silent zero
Mutating `SELECT targetUserId, ip` → `SELECT userId, ip` (leaving the WHERE alone) leaves **243 green**. Rows come back keyed `userId` → `Number(undefined)` = `NaN` → dropped → zero rows, no throw, `ipsRead` stays true, no warning. Existing tests pin the WHERE clause and the verb, never the projection.

### 🟡 F4 — my question answered: the discriminator DOES exist, and the column choice was already settled in-repo
I asked whether any counter separates "no ring found" from "query wrong". **It does** — `evidence_distinct_registration_ips: 0` with `evidence_registration_ips: 1` and `cohort_size > 0` is the wrong-column signature. The PR under-claims its own coverage. And the `targetUserId` choice is corroborated with a *measurement* already in this repo: `user-signals.service.ts:33-34` — *"Use `targetUserId`, NOT `userId`: for Login and Registration rows `userId` is empty ~95% of the time (30M of 31.5M logins), so filtering on it silently finds nothing."* **The author's headline risk is materially lower than the PR body claims.** Residual: the human-facing summary says nothing when the read ran and matched nothing.

### 🟡 F5 — `content-templating` collides on this site's commonest long comment shape
Executed against the shipped normaliser: `Steps: 20, Sampler: Euler a, CFG scale: 7, Seed: 1234567890, Size: 512x768` and the same line with different numbers produce an **identical** fingerprint. A generation-parameter paste is one of the most ordinary comments on civitai; six such accounts in a day → reported as a ring, quoting the shared text. The `MIN_FINGERPRINT_CHARS`/`TOKENS` floors defend only against *short* text and cannot help here by construction.

### 🟡 F6 — the domain list is Anglophone-biased in the false-positive direction, and walkable in one keystroke
`hotmail.com`/`.co.uk` and `yahoo.com`/`.co.uk` are listed; `hotmail.fr/de/es/it`, `foxmail.com`, `yandex`, `web.de`, `free.fr`, `libero.it`, `uol.com.br`, `daum.net`, `rediffmail.com` and ~25 more are not. Nine new posting accounts a day on `hotmail.fr` is ordinary, so the standing false positive lands systematically on non-English-speaking users. Sharpest case: **`proton.me` and `protonmail.com` are both listed, but `pm.me` — Proton's own short alias, offered to every paid account — is not.** The evasion is one keystroke: register the ring on `gmail.com`.

### 🟡 F7 — a code comment asserts a counter that does not exist
`clustering.ts:75` claims `domains_suppressed_common` "is what makes the first measurable". The identifier appears nowhere in `src/`, `packages/` or `apps/` — verified with a positive control. The PR body discloses this; the code does not, and the code is what the next maintainer reads.

### 🟡 F8 — a failing content read kills the whole run; a failing IP read only degrades it
The IP loop has try/catch, a degrade path and a source flag. `listContentSamples` has none, so a timeout on a busy replica propagates out and **no report is filed at all** — losing the velocity heuristic's day too, and looking exactly like a dead producer. Contradicts the module's own stated philosophy at `evidence.ts:399`.

### 🟡 F9 — no time predicate on a daily 50-chunk ClickHouse scan
Sibling code treats `time` as the pruning column. Every cohort account is younger than the window, so `AND time >= '<createdAfter>'` is semantically free and is the difference between 50 pruned reads and 50 unpruned ones.

### 🟢 F10–F13
Four ledger escapes verified by *executing* the shipped regexes (`ch?.$exec`, `ch['$exec']`, `dbWrite?.user.update`, an interpolated first token) — optional chaining is the idiomatic spelling for this client, since `clickhouse` is typed `| undefined`. Distribution counters duplicated into every batch, so the buckets sum to k×`cohort_size` across board rows. `LIMIT ids.length * 4` is a per-*result* cap, not the per-account cap its comment claims, and truncation is silent. Four surviving mutants on untested constants, including the entire budget-exhausted summary sentence.

### Claims re-verified
243/8 at HEAD and 131/6 at base both reproduce; all four behavioural base-reds reproduce by name; the labelled invariant guard does pass at base, correctly not counted; the bucket-sum assertion is **real, not vacuous**; Prettier clean with a positive control. Typecheck: my absolute count was 721 not 758 (different `node_modules` resolution) — but **the claim was equality with base, and the error sets are byte-identical**. The 35-mutant table was not reproduced as such; the auditor ran **62 of its own**, killing 55, with 2 provable equivalents and the other 5 being F2/F3/F13 — each sandbox given a fresh `node_modules/.vite` so no mutant could be served from a stale transform cache.

### What the threshold actually admits
Blend is the mean of three, so one signal must reach 0.45: **6 items within 15 minutes of signup** (velocity's cheapest crossing, because `MIN_AGE_HOURS` floors the divisor), **6** members on one IP, **9** on one uncommon domain, **6** on one fingerprint. Worth stating because the PR's NAT reassurance cites *ten* — that is the top of the ramp, not where reporting starts. Day one's reported population will be dominated by newcomers doing a normal first-session batch upload.

**Blocks a real cron:** F1, F2, F3. **Blocks merge:** none.

ZacxDev:

Commit `409f9148be`, 15 files. Payload this round: **489** (plus 778 scaffolding).

```audit-claims round=1 audited=97b2fd8399..409f9148
1. F2 — `run.test.ts` drives a six-account ring through the PRODUCTION registry with a real EvidenceReader and asserts the EMITTED finding's own sub-scores, parsed from the `Per-heuristic:` clause, with a no-evidence control arm beside it.
2. F2 — the job-wiring half is asserted on the job SOURCE, because `evidence?` is deliberately optional so deleting the line was equally green.
3. F1 — `PUBLIC_IP_ONLY` was SHARED, not copied: moved into a new `packages/civitai-shared/src/clickhouse-ip-filters.ts` consumed by both trees, with `INTERNAL_IP_RANGE` re-exported so the moderator's three other importers are untouched. The moderator's emitted SQL is claimed byte-unchanged and its 23 tests still pass.
4. F3 — the SELECT projection is pinned, with the GROUP BY alongside it.
5. F8 — content-read try/catch, `sources.contentSamples`, a counter, and a summary sentence distinct from an exhausted budget.
6. F4 — a summary clause for the read-ran-and-matched-nothing case; F7 — `domains_suppressed_common` now exists; F9 — `AND time >= …`; F12 — `LIMIT 4 BY targetUserId` plus `ORDER BY`; F13 — six constants pinned and the whole budget sentence asserted.
7. F10 — both ledger regexes widened for `?.` plus a bracket key for ClickHouse, each with its own control AND its own negative control.
8. 21 mutants planted, 21 killed, fresh `node_modules/.vite` per mutant. The audit's own three are among them, each killed by the assertion written for it.
9. Ten new cases are genuine regression tests; SIX are invariant guards that pass at base and are LABELLED as such rather than counted as coverage.
10. F5 — deliberately NOT patched, on measured arithmetic; F6 — extended to ~120 entries and labelled a patch, with a negative control that the predicate was not widened to `() => true`.
```

**F5 is a refusal backed by measurement, not taste.** Both remedies I proposed were tested against the shipped normaliser and both fail: mask-token ratio does not separate the classes, it **interleaves** them — a parameter paste including the prompt scores 6/21 and `check out linkmask for nummask free buzz` scores 2/7, *the same 0.286*, while `free buzz linkmask nummask` is 0.500, above the longest paste's 0.389. No cut exists in either direction. Raising `CLUSTER_ZERO_AT` cannot bound the innocent cluster either. So the collision is pinned as a regression case, stated in full in `similarity.ts`, and a new `heuristic:<id>:sole_signal` counter was added over the *reported* members — the population a collision inflates, and the one `fired` cannot see. That counter is meant to set the fix.

🔴 **Self-caught, and worth recording:** the first positive control was mutating `BOT_ACCOUNT_DETECTOR`, and it **survived** — because `run.test.ts` asserts against the same imported constant, so the mutant moves both sides of the comparison. An equivalent mutant, discarded rather than reported as a finding. That is the vacuous-assertion shape catching itself.

**PR body corrected:** the `targetUserId` risk was overstated twice — the choice is settled by a counted measurement at `user-signals.service.ts:33-34`, and the discriminator already existed as the `evidence_registration_ips: 1` / `evidence_distinct_registration_ips: 0` pair.

**Not verified:** the ClickHouse statement has still never executed, so `LIMIT n BY` after `GROUP BY`/`ORDER BY` rests on in-repo precedent (`~/server/redis/caches.ts:1975`, live code, same shape), not a live parse — the mocked `$query` never parses the string. CI shards 3/4 remain pre-existing red on the stale `BASE_TESTS` constant, confirmed identical at base.

ZacxDev:

Range `97b2fd8399..409f9148be`, one commit. **No 🔴. 2 🟡, 1 🟢.** Verdict: **safe to merge**; one finding to settle before a real cron.

### The two claims I asked to be verified rather than accepted — both hold

**The byte-unchanged claim was checked directly, not via "its tests pass".** The base commit's own `PUBLIC_IP_ONLY` expression text (pulled with `git show 97b2fd8399:…`) was executed against the shipped `publicIpOnlySql()`: **both 378 chars, `===` true**, with a negative control (`publicIpOnlySql('addr')` → false). `INTERNAL_IP_RANGE` identical on both sides. `reactor-lookup.test.ts` 23 passed. A shared-code extraction touching a live moderator surface, verified as identical rather than merely green.

**F5's arithmetic holds, so the refusal stands and is not re-litigated.** Executing the shipped normaliser: a parameter paste with prompt is **6/21 = 0.2857** and `check out <link> for <num> free buzz` is **2/7 = 0.2857** — *exactly equal* — while `free buzz <link> <num>` is **0.5000**, above the longest paste's **0.3889**. The ordering inverts. No threshold separates the classes in either direction. Declining to patch was the right call.

Also re-verified: the seam mutant (`signals: emptyCohortSignals()`) now kills **the seam test and nothing else** — 1 failed / 270 passed. Both ledger regexes reverted to base text → exactly the 2 new cases go red. And the base-failure split was **understated**: 20 cases red at base, and all 9 that pass carry explicit labels ("THIS IS NOT A GUARD, IT IS A RECORD", "THE MUTANT THIS EXISTS FOR"). None is a mislabelled regression test.

### 🟡 1 — `sole_signal` does not count the population it was created to measure
`scoring.ts:194-196`. F5's refusal defers the fix to this counter, whose docstring asserts *"Nothing about such an account fires the other two heuristics, so it lands here and nowhere else."* False for a routine case, because the filter is `score > 0` and a **trace** score disqualifies the finding.

Worked from shipped constants: a member 40 min old with 6 parameter-paste comments, fingerprint cluster of 6 → velocity `rampScore(9.0, 4, 40)` = **0.1389**, templating **0.5**, blend **0.2130** → reported. `fired.length === 2`, so it increments **no** `sole_signal` key. An operator reading `content-templating:sole_signal = 0` concludes the known collision produced no reports this run. **The bias is toward the reassuring direction, on the one number the F5 decision was deferred to.**

Follow-up; blocks *acting on the counter*, not merge. Closing condition: correct the docstring to what it measures, or change the predicate to a dominance test — with a unit case at velocity 0.1389 + templating 0.5, red first.

### 🟡 2 — the `time >=` literal carries no timezone (flagged, **not** asserted)
`evidence.ts:193-212`. `YYYY-MM-DD hh:mm:ss` with no zone; ClickHouse resolves it in the column's declared timezone or the server's. If `userActivities.time` is a bare `DateTime` and the server is west of UTC, the bound lands hours late and the oldest slice of the window is dropped — silently, with `sources.registrationIps` still true and cluster sizes understated. Understating is the direction that produces the confident zero this module is written against.

🔴 **But this is not introduced here.** `src/server/services/blocks/app-views.service.ts:129,226` already compares a `time DateTime` column to that exact literal format in production, and the `toISOString().slice(0,19)` spelling has ≥5 in-repo precedents. If the assumption is wrong it is wrong repo-wide today. The second-truncation errs *loosely* (bound moves earlier), which is the safe direction.

**One query settles it, and it needs ClickHouse Cloud access I deliberately did not use** — reaching an external vendor endpoint with production credentials from a workstation is a different class of action from the in-cluster reads elsewhere in this review:
```sql
SELECT name, type FROM system.columns WHERE database='default' AND table='userActivities' AND name='time';
SELECT timezone();
```
If the type is a bare `DateTime` and `timezone()` is not UTC, wrap as `toDateTime('…','UTC')`.

### 🟢 3 — `publicIpOnlySql(column)` interpolates its parameter with no identifier guard
`clickhouse-ip-filters.ts:51-56`. The precondition is a comment only, while its own sibling `clickhouse-filters.ts:14` exists precisely because *"Interpolated into ClickHouse queries, which do NO escaping"* — the codebase's convention at this exact seam is a regex guard. Both current call sites use the default `'ip'`, so it is unreachable today; it is hardening on a **new, shared, exported** API. Closing condition: reject a column not matching `/^[A-Za-z_][A-Za-z0-9_]*$/`, with a test that watches the throw.

### Checked and deliberately not filed
The `Per-heuristic:` parse is anchored on a following literal, non-greedy, asserts all three ids first; `toFixed(2)` is not locale-sensitive; ids are exact map keys. The widened `SEGMENT` over-matches only text that is not valid JS. `LIMIT 4 BY targetUserId` matches ClickHouse clause order and the in-repo precedent at `caches.ts:1966-1975` — **neither is a live parse**. Worst-case summary is **1596 chars** against the schema's 2,000 cap; headroom is ~400 and the two clauses added this round are 230–300 each, so a *third* would 400 the whole report. Worth knowing.

### Regression control for the range
Full `unit*` at HEAD: 334 failing files, all `PrismaClientInitializationError` / missing-package from a hand-built `node_modules`. Control run with the range's payload reverted to base: **338** failing files, and `comm` of the two sets is **empty in the HEAD direction** — the only difference is the four detector test files going red when the payload is reverted. **Zero new failures attributable to this round.**

### Ledger
`round 2 · payload lines changed THIS round: 489 (since round 1: 489)`. Honest qualifier from the auditor: of those 489, **247 are code or data and 242 are comment prose**, and ~130 of the 247 are `COMMON_EMAIL_DOMAINS` entries — so the logic payload is roughly 117 lines. Not a zero-payload round; the gate does not fire.

PR #4644, opened 2026-09-05
fix(blocks): stop an identical chat ask being answered from a stored reply at full price

ZacxDev:
## What

`buildStep` for the `chat-completion` block step now emits a fresh random `seed` per submit.

## Why

The platform serves a step's result from storage when a later step arrives with a **byte-identical input**, skipping execution entirely. That is deliberate, and it is sound for a step whose output is a pure function of its input — a seeded image generation asked for the same picture twice.

A chat completion at `temperature > 0` is not such a step. The same input has many correct answers, so reusing one is not "the same result", it is **last week's result**.

`buildChatCompletionInput` emitted only semantic fields — `model`, `messages`, `maxTokens`, `temperature`, `tools` — so two identical asks hashed identically **by construction**, and a block had no field with which to ask for a fresh draw.

Measured in production before this change:

- nearly **half** of tool-using turns had at least one round served from storage
- the longest gap was a **whole answer more than twenty hours old**, replayed verbatim
- every replay was **charged in full**

The question that surfaced it — *"what are the most popular models"* — is exactly the shape that breaks: its correct answer changes daily, and the app had no way to refresh it by asking again.

**It compounds with a moderation withhold, which is the half that looks like a different bug.** Asking again after a withhold resubmits the same input, replays the same refused text and charges again — so the withhold reads as permanent and the retry reads as broken.

## Design notes

- **`seed`, not `user`.** `seed` already means *resample* on the input contract this entry is anchored to. `user` was rejected deliberately: it is an end-user identifier forwarded upstream, and randomising it would corrupt the abuse signal it exists to carry.
- **The param surface is unchanged.** A caller-supplied `seed` is still refused by `.strict()` — otherwise a block could pin itself back onto a stored reply and re-enable exactly the replay this removes. A cross-reference now sits on that entry in the rejected-field list, at the line an edit would touch.
- **Web Crypto, not `node:crypto`.** This module is reachable from client code — `workflow.schema` imports the step registry for `REGISTERED_STEP_IDS`, and `BlockGenerationSourceUploadModal.tsx` imports that schema. A `~/server/` path is not a guarantee of server-only bundling. Whether a `crypto` import would actually break the client build was **not measured**; `globalThis.crypto.getRandomValues` removes the question rather than answering it. Note that neither `typecheck` nor the unit suites could have caught a bad choice here — neither bundles.
- **`2 ** 31` must stay a power of two.** The draw is a uniform 32-bit value reduced modulo the bound; 2^32 is an exact multiple of 2^31, so it is exactly uniform. A non-power-of-two bound would introduce silent modulo bias that no test here would see. Documented at the constant.

## Cost, stated

This gives up the execution saving on a repeated identical ask **for this entry**. That saving is real and is being spent deliberately: a stale answer was charged at full price either way, so the storage reuse was saving execution while still billing the reader.

Consequence worth naming: a duplicate submit of an identical payload now costs a real execution rather than a stored read.

## Tests

Four cases in the `buildStep` describe, each pinning a **different half** — "it varies" is not the whole property:

| case | pins |
|---|---|
| varies the seed across submits of identical params | it is not replayed |
| varies **only** the seed | the prompt it is attached to is untouched |
| draws from the full 2^31 space | the range, not merely that it moves |
| (cross-reference only) | a caller-supplied `seed` stays refused — already pinned by the existing rejected-field list, so deliberately **not** re-asserted as new coverage |

**Mutation-checked**, each mutant isolated to the call site:

| mutant | result |
|---|---|
| no seed emitted (the pre-change build) | RED — distinctness + range (+ the 3 key-set guards) |
| constant seed `0` | RED — distinctness |
| narrowed range (`% 1000`) | RED — **range only** |
| seed derived from `params` | RED — distinctness |
| buster reaches the prompt | RED — "varies only the seed" |

The narrowed-range mutant is killed by the range case **alone** — distinctness survives it — so that case is not redundant.

## Verification

- `pnpm typecheck` — **0 type errors** (80s, heap cap 8192 MB)
- `npx vitest run src/server/services/blocks/ src/server/routers/__tests__/blocks.router.workflow.test.ts src/server/schema/blocks/__tests__/ src/server/services/__tests__/no-unguarded-billable-submit.test.ts` — **180 files / 4973 tests passed**
- `pnpm prettier:check` clean, with a negative control (a deliberately misformatted file was rejected by the same invocation), since "all matched files use Prettier code style" is also what zero checked files prints
- Three pre-existing exact-key-set assertions outside the file I edited failed on the first run and were updated — found by running the consumers, not the changed file

https://claude.ai/code/session_01Duou4UyL9j5MDF7pSpy1WK

### DIFF
--- a/src/server/routers/blocks.router.ts
+++ b/src/server/routers/blocks.router.ts
@@ -7051,10 +7051,26 @@ async function assertStepRequestAllowed(claims: BlockClaims): Promise<number> {
  * at load time.
  *
  * 🔴 SHARED BY THE QUOTE AND THE SUBMIT ON PURPOSE. Both the estimate's
- * `whatif:true` quote and the real submit go through here, so the two can never
- * price different things and neither can be given a step the other would have
- * refused. Extracting it is what makes "estimate quotes what submit bills" a
- * property of the code rather than of two call sites staying in step.
+ * `whatif:true` quote and the real submit go through here, so neither can be
+ * given a step the other would have refused. Extracting it is what makes
+ * "estimate quotes what submit bills" a property of the code rather than of two
+ * call sites staying in step.
+ *
+ * 🔴 BUT THIS IS NO LONGER A PURE FUNCTION, SO "THE TWO CAN NEVER PRICE
+ * DIFFERENT THINGS" IS NOT THE STRUCTURAL GUARANTEE IT READS AS. The estimate
+ * (`estimateStepWorkflow`) and the submit (`submitStepWorkflow`) are SEPARATE
+ * requests, each calling this helper once. `chat-completion`'s `buildStep`
+ * emits a fresh per-submit `seed` (see its `CHAT_COMPLETION_SEED_EXCLUSIVE_MAX`
+ * for why), so the two requests build steps that DIFFER in that field.
+ *
+ * The guarantee still holds today, but it now rests on an ORCHESTRATOR fact
+ * that nothing in this repo pins: chat-completion is priced from a token
+ * estimate and the price does not depend on the seed. **If the orchestrator
+ * ever prices a dedupe-eligible step cheaper — an obvious optimisation for a
+ * content-addressed store — the estimate would quote one seed's price and the
+ * submit reserve another's, and nothing here would go red.** Any future
+ * `buildStep` that varies a PRICED field breaks this outright. Keep
+ * per-request variation confined to fields the quote cannot see.
  */
 function buildStepOrchestratorStep(
   step: ReturnType<typeof resolveBlockStep>,
--- a/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
+++ b/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
@@ -332,6 +332,13 @@ describe('chat-completion — buildStep', () => {
     expect(
       Object.keys(chatCompletionStep.buildStep({ ...VALID_PARAMS, temperature: 0.7 }).input).sort()
     ).toEqual(['maxTokens', 'messages', 'model', 'seed', 'temperature']);
+    // 🔴 `temperature: 0` DROPS `seed`, and that is the one conditional key on
+    // this surface. Asserted here rather than as its own case because THIS test
+    // is the one claiming to pin the exact key set — leaving the conditional
+    // uncovered would make its name wider than what it checks.
+    expect(
+      Object.keys(chatCompletionStep.buildStep({ ...VALID_PARAMS, temperature: 0 }).input).sort()
+    ).toEqual(['maxTokens', 'messages', 'model', 'temperature']);
   });
 
   it('forwards the parsed params verbatim', () => {
--- a/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
+++ b/src/server/services/blocks/steps/__tests__/chat-completion.step.test.ts
@@ -4,6 +4,7 @@ import {
   CHAT_COMPLETION_MAX_OUTPUT_TOKENS,
   CHAT_COMPLETION_MODELS,
   CHAT_COMPLETION_PRICE_BUZZ,
+  CHAT_COMPLETION_SEED_EXCLUSIVE_MAX,
   chatCompletionStep,
   MAX_TOOL_ROUNDS,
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Correcting two claims in the PR body above

Round-1 audit found both. Recording them here rather than silently editing the body, since a reviewer may already have read the wrong version. Fixed in `d23d671786`.

**1. The mutation table's headline claim does not reproduce.** The body says the narrowed-range mutant (`% 1000`) is *"killed by the range case **alone** — distinctness survives it"*. That was my measurement, and it is not reliable: the distinctness case tolerates one collision in 32 draws, and 32 draws from 1000 values has ≈0.5 expected colliding pairs, so the row is a coin flip. My run saw 1 failure; the audit's saw 2.

The conclusion — that the range case is *not* redundant with distinctness — is still correct, and the mutant that shows it cleanly is **`% (2 ** 20)`**: 1 failure, the range case, distinctness green. Verified.

Two further rows were over-precise in the safe direction: `constant seed 0` is red on distinctness **and** range, not distinctness alone; `no seed emitted` produces **6** red, not 5 — the sixth is `forwards the parsed params verbatim`.

**2. "Three pre-existing exact-key-set assertions *outside the file I edited*" is wrong.** All of them live in this entry's own test file, which the PR already changes. `git diff --stat` was 2 files. The sentence claims blast-radius discovery across consumers that did not happen — running the wider suite was still what surfaced them, but they were never outside the diff.

## What the audit found that the body did not claim at all

- **`buildStep` is not server-only.** `assertStepInvariants` runs at module scope and calls it once per variant — measured at **3 calls** on registry import, before any request. My docstring asserted the opposite. That strengthens the Web Crypto choice rather than weakening it, and it is now stated correctly.
- **`buildStepOrchestratorStep` stopped being a pure function**, and its comment still advertised "the two can never price different things" as structural. Still true today, but only because chat-completion is priced from a token estimate that ignores the seed — an orchestrator fact this repo does not pin. Comment now says what it rests on.
- **The "MUST STAY A POWER OF TWO" invariant had no enforcement.** The range case compared the constant against a value derived from that same constant — true by construction. `2 ** 24` and `1_500_000_000` both survived the full 86-test file. A new case pins the bound; both mutants now die to it by name.

ZacxDev:
## One user-visible behaviour change this PR's body did not state

The round-1 audit flagged it as unverified; it is settled by reading the path, and it is worth having on the record before merge.

On a reuse hit the orchestrator's `GenerateJobsAsync` calls `UpdateOutputAsync` and then `yield break` — **no job is ever created**, so the step acquires its output and completes with no provider round-trip. So a repeated identical ask used to return **near-instantly**; after this change it takes a normal generation.

That is intended, not a regression. It is the same fact as "the reader was charged full price for a stored reply", seen from the latency side rather than the money side — you cannot stop serving the stale answer and also keep the speed of serving it. But it is a real change for the repeated-ask class:

- a block that resubmits an identical payload (a retry, a double-submit) now waits for a real generation instead of getting an immediate snapshot
- and it costs a real provider execution where it previously cost a stored read

Unchanged: the Buzz charged to the user, which was always the full quote either way — the quote is computed before job generation and never consulted the stored reply.

`temperature: 0` callers keep the old behaviour entirely, per the gate added in `bb37acb299`.

## Still not attributed to this PR: `preview / component-tests`

Traced to `main`, not to this change. `2cbacbc4d5` (2026-09-04) added `<PricingSlotHistory />` to `ModelVersionUpsertForm`; that component calls `trpc.modelVersion.getPricingSlots.useQuery`, and the test's `vi.mock('~/utils/trpc', …)` hand-lists only `getLicensingRoots`, `getUserEarlyAccessVersions`, `getPricingAllowance` and `upsert`. So `trpc.modelVersion.getPricingSlots` is `undefined`, `.useQuery` throws, hydration fails, and all 24 cases then time out on their locators.

The test file's own comment already warns about this exact failure mode — *"Hand-listed, so every router entry the component reads has to appear here or it throws on render — the whole suite went red on this one when the allowance counter was added."*

Sibling PRs pass because their merge-bases predate that commit; this PR is the first to run the check on a base containing it. **Fix is one line in that mock's router list** — deliberately not folded into this PR, which is unrelated to it.

PR #2649, opened 2026-06-19
perf(redis): compress + split tensor-metadata cache (kill the ~80GiB next-redis-cluster whale)

ZacxDev:
## The whale

`packed:caches:tensor-metadata` was measured at **~71-80% (~80 GiB)** of the hot `next-redis-cluster`, which is pinned at its memory cap and LRU-evicting hot feed/metrics caches. Root cause: PR #2500 caches the **full** parsed tensor array (`parseModelTensorMetadata` → ~335 KB msgpack, mostly repetitive tensor-name strings) in `redis.packed` with a 1-month TTL. The route is cold (~0.5 req/s total), the blob is highly compressible, and regeneration is cheap (header range-fetch only).

Measured with Node's built-in `zlib` on a real blob: **brotli quality 6 = 64.9× (80 GiB → ~1.2 GiB), decompress ~0.10 ms, compress ~0.87 ms** — cheapest decompress of all codecs tested. Fleet read-path cost is negligible (~0.12 CPU-ms/s).

### A. Opt-in brotli compression for the full-blob cache (back-compat safe)
- New `~/server/redis/packed-compression` using Node built-in **zlib brotli, quality 6** (`brotliCompressSync` / `brotliDecompressSync`, `BROTLI_PARAM_SIZE_HINT` set). **No new dependency.**
- Threaded a `compress?: boolean` option through `fetchThroughCache` → `redis.packed.set`. Compression happens on the msgpack-**packed** buffer (compress after pack on write; decompress before unpack on read).
- **Zero-downtime migration via a sentinel prefix byte** (`0x01` = brotli). The ~220k existing keys are uncompressed raw msgpack. New compressed writes carry the sentinel; the reader strips + inflates sentinel-tagged values and passes legacy raw-msgpack through untouched.
  - **Sentinel safety (documented in code):** values stored through `fetchThroughCache` are ALWAYS the `{ data, cachedAt }` wrapper object → the msgpack first byte is a map marker (`0x80–0x8f` fixmap / `0xde` map16 / `0xdf` map32), **never** a fixint `0x01`. So the sentinel cannot collide with any legacy value on this path. Compression is intentionally NOT enabled for any caller that stores a bare scalar.
- Enabled `compress: true` **only** for the tensor-metadata full-blob cache.

### B. Split summary / full so the hot summary path never touches the big blob
- Added `REDIS_KEYS.CACHES.TENSOR_METADATA_SUMMARY` (`'packed:caches:tensor-metadata-summary'`).
[...]

PR #4649, opened 2026-09-05
perf(redis): brotli-compress imageMetaCache values at rest (#4588)

### LINKED ISSUE #4588: perf(redis): image-meta swings ~5→27 GiB on the hot shard and is stored uncompressed — brotli measures 3.04x
> **Corrected 2026-09-03.** The first version of this issue said `image-meta` was "~65% of the cache dataset". That was measured during a peak and is wrong as a steady-state figure — at rest it is ~23%. The corrected framing is below, and it is a *sharper* argument for the same fix, not a weaker one. See the correction comment for the full diff of what changed.

## Summary

`packed:caches:image-meta` is the **elastic** component of the redis cache — the one that swings hardest with load — and it is stored **uncompressed**. On the hot shard it moves roughly **5.3 GiB at rest → 27.3 GiB at peak**, and it is what pushes that shard into its memory cap.

Its values compress ~3×. Compressing them would cut the peak enough to keep the shard under its cap, which is the whole problem. This is the same shape and the same fix as the `tensor-metadata` whale that #2649 resolved in June.

## What happened

On 2026-09-03 the hot shard filled from 25.2 → 40.0 GiB in about three hours, hit its cap, evicted **57.1M keys** over the following ~6–10 hours, then drained back down on its own. That was the **first such excursion in 14 days** — over that window the shard oscillates 23–36 GiB and does not normally reach cap. So this is an episodic capacity problem, not a permanently-full cache.

`image-meta` is what moved. Sampling 20,000 keys by `MEMORY USAGE` on the hot shard, at the peak and again at the trough:

| prefix | at peak (40.0 GiB used) | at rest (22.9 GiB used) |
|---|---|---|
| `packed:caches:image-meta` | **68.3%** (~27.3 GiB) | 23.0% (~5.3 GiB) |
| `packed:caches:tag-ids-for-images` | 12.2% | **25.6%** (~5.9 GiB) |
| `packed:caches:images-for-model-version-2` | 4.7% | 9.4% |
| `metrics:Image` | 2.9% | 6.0% |

Every other prefix's *share* rises at rest simply because `image-meta` shrank. In absolute terms `image-meta` is the one that goes ~5× — everything else is roughly flat. Mean value size is stable across both samples (3,419 B at peak, 3,594 B at rest), so what swings is the **number** of live keys, not their size.

## Why the TTL lever is already spent

Stated up front so nobody re-derives it: **1200/1200 sampled keys carry a TTL**, max 4.99h — matching `ttl: CacheTTL.hour * 4` + `staleWhileRevalidateTtl: CacheTTL.hour` in `src/server/redis/caches.ts`. There is no leaked or dead data to bust, and that file's own comment (correctly) forbids shortening the logical 4h:

> Do NOT shorten the logical 4h ttl: that would convert the hot working set into misses and stampede dbRead.

The SWR tail was already trimmed 8h → 5h in an earlier pass. Compression is what's left.

## Compression measurements

Measured on **400 real values** pulled from the live cache:

| strategy | ratio | bytes saved |
|---|---|---|
| brotli q=5, every value | **3.04×** | **67.2%** |
| brotli q=11, every value | 3.35× | 70.1% |
| zlib -6, every value | 2.84× | 64.8% |
| brotli q=5, only values > 4 KiB (5.8% of keys) | 1.96× | **48.9%** |
| brotli q=5, only values > 2 KiB (19.2% of keys) | 2.21× | 54.8% |

Size distribution over 1,200 keys — note the fat tail, which is why a threshold works so well: p50 **1.6 KB**, p90 3.2 KB, p99 **28.8 KB**, max **328 KB**; the top 1% of keys hold **28.8%** of the prefix's bytes and the top 10% hold **58.2%**.

**The payoff is on the peak, not the average.** At full q=5, `image-meta`'s peak contribution drops ~27.3 → ~9 GiB, which takes the hot shard's peak from 40 GiB to roughly 22 GiB — comfortably under its cap, so the excursion stops happening. At rest it saves a more modest ~3.5 GiB, which is real but not the point.

## Suggested implementation

The primitive already exists and is back-compat safe — `redis.packed.get/set` take `{ compress }`, entries are **sentinel-tagged**, and reads transparently decode both compressed and legacy-uncompressed values, so enabling it needs no key-bust or migration.

⚠️ But the option is currently surfaced only on `fetchThroughCache` (`src/server/utils/cache-helpers.ts:256`). **`imageMetaCache` is a `createCachedObject`, which does not accept it** — so this is not a one-line flag flip. The work is:

1. Thread `compress` through `createCachedObject`'s options into its `redis.packed` read and write paths, mirroring how `fetchThroughCache` already does it (including the bust symmetry note at `cache-helpers.ts:400-409` — a compressed key busted with `compress=false` throws).
2. Enable it on `imageMetaCache` in `src/server/redis/caches.ts`.

**Consider a size threshold rather than compressing everything.** The existing comment on the option says as much — *"Only enable for LARGE, highly-compressible blobs … most cached values are tiny and would only pay the codec overhead"* — and this prefix's p50 is only 1.6 KB. **> 4 KiB captures 48.9% of the savings from 5.8% of writes** and looks like the best CPU-per-byte-saved point; compressing everything is worth ~18pp more if the CPU proves cheap.

Please benchmark the decode cost on the read path before choosing — #2649's own service notes call out that a popular file must be brotli-decompressed **plus** msgpack-unpacked per request, and that's what motivated its per-pod LRU. `image-meta` is read far more often than tensor-metadata, so decode cost matters more here — another argument for the threshold.

Note that `imageMetaCache` deliberately has **no per-pod L1** (the `hideMeta` privacy gate — a cross-pod `refresh()` can't be observed by an L1, which would risk serving a stale prompt after an owner hides it). Preserve that; don't "fix" decode cost by adding an L1 here.

## Priority

This is **not** urgent — the cache self-recovered and is currently at 57% of cap with zero evictions. It is worth doing because the excursion cost 57.1M evicted keys (i.e. a miss storm against the DB) and will recur under the same load.

## Closing condition

A PR enabling compression on `imageMetaCache` is merged to `main`, **and** a re-measurement with the same method (sample 20k keys by `MEMORY USAGE`, aggregate by prefix with trailing id segments collapsed) taken **during a peak** shows `packed:caches:image-meta` holding **below 40%** of the hot shard's dataset.

Checked by: the merged PR (mechanical), plus the re-measurement run by whoever merges it. ⚠️ The comparison must be peak-to-peak — the 68.3% peak figure is the before-value; a trough sample will read ~23% whether or not this is fixed, so measuring at rest would falsely show success.

Alternative valid close: if the decode-cost benchmark shows compression is a net loss on the read path, close as **declined** with the benchmark numbers attached. Checked by: @ZacxDev reading the posted benchmark.

## Context

- Precedent: #2649 (brotli-compress + summary-split for `tensor-metadata`) — same playbook, different prefix.
- Cache-side operational detail and alert history live in the private infra repo; ping @ZacxDev if you need it.

https://claude.ai/code/session_01Ar44488ch8PeZSMo3Hko8a

ZacxDev:
Closes #4588.

`packed:caches:image-meta` was stored uncompressed. It is the **elastic** component of the packed cache — its resident set swings ~5× under load and is what drives the cache to its memory cap — and its values are prompt / generation-parameter text, i.e. highly repetitive. Measured on 336 real values sampled **at a peak** (issue #4588, [benchmark comment](https://github.com/civitai/civitai/issues/4588#issuecomment-5549843539)): brotli q=6 gives **3.09×, 67.6 % of bytes saved**, at **under 2 % of app CPU** as a deliberately pessimistic upper bound (that bound attributes *all* cache read+write traffic to this one prefix; the true figure is a fraction of it). The codec is already `promisify(zlib.brotli*)`, so it runs on the libuv threadpool and never blocks the event loop.

### 1. `packed.mGet` gains a `compress` option — this was a hard blocker, not a nicety

The issue describes the work as "thread `compress` through `createCachedObject`". That is necessary but **not sufficient**: `redis.packed.mGet` had no `compress` option at all — only `packed.get` did — and `createCachedArray` reads **exclusively** through `mGet`. Enabling compression without this would have written sentinel-tagged values that the batched read decodes on the general msgpack path: `unpack()` throws, `safeUnpack` **UNLINKs the entry**, and the read reports a miss. A permanent miss+evict loop, with no error surfaced anywhere and a *higher* DB load than before the change.

`mGet` now routes through `safeUnpackCompressed` when opted in, exactly as `get` does.

### 2. `createCachedArray` / `createCachedObject` gain `compress`, threaded to **all nine** packed call sites

Symmetry is the entire contract — one missed site is silent cache corruption on exactly the path that was missed. The full ledger, all in `packages/civitai-redis/src/cached-array.ts`:

| # | site | op | how it is covered |
|---|---|---|---|
| 1 | `fetch` batch read | `mGet` | behavioural (round-trip + mixed-population) |
| 2 | `fetch` value write | `set` | sentinel-byte assertion |
| 3 | `fetch` `notFound` marker write | `set` (NX) | sentinel-byte + no-re-query assertion |
| 4 | `bust` debounce-marker write | `set` | sentinel byte + marker decodes to `{ debounce: true }` |
| 5 | `invalidate` read | `mGet` | behavioural (SWR `bust` rewrites the entry) |
| 6 | `invalidate` rewrite | `set` | sentinel byte |
| 7 | `refresh` rewrite | `set` | sentinel byte + subsequent read hits |
| 8 | `update` read | `mGet` | behavioural (`update` returns `true`) |
| 9 | `update` rewrite | `set` | sentinel byte + value round-trips |

`redis.del` / `setNxKeepTtlWithEx` carry no codec and are untouched. A **seam-ledger test** asserts the count is exactly 9 in both directions and that every one passes the options object, so a tenth call site added later cannot be silently asymmetric.

**Sentinel safety re-verified for this new caller.** The 0x01 brotli sentinel is only collision-free because every value on the compress-aware path is an object (a msgpack map marker is `0x80–0x8f`/`0xde`/`0xdf`, never `0x01`). Everything `createCachedArray` writes is a record — `{ ...result, cachedAt }`, `{ [idKey]: id, notFound: true, cachedAt }`, `{ [idKey]: id, debounce: true }` — and `T extends object`, so a bare scalar cannot reach the write path. The doc comments in `packed-compression.ts` and `client.ts` are updated to name both callers rather than just `fetchThroughCache`.

### 3. `imageMetaCache` opts in — **all values, not a threshold**

The issue body suggests `> 4 KiB` as "the best CPU-per-byte-saved point". Measured, it is not a good trade:

| threshold | keys touched | bytes saved (whole prefix) |
|---|---|---|
| **all values** | 100 % | **67.6 %** |
| `> 1 KiB` | 78.6 % | 66.6 % |
| `> 2 KiB` | 31.8 % | 57.5 % |
| `> 4 KiB` | 6.0 % | **47.5 %** |

`> 4 KiB` gives up **~20 percentage points** — the difference between clearing the cap at peak and not — to buy back a fraction of a percent of app CPU. That trades away the resource that is scarce for one that is not. `> 1 KiB` is within 1 pp of "all", and since the primitive has no threshold support, "all" is also the least code. (Codec and quality are unchanged: the primitive hardcodes q=6, which measures an identical ratio to q=5.)

**Back-compat needs no key-bust or migration.** Entries are sentinel-tagged and the compress-aware read discriminates on the first byte, so legacy uncompressed entries written before the flag still decode. Covered by a test that serves a **mixed population** — one legacy raw entry and one freshly compressed entry — in a single `fetch`, which is the state of the cache for the whole TTL after the flag flips. Note the asymmetry: turning the flag back **off** is *not* free (a compressed key read on the general path throws and is evicted), so that direction would need a bust. Called out in the option's doc comment.

**The deliberate absence of a per-pod L1 on this cache is preserved and now guarded.** `meta` is gated on `hideMeta`, and a cross-pod `refresh()` cannot be observed by an L1, so one would keep serving a hidden prompt. Compression changes decode cost, which is exactly the argument someone would reach for an L1 with — a test now pins its absence so that reasoning has to come back through review.

## Tests

`packages/civitai-redis/src/__tests__/cached-array-compress.test.ts` drives the **real** `createCacheRedis()` client against an in-memory transport (only `redis`'s client factory is faked). Everything above the socket — the packed wrappers, msgpack, brotli, the cache builders — is the real module. That is deliberate: a suite that mocks `redis.packed` would pass against a client whose `mGet` ignores the flag entirely, i.e. against the exact production defect. Driving the real client also lets the tests assert the **on-disk bytes**, which is the only way to tell "compression is on" from "the round-trip happens to work anyway".

Plus `src/server/redis/__tests__/image-meta-cache-compress.test.ts`, a config guard on the declaration site (`caches.ts` builds 20+ module-scope singletons behind the env/clickhouse import graph, so importing it from a unit test is not viable, and the cache objects do not expose the options they were built with). It carries two positive controls: the extractor is brace-scoped to `imageMetaCache` and proven so against an uncompressed neighbour, and a missing declaration throws rather than passing vacuously.

One existing assertion, `cache-helpers-update.test.ts`, was updated for the new packed-options argument.

### Red/green matrix

Base ref `d197ecbea2` (`origin/main`), HEAD `44f5a6ee74` (matrix measured at `d84df61ec6`; the follow-up commit is test-only and adds no new red/green rows — its evidence is the mutation table below it). Both suites run at the base with the test files copied in.

| suite | at `d197ecbea2` | at HEAD |
|---|---|---|
| `cached-array-compress.test.ts` | **10 failed** / 4 passed (14) | 14 passed |
| `image-meta-cache-compress.test.ts` | **1 failed** / 5 passed (6) | 6 passed |

The 5 that are green at base are **invariant guards**, labelled as such in the file and not counted as regression coverage:

- *round-trips a value through a compressed cache* — with `compress` ignored, reads and writes are still symmetric, just uncompressed. It pins that enabling compression does not break the read-through contract; the regression coverage is the sentinel assertion.
- *finds exactly the 9 known redis.packed call sites* — the nine sites exist at base too. Its job is to keep the ledger honest: a detector that goes blind fails here rather than reporting zero violations over zero call sites.
- the two *compress:false / default is unchanged* tests — green at base by construction, which is the point. They go red only if someone flips the default on.
- *a missing declaration throws* (the config guard's own positive control).

### Mutation results — 13 mutants, **0 survivors**

Each mutant applied alone to a clean tree, both suites run, tree restored. An unmutated control run (14/14 and 6/6 green) was recorded first, so "killed" is attributable.

| # | mutation | killed by | failure |
|---|---|---|---|
| M1 | drop the flag from the `fetch` value **write** | sentinel + mixed-population + ledger | `expected 222 to be 1` (0xde map marker, not the sentinel) |
| M2 | drop it from `fetch`'s **mGet** | 5 behavioural + ledger | `lookupFn called 2 times` — the miss+evict loop |
| M3 | drop it from the `notFound` write | notFound test + ledger | `expected 222 to be 1` |
| M4 | drop it from `bust`'s debounce write | bust test + ledger | `expected 222 to be 1` |
| M5 | drop it from `invalidate`'s mGet | SWR-bust test + ledger | `entry was evicted — invalidate read it on the non-compress path` |
| M6 | drop it from `invalidate`'s write | SWR-bust test + ledger | `expected 222 to be 1` |
| M7 | drop it from `refresh`'s write | refresh test + ledger | `expected 222 to be 1` |
| M8 | drop it from `update`'s mGet | update test + ledger | `update() found no entry — the compressed read decoded on the wrong path` |
| M9 | drop it from `update`'s write | update test + ledger | `expected 222 to be 1` |
| M10 | accept the option but hardcode `{ compress: false }` | 8 tests | 7× sentinel + the "derived, not hardcoded" guard |
| M11 | **the seam** — `client.ts` `mGet` ignores the flag (the pre-#4588 state) | 7 tests | `lookupFn called 2 times`; `expected null to deeply equal {…}` |
| M12 | `decompressPacked` skips the sentinel branch | 7 tests | same set as M11 (same mechanism, one layer lower) |
| M13 | remove `compress: true` from `imageMetaCache` | config guard | `expected '{\n  key: REDIS_KEYS.CACHES.IMAGE_MET…' to match /compress:\s*true/` |

Two things worth stating plainly rather than leaving implicit:

- **M11 and M12 produce identical failure sets.** They are the same mechanism at two layers, so neither is independent evidence about the other.
- **A dropped flag on a WRITE site (M1, M3, M4, M6, M7, M9) is behaviourally invisible** — the value is written uncompressed and the compress-aware read decodes it fine as "legacy". Those mutants are killed *only* by the on-disk sentinel assertions and the ledger. This is exactly why the tests assert bytes rather than round-trips: a suite built on round-trips alone would have scored six survivors while reporting full coverage.

### Commands run

| command | result |
|---|---|
| `pnpm typecheck` | `OK — 0 type errors in 68s` |
| `npx vitest run --project '@civitai/redis'` | 18 files, **170 passed** |
| `pnpm test:packages:run` | 87 files, 1178 passed / **8 failed** — all in `@civitai/db-queries` `*.explain.test.ts`, which need a live Postgres; unrelated |
| `pnpm test:unit:run` | 1646 files, **38 212 passed** / 4 files failed — `apps/moderator`'s `.svelte-kit/tsconfig.json` is not generated in a fresh checkout; the identical 4 fail at `d197ecbea2` |
| `npx prettier --check` (changed files) | clean |
| `npx eslint` (changed files) | 6 pre-existing `no-explicit-any` in `caches.ts` (lines 305–649, untouched by this PR); byte-identical at `d197ecbea2` |

### Round-1 audit follow-up — `packed.mGet`'s index contract (commit `44f5a6ee74`)

An adversarial audit reproduced the 13-mutant table and the red/green matrix above exactly, and found **no blocking issue** — but it wrote six mutants of its own, and two of them survived the full package project against code *this PR adds*: the `if (packedOptions?.compress)` branch in `mGet` had no index-alignment guard.

| mutation | why it matters |
|---|---|
| `client.unlink(keys[i])` → `keys[0]` | a decode failure evicts the **first** key of the batch, not the offending one — so one undecodable entry repeatedly evicts a healthy neighbour and never evicts itself. Reachable in production: the mixed-fleet rollout described below is precisely when this path meets values it cannot decode. |
| reverse the returned array | `mGet`'s contract is positional. `createCachedArray` re-keys the batch by `idKey` immediately after the read, so it cannot observe the break — the next caller would. |

Fixed test-only, one test per branch: a **3-key batch whose MIDDLE key holds an undecodable value**, asserting the results are `[value, null, value]` *in that order* and that **exactly** the offending key was evicted. Failure in the middle means an off-by-one and a first-element collapse land on different keys; the two good values are pairwise distinct so a reversal is detectable (a reversal keeps the `null` in the middle, so only distinct outer values expose it); eviction is asserted as an exact set, so a wrong, extra, or absent eviction all fail alike. A positive control pins that the corrupt fixtures really are undecodable on their own path — a fixture that silently decoded would make both tests pass while proving nothing.

**The general (non-compress) branch is guarded too.** It is pre-existing code this PR does not modify, and the guard adds no production diff — but it carries the identical `keys[i]` shape one line below the branch that was added, so an edit to one is overwhelmingly likely to be made to both. Closing the class rather than the instance.

Each mutant was run against the **full package project**, both with and without the new guard, so the kills are attributable:

| mutant | without the guard | with the guard |
|---|---|---|
| compress, `unlink keys[0]` | **SURVIVED** 170/170 | KILLED — `only the undecodable key may be evicted` |
| compress, reversed array | **SURVIVED** 170/170 | KILLED — the order assertion |
| general, `unlink keys[0]` | **SURVIVED** 170/170 | KILLED — `only the undecodable key may be evicted` |
| general, reversed array | **SURVIVED** 170/170 | KILLED — the order assertion |

Each fails **exactly one** test, the one owning that branch, with no collateral from another assertion. Package project: 170 → **173 passing**.

### A pre-existing gap this PR does NOT close

`packed.get`'s compress branch has no client-level test, so deleting `if (packedOptions?.compress)` from **`get`** (as opposed to `mGet`) survives the suite. That is #2649's code, not this PR's, and closing it would expand the diff into unrelated territory — recorded here so it is visible rather than silently inherited.

## Not done, deliberately

- **No size threshold** — see the table above.
- **No codec or quality change.** q=6 vs q=5 measured an identical 3.09× at ~20 % more CPU on a negligible base.
- **No switch to zstd.** It would win on a clean slate (97 % of the saving for roughly a third of the CPU), but the brotli path is already sentinel-tagged and back-compat safe, and CPU is not the binding constraint. Recorded so it is not re-litigated.
- **No per-pod L1** on this cache — see above.

## One second-order effect worth naming

During the rollout window, a pod running the OLD image that reads an `image-meta` key a NEW pod already wrote compressed will fail to decode it, **evict it, and refetch from origin**. That is the documented and intended behaviour of the sentinel design (see the mixed-fleet note at the bottom of `packed-decode-paths.test.ts`) — fail-open, no corruption, no 5xx — but it does mean a transient bump in `image-meta` misses against the DB while the fleet converges, on top of the normal deploy churn. It is bounded by the deploy duration, not by the 4 h TTL.

Verifying the issue's close condition is a peak-to-peak re-measurement, which is a runtime observation and not something this PR can assert.

### DIFF
--- a/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
+++ b/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
@@ -83,7 +83,7 @@ vi.mock('redis', () => {
 
 import { createCacheBuilders, type CachedLookupOptions } from '../cached-array';
 import { createCacheRedis } from '../client';
-import { PACKED_BROTLI_SENTINEL, compressPacked } from '../packed-compression';
+import { PACKED_BROTLI_SENTINEL, compressPacked, decompressPacked } from '../packed-compression';
 import type { RedisKeyTemplateCache } from '../client';
 
 const redis = createCacheRedis();
@@ -388,5 +388,78 @@ describe('the symmetry hazard this flag exists to prevent', () => {
     store.set(key, await compressPacked(Buffer.from(pack({ id: 99, blob: bigBlob }))));
     const [viaCompress] = await redis.packed.mGet<Row>([key], { compress: true });
     expect(viaCompress).toEqual({ id: 99, blob: bigBlob });
+  });
+});
+
+describe('packed.mGet index alignment — the OTHER half of the seam', () => {
+  /**
+   * `mGet`'s contract is POSITIONAL: result[i] belongs to keys[i], and a decode failure at i must
+   * evict keys[i] — nobody else. Both halves are unguarded by everything above, because
+   * `createCachedArray` re-keys the batch by `idKey` immediately after the read and so cannot
+   * observe either defect. The next caller would.
+   *
+   * The eviction half is not hypothetical for this PR: during a mixed-fleet rollout a pod WILL
+   * meet entries it cannot decode (see the compress-symmetry test above), so a mis-indexed
+   * `unlink` would repeatedly evict a healthy NEIGHBOUR while leaving the bad entry in place —
+   * a self-sustaining eviction of good data that looks like ordinary cache churn.
+   *
+   * Shape of the fixture, which is what makes both defects visible:
+   *  - THREE keys, failure in the MIDDLE, so an off-by-one and a first-element collapse land on
+   *    different keys and neither can be satisfied by luck;
+   *  - the two good values are PAIRWISE DISTINCT, so a reversed array is detectable (a reversal
+   *    keeps `null` in the middle — only distinct outer values expose it);
+   *  - eviction is asserted as an exact SET over the store, so it fails whether the wrong key is
+   *    evicted, an extra key is evicted, or nothing is evicted at all.
+   */
+  const K = (n: number) => `${KEY}:align-${n}` as RedisKeyTemplateCache;
+  const first = { id: 101, blob: 'first-value' };
+  const last = { id: 103, blob: 'last-value' };
+  // Sentinel-tagged but NOT a valid brotli stream: reaches the compress-aware decode and throws
+  // inside brotliDecompress, which is the realistic corruption on this path.
+  const CORRUPT_COMPRESSED = Buffer.from([PACKED_BROTLI_SENTINEL, 0xff, 0xff, 0xff, 0xff]);
+  // A truncated msgpack map16 header — no sentinel byte, so it is a corrupt LEGACY value and
+  // fails inside `unpack` on the general path.
+  const CORRUPT_PLAIN = Buffer.from([0xde, 0x00, 0x05]);
+
+  it('COMPRESS branch: positional order is preserved and ONLY the failing key is evicted', async () => {
+    store.set(K(1), await compressPacked(Buffer.from(pack(first))));
+    store.set(K(2), CORRUPT_COMPRESSED);
+    store.set(K(3), await compressPacked(Buffer.from(pack(last))));
+
+    const got = await redis.packed.mGet<Row>([K(1), K(2), K(3)], { compress: true });
+
+    // ORDER: index 0 is the FIRST key's value, index 2 is the LAST key's. Reversing the array
+    // keeps the null in the middle, so only these two distinct outer values can catch it.
+    expect(got).toEqual([first, null, last]);
+
+    // EVICTION: exactly the offending key, and nothing else. An exact set rather than three
+    // independent `has` calls, so a wrong/extra/absent eviction all fail the same assertion.
+    const survivors = [K(1), K(2), K(3)].filter((k) => store.has(k));
+    expect(survivors, 'only the undecodable key may be evicted').toEqual([K(1), K(3)]);
+  });
+
+  it('NON-COMPRESS branch: same contract, same fixture shape', async () => {
+    // The general path is PRE-EXISTING code this PR did not touch, but it carries the identical
+    // `keys[i]` shape one line below the branch that was added — so a future edit to one half is
+    // overwhelmingly likely to be made to both. Guarded here to close the class rather than the
+    // instance; this is test-only and expands no production diff.
+    store.set(K(4), Buffer.from(pack(first)));
+    store.set(K(5), CORRUPT_PLAIN);
+    store.set(K(6), Buffer.from(pack(last)));
+
+    const got = await redis.packed.mGet<Row>([K(4), K(5), K(6)]);
+
+    expect(got).toEqual([first, null, last]);
+    const survivors = [K(4), K(5), K(6)].filter((k) => store.has(k));
+    expect(survivors, 'only the undecodable key may be evicted').toEqual([K(4), K(6)]);
+  });
+
+  it('POSITIVE CONTROL: the corrupt fixtures really are undecodable on their own path', () => {
+    // Without this, a fixture that silently DECODED would make both tests above pass while
+    // proving nothing — the null would never be produced and no eviction would ever be attempted.
+    expect(CORRUPT_COMPRESSED[0]).toBe(PACKED_BROTLI_SENTINEL);
+    expect(CORRUPT_PLAIN[0]).not.toBe(PACKED_BROTLI_SENTINEL);
+    expect(() => unpack(CORRUPT_PLAIN)).toThrow();
+    return expect(decompressPacked(CORRUPT_COMPRESSED)).rejects.toThrow();
   });
 });
--- a/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
+++ b/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
@@ -1,0 +1,392 @@
+import { readFileSync } from 'fs';
+import { pack, unpack } from 'msgpackr';
+import path from 'path';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * End-to-end coverage for `compress` on createCachedArray / createCachedObject (issue #4588,
+ * shipped for `imageMetaCache`).
+ *
+ * WHY THIS FILE DRIVES THE REAL CLIENT: the hazard here is a SEAM, not a component. The flag has
+ * to be symmetric across TWO modules — `cached-array.ts` must pass `{ compress }` to every one of
+ * its nine `redis.packed` reads and writes, AND `client.ts`'s `packed.mGet` must actually honour
+ * it (before #4588 it could not — only `packed.get` took the option, and `createCachedArray` reads
+ * exclusively through `mGet`). A test that mocks `redis.packed` proves neither half: it would pass
+ * against a client whose mGet ignores the flag entirely, which is the exact production defect —
+ * a compressed value decoded on the general msgpack path throws, the entry is EVICTED, and the
+ * read reports a MISS. Permanent miss+evict loop, no error anywhere.
+ *
+ * So the only thing faked here is the TRANSPORT: `redis`'s client factory is replaced by an
+ * in-memory server holding raw Buffers. Everything above the socket — `createCacheRedis`,
+ * `client.packed.set/mGet/get`, the msgpack + brotli codec, `createCacheBuilders` — is the real
+ * module. That also lets these tests assert the ON-DISK BYTES (the 0x01 brotli sentinel), which is
+ * the only way to tell "compression is on" from "the round-trip happens to work anyway".
+ */
+
+// process.env must be populated BEFORE ../client is imported: loadRedisEnv() validates it.
+vi.hoisted(() => {
+  process.env.REDIS_URL ??= 'redis://localhost:6379';
+  process.env.REDIS_SYS_URL ??= 'redis://localhost:6379';
+  process.env.REDIS_CLUSTER = 'false';
+});
+
+/** The in-memory "server": key -> raw stored bytes, exactly as a real Redis would hold them. */
+const { store } = vi.hoisted(() => ({ store: new Map<string, Buffer>() }));
+
+vi.mock('redis', () => {
+  const toBuffer = (v: unknown): Buffer =>
+    Buffer.isBuffer(v) ? v : Buffer.from(v as string, 'binary');
+
+  // A single fake node client. `withTypeMapping` returns a Buffer-mode VIEW over the same store
+  // (mirroring node-redis: same connection, different reply decoding).
+  const makeClient = (bufferMode = false): Record<string, unknown> => {
+    const self: Record<string, any> = {
+      connect: () => Promise.resolve(self),
+      on: () => self,
+      withTypeMapping: () => makeClient(true),
+      get: async (key: string) => {
+        const v = store.get(key);
+        if (v === undefined) return null;
+        return bufferMode ? v : v.toString('binary');
+      },
+      set: async (key: string, value: unknown, options?: { NX?: boolean; XX?: boolean }) => {
+        const exists = store.has(key);
+        if (options?.NX && exists) return null;
+        if (options?.XX && !exists) return null;
+        store.set(key, toBuffer(value));
+        return 'OK';
+      },
+      del: async (key: string) => (store.delete(key) ? 1 : 0),
+      unlink: async (key: string) => (store.delete(key) ? 1 : 0),
+      // Backs client.setNxKeepTtlWithEx's EVAL (SET NX KEEPTTL + EXPIRE). TTLs are irrelevant to
+      // the codec contract under test, so the lock is modelled as plain set-if-absent.
+      eval: async (_script: string, opts: { keys: string[] }) => {
+        const key = opts.keys[0];
+        if (store.has(key)) return 0;
+        store.set(key, Buffer.from('1'));
+        return 1;
+      },
+      scanIterator: async function* () {
+        /* unused */
+      },
+    };
+    return self;
+  };
+
+  return {
+    createClient: () => makeClient(false),
+    createCluster: () => makeClient(false),
+    createSentinel: () => makeClient(false),
+    RESP_TYPES: { BLOB_STRING: 36 },
+  };
+});
+
+import { createCacheBuilders, type CachedLookupOptions } from '../cached-array';
+import { createCacheRedis } from '../client';
+import { PACKED_BROTLI_SENTINEL, compressPacked } from '../packed-compression';
+import type { RedisKeyTemplateCache } from '../client';
+
+const redis = createCacheRedis();
+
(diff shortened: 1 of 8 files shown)


ZacxDev:
**Round-1 adversarial audit — claims block.** Verdict was *safe to merge, no blocking findings*. Recording what round 1's fixes claim to have addressed, so a round-2 delta re-audit has a machine-readable anchor rather than re-auditing the whole PR.

The audit independently reproduced the implementation's 13-mutant table (all killed, each for its own guard's reason) and its red/green matrix (10-of-14, 1-of-6) exactly. It then wrote six mutants of its own, of which **three survived** the full 170-test package project — two in code this PR adds, one pre-existing. The fix commit `44f5a6ee74` is **test-only: zero production files changed**.

```audit-claims round=1 audited=d84df61ec6d430622d69d76fe45405432efcb3af..44f5a6ee
1. F2 FIXED: the compressed mGet branch's positional/index contract is now guarded by a 3-key batch test whose middle key is undecodable, asserting result order [first, null, last] and an exact survivor set; the two surviving mutants (unlink(keys[0]) instead of keys[i]; reversed result array) are claimed killed, each by exactly one test and by its own assertion.
2. F2 CLASS-CLOSED: the identical keys[i] shape in the adjacent non-compress mGet branch received the same guard, so the class is covered rather than the reported instance. Claimed zero production diff.
3. ATTRIBUTION CONTROL claimed: all four mutants were re-run with the new test reverted to d84df61ec6 and each SURVIVED 170/170, so the kills are attributed to the new guard and not to a pre-existing assertion.
4. POSITIVE CONTROL claimed: the corrupt fixtures are asserted to be genuinely undecodable on their own path, so neither new test can pass by silently decoding and never producing a null or attempting an eviction.
5. F5 NOT FIXED, deliberately: packed.get's compress branch still has no client-level test (deleting its compress branch survives). Pre-existing code from the earlier compression PR, recorded in the PR body as a known gap and held out of scope.
6. F1 NOT FIXED, deliberately: rollback remains asymmetric (turning compress off needs a key-bust) because the general decode path never attempts brotli. Deferred as a design decision, not an oversight.
7. F4 NOT FIXED, context only: rollout-window misses land on the hard-miss path, which is not single-flighted.
8. F3 NOT ADDRESSED HERE: a deployment-configuration issue found by the same audit is tracked outside this repository and is not a code change to this PR.
```

Round 2 will be dispatched as a delta against `d84df61ec6..44f5a6ee74`, and asked to state per prior claim: actually fixed / partially / not / made worse — treating each line above as a claim to check against the diff rather than a fact.

PR #4651, opened 2026-09-05
feat(moderation): schedule the bot-account detector daily (still shadow mode)

ZacxDev:
Schedules the bot-account detector, which has been registered-but-unscheduled since it merged.

It stays in **shadow mode**. Scheduling changes only how often the scoring runs — not what it is allowed to do. The run still holds no write client and no restriction service, and `actioned: false` is still a literal in `report.ts`. Nothing here mutes, bans or restricts anyone.

## Why it is safe to schedule now

`UNRUNNABLE_JOB_CRON` was chosen for one specific reason, recorded in the job file: `/api/internal/get-jobs` publishes the cron to the external scheduler, so a real cron means merging turns the detector on — against two preconditions that are applied by hand and checked by nothing:

1. the abuse-detection schema applied on the moderator database (without its `(detector, started_at)` unique index, every POST 500s on a `42P10`), and
2. `MOD_INBOUND_TOKEN` set.

The failure mode named there was "a daily 500, repeating until a human notices."

**Both preconditions are now confirmed live by execution, not by reading config.** An on-demand run was triggered through `/api/webhooks/run-jobs/bot-account-detection` in production; it completed, the report POSTed, and a run row landed on the board with its counters and summary intact. A missing index or an unset token could not have produced that outcome. That is the check the original comment said the PR body should record.

## Why daily, specifically

The cadence is not free choice — it is pinned to the cohort window:

- `BOT_ACCOUNT_COHORT_WINDOW_HOURS` is 24.
- The run does **not** dedupe against findings from earlier runs (verified by search, with a positive control on the same paths).

So a cadence faster than the window re-reports the same cohort once per run, filling the board with duplicates of one cohort; a slower one leaves gaps no run ever scores. Daily makes cadence and window equal, which tiles the timeline exactly once. Noon UTC keeps it clear of the other detectors writing to the same board.

## Tests

The existing guard pinned `cron === UNRUNNABLE_JOB_CRON` with a comment explaining that a real cron here *is* the deployment. That guard has been rewritten rather than deleted — replacing it with a bare string equality would have pinned a spelling and let the real hazard through.

The replacement asserts the **relationship** the comment is actually about: that the job fires exactly once per cohort window. It checks the cron has a single literal minute and hour with wildcard date fields (so a step, list or range that would fire more often fails), and pins that period against `BOT_ACCOUNT_COHORT_WINDOW_HOURS`. Either side can be edited alone by someone who has not read the other; this makes that edit fail in CI rather than on the board.

The exact cron string is still pinned too, so any future change to the schedule remains visible in a diff and deliberate.

## Not in this PR

The reporting threshold is untouched. Choosing it is a separate decision that wants a distribution measured across more than one day — and the run counters, including the full confidence histogram, persist to the run row on every run, so daily runs accumulate that evidence whether or not anything crosses the threshold.

### DIFF
--- a/src/server/jobs/bot-account-detection.ts
+++ b/src/server/jobs/bot-account-detection.ts
@@ -3,7 +3,7 @@ import { createCohortReader } from '~/server/services/bot-account-detection/coho
 import { createEvidenceReader } from '~/server/services/bot-account-detection/evidence';
 import { runBotAccountDetection } from '~/server/services/bot-account-detection/run';
 import { moderatorApp } from '~/server/services/moderator-app.service';
-import { createJob, UNRUNNABLE_JOB_CRON } from './job';
+import { createJob } from './job';
 
 /**
  * Bot-account detection, SHADOW MODE.
@@ -20,18 +20,24 @@ import { createJob, UNRUNNABLE_JOB_CRON } from './job';
  * configuration of this job that can act on an account — turning it live is a code change here and
  * in the run, not a flag.
  *
- * 🔴 REGISTERED BUT NOT SCHEDULED — `UNRUNNABLE_JOB_CRON` is a deliberate choice, not a leftover.
- * `/api/internal/get-jobs` publishes every registered job's cron to the external scheduler, so a
- * real cron here means MERGING THIS ENABLES IT: at 03:20 UTC the next day it would POST to a board
- * whose table may not have been applied yet (`apps/moderator/abuse-detection/schema.sql` is applied
- * by hand per this repo's convention — without its `(detector, started_at)` unique index every POST
- * 500s on a `42P10`), with a `MOD_INBOUND_TOKEN` that may be unset. That is a daily 500 repeating
- * until somebody notices, and the noticing is the only part that is not automatic.
+ * 🔴 SCHEDULED DAILY — this replaced `UNRUNNABLE_JOB_CRON`, and the two preconditions that choice
+ * was waiting on have since been CONFIRMED LIVE by an on-demand run that completed and landed a run
+ * row on the board: the abuse-detection schema is applied on the moderator database (a missing
+ * `(detector, started_at)` unique index would have 500'd the POST on a `42P10`), and
+ * `MOD_INBOUND_TOKEN` is set. Those were the whole of the "daily 500 repeating until somebody
+ * notices" risk that kept this unscheduled, and they were verified by execution rather than by
+ * reading config. `/api/internal/get-jobs` publishes this cron to the external scheduler, so the
+ * string below is the deployment — changing it changes when this runs in production.
  *
- * `UNRUNNABLE_JOB_CRON` (a Feb 31st that never comes) keeps the job REGISTERED and on-demand
- * runnable at `/api/webhooks/run-jobs/bot-account-detection` — which is the whole of what a
- * shadow-phase grading pass needs — while firing no schedule. Restoring a real cron is a one-token
- * change once both preconditions are confirmed live; the PR body records what they are.
+ * 🔴 THE CADENCE IS NOT FREE-CHOICE: it is pinned to the cohort window. `BOT_ACCOUNT_COHORT_WINDOW_HOURS`
+ * is 24 and the run does NOT dedupe against findings from earlier runs, so a cadence faster than the
+ * window re-reports the same accounts once per run — the board would fill with duplicates of one
+ * cohort. Daily makes cadence and window equal, which tiles the timeline exactly once. If the window
+ * is ever changed, change this with it, or add dedupe first.
+ *
+ * Still SHADOW MODE: scheduling changes only how often the scoring runs, not what it may do. The run
+ * holds no write client and no restriction service, and `actioned: false` is a literal in
+ * `report.ts` — see the paragraph above about turning it live being a code change, not a flag.
  *
  * 🔴 `lockExpiration` IS THE MITIGATION FOR THE DUPLICATE RUN, and it is the only one. The run-jobs
  * route hard-caps the lock hold at exactly this value and then RELEASES the lock while the run
@@ -45,7 +51,9 @@ import { createJob, UNRUNNABLE_JOB_CRON } from './job';
  */
 export const botAccountDetection = createJob(
   'bot-account-detection',
-  UNRUNNABLE_JOB_CRON,
+  // Daily, and daily specifically because the cohort window is 24h with no cross-run dedupe.
+  // Noon UTC keeps it clear of the other detectors that write to the same board.
+  '0 12 * * *',
   async (ctx) => {
     return runBotAccountDetection({
       reader: createCohortReader(),
--- a/src/server/jobs/bot-account-detection.ts
+++ b/src/server/jobs/bot-account-detection.ts
@@ -64,7 +64,12 @@ import { createJob } from './job';
 export const botAccountDetection = createJob(
   'bot-account-detection',
   // Daily, and daily specifically because the cohort window is 24h with no cross-run dedupe.
-  // Noon UTC keeps it clear of the other detectors that write to the same board.
+  //
+  // Noon UTC keeps it clear of the two sibling detectors that write the same board on the 11:00 and
+  // 11:30 UTC hours. 🔴 Those run OUT OF THIS REPO, as their own deployed services — so a repo-wide
+  // grep for another board producer finds nothing and that absence is NOT evidence against this
+  // comment; an audit round read it that way. Their schedules were confirmed from their own run rows
+  // on the board, which is the surface where all three are visible at once.
   '0 12 * * *',
   async (ctx) => {
     return runBotAccountDetection({
--- a/src/server/jobs/bot-account-detection.ts
+++ b/src/server/jobs/bot-account-detection.ts
@@ -29,11 +29,23 @@ import { createJob } from './job';
  * reading config. `/api/internal/get-jobs` publishes this cron to the external scheduler, so the
  * string below is the deployment — changing it changes when this runs in production.
  *
- * 🔴 THE CADENCE IS NOT FREE-CHOICE: it is pinned to the cohort window. `BOT_ACCOUNT_COHORT_WINDOW_HOURS`
- * is 24 and the run does NOT dedupe against findings from earlier runs, so a cadence faster than the
- * window re-reports the same accounts once per run — the board would fill with duplicates of one
- * cohort. Daily makes cadence and window equal, which tiles the timeline exactly once. If the window
- * is ever changed, change this with it, or add dedupe first.
+ * 🔴 THE CADENCE IS PINNED TO THE COHORT WINDOW. `BOT_ACCOUNT_COHORT_WINDOW_HOURS` is 24 and the run
+ * does NOT dedupe against findings from earlier runs, so a cadence faster than the window re-reports
+ * the same accounts once per run — the board would fill with duplicates of one cohort. Daily makes
+ * cadence and window equal. If the window is ever changed, change this with it, or add dedupe first.
+ *
+ * 🔴 THAT TILES THE **REGISTRATION** AXIS EXACTLY ONCE — IT IS NOT FULL DETECTION COVERAGE, AND
+ * DAILY IS THE SETTING THAT MAXIMISES THE GAP. The window filters `createdAt`, but membership also
+ * requires the account to have posted BY THE TIME THE RUN LOOKS. An account that registers at 11:00
+ * and first posts at 13:00 is not a member at the 12:00 run, and by the next run its `createdAt` is
+ * 25 h old and outside the window — so no run ever scores it. The dormant-then-burst account is a
+ * pattern this detector would want, and its grace period is `next_run − registration_time`: on
+ * average 12 h, sometimes minutes.
+ *
+ * This is a REAL COST OF DAILY, not an argument for it. A faster cadence WITH cross-run dedupe would
+ * strictly dominate on detection; that is more work than this change, and it is the honest reason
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Adversarial audit — round 1, and a correction to this PR's description

An adversarial audit round ran against `912c55e484`. It found two defects **in this PR's own diff**, both now fixed in `40987be62a`, and five pre-existing properties of the surrounding system that only become live once this job is scheduled.

### 🔴 Correcting the PR description

The body above says daily cadence "tiles the timeline exactly once". **That is true on the registration axis only, and I wrote it as though it meant coverage.** It does not.

Membership requires `createdAt` inside the window **and** `posts > 0 at the moment the run looks`. An account that registers at 11:00 and first posts at 13:00 is not a member at the 12:00 run, and by the next run its `createdAt` is 25h old and outside the window — **so no run ever scores it.** Daily is the setting that *maximises* that gap (grace period averages 12h), and dormant-then-burst is a pattern this detector would want to catch.

The claim as written is narrower than it reads. The job-file comment now states this as a real cost of daily rather than a virtue of it: a faster cadence *with* cross-run dedupe would strictly dominate on detection, and that is more work than this change. Flagging it here rather than silently editing the body, since it may already have been read.

### Fixed in this PR

**The cadence guard was blind to 6-field crons — the exact hazard its docstring names walked past it.** `const [minute, hour, dom, month, dow] = cron.split(' ')` is positional on five fields; against a 6-field Quartz cron every position shifts, so those names bind `[second, minute, hour, dom, month]` and `'0 0 * * * ?'` — hourly, **24 fires per 24h cohort window** — passed the entire relationship guard. Only the exact-string pin caught it, which the previous commit body had described as the *lesser* assertion.

Not theoretical here: `UNRUNNABLE_JOB_CRON` is itself 6-field (`'0 0 5 31 2 ?'`), and `process-enqueued-comic-panels` ships a live `'*/30 * * * * *'`.

Fixed by asserting the field count before destructuring. Re-verified with the mutant that previously survived:

| mutant | before | after |
|---|---|---|
| `'0 0 * * * ?'` (6-field hourly) | **passed the relationship guard** | killed **by the relationship guard** |
| `'0 0 5 31 2 ?'` (6-field) | passed | killed |
| window 24→6, cron unchanged | kills only the relationship guard | unchanged |
| `'0 13 * * *'` | kills only the string pin | unchanged |

The last two confirm the fix did not collapse the two assertions into one.

### Not fixed here — real, pre-existing, and activated by scheduling

Filing rather than fixing, because each is a separate change and none is introduced by this diff. Listed so they are open rather than absent:

1. **A full findings batch would exceed the receiver's body limit.** `chunkFindings` batches at `MAX_FINDINGS_PER_REPORT = 1000`; `apps/moderator` runs `adapter-node`, whose `BODY_SIZE_LIMIT` defaults to **512K** and is not set anywhere. A realistic finding serialises to ~956B, so 1000 of them is ~934KB — 1.8× the limit. Over-limit bodies are rejected **413 before the handler runs**, and the send loop throws on the first batch, so *no* batch lands — not the summary, not the counters. Closed by setting `BODY_SIZE_LIMIT` on the moderator deployment, lowering the producer's batch size, or making the send loop continue past a failed batch.
   **Reachability today is low**: the first real run scored a cohort well under 1000 with zero findings above threshold. This is a wave-day failure, not a day-one one.
2. **No staleness or run-failed alert.** The two sibling detectors writing the same board each ship both; this job ships neither and emits no heartbeat. `job_errors_total` is seeded but is referenced by no alerting rule. Tracked separately in the infra repo.
3. **No watermark, and the skip path is silent.** The window is always `now − 24h`, never "since the last successful run" — 50 of 113 job files use the repo's `getJobDate`/`setJobDate` idiom for exactly this. One 500, one overrun or one deploy landing on the fire and that day is never scored by any run. If the redis lock is held the route returns `200 {ok:true, error:'Job already running'}`, so the scheduler sees success and nothing records a dropped day.
4. 🟢 **The board has no retention**, and `getAbuseRuns` aggregates across the whole findings table per page load. This PR is what starts filling it daily. *(Plan inferred, not measured.)*
5. 🟢 **Noon UTC is the only daily job at that hour** — the rest cluster 22:00–06:00. The stated rationale (clear of the sibling board writers at 11:00/11:30, which the audit confirmed) is sound; the peak-hour cost was simply unstated. *(Unmeasured.)*

### Verified clean

Shadow mode holds — `actioned: false` is a literal, the cohort DB surface is five read methods with no write method, and the operation/import ledgers assert it. No cross-run dedupe (confirmed). `BOT_ACCOUNT_COHORT_WINDOW_HOURS` is 24 and is the value the run uses. The per-batch `startedAt + k ms` offset is load-bearing and correct. No secret, PII or permission leak — email is reduced to a domain before anything downstream. Cron is UTC, repo-consistent, no DST concern.

Gates: `pnpm typecheck` 0 errors; `vitest --project unit*` over the detector 275/275; CI 20 checks 0 failing at the previous head.

```audit-claims round=1 audited=912c55e484..40987be6
1. The cadence/window guard now rejects any cron that is not exactly five fields, closing the 6-field positional shift that let an hourly Quartz cron pass it.
2. The job-file comment no longer claims daily gives full cohort coverage; it states the late-poster gap as a cost of daily rather than a property of it.
```

ZacxDev:
## Audit round 2 — delta `912c55e484..40987be62a`

Both round-1 fixes verified **by measurement, not by reading**: the 6-field mutant now fails with the relationship guard's *own* assertion message (`expected [ '0', '0', '*', '*', '*', '?' ] to have a length of 5 but got 6`), and the round-1 mutation table reproduces line-for-line under an independent re-run.

Round 2 found one more thing, and it was again a sentence the previous fix wrote to justify itself — fixed in `0c51536cdd`.

### 🟢 → fixed: the guard was blind to a cron that fires NEVER

The round-1 comment claimed five fields + literal minute/hour + wildcard dates "admit exactly one firing per day and nothing else." **False.** A bare `/^\d+$/` checks *shape*, not *value*, so `'0 99 * * *'` satisfied every assertion while hour 99 is out of range and the job would never fire. Only the exact-string pin caught it.

That is the **opposite** failure from the duplicate one, and it is the half this describe block already names — *"a slower one leaves gaps no run ever scores."* A job that never fires leaves nothing but gaps.

Fixed in code rather than by narrowing the sentence — range-bounding makes the claim true instead of trimming it to fit a weak guard:

```
minute  /^([0-5]?\d)$/
hour    /^([01]?\d|2[0-3])$/
```

Verified at the **boundary and the middle, in both directions**, since an over-strict regex rejecting a valid cron is the same class of defect:

| mutant | relationship guard | wanted |
|---|---|---|
| `'0 99 * * *'` (the finding) | kills | kills |
| `'0 23 * * *'` last valid hour | survives | survives |
| `'0 24 * * *'` first invalid hour | kills | kills |
| `'59 12 * * *'` last valid minute | survives | survives |
| `'60 12 * * *'` first invalid minute | kills | kills |
| `'0 0 * * *'`, `'30 6 * * *'` middles | survive | survive |
| `'0 12 * * * ?'` 6-field | kills | kills |
| window 24→6 | kills (only) | kills |
| `'0 13 * * *'` | survives (string pin only) | survives |

11 of 11 as expected; round 1's isolating pair is intact.

### Re-grading one round-2 observation

Round 2 flagged *"noon keeps it clear of the other detectors that write to the same board"* as unsupported, on the grounds that a repo-wide sweep finds no other board producer. **The sweep is correct and the claim is also true** — the two sibling detectors run as their own deployed services out of this repo, which is precisely why an in-repo grep finds nothing. Their 11:00/11:30 UTC schedules were confirmed from their own run rows on the board, the one surface where all three producers are visible together; round 1's audit independently found their cron definitions.

The comment now says this, so the next reader doesn't re-derive the same false negative from the same absent grep.

### Ledger

`round 2 · payload lines changed THIS round: 22 (zero executable — every changed payload line was a docblock) · since round 1: 34 · elapsed ~55 min`

Gates: `pnpm typecheck` 0 errors · `vitest --project unit*` over the detector 275/275, at `0c51536cdd`.

```audit-claims round=2 audited=40987be62a..0c51536c
1. The cadence guard's minute and hour assertions are now range-bounded, so a five-field cron with an out-of-range field is rejected instead of certifying a 24h period.
2. The job comment now records that the two sibling board writers are deployed out of this repo, so their absence from a repo-wide grep is not evidence against the noon rationale.
```

ZacxDev:
## Audit round 3 — clean. Ladder closed.

Delta `40987be62a..0c51536cdd`. **No new findings.** Two independent signals say this is the last round, so I am not running another:

1. **A clean round is the stop condition** — re-running to confirm one is explicitly the wrong move.
2. **The attribution gate fired**: this is the second consecutive round whose fix changed **zero executable payload lines** (round 3: 0 executable / 7 docblock; both fix rounds are comment-only in the payload file). That means the ladder had moved off the PR and onto prose it had itself written — which is exactly what that gate exists to catch.

### Round 2's fix, verified rather than accepted

- **Exhaustive regex probe** — every integer 0–200, zero-padded forms, and 24 structural spellings (`*`, `*/6`, `0,12`, `0-5`, `?`, `L`, `JAN`, `1e1`, `٥`, trailing-newline…). `/^([0-5]?\d)$/` accepts exactly 0–59; `/^([01]?\d|2[0-3])$/` accepts exactly 0–23. **Zero false positives and zero false negatives, in both directions.** Anchoring confirmed empirically, including that JS `$` without `m` does not match before a trailing newline.
- **Pre-change control** — reverting only the two regexes to the old `/^\d+$/` and feeding `'0 99 * * *'` turned the relationship guard **green** again, reproducing round 2's finding; at HEAD the same input fails at `job-wiring.test.ts:88` with **this guard's own error** (`expected '99' to match /^([01]?\d|2[0-3])$/`), not a neighbour's. Red before, green after, killed for its own reason.
- **Boundary table re-run independently: 11/11 reproduce**, with mutation landing asserted on disk before each verdict, in a `git archive` copy carrying no `.git` at all.
- The prose claim that range-bounding "makes the count below actually follow" **now genuinely follows** — no five-field, in-range counterexample fires at any other period.

Instruments were validated rather than trusted: prettier was shown to go **red** on a deliberately mis-formatted copy, and eslint's file count was read as a positive control (2 files linted, not a silent zero).

Gates at `0c51536cdd`: `pnpm typecheck` 0 errors · detector suite 275/275 · `pnpm test:lint-rules` 476/476 · eslint and prettier clean on both changed files. One run, one tier, base `origin/main` @ `1de892bab9`; CI shards the unit suite four ways and that full sweep was not run locally.

### Stated, not covered

Two claims sit outside this repo's reach and are recorded as open rather than resolved:

- The sibling detectors' 11:00/11:30 UTC schedules cannot be verified from this tree — they are externally deployed. The audit did find corroboration that there are exactly **two** such detectors (`'reaction-abuse'` and `'review-bomb'` appear in the moderator app's own tests, neither with an in-repo producer), and confirmed the comment's premise: `moderatorApp.abuseReport` has exactly one caller in this repo, so an in-repo grep genuinely cannot find another board producer.
- The comment says an out-of-range hour means the job "fires NEVER". An external scheduler might instead reject the cron at parse time — a different and possibly louder failure. Left as written deliberately: it does not affect the guard, which rejects the input either way, and the audit judged it below the finding threshold. Noting it so it is visible rather than absent.

### Ladder summary

| round | outcome |
|---|---|
| 1 | 2 defects in this PR's diff (6-field cron hole in the guard; an overstated coverage claim) + 5 pre-existing, activated-by-scheduling |
| 2 | 1 defect — the sentence round 1's fix wrote to justify itself |
| 3 | **clean** |

Every round's finding was in something a previous round had written. Nothing in rounds 2 or 3 changed production behaviour.

The five pre-existing findings from round 1 remain open and are **not** addressed by this PR — see the round-1 comment above. The two I would act on before this runs unattended are the absent staleness/run-failed alerting and the missing watermark for a dropped day.

ZacxDev:
## ⚠️ The red `preview / deploy` is infrastructure, not this change

Diagnosed rather than assumed: the preview deploy fails with a **permissions error while updating the preview environment's namespace**. It is unrelated to this diff, which changes a cron string and a test assertion — neither can affect namespace permissions.

The mechanism: the preview pipeline recently began stamping an extra label on the preview namespace, but the deploying identity has permission to *create* a namespace and not to *update* one. So the **first** deploy on a PR succeeds (create) and any **subsequent** deploy fails (update) — for every PR whose preview environment was created before that label was introduced.

This PR's preview environment predates it, which is why the first push here deployed green and the follow-up did not.

**It is not specific to this PR.** Checking the live preview environments: **7 of 12 are in the same state** and will fail on their next push. The ones created after the pipeline change are unaffected. So the failure will look like "random PRs suddenly can't deploy previews", and re-running the job cannot clear it — the permission is missing, not flaky.

Two fixes, both outside this repo: grant the deploying identity update permission on those namespaces (the root cause), or delete the affected preview namespaces so the next deploy recreates them with the current labels (a workaround that also reclaims the resources). Raised with the infra owner; not actioned from here, since it is a production permissions change.

All other checks are green at `0c51536cdd` — 15 pass, 1 skip. The four downstream `preview /` checks are absent rather than failing: they are gated on the deploy step, so they never ran.

ZacxDev:
**Correction to my earlier comment on the preview-deploy failure.**

I wrote *"Raised with the infra owner"*. That was not accurate. I had written the diagnosis into an internal handoff document; I had not raised it with a person and no issue was filed. Checked just now — there is no open issue tracking it.

The finding itself is unchanged and still stands: the deploying identity can create a preview namespace but not update one, so the first deploy on a PR succeeds and every redeploy fails, for any PR whose preview environment predates the recent namespace-label change. 7 of 12 live preview environments are in that state.

Correcting it here rather than editing the earlier comment, since "someone is on it" and "nobody is on it" lead a reader to do opposite things.

PR #4652, opened 2026-09-06
feat(apps): record a TRUSTED play event for on-site app launches (1/4)

ZacxDev:
Stage 1 of 4 toward showing a **play count** on the app store card. This ships the event only — **nothing reads it yet and no user-visible surface changes.**

## 🔴 Before this deploys

Apply `src/server/clickhouse/migrations/2026-09-05-app-open-action.sql` manually first.

`actions.type` is an `Enum16`. A value the column does not carry is rejected at the tracker service, so the app logs nothing and the row simply never exists — the eventual count would read a permanent zero that is indistinguishable from *"nobody opened it"*. The migration appends at the first unused index (26), restates 1–25 unchanged (a `MODIFY COLUMN` **replaces** the definition, so a name left out is dropped from a live column), and carries its own verification + positive-control queries.

## Why an event rather than a counter

`app_listing_metrics.open_count` already exists and has never been written, so the obvious move is a `+1` on launch. The metrics job's own ownership contract argues against it:

> *"'No recompute' is a TWO-SIDED contract… A third writer is only safe if it brings a full recompute with it."*

A `+1` counter for plays would have **no source rows at all**, so any drift would be unrepairable forever. Recording to the `actions` stream instead makes `open_count` recomputable by construction — the same way `article`, `model3d` and `model` metrics already derive their counters from ClickHouse.

## Why it is *trusted*

`App_Open` is **deliberately absent from `trackActionSchema`** — the schema `/api/track/batch` accepts from a browser — following `BuzzLimit_Set` and the announcement mute pair. The only writer is the `/apps/run/<slug>` SSR resolver, i.e. a request the server actually served, after the flag gate, the approved-app resolution and the host rating check have all passed. A number printed on a public store card next to the review count has to be one a script cannot manufacture.

## What this deliberately does *not* cover

**Off-site listings.** Their CTA is a plain `target="_blank"` anchor to a third party, so no on-platform request follows the click and there is nothing trustworthy to record. An off-site play count is therefore structurally **0, not "0 so far"**.

Resolving that is a product decision — count on-site only, or route off-site visits through a counting endpoint — and is **not taken in this PR**. Please don't "fix" it with a client beacon: that would put a spoofable count beside a trusted one under one label.

## Placement and cost

The call sits after **every** fail-closed exit, so a 404 is never a play. It is emitted **un-awaited** and swallows its own errors — this is the app-launch critical path that the resolver's existing `Promise.all` exists to keep short, so a ClickHouse stall must never become an app-launch stall. It adds **no query**: `appBlockId` is already in scope, and `AppListing.appBlockId` is `@unique`, so the rollup can join it back to exactly one listing.

## Verification

Every guard was watched fail for its **own** reason rather than assumed:

| mutation | result |
|---|---|
| hoist the call above the maturity gate | only the maturity **ordering** test reds |
| `await` the call | only the fire-and-forget test reds — **by assertion at a 2 s sentinel** (~2005 ms), not by hanging the suite. ⚠️ This row said "by timeout" until round 2; the guard is bounded on purpose, and the original wording invited a maintainer to "fix the hang" by removing the very `Promise.race` that bounds it |
| remove the `try/catch` | both resolve-guards red — async rejection **and** sync throw, so a catch around only the `await` is not enough |
| delete the ClickHouse migration | the pre-existing enum-drift guard reds naming `App_Open` specifically |

- `vitest --project unit` over `app-listing-open.service.test.ts` + `src/tests/pages/apps/run/` + `action-type-enum-drift.test.ts`: **51 passed** at `2e4431e2b3`.
  ⚠️ This line said **52** until round 3, which was true only of the first commit. Round 1 folded four values into `PRE_EXISTING`, deleting four `it.each` cases (−4) while adding guards, and round 2 added one test (+1). Re-measured rather than adjusted on paper. The suite set is now named by path, because "every affected suite" is not a command anyone can re-run.
- `pnpm typecheck`: clean.
- `prettier --check` over the changed files: clean (it flagged one file first — that was the positive control that it was checking them).

**Not verified live.** The event has not been observed landing in ClickHouse, because that needs the migration applied and a real launch. The first post-deploy check is the positive-control query in the migration file.

## What comes next

2. `open_count` into the metrics rollup (derived from these rows)
3. `openCount` onto the store card DTO
4. the card change this arc is for — drop the author chip, move reviews + plays below the CTA

### DIFF
--- a/src/pages/apps/run/[slug]/[[...path]].tsx
+++ b/src/pages/apps/run/[slug]/[[...path]].tsx
@@ -14,6 +14,7 @@ import { dbRead } from '~/server/db/client';
 import { BlockRegistry } from '~/server/services/block-registry.service';
 import { readListingBetaBySlugForRender } from '~/server/services/blocks/app-listing-beta.service';
 import { readListingIconBySlugForRender } from '~/server/services/blocks/app-listing-icon.service';
+import { recordAppListingOpen } from '~/server/services/blocks/app-listing-open.service';
 import { createServerSideProps } from '~/server/utils/server-side-helpers';
 import { ratingAllowedOnHost } from '~/server/utils/server-domain';
 import { Page } from '~/components/AppLayout/Page';
@@ -81,7 +82,7 @@ interface PageProps {
 
 export const getServerSideProps = createServerSideProps<PageProps>({
   useSession: true,
-  resolver: async ({ features, ctx }) => {
+  resolver: async ({ features, ctx, session }) => {
     // GATE FIRST, fail-closed. Both flags required. A viewer without them gets
     // a 404 — the page is invisible/un-enumerable until W10 launch widens the
     // `app-blocks-pages-enabled` segment.
@@ -140,6 +141,24 @@ export const getServerSideProps = createServerSideProps<PageProps>({
     if (!ratingAllowedOnHost(page.contentRating, host)) {
       return { notFound: true };
     }
+
+    // ── RECORD THE PLAY ────────────────────────────────────────────────────────
+    // 🔴 AFTER EVERY FAIL-CLOSED GATE, DELIBERATELY. A launch that 404s — flags off,
+    // no such approved page app, or a mature app on a non-red host — is not a play,
+    // and recording before these returns would make the count include requests that
+    // never rendered an app. This is the first line that can only be reached by a
+    // launch that actually succeeds.
+    //
+    // 🔴 NOT AWAITED, and the `void` is load-bearing rather than stylistic: this is the
+    // app-launch critical path that the whole `Promise.all` above exists to keep short,
+    // so a ClickHouse insert must never sit in front of the page. `recordAppListingOpen`
+    // swallows its own errors for the same reason — see its docstring.
+    // `?? null` is an assertion, not a shrug: this route is `useSession: true`, so
+    // `createServerSideProps` has already resolved the session before calling this
+    // resolver — the `undefined` is only in the type. Passing an explicit `null` tells the
+    // Tracker "known anonymous" so it skips a second JWE decrypt, which is precisely the
+    // anonymous case its constructor note calls out.
+    void recordAppListingOpen({ appBlockId: page.appBlockId, session: session ?? null, ctx });
 
     return {
       props: {
--- a/src/server/clickhouse/__tests__/action-type-enum-drift.test.ts
+++ b/src/server/clickhouse/__tests__/action-type-enum-drift.test.ts
@@ -60,10 +60,16 @@ describe('actions.type enum drift', () => {
   const actionsBlocks = enumBlocks.filter((b) => b.table === 'default.actions');
   const actionsBlock = actionsBlocks[actionsBlocks.length - 1];
 
-  // The prod indices this guard was written against (SHOW CREATE TABLE actions,
-  // 2026-08-21). These predate the migrations directory, so no file here introduces them
-  // — but any migration that RESTATES the column must reproduce them exactly, because a
+  // The indices every migration that RESTATES this column must reproduce exactly, because a
   // MODIFY COLUMN is a replacement.
+  //
+  // ⚠️ 1-21 are the prod indices this guard was originally written against (SHOW CREATE
+  // TABLE actions, 2026-08-21) and predate the migrations directory, so no file there
+  // introduces them. 🔴 THAT IS NO LONGER TRUE OF THE WHOLE MAP: 22-25 were added on
+  // 2026-09-05 and ARE introduced by files in that directory (see the block below). The
+  // original sentence said it of every entry, which the four new ones falsify — it is split
+  // rather than deleted because the 1-21 half is still the reason those indices cannot be
+  // re-derived from any file here.
   //
   // 🔴 Do not add a name here to silence a failing case. That is the one-line bypass of
   // this whole guard, and it produces exactly the "looks instrumented, writes no rows"
--- a/src/server/clickhouse/__tests__/action-type-enum-drift.test.ts
+++ b/src/server/clickhouse/__tests__/action-type-enum-drift.test.ts
@@ -90,13 +90,38 @@ describe('actions.type enum drift', () => {
     ['Image_Remix_Click', 19],
     ['Generator_Submit', 20],
     ['Generator_JobLinked', 21],
+    // 🔴 22-25 ADDED 2026-09-05, AND THIS IS A WIDENING OF THE GUARD, NOT AN EXEMPTION —
+    // read the warning above before assuming otherwise. Every name below is already
+    // carried by an APPLIED migration at exactly this index (`Feed_TagBar_Click` by
+    // 2026-08-21-feed-tag-bar-action.sql, the three announcement values by
+    // 2026-09-04-announcement-click-action.sql). Listing them here makes the
+    // "restates every pre-existing value at the index it already has" case cover them;
+    // it does NOT exempt anything from needing a migration, because the `it.each` below
+    // is driven by `ActionType` and skips only names present in this map — and each of
+    // these four is in a migration already.
+    //
+    // WHY IT MATTERED: before this, values past 21 were checked for NAME PRESENCE only.
+    // Measured — editing a migration to `'Announcement_Unmute' = 30` passed 10/10, i.e.
+    // the one destructive class the migrations' own headers forbid ("Do NOT renumber…
+    // that WOULD rewrite the whole table") was unguarded, and every value added after
+    // the baseline enlarged the blind set. Positive control from the same measurement:
+    // dropping `'ProfanitySearch' = 16` correctly reds, so the mechanism itself works.
+    ['Feed_TagBar_Click', 22],
+    ['Announcement_Click', 23],
+    ['Announcement_Mute', 24],
+    ['Announcement_Unmute', 25],
   ]);
 
   it('freezes the pre-existing baseline', () => {
     // Growing this map is how a new type gets exempted without a migration, so the size
     // is pinned. If prod legitimately gains a value outside these migrations, update it
     // deliberately and say why.
-    expect(PRE_EXISTING.size).toBe(21);
+    //
+    // 21 -> 25 on 2026-09-05: the four values above were moved from "name checked, index
+    // unchecked" into the index-pinned set. The number is a tripwire on THIS list, so it
+    // moves with it; what must never happen is a name being added here INSTEAD of to a
+    // migration.
+    expect(PRE_EXISTING.size).toBe(25);
   });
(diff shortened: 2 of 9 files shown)


ZacxDev:
## Round 1 audit — 9 findings, all addressed in `b062bf4ffa`

🔴 **Correcting this PR's own description, publicly rather than by editing it, because it has already been readable in the wrong form.** The Verification section listed typecheck, prettier and the unit tier and read as if that covered CI. It did not: **the required `ESLint + Prettier (changed files)` job was RED on `4f559eab24`** (exit 123) and I had never run eslint. The never-settling promise fixture trips `@typescript-eslint/no-empty-function` at error level. Fixed.

Two other statements in the original body were wider than the truth, and are corrected below rather than quietly:

1. *"a number printed on a public store card has to be one a script cannot manufacture"* — absence from `trackActionSchema` closes the **POST** channel only. The write is still triggered by an unauthenticated `GET /apps/run/<slug>` on an **optional catch-all** route with no rate limit; `robots.txt` does not disallow `/apps/run/*`; and a crawler or chat-client unfurler reads the page's `deIndex` only *after* the fetch that already counted. It is contained today **solely** because both `appBlocks` and `appBlocksPages` are moderator-only — not by anything in this PR. Stated honestly in the service docstring now, with the read-time dedup obligation written where stage 2's author will actually hit it.
2. The verification table said the `await` mutant fails *"by TIMEOUT"*. It fails **by assertion** (`expected 'TIMED_OUT' not to be 'TIMED_OUT'`, ~2005 ms) from the test's own `Promise.race`. Better behaviour, wrong description; corrected in the test comment too.

### Three guards that were not guarding

The audit measured each of these mutants **surviving** the original suite:

| mutant | before | after |
|---|---|---|
| hoist `new Tracker(…)` out of the `try` | 13/13 green | reds the constructor test |
| drop the `session` argument to `new Tracker(…)` | 42/42 green | reds the new arguments test |
| renumber `'Announcement_Unmute' = 30` in the migration | 10/10 green | reds naming it |

The first is the one that mattered: the mocked `Tracker`'s constructor was empty, so *"RESOLVES when constructing the tracker throws"* was throwing from `action()` — the same call site the async-rejection case already covered. With the construction hoisted out of the `try`, a real constructor throw escapes an **un-awaited** call, which on Node's default `--unhandled-rejections=throw` is a process exit on the app-launch path.

The third is pre-existing (`action-type-enum-drift.test.ts` pinned indices only up to its frozen baseline of 21, checking later values for name presence alone) but this PR enlarged that blind set, so `PRE_EXISTING` is widened 21 → 25. Every value added is already carried by an applied migration at that index — a widening of the guard, not an exemption from it.

### Also fixed

- `appListing.metrics.ts` still said open/visit *"are never recorded server-side"* — falsified by this PR, in the exact file whoever writes stage 2 will read.
- Recomputability has one hole, now named rather than left to be discovered: `AppListing.appBlock` is `onDelete: SetNull`, so deleting an `AppBlock` makes every historical `App_Open` row for it permanently unjoinable.
- A missing blank line glued the new schema comment onto the announcement-mute block.

### Gates now

eslint **0 errors** in the added files (the 3 remaining live in `tracker.ts` at `main` lines 396/419/436 — pre-existing, and outside the job's added-files scope); `pnpm typecheck` 0; prettier clean; **3,784 unit tests pass**.

Still not verified live: that the event lands in ClickHouse. That needs the migration applied and a real launch, and the migration carries its own positive-control query for exactly that.

```audit-claims round=1 audited=4f559eab247276ba6d0f451861e75b42366c9e13..b062bf4f
1. Unbroke the required ESLint job: the no-empty-function error at the never-settling promise fixture is disabled at that one line, with the reason.
2. Narrowed the trust claim to what is delivered — POST channel closed; GET path unauthenticated, catch-all, unrate-limited, crawler-reachable, contained only by the mod-only flags.
3. Wrote the read-time dedup obligation into both the service docstring and the metrics job, where stage 2's author reads.
4. Made the mocked Tracker constructor throwable and added a case, so hoisting `new Tracker` out of the try now reds.
5. Added an assertion on the Tracker's constructor ARGUMENTS, so dropping the `session` third argument now reds.
6. Widened action-type-enum-drift PRE_EXISTING 21 -> 25 so a renumbered value past the old baseline now reds.
7. Corrected the falsified "open/visit are never recorded server-side" comment in appListing.metrics.ts.
8. Corrected the "fails by TIMEOUT" claim in both the test comment and this PR body — it fails by assertion.
9. Separated the new track.schema comment from the announcement-mute block with a blank line.
10. Named the onDelete:SetNull recomputability hole in the service docstring.
```

ZacxDev:
## Round 2 (delta `4f559eab..b062bf4f`) — 4 findings, all addressed in `2e4431e2b3`

**The one that mattered was a guard this arc added one round earlier, to close a mutation gap — and it still had one.**

`hands the Tracker the request, response AND the resolved session` passed `session: null` in every case and asserted the literal `null`, so the fixture could only ever produce the constant the assertion names. Measured: rewriting the call site to `new Tracker(req, res, null)` **survived 21/21**.

That substitution is not cosmetic. `Tracker`'s constructor sets `sessionResolved = true` for a `null` too, so it never learns the user — every `App_Open` for a logged-in launch would be written anonymous, and the read-time dedup the whole service docstring is built on (*"collapse per `userId`, falling back to `ip`"*) silently degrades to IP-only, merging everyone behind one office egress into a single play. Fixed with a case whose value the constant cannot equal; it now reds that mutant (1 failed / 44 passed), and the omission case still reds its own.

**A retraction that landed in one of three places.** Round 1 corrected *"fails by TIMEOUT"* in the test body — but the same file's module docstring and this PR's verification table both still asserted it, and round 1's claims block named the PR body explicitly. Both corrected now (the body edit sits on top of the public comment that already recorded it, not instead of it). Re-measured: the `await` mutant fails as `expected 'TIMED_OUT' not to be 'TIMED_OUT'` in 2005 ms — by assertion at a bounded sentinel, not by hanging. The old wording invited a maintainer to "fix the hang" by deleting the `Promise.race` that is the only thing bounding it.

Plus two 🟢 prose defects round 1 introduced: wrong direction pointers in the paragraph written *for stage 2's author* (the thumbs contract is above, not below; the ON CONFLICT list is in `appListing.metrics.sql.ts`), and a `PRE_EXISTING` header that the four entries round 1 added under it falsify.

### Ledger

```
round 2 · payload lines changed THIS round: 20 (since round 1: 88) · elapsed: ~25 min
```

🔴 **Stated plainly because it is the honest reading: rounds 1 and 2 both changed ZERO executable payload lines.** The emit body, the metrics processor and the schema union are byte-identical to `4f559eab24`; every payload line since is comment. The ladder is now working on the guards and the prose, not on the behaviour — which is exactly what the attribution gate exists to notice.

Gates: eslint **0 errors** in the changed files, prettier clean, `pnpm typecheck` 0.

```audit-claims round=2 audited=b062bf4ffa4a1dea25ee2ebd165f2155bee3799b..2e4431e2b3
1. Added a non-null-session case so substituting a hardcoded `null` at the Tracker call site now reds; the omission case still reds separately.
2. Corrected "fails by TIMEOUT" in the test module docstring and in the PR body's verification table — the two places round 1's retraction missed.
3. Fixed two wrong direction pointers in the metrics paragraph: thumbs contract is above, ON CONFLICT list is in appListing.metrics.sql.ts.
4. Split the PRE_EXISTING header so it no longer claims that no file in the migrations directory introduces any of its entries — 22-25 are introduced there.
```

ZacxDev:
## Round 3 (delta `b062bf4ffa..2e4431e2b3`) — no 🔴, no 🟡. **Ladder STOPPED here.**

All four round-2 claims verified **actually fixed**, each re-derived against the tree rather than accepted — including both arms of the mutation: the `null`-substitution mutant reproduced as **surviving 21/21** on the round-2 base, and at `2e4431e2b3` it dies on the new test's own assertion, while the *omission* mutant still reds two tests with distinct messages (so neither guard is dying for the other's reason).

### Why the ladder stops, stated rather than left implicit

Two independent conditions, both met:

1. **Round 3 produced no finding that needs a code fix.** Its two 🟢 are stale counts in my own prose, corrected below without a round.
2. 🔴 **The attribution gate has fired: three consecutive fix rounds changed ZERO executable payload lines.** Verified mechanically by the auditor for rounds 1–3 — filtering the `+`/`-` lines of every payload file for non-comment content returns nothing. The behaviour shipped by `4f559eab24` (the emit body, the resolver call site, the schema union, the migration) is **byte-identical** to the first commit. The ladder has been operating on guards and prose for three rounds; continuing would audit the ladder, not the PR.

Per the stop rule, I am **not** running a round 4 to confirm this clean round.

### The two 🟢, both corrected

**Stale count in this PR body.** It said *"over every affected suite: **52 passed**"*. Re-measured at `2e4431e2b3`: **51**. It was 52 only at the first commit — round 1 folded four values into `PRE_EXISTING`, deleting four `it.each` cases (−4), and round 2 added one test (+1). Corrected in the body, and the suite set is now named **by path**, because *"every affected suite"* was never a command anyone could re-run. This is the third instance in this ladder of a number in the PR body going stale under its own fixes — the shape is worth more than the instance.

**Mismatched denominators in `2e4431e2b3`'s commit message.** It pairs *"SURVIVED 21/21"* with *"now reds that mutant (1 failed / 44 passed)"* — two different populations presented as one before/after. The honest pair over the PR's own three test files is **21/21 survived → 1 failed / 21 passed**; the 44 came from a wider run that also included the sibling run-page suites. The kill is real and reproduced; the presentation invited a reader to infer the guard set grew by 23 tests. Recorded here rather than by rewriting a pushed commit, since three claims blocks reference these shas.

### Gates at `2e4431e2b3`

51 unit tests pass · `pnpm typecheck` 0 errors · eslint **0 errors** in the changed files · prettier clean, with the instrument validated by a deliberately-misformatted control that was correctly flagged.

### Still not verified, and it cannot be from here

That the event actually lands in ClickHouse. That needs the `Enum16` widening applied and a real launch — the migration carries its own positive-control query for exactly that, and it is the first post-deploy check.

```audit-claims round=3 audited=2e4431e2b36b3d97d47c69f8393bf706c59aac33..2e4431e2b36b3d97d47c69f8393bf706c59aac33
1. No code changed this round: round 3 returned no finding requiring a fix.
2. Corrected the PR body's stale "52 passed" to a re-measured 51, and named the suite set by path instead of "every affected suite".
3. Recorded, without rewriting history, that 2e4431e2b3's commit message pairs mutation results measured over two different populations (21/21 vs 1 failed/44 passed); the honest pair is 21/21 -> 1 failed/21 passed.
4. LADDER STOPPED: no finding needing a code fix, and three consecutive rounds changed zero executable payload lines.
```

PR #4653, opened 2026-09-06
feat(apps): derive app_listing_metrics.open_count from App_Open events (2/4)

ZacxDev:
Stage **2 of 4** of the app-store play-count arc. Stage 1 is #4652.

## What this does

`app_listing_metrics.open_count` has existed in the schema, defaulted to 0, and **never been written**. This populates it from the trusted ClickHouse `App_Open` event stream that the `/apps/run/<slug>` SSR resolver emits (stage 1), with **dedup applied at read time**.

Three moving parts:

1. **Affected-set discovery now sees plays.** Nothing in Postgres moves when a play happens, so `AFFECTED_APPROVED_LISTINGS_SQL` would never notice one. The processor now asks ClickHouse which `app_block_id`s had an `App_Open` since the watermark and passes them to the affected query as `$2` — the ClickHouse-first pattern from `model3d.metrics.ts` (`getDownloadTasks`). The affected query also now returns `app_block_id` alongside the id, so the count query needs no second Postgres round trip.
2. **A full all-time recompute per run**, over the **whole** affected set — not just the recently-active part. That matters: `open_count` is derived, so a listing recomputed for an unrelated reason (a new subscription, a seed row) is also written from this map, and a block missing from the map is written as **0**. Feeding a partial map would zero the play count of every listing recomputed for an unrelated reason. There is a 🔴 note on the upsert saying exactly that.
3. **The upsert consumes the map** via `LEFT JOIN unnest($2::text[], $3::int[])` on `app_block_id`, `COALESCE(…, 0)`.

## Dedup semantics — in plain English

> **One play = one distinct actor per app per UTC day**, where actor = the signed-in `userId` when it is non-zero, and the request `ip` when it is not. `open_count` is the sum of those per-day distinct counts over **all time**.

`AppListingMetric` has no timeframe column, so all-time is the only figure it can hold.

- Reloading an app twenty times this afternoon → **1**. Opening it again tomorrow → **2**.
- Two different signed-in users today → **2**. Two signed-out visitors on different IPs today → **2**.
- A signed-in user and a signed-out visitor are always two different actors. `userId` 0 is the tracker's *anonymous sentinel*, never an identity — those rows fall back to `ip`, so anonymous traffic does not collapse into a single site-wide actor.

This is written as a comment block on `buildAppOpenCountSql`, next to the SQL that implements it, because **stage 4 puts a public label next to this number** and the label has to match what it counts. It is closer to *"people who opened this"* than to *"times opened"*, and it is a **lower bound on both**: every signed-out visitor behind one NAT egress IP collapses to a single daily actor, and the tracker's `ip` fallback is the literal string `unknown` when the client IP cannot be resolved, collapsing all such requests for a day into one.

**Undercounting is the deliberate direction.** The emit is an unauthenticated, unrate-limited `GET` on an optional catch-all route with no `robots.txt` disallow, so the raw row count is inflatable by a refresh loop, a crawler or a chat-client link unfurler. An inflated public number is worse than a conservative one.

**If you want a different window, say so now** — the definition is one expression in `buildAppOpenCountSql` and one mirror in the pure spec. The two obvious alternatives and why I did not take them: dropping the daily bucket (`uniqExact` over all time) makes a returning user worth 1 forever, which reads badly next to "installs"; using a rolling 30-day window would make a public number that silently *decreases*, which is worse on a store card than one that only grows.

## On-site only

Off-site listings have no trustworthy source at all — their CTA is a plain third-party anchor, so no on-platform request follows the click. Both counters gate on `kind = 'onsite' AND app_block_id IS NOT NULL`, and this is **asserted**, not left to fall out of the join: a test counts the gate occurrences (2, one per counter) and another feeds an off-site listing that *does* carry an `app_block_id` and requires 0.

🔴 **A `0` in this column therefore means "no plays OR not measurable"**, and nothing in the column distinguishes them. **Stage 3 must project `openCount` as `null` for `kind='offsite'` rather than 0** — the operator's decision is that off-site cards omit the stat entirely rather than show a zero that reads as "nobody used it". Documented in both `appListing.metrics.ts` and `appListing.metrics.sql.ts`.

## Ownership contract

`app-listing-review.service.ts` says *"'No recompute' is a TWO-SIDED contract … A third writer is only safe if it brings a full recompute with it."* This is that third writer, and the recompute it brings is total. **There is no `+1` path anywhere**, and a test pins that: the ON CONFLICT set must not match `/"open_count"\s*\+/` and must not reference the column's own current value.

The INSERT column list and the ON CONFLICT DO UPDATE set now name `install_count`, `open_count`, `updated_at` — and still **never** `thumbs_up_count` / `thumbs_down_count`. `connect_count` / `visit_count` / `tipped_*` remain absent. I also corrected the now-stale claim in `app-listing-review.service.ts` that the processor "names only `install_count`".

## An existing guard was deliberately narrowed

`appListing.metrics.test.ts` asserted *"does not write connect/open/visit/tipped counters (no reader; feature not live)"*. `open` now has a reader and a trusted source, so `open_count` moved **out** of that not-written list and **into** the positive ON CONFLICT assertion. The other three are unchanged and still asserted absent. The narrowing is explained inline at the assertion and in the commit message.

## 🔴 One non-obvious design decision, with the measurement behind it

The ClickHouse reads use **`toString(type) = 'App_Open'`, not `type = 'App_Open'`**. `actions.type` is an `Enum16`, and comparing it to a literal the enum does not carry is a **hard query error**, not an empty result. Measured directly against ClickHouse 26.7.5 with a pre-migration enum definition:

| form | result |
|---|---|
| `PREWHERE type = 'App_Open'` | `Code: 691. DB::Exception: Unknown element 'App_Open' for enum … (UNKNOWN_ELEMENT_OF_ENUM)` |
| `PREWHERE toString(type) = 'App_Open'` | `0` — clean, no error |

The bare form would make this processor **throw on every run — taking `install_count` down with it** — on any deploy where #4652's widening migration has not been applied. `toString` simply matches nothing, which is precisely the inert behaviour this rollup wants until the migration lands. Both queries also apply the type predicate in `PREWHERE`, because the count query is intentionally unbounded in time and nothing prunes it otherwise.

### Live SQL — both engines, not just string assertions

**Postgres 17**, throwaway instance, real tables, `PREPARE` + `EXECUTE` of the actual exported SQL strings:

- The upsert and the affected query both **parse and execute** (the `LEFT JOIN unnest($2::text[], $3::int[]) AS oc(...)` form is valid).
- Fixture: `apl_on` (on-site, 3 subs / 1 disabled) with a **pre-existing metric row created by the thumbs writer** carrying `thumbs 17/4` and `connect/visit/tipped 91/92/93/94`.
- Result: `install_count 2`, `open_count 4` (from the supplied map), and **`thumbs_up 17`, `thumbs_down 4`, `connect 91`, `visit 92`, `tipped 93`, `tipped_amount 94` all unchanged**. That is the ownership contract verified against a real database, not just against a regex.
- `apl_off` / `apl_noblock` → `0 / 0`. `apl_pending` (non-approved) excluded — `INSERT 0 4`, not 5.
- Play-only affected arm: with a watermark set to **2999-01-01** (so no subscription arm can fire) and every listing already holding a metric row (so no seed arm can fire), passing `ARRAY['apb_beta']` returned exactly `apl_on2`. Positive control that the new arm is what selected it.
- **Empty parallel arrays** (`$2 = '{}'`, `$3 = '{}'` — the batch where every listing has a null `app_block_id`): executes cleanly, `open_count` COALESCEs to 0, thumbs preserved. That case is not exercised by any string assertion and would have surfaced only in production.
- The same probe also **demonstrates the hazard the design note warns about**: re-running the upsert for `apl_on` with an *empty* map rewrote its `open_count` from 4 back to 0. That is precisely why the processor queries ClickHouse for the **whole** affected set rather than only the recently-active blocks — a partial map is not a no-op, it is a zeroing.

**ClickHouse 26.7.5** (`clickhouse-local`), real `MergeTree` `actions` table with an `Enum16` `type` and a hand-built fixture:

- `apb_alpha` → **4** (one user refresh-looping 3× on 09-05 = 1, a second user = +1, one anon IP twice = +1, the first user again on 09-06 = +1).
- `apb_beta` → **4** (two anon actors on different IPs, plus `userId 5` and the literal ip `'5'` which must **not** collapse).
- A `Generator_Submit` row on the same app was not counted; a row with `details = '{}'` was dropped; `apb_gamma` (not in the `IN` list) was absent from the count but present in the recent-blocks result.

### Mutation matrix — 27 mutants, 26 killed

Each mutant was applied to the production source, the suite run, the failing test names and assertion messages recorded, and the source restored. Baseline before **and** after the sweep: `Tests 48 passed (48)`.

| # | Mutant | Result | Test that reds | Message |
|---|---|---|---|---|
| M1 | actor key ignores the anonymous sentinel (always `u:<userId>`) | KILLED | `two ANONYMOUS actors on different IPs count 2` | `expected 1 to be 2` |
| M2 | actor key drops the `u:` / `i:` prefixes | KILLED | `a userId and an ip that render the same string are still two actors` | `expected '5' not to be '5'` |
| M3 | no dedup — count raw rows per day | KILLED | `a refresh loop by ONE signed-in actor collapses to 1` | `expected 4 to be 1` |
| M4 | no per-day bucketing — one bucket for all time | KILLED | `the same actor on two different UTC days counts 2` | `expected 1 to be 2` |
| M5 | day bucket uses the LOCAL calendar day, not UTC | KILLED | `the day boundary is UTC, not local` | `expected '2026-09-05' to be '2026-09-06'` |
| M6 | spec: `measurable` drops the `kind==='onsite'` gate | KILLED | `OFF-SITE listing carrying a stray app_block_id STILL gets 0` | `expected 4 to be +0` |
| M7 | spec: `measurable` drops the non-null `app_block_id` gate | **SURVIVED** | — | see below |
| M8 | spec: approved-only filter removed | KILLED | `non-approved listings … are excluded` | `expected [ 'apl_draft', … ] to deeply equal [ 'apl_ok' ]` |
| M9 | spec: install count stops filtering on `enabled` | KILLED | `installCount = its ACTIVE (enabled) subscription count` | `expected 3 to be 2` |
| M10 | upsert: `open_count` dropped from the ON CONFLICT set | KILLED | `the ON CONFLICT DO UPDATE writes install_count / open_count / updated_at` | `expected 'ON CONFLICT…' to contain '"open_count" = EXCLUDED."open_count"'` |
| M11 | upsert: `thumbs_up_count` added to the ON CONFLICT set | KILLED | `NEVER writes thumbs_up_count / thumbs_down_count` | `expected ' INSERT INTO…' not to contain 'thumbs_up_count'` |
| M12 | upsert: `connect_count` re-added to the INSERT list | KILLED | `does not write connect/visit/tipped counters` | `expected ' INSERT INTO…' not to contain 'connect_count'` |
| M13 | CH **count** query: bare Enum comparison | KILLED | `matches the action type by toString(type), NOT by Enum comparison` | `expected '…' to contain 'toString(type) = \'App_Open\''` |
| M14 | affected query: the `$2` play arm removed | KILLED | `has an arm keyed on the ClickHouse-supplied app_block_id list ($2)` | `expected ' SELECT al.id, …' to contain 'al."kind" = \'onsite\' AND al."app_bl…'` |
| M15 | upsert: `LEFT JOIN unnest` → `INNER JOIN` | KILLED | `is a LEFT join, so a listing absent from the map is written 0` | `expected ' INSERT INTO…' not to contain 'INNER JOIN unnest'` |
| M16 | CH count query: `uniqExact` → `count()` | KILLED | `the count query buckets by UTC day and counts DISTINCT actors per bucket` | `expected '…' to contain 'uniqExact(if(userId != 0, …'` |
| M17 | `escapeClickhouseString` becomes the identity function | KILLED | `escapes quotes and backslashes in the IN list rather than dropping the id` | `expected 'ap\'b' to be 'ap\\\'b'` |
| M18 | spec: `openCount` hardcoded to 0 | KILLED | `an on-site listing gets its app block's deduped all-time play count` | `expected +0 to be 3` |
| M19 | upsert: `open_count` gate loses `kind==='onsite'` | KILLED | `both counters gate on kind = onsite AND a non-null app_block_id` | `expected [ Array(1) ] to have a length of 2 but got 1` |
| M20 | upsert: `open_count` becomes an **increment** | KILLED | `open_count is DERIVED from the supplied map, never incremented` | `expected ' INSERT INTO…' not to match /"open_count"\s*\+/` |
| M21 | CH **recent** query: bare Enum comparison | KILLED | `matches the action type by toString(type)…` | `expected '\n    SELECT DISTINCT…' to contain 'toString(type) = \'App_Open\''` |
| M22 | CH recent query: loses the watermark bound | KILLED | `the recent-blocks query is bounded by the watermark and drops empty ids` | `expected '…' to contain 'time >= parseDateTimeBestEffort(…'` |
| M23 | CH recent query: stops dropping empty `appBlockId` | KILLED | same as M22 | `expected '…' to contain 'JSONExtractString(details, \'appBlock…'` |
| M24 | affected query: stops returning `app_block_id` | KILLED | `returns app_block_id so the caller can query ClickHouse without a second round trip` | `expected ' SELECT al.id FROM…' to contain 'SELECT al.id, al."app_block_id"'` |
| M25 | spec: stops dropping empty-`appBlockId` rows | KILLED | `a row whose details carried no appBlockId is dropped, not counted under ""` | `expected true to be false` |
| M26 | CH count query: drops the app-block `IN` filter | KILLED | `the count query filters to the requested app block ids` | `expected '…' to contain 'WHERE JSONExtractString(details, \'ap…'` |
| M27 | CH count query: gains a 30-day window | KILLED | `the count query is all-time (no time predicate)` | `expected '…' not to contain 'time >='` |

**M7 survived, and I am reporting it as a real result rather than papering over it.** With `measurable = l.kind === 'onsite'` alone, a null-`appBlockId` listing still produces `0/0`: no subscription row has a null `app_block_id`, and `Map.get(null)` is `undefined`. The same holds in SQL — `bus.app_block_id = al.app_block_id` never matches NULL, and the LEFT JOIN COALESCEs to 0. So that conjunct is **redundant by construction, not merely untested**: no observable behaviour depends on it. I kept it (it mirrors the SQL's `IS NOT NULL` and states intent) and added a comment saying it is *not* a behavioural guard, so nobody later reads it as coverage it does not provide.

Two other guardrails on the sweep itself: the spec⇄SQL lockstep block has an explicit **positive control** (`the parser found something to compare`) so a regex that matched nothing cannot make the equality assertions pass vacuously, and it pairs the derived comparison with a literal (`and that set is install_count + open_count`) so both sides drifting together is still caught.

### Gates

All run from the worktree through the repo's dev shell.

| Gate | Command | Result |
|---|---|---|
| Unit (target file) | `npx vitest run --project unit src/server/metrics/__tests__/appListing.metrics.test.ts` | **48 passed (48)** |
| Unit (adjacent contracts) | `npx vitest run --project unit …/app-listing-review.mod-exclude.test.ts …/action-type-enum-drift.test.ts` | **40 passed (40)**, 2 files |
| Unit (full project) | `npx vitest run --project unit` | **1637 files passed / 3 skipped**, **38137 tests passed / 33 skipped**, 0 failed |
| Typecheck | `pnpm typecheck` | `typecheck: OK — 0 type errors in 54s` |
| Typecheck negative control | same, with a planted return-type error | `typecheck: 3 type error(s)`, rc 2 — the gate can go red |
| ESLint | `npx eslint --format unix <the 4 changed files>` | no output (0 problems) |
| Prettier | `npx prettier --check <the 4 changed files>` | `All matched files use Prettier code style!` (it flagged one file on the first run; fixed with `--write`, so this verdict is about files it actually read) |

## 🔴 Inert until #4652 merges

This branch is based **directly on `origin/main`** and imports nothing from #4652 — it matches the literal string `'App_Open'` in ClickHouse. That makes the two PRs genuinely independent and avoids a stacked PR.

The consequence is that **`open_count` computes and writes 0 until #4652 merges and its `Enum16` widening migration (`2026-09-05-app-open-action.sql`) is applied to production.** No `App_Open` row can exist before both of those happen: the tracker's insert is rejected at the tracker service for a value the column does not carry. This is safe rather than merely harmless — the `toString(type)` form above means the query returns an empty set instead of erroring in the meantime.

**Merge-order note:** both PRs edit the header comment block of `src/server/metrics/appListing.metrics.ts`, so whichever merges second will need a trivial comment-only conflict resolution. #4652's version of that block states the two obligations this PR discharges (dedupe at read time; add `open_count` to the ON CONFLICT list); this PR's version documents them as discharged. Keep this PR's version when resolving.

## Not verified

- **No production ClickHouse or Postgres was touched.** The Postgres and ClickHouse results above are from throwaway local instances with hand-built fixtures matching the documented schemas — they prove the SQL parses, executes and computes the right numbers, but not that production's `actions` table has exactly the column types I assumed (`userId` numeric, `ip` String, `details` String, `time` DateTime), which I read out of `tracker.ts` rather than from the live cluster.
- **Query cost on the real table is unmeasured.** The all-time count query has no time bound by design. `PREWHERE toString(type)` is there to keep it to a narrow column scan, and it only runs when the affected set is non-empty (this processor is on a `*/5 * * * *` schedule), but I have not run it against production's 92.8M-row `actions`. Worth watching the first time real rows exist.

https://claude.ai/code/session_01PXFskJ3JuYmGAmTFj8cFDU

### DIFF
--- a/src/server/jobs/update-metrics.ts
+++ b/src/server/jobs/update-metrics.ts
@@ -24,8 +24,11 @@ const metricSets = {
 const metricSchedules: Record<string, string> = {
   'model-collections': '*/5 * * * *', // every 5 minutes
   basemodels: '*/5 * * * *', // every 5 minutes
-  // Install/connect counts don't need minute-freshness — the store `popular`
-  // sort tolerates a few minutes of lag. Off-peak the affected set is empty.
+  // Install and play (open) counts don't need minute-freshness — the store
+  // `popular` sort tolerates a few minutes of lag. Off-peak the affected set is
+  // empty. The interval is also a cost knob: each run's play count is an unbounded
+  // ClickHouse scan (see appListing.metrics.sql.ts), so tightening this multiplies
+  // that scan rather than the useful work.
   'app-listings': '*/5 * * * *', // every 5 minutes
 };
 
--- a/src/server/metrics/__tests__/appListing.metrics.processor.test.ts
+++ b/src/server/metrics/__tests__/appListing.metrics.processor.test.ts
@@ -1,0 +1,118 @@
+import client from 'prom-client';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * 🔴 THE SEAM, NOT EITHER SIDE OF IT.
+ *
+ * appListing.metrics.test.ts proves the soft-degrade POLICY (a query rejection
+ * degrades; a builder or mapping throw propagates). appListing.metrics.prom.test.ts
+ * proves the SIGNAL exists on the registry `/api/metrics` scrapes and increments.
+ * Both pass with the processor's `onDegrade` never calling the emitter at all — the
+ * counter would be a correct, registered, permanently-zero series, and "blipped
+ * once" vs "dead for a week" would still be indistinguishable, which is the entire
+ * reason the counter was added.
+ *
+ * So this file drives the REAL `appListingMetrics.update()` with a real (unmocked)
+ * `appListing.metrics.sql` and a real prom registry, and mocks only the infra edges
+ * the processor talks to: ClickHouse, Postgres, the job-date/queue plumbing.
+ */
+
+const { chQuery, pgCancellableQuery } = vi.hoisted(() => ({
+  chQuery: vi.fn(),
+  pgCancellableQuery: vi.fn(),
+}));
+
+vi.mock('~/server/clickhouse/client', () => ({ clickhouse: { query: chQuery } }));
+vi.mock('~/server/jobs/job', () => ({
+  // An hour ago, so base.metrics' `shouldUpdate` gate is open.
+  getJobDate: async () => [new Date(Date.now() - 3_600_000), async () => undefined],
+}));
+vi.mock('~/server/redis/queues', () => ({
+  checkoutQueue: async () => ({ content: [], commit: async () => undefined }),
+  addToQueue: async () => undefined,
+}));
+vi.mock('~/server/db/pgDb', () => ({
+  pgDbWrite: { cancellableQuery: pgCancellableQuery },
+  pgDbRead: { cancellableQuery: pgCancellableQuery },
+  pgDbReadLong: { cancellableQuery: pgCancellableQuery },
+}));
+
+import { appListingMetrics } from '../appListing.metrics';
+
+const METRIC = 'civitai_app_listing_open_discovery_degraded_total';
+const jobContext = { checkIfCanceled: () => undefined, on: () => undefined } as never;
+
+/** A @clickhouse/client ResultSet stand-in. */
+const chRows = (rows: unknown[]) => ({ json: async () => rows });
+/** A pg `cancellableQuery` stand-in. */
+const pgRows = (rows: unknown[]) => ({ result: async () => rows, cancel: async () => undefined });
+
+async function degradeCount(): Promise<number> {
+  const metric = client.register.getSingleMetric(METRIC) as
+    | { get(): Promise<{ values: Array<{ value: number }> }> }
+    | undefined;
+  if (!metric) return Number.NaN;
+  return (await metric.get()).values[0]?.value ?? Number.NaN;
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  // `resetMetrics()`, NOT `removeSingleMetric()`: the counter is registered eagerly
+  // at the prom module's scope (which this file's import of the processor pulls in),
+  // and removing it would delete exactly the "healthy pod reports 0" property the
+  // negative control below is asserting.
+  client.register.resetMetrics();
+});
+
+describe('appListingMetrics.update — the degrade signal is actually wired', () => {
+  it('🔴 a ClickHouse DISCOVERY failure increments the counter AND the run continues', async () => {
+    // The regression the soft path exists to prevent is `install_count` being held
+    // hostage by a ClickHouse blip, so both halves are asserted together: the signal
+    // fires, and the Postgres affected-set query still runs.
+    chQuery.mockRejectedValue(new Error('MEMORY_LIMIT_EXCEEDED'));
+    pgCancellableQuery.mockResolvedValue(pgRows([]));
+
+    await expect(appListingMetrics.update(jobContext)).resolves.toBeUndefined();
+
+    expect(await degradeCount()).toBe(1);
+    expect(pgCancellableQuery).toHaveBeenCalledTimes(1);
+    expect(pgCancellableQuery.mock.calls[0][0]).toContain('app_listings');
+  });
+
+  it('🔴 NEGATIVE CONTROL: a HEALTHY run leaves the counter at 0', async () => {
+    // Without this the counter could be incremented unconditionally — a series that
+    // is always non-zero cannot distinguish a degrade from a normal cycle, which is
+    // the same blindness with the opposite sign.
+    chQuery
+      .mockResolvedValueOnce(chRows([{ appBlockId: 'apb_x' }])) // discovery
+      .mockResolvedValueOnce(chRows([{ appBlockId: 'apb_x', openCount: 4 }])); // count
+    pgCancellableQuery
+      .mockResolvedValueOnce(pgRows([{ id: 'apl_1', app_block_id: 'apb_x' }])) // affected
+      .mockResolvedValueOnce(pgRows([])); // upsert
+
+    await appListingMetrics.update(jobContext);
+
+    expect(await degradeCount()).toBe(0);
+    // Positive control on the fixture: the run really did do its work, so the 0
+    // above is "nothing degraded", not "nothing ran".
+    expect(chQuery).toHaveBeenCalledTimes(2);
+    expect(pgCancellableQuery).toHaveBeenCalledTimes(2);
+    expect(pgCancellableQuery.mock.calls[1][0]).toContain('ON CONFLICT');
+  });
+
+  it('🔴 the COUNT read fails HARD and does NOT touch the degrade counter', async () => {
+    // The asymmetry, at the seam: `open_count` is derived, so a partial count map is
+    // written over live counts as 0. That must fail the run, not degrade it — and it
+    // must not be reported on the counter that means "we degraded and carried on".
+    chQuery
+      .mockResolvedValueOnce(chRows([{ appBlockId: 'apb_x' }])) // discovery succeeds
+      .mockRejectedValueOnce(new Error('Code: 62. DB::Exception: Max query size exceeded'));
+    pgCancellableQuery.mockResolvedValueOnce(pgRows([{ id: 'apl_1', app_block_id: 'apb_x' }]));
+
+    await expect(appListingMetrics.update(jobContext)).rejects.toThrow('Code: 62');
+
+    expect(await degradeCount()).toBe(0);
+    // No upsert ran — the counts stay at their last good values.
+    expect(pgCancellableQuery).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/server/metrics/__tests__/appListing.metrics.prom.test.ts
+++ b/src/server/metrics/__tests__/appListing.metrics.prom.test.ts
@@ -1,0 +1,144 @@
+import client from 'prom-client';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import {
+  ensureRegisterAppListingMetrics,
+  recordAppListingOpenDiscoveryDegrade,
+} from '../appListing.metrics.prom';
+
+/**
+ * The REAL prom-client side of the AppListing play-discovery degrade signal.
+ *
+ * 🔴 WHY THIS SIGNAL EXISTS AT ALL. `fetchRecentlyOpenedBlockIds` fails SOFT: when
+ * ClickHouse cannot answer, the run recomputes `install_count` only and returns an
+ * empty play-candidate set. That degraded state is by construction
+ * INDISTINGUISHABLE from a quiet cycle — both produce an empty list and an
+ * unchanged `open_count` — so "blipped once" and "dead for a week" look identical
+ * from the outside. One Axiom warning per run cannot express that difference as an
+ * alert; a counter's rate over time can.
+ *
+ * 🔴 WHY IT IS TESTED AGAINST THE REAL REGISTRY. A metric-name typo or a registry
+ * the scrape does not read would produce an alert rule that silently never fires —
+ * the same class of defect the counter exists to remove. So this file drives the
+ * actual default registry that `/api/metrics` collects, rather than a spy.
+ */
+
+const METRIC = 'civitai_app_listing_open_discovery_degraded_total';
+
+/**
+ * 🔴 CAPTURED AT MODULE SCOPE, ON PURPOSE — this is the ONLY place the eager
+ * registration is observable.
+ *
+ * ES imports are evaluated before this line, so this records the registry state
+ * produced by NOTHING BUT importing the module. Every `it` below runs after a
+ * `beforeEach` that removes the metric, and the first one to call the emitter
+ * registers it again — so inside a test body "is it registered?" is answered by
+ * whatever ran earlier in the file, not by the module. Measured: deleting the
+ * module-scope `ensureRegisterAppListingMetrics()` call left the whole suite GREEN
+ * until this capture existed.
+ */
+const REGISTERED_BY_IMPORT_ALONE = client.register.getSingleMetric(METRIC) !== undefined;
+
+/** Read the counter's single (unlabelled) series from the default registry. */
+async function read(): Promise<number> {
+  const metric = client.register.getSingleMetric(METRIC) as
+    | { get(): Promise<{ values: Array<{ value: number }> }> }
+    | undefined;
+  if (!metric) return Number.NaN;
+  const { values } = await metric.get();
+  return values[0]?.value ?? Number.NaN;
+}
+
+beforeEach(() => {
+  // Full isolation, not `resetMetrics()`: two tests below deliberately re-register
+  // the name (once absent, once with an incompatible TYPE), and `resetMetrics` only
+  // zeroes values — it would leave those registrations standing for the next test.
+  client.register.removeSingleMetric(METRIC);
+});
+
+describe(METRIC, () => {
+  it('is registered on the DEFAULT registry — the one /api/metrics scrapes', () => {
+    ensureRegisterAppListingMetrics();
+    expect(client.register.getSingleMetric(METRIC)).toBeDefined();
+  });
+
+  it('🔴 is registered by IMPORTING the module — a never-degraded pod is 0, not absent', () => {
+    // If registration only happened inside the emitter, the series would not exist
+    // until the first degrade: a healthy pod would scrape `no data`, which reads as
+    // "nothing ever degraded" but equally means "the instrument was never wired" —
+    // precisely the ambiguity this counter exists to remove. The processor imports
+    // this module, so import-time registration is what makes the 0 real in prod.
+    expect(REGISTERED_BY_IMPORT_ALONE).toBe(true);
+  });
+
+  it('reports an observable 0 before anything degrades, not `no data`', async () => {
+    // An absent series is ambiguous — it reads as "nothing ever went wrong" when it
+    // may equally mean "nobody loaded the module". An unlabelled prom-client counter
+    // materialises its single series at registration, which is what removes that.
+    ensureRegisterAppListingMetrics();
+    expect(await read()).toBe(0);
+  });
+
+  it('increments once per degraded run', async () => {
+    ensureRegisterAppListingMetrics();
+    expect(await read()).toBe(0);
+    recordAppListingOpenDiscoveryDegrade();
+    expect(await read()).toBe(1);
+    recordAppListingOpenDiscoveryDegrade();
+    recordAppListingOpenDiscoveryDegrade();
+    expect(await read()).toBe(3);
+  });
+
+  it('registers itself when the emitter is the first caller', async () => {
+    // The processor's onDegrade calls only the emitter — it never calls the
+    // ensure-register helper first. If registration were left to some other module,
+    // the very first degrade would be lost.
+    client.register.removeSingleMetric(METRIC);
+    expect(client.register.getSingleMetric(METRIC)).toBeUndefined();
+    recordAppListingOpenDiscoveryDegrade();
+    expect(await read()).toBe(1);
+  });
+
+  it('🔴 declares NO labels — persistence is read from the rate, attribution from the log', async () => {
+    // Asserted against the DECLARED labelNames, not the emitted series: prom-client
+    // omits a declared-but-never-supplied label from the output, so inspecting
+    // `values[].labels` stays green while the metric is declared wide open. The
+    // natural label here would be the ClickHouse error message — an unbounded
+    // population that would blow the cardinality budget. It lives in the Axiom
+    // warning instead.
+    const metric = ensureRegisterAppListingMetrics().openDiscoveryDegradedTotal as unknown as {
+      labelNames: string[];
+    };
+    expect([...metric.labelNames]).toEqual([]);
+
+    recordAppListingOpenDiscoveryDegrade();
+    const values = await (
+      client.register.getSingleMetric(METRIC) as unknown as {
+        get(): Promise<{ values: Array<{ labels: Record<string, string> }> }>;
+      }
+    ).get();
+    expect(values.values).toHaveLength(1);
+    expect(values.values[0].labels).toEqual({});
+  });
+
+  it('is get-or-create — a second import reuses the instance instead of throwing', () => {
+    const first = ensureRegisterAppListingMetrics().openDiscoveryDegradedTotal;
+    const second = ensureRegisterAppListingMetrics().openDiscoveryDegradedTotal;
+    expect(second).toBe(first);
+  });
+
+  it('🔴 NEVER THROWS — a metrics failure must not fail the run it is observing', () => {
+    // The emitter sits inside the catch that keeps `install_count` running through a
+    // ClickHouse outage, so a registry collision must not convert a degrade into a
+    // failed run. Claim the name with an INCOMPATIBLE type (a Histogram has
+    // `observe`, not `inc`), so the get-or-create's unchecked cast hands the emitter
+    // an object whose `.inc` is undefined.
+    new client.Histogram({ name: METRIC, help: 'collision', registers: [client.register] });
+    // Positive control: without this, a future prom-client where the collision does
+    // NOT produce a throwing call site would make the assertion below vacuous.
+    const claimed = client.register.getSingleMetric(METRIC) as unknown as { inc?: unknown };
+    expect(typeof claimed.inc).toBe('undefined');
+
+    expect(() => recordAppListingOpenDiscoveryDegrade()).not.toThrow();
+  });
+});
--- a/src/server/metrics/__tests__/appListing.metrics.test.ts
+++ b/src/server/metrics/__tests__/appListing.metrics.test.ts
@@ -5,6 +5,8 @@ import {
   APP_LISTING_BATCH_SIZE,
   APP_LISTING_METRIC_UPSERT_SQL,
   APP_OPEN_ACTION_TYPE,
+  APP_OPEN_CH_CHUNK_SIZE,
+  APP_OPEN_COUNT_QUERY_MARKER,
   appOpenActorKey,
   appOpenUtcDay,
   buildAppOpenCountSql,
@@ -587,6 +589,14 @@ describe('affected-set selection — arms, including the repair arm', () => {
   });
 
   it('REPAIR: also fires on a non-zero install_count, and on an OFF-SITE row', () => {
+    // 🔴 THE PREDICATE IS "NULL JOIN KEY + NON-ZERO COUNT", NOT "the AppBlock was
+    // deleted". `app_block_id IS NULL` is not a kind discriminator (schema.prisma
+    // on `AppListing.appBlockId`), so the arm also covers a natively-created
+    // off-site listing that never had a block to lose — `apl_offsite` here IS that
+    // population. In production its counters are written 0 by the same upsert, so
+    // the arm is a no-op over it; a row that does hold a non-zero count is wrong by
+    // definition and wants the same clearing. Do not read a firing of this arm as
+    // evidence of a deletion.
     const out = selectAffectedApprovedListings(
       affectedInput({
         listings: [
@@ -813,11 +823,17 @@ describe('spec ⇄ SQL lockstep', () => {
  * so one over-long query takes `install_count` down with it — and two live triggers
  * reach that size: the seed arm on a fresh or restored `app_listing_metrics`, and
  * the store simply growing past ~7,700 approved on-site listings.
+ *
+ * 🔴 BUT CHUNKING SMALLER IS NOT "SAFER" — IT IS LINEARLY MORE EXPENSIVE, AND EVERY
+ * ASSERTION IN THIS BLOCK IS RELATIVE TO THE CONSTANT, SO NONE OF THEM CAN SEE IT.
+ * Each chunk is an unbounded `actions` scan whose cost does not shrink with the `IN`
+ * list, so the chunk COUNT is the cost. The absolute block below is the guard that
+ * survives the constant moving in either direction.
  */
 describe('fetchAppOpenCounts — chunking and merge', () => {
-  // Deliberately NOT a multiple of the batch size: the last chunk is partial, so a
+  // Deliberately NOT a multiple of the chunk size: the last chunk is partial, so a
   // mutant that drops or double-counts a trailing chunk cannot land on the boundary.
-  const BLOCK_COUNT = APP_LISTING_BATCH_SIZE * 2 + 37;
+  const BLOCK_COUNT = APP_OPEN_CH_CHUNK_SIZE * 2 + 37;
   const allIds = Array.from({ length: BLOCK_COUNT }, (_, i) => `apb_${i}`);
   // Per-id distinct, and distinct from the batch size, the chunk count and the
   // index — a mutant returning a constant or the wrong chunk's numbers cannot pass.
@@ -843,12 +859,12 @@ describe('fetchAppOpenCounts — chunking and merge', () => {
     expect(idsIn(buildAppOpenCountSql(['apb_a', 'apb_b']))).toEqual(['apb_a', 'apb_b']);
   });
 
-  it('splits into ceil(n / APP_LISTING_BATCH_SIZE) queries, none over the batch size', async () => {
+  it('splits into ceil(n / APP_OPEN_CH_CHUNK_SIZE) queries, none over the chunk size', async () => {
     const { queries, runQuery } = recordingRunner();
     await fetchAppOpenCounts(allIds, runQuery);
-    expect(queries).toHaveLength(Math.ceil(BLOCK_COUNT / APP_LISTING_BATCH_SIZE));
+    expect(queries).toHaveLength(Math.ceil(BLOCK_COUNT / APP_OPEN_CH_CHUNK_SIZE));
     for (const sql of queries) {
-      expect(idsIn(sql).length).toBeLessThanOrEqual(APP_LISTING_BATCH_SIZE);
+      expect(idsIn(sql).length).toBeLessThanOrEqual(APP_OPEN_CH_CHUNK_SIZE);
     }
   });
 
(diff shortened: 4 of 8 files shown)


ZacxDev:
## Round-1 audit fixes applied — `7453cbd1af`

All seven findings addressed in one commit. Each new or changed guard was watched fail for its own reason; matrix below.

---

### 1. 🔴 The UTC-day guard was vacuous on a UTC runner

**Reproduced the reported vacuity, then closed it.** `vitest.config.mts` is untouched — the fix is local to the test file, so no other suite's ambient zone moves.

The old assertion only fails on a runner whose local zone differs from UTC. Replaced with a zone-varying block that runs `appOpenUtcDay` **and** `computeAppOpenCounts` under four ambient zones (`UTC`, `America/Chicago`, `Asia/Kolkata`, `Pacific/Kiritimati`), chosen to straddle the two instants in both directions: UTC−5 pulls a just-after-midnight instant back a day, UTC+14 pushes a just-before-midnight one forward. Whatever zone the runner is in, at least two of the four disagree with it.

`process.env.TZ` is set/restored around each call. Node picking that up at runtime was **verified empirically, not assumed** (Node v24.19.0, forks pool) and is pinned by a **positive control** test: if a Node/pool combination ignored the override, every zone case would silently collapse into a re-run of the ambient zone, which is the same vacuity wearing four names. The original single-zone assertion is kept, relabelled as the plain statement of expected values rather than as the guard.

**Acceptance criterion, met:**

| ambient TZ | `toLocaleDateString('en-CA')` mutant | before | after |
|---|---|---|---|
| `UTC` | | **SURVIVED** (48 passed) | **KILLED** — 6 failed / 72 passed |
| `America/Chicago` | | KILLED (3 failed / 45 passed) | **KILLED** — 9 failed / 69 passed |

Suite is green under both zones unmutated (78/78 each).

### 2. The SCOPE comment claimed a reader that does not exist

Confirmed by enumeration, not sampling: the only non-metrics occurrences of `open_count`/`openCount` in the tree are generated declarations — `packages/civitai-db-schema/src/kysely/types.ts`, `.../src/models.ts`, `prisma/schema.prisma`, its migration, and a schema-drift fixture. No reader.

Rewritten so the file no longer states a rule and silently exempts itself. The exception is now argued explicitly: a **trusted server-side source** exists today, the consumer lands in stage 3, and populating ahead is safe **because** the value is derived and idempotent — every run recomputes it in full, so there is no accumulating state to get wrong while unread and no backfill to get right later. It also states why this cannot be cited for `connect`/`visit`/`tipped`: those have no source at all, so populating one would mean inventing the number. Mirrored in `appListing.metrics.ts` and in the narrowed test's comment (which had said "Stage 2 gave it a reader").

### 3. A ClickHouse blip took `install_count` down with it

Handling is now asymmetric, and both halves say so in their own docstrings.

- **Discovery read → SOFT.** Degrades to "no new plays this run", logs to Axiom (`app-listing-metrics-open-discovery-degraded`), and lets `install_count` proceed. Self-healing: `setLastUpdate()` still runs, but the next run's watermark still covers the gap. This is the read that sits before `if (!affected.length) return`, so on a quiet cycle a throw here used to abort `update()` with nothing to do at all.
- **Count read → HARD.** Unchanged and now documented as deliberate: `open_count` is derived, so a block missing from the map is *written as 0*. A partial answer does not degrade the run, it zeroes live numbers.
- A **cancelled job is not a ClickHouse failure** — `onDegrade` may throw to veto the degrade, and the processor's calls `checkIfCanceled()` first so a cancellation is rethrown rather than swallowed into "no plays".

The docstring at the old `:146-155` that argued fail-hard for the discovery read on count-read grounds is replaced.

### 4. The count query was unchunked past ≈7,700 ids

Chunked at the shared batch size (200 — now single-sourced as `APP_LISTING_BATCH_SIZE` and consumed by both the Postgres upsert batching and the ClickHouse reads, so the two cannot drift). Chunks run sequentially, not concurrently: each is a full-table scan, so parallelism multiplies peak ClickHouse memory for no useful win.

The chunk/merge lives in `fetchAppOpenCounts`, which takes an injected `runQuery` so it stays dependency-free and is exercisable without a ClickHouse client. Tested with `BATCH_SIZE * 2 + 37 = 437` ids — deliberately **not** a multiple of the batch size, so the last chunk is partial and a mutant cannot land on its own boundary. Expected counts are `id * 3 + 11`, distinct per id and distinct from the batch size, the chunk count and the index, so a mutant returning a constant or the wrong chunk's numbers cannot pass. Asserted: query count, no chunk over the batch size, merged map complete and per-id correct, every id asked exactly once, dedup/empty-drop, zero queries on an empty set, and a chunk failure propagating. Plus a positive control that the test's id extractor can actually see ids in a query.

### 5. Losing the join key froze a published count

Added a repair arm: an approved listing whose `app_block_id IS NULL` while its metric row still carries a non-zero `open_count` **or** `install_count` is selected, so the existing `measurable` gate recomputes it to 0. Self-terminating — the run it fires on writes 0, after which the predicate stops matching. Deliberately not gated on `kind`: an off-site row with a stale non-zero count is the same defect.

**Verified against real Postgres 17**, not only against the spec mirror. The extracted production SQL strings were run against a throwaway cluster with the orphan seeded at `open_count = 777`:

| step | result |
|---|---|
| affected query | selects `apl_orphan` (+ `apl_seed` via the seed arm); does **not** select the settled, already-zeroed, or non-approved rows |
| upsert with an empty play map | `install_count → 0`, `open_count → 0`, `thumbs_up_count` still `9` (ownership contract holds) |
| affected query, second pass | `apl_orphan` no longer selected — self-terminating |
| **negative control:** same query, repair arm deleted, count restored to 777 | `apl_orphan` **not selected** — the frozen-count defect reproduced |

In the unit suite this is pinned two ways, because a structural assertion alone would leave the behaviour unpinned: the affected query's arms are now mirrored by an executable spec (`selectAffectedApprovedListings`) with per-arm tests including a settled-listing negative control, and the SQL side pins the **whole normalised arm string** rather than a keyword, so a reworded-but-equivalent arm fails and gets re-read.

### 6. Documented, not fixed: the count read is structurally unbounded

Written into the code comment beside the query, per the instruction — no issue filed. It records that primary-key usage is `type` only with the `IN` filter applied above `ReadFromMergeTree`, so `details`/`userId`/`ip`/`time` are read for every `App_Open` row ever written regardless of how few blocks are asked about, every 5 minutes, with no TTL on `actions`. It names the standard remedy — an `AggregatingMergeTree` MV keyed `(appBlockId, day)` over `uniqExactState`, turning the scan into a point lookup — and **concrete triggers** rather than "when it hurts": query wall time > ~5 s or `read_rows` > ~100M in `system.query_log`; `App_Open` passing ~1% of `actions`; or the schedule tightening below 5 minutes / the figure gaining a timeframe variant. The schedule comment in `update-metrics.ts` now cross-references it as a cost knob.

### 7. Cheap correctness nits — all four

- Both ClickHouse queries now run through `chCancellableQuery`, which registers `jobContext.on('cancel', …)` and passes an `abort_signal`. `ctx.ch.$query` exposes no abort handle, which is why the Postgres queries beside them had one and these did not; dropping to `ctx.ch.query` is what makes it reachable.
- The batch `Map` comment now says it is **defensive, not load-bearing** — the `ON CONFLICT … cannot affect row a second time` error is real and a widened join is how you'd reach it, but `AppListing.appBlockId` is `@unique` (schema.prisma:2908), so no batch can contain a duplicate.
- `update-metrics.ts` schedule comment no longer says "Install/connect counts".
- The counter guard now asserts `not.toContain('tipped')`. Verified the gap was real: `'tipped_count' in 'tipped_amount_count'` is `False`, so the old spelling asserted nothing about the amount column.

---

## Mutation matrix

Every mutant applied to the real source, suite re-run, mutant reverted, green re-confirmed.

| # | mutant | tests that red | failing assertion |
|---|---|---|---|
| 1 | `appOpenUtcDay` → `toLocaleDateString('en-CA')`, **TZ=UTC** | 6 (`bucket boundaries stay UTC under TZ=…` ×3, `the dedup buckets by UTC day … under TZ=…` ×3) | `expected '2026-09-05' to be '2026-09-06'`; `expected 1 to be 2` |
| 1b | same mutant, **TZ=America/Chicago** | 9 (the 6 above + `the same anonymous IP twice in a day counts 1`, `the same actor on two different UTC days counts 2`, `the day boundary is UTC, not local`) | `expected 2 to be 1`; `expected '2026-09-05' to be '2026-09-06'` |
| 2 | count read **unchunked** (one query for the whole set) | 1 — `splits into ceil(n / APP_LISTING_BATCH_SIZE) queries, none over the batch size` | `expected [ Array(1) ] to have a length of 3 but got 1` |
| 3 | chunks **no longer merge** (`counts.clear()` per iteration) | 1 — `the merged map is COMPLETE and per-id correct across every chunk` | `expected 37 to be 437` |
| 4 | **repair arm removed from the SQL** | 1 — `has the REPAIR arm — a lost join key with a non-zero count still published` | `expected ' SELECT al.id, al."app_block_id" FROM…' to contain 'OR ( al."app_block_id" IS NULL AND EX…'` |
| 5 | **repair arm removed from the executable spec** | 3 — `REPAIR: a NULL join key with a non-zero published open_count is selected`, `REPAIR: also fires on a non-zero install_count, and on an OFF-SITE row`, `END TO END: the repaired listing then recomputes to 0` | `expected [] to deeply equal [ 'apl_orphan' ]` |
| 6 | **discovery read fails hard again** (try/catch removed) | 2 — `a ClickHouse failure degrades to "no new plays" so install_count still runs`, `onDegrade MAY VETO by throwing` | `Error: MEMORY_LIMIT_EXCEEDED`; `expected [Function] to throw error including 'Job was canceled' but got 'The user aborted a request.'` |
| 7 | upsert writes **`tipped_amount_count`** | 4 — incl. `does not write connect/visit/tipped counters` | `expected ' INSERT INTO "app_listing_metrics" ( …' not to contain 'tipped'` |

Mutant 4 also verified **against real Postgres** (see finding 5's table) — with the arm deleted the orphan row is not selected at all.

Honest note on mutant 7: three of the four reds are the spec⇄SQL lockstep guards, which fire on any counter-set change. The widened `tipped` guard did fire **with its own assertion**, and the isolated claim — that the old `tipped_count` spelling could not have caught `tipped_amount_count` — was checked directly (`False`).

Mutant 2's second assertion (no chunk over the batch size) is never reached under that mutant, since the query-count assertion fires first. It is a redundant guard on the same property, not independent coverage.

## Gates

Run from the worktree via `direnv exec … env -C …`; explicit file arguments throughout (no unquoted variable holding a file list).

| gate | command | result |
|---|---|---|
| unit — file under change | `npx vitest run --project unit src/server/metrics/__tests__/appListing.metrics.test.ts` | **78 passed (78)**, 1 file — green under `TZ=UTC` **and** `TZ=America/Chicago` |
| unit — all metrics tests | `npx vitest run --project unit src/server/metrics/__tests__/` | **201 passed (201)**, 10 files |
| unit — app-listing-review (the other writer of this table) | `npx vitest run --project unit src/server/services/blocks/__tests__/app-listing-review.{service,store-scope,mod-exclude}.test.ts` | **69 passed (69)**, 3 files |
| typecheck | `pnpm typecheck` | `OK — 0 type errors in 17s` |
| eslint | `npx eslint --format unix <4 files>` | 0 problems |
| prettier | `npx prettier --check <4 files>` | `All matched files use Prettier code style!` |

**Both linters were positive-controlled**, because "0 problems" and "All matched files use Prettier code style!" are also what a tool wired to nothing prints. A deliberately bad file was added to the *same* invocation: prettier reported `[warn] …__prettier_eslint_control.ts` while still passing the four real files, and eslint reported `1 problem … [Error/prefer-const]`. The control file was deleted; `git status` is clean apart from the four changed paths. The typecheck instrument's negative control came for free — it went red on a real `TS2322` mid-work (`ResultSet.json<T>()` resolves to `T`, not `T[]`) and green after the fix.

## Not verified

- **No live ClickHouse or production Postgres run.** The `Code: 62` ceiling and the `EXPLAIN indexes=1` key-usage finding are taken from the audit's measurements, not re-measured here. Finding 5 *was* re-measured, against a throwaway Postgres 17 cluster (audit used 16), so the DDL is my minimal stand-in for the three tables rather than the real schema.
- **The Axiom log line has not been observed firing** — the degrade path is unit-tested through `onDegrade`, but nothing has exercised the real `logToAxiom` call.
- **`abort_signal` cancellation is not integration-tested.** The wiring is typechecked and follows the client's documented surface; no test drives a real cancelled job. Note it registers one listener per query, so a chunked count read adds one per chunk — the existing Postgres batch loop already does the same.

---

```audit-claims round=1 audited=2d3554d0a9f5401584712cee2a6893167d17c887..7453cbd1af
1. appOpenUtcDay's UTC-day property is now asserted under four ambient timezones with a positive control, and the toLocaleDateString mutant is killed under both TZ=UTC and TZ=America/Chicago.
2. The SCOPE comment no longer claims open_count has a reader; it states the consumer lands in stage 3 and argues the exception from source + derived/idempotent value, in both the SQL module and the processor and the narrowed test comment.
3. The discovery read fails soft (degrade to no-new-plays, log to Axiom) while the count read still fails hard, and a cancelled job is rethrown rather than swallowed; the docstring arguing fail-hard on count-read grounds is corrected.
4. The ClickHouse count read is chunked at APP_LISTING_BATCH_SIZE with the per-chunk maps merged, and a test exercises 437 ids across 3 chunks proving the merged map is complete and per-id correct.
5. The affected query has a repair arm selecting an approved listing with a NULL app_block_id and a non-zero metric row, so the recompute writes 0; verified against real Postgres including the arm-deleted negative control, and mirrored by an executable spec with per-arm tests.
6. The count read's unbounded-scan cost is documented in the code comment beside the query, naming the AggregatingMergeTree MV remedy and numeric triggers; no issue was filed and no MV was built.
7. Both ClickHouse queries register a cancel listener via abort_signal; the batch Map comment is labelled defensive rather than load-bearing; the schedule comment no longer says install/connect; the counter guard asserts on 'tipped' so tipped_amount_count is covered.
8. APP_LISTING_BATCH_SIZE is single-sourced in the SQL module and consumed by both the Postgres upsert batching and the ClickHouse count chunking.
9. The processor's stale "RECOMPUTABILITY HAS ONE HOLE" comment is updated to say the hole now fails to zero rather than freezing.
```

ZacxDev:
## Round-2 audit fixes — `2b100da9a4`

All five findings addressed. Every number below was re-derived, not carried over.

---

### 1 🟡 The round-1 chunking made the dominant cost ~39× worse

**Fix.** Split the constant. `APP_LISTING_BATCH_SIZE = 200` now covers *only* the Postgres upsert batches; the ClickHouse count read gets `APP_OPEN_CH_CHUNK_SIZE = 2_000`. Both constants carry an explicit statement of the two-directional pressure (upsert optimum pushes small: shorter row-lock hold under `limitConcurrency(tasks, 5)`; ClickHouse optimum pushes large: each chunk is a full `actions` scan whose cost is flat in the `IN` list, bounded only by `max_query_size`).

**The arithmetic, re-derived.** An id is `apb_` + a 26-char ULID = 30 chars, rendered as `'<id>', ` = **34 bytes**; the rest of the query is a fixed **460 bytes**:

```
bytes(n) = 460 + 34n
```

That model **reproduces round-1's live measurement exactly** — `bytes(7000) = 238,460`, the figure the ceiling note reports from ClickHouse 26.8.2.7 — which is what licenses using it to state an exact ceiling instead of the 7,000/8,000 bracket:

| n | bytes | vs `max_query_size` 262,144 |
|---|---|---|
| 2,000 (**adopted**) | 68,460 | 3.83× under |
| 3,841 | 131,054 | last value under the ½ budget |
| 7,000 | 238,460 | 1.10× under (round-1's measured OK) |
| **7,696** | **262,124** | **exact hard ceiling** |
| 7,697 | 262,158 | over → `Code: 62` |
| 8,000 | 272,460 | over (round-1's measured failure) |

Seed path (~7,700 approved on-site listings): `ceil(7700/2000) = 4` scans, against `ceil(7700/200) = 39`. That is the 39× the audit named. 2,000 adopted as suggested — 4,000 would only buy 4→2 scans while spending most of the headroom, and `max_query_size` is a per-server *setting* that can be lowered under us.

**Absolute guards.** Every prior chunking assertion was relative (`<= APP_OPEN_CH_CHUNK_SIZE`), so all of them stay green for any value of the constant. New block `🔴 ABSOLUTE bounds on the ClickHouse chunk size`, built **only from literals** that do not move when either constant moves — bracketing from both sides:

- **upper** — one full chunk of production-shaped ids must fit in `262_144 / 2`;
- **upper** — the chunk size must not exceed `7_000` (the largest list measured safe live);
- **lower** — the ~7,700-listing seed path must cost **at most 4** scans;
- plus: the two constants must not be equal, and `fetchAppOpenCounts` must *default* to the ClickHouse one.

A positive control asserts the byte measurement can actually exceed the ceiling (`bytes(8000) > 262,144`) and that `realisticIds` really produces 30-char ids — without it, a short-id fixture would make every "under budget" assertion pass over a tiny query.

### 2 🟡 The soft discovery path swallowed permanent and programming errors, with no aggregable signal

**Narrowed the `try`.** `fetchRecentlyOpenedBlockIds` now builds the SQL and runs `rows.map(...)` *outside* the `try`; only the awaited query is inside. A builder throw or a non-array result now propagates instead of returning `[]` on every run forever.

**Added the aggregable signal.** New module `src/server/metrics/appListing.metrics.prom.ts` exposing `civitai_app_listing_open_discovery_degraded_total`, wired into the processor's `onDegrade` beside the existing Axiom warning. In the comment: **what it distinguishes** — a degraded run is by construction indistinguishable from a quiet one (both yield an empty candidate set and an unchanged `open_count`), so only the counter's **rate over time** separates "blipped once" from "dead for a week"; the Axiom warning keeps the per-run error attribution that a deliberately unlabelled counter does not carry. It is unlabelled by design (the natural label would be the ClickHouse error message — unbounded cardinality) and the emitter never throws.

**A defect this surfaced in my own work.** The first draft registered the counter lazily inside the emitter, so a never-degraded pod would scrape `no data`, not `0` — the exact ambiguity the counter was added to remove. Caught by the seam test returning `NaN` on the healthy-run case. Fixed with eager module-scope registration; see M11 below for why that line needed its own guard.

**Closed the isolation seam.** New `appListing.metrics.processor.test.ts` drives the *real* `appListingMetrics.update()` (real `appListing.metrics.sql`, real prom registry, mocked infra edges only) and pins: a discovery failure increments the counter **and** the Postgres affected query still runs; a healthy run leaves it at 0; and the hard-failing count read rejects the run without touching the counter. Neither of the other two spec files can see that wiring — both pass with the emitter never called.

### 3 🟢 The `abort_signal` docstring claimed a teardown the client cannot deliver

Corrected to state what abort actually does — ends the Node-side HTTP request, so the promise rejects, the response stream stops being consumed and the socket is released — and that it does **not** cancel the query inside ClickHouse. Names the setting that would make the stronger claim true (`cancel_http_readonly_queries_on_client_close`, default 0) and records that `createClickhouseClient` sets only `async_insert` / `wait_for_async_insert` / `output_format_json_quote_64bit_integers`. Adds the operational consequence, since this is load-bearing for sizing against finding 1: **size ClickHouse for the scans running to completion.**

### 4 🟢 The MV trigger's diagnostic query could not separate the expensive read from the cheap one

Introduced `APP_OPEN_COUNT_QUERY_MARKER = 'sum(dailyActors)'`, **interpolated into** `buildAppOpenCountSql` rather than spelled there and quoted in the comment — so renaming `dailyActors` moves the constant and the runbook cannot silently grade the wrong queries. The runbook query is now bounded, discriminating and ordered:

```sql
SELECT event_time, query_duration_ms, read_rows, memory_usage
  FROM system.query_log
 WHERE type = 'QueryFinish'
   AND event_date >= today() - 7
   AND query LIKE '%sum(dailyActors)%'
 ORDER BY query_duration_ms DESC
 LIMIT 20
```

**Verified mechanically, not by eye** — new `MV-trigger diagnostic` block asserts the marker is present in the count query, **absent** from the discovery query, and that the count query is literally built from the constant. Its positive control asserts `App_Open` alone is in **both** queries, i.e. it pins the ambiguity being fixed rather than assuming it.

### 5 🟢 The repair arm's prose asserted something the schema forbids

Rewritten in all four places (SQL docstring, inline SQL comment, processor mirror, test) to describe the real predicate — **NULL join key AND a non-zero published count** — and both populations it covers: (1) the frozen-number case, a listing whose backing AppBlock was deleted (`onDelete: SetNull`), which is why the arm exists; (2) a natively-created off-site listing, which never had a block to lose, quoting the schema's own "It is NOT a kind discriminator: discriminate on `kind`, never on appBlockId nullness". Population 2 is expected to be empty (the upsert writes those rows 0), so the arm is a no-op over it — no behavioural change. Explicitly warns off the follow-up the audit predicted: *do not build an alert reading "repair arm fired ⇒ an AppBlock was deleted"*.

---

## Mutation matrix

13 mutants, applied one at a time by a harness that asserts the anchor is unique, restores byte-for-byte (sha256-verified) and re-runs green after each. **All 13 killed.** Baseline: 104 passed.

| # | Mutant | Killed by | Message |
|---|---|---|---|
| M1 | `APP_OPEN_CH_CHUNK_SIZE` → 8,000 (past the hard ceiling) | ABSOLUTE byte + id bounds | `expected 272460 to be less than 131072`; `expected 8000 to be less than or equal to 7000` |
| M1b | → 4,000 (past the ½ budget, under the hard ceiling) | ABSOLUTE byte bound | `expected 136460 to be less than 131072` |
| M2 | re-widen the `try` to cover builder + `rows.map` | builder + mapping propagation tests | `promise resolved "[]" instead of rejecting` (×2) |
| M3b | marker reverted to the ambiguous `'App_Open'` | marker-absent-from-discovery | `expected '...SELECT DISTINCT...' not to contain 'App_Open'` |
| M4 | `fetchAppOpenCounts` default → `APP_LISTING_BATCH_SIZE` | LOWER bound + default test | `expected 39 to be less than or equal to 4` |
| M5 | `APP_OPEN_CH_CHUNK_SIZE` → 200 (re-merged) | separate-constants + LOWER bound | `expected 200 not to be 200`; `expected 39 to be less than or equal to 4` |
| M6 | emitter body no-op (drop `.inc()`) | increments + seam tests | `expected +0 to be 1` |
| M7 | counter renamed (drop `_total`) | 9 tests incl. registry presence | `expected undefined to be defined` |
| M8 | counter gains an `error` label | declared-labels guard | `expected [ 'error' ] to deeply equal []` |
| M9 | emitter rethrows instead of swallowing | never-throws guard | `expected [Function] to not throw ... '...inc is not a function'` |
| M10 | remove `recordAppListingOpenDiscoveryDegrade()` from the processor | **seam** tests | `expected NaN to be 1` |
| M11 | remove eager module-scope registration | import-time registration guard | `expected false to be true` |
| M12 | make the count read fail **soft** | fail-hard asymmetry guard | `promise resolved "undefined" instead of rejecting` |

🔴 **M11 SURVIVED on its first run** and is worth flagging. The eager-registration line — the thing that makes "a healthy pod reports 0" true — was unguarded: the processor test's *first* test registers the counter via the emitter, so by the time the negative control ran, "is it registered?" was answered by test order, not by the module. Fixed by capturing `client.register.getSingleMetric(METRIC) !== undefined` **at module scope** in the prom spec, before any `beforeEach` can remove it. M11 then died on its own assertion. The full battery was re-run after that change and after the prettier pass; 13/13 killed both times.

M3's first form (appending `/*App_Open*/` to the marker) died to the *byte-arithmetic* and *SQL-shape* tests but **not** to the discrimination guard — it made the marker ambiguous without putting the whole marker string into the discovery query. Replaced with M3b, which is the mutation the guard is actually for. Reporting both since the first one was a badly-chosen mutant, not a passing result.

## Gates

| Gate | Command | Result |
|---|---|---|
| Unit (targeted) | `npx vitest run --project unit <3 files>` | **104 passed / 104**, 3 files |
| Unit (collateral) | `npx vitest run --project unit src/server/metrics/__tests__/ src/server/services/blocks/__tests__/app-listing-review.mod-exclude.test.ts` | **258 passed / 258**, 13 files |
| Typecheck | `pnpm typecheck` | `rc=0` — `typecheck: OK — 0 type errors in 99s` |
| Typecheck — instrument validated | same, with an injected `const x: number = "str"` | `rc=2` — `error TS2322` + `typecheck: 1 type error(s)` |
| ESLint | `npx eslint --format unix <6 files + 1 deliberately bad control>` | **1 problem, on the control file only** — 0 on all 6 changed files |
| Prettier | `npx prettier --check <6 files + the same control>` | control flagged; the 2 real issues it found in my test files were fixed, re-check `All matched files use Prettier code style!` |

Both linter positive controls ran **in the same invocation** as the real files, so the clean verdict on the 6 files is a verdict the instrument was demonstrably able to fail.

## Not verified

- **No live ClickHouse.** The `Code: 62` ceiling, the 2,000-id chunk and the 4-scan seed path are verified against the query **builder's byte output** and round-1's recorded live measurement, which the `460 + 34n` model reproduces exactly. Nothing here re-ran a query against a real server.
- **The 34-bytes-per-id assumption** rests on ids being `apb_` + a 26-char ULID. It is corroborated by reproducing round-1's 238,460 at 7,000 ids, but not by sampling production ids.
- **The `~7,700 approved on-site listings` seed figure** is carried forward from this file's existing docstrings; I did not re-count it against the database.
- **The counter is not observable until the AppListing processor module is loaded** on a pod — i.e. on the pool that runs metric jobs. Stated in the module comment rather than papered over. No alert rule is added by this PR.

```audit-claims round=2 audited=7453cbd1af24d4bc5f26ec427e10598d75d9546b..2b100da9a4cc3ba28faf69c91503731fbcf85085
1. The ClickHouse count read is chunked at its own constant APP_OPEN_CH_CHUNK_SIZE = 2,000, no longer at the Postgres APP_LISTING_BATCH_SIZE = 200.
2. Both constants carry a comment stating the two-directional pressure and forbidding re-merging them.
3. The chunk size is bounded by test-local LITERALS on both sides: a full chunk under 131,072 bytes, the chunk size at most 7,000 ids, and the ~7,700-listing seed path at most 4 ClickHouse scans.
4. The bytes(n) = 460 + 34n model, the 7,696-id exact ceiling and the 238,460-byte figure at 7,000 ids are asserted by test against the real query builder.
5. fetchAppOpenCounts defaults to APP_OPEN_CH_CHUNK_SIZE, pinned by a test that fails if the default reverts to the Postgres batch size.
6. The try in fetchRecentlyOpenedBlockIds wraps only the awaited query; the SQL builder and rows.map are outside it.
7. A builder throw and a non-array query result each propagate rather than degrading, pinned by two tests with a degrading negative control in the same block.
8. A Prometheus counter civitai_app_listing_open_discovery_degraded_total is registered on the default registry that /api/metrics scrapes, and is incremented by the processor's onDegrade.
9. That counter is registered at module import, so a pod that has loaded the processor reports 0 rather than an absent series.
10. The counter declares no labels, never throws, and is not incremented by the hard-failing count read.
11. The processor seam is pinned by a test driving the real appListingMetrics.update(): the counter moves on a discovery failure, stays 0 on a healthy run, and stays 0 when the count read fails hard.
12. The abort_signal docstring now says abort ends the Node-side HTTP request only, does not cancel the query server-side, and names cancel_http_readonly_queries_on_client_close as what would.
13. The MV-trigger runbook query filters on sum(dailyActors), bounds event_date to 7 days, orders by query_duration_ms and limits to 20.
14. APP_OPEN_COUNT_QUERY_MARKER is interpolated into the count query rather than duplicated, and a test asserts it is present there and absent from the discovery query, with a positive control that App_Open alone is in both.
15. The repair-arm prose in the SQL docstring, the inline SQL comment, the processor mirror and the test describes the predicate as "NULL join key AND a non-zero published count" and names both populations it covers.
16. That prose states that app_block_id nullness is not a kind discriminator and warns against an alert inferring an AppBlock deletion from the arm firing.
17. No behavioural change to the repair arm, the affected-set SQL, the upsert SQL or the compute spec.
```

ZacxDev:
## Round 3 (delta `7453cbd1af..2b100da9a4`) — no 🔴, no 🟡. **Ladder STOPPED**, fixes in `278f73b974`.

All 17 claims from round 2's fix commit hold — 16 fully, 1 partially — and the auditor re-derived the numbers rather than accepting them: the byte model `460 + 34n` is exact for every n ≥ 1 against the real builder output, `238,460 @ 7,000` reproduces the recorded live figure, and the 7,696 / 7,697 ceiling, the 3.83× margin and the 39 → 4 scan counts all check out.

The largest new surface — the Prometheus counter — was checked against the things that make a metric silently useless here: no name collision, registered on `client.register` only (so `/api/metrics` concatenating two registries **cannot** emit a duplicate `# HELP`/`# TYPE` and reject the whole response), double-registration impossible by construction, and correctly *absent* from the server-graph singleton watchlist because it holds no module-scope state — verified against two existing siblings with the identical shape.

### Why the ladder stops here

Round 3 produced **no finding requiring a code fix**. Both 🟢 were text, fixed in `278f73b974` without another round; re-running a round to confirm a clean round is exactly what the stop rule forbids. The attribution gate is *not* what ended it — this ladder changed executable payload in every round (round 3: a new emitter module, two new constants, restructured control flow), so it never left the PR.

### The two 🟢, and a third I introduced while fixing them

**A docstring with two wrong numbers.** It said *"That 39x is the cost of merging them"* about a comparison it had just stated as 39 scans vs 4 — which is **9.75×**. The 39× is right against a *different* denominator (the one unchunked query that preceded chunking), so one figure was doing duty for two comparisons. It also offered *"4 → 2 scans at n = 4,000"* as an available trade; n = 4,000 is 136,460 bytes and **cannot be chosen at all** — it exceeds the 131,072-byte literal budget added one commit earlier, so setting it reds the UPPER BOUND test.

🔴 **My first replacement for that sentence was wrong in the same shape**, and it is recorded rather than quietly fixed because it is the trap this ladder keeps finding: I wrote that raising the constant buys nothing short of n ≥ 3,850, having checked only the 2-scan step. Re-deriving showed **3 scans is reachable in budget from n ≥ 2,567** (87,738 bytes), so the honest claim is "at most one scan is available, 4 → 3". A third figure I wrote ("28% of the headroom") was also wrong — 28% is the query-size increase, the budget share is 14.7%. Both now carry the arithmetic they came from.

**A test title claiming what its assertions cannot observe.** *"the marker is what the count query is actually built from (no second spelling)"* — measured: replacing the interpolated marker in `buildAppOpenCountSql` with a literal `sum(dailyActors)`, i.e. exactly the second spelling the title forbade, leaves the file **93/93 green**. `toContain` cannot distinguish interpolation from duplication. The title now claims only what it checks, and the comment records where the hazard *is* covered — renaming the aggregate in the SQL while leaving the constant reds three assertions in that block.

### One self-report that did not reproduce

`2b100da9a4`'s message says *"13 mutants, all killed, each on its own assertion"*. One did not reproduce as isolated: deleting the `recordAppListingOpenDiscoveryDegrade()` call also leaves its import unused, the SSR transform elides the dead import, the prom module never loads, and **three** tests die — two with `expected NaN to be +0`, i.e. for the wrong reason. Isolating the mutation kills exactly one, with `expected +0 to be 1`. **The guard is correct and correctly scoped; only the mutation as reported was not.**

### Gates

104 tests across the three metrics suites · `pnpm typecheck` 0 errors · eslint 0 errors · prettier clean.

### Carried forward, not fixed here

- The count read is **structurally unbounded** — by `EXPLAIN`, it scans every `App_Open` row ever written, and `actions` has no TTL. Documented in code with the `AggregatingMergeTree` remedy and numeric triggers; the trigger's diagnostic query was fixed this round to discriminate the expensive read from the cheap one.
- Everything about production ClickHouse here is reasoned from the repo's documented graph model plus local probes — **not** from a live scrape. The `Code: 62` ceiling and the `EXPLAIN` key-usage finding were measured on throwaway instances.
- Stage 3 must project `openCount` as **`null`** for `kind='offsite'`, never `0`. Nothing mechanical enforces that yet; it is a required guard for stage 3's review.

```audit-claims round=3 audited=2b100da9a4cc3ba28faf69c91503731fbcf85085..278f73b974
1. Corrected the chunk docstring's "39x" to 9.75x for the stated comparison, and named both denominators so they are not re-merged.
2. Corrected the claim that n = 4,000 is an available trade — it exceeds the 131,072-byte budget and reds the UPPER BOUND test.
3. Recorded that my own first correction was also wrong (3 scans IS reachable from n >= 2,567) and stated the arithmetic behind every figure.
4. Narrowed the marker test's title to what its assertions observe, and documented that the drift hazard is covered by a sibling assertion.
5. Recorded that one of 2b100da9a4's 13 reported mutants was not isolated; the guard itself is correct.
6. LADDER STOPPED at round 3: no finding required a code fix.
```

ZacxDev:
## Round 4 (delta `2b100da9a4..278f73b974`) — fixes in `9887ae653d`. **Ladder STOPPED.**

This round audited a 33-line, comment-and-test-title-only commit. That is normally not worth a round; it was run because the commit rewrote a paragraph whose arithmetic had already been wrong **three times**. **It was wrong a fourth time**, so the round paid for itself.

### The correction contained the defect it was correcting

The note justifying the retitled marker test claimed renaming the aggregate *"reds three assertions in this block (including the one above, which requires the marker to be absent from the discovery query)"*. Measured, on both readings of "rename the aggregate":

| mutation | tests red | in that block |
|---|---|---|
| alias only (`AS dailyActors` → `AS dailyActrs2`, constant untouched) | **1** — `the count query buckets by UTC day…` | **0** |
| de-interpolate **and** rename | 3 | 2 |

And the assertion the note credited — `🔴 the marker is ABSENT from the cheap discovery query` — **passes under both, and structurally always will**: `recent` never contains `dailyActors` in any spelling, so its `not.toContain` is satisfied whatever the count query says. It provides **none** of that coverage, and naming it pointed maintainers away from the sibling that actually pins the alias literal.

A coverage description wider than its implementation is precisely what the retitling was fixing. It is now written per-mutation with measured counts instead of as a summary claim.

### Two more claims that did not survive derivation

**The "TWO DENOMINATORS" note asserted a census it had not done.** It attributed the 39× to `buildAppOpenCountSql`'s docstring, which quotes no such figure — the 39× is at `APP_LISTING_BATCH_SIZE`'s docstring — and two sites added one commit earlier still carried a bare `~39x` reading as the 9.75× comparison (`appListing.metrics.ts` and the absolute-bounds note in the test file). Both corrected; the note now enumerates every site rather than asserting they agree.

🔴 **`~7,700` had no provenance, and both thresholds this branch added (2,567 and 3,850) derive from it.** It is this file's own hard ceiling (7,696) rounded up — the largest population an *unchunked* query could have served — silently reused as a *seed population*, and cited as *"the seed/restore path this file's docstrings name"*, which is a self-reference, not a source. The real seed size is today's approved on-site listing count, recorded nowhere in this repo. Now stated as an order of magnitude, with the derived thresholds marked exact **given** that input and moving with it.

### Why the ladder stops here — on the gate, not on a clean round

Round 4's findings were real, so the findings-keyed rule alone would continue to round 5. It stops instead because **this is the second consecutive fix round with zero executable payload**. `278f73b974` was the first, and round 4 verified that mechanically rather than assuming it: both versions transpiled with `removeComments: true` and byte-diffed (7,528 B identical), with a positive control confirming the stripper could see the one real change.

Two such rounds is the documented signal that the ladder is auditing **its own prose** rather than the PR. The behaviour shipped by `2b100da9a4` is unchanged and was audited clean at that tip; everything since has been comment accuracy.

**What that means for a reader:** the remaining risk in this PR is not in its code. It is that the docstrings now carry numeric guidance — chunk sizes, scan counts, a seed population — that has been wrong four times and is derived from an input nobody has measured. Treat those numbers as reasoning, not as measurements, and re-derive before sizing a production constant from them.

### Gates

104 tests across the three metrics suites · `pnpm typecheck` 0 errors in 74 s · eslint 0 errors · prettier clean.

### Carried forward, unchanged

- The count read is **structurally unbounded** — it scans every `App_Open` row ever written, and `actions` has no TTL. Documented with the `AggregatingMergeTree` remedy and a discriminating diagnostic query.
- Everything about production ClickHouse here is reasoned from local probes and the repo's documented graph model, **not** a live scrape.
- Stage 3 must project `openCount` as **`null`** for `kind='offsite'`, never `0`. Nothing mechanical enforces it yet — a required guard for stage 3's review.

```audit-claims round=4 audited=278f73b974e0a1dc5ed9a4724e0f9936ab4f9595..9887ae653d
1. Rewrote the marker test's coverage note per-mutation with measured red counts (1 red / 0 in-block for an alias-only rename; 3 red / 2 in-block when de-interpolated).
2. Removed the credit given to "the marker is ABSENT from the discovery query", which cannot red on a count-query rename, and named the sibling that actually pins the alias.
3. Corrected the TWO DENOMINATORS note's file attribution and converted it to a full census of every 39x/9.75x site.
4. Corrected the two bare ~39x sites in appListing.metrics.ts and the test file's absolute-bounds note to 9.75x.
5. Stated 7,700's real provenance — the rounded hard ceiling, not a measured population — and marked the derived thresholds as moving with it.
6. Dropped a hardcoded "93/93 GREEN" that goes stale when a test is added.
7. LADDER STOPPED on the attribution gate: second consecutive fix round with zero executable payload, verified by comment-stripped byte-diff.
```

PR #4654, opened 2026-09-06
feat(metrics): measure the brotli codec behind redis.packed compression

ZacxDev:
## Problem

The opt-in brotli codec on `redis.packed` values (`compressPacked` /
`decompressPacked`) has **no cost signal at all**. That is not for lack of
metrics nearby — it is that the two signals which look like they cover it both
give a confidently wrong answer:

**CPU profiles cannot see it.** The codec is
`util.promisify(zlib.brotliCompress / zlib.brotliDecompress)`, so it runs on the
libuv threadpool. Threadpool work carries no JS stack, so no `brotli*` frame is
ever sampled. Searching a profile for the codec returns **zero**, which reads as
"the codec is free" rather than "the profiler is structurally blind to it".

**`civitai_app_redis_command_duration_seconds` does not cover it either.** That
observation is made inside the redis command wrapper and closes when the round
trip closes. Compression happens before the write, and decompression strictly
*after* the read — in `client.packed.get` / `client.packed.mGet`, past the point
where the command histogram has already been observed. And it fails in the worse
direction: a compressed payload is **smaller on the wire**, so enabling
compression makes that histogram *improve* while adding codec work it cannot
observe. Reading it as a codec cost gives you the answer backwards.

So today the question "what does compression cost us?" cannot be answered from
any existing signal.

## Change

Adds `civitai_app_packed_codec_duration_seconds` — a histogram of the codec's
elapsed time.

| label | values |
|---|---|
| `op` | `compress` \| `decompress` |
| `cache_name` | the cache **prefix**, e.g. `packed:caches:image-meta` |

### What the number is, precisely

**Elapsed time as the caller sees it, not CPU time.** `performance.now()` starts
*before* the promisified call is enqueued and stops when its callback resolves,
so a sample is libuv threadpool **queue wait + codec work**. That is the right
quantity for "what did compression cost this request", and it is the only window
reachable from JS — the enqueue and the threadpool hand-off are both below the
promisified boundary. But it means the value rises when the threadpool is busy
with unrelated work (other brotli calls, fs, dns, crypto) at unchanged codec
cost. The help string and the `PackedCodecTimer` doc comment both say this
outright, so a rise is read as "the codec **path** got slower", not "brotli got
more expensive".

### What `cache_name` does and does not join to

`cache_name` is the cache **prefix**, chosen so cardinality stays bounded by the
small set of caches that opt into `compress` (two today). Whether it *joins* to
cache-traffic counters depends on the caller, and it holds for only one of the
two consumers — stated here because an earlier revision of this description
claimed it held for both:

* **`createCachedArray` / `createCachedObject` — joins.** The builder passes its
  own `key` as this label *and* passes the same `key` as `cache_name` to
  `civitai_app_cache_hit_total` / `civitai_app_cache_miss_total`, so the two label
  values are equal by construction.
* **`fetchThroughCache` — does not join.** It emits no hit/miss counters at all,
  so there is no cache-traffic series carrying this `cache_name` to join against.
  The nearest counter a reader might reach for is the tensor-metadata in-process
  LRU in `tensor-metadata.service`, which labels itself `tensor-metadata-full` — a
  different value naming a different cache. The label still earns its place on
  this consumer by separating one cache's codec cost from another's; it is simply
  not a join key today, and no dashboard should be built as though it were.

`fetchThroughCache` takes an explicit `cacheName` option rather than reusing its
`key`, because that key is routinely `${prefix}:${id}` — labelling with it would
be unbounded prom cardinality. A compress-aware call site that supplies no name
reports the explicit literal `unknown`, so the label is present on every sample.

### Buckets are cut for the values actually measured

The first draft of this metric put the floor at `0.0005` (0.5 ms) on a "sub-ms
for a typical record" intuition. Review measured that as the wrong floor by more
than an order of magnitude, and it would have shipped a permanently-misleading
quantile: `imageMetaCache` — the compressed cache that carries the sample volume —
holds ~0.5–4 KB values that decompress in **~0.026 ms**, putting **99.4% of
samples at or below the old first edge** (measured in review; p50 error 9.6×), so
`histogram_quantile` would have answered with that edge no matter what the codec
did.

> ⚠️ **Superseded by Round 3 (B1/B2).** This section's fix — a floor of
> `0.00002` with three more edges below the old one — was measured against
> production values in round 3 and found to be *the same mistake one order of
> magnitude smaller*: 38.5% of real decompress samples land at or below 20 µs.
> The shipped edge list, the measurements it is cut against, and the guard that
> pins it are in **Round 3 → B1 and B2**. Kept here because the reasoning error
> it describes is the one round 3 repeats at finer scale.

### A metrics fault must never destroy a cache entry

The timer call sits inside `decompressPacked`, which sits inside
`safeUnpackCompressed`'s `try` — and that `catch` does not mean "something went
wrong", it means **"the stored bytes are corrupt"**, so it logs, `unlink`s the key
and returns `null`. So a bridge whose `observe()` throws (duplicate metric
registration, a rejected label value, an exhausted registry) would not merely
lose a sample: a **valid entry is reported as a miss and deleted**, on every read,
for as long as the fault lasted — an observability change silently becoming a
cache-eviction bug that presents as a cache with a mysterious 0% hit rate.

`packedCodecTimer`'s body is now wrapped in `try { … } catch {}` — one place,
both ops, all three call sites. Three tests drive a **throwing** bridge (the
pre-existing test only covered bridge *absence*, which short-circuits at `?.` and
is a strictly weaker claim) and assert that no key is unlinked and no value is
lost, with a call counter as the positive control that the throwing path was
really entered.

### The legacy path is deliberately not recorded

`decompressPacked` returns a pre-compression (raw msgpack, no sentinel byte)
value untouched after a single byte comparison. Those are **not** recorded as
`decompress` observations. Counting them would fill the histogram with near-zero
samples measuring the cost of *not* decompressing; the p50 would then answer "how
much of this cache is still legacy", which is not the question the metric exists
for.

I chose *skip* over *a separate outcome label* because the passthrough is not a
codec call at all — it is a branch and a return, and giving it a bucket invites
someone to aggregate across the label and get the diluted number back. The
sentinel discrimination stays inside `decompressPacked`, the one function that
already makes it; the caller never re-derives that predicate, so the two cannot
drift apart.

### Wiring

Follows the existing plumbing exactly, with no parallel registry:

* the metric is declared in `@civitai/telemetry`,
* published on the `__civitaiRedisMetrics` globalThis bridge by the app's prom
  shim, alongside `redisCommandDuration` and friends,
* read **per call** by `@civitai/redis` — that package must not statically import
  `prom-client`, since it is reachable from the client bundle, and the bridge is
  published after it loads, so a captured-once handle would pin `undefined`
  forever.

Instrumentation no-ops entirely when the bridge is absent, and swallows a bridge
that throws (both covered by tests).

The timing itself lives in `packed-compression.ts` behind an optional injected
callback, rather than being wrapped at the call sites — that keeps the
codec/legacy discrimination in one place and lets it be unit-tested without a
redis client.

### `fetchThroughCache`'s retry no longer drops the label

The lock-loser retry path re-enters `fetchThroughCache` with a **hand-respelled**
options object rather than passing `options` through, and it threaded `compress`
but not `cacheName`. So a lock loser's retry recorded its samples under
`cache_name="unknown"` even though the caller had named the cache — invisible,
because nothing about the returned value changes. Fixed, and pinned by a test
that drives the real retry.

### Also

`redisCommandDuration`'s round-trip-only scope is now stated in a comment at both
its declaration and its observe site, so the next reader does not mistake it for
a decode cost. **Its semantics are unchanged** — no behaviour, buckets or labels
were touched.

## Tests

`packages/civitai-redis/src/__tests__/packed-codec-timing.test.ts` — 6 tests on
the codec's own timing contract:

* `compressPacked records exactly one compress sample`
* `decompressPacked records exactly one decompress sample when brotli actually ran`
* `decompressPacked records NOTHING for a legacy raw-msgpack value (no sentinel)`
* `decompressPacked records NOTHING for an empty buffer`
* `records SECONDS: 0 < sample <= the caller-observed elapsed time, both ops`
* `is optional: both functions still work with no timer passed`

The seconds assertion is bounded against the caller's *own* wall clock rather
than a fixed threshold, so it is load-independent — a millisecond value overshoots
the elapsed-seconds bound by ~1000x on any machine.

`packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts` — 12 tests, a
**seam** test rather than a component test. It drives the real client, the real
codec, the real cache builders and the real bridge lookup against an in-memory
transport (same approach as the existing `cached-array-compress.test.ts`). A test
that mocked `redis.packed` would pass against a client that instruments nothing,
which is exactly the defect worth catching:

* `a compressed set observes exactly one compress sample under the given cache_name`
* `an UNCOMPRESSED set observes nothing (no codec ran)`
* `a compressed get observes exactly one decompress sample under the given cache_name`
* `a LEGACY (uncompressed) value read through the compress-aware get observes nothing`
* `a compressed mGet observes one decompress per COMPRESSED value only`
* `an UNCOMPRESSED get observes nothing`
* `falls back to the explicit unnamed-cache label when no cacheName is given`
* `does not throw when the app has published no metrics bridge`
* `createCachedObject labels with the cache PREFIX, never the per-id key`
* `a throwing observe() neither loses the value nor unlinks the key on a compressed get` *(new)*
* `a throwing observe() does not fail a compressed set` *(new)*
* `a throwing observe() leaves a whole compressed mGet batch intact` *(new)*

`packages/civitai-telemetry/src/__tests__/packed-codec-buckets.test.ts` — 4 tests
(new file) on the bucket-resolution invariant, including the negative control
described above.

`src/server/utils/__tests__/cache-helpers-cache-name.test.ts` — 5 tests (new
file) against the **real** `fetchThroughCache` with redis mocked. `cacheName` is
observational, so nothing else in that directory can see it go missing:

* `threads cacheName to the packed READ`
* `threads cacheName to the packed WRITE`
* `never substitutes the per-id key for the cache name`
* `passes cacheName undefined when the caller names no cache (no second fallback literal)`
* `preserves cacheName across the lock-loser retry`

`src/tests/api/v1/model-files/tensor-metadata-cache-name.test.ts` — 4 tests (new
file) driving the **real handler module**, so the assertion is about the
production call site rather than a copy of it. The sibling
`tensor-metadata-cache-headers.test.ts` stubs `getFullTensorAnalysisCached` with a
resolved value, which short-circuits the composition — which is precisely why that
file cannot see a dropped option here.

`src/server/utils/__tests__/tensor-metadata-cache-split.test.ts` — its
`makeFetchers` model of the endpoint had drifted (it omitted `cacheName` while
production passes it). Re-synced, and it now records and asserts the label
alongside `compress`. Its header notes that it is a *model* and points at the
handler-driving test above as the thing that pins the real call site.

Two existing structural assertions were widened rather than loosened:
`cached-array-compress.test.ts`'s packedOptions ledger now pins the full new
binding, and `cache-helpers-update.test.ts` now asserts `cacheName` equals the
cache **prefix**, not the entry key — so a regression to per-id labelling fails
there too.

## Mutation table

Every mutant was applied to the committed code, run, and killed by a **named**
assertion (not merely "a test failed"). M1–M7 are from the original round; M8–M12
cover the fixes above.

| # | Mutant | Killed by | Failing assertion |
|---|---|---|---|
| M1 | delete the bridge `observe` call — timer becomes a no-op | 5 tests | `expected [] to deeply equal [ { op: 'compress', … } ]` |
| M2a | `compress` records the label `decompress` | 4 tests | `expected [ 'decompress' ] to deeply equal [ 'compress' ]` |
| M2b | `decompress` records the label `compress` | 5 tests | `expected [ 'compress' ] to deeply equal [ 'decompress' ]` |
| M3 | record the legacy early return as `decompress` | 4 tests | `expected [ { op: 'decompress', seconds: +0 } ] to deeply equal []` |
| M4 | cache builders stop passing `cacheName` | 2 tests | `expected Set{ 'unknown' } to deeply equal Set{ 'packed:caches:test-codec-builder' }` |
| M5 | report milliseconds instead of seconds | 1 test | `expected 1.8352319999999906 to be less than or equal to 0.0018524749999999984` |
| M6 | write path stops passing a timer | 3 tests | `expected [] to deeply equal [ { op: 'compress', … } ]` |
| M7 | read path stops passing a timer | 3 tests | `expected [] to deeply equal [ { op: 'decompress', … } ]` |
| M8 | `cache-helpers.ts` `const cacheName = options.cacheName` → `= undefined` | 4 tests, 2 files | `the packed READ carries the caller cacheName: expected undefined to be 'packed:caches:cache-name-probe'` |
| M9 | drop `cacheName: REDIS_KEYS.CACHES.TENSOR_METADATA` at the endpoint call site | 2 tests, 1 file | `the FULL cache is named, so its codec samples do not fall into the client "unknown" bucket: expected undefined to be 'packed:caches:tensor-metadata'` |
| M10 | remove the `try/catch` from `packedCodecTimer` | 3 tests | `the healthy entry was NOT unlinked by a metrics fault: expected false to be true` |
| M11 | drop `cacheName` from the lock-loser retry recursion **only** | 1 test | `the RETRY read still carries cacheName: expected undefined to be 'packed:caches:cache-name-probe'` |
| M12 | revert the bucket list to the pre-fix edges | 1 test | `the first edge sits BELOW the measured typical decompress, so a typical sample is resolvable: expected 0.0005 to be less than 0.000026` |

M8 and M9 are the two mutants that previously survived the entire green
`unit*` tier byte-identically. M10's assertion reproduces the defect exactly: the
key really is deleted (`expected false to be true` on `store.has(key)`), and the
same run also shows the read returning `null` for a valid entry. M11 is isolated
to the single line — it kills exactly one test and leaves the other 20 green, so
it cannot be dying for another guard's reason.

## Known open — deliberately not fixed here

Recorded so they read as open rather than absent:

* **`bustFetchThroughCache` has no `cacheName` option**
  (`src/server/utils/cache-helpers.ts`). Latent only: the comment at its
  `compress` parameter records that no current caller busts a compressed key, so
  no codec runs on that path and there is nothing to label. It becomes real the
  moment a compressed cache gains a bust call site.
* **`cacheName?: string` is unvalidated** (`packages/civitai-redis/src/client.ts`).
  `fetchThroughCache` takes a free-form string right next to a `key` that is
  routinely `${prefix}:${id}` — a confusable pair, and passing the key would be
  unbounded prom cardinality. The defences today are doc comments plus the tests
  above that assert the label is *not* the key at each call site; nothing
  structurally prevents it. A branded type or a registry of permitted names would,
  at the cost of touching every packed call signature.
* **The codec's dispatch overhead is tracked as civitai#4656 and is explicitly out
  of scope for this PR.** The codec is `promisify(zlib.brotli*)` unconditionally,
  so every call pays a libuv threadpool round trip: measured at p50 17.7 µs for a
  1-byte payload against a p50 21.9–30.2 µs real decompress, i.e. **~63% of an
  async decompress is dispatch, not brotli**. Making the choice size-conditional
  (sync below a threshold, async above it) would remove most of that, but it is a
  change to the codec's execution model — a different review, with its own
  event-loop-blocking argument to make. This PR only *measures* the path; the new
  bucket floor at 5 µs sits below the dispatch floor precisely so that overhead is
  visible as its own band rather than hidden in bucket one.

## Verification

> **The numbers that were here have been deleted rather than maintained.** This
> section carried a round-1 snapshot (`1645/38249`, `1203/8`, `@civitai/redis`
> 191, `@civitai/telemetry` 32) that every later round then contradicted in a
> table below it — two copies of the same claim, one of them always stale. There
> is now **one** authoritative set of counts, at the head of the branch:
> **Round 3 → Round-3 verification**. Round 2's table is kept as history and is
> labelled with the ref it was measured at.

---

# Round 2 — delta-audit fixes (`1243de2f7b`, `f098314ea5`)

A second adversarial pass over the round-1 delta found three guards that claimed
more than they delivered and one shipped help string that gives a wrong
diagnosis instruction. All five are fixed here, plus a guard on the seam the
whole metric hangs on.

## R1 — the bucket guard pinned a constant, not the histogram (2 surviving mutants)

`packed-codec-buckets.test.ts` imported `PACKED_CODEC_DURATION_BUCKETS` and never
read the **registered** metric. Two mutants shipped past the full package tier
(223/223 at the time) and past all four affected `unit*` files:

* **N1** — leave the constant correct, set `buckets:` on `packedCodecDuration` to
  the pre-fix literal `[0.0005, …, 1]`. That ships the exact histogram this guard
  exists to prevent, with nothing red anywhere.
* **N2** — coarsen the constant to `[0.00002, 0.00005, 0.0001, 0.00025, 1]`. The
  old `coversLargeTail` only asserted `last >= 0.036`, so a single 0.25 ms → 1 s
  bucket passed both predicates while being unable to resolve the ~36 ms tail at
  all.

The guard now **observes a sample and reads the `le` set back off prom-client**
(`packedCodecDuration.get()`), so the verdict comes from the registered metric and
the constant is checked separately for agreement with it. The last-edge-only tail
check is replaced by a **max adjacent edge ratio** predicate (`4x`; the shipped
list's widest pair is `2.5x`, so the bound is not pinning today's number). Both
mutants now die — each on its own assertion, with the other predicates still
passing, so neither is dying for a neighbour's reason.

> ⚠️ **Superseded by Round 3 (B2).** The parenthetical above is the second half of
> a claim whose first half — that 4 "is not a multiple of any step in the list, so
> it cannot be satisfied by accident" — is **false**: two consecutive 2× steps
> compose to exactly 4×, so deleting a single interior edge satisfies the bound.
> The bound is replaced in round 3 by an equality on the ordered ratio sequence.

The comment above the export ("Exported so the resolution invariant can be checked
mechanically rather than by eye") now describes what the test actually does.

## R2 — the observe-failure catch was permanently silent, and its stated reason was false

`packedCodecTimer`'s `catch` is correct and stays. But it swallowed with no log,
no counter and no breadcrumb — so a bridge that throws on every `observe` yields a
histogram that never has data, which is byte-for-byte what *"no cache opted into
compress"* looks like. On the one metric that exists to answer whether compression
pays, a reassuring zero is indistinguishable from a probe wired to nothing.

It now **counts every swallowed failure and reports at most one line per minute**,
naming the op, the cache and the underlying error. Rate-limited rather than
one-shot: a throwing bridge throws on every codec call (unthrottled = per-read log
spam), while a strict one-shot goes quiet for the life of the process however long
the fault lasts.

The old comment asserted that the logger on this path *"is the one that reports
'corrupt entry', which is precisely the wrong claim"* — and therefore that no log
line was possible. **That was false.** `log()` in this module is general-purpose,
called with ~20 distinct messages; the corrupt-entry line is one of its messages,
not its identity. The comment is corrected to what is true, and the breadcrumb's
two real limits are stated instead of a new invented reason: `log()` defaults to a
**no-op** and is only wired by the app shim (`createLogger('redis')`), itself gated
on the app's `LOGGING` config. It is a breadcrumb for whoever goes looking, not an
alert.

## R3 — the shipped `help` string gave a wrong diagnosis instruction

Both the doc block and the `help` said a sample was *"threadpool queue wait + codec
work"* and enumerated inflation causes as threadpool-side only. But the clock stops
at the **`await` continuation on the JS thread**, so the sample also absorbs
**event-loop delay**, and it does so at roughly 1:1.

Measured on this tree:

| what | measurement |
|---|---|
| decompress sample while the JS thread was blocked 50 ms | **50.79 ms** (same call is ~0.02 ms unblocked) |
| threadpool round-trip floor, 1-byte payload, no real codec work | **~11 µs p50** on one machine, **~22 µs p50** on another |
| typical `imageMetaCache`-shaped decompress | **~25–36 µs p50** |

So roughly **half or more of a typical sample is the promise/threadpool hand-off,
not brotli**. Both facts — event-loop delay, and the ~10–25 µs dispatch floor — are
now named in `packed-compression.ts`, in the `RedisMetricsBridge` field doc, and in
the shipped `help` string, so nobody reads the low buckets as codec time or a rise
as "brotli got more expensive".

## R4 — a prose guard replaced with a structural one

`fetchThroughCache`'s lock-loser retry re-spelled all five options by hand, with a
comment warning that a sixth would be silently dropped. It now spreads:

```ts
return fetchThroughCache(key, fetchFn, { ...options, retryCount: retryCount - 1 });
```

Behaviour-identical — every default is a **constant** (`CacheTTL.sm`, `10`, `3`,
`false`, `undefined`) applied with `??`, so an option passed through as `undefined`
re-derives exactly the value the resolved local held — and it makes the whole class
impossible rather than pinning it with prose.

## R5 — swept the sibling the round-1 reword missed

`packages/civitai-telemetry/src/client.ts` still opened with *"Wall time of the
brotli codec…"*, the exact framing rewritten 48 lines below and in
`packed-compression.ts`. Same file, same commit. The `RedisMetricsBridge` field doc
in `packages/civitai-redis/src/client.ts` carried the same phrasing and is swept
too.

## R6 — a guard on the seam the metric hangs on

`src/server/prom/client.ts` publishes `packedCodecDuration` onto the globalThis
bridge. **Delete that one line and every codec sample vanishes with no error
anywhere** — the consumer reads it through `?.`, and the package's own suite
installs a *fake* bridge (correctly: it is testing the client, not the app), so
nothing in either package can see it. The observable result is a histogram with no
series: the same reassuring zero as R2.

New `src/server/prom/__tests__/redis-metrics-bridge.test.ts` loads the **real** app
shim and asserts two things, because either alone is walkable:

* a **ledger** of the published handle set, asserted for equality so it fails when
  the set shrinks (the deletion) *or* grows, plus that no key holds `undefined`
  (which satisfies a key check and still short-circuits at the consumer's `?.`);
* a **behavioural** check that the published handle really writes to the scraped
  exposition text, pinned as whole `_count` / `_sum` / `_bucket` lines with a
  negative control first (the probe series must not exist before the observe).

## Round-2 mutation table

Applied to the committed code, run, restored, and re-verified green after each.

| # | Mutant | Killed by | Failing assertion |
|---|---|---|---|
| N1 | constant correct, `buckets:` on the histogram reverted to the pre-fix literal | 2 tests | `prom-client reported bucket edges for the registered metric: expected 11 to be 15` and `the first REGISTERED edge sits BELOW the measured typical decompress: expected 0.0005 to be less than 0.000026` |
| N2 | constant coarsened to `[0.00002, 0.00005, 0.0001, 0.00025, 1]` | 1 test | `the widest adjacent edge ratio must stay at or under 4x…: expected 4000 to be less than or equal to 4` |
| N3 | delete `packedCodecDuration` from the globalThis publish | 2 tests | `expected [ 'redisCommandDuration', …(6) ] to deeply equal [ 'packedCodecDuration', …(7) ]` |
| N3b | publish the key but wire it to `redisCommandDuration` | 1 test (ledger half correctly passes) | the scraped-exposition assertion on `…_count{op="decompress",cache_name="packed:caches:bridge-seam-probe"} 1` |
| N4 | breadcrumb never logs (`if (false)`) | 1 test | `the FIRST swallowed failure is reported: expected [] to have a length of 1 but got +0` |
| N5 | breadcrumb unthrottled (`if (true)`) | 1 test | `a second failure inside the window is throttled, not logged: expected […(2)] to have a length of 1 but got 2` |
| N6 | breadcrumb one-shot forever (`lastLoggedAt === 0`) | 1 test | `the breadcrumb returns after the throttle window: expected [Array(1)] to have a length of 2 but got 1` |
| N7 | count only the failures that were logged (increment moved inside the throttle) | 1 test | `the throttled failures are still counted, so the line reports 3 rather than 1: expected '…failures=2…' to contain 'failures=3'` |
| N8 | drop the `...options` spread from the retry recursion | 1 test | `the RETRY read still carries cacheName: expected undefined to be 'packed:caches:cache-name-probe'` |

N1 and N2 are the two mutants that previously survived a fully green tier. N3b is
the isolation control for R6: it leaves the ledger half green and kills only the
behavioural half, so the two halves are covering different things.

## Round-2 verification (corrected numbers)

The earlier "1206 passed / 8 failed" figure in this PR body was wrong. Re-measured
from a clean run of each ref:

| ref | `pnpm test:packages:run` | files |
|---|---|---|
| merge base `e695de5aae` | 1189 total / **1181 passed** / 8 failed | 84 of 87 |
| round-1 head `f81267c702` | 1211 total / **1203 passed** / 8 failed | 87 of 90 |
| this head `1243de2f7b` (`f098314ea5` is comment-only) | 1216 total / **1208 passed** / 8 failed | 87 of 90 |

head − base = **+27 tests / +3 files**: the three package test files this PR adds
(6 + 12 + 4 at round 1) plus the 5 tests round 2 adds to two of them. The 8
failures are identical at every ref — the `@civitai/db-queries` DB-backed EXPLAIN
suites, all from a single `connect ECONNREFUSED …:15432` because no local Postgres
is running. Connection refusals, not assertions.

Also at `1243de2f7b` (unchanged by the comment-only `f098314ea5`):

* `pnpm typecheck` — **0 type errors** (via `scripts/typecheck.mjs`, which sets the
  8 GB heap cap; a bare `tsc --noEmit` OOMs at exit 134 on this tree).
* `pnpm test:unit:run` — **1646 files / 38251 tests passed**, 3 files + 33 tests
  skipped, **0 failures**.
* `@civitai/redis` **20 files / 193 tests**, `@civitai/telemetry` **3 files / 35
  tests** (191 and 32 at the round-1 head).
* `prettier --check` clean and `eslint` **0 errors** on all 7 changed files; the 2
  remaining `no-explicit-any` warnings in `cache-helpers.ts` are pre-existing and
  untouched by this change.

---

# Round 3 — final audit round (`d3bf7ba68a`)

Four fixes. Two of them are guards from round 2 that were **satisfiable while the
defect they name is present**; the other two close a hole opened by round 2's own
breadcrumb and pin a constant that was bounded from only one side.

## B1 — the bucket edges are re-cut against production measurements

48 **real** `image-meta` values were pulled off the live cache and timed through
the same `promisify(zlib.brotli*)` shape the codec uses (100% brotli-sentinelled;
compressed p50 883 B, uncompressed p50 1,671 B). Per-value codec timings, in
microseconds:

| | p50 | p90 | p99 |
|---|---|---|---|
| dispatch floor (1-byte payload — threadpool round trip, no codec work) | 17.7 | 38.6 | 140.9 |
| decompress (async, the shape that ships) | 21.9–30.2 | 41.9–66.2 | 182.7 |
| compress q6 | 111.6 | 216.3 | 3,803 |

Against those, the round-1 floor of `0.00002` (20 µs) was **still inside the
sample population**: **38.5% of decompress samples land at or below it**. That is
the same mistake the original `0.0005` floor made — a first edge placed where
typical samples land — one order of magnitude smaller and therefore much easier to
call settled. Everything at or below the first edge is indistinguishable, so a
floor inside the population reports a stable, plausible number that barely moves
whatever the codec does.

**The new list is 21 edges:**

```
5, 10, 15, 20, 30, 50, 75, 100, 150, 250 µs,
0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500 ms, 1 s
```

* the floor is **below the measured 17.7 µs dispatch floor**, so bucket one means
  *"impossibly fast — suspect the instrument"* rather than *"typical"*;
* the decompress bulk (10–70 µs) and the compress bulk (100–250 µs) get the finest
  steps in the list, 1.33×–1.67×, because that is where the readable quantiles
  live;
* the ms decade covers compress p99 (~3.8 ms) and the ~36 ms large-blob tail at
  the usual 1-2-5 steps, widest adjacent pair 2.5×.

Cost: 21 edges instead of 15, i.e. **+6 edges × 2 ops × the small set of
compress-aware caches**. Settled before merge, as before — re-cutting resets every
series' history.

## B2 — the bucket guard's bound was satisfiable by deleting an edge, and said it was not

Round 2 replaced a last-edge-only tail check with a bound on the widest adjacent
edge ratio (`<= 4`), and its comment asserted that 4 *"is not a multiple of any
step in the list, so it cannot be satisfied by accident"*.

**That sentence is false.** Deleting an edge does not invent a new ratio — it
**multiplies the two steps either side of it**. A list built from 2× and 2.5×
steps therefore composes to 4×, 5× and 6.25×, and the smallest of those is exactly
the bound. Demonstrated on the round-2 list: deleting any one of `0.0005`,
`0.005`, `0.05` or `0.5` leaves its neighbours at a ratio of **exactly 4**, and
**all 35 tests in the package stay green** — a histogram with half the resolution
across a whole decade, shipped past the guard written to prevent it.

Raising the bound to 3 would close that particular hole and leave the same shape
of guard for the next re-cut to reopen. So the bound is **replaced by an equality
on the ordered sequence of adjacent ratios**, which has no composition hole:
deleting any edge shortens the sequence, changing one moves an entry, inserting
one lengthens it. The comparison is rounded to 3 decimals because 1.5× and 1.667×
are not exact in IEEE754 — ~0.05% resolution, far finer than any step the list
uses.

The old `MAX_EDGE_RATIO` survives, renamed `CONTROL_MAX_EDGE_RATIO` and demoted to
what it can honestly do: discriminating the known-bad **control** lists at the
bottom of the file, which are nowhere near it (2.5× and 4000×). Its comment now
states plainly that no claim is made about it being unreachable by composition —
**it is reachable, and that is why it is no longer the guard.**

A separate test pins the floor below the measured dispatch floor, with a **negative
control** (the round-1 20 µs list, which passes the weaker "below the typical
value" test and fails this one) and a **positive control** (the shipped list must
be accepted, or the assertion is a claim about a predicate that always rejects).

## B3 — the bridge ledger covered only the publisher, and its docstring said otherwise

`redis-metrics-bridge.test.ts` claimed its ledger caught drift *"or vice versa"* —
a handle added to `@civitai/redis` that the app never publishes, which is the
silently-dead-metric case the whole seam exists to prevent. **It cannot.**
`PUBLISHED_HANDLES` is a hand-written literal compared only against
`Object.keys(globalThis.__civitaiRedisMetrics)`, and one person edits both, so
both sides move together.

Demonstrated: add a ninth field to `RedisMetricsBridge` and a real read of it at
the codec timer, leave the publish at eight — **bridge test 2/2 green,
`@civitai/redis` 194/194 green**. Every consumer read is optional-chained, so the
result is a permanently empty series indistinguishable from "the feature is off".

`RedisMetricsBridge` is now **exported** (a type export, erased at build — it adds
nothing to any bundle) and the published object is pinned to it at compile time:

```ts
const redisMetricsBridge = { … } satisfies RedisMetricsBridge &
  Record<keyof RedisMetricsBridge, unknown>;
```

The `Record` half is load-bearing. Most handles are declared **optional** over
there, so a plain `satisfies RedisMetricsBridge` accepts every one of them being
missing — i.e. it accepts exactly the defect. `Record<keyof …, unknown>` makes each
key required at the publish site; intersecting the real type back in keeps the
value shapes checked; and excess-property checking on the literal covers the other
direction.

The check lives in `src/server/prom/client.ts` and **deliberately not** in the test
file: `src/**/__tests__/**` is excluded in `tsconfig.json`, so a type-level
assertion written there would never be evaluated by `pnpm typecheck` — coverage
that reads as coverage while providing none. The docstring now says what the file
covers, what it does not, and where the other direction is enforced.

## B4 — round 2's breadcrumb put unprotected code inside the load-bearing catch

The `catch` in `packedCodecTimer` is documented as 🔴 load-bearing because a throw
escaping it lands in `safeUnpackCompressed`'s catch, **which unlinks the key**.
Round 2 added the breadcrumb *inside* it — an unprotected `log(...)` plus
`err instanceof Error ? err.message : String(err)`. `log` is **injected by the
app** and is therefore arbitrary code, and `String(err)` invokes a
`toString`/`Symbol.toPrimitive` on a value we did not construct. Either throwing
reintroduces the whole observability-becomes-cache-eviction bug the catch exists
to stop, one layer further in, by the very line that reports the swallow.

Now a nested `try { … } catch {}`. The failure counter is still incremented outside
it, so a later successful line carries the count. Low reachability is not totality;
the nesting is what makes the catch total.

## B5 — the throttle constant was pinned only from above

The round-2 test crossed the throttle window and asserted the line returns. That
pins the interval from **above** only — "no larger than the advance" — which every
value down to 1 ms satisfies. Measured: setting
`PACKED_CODEC_OBSERVE_LOG_INTERVAL_MS` to `1` left `@civitai/redis` **193/193
green**, so a `60_000 → 60` unit slip would restore near-per-call log spam under a
sustained fault with nothing red anywhere.

A `+59_999` advance that must **still be throttled** now pins it from below. The
two together bracket the constant into `(59_999, 60_001]`.

## B6 — the message count is deleted rather than corrected a fourth time

The observe-failure comment's claim about how many distinct messages this module's
`log()` carries has been wrong three times: `~20` in round 2, then `35` in a commit
made **solely** to correct it (stating "Counted: 35"); measured at the round-2 head
it is neither — 49 call sites, 47 distinct texts.

**The figure is removed, not re-corrected.** The only claim it ever supported is
that this logger already carries plenty of unrelated traffic, which needs no
number; a count of log sites drifts on every edit and nothing checks it. The
comment now says outright that no count is asserted, and why. Its **other** claim —
that the corrupt-entry line is `Packed (compressed) unpack failed, evicting bad
cache entry` in `safeUnpackCompressed`, not this logger's identity — is correct and
is kept.

## Round-3 mutation table

Applied to the committed code, run, restored, and re-verified green after each.
Bucket mutants were applied by a harness that edits **code lines only** — the first
version of it matched inside the constant's block and silently deleted the `0.5` in
a **comment**, leaving the array untouched and scoring that mutant SURVIVED. The
harness now parses the edge list from non-comment lines and asserts the count drops
by exactly one.

| # | Mutant | Killed by | Failing assertion |
|---|---|---|---|
| P1 | delete edge `0.00003` | 1 test | `the registered edges have exactly the adjacent-ratio sequence this list was designed with…: expected [ 2, 1.5, 1.333, 2.5, 1.5, …(14) ] to deeply equal [ 2, 1.5, 1.333, 1.5, 1.667, …(15) ]` |
| P2 | delete edge `0.000075` | 1 test | same assertion — `expected [ 2, 1.5, 1.333, 1.5, 1.667, 2, …(13) ]` |
| P3 | delete edge `0.00015` | 1 test | same assertion — 20 ratios where 20 are expected but one entry composed |
| P4 | delete edge `0.0005` | 1 test | same assertion |
| P5 | delete edge `0.005` | 1 test | same assertion |
| P6 | delete edge `0.05` | 1 test | same assertion |
| P7 | delete edge `0.5` | 1 test | same assertion |
| P8 | revert the whole list to the round-1 20 µs edge set | 3 tests | `the first REGISTERED edge must sit below the 0.0000177s threadpool dispatch floor…: expected 0.00002 to be less than 0.0000177`, plus the ratio-sequence assertion and the dispatch-floor **positive control** |
| P9 | add a 9th handle to `RedisMetricsBridge` + a real read of it, publish left at 8 | `pnpm typecheck` | `src/server/prom/client.ts(62,3): error TS1360: Type '{ … }' does not satisfy the expected type 'RedisMetricsBridge & Record<keyof RedisMetricsBridge, unknown>'` |
| P10 | remove the nested `try/catch` from `packedCodecTimer` (body kept in a bare block) | 1 test | `the healthy entry survived a throwing LOGGER, not merely a throwing observe(): expected false to be true` |
| P11 | `PACKED_CODEC_OBSERVE_LOG_INTERVAL_MS = 1` | 1 test | `a failure just BELOW the throttle window is still throttled — the window is ~60s, not milliseconds: expected […(2)] to have a length of 1 but got 2` |

**P1–P7 are the mutants the round-2 bound could not see.** Four of them
(`0.0005`, `0.005`, `0.05`, `0.5`) left the round-2 list at a widest adjacent ratio
of exactly 4 and passed a `<= 4` bound.

**P9 has a control, because a red arm alone does not attribute.** With the
`satisfies` clause removed and the mutant otherwise identical, `pnpm typecheck`
returns **0 errors** — so the `satisfies` is what kills it, not something else in
the tree. And under P9 the runtime suites stay green (`@civitai/redis` 194/194, the
bridge test 2/2), which is the point: no runtime test on either side can observe
this defect.

**P10's two positive controls run before the killing assertion** — the throwing
`observe()` was reached, and the throttle then admitted the breadcrumb line and the
injected logger actually threw. Without the second, the test would also pass
against a build where the breadcrumb never fires at all.

## Round-3 verification

Measured at `d3bf7ba68a` in a clean worktree off this branch.

| | before (`f098314ea5`) | after (`d3bf7ba68a`) |
|---|---|---|
| `@civitai/telemetry` | 3 files / **35** tests | 3 files / **38** tests |
| `@civitai/redis` | 20 files / **193** tests | 20 files / **194** tests |
| `src/server/prom/__tests__/redis-metrics-bridge.test.ts` | 1 file / **2** tests | 1 file / **2** tests |
| `pnpm test:packages:run` | 1216 total / 1208 passed / 8 failed *(derived)* | **1220 total / 1212 passed / 8 failed** *(measured)* |
| `pnpm typecheck` | 0 errors | **0 errors** (77 s, 8 GB heap cap) |

The first four rows' "before" figures were measured directly at the pristine
`f098314ea5` worktree. The `test:packages:run` "before" is marked *derived*: it is
the measured 1220 minus the 4 tests this round adds, all of which are in the two
projects measured directly on both sides. It agrees with round 2's independently
measured 1216, which is the cross-check.

The 8 failures are identical before and after: the `@civitai/db-queries` DB-backed
EXPLAIN suites, all from a single `connect ECONNREFUSED …:15432` because no local
Postgres is running. Connection refusals, not assertions.

`prettier --check` clean and `eslint` **0 errors** on all 6 changed files. The
eslint run was positive-controlled — the same invocation on a deliberately bad file
reports `1 problem`, so the clean verdict is about the code and not about a
zero-file glob.

### CI is the authoritative verdict, and it is green on this head

`gh run view 34046090182` at `headSha=d3bf7ba68a`: **20 checks, 19 SUCCESS + 1
SKIPPED, every one terminal** — including *App unit tests + typecheck*, *Package
unit tests*, *ESLint + Prettier (changed files)*, all four *Unit tests* shards,
`tekton / typecheck`, and the preview deploy + smoke suite (66 passed, 0 flaky).
That matters for the section below: CI runs in a checkout that has none of the two
local environment faults, and it agrees.

### `pnpm test:unit:run`, and the two environment faults in it

**1649 files / 38269 tests — 38222 passed, 45 skipped, 2 failed.** Neither failure
is attributable to this change; both were run down to a cause rather than re-run
until green:

* **4 of the 5 failing files** fail on
  `TSConfckParseError: failed to resolve "extends":"./.svelte-kit/tsconfig.json"`
  in `apps/moderator/tsconfig.json` and `apps/auth/tsconfig.json`. `.svelte-kit/`
  is a **generated** SvelteKit artifact and is gitignored
  (`apps/moderator/.gitignore:8`), so it is present in a long-lived clone and
  absent in a fresh checkout. `moderator-restriction-vocabulary.test.ts` passes
  **21/21** in a clone that has it. This is a property of the checkout, identical
  on `main`.
* **`eventloop-watchdog.capture.test.ts > 🔴 classifies an idle wedge as starved`**
  took **12,045 ms** in the full-suite run and **passes in isolation** — a
  wall-time load flake, not an assertion failure.

A first attempt at this tier is discarded and not quoted: the worktree shares the
base clone's `node_modules` by symlink, hence also `node_modules/.vite` and
`.experimental-vitest-cache`, and a concurrent run in the base clone rewrote that
cache mid-run. 814 of 1649 files then failed to **collect**
(`Cannot find module …/deps_ssr/prom-client.js`, `ENOENT …/.experimental-vitest-cache/…`)
— roughly 8,000 tests never executed, while the summary still printed a pass count
for the rest. The run above was re-taken with nothing concurrent.

The one app file this round touches — `src/server/prom/client.ts` — is covered by
`redis-metrics-bridge.test.ts`, 2/2, against the real module, and the change there
is a named `const` plus a `satisfies` clause: runtime-identical to what it replaced,
and `pnpm typecheck` is the only gate that can see the guard it adds.

### Diff shape

6 files, **+332 / −59**. Of the added lines, **114 are executable and 218 are
comment or blank**:

| file | added exec | added comment/blank | kind |
|---|---|---|---|
| `packages/civitai-telemetry/src/client.ts` | 3 | 32 | payload |
| `packages/civitai-redis/src/client.ts` | 12 | 32 | payload |
| `src/server/prom/client.ts` | 5 | 22 | payload |
| `packages/civitai-telemetry/src/__tests__/packed-codec-buckets.test.ts` | 49 | 76 | scaffolding |
| `packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts` | 45 | 31 | scaffolding |
| `src/server/prom/__tests__/redis-metrics-bridge.test.ts` | 0 | 25 | scaffolding |

Payload total: **20 executable / 86 comment**. The ratio is deliberate — three of
the four fixes are corrections to *claims* a previous round made, and the
correction has to be the sentence as much as the code, or the next reader
re-derives the same wrong justification.

### DIFF
--- a/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
+++ b/packages/civitai-redis/src/__tests__/cached-array-compress.test.ts
@@ -367,8 +367,12 @@ describe('SEAM LEDGER: every redis.packed call site in cached-array.ts is compre
     expect(missing).toEqual([]);
   });
 
-  it('packedOptions is derived from the option, not hardcoded', () => {
-    expect(src).toMatch(/const packedOptions = \{ compress \} as const;/);
+  // `cacheName: key` joined the binding when the codec duration histogram landed. It is pinned in
+  // the same assertion because it carries its own hazard: `key` is the cache PREFIX, and the label
+  // must never become the per-id `${key}:${id}` (that would be unbounded prom cardinality). Both
+  // halves are the shorthand/identifier form, i.e. derived from the caller, never hardcoded.
+  it('packedOptions is derived from the options, not hardcoded', () => {
+    expect(src).toMatch(/const packedOptions = \{ compress, cacheName: key \} as const;/);
   });
 });
 
--- a/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
+++ b/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
@@ -197,6 +197,8 @@ describe('packed codec metrics — labels and wiring', () => {
   });
 
   it('does not throw when the app has published no metrics bridge', async () => {
+    // NOTE: absence is a no-op — `?.` short-circuits and no callback runs. That is a strictly
+    // weaker claim than the throwing-bridge block below, which is where the damage lives.
     removeBridge();
     await expect(
       redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE })
@@ -253,5 +255,86 @@ describe('packed codec metrics — labels and wiring', () => {
     );
     expect(readLabels).toEqual(new Set([BUILDER_CACHE]));
     expect(observed.filter((o) => o.op === 'decompress')).toHaveLength(2);
+  });
+});
+
+/**
+ * 🔴 A METRICS FAULT MUST NOT DESTROY A CACHE ENTRY.
+ *
+ * The timer call sits INSIDE `decompressPacked`, which sits inside `safeUnpackCompressed`'s try —
+ * and that catch does not mean "something went wrong", it means "the stored bytes are corrupt", so
+ * it logs, UNLINKS the key and returns null. A bridge whose `observe()` throws (duplicate metric
+ * registration, a rejected label value, an exhausted registry) therefore does not merely lose a
+ * sample: a VALID entry is reported as a miss and DELETED, on every read, for as long as the fault
+ * lasts — an observability change turning into a cache-eviction bug that presents as a cache with
+ * a mysterious 0% hit rate.
+ *
+ * The "no metrics bridge" test above cannot see this: absence short-circuits at `?.` and no
+ * callback runs at all. Absence and throw are different mechanisms, and only one of them is
+ * dangerous.
+ */
+describe('packed codec metrics — a metrics fault must never fail a read or a write', () => {
+  /**
+   * Counts every call BEFORE throwing. The count is the positive control: without it, all three
+   * tests below would also pass against a client that had stopped calling the timer entirely —
+   * "nothing threw" is exactly what a dead wire looks like too. Asserting `attempts > 0` says the
+   * throwing path was really entered and really swallowed.
+   */
+  let attempts = 0;
+  const BOOM = 'prom registry exploded';
+
+  function installThrowingBridge() {
+    attempts = 0;
+    (globalThis as unknown as { __civitaiRedisMetrics?: unknown }).__civitaiRedisMetrics = {
+      packedCodecDuration: {
+        observe: () => {
+          attempts += 1;
+          throw new Error(BOOM);
+        },
+      },
+    };
+  }
+
+  it('a throwing observe() neither loses the value nor unlinks the key on a compressed get', async () => {
+    store.set(KEY, await compressPacked(Buffer.from(pack(record))));
+    installThrowingBridge();
+
+    const got = await redis.packed.get<typeof record>(KEY, {
+      compress: true,
+      cacheName: DIRECT_CACHE,
+    });
+
+    expect(attempts, 'the throwing observe() was actually reached').toBe(1);
+    // Asserted BEFORE the value: the deletion is the damaging half, and it outlives the request.
+    expect(store.has(KEY), 'the healthy entry was NOT unlinked by a metrics fault').toBe(true);
+    expect(got, 'a valid entry is still returned when the metrics bridge throws').toEqual(record);
+  });
+
+  it('a throwing observe() does not fail a compressed set', async () => {
+    installThrowingBridge();
+
+    await expect(
+      redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE })
+    ).resolves.toBeDefined();
+    expect(attempts, 'the throwing observe() was actually reached').toBe(1);
+    expect(store.has(KEY), 'the compressed write still landed').toBe(true);
+  });
+
+  it('a throwing observe() leaves a whole compressed mGet batch intact', async () => {
+    store.set(KEY, await compressPacked(Buffer.from(pack(record))));
+    store.set(KEY_LEGACY, await compressPacked(Buffer.from(pack(record))));
+    installThrowingBridge();
+
+    const got = await redis.packed.mGet<typeof record>([KEY, KEY_LEGACY], {
+      compress: true,
+      cacheName: DIRECT_CACHE,
+    });
+
+    expect(attempts, 'the throwing observe() was reached once per compressed value').toBe(2);
+    expect(
+      [store.has(KEY), store.has(KEY_LEGACY)],
+      'no key in the batch was unlinked by a metrics fault'
+    ).toEqual([true, true]);
+    expect(got, 'every entry in the batch still decodes').toEqual([record, record]);
   });
 });
--- a/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
+++ b/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
@@ -74,7 +74,11 @@ vi.mock('redis', () => {
 });
 
 import { createCacheBuilders } from '../cached-array';
-import { createCacheRedis, PACKED_CODEC_UNNAMED_CACHE } from '../client';
+import {
+  createCacheRedis,
+  PACKED_CODEC_UNNAMED_CACHE,
+  __resetPackedCodecObserveFailureState,
+} from '../client';
 import type { RedisKeyTemplateCache } from '../client';
 import { compressPacked } from '../packed-compression';
 
@@ -336,5 +340,106 @@ describe('packed codec metrics — a metrics fault must never fail a read or a w
       'no key in the batch was unlinked by a metrics fault'
     ).toEqual([true, true]);
     expect(got, 'every entry in the batch still decodes').toEqual([record, record]);
+  });
+});
+
+/**
+ * 🔴 A SWALLOWED FAILURE MUST NOT BE A SILENT ONE.
+ *
+ * The block above proves the catch protects the cache. It says nothing about whether anyone would
+ * ever find out. A bridge whose `observe()` throws on every call yields a histogram that never has
+ * data — which is byte-for-byte what "no cache opted into compress" looks like, on the only metric
+ * that can answer whether compression is worth its cost. A reassuring zero is indistinguishable
+ * from a probe wired to nothing, so the catch has to say so.
+ *
+ * It is rate-limited rather than one-shot: a throwing bridge throws on EVERY codec call, so an
+ * unthrottled line would be per-cache-read log spam, while a strict one-shot would go quiet for the
+ * whole life of the process no matter how long the fault lasted. The three properties below are
+ * asserted in one test on purpose — the throttle is module-scoped state, so "did not log" only
+ * means "throttled" relative to a known earlier log in the same sequence.
+ */
+describe('packed codec metrics — a swallowed observe failure leaves a breadcrumb', () => {
+  const BREADCRUMB = 'packed codec metrics observe FAILED';
+  const BOOM_MSG = 'prom registry exploded';
+  let attempts = 0;
+
+  function installThrowingBridgeHere() {
+    attempts = 0;
+    (globalThis as unknown as { __civitaiRedisMetrics?: unknown }).__civitaiRedisMetrics = {
+      packedCodecDuration: {
+        observe: () => {
+          attempts += 1;
+          throw new Error(BOOM_MSG);
+        },
+      },
+    };
+  }
+
+  it('logs the first failure, throttles the next, and speaks again once the window has passed', async () => {
+    __resetPackedCodecObserveFailureState();
+    const lines: string[] = [];
+    // `log` is module-scoped in ../client and injected here; this rebinds it for the rest of the
+    // file, which is why the bridge is restored in the `finally` below.
+    createCacheRedis({ log: (msg: string) => lines.push(String(msg)) });
+    installThrowingBridgeHere();
+
+    // A controlled clock: the throttle reads Date.now(). Nothing else on this path does — the
+    // deadline wrapper uses setTimeout — so freezing it cannot stall the client.
+    let clock = 1_700_000_000_000;
+    const nowSpy = vi.spyOn(Date, 'now').mockImplementation(() => clock);
+    const breadcrumbs = () => lines.filter((l) => l.includes(BREADCRUMB));
+
+    try {
+      await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
+
+      // Positive control: the throwing path was actually entered. Without it every "logged N
+      // times" assertion below would also hold against a client that stopped timing entirely.
+      expect(attempts, 'the throwing observe() was actually reached').toBe(1);
+      expect(breadcrumbs(), 'the FIRST swallowed failure is reported').toHaveLength(1);
+      // The line has to be diagnosable on its own: which op, which cache, and the underlying
+      // error. A bare "metrics failed" would not distinguish this from any other swallow.
+      expect(breadcrumbs()[0]).toContain('op=compress');
+      expect(breadcrumbs()[0]).toContain(DIRECT_CACHE);
+      expect(breadcrumbs()[0]).toContain(BOOM_MSG);
+      expect(breadcrumbs()[0]).toContain('failures=1');
+
+      // Same window → counted, not logged.
+      await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
+      expect(attempts, 'the second failure really happened').toBe(2);
+      expect(
+        breadcrumbs(),
+        'a second failure inside the window is throttled, not logged'
+      ).toHaveLength(1);
+
+      // Past the window → speaks again, and carries the count accumulated while it was quiet.
+      clock += 60_001;
+      await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
+      expect(breadcrumbs(), 'the breadcrumb returns after the throttle window').toHaveLength(2);
+      expect(
+        breadcrumbs()[1],
+        'the throttled failures are still counted, so the line reports 3 rather than 1'
+      ).toContain('failures=3');
+    } finally {
+      nowSpy.mockRestore();
+      createCacheRedis({ log: () => undefined });
+      installBridge();
+      __resetPackedCodecObserveFailureState();
+    }
+  });
+
+  it('NEGATIVE CONTROL: a healthy bridge logs no breadcrumb at all', async () => {
+    __resetPackedCodecObserveFailureState();
+    const lines: string[] = [];
+    createCacheRedis({ log: (msg: string) => lines.push(String(msg)) });
+    installBridge();
+
+    try {
+      await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
+      expect(observed, 'the healthy bridge really did receive the sample').toHaveLength(1);
+      expect(lines.filter((l) => l.includes(BREADCRUMB))).toHaveLength(0);
+    } finally {
+      createCacheRedis({ log: () => undefined });
+      __resetPackedCodecObserveFailureState();
+    }
   });
 });
--- a/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
+++ b/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
@@ -354,9 +354,11 @@ describe('packed codec metrics — a metrics fault must never fail a read or a w
  *
  * It is rate-limited rather than one-shot: a throwing bridge throws on EVERY codec call, so an
  * unthrottled line would be per-cache-read log spam, while a strict one-shot would go quiet for the
- * whole life of the process no matter how long the fault lasted. The three properties below are
- * asserted in one test on purpose — the throttle is module-scoped state, so "did not log" only
- * means "throttled" relative to a known earlier log in the same sequence.
+ * whole life of the process no matter how long the fault lasted. The properties below are asserted
+ * in one test on purpose — the throttle is module-scoped state, so "did not log" only means
+ * "throttled" relative to a known earlier log in the same sequence. The window is walked from BOTH
+ * sides (just under it, then just over it) so the interval constant is bracketed rather than merely
+ * bounded from above; see the comment at the 59_999 advance for why the one-sided version was inert.
  */
 describe('packed codec metrics — a swallowed observe failure leaves a breadcrumb', () => {
   const BREADCRUMB = 'packed codec metrics observe FAILED';
@@ -411,16 +413,84 @@ describe('packed codec metrics — a swallowed observe failure leaves a breadcru
         'a second failure inside the window is throttled, not logged'
       ).toHaveLength(1);
 
+      // 🔴 LOWER BOUND ON THE INTERVAL, and it is the half that is easy to leave out. Crossing the
+      // window (below) pins the constant only from ABOVE — it says "no larger than the advance",
+      // which every value down to 1 ms satisfies. Measured: with the interval set to 1 the whole
+      // package stayed 193/193 green, so a 60_000 → 60 unit slip would restore near-per-call log
+      // spam under a sustained fault with nothing red. This advance stops JUST SHORT of the window
+      // and must still be throttled, which pins it from BELOW; the two together bracket the
+      // constant into (59_999, 60_001].
+      clock += 59_999;
+      await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
+      expect(attempts, 'the third failure really happened').toBe(3);
+      expect(
+        breadcrumbs(),
+        'a failure just BELOW the throttle window is still throttled — the window is ~60s, not milliseconds'
+      ).toHaveLength(1);
+
       // Past the window → speaks again, and carries the count accumulated while it was quiet.
-      clock += 60_001;
+      clock += 2; // 60_001 since the first (and so far only) breadcrumb
       await redis.packed.set(KEY, record, undefined, { compress: true, cacheName: DIRECT_CACHE });
       expect(breadcrumbs(), 'the breadcrumb returns after the throttle window').toHaveLength(2);
       expect(
         breadcrumbs()[1],
-        'the throttled failures are still counted, so the line reports 3 rather than 1'
-      ).toContain('failures=3');
+        'the throttled failures are still counted, so the line reports 4 rather than 1'
+      ).toContain('failures=4');
     } finally {
       nowSpy.mockRestore();
+      createCacheRedis({ log: () => undefined });
+      installBridge();
+      __resetPackedCodecObserveFailureState();
+    }
+  });
+
+  // 🔴 THE REPORTING PATH IS ITSELF INSIDE THE LOAD-BEARING CATCH, and it has to be, because `log`
+  // is INJECTED BY THE APP and is therefore arbitrary code. A throw from it escapes the swallow and
+  // lands in `safeUnpackCompressed`'s catch — which UNLINKS THE KEY. That is the same
+  // observability-change-becomes-cache-eviction bug the outer catch exists to stop, reintroduced
+  // one layer further in by the very line that reports the swallow. Verified by mutation: remove
+  // the nested try/catch in `packedCodecTimer` and this test deletes a healthy entry.
+  it('a throwing log() on the breadcrumb path still does not unlink a healthy entry', async () => {
+    __resetPackedCodecObserveFailureState();
+    store.set(KEY, await compressPacked(Buffer.from(pack(record))));
+    // Throws ONLY on the breadcrumb line, not on every call: `createCacheRedis` logs during client
+    // construction, so a globally-throwing logger would blow up before reaching the path under
+    // test — and a real logger failing on one payload (an over-long line, a serialiser choking on
+    // an interpolated value) is the shape this is actually defending against anyway.
+    let breadcrumbThrows = 0;
+    createCacheRedis({
+      log: (msg: string) => {
+        if (!String(msg).includes(BREADCRUMB)) return;
+        breadcrumbThrows += 1;
+        throw new Error('the injected logger exploded');
+      },
+    });
+    installThrowingBridgeHere();
+
+    try {
+      const got = await redis.packed.get<typeof record>(KEY, {
+        compress: true,
+        cacheName: DIRECT_CACHE,
+      });
+
+      // Positive controls, both needed. `attempts` says the observe threw, so the outer catch ran;
+      // `breadcrumbThrows` says the throttle then ADMITTED the line, the logger really was called,
+      // and it really threw. Without the second, this test would also pass against a build where
+      // the breadcrumb never fires at all — i.e. against no guard being exercised.
+      expect(attempts, 'the throwing observe() was reached, so the swallow path really ran').toBe(
+        1
+      );
+      expect(
+        breadcrumbThrows,
+        'the breadcrumb line was actually emitted and the injected logger actually threw'
+      ).toBe(1);
+      // Asserted before the value: the deletion is the damaging half and it outlives the request.
+      expect(
+        store.has(KEY),
+        'the healthy entry survived a throwing LOGGER, not merely a throwing observe()'
+      ).toBe(true);
+      expect(got, 'a valid entry is still returned').toEqual(record);
+    } finally {
       createCacheRedis({ log: () => undefined });
       installBridge();
       __resetPackedCodecObserveFailureState();
--- a/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
+++ b/packages/civitai-redis/src/__tests__/packed-codec-metrics.test.ts
@@ -1,0 +1,257 @@
+import { pack } from 'msgpackr';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * SEAM coverage for the packed brotli codec's prom instrumentation
+ * (`civitai_app_packed_codec_duration_seconds`, labels `op` + `cache_name`).
+ *
+ * WHY THIS DRIVES THE REAL CLIENT. The unit test beside this one
+ * (./packed-codec-timing.test.ts) proves `compressPacked`/`decompressPacked` CALL their timing
+ * callback. That is a claim about one module. The thing that can actually be wrong in production
+ * is the SEAM: whether `client.packed.set/get/mGet` pass a callback at all, whether it reaches the
+ * globalThis metrics bridge the app publishes, and whether the `cache_name` it carries is the
+ * bounded cache PREFIX rather than a per-id key. A test that mocked `packed` would pass against a
+ * client that instruments nothing.
+ *
+ * So the only faked thing here is the TRANSPORT: `redis`'s client factory is replaced by an
+ * in-memory server holding raw Buffers (same approach as ./cached-array-compress.test.ts).
+ * `createCacheRedis`, `client.packed.*`, the msgpack + brotli codec, `createCacheBuilders` and the
+ * bridge lookup are all the real modules.
+ */
+
+// process.env must be populated BEFORE ../client is imported: loadRedisEnv() validates it.
+vi.hoisted(() => {
+  process.env.REDIS_URL ??= 'redis://localhost:6379';
+  process.env.REDIS_SYS_URL ??= 'redis://localhost:6379';
+  process.env.REDIS_CLUSTER = 'false';
+});
+
+/** The in-memory "server": key -> raw stored bytes, exactly as a real Redis would hold them. */
+const { store } = vi.hoisted(() => ({ store: new Map<string, Buffer>() }));
+
+vi.mock('redis', () => {
+  const toBuffer = (v: unknown): Buffer =>
+    Buffer.isBuffer(v) ? v : Buffer.from(v as string, 'binary');
+
+  const makeClient = (bufferMode = false): Record<string, unknown> => {
+    const self: Record<string, any> = {
+      connect: () => Promise.resolve(self),
+      on: () => self,
+      withTypeMapping: () => makeClient(true),
+      get: async (key: string) => {
+        const v = store.get(key);
+        if (v === undefined) return null;
+        return bufferMode ? v : v.toString('binary');
+      },
+      set: async (key: string, value: unknown, options?: { NX?: boolean; XX?: boolean }) => {
+        const exists = store.has(key);
+        if (options?.NX && exists) return null;
+        if (options?.XX && !exists) return null;
+        store.set(key, toBuffer(value));
+        return 'OK';
+      },
+      del: async (key: string) => (store.delete(key) ? 1 : 0),
+      unlink: async (key: string) => (store.delete(key) ? 1 : 0),
+      eval: async (_script: string, opts: { keys: string[] }) => {
+        const key = opts.keys[0];
+        if (store.has(key)) return 0;
+        store.set(key, Buffer.from('1'));
+        return 1;
+      },
+      scanIterator: async function* () {
+        /* unused */
+      },
+    };
+    return self;
+  };
+
+  return {
+    createClient: () => makeClient(false),
+    createCluster: () => makeClient(false),
+    createSentinel: () => makeClient(false),
+    RESP_TYPES: { BLOB_STRING: 36 },
+  };
+});
+
+import { createCacheBuilders } from '../cached-array';
+import { createCacheRedis, PACKED_CODEC_UNNAMED_CACHE } from '../client';
+import type { RedisKeyTemplateCache } from '../client';
+import { compressPacked } from '../packed-compression';
+
+const redis = createCacheRedis();
+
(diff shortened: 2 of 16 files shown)


ZacxDev:
Adversarial audit round 1 (first, full) ran against `d172dc74cb`; fixes landed at `f81267c7`. Eight findings, none deploy-blocking. What the fix round claims to have addressed:

```audit-claims round=1 audited=d172dc74cb..f81267c7
1. Histogram bucket floor lowered from 0.5ms by prepending 0.00002/0.00005/0.0001/0.00025, plus a guard test pinning that the first edge sits below a typical decompress sample.
2. packedCodecTimer body wrapped in try/catch, so a throwing metrics bridge can no longer be caught by safeUnpackCompressed's corrupt-entry handler and unlink a healthy cache entry.
3. fetchThroughCache's retry recursion now threads cacheName, which it previously dropped while threading compress.
4. Test coverage added for the fetchThroughCache half of the codec seam, killing two mutants that previously survived the full unit tier (making the option inert; dropping it at the tensor-metadata call site).
5. The "cache_name is the same value the cache hit/miss counters already carry" claim corrected to name the consumer for which the join does not hold; no replacement rationale was substituted.
6. Help string and codec docs now state the sample includes libuv threadpool queue wait, not codec CPU alone.
7. Findings 5 (bustFetchThroughCache has no cacheName option) and 8 (cacheName is unvalidated) recorded in the PR body as known-open and deliberately not fixed.
```

Ledger: round 1 · payload lines changed this round: 84 added / 16 removed across 4 source files · scaffolding: 535 added / 7 removed across 5 test files.

ZacxDev:
Adversarial audit round 2 (delta) ran against `f81267c702`; fixes landed at `f098314ea5`. Five findings, no deploy-blockers; verdict was "safe to merge" but the findings were real, so the ladder continued. What this fix round claims to have addressed:

```audit-claims round=2 audited=f81267c702..f098314ea5
1. Bucket guard rewritten to observe a sample and read the REGISTERED le set back off prom-client rather than asserting on the exported constant, plus a max-adjacent-edge-ratio predicate replacing the last-edge-only check; the two mutants that previously survived (buckets reverted at the registration site; constant collapsed to a single wide bucket) now both fail.
2. The packedCodecTimer catch now emits a rate-limited breadcrumb carrying a failure count, so a bridge that throws on every observe is no longer indistinguishable from no cache opting into compression; the comment's false claim that the module logger can only report "corrupt entry" was corrected.
3. The codec docs and the shipped help string now name event-loop delay as an inflation path alongside threadpool queue wait, and name the dispatch floor so the low buckets are not read as codec time.
4. fetchThroughCache's retry recursion switched to spreading the options object, so a future sixth option cannot be silently dropped.
5. The "Wall time of the brotli codec" framing was swept at the sibling site 48 lines from where it had already been corrected.
6. The PR body's test:packages:run figure was corrected against a re-measurement, with a base/head table replacing the single quoted number.
7. A guard was added for the metrics-bridge publish line, which was previously unpinned: deleting it made every codec sample vanish with no error because absence short-circuits at the optional call.
8. A follow-up commit corrected this fix round's own comment, which asserted "~20 distinct messages" where the counted value is 35.
```

Ledger: round 2 · payload lines changed this round: 110 added / 37 removed across 4 source files · scaffolding: 326 added / 14 removed across 3 test files · since round 1: payload 194 added / 53 removed.

ZacxDev:
Adversarial audit round 3 (delta) ran against `f098314ea5`; fixes landed at `d3bf7ba68a`. Verdict was "safe to merge" with no deploy-blockers. What this fix round claims to have addressed:

```audit-claims round=3 audited=f098314ea5..d3bf7ba68a
1. Bucket edges re-cut against measured codec timings on real cached values, with the first edge moved below the measured threadpool dispatch floor so bucket 1 means "impossibly fast" rather than "typical"; previously 38.5% of decompress samples landed at or below the first edge.
2. The bucket guard now asserts the exact ordered adjacent-ratio sequence instead of a maximum-ratio bound, closing the composition hole where two consecutive 2x steps satisfied a 4x bound and four interior edges could be deleted with the suite green.
3. The old maximum-ratio bound was kept only as a discriminator for two known-bad control lists, with its comment stating plainly that no claim is made about it being unreachable by composition, because it is reachable.
4. The metrics-bridge ledger gained a compile-time exhaustiveness check, so a handle added to the bridge type but never published now fails typecheck rather than shipping as a silently dead metric.
5. The load-bearing catch regained totality via a nested try/catch, so a throwing logger inside it can no longer reach the caller that unlinks the cache entry.
6. The throttle interval gained a lower-bound assertion; it was previously pinned only from above, so a unit slip to milliseconds passed.
7. The message count in the observe-failure comment was deleted rather than corrected a fourth time, after being wrong three times in the same sentence.
8. The PR body's duplicated prose figures were removed in favour of the single measured table they had come to contradict.
9. The shipped help string's dispatch-floor figure was corrected to the measured value.
```

Ledger: round 3 · payload lines changed this round: 20 executable / 86 comment across 3 source files · scaffolding: 94 executable / 132 comment across 3 test files.

---

## The audit ladder is CLOSED here, and this is why it is not being closed on convergence

Three rounds, no round returned a deploy-blocker, and every claimed fix was verified real with its killing mutant independently re-derived. The ladder is being stopped on the **stated criterion for a payload that has become prose**, not because findings ran out — and the difference matters, so it is recorded rather than left implicit:

- **The rounds will not stop on their own.** Round 3 reported 5 of its 8 findings against scaffolding or prose the ladder itself had written in round 2. This round's payload is **20 executable lines against 86 comment lines**, and three of its four fixes correct a *claim a previous round made* rather than code. Each round now writes the prose that becomes the next round's findings, so the attribution gate — which counts payload lines — cannot fire, and a findings-keyed stop cannot either.
- **The blast radius of what remains is "a sentence in a comment is imprecise."** No finding in any round could break a request path.
- **The recurring shape was swept at every site, not just the reported one:** the "claim corrected at one site, missed at its sibling" defect appeared in round 2 and again in round 3 one layer up in the PR body; both are now closed, and `grep "Wall time"` over `src/` and `packages/` returns zero.

## Known open — deliberately not fixed, so this reads as open rather than absent

1. **The observe-failure breadcrumb depends on the app's logging configuration.** If that channel is off, a metrics bridge throwing on every observe still produces an empty histogram, which is indistinguishable from "no cache opted into compression". Decision taken: leave it — the in-code comment states the dependency honestly rather than hiding it. Closing condition: either the deployment enables that channel, or a config-independent signal is added; checked by whoever next edits `packedCodecTimer`.
2. **`bustFetchThroughCache` has no `cacheName` option** and `cacheName` is unvalidated free-form text next to a key that is routinely `prefix:id`. Both latent — no compressed caller reaches the first today, and both live callers pass a constant to the second.
3. **The codec's dispatch overhead is tracked separately as #4656** — roughly two thirds of an async decompress is threadpool dispatch rather than compression, which is a hot-path optimisation and deliberately not folded into an observability PR.

Verification of this final round, done independently of the fix: the shipped edge set was checked against timings of **48 real cached values** — every measured percentile now lands in a distinct bucket, nothing falls at or below the first edge, and the largest single bucket holds 36.7% where one previously held the p50 hostage. The `0.5` edge deletion — the mutant whose first harness mis-scored it by editing a comment instead of the array — was re-run with the comment confirmed untouched, and fails on the ratio-sequence assertion.

PR #4658, opened 2026-09-06
feat(moderation): read-through cache for external prompt-moderation verdicts (off by default)

ZacxDev:
Ships the verdict cache the dark probe was built to size. **Off by default** — it does nothing until an allowlisted namespace *and* a positive TTL are both configured.

The probe's own source said this had to be its own change with its own review, because caching a moderation verdict is a **trust-and-safety** decision rather than a performance one. That decision was taken deliberately, not arrived at.

## Why

`moderatePrompt` is an inline, serial outbound call on the generation submit path — **~205 ms at p50 ~188 ms**, and **78–88 %** of the app-side time of `generateFromGraph`. Three ways to make the *call* cheaper were tested in production and all three are dead:

| lever | why it's dead |
|---|---|
| clamp the slow tail | the distribution is a **floor**, not a tail — zero calls under 50 ms, 0.46 % under 100 ms |
| reuse the connection | the classifier answers from a nearby edge; the whole handshake is ~7–24 ms |
| overlap with the submit | impossible — a flagged prompt must never reach `submitWorkflow` |

Best infra lever: ~10 ms. Not making the call: **~70 ms**. The probe measured **34.3 %** repeats at a 5m window over **224,989 observations**, load-invariant across the full weekly range (+1.63 points over a ~33 % swing in volume).

## The staleness answer is mechanical, not procedural

The verdict is **not a function of the prompt alone**. `moderatePrompt` derives `flagged`/`categories` from the raw scores using `EXTERNAL_MODERATION_THRESHOLD` and, when set, `EXTERNAL_MODERATION_CATEGORIES`, and asks a specific `model`. So the key carries a digest of all four:

```
generation:moderation-verdict:<namespace>:<policyDigest>:<promptDigest>
```

Change the threshold, the category map or the model and **every previously-cached verdict becomes unreachable at once** — no flush step, nothing for an operator to remember; old entries age out on their own `EX`. The digest is stable under a category-key *reorder*, so an unrelated config reshuffle doesn't flush the cache.

The residual exposure is a classifier that changes behind a stable model name (`omni-moderation-latest` is a moving target by construction). Only the TTL bounds that — hence the 3600 s cap.

## Deliberately not cached

- **Failures.** Only an `ok` outcome is stored. The caller is fail-**soft** on error — it proceeds with `flagged:false` and the local regex audit still gates — so caching one transient gateway error would be TTL-long silent under-moderation of that exact prompt.
- **Cache errors.** A Redis failure resolves to a miss and the classifier is called. There is no path from a cache problem to a weaker verdict; worst case is the latency we already pay.

## Arming

Arming takes **two** inputs, both required: `EXTERNAL_MODERATION_CACHE_NAMESPACE=prod` (a closed allowlist) **and** a positive `EXTERNAL_MODERATION_CACHE_TTL_SECONDS`. Unarmed there are **no metric series at all**.

⚠️ **This section originally said the opposite** — that the TTL was the sole switch, that a second input would be a thing that could disagree with the first, and that reading it per call made it "a real kill switch rather than one needing a rollout." Audit rounds 1 and 2 retracted all three. Deployments share one sysRedis and the PR-preview task copies `civitai-cfg` wholesale, so a namespace segment is mandatory; and `env` is a module-load snapshot, so **disabling requires a rollout**. Corrected here as well as in the code, because this is the first thing a reviewer reads.

## ⚠️ One thing to know about the metrics

**A hit is not observed on `external_moderation_duration_seconds`.** Recording hits as ~0 ms samples would drag down the p50 of the one instrument that measures what a classifier call costs — the same instrument any future decision about the model has to be graded on.

Consequence: once armed, that histogram's `_count` is **classifier calls**, while the new `external_moderation_cache_total` counts moderation **checks**. The two diverge by exactly the hit count. Anyone comparing a before/after on that histogram needs to know the denominator changed meaning on the day the cache armed.

## Also: a latent type bug

Widens `moderatePrompt`'s return from `{ flagged: false }` to `boolean`. That literal was a **bug, not a safety property** — the function returns `flagged: true` whenever the classifier flags, and only type-checked because `results[0].flagged` comes off an untyped `res.json()`, so `any` flowed past the annotation. Any consumer TS had narrowed to "never flagged" was being told something false about a moderation gate. Typecheck is clean after the widening.

## Verification

- **68/68** across all four moderation suites (28 in the new one)
- **Typecheck OK, 0 errors in 103 s** — content read, not an exit code
- **Prettier clean**, confirmed against a positive control first, since "All matched files use Prettier code style!" is also what it prints when the glob matched nothing

Mutation-tested, each mutant the narrowest expression that can be wrong so it dies for its own guard's reason:

| mutation | result |
|---|---|
| policy digest → a hex constant (`policyDigest`'s body) | 4 tests fail |
| namespace segment dropped from the key | 4 tests fail, incl. the two-deployment case |
| namespace allowlist widened to any non-empty string | 2 tests fail |
| **write-side** arming guard removed | fails alone with `expected '…:null:2c…' to contain ':prod:'` |
| the sys-read deadline wrapper removed | hangs its test to a vitest timeout |
| a failed call is cached | "failure must cost a retry" fails, alone |
| deserialization loosened to truthiness | malformed-value test fails, alone — `f: 0` must not read as "not flagged" |
| a hit observed on the histogram | no-latency-sample test fails, alone |
| TTL arming guard removed | 2 tests fail (both TTL cases) |

**Every row above was re-measured at the current head.** Two of them had drifted since they were written — the segment row gained a fourth kill from the new keyspace assertion, and the TTL row stopped being "alone" once `moderationCacheEnabled()` began routing through `armedCache()`. Both drifted in the understating direction, and both were caused by the very commit that claimed to have corrected this table.

🔴 Two mutation-testing mistakes of mine, both caught by re-running rather than assuming. The first policy mutation hit the **read path alone**, desynchronising the keys and taking out 7 tests — proving the *key* matters, not the *policy segment*. And in round 2 the **write-side arming guard survived** a green 28/28 suite; the obvious fix (`expect(set).not.toHaveBeenCalled()`) was *also* vacuous, because the write is fire-and-forget behind a lazy `import()` and had not happened yet either way. An absence you did not wait for is not an absence.

## Rollout

Merge is inert. To arm, set **both** `EXTERNAL_MODERATION_CACHE_NAMESPACE=prod` and a positive TTL on one deployment, then read `external_moderation_cache_total`; the hit rate should land at or **below** 34.3 %, because the probe claimed its slot when a request *started* whereas this cache cannot store a verdict until the classifier answers ~200 ms later.

https://claude.ai/code/session_01SsSvftTXXvxBixDoMq7FTM

### DIFF
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -2007,8 +2007,9 @@ export const REDIS_SYS_KEYS = {
      * segment after THAT is a truncated SHA-256 of the exact string sent to the classifier; the
      * prompt itself is never stored. Every key carries an EX.
      *
-     * Written only while EXTERNAL_MODERATION_CACHE_TTL_SECONDS is a positive integer — that TTL is
-     * the arming switch, and there is deliberately no separate boolean to disagree with it.
+     * Written only while BOTH EXTERNAL_MODERATION_CACHE_TTL_SECONDS is a positive integer AND
+     * EXTERNAL_MODERATION_CACHE_NAMESPACE names an allowlisted deployment — two inputs, because one
+     * says how long entries live and the other says where; neither is inferable from the other.
      * See src/server/integrations/moderation-verdict-cache.ts.
      */
     MODERATION_VERDICT: 'generation:moderation-verdict',
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1999,7 +1999,9 @@ export const REDIS_SYS_KEYS = {
      * Read-through cache of external prompt-moderation VERDICTS — unlike MODERATION_CACHE_PROBE
      * above, this one IS read back, and a hit skips the classifier call entirely.
      *
-     * The segment after this prefix is a digest of the POLICY that produced the verdict (model,
+     * The segment after this prefix is the DEPLOYMENT namespace — several deployments share one
+     * sysRedis and sys keys carry no environment segment, so without it a preview could serve a
+     * verdict to production. The segment after THAT is a digest of the POLICY (model,
      * score threshold, category map), not a deployment label: changing any of them changes every
      * key, so a policy change invalidates the whole cache atomically with no flush step. The
      * segment after THAT is a truncated SHA-256 of the exact string sent to the classifier; the
--- a/packages/civitai-redis/src/client.ts
+++ b/packages/civitai-redis/src/client.ts
@@ -1995,6 +1995,21 @@ export const REDIS_SYS_KEYS = {
      * See src/server/integrations/moderation-cache-probe.ts.
      */
     MODERATION_CACHE_PROBE: 'generation:moderation-cache-probe',
+    /**
+     * Read-through cache of external prompt-moderation VERDICTS — unlike MODERATION_CACHE_PROBE
+     * above, this one IS read back, and a hit skips the classifier call entirely.
+     *
+     * The segment after this prefix is a digest of the POLICY that produced the verdict (model,
+     * score threshold, category map), not a deployment label: changing any of them changes every
+     * key, so a policy change invalidates the whole cache atomically with no flush step. The
+     * segment after THAT is a truncated SHA-256 of the exact string sent to the classifier; the
+     * prompt itself is never stored. Every key carries an EX.
+     *
+     * Written only while EXTERNAL_MODERATION_CACHE_TTL_SECONDS is a positive integer — that TTL is
+     * the arming switch, and there is deliberately no separate boolean to disagree with it.
+     * See src/server/integrations/moderation-verdict-cache.ts.
+     */
+    MODERATION_VERDICT: 'generation:moderation-verdict',
   },
   TRAINING: {
     STATUS: 'training:status',
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -656,7 +656,20 @@ export const serverSchema = z
     // bound on a classifier whose model can change behind a stable name; the measured value of a
     // longer window is small anyway (12x the TTL bought ~7 points of hit rate).
     // See src/server/integrations/moderation-verdict-cache.ts.
-    EXTERNAL_MODERATION_CACHE_TTL_SECONDS: z.coerce.number().int().min(0).max(3600).default(0),
+    // The deployment this cache writes under. REQUIRED to arm it — several civitai-web deployments
+    // share one sysRedis and sys keys carry no environment segment, and the PR-preview task copies
+    // civitai-cfg WHOLESALE, so the TTL below is inherited by every open preview. A closed
+    // allowlist lives in moderation-verdict-cache.ts (one rule, one place); anything outside it is
+    // OFF and logged once. See that module for why the policy digest cannot substitute for this.
+    EXTERNAL_MODERATION_CACHE_NAMESPACE: z.string().trim().optional().default(''),
+    // 🔴 `.catch(0)`, NOT `.default(0)` — the same rule TRPC_MAX_BATCH_SIZE and
+    // EXTERNAL_MODERATION_TIMEOUT_MS carry above. `src/env/server.ts` THROWS on any invalid field,
+    // and env is parsed only at container start, so a typo here (`=off`, `=false`, `=3600s`,
+    // `=7200` over the cap) does nothing visible at the time and then CrashLoops the whole fleet at
+    // the next rollout, hours detached from the change — during the very incident this lever exists
+    // to end. `.catch(0)` degrades an unparseable value to OFF, which is the safe direction for a
+    // cache in front of a moderation gate.
+    EXTERNAL_MODERATION_CACHE_TTL_SECONDS: z.coerce.number().int().min(0).max(3600).catch(0),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
 
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -647,6 +647,16 @@ export const serverSchema = z
     // as perfectly good namespaces, so the likeliest spelling of "turn this off" would ARM the
     // probe. See src/server/integrations/moderation-cache-probe.ts.
     EXTERNAL_MODERATION_CACHE_PROBE: z.string().trim().optional().default(''),
+    //
+    // Seconds to hold a cached external-moderation verdict. THIS IS THE ARMING SWITCH for the
+    // verdict cache: absent or 0 means the cache is inert and emits no metric series at all.
+    // There is deliberately no separate boolean — a second on/off input is a second thing that can
+    // disagree with the first, which is the trap the probe's namespace allowlist above exists to
+    // avoid. Capped at 3600 because a cached verdict is a STALE verdict and the TTL is the only
+    // bound on a classifier whose model can change behind a stable name; the measured value of a
+    // longer window is small anyway (12x the TTL bought ~7 points of hit rate).
+    // See src/server/integrations/moderation-verdict-cache.ts.
+    EXTERNAL_MODERATION_CACHE_TTL_SECONDS: z.coerce.number().int().min(0).max(3600).default(0),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
 
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -648,20 +648,33 @@ export const serverSchema = z
     // probe. See src/server/integrations/moderation-cache-probe.ts.
     EXTERNAL_MODERATION_CACHE_PROBE: z.string().trim().optional().default(''),
     //
-    // Seconds to hold a cached external-moderation verdict. THIS IS THE ARMING SWITCH for the
-    // verdict cache: absent or 0 means the cache is inert and emits no metric series at all.
-    // There is deliberately no separate boolean — a second on/off input is a second thing that can
-    // disagree with the first, which is the trap the probe's namespace allowlist above exists to
-    // avoid. Capped at 3600 because a cached verdict is a STALE verdict and the TTL is the only
-    // bound on a classifier whose model can change behind a stable name; the measured value of a
-    // longer window is small anyway (12x the TTL bought ~7 points of hit rate).
-    // See src/server/integrations/moderation-verdict-cache.ts.
-    // The deployment this cache writes under. REQUIRED to arm it — several civitai-web deployments
-    // share one sysRedis and sys keys carry no environment segment, and the PR-preview task copies
-    // civitai-cfg WHOLESALE, so the TTL below is inherited by every open preview. A closed
-    // allowlist lives in moderation-verdict-cache.ts (one rule, one place); anything outside it is
-    // OFF and logged once. See that module for why the policy digest cannot substitute for this.
+    // 🔴 THE VERDICT CACHE NEEDS BOTH OF THE NEXT TWO VARIABLES. Neither alone arms it: this one
+    // says WHERE entries live, the TTL below says HOW LONG. Setting only one leaves the cache inert
+    // and emitting no metric series, which looks identical to "not configured" — so if you set a
+    // TTL and see no `external_moderation_cache_total`, check this variable before concluding the
+    // metric is broken.
+    //
+    // ⚠️ An earlier revision of this block said the TTL was "THE ARMING SWITCH" with "deliberately
+    // no separate boolean". That was retracted across the module, the counter help text, the redis
+    // key registry and the PR body — and survived HERE, five lines above the comment contradicting
+    // it, which is the surface an operator actually reads. Stated once, in full, at both fields.
+    //
+    // The deployment this cache writes under. Several civitai-web deployments share one sysRedis
+    // and sys keys carry no environment segment, and the PR-preview task copies civitai-cfg
+    // WHOLESALE, so the TTL below is inherited by every open preview. A closed allowlist lives in
+    // moderation-verdict-cache.ts (one rule, one place); anything outside it is OFF and logged
+    // once. See that module for why the policy digest cannot substitute for this.
     EXTERNAL_MODERATION_CACHE_NAMESPACE: z.string().trim().optional().default(''),
+    // Seconds to hold a cached external-moderation verdict — the second of the two required inputs
+    // described above. Capped at 3600 because a cached verdict is a STALE verdict and the TTL is
+    // the only bound on a classifier whose model can change behind a stable name; the measured
+    // value of a longer window is small anyway (12x the TTL bought ~7 points of hit rate).
+    // See src/server/integrations/moderation-verdict-cache.ts.
+    //
+    // ⚠️ This descriptive comment was ORPHANED for two commits — it sat above the NAMESPACE
+    // declaration, so a reader of that field was told it is measured in seconds and capped at 3600.
+    // Introduced by inserting the namespace field between this text and the field it describes.
+    //
     // 🔴 `.catch(0)`, NOT `.default(0)` — the same rule TRPC_MAX_BATCH_SIZE and
     // EXTERNAL_MODERATION_TIMEOUT_MS carry above. `src/env/server.ts` THROWS on any invalid field,
     // and env is parsed only at container start, so a typo here (`=off`, `=false`, `=3600s`,
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -651,7 +651,9 @@ export const serverSchema = z
     // 🔴 THE VERDICT CACHE NEEDS BOTH OF THE NEXT TWO VARIABLES. Neither alone arms it: this one
     // says WHERE entries live, the TTL below says HOW LONG. Setting only one leaves the cache inert
     // and emitting no metric series, which looks identical to "not configured" — so if you set a
-    // TTL and see no `external_moderation_cache_total`, check this variable before concluding the
+    // TTL and see no `civitai_app_external_moderation_cache_total` (PROM_PREFIX is prepended at
+    // registration, so the bare name in the counter's help text is NOT queryable), check this
+    // variable before concluding the
     // metric is broken.
     //
     // ⚠️ An earlier revision of this block said the TTL was "THE ARMING SWITCH" with "deliberately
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Correction to this PR's description — the mutation table overstates one row

The body says the isolated policy mutation "fails exactly **2** tests." That is accurate for the mutation I actually ran (I replaced the `policyDigest(...)` **call site** with a non-hex string, so the three `policyDigest` unit tests — which call the function directly — were untouched, and the key-shape regex `[0-9a-f]{16}` rejected the non-hex value).

It is **not** accurate for the mutation the sentence describes. Mutating `policyDigest`'s **body** to return a 16-hex-char constant fails **4** tests — the three `changes when the THRESHOLD / MODEL / CATEGORY MAP changes` cases plus the invalidation test — and the key-shape test **passes**, because a hex constant still satisfies the regex.

Correcting publicly rather than silently editing the body, since a reviewer may already have read the wrong version. The direction of the error is harmless (it understates the kill), but the row describes a mutation I did not run.

---

## Round 1 audit: 5 findings, all fixed in `394463e63e`

3 🔴 / 2 🟡. Every finding was verified against the tree before fixing rather than taken on the auditor's word — all five reproduced.

| | finding | fix |
|---|---|---|
| F1 🔴 | the awaited `sysRedis.get` had no deadline; the sys client has no socketTimeout, so a silent half-open parks the submit path ~11 min | wrapped in `withSysReadDeadline` |
| F2 🔴 | the key carried no deployment segment; deployments share one sysRedis and the preview task copies `civitai-cfg` wholesale | added `EXTERNAL_MODERATION_CACHE_NAMESPACE` + closed allowlist, in the key |
| F3 🔴 | "a kill switch that does not need a rollout" was false — `env` is a module-load snapshot | claim deleted, truth stated |
| F4 🟡 | `.default(0)` CrashLoops the fleet on an operator typo; this schema mandates `.catch()` in 🔴 three times | `.catch(0)` |
| F5 🟡 | half the strict deserialization guard had no test — the mutant survived 61/61 green | test added |

**F2 changes the operator model:** arming now takes two inputs, `EXTERNAL_MODERATION_CACHE_NAMESPACE=prod` **and** a positive `EXTERNAL_MODERATION_CACHE_TTL_SECONDS`. The original design argued against a second input; that argument was wrong here. These are not two on/off switches — one names *where* entries live, one says *how long*, neither is inferable from the other, and both are necessary by construction. The dark probe already established the rule for a write-only instrument; a cache that serves verdicts on a moderation gate inherits it a fortiori.

**Verification:** 68/68 (was 61). Typecheck OK, 0 errors in 115 s. Prettier clean, instrument-validated. Every new guard mutation-tested and killed by its own test — including the deadline wrapper, whose mutant hangs its test to a vitest timeout rather than passing.

**Two instrument defects found while doing this, both mine:** the F1 test's first draft used `mockImplementation` on a shared-module mock, which `vi.restoreAllMocks()` does not reset, so it leaked into every later case and broke the invalidation test three describes down (scoped to `mockImplementationOnce`); and the mutation runner crashed on a bad `or [None]` — loudly, which is the right direction.

**Still not covered, stated rather than implied:** the live shared-sysRedis topology and the preview ConfigMap key list are quoted from in-repo comments, not re-measured against the cluster. No real-traffic hit rate or sysRedis capacity figure at a 3600 s TTL has been measured.

```audit-claims round=1 audited=2ac6e6e5c6..394463e6
1. F1: the awaited sysRedis read in readCachedVerdict is now wrapped in withSysReadDeadline, and the module comment claiming the worst case is today's latency is corrected.
2. F2: the verdict key now carries a deployment namespace segment gated by a closed allowlist (CACHE_NAMESPACES / EXTERNAL_MODERATION_CACHE_NAMESPACE); the cache arms only when a known namespace AND a positive TTL are both present.
3. F3: the "read per call, so it is a kill switch that does not need a rollout" claim is deleted and replaced with a statement that env is a module-load snapshot and disabling requires a rollout.
4. F4: EXTERNAL_MODERATION_CACHE_TTL_SECONDS uses .catch(0) instead of .default(0), and the test now pins degrade-to-OFF instead of pinning the throw.
5. F5: a test now feeds a non-string category element and asserts result=error plus a re-call, covering the half of the strict deserialization guard that previously had none.
6. New tests added for the namespace allowlist (including on/off spellings), the deadline wrapper under a never-settling get, and two-namespace key separation; all four new guards mutation-tested and killed by their own test.
```

ZacxDev:
## Round 2 audit: no 🔴 — the round-1 fixes are real. 2 🟡 / 3 🟢, all fixed in `eb5ce7753c`

Round 2 confirmed all three round-1 deploy-blockers genuinely closed, re-derived every load-bearing fact, and reproduced all four self-reported mutation kills. What it found was one real gap and a sweep of stale prose.

### 🟡-1 — the write-side arming guard survived mutation, and my "every new guard mutation-tested" claim was overstated

F2 added **two** guards; I mutated only the read one. Removing the namespace check from the *write* path left **28/28 green**: a deployment with a rejected namespace stopped being inert and wrote `generation:moderation-verdict:null:<policy>:<digest>` into one shared un-namespaced keyspace.

🔴 **The obvious fix was also vacuous, and only re-running the mutant caught it.** Adding `expect(sysRedis.set).not.toHaveBeenCalled()` passes whether or not the guard exists — `writeCachedVerdict` is fire-and-forget behind a lazy `import()`, so the write hasn't happened *yet* either way. **An absence you did not wait for is not an absence.** The test now orders it (a third armed call whose write is awaited, then the whole keyspace asserted), and the mutant dies alone with `expected 'generation:moderation-verdict:null:2c…' to contain ':prod:'`.

**Root cause of that gap:** the round-1 fix open-coded the arming predicate at both live sites and left `moderationCacheEnabled()` with **zero production callers** — so seven test assertions aimed at a function nothing executed, while the two predicates actually gating the cache were unasserted. Consolidated into `armedCache()`, which returns the namespace and TTL; returning the *values* is what makes one predicate sufficient, and needing them is why the sites open-coded it.

### 🟡-2 — five sibling claims still asserted the retracted version

Two are operator-facing:

1. **The Prometheus counter HELP text** (ships to Grafana) said "unarmed (no `EXTERNAL_MODERATION_CACHE_TTL_SECONDS`) there are no series" — conflating *no TTL configured* with *TTL set, namespace rejected*. That is the exact two-states-confused failure the arming design exists to prevent.
2. **The redis key-registry docblock**, three lines below a sentence this ladder had already edited, still called the TTL the arming switch.

Plus: `ttlSeconds()`'s docstring repeated the "no second on/off input" argument that `cacheNamespace` retracts **35 lines above it in the same file**; it claimed a non-numeric value is "rejected by the env schema, not silently coerced", which round 1's own `.catch(0)` had just made false; and `buildVerdictKey`'s docstring declared a 3-segment key while emitting four.

🟢-4 renamed three tests whose names asserted pre-fix behaviour (including one whose own body now asserts the opposite). 🟢-5 carried the "before adding a member, check both" invariant onto the duplicated allowlist — the probe holds the only other copy and is scheduled for deletion, so the rule was about to be lost off the higher-stakes surface. 🟢-3 corrected in the PR body above.

### Two claims in the round-1 commit message were false

They can't be amended after a push, so they're retracted in `eb5ce7753c`'s message and here:

- **"the only unwrapped awaited sys read in the tree"** — false; ~50 remain (`system-cache.ts`, `block-tip-rate-limit.ts`, `blocks.router.ts`, …). None is on the generation submit path, so the *fix* is right and the *claim* was not. The code comment makes the narrower, defensible version.
- **"this schema documents that rule three times in 🔴"** — three `.catch()` rationales exist, but only one carried a 🔴 before that commit.

### And I repeated the pattern I was fixing

While correcting the PR body I swept some claims and left three siblings stale in the same edit — the body's opening line, the `61/61`, and the `147 s`. Fixed in a follow-up edit and swept for others.

**Verification:** 68/68. Typecheck OK, 0 errors in 103 s. Prettier clean. All six mutants re-confirmed killed, each by its own test.

**Still not verified, stated rather than implied:** the cluster-topology facts the F2 design rests on (shared sysRedis, the preview `civitai-cfg` wholesale copy, the ~10-previews figure) are quoted from in-repo comments and live in `talos-infra`; no real-traffic hit-rate or capacity number at a 3600 s TTL has been measured.

```audit-claims round=2 audited=394463e63e..eb5ce775
1. 🟡-1a: the namespace test now asserts the WRITE side too, ordered behind a third armed call whose write is awaited, and the write-side arming mutant now fails alone instead of surviving 28/28.
2. 🟡-1b: the arming predicate is consolidated into armedCache(), used by both production call sites; moderationCacheEnabled() is a derived view of it rather than a third copy.
3. 🟡-2: five stale sibling claims corrected — the Prometheus HELP text, the redis key-registry docblock, ttlSeconds()'s retracted "arming switch" paragraph, its "rejected by the env schema" sentence, and buildVerdictKey's 3-segment key shape.
4. 🟢-4: three test names/comments that asserted pre-fix behaviour renamed; console.error stubbed in the on/off-spellings case.
5. 🟢-5: the "before adding a member, check both" invariant plus the frozen-array and one-object rationales carried onto CACHE_NAMESPACES from the probe.
6. 🟢-3: the PR body's key shape, arming paragraph, mutation table, test count and typecheck timing corrected; the retracted claims are quoted in a retraction note rather than deleted.
7. Two false claims in the round-1 commit message (the "only unwrapped sys read in the tree", and "three times in 🔴") are retracted in the round-2 commit message, since a pushed message cannot be amended.
```

ZacxDev:
## Round 3 audit: no 🔴 again — 2 🟡 / 3 🟢, all prose, all fixed in `44efcd488c`

Round 3 independently reproduced the round-2 regression matrix: the write-side arming mutant is **red at HEAD and green at `394463e63e`**, confirming that test is a genuine regression test and not an invariant guard. The payload has been clean since round 2.

### 🟡-1 — the retracted arming claim survived **verbatim** in the env schema

`src/env/server-schema.ts` still said *"THIS IS THE ARMING SWITCH … there is deliberately no separate boolean"* — the exact sentence rounds 1 and 2 retracted in the module, the counter HELP text, the redis key registry and the PR body — sitting **five lines above the comment that contradicts it**, on the one surface an operator reads to learn what these variables do.

Failure: read it, set only the TTL, see no `external_moderation_cache_total`, conclude the metric is broken. That is the two-states-confused outcome the F2 design exists to prevent, reached from the authoritative config surface.

🔴 **Third recurrence of the same shape on this PR, so this round closed the SHAPE rather than the instance.** Every site stating the cache's arming condition was enumerated by grep and corrected; the sweep now returns none. The two the audit hadn't named were `moderation.ts`'s production call-site comment and the module header's own enumeration.

**I also orphaned a comment block two commits ago**, which round 3 caught: the TTL's descriptive text ("Seconds to hold…", "Capped at 3600") sat above the **namespace** declaration, because I inserted that field between the text and the field it describes. A reader of the namespace variable was told it's measured in seconds. Re-attached.

### 🟡-2 — the carried-over invariant said "CHECK BOTH" and stated one check

The probe's rule is (1) the member names ONE running population **and** (2) something actually **scrapes** it. Only (1) came across. A heading promising two checks and supplying one is precisely what stops the next person looking for the missing half — and (2) is the criterion that removed `next-stage` from the probe's list. For the write-only probe an unscraped namespace meant an unreadable measurement; for this cache it means **serving cached verdicts on a moderation gate with zero observability**.

Same gap in the HELP text this ladder rewrote: it listed two causes of an absent series and omitted the third, *armed but unscraped*. Its discriminator also over-claimed — "told apart only by the one-off error the namespace resolver logs" is unsound, since the namespace is resolved whether or not a TTL is set, so that log separates a rejected namespace from the other two causes, not the two named states from each other.

### 🟢-3 — two mutation-table rows had drifted, caused by the commit that claimed to fix that table

Re-measured **every** row directly rather than taking the audit's word: the namespace-segment row is **4** kills not 3 (the new keyspace assertion adds one), and the TTL-arming row is **2** not "alone" (`moderationCacheEnabled()` now routes through `armedCache()`). Both understate, so harmless in direction. Body corrected, with a note that every row was re-measured at the current head.

### 🟢-5 — the round-1 correction comment now quotes body text that no longer exists

[That comment](https://github.com/civitai/civitai/pull/4658#issuecomment-5561747650) opens by quoting *"fails exactly 2 tests"* from the body; the round-2 body edit removed that phrase. Its stated purpose was "correcting publicly rather than silently editing the body" — and the body was then edited anyway. Recording it here rather than editing that comment: it is an accurate record of what the body said **when it was written**, and rewriting it would destroy the very history it exists to preserve.

**Verification:** 68/68. Typecheck OK, 0 errors in 102 s. Prettier clean. The two drifted counts re-measured directly.

**Still not verified:** every figure sourced from the dark probe (34.3 %, 224,989, ~205 ms, 78–88 %) and every cluster-topology fact behind F2 — those live in `talos-infra` and are out of this repo's reach.

```audit-claims round=3 audited=eb5ce7753c..44efcd48
1. 🟡-1: the retracted "THIS IS THE ARMING SWITCH / no separate boolean" text is removed from src/env/server-schema.ts and replaced with a statement that both variables are required.
2. 🟡-1b: every remaining site stating the cache's arming condition was enumerated and corrected — moderation.ts's call-site comment and the module header's staleness enumeration; a repo-wide grep for the shape now returns no hits.
3. 🟡-1c: the orphaned comment block in server-schema.ts is re-attached to the TTL field it describes, so the namespace field is no longer documented as seconds-and-capped-at-3600.
4. 🟡-2: the CACHE_NAMESPACES invariant now states BOTH checks, the second being that something scrapes the deployment; the counter HELP text gains the armed-but-unscraped cause and its discriminator is narrowed to what the log actually separates.
5. 🟢-3: every mutation-table row in the PR body was re-measured at the current head; the namespace-segment row is corrected to 4 and the TTL-arming row to 2, with a note that drift was caused by the commit claiming to correct the table.
6. 🟢-5: the round-1 correction comment is left unedited by choice, and the reason is recorded in the round-3 comment rather than by rewriting it.
```

ZacxDev:
## Round 4: no 🔴, no 🟡 — 3 🟢, all fixed in `632828c3a7`. **The ladder stops here.**

Round 4 independently re-derived **all nine mutation-table rows from scratch** and every one matched, so that table is now corroborated by someone who didn't write it. It also confirmed the round-3 sweep, the re-attached comment block, and the retraction notes.

### What it found

**🟢-1 — the mirrored arming claim.** `cacheNamespace()`'s summary said *"or `null` when it is not armed"* — an equivalence true in one direction only: with `NAMESPACE=prod` and `TTL=0` it returns `'prod'` while the cache is off. This is the **mirror** of the TTL-only claim retracted three times, written for the *other half*, and it survived every sweep because every sweep looked for the TTL wording.

**🟢-2 — a sentence added last round named a series that does not exist.** It told an operator *"set a TTL and see no `external_moderation_cache_total`"* — but `PROM_PREFIX` (`civitai_app_`) is prepended at registration, so the queryable name is `civitai_app_external_moderation_cache_total`. Someone following that instruction queries a name returning nothing and concludes the cache is broken — the exact outcome the sentence was added to prevent. (The bare form is correct *inside* the counter's help text, where it's the registration argument.)

**🟢-3 — the narrowed discriminator was ambiguous.** *"A rejected namespace logs one error line, which separates it from the other two"* is false when a deployment has both a typo'd namespace and no TTL — that also logs. Rather than reverse the comparative, I deleted it: the help text now states what the log actually does and claims nothing about which states it separates.

**This time the sweep was by SHAPE, not by instance** — two greps, one for every comment asserting an arming *state* and one for every metric name in prose an operator could copy. Shape A returned 5 hits (4 correct, 1 wrong); Shape B returned 3 (2 legitimate, 1 wrong). Both re-run clean.

---

## Why this is the last round

**The payload has not changed since `eb5ce7753c`.** Round 4's diff was 44 lines with **zero executable**; this one is the same shape.

Rounds 3 and 4 found only prose, and **two of round 4's three findings were prose about the prose corrections** — 🟢-1 is the sibling of a comment fixed in round 2, 🟢-3 critiques a sentence written in round 3 to critique a sentence written in round 2. A round 5 would, by construction, audit round 4's edits to round 3's edits.

Neither automatic stop rule can fire here, and that is *why* this needs stating explicitly rather than being left implicit:

- the **findings-keyed** rule can't fire because the payload is prose, so "fixed a defect" and "reworded a warning" are the same edit;
- the **attribution gate** can't fire for the same reason — comment lines in shipped source count as payload, so no round reads as zero.

So I'm stopping on the stated criterion: **no 🔴, no blast radius beyond "the document contains a false sentence", and the recurring shape swept at every site rather than at the one reported.**

**Nothing is deliberately left unfixed.** All three of round 4's findings are fixed in `632828c3a7`. Recording that explicitly, because a report that *ended* a ladder on this criterion is otherwise indistinguishable from one that converged.

---

## Ladder summary

| round | findings | what changed |
|---|---|---|
| 1 | **3 🔴**, 2 🟡 | unbounded sys read on the submit path; no deployment segment in the key; a false kill-switch claim; `.default()` vs `.catch()`; an untested guard half |
| 2 | 2 🟡, 3 🟢 | the write-side arming guard survived mutation; five stale sibling claims |
| 3 | 2 🟡, 3 🟢 | the retracted claim verbatim in the env schema; an orphaned comment block; "CHECK BOTH" listing one check |
| 4 | **3 🟢** | the mirrored arming claim; an unqueryable metric name; an ambiguous discriminator |

**Round 1 is the only round that found a code defect that could hurt in production**; round 2 found the only other real one (an unpinned guard). Everything after was documentation.

**Final state:** 68/68 across four moderation suites, typecheck 0 errors, prettier clean, nine mutants each killed by their own test and all nine independently re-derived by round 4.

**Still unverified, and out of this repo's reach:** every figure from the dark probe (34.3 %, 224,989 observations, ~205 ms, 78–88 %) and every cluster-topology fact behind the namespace design (shared sysRedis, the preview `civitai-cfg` wholesale copy, which deployments are scraped). Those live in the private infrastructure repo. Anyone arming this cache should confirm them there first.

ZacxDev:
## Correction: the "final state" line above overstated what had been verified

The round-4 comment closed with *"68/68 across four moderation suites, typecheck 0 errors, prettier clean"* and presented that as the gate. **It was three tiers, not four.** CI then went red on the fourth:

```
moderation-verdict-cache.test.ts:336:71: Unexpected empty arrow function
[@typescript-eslint/no-empty-function]
```

`new Promise(() => {})` in the deadline test. Fixed in `88f3adae7a` as `new Promise(() => undefined)` — identical behaviour, non-empty body.

**ESLint was never run locally, once, across five commits.** Every round reported the same three commands and called it complete, and the repo's CI lints **added** files specifically — which is exactly what this PR adds — so the one tier that was skipped is the one that had a finding. Four adversarial audit rounds didn't catch it either, because each was briefed with those same three commands.

That is the same shape as most of what this ladder found in the prose: a completeness claim that hadn't been earned. It is worth recording against a PR whose whole review history is about exactly that.

Now run properly, with the instrument validated first — appending `const __ctl = () => {}` produced `1 problem / no-empty-function`, so the clean result is a real zero rather than a silent no-op:

- `pnpm exec eslint --format unix` over all five changed files → **rc 0**
- prettier → clean · vitest → **68/68** · typecheck → 0 errors

**CI on `88f3adae7a`: 18 pass, 1 skipping, 0 fail.**

This does not reopen the audit ladder — it was a one-line mechanical fix in a test, not a finding — but the "final state" summary above should be read with this correction attached.

PR #4659, opened 2026-09-06
feat(apps): project openCount onto the store CARD DTO — number on-site, null off-site

ZacxDev:
Stage 3 of the app-store play-count arc: project `openCount` onto the store **card** DTO so a `/apps` grid tile can say how used an app is.

## What it does

| File | Change |
|---|---|
| `src/server/schema/blocks/app-listing-read.schema.ts` | `ListingCard` gains `openCount: number \| null`, with the allowlist justification every field on this public DTO carries. |
| `src/server/services/blocks/app-listing.service.ts` | `listingHydrateSelect` selects `metric.openCount`; `projectListingCard` projects it through a new `cardOpenCount` helper. |
| `src/server/services/blocks/__tests__/app-listing.service.test.ts` | 13 new cases + `openCount` added to the exact-key-set allowlist assertion. |

**Scope is the CARD projection only.** The detail (`:516`) and `listMine` (`:1025`) reads are untouched — they keep projecting `installCount` and nothing else.

## 🔴 The null-vs-zero rule, and why it is a truth claim

```
on-site  → a NUMBER.  row.metric?.openCount ?? 0
off-site → null, ALWAYS, whatever the column holds
```

An **off-site** listing's CTA is a plain `target="_blank"` anchor to a third party. No on-platform request follows the click, so there is **nothing trustworthy to count** — its play count is *structurally absent*, not zero. Stage 4 renders `null` by omitting the stat row entirely; a `0` would render as **"nobody has ever used this app"**, a false statement about an app we simply cannot measure.

`app_listing_metrics.open_count` is `Int NOT NULL DEFAULT 0`, so an off-site row genuinely **does** carry a literal `0` in the column. Discriminating on `kind` is the only thing that keeps that value off the wire; `row.metric?.openCount ?? 0` alone would silently emit the wrong answer for every off-site card.

**The mirror hazard is equally real: do NOT over-null.** An on-site listing nobody has opened yet is a genuine **`0`** and must stay `0` — a missing metric row means "no plays recorded yet", which is 0 (the same COALESCE-to-0 reading `installCount` documents).

**The discriminator is `kind`, never `appBlockId` nullness.** `schema.prisma` says so at the `appBlockId` field: a natively-created off-site listing also leaves it NULL, while the `#2821` off-site rows *do* carry one. The two predicates disagree on real rows **in both directions**, so there are two tests for it rather than one.

The positive `=== 'onsite'` test (rather than `!== 'offsite'`) is deliberate: it fails **closed** to `null` for any kind added later, because an omitted stat row is honest about an unmeasured app while a `0` is a false claim about it.

## Type-forced widening (declared, not silent)

`openCount` is a required field on `ListingCard`, which forces two things beyond the card projection. Both are called out because they are the only places this PR touches that the brief did not name:

- **`src/components/Apps/reviewListingPreview.ts`** — `buildListingCardPreview` returns a `ListingCard`, so it must supply the field. It gets **`null`** — but 🔴 **not for the reason this line originally gave.** It said *"a listing still in the review queue has not been openable by anyone"*, which is FALSE: an `onsite` review row is a listing-MEDIA revision of an already-approved, already-LIVE app, which may have any number of real opens. The correct reason is narrower and covers BOTH kinds: this builder never reads an `AppListingMetric`, so it cannot measure the row at all, and the DTO's definitions put a producer that cannot measure in the `null` case. Round 1 changed this to a kind-discriminated `0`; round 2 reverted it to `null` by operator decision — truth over shape-parity with `cardOpenCount`. See the round-2 comment. Same own-the-limitation shape as the `isBeta: false` / `sourceRepoUrl: null` placeholders already in that file.
- **11 browser-tier fixture files** under `src/components/Apps/` — one line each (`openCount: 0` for fixed-on-site fixtures, `openCount: kind === 'onsite' ? 0 : null` for the three that parameterise kind). Pure fixture compliance; no assertion changed.

## Fixtures cannot pass by accident

Every off-site fixture carries a **non-zero** `open_count` (8123, 4517, 9902, 3311, 7604), and the on-site counts are pairwise distinct (4213, 6178, 5290, 2748). A `0`-seeded off-site fixture could not tell the correct projection apart from the naive `?? 0` — both would return the same thing and the suite would be green over a broken projection. None of the values equals a constant any assertion names.

## Mutation matrix

Every mutant applied to the **final** tree (post-prettier), each run individually, tree restored and checksum-verified byte-identical afterwards. Baseline: **99 passed** at the time of that run; the suite is **100** as of `689e82278b`.

| # | Mutant | Result | Killed by (test → its own assertion message) |
|---|---|---|---|
| M1 | `cardOpenCount` → `return row.metric?.openCount ?? 0;` (no `kind` discrimination) | **6 failed / 93 passed** (7 as of `689e82278b` — the fail-closed test added in round 1 also kills M1) | `🔴 OFF-SITE whose metric row holds a NON-ZERO count: null` → `AssertionError: expected 8123 to be null`; also the 4-shape sweep (`expected 4517 to be null`), the no-metric case (`expected +0 to be null`), the kind-pair case (`expected 5290 to be null`), the appBlockId case and the JSON-wire case |
| M2 | return `null` whenever the count is 0 | **3 failed / 96 passed** | `ON-SITE whose metric row holds a literal 0: 0, not null (do NOT over-null)` → `AssertionError: expected null to be +0 // Object.is equality`; also the no-metric-row and omitted-column on-site cases, same message |
| M3 | `if (!row.appBlockId) return null;` (discriminate on `appBlockId` nullness) | **4 failed / 95 passed** | `🔴 ON-SITE with a NULL appBlockId is still a NUMBER` → `expected null to be 6178`; `🔴 OFF-SITE WITH a backing appBlockId is still NULL` → `expected 8123 to be null`; plus the 4-shape sweep (`{"appBlockId":"ab_7",…}: expected 3311 to be null`) and the kind-pair case |
| M4 | drop `openCount: true` from `listingHydrateSelect` | **1 failed / 98 passed** | `the shared hydrate select actually asks for the column` → `expected undefined to be true` |
| M5 | narrow the DTO to `openCount: number;` | **1 failed / 98 passed** (vitest) **+ typecheck rc=2, 5 errors** | vitest: `🔴 the DTO declares openCount: number \| null (source tier)` → `expected '\n  id: string;…' to match /^\s*openCount: number \| null;$/m`. typecheck: `app-listing.service.ts(354,5): TS2322 Type 'number \| null' is not assignable to type 'number'`, `reviewListingPreview.ts(124,5): TS2322 Type 'null' is not assignable to type 'number'`, + 3 fixture files |
| M6 | delete the `openCount:` line from the card projection | **12 failed / 87 passed** | `projects the exact public card allowlist` → `expected [ …(12) ] to deeply equal [ …(13) ]`, plus all 11 openCount cases (`expected undefined to be …`) |

Notes on the battery:

- **M1 and M3 are separated, not conflated.** `🔴 ON-SITE with a NULL appBlockId is still a NUMBER` correctly **survives** M1 (M1 returns the number for on-site too) and kills M3 — so that test is genuinely pinning the `appBlockId` half rather than riding along on the `kind` half.
- **The type guard needed two tiers, and the obvious one was inert.** `tsconfig.json` excludes `src/**/__tests__/**` from the root program, so a `const x: ListingCard['openCount'] = null;` written inside the service test file would compile-error **nowhere**. Measured directly: `tsc --listFilesOnly` returns **0** hits for `app-listing.service.test.ts` against **1** each for `app-listing.service.ts` and a sibling `.browser.test.tsx` (positive controls) out of 12,479 files. Hence the source-parse test — which carries its own positive control (`reviewCount: number;` must be found in the parsed body, so a regex that matched nothing cannot report "no violation" for a file it never read). The typecheck tier is covered by the production call sites, as M5 shows.

## Gates

All run in the worktree via `direnv exec …/civitai env -C <wt> …`.

| Gate | Command | Result |
|---|---|---|
| Unit (target suite) | `npx vitest run --project unit src/server/services/blocks/__tests__/app-listing.service.test.ts` | **rc=0 — 100 passed (100)** as of `689e82278b`, 1 file |
| Unit (related suites) | same + `reviewListingPreview` / `appListingGrid` / `appListingCardView` / `related-listings` / `recentAppsRail` / `appListingStatChips` | **rc=0 — 7 files, 312 passed (312)** |
| Unit (whole project) | `npx vitest run --project unit` | 1636 passed / 1 failed / 3 skipped files; **38,111 passed, 1 failed, 33 skipped**. The single failure is `src/server/__tests__/eventloop-watchdog.capture.test.ts`, a **flake** — see below |
| Typecheck | `pnpm typecheck` | **rc=0, 0 errors** (`grep -ac 'error TS'` = 0) |
| ESLint | `npx eslint --format unix <16 changed files>` | **0 errors on the changed lines**; 4 pre-existing `no-unused-vars` *warnings* on `_a` at lines 25–29 of the service test (the `vi.hoisted` mock block, untouched by this PR) |
| Prettier | `npx prettier --check <16 changed files>` | **15 clean, 1 pre-existing non-blocking warning** — corrected in round 1, see below |

**Both linters were positive-controlled in the same invocation** against a deliberately-bad throwaway file: prettier flagged `[warn] src/__lintprobe__/probe.ts`; eslint reported 4 findings on the probe (2 Error, 2 Warning) alongside the real files. The probe was deleted before staging.

> #### ⚠️ Correction (round-1 audit): the Prettier row above originally read "clean", and that was wrong for 1 of 15 files.
>
> `npx prettier --check` reports `[warn] src/server/schema/blocks/app-listing-read.schema.ts`. The offending hunk is at `:104-108` (`ListAllListingsForModerationInput`) — **outside this PR's diff and byte-identical on `origin/main`** (this PR's only hunk in that file is at `+211,30`). Verified by extracting the base blob and diffing Prettier's own output: exactly **one** hunk, and it is that one.
>
> **Deliberately not reformatted** — that would be scope creep. CI is correctly green: `lint.yml` runs Prettier `continue-on-error: true` for **modified** files and blocking only for **added** files, and this PR adds none.

### The one red, and why it is not this PR

`eventloop-watchdog.capture.test.ts` failed once in the full run (`🔴 classifies a wedge as starved even while OTHER threads burn CPU`, **12,129 ms**) and once in isolation — but on a **different test** (`🔴 classifies a CPU-burning wedge as executing`, **11,949 ms**). Two further isolated runs: **9 passed (9)** both times. A different test failing each time, at ~12 s wall on a timing classifier, is the flake signature and not an assertion failure. The file imports only `~/server/eventloop-watchdog` and contains **0** occurrences of `listing`/`Listing` — there is no path from this change to it.

## 🔴 What I could NOT verify

**The browser test tiers did not run on this machine.** The `component` (`*.browser.test.tsx`) and `geometry` (`*.geometry.test.tsx`) projects both abort at browser launch:

```
browserType.launch: Executable doesn't exist at
  /nix/store/…-playwright-browsers/chromium_headless_shell-1200/chrome-headless-shell-linux64/chrome-headless-shell
```

The store path ships `chromium_headless_shell-**1228**`, not `-1200` — a pinned-browser version mismatch in this environment. **Control run:** an *untouched* geometry test (`src/tests/geometry/geometryHarness.geometry.test.tsx`) fails identically, so this is the environment, not this branch.

Consequence, stated plainly: the 11 `.browser.test.tsx` / `.geometry.test.tsx` fixture files this PR edits are covered here **only by `pnpm typecheck`**, which does include them — they produced 11 of the 12 original TS2741 errors when the field was first added, and they compile clean now. Their assertions were not re-executed. CI's browser tier is the check that closes this.

## Independence from the rest of the arc

This branch is based directly on `origin/main`, **not stacked**. It is genuinely independent of the two open PRs: `app_listing_metrics.open_count` already exists in the schema (`Int NOT NULL DEFAULT 0`), so nothing here needs them.

It simply **reads `0` for every on-site listing until #4653's rollup populates the column, and until #4652's events exist to feed that rollup.** Off-site cards read `null` permanently, by design. Nothing in this PR changes what the column contains — it only decides how the column is projected.

## 🔴 MERGE-ORDER CONSTRAINT — read before shipping Stage 4

Independence from the arc is **not** independence from merge order.

Nothing writes `open_count` yet: `appListing.metrics.sql.ts` populates only `install_count`, and its own suite asserts `expect(upsert).not.toContain('open_count')`. So `cardOpenCount` returns a **literal `0` for every on-site listing in production today**, and the DTO has no third state for "measurable but not yet measured" — `null` means unmeasurable, `0` means measured-as-none, and the honest answer right now is neither.

> **Stage 4 (the renderer) must not merge before #4653, or must gate the rendered stat row behind a flag.**

Shipping the renderer first makes the public `/apps` grid print **"0 plays" on every on-site app, including heavily-used ones** — by this PR's own written standard the worst outcome in the arc, on the public surface. The flag is deliberately **not** built here.

This constraint is also recorded in a comment at `cardOpenCount` so it cannot be missed from the code side.

## Round-1 audit fixes — `0f4ada7c09`

Five findings addressed; full write-up, mutation matrix and `audit-claims` block in the [round-1 comment](https://github.com/civitai/civitai/pull/4659#issuecomment-5561570563). Summary:

1. ~~**`buildListingCardPreview` no longer over-nulls the on-site case** — it now discriminates on kind (`onsite → 0`, `offsite → null`) like `cardOpenCount`.~~ 🔴 **SUPERSEDED at `689e82278b` (round 2): reverted to an unconditional `null` by operator decision**, and re-stated at `a49eaad27f` (round 3) as a deliberate OVERRIDE of the DTO's rule rather than a derivation from it. What survives from round 1: the comment no longer claims a review-queue listing could not have been opened, and the file has tests it previously lacked — it contained zero occurrences of `openCount` before. See the round-2 and round-3 comments.
2. **The "fails CLOSED for any future kind" claim is now guarded** by an unknown-`kind` test. The `=== 'offsite'` mutant survived 99/99 before and is killed by exactly that one test now.
3. ~~**Both `AppListingMetric` module headers** record that `open_count` now has a consumer.~~ 🔴 **REVERTED at `689e82278b` (round 2).** Those notes manufactured a conflict with **#4653** that did not previously exist — measured, one conflict hunk in each file, with contradictory content — so both were removed. **This PR now touches no file in common with #4653** (16 paths vs 8, intersection empty). #4653 owns those headers.
4. **The Prettier gate result is corrected** (see the correction box above).
5. **The merge-order constraint is recorded** (this section, plus the code comment).

Additionally, the `ListingCard.openCount` **allowlist justification was rewritten**. It claimed "the exact same argument as `ListingDetail.installCount`, one step weaker", which flattered it on an axis it never named: `openCount` is a **card** field and so is **bulk-enumerable by an anonymous caller** over `GET /api/v1/apps` at up to 50 cards per page, where `installCount` is detail-only and costs one request per app; and no sort exposes `open_count`, so it is a genuinely new public fact rather than added precision. **Both deltas were raised and accepted deliberately** — printing plays on a public card was already the decision, so bulk readability is a consequence of it, and the `popular` sort already makes relative ordering by usage publicly inferable. Recorded in-comment so the next reviewer gets the accurate comparison.

### DIFF
--- a/src/components/Apps/AppListingCard.browser.test.tsx
+++ b/src/components/Apps/AppListingCard.browser.test.tsx
@@ -232,6 +232,7 @@ function base(over: Partial<ListingCard>): ListingCard {
     creator: { id: 5, username: 'alice', image: null },
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: 'blk-1',
--- a/src/components/Apps/AppListingCardSkeleton.geometry.test.tsx
+++ b/src/components/Apps/AppListingCardSkeleton.geometry.test.tsx
@@ -90,6 +90,7 @@ function makeCard(id: string, name: string, creator: string): ListingCard {
     creator: { id: 7331, username: creator, image: null },
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `blk-${id}`,
--- a/src/components/Apps/AppListingDetailBody.browser.test.tsx
+++ b/src/components/Apps/AppListingDetailBody.browser.test.tsx
@@ -343,6 +343,7 @@ function relatedCard(id: string, name: string): ListingCard {
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `ab-${id}`,
--- a/src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.browser.test.tsx
@@ -28,6 +28,7 @@ function makeCard(id: string, name: string, kind: 'onsite' | 'offsite' = 'onsite
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: kind === 'onsite' ? 0 : null,
     kindData:
       kind === 'onsite'
         ? {
--- a/src/components/Apps/AppListingsMarketplaceBody.columns.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.columns.browser.test.tsx
@@ -64,6 +64,7 @@ function makeCard(id: string, name: string, tagline: string | null = 'tag'): Lis
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `blk-${id}`,
--- a/src/components/Apps/AppListingsMarketplaceBody.hydration.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.hydration.browser.test.tsx
@@ -56,6 +56,7 @@ function makeCard(id: string, name: string, kind: 'onsite' | 'offsite' = 'onsite
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: kind === 'onsite' ? 0 : null,
     kindData:
       kind === 'onsite'
         ? {
--- a/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
@@ -58,6 +58,7 @@ function makeCard(id: string, name: string): ListingCard {
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `blk-${id}`,
--- a/src/components/Apps/AppListingsMarketplaceBody.recentsReconcile.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.recentsReconcile.browser.test.tsx
@@ -66,6 +66,7 @@ function makeOnsiteCard(args: {
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: args.appBlockId === undefined ? `blk_${args.slug}` : args.appBlockId,
--- a/src/components/Apps/AppListingsMarketplaceBody.storeGate.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.storeGate.browser.test.tsx
@@ -44,6 +44,7 @@ function makeCard(id: string, name: string): ListingCard {
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `blk-${id}`,
--- a/src/components/Apps/AppListingsMarketplaceBody.stretch.geometry.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.stretch.geometry.test.tsx
@@ -110,6 +110,7 @@ function makeCard(id: string, name: string, tagline: string | null): ListingCard
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: 0,
     kindData: {
       kind: 'onsite',
       appBlockId: `blk-${id}`,
--- a/src/components/Apps/AppListingsMarketplaceBody.urlState.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.urlState.browser.test.tsx
@@ -50,6 +50,7 @@ function makeCard(id: string, name: string, kind: 'onsite' | 'offsite' = 'onsite
     creator: null,
     recommend: { recommendedCount: 0, notRecommendedCount: 0, recommendPct: null },
     reviewCount: 0,
+    openCount: kind === 'onsite' ? 0 : null,
     kindData:
       kind === 'onsite'
         ? {
--- a/src/components/Apps/__tests__/reviewListingPreview.test.ts
+++ b/src/components/Apps/__tests__/reviewListingPreview.test.ts
@@ -124,6 +124,39 @@ describe('buildListingCardPreview', () => {
     );
   });
 
+  /**
+   * 🔴 THE PLAY COUNT IS KIND-DISCRIMINATED HERE TOO, and this builder is the SECOND
+   * producer of the field on the SAME screen: `ListingPreviewSection` renders
+   * `previewQuery.data?.card ?? buildListingCardPreview(request)`, and the primary side
+   * (`getListingPreviewForReview` → `projectListingCard` → `cardOpenCount`) returns a
+   * NUMBER for an on-site row. An unconditional `null` here would make the two
+   * producers disagree about which kind is measurable.
+   *
+   * The DTO's own rule (`app-listing-read.schema.ts`, `ListingCard.openCount`) is what
+   * settles it: off-site is ABSENT (`null`), on-site with nothing recorded is a genuine
+   * `0`. And an `onsite` review row is a listing-MEDIA revision of an already-approved,
+   * already-live app — so "nobody could have opened it" is not true of it.
+   *
+   * 🔴 BOTH kinds are asserted, so NO SINGLE CONSTANT passes: a hardcoded `null` reds
+   * the on-site case, a hardcoded `0` reds the off-site case.
+   */
+  it('🔴 openCount: 0 on an on-site row, null off-site — no single constant satisfies both', () => {
+    const onsite = buildListingCardPreview(row({ id: 'r6', kind: 'onsite' }));
+    expect(onsite.kind).toBe('onsite');
+    expect(onsite.openCount).toBe(0);
+    expect(onsite.openCount).not.toBeNull();
+
+    const offsite = buildListingCardPreview(row({ id: 'r6b', kind: 'offsite' }));
+    expect(offsite.kind).toBe('offsite');
+    expect(offsite.openCount).toBeNull();
+
+    // A row carrying no `kind` at all defaults to off-site (the review adapters'
+    // backward-compatible default) — so it takes the off-site answer, not the on-site one.
+    const defaulted = buildListingCardPreview(row({ id: 'r6c' }));
+    expect(defaulted.kind).toBe('offsite');
+    expect(defaulted.openCount).toBeNull();
+  });
+
   it('passes resolved image URLs straight through when provided', () => {
     const card = buildListingCardPreview(row({ id: 'r5' }), {
       iconUrl: 'https://cdn/icon.png',
--- a/src/components/Apps/__tests__/reviewListingPreview.test.ts
+++ b/src/components/Apps/__tests__/reviewListingPreview.test.ts
@@ -128,11 +128,18 @@ describe('buildListingCardPreview', () => {
    * 🔴 THE PLAY COUNT IS `null` HERE FOR EVERY KIND, and the reason is what this test
    * exists to pin — because an earlier round asserted the OPPOSITE and was wrong.
    *
-   * This builder never reads an `AppListingMetric`, so it cannot measure the row at
-   * all, for any kind. The DTO's own definitions (`app-listing-read.schema.ts`,
-   * `ListingCard.openCount`) then decide it: `null` = "not measurable on this surface",
-   * `0` = "measured, and the answer is none". A producer that cannot measure is in the
-   * `null` case.
+   * 🔴 THIS OVERRIDES THE DTO'S RULE RATHER THAN FOLLOWING IT. The DTO says an on-site
+   * listing with no metric row is a genuine `0` and must not be over-nulled, and that is
+   * exactly this builder's on-site input — so the written rule yields `0` here. We return
+   * `null` anyway, on a judgement about this surface: `OffsitePendingRow` carries no
+   * `AppListingMetric` and this builder never reads one, so unlike `cardOpenCount` it can
+   * never become right later, and a `0` is wrong permanently on a moderator screen for an
+   * app that may have 40,000 plays. Operator call, 2026-09-06: truth over parity.
+   *
+   * 🔴 Do NOT restate that as "a producer that cannot measure returns null". An earlier
+   * draft did, and it is wider than the decision: `cardOpenCount` cannot measure today
+   * either, so that phrasing licenses nulling on-site rows on the PUBLIC card DTO — the
+   * defect this PR's own mutant M2 exists to catch.
    *
    * 🔴 IT DELIBERATELY DISAGREES WITH `cardOpenCount`, which returns a NUMBER for an
    * on-site row, and that divergence is the point. Making the two agree on shape (the
--- a/src/components/Apps/__tests__/reviewListingPreview.test.ts
+++ b/src/components/Apps/__tests__/reviewListingPreview.test.ts
@@ -125,33 +125,37 @@ describe('buildListingCardPreview', () => {
   });
 
   /**
-   * 🔴 THE PLAY COUNT IS KIND-DISCRIMINATED HERE TOO, and this builder is the SECOND
-   * producer of the field on the SAME screen: `ListingPreviewSection` renders
-   * `previewQuery.data?.card ?? buildListingCardPreview(request)`, and the primary side
-   * (`getListingPreviewForReview` → `projectListingCard` → `cardOpenCount`) returns a
-   * NUMBER for an on-site row. An unconditional `null` here would make the two
-   * producers disagree about which kind is measurable.
+   * 🔴 THE PLAY COUNT IS `null` HERE FOR EVERY KIND, and the reason is what this test
+   * exists to pin — because an earlier round asserted the OPPOSITE and was wrong.
    *
-   * The DTO's own rule (`app-listing-read.schema.ts`, `ListingCard.openCount`) is what
-   * settles it: off-site is ABSENT (`null`), on-site with nothing recorded is a genuine
-   * `0`. And an `onsite` review row is a listing-MEDIA revision of an already-approved,
-   * already-live app — so "nobody could have opened it" is not true of it.
+   * This builder never reads an `AppListingMetric`, so it cannot measure the row at
+   * all, for any kind. The DTO's own definitions (`app-listing-read.schema.ts`,
+   * `ListingCard.openCount`) then decide it: `null` = "not measurable on this surface",
+   * `0` = "measured, and the answer is none". A producer that cannot measure is in the
+   * `null` case.
    *
-   * 🔴 BOTH kinds are asserted, so NO SINGLE CONSTANT passes: a hardcoded `null` reds
-   * the on-site case, a hardcoded `0` reds the off-site case.
+   * 🔴 IT DELIBERATELY DISAGREES WITH `cardOpenCount`, which returns a NUMBER for an
+   * on-site row, and that divergence is the point. Making the two agree on shape (the
+   * earlier round's `0`) bought kind-parity at the price of showing a moderator
+   * "0 plays" for an app with 40,000 — permanently, since this builder never gains a
+   * metric row and `previewQuery` is `retry: false`. Operator call, 2026-09-06: truth
+   * over parity. If you are about to "restore parity", read that trade first.
+   *
+   * 🔴 ON-SITE IS THE LOAD-BEARING CASE. `kind: 'onsite'` is asserted explicitly so the
+   * mutant that reinstates a kind-discriminated `0` reds HERE rather than silently
+   * passing an all-null assertion.
    */
-  it('🔴 openCount: 0 on an on-site row, null off-site — no single constant satisfies both', () => {
+  it('🔴 openCount: null for EVERY kind — this builder can never measure, so it never claims 0', () => {
     const onsite = buildListingCardPreview(row({ id: 'r6', kind: 'onsite' }));
     expect(onsite.kind).toBe('onsite');
-    expect(onsite.openCount).toBe(0);
-    expect(onsite.openCount).not.toBeNull();
+    expect(onsite.openCount).toBeNull();
 
     const offsite = buildListingCardPreview(row({ id: 'r6b', kind: 'offsite' }));
     expect(offsite.kind).toBe('offsite');
     expect(offsite.openCount).toBeNull();
 
     // A row carrying no `kind` at all defaults to off-site (the review adapters'
-    // backward-compatible default) — so it takes the off-site answer, not the on-site one.
+    // backward-compatible default). Same answer, different route to it.
     const defaulted = buildListingCardPreview(row({ id: 'r6c' }));
     expect(defaulted.kind).toBe('offsite');
     expect(defaulted.openCount).toBeNull();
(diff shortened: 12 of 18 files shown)


ZacxDev:
## Round-1 audit fixes — `0f4ada7c09`

All five findings addressed. Every new guard was watched fail on its own assertion before being kept.

### 🟡 1 — `reviewListingPreview.ts` over-nulled the on-site case

`buildListingCardPreview` returned `openCount: null` unconditionally. Now:

```ts
openCount: listingKind(row) === 'onsite' ? 0 : null,
```

The comment's stated reason was factually wrong and has been rewritten. It claimed *"a listing still in the review queue has not been openable by anyone"* — but `OffsitePendingRow.kind === 'onsite'` is a listing-**media** revision of a first-class, already-approved, already-live app, which may have any number of real opens. `null` asserted "structurally unmeasurable", which is false for an on-site app.

This now agrees with all three things that disagreed with it before: the DTO rule this PR wrote (an on-site listing nobody has opened is a genuine `0`), the sibling `installCount: 0` 54 lines below, and the primary producer this builder is the fallback *from* (`projectListingCard`, which returns a number for on-site) — the two feed the same component via `previewQuery.data?.card ?? buildListingCardPreview(request)`.

`reviewListingPreview.test.ts` previously contained **zero** occurrences of `openCount`; it now asserts both kinds plus the absent-`kind` default (→ off-site).

### 🟡 2 — "fails CLOSED for any future kind" is now guarded, not just asserted

Added an unknown-`kind` (`'embedded'`) hydrated row carrying a distinct non-zero count, asserting `openCount === null`. The mutant that removes the property now dies (matrix below). `cardOpenCount`'s comment records that the property is guarded rather than merely stated.

### 🟢 3 — both metrics headers updated

`appListing.metrics.ts` and `appListing.metrics.sql.ts` both said `open_count` has no reader and should be populated with the PR that ships its consumer. Added one additive note to each — the consumer is this PR's card DTO, and the column is still unpopulated. Kept minimal and additive so the conflict with the rollup PR stays trivial.

### 🟢 4 — corrected gate result

**The PR body's "Prettier — clean" was wrong for 1 of 15 files.** Corrected to: **14 clean, 1 pre-existing non-blocking warning.**

Verified by extracting the base blob rather than taking the finding on trust — `prettier` on `app-listing-read.schema.ts` produces exactly **one** hunk, at `:104-108` (`ListAllListingsForModerationInput`), and that hunk is **byte-identical on `origin/main`**. This PR's only hunk in that file is at `+211,30`, nowhere near it. Not reformatted — that would be scope creep.

CI is correctly green: `lint.yml` runs Prettier `continue-on-error: true` for **modified** files and blocking only for **added** ones, and this PR adds no files.

Re-verified after my own comment edit: still exactly one hunk, still only the pre-existing one.

### 🟡 5 — merge-order constraint recorded

Nothing writes `open_count` yet — `appListing.metrics.sql.ts` populates only `install_count`, and its own suite asserts `expect(upsert).not.toContain('open_count')`. **So `cardOpenCount` returns a literal `0` for every on-site listing today**, and the DTO has no third state for "measurable but not yet measured".

> **Stage 4 (the renderer) must not merge before the rollup PR, or must gate the rendered stat row behind a flag.** Otherwise the public `/apps` grid prints "0 plays" on every on-site app including heavily-used ones — by this field's own written standard the worst outcome in the arc, on the public surface.

Recorded in the PR body and in a comment at `cardOpenCount` so the renderer's author cannot miss it. The flag is deliberately not built here.

### Allowlist justification corrected (operator decision recorded)

The card DTO's justification claimed *"the exact same argument as `ListingDetail.installCount`, one step weaker and still sufficient"*. That flattered it on an axis it never named. Rewritten to state both real deltas:

1. **A new public fact, not added precision.** `installCount` is already publicly observable — `popular` is `install_count DESC`, so the ordering is derivable already. No sort exposes `open_count` (`listingSortSchema` = top-rated | popular | newest | name).
2. **Bulk-enumerable by an anonymous caller**, because this is a **card** field while `installCount` is **detail**-only. The card DTO is the body of `GET /api/v1/apps` — public, anon-capable, up to 50 cards per page — so an unauthenticated client can page the play count of the entire approved catalog. `installCount` costs one request per app.

Both were raised and **accepted deliberately**: printing plays on a public card was already the decision, so bulk readability is a consequence of it rather than a new decision, and the `popular` sort already makes relative ordering by usage publicly inferable. Recorded in-comment so the next reviewer gets the accurate comparison instead of the flattering one, and does not re-open a settled question.

### Mutation matrix

Each mutant applied in isolation and reverted (md5-verified) before the next.

| # | Mutant | Test that reds | Its own message |
|---|---|---|---|
| M1 | `cardOpenCount`: `row.kind !== 'onsite'` → `row.kind === 'offsite'` | `🔴 an UNKNOWN future kind projects null — the discriminator FAILS CLOSED, never to 0` | `expected 1487 to be null` @ `app-listing.service.test.ts:1254` |
| M2 | preview ternary → `openCount: null` | `🔴 openCount: 0 on an on-site row, null off-site` (**on-site** half) | `expected null to be +0` @ `reviewListingPreview.test.ts:146` |
| M3 | preview ternary → `openCount: 0` | same test, **off-site** half | `expected +0 to be null` @ `reviewListingPreview.test.ts:151` |

- **M1 was the audit's surviving mutant.** Under it exactly **one** test fails and the other **99** service tests still pass — reproducing the reported 99/99 survival and confirming the new row is the only thing in the suite that can see the difference.
- **M2 and M3 red on different assertion lines**, so the pair pins a relationship: no single constant satisfies both.
- Each died on its own assertion with its own message, not a neighbour's.

### Gate results

| Gate | Result |
|---|---|
| Unit tests (3 affected suites) | **126/126 pass** |
| `pnpm typecheck` | **0 errors** in 100s. Positive control: narrowing `openCount` to `number` → **rc=2, exactly 5 errors**, incl. `app-listing.service.ts:371` and `reviewListingPreview.ts:139` |
| ESLint (7 touched files) | **4 problems, all pre-existing** — `_a` unused-vars at `app-listing.service.test.ts:25-29`, byte-identical to `origin/main`. **0 new.** Positive-controlled against a deliberately bad file (3 problems reported; note its rc was **0**, so this was counted, not read off an exit code) |
| Prettier (7 touched files) | **6 clean, 1 pre-existing hunk** (finding 4). Positive-controlled on the same bad file |

**Could not verify:** the `component` / `geometry` browser tiers do not run in this environment (chromium store mismatch) — environment, not a defect. Those suites are unchanged by this commit apart from the type-level `openCount` fixture additions already in the PR.

```audit-claims round=1 audited=05353a305a687eb8b9b6a58ef7bb51605c2251d1..0f4ada7c09212fa955973f762da55e686ceaf472
1. buildListingCardPreview returns 0 for an on-site row and null for an off-site row, instead of null unconditionally.
2. The comment at that field no longer claims a review-queue listing could not have been opened by anyone.
3. reviewListingPreview.test.ts asserts openCount for on-site, off-site, and absent-kind rows; it previously asserted none.
4. app-listing.service.test.ts asserts that a row with an unknown kind projects openCount === null.
5. The kind !== 'onsite' -> kind === 'offsite' mutant is killed by that test and by no other test in the file.
6. cardOpenCount's comment states the fail-closed property is guarded by that test.
7. cardOpenCount's comment states the merge-order constraint: the renderer must not merge before the rollup, or must gate the stat row behind a flag.
8. appListing.metrics.ts and appListing.metrics.sql.ts each record that open_count now has a consumer and is still unpopulated.
9. The ListingCard.openCount allowlist justification states that openCount is a card field, bulk-enumerable by an anonymous caller over GET /api/v1/apps at up to 50 per page, while installCount is detail-only.
10. That justification states no sort exposes open_count, so the number is a new public fact rather than added precision.
11. That justification records that both deltas were raised and accepted deliberately.
12. The PR body's Prettier claim is corrected to 14 clean, 1 pre-existing non-blocking warning.
13. The pre-existing Prettier hunk at app-listing-read.schema.ts:104-108 was not reformatted.
14. No production behaviour changed for on-site or off-site rows projected by projectListingCard.
```

ZacxDev:
## Round 2 (delta `05353a305a..0f4ada7c09`) — 3 findings, fixed in `689e82278b`

Two of the three were defects **round 1 introduced**, which is exactly why the round existed.

### 🟡 The preview goes back to `null` — round 1's premise was right, its conclusion was not

Round 1 correctly established that *"a listing still in the review queue has not been openable by anyone"* is **false**: an `onsite` review row is a listing-MEDIA revision of an already-approved, already-live app. It then concluded the value should be `0`. That did not follow.

`buildListingCardPreview` never reads an `AppListingMetric`. It cannot measure the row **at all, for any kind, ever**. The DTO's own definitions settle it — `null` = "not measurable on this surface", `0` = "measured, and the answer is none" — and a producer that cannot measure is in the `null` case.

Returning `0` bought shape-parity with `cardOpenCount` at the price of showing a moderator **"0 plays" for an app with 40,000** — and *permanently*, because this builder never gains a metric row and `previewQuery` is `retry: false`, so a single error pins the fallback for that mount. **Operator decision: truth over parity.** Accepted consequence: the stat row can appear when `previewQuery` resolves. A row that appears is recoverable; a wrong number that never corrects itself is not.

🔴 **The value returns to where it started, but the two things that were genuinely wrong before round 1 do not.** The comment's false premise stays corrected, and the file keeps the tests it did not have — it contained **zero** occurrences of `openCount` before round 1. Net, this is better than either prior state.

Mutation-verified both ways, each applied alone and reverted: reinstating the kind-discriminated `0` → 1 red, `expected +0 to be null`; an unconditional `0` → same guard, same message.

The sibling `installCount: 0` on that surface is a false zero of the same class. **Deliberately left alone** — out of scope, and inert (`buildListingStatChips` returns `[]` under `preview`) — with a note in the code not to cite it as precedent for reverting this.

### 🟡 Round 1 manufactured the merge conflict it then called trivial — my instruction, my error

Round 1 added a note to `appListing.metrics.ts` and `appListing.metrics.sql.ts` and claimed *"minimal and additive so the conflict stays trivial"*. Measured with a 3-way `git merge-file` against #4653's tip, with a positive control (theirs := base → rc 0, 0 hunks): **one conflict hunk in each file, rc 1**.

Worse than the count: the content is **contradictory, not adjacent**. This PR's note said `open_count` is NOT populated; #4653 rewrites the same paragraph to say it **is**, from the ClickHouse `App_Open` stream. Keeping both sides — the reflex when resolving — ships a file whose comment denies what its own SQL does.

**Before that commit this PR touched neither file.** Both notes are removed. Verified disjoint again by comparing `git diff origin/main...HEAD --name-only` on each branch (16 files vs 8, **no shared path**) — note the two-dot form is useless here, it is polluted by base drift. #4653 already updates those headers correctly and should own them; the transient staleness if this merges first is one comment, and cheaper than a contradictory conflict.

### 🟢 Stale self-reported numbers in the PR body, corrected

Round 1 corrected one number in the body while invalidating its neighbours. Re-derived and fixed: **16** changed files (not 15), prettier **15 clean + 1 pre-existing** (not 14+1), service-suite baseline **100** (not 99), and M1 now reds **7** tests (not 6) because round 1's fail-closed test also kills it. The Prettier warning remains the same single pre-existing hunk at `app-listing-read.schema.ts:104-108`, outside this PR's diff and deliberately not reformatted.

### Gates at `689e82278b`

116 tests across the two affected suites · `pnpm typecheck` **OK — 0 type errors in 85 s** (read from the tool's own terminal marker, not from an absence of matches — an earlier read of this same log returned a "clean" zero while the run was still in progress) · eslint **0 output lines** (counted, not `rc` — this repo's eslint has returned `rc=0` alongside findings) · prettier as above.

### Still carried forward

**Merge order.** Nothing writes `open_count` yet, so `cardOpenCount` returns a literal `0` for every on-site listing today. **Stage 4 must not merge before #4653**, or the rendered row must be flagged — otherwise the public grid shows "0 plays" on every on-site app, including heavily-used ones. Recorded at `cardOpenCount` and in the body.

```audit-claims round=2 audited=0f4ada7c09212fa955973f762da55e686ceaf472..689e82278b
1. Reverted buildListingCardPreview to an unconditional null, keeping round 1's correction of the false "nobody could have opened it" premise and the tests it added.
2. Recorded the reason in code: this builder never reads a metric row, so by the DTO's definitions it is in the null case, and the divergence from cardOpenCount is deliberate.
3. Removed both metrics-header notes, so this PR and #4653 touch no file in common again.
4. Corrected four stale self-reported numbers in the PR body (16 files, 15 clean, baseline 100, M1 reds 7).
5. Corrected the PR body paragraph that still described the preview value as null for the refuted reason.
```

ZacxDev:
## Round 3 (delta `0f4ada7c09..689e82278b`) — fixes in `a49eaad27f`. **Ladder STOPPED.**

Two 🟡 and one 🟢, all prose, all written by the previous round.

### 🔴 The one that mattered: my comment sourced the decision to a file that forbids it

The round-2 comment said:

> *"The DTO's own definitions then decide it: `null` = not measurable on this surface, `0` = measured and the answer is none. **A producer that cannot measure is in the `null` case.**"*

The DTO — in a hunk **this same PR writes** — says the opposite for exactly this input: an on-site listing with no metric row is a genuine `0` and must **not** be over-nulled, and `null` there means "structurally unmeasurable", a property of the **kind**, not of the producer. It has no "producer that cannot measure" case at all. `buildListingCardPreview`'s on-site input *is* "on-site, no metric row", so the written rule yields `0`. We return `null` anyway.

That is fine as a decision and wrong as a derivation — and the danger isn't the mislabel. **`cardOpenCount` also cannot measure today**, because nothing writes `open_count` yet. So "a producer that cannot measure returns `null`" licenses returning `null` for on-site rows on the **public** `GET /api/v1/apps` card DTO — which is **mutant M2 in this PR's own matrix**, and precisely what the DTO's "do not over-null" paragraph exists to forbid. A later round applying the rule as written would break the public card while believing it was following the file's own convention.

Both sites now state it honestly: the DTO's rule yields `0` here, this is a deliberate override **scoped to this builder**, and the reason is a judgement about this surface — unlike `cardOpenCount` it can never become right later, so a `0` is permanently wrong on a moderator screen. Each carries an explicit **"do not generalise this into a rule"** naming the failure it would cause.

### 🟢 And the sibling that argued for itself with the refuted premise

`installCount: 0` fifty lines below still justified itself with *"an unapproved listing has never been installable"* — the same premise these rounds spent two iterations refuting, and false for the same input. Left in place (out of scope, and **inert on both consumers** — `buildListingStatChips` returns `[]` under `preview`, and `buildListingDetailRows`' install row is `!opts.preview`-gated, a second consumer the earlier note failed to name) but no longer arguing from a refuted claim, and no longer readable as precedent for reverting `openCount`.

### PR body corrected

Its round-1 section still described the tree in the present tense — *"now discriminates on kind"* and *"both module headers record…, the conflict stays trivial"* — both false at head, so the body asserted and denied the same two facts. Both items are struck through and SHA-scoped.

### Why the ladder stops — on the oscillation, not on a clean round

Round 3's findings were real, so the findings rule alone would continue. It stops on what the audit measured instead:

> **Filtering comments out of `05353a305a..689e82278b` across all five payload files returns an empty diff.**

Round 1 changed `openCount: null` → a ternary. Round 3 changed it back. **Two full fix rounds produced zero net executable change**, and the value has oscillated `null → 0 → null`. The rounds are undoing each other — the attribution gate's condition arriving by a different route than two consecutive zero-payload rounds. The behaviour this PR ships is **byte-identical to its first commit**; everything since is comments and tests.

That is not wasted: round 1 caught a genuinely false comment, round 2 caught a manufactured merge conflict and a permanent false zero, round 3 caught a rule wider than its decision. But the remaining risk is comment accuracy on a field **no renderer reads at this ref**, and another round would audit prose about prose.

### Gates

Round 3's own audit ran on the **merged tree** (`origin/main` at `6ff42aed42`, which now contains #4652, + this branch): typecheck **OK — 0 type errors**, **5,953 unit tests** across 254 files, eslint 4 pre-existing warnings byte-identical to `main`, prettier 15 clean + 1 pre-existing. My fix commit: 116 tests across the two affected suites, eslint 0, prettier clean.

### Still carried forward

🔴 **Merge order.** Nothing writes `open_count` yet, so `cardOpenCount` returns a literal `0` for every on-site listing. **Stage 4 must not merge before #4653**, or the rendered row must be flagged — otherwise the public grid shows "0 plays" on every on-site app. Recorded at `cardOpenCount`.

🔴 **#4652 is merged (`6ff42aed42`) but its ClickHouse `Enum16` migration is NOT yet applied.** It must be applied before the next `release` cut that includes it, or `App_Open` rows are silently rejected at the tracker.

```audit-claims round=3 audited=689e82278bf699042a2ab0fc32ee4d6c9bfab5ca..a49eaad27f
1. Re-stated the preview's null as a deliberate OVERRIDE of the DTO's rule, not a derivation from it, in both the builder comment and the test docblock.
2. Removed the over-wide "a producer that cannot measure returns null" phrasing and recorded the public-card failure it would license.
3. Stopped installCount: 0 justifying itself with the refuted "never been installable" premise, and named the second !opts.preview-gated consumer the earlier note missed.
4. Struck through and SHA-scoped the two superseded round-1 items in the PR body.
5. LADDER STOPPED: two fix rounds produced zero net executable change and the value oscillated null -> 0 -> null.
```

PR #4660, opened 2026-09-06
perf(blocks): drop unplayable collections from discovery via a bounded, order-pinned sample

ZacxDev:
> **Revised.** The first revision of this PR computed an **exact clamped item count** over the discovery over-fetch window. That was measured against a production-scale replica and is a **~31× regression on the discovery front door**, so it is gone. The correctness work survives; the expensive half is replaced by a bounded, order-pinned sample. The measurements and what they force are in [Cost](#cost--why-the-exact-count-is-gone).

## The problem

On a SFW domain, App Blocks collection discovery is dominated by collections that are overwhelmingly mature, and every card advertises an item count the player will never serve.

Measured on the live first page (24 rows):

| | advertised | playable at a SFW ceiling | |
|---|---|---|---|
| rank 2 — a mature video collection | 2,078 | 216 | 10% |
| rank 13 — a mature image collection | 2,080 | 19 | 1% |
| a large contest collection | 34,577 | ~97% | fine |

**A collection's own `nsfwLevel` cannot separate these.** It is a bitmask OR-ed over the collection's items, so the 97%-safe contest collection and the 1%-safe mature one carry the **identical** value (29 = PG|R|X|XXX), and both *intersect* a SFW ceiling.

**Gating harder on it is not an option either.** Strict containment (`nsfwLevel & ~browsingLevel === 0`) was measured against the live population and **rejected**: it drops **21 of the 24 rows on page 1 (87.5%)**, including every contest collection, and takes the whole population from 29,095 to 12,389 (−57.4%). This PR deliberately does not implement it.

The separating signal is **how much of the collection survives the ceiling**.

## Cost — why the exact count is gone

Measured on a production-scale replica, over the real page-1 over-fetch window under this endpoint's default sort (`Most Followers`): **97 collections, 298,469 accepted items, largest 34,577, mean 3,077.** Steady-state, cold run discarded, the id-selection subquery removed so only the count is timed.

| query | run 1 | run 2 |
|---|---|---|
| **before** — 1 unclamped count, 24 ids (what `main` runs) | **85 ms** | **82 ms** |
| unclamped, 97 ids | 84 ms | 78 ms |
| **exact CLAMPED, 97 ids (revision 1 of this PR)** | **2829 ms** | **2563 ms** |
| exact clamped, 24 ids | 2076 ms | 2079 ms |
| **THIS QUERY, as shipped (`ORDER BY id DESC`)** | **399 ms** | **412 ms** |
| ...the same with a bare `LIMIT`, no `ORDER BY` | 102 ms | 107 ms |
| unordered sample, cap 500/collection | 354 ms | — |

**Mechanism.** The unclamped count is an `Index Only Scan` on `CollectionItem_collectionId_status_covered` — no heap access at all. `LEFT JOIN "Image"` forfeits that covering index and becomes a `Nested Loop Left Join` over ~300k rows, `Buffers: shared hit=1713835`.

**Narrowing `OVERFETCH` is not the fix**, and revision 1's PR body was wrong to name it as the first knob: clamping 24 ids instead of 97 costs 2076 ms against 2829 ms — **~27%, not the order of magnitude needed.** The cost is dominated by a handful of very large collections that are on page 1 *because* the sort ranks them there.

🔴 **The last three rows are a CORRECTION.** An earlier revision of this PR quoted **257 ms** for the sample and said it was measured with this ordering. It was not — 257 ms was an **unordered** `LIMIT 200`, and the order pin (the thing that makes the 1-in-84 verdict deterministic rather than a coin toss) was added afterwards. Re-measured on the shipped shape it is **~400 ms**: still ~7× cheaper than the exact clamp it replaces, so the decision is unchanged, but ~4.7× the unclamped baseline rather than the ~3× the old number implied. **The figure was never asserted on by any test**, which is exactly how it travelled from a probe into a comment, a commit message and this body unchallenged.

**Where the pin's cost comes from**, since ~4× is more than it looks: no index gives `(collectionId, status)` rows in `id` order, so the LATERAL reads the **whole collection** through the covering index — mean 3,077 rows, 34,577 at the tail — and quicksorts it before taking 200. The cap bounds what the JOIN and the count touch, **not what the scan reads**. An index on `(collectionId, status, id DESC)` would remove the sort; that is a migration, applied by hand per environment here, so it is left as a follow-up rather than smuggled into this PR.

**Accuracy of the sample**, over the same 97 collections: a 200-item sample and the exact count agree on the 20% floor for **97 of 97** (63 keeps either way, **0 disagreements**). 500 buys nothing measurable and widens the worst case, so the cap is **200**.

### 🔴 It is order-sensitive, and that shapes the design

Comparing an **oldest-200** sample against a **newest-200** sample over the **84** collections holding more than 200 items:

- **1 order disagreement** — one collection's floor verdict flips depending on which end you read.
- **max fraction gap 0.800** — one collection differs by 80 percentage points between its two ends. Its character changed over time.
- mean gap 0.038.

So the sample **pins its order explicitly** — `ORDER BY ci."id" DESC` inside the LATERAL — and never a bare `LIMIT`, which takes *physical* row order: neither stable nor reviewable, and it would make that 1-in-84 verdict a coin flip. **Newest-first is the defensible end:** it is what a viewer opening the collection sees first, and a collection that has recently gone mature should be judged on what it is *now*, not what it was.

Both numbers are recorded in the code next to `PLAYABLE_SAMPLE_SIZE`, so nobody has to re-measure to find them.

## What this does

1. **`getCollectionPlayableSample` (new, in `block-collections.service.ts`).** One bounded `CROSS JOIN LATERAL` per candidate collection: the newest `PLAYABLE_SAMPLE_SIZE` (**200**, exported) accepted items, and how many of them the ceiling permits. Returns `{sampled, playable}` per id.
2. **Public discovery drops collections below `MIN_PLAYABLE_FRACTION` (0.2)**, computed from that sample.
3. **`mode=public` `itemCount` is the ADVERTISED (unclamped) count** — the same query, over the same `limit`-sized id list, that this endpoint has always run. It is not a lie: it is the collection's real size, and the floor now guarantees most of it is playable. A **sample-extrapolated estimate is deliberately not substituted** — a field that reads as exact must not carry one.
4. **`mode=mine` keeps the exact clamped count.** That branch counts only the subject's own, already-sliced collections: a bounded population with no over-fetch window and no popularity-ranked giants, so it does not have this cost profile. The reasoning is stated at the call site.
5. **`getCollectionItemCount` keeps its optional `browsingLevel`** — the LEFT JOIN, the `imageId IS NULL` escape, the fail-closed orphan handling and its whole PGlite suite are unchanged. **It has one production caller: `mode=mine`.** It is not dead code, but it is no longer on the hot discovery path.

### 🔴 The floor is a ranking heuristic, not a safety gate

Stated where it is defined, in both the constant's doc and the sample function's. It decides whether a card is worth *showing*, on the theory that a card promising 2,080 items and opening onto 19 is a bad card. **It protects nobody from mature media.** Per-item maturity is enforced on the **detail endpoint**, where the media is actually read, and on the cover via `getFallbackCoverImages`. A collection this floor keeps can still be 79% mature; a collection it drops was never unsafe. And it is now an **approximation** of that heuristic — the fraction is over the collection's recent items, not the whole collection.

### The sample is not image-only

`nsfwLevel` lives on `Image`, so an inner `JOIN "Image"` would score every model / post / article item as unplayable — and those collections would not merely rank lower, they would report `sampled: 0` and fall out. Hence `LEFT JOIN` plus an explicit `s."imageId" IS NULL` escape. An item whose `Image` row is gone is **sampled but not playable** (fail-closed, matching `getFallbackCoverImages`).

Covered by executing the real SQL on an in-process Postgres (PGlite), not by string-matching it — `JOIN` and `LEFT JOIN`, and `LIMIT` with and without `ORDER BY`, contain the same fragments in the same order, so no text assertion can tell them apart.

### The floor is discovery-only

`mode=mine` never drops. The subject created or bookmarked every collection in that list; hiding one for being mostly mature makes it look deleted — the defect class the detail surface already settled the other way (an empty own-collection is shown *disabled with a reason*, never hidden). That branch does not even issue the sample query.

### A collection with nothing sampled is kept

`0 / 0` is `NaN` and every `NaN` comparison is false, so without an explicit branch it would be silently dropped by a filter that has nothing to say about it. The floor detects a *maturity mismatch*; a collection with no items has none.

## Pagination

The public branch already over-fetches `limit * 4 + 1` and walks rows until the page fills, so **the floor is applied inside that walk, not after the page is chosen.** A dropped row is walked past exactly like a maturity-clamped one, so the page still fills toward `limit` and the keyset cursor still advances.

**The cost, stated plainly:** the floor drops more rows than the ceiling alone did, so the "consumed the entire over-fetch without filling `limit`" branch is no longer rare — **a page can come back short**, occasionally empty. What it never does is come back *unwalkable*: that branch sets the cursor to the last fetched row, advancing by `OVERFETCH - 1` rows per request. Filtering after the slice was rejected precisely because a page that filtered to empty would emit no cursor and **terminate the feed** while qualifying collections remained. A test drives exactly that page and asserts the cursor advances.

**Per-request query cost after this change:** one `getAllCollections` (unchanged) + **one bounded sample over the ceiling-passing candidates** (~400 ms measured, as shipped) + the unclamped count over the final page (~85 ms, unchanged from `main`) + followed/fallback-cover lookups (unchanged). Revision 1's two grouped counts, one of them clamped over the whole over-fetch window, are gone.

## Call-site ledger

`collection-item-count-clamp-wiring.test.ts` enumerates every `getCollectionItemCount` call site in `src/` from the **tree** via the TypeScript AST and pins, per `file::enclosingFunction`, **in source order**, whether each one clamps.

It now guards the **cost** as much as the correctness, because the optional parameter is silent in both directions:

- Adding it to the public branch restores the ~31× regression. Nothing about the response changes shape — it just gets slow.
- Dropping it from `mode=mine` sends a viewer's own card back to advertising items their ceiling hides.

Both are mutants below, and each is killed.

## Verification

- **Typecheck** (`node scripts/typecheck.mjs`, run from the worktree): **0 errors**.
- **Unit** (`node scripts/test-unit-run.mjs`): the four affected files are **65/65 green**. **Full unit suite: 38,256 passed / 26 failed / 33 skipped over 1,649 files.** The 26 failures are in three files this PR does not touch (`recent-source-images.store`, `sticker-reveal.store`, `AppBlocks/hiddenBlocks`); a **pristine `origin/main` worktree reproduces exactly the same 26**, so they are pre-existing and unrelated.
- **Red at base**: the new/changed tests were run against two bases with only the three production files reverted.
  - **at `origin/main` = `1de892bab9`** (feature absent): **30 failed / 35 passed** of 65 — the whole new sample suite (14/14), the ledger's pin, 11 endpoint tests, and the clamped half of the `getCollectionItemCount` PGlite suite (4 of 11; the unclamped half correctly still passes, which is the "existing callers are unaffected" claim). `main` has since advanced to `72022e4be2` (#4651, unrelated to collections and touching none of these files).
  - **at `e1a58307d6`** (revision 1 — the exact clamped count): **20 failed / 45 passed**. The interesting ones: the ledger fails on the old three-call clamped shape, and *"the count query is NOT clamped"* fails — i.e. the new suite catches the design this revision replaces, not just its absence.
- **Mutation**: **15 targeted mutants, every one `cmp`-verified to have landed before running** (a no-op edit reads exactly like a survivor). **All 15 killed, zero survivors**, each by an assertion that names the thing broken:

  | mutant | killed by |
  |---|---|
  | drop `ORDER BY` from the LATERAL | "samples the NEWEST items" + "reading the wrong end FLIPS THE FLOOR VERDICT" |
  | `ORDER BY id DESC` → `ASC` | same two |
  | `LEFT JOIN "Image"` → `JOIN` | "keeps NON-IMAGE items unconditionally" (+6) |
  | `PLAYABLE_SAMPLE_SIZE` 200 → 201 | "PLAYABLE_SAMPLE_SIZE is 200 — pinned absolutely" (only that one; every other fixture derives from the constant, which is why the absolute pin exists) |
  | drop the `LIMIT` | "the cap is PER COLLECTION" + "the DEFAULT sample size is PLAYABLE_SAMPLE_SIZE" |
  | `LIMIT n` → `LIMIT n + 1` | "the DEFAULT sample size is PLAYABLE_SAMPLE_SIZE" |
  | drop the `imageId IS NULL` escape | "keeps NON-IMAGE items unconditionally" (+5) |
  | drop the `nsfwLevel = 0` (unrated) clause | "a MIXED collection keeps permitted images…" (+3) |
  | swap `meetsPlayableFloor` args | the 5 public-floor endpoint tests |
  | public `itemCount` ← sampled playable | "the count query is NOT clamped" + "DROPS a collection below the floor" |
  | public count gains `browsingLevel` | the **ledger** + "the count query is NOT clamped" |
  | drop the `sampled <= 0` branch | "nothing sampled is KEPT" (+10) |
  | `>=` → `>` at the floor | "inclusive at the floor and exclusive below it" |
  | endpoint passes its own cap literal | "the SAMPLE covers the CEILING-PASSING candidates" (no third argument) |
  | `mine` count loses `browsingLevel` | the **ledger** + "`mode=mine` itemCount is the EXACT CLAMPED count" |

- **Merged-tree check with the one overlapping open PR, #4597** (`feat/allow-creators-to-archive-collections`; overlap is `collection.service.ts`). Re-run for this revision, on `main@72022e4be2` + #4597 + this branch: **merges with no conflict**, and both PRs' suites are **72/72 green together**. Typecheck on that tree reports 2 errors, both inside #4597's own files (`archivedAt` missing from the generated Prisma client) — and the **control** (`main` + #4597 **without** this PR) reports **the identical 2**, so this PR contributes none. **Not based on #4597.**
- **Formatting**: `src/pages/api/v1/blocks/collections/index.ts` and `src/tests/api/v1/blocks/collections-endpoint.test.ts` do not satisfy `prettier --check` — **they already did not on `origin/main`**, so this PR neither introduces nor fixes it, and reformatting them would bury the diff.

## Tests kept, rewritten, deleted

- **Kept unchanged:** `collection-item-count-clamp.behavior.test.ts` (the whole PGlite suite for `getCollectionItemCount` — the function is unchanged), and every pre-existing endpoint test (authz, sort aliases, cover fallback/clamp, pagination, `mode=mine` read-split, 400 shape).
- **Rewritten:** the five floor tests in `collections-endpoint.test.ts`, from "two count queries" onto the sample; the `meetsPlayableFloor` unit block, for the `sampled` rename; the ledger's expectation row (three calls → two) and its rationale.
- **Added:** `collection-playable-sample.behavior.test.ts` (14 PGlite tests), a test pinning that the public count is **not** clamped and covers only the final page, and a test pinning that the endpoint passes **no** cap of its own.
- **Deleted:** nothing.

## Notes

- No Prisma migration — nothing schema-level changed.
- Two endpoint tests pass at base *and* at HEAD by construction (`mode=mine` never drops; a collection with nothing sampled is never dropped), because with no floor at all there is nothing to drop. They are **invariant guards** with respect to `main`, not regression coverage; their evidence is the mutation run.
- **The measurements in this body are not reproducible from CI.** They were taken by hand against a production-scale replica. The accuracy and order-sensitivity figures (97/97, 1-in-84, 0.800) come from the same session and are recorded in the code so they survive this PR.

### DIFF
--- a/src/pages/api/v1/blocks/collections/index.ts
+++ b/src/pages/api/v1/blocks/collections/index.ts
@@ -45,8 +45,20 @@ import {
  * can't surface a mature collection in discovery. (Per-item maturity is enforced
  * on the detail endpoint where the media is actually read.)
  *
+ * 🔴 THE COLLECTION-LEVEL `nsfwLevel` CANNOT SEPARATE A MOSTLY-SAFE COLLECTION
+ * FROM A MOSTLY-MATURE ONE, which is why the clamped count below exists. It is a
+ * bitmask OR-ed over the collection's items, so a contest collection that is 97%
+ * safe and a mature collection that is 1% safe both carry the same value (29 =
+ * PG|R|X|XXX) and both INTERSECT a SFW ceiling. Gating harder on that value is not
+ * an option: strict containment (`nsfwLevel & ~browsingLevel === 0`) was measured
+ * against the live population and drops 87.5% of the first discovery page,
+ * including the large contest collections that are the best content on a SFW
+ * domain. The separating signal is the CLAMPED ITEM COUNT — see
+ * `MIN_PLAYABLE_FRACTION`.
+ *
  * Response: `{ items: [{ id, name, description, coverImageUrl, itemCount,
- *   curator:{ userId, username }, isPublic, followed }], nextCursor }`.
+ *   curator:{ userId, username }, isPublic, followed }], nextCursor }`, where
+ * `itemCount` is CLAMPED to the token's ceiling in BOTH modes.
  */
 
 export const config = { api: { responseLimit: false } };
@@ -60,6 +72,38 @@ const SORT_ALIAS: Record<string, CollectionSort> = {
   newest: CollectionSort.Newest,
   popular: CollectionSort.MostContributors,
 };
+
+/**
+ * The PLAYABLE-FRACTION FLOOR: the share of a collection's advertised (accepted)
+ * items that must survive the viewer's maturity ceiling for that collection to be
+ * worth surfacing in PUBLIC discovery.
+ *
+ * A collection at 1% is not a collection the viewer can browse; it is a card
+ * promising 2,080 items that opens onto 19. Twenty percent is the point measured
+ * to separate the two live populations — the mostly-safe contest collections sit
+ * far above it and the mature ones far below — without needing a second signal.
+ *
+ * 🔴 DISCOVERY ONLY. It is deliberately NOT applied in `mode=mine`; see the drop
+ * comment on the public branch and the note above the `mine` branch.
+ */
+export const MIN_PLAYABLE_FRACTION = 0.2;
+
+/**
+ * Does `playable` of `advertised` items clear {@link MIN_PLAYABLE_FRACTION}?
+ *
+ * 🔴 THE ZERO CASE IS AN EXPLICIT BRANCH, NOT A DIVISION. An empty collection
+ * gives `0 / 0 = NaN`, and every comparison against NaN is false — so without
+ * this guard an empty collection would be silently DROPPED by a filter that has
+ * nothing to say about it. This filter exists to catch a MATURITY MISMATCH
+ * between what a card advertises and what it can serve, and a collection
+ * advertising nothing has no mismatch to detect. Whether an empty collection
+ * belongs in discovery at all is a separate product question this must not decide
+ * as a side effect.
+ */
+export function meetsPlayableFloor(advertised: number, playable: number): boolean {
+  if (advertised <= 0) return true;
+  return playable / advertised >= MIN_PLAYABLE_FRACTION;
+}
 
 const querySchema = z.object({
   mode: z.enum(['public', 'mine']).default('public'),
@@ -160,6 +204,33 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
         },
       });
 
+      // The playable-fraction floor needs BOTH counts, and it decides which rows
+      // the walk below consumes — so the counts have to be resolved BEFORE the
+      // walk, over every row that could still appear, not after the page is
+      // chosen. Only rows that already clear the collection-level ceiling can
+      // appear, so those are the only ids worth counting.
+      //
+      // COST: two grouped counts over up to `OVERFETCH` ids instead of one over
+      // `limit`. Both are the same indexed GROUP BY the page already ran.
+      const ceilingOk = (c: (typeof rows)[number]) =>
+        collectionWithinCeiling(c.nsfwLevel ?? 0, browsingLevel);
+      const candidateIds = rows.filter(ceilingOk).map((c) => c.id);
+      const [advertisedRows, playableRows] = await Promise.all([
+        getCollectionItemCount({
+          collectionIds: candidateIds,
+          status: CollectionItemStatus.ACCEPTED,
+        }),
+        // The SAME count, clamped to this token's ceiling — what the player will
+        // actually serve.
+        getCollectionItemCount({
+          collectionIds: candidateIds,
+          status: CollectionItemStatus.ACCEPTED,
+          browsingLevel,
+        }),
+      ]);
+      const advertisedMap = new Map(advertisedRows.map((c) => [c.id, Number(c.count)]));
+      const playableMap = new Map(playableRows.map((c) => [c.id, Number(c.count)]));
+
       const items: typeof rows = [];
       let firstUnconsumedId: number | undefined;
       for (let i = 0; i < rows.length; i++) {
@@ -167,7 +238,19 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
           firstUnconsumedId = rows[i].id;
           break;
         }
-        if (collectionWithinCeiling(rows[i].nsfwLevel ?? 0, browsingLevel)) items.push(rows[i]);
+        if (!ceilingOk(rows[i])) continue;
+        // 🔴 THE DROP, AND WHY IT IS INSIDE THIS WALK RATHER THAN AFTER IT. Both
+        // count maps are absent-means-zero (GROUP BY emits no row for a collection
+        // with no accepted items), which `meetsPlayableFloor` reads as "empty, keep".
+        // Filtering here means a dropped row is walked PAST like a clamped-out one,
+        // so the page still fills to `limit` and `firstUnconsumedId` still advances
+        // — filtering the sliced page afterwards would return short pages and, on a
+        // page that filtered to empty, emit no cursor at all and TERMINATE the feed
+        // while qualifying collections remained.
+        const advertised = advertisedMap.get(rows[i].id) ?? 0;
+        const playable = playableMap.get(rows[i].id) ?? 0;
+        if (!meetsPlayableFloor(advertised, playable)) continue;
+        items.push(rows[i]);
       }
 
       let nextCursor: number | undefined;
@@ -178,7 +261,10 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
         // Consumed the ENTIRE over-fetch without filling `limit` (a very heavy
         // clamp) yet the source returned a full batch → more may remain. Resume
         // from the last fetched row (inclusive → re-fetched next page; the client
-        // dedups by id). Rare (needs the clamp to drop most of 4×limit+1 rows).
+        // dedups by id). No longer rare: the playable floor drops far more rows
+        // than the ceiling alone did, so a page CAN come back short — but it comes
+        // back with a cursor that advanced by OVERFETCH-1 rows, which is what
+        // keeps the rest of the feed reachable. Short page, live feed.
         nextCursor = rows[rows.length - 1]?.id;
       }
       // else: rows.length < OVERFETCH and the page wasn't over-consumed → the
@@ -195,12 +281,13 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
       const primaryCoverUsable = (c: (typeof items)[number]) =>
         !!c.image?.url && collectionWithinCeiling(c.image.nsfwLevel ?? 0, browsingLevel);
       const missingCoverIds = items.filter((c) => !primaryCoverUsable(c)).map((c) => c.id);
-      const [countRows, followed, fallbackCovers] = await Promise.all([
-        getCollectionItemCount({ collectionIds: ids, status: CollectionItemStatus.ACCEPTED }),
+      const [followed, fallbackCovers] = await Promise.all([
         getFollowedCollectionIds(subjectUserId, ids),
         getFallbackCoverImages(missingCoverIds, browsingLevel),
       ]);
-      const countMap = new Map(countRows.map((c) => [c.id, Number(c.count)]));
+      // No third count query here: `playableMap` was already resolved above for the
+      // floor, over a SUPERSET of `ids`, and it is the clamped number the card must
+      // advertise.
 
       res.status(200).json({
         items: items.map((c) => ({
@@ -208,7 +295,7 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
           name: c.name,
           description: c.description ?? null,
           coverImageUrl: toCoverImageUrl(primaryCoverUsable(c) ? c.image : fallbackCovers.get(c.id)),
-          itemCount: countMap.get(c.id) ?? 0,
+          itemCount: playableMap.get(c.id) ?? 0,
           curator: { userId: c.userId, username: c.user?.username ?? null },
           isPublic: c.read === CollectionReadConfiguration.Public,
           followed: followed.has(c.id),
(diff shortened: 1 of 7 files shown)


ZacxDev:
### Local verification (runner: `node scripts/test-unit-run.mjs` → `node_modules/.bin/vitest run --project 'unit*'`)

**Full unit suite on the committed tree:** `Tests 1 failed | 38266 passed | 33 skipped (38300)`, `Test Files 1 failed | 1644 passed`.

The single failure is `src/server/__tests__/eventloop-watchdog.capture.test.ts > 🔴 classifies a wedge as starved even while OTHER threads burn CPU`. Three independent measures say it is not this PR:

- the file contains **0** references to `getCollectionItemCount`, `collection`, `browsingLevel` or the floor;
- run alone on an unloaded box it passes **9/9**;
- all four CI `Unit tests` shards pass here, which covers that file.

It is a CPU-starvation classifier asserting a 12s timing property, and my box was running the full suite concurrently with CI. Reported rather than waved away — I did not bisect it against `main`.

**Typecheck:** `node scripts/typecheck.mjs` → 0 errors. **ESLint:** 0 errors (warnings only, all pre-existing `any` in the test file's house style).

### Red-at-base / green-at-HEAD

Reverting **only** the two production files (`collection.service.ts`, `blocks/collections/index.ts`) and keeping the tests: **16 of the 27 new tests fail**, 50/50 pass at HEAD.

The 11 that pass at base do so because with no floor at all there is nothing to drop — they are **invariant guards, not regression coverage**, and I am labelling them as such rather than counting them:

| green at base | proven instead by |
|---|---|
| `mode=mine`: FAR below the floor is still returned | mutant M9 (floor applied to `mine`) |
| a collection advertising ZERO items is not dropped | mutants M6/M7 (zero branch removed / inverted) |
| keeps NON-IMAGE items unconditionally | mutants M1/M2 (inner join / escape dropped) |
| no UNEXPECTED call site | mutant M13 (a new unpinned caller) |
| 3 unclamped-baseline tests, "ceiling permits everything" | positive controls, by design |
| the status filter composes with the clamp | **nothing** — genuine invariant guard, unproven |
| a collection with nothing to count stays absent | **nothing** — genuine invariant guard, unproven |
| no call site passes an unreadable argument | **nothing** — genuine invariant guard, unproven |

### Mutation results — 13 mutants, 13 killed, 0 survivors

Every edit was `cmp`-verified against a backup before running, so a no-op edit could not be scored as a survivor.

| mutant | killed by | assertion |
|---|---|---|
| M1 `LEFT JOIN` → `JOIN` | keeps NON-IMAGE items unconditionally | `expected +0 to be 3` |
| M2 drop the `imageId IS NULL` escape | same | `expected +0 to be 3` |
| M3 `!= null` → truthiness (WHERE) | `browsingLevel: 0` clamps to unrated-only | `expected 8 to be 4` |
| M3b `!= null` → truthiness (JOIN) | same | — |
| M4 bitwise → `<=` threshold | the clamp is bitwise, not a threshold | `expected 7 to be 5` |
| M5 `>=` → `>` | the boundary test | `expected [] to deeply equal [ 10 ]` |
| M6 delete the zero branch | ZERO-advertised not dropped | `expected [ 10 ] to deeply equal [ 10, 8 ]` |
| M7 zero branch returns `false` | same | same |
| M8 advertise the unclamped count | DROPS-below-floor test | `expected 34577 to be 33500` |
| M9 apply the floor to `mine` | mine: FAR below floor still returned | `expected [ 21 ] to deeply equal [ 21, 20 ]` |
| M10 `mine` loses its clamp | mine itemCount is clamped **+ the ledger** | ledger + count assertion |
| M11 filter after the walk | EVERY-row-dropped page advances the cursor | `expected 9 to be 6` |
| M12 public loses its clamp | 4 floor tests **+ the ledger** | ledger + floor assertions |
| M13 a new unpinned caller | no UNEXPECTED call site | ledger |

The harness's own positive control is the clean run before and after each mutant: 50/50 green, and each mutant produced a **distinct** failure set, so no stale-transform-cache effect.

### What I could not verify

- **No live/production check.** Everything above is unit-level. The numbers in the description (2,078/216, 2,080/19, the 87.5% containment rejection) are from the measurement that motivated the PR, not re-derived by me.
- **20% is a judgement, not a fit.** It cleanly separates the two live populations that were measured, but I did not sweep the threshold against the full catalog. If real short-page rates come in high, the knob to turn first is the `4x` over-fetch, then this.
- **`itemCount` is a semantic change for block authors** — no in-repo consumer reads it, so nothing here catches a block that depended on the unclamped number.

ZacxDev:
## 🔴 Cost measurement: the clamped count is ~31× the unclamped one on the real discovery page

I ran the performance probe this change was gated on, against a production-scale replica. **The logic is right; the cost is not affordable as written.** Numbers below are steady-state (two consecutive runs, cold-cache first run discarded), with the id-selection subquery removed so only the count is being timed.

Page-1 over-fetch window under the endpoint's default sort (`Most Followers`): **97 collections, 298,469 accepted items, largest 34,577, mean 3,077.**

| query | run 1 | run 2 |
|---|---|---|
| **before** — 1 unclamped count over the sliced page (24 ids) | **85 ms** | **82 ms** |
| after, arm a — unclamped, 97 ids | 84 ms | 78 ms |
| **after, arm b — CLAMPED, 97 ids** | **2829 ms** | **2563 ms** |
| variant — clamped over only 24 ids | 2076 ms | 2079 ms |

**Mechanism:** the unclamped count is an *Index Only Scan* on `CollectionItem_collectionId_status_covered` — no heap access. Adding `LEFT JOIN "Image"` forfeits the covering index and becomes a **Nested Loop Left Join** over ~300k rows, `Buffers: shared hit=1713835`.

### Two things this refutes

1. **Widening/narrowing `OVERFETCH` is not the knob.** The PR body flags the `4x` multiplier as the first thing to turn. Measured: clamping over 24 ids instead of 97 costs 2076 ms vs 2829 ms — **~27%, not the order of magnitude needed**. The cost is dominated by a handful of very large collections that are on page 1 *by definition*, since the sort that surfaces them is the one that ranks them.

2. **A "realistic-looking" probe gives the wrong answer here.** Measured over the 97 *newest* public Image collections the delta is only +39% (210 → 291 ms) — because new collections are small (mean ~20 items). The sort the endpoint actually uses returns collections **150× larger** on average. One measurement point was not a general claim; the two differ by an order of magnitude.

### Directions, none of them free

- **Denormalise `nsfwLevel` onto `CollectionItem`** — restores a covering index. Needs a migration, which in this deployment is applied by hand per environment, so it is an operator decision rather than a CI one.
- **Precompute a per-collection playable fraction** on a schedule; discovery then reads a column. Staleness is bounded and probably acceptable for a ranking filter.
- **Sample rather than count** — the floor only needs a fraction, not an exact number, so a bounded sample per collection would do.
- **Cache per (collection, ceiling)** — the ceiling space is tiny.

I have not implemented any of these; picking one is a design call. The correctness work here — the LEFT JOIN escape for non-image items, the discovery-only floor, the drop inside the pagination walk — stands and should survive whichever direction is chosen.

PR #4661, opened 2026-09-06
fix(search): re-index collections when a model or image is removed

ZacxDev:
## The defect

Collection search results render broken thumbnails for models and images that have been deleted.

`collections_v3` denormalizes images onto every collection document. Nothing about removing that media ever reached the index, so the document kept a snapshot of things that no longer exist.

**Why nothing reached it.** `prepareBatches` in `src/server/search-index/collections.search-index.ts` filters candidate collections on `c."createdAt" >= lastUpdatedAt`, so the incremental sweep only ever revisits **newly created** collections. An existing collection is rebuilt only when something enqueues it explicitly — and no removal path did. `model.service.ts` had exactly one `collectionsSearchIndex.queueUpdate` call before this PR (the featured-models path); `image.service.ts` had none.

`collection-index-sync.ts` does not cover this: it runs the opposite direction (collection membership → `images_v6.collectionIds`).

## An image reaches a collection document by seven routes

Only one of them is a `CollectionItem.imageId` row. Resolving by that column alone — the obvious reading, and what the first version of this PR did — misses six, including the two commonest. The collection that prompted this work has two items, both model-type with `imageId` NULL; its thumbnail arrives via `modelItemImage`, so the original fix resolved **zero** collections for it.

| # | Source | Route |
|---|---|---|
| 1 | `imageItemImage` CTE | `CollectionItem.imageId` → the image directly |
| 2 | `postItemImage` CTE | `CollectionItem.postId` → that post's **first** image |
| 3 | `modelItemImage` CTE | `CollectionItem.modelId` → image → Post → ModelVersion → Model |
| 4 | `articleItemImage` CTE | `CollectionItem.articleId` → `Article.coverId` |
| 5 | `model3dItemImage` CTE | `CollectionItem.model3dId` → `Model3D.thumbnailImageId` |
| 6 | the document's cover field | `Collection.imageId` |
| 7 | a separate `findMany` in `pullData` | `User.profilePictureId` → `Collection.userId` — **partial, see below** |

Route 6 matters most on screen — `CollectionCard` renders `if (data.image) return [data.image]`, so a cover wins over every item image. Route 7 is invisible to anyone enumerating from the CTE list, which is how it was missed twice: profile pictures are fetched by their own `db.image.findMany` and shipped as `user.profilePicture` on **every** collection document. The code comment now says to enumerate from `pullData` and `transformData`, not from the CTEs.

⚠️ **Route 7 is only partly covered, deliberately.** The leg matches `u."profilePictureId" IN (ids)` — images that are someone's avatar *right now*. The usual way an avatar image dies is the deferred reaper, and that path **cannot** match: `remove-replaced-images` calls `getStillReferencedImageIds`, which filters out every id that is still a `profilePictureId`, before calling `deleteImages`. So for a *replaced* avatar this leg resolves zero collections, by construction, every time. `deleteUser` reaches the same outcome by a different mechanism — it nulls `profilePictureId` inside its own transaction, so by the time that user's images are destroyed (later, by `remove-deleted-user-images`) nothing matches this leg either. It does **not** delete the user's collections; `removeAllContent` is the function that does that.

The leg is kept because it is correct for what it does match — a moderator or ingestion delete of a live avatar — and its cost scales with the number of bound ids — measured 9.47 at one, 18.94 at two, 28.41 at three — so at the 100-id batch `deleteImages` uses it is ~950, not 9.47. Bounded per id because `profilePictureId` is `@unique`. **The remaining gap is real and is not closed here:** a user changes their avatar, the old `Image` row and its object are destroyed 30 days later, and every collection they own still ships `user.profilePicture` pointing at a dead url. Closing it needs an enqueue at *replacement* time, which is a different trigger from removal. Tracked separately.

The joins for routes 2–5 mirror the CTEs' own conditions, including `m."userId" = p."userId"`, so this and the index agree on which image a collection would show.

## What this changes

A leaf module, `src/server/services/collection-media-index.ts`, resolving the affected collections and queueing them for rebuild, wired into the image-delete paths and the permanent model delete.

**`Update`, not `Delete`.** The collection still exists; only its document needs rebuilding. A `Delete` would evict a real collection.

**Resolve before the delete.** `CollectionItem.modelId`/`.postId`/`.imageId` are all `onDelete: Cascade`, so once the row is gone there is nothing left to read. For `permaDeleteModelById` the resolve is deliberately **outside** the transaction: a failed statement aborts a Postgres transaction outright, so a bookkeeping read placed inside could take the whole delete down with it. It gains a `postId` leg because that path hard-deletes Posts in the same transaction.

**Non-throwing by contract.** Every exported function logs and returns rather than throwing — these run around deletes that have already committed or are about to, so a failure must cost a stale document, never the user's delete.

**Volume.** Ids are deduped, the enqueue is chunked at 500 per queue write, and the resolve caps at 10,000 distinct collections. The cap applies to distinct *collections*, not membership rows — hence raw `SELECT DISTINCT … LIMIT` rather than Prisma's `distinct`, which has historically been applied after rows come back. Truncation is logged (`collection-media-index-enqueue-truncated`) **without a figure**: every lookup stops at `LIMIT cap + 1`, so any number quoted would be `1` however large the real overflow, and real models sit far past the cap (one measured in **86,156** distinct collections).

## Plans

`EXPLAIN` (no `ANALYZE`), read replica, `pg_is_in_recovery()` confirmed.

🔴 **The image-resolve cost scales with the number of bound ids, so each row below says how many it was measured at.** `deleteImages` batches **100** ids (`Limiter({ batchSize: 100 })`, concurrency 5), so the 100-id row is the one that describes the bulk path; the 1-id row describes a single-image delete.

| Query | bound ids | Cost | Notes |
|---|---|---|---|
| **image resolve as it ships today** (gate false, six legs) | 1 | **9,924** | every leg an index scan |
| **image resolve as it ships today** (gate false, six legs) | **100** | **2,017,779** | the bulk path — 203× the 1-id figure |
| image resolve with the cover leg on, index absent | 1 | 461,456 | 46.5× the 1-id six-leg figure |
| image resolve with the cover leg on, index absent | **100** | **2,487,317** | **+23%** over six legs, not 46.5× |
| model cascade, three legs | n/a (one model) | 89.68 | vs **3,336,745** for the `OR` form it replaced |
| cover-index gate | n/a (catalog) | 5.01 | two index scans on the catalogs |

**What the gate is and is not worth.** At one bound id the cover leg dominates (46.5×), which is where the "47×" figure in earlier revisions of this PR came from. At the batch size the code actually uses, the marginal cost of the cover leg is **+23%** — the other six legs have grown to dominate it. The gate is still right, but for a reason the ratio does not carry: what it prevents is a **parallel sequential scan of the 2,817 MB `Collection` heap**, up to five concurrently, on the primary, potentially while a `CREATE INDEX CONCURRENTLY` is building on that same table. Weigh it on that, not on a multiplier.

The cascade query originally bridged its columns with a single `OR`, which is not sargable against a ~208M-row table: Postgres used none of the indexes and scanned an unrelated one end to end (43.3M estimated rows) with the whole condition as a post-scan `Filter`. Split into `UNION`ed legs, every branch is an index scan. Verified to survive parameterisation — the shape Prisma sends — at cost 74.52 for the two-leg form.

Five legs are index probes — four single (`CollectionItem_imageId_lookup`, `_postId_lookup`, `_modelId`, `_model3dId_idx`) plus the avatar leg, which is two probes joined (`User_profilePictureId_key` then `Collection_userId_idx`; per-id cost above). The **article leg is not**: `CollectionItem_article_idx` is `("collectionId","articleId")` and nothing leads on `articleId`, so it scans that partial index: row estimate **271** per outer row, against **6** for the `postId` leg. (An earlier revision of this line said "2,799 rows" — that was the leg's *cost* span, not a row count.) It is bounded — the index is 7,712 kB — so it stays rather than getting one of its own.

## The cover leg is gated, and ships with a migration

`Collection.imageId` has **no index**. Without one the leg is a parallel sequential scan of the ~17.3M-row / **2,817 MB** heap. So it is gated on the index being present *and usable*, and ships with `20260906190000_collection_imageid_idx`.

The gate checks `indisvalid AND indisready`, joined through `pg_index` and schema-qualified — not merely that the name exists in `pg_class`:

> A `CREATE INDEX CONCURRENTLY` row appears in `pg_class` from the **start** of the build and stays there permanently if the build **fails**. A name-only check therefore returns true throughout a multi-minute build, and forever after a failure — switching the leg on in exactly the window the gate exists to cover, so every image delete would pay the seq scan concurrently with the build, on the primary, up to five at a time (`deleteImages` uses `Limiter({ batchSize: 100 })`, default concurrency 5).

The migration deliberately has **no `IF NOT EXISTS`**: a failed `CONCURRENTLY` build leaves an INVALID index occupying the name, and `IF NOT EXISTS` would turn the retry into a silent no-op that prints `CREATE INDEX` and repairs nothing. A `DROP INDEX CONCURRENTLY` recovery step is documented in the migration.

**Not verified:** the post-index plan for the cover leg. `hypopg` is available but not installed, and `CREATE EXTENSION`/`CREATE INDEX` are writes — impossible on a replica and not something to do to a primary for a test. The seq-scan plan is measured; the indexed plan is predicted.

## Soft delete and unpublish: deliberately no enqueue

Earlier revisions queued a rebuild on `deleteModelById` and `unpublishModelById`. Both were **provably no-ops**: the index's image CTEs filter only on `i."ingestion" = 'Scanned' AND i."needsReview" IS NULL` — no `Model.status`, no `Model.deletedAt`, no `Post.publishedAt` — so a soft delete or unpublish leaves the Model, Post and Image rows intact and the rebuild re-emits the same image. Verified in production: a model with `status=Deleted` and `deletedAt` set is an ACCEPTED item in 41 collections and its `modelItemImage` join still returns live images.

They have been **removed**. The cost was up to 10,000 Meilisearch enqueues per soft delete for no change in output, and the justification that kept them ("correct once the image is genuinely gone") is served elsewhere — when the image is genuinely removed, the image-delete paths fire their own enqueue.

**Open question for a reviewer:** if the product answer is *"hide a soft-deleted model's images from collection cards"*, the CTE predicate and this enqueue are needed **together** — neither alone does anything. Adding a status predicate to a shared index CTE changes collection-card semantics for every soft-deleted model on the site, which is a product decision, not one to make inside this PR. A test pins the absence and records this.

## Corrections to claims this code was making

- **The indexes are btree, not hash.** Every index on `CollectionItem.modelId`/`.imageId` is `USING btree`; the Prisma schema annotates them `type: Hash`. The deployed ones are what the planner uses.
- **`Collection.imageId` has no foreign key.** `Collection` carries exactly one (`Collection_userId_fkey`), so the schema's `onDelete: SetNull` is not a guarantee the database makes — a cover can and does point at a deleted `Image`. No behavioural consequence here (the resolve runs before the delete either way), but the premise was wrong.
- **`truncated`, not a count** — see Volume above.

## Test matrix

34 tests. Every guard was watched failing on the behaviour it pins, and eleven mutants across three rounds were each killed by their designated guard:

| Mutation | Killed by |
|---|---|
| cascade `OR` form restored | `splits the columns into UNIONed legs…` |
| cascade without the `postId` leg | `resolves the posts the cascade will delete…` |
| image resolve reduced to the `imageId` leg | `resolves every route by which an image reaches…` |
| the avatar leg dropped | `resolves the owner-avatar route, which is not one of the index CTEs` |
| gate reverted to a name-only `pg_class` check | `asks whether the index is VALID and READY, not merely named` |
| cover probe moved back inside the main try | `still resolves the other six legs when the catalog probe fails` |
| cover leg run ungated | `omits the cover leg and warns when the index is not usable` |
| soft-delete enqueue re-added | `does not queue a collections rebuild, which would be a no-op` |
| raw `Error` instead of `safeError` | `never throws when the queue write fails…` |
| a number quoted for the overflow | `does not quote a number it cannot know for the overflow` |
| image-leg failure discarding results | `still resolves the other six legs…` |

🔴 **One mutant survived on first run** and is worth recording: the `safeError` guard asserted `objectContaining({ message })`, which a live `Error` also satisfies, so it could never fail. It now asserts after a JSON round-trip — the transformation that actually decides whether the cause survives. Two other bugs were in the test harness rather than the code: `sqlOf` joined only the template strings, rendering an interpolated `Prisma.sql` fragment as `?`; and an assertion used `vi.spyOn` on an intra-module call, which cannot be intercepted.

Gates: `pnpm typecheck` 0 errors; `pnpm test:lint-rules` 476 passed; `src/server/services/__tests__/` + the Flipt ledger 394 files / 5598 passed.

## Known and tracked, not fixed here

A collection whose **last** item is hard-deleted stops matching the index's `EXISTS` clause, so `pullData` returns nothing and `pushData` is upsert-only — the stale document survives permanently. Real and not rare — roughly one collection in five holds exactly one item (two `TABLESAMPLE` runs put it at 19.2% and 22.0%; the exact share does not matter here) — but it needs a `deleteDocs` path and is a defect in the index sync's push semantics rather than in these enqueues. Tracked separately.

Backfilling already-stale documents is also out of scope; this stops new ones being created.

### DIFF
--- a/packages/civitai-db-schema/prisma/migrations/20260906190000_collection_imageid_idx/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260906190000_collection_imageid_idx/migration.sql
@@ -2,17 +2,28 @@
 --
 -- The collections search index denormalizes a cover image onto every collection document, and
 -- CollectionCard renders `if (data.image) return [data.image]` — the cover wins over item images.
--- `Collection.imageId -> Image` is `onDelete: SetNull`, so deleting a cover image leaves the
--- document holding a dead thumbnail with no CollectionItem row pointing at it. The reindex enqueue
--- therefore has to look the collection up BY that column.
+-- Deleting a cover image otherwise leaves the document holding a dead thumbnail with no
+-- CollectionItem row pointing at it, so the reindex enqueue has to look the collection up BY that
+-- column. (`Collection.imageId` is not a foreign key in the deployed database, so nothing clears
+-- it on delete either.)
 --
--- Without this index that lookup is a parallel sequential scan of a ~17.3M-row / 4.9 GB table
--- (measured cost 451,531, against ~10,000 for every other leg of the same query combined), on a
--- user-facing image-delete path. With it the leg is an index probe.
+-- Without this index that lookup is a parallel sequential scan of the ~17.3M-row / 2,817 MB heap:
+-- measured, the seven-leg resolve costs 461,408 with the leg enabled and unindexed, against 9,876
+-- with it disabled. That is on a user-facing image-delete path which `deleteImages` runs at up to
+-- five concurrent batches.
 --
--- 🔴 APPLY THIS BEFORE DEPLOYING the collections-reindex change that reads it. Migrations here are
--- applied by hand, so the ordering is a human step, not something the deploy enforces.
+-- 🔴 NO `IF NOT EXISTS`, DELIBERATELY. `CREATE INDEX CONCURRENTLY` leaves an INVALID index behind
+-- when it fails, and that index still occupies the name — so `IF NOT EXISTS` would turn the retry
+-- into a silent no-op that prints `CREATE INDEX` and repairs nothing. Without it the retry errors
+-- loudly and you take the recovery below. The application gate reads `indisvalid AND indisready`
+-- rather than mere existence, so a half-built or failed index does NOT switch the cover leg on.
 --
--- CONCURRENTLY cannot run inside a transaction block; run this statement on its own.
-CREATE INDEX CONCURRENTLY IF NOT EXISTS "Collection_imageId_idx" ON "Collection" ("imageId")
+-- Recovery if this fails partway (check: `SELECT indisvalid, indisready FROM pg_index
+-- WHERE indexrelid = '"Collection_imageId_idx"'::regclass;` — anything but `t | t` is unusable):
+--
+--   DROP INDEX CONCURRENTLY "Collection_imageId_idx";
+--   -- then re-run the CREATE below
+--
+-- CONCURRENTLY cannot run inside a transaction block; run each statement on its own.
+CREATE INDEX CONCURRENTLY "Collection_imageId_idx" ON "Collection" ("imageId")
   WHERE "imageId" IS NOT NULL;
--- a/packages/civitai-db-schema/prisma/migrations/20260906190000_collection_imageid_idx/migration.sql
+++ b/packages/civitai-db-schema/prisma/migrations/20260906190000_collection_imageid_idx/migration.sql
@@ -1,0 +1,18 @@
+-- Index `Collection."imageId"` so a deleted image can find the collections whose COVER it is.
+--
+-- The collections search index denormalizes a cover image onto every collection document, and
+-- CollectionCard renders `if (data.image) return [data.image]` — the cover wins over item images.
+-- `Collection.imageId -> Image` is `onDelete: SetNull`, so deleting a cover image leaves the
+-- document holding a dead thumbnail with no CollectionItem row pointing at it. The reindex enqueue
+-- therefore has to look the collection up BY that column.
+--
+-- Without this index that lookup is a parallel sequential scan of a ~17.3M-row / 4.9 GB table
+-- (measured cost 451,531, against ~10,000 for every other leg of the same query combined), on a
+-- user-facing image-delete path. With it the leg is an index probe.
+--
+-- 🔴 APPLY THIS BEFORE DEPLOYING the collections-reindex change that reads it. Migrations here are
+-- applied by hand, so the ordering is a human step, not something the deploy enforces.
+--
+-- CONCURRENTLY cannot run inside a transaction block; run this statement on its own.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "Collection_imageId_idx" ON "Collection" ("imageId")
+  WHERE "imageId" IS NOT NULL;
--- a/src/server/flipt/__tests__/flipt-eval-context.test.ts
+++ b/src/server/flipt/__tests__/flipt-eval-context.test.ts
@@ -152,14 +152,14 @@ const ENTITY_WITHOUT_CONTEXT_LEDGER: Record<string, string> = {
   'server/services/article-rating-review.helpers.ts:482':
     'article-rating-dispute has no segment rollouts; background path with no SessionUser',
   // flag `feed-fetch-filter-in-post`: enabled=true, 0 rules, 0 rollouts.
-  'server/services/image.service.ts:3242':
+  'server/services/image.service.ts:3267':
     'feed-fetch-filter-in-post has no segment rollouts; hot feed path',
   // flag `feed-image-existence`: enabled=true, 0 rules, 0 rollouts. Three sites.
-  'server/services/image.service.ts:3286':
+  'server/services/image.service.ts:3311':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:4055':
+  'server/services/image.service.ts:4080':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:4865':
+  'server/services/image.service.ts:4890':
     'feed-image-existence has no segment rollouts; hot feed path',
   // flags `model-text-moderation-xguard` / `-apply`: both enabled=true, 0 rules,
   // 0 rollouts (checked against flipt-state and against the evaluation API on
@@ -219,7 +219,7 @@ describe('flipt evaluation context — source gate', () => {
     // these and fail the second, and vice versa.
     const bySite = new Map(calls.map((c) => [c.site, c]));
     expect(bySite.get('server/services/feedback.service.ts:43')?.argc).toBe(3);
-    expect(bySite.get('server/services/image.service.ts:3286')?.argc).toBe(2);
+    expect(bySite.get('server/services/image.service.ts:3311')?.argc).toBe(2);
   });
 
   it('adds no Flipt evaluation that names an entity but passes no context', () => {
--- a/src/server/services/__tests__/collection-media-index.test.ts
+++ b/src/server/services/__tests__/collection-media-index.test.ts
@@ -139,6 +139,30 @@ describe('getCollectionIdsForModelCascade', () => {
       dropped: 0,
     });
   });
+
+  // `CollectionItem` is a ~208M-row table. Bridging the two columns with a single
+  // `OR` makes the predicate non-sargable: Postgres can use NEITHER the modelId nor
+  // the imageId index, and falls back to scanning an unrelated index end to end with
+  // the whole condition as a post-scan Filter (measured: 43.3M estimated rows, total
+  // cost 3,336,745). Split into two UNIONed legs, each probes its own index
+  // (measured: cost 73.76, both legs index-only scans; 74.52 for the parameterised
+  // generic plan, which is the shape Prisma sends).
+  //
+  // The guard is "no OR anywhere in this statement" rather than a spelling check on
+  // the good form: this query has no other legitimate use for one, so any OR that
+  // appears here is the non-sargable shape coming back.
+  it('splits the two columns into UNIONed legs so each can use its own index', async () => {
+    dbMock.dbWrite.$queryRaw.mockResolvedValueOnce(rows(COLLECTION_A));
+
+    await getCollectionIdsForModelCascade({ modelId: MODEL_ID });
+
+    const sql = sqlOf(dbMock.dbWrite.$queryRaw.mock.calls[0]);
+    expect(sql).toMatch(/\bUNION\b/);
+    // Two independent scans of the table, one per column.
+    expect(sql.match(/FROM "CollectionItem"/g)).toHaveLength(2);
+    // The non-sargable bridge must not come back.
+    expect(sql).not.toMatch(/\bOR\b/);
+  });
 });
 
 describe('enqueueCollectionRebuild', () => {
@@ -203,6 +227,25 @@ describe('enqueueCollectionRebuild', () => {
         cap: 3,
       })
     );
+  });
+
+  // Every lookup here stops at `LIMIT cap + 1`, so the count that comes back is only
+  // ever "at least one more than the cap" — the true overflow is never observed and
+  // cannot be. A message asserting an exact figure would be a claim the query was
+  // never in a position to make, so the wording is pinned as a lower bound.
+  it('words the truncation warning as a lower bound, not an exact count', async () => {
+    await enqueueCollectionRebuild({
+      collectionIds: [COLLECTION_B],
+      dropped: 4,
+      source: 'model-perma-delete',
+      cap: 2,
+    });
+
+    const warning = loggingMock.logToAxiom.mock.calls
+      .map((c) => c[0] as { name?: string; message?: string })
+      .find((a) => a?.name === 'collection-media-index-enqueue-truncated');
+
+    expect(warning?.message).toContain('at least');
   });
 
   it('never throws when the queue write fails, so post-delete cleanup still runs', async () => {
--- a/src/server/services/__tests__/collection-media-index.test.ts
+++ b/src/server/services/__tests__/collection-media-index.test.ts
@@ -1,8 +1,8 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
 /**
- * Unit coverage for the collections-reindex enqueue used by every model/image
- * removal path.
+ * Unit coverage for the collections-reindex enqueue used by every model/image removal
+ * path.
  *
  * The defect it exists for: `prepareBatches` in collections.search-index.ts filters on
  * `c."createdAt" >= lastUpdatedAt`, so the incremental sweep only revisits NEWLY
@@ -20,6 +20,7 @@ const { mockCollectionsQueueUpdate } = vi.hoisted(() => ({
 }));
 
 vi.mock('~/server/search-index', () => ({
+  articlesSearchIndex: { queueUpdate: vi.fn() },
   collectionsSearchIndex: { queueUpdate: mockCollectionsQueueUpdate },
   imagesMetricsSearchIndex: { queueUpdate: vi.fn() },
   imagesSearchIndex: { queueUpdate: vi.fn() },
@@ -27,6 +28,7 @@ vi.mock('~/server/search-index', () => ({
 }));
 
 import {
+  COVER_INDEX_NAME,
   enqueueCollectionRebuild,
   getCollectionIdsForMedia,
   getCollectionIdsForModelCascade,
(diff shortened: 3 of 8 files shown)


ZacxDev:
**Round 1 adversarial audit — complete.** No 🔴 deploy-blockers. The audit reproduced the red/green matrix exactly, independently re-ran every mutation claim in the PR body, added four mutants of its own (each killed by its own guard), and reproduced the `EXPLAIN` figures for the `OR` → UNION fix. It found the change narrower than the PR claimed; round 2's fixes are below.

Deliberately **not** fixed here and tracked separately: a collection whose *last* item is hard-deleted stops matching the index's `EXISTS` clause, so the upsert-only push path never rewrites or removes its document. That is a defect in the index sync's push semantics rather than in these enqueues.

```audit-claims round=1 audited=bbd13178ed..d5262294
1. Image removal now resolves collections that show the image indirectly, via all five item CTEs (imageId, postId, modelId, articleId, model3dId) plus the collection cover, not only CollectionItem.imageId.
2. permaDeleteModelById's cascade resolver gained the postId leg, for the Posts it hard-deletes whose CollectionItem.postId cascades.
3. The cover-image leg (Collection.imageId) is added but gated at runtime on the existence of an index on that column, and ships migration 20260906190000_collection_imageid_idx (CREATE INDEX CONCURRENTLY) to create it; when the index is absent the leg is skipped and logged.
4. The comments at model.service.ts soft-delete and unpublish no longer claim the rebuild removes a deleted/unpublished model's thumbnail; the wiring is kept and the underlying product question is raised in the PR body. The index CTEs were not modified.
5. logFailure now routes the error through safeError, so the cause survives JSON serialisation instead of stringifying to {}.
6. The truncation signal changed from a `dropped` count (structurally always 1) to a `truncated` boolean, and the operator-facing warning no longer quotes a figure.
7. Six nits fixed: case-insensitive OR guard, a try/catch and a failure test for deleteImages, the unwrapped await in permaDeleteModelById, the image-leg catch no longer discarding model-leg ids, and articlesSearchIndex added to the search-index mock factory.
8. Two PR-body inaccuracies corrected: the retracted "hash index" sentence, and the mutation table's mislabelling of the cap test as the truncation test.
```

ZacxDev:
**Round 2 delta re-audit — complete.** Range `bbd13178ed..d52622942b`. One 🔴 and five smaller findings; all fixed in round 3.

The 🔴 was worth the round: the cover-leg gate matched on `pg_class.relname` alone, and a `CREATE INDEX CONCURRENTLY` index is present in `pg_class` from the *start* of the build — and stays there permanently if the build fails. So in the exact deploy-then-migrate ordering the gate exists to protect, it flipped to `true` when the build began, switching the cover leg on to run a sequential scan concurrently with the index build. The auditor reproduced all three states (mid-build, post-failure, retry) on a throwaway PostgreSQL instance rather than arguing it.

The audit also re-derived every cost figure, row count and percentage the PR asserts, against the then-current head.

```audit-claims round=2 audited=d52622942b..3c359e00
1. The cover-index gate now joins pg_index and requires indisvalid AND indisready, schema-qualified to public, so a half-built or failed CONCURRENTLY index no longer switches the cover leg on.
2. The migration dropped IF NOT EXISTS, so a retry over a failed build errors loudly instead of reporting success and repairing nothing; a DROP INDEX CONCURRENTLY recovery step and an indisvalid/indisready check query are documented alongside it.
3. The catalog probe now has its own catch setting withCover = false, so a failing probe no longer discards the six legs that do not depend on the cover index.
4. A seventh route was found and covered: user.profilePicture, which reaches the document via a separate db.image.findMany rather than a CTE. The enumeration comment now says to derive routes from pullData/transformData rather than from the CTEs, naming the method error that hid routes 6 and 7.
5. The new perma-delete catch in model.service.ts routes its error through safeError.
6. The soft-delete and unpublish enqueue call sites were REMOVED rather than re-justified, along with queueCollectionsForMedia and the model-id leg of the image resolver; a test pins their absence and records that hiding soft-deleted art would need the CTE predicate and the enqueue together.
7. Three premises corrected: Collection.imageId has no foreign key in the deployed database; the article leg is an index scan, not a probe; the enqueue-failure assertion now asserts across a JSON round-trip so it cannot pass without safeError.
8. The PR body was rewritten rather than appended to, clearing round-1 statements that later sections contradicted; the table figure is now 2,817 MB heap (not 4.9 GB total relation size) and the single-item share is 22.0% from a TABLESAMPLE rather than 24.6% from a heap-order LIMIT.
```

ZacxDev:
**Round 3 delta re-audit — complete.** Range `d52622942b..3c359e0082`. Every claim round 3 made verified true; all four mutation-testable ones died to their own designated guard. The 🔴 gate fix from round 3 is confirmed correct — no realistic fail-closed path, `dbWrite` confirmed as the right connection (gate and legs hit the same node, so no replica/primary catalog split), probe cost 5.01 reproduced. The auditor independently re-enumerated `pullData`/`transformData` and found **no eighth Image-row route**, and read `scripts/local-dev/run_migrations.ts` to confirm dropping `IF NOT EXISTS` behaves as intended (a throw is recorded as `failedMigrations` without being marked applied, so a retry re-attempts).

Four findings, none blocking — one scope claim and three pieces of drift left by the previous round's own fixes. All fixed in round 4.

Filed separately rather than fixed here: replacing an avatar leaves a dead `user.profilePicture` on every collection that user owns. The leg added in round 3 covers deletion of a *live* avatar and cannot fire on the replacement path, because `remove-replaced-images.ts` filters out every id still referenced as a `profilePictureId` before calling `deleteImages`. Fixing it needs an enqueue at replacement time — a different trigger.

```audit-claims round=3 audited=3c359e0082..de31326dfc
1. The avatar leg is kept but its claim is corrected: the code comment and PR body now state it covers deletion of a live avatar only, and name the remaining replacement-path gap as needing an enqueue at replacement time, out of this PR's scope.
2. The vacuous soft-delete guard was deleted rather than re-pointed: re-pointing it at permaDeleteModelById was mutation-tested and the mutant SURVIVED, because enqueueCollectionRebuild is non-throwing and catches internally so the outer catch is unreachable through the real callee. The call site now records that the catch is defence-in-depth with no coverage, and what exercising it would cost.
3. The model-removal test file's docstring now states that the soft-delete and perma-delete paths want opposite things and why, and that unpublishModelById is no longer exercised.
4. The units error is corrected in both the code comment and the PR body: 271 vs 6 are the row estimates, 2,799 was the article leg's cost span; the index size is independently confirmed at 7,712 kB and the retracted figure is named so a reader who saw it knows it was withdrawn.
5. One sentence added to the gate describe recording that its assertions are spelled rather than structural, and why a unit test cannot do better with $queryRaw behind a mock. No speculative structural guard was added.
6. All five round-3 mutants still die to their designated guards after these edits; the file is at 34 tests (one deleted, none added).
```

ZacxDev:
**Round 4 delta re-audit — complete**, and **the audit ladder stops here.** Range `3c359e0082..de31326dfc`. Two findings, both prose, neither blocking; both fixed, plus one more found while sweeping.

The round-4 audit independently reproduced the `permaDeleteModelById` mutation result (the mutant survives the re-pointed test, so deleting rather than re-pointing was correct), verified the `remove-replaced-images` → `getStillReferencedImageIds` mechanism rather than trusting the summary, and re-measured the article index at exactly 7,712 kB.

## Why the ladder is stopping deliberately rather than on a clean round

Stating this explicitly, because a report that ends on this criterion otherwise looks identical to one that converged, and those mean opposite things.

Neither normal stop condition can fire here:

- **The findings rule** (continue while rounds find things) would run indefinitely — rounds 3, 4 and 5 each found real defects, and every one was a sentence a previous round wrote while fixing the round before it.
- **The attribution gate** (two consecutive rounds changing zero payload lines) cannot fire, because comments in a shipped source file count as payload by the letter. Round 4's 27 payload lines were **100% comments**; round 5's and this one's are too. The payload is now prose, so the gate is structurally inert.

That combination permits stopping only once the recurring **shape** is swept at every site rather than at the one reported. The shape: *a comment or PR-body sentence that names a specific function, job, index or table and asserts something untrue about it.* Five instances across three rounds — "seven routes", "five legs are index probes", "2,799 rows per outer row", "still runs the trailing cleanup", "`deleteUser` deletes the user's collections".

The sweep is done: every claim in `collection-media-index.ts`, the touched blocks of `model.service.ts`, the migration comment, the test docstrings and this PR body was checked against the current head. It caught two beyond the reported ones — `permaDeleteModelById`'s cascade note omitted `Post` (the transaction destroys all three), and the avatar cost figure was quoted without the input size it was measured at.

**No 🔴 remains, and the blast radius of anything left is "a comment could be clearer".**

## Deliberately open, tracked, not missed

1. **A collection whose last item is hard-deleted keeps its document forever.** It stops matching the index's `EXISTS` clause, and the push path is upsert-only, so nothing rewrites or removes it. Defect in the sync's push semantics, not in these enqueues.
2. **Replacing an avatar leaves a dead `user.profilePicture`** on every collection that user owns. The leg here covers deletion of a *live* avatar and cannot fire on the replacement path. Needs an enqueue at replacement time.
3. **No backfill.** This stops new stale documents; roughly 850 collections already carry one.
4. **The post-index plan for the cover leg is predicted, not measured** — `hypopg` is not installed and `CREATE INDEX` cannot run on a replica. The seq-scan plan it avoids *is* measured.

## 🔴 Before deploying

**Apply `20260906190000_collection_imageid_idx` by hand first.** Migrations are not auto-applied here. The runtime gate makes deploy-then-migrate *safe* (the cover leg stays off until the index is valid and ready), but applying it first is what makes the leg actually do anything.

```audit-claims round=4 audited=de31326dfc..f37baf72
1. The deleteUser claim is corrected: it does NOT delete the user's collections — removeAllContent does, and it is a separate moderator endpoint — and deleteUser reaches the same outcome by nulling profilePictureId inside its own transaction.
2. permaDeleteModelById's cascade note now says CollectionItem cascades from Model, Post AND Image, all three of which that transaction destroys; it previously named only Model and Image.
3. The PR body's test count is corrected from 35 to 34, matching the file after round 4's deletion.
4. Every named-entity and measured claim in the module, the touched service blocks, the migration comment, the test docstrings and the PR body was swept against the current head; the overflow example is re-measured at 86,156 collections.
5. The avatar leg's cost is no longer quoted bare: it scales with bound ids (9.47 at one, 18.94 at two, 28.41 at three), so the note gives the ~950 figure for the 100-id batch deleteImages actually uses, and the duplicate in the function docstring now refers back rather than restating.
```

ZacxDev:
**Round 5 delta re-audit — complete.** Range `de31326dfc..f37baf7216` (the two comments-only commits). **The commits themselves audited clean** — every claim in their diff was checked by reading the named entity or re-measuring the number, and all of them hold: the `deleteUser`/`removeAllContent` correction (both halves verified, including that `deleteUser` nulls `profilePictureId` inside its `$transaction`), the `Model`/`Post`/`Image` cascade note (verified against live `pg_constraint`, not just the Prisma schema), the 34-test count (counted by enumeration), and the avatar cost scaling (reproduced at 9.47 / 18.94 / 28.41, and 946.95 at 100 real image ids — so the ~950 extrapolation is exactly linear).

The auditor confirmed the range is comments-only with a positive control: the non-comment filter returns 0 here and **7** on `de31326dfc~1..de31326dfc` through the identical pipeline, so the zero is measured rather than wired to nothing.

**One finding, in this description rather than in the code — now fixed.** The sweep commit verified every plan cost on the wrong axis: it re-ran each at whatever id count it was originally measured at and confirmed no drift, which preserved in the Plans table the exact defect its sibling commit fixed for the avatar leg. `9,876` was a **one-bound-id** measurement, stated with no id count.

Re-measured on the replica (PG 18.3, `pg_is_in_recovery()` true, no `ANALYZE`), with 100 real recent image ids:

| bound ids | six legs (ships) | seven legs (cover on, index absent) | ratio |
|---|---|---|---|
| 1 | 9,924 | 461,456 | 46.5× |
| **100** | **2,017,779** | **2,487,317** | **1.23×** |

`deleteImages` batches 100 ids at concurrency 5, so the bulk path costs **203×** the figure the table quoted. And the cover leg's marginal cost there is **+23%**, not 46.5×.

That does not weaken the gate — what it prevents is a parallel sequential scan of the 2,817 MB `Collection` heap, up to five concurrently, on the primary, potentially during a `CREATE INDEX CONCURRENTLY` on that same table. But the multiplier a reviewer would have weighed it on was off by ~38×. The Plans table now states the bound-id count per row and gives both figures.

**This closes the audit ladder.** Six rounds: two found code defects (a non-sargable `OR` costing a 43M-row scan; a gate that failed open for the duration of an index build), four found claims — and every one of those four was a sentence a previous round wrote while fixing the round before it. No round ever returned zero findings; rounds 3–6 were stopped on the prose-payload criterion recorded above, and this last one needed no code change at all.

Open and tracked, not missed: the single-item-collection document leak, the avatar-**replacement** path, no backfill of the ~850 already-stale documents, and the cover leg's post-index plan (predicted; the seq scan it avoids is measured).

🔴 **Apply `20260906190000_collection_imageid_idx` by hand before deploying.** Migrations are not auto-applied here; the gate makes the ordering safe, but applying it is what makes the cover leg do anything.

PR #4662, opened 2026-09-06
Moderation-only retraction of the shared stored object on a blocked-image takedown

ZacxDev:
## What this does

The image cache service can delete a **shared, content-addressed stored object**. That object is shared by every byte-identical image across **all** owners, so deleting it removes the full-resolution original for every one of them — irreversibly, and without any way to enumerate or notify them. Since the cache service's current release that deletion happens only when the request carries `retractPublicBlobs=true`, which defaults off and which nothing sent.

This PR makes the application send it, from **an affirmative per-image human takedown only**.

> ⚠️ **This description was rewritten after round 2 inverted the design.** An earlier version of this body described a *negative* discriminator (retract unless a row carried an account-deletion marker) and listed four writers as retracting that do not. If you read this PR before that rewrite, re-read this section — the mechanism and the writer table below both changed.

### The discriminator is positive, and server-only

`removeBlockedImages` retracts for an image only where a **`ModActivity`** row exists with:

- `activity ∈ ('review', 'bulkRemove')` (`MODERATOR_TAKEDOWN_ACTIVITIES`), and
- `entityType = 'image'`, and
- `createdAt >= the queue row's createdAt`.

`Image.metadata` no longer participates in the decision at all. `ModActivity` has three writers — `trackModActivity`, the moderator app's `recordModActivity`, and the auth hub's impersonate-only copy — and **no client path reaches it in any polarity**: every writer of the `image` + `review`/`bulkRemove` combination sits behind `moderatorProcedure`, a `ModEndpoint`, a `WebhookEndpoint`, or requires an explicit `moderatorId`.

The timestamp comparison is load-bearing, not decoration: three of the six `review` writers are the *un-block* direction, so without it a moderator who once approved an image would license retraction on a later automated block.

**Why positive:** the first design was negative, and every writer that did not explicitly opt out retracted by default. That was reachable from ordinary self-service account deletion, from an ML classifier's false positive, from a whole-library ban — and, most sharply, from **any logged-in user filing a CSAM report on someone else's image** (`report.create` is `guardedProcedureAllowUnverifiedEmail`). Inverting the polarity closes that class rather than enumerating its members.

## What reaches the retracting call

Re-derived over every `ingestion='Blocked'` write in `src`, `apps` and `packages`, including the writes the DB trigger `trg_blocked_image_delete_queue` picks up.

| Writer | Per-image `ModActivity` | Retracts |
|---|---|---|
| `handleBlockImages` with an id list + `moderatorId` | `review` (+ `bulkRemove` from the mod app) | **yes** |
| `blockImage` (`apps/moderator`) | `review` | **yes** |
| `setTosViolationHandler` | `review` — added in this PR | **yes** |
| `handleBlockImages` with `userId` and no ids | none (only a `user`/`removeContent` row) | no |
| `toggleBan` remove-all-media | none | no |
| `softDeleteUser` (CSAM, whole library) | none | no |
| `report.service` CSAM branch | none | no |
| `image-scan-result.service` × 3 outcomes | none | no |
| `api/webhooks/image-scan-result` rule engine + prompt audit | none | no |
| `new-order.service` (Knights game) | none — calls `handleBlockImages` with no `moderatorId` | no |
| `remove-deleted-user-images` grace block | none | no |
| **any future writer** | none by default | **no** |

`setTosViolationHandler` recorded no `ModActivity` at all before this PR and would have been silently demoted; it now writes one. That is an additive change to the moderator audit log — every reader of that log was walked, and none of them (queue-stats, minor-hash, the account-history buckets) is perturbed. The Appeals card's "Removed … by X" lateral now populates where it was previously blank.

### Operator decisions taken here, deliberately

Two writers were **excluded by decision**, not by accident:

- **Automated blocks do not retract** — the content-rating webhook, the prompt audit and the moderation rule engine. Retraction is a rare, deliberate act, and a classifier false positive that nobody appeals inside the retention window would otherwise destroy every other owner's byte-identical original.
- **The ban's remove-all-media branch does not retract** — its scope is the banned account's *entire library* rather than the violating content, so the collateral is multiplied by every image they ever uploaded. This is the same reasoning that leaves `api/mod/delete-user-images` off.

Both still hard-delete the row. Only the shared bytes are spared.

## Flows that call `deleteImages` and deliberately do not retract

`deleteImageById` is given **no option at all** — it is the user-facing delete and has no moderation caller. `remove-replaced-images`, `remove-deleted-user-images`, post deletion's direct `deleteImageFromS3`, the `hideMeta` `purgeResizeCache` (which *refuses* retraction on a partial scope and logs `image-blob-retraction-refused`), and article/collection/user cover cleanup all keep variant-only invalidation.

`api/mod/delete-user-images` and the `admin/delete-images` webhook are moderator-adjacent but stay off: the first deletes a whole library including never-moderated images, and the second has no session to attribute intent to.

## Account-deletion restore breadcrumbs

`remove-deleted-user-images`' grace path records each image's prior `ingestion` in metadata so a reinstated account can be restored. Moderation block sites now clear that breadcrumb when they block, so a takedown cannot later be un-done by an account restore: `handleBlockImages`, `setTosViolationHandler`, `softDeleteUser`, `toggleBan`, and inline in the moderator app (which cannot import `src/`) with a drift guard pinning both branches of its SQL.

🔴 **`toggleBan`'s clear is scoped to the ids its own UPDATE blocks**, not account-wide. Account-wide looks equivalent and is not: `handleUnblockImages` (`/api/mod/restore-images`) sets `ingestion='Scanned'` and strips only `ruleId`/`ruleReason`, so a row can be **not-Blocked and still breadcrumbed** — and an account-wide clear would strip the restore record off rows the ban never touched, leaving a reinstated user's library permanently blocked with `imagesPendingRestore` reading 0. That regression was introduced and caught inside this PR's review; do not widen the scope back.

## 🔴 A consequence that is known and accepted, not overlooked

The stored object is content-addressed, so it is shared by **every byte-identical image, of
every owner**. Retracting it removes the full-resolution original for all of them at once. Those
other images keep live database rows and will serve a broken original — an orphaned-row symptom,
deliberately reintroduced.

This is accepted: a bit-identical copy of content that must not exist is the same content, and a
takedown that leaves copies serving is not a takedown.

**It is not fixed here, and it cannot be from this codebase.** The app stores no content-hash key
per image, and the `pHash` it does store is a *perceptual* hash — a similarity signal, not a
byte-identity key — so it is not a substitute. `Image.hash` is a blurhash placeholder, not a
content hash either. Enumerating the affected rows would need a content-hash column that does not
exist; building that fan-out is separate work and is out of scope here.

The only attribution trail is a log line emitted per retracted image in `deleteImageFromS3`:
`image-blob-retraction-requested`, carrying the image id and its key. It is emitted *before* the
request, so a request that is made but never confirmed still leaves a record. It is asserted by a
test in both directions (present on retraction, absent on an ordinary delete).

## A second consumer of the same endpoint

`apps/moderator/src/lib/server/image-deletion.ts` open-codes its own POST to
`/admin/invalidate` with no `X-Admin-Secret`, no `res.ok` check and no retraction parameter, so it
invalidates variants only and none of this reaches it. It is **not** wired up here — that is a
separate change with its own risk surface — but it is now named in the comment beside this
caller's header, which previously claimed that sending the header made enabling the destructive
mode "a change on ONE side, not a synchronised deploy across two services". That was not true of
the tree, and it predated this branch. Both that comment and the `!res.ok` one now assert only
what is observable from this repo; what the service does with a missing header is not.

## Tests

New:

- `src/server/services/__tests__/moderation-blob-retraction.test.ts` (17 tests) — the layer-by-layer
  opt-in.
- `src/server/services/__tests__/no-unmoderated-blob-retraction.test.ts` (5 tests) — a
  source-scanning ledger of the call sites.
- `src/server/jobs/__tests__/account-deletion-no-blob-retraction.test.ts` (2 tests) — the
  reachability regression: it drives account deletion → grace block → job queue → retention purge
  and asserts the purge does not request retraction for those images, while a moderator-blocked
  image in the same batch still does.

Two tests added to `remove-blocked-images-retention.test.ts`.

The central claim is the **negative** one — an ordinary user delete must never send the parameter.
Every "does not send it" assertion is preceded by an assertion that the invalidate request
actually happened, so a path that throws before reaching the service cannot satisfy it vacuously.
The job-level test carries the same discipline one level up: it asserts each image *was* deleted
before asserting how, so a path that silently stops deleting cannot satisfy "did not retract".

Because the set-of-call-sites claim is not behavioural, the ledger guard is that half: it asserts
the exact set of files that mention the option and the exact set of **call sites** that turn it
on, and it fails in **both** directions — a new site starting to retract, or the moderation flow
silently ceasing to.

### Ledger guard: two walks it used to allow

Both were demonstrated fully green against the previous version of that file, which ledgered file
NAMES over a walk rooted at `src`:

1. a **second** retracting call site appended to a file already on the ledger — `image-ingestion.ts`
   hosts several jobs, so that is exactly where a copy-paste lands;
2. a setter **outside `src`** — `apps`, `packages` and `scripts` were invisible to the walk.

It now counts call sites per file (`path ×N`) and walks `src`, `apps`, `packages`, `scripts`,
`test` and `tests`, ~7,900 files. Re-measured after the change:

| Mutation | Old guard | New guard |
|---|---|---|
| second `retractPublicBlobs: true` in `image-ingestion.ts` | — | red: `×1` vs `×2` |
| `retractPublicBlobs: true` in `packages/civitai-shared/src/...` | **green** | red, on both the setter and the mentions ledger |
| moderation flow stops asking (shrink direction) | red | red: `[]` vs `["…image-ingestion.ts ×1"]` |
| walk re-narrowed to `SCAN_ROOTS = ['src']` | — | red: `['packages','apps']` contributed no files |

That last row is the point of the new breadth control: it reads the roots it requires out of
`pnpm-workspace.yaml`, not out of the guard's own `SCAN_ROOTS`, so re-narrowing the walk cannot
approve itself.

### Red -> green

Service layer, measured with the tests present and the implementation absent (i.e. at the base
commit for `image.service.ts`), then with the implementation:

| | base | HEAD |
|---|---|---|
| `moderation-blob-retraction.test.ts` | **6 failed / 11 passed (17)** | **17 passed** |

The six red at base:

1. `purgeResizeCache > asks for retraction when the option is explicitly true`
2. `purgeResizeCache > changes nothing but the added parameter`
3. `purgeResizeCache > refuses to retract on a partial-scope purge`
4. `deleteImageFromS3 > passes retraction through when asked`
5. `deleteImageFromS3 > records the retraction request against the image id`
6. `deleteImages > asks for retraction when the moderation flow says so`

Job layer, `account-deletion-no-blob-retraction.test.ts`, against the pre-split call site:

```
× deletes the grace images without asking for retraction, while still retracting a takedown
AssertionError: an account-deletion grace block reached the retracting call: a user deleting
their own account must not destroy other owners' stored bytes: expected true to be false
Tests  1 failed | 1 passed (2)
```

**The eleven service-layer tests that pass at base are invariant guards, not regression
coverage** — at base the parameter did not exist, so nothing could send it. They are stated as
such and are validated by mutation instead (M1, M12, M13).

### Mutation results

Every mutant below was applied to the real source, run, and reverted.

Service layer — all 13 killed, and the reported failure is **that guard's own assertion** in each
case. ⚠️ These were measured on the previous round and are **not** re-run here; the job-level
split changed the call site M9 restores, so read M9 as superseded by J1 below:

| # | Mutation | Killed by |
|---|---|---|
| M0 | drop `retractParam` from the query (positive control) | 4 tests, `to contain '&retractPublicBlobs=true'` |
| M1 | `purgeResizeCache` default flips to `true` | `does not ask for retraction when the option is omitted` — `not to contain` |
| M2 | remove the `scope === 'all'` conjunct | `refuses to retract on a partial-scope purge` — `not to contain` |
| M3 | silence the refusal log | `refuses to retract on a partial-scope purge` — `logToAxiom to be called with` |
| M4 | `deleteImageFromS3` stops forwarding to `purgeResizeCache` | `passes retraction through when asked` |
| M5 | `deleteImages` stops forwarding to `deleteImageFromS3` | `asks for retraction when the moderation flow says so` |
| M6 | drop the attribution log | `records the retraction request against the image id` |
| M7 | `retractPublicBlobs=1` instead of `=true` | 4 tests, exact-literal assertion |
| M8 | omit the `&` separator | 4 tests, exact-literal assertion |
| M9 | moderation job stops asking (= the base call site) | both job tests + the setter ledger |
| M10 | options object passed in `updatePosts`'s position | `leaves post updating on while doing so` — `{…} to be true` |
| M11 | a second, non-moderation file starts retracting | the setter ledger + the plumbing ledger |
| M12 | `deleteImageFromS3` default flips to `true` | 3 tests incl. `deleteImageById never retracts` |
| M13 | `deleteImages` default flips to `true` | `does not ask for retraction by default` + the `updatePosts`-only case |

Job layer — the split and the marker seam:

| # | Mutation | Killed by |
|---|---|---|
| J1 | restore the pre-split call site (everything retracts) | `an account-deletion grace block reached the retracting call` |
| J2 | invert both filters | same assertion |
| J3 | reader looks up a different metadata key | same assertion |
| J4 | drop the retracting call entirely | `a moderator takedown stopped asking for retraction` |
| J5 | **writer** stamps a different metadata key | both tests: `expected undefined to be 'Scanned'` and the grace assertion |

J3 and J5 are the seam in both directions: the test computes the marker from the key each job
actually **binds**, not from the shared constant, so the writer and the reader agreeing is what is
being asserted rather than assumed.

### Counts

- `pnpm typecheck` — **0 type errors** (heap cap 8192 MB).
- `pnpm run test:lint-rules` — **33 files, 482 tests, 0 failed**.
- `pnpm run test:unit:run` (full), run twice:
  - mid-round, on the split + ledger changes — **1651 files: 1648 passed, 3 skipped; 38300 tests: 38300 passed, 33 skipped, 0 failed**. Zero failing-test lines, no timeout panic.
  - on the final head — **1 file / 2 tests failed**: both in
    `src/server/__tests__/eventloop-watchdog.capture.test.ts`, the CPU-timing test that deliberately
    saturates threads. Load flake, and measured as one rather than assumed: that run's `tests` phase
    was 260.79s against 184.17s for the green run above (+42% across the whole suite, the load
    signature rather than one assertion's), and the file alone gives **1 failed / 8 passed at 20.34s
    while the tail of the suite was still winding down, then 9 passed at 9.59s and 9 passed at
    9.34s**. It is in an unrelated module graph — nothing in it touches the image service, the jobs
    or the queue — and CI's four unit shards plus `App unit tests + typecheck` are green on the same
    commit. It flaked on the previous round too. Recorded as a pre-existing flaky test, not as
    evidence about this change, and not fixed here.
- `eslint` on the touched files — **0 new problems**. `image.service.ts` carries 3 pre-existing
  errors (`consistent-type-imports` L205, `prefer-const`, `no-module-scope-cache`), all present at
  the base commit and none in the changed regions.
- `pnpm run prettier:write` — 5 files reformatted, then `prettier:check` clean.
- CI on the current head — **20 checks, all terminal: 19 SUCCESS + 1 SKIPPED, 0 failures.**

## Also in this diff

`src/server/flipt/__tests__/flipt-eval-context.test.ts` keys its ledger on `file:line`, and the
comment changes in `image.service.ts` moved four entries. The line numbers are updated; no entry
was added or removed.

## Deployment

**This is inert until deployed** — nothing changes for anyone until the app ships. Once it does,
the moderation takedown job gains a destructive, cross-account capability that it did not have
before, with the collateral described above. Worth watching
`image-blob-retraction-requested` after the first run of the hourly job — and worth checking that
the retracted count in that job's return value is not tracking total deletions, which is what a
regression in the split would look like.


ZacxDev:
Round 1 adversarial audit complete (1 🔴, 3 🟡). Fixes landed in `e73349e857`…`bdfed26f08`.

```audit-claims round=1 audited=bb05a7d325..bdfed26f
1. `removeBlockedImages` no longer retracts for images that reached it via account deletion: the query derives `fromAccountDeletion` from the existing `accountDeletionPriorIngestion` metadata marker and splits the batch, so only unmarked rows are passed `retractPublicBlobs: true`. Both populations are still hard-deleted.
2. A regression test drives the account-deletion grace path through to the retraction decision and asserts retraction is not requested; claimed red on the pre-fix call site.
3. The ledger guard was widened past `src/` to the workspace roots and now asserts call-site counts (`path xN`) rather than file names, so a second setter inside an already-listed file and a setter outside `src/` both trip it; the both-directions property is claimed to hold.
4. The two false comments about a one-sided deploy were corrected to assert only what is observable from this repo, and the second consumer in `apps/moderator` is now named. `apps/moderator` itself was deliberately NOT wired.
5. The comment claiming every image reaching the retracting call site was blocked by moderation, and the PR body's flow table, were rewritten; the table is now a queue-writer table.
6. The queue-writer enumeration was corrected from 4 to 7 writers; the claim retained is that `remove-deleted-user-images` is the only non-moderation writer.
```

### 🔴 Open question raised BY the fix round, deliberately not decided

`toggleBan`'s "remove all media" branch blocks a banned user's **entire library** with `blockedFor='moderated'` and relies on this job to hard-delete it after the appeal window — so it now retracts a whole library. That is in tension with this PR's own reason for leaving `api/mod/delete-user-images` **off** (whole-library ⇒ collateral multiplied by every image). The difference is the route, not the intent. Left **ON**; it is a policy call for whoever owns the ban flow.

### Residual design note for the next round

The discriminator is **negative**: absence of the account-deletion marker means "retract". A future non-moderation blocking flow that does not set that marker would therefore retract silently — the shape of the defect this round fixed. A positive takedown marker on the queue row would fail closed instead; not done here.

Ledger — round 1 · payload lines changed THIS round: `image-ingestion.ts` (the split + query) and the ledger/regression suites; scaffolding is the two test files.

ZacxDev:
Round 2 (delta) complete: 3 🔴, 2 🟡, 1 🟢 — all closed. Fixes in `ecad90d5dc`…`673d80b198`.

```audit-claims round=2 audited=bdfed26f08..673d80b1
1. The retraction discriminator was INVERTED from negative to positive: `Image.metadata` no longer participates at all. The job retracts only where a `ModActivity` row with activity in ('review','bulkRemove'), entityType='image', exists dated at or after the queue row.
2. A forged `accountDeletionPriorIngestion` from the upload path (both a string value and JSON null) is claimed to move the decision in neither direction, driven through the real `createImage`.
3. Automated blocks (content-rating webhook, prompt audit, moderation rule engine, and the second copy of that pipeline in the webhook route) are claimed not to retract, driven through the real `processImageScanWorkflow`.
4. `toggleBan`'s remove-all-media branch is claimed not to retract, driven through the real `toggleBan`.
5. `setTosViolationHandler` now records a `ModActivity` row it previously did not — an additive change to the moderator audit log, without which that takedown would have been silently demoted.
6. Account-deletion markers are cleared at the moderation block sites (`handleBlockImages`, `setTosViolationHandler`, `softDeleteUser`, `toggleBan`, and inline in the moderator app with a drift guard).
7. The ledger's workspace-root parse is scoped to the `packages:` block, tested against the shape `pnpm approve-builds` writes.
8. No schema change was required; no Prisma migration is part of this PR.
```

### Disclosed by the fix round, worth a reader's attention

**`report.service`'s CSAM branch is reachable by any logged-in user** (`report.create` is `guardedProcedureAllowUnverifiedEmail`). Under the previous opt-out predicate a stranger filing a CSAM report on someone else's image would have queued it for retraction. Closed by the polarity change; recorded because it shows why the polarity mattered rather than being a style preference.

**Accepted imprecision, documented in-code:** automation blocks → a moderator unblocks (`review`) → automation re-blocks, all inside one retention window. The surviving queue row keeps the first block's timestamp, so the unblock dates after it and licenses retraction. Closing it needs `review` split into distinct block/unblock activities, which changes the vocabulary the account-history panel buckets on. Not done.

**The backfilled queue population cannot retract** — rows inserted by the 2026-08-06 completeness migration carry that migration's `createdAt`, later than any moderator activity on them. Fail-closed.

Ledger — round 2 · payload changed THIS round: the discriminator rewrite in `image-ingestion.ts`, marker-clearing at 4 moderation block sites + the moderator app, and one added `trackModActivity`; scaffolding is the new reachability/marker suites and the ledger guard.

ZacxDev:
Round 3 (delta) complete: 1 🟡 behavioural + 3 🟡/🟢 guard-and-prose. All closed. Fixes in `2f6502961d`…`4dc99ebbd7`.

```audit-claims round=3 audited=673d80b198..4dc99ebb
1. `toggleBan` no longer clears account-deletion breadcrumbs account-wide: it reads the ids its own UPDATE will block and clears only those, inside a try/catch so the read can never stop the media block.
2. `softDeleteUser`'s account-wide clear is retained deliberately (its UPDATE carries no ingestion predicate, so the sets coincide) and now says why.
3. Assertions were added for `setTosViolationHandler` and `softDeleteUser`, and for both scope directions of `toggleBan`; the docstring now names only what is actually pinned.
4. The `MODERATOR_TAKEDOWN_ACTIVITIES` docstring was rewritten to match the code (it compares against the queue row, not the re-block) and the retraction window now carries its mechanism rather than a bare duration.
5. The moderator-app drift guard asserts both ternary arms separately.
6. The test fake now reads which arrow a statement used and throws on an unrecognised shape; `$queryRaw`/`$executeRaw` return promises as Prisma does.
7. The `ON CONFLICT` comments record that the unique-index drop is applied in production while `containers/db/docker-init` still creates it, so local databases diverge.
```

### The premise correction that changed the fix

The brief for this round asserted the two row sets were disjoint by construction, and proposed removing `toggleBan`'s clear entirely. **That was wrong.** `handleUnblockImages` (`image.service.ts:796`, reachable via `/api/mod/restore-images`) sets `ingestion = 'Scanned'` and strips only `ruleId`/`ruleReason` — it leaves the account-deletion breadcrumb in place. So a row can be not-Blocked *and* breadcrumbed, and that row is one the ban blocks. Removing the call would have reintroduced the round-2 regression in exactly that window. Verified independently at the tree before accepting the narrowing.

Two further corrections found while fixing: the round-3 docstring named `handleAcceptImages`, **a function that does not exist** anywhere in the repo (it is `handleUnblockImages`); and `remove-deleted-user-images` has *two* breadcrumb-writing statements, not one.

### Mutation

The three mutants that SURVIVED round 3 are now killed, each by its own assertion — the `remixSource`-branch strip, `->` narrowed to `->>`, and deleting a marker-clear. 10/10 killed, 0 survived.

### Known gap, stated rather than papered over

No single test spans `toggleBan` → `restoreUser` end-to-end. The breadcrumb count is asserted through the real `countPendingAccountDeletionImageRestores` against the fake DB, and `restoreUser`'s gating on that count is covered separately with a mocked counter. That seam is open; wiring `restoreUser` into the reachability fixture was judged not worth the fragility.

Ledger — round 3 · payload: the `toggleBan` narrowing in `user.service.ts` plus docstring corrections in `image-ingestion.ts` and `moderator.service.ts`; scaffolding is the marker-clearing and reachability suites.

ZacxDev:
**Round 4 (delta) is clean of code defects — the audit ladder ends here.** No round 5: the process forbids running one to confirm a clean round.

`round 4 · payload lines changed THIS round: 147 (since round 1: 483)` — of which 83 were comment-only; the single behavioural change was the `toggleBan` narrowing.

### What the four rounds found

Each round found something real, and three of the four were introduced or left open by the previous round's fix:

1. **R1** — ordinary self-service account deletion reached retraction (the "moderation only" invariant was true of the direct callers and false of the queue-fed graph).
2. **R2** — the fix for R1 keyed on `Image.metadata`, a field **any uploader writes at upload time**, so anyone could pre-exempt their own image; and the negative polarity also admitted ML classifier output, whole-library bans, and a **stranger-filed CSAM report**.
3. **R3** — the polarity inversion was correct, but its marker-clearing stripped account-deletion restore breadcrumbs **account-wide**, silently destroying the restore state of a self-deleted library.
4. **R4** — clean. The one 🟡 was this description, not the tree.

None of these was visible to the type checker, the ~38,300-test suite, or CI, all of which were green throughout.

### The description was stale and is now corrected

It described the pre-R2 negative discriminator and listed **four writers as retracting that do not** — including `report.service`'s CSAM branch, the stranger-reachable one. Anyone who read this PR before now should re-read "What this does" and the writer table. Counts and the CI line are updated to the current head.

### Verified independently, not taken from a report

The `ModActivity` append-only migration **is applied in production** — `pg_indexes` on `ModActivity` returns four indexes and the only unique one is the primary key on `id`. That was the one premise the round-4 auditor flagged as unverifiable from the repo, and it is what the `ON CONFLICT DO NOTHING` reasoning rests on. Note `containers/db/docker-init` still creates the dropped index, so **local databases diverge from production here** — a developer reproducing locally would see a demotion that cannot happen in prod.

### Still open, for the merger — not defects

- **No single test spans `toggleBan` → `restoreUser`.** The breadcrumb count is asserted through the real counter against the fake DB; `restoreUser`'s gating on it is covered separately with a mocked counter. Wiring the two was judged not worth the fixture fragility.
- **The read/UPDATE/clear triple in `toggleBan` is three statements, not one.** A `$queryRaw` failure or a concurrent un-block leaves a stale breadcrumb on a blocked row — logged (`ban-user-grace-marker-scan`), recoverable, and erring toward keeping an image up rather than destroying bytes. Folding the strip into the block statement would remove both windows; deliberately not done, because the destructive UPDATE's scope is kept literal in one place.
- **Merging is not deploying.** This is inert until the app ships, and it then grants the takedown job a destructive cross-account capability it did not have.

PR #4663, opened 2026-09-06
feat(blocks): publish coverNsfwLevel on the collections list response

ZacxDev:
## What

`GET /api/v1/blocks/collections` returns a `coverImageUrl` per collection but no maturity level for that cover. This adds an optional `coverNsfwLevel` alongside it, on both `mode=public` and `mode=mine`.

Verified against the live endpoint before starting: the first item's keys were exactly `[coverImageUrl, curator, description, followed, id, isPublic, itemCount, name]`.

Additive and non-breaking — no existing field changed or removed, and no schema change (`Image.nsfwLevel` already exists).

## Why now

The consumer already exists and is already written against this field. A first-party collections block declares `coverNsfwLevel?: number` in its types and reads it at three call sites behind `nsfwLevel !== undefined ? shouldBlur(nsfwLevel) : !isSfw`. Today it is always `undefined`, so every card falls back to a domain-wide ceiling instead of gating on the cover it is actually painting.

## The three things that make this subtle

**1. The level must describe the cover ACTUALLY RETURNED.** The cover is `primaryCoverUsable(c) ? c.image : fallbackCovers.get(c.id)` — the primary `Collection.image` when its own level clears the token's clamped ceiling, otherwise the maturity-clamped fallback item. Publishing the primary's level beside a fallback url is *worse than publishing nothing*: the consumer treats a supplied level as authoritative and an absent one as "fall back to the ceiling", so a mismatched level is a lie it will gate on — and the mismatch case is exactly the one where the primary was rejected for being mature.

Rather than test for that, `toCoverFields()` makes it unrepresentable: it takes **one** image and returns **both** fields from it, so the caller picks the image once and the two cannot drift.

**2. `getFallbackCoverImages` did not return the level.** It selected `url` and `type` only — `nsfwLevel` appeared solely in the `WHERE`. For a fallback cover *that row is the served image*, so its own level now travels with it. Its only two call sites are the two mapping sites in this endpoint; both are updated in this PR.

**3. No cover ⇒ the field is ABSENT, never `0`.** `0` is a real level meaning *unrated*, and the consumer's guard branches on `undefined` vs a real value, so emitting `0` for "there is no cover" would silently move it onto a different code path. The presence of the level is decided by the same expression as the presence of the url (`coverImageUrl === null`), so they cannot disagree.

## Naming, deliberately

It is `coverNsfwLevel`, **not** `nsfwLevel`. A Collection has its own `nsfwLevel` — a bitmask OR-ed over its items — and this endpoint's existing doc comment already records at length that it cannot separate a 97%-safe contest collection from a 1%-safe mature one (both carry `29`). A field named `nsfwLevel` on this response would invite exactly that confusion. Both doc comments say so explicitly, and a test asserts the item's key set contains no bare `nsfwLevel`.

## Semantic, and a consumer-side mismatch to fix separately

The fallback cover is selected `ORDER BY ci."createdAt" DESC` — the **newest** clamped accepted item. It is deliberately not a "best" or "highest-rated representative"; nothing on this path ranks items. The server-side doc comment now states the semantic as *the level of the image being served as the cover*.

⚠️ The consumer's own type comment currently describes the cover as "its highest-rated representative item", which does not match what the server selects. That is a comment fix in the consuming repo, not here — flagging it so it can be corrected there.

## Verification

Runner: `node scripts/test-unit-run.mjs` (→ `vitest run --project 'unit*'`); typecheck `node scripts/typecheck.mjs`.

- **Full unit suite:** `1649 passed | 3 skipped (1652)` files, `38311 passed | 33 skipped (38344)` tests.
- **Typecheck:** `0 type errors`.
- **Adjacent suites** (`src/server/services/blocks/__tests__`, `collection-item-count-clamp*`, `src/tests/api/v1`): `210 passed` files / `4628 passed` tests.
- **ESLint** on the four changed files: 0 errors (51 pre-existing-style `no-explicit-any` warnings in the test file).

**Red at base.** Both test files were run against `origin/main`'s sources with the new tests in place: **19 failed / 40 passed**. Every new or changed assertion is red at base; the 40 are the pre-existing tests, unchanged.

One new test initially passed at base and was strengthened rather than counted: an "is the key absent?" assertion is satisfied for free by a response that publishes no level anywhere. It now includes a cover-bearing sibling in the same page, which proves the response *does* publish levels, and it is red at base.

Two assertions are **invariant guards, not regression coverage**, and are labelled as such in-file: "no bare `nsfwLevel` key" held before this change too.

**Mutation-checked — 13 mutants, all killed**, each by the intended test with that test's own assertion, on inputs no earlier check rejects. Every mutant edit was `cmp`-verified to have actually changed the file before the run.

| # | Mutation | Killed by |
|---|---|---|
| M1 | no-cover emits `0` instead of omitting | `NO COVER → the level field is ABSENT, not 0` |
| M2 | truthiness guard drops a real `0` | `an UNRATED (0) cover publishes a real 0` |
| M3 | absent level normalises to 1, not 0 | `a cover with a null/absent level is UNRATED (0)` |
| M4 | fallback map drops the level | 4 tests incl. the seam test |
| M5 | null column level becomes 1 | `a NULL nsfwLevel column reads as the real level 0` |
| M6 | SQL stops selecting `nsfwLevel` | `SELECTS the item nsfwLevel` |
| M7 | public branch levels the PRIMARY unconditionally | mature-clamped / no-cover / mixed-page |
| M8 | public branch levels the FALLBACK unconditionally | usable-primary / no-cover / mixed-page |
| M9 | mine branch levels the PRIMARY unconditionally | `mode=mine: the served cover's level…` |
| M10 | field renamed to `nsfwLevel` | 7 tests |
| M11 | mine branch levels the FALLBACK unconditionally | `mode=mine: the served cover's level…` |
| M12 | mine branch coerces no-cover to `0` | `mode=mine: the served cover's level…` |

M9's first anchor was a **substring of the public branch's more-indented line**, so it silently mutated the public branch and reported the public tests as killers. Caught by the anchor-occurrence count in the harness, re-anchored on the preceding comment line, and re-run — the corrected M9/M11/M12 are each killed by the `mode=mine` test alone, which is what proves that branch is independently covered.

Coverage includes: primary-cover, fallback-cover, no-cover, unrated (`0`) published as a real `0`, both modes, and a mixed page asserting item-by-item that each published level belongs to the image its url came from.

## Not verified

Only the SQL text is asserted for the `nsfwLevel` projection — `dbRead.$queryRaw` is mocked, as it already was for this function, so the column is not exercised against a live Postgres. The `?? 0` normalisation of a null column read is covered at the mapping layer, not against a real null row.

## Overlap with open PRs

Based directly on `main`, not stacked. Checked both adjacent open PRs for file overlap: **none**.

- **#4661** (`zach/collections-index-stale-images`) — touches `collection-media-index.ts`, `image.service.ts`, `model.service.ts`, a migration. No shared file.
- **#4597** (`feat/allow-creators-to-archive-collections`) — touches `collection.service.ts`, controllers/routers/schema/selectors, a migration. No shared file; this PR does not modify `collection.service.ts`.

### DIFF
--- a/src/pages/api/v1/blocks/collections/index.ts
+++ b/src/pages/api/v1/blocks/collections/index.ts
@@ -18,7 +18,7 @@ import {
   getFallbackCoverImages,
   getFollowedCollectionIds,
   hydrateBlockSubject,
-  toCoverImageUrl,
+  toCoverFields,
 } from '~/server/services/blocks/block-collections.service';
 import { resolveCatalogBrowsingLevel } from '~/server/utils/block-catalog-maturity';
 import { checkBlockCatalogRateLimit } from '~/server/utils/block-catalog-rate-limit';
@@ -57,8 +57,23 @@ import {
  * domain. The separating signal is HOW MUCH OF THE COLLECTION SURVIVES THE
  * CEILING, sampled — see `MIN_PLAYABLE_FRACTION` and `PLAYABLE_SAMPLE_SIZE`.
  *
- * Response: `{ items: [{ id, name, description, coverImageUrl, itemCount,
- *   curator:{ userId, username }, isPublic, followed }], nextCursor }`.
+ * Response: `{ items: [{ id, name, description, coverImageUrl, coverNsfwLevel?,
+ *   itemCount, curator:{ userId, username }, isPublic, followed }], nextCursor }`.
+ *
+ * 🔴 `coverNsfwLevel` IS THE LEVEL OF THE COVER BEING SERVED — NOT THE
+ * COLLECTION'S `nsfwLevel`, AND THE NAME IS DELIBERATE. The collection-level value
+ * is the OR-ed bitmask described above, which cannot separate a 97%-safe contest
+ * collection from a 1%-safe mature one; a field named `nsfwLevel` on this response
+ * would invite exactly that confusion. This one describes a SINGLE IMAGE: whichever
+ * image `coverImageUrl` resolves to — the primary `Collection.image` when it is
+ * usable, otherwise the maturity-clamped fallback, which is the collection's
+ * NEWEST permitted accepted item (not a "highest-rated representative"; nothing
+ * here ranks items). Both fields are produced together by `toCoverFields` from the
+ * one chosen image, so they can never describe different images.
+ *
+ * 🔴 IT IS OMITTED WHEN THERE IS NO COVER, AND `0` IS A REAL LEVEL. A consumer
+ * reads `undefined` as "no claim — use your own domain ceiling" and any supplied
+ * value as authoritative, so `0` (unrated) must never stand in for "absent".
  *
  * 🔴 `itemCount` DIFFERS BY MODE, DELIBERATELY. In `mode=mine` it is CLAMPED to
  * the token's ceiling — that branch counts only the subject's own collections, a
@@ -314,7 +329,10 @@ const baseHandler = withAxiom(async function handler(req: NextApiRequest, res: N
           id: c.id,
           name: c.name,
           description: c.description ?? null,
-          coverImageUrl: toCoverImageUrl(primaryCoverUsable(c) ? c.image : fallbackCovers.get(c.id)),
+          // ONE image in, BOTH cover fields out — the url and the level of that
+          // same image. Splitting these back into two expressions is how a
+          // fallback cover ends up advertising the primary's level.
+          ...toCoverFields(primaryCoverUsable(c) ? c.image : fallbackCovers.get(c.id)),
           itemCount: countMap.get(c.id) ?? 0,
           curator: { userId: c.userId, username: c.user?.username ?? null },
           isPublic: c.read === CollectionReadConfiguration.Public,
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Consumer-side check: what publishing this field actually changes on screen

Neither CI nor the unit tests can answer this, so I measured it against a production-scale replica. The concern is specific: the consumer treats an **absent** level as "fall back to the domain ceiling" (not blurred on a SFW domain) but a supplied `0` as a **real** rating — *unrated* — which is fail-closed and **does** blur. So publishing the field turns every unrated cover from unblurred into blurred.

Page-1 discovery window under the endpoint's default sort, SFW ceiling (3), reconstructing the exact cover each row would be served (primary when itself within the ceiling, else the clamped fallback):

| | |
|---|---|
| collections on page 1 | 24 |
| cover came from the **primary** image | 19 |
| cover came from the **clamped fallback** | 5 |
| **no cover at all** (field correctly absent) | **0** |
| **unrated (`0`) — would newly blur** | **0** |
| rated level published | **24** |

Served cover levels: **22 at level 1**, **2 at level 2**. Both are below the blur threshold, so **no card on the current page 1 changes appearance.** The field just stops being a guess.

### What this does not establish

- **One population, one ceiling.** This is page 1 on a SFW ceiling. The case the field exists for is a **red-capable** ceiling, where a mature cover should blur and currently falls through to `!isSfw` — I have no red-domain token and did **not** measure it. The unit tests cover that path; the live population does not exercise it.
- The `nsfwLevel` SQL projection is asserted as text in the suite (`$queryRaw` is mocked, as it already was for this function), so the added column is not exercised against a live Postgres by the tests. This probe does exercise the equivalent query shape, which is the closest independent evidence available without a block token.
- Cover composition drifts. A future re-check should re-run the probe rather than cite this table.

PR #4668, opened 2026-09-07
feat(apps): surface Build apps as an /apps/* subnav tab, collapse the dropdown to one entry

ZacxDev:
## What

`/apps/get-started` ("App builders") stops being a second row in the user-avatar dropdown and becomes the first tab of the shared `/apps/*` subnav.

## 1. The dropdown consolidation

`useGetMenuItems` carried **two adjacent** App-Blocks rows: "Build apps" → `/apps/get-started` (on `appsNav.getStarted`) and "Apps" → `/apps` (on `appsNav.marketplace`). A moderator holds both flags, so they saw two near-identical rows for one product.

They are now **one** row:

```ts
href: appsNav.marketplace ? '/apps' : '/apps/get-started',
visible: appsNav.marketplace || appsNav.getStarted,
```

The fallback is not cosmetic. A viewer holding `appBlocksGetStarted` **without** a store flag cannot load `/apps` at all — its `getServerSideProps` runs `resolveAppsPageAccess`, which returns `notFound`. Pointing them at `/apps` would be a menu entry into a 404, so they land directly on the one page they are entitled to. Store-access users get the richer marketplace landing and reach Build apps from the subnav.

`appsNavVisibility` is unchanged and still returns both booleans — deleting `getStarted` would send that cohort at `/apps`. The now-expired `newUntil: 2026-08-01` is dropped; `IconCode` moves from `hooks.tsx` to `AppsSubNav.tsx`.

## 2. The gate widening — required, not optional

`AppsSubNav`'s whole-bar gate goes from

```ts
if (!hasAppsStoreAccess(features)) return null;
```

to

```ts
if (!hasAppsStoreAccess(features) && !features.appBlocksGetStarted) return null;
```

Without the second term the container returns `null` for **exactly the cohort the new tab exists for**, so the tab would be invisible to them on the one `/apps/*` page they can load — the empty chrome band that `pages/apps/get-started.tsx` had a `TODO(launch)` about. It is the union of the two pages' own gates (`resolveAppsPageAccess` and `resolveGetStartedAccess`).

Safe to apply on **first paint**, for the same reason already written out for `appBlocksAuthor` a few lines below — now stated rather than left implied: `appBlocksGetStarted` is resolved server-side in `_app`'s `getInitialProps`, serialized into `pageProps.flags`, and frozen by `useState(initialFlags)` in `FeatureFlagsProvider`. It is not `toggleable: true`, so the client overlay cannot move it. Server render and first client render compute the same boolean, so there is no tab-set hydration mismatch.

The `enabled:` predicate on `blocks.getNavSummary` is deliberately **not** widened with it (it must mirror the gate on the *procedure*, `enforceAppBlocksFlag`). Pinned by a test.

~~`visible: () => true` on the new row is deliberate: the flag decision lives in the bar gate, keeping per-item predicates about viewer STATE, which is what every other row does.~~ **Withdrawn in round 2 — this was the defect.** The bar gate is an OR, so it admits store-access viewers who do *not* hold `appBlocksGetStarted`, and an unconditional row offered them a tab into a 404. The row is now `visible: (_s, c) => c.canGetStarted`, the flag threaded onto `AppsNavContext` exactly the way `appBlocksAuthor` is threaded as `isAuthor`. See **Review round 2** below.

## 3. ~~The 1 → 2 tab visibility side effect~~ — **reverted in round 2**

Round 1 made "Build apps" unconditional, taking the always-on tab set from one (Marketplace) to two and making the `links.length < 2` collapse unreachable through the container. Round 2 gates the row on `canGetStarted`, so **none of that happens any more**:

- A store-visible non-author with no installs and no `appBlocksGetStarted` qualifies for Marketplace alone and the collapse hides the bar — **exactly `main`'s behaviour**, no visibility change for that cohort. Same for a logged-out viewer with the store flag lit.
- The `< 2` collapse is therefore reachable through the container again, by a live cohort. The tests round 1 had to re-label as "pinning the view, not a cohort" are re-labelled back to what they are: coverage of a real viewer. Mutation-checked — deleting `links.length < 2` turns **6** of them red.

Geometry pins checked, and one needed a fix: `AppsPageLayout.geometry.browser.test.tsx` asserted `tabWidth ≈ 133.27` on `document.querySelector('[role="tab"]')` — the FIRST tab. 133.27px is a measurement of the string "Marketplace"; with "Build apps" leading the bar the same lookup silently became a different box (**122.39px measured**). Rather than re-baseline to a number nobody has a reason for, the selector is re-pointed to `[role="tab"][href="/apps"]` so the pin still fails for the reason it was written (a `padding` shorthand instead of `paddingBlock`). `AppsPageLayout.chromeAlignment.browser.test.tsx` and `__tests__/appsPageLayoutRender.test.ts` are unaffected (nav/body geometry and a mocked subnav respectively) and pass unchanged.

## 4. The app-block chrome: EXCLUDED, and the exclusion is now asserted

`chromeNavAlignsWithSubNav.test.ts` treats `SUB_NAV_LINKS` as source of truth, but its rule is **one-directional** — every *chrome* destination must exist in the subnav, never the reverse. The chrome is already a strict subset (`Create` / `Invites` / `Revenue` have always been subnav-only), so adding a subnav row keeps the guard green whether the omission was a decision or an oversight.

`/apps/get-started` is excluded from the chrome, for two reasons:

1. **Audience** — that menu opens over a *running* app; a developer-onboarding page is a store-front concern.
2. **Gating** — the chrome's platform nav has no feature-flag plumbing at all (its only condition is `isModerator`), while `/apps/get-started` sits behind the `appBlocksGetStarted` **kill switch**. Mirroring it would render it unconditionally and defeat the switch.

Rather than loosen the guard silently, the excluded SET is now asserted, failing when it grows or shrinks:

```
['/apps/get-started', '/apps/submit', '/apps/invites', '/apps/revenue']
```

## Red/green matrix

Every new assertion was watched fail on pre-change code with the mutation **isolated** to one hunk. Base = `origin/main` @ `19210f04ad`; HEAD = `aec15e652b`.

**M1 — `hooks.tsx` reverted to base (two `/apps` rows), everything else at HEAD**
`vitest --project unit appsMenuEntry.test.ts` → **5 failed | 6 passed (11)**

| assertion | base | HEAD |
|---|---|---|
| exactly one `/apps` row | RED | green |
| labelled "Apps" | RED | green |
| href: `/apps` with store access, `/apps/get-started` without | RED | green |
| visible: `marketplace \|\| getStarted` across all 4 combinations | RED | green |
| the expired `newUntil: 2026-08-01` is gone | RED | green |
| *the 4 extractor/evaluator controls* | green | green | ← instrument controls, base-independent by design |
| *"the href it resolves to is one the flags admit"* | green | green | ← **INVARIANT GUARD**, not regression coverage: the old two-row shape was also internally consistent |

**M2 — only the bar-gate term reverted** (`if (!hasAppsStoreAccess(features)) return null`), the subnav row kept
`vitest --project component AppsSubNav.storeGate.browser.test.tsx` → **2 failed | 13 passed (15)**

| assertion | base gate | HEAD |
|---|---|---|
| `appBlocksGetStarted` ONLY (no store flag) → bar **and** Build apps tab render | RED | green |
| the get-started term does not enable the summary query | RED | green | ← its precondition (the bar renders) is supplied by the gate, so its red here is honest but not specific to the query |
| store access but no `appBlocksGetStarted` → bar still renders | green | green | ← **no-regression case**; it goes red if the OR is turned into an AND |
| neither → no bar | green | green | ← **invariant guard** (the gate is still a gate) |
| *the other 11* | green | green |

The 13 survivors are the point: reverting only this term killed **only** the get-started cohort's tests.

**M3 — only the `SUB_NAV_LINKS` row removed**, the widened gate kept
`vitest --project component` over the 4 AppsSubNav suites + geometry → **27 failed | 52 passed (79)**

Red included: Build apps is the first tab; its href is `/apps/get-started`; a non-author with an empty summary gets a two-tab bar; the SSR HTML contains `/apps/get-started`; every `['Build apps', …]` tab-set literal in the hydration/storeGate suites. Green: the geometry file — confirming `marketplaceTabWidth` is an **invariant guard** for this change once the selector was re-pointed (it was 122.39-vs-133.27 RED before the re-point, which is the evidence it was a real pin and not noise).

**M4 — the chrome exclusion ledger, two runs.**
Against base `AppsSubNav.tsx` it goes red on its own row-count control (`expected 7 to be >= 8`), which proves it notices but not that the *set* assertion is reachable. So a second, isolated mutation: an extra `/apps/foo` subnav row (count stays ≥ 8, chrome stays 4). Result — **1 failed | 7 passed**, failing with *this* assertion's own message: `expected [ '/apps/get-started', …(4) ] to deeply equal [ '/apps/get-started', …(3) ]`.

## Verification

Run in a clean worktree off `origin/main`, via the repo's own scripts.

- `pnpm typecheck` → **`OK — 0 type errors in 58s (heap cap 8192 MB)`**. Positive control: the same command reported **10** errors minutes earlier (an uninitialised `event-engine-common` submodule in the fresh worktree), so it can go red.
- `eslint` on the 14 changed files → **0 errors, 0 warnings**. Positive control: the same invocation with `src/utils/zod-helpers.ts` appended reports **6 warnings** on that file, so the zero is a fact about the code, not a wiring failure. (`pnpm lint` repo-wide is 358 errors / 4158 warnings — all pre-existing, none in these files.)
- `pnpm prettier:check` → `All matched files use Prettier code style!`
- `vitest --project unit src/components/{Apps,AppLayout,AppBlocks} src/pages/apps` → **122 files, 2353 tests, all passed**
- `vitest --project component` over the 8 AppsSubNav / AppsPageLayout / GetStartedBody suites → **8 files, 102 tests, all passed**
- `vitest --project component` over the 3 AppBlockChrome suites → **3 files, 39 tests, all passed**

Counts read from the runner's own summary lines, not from exit codes.

### CI

**18/18 checks pass** (1 `skipping` by design), `mergeStateStatus=CLEAN`. That includes `preview / component-tests` — the browser-mode `component` tier — on CI's own browser build, which supersedes the local caveat below as evidence for the `.browser.test.tsx` changes.

`Unit tests (1)` went **red once and passed on re-run of the identical commit**. The single failure was `src/server/services/blocks/__tests__/app-spend-cap.service.test.ts > reserveAppSpend — FAIL-CLOSED > enforces the STRICTEST ceilings when the resolver degrades` (`expected true to be false`) — a file this PR does not touch, in a subsystem it does not touch, and 32/32 green locally at HEAD. The mechanism is a **wall-clock race in that test**, not a flake diagnosis by elimination: `appSpendVelocityKey` derives its Redis key from `Math.floor(Date.now()/1000/60)`, the test issues `velocityMaxGens` (= 120) sequential reservations with **no fake timers**, and `beforeEach` clears the store, so a minute-boundary crossing anywhere inside the loop resets the counter and makes the 121st call `allowed: true` — exactly the observed assertion. Worth fixing with `vi.useFakeTimers()` (or a pinned `Date.now`) in a separate PR; deliberately not folded into a nav change.

One caveat I cannot rule out: the unit suite is sharded with vitest `--shard=N/4`, so adding a test file can reassign files between shards. If that test is order- or load-sensitive beyond the clock race, this PR could be the thing that surfaced it. The re-run passing on the same file set argues against a deterministic coupling.

⚠️ **Environment caveat on the browser tier.** In this checkout the `component`/`geometry` projects cannot launch out of the box: the nix `playwright-browsers` derivation ships `chromium_headless_shell-**1228**` while the pinned `playwright-core@1.57.0` looks for `**1200**`, and the generic-Linux binary in `~/.cache/ms-playwright` will not run on NixOS. Every browser-tier number above was produced with `PLAYWRIGHT_BROWSERS_PATH` pointed at a directory of symlinks aliasing `…-1200` → the nix `…-1228` build. The runs are real (they went red where predicted and green after the fixes), but they were **not** on the exact browser build CI uses. Same inputs in the primary clone, so this is a pre-existing toolchain drift, not something this branch introduces.

## Known, unchanged by this PR

The Marketplace tab is unconditional, so a viewer admitted by the **get-started term alone** sees a tab that `/apps` answers with `notFound`. ~~Not reachable today — `appBlocksGetStarted` is staged mod-only and a moderator holds the store flags~~ **— struck in round 4. This is the SAME live-Flipt claim round 3 withdrew in the `hooks.tsx` comment, at a site round 3's sweep did not reach; it is not observable from this repo. The consequence described below is unchanged and still real — what is withdrawn is the confidence that the cohort is empty.** — and the trigger is a **Flipt toggle, not a deploy** (`getFeatureFlags` returns the Flipt answer before it evaluates `availability`). Gating Marketplace on `hasAppsStoreAccess` is **not** the fix: it drops that viewer to one tab and the `< 2` collapse hides the whole bar, deleting the Build apps tab for the only cohort it exists for. The real fix is a get-started-aware `/apps` landing, which is a product decision. Recorded in both `AppsSubNav.tsx` and `pages/apps/get-started.tsx` rather than left for someone to rediscover.

---

# Review round 2 — `bbd668f747` + `019bc59cd4`

Everything above describes round 1 (`aec15e652b`). Sections struck through there were changed by this round.

## 🔴 The defect: the tab was ungated

`SUB_NAV_LINKS`' new row was `visible: () => true` while the whole-bar gate is an **OR**:

```ts
if (!hasAppsStoreAccess(features) && !features.appBlocksGetStarted) return null;
```

A viewer with **store access but without `appBlocksGetStarted`** passes that gate and was shown a "Build apps" tab whose page answers 404 — `resolveGetStartedAccess` returns `{ notFound: true }`, and `pages/apps/get-started.tsx:44` renders `<NotFound/>` on the same condition. Two real consequences:

1. A non-mod `app-dev-testers` member holds `appBlocks` (so `hasAppsStoreAccess`) but **not** `appBlocksGetStarted` → offered a 404. Pre-PR they had no route to that page at all; the dropdown row was correctly gated on `appsNav.getStarted`.
2. Flipping `app-blocks-get-started` **off** in Flipt — the flag's stated purpose as a kill switch — no longer removed the nav entry, because every mod holds `appBlocks`. **§4 of this PR cites exactly that hazard** as the reason to exclude the route from the app-block chrome nav. The subnav was doing the thing that argument forbids.

### The fix

`AppsNavContext` gains `canGetStarted`, derived in the container from `features.appBlocksGetStarted`; the row reads `visible: (_s, c) => c.canGetStarted`. This is the pattern already established for `isAuthor` (`isAppDeveloper(user, { appBlocksAuthor: features.appBlocksAuthor })`) — a flag threaded into the context and turned into a viewer fact, not read raw inside a predicate. `appBlocksGetStarted` is verified **not** `toggleable` (absent from `computeUserFeatureFlagsOverlay`; defined at `feature-flags.service.ts:561`), so it is SSR-frozen and hydration-safe outside the `useIsClient` deferral — the same argument §2 already makes for the bar gate. Not deferred behind `useIsClient()`.

`canGetStarted` is deliberately **not** session-scoped, unlike `isAuthor`: `resolveGetStartedAccess` reads the flag and consults no user, so folding it into the `currentUser` branch would hide the tab from a logged-out viewer the page would serve, and — Marketplace being their only other tab — the `< 2` collapse would then hide the whole bar from them. Pinned by a test in both suites; mutation M2 below.

### Resulting behaviour, each pinned

| cohort | tabs | bar |
|---|---|---|
| store access only | Marketplace | **none** (`< 2` collapse) = exactly `main` |
| `appBlocksGetStarted` only | Build apps + Marketplace | renders |
| both | Build apps + Marketplace (+ capability/summary tabs) | renders |
| logged out, no flag | Marketplace | none |
| logged out, WITH flag | Build apps + Marketplace | renders |

## Red/green matrix — round 2

Every mutation isolated to the narrowest expression that can be wrong. Base for M1/M2/M5 = the round-1 HEAD `aec15e652b`'s expression, applied to the round-2 tree so only that expression differs.

**M1 — the pre-fix predicate: `visible: (_s, c) => c.canGetStarted` → `visible: () => true`**
`vitest --project component` over the 4 AppsSubNav suites → **13 failed | 69 passed (82)**. Each on its own assertion:

| test | message |
|---|---|
| `browser` › a non-author with an EMPTY summary and no get-started renders NOTHING | `expected [ <a …(12)>…(2)</a> ] to have a length of +0 but got 1` |
| `browser` › Build apps is ABSENT for a store-visible viewer WITHOUT the flag | `expected [ <a …(11)>…(2)</a> ] to have a length of +0 but got 1` |
| `browser` › a summary flag alone brings the bar back (one install, no get-started) | `expected […] to have a length of 2 but got 3` |
| `browser` › the author capability alone brings it back too | `expected […] to have a length of 2 but got 3` |
| `browser` › a NON-author invitee gets no sub-nav at all | `expected [ <nav …> ] to have a length of +0 but got 1` |
| `hydration` › a logged-out viewer never gets Create | `expected [ 'Build apps', 'Marketplace' ] to deeply equal []` |
| `hydration` › a non-author with an empty summary renders no nav at all | `expected [ <nav …> ] to have a length of +0 but got 1` |
| `hydration` › one install is enough to bring the bar back | `expected [ 'Build apps', 'Marketplace', …(1) ] to deeply equal [ 'Marketplace', 'Installed' ]` |
| `storeGate` › BOTH flags true → renders (today's live cohort) | `expected [ 'Build apps', …(6) ] to deeply equal [ 'Marketplace', 'Create', …(4) ]` |
| `storeGate` › POSITIVE CONTROL | `expected [ 'Build apps', …(2) ] to deeply equal [ 'Marketplace', 'Create' ]` |
| `storeGate` › appListings-only NON-author, no get-started → no bar at all | `expected [ 'Build apps', 'Marketplace' ] to deeply equal []` |
| `storeGate` › a logged-out viewer with the store flag lit gets no bar | `expected [ 'Build apps', 'Marketplace' ] to deeply equal []` |
| `storeGate` › store access but NO appBlocksGetStarted → a bar with NO Build apps tab | `expected [ 'Build apps', …(2) ] to deeply equal [ 'Marketplace', 'Create' ]` |

The `ssrHydration` suite stayed green, correctly: every cohort in it holds the flag, so `() => true` produces identical output. It is not the guard for this.

**M2 — session-scope the derivation: `canGetStarted: !!features.appBlocksGetStarted` → `currentUser ? … : false`**
→ **2 failed | 37 passed (39)**

| test | message |
|---|---|
| `hydration` › a logged-out viewer WITH appBlocksGetStarted does get Build apps (positive control) | `Cannot find element with locator: getByRole('tab', { name: 'Build apps' })` |
| `ssrHydration` › logged-out: no SESSION-scoped tab in the server HTML | `expected '<style …' to contain '/apps/get-started'` |

**M5 — delete `if (links.length < 2) return null;`**
→ **6 failed | 71 passed (77)**, confirming the collapse tests are covering a live cohort again rather than a guard nothing exercises. Reds include `storeGate › appListings-only NON-author, no get-started → no bar at all` (`expected [ 'Marketplace' ] to deeply equal []`).

**Geometry diagnostic control** — mocking `appBlocksAuthor: false` in `AppsPageLayout.geometry.browser.test.tsx` (so the bar collapses) now fails with the intended diagnosis on all 3 tests instead of a bare `TypeError`:

```
`nav[aria-label="App sections"]` is not in the document. The geometry here is measured off the
`/apps` sub-nav, which renders only for a viewer with ≥2 qualifying tabs — check the
`useFeatureFlags` / `useCurrentUser` mocks above before re-baselining a number.
```

## 🟡 The chrome ledger's control masked its own assertion

`chromeNavAlignsWithSubNav.test.ts` pinned its parse controls **on the live counts** (`expect(inChrome.size, 'parsed no items out of the chrome platform nav').toBe(4)`, `expect(subnav.length).toBeGreaterThanOrEqual(8)` with 8 rows live). Adding a chrome entry therefore failed on the *control* — `parsed no items out of the chrome platform nav: expected 5 to be 4`, a message that **states the opposite of what happened** — and the `toEqual` ledger never ran. The ledger's SHRINK direction was unproven; it could only fire if an item were added AND another removed in one change.

~~All three count controls (`:289`, `:290`, `:344`, `:345`, `:414`) are now `toBeGreaterThanOrEqual(1)` — a real parse control — and the `toEqual` ledgers own their sets in both directions.~~ **Half withdrawn in round 3 — this was the round-2 defect.** Two of the three relaxations were right; the third relaxed a floor whose set *nothing* owned, deleting a live detection. Fixed below in **Review round 3**. (The cited line numbers were also stale: at `019bc59cd4` the three floors are at `:293`, `:294`, `:356`, `:357`, `:431`.) The two ledgers named in the table below were and remain correct — re-confirmed in round 3. Re-mutated:

| direction | mutation | fails on | message |
|---|---|---|---|
| **GROW** | add a `SUB_NAV_LINKS` row (`/apps/news`) | the exclusion ledger's own `toEqual` | `a store subnav destination is missing from the app-block chrome…: expected [ '/apps/get-started', …(4) ] to deeply equal [ '/apps/get-started', …(3) ]` — `+ "/apps/news"` |
| **SHRINK** | add `<ChromeSurfaceItem href="/apps/get-started">Build apps</ChromeSurfaceItem>` | the exclusion ledger's own `toEqual` | `…expected [ '/apps/submit', …(2) ] to deeply equal [ '/apps/get-started', …(3) ]` — `- "/apps/get-started"` |

SHRINK also (correctly, independently) reds the sibling `the four shared destinations are the expected ones` ledger. Before the fix it died only on the control.

## 🟡 Stale comments the round-1 diff invalidated

`pages/apps/[appBlockId]/edit.tsx` and `pages/apps/listing/[appListingId]/edit.tsx` carried byte-identical notes ("Only 'Marketplace' is unconditional… gets an EMPTY chrome band") both pointing at `pages/apps/get-started.tsx`, whose note round 1 rewrote to say the empty-band case was closed — so chasing the pointer found a contradiction. Repaired **against the code**, and the round-2 fix makes most of both notes true again: the empty-band case is closed for `/apps/get-started` and still open for the other `/apps/*` pages, and the tab enumeration now includes the get-started capability.

**Sweep** (whole tree, `command grep -rn` so `.gitignore` is not honoured, plus a positive control that the pattern matches the two known sites): `"EMPTY chrome band"` → 2 sites, both fixed. `"is unconditional"` → 34 hits, 4 in this subsystem (2 edit pages fixed, `get-started.tsx` still correct about Marketplace, `IframeHost.tsx` correct about the chrome). `"Build apps"`, `"always-on"`, `"two unconditional"`, `"no longer reachable"`, `"collapse floor"` across `.ts/.tsx/.md/.mdx` including `docs/` and `claudedocs/` → no further sites. Also updated: `IframeHost.tsx` §4 (notes that the subnav's row *does* read the kill switch, which is what makes the exclusion a difference between surfaces rather than an inconsistency) and the ledger's own block comment.

**One site deliberately left**: `server/routers/blocks.router.ts:2607` says the sub-nav "degrades to just the always-on tabs" (plural). That sentence was already loose before this PR — it is a claim about the summary query, not about the flags, and this PR's diff did not invalidate it. Flagging rather than silently rewording pre-existing prose.

## 🟢 Small ones

- `AppsPageLayout.geometry.browser.test.tsx`: `tabWidth` was computed and asserted nowhere — removed, with a note saying why, so a reader does not believe the first-tab width is still pinned. `marketplaceTabWidth` remains the inline-axis pin. All element lookups now go through a `required()` helper that fails with the diagnosis above rather than an anonymous `TypeError`. The file header note ("an anonymous / non-author viewer would render NO `<nav>` at all") is **true again** after the round-2 fix, and is rewritten from the post-fix code: Marketplace is the only unconditional row, the author capability supplies the second tab, and `appBlocksGetStarted` is deliberately unset so the rendered bar is identical to the one the 133.27px baseline came from.
- `tests/preview-apps-marketplace.spec.ts:51-52` documented "first tab is Marketplace (AppsSubNav.tsx:55)" — both halves wrong, assertion unaffected. It now names the row it keys on, notes that "Build apps" precedes it for any viewer holding `appBlocksGetStarted` (which the spec's `mod` role does), and says the assertion selects by name so order is irrelevant.

## 🟢 Judgement call: the dropdown row's label for the get-started-only cohort

**Decision: accept the mismatch, document it, do not branch the label.** For a viewer with `appBlocksGetStarted` and no store flag the single dropdown row reads **"Apps"** with `IconPlugConnected` while navigating to `/apps/get-started` (builder onboarding). Reasons, now recorded at the call site in `hooks.tsx`:

1. **"Apps" names the product area**, which both destinations are inside. The row is a door into `/apps*`, not a description of the landing page.
2. Branching the label *and* the icon (`IconPlugConnected` vs the subnav's `IconCode`) re-splits one row into two presentations — the duplication this consolidation exists to remove — and would need the machine-checked `appsMenuEntry.test.ts` label ledger updated for a cohort nobody is in.
3. ~~**The cohort is empty today.** `appBlocksGetStarted` is `availability: ['mod']` and every moderator holds a store flag, so no real viewer takes this branch; a conditional label would be copy nobody could see in production.~~ **Withdrawn in round 3 — and round 3's sweep struck the ledger sentence and the stale line citations in this section and missed THIS one. Struck in round 4.** "Every moderator holds a store flag" is a claim about **live Flipt**, which this repo cannot settle: `getFeatureFlags` returns Flipt's answer *before* it evaluates `availability`, so `['mod']` is the Flipt-DOWN fallback, not the live population. The **decision** (accept the label mismatch, do not branch it) still stands on reasons 1 and 2; the *ground* offered here does not.

~~**Re-decide it when that stops being true** — the trigger is `app-blocks-get-started` being widened past `['mod']`, in Flipt or in `feature-flags.service.ts`. That trigger is written into the comment beside the ternary, not left in a review thread.~~ **Both halves withdrawn.** (a) The trigger is **bidirectional** — the condition is `!marketplace && getStarted`, so the store flags *narrowing* reaches it too (round 3, below). (b) The final clause is now flatly false: the comment beside the ternary no longer says what this sentence says it says. Round 3 rewrote it to name both directions; round 4 rewrote it again to name **all three** store flags behind `marketplace` (`app-listings` ‖ `app-blocks-enabled` ‖ `app-listings-public-external`) and to scope the `availability` claim. **Read the trigger in `hooks.tsx`, not here.**

## Verification — round 2

Repo's own scripts, in a clean worktree off the PR branch.

- **Merged-tree check.** `origin/main` moved (`f311298e13`, #4666 touches `IframeHost.tsx` in disjoint regions). A worktree off `origin/main` with this branch merged: `node scripts/typecheck.mjs` → **`OK — 0 type errors in 59s`**; `vitest --project component src/components/Apps src/components/AppBlocks` → **123 files, 1720 tests, all passed**; full `vitest --project unit` → **1644 files passed, 38295 tests passed**, with **6 files / 1 test failing**. Those 6 are **identical at plain `origin/main`** with this branch un-merged (measured — same run, same 6 files, same 1 assertion): `pending-review-mute`, `creator-shop-rejected-terminal`, `user-settings-race.behavior`, `feature-flags.early-adopter.seam`, `restriction-type-seam`, `moderator-restriction-vocabulary` — all `apps/moderator`-adjacent and unrelated to this diff.
- `node scripts/typecheck.mjs` on the branch worktree → 0 errors in `src/`. (It also reported 31 `packages/**` "Cannot find module 'kysely'" errors, which are an artifact of a stale generated Prisma/Kysely client in that worktree — running `scripts/generate-slim-schema.js` + `prisma generate` in the merged worktree takes the count to a clean 0, which is the number to read.)
- `eslint` on the 12 changed `src/` files → **0 problems**. Positive control: the same invocation on `src/utils/zod-helpers.ts` reports **6 problems**, so the zero is a fact about the code. (Array-quoted `"${FILES[@]}"`, and the file count was read back as 12 — a space-separated `$FILES` in zsh checks zero files and prints a confident pass.)
- `prettier --check` on the 13 changed files → `All matched files use Prettier code style!`. Negative control: the same invocation on a deliberately-misformatted scratch file reports `Code style issues found`.
- `vitest --project unit src/components/{Apps,AppBlocks,AppLayout}` → **122 files, 2353 tests, all passed**.
- `vitest --project component src/components/Apps src/components/AppBlocks` → **121 files, 1687 tests, all passed** (branch worktree).

Counts read from the runner's own summary lines, never from an exit code.

⚠️ **Browser-tier caveat, updated.** The nix `playwright-browsers` derivation ships `chromium_headless_shell-1228` while the pinned `playwright-core@1.57.0` wants `1200`. Every browser-tier number above was produced with `PLAYWRIGHT_BROWSERS_PATH` pointed at a directory of symlinks aliasing `…-1200` → the nix `…-1228` build, in a worktree with a real `pnpm install` (a `node_modules` symlink to the primary clone makes vitest fail to find its runner) and a generated Prisma client. The runs are real — they went red exactly where predicted in M1/M2/M5 and green after — but they are **not** on the browser build CI uses. `preview / component-tests` and `Geometry tests` on CI remain the authority for that tier.

### CI — round 2

At `019bc59cd4`: **20 checks, all terminal — 19 `SUCCESS`, 1 `SKIPPED`** (`Typecheck (main pushes / fork PRs / non-main base)`, skipped by design), `mergeStateStatus` `CLEAN`, no failures and no re-runs. That includes **`preview / component-tests`** and **`Geometry tests`** — the two browser-mode tiers — on CI's own browser build, which supersedes the local `PLAYWRIGHT_BROWSERS_PATH` caveat above as evidence for the `.browser.test.tsx` changes. Also green: `App unit tests + typecheck`, `Unit tests (1..4)`, `tekton / typecheck`, `tekton / fixture-bootstrap`, `preview / auth-tests`, `preview / render-check`, `preview / deploy`, `ESLint + Prettier (changed files)`, `Package unit tests`, `Schema drift gate`, `event-engine-common pin`, both `Windows dev-env` jobs.

The `app-spend-cap.service.test.ts` wall-clock race described in round 1's CI note did **not** recur on this run.

---

# Review round 3 — `684be6cc8f` + `136fffd647`

Round 2 relaxed three count controls in `chromeNavAlignsWithSubNav.test.ts` from exact counts to `toBeGreaterThanOrEqual(1)`. Two of those were right. The third was not, and the round-2 note above claiming otherwise is struck through.

## 🟡 The relaxed floor deleted a live detection

For the two floors in `every platform-nav destination…` and `the subnav rows deliberately ABSENT…`, relaxing is safe: something else owns their sets outright (the exact 4-item platform nav; the exact excluded set). **Nothing owned `parseAllChromeLinks`' set.** Its comment said *"Parse control, not a ledger — same reasoning as the two above"*, and that reasoning does not transfer.

Rules (a) and (b) in that test are **per-link**, so a new chrome item pointing at a route `SUB_NAV_LINKS` already carries, wearing that row's own glyph, satisfies both. Measured on `019bc59cd4`:

| tree | mutation | result |
|---|---|---|
| `019bc59cd4` | add `<ChromeSurfaceItem href="/apps/get-started" leftSection={<IconCode …/>}>Build apps</ChromeSurfaceItem>` to the **⋮ overflow** | **8 passed — mutant SURVIVED** |

Not bookkeeping. The **DELIBERATE SUBSET** note above the platform nav argues that a surface which does not honour a kill switch must not advertise the route it switches off — ~~this chrome has no feature-flag plumbing (its only condition anywhere is `isModerator`)~~ **— corrected in round 4: that is true of the platform-nav SLICE and false of the surface (d) governs. `chromeBody()` spans `AppBlockChrome` through `ChromeDesktopLeadingGroup`, which renders `<ChromeReviewMenuItem>` gating on `hasAppsStoreAccess(useOptionalFeatureFlags())` and again through `useCanReviewListing` → `resolveClientStoreScope`. What IS true, and is what the argument needs: ~~every literal-href item in this chrome renders unconditionally except the moderator-gated `/apps/review`~~ — corrected again in round 5, see below: **no literal-href item in this chrome is gated by a FEATURE FLAG** except the moderator-gated `/apps/review`. Two ARE conditional — the compact back chevron under `compact`, the breadcrumb crumb under `isPage` — but on LAYOUT, not on a flag, which is what the argument turns on.** A maintainer adds "Build apps" to the ⋮ menu, the guard stays green, `app-blocks-get-started` is later switched off, and the chrome keeps advertising a page that answers `notFound`. The overflow is the same surface as the platform nav; only the enumeration stopped short of it.

**Fix:** a `toEqual` ledger — **(d)** — over ~~the whole literal-href set~~ **the literal-href set its parser could see — which round 4 measured to be a strict SUBSET of the chrome's actual link set; see round 4 below** — placed *after* (a)/(b)/(c) so the more specific rules keep their own messages. Sorted, so a re-order cannot report a route change that did not happen; duplicates kept, since `/apps/installed` legitimately appears twice and a Set would hide a third item hung on an already-listed route. `toBe(5)` is deliberately **not** restored — that is the misdirection round 2 correctly removed (a maintainer adding a legitimate destination was told the scanner had broken).

The floor's comment now says what is true: this one has **no sibling `toEqual` to inherit its set from**, and points at (d).

### Mutation matrix

Each mutation isolated to one hunk. Runner: `vitest --project unit chromeNavAlignsWithSubNav.test.ts` (8 tests). Baseline both sides: **8 passed**.

| # | mutation | pre-fix | post-fix | fails on / message |
|---|---|---|---|---|
| **A** | **GROW outside every enumerated region** — add `/apps/get-started` "Build apps" to the ⋮ overflow | **8 passed — SURVIVED** | **1 failed \| 7 passed** | **(d)'s own message** — `the set of routes the app-block chrome links to has changed…: expected [ '/apps', '/apps/get-started', …(4) ] to deeply equal [ '/apps', '/apps/installed', …(3) ]`. (a) and (b) both pass, so (d) is the **only** thing that catches it. |
| **B** | **SHRINK** — delete the platform-nav `Review` item | — | **4 failed \| 4 passed** | (d) fires **with its own message**: `…expected [ '/apps', '/apps/installed', …(2) ] to deeply equal [ '/apps', '/apps/installed', …(3) ]`. Three other tests also fire, correctly — a platform-nav item is owned by them too. |
| **C** | **SHRINK** — delete the ⋮ overflow's `Manage apps` | — | **1 failed \| 7 passed** | Dies on **(c)**, which precedes (d) in the same test. Recorded rather than engineered around: **every existing link is already owned by some assertion for SHRINK**, so (d)'s unique contribution is GROW outside the platform-nav slice — which is exactly the gap A demonstrates. |

**Re-confirmed — the two ledgers that were already correct still die on their own `toEqual`, no regression:**

| ledger | direction | mutation | result |
|---|---|---|---|
| exclusion ledger | **GROW** | add a `SUB_NAV_LINKS` row (`/apps/changelog`) | **1 failed \| 7 passed** — `a store subnav destination is missing from the app-block chrome…: expected [ '/apps/get-started', …(4) ] to deeply equal [ '/apps/get-started', …(3) ]` |
| exclusion ledger | **SHRINK** | add `/apps/get-started` to the **platform nav** | fires — `…expected [ '/apps/submit', …(2) ] to deeply equal [ '/apps/get-started', …(3) ]` |
| `expected glyphs` | **GROW** | same mutant | fires — `expected [ { href: '/apps', …(2) }, …(4) ] to deeply equal [ …, …(3) ]` |
| `expected glyphs` | **SHRINK** | delete `Review` | fires — `expected [ { href: '/apps', …(2) }, …(2) ] to deeply equal [ …, …(3) ]` |

## 🟢 The re-decide trigger was narrower than the branch's reachability

`hooks.tsx` recorded the trigger as "`app-blocks-get-started` widened past `['mod']`". The branch is `!marketplace && getStarted`, so **either side moving reaches it** — the store flags *narrowing* (`app-blocks-enabled` / `app-listings` turned off in Flipt while get-started stays on) puts **every moderator** on a row labelled "Apps" with `IconPlugConnected` navigating to developer onboarding. Both directions are now named.

The stated ground was also a live-Flipt claim presented as if derived from `availability`. This PR's own `pages/apps/get-started.tsx` establishes that `getFeatureFlags` returns the Flipt answer **before** evaluating `availability` ("Flipt overrides role checks (both enable AND disable)"), so `availability` is only the Flipt-**down** fallback. The comment now says the cohort's emptiness is observable **only in live Flipt** and that the repo cannot settle it — what the code *does* establish is that all four App-Blocks flags are Flipt-backed runtime toggles, so the branch is reachable with no PR and no deploy (verified: `feature-flags.service.ts:510, 520, 561, 571`).

Consequence is cosmetic; the defect was the decision record. Comment-only, no behaviour change.

## 🟢 Audit of the verification counts above — one correction, and one challenge that did not hold

The round-2 figure **"123 files / 1720 tests"** was challenged as understating the `component` tier, on the ground that the `component` project's include (`src/**/*.browser.test.tsx`) matches **217 files / 2403 tests**. **Re-measured, and the challenge does not hold — the two numbers are different populations, and both are right:**

| claim | command | measured | verdict |
|---|---|---|---|
| 123 files (merged tree) | `vitest --project component src/components/Apps src/components/AppBlocks` | **123** — `git ls-tree` of the merge of `f311298e13` + branch, filtered to `*.browser.test.tsx` under those two dirs. Still **123** against today's `origin/main` (`d6e5c4eec0`). | ✅ reproduces exactly |
| 121 files / 1687 tests (branch) | same command, branch worktree | **121 files, 1687 tests, all passed** — run in full | ✅ reproduces exactly |
| — | `vitest --project component` **unfiltered** | **217 files, 2403 tests, all passed** — run in full at this branch | ✅ correct, but a *different* population: the body never claimed a whole-tier count, and the command carrying the path filter is printed next to the number |
| 122 files / 2353 tests | `vitest --project unit src/components/{Apps,AppBlocks,AppLayout}` | **122 files, 2353 tests, all passed** — run in full | ✅ reproduces exactly |
| 12 / 13 / 14 changed files | `git diff --name-only` | round 2 = 12 `src/` of 13 total; round 1 = 14 | ✅ all three reproduce |
| 6 eslint problems on `zod-helpers.ts` | positive control | **6 problems (0 errors, 6 warnings)** | ✅ reproduces |

**The one number that does not reproduce** is the line-number citation `:289, :290, :344, :345, :414` in the round-2 section — at `019bc59cd4` the three floors sit at **`:293`, `:294`, `:356`, `:357`, `:431`**. Struck through above.

**Not re-measured:** the *1720 test count* for the merged tree (the file-count half reproduces exactly; re-running the test half needs a fully installed merged worktree, which I did not build). Stated rather than restated as if measured.

## Round-3 verification

Instruments validated before every zero.

- `vitest --project unit chromeNavAlignsWithSubNav.test.ts` → **8 passed**; with the whole `AppLayout` tree → **32 passed**.
- `vitest --project component` (unfiltered, real browser) → **217 files, 2403 tests, all passed**.
- `node scripts/typecheck.mjs` → **`OK — 0 type errors in 81s`**. Positive control: the *identical command* minutes earlier reported **2848 errors**, off a Prisma client generated **2025-12-08**; `prisma generate` under the flake's `PRISMA_*` engine paths takes it to 0. None of the 2848 named either changed file — the instrument can go red, and the zero is a fact about the code.
- `eslint` on both changed files → **0**. Positive control: appending `src/utils/zod-helpers.ts` reports **6 problems**. Array-quoted, file count read back as **2** (a space-separated `$FILES` in zsh checks zero files and prints a confident pass).
- `prettier --check` → clean. Negative control: a deliberately misformatted scratch file reports `Code style issues found`.
- **Merged-tree check.** `origin/main` moved to `d6e5c4eec0`, and #4666 touches **`IframeHost.tsx` — the very file the new ledger scans**, so this was not a formality. The guard including (d) was run against the *merged* `IframeHost.tsx` + `AppsSubNav.tsx`: **8 passed**. `gh pr view` reports `MERGEABLE` / `CLEAN`.

⚠️ Browser-tier caveat unchanged from round 2: the nix `playwright-browsers` derivation ships `chromium_headless_shell-1228` while the pinned `playwright-core@1.57.0` wants `1200`, so the 217-file run used `PLAYWRIGHT_BROWSERS_PATH` aliasing `…-1200` → the nix `…-1228` build. Real runs, not CI's exact browser build. `preview / component-tests` and `Geometry tests` remain the authority for that tier.

---

# Review round 4 — `474bf91fc6`

Round 3 added the **(d)** ledger and claimed it owned "the whole literal-href set". It did not — its parser owned a strict subset, and the message it printed described the wider set. Two ordinary element shapes walked through it with all 8 tests green.

## 🟡 (d)'s message claimed the chrome's route set; the parser owned a subset

`parseAllChromeLinks` matched `<ChromeSurfaceItem …>text</ChromeSurfaceItem>` and then dropped anything lacking **both** a literal `href="…"` **and** a `leftSection={<IconX`. Measured on `136fffd647`, both mutants **SURVIVED — 8 passed**:

| # | mutation (added to the ⋮ overflow) | why it is not exotic |
|---|---|---|
| **P1** | `<ActionIcon component={Link} href="/apps/get-started" …>` | this is the element shape the chrome **already uses** for its `/apps` back-link (`IframeHost.tsx`, compact-mode chevron) |
| **P2** | `<ChromeSurfaceItem href="/apps/get-started">Build apps</ChromeSurfaceItem>` with no `leftSection` | `ChromeSurface.tsx` types it `leftSection?: ReactNode`, so it compiles — and it is the very element (d) claims to enumerate |

Either one is a door out of a running app into a route the `app-blocks-get-started` kill switch can turn off, offered unconditionally, with the guard green. Verbatim the hazard (d)'s own message names.

**Option taken: WIDEN the parser** (the preferred one), not narrow the message. Narrowing would have bought accuracy by shrinking coverage, and the coverage is the point — the sentence is right, so the parser was made as wide as the sentence.

- It scans **tags**, not one tag name: any element carrying a literal `href="…"`.
- `leftSection` is **optional metadata**, not a condition of inclusion.
- An opening tag ends at the first `>` at **brace depth 0**, and scanning resumes after it — so a nested `<Icon…/>` inside the attribute region is never opened as an element of its own.
- `href` is matched against the attribute list with every `{…}` expression **stripped**, which is what keeps `href={`/apps/run/${r.blockId}`}` out (not a platform route) and stops a nested `prop={<Foo href="…"/>}` being attributed to the outer tag.
- `label` falls back to `aria-label` (the back chevron has no text at all, and a null label makes (a)/(b)'s messages name nothing).

**What that widening COSTS, stated where it is paid.** The set grows from 5 to **7**: the compact back chevron and the breadcrumb crumb, both `/apps`, both `leftSection`-less. Rule **(b)** (same route ⇒ same glyph) is therefore now scoped to links that HAVE a `leftSection` glyph, and says so at the call site — an element with no glyph cannot be drawing the route with the *wrong* one, and its presence in the chrome is still owned outright by (d). `icon` is deliberately the `leftSection` glyph **only**, never "any Icon inside the element": the back chevron renders `<IconChevronLeft/>` as its child, which is a directional affordance and not this route's glyph, and scoring it as one would make (b) red on correct code.

Both shapes are pinned as **parser fixtures** in the positive-control test, not merely described — a widened parser that silently stopped matching them again would restore the hole while looking identical. A negative control for the expression-stripping is pinned alongside.

### Mutation matrix

Both guards run **in the same vitest invocation**: `chromeNavAlignsWithSubNav.test.ts` (round 4) and a byte copy of the `136fffd647` version. 16 tests, baseline **16 passed**. Each mutation isolated to one hunk; sources restored by file copy between runs.

| # | mutation | pre-fix (`136fffd647`) | post-fix | fails on / message |
|---|---|---|---|---|
| **P1** | `<ActionIcon component={Link} href="/apps/get-started" aria-label="Build apps">` in the ⋮ overflow | **8 passed — SURVIVED** | **1 failed \| 7 passed** | **(d)'s own message** — `the set of routes the app-block chrome links to has changed…: expected [ '/apps', '/apps', '/apps', …(5) ] to deeply equal [ '/apps', '/apps', '/apps', …(4) ]`. (a), (b) and (c) all pass, so (d) is the **only** thing that catches it. |
| **P2** | `<ChromeSurfaceItem href="/apps/get-started">Build apps</ChromeSurfaceItem>` (no `leftSection`) in the ⋮ overflow | **8 passed — SURVIVED** | **1 failed \| 7 passed** | **(d)'s own message**, same diff. (a)/(b)/(c) pass. |

**No-regression re-confirmations** — every case round 3 recorded, re-measured on the round-4 tree:

| # | mutation | result | fails on |
|---|---|---|---|
| **A** | ⋮-overflow **GROW**: `<ChromeSurfaceItem href="/apps/get-started" leftSection={<IconCode …/>}>Build apps</…>` | 1 failed \| 7 passed, **both** trees | **(d)'s own message**; (a)/(b)/(c) pass — unchanged from round 3 |
| **B** | **SHRINK** the platform-nav `Review` item | 4 failed \| 4 passed | **(d)'s own message** — `…expected [ '/apps', '/apps', '/apps', …(3) ] to deeply equal [ …(4) ]`; the exclusion ledger, `expected glyphs` and the label test also fire, correctly |
| **C** | **SHRINK** the ⋮ overflow's `Manage apps` | 1 failed \| 7 passed | **(c)** — `the chrome should still offer BOTH /apps/installed entries…: expected [ 'Installed apps' ] to deeply equal [ 'Installed apps', 'Manage apps' ]` |
| — | exclusion ledger **GROW**: add a `SUB_NAV_LINKS` row (`/apps/changelog`) | 1 failed \| 7 passed | its own `toEqual` — `a store subnav destination is missing from the app-block chrome…: expected [ '/apps/changelog', …] to deeply equal […]` |
| — | exclusion ledger **SHRINK**: add `/apps/get-started` to the **platform nav** | 3 failed \| 5 passed | its own `toEqual` — `…expected [ '/apps/submit', …(2) ] to deeply equal [ '/apps/get-started', …(3) ]` |
| — | `expected glyphs` **GROW** | same mutant | fires — `expected [ { href: '/apps', …(2) }, …(4) ] to deeply equal [ …, …(3) ]` |
| — | `expected glyphs` **SHRINK** | mutation **B** | fires — `expected [ { href: '/apps', …(2) }, …(2) ] to deeply equal [ …, …(3) ]` |

## 🟡 A claim that was true-and-scoped and got restated unscoped

`(d)`'s message and block comment asserted *"this surface has no feature-flag plumbing at all — the only condition anywhere on it is `isModerator`"*. At `019bc59cd4` that sentence was scoped to the **platform nav**, where it holds. Round 3 restated it for (d), which is deliberately the **whole-chrome** rule — and there it is false.

`chromeBody()` spans `AppBlockChrome` through `ChromeDesktopLeadingGroup`, which renders `<ChromeReviewMenuItem>` (`IframeHost.tsx`). That item calls `useOptionalFeatureFlags()` and returns `null` without `hasAppsStoreAccess(features)` (`ChromeReviewEntry.tsx`), then narrows again through `useCanReviewListing` → `resolveClientStoreScope(features)` (`ReviewListingButton.tsx`). Other conditions on the surface: `{recentApps.length > 0 && …}`, `{appBlockId && …}`, `{showHide && …}`.

This is not a nit — the message is the argument a maintainer is handed when (d) goes red, and it told them the surface *cannot* honour a kill switch, foreclosing the option the repo demonstrates a few lines away. **Rewritten from the code**, and it now says what is true of the surface (d) governs:

> NO literal-href item in this chrome is **gated by a FEATURE FLAG** except the moderator-gated `/apps/review`, so a flag-gated destination added here **as a plain link** keeps being offered after its flag goes down … The surface **CAN** read flags — `ChromeReviewMenuItem` gates itself on `hasAppsStoreAccess(useOptionalFeatureFlags())` — so "add it GATED, the way that item is" is a real third option alongside excluding it; adding it ungated is not.

Note what changed and what did not: the **exclusion** of `/apps/get-started` still stands, and the DELIBERATE SUBSET note it rests on is *scoped to the platform nav*, where it remains accurate. What is corrected is the whole-chrome generalisation of it.

## 🟢 Two of three store flags named, and an over-wide `availability` sentence

`hooks.tsx` said the store-flags-narrowing trigger was *"`app-blocks-enabled` / `app-listings` turned off in Flipt … Then EVERY moderator takes this branch."* `marketplace` is `hasAppsStoreAccess` = `appListings || appBlocks || **appListingsPublicExternal**` (`shared/utils/app-blocks-access.ts`) — three flags, and the sibling `appsNavVisibility.ts` spells all three, twice. A mod holding `app-listings-public-external` keeps `marketplace === true` and does **not** take the branch, so the trigger was under-specified in exactly the direction round 3 added. All three are now named, with the "any ONE of them is enough to stay out of this branch" consequence spelled out.

Same paragraph: *"`availability` is only the Flipt-DOWN fallback"* is over-wide. `availability` also carries env / region / **server-colour** terms that run **before** Flipt and that Flipt cannot override (`feature-flags.service.ts` — *"Server/domain restrictions always apply"*). It happens to be accurate for a `['mod']` flag, which is the case at hand, so the sentence is **scoped** rather than deleted, with an explicit "do not generalise this to a flag that names a colour domain".

Comment-only. No behaviour change in `hooks.tsx`.

## 🟢 PR-body sweep — the site a reader hits first, three rounds running

Round 3 struck the round-2 ledger sentence and the stale line citations in its own section and **left point 3 and the re-decide sentence unstruck** — text a reviewer reading top-to-bottom takes as current. Both are struck above.

The sweep was then run over the **whole body** as a pattern battery with controls, not by eye: positive control `Marketplace` → **22 hits** (the pattern can match), negative control `zzz-no-such-string` → **0** (the zero means something). Eight retraction signatures, every hit read:

| pattern | hits | disposition |
|---|---|---|
| `feature-flag plumbing` / `only condition` / `isModerator` | 2 | `:58` **left** — scoped to "the chrome's platform nav", where it is true. `:285` **corrected** (whole-chrome claim). |
| `cohort is empty` / `empty today` | 1 | `:243` **struck** |
| `widened past` / `staged mod-only` | 3 | `:245` **struck**; `:129` **struck**; `:312` is round 3's own correction, current |
| `written into the comment` | 1 | `:245` **struck** |
| `whole literal-href set` | 1 | `:287` **corrected** |
| `moderator holds the store flags` | 2 | `:243`, `:129` — both **struck** |
| check counts | 2 | `:264` verified correct (20/19/1 at `019bc59cd4`); `:119` is a round-1 count against a round-1 commit — **not re-derivable now**, flagged rather than silently rewritten |

**The one the sweep found that the brief did not name:** `:129`, in round 1's *"Known, unchanged by this PR"* — *"Not reachable today — `appBlocksGetStarted` is staged mod-only and a moderator holds the store flags"*. That is the same live-Flipt claim round 3 withdrew inside the `hooks.tsx` comment, at a body site round 3's sweep never reached. Struck. The *consequence* recorded there is unchanged and still real; the confidence that nobody is in the cohort is not.

## 🟢 Correction to round 3's claim 11

Round 3's review comment recorded *"CI at `136fffd647` → 19 checks terminal, 18 SUCCESS + 1 SKIPPED"*. Re-read: the rollup has **20 entries — 19 `SUCCESS`, 1 `SKIPPED`** (`Typecheck (main pushes / fork PRs / non-main base)`, skipped by design). Wrong in the safe direction, but it is a count in prose. Corrected here and in that comment.

## Round-4 verification

Instruments validated before every zero.

- **`node scripts/typecheck.mjs` → `OK — 0 type errors`. 🔴 That number is NOT evidence about the changed test file, and the negative control is what proved it.** A deliberate `const NEGATIVE_CONTROL: number = src;` planted in `chromeNavAlignsWithSubNav.test.ts` was reported as **0 errors, exit 0** — the root `tsconfig.json` carries `src/**/__tests__/**` in its `exclude`, so that command is **structurally blind** to every file under a `__tests__/` directory. The repo already knows this and ships the instrument for it: `scripts/ci/typecheck-tests-gate.mjs` / `tsconfig.tests.json`.
- **`node scripts/typecheck.mjs -p tsconfig.tests.json` → 1160 errors across the test tree, `0` of them in either changed file.** Negative control on the *same* command with the same planted error → **1161**, the extra one being `src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts(261,9): error TS2322`. So the instrument can see this file, and the zero for it is a fact about the code.
  ⚠️ The 1160 are **pre-existing and in unrelated subsystems** (`blocks.router.workflow`, `apps-shared.router`, `publish-request.reviewSandbox`, `database-target`, …) — real drift from other merged work, arithmetically not attributable to this diff (my two files contribute 0). `scripts/ci/typecheck-tests-gate.mjs` is red on this tree for that reason. **Flagged, not fixed — out of scope for a nav PR**, and CI is green on the branch.
- `vitest --project unit chromeNavAlignsWithSubNav.test.ts` → **8 passed**; `--project unit src/components/{AppBlocks,Apps,AppLayout}` → **122 files, 2353 tests, all passed** (reproduces round 2 and round 3 exactly).
- `eslint` on both changed files → **0 problems**; file count read back as **2** (bash array, not a space-separated `$FILES`, which in zsh checks zero files and prints a confident pass). Positive control: appending `src/utils/zod-helpers.ts` → **6 problems**.
- `prettier --check` → clean. **Negative control is free here: the first run of the round-4 file FAILED** (`Code style issues found`) and was fixed with `--write`, so this instrument was watched red on this very change.
- **Merged-tree check.** `origin/main` is `d6e5c4eec0` (re-fetched; unmoved since round 3), `gh pr view` → `MERGEABLE` / `CLEAN`, `git merge-tree --write-tree` exit **0** (branching on the exit code — that command prints only a tree OID on success and emits no conflict markers, so grepping for them is meaningless). #4666 **does** change `IframeHost.tsx`, the file (d) scans, so this is not a formality: the round-4 guard was run against the **merged** `IframeHost.tsx` + `AppsSubNav.tsx` → **8 passed**, and the merged file still holds exactly the 7 literal-href sites the ledger names. Mutation control on the merged tree: the **P2** shape → **1 failed | 7 passed**, on (d)'s own message.
- Browser tier: **not run this round** — round 4 changes no `.browser.test.tsx` and no rendered output. The NixOS Playwright drift is unchanged (nix ships `chromium_headless_shell-1228`, pinned `playwright-core@1.57.0` wants `1200`). `preview / component-tests` and `Geometry tests` on CI remain the authority for that tier.

Counts read from the runner's own summary lines, never from an exit code.

### CI — round 4

At `474bf91fc6`: **20 checks, all terminal — 19 `SUCCESS`, 1 `SKIPPED`** (`Typecheck (main pushes / fork PRs / non-main base)`, skipped by design), no failures and no re-runs. Polled by requiring a **terminal** conclusion on every entry *and* a minimum entry count of 20 — an empty conclusion is BUSY, not settled, and a rollup that has registered only some of its checks settles instantly and correctly-looking. Green includes **`preview / component-tests`** and **`Geometry tests`** — the two browser-mode tiers on CI's own browser build, which is the authority for that tier — plus `App unit tests + typecheck`, `Unit tests (1..4)`, `tekton / typecheck`, `tekton / fixture-bootstrap`, `preview / auth-tests`, `preview / render-check`, `preview / smoke-tests`, `preview / deploy`, `ESLint + Prettier (changed files)`, `Package unit tests`, `Schema drift gate`, `event-engine-common pin`, both `Windows dev-env` jobs. `mergeable` `MERGEABLE`.

The `app-spend-cap.service.test.ts` wall-clock race from round 1's CI note did **not** recur.

### DIFF
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -564,9 +564,12 @@ export function AppBlockChrome({
           `/apps/invites` and `/apps/revenue`, and none of them belong in a menu
           that opens over a RUNNING app. `/apps/get-started` additionally sits
           behind the `appBlocksGetStarted` kill switch, which this section has no
-          way to read — every item here is unconditional except Review. That
-          excluded SET is itself asserted by the same guard, so adding a subnav row
-          fails there until someone decides which side it lands on.
+          way to read — every item here is unconditional except Review. (The SUBNAV
+          does read it: its row is `visible: (_s, c) => c.canGetStarted`. That is the
+          difference, and it is the whole argument — a surface that cannot honour a
+          kill switch must not advertise the route it switches off.) That excluded
+          SET is itself asserted by the same guard, so adding a subnav row fails
+          there until someone decides which side it lands on.
 
           The LABELS are deliberately NOT all identical: the subnav's tabs sit under
           an "Apps" heading and can afford one-word labels ("Installed", "Review"),
--- a/src/components/AppBlocks/IframeHost.tsx
+++ b/src/components/AppBlocks/IframeHost.tsx
@@ -559,6 +559,15 @@ export function AppBlockChrome({
           which reads BOTH tables and fails when they drift — including when the
           subnav changes and this menu does not.
 
+          🔴 THIS SECTION IS A DELIBERATE SUBSET, NOT A MIRROR. The subnav also
+          carries `/apps/get-started` ("Build apps"), `/apps/submit`,
+          `/apps/invites` and `/apps/revenue`, and none of them belong in a menu
+          that opens over a RUNNING app. `/apps/get-started` additionally sits
+          behind the `appBlocksGetStarted` kill switch, which this section has no
+          way to read — every item here is unconditional except Review. That
+          excluded SET is itself asserted by the same guard, so adding a subnav row
+          fails there until someone decides which side it lands on.
+
           The LABELS are deliberately NOT all identical: the subnav's tabs sit under
           an "Apps" heading and can afford one-word labels ("Installed", "Review"),
           whereas these items stand alone over a running app and need the noun
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -579,9 +579,14 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     // header before trusting the sentence below.
     //
     // That is not bookkeeping. `/apps/get-started` is governed by the `appBlocksGetStarted`
-    // kill switch, and EVERY literal-href item in this chrome is rendered
-    // unconditionally except the moderator-gated `/apps/review` — no flag decides whether
-    // a LINK here is offered. The DELIBERATE SUBSET note above the platform nav in
+    // kill switch, and NO literal-href item in this chrome is gated by a FEATURE FLAG
+    // except the moderator-gated `/apps/review` — no flag decides whether a LINK here is
+    // offered. (Two ARE conditional, on LAYOUT rather than a flag: the compact back
+    // chevron renders only under `compact`, the breadcrumb crumb only under `isPage`.
+    // Neither can be switched off in Flipt, which is what this argument turns on — so
+    // say "gated by no flag", never "rendered unconditionally". Four earlier drafts of
+    // this sentence overstated it in exactly that way.)
+    // The DELIBERATE SUBSET note above the platform nav in
     // `IframeHost.tsx` is the argument that a surface which does not honour a kill switch
     // must not advertise the route it switches off, and the overflow is the same surface,
     // so the same argument governs it; only the enumeration stopped short.
@@ -602,8 +607,8 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     expect(
       links.map((l) => l.href).sort(),
       'the set of routes the app-block chrome links to has changed. Adding one is a ' +
-        'product decision rather than a detail: every literal-href item in this chrome is ' +
-        'rendered UNCONDITIONALLY except the moderator-gated `/apps/review`, so a ' +
+        'product decision rather than a detail: NO literal-href item in this chrome is ' +
+        'gated by a FEATURE FLAG except the moderator-gated `/apps/review`, so a ' +
         'flag-gated destination added here as a plain link keeps being offered after its ' +
         'flag goes down (that is why `/apps/get-started` is excluded; see the DELIBERATE ' +
         'SUBSET note in `IframeHost.tsx`). The surface CAN read flags — ' +
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -425,9 +425,15 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     const links = parseAllChromeLinks(chromeBody());
     const bySubNavHref = new Map(subnav.map((e) => [e.href, e]));
 
-    // Parse control, not a ledger — same reasoning as the two above: a floor pinned on
-    // the live count (5) would take the failure away from the (a)/(b) rules below the
-    // moment the chrome gains an item, and report it as a parse error.
+    // 🔴 PARSE CONTROL ONLY — AND, UNLIKE THE OTHER TWO FLOORS IN THIS FILE, IT HAS NO
+    // SIBLING `toEqual` ELSEWHERE TO INHERIT ITS SET FROM. The floors in the two tests
+    // above are safe to relax because something else owns their sets outright (the exact
+    // 4-item platform nav; the exact excluded set). Nothing owned THIS parser's set, so
+    // relaxing this one alone deleted a live detection — see (d), which was added to
+    // restore it. A zero here is still indistinguishable from a regex wired to nothing,
+    // and ruling that out is all this floor does; pinned on the live count it would
+    // instead steal the failure from the rules below and report a nav change as a parse
+    // error.
     expect(links.length, 'parsed no literal-href items out of the chrome').toBeGreaterThanOrEqual(
       1
     );
@@ -461,6 +467,46 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
         'their icons, not about removing one of them (that would be a behaviour change).'
     ).toEqual(['Installed apps', 'Manage apps']);
     expect(new Set(installed.map((l) => l.icon)).size).toBe(1);
+
+    // (d) 🔴 THE LEDGER — the SET of routes the chrome links to, owned outright, so this
+    // fails when it GROWS or SHRINKS.
+    //
+    // (a) and (b) are PER-LINK and structurally cannot see an ADDITION: a new item
+    // pointing at a route `SUB_NAV_LINKS` already carries, wearing that row's own glyph,
+    // satisfies both. Nor does anything else here close the gap — the `expected glyphs`
+    // test enumerates the PLATFORM-NAV slice only, and (c) enumerates the two
+    // `/apps/installed` LABELS only. So an item added to the ⋮ overflow was invisible to
+    // every assertion in this file. Measured: adding `<ChromeSurfaceItem
+    // href="/apps/get-started" leftSection={<IconCode …/>}>Build apps</ChromeSurfaceItem>`
+    // to the overflow passed all 8 tests.
+    //
+    // That is not bookkeeping. `/apps/get-started` is governed by the `appBlocksGetStarted`
+    // kill switch, and this chrome has no flag plumbing at all — the DELIBERATE SUBSET note
+    // above the platform nav in `IframeHost.tsx` is the argument that a surface which
+    // cannot honour a kill switch must not advertise the route it switches off. The
+    // overflow is the same surface, so the same argument governs it; only the enumeration
+    // stopped short.
+    //
+    // SORTED, so a re-ORDER cannot report a route change that did not happen — this rule
+    // is about the SET, and the platform-nav slice's own `toEqual` is what governs order
+    // there. DUPLICATES KEPT: `/apps/installed` legitimately appears twice, and collapsing
+    // to a Set would hide a third item hung on an already-listed route.
+    expect(
+      links.map((l) => l.href).sort(),
+      'the set of routes the app-block chrome links to has changed. Adding one is a ' +
+        'product decision rather than a detail: this surface has no feature-flag plumbing ' +
+        'at all — the only condition anywhere on it is `isModerator` — so a flag-gated ' +
+        'destination added here keeps being offered after its flag goes down (that is why ' +
+        '`/apps/get-started` is excluded; see the DELIBERATE SUBSET note in ' +
+        '`IframeHost.tsx`). Removing one deletes a door out of a running app. Update this ' +
+        'list deliberately, WITH the reason.'
+    ).toEqual([
+      '/apps', // Marketplace — platform nav
+      '/apps/installed', // Installed apps — platform nav
+      '/apps/installed', // Manage apps — ⋮ overflow; the pair (c) governs their labels
+      '/apps/mine', // My apps — platform nav
+      '/apps/review', // Review — platform nav, moderator-gated
+    ]);
   });
 
   it('the whole-chrome parser sees BOTH sections — positive control', () => {
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -284,10 +284,14 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     const subnav = parseSubNav(code(read(SUBNAV)));
     const nav = parsePlatformNav(code(read(CHROME)));
 
-    // Both tables were actually read. A zero here is indistinguishable from a
-    // parser wired to nothing, so it must never be the thing that makes this pass.
-    expect(subnav.length, 'parsed no rows out of SUB_NAV_LINKS').toBeGreaterThanOrEqual(7);
-    expect(nav.length, 'parsed no items out of the chrome platform nav').toBe(4);
+    // 🔴 PARSE CONTROLS ONLY — deliberately far below the live counts. A zero is
+    // indistinguishable from a parser wired to nothing, so it must never be the thing
+    // that makes this pass; but a control pinned ON the live count STEALS the failure
+    // from the rule below, reporting a parse problem for what is really a nav change.
+    // The exact chrome set is owned by the `expected glyphs` test; the exact excluded
+    // set by the ledger.
+    expect(subnav.length, 'parsed no rows out of SUB_NAV_LINKS').toBeGreaterThanOrEqual(1);
+    expect(nav.length, 'parsed no items out of the chrome platform nav').toBeGreaterThanOrEqual(1);
 
     const bySubNavHref = new Map(subnav.map((e) => [e.href, e]));
 
@@ -328,7 +332,9 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
    *      condition is `isModerator`), while `/apps/get-started` is governed by the
    *      `appBlocksGetStarted` KILL SWITCH. Mirroring the entry here would render it
    *      unconditionally and defeat the switch — the chrome would keep offering a page
-   *      that has been turned off.
+   *      that has been turned off. The subnav's own row IS gated on that flag
+   *      (`visible: (_s, c) => c.canGetStarted`), which is what makes this a difference
+   *      between the two surfaces rather than an inconsistency.
    *
    * So this asserts the excluded SET, and fails when it GROWS (a new subnav row nobody
    * decided about) or SHRINKS (an entry added to the chrome without updating the note).
@@ -339,10 +345,18 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     const nav = parsePlatformNav(code(read(CHROME)));
     const inChrome = new Set(nav.map((e) => e.href));
 
-    // Positive control: the sets were really read. A zero on either side would make
-    // the difference below trivially "everything" or "nothing".
-    expect(subnav.length, 'parsed no rows out of SUB_NAV_LINKS').toBeGreaterThanOrEqual(8);
-    expect(inChrome.size, 'parsed no items out of the chrome platform nav').toBe(4);
+    // 🔴 PARSE CONTROLS, NOT LEDGERS — and the floors are deliberately far below the
+    // live counts (8 and 4). Pinned ON those counts they MASK the ledger they were
+    // meant to protect: adding a chrome item made `toBe(4)` fail first, with the
+    // message "parsed no items out of the chrome platform nav: expected 5 to be 4" —
+    // which states the opposite of what happened — and the `toEqual` below never ran,
+    // leaving the ledger's SHRINK direction unproven. A zero on either side would make
+    // the difference below trivially "everything" or "nothing", and that is all these
+    // two are here to rule out; the SET is the `toEqual`'s to own, in both directions.
+    expect(subnav.length, 'parsed no rows out of SUB_NAV_LINKS').toBeGreaterThanOrEqual(1);
+    expect(inChrome.size, 'parsed no items out of the chrome platform nav').toBeGreaterThanOrEqual(
+      1
+    );
 
     expect(
       subnav.map((e) => e.href).filter((h) => !inChrome.has(h)),
@@ -411,7 +425,12 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     const links = parseAllChromeLinks(chromeBody());
     const bySubNavHref = new Map(subnav.map((e) => [e.href, e]));
 
-    expect(links.length, 'parsed no literal-href items out of the chrome').toBe(5);
+    // Parse control, not a ledger — same reasoning as the two above: a floor pinned on
+    // the live count (5) would take the failure away from the (a)/(b) rules below the
+    // moment the chrome gains an item, and report it as a parse error.
+    expect(links.length, 'parsed no literal-href items out of the chrome').toBeGreaterThanOrEqual(
+      1
+    );
 
     // (a) Every route the chrome links to is a route the store actually has.
     for (const link of links) {
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -471,12 +471,13 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     // (d) 🔴 THE LEDGER — the SET of routes the chrome links to, owned outright, so this
     // fails when it GROWS or SHRINKS.
     //
-    // (a) and (b) are PER-LINK and structurally cannot see an ADDITION: a new item
-    // pointing at a route `SUB_NAV_LINKS` already carries, wearing that row's own glyph,
-    // satisfies both. Nor does anything else here close the gap — the `expected glyphs`
-    // test enumerates the PLATFORM-NAV slice only, and (c) enumerates the two
-    // `/apps/installed` LABELS only. So an item added to the ⋮ overflow was invisible to
-    // every assertion in this file. Measured: adding `<ChromeSurfaceItem
+    // (a) and (b) are PER-LINK, so they catch an addition only when the added link is
+    // itself wrong — an invented route, or the store's route under the wrong glyph. A new
+    // item pointing at a route `SUB_NAV_LINKS` already carries, wearing that row's own
+    // glyph, satisfies both and is invisible to them. Nor does anything else here close
+    // the gap — the `expected glyphs` test enumerates the PLATFORM-NAV slice only, and (c)
+    // enumerates the two `/apps/installed` LABELS only. So an item added to the ⋮ overflow
+    // was invisible to every assertion in this file. Measured: adding `<ChromeSurfaceItem
     // href="/apps/get-started" leftSection={<IconCode …/>}>Build apps</ChromeSurfaceItem>`
     // to the overflow passed all 8 tests.
     //
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -307,6 +307,60 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     }
   });
 
+  /**
+   * 🔴 THE EXCLUSION LEDGER — the half this guard could not previously express.
+   *
+   * The rule above is ONE-DIRECTIONAL: every chrome destination must exist in the
+   * subnav, never the reverse. That is correct — the chrome is deliberately a strict
+   * SUBSET (Create / Invites / Revenue have always been subnav-only) — but it means a
+   * new `SUB_NAV_LINKS` row is scored identically whether its absence from the chrome
+   * was a decision or an oversight. Both readings pass, silently, which is the exact
+   * shape of an unpinned decision.
+   *
+   * `/apps/get-started` ("Build apps") is the row that made this worth writing. It was
+   * EXCLUDED from the chrome on purpose, for two reasons:
+   *
+   *   1. AUDIENCE. This menu opens over a RUNNING app — the viewer is consuming an app,
+   *      not looking for a developer-onboarding page. The four shared destinations are
+   *      the store's consumption + ownership surfaces; a "start building" call to action
+   *      is a store-front concern.
+   *   2. GATING. The chrome's platform nav has no feature-flag plumbing at all (its only
+   *      condition is `isModerator`), while `/apps/get-started` is governed by the
+   *      `appBlocksGetStarted` KILL SWITCH. Mirroring the entry here would render it
+   *      unconditionally and defeat the switch — the chrome would keep offering a page
+   *      that has been turned off.
+   *
+   * So this asserts the excluded SET, and fails when it GROWS (a new subnav row nobody
+   * decided about) or SHRINKS (an entry added to the chrome without updating the note).
+   * It is the deliberate alternative to loosening the guard.
+   */
+  it('🔴 the subnav rows deliberately ABSENT from the chrome are exactly these', () => {
+    const subnav = parseSubNav(code(read(SUBNAV)));
+    const nav = parsePlatformNav(code(read(CHROME)));
+    const inChrome = new Set(nav.map((e) => e.href));
+
+    // Positive control: the sets were really read. A zero on either side would make
+    // the difference below trivially "everything" or "nothing".
+    expect(subnav.length, 'parsed no rows out of SUB_NAV_LINKS').toBeGreaterThanOrEqual(8);
+    expect(inChrome.size, 'parsed no items out of the chrome platform nav').toBe(4);
+
+    expect(
+      subnav.map((e) => e.href).filter((h) => !inChrome.has(h)),
+      'a store subnav destination is missing from the app-block chrome. If that is ' +
+        'deliberate, add it here WITH the reason; if it is not, add it to the chrome ' +
+        '(the subnav is the source of truth, so the chrome follows).'
+    ).toEqual([
+      // Developer onboarding — a store-front concern, and flag-gated by a kill switch
+      // the chrome cannot honour. See the block comment above.
+      '/apps/get-started',
+      // Authoring + owner-management surfaces. Pre-existing exclusions: the chrome is
+      // navigation for someone RUNNING an app, not managing one.
+      '/apps/submit',
+      '/apps/invites',
+      '/apps/revenue',
+    ]);
+  });
+
   it('the four shared destinations are the expected ones, drawn with the expected glyphs', () => {
     // The relationship test above is the real guard; this one names the resolved
     // values so a failure reads as a diff rather than sending you to two files.
--- a/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
+++ b/src/components/AppBlocks/__tests__/chromeNavAlignsWithSubNav.test.ts
@@ -200,29 +200,112 @@ function chromeBody(): string {
   return nextExport === -1 ? rest : rest.slice(0, nextExport);
 }
 
-function parseAllChromeLinks(src: string): NavEntry[] {
-  return [...src.matchAll(/<ChromeSurfaceItem\b([\s\S]*?)<\/ChromeSurfaceItem>/g)]
-    .map((m) => {
-      const block = m[1];
-      const href = /href="([^"]+)"/.exec(block)?.[1];
-      const icon = /leftSection=\{<(Icon\w+)/.exec(block)?.[1];
-      if (!href || !icon) return null;
-      let depth = 0;
-      let tagEnd = -1;
-      for (let i = 0; i < block.length; i += 1) {
-        const ch = block[i];
-        if (ch === '{') depth += 1;
-        else if (ch === '}') depth -= 1;
-        else if (ch === '>' && depth === 0) {
-          tagEnd = i;
-          break;
-        }
+/** A link site in the chrome. `label` and `icon` are OPTIONAL because a real link in
+ *  this file may carry neither — see `parseAllChromeLinks`. */
+type ChromeLink = { tag: string; href: string; label: string | null; icon: string | null };
+
+/** Drop every `{…}` expression from an attribute list, so `href="…"` is matched only
+ *  when it is a LITERAL attribute of this tag. Two things this rules out: a
+ *  `href={`/apps/run/${r.blockId}`}` template (not a platform route), and a nested
+ *  `someProp={<Foo href="…"/>}` being attributed to the OUTER element. */
+function stripAttrExpressions(attrs: string): string {
+  let out = '';
+  let depth = 0;
+  for (const ch of attrs) {
+    if (ch === '{') depth += 1;
+    else if (ch === '}') depth -= 1;
+    else if (depth === 0) out += ch;
+  }
+  return out;
+}
+
+/**
+ * EVERY element in the chrome that carries a LITERAL `href="…"` — whatever tag it is
+ * written as, and whether or not it carries a `leftSection` glyph.
+ *
+ * 🔴 ROUND 4 WIDENED THIS, BECAUSE (d)'s MESSAGE CLAIMED A SET THIS PARSER DID NOT OWN.
+ * The previous version matched `<ChromeSurfaceItem …>text</ChromeSurfaceItem>` and then
+ * dropped anything without BOTH a literal href AND a `leftSection={<IconX`. Two ordinary
+ * shapes walked straight through it, both measured surviving as mutants:
+ *
+ *   • `<ActionIcon component={Link} href="/apps/get-started" …>` in the ⋮ overflow — and
+ *     that is not an exotic shape, it is the one the chrome ALREADY uses for its `/apps`
+ *     back-link (the compact-mode chevron);
+ *   • `<ChromeSurfaceItem href="/apps/get-started">Build apps</ChromeSurfaceItem>` with no
+ *     `leftSection` — which compiles, because `ChromeSurface.tsx` types that prop
+ *     `leftSection?: ReactNode`, and which is the very element (d) claims to enumerate.
+ *
+ * Either one is a door out of a running app into a flag-gated route, offered
+ * unconditionally, with all 8 tests green. So the parser now scans TAGS rather than one
+ * tag name, and treats the glyph as optional metadata rather than a condition of
+ * inclusion. What that costs is stated where it is paid: rule (b) below can only compare
+ * a glyph an element actually HAS, so it now skips the icon-less links that (d) still
+ * counts.
+ *
+ * 🔴 `icon` IS THE `leftSection` GLYPH ONLY — deliberately not "any Icon inside the
+ * element". The back chevron renders `<IconChevronLeft/>` as its CHILD; that is a
+ * directional affordance, not this route's glyph, and scoring it as one would make the
+ * same-route rule red on correct code. The rule (b) enforces is about the picture in a
+ * menu ROW, which is what `leftSection` is.
+ *
+ * Mechanics: an opening tag ends at the first `>` at brace depth 0 (`leftSection={<Icon
+ * … />}` contains a `>` that is not the end of the tag), and scanning resumes AFTER that
+ * `>` — so a nested `<Icon…>` inside the attribute region is never itself opened as an
+ * element. The label is the element's text with nested tags removed, falling back to
+ * `aria-label` (the back chevron has no text at all, and a null label would make (a)/(b)'s
+ * failure messages name nothing). `</tag>` is matched by name, which is exact here because
+ * none of these tags nest inside themselves; if one ever does, the LABEL is what goes
+ * wrong, never the href set (d) owns.
+ */
+function parseAllChromeLinks(src: string): ChromeLink[] {
+  const out: ChromeLink[] = [];
+  let i = 0;
+  while (i < src.length) {
+    if (src[i] !== '<' || !/[A-Za-z]/.test(src[i + 1] ?? '')) {
+      i += 1;
+      continue;
+    }
+    const tag = /^[A-Za-z][\w.]*/.exec(src.slice(i + 1))?.[0];
+    if (!tag) {
+      i += 1;
+      continue;
+    }
+    let depth = 0;
+    let tagEnd = -1;
+    for (let j = i + 1 + tag.length; j < src.length; j += 1) {
+      const ch = src[j];
+      if (ch === '{') depth += 1;
+      else if (ch === '}') depth -= 1;
+      else if (ch === '>' && depth === 0) {
+        tagEnd = j;
+        break;
       }
-      if (tagEnd === -1) return null;
-      const label = block.slice(tagEnd + 1).trim();
-      return label ? { href, label, icon } : null;
-    })
-    .filter((e): e is NavEntry => e !== null);
+    }
+    if (tagEnd === -1) break;
+    const attrs = src.slice(i + 1 + tag.length, tagEnd);
+    const selfClosing = src[tagEnd - 1] === '/';
+    i = tagEnd + 1;
+
+    const href = /\bhref="([^"]+)"/.exec(stripAttrExpressions(attrs))?.[1];
+    if (!href) continue;
+    const icon = /\bleftSection=\{<(Icon\w+)/.exec(attrs)?.[1] ?? null;
+
+    let label: string | null = null;
+    if (!selfClosing) {
+      const close = src.indexOf(`</${tag}>`, tagEnd);
+      if (close !== -1) {
+        label =
+          src
+            .slice(tagEnd + 1, close)
+            .replace(/<[^>]*>/g, '')
+            .trim() || null;
+      }
+    }
+    if (!label) label = /\baria-label="([^"]+)"/.exec(attrs)?.[1] ?? null;
+
+    out.push({ tag, href, label, icon });
+  }
+  return out;
 }
 
 describe('the app-block chrome platform nav agrees with the store subnav', () => {
@@ -449,7 +532,13 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     }
 
     // (b) …and draws it with the store's glyph, wherever in the chrome it appears.
-    for (const link of links) {
+    //
+    // 🔴 SCOPED TO THE LINKS THAT HAVE A `leftSection` GLYPH, which since round 4 is a
+    // SUBSET of `links` rather than all of it. An element with no glyph cannot be drawing
+    // the route with the WRONG one, so there is nothing here to compare; its presence in
+    // the chrome is still owned outright by (d). The two live examples are the compact
+    // back chevron and the breadcrumb crumb, neither of which is a menu row.
+    for (const link of links.filter((l) => l.icon !== null)) {
       expect(
         link.icon,
         `same-route icon drift: "${link.label}" links to \`${link.href}\` with \`${link.icon}\`, ` +
@@ -477,32 +566,55 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
     // glyph, satisfies both and is invisible to them. Nor does anything else here close
     // the gap — the `expected glyphs` test enumerates the PLATFORM-NAV slice only, and (c)
     // enumerates the two `/apps/installed` LABELS only. So an item added to the ⋮ overflow
-    // was invisible to every assertion in this file. Measured: adding `<ChromeSurfaceItem
-    // href="/apps/get-started" leftSection={<IconCode …/>}>Build apps</ChromeSurfaceItem>`
-    // to the overflow passed all 8 tests.
+    // was invisible to every assertion in this file. Measured on the PRE-(d) tree: adding
+    // `<ChromeSurfaceItem href="/apps/get-started" leftSection={<IconCode …/>}>Build
+    // apps</ChromeSurfaceItem>` to the overflow passed all 8 tests. (d) kills that one now.
+    //
+    // 🔴 AND (d) ALONE WAS STILL NOT ENOUGH, WHICH IS WHY THIS COMMENT IS NOT THE END OF
+    // THE STORY. Round 4 measured two FURTHER shapes surviving with (d) in place, because
+    // the PARSER feeding it required both a literal `href` and a `leftSection` glyph — an
+    // `<ActionIcon component={Link} href>` and a `leftSection`-less `<ChromeSurfaceItem>`.
+    // Both are ordinary; the first is the shape the chrome already uses for its `/apps`
+    // back-link. The ledger is only ever as wide as `parseAllChromeLinks` — read its
+    // header before trusting the sentence below.
     //
     // That is not bookkeeping. `/apps/get-started` is governed by the `appBlocksGetStarted`
-    // kill switch, and this chrome has no flag plumbing at all — the DELIBERATE SUBSET note
-    // above the platform nav in `IframeHost.tsx` is the argument that a surface which
-    // cannot honour a kill switch must not advertise the route it switches off. The
-    // overflow is the same surface, so the same argument governs it; only the enumeration
-    // stopped short.
+    // kill switch, and EVERY literal-href item in this chrome is rendered
+    // unconditionally except the moderator-gated `/apps/review` — no flag decides whether
+    // a LINK here is offered. The DELIBERATE SUBSET note above the platform nav in
+    // `IframeHost.tsx` is the argument that a surface which does not honour a kill switch
+    // must not advertise the route it switches off, and the overflow is the same surface,
+    // so the same argument governs it; only the enumeration stopped short.
+    //
+    // 🔴 THAT IS NOT "THIS SURFACE CANNOT READ FLAGS" — it can, and the repo demonstrates
+    // it a few lines away. `<ChromeReviewMenuItem>` (`IframeHost.tsx`) calls
+    // `useOptionalFeatureFlags()` and returns null without `hasAppsStoreAccess(features)`,
+    // and its `useCanReviewListing` narrows again through `resolveClientStoreScope`. It
+    // carries no `href`, which is why it is not in this set — and it is exactly the shape a
+    // gated door WOULD take. So a maintainer this ledger stops has two honest options, not
+    // one: exclude the route, or add it gated the way that item is. What is not an option
+    // is adding it as a plain link.
     //
     // SORTED, so a re-ORDER cannot report a route change that did not happen — this rule
     // is about the SET, and the platform-nav slice's own `toEqual` is what governs order
-    // there. DUPLICATES KEPT: `/apps/installed` legitimately appears twice, and collapsing
-    // to a Set would hide a third item hung on an already-listed route.
+    // there. DUPLICATES KEPT: `/apps` and `/apps/installed` each legitimately appear more
+    // than once, and collapsing to a Set would hide an extra item hung on a listed route.
     expect(
       links.map((l) => l.href).sort(),
       'the set of routes the app-block chrome links to has changed. Adding one is a ' +
-        'product decision rather than a detail: this surface has no feature-flag plumbing ' +
-        'at all — the only condition anywhere on it is `isModerator` — so a flag-gated ' +
-        'destination added here keeps being offered after its flag goes down (that is why ' +
-        '`/apps/get-started` is excluded; see the DELIBERATE SUBSET note in ' +
-        '`IframeHost.tsx`). Removing one deletes a door out of a running app. Update this ' +
-        'list deliberately, WITH the reason.'
+        'product decision rather than a detail: every literal-href item in this chrome is ' +
+        'rendered UNCONDITIONALLY except the moderator-gated `/apps/review`, so a ' +
+        'flag-gated destination added here as a plain link keeps being offered after its ' +
+        'flag goes down (that is why `/apps/get-started` is excluded; see the DELIBERATE ' +
+        'SUBSET note in `IframeHost.tsx`). The surface CAN read flags — ' +
+        '`ChromeReviewMenuItem` gates itself on `hasAppsStoreAccess(useOptionalFeatureFlags())` ' +
+        '— so "add it GATED, the way that item is" is a real third option alongside ' +
+        'excluding it; adding it ungated is not. Removing one deletes a door out of a ' +
+        'running app. Update this list deliberately, WITH the reason.'
     ).toEqual([
       '/apps', // Marketplace — platform nav
+      '/apps', // the compact back chevron — `<ActionIcon component={Link}>`, no leftSection
+      '/apps', // the breadcrumb's first crumb — `<Anchor component={Link}>`, no leftSection
       '/apps/installed', // Installed apps — platform nav
       '/apps/installed', // Manage apps — ⋮ overflow; the pair (c) governs their labels
       '/apps/mine', // My apps — platform nav
@@ -528,6 +640,41 @@ describe('the app-block chrome platform nav agrees with the store subnav', () =>
         '<ChromeSurfaceItem href={`/apps/run/${r.blockId}`} leftSection={<IconApps />}>x</ChromeSurfaceItem>'
       )
     ).toHaveLength(0);
+
+    // 🔴 THE TWO SHAPES ROUND 4 ADDED, AS FIXTURES. Both were measured SURVIVING as
+    // mutants against the pre-round-4 parser (8 passed, `/apps/get-started` invisible to
+    // every assertion in this file). A widened parser that silently stopped matching them
+    // again would restore that hole while looking exactly like this — so the shapes are
+    // pinned here, not merely described in the comment above.
+    //
+    // P1: the `<ActionIcon component={Link} href>` shape, which the chrome ALREADY uses
+    // for its `/apps` back-link. No text child and no `leftSection`, so the label comes
+    // from `aria-label` and the icon is null.
+    expect(
+      parseAllChromeLinks(
+        '<ActionIcon component={Link} href="/apps/get-started" aria-label="Build apps">' +
+          '<IconCode size={16} stroke={1.5} /></ActionIcon>'
+      )
+    ).toEqual([{ tag: 'ActionIcon', href: '/apps/get-started', label: 'Build apps', icon: null }]);
+
+    // P2: a `ChromeSurfaceItem` with NO `leftSection` — legal, because the primitive types
+    // it `leftSection?: ReactNode`. The old parser required the glyph for INCLUSION.
+    expect(
+      parseAllChromeLinks(
+        '<ChromeSurfaceItem href="/apps/get-started">Build apps</ChromeSurfaceItem>'
+      )
+    ).toEqual([
+      { tag: 'ChromeSurfaceItem', href: '/apps/get-started', label: 'Build apps', icon: null },
+    ]);
+
+    // …and the negative control for `stripAttrExpressions`: an href nested inside ANOTHER
+    // element in the attribute region is that element's, not this one's. Without the
+    // strip, the outer tag would be reported as linking to `/apps/get-started`.
+    expect(
+      parseAllChromeLinks(
+        '<ChromeSurfaceItem leftSection={<Foo href="/apps/get-started" />}>x</ChromeSurfaceItem>'
+      )
+    ).toEqual([]);
   });
 
   it('the breadcrumb’s first crumb reads "Marketplace" and still links to /apps', () => {
--- a/src/components/AppLayout/AppHeader/appsMenuEntry.test.ts
+++ b/src/components/AppLayout/AppHeader/appsMenuEntry.test.ts
@@ -1,0 +1,274 @@
+import fs from 'fs';
+import path from 'path';
+import ts from 'typescript';
+import { describe, expect, it } from 'vitest';
+
+/**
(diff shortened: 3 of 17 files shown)


ZacxDev:
Round 1 adversarial audit complete; fixes pushed. Claims block for the next round:

```audit-claims round=1 audited=aec15e652b..019bc59c
1. 🔴-1 CLAIMED FIXED: the "Build apps" tab is no longer unconditional — `AppsNavContext` gains `canGetStarted` (AppsSubNav.tsx:80-88), derived in the container at :485-490 as `!!features.appBlocksGetStarted`, and the tab reads `visible: (_s, c) => c.canGetStarted` at :170-175.
2. 🔴-1 CLAIMED: `canGetStarted` is deliberately NOT session-scoped (unlike `isAuthor`), on the stated ground that `resolveGetStartedAccess` consults no user, so folding it into the `currentUser` branch would hide the tab from a logged-out viewer the page would serve.
3. 🔴-1 CLAIMED: a store-access-only viewer now falls back to exactly `main`'s behaviour (Marketplace alone → the `< 2` collapse hides the bar), and the `app-blocks-get-started` Flipt kill switch again removes the nav entry.
4. 🔴-1 CLAIMED: the `storeGate` browser suite was reworked so the case that previously asserted `['Build apps','Marketplace']` for the store-only cohort now asserts `['Marketplace','Create']` with no Build apps tab; logged-out and appListings-only-non-author cases assert no bar; a new pair pins the get-started-only cohort.
5. 🔴-1 CLAIMED: test blocks the previous round had relabelled as "pinning the view, not a cohort" are relabelled back, on the ground that the `< 2` collapse is reachable again (mutation M5).
6. 🔴-1 CLAIMED: the docstring at AppsSubNav.tsx:132-166 and the collapse notes at :124-129 and :383-388 were rewritten from the post-fix code rather than from the prior intent.
7. 🟡-1 CLAIMED FIXED: all three count controls in chromeNavAlignsWithSubNav.test.ts (:289/290, :344/345, :414) are now `toBeGreaterThanOrEqual(1)`, so the `toEqual` ledgers own their sets and fire on both GROW and SHRINK with their own assertion's message.
8. 🟡-2 CLAIMED FIXED + SWEPT: both `edit.tsx` notes, the `get-started.tsx` referent, IframeHost.tsx §4 and the ledger block comment were repaired against the code; a `command grep -rn` sweep with positive controls found no further sites. One site deliberately left: server/routers/blocks.router.ts:2607, on the ground it was already loose pre-PR and is not invalidated by this diff.
9. 🟢-2 CLAIMED FIXED: `tabWidth` removed from AppsPageLayout.geometry.browser.test.tsx; all lookups behind a `required()` helper producing the intended diagnosis instead of a bare TypeError; the :47 header note rewritten from post-fix code on the ground it is true again.
10. 🟢-3 CLAIMED FIXED: tests/preview-apps-marketplace.spec.ts:51-52 now names the row it keys on and states order is irrelevant to the by-name assertion.
11. 🟢-4 CLAIMED DECIDED (accept, no code change): the surviving dropdown row stays labelled "Apps" with IconPlugConnected for both destinations; rationale and a re-decide trigger (`app-blocks-get-started` widened past `['mod']`) recorded at hooks.tsx:261-276. Asserted ground: the mismatched cohort is empty today because availability is `['mod']` and every mod holds a store flag.
12. VERIFICATION CLAIMED: merged tree vs origin/main `f311298e13` → 0 type errors; component tier 123 files / 1720 tests; full unit tier 38295 passed with 6 files / 1 test red measured identical at un-merged origin/main. CI at 019bc59cd4 → 20 checks terminal, 19 SUCCESS + 1 SKIPPED, mergeStateStatus CLEAN.
13. VERIFICATION CAVEAT CLAIMED: the local browser tier ran only via a PLAYWRIGHT_BROWSERS_PATH shim aliasing chromium_headless_shell-1200 to the nix -1228 build, so CI's two green browser tiers are the stated authority; the 31 `packages/**` "Cannot find module 'kysely'" typecheck errors are claimed to be a stale generated-client artifact, clean after regeneration.
```

ZacxDev:
## Round 3 pushed (`684be6cc8f`) — one round-2 claim withdrawn, and one challenge to the numbers that did *not* hold

Two things a reviewer who already read this PR should know, because both concern claims that were in the body before now.

### 1. 🟡 Withdrawn: round 2 said all three relaxed floors had their sets owned. Only two did.

Round 2 relaxed three count controls in `chromeNavAlignsWithSubNav.test.ts` to `toBeGreaterThanOrEqual(1)` and reported that *"the `toEqual` ledgers own their sets in both directions"*. That is true for two of them — the exact 4-item platform nav, and the exact excluded set. It was **false for `parseAllChromeLinks`**, whose set nothing owned, and whose comment claimed *"same reasoning as the two above"*. The relaxation deleted a live detection.

Rules (a) and (b) in that test are **per-link**, so a new chrome item pointing at a route `SUB_NAV_LINKS` already carries, wearing that row's own glyph, satisfies both. Measured at `019bc59cd4`: adding a "Build apps" item for `/apps/get-started` to the **⋮ overflow** — outside the platform-nav slice the other ledgers enumerate — **passed all 8 tests**.

That matters because this chrome has no feature-flag plumbing at all (its only condition anywhere is `isModerator`), which is precisely the DELIBERATE SUBSET argument for keeping `/apps/get-started` off it: someone adds "Build apps" to the ⋮ menu, the guard stays green, `app-blocks-get-started` is later switched off, and the chrome keeps advertising a page that answers `notFound`.

Fixed by adding a `toEqual` ledger **(d)** over the whole literal-href set, after (a)/(b)/(c) so the specific rules keep their messages. `toBe(5)` deliberately **not** restored — that was the misdirection round 2 correctly removed. The mutant that previously survived now dies on (d)'s own assertion, with (a) and (b) still passing:

```
expected [ '/apps', '/apps/get-started', …(4) ] to deeply equal [ '/apps', '/apps/installed', …(3) ]
```

Both directions measured, and both pre-existing ledgers re-confirmed against GROW and SHRINK — full matrix in the **Review round 3** section of the body. One limitation recorded rather than engineered around: every *existing* link is already owned by some assertion for SHRINK, so (d)'s unique contribution is GROW outside the platform-nav slice.

### 2. 🟢 Correction to a correction: the "123 files" figure was challenged, and it holds

A review note flagged the body's **"123 files / 1720 tests"** as understating the `component` tier by 94 files, on the ground that the `component` project's include matches **217 files / 2403 tests**. **I re-measured both, and the challenge does not hold** — they are different populations and both numbers are right. The body's figure comes from a *path-filtered* run whose filter is printed next to it:

```
vitest --project component src/components/Apps src/components/AppBlocks
```

- **123** — exactly what `*.browser.test.tsx` under those two directories counts in the merged tree the body measured against, and still 123 against today's `origin/main`.
- **121 files / 1687 tests, all passed** — the same command in the branch worktree, run in full. Exact reproduction of the body's other figure.
- **217 files / 2403 tests, all passed** — `--project component` *unfiltered*, run in full. Correct, and a different question.

So no number needed correcting there. Everything else in the verification section also reproduces: 122 files / 2353 tests on the unit filter, the 12/13/14 changed-file counts, and the 6-problem eslint positive control.

**One thing genuinely was wrong** and is struck through in the body: the line-number citation `:289, :290, :344, :345, :414`. At `019bc59cd4` those floors sit at `:293`, `:294`, `:356`, `:357`, `:431`.

**Not re-measured, and flagged rather than restated:** the *1720 test count* for the merged tree. The file-count half reproduces exactly; re-running the test half needs a fully installed merged worktree, which I did not build.

### Round-3 verification

`--project component` unfiltered → **217 files / 2403 tests, all passed**. `node scripts/typecheck.mjs` → **0 errors**, with the identical command minutes earlier reporting **2848** off a Prisma client generated 2025-12-08 as its positive control (none of the 2848 named either changed file). eslint 0 with `zod-helpers.ts` reporting 6 as the control; prettier clean with a misformatted scratch file as the negative control.

**Merged-tree check was not a formality this time:** `origin/main` moved to `d6e5c4eec0`, and #4666 touches `IframeHost.tsx` — the very file the new ledger scans. The guard including (d) was run against the *merged* sources: **8 passed**.

Also in this push: `hooks.tsx`'s re-decide trigger was narrower than the branch's reachability (the branch is `!marketplace && getStarted`, so the store flags *narrowing* reaches it too, putting every moderator on an "Apps" row that navigates to developer onboarding), and its stated ground was a live-Flipt claim presented as derived from `availability` — which is only the Flipt-down fallback. Both corrected; comment-only, no behaviour change.

ZacxDev:
Round 2 delta re-audit complete; fixes pushed. Claims block for the next round:

```audit-claims round=2 audited=019bc59cd4..136fffd647
1. 🟡-A CLAIMED FIXED: chromeNavAlignsWithSubNav.test.ts:471-509 adds ledger (d), a `toEqual` over `parseAllChromeLinks`' hrefs, placed AFTER (a)/(b)/(c) so the specific rules keep their messages. Sorted; duplicates deliberately kept (`/apps/installed` legitimately appears twice, and a Set would hide a third item on an already-listed route). `toBe(5)` was NOT restored.
2. 🟡-A CLAIMED: the false comment at :428-436 now states this floor has no sibling `toEqual` to inherit its set from, and points at (d); the "same reasoning as the two above" sentence is gone.
3. 🟡-A CLAIMED (stated limitation, not engineered around): (d)'s SHRINK direction is reachable but NOT uniquely load-bearing — every existing link is already owned by some assertion for SHRINK, so no shrink was constructible that only (d) catches. (d)'s unique contribution is claimed to be GROW outside the platform-nav slice.
4. 🟡-A CLAIMED mutation matrix: GROW in the ⋮ overflow (`/apps/get-started` "Build apps") was 8 passed / SURVIVED pre-fix, and post-fix is 1 failed / 7 passed dying on (d)'s own message, with (a) and (b) passing so (d) is the only catcher. SHRINK deleting platform-nav Review → 4 failed / 4 passed, (d) fires with its own message. SHRINK deleting ⋮ "Manage apps" → dies on (c), which precedes (d).
5. 🟡-A CLAIMED no regression: the exclusion ledger still fires on its own `toEqual` for GROW and SHRINK, and `expected glyphs` still fires for both directions.
6. 🟢-B CLAIMED FIXED (comment-only): hooks.tsx:252-276 now names both routes into the `!marketplace && getStarted` branch — get-started widening, or `app-blocks-enabled`/`app-listings` narrowing — and restates the ground as: `availability` is only the Flipt-DOWN fallback, so the cohort's emptiness is observable ONLY in live Flipt. Claimed verified that all four App-Blocks flags carry a `fliptKey` (feature-flags.service.ts:510,520,561,571).
7. 🟢-C CLAIMED NOT A DEFECT: the body's "123 files" came from a path-filtered run (`--project component src/components/Apps src/components/AppBlocks`) printed beside the figure; the round-2 audit compared it against the unfiltered project include. Claimed both numbers are correct for different populations, so the number was deliberately not "corrected".
8. 🟢-C CLAIMED re-measured: 123 files at the merged tree and still 123 at today's origin/main; 121 files / 1687 tests exact; unfiltered tier 217 files / 2403 tests all passed; unit 122 files / 2353 tests exact; the changed-file counts and the 6-error eslint control all reproduce.
9. 🟢-C CLAIMED CORRECTED: the line citations `:289,:290,:344,:345,:414` are actually `:293,:294,:356,:357,:431` — struck through in the PR body. Also struck: the round-2 sentence "the `toEqual` ledgers own their sets in both directions", half-withdrawn because that was the defect. Comment posted at #issuecomment-5565369320.
10. CLAIMED about the red/green standard: (d) would PASS at 019bc59cd4 because the tree is unchanged; what goes red is the mutant that pre-fix accepted, claimed measured both ways.
11. VERIFICATION CLAIMED: merged-tree guard run against the MERGED `IframeHost.tsx` + `AppsSubNav.tsx` (origin/main moved to d6e5c4eec0 and #4666 touches IframeHost.tsx, the file (d) scans) → 8 passed. ~~CI at 136fffd647 → 19 checks terminal, 18 SUCCESS + 1 SKIPPED~~, MERGEABLE/CLEAN. **CORRECTED in round 4 (re-read from the rollup): 20 entries, 19 `SUCCESS` + 1 `SKIPPED` (`Typecheck (main pushes / fork PRs / non-main base)`, skipped by design).** Wrong in the safe direction, but it is a count in prose.
12. VERIFICATION CLAIMED: local typecheck first reported 2848 errors from a Prisma client generated 2025-12-08; after `prisma generate` under the flake's PRISMA_* engine paths → 0 type errors in 81s, with that failing run offered as the positive control for the zero. None of the 2848 named either changed file.
13. NOT RE-MEASURED, claimed flagged rather than restated: the 1720 test count for the merged tree — the file-count half reproduces exactly, the test half needs a fully installed merged worktree that was not built.
```

ZacxDev:
Round 3 delta re-audit complete; fixes pushed. Claims block for the next round:

```audit-claims round=3 audited=136fffd647..474bf91fc6
1. F1 CLAIMED FIXED by WIDENING the parser (not narrowing the message): chromeNavAlignsWithSubNav.test.ts:203-303 now scans tags rather than one tag name, treats `leftSection` as optional metadata, ends an opening tag at the first `>` at brace depth 0 and resumes after it, and matches `href` against the attribute list with every `{…}` stripped (`stripAttrExpressions`, :208).
2. F1 CLAIMED: the owned set grows 5 → 7 (:598-606) — the compact back chevron (`ActionIcon component={Link}`) and the breadcrumb crumb (`Anchor component={Link}`), both `/apps`, both glyph-less.
3. F1 CLAIMED COST, stated where it is paid: rule (b) is now scoped to links that HAVE a `leftSection` glyph (:535-541), on the ground that a glyph-less element cannot draw the route with the WRONG glyph and its presence is still owned by (d). `icon` is claimed to be deliberately the `leftSection` glyph only, never "any Icon inside the element", because the back chevron's child `IconChevronLeft` is a directional affordance that would make (b) red on correct code.
4. F1 CLAIMED: both new shapes are pinned as parser fixtures at :635-671, plus a negative control for the expression-strip.
5. F1 CLAIMED mutation results: P1 (`ActionIcon component={Link} href="/apps/get-started"` in the ⋮ overflow) and P2 (`ChromeSurfaceItem` with no `leftSection`) were each 8 passed / SURVIVED pre-fix and are each 1 failed / 7 passed post-fix, dying on (d)'s own message with (a)/(b)/(c) passing. Pre and post were run in ONE vitest invocation against a byte copy of 136fffd647, baseline 16 passed.
6. F1 CLAIMED no regression, all re-measured on the round-4 tree: ⋮-overflow GROW with a glyph → (d)'s own message; SHRINK platform-nav Review → (d)'s own message plus three others; SHRINK ⋮ "Manage apps" → dies on (c); exclusion ledger GROW and SHRINK → its own `toEqual`; `expected glyphs` fires both ways.
7. F2 CLAIMED FIXED, derived from code: :573-591 and the message at :598 now say every literal-href item renders unconditionally except the moderator-gated `/apps/review`, and that the surface CAN read flags — so the ledger offers a maintainer the "add it GATED, the way ChromeReviewMenuItem is" option instead of foreclosing it. Claimed verified that `chromeBody()` spans IframeHost.tsx:329-1089, including ChromeDesktopLeadingGroup (:900) and ChromeReviewMenuItem (:688), which gates on `hasAppsStoreAccess(useOptionalFeatureFlags())` (ChromeReviewEntry.tsx:101-102) and `useCanReviewListing` → `resolveClientStoreScope` (ReviewListingButton.tsx:89-94).
8. F3 CLAIMED FIXED: hooks.tsx:271-282 names all three store flags (`app-listings` ‖ `app-blocks-enabled` ‖ `app-listings-public-external`, keys claimed verified at feature-flags.service.ts:510,520,544) and states any ONE keeps `marketplace === true`. :252-265 scopes the `availability` sentence to the ROLE terms only, noting env/region/server-colour terms run before Flipt and cannot be overridden by it (feature-flags.service.ts:790-805), with an explicit "do not generalise".
9. F4 CLAIMED FIXED plus one site the brief did not name: PR body :243 and :245 struck as briefed; **:129 struck** — round 1's "a moderator holds the store flags", the same live-Flipt claim round 3 withdrew in the hooks.tsx comment but never reached in the body. :285 and :287 corrected (F2's body twin, and "over the whole literal-href set" → "the set its parser could see — a strict subset"). :58 deliberately left, claimed true because scoped to the platform nav. :119's round-1 "18/18 checks" claimed not re-derivable now and flagged rather than rewritten.
10. F4 CLAIMED sweep method: a pattern battery with controls, not by eye — positive control `Marketplace` → 22 hits, negative control `zzz-no-such-string` → 0, eight retraction signatures, every hit read and dispositioned; post-edit re-sweep found 0 unmarked retracted sites.
11. CLAIMED: the claim-11 count correction (20 rollup entries, 19 SUCCESS + 1 SKIPPED, not "19 terminal / 18 SUCCESS") was in the round-3 REVIEW COMMENT, not the body; patched in place at issue comment 5565469635 and recorded in the round-4 body section.
12. 🔴 INSTRUMENT CLAIM: `node scripts/typecheck.mjs` is structurally BLIND to the changed guard — the root tsconfig.json excludes `src/**/__tests__/**`, and a deliberately planted `const NEGATIVE_CONTROL: number = src;` was ALSO reported as 0 errors / exit 0. The correct instrument is claimed to be `node scripts/typecheck.mjs -p tsconfig.tests.json` → 1160 errors, 0 in either changed file; the same command with the planted error → 1161, the extra being chromeNavAlignsWithSubNav.test.ts(261,9) TS2322.
13. VERIFICATION CLAIMED: guard 8 passed; `src/components/{AppBlocks,Apps,AppLayout}` → 122 files / 2353 tests all passed (claimed to reproduce rounds 2 and 3 exactly); eslint 0 problems with file count read back as 2 and a 6-problem positive control; prettier watched RED on this change then clean. Merged tree vs origin/main d6e5c4eec0 (re-fetched, unmoved) → merge-tree exit 0 branched on exit code not a marker grep, guard 8 passed against the MERGED IframeHost.tsx which still holds exactly the 7 ledger sites, merged-tree P2 mutation → 1 failed / 7 passed on (d)'s message. CI at 474bf91fc6 → 20 checks all terminal, 19 SUCCESS + 1 SKIPPED, polled by requiring a terminal conclusion on every entry AND a minimum of 20, with `.github/workflows` confirmed present on origin/main first.
14. REPORTED, NOT FIXED (claimed out of scope): `scripts/ci/typecheck-tests-gate.mjs` is RED on this tree with 1160 pre-existing test-tree type errors across 11 baselined files in unrelated subsystems; claimed arithmetically not attributable to this PR (changed files contribute 0) and CI green on the branch, but whether it is ALSO red on origin/main was NOT established.
15. REPORTED, DELIBERATELY NOT TOUCHED: IframeHost.tsx:566-567 still says the platform-nav section has "no way to read [the kill switch] — every item here is unconditional except Review", which ignores `{recentApps.length > 0 && …}` in the same const; same imprecision at the exclusion ledger's comment :334-337. Claimed left because both are FLAG-accurate, which is what the argument rests on.
16. NOT RUN: the browser tier — claimed round 4 changes no `.browser.test.tsx` and no rendered output; NixOS Playwright drift unchanged; CI's two browser tiers green.
```

ZacxDev:
## Audit ladder closed after round 4 — `a0402d1be2`

Round 4 returned one LOW finding (F-1): the route ledger's comment and assertion message said every literal-href chrome item is "rendered UNCONDITIONALLY" except `/apps/review`, while two of the seven are conditional on layout (`compact`, `isPage`). Fixed in `a0402d1be2` at both source sites and both PR-body sites, using the narrow claim the code supports — "gated by no FEATURE FLAG" — with the two layout-conditional sites named in the comment.

**Why the ladder stops here rather than running a round 5**, stated explicitly so this reads as a decision and not as convergence:

1. **F-1 was a false sentence with no behavioural, detection or decision consequence.** The inference it supports is unaffected — neither `compact` nor `isPage` is a feature flag, and the only `useFeatureFlags()` call in the file sits outside `chromeBody()`. Under this ladder's own rule, a reword that changes nothing a reader does is not a finding.

2. **No 🔴 and no blast radius beyond "the document contains a false sentence".** The 🔴 from round 1 is closed and mutation-proven: reverting `visible` to `() => true` turns 13 tests red across 3 files, and the store-access-only cohort's tab set is byte-identical to `main`'s by construction.

3. **The payload is now prose, so the attribution gate cannot fire.** Rounds 2, 3 and 4 changed 201, 28 and 34 payload lines with **0 executable** in each of the last three. Comments in shipped source count as payload, so the two-zero-rounds gate is structurally inert here while the ladder does exactly what that gate exists to catch.

4. **The recurring SHAPE was swept at every site, not just the reported one.** This was the fourth draft of that sentence to overstate in the same direction — each rewritten while fixing the previous round. The comment now records the dead drafts and the instruction not to reach for "unconditionally" again, which is what stops a fifth being derived. Body sweep re-run with controls; the three surviving uses of the word are accurate as scoped (`:58` platform-nav, `:364` describing the hypothetical added link).

### Open, deliberately not fixed

- **`IframeHost.tsx:566-567` and `:334-337`** say the platform-nav section's "only condition is `isModerator`", which ignores `{recentApps.length > 0 && …}` in the same const. Both are *flag*-accurate, which is what the argument rests on. Left as-is; noted here so it is visible rather than absent.
- **PR body `:119`'s "18/18 checks"** is a round-1 count against a round-1 commit and is not re-derivable now. Flagged rather than silently rewritten. The current rollup at `a0402d1be2` is the authority.
- **`scripts/ci/typecheck-tests-gate.mjs` is red on this tree** (11 baselined files worse). Not this PR's: `origin/main` fails the same gate harder (12 worse, 3 entries the PR tree does not have), and 0 errors name any changed file.

### One thing worth carrying past this PR

`node scripts/typecheck.mjs` is **structurally blind to `src/**/__tests__/**`** — the root `tsconfig.json` excludes it. Confirmed by file-loading counts (default project 12,515 files, this guard at 0; tests project 14,607, at 1) and by a planted `const NEGATIVE_CONTROL: number = …`, which the default run reported as 0 errors. Any earlier round's typecheck zero for a test file was vacuous. The repo already documents this at `scripts/ci/typecheck-tests-gate.mjs:9-10`, so it is a re-discovery, not a new hole — but it is worth knowing before quoting a typecheck zero for a test file again. Use `-p tsconfig.tests.json`.

Verification for `a0402d1be2`: guard 8/8; eslint 0 on the changed file with a 6-problem positive control in the same invocation; prettier clean with a deliberately misformatted control watched to fail first.

PR #4680, opened 2026-09-07
feat(moderation): dark shadow probe to measure whether a cheaper classifier model agrees

ZacxDev:
feat(moderation): dark shadow probe to measure whether a cheaper classifier model agrees

The prompt classifier costs ~205ms of wall time on every generation submit and is
78-88% of the app-side time of orchestrator.generateFromGraph. Roughly 175-180ms of
that is the model's own inference, which makes it the largest single component and
the only one nobody has tried to move — every infra-side lever (tail clamping,
connection reuse, overlapping the submit) was measured and all three are dead.

The blocker on trying a cheaper model was never latency, it was quality, and there
was no instrument for it. The duration histogram grades a candidate's SPEED within
one scrape but cannot express whether it AGREES with the incumbent, so the question
kept being put to a moderation-quality owner with no data attached. This produces
that data and nothing else.

On a sampled share of calls it issues a second, non-blocking request to a candidate
model and records how the two verdicts compare. The verdict of record is always the
incumbent's; nothing reads a shadow result back.

The two disagreement directions are separate outcomes rather than one "diverged"
label, and that is the main design decision here. candidate_permissive means the
incumbent flagged and the candidate did not — the candidate would have let a prompt
through a fail-closed gate, which is the number the decision actually turns on.
candidate_strict is the reverse and is a false-positive cost. Folded into one label
they cancel: a candidate 2% more permissive and 2% stricter would read as agreeing
perfectly.

Off by default and inert on merge. It needs both EXTERNAL_MODERATION_SHADOW_MODEL
and a positive EXTERNAL_MODERATION_SHADOW_SAMPLE; neither alone does anything. No
series at all therefore means "not armed", never "the models agree". The sample rate
is the spend control, because every counted comparison is a second billable request.

No prompt is stored or logged. The prompt goes to the candidate in the request body
exactly as the live call already sends it to the same vendor, and is then dropped;
only a bounded outcome label reaches the metric. This shape was chosen specifically
so that comparing models does not require building a prompt corpus.

The verdict derivation moves into its own module so both callers share it. A shadow
comparison that reimplemented that reduction would compare a policy-applied verdict
against a raw one and report the policy's effect as the model's disagreement — a
wrong answer that looks entirely plausible. It is a behaviour-preserving extraction;
the separate module avoids an import cycle, since the probe is imported by
moderation.ts.

Verification. 91 tests pass across the six moderation suites, including the four
pre-existing ones. Typecheck 0 errors, lint 0 problems. Eight mutants were run and
all eight were killed by a named test: sending the incumbent model instead of the
candidate, shadowing the raw prompt instead of the prepared one, collapsing the two
divergence directions, deleting either arming guard, observing the shadow request on
the duration histogram, and two on the extracted policy. The sample-arming guard is
worth a note — it is redundant against sample=0, since Math.random() >= 0 is always
true, so deleting it leaves that test green. NaN is the only case it uniquely catches,
and there is a test for exactly that, which is what kills the mutant.

### DIFF
--- a/src/env/server-schema.ts
+++ b/src/env/server-schema.ts
@@ -685,6 +685,23 @@ export const serverSchema = z
     // to end. `.catch(0)` degrades an unparseable value to OFF, which is the safe direction for a
     // cache in front of a moderation gate.
     EXTERNAL_MODERATION_CACHE_TTL_SECONDS: z.coerce.number().int().min(0).max(3600).catch(0),
+
+    // DARK SHADOW PROBE — compare a CANDIDATE classifier model against the incumbent on a sampled
+    // share of calls, to produce the disagreement rate the "cheaper model?" decision needs. Both
+    // fields are required to arm and neither defaults on; see
+    // src/server/integrations/moderation-shadow-probe.ts.
+    //
+    // 🔴 SAMPLE IS THE SPEND CONTROL, NOT A CONVENIENCE. Every sampled call issues a SECOND billable
+    // classifier request against the production credential, so 1.0 doubles the moderation bill for
+    // as long as it is armed. It is a fraction in [0,1], not a percentage.
+    //
+    // Both use `.catch(...)` rather than `.default(...)`, the same rule the TTL above carries:
+    // `src/env/server.ts` THROWS on an invalid field and env is parsed only at container start, so a
+    // typo would do nothing visible now and CrashLoop the fleet at the next rollout. Degrading to
+    // OFF is the safe direction for a probe that spends money. A non-numeric SAMPLE coerces to NaN,
+    // fails `.min(0)` and lands on 0 = off.
+    EXTERNAL_MODERATION_SHADOW_MODEL: z.string().default('').catch(''),
+    EXTERNAL_MODERATION_SHADOW_SAMPLE: z.coerce.number().min(0).max(1).catch(0),
     BLOCKED_IMAGE_HASH_CHECK: zc.booleanString.optional().default(false),
     MODERATION_KNIGHT_TAGS: commaDelimitedStringArray().default([]),
 
--- a/src/server/integrations/__tests__/moderation-shadow-probe.test.ts
+++ b/src/server/integrations/__tests__/moderation-shadow-probe.test.ts
@@ -48,7 +48,7 @@ vi.mock('~/env/server', () => ({ env }));
 import { serverSchema } from '~/env/server-schema';
 import { extModeration } from '~/server/integrations/moderation';
 import {
-  candidateVocabularyCovers,
+  candidateVerdictIsDeterminate,
   classifyShadowOutcome,
 } from '~/server/integrations/moderation-shadow-probe';
 
@@ -130,6 +130,7 @@ beforeEach(() => {
   // entries, so a case that sets one would otherwise leak into every later case in this file.
   process.env.CACHE_KEY_NAMESPACE = '';
   delete process.env.IS_PREVIEW;
+  delete process.env.NEXT_PUBLIC_IS_PR_PREVIEW;
 });
 
 afterEach(() => {
@@ -296,18 +297,18 @@ describe('shadow probe — vocabulary gate (audit round 1, F1)', () => {
   it('coverage is by OWN property, so a prototype key cannot fake it', () => {
     // `in` walks the prototype chain, so `'toString' in {}` is true and the vocabulary would read
     // as covered on a response carrying nothing of the sort.
-    expect(candidateVocabularyCovers({ categories: {} }, { toString: 'x' })).toBe(false);
-    expect(candidateVocabularyCovers({ categories: { toString: true } }, { toString: 'x' })).toBe(
-      true
-    );
+    expect(candidateVerdictIsDeterminate({ categories: {} }, { toString: 'x' })).toBe(false);
+    expect(
+      candidateVerdictIsDeterminate({ categories: { toString: true } }, { toString: 'x' })
+    ).toBe(true);
   });
 
   it('treats a missing or non-object categories field as not covered', () => {
-    expect(candidateVocabularyCovers({}, PROD_MAP)).toBe(false);
-    expect(candidateVocabularyCovers({ categories: null }, PROD_MAP)).toBe(false);
-    expect(candidateVocabularyCovers({ categories: 'nope' }, PROD_MAP)).toBe(false);
+    expect(candidateVerdictIsDeterminate({}, PROD_MAP)).toBe(false);
+    expect(candidateVerdictIsDeterminate({ categories: null }, PROD_MAP)).toBe(false);
+    expect(candidateVerdictIsDeterminate({ categories: 'nope' }, PROD_MAP)).toBe(false);
     // Threshold mode: no map, nothing to cover.
-    expect(candidateVocabularyCovers({}, undefined)).toBe(true);
+    expect(candidateVerdictIsDeterminate({}, undefined)).toBe(true);
   });
 
   it('is DETERMINACY, not coverage: a present TRUE key settles the verdict despite a missing one', () => {
@@ -316,7 +317,7 @@ describe('shadow probe — vocabulary gate (audit round 1, F1)', () => {
     // makeable and must not be thrown away as `incomparable`. Unreachable on production's
     // single-key map; reachable the moment a second category is configured, which needs no code.
     const twoKeys = { 'sexual/minors': 'a', violence: 'b' };
-    expect(candidateVocabularyCovers({ categories: { 'sexual/minors': true } }, twoKeys)).toBe(
+    expect(candidateVerdictIsDeterminate({ categories: { 'sexual/minors': true } }, twoKeys)).toBe(
       true
     );
   });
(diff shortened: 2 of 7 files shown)


ZacxDev:
## Audit round 1 — fixes landed

Round 1 was a first, full adversarial audit at `10827891b7`. It returned 3 🟡 and 1 🟢, no 🔴. Fixes are in `9da5a6bb0e`.

**F1 (the one that mattered).** Production runs the classifier in category-map mode with a single key, so the app's verdict reduces to whether the response carries that key. A candidate model naming the same category differently returned a verdict of `false` on every call, so every incumbent flag booked as `candidate_permissive` at zero errors — meaning a candidate in perfect agreement and one that is completely blind produced the identical, plausible-looking number. The module's own comment made this worse by telling the reader an unusable candidate shows up as ~100% `error`.

**F4 was not a defect** and has not been treated as one; it is now documented in place so the mutation claim does not imply coverage it lacks.

```audit-claims round=1 audited=10827891b748c8b952e8b57b57ce8f36b412a2ea..9da5a6bb
1. F1: added an `incomparable` outcome and a vocabulary gate that runs before the comparison, so a candidate whose response lacks the configured category-map keys is not recorded as a disagreement.
2. F1: corrected the module comment asserting that an unusable candidate surfaces as ~100% `error`; that reasoning holds for an unrecognised model name and not for a recognised model with a different category vocabulary.
3. F1: vocabulary coverage is tested by own-property rather than `in`, and requires every mapped key, not one of them.
4. F2: the probe now refuses to arm when `IS_PREVIEW === 'true'`, so config inherited from staging into a PR preview cannot arm it.
5. F3: documented that the probe doubles the request rate against a shared vendor quota and that a 429 wave would reach the fail-soft live call; documented that there is no in-process kill switch and disarming needs a pod rollout.
6. F4: documented that the probe-boundary clamp is behaviourally redundant with the metrics-layer clamp and is killed by no mutant of its own.
```

Verification for this round: 98 tests pass (6 suites), typecheck 0 errors, lint 0 problems, and 5 mutants on the new guards each killed by a named test (vocabulary gate disabled; own-property swapped for `in`; preview guard deleted; `every` swapped for `some`; `incomparable` collapsed into `error`).

ZacxDev:
## Audit round 2 (delta) — fixes landed

Delta re-audit of `10827891b7..9da5a6bb0e`. Returned 1 🟡 + 1 🟢. Fixes in `e87f0c37d5`.

**F1 was written by round 1's own fix**, which is the shape this ladder exists to catch. The `IS_PREVIEW` guard refused the staging rehearsal its comment promised to protect — `civitai-next`, the config source previews copy from, sets `IS_PREVIEW=true` itself. Verified live: `civitai-next` → `IS_PREVIEW=true`, `CACHE_KEY_NAMESPACE=next`; a live preview → `IS_PREVIEW=true`, `CACHE_KEY_NAMESPACE=preview`. The comment's `isNonProductionDatabase()` comparative was also inverted, and has been deleted rather than reversed.

The new discriminator is not an invention: `cache-key-prefix.ts` and `env/database-target` both document this exact `IS_PREVIEW` conflation and both resolved it with an explicit namespace.

```audit-claims round=2 audited=9da5a6bb0e46688d2eed20d680e364a36b93607c..e87f0c37
1. F1: the preview guard now discriminates on `CACHE_KEY_NAMESPACE === 'preview'` instead of `IS_PREVIEW === 'true'`, so the standing non-production deployment (namespace `next`) can be armed.
2. F1: deleted the inverted `isNonProductionDatabase()` comparative rather than reversing it, and replaced it with the measured deployment facts.
3. F1: a narrow `IS_PREVIEW` fallback remains for the transitional state `cache-key-prefix.ts` warns about — namespace unset on a deployment that carries the flag — so an unset namespace cannot fail open.
4. F1: the namespace is read from `process.env` with the same `?.trim() ?? ''` normalisation as `cache-key-prefix.ts`, because the exported constant is module-eval-time and this is a per-call guard.
5. F2: the vocabulary gate now tests DETERMINACY rather than coverage — one present `true` mapped key settles the verdict, so that comparison is no longer discarded as `incomparable`. Unreachable on production's single-key map; reachable on adding a second category.
```

Verification: 102 tests pass (6 suites), typecheck 0, lint 0. Five mutants each killed by a named test, including **restoring the exact F1 regression** — now caught in both directions, by the preview-refusal test and by the staging-arm test.

ZacxDev:
## Audit round 3 (delta) — fixes landed

Delta re-audit of `9da5a6bb0e..e87f0c37d5`. Returned 2 🟡 + 2 🟢. Fixes in `8c7dcda559`.

**F1 is the third round running in which the defect was the sentence a fix wrote to explain itself.** Round 2's comment claimed it followed the precedent of `cache-key-prefix.ts` and `env/database-target`. Both diagnose the `IS_PREVIEW` overload — that half stands — but each resolved it by introducing *its own dedicated variable*, and each says so in terms. Reusing one of their variables is the opposite of that precedent, and the sibling probe on this path had already rejected that exact reuse three days earlier. **Retracted in place rather than replaced**, with a note not to reach for a fresh justification for the same shape.

**F4 is settled, not left open.** Production is single-key — read live: `sexual/minors`, 41 bytes, matching the 41-byte ciphertext the audit measured independently. So the determinacy change is latent, not live.

```audit-claims round=3 audited=e87f0c37d5e1dc158cbfdaa73c74306de563d6fb..8c7dcda5
1. F1: retracted the precedent claim in place, and recorded that both cited modules introduced their own dedicated variable rather than reusing another's.
2. F1: added `NEXT_PUBLIC_IS_PR_PREVIEW === 'true'` as a second refusal clause, checked ALONGSIDE the namespace, because a per-preview keyspace (`preview-<PR>`) would silently un-refuse every preview.
3. F1: stated plainly what the guard rests on — two signals this module does not own and no gate pins — and recorded that the structurally correct dedicated variable is deliberately not taken, because it would fail open until an infra strip-filter change lands.
4. F2: `incomparable` now documented as "verdict not DECIDABLE", not "vocabularies differ"; the metric HELP records that `incomparable == 0` does NOT imply vocabulary agreement, only that every counted comparison was decidable.
5. F2: renamed the predicate `candidateVocabularyCovers` → `candidateVerdictIsDeterminate` to match what it tests.
6. F3: pinned the `.trim()` normalisation with a whitespace case — a padded `" preview"` matches neither the literal nor empty, so without it neither clause fires and the probe arms in a preview.
7. F4: settled the single-key premise by live measurement rather than leaving it unverified.
```

Verification: 104 tests pass (6 suites), typecheck 0, lint 0. Three mutants each killed by a named test — **including the `.trim()` mutant that SURVIVED round 3**, which now dies.

Also fixed while in the file: the backtick-for-apostrophe in the metric HELP that round 1 flagged as below the findings bar.

ZacxDev:
## Audit round 4 (delta) — fixes landed, and **the ladder ends here**

Delta re-audit of `e87f0c37d5..8c7dcda559`. Returned 2 🟡 + 1 🟢, **no 🔴**. Fixes in `f2c092b480`.

```audit-claims round=4 audited=8c7dcda559ee1f0b45e6dc1e0e4b4a2b0e5d3e3a..f2c092b4
1. F5: the remediation instruction now names the preview deploy task's jq override chain, not its strip filter — the strip filter only removes inherited explicit env entries and is a no-op for a variable that does not exist yet, so following the old sentence would have armed every preview.
2. F6: deleted the `contest-score.service.ts` citation rather than replacing it. That file reads the variable as an explicitly legacy fallback for a database-class question, de-authorised once DATABASE_ENVIRONMENT is set — a counter-example, not a precedent. No precedent is claimed now.
3. F7: swept the renamed meaning at all three remaining sites, including the describe block whose title contradicted its own test names.
4. From the sweep rather than the report: an already-off-by-one line-number cross-reference is now unnumbered; the ambiguous "no test pins these" now distinguishes this module READING the signals from infra SUPPLYING them; the metric HELP no longer disclaims an inference that is true under a single-key policy.
```

`round 4 · payload lines changed THIS round: 49 (probe 29+/15−, metrics 3+/2−) · scaffolding: 2`. 104 tests, typecheck 0, lint 0. **No behavioural change in this commit.**

---

### 🔴 Why this ladder is being stopped deliberately, rather than having converged

Stating this explicitly, because a report that *ended* a ladder on this basis is otherwise indistinguishable from one that ran out of findings — and those are opposite meanings.

Round 4 found real things, so the findings-keyed stop rule did **not** fire. I am stopping on the prose criterion instead, and here is the named reason the rounds would not stop on their own:

- **Every round-4 finding was prose-only.** Rounds 2 and 3 each found a genuine behavioural defect; round 4 found none.
- **The payload has gone to commentary.** Round 4's fixes were **1 genuinely new executable line** against ~52 comment lines. This round's fixes are **0** executable lines.
- **The findings are now about the ladder's own writing.** F5 and F6 were defects in the *retraction paragraph* — i.e. commentary about an earlier round's commentary. A round 5 would re-read the paragraph this commit just rewrote, and there is no reason to expect that to terminate.
- No 🔴 at any point in rounds 2–4, and no blast radius beyond "a sentence in this file is wrong".

Per the ladder's own attribution rule, I also swept the recurring **shape** — an unverified claim about another file — at *every* site in the module rather than only the two that were reported, which is what should stop it regenerating.

### What I am deliberately NOT fixing — open, not absent

1. **The guard's coupling itself.** It rests on `CACHE_KEY_NAMESPACE` and `NEXT_PUBLIC_IS_PR_PREVIEW`, both owned by infra, and **no gate in either repo pins their supply**. A dedicated variable is the structurally correct fix and is *not* taken, because a preview would never receive a brand-new variable, so the probe would either refuse everywhere or fail open in every preview. Closing it means a coordinated change in both repos; the code comment says so at the site.
2. **`=== 'true'` is not pinned.** Weakening it to a bare truthiness check passes all 104 tests. The direction is over-refusal — the probe stays dark, no spend, no metric — nothing sets these variables to a non-`'true'` value, and the pre-existing `IS_PREVIEW` clause has identical laxity. Left as-is.
3. **Whether the infra gate `validate-preview-env-shadow.sh` would reject a future dedicated variable** — read at the header/computation level across rounds, never exercised against a hypothetical change.

### Ladder summary

| round | findings | behavioural? | outcome |
|---|---|---|---|
| 1 (full) | 3 🟡 1 🟢 | **yes** — a candidate in perfect agreement was reported as missing every flag | fixed |
| 2 (delta) | 1 🟡 1 🟢 | **yes** — the preview guard blocked the staging arm it promised to protect | fixed |
| 3 (delta) | 2 🟡 2 🟢 | **yes** — an unguarded `.trim()` would arm the probe in a preview | fixed |
| 4 (delta) | 2 🟡 1 🟢 | **no** — all prose | fixed; ladder stopped |

Behaviour is unchanged since `8c7dcda559`. The PR is inert on merge and, in my view, ready for a merge decision.

PR #4688, opened 2026-09-07
fix(clickhouse): widen reaction/report enums and guard every tracker enum column

ZacxDev:
## The class

The ClickHouse tracker rejects a row **client-side**, while serializing the batch, when the target column's enum does not carry the value. The app POSTs fire-and-forget, so:

- the caller sees a successful send,
- nothing is logged app-side,
- `SHOW CREATE TABLE` is clean, because the DDL genuinely is correct,
- and the metric reads `0` — indistinguishable from "nobody did this".

`actions.type` has had a drift guard since 2026-08-21. **Nothing else did**, and that narrowness is why the gaps below survived.

## What was found (measured on production, 2026-09-07)

| column | missing | evidence |
|---|---|---|
| `reactions.nsfw` | `Blocked` | **7 of 10 tracker rejections in a 2h38m window, all real users** — on the order of ~387 rows/day discarded |
| `reports.reason` | `Spam`, `StickerPlacement` | same sweep; both reachable from the report form |
| `reactions.type` | `Post_Create`, `Post_Delete` | same sweep; emitted by the reaction controller's `case 'post':` branch |
| `reports.entityType` | `challenge`, `comicProject`, `model3d`, `model3dReview` | **found while writing the new guard** — not in the original sweep |

## 🔴 Production state — what is applied and what is not

**Already applied by hand on 2026-09-07T23:00:20Z, tracker restarted** (recorded in the migration for history; re-applying is a no-op):

- section 1 — `reactions.nsfw` += `'Blocked' = 5`
- section 2 — `reports.reason` += `'Spam' = 8`, `'StickerPlacement' = 9`

**Also already applied — `2026-09-08T04:04:14Z`, i.e. AFTER this PR was first written:**

- section 5 — `reports.entityType` += `'challenge' = 13` … `'model3dReview' = 16`

  ⚠️ Corrected in `b45986c91d`. An earlier revision of this PR body and of the migration's section 5 header said this was unapplied and that reports on those four entity types "are dropped by the tracker today". **That is no longer true.** The DDL in the file matches live byte-for-byte, so nothing about the statement changed — only its stated status. A tracker restart after that ALTER was **not** observed, so whether the tracker is actually writing those four values is still open; the migration carries a positive-control query for it rather than assuming.

**NOT applied. Needs a human to run it, then restart the tracker service so it re-reads the column schema:**

- section 3 — `reactions.type` += `'Post_Create' = 17`, `'Post_Delete' = 18`
- section 4 — `reactions_owner_scores_mv` `MODIFY QUERY`

The sections are numbered in the order they were written, not the order they were applied, which is why an applied section sits below two unapplied ones. Each section header carries its own status.

This PR applies **no** DDL. Nothing auto-runs it.

## 🔴🔴 Sections 3 and 4 are ONE operation. Apply both or neither.

`reactions_owner_scores_mv` scores `type IN (<the '*_Create' list>)` as **+1 and everything else as −1**. There is no unknown arm.

So widening `reactions.type` **without** widening that list does not merely fail to count post reactions — **every post reaction would DECREMENT the content owner's score.** That converts the current, recoverable failure (rows are dropped; nothing was written) into an unrecoverable one (rows land with the wrong sign into an aggregate with no source of truth to rebuild from).

If you can only do one, do **neither**. Another day of dropped post reactions is strictly better than an hour of corrupted owner scores.

**"In one sitting" is not, by itself, protection.** The tracker rebuilds its column serializers from the schema it reads at connect time, so the moment any tracker pod restarts after section 3 lands it starts *accepting* `Post_Create` with section 4 still unapplied — and that deployment has been observed replacing its pods unprompted on the order of every few hours. So the interval between the two statements is a real window, and working quickly narrows it rather than closing it. A materialized view does not backfill: rows landing in that window are scored −1 **permanently**, and section 4 repairs none of them. The migration's POST-APPLY block now carries the query that detects it.

The two statements sit next to each other in the migration file, and `tracker-enum-drift.test.ts` asserts **set equality** between the column's `*_Create` values and the view's +1 list, in both directions.

### The MV statement is pinned WHOLE, not through its IN list

`MODIFY QUERY` replaces the **entire SELECT** of a live view maintaining every user's all-time reaction score (`reactions_owner_scores`, read by `getReactionTasks` in `src/server/metrics/user.metrics.ts`). The set-equality check above covers the IN list and nothing else — and the two most damaging edits both leave the IN list untouched:

- swapping the `multiIf` arms, `, 1, -1)` → `, -1, 1)`, scores **every** reaction −1;
- moving `parseDateTimeBestEffort('2024-04-27')` forward disables the self-reaction exclusion, so a user can farm their own score.

Applied together, those two left the suite **fully green**. So `b45986c91d` adds three things: the live pre-image is recorded verbatim in the migration's section 4 header (giving a rollback text, and making the former "restated byte-for-byte from the live definition" comment a checkable claim instead of load-bearing prose); the post-apply step is a **full diff against that pre-image** rather than a grep for one value; and the guard pins the **whole whitespace-normalised statement** against that pre-image plus the single added value. A companion assertion requires the +1 list to come from exactly one `type IN (...)` occurrence. ⚠️ Corrected after review: pinning the statement makes that assertion **strictly dominated**, not newly reachable — a body with more than one `type IN (...)` list also differs from `EXPECTED_MV_SELECT`, so the pin fires on it too (re-running M14 without skipping the pin fails **2** tests, not 1). It is kept for a specific, actionable failure message rather than for unique coverage.

## Why the type system never saw the `reactions.type` gap

`getTrackerEvent` builds its `type` as `` `Post_${action}` ``. A template-literal expression infers as plain `string` unless something contextually types it — and the call site laundered it through `as ReactionType`. So the one branch emitting a value the column has never carried type-checked for as long as it has existed.

The cast is removed and `getTrackerEvent` now carries an explicit return type, which contextually types every branch's constructed string against the union. All eight branches were enumerated (`image`, `post`, `article`, `commentOld`, `comment`, `question`, `answer`, `bountyEntry` × `Create`/`Delete`); `post` was the only one outside the union.

`ReactionType` and `ReportType` changed from bare unions to `as const` arrays (the `ActionType` pattern already in this file). A union cannot be enumerated at runtime, which is exactly why those domains were structurally unguardable.

## The durable fix

New `src/server/clickhouse/__tests__/tracker-enum-drift.test.ts` — **31 tests** — checks **seven** columns, each against the app-side domain that feeds it:

| column | domain |
|---|---|
| `reactions.type` | `ReactionType` (tracker) |
| `reactions.reaction` | `ReviewReactions` (Prisma) |
| `reactions.nsfw` | `NsfwLevelDeprecated` |
| `reports.type` | `ReportType` (tracker) |
| `reports.entityType` | `ReportEntity` |
| `reports.reason` | `ReportReason` (Prisma) |
| `reports.status` | `ReportStatus` (Prisma) |

Per column it asserts: every app value is in the effective definition (newest migration that states the column, else the measured production baseline); every migration restating the column reproduces the baseline names **at their existing indices** (a `MODIFY COLUMN` replaces the definition — a dropped name deletes a live value, a moved index silently remaps rows); indices are distinct; and the same verbatim `POST-APPLY:` marker the `actions.type` guard pins.

Positive controls, because every `it.each` here is driven by parsed data and a regex that matches nothing would make the whole file a green meaning "scanned zero columns": the directory has ≥8 `.sql` files; ≥5 **multi-line** `ALTER … MODIFY COLUMN` blocks parsed (these statements span two lines — a single-line regex matches zero of them); a block exists for each column this change defines, named explicitly rather than counted; the MV `MODIFY QUERY` was found; and ≥1 migration exists for the marker check to run over.

## Red-at-base / green-at-HEAD

`origin/main` @ `14302e9a9d`.

**Environmental control first** (a fresh worktree has neither `node_modules` nor the `event-engine-common` submodule; both were installed/initialised before anything below was believed): after `pnpm install` + `git submodule update --init event-engine-common`, `pnpm typecheck` on the **unmodified** tree reports `typecheck: OK — 0 type errors`. Every result below is against that baseline.

| state | result |
|---|---|
| base code + new test file only | **10 failed / 18 passed (28)** |
| base code + the `as const` conversion (16 values, no `Post_*`), no migration file | **8 failed / 20 passed (28)** |
| HEAD | **31 passed (31)**; whole `src/server/clickhouse` dir 7 files / **66 passed**; `typecheck: OK — 0 type errors` |

(The `29`/`64` figures were HEAD before `b45986c91d`, which added two assertions.)

The three live gaps red at base with their own messages, independent of any other part of this change:

```
default.reactions.nsfw cannot accept 1 value(s) the app emits: Blocked.
default.reports.reason cannot accept 2 value(s) the app emits: Spam, StickerPlacement.
default.reports.entitytype cannot accept 4 value(s) the app emits: challenge, comicProject, model3d, model3dReview.
```

**Stated honestly:** `reactions.type` and `reports.type` are **green at base**, and that is correct — at base those unions were fully covered by the live column, so there was no drift to see. Their guards are exercised by mutation instead (M1, M3, M10 below). In the first state, those two columns fail with `Cannot read properties of undefined` rather than an assertion, because `ReactionType`/`ReportType` are type-only exports at base; that is a red for the wrong reason and is not counted as evidence.

## Mutation matrix

Every mutation was applied to the real tree, run, and reverted. Harness control: unmutated → `29 passed`, no failures.

| # | mutation | killed by | isolated? |
|---|---|---|---|
| M1 | drop `Post_Create`/`Post_Delete` from the `ReactionType` array | `pnpm typecheck` → `reaction.controller.ts(89,11): error TS2322: Type '"Post_Create" \| "Post_Delete"' is not assignable to …` — **1 type error**, at the `post` branch | yes |
| M2 | drop `'Post_Create'` from the MV +1 list | `scores every reactions.type *_Create value +1 in reactions_owner_scores_mv` → "the +1 list and the reactions.type '*_Create' values disagree … apply both or neither" | 1 failed / 28 passed |
| M3 | drop `'Post_Delete' = 18` from the `reactions.type` ALTER | `every default.reactions.type value the app can emit exists in the column` → "cannot accept 1 value(s) the app emits: Post_Delete." | 1 failed / 28 passed |
| M4 | renumber `'Image_Create' = 1` → `= 99` | `every migration restating default.reactions.type reproduces the baseline indices exactly` → "moves 'Image_Create' from 1 to 99" | 1 failed / 28 passed |
| M5 | drop `'Blocked' = 5` from the `nsfw` ALTER | `every default.reactions.nsfw value …` → "cannot accept 1 value(s) the app emits: Blocked." | 1 failed / 28 passed |
| M6 | drop `'StickerPlacement' = 9` from the `reason` ALTER | `every default.reports.reason value …` → "cannot accept 1 value(s) the app emits: StickerPlacement." | co-run with M7 |
| M7 | reword the `POST-APPLY:` line to a sentence that still says "restart" | `… carries the POST-APPLY tracker-restart marker verbatim` → "widens a tracker-written enum but does not carry the post-apply marker." | co-run with M6 |
| M8 | narrow the parser's `\s+` to `[ \t]+` (the single-line-regex trap) | `parses multi-line ALTER ... MODIFY COLUMN statements` → "expected 0 to be greater than or equal to 5", plus 7 downstream | control, not isolated — by design |
| M9 | add `Widget = 'widget'` to `ReportEntity` | `every default.reports.entitytype value …` → "cannot accept 1 value(s) the app emits: widget." | co-run with M10 |
| M10 | add `'Reopened'` to the `ReportType` array | `every default.reports.type value …` → "cannot accept 1 value(s) the app emits: Reopened." | co-run with M9 |

M10 matters specifically: `reports.type` is a column **no migration touches**, so it is checked purely against the recorded baseline. The mutation proves that path is live rather than vacuous.

M8 is the positive control for the multi-line trap. Under it the suite reports **28** tests rather than 29, because the marker `it.each` is driven by parsed files and had zero cases — which is exactly what the "found migrations touching these columns to check for the marker" assertion exists to catch, and it fired.

### Mutation matrix — round 2 (`b45986c91d`)

Run in a `cp -a` copy with its `.git` removed, against a **31-pass harness control**. M1–M10 above predate these assertions and were not re-run.

| # | mutation | killed by | result |
|---|---|---|---|
| M11 | `multiIf(…, 1, -1)` → `multiIf(…, -1, 1)` in the MV `MODIFY QUERY` | `reactions_owner_scores_mv MODIFY QUERY is the live view plus exactly the one added value` → "does not match the live view definition with 'Post_Create' added and nothing else changed" | **1 failed / 30 passed** — isolated |
| M12 | `parseDateTimeBestEffort('2024-04-27')` → `('2099-01-01')` | same assertion, same message | **1 failed / 30 passed** — isolated |
| M13 | M11 **and** M12 together — the exact pair that was previously fully green | same assertion | **1 failed / 30 passed** |
| M14 | add a second `type IN (…)` list to the MV `WHERE`, **with the statement pin skipped** so it cannot shadow | `derives the +1 list from exactly one type IN (...) list` → "the MODIFY QUERY contains 2 `type IN (...)` lists" | **1 failed / 29 passed / 1 skipped** — reachable on its own |
| M15 | pass `entityType: 'widget'` (outside `ReportEntity`) at the only `Tracker.report` call site | `pnpm typecheck` → `report.controller.ts(49,9): error TS2322: Type '"widget"' is not assignable to type 'ReportEntity'` | **1 type error** |
| M15-control | the identical mutation with `entityType` reverted to the old bare `string` | nothing | **`typecheck: OK — 0 type errors`** |

M11–M13 matter most: under every one of them the **pre-existing coupling test stayed green**, alongside all 29 other assertions. That is the measurement showing the IN-list-only guard could not see any of these changes, and it is why the pin is a whole-statement equality rather than a wider regex.

M15-control is the negative control for the `ReportEntity` change: the same out-of-domain value is **silently accepted** under the old typing and rejected under the new one, which is the type-laundering this PR removes from the reaction path, left on the report path.

## Not verified

- **This PR runs no DDL.** Sections 3 and 4 are unapplied by design; nobody has observed a `Post_Create` row land, or the widened MV recompute. Sections 1, 2 and 5 were applied by hand outside this PR and are asserted here only as history.
- **Section 5's tracker restart was not observed.** The ALTER landed `2026-09-08T04:04:14Z`; whether the tracker has re-read the schema and is actually writing `challenge` / `comicProject` / `model3d` / `model3dReview` is **open**. The migration carries a query for it, and states why a zero from that query would be ambiguous rather than reassuring.
- The enum definitions and the MV pre-image were transcribed from live `SHOW CREATE TABLE` reads (`2026-09-07`, MV re-read `2026-09-08`). They should be re-read immediately before applying rather than trusted from this file.
- **The suite cannot see production, and is green on the very defect it was written for.** It reads migration *text*: `effectiveArms` returns the newest migration that **states** a column, applied or not. Right now `Post_Create` is stated, unapplied, and rejected in production, and every assertion passes. The `effectiveArms` docstring says this plainly. It is a **disclosure, not a repair** — no change to this file could close it, because the evidence lives in a system the test does not talk to.
- No integration test exercises the tracker's serializer. If production drifts by some path other than a migration in this directory, the baselines here go stale silently.
- The claim that the tracker deployment rolls its pods unprompted "every few hours" comes from operational observation, not from a measured distribution over time.

### DIFF
--- a/src/server/clickhouse/__tests__/tracker-enum-drift.test.ts
+++ b/src/server/clickhouse/__tests__/tracker-enum-drift.test.ts
@@ -1,0 +1,422 @@
+import fs from 'fs';
+import path from 'path';
+import { describe, expect, it } from 'vitest';
+
+import { ReactionType, ReportType } from '~/server/clickhouse/tracker';
+import { NsfwLevelDeprecated } from '~/shared/constants/browsingLevel.constants';
+import { ReportEntity } from '~/shared/utils/report-helpers';
+import { ReportReason, ReportStatus, ReviewReactions } from '~/shared/utils/prisma/enums';
+
+/**
+ * Enum drift on every tracker-written column EXCEPT `actions.type`.
+ *
+ * `actions.type` has had a guard since 2026-08-21 (./action-type-enum-drift.test.ts).
+ * Nothing else did — and that narrowness is the entire reason this file exists. A sweep of
+ * tracker rejections over a 2h38m window on 2026-09-07 found three live gaps, none of them
+ * on `actions`:
+ *
+ *   * `reactions.nsfw` lacked 'Blocked' — 7 of 10 rejections in that window, all real
+ *     users, on the order of ~387 rows/day silently discarded.
+ *   * `reports.reason` lacked 'Spam' and 'StickerPlacement'.
+ *   * `reactions.type` lacks 'Post_Create' and 'Post_Delete'.
+ *
+ * A fourth, `reports.entityType`, was found by writing this guard.
+ *
+ * 🔴 WHY NOTHING ELSE CATCHES IT. The rejection is CLIENT-SIDE, inside the tracker service,
+ * while it serializes the batch — the app POSTs fire-and-forget, so the caller sees success,
+ * nothing is logged here, and the row never exists. `SHOW CREATE TABLE` is clean because the
+ * DDL genuinely is. Typecheck was clean too: the reaction controller built its `type` with a
+ * template literal (inferred `string`) and laundered it through `as ReactionType` at the call
+ * site, so the one branch emitting a value the column has never carried type-checked for as
+ * long as it has existed. That cast is gone; this file is what stops the class coming back.
+ *
+ * THE PROPERTY: for every enum column the tracker writes, every value the APP can emit must
+ * be present in the column's effective definition — the newest migration that states it, or,
+ * for a column no migration has touched, the measured production baseline below.
+ *
+ * Containment, not equality: a column is allowed to be WIDER than the app domain
+ * (`reactions.nsfw` carries 'Undefined', which no app value maps to). Narrower is the defect.
+ */
+
+const MIGRATIONS_DIR = path.resolve(__dirname, '../migrations');
+
+const sqlFiles = fs
+  .readdirSync(MIGRATIONS_DIR)
+  .filter((f) => f.endsWith('.sql'))
+  .sort()
+  .map((f) => ({ name: f, raw: fs.readFileSync(path.join(MIGRATIONS_DIR, f), 'utf-8') }));
+
+/**
+ * 🔴 Comments are stripped BEFORE anything is parsed. These migrations carry their values in
+ * prose as well as in DDL — a rationale, a verification snippet, a POST-APPLY block — so a
+ * parser fed the raw file stays green on a migration whose ALTER was deleted and whose
+ * comments were not. Measured on the sibling guard: removing an arm left it passing.
+ */
+const stripComments = (raw: string) =>
+  raw
+    .split('\n')
+    .filter((line) => !line.trim().startsWith('--'))
+    .join('\n');
+
+/**
+ * 🔴 MULTI-LINE BY CONSTRUCTION. Every migration in this directory writes the statement as
+ * `ALTER TABLE default.reactions\n  MODIFY COLUMN ...`, and the enum arms are one per line.
+ * A regex anchored to a single line matches ZERO of them — and a guard built on it scans an
+ * empty set and passes on everything, which is the same reassuring zero this file exists to
+ * stop. `\s+` spans newlines; `parsedBlockCount` below is the control that proves it did.
+ *
+ * Enum8 and Enum16 are both accepted: `actions.type` is Enum16, every column here is Enum8,
+ * and a guard that silently skipped one width would have exactly the blind spot above.
+ */
+const ENUM_COLUMN_RE =
+  /ALTER\s+TABLE\s+([\w.]+)\s+MODIFY\s+COLUMN\s+`?(\w+)`?\s+Enum(?:8|16)\s*\(([^)]*)\)/gi;
+
+type EnumBlock = {
+  file: string;
+  column: string; // "<table>.<column>", lower-cased
+  arms: Map<string, number>;
+  multiLine: boolean;
+};
+
+const enumBlocks: EnumBlock[] = sqlFiles.flatMap(({ name, raw }) =>
+  [...stripComments(raw).matchAll(ENUM_COLUMN_RE)].map((m) => ({
+    file: name,
+    column: `${m[1]}.${m[2]}`.toLowerCase(),
+    arms: new Map<string, number>(
+      [...m[3].matchAll(/'([A-Za-z0-9_]+)'\s*=\s*(\d+)/g)].map((a) => [a[1], Number(a[2])])
+    ),
+    multiLine: m[0].includes('\n'),
+  }))
+);
+
+/**
+ * The `reactions_owner_scores_mv` +1 list. Parsed, not substring-matched, because the
+ * property that matters is set equality against the column's '*_Create' values — see the
+ * coupling test at the bottom of this file for why that is not cosmetic.
+ */
+const MV_QUERY_RE =
+  /ALTER\s+TABLE\s+default\.reactions_owner_scores_mv\s+MODIFY\s+QUERY([\s\S]*?);/gi;
+
+const mvBlocks = sqlFiles.flatMap(({ name, raw }) =>
+  [...stripComments(raw).matchAll(MV_QUERY_RE)].map((m) => {
+    const inList = /type\s+IN\s*\(([^)]*)\)/i.exec(m[1]);
+    return {
+      file: name,
+      plusOne: new Set<string>(
+        inList ? [...inList[1].matchAll(/'([A-Za-z0-9_]+)'/g)].map((a) => a[1]) : []
+      ),
+    };
+  })
+);
+
+/** The newest block wins: `MODIFY COLUMN` replaces the definition, and files are date-named. */
+const lastBlockFor = (column: string) => {
+  const matching = enumBlocks.filter((b) => b.column === column);
+  return matching[matching.length - 1];
+};
+
+type ColumnSpec = {
+  /** "<table>.<column>", lower-cased — the key the parser produces. */
+  column: string;
+  /** Where the app-side domain lives, for the failure message. */
+  domain: string;
+  /** Every value the app can emit into this column. */
+  appValues: readonly string[];
+  /**
+   * The production definition as it stood BEFORE any migration in this directory touched
+   * the column, measured with `SHOW CREATE TABLE` on 2026-09-07.
+   *
+   * 🔴 Do NOT add a name here to silence a failing case. That is the one-line bypass of this
+   * whole guard and it reproduces precisely the "looks instrumented, writes no rows" outcome
+   * the file exists to stop — the value would be in the app and still absent from the column.
+   * A new value belongs in a migration.
+   */
+  baseline: ReadonlyMap<string, number>;
+};
+
+const COLUMNS: readonly ColumnSpec[] = [
+  {
+    column: 'default.reactions.type',
+    domain: 'ReactionType (src/server/clickhouse/tracker.ts)',
+    appValues: ReactionType,
+    baseline: new Map([
+      ['Image_Create', 1],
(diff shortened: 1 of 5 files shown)


ZacxDev:
**Correction to this PR's description**, surfaced by a round-2 delta audit. Posting rather than only editing the body, since a reviewer may already have read the wrong version.

The description said, of the companion assertion added in `b45986c91d`:

> A companion assertion requires the +1 list to come from exactly one `type IN (...)` occurrence, which only becomes reachable once the statement is pinned.

**That is backwards.** Pinning the whole normalised statement makes that assertion *strictly dominated*, not newly reachable: any MV body carrying more than one `type IN (...)` list necessarily also differs from `EXPECTED_MV_SELECT`, so the pin fires on it too. The assertion was reachable as a unique signal *before* the pin existed, and is no longer.

Measured, not reasoned: re-running M14 **without** skipping the pin fails **2** tests, not 1. The M14 row in the mutation table is itself accurate — it states the pin was skipped so it could not shadow — so the table and this sentence disagreed, and the table was the correct one.

The assertion is worth keeping regardless, for a reason the original sentence did not give: it produces a specific, actionable failure message (`the MODIFY QUERY contains 2 type IN (...) lists`) where the pin alone would only report a whole-statement mismatch. That is defence in depth and better diagnostics — not unique coverage.

The body has been corrected to say this. No code changed for it.

---

Separately, `2608b83860` fixes an operator-facing defect the same audit found: the MV pin's failure message told the reader that `EXPECTED_MV_SELECT` and the migration's section 4 header hold "the same pre-image" and to "re-read it, update BOTH". They are deliberately different strings — `EXPECTED_MV_SELECT` is the **post**-image (with `'Post_Create'`), the section 4 header is the **pre**-image (without it) and is the rollback text. After section 4 is applied a fresh `SHOW CREATE TABLE` returns the post-image, so an operator following that instruction would overwrite the rollback text with the state it exists to undo.

PR #4691, opened 2026-09-08
fix(csam): bound archiver memory when archiving a large image library

ZacxDev:
## Problem

Archiving a CSAM report for a user with a very large image library exhausts the container's memory. The process is killed partway through, taking the rest of that hour's cron work down with it, and because the job is rescheduled it repeats on every run — so the archive job never completes once such a report enters the queue.

There are **two independent unbounded paths** in `archiveCsamDataForReport()`, and round 1 of this PR only fixed one of them. Both are fixed now.

### 1. The archiver queue (`archiveImages` / `archiveGeneratedImages`)

These download every image the reported user owns, fully buffer each one (`blob.arrayBuffer()` → `Buffer.from`), and call `archive.append(buffer, …)`.

`archive.append()` is fire-and-forget: it pushes the source onto an **unbounded** internal queue drained by a single worker (`archiver@6.0.2`, `lib/core.js` — `this._queue = async.queue(…, 1)`). With a fast producer (network) in front of a slow consumer (level-9 deflate over already-compressed JPEG/WebP), every source buffer is retained at once and resident memory tracks *total downloaded bytes*.

The existing `plimit(10)` bounds concurrent **downloads** only — nothing bounds the compressor. And `NODE_OPTIONS=--max-old-space-size` offers no protection: the retained buffers are external/off-heap, so the V8 old-space cap never engages before the container limit does.

Measured directly against the installed `archiver@6.0.2` before writing the fix:

```
unbounded peak outstanding entries: 40 of 40    <- every appended buffer retained
bounded peak outstanding:            4          <- awaiting 'entry' caps it, all 40 still processed
```

### 2. The base evidence bundle (`archiveBaseReportData`) — new in round 2

```ts
const images = await dbRead.image.findMany({ where: { userId } });   // every row, no pagination
const models = await dbRead.model.findMany({ where: { userId } });   // ditto
const modelVersions = await dbRead.modelVersion.findMany({ … });     // ditto
…
await fsAsync.writeFile(outPath, JSON.stringify({ user, reportId, models, modelVersions, images }, replacer));
```

Two ceilings, and the second is worse than the first:

- **Soft:** all three full row arrays and the rendered string are resident simultaneously.
- **Hard:** `JSON.stringify` returns **one JavaScript string**, and a JS string cannot exceed V8's maximum length (536,870,888 characters on 64-bit — `require('buffer').constants.MAX_STRING_LENGTH`, measured on this Node build). A bundle above that throws `RangeError: Invalid string length` — on every attempt, forever, regardless of how much memory the host has. Such a report can never be archived.

> ⚠️ **Correcting round 1.** This PR's previous description listed the base bundle under "deliberately out of scope", on the grounds that it is *"orders of magnitude too small to be the cause here."* **That claim is withdrawn.** It was derived from one account, and it does not generalise: the row count is unbounded by construction, `Image` rows are wide (including a `meta` JSON blob), and the string ceiling is a hard wall rather than a matter of degree. It is now fixed.

## Changes

**1. Backpressure on the archiver.** New `src/server/utils/archive-helpers.ts`. `createBoundedArchive()` wraps an archiver instance and waits on its `'entry'` event before each append, capping appended-but-unprocessed entries at 8. Peak retained bytes become a function of that cap rather than of how many images the account has.

> The doc comment on that constant now states the **real** peak: each of the 10 `p-limit` slots also holds its `blob` + `arrayBuffer` + `buffer` while parked in `append()`, so the peak is about `8 + 3 × 10 = 38` image-sizes, not 8. Still a constant — which is the whole point — but round 1 understated it by ~5×.

**2. Compression level 9 → 1.** Deflate cannot meaningfully shrink already-compressed image bytes, so level 9 was spending a lot of CPU for close to nothing — while being exactly the slow consumer that let the queue grow. Nothing in the repo reads these archives back or depends on their size; they are uploaded as-is.

**3. `finalize()` is now awaited.** It was never awaited before `fs.createReadStream(outPath)` opened the file for upload, which can upload a truncated or empty archive.

> ⚠️ **Awaiting `archive.finalize()` alone is NOT sufficient.** Its promise resolves when the zip module ends, not when the sink has flushed. Measured: a file was **24254 bytes** when `finalize()` resolved and **25342 bytes** once the write stream emitted `close`. The helper waits for both.

**4. `archiveGeneratedImages()`** had the identical unbounded pattern — same fix.

**5. The base bundle is now paged and streamed.** New `src/server/utils/json-stream-helpers.ts`.

- `scanPagesById()` in the service keyset-pages each `findMany` (`id > cursor ORDER BY id ASC`, 500 rows/page). Keyset rather than offset: offset paging degrades quadratically and can skip or duplicate rows if the set shifts mid-scan.
- `writeJsonObject()` serialises the document straight to the output stream, one array element at a time, so no single string above one row's JSON is ever built. `pipeline` supplies backpressure, and it is also what guarantees the file is whole when the call resolves — it waits on the sink via `stream.finished`, which for an `autoDestroy` Writable (the default, and what `fs.createWriteStream` is) means `'close'`. Round 2 added a second `await` on `'close'` on top of that under a comment claiming `pipeline` gave no such guarantee; **that claim was measured false and the line is now removed** — see round 3 below. The archive's own `'close'` wait is a different case and is kept.

🔴 **The EVIDENCE in the bundle is unchanged.** Same five properties in the same order, the same `where` clauses and **no `select`** (every column still lands in the archive), the same bigint replacer, and the same row set. The byte-identity test asserts that the streamed serialiser produces exactly what `JSON.stringify` produces — it is a claim about the *serialiser*, over one fixed row order.

Two things are genuinely new, and neither adds or removes a row. The queries now also carry `orderBy: { id: 'asc' }`, `take` and an `id > cursor` predicate, because they are paged. And as a consequence, element order within each array is `id ASC` where before it was whatever the query plan happened to yield — so against real data the file is **set-identical but not byte-identical** to what the old code wrote. Determinism is the right property for an evidence artefact, but it is a change, and anything diffing two bundles across this commit will see it.

The one behavioural trade worth naming: `archiveBaseReportData` and `archiveImages` used to share a single in-memory snapshot of the image rows and now run two independent scans. That does not weaken the bundle — `archiveImages` already tolerates listing an image in `data.json` that is missing from `images.zip`, because `fetchBlob` returning null makes it skip the entry silently — and it is what makes the memory bound hold for both callers.

**6. `getCsamsToArchive()` has a deterministic `ORDER BY`, and its comment now says what that actually does.**

> ⚠️ **Correcting round 1.** The comment claimed the ordering stops a bad report sitting "at the front of every batch". That is backwards: a chronically-failing report is by construction among the *oldest* rows still matching the filter, so `createdAt asc` **pins it to the head of every run** — the ordering makes head-of-line blocking deterministic, it does not cure it. The ordering is kept (it is right for auditability and reproducibility) and the comment now says so, and says explicitly that per-report failure isolation is not in scope.

**7. The `'error'` listener no longer rethrows from inside the handler.** `archive.on('error', err => { throw err })` surfaces as an uncaught exception from inside archiver's async internals rather than as a rejection the caller can act on. The error is captured and re-thrown from `append()`/`finalize()` instead.

> ⚠️ **Correcting round 1.** The previous description said *"the outcome for `archivedAt` is unchanged"*. **That is false, and it was measured false.** Before this PR, throwing from the `'error'` handler surfaced synchronously out of `append()` into `archiveGeneratedImages`'s inner `catch (e) {}` — so a bad entry was silently dropped and the report still completed and was stamped. With the error latched and rethrown from `finalize()`, that inner catch can no longer swallow it and the report is never stamped.
>
> Measured on the same input (a `previewUrl` whose basename is empty, e.g. `https://host/blob/` or `https://host/?x`):
>
> | | pre-PR | round 1 |
> |---|---|---|
> | entry names appended | `['', '', 'ok-2.jpeg']` | archiver raises `ENTRYNAMEREQUIRED` |
> | `archiveCsamDataForReport` | resolves | rejects |
> | `archivedAt` | stamped | never stamped — report retries forever |
>
> **The new direction is kept**: for an evidence subsystem, failing loudly beats silently dropping evidence. But "loud" must not mean "permanently unarchivable", so the one known input class that produces an empty name is given a deterministic fallback name (`zipEntryNameForUrl`) instead of poisoning the report. The basename expression is evaluated first and returned verbatim whenever it is non-empty, so every URL that already produced a usable name keeps byte-identical naming.

**8. `uploadStream()` rejects with a real `Error`, and `process-csam.ts` extracts the message safely.** `uploadStream` used to `reject()` with no argument. The rejection value is then `undefined`, and the batch loops' `catch (e: any) { … message: e.message }` raise a **`TypeError` inside the catch block** — which nothing catches, so it escapes the `for` loop and **abandons every remaining report in that batch**. Both loops (`send-csam-reports` and `archive-csam-reports`) had it. Both ends are fixed: the rejection now carries a real `Error` with the original as `cause`, and the catch uses a total `errorMessage(e)`.

**9. Test-suite hygiene.** `csam-archive-backpressure.test.ts` was writing real zips into `<repo>/csam` — which is **not** gitignored — and unconditionally `rm -rf`-ing that path in `afterAll`. That is the same path the service uses as `baseDir` in dev, so a unit run could delete a developer's local scratch files. The suite now redirects `baseDir` into its own `mkdtemp` directory.

## Test coverage

Four files:

- `src/server/utils/__tests__/archive-helpers.test.ts` — pins the archiver mechanism against the **real** archiver with a slow sink. Two **negative controls** (a plain `archive.append` loop, and an un-awaited `finalize`) prove the measurement can actually observe unbounded growth and a short file. Plus `zipEntryNameForUrl`, including a test that runs both arms through the real archiver: the empty name it replaces is rejected, the fallback is accepted.
- `src/server/utils/__tests__/json-stream-helpers.test.ts` — **byte-identity** with `JSON.stringify` over a fixture carrying bigints, `Date`s, a `U+2028`, escaped keys, nested nulls, an `undefined` property, and an unserialisable array element; plus the ceiling proof (below).
- `src/server/services/__tests__/csam-archive-backpressure.test.ts` — pins the **seam**: that the service actually routes through the bounded appender and the paged scans. A correct helper nobody calls fixes nothing, and no isolated test can see that.
- `src/server/jobs/__tests__/process-csam.test.ts` — batch integrity for both cron loops.

### Proving the string ceiling is gone

`JSON.stringify` is temporarily replaced with a wrapper that throws `RangeError: Invalid string length` above a small test-chosen cap. That is not an approximation — it is V8's own failure, made reachable from a fixture that fits in a unit test. The fixture overshoots the cap ~10× while any single row is ~40× under it, so a per-row serialiser is safe by a wide margin and a one-shot serialiser is not.

- **Negative control:** the pre-fix `JSON.stringify(document, replacer)` throws at that cap.
- **Under test:** `archiveCsamDataForReport` completes, the longest string ever produced stays under the cap, and the uploaded `data.json` is compared **byte for byte** against what the pre-fix serialiser produces for the same fixture.

Every new behavioural test was run against the pre-change code and watched to fail. Actual output:

**A — base bundle** (service reverted to `main`, i.e. commit before this branch):

```
× archives a library whose bundle exceeds the maximum JavaScript string length
    RangeError: Invalid string length
      ❯ archiveBaseReportData src/server/services/csam.service-new.ts:1186:12
× never issues an unbounded scan for the rows that go into the bundle
    AssertionError: image.findMany call without a page size: expected 'undefined' to be 'number'
× walks every page rather than archiving only the first
    AssertionError: expected [ { id: 1, …(8) }, …(499) ] to have a length of 1001 but got 500
× filters model versions by the COMPLETE set of model ids, not a partially-scanned one
    AssertionError: expected [ 1, 2, …(491) ] to have a length of 501 but got 500
```

> ⚠️ One test in that group is **not** red-at-base coverage and is labelled as such: *"archives images from EVERY page"*. Against pre-PR code it fails by 60 s timeout (the un-awaited-`finalize` defect trips first), and against round-1 it **passes**, because a fixture answering an unpaged `findMany` with every row cannot distinguish an unpaged read from a correctly paged one. It is a guard on the new paging code — proved reachable by the `break`-after-first-page mutant below — while the unpaged read itself is caught by *"never issues an unbounded scan"*, which fails at base on its own assertion.

**B — batch abort** (job + service reverted to `main`):

```
× continues the batch when a report rejects with a non-Error value
    Caused by: TypeError: Cannot read properties of undefined (reading 'message')
      ❯ archiveCsamReportDataJob src/server/jobs/process-csam.ts:49:22
× still logs something identifiable for the failed report        (same TypeError)
× survives a rejection that is a string / null / a plain object  (3 cases)
× send-csam-reports: continues the batch …                       (same TypeError, line 27)
Tests  6 failed | 2 passed (8)

× rejects with a real Error, so the batch loop can read .message without throwing
    AssertionError: expected undefined to be an instance of Error
```

The service's own `console.log(e)` printed a bare `undefined` in that last run — direct confirmation that the rejection value is `undefined`.

**D — unnameable generated-image URL** (service at round-1 HEAD, i.e. the commit that introduced the behaviour):

```
× archives the entry under a fallback name instead of poisoning the report forever
    ArchiverError: entry name must be a non-empty string value   { code: 'ENTRYNAMEREQUIRED' }
      ❯ Object.append src/server/utils/archive-helpers.ts:155
      ❯ src/server/services/csam.service-new.ts:1295
```

Against **pre-PR** the same test fails differently and that difference is the point — `expected [ '', '', 'ok-2.jpeg' ] to deeply equal [ 'unnamed-0-blob', 'unnamed-1', 'ok-2.jpeg' ]`, reached *after* the call resolved. Two entries silently dropped, report completed. That is the archivedAt behaviour change, measured.

At HEAD: **64 passed (64)** across the four files (59 after round 2, plus round 3's five).

### Mutation checks

Every new guard was broken on purpose and confirmed to fail with **that guard's own** error.

| Mutation | Result |
|---|---|
| `serializeJsonObject`: drop the undefined-property omission | ✅ killed — `expected '{"a":1,"b":undefined,"c":3}' to be '{"a":1,"c":3}'` |
| `serializeJsonObject`: array element `undefined` dropped instead of `null` | ✅ killed — `expected '{"xs":[1,,,4]}' to be '{"xs":[1,null,null,4]}'` |
| `serializeJsonObject`: always emit a leading comma | ✅ killed — `expected '{,"a":1…' to be '{"a":1…'` |
| `writeJsonObject`: stop awaiting the sink close | ⚪️ **survived — the round-2 entry claiming an `ENOENT` kill does not reproduce.** Deleting `await closed` left all in-range tests green, because `pipeline` already waited for `'close'`. The line has been removed rather than re-justified; see round 3. |
| `scanPagesById`: never advance the cursor | ✅ killed — `expected [ undefined, undefined, undefined ] to deeply equal [ undefined, 500, 1000 ]`, plus a 60 s timeout on the infinite re-read |
| `scanPagesById`: stop after the first page | ✅ killed — `to have a length of 1001 but got 500` |
| `scanPagesById`: drop the page size (unbounded `findMany`) | ✅ killed — same |
| `archiveImages`: `break` after the first page | ✅ killed — `to have a length of 503 but got 500` |
| `zipEntryNameForUrl`: remove the empty-basename fallback | ✅ killed — `expected '' to be 'unnamed-4-blob'` (+ the seam test) |
| `zipEntryNameForUrl`: always use the fallback | ✅ killed — `expected 'unnamed-3-blob-7.jpeg' to be 'blob-7.jpeg'` |
| `process-csam`: read `e.message` directly again | ✅ killed — `promise rejected "TypeError: …" instead of resolving` |
| `errorMessage()` returns a constant | ✅ killed — `expected 'error' to be 'synthetic archive failure'` |
| `uploadStream`: bare `reject()` again | ✅ killed — `expected undefined to be an instance of Error` |
| `uploadStream`: drop the `cause` | ✅ killed — `expected undefined to be Error: synthetic s3 failure` |
| Test harness: remove the tmpdir redirect | ✅ killed — the suite's own scratch-dir guard |
| `append`'s `while (pending >= max)` re-check → `if` | ✅ killed — `expected 17 to be less than or equal to 8` (seam) and `expected 14 to be less than or equal to 5` (helper) |
| `releaseWaiters`' `while` → `if` | ✅ killed **(round 3)** — `expected { raced: 'HUNG', settled: 3 } to deeply equal { raced: 'all-settled', settled: 5 }`. Round 2 recorded this as a harmless survivor; it breaks the helper's settle-always contract and had no test. Severity is scoped in round 3 below. |
| Remove the `modelScanComplete` ordering assertion | ⚪️ **survived, and correctly so** — see below |

Two survivors in the first sweep were real gaps and both were fixed:

1. **`archiveImages` page-walking had no test at all.** The bundle's page-walk test uses an `ExternalLink` report, which never reaches `archiveImages`. A test crossing a full page boundary (503 images against a 500-row page) was added; the `break`-after-first-page mutant is now killed.
2. **A test was asserting on a live-mutating array.** `modelVersions` was filtered with `modelId: { in: modelIds }`, passing the array by reference. The recorded mock call argument therefore kept mutating *after* the call, so a scan that genuinely ran with an **empty** id set read back as complete. The production code now passes a copy (`[...modelIds]`).

   ⚠️ **Correcting round 2.** The comment round 2 wrote beside that copy justified it as a *race* — `modelIds` "is still being appended to by the generator above at the moment this closure is built". That is false, and the ordering invariant asserted at the top of `streamModelVersions` is what makes it false: `streamModelVersions` throws unless `modelScanComplete`, so every `findMany` is issued strictly after the last push. Two comments in one function asserting contradictory concurrency models. The copy is kept, because it does have one effect and that effect is now what the comment claims — it is what lets the *COMPLETE set* assertion observe the hazard. Measured as a two-arm control, with the invariant removed and the entry order reversed so the hazard is reachable: against the copy the test fails `expected [] to have a length of 501`; against a live reference it passes vacuously.

On the `modelScanComplete` assertion: removing it alone is unobservable, because in correct code it never fires — it is an **invariant guard, not regression coverage**, and it is labelled as one. Its reachability was proved separately by mutating the hazard it names (serialising `modelVersions` before `models`), which fails with the guard's own message:

```
Error: archiveBaseReportData: model versions were streamed before the model scan finished —
       modelIds is incomplete and the bundle would be missing evidence
```

and with the guard *also* removed the same hazard is still caught, by the byte-identity test and the model-id test — so the hazard has coverage that does not depend on the assertion. The assertion's job is to turn a silently-incomplete evidence bundle into a loud, self-describing failure.

### Gates

- `pnpm typecheck` — `OK — 0 type errors in 119s (heap cap 8192 MB)`
- `pnpm run test:unit:run` (full suite) — `Test Files 4 failed | 1677 passed | 3 skipped (1684)`, `Tests 28 failed | 38773 passed | 33 skipped (38834)`. **All 28 failures are pre-existing local-environment noise outside this diff's dependency closure**, verified three ways: none of the four files is in this branch's diff or its working tree, none imports anything this PR touches, and 26 of the 28 fail with `TypeError: Cannot read properties of undefined (reading 'setItem')` — `window.localStorage` is undefined on this box's Node build (`ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`). The other two are `eventloop-watchdog.capture.test.ts`, a known local-only failure that is green in CI.
- `eslint` — 8 files linted; **0 errors and 0 warnings in every new file**. `csam.service-new.ts` reports 1 error + 10 warnings, all pre-existing and verified present at base (the error is the `import('stream/web')` type annotation in `getTrainingDataZipStream`, untouched by this diff).
- `prettier --check` clean (validated with a deliberately-unformatted control file, which it correctly flags).
- **CI on this head (`05f6369`): 19 checks registered, 18 `pass`, 1 `skipping`, 0 failures** — including all four `Unit tests` shards, `App unit tests + typecheck`, `ESLint + Prettier (changed files)`, `tekton / typecheck` and the four `preview / *` jobs. That also settles the four local failures above: the same shards are green here, so they are this box's environment, not the tree.

---

## Round 3 — a delta audit of round 2

Round 2 shipped four claims about its own code that do not survive re-checking. The code was
mostly right; the prose describing it was not, and in one case the prose actively recommended
breaking working code. Everything below was re-measured, not reasoned about.

**1. `releaseWaiters`' `while` is what makes the failure path settle, and round 2 said it was not.**

`releaseWaiters` runs on `entry` and on `error` only, and once a failure is latched `append()`
rejects immediately, so no new entries are produced. If the error is the last event the archive
emits, it is therefore also the last wake-up any parked waiter gets, and releasing one per event
strands the rest forever — breaking this helper's contract that an `append()` always settles.
Measured against the module with a stub emitting a single `error` and no `entry`: 2 of 5 appends
never settle. Round 2's comment explained the loop purely in terms of the success path, said "the
bound is upheld anyway, by the `while` re-check in `append()`", and its mutation table recorded
the `while`→`if` mutant as a harmless survivor — together, an invitation to make the edit. All
three surfaces are corrected and a regression test pins it; it survived only because nothing
tested the failure path.

> ⚠️ **Scoping this honestly, because the first draft of this section overstated it.** The
> paragraph above is a claim about *this module's contract*, not a reproduced production hang, and
> an earlier draft asserted both "archiver emits `error` at most once" and "`archiveCsamDataForReport`
> never settles for that report". **Neither is established.** `archiver@6` has **23 distinct
> `emit('error')` sites** and guarantees nothing in either direction, and three of its real failure
> paths were probed without reaching the stranding shape: an empty entry name emits `error` and
> then goes on emitting **8 more** `entry` events (so an `if` would drain there); a directory entry
> emits no `error` at all; an erroring source stream stalls the archiver with **no `error` event**,
> which no variant of this loop can rescue. The `while` stays because it is free and is what keeps
> the contract true whichever site fires — not because a wedge has been observed.

**2. `writeJsonObject`'s second `await` on the sink's `'close'` is removed.**

Its doc-block asserted that `pipeline` resolving is no promise a later `fs.createReadStream` sees
every byte. Measured false on **node v24.19.0** (the pinned dev shell) and again on a **node 26**
host: `pipeline` ends the sink and waits on it via `stream.finished`, which for an `autoDestroy`
Writable — the default, and what `fs.createWriteStream` is — means waiting for `'close'`. Over 20
rounds of a 4 MB document, `'close'` had already fired and the file was byte-complete every time
`pipeline` resolved. The mutation-table row claiming an `ENOENT` kill does not reproduce: deleting
the line left every in-range test green.

It was also **not** harmless belt-and-braces. This signature accepts any `Writable`, and one built
with `autoDestroy: false` never emits `'close'` at all, so the extra await hung forever. Removed,
with a test that reports `HUNG` if it is put back.

> The equivalent wait in `archive-helpers.ts` is **genuinely load-bearing and is left alone** —
> there the awaited promise is archiver's `finalize()`, which resolves when the zip module ends
> and knows nothing about the sink (measured in round 1: 24254 bytes vs 25342). Its doc now says
> so, and records the same `autoDestroy` requirement.

**3. The `[...modelIds]` copy is not defending against a race.**

Round 2's comment said `modelIds` "is still being appended to by the generator above at the moment
this closure is built". The ordering invariant asserted at the top of `streamModelVersions` throws unless
`modelScanComplete`, so every `findMany` is issued strictly after the last push — two comments in
one function asserting contradictory concurrency models. Reverting the copy to a live reference
leaves all 59 round-2 tests green.

The copy is kept because it does have exactly one measured effect, and that is now what the comment
claims: it is what lets the *COMPLETE set* assertion observe the hazard the invariant names.
Two-arm control, with the invariant removed and the entry order reversed so the hazard is
reachable — copy: `expected [] to have a length of 501`; live reference: passes vacuously.

**4. `serializeJsonObject`'s comma bookkeeping for an `items` entry was unpinned.**

Removing `wroteEntry = true;` from the `items` branch left everything green. Production is
unaffected today because two `value` entries precede the arrays, but the helper is exported with a
byte-identity contract and `[['xs',{items:[1]}],['ys',{value:2}]]` yields `{"xs":[1]"ys":2}` —
invalid JSON, no test red. Pinned.

**5. Smaller corrections.**

- `archive-helpers.ts` referred to a "file header" that does not exist. The reason is stated inline.
- The in-source 🔴 "bundle is unchanged" block overstated on the two axes the description above now
  states precisely: the queries do carry `orderBy`/`take`/`id > cursor`, and element order is
  `id ASC` where it was plan-order, so the document is set-identical but not byte-identical.
- `scanPagesById` now names its index precondition rather than leaving a bare "offset paging
  degrades quadratically" rationale. Read off `packages/civitai-db-schema/prisma/schema.full.prisma`:
  `Image` has `@@index([userId, id])`, so the one genuinely large scan is served directly; `Model`
  has no `userId` index and `ModelVersion`'s only `modelId` index is a **hash** index, which cannot
  serve a range or an ordering. Accepted, because those two row counts are small next to the image
  count. No `EXPLAIN` was run and the comment says so.

**6. Resource leak closed on both triggers.**

`archiveImages` now pulls DB pages *inside* the archiver's lifetime, where the pre-change code
fetched every row before any archiver existed. A `dbRead` failure on page 2+ therefore escapes
without `finalize()`, leaving the write-stream fd and queued buffers open while the caller's
handler removes the directory underneath them. The leak shape was pre-existing (a
`fetchBlob`/`append` failure did it already); round 2 added the DB trigger. Both call sites now
release through a shared `closeArchiveOnFailure`, each with its own regression test.

> ⚠️ Recorded honestly in that helper rather than dressed up: only `output.destroy()` is observed
> by the tests. Removing `archive.abort()` leaves **all 64 tests green**, because the bounded
> appender caps the backlog at `MAX_PENDING_ARCHIVE_ENTRIES` — too shallow to observe without a
> timing-dependent assertion. Its effect was measured separately on an *unbounded* queue, where it
> is unmissable: sink destroyed and no `abort()` → the archiver went on to deflate **60 of 60**
> queued entries; with `abort()` → **0**. Real work, bounded here, and marked untested rather than
> covered.

### Round 3 — red at pre-fix / green at HEAD

Each new test was run against the pre-fix code and watched to fail. Actual output:

| Test | Pre-fix mutation | Red at pre-fix | Green at HEAD |
|---|---|---|---|
| `settles EVERY parked append when the archiver errors` | `releaseWaiters`' `while` → `if` | `expected { raced: 'HUNG', settled: 3 } to deeply equal { raced: 'all-settled', settled: 5 }` | ✅ |
| `resolves for a sink that never emits close` | re-add `await closed` | `expected { raced: 'HUNG', closed: false } to deeply equal { raced: 'resolved', closed: false }` | ✅ |
| `separates an items entry from whatever follows it` | drop `wroteEntry = true` in the `items` branch | `expected '{"xs":[1]"ys":2}' to be '{"xs":[1],"ys":2}'` | ✅ |
| `closes the archive … when a page fetch fails mid-archive` | drop the release in `archiveImages` | `expected false to be true` on `zipSinks[0].destroyed` | ✅ |
| `closes the archive … when a download fails` | drop the release in `archiveGeneratedImages` | `expected false to be true` on `zipSinks[0].destroyed` | ✅ |

Each release-site mutant failed **only its own** test, so neither is killed by the other's error.
Splitting `closeArchiveOnFailure`: dropping `output.destroy()` kills both cases; dropping
`archive.abort()` kills nothing — stated as such above rather than claimed as coverage.

Also fixed: a timing dependency in `json-stream-helpers.test.ts`. A sink whose `open()` was still
in flight could create its file under the suite's tmpdir after `afterAll` had begun removing it,
seen once as an `ENOTEMPTY` failure of the *suite* (not of any test) under parallel load. The test
now waits for that sink to close. Seen exactly once and not reproduced in 9 subsequent group
runs (4 before the fix, 5 after) — so the fix removes a real timing dependency, but is not proven
to be the cause of that one occurrence.

### Round 3 gates

Run inside the pinned dev shell (`direnv exec . …`, node **v24.19.0**), which is why the 28
`window.localStorage` failures reported for round 2 are absent — they were the host node, not the
tree.

- `pnpm typecheck` — `OK — 0 type errors in 129s (heap cap 8192 MB)`
- `pnpm run test:unit:run` (full suite) — `Test Files 1681 passed | 3 skipped (1684)`,
  `Tests 38806 passed | 33 skipped (38839)`, **0 failed**
- In-range four files — `64 passed (64)`, run 5× clean after the tmpdir fix (and 4× clean before
  it; the `ENOTEMPTY` was seen once and never reproduced)
- `prettier --check` — clean on all six changed files, and validated against a deliberately
  unformatted control file that it correctly flags
- `eslint` — **0 errors and 0 warnings in every changed helper and test file**;
  `csam.service-new.ts` reports the same 1 error + 10 warnings as round 2, all pre-existing (the
  error is the `import('stream/web')` annotation at a line this diff does not touch)

## Deliberately out of scope

- `src/server/services/csam.service.ts` (the pre-`-new` module) carries the same unbounded patterns in its own copy of these functions. Its `archiveCsamDataForReport` is not what the cron job calls, so it is not the live path — flagging rather than fixing, since the right move is probably deleting the dead half.
- **Per-report failure isolation.** Ordering the batch does not provide it (see change 6), and the fix in change 8 only covers a report that *throws* — a report that kills the process still blocks the queue. A circuit breaker (skip after N consecutive failures) would fix it, but skipping a report means not archiving evidence, which is a policy decision rather than an engineering one.
- **Narrowing the bundle's `select`.** It would cut the row width substantially, but it changes what evidence is in a NCMEC archive. Not done here, and not something to do quietly.

### DIFF
--- a/src/server/jobs/__tests__/process-csam.test.ts
+++ b/src/server/jobs/__tests__/process-csam.test.ts
@@ -1,0 +1,125 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * Batch integrity for the hourly CSAM jobs.
+ *
+ * The defect: both loops caught per report so that one bad report could not stop the rest, then
+ * read `e.message` inside the catch block. A rejection whose value is not an object — which is
+ * exactly what `uploadStream`'s bare `reject()` produced — makes that read throw a `TypeError`
+ * *inside the catch*, where nothing catches it. It escapes the `for` loop and abandons every
+ * remaining report in that batch, silently, having logged nothing.
+ *
+ * All ids and messages below are invented.
+ */
+
+const mocks = vi.hoisted(() => ({
+  getCsamsToArchive: vi.fn(),
+  getCsamsToReport: vi.fn(),
+  archiveCsamDataForReport: vi.fn(),
+  processCsamReport: vi.fn(),
+  getCsamsToRemoveContent: vi.fn(async () => []),
+}));
+
+vi.mock('~/server/services/csam.service-new', () => mocks);
+
+import { csamJobs } from '~/server/jobs/process-csam';
+import { loggingMock } from '~/__tests__/mocks/logging.mock';
+
+const archiveJob = csamJobs.find((job) => job.name === 'archive-csam-reports');
+const sendJob = csamJobs.find((job) => job.name === 'send-csam-reports');
+
+const reports = [{ id: 8801 }, { id: 8802 }, { id: 8803 }];
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mocks.getCsamsToArchive.mockResolvedValue(reports);
+  mocks.getCsamsToReport.mockResolvedValue(reports);
+  mocks.archiveCsamDataForReport.mockResolvedValue(undefined);
+  mocks.processCsamReport.mockResolvedValue(undefined);
+});
+
+/** The exact pre-fix failure: `Promise.reject()` with no argument. */
+const rejectWithUndefined = () => Promise.reject();
+
+describe('archive-csam-reports batch isolation', () => {
+  it('POSITIVE CONTROL: a healthy batch reaches every report', async () => {
+    await archiveJob!.run({}).result;
+    expect(mocks.archiveCsamDataForReport).toHaveBeenCalledTimes(reports.length);
+  });
+
+  it('continues the batch when a report rejects with a non-Error value', async () => {
+    mocks.archiveCsamDataForReport.mockImplementationOnce(rejectWithUndefined);
+
+    // Pre-fix: `e.message` throws `TypeError: Cannot read properties of undefined (reading
+    // 'message')` from inside the catch, the job promise rejects, and reports 8802 and 8803 are
+    // never attempted.
+    await expect(archiveJob!.run({}).result).resolves.not.toThrow();
+
+    expect(mocks.archiveCsamDataForReport).toHaveBeenCalledTimes(reports.length);
+    expect(mocks.archiveCsamDataForReport).toHaveBeenLastCalledWith(reports[2]);
+  });
+
+  it('still logs something identifiable for the failed report', async () => {
+    mocks.archiveCsamDataForReport.mockImplementationOnce(rejectWithUndefined);
+
+    await archiveJob!.run({}).result;
+
+    // A message of `undefined` would make the log line useless, which is the other half of the
+    // fix — extracting safely is not the same as extracting nothing.
+    expect(loggingMock.logToAxiom).toHaveBeenCalledWith(
+      expect.objectContaining({ name: 'csam-report', subType: 'archive-data' })
+    );
+    const logged = loggingMock.logToAxiom.mock.calls
+      .map(([arg]) => arg as { subType?: string; message?: unknown })
+      .filter((arg) => arg?.subType === 'archive-data');
+    expect(logged).toHaveLength(1);
+    expect(typeof logged[0].message).toBe('string');
+    expect(logged[0].message).toBe('undefined');
+  });
+
+  it.each([
+    ['a string', 'plain string failure', 'plain string failure'],
+    ['null', null, 'null'],
+    ['a plain object', { code: 'ENOENT' }, '{"code":"ENOENT"}'],
+  ])('survives a rejection that is %s', async (_label, value, expected) => {
+    mocks.archiveCsamDataForReport.mockImplementationOnce(() => Promise.reject(value));
+
+    await expect(archiveJob!.run({}).result).resolves.not.toThrow();
+
+    expect(mocks.archiveCsamDataForReport).toHaveBeenCalledTimes(reports.length);
+    const logged = loggingMock.logToAxiom.mock.calls
+      .map(([arg]) => arg as { subType?: string; message?: unknown })
+      .filter((arg) => arg?.subType === 'archive-data');
+    expect(logged[0].message).toBe(expected);
+  });
+
+  it('preserves the message of a real Error', async () => {
+    mocks.archiveCsamDataForReport.mockImplementationOnce(() =>
+      Promise.reject(new Error('synthetic archive failure'))
+    );
+
+    await archiveJob!.run({}).result;
+
+    const logged = loggingMock.logToAxiom.mock.calls
+      .map(([arg]) => arg as { subType?: string; message?: unknown })
+      .filter((arg) => arg?.subType === 'archive-data');
+    expect(logged[0].message).toBe('synthetic archive failure');
+  });
+});
+
+describe('send-csam-reports batch isolation', () => {
+  it('continues the batch when a report rejects with a non-Error value', async () => {
+    // The identical shape in the sibling loop. Fixing one and not the other would leave the
+    // batch-abort live on the path that actually files the NCMEC report.
+    mocks.processCsamReport.mockImplementationOnce(rejectWithUndefined);
+
+    await expect(sendJob!.run({}).result).resolves.not.toThrow();
+
+    expect(mocks.processCsamReport).toHaveBeenCalledTimes(reports.length);
+    const logged = loggingMock.logToAxiom.mock.calls
+      .map(([arg]) => arg as { subType?: string; message?: unknown })
+      .filter((arg) => arg?.subType === 'send-report');
+    expect(logged).toHaveLength(1);
+    expect(logged[0].message).toBe('undefined');
+  });
+});
--- a/src/server/jobs/process-csam.ts
+++ b/src/server/jobs/process-csam.ts
@@ -9,6 +9,26 @@ import {
 } from '~/server/services/csam.service-new';
 import { logToAxiom } from '~/server/logging/client';
 
+/**
+ * Reads a message off an unknown thrown value without assuming it has one.
+ *
+ * 🔴 This is a batch-integrity guard, not a formatting nicety. The loops below catch per report
+ * so that one bad report cannot stop the rest, but `e.message` on a non-object rejection throws
+ * a `TypeError` *inside the catch block*, where nothing catches it — it escapes the `for` loop
+ * and abandons every remaining report in the batch. `uploadStream` used to reject with a bare
+ * `reject()`, i.e. `undefined`, which is exactly that case. Both ends are fixed; this is the end
+ * that holds for any future thrower.
+ */
+function errorMessage(e: unknown): string {
+  if (e instanceof Error) return e.message;
+  if (typeof e === 'string') return e;
+  try {
+    return JSON.stringify(e) ?? String(e);
+  } catch {
+    return String(e);
+  }
+}
+
 const sendCsamReportsJob = createJob(
   'send-csam-reports',
   '0 */1 * * *',
@@ -18,13 +38,13 @@ const sendCsamReportsJob = createJob(
     for (const report of reports) {
       try {
         await processCsamReport(report);
-      } catch (e: any) {
+      } catch (e) {
         if (isDev) console.log(e);
         logToAxiom({
           name: 'csam-report',
           type: 'error',
           subType: 'send-report',
-          message: e.message,
+          message: errorMessage(e),
         });
       }
     }
@@ -41,12 +61,12 @@ const archiveCsamReportDataJob = createJob(
     for (const report of reports) {
       try {
         await archiveCsamDataForReport(report);
-      } catch (e: any) {
+      } catch (e) {
         logToAxiom({
           name: 'csam-report',
           type: 'error',
           subType: 'archive-data',
-          message: e.message,
+          message: errorMessage(e),
         });
       }
     }
--- a/src/server/services/__tests__/csam-archive-backpressure.test.ts
+++ b/src/server/services/__tests__/csam-archive-backpressure.test.ts
@@ -1,0 +1,338 @@
+import fs from 'fs';
+import path from 'path';
+import { afterAll, afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { Archiver } from 'archiver';
+import type * as ArchiverModule from 'archiver';
+
+/**
+ * Seam coverage for the CSAM evidence archiver.
+ *
+ * `archive-helpers.test.ts` pins the mechanism in isolation. This file pins the thing that
+ * isolation cannot see: that `archiveCsamDataForReport` actually routes its appends through the
+ * bounded appender. A perfectly-correct helper nobody calls fixes nothing.
+ *
+ * The defect being guarded: every image belonging to the reported user was downloaded, fully
+ * buffered, and handed to `archive.append()` with nothing bounding the archiver's queue, so
+ * resident memory tracked total downloaded bytes and the process was killed on its memory limit
+ * partway through. All numbers below are invented.
+ */
+
+/** Instrumentation over the real archiver: counts appended vs. processed entries per archive. */
+type ArchiveProbe = {
+  appended: number;
+  processed: number;
+  peakOutstanding: number;
+  peakRetainedBytes: number;
+  entryNames: string[];
+  options: { zlib?: { level?: number } } | undefined;
+};
+const probes: ArchiveProbe[] = [];
+
+vi.mock('archiver', async (importOriginal) => {
+  const actual = await importOriginal<{ default: typeof ArchiverModule }>();
+  const create = (actual.default ?? actual) as unknown as (
+    format: string,
+    options?: unknown
+  ) => Archiver;
+  return {
+    default: (format: string, options?: unknown) => {
+      const archive = create(format, options);
+      const probe: ArchiveProbe = {
+        appended: 0,
+        processed: 0,
+        peakOutstanding: 0,
+        peakRetainedBytes: 0,
+        entryNames: [],
+        options: options as ArchiveProbe['options'],
+      };
+      probes.push(probe);
+      let retained = 0;
+      const sizes: number[] = [];
+      const originalAppend = archive.append.bind(archive);
+      archive.append = ((source: Buffer, data: { name: string }) => {
+        probe.appended++;
+        probe.entryNames.push(data.name);
+        sizes.push(Buffer.isBuffer(source) ? source.byteLength : 0);
+        retained += Buffer.isBuffer(source) ? source.byteLength : 0;
+        const outstanding = probe.appended - probe.processed;
+        if (outstanding > probe.peakOutstanding) probe.peakOutstanding = outstanding;
+        if (retained > probe.peakRetainedBytes) probe.peakRetainedBytes = retained;
+        return originalAppend(source, data as never);
+      }) as typeof archive.append;
+      archive.on('entry', () => {
+        retained -= sizes[probe.processed] ?? 0;
+        probe.processed++;
+      });
+      return archive;
+    },
+  };
+});
+
+vi.mock('~/utils/file-utils', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  fetchBlob: (...args: unknown[]) => mockFetchBlob(...args),
+}));
+
+vi.mock('@aws-sdk/lib-storage', () => ({
+  Upload: class {
+    private readonly body: NodeJS.ReadableStream;
+    constructor({ params }: { params: { Body: NodeJS.ReadableStream } }) {
+      this.body = params.Body;
+    }
+    async done() {
+      // Drain, so the PassThrough the service pipes into does not stall the caller.
+      for await (const _chunk of this.body) {
+        void _chunk;
+      }
+      return {};
+    }
+  },
+}));
+
+vi.mock('@aws-sdk/client-s3', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  S3Client: class {
+    send() {
+      return Promise.resolve({});
+    }
+    destroy() {
+      return undefined;
+    }
+  },
+}));
+
+vi.mock('~/server/http/orchestrator/flagged-consumers', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  getConsumerStrikes: (...args: unknown[]) => mockGetConsumerStrikes(...args),
+}));
+
+let mockFetchBlob: (...args: unknown[]) => unknown = () => null;
+let mockGetConsumerStrikes: (...args: unknown[]) => unknown = async () => [];
+
+import { archiveCsamDataForReport, getCsamsToArchive } from '~/server/services/csam.service-new';
+import { dbMock } from '~/__tests__/mocks/db.mock';
+import { setEnv } from '~/__tests__/mocks/env.mock';
+import {
+  MAX_PENDING_ARCHIVE_ENTRIES,
+  MEDIA_ARCHIVE_COMPRESSION_LEVEL,
+} from '~/server/utils/archive-helpers';
+
+// Overshoot the bound by a wide, non-multiple margin so the fixture cannot sit on the boundary
+// it is testing. Invented figures — the real report was a different size entirely.
+const IMAGE_COUNT = 53;
+const IMAGE_BYTES = 97 * 1024;
+
+const csamBaseDir = path.join(process.cwd(), 'csam');
+
+function fakeBlob(bytes: number) {
+  return {
+    type: 'image/jpeg',
+    arrayBuffer: async () => {
+      const buffer = Buffer.alloc(bytes, 0xab);
+      return buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength);
+    },
+  };
+}
+
+beforeEach(() => {
+  probes.length = 0;
+  mockFetchBlob = async () => fakeBlob(IMAGE_BYTES);
+  mockGetConsumerStrikes = async () => [];
+  setEnv({
+    CSAM_UPLOAD_KEY: 'test-key',
+    CSAM_UPLOAD_SECRET: 'test-secret',
+    CSAM_UPLOAD_REGION: 'test-region',
+    CSAM_UPLOAD_ENDPOINT: 'https://example.invalid',
+    CSAM_BUCKET_NAME: 'test-bucket',
+  });
+});
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+afterAll(() => {
+  fs.rmSync(csamBaseDir, { recursive: true, force: true });
+});
+
+describe('getCsamsToArchive', () => {
+  it('orders the batch deterministically so one bad report cannot permanently starve the rest', async () => {
+    await getCsamsToArchive();
+    const args = dbMock.dbRead.csamReport.findMany.mock.calls.at(-1)?.[0];
+    expect(args?.orderBy).toBeDefined();
+    expect(args?.orderBy).toEqual([{ createdAt: 'asc' }, { id: 'asc' }]);
+  });
+});
+
+describe('archiveCsamDataForReport (Image report)', () => {
+  const report = {
+    id: 4242,
+    userId: 909090,
+    reportedById: 5,
+    reportId: 77,
+    type: 'Image',
+    details: {},
+    images: [],
+  };
+
+  function seedDb() {
+    dbMock.dbRead.image.findMany.mockResolvedValue(
+      Array.from({ length: IMAGE_COUNT }, (_, i) => ({
+        id: i + 1,
+        url: `image-uuid-${i}`,
+        name: `picture-${i}.jpeg`,
+        type: 'image',
+        width: 1024,
+        userId: report.userId,
+      }))
+    );
+    dbMock.dbRead.user.findUnique.mockResolvedValue({
+      id: report.userId,
+      name: 'n',
+      email: 'e',
+      username: 'u',
+    });
+    dbMock.dbRead.model.findMany.mockResolvedValue([]);
+    dbMock.dbRead.modelVersion.findMany.mockResolvedValue([]);
+    dbMock.dbWrite.csamReport.update.mockResolvedValue({});
+  }
+
+  it('holds a bounded number of image buffers no matter how many images the user has', async () => {
+    seedDb();
+
+    await archiveCsamDataForReport(report as never);
+
+    const imageArchive = probes.find((p) => p.appended > 0);
+    expect(
+      imageArchive,
+      'no archive received any entries — the probe is wired to nothing'
+    ).toBeDefined();
+
+    // Positive control on the instrument, and it must come FIRST: a bounded reading is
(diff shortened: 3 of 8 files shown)


ZacxDev:
Adversarial audit round 1 ran against `e552573204`; its findings were fixed in `05f6369c8c`. Recording what that fix round **claims** to have addressed, so the next round can check each against the diff rather than take it on trust.

```audit-claims round=1 audited=e552573204d6b4539fc4f9a537d298814423d061..05f6369c
1. Unbounded base-bundle serialisation: the single unpaged image/model/modelVersion findMany plus one JSON.stringify is replaced by keyset-paged scans (id > cursor ORDER BY id ASC) streamed via writeJsonObject, so no single JS string is built and no caller holds more than one page.
2. Bundle contents are claimed UNCHANGED by that rewrite: same where clauses, no select added, same bigint replacer, same five properties in the same order; asserted by byte-identity against the previous serialiser both at unit level and on the uploaded data.json.
3. Batch abort: process-csam.ts read e.message unguarded, so a bare reject() (undefined) threw a TypeError inside the catch and abandoned the rest of the batch. Claimed fixed at both ends — safe message extraction, and the uploadStream rejection now carries a real Error.
4. The getCsamsToArchive ORDER BY comment claimed to cure head-of-line blocking; corrected to state it makes the blocking deterministic and that per-report failure isolation is deliberately out of scope.
5. The error-latching behaviour change (one bad entry now fails the whole report instead of being silently dropped) is kept deliberately, the PR body's "outcome-neutral for archivedAt" claim is corrected, and an empty-basename previewUrl is given a fallback entry name.
6. archive-helpers: the releaseWaiters comment is corrected (it releases every waiter whenever a slot is free; correctness rests on the while re-check in append), and the documented peak is restated as ~38 image-sizes rather than 8.
7. Test scaffolding no longer writes zips into a non-gitignored process.cwd()/csam that afterAll rm -rf'd; it uses mkdtemp under os.tmpdir().
```

Two judgement calls in that round were flagged rather than decided silently, and are the parts most worth a second opinion:

- **`archiveBaseReportData` and `archiveImages` now run two independent paged scans** where they previously shared one in-memory snapshot. Accepted on the grounds that the archive step already tolerates `data.json` listing an image the zip lacks (`if (!blob) return`), and that the pre-existing snapshot→download window was already larger. Documented at the `scanUserImages` definition.
- **Row order in the bundle is now deterministic (`id ASC`)** where it was previously whatever the query plan produced. The row set and every column are unchanged, so this was not treated as a content change, but the bytes of a given historical run would differ.

ZacxDev:
Adversarial audit round 2 ran against `05f6369c8c` and returned eight findings; they were addressed across `93886e8402`, `659f14d2b8` and `818ba685ad`. Recording what that fix round **claims** to have addressed, so the next round can check each against the diff rather than take it on trust.

```audit-claims round=2 audited=05f6369c8c282cc6a95b69fd9cc78c482c57e12d..818ba685
1. releaseWaiters keeps its while loop and gains a regression test pinning that every parked append settles when the archiver errors; the comment and PR body no longer claim the loop is a harmless survivor. The SEVERITY claim is explicitly retracted — archiver has many error-emit sites and no one-error guarantee, and none of three probed real failure paths reaches the stranding shape, so the guard is justified only as keeping the module's settle-always contract true, not by any observed wedge.
2. The await on sink close in json-stream-helpers is DELETED rather than kept: measured on two Node versions, close had already fired and the file was byte-complete when pipeline resolved, and for autoDestroy:false close never fires at all so the await would hang forever. The equivalent wait in archive-helpers is load-bearing and is kept, now with a caveat distinguishing the two.
3. The false race comment beside the model-id copy is rewritten. A two-arm control (invariant removed, entries reversed) shows the copy is not purely defensive, so the previous round's claim about it stands; only the comment was wrong.
4. serializeJsonObject's comma bookkeeping for a leading items entry is pinned by a test; previously an items-first object followed by another entry emitted syntactically invalid JSON with no test red.
5. The dead "see the file header" cross-reference in archive-helpers is replaced with the reason stated where the reader is.
6. The in-source block asserting the bundle is unchanged is corrected on both axes: the queries now carry orderBy, take and an id-cursor predicate, and array element order is now id ASC rather than plan order — stated for models and modelVersions, not only images.
7. The archiver and its sink are now released when a DB page fetch fails and when a download fails, closing a leak whose trigger the previous round widened; both paths have tests, and each release-site mutant fails only its own test.
8. The keyset-paging performance rationale now names its index precondition, read from the Prisma schema rather than from a query plan.
```

Two things this round deliberately left unproven rather than papering over, both recorded in-source:

- `archive.abort()` has a measured effect on an unbounded-queue probe, but no test in the suite can observe it at the bounded queue depth — so it is recorded as untested rather than counted as covered.
- A single `ENOTEMPTY` in a test `afterAll` was seen once and never reproduced (4 runs before, 5 after). A real dangling-`open()` timing dependency near it was removed, but that is **not** claimed as the cause.

ZacxDev:
## Audit ladder closed — one finding deliberately left open

Three adversarial rounds ran against this PR (round 1 full, rounds 2 and 3 delta). The final round returned **no deploy-blocking findings** and an advisory verdict of *safe to merge*, having re-measured rather than trusted the previous rounds' tables.

`a2f1e6f3d7` fixes the two prose findings from round 3. **One finding is not fixed, and this comment is the record of it** — so it reads as open rather than absent.

### OPEN — `closeArchiveOnFailure` can leave a parked `append()` unsettled

`src/server/services/csam.service-new.ts` — `closeArchiveOnFailure` calls `archive.abort()` (which emits no `error`, so `createBoundedArchive` never latches `failure`) and `output.destroy()` (which stops further `entry` events). An `append()` parked against the pending-entry bound at that moment therefore has no remaining wake-up source and never settles.

That contradicts the contract this module now documents in `archive-helpers.ts` — *an `append()` either resolves or rejects* — and the guard whose whole purpose is to uphold it.

**Why it is not fixed here.** It is roughly one line (latch the failure, or emit it onto `archive`, before `abort()`). But nobody has reproduced it: the audit measured it with a specific harness, and two independent operator probes against the real archiver failed to create the precondition at all — in the instrumented one the bound was reached (`peak: 2`) yet every append had already settled by the time the failure path fired (`parkedAtCloseTime: 0`), making the result vacuous rather than a refutation. Changing code in response to a review resets the verification gate, and that is not worth spending on a defect that may be unreachable.

**Impact, stated honestly:** no unhandled rejection (`Promise.all` has already attached handlers to every element), no caller hang (the enclosing `catch` rethrows immediately), no evidence loss, and the object graph becomes unreachable so retained buffers are collected. It is a latent contract violation, not an observed failure.

**Closing condition.** This item is closed when **either**:
1. a test reproduces a parked `append()` that never settles after `closeArchiveOnFailure`, and a fix lands making it settle — evidenced by that test failing before the fix and passing after; **or**
2. a maintainer establishes that the case is unreachable in this code path and records why, at which point the contract comment in `archive-helpers.ts` should be narrowed to match and this item deleted.

Either outcome resolves it. What should *not* happen is it sitting here indefinitely with neither.

### Also recorded, unrelated to this PR

`csam.service-new.ts` on `main` carries a CSAM report id in a code comment (introduced by an earlier, unrelated commit). This is a public repository. Worth a separate one-line cleanup PR; deliberately not folded into this diff.

### Two things this PR does not claim

- The `releaseWaiters` loop is justified as cheap insurance keeping the module's settle-always contract true — **not** by any observed production hang. An earlier round asserted a severity that was retracted once `archiver`'s error semantics were actually measured.
- `archive.abort()` has a measured effect on an unbounded queue but no test in this suite can observe it at the bounded depth; it is recorded in-source as untested rather than counted as covered.

PR #4692, opened 2026-09-08
fix(images): observe whether a created Image row's media actually exists, and stop the upload hook lying about a refused PUT

ZacxDev:
> **Corrected after review (see the PR comment for the full list).** Three claims in the
> first version of this body were wrong or under-stated and have been fixed here and in the
> code: the server probe does **not** cover every `Image` row; a non-zero `present` count is
> **not** a sufficient positive control on its own (the missing half has now been measured);
> and the consumer count was **24**, derived from a literal grep that could not see aliased
> calls — it is re-derived below.

## The defect

An `Image` row is written from client-supplied JSON. The media key (`Image.url`) arrives over the wire and is never checked against storage, so a key whose object never landed produces a complete, healthy-looking row whose media 404s forever. There is no repair path — nothing later re-derives the bytes.

The dominant producer is `useCFImageUpload`. **XHR's `error` event is network-only**, so a PUT that reaches the store and is *refused* — an expired presign, a malformed request, a briefly unavailable store — completes normally and lands on `loadend`, where the hook resolves and hands the caller a key with nothing behind it. In a sampled production window, 7 of 10 defective rows had a key an upload endpoint had signed 2.0–23.3 s before the row was written, with zero bytes ever stored.

## The shape I chose, and why

**Server half — an observe-only existence check in `createImage`.** Nothing branches on the verdict: rejecting an image creation is a *new* user-facing failure mode, and the honest way to size it is to let the `absent` rate accumulate against a real denominator first. There is deliberately **no env flag** either — an enforcement branch nobody exercises is an untested one, so enforcement is a separate change gated on this measurement.

### 🔴 What that half actually covers — corrected

The first version of this body and both code comments said *"every `Image` row in the app goes through `createImage`"*, and used that to justify the whole approach. **That is false.** `createImage` is the widest single funnel — post images, model-version and collection paths, comics, cover images, thumbnails — and it reaches every session regardless of which bundle it loaded, which a client-side fix cannot. But five write paths reach the `Image` table without passing through it, so their rows are in neither the numerator nor the denominator of anything this emits:

| bypass | write |
|---|---|
| `article.service.ts` `linkArticleContentImages` | `tx.image.createManyAndReturn` |
| `image.service.ts` `createEntityImages` | `dbClient.image.createMany` |
| `image.service.ts` `updateEntityImages` | `dbClient.image.createMany` |
| `blocks/app-listing-assets.service.ts` | `dbWrite.image.create` |
| `pages/api/admin/temp/migrate-article-images.ts` | `tx.image.createManyAndReturn` (one-off admin backfill) |

**The article path is the highest-risk of those, because it is fed by the exact hook defect this PR describes.** `TipTap/EdgeMediaNode.tsx` and `libs/tiptap/extensions/CustomImage.tsx` write the upload's key into article content off a promise that resolves even when the PUT was refused; `linkArticleContentImages` then materialises that content into rows — unprobed.

**Extending the probe to it is a separate change, for a structural reason.** `linkArticleContentImages` does its `createManyAndReturn` *inside* a `dbWrite.$transaction` callback, and this repo has a lint rule (`local-rules/no-io-in-transaction`) whose whole purpose is to keep network IO out of that callback, because it burns the transaction's timeout budget. Adding a HEAD there means restructuring the transaction first — not adding an argument.

**I am not replacing the old justification with a better-sounding one.** Both halves of this change are partial — the client half only reaches sessions on the new bundle, this half only reaches this funnel — and **nothing here establishes which of the two leaves more rows uncovered.** The reason the server half ships first is that it is the one that does not depend on bundle rollout, not that it is complete.

**Client half — partial, and the partiality is the decision.** The `loadend` branch updated nothing on a refused PUT, so the tracked file asserted `uploading` forever. That is a second, independent bug: `ImageUpload` renders `status === 'error'` (a badge it never got to show) and `ChallengeSubmitModal` both renders it *and* gates its submit on `status === 'uploading'`, so a refused PUT there wedges the submit with a spinner that can never clear. This PR fixes that — a pure status-reporting fix with **no change to the promise contract**, so no consumer can break.

**The promise still resolves on a refused PUT, and that is deliberate.** Making it reject is the change that stops callers persisting a key with nothing behind it, and it is *not* a one-line change.

### Consumer audit — why the reject is not in this PR

🔴 **The "24 call sites" in the first version of this body was wrong.** It came from a literal `uploadToCF(` grep, which sees only the sites whose local binding is *named* `uploadToCF` and is blind to aliased destructures and to sites that hand the function on as a value. Re-derived:

**Method** (reproducible): enumerate every non-test file under `src/**` that mentions `uploadToCF`; for each, collect the local binding names introduced by the destructure, *including aliases* (`const { uploadToCF: uploadEnhanceToCF } = useCFImageUpload()` — there are **7** such alias sites); then per file match `<name>(` for invocations and bare `<name>` for value handoffs. Classify each invocation by walking the brace stack for an enclosing `try{}catch` (a `try{}finally` does **not** count — two comics sites are exactly that), plus a chained `.catch(` found by balanced-paren chain walking, plus `await`/`return`. Then hand-read the caller frame of every site the walk did *not* put in `caught`, because a lexical walk cannot see a rejection caught one frame up.

That last step mattered twice, and both were instrument defects rather than code findings: a first pass counted `try{}finally` as catching (2 false *caught*), and a `[^;]*` chain regex stopped at the first `;` inside a `.then` body and so could not see the trailing `.catch` on the two TipTap sites (2 false *floating*).

**42 call sites: 32 direct invocations + 10 handoffs.**

| direct invocation, by how a rejection would land | count |
|---|---|
| **Caught** (enclosing `try`/`catch`, chained `.catch`, or caught one frame up) | 18 |
| **Uncaught, awaited** — unhandled rejection, silent no-op or stuck UI | 10 |
| **Uncaught inside `Promise.all`** (`ImageUpload.tsx:118`) | 1 |
| **Floating** (fire-and-forget, no `.catch`) | 3 |

| handoff — the function passed as a value | count | rejection handling |
|---|---|---|
| `fetchAndUploadGeneratorImage(url, base, uploadFn)` | 5 | all 5 wrapped in `try`/`catch` by their callers |
| `openGeneratorImagePicker({ uploadFn })` | 5 | caught inside the helper (`Promise.allSettled` + `try`/`catch`) |

So **14 of the 32 direct sites do not catch**, and all 10 handoffs do. **The correction does not invert the scoping decision** — the sites the old grep missed are mostly *uncaught-awaited*, and the handoff surface it also missed is uniformly caught. The blocker is unchanged and is the batch case: `ImageUpload.tsx` awaits `uploadToCF` inside an uncaught `Promise.all`, so one rejection aborts the whole batch — including files that uploaded fine — and their entries keep their local `blob:` URLs, which the `useEffect` then hands to `onChange` **as the persisted image values**. That is data corruption, not a no-op, in the highest-traffic upload component.

So the reject needs its own PR with the consumer work attached. Given an earlier complete version of this arc was closed as too extensive for the root cause, splitting there rather than growing this one is the deliberate call.

## Files touched

| file | why |
|---|---|
| `src/server/utils/created-image-media-probe.ts` | **new** — the probe. Four-valued verdict (`present` / `absent` / `unknown` / `not-applicable`), never throws, bounded by a 2 s abort, bucket resolved through `getImageUploadBackend()` so it asks the store the key was actually minted into. |
| `src/server/utils/__tests__/created-image-media-probe.test.ts` | **new** — 25 cases over the predicate, all four verdicts, and the delivered timeout budget. |
| `src/server/services/image.service.ts` | the call site: probe, log, continue. Mostly rationale — including the bypass list, the measured 404 fact, the loop cost, and this log's removal condition. |
| `src/server/services/__tests__/create-image-media-verify.test.ts` | **new** — the seam: one line per call whatever the verdict, url present/omitted by verdict, probe ordered before the write, and the row still created on `absent`. |
| `src/hooks/useCFImageUpload.tsx` | mark the tracked file errored on a refused PUT; widen the success test to 2xx; guard `loadend` so it cannot overwrite `aborted` with `error`. |
| `src/hooks/__tests__/useCFImageUpload.test.ts` | **new** — drives a fake XHR through every terminal sequence the spec mandates. |
| `src/server/flipt/__tests__/flipt-eval-context.test.ts` | ledger line pins move by 1 for the single new import. Mechanical. |

## Reading the measurement

The log is `create-image-media-verify`, one line per `createImage` call.

🔴 **`present` is the positive control — do not read the `absent` count without it.** A probe that can never answer emits zero `absent` verdicts, and "zero absent" is exactly what a clean result looks like. Backend resolution throws without credentials; both that and an abort land on `unknown`, and both look clean. So: require a non-zero `present` count, check that `unknown` is a small share of `present + absent + unknown`, and only then read the `absent` rate against the total line count.

🔴 **But `present` alone is NOT sufficient, and the first version of this body said it was.** `present` proves the credential can read an object that *exists*; it does not prove that a key which does **not** exist comes back as a 404 rather than a 403. `isNotFoundError` in `src/utils/s3-utils.ts` accepts only `NotFound`, `NoSuchKey` and a 404 status — so under a 403-on-missing credential every real defect would route to `unknown`, `absent` would sit at zero, and the documented control above would still read healthy.

**That half was measured rather than reasoned about.** On 2026-09-08, HeadObject was issued against the production image bucket with the production upload credential:

- a key that does **not** exist → `name=NotFound`, `$metadata.httpStatusCode=404` (no `Code`);
- a key that **does** exist → success with a non-zero `ContentLength` (the positive control).

So `absent` is reachable and the counter is not structurally pinned at zero. 🔴 **That measurement is contingent on the credential it was taken with** — a key rotation, a permissions change, or a move to a different bucket can reintroduce the hazard silently. The symptom is `absent` going to zero while `present` stays healthy; if you see that shape, re-take the measurement before concluding the defect is gone.

Settle any individual `absent` by taking its logged key and HEADing the store by hand — which is what the `url` field is for. `url` is logged only for a probed verdict, where the predicate has already established it is a 36-character uuid. `not-applicable` is by definition the arbitrary caller-supplied strings that failed that predicate, so those are omitted rather than shipped to the log sink.

### 🔴 When this log comes out

It is unsampled — one line per `createImage` call — which is the only shape that gives `absent` a denominator, and also the reason it must not live here indefinitely.

**Closing condition:** a merged PR that reads the `present` / `absent` / `unknown` counts for `create-image-media-verify`, quotes those three numbers in its body, and then either (a) adds the enforcement branch this measurement exists to size, or (b) deletes this call outright because the rate does not justify one. Either way that PR removes or samples down the `logToAxiom`.

**Mechanical check that it is closed:** the string `create-image-media-verify` no longer appears in `src/server/services/image.service.ts`.

## Verification

Every guard was watched fail before it passed. 12 mutants, each killed by the specific test for that guard, with that test's own assertion:

| mutation | test that failed | message |
|---|---|---|
| restore the original `loadend` branch | 6 hook tests | `expected [ 'uploading' ] to deeply equal [ 'error' ]` |
| remove the `loadend` terminal guard | canceled-upload test | `expected [ 'error' ] to deeply equal [ 'aborted' ]` |
| narrow 2xx back to `=== 200` | 201 / 204 tests | `expected [ 'error' ] to deeply equal [ 'success' ]` |
| drop the zero-length check | zero-length test | `expected 'present' to be 'absent'` |
| `size === 0` → `!size` | `size: null` test | `expected 'absent' to be 'present'` |
| catch returns `absent` | both fail-open tests | `expected 'absent' to be 'unknown'` |
| widen the predicate to any string | not-applicable test | `expected 'absent' to be 'not-applicable'` |
| move backend resolution outside the try | backend-unresolvable test | `promise rejected "Error: no credentials" instead of resolving` |
| log only on an `absent` verdict | 5 seam tests | `expected [] to have a length of 1` |
| log `url` unconditionally | not-applicable url test | `expected 'some-file…png' to be null` |
| throw on `absent` | 3 seam tests | `promise rejected "Error: media absent" instead of resolving` |
| relocate the probe below the write | ordering test | `expected 42 to be less than 41` |

🔴 **One of those twelve was added this round, because the timeout budget was pinned as a constant rather than as behaviour.** The old tests asserted that `CREATED_IMAGE_MEDIA_PROBE_TIMEOUT_MS === 2000` and that *an* `AbortSignal` was passed — neither of which can see a probe that declares a 2 s budget and then bounds the request with something else. Measured: mutating the call to `AbortSignal.timeout(600000)` left **all 24 probe tests green, rc 0** — a 10-minute deadline on a user-facing mutation, invisible.

The new test (`hands the store a signal carrying THAT budget, not merely some signal`) spies on `AbortSignal.timeout` and asserts the delivered value against a **literal**, so a mutant that moves the constant *and* the call site together still fails. Both directions were watched go red:

| mutant | result |
|---|---|
| `AbortSignal.timeout(CREATED_IMAGE_MEDIA_PROBE_TIMEOUT_MS)` → `AbortSignal.timeout(600000)` | `1 failed \| 24 passed (25)` — only the new test, `AssertionError: expected 600000 to be 2000 // Object.is equality` at `created-image-media-probe.test.ts:160` |
| the constant itself `2000` → `600000` | `2 failed \| 23 passed (25)` — the new test **and** the pre-existing constant test, same message |
| unmutated | `25 passed (25)` |

- `pnpm typecheck` — **0 type errors in 152 s** (heap cap 8192 MB).
- `pnpm vitest run --project unit` — **1677 files, 38,642 passed**, 33 skipped, **3 failed**. All 3 are in `eventloop-watchdog.capture.test.ts` (`starvation classifier`) and `moderation.instrumentation.test.ts` — timing tests that race real wall-clock deadlines. 🔴 **Confirmed pre-existing and load-sensitive, not caused by this change, by control rather than by assertion:** a *different* test in that describe block fails on each run, and running the file at the branch head with all six of my edits reverted failed **on both control runs, with two different tests**. `moderation.instrumentation.test.ts` passes in isolation. The four test files this PR touches: **56 passed (56)**.
- `prettier --check` on all 6 changed source files — clean. Negative control: an added `const   x   =    1` made it report `[warn] src/hooks/useCFImageUpload.tsx`, so the check does go red on these paths.
- `eslint` on all 6 files — my new files are clean; the 3 errors and 14 warnings reported all sit in pre-existing untouched regions of `image.service.ts` (lines 211, 4052, 5657) and `useCFImageUpload.tsx:234`.

### 🔴 CI is only PARTIALLY reporting on this head, and the reason is a conflict with `main`

At the current head, `gh pr checks` reports **7 checks, all `SUCCESS`** (stable across two reads):
the five `preview / *` statuses plus `tekton / fixture-bootstrap` and `tekton / typecheck`.
The previous head also carried **13 GitHub-Actions check-runs** (`App unit tests + typecheck`,
`ESLint + Prettier`, the four `Unit tests (n)` shards, `Package unit tests`, `Geometry tests`,
`Schema drift gate`, `event-engine-common pin`, both `Windows dev-env` jobs, and a skipped
`Typecheck`). **Those 13 have not run on this head, and will not until the conflict below is
resolved** — all four workflows are `pull_request`-triggered, and GitHub cannot build the merge
ref for an unmergeable PR, so no event dispatches. So this is **not** a green CI claim: it is
7 green and 13 unrun.

**The conflict is pre-existing, not introduced by this round.** `git merge-tree --write-tree`
against `origin/main` conflicts at the *previous* head as well as this one, in the same single
file — `src/hooks/useCFImageUpload.tsx`, against `44ff54d2d9` (#4573, "relay an image upload
through our own origin when the direct PUT cannot resolve the storage host"), which landed on
`main` and edits the same upload path this PR does.

I deliberately did **not** resolve it: #4573's origin-relay change and this PR's `loadend`
rewrite touch the same terminal-event handling, so merging them is a semantic call about which
statuses a relayed PUT should report — not a mechanical one, and not something to fold silently
into a correction round.

## Deliberately left out

- **The reject in `uploadToCF`** — see the consumer audit above.
- **Extending the probe to the five bypass paths**, above all `linkArticleContentImages`. That one is blocked on the `no-io-in-transaction` constraint, not on effort.
- **An enforcement flag.** Observe-only is enforced by construction, not by a default-off branch.
- **Consolidating the media-key predicate.** #4490 is adding an identical `isProbeableMediaKey` to `@civitai/shared`; this repo forbids stacked PRs, so importing it here would base this change on that unmerged branch. The duplicate is documented at the declaration and must be consolidated by whichever lands second. Closing condition is mechanical: `created-image-media-probe.ts` no longer declares `MEDIA_KEY_RE`.
- **Browser-project tests were not run locally** (Playwright browsers are not installed in my worktree). All 9 `.browser.test.tsx` files that touch this hook `vi.mock` it wholesale, so they stub its internals entirely and cannot be affected — but that is reasoning, not a run. CI covers it.

## Overlap with #4490

Both PRs touch `image.service.ts` and both shift the flipt-eval ledger. The code regions are far apart — #4490 is in `resolveIngestionError` (~line 8060+) and its import edit is at the s3-utils import; mine is in `createImage` (~line 6100) with its import ~20 lines away — so a textual conflict is unlikely, but **whichever merges second must re-run the flipt-eval gate and take the line numbers it prints.** Both are based directly on `main`; neither is stacked.

*(The two claims about #4490's own contents in that paragraph are carried over from the first version of this body and were **not** re-verified this round — I deliberately did not read that branch, since nothing here may be based on it. Treat them as approximate. What I did verify: this round's edits all sit **below** the four ledger-pinned lines, so the gate is unchanged and still green — 12 passed.)*

## Cost, and two measurement caveats a reader should know

**One HEAD per `createImage` call, bounded at 2 s per network attempt.** Most call sites are single-image (a post image, a cover, a block upload, a thumbnail), where this is one bounded round-trip on a mutation that already does several.

🔴 **Two `comics.router.ts` procedures call `createImage` once per input item**, and the first version of this body named only one of them:

| procedure | loop | cap |
|---|---|---|
| `bulkCreatePanels` | `for (let i = 0; i < input.panels.length; i++)` | `panels` `.max(20)` |
| `addReferenceImages` | `for (const img of input.images)` | `images` `.max(10)` |

Both are sequential, so they add up to 20 and up to 10 sequential HEADs respectively.

🔴 **And the 2 s bound is per-attempt, not wall-clock — so the degraded case is bigger than 20 × 2 s.** As `src/utils/s3-utils.ts` records on `checkFileExists` and `headObject`, the signal is shared by every retry attempt but the SDK sleeps *between* attempts on a plain, non-abort-aware timer: a deadline landing mid-backoff lets that sleep run to completion and only the next attempt short-circuits. Worst case per call is therefore the budget **plus one backoff**. `getB2ImageS3Client` sets neither `maxAttempts` nor a request-handler timeout, so the backoff length is whatever the SDK default schedule produces — s3-utils records one measurement against the installed SDK, a 300 ms budget with a ~5 s backoff in flight returning in ~4.7 s. I have not measured this probe's own degraded wall time, so I am not quoting a total; what I will say is that against a degraded store a full 20-panel import adds twenty of those, not forty seconds.

The reason I judged it acceptable rather than restructuring: both loops are *already* N sequential IO round-trips per item — a DB write plus an ingestion call — so this is a proportional addition to an existing sequential-IO path rather than a new failure class, and it is abort-bounded where the existing steps are not. But "proportional" is a claim about shape, not about the absolute number, and the absolute number stays unbounded until `getB2ImageS3Client` sets a `maxAttempts`. If a reviewer disagrees, the cheap fix is to stop awaiting the probe (nothing branches on the verdict) — I kept it awaited so the verdict describes the store's state *at the moment the row is written*, which is the measurement enforcement would later act on.

🔴 **Cover images are a blind spot in this log, and their `absent` count will read as zero for the wrong reason.** `resolveCoverImageId` already runs an *enforcing* existence check immediately before calling `createImage`, so a definitively-absent cover is rejected before it ever reaches this probe. Cover-image defects therefore cannot appear here — that zero is a pre-filter, not evidence. That path also now does two HEADs for the same key; harmless (both bounded, both fail-open) but real. Consolidating the two is a natural follow-up once the shared predicate lands.

### DIFF
--- a/src/hooks/__tests__/useCFImageUpload.test.ts
+++ b/src/hooks/__tests__/useCFImageUpload.test.ts
@@ -188,11 +188,12 @@ describe('useCFImageUpload — the tracked file must not lie about a refused PUT
      * 🔴 THE REMAINING HALF OF THE DEFECT, PINNED RATHER THAN FIXED. `uploadToCF` still
      * RESOLVES on a refused PUT, handing the caller a media key with nothing behind it —
      * which is how an `Image` row gets written for media that never landed. Making it
-     * reject is the follow-up, and it is not a one-line change: 11 of 24 call sites do not
-     * catch, and `ImageUpload`'s uncaught `Promise.all` would propagate local `blob:` URLs
-     * into `onChange` as image values. Asserting the current behaviour makes that follow-up
-     * a deliberate edit to this expectation instead of a silent drift, and the observe-only
-     * probe in `createImage` covers the exposure server-side meanwhile.
+     * reject is the follow-up, and it is not a one-line change: 14 of the 32 direct call
+     * sites do not catch, and `ImageUpload`'s uncaught `Promise.all` would propagate local
+     * `blob:` URLs into `onChange` as image values. Asserting the current behaviour makes
+     * that follow-up a deliberate edit to this expectation instead of a silent drift, and
+     * the observe-only probe in `createImage` covers the `createImage` funnel server-side
+     * meanwhile — not every `Image` row; see that call site for the paths it misses.
      */
     expect(settled.error).toBeUndefined();
     expect(settled.value).toEqual({
--- a/src/hooks/__tests__/useCFImageUpload.test.ts
+++ b/src/hooks/__tests__/useCFImageUpload.test.ts
@@ -1,0 +1,237 @@
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import * as React from 'react';
+import type { act as actType } from 'react-dom/test-utils';
+import { createRoot } from 'react-dom/client';
+
+// React 18.3 exposes `act` on the `react` export, but our @types/react predates that typing.
+const act = (React as unknown as { act: typeof actType }).act;
+// Without this React warns on every `act(...)` and does not flush its queue the same way.
+(globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
+
+vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => null }));
+vi.mock('~/utils/notifications', () => ({ showErrorNotification: vi.fn() }));
+vi.mock('~/utils/media-preprocessors', () => ({
+  preprocessFile: vi.fn(async () => ({
+    type: 'image',
+    objectUrl: 'blob:fake-object-url',
+    meta: {},
+    metadata: { width: 64, height: 48, hash: 'LKO2' },
+  })),
+  auditImageMeta: vi.fn(async () => ({ blockedFor: undefined })),
+}));
+
+import { useCFImageUpload } from '~/hooks/useCFImageUpload';
+
+/** The key the (stubbed) sign endpoint hands back. Distinct from every other literal here. */
+const UPLOAD_ID = 'd47b16f0-9a25-4e88-8c31-71fe0a3b62d4';
+
+type Listeners = Record<string, Array<() => void>>;
+
+/**
+ * A hand-rolled XHR that lets a test decide which terminal event sequence the browser
+ * delivers. The sequences are the ones the spec mandates and they are the whole point of
+ * the code under test: `error` and `abort` are EACH FOLLOWED BY `loadend`, while a
+ * completed-but-refused request delivers `loadend` alone.
+ */
+class FakeXHR {
+  static last: FakeXHR;
+  readyState = 0;
+  status = 0;
+  upload = { addEventListener: (t: string, f: () => void) => this.add(this.uploadListeners, t, f) };
+  private listeners: Listeners = {};
+  private uploadListeners: Listeners = {};
+
+  constructor() {
+    FakeXHR.last = this;
+  }
+  private add(bag: Listeners, type: string, fn: () => void) {
+    (bag[type] ??= []).push(fn);
+  }
+  addEventListener(type: string, fn: () => void) {
+    this.add(this.listeners, type, fn);
+  }
+  open() {
+    /* no-op */
+  }
+  send() {
+    /* no-op — the test drives the terminal events itself */
+  }
+  abort() {
+    /* no-op */
+  }
+  private fire(type: string) {
+    for (const fn of this.listeners[type] ?? []) fn();
+  }
+  /** A request that completed with this status, i.e. `load` -> `loadend`, no `error`. */
+  completeWith(status: number) {
+    this.readyState = 4;
+    this.status = status;
+    this.fire('loadend');
+  }
+  /** A transport-level failure: `error` then `loadend`. */
+  networkError() {
+    this.readyState = 4;
+    this.status = 0;
+    this.fire('error');
+    this.fire('loadend');
+  }
+  /** A user cancel: `abort` then `loadend`. */
+  cancel() {
+    this.readyState = 0;
+    this.status = 0;
+    this.fire('abort');
+    this.fire('loadend');
+  }
+}
+
+type Harness = {
+  upload: (file: File) => Promise<unknown>;
+  statuses: () => string[];
+  unmount: () => void;
+};
+
+async function mountHook(): Promise<Harness> {
+  let api: ReturnType<typeof useCFImageUpload> | undefined;
+  function Probe() {
+    api = useCFImageUpload();
+    return null;
+  }
+  const container = document.createElement('div');
+  const root = createRoot(container);
+  await act(async () => {
+    root.render(React.createElement(Probe));
+  });
+  return {
+    upload: (file) => api!.uploadToCF(file),
+    statuses: () => api!.files.map((f) => f.status),
+    unmount: () => act(() => root.unmount()),
+  };
+}
+
+/** Start an upload, let it reach the point where the XHR has been sent, and hand back both. */
+async function startUpload(h: Harness) {
+  const settled: { value?: unknown; error?: unknown } = {};
+  let promise!: Promise<unknown>;
+  await act(async () => {
+    promise = h.upload(new File(['x'], 'a.png', { type: 'image/png' }));
+    promise.then(
+      (value) => (settled.value = value),
+      (error) => (settled.error = error)
+    );
+    // Let the sign fetch + preprocess microtasks drain so the XHR exists.
+    await new Promise((r) => setTimeout(r, 0));
+  });
+  return { promise, settled, xhr: FakeXHR.last };
+}
+
+describe('useCFImageUpload — the tracked file must not lie about a refused PUT', () => {
+  beforeEach(() => {
+    vi.stubGlobal('XMLHttpRequest', FakeXHR);
+    vi.stubGlobal(
+      'fetch',
+      vi.fn(async () => ({ json: async () => ({ id: UPLOAD_ID, uploadURL: 'https://store/put' }) }))
+    );
+  });
+  afterEach(() => vi.unstubAllGlobals());
+
+  it('marks the file successful on a 200', async () => {
+    const h = await mountHook();
+    const { settled, xhr } = await startUpload(h);
+
+    await act(async () => {
+      xhr.completeWith(200);
+      await new Promise((r) => setTimeout(r, 0));
+    });
+
+    expect(h.statuses()).toEqual(['success']);
+    expect(settled.value).toEqual({
+      url: 'https://store/put',
+      id: UPLOAD_ID,
+      objectUrl: 'blob:fake-object-url',
+      type: 'image',
+    });
+    h.unmount();
+  });
+
+  it.each([201, 204])('marks the file successful on a %i', async (status) => {
+    // 🔴 A 2xx is a successful PUT by any reading. Reporting one as an error would put a
+    // failure badge on an upload that worked.
+    const h = await mountHook();
+    const { xhr } = await startUpload(h);
+
+    await act(async () => {
+      xhr.completeWith(status);
+      await new Promise((r) => setTimeout(r, 0));
+    });
+
+    expect(h.statuses()).toEqual(['success']);
+    h.unmount();
+  });
+
+  it.each([400, 403, 500, 503])('marks the file errored on a refused %i', async (status) => {
+    // 🔴 THE DEFECT THIS FIXES. A refused PUT never fires `error` — it completes and lands
+    // on `loadend`. Before this change the branch updated nothing, so the tracked file
+    // asserted `uploading` forever: `ImageUpload` never showed its error badge and
+    // `ChallengeSubmitModal`'s submit gate (`some(f => f.status === 'uploading')`) could
+    // never clear.
+    const h = await mountHook();
+    const { settled, xhr } = await startUpload(h);
+
+    await act(async () => {
+      xhr.completeWith(status);
+      await new Promise((r) => setTimeout(r, 0));
+    });
+
+    expect(h.statuses()).toEqual(['error']);
+    /**
+     * 🔴 THE REMAINING HALF OF THE DEFECT, PINNED RATHER THAN FIXED. `uploadToCF` still
+     * RESOLVES on a refused PUT, handing the caller a media key with nothing behind it —
+     * which is how an `Image` row gets written for media that never landed. Making it
+     * reject is the follow-up, and it is not a one-line change: 11 of 24 call sites do not
+     * catch, and `ImageUpload`'s uncaught `Promise.all` would propagate local `blob:` URLs
+     * into `onChange` as image values. Asserting the current behaviour makes that follow-up
+     * a deliberate edit to this expectation instead of a silent drift, and the observe-only
+     * probe in `createImage` covers the exposure server-side meanwhile.
+     */
+    expect(settled.error).toBeUndefined();
+    expect(settled.value).toEqual({
+      url: 'https://store/put',
+      id: UPLOAD_ID,
+      objectUrl: 'blob:fake-object-url',
+      type: 'image',
+    });
+    h.unmount();
+  });
+
+  it('marks the file errored and rejects on a network failure', async () => {
+    const h = await mountHook();
+    const { settled, xhr } = await startUpload(h);
+
+    await act(async () => {
+      xhr.networkError();
+      await new Promise((r) => setTimeout(r, 0));
+    });
+
+    expect(h.statuses()).toEqual(['error']);
+    expect((settled.error as Error).message).toBe('Upload failed (status 0)');
+    h.unmount();
+  });
+
+  it('keeps a canceled upload reported as aborted, not errored', async () => {
+    // 🔴 `abort` is FOLLOWED by `loadend`. Without the terminal guard the `loadend` handler
+    // re-decides the status and overwrites the user's cancel with a failure — this test
+    // fails without it.
+    const h = await mountHook();
+    const { settled, xhr } = await startUpload(h);
+
+    await act(async () => {
+      xhr.cancel();
+      await new Promise((r) => setTimeout(r, 0));
+    });
+
+    expect(h.statuses()).toEqual(['aborted']);
+    expect((settled.error as Error).message).toBe('Upload canceled');
+    h.unmount();
+  });
+});
--- a/src/hooks/__tests__/useCFImageUpload.test.ts
+++ b/src/hooks/__tests__/useCFImageUpload.test.ts
@@ -76,9 +76,27 @@ class FakeXHR {
     this.fire('error');
     this.fire('loadend');
   }
-  /** A user cancel: `abort` then `loadend`. */
+  /**
+   * A user cancel: `abort` then `loadend`.
+   *
+   * 🔴 `readyState` is 4 — the same value `src/utils/__tests__/upload-settlement.test.ts`
+   * uses, and for the same reason: `abort()` runs the request-error steps, which set the
+   * state to DONE and fire both events, and only afterwards is the state reset to UNSENT.
+   * This stub said 0, i.e. two stubs of the same browser event disagreed about it, and one
+   * of them had to be wrong.
+   *
+   * ⚠ THE SCOPE OF THAT FIX, MEASURED RATHER THAN ASSUMED. A round-2 audit predicted the
+   * old `0` made the cancel case pass for the wrong reason — `success` computing false via
+   * the `readyState === 4` clause instead of via `status === 0`. That prediction does not
+   * hold here: deleting the `aborted` guard from `attachUploadSettlement` turns this case
+   * red under BOTH values, identically (`expected [ 'error' ] to deeply equal [ 'aborted' ]`,
+   * 1 failed | 8 passed), because with the guard gone `success` is false either way and
+   * `onError` overwrites the cancel regardless of which clause produced the false. So this
+   * is a FIDELITY fix — the stub now models the browser and agrees with its sibling — and
+   * not a discrimination fix. The case was already discriminating.
+   */
   cancel() {
-    this.readyState = 0;
+    this.readyState = 4;
     this.status = 0;
     this.fire('abort');
     this.fire('loadend');
@@ -188,9 +206,19 @@ describe('useCFImageUpload — the tracked file must not lie about a refused PUT
      * 🔴 THE REMAINING HALF OF THE DEFECT, PINNED RATHER THAN FIXED. `uploadToCF` still
      * RESOLVES on a refused PUT, handing the caller a media key with nothing behind it —
      * which is how an `Image` row gets written for media that never landed. Making it
-     * reject is the follow-up, and it is not a one-line change: 14 of the 32 direct call
-     * sites do not catch, and `ImageUpload`'s uncaught `Promise.all` would propagate local
-     * `blob:` URLs into `onChange` as image values. Asserting the current behaviour makes
+     * reject is the follow-up, and it is not a one-line change: most direct call sites do
+     * not catch, and `ImageUpload`'s uncaught `Promise.all` would propagate local `blob:`
+     * URLs into `onChange` as image values.
+     *
+     * ⚠ There was a precise count here ("14 of the 32 direct call sites"). It is gone
+     * deliberately, and NOT because the qualitative claim weakened — the `Promise.all`
+     * consequence above is verified at `src/components/ImageUpload/ImageUpload.tsx`. The
+     * count is gone because it was derived four times and moved every time (24 → ~38 → 42
+     * → 32/14), a round-2 audit independently got a different numerator, and nothing in
+     * this repo asserts on any of those numbers, so the figure could only ever rot. Do not
+     * reinstate one without a test that fails when it drifts.
+     *
+     * Asserting the current behaviour makes
      * that follow-up a deliberate edit to this expectation instead of a silent drift, and
      * the observe-only probe in `createImage` covers the `createImage` funnel server-side
      * meanwhile — not every `Image` row; see that call site for the paths it misses.
--- a/src/server/flipt/__tests__/flipt-eval-context.test.ts
+++ b/src/server/flipt/__tests__/flipt-eval-context.test.ts
@@ -152,14 +152,14 @@ const ENTITY_WITHOUT_CONTEXT_LEDGER: Record<string, string> = {
   'server/services/article-rating-review.helpers.ts:482':
     'article-rating-dispute has no segment rollouts; background path with no SessionUser',
   // flag `feed-fetch-filter-in-post`: enabled=true, 0 rules, 0 rollouts.
-  'server/services/image.service.ts:3267':
+  'server/services/image.service.ts:3268':
     'feed-fetch-filter-in-post has no segment rollouts; hot feed path',
   // flag `feed-image-existence`: enabled=true, 0 rules, 0 rollouts. Three sites.
-  'server/services/image.service.ts:3311':
+  'server/services/image.service.ts:3312':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:4080':
+  'server/services/image.service.ts:4081':
     'feed-image-existence has no segment rollouts; hot feed path',
-  'server/services/image.service.ts:4890':
+  'server/services/image.service.ts:4891':
     'feed-image-existence has no segment rollouts; hot feed path',
   // flags `model-text-moderation-xguard` / `-apply`: both enabled=true, 0 rules,
   // 0 rollouts (checked against flipt-state and against the evaluation API on
@@ -219,7 +219,7 @@ describe('flipt evaluation context — source gate', () => {
     // these and fail the second, and vice versa.
     const bySite = new Map(calls.map((c) => [c.site, c]));
     expect(bySite.get('server/services/feedback.service.ts:43')?.argc).toBe(3);
-    expect(bySite.get('server/services/image.service.ts:3311')?.argc).toBe(2);
+    expect(bySite.get('server/services/image.service.ts:3312')?.argc).toBe(2);
   });
 
   it('adds no Flipt evaluation that names an entity but passes no context', () => {
--- a/src/server/services/__tests__/create-image-media-verify.test.ts
+++ b/src/server/services/__tests__/create-image-media-verify.test.ts
@@ -1,0 +1,132 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { CreatedImageMediaVerdict } from '~/server/utils/created-image-media-probe';
+
+const { probeMock } = vi.hoisted(() => ({
+  probeMock: vi.fn<() => Promise<CreatedImageMediaVerdict>>(),
+}));
+
+vi.mock('~/server/utils/created-image-media-probe', () => ({
+  probeCreatedImageMedia: probeMock,
+}));
+
+import { dbMock } from '~/__tests__/mocks/db.mock';
+import { loggingMock } from '~/__tests__/mocks/logging.mock';
+import { createImage } from '~/server/services/image.service';
+
+/**
+ * Distinct from every constant asserted below, so a mutant that logs a hardcoded key or the
+ * wrong field cannot pass by coincidence.
+ */
+const KEY = '3f6c2b91-0d84-4a15-9e70-c2b8a4d15e33';
+const USER_ID = 8113;
+const POST_ID = 55207;
+const CREATED_ID = 90210;
+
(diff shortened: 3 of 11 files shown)


ZacxDev:
## Correction round — three claims in the original body were wrong, and I have edited it

A reviewer may already have read the first version. Editing the body silently would hide
what changed, so here is the list. Pushed as one commit on top of the branch; no behaviour
changed except one new test.

### 1. 🔴 "Every `Image` row in the app goes through `createImage`" — **false**

This was the stated justification for the whole server-half approach, and it appeared in
the body *and* in two code comments. Five write paths reach the `Image` table without
passing through `createImage`, so their rows are in neither the numerator nor the
denominator of anything the probe emits:

1. `article.service.ts` `linkArticleContentImages` — `tx.image.createManyAndReturn`
2. `image.service.ts` `createEntityImages` — `dbClient.image.createMany`
3. `image.service.ts` `updateEntityImages` — `dbClient.image.createMany`
4. `blocks/app-listing-assets.service.ts` — `dbWrite.image.create`
5. `pages/api/admin/temp/migrate-article-images.ts` — a one-off admin backfill

**The article path is the highest-risk**, because it is fed by the exact hook defect this
PR describes: the TipTap nodes write the upload key into article content off a promise that
resolves even when the PUT was refused, and `linkArticleContentImages` then materialises
that content into rows — unprobed.

**I fixed the claim, not the coverage.** `linkArticleContentImages` writes inside a
`dbWrite.$transaction` callback, and `local-rules/no-io-in-transaction` exists precisely to
keep network IO out of there, so extending the probe means restructuring the transaction —
a separate change, recorded as such at the call site.

🔴 **And I did not swap in a better-sounding justification.** Both halves of this PR are
partial (the client half only reaches new bundles; this half only reaches this funnel) and
**nothing establishes which leaves more rows uncovered.** The server half ships first
because it does not depend on bundle rollout, not because it is complete.

### 2. 🔴 The documented positive-control protocol was insufficient — and I measured the missing half

The body and the code told you to require a non-zero `present` count before trusting
`absent`. That is not enough in principle: `present` proves the credential can read an
object that **exists**; it does not prove a **missing** key answers 404 rather than 403.
`isNotFoundError` (`src/utils/s3-utils.ts`) accepts only `NotFound` / `NoSuchKey` / 404, so
a 403-on-missing credential would route every real defect to `unknown` and pin `absent` at
zero — while satisfying the documented control exactly.

Measured on **2026-09-08** with HeadObject against the production image bucket using the
production upload credential:

- a key that does **not** exist → `name=NotFound`, `$metadata.httpStatusCode=404`
- a key that **does** exist → success with a non-zero `ContentLength` (positive control)

So `absent` is reachable and the counter is not structurally stuck. Stated at that scope in
the body and the code — **this bucket, this credential, as of that date** — with the
reader's warning kept: a credential rotation or bucket move can reintroduce the hazard, and
the symptom is `absent` going to zero while `present` stays healthy.

### 3. 🔴 The consumer count was wrong, in three places

`24` was quoted in the body table, in `useCFImageUpload.tsx`, and in
`useCFImageUpload.test.ts`. It came from a literal `uploadToCF(` grep, which sees only the
sites whose local binding is *named* `uploadToCF` and is blind to the **7** aliased
destructures (`const { uploadToCF: uploadEnhanceToCF } = …`) and to every site that hands
the function on as a value.

Re-derived — **42 call sites: 32 direct invocations + 10 handoffs**, of which **14 of the
32 direct sites do not catch** and all 10 handoffs do. Method and the full breakdown are in
the body's consumer audit, including the two instrument defects the derivation hit (a first
pass counted `try{}finally` as catching; a chain regex could not see a `.catch` that
followed a `.then` containing a `;`).

**The scoping decision is unchanged** — the newly-found direct sites are mostly
uncaught-awaited and the handoff surface is uniformly caught, so making `uploadToCF` reject
is still a follow-up with consumer work attached, blocked on the same `ImageUpload.tsx`
`Promise.all` case.

### 4. The timeout budget was pinned as a constant, not as behaviour

The old tests asserted `CREATED_IMAGE_MEDIA_PROBE_TIMEOUT_MS === 2000` and that *an*
`AbortSignal` was passed. Neither can see a probe that declares 2 s and then bounds the
request with something else: mutating the call to `AbortSignal.timeout(600000)` left **all
24 probe tests green, rc 0** — a 10-minute deadline on a user-facing mutation, invisible.

New test: `hands the store a signal carrying THAT budget, not merely some signal`. It spies
on `AbortSignal.timeout` and asserts the delivered value against a **literal**, so a mutant
that moves the constant *and* the call site together still fails. Both directions watched
go red:

| mutant | result |
|---|---|
| `AbortSignal.timeout(CREATED_IMAGE_MEDIA_PROBE_TIMEOUT_MS)` → `(600000)` | `1 failed \| 24 passed (25)` — only the new test: `AssertionError: expected 600000 to be 2000 // Object.is equality` at `created-image-media-probe.test.ts:160` |
| the constant `2000` → `600000` | `2 failed \| 23 passed (25)` — the new test **and** the pre-existing constant test |
| unmutated | `25 passed (25)` |

### 5. Degraded-store latency was understated, and one loop was unnamed

The body named `bulkCreatePanels` (`panels` `.max(20)`) but not `addReferenceImages`
(`images` `.max(10)`) — both call `createImage` once per input item, sequentially. And the
2 s bound is **per network attempt, not wall-clock**: as `s3-utils` records, the SDK sleeps
between retries on a non-abort-aware timer, so worst case per call is the budget plus one
backoff, and `getB2ImageS3Client` sets neither `maxAttempts` nor a request-handler timeout.
Both loops and the degraded sizing are now stated at the call site. I did **not** quote a
degraded wall-clock total, because I have not measured this probe's own.

### 6. The new per-mutation log had no removal condition

`create-image-media-verify` is unsampled — one line per `createImage` call — and nothing
recorded when it comes out. Closing condition now written at the call site and in the body:
a merged PR that reads the `present` / `absent` / `unknown` counts, quotes those three
numbers in its body, and then either adds the enforcement branch or deletes the call.
Mechanical check that it is closed: the string `create-image-media-verify` no longer appears
in `image.service.ts`.

### Deliberately not done

- **F6** (merging two populations under `not-applicable`) — the 🟢 the audit marked
  optional. Skipped; the verdict still merges "not a media key we own" with "a key shape we
  decline to probe", and nothing here changes that.
- **Extending the probe to the five bypass paths.** Blocked on the transaction constraint
  above, not on effort.
- **Reading #4490.** This PR must not be based on it, so the two claims about its internals
  in the "Overlap" section are carried over from the original body and were not re-verified
  this round. They are flagged as such there.

ZacxDev:
🔴 **Heads-up on CI, added to the body after the correction round pushed.** This head reports **7 checks, all green** — but the **13 GitHub-Actions check-runs the previous head carried have not run**, because the PR now **conflicts with `main`** and all four workflows are `pull_request`-triggered (GitHub cannot build a merge ref for an unmergeable PR, so no event dispatches). Please do not read the green as a full pass.

The conflict is **pre-existing, not from this round**: `git merge-tree --write-tree` against `origin/main` conflicts at the previous head too, in the same single file — `src/hooks/useCFImageUpload.tsx`, against `44ff54d2d9` (#4573, the origin-relay upload change).

I did not resolve it: #4573's relay path and this PR's `loadend` rewrite touch the same terminal-event handling, so which statuses a *relayed* PUT should report is a semantic call, not a mechanical merge — and not something to fold silently into a correction round.

ZacxDev:
Round 1 adversarial audit ran against `493202b8a6`; its findings were fixed in `e4453cab89`, and the branch was then rebased onto `main` (force-push `e4453cab89` → `000dad1040`) because a conflict introduced by #4573 was leaving 13 of the 20 checks structurally unrun.

Recording what this round's fixes **claim** to have addressed, so the next round can check each against the diff rather than against this list.

```audit-claims round=1 audited=493202b8a6c8dc7d0f76a29b1dd1b1e6e1c1d0a0..000dad10
1. F1 — the claim "every Image row goes through createImage" is retracted in both places that asserted it; the five bypassing call sites are enumerated, the article/TipTap path is named as the highest-risk uncovered producer, and the $transaction + no-io-in-transaction constraint is recorded as why extending coverage is a separate change. No replacement justification is offered for the retracted claim.
2. F2 — the positive-control protocol no longer rests on `present > 0`; it states the measured HeadObject behaviour (missing key -> NotFound/404, so `absent` is reachable), scoped to that bucket/credential/date, and keeps a credential-rotation caveat.
3. F3 — the probe's timeout budget is pinned behaviourally, not just as a constant: a mutant that raises the real budget now fails a test.
4. F4 — the "call sites that do not catch" figure was re-derived, and after the rebase is no longer quoted in source at all; it was unpinned by construction and is deliberately not reinstated.
5. F5 — the second bounded loop (addReferenceImages, max 10) is named alongside bulkCreatePanels, and the degraded-store case is sized as budget-plus-one-backoff per call rather than only the healthy case.
6. F7 — the new per-mutation log carries a closing condition naming what ends it and the mechanical check for it.
7. F6 was deliberately not done (a nit: `not-applicable` merges two populations).
8. Post-audit, unaudited: the client fix was RELOCATED into src/utils/upload-settlement.ts, because #4573 moved terminal-event handling out of the hook. It widens the success test from `=== 200` to 2xx, adds the missing `onError()` on the non-2xx branch, and adds an `aborted` flag so the new error branch cannot overwrite a user cancel. The hook's own source is now identical to main.
```

Verification carried out on the rebased head: `merge-tree` exits 0 against current `main`; 20 checks, all terminal, none failing, stable across two spaced reads; typecheck 0 errors; unit suite 38,813 passed with a single failure in `eventloop-watchdog.capture.test.ts`, a file this branch does not touch and which fails nondeterministically in isolation on the same tree (1 of 3 identical runs, a different test each time).

Item 8 is new payload that no audit round has seen.

ZacxDev:
## Corrections to what I said earlier on this PR

Three claims I made on this PR were wrong. Correcting them here rather than editing the earlier comment, because a reviewer may already have read the wrong version.

**1. "No longer quoted in source at all" was FALSE.**

Item 4 of my audit-claims block said the consumer-count figure *"after the rebase is no longer quoted in source at all"*. It was. It was still at `src/hooks/__tests__/useCFImageUpload.test.ts:191`, reading *"14 of the 32 direct call sites do not catch"*. The rebase removed it from `useCFImageUpload.tsx` only; I checked with a search piped through `head -5` and the test file sorted sixth. I turned a truncated list into a negative claim, in a commit message and in a public comment.

It is removed now, and no number replaces it. That figure was derived four times and moved every time (24 → ~38 → 42 → 32 direct / 14 uncaught), a fresh derivation got 13 rather than 14 for the numerator, and nothing in the repo asserts on it — so it was unpinned by construction and could only rot. The comment now states the shape instead ("most direct call sites do not catch"), which is the part that is actually verified. The `ImageUpload` `Promise.all` consequence it supported is verified and kept.

**2. The red count was understated: 6 of 9, not 4 of 9.**

The last commit message said "with the source fix reverted, 4 of the 9 hook tests fail". Re-measured against the parent version of `upload-settlement.ts`:

```
6 failed | 3 passed (9)
  expected [ 'uploading' ] to deeply equal [ 'error' ]    x4   (400/403/500/503)
  expected [ 'uploading' ] to deeply equal [ 'success' ]  x2   (201/204)
```

The 201/204 pair was missed. Conservative direction, still wrong.

**3. The bypassing-write-path enumeration said five. It is six.**

`duplicateImage` in `src/server/jobs/daily-challenge-processing.ts` is a raw `INSERT INTO "Image" (…) SELECT … FROM "Image"` that re-owns an existing row's columns — `url` included — to another user for the challenge cover. It reaches the table without `createImage` and was not in the list.

I re-derived the whole population rather than patching the count: every `.image.(create|createMany|createManyAndReturn|upsert)` call and every raw `INSERT INTO "Image"` under `src/`, excluding tests, plus checks for Prisma nested creates, bracket-notation model access, and the two dynamic-table INSERTs (both metric/vote tables, not `Image`). Six is exhaustive at this ref. The comment now names the population it asserts over, since the claim has been wrong once.

Its risk is *lower* than the article path, and the comment says so: the `url` is copied from a row that already exists rather than freshly minted from a client upload, so it can propagate an existing bad key but cannot create a new one.

---

## Also in this round

**The probe is now actually bounded.** The 2 s `AbortSignal` bounds each network *attempt*; the SDK's between-attempt sleep is not abort-aware, so against a degraded store one probe could run for the budget plus a full backoff — on a user-facing mutation that `comics.router` calls in loops of 20 and 10.

Not fixed by adding `maxAttempts` to `getB2ImageS3Client`: that factory is shared with the live upload-completion and abort endpoints, the announcement media check and the server-side upload/delete paths, all of which want their retries. It is byte-identical — the `s3-utils.ts` diff is a single 67-line insertion with zero deleted lines. The probe got its own `maxAttempts: 1` client instead, and `probeCreatedImageMedia` additionally races itself against a 3 s deadline covering both the backend resolution and the head call, so the bound holds without trusting the SDK. Worst case per call is 3 s.

**A test comment overstated the fix's reach.** It claimed `ImageUpload` renders the hook's `error` status. It does not — it reads only `progress`; the `status === 'error'` there is `CustomFile.status`, the post-upload ingestion status, a different type over a different value set. So on the `ImageUpload` path a refused PUT is still silent to the user after this PR, and the comment now says that plainly so nobody verifying there concludes the fix is broken. The `ChallengeSubmitModal` half is verified end to end and stands.

**A stub was wrong about the browser, but its predicted consequence did not reproduce.** The hook test's `cancel()` set `readyState = 0` where the sibling stub correctly sets `4`. Fixed. The prediction was that this made the cancel case pass for the wrong reason — measured, it does not: deleting the `aborted` guard turns that case red under *both* values, identically (`expected [ 'error' ] to deeply equal [ 'aborted' ]`, 1 failed | 8 passed). It is a fidelity fix, not a discrimination fix, and the comment now says which.

**Recorded, not changed:** nothing in production calls `TrackedFile.abort()` — enumerated every non-test file that calls `useCFImageUpload` and grepped each for `abort`; the only hit cancels a generation, not an upload. The guard stays (it is reachable by the module's contract and is what the cancel test exercises), with one line noting its trigger is currently unwired so it is not over-trusted.

Each new or changed guard was watched to fail: the deadline race (2 failed, `Test timed out in 60000ms`), the deadline value in both directions (`expected 'unknown' to be 'present'` when shrunk; `expected 30000 to be 3000` when raised), `maxAttempts` on the probe client (`expected 3 to be 1`), and the seam that pins which backend production resolves (`expected 'shared-backend-bucket' to be 'probe-backend-bucket'`).

ZacxDev:
### Correction to the correction commit

The commit I just pushed (`3696978`) contains a miscount, in the paragraph asserting that its own sweeps were complete. Correcting it here because it cannot be amended without a force-push.

**It says** the repo-wide sweep for the retracted figure "returned nine hits, all unrelated subjects, plus the labelled retraction note" — i.e. ten lines. **It returned nine lines total: eight unrelated, plus the retraction note.**

```
$ git grep -nE '(14 of the|32 direct|of the 32|42 direct)' | wc -l
9
```

The eight are `Dockerfile:63`, `apps/creator-studio/src/lib/server/analytics.ts:338`, `docs/moderator-app/retool-parity-checklist.md:725`, `eslint-local-rules.js:1062`, `packages/civitai-db-schema/src/schema-drift/README.md:247`, `src/instrumentation.node.ts:137`, `src/server/services/__tests__/no-hand-typed-redis-key-constants.test.ts:94`, `test/geometry-setup.tsx:45` — all about unrelated subjects. The ninth is `src/hooks/__tests__/useCFImageUpload.test.ts:213`, the note that records the figure as retracted.

The conclusion is unchanged (the figure appears nowhere as a live claim); the arithmetic reporting it was wrong.

### One scope note on the six-path enumeration

The comment scopes its population to `src/`, which is deliberate but worth closing out. Extending the same two searches to `packages/` and `apps/` finds exactly two more writes to the `Image` table, neither a runtime path: `packages/civitai-db-schema/prisma/seed.ts` (two `prisma.image.create` calls in the dev seed) and a 2023 migration's one-off `INSERT INTO "Image"`. So the `src/`-scoped six is the complete set of production write paths at this ref.

ZacxDev:
### `Unit tests (2)` is red on this head, and it is red on `main` too

Not from this branch. Recording the control rather than asserting it.

The failure is `src/server/services/__tests__/no-direct-shared-module-mock.test.ts` → *"adds no new direct mock of a module that has a canonical mock"*, naming exactly one file:

```
+   "src/server/routers/__tests__/article.router.scan-status-authz.test.ts",
```

That file **does not exist on this branch** — `git cat-file -e HEAD:<path>` fails, and `git log origin/main -- <path>` shows it arriving on `main` in `1d8ad2378` ("fix(articles): scope getScanStatus to the article's author"). It does `vi.mock('~/server/db/client', …)`, a CANONICAL specifier, and it is not in `src/__tests__/mocks/direct-mock-allowlist.json`. That allowlist is byte-identical between `main` and this branch (`git diff origin/main HEAD -- <allowlist>` is empty), so nothing here could have caused it.

Actions runs the merge ref, which is why this branch's rollup shows it.

The discriminating control, run rather than reasoned about — a clean worktree of `origin/main` at `02ad4701a`, no changes from this branch present:

```
FAIL src/server/services/__tests__/no-direct-shared-module-mock.test.ts
  > adds no new direct mock of a module that has a canonical mock
AssertionError: ... expected [ Array(1) ] to deeply equal []
+   "src/server/routers/__tests__/article.router.scan-status-authz.test.ts",
Test Files  1 failed (1)   Tests  1 failed | 2 passed (3)
```

Identical failure, identical single file. On this branch's tree alone the same guard is green (3 passed).

I have not fixed it here: it is someone else's file, the fix is either an allowlist entry or a migration to the canonical mock, and folding an unrelated `main` repair into a correction round is how the next round gets a finding. It does need fixing on `main` — a permanently-red shard trains everyone to click through.

The rest of the rollup is green, including the four `eventloop-watchdog.capture.test.ts` cases that flake locally (all 9 passed in CI on this head).

ZacxDev:
### One imprecision in my previous comment, and the settled rollup

**Imprecision.** I wrote *"the four `eventloop-watchdog.capture.test.ts` cases that flake locally"*. Four is the number of cases in that file's `starvation classifier` block; it is not the number I watched flake. Three distinct cases failed across my local runs — `exposes the ratio distribution and the threshold echo` in one full-suite run, and `classifies an idle wedge as starved, and skips the capture` plus `classifies a wedge as starved even while OTHER threads burn CPU` in another. The point stands (nondeterministic, different cases each run, all 9 green in CI on this head); the count attached to it was loose.

**Rollup** on `3696978`, two spaced reads — `19:51:25Z` and `19:55:01Z` UTC, identical:

| | |
|---|---|
| checks | 20, every one terminal, both reads |
| failing | 1 — `Unit tests (2)`, shown above to fail identically on a clean `origin/main` worktree |
| everything else | SUCCESS, plus one deliberate SKIPPED (`Typecheck (main pushes / fork PRs / non-main base)`) |

Local, on this branch's tree: typecheck 0 errors; unit suite 38,965 passed / 33 skipped / 2 failed, both in `eventloop-watchdog.capture.test.ts`, which this branch does not touch.

PR #4685, opened 2026-09-07
feat(apps): consolidate the build funnel into one state-aware /apps/build

ZacxDev:
## What

One subnav item **Build** → one route **`/apps/build`**, replacing `Build apps` (`/apps/get-started`), `Create` (`/apps/submit`) and `My apps` (`/apps/mine`).

Those three were largely one another's content. `/apps/get-started` was 100% static marketing with one `useDisclosure`, zero network, and every CTA pointing off-platform (GitHub, two npm pages) — no in-product next step at all. `/apps/submit`'s on-platform branch (`CliSubmitCta`) was the same `cliCommands.ts` copy-paste wall a second time: no form, no upload, no mutation. Only `/apps/mine` did real work. A developer's path through the funnel was three tabs that mostly showed each other.

`/apps/build` is state-aware instead:

| State | Condition | Content |
|---|---|---|
| **A · pitch** | not `isAuthor` | `GetStartedBody` + an access CTA. The only public-facing state. |
| **B · first-app** | author, no apps | short pitch + CLI quickstart + `[Create your first app →]` |
| **C · workbench** | author && (`hasEditableApps` ‖ `hasSubmissions`) | `MyAppsBody` + `[+ New app]`; pitch demoted to a collapsed resources strip |

No `blocks.*` / `appListings.*` server changes. `GetStartedBody` and `MyAppsBody` are reused as-is.

## The gate — one predicate, two callers

`canAccessAppsBuild(user, features)` in `~/shared/utils/app-blocks-access`:

```
hasAppsStoreAccess(features) && (isAppDeveloper(user, { appBlocksAuthor }) || appBlocksGetStarted)
```

It is called by **both** the `Build` row in `SUB_NAV_LINKS` **and** `resolveBuildPageAccess`, which the page's `getServerSideProps` calls. Nothing restates it.

This defect class has shipped twice — a tab offered to a cohort whose page answers `notFound` (#3899 for `Create`; again as a deploy-blocking finding on #4668) — and both times the tab's rule and the page's rule were written separately and drifted. One function cannot disagree with itself.

`appBlocksGetStarted` keeps its kill-switch meaning exactly: it decides whether a **store-visible non-author** is shown the pitch. Flip `app-blocks-get-started` off and the pitch (and its tab) disappear for non-authors, no deploy. It cannot switch an author out of their own workbench, which is correct — the switch is on the funnel, not on authoring.

### Requested finding: are the flags `toggleable`?

[...]

PR #4694, opened 2026-09-08
fix(apps): stop /apps/build flashing the "first app" screen at authors who have apps

ZacxDev:
Closes clawgate #530. Closes audit finding **F6** from #4685 (`AppsBuildBody` had no rendered test in any tier, and `viewed.current` was untested).

## The bug

`blocks.getNavSummary` decides state B (`first-app`) vs state C (`workbench`), and it is client-only (tRPC runs `ssr: false`). `resolveAppsBuildState` returns `first-app` whenever both of its summary booleans are false — which is exactly what an unresolved query looks like. So **every author who has apps rendered "Ship your first app"** on the server render and on the first client paint, then swapped to their workbench. The wrong screen, briefly, on every visit.

The analytics could not see it — the `view` event fires only when `settled` — which is why a four-round and a two-round audit ladder both closed clean over it. Nothing that was being measured moved.

## The fix

While the summary query is still pending, render a neutral `AppsBuildBodySkeleton` instead of guessing a state. The **same `settled` boolean** gates the render and the `view` event, so the two cannot disagree about which screen a viewer saw.

The skeleton is deliberately **neutral, not a workbench mock-up**: the destination is unknown by construction while it renders, so committing to C's shape would be the same defect one screen over. It does **not** claim zero layout shift, and says so in the file — contrast `AppListingCardSkeleton`, which knows exactly what it is reserving for.

## 🔴 One behavioural correction rides along, and it is what stops the fix hanging

The old predicate was `settled = !isAuthor || (isClient && isFetched)`. `isFetched` never goes true for a query that never ran — and the query's `enabled` requires `appBlocks`, while the **page** gate is `hasAppsStoreAccess` (`appListings || appBlocks || appListingsPublicExternal`). So an author holding `appListings` alone reaches this page with the summary query switched **off**, and that predicate is `false` for them forever.

Consequences: today they post **no `view` event at all** — the same silently-truncated denominator the existing comment already rejects for the error cohort — and under a wait keyed on `isFetched` they would have sat under the skeleton **permanently**. `settled` now keys on whether there is anything to wait for, so they settle on the first paint, exactly like a non-author.

> 🔴 **Corrected 2026-09-08 (audit round 4).** This paragraph used to end: *"The all-false summary is that viewer's correct and final answer (the `enabled` comment says so)."* **Both halves are false and the code now says so.** `appListings.listMine` is `appDeveloperProcedure` → `hasAppBlocksAuthor`, gated on `appBlocksAuthor` rather than `appBlocks`, and everyone in this cohort holds `appBlocksAuthor` by construction — so their workbench *would* have rendered. The summary is **FINAL** (no further answer is coming) but **not CORRECT**: that cohort can own apps and is shown "Ship your first app" permanently. That is **pre-existing and unchanged by this PR**, and it is disclosed at the `enabled` gate in `AppsBuildBody.tsx`. Left in place rather than edited away because GitHub seeds the squash-merge message from this body, and a merged history asserting the cohort is correctly served is exactly how the follow-up never gets opened.

## Disclosed trade

An author who genuinely has nothing now sees a brief skeleton before state B, where before they got state B immediately. Before the query lands the two states are indistinguishable by construction, so this is not avoidable — and showing established authors the wrong screen on every visit is the worse half of the trade.

## Test matrix — red at base, green at HEAD

New rendered suite `AppsBuildBody.browser.test.tsx` (12 tests) — the first rendered coverage this component has had in any tier — plus `AppsBuildBodySkeleton.ssr.browser.test.tsx` (4).

> ⚠️ **Corrected 2026-09-08.** This said `(3)`, and `+3 SSR = 15` below. It was true when written and was falsified by **this PR's own audit round 1**, which added a fourth SSR test. Same for the `222`-file figure below. See the round-3 comment — a claim true at one commit and falsified by a later commit of the same PR sits inside no audit round's range, which is exactly how all three survived two rounds of re-derivation.

| tree | result |
|---|---|
| `552202a9f9` (`origin/main`), new spec applied to base source | **4 failed \| 8 passed** |
| this branch | **12 passed** (+4 SSR = 16) |

The four that are red at base are the skeleton assertions and the mount→settle transition. The rest are controls and unchanged-path guards, and they pass on both sides on purpose.

### Mutation-checked — each mutant killed by its own assertion

| mutant | killed by | failure |
|---|---|---|
| `settled` reverted to the naive `!isAuthor \|\| (isClient && isFetched)` | *"author whose summary query is DISABLED settles immediately"* — **only** that test | `Cannot find element with locator: getByTestId('apps-build-first-app')` (stuck under the skeleton) |
| the `settled` term dropped from the **analytics** gate alone | both view-event tests | `[event] != []`, and the recorded state is `first-app` rather than `workbench` |
| `component="span"` dropped from the skeleton's bar | *"no `<div>` descends from a `<p>`"* | `expected 5 to be +0` — one per `TextLineSkeleton` |

### 🔴 Two defects in my own test, both found only by running it at the base commit

Recorded because the second is a trap for anyone writing a "was this ever painted" assertion in this codebase:

1. `takeRecords()` was being **drained** (`.forEach(() => undefined)`) rather than processed — the callback is a microtask, so records still queued when the awaited assertion resolved were discarded unread.
2. **Every branch of `AppsBuildBody` returns a Mantine `Stack`**, so React reconciles the same `<div>` **in place** and state B is never an *added node*. A `childList`-only MutationObserver reports a clean zero over a first paint that really did render state B. The measured trace is a single attribute record: `attr data-testid old=apps-build-first-app new=null`.

Both versions passed against base code that paints state B on every author's first render. The watcher now observes `attributes` with `attributeOldValue`, and its positive control drives a real B→workbench swap so it exercises the same branch the absence assertion depends on.

## Verification scope

- `--project component src/components/Apps/AppsBuildBody.browser.test.tsx src/components/Apps/AppsBuildBodySkeleton.ssr.browser.test.tsx` — **16 passed**, ~2s (warm).
- `--project unit src/components/Apps/__tests__ src/shared/utils` — **2004 passed / 95 files**. No call-site ledger broke.
- `pnpm run test:component` (full tier — **223** `.browser.test.tsx` files at HEAD, 221 at base; this said `222`, which was wrong at both ends of the sentence that names them) — my file **passed** (12 tests, 612ms). The run also reported 12 files / 22 tests failing elsewhere; a warm re-run of exactly those 12 gave **2 files / 2 tests**, and a third pass gave **0**. That is drift, not convergence — the cold-cascade signature — and none of the 12 references any module this PR touches (checked with a positive control on the grep).
- `eslint` on all five changed files — clean.
- `pnpm run typecheck` — **0 errors under `src/`**. 31 remain in `packages/*`, all `Cannot find module 'kysely' / '@opentelemetry/context-async-hooks'`: this throwaway worktree has no per-package `node_modules`, which the base clone does. A worktree-setup artifact.

### NOT verified

- **Not exercised in a real browser against production.** No click-path repro of the original flash; the evidence is the red-at-base spec, not a live observation.
- **The back-navigation question in #530 is unanswered, not answered.** I did not measure whether a client-side return to `/apps/build` flashes. The reason to expect it does not is that `isFetched` is query state rather than observer state, so a cached query settles a new observer immediately — that is derived from React Query's state model, **not measured**. `placeholderData` deliberately not added; if a flash is seen in practice that is the follow-up.
- **No screen-reader verification** of the skeleton's live region. The claim is only that the markup *can* announce (a `role="status"` region, not marked busy, with non-hidden text).
- The full component tier was **not** measured at base, so the 12-file failure above is reported as *not attributable to this change*, not as *pre-existing*.

https://claude.ai/code/session_01LvPvg8htuqabcpLXy2m9hc

### DIFF
--- a/src/components/Apps/AppsBuildBody.browser.test.tsx
+++ b/src/components/Apps/AppsBuildBody.browser.test.tsx
@@ -1,0 +1,410 @@
+import { useEffect, useReducer } from 'react';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { page } from 'vitest/browser';
+// `test/` lives outside `src`, so the `~` alias doesn't reach it — relative import.
+import { renderWithProviders } from '../../../test/component-setup';
+// Type-only namespace import (NOT `typeof import('...')`, which
+// @typescript-eslint/consistent-type-imports rejects) so the spread below keeps the
+// real module's type.
+import type * as TrpcMod from '~/utils/trpc';
+
+/**
+ * `AppsBuildBody` — THE UNSETTLED WINDOW. (clawgate #530; closes audit finding F6 from
+ * civitai#4685, which recorded that this component had no rendered test in any tier.)
+ *
+ * 🔴 THE BUG THIS FILE IS THE REGRESSION TEST FOR. `blocks.getNavSummary` is client-only
+ * (tRPC runs `ssr: false`), and `resolveAppsBuildState` reads its two all-false booleans as
+ * `first-app`. So on `main` before this change EVERY author who has apps rendered state B —
+ * "Ship your first app" — on the server and on the first client paint, and only swapped to
+ * their workbench once the query landed. The wrong screen, briefly, on every visit.
+ *
+ * The analytics could not see it (`view` fires only when `settled`), which is exactly why
+ * it survived a four-round and a two-round audit ladder: nothing that was measured moved.
+ * It takes a RENDERED test, which is what this is.
+ *
+ * ── WHY THE MOCKS ARE SHAPED LIKE THIS ──────────────────────────────────────────
+ * `mocks.isFetched` and `mocks.navSummary` are the query's two observable outputs, driven
+ * independently on purpose: `isFetched` goes true on success OR error while `data` stays
+ * `undefined` forever on an error, and the component keys its wait on the FORMER. Driving
+ * them as one boolean would make the error cohort — the population most worth seeing —
+ * untestable, and it is a cohort this component has already been corrected about once.
+ *
+ * 🔴 THE `getNavSummary` MOCK HONOURS `enabled`, AND THAT IS LOAD-BEARING RATHER THAN
+ * FIDELITY FOR ITS OWN SAKE. A disabled query never runs, so React Query never sets
+ * `isFetched` on it. A mock that ignored `enabled` would return whatever `mocks.isFetched`
+ * says and the permanent-skeleton case below would be structurally unobservable — it would
+ * pass while production hung. (The sibling `AppsSubNav.hydration.browser.test.tsx`
+ * deliberately IGNORES `enabled` because its subject is the `useIsClient` deferral; the
+ * choice is per-suite, not a house style.)
+ */
+
+/** A summary for an author who HAS apps — the cohort the bug hit. */
+const SUMMARY_WITH_APPS = {
+  hasInstalls: false,
+  hasSubmissions: false,
+  hasApprovedApps: false,
+  isReviewer: false,
+  hasEditableApps: true,
+  hasPendingInvites: false,
+};
+
+/** The all-false summary — an author with genuinely nothing yet (real state B). */
+const EMPTY_SUMMARY = { ...SUMMARY_WITH_APPS, hasEditableApps: false };
+
+type TrackedAction = { type: string; details: { action: string; state: string } };
+
+const mocks = vi.hoisted(() => ({
+  isClient: true,
+  isFetched: false,
+  navSummary: undefined as undefined | Record<string, boolean>,
+  flags: { appBlocks: true, appBlocksAuthor: true } as Record<string, boolean>,
+  user: { id: 7, username: 'author', isModerator: false } as null | {
+    id: number;
+    username: string;
+    isModerator?: boolean;
+  },
+  tracked: [] as TrackedAction[],
+  /** What `enabled:` the component actually passed — asserted, not assumed. */
+  lastEnabled: undefined as undefined | boolean,
+}));
+
+vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => mocks.isClient }));
+vi.mock('~/providers/FeatureFlagsProvider', () => ({ useFeatureFlags: () => mocks.flags }));
+vi.mock('~/hooks/useCurrentUser', () => ({ useCurrentUser: () => mocks.user }));
+
+vi.mock('~/components/TrackView/track.utils', () => ({
+  useTrackEvent: () => ({
+    trackAction: (payload: TrackedAction) => {
+      mocks.tracked.push(payload);
+      return Promise.resolve();
+    },
+  }),
+}));
+
+// Spread the REAL module and override only `trpc` (local-rules/no-wholesale-module-mock):
+// a hand-written replacement silently breaks every importer the day '~/utils/trpc' grows an
+// export this factory omits — the whole FILE then fails to load with 0 tests collected and
+// no failing assertion.
+vi.mock('~/utils/trpc', async (importOriginal) => ({
+  ...(await importOriginal<typeof TrpcMod>()),
+  trpc: {
+    blocks: {
+      getNavSummary: {
+        useQuery: (_input: unknown, opts?: { enabled?: boolean }) => {
+          mocks.lastEnabled = opts?.enabled;
+          // A query that never ran never becomes `isFetched`. See the header.
+          if (opts?.enabled === false) return { data: undefined, isFetched: false };
+          return { data: mocks.navSummary, isFetched: mocks.isFetched };
+        },
+      },
+      withdrawPublishRequest: { useMutation: () => ({ mutate: () => undefined, isPending: false }) },
+    },
+    // `MyAppsBody` — mounted by the workbench state. Held in its loading branch; this
+    // suite is about WHICH screen renders, not about the table's contents.
+    appListings: {
+      listMine: { useQuery: () => ({ data: undefined, isLoading: true, error: null }) },
+      listMyOrphanedSubmissions: {
+        useQuery: () => ({ data: undefined, isLoading: true, error: null }),
+      },
+    },
+    useUtils: () => ({
+      appListings: {
+        listMine: { invalidate: () => undefined },
+        listMyOrphanedSubmissions: { invalidate: () => undefined },
+      },
+    }),
+  },
+}));
+
+const { AppsBuildBody } = await import('./AppsBuildBody');
+
+const SKELETON = 'apps-build-skeleton';
+const FIRST_APP = 'apps-build-first-app';
+const PITCH = 'apps-build-pitch';
+const WORKBENCH_CTA = 'apps-build-new-app';
+
+/**
+ * 🔴 RENDER BARRIER — required before every "renders nothing" assertion. `render()` commits
+ * through a React 18 concurrent root on a LATER task, so a synchronous absence assertion
+ * right after `renderWithProviders` reads an EMPTY container and passes whatever the
+ * component does. Render the sentinel alongside and await it first. (Same reasoning, and
+ * the same mutation-found hole, as `AppsSubNav.hydration.browser.test.tsx`.)
+ */
+const RENDER_BARRIER = 'render-barrier';
+const RenderBarrier = () => <div data-testid={RENDER_BARRIER} />;
+
+async function renderBody(node = <AppsBuildBody />) {
+  renderWithProviders(
+    <>
+      <RenderBarrier />
+      {node}
+    </>
+  );
+  await expect.element(page.getByTestId(RENDER_BARRIER)).toBeInTheDocument();
+}
+
+const seen = (testId: string) => page.getByTestId(testId).elements().length;
+
+beforeEach(() => {
+  mocks.isClient = true;
+  mocks.isFetched = false;
+  mocks.navSummary = undefined;
+  mocks.flags = { appBlocks: true, appBlocksAuthor: true };
+  mocks.user = { id: 7, username: 'author', isModerator: false };
+  mocks.tracked = [];
+  mocks.lastEnabled = undefined;
+});
+
+describe('AppsBuildBody — the unsettled window renders a skeleton, never state B', () => {
+  test('🔴 author, summary NOT yet fetched → skeleton, and NO "first app" content', async () => {
+    // The exact production shape: an author whose `getNavSummary` has not landed. On `main`
+    // this rendered state B; this is the assertion that is RED at the base commit.
+    await renderBody();
+
+    await expect.element(page.getByTestId(SKELETON)).toBeInTheDocument();
+    expect(seen(FIRST_APP), 'state-B content rendered during the unsettled window').toBe(0);
+    // Belt and braces on the COPY as well as the testid: a refactor that moved the
+    // `data-testid` off that block would otherwise silently defeat the line above.
+    expect(page.getByText('Ship your first app').elements()).toHaveLength(0);
+    expect(seen(WORKBENCH_CTA)).toBe(0);
+  });
+
+  test('the SERVER render (isClient=false) is the skeleton too, so first paint matches it', async () => {
+    // `useIsClient()` is false on the server AND on the first client paint. Both must
+    // produce the same output or hydration bails — the #418/#425 incident this page's
+    // deferral exists to avoid. Asserting the pre-mount render pins that the fix did not
+    // introduce a server/client split of its own.
+    mocks.isClient = false;
+    mocks.isFetched = true; // even with the query "landed", pre-mount must not use it
+    mocks.navSummary = { ...SUMMARY_WITH_APPS };
+    await renderBody();
+
+    await expect.element(page.getByTestId(SKELETON)).toBeInTheDocument();
+    expect(seen(FIRST_APP)).toBe(0);
+    expect(seen(WORKBENCH_CTA)).toBe(0);
+  });
+
+  test('settled WITH apps → the workbench, and no skeleton left behind', async () => {
+    mocks.isFetched = true;
+    mocks.navSummary = { ...SUMMARY_WITH_APPS };
+    await renderBody();
+
+    await expect.element(page.getByTestId(WORKBENCH_CTA)).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+    expect(seen(FIRST_APP)).toBe(0);
+  });
+
+  test('settled with NOTHING → state B really does render (the skeleton is not a permanent stand-in)', async () => {
+    // The positive control for every `seen(FIRST_APP)).toBe(0)` above: the same reader DOES
+    // observe state B when it is the settled answer. Without this the absences could be
+    // satisfied by a component that renders state B never, or by a query wired to nothing.
+    mocks.isFetched = true;
+    mocks.navSummary = { ...EMPTY_SUMMARY };
+    await renderBody();
+
+    await expect.element(page.getByTestId(FIRST_APP)).toBeInTheDocument();
+    await expect.element(page.getByText('Ship your first app')).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+  });
+
+  test('settled by an ERROR (isFetched true, data undefined) → state B, not a stuck skeleton', async () => {
+    // `isFetched`, not `data !== undefined`, is the wait's key — deliberately, so an author
+    // whose summary call FAILS still settles. Keyed on the data they would sit under the
+    // skeleton forever, and they are the cohort most worth not stranding.
+    mocks.isFetched = true;
+    mocks.navSummary = undefined;
+    await renderBody();
+
+    await expect.element(page.getByTestId(FIRST_APP)).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+  });
+});
+
+describe('AppsBuildBody — states that must NOT gain a loading frame', () => {
+  test('a NON-author gets the pitch immediately, with no skeleton, even pre-mount', async () => {
+    // State A is the only PUBLIC-facing and deliberately-indexable state. Putting a loading
+    // frame in front of it would put a skeleton in the SSR HTML a crawler reads.
+    mocks.flags = { appBlocks: true, appBlocksAuthor: false };
+    mocks.isClient = false;
+    mocks.isFetched = false;
+    await renderBody();
+
+    await expect.element(page.getByTestId(PITCH)).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+  });
+
+  test('logged out → the pitch, no skeleton', async () => {
+    mocks.user = null;
+    await renderBody();
+
+    await expect.element(page.getByTestId(PITCH)).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+  });
+
+  /**
+   * 🔴 THE PERMANENT-SKELETON HAZARD, PINNED. The PAGE gate is `hasAppsStoreAccess`
+   * (`appListings || appBlocks || appListingsPublicExternal`) while the summary query's
+   * `enabled` requires `appBlocks` — so an author holding `appListings` alone reaches this
+   * page with the query switched OFF. `isFetched` is then false forever, and a wait keyed on
+   * `isAuthor && !(isClient && isFetched)` would render the skeleton until they navigated
+   * away. The fix keys the wait on `enabled` instead; the all-false summary is that viewer's
+   * correct and final answer, so they settle on the first paint.
+   */
+  test('🔴 author whose summary query is DISABLED settles immediately — no permanent skeleton', async () => {
+    mocks.flags = { appBlocks: false, appListings: true, appBlocksAuthor: true };
+    mocks.isFetched = false;
+    mocks.navSummary = undefined;
+    await renderBody();
+
+    // The wiring this case depends on, asserted rather than assumed: if the component ever
+    // stopped disabling the query for this viewer, the test below would be about a different
+    // situation than the one its name claims.
+    expect(mocks.lastEnabled, 'the summary query should be disabled for this viewer').toBe(false);
+    await expect.element(page.getByTestId(FIRST_APP)).toBeInTheDocument();
+    expect(seen(SKELETON)).toBe(0);
+  });
+});
+
+/**
+ * The `view` analytics event, which #530 fenced explicitly: it must still fire exactly once
+ * per mount and only when settled. These drive the REAL `useEffect` + `viewed` ref — the
+ * thing F6 recorded as untested — through a real mount→settle transition.
+ */
+describe('AppsBuildBody — the view event is unchanged by the skeleton', () => {
+  /** Flips the query to "landed with apps" AFTER mount, i.e. a real settle. */
+  function SettleAfterMount() {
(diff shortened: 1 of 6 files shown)


ZacxDev:
## 🔴 Correction to this PR's description — a sentence in it is false

A blind adversarial audit (round 1) found that the description's justification for the `settled` change is **wrong**, and I am correcting it here rather than silently editing the body, because anyone who has already read it read the wrong version.

**The body says** the all-false summary is *"that viewer's correct and final answer (the `enabled` comment says so)"*, citing the pre-existing comment's claim that *"the workbench they would resolve to reads `appListings.listMine`, and its Withdraw action carries `enforceAppBlocksFlag` too."*

**The `listMine` half of that is false.** `appListings.listMine` is `appDeveloperProcedure` (`src/server/routers/app-listings.router.ts:1082`), and `appDeveloperProcedure = protectedProcedure.use(hasAppBlocksAuthor)` (`src/server/trpc.ts:499-508`) — gated on **`appBlocksAuthor`**, not `appBlocks`. Every viewer in the cohort under discussion holds `appBlocksAuthor` by construction, because that is what made `isAuthor` true. So `listMine` **succeeds** for them and their workbench would have rendered. Only `withdrawPublishRequest` would refuse.

**What is actually true:** all-false is the only answer that *query* can give them — `getNavSummary` is `appBlocks`-gated and `enforceAppBlocksFlag` returns an empty result rather than throwing for a query — but it is **not a true description of their account**. A viewer holding `app-listings` and not `app-blocks-enabled` reaches `/apps/build` (the page gate is `hasAppsStoreAccess`, a three-flag OR) and is shown "Ship your first app" **permanently**, even when they own apps.

**That is pre-existing and this PR does not change that render** — before and after, the cohort resolves to `first-app` on the first paint and stays there. So it is not a regression, and the PR's fix is unaffected. But the PR's *headline* ("stop `/apps/build` flashing the first-app screen at authors who have apps") does not hold for this one cohort, and the reason I gave for that being acceptable was not a reason. `settled` for them is **FINAL** (no further answer is coming) but not **CORRECT**.

Both comments now say so, and the retraction is recorded in the file so it is not re-derived. Fixing the underlying gap means widening what the summary can answer — a server change to a `blocks.*` procedure — which this PR deliberately does not touch.

## The other five findings

| # | severity | what | disposition |
|---|---|---|---|
| 1 | 🟡 `scaffolding` | `__tests__/appsBuildState.test.ts` still taught the retracted belief — in a block title, a docstring and a test name — and it is in the **gated** tier while the corrected prose was in the ungated one | **fixed** |
| 2 | 🟡 `payload` | the false justification above | **fixed** (retracted in place) |
| 3 | 🟡 `payload` | for an author whose query IS enabled, nothing bounds the wait: a client that never mounts, or a hang with no rejection, now renders the skeleton indefinitely where the old code rendered state B. No timeout to lean on — the tRPC links set no `AbortSignal`, and `queryRetry` fires only on a rejection | **disclosed, not fixed** — needs an already-broken client, and a deadline would put a second predicate in front of the render, which is the split the design exists to prevent. Written down at the gate |
| 4 | 🟡 `scaffolding` | both specs are `component`-tier, which `.github/workflows/lint.yml:789` states is **ungated** — so the guard for this bug could not fail a future PR | **fixed**: `resolveAppsBuildSettled` extracted to `appsBuildState.ts`, eight-row truth table pinned in the **blocking** `unit` tier |
| 5 | 🟢 `payload` | the skeleton claimed it "must not look like either B or C" four paragraphs above a header that deliberately mirrors C's geometry | **fixed** — the absolute was wrong, not the header; narrowed to neutrality of *content* |
| 6 | 🟢 `scaffolding` | the `role="status"` / no-`aria-busy` claim was prose only | **fixed**, with the role assertion as the positive control for the absence |

The audit also re-derived every number in the description — the red-at-base matrix, all three mutation results, the typecheck and lint counts — and they hold.

## Verification of this fix round

- **`unit` tier (blocking): 25 passed**, up from 22.
- Browser specs: **16 passed** (12 + 4), twice, warm 1.7s.
- eslint 6 files 0 errors 0 warnings · prettier clean · typecheck **0 errors under `src/`**.
- Two mutants, each killed by its own assertion:
  - `resolveAppsBuildSettled` → the naive `isClient && isFetched`: **3 of the 4 disabled rows fail in the blocking tier**, `expected false to be true`.
  - `role="status"` dropped: only the a11y test fails, `expected '<style…' to contain 'role="status"'`.

**Still not verified**, unchanged from the description: no live browser repro against production, no back-navigation measurement, no screen-reader check, and the full 222-file `component` tier was measured at neither base nor HEAD.

```audit-claims round=1 audited=7f459711c5..1f3fa7c6
1. The `enabled` comment's `listMine` justification is retracted in place, and the `settled` comment no longer claims the all-false summary is CORRECT for the disabled cohort — only FINAL.
2. `resolveAppsBuildSettled` is extracted into `appsBuildState.ts` and its 8-row truth table is pinned in the BLOCKING `unit` tier; the call site no longer inlines the expression.
3. `__tests__/appsBuildState.test.ts`'s SSR block no longer asserts that an author's server HTML is state B — title, docstring and test name all changed.
4. The never-settling failure mode (permanent skeleton for an enabled author whose query hangs) is disclosed in a comment at the render gate; behaviour unchanged.
5. `AppsBuildBodySkeleton`'s "must not look like either B or C" is narrowed to neutrality of CONTENT, reconciling it with the header row that mirrors C's geometry.
6. The skeleton's `role="status"`-present / `aria-busy`-absent claim is now asserted in the SSR spec.
```

ZacxDev:
## Round 2 — the fix round's own prose was the finding, twice

Round 2 caught round 1 doing the exact thing round 1 was fixing. **No runtime behaviour changes in `40ba2aa139`** — it is a correction of load-bearing prose plus one over-strict assertion.

### 🔴 Correcting my own round-1 comment: nothing in this repo blocks a merge on a check

Round 1's finding was *"the guard could not fail a future PR"*. My fix moved the guard from the `component` tier to the `unit` tier and asserted, in **three** places, that `unit` is "the tier CI blocks on". **That is false**, and the authority I cited for it says so:

- `.github/workflows/lint.yml:405` — `continue-on-error: ${{ github.event_name == 'pull_request' }}` on the `unit` job, with the deliberate *"REPORT-ONLY to start… FLIP TO BLOCKING once a couple of weeks of runs show a clean pass rate"* note at `:357`.
- `main` has branch protection with **no required status checks**. Read live: `gh api repos/civitai/civitai/branches/main/protection` → `"required_status_checks": null`.

So the round-1 fix moved the guard from a report-only preview job to a report-only-on-PR lint job, while claiming the opposite — the same shape as the finding it was answering.

**What the move does buy, stated honestly:** `component` is report-only *everywhere* (its only CI home is `preview / component-tests`), whereas `unit` runs on `push: [main]`, where that `continue-on-error` expression evaluates **false**. A regression in this predicate therefore reds the `main` build **after it lands**, instead of being observed nowhere. **Catch-after-landing, not block-before-merge.** All three comments now say that, each recording what it used to say so nobody re-derives the stronger claim.

### 🔴 And my round-1 correction over-swung into its mirror image

The test docstring went from *"an author's server HTML is state B"* to a bare *"an author's server HTML is the skeleton"*. **Both absolutes are false, in opposite directions.** An author *without* `appBlocks` — admitted to the page by `appListings` — has `summaryEnabled` false, settles immediately, and state B really is their server HTML, permanently. That is the same cohort the round-1 retraction is about; the correction and the docstring contradicted each other in one commit. Now qualified.

### 🟡 Two counts in my round-1 verification section were wrong

| I wrote | actually |
|---|---|
| "`unit` tier (blocking): 25 passed, up from **22**" | **16** at `7f459711c5` — measured by running the base file verbatim. The delta is **+9**, not +3. `22` was the pass count of the *mutant* run (`3 failed \| 22 passed`); I transposed it. |
| "eslint **6** files" | 6 is what I ran; the round-1 range is **5** files. |

The "(blocking)" label in that same line is wrong for the reason above.

### The rest

| # | severity | what | disposition |
|---|---|---|---|
| D | 🟢 | truth-table row 6 carried a *production* justification for an input that cannot occur — `IsClientProvider` flips `isClient` once app-wide, so `isClient === false` implies an empty tRPC cache | **fixed** — relabelled COMBINATORIAL, matching how rows 1–4 were already framed. The row stays as mutation-hardening |
| E | 🟢 | `aria-busy` absence asserted document-wide while the claim is about the `role="status"` container — a Mantine bump emitting it on a child `<Skeleton>` would red a test whose claim still held | **fixed** — scoped to the region's own tag, and re-checked that narrowing did not blind it |
| C | 🟢 | `resolveAppsBuildSettled` has no call-site ledger where its two siblings do | **deliberately not fixed.** Given the finding above, a ledger inherits the same non-enforcement — scaffolding with no guarantee. The instruction stays and now says plainly that nothing enforces it |

Round 2 also re-derived the round-1 retraction from source and confirmed it: `listMine` → `appDeveloperProcedure` → `hasAppBlocksAuthor` reads `appBlocksAuthor` only; `enforceAppBlocksFlag` returns the all-false shape for a query rather than throwing; and the cohort's **render** is unchanged by this PR (on `main`, `settled` was wired only to the `view` effect — there was no render gate). It correctly noted that the cohort's **analytics** do change, which is intended and documented.

### Verification of this round

- `unit` 25 passed (unchanged — this commit adds no test) · browser specs **16 passed, twice**, warm 2.9–3.0s.
- eslint 4 files 0/0 · prettier clean · typecheck **0 errors under `src/`**.
- Mutant: `aria-busy="true"` put back **on the live region** → only the a11y test fails, `expected '<div style="--stack-gap…' not to contain 'aria-busy'`. The scoped assertion still kills the hazard it was narrowed around.

**Still not verified**, unchanged: no live browser repro, no back-navigation measurement, no screen-reader check, and the full 222-file `component` tier measured at neither end. Round 2 additionally flagged that the round-1 disclosure's claim about tRPC links setting no `AbortSignal` is **unchecked** — nothing rides on it (it is a comment with no behaviour attached), but treat it as unverified.

```audit-claims round=2 audited=1f3fa7c64e..40ba2aa139
1. The "BLOCKING `unit` tier" claim is retracted in all three places that made it (appsBuildState.ts, AppsBuildBody.tsx, __tests__/appsBuildState.test.ts) and replaced with catch-after-landing-on-main, citing lint.yml:405 and the live required_status_checks: null.
2. The test docstring's bare "an author's server HTML is the skeleton" is qualified to authors whose summary query is ENABLED; the appBlocks-disabled cohort is named as still rendering state B permanently.
3. Truth-table row 6's justification no longer claims a production hydration hazard; it is labelled COMBINATORIAL, not reachable.
4. The aria-busy absence assertion is scoped to the element carrying role="status" instead of the whole serialized document.
5. No call-site ledger was added for resolveAppsBuildSettled; the comment now states that nothing enforces the do-not-re-inline instruction.
6. The wrong counts (22 -> 16 base tests, 6 -> 5 files) are corrected in this comment; no code depended on them.
```

ZacxDev:
## Round 3 — the sweeps finished, one blinded guard widened, and a stop rationale

Commits `c2ef4a3093` and `d841286ea4`. **Zero executable payload lines changed in either** — mechanically confirmed: stripping comment lines from the range diff of both production files leaves nothing.

### Round 3's meta-finding, which is the one worth keeping

**The analysis was right in rounds 1 and 2; the sweep was not.** Round 1 retracted a false claim at **3 of its 4** sites. Round 2 qualified an over-swung absolute at **1 of its 3**. Both times the missed copy sat in a file the same commit was already editing. That is a sweep failure, not a reasoning failure, and it has a mechanical fix — which is what this round used.

| # | finding | fix |
|---|---|---|
| F1 🟡 | the retracted *"the tier CI blocks on"* survived as the **summary line** of the docstring whose body retracts it (`__tests__/appsBuildState.test.ts:180`), eight lines above the paragraph saying these rows block nothing | reworded to "in the `unit` tier", full stop. A summary line is what a hurried reader takes away, so it was the worst surviving copy |
| F2 🟡 | the unqualified *"an author's first paint is a skeleton"* survived in **two** more places — `AppsBuildBody.tsx:62` (the component's own headline) and the SSR spec's *"on EVERY author's every visit"* | both qualified. The correcting paragraph was 40 lines below the headline, which is the reading order that lets an absolute win |
| F3 🟢→🔴 | **my round-2 narrowing blinded the guard, and my check for that was the wrong check** — see below | assertion widened to region **and** ancestors; both mutants die with their own message |

### 🔴 F3 is the one I'd want a reviewer to read

Round 2 scoped the `aria-busy` absence from the whole document to the `role="status"` tag, then "verified" that the narrowing was safe by **re-running the mutation it kept**. Measured this round: `aria-busy="true"` on an **ancestor** of the live region **survives** the scoped form (4 passed) and dies under the document-wide one. The narrowing's stated justification only ever argued about *descendants*.

> **The reusable rule, now written into the test: when you NARROW a guard, mutate the case you EXCLUDED, not the case you kept.** Re-testing the retained half proves nothing about the half you dropped.

Both mutants now die with their **own** named message, not one killing the other:
- ancestor → `an ANCESTOR of the live region is marked busy, which suppresses it`
- region → `the live region itself is marked busy`

**And then I did the same thing again, and caught it myself** (`d841286ea4`). The comment I wrote justifying the widened assertion asserted that ARIA treats a busy ancestor as suppressing a contained live region. I checked that against neither the spec nor a screen reader — I wrote it to explain a guard I had just widened, which is exactly the mechanism that produced the false claims in rounds 1 and 2. It is retracted for the justification I can stand behind: *the form this replaced did cover the ancestor case, and the narrowing dropped it for a reason that only argued about descendants.* The note tells the next reader to **delete** the assertion if a busy ancestor turns out to be harmless — not to reword it — so nobody is invited to invent a fourth rationale.

### The once-per-ladder sweep (ignores commit ranges, by design)

This is the class no delta round can see: a claim **true when written** and **falsified by a later commit of the same PR** sits inside no round's range. Three of these were live:

| claim | reality | corrected |
|---|---|---|
| body: "(3) SSR tests", "+3 SSR = 15", "15 passed" | **4** and **16** — falsified by *round 1's own commit*, which added the a11y test | PR body, with the correction stated in it rather than made silently |
| body + both comments: "the full **222**-file component tier" | **221** at base, **223** at HEAD — wrong at *both* ends of the sentence naming them | PR body |
| `lint.yml:357` cited for a sentence at `:365` | `:357` anchors "REPORT-ONLY to start"; `:365` is "FLIP TO BLOCKING" | now `:357-365` |

Round 1's comment said *"the audit re-derived every number in the description … and they hold."* **That sentence was made false by round 1's own edit** — which is the cleanest possible illustration of why this sweep has to ignore ranges.

Two comments that had stopped describing their code are also fixed: the "found by the SAME matcher" note (the assertion is a regex now) and a paraphrase that dropped a nesting level from `next({ ctx: { _appBlocksDisabled: true } })`.

**Sweep evidence — mechanical, not asserted.** Across all six PR files:

```
"blocks on|BLOCKING tier"        -> 5 hits, ALL retraction prose
unqualified skeleton absolute    -> 0  (every remaining "EVERY author" describes the
                                        PRE-change state, which is true — disabled
                                        cohort included; verified against origin/main)
"222" / "15 passed" / "+3 SSR"   -> 0 in files; corrected in the PR body
```

### Verification

unit **25 passed** · browser specs **16 passed**, twice, warm 2.3–2.5s · eslint 4 files 0/0 · prettier clean · typecheck **0 errors under `src/`** · both `aria-busy` mutants killed by their own assertions.

### 🔴 Why I am stopping the ladder here — stated, not left implicit

Three rounds, every one of which found something real, so no round was wasted. I am stopping on the **prose-payload criterion** rather than on a clean round, and the honest reason the rounds would not stop on their own is that **rounds 2 and 3 changed zero executable payload lines** — each was auditing the prose the previous round wrote, and a round 4 would audit round 3's prose. The conditions:

- **No 🔴 outstanding.** F3 had real blast radius and is fixed and double-mutation-verified.
- **No blast radius beyond "a document contains a false sentence"** for anything that remains.
- **The recurring shape swept at every site**, mechanically, with the grep counts above — this is the specific generator of rounds 2's and 3's findings, and it is the first round in which it was addressed with a control rather than with more care.

**What I am NOT fixing, so it reads as open rather than absent:**

1. **Every `lint.yml:NNN` citation in these files is unpinned by construction.** Nothing asserts those line numbers and `lint.yml` is ~928 lines and actively edited, so all six rot silently. Left as-is because the alternative — quoting text instead of lines — was not obviously better and is a judgement call for this repo's owners.
2. **The `app-listings`-without-`app-blocks` author cohort still sees state B permanently.** Pre-existing, unchanged by this PR, documented at the `enabled` gate. Fixing it means widening a `blocks.*` procedure — a server change this PR deliberately does not make.
3. **`#418/#425`** (`AppsBuildBody.tsx:59`) resolve to nothing in this repo; they are pre-existing at `552202a9f9` and are presumably tracker numbers written without a prefix.

**Still not verified**, unchanged throughout: no live browser repro against production, no back-navigation measurement, no screen-reader check, and the full `component` tier measured at neither end.

```audit-claims round=3 audited=40ba2aa139..d841286ea4
1. The retracted "tier CI blocks on" is gone from the truth-table docstring's SUMMARY line — the 4th site, which round 1 and round 2 both missed.
2. The unqualified "an author's first paint is a skeleton" is qualified at both surviving sites (AppsBuildBody.tsx headline, SSR spec header).
3. The aria-busy assertion covers the live region AND its ancestors; the ancestor mutation survived the round-2 form and now dies with its own message.
4. The ARIA justification I wrote for that widening is retracted as unverified; the assertion is kept on the narrower ground that the previous form covered the case and the narrowing dropped it.
5. The PR body's SSR/total test counts (3/15 -> 4/16) and the 222-file tier figure (221 base / 223 HEAD) are corrected in the body itself, with the correction visible rather than silent.
6. lint.yml:357 is now cited as :357-365; two comments that had stopped describing their code are fixed.
7. The retraction sweep was done by grep across all six files and the counts are published above — the first round in which completeness is evidenced rather than asserted.
```

ZacxDev:
## Round 4 — two findings, both fixed. **Not a clean round.**

Commit `aa467d2e67`, plus a body edit. Stating the headline plainly because a merge was gated on it: **round 4 did not come back clean.** Four rounds, four rounds with findings.

### 🟡 F1 — the PR body still asserted the claim the code retracts, and cited the retraction as its authority · `scaffolding`

The body read: *"The all-false summary is that viewer's **correct and final answer** (the `enabled` comment says so)."* `AppsBuildBody.tsx:166-171` says the opposite — **FINAL is true, CORRECT is not** — and `:103-110` records that cohort as shown "Ship your first app" permanently even when they own apps. The parenthetical citation was inverted: the `enabled` comment now says the reverse of what the body claimed it said.

**Why this was worth catching before merge:** GitHub seeds the squash-merge message from the body. A merged history asserting that cohort is correctly served is exactly how the follow-up never gets opened. **Corrected in the body, in place, with the retraction left visible rather than edited away.**

This is the ladder's signature shape landing in the one artifact the six-file grep sweep structurally does not cover.

### 🟢 F2 — the new ancestor assertion had no control of its own, and the comment above it claimed otherwise · `scaffolding`

`expect(html).toContain('role="status"')` proves the *document* is non-empty; it constrains nothing about `html.slice(0, region.index)`. If `region.index` were ever `0` that slice is `''` and `not.toContain` passes for free — the precise vacuous-absence hazard the comment four lines up warns about, reintroduced one assertion later. Measured: `region.index` is 882 today and the slice holds MantineProvider's `<style data-mantine-styles>` preamble, so it is **not** vacuous now — but nothing would have said so if Mantine stopped emitting it. Now asserted, and mutation-checked:

- slice forced to `html.slice(0, 0)` → `the ancestor slice is empty — the assertion below is vacuous: expected 0 to be greater than 0`
- ancestor mutant still dies for its own reason → `an ANCESTOR of the live region is marked busy, which suppresses it`

Two comments corrected alongside, because a comment is a claim: *"over-broad by exactly the siblings"* **understated** it (the slice is raw markup, so a preceding sibling's whole subtree is in it), and *"AND EVERY ANCESTOR"* is scoped to ancestors **inside this component** — the spec renders `AppsBuildBodySkeleton` alone, so a busy ancestor contributed by `AppsPageLayout` is out of reach and guarded nowhere.

### 🔴 And a correction to my own round-3 evidence

Round 3 published `"blocks on|BLOCKING tier" -> 5 hits, all retraction prose` as proof the sweep was complete. Round 4 could not reproduce the 5 under any pattern it tried. Reconciled:

| pattern | hits |
|---|---|
| what I actually **ran** (`blocks on\|BLOCKING (.unit.\|tier)\|what actually blocks`) | **5** |
| what I **published** (`blocks on\|BLOCKING tier`) | **4** |

**The evidence was sound; the transcription was not.** I simplified the pattern when writing it up, which made a published count unreproducible from the published command — and unreproducible evidence is not evidence. The *claim* it supported does hold under every pattern: all hits are retraction prose, negations, or accurate quotes of `lint.yml`'s own "FLIP TO BLOCKING" TODO. **No surviving assertion that any tier blocks.** Publish the exact command next time, not a tidied one.

### Round 4's independent re-measurements — all round-3 claims hold

Every round-3 fix verified **actually fixed**, by the auditor's own runs rather than my report: the truth-table summary line, both qualifier sites, the widened `aria-busy` guard (including a **negative control** — deleting only the ancestor assertion makes the ancestor mutant survive with "4 passed", reproducing my figure exactly), the ARIA retraction's honesty, all four corrected body counts (SSR **4**, total **16**, `.browser.test.tsx` **221** at base / **223** at HEAD), and the `lint.yml:357-365` span. Nothing was made worse.

### The ladder stops here, and the gate — not my judgement — is why

The attribution gate's condition is now **met and independently verified**: rounds 2, 3 and 4 each changed **zero executable payload lines** (all 15 executable lines in this PR since round 1 come from round 1's own `resolveAppsBuildSettled` extraction). Two consecutive zero-payload rounds means the ladder has left the PR and is auditing its own scaffolding. Round 4's auditor independently recommended no round 5.

**So `aa467d2e67` is unaudited by anyone**, and I am saying that rather than implying the PR is now clean. It is one test control and two comment corrections, each mutation-checked above.

**What remains open, recorded so it reads as open rather than absent:**

1. The **`app-listings`-without-`app-blocks` author** still sees state B permanently. Pre-existing, unchanged here, disclosed at the `enabled` gate. Closing condition if filed: that cohort reaches the workbench, pinned by a rendered test, verified by a merged PR. Needs a server change to `blocks.getNavSummary`.
2. Every **`lint.yml:NNN` citation** in these files is unpinned by construction and will rot silently.
3. A **page-level busy ancestor** (from `AppsPageLayout`) is guarded nowhere; this spec cannot reach it.
4. **`#418/#425`** (`AppsBuildBody.tsx:59`) resolve to nothing in this repo — pre-existing at `552202a9f9`.

**Still not verified** throughout: no live browser repro against production, no back-navigation measurement, no screen-reader check, whether a busy ancestor actually suppresses a live region (honestly unverified in the code too), and the full `component` tier at either end.

```audit-claims round=4 audited=d841286ea4..aa467d2e67
1. The PR body no longer asserts the all-false summary is the disabled cohort's "correct and final answer"; the retraction is in the body itself, visible, because GitHub seeds the squash message from it.
2. The ancestor aria-busy assertion has its own emptiness control, mutation-checked by forcing the slice to length 0.
3. "over-broad by exactly the siblings" is corrected to include the siblings' subtrees; "every ancestor" is scoped to ancestors inside this component, with the page-level gap named as unguarded.
4. Round 3's published grep count is reconciled: 5 was the pattern actually run, 4 is the pattern published — the claim holds under both, the transcription did not.
5. No code outside tests and comments changed in this round; the attribution gate's two-consecutive-zero-executable-payload condition is met and the ladder is stopped.
```

PR #4695, opened 2026-09-08
feat(upload): count image-upload relay usage, so a successful fallback stops being invisible

ZacxDev:
## Premise, verified before writing anything

The fallback image-upload relay (`/api/v1/image-upload/relay`, #4573) exists to rescue clients whose direct PUT to the storage host fails at the network layer. It shipped with no way to tell whether it has ever helped anyone. Each leg of that was checked against the code that produces the signal, not assumed:

| Claim | How it was checked | Result |
|---|---|---|
| A successful relay is not visible in the request-log stream | The access log is filtered to `4xx/5xx OR >5s` | Confirmed — a fast 200 is never written, so today's only signal is biased **entirely toward the rejection branches** |
| Traces cannot see it | `TraceIdRatioBasedSampler` in `src/instrumentation.node.ts`, default ratio `0.1` (env-overridable) | Confirmed — at this route's event rate, "sampled zero" and "never happened" are the same observation, indefinitely |
| The media-location registry cannot separate relayed from direct | Both paths register the same backend, deliberately (same bucket, same backend — that is the point of the route) | Confirmed |
| **Is there already a counter on this route?** | `git grep -i relay` across `src/` and `packages/`; read `src/server/prom/*` and `src/pages/api/metrics.ts` | **No.** The route calls `instrumentApiResponse`, which counts only `status >= 500` into `civitai_app_http_errors_total` — it can see the store blowing up and nothing else. Not a usage signal, never meant to be one |
| **Is there an app-metrics surface to reuse?** | `src/server/prom/` (`client.ts`, `b2-put.metrics.ts`, `http-errors.ts`, …), `src/pages/api/metrics.ts` | **Yes** — `prom-client` on the default registry, scraped at `/api/metrics`. No new library and no new endpoint were introduced |
| Is someone else already doing this? | `gh pr list --state open` | No overlap. #4692 touches `upload-settlement.ts` / `image.service.ts`, not this route and not a metric |

So the premise held, and the work is a gap rather than a duplicate.

## What this adds

`civitai_image_upload_relay_total{outcome}` — a new module `src/server/prom/image-upload-relay.metrics.ts`, following the existing pattern in that directory (get-or-create against the default registry, fail-soft emitter, code-owned label union).

**One increment per invocation, structurally.** The response logic moved into `runRelay`, which *returns* its outcome; the exported handler is the only thing that emits. That is deliberate rather than cosmetic — "someone remembers to add an emit call beside each of the ten `res.status(...)` sites" is not a mechanism. Because the return type is a closed union with no `undefined` in it, **a new branch that settles the response without naming an outcome is a compile error** (verified below). The wrapper's `catch` records `handler_error` and re-throws the original error unchanged, so the guarantee also covers the unexpected without changing what the framework sees.

**Seeded at 0, from `/api/metrics`.** prom-client materialises a child only on its first `inc()`, and this route almost never runs on a given pod. Without seeding a scrape returns nothing and PromQL answers `no data` — indistinguishable from "the instrument was never wired", which is the exact ambiguity the counter exists to remove. `ensureRegisterImageUploadRelayMetrics()` is called explicitly from `src/pages/api/metrics.ts`, mirroring its two neighbours there. Both halves are required; neither works alone, and there is a dedicated test for that seam because neither the module's own suite nor the route's suite can see it.

### Cardinality bound

**Exactly one label, over a closed union of 11 values → 11 series per pod, fixed.** Nothing caller-influenced can move that number.

`success`, `method_not_allowed`, `forbidden_origin`, `unauthorized`, `busy`, `too_large`, `truncated`, `empty`, `read_error`, `store_error`, `handler_error`.

Deliberately **no** user id, object key, bucket, hostname, path, content type or byte size. `recordImageUploadRelay` narrows at runtime and **drops** an unknown value rather than relabelling it — relabelling to `unknown` would break the fixed-series claim, and relabelling to `success` would invent evidence that the relay rescued an upload. The bound therefore rests on code, not on erased types. Pinned by a test that asserts `labelNames` and the exact series count.

`too_large` and `truncated` are kept apart on purpose: folding them would let a truncation regression — the defect that stored corrupt images and answered 200 — hide inside the size-limit count.

## Mutation matrix

Every mutation was applied in isolation to a pristine copy, run, and reverted. Baseline: **48 passed, 0 failed.**

| # | Mutation | Result | Which test(s) went red |
|---|---|---|---|
| M1 | Delete the increment in the handler | **12 failed** / 36 passed | every per-branch case + `records EXACTLY ONE outcome per invocation` |
| M2 | Hardcode the label: `recordImageUploadRelay('success')` | **11 failed** / 37 passed | every per-branch case **except** `counts a SUCCESSFUL relay`, which the mutant gets right — this is the discriminating shape a naive "did my outcome move?" test would miss |
| M3 | Fold `truncated` into `too_large` | **1 failed** / 47 | `counts a short body as truncated, NOT as too_large` |
| M4 | No-op the seeding loop | **8 failed** / 40 | `SEEDS every outcome at 0…`, the cardinality-bound case, and 🔴 `/api/metrics … exposes ALL eleven outcome series at 0` |
| M5 | Remove the runtime narrowing in the emitter | **1 failed** / 47 | `DROPS an unknown outcome rather than relabelling it` |
| M6 | Emit twice per invocation | **12 failed** / 36 | every per-branch case + the exactly-one seam guard |
| M7 | Remove the `handler_error` emit from the wrapper's catch | **1 failed** / 47 | `counts an unexpected throw as handler_error and RE-THROWS it unchanged` |
| M8 | Remove `ensureRegisterImageUploadRelayMetrics()` from `/api/metrics.ts` | **2 failed** / 46 | the two 🔴 cases in the seam suite — the module half alone does **not** cover this |
| M9 | Delete `return 'success'` from the success branch | **compile error** | `relay.ts(195,4): error TS2366: Function lacks ending return statement…` — this is what makes an uncounted branch impossible rather than merely tested-for |
| M10 | Add a second (`user_id`) label | **1 failed** / 47 | `holds the cardinality bound at exactly one label over a closed union` |
| M11 | Register on a private registry instead of the default | **24 failed** / 24 | including the seam suite — proves the tests read the registry `/api/metrics` actually serves, not an in-memory handle |

Two isolation checks worth stating: M1 killed **only** route-level cases and left the metrics-module suite green (correct — different module), and M8 killed **only** the seam suite (correct — the seam is the endpoint wiring). A mutation that killed everything indiscriminately would tell you nothing about which guard is doing the work.

The suites also carry their own positive controls: the seam suite asserts the scrape body is non-empty and carries unrelated metrics before asserting anything about this one, so a handler wired to nothing cannot pass by returning `""`.

## What I verified, and what I did not

**Verified**

- Cold typecheck on the merged tree (deleted `tsconfig.tsbuildinfo` first — a warm run reports stale incremental errors): **0 errors, 85s**.
- The premise, per the table at the top — including that no counter already exists and that no new metrics surface was needed.
- The mutation matrix above, every row run rather than reasoned about.
- Merged-tree gate: current `origin/main` merged into this branch and both typecheck and the unit tier re-run **there**, not on the branch alone. The lockfile did not move in the merge.
- Prettier and ESLint clean on every touched file. All were already prettier-clean on `main`, so no unrelated reformatting rides along.

**NOT verified**

- **No production observation.** Nothing here proves the counter increments on a real pod, because the event it counts requires a client that cannot resolve the storage host. The route-level tests drive the real emitter and read the real registry, and the seam test scrapes the real `/api/metrics` handler — but that is the unit tier, not production. The first honest production check is that `civitai_image_upload_relay_total` appears in a scrape with all eleven series at 0.
- **The browser/component tier was not run locally** — it cannot run on this host (a Playwright build pin mismatch), so my local claim covers the `unit*` tier only. CI ran it and it passed (`preview / component-tests`); that is CI's evidence, not mine.
- **No dashboard or alert.** Deliberately out of scope. Note for whoever adds one: this is a **rare per-pod** counter, so read it with `sum()` / `max_over_time()`, **not** `rate()` on a single child — a pod that relays once creates its child at 1 and never touches it again, so `rate()` over that child is structurally 0 and a threshold keyed on it would silently never fire. The metric's HELP text says so too.
- **No behaviour change to the relay is claimed or intended.** The response logic is byte-identical apart from each `return;` becoming `return '<outcome>';`; the existing suite for this route still passes unchanged.

## Test counts

Counts read off each run's own summary line, not from an exit code.

- Targeted (the three suites this touches): **48 passed / 48**, 0 failed.
- Full `unit*` tier, **on the merged tree** (`origin/main` @ `38179c58d0` merged in): **1687 test files passed | 3 skipped (1690)**, **38962 tests passed | 33 skipped (38995)**, **0 failed**, 275s.
- Cold typecheck on the same tree: **0 errors**, 85s.

`origin/main` was re-checked immediately before opening this PR and had not moved past the merged commit.

CI on this PR: **19 checks, 18 pass / 1 skipped / 0 failed** — including all four `Unit tests` shards, `App unit tests + typecheck`, `tekton / typecheck`, `ESLint + Prettier (changed files)`, `Geometry tests`, `Package unit tests`, `Schema drift gate`, and `preview / component-tests` (the browser tier this host cannot run).

### DIFF
--- a/src/__tests__/pages/api/v1/image-upload/relay.test.ts
+++ b/src/__tests__/pages/api/v1/image-upload/relay.test.ts
@@ -561,8 +561,8 @@ describe('image-upload relay', () => {
   // Usage counter
   // -------------------------------------------------------------------------
   //
-  // WHY THESE EXIST. A successful relay is INVISIBLE in production: a 200 is filtered
-  // out of the access-log stream (it logs 4xx/5xx or >5s only), traces are head-sampled
+  // WHY THESE EXIST. A successful relay is INVISIBLE in production: our request-log
+  // stream does not retain fast 2xx responses, traces are head-sampled
   // at 0.1 against an event rate of a handful over months, and the media-location
   // registry records the same backend for a relayed and a direct upload. So the route
   // shipped with no way to answer "did this ever help anyone?", and
--- a/src/__tests__/pages/api/v1/image-upload/relay.test.ts
+++ b/src/__tests__/pages/api/v1/image-upload/relay.test.ts
@@ -76,6 +76,17 @@ import handler, {
   __getInFlightForTest,
   __resetInFlightForTest,
 } from '~/pages/api/v1/image-upload/relay';
+// 🔴 DELIBERATELY NOT MOCKED. The usage counter is the only production evidence that
+// this fallback ever works, so the cases below drive the REAL emitter and read the real
+// prom registry. A mock here would assert that the route called a function, which is
+// the weaker claim — it cannot see a wrong label, a folded series, or a registration on
+// a registry nothing scrapes.
+import client from 'prom-client';
+import {
+  IMAGE_UPLOAD_RELAY_METRIC,
+  IMAGE_UPLOAD_RELAY_OUTCOMES,
+  __resetImageUploadRelayMetricsForTest,
+} from '~/server/prom/image-upload-relay.metrics';
 
 /** Records the order in which the handler wrote a status vs destroyed the request. */
 type Trace = string[];
@@ -182,6 +193,9 @@ beforeEach(() => {
   // A case that leaves a slot held would otherwise make the NEXT case fail for an
   // unrelated reason — measured as a false cascading failure during round-2 mutation.
   __resetInFlightForTest();
+  // The counter is module-scoped and process-wide; without this every case inherits
+  // the previous one's counts and the "and only that outcome" assertions go vacuous.
+  __resetImageUploadRelayMetricsForTest();
   prodFlag.value = false;
   mockGetServerAuthSession.mockResolvedValue(authed);
   mockUploadImageBufferToStore.mockResolvedValue({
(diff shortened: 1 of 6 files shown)


ZacxDev:
Applied five audit fixes in `395c63b81a3a75080c71d91267b964112e76bc98`. All comment / HELP-string / test-line edits — **no behaviour change to the route**, re-proved below.

### What changed

**🟡-1 — the documented query could not answer the question the metric exists for** (`image-upload-relay.metrics.ts`, module header + HELP)

The module told readers `sum(civitai_image_upload_relay_total)` was the request count and `…{outcome="success"}` was the rescue count. That is an instant vector over *currently-live pods only* — prom-client counters live in the Node heap and die with the process, and processes here are replaced routinely, so a relay that rescued someone on Monday reads 0 on Tuesday. The metric's whole job is the historical "has this fallback ever helped anyone?", and that read cannot answer it.

Now split into two explicitly-labelled questions: `sum(...)` for "is it being used right now" (with the live-pod-only caveat stated), and `sum(increase(...[30d]))` — or `sum(max_over_time(...[30d]))` — for "has it ever helped", both of which aggregate over child series that existed anywhere in the range including pods long since replaced.

Also corrected the 🔴 ALERTING rationale, which argued *against* `increase()` on the grounds that "a pod that relays once creates its child series at 1 and never touches it again". **The seeding this PR adds invalidates that**: every child is materialised at 0 on the pod's first scrape (seeding + its call site in `src/pages/api/metrics.ts`), so a later relay is a visible `0 -> 1` delta and `increase()` reports it. The note now says so and warns against removing the seeding to "simplify" an alert.

**🟡-2 — `handler_error` was documented as something it is not** (same file, outcome doc + HELP)

It read "a non-zero here is a bug in the route, not a property of the traffic." `runRelay` awaits three things outside its inner `try`, and `getServerAuthSession` is the one with a network dependency: `~/server/auth/get-server-auth-session.ts` fail-softs `getHubSession` (`:92`) and `getLegacySession` (`:102`) with `.catch(() => null)`, but awaits `getSessionFromBearerToken(token)` (`:79`) and `maybeRollHubCookie(...)` (`:97`) **unguarded** — the latter a network write to the auth hub. During an auth-hub incident this counter goes non-zero and the old prose sent the investigator to the wrong file. The doc now names the dependency and says to check its health first.

**🟡-3 — infrastructure disclosure in a public repo** (four sites)

`src/server/prom/image-upload-relay.metrics.ts:8` and its HELP string, `src/pages/api/v1/image-upload/relay.ts` (wrapper docblock), and the usage-counter preamble in `src/__tests__/pages/api/v1/image-upload/relay.test.ts` all stated the ingress access-log filter predicate. That is configuration from a private infrastructure repo, was not in `src/` before this PR, and this repo is public — and it discloses that a fast 2xx on *any* endpoint leaves no access-log record, which is a forensics blind spot rather than a fact about this route. All four now state only the consequence: *"our request-log stream does not retain fast 2xx responses."* The predicate is not paraphrased anywhere.

**🟢-4 — a comment claiming more than the compiler enforces** (`relay.ts`)

It said a newly-added branch that *settles the response* without naming an outcome is a compile error. TypeScript enforces that only for branches that **return**; `if (bad) { res.status(400).json(…); }` with no `return` compiles, double-writes the response, and is counted under whatever outcome control falls through to. Narrowed to "RETURNS without naming an outcome", with the residual hole named explicitly.

**🟢-5 — a test that passed only by ordering** (`src/server/prom/__tests__/image-upload-relay.metrics.test.ts`)

`never throws` read the metric handle off the registry *before* calling `ensureRegisterImageUploadRelayMetrics()`, so it passed only because earlier tests in the file had already registered the counter. Swapped the two lines.

Additionally strengthened it: the old stub threw on *every* `inc`, so it fired inside `seedAllSeries` — the first `inc` — and never reached `imageUploadRelayTotal.inc({ outcome })`. It proved the `try/catch` swallows *a* throw from somewhere in the emit path, which is weaker than the guard's actual claim. The stub now throws only when called without a second argument (seeding always passes `0`; the real increment never does), plus a positive control asserting the throwing call was reached, plus an assertion that no phantom count was left behind.

### Verification

| Check | Result |
|---|---|
| Cold typecheck (`tsconfig.tsbuildinfo` deleted, `pnpm typecheck`) | **0 type errors** in 95s |
| Targeted suites | **3 files / 48 tests passed** (`image-upload-relay.metrics.test.ts`, `relay.test.ts`, `metrics-endpoint-image-upload-relay-seeding.test.ts`) |
| ESLint (`--max-warnings=0`, JSON output read for the file count) | **4 files linted, 0 problems** |
| Prettier | clean on all 4 touched files; `origin/main`'s copies confirmed clean **first**, and a positive control (deliberate malformation) confirmed the check can go red |
| Component (browser) tier | **not run** — cannot run on this host |

**Behaviour-equivalence re-check.** Normalised the base and head bodies of `relay.ts` (comments and blank lines stripped, every `return <x>;` rewritten to `return;`) and diffed:

* vs `origin/main` — the only statement-level differences are the two added `return`s (after `res.status(200).json({ id: key })` and after `handleEndpointError(res, e)`) plus the metrics import and the `imageUploadRelay` / `runRelay` split itself. Every other branch normalises to the `return;` that was already there.
* vs the previous branch tip `92989abb6e` — **IDENTICAL** (144 normalised statement lines each side). These edits moved no statement.

**Mutation check on 🟢-5.** Isolation run (`-t "never throws"`): **red before** the fix (`TypeError: Cannot read properties of undefined (reading 'inc')`), **green after**. Then deleted the `try/catch` in `recordImageUploadRelay` and re-ran — the test fails with `expected [Function] to not throw an error but 'Error: registry exploded' was thrown`, i.e. it dies for *this* guard's specific reason, at the final increment. Mutation reverted; file restored byte-identical.

### Infra-disclosure sweep — other findings

Swept all 841 added lines of the branch diff for thresholds, retention rules, sampling rates, filter predicates and topology. Nothing else introduced by this PR discloses infra configuration. Two notes:

* `traces are head-sampled at 0.1` — **not** a disclosure: `0.1` is the in-repo default in `src/instrumentation.node.ts:184` on `main`. Left as-is.
* Two **pre-existing on `main`**, outside this PR's diff, reported but not changed:
  * `src/pages/api/v1/image-upload/relay.ts:115` (pre-existing hunk) — *"`/api/v1/*` is served by a pool whose replica floor is in the dozens, and there is no session affinity"*: deployment topology.
  * `src/shared/constants/trpc.constants.ts:45-50` — names the CDN vendor as the request-population source, *"not a sampled or filtered access log"*, the 17mb Next body limit, and a measured `~2.87M tRPC requests` over a named 24h window.

PR #4699, opened 2026-09-08
feat(apps): migrate /apps/installed to /apps/activity — widen past the slot flag, default to the feed, Build to the end of the sub-nav

ZacxDev:
Renames `/apps/installed` to `/apps/activity`, widens the page past the model-slot flag, defaults it to the activity feed, moves `Build` to the end of the sub-nav, and makes the activity table readable.

The page stopped being an installs surface some time ago — it hosts four tabs, three of which are not installs. This makes the route say so, and makes the gate honest about who has activity.

## What changed, per acceptance criterion

**1 — Route + label.** `src/pages/apps/installed.tsx` → `src/pages/apps/activity.tsx`, with a `statusCode: 301` redirect in `next.config.mjs` (not `permanent: true` — Next maps that to 308, which preserves the request method; these are GET-only pages whose inbound links are bookmarks). The page component is **moved, not stubbed**: a stub whose only job is to redirect reads as a live route, and `appsPageWidths`' fs-walk would then demand a width classification for a page that never renders. Same shape as the `/apps/build` consolidation. Sub-nav row now reads **Activity**. Every in-repo link is repointed and the prose swept.

**2 — Page gate widens to `appBlocks || appBlocksPages`.** Extracted as `resolveActivityPageAccess` (pure, node-importable — the `resolveAppsPageAccess` / `resolveBuildPageAccess` pattern) over a new shared `canAccessAppsActivity`. `appBlocks` is the model-**slot** flag; someone who has only ever run a full-page app (`/apps/run/<slug>`, gated on `appBlocksPages`) has generations, scope-gated API calls and Buzz spends recorded against them and **no slot install at all**. Refusing them a page called "Activity" was the dishonest half of the rename. The page body re-checks with the same predicate, so the SSR gate and the client gate cannot drift.

Verified before relying on it: `appBlocksPages` exists in `feature-flags.service.ts` (`availability: ['mod']`, fliptKey `app-blocks-pages-enabled`) and is **not** `toggleable`, so `computeUserFeatureFlagsOverlay` cannot move it between the server render and the first client paint — which is what lets one predicate serve an SSR-resolved gate and a client re-check.

**3 — Installs tab narrows to `features.appBlocks`.** The tab's content is slot subscriptions and per-model installs, which is exactly what that flag governs. Non-vacuous only because of (2). The **panel** is gated with the tab, not just the tab: a `Tabs.Panel` whose `value` has no tab is unreachable through the UI but still mounts its children on every render.

**4 — Tabs are URL-backed.** Controlled `value` + `onChange` writing `?tab=` through `router.replace(..., { shallow: true })` — no `getServerSideProps` re-run, no history entry per click. The default tab **drops** the key rather than writing `?tab=activity`. Resolution rules live in a pure `appsActivityTabs.ts` so they run in the blocking node tier: an unknown / repeated / prototype-key value falls back, and so does `?tab=subscriptions` for a viewer without the slot flag (handing that to `Tabs.value` selects a tab that is not in the list — a bar with nothing active over an empty panel).

**5 — Sub-nav order: Marketplace → Activity → Invites → Revenue → Build → Review.** `Build` led the table on the argument that it is "the front door for someone who has not built anything yet"; it is the narrowest-audience row here (`canAccessAppsBuild`), so leading with it put the smallest cohort's destination in the position that reads as what the bar is for. The stale comment is **rewritten**, not relocated. `chromeNavAlignsWithSubNav.test.ts` and `subNavRowsMatchPageGates.test.ts` are green; the chrome's two `/apps/installed` items are repointed and the platform-nav one relabelled to "App activity".

**6 — Header CTA removed.** Marketplace is a sub-nav tab. The three in-empty-state anchors stay — an empty panel with no way forward is a dead end.

**7 — `When` is relative.** The existing `DaysFromNow`, with the absolute stamp still in the tooltip (and now also in the `<time datetime>` attribute).

**8 — App name links to the store detail, GATED.** `getListingDetailHref(slug)`, gated on `hasAppsStoreAccess` read through `useOptionalFeatureFlags` so an absent provider fails **closed**. This is `AppNameCrumb`'s solution copied rather than re-derived. `/apps/store-preview/<slug>` `getServerSideProps`-gates on that predicate and answers `notFound`, and `appListings.getAppDetail` throws NOT_FOUND at scope `none` — an ungated link is a 404 affordance, the class #4668 shipped and #4685 exists to prevent.

**9 — Column ledger re-measured, not inherited.** See below.

## Test matrix — RED before, GREEN after

Every new test was run against `origin/main`'s code with the rest of this branch in place.

| file | tier | at `origin/main` | here |
|---|---|---|---|
| `resolveActivityPageAccess.test.ts` | unit (blocking) | the `appBlocksPages`-without-`appBlocks` case is red — main's gate is `if (!features?.appBlocks) return { notFound: true }` | 7/7 green |
| `appsActivityTabs.test.ts` | unit (blocking) | new module; new-feature coverage, labelled as such in the file | 12/12 green |
| `apps-activity-redirect.test.ts` | unit (blocking) | red — no rule, and `installed.tsx` still exists | 6/6 green |
| `AppActivityPanel.storeGate.browser.test.tsx` | component | **8 of 8 red** (`Cannot find element with locator: getByTestId('app-activity-app-name')` — main renders a plain `<Text>` with no testid) | 8/8 green |
| `AppActivityPage.browser.test.tsx` | component | **9 of 11 red** | 11/11 green |

The two that pass at `origin/main` are the positive-direction cases main already satisfies (`Installs` present with `appBlocks`; no tabs at all without either flag). Their red siblings are what carry the claim. The 9 reds, with their messages:

```
expected 'Installs' to be 'Recent activity'          (opens on the feed — main defaults to subscriptions)
expected 'Installs' to be 'Apps & permissions'       (?tab= selects on load — main's tabs read no query)
expected 'Installs' to be 'Recent activity'          (unknown ?tab falls back)
expected "vi.fn()" to be called 1 times, but got 0   (switching a tab rewrites the URL)
TypeError: undefined is not iterable                 (…and the default tab drops the key)
Cannot find element … Recent activity                (Installs ABSENT without appBlocks — main 404s the page)
Cannot find element … Recent activity                (…and the page still LOADS for that viewer)
expected [ 'Browse marketplace', …(3) ] to deeply equal [ 'Browse the marketplace', …(2) ]
expected true to be false                            (the removed header CTA is back)
```

## Criterion 8 — mutation check

Deleting `!canSeeStore ||` from `ActivityAppName`'s early return:

```
× 🔴 an ineligible viewer gets PLAIN TEXT, not a link
    AssertionError: an ineligible viewer must not be given an anchor:
    expected 'A' not to be 'A' // Object.is equality
× 🔴 FAILS CLOSED with no FeatureFlags provider at all
    AssertionError: expected 'A' not to be 'A' // Object.is equality

Tests  2 failed | 6 passed (8)
```

Both die on **their own** assertion, not on another test's error, and the three eligible-viewer cases stay green — which is what makes the red attributable to the gate rather than to the feature existing. Asserted as "not an anchor, no `href`" rather than as a tag literal, because Mantine's `<Text>` renders a `<p>` and pinning `'P'` would make the test about Mantine's default element; the eligible-viewer case is the positive control that keeps `not.toBe('A')` from being vacuous.

## Criterion 9 — the `appsWideLayout` EXEMPT entry, re-measured

The exemption rested on a measurement taken while `When` rendered a fixed-width `YYYY-MM-DD HH:mm` stamp, so it no longer described the table. Redone at 768 / 1200 / 1440 / 2560 on the same rich `tip` fixture, `AppsWideLayout.geometry.test.tsx` — `When | App | Action | Detail | Status`, natural layout, cell widths in px:

```
before   @768    119.67 | 177.34 | 194.36 | 179.09 |  65.53
         @1200   189.94 | 281.44 | 308.44 | 284.23 | 103.95
         @1440   228.95 | 339.27 | 371.81 | 342.63 | 125.34
         @2560   411.09 | 609.14 | 667.58 | 615.19 | 225.00

after    @768    108.39 | 180.58 | 197.91 | 182.38 |  66.75
         @1200   172.02 | 286.59 | 314.08 | 289.44 | 105.88
         @1440   207.38 | 345.47 | 378.63 | 348.91 | 127.63
         @2560   372.33 | 620.30 | 679.80 | 626.45 | 229.13

ROW HEIGHT  36.19 at all four widths, before AND after.
```

`When` gave up ~10% of its width at every viewport (38.76px at 2560) and the other four absorbed it proportionally — the table stayed at its natural distribution, it did not develop a surplus. The `no-surplus` argument is a claim about max-content sum vs container width at 768, and a narrower `When` makes that sum *smaller*, i.e. further from the case a ledger could help. **Control on the instrument:** the `before @2560` row reproduces the `natural @2560` figures already recorded in that comment, exactly.

Two measurement repairs came out of this and are worth reading:

- the geometry fixture used an absolute `createdAt`, which under `DaysFromNow` renders a string that **grows with wall-clock time** ("7 days ago" → "a year ago") and would have quietly moved the widths this file measures. Now a fixed offset from `now`.
- the `…each ONE line box` arm reads `lineCount` = `Range.getClientRects().length`, which returns one rect **per box** in the range. Once `When` became `<Text><time>…</time></Text>` the wrapper's range held two boxes and returned 2 **with nothing having wrapped**. The read drills to the `<time>`; the row-height arm staying at 36.19 across all four widths is the independent proof the cell is still one line.

## Also in here

`appsStoreAccessCallSites`' blanking-ratio bound moved 0.55 → 0.65. Measured on `origin/main` it sat at **0.5486** against a 0.55 bound — 0.0014 of headroom, so any added line of documentation (or one longer JSX string, which masks as blanks too) tripped it. It now carries a **negative control on the real file** — the historical apostrophe desync reproduced, blanking the tail — proving the bound can still fire. The `AppActivityPanel.tsx` entry was added to `STORE_GATE_SITES` with its reason.

## Gates run

- `pnpm typecheck` — **0 errors under `src/`**. 31 remain under `packages/`, all `Cannot find package 'kysely'` from a local `packages/*/node_modules` gap, identical on a clean `origin/main` worktree.
- `vitest --project unit` (full) — **34,638 passed / 30 failed**. All 30 are in `src/server/**`, `src/tests/**`, `src/test-utils/**`, none in a file this PR touches. Confirmed pre-existing: the same 6 files reproduce **23 identical failures** in a clean `origin/main` worktree sharing the same `node_modules`; the remaining 7 are a timing-sensitive watchdog test and a suite added by `e76bc60be7`, which this branch is rebased onto.
- `vitest --project component` on the 11 files in scope — **148 passed / 0 failed**.
- `vitest --project geometry` on `AppsWideLayout.geometry.test.tsx` — **34 passed / 0 failed**.
- `eslint` on the 7 added files — clean. `prettier --list-different` on the added files + the two modified files that were prettier-clean at `origin/main` — clean. The other 11 modified files are prettier-nonconforming **at `origin/main`** (measured per file via `--stdin-filepath`), and are deliberately left alone: CI treats modified files as report-only precisely so a 3-line change does not become a 200-line reformat.

## ⚠️ Superseded below

The two bullets that used to close this PR ("`blocks.*` untouched" and "the Activity row's
predicate is still `s.hasInstalls`") are **no longer true** — see the follow-up section.
They are named rather than deleted because they were the recorded justification, and the
audit found the justification itself was wrong.

---

# Audit follow-ups (commit `5107853cba`)

The operator approved a scope expansion into `blocks.*` for this. What the audit found:

🔴 **The justification for criterion 2 was false, and the real defect was elsewhere.**
Widening the page gate to `appBlocks || appBlocksPages` was argued from "someone who ran a
full-page app has activity and no slot install, so they were refused". They were not:
`src/pages/apps/run/[slug]/[[...path]].tsx:89` requires **BOTH** flags, so a page-app
runner always held `appBlocks` and the old page gate never refused them. The page gate is
kept as-is (operator's call). The **actual** defect was the sub-nav row — `hasInstalls` is
`dbRead.blockUserSubscription.findFirst(...) !== null`, and full-page apps are stateless by
design (`/apps/run`'s Decision 2: *"no `block_user_subscriptions` row, no migration"*), so a
viewer with both flags who runs page apps has activity and **no Activity tab**.

## What changed

| # | Item | Status |
|---|---|---|
| 1 | `getNavSummary` reports page-app activity (`hasActivity`) | done |
| 2 | Activity row keys on `s.hasInstalls \|\| s.hasActivity` | done |
| 3 | Empty-state `/apps` anchors gated on `hasAppsStoreAccess` (both sites) | done |
| 4 | `ActivityAppName` never links a listing that does not exist — fixed **in the service** | done |
| 5 | `subNavRowsMatchPageGates` pins the Activity row; header claim corrected | done |
| 6 | Two stale comment claims corrected | done |
| 7 | Double tooltip removed; `live` passed | done |
| 8 | Run-frame drawer: link **suppressed**, not left | done (decided, below) |

### 1. The `hasActivity` predicate, and why it matches the feed

```ts
hasActivity: buzzActivity !== null || scopeActivity !== null
```

built from two bounded probes added to the existing `Promise.all`:

```ts
dbRead.blockBuzzAttribution.findFirst({ where: { userId: user.id }, select: { id: true } }),
dbRead.blockScopeInvocation.findFirst({
  where: {
    userId: user.id,
    OR: [{ appBlockId: { not: null } }, { syntheticAppId: { not: null } }],
  },
  select: { id: true },
}),
```

These are **the two tables the feed itself walks** — `listMyAppActivity` reads
`block_buzz_attribution` on the SPENDER's `userId`
(`user-app-surface.service.ts:222`), `listMyScopeInvocations` reads
`block_scope_invocations` on the same column (`:392`). The scope probe copies the feed's
own `WHERE`, **external-OAuth exclusion included**: the global feed keeps `app-block` and
synthetic dev-tunnel rows and drops external-OAuth rows (both columns null). Without that
term an external-OAuth-only viewer would be handed an Activity tab whose feed says "No
activity yet" — the mirror of the defect being fixed. `getNavSummary`'s all-false
short-circuit shape gains the key too.

### 4. The slug fix is server-side

`user-app-surface.service.ts:261` and `:450` emitted
`appSlug: r.appBlock?.blockId ?? r.appBlockId`. Those two columns are not
interchangeable: `AppBlock.blockId` is the slug `AppListing.slug` mirrors
(`app-listing-mapper.ts` writes `slug: ab.blockId`); `appBlockId` is the **foreign key**.
So an unresolved join produced `/apps/store-preview/<pk>` — a guaranteed 404, offered
exactly on the rows where the app is least resolvable. That join really does come back
null: a pre-approval App-Dev-Tunnel spend writes `appBlockId: null` + `syntheticAppId`.

Both sites now emit `null`, the item types are `string | null`, `ActivityAppName` renders
plain text when absent, and the slug badge beside it is dropped rather than printing the
primary key. `appName` keeps its `?? r.appBlockId` tail on purpose — a display string with
no navigational meaning. **No per-row client fetch was added.**

A pre-existing test asserted the OLD behaviour
(`expect(result.items[0].appSlug).toBe('apb_x')`) and went red on the fix — that is the
change's own red→green, in the inverse direction. It is rewritten with the reason the old
expectation was the defect written down.

### 8. The drawer decision (item 8)

`AppPermissionsActivityDrawer` mounts the shared panel **over a running full-page app**, so
the new app-name link would top-level-navigate the user out of it — from a panel they
opened to read, with no warning and no route back to their in-app state. The column is also
redundant there (every row is the same app).

**Decision: suppress the link in per-app drill-down mode.** `ActivityAppName` gains
`linkable`, and the panel passes `linkable={!appBlockId}` — reusing the existing
"plain text" branch rather than adding a second rendering of the column. Not
`target="_blank"`: a second mechanism for one column is how the two come to disagree.
Written down on both components and pinned by a test in
`AppPermissionsActivityDrawer.browser.test.tsx` (the seam), with a positive control showing
the same viewer + same row DOES get a link on the whole-account feed.

## Red→green matrix

Every behavioural change was watched to fail without it.

| Change | Red without it | Green with it |
|---|---|---|
| `hasActivity` on the row | `subNavRowsMatchPageGates` › *Activity is visible exactly when…* + *the page ADMITS the cohort…* fail; `AppsSubNav.browser` › *Activity shows for hasActivity with NO installs* fails | unit 10/10, component 48/48 |
| empty-state anchor gate | `AppActivityPage.browser` › *a page-apps-only viewer gets NO marketplace anchor…* and `AppActivityPanel.storeGate` › *a page-apps-only viewer gets NO marketplace link* + *FAILS CLOSED…* fail (3) | 30/30 across the two files |
| `appSlug` → null | pre-existing `user-app-surface.orchestration` › *falls back to appBlockId…* went RED on the fix (it pinned the defect) | rewritten + 2 new cases; file green |
| drawer `linkable` | drawer › *the App name is PLAIN TEXT in the drawer…* and panel › *appBlockId set ⇒ plain text…* fail (2) | 25/25 |
| tooltip removal | see the mutation note below | 17/17 |

## Mutation checks (exact messages)

**A — delete the `hasActivity` term** (`visible: (s) => s.hasInstalls`):

```
subNavRowsMatchPageGates › Activity is visible exactly when the viewer has installs OR activity
AssertionError: hasInstalls=false hasActivity=true (others=false store=false author=false):
the Activity TAB is hidden. …: expected false to be true
```

and, in the browser tier, `AppsSubNav.browser.test.tsx:225` —
`await expect.element(tab('Activity')).toBeInTheDocument()` → `Matcher did not succeed in
time`, **1 failed / 47 passed**. The `Activity shows when hasInstalls` case stays green, so
the red is attributable to this term.

**B — delete the `canSeeStore &&` guard from both empty states**: 3 tests fail on their own
assertions, 27 stay green:

```
AppActivityPage › a page-apps-only viewer gets NO marketplace anchor anywhere on the page
AssertionError: a link into a `notFound` was offered to a viewer with no store access:
expected [ 'Browse the marketplace', …(1) ] to deeply equal []

AppActivityPanel.storeGate › a page-apps-only viewer gets NO marketplace link
AssertionError: a link into a `notFound` was rendered for a store-less viewer:
expected <a …(6)></a> to have a length of +0 but got 1
```

**C — drop `linkable={!appBlockId}`**: 2 fail, 23 pass —
`AssertionError: the drawer must not offer a link that navigates out of the running app:
expected 'A' not to be 'A'`.

🔴 **D — re-add the `<Tooltip>`: the first version of that guard SURVIVED (17/17 green).**
A **closed** Mantine `Tooltip` adds no attribute and no element, so wrapped and unwrapped
markup are byte-identical in the DOM — the structural guard (counting `[title]` nodes and
`aria-describedby`) could not see the hazard it claimed to cover. It is rewritten as a real
hover: `.hover()` the cell, poll for `[role="tooltip"]`. Under the re-added wrapper it now
fails —
`AssertionError: a second tooltip was re-added over DaysFromNow… expected [ Array(1) ] to
deeply equal []` — and passes with the wrapper gone.

## Gate results (counts, not exit codes)

- `pnpm typecheck` — **0 type errors**, and the instrument was negative-controlled: removing
  `hasActivity` from one typed fixture produced `1 type error(s)` / exit 2
  (`TS2741: Property 'hasActivity' is missing … but required in type 'AppsNavSummary'`).
- `vitest --project unit` (**full tier**) — **38,894 passed / 1 failed / 33 skipped** across
  1,689 files. The one failure is `src/server/__tests__/eventloop-watchdog.capture.test.ts`,
  a timing test unrelated to this diff. Controlled by alternating runs against a clean
  `HEAD` worktree: base run 1 **2 failed**, mine run 1 **1 failed**, base run 2 **9/9
  passed**, mine run 2 **9/9 passed** — flaky at base as well as here.
- `vitest --project unit` scoped to `components/Apps`, `components/AppBlocks`,
  `server/routers/__tests__`, `server/services/blocks/__tests__` — **348 files / 7,585
  passed / 0 failed**.
- `vitest --project component` on every browser test under `components/Apps` +
  `components/AppBlocks` — **127 files / 1,783 passed / 0 failed**.
- `vitest --project geometry` on `AppsWideLayout.geometry.test.tsx` — **34 passed / 0
  failed**.
- `prettier` — the 3 files this commit newly made non-conforming were written; the other 5
  touched files were **already** non-conforming at this branch's HEAD (measured two ways:
  `--stdin-filepath` on `git show HEAD:<file>`, and `--list-different` on a clean
  `git worktree` of HEAD — both name the same 5). CI treats modified files as report-only.
- `eslint` on all 18 changed files — **16 errors, 3 warnings, all pre-existing** (JSX
  apostrophes, empty arrow functions, the drawer test's pre-existing wholesale `~/utils/trpc`
  mock, `blocks.router.ts:5347/8135`, `user-app-surface.service.ts:16`). Every reported line
  is outside this commit's hunks — one that *was* inside (`onClose={() => {}}` in a new test)
  was changed to `vi.fn()` so this commit adds none.

## Fixtures updated (all green)

`AppsSubNav.{browser,hydration,ssrHydration,storeGate}.browser.test.tsx`,
`AppsBuildBody.browser.test.tsx`, `AppActivityPage.browser.test.tsx`,
`blocks.router.getNavSummary.test.ts`, `subNavRowsMatchPageGates.test.ts`. Also
`appsStoreAccessCallSites.test.ts` — the ledger **grew** by `pages/apps/activity.tsx`, with
a carve-out explaining why a runtime-gated page can still hold a store-gated link.

## Still not done / not verified

- **The page gate stays `appBlocks || appBlocksPages`** — operator's call, not re-litigated.
- **A residual narrowness survives, deliberately.** Both summary flags come from
  `getNavSummary`, whose `enabled` and whose server middleware gate on `appBlocks`, while the
  page admits `appBlocks || appBlocksPages`. A viewer holding `appBlocksPages` **without**
  `appBlocks` still gets an all-false summary and no tab. That is the safe direction of the
  #3899 / #4668 class (an unreachable page, never a tab into a 404); closing it means moving
  the procedure's own gate, which widens every flag on the summary. Recorded on the row.
- **`live` on `DaysFromNow` is not pinned by a test.** Asserting it would need fake timers
  against dayjs; the change is a one-word prop and the stale-relative-string behaviour it
  fixes is not covered.
- **No live/manual verification.** Everything above is tests, typecheck and lint on this
  machine — the click path on a real `/apps/activity` for a page-apps-only viewer has not
  been exercised.
- The two added `getNavSummary` probes are asserted to be `findFirst` + `select: { id }` +
  user-scoped, but their **cost against production data has not been measured**.

## Base sync (`fef6336dd5`)

`origin/main` was merged into this branch — **no force-push, no rewrite, and none of this
PR's files are touched by main's side** (`git diff HEAD^1 HEAD -- <this PR's paths>` is
empty). It was needed because CI was red on a failure this branch does not contain:

`Unit tests (2)` failed on `no-direct-shared-module-mock` naming
`src/server/routers/__tests__/article.router.scan-status-authz.test.ts` — a file that **does
not exist on this branch**. GitHub's `refs/pull/4699/merge` was pinned at base `02ad4701ad`,
the commit immediately *before* #4709 (`2a19acf8ef`, "use the canonical db mock in the
scan-status authz test") fixed exactly that file. Verified rather than inferred:
`git merge-base --is-ancestor 2a19acf8ef refs/pull/4699/merge` → **NO**, and that ref's copy
of the file still carries the `vi.mock('~/server/db/client', …)` the guard rejects. Re-running
the job reused the same stale merge ref and failed identically; only moving the head
recomputes it.

Re-gated on the MERGED tree afterwards: `pnpm typecheck` **0 errors**; `vitest --project unit`
over `components/Apps`, `components/AppBlocks`, `server/routers/__tests__`,
`server/services/blocks/__tests__` **plus the guard test itself** — **350 files / 7,593 passed
/ 0 failed**; the 9 in-scope browser files — **156 passed / 0 failed**. PR checks now read
**18 pass / 0 fail** (1 `skipping`, a main-pushes-only job).

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -421,6 +421,26 @@ export default defineNextConfig(
           destination: '/apps/build',
           statusCode: 301,
         },
+        // ── `/apps/installed` → `/apps/activity` ───────────────────────────────────
+        // The page was renamed when it stopped being an installs surface: it now opens
+        // on the activity feed, and its gate widened from `appBlocks` (the model-slot
+        // flag) to `appBlocks || appBlocksPages`, so a viewer whose only app usage is a
+        // full-page app reaches it. The page component is MOVED (`installed.tsx` →
+        // `activity.tsx`), not stubbed — same reason as the three rules above.
+        //
+        // 🔴 `statusCode: 301`, matching its neighbours and for the same reason: Next
+        // maps `permanent: true` to 308, and these are GET-only pages whose inbound
+        // links are bookmarks and search results. The two keys are mutually exclusive
+        // in Next's schema.
+        //
+        // Next preserves the query string across a redirect, so an inbound
+        // `/apps/installed?tab=permissions` keeps its `?tab=` — which only became a
+        // meaningful statement when the tabs went URL-backed in the same change.
+        {
+          source: '/apps/installed',
+          destination: '/apps/activity',
+          statusCode: 301,
+        },
         {
           source: '/api/download/training-data/:modelVersionId',
           destination: '/api/download/models/:modelVersionId?type=Training%20Data',
--- a/src/__tests__/pages/apps-activity-redirect.test.ts
+++ b/src/__tests__/pages/apps-activity-redirect.test.ts
@@ -1,0 +1,139 @@
+import fs from 'fs';
+import path from 'path';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * `/apps/installed` → `/apps/activity`, as a REAL redirect rather than a stub page.
+ *
+ * The page stopped being an installs surface: it opens on the activity feed, its
+ * `Installs` tab is gated on the model-slot flag, and its own gate widened to
+ * `appBlocks || appBlocksPages` so a viewer whose only app usage is a full-page app
+ * reaches it. The route name followed.
+ *
+ * 🔴 THIS READS THE ACTUAL `next.config.mjs` EXPORT AND CALLS `redirects()`. A text scan
+ * of the file would pass on a commented-out entry, on an entry in the wrong array, and
+ * on one whose object never reaches the returned list — all three of which look right in
+ * a diff. Invoking the function is the only check that the rule is one Next will serve.
+ *
+ * Structure and controls deliberately mirror `apps-build-redirects.test.ts`, which is
+ * the same job for the `/apps/build` consolidation; that file owns its own three routes
+ * and this one owns this route, so neither's ledger can absorb the other's silently.
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const nextConfig = (await import('../../../next.config.mjs')).default as any;
+
+type Redirect = {
+  source: string;
+  destination: string;
+  permanent?: boolean;
+  statusCode?: number;
+};
+
+const redirects = (await nextConfig.redirects()) as Redirect[];
+
+const RETIRED_ROUTE = '/apps/installed';
+const DESTINATION = '/apps/activity';
+
+describe('the /apps/installed → /apps/activity redirect', () => {
+  it('POSITIVE CONTROL: the redirect list is populated and reachable', () => {
+    // 🔴 Without this, every assertion below could be satisfied by a `find` over an empty
+    // array producing `undefined`. A reassuring "no match" is indistinguishable from a
+    // probe wired to nothing.
+    expect(Array.isArray(redirects)).toBe(true);
+    expect(redirects.length).toBeGreaterThan(5);
+    expect(redirects.some((r) => r.source === '/discord')).toBe(true);
+  });
+
+  it('🔴 /apps/installed 301s to /apps/activity', () => {
+    const rule = redirects.find((r) => r.source === RETIRED_ROUTE);
+    expect(rule, `no redirect rule for ${RETIRED_ROUTE}`).toBeDefined();
+    expect(rule!.destination).toBe(DESTINATION);
+    // 🔴 301, not `permanent: true` — Next maps `permanent` to 308, which preserves the
+    // request METHOD. This is a GET-only page whose inbound links are bookmarks and
+    // search results; 301 is the status those consumers cache and rewrite on. The two
+    // options are mutually exclusive in Next's schema.
+    expect(rule!.statusCode).toBe(301);
+    expect(rule!.permanent).toBeUndefined();
+  });
+
+  it('🔴 it does not land on another redirect', () => {
+    // A 301→301 costs a round trip, and some link-equity and bookmark-rewriting
+    // consumers stop following after the first hop.
+    const sources = new Set(redirects.map((r) => r.source));
+    expect(sources.has(DESTINATION)).toBe(false);
+    // Guard-the-guard: a `sources` set built from an empty list makes the check above
+    // trivially pass. Prove the lookup can hit.
+    expect(sources.has(RETIRED_ROUTE)).toBe(true);
+  });
+
+  it('🔴 the page MOVED — installed.tsx is gone and activity.tsx exists', () => {
+    // A stub whose only job is to redirect is dead code that reads as a live route, and
+    // `appsPageWidths`' fs-walk would then demand a width classification for a page that
+    // never renders. So the component is moved, not emptied.
+    const pagesDir = path.resolve(__dirname, '../../pages/apps');
+    expect(fs.existsSync(pagesDir)).toBe(true); // control: looking in the right place
+    expect(fs.existsSync(path.join(pagesDir, 'installed.tsx'))).toBe(false);
+    expect(fs.existsSync(path.join(pagesDir, 'activity.tsx'))).toBe(true);
+  });
+
+  it('no in-app link still points at /apps/installed', () => {
+    // A link to a redirected route costs a round trip and re-renders the whole shell. The
+    // redirect is for BOOKMARKS and notification URLs already in the wild, not for links
+    // this codebase is still emitting today.
+    const srcDir = path.resolve(__dirname, '../..');
+    const offenders: string[] = [];
+    const linkish = new RegExp(`(?:href|destination)\\s*[=:]\\s*['"]${RETIRED_ROUTE}['"]`);
+    const pushy = new RegExp(`router\\.(?:push|replace)\\(\\s*['"]${RETIRED_ROUTE}['"]`);
+    const walk = (dir: string) => {
+      for (const name of fs.readdirSync(dir, { withFileTypes: true })) {
+        const full = path.join(dir, name.name);
+        if (name.isDirectory()) {
+          if (name.name === 'node_modules' || name.name === '__screenshots__') continue;
+          walk(full);
+          continue;
+        }
+        if (!/\.(ts|tsx)$/.test(name.name)) continue;
+        // This file names the retired route as DATA. Excluding self is what keeps the
+        // guard from reporting itself.
+        if (full === __filename) continue;
+        const text = fs.readFileSync(full, 'utf8');
+        // Only LINK-shaped occurrences: a quoted string literal in an href/destination
+        // position. Prose in comments that explains the rename is documentation, not a
+        // link — and this change deliberately left some of it.
+        if (linkish.test(text) || pushy.test(text)) {
+          offenders.push(`${path.relative(srcDir, full)} → ${RETIRED_ROUTE}`);
+        }
+      }
+    };
+    walk(srcDir);
+    expect(offenders).toEqual([]);
+  });
+
+  it('🔴 POSITIVE CONTROL for the link walk: it CAN find a link', () => {
+    // The walk above returning `[]` is indistinguishable from a walk whose regexes match
+    // nothing. Feed the same two patterns a string they MUST hit. Without this the whole
+    // sweep is a reassuring zero — the exact failure mode a redirect guard is worst at,
+    // because a missed link still WORKS (via the redirect) and so is never reported.
+    const sample = `<Link href="/apps/installed">x</Link>; router.push('/apps/installed');`;
+    expect(
+      new RegExp(`(?:href|destination)\\s*[=:]\\s*['"]${RETIRED_ROUTE}['"]`).test(sample)
+    ).toBe(true);
+    expect(
+      new RegExp(`router\\.(?:push|replace)\\(\\s*['"]${RETIRED_ROUTE}['"]`).test(sample)
+    ).toBe(true);
+    // …and that the walk reaches a plausible number of files.
+    let count = 0;
+    const walk = (dir: string) => {
+      for (const name of fs.readdirSync(dir, { withFileTypes: true })) {
+        const full = path.join(dir, name.name);
+        if (name.isDirectory()) {
+          if (name.name === 'node_modules' || name.name === '__screenshots__') continue;
+          walk(full);
+        } else if (/\.(ts|tsx)$/.test(name.name)) count += 1;
+      }
+    };
+    walk(path.resolve(__dirname, '../..'));
+    expect(count).toBeGreaterThan(500);
+  });
+});
--- a/src/components/AppBlocks/AppBlockChromeMobileShell.browser.test.tsx
+++ b/src/components/AppBlocks/AppBlockChromeMobileShell.browser.test.tsx
@@ -377,11 +377,15 @@ describe('AppBlockChrome mobile shell', () => {
   });
 
   test(`REGRESSION — at ${PHONE[0]}x${PHONE[1]} the platform nav is inside the ⋮ sheet`, async () => {
-    // 🔴 RED AT `origin/main`, BEHAVIOURALLY. The failing assertion is the presence of
-    // "Installed apps" after opening the ⋮. That label exists ONLY in the platform-nav
-    // section — the ⋮ menu's own item for the same route is worded "Manage apps" — so
-    // at main, where the ⋮ carries app actions only, it is not reachable from this
-    // trigger at any width.
+    // 🔴 RED AT THE REVISION THAT INTRODUCED IT, BEHAVIOURALLY. The failing assertion is
+    // the presence of the platform nav's own label for the activity route after opening
+    // the ⋮. That label exists ONLY in the platform-nav section — the ⋮ menu's own item
+    // for the same route is worded "Manage apps" — so at the pre-fold revision, where the
+    // ⋮ carries app actions only, it is not reachable from this trigger at any width.
+    //
+    // ⚠️ THE LABEL MOVED WITH THE ROUTE: it read "Installed apps" until `/apps/installed`
+    // became `/apps/activity` and it became "App activity". Same item, same section, same
+    // claim — only the word changed.
     //
     // 🔴 IT IS ASSERTED BY LABEL, NOT BY A TESTID, ON PURPOSE. A testid on the new
     // sheet would go red at main merely by not existing, which proves nothing. A user-
@@ -399,7 +403,7 @@ describe('AppBlockChrome mobile shell', () => {
     expect(sheet, 'the ⋮ must have opened SOME surface').not.toBeNull();
 
     const labels = rowLabels(sheet as HTMLElement);
-    for (const label of ['Marketplace', 'Installed apps', 'My apps']) {
+    for (const label of ['Marketplace', 'App activity', 'My apps']) {
       expect(
         labels,
         `at ${PHONE[0]}px "${label}" must be reachable from the ⋮ — the platform nav folds into ` +
--- a/src/components/AppBlocks/AppBlockChromePlatformNav.browser.test.tsx
+++ b/src/components/AppBlocks/AppBlockChromePlatformNav.browser.test.tsx
@@ -2,7 +2,7 @@ import { describe, expect, test, vi, beforeEach } from 'vitest';
 import { page } from 'vitest/browser';
 
 // Part A: the app icon opens a Menu of the Civitai App PLATFORM's own pages
-// (Marketplace / Installed apps / My apps / Review). "Review" is gated on
+// (Marketplace / App activity / My apps / Review). "Review" is gated on
 // the viewer's moderator flag. Part B: the ⋯ menu gains a "Permissions &
 // activity" item (only when an appBlockId is threaded) that opens a per-app
 // transparency drawer.
@@ -88,8 +88,8 @@ describe('AppBlockChrome platform-nav menu (Part A)', () => {
     const home = page.getByRole('menuitem', { name: 'Marketplace' }).element();
     expect(home.getAttribute('href')).toBe('/apps');
 
-    const installed = page.getByRole('menuitem', { name: 'Installed apps' }).element();
-    expect(installed.getAttribute('href')).toBe('/apps/installed');
+    const installed = page.getByRole('menuitem', { name: 'App activity' }).element();
+    expect(installed.getAttribute('href')).toBe('/apps/activity');
 
     // Was "My submissions" → `/apps/my-submissions`, then `/apps/mine`; that table is now
     // state C of the consolidated `/apps/build`, and this item was REPOINTED rather than
--- a/src/components/AppBlocks/AppPermissionsActivityDrawer.browser.test.tsx
+++ b/src/components/AppBlocks/AppPermissionsActivityDrawer.browser.test.tsx
@@ -18,6 +18,11 @@ const m = vi.hoisted(() => ({
   grantsSpy: undefined as unknown as ReturnType<typeof vi.fn>,
   buzzSpy: undefined as unknown as ReturnType<typeof vi.fn>,
   scopeSpy: undefined as unknown as ReturnType<typeof vi.fn>,
+  // Store-visibility flags for `AppActivityPanel`'s `App` column. `null` (the default,
+  // and what this file rendered under before) is what `useOptionalFeatureFlags` returns
+  // outside a provider, which fails CLOSED — so a link test run at the default would be
+  // vacuous. The seam test below sets a store-ELIGIBLE viewer deliberately.
+  flags: null as null | Record<string, boolean>,
 }));
 
 // 🔴 `AppActivityPanel`'s `When` column renders `DaysFromNow`, which reads
@@ -29,6 +34,13 @@ vi.mock('~/providers/IsClientProvider', () => ({ useIsClient: () => true }));
 
 vi.mock('~/hooks/useCurrentUser', () => ({
   useCurrentUser: () => m.user,
+}));
+
+vi.mock('~/providers/FeatureFlagsProvider', () => ({
+  useFeatureFlags: () => m.flags ?? {},
+  useOptionalFeatureFlags: () => m.flags,
+  useFeatureFlagsReady: () => true,
+  FeatureFlagsProvider: ({ children }: { children: unknown }) => children,
 }));
 
 vi.mock('~/utils/trpc', () => {
@@ -75,6 +87,7 @@ import { renderWithProviders } from '../../../test/component-setup';
 
 beforeEach(() => {
   m.user = { id: 1, username: 'viewer', isModerator: false };
+  m.flags = null;
   m.grants = [];
   m.buzz = [];
   m.scopes = [];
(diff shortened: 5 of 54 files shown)


ZacxDev:
## Audit round 1 — findings fixed, and what I re-verified myself

The blind round-1 audit found that **criterion 2's stated rationale was false**, and the operator chose to fix it properly rather than revert. Fixes are in; below is what **I** re-ran rather than relayed.

### Independently re-verified

| claim | my measurement |
|---|---|
| `hasActivity` term is load-bearing on the Activity row | **Mutation**: `visible: (s) => s.hasInstalls \|\| s.hasActivity` → `s.hasInstalls`. Blocking unit tier: **2 failed / 8 passed**, both on their own assertions — `hasInstalls=false hasActivity=true …: the Activity TAB is hidden` and `a viewer the page serves — page-app activity, zero slot installs — has no tab`. Baseline 10/10. |
| the rewritten double-tooltip guard actually works | 🔴 **My first attempt at this mutation was INVALID and I am recording that.** I re-added `<Tooltip>` without its import → `ReferenceError: Tooltip is not defined` → 14 failures, all at the 15 s matcher ceiling. That is a mount crash, not a guard firing; it proved nothing. Redone **with** the import: **1 failed / 16 passed**, twice (6.5 s / 7.2 s, warm), killed by exactly the tooltip test with its own message — `a second tooltip was re-added over DaysFromNow …: expected [ Array(1) ] to deeply equal []`. |
| the merge-of-main was not a rewrite | `fef6336dd5` is a real merge commit; `2a19acf8ef` (#4709 — the fix for the file CI was complaining about) is on main's side, which corroborates the stale-`refs/pull/4699/merge` diagnosis. No force-push, base still `main`, not stacked. |
| current state | **19/19 checks: 18 pass, 1 skipping. `MERGEABLE` / `CLEAN`.** Touched suites re-run: unit 54/54, component 78/78. |

### The implementer self-reported a vacuous guard, and that is worth naming

Its first double-tooltip guard **survived** the mutation (17/17 green): a *closed* Mantine `Tooltip` adds nothing to the DOM, so a structural check counting `[title]` / `aria-describedby` was measuring nothing. It rewrote it as a real hover polling `[role="tooltip"]`, and that version dies. A guard that has never been watched fail is a claim, not a test — this one was caught before merge because the mutation was actually run.

### What is fixed

1. **`getNavSummary` gains `hasActivity`** — true when the viewer has a row in either table the feed actually walks (`block_buzz_attribution` as spender, `block_scope_invocations`), the scope probe mirroring the feed's own `WHERE` including the external-OAuth exclusion, so the flag cannot light a tab over an empty page.
2. **The Activity row keys on `hasInstalls || hasActivity`** — installs are slot subscriptions; page apps are stateless by design and write none. This is the defect the audit actually identified.
3. **Both empty-state `/apps` anchors gated** on `hasAppsStoreAccess` (`appBlocksPages` is not one of its disjuncts), via `useOptionalFeatureFlags` so an absent provider fails closed.
4. **`appSlug` is a listing slug or `null`, never an `AppBlock` primary key** — fixed at the service, both feeds. The old `?? r.appBlockId` fallback produced `/apps/store-preview/<pk>`, a guaranteed 404 on exactly the rows where the app is least resolvable.
5. **`subNavRowsMatchPageGates` now pins the Activity row** against the page's own resolver; its header claim that summary-driven rows have "no second copy to disagree with" is corrected — *which* fields a row ORs together is exactly such a copy, and that blind spot is why none of this was caught earlier.
6. Two stale comment claims, the double tooltip, and `live` on `DaysFromNow`.
7. The run-frame permissions drawer passes `linkable={false}` — the shared panel mounts **over a running full-page app** there, so a top-level link would navigate the user out of it.

### 🔴 Residual, recorded rather than closed

A viewer holding `appBlocksPages` **without** `appBlocks` still gets no tab: both summary flags come from `getNavSummary`, which is itself `appBlocks`-gated, while the page admits either. Safe direction — an unreachable page, never a tab into a 404 — and closing it means moving that procedure's own gate, which would widen **every** flag on the summary. The operator kept the page-gate widening knowingly; the row comments record that its original justification was false.

### Not verified

No live click-path on `/apps/activity` for any cohort. `live` on `DaysFromNow` is not pinned by a test (needs fake timers against dayjs). The two new probes' cost against production data is unmeasured — only their shape (`findFirst`, `select: { id }`, user-scoped) is asserted. And the round-1 audit could not confirm a 301 *observed* — only the rule in the invoked config export; query-string preservation across the redirect is tested nowhere.

A delta re-audit of this fix round is running.

ZacxDev:
Round-1 claims block, posted separately because my fix-round comment above omitted it — and without a parseable block a delta re-audit silently degrades into a blind full audit, which would then read as covered. The tooling refused the round-2 brief for exactly that reason; this is the fix.

```audit-claims round=1 audited=0f5ef634cf..fef6336d
1. `blocks.getNavSummary` gains `hasActivity`, true when the viewer has a row in either table the activity feed walks (block_buzz_attribution as spender, block_scope_invocations), with the scope probe mirroring the feed's own WHERE including its external-OAuth exclusion.
2. The Activity sub-nav row keys on `hasInstalls || hasActivity`, so a viewer whose only signal is stateless page-app usage gets a tab; both terms are documented at the row.
3. The two empty-state `/apps` anchors are gated on `hasAppsStoreAccess` via `useOptionalFeatureFlags`, so the cohort the widened page admits is no longer offered links into a notFound; the CTA is omitted rather than reworded.
4. `appSlug` is emitted as a listing slug or null — never an AppBlock primary key — at both feed sites in user-app-surface.service.ts; `ActivityAppName` renders plain text when it is absent.
5. `subNavRowsMatchPageGates` now pins the Activity row against `/apps/activity`'s own resolver, and its header claim that summary-driven rows have "no second copy to disagree with" is corrected.
6. Two stale comment claims are corrected (AppsSubNav's getNavSummary `enabled` rationale, and the test header), the Mantine Tooltip is removed from the When cell in favour of DaysFromNow's own `<time title>`, and `live` is passed.
7. The run-frame permissions drawer passes `linkable={false}` so the app-name link cannot top-level-navigate a user out of a running full-page app.
8. The page gate stays `appBlocks || appBlocksPages` per the operator's decision; the row comments record that its original justification was false, and the residual (a pages-only viewer still gets no tab, since getNavSummary is itself appBlocks-gated) is documented rather than closed.
9. `origin/main` was merged into the branch as `fef6336dd5` — a real merge commit, no force-push, no rewrite — to clear a stale `refs/pull/4699/merge` that was failing CI on a file not on this branch.
```

ZacxDev:
## Audit round 2 — findings fixed (`e183310da0`)

Two of five findings were actionable. **The headline one was demonstrated rather than argued, and it is fixed the same way.**

**F2 (guard) — "the probe MIRRORS the feed" was two copies of one literal.** `getNavSummary`'s `hasActivity` probe and `listMyScopeInvocations` each asserted their *own* `where` against a hand-written literal in their own suite; neither read the other. The audit tightened the **feed's** clause and updated the **feed's own** literal — exactly the edit a developer making that change would make — and **both suites stayed green at 65/65** while the probe silently over-matched.

Fixed by construction: `GLOBAL_SCOPE_ACTIVITY_OR` is exported once from `user-app-surface.service.ts`; the feed spreads it, `blocks.router.ts` imports it. **Proof, same mutation:** tightening the predicate at its single source now reds **three tests across all three files** where it previously reddened none. The new `scopeActivityPredicate.test.ts` deliberately does *not* re-spell the clause as a third literal — that would recreate the defect — it asserts the shape and that neither call site carries its own copy.

**F1 (perf) — NOT fixed, because measurement inverts the analysis.** The audit was structurally right that `block_buzz_attribution` has no index leading on `user_id` (seven live indexes, none on it) and right that it could not reach a DB. Measured on prod:

| table | rows | size | probe plan |
|---|---|---|---|
| `block_buzz_attribution` (flagged) | **0** | 72 kB | Seq Scan, cost 0.00, **0.021 ms** |
| `block_scope_invocations` (cleared) | **863,305** | 314 MB | Bitmap Heap Scan, **6.0 ms**, 48 buffers read |

The severity ordering is the reverse of the structural ordering. Neither blocks a merge on a query cached 60 s per nav. The index is worth adding when that table grows — and per this repo's rule 8 it is a hand-applied migration, not a code change.

**F5 (comment)** — the documented residual said the `appBlocksPages`-without-`appBlocks` cohort "still gets an all-false summary and no tab", inviting a future widening of a `blocks.*` gate to close a gap that is effectively **empty**: that cohort is refused by `/apps/run` itself and cannot install, so all-false is the *correct* tab set for them. Corrected in place.

**F3, F4** — left deliberately, reasons recorded: F3's severity is bounded by the browser tier not gating a merge; F4 (`live` gives each feed row its own 15 s interval) is real but harmless at these volumes.

**One error of mine, caught by the gate:** my `as const` on the shared constant made it a readonly tuple Prisma rejects — 1 typecheck error, mine, fixed. And the 3 eslint errors on these files are **exactly** the 3 present at branch HEAD, measured via `--stdin-filename` rather than asserted.

**Gate:** unit **7,246 passed / 309 files** (routers + blocks services + Apps), component 78 passed on the touched specs, prettier clean, typecheck **0 errors under `src/`**, CI **19 pass / 1 skipping**.

```audit-claims round=2 audited=fef6336dd5..e183310d
1. The scope-activity predicate is single-sourced as `GLOBAL_SCOPE_ACTIVITY_OR`, exported from user-app-surface.service.ts and imported by blocks.router.ts, so the probe and the feed cannot drift.
2. `scopeActivityPredicate.test.ts` pins the shape of that predicate and asserts neither call site re-spells the clause; it deliberately does not restate the clause as a third literal.
3. The drift the round-2 audit demonstrated (tighten the feed + its own literal, 65/65 green) now reds three tests across three files.
4. The residual comment about the appBlocksPages-without-appBlocks cohort is narrowed: that cohort can generate no activity at all, so all-false is correct for them and the procedure's gate should NOT be widened for it.
5. F1 was measured rather than fixed: block_buzz_attribution is empty on prod (Seq Scan, 0.021 ms) while the cleared scope probe costs 6.0 ms on 863k rows; no index added.
6. F3 and F4 are knowingly not fixed, with the reasons recorded.
```

PR #4702, opened 2026-09-08
fix(search): stop a Meilisearch task wait timing out into a silent success

JustMaier:
Two defects in `src/server/meilisearch/util.ts`. They are in one PR because separating them is unsafe — see the last section.

## 1. A task-wait timeout became a silent success (the wider blast radius)

`waitForTasksWithRetries` did not forward its `client` when it recursed:

```ts
return waitForTasksWithRetries(taskUids, remainingRetries - 1);   // client dropped
```

Every retry therefore fell back to the module-level `searchClient` — a **different instance** from the one the metrics processors pass, and `null` whenever `SEARCH_HOST` is unset, where the `if (!client) return []` guard at the top returns an **empty array as if the tasks had succeeded**. A timeout turns into a success with no error anywhere.

## 2. An index reset died on any backlog, and looked alive while dead

`getOrCreateIndex` awaited index creation with a bare `client.waitForTask`, which takes the client default of **5s** and never retries. Index creation is a *queued* Meilisearch task, so that wait is for queue position, not for work. A `search-index-sync-*-reset` fired while the queue had a backlog threw `MeiliSearchTimeOutError` and built nothing — while the creation task **stayed enqueued**, so the dead job read as a running one. That cost two hours of misdiagnosis today.

It now uses the escalating retry budget (5s, 10s, … ≈75s total).

**This does not make a reset immune to a long backlog, and is not meant to.** One `images_v6` batch on this instance has taken 51 minutes; nothing in-process should wait that long. A quiet queue remains a precondition for a reset.

## Why they ship together

Before this change **nothing in the repo passed a client to `waitForTasksWithRetries`** — its only call sites are inside `util.ts`, and the pre-change one called `client.waitForTask` directly. Defect 1 was latent.

Fixing the wait is what makes it reachable: the search-index processors pass their own client to `getOrCreateIndex` (`base.search-index.ts:355`; the metrics indexes pass `metricsSearchClient`), and that client is now forwarded into the helper. Fixing #2 without #1 would have **introduced** the silent success rather than exposed one.

## Tests

`src/server/meilisearch/__tests__/get-or-create-index.test.ts`, 3 tests. Both defects have a measured revert:

| revert | result |
|---|---|
| drop `client` from the retry call | `Tests 2 failed \| 1 passed (3)` — `expected "vi.fn()" to be called 3 times, but got 1 times` — 1.36s |
| restore the bare `client.waitForTask` | `Tests 3 failed (3)` — `expected "vi.fn()" to be called 1 times, but got 0 times` — 1.37s |

Two notes for whoever edits these next, both also written into the file:

- The client-forwarding test asserts an **exact** call count. `toHaveBeenCalled()` cannot fail here — the first attempt uses the caller's client whether or not the retry forwards it. The original draft of this test asserted that and was vacuous.
- The tests use **fake timers** deliberately. `getOrCreateIndex` is wrapped in `withRetries(fn, 3, 60000)`, so anything thrown out of it sleeps 3×60s of *real* time; before the fake timers a revert took over two minutes per test and read as a wedged runner rather than a failure.

## Verification

- `pnpm exec eslint` on both touched files — 0 errors (6 pre-existing-style `any`/unused-mock-import warnings)
- Full unit suite — see comment below
- Nothing here writes to Meilisearch. Search on prod flapped three times this afternoon; a separate stability gate covers the live steps, and this PR is not one of them.

https://claude.ai/code/session_01T7ZW9caRhbAUDQGTXxGwgn

### DIFF
--- a/src/server/meilisearch/__tests__/get-or-create-index.test.ts
+++ b/src/server/meilisearch/__tests__/get-or-create-index.test.ts
@@ -1,0 +1,89 @@
+import { MeiliSearchTimeOutError } from 'meilisearch';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+/**
+ * Index creation is a QUEUED Meilisearch task, so waiting on it is waiting for queue position rather
+ * than for work. `getOrCreateIndex` used to call `client.waitForTask` bare, which takes the client
+ * default of 5s and never retries — so a `search-index-sync-*-reset` fired while the queue had a
+ * backlog threw MeiliSearchTimeOutError about four minutes in and built nothing, while the creation
+ * task sat enqueued and made the dead job look like a running one.
+ *
+ * These pin that a creation wait survives a timeout instead of ending the job. They do NOT claim the
+ * reset is immune to a long backlog: the retry budget is finite (~75s) and one images batch on prod
+ * has taken 51 minutes. A quiet queue is still a precondition.
+ */
+
+vi.mock('~/server/meilisearch/client', () => ({ searchClient: null, metricsSearchClient: null }));
+import { getOrCreateIndex } from '~/server/meilisearch/util';
+import { redisMock } from '~/__tests__/mocks/redis.mock';
+import { loggingMock } from '~/__tests__/mocks/logging.mock';
+
+const notFound = Object.assign(new Error('index_not_found'), { code: 'index_not_found' });
+
+function fakeClient({ timeouts }: { timeouts: number }) {
+  let remaining = timeouts;
+  const index = { uid: 'test_index', update: vi.fn() };
+  return {
+    created: false,
+    getIndex: vi.fn(function (this: any) {
+      // Not found until createIndex has run — the branch under test.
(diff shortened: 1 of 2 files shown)


JustMaier:
Full unit suite on `fe975d0ceb`, `--max-workers=4`:

```
Test Files  1685 passed | 3 skipped (1688)
     Tests  38949 passed | 35 skipped (38984)
     Errors  1 error
```

Zero test failures; the totals account for every file. The one error is infrastructure, not an assertion — a forks worker failed to start for `src/shared/tiptap/__tests__/blurb.node.test.ts` (`Timeout waiting for worker to respond`), which is what makes the run exit 1. That file passes on its own, run together with the submodule canary:

```
Test Files  2 passed (2)
     Tests  385 passed (385)     # blocks.router.workflow.test.ts + blurb.node.test.ts
```

The canary is there because `blocks.router.workflow.test.ts` collects **zero** tests in a worktree without `event-engine-common` initialised, and a zero-collection run still reads as a pass. 385 is nonzero, so this worktree's run means something.

PR #4703, opened 2026-09-08
feat(search): hide paid models from model search, behind a flag

JustMaier:
Closes the search half of ClickUp `868m1r2u7`. The card badge (#4678) and the feed `Hide Paid` filter (#4690) already shipped; this adds the same choice to `/search/models`, behind a Flipt flag that is off.

"Paid" here means a **gated** model — a live `PaidAccess` row — which is Justin's ruling, not the ~31K licensing-fee models. ~3,181 models on prod as of tonight.

## Do not enable the flag before the index settings have APPLIED

Measured against live prod Meilisearch, not reasoned about:

```
baseline site filter                            200, 21,140 hits, processingTimeMs 23
baseline + AND NOT hasActivePaidAccess = true   400 Attribute hasActivePaidAccess is not filterable
```

Turning `paid-model-search-filter` on before `updateFilterableAttributes` has applied makes **every models search 400 for anyone who ticks the box** — not just the paid results, the whole result set. Bounded and self-healing (the clause is only sent when the box is ticked, and the state is per-session), but loud.

The only load-bearing constraint is that **both the settings change and the backfill precede the flag**. Flag before the settings is the 400 above; flag before the backfill hides nothing at all, which reads as "the feature doesn't work". Between themselves the order does not matter much, and in production the backfill went first — see below.

## The polarity is the whole feature

The clause is `NOT hasActivePaidAccess = true`, **never** an equality.

Only GATED models carry the attribute. Around 700K model documents do not have it — the deployed indexer writes it, but existing documents predate the field, and the card reads `!!data.hasActivePaidAccess` so absent means false. In Meilisearch an **equality matches only documents that HAVE the attribute**, while a **negation also matches documents that lack it**. `= false` would hide nearly the whole index and "hide paid" would silently return almost nothing — precisely the failure this ticket exists to prevent. `!= true` reads as the opposite but excludes absent documents the same way; both spellings are forbidden by a guard named for the decision.

Measured on the live index against `cannotPromote`, which is sparse the same way:

| filter | hits |
|---|---|
| `cannotPromote EXISTS` | 1,853 |
| `cannotPromote = true` | 1,115 |
| `NOT cannotPromote = true` | at least 100,000 |

If a negation skipped absent documents, that last number could not have exceeded 1,853 minus 1,115 = 738.

## Two admin endpoints, both inert by default

**`queue-paid-models-reindex`** queues the gated model ids for a document rewrite. Only gated documents need it.

It reports `landed` by **reading the queue back**, not by the enqueue returning: `addToQueue` fails open — on a degraded sysRedis it parks the ids in Postgres and returns `false` — and neither wrapper propagates that boolean. `landed === gatedModelCount` proves the ids are queued. A lower number is **ambiguous**, because the 15-minute sync may have checked the queue out destructively in between; it means re-run, not diagnose.

**`apply-models-index-filterable-attributes`** applies the filterable list to the live index.

It exists because the step "add the attribute to `modelsFilterableAttributes`" had no mechanism behind it. Each index's `onIndexSetup` is the only in-repo caller of `updateFilterableAttributes`, and it runs in exactly one place — `reset()` in `base.search-index.ts` — against the `_NEW` **swap** index. Adding an attribute to the list is **inert on the live index** until a full ~705K-document rebuild, or this.

The call **replaces** the list rather than merging, so it refuses two ways:

- a write that would **remove** an attribute the live index has refuses without `allowRemove`. Drift is not directional, and removing a filterable attribute makes every query using it answer 400. A point-in-time read at 19:16Z showed live 21 attributes against 22 desired, added `hasActivePaidAccess`, removed none — but that is true of one read, not of run time, which is what the refusal is for.
- a write while a settings task is **already enqueued** refuses without `force`. The call returns on *enqueue* while `getFilterableAttributes` reports *applied*, and `WebhookEndpoint` has no method guard, so a browser reload would otherwise fire a second full reindex.

It uses `searchClient.index(name)` rather than `getOrCreateIndex`, so a typo is an error rather than a new empty index.

**The environment it would land in**, measured on prod at 19:16Z: `models_v9` holds **705,068 documents**; task 1742909 was received 16:40:21Z and did not start until **17:27:19Z**, a 46m58s queue wait; **522 enqueued, 5 processing, 563 failed** globally; the newest *succeeded* `models_v9` task was 16:40Z, so the models index had landed nothing for about 2h40m. The duration of the settings task itself remains unmeasured — nobody has run one. All point-in-time: re-read before running rather than quoting these forward.

Neither endpoint is reachable from a startup hook, a cron, a setup path, or a test against a live client, and `dryRun` defaults to `true` on both.

**`apply-models-index-filterable-attributes` has never been run, in either mode.**

**The backfill, however, is already done** — and not by the endpoint in this PR, which is not deployed. On **2026-09-08 at 19:46Z** I queued the 3,180 gated model ids through `/api/internal/search-index-update`, which is already on `main` and shipped in 5.1.80. Four POSTs of ≤1000 ids, all 200. The 15-minute sync drained the queue at 19:46:12–19:46:14 and Meilisearch enqueued four `documentAdditionOrUpdate` tasks.

⚠️ **Do not re-run the backfill after merging.** The rollout below reflects what is actually left.

## Rollout

~~1. backfill the gated documents~~ — **DONE 2026-09-08 19:46Z**, via `/api/internal/search-index-update`, 3,180 ids. Four Meilisearch tasks enqueued; at the time of writing they had not yet started.

Remaining:

1. confirm the four `documentAdditionOrUpdate` tasks reached `succeeded`, and that sampled gated documents now carry `hasActivePaidAccess`
2. merge, deploy
3. `apply-models-index-filterable-attributes?dryRun=false`, poll the returned `taskUid` until **applied** (not merely enqueued)
4. read `getFilterableAttributes` back and see the attribute
5. merge `civitai/flipt-state` #83, then turn `paid-model-search-filter` on

`queue-paid-models-reindex` is **not** a pending step. It stays because the set of gated models changes with the wall clock — `endsAt` is in the predicate — so this is a reconciler rather than a one-shot, and because it is the only thing that reads the queue back. Nine other `admin/temp` endpoints queue search-index updates and every one of them reports success from the call returning, which `addToQueue`'s fail-open makes worthless. Run it `dryRun=true` after deploy if you want a second reading on tonight's backfill — but note `alreadyQueued` is computed against a queue the sync drains destructively, so it reads near zero more than a tick after any enqueue regardless of what happened.

Steps 2 and 4 are held behind a search-stability gate: prod Meilisearch has flapped repeatedly today, most recently an ECONNRESET/408 episode from 19:33:28Z to 19:37:33Z, confirmed by an external 503 in the same minute.

## Tests, with the revert measured for each

| mutation | result |
|---|---|
| clause to `hasActivePaidAccess = false` | 3 failed, 19 passed (22) |
| clause to `hasActivePaidAccess != true` (present-only) | 3 failed, 19 passed (22) |
| flag gate dropped from the helper | 1 failed |
| checkbox default flipped to ticked | 1 failed |
| clause deleted from the filters array | 1 failed |
| `hasActivePaidAccess` removed from the filterable list | 2 failed |
| apply writes only the missing attrs, not the whole list | 1 failed |
| apply's removal refusal deleted | 1 failed |
| apply's pending-task refusal deleted | 1 failed |
| `dryRun` default flipped, either endpoint | 1 failed each |
| `WebhookEndpoint` wrapper removed, either endpoint | 1 failed each |
| `slice(i, chunkSize)` slip in the chunking loop | 1 failed |
| `landed` inferred from the call, not the read-back | `expected 3 to be +0` |
| queue endpoint switched to the cached `getGatedModelIds` | 5 failed |

Seventeen mutations, all red. One was **green** on the first pass — flipping the checkbox default to ticked, which would have silently removed paid models from everyone's search — and is now pinned.

## Verification

- `pnpm exec eslint` on every touched file — 0 errors
- Full unit suite — see comment below

## Known and deliberate

`hidePaid` is component-local state, so it is absent from the URL and `ClearRefinements` does not clear it, where the feed's equivalent is URL-persisted. The half a user actually notices is **browser-back**: every other control in that sidebar is restored from the URL on a back-navigation into search, and this one resets to unticked, so the paid models reappear. The same-named control therefore behaves two ways — linkable and clearable on `/models`, neither on `/search/models`. Raised by two reviewers; left as-is for this PR because moving it into InstantSearch's `uiState` is a separate change to the search routing layer, and `ToggleRefinement` cannot express this filter — it would force the equality-on-`false` bug above.

**Staleness, stated for the record.** Within the search page the card and the filter read the *same* document attribute, so they cannot disagree — a stale document is stale in both directions at once, and the "card says Paid but Hide Paid left it on screen" divergence this ticket exists to close cannot occur there. What remains is cross-surface: search can lag the model page and the feed by the index-sync latency, which is normally ~1 minute for the expiry job plus up to 15 for the sync, but is in practice bounded by the Meilisearch backlog — measured at 46m58s of queue wait today.

https://claude.ai/code/session_01T7ZW9caRhbAUDQGTXxGwgn

### DIFF
--- a/src/components/Search/__tests__/hide-paid-search-filter-negation.test.ts
+++ b/src/components/Search/__tests__/hide-paid-search-filter-negation.test.ts
@@ -1,7 +1,10 @@
 import { readFileSync, readdirSync, statSync } from 'fs';
 import path from 'path';
 import { describe, expect, it } from 'vitest';
-import { HIDE_PAID_MODELS_FILTER } from '~/components/Search/paid-model-search-filter';
+import {
+  HIDE_PAID_MODELS_FILTER,
+  paidModelsSearchFilterClause,
+} from '~/components/Search/paid-model-search-filter';
 import { joinFilterClauses } from '~/components/Search/search-filters';
 import { MODELS_SEARCH_INDEX } from '~/server/common/constants';
 import { filterableAttributesByIndex } from '~/server/search-index/filterable-attributes';
@@ -25,7 +28,7 @@ import { filterableAttributesByIndex } from '~/server/search-index/filterable-at
 const repoRoot = path.resolve(__dirname, '../../../..');
 const CALL_SITE = 'src/pages/search/models.tsx';
 
-// This file states the forbidden spelling in order to forbid it, so it is the one exemption.
+// This file states the forbidden spellings in order to forbid them, so it is the one exemption.
 // Exactly one, asserted below: widening this list is the edit that must be visible in a diff.
 const ALLOWLIST = [
   'src/components/Search/__tests__/hide-paid-search-filter-negation.test.ts',
@@ -33,14 +36,14 @@ const ALLOWLIST = [
 
 describe('hide-paid search filter', () => {
   it('is a negation, not an equality on false', () => {
-    expect(HIDE_PAID_MODELS_FILTER).toMatch(/^NOT\s/);
-    expect(HIDE_PAID_MODELS_FILTER).toContain('hasActivePaidAccess');
-    expect(HIDE_PAID_MODELS_FILTER).not.toContain('false');
+    // The exact string is the load-bearing assertion, not the shape checks around it: `NOT
+    // hasActivePaidAccess EXISTS` would satisfy every property below and hide the wrong documents.
+    expect(HIDE_PAID_MODELS_FILTER).toBe('NOT hasActivePaidAccess = true');
   });
 
   it('survives the parenthesising every clause gets before it reaches Meilisearch', () => {
-    // joinFilterClauses wraps each clause in `(...)`. A clause that is only valid unwrapped would
-    // fail as a 400 at runtime and never in a unit test that looked at the constant alone.
+    // joinFilterClauses wraps each clause in `(...)`. A clause only valid unwrapped would fail as a
+    // 400 at runtime and never in a unit test that looked at the constant alone.
     expect(joinFilterClauses([HIDE_PAID_MODELS_FILTER])).toBe('(NOT hasActivePaidAccess = true)');
   });
 
@@ -48,6 +51,23 @@ describe('hide-paid search filter', () => {
     // Without this the whole models search answers 400 invalid_search_filter the moment anyone
     // ticks the box — not just the paid filter, the entire result set.
     expect(filterableAttributesByIndex[MODELS_SEARCH_INDEX]).toContain('hasActivePaidAccess');
+  });
+});
+
+describe('paidModelsSearchFilterClause', () => {
+  it('applies the clause only when the flag is on AND the box is ticked', () => {
+    expect(paidModelsSearchFilterClause(true, true)).toBe(HIDE_PAID_MODELS_FILTER);
+  });
+
+  it('emits nothing when the feature flag is off, whatever the box says', () => {
+    // Dropping the flag from the gate would ship the filter to everyone before the index settings
+    // and the backfill land, which breaks the entire models search rather than just this filter.
+    expect(paidModelsSearchFilterClause(false, true)).toBeNull();
+  });
+
+  it('emits nothing when the box is unticked — paid models are shown by DEFAULT', () => {
+    // A default of `true` would silently remove paid models from everyone's search.
+    expect(paidModelsSearchFilterClause(true, false)).toBeNull();
   });
 });
 
@@ -61,27 +81,42 @@ function walk(dir: string, out: string[] = []) {
   return out;
 }
 
-describe('no equality-on-false spelling anywhere', () => {
-  it('src/ never writes `hasActivePaidAccess = false`', () => {
+// Each alternative is a spelling that is valid Meilisearch and matches only documents that HAVE the
+// attribute. `= false` is the obvious one; `!= true` reads as the opposite but excludes absent
+// documents the same way, and the attribute may be quoted.
+const FORBIDDEN = /"?hasActivePaidAccess"?\s*(=\s*false|!=\s*true)/;
+
+describe('no equality-on-present-only spelling anywhere', () => {
+  it('src/ never writes hasActivePaidAccess as an equality that skips absent documents', () => {
     const offenders = walk(path.join(repoRoot, 'src'))
-      .filter((file) => /hasActivePaidAccess\s*=\s*false/.test(readFileSync(file, 'utf8')))
+      .filter((file) => FORBIDDEN.test(readFileSync(file, 'utf8')))
       .map((file) => path.relative(repoRoot, file).split(path.sep).join('/'))
       .filter((file) => !ALLOWLIST.includes(file as (typeof ALLOWLIST)[number]));
 
     expect(ALLOWLIST).toHaveLength(1);
-
     expect(
       offenders,
       'Use a NOT clause. An equality does not match documents that lack the attribute.'
     ).toEqual([]);
   });
 
-  it('the call site uses the shared constant rather than an inline string', () => {
-    // The constant carries the reasoning. Inlining the string is how the reasoning gets lost and
-    // the next edit reaches for `= false`.
+  it('the call site actually WIRES the clause in, not merely imports it', () => {
+    // `toContain('HIDE_PAID_MODELS_FILTER')` was satisfied by the import line alone, so deleting the
+    // clause from the filters array left every suite green while the checkbox did nothing. Pin the
+    // call, which the import cannot satisfy.
     const source = readFileSync(path.join(repoRoot, CALL_SITE), 'utf8');
 
-    expect(source).toContain('HIDE_PAID_MODELS_FILTER');
-    expect(source).not.toContain("'NOT hasActivePaidAccess");
+    expect(source).toContain(
+      'paidModelsSearchFilterClause(features.paidModelSearchFilter, hidePaid)'
+    );
+    expect(source).not.toMatch(/['"`]NOT hasActivePaidAccess/);
+  });
+
+  it('shows paid models by DEFAULT — the box starts unticked', () => {
+    // The helper test covers "unticked emits nothing", but nothing read the initial value, so
+    // useState(true) — every user's model search silently losing paid models — was invisible.
+    const source = readFileSync(path.join(repoRoot, CALL_SITE), 'utf8');
+
+    expect(source).toContain('const [hidePaid, setHidePaid] = useState(false)');
   });
 });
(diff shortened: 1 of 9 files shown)


JustMaier:
Full unit suite on `7ed67fca8c`, `--max-workers=4`:

```
Test Files  1 failed | 1696 passed | 3 skipped (1700)
     Tests  1 failed | 39059 passed | 35 skipped (39095)
```

Totals account for every file. **The one failure is not from this branch.** `no-direct-shared-module-mock` flags `src/server/routers/__tests__/article.router.scan-status-authz.test.ts`, added today by `1d8ad23785`, which is in `origin/main` and is not one of the nine files this PR touches — so `main` is red on that guard independently of this change.

That guard is not blind to new files: earlier this evening it flagged a test file of mine on the other branch, and I ran the codemod for it. Here it is correctly pointing somewhere else.

Submodule canary:

```
Test Files  1 passed (1)
     Tests  370 passed (370)     # blocks.router.workflow.test.ts
```

Nonzero, so the worktree's run means something — that file collects **zero** tests without `event-engine-common` initialised, and a zero-collection run still reads as a pass.

JustMaier:
Updated to `1d2a022024` — a **child** of `7ed67fca8c`, pushed as a fast-forward, so the five review lanes' findings still resolve against the commit each was pinned to.

Full unit suite on `1d2a022024`, `--max-workers=4`:

```
Test Files  1 failed | 1696 passed | 3 skipped (1700)
     Tests  1 failed | 39070 passed | 35 skipped (39106)
```

Totals account for every file. **The one failure is not from this branch.** `no-direct-shared-module-mock` flags `src/server/routers/__tests__/article.router.scan-status-authz.test.ts`, added to `main` today by `1d8ad23785`, which is not among this PR's files — so `main` is red on that ratchet independently. That guard is not blind to new files: it flagged a test file of mine earlier this evening on the other branch and I ran the codemod for it.

Submodule canary:

```
Test Files  1 passed (1)
     Tests  370 passed (370)     # blocks.router.workflow.test.ts
```

Nonzero, so the run means something — that file collects **zero** tests without `event-engine-common` initialised, and a zero-collection run still reads as a pass.

## What the review round changed

Five lanes reviewed `7ed67fca8c`; all five reported the SHA and a clean tree, except the perf lane, which correctly reported the tree had gone dirty **while it was reading** (my fault — I started fixing before it finished) and refused to attribute its findings to code it had not read.

Confirmed and fixed:

- **`updateFilterableAttributes` replaces rather than merges**, so the endpoint could delete filterable attributes the live index has. Now refuses without `allowRemove`. Separately, the call returns on *enqueue* while the read reports *applied*, and the route has no method guard — a browser reload would have queued a second full 705K-document reindex. Now refuses without `force`.
- **A test passed on the mutation that breaks all of model search.** `toContain` cannot separate `updateFilterableAttributes(desired)` from `updateFilterableAttributes(missing)` on a replace-semantics write. Now `toEqual`.
- **The feature was deletable with every test green** — the call-site guard was satisfied by the import line alone. The clause is now a function whose call is pinned, and the flag gate and the default have their own tests.
- **Neither endpoint suite proved a bad token is rejected**; deleting `WebhookEndpoint` reddened nothing. Both now assert 401 and that nothing was written.
- The chunking loop's `slice(i, chunkSize)` slip was invisible at the default chunk size. Now covered.
- `before.includes` on the dry-run path was O(n·m) against a queue documented as reaching ~211K ids. Now a `Set`.
- The `landed` comment claimed a low count meant sysRedis dropped ids. It can also mean the 15-minute sync checked the queue out in between, so it now says equal-to-count is proof and below-count means re-run.

Refuted, after reading the code rather than accepting the claim: one lane reported that nothing re-indexes a model when its gate starts or expires. Gate writes reach `modelsSearchIndex.queueUpdate` through `writeModelVersionGateAndGoal` → `bustMvCache` → `bustModelLevelVersionCaches`, and timed expiry is handled by `process-ending-early-access`, which queues the affected model ids. The lane's grep was scoped to `paid-access.service.ts`, where the invalidation does not live.

Not changed, deliberately: `chunkSize` stays 1000 (the argument for 10000 was about fail-open windows rather than speed, and a partial park is recoverable either way), and `hidePaid` stays component-local — see **Known and deliberate** in the description.

JustMaier:
Second review round. Updated to `07fcdec231`, a fast-forward child of `1d2a022024`, so both previously-reviewed SHAs still resolve.

Five lanes re-read `1d2a022024`; all five confirmed the SHA and a clean tree. **Reuse, safety and perf: no findings.** Tests found seven, intent found five. Eleven mutations fixed, every one of which was green before.

## The pattern the tests lane found

I had asserted that mocks were **called** without asserting **what they were called with** — and the unasserted arguments were the consequential ones.

| mutation | production consequence | now |
|---|---|---|
| drop `readOnly` from `getQueue(...)` | checks the models queue out **destructively** on every call, dry run included, consuming work the 15-minute sync was about to take | red |
| point `getQueue` at another index, or at the Delete queue | `landed` permanently 0 | red |
| `landed = after.length` | both fixtures had made queue-depth and our-ids-landed the same number, so the landed-3/landed-0 pair only *looked* like a control | red |
| retarget `searchClient.index(...)` | writes the models list onto another index, wiping its filterable attributes | red |
| point `getTasks` at another task type, or drop `processing` | the double-reindex guard silently stops guarding while its 409 test keeps passing | red |
| `updateFilterableAttributes(missing)` | strips every other filterable attribute | red |
| `allowRemove` writing `[...desired, ...extra]` | reports `removed` while removing nothing; operator re-runs forever | red |
| negate `alreadyQueued` | tells an operator the work is already done | red |

## From the safety lane

`pendingSettingsTaskUids` now returns `null` rather than throwing when the tasks API is unreachable. That call newly requires `tasks.get` on `SEARCH_API_KEY`, and an unguarded throw took out the **dry run** as well as the write. `null` is deliberately distinct from `[]` — *could not determine* versus *checked, nothing pending* — and the write path refuses on not-knowing.

## From the intent lane

The description was factually wrong and has been corrected: it claimed the backfill had not been run, and its rollout step 2 told the next operator to run it. It **was** run, at 19:46Z, through `/api/internal/search-index-update`. Following the old runbook would have re-queued 3,180 ids.

## Verification

- `pnpm exec eslint` on every touched file — 0 errors
- Full unit suite on `07fcdec231`, `--max-workers=4`:

```
Test Files  1 failed | 1696 passed | 3 skipped (1700)
     Tests  1 failed | 39076 passed | 35 skipped (39112)
```

The single failure is `article.router.scan-status-authz.test.ts`, added to `main` today by `1d8ad23785` and not among this PR's files — `main` is red on that ratchet independently.

- Canary: `Tests 370 passed (370)` on `blocks.router.workflow.test.ts`, so the submodule is initialised and the run means something.

**28 mutation controls across both rounds.** One was green on the first pass — flipping the checkbox default to ticked — and is now pinned.

PR #4704, opened 2026-09-08
fix(search-index): surface updateSync task failures instead of reporting success

ZacxDev:
## The problem

A production backfill submitted through `GET /api/mod/update-index` reported success while ~97% of the work failed on a database statement timeout. Every call returned HTTP 200 `{"status":"ok"}`; in reality only one of the five batches indexed anything, and ~1,200 of ~1,239 documents were never written. Nothing in the response, the status code, or the caller's view distinguished that from a clean run.

The silence is layered, and every layer has to be crossed for the caller to see anything:

- `processSearchIndexTask` catches the exception, logs it, and returns the string `'error'`.
- `updateSync` runs its workers with `await Promise.all(workers)` and then returns `undefined` unconditionally — it never asks whether any task ended in `'error'`.
- The endpoint awaits `updateSync(...)` and unconditionally sends `200 {"status":"ok"}`.

## What this changes

**1. `updateSync` reports its outcome.** It now resolves to `{ indexName, totalTasks, failedTasks, failedIds }` instead of `undefined`. This builds on what the queue already does rather than adding a parallel mechanism: `getTaskQueueWorker` already routes an `'error'` result to `TaskQueue.failTask`, which already counts a permanently-failed task; the queue now also keeps the failed tasks so the caller can be told *how much* was lost, not just that something was. Existing callers (`refresh-auction-cache`, `handle-auctions`, `reindex-recent-scheduled-images`) ignore the return value and are unaffected.

Two supporting fixes in `TaskQueue`, both of which the count depends on to be truthful:

- **`failTask` could drop a retry entirely.** It removed the task from `processing` and *then* awaited the retry backoff. During that window `isQueueEmpty()` returned true, so with a single in-flight task every worker exited before the task was re-queued — the retry never ran, and because `stats.failed` is only incremented once retries are exhausted, nothing was ever recorded as failed. This is exactly the shape of the production incident: one batch per call, one failure, zero retries, zero failures recorded. A `retrying` counter now keeps the task visible to `isQueueEmpty()` across the backoff, and the worker awaits `failTask` rather than dropping its promise.
- **A failure after the pull step lost its id count.** The transform and push tasks no longer carry the id list, so tasks now carry an `idCount` through the `pull -> transform -> push` handoff and a failure at any step can still be attributed to a number of documents.

**2. The endpoint reports it.** If any batch failed, the response is a non-2xx carrying the index name and the failed batch/id counts:

```json
{ "status": "error", "index": "collections_v3", "failedTasks": 4, "totalTasks": 5,
  "failedIds": 1200, "error": "1200 ids in 4 of 5 batches failed to index" }
```

The all-succeeded response is byte-for-byte what it was (`200 {"status":"ok"}`), so existing callers see no change on the happy path, and a `processQueues` run — which produces no sync result — is also unchanged. The nine-arm `switch` became a lookup keyed by the same zod enum (checked against the lookup's keys with `satisfies`, so the two cannot drift), which is what lets the result be captured in one place instead of nine.

**3. The sync chunk size is per-index.** `updateSync` hardcoded `chunk(items, 500)`. It is now an optional `updateSyncChunkSize` on the processor, defaulting to 500 — so every index behaves exactly as before unless it opts in.

The collections index opts in at **25**. Measured against production data for the affected collections: a 25-id targeted pull runs in **~0.18s**, but the query planner flips around 50 ids and a 50-id pull already costs **~1.41s** — essentially the same as a 300-id pull at **~1.45s**. So the cost is not linear in batch size; what matters is staying on the fast side of that flip. Against a ~10s statement timeout that these queries demonstrably hit under normal load, 25 leaves roughly two orders of magnitude of headroom, where 300 and 500 do not.

## Adjacent, deliberately not fixed here

`processQueues` shares `processSearchIndexTask` and calls `queuedUpdates.commit()` after `await Promise.all(workers)` regardless of whether any task errored — so a failed background batch clears its queued ids as well, and those ids are not retried by a later run. That is a real second instance of the same silent-failure shape, but it is a background cron with a different blast radius and a different fix (the commit has to become conditional, and the ids of failed tasks have to be preserved rather than merely counted). Worth a reviewer knowing about; deliberately out of scope for this PR.

## Tests

`src/server/search-index/__tests__/update-sync-failure-reporting.test.ts` and `src/__tests__/pages/api/mod/update-index.test.ts`.

The regression specs were run against the parent commit (`87f3073240`, with only the test files added) and watched fail — 8 failed / 3 passed:

- `reports the failed batch when the pull step throws, instead of resolving silently` — `Cannot read properties of undefined`, i.e. `updateSync` resolved to nothing. Its wall time there was **1.0s**, not the ~4s of four attempts, which is the dropped-retry bug showing up directly in the timing.
- `attributes a failure at the push step back to the ids that were never written`
- `reports zero failures when every batch succeeds`
- `returns an all-zero result for an empty item list`
- `splits into batches of the configured size` / `falls back to the default chunk size when a processor sets none` / `keeps the collections index well below the default`
- `does NOT return 200 ok when a batch failed to index` — `expected 200 not to be 200`

The 3 that passed at the parent commit are the ones that pin *unchanged* behaviour (the 200 success shape, id routing, the `processQueues` path) — invariant guards, not regression coverage.

The batching spec uses 60 ids against a chunk size of 25, deliberately not a multiple, so it asserts `[25, 25, 10]` — an off-by-one in the batching changes the final batch rather than hiding in a round number.

All 11 pass on this branch.

### DIFF
--- a/src/__tests__/pages/api/mod/update-index.test.ts
+++ b/src/__tests__/pages/api/mod/update-index.test.ts
@@ -42,7 +42,8 @@ vi.mock('~/server/jobs/job', () => ({
   inJobContext: (_res: NextApiResponse, fn: (jobContext: unknown) => Promise<void>) => fn({}),
 }));
 
-// The real index objects open a Meilisearch client and hit the database at module load.
+// Replaced with spies because which of these objects the handler reaches for is the thing under
+// test; the real ones would talk to Meilisearch and the database once called.
 vi.mock('~/server/search-index', () => indexMocks);
 
 import {
--- a/src/__tests__/pages/api/mod/update-index.test.ts
+++ b/src/__tests__/pages/api/mod/update-index.test.ts
@@ -1,10 +1,37 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-const { updateSync, processQueues } = vi.hoisted(() => ({
-  updateSync: vi.fn(),
-  processQueues: vi.fn().mockResolvedValue(undefined),
-}));
+/**
+ * One mock per export, each with its OWN spies. Sharing a single object across all nine made the
+ * lookup table in the handler unobservable: any index could be mapped to any other and every
+ * assertion still passed.
+ */
+const { indexMocks } = vi.hoisted(() => {
+  const exportNames = [
+    'modelsSearchIndex',
+    'usersSearchIndex',
+    'imagesSearchIndex',
+    'articlesSearchIndex',
+    'imagesMetricsSearchIndex',
+    'collectionsSearchIndex',
+    'bountiesSearchIndex',
+    'toolsSearchIndex',
+    'comicsSearchIndex',
+  ] as const;
+
+  type ExportName = (typeof exportNames)[number];
+  const mocks = {} as Record<
+    ExportName,
+    { updateSync: ReturnType<typeof vi.fn>; processQueues: ReturnType<typeof vi.fn> }
+  >;
+  for (const name of exportNames) {
+    mocks[name] = { updateSync: vi.fn(), processQueues: vi.fn() };
+  }
+  return { indexMocks: mocks };
+});
+
+type IndexExportName = keyof typeof indexMocks;
+const indexExportNames = Object.keys(indexMocks) as IndexExportName[];
 
 // ModEndpoint wraps the handler in mod-auth; the handler itself is what is under test.
 vi.mock('~/server/utils/endpoint-helpers', () => ({
@@ -16,23 +43,40 @@ vi.mock('~/server/jobs/job', () => ({
 }));
 
 // The real index objects open a Meilisearch client and hit the database at module load.
-vi.mock('~/server/search-index', () => {
-  const index = { updateSync, processQueues };
-  return {
-    modelsSearchIndex: index,
-    usersSearchIndex: index,
-    imagesSearchIndex: index,
-    articlesSearchIndex: index,
-    imagesMetricsSearchIndex: index,
-    collectionsSearchIndex: index,
-    bountiesSearchIndex: index,
-    toolsSearchIndex: index,
-    comicsSearchIndex: index,
-  };
-});
+vi.mock('~/server/search-index', () => indexMocks);
 
-import { COLLECTIONS_SEARCH_INDEX } from '~/server/common/constants';
+import {
+  ARTICLES_SEARCH_INDEX,
+  BOUNTIES_SEARCH_INDEX,
+  COLLECTIONS_SEARCH_INDEX,
+  COMICS_SEARCH_INDEX,
+  IMAGES_SEARCH_INDEX,
+  METRICS_IMAGES_SEARCH_INDEX,
+  MODELS_SEARCH_INDEX,
+  TOOLS_SEARCH_INDEX,
+  USERS_SEARCH_INDEX,
+} from '~/server/common/constants';
 import handler from '~/pages/api/mod/update-index';
+
+/**
+ * The routing this file pins: every `index` query value and the module export that must handle
+ * it. Written out by hand rather than derived from the handler, so a copy-paste slip in the
+ * handler's lookup table disagrees with it instead of being mirrored by it.
+ */
+const expectedRouting: Array<{ index: string; exportName: IndexExportName }> = [
+  { index: MODELS_SEARCH_INDEX, exportName: 'modelsSearchIndex' },
+  { index: USERS_SEARCH_INDEX, exportName: 'usersSearchIndex' },
+  { index: IMAGES_SEARCH_INDEX, exportName: 'imagesSearchIndex' },
+  { index: ARTICLES_SEARCH_INDEX, exportName: 'articlesSearchIndex' },
+  { index: METRICS_IMAGES_SEARCH_INDEX, exportName: 'imagesMetricsSearchIndex' },
+  { index: COLLECTIONS_SEARCH_INDEX, exportName: 'collectionsSearchIndex' },
+  { index: BOUNTIES_SEARCH_INDEX, exportName: 'bountiesSearchIndex' },
+  { index: TOOLS_SEARCH_INDEX, exportName: 'toolsSearchIndex' },
+  { index: COMICS_SEARCH_INDEX, exportName: 'comicsSearchIndex' },
+];
+
+const updateSync = indexMocks.collectionsSearchIndex.updateSync;
+const processQueues = indexMocks.collectionsSearchIndex.processQueues;
 
 const runRequest = async (query: Record<string, string>) => {
   const req = {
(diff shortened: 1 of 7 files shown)


ZacxDev:
Note on CI: `Unit tests (2)` is red here for a reason unrelated to this PR. The failure is `no-direct-shared-module-mock > adds no new direct mock of a module that has a canonical mock`, and the file it names is `src/server/routers/__tests__/article.router.scan-status-authz.test.ts`, which this branch does not touch — it needs the generated allowlist regenerated.

The same shard already fails on `main` at this branch's base commit (`87f3073240`), with the same single failure, so it is a pre-existing red rather than something this change introduced. Every other check on this PR passes.

ZacxDev:
Round 1 adversarial audit ran against `f80aed4c2a` (0 blocking, 6 should-fix, 3 nits). Four were fixed in `f908072cd1`; the rest are recorded below as deliberately not addressed in this PR.

The audit independently reproduced this PR's red-then-green claim (8 failed / 3 passed with the payload files restored, matching the description), confirmed `failedIds` is reachable and branched on, and found no task lost at 0 tasks, 1 task, or fully-exhausted retries.

```audit-claims round=1 audited=f80aed4c2a..f908072c
1. failTask now retains a {type, idCount, retries} summary instead of the whole Task, so a permanently-failed batch no longer holds its document payload (or intermediate pull data) for the life of the queue.
2. The index-routing table in the mod update-index handler is now observable in tests: nine separate mocks with their own spies, asserting the correct export was called and no other one was.
3. failedIds is now exercised with a PARTIAL failure, so it can no longer be replaced by "every id in the run" without a test going red.
4. updateSyncChunkSize is clamped to a positive integer, so a 0 or negative value can no longer produce zero batches and report success with nothing written.
```

**Deliberately not addressed here** — recorded so they read as open rather than absent:

- `processQueues()` and `update()` both call `queuedUpdates.commit()` after `Promise.all(workers)` regardless of whether any task failed, so a failed batch still clears its queue entry and is never retried. `update()` is the one that runs on a schedule for collections. Unchanged by this PR, and now the only remaining path with the silent-drop shape this PR removes from `updateSync`.
- `updateSyncChunkSize: 25` also multiplies `pushData` and cleanup round-trips, not just the pull query; the field's doc comment describes only the query effect.
- The `satisfies` check proves every enum member has a handler, but not the reverse — a key added without an enum entry still compiles.
- `failTask` is now awaited inside a promise executor with no try/catch, so a throw there would hang rather than reject. Not reachable today.

ZacxDev:
Round 2 delta re-audit ran against `f80aed4c2a..f908072cd1`. All four round-1 fixes verified as landed — claim 1 by actual heap reachability (WeakRef + forced GC) with both controls, the other three by mutation. Round 2 then found 3 should-fix and 3 nits, all of them guards or comments narrower than the prose written around them rather than defects in shipped behaviour. Those are fixed in `ccf40f143b`.

`ccf40f143b` changes **no executable production line** — the only two source-file edits are comments.

```audit-claims round=2 audited=f908072cd1..ccf40f14
1. The holdsReference test walker now follows Set entries and Map keys and values, so it is no longer blind to the one structure on the queue that holds live tasks; its docstring now states what it does not follow.
2. The retention guard is now queue-wide (holdsReference against the queue itself), so a mutant that keeps the summary but also retains the whole task on another queue field is caught; the comment was reworded to state exactly what the guard pins rather than overclaiming.
3. NaN, Infinity and -Infinity are now covered cases for the updateSyncChunkSize clamp, with exact expected values instead of a >= 1 assertion that 1, 500 and Infinity all satisfy.
4. The clamp comment now says non-finite rather than non-numeric, and records that Infinity previously yielded one batch of everything.
5. The failedTasks docblock no longer claims only-reported-fields are kept, and documents retries as a retry count rather than an attempt count.
6. The document batch size quoted in the test rationale corrected from five figures to 100,000.
```

Still deliberately not addressed, unchanged from the round-1 comment above: the `processQueues()`/`update()` unconditional `commit()` after `Promise.all(workers)`; the `updateSyncChunkSize: 25` round-trip multiplication; the one-directional `satisfies`; and the `failTask`-throws hang shape.

One claim in this branch remains unverifiable from the repo: a test comment stating the batch that timed out in production was 300 ids. It is round-1 text, no artifact in the tree settles it, and it should not be read as confirmed.

ZacxDev:
Round 3 delta re-audit ran against `f908072cd1..ccf40f143b`. All six round-2 fixes verified as landed, two of them mutation-proven armed — notably the `auditTrail` mutant that **survived round 2's tests entirely** (4 passed) is **killed by round 3's** (3 failed, with the guard's own assertion message in ~1.5 s, not by timeout). Round 3 returned one finding: a test comment naming the wrong case as load-bearing. Fixed in `dae389dd54`, along with a full sweep of every comment the previous two commits wrote.

```audit-claims round=3 audited=ccf40f143b..dae389dd54
1. The comment naming which assertion is load-bearing for the chunk-size clamp now names Infinity rather than -Infinity, re-derived by running the finiteness mutant with the exact-size assertion removed and observing which rows survive.
2. The claim that a non-finite chunk size can arrive from config or an env var is deleted; the comment now states the guard is defensive with no caller reaching it, rather than substituting a new justification.
3. The claim that the value used to yield a single batch of everything is deleted; it described an intermediate commit inside this PR, not any released behaviour.
4. The list of chunk-size values that produce no batches now includes -Infinity, and names Infinity as the exception, measured against the vendored lodash-es.
5. The retention sentence in failTask is scoped to the give-up path, since the retry branch legitimately returns with the task back on the queue.
6. The updateSyncChunkSize docstring no longer claims each batch is one pullData call, which is false for any processor setting pullSteps.
7. Two comments crediting either the retrying counter or the awaited failTask as the load-bearing mechanism now state they are redundant: removing either alone leaves the suite green, and only removing both reproduces the original drop.
8. An unsupported claim that the real index modules open a client at module load was deleted after probing the import, not reworded.
```

**Ladder stopped here, and this is the reason** — recorded so this is distinguishable from a round that simply converged. The stop is the attribution gate, not a round cap: `ccf40f143b` and `dae389dd54` each changed **zero executable lines** (round 3 measured 26 changed lines / 0 non-comment; the final sweep measured 56 / 0, with a positive control of 666 non-comment lines on the PR's own full diff, so the zero is a real reading). Two consecutive fix rounds with zero payload change means the ladder is auditing its own scaffolding rather than the PR. Round 1 changed 20 executable lines and caught a genuine production hazard; rounds 2 and 3 found only prose, with decreasing severity.

**Known open, deliberately, and not addressed by any round:**

- Neither half of the retry fix is individually pinned. Measured over four shapes: `retrying` removed alone → suite green; the awaited `failTask` removed alone → suite green; both removed → 3 failed. They are redundant rather than alternatives, which is now stated in the comments instead of each crediting the other. A future "simplify" that removes both would restore the original silent drop with a green suite.
- `processQueues()` and `update()` still call `commit()` after `Promise.all(workers)` regardless of failure, so a failed batch clears its queue entry and is never retried. `update()` is the scheduled path.
- `updateSyncChunkSize: 25` also multiplies push and cleanup round-trips, not just the pull query.
- The `satisfies` check is one-directional.
- One comment in `collections.search-index.ts` records production timings (~0.18 s at 25 ids, a planner flip near 50, statement-timeout losses above it). Those were measured directly against the production primary, but nothing in this repo can re-derive them — treat them as an external measurement, not a repo-verifiable claim.

PR #4711, opened 2026-09-08
fix(indexes): rebuild owned documents when a user's avatar changes

manuelurenah:
Closes the residue that #4661 deliberately left out of scope — replacing an avatar left a dead `user.profilePicture` on every document the user owns — and the two siblings of it that turned up while verifying the first.

## The defect

Five search indexes ship `user.profilePicture`. Three of them denormalize it into every document the user owns and cannot notice when it changes: **collections**, **bounties**, **comics**. Each resolves the avatar by a separate fetch in its own `pullData` rather than by a join, and each `prepareBatches` filters on `"createdAt" >= lastUpdatedAt` — so an existing document is rebuilt only by an explicit enqueue. Nothing enqueued on an avatar change. The document keeps the old image, and 30 days later `remove-replaced-images` destroys it, leaving every card that user owns rendering a dead url.

(`users` is already enqueued unconditionally at both call sites. `articles` fetches the avatar live at query time and is not affected. `metrics_images_v1`, the feed, does not carry it.)

**#4661's avatar leg cannot fire on this path.** It matches `u."profilePictureId" IN (ids)` — images that are somebody's avatar *right now* — and `remove-replaced-images` calls `getStillReferencedImageIds` to drop every id that is still a `profilePictureId` before it calls `deleteImages`. On the replacement path it resolves zero collections, by construction, every time. That leg is untouched here and still needed.

## Evidence it is live

Of **120** indexed public collections owned by users with a pending `ReplacedImageDelete` row, **100 carry a `user.profilePictureId` that no longer matches the database**. The 20 fresh ones had been rebuilt incidentally by some unrelated enqueue — none by `createdAt`, since every sampled collection predates its owner's current avatar.

## The change

New leaf `src/server/services/owner-avatar-index.ts` — `queueOwnerAvatarReindex({ userId, source })` — called from the two places `User.profilePictureId` changes:

- **`updateUserHandler`**, when the id changes. That includes the *first* avatar a user sets: those documents ship `profilePicture: null` until something rebuilds them, so the picture is missing rather than stale.
- **`deleteUser`**, which nulls the column with its own `dbWrite.user.update` and leaves the account's collections and comics standing. (`removeAllContent` is the separate endpoint that deletes content; it removes collections and bounties, never a `ComicProject`.) None of the three indexes filters on `deletedAt`, so those documents kept an avatar `remove-deleted-user-images` later destroys.

Non-throwing by contract — it runs after a write that has already committed, so a throw would report a successful profile save, or a completed account deletion, as an error. In `deleteUser` the call site *also* catches: subscription cancellation and `invalidateSession` run below it, and those should not depend on a contract that lives in another file.

## Cost

| | max per user | index probe (prod replica) |
|---|---|---|
| collections | 8,987 | `Collection_userId_idx` — cost 6.77, 0.51 ms |
| bounties | 67 | `Bounty_userId_idx` — cost 3.57, 0.02 ms |
| comics | 48 | `ComicProject_userId_idx` — cost 2.50, 0.02 ms |

Each lookup is bounded in SQL at `LIMIT cap + 1`; past the cap `owner-avatar-index-truncated` says so rather than truncating silently. The enqueue is deliberately **unchunked** — `addToQueue` already splits at `QUEUE_ADD_CHUNK_SIZE = 10_000`, which is the cap here, so chunking above it would only turn one `sAdd` into many bucket-read round trips.

## Why `images` is excluded

`images` carries the same denormalized avatar and has the same sweep, so it is stale the same way — but its fan-out is per **image**, not per project. One sampled heavy uploader owns **831,541**. Queueing that behind a profile save is a six-figure rebuild; it needs its own mechanism (not a fourth entry in the list). A test pins the exclusion so it is not added casually.

## Tests

18 across three files. Watched failing on pre-change code: removing the `updateUserHandler` trigger fails 5 of 6 controller tests in ~4 ms; removing the `deleteUser` call fails both its tests in 6 ms and 2 ms; removing the `deleteUser` `.catch` fails its guard test in 3 ms. The negative controls (`queues nothing when the picture is unchanged`, `does not fan out to the images index`) correctly stay green throughout.

## Known gaps, recorded in the module header, not closed here

- **No backfill.** Documents already carrying a dead avatar stay stale until something else rebuilds them.
- **`Image.onDelete: SetNull`** clears `profilePictureId` in the database with no application code involved, so a moderator or ingestion delete of a *live* avatar reaches only #4661's collections leg — bounties and comics keep the dead url.
- **`/intent/avatar`** writes the legacy `User.image` column, which all three indexes also denormalize and no predicate here compares. Latent: one non-deleted prod account carries it, owning no indexed collection.
- **The pre-update value is read from the replica.** An A→B→A avatar change inside the replication window compares equal and skips the fan-out. The same handler documents this hazard for `leaderboardShowcase` 40 lines above and chooses unconditional work instead; here unconditional work would mean up to 8,987 enqueues on every profile save, so it is left as a deliberate trade rather than fixed silently.
- **The reaper is unchanged.** `remove-replaced-images` filtering out still-referenced ids is correct behaviour.

## Verifier

```
GET /indexes/collections_v3/documents/<collectionId>?fields=id,user
```

For a user who has changed their avatar, `user.profilePicture` must name the current image. Positive control that the query works and the field is populated: the same call for a collection whose owner has never changed their avatar returns a live `user.profilePicture`.

ClickUp: https://app.clickup.com/t/868m22z20

https://claude.ai/code/session_01SbBei7KWcAqD5K7e5oiPLk

### DIFF
--- a/src/server/controllers/__tests__/user.controller.avatar-collection-reindex.test.ts
+++ b/src/server/controllers/__tests__/user.controller.avatar-collection-reindex.test.ts
@@ -1,0 +1,144 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type * as UserService from '~/server/services/user.service';
+
+/**
+ * Changing an avatar must re-index the collections that user owns — see the route-7 note in
+ * collection-media-index.ts for why the removal-side leg cannot cover a replacement.
+ */
+
+const { mockUpdateUserById, mockGetUserById, mockIngestImage, mockCollectionsQueueUpdate } =
+  vi.hoisted(() => ({
+    mockUpdateUserById: vi.fn(),
+    mockGetUserById: vi.fn(),
+    mockIngestImage: vi.fn(),
+    mockCollectionsQueueUpdate: vi.fn(),
+  }));
+
+vi.mock('~/server/services/orchestrator/civitai', () => ({
+  invalidateCivitaiUser: vi.fn().mockResolvedValue(undefined),
+}));
+vi.mock('~/utils/signal-client', () => ({
+  signalClient: { send: vi.fn().mockResolvedValue(undefined) },
+}));
+vi.mock('~/server/services/user.service', async (importOriginal) => ({
+  ...(await importOriginal<typeof UserService>()),
+  getUserById: mockGetUserById,
+  updateUserById: mockUpdateUserById,
+  updateLeaderboardRankForUsers: vi.fn(),
+  equipCosmetic: vi.fn(),
+  unequipCosmeticByType: vi.fn(),
+  createUserReferral: vi.fn(),
+  isUsernamePermitted: vi.fn(async () => true),
+  queueModelMetricPrivacyReindex: vi.fn(),
+}));
+vi.mock('~/server/services/image.service', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  ingestImage: mockIngestImage,
+  deleteImageById: vi.fn(),
+  deleteImages: vi.fn(),
+  queueReplacedImageDeletion: vi.fn(),
+}));
+vi.mock('~/server/search-index', () => ({
+  usersSearchIndex: { queueUpdate: vi.fn() },
+  collectionsSearchIndex: { queueUpdate: mockCollectionsQueueUpdate },
+}));
+vi.mock('~/server/cloudflare/client', () => ({ purgeCache: vi.fn(() => ({ catch: vi.fn() })) }));
+
+import { dbMock } from '~/__tests__/mocks/db.mock';
+import { loggingMock } from '~/__tests__/mocks/logging.mock';
+import { SearchIndexUpdateQueueAction } from '~/server/common/enums';
+import { updateUserHandler } from '~/server/controllers/user.controller';
+
+const USER_ID = 5;
+const OLD_PICTURE_ID = 42;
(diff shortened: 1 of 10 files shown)


manuelurenah:
Closing: this treats the symptom, not the cause.

Investigation while reviewing it found the fan-out reaches only 2 of the 5 indexes that denormalize the avatar, and misses the two largest:

| index | docs | max/user | avatar staleness sampled |
|---|---|---|---|
| images_v6 | 63,839,039 | 831,541 | not sampled |
| models_v9 | 705,747 | 14,372 | 69 / 120 stale |
| collections_v3 | 95,907 | 8,987 | 100 / 120 stale |
| bounties_v3 | 9,096 | 67 | — |
| comics_v1 | 0 | 48 | index is empty, so this leg was inert |

`models_v9` carries the full nested `profilePicture` blob and was missed here. Neither models nor images can be fixed by enqueueing — 14,372 models for one owner already exceeds this PR's 10,000 cap, and 831,541 images is not a fan-out at all.

The real fix is to stop storing display-only user data in the documents and hydrate it at read time through a server-side search proxy, which is the pattern `getAllImagesIndex` already uses for the image feed. Tracked on the ClickUp task; a plan doc follows.

The defect described in 868m22z20 is real and still open — this just isn't the shape of its fix.

PR #4716, opened 2026-09-09
fix(clickhouse): clamp pageView numeric fields so an outlier stops destroying the row

ZacxDev:
## The bug

`pageViews` is the last table still losing rows to the ClickHouse tracker's silent client-side rejection — and unlike the six enum gaps before it, the cause is **integer width overflow**.

A value too wide for its column is rejected while the batch is serialized, *before* ClickHouse is asked. The POST is fire-and-forget and returns success, the app logs nothing, `SHOW CREATE TABLE` looks fine, and the row simply never exists.

Measured on production by reading the rejected rows out of the dead-letter queue — **26 of 26 sampled rows explained, no residual**:

| cause | n | column | ceiling | observed |
|---|---|---|---|---|
| `duration` over `UInt32` | 21 | `UInt32` | 4,294,967,295 | 4.32e9 – 1.24e10 (≈50–144 **days** in ms) |
| `windowWidth`/`windowHeight` over `Int16` | 5 | `Int16` | 32,767 | up to 183,800 × 102,200 px |

The corroborating tell: the live `duration` maximum sat at **4,264,810,774 — 99.3% of the ceiling and never above it**. A maximum pinned just under a type bound is the signature of a silently clipped tail.

## Why clamp

**Not widen the column.** `UInt32`→`UInt64` rewrites data on a table taking ~55.7M rows/7d, to preserve values that are junk on their face (a 144-day "page view", a 183,800-pixel window).

**Not drop the row.** The outlier is *one metric* on an otherwise-good row — `path`, `host`, `country`, `userId` are all fine — so saturating that field keeps the page view countable.

Saturation is also honest here rather than lossy: the stored distribution was **already** clipped at this bound, just by silent loss instead of by an explicit clamp.

## One design detail worth reviewing

The three clamped fields are **destructured out of the spread**, not overwritten after it:

```ts
const { duration, windowWidth, windowHeight, ...rest } = values;
```

That's load-bearing. With `...values` still carrying them, the clamp would depend on key **order** — move the spread, or let a merge reorder it, and the raw values win again silently, which is the exact failure this guards. Destructuring makes the raw values unreachable at the emit site, so no reordering can reintroduce it.

## Tests

Expectations are **literal** — the real rejected production values and the ClickHouse type bounds — never derived from the implementation. The boundary is pinned from both sides (the exact ceiling must pass through unchanged; one past it must clamp), because either half alone is satisfied by an implementation off by one.

**Every guard was watched to fail, in isolation:**

| mutation | result |
|---|---|
| revert only the emit wiring | **3 failed** (wiring tests), 16 passed — unit tests correctly unaffected |
| remove the ceiling clamp | **10 failed** |
| remove the negative guard | **1 failed**, precisely the negatives test |

Suite: **87 passed across 8 files** (base: 68 across 7).

## Two defects in my own test, caught by running it

Both are recorded in comments, because they're the reusable shapes rather than one-off slips:

1. The source slice was bounded at the first `\n  }` — which matches the `}) {` closing `pageView`'s own **parameter type**. Every structural assertion was running against an **empty string**. It failed loudly, but only because the positive assertions were there; a `.not.toContain` check alone would have gone green on an empty slice.
2. `.not.toContain('...values')` matched the **comment** that explains the hazard, failing against correct code. A grep can't tell code from prose. The block is comment-stripped now, behind a positive control proving the stripper hasn't eaten the code.

## Not claimed

This stops the *loss*. It does not fix the upstream producer: the durations increase monotonically within a session, which reads as an accumulator that never resets, and the absurd window dimensions suggest instrumented or headless clients. Both are client-side and unconfirmed — worth a look, but the row-destroying behaviour is closed either way by clamping at the boundary.

### DIFF
--- a/src/server/clickhouse/__tests__/tracker-pageview-clamp.test.ts
+++ b/src/server/clickhouse/__tests__/tracker-pageview-clamp.test.ts
@@ -92,19 +92,31 @@ describe('pageView emits through the clamp', () => {
   // An earlier version of this comment said "a behavioural test would need the whole
   // tracker/session/actor apparatus", and used that to justify pinning the wiring by
   // source text alone. It was FALSE, and it is recorded here rather than quietly deleted
-  // because it caused a real defect: that apparatus already existed 40 lines away in
+  // because it caused a real defect: that apparatus already existed, in this directory, in
   // `tracker.blockRender.test.ts`, and because nothing asserted on the VALUES that get
   // POSTed, the suite could not see the two constants in `tracker.ts` that decide what
-  // reaches ClickHouse. Measured: setting `INT16_MAX` to 65_535 left this suite at
-  // 87 passed, fully green, while every window dimension in (32767, 65535] would go
-  // straight at an `Int16` column and destroy its row.
+  // reaches ClickHouse. Measured at the time: setting `INT16_MAX` to 65_535 left the
+  // whole `src/server/clickhouse/` suite at 87 passed, fully green, while every window
+  // dimension in (32767, 65535] would go straight at an `Int16` column and destroy its
+  // row. (87 was the directory total BEFORE `tracker.pageView.test.ts` existed; today
+  // the same mutation is caught. Re-deriving that number now gives a different total —
+  // it is recorded as the historical measurement, not as a reproducible one.)
   //
   // 🔴 SO: IF YOU ARE GUARDING A NEW NARROW COLUMN, ADD A WIRE ASSERTION THERE, NOT A
-  // SOURCE-TEXT ASSERTION HERE. What survives in this file is only what a wire test
-  // genuinely cannot see: that a FUTURE edit could reintroduce `...values` after the
-  // clamps. A spread placed BEFORE the clamps is behaviourally identical to the
-  // destructure, so the wire test cannot distinguish them — that, and only that, is what
-  // these structural assertions are for.
+  // SOURCE-TEXT ASSERTION HERE.
+  //
+  // Of the structural assertions below, exactly which are irreplaceable — stated per
+  // test, because an earlier version of this comment said "only what a wire test
+  // genuinely cannot see" and that was FALSE for one of them:
+  //   * `does NOT spread the raw values object` — UNIQUE. A spread placed BEFORE the
+  //     clamps is behaviourally identical to the destructure, so the wire test passes on
+  //     it (measured: 9/9); only source text catches that a future edit could put it
+  //     back. A spread AFTER the clamps IS caught by the wire test (measured: 5 failed).
+  //   * `the ledger of clamped fields` — UNIQUE. A newly-added unclamped column is
+  //     absent from the wire test's fixture, so only this sees the set change.
+  //   * `clamps all three narrow columns` — REDUNDANT with the wire test, which fails
+  //     5/9 if any clamp is removed. Kept as a cheap locality check, not because it is
+  //     irreplaceable. Do not cite it as a reason to add more source-text assertions.
   //
   // They assert a RELATIONSHIP (raw values unreachable at the emit site), not the
   // presence of a word — a guard that merely checked for the string "clampToColumn"
--- a/src/server/clickhouse/__tests__/tracker-pageview-clamp.test.ts
+++ b/src/server/clickhouse/__tests__/tracker-pageview-clamp.test.ts
@@ -1,0 +1,150 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'fs';
+import path from 'path';
+
+import { clampToColumn } from '~/server/clickhouse/tracker';
+
+/**
+ * `pageViews` was the last table still losing rows to the ClickHouse tracker's silent
+ * client-side rejection, and unlike the six enum gaps before it the cause was INTEGER
+ * WIDTH OVERFLOW: a value too large for its column destroys the whole row while the
+ * fire-and-forget POST still reports success.
+ *
+ * 🔴 The expectations here are LITERAL, taken from the production rows that were actually
+ * rejected (read out of the dead-letter queue on 2026-09-08) and from the ClickHouse type
+ * bounds — never derived from the implementation they test.
+ */
+
+const UINT32_MAX = 4_294_967_295;
+const INT16_MAX = 32_767;
+
+describe('clampToColumn', () => {
+  it('passes through a value the column can hold', () => {
+    expect(clampToColumn(13_178, UINT32_MAX)).toBe(13_178);
+    expect(clampToColumn(1_920, INT16_MAX)).toBe(1_920);
+  });
+
+  it('accepts the exact ceiling unchanged', () => {
+    // The off-by-one that would matter in the SAFE direction: clamping the boundary
+    // itself would corrupt legitimate maximum values on every row.
+    expect(clampToColumn(UINT32_MAX, UINT32_MAX)).toBe(UINT32_MAX);
+    expect(clampToColumn(INT16_MAX, INT16_MAX)).toBe(INT16_MAX);
+  });
+
+  it('clamps one past the ceiling', () => {
+    // The other half of the pair. Together these two pin the boundary exactly; either
+    // alone is satisfied by an implementation off by one.
+    expect(clampToColumn(UINT32_MAX + 1, UINT32_MAX)).toBe(UINT32_MAX);
+    expect(clampToColumn(INT16_MAX + 1, INT16_MAX)).toBe(INT16_MAX);
+  });
+
+  describe('the real rejected production values', () => {
+    // Each of these destroyed a real pageViews row. Pinning the actual numbers means a
+    // regression is reported in the terms of the incident, not as an abstract bound.
+    const rejectedDurations = [
+      4_323_369_542, // smallest observed overflow — only 0.66% over the ceiling
+      4_742_447_225,
+      5_271_443_941,
+      12_434_883_238, // largest observed — ~2.9x the ceiling
+    ];
+    it.each(rejectedDurations)('duration %d becomes the UInt32 ceiling', (value: number) => {
+      expect(value).toBeGreaterThan(UINT32_MAX); // the fixture is genuinely a violation
+      expect(clampToColumn(value, UINT32_MAX)).toBe(UINT32_MAX);
+    });
+
+    const rejectedDimensions = [51_600, 100_000, 102_000, 183_800, 211_900];
+    it.each(rejectedDimensions)(
+      'window dimension %d becomes the Int16 ceiling',
+      (value: number) => {
+        expect(value).toBeGreaterThan(INT16_MAX);
+        expect(clampToColumn(value, INT16_MAX)).toBe(INT16_MAX);
+      }
+    );
+  });
+
+  it('collapses non-finite input to 0 rather than emitting NaN', () => {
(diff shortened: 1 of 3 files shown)


ZacxDev:
Adversarial audit round 1 complete. Findings and the fixes that landed in `e6574d16`.

**Round 1 findings:** 0 🔴 · 2 🟡 · 3 🟢. Verdict was advisory "safe to merge"; the two 🟡s were fixed rather than shipped, because one of them meant the guard protecting this change could not see the constants that decide it.

Both 🟡s were reproduced independently before being fixed — the survived mutation was re-run here (87 passed with a wrong `INT16_MAX`, matching the report exactly).

Not fixed in this round, recorded so it reads as open rather than absent:
- **`ads` (Bool) can still destroy a row.** `/api/internal/ping` type-asserts its JSON body instead of parsing it, so `ads: "true"` is forwarded and rejected exactly as an oversized integer was. Out of this PR's scope (different file, different mechanism); the code comment now says explicitly that this change closes the numeric-width half only. Closing condition: a `zod` parse of that endpoint's body, or a named reader dismissing it in writing.
- **JSON key order on the wire** is unverified — the tracker service is not in this repo. Almost certainly inert for any name-based decoder; flagged rather than claimed covered.

```audit-claims round=1 audited=52500b115c..e6574d16
1. Added tracker.pageView.test.ts, a behavioural test exercising the real Tracker and asserting the values POSTed on the wire; claims both constant mutations (INT16_MAX 32767->65535, UINT32_MAX widened) now fail the suite, where previously both survived fully green.
2. Removed the claim that a behavioural test "would need the whole tracker/session/actor apparatus"; claims that apparatus already existed in tracker.blockRender.test.ts in the same directory.
3. Retracted in place the comment claim that saturation is honest because the stored distribution "was ALREADY clipped at this bound, just by silent loss instead"; claims dropping removes a row from the distribution while clamping inserts it at the maximum, and that these are different operations with different means.
4. Added a statement that duration == 4294967295 is a saturation sentinel that aggregates should exclude, and that the "max pinned just under a type bound" diagnostic is destroyed for this column.
5. Removed the emitBlock.length < rawBlock.length assertion from the positive control; claims the suite no longer goes red when only the // comments inside pageView are deleted with the code left byte-identical.
6. Extended the comment stripper to remove /* */ blocks as well as // lines; claims a JSDoc rewrite of the in-method comment no longer fails the .not.toContain('...values') assertion.
7. Corrected the observed window extremes to widths up to 183,800 and heights up to 211,900 from different rows; claims no single row was 183,800 x 102,200.
8. Added an explicit statement that this change closes the numeric-width half only, and that a wrong-TYPE value (e.g. ads: "true" into a Bool column) still destroys the row via /api/internal/ping's unparsed body.
```

ZacxDev:
Adversarial audit round 2 (delta, `52500b115c..e6574d16`) complete. Fixes in `71ff5c955b`.

**Round 2 findings:** 0 🔴 · 1 🟡 · 1 🟢.

🔴 **Correction to the round-1 claims block above: claim 2 was FALSE.** It asserted the "a behavioural test would need the whole tracker/session/actor apparatus" sentence had been removed. It had not — the round-1 commit touched only lines 111–136 of that file, and the sentence was still in the tree at line 89. Correcting publicly rather than editing the earlier comment, since it has already been read.

That claim being false mattered more than a stale comment usually would: the sentence is the *justification* for the defect round 1 fixed. It tells the next reader a wire-level test is out of reach — the reasoning that left the suite blind to the two constants in the first place.

**Round-2 disposition of round 1's claims** (from the auditor, verified independently where noted): 1 ✅ · 2 ❌ **not fixed** · 3 ✅ · 4 ✅ · 5 ✅ · 6 ✅ · 7 ✅ as written (the production fact behind it is not checkable from a sandbox) · 8 ✅.

The auditor independently killed the guard with **15 mutations**, including four this ladder had not tried: narrowing rather than widening each constant, swapping the two constants at the call sites, and respreading `...values` *after* the clamps. Two survivors were confirmed genuinely equivalent (`Math.trunc`→`Math.floor`, identical on every non-negative input; and reordering the finite/negative guards).

**Ledger — round 2:** payload lines changed this round: **31** (since round 1: 31) · scaffolding: 161.

⚠️ One number, one name: "payload lines" above is the round-1 classification (`tracker.ts` = payload) applied consistently, and it is the number any stop would be taken on. A separate observation, not that number: **all 31 of those lines are comments** — zero executable payload changed in round 2, and the same is true of round 3's fixes. Recorded because it is the signal the attribution gate exists to surface, even though the gate does not fire numerically.

Still open, unchanged from round 1 and recorded so it reads as open rather than absent:
- **`ads` (Bool) can still destroy a row** — `/api/internal/ping` gives `JSON.parse`'s `any` a type annotation and validates nothing at runtime, so `ads: "true"` is forwarded into a `Bool` column and rejected exactly as an oversized integer was. Out of this PR's scope. Closing condition: a `zod` parse of that endpoint's body, or a named reader dismissing it in writing.
- **JSON key order on the wire** — unverified; the tracker service is not in this repo.
- **The production figures** this PR quotes (the DLQ sample, ~55.7M rows/7d, the live max) are not checkable from a sandbox. They were measured against production in-session; the auditor confirmed only that they are internally consistent.

```audit-claims round=2 audited=e6574d16..71ff5c955b
1. Rewrote the false "a behavioural test would need the whole tracker/session/actor apparatus" sentence at tracker-pageview-clamp.test.ts:89; claims it now records that the premise was false, points at tracker.pageView.test.ts, and states what the structural assertions are still for (a spread placed BEFORE the clamps is behaviourally identical to the destructure, so only source text can catch a future reorder).
2. Swept the directory for other claims that a behavioural test is infeasible; claims one hit, the one reported.
3. Widened the saturation-sentinel warning from `duration` alone to all three clamped columns; claims windowWidth/windowHeight == 32767 are sentinels by the same argument, and states the leverage difference (~5.9 orders of magnitude vs ~17x).
4. Corrected "4-5 orders of magnitude" to ~5.9; claims the earlier figure understated and therefore weakened its own argument.
5. Corrected "no longer distinguishes a clipped tail from a healthy one"; claims a max EQUAL to the bound still proves clipping occurred and what dies is the just-under-the-bound signature.
6. Corrected "type-asserts its JSON body" to a type ANNOTATION over JSON.parse's any with no runtime validation; claims this was verified by reading ping.ts and that behaviour is identical either way.
7. Claims the suite is 96/9 green after these edits, that INT16_MAX 32767->65535 still kills 2 tests, and that the comment-only-deletion control is still green.
```

ZacxDev:
Adversarial audit round 3 (delta, `e6574d16..71ff5c955b`) complete. Fixes in `8a299513dc`.

**Round 3 findings:** 0 🔴 · 0 🟡 · 3 🟢 — and all three were prose written by round 2's fix. Each was a *measurable* claim the auditor refuted by running it, not a matter of taste, so all three were fixed and every replacement claim was itself verified by running it.

The auditor also independently re-killed the guard (INT16_MAX widened, all three clamps removed, `...values` after the clamps, `...values` before the clamps, plus two positive controls) and confirmed 96/9 green, 14/14 CI checks passing, `mergeStateStatus: CLEAN`.

## 🔴 STOPPING THE LADDER HERE — on the stated criterion, not because a round came back clean

Round 3 was **not** clean, so this is not the findings-keyed stop. I am stopping on the prose criterion, and per the rule that criterion is only usable if I can **name why the rounds will not stop on their own**. Naming it:

- **The executable payload has been frozen since round 1** and is mutation-verified from two independent directions (round 2's auditor ran 15 mutations; round 3's ran 4 more plus 2 controls; the only survivors were confirmed genuinely equivalent).
- **Rounds 2 and 3 each changed ZERO executable payload** — verified mechanically both times, not asserted. Two consecutive rounds.
- **Every finding in rounds 2 and 3 was a defect in prose that the PREVIOUS round's fix wrote.** Round 2 found that round 1's claims block was false. Round 3 found three sentences round 2 wrote while fixing round 1. Round 4 would audit the sentences round 3 just wrote.
- That is self-sustaining rather than converging: each fix round's output *is* new prose, and new prose is the next round's audit surface. It has no natural terminus.

The escape-hatch conditions are met: no 🔴, no blast radius beyond "a comment contains a false sentence", and the recurring shape swept at every site rather than only where reported — this round I swept both files for distance claims, non-reproducing numbers and unqualified assertions, and fixed three more that nobody had reported (the "40 lines away" distance claim, the "87 passed" measurement whose referent no longer exists, and "~5e3 ms typical" which is an assumption rather than a measurement).

## Recorded as OPEN, not absent

- **`ads` (Bool) can still destroy a row.** `/api/internal/ping` gives `JSON.parse`'s `any` a type annotation and validates nothing at runtime, so `ads: "true"` is forwarded into a `Bool` column and silently rejected. Out of this PR's scope. **Closing condition:** a `zod` parse of that endpoint's body, or a named reader dismissing it in writing.
- **Not fixed, deliberately:** the `~5.9 orders of magnitude` / `~17×` pair mixes units (cosmetic — both figures are given plainly enough to convert); and the comment does not warn that adding a fourth clamped column will turn the ledger test red and force an edit in that file (a loud, self-explaining failure).
- **Unverifiable from a sandbox, by construction:** every production figure this PR quotes — the DLQ sample (21 of 26 / 5 of 26), `~55.7M rows/7d`, the live max `4,264,810,774`, the `~5e3 ms` baseline. These were measured against production in-session; all three auditors confirmed only that they are internally consistent, which is a consistency check and not a verification.
- **JSON key order on the wire** — unverified; the tracker service is not in this repo. Only relevant to a caller ordering the clamped fields early, which `ping.ts` does not.

## Ledger — final

```
round 3 · payload lines changed THIS round: 42 (since round 1: 73) · elapsed: 3 rounds
```
One number, one name: "payload lines" is the round-1 classification (`tracker.ts` = payload) applied unchanged throughout, and it is the number any stop would be taken on. Separate observation under a different name: **zero executable payload for two consecutive rounds** — that is the signal the attribution gate exists to surface, and it is what the stop above rests on, since the gate cannot fire numerically when the payload edits are comments inside a payload file.

```audit-claims round=3 audited=71ff5c955b..8a299513dc
1. Replaced the sweeping claim "only what a wire test genuinely cannot see" with a per-test disposition; claims two structural assertions are unique and one (clamps all three narrow columns) is redundant with the wire test and labelled as such.
2. Claims the spread-BEFORE-clamps variant passes the wire test 9/9 while failing the directory suite 1, and that removing all three clamps fails the wire test 5 of 9 — both measured by running them.
3. Replaced "what dies is the JUST-UNDER-the-bound signature" with SHADOWS rather than destroys; claims the clamp is the identity below the bound so max(duration) WHERE duration < 4294967295 recovers the tail exactly, using the exclusion the same block already mandates.
4. Replaced the "(checked: ... appear outside __tests__ only on the write path)" parenthetical; claims the reproducible method is to enumerate readers of the pageViews TABLE and then check their columns, and warns that the column-name grep does not reproduce because `duration` alone appears in 271 non-test files.
5. Replaced the "40 lines away" distance claim about another file with "in this directory".
6. Labelled the "87 passed" figure as a historical measurement whose referent (the directory total before tracker.pageView.test.ts existed) no longer reproduces.
7. Labelled "~5e3 ms typical page duration" as an assumption rather than a measurement, and claims the conclusion survives any plausible baseline (5.1 OOM even at 30 s).
8. Claims the suite is 96/9 green after these edits and the INT16_MAX mutation still dies.
```

PR #4717, opened 2026-09-09
perf(apps): cache the /apps store catalog query behind a bust tag

ZacxDev:
`appListings.listAvailable` — the query behind the whole `/apps` store grid — had **no server-side cache**. Every page load was a cold DB read of a keyset query that joins `app_listing_metrics` and computes a Bayesian sort key. This adds one.

Refs #529.

> **Numbers in this body were re-derived at the current head.** Every count, ratio and cross-reference below was measured against this branch on 2026-09-08, not carried forward from an earlier round. Where a claim from an earlier round turned out to be unpinned or wrong, it says so rather than being quietly replaced.

---

## The design, and why not the alternatives

The cache is **`queryCache` + a bust tag**, wrapping the raw keyset query inside `listAvailableListings`. Hydration and the `nextCursor` derivation are untouched — `nextCursor` comes off the (now cached) `idRows`, exactly as `getPostsInfinite` does it in `src/server/services/post.service.ts`.

**What is cached is `{ id, sort_key }` per row, and nothing else.** Every projection field on a card — tagline, description, icon/cover/screenshot URLs, the owner chip, the beta badge — comes from a **live** hydration below the cache on every request. That distinction decides which mutations need a bust, and getting it wrong in either direction is the main way this change could have been wrong. It is stated in the tag's own module doc (`app-listing-cache.constants.ts`), which is the first thing a maintainer reads.

**Two of the query's axes are viewer-varying, and they are LITERAL key segments, not hash input.**

`cache-helpers`'s `queryCache` builds its key as `[key, version, hashifyObject(query)].join(':')`. `hashifyObject` → `hashify` is a **32-bit linear rolling hash**, and `decodeListingCursor` admits `cursorSortKey` / `cursorId` as arbitrary free strings straight into the hashed statement as bound params (the read schema bounds `cursor` only by `z.string().max(128)`). So a collision against a chosen target is *constructible*, not something to brute-force. An earlier revision of this change relied on "every axis is interpolated into the statement, therefore every axis is in the key", which silently assumes the hash is injective. It is not.

The fix is to lift the two axes that are **security boundaries** out of the hashed payload and into the literal key — `listAvailableAppListings:<scope>:<red|sfw>`:

- `scope` — the public/onsite boundary. A collision would serve on-site apps to anonymous callers of `GET /api/v1/apps`, i.e. civitai#3983 re-opened through the cache rather than through a `??`.
- `redCapable` — the maturity gate. A collision would serve `r`/`x` cards onto a SFW domain.

With both in the prefix, a hash collision can only ever mix two pages **within** one viewer class.

🔴 **The residual is stated as a residual.** The remaining axes (`kind`, `category`, `sort`, `cursor`, `limit`) stay in the hash, and a constructed collision across them is **cross-user cache poisoning of the shared `/apps` grid** — the attacker's crafted-cursor request MISSES, so it is the request that WRITES the colliding key, and every later reader deriving that key HITS it. One request can pin the store's first page to an arbitrary filtered or empty result for up to 180s for everyone in that class. (A previous round's comment described this as "the attacker's own page served back to themselves", which was wrong; it is fixed in the code comment too.) It is **not** a disclosure boundary — every row in a poisoned page came from a statement carrying the same `scope` and `redCapable` predicates. It is accepted deliberately: containing the rest means putting `cursor`, a free 128-byte string, into the literal key, which makes the keyspace and the `cache_name` metric label request-controlled and unbounded.

**Not `fetchThroughCache`** — it has no `tag` option, so `bustCacheTag` cannot drive it and a moderator approve/delist would be invisible until the TTL expired.

**Not `clearCacheByPattern`** — banned for bust-on-mutation by its own header in `cache-helpers.ts`: a prior use ran a cluster SCAN over a ~60M-key shard → redis timeouts → 504 waves, and was reverted.

**Not a generation counter** — a counter folded into the key leaves the superseded entries in redis to expire on their own, so a bust *multiplies* the keyspace instead of reclaiming it. The tag set holds the exact keys to delete.

**Not widening `hashify`** — it keys caches, DOM ids and de-dup across the codebase; changing its output is a global blast radius. The containment belongs at the one call site that has a security boundary.

**TTL = `CacheTTL.sm` (180s).** The catalog is mod-gated and low-churn, so the TTL is not the freshness mechanism — the busts are. It is the backstop for the path that has no mutation to hang a bust on (a row that ages into visibility). It is explicitly **not** a redis-outage backstop: `queryCache` has no fail-open, so a redis outage 500s `/apps` rather than degrading to a live read. That is the dependency this change accepts.

---

## The leaf constants module

New: `src/server/services/blocks/app-listing-cache.constants.ts`, importing **nothing**. Exports `APP_LISTING_CATALOG_TAG` (new) and `APP_LISTING_RECOMMEND_MEAN_TAG` (the *existing* tag string, previously duplicated as a private literal in `app-listing.service.ts` and `app-listing-review.service.ts` — a producer and its only buster each spelling it independently). Both sites now read the shared const.

It is not in `app-listing.service.ts` on purpose. `app-listings.router.ts` and `blocks.router.ts` deliberately `await import()` their services to keep the Prisma client out of the router import graph — `blocks.router.ts` has **8** such sites for `user-app-surface.service` alone (re-counted at head). Exporting a constant from a heavy service makes those lazy imports **graph-inert**: nothing errors, and the cost only reappears later when someone adds an import to the service. Precedent and the measured version of that failure: `src/server/services/blocks/scope-activity-predicate.ts`.

---

## Bust sites, and the rule

🔴 **The rule is "bust when a cached axis or catalog membership moves" — NOT "every listing-state mutation busts".** Earlier rounds of this PR invoked the latter as a "UNIFORM RULE" at five call sites. It is false, and stating it that way makes a reader's model of the cache wrong in the expensive direction: it implies a writer without a bust is a bug, when a whole enumerated list of them are deliberate and correct. All five sites now say what they actually are — cheap defence-in-depth on a path that is inert today — and name which.

The cached statement reads exactly `al.id`, `al.status`, `al.revision_of_id`, `al.kind`, `al.category`, `al.content_rating`, `al.app_block_id` (the join key onto `app_blocks` for the deploy gate), `ab.current_version_deployed_at`, and the `sort_key` inputs (`al.name`, `al.created_at`, the `app_listing_metrics` rollup). A write that moves one of those needs a bust. A write that moves anything else does not.

`al.app_block_id` was missing from this list until round 4. It is latent rather than live: the only write of that column onto an **existing** row is `approveRequest`'s link step, guarded `where: { status: 'draft' }` (`publish-request.service.ts`), and it busts anyway — every other write of it is a `create` that mints at `status:'draft'`. It is listed so the next writer of the column is checked against the rule rather than against a list it is missing from.

🔴 **Two of those axes do not live on `app_listings`, and the rule is violated on them BY DESIGN.** The `app_listing_metrics` rollup feeds `sort='top-rated'` (thumbs) and `sort='popular'` (`install_count`), and it has two live writers that do **not** bust: `app-listing-review.service.ts` on every review vote (it busts only the recommend-MEAN tag) and `src/server/metrics/appListing.metrics.sql.ts` on the metric job (it busts nothing). That is the right trade — busting per vote or per metric pass would defeat the cache, while the cost is bounded at one `CacheTTL.sm` window of **sort** lag, with no row appearing, disappearing or changing maturity — but nothing said so, so a reader applying the rule to a metrics writer reached the wrong conclusion. It is now stated in all four places the axis list appears, and the ledger suite's blind-spot section names it as a blind spot the writer scan structurally cannot cover (the writers are on a different table, so no `appListing.` delegate call exists to enumerate).

One exported helper, `bustAppListingCatalogCache()`, fired post-commit as `await bustAppListingCatalogCache().catch(() => undefined)` — the posture of `bustRecommendMeanCache` in `app-listing-review.service.ts`. A cache-bus outage must never fail a mutation that already committed.

**19 bust sites**, pinned by name in the ledger suite. Load-bearing ones:

| Site | Cached axis it moves |
|---|---|
| `approveRequest` (`publish-request.service.ts`) | mints/flips to `approved` — membership |
| `approveExternalRequest`, `applyApprovedRevision` (`offsite-listing.service.ts`) | membership; and `applyApprovedRevision` writes `contentRating` / `category` / `name` onto the **live parent** |
| `delistListing`, `relistListing`, `purgeListing`, `resetListingToPending`, `resetOnsiteListingToPending`, `unpublishOwnListing`, `republishOwnListing` (`offsite-moderation.service.ts`) | membership |
| `updateListing`'s `approved` trivial-only branch (`offsite-listing.service.ts`) | `category` on a live row |
| `setListingIcon` / `setListingCover` / `addListingScreenshot` (`app-listing-assets.service.ts`) | `content_rating`, via the in-tx `reDeriveContentRatingForModLiveEdit` |
| `backfillAppListings` (`app-listing-backfill.service.ts`) | mints at `status:'approved'`. Guarded `if (!dryRun && result.created > 0)` |
| 🔴 `watchApplyJobAndRecord` (`build-callback.ts`) | **non-tRPC.** Writes `currentVersionDeployedAt` — the onsite deploy gate in the cached query. Without it a freshly-deployed onsite app is invisible for the whole TTL |

The remaining sites (`claimListing`, `rejectExternalRequest`, `submitListingRevision`, `updateListing`'s `removed` / `draft`-`pending` / material-shadow branches) are **inert today** and each says so, with what would make it load-bearing.

⚠️ Two justifications from earlier rounds were **wrong and are corrected in the code**, not quietly dropped:
- the icon/cover busts were explained as "the card's `iconUrl` moved". `iconUrl` is hydrated live and can never be stale; the **content-rating re-derive** is the whole reason those busts exist.
- `rejectExternalRequest`'s bust was explained as "both outcomes are catalog membership changes". Neither is — it reaches the table only through `closeTerminalListing`, which deletes a `draft` or flips `pending`→`removed`, both already outside the approved-only query.

🔴 **`BlockRegistry.setMarketplaceMeta` has NO bust, deliberately.** It writes `app_blocks.category`; the catalog statement filters `app_listings.category`. Different table, different column. Its comment now names the exact ledger row to add — `'src/server/services/block-registry.service.ts::BlockRegistry::setMarketplaceMeta'` — if the cached query ever grows a predicate over an `app_blocks` column it writes.

**`approveRequest` uses a lazy `await import`**, not a static one, and the `try` wraps the **import** as well as the call: `publish-request.service.ts` deliberately keeps `~/server/db/client` and `~/env/server` out of its static graph so its pure-helper suites can load it without Prisma, and a module-load failure anywhere in `app-listing.service`'s static graph would otherwise propagate out of `approveRequest` after the DB writes had committed.

---

## 🔴 The freshness guard: enumerate WRITERS, not busters

`src/server/services/blocks/__tests__/app-listing.catalog-bust-ledger.test.ts` (**11 tests**).

An earlier round of this file pinned only the SET OF BUST CALL SITES and its header claimed that went red "when a new listing mutation appears without a bust". **It did not** — a mutation with no bust adds no call site, so the pinned set is unchanged and the ledger stays green. Measured on this branch: appending an `export async function forgetfulDelist()` that flips an approved row to `removed` with no bust left the bust-only ledger at **5 passed / 0 failed**. That is the same zero-call-site condition as the original `applyApprovedRevision` defect the header cited as proof it would have caught it, so the guard was reading as coverage while providing none.

The primary assertion is now **inverted**. The scan enumerates every writer of the `AppListing` table (`<client>.appListing.{create,createMany,update,updateMany,upsert,delete,deleteMany}`, `tx.` forms included) across `src`/`apps`/`packages`/`scripts`, resolves each to its enclosing function, and requires that function to be **either** a bust call site **or** on an `EXEMPT` map with a one-line reason. A new un-busted mutation fails by construction. The bust set is still pinned separately, for the direction the writer scan cannot see: deleting a bust.

**`EXEMPT` (11), each reason re-derived against the code at this commit:**

| Writer | Why it does not bust |
|---|---|
| `backfillListingAssets` | writes only `coverId` / `iconId` — hydrated projection fields |
| `reDeriveContentRatingForModLiveEdit` | does raise `content_rating`, but is an in-tx helper whose callers all bust post-commit (a bust inside the tx would fire on a rollback too). **The caller set is now pinned mechanically** in `CALLER_BUSTS` — see round 4 below |
| `acceptTransfer` | writes only `userId`; the statement never reads ownership |
| `beginListingRevision`, `submitExternalListing`, `submitVersion` | mint at `status:'draft'` |
| `closeTerminalListing`, `closeOnsiteResetListingOnWithdraw`, `routeRepublishToReviewInTx` | move between non-`approved` statuses |
| `updateRevisionDraft` | writes the draft shadow plus `is_beta` on the parent; `is_beta` is not in the cached statement |
| `deleteOnsiteDraftListingForSlug` | deletes only `{kind:'onsite', appBlockId:null, status:'draft'}` |

A hygiene test keeps `EXEMPT` from rotting: every entry must still be a current writer, must not also bust, and must carry a real reason.

### 🔴 The attributor is a backward parse, and it fails loudly

Three pattern-list versions each mis-attributed **silently**, which is the worst output this file can produce — it sends the reader to an unrelated function:

1. a two-space-indented `name(` alternative added to name class methods also matched `  if (` and `  switch (`, so four real mutations were attributed to a function called `if`;
2. removing that and anchoring at column 0 made class methods invisible instead — re-adding the bust inside `BlockRegistry.setMarketplaceMeta` reported it as `block-registry.service.ts::resolveRenderMode`, an unrelated top-level helper, **while a comment at that call site instructs a maintainer to add that function to the ledger**;
3. a leftmost-match regex over a lookback window jumped to the *previous* function whenever a signature contained an object-type parameter (`function f(opts: {`), so `delistListing` was reported as `flipBackingBlockStatus`.

It now walks backward from each `{` with real bracket matching, nests frames by brace depth, and names class methods `Class::method`. When it cannot name a frame it emits `<unattributed>` and a test fails on that — which is how `acceptTransfer` and `updateRevisionDraft` were found (a character-class scan could not step back over `Promise<{ … }>`).

**Mutation results (all three verbatim in the PR discussion):**

| Mutant | Result |
|---|---|
| new un-busted writer `forgetfulDelist` appended to `offsite-moderation.service.ts` | **1 failed** — `🔴 every AppListing writer either busts the catalog cache or is EXEMPT`, naming `…offsite-moderation.service.ts::forgetfulDelist`. Mutant compiled (root typecheck still 13, none in that file) |
| `claimListing`'s bust deleted | **2 failed** — the coverage guard names `…::claimListing`, and the ledger comparison reports `REMOVED: ["…::claimListing"]` |
| bust re-added inside `BlockRegistry.setMarketplaceMeta` | **1 failed** — `ADDED: ["src/server/services/block-registry.service.ts::BlockRegistry::setMarketplaceMeta"]`, i.e. the exact string that call site's comment tells you to add |

---

## The key guard

`src/server/services/blocks/__tests__/app-listing.catalog-cache.test.ts` (**13 tests**). It asserts on the **derived cache key** — the real `hashifyObject` applied to the real `Prisma.Sql` captured on its way into `queryCache`, joined the way `cache-helpers` joins it — with the hash segment **removed**, because that is the half a collision cannot touch. A first test pins that the derivation is deterministic for an unchanged shape, so "these differ" is not vacuously true, and a drift-guard fails if `cache-helpers` stops keying on the whole `Prisma.Sql`.

The viewer-class product is now enumerated in full and its size is **derived from `STORE_VISIBILITY_SCOPES`**, not written down — which also pins the `cache_name` label-cardinality claim `catalogPageCache` makes. Adding a scope fails this test until the new classes are enumerated.

⚠️ An earlier round's comment quoted "**six tuning characters** in the cursor are enough to collide the two scope statements". That figure was never derived or pinned by anything, so it is **gone rather than replaced**. What is true and sufficient: the hash is 32-bit and linear and the attacker controls two free strings in the hashed payload, so a collision is constructible; and the guard puts the boundary outside the hash entirely, which makes the exact cost irrelevant.

---

## Client

`AppListingsMarketplaceBody.tsx` — the `useInfiniteQuery` had no `staleTime`, so react-query's default of **0** refetched on every remount and every window focus, on a page viewers leave and immediately return to. Now `staleTime: 60 * 1000`.

The **property** is that the client window stays strictly shorter than the server TTL, so the client can never be the longer of the two staleness sources; a client window outliving the server TTL would make the explicit busts invisible for the difference. That inequality — and only that — is what the browser spec asserts, against the real `CacheTTL.sm` rather than a literal. The 3× headroom is a judgement, not a pinned ratio (an earlier round's body called it "deliberately a third of the server TTL", which reads as a pinned relationship and is not one).

Mutating `staleTime` to `600 * 1000`:

```
AssertionError: the client staleTime is >= the server-side TTL (CacheTTL.sm = 180s) on
`listAvailableListings`. A client window longer than the server one makes the explicit
catalog busts invisible for the difference.: expected 600000 to be less than 180000
```

It does not interact with `placeholderData: keepPreviousData` (untouched): `staleTime` governs refetch of *this* key, `keepPreviousData` governs what renders while a *different* key loads.

---

## Round 4 — three findings, all in the guard, none in the payload

A fourth adversarial round found **no payload defect** and confirmed all 11 `EXEMPT` reasons hold. Every finding was in the ledger guard itself, which is round 3's whole deliverable.

### F1 🔴 The ledger could be walked silently by an unbalanced brace in a regex literal

`buildFrames` nests frames by brace depth over text from `stripCommentsAndStrings`, which removes comments, template literals and quoted strings — but **not regex literals**. A `{` or `}` inside one survives, every frame after it re-nests, and a write is then attributed to a *different real function*: no `<unattributed>`, no error, and the coverage guard reads as satisfied because the name it landed on is in `LEDGER`/`EXEMPT`.

Demonstrated end-to-end, with the mutation isolated to the brace:

| Arm | Ledger file | Result |
|---|---|---|
| `const _jsonHead = /^\s*\{/;` in `acceptTransfer` + un-busting `forgetfulDelist()` appended, both in `app-ownership-transfer.service.ts` | at `0960cebd8f` | **8 passed / 0 failed** — the new writer folded into `acceptTransfer`, which is `EXEMPT`, so it read as covered |
| same mutant | with F1 | **1 failed** — `🔴 every scanned file brace-balances after stripping`, reporting `src/server/services/blocks/app-ownership-transfer.service.ts (delta +1, min depth 0)` |
| same mutant, regex **balanced** (`/^\s*\{\}/`) | with F1 | **1 failed** — `🔴 every AppListing writer either busts the catalog cache or is EXEMPT`, naming `…app-ownership-transfer.service.ts::forgetfulDelist`. The brace is the variable, not the mutation |

`app-ownership-transfer.service.ts` was chosen because it holds **no** bust sites, so the `LEDGER` equality pin cannot see the collapse — that is the silent case. Reachability is real: **39 of the 5,123** scanned files do not brace-balance after stripping, one of them (`src/pages/apps/[appBlockId]/index.tsx`, delta +2) already inside the scan population via `dbRead.appListing.findFirst`. None contributes a site today, so nothing is mis-attributed now; each becomes a red build the day it gains a writer.

**The fix is a brace-balance counter in `buildFrames` that fails loudly**, not a regex-literal stripper. A balance check cannot be walked by a construct nobody thought of, whereas a stripper only closes the shape we happened to see — and `/` is genuinely ambiguous with division in a module two **other** guards also depend on. `minDepth` is tracked alongside the delta so a `+1/−1` pair that cancels is still caught. The imbalanced file is **recorded, not skipped**: attribution still runs, so exactly one test goes red and it is the one naming the real cause.

### F2 🟡 An exemption resting on an unasserted caller-set claim

`reDeriveContentRatingForModLiveEdit`'s `EXEMPT` reason said it "is an in-tx helper with exactly three callers … and all three bust post-commit". True, and **nothing asserted it** — and the dangerous caller is invisible to the writer scan *by construction*, because it issues no `appListing.` write of its own.

| Arm | Result |
|---|---|
| a 4th caller running the helper inside `dbWrite.$transaction` with no bust (`relabelListingAssetsInTx`, reads via `findUniqueOrThrow`) at `0960cebd8f` | **8 passed / 0 failed**; the function never appears in `WRITERS` |
| same mutant with F2 | **1 failed** — `🔴 every CALLER of an exempt-by-caller helper busts, and the caller set is pinned`, naming `…app-listing-assets.service.ts::relabelListingAssetsInTx` |
| same mutant **plus** a post-commit bust | **2 failed** — the SITES half passes and only the bookkeeping pin fires (`the caller set … changed`, 4 vs 3), alongside the ordinary `LEDGER` "a bust nobody recorded" red. Both assertions are separately reachable |

If that caller had landed, a moderator raising an approved listing `g`→`r` would sit in the cached SFW page for the full `CacheTTL.sm` — verbatim the narrative this file's header opens with.

`CALLER_BUSTS` is a map, so a second helper exempted on the same argument is a one-line addition, and a hygiene assertion requires every pinned helper to still be an `EXEMPT` row.

**Secondary, same row:** the `EXEMPT` preamble said "the bar for a row here is that the write provably cannot move a CACHED axis". This row moves one; it is exempt on a **caller** argument. The preamble now states both bars, says which one each kind of row claims, and requires a bar-2 row to carry a `CALLER_BUSTS` pin. (`routeRepublishToReviewInTx` also cites its caller, but independently clears bar 1 — its `where` selects `status:'removed'`, so it moves no catalog member — so it needs no pin.)

### F3 🟡 The cached-axis list was incomplete in four places

`al.app_block_id` and the off-table metric writers, corrected in the ledger suite, `app-listing.service.ts`, `app-listing-cache.constants.ts` and this body — see "Bust sites, and the rule" above.

### Also

- `app-listing.catalog-cache.test.ts` said "the five shapes"; the enumerated product is **six** (`STORE_VISIBILITY_SCOPES.length * 2` = 3 × 2, and six literal rows).
- The `CALL_RE` comment credited its `\(` with excluding the buster's own declaration. It does not — a declaration's parameter list matches `name\s*\(` exactly like a call; the `\bfunction\s+$` lookback in `scan` is what drops it, as that line's own comment says. All three exclusions are now named separately.
- `attribute()` can return `<module scope>`, which the "every site resolves to a named enclosing function" test does not filter. That is deliberate and now documented: it is a *correct* answer (a genuine top-level write), not a parse failure, and it needs no guard of its own — neither list holds a `<module scope>` row today, so a new one lands in neither `LEDGER` nor `EXEMPT` and the headline coverage guard reports it by name. Nothing *mechanically* stops someone typing such a row in; the point is that doing so is a deliberate act, the same contract every other row is under, rather than a silent pass.

---

## Verification (re-run at the current head)

**Typecheck — base-vs-HEAD, `NODE_OPTIONS=--max-old-space-size=12288`.** Both runs in the **same** working tree, swapping the changed files between their base and head contents: a separate worktree is not a valid instrument here, because `packages/` resolves through the primary clone's `node_modules` symlink and an unbuilt worktree reported **54** root errors instead of 13, purely from build state.

| Config | base | HEAD | |
|---|---|---|---|
| `node scripts/typecheck.mjs` | 13 | 13 | **byte-identical output** |
| `node scripts/typecheck.mjs -p tsconfig.tests.json` | 1205 | 1205 | identical after normalising line numbers |

The 1205 is also the **positive control**: a config that reports a large non-zero count is demonstrably wired to the code. Re-measured for round 4 (base = `0960cebd8f`, swapped in the same tree, restored and sha256-verified afterwards): the main config's 13 error lines are **identical as a sorted set**; the tests config is 1205 vs 1205 and **identical on `(file, line, column, TS####, order-insensitive message tokens)`**. Its residual textual diff is 36 lines in two unrelated files (`resolve-nav-items.test.ts`, `process-vault-items.test.ts`) where tsc printed a union's members and an object type's properties in a different order — same location, same code, same members. The 3 pre-existing errors in `offsite-listing.onsite-revision.service.test.ts` are present identically on base.

**Tests.**

- Blocks tier, `vitest run --project unit src/server/services/blocks` — **168 files / 4225 tests, all green** at this head, against **168 / 4222** with the four changed files swapped back to their previous contents in the **same tree**. The +3 are round 4's new ledger assertions (that file goes 8 → 11 tests). ⚠️ An earlier round of this body quoted `165 files / 3885 tests` for this tier; that did not reproduce and is corrected here — 165 is the count of `__tests__/*.test.ts` *files on disk*, while the `unit` project's `src/**/*.test.ts` include selects 168 under this directory. A cold run of the same command in a `cp -a` copy reported `28 failed | 140 passed` with **3845 tests passed and 0 failed** — the dep-optimizer cascade (`Cannot find module …/node_modules/.vite/…/deps_ssr/prom-client.js`), i.e. suites that fail to *import* rather than to assert; 3885 is close enough to 3845 that the old figure may have been taken from a partially-cascaded run. Warm runs in the worktree were 57.8s and 59.7s.
- `AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx` — **4 passed** warm. The browser tier cold-cascades: the first run failed to import (stale vite dep-optimizer cache) and the second reported 4 failed with `Cannot read properties of null (reading 'useSyncExternalStore')`, a duplicated-React symptom of the same cache. An untouched control spec passing in between is what identified it as environmental. Wall time discriminates: 18.05s on the poisoned run vs 7.08s warm.
- `npx prettier --check` on all **22** changed files → clean. `npx eslint` → 2 errors, both pre-existing `no-explicit-any` on `build-callback.ts`'s `safe()` helper, unchanged by this PR.

**CI.** `tekton / typecheck` shows ERROR at the previous head. It is **not** a type error: the TaskRun's own condition reads `TaskRun cancelled as the PipelineRun it belongs to has timed out`, and the PipelineRun reads `failed due to tasks failed to finish within "1h0m0s"`. Two pipelineruns for this PR overlapped (`pr-check-4717-hxh8t` at 02:25:19Z on `abf7e5b02`, `pr-check-4717-wdglh` at 02:27:20Z on `37aef4ed3`); the earlier one's typecheck **passed**, the later one's was starved and cancelled at the hour. The only commit between those two revisions is a prettier-formatting commit. The same check passes on all 7 other open PRs sampled (`gh pr list --state open --limit 8`). This body's CI claims must be re-read after this round's push, which triggers a fresh run.

**Test-scaffold changes this forced** (each is the documented one-key-`vi.mock` trap):

- `app-listing.service.test.ts`, `app-listing.public-scope.test.ts`, `app-listing.deploy-gate.test.ts`, `app-listing.moderation.service.test.ts`, `app-collaborator.public-byline-offsite.test.ts` — their cache-helpers factories exported only `queryCache`; the service now imports `bustCacheTag` too, so each would have died with `No "bustCacheTag" export is defined on the … mock`. Their `~/server/common/constants` stubs also gained `CacheTTL.sm`.
- `publish-request.listingSync.test.ts`, `publish-request.notify.service.test.ts` — these did not mock cache-helpers at all; `approveRequest`'s lazy import of the listing service pulled the real module, whose body builds a logger off the real env, throwing `TypeError: Cannot read properties of undefined (reading 'includes')` *inside* the approve. Stubbed.
- `offsite-listing.onsite-revision.service.test.ts` uses a **partial** mock (spread `importOriginal()`, override `bustCacheTag`) so it cannot go stale when that graph reaches for a third export.
- The new suites use the **canonical `dbMock`** rather than a per-file `~/server/db/client` mock, which `no-direct-shared-module-mock` forbids. (That guard is textual — it also fires on the forbidden call written as a code sample in a comment.)

### DIFF
--- a/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
@@ -334,4 +334,45 @@ describe('🔴 a sort change re-orders the store without emptying it', () => {
     // component's OTHER options too.
     expect(typeof captured?.getNextPageParam).toBe('function');
   });
+
+  /**
+   * 🔴 THE CLIENT HALF OF THE `/apps` CACHE (#529).
+   *
+   * The server now caches this query's keyset page for 180s (`CacheTTL.sm`) behind the
+   * `app-listing:catalog` bust tag. The client half is a `staleTime`: without one,
+   * react-query's default is ZERO, so every remount and every window refocus refetches
+   * a catalog that has not moved — and `/apps` is a page viewers leave and come
+   * straight back to.
+   *
+   * Asserted HERE, through the same capture as `placeholderData` above, for the same
+   * reason: every sibling spec mocks the query with a literal-returning function, so
+   * against those this option is inert and a guard would read as coverage while
+   * providing none. This reads the object the COMPONENT passed.
+   *
+   * The bound is asserted, not just presence: the value must be a positive number and
+   * must stay STRICTLY BELOW the server's 180 000 ms TTL, so the client can never be
+   * the longer of the two staleness sources — a client window outliving the server TTL
+   * would make `bustAppListingCatalogCache()` invisible for the difference.
+   */
+  test('the component supplies a staleTime, bounded below the server TTL', async () => {
+    mocks.sort = 'top-rated';
+    mocks.pending = [];
+    mocks.lastOptions = null as Record<string, unknown> | null;
+    renderWithProviders(<AppListingsMarketplaceBody />);
+    await expect.poll(() => mocks.lastOptions !== null).toBe(true);
+    const captured: Record<string, unknown> | null = mocks.lastOptions;
+    const staleTime = captured?.staleTime;
+    expect(
+      typeof staleTime,
+      'the store query passes no `staleTime`, so react-query refetches the whole ' +
+        'catalog on every remount and every window refocus (its default is 0). See #529.'
+    ).toBe('number');
+    expect(staleTime as number).toBeGreaterThan(0);
+    expect(
+      staleTime as number,
+      'the client staleTime is >= the server-side TTL (CacheTTL.sm, 180s) on ' +
+        '`listAvailableListings`. A client window longer than the server one makes the ' +
+        'explicit catalog busts invisible for the difference.'
+    ).toBeLessThan(180 * 1000);
+  });
 });
--- a/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx
@@ -42,6 +42,13 @@ import { describe, expect, test, vi } from 'vitest';
 import { renderWithProviders } from '../../../test/component-setup';
 import type * as TrpcMod from '~/utils/trpc';
 import type { ListingCard } from '~/server/schema/blocks/app-listing-read.schema';
+// 🔴 THE REAL CONSTANT, not a literal. The bound below is a claim about the SERVER's
+// TTL on `listAvailableListings`; hardcoding `180 * 1000` means lowering that TTL
+// leaves this guard passing while the property it names (client window strictly
+// shorter than server window) is false. `CacheTTL` is a pure const object and
+// `~/server/common/constants` is already imported by components + sibling browser
+// specs, so it loads here.
+import { CacheTTL } from '~/server/common/constants';
 
 function makeCard(id: string, name: string): ListingCard {
   return {
@@ -350,9 +357,10 @@ describe('🔴 a sort change re-orders the store without emptying it', () => {
    * providing none. This reads the object the COMPONENT passed.
    *
    * The bound is asserted, not just presence: the value must be a positive number and
-   * must stay STRICTLY BELOW the server's 180 000 ms TTL, so the client can never be
-   * the longer of the two staleness sources — a client window outliving the server TTL
-   * would make `bustAppListingCatalogCache()` invisible for the difference.
+   * must stay STRICTLY BELOW the server's TTL — DERIVED from `CacheTTL.sm`, the same
+   * constant the service passes — so the client can never be the longer of the two
+   * staleness sources. A client window outliving the server TTL would make
+   * `bustAppListingCatalogCache()` invisible for the difference.
    */
   test('the component supplies a staleTime, bounded below the server TTL', async () => {
     mocks.sort = 'top-rated';
@@ -368,11 +376,17 @@ describe('🔴 a sort change re-orders the store without emptying it', () => {
         'catalog on every remount and every window refocus (its default is 0). See #529.'
     ).toBe('number');
     expect(staleTime as number).toBeGreaterThan(0);
+    // `CacheTTL` is in SECONDS; react-query's `staleTime` is in MILLISECONDS.
+    const serverTtlMs = CacheTTL.sm * 1000;
+    // Positive control on the derivation itself: a `CacheTTL` that failed to load (or an
+    // `sm` renamed away) would give `NaN`, and `toBeLessThan(NaN)` fails loudly rather
+    // than passing — but pin the unit anyway so a seconds/ms mix-up cannot pass either.
+    expect(serverTtlMs).toBeGreaterThan(1000);
     expect(
       staleTime as number,
-      'the client staleTime is >= the server-side TTL (CacheTTL.sm, 180s) on ' +
+      `the client staleTime is >= the server-side TTL (CacheTTL.sm = ${CacheTTL.sm}s) on ` +
         '`listAvailableListings`. A client window longer than the server one makes the ' +
         'explicit catalog busts invisible for the difference.'
-    ).toBeLessThan(180 * 1000);
+    ).toBeLessThan(serverTtlMs);
   });
 });
--- a/src/components/Apps/AppListingsMarketplaceBody.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.tsx
@@ -257,6 +257,27 @@ export function AppListingsMarketplaceBody() {
       enabled: hasAppsStoreAccess(features),
       getNextPageParam: (lastPage) => lastPage.nextCursor,
       /**
+       * 🔴 60s. Without a `staleTime` react-query's default is ZERO, which means
+       * every remount and every window refocus refetches — and the `/apps` grid is
+       * a page viewers leave and come straight back to (open an app, hit back;
+       * tab away, tab back). Each of those was a fresh network round trip for a
+       * catalog that had not moved.
+       *
+       * WHY 60 AND NOT MORE: it is deliberately a THIRD of the server-side TTL
+       * (`CacheTTL.sm`, 180s, on `listAvailableListings`'s keyset query), so the
+       * client is never the longer of the two staleness sources. The moderator-
+       * visible worst case for a listing entering or leaving the store is bounded
+       * by the server side, which every listing-state mutation busts explicitly
+       * (`bustAppListingCatalogCache`) — a client window that outlived the server
+       * TTL would have made those busts invisible for the difference.
+       *
+       * It does NOT interact with `placeholderData` below: `staleTime` governs
+       * whether a cached entry for THIS key is refetched; `keepPreviousData`
+       * governs what is rendered while a DIFFERENT key loads. A sort/filter change
+       * is a new key, so it still fetches immediately regardless of this value.
+       */
+      staleTime: 60 * 1000,
+      /**
        * 🔴 KEEP THE PREVIOUS RESULT WHILE A NEW ONE LOADS. The query key is
        * `{kind, category, sort, limit}`, so every sort/filter change is a NEW
        * key — and without this `data` goes `undefined` for the duration of the
--- a/src/components/Apps/AppListingsMarketplaceBody.tsx
+++ b/src/components/Apps/AppListingsMarketplaceBody.tsx
@@ -263,9 +263,13 @@ export function AppListingsMarketplaceBody() {
        * tab away, tab back). Each of those was a fresh network round trip for a
        * catalog that had not moved.
        *
-       * WHY 60 AND NOT MORE: it is deliberately a THIRD of the server-side TTL
-       * (`CacheTTL.sm`, 180s, on `listAvailableListings`'s keyset query), so the
-       * client is never the longer of the two staleness sources. The moderator-
+       * WHY 60 AND NOT MORE: the PROPERTY is that the client window stays strictly
+       * shorter than the server-side TTL (`CacheTTL.sm`, 180s at the time of writing,
+       * on `listAvailableListings`'s keyset query), so the client is never the longer
+       * of the two staleness sources. That inequality — and only that — is what
+       * `AppListingsMarketplaceBody.keepPreviousData.browser.test.tsx` asserts, against
+       * the real constant. The specific 3× headroom is a judgement, not a pinned
+       * ratio: pick any value under `CacheTTL.sm`. The moderator-
        * visible worst case for a listing entering or leaving the store is bounded
        * by the server side, which every listing-state mutation busts explicitly
        * (`bustAppListingCatalogCache`) — a client window that outlived the server
--- a/src/pages/api/internal/blocks/build-callback.ts
+++ b/src/pages/api/internal/blocks/build-callback.ts
@@ -4,6 +4,7 @@ import type { Readable } from 'node:stream';
 import { withAxiom } from '@civitai/next-axiom';
 import { env } from '~/env/server';
 import { dbWrite } from '~/server/db/client';
+import { bustAppListingCatalogCache } from '~/server/services/blocks/app-listing.service';
 import { redis, REDIS_KEYS } from '~/server/redis/client';
 import { isAppBlocksPipelineEnabled } from '~/server/services/app-blocks-flag';
 import { setCommitStatus } from '~/server/services/blocks/forgejo.service';
@@ -323,7 +324,9 @@ export default withAxiom(async function handler(req: NextApiRequest, res: NextAp
   // window. A replayed success callback short-circuits here before triggering
   // another apply Job or touching commit status.
   if (!(await markApplyTriggered(body.appBlockId, body.sha))) {
-    res.status(200).json({ ok: true, applied: false, reason: 'duplicate callback (replay-guarded)' });
+    res
+      .status(200)
+      .json({ ok: true, applied: false, reason: 'duplicate callback (replay-guarded)' });
     return;
   }
 
@@ -413,6 +416,13 @@ async function watchApplyJobAndRecord(args: {
         where: { id: args.appBlockId },
         data: { currentVersionDeployedAt: new Date() },
       });
+      // 🔴 Catalog bust: `currentVersionDeployedAt` IS the onsite DEPLOY GATE in
+      // `listAvailableListings` (`al.kind <> 'onsite' OR ab.current_version_deployed_at IS
+      // NOT NULL`). This is the ONE non-tRPC writer of it, so without a bust here a
+      // freshly-deployed onsite app stays INVISIBLE in the store for the whole TTL —
+      // the deploy succeeds and the app simply is not there.
+      await bustAppListingCatalogCache().catch(() => undefined);
+
       await markRequestDeployState(args.slug, args.sha, 'live');
       await safe(setCommitStatus, {
         slug: args.slug,
--- a/src/server/services/block-registry.service.ts
+++ b/src/server/services/block-registry.service.ts
@@ -3504,9 +3504,15 @@ export class BlockRegistry {
     // `current_version_deployed_at`, via the deploy gate. A bust here was measured inert.
     //
     // If the catalog query ever grows a predicate over an `app_blocks` column this
-    // function writes, add the bust back — and add this function to the ledger in
+    // function writes, add the bust back — and add the row
+    // `'src/server/services/block-registry.service.ts::BlockRegistry::setMarketplaceMeta'`
+    // to `LEDGER` in
     // `~/server/services/blocks/__tests__/app-listing.catalog-bust-ledger.test.ts`, which
     // is what will fail and make that a deliberate decision rather than an omission.
+    // (That spelling is exact and verified: the ledger's attributor names class methods
+    // `Class::method`. An earlier version of it was anchored at column 0 and reported an
+    // added bust here as an unrelated top-level helper, which made this instruction
+    // unfollowable.)
 
     return {
       appBlockId: updated.id,
--- a/src/server/services/block-registry.service.ts
+++ b/src/server/services/block-registry.service.ts
@@ -1,7 +1,6 @@
 import { Prisma } from '@prisma/client';
 import { env } from '~/env/server';
 import { dbRead, dbWrite } from '~/server/db/client';
-import { bustAppListingCatalogCache } from '~/server/services/blocks/app-listing.service';
 import { redis, REDIS_KEYS, REDIS_SYS_KEYS, sysRedis } from '~/server/redis/client';
 import { manifestSettingsSchema } from '~/server/schema/blocks/manifest-settings.meta.schema';
 import { SLUG_REGEX } from '~/server/schema/blocks/publish-request.schema';
@@ -3495,8 +3494,19 @@ export class BlockRegistry {
         featuredOrder: true,
       },
     });
-    // Catalog bust: `category` is a FILTER AXIS of the cached page, and a mod sets it here.
-    await bustAppListingCatalogCache().catch(() => undefined);
+    // 🔴 NO CATALOG BUST HERE, DELIBERATELY — and the reason is not "category doesn't
+    // matter", it is that this writes the WRONG TABLE for that cache. The `/apps` catalog
+    // statement filters `app_listings.category`; this function writes
+    // `app_blocks.category`. The two are synced in ONE direction and at ONE moment —
+    // `approveRequest` copies the block's curated scalars onto the listing on approve, and
+    // busts there. Nothing in the cached statement reads any column this function writes
+    // (`category`, `featured`, `featuredOrder`); of `app_blocks` it reads only
+    // `current_version_deployed_at`, via the deploy gate. A bust here was measured inert.
+    //
+    // If the catalog query ever grows a predicate over an `app_blocks` column this
+    // function writes, add the bust back — and add this function to the ledger in
+    // `~/server/services/blocks/__tests__/app-listing.catalog-bust-ledger.test.ts`, which
+    // is what will fail and make that a deliberate decision rather than an omission.
 
     return {
       appBlockId: updated.id,
--- a/src/server/services/block-registry.service.ts
+++ b/src/server/services/block-registry.service.ts
@@ -1,6 +1,7 @@
 import { Prisma } from '@prisma/client';
 import { env } from '~/env/server';
 import { dbRead, dbWrite } from '~/server/db/client';
+import { bustAppListingCatalogCache } from '~/server/services/blocks/app-listing.service';
 import { redis, REDIS_KEYS, REDIS_SYS_KEYS, sysRedis } from '~/server/redis/client';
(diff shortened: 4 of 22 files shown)


ZacxDev:
## 🔴 Correcting this PR's own safety claim — the original body was WRONG, and a reviewer may have already believed it

Leaving the body as-published and correcting it here rather than editing it silently, because the wrong version is what any earlier reviewer read.

### What the body claimed

> Every axis this page varies on … is interpolated into that one statement, so **each is in the cache key by construction**. A hand-assembled key is a list someone has to keep complete; this one *cannot* omit an axis…

and, on the maturity axis:

> The SQL-hash key covers it **by construction**, with nothing enumerated.

### Why that is false

"Distinct statements ⇒ distinct keys" is only true if the hash is **injective**. It is not.

`queryCache` (`src/server/utils/cache-helpers.ts:111-113`) builds the key as
`[key, version, hashifyObject(query)].join(':')`. `hashifyObject` → `hashify`
(`src/utils/string-helpers.ts:277-285`) is a **32-bit** rolling hash:

```
hash = (hash << 5) - hash + chr;
hash |= 0;
```

32 bits, and **linear** — so a collision against a *chosen target* is constructed algebraically, not brute-forced.

That would still be academic if the hashed bytes were all server-controlled. They are not. `decodeListingCursor` (`src/server/services/blocks/app-listing.service.ts:119-150`) slices `cursorSortKey` and `cursorId` out of a lenient base64url decode as **arbitrary free strings** — only `cursorMean` is range-validated — and both are bound into the keyset statement. The router validates `cursor` only as `z.string().max(128)`. That is ample tuning room.

Verified independently: two statements differing **exactly** at the scope predicate (`TRUE` vs `al.kind = 'offsite'`) collide on the identical 32-bit hash with **six** tuning characters placed in the cursor.

Both viewer shapes are live and shared one module-scope cache instance:

* `app-listings.router.ts:1756-1763` → `scope: 'full'`
* `src/pages/api/v1/apps/index.ts:120-131` → `scope: 'public-external'`, the anonymous REST caller

And hydration (`findMany({ where: { id: { in: pageIds } } })`) re-applies **neither** the scope gate nor the maturity gate. So a constructed collision serves on-site listings into the public REST response — civitai#3983 reopened, arrived at through the cache instead of through a `??` — and the reverse direction is cache poisoning.

### What the fix actually guarantees (weaker, and true)

Pushed in `abf7e5b024`. `queryCache` is now constructed **per viewer class** inside `listAvailableListings`, with the two security-boundary axes as **literal, un-hashed key segments**:

```ts
const cacheable = queryCache(
  dbRead,
  `listAvailableAppListings:${scope}:${redCapable ? 'red' : 'sfw'}`,
  'v1'
);
```

The guarantee is now precisely this, and no more:

* **A hash collision can only ever mix two pages WITHIN one viewer class.** The class boundary — `scope` (public/onsite) and `redCapable` (maturity) — is no longer hash-dependent, because it is not in the hash at all.
* The remaining axes (`kind`, `category`, `sort`, `cursor`, `limit`) **stay in the 32-bit hash**, and are still collidable. That is accepted deliberately: they are ordinary correctness, not a disclosure boundary. The worst a constructed collision buys there is the attacker's own page served back to themselves under a different filter, inside their own viewer class.
* `hashify` itself is **untouched**. It keys caches, DOM ids and de-dup across the codebase, so widening it is a global blast radius; the containment belongs at the one call site that has a security boundary.

### The guard, and why the old one was not enough

The previous guard suite proved that distinct viewer shapes produce distinct **statements** — level (1). It never tested the statement→key step, which is exactly where the defect lived. Both levels are now asserted, and the new ones assert on the key **with its hash segment removed** — the half a collision cannot touch:

```
× 🔴 `scope` is a LITERAL key segment — a hash collision cannot cross the store-scope boundary
  AssertionError: the store SCOPE is not in the un-hashed part of the cache key, so `full` and
  `public-external` are separated ONLY by a 32-bit hash over a statement whose cursor bytes the
  caller supplies. That hash is constructible: a crafted cursor serves the full catalog (on-site
  apps included) to an anonymous /api/v1/apps caller. Put `scope` back into the `key` string
  passed to `queryCache`.
    expected 'listAvailableAppListings:v1' not to be 'listAvailableAppListings:v1'

× 🔴 every (scope × redCapable) viewer class has a distinct HASH-INDEPENDENT prefix
  AssertionError: two viewer classes share a hash-independent cache-key prefix, so only a 32-bit
  hash separates them: ["listAvailableAppListings:v1","listAvailableAppListings:v1",
  "listAvailableAppListings:v1","listAvailableAppListings:v1","listAvailableAppListings:v1"]
    expected 1 to be 5
```

Watched red with the axes moved back into the hash (4 of 13 red), green at HEAD (13/13). **The pre-existing derived-key cases stayed GREEN under that same mutation** — which is the whole point: they were necessary and never sufficient, and the body read them as sufficient.

Everything the body says about the *alternatives* (`fetchThroughCache` has no `tag`; `clearCacheByPattern` is banned; a generation counter multiplies the keyspace) stands unchanged.

One further correction while I am at it: the body and the code comment described the 180s TTL as "the backstop for … a redis bus failure". That is also false — `queryCache` has no try/catch and no fail-open (unlike `fetchThroughCache`), so a redis outage is a **500 on `/apps` and on `GET /api/v1/apps`**, not a degraded live read. The comment now states that dependency plainly instead of claiming a resilience the code does not have.

PR #4718, opened 2026-09-09
feat(rewards-abuse): gate clustering on per-account reward persistence

JustMaier:
## What

The nightly `rewards-abuse-prevention` job asks whether a set of accounts share an IP exclusively. It never asks whether any individual account behaves like a farm. Sharing an IP is the shape of a reward farm and also the shape of a household, so that single question cannot tell them apart.

This adds the second question, per **account**: over a configurable window, on how many days did this account reach its own daily cap for the award types being watched. Each account has its own cap, so an IP's total cannot answer it — a busy household looks busy at the IP while no individual pegs anything.

Three sysRedis options under `rewards:abuse-limits`, all no-ops by default:

| option | default | |
|---|---|---|
| `min_cap_days` | `0` | 0 disables the gate entirely — no CTE, no predicate |
| `cap_days_window` | `30` | bounded, so a mistyped window fails validation rather than the nightly run |
| `cap_day_awarded` | none | required once `min_cap_days` is set |

**Nothing here enables enforcement.** `dryRun` is untouched.

## The day window was wrong, and it had to be fixed for any of this to mean anything

`createdDate > subtractDays(now(), 1)` reads as "the last 24 hours" and is not. `createdDate` is a Date, so ClickHouse compares it at midnight, and at the 03:00 cron the predicate resolves to "createdDate is today" — a few hours of data, measured against thresholds that only mean something over a day. It also made a run's window depend on the hour it fired, so a hand-triggered run and a scheduled one were not the same job.

The clustering day is now the previous **complete** day. The cap-day aggregate had a quieter version of the same bug and is now a `BETWEEN` over complete days ending on the clustering day, so `cap_days_window: 30` means thirty whole days rather than twenty-nine and a fraction.

This is pre-existing and predates the gate. It is in this PR because picking a threshold against the old window would have been meaningless.

## Two things that are easy to undo by accident

**`max_user_count` must read a count the gate cannot shrink.** It is the one threshold that is an *upper* bound. Gate the column it compares and an IP too large to be a farm drops under the ceiling and gets flagged for the first time — the gate widens the net instead of narrowing it. With the ceiling reading an ungated count, the gated set is a strict subset of the ungated one.

**`user_ids` drives the disable write and is gated in its own right.** It used to be an ungated `array_agg`, correct only because the `ip_user_count = user_count` having-clause forced the two sets to coincide. Deleting that clause reads as a tidy-up once a persistence gate exists, and without this change it would quietly turn the disable list into every account that touched the IP — including the casual household earners the gate exists to protect.

## Verification

- Full unit suite green, `test:lint-rules` green, typecheck clean, lint clean (the two `no-explicit-any` warnings in the test file are pre-existing on `main`).
- The **emitted** query string — not a transcription — was dumped for all three config shapes, `EXPLAIN SYNTAX`'d, and both gated shapes executed read-only against production. The diff introduces a CTE referenced from inside `If`-aggregate arguments, `groupUniqArrayIf`, and integer/`Decimal` arithmetic inside an `if()`; none of that is obviously valid by inspection.
- After the window fix, the emitted query returns for the previous complete day exactly what a hand-written query for that date returned before it — the control on the change being the one intended.
- Mutation controls, each restore verified by `git hash-object`: gate forced off, gate moved into the outer `WHERE`, either branch's ceiling reverted to the gated count, `cap_day_awarded` given a default back, the multiplied ceiling flattened, the exclusivity having-clause deleted, the type filter dropped from the outer `WHERE`, the CTE measuring a different award family, and the day window reverted. All red. Three of those tests exist *because* the mutation initially passed.

## Review lanes

All five ran; the test lane ran alone.

| lane | outcome |
|---|---|
| correctness | Found the `max_user_count` widening. Fixed. Also flagged a non-finite literal path on a new numeric field — closed by `.int()`. |
| perf | No fix required. Corrected my cost claim: my timing was fully cache-served, so it is a warm floor. Confirmed the `time` bound is load-bearing, the CTE is evaluated once, and the type prefix uses the primary key. Its `.max()` suggestion is taken. |
| reuse | The gate is genuinely new — no existing helper answers cap attainment over a window. Its divergence findings are addressed in the schema comments; `status = 'capped'` was considered and rejected, see below. |
| intent | Confirmed scope, defaults and that enforcement is untouched. Found the mismatch between `cap_day_awarded`'s default and `award_types`' default, and a multiplier case where a cap-pegger scored no cap-days. Both fixed. |
| tests | Found three mutations the suite passed clean on, two on the disable path. All fixed; one of my fixes for it also passed its own mutation and had to be scoped again. |
| comment-review | Two comments were false, one of them made false by my own fix in this branch. Dated measurements removed. |
| docs-drift | No public doc drifted; the guard counts in `CLAUDE.md` are untouched and correct. Corrections landed in the private notes. |

**`status = 'capped'` was considered and not used.** A capped row means the account was *refused* after reaching the cap, not that it reached it — an account landing exactly on the cap and stopping writes none. It is also excluded by the existing `awardAmount > 0` filter, so it cannot be added as a second signal without changing that filter, and it would strand the thresholds this was modelled against.

**No typed audit columns were added** for the new options. They reach `rewards_abuse_decisions` inside the existing `config` JSON, and adding columns needs a hand-applied migration to land before the code or the writes vanish silently.

## `dryRun` now defaults to true

The safety feature in this PR is opt-in while its amplification is not: `min_cap_days` defaults to off, but the day-window fix widens what the **ungated** detector sees for everyone. A config written before this branch would therefore select a materially larger set, and if it also said `dryRun: false` it would act on all of it.

So the default is flipped. A config that is missing, half-written or reverted can no longer disable anyone; enforcing takes an explicit `dryRun: false`. The live config sets it explicitly, so nothing changes today. Five existing tests were relying on the old default to reach the destructive path and now say `dryRun: false` where they mean it.

## This is not sufficient for automatic enforcement, and should not be switched on as-is

An adversarial pass and a round of production validation both concluded that the persistence signal narrows the flagged population without changing its composition enough to act on automatically, and that single-day exclusivity is less stable than it looks. The signal that does separate the two populations is account shape, for which the job already has `user_conditions` on the destructive UPDATE.

The recommendation is to merge this, leave `dryRun` on, and treat the cluster list as input to moderator review rather than as an automatic action — which is what the original detector notes concluded before any of this was built. The evidence sits with the ticket and the private notes; it is deliberately not in this repo.

## Still to do after this merges

1. Set `min_cap_days` and `cap_day_awarded` in sysRedis — the gate does nothing until then.
2. A dry pass at that threshold, read back from the decision table rather than trusted from the response.
3. Deploy; the config is only meaningful against the fixed window.
4. Removing `dryRun` is a separate, deliberate decision.

Detection thresholds, the modelling behind them and the measured effect are deliberately kept out of this repo — see the private working notes.

### DIFF
--- a/src/server/jobs/__tests__/rewards-abuse-prevention.test.ts
+++ b/src/server/jobs/__tests__/rewards-abuse-prevention.test.ts
@@ -183,6 +183,16 @@ describe('rewards-abuse-prevention — detection shape', () => {
     expect(sql).not.toContain('AND startsWith');
   });
 
+  it('clusters over the previous COMPLETE day, not whatever is left of today', async () => {
+    await runWith({ award_types: ['dailyBoost'] });
+
+    // `createdDate > subtractDays(now(), 1)` reads as "the last 24 hours" and is not: createdDate
+    // is a Date, so the comparison lands on midnight and the predicate collapses to "today" —
+    // three hours of data at the 03:00 cron, against thresholds that assume a day.
+    expect(outerWhereOf(sqlOf())).toContain('createdDate = subtractDays(toDate(now()), 1)');
+    expect(outerWhereOf(sqlOf())).not.toContain('createdDate > subtractDays(now(), 1)');
+  });
+
   it('emits a cluster-size ceiling when max_user_count is set', async () => {
     await runWith({ max_user_count: 5 });
 
@@ -286,7 +296,7 @@ describe('rewards-abuse-prevention — scan bounds and config safety', () => {
     await runWith({});
 
     const sql = sqlOf();
-    expect(sql).toContain('createdDate > subtractDays(now(), 1)');
+    expect(sql).toContain('createdDate = subtractDays(toDate(now()), 1)');
     // `createdDate` is MATERIALIZED and prunes nothing; without this the scan reads the
     // whole table rather than the window it claims to.
     expect(sql).toContain('time > subtractDays(now(), 3)');
@@ -421,14 +431,18 @@ describe('rewards-abuse-prevention — per-account persistence gate', () => {
     const sql = sqlOf();
     expect(sql).toContain('GROUP BY toUserId, day');
     expect(sql).toContain('HAVING countIf(day_awarded >= day_cap) >= 10');
-    expect(sql).toContain('createdDate > subtractDays(now(), 30)');
+    expect(sql).toContain(
+      'createdDate BETWEEN subtractDays(toDate(now()), 30) AND subtractDays(toDate(now()), 1)'
+    );
   });
 
   it('bounds the wider window on the partition key as well as on createdDate', async () => {
     await runWith({ ...prefixConfig, min_cap_days: 10, cap_days_window: 14 });
 
     const sql = sqlOf();
-    expect(sql).toContain('createdDate > subtractDays(now(), 14)');
+    expect(sql).toContain(
+      'createdDate BETWEEN subtractDays(toDate(now()), 14) AND subtractDays(toDate(now()), 1)'
+    );
     // Why wider than the window: see DATE_BOUND_SLACK_DAYS.
     expect(sql).toContain('time > subtractDays(now(), 17)');
   });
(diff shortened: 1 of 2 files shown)


JustMaier:
Closing on Justin's call. The job this was improving has disabled nobody in six months, and the validation on this branch concluded its signal cannot be enforced automatically without hitting real users. Rather than keep a gate nobody will switch on, the job is being removed entirely and the problem retried from a different angle.

Branch tip `21641e8252` is kept for reference; the analysis lives with the ticket and the private notes.

PR #4719, opened 2026-09-09
feat(app-blocks): a host-side backstop when a block token is missing consent-gated scopes

ZacxDev:
feat(app-blocks): a host-side backstop when a block token is missing consent-gated scopes

The mint already computes everything needed to prompt a viewer for a withheld
scope, and then drops it on the floor.

`block-tokens` computes `missingScopes` + `needsConsent` from `partitionByConsent`,
and the comment beside it says the signal is "reported back to the host as
`needs_consent` so it can surface a re-consent prompt." It was not surfaced.
In `PageBlockHost` the value was a prop, a destructure, a comment, and then
`data-needs-consent={…}` — an attribute nothing reads. `git grep` returns exactly
one non-test hit: the line that writes it.

`missingScopes` DOES have a consumer (`resolveRequestConsent`). It is `needsConsent`
alone that dead-ends. So the consent modal opened only when a block PULLED it via
`REQUEST_CONSENT`, and an app that never asked left the viewer with a
working-looking control that could not succeed and no way to learn why.

Not theoretical. Measured on the live `playable-collections` app: a real 2-Buzz
tip was refused on both legs at the scope gate, the Buzz ledger confirms nothing
moved, and the viewer saw only "None of that tip came back confirmed" — no
prompt, no explanation, permanently. That viewer had no `app_user_scope_grants`
row, and `getGrantedScopes` is fail-closed, so every consent-gated scope was
withheld. `social:tip:self` was granted on ONE row in the whole grants table, so
this was the ORDINARY path a viewer took, not an edge case.

## What this adds

A dismissible, non-modal notice in the host's own chrome when `needsConsent` is
true, with a Review button that opens the existing `BlockConsentModal`. Zero app
changes; covers every block at once.

🔴 A NOTICE, NOT AN AUTO-OPENED MODAL, and that is a decision rather than a
shortcut. A block can be fully usable unconsented — `collections:read:self` is
consent-exempt, so `playable-collections` browses public collections fine with no
grant at all — and an unconditional modal would interrupt every viewer of every
app that merely REQUESTS a consent-gated scope.

Gated through `resolveRequestConsent`, the SAME predicate the block-initiated path
uses, so the two cannot disagree about when consent is offerable: `ready` plus a
non-empty missing set. The `dialogStore.trigger` call is extracted into one
`openConsentModal` used by both callers — a second inline copy would be the same
rule in two places and would drift on the first prop that changes.

Dismissal is per-mount and deliberately not persisted: the notice is the only
thing between "the app silently cannot do the thing" and the viewer, so a
remembered dismissal would make the hole permanent for exactly the people who
have already met it once.

## Verification

- `--project component src/components/AppBlocks/` — 43 files / 532 tests pass.
  Wider than the changed file on purpose: the default `PageBlockHost` fixture
  already carries `needsConsent: true`, so a notice that rendered wrongly would
  have disturbed neighbouring host suites.
- RED/GREEN watched, not assumed. With the notice's render condition forced
  false, 3 of the 5 new tests fail by name — the three that assert it EXISTS.
  The other two assert ABSENCE (nothing missing; not-yet-ready) and correctly
  stay green under that mutant: they are CONTROLS, not regression coverage, and
  are labelled as such.
- `typecheck` — the changed file is clean. 🔴 The repo-wide count in a fresh
  worktree is ~6450 because the Prisma client is not generated there (1490
  `typeof Prisma` errors, no `.prisma/client`), so the meaningful reading is the
  DELTA: 6452 → 6450 across this change, i.e. it introduced none and removed the
  two it had. Those two were real — the first version passed `Status` where
  `HostStatus` was wanted; it now goes through `toHostGateStatus`, the existing
  converter the message handlers use, because this component's `Status` carries
  an extra `'error'` member. A cast would have compiled and silently disagreed
  with every other gate.

## Not in scope

- Which scopes are consent-gated. `CONSENT_EXEMPT_SCOPES` is fail-closed and
  correct; untouched.
- The app-initiated `REQUEST_CONSENT` path — strictly additive beside it.
- The app-side fix for the app that surfaced this, which asks for its own money
  scopes at press time. That is better UX than any host-level prompt and shipped
  separately; this is the backstop for apps that do not.

## Not verified

Visual review. The notice reuses Mantine primitives already imported here and the
suite asserts structure and behaviour, but nobody has looked at it rendered.


ZacxDev:
## Adversarial audit — round 1

Round 1 ran against `1582766528`. **No 🔴.** Six 🟡 and four 🟢. Fixed in `7f0348d96d`: three 🟡 and four 🟢. Three 🟡 are deliberately left — they are judgement calls for a reviewer, listed below rather than quietly dropped.

### Fixed

1. **The dismissal latch was keyed on nothing** — and it re-opened the hole this PR exists to close. `PageBlockHost` is not remounted between two `/apps/run/<slug>` pages (no `key` in `_app.tsx`, none at the call site), and the chrome this component draws is itself that one-click path. So dismissing app A's notice suppressed app B's for the rest of the SPA session, for apps the viewer had never seen it for — while the comment beside it justified non-persistence by saying a remembered dismissal "would make the hole permanent again for exactly the people who have already met it once". Now keyed on `appBlockId`.
2. **No `reviewMode` gate.** The block-initiated path calls that guard absolute; the notice had none. Unreachable in production only because `ReviewBlockPreviewHost` hardcodes `missingScopes={[]}` — a caller-side accident, not a guard, and untested.
3. **Two clicks stacked two consent modals.** `dialogStore.trigger` dedupes on `id` alone and fell back to `Date.now()`. Latent until this PR added the first human-clickable trigger. Stable id.
4. **`needsConsent &&` had no killing mutation**, and the negative control varied two conditions at once so it could attribute neither. Split into two controls; the term is now killable. The new case pins a state the mint never produces and says so.
5. **A comment claimed the `data-needs-consent` attribute "stays for debugging and for tests".** No test reads it. Corrected, naming the false half.
6. **A pre-existing comment had been split mid-sentence** by the `openConsentModal` insertion. Opener moved above it.
7. **Prettier** — both changed files pass `--check`.

### Verification

- `--project component src/components/AppBlocks/` → **43 files / 536 tests** (was 532).
- All four new guards mutation-killed, each by exactly one test, by name.
- Typecheck: changed files clean.

### 🔴 Correcting this PR's own body

It said the worktree install "completed". It did not — `pnpm install --frozen-lockfile` **fails at the Prisma postinstall on NixOS** (404 fetching the engine checksum). The test counts stand, because the component tier does not need a generated client, but the install was reported clean because a trailing shell command swallowed pnpm's status. Copying a generated `node_modules/.prisma` in is what makes the tier runnable, and it drops the repo-wide typecheck from ~6450 to ~2869 — confirming the missing client was the bulk of that number.

### Not addressed — reviewer's call, recorded so they are open rather than absent

- **Height.** The notice is ~89px at phone width against `FILL_MIN_HEIGHT_PX = 300`. Nothing clips, but it lands in the case that floor exists for, and the geometry tier pins `needsConsent: false`, so no tier can see it.
- **No kill switch.** Every logged-in viewer of an app declaring a consent-gated scope sees this on first load. The only lever kills the whole App Blocks surface. Rollback is a revert plus a release.
- **`IframeHost` has the identical hole** on the live `model.sidebar_top` surface and is untouched. It should have been in the original "not in scope" list; it now is.
- **Visual review** — still nobody has looked at it rendered.

### Audit hygiene

The auditor disclosed two false readings of its own before the one it reported: symlinking the base clone's `node_modules` produced duplicate-React failures, and a plain install produced `Tests no tests`. Only the third run — a real install plus a copied `.prisma` — matched. It also noted the `component` tier is **ungated in CI** (`.github/workflows/lint.yml` says so explicitly), so these tests block nothing automatically.

```audit-claims round=1 audited=1582766528ca7be62e681c106d5852a2b3cd9ff4..7f0348d9
1. Dismissal is now keyed on appBlockId, so dismissing one app's notice no longer suppresses another's across an SPA navigation.
2. The notice is gated on !reviewMode, matching the absolute guard the block-initiated REQUEST_CONSENT path already applies.
3. openConsentModal passes a stable dialog id, so repeated clicks open one modal rather than stacking.
4. The needsConsent term is now killable: the negative control was split so each of needsConsent and missingScopes is varied alone.
5. The data-needs-consent comment no longer claims a test reads it; no test does.
6. The lazy-consent comment split by the openConsentModal insertion is contiguous again.
7. Both changed files pass prettier --check.
```

ZacxDev:
## Adversarial audit — round 2, and the ladder STOPS here

Round 2 ran against `7f0348d96d`. **No 🔴, and no defect in shipped behaviour.** Two 🟡 and two 🟢 — and **all four were sentences the round-1 fix wrote while fixing round 1's findings.** Fixed in `5a33876381`.

### What it found

1. **🟡 The `rerender` claim was false.** Round 1 added a bespoke harness justified by "vitest-browser-react's handle exposes no `rerender`". It does — and `useBlockToken.browser.test.tsx` already drives this exact app-A→app-B soft-nav with it. The real cause was a missing `await`. Left alone, that comment would have taught the next author to hand-roll harnesses against the repo's convention. Now uses `rerender`; harness gone.
2. **🟡 `!reviewMode` is STRICTER than the path it claimed to mirror.** That path's absolute rule is *never a MODAL*, and it deliberately emits a PASSIVE notice because silence "reads as 'this app is broken'". Suppressing the notice entirely re-creates that silence. **Behaviour kept** — unreachable today, safety property holds either way — but the comment no longer claims a mirror, and it names the correct shape (notice without the button) for whoever threads those props from `mintData`.
3. **🟢 One split control was an unlabelled fiction.** Round 1 flagged one half as unreachable and left its equally-unreachable sibling unflagged. Both labelled.
4. **🟢 The environment claim in this PR's body is wrong.** `pnpm install --frozen-lockfile` does not fail "on NixOS" — it fails **outside the flake dev shell**. Inside it: `✔ Generated Prisma Client`, `Done in 40.6s`, rc 0. The derived "~2869 type errors is normal" was an artefact of that half-populated tree; the real floor is **10**, none in the changed files. Re-verified here, not taken from the audit.

### The sweep

The recurring shape is *a claim asserted without checking*, so it was swept at every site rather than the reported one:

- **Two measurements had been restated as mine** when they were the round-1 auditor's ("two clicks 30ms apart"; the two-armed dismissal control). Both now attributed, and the comment says which arm the committed test carries and which the mutation covers.
- **Three structural facts the fix rests on** had been taken from the audit on trust. Re-verified and cited: `_app.tsx:175` renders `<Component>` unkeyed; the run page renders `<PageBlockHost` unkeyed; `IframeHost.tsx:642` links "Recently run" to `/apps/run/<id>`. All three hold.

### Verification

43 files / **536 tests** inside the flake shell · all four guards **re-mutated after the refactor** (the per-app test changed shape, so its discrimination is re-proven, not assumed) — each 1 failed / 40 passed, each naming its own test · typecheck 10 repo-wide, **0** in the changed files · prettier clean.

### 🔴 Why the ladder stops at round 2, stated rather than left implicit

The findings-keyed rule would run a round 3, because round 2 produced findings that needed fixing. I am stopping instead, on the prose-payload criterion, and the reason has to be written down or a stopped ladder is indistinguishable from a converged one:

- **No 🔴, in either round.**
- **Blast radius is "the document contains a false sentence."** Round 2 changed no shipped behaviour — its one behavioural finding was kept as-is and only its comment corrected.
- **The recurring shape was swept at every site**, not just the reported ones (above).
- **The rounds will not stop on their own.** Round 2's findings were entirely prose round 1 wrote; round 3 would audit prose round 2 wrote. Every fix round composes new explanation, so "fixed a defect" and "reworded a warning" are the same edit here — and the payload-attribution gate cannot fire, because the comments legitimately live in the payload file.

**Recorded as OPEN, not absent** — three reviewer's-call items carried unchanged from round 1: the notice's height against `FILL_MIN_HEIGHT_PX` (invisible to the geometry tier, which pins `needsConsent: false`); **no kill switch** on UI that greets every logged-in viewer of an app declaring a consent-gated scope; and **`IframeHost` has the identical hole** on the live `model.sidebar_top` surface. Plus: **nobody has looked at this rendered.**

```audit-claims round=2 audited=7f0348d96da5b7204130b6e5713c0858fea4cb44..5a338763
1. The TwoApps harness is replaced by the repo's rerender convention, and the false "no rerender" claim is gone.
2. The !reviewMode comment no longer claims to mirror the block path; it states that it is stricter, why that is accepted, and what shape to use if the review props are ever threaded.
3. Both halves of the split negative control are labelled as states the mint never produces.
4. The environment claim is corrected: install fails outside the flake dev shell, not on NixOS; the typecheck floor is 10, not ~2869.
5. Two measurements borrowed from the round-1 audit are attributed rather than restated as the author's own.
6. Three structural facts (_app.tsx unkeyed Component, unkeyed PageBlockHost, IframeHost Recently-run links) are re-verified independently and cited.
```

